PR #2441, opened 2014-04-29
Pipeline GUI Tool

tomspilman:
This is the first commit of the Pipeline GUI tool for nice visual editing of the new MonoGame content pipeline command files.  The tool looks like this:

pipeline

It currently has the following basic features which makes it usable immediately:
- Loading, saving, and creating new pipeline project files.
- Adding and removing existing content to the project.
- Change importers and processors as well as its properties.
- Build, rebuild, and clean content.
- No dependence on Microsoft XNA or Visual Studio.

It was built using C# and WinForms specifically for Windows support.  We have a simplified MVC design intended to allow someone to eventually implement a new IView for MacOS and Linux.

Our next steps here are to get feedback from users, iterate on the core features, and improve the functionality.  A few things on our roadmap for the tool:
- Editing custom processor assembly lists.
- System for per-platform content/settings.
- Support for linked content files.
- Better property editors.
- Undo/redo stack.
- Actual help docs.
- Recent file lists.
- Proper unloading of processor assemblies.
- Adding tool to MG installer and registering file extensions.
- Triggering .mgcb building directly from Visual Studio build.
- Import existing .contentproj files.

Thanks to @jamesford42 for making all the treeview and property grid work correctly.

### DIFF
--- a/default.build
+++ b/default.build
@@ -176,6 +176,8 @@
       <exec program="msbuild " commandline="Tools\2MGFX\2MGFX.sln /t:Build /p:Configuration=Release" />
       <exec program="msbuild " commandline="Tools\MGCB\MGCB.sln /t:Clean /p:Configuration=Release" />
       <exec program="msbuild " commandline="Tools\MGCB\MGCB.sln /t:Build /p:Configuration=Release" />
+      <exec program="msbuild " commandline="Tools\Pipeline\Pipeline.sln /t:Clean /p:Configuration=Release" />
+      <exec program="msbuild " commandline="Tools\Pipeline\Pipeline.sln /t:Build /p:Configuration=Release" />
     </if>
   </target>
 


jarnoburger:
Situation:
When you have pressed the Build button.. 
You see the 'build-log" in the textbox on the right.. 

Issue:
I am missing info about which keys i could use to get to my content.

Possible solution :
- if i would build content in Content\Fonts\Candera.SpriteFont , that the build log would say : You now have a <Font> reachable at Fonts\Candera.. 
- string.format("You know have a <{0}> reachable at {1}", resourceTypename, location);

tomspilman:
> I am missing info about which keys i could use to get to my content.

That is a problem that documentation should solve.

> the build log would say : You now have a reachable at Fonts\Candera.. 

We don't want to clutter up the build log with that.

cra0zy:
I've been playing around with Xwt yesterday and today, and it's really good. Also it's naming scheme and the way everything works is similar to Gtk, so for someone like me it makes it very easy to use.

tomspilman:
> Xwt yesterday and today, and it's really good. Also it's
> naming scheme and the way everything works is similar to Gtk

Because it was originally designed by Xamarin to replace their usage of Gtk.  However i've been told Xamarin are moving away from Xwt themselves... this is why we decided on Gtk.

cra0zy:
> Because it was originally designed by Xamarin to replace their usage of Gtk. However i've been told Xamarin are moving away from Xwt themselves... this is why we decided on Gtk.

Even tho Xamarin decided to abandon it, the project itself is still quite active.

Is the OS X Pipeline going to stick with Gtk, or are there some plans for Cocoa?

tomspilman:
> Is the OS X Pipeline going to stick with Gtk

Absolutely... the last thing we want is a 3rd GUI toolkit to maintain if we can avoid it.

KonajuGames:
We started the Mac and Linux Pipeline Tool with Xwt but found it lacking in
features and documentation, and had a crash that we couldn't figure out at
that time. It also suffers the same affliction as most other cross platform
GUI toolkits: it just doesn't look or behave like a native application. On
Linux it is a wrapper around Gtk that has less features than Gtk. On Mac it
doesn't expose enough of Cocoa to implement everything we need in the tool.

cra0zy:
> We started the Mac and Linux Pipeline Tool with Xwt but found it lacking in
> features and documentation

Agreed on documentation, but... *caugh* https://github.com/cra0zy/MonoGame/tree/xwtpipeline *caugh*.... people do tell me that I am stubborn.... try it out. It's only Windows and Linux version, I don't own a Mac, so can't really build for it.

The only thing I found missing feature wise are accels for MenuItems.

Also I won't setup any PRs with it, I just built it to prove that Xwt was a viable way of doing it.

PR #2516, opened 2014-05-11
Resumer improvement

KakCAT:
Hi!

this tries to implement https://github.com/mono/MonoGame/issues/2463

### DIFF
--- a/MonoGame.Framework/Android/IResumeManager.cs
+++ b/MonoGame.Framework/Android/IResumeManager.cs
@@ -59,5 +59,13 @@ namespace Microsoft.Xna.Framework
         /// Called whilst the game is resuming. Draw something to the screen here.
         /// </summary>
         void Draw();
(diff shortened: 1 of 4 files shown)


tomspilman:
@dellis1972 @KonajuGames - We never got anyone to test this independently.  Is any of this of value 2 years later?

KakCAT:
This PR was submitted because at the time, each time the app was deactivated it had to reload all the assets.  A few months later Xamarin fixed something and https://github.com/mono/MonoGame/pull/2753 was commited, making this PR obsolete... unless Android does something weird with the app like saving the app status but unloading the assets, which would result in a very slow resume like in the old days. As AFAIK nobody has reported this problem in 2 years I think it's safe to just discard this PR.

tomspilman:
Thanks for the explanation @KakCAT !

PR #2590, opened 2014-05-20
Model importer and processor improvements

### LINKED ISSUE #2182: FBX importer does not work
We have tried FBX files with the following properties:
- 2014, binary format
- 2011, ASCII format
- with bones
- without bones
- with animations
- without animations

So far, none of these have imported successfully. There is a crash on line 201 of ModelReader:

```
                mesh.Name = name;
                mesh.ParentBone = bones[parentBoneIndex]; /// here
                mesh.ParentBone.AddMesh(mesh);
```

Because the parentBoneIndex is -1.  I modified the code to map based on the bone name instead of the index, which got past this point, but the model then half rendered and vertices were missing from the mesh.

Is there a specification on what forms of FBX are supported in the MonoGame content pipeline?

JamesLupiani:
I've addressed some bugs with model importing and processing, now the output looks much closer to XNA. It can properly build the dude and ship assets from XNA samples, so long as the source files are updated to the newer FBX versions first. This should also resolve #2182, which had a missing triangles problem.

I also filled in some of the gaps for material processing necessary to get things building and models written out, so the ModelProcessor properly chains to build materials and textures.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/ModelProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/ModelProcessor.cs
@@ -15,8 +15,6 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
     [ContentProcessor(DisplayName = "Model - MonoGame")]
     public class ModelProcessor : ContentProcessor<NodeContent, ModelContent>
     {
-        private readonly Dictionary<MaterialContent, MaterialContent> _processedMaterials = new Dictionary<MaterialContent, MaterialContent>();
-
         private ContentIdentity _identity;
         private ContentBuildLogger _logger;
 


nkast:
https://github.com/mono/MonoGame/blob/40347ddcc849117773677dc7eba96e6d37f2b023/MonoGame.Framework.Content.Pipeline/Processors/ModelProcessor.cs#L230-L232 

This code add a white Color channel at every generated Model. Not only this waste bytes but also cause problem when I expect to read a VertexPositionNormalTexture under XNA with GetData() but the underling structure is VertexPositionNormalTextureColor when importing with MG.

We never show XNA add a white color channel, so this is probably a bug of XNA & some specific 3D program. Under lightwave we were able to generate/export fbx with or without color channels.

Can we reverse this? Maybe add a switch 'GenerateWhiteVertexChannel' in the ModelProcessor for the case of XNA ported codebase with models/FBX that cause this error while leave all other models unaffected.

DDReaper:
Interesting, now that this has been merged, will have to try the Custom model Importer through MGCB again.  Last time it was reporting issues reading Material content (#4187)

nkast:
So, this has been moved here since then: https://github.com/mono/MonoGame/blob/40347ddcc849117773677dc7eba96e6d37f2b023/MonoGame.Framework.Content.Pipeline/Processors/ModelProcessor.cs#L230-L232

I wish to post a PR to remove this.
@JamesLupiani, is that ok?
I have never seen XNA put a color channel, while Monogame currently seems to add a color channel to every model. Not to mention that all Vertex definitions are incompatible with this. Instead of  VertexPositionTexture, VertexPositionNormalTexture,etc we would have VertexPositionTextureColor, VertexPositionNormalTextureColor, etc. 
When the FBX had a color channel it usually comes as VertexPositionColorTexture.

JamesLupiani:
Hmm. Are you saying it's wrong, or that the behavior is different with Lightwave? I mostly tested with the XNA sample assets and some 3DS Max models we had, but I'd say the ModelProcessor was the least developed portion. I think I still had some unmerged fixes myself https://github.com/JamesLupiani/MonoGame/commit/fbc0f56287cb23c25ccd9b2c18580089e362542a.

I'd say if XNA puts them in with the sample assets, it should probably be the default behavior until it's time to break compatibility. If they aren't there even with the old samples it should definitely be removed. Getting the wrong vertex format is an error.

On the other hand, if some cases do add them or if nobody's willing to test, the switch is the way to go.

nkast:
I just tested the tank.fbx from the 'CameraShake' xna catalog sample and the vertexDeclaration is Position/Normal/Texture. I have never seen the behavior you describe with XNA but there can be only two cases. (A) XNA misinterpreted the .fbx samples as having vertexcolor or (b) the .fbx indeed had a vertexcolor and assimp failed to export it.

For (A) I think is wrong to propagate the bug on every MG model from now on. If there is legacy code that need to work on XNA I believe there are ways to workaround it. (re-exporting/modifying the .fbx, using MG XNBs,)
For (B), well there is no use to replace the channel with #FFFFFF, again the FBX must me updated or fixing assimp.

On every other case it waste memory/bandwidth and also breaks XNA code like: `GetData<VertexPositionNormalTexture>()`.

nkast:
So, is it ok with you to remove this? Will it cause any problems within your projects?

JamesLupiani:
You can remove it as far as I'm concerned. I'm not actively able to participate at this time. Due diligence was all I was looking for.

PR #2613, opened 2014-05-24
Issue2612 bugfix

thefex:
https://github.com/mono/MonoGame/issues/2612

Looks like it fix crashes after restart (deacitvation/activation) on windows phone.

edit: To simulate crash debug on device (didn't test on emulator), set breakpoint in GraphicsDevice.DirectX.UpdateDevice(..,..), go back to home screen and back to game few times ( should crash after SharpDx.Dispose(ref _d3dContext) )


tomspilman:
Closing this as it never solved the original issue and we're moving forward to Windows 10 now.

PR #2621, opened 2014-05-25
IntermediateSerializer Skeleton

tomspilman:
This PR adds the initial skeleton of the various IntermediateSerializer classes as described from the XNA docs.  The implementation does not work yet.

I'm also starting to port over the work @JamesLupiani  did in the XmlImporter (see https://github.com/mono/MonoGame/pull/2380) which basically implements what the IntermediateSerializer classes need to do.

Eventually we should be able to replace the current `XmlImporter` implementation with a single call to `IntermediateSerializer.Deserialize()`.  This is in fact how i'm testing that this works.

My next step is to implement the overloaded `ContentTypeSerializer` which is what seems to drive the entire XML serialization process.  This looks like a very similar design to how the `ContentTypeWriter`/`ContentTypeReader` system works.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/IntermediateReader.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/IntermediateReader.cs
@@ -3,14 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
-using System.Collections;
-using System.Collections.Generic;
-using System.Globalization;
-using System.Linq;
-using System.Reflection;
 using System.Xml;
-using System.Xml.Linq;
-using Microsoft.Xna.Framework.Utilities;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
 {


tgjones:
I've just run into this issue - the methods in `IntermediateWriter` are not yet implemented.

> I included a bunch of test cases from Sean Hargreaves' examples with the XmlImporter PR. We should be able to run all of the test cases in Test\ContentPipeline\AssetTestClasses.cs through the IntermediateSerializer and get XML output matching the ones in \Test\Assets\Xml. Once those can make a round trip read-and-write it should be good to go.

Is that the next step? To be able to serialize the test cases in `AssetTestClasses.cs`, and get XML output matching the ones in `Test/Assets/Xml`? If so, I'll have a go at this.

tomspilman:
> Is that the next step?

I think that is about correct.

Note we already have the tests for `IntermediateReader`...
https://github.com/mono/MonoGame/blob/develop/Test/ContentPipeline/IntermediateDeserializerTest.cs
The `IntermediateWriter` tests will be different of course.

PR #2651, opened 2014-05-31
Render faster if possible on fixed time step

hach-que:
Currently when MonoGame is using a fixed time step, it won't ever render faster than 60FPS, despite the GPUs capability to do so.

This modifies the tick so that if SynchronizeWithVerticalRetrace is off, it will continue to render frames until the time is used up (instead of sleeping).

This is useful for diagnosing whether rendering is responsible for causing slow frames by giving an accurate representation of how fast the current frame can be rendered.


hach-que:
Closing off this PR so I can resubmit it from a new repository, and with the changes @tomspilman requested.

PR #2691, opened 2014-06-09
GraphicsDeviceManager.ApplyChanges changing window size.

jamesford42:
GraphicsDeviceManager.ApplyChanges should not be triggering the window to resize, this is only a backbuffer resize operation.

In standard XNA one can hook into Window.OnClientBoundsChanged and initiate a backbuffer resize with GraphicsDeviceManager.ApplyChanges. But in MonoGame this triggers an infinite loop do to this line.


tomspilman:
All this code has been refactored since then... closing.

PR #2702, opened 2014-06-10
PipelineTool PropertyGrid Usability

jamesford42:
ContentItem.Processor property is expanded by default in the property grid.

When PropertyGrid updates in response to a selection change:
- selected property / griditem is synchronized
- expanded griditems are synchronized

Selection.Modified signature changed.


tomspilman:
The property grid is totally different now... closing.

PR #2842, opened 2014-07-22
MacOS: use common OpenGL/AL codepath (1/2)

thefiddler:
This PR switches MacOS from MonoMac.OpenGL/AL to the corresponding OpenTK namespaces. This allows us to use a common codepath for the WinGL/Linux/MacOS platforms, which reduces the support burden (see comments in #2782.)

Changes:
- `MonoMac.OpenGL/AL` namespaces converted to `OpenTK.Graphics.OpenGL` and `OpenTK.Audio.OpenAL` respectively.
- `#if MONOMAC` regions regarding API differences between MonoMac and OpenTK are eliminated.
- `MacGamePlatform` and `GameWindow` are not affected - we are still using MonoMac for platform support.
- Unit tests remain unaffected (87/166 passing on my system.)

This PR does not introduce any functional differences and should be safe to apply as-is. I'll issue a separate PR for switching `MacGamePlatform` to `OpenTKGamePlatform`, as that is much more invasive change and should be reviewed separately.

### DIFF
--- a/Build/Projects/FrameworkReferences.definition
+++ b/Build/Projects/FrameworkReferences.definition
@@ -23,6 +23,9 @@
     <Binary
       Name="Tao.Sdl"
       Path="ThirdParty/GamepadConfig/Tao.Sdl.dll" />
+    <Binary
+      Name="OpenTK"
+      Path="ThirdParty/Dependencies/OpenTK.dll" />
   </Platform>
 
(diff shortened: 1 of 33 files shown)


cra0zy:
@tomspilman This can be closed, DesktopGL already made a common code path, plus SDL branch is in it's works.

PR #2849, opened 2014-07-23
MacOS: use common OpenGL/AL codepath (2/2)

thefiddler:
This PR switches MacOS from MacGamePlatform to OpenTKGamePlatform, which is shared with WindowsGL and Linux. It the second part of #2842.

This PR requires OpenTK >= 1.1.4c (https://github.com/Mono-Game/MonoGame.Dependencies/pull/23). Previous versions will work but will face stability issues when running the unit tests.

Current status: 149/166 unit tests passing. We are currently failing all 8 shader tests (cannot find content) and 3d texture tests (not implemented.)

I have also temporarily disabled VideoPlayer on MacOS, as that was tightly bound to MacGamePlatform.

Questions:
- WindowsGL somehow executes more tests than MacOS. Any idea how to enable the missing tests?
- Is there a plan for VideoPlayer on Linux/WindowsGL (libav or ffmpeg or something similar)?
- If so, would it make sense to use the same approach for MacOS, or do we prefer to keep a separate QT implementation there?

Feedback and testing appreciated!


cra0zy:
@tomspilman This can be closed, DesktopGL already made a common code path, plus SDL branch is in it's works.

cra0zy:
@tomspilman You skipped the 2nd one.

tomspilman:
Got confused with the message... looked like the same one. :)

PR #2900, opened 2014-08-09
Display orientation can be landscape or portrait on WP8

aberus:
Both properties 
<code class='language-cs'>GraphicsDevice.PresentationParameters.DisplayOrientation</code> and
<code class='language-cs'>TouchPanel.DisplayOrientation</code> 
always return value <code class='language-cs'>Portrait</code> even if we are in landscape mode.

<code class='language-cs'>Window.CurrentOrientation</code> returns correct value.


tomspilman:
@nkast @Nezz - So looks like we dropped the ball on this... is this of value still almost 2 years later?

aberus:
I agree, I'm closing pull request.

PR #2998, opened 2014-09-18
Vertex texture fetch, skeleton meshes without animations, Pipeline Tool changes

DissidentDan:
*Adding support for vertex texture fetch to Windows (Direct3D 11) target. This code should theoretically also work for OpenGL, but it cannot be tested due to the currently-used version of MojoShader not supporting texture fetch instructions in vertex shaders.

*Support for importing skeleton meshes that don't have animations in the file. This is useful for having separate files for the mesh and the animations

*Adding functionality to the Pipeline Tool to add all files from a folder in one step.

*Pipeline Tool shows exception info from MGCB in the output pane

*Propagating some fullscreen flags further down the line (as fullscreen support is not fully implemented, you may want to skip this one)

*You may also want to skip the changes to the references for the Yeti media projects. I found them necessary to compile, iirc.


tomspilman:
This one is tricky to deal with because of all the mixed features... even worse than we're looking at it over a year an a half later.

Most of the stuff has been addressed or is no longer used.


> Adding support for vertex texture fetch to Windows (Direct3D 11) target.

This was handled in https://github.com/mono/MonoGame/pull/3608.


> Support for importing skeleton meshes that don't have animations in the file. This is useful for having separate files for the mesh and the animations

@mgarstenauer  - Do you think the changes to the OpenAssetImporter.cs here are something we should extract and resubmit?  Or did we already cover this in some other PR?

mgarstenauer:
The OpenAssetImporter changes were covered in #3336. I also made a quick test using a file without animations: The skeleton is imported correctly.

tomspilman:
Thanks @mgarstenauer !

PR #3008, opened 2014-09-22
Set SamplerState.GraphicsDevice to null on reset

thefiddler:
The predefined SamplerState are statically initialized on first access
and their GraphicsDevice property set when Activate() is called.
However, SamplerState.GraphicsDevice is never cleared when the
GraphicsDevice is disposed. This leads to an assertion when a
SamplerState is re-activated on a new GraphicsDevice.

This PR adds a static SamplerState.PlatformReset() method that is
invoked when a GraphicsDevice is disposed, allowing SamplerStates to be
re-initialized when activated on a new GraphicsDevice.


tgjones:
This will be fixed as part of #3472.

tomspilman:
@tgjones - I don't mind closing this PR and depending on your fix in #3472.

PR #3009, opened 2014-09-22
Adding support for 'Asset Name'.

jamesford42:
Content items now have a property 'Asset Name': the name given to the output/copied file.
Importing a contentproj will now import the 'Asset Name' of content items.
Syntax changed from "/build sourcefile" to "/build sourcefile;assetName". /copy changed in same way.
/copy items now get cleaned like regular items.
/copy items now get logged to build output like regular items.
Number of skipped items is now logged alongside success/fail counts.
/launchdebugger option will now break as soon as the option is set rather than after parsing the entire file.


KonajuGames:
@tomspilman What were the changes you wanted to make to this?

@jamesford42 This will need latest from develop merged into it.

tomspilman:
I'll get this fixed up here soon.... sorry I let it slip.

KakCAT:
Hello,

is this PR superseded with another PR?

tomspilman:
> is this PR superseded with another PR?

No... I just forgot about it.  I'll see if I can get to it this weekend.

jamesford42:
I'm really not sure what these 'snapshot dependency' errors are.

tomspilman:
> I'm really not sure what these 'snapshot dependency' errors are.

You stopped reading too soon.  The error is "Snapshot dependency failed to start: ... Test Windows".  This means: This build failed because Test Windows failed.   So if you look at the failing Test Windows build you see this in the output log:

'''
Failed to build patch for build #3.6.0.1357 {build id=32697, buildTypeId=MonoGame_TestWindows}, VCS root: "Develop" {instance id=5, parent internal id=1, parent id=MonoGame_Develop, description: "https://github.com/MonoGame/MonoGame.git#develop"}, due to error: 'build patch' command failed.
stderr: The repository 'https://github.com/MonoGame/MonoGame.git' has a submodule in the '5a653afcb3b4d2b8f450ce2997112d1479a9f9f9' commit at the 'PSVita' path, but has no entry for this path in .gitmodules configuration
exit code: 1
'''

My guess is whatever you did in this commit:

https://github.com/MonoGame/MonoGame/pull/3009/commits/ecd6ebb6d1d118b16b4dffb585c17642b12f8db9

Screwed up your branch.

IMO you should create a new branch and copy over your changes manually and resubmit this PR.

Jjagg:
> My guess is whatever you did in this commit:
ecd6ebb
Screwed up your branch.

You can probably just [revert that commit](https://help.github.com/desktop/guides/contributing/reverting-a-commit/).

jamesford42:
Actually that commit FIXED the build. At least the "Build Windows" and "Build Mac" parts of it. Those are old submodules from really old versions of monogame (and the psvita folder should have never been a submodule). I don't know how exactly they got added as submodules in my feature branch, but for sure none of them should actually be submodules of monogame in the develop branch, right?

tomspilman:
>  but for sure none of them should actually be submodules of monogame in the develop branch, right?

Yes...  Vita should not be a submodule on the public repo.

As I said before I think your branch here is screwed up enough now that your best bet is to recreate it.

MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
```diff
@@ -162,6 +168,9 @@ public bool NeedsRebuild(PipelineManager manager, PipelineBuildEvent cachedEvent
             if (cachedEvent == null)
                 return true;
 
+            // jcf: Not sure if we need to deal with 'file exists in output but with different case'
+            //      here, or if that should be done elsewhere.
+
```
tomspilman: Yes I think we should.  This should handle all cases of a file being dirty.

Now you could argue a different case is actually a different file... which is how things work on non-Windows OSs.  However for ease of use and portability, I really think MG and the pipeline should try to work as file case does not matter.

MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
```diff
@@ -176,7 +185,7 @@ public bool NeedsRebuild(PipelineManager manager, PipelineBuildEvent cachedEvent
 
             // If the source file is newer than the dest file
             // then it must have been updated and needs a rebuild.
-            if (sourceWriteTime >= destWriteTime)
+            if (sourceWriteTime > destWriteTime)
```
tomspilman: Why?

MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
```diff
@@ -32,6 +33,11 @@ public PipelineBuildEvent()
         }
 
         /// <summary>
+        /// True if this item is a simple file copy, doing no importing or processing.
+        /// </summary>
+        public bool CopyOnly { get; set; }
```
tomspilman: Instead of making this an additional property that has to be propagated thru the system.... make it implicit based on having an importer and/or processor defined.

MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
```diff
@@ -716,65 +797,120 @@ public object ProcessContent(PipelineBuildEvent pipelineEvent)
...
 
         private void WriteXnb(object content, PipelineBuildEvent pipelineEvent)
         {
             // Make sure the output directory exists.
             var outputFileDir = Path.GetDirectoryName(pipelineEvent.DestFile);
+            
+            var wrongCase = PathHelper.DirectoryExistsWithDifferentCase(outputFileDir);
+            if (wrongCase.HasValue && wrongCase.Value == true)
+            {
+                throw new Exception(string.Format("Failed to write xnb '{0}' because the path already exists in a different case!", pipelineEvent.DestFile));
```
tomspilman: This seems bad...  I don't think we want to get into this mess of failing builds because of folder name case.

Tools/MGCB/BuildContent.cs
```diff
@@ -15,10 +16,20 @@ namespace MGCB
...
-        public bool LaunchDebugger = false;
+        public void OnLaunchDebugger(bool val)
+        {
+            LaunchDebugger = val;
+
+            if (LaunchDebugger)
+            {
+                Debugger.Launch();
+            }
+        }
```
tomspilman: Please pull this out into another PR.  Checking to see if a big PR works and doesn't break things is hard enough without packing in a bunch of separate features.

Tools/MGCB/BuildContent.cs
```diff
@@ -261,149 +299,201 @@ public void Build(out int successCount, out int errorCount)
 
             // If the target changed in any way then we need to force
             // a fuull rebuild even under incremental builds.
+            // jcf: what? why?
```
tomspilman: `previousContent` is what the intermediate data says was previously built.  If you decided to change what platform you are building for, but didn't change the folder it build to...  then the output and intermediate folders would have content for a different platform in it.  This code is to force a clean if you changed these settings and then wrote to the same folder.

Tools/MGCB/CommandLineParser.cs
```diff
@@ -187,6 +187,9 @@ public bool Parse(IEnumerable<string> args)
             {
                 showUsage = false;
 
+                // echo each command, at least if we aren't being run from a command window...
+                Console.WriteLine(">"+arg);
```
tomspilman: Lets not do this.

tomspilman:
So after reviewing this we will not merge this as is.

We need to at least split out the different mixed features in here.

 - Debug launch change.
 - Asset renaming.
 - File case changes.
 - Skipped file count info.

Then we can at least get the 2 or 3 easy and non-controversial features merged.

As is I don't think any of this will be merged.

MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
jamesford42: If your output file is the source file copied to output, then the write times will exactly equal (and always will). Without this change, copy items always evaluate as needing a rebuild.

MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
jamesford42: Its just this item that has failed.

jamesford42: I am using this feature exactly for removing case present in the source asset name and folder structure from the output. If give the copy item "Data/foo.dat" the asset name "data/foo.dat" and it says it built successfully, but when I look in bin/ its actually "Data/foo.dat" then it hasn't done what I asked and silently failed in a way that I wont realize until I crash trying to File.Open("data/foo.dat") on an actual ps4 (actually its even worse, it wont fail until I try it on an installed package).

KonajuGames: I'd the intent that Asset Name can specify a different path than the source file? Is that what XNA allowed? I don't think we would want that. I only ever used a name which renamed the XNB in the same relative path as the source file. The Asset Name should just rename the XNB.

MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
KonajuGames: If looking at the code can't explain why the change was needed, it should have a simple comment stating why it is the way it is.

MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
jamesford42: Yes, that is the intention. Think of asset name as what you would pass into Content.Load(...) at runtime. It is the filename and path relative to the content root.

In xna what you saw in the IDE was just the 'name'. But it did know the input and output paths relative to the project root / output root (respecively). Because this is how including an item "as a link" worked.

MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
jamesford42: The comment was already there. And it does accurately describe what the code was "suppose" to be doing. Technically this line is fixing a preexisting bug.

MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
tomspilman: > I'd the intent that Asset Name can specify a different path than the source file

I think so.  I think this feature is trying to replace both the asset name feature of XNA as well as the "link" feature of MSBuild scripts for including files from outside the content project folder.

Maybe there is a better way to support both... but I think we need both features.

Issue #2334, opened 2014-04-07
Android RenderTarget2D.GetData/SetData Inconsistencies and Problems On Pause/Resume


JRC-BitBull:
Hi,

I am writing some code to get around the annoying issue of references to RenderTarget2D objects being lost when an Android app is sent to the background. What I am trying to do is cache the raw data for these offscreen images when the app is sent to background using RenderTarget2D.GetData and then recreate them when the app is sent to the foreground using RenderTarget2D.SetData.

This is all kind of working apart from when the texture is recreated with SetData the colours are going all weird like the byte order has been changed or something. if I do the Get/Set process again on recreation it flips everything back to normal and things look fine!

Is this a bug or something I am doing wrong? My render targets are created with SurfaceFormat.Color so I'm guessing UInt32 is the correct format to use for storing the data here? The code is roughly analogous to this kind of thing...

```
UInt32[] data;
// on pause
data	= new UInt32[ texture.Width*texture.Height  ];
texture.GetData (data);

// on resume
// first recreate the texture and then...
texture.SetData	(data);

// all looks weird unless I add this again...
texture.GetData (data);
texture.SetData	(data);
```

Any insight appreciated...

Something else problematic I have noticed is that after repeatedly sending the app to background and returning it to foreground I start to lose the GL context on recreated textures even if I appear to have recreated them properly. This is also the case with Texture2D objects created using LoadContent (ie when MonoGame is handling the recreation of the texture).

I don't know if it's anything to do with it but I noticed in the logs that each time I send the app to background and resume the following piece of output in the logs (which only appears once when the app is first launched) appears again as expected bit is duplicated for each background/foreground cycle, so you end up with this piece of output repeated many, many times which doesn't seem right... though it could be pilot error!

```
[MonoGame] Supported extensions:
[MonoGame] GL_OES_EGL_image
[MonoGame] GL_OES_depth24
[MonoGame] GL_OES_depth32
[MonoGame] GL_OES_element_index_uint
[MonoGame] GL_OES_texture_float
[MonoGame] GL_OES_texture_float_linear
[MonoGame] GL_OES_compressed_paletted_texture
[MonoGame] GL_OES_compressed_ETC1_RGB8_texture
[MonoGame] GL_OES_depth_texture
[MonoGame] GL_OES_texture_half_float
[MonoGame] GL_OES_texture_half_float_linear
[MonoGame] GL_OES_packed_depth_stencil
[MonoGame] GL_OES_vertex_half_float
```

Issue #2297, opened 2014-03-25
GetData at Texture2D.cs for Android


grishaandrianov:
I suppose at GetData<T> (Texture2D.cs) it should be like this for android 
(https://github.com/Cocos2DXNA/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.cs)
after the "// Loop through and extract the data but we need to load it":
```C#
		var dataRowColOffset = 0;
		var sz = 0;
		var pixelOffset = 0;
		var dataElement = startIndex;
		for (int y = r.Top; y < r.Bottom; y++)
		{
		    for (int x = r.Left; x < r.Right; x++)
		    {
                        var result = new Color(0, 0, 0, 0);
                        dataRowColOffset = ((y * width) + x);
```
and after the "switch (Format)":
```C#
		data[dataElement++] = (T)(object)result;
```

Issue #1538, opened 2013-03-13
Time to Solve the GetData problem


SamuelEnglard:
Ok we all agree that Writing data from the CPU to the GPU is just bad performance wise so lets come up with all the reasons you'd want to and what can be done instead

Collision detection: As has been said before just do this with off screen data.
Screen Capture: Best solution I have to this would be to create a sub class of `RenderTarget2D` that internally stores it's data "CPU side" (aka doesn't actually draw to screen ever) so reading it back is much quicker.

Any other use cases?

Issue #1536, opened 2013-03-13
[Android] GetData on Nexus 7 leaves graphics in weird state


totallyeviljake:
GetData on the Nexus 7 is not supported. So if you capture the GL error (InvalidEnum) and attempt to recover, you can't. The blending state is so distorted that the screen flickers between updates and your sprites have no blending (all black) and appear as black boxes. Eventually the draw ceases to work properly but the game continues.

I am adding this issue as a reminder that GetData is dicey business and that there is a side effect should GetData fail.

PR #3054, opened 2014-10-08
Fix Texture2D.GetData() on GLES platforms

prollin:
Fix and Simplified Texture2D.GetData()
Works on both iOS and Android (Tested on iPad Mini 1st Gen, iPhone 6, Nexus 7 1st Gen and Nexus 5)
Removed superfluous large memory allocations

WARNING: Only renderable surface formats are supported. The previous code looked like it supported GetData on DXT texture but it will fail since opengl doesn't support attaching a compressed texture to a FBO.


tgjones:
> Fixes:
> #2297
> #1536
> #1405

> Probably Fixes:
> #2334
> #1538

All these are still open - I guess at least the first 3 can be closed now?

PR #3111, opened 2014-10-23
Hooks into ViewController

nikhilsinha:
I needed to add hooks into ViewDidAppear and ViewDidDisappear in the iOSGameViewController in order to use Flurry's ads. I created an interface for dependency injections.


tomspilman:
There never got merged.  We can revisit things with requirements for new ad systems as needed.

PR #3138, opened 2014-10-29
Rewrote content builder targets

Nezz:
### What it does
1. Automatically builds MonoGame Content projects before the project is built
2. Builds incrementally to allow faster development
3. Supports multiple projects in the same directory (like resolution packs), whitespaces in paths (like the Visual Studio 2013 folder)
4. Outputs built content to `ProjectDirectory\Content\PlatformName` (avoiding clash of contents built for different platforms)
5. Content files will be added to the project and copied to the output directory
6. Content files will be automatically packaged with the rest of the app (including Android)

TL;DL
Works as seamlessly as XNA content pipeline
### How to get it to work

Add the following lines to the csproj file using a text editor:
#### At the end of the first `<PropertyGroup>`:

``` XML
<MonoGamePlatform>PlatformName</MonoGamePlatform>
```

(PlatformName shuold be iOS, Android, WindowsPhone, etc)
#### Somewhere after an `</ItemGroup>`:

``` XML
  <ItemGroup>
    <MonoGameContentReference Include="PathToYour.mgcb">
    </MonoGameContentReference>
  </ItemGroup>
```
#### At the end of the file, before `</Project>`

``` XML
  <Import Project="PathTo\MonoGame.Content.Builder.targets" />
```

Thanks to @jamesford42 and @tomspilman for the help.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
@@ -17,108 +17,89 @@
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <PropertyGroup>
-    <CoreBuildDependsOn>
-      MonoGame_BuildContentProjects;
-      $(CoreBuildDependsOn)
-    </CoreBuildDependsOn>
+  <!-- Add MonoGameContentReference to item type selection in Visual Studio -->
 
-    <CoreCleanDependsOn>
-      MonoGame_CleanContentProjects;
-      $(CoreCleanDependsOn)
-    </CoreCleanDependsOn>
-  </PropertyGroup>
+  <ItemGroup>
+    <AvailableItemName Include="MonoGameContentReference" />
+  </ItemGroup>
+
+  <!-- Get all Mono Game Content References and store them in a list -->
+
+  <ItemGroup>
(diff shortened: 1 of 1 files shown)


tgjones:
>> that doesn't work for me now. If I change a processor mgcb thinks nothing has changed and builds nothing.

> Then that is a bug... the plan was always to force a rebuild in those cases by default.

I just hit this bug - when I change my `ContentProcessor`, it doesn't cause a content rebuild. Has an issue already been raised for this? I couldn't find one.

tomspilman:
> I just hit this bug - when I change my  ContentProcessor , it 
> doesn't cause a content rebuild.

There isn't an issue up for it yet.

You can see here the TODOs where the checks belong...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs#L189

We would start by storing the timestamp of the assembly in PipelineBuildEvent (eventually we want to store a per-processor/importer version... but that is for another time).  The complication is that within this method it needs to resolve the processor/importer assembly which it doesn't have.  

What I suggest is passing `PipelineManager` thru to that method.  We would add a special helper method to return the timestamps by processor/importer name.  Then within `ProcessorInfo` and `ImporterInfo` cache the timestamp of the assembly... so the file access is done only once within `PipelineManager.ResolveAssemblies()`.

PR #3192, opened 2014-11-20
Change OpenGL resources dispose behavior

### LINKED ISSUE #3188: Null reference to RenderTarget2D.GraphicsDevice during Dispose()
My game allows the user to double-tap to load the next game screen. With the latest MonoGame source a bug came out of nowhere: when I double-tap to skip to the next screen, a null reference error popped up:
(inside MonoGame.Framework/Graphics/RenderTarget2D.OpenGL.cs)

```
    protected override void Dispose(bool disposing)
    {
        if (!IsDisposed)
        {
            GraphicsDevice.AddDisposeAction(() => 
            {
                this.GraphicsDevice.PlatformDeleteRenderTarget(this);
            });
        }

        base.Dispose(disposing);
    }
```

The error shows that "this.GraphicsDevice" is null. 

My guess is that the reference to the GraphicsDevice of this RenderTarget2D has been cleared before calling this method. Thing were working properly with the MonoGame source back in September. Kind of wondering if there can be a quick fix to this issue... Thanks very much! =)

prollin:
Simplified dispose behavior for opengl resources. It shouldn't be MonoGame's responsibility to delay resource disposal (should be done by a higher level rendering component). This PR should fix #3188

Note that this PR keeps `GraphicsDevice.AddDisposeAction()` to be used by the `GraphicsDevice.PlatformDispose` method. This is on purpose since there might be some corner cases there that I didn't get a chance to test yet.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
@@ -372,7 +372,7 @@ namespace Microsoft.Xna.Framework.Graphics
         /// This allows GL resources to be disposed from other threads, such as the finalizer.
         /// </summary>
         /// <param name="disposeAction">The action to execute for the dispose.</param>
-        static internal void AddDisposeAction(Action disposeAction)
+        static private void AddDisposeAction(Action disposeAction)
         {
             if (disposeAction == null)
                 throw new ArgumentNullException("disposeAction");
--- a/MonoGame.Framework/Graphics/OcclusionQuery.cs
+++ b/MonoGame.Framework/Graphics/OcclusionQuery.cs
@@ -63,11 +63,11 @@ namespace Microsoft.Xna.Framework.Graphics
             if (!IsDisposed)
             {
 #if OPENGL
-                GraphicsDevice.AddDisposeAction(() =>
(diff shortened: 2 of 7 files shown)


prollin:
@rtabbara could you paste a more complete stack trace? Thanks

rtabbara:
Thanks for looking at it again @prollin. The graphics resource finalizer kicks things off so that is the full stack trace. 

And for what it's worth, for me, within the CocosSharp WinGL tests, <code>BackgroundContext</code> was not null at the point of the exception being raised. I think the issue is that the order of finalizers is non-deterministic, so there really is no guarantee that BackgroundContext/ Window/ resources etc. are Disposed of in any consistent order. 

In general, I believe that the correct practice is that finalisers shouldn't be referencing other objects that have their own finalisers. I understand that this change was done with explicit Dispose() calls in mind, but I think we still need to nicely handle when this isn't necessarily the case.

tomspilman:
>  I understand that this change was done with explicit Dispose() calls in mind

Yes it was.

In general depending on finalizers to clean up graphics resources is a very dangerous practice.  We've even considered making finalizers spit out nasty warnings when they are run to alert the user of this.

Still it probably shouldn't crash.

PR #3197, opened 2014-11-22
Use FFmpeg for audio compilation

hach-que:
This rewrites the audio compilation part of the content pipeline to use FFmpeg instead of NAudio / MonoMac's Audio Toolkit.

I've tested compilation of audio assets on Linux and sound effects playback successfully (although music is being converted to Ogg format instead of Wave, but that's a separate issue).  As I don't have access to any other platforms (only Windows in a VM with no Audio), I'll need other people to verify that the audio compilation works and results in XNB files that can be played back

I'm pretty sure the format names for FFmpeg are correct for the conversion, but I'm less sure about the muxer names (which are the wrappers around the encoded audio data).  For PCM, the muxer had to be "s16le" which means "raw PCM data"; I'm not sure whether the ADPCM or IMA ADPCM formats should be using the Wave muxer or just contain the raw data.


tomspilman:
Ok finally was able to run my own test here and it seems like this works great on Windows.  Any bugs will be just specific issues with formats and parsing of wave header data.  Things we can fix in subsequent PRs.

@dellis1972 @hach-que - Do either of you have time to track down the source of the distortion on Mac?

tomspilman:
Also at some point someone should delete the NAudio files from the dependency repo.

MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
```diff
@@ -134,229 +114,188 @@ int QualityToBitRate(ConversionQuality quality)
...
+                    case ConversionFormat.Adpcm:
+                        // ADPCM Microsoft 
+                        ffmpegCodecName = "adpcm_ms";
+                        ffmpegMuxerName = "wav";
+                        format = 0x0002; /* WAVE_FORMAT_ADPCM */
+                        break;
+                    case ConversionFormat.Pcm:
+                        // PCM signed 16-bit little-endian
+                        ffmpegCodecName = "pcm_s16le";
+                        ffmpegMuxerName = "s16le";
```
tomspilman: @hach-que - Do you remember why is this `s16le` and not `wav` here?  

This seems to generate a raw audio file without a header.  Looking at the other formats we have a mix of files with and without headers.

Seems like if we're actually writing out to a file... we want it to always include headers in the written file.

KonajuGames: The raw data format is used because it is that raw data that is put in the XNB file.

tomspilman: So if we're writing to an output file... we probably should use "wav" there.

MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
```diff
@@ -134,229 +114,188 @@ int QualityToBitRate(ConversionQuality quality)
...
-        /// </summary>
-        public void Dispose()
-        {
-            Dispose(true);
-            GC.SuppressFinalize(this);
-        }
+                string ffprobeStdout, ffprobeStderr;
+                var ffprobeExitCode = ExternalTool.Run(
+                    "ffprobe",
+                    string.Format("-i \"{0}\" -show_entries streams -v quiet -of flat", temporarySource),
```
tomspilman: Note we're probing the original source file and not the converted output file here.  This means the format is likely incorrect.

PR #3211, opened 2014-11-27
Center OpenTK game window when it is resized to mimic the behavior of WinFormsGameWindow

AngeloG:


cra0zy:
You are missing calculations if the window goes out of screen. Unlike windows on linux it's totally allowed.

AngeloG:
You're right! I'll have to look into that. Should be a relatively simple fix.

PR #3310, opened 2014-12-23
Fullscreen enabling for WindowsDX

Shqrdx:
This experimental commit enables IsFullscreen in constructor, and in another methods with ApplyChange and also fixes ToggleFullScreen. Only for WindowsDX platform. Tested with XNA.
 
This is "soft" version - The window is just stretch up and loses borders.

Because these things already arent working it doesnt matter if my changes break them a bit.

### DIFF
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -635,8 +635,6 @@ namespace Microsoft.Xna.Framework
 
 #if !(WINDOWS && DIRECTX)
 
-            // FIXME : FULLSCREEN
-
             if (GraphicsDevice.PresentationParameters.IsFullScreen)
                 Platform.EnterFullScreen();
             else
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -634,6 +634,9 @@ namespace Microsoft.Xna.Framework
 			Platform.BeginScreenDeviceChange(GraphicsDevice.PresentationParameters.IsFullScreen);
 
 #if !(WINDOWS && DIRECTX)
+
+            // FIXME : FULLSCREEN
+
             if (GraphicsDevice.PresentationParameters.IsFullScreen)
                 Platform.EnterFullScreen();
             else
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -632,11 +632,13 @@ namespace Microsoft.Xna.Framework
         internal void applyChanges(GraphicsDeviceManager manager)
         {
 			Platform.BeginScreenDeviceChange(GraphicsDevice.PresentationParameters.IsFullScreen);
+
+#if !(WINDOWS && DIRECTX)
             if (GraphicsDevice.PresentationParameters.IsFullScreen)
                 Platform.EnterFullScreen();
             else
                 Platform.ExitFullScreen();
-
+#endif
             var viewport = new Viewport(0, 0,
 			                            GraphicsDevice.PresentationParameters.BackBufferWidth,
 			                            GraphicsDevice.PresentationParameters.BackBufferHeight);
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -663,7 +663,7 @@ namespace Microsoft.Xna.Framework.Graphics
                                             PresentationParameters.BackBufferWidth,
                                             PresentationParameters.BackBufferHeight,
                                             format,
-                                            SwapChainFlags.None);
+                                            SwapChainFlags.AllowModeSwitch);
 
                 if (useFullscreenParameter)
                 {
@@ -691,7 +691,7 @@ namespace Microsoft.Xna.Framework.Graphics
                     ModeDescription =
                     {
                         Format = format,
-                        Scaling = DisplayModeScaling.Stretched,
+                        Scaling = DisplayModeScaling.Unspecified,
                         Width = PresentationParameters.BackBufferWidth,
                         Height = PresentationParameters.BackBufferHeight,                        
                     },
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -666,10 +666,9 @@ namespace Microsoft.Xna.Framework.Graphics
                                             SwapChainFlags.None
                                             );
 
+                // This force to switch to fullscreen mode when hardware mode enabled(working in WindowsDX mode).
                 if (useFullscreenParameter)
                 {
-                    // this force to switch to fullscreen mode when hardware mode enabled(working in WindowsDX mode).
-
                     _swapChain.SetFullscreenState(PresentationParameters.IsFullScreen, null);
                 }
 
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -668,6 +668,8 @@ namespace Microsoft.Xna.Framework.Graphics
 
                 if (useFullscreenParameter)
                 {
+                    // this force to switch to fullscreen mode when hardware mode enabled(working in WindowsDX mode).
+
                     _swapChain.SetFullscreenState(PresentationParameters.IsFullScreen, null);
                 }
 
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -691,7 +691,11 @@ namespace Microsoft.Xna.Framework.Graphics
                     ModeDescription =
                     {
                         Format = format,
+#if WINRT
+                        Scaling = DisplayModeScaling.Stretched,
+#else
                         Scaling = DisplayModeScaling.Unspecified,
+#endif
                         Width = PresentationParameters.BackBufferWidth,
                         Height = PresentationParameters.BackBufferHeight,                        
                     },
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -590,7 +590,7 @@ namespace Microsoft.Xna.Framework.Graphics
             _d3dContext = _d3dDevice.ImmediateContext.QueryInterface<SharpDX.Direct3D11.DeviceContext>();
         }
 
-        internal void CreateSizeDependentResources()
+        internal void CreateSizeDependentResources(bool useFullscreenParameter = false)
         {
             _d3dContext.OutputMerger.SetTargets((SharpDX.Direct3D11.DepthStencilView)null,
                                                 (SharpDX.Direct3D11.RenderTargetView)null);
@@ -665,6 +665,11 @@ namespace Microsoft.Xna.Framework.Graphics
                                             format,
                                             SwapChainFlags.None);
 
+                if (useFullscreenParameter)
+                {
+                    _swapChain.SetFullscreenState(PresentationParameters.IsFullScreen, null);
+                }
+
                 // Update Vsync setting.
                 using (var dxgiDevice = _d3dDevice.QueryInterface<SharpDX.DXGI.Device1>())
                 {
@@ -696,7 +701,7 @@ namespace Microsoft.Xna.Framework.Graphics
                     Usage = SharpDX.DXGI.Usage.RenderTargetOutput,
                     BufferCount = 2,
                     SwapEffect = SharpDXHelper.ToSwapEffect(PresentationParameters.PresentationInterval),
-                    IsWindowed = true,
+                    IsWindowed = useFullscreenParameter ? PresentationParameters.IsFullScreen : true,
                 };
 
                 // Once the desired swap chain description is configured, it must be created on the same adapter as our D3D Device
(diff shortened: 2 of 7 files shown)


Shqrdx:
@tomspilman Your thoughts ?

tomspilman:
> This is "soft" version - The window is just stretch up and loses borders.

This is the part that bothers me.

On one hand the idea of supporting a fullscreen without changing the desktop resolution is an interesting idea.  This is probably something we should support in some way.

But this is not the feature people looking for fullscreen support are looking for.  They want the old desktop resolution changing fullscreen mode.

So the question is how does this evolve into the actual fullscreen mode feature.

@KonajuGames ?

MonoGame.Framework/Windows/WinFormsGamePlatform.cs
```diff
@@ -145,14 +170,39 @@ public override bool BeforeUpdate(GameTime gameTime)
...
+        {
+#if (WINDOWS && DIRECTX)
+            // FIXME : FULLSCREEN
+
+            if (_window._form.Location.X != 0 && _window._form.Location.Y != 0)
+            {
+                _savedWindowedLocX = _window._form.Location.X;
+                _savedWindowedLocY = _window._form.Location.Y;
+            }
+            _window.IsBorderless = true;
```
tomspilman: Shouldn't we just be using `_window._form.WindowState = FormWindowState.Maximized`?

theZMan:
I agree - on platforms that support proper full screen we should continue to implement it.

There's big advantages to the 'soft' approach - its faster to switch, you can still use common dialogs, tray pop ups and other apps still work, alt-tab doesn't have huge switches, Desktop icons are not resized etc...  

Its used to be that 'hard' full screen was faster - certainly in the DX9 time frame and I think in Win7/8 desktop compositing is disabled so there may still be a performance benefit. Plus its the only way to do full screen at a lower resolution which may be the only way to run some games.

So 'soft' should be a MonoGame enhancement and/or option on platforms that don't support it.

Shqrdx:
In original XNA the actual resolution of back buffer is not changed after switching to fullscreen. The only choose here is between "hard" switch on device or "soft" switch on window. So, @theZMan wants a "hard", whats says others ?

MonoGame.Framework/Windows/WinFormsGamePlatform.cs
Shqrdx: Good catch ! I totally forgot about it.

Shqrdx:
I have an idea - implementing GraphicsDeviceManager.UseHardwareFullscreenSwitch property which is a hint for internal routines what choose "hard" or "soft"...

Shqrdx:
After some tests I start thinking @theZMan version is more correct in this case, so I enabled it by default.

Shqrdx:
Ok, so now its fully working ! Both "hard" and "soft" switches works fine as intended with XNA. The new property can be changed only in constuctor so it will be safe. Last fix also protect WINRT platform(so you will be sure that nothing is changed for that platform).

Shqrdx:
Ok, so last fix completly makes WINRT platform untouched.

Shqrdx:
These last fixes I think is final - now you we change resolution and switch fullscreen without restrictions and bugs.

Shqrdx:
I even prepare a demo code for testing my changes. For correct result - texture must be any 1920x1080 texture and font doesnt matter.
<code>
    public class Game1 : Game
    {
        SpriteBatch spriteBatch;
        private KeyboardState lks;
        private Texture2D texture;
        private SpriteFont font;

        public Game1() : base()
        {

            graphics.PreferredBackBufferWidth = 640;
            graphics.PreferredBackBufferHeight = 480;

        }

        {
            base.Initialize();
        }

        {
            texture = Content.Load<Texture2D>("texture");
            font = Content.Load<SpriteFont>("font");
        }

        /// <summary>
        /// Allows the game to run logic such as updating the world,
        /// checking for collisions, gathering input, and playing audio.
        /// </summary>
        /// <param name="gameTime">Provides a snapshot of timing values.</param>
        {
            var ks = Keyboard.GetState();

            if (lks.IsKeyUp(Keys.Escape) && ks.IsKeyDown(Keys.Escape))
            {
                Exit();
            }

            if (lks.IsKeyUp(Keys.Space) && ks.IsKeyDown(Keys.Space))
            {
                graphics.ToggleFullScreen();
            }

            if (lks.IsKeyUp(Keys.D1) && ks.IsKeyDown(Keys.D1))
            {
                graphics.PreferredBackBufferWidth = 640;
                graphics.PreferredBackBufferHeight = 480;

                graphics.ApplyChanges();
            }

            if (lks.IsKeyUp(Keys.D2) && ks.IsKeyDown(Keys.D2))
            {
                graphics.PreferredBackBufferWidth = 800;
                graphics.PreferredBackBufferHeight = 600;

                graphics.ApplyChanges();
            }

            if (lks.IsKeyUp(Keys.D3) && ks.IsKeyDown(Keys.D3))
            {
                graphics.PreferredBackBufferWidth = 1024;
                graphics.PreferredBackBufferHeight = 768;

                graphics.ApplyChanges();
            }

            if (lks.IsKeyUp(Keys.D4) && ks.IsKeyDown(Keys.D4))
            {
                graphics.PreferredBackBufferWidth = 1920;
                graphics.PreferredBackBufferHeight = 1080;

                graphics.ApplyChanges();
            }

            lks = ks;

            base.Update(gameTime);
        }

        /// <summary>
        /// This is called when the game should draw itself.
        /// </summary>
        /// <param name="gameTime">Provides a snapshot of timing values.</param>
        {

            spriteBatch.Begin();


            spriteBatch.Draw(texture, GraphicsDevice.Viewport.Bounds, Color.White);

            spriteBatch.DrawString(font, "graphics.IsFullScreen/GD.IsFullScreen = " + graphics.IsFullScreen + " , " + GraphicsDevice.PresentationParameters.IsFullScreen + "\n" +
                                         "Viewport : " + GraphicsDevice.Viewport.ToString() + "\n" +
                                         "BackBuffer : " + GraphicsDevice.PresentationParameters.BackBufferWidth + " , " + GraphicsDevice.PresentationParameters.BackBufferHeight
                                         ,
            Vector2.Zero, Color.White);

            spriteBatch.End();

            base.Draw(gameTime);
        }
    }
</code>

Shqrdx:
The last fix if for minimizing routine - if user pressed alt+tab without that fix - he cannot restore fullscreen state correctly first time. This fix is fixing it )

Shqrdx:
@tomspilman Your thoughts ?

MonoGame.Framework/Game.cs
```diff
@@ -632,11 +632,16 @@ private void Platform_ApplicationViewChanged(object sender, ViewStateChangedEven
         internal void applyChanges(GraphicsDeviceManager manager)
         {
 			Platform.BeginScreenDeviceChange(GraphicsDevice.PresentationParameters.IsFullScreen);
+
+#if !(WINDOWS && DIRECTX)
+
+            // FIXME : FULLSCREEN
```
tomspilman: What is this FIXME for?  What code is missing here?  I'm ok with adding notes like this, but they need to clearly explain what is needed to be done.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -663,7 +663,13 @@ internal void CreateSizeDependentResources()
                                             PresentationParameters.BackBufferWidth,
                                             PresentationParameters.BackBufferHeight,
                                             format,
-                                            SwapChainFlags.None);
+                                            SwapChainFlags.None
+                                            );
+
+                if (useFullscreenParameter)
```
tomspilman: Can you add a comment here as to what this is doing?  Like I don't understand why we need `useFullscreenParameter` and don't just obey the `PresentationParameters.IsFullScreen` value.

MonoGame.Framework/GraphicsDeviceManager.cs
```diff
@@ -34,6 +34,14 @@ public class GraphicsDeviceManager : IGraphicsDeviceService, IDisposable, IGraph
         private DisplayOrientation _supportedOrientations;
         private bool _synchronizedWithVerticalRetrace = true;
         private bool _drawBegun;
+        private bool _hardwareModeSwitch = true;
+
+#if WINDOWS && DIRECTX
+
+        // FIXME : FULLSCREEN
```
tomspilman: Another FIXME that needs an explanation.

MonoGame.Framework/GraphicsDeviceManager.cs
```diff
@@ -286,6 +295,24 @@ public void ApplyChanges()
             //
             TouchPanel.DisplayWidth = _graphicsDevice.PresentationParameters.BackBufferWidth;
             TouchPanel.DisplayHeight = _graphicsDevice.PresentationParameters.BackBufferHeight;
+
+#if WINDOWS && DIRECTX
+
+            // FIXME : FULLSCREEN
```
tomspilman: Another FIXME that needs an explanation.

MonoGame.Framework/GraphicsDeviceManager.cs
```diff
@@ -368,6 +395,12 @@ private void Initialize()
         public void ToggleFullScreen()
         {
             IsFullScreen = !IsFullScreen;
+#if WINDOWS && DIRECTX
+
+            // FIXME : FULLSCREEN
```
tomspilman: Another FIXME that needs an explanation.

MonoGame.Framework/Windows/WinFormsGamePlatform.cs
```diff
@@ -118,8 +118,21 @@ public override bool BeforeRun()
         public override void BeforeInitialize()
         {
             _window.Initialize(Game.graphicsDeviceManager.PreferredBackBufferWidth, Game.graphicsDeviceManager.PreferredBackBufferHeight);
-
-            base.BeforeInitialize();
+            base.BeforeInitialize();
+
+#if (WINDOWS && DIRECTX)
+
+            // FIXME : FULLSCREEN
```
tomspilman: Another FIXME that needs an explanation.

MonoGame.Framework/Windows/WinFormsGamePlatform.cs
```diff
@@ -145,14 +158,64 @@ public override bool BeforeUpdate(GameTime gameTime)
...
-
-        public override void EnterFullScreen()
-        {
+        }
+
+
+        public override void EnterFullScreen()
+        {
+#if (WINDOWS && DIRECTX)
+            // FIXME : FULLSCREEN
```
tomspilman: Another FIXME that needs an explanation.

MonoGame.Framework/Windows/WinFormsGamePlatform.cs
```diff
@@ -145,14 +158,64 @@ public override bool BeforeUpdate(GameTime gameTime)
...
+            _alreadyInWindowedMode = false;
+            _alreadyInFullScreenMode = true;
+#endif
         }
 
         public override void ExitFullScreen()
-        {
+        {
+#if (WINDOWS && DIRECTX)
+            // FIXME : FULLSCREEN
```
tomspilman: Another FIXME that needs an explanation.

MonoGame.Framework/Windows/WinFormsGameWindow.cs
```diff
@@ -338,9 +353,13 @@ private void OnClientSizeChanged(object sender, EventArgs eventArgs)
 
                 var newWidth = _form.ClientRectangle.Width;
                 var newHeight = _form.ClientRectangle.Height;
+#if !(WINDOWS && DIRECTX)
+
+                // FIXME : FULLSCREEN
```
tomspilman: Another FIXME that needs an explanation.

tomspilman:
Generally ok, but I need to run your test case here and verify it is working.

Shqrdx:
@tomspilman I removed all "FIXME" cases. Of course my patch requires testing. So I'll be wait.

tomspilman:
>  I removed all "FIXME" cases. 

I guess that means there is nothing left to fix in those blocks now then?

Shqrdx:
@tomspilman Nothing, it was fault.

Shqrdx:
@dellis1972 @tomspilman Any progress of testing ?

willmotil:
so is this fixed now ?

i made a full methodical xna behavioral test a while back you should get the same results
against mono game

to perform that test make a mono-game project, drop the code in this link (below the output) into a game1 class. set your project to be of type console and run it 
, 
https://onedrive.live.com/?cid=BEE9E3EE9236C993&id=BEE9E3EE9236C993!261
the code exists at the bottom of the link after the xna output

you should get the same behavior even if you change it up a bit

KonajuGames:
Many larger games such as those from Valve support the "soft" fullscreen as described here, but it is a good thing to still support the exclusive fullscreen as well to maintain that compatibility.  If this PR supports both, I'm happy with it.

Shqrdx:
@KonajuGames Yes, it support both just switch graphics.HardwareModeSwitch in constructor.

Shqrdx:
@tomspilman Why is ignorring this useful commit ?

tomspilman:
> Why is ignorring this useful commit ?

I'm not ignoring it.

There are PRs that come in that are uncomplicated with no questionable features which I can glance over and know with pretty much certainty that it will work.  When I see these I try to merge immediately.

There are other PRs which either have features that need discussion or are complicated enough that they require testing before they can be merged.

This PR needs testing and unfortunately i don't have a lot of time for that at this moment.  Until i or someone else has time to test it will have to wait.

I wish i could spend 100% of my time on MonoGame, but unfortunately i cannot afford to.

Shqrdx:
@tomspilman But fullscreen state already is in unusable state so whats a point to not test it publicly by mergin it ? Any errors here can be fixed easy anyway ...

Shqrdx:
I rebase it, so I close this PR.

PR #3312, opened 2014-12-24
Added WindowsPhone81 platform to Content Pipeline

Nezz:
Should it have a different identifier?


KonajuGames:
Given the 1080p example, you still wouldn't need a separate platform identifier because you would be checking the resolution of the device before loading any 1080p-specific content.  Since WP8.0 won't report 1080p, you just simply won't load that content.

If there was a need for different processing for the same data, then it would be reasonable to have the extra identifier.

Nezz:
I don't want to include it in my XAP/APPX either.

tomspilman:
> I don't want to include it in my XAP/APPX either.

I think this is where having more than one "configuration" needs to come into play.  The configuration is like "Debug" or "Release" ...  in old XNA no one really used them.  However it has been our plan to support them for things like demo builds, adding dlc content, HD content, etc.  The idea was to be able to change properties or include content under only one particular config.

Like ideally you can tell it to build for WindowsPhone8, but pass a configuration identifier like "HDContent" that makes it include the additional 1080p content.

PR #3315, opened 2014-12-25
Monogame fixes for border, resize and fullscreen on Linux.

cra0zy:
This will allow for the border to remain fixed and not break the window resizing. Also re enabled fullscreen. And a fast test/example of the patch in action: http://youtu.be/PBlQMNuzijE


dellis1972:
Ideally this should be an OpenTK thing, but if we can't get it fixed we need to at least provide a working workaround.

Issue #3230, opened 2014-12-05
double click mgdb from VS sometimes opens wrong file


theZMan:
Sometimes the GUI tool opens the most recent mgcb not the one you double clicked on

In my case it was a brand new DX Windows project.

nkast:
I found a couple of entries in the registry, one of them missing double quotes around the pipilinepath and a default parameter.

Save the following two lines into a text file pipeline.reg , right click and 'Merge'. That fixed it for now.

[HKEY_CLASSES_ROOT\MonoGame.ContentBuilderFile\shell\open\command]
@="\"C:\\Program Files (x86)\\MSBuild\\MonoGame\\v3.0\\Tools\\Pipeline.exe\" \"%1\""

tomspilman:
@nkast - That bug is coming from here then:

https://github.com/mono/MonoGame/blob/develop/Installers/Windows/MonoGame.nsi#L107

Just need someone to tweak it, build and run the installer to test it.

tomspilman:
Anyone have time to investigate this one?

nkast:
the error must be at line 65, it has to out output `"%1"` at the end of the registry key and possibly wrap the path with quotes as well. But I am not familiar with the sintax of this installer.
https://github.com/mono/MonoGame/blob/develop/Installers/Windows/MonoGame.nsi#L65
the output should look like this 
`"C:\Program Files (x86)\MSBuild\MonoGame\v3.0\Tools\Pipeline.exe" "%1"`

nkast:
I tried NSIS 3.0b1 and 2.46 with the MonoGame.nsi and in both cases I got "!include: could not find: "header.nsh"". I am only guessing here, it's not clear from the teamcity build logs what type of installer the *.nsi is for.

I could add the "%1" and push a branch, wait for teamcity to build and then test the installer, but in case it fails it means more garbage PR. I don't know how .NSI handle quotes,double-quotes, escapes, etc.

PR #3317, opened 2014-12-27
Important pipeline bug fix.

Shqrdx:
This is simple and effecient solution for #3230 problem. 

How could I found this solution ? I insert changed Pipeline.exe to original tools folder and found that Main function misses the arguments completely at some point, for reason I dont solved. Probably the thing called IsolationStorage is guilty.. I still think its a temporary fix and whole pipeline should be checked to resolve this problem completly.

But for now this patch removes this problem from developer eyes.

### DIFF
--- a/Tools/Pipeline/Program.cs
+++ b/Tools/Pipeline/Program.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


tomspilman:
Yeah... same here.  Gonna close this for now.

PR #3336, opened 2014-12-30
Improve FBX support

mgarstenauer:
I have spent some time refactoring the OpenAssetImporter. The goal was to improve FBX support. If a model can be previewed in the [Assimp viewer](http://assimp.sourceforge.net/main_viewer.html), then MonoGame should be able to import it.

Still, there are tons of FBX models that Assimp does not support. I think this is the best we can do in MonoGame, everything else needs to be fixed in Assimp.

**DISCLAIMER:** I have mostly tested the FBX importer with our custom model processor. Our model processor now works with the XNA content pipeline and the MonoGame content pipeline (same code).
Please test your models with MonoGame's ModelProcessor!

**Notes about XNA vs. MonoGame/Assimp:**
XNA was built with the 2006.11 version of the FBX exporter. It may or may not support newer FBX versions. This is totally unpredictable. It depends on the internal structure of the FBX file -- the actual file version is secondary.

Assimp supports FBX version 7.1 (FBX 2011), 7.2 (FBX 2012), and 7.3 (FBX 2013). That means most original XNA models need to be converted to a newer version.
Assimp is not as powerful as the FBX SDK. It is best to check each FBX file in the [Assimp viewer](http://assimp.sourceforge.net/main_viewer.html). If the model does not look right in the viewer, then it can't be imported with MonoGame. The model needs to be fixed in 3dsmax (or another tool).

For example: The XNA's Dude.fbx from the [SkinningSample](http://xbox.create.msdn.com/en-US/education/catalog/sample/skinned_model) needs to be converted to a newer FBX version. But even then it can't be loaded by Assimp. We had to fix pivots/transforms and animations manually in 3dsmax.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
@@ -4,35 +4,139 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using Assimp;
-using Assimp.Configs;
-using Microsoft.Xna.Framework.Content.Pipeline.Graphics;
-using Microsoft.Xna.Framework.Content.Pipeline;
-using Microsoft.Xna.Framework;
-using System.IO;
 using System.Diagnostics;
-using Quaternion = Microsoft.Xna.Framework.Quaternion;
-using System.Text;
-using Assimp.Unmanaged;
+using System.Linq;
 using System.Reflection;
-using System.Collections.Generic;
+using Assimp;
+using Assimp.Unmanaged;
+using Microsoft.Xna.Framework.Content.Pipeline.Graphics;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline
 {
-    [ContentImporter(".fbx", ".x", DisplayName = "Open Asset Import Library - MonoGame", DefaultProcessor = "ModelProcessor")]
+    [ContentImporter(".dae", ".fbx", ".x", DisplayName = "Open Asset Import Library - MonoGame", DefaultProcessor = "ModelProcessor")]
     public class OpenAssetImporter : ContentImporter<NodeContent>
     {
+        // Assimp has a few limitations (not all FBX files are supported):
+        // FBX files reference objects using IDs. Therefore, it is possible to resolve
+        // bones even if multiple bones/nodes have the same name. But Assimp references
+        // bones only by name!
+        // --> Limitation #1: A model cannot have more than one skeleton!
+        // --> Limitation #2: Bone names need to be unique!
+        //
+        // Bones are represented by regular nodes, but there is no flag indicating whether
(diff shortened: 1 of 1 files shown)


KonajuGames:
Alias provide a free FBX converter utility on their website. Installing
Blender just to convert a FBX file when their import and export are
incomplete seems a bit of overkill.

mgarstenauer:
> Should we inquire to the Assimp team on support for older FBX formats?

Probably not worth the effort, IMHO. I guess all current 3D tools output FBX v7.x, which can be read by Assimp. Old XNA models can be converted with the [Autodesk FBX Converter](http://usa.autodesk.com/adsk/servlet/pc/item?siteID=123112&id=10775920).

>> But even then it can't be loaded by Assimp. We had to fix pivots/transforms 
and animations manually in 3dsmax.

> Is that something that should be fixed within Assimp?

This is a serious problem: Most animated models off the web do not follow any rules. Most of them have some quirks when opened in the Assimp viewer.

As I see, our options are:
- Test which settings work with Assimp and then stick with these settings. Of course, this requires that you create your own models or know how to fix animated models in 3dsmax (or another tool).
- Improve FBX support in Assimp. But I guess this is really hard. There is no official FBX specification - at least none that I am aware of.
- Use the Autodesk FBX SDK directly. Though I am not sure if the license allows it. The Blender team, for example, is developing their own FBX importer/exporter because of licensing issues.

tomspilman:
> Old XNA models can be converted with the Autodesk FBX Converter.

 Didn't know that existed...  seems like the right thing to point people at.  We should add this into the docs on model support before too long.

> Most animated models off the web do not follow any rules.

 I see... you are talking about them being exported with different standards.

 But the XNA's Dude.fbx should follow the correct standards for XNA...  so I guess the conversion to a newer FBX version introduced the "quirks".

 If we can eventually identify some common "quirks" we should look to add some more properties into the processor to enable those fixes.

tomspilman:
Merging...  thanks to @mgarstenauer and DigitalRune for this huge contribution to MonoGame!

tattries:
Models that worked a few weeks ago throw exception with latest build. Error is System.AccessViolationException in SharpdDX.dll "Attempted to read or write protected memory. This is often an indication that other memory is corrupt."
StackTrace is  at 
SharpDX.Interop.memcpy(Void* pDest, Void* pSrc, Int32 count)
   at SharpDX.Utilities.CopyMemory(IntPtr dest, IntPtr src, Int32 sizeInBytesToCopy)
   at Microsoft.Xna.Framework.Graphics.VertexBuffer.PlatformSetDataInternal[T](Int32 offsetInBytes, T[] data, Int32 startIndex, Int32 elementCount, Int32 vertexStride, SetDataOptions options, Int32 bufferSize, Int32 elementSizeInBytes)
   at Microsoft.Xna.F

Could this be related to FBX changes?

tgjones:
This is related to #3356. I'll submit a fix shortly.

PR #3340, opened 2014-12-31
SpriteBatch new comments with examples.

Shqrdx:
As I promised I continue edit SpriteBatch. Please, find errors if you have some time. 

@tomspilman you said we need some examples, so I writed example to almost every method in this class.


Shqrdx:
Hmm.. maybe instead of pushing these example to *.cs file, I can create a page in docs.. Ok, I close it, cuz its a mess.

PR #3342, opened 2015-01-01
Added building of Pipeline .pkg installer for MacOS

dellis1972:


dellis1972:
Don't merge yet there is a PR in the Dependencies that needs merging first

dellis1972:
https://github.com/Mono-Game/MonoGame.Dependencies/pull/36

tomspilman:
@dellis1972 - So what is the location and names of the package files this generates?  I'll make sure they appear in the build artifacts.  Also what should i put on the MonoGame website download page?

dellis1972:
@tomspilman You will need to include the artifact Pipeline.Installer.pkg which should be located in 

Tools/Pipeline/bin/MacOS/AnyCPU/Release/

Also yes we should put this (Pipeline.Installer.pkg) on the download page.

Note this package will only include the Pipeline tool (and MGCB which is bundled) It WILL need mono installed on the system to work, So we need to mention that too. 

I'm still working on the Xamarin Studio Addin

tomspilman:
I see a Pipeline.MacOS.pkg in there too...  is that something different?

cra0zy:
@dellis1972 Do you need any help with xamarin studio addin?

@tomspilman @dellis1972 I was thinking of a way to setup installation for linux and the best way I found would be to use something like makeself.sh, basically creates self extracting archive and runs a script afterwards. Do you guys have something specific you want me to setup for installation(like gui or distribution specific packages), or would that be good enough?

tomspilman:
> Do you guys have something specific you want me to setup for installation

I have no clue.  Seems like whatever is considered a "good" installation process for Linux desktop users would be preferred.

@dellis1972 might have a better opinion on it.

dellis1972:
@tomspilman the Pipeline.MacOS.pkg can be ignored, its packaged withing the Installer.pkg

@cra0zy my experience of installers on Linux is not great, so I think for now a simple makeself.sh script might do the trick. We currently don't have a Linux based build bot though, so not sure how we'd run that as part of the CI process.

Also @cra0zy I'm almost done with the new XS extension, just need to tidy it up and get the PR up :) but thanks for the offer :) Might need your help testing on Linux though.

cra0zy:
@dellis1972 I know that you use windows servers to make linux stuff, but windows can use Cygwin to run linux sh files which makes makefile.sh quite nice to use.

kjpou1:
Congrats guys.  Great work!!!!

Can't wait for the XS extension either but this is one nail in the coffin of the Pipeline support on non windows.

tomspilman:
@dellis1972 - I have updated the download page...  give me some feedback on it:  http://www.monogame.net/downloads/

@cra0zy - When you work out the install process for Linux let me know.  If it is just a script + binaries included in a tar.gz archive I can easily work that into the build process on Windows.

@kjpou1 - Thanks...  it has been a very long time coming!

dellis1972:
Looks good to me :)

-----Original Message-----

cra0zy:
@tomspilman I've actually finished the linux installer the moment I said "Found a bug while I was making linux installer scripts..." on the PR, it generates .run file but all you need is cygwin, and to run one my scripts which will generate it. I can only do 1 pull request at a time because of my setup (multiple local branches from cra0zy:develop, merge them up when I want to add a PR), I'm still quite new when it comes to using git.

cra0zy:
Side note, monodevelop addin can be installed without the need for mpack package. All that is needed is for the addin to be copied to monodevelop addin folder which for linux is: ~/.local/share/MonoDevelop-5.0/LocalInstall/Addins, so we could include the addin with the installers.

tomspilman:
> it generates .run file but all you need is cygwin, and to run one my scripts which will generate it.

I really want to avoid installing Cygwin on the build box.  My experience with it has always been negative and I don't want to require others to use it.

Can we write the script to generate the installer in something the build already supports like Nant?  What does the script do other than package the files into a tar file?

cra0zy:
To explain, I've created several scripts:
generatelinuxinstaller.sh - copies all the needed files to Data, and creates .run package using that. .run package will extract to temp folder and run postinstall.sh. All the linux user would have to do is "sudo ./monogame-pipeline.run"
postinstall.sh - this is the script that actually does the installation, copies everything, creates launcher....
uninstall.sh - uninstallation

Now if you want to replace generatelinuxinstaller.sh, someone would need to copy all the needed files and added them to archive, also I really don't know any way to create that nice self extracting archive for linux, so the person extracting the archive would need to manually launch postinstall.sh to install the pipeline.

cra0zy:
I know what I can do for the installation, I can create an installer by adding all the needed files as embeded resources, and then just copying them to tmp dir. It will still be one file, it will just have to be run with "sudo mono monogame-pipeline.run".

dellis1972:
@cra0zy I dropped you a mail about getting a screen shot for Linux of the Pipeline tool. It might have gone into spam though, so I just wanted to give you a heads up here too :)

cra0zy:
Bad idea, files end up being too big...

Issue #3271, opened 2014-12-13
iOS 8.1 orientation broken


Nezz:
Landscape orientation stopped working as of iOS 8.1. The screen does not autorotate to landscape.

PR #3344, opened 2015-01-02
iOS: robust rotation handling for iPad/iPhone devices

rtabbara:
* Should address issue raised by #3271
* In particular, both iPad/iPhone devices should correctly autorotate
* The main problem was that previously, iPad devices would not receive TraitCollectionDidChange calls after a rotation, so this is an alternate solution


tomspilman:
@Nezz @TheRoboZ @Andreyul

What are your thoughts on this PR?

dellis1972:
@tomspilman this looks ok to me, but i'll let someone with a bit more experience on iOS weigh in :)

Nezz:
I'll check this out tomorrow.

Nezz:
Looking into it now...

Nezz:
Works fine with an iPad running iOS 8.1.
Doesn't work correctly with an iPhone 4S running iOS 7.1. The GraphicsDevice.Viewport is 640x960 instead of 960x640

rtabbara:
Fwiw, this change only impacts iOS 8+, so if indeed iOS 7 rotation is
broken then we'll need to do something extra. I suppose I shouldn't have
called the PR robust.

Nezz:
The iPad running iOS 8 is broken too as I can still reproduce #3271

rtabbara:
Thanks @Nezz for looking into this, I'll close this for now and take a look at it again with the Unified changes incorporated.

PR #3346, opened 2015-01-02
Viewport improvements

Shqrdx:
Changes : 
  -Comments for all things inside Viewport struct.
  -New constructor which accept full Viewport description.
  -Change TitleSafeArea a bit.(I still thinks that is subject to change for platforms like PS4 or XBOX)

PS If this change is not correct I remove it quickly.

### DIFF
--- a/MonoGame.Framework/Graphics/Viewport.cs
+++ b/MonoGame.Framework/Graphics/Viewport.cs
@@ -19,11 +19,11 @@ namespace Microsoft.Xna.Framework.Graphics
 		private int height;
 		private float minDepth;
 		private float maxDepth;
-		
-		#region Properties
+
+        #region Properties
 
         /// <summary>
-        /// The height of this viewport.
+        /// The height of the bounds in pixels.
         /// </summary>
         [DataMember]
         public int Height
@@ -65,7 +65,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		}
 
         /// <summary>
-        /// The width of this viewport.
+        /// The width of the bounds in pixels.
         /// </summary>
         [DataMember]
         public int Width
--- a/MonoGame.Framework/Graphics/Viewport.cs
+++ b/MonoGame.Framework/Graphics/Viewport.cs
@@ -140,7 +140,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		}
 
         /// <summary>
-        /// Gets the actual visible area of this <see cref="Viewport"/> which can differs from <see cref="Bounds"/> on some platforms.
+        /// Returns the subset of the viewport that is guaranteed to be visible on a lower quality display.
         /// </summary>
 		public Rectangle TitleSafeArea 
 		{
--- a/MonoGame.Framework/Graphics/Viewport.cs
+++ b/MonoGame.Framework/Graphics/Viewport.cs
@@ -1,4 +1,4 @@
-// MIT License - Copyright (C) The Mono.Xna Team
+// MonoGame - Copyright (C) The MonoGame Team
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Graphics/Viewport.cs
```diff
@@ -1,55 +1,18 @@
...
-code form, you may only do so under a license that complies with this license.
-(E) The software is licensed "as-is." You bear the risk of using it. The contributors give no express warranties, guarantees
-or conditions. You may have additional consumer rights under your local laws which this license cannot change. To the extent
-permitted under your local laws, the contributors exclude the implied warranties of merchantability, fitness for a particular
-purpose and non-infringement.
-*/
-#endregion License
-
-using Microsoft.Xna.Framework;
+// MIT License - Copyright (C) The Mono.Xna Team
```
tomspilman: This is the wrong license header.  It should be the MonoGame one.

MonoGame.Framework/Graphics/Viewport.cs
```diff
@@ -58,6 +21,10 @@ public struct Viewport
 		private float maxDepth;
 		
 		#region Properties
+
+        /// <summary>
+        /// The height of this <see cref="Viewport"/>.
```
tomspilman: We don't need to include `of this <see cref="Viewport"/>` on every summary.  Also the link is redundant as it is just linking to the same docs they are already reading.

tomspilman: For example a more useful description could be...
> The height of the bounds in pixels.

Shqrdx:
@tomspilman Whats I must change next ?

MonoGame.Framework/Graphics/Viewport.cs
```diff
@@ -172,11 +173,42 @@ public Viewport(int x, int y, int width, int height)
 		    this.minDepth = 0.0f;
 		    this.maxDepth = 1.0f;
 		}
-		
+
+        /// <summary>
+        /// Creates a new instance of <see cref="Viewport"/> struct.
+        /// </summary>
+        /// <param name="x">The x coordinate of the upper-left corner, of the view bounds, in pixels.</param>
```
tomspilman: The commas are not necessary here.

Also use this same description for the X/Y property as well.

MonoGame.Framework/Graphics/Viewport.cs
```diff
@@ -154,15 +138,32 @@ public Rectangle Bounds
...
 			}
 		}
 		
+        /// <summary>
+        /// Creates a new instance of <see cref="Viewport"/> struct.
+        /// </summary>
+        /// <param name="x">The x coordinate of the view bounds.</param>
+        /// <param name="y">The y coordinate of the view bounds.</param>
+        /// <param name="width">The width of the view bounds.</param>
+        /// <param name="height">The height of the view bounds.</param>
```
tomspilman: The param descriptions here should match the ones in the other constructor.

MonoGame.Framework/Graphics/Viewport.cs
```diff
@@ -154,15 +138,32 @@ public Rectangle Bounds
 				height = value.Height;
 			}
 		}
-		
+
+        /// <summary>
+        /// Gets the actual visible area of this <see cref="Viewport"/> which can differs from <see cref="Bounds"/> on some platforms.
```
tomspilman:
```
Returns the subset of the viewport that is guaranteed to be visible on a lower quality display.
```

Shqrdx:
@tomspilman Now ok ?... cuz I wanna sleep : p

tomspilman:
@Shqrdx - Don't feel like you need to stay up...  this can always wait. ;)

Shqrdx:
@tomspilman Thank you ! :)

PR #3348, opened 2015-01-03
Plane improvements

tgjones:
I've implemented the missing Plane.Transform methods, to transform a `Plane` by a `Matrix` and a `Quaternion`.

(I think this is my first pull request to MonoGame!)


tomspilman:
This is great!  Thanks for the fix and the tests to match!

Issue #3349, opened 2015-01-03
ContentProcessorContext.BuildAndLoadAsset fails when trying to build .fx files


tgjones:
I have a ContentProcessor that dynamically builds an effect file, and then calls `BuildAndLoadAsset` like this:

``` csharp
var tempFxFile = @"C:\Users\Tim\AppData\Local\Temp\DynamicallyGeneratedEffect.fx";
CompiledEffectContent compiledEffect = context.BuildAndLoadAsset<EffectContent, CompiledEffectContent>(
    new ExternalReference<EffectContent>(tempFxFile),
    "EffectProcessor");
```

It works fine in XNA. But in MonoGame, it crashes with this message: "The path is not of a legal form.". Here is the stack trace:

```
mscorlib.dll!System.IO.Path.NormalizePath(string path, bool fullCheck, int maxPathLength, bool expandShortPaths)	Unknown
 	mscorlib.dll!System.IO.Path.GetFullPath(string path)	Unknown
 	MonoGame.Framework.Content.Pipeline.dll!TwoMGFX.ShaderInfo.FromString(string effectSource, string filePath, TwoMGFX.Options options)	Unknown
 	MonoGame.Framework.Content.Pipeline.dll!Microsoft.Xna.Framework.Content.Pipeline.Processors.EffectProcessor.Process(Microsoft.Xna.Framework.Content.Pipeline.Graphics.EffectContent input, Microsoft.Xna.Framework.Content.Pipeline.ContentProcessorContext context)	Unknown
 	MonoGame.Framework.Content.Pipeline.dll!Microsoft.Xna.Framework.Content.Pipeline.ContentProcessor<Microsoft.Xna.Framework.Content.Pipeline.Graphics.EffectContent,Microsoft.Xna.Framework.Content.Pipeline.Processors.CompiledEffectContent>.Microsoft.Xna.Framework.Content.Pipeline.IContentProcessor.Process(object input, Microsoft.Xna.Framework.Content.Pipeline.ContentProcessorContext context)	Unknown
 	MonoGame.Framework.Content.Pipeline.dll!MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.ProcessContent(MonoGame.Framework.Content.Pipeline.Builder.PipelineBuildEvent pipelineEvent)	Unknown
[...]

PR #3350, opened 2015-01-04
Possible fix for PipelineProcessorContext.BuildAndLoadAsset bug

tgjones:
I don't know if this is the correct fix - changing `ResolveOutputFilepath` to `internal` makes me uncomfortable. But more generally, it seems to me (without much knowledge of this code) that `BuildAsset` and `BuildAndLoadAsset` could/should share more code than they are currently sharing.

I'm happy to rework this, if you can give me some pointers on a better way to fix it.

In any case, this fixes the issue I raised in #3349, so I'm now able to build dynamically generated effects. Yay!

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
@@ -374,7 +374,7 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
             return result;
         }
 
-        private void ResolveOutputFilepath(string sourceFilepath, ref string outputFilepath)
+        internal void ResolveOutputFilepath(string sourceFilepath, ref string outputFilepath)
         {
             // If the output path is null... build it from the source file path.
             if (string.IsNullOrEmpty(outputFilepath))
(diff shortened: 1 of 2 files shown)


tomspilman:
Note I wrote a lot of this code, but even I get confused by it all.

> But more generally, it seems to me (without much knowledge of this code) that 
> `BuildAsset` and `BuildAndLoadAsset` could/should share more code than 
> they are currently sharing.

There are possibly some differences in there that need to be fixed, but they do very different things.  `BuildAsset` actually writes an XNB to disk where as `BuildAndLoadAsset` does not.

This leads us to my next question...  why do we need to set the `DestFile` when `BuildAndLoadAsset` never writes this content to disk?

Your bug reported in #3349 originates from here...

https://github.com/mono/MonoGame/blob/develop/Tools/2MGFX/ShaderInfo.cs#L81

It looks like I added the `OutputFilePath` property in https://github.com/mono/MonoGame/pull/2567.  It is unused in the public repo, but we do use in the private PS4 repo for writing output debug files.

So I suggest we close this PR and instead try a new PR which only sets the `ShaderInfo.OutputFilePath` if `options.OutputFile` is not null.  This should fix your issue specific to the effect processor without changing asset processing as a whole.

tgjones:
Ah, that's the source of my confusion - somewhere along the way, I had forgotten that `BuildAndLoadAsset` doesn't write to disk. Thanks for clearing that up.

I'll submit a new PR with your proposed fix.

PR #3354, opened 2015-01-05
Fix dictation on i os 8

panayiotispanteli:


panayiotispanteli:
Sorry. I made this pull request against the wrong base fork.

tomspilman:
@panayiotispanteli - I'm not sure what this PR is actually for.  There are lots of unrelated changes in here as well as stuff that we cannot merge like MonoMac binaries.

Can you explain what you're trying to do and maybe we can work out a path forward.

tomspilman:
@panayiotispanteli - No prob...  I was confused there for a second. :)

PR #3356, opened 2015-01-05
Implement vertexStride in VertexBuffer.SetData for DirectX platform.

tgjones:
This code roughly mirrors the equivalent path in `GetData`, although it is only used when `vertexStride != elementSize`, so it doesn't affect existing usage. An alternative implementation (which I actually tried first), which avoids the need for a staging buffer, would call `UpdateSubresource` once for each element, but I guess that would be very slow for large arrays.

Note that this only implements the `vertexStride` parameter for non-dynamic `VertexBuffer`s (which is actually harder than it would be for dynamic `VertexBuffer`s, since you wouldn't need the staging buffer for a dynamic `VertexBuffer`).

I changed `Game.RunOneFrame()` to `Game.Run()` for these tests, because `Game.RunOneFrame()` doesn't seem to call `Draw()` when running tests in XNA, so the tests gave false positives. If this is really the case (I'd be happy to be wrong!), then we should update other tests to match, or figure out why `Game.RunOneFrame()` isn't calling `Draw()`.

### DIFF
--- a/MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
+++ b/MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
@@ -138,18 +138,52 @@ namespace Microsoft.Xna.Framework.Graphics
                 var startBytes = startIndex * elementSizeInBytes;
                 var dataPtr = (IntPtr)(dataHandle.AddrOfPinnedObject().ToInt64() + startBytes);
 
-                var box = new SharpDX.DataBox(dataPtr, 1, 0);
+                var d3dContext = GraphicsDevice._d3dContext;
+                lock (d3dContext)
+                {
+                    if (vertexStride == elementSizeInBytes)
+                    {
+                        var box = new SharpDX.DataBox(dataPtr, 1, 0);
 
-                var region = new SharpDX.Direct3D11.ResourceRegion();
-                region.Top = 0;
-                region.Front = 0;
-                region.Back = 1;
-                region.Bottom = 1;
-                region.Left = offsetInBytes;
-                region.Right = offsetInBytes + (elementCount * elementSizeInBytes);
+                        var region = new SharpDX.Direct3D11.ResourceRegion();
+                        region.Top = 0;
+                        region.Front = 0;
+                        region.Back = 1;
+                        region.Bottom = 1;
+                        region.Left = offsetInBytes;
+                        region.Right = offsetInBytes + (elementCount * elementSizeInBytes);
 
-                lock (GraphicsDevice._d3dContext)
-                    GraphicsDevice._d3dContext.UpdateSubresource(box, _buffer, 0, region);
+                        d3dContext.UpdateSubresource(box, _buffer, 0, region);
+                    }
+                    else
+                    {
+                        // Copy the buffer to a staging resource, so that any elements we don't write to will still be correct.
+                        var stagingDesc = _buffer.Description;
+                        stagingDesc.BindFlags = SharpDX.Direct3D11.BindFlags.None;
+                        stagingDesc.CpuAccessFlags = SharpDX.Direct3D11.CpuAccessFlags.Read | SharpDX.Direct3D11.CpuAccessFlags.Write;
+                        stagingDesc.Usage = SharpDX.Direct3D11.ResourceUsage.Staging;
+                        stagingDesc.OptionFlags = SharpDX.Direct3D11.ResourceOptionFlags.None;
+                        var stagingBuffer = new SharpDX.Direct3D11.Buffer(GraphicsDevice._d3dDevice, stagingDesc);
+
+                        lock (GraphicsDevice._d3dContext)
+                            d3dContext.CopyResource(_buffer, stagingBuffer);
+
+                        // Map the staging resource to a CPU accessible memory
+                        var box = d3dContext.MapSubresource(stagingBuffer, 0, SharpDX.Direct3D11.MapMode.Read, SharpDX.Direct3D11.MapFlags.None);
+
+                        for (int i = 0; i < data.Length; i++)
+                            SharpDX.Utilities.CopyMemory(box.DataPointer + i * vertexStride + offsetInBytes, dataPtr + i * elementSizeInBytes, elementSizeInBytes);
+
+                        // Make sure that we unmap the resource in case of an exception
+                        d3dContext.UnmapSubresource(stagingBuffer, 0);
+
+                        // Copy back from staging resource to real buffer.
+                        lock (GraphicsDevice._d3dContext)
+                            d3dContext.CopyResource(stagingBuffer, _buffer);
+
+                        stagingBuffer.Dispose();
+                    }
+                }
 
                 dataHandle.Free();
             }
--- a/MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
+++ b/MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
@@ -139,50 +139,53 @@ namespace Microsoft.Xna.Framework.Graphics
                 var dataPtr = (IntPtr)(dataHandle.AddrOfPinnedObject().ToInt64() + startBytes);
 
                 var d3dContext = GraphicsDevice._d3dContext;
-                lock (d3dContext)
+                
+                if (vertexStride == elementSizeInBytes)
                 {
-                    if (vertexStride == elementSizeInBytes)
-                    {
-                        var box = new SharpDX.DataBox(dataPtr, 1, 0);
+                    var box = new SharpDX.DataBox(dataPtr, 1, 0);
 
-                        var region = new SharpDX.Direct3D11.ResourceRegion();
-                        region.Top = 0;
-                        region.Front = 0;
-                        region.Back = 1;
-                        region.Bottom = 1;
-                        region.Left = offsetInBytes;
-                        region.Right = offsetInBytes + (elementCount * elementSizeInBytes);
+                    var region = new SharpDX.Direct3D11.ResourceRegion();
+                    region.Top = 0;
+                    region.Front = 0;
+                    region.Back = 1;
+                    region.Bottom = 1;
+                    region.Left = offsetInBytes;
+                    region.Right = offsetInBytes + (elementCount * elementSizeInBytes);
 
+                    lock (d3dContext)
                         d3dContext.UpdateSubresource(box, _buffer, 0, region);
-                    }
-                    else
-                    {
-                        // Copy the buffer to a staging resource, so that any elements we don't write to will still be correct.
-                        var stagingDesc = _buffer.Description;
-                        stagingDesc.BindFlags = SharpDX.Direct3D11.BindFlags.None;
-                        stagingDesc.CpuAccessFlags = SharpDX.Direct3D11.CpuAccessFlags.Read | SharpDX.Direct3D11.CpuAccessFlags.Write;
-                        stagingDesc.Usage = SharpDX.Direct3D11.ResourceUsage.Staging;
-                        stagingDesc.OptionFlags = SharpDX.Direct3D11.ResourceOptionFlags.None;
-                        var stagingBuffer = new SharpDX.Direct3D11.Buffer(GraphicsDevice._d3dDevice, stagingDesc);
+                }
+                else
+                {
+                    // Copy the buffer to a staging resource, so that any elements we don't write to will still be correct.
+                    var stagingDesc = _buffer.Description;
+                    stagingDesc.BindFlags = SharpDX.Direct3D11.BindFlags.None;
+                    stagingDesc.CpuAccessFlags = SharpDX.Direct3D11.CpuAccessFlags.Read | SharpDX.Direct3D11.CpuAccessFlags.Write;
+                    stagingDesc.Usage = SharpDX.Direct3D11.ResourceUsage.Staging;
+                    stagingDesc.OptionFlags = SharpDX.Direct3D11.ResourceOptionFlags.None;
+                    var stagingBuffer = new SharpDX.Direct3D11.Buffer(GraphicsDevice._d3dDevice, stagingDesc);
 
-                        lock (GraphicsDevice._d3dContext)
-                            d3dContext.CopyResource(_buffer, stagingBuffer);
+                    lock (d3dContext)
+                    {
+                        d3dContext.CopyResource(_buffer, stagingBuffer);
 
                         // Map the staging resource to a CPU accessible memory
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
```diff
@@ -138,18 +138,52 @@ void GenerateIfRequired()
                 var startBytes = startIndex * elementSizeInBytes;
                 var dataPtr = (IntPtr)(dataHandle.AddrOfPinnedObject().ToInt64() + startBytes);
 
-                var box = new SharpDX.DataBox(dataPtr, 1, 0);
+                var d3dContext = GraphicsDevice._d3dContext;
+                lock (d3dContext)
```
tomspilman: I'm confused...  why did this lock need to move before the region/box creation?

MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
```diff
@@ -138,18 +138,52 @@ void GenerateIfRequired()
...
+                    {
+                        // Copy the buffer to a staging resource, so that any elements we don't write to will still be correct.
+                        var stagingDesc = _buffer.Description;
+                        stagingDesc.BindFlags = SharpDX.Direct3D11.BindFlags.None;
+                        stagingDesc.CpuAccessFlags = SharpDX.Direct3D11.CpuAccessFlags.Read | SharpDX.Direct3D11.CpuAccessFlags.Write;
+                        stagingDesc.Usage = SharpDX.Direct3D11.ResourceUsage.Staging;
+                        stagingDesc.OptionFlags = SharpDX.Direct3D11.ResourceOptionFlags.None;
+                        var stagingBuffer = new SharpDX.Direct3D11.Buffer(GraphicsDevice._d3dDevice, stagingDesc);
+
+                        lock (GraphicsDevice._d3dContext)
```
tomspilman: Why is there a lock inside the same lock?

MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
```diff
@@ -138,18 +138,52 @@ void GenerateIfRequired()
...
+                        var box = d3dContext.MapSubresource(stagingBuffer, 0, SharpDX.Direct3D11.MapMode.Read, SharpDX.Direct3D11.MapFlags.None);
+
+                        for (int i = 0; i < data.Length; i++)
+                            SharpDX.Utilities.CopyMemory(box.DataPointer + i * vertexStride + offsetInBytes, dataPtr + i * elementSizeInBytes, elementSizeInBytes);
+
+                        // Make sure that we unmap the resource in case of an exception
+                        d3dContext.UnmapSubresource(stagingBuffer, 0);
+
+                        // Copy back from staging resource to real buffer.
+                        lock (GraphicsDevice._d3dContext)
```
tomspilman: Another lock within a lock to the same resource.

Test/Framework/Visual/VertexBufferTest.cs
```diff
@@ -121,7 +121,36 @@ public void GetPosition(bool dynamic)
                 Assert.AreEqual(savedData[2].Position, readData[2]);
                 Assert.AreEqual(savedData[3].Position, readData[3]);
             };
-            Game.RunOneFrame();
+            Game.Run();
```
tomspilman: I'm not sure why this works for you.

`Run()` would run the game and not return until Exit() is called.  `RunOneFrame()` is supposed to run one fame and return.

So if `RunOneFrame()` isn't working for you we need to fix it.  We shouldn't switch things to `Run()`.

Test/Framework/Visual/VertexBufferTest.cs
```diff
@@ -121,7 +121,36 @@ public void GetPosition(bool dynamic)
...
+        }
+
+        //[TestCase(true)]
```
tomspilman: I guess the dynamic case will be enabled once we support stride in `DynamicVertexBuffer.SetData`?

MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
tgjones: I was trying to simplify it by just having one lock, but then I messed up by copying-and-pasting several more locks into the "else" block. I'll fix it by going back to more atomic locking.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.DirectX.cs
tgjones: That's my bad (although nested locks on the same object don't have any effect, but it's still untidy). I'll fix it.

Test/Framework/Visual/VertexBufferTest.cs
tgjones: It works because `VisualTestFixtureBase` sets `_game.ExitCondition = x => x.DrawNumber > 1;`.

Test/Framework/Visual/VertexBufferTest.cs
tgjones: Yes - although these tests also use `GetData`, which isn't implemented at all for dynamic vertex buffers in `VertexBuffer.DirectX.cs`.

Test/Framework/Visual/VertexBufferTest.cs
tgjones: Hmm, in the XNA test project, no window at all is created for tests which call `RunOneFrame`, and because there is no window, `Draw` is never called. So all those tests, specifically those in IndexBufferTest, Texture2DTest, Texture3DTest, VertexBufferTest, result in false positives.

Tests which call `Run`, and rely on the the `ExitCondition`, do work correctly in the XNA test project - they open a window, and call `Draw` as expected.

Will keep investigating...

Test/Framework/Visual/VertexBufferTest.cs
tomspilman: > although these tests also use `GetData`, which isn't implemented at all for dynamic vertex buffers 

We can maybe implement `GetData`... although I don't want to encourage people to use it on any regular basis.

Another approach is to validate the results by rendering something.

Test/Framework/Visual/VertexBufferTest.cs
tomspilman: What is weird is that i'm not having that problem here.  I can run `Texture2DTest` for instance and it seems to be working.

How are you running the tests?  Maybe we are doing it differently?

Test/Framework/Visual/VertexBufferTest.cs
tgjones: Presumably we do eventually want `GetData` for dynamic vertex buffers (and `SetData` with a `vertexStride`), for XNA compatibility, whether or not it's a good idea to actually use it? But I was hoping to leave that for the future - I haven't ever used `DynamicVertexBuffer` myself.

Test/Framework/Visual/VertexBufferTest.cs
tgjones: You mean you can step into, say, `Texture2DTest.ShouldSetAndGetData`, and see it actually go into the `Game.DrawWith` event handler?

I'm using ReSharper. Should I be using something else?

tgjones:
Do you want me to do any more to this? I can revert to using `Game.RunOneFrame` if you prefer, although as far as I can tell, that means the tests won't (really) be run under XNA, and I haven't yet found any way of making it work. Did you say that you can actually see it going into the `DrawWith` event handler under XNA?

`Game.Run()` does essentially the same thing as `Game.RunOneFrame()`, because  `VisualTestFixtureBase` sets `_game.ExitCondition = x => x.DrawNumber > 1`.

tgjones:
I've just pushed an experiment, to see if `RunOneFrame` works under XNA on the build server. Please don't merge this pull request yet!

tgjones:
Well, that test passed on TeamCity, so I guess I'm just running the tests wrong!

Reading [this](https://github.com/mono/MonoGame/blob/develop/Test/README.md) more closely indicates that I shouldn't be using ReSharper to run the tests, but when I run the `MonoGame.Tests.XNA` project as a console app, these `RunOneFrame` tests still don't work properly for me (and lots of other tests fail too). Maybe I should use the NUnit runner GUI? @tomspilman, how do you run MonoGame tests? I assume I'm just making some kind of beginner mistake.

(I've removed that experimental commit, so this pull request is good to go again.)

tomspilman:
> Do you want me to do any more to this? 

Let me do some tests here.

For sure we want to fix `Game.RunOneFrame` if it is broken... we don't want to replace it.


> I've just pushed an experiment, to see if `RunOneFrame` works under XNA on the build server

The build server does not execute the XNA tests.  So your experiment didn't do anything really.


> Reading this more closely indicates that I shouldn't be using ReSharper to run the tests,

 I use ReSharper to run tests... it is generally fine, but there are some rare times where I see a random failure that I can't account for.

tgjones:
I have gone back to `RunOneFrame` for these tests. Is there any chance this PR can be merged, and then we deal with the `RunOneFrame` / XNA thing as a separate issue? It affects more than just these tests.

tomspilman:
Yeah...  that makes sense.  I've just been so busy I've not been able to get into it.

mgarstenauer:
> it is only used when vertexStride != elementSize, so it doesn't affect existing usage.

Not quite true. I can no longer load simple models using the `ContentManager`. The `VertexBufferReader` calls 

    buffer.SetData( data );

where `data` is a byte buffer. Since byte != vertexStride it now runs into the new branch and causes an AccessViolationException in `SharpDX.Utilities.CopyMemory`.

@tgjones, @tomspilman Did you try loading a model using the content pipeline?

tgjones:
Oops. I'll take a look at this.

mgarstenauer:
Perhaps `VertexBuffer.SetData<T>(T[])` is calling `SetDataInternal<T>` with the wrong parameters?

tgjones:
I think you're right - maybe we should pass some kind of "null" placeholder value for `vertexStride`, to tell it to use the original code path? Since `vertexStride` should only be used when you explicitly specified it.

I copied that conditional from the `GetData` code, so I wonder if it has the same bug?

tgjones:
Either way, this sounds a good time for a test case :) I'll write a failing test, and see if @tomspilman weighs in with his thoughts on the best solution.

tgjones:
It's getting towards my late evening, so I might not have time to fix this today. Here's a failing test:

``` csharp
//[TestCase(true)]
[TestCase(false)]
public void ShouldSetAndGetDataBytes(bool dynamic)
{
    Game.DrawWith += (sender, e) =>
    {
        var vertexBuffer = (dynamic)
            ? new DynamicVertexBuffer(Game.GraphicsDevice, typeof(VertexPositionTexture), savedData.Length, BufferUsage.None)
            : new VertexBuffer(Game.GraphicsDevice, typeof(VertexPositionTexture), savedData.Length, BufferUsage.None);
        var savedDataBytes = ArrayUtil.ConvertFrom(savedData);
        vertexBuffer.SetData(savedDataBytes);

        var readData = new VertexPositionTexture[4];
        vertexBuffer.GetData(readData, 0, 4);
        Assert.AreEqual(savedData, readData);
    };
    Game.RunOneFrame();
}
```

And it relies on this utility method:

``` csharp
public static byte[] ConvertFrom<T>(T[] source) where T : struct
{
    var sizeOfSource = Marshal.SizeOf(typeof(T));
    var count = source.Length;
    var dest = new byte[sizeOfSource * count];

    var pinned = GCHandle.Alloc(dest, GCHandleType.Pinned);
    var pointer = pinned.AddrOfPinnedObject();

    for (var i = 0; i < count; i++, pointer += sizeOfSource)
        Marshal.StructureToPtr(source[i], pointer, true);

    pinned.Free();

    return dest;
}
```

I think we could pass `-1` as the `vertexStride` from the two overloads of `SetData` that don't have a `vertexStride` parameter, and check for that value in `PlatformSetDataInternal`. But that won't work as-is because `VertexBuffer.SetDataInternal` is doing some range-checking on `vertexStride`.

mgarstenauer:
@tgjones Thx for the quick response.

> I copied that conditional from the GetData code, so I wonder if it has the same bug?

`GetData<T>(T[])` seems to work.

        public void GetData<T>(T[] data, int startIndex, int elementCount) where T : struct
        {
            var elementSizeInByte = Marshal.SizeOf(typeof(T));
            this.GetData<T>(0, data, startIndex, elementCount, elementSizeInByte);
        }

It passes `elementSizeInByte` instead of `VertexDeclaration.VertexStride`. I suggest we do the same in SetData<T>(T[]). 

Except then `SetDataInternal<T>` runs into this exception:

            if (... || (vertexStride < VertexDeclaration.VertexStride))
                throw new ArgumentOutOfRangeException("... The vertex stride is too small for the type of data requested.");

Is this check required? If I pass elementSizeInByte in `SetData<T>(T[])` and remove this check in `SetDataInternal<T>`, it works again.

tgjones:
@mgarstenauer No problem - I'm embarrassed that I missed this bug.

That sounds like a sensible solution. I think that check is implementing the rules defined under "Exceptions" [here](http://msdn.microsoft.com/en-us/library/bb975969.aspx). But I'm not at all sure that it's checking the right thing.

tomspilman:
I can't say I know off the top of my head what is the right solution here.

First step is to get a few tests in there which we can verify work in XNA and fail under MonoGame.  From that we should be able to work out the right fix.

The docs are not clear, but I guess the `vertexStride` parameter passed to this method...

 http://msdn.microsoft.com/en-us/library/bb975969(v=xnagamestudio.40).aspx

... is not meant to actually be `VertexDeclaration.VertexStride` as the VB already knows that.  It must be there to define how the `data` is supposed to be copied.  Whatever that behavior is should tell us what to pass to `vertexStride` for the `SetData` methods that don't define one explicitly.

tgjones:
I can put some tests together tomorrow.

In the meantime, in case it helps: the reason I worked on this in the first place is, in XNA, I was passing just the position components (Vector3[]), or just the normals, etc., of my vertices, with a byteOffset, and with vertexStride set to the same as the declaration.

I'm struggling to think why you'd use something else for vertexStride. Perhaps it's left over from when you could stuff multiple sets of vertices into the same buffer, and draw them with different declarations, in XNA 3.x and below.

tgjones:
I went straight to the source, and asked Shawn Hargreaves about this. [Here is his reply](https://twitter.com/shawnhargreaves/status/553258905589907456). As you can see from my reply to that, I didn't completely understand what he meant, but thinking about it more: I *think* that `vertexStride` is a poor name. I think what it actually is is "source data stride". If so, then I understand what he meant about padding.

I haven't had a chance to write tests for this today, but I'll definitely get to it tomorrow.

tomspilman:
Thanks @tgjones.  Be sure to prioritize this over other PRs...  if we don't get a fix tomorrow i'll have to roll back your changes.

PR #3357, opened 2015-01-05
Add PlaneTest file to XNA test project

tgjones:
(Fortunately, the new `Plane.Transform` methods seem to match XNA!)


tomspilman:
Great... thanks for adding it!

I should maybe look into adding the XNA tests into the automated build soon.

PR #3359, opened 2015-01-05
Linux Monogame Installer[WIP]

cra0zy:
So the installer is a little too big, it should probably be added to an archive for compression before adding it to a download page. Do tell me what I can do to improve this.


cra0zy:
@tomspilman Can you move the linux files to be built using the mac server?

tomspilman:
> Can you move the linux files to be built using the mac server?

Like what... all the Linux code?  I don't know if they are any negative implications to that.  @dellis1972 ??

tomspilman:
So maybe I need a bit of explanation as to what is happening here.

The .csproj here is an EXE that is actually doing the install?  It has embedded within it all the binaries and stuff?

dellis1972:
not off the top of my head. In theory it should work

cra0zy:
Don't merge this, this is also WIP, but the idea is to reuse the code I wrote for makeself, basically this generates an exe that will self extract itself to /tmp and launch install.sh. If the linux code is moved to mac server then I can use makeself to generate native .run package for linux.

cra0zy:
If linux stuff remain on windows server, merge this, if you manage to move linux stuff to mac server, I'll make adjustments to this so it generates native .run files.

cra0zy:
@tomspilman if moving linux stuff to mac server is a bad idea, I could compile linux installer of monogame for every major release/major change using my pc, since I have linux, and then send it to you

dellis1972:
I can try to create a PR which builds the linux stuff on mac over the
weekend, we can see what happens then.
But we will need to update the Windows installer as well to get the linux
stuff from the mac server.

cra0zy:
@dellis1972 Thanks for the info.

tomspilman:
> But we will need to update the Windows installer as well to get the linux stuff from the mac server.

That is the complication.  We package Linux binaries for Windows right now... so we would need to copy the artifacts from the Mac build over to the Windows build to include them into the installer.

That said... is this useful?  How many people build Linux support on Windows?  You can't test it on Windows... so why have a template and support building it?

tomspilman:
So should I merge this for now?

cra0zy:
@tomspilman let me add makeself to dependencies, then merge

cra0zy:
@tomspilman all done, you can merge it, also if you want to connect something to the downloads page temporarily as a linux installer, before the process gets automated, use this: https://mega.co.nz/#!yBoSDIBL!eX7DpV1HrjNedWZq0nt93kDAcVANRO2gegrQMyOS2n8

tomspilman:
Ok... merging!

tomspilman:
@dellis1972 @cra0zy @KonajuGames - How useful is it to actually include a Linux template and Linux binaries on the Windows installer?  Has this ever been useful?

cra0zy:
It's nice to be able to compile linux binary from windows, but I don't think it's very useful, you still have to start up linux to test your game on it.

PR #3361, opened 2015-01-05
Rework of Xamarin Studio/MonoDevelop Addin

dellis1972:
Includes new templates and new Shared Projects support
Includes Pipeline Tool support
Removes old ProjectTemplates

This is just a Test PR for now to check for any build issues.

Also worth noting that once this is in , the old MonoDevelopContent addin will no longer be needed and will also be removed probably as part of the PR as well. This is because we don't want to support the old xna content projects moving forward , only the .mgcb ones which this new addin will take care of. 

I do need some guidance on where the Pipeline.exe will be installed on linux.. so we can start it correctly from the IDE.


dellis1972:
Note all the templates are now Nuget based. But we do include the nuget package in the addin so it will work offline if required (this is how Xamarin Studio handles nuget).

I will also look at packaging this addin with the Pipeline Mac installer.

tomspilman:
> I will also look at packaging this addin with the Pipeline Mac installer.

Will this then replace the mpack we have for MonoDevelop then?

dellis1972:
It can replace the mpack. I think having a proper Mac installer is more important now (same for linux)

cra0zy:
On linux I made it so that the path where the pipeline will be installed can be changed during the installation, but launching /bin/monogame-pipeline will always launch monogame pipeline(or typying monogame-pipeline in terminal). Hope it helps.

cra0zy:
@tomspilman Sorry for asking again, but can you make mac server do the build of linux files, If you can, then I can make a proper linux installer, if not, the PR I submitted is closes thing to installer I can make windows build.

dellis1972:
@cra0zy could you add a script to run the mgcb as well? e.g /bin/mgcb

cra0zy:
@dellis1972 already did

dellis1972:
@cra0zy awesome :)

tomspilman:
> It can replace the mpack. I think having a proper Mac installer is more important now (same for linux)

Great!

dellis1972:
@tomspilman ok some more changes. First is the location of the Pipeline.Installer.pkg it will now be in 

Installers/Pipeline.Installer.pkg

Now given that it now includes installing the Xamarin Studio Addin, I'm thinking of renaming the file to just MonoGame.Installer.pkg … comments?

I also moved the building of the installer into the Installers directory default.build file which makes more sense. 

I had to make some changes to the start up code of the Pipeline app on mac because it was always getting a arg like -psn_0_xxxxxx which was messing up the auto open. 

The integration with XS is done, if you have a .mgcb file in a project and double click on it , it will open in the Pipeline app :) 

The new Templates have all been moved to the IDE/MonoDevelop folder. I have done, Mac, Android, iOS, iOS+Android+Shared, there is a Windows/Linux project as well. All the new templates use nuget to get the binaries. We ship the current stable nuget package with the addin so it will work offline.

A Video of the XS integration is over at http://www.screencast.com/t/ygygS0JOw

dellis1972:
Still have a few bits to tidy up then we can look at merging this in

tomspilman:
> thinking of renaming the file to just MonoGame.Installer.pkg … comments?

 Seems great to me.

> The new Templates have all been moved to the IDE/MonoDevelop folder

 I guess that means that the Xamarin Studio templates are still an IDE extension?

dellis1972:
Yes, and extension is the only way to do templates in XS and MonoDevelop :-/

-----Original Message-----

cra0zy:
@dellis1972 I'm guessing I should add installation of xs addin to pipeline installer for linux, just 1 quick question, did you just setup copying the addin to xs addin folder or did you do something more?

tomspilman:
@dellis1972 - How long before you think this can be ready?

dellis1972:
@tomspilman its a bit of a catch 22. It is relying on Nuget packages, but we need the new ones now especially since iOS is now Unified rather than classic as the framework versions changed from MonoTouch to Xamarin.iOS. We need to coordinate a Nuget release at the same time so we can make sure this all works. So we'll need to chat with @ddreaper so make sure I get the package names right etc.

tomspilman:
> It is relying on Nuget packages

I'm confused.

How is this an installer for Mac if it depends on NuGet?  Are the assemblies coming from the installer or from NuGet?

dellis1972:
The new XS templates are designed to use Nuget to get their references.
This is good for a couple of reasons , updates and it means that the
references will actually work. With the old addin, because of the way XS
works we could not auto add references for all the project types without
putting in some horrible hacks (see
https://github.com/dellis1972/MonoGameMonoDevelopAddin/blob/master/MonoDevelop.MonoGame/MonoGameProject.cs#L119).
Also this only worked on "MonoGame/XNA" project types, it cannot work for
Android/iOS or Mac because those project types are Xamarin ones not ours.
We might be able to derive new project types for Android/iOS but we can't
for Mac as the MonoMac project type is sealed as is the Xamarin.Mac one.

XS has built in support for Nuget so the new templates use that, but to
support offline installation we package the "latest" nuget packages with
the Addin so that if the user is offline they can still work.

We could go down the route of putting MonoGame in
/Library/Frameworks/MonoGame.framework/ which would be nice...but that will
require allot more research/work and things like MSBuild tasks, integration
with mono so that the assemblies can in the GAC/can be resolved without all
the hacks.

If we want this in the 3.3 release but not with Nuget, I'll have to put in
those hacks and we'll have to accept that the templates won't build out of
the box because of missing assembly references for projects like Mac/iOS
and Android. Or just put in basic templates and let the users figure out
the references themselves.

tomspilman:
> The new XS templates are designed to use Nuget to get their references.

That sounds horrible to me honestly.  I really hate this NuGet-everything kick people seem to be on.

That means your project references can be out of sync with the Pipeline tool, documentation once we start installing that, any other stuff we package together because they are versioned together.


> to support offline installation we package the "latest" nuget packages
> with the Addin so that if the user is offline they can still work.

As long as the installer is doing that I think it is fine.

This means once we get the TeamCity server generating the NuGet packages we will have to ensure they are included in the Mac installer.

dellis1972:
I am also wary of Nuget, I've been bitten by server outages before. That
said for this particular problem getting round bugs in MonoDevelop/XS and a
cross platform references solution, it works. Getting that assembly
resolving code working cross platform (since MD/XS runs on Mac/Linux &
Windows) was a nightmare. Nuget solves that problem..

I'll try and think of some other ways to fix this references issue if I
can.. but I'd rather not get into the whole one mpack for each platform
again as it was a PITA to maintain. There should be one set of templates
that work in XS/MD on all the platforms.

DDReaper:
I share your sentiment @dellis1972 , as good as NuGet is for keeping things up to date and letting you know about updates. If you twant to use it offline (even if you have downloaded the package before) you are stuffed.

SO packaging a version of the NuGet package complete with the template is a fantastic idea.

Keeping things similar, should this also be done for the other templates we ship on windows, to keep conformity?

tomspilman:
> Keeping things similar, should this also be done for the other templates we ship on windows

IMO no...  there is nothing broken with how to the templates work now on Windows.

At the very least we should wait to see what fallout if any occurs from using this on Mac/Linux.

dellis1972:
I have uploaded the .mpack for the new add and templates , I could use
someone testing it on linux.

https://dl.dropboxusercontent.com/u/18881050/MGCB/MonoDevelop.MonoGame_3.2.0.mpack

Under linux you will probably only get a couple of templates

cra0zy:
pick me! pick me!

dellis1972:
@cra0zy go for it :)

cra0zy:
Tested it and here are the notes:
<ul>
<li>Visible projects are OpenGL and Shared Project</li>
<li>For opengl project OpenTK and Tao.Sdl config files aren't getting copied, this makes game not start but instead start complaining about sdl.dll missing</li>
<li>There are sdl.dll and sdl_mixer.dll files in the project directory</li>
<li>Shared project doesn't have a reference to monogame</li>
</ul>

dellis1972:
@cra0zy simon will sort the .config files are they are part of the nuget
package, so are the sdl.dll files not sure what we can do with that.
The Shared project does not have a reference, the idea behind a shared
project is you reference it from another project which provides those
references.
I can add a OpenGL / Shared project combination which will make is easier
for devs

DDReaper:
If shared projects on iOS/Mac are the same as the shared projects on Windows, then you don't put DLL's like MG in there.  The shared project is able to read from references in each platform and you only put common code in there.
Each platform would have the specific version of MG for that platform installed.

Question is, are the .config files also needed for MacOS, this would be a bit of a problem as I wont be able to automate the setting of the CopyAction (since I'm told PS won't work) but I can include them, so  setting them to copy would be manual.

cra0zy:
@dellis1972 Could you include the old MonoGameProject.cs class as MonoGameProjectLegacy.cs, just so that the old projects are openable?

dellis1972:
Its the same class so they will still be openable. or at least they should
be :)

cra0zy:
sweet :)

cra0zy:
Is the addin complete now?

dellis1972:
Needs testing :)

-----Original Message-----

DDReaper:
So where would we find the new update to test @dellis1972 ? Looked on TeamCity and couldn't see it.

dellis1972:
It will be part of the Mac archives

http://teamcity.monogame.net/viewLog.html?buildId=5267&buildTypeId=MonoGame_DevelopMac&tab=artifacts#!xvtkz

dellis1972:
Darn , its not... doh..

dellis1972:
I'll have to upload a new one to drop box.

DDReaper:
Thank goodness for that, thought I was losing it there.

Saw the MonoDevelop stuff disappear of teamcity and even the main mg downloads page a while back. Must have been cleaned up

eindoofus:
I have the following installed in OS X: Mono Framework, MonoDevelop, MonoGame Pipeline Tool for Mac, and Xamarin Studio. Downloaded the zip at http://teamcity.monogame.net/viewLog.html?buildId=5267&buildTypeId=MonoGame_DevelopMac&tab=artifacts#!xvtkz and tried to install the mpack file using the Xamarin Studio Add-In Manager. Received the following error:
> The installation failed! 
> The required addin 'MonoDevelop.Autotools,4.0' is disabled.

Within the Add-In Manger there is now a gray "MonoGame Content Support" icon and clicking on it displays the following:

> ! This add-in can't be loaded due to missing dependencies
> MonoGame Content Support
> Version 3.0.0
>
> The following depedencies required by this add-in are not available:
> Required: MonoDevelop.Core v4.0, found v5.7.1
> Required: MonoDevelop.Ide v4.0, found v5.7.1
> MonoGame Compiled Content Support

I appreciate the work community is doing to make this framework available to coders such as myself, but being new to game development I've just spent most of the weekend trying to figure out how to get the Mono.Sample Platformer2D to run on my Mac and have yet to succeed at such a trivial task. 

I've followed every posted solution available and to me it seems to me that at the current moment MonoGame is not stable as a cross-platform solution.

xanather:
@eindoofus The mpack is not exactly required, it is possible to compile MonoGame yourself (Github allows you to download the .zip file) and then reference the compiled .dll's needed for running the game (i.e. the compiled MonoGame.Framework.dll). Make sure you have the required dependencies for the target platform https://github.com/Mono-Game/MonoGame.Dependencies, put them into the output directory of your project.

I think MonoGame is currently transitioning to fully supporting Content compiling, I feel that creating a add-in for such a thing can be difficult.

dellis1972:
ok latest addin uploaded for testing

https://www.dropbox.com/s/73eeppcx1g3uycl/MonoDevelop.MonoGame_3.3.0.mpack?dl=0

cra0zy:
Small problem with monogame pipeline library and PCL, they both can't find nuget packages. Also is there a way to put OpenTK.dll.config and Tao.Sdl.config to copy if newer action automatically?

dellis1972:
I can't set the copy if newer action in the templates. I'll check the PCL
and pipeline references

dellis1972:
OK, I figured out how to set the Copy if Newer value :)

dellis1972:
the PCL issue is because the nuget package id has a -beta rather than -Beta in it. This probably doesn't matter on windows but on Mac and Linux it means the package can't be found.

DDReaper:
Crafted some new NuGet's as per @dellis1972 request. with the -Beta fix in.
Using updated dll's as well.

dellis1972:
OK new build 

https://www.dropbox.com/s/73eeppcx1g3uycl/MonoDevelop.MonoGame_3.3.0.mpack?dl=0

should fix the Pipeline project. might fix the PCL one too, but I get errors locally about the profile not being supported but I think that is just my system.

Should also have the dll.config copy if newer issue fixed. But because the nuget also provides those files you get a warning about a file being updated outside the ide :/

cra0zy:
On linux I had to install referenceassemblies-pcl to get PCL working, maybe mac and windows have something similar.

DDReaper:
Right did some quick tests in XamarinStudio (the plugin didn't like my Unity version of MD)
Most projects worked completely fine (Although it found a MonoGame.Framework.Android V3.3.0 which doesn't exist on NuGet??)

The pipeline project installs but when it looks for the package on NuGet it simply states 
> Unable to find package 'MonoGame.Framework.Content.Pipeline.Portable'.

Can't see any other debug messages unless there is a better way of getting a detailed log out of XS

Other than that it was all fine.

cra0zy:
@dellis1972 I think you should put this PR up for merging. If there are bugs you won't be the only one who will have to solve them, you would get help.

dellis1972:
This is ready for merging I think, just waiting on @tomspilman  or @KonajuGames to review and merge it :)

cra0zy:
Oh, can't wait for this to get merged :D

IDE/MonoDevelop/MonoDevelop.MonoGame.iOS/templates/Shared/Activity1.cs
```diff
@@ -9,8 +9,9 @@
 using Android.OS;
 
 using Microsoft.Xna.Framework;
+using ${ProjectName};
 
-namespace ${Namespace}
+namespace ${ProjectName}.Droid
```
KonajuGames: Why the `.Droid` addition to the namespace?  It just necessitates the additional `using` statement above it, and "Droid" is not an abbreviation that Google approve of for "Android".

Also, a project name might not be a valid namespace if I remember correctly.

IDE/MonoDevelop/MonoDevelop.MonoGame.iOS/templates/Shared/Program.cs
```diff
@@ -0,0 +1,83 @@
...
+using MonoMac.Foundation;
+#elif __IOS__
+using Foundation;
+using UIKit;
+#endif
+using ${ProjectName};
+#endregion
+
+
+namespace ${ProjectName}.iOS
```
KonajuGames: The platform addition to the namespace again.

Same with the project name might not be a valid namespace.

IDE/MonoDevelop/MonoDevelop.MonoGame.Android/Properties/AssemblyInfo.cs
```diff
@@ -1,27 +1,22 @@
...
-[assembly: AssemblyCopyright("MonoGame Team")]
-[assembly: AssemblyTrademark("")]
-[assembly: AssemblyCulture("")]
-
+[assembly: AssemblyTitle ("MonoDevelop.MonoGame.Android")]
+[assembly: AssemblyDescription ("")]
+[assembly: AssemblyConfiguration ("")]
+[assembly: AssemblyCompany ("")]
+[assembly: AssemblyProduct ("")]
+[assembly: AssemblyCopyright ("Dean")]
```
KonajuGames: Shouldn't the copyright be left as "MonoGame Team"?

IDE/MonoDevelop/MonoDevelop.MonoGame.Android/MonoDevelop.MonoGame.Android.csproj
```diff
@@ -0,0 +1,107 @@
...
+    <Optimize>true</Optimize>
+    <OutputPath>..\bin\Release</OutputPath>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <ConsolePause>false</ConsolePause>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="MonoDevelop.Ide">
+      <HintPath>..\..\..\..\..\..\..\..\..\Applications\Xamarin Studio.app\Contents\Resources\lib\monodevelop\bin\MonoDevelop.Ide.dll</HintPath>
```
KonajuGames: All the `..\..\..\..\` would be specific to your machine, and `Applications\Xamarin Studio.app` is specific to Mac. How does this work on another directory structure and Linux/Windows?

KonajuGames:
The Android splash screen is very low resolution and dull grey.  Can it use the same white splash as the VS Android template? And default the project to landscape as all other templates do.
MonoGame\ProjectTemplates\VisualStudio2010\Android\Resources\Drawable\Splash.png

I would like to remove the small "mono for Android" logo from the lower-left though.

IDE/MonoDevelop/MonoDevelop.MonoGame.iOS/Properties/AssemblyInfo.cs
```diff
@@ -1,26 +1,22 @@
...
-[assembly: AssemblyCopyright("${AuthorCopyright}")]
-[assembly: AssemblyTrademark("${AuthorTrademark}")]
-[assembly: AssemblyCulture("")]
-
+[assembly: AssemblyTitle ("MonoDevelop.MonoGame.iOS")]
+[assembly: AssemblyDescription ("")]
+[assembly: AssemblyConfiguration ("")]
+[assembly: AssemblyCompany ("")]
+[assembly: AssemblyProduct ("")]
+[assembly: AssemblyCopyright ("Dean")]
```
KonajuGames: Is this the assembly info for the plug-in or the template for new projects?  If for the plug-in, the copyright should be "MonoGame Team" like our standard file header.  If for a template, it should be the token so the IDE fills it in as appropriate for the user.

IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+#if MONOMAC
+using MonoMac.AppKit;
+using MonoMac.Foundation;
+#elif __IOS__
+using MonoTouch.Foundation;
+using MonoTouch.UIKit;
+#endif
+#endregion
+
+namespace ${ProjectName}
```
KonajuGames: Should be `${Namespace}`?

Tools/Pipeline/Program.cs
```diff
@@ -45,11 +45,19 @@ static void Main(string [] args)
...
+			#if LINUX
 			if (args != null && args.Length > 0)
 			{
 				var projectFilePath = string.Join(" ", args);
 				win.OpenProjectPath = projectFilePath;
 			}
+			#elif MONOMAC
+			var project = Environment.GetEnvironmentVariable("MONOGAME_PIPELINE_PROJECT");
+			if (!string.IsNullOrEmpty (project)) {
+				System.IO.File.WriteAllText("/Users/dean/Pipeline.log", string.Format ("ENV : {0}", project));
```
KonajuGames: Debugging code that shouldn't be here?

KonajuGames:
That is a /lot/ of files and changes to look through. :)

cra0zy:
https://www.youtube.com/watch?v=-dDbSyfU_04

IDE/MonoDevelop/MonoDevelop.MonoGame.Android/MonoDevelop.MonoGame.Android.csproj
dellis1972: I'll fix up the paths for Mac, BUT this addin is only built on Mac, not on linux or windows. The resulting .mpack can then be used on all platforms.

cra0zy: Wouldn't it be easier to include MonoDevelop.Ide.dll in MonoGame Dependencies?

dellis1972:
Closing this PR, reopened a new one

IDE/MonoDevelop/MonoDevelop.MonoGame.Android/MonoDevelop.MonoGame.Android.csproj
dellis1972: No. Its constantly updated.

Also looks like some recent changes to the pipeline app on Linux have completely broken it on mac. We shouldn't be including .so files from a Linux box for the mac deployment. I get a DLL not found error now , its looking for libc.dylib for some reason :-/

-----Original Message-----

cra0zy: 1) so libs
Ups, .so files were only meant for linux, I accidentally left MacOS(copy pasted using configs above). I'll setup a PR that fixes that right away.

2) MonoDevelop.Ide.dll
Even if it is constantly updated, we would not need to update it that often because it does not use a local copy... then again there is no way to disable local copy with protobuild and Mac server will probably do the build anyway...

dellis1972: i don't know whats causing the crash but here is the stack trace... 

Unhandled Exception:
System.DllNotFoundException: libc.dylib
  at (wrapper managed-to-native) Mono.Unix.Native.Stdlib:free (intptr)
  at Mono.Unix.UnixMarshal.FreeHeap (IntPtr ptr) [0x00000] in <filename unknown>:0 
  at Mono.Unix.Catalog.GetString (System.String s) [0x00000] in <filename unknown>:0 
  at MonoGame.Tools.Pipeline.MainWindow.Build () [0x00000] in <filename unknown>:0 
  at MonoGame.Tools.Pipeline.MainWindow..ctor () [0x00000] in <filename unknown>:0 
  at MonoGame.Tools.Pipeline.Program.Main (System.String[] args) [0x00000] in <filename unknown>:0 
[ERROR] FATAL UNHANDLED EXCEPTION: System.DllNotFoundException: libc.dylib
  at (wrapper managed-to-native) Mono.Unix.Native.Stdlib:free (intptr)
  at Mono.Unix.UnixMarshal.FreeHeap (IntPtr ptr) [0x00000] in <filename unknown>:0 
  at Mono.Unix.Catalog.GetString (System.String s) [0x00000] in <filename unknown>:0 
  at MonoGame.Tools.Pipeline.MainWindow.Build () [0x00000] in <filename unknown>:0 
  at MonoGame.Tools.Pipeline.MainWindow..ctor () [0x00000] in <filename unknown>:0 
  at MonoGame.Tools.Pipeline.Program.Main (System.String[] args) [0x00000] in <filename unknown>:0

cra0zy: I have no idea of whats causing the exception... Also I can't test anything I find using Google for "System.DllNotFoundException: libc.dylib" because I don't have a Mac.

PR #3367, opened 2015-01-07
Check Importer and Processor DLL timestamps when deciding whether content needs rebuild

tgjones:
Fixes the issue, mentioned in #3138, where updating a `ContentImporter` or `ContentProcessor` used for a content item, didn't force a rebuild of that content item.

I have followed the implementation suggested by @tomspilman [here](https://github.com/mono/MonoGame/pull/3138#issuecomment-68819320).

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
@@ -143,7 +143,7 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
                 serializer.Serialize(textWriter, this);
         }
 
-        public bool NeedsRebuild(PipelineBuildEvent cachedEvent)
+        public bool NeedsRebuild(PipelineManager manager, PipelineBuildEvent cachedEvent)
         {
             // If we have no previously cached build event then we cannot
             // be sure that the state hasn't changed... force a rebuild.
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks @tgjones !

This looks good at first glance...  i'll look it over a bit more in the morning and merge it.

tgjones:
Great!

tomspilman:
Looks good... merging!  Thanks again!

MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
```diff
@@ -180,13 +180,19 @@ public bool NeedsRebuild(PipelineBuildEvent cachedEvent)
                 cachedEvent.DestFile != DestFile)
                 return true;
 
+            // Did the importer assembly change?
+            if (manager.GetImporterAssemblyTimestamp(cachedEvent.Importer) > cachedEvent.DestTime)
```
tomspilman: You know...  I think we might need to keep track of the assembly timestamps explicitly here instead of using the `DestTime`.  You could have a case where the assembly is rolled back to an older version with an older timestamp.

tgjones: I hadn't thought of that scenario, but yes, using `DestTime` did feel a bit brittle. I'll put together another pull request with that change.

PR #3368, opened 2015-01-07
Rectangle improvements

Shqrdx:
Changes:
 -Added missed XNA overloads to Rectangle.Contains and float version of Inflate
 -Small unification comment fixes for Color
 -RectangleTest finally !

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -296,7 +296,7 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
-        /// Gets or sets the blue component of this <see cref="Color"/>.
+        /// Gets or sets the blue component.
         /// </summary>
         [DataMember]
         public byte B
@@ -315,7 +315,7 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
-        /// Gets or sets the green component of this <see cref="Color"/>.
+        /// Gets or sets the green component.
         /// </summary>
         [DataMember]
         public byte G
@@ -334,7 +334,7 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
-        /// Gets or sets the red component of this <see cref="Color"/>.
+        /// Gets or sets the red component.
         /// </summary>
         [DataMember]
         public byte R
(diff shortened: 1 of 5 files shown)


MonoGame.Framework/Color.cs
```diff
@@ -296,7 +296,7 @@ public Color(float r, float g, float b, float alpha)
         }
 
         /// <summary>
-        /// Gets or sets the blue component of <see cref="Color"/>.
+        /// Gets or sets the blue component of this <see cref="Color"/>.
```
tomspilman: I really dislike this pattern of referencing the class itself from every member.  

Like if you can see "this Color" "this Color" "this Color" "this Color" "this Color" repeated in every member that means the information is redundant.

Saying...
```
/// Gets or sets the blue component.
```
... is enough.  The user understands it is the blue component of "this Color".

Just noticed this appearing a lot and i'd like to see us stop doing it.

MonoGame.Framework/Color.cs
```diff
@@ -1732,19 +1732,19 @@ public static Color Multiply(Color value, float scale)
...
         }
 
-    /// <summary>
-    /// Gets a three-component <see cref="Vector3"/> representation for this object.
-    /// </summary>
-    /// <returns>A three-component <see cref="Vector3"/> representation for this object.</returns>
+        /// <summary>
+        /// Gets a <see cref="Vector3"/> representation for this object.
+        /// </summary>
+        /// <returns>A <see cref="Vector3"/> representation for this object.</returns>
```
tomspilman: This one has the generic and unhelpful "for this object".  Instead we should give them useful information here:
```
/// Returns the color in a 0-to-1 normalized <see cref="Vector3"/> format.
```
Is a much more useful thing to say here.

I guess my greater point is we shouldn't waste words...  extra words take time for us to write and for developers to read...  they all should be useful words.

tomspilman:
@Shqrdx - Did you run the rectangle tests on the XNA version of the test project?  You have to manually add the `MonoGame.Tests.XNA.csproj` project to the solution and add the test code to it.

MonoGame.Framework/Color.cs
Shqrdx: Ok..

Shqrdx:
@tomspilman Of course, I tested it with XNA, but cannot add this test to actual XNA project because it contains elements which is not existed in XNA.

tomspilman:
> cannot add this test to actual XNA project because it contains elements which is not existed in XNA.

You can add a `#if !XNA` around the non-XNA parts.  We do that in a few other of the tests.

Shqrdx:
@tomspilman Finished !

tomspilman:
Great... merging!

PR #3371, opened 2015-01-08
Additional Rectangle imrovements

Shqrdx:
This patch is not important and works like finished brush move for painter.

Changes : 
Added missed XNA Contains overload - http://msdn.microsoft.com/en-us/library/bb447712.aspx
Change test - first we must check methods with "ref" because it can change parameter inside itself.
Added test for new overload.

### DIFF
--- a/MonoGame.Framework/Rectangle.cs
+++ b/MonoGame.Framework/Rectangle.cs
@@ -290,6 +290,16 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
+        /// Gets whether or not the provided <see cref="Rectangle"/> lies within the bounds of this <see cref="Rectangle"/>.
+        /// </summary>
+        /// <param name="value">The <see cref="Rectangle"/> to check for inclusion in this <see cref="Rectangle"/>.</param>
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Rectangle.cs
```diff
@@ -290,6 +290,16 @@ public bool Contains(Rectangle value)
         }
 
         /// <summary>
+        /// Gets whether or not the provided <see cref="Rectangle"/> lies within the bounds of this <see cref="Rectangle"/>.
+        /// </summary>
+        /// <param name="value">The <see cref="Rectangle"/> to check for inclusion in this <see cref="Rectangle"/>.</param>
+        /// <param name="result"><c>true</c> if the provided <see cref="Rectangle"/>'s bounds lie entirely inside this <see cref="Rectangle"/>; <c>false</c> otherwise. As an output parameter.</param>
+        public void Contains(ref Rectangle value,out bool result)
```
tomspilman: I don't see the point to a `out bool` here.  It can simply return `bool` and be just as efficient.

Shqrdx: http://msdn.microsoft.com/en-us/library/bb447712.aspx

tomspilman: Weird... didn't expect that to be an XNA API.

PR #3372, opened 2015-01-08
Cache timestamp for importer and processor DLLs

tgjones:
and fix bug where processor timestamps weren't being set (oops) - didn't notice it before because I was testing with a single DLL that contained both the importer and processor.

Fixes potential issue raised by @tomspilman [here](https://github.com/mono/MonoGame/pull/3367#discussion_r22638244).


tomspilman:
Looks good... merging!

PR #3373, opened 2015-01-08
Simplified iOS version checks

Nezz:


tomspilman:
Seems reasonable... merging!

PR #3374, opened 2015-01-08
Modern MessageBox and KeyboardInput

### LINKED ISSUE #3313: Guide doesn't work on WP8.1
Because BeginInvoke is not supported.

Nezz:
Uses async/await
Implemented for Android, iOS, Windows Phone 8, Windows 8 and Windows Phone 8.1
Added cancellation support (works everywhere except on WP8 where we can make it work if we use our custom XAML implementation instead of the XNA one)
Fixes #3313


Nezz:
@tomspilman Does this look good to you?

Nezz:
This needs to be updated to work with the new Xamarin Unified API.

Nezz:
Done.

Nezz:
The build has failed because it didn't contain the other parts for the unified api. Looks like i'll have to rebase it.
```error CS0246: The type or namespace name `NSNotification' could not be found. Are you missing `MonoTouch.Foundation' using directive?```

PR #3376, opened 2015-01-09
Buffer fixes

Shqrdx:
Changes :
 -Licenses are changed to a good look
 -Exceptions are fixed
 -IsContentLost is useless, however for backward compatibility its now returns IsDisposed.
 -SuppressUserOffsetWarningForNonDXBuilds is internal API property(internal API should not use properties, only fields) so I removed it.
 -Additional check to SetData in IndexBuffer - why? - I've got several blue screens after testing a project which uses index buffer before I found a reason - and thats why I added this

### DIFF
--- a/MonoGame.Framework/Graphics/Vertices/DynamicIndexBuffer.cs
+++ b/MonoGame.Framework/Graphics/Vertices/DynamicIndexBuffer.cs
@@ -12,6 +12,8 @@ namespace Microsoft.Xna.Framework.Graphics
         /// Special offset used internally by GraphicsDevice.DrawUserXXX() methods.
         /// </summary>
         internal int UserOffset;
+
+        public bool IsContentLost { get { return IsDisposed; } }
 		
 		public DynamicIndexBuffer(GraphicsDevice graphicsDevice, IndexElementSize indexElementSize, int indexCount, BufferUsage usage) :
 			base(graphicsDevice, indexElementSize, indexCount, usage, true)
--- a/MonoGame.Framework/Graphics/Vertices/DynamicIndexBuffer.cs
+++ b/MonoGame.Framework/Graphics/Vertices/DynamicIndexBuffer.cs
@@ -1,43 +1,7 @@
-// #region License
-// /*
-// Microsoft Public License (Ms-PL)
-// MonoGame - Copyright © 2009 The MonoGame Team
-// 
-// All rights reserved.
-// 
-// This license governs use of the accompanying software. If you use the software, you accept this license. If you do not
-// accept the license, do not use the software.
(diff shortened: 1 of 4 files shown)


Shqrdx:
@tomspilman Should we keep IsContentLost for DynamicVertexBuffer and DynamicIndexBuffer or it could get safely removal ?

tomspilman:
> Should we keep IsContentLost for DynamicVertexBuffer and DynamicIndexBuffer or it could be safely removal ?

It is part of the XNA API...

 http://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.dynamicvertexbuffer_members(v=xnagamestudio.40).aspx

http://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.dynamicindexbuffer_members.aspx

So they should be kept and made to work on platforms where content can be lost.

Shqrdx:
Ok, I marked it obsolete.. its ok ?

Shqrdx:
https://github.com/mono/MonoGame/issues/3379 - this is reopened commit

tomspilman:
> Ok, I marked it obsolete.. its ok ?

Why do you think it is obsolete?

Dynamic buffers can loose data if the hardware resets.... that is why this property and the `ContentLost`...

http://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.dynamicvertexbuffer.contentlost(v=xnagamestudio.40).aspx

... events exist.

Shqrdx:
Hmm. ok, I return it to normal(e.g unchanged)

Shqrdx:
Because it only return false.

tomspilman:
> Because it only return false and do nothing more.

Doesn't mean the API shouldn't exist...  just means the API isn't implemented fully/correctly.

Shqrdx:
Ok, I just fixed bugs and doesnt change API - my PR now is correct ?https://github.com/mono/MonoGame/issues/3379

Shqrdx:
@tomspilman I just didnt notice that DX11 or OpenGL have lost dynamic buffer states.. only DX9 does

tomspilman:
DX11 can...  usually it is caused by driver failure or dragging a game window across monitors that are powered by totally separate GPUs.  All of that is very rare now days... so it doesn't come up very often as a concern.  Still we need to fix it.

OpenGL I don't know.  I suspect maybe it does, but maybe it is driver dependent.

All other platforms should just return `false`.  You can then write you game code to handle lost events and work against any platform you build for.

Shqrdx:
Yeah these classes missed ContentLost event also, where and how to put this event is a difficult question.

PR #3377, opened 2015-01-09
Implement IntermediateSerializer.Serialize

tgjones:
and supporting code, including `IntermediateWriter` and `ReflectiveSerializer.Serialize`.

Related to #2621.

I made some changes to the existing XML test files, for two reasons:

* To make them round-trippable, and
* To remove namespace alias usage from most test files, except for the one specifically related to namespaces

The code should be mostly okay, but `IntermediateSerializer.WriteNamespaces` could do with some cleaning up. That method was the most complicated to write out of all this; it took me a while to figure out XNA's behaviour.

On the bright side, I've ended up with a number of test cases dealing with nested namespaces, ambiguous namespaces, and nested ambiguous namespaces, so it shouldn't be too hard to go in and do some internal refactoring on that `WriteNamespaces` method, without too much risk of breaking anything.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/IntermediateWriter.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/IntermediateWriter.cs
@@ -12,9 +12,9 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
 {
     public sealed class IntermediateWriter
     {
-        private readonly List<object> _writtenObjects;
+        private readonly Stack<object> _currentObjectStack;
         private readonly Dictionary<object, string> _sharedResources;
-        private readonly List<ExternalReference> _externalReferences;
+        private readonly Dictionary<object, ExternalReference> _externalReferences;
         private readonly string _filePath;
 
         internal IntermediateWriter(IntermediateSerializer serializer, XmlWriter xmlWriter, string filePath)
@@ -23,9 +23,9 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
             Xml = xmlWriter;
             _filePath = filePath;
 
-            _writtenObjects = new List<object>();
+            _currentObjectStack = new Stack<object>();
             _sharedResources = new Dictionary<object, string>();
-            _externalReferences = new List<ExternalReference>();
+            _externalReferences = new Dictionary<object, ExternalReference>();
         }
 
         public XmlWriter Xml { get; private set; }
@@ -34,13 +34,14 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
 
         public void WriteExternalReference<T>(ExternalReference<T> value)
         {
-            var externalReference = new ExternalReference
-            {
-                ID = "#External" + (_externalReferences.Count + 1),
-                TargetType = typeof(T).FullName,
-                FileName = MakeRelativePath(value.Filename)
-            };
-            _externalReferences.Add(externalReference);
+            ExternalReference externalReference;
+            if (!_externalReferences.TryGetValue(value, out externalReference))
+                _externalReferences.Add(value, externalReference = new ExternalReference
+                {
+                    ID = "#External" + (_externalReferences.Count + 1),
+                    TargetType = typeof(T).FullName,
+                    FileName = MakeRelativePath(value.Filename)
+                });
 
             Xml.WriteElementString("Reference", externalReference.ID);
         }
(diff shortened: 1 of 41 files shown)


MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/IntermediateWriter.cs
```diff
@@ -24,37 +34,180 @@ internal IntermediateWriter(IntermediateSerializer serializer, XmlWriter xmlWrit
...
         public void WriteExternalReference<T>(ExternalReference<T> value)
         {
-            throw new NotImplementedException();
+            var externalReference = new ExternalReference
+            {
+                ID = "#External" + (_externalReferences.Count + 1),
+                TargetType = typeof(T).FullName,
+                FileName = MakeRelativePath(value.Filename)
+            };
+            _externalReferences.Add(externalReference);
```
tomspilman: I suspect that the same external reference can be written multiple times in the same file.  I think you need to look for duplicates and avoid adding them to the list twice.  The tests must be passing because it doesn't test writing the same reference more than once.

Test/ContentPipeline/IntermediateDeserializerTest.cs
```diff
@@ -266,27 +266,6 @@ public void MathTypes()
...
-                Assert.AreEqual(-10000000, primitiveTypes.Long);
-                Assert.AreEqual(10000000, primitiveTypes.ULong);
-                Assert.AreEqual(1234567.0f, primitiveTypes.Float);
-                Assert.AreEqual(1234567890.0, primitiveTypes.Double);
-                Assert.AreEqual(null, primitiveTypes.NullChar);
-                Assert.AreEqual(' ', primitiveTypes.NotNullChar);
-            });
-        }
-
-        [Test]
```
tomspilman: Try to avoid moving functions around within an existing source file.  It makes it very difficult to track what the actual changes are on the GitHub diff view as it will show as the entire function as been deleted and re-added elsewhere.

tomspilman:
This is great... a huge missing piece!

MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/IntermediateWriter.cs
tgjones: Good point - I wondered about it too, but forgot to write a test for it. I'll fix that.

Test/ContentPipeline/IntermediateDeserializerTest.cs
tgjones: Okay, fair enough. (It wasn't a random move though; I moved it to its expected position in the file, since this is test case 18.)

tgjones:
`IntermediateSerializer.Serialize` is quite useful for level editor-type applications - I'm using it to save game assets (scene files, materials, etc.) back to disk as XML, ready to be fed in to the content pipeline.

tomspilman:
> `IntermediateSerializer.Serialize`  is quite useful for level editor-type applications 

Interesting..  I've never considered using it before.  But now that we have a cross platform version it is something worth considering.

Test/ContentPipeline/IntermediateDeserializerTest.cs
tomspilman: > I moved it to its expected position in the file, since this is test case 18

Yeah I had seen that before, but every time I noticed I was also modifying the function... so I fought the temptation to move it.

tgjones:
>> IntermediateSerializer.Serialize is quite useful for level editor-type applications

>Interesting.. I've never considered using it before. But now that we have a cross platform version it is something worth considering.

Now that we have a cross platform version of the content pipeline, you mean? True, although in my case, my editor is WPF-based, so pretty much Windows only at this point. I'd love to use Xwt or Eto or something, but none of them yet provide out-of-the-box controls advanced enough for a good editor experience (docking, etc.).

tgjones:
If the tests pass, I think this is ready to go. I'd like to come back to it later and refactor the namespace-aliasing implementation, but I believe it's working correctly now, and that code is all internal and quite isolated.

My latest commit fixes a bug in `IntermediateReader` when external references are referenced from multiple places. I also fixed rather a basic problem with the previous implementation of cycling reference detection.

(Interestingly, it turns out that when XNA deserializes `ExternalReference<T>` objects, it creates a new one every time, even for those that shared an ID in the XML file.)

tgjones:
> I'd like to come back to it later and refactor the namespace-aliasing implementation, but I believe it's working correctly now, and that code is all internal and quite isolated.

I've gone ahead and done this internal refactoring - it was too painful to leave the code as it was :)

tomspilman:
This looks great!  Merging!

PR #3378, opened 2015-01-10
Fix VertexBuffer.SetData bug introduced in #3356

tgjones:
... and add more `SetData` / `GetData` tests to tease out XNA behaviour surrounding `startIndex` / `elementCount` / `vertexStride` parameters.

Fixes the bug introduced in #3356.

Hopefully these tests go some way towards making it clear how these parameters interact. As I understand it, `vertexStride` controls the padding left between source data elements when they are copied to the vertex buffer. It must be `>= sizeof(T)`.

In my opinion, that means it's a somewhat poor choice of name, because if you're passing in a `byte[]`, the value you would choose for `vertexStride` (probably `1`) doesn't really have anything to do with vertex stride. But that's for another time.

I'm sorry for introducing this bug; I'll try to do more testing next time. If you spot anything amiss in this PR, let me know and I'll fix it ASAP.

**IMPORTANT**: @tomspilman , if you want the tests to actually run under XNA, you'll need to change the `RunOneFrame` calls to `Game.Run()`, as per our previous discussion. (I'm not sure why you don't want to use `Game.Run()`, since it has the same effect of running one frame and then exiting. In XNA, `RunOneFrame` seems to assume that the game has already been initialized, which MonoGame does at the start of `RunOneFrame`, but XNA (apparently) doesn't. So it might be non-trivial to get it working in the XNA tests. But that's a separate issue.)

### DIFF
--- a/MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
+++ b/MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
@@ -83,17 +83,72 @@ namespace Microsoft.Xna.Framework.Graphics
             this.GetData<T>(0, data, 0, data.Length, elementSizeInByte);
         }
 
+        /// <summary>
+        /// Sets the vertex buffer data, specifying the index at which to start copying from the source data array,
+        /// the number of elements to copy from the source data array, and the space to leave between elements from the source data array.
+        /// </summary>
+        /// <typeparam name="T">Type of elements in the data array.</typeparam>
+        /// <param name="offsetInBytes">Offset in bytes from the beginning of the vertex buffer to the start of the copied data.</param>
+        /// <param name="data">Data array.</param>
+        /// <param name="startIndex">Index at which to start copying from <paramref name="data"/>.
+        /// Must be within the <paramref name="data"/> array bounds.</param>
+        /// <param name="elementCount">Number of elements to copy from <paramref name="data"/>.
+        /// The combination of <paramref name="startIndex"/> and <paramref name="elementCount"/> 
+        /// must be within the <paramref name="data"/> array bounds.</param>
+        /// <param name="vertexStride">The space to leave between elements copied from <paramref name="data"/>.
+        /// If you specify <c>0</c> for this parameter, it will be treated as if you had specified <c>sizeof(T)</c>.
+        /// With the exception of <c>0</c>, you must specify a value greater than or equal to <c>sizeof(T)</c>.</param>
+        /// <example>
+        /// If <c>T</c> is <c>VertexPositionTexture</c>, but you want to set only the position component of the vertex data,
+        /// you would call this method as follows:
+        /// <code>
+        /// Vector3[] positions = new Vector3[numVertices];
+        /// vertexBuffer.SetData(0, positions, 0, numVertices, vertexBuffer.VertexDeclaration.VertexStride);
+        /// </code>
+        /// 
+        /// Continuing from the previous example, if you want to set only the texture coordinate component of the vertex data,
+        /// you would call this method as follows (note the use of <paramref name="offsetInBytes"/>:
+        /// <code>
+        /// Vector2[] texCoords = new Vector2[numVertices];
+        /// vertexBuffer.SetData(12, texCoords, 0, numVertices, vertexBuffer.VertexDeclaration.VertexStride);
+        /// </code>
+        /// </example>
+        /// <remarks>
+        /// If you provide a <c>byte[]</c> in the <paramref name="data"/> parameter, then you should almost certainly
+        /// set <paramref name="vertexStride"/> to <c>1</c>, to avoid leaving any space between the <c>byte</c> values
+        /// when they are copied into the vertex buffer.
+        /// </remarks>
         public void SetData<T>(int offsetInBytes, T[] data, int startIndex, int elementCount, int vertexStride) where T : struct
         {
             SetDataInternal<T>(offsetInBytes, data, startIndex, elementCount, vertexStride, SetDataOptions.None);
         }
-        		
+
+        /// <summary>
+        /// Sets the vertex buffer data, specifying the index at which to start copying from the source data array,
+        /// and the number of elements to copy from the source data array. This is the same as calling 
+        /// <see cref="SetData{T}(int, T[], int, int, int)"/>  with <c>offsetInBytes</c> equal to <c>0</c>,
+        /// and <c>vertexStride</c> equal to <c>sizeof(T)</c>.
+        /// </summary>
+        /// <typeparam name="T">Type of elements in the data array.</typeparam>
+        /// <param name="data">Data array.</param>
+        /// <param name="startIndex">Index at which to start copying from <paramref name="data"/>.
+        /// Must be within the <paramref name="data"/> array bounds.</param>
+        /// <param name="elementCount">Number of elements to copy from <paramref name="data"/>.
+        /// The combination of <paramref name="startIndex"/> and <paramref name="elementCount"/> 
+        /// must be within the <paramref name="data"/> array bounds.</param>
 		public void SetData<T>(T[] data, int startIndex, int elementCount) where T : struct
         {
             var elementSizeInBytes = Marshal.SizeOf(typeof(T));
             SetDataInternal<T>(0, data, startIndex, elementCount, elementSizeInBytes, SetDataOptions.None);
 		}
 		
+        /// <summary>
+        /// Sets the vertex buffer data. This is the same as calling <see cref="SetData{T}(int, T[], int, int, int)"/> 
+        /// with <c>offsetInBytes</c> and <c>startIndex</c> equal to <c>0</c>, <c>elementCount</c> equal to <c>data.Length</c>, 
+        /// and <c>vertexStride</c> equal to <c>sizeof(T)</c>.
+        /// </summary>
+        /// <typeparam name="T">Type of elements in the data array.</typeparam>
+        /// <param name="data">Data array.</param>
         public void SetData<T>(T[] data) where T : struct
         {
             var elementSizeInBytes = Marshal.SizeOf(typeof(T));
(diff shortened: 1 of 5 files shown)


MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
```diff
@@ -54,13 +54,18 @@ internal protected override void GraphicsDeviceResetting()
 
         public void GetData<T> (int offsetInBytes, T[] data, int startIndex, int elementCount, int vertexStride) where T : struct
```
tomspilman: If we think `vertexStride` is an inaccurate name, we should consider changing it.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
```diff
@@ -54,13 +54,18 @@ internal protected override void GraphicsDeviceResetting()
 
         public void GetData<T> (int offsetInBytes, T[] data, int startIndex, int elementCount, int vertexStride) where T : struct
         {
+            var elementSizeInBytes = Marshal.SizeOf(typeof(T));
+
+            if (vertexStride == 0)
+                vertexStride = elementSizeInBytes;
```
tomspilman: Is that accurate?  In XNA if you pass 0 to `vertexStride` it works anyway?

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tgjones: For now, I could add some XML doc comments to explain it in more detail? Unless you feel strongly about it, I'd prefer to stick with XNA's parameter names until there is a broader decision to move away from the XNA API.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tgjones: Yes - XNA appears to follow this same behaviour of treating `0` as `sizeof(T)`. This case is specifically covered in the unit tests, because it surprised me too.

tomspilman: This is good information for docs on the method.  Stuff that XNA doesn't even mention.

tgjones: Okay - let me add some XML doc comments before you merge this.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tomspilman: > I'd prefer to stick with XNA's parameter names until there is a broader
> decision to move away from the XNA API.

There is a plan in the future to move beyond XNA4, but that isn't my point here.

We do try to fix mistakes in XNA when we can.  Especially mistakes that generally don't hurt compatibility.  In this case the code will work 100% the same...  it will just make more sense.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tgjones: Talking of which, what is MonoGame's policy on doc comments? From what I've seen so far, I guess you don't / can't copy XNA's doc comments?

tomspilman:
My last worry with this PR is that we're breaking non-DirectX platforms.  

I suspect the OpenGL platform expects `vertexStride` to match the VertexDeclaration in most cases... so it could be the logic for doing the copy will be broken.

Unfortunately the OpenGL based unit tests are still not in good shape.  So they don't run right now for us to check.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tomspilman: > Talking of which, what is MonoGame's policy on doc comments? 

The same as code... we cannot use someone else's copyrighted materials.  So you cannot cut and paste over XNA docs.  And really we don't want the XNA docs... a lot of them are quite poor really.  Leaving tons of questions... we should strive for better.

tgjones:
@tomspilman  - that's true. I think the OpenGL platform is the same as DirectX was before my original PR - i.e. it supports a different `vertexStride` in `GetData` ([see here]((https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Vertices/VertexBuffer.OpenGL.cs#L115)), but not `SetData`.

I've said this before, and I was wrong last time, but I don't *think* my changes should affect OpenGL - you weren't previously able to supply a custom `vertexStride`, and you still can't. If you try to, it will probably crash, but that would have been the case before too.

tomspilman:
> you weren't previously able to supply a custom  vertexStride , and you still can't.

I was more concerned about the cases where you can't.  

Like `GetData(T[])` internally passes `sizeof(T)` where as before it passed `VertexDeclaration.VertexStride` down to `PlatformGetData`.  That has a high probability of affecting OpenGL if it uses the `vertexStride` parameter at all.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tgjones: Okay, cool. I assumed so, but it was worth checking.

tgjones:
> Like GetData(T[]) internally passes sizeof(T) where as before it passed VertexDeclaration.VertexStride down to PlatformGetData.

Do you mean `SetData`? I didn't change what `GetData` passes. From what I can see, `VertexBuffer.SetBufferData` in the OpenGL platform - [here](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Vertices/VertexBuffer.OpenGL.cs#L144) - doesn't use `vertexStride` at all.

Of course it would better to actually test this :) Is there an easy way of doing that?

tomspilman:
> Do you mean `SetData`?

Yeah...  that is what I meant.  

The ones where you have replaced `VertexDeclaration.VertexStride` with ` sizeof(T)`.

> From what I can see, `VertexBuffer.SetBufferData` in the OpenGL platform - here - doesn't use `vertexStride` at all.

Good...  at least that eliminates that concern for now.

It does however look like it might break PSM...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Vertices/VertexBuffer.PSM.cs#L35


> Of course it would better to actually test this :) Is there an easy way of doing that?

Easy...  possibly not.  The easiest way would be to generate the Windows OpenGL unit test project in Protobuild, ignore any other broken test cases, focus only on yours.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tgjones: I've added comments. Might have got a bit carried away, but hopefully verbosity is better than "paramX specifies X".

tomspilman: Naw...  they are ok.  My only gripe would be that the summary blocks got a little bit long.  They are supposed to be short and to the point.  Remarks are for the deeper details of the method.

tgjones:
I've run the OpenGL tests, both on the current version, and also on the commit prior to any of my changes (801d1ff84c6b20baf635198b913466a9c59ce5fb).

The tests that were failing previously are still failing, and the tests that were passing previously are still passing. So that's good - I don't think I've broken anything that was previously working.

Some of the new tests, unrelated to `vertexStride`, actually fail on the old version, so that's helpful to uncover a bug or two there. And of course the `SetData(..., vertexStride)`-related tests fail on both old and new versions. (The failing tests should be looked at - I've done some experimental fixes, and it doesn't look hard to get most of the tests passing. But I don't want to do that until this current PR is considered stable.)

tomspilman:
>  I don't think I've broken anything that was previously working.

Great... for this PR that is all you need to worry about.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tgjones: I couldn't think of shorter ways of saying what I wanted to say, without resorting to "specifies the {parameterName}" type comments, which don't help anybody. And when you seem them rendered in a tooltip, without all the markup, I don't think they're too bad. I'm open to suggestions though!

tgjones:
... except for PSM, as you pointed out - looks like that will indeed be affected. Do you want me to look at that? I don't know anything about PSM - I don't have the SDK installed.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
tomspilman: I agree... good docs are really hard to write.

tomspilman:
>  I don't know anything about PSM - I don't have the SDK installed.

I don't either.  Usually we sort of blindly fix it and let the build server tell us if it compiles.  We really don't have very many people that use it.

I think if you just replace `vertexStride` with the stride from `VertexDeclaration` it should behave the same as before for most cases.

Shqrdx:
@tgjones Eh.. I guess my PR https://github.com/mono/MonoGame/pull/3376 is not as much as yours
Could you put   
    switch (IndexElementSize)
  {
                case IndexElementSize.SixteenBits:
                    if (typeof(T) != typeof(short) && typeof(T) != typeof(ushort))
                   {
                        throw new ArrayTypeMismatchException("Invalid type using when short or ushort need to be used.");
                    }
                   break;
                case IndexElementSize.ThirtyTwoBits:
                    if (typeof(T) != typeof(int) && typeof(T) != typeof(uint))
                    {
                        throw new ArrayTypeMismatchException("Invalid type using when int or uint need to be used.");
                    }
                    break;
           } 
to IndexBuffer.SetDataInternal. Thats will be very helpful. Thanks.

tomspilman:
@Shqrdx - It would be good to add some unit tests to yours following @tgjones example here.

tomspilman:
Ok... gonna merge this.  Should fix #3356 and it has better functionality too.  Thanks @tgjones !

tgjones:
Great! Thanks for merging it.

By the way, do you have any plans to run tests for more platforms (including XNA) on TeamCity, or are there technical reasons for not doing so? Does it only run Direct3D platform tests so far?

tomspilman:
> By the way, do you have any plans to run tests for more platforms (including XNA) on TeamCity

Yes.  We don't because they don't yet successfully run with the existing tests.

The sooner we can run the Windows OpenGL tests the better.  Then after we can get the Mac platform working followed by Linux ideally.

iOS / Windows Phone are bigger challenges as we need to setup the correct hardware and maybe even hack it to make it work in an automated test environment.

PR #3379, opened 2015-01-10
Buffer fixes

Shqrdx:
Reopened https://github.com/mono/MonoGame/pull/3376

### DIFF
--- a/MonoGame.Framework/Graphics/Vertices/IndexBuffer.cs
+++ b/MonoGame.Framework/Graphics/Vertices/IndexBuffer.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Xna.Framework.Graphics
 {
     public partial class IndexBuffer : GraphicsResource
     {
-        private bool _isDynamic;
+        private readonly bool _isDynamic;
 
         public BufferUsage BufferUsage { get; private set; }
         public int IndexCount { get; private set; }
@@ -24,7 +24,7 @@ namespace Microsoft.Xna.Framework.Graphics
         {
 			if (graphicsDevice == null)
             {
-                throw new ArgumentNullException("GraphicsDevice is null");
+                throw new ArgumentNullException("graphicsDevice");
             }
 			this.GraphicsDevice = graphicsDevice;
 			this.IndexElementSize = indexElementSize;	
@@ -63,7 +63,7 @@ namespace Microsoft.Xna.Framework.Graphics
                         throw new NotSupportedException("The profile does not support an elementSize of IndexElementSize.ThirtyTwoBits; use IndexElementSize.SixteenBits or a type that has a size of two bytes.");
                     return IndexElementSize.ThirtyTwoBits;
                 default:
-                    throw new ArgumentOutOfRangeException("Index buffers can only be created for types that are sixteen or thirty two bits in length");
+                    throw new ArgumentOutOfRangeException("type","Index buffers can only be created for types that are sixteen or thirty two bits in length");
             }
         }
 
@@ -78,7 +78,7 @@ namespace Microsoft.Xna.Framework.Graphics
         public void GetData<T>(int offsetInBytes, T[] data, int startIndex, int elementCount) where T : struct
         {
             if (data == null)
-                throw new ArgumentNullException("data is null");
+                throw new ArgumentNullException("data");
             if (data.Length < (startIndex + elementCount))
                 throw new InvalidOperationException("The array specified in the data parameter is not the correct size for the amount of data requested.");
             if (BufferUsage == BufferUsage.WriteOnly)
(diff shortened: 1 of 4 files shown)


MonoGame.Framework/Graphics/Vertices/IndexBuffer.cs
```diff
@@ -115,10 +115,26 @@ internal protected override void GraphicsDeviceResetting()
...
+                        throw new ArrayTypeMismatchException("Invalid type using when short or ushort need to be used.");
+                    }
+                    break;
+                case IndexElementSize.ThirtyTwoBits:
+                    if (type != typeof(int) || type != typeof(uint))
+                    {
+                        throw new ArrayTypeMismatchException("Invalid type using when int or uint need to be used.");
+                    }
+                    break;
+            }
```
tomspilman: Are we sure `IndexBuffer` only allows `T` to be 16bits or 32bits?

I would feel better about this if we added a unit test and ran it on XNA to verify.

Shqrdx: Its my fault - || instead &&

Shqrdx: I doubt if someone put exotic things like Color or float to IndexBuffer. It can easy call blue screen on Windows.

tomspilman: So the failure above shows at least shows something is wrong there.

We need to add a test where the `short[]` array is copied into a `byte[]` for instance and then `SetData`/`GetData` is called.  I suspect `IndexBuffer` doesn't care what `T` is as long as it can be copied to the buffer properly.

Shqrdx: Are you sure thats good idea to put byte array to thing intended to accept short,ushort,int,uint... ? I dont know.

Shqrdx: however In XNA its accepted...

Shqrdx: Maybe I add byte and ubyte to exception list ?

tomspilman: > I doubt if someone put exotic things like Color or float to IndexBuffer.

It isn't about doubt or what is logical...  it is what XNA supports.

I just tested this...

```
            var savedData = new short[] { 1, 2, 3, 4 };
            var asBytes = new byte[savedData.Length * sizeof(short)];
            Buffer.BlockCopy(savedData, 0, asBytes, 0, asBytes.Length);
            var indexBuffer = new IndexBuffer(GraphicsDevice, IndexElementSize.SixteenBits, savedData.Length, BufferUsage.None);
            indexBuffer.SetData(asBytes);
            var readData = new short[4];
            indexBuffer.GetData(readData, 0, 4);
            Assert.AreEqual(savedData, readData);
```

That works on XNA.  It would fail under your change.

tomspilman: > Maybe I add byte and ubyte to exception list ? 

No...  that isn't the fix.

The type of `T` does not matter.  XNA is simply copying an array of bytes as far as it cares.

Shqrdx: Ok, I delete this change.

tomspilman: Just to be complete... here is the same test with an array of float passed to `SetData`.

```
            var savedData = new short[] { 1, 2, 3, 4 };
            var asFloat = new float[(savedData.Length * sizeof(short)) / sizeof(float)];
            Buffer.BlockCopy(savedData, 0, asFloat, 0, savedData.Length * sizeof(short));
            var indexBuffer = new IndexBuffer(GraphicsDevice, IndexElementSize.SixteenBits, savedData.Length, BufferUsage.None);
            indexBuffer.SetData(asFloat);
            var readData = new short[4];
            indexBuffer.GetData(readData, 0, 4);
            Assert.AreEqual(savedData, readData);
```
There is nothing magical about float that would make this fail or any value of `T`.  XNA and MonoGame are simply copying one block of memory to another.

Shqrdx: Just notice that incorrect putting calling blue screen very quickly (on WinDX platform). However its more like SharpDX problem than MonoGame.

tomspilman: > Just notice that incorrect putting calling blue screen very quickly

I think you mean a crash when calling `SetData`.  Yes... that is what https://github.com/mono/MonoGame/pull/3378 fixed.

Shqrdx: Ah.. than my apologize. Tested it now its fine now...

Shqrdx:
@tomspilman Ok, now everything ok ?

Issue #3322, opened 2014-12-29
Pipeline tool - add indicator for missing files.


dneelyep:
It would be useful to add an indicator that a file does not exist in the project structure pane when the Pipeline tool can't find a file

For example:
If I have this project structure:
Models\
    ModelA
    ModelB

and I then move all the files into a new folder like so:
Models2\
    ModelA
    ModelB

it would be useful if a little red circle showed in the Pipeline UI, indicating that the tool can't find the file. Not sure how tricky this would be to implement, but it would just be a nice-to-have thing - not super high priority I wouldn't think.

cra0zy:
This can be closed.

PR #3384, opened 2015-01-12
Pipeline indicator for missing files

cra0zy:
I liked the idea from #3322 so I added it.


cra0zy:
@tomspilman @dellis1972 That failed test makes no sense, did I do something wrong?

tgjones:
@cra0zy No, it's not you - that test has been failing intermittently ever since I added it.

@tomspilman I'll take a look at it now. I don't think I materially changed `GetData` (only `SetData`) so I suspect this new test probably exposes some underlying issue that has been present for a while. I'll know more when I find out what the actual problem is.

cra0zy:
@tgjones Thanks for the info!!!

tgjones:
I think I've fixed this intermittent test failure: #3402. (It turned out to be a long-standing issue in a little-if-at-all-used code path in `GetData`, that happened to be uncovered by the tests I wrote for my `SetData` changes.)

KonajuGames:
@cra0zy If you merge latest from develop into this branch, it will re-run the test.

cra0zy:
It would be nice if someone could merge this. After this PR I am changing to using other branches for PR instead of develop branch so I can setup multiple PRs at the same time.

KonajuGames:
I'll merge this now.  We can fix the path separators in a later PR (there are constants in .NET for path separators on a platform).

cra0zy:
Thanks @KonajuGames !!!

PR #3385, opened 2015-01-12
Texture parameters on effects can now be explicitly set to null

### LINKED ISSUE #3351: Textures should always be assigned to device, even if they are null
I'm not sure if I'm right about this, which is why I'm creating an issue, not a pull request. If it is indeed a bug, then I can create a pull request.

It seems to me that this line is wrong:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/EffectPass.cs#L118

I think we should remove the null check, and always set the texture, even if it is null.

For example, if the previously bound texture is a TextureCube (well, Texture2D array, in Direct3D 11 terminology), and the new texture is a Texture2D, but it just happens to be null, then it's still necessary to set that texture slot to null, otherwise Direct3D will complain that the currently bound texture has the wrong dimension for the current shader. (I have this exact situation, which is how I found this.)

tgjones:
Fixes #3351.

Please feel free to review this thoroughly before merging :) I haven't changed many lines of code in `EffectPass`, but it's quite a fundamental change. I hope I've understood the `_stateKey` system correctly - I based it on how `ConstantBuffer.Update` works.

### DIFF
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -125,6 +125,7 @@
     
     <Compile Include="Framework\Visual\BlendStateTest.cs" />
     <Compile Include="Framework\Visual\DepthStencilStateTest.cs" />
+    <Compile Include="Framework\Visual\EffectTest.cs" />
     <Compile Include="Framework\Visual\RasterizerStateTest.cs" />
     <Compile Include="Framework\Visual\SamplerStateTest.cs" />
     <Compile Include="Framework\Visual\ScissorRectangleTest.cs" />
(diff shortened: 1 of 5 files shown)


KonajuGames:
It looks reasonable to me.  I don't fully understand the stateKey system either though.

tgjones:
@tomspilman @KonajuGames Let me know if there's something more you'd like to see for this PR - any specific additional tests you can think of?

tgjones:
I guess this isn't merged yet because it's a simple-but-fundamental change that requires more testing? If anyone can think of more test scenarios, I'd be happy to add more unit tests.

tomspilman:
The issue for me is that I don't see how the state key technique reliably solves this problem.

Just add a comment in the code where the state key is used as what that code is doing and how it solves the issue.

tgjones:
I've added some comments - hopefully they make sense.

tomspilman:
Ok...  I now remember how that works.  Thanks!

PR #3388, opened 2015-01-13
Linux installer is now checking installation priviledge for sudo or root

christallire:
little addition to nice bash script.. :)


tomspilman:
@cra0zy - Thoughts?

cra0zy:
It's good.

tomspilman:
Thanks @cra0zy  and @christallire !

PR #3391, opened 2015-01-14
Xam.iOS Unified: small fixes so that MG builds

rtabbara:
* Texture2D.OpenGL : Making sure we don't cast vars to ints prematurely. CGBitmapContext constructor is expecting nints and although I think this conversion should be occurring implicitly, it appears that one runs into issues if we cast and then pass width/height vars as ints.
* KeyboardInputView: UILabel.StringSize was deprecated and is removed in unified API, so need an alternate approach
* SignedInGamer: Conversion from DateTime to NSDate needs to be explicit. A suggested solution in http://developer.xamarin.com/guides/cross-platform/macios/unified/ is to add extension methods which have been included, but people may prefer for them to be located somewhere else.


dellis1972:
Looks ok to me. 
Anyone else familiar with iOS want to comment?

dellis1972:
@tomspilman can we test the build on this one again. I merged the Addin fixes so with these fixes we should be gold.

danzel:
I think you need to rebase on master or merge from master. The build bot builds the remote branch, not the merged result AFAIK

tomspilman:
> The build bot builds the remote branch, not the merged result 

Yes... this is correct.  Waiting on @rtabbara  or @kjpou1 to merge/rebase from develop to get @dellis1972 fixes.

tomspilman:
I'm going to kick off the build again... that last failure wasn't your fault.

rtabbara:
Done. Thanks @tomspilman and @dellis1972 .

tomspilman:
Ok...  all good to merge!

PR #3395, opened 2015-01-14
Curve improvements

Shqrdx:
Changes : 
  -Comments for Curve,CurveKeyCollection,CurveKey.
  -Added missed [TypeConverter(typeof(ExpandableObjectConverter))] to these classes.
  -Added possibility of serialization to these classes.
  -Small bug fixes.**
    *Renamed private fields to version with prefix "_".
    *Changed loop i++ (which is less native to computers than ++i).
  -New tests for these classes.
  -cs files now sorted in MonoGame.Tests.definition

### DIFF
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -84,23 +84,26 @@
         <Platforms>iOS</Platforms>
     </Compile>
 
+    <Compile Include="Framework\Bounding.cs" />
     <Compile Include="Framework\ColorTest.cs" />
+	<Compile Include="Framework\CurveKeyCollectionTest.cs" />
+	<Compile Include="Framework\CurveKeyTest.cs" />
+	<Compile Include="Framework\CurveTest.cs" />
+    <Compile Include="Framework\GamePadStateTest.cs" />
+    <Compile Include="Framework\GameTest.cs" />	
+    <Compile Include="Framework\GameTest+Methods.cs" />
+    <Compile Include="Framework\GameTest+Properties.cs" />
     <Compile Include="Framework\GestureRecognizerTest.cs" />
+    <Compile Include="Framework\MathHelperTest.cs" />
     <Compile Include="Framework\MockWindow.cs" />
     <Compile Include="Framework\PackedVectorTest.cs" />
-    <Compile Include="Framework\TouchCollectionTest.cs" />
-    <Compile Include="Framework\TouchPanelTest.cs" />
-    <Compile Include="Framework\TouchLocationTest.cs" />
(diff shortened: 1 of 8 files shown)


tomspilman:
This failure isn't your fault...  @dellis1972 needs to fix the build after the XamarinStudio update broke the old MonoDevelop plugin stuff.  Things will be back to working later today.

Shqrdx:
@tomspilman Oh, glad to hear this !

cra0zy:
@Shqrdx If you have some time, would you mind giving me a hand in adding win forms with #3384?

Shqrdx:
@cra0zy You mean indicator to missed files ? Of course, but first show me the screenshot of your work on linux. If you have skype you can contact with me(xaron-7).

cra0zy:
@dellis1972 So with monodevelop 5.7 / xamarin studio 5.7 all of my custom addins are broken. Since your monodevelop server addin is also out If you figure out what's causing it, please do tell me.

polsys:
In the tests you seem to test for floating-point number equality with `Assert.AreEqual`... and this is almost always a big no-no. This also affects `Vector2` tests, they failed under XNA thanks to rounding differences. Some of the asserts in this PR really seem to work around rounding errors, and could fail on another machine, build configuration or code change:
```C#
// Should be 0.800f with reasonable rounding margin
Assert.AreEqual(0.799999952f, curve3.Keys[0].TangentOut);
```
I think something like https://github.com/mono/MonoGame/blob/develop/Test/Runner/Utility.cs#L50-73 should be implemented for curves and vectors.

As a side note, `i++` and `++i` should produce the same result when used in a loop, so there should be no technical reason for changing them. Readability is of course completely different beast.

dellis1972:
@cra0zy on windows it seems that all the core assemblies for XS has been updated to .net 4.5 as a result a 4.0 addin will not be able to load the XS references (e.g MonoDevelop.Core). Not sure if this is the same on Mac or Linux though.

cra0zy:
@dellis1972 Thanks, now it works

Shqrdx:
@polsys 1. Yes, but I tested it - its fine - if you for example put double or forgot add ".f" to it it does fail.
2. Of course it is just aesthetical reason - I readed about it few weeks ago - in c++ its common spelling, computer assembly generation for ++i is more faster than i++. So why not ? It is goal to identical targets.

Shqrdx:
Ok. I implement something like https://github.com/mono/MonoGame/blob/develop/Test/Runner/Utility.cs#L50-73 for vectors and curves if we (really) need it.

tomspilman:
@Shqrdx - Merge or rebase the latest develop changes into this branch.  That will fix the build error.

PR #3402, opened 2015-01-15
Fix intermittent VertexBuffer.GetData bug when vertexStride != sizeof(T)

tgjones:
and remove unused DataPointer variable.

This bug has been present since at least February 2014, but I suspect it wasn't noticed because not many people, if any, used the overload of `GetData` that specifies a `vertexStride`.

The tests I added recently exercised this code path, and caused the loop to overrun the buffer, but it was intermittent, depending on what happened to be in the memory following the buffer.


KonajuGames:
Thanks @tgjones.  Merging.

PR #3408, opened 2015-01-16
Fixed Windows Installer Version Number

tomspilman:
We now pull the Windows installer version number from the TeamCity environment variable.  This ensures we can update the build version from TeamCity and have the build reflect that change.


Shqrdx:
I found bug inside installer - it suggest install MonoGame to all Visual studio versions , however I have only VS2013 Community edition...

tomspilman:
> it suggest install MonoGame to all Visual studio versions 

 Check your documents folder.  Do you have a "Visual Studio 2010" and "Visual Studio 2012" folder in there?  That is what it is looking for to detect if it should install templates for them.

> however I have only VS2013 Community edition

I haven't tested VS2013 community edition at all, so I don't know if we properly install templates for it or not.

Shqrdx:
> Do you have a "Visual Studio 2010" and "Visual Studio 2012" folder in there? 
Nope, only Visual Studio 2013

tomspilman:
> Nope, only Visual Studio 2013

Weird...  i'll investigate some.

Shqrdx:
image

tomspilman:
Thanks @Shqrdx ... that helps!

PR #3410, opened 2015-01-16
Metro to Modern transformation

Shqrdx:
All classes begins with Metro should be renamed to Modern(without UI) because Microsoft is changed it long time ago - http://www.theverge.com/2012/8/10/3232921/microsoft-modern-ui-style-metro-style-replacement
Also I remove not used references and fix licenses.

### DIFF
--- a/MonoGame.Framework/Windows8/ModernFrameworkView.cs
+++ b/MonoGame.Framework/Windows8/ModernFrameworkView.cs
@@ -1,4 +1,4 @@
-﻿// MIT License - Copyright (C) The Mono.Xna Team
+﻿// MonoGame - Copyright (C) The MonoGame Team
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
--- a/MonoGame.Framework/Windows8/ModernFrameworkView.cs
+++ b/MonoGame.Framework/Windows8/ModernFrameworkView.cs
@@ -1,18 +1,18 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿// MIT License - Copyright (C) The Mono.Xna Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using Windows.ApplicationModel.Activation;
 using Windows.ApplicationModel.Core;
 using Windows.UI.Core;
-using Windows.ApplicationModel.Activation;
 
 namespace Microsoft.Xna.Framework
 {
-    class MetroFrameworkView<T> : IFrameworkView
+    class ModernFrameworkView<T> : IFrameworkView
         where T : Game, new()
     {
-        public MetroFrameworkView(Action<T, IActivatedEventArgs> gameConstructorCustomizationDelegate)
+        public ModernFrameworkView(Action<T, IActivatedEventArgs> gameConstructorCustomizationDelegate)
         {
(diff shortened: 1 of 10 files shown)


Shqrdx:
@tomspilman I think its good internal change.. what do you think ?

MonoGame.Framework/Windows8/ModernGameWindow.cs
```diff
@@ -1,62 +1,19 @@
...
-code form, you may only do so under a license that complies with this license.
-(E) The software is licensed "as-is." You bear the risk of using it. The contributors give no express warranties, guarantees
-or conditions. You may have additional consumer rights under your local laws which this license cannot change. To the extent
-permitted under your local laws, the contributors exclude the implied warranties of merchantability, fitness for a particular
-purpose and non-infringement.
-*/
-#endregion License
+﻿// MIT License - Copyright (C) The Mono.Xna Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
```
tomspilman: Note... XnaTouch is the original name of MonoGame.  So this should have a MonoGame license header in it.

MonoGame.Framework/Windows8/ModernHelper.cs
```diff
@@ -1,14 +1,15 @@
-﻿using System;
-using System.Collections.Generic;
+﻿// MIT License - Copyright (C) The Mono.Xna Team
```
tomspilman: This for sure should have a MonoGame license header...  I wrote this code myself for MonoGame.

MonoGame.Framework/Windows8/ModernGamePlatform.cs
```diff
@@ -1,90 +1,20 @@
-#region License
-/*
-Microsoft Public License (Ms-PL)
-MonoGame - Copyright © 2009-2011 The MonoGame Team
+// MIT License - Copyright (C) The Mono.Xna Team
```
tomspilman: This should use the MonoGame license header.

MonoGame.Framework/Windows8/ModernFrameworkView.cs
```diff
@@ -1,18 +1,18 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+﻿// MIT License - Copyright (C) The Mono.Xna Team
```
tomspilman: This should use the MonoGame license header.... I wrote this code for MonoGame.

tomspilman:
What do you think @KonajuGames ?

IMO this change is unnecessary really.  Metro/Modern... who cares.

The bigger concern is Windows 10 is coming... will it be renamed again?  I don't know.  So why make this change now?

Shqrdx:
Damn, i didnt see this...

Shqrdx:
@tomspilman Yes, its unnecessary. But small things together can create a beautiful picture.

tomspilman:
> But small things together can create a beautiful picture. 

How beautiful the code looks like is irrelevant...  this change is more likely to cause conflicts which is more important.

KonajuGames:
I think the name change is not really necessary since it is internal. It
has had that name for the several years since the change, and we discussed
changing it back then and decided to leave it as was. Even Microsoft still
have things that refer to it as Metro. It would also cause issues for
anyone with local changes to any of those files. And no-one really liked
the Modern name anyway.  Microsoft don't even use the name much these days,
preferring the term Windows Store App, but everyone still knows what Metro
means. It's also a nice piece of history, saying we supported apps for
Windows 8 back at the beginning.

PR #3411, opened 2015-01-16
Fix bug in IntermediateSerializer.Serialize when value of statically-known type is null

tgjones:
This was a minor bug in the `IntermediateSerializer.Serialize` implementation that was merged a few days ago.

I've fixed it and added a test case for it.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ContentTypeSerializerT.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ContentTypeSerializerT.cs
@@ -43,8 +43,9 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
 
         protected internal override void ScanChildren(IntermediateSerializer serializer, ChildCallback callback, object value)
         {
-            var cast = value == null ? default(T) : (T)value;
-            ScanChildren(serializer, callback, cast);
+            if (value == null)
+                return;
+            ScanChildren(serializer, callback, (T)value);
         }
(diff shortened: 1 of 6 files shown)


tgjones:
Interesting... have I just accidentally uncovered a bug? After deserialisation, the new property I added, `H`, is correctly null. But after sending it through the the content pipeline, it becomes an empty list - which is what causes the test failure.

That test passes on XNA. It seems like [this line in ReflectiveReader](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentReaders/ReflectiveReader.cs#L161) always creates an object, even if it should be null, but I'm not at all familiar with that code so there might be some subtleties that I'm missing.

tomspilman:
> It seems like this line in ReflectiveReader always creates an object, even if it should be null

Interesting.

Well in your test case `H` is a field and not a property.  You can see here...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentReaders/ReflectiveReader.cs#L152

Seems like maybe instead of calling `elementType.GetDefaultConstructor();` we should just be getting the value from the field like we do for properties.  So something like this...

```
            Func<object, object> construct = parent => null;
            if (property != null && !property.CanWrite)
                construct = parent => property.GetValue(parent, null);
            else
                construct = parent => field.GetValue(parent, null);
```

tgjones:
I've simplified it as you suggested (the variable declaration already sets the default, so it's even simpler).

tomspilman:
Looks good...  unit tests FTW!  Merging!

PR #3412, opened 2015-01-16
Allow line break as element separator during deserialization

### LINKED ISSUE #3404: Importing XML through the MGCB processor - errors
From one fire to the next.  I'm converting an old XNA 4 project to MonoGame and running in to issues with the XML importer.  

SO far I have two issues (which were not an issue in the original XNA XML content importer.

1: "Not have enough entries in space-separated list!"
this seems related to how certain data types are imported, e.g. A vector 2 is expected to be 2 integers space separated.
The source XML hasn't changed but with the MGCB processor it is  falling over.

2: The Xml element `<item>` is required!
Although this imported fine in XNA, which ignored missing values.  With the MGCB importer, it seems to be forcing the XML de-serialisation to make all fields mandatory

Any ideas on these.  I would rather not have to go and edit the hundreds of XML configuration files used in this project, especially since they worked fine previously.

tgjones:
Fixes #3404.

I don't know enough about this code to know whether this is a safe change - @tomspilman, does this look okay to you? I'm not sure what other types of data we might be deserializing, and whether a line break is always an acceptable separator.


Build/Projects/MonoGame.Tests.definition
```diff
@@ -509,6 +509,9 @@
     <Content Include="Assets\Xml\12_MathTypes.xml">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+	<Content Include="Assets\Xml\12_MathTypesOutput.xml">
```
tomspilman: You know I think a better idea here is to rename this like `21_CustomFormatting.xml`.  And have a test focus on custom/alternate formatting during deserialization for different types.

MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ElementSerializerT.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
 {
     abstract class ElementSerializer<T> : ContentTypeSerializer<T>
     {
-        private static readonly char [] _seperators = { ' ', '\t' };
+        private static readonly char [] _seperators = { ' ', '\t', '\n' };
```
tomspilman: I think this is safe.  If all our tests pass with it then at least for now it should be fine.

Build/Projects/MonoGame.Tests.definition
tgjones: I like that - I've just pushed another commit for this.

tgjones:
I've also handled XML comments inside elements - I'm not sure if this is particularly standard according to the XML spec, but the XNA deserializer handles it, so we should too.

tomspilman:
Cool... this looks like a great fix and improvement to the tests.

PR #3418, opened 2015-01-17
iOS 8 rotation handling for iPad and when Info.plist is > SupportedOrientations

rtabbara:
Having another crack of this as I realised I previously didn't correctly read the original issue encountered in #3271 by @Nezz. So there were actually two separate issues:

1. iOS 8 rotation problems when targeting iPad when you allow *both* landscape/portrait rotations. This was something other people in the discussion of #3271 were encountering and that was what my previous fix was targeting.

2. The original issue @Nezz was describing was the case where you're Info.plist is a superset of those specified in the root view controller's supported orientations. So e.g. Info.plist is portrait/landscape but you want the root view controller to only support landscape orientations. In this instance, the behaviour should be that the app launches in landscape.

So I've appended my original fix for 1. along with another fix for 2. I've tried a number of iPad/iPhone simulated devices targeting iOS 7.1/8.1 so I'm hopeful that this solves the issues.

### DIFF
--- a/MonoGame.Framework/iOS/iOSGamePlatform.cs
+++ b/MonoGame.Framework/iOS/iOSGamePlatform.cs
@@ -117,7 +117,6 @@ namespace Microsoft.Xna.Framework
             game.Services.AddService (typeof(UIViewController), _viewController);
             Window = new iOSGameWindow (_viewController);
 
-            _mainWindow.RootViewController = _viewController;
             _mainWindow.Add (_viewController.View);
 
             _viewController.InterfaceOrientationChanged += ViewController_InterfaceOrientationChanged;
@@ -201,6 +200,11 @@ namespace Microsoft.Xna.Framework
         {
             // Show the window
             _mainWindow.MakeKeyAndVisible();
+
+            // In iOS 8+ we need to set the root view controller *after* Window MakeKey
+            // This ensures that the viewController's supported interface orientations
+            // will be respected at launch
+            _mainWindow.RootViewController = _viewController;
 
             BeginObservingUIApplication();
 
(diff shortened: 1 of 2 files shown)


Nezz:
I will have a look at it soon.

Nezz:
The game still launches in portrait at first for me.

iPhone 4S running iOS 7.1
```
2015-01-18 13:55:22.688 TestiOS[1114:60b] this.GraphicsDevice.Viewport={X:0 Y:0 Width:640 Height:960 MinDepth:0 MaxDepth:1}
2015-01-18 13:55:23.218 TestiOS[1114:60b] this.GraphicsDevice.Viewport={X:0 Y:0 Width:960 Height:640 MinDepth:0 MaxDepth:1}
```

iPad 2 running iOS 8.1
```
2015-01-18 14:01:44.775 SongArciOS[1852:1508517] this.GraphicsDevice.Viewport={X:0 Y:0 Width:768 Height:1024 MinDepth:0 MaxDepth:1}
2015-01-18 14:01:45.730 SongArciOS[1852:1508517] this.GraphicsDevice.Viewport={X:0 Y:0 Width:1024 Height:768 MinDepth:0 MaxDepth:1}
```

Nezz:
It's still way better than as it is now because currently the screen doesn't rotate at all. I'm fine with merging this if @rtabbara gives it a go.

rtabbara:
@Nezz Yep, can replicate this behaviour, but ultimately the app will not launch in portrait which is a good thing at least. It just means that the viewport is going to be set twice at initialisation. 

Honestly, I think this is the best we can do because if you see how the graphics device is first initialised [here](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/GraphicsDeviceManager.cs#L268-L276), it's making use of the window's current orientation to determine the buffer width/height. At that point in time there's no root view controller attached to the window, so we can't make any assumptions as to what the desired supported orientations are going to be.

Nezz:
Okay, @tomspilman can you merge this?

tomspilman:
Thanks @Nezz ... merging!

PR #3420, opened 2015-01-17
Fixed SpriteSortMode.Texture

### LINKED ISSUE #3272: SpriteSortMode.Texture doesn't sort.
In SpriteBatcher.DrawBatch, _batchItemList.Sort(CompareTexture) is used to sort the list on Texture. CompareTexture however is defined as such:

static int CompareTexture(SpriteBatchItem a, SpriteBatchItem b)
{
    return ReferenceEquals(a.Texture, b.Texture) ? 0 : 1;
}

As far as I know the Sort(System.Comparison) works the same as the IComparable based sort. In the doc regarding IComparable.CompareTo it states: "If A.CompareTo(B) returns a value other than zero, then B.CompareTo(A) must return a value of the opposite sign." Which is probably the reason why SpriteSortMode.Texture doesn't actually sort the list (and therefore will cause excessive draw calls).

http://msdn.microsoft.com/en-us/library/system.icomparable.compareto.aspx

polsys:
Fixes #3272.

* Added SpriteSortMode test cases to SpriteBatch tests.
* Introduced an internal property to `Texture2D` to use as the comparison key. Because this key is determined by the load order, the tests should stay deterministic, and getting it is an O(1) operation.


polsys:
Is it just me, or do the XNA tests have trouble with capturing visual test frames? I had to test this against a separate XNA project.

When poking around I noticed that on XNA, `VisualTestGame.Update()` (in `TestGameBase`) got called but `Draw()` was never called. On MonoGame, `Draw()` was called too.

tomspilman:
> Is it just me, or do the XNA tests have trouble with capturing visual test frames

It does...  it often works, but every once in a while it fails.

I've not been able to fully determine the reason for this.  

One issue is that NUnit runs all tests within the same process/thread.  So you are creating a Game instance, destroying it, then recreating it in the next test.  XNA has issues with that... namely it leaves the WM_QUIT message on the message pump causing the next Game instance to immediately shutdown again.  This is what this code is trying to solve...
 
 https://github.com/mono/MonoGame/blob/develop/Test/Framework/TestGameBase.cs#L326

And that seems to work.  Still we have cases where the XNA Game class either instantly quits or just doesn't render any frames.  I can't tell if it has anything to do with the message pump issue or is some other problem entirely.

This by the way is mostly why we're not yet running the XNA tests on the TeamCity server.

tgjones:
(@tomspilman another, separate, issue being that RunOneFrame doesn't work in the XNA tests? By the way, do you want me to raise an issue for that?)

polsys:
I tested a bit more... It appears that on my computer the visual tests stop failing when `GameTest` is excluded from the project. 

`MockGame` does not swallow `WM_QUIT` messages. Adding a `AbsorbQuitMessage` call, copypasting the code from `TestGameBase`, to `MockGame.EndRun()` made the tests work for me.

tomspilman:
> On my computer the visual tests stopped failing when `GameTest` was excluded from the project. 

@polsys - How did you exclude it?  It seems like excluding it would remove some of the tests it depends on.


>  By the way, do you want me to raise an issue for that?)

@tgjones - Sure... lets do that.

polsys:
> How did you exclude it?

I mean that I excluded it from the Visual Studio project while trying to narrow down the issue. All the remaining tests started passing when `GameTest` files weren't included.

Because `GameTest` uses `MockGame` and the visual tests use `VisualTestGame`, I would suspect that `MockGame` is at fault.

polsys:
XNA tests aside, do the changes look OK?

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -146,6 +149,18 @@ private static void ConvertToABGR(int pixelHeight, int pixelWidth, int[] pixels)
...
+        /// Returns a code unique to this <see cref="Texture2D"/> instance, used for sorting
+        /// textures with <see cref="SpriteSortMode.Texture"/>.
+        /// </summary>
+        internal int ComparisonCode
+        {
+            get { return _comparisonCode; }
+        }
+
+        private int _comparisonCode;
+        private static int _nextComparisonCode;
```
tomspilman: This should be moved to `Texture` so it is available to all derived texture classes.

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -52,6 +52,9 @@ protected Texture2D(GraphicsDevice graphicsDevice, int width, int height, bool m
             this.height = height;
             this._format = format;
             this._levelCount = mipmap ? CalculateMipLevels(width, height) : 1;
+
+            this._comparisonCode = _nextComparisonCode;
+            _nextComparisonCode++;
```
tomspilman: We should be more careful about overflow here.  Maybe do this instead...
```
this._comparisonCode = unchecked(_nextComparisonCode++);
```

tomspilman: Actually this would be better...

```
this._comparisonCode = Interlocked.Increment(ref _nextComparisonCode);
```

This not only handles overflow correctly, but it also makes this thread safe which is important.
```

File: MonoGame.Framework/Graphics/SpriteBatcher.cs [hunk 3]
```diff
@@ -163,14 +163,14 @@ private void EnsureArrayCapacity(int numBatchItems)
...
+        /// Comparison of the underlying Texture objects for each given SpriteBatchitem.
         /// </summary>
         /// <param name="a"></param>
         /// <param name="b"></param>
-        /// <returns>0 if they are not reference equal, and 1 if so.</returns>
+        /// <returns>0 if they are equal, -1 or 1 if not.</returns>
 	    static int CompareTexture ( SpriteBatchItem a, SpriteBatchItem b )
 		{
-            return ReferenceEquals( a.Texture, b.Texture ) ? 0 : 1;
+            return a.Texture.ComparisonCode.CompareTo(b.Texture.ComparisonCode);
```
tomspilman: We should consider making `Texture` inherit from [IComparable<T>](https://msdn.microsoft.com/en-us/library/4d7sx9hd(v=vs.110).aspx).  This way we solve this problem for us in `SpriteBatch` and give end users a way to do the same type of sorting in their own code without having to solve this issue themselves.

tomspilman:
So I agree that this is the right thing concept.  We need some numerical constant to sort by here and since there is no consistent "pointer" value (probably what MS XNA does) we have to make one.

However I have a few problems with the implementation here.  I've put my notes inline.

MonoGame.Framework/Graphics/Texture2D.cs
tomspilman: Also `_comparisonCode` should be marked as `readonly` if possible.

polsys:
Thanks for great feedback, I'll make the changes when I get to my dev PC later today.

I for one prefer making all `Texture` classes comparable instead of patching in support for this specific use case, if we are okay with that addition. In that case I'll remove the internal `ComparisonCode` property in favor of `CompareTo`.

polsys:
Done with the changes. Have to say that the code looks much more readable now with `IComparable`, with just a simple
```
a.Texture.CompareTo(b.Texture)
```

tomspilman:
Yeah...  this is a lot cleaner and a more useful feature.

@KonajuGames - Thoughts?

prollin:
Texture doesn't implement IComparable in XNA. This is probably an harmless deviation from the original XNA behavior though.

Regardless of implementing IComparable or not, It might be slightly better to compare the native texture handle to be future proof with this: if MonoGame is extended to support creating Texture instances from existing native handles (useful for interop with 3rd party libraries) the proposed implementation wouldn't be optimal.

tomspilman:
> This is probably an harmless deviation from the original XNA behavior though.

That is how I felt about it.  Also it makes it easy for those implementing their own rendering systems to sort by texture...  instead of having to invent their own work around.

>  if MonoGame is extended to support creating Texture 
> instances from existing native handles 

 I see what you are saying.

 However do you think it is a problem that sorting by texture handle might produce different results for OpenGL vs DirectX?

KonajuGames:
> However do you think it is a problem that sorting by texture handle might produce different results for OpenGL vs DirectX?

If your visual end result is dependent on texture draw order, then you shouldn't be using SpriteSortMode.Texture.  You would use FrontToBack, BackToFront or Immediate.  The only guarantee with Texture is that batches with the same texture get rendered together.

The native handle for textures does make sense for a sort criteria.  It is already guaranteed to be unique per texture and no extra maintenance or class members are required.

prollin:
@KonajuGames summarized what I was about the respond perfectly.

We could add a generic `GetPlatformTexture()` method (or readonly property) returning an `IntPtr` with the encoded native pointer/name/id if we wanted to do the comparison in the generic `Texture` class (something that could be extended to all `GraphicsResource` classes later on)

tomspilman:
> We could add a generic `GetPlatformTexture()` method (or readonly property) returning an `IntPtr`

Yeah...  I think that is reasonable.

So the `CompareTo` becomes:

```
public int CompareTo(Texture other) 
{ 
    if (other == null) 
       return 1; 
    return GetPlatformTexture().CompareTo(other.GetPlatformTexture());
} 
```

Then we add a new abstract method:
```
protected abstract IntPtr GetPlatformTexture();
```
Except that abstract is now a second internal virtual call...  not very optimal if you are going to be doing a sort on it IMO.

So maybe a better design is closer to the "readonly property" @prollin suggests that has to be set during texture construction to a valid `IntPtr` based on the native handle.

Thoughts?

@polsys - Sorry to make more work for you here again.

KonajuGames:
I was thinking that a common name for this `IntPtr` property in .NET is `Handle`.

tomspilman:
> I was thinking that a common name for this `IntPtr` property in .NET is  Handle .

Sure... probably should be `TextureHandle` or `GetTextureHandle` depending on how it is done.

polsys:
XNA probably does the comparison internally with the native pointer, and initially I planned to do so. However, I chose to add a new member because
* it required no changes to platform-specific portions
* as a bonus, made unit testing this `SpriteSortMode` possible, as the draw order depended on the load order

This is of course just my two cents, and I'm fine with making `CompareTo` use native handles, if decided so. Both ways have their own pros and cons, though minor IMO.

mgarstenauer:
The addition of an `IntPtr Texture.Handle` property would also help us in our engine -- a useful addition!

Whereas, IMHO, I don't think adding `CompareTo` to `Texture` is necessary or helpful for people who implement their own rendering system. Usually, when sorting draw calls the texture is just one of many sort criteria. For efficiency all sort criteria are added into a sort key (similar to [this article](http://realtimecollisiondetection.net/blog/?p=86)). The draw calls are then sorted by comparing the keys -- calling a `texture.CompareTo` method would be inefficient in this case. 
Another aspect: Textures aren't comparable by nature. What is the order of two textures? I don't think that this can be generalized. Textures can be ordered by native pointer, size, format, order of loading, etc. Therefore, I would not add `IComparable<T>`.

tomspilman:
> it required no changes to platform-specific portions

Yeah...  that is always a bit daunting and makes the PR bigger, but it really isn't a good reason to not do it.


> made unit testing this `SpriteSortMode` possible, as the draw order depended on the load order

That is a more compelling reason IMO to consider not using the native pointer.

How would we unit test this if it did use the native pointer?  Could it be done?

If a test isn't possible then that makes me lean towards leaving the implementation as is.


Also thinking about this more...  even though code shouldn't depend on the order in `SpriteSortMode.Texture`, I assure you there are games written right now where it probably does.  Are users not better off with ports to other platforms working consistently here?

tomspilman:
> The addition of an `IntPtr Texture.Handle` property would also help us in our engine

Yeah...  i see what you are saying.  Having something exposed that can be used as a unique identifier for sorting is more useful than just implementing `IComparable<T>`.

However i'm still sort of dubious that this should be the texture handle.   Like for something like a DirectX texture there are two objects...  the resource and the view.  Which one is the handle?  And for a render target there are 4 objects.  The concept of just a simple handle becomes confusing itself.

Wow...  this whole thing which seemed simple got complex quick!  Still I love having all of you commenting on this... if not we probably would have already made a bad decision here.

mgarstenauer:
> Like for something like a DirectX texture there are two objects... the resource and the view. Which one is the handle?

Good point.

Here is another solution (as used in our engine): For consistent result on all platforms assign a sort ID on the fly to each texture.
- Add an internal field/property `int Texture.Id`.
- Before `SpriteBatch.Begin` all IDs are cleared (e.g. -1).
- Assign a new ID (increment counter) to each texture when `SpriteBatch.Draw` is called.
- In `SpriteBatch.End` sort textures by ID, submit the batch, and clear the IDs.

This way sort order is identical on all platforms. It depends only on order of the `SpriteBatch.Draw` calls, which is consistent unless you change your game logic ... or add multi-threading which is a whole other topic.

Another advantage is that the range of IDs is minimal (0 to *number textures in batch* - 1). This can be useful when you want to squeeze the ID into a larger sort key, but only have a few bits for storing the ID.

tomspilman:
> For consistent result on all platforms assign a sort ID on the fly to each texture.

Well that is basically what this PR is already doing except it isn't "on the fly"...  it is done at texture creation and is totally thread safe.

To solve your need for an identifier for more complex sorting we could instead expose this as `Texture.Id` or `Texture.SortKey` or something like that instead of using `IComparable<T>`.

MonoGame.Framework/Graphics/Texture.cs
```diff
@@ -4,14 +4,18 @@
...
 namespace Microsoft.Xna.Framework.Graphics
 {
-	public abstract partial class Texture : GraphicsResource
+	public abstract partial class Texture : GraphicsResource, IComparable<Texture>
 	{
 		internal SurfaceFormat _format;
 		internal int _levelCount;
 
+        private readonly int _comparisonCode = Interlocked.Increment(ref _lastComparisonCode);
+        private static int _lastComparisonCode;
```
tomspilman: Was thinking maybe we want this to default to `Int32.MinValue`.  This way we don't get a sudden change in sort order if the count ever rolls over from `Int32.MaxValue`.

I was curious and it would take creating one new texture per second for over 136 years for this value to fully roll over.  So i'm not very concerned about the numerical limits here.

mgarstenauer:
> Well that is basically what this PR is already doing except it isn't "on the fly"... it is done at texture creation and is totally thread safe.

Correct, the only difference is that the IDs "on the fly" have minimal range, which was important in some of my cases. 

> To solve your need for an identifier for more complex sorting we could instead expose this as Texture.Id or Texture.SortKey or something like that instead of using IComparable<T>.

Yes, but adding a new property probably needs some more consideration. For this PR I suggest to keep the API as is (i.e. keep the Id internal and avoid adding `IComparable<T>`.)

MonoGame.Framework/Graphics/Texture.cs
polsys: I'm not sure if initializing it to `Int32.MinValue` would be of any real benefit... If you have created over two billion `Texture` instances, a change in sort order would be the least of your problems :)

polsys:
I have to agree with @tomspilman on texture handles... If there is no single 'handle' concept, which one to use? What if some platform has 64-bit handles, or an entirely different concept (a complex .NET object, for example)? This is why I preferred not messing with platform code.

This is surprisingly complex... but important to do well with such a core feature as `SpriteBatch`.

prollin:
> What if some platform has 64-bit handles, or an entirely different concept (a complex .NET object, for example)? This is why I preferred not messing with platform code.

An `IntPtr` handle still works here; for a complex .NET object, a good trick is to use GCHandle.ToIntPtr(GCHandle.Alloc(<object>, GCHandleType.Normal))`

I am not sure I understand why unit tests can't work with this.

Regarding D3D, I am pretty sure the view is what you want to use since it is what is bound to the pipeline and would result to a state change.

polsys:
This has been a great discussion, but we have to keep in mind this is only about comparing `Texture` equality. It really does not matter where the comparison key comes from, as long as it is unique to the instance. For that reason, I'd avoid adding code just for this to every `Texture` derivative for every platform. Introducing some new `NativeHandle` member is IMO out of the scope of this PR.

I agree with @mgarstenauer that textures really aren't comparable, and therefore implementing `IComparable` might cause confusion. Yet I guess most of the cases they are sorted to implement batching like this. At least I could write a doc comment explaining the behavior and warning not to rely on the sort order - we aren't letting users see the key anyway. Or we could make `CompareTo` internal and not cause any visible API change (nor benefit) to users.

This gets really hard to test if there's no deterministic sort order (if going with the native handle option, there could be some mock `Texture2D` giving predictable 'handles', but that would get complex and only test `SpriteBatch`, not real-world, behavior). Of course, having this testable is quite minor now, since the test will mostly track sorting regressions. A unit test would have prevented this bug in the first place, though.

tomspilman:
> Introducing some new `NativeHandle` member is IMO out of the scope of this PR.

I agree... this is about an identifier for sorting only.  If in the future a PR adds some `NativeHandle` feature we can then discuss replacing the identifier with the value of it.

> therefore implementing `IComparable` might cause confusion.

That is true... we don't want people to think it is doing pixel comparison or anything, but also the `CompareTo` call isn't useful for others needing to sort by multiple states.

So maybe just making the "comparison code" value a public getter removes the confusion and makes it more useful to engine developers.  I would suggest an name that is more accurate to its use...  `SortCode`, `SortingId`, or `SortingKey` or something like that.  

Then we can document that is it is a consistent value specifically for sorting by texture.

> This gets really hard to test if there's no deterministic sort order

I agree.... and we have enough hard problems to test.  We should just leave any `NativeHandle` discussion for when someone spends the time to submit a `NativeHandle` PR.

prollin:
@polsys the overhead of properly comparing textures based on their native handle is a one time implementation cost and each platform implementations seem straightforward (only 2 implementations afaik: OpenGL and DirectX).

The point of sorting by texture is to minimize the performance cost of swapping from one texture to another: the only way we can guaranty the texture is the same, is by comparing the native resource.

IComparable is confusing here, what is being compared? I would bet this is why XNA didn't use it.

I would suggest to add an internal CompareTo method that simply compare the existing native handle; this solution requires less code that any solution provided here and is fool proof.

tomspilman:
> would suggest to add an internal CompareTo method that simply compare the existing native handle; this solution requires less code that any solution provided here and is fool proof.

I disagree.

 - What is the native handle for each platform isn't even clear.  
 - It makes a non-platform specific feature platform specific.
 - It is harder to unit test.

Once someone implements some sort of native handle feature we can replace the generation of the sorting identifier with it and fix the unit test.

tomspilman:
Also while I love the discussion and it usually helps...  but this PR is about to derail.  We've seen this happen tons of times... no one agrees on a direction.  Any minute now this thread will go silent and nothing will get merged.

In these cases moving forward  @KonajuGames or i will just make an executive decision... right or wrong.  This lets things move forward and if issues occur a new PR can be submitted to fix them.

prollin:
> What is the native handle for each platform isn't even clear.
OpenGL is well defined and DirectX can simply sort by both View and Resource if necessary.

> It makes a non-platform specific feature platform specific.
As a matter of fact, it is a platform specific feature; some platform might benefit from a custom sorting mechanism here

> It is harder to unit test.
There are solutions, but unit testing graphics stuff is always hard; the best way I know to properly unit test GPU related features is to record and compare the OpenGL/DirectX command list sent to the driver.
But again, I don't see how this would affect unit testing here since the only issue for a problem in the sorting code would be a slight performance degradation (which can be tested separately)

> Also while I love the discussion and it usually helps... but this PR is about to derail. We've seen this happen tons of times... no one agrees on a direction.
As simple as this PR meant to be, the proposed solution seems flawed and could create confusion; if it derails then it probably means it needs to be modified and not merged.

polsys:
Currently `SpriteSortMode.Texture` does not work at all, and that's what this PR is trying to fix. If some platform needs special sorting, I'm all in for implementing such behavior. But AFAIK current platforms benefit lots from just simple texture sorting.

I currently fail to see how using a native handle would improve performance, unless there are multiple `Texture` instances pointing to same native object, which probably is not the common use case. And there would be more than 2 implementations: all different `Texture` derivatives (`Texture2D`, `RenderTarget2D`...) on DirectX, OpenGL, PS4 (?) and so.

I'll remove the `IComparable` implementation... it would have confused more than helped. The `SortingKey` property should solve this current bug well enough. Since this raises so much discussion, should it be an internal property at least until the implementation has stabilized?

mgarstenauer:
I agree with @polsys. Remove interface `IComparable`, keep `SortingKey` internal.
Then I say it is good to merge.

Public `NativeHandle`/`SortingKey` properties need to be discussed in the future at broader level. When I think of it, this also affects other graphics resources (vertex buffers, effects, ...). We need to be able to sort them as well.

prollin:
@polsys @mgarstenauer seems like a good compromise.

> I currently fail to see how using a native handle would improve performance, unless there are multiple Texture instances pointing to same native object, which probably is not the common use case

It is less about performance and more about correctness. I honestly doubt it will ever be an issue with this implementation but it seemed to be more fool proof to use the native handle.

tomspilman:
@KonajuGames - If you are good with this then lets merge it.

KonajuGames:
I'm good with it.  Merging,

polsys:
A big thank you to everybody who provided feedback - this was a good and educating discussion.

PR #3421, opened 2015-01-17
WinRT MediaLibrary speedup

Nezz:
The async IO is very slow on WinRT. When parsing the storage folder for music we made two calls: `GetFilesAsync` to get the files and `GetFoldersAsync` to get subfolders to parse. This PR replaces these two calls with a single `GetItemsAsync` call, which makes the MediaLibrary load almost twice as fast.


Nezz:
@mgbot Do you work! :D

tomspilman:
Thanks... merging!

PR #3425, opened 2015-01-18
Modern MessageBox and KeyboardInput

Nezz:
Rebase of #3374


Nezz:
@tomspilman Anything against the merge?

MonoGame.Framework/Input/MessageBox.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Xna.Framework.Input
+{
+    public partial class MessageBox
+    {
+        public static bool IsVisible { get; private set; }
+
+        public static async Task<int?> Show(string title, string description, IEnumerable<string> buttons)
```
tomspilman: Can we get some docs for this?

MonoGame.Framework/Input/MessageBox.cs
```diff
@@ -0,0 +1,38 @@
...
+                throw new ArgumentException("Invalid number of buttons: one to three required", "buttons");
+
+            var result = await PlatformShow(title, description, buttonsList);
+
+            IsVisible = false;
+
+            return result;
+        }
+
+        public static void SetResult(int result)
```
tomspilman: What is this function for?  Is this really just a `Cancel` or `Close`?

tomspilman:
Note this is a fix for https://github.com/mono/MonoGame/issues/3313.

MonoGame.Framework/Input/MessageBox.cs
tomspilman: Maybe add a extremely simple `<example>` block to the docs too?

tomspilman:
I assume here at some later date someone should add support for the remaining platforms.  Window, Linux, and Mac desktops?

MonoGame.Framework/Input/MessageBox.cs
Nezz: Yes. They are mainly useful for testing. Since they are supported on all mobile mobile platforms (except WP8.0, which will soon be gone) I decided to add them.

Nezz:
I added the documentation.

Yes, it would be great if support could be added. The existing implementations can be easily reused.

MonoGame.Framework/Input/MessageBox.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <param name="result">Result to return</param>
+        /// <exception cref="System.Exception">Thrown when the message box is not visible</exception>
+        /// <example>
+        /// <code>
+        /// var colorTask = MessageBox.Show("Color", "What's your favorite color?", new[] { "Red", "Green", "Blue" });
+        /// MessageBox.SetResult(0);
+        /// var color = await colorTask;
+        /// </code>
+        /// </example>
+        public static void SetResult(int result)
```
tomspilman: Do you mind renaming this then to either `Close` or `Cancel` then to better represent what it does?

Also shouldn't it be.... `int? result`... to be able to emulate all possible results from `Show`?

Nezz: Okay, I will rename it to `Cancel`.

That's correct, it should be `int?`.

Nezz:
Done.

tomspilman:
Thanks @Nezz.

@KonajuGames  - If you are good with this then lets merge it.

KonajuGames:
I'm good with this.  The only minor comment would be that the defacto standard for async/await naming is the addition of `Async` to the method name.  Since we don't have any blocking non-async versions, I don't think this is a blocking (pun intended) issue.

tomspilman:
Ok... merging!

tomspilman:
It would be good if someone could write a doc page on these APIs with more detailed examples.

MonoGame.Framework/Input/MessageBox.cs
```diff
@@ -0,0 +1,63 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Xna.Framework.Input
+{
+    public partial class MessageBox
```
tomspilman: @Nezz - Can we make this a static class?  Or is partial not allowed on static classes?

My problem is that in the docs...
 
  http://www.monogame.net/documentation/?page=T_Microsoft_Xna_Framework_Input_MessageBox

... it shows you can create an instance of `MessageBox` which is obviously not the intended use.

If not making it a static class... maybe make the default constructor private?

tomspilman: Same issue with `KeyboardInput`.

KonajuGames: Partial is allowed on static classes.  We use it already in other MonoGame code.

Nezz: It didn't build on one of the platforms, but I'll check again.

PR #3426, opened 2015-01-18
linux fixes and improvements

cra0zy:


dellis1972:
Looks good. i assume we use 32bit sharp font on Linux because the mono on that doesn't support x64 yet :/

PR #3431, opened 2015-01-19
Opening broken .mgcb files throws exceptions

emarcotte:
Sorry if this is a duplicate pull request... I didn't see another like it. When opening a broken MGCB file in pipeline.exe I am getting:

    System.NullReferenceException: Object reference not set to an instance of an object
      at MonoGame.Tools.Pipeline.PipelineController+<OpenProject>c__AnonStorey0.<>m__0 (System.String msg, System.Object[] args) [0x00027] in /home/emarcotte/src/MonoGame/Tools/Pipeline/Common/PipelineController.cs:218 
      at MGCB.MGBuildParser.ShowError (System.String message, System.Object[] args) [0x00020] in /home/emarcotte/src/MonoGame/Tools/MGCB/CommandLineParser.cs:480 
      at MGCB.MGBuildParser.ParseArgument (System.String arg) [0x0008b] in /home/emarcotte/src/MonoGame/Tools/MGCB/CommandLineParser.cs:336 
      at MGCB.MGBuildParser.Parse (IEnumerable`1 args) [0x00027] in /home/emarcotte/src/MonoGame/Tools/MGCB/CommandLineParser.cs:190 
      at MonoGame.Tools.Pipeline.PipelineProjectParser.OpenProject (System.String projectFilePath, MGCB.ErrorCallback errorCallback) [0x00053] in /home/emarcotte/src/MonoGame/Tools/Pipeline/Common/PipelineProjectParser.cs:238 
      at MonoGame.Tools.Pipeline.PipelineController.OpenProject (System.String projectFilePath) [0x0006a] in /home/emarcotte/src/MonoGame/Tools/Pipeline/Common/PipelineController.cs:219 
      at MonoGame.Tools.Pipeline.MainWindow.OnShowEvent () [0x00063] in /home/emarcotte/src/MonoGame/Tools/Pipeline/Gtk/MainWindow.cs:115 
      at MonoGame.Tools.Pipeline.Program.Main (System.String[] args) [0x00045] in /home/emarcotte/src/MonoGame/Tools/Pipeline/Program.cs:53 

It seems like the `View` property isn't being set on the `PipelineController` but is required by some other code (I didn't chase down this thread very far). The _view is required by the constructor, and seems to work properly when I switch to using it. I can try to dig down further into removing `View` if it isn't needed.

Here's a sample .mgcb file that caused the error:

    /outputDir:bin
    /intermediateDir:obj
    /platform:Linux
    /config:AnyCPU
    /profile:Reach
    /compress:False
    /incremental

The /incremental is invalid and it crashes trying to read it. This happens both at startup (opening file from history) and just doing file -> open. After this change the error message is displayed in the middle.

### DIFF
--- a/Tools/Pipeline/Common/IController.cs
+++ b/Tools/Pipeline/Common/IController.cs
@@ -44,7 +44,7 @@ namespace MonoGame.Tools.Pipeline
         /// <summary>
         /// The view this controller is attached to.
         /// </summary>
-        IView View { get; set; }
+        IView View { get; }
 
         /// <summary>
         /// Triggered when the project starts loading.
--- a/Tools/Pipeline/Common/PipelineController.ExcludeAction.cs
+++ b/Tools/Pipeline/Common/PipelineController.ExcludeAction.cs
@@ -30,7 +30,7 @@ namespace MonoGame.Tools.Pipeline
 
             public void Do()
             {
-                _con._view.BeginTreeUpdate();
+                _con.View.BeginTreeUpdate();
 
                 foreach (var obj in _state)
                 {
@@ -40,19 +40,19 @@ namespace MonoGame.Tools.Pipeline
                         if (item.OriginalPath == obj.SourceFile)
                         {
                             _con._project.ContentItems.Remove(item);
-                            _con._view.RemoveTreeItem(item);
+                            _con.View.RemoveTreeItem(item);
                             break;
                         }
                     }
                 }
 
-                _con._view.EndTreeUpdate();
+                _con.View.EndTreeUpdate();
                 _con.ProjectDirty = true;
             }
 
             public void Undo()
             {
-                _con._view.BeginTreeUpdate();
+                _con.View.BeginTreeUpdate();
 
                 foreach (var obj in _state)
                 {
@@ -64,10 +64,10 @@ namespace MonoGame.Tools.Pipeline
                     item.ResolveTypes();
 
                     _con._project.ContentItems.Add(item);
-                    _con._view.AddTreeItem(item);                        
+                    _con.View.AddTreeItem(item);
                 }
 
-                _con._view.EndTreeUpdate();
+                _con.View.EndTreeUpdate();
                 _con.ProjectDirty = true;
             }
         }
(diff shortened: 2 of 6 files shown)


tomspilman:
What is weird is why do we have...

https://github.com/emarcotte/MonoGame/blob/develop/Tools/Pipeline/Common/PipelineController.cs#L21
```
private readonly IView _view;
```

... and...

https://github.com/emarcotte/MonoGame/blob/develop/Tools/Pipeline/Common/PipelineController.cs#L62
```
public IView View { get; set; }
```

... in the same class?

I suspect this is the actual bug here.

@cra0zy @dellis1972 @jamesford42 ???

emarcotte:
Yep. I took the path of "getting the error fixed." I'm willing to do some more clean-up if someone is willing to tell me which one is the proper route!

KonajuGames:
I'm guessing the public property should reference the private field.​  And
that would require removal of the set in the public property as the private
field is read only.

tomspilman:
Let's see what the others say.  They have been actively working in that code lately.

dellis1972:
That new View property (the get;set one) was added in
https://github.com/emarcotte/MonoGame/commit/b9e06ea76fc16e3dc22a767b19abe8b96c62015c

and is used

https://github.com/emarcotte/MonoGame/commit/b9e06ea76fc16e3dc22a767b19abe8b96c62015c#diff-3855c40038919558ecfd4fbdf7c1c113R74

but is also set in the constructor. My guess is one of them can go :)

tomspilman:
> My guess is one of them can go :)

So that is weird...  in the Windows form code it does...

```
_controller.View = this;
```
But the controller is passed the IView within the constructor too...

https://github.com/emarcotte/MonoGame/blob/develop/Tools/Pipeline/Common/PipelineController.cs#L72

Seems like we have two different ways of setting the IView which is odd to me.

cra0zy:
I agree with @dellis1972

emarcotte:
I removed _view, since it seems to be used in a bunch of other files (even if they're internal or whatever, it seems odd to hide variables). I don't have windows to test on, but I ran a few tests on linux and things seem reasonable.

tomspilman:
This seems reasonable to me.

@cra0zy @dellis1972 - This reminds me we need to figure out some sort of automated testing solution for the Pipeline tool.  Eventually it will be complex enough that testing for bugs will become an issue.  We should consider our options now.

tomspilman:
Thanks @emarcotte !

tomspilman:
> Opening a broken file then pressing save does still result in an error tho :)

Well that is a separate issue to discuss.  Can you give details on that crash?

emarcotte:
Yep. I meant to log a task for that last night but then I remembered I had to prepare for work :-1:

PR #3436, opened 2015-01-20
Multi Game/GraphicsDevice within single process feature

### LINKED ISSUE #1823: Multi-Device Rendering
I'm testing out having multiple sub windows each with their own GraphicsDevice.  I was able to locate a bunch of places where monogame isn't multi-device friendly, static cache in Effect, Sampler States, RenderStates ( Blend, Raster, etc).  I've managed to get around all of them, but now I seem to crash on one I can't figure out. Does anyone know what could cause this to crash:

```C#
//In TextureCollection.cs

#elif DIRECTX
                if (_textures[i] == null || _textures[i].IsDisposed)
                    pixelShaderStage.SetShaderResource(i, null);
                else
                    //HERE
                    pixelShaderStage.SetShaderResource(i, _textures[i].GetShaderResourceView());
```

It seems to me like the texture is loaded from the previous device context, but I was able to see it for one frame before it crashed.

Any ideas?

Also, would my multi-device rendering changes be useful to check in? As it stands, without them the Debug.Assertions just fail.

christallire:
this commit resolves #1823

- FEATURE: GraphicsDeviceContext added for backward compatibility with XNA framework
(Squashed)
- REFACTOR: EffectCache moved from Effect to GraphicsDeviceContext
- REFACTOR: Moved ResetStates in GraphicsDevice to GraphicsDeviceContext
- REFACTOR: Default BlendState are moved to GraphicsDeviceContext
- TASK: Defensive code against usage of GraphicsResource.Current
- FIX: (Windows) Closing game window shouldn't close whole application
- TASK: Build definitions update for GraphicsDeviceContext
- WORKAROUND: Hacks to workaround usage of static member _allWindows in WinFormsGameWindow
- TASK: DebugAssertion for GraphicsDeviceContext added to GraphicsResource
- REFACTOR: Proper disposal of GraphicsResource

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsResource.cs
+++ b/MonoGame.Framework/Graphics/GraphicsResource.cs
@@ -40,6 +40,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
@@ -115,9 +116,19 @@ namespace Microsoft.Xna.Framework.Graphics
                 {
                     var target = resource.Target;
                     if (target != null)
+                    {
+                        // Do not dispose another device's graphics resource
+                        if (target is GraphicsResource)
+                        {
+                            if ((target as GraphicsResource).GraphicsDevice != GraphicsDeviceContext.Current.GraphicsDevice)
+                                continue;
+                        }
                         (target as IDisposable).Dispose();
+                    }
+
+                    // We can remove weak handle from collection safely since we have cloned resources dictionary
+                    resources.Remove(resource);
                 }
-                resources.Clear();
             }
         }
 
@@ -193,6 +204,23 @@ namespace Microsoft.Xna.Framework.Graphics
         public override string ToString()
         {
             return string.IsNullOrEmpty(Name) ? base.ToString() : Name;
+        }
+
+        protected static void ThrowIfGraphicsDeviceContextNull()
+        {
+            if (GraphicsDeviceContext.Current == null)
+                throw new InvalidOperationException("The graphics device context couldn't be null");
+        }
+
+        [Conditional("DEBUG")]
+        internal static void DebugAssertGraphicsDeviceContext<T>(Utilities.ObjectFactoryWithReset<T> objectFactoryWithReset)
+            where T : GraphicsResource, new()
+        {
+            var graphicsDeviceContext = GraphicsDeviceContext.Current;
+            var value = objectFactoryWithReset.Value;
+
+            Debug.Assert(graphicsDeviceContext != null, "The graphics device context couldn't be null");
+            Debug.Assert(value.GraphicsDevice == null || graphicsDeviceContext.GraphicsDevice == value.GraphicsDevice, "Bounded GraphicsResource is different from current graphics context");
         }
 	}
 }
--- a/MonoGame.Framework/Windows/WinFormsGamePlatform.cs
+++ b/MonoGame.Framework/Windows/WinFormsGamePlatform.cs
@@ -134,7 +134,7 @@ namespace MonoGame.Framework
         
         public override void Exit()
         {
-            Application.Exit();
+            _window.Dispose();
         }
 
         public override bool BeforeUpdate(GameTime gameTime)
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -41,11 +41,13 @@ purpose and non-infringement.
 using System;
 using System.Collections.Generic;
 using System.Drawing;
+using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Windows.Forms;
 using Microsoft.Xna.Framework;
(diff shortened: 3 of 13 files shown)


tomspilman:
I'm unsure of these changes.

At first glance it seems you misinterpreted what I said about `GraphicsDeviceContext`.  My comment was aimed specifically at multithreaded rendering.  The plan is to move most of the guts of `GraphicsDevice` into `GraphicsDeviceContext` and then `GraphicsDevice` would inherit from it ensuring XNA compatibility.

The stuff for `WinFormsGameWindow` seems safe on its own as an improvement.

The rest has some good ideas...   yes we cannot store graphics device specific stuff in random statics.  That does need to be moved into the GraphicsDevice/Context.   Still I would have to review it much more in depth and it would probably require a bunch of changes before it could be merged.

@KonajuGames ?

tomspilman:
Also in general with big changes you really should communicate what you plan to do very clearly ahead of time.  I really hate having to say no when people submit stuff that took them a lot of effort to write.  Still when I don't think it is the right path forward I have no choice.

christallire:
Yes, Indeed. I should've communicate with you guys.
My bad, I hate to making excuses but I'm middle of something and it's requirement is rapidly expanding and I'm running out of time.

Although previous decisions I made (which is using MonoGame instead of Unity) is forcing me to implement such features yet no one requires,
But I'm still thinking it's good decision because I can contribute.

I'm going to tell little bit explanation about why i'm using GraphicsDeviceContext and .Current thread static member.
I got general idea from OperationContext.Current from WCF (windows communication foundation)
I needed to change as least as possible to keep source unchanged safe and make sure get this working with XNA compatibility.

Unless thread is not executing another Game's RunLoop, it's guaranteed to not interfere with another Game.

Previously i used AppDomain to run multiple games but it was bad choice since I had only one connection to connect to server and it's making everything difficult.

I'm looking forward to your comments.

MonoGame.Framework/Graphics/GraphicsResource.cs
```diff
@@ -115,9 +116,19 @@ internal static void DisposeAll()
...
+                        if (target is GraphicsResource)
+                        {
+                            if ((target as GraphicsResource).GraphicsDevice != GraphicsDeviceContext.Current.GraphicsDevice)
+                                continue;
+                        }
                         (target as IDisposable).Dispose();
+                    }
+
+                    // We can remove weak handle from collection safely since we have cloned resources dictionary
+                    resources.Remove(resource);
```
tomspilman: @tgjones  - This is another case of a `static` that is graphics device specific.  If you get some time maybe you can move this one into `GraphicsDevice` too?

MonoGame.Framework/Windows/WinFormsGameWindow.cs
```diff
@@ -195,7 +198,35 @@ internal WinFormsGameWindow(WinFormsGamePlatform platform)
 
             _form.KeyPress += OnKeyPress;
 
-            _allWindows.Add(this);
+            RegisterToAllWindows();
+        }
+
+        private void RegisterToAllWindows()
```
tomspilman: @christallire - Can you submit the changes to `WinFormsGameWindow` in a separate PR?

MonoGame.Framework/Windows/WinFormsGamePlatform.cs
```diff
@@ -134,7 +134,7 @@ public override void StartRunLoop()
         
         public override void Exit()
         {
-            Application.Exit();
+            _window.Dispose();
```
tomspilman: @christallire - What is this change for?  If you have a single window game will this still close the application correctly?

christallire: Yeah. But it is already been done @ https://github.com/KonajuGames/MonoGame/commit/d6a65c8d3a3b86b40366a0466e46f35b17decf64

MonoGame.Framework/Windows/WinFormsGameWindow.cs
christallire: Sure, i'll make a PR for it

MonoGame.Framework/Graphics/GraphicsResource.cs
tgjones: Yep, I've got it queued up after the `EffectCache` PR. Changing how `GraphicsResource` are disposed of will result in a subtle - and I think harmless - change of behaviour, but we can discuss that in the PR.

tomspilman:
@christallire - So I'm closing this PR.

With @tgjones changes in...

  https://github.com/mono/MonoGame/pull/3472
  https://github.com/mono/MonoGame/pull/3511
  https://github.com/mono/MonoGame/pull/3516
  https://github.com/mono/MonoGame/pull/3520
  https://github.com/mono/MonoGame/pull/3526

... I think we've addressed nearly all the issues you have here.

Also you've resubmitted a separate PR for https://github.com/mono/MonoGame/pull/3525.

All that remains is the `GraphicsResource.DisposeAll()` issue which I expect to see fix here soon.

As far as multithreaded rendering and multiple `GraphicsContext` objects...  that is something we're still working towards, but it will be different than how you approached it here.

PR #3437, opened 2015-01-20
Corrected thumbstick output range for GamePadDeadZone.Circular

mrhelmut:
Corrected some overclamping in thumbstick deadzone calculation which added stutter when GamePadDeadZone.Circular is used.


tomspilman:
Looks good to me.... Merging!

PR #3440, opened 2015-01-21
Application run loop workaround to support mono windows

christallire:
This change will make MonoGame running on Mono-Windows, which will help deploying MonoGame based games without .Net framework.


tomspilman:
If it works the same as what was there before then it should be fine.

Anyone have any time to test this change?

@KonajuGames - Thoughts?

KonajuGames:
I think it looks ok too.  Does the same job and increases compatibility.

christallire:
I found some issues while testing intensively. I'll make more changes to
resolve it
2015. 1. 22. 오전 9:26에 "KonajuGames" <notifications@github.com>님이 작성:

Issue #3442, opened 2015-01-21
Pipeline: select more files and right click


vancura:
Select more files in the tree and right click. Boom.

dellis1972:
This is on Mac OS, should also happen on Linux.

cra0zy:
@tomspilman This can be closed.

PR #3444, opened 2015-01-21
Fix for #3442 

cra0zy:
Fix for #3442


danzel:
@cra0zy Thanks for all of the work you've been putting in to MonoGame.
Do you think you could write more than "fix" in your commit messages?
When looking through history it is useful to know what a commit is actually doing, like "Fixes crash when right clicking the tree view with multiple items selected" or something like that.
Thanks!

cra0zy:
@danzel Sorry about that, from now on I'll put in a better suited explanations for commit messages.

danzel:
Awesome, keep up the good work :)

Issue #3447, opened 2015-01-21
Linux SpriteFont Processor doesn't use family names


CallumDev:
When building a spritefont file in Pipeline on Linux, mgcb searches for the font by filename, but not by family name. This means that a spritefont description won't build when asking for 'Arial', as the filename on most Linux systems is 'arial.ttf'. 

A way of fixing this would be to call into fontconfig, but should such a wrapper be implemented inside MonoGame or as a separate project?

cra0zy:
What do you mean by fontconfig?

CallumDev:
Fontconfig as in the library used for font configuration on Linux (libfontconfig.so)
[This](https://gist.github.com/CallumDev/7c66b3f9cf7a876ef75f) is an example of how it would be used

cra0zy:
I found another solution, run in terminal: "fc-list : file family", now the only question is which is better.

cra0zy:
@tomspilman This can be closed.

PR #3450, opened 2015-01-21
Fix for #3447

cra0zy:
Fix for #3447

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
+++ b/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
@@ -151,6 +151,9 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         private static void EnsureExecutable(string path)
         {
 #if LINUX || MONOMAC
+            if(path == "/bin/bash")
+                return;
+
             try
             {
                 var p = Process.Start("chmod", "u+x '" + path + "'");
--- a/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
@@ -76,7 +76,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
 #else
             System.Diagnostics.Process proc = new System.Diagnostics.Process();
             proc.StartInfo.FileName = "/bin/bash";
-            proc.StartInfo.Arguments = "-c \"fc-list : file family\"";
+            proc.StartInfo.Arguments = "-c \"fc-list : file family style\"";
             proc.StartInfo.UseShellExecute = false; 
             proc.StartInfo.RedirectStandardOutput = true;
             proc.Start();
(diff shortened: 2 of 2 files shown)


cra0zy:
@CallumDev better?

tomspilman:
Waiting on @CallumDev to reply before merging.

CallumDev:
Fails now on Condensed fonts (DejaVu Sans Condensed)
`/usr/share/fonts/TTF/DejaVuSansCondensed-Bold.ttf: DejaVu Sans,DejaVu Sans Condensed:style=Condensed Bold,Bold`

CallumDev:
Something like this (using fc-match), might be a better option

        string FindFontFileFromFontName(string fontname, string style)
		{
			System.Diagnostics.Process proc = new System.Diagnostics.Process();
			proc.StartInfo.FileName = "/bin/bash";
		        proc.StartInfo.Arguments = 
				string.Format ("-c \"fc-match -f '%{{file}}:%{{family}}\\n' '{0}:style={1}'\"", fontname, style);
			proc.StartInfo.UseShellExecute = false; 
			proc.StartInfo.RedirectStandardOutput = true;
			proc.Start();
			proc.WaitForExit ();
		        string s = proc.StandardOutput.ReadToEnd ().Trim ();
			var split = s.Split (':');
			//check font family, fontconfig might return a fallback
			if (split [1].Contains (",")) { //this file defines multiple family names
				var families = split [1].Split (',');
				foreach (var f in families) {
					if (f.ToLowerInvariant () == fontname.ToLowerInvariant ())
						return split [0];
				}
				//didn't find it
				return String.Empty;
			} else {
				if (split [1].ToLowerInvariant () != fontname.ToLowerInvariant ())
					return String.Empty;
			}
			return split [0];
		}

If you use it as `FindFontFileFromFontName("Arial","italic");` it'll return the name of the file if it can find it, or `String.Empty` otherwise

cra0zy:
@CallumDev What if someone writes font file name instead of the font family name... Anyway with my code I forgot to take into account that there can be styles like "Condensed Bold" and "Bold", if you would search for bold both would come up as correct.

KonajuGames:
We have a `ExternalTool` utility class in the Pipeline project with a `Run` method for running extenral runs and retrieving the output.  Saves duplicating the code for starting a process and waiting on it.

KonajuGames:
It should check if the specified font is a filename by checking if a file by that name exists (useful for people who keep the TTFs in their project).  If not, assume it is a family name and pass it through this function to convert it to a font filename.

cra0zy:
@KonajuGames Should it check for the filename only in the project dir, or should it check for the filename threw the system fonts too?

KonajuGames:
If you want to use a system font, specify the font family.  If you have the TTF file in your project, then you specify the filename relative to the .spritefont file.

cra0zy:
Once this is ok I am gonna rebase it.

CallumDev:
This looks good

tomspilman:
Ok... merging!

PR #3453, opened 2015-01-22
Removed the RGBA->BGRA swap in FontTextureProcessor

### LINKED ISSUE #3430: SpriteFont Texture Color are swapped from RGBA to BGRA even if processor is set to Color
This is a reproducible issue in Windows (dx AND gl).

https://www.dropbox.com/s/yuiip7y8nibry9c/FontBuildIssue.zip?dl=0

This is a project that basically shows the issue, the font is a number font from our game, i colored it to show the problem: 
1) at the top you can see the original texture from wich the font should be drawn (SpriteBatch.Draw)
2) the second and third line shows the actual result: R and B are swapped. (SpriteBatch.DrawString)

colorfontissue

KonajuGames:
It is already done elsewhere.

Fixes #3430

### DIFF
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -141,6 +141,9 @@
     <Compile Include="ContentPipeline\FontDescriptionTests.cs">
         <Platforms>Windows,MacOS,Linux</Platforms>
     </Compile>
+    <Compile Include="ContentPipeline\FontTextureProcessorTests.cs">
+        <Platforms>Windows,MacOS,Linux</Platforms>
+    </Compile>
     <Compile Include="ContentPipeline\IntermediateDeserializerTest.cs">
         <Platforms>Windows,MacOS,Linux</Platforms>
     </Compile>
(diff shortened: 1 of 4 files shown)


tomspilman:
Could we make a unit test for this?  Something like I wrote for the `TextureProcessor`...

I know you are busy with stuff, but if we're going to ask others to write unit tests we have to hold ourselves to the same standard.

tomspilman:
Note a recommended setup up unit tests for `FontTextureProcessor` in https://github.com/mono/MonoGame/pull/3454 as well.  Lets make sure two people don't go off and do the same work.

KonajuGames:
I will get the unit test added today.

tomspilman:
Thanks @KonajuGames .

tomspilman:
The test looks good.  I assume you flipped the bool in `ToXnaBitmap()` back to true to validate your test catches the original bug?

KonajuGames:
I did, and it caught it.​

tomspilman:
Great!  Merging!

PR #3454, opened 2015-01-22
Calculate line height smaller than line spacing correctly

Nezz:
When using SpriteFont textures, LineSpacing is the height of the tallest sprite (untrimmed).

MeasureString works incorrectly because it initializes the `finalLineHeight` variable with LineSpacing, then looks for taller sprites:

``` C#
if (currentGlyph.Cropping.Height > finalLineHeight)
    finalLineHeight = currentGlyph.Cropping.Height;
```

This will obviously never happen with SpriteFont textures. This PR modifies MeasureString so it initializes `finalLineHeight` with 0 and resets it to 0 at a line break.

I haven't tested it with SpriteFont descriptions where LineSpacing might be a correct value, so please test it. 

The biggest change is when you measure small characters, like a space which is represented as a 1x1 transparent pixel. It will return a height of 1:
`this.font.MeasureString(" ")` -> {18, 1}

How is the result of MeasureString "definied"?

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteFont.cs
+++ b/MonoGame.Framework/Graphics/SpriteFont.cs
@@ -233,7 +233,7 @@ namespace Microsoft.Xna.Framework.Graphics
                 if (c == '\n')
                 {
                     fullLineCount++;
-                    finalLineHeight = LineSpacing;
+                    finalLineHeight = 0;
 
                     offset.X = 0;
                     offset.Y = LineSpacing * fullLineCount;
--- a/MonoGame.Framework/Graphics/SpriteFont.cs
+++ b/MonoGame.Framework/Graphics/SpriteFont.cs
@@ -214,7 +214,7 @@ namespace Microsoft.Xna.Framework.Graphics
                 defaultGlyph = _glyphs[DefaultCharacter.Value];
 
 			var width = 0.0f;
-			var finalLineHeight = (float)LineSpacing;
+			var finalLineHeight = 0;
 			var fullLineCount = 0;
             var currentGlyph = Glyph.Empty;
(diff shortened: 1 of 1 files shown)


tomspilman:
> I haven't tested it with SpriteFont descriptions where LineSpacing might be a correct value, so please test it. 

Yeah... we cannot merge this change until we are 100% certain it doesn't further break SpriteFont rendering.  I just don't want to see SpriteFont get any less accurate than it is now.

We could setup a unit test for the `FontTextureProcessor` and `FontDescriptionProcessor` similar to what I did for the `TextureProcessor`....

This could at least get you some validation that the processor returns the correct metrics as XNA does.

Note a recommended setup up unit tests for `FontTextureProcessor` in https://github.com/mono/MonoGame/pull/3453 as well.  Lets make sure two people don't go off and do the same work.

Nezz:
Can you do that? I don't have any time I could allocate for this as it now works the way we expect it to work and I have to move on to other work items.

tomspilman:
Looks like @KonajuGames is going to add the tests for his PR.  Should just be a simple merge/rebase from develop to pull the tests into this PR then.  Or @KonajuGames can merge your changes into his PR if needed as well.

nkast:
@Nezz This looks to be the same issue as https://github.com/mono/MonoGame/issues/3941 & fix https://github.com/mono/MonoGame/pull/3945.

aaronfranke:
This PR has been in limbo for over 5 years. That's ridiculous. MonoGame developers, I'm calling you out - you need to review these, and either merge or reject them. Don't let these stay in limbo.

@tomspilman @harry-cpp @dellis1972 @Jjagg

PR #3458, opened 2015-01-23
Handle software buttons on WP8

### LINKED ISSUE #3457: [WP8] Software buttons are not handled by MonoGame
Ouch.

wp_ss_20150120_0002

Nezz:
Works fine with DrawingSurfaceBackgroundGrid. Please test with DrawingSurface.

Fixes #3457

### DIFF
--- a/MonoGame.Framework/WindowsPhone/WPGameWindow.cs
+++ b/MonoGame.Framework/WindowsPhone/WPGameWindow.cs
@@ -56,6 +56,7 @@ namespace MonoGame.Framework.WindowsPhone
         private DisplayOrientation _supportedOrientations;
         private DisplayOrientation _orientation;
         private Rectangle _clientBounds;
+        private Rectangle _actualBounds;
         private Game _game;
 
         #region Internal Properties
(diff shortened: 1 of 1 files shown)


Nezz:
My solution can't handle resuming, so don't merge this yet.

tomspilman:
Ok... was waiting to see if someone could jump in and test the DrawingSurface case as well.

tomspilman:
Anyone?

Nezz:
My approach doesnt work as expected in some cases like rendertarget reset. Os there a way to make graphics device act like the viewport is smaller?

tomspilman:
> My approach doesnt work as expected in some cases like rendertarget reset. 

Hum...   yeah that would be an issue.

Why can we not change the render target size instead?

Nezz:
The backbuffer's size is fixed. We have to force the game to make it look like the resolution is smaller so we won't draw under the area of the software button bar. Since the user can hide and show these we have to do that resizing while the game is running (like we do with Windows 8 because of snap).

As I've seen this is a general issue. Android added two new approaches for handling software buttons in KitKat and both of them require us to change the resolution while the game is running, but that's not supported at the moment. At least in that case the backbuffer resolution is changing, so we don't need a hack like we do on WP8.

tomspilman:
I'm not sure the best way to solve this.

Faking viewport size and forcing offsets seems like a possible solution.  However it has to be faked in the presentation parameters, graphics device manager, and touch inputs.

Offscreen render targets is only a partial solution and has performance implications.

Not sure of any others.

KonajuGames:
Android games I have seen such as Angry Birds Epic and Crossy Road simply
draw the software buttons over the game when the use skews them by sweeping
down from the top of the screen. No resizing of the view.

Nezz:
Yes, but that's only after KitKat. It's called immersive mode.

However, on WP8 there is no such thing. The emulator allows you to show and hide the software keyboard bar, but the devices I have seen don't.

PR #3459, opened 2015-01-23
Gtk Pipeline UI Improvements

### LINKED ISSUE #3456: Pipeline Tool Gtk# - Save option never enabled after editing.
When altering properties of a Processor or Importer the save button is never enabled. The only way to save changes is to hit the build button or use save as.

cra0zy:
This will fix #3456 plus few other things like "Delete" enabled when nothing is selected, "????" apearing in propertiesgrid when selecting empty space...

### DIFF
--- a/Tools/Pipeline/Gtk/MainWindow.cs
+++ b/Tools/Pipeline/Gtk/MainWindow.cs
@@ -253,16 +253,16 @@ namespace MonoGame.Tools.Pipeline
 
 		public void AddTreeItem (IProjectItem item)
 		{
-			projectview1.AddItem (projectview1.GetBaseIter(), item.OriginalPath, item.Exists);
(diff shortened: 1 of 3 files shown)


tomspilman:
Is littering `UpdateMenus()` all over the UI code really the best way to keep things updated in GTK?  On Windows usually you hook into either a menu activation event or an idle event to do this processing.

cra0zy:
I don't know of better way with Gtk. 2 problems with what you said are:
1) mac doesn't use menubar, that's why #3456 was occuring
2) The closest event to menubar activated is checking "File" submenu activated which will inturn make Ctrl + S not work until "File" submenu gets clicked...

tomspilman:
Weird...  you would think there would be a better way.  Merging!

PR #3461, opened 2015-01-24
Change Game.RunOneFrame() to Game.Run() in visual tests

### LINKED ISSUE #3423: Tests that use Game.RunOneFrame aren't actually executed in XNA
A pattern MonoGame uses in some of its tests, specifically the ones in fixtures derived from `VisualTestFixtureBase`, is to register an event handler with `Game.DrawWith`, and then call `Game.RunOneFrame()`. But I noticed while working in `VertexBufferTest.cs` that, when running these tests under XNA, the test window is never shown and the `DrawWith` event handler is never called, and so the test gives a false positive.

Only a few tests use this pattern:

* `IndexBufferTest.cs`
* `Texture2DTest.cs`
* `Texture3DTest.cs`
* `VertexBufferTest.cs`

The simplest fix is to replace these calls to `Game.RunOneFrame()` with `Game.Run()`. This works because the tests use a custom `Game` class, which defines a new `Run` method, which exits the game after some condition evaluates to true. By default, that condition is "exit after drawing one frame". So within the context of these tests, `Game.Run()` does exactly the same thing as `Game.RunOneFrame()`.

If, instead, we want to find the root cause so that we can continue using `RunOneFrame`, then... looking in the debugger immediately after `Game.RunOneFrame()` has been called, I can see that `IsActive`, as well as `doneFirstUpdate`, `doneFirstDraw`, etc. are all `false`. I have tried calling `Initialize` before `RunOneFrame`, but that doesn't help.

I can see in MG's implementation of `RunOneFrame()` that we initialise the game if it hasn't already been initialised. My best guess is that XNA doesn't do this - it expects the game to already be initialised (which it does in `Run`, but apparently not in `RunOneFrame`) . But I don't know enough about the XNA `Game` class to know how to do that initialisation. (Or put another way, I'm not sure how XNA's `RunOneFrame` is intended to be used at all.)

tgjones:
Fixes #3423. Tests that used Game.RunOneFrame() were giving false positives when run in XNA.

### DIFF
--- a/Test/Framework/Visual/IndexBufferTest.cs
+++ b/Test/Framework/Visual/IndexBufferTest.cs
@@ -25,7 +25,7 @@ namespace MonoGame.Tests.Visual
                 indexBuffer.GetData(readData, 0, 4);
                 Assert.AreEqual(savedData, readData);
             };
-            Game.RunOneFrame();
+            Game.Run();
         }
 
         [Test]
@@ -44,7 +44,7 @@ namespace MonoGame.Tests.Visual
                 Assert.AreEqual(0, readData[2]);
                 Assert.AreEqual(0, readData[3]);
             };
-            Game.RunOneFrame();
+            Game.Run();
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


tomspilman:
Great...  Thanks!

PR #3466, opened 2015-01-26
Small fixes

jamesford42:


MonoGame.Framework/Utilities/FileHelpers.cs
```diff
@@ -59,5 +59,23 @@ public static string ResolveRelativePath(string filePath, string relativeFile)
...
             return NormalizeFilePathSeparators(localPath);
         }
+
+        /// <summary>
+        /// Returns a path relative to the base path.
+        /// </summary>
+        /// <param name="basePath">The path to make relative to.  Must end with directory seperator.</param>
+        /// <param name="path">The path to be made relative to the basePath.</param>
+        /// <returns>The relative path or the original string if it is not absolute or cannot be made relative.</returns>
+        public static string GetRelativePath(string basePath, string path)
```
tomspilman: If you are going to add this you need to add a unit test for it.

https://github.com/mono/MonoGame/blob/develop/Test/Framework/UtilitiesTest.cs#L45

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -254,8 +255,14 @@ protected virtual Stream OpenStream(string assetName)
 			try
             {
                 string assetPath = Path.Combine(RootDirectory, assetName) + ".xnb";
-                stream = TitleContainer.OpenStream(assetPath);
 
+                // For Win8 metro applications you do not have access to Environment.CurrentDirectory nor does Path.DirectorySeparatorChar exist.
+#if !WINRT
+                if (Path.IsPathRooted(assetPath))
+                    assetPath = FileHelpers.GetRelativePath(Environment.CurrentDirectory + Path.DirectorySeparatorChar, assetPath);
```
tomspilman: This fix won't work if the absolute path is on another drive letter.

I suspect we need to just not call `TitleContainer.OpenStream` if the path is absolute... instead calling `File.OpenRead()` directly.

Also this should only be allowed on desktop platforms Linux, MacOS, Windows (not metro) and Android according to https://github.com/mono/MonoGame/issues/3165.

jamesford42: I agree, File.OpenRead ifdef'd for desktop platforms is a better approach. I'll make that change in a bit.

tomspilman: Can you get to this soon?

jamesford42:
@tomspilman Commited change as per your comment.

tomspilman:
Thanks... merging!

PR #3469, opened 2015-01-27
LZ4 moved to more consistent place

Shqrdx:
Instead of separation from main framework I suggest insertion the LZ4 namespace into Content.Pipeline itself. Than it will not be an eyesore in documentation.

### DIFF
--- a/MonoGame.Framework/Utilities/Lz4Stream/Lz4DecoderStream.cs
+++ b/MonoGame.Framework/Utilities/Lz4Stream/Lz4DecoderStream.cs
@@ -5,7 +5,7 @@
 using System;
 using System.IO;
 
-namespace Lz4
+namespace Microsoft.Xna.Framework.Content.Pipeline.LZ4
 {
 	public class Lz4DecoderStream : Stream
 	{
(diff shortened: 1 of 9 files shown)


MonoGame.Framework/Utilities/Lz4Stream/Lz4DecoderStream.cs
```diff
@@ -5,7 +5,7 @@
 using System;
 using System.IO;
 
-namespace Lz4
+namespace Microsoft.Xna.Framework.Content.Pipeline.LZ4
```
KonajuGames: This namespace should not have `Content.Pipeline` in it as it is not in the content build pipeline.

KonajuGames:
If we're going to alter these namespaces to fit, then they should be fully qualified rather than `LZ4` being at the assembly namespace level.  They're in a `Utilities` folder, so the namespace should include that to help denote these are utility types.

I would also be inclined to change the visibility level of the types from `public` to `internal` as they're not really intended to be used outside of the content pipeline and content manager.

@tomspilman ?

Shqrdx:
@KonajuGames And now ? How it is ?

MonoGame.Framework/Utilities/Lz4Stream/Lz4DecoderStream.cs
```diff
@@ -1,13 +1,17 @@
-﻿#define CHECK_ARGS
+﻿// MIT License - Copyright (C) The Mono.Xna Team
```
KonajuGames: The license and copyright header should not be added as this code came from somewhere else. It wasn't written by us.

tomspilman:
This seems reasonable to me.  @KonajuGames ?

KonajuGames:
Merging. Thanks @Shqrdx

PR #3470, opened 2015-01-27
New enum docs.

Shqrdx:
Docs for 
-VertexElementUsage
-VertexElementFormat
-TextureFilter
-TextureAddressMode
-SurfaceFormat
Fix licenses for all of those enums.
SurfaceFormat_Legacy is renamed to SurfaceFormatLegacy and marked internal instead of public.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteEffects.cs
+++ b/MonoGame.Framework/Graphics/SpriteEffects.cs
@@ -1,53 +1,28 @@
-#region License
-/*
-Microsoft Public License (Ms-PL)
-MonoGame - Copyright © 2009 The MonoGame Team
-
-All rights reserved.
-
-This license governs use of the accompanying software. If you use the software, you accept this license. If you do not
-accept the license, do not use the software.
-
-1. Definitions
-The terms "reproduce," "reproduction," "derivative works," and "distribution" have the same meaning here as under 
-U.S. copyright law.
-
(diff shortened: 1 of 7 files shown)


MonoGame.Framework/Graphics/SpriteEffects.cs
```diff
@@ -1,53 +1,28 @@
...
+    /// </summary>
     [Flags]
     public enum SpriteEffects
     {
+        /// <summary>
+        /// No options specified.
+        /// </summary>
 		None = 0,
+        /// <summary>
+        /// Rotate 180 degrees around the Y axis before rendering.
```
tomspilman: Are you sure this is rotation?  I think this is not rotation at all...  it is literally flipping the sprite so the top and bottom are switched, but the left and right are not.

Shqrdx: > Flip sprite to 180 degrees around the Y axis before rendering.
It is more good ?

tomspilman: Oh... I guess Y axis is the important bit here.  It is trying to describe flipping as a rotation around a single axis.  That seems confusing to me.  Maybe a better description:

> Render the sprite reversed along the X axis.

@KonajuGames ?

tomspilman:
Looks all good to me...  i'll merge once it finishes building.

Issue #3449, opened 2015-01-21
Request: remove effects from Effect.EffectCache when effect is disposed


tgjones:
Effects are cached in a private static field in `Effect.EffectCache`. But the only code that calls the internal `Effect.FlushCache()` is `Game.Dispose()`.

It took me a while to figure out that this is the root cause of why some of my visual tests are failing. I'm not using a `Game` object - I'm just creating lower-level objects like `GraphicsDevice` directly. But because I'm disposing the `GraphicsDevice` and other graphics resources after each test, any effects left in the cache are unusable. Any subsequent tests that use the same `Effect` don't work.

I'm not sure what the best solution is - it seems like cached effects should be removed when the `GraphicsDevice` used to create them (rather than the `Game`) is disposed? Or at least, when the `Effect` itself is disposed (I think `GraphicsDevice.Dispose` already calls `Dispose` on `GraphicsResource` objects, so that may be enough).

For now, I've worked around it with this, but it's not pretty:

``` csharp
typeof(Effect).GetMethod("FlushCache", BindingFlags.NonPublic | BindingFlags.Static).Invoke(null, null);
```

**EDIT**: I've just run into a similar issue with state objects - I get exceptions because they are associated with a different graphics device. Basically, I think I need to be able to [what Game is doing here](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Game.cs#L120) (or even better, have it happen implicitly when `GraphicsDevice` is disposed):

``` csharp
if (_graphicsDeviceManager != null)
{
    Effect.FlushCache();
    BlendState.ResetStates();
    DepthStencilState.ResetStates();
    RasterizerState.ResetStates();
    SamplerState.ResetStates();

    ...
}
```

tomspilman:
Yeah...  what needs to happen is similar to what https://github.com/mono/MonoGame/pull/3436 is doing.

We need to move the effect cache and the default render states into `GraphicsDevice`.  This should be as simple as moving `Effect.EffectCache` into `GraphicsDevice.EffectCache` and making it non-`static`.

This would be a good first step to better support for multiple `GraphicsDevice`s and would be a good first step towards support for multiple graphics contexts in the future.

tgjones:
Ah, cool. I saw that PR but missed that it included a fix for this. Should I wait until #3436 is resolved one way or another?

tomspilman:
> Should I wait until #3436 is resolved one way or another?

I don't think we'll merge that PR as is.

[...]

PR #3472, opened 2015-01-29
Create default BlendStates (Opaque, AlphaBlend, etc.) per GraphicsDevice

tgjones:
When this PR is finished, it will provide a fix for #3449.

This first batch of commits only affects `BlendState` - if this looks good, then I'll carry on with the other render states, and finally the effect cache.

I have added some tests and tightened up `BlendState` mutability.

I have refactored how blend states get applied, which required some changes to the platform-specific `GraphicsDevice` files. This allows us to centralise the handling of `GraphicsDevice`-specific clones of render states. (I actually wanted to have a `BlendState.Apply` method, which had the `if`/`else` for default blend states, and then called `BlendState.PlatformApply`, but this broke when compiled in 2MGFX, since there are no platform-specific files there. So for now, that code is in `GraphicsDevice.Apply`.)


Test/Framework/Visual/BlendStateTest.cs
```diff
@@ -0,0 +1,63 @@
...
+            Game.Run();
+        }
+
+        [Test]
+        public void ShouldNotBeAbleToMutateStateObjectAfterBindingToGraphicsDevice()
+        {
+            Game.DrawWith += (sender, e) =>
+            {
+                var blendState = new BlendState();
+                Assert.DoesNotThrow(() => blendState.AlphaBlendFunction = BlendFunction.ReverseSubtract);
```
tomspilman: If we are going to test one property we should test all of them.  I know they are implemented in a common pattern, but we should still test that all the setters are working correctly.

tomspilman:
Note I have an actual blend state visual test here I can share...

 https://gist.github.com/tomspilman/a22634d642b3772c4e6c

It draws a grid with the possible blend state combinations.  It could be combined with an image comparison to both validate blends work and that their results are accurate across platforms.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -262,6 +278,35 @@ public DepthStencilState DepthStencilState
...
+                // instead we use our GraphicsDevice-specific version of them.
+                var newBlendState = _blendState;
+                if (ReferenceEquals(_blendState, BlendState.Additive))
+                    newBlendState = BlendStateAdditive;
+                else if (ReferenceEquals(_blendState, BlendState.AlphaBlend))
+                    newBlendState = BlendStateAlphaBlend;
+                else if (ReferenceEquals(_blendState, BlendState.NonPremultiplied))
+                    newBlendState = BlendStateNonPremultiplied;
+                else if (ReferenceEquals(_blendState, BlendState.Opaque))
+                    newBlendState = BlendStateOpaque;
```
tomspilman: Maybe all this code should be moved to the `BlendState` setter.  This way it only occurs once on set and not on every draw call.  This would mean you need to keep two temporaries....`_blendState` and `_realBlendState` or something.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -262,6 +278,35 @@ public DepthStencilState DepthStencilState
             }
         }
 
+        internal void ApplyState(bool applyShaders)
```
tomspilman: I don't remember where the `applyShaders == false` case is used anymore.  Could it be it isn't needed?

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -262,6 +278,35 @@ public DepthStencilState DepthStencilState
...
+                else if (ReferenceEquals(_blendState, BlendState.NonPremultiplied))
+                    newBlendState = BlendStateNonPremultiplied;
+                else if (ReferenceEquals(_blendState, BlendState.Opaque))
+                    newBlendState = BlendStateOpaque;
+
+                // Blend state is now bound to a device... no one should
+                // be changing the state of the blend state object now!
+                newBlendState.BindToGraphicsDevice(this);
+
+                newBlendState.PlatformApplyState(this);
```
tomspilman: What is the benefit of separating the `BindToGraphicsDevice` and `PlatformApplyState`?

Test/Framework/Visual/BlendStateTest.cs
tgjones: Fair enough. I always go back and forth on this, in my own code - there is definitely some value in the repetitive `A.x == B.x` style of tests, but I'm never sure whether it's worth it. But more than happy to follow that rule in MG.

MonoGame.Framework/Graphics/GraphicsDevice.cs
tgjones: Good call - and that will remove the need for the separate `BindToGraphicsDevice` and `MarkBound` methods.

Test/Framework/Visual/BlendStateTest.cs
tomspilman: > I always go back and forth on this

I do some as well, but our goal is 100% code coverage... even if the test has a low likelihood of failing.

MonoGame.Framework/Graphics/GraphicsDevice.cs
tgjones: The only usage I can find is [here](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs#L259).

tomspilman: Ok... looks like the reason I did that was to optimize the state changes in that one case of Clear().  Since Clear() is pretty infrequent seems like a pretty tiny win for the extra complication.

We can maybe remove it... not sure yet.

MonoGame.Framework/Graphics/GraphicsDevice.cs
tgjones: I just realised that it's not done on every draw call - only when `_blendStateDirty == true`. But I still think it's worth making this change - the code becomes a bit cleaner.

tgjones:
@tomspilman if the tests pass, then I think I've incorporated all of your suggestions, and this is ready to review again. Let me know if you think I'm heading in the right direction - if so, I'll carry on with the other render states and effect cache.

tomspilman:
> Let me know if you think I'm heading in the right direction

I think generally yes.

However I'm not sure about making one test per-blend state.  While this way it will point out specifically which blend failed....  it seems and looks like a lot more complexity to have a couple dozen reference images.  If we decide to add one more degree of testing it could explode into 100s of reference images.

@KonajuGames  @Nezz - Thoughts?

Nezz:
In my opinion the more small, atomic tests we have the better.

tgjones:
I'll wait till there's a decision before carrying on; if we decide to have fewer tests / reference images, then I'll modify my last commit, to avoid bloating the MG repo.

That said, my vote is also for small, atomic tests, even if it means there are a lot of them.

tomspilman:
@tgjones - FYI... there is some sort of conflict in this PR with the latest changes to the develop branch.

tgjones:
Uh-oh, here's where my limited Git knowledge manifests itself. Should I be doing a rebase?

tomspilman:
> Should I be doing a rebase?

Or merge from develop.  It will mark conflicted files and you just go fix them up.

It happens when some other PR makes a change to the same lines of code as you do.

Nezz:
Yes, that's what you need.

tgjones:
It's not the first time I've done a rebase, but I do it infrequently enough to forget how to do it between each time :) Hopefully I got it right.

tomspilman:
I was gonna say...  this PR has 182 changed files now. so you must of messed up the rebase.  But actually you did not... the rebase looks fine.  You really did add/change 182 files in this PR.

@KonajuGames - Thoughts on having all these individual images and test for blend state?

KonajuGames:
I'll have a closer look soon when I'm not on a phone.

KonajuGames:
It's a comprehensive set of tests, and I think @tgjones has done a great job in setting them up.

Do we know (roughly) how much extra time this adds to a test pass?

KonajuGames:
Never mind.  The TeamCity posts tell me that information.
Before: 376 tests at ~16 minutes.
After: 546 tests at ~22 minutes.

KonajuGames:
I agree with @tgjones and @Nezz that these smaller, atomic tests are a good thing.

tgjones:
So shall I carry on with this? Or should I make separate PRs for the other render states (this one is going to get kind of large, which is fine with me, but the admins might prefer a smaller diff)?

The tests could definitely be sped up if we used `TestFixtureSetUp` instead of `SetUp`, but that's a slippery slope - I like keeping tests as isolated as possible.

KonajuGames:
Because of their size, keep them as separate PRs.​

tomspilman:
> I agree with @tgjones and @Nezz that these smaller, atomic tests are a good thing.

I guess i'm out voted here. :)

It bothers me that of 546 test... 170 are dedicated to blend modes.  We added 6 minutes to the unit test run time for a test that takes under 1 second if it was done as a batch.

It really seems like overkill and for no good reason.

Like look at the unit test for serialization.  We didn't make one XML file per field we were testing.

tomspilman:
> Or should I make separate PRs for the other render states 

Separate PRs... it will just get too big and complex to test.

Especially since all the work is easily separated.

KonajuGames:
Looking at the sample images, are you suggesting the different blend modes
could be rendered onto one screenshot?​  Given the sample texture used,
that shouldn't be difficult.  I didn't think of that until now.  I blame
the lack of sleep. :)

tgjones:
> It really seems like overkill and for no good reason.

I can definitely see that. I'm coming around to your viewpoint :) Mainly because of the time it adds to the test suite.

KonajuGames:
Render them in a grid, one blend state per cell.  You might fit all combinations onto one screenshot?  The diff between images should highlight the failed states.  Perhaps a small text item above each rendered texture to identify the blend state used.

tomspilman:
Not near my work PC, so I can't share a screenshot... but yeah.  The original code I shared rendered all the blend states by drawing a grid of sprites.

And not like either extreme are the only options... one test per blend... or all blends in one test.  We could batch them into 4 or 5 tests if it makes sense.  But 100s of tests seems like it just sets us up for maintenance hassles in the future.

Yes if the test fails it won't say what specific blend failed...  it will just say that some blend mode failed.  But the point of the tests isn't do debug things for us.  It is there to signal that something is broken is all.

tgjones:
> Render them in a grid, one blend state per cell. 

@KonajuGames Yep, that's what @tomspilman's original test code did. It was me that tried to be clever and separate them into separate test cases.

tomspilman:
>  I'm coming around to your viewpoint :) Mainly because of the time it adds to the test suite.

Yeah... I am glad @KonajuGames looked that up.  It is a concern as our hardware isn't free and build times do matter for us.  We can't end up with 2 hours of unit tests to run on each PR.

tgjones:
> Yes if the test fails it won't say what specific blend failed... it will just say that some blend mode failed. But the point of the tests isn't do debug things for us. It is there to signal that something is broken is all.

That's a good point. I'll modify my last commit to do everything in one screenshot. I (now) think that's the right way to go - this was a useful discussion.

tomspilman:
> this was a useful discussion.

I agree... It forced me to think more about it.

Unit testing purists will always say lots of little tests with no overlap in functionality.  They will also preach interface abstractions and mocking to fake objects.  While they are good general practices, it doesn't need to be all or nothing.  The reality of shipping code means finding a balance between the purist/academic view and real world experience.

Nezz:
To get the best of two worlds do the rendering in `[TestFixtureSetUp]` (or `[OneTimeSetUp]` in NUnit 3), set the result as a class variable then every individual test should only test a specific area of the result. This way you would get the speed of batching, but tests will be atomic.

tgjones:
If the tests pass, then as far as I'm concerned this is ready to go. I'll create separate PRs for the equivalent changes to `SamplerState` and `RasterizerState`. (And `Effect.EffectCache`, which was the thing that got me started on all this!)

tgjones:
@tomspilman do you mind checking if the new reference image looks like what you'd expect? I think there was a bug in the gist - `_states[x * y]` should be `_states[(y * width) + x]` - which I've fixed here, but it would be useful to have somebody verify that.

tgjones:
Let me know if there's anything more to do here - as far as I'm concerned, this can be merged.

tomspilman:
Sorry... forgot about this one for a few days there.  I'm going to merge this now so you can move forward with the other states following the same pattern.

tgjones:
Awesome!

PR #3474, opened 2015-01-30
- Application run loop workaround to support mono windows

christallire:
This change will make MonoGame running on Mono-Windows, which will help deploying MonoGame based games without .Net framework.

Reworked for PR #3440 to cleaning up changes


KonajuGames:
Thanks @christallire . Merging.

PR #3475, opened 2015-01-30
- Mono workaround: Uri.LocalPath is not always same with .net platform

christallire:
as with version of mono 3.12.0 
Uri.LocalPath returns "///" or "//" before actual LocalPath.

```
Mono: ///Content/Spirtes/a.xnb
Windows: Content/Spirtes/a.xnb
```

So i did eliminated in order to "///" then "//" to prevent unwanted "/" left.

### DIFF
--- a/MonoGame.Framework/TitleContainer.cs
+++ b/MonoGame.Framework/TitleContainer.cs
@@ -116,7 +116,15 @@ namespace Microsoft.Xna.Framework
         // this same logic is duplicated all over the code base.
         internal static string GetFilename(string name)
         {
-            return FileHelpers.NormalizeFilePathSeparators(new Uri("file:///" + name).LocalPath.Substring(1));
+            var path = new Uri("file:///" + name).LocalPath.Replace("///", "")
+                                                 .Replace("//", "")
+                                                 .Replace(@"\\\", "")
+                                                 .Replace(@"\\", "");
+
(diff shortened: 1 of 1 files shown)


tomspilman:
>  with version of mono 3.12.0 

Is that an older or newer version of Mono?

chris-sung:
It's latest stable release of mono

dellis1972:
I'll test the LocalPath on my mac and see if I get the same issue. It looks like a bug in mono itself.. So we should probably raise a bug over at bugzilla.xamarin.com 

If I do get the same results then this seems like a reasonable work around until its fixed

dellis1972:
This might be a linux bug the following code on a mac (mono 3.12.0) 

var uri  = new Uri ("file:///Users/dean/Desktop/Test.mgcb");
			Console.WriteLine ( uri.LocalPath);
			Console.ReadLine ();

produces 

/Users/dean/Desktop/Test.mgcb

as expected. I'll try this on my linux machine later.

chris-sung:
I did it on a windows.

chris-sung:
```
			var uri = new Uri("file:///Users/dean/Desktop/Test.mgcb");
			Console.WriteLine(uri.LocalPath);
			Console.ReadLine();
```

I did same test on windows.

It produces

```
\\\Users\dean\Desktop\Test.mgcb
```

PR #3476, opened 2015-01-30
- FIX for exception is thrown when device feature is not supported by PresentationParameters.MultiSam...

christallire:
...pleCount

```
multisampleDesc.Quality = (int)SharpDX.Direct3D11.StandardMultisampleQualityLevels.StandardMultisamplePattern;
+                
```

This code throws error when device is either very old or only supporting for software rendering.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -632,6 +632,7 @@ namespace Microsoft.Xna.Framework.Graphics
             {
                 //Find the maximum supported level coming down from 32, 16, 8, 4, 2, 1, 0
                 var maxLevel = 32;
+                var quality = (int) SharpDX.Direct3D11.StandardMultisampleQualityLevels.StandardMultisamplePattern;
                 while (maxLevel > 0)
                 {
                     if (_d3dDevice.CheckMultisampleQualityLevels(Format.R32G32B32A32_Typeless, maxLevel) > 0)
@@ -643,10 +644,11 @@ namespace Microsoft.Xna.Framework.Graphics
                 if (PresentationParameters.MultiSampleCount > maxLevel)
                 {
                     targetLevel = maxLevel;
+                    quality = maxLevel - 1;
                 }
 
                 multisampleDesc.Count = targetLevel;
-                multisampleDesc.Quality = (int)SharpDX.Direct3D11.StandardMultisampleQualityLevels.StandardMultisamplePattern;
+                multisampleDesc.Quality = quality;
             }
 
             // Use BGRA for the swap chain.


tomspilman:
I'm not sure what exactly this is fixing.

@tgjones @KonajuGames - This change make sense to you guys?

tgjones:
Disclaimer: everything I know about this, I just learnt in the last 15 minutes :)

From reading the MSDN docs, it looks like the code was previously not quite right, but this is also not quite the right fix. From the [MSDN docs here](https://msdn.microsoft.com/en-us/library/windows/desktop/bb173072%28v=vs.85%29.aspx) about `SampleDescription.Quality`:

> The image quality level. The higher the quality, the lower the performance. The valid range is between zero and one less than the level returned by ID3D10Device::CheckMultisampleQualityLevels for Direct3D 10 or ID3D11Device::CheckMultisampleQualityLevels for Direct3D 11.
For Direct3D 10.1 and Direct3D 11, you can use two special quality level values. [...]

And [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476218(v=vs.85).aspx) it says:

> An app calls ID3D11Device::CheckMultisampleQualityLevels to get the number of quality levels available during multisampling. A 0 quality level means the hardware does not support multisampling for the particular format. If the number of quality levels is greater than 0 and the hardware supports the fixed sample patterns for the sample count, the app can request the fixed patterns by specifying quality level as either D3D11_STANDARD_MULTISAMPLE_PATTERN or D3D11_CENTER_MULTISAMPLE_PATTERN. The app can call ID3D11Device::CheckFormatSupport to check for support of the standard fixed patterns. [...]

So this change in this PR is not right:

``` csharp
quality = maxLevel - 1;
```

It should subtract 1 from the return value of `CheckMultisampleQualityLevels`, not `maxLevel`. However, that would mean we'd use the maximum quality level, which is a change from the current behaviour, where we just use the "standard" pattern, whatever that is - I think it means it's up to the hardware.

But - the docs make it sound like it's possible for the hardware not to support those standard patterns, in which case I guess you can't use the standard pattern enum values. And that's presumably what this PR is for. However, the docs I quoted above go on to say:

> The runtime defines the following standard sample patterns for 1(trivial), 2, 4, 8, and 16 sample counts. Hardware must support 1, 4, and 8 sample counts. Hardware vendors can expose more sample counts beyond these. However, if vendors support 2, 4(required), 8(required), or 16, they must also support the corresponding standard pattern or center pattern for each of those sample counts.

... which makes it sound like if hardware reports `> 0` for `CheckMultisampleQualityLevels`, then it *must* support the standard patterns.

So, in conclusion, the code we have now looks fine to me :) But the docs, rather unclearly, make it sound some hardware might not support the fixed patterns. But I think we'd want to see some concrete examples of that, with concrete hardware specs, before making this change.

@christallire, have you seen hardware (or reference device, etc.) that requires this change? What prompted this PR?

tgjones:
Incidentally, although the docs say

> The app can call ID3D11Device::CheckFormatSupport to check for support of the standard fixed patterns.

... I haven't been able to find how to do this. None of the `FormatSupport` flags look related.

chris-sung:
I was testing it on various platforms on vmware without 3d acceleration.
In this case it was vanilla windows 7, strangely enough it only supports hardware acceleration but reference, software, warp devices types.

I fixed it according to some sharpdx documents said i should set it to max quality levels - 1
i thought many samples sets those values to (1, 0) so i thought it makes sense.

I'll go find the document mentioned it and it's okay to just close issue if this is wrong.

tgjones:
I think `Quality = 0` would be alright, although that would mean it's up to the driver, rather than specifying one of the standard patterns. The part that doesn't look right to me is `maxLevel - 1`, which would be mean it would be `31`, `15, etc.

chris-sung:
You're right.

http://sharpdx.org/documentation/api/f-sharpdx-dxgi-sampledescription-quality
...And yes I did misinterpreted that document.

It passed test because that hardware device it only supports (1, 0).
But still, 1 sample count and StandardMultisamplePattern just doesn't fit in this test environment.

Thanks for your advice, if you think i still can fix this issue, i'll go for it.

tgjones:
@christallire On what line of code does it crash? When it's actually creating the `SwapChain`?

chris-sung:
@tgjones Yes, throws InvalidArguementException In CreateSwapChain.

tgjones:
I had another idea. @christallire, do you still have access to the VM where the problem occurred? If so, can you try this:

Change this line:

``` csharp
if (_d3dDevice.CheckMultisampleQualityLevels(Format.R32G32B32A32_Typeless, maxLevel) > 0)
```

to this:

``` csharp
if (_d3dDevice.CheckMultisampleQualityLevels(format, maxLevel) > 0)
```

And then move the line that declares and sets the `format` variable (currently line 653) to above that block.

It might help - checking multisample quality levels for one format, and then creating the swap chain with a completely different format, doesn't look right to me.

christallire:
@tgjones i tested it and still get same exception.
The format was B8G8R8A8_UNorm.

tgjones:
Can you post more details about the virtual machine you're testing on? I'll try to find time to recreate it here and do some debugging. You said Windows 7, but x86 / x64? RTM or SP1? Any drivers installed? Is it using VMware's Direct3D drivers, or just the ones installed with Windows? etc.

christallire:
Windows 7 SP1 Vanilla, No drivers/directx manually installed, except VMware tools.
I can see accelerate 3D graphics option enabled in Virtual Machine Settings -> Hardware -> Display.

tomspilman:
Anything new here?

In general we don't spend a lot of time making MonoGame run on virtual machines because of how error prone that usually is.  However if a simple fix that doesn't hurt the normal uses of MonoGame is possible... we will merge it.

tgjones:
I meant to spin up a VM to investigate this, because there definitely is an issue with MG here, but I haven't done that yet.

This PR as it stands isn't the right fix - but I've spent a fair amount amount of time trawling MSDN and I think it might actually be missing some documentation about how to check for support of the standard multisample pattern. So I don't know what the *right* fix is.

tgjones:
Based on @christallire's findings, one rather hacky solution would be to catch the exception thrown when creating the swap chain, and try again with quality set to 0.

PR #3477, opened 2015-01-30
- FIX: for OpenALSoundController -- collection is modified when multiple...

christallire:
... thread is accessing it

Although I think this is lazy way to manage sound sources via static purgeMe member variables,
I just followed the way it uses resources just written above.


tomspilman:
@christallire - Looks like maybe your OpenALException branch is a few weeks old.  If you merge/rebase with the latest develop this PR should build correctly.

chris-sung:
Oops, my bad. I'll rebase it

PR #3479, opened 2015-01-31
Improves BoundingBox/Plane intersection performance

prollin:
Inlining Vector3.Dot calls inside BoundingBox.Intersects results in significant performance improvement , especially when performing frustum culling on mobile devices.

I thought about using [`[MethodImpl(MethodImplOptions.AggressiveInlining)]`](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.methodimploptions(v=VS.110).aspx) attribute but it would require to switch MonoGame to use .NET4.5 and it would not guaranty that it would be inlined anyway. I believe it is safer to simply inline/unroll MonoGame core math algorithms to ensure the best possible performance on all platforms.

### DIFF
--- a/MonoGame.Framework/BoundingBox.cs
+++ b/MonoGame.Framework/BoundingBox.cs
@@ -472,14 +472,14 @@ namespace Microsoft.Xna.Framework
                 negativeVertex.Z = Max.Z;
             }
 
-            var distance = Vector3.Dot(plane.Normal, negativeVertex) + plane.D;
+            var distance = plane.Normal.X * negativeVertex.X + plane.Normal.Y * negativeVertex.Y + plane.Normal.Z * negativeVertex.Z + plane.D;
             if (distance > 0)
             {
                 result = PlaneIntersectionType.Front;
                 return;
             }
 
-            distance = Vector3.Dot(plane.Normal, positiveVertex) + plane.D;
+            distance = plane.Normal.X * positiveVertex.X + plane.Normal.Y * positiveVertex.Y + plane.Normal.Z * positiveVertex.Z + plane.D;
             if (distance < 0)
             {
                 result = PlaneIntersectionType.Back;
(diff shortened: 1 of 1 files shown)


mgarstenauer:
When we optimize/inline code in our projects, we leave the original code as a comment. Example:

    // Original:
    // var distance = Vector3.Dot(plane.Normal, negativeVertex) + plane.D;
    
    // Inlined:
    var distance = plane.Normal.X * negativeVertex.X + plane.Normal.Y * negativeVertex.Y + plane.Normal.Z * negativeVertex.Z + plane.D;

This helps readability, and once Vector3 supports SIMD (e.g. System.Numerics) we can easily switch back to the original implementation.

prollin:
@mgarstenauer I would say that this kind of sensitive code need to always be inlined; even with SIMD there is still a cost associated with calling a method (even if it is static). When MonoGame supports System.Numerics, we will still need to inline that code in place IMHO.

Adding a comment to describe what the code is essentially doing in a human readable way is definitely a good idea though :)

MonoGame.Framework/BoundingBox.cs
```diff
@@ -472,14 +472,14 @@ public void Intersects(ref Plane plane, out PlaneIntersectionType result)
                 negativeVertex.Z = Max.Z;
             }
 
-            var distance = Vector3.Dot(plane.Normal, negativeVertex) + plane.D;
+            var distance = plane.Normal.X * negativeVertex.X + plane.Normal.Y * negativeVertex.Y + plane.Normal.Z * negativeVertex.Z + plane.D;
```
tomspilman: Just add a comment above this...

```
// Inline Vector3.Dot(plane.Normal, negativeVertex) + plane.D;
```

tomspilman:
Great... thanks @prollin !

PR #3487, opened 2015-02-02
Throw helpful exception when vertex format doesn't match vertex declaration

### LINKED ISSUE #3235: Better error when VertexFormat doesn't match shader
During MVA demo had a cased where a models vertex format mismatched with the shader input format.

The error you get is below. @tomspilman says he can make that better.

An unhandled exception of type 'SharpDX.SharpDXException' occurred in SharpDX.dll

Additional information: HRESULT: [0x80070057], Module: [General], ApiCode: [E_INVALIDARG/Invalid Arguments], Message: The parameter is incorrect.

tgjones:
Fixes #3235.

As discussed in #3235, we don't have access to the input elements from the vertex shader, so the best we can do is tell the user the elements in the current vertex declaration.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Diagnostics;
+using System.Linq;
 
 using SharpDX;
 using SharpDX.Direct3D;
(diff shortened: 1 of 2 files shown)


tomspilman:
Looks good to me... and a unit test!

tgjones:
Following your lead, I'm firmly subscribed to the theory that MonoGame can only support the number of platforms that it does, with the help of a large suite of unit tests :) So I'm doing my (little) bit, and adding some tests with each PR.

tomspilman:
Absolutely....  and if everyone added even a little to the unit tests on each PR we will be in really good shape within a year.

tomspilman:
Merging!

PR #3489, opened 2015-02-02
Pass EffectProcessor Defines through to shader compiler.

### LINKED ISSUE #3096: fx technique preprocessor defines
Looks like whatever custom parsing we do to support the .fx syntax for defining a technique / pass / etc is not setup to obey one or more of the following: #if, #elif, #endif. Alternately the problem could be that defines from the processor are not being passed down to this step.

tgjones:
Fixes #3096.


tomspilman:
Looks great to me!  Thanks for the fix/feature!

tomspilman:
Merged!

Issue #1995, opened 2013-09-15
sRGB Gamma Support


tomspilman:
We would like to see gamma correct rendering support added to MonoGame.  Currently we do this entirely in the shaders...  doing the correct conversion to linear space and back to gamma as appropriate.  The problem is that this wastes shader instructions.

The better way to do this in DX11/GL is to create the textures and render targets using the _SRGB formats.  In this mode the GPU will automatically convert to linear and back to gamma as needed entirely in hardware and cheaper than using shader instructions.  It also allows the hardware to do gamma correct filtering.

To add this to MonoGame we would have to:
- Add the _SRGB formats to the PixelFormat enum.
- Fix the Texture/RenderTarget classes to create SRGB formats.
- Fix ContentManager.Load<> To Support Loading sRGB formats.

Thoughts?

tgjones:
This (and #1996) looks like an interesting issue to revive. I admit selfish reasons - I haven't yet dealt with gamma correction in my engine, and if I can do using built-in techniques, without custom shader support, that will be easier! So, I'd like to help...

> I'll take a stab at adding support this week.

@tomspilman did you get anywhere with that? #1996 seems like the more controverial change; once that's in place, the other changes in `SurfaceFormat` and the Texture / RenderTarget classes seem easier.

tomspilman:
> This (and #1996) looks like an interesting issue to revive

I agree.


> I admit selfish reasons 

I think most of our PRs are features or fixes added because some developer needed it.  So selfish reasons are the norm around here.

As long as it is a good PR it doesn't matter.

> did you get anywhere with that?

I don't remember and I can't find a branch for it right now.  Really anything I had is so old right now it probably won't be useful.

None of these changes are hard really.

tomspilman:
FYI on OpenGL support from Wikipedia:

> OpenGL supports the textures with sRGB gamma encoded color components (first introduced with EXT_texture_sRGB extension, added to the core in OpenGL 2.1) and rendering into sRGB gamma encoded framebuffers (first introduced with EXT_framebuffer_sRGB extension, added to the core in OpenGL 3.0).

http://en.wikipedia.org/wiki/SRGB

So there are options to support it there.

tgjones:
[...]

Issue #1996, opened 2013-09-15
ContentManager Loading Extension


tomspilman:
I have a proposal for a new extension to the `ContentManager`:

```
public abstract class ContentOptions
{
};

public class ContentManager
{
   public virtual T Load<T>(string assetName, ContentOptions options)
}
```

This is not for working around the content pipeline.  We wouldn't use this or recommend it be used to do runtime processing of content.  This is simply for passing options that are simple runtime choices down thru the ContentManager and down to the various content readers.

For example it could be used for textures:

```
var options = new TextureOptions();
options.sRGBFormat = true;
options.PreserveSystemCopy = true;
var texture = contentManager.Load<Texture2D>("myTexture",options);
```

sRGB support is simply creating the Texture2D using an sRGB format...  no extra processing is done.  While you could add support into a custom texture processor...  it is sort of overkill for something that is a runtime choice.

The issues with supporting Texture2D.GetData() on some platforms is easily solved by keeping around the system memory copy of the texture data.  By default the ContentManager throws that away after creating the GPU texture resource.  If for example we kept these around we could use them to implement a fast GetData() on textures.

PR #3492, opened 2015-02-03
Implement sRGB gamma support

tgjones:
Implements (or will, when it's finished) #1995.

So far, this only works for DirectX. I'll add OpenGL support in a later commit, once things are working properly for DirectX.

This initial commit is just the runtime support. We need to figure out how to tell `Texture*Reader` to set the relevant sRGB `SurfaceFormat` when loading textures.


MonoGame.Framework/Graphics/SurfaceFormat.cs
```diff
@@ -104,6 +104,32 @@ public enum SurfaceFormat
...
+        Dxt1SRgb = 33,
+        /// <summary>
+        /// DXT3. sRGB texture format with compression. Surface dimensions must be a multiple of 4.
+        /// </summary>
+        Dxt3SRgb = 34,
+        /// <summary>
+        /// DXT5. sRGB texture format with compression. Surface dimensions must be a multiple of 4.
+        /// </summary>
+        Dxt5SRgb = 35,
+
```
tomspilman: This is starting to feel like an explosion of formats with `SRgb` appended.

Can just about any format be SRGB?  Maybe it needs to be some separate flag?

tgjones:
I've read through the discussion in #1996. Although a more extensible loading mechanism sounds good in general, I wasn't sure if it's really necessary for sRGB support.

As I understand it, for any given texture, you'd want to always - or never - use the sRGB format. (For most mesh textures, you'd use sRGB; for any generated stuff like heightmaps or normal maps, you wouldn't).

If that's true (is it?), then I don't see the harm in setting an IsSRgbTexture property on the texture processor. Just my $0.02 to get the discussion going.

MonoGame.Framework/Graphics/SurfaceFormat.cs
tgjones: No, at least according to SharpDX - these are all but one of the SRgb formats that the SharpDX `Format` enum exposes. (The only other one is `BC7_UNorm_SRgb`, but we don't support the non-SRgb version of that.)

KonajuGames: How many more would be added when OpenGL is added to the mix?​

tomspilman: Ok... totally do not have my brain in SRGB thinking right now.  Yes it is only supported on a subset of texture formats.

tgjones: I *think* there's a similar story with OpenGL - as in only a few formats support sRGB, and perhaps not the same subset as Direct3D - but I haven't checked in detail.

tgjones: > How many more would be added when OpenGL is added to the mix?​

From the little reading about it that I've done so far... I don't know :) There are a confusing array of OpenGL extensions covering sRGB support, and I don't know which ones we will use.

KonajuGames: > There are a confusing array of OpenGL extensions

So it's business as usual then. :)

tgjones: I think so, although I should say I'm not at all an OpenGL expert, so I might be making it sound worse than it is. (I'm comfortable-ish with basic OpenGL usage, but I don't think I've ever used an extension before.)

tomspilman:
> Although a more extensible loading mechanism sounds good in general, I wasn't sure if it's really necessary for sRGB support.

Yeah... this is a good separation.  We can leave the more controversial changes for content loading for another discussion.

> As I understand it, for any given texture, you'd want to always - or never - use the sRGB format.

Yes.  It depends on what the data is in the texture.

First using sRGB isn't required... for most games it won't be important.  However if you are doing complex lighting models and/or HDR you want to be sure things are gamma correct.  This is where sRGB comes in.

If it is like a diffuse texture for texturing a mesh or for a sprite or something....  you would want sRGB so that the GPU will convert from gamma to linear when sampling the texture in the pixel shader.  You also want your render target you are rendering your diffuse data to be sRGB if the results are displayed back to the user.

Things like normal maps don't need sRGB.  The data in them is mathematical and not visual... there is no gamma curve there.  So those don't get an sRGB format and the GPU will not do any conversion when sampling them in the pixel shader.

> I don't see the harm in setting an IsSRgbTexture property on the texture processor.

It would be worth discussing as part of the content support part of things.

MonoGame.Framework/Graphics/SurfaceFormat.cs
tomspilman: SRGB support is a hardware feature, so OGL should match up pretty identically.

The complexity becomes when some GPU hardware doesn't support particular sRGB formats.

KonajuGames:
> > I don't see the harm in setting an IsSRgbTexture property on the texture processor.
> It would be worth discussing as part of the content support part of things.

In the TextureProcessor, we don't select a specific format, but rather Color, DxtCompressed, Compressed (which will select the best compression for the target platform), PvrCompressed or AtiCompressed (future support).  Here an extra property `IsSRgb` would be better than doubling the size of the `TextureProcessorCompressionFormat` enum with an extra Srgb member for each value.

tomspilman:
> Here an extra property  IsSRgb  would be better than doubling the size of the  TextureProcessorCompressionFormat  enum 

For certain that would be the path.

KonajuGames:
> For certain that would be the path.

But for the `TextureFormat` enum, I'd leave it as they are in this PR.

tgjones:
I've added support for sRGB in the content pipeline - basically just `TextureProcessor.IsSRgb`, and then applying that in `Texture2DWriter` and `TextureCubeWriter`.

Let me know if you have better ideas for how it should be done. This way seemed like it touched the least amount of code.

tomspilman:
> basically just `TextureProcessor.IsSRgb`

I'm not sure if we want to use the terminology `IsSRgb` here.

First sRGB is more of a programmer's term than one aimed at content creators.  Assuming that is what we want to do I'm thinking either it should be `UseSRgb` or `MakeSRgb` or `UseSRgbFormat` which more correctly describes what the setting is doing.

Note: Unreal4 has an `sRGB` checkbox on textures, but that term could be still use for legacy reasons.

The term most photographers and artists know is "gamma".  If we go that way then we probably want to `IsGammaCorrect` or `IsGammaCorrected` which defines that the source texture contains gamma corrected data.

> and then applying that in `Texture2DWriter` and `TextureCubeWriter`.

Well the option needs to appear in [MaterialProcessor](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/MaterialProcessor.cs) and [ModelProcessor](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/ModelProcessor.cs) as well.  Both of those processors internally process textures.  Really... any processor that includes `TextureProcessorOutputFormat` probably also requires this sRGB/Gamma property as well.  In the case where the processor is working on specular or normal map textures it can never use an sRGB format as those are generally not gamma corrected textures.

Also brings up questions about `FontDescriptionProcessor` and `FontTextureProcessor`.  Both internally process a texture for fonts...  i'm pretty sure that texture is gamma correct.  So this means we need to option to enable gamma support in the texture format.

So this is all more complex...  which is why we discussed keeping it for a separate PR. ;)

MonoGame.Framework.Content.Pipeline/Serialization/Compiler/TextureCubeWriter.cs
```diff
@@ -20,6 +20,8 @@ protected internal override void Write(ContentWriter output, TextureCubeContent
             if (!level0.TryGetFormat(out format))
                 throw new Exception("Couldn't get format for TextureCubeContent.");
 
+            format = value.GetActualFormat(format);
```
tomspilman: What happens if the platform doesn't support sRGB like PSM or OpenGL at the moment?  We want the pipeline to fail during texture creation not the runtime.

tgjones:
> which is why we discussed keeping it for a separate PR. ;)

Oops, I completely missed / misunderstand that discussion ;-) In that case, I'll keep this PR just for the runtime changes, and then make a separate PR for the content pipeline changes.

Shall I add OpenGL runtime support as part of this PR, or separate that out too?

> I'm not sure if we want to use the terminology `IsSRgb` here. ... The term most photographers and artists know is "gamma".

I'll defer to wiser heads; I really don't know much about sRGB. But it would be nice to make it clear that there's a connection between setting the sRGB property on textures at content build time, and creating the back buffer with an sRGB format at runtime. Using something that mentions "sRGB", rather than gamma, would help with that.

> I'm thinking either it should be `UseSRgb` or `MakeSRgb` or `UseSRgbFormat`

I was thinking that what you're saying with `IsSRgb` is "this texture **is already** sRGB-encoded". In my mind, "make" (and to a lesser extent "use") implies that the content pipeline is going to do some kind of conversion for you, which isn't the case.

(Just points for discussion, I'll be happy to change it!)

MonoGame.Framework.Content.Pipeline/Serialization/Compiler/TextureCubeWriter.cs
tgjones: Good point - if we know for sure that sRGB isn't supported by the target platform, we should fail at content build time.

But we'll still have to support the case where textures are built with sRGB=true, but it isn't supported at runtime (i.e. once we implement it for OpenGL, not all graphics cards will necessarily support the required extensions). In that case, we'll have to fallback to using a non-sRGB texture format.

KonajuGames: Is the texture processing different if the texture is flagged as sRGB?  Is
there a difference in the pixel data?  Or is just a different texture
format value?​

tomspilman: > Is the texture processing different if the texture is flagged as sRGB? 

No it is not... the texture data is 100% the same.  Pretty much all game textures are gamma corrected and in the non-linear sRGB format.  This is just letting the GPU hardware know this fact so that it can correctly decode it to a linear format before processing.

This is why I originally was looking for a solution that didn't involve adding this to the pipeline.  It doesn't change the pipeline at all really...  it is more of an issue on how the texture is used in your game.

tgjones: > This is why I originally was looking for a solution that didn't involve adding this to the pipeline.

Yep, and having implemented [a version of] the content pipeline changes, I'm more inclined to that way of thinking too. But I keep coming back to: for any given texture, it either is or it isn't sRGB-encoded, so it makes sense to specify it once at content build time, and let the runtime figure out whether it's supported by the hardware.

tgjones: On a related note, if it's possible for an OpenGL driver to support framebuffer sRGB but not texture sRGB - *do any OpenGL experts here know if that can happen?* - then that's awkward. I don't know if we just disable sRGB totally if either type isn't available. Otherwise MG users will have to do some gruntwork to make sure they're creating textures and framebuffers in a consistent colour space, which sounds a bit lower-level than XNA / MG usually aims for.

tomspilman: > if it's possible for an OpenGL driver to support framebuffer sRGB but not texture sRGB 

I'm not an expert... but I highly doubt it.  If the sampler has the ability to decode gamma, I would think it would be able to encode gamma.

tgjones:
I've removed the content pipeline changes from this PR, and added OpenGL support. I think it makes sense to do all the runtime changes in one PR, because a lot of it is shared across platforms.

With these changes, the sRGB tests in `Texture2DTest` now pass for both Direct3D and OpenGL.

I'd feel more confident about this once I know the answers to my questions [here](https://github.com/mono/MonoGame/issues/1995#issuecomment-74408429). @tomspilman, do we have a resident OpenGL expert? I'm working a bit blind here - I'm comfortable with Direct3D, but not OpenGL.

tgjones:
Finally! I should probably get myself setup for local Android and iOS compilation to avoid relying on the build server to find syntax errors.

tomspilman:
> I should probably get myself setup for local Android and iOS compilation

Really it is sort of impossible to require contributors to have access to all platforms at all times.  I'm totally fine with people using the build server and unit tests to verify other platforms work. ;)

tomspilman:
> I've removed the content pipeline changes from this PR, and added OpenGL support. 

That makes sense... we can work out how best to add it to the pipeline separately.


> I'd feel more confident about this 

Well as long as your changes didn't break existing rendering features then it should be safe.  If someone reports a bug later with sRGB support it should be easy to address.

tomspilman:
So what is the status of this PR?  Are we ready to merge?  Or are we waiting on something?

tgjones:
It's done - it's possible that I've missed something (there are quite a few places with `switch` statements for the texture formats) but as you said as long as I don't break existing functionality, it's safe enough.

The check for sRGB format in OpenGL (full and ES) has been discussed in the other thread, but so far no-one has explicitly said "looks good", so that's the part I'm most unsure of.

prollin:
@tgjones should an exception be thrown when the sRGB format (or actually any format) isn't supported instead of falling back to another format (see goto statements)?

tgjones:
@prollin Good question.

It partly depends on what we decide on for the content pipeline part of this. If we're embedding the sRGB `SurfaceFormat` into compiled textures, then we *do* need to fallback to non-sRGB formats, because we have no idea what will actually be available at runtime, and we can't throw an exception in this case.

And it depends on how many knobs and dials we want to expose to developers:

* Attempt to make it automatic - if sRGB isn't available, fallback to non-sRGB, for both backbuffer and textures. Obviously the actual output on the screen would be different if the fallback is used.
* Add a public `SupportsSRgb` property somewhere, and leave it up to developers to create the backbuffer and textures in the right format. They'd also have the chance to use different shaders if sRGB isn't available on the current hardware. But XNA 4 / MG moved away from this kind of granular capabilities checking, so I'm not sure about this. @tomspilman?

tgjones:
@tomspilman @prollin I'm tempted to suggest that this PR is merged as-is, unless anyone has any objections. We have another discussion in front of us for the content pipeline part of these changes, and I'd like to move on to that.

tomspilman:
Ok...  i'll review this once again in the morning then merge if nothing jumps out at me.

tomspilman:
@tgjones - Can you fix the conflict in this one?

tgjones:
@tomspilman Done.

tomspilman:
Merged!

PR #3499, opened 2015-02-06
WP8 Texture2D.FromStream ecception fix

Nezz:
Catch disposed stream exceptions in WP8 Texture2D.FromStream. It cannot be caught elsewhere because it runs on a different thread.


danzel:
Could we record the exception object and re-throw it outside of the BlockOnUIThread call?
Looks like if it fails currently you'll get a null reference on the next line when trying to use bitmap, which won't be very useful to the user.

Nezz:
To be honest that happens all the time because it also occus when the BlockOnUIThread times out. If you want, feel free to make the change :)

PR #3503, opened 2015-02-07
Content pipeline fixes - getter-only properties and polymorphic generic types

### LINKED ISSUE #3404: Importing XML through the MGCB processor - errors

tgjones:
* Added `ContentTypeReader.CanDeserializeIntoExistingObject`, to match `ContentTypeWriter` and `ContentTypeSerializer` properties of the same name
* Fixed a bug in `ContentCompiler.GetTypeWriter` if it was called multiple times for different closed generic types based on the same open generic type
* Added support for polymorphic generic types in `IntermediateSerializer` - i.e. storing a `List<string>` in property of type `object`
* Implemented "proper" check in `ReflectiveSerializer` for whether we can deserialize into an existing object, instead of basing it on being one of the `System.*` types - fixes 2nd issue in #3404 (note that this was already done properly in `ReflectiveWriter`, and in fact there would be some value in sharing some of the code to check whether properties can be serialized)
* Used `property.GetIndexParameters()` in `ReflectiveReader` instead of the more complicated previous check
* Added tests, or improved existing tests, for most of the above

These issues are fairly closely related, so hopefully it's okay bundling them into one PR. (Actually, the fix for the first problem uncovered another bug, and the fix for that uncovered a subsequent bug, etc.)

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
@@ -60,10 +60,17 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
                     if (prop.GetGetMethod() == null)
                         return false;
 
-                    // If there is no public setter, and we don't have a type serializer 
+                    // If there is a setter, but it's private, then don't include this element
+                    // (although technically we could, as long as we have a serializer with
+                    // CanDeserializeIntoExistingObject=true for this property type)
+                    var setter = prop.GetSetMethod(true);
+                    if (setter != null && !setter.IsPublic)
+                        return false;
+
+                    // If there is no setter, and we don't have a type serializer 
                     // that can deserialize into an existing object, then we have no way 
                     // for it to be deserialized.
-                    if (prop.GetSetMethod() == null && !serializer.GetTypeSerializer(prop.PropertyType).CanDeserializeIntoExistingObject)
+                    if (setter == null && !serializer.GetTypeSerializer(prop.PropertyType).CanDeserializeIntoExistingObject)
                         return false;
 
                     // Don't serialize or deserialize indexers.
@@ -125,6 +132,11 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
                 if (GetElementInfo(serializer, field, out info))
                     _elements.Add(info);                
             }
+        }
+
+        public override bool CanDeserializeIntoExistingObject
+        {
+            get { return TargetType.IsClass; }
         }
 
         protected internal override object Deserialize(IntermediateReader input, ContentSerializerAttribute format, object existingInstance)
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
@@ -3,7 +3,6 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
-using System.CodeDom;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -136,7 +135,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
 
         public override bool CanDeserializeIntoExistingObject
         {
-            get { return TargetType.IsClass; }
+            get { return TargetType.IsClass && TargetType.BaseType != null; }
         }
 
         protected internal override object Deserialize(IntermediateReader input, ContentSerializerAttribute format, object existingInstance)
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
@@ -60,10 +60,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
                     if (prop.GetGetMethod() == null)
                         return false;
 
-                    // If there is no public setter and the property is a system
-                    // type then we have no way for it to be deserialized.
-                    if (prop.GetSetMethod() == null &&
-                        prop.PropertyType.Namespace == "System")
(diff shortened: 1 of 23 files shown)


tgjones:
Looks like `Type.IsClass` isn't available on WinRT - I'll have a look.

tomspilman:
> in fact there would be some value in sharing some of the code to check whether properties can be serialized

Yeah...  it seems like some form of `MemberSerializationEnumerator` or something like that could be useful.  It would return all the types in the correct order for serialization and deserialization.

>  hopefully it's okay bundling them into one PR.

 As long as all the changes are closely related... like all about serialization...  it is cool.

MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
```diff
@@ -127,6 +134,11 @@ protected internal override void Initialize(IntermediateSerializer serializer)
             }
         }
 
+        public override bool CanDeserializeIntoExistingObject
+        {
+            get { return TargetType.IsClass; }
```
tomspilman: I think this actually is more complex...  like I suspect `Array` types can't support it.  Don't know if we have a test that covers that case.

Test/ContentPipeline/AssetTestClasses.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Xna.Framework.Content.Pipeline.Graphics;
 using Microsoft.Xna.Framework.Graphics;
 using System.Collections.Generic;
+using System.Runtime.Remoting.Messaging;
```
tomspilman: I suspect this namespace doesn't need to be there.

MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
tgjones: You're right, `Array` types don't support it, but they're handled separately by `ArraySerializer`. I think this is correct for `ReflectiveSerializer`.

tgjones: Oops - actually you're right :) For "polymorphic" properties, we *do* need to check if it's an array, and handle that differently.

tgjones: Hmm, I think I've uncovered an edge-case-bug in XNA - it will happily serialize getter-only properties if the type of the property is `IList`, but will throw an exception when you try to deserialize (that's correct, you shouldn't be able to deserialize into a getter-only array property, whatever the property type may be declared as).

I think we could be a bit cleverer, and not serialize those properties in the first place. That would be a change from XNA behaviour, but this is really an edge case, and I don't think the XNA behaviour is correct. To do that, we'll have to do two checks - once when we're deciding which elements to serialize in `ReflectiveSerializer.Initialize`, and once in `ReflectiveSerialize.Serialize`, when we know the actual type of the object.

tgjones: Okay, I think I'm getting myself mixed up now. We should probably just copy XNA's behaviour, even though it's not that helpful. Otherwise we'll find ourselves having to check the actual type returned from every getter-only property when deserializing, to see whether the XML element is required or not.

tgjones:
So... XNA's behaviour when serializing polymorphic array types is a bit weird, but I've tried to match it, with a corresponding test.

MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
```diff
@@ -127,6 +133,11 @@ protected internal override void Initialize(IntermediateSerializer serializer)
             }
         }
 
+        public override bool CanDeserializeIntoExistingObject
+        {
+            get { return TargetType.IsClass && TargetType.BaseType != null; }
```
tgjones: This is a bit of a guess - it makes all our tests pass, but we might just be missing test cases.

tgjones:
It looks like something's up with the Mac build agent:

> [Step 1/1] /Applications/buildAgent/work/f26e5a04a66fcbdb/default.build(110,8):
External Program Failed: /Applications/Xamarin Studio.app/Contents/MacOS/mdtool (return code was 1)
[01:27:39][Step 1/1] Step Running NAnt Script (NAnt) failed

tomspilman:
> It looks like something's up with the Mac build agent:

That is normal...  it means something in the build failed.  You have to look at the build log for details:

> MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/IntermediateSerializer.cs(167,34): error CS0246: The type or namespace name `NonGenericIListSerializer' could not be found. Are you missing an assembly reference?

MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
tomspilman: We can only work with the test cases we have.  If we find another case that fails in the future we'll just improve the test and add another fix.

tgjones:
Oops - I was looking in the wrong place.

Hopefully it will build okay this time.

tomspilman:
Ok... merging this.  Thanks for all your help here @tgjones !  These serializers are so much more complex than we thought to begin with.

tgjones:
Cool, thanks for merging!

> These serializers are so much more complex than we thought to begin with.

Indeed - and it's hard / basically impossible to find these edge cases until somebody reports a bug. No doubt there are a few more edge cases waiting to be found.

PR #3504, opened 2015-02-07
Updated protobuild to disable "Generate overflow checks" on iOS

### LINKED ISSUE #3438: [iOS] Disable "Generate overflow checks"
After upgrading to the unified API we started getting overflow exceptions in MonoGame here:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentReaders/Texture2DReader.cs#L195

Here ushort is cast to byte. Disabling "Generate overflow checks" resolves the issue.

We need to stop adding `<CheckForOverflowUnderflow>True</CheckForOverflowUnderflow>` to the debug build configurations.
@hach-que Can you remove it?
https://github.com/hach-que/Protobuild/blob/fea0c15222535b3dac5a536f3d8dbd04bc5d6179/Protobuild.Internal/BuildResources/GenerateProject.CSharp.xslt#L496

`<AllowUnsafeBlocks>True</AllowUnsafeBlocks>` can be removed too because the default value is true and it is a project level toggle, not a configuration level one.

proj

Nezz:
Fixes #3438


tomspilman:
Thanks... merging!

PR #3510, opened 2015-02-10
Fixes redundant OpenGL calls

prollin:
Fixes a bunch of redundant OpenGL calls. There are more to be fixed but this is a good start that shouldn't get in the way of other PR.


tomspilman:
So two concerns here.

This technique assumes that no other lib/service will sneak in a change to the GL state on us.  Now we haven't run into this problem yet, but i don't trust Android to not show this problem some day.  The protection could be a forced state reset before each frame.

Second we can't hold device state in statics.  This is the issue we're starting to resolve with #

tomspilman:
Sorry... mistap on my phone. :(

MonoGame.Framework/Graphics/States/BlendState.OpenGL.cs
```diff
@@ -14,43 +14,78 @@ namespace Microsoft.Xna.Framework.Graphics
 {
     public partial class BlendState
     {
-        internal void PlatformApplyState(GraphicsDevice device)
+        private static bool _currentBlendEnable = false;
+        private static BlendState _currentBlendState = new BlendState();
```
tomspilman: @tgjones - So considering your work in https://github.com/mono/MonoGame/pull/3472, how do you think something like this should be done?  I suspect state like this just needs to live in `GraphicsDevice`.

tgjones: Yes, I agree - it should be an internal field / property in `GraphicsDevice`. But since this is specific to OpenGL, probably it should live in `GraphicsDevice.OpenGL.cs`.

(And I might suggest renaming it, since `GraphicsDevice` already has a "current" blend state, even if that's not the actual name of the variable. Perhaps "_deviceBlendState" or something of that nature?)

prollin:
> The protection could be a forced state reset before each frame.

That would be inefficient and defeat the purpose. I believe a better option in this case is to offer an easy way to manually force reseting the states; This PR does open such a possibility with the `force` parameter for each `PlatformApply()` method.

> Second we can't hold device state in statics

Fixed in last commit, let me know if there are other issues

tomspilman:
@prollin - Can you merge in the latest changes from develop?  Just want to be sure your work is compatible with the work @tgjones did.

prollin:
@tomspilman done :)

tomspilman:
Cool...  i'll feel so much better about merging stuff like this as soon as we have the OGL unit tests running.  Thanks @prollin !

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
```diff
@@ -227,6 +236,11 @@ private void PlatformInitialize()
                     "MonoGame requires either ARB_framebuffer_object or EXT_framebuffer_object." +
                     "Try updating your graphics drivers.");
             }
+
+            // Force reseting states
+            this.BlendState.PlatformApplyState(this, true);
```
tgjones: I think it would be better to change these to `_actualBlendState`, etc. That will use the `GraphicsDevice`-specific instances of static state objects.

I can see why it works, as it is now - because it happens that we don't create any device-specific objects in `BlendState.OpenGL.cs`. But that might change in the future.

PR #3511, opened 2015-02-10
Create default RasterizerState objects (CullClockwise, CullCounterClockwise, None) per GraphicsDevice

tgjones:
Continues from #3472, which did the same thing for default `BlendState` objects.

I've added some tests for most properties, except for `DepthBias` and `SlopeScaleDepthBias`, which are a little hard to test. Open to ideas though!

I know that we decided to use a single test for all the `BlendState` visual tests, but I think it makes sense to split these tests out - there are only 7 of them, and they're not as directly related to each other as the `BlendState` tests were.


tomspilman:
This looks great to me.  @KonajuGames ?

KonajuGames:
Looks good to me.  Nice work, @tgjones.

PR #3513, opened 2015-02-10
Fixed key combo incorrect behavior

Shqrdx:
I finally found the fix for https://github.com/mono/MonoGame/issues/3501. The users of MonoGame are able to write their own key combo instead of fight with predefined one. Without this fix DXGI will intercept some key pressing(Alt+enter or printscreen for example) and do something uncontrollable for developers.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -719,7 +719,7 @@ namespace Microsoft.Xna.Framework.Graphics
                 using (var dxgiFactory = dxgiAdapter.GetParent<SharpDX.DXGI.Factory1>())
                 {
                     _swapChain = new SwapChain(dxgiFactory, dxgiDevice, desc);
-
+                    dxgiFactory.MakeWindowAssociation(PresentationParameters.DeviceWindowHandle, WindowAssociationFlags.IgnoreAll);
                     // If VSync is disabled, Ensure that DXGI does not queue more than one frame at a time. This 
                     // both reduces latency and ensures that the application will only render 
                     // after each VSync, minimizing power consumption.


MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -719,7 +719,7 @@ internal void CreateSizeDependentResources(bool useFullscreenParameter = false)
                 using (var dxgiFactory = dxgiAdapter.GetParent<SharpDX.DXGI.Factory1>())
                 {
                     _swapChain = new SwapChain(dxgiFactory, dxgiDevice, desc);
-
+                    dxgiFactory.MakeWindowAssociation(PresentationParameters.DeviceWindowHandle, WindowAssociationFlags.IgnoreAll);
```
tomspilman: So this disables `Alt+Enter` only it seems from the docs:

 https://msdn.microsoft.com/en-us/library/windows/desktop/bb174540(v=vs.85).aspx

Which I agree... we don't want the DX system controlling this, we want the GameWindow managing this itself.

@KonajuGames  @tgjones ???

KonajuGames: I would agree with that.  I was wondering where the `Alt+Enter` key combo was being handled.

tomspilman:
Anyone have some time to test this change before we merge?

Shqrdx:
It enables manual Alt + Enter processing for KeyboardState in Update(before DXGI intercept it and force fullscreen toggling) and does not damage screen resolution switches. It does not effect to Windows8 platform.

Shqrdx:
I guess detailing testing is not required because this method called only once, when D3D/swapchain  is created and it just an adjustment to a property of DXGI.

Shqrdx:
Ok, how to test it - create simple windows project from debug, run it press alt + enter combination and see the result. Joking )

tomspilman:
I had to check the a default XNA game...  yeah...  Alt+Enter doesn't work there either.

So this just makes it work like XNA does.

I'm good with it if @KonajuGames is good with it.

KonajuGames:
I'm good with it.  Merging.

PR #3516, opened 2015-02-11
Create default DepthStencilState objects (Default, DepthRead, None) per GraphicsDevice

tgjones:
Continues from #3472 and #3511, which did the same thing for `BlendState` and `RasterizerState`, respectively.

Only one visual test this time. We'd have to write custom shaders to do proper tests of the other `DepthStencilState` properties, and there isn't yet (AFAIK) a convenient method to compile and load a custom effect as part of a visual test.


tomspilman:
> there isn't yet (AFAIK) a convenient method to compile and load a custom effect as part of a visual test.

Yes there isn't...  but also should there be?  Like is compiling an effect to run a test that has nothing to do with compiling an effect a good idea?

I mean I do like the idea of getting a little more complete system testing by doing that.  But the traditional "unit tests should be small and focused" part of me wonders if it is bad in the long run.

tgjones:
> But the traditional "unit tests should be small and focused" part of me wonders if it is bad in the long run.

I do agree, in principle (although these visual tests aren't really "unit" tests at all). As long as we have tests specifically for compiling effects, which will fail more atomically if something goes wrong, then I don't see a problem with compiling effects as part of a test for something else. Some things are hard / impossible to test without custom effects. Same with textures - I wanted to compile a texture as part of an sRGB visual test.

There are some pre-compiled `.xnb`s in the test assets folder - we could go further down that path, but personally I find them difficult to work with, without a nice way of quickly regenerating those `.xnb` files - maybe we could improve that, and then it wouldn't be a bad option.

tomspilman:
> we could go further down that path, but personally I find them difficult to work with

Yeah I agree.

We should just write a utility method that compiles an FX file and returns an `Effect`.  We can then use that from various places in the test framework as needed.

You just need to call `EffectProcessor` which returns a `CompiledEffectContent` which internally holds the byte array you can pass to the `Effect` constructor.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -16,8 +16,14 @@ public partial class GraphicsDevice : IDisposable
         private bool _isDisposed;
 
         private BlendState _blendState;
-        private BlendState _actualBlendState;
-        private DepthStencilState _depthStencilState = DepthStencilState.Default;
+        private BlendState _actualBlendState;
```
tomspilman: Do you mind cleaning up the blend state declarations?  Like here you have the `_blendState` and `_actualBlendState`, but have all the `_blendStateAdditive`/`_blendStateAlphaBlend`/etc down around line 300.

Test/Framework/Components/Two3DCubesComponent.cs
```diff
@@ -0,0 +1,234 @@
...
+object code form, you may only do so under a license that complies with this
+license.
+
+(E) The software is licensed "as-is." You bear the risk of using it. The
+contributors give no express warranties, guarantees or conditions. You may have
+additional consumer rights under your local laws which this license cannot
+change. To the extent permitted under your local laws, the contributors exclude
+the implied warranties of merchantability, fitness for a particular purpose and
+non-infringement.
+*/
```
tomspilman: Go ahead and put the short MonoGame Team copyright block we have in some of the other files here.

tomspilman:
Looks good overall... will merge when I wake up. :)

MonoGame.Framework/Graphics/GraphicsDevice.cs
tgjones: I was wondering if anyone (else) would notice that :-) Yeah, I planned to move those, just wasn't sure which PR to fit that into.

tgjones:
> We should just write a utility method that compiles an FX file and returns an Effect.

I've added an `EffectUtility.CompileEffect` test method, but I haven't actually used it yet. I had temporarily forgotten how stencil buffers work :) So I thought I needed a custom effect to test it, but obviously I don't. I still think it will be useful though, so I've left it in.

tgjones:
(By the way, I notice now that I've installed the EditorConfig VS extension, it makes whitespace changes beyond what I intend - is that okay? It's kind of annoying, I like to keep whitespace changes to a minimum in PRs. But if I disable EditorConfig, that kind of defeats the purpose of MonoGame having an `.editorconfig` file?)

tgjones:
@tomspilman does this look okay?

(I've disabled the EditorConfig VS extension for now, to prevent  those unwanted whitespace changes for future PRs.)

tomspilman:
> prevent those unwanted whitespace changes 

Yea... whitespace changes are bad for us.  They basically can cause false code conflicts and make simple changes unintelligible in a diff.

Small ones happen at times and that is fine, just don't go on "code cleanup" sprees.

tomspilman:
@tgjones - Note we have https://github.com/mono/MonoGame/issues/1512 and https://github.com/mono/MonoGame/issues/1511 which cover our plans for code cleanup.

tgjones:
> Yea... whitespace changes are bad for us.

In that case, it might be worth removing the `.editorconfig` file from the MG repo until after someone has done a full cleanup pass? As long as that file is present, then the EditorConfig VS extension will make whitespace changes (as I understand it).

tomspilman:
> it might be worth removing the .editorconfig  file from the MG repo 

You know I don't remember how that got there.

tgjones:
It was added in #2039.

I'm guessing that few, if any, MG contributors have the EditorConfig VS extension installed - the tell-tale sign is whitespace being removed from the ends of lines that aren't part of the intended change.

PR #3518, opened 2015-02-11
Content readers cache

nkast:
Here is a memoryPool for byte[] objects used by ContentReaders. I found that this can reduce load/resume times down to 70%.
Haven't tested it on other than DX platforms and probably needs some more work but this thread on community reminded me I need to submit this.
http://community.monogame.net/t/loading-huge-content/1942
Feel free to propose any changes to names, functionality, and of course feedback on other platforms, profile data, etc.

contentreadersbufferpool


MonoGame.Framework/Content/ContentReaders/EffectReader.cs
```diff
@@ -31,11 +36,19 @@ private static bool Contains(string search, string[] arr)
...
+            Effect effect = existingInstance;
+
+            var dataSize = (int)input.ReadUInt32();
+            byte[] data = MemoryPool.Current.GetPooledBuffer(dataSize);
+            input.Read(data, 0, dataSize);
+
+            if (effect == null)
+            {
+                effect = new Effect(input.GraphicsDevice, data, 0, dataSize);
+                MemoryPool.Current.PoolBuffer(data);
```
tomspilman: This looks like a memory leak.  If `effect != null` the `data` is never returned the to pool.

MonoGame.Framework/Content/ContentReaders/Texture3DReader.cs
```diff
@@ -1,30 +1,35 @@
...
                 texture = existingInstance;
             
             for (int i = 0; i < levelCount; i++)
             {
-                int dataSize = reader.ReadInt32();
-                byte[] data = reader.ReadBytes(dataSize);
+                int dataSize = input.ReadInt32();
+                byte[] data = MemoryPool.Current.GetPooledBuffer(dataSize);
+                input.Read(data, 0, dataSize);
                 texture.SetData(i, 0, 0, width, height, 0, depth, data, 0, dataSize);
```
tomspilman: Leak again?  `MemoryPool.Current.PoolBuffer(data)` is never called.

tomspilman:
I think maybe this is a little over engineered.

`MemoryPool` is never even used as a pool in this PR.  It just pulls out an array and returns it to the pool over and over, getting a larger array from time to time.  This is more of a temporary/scratch memory buffer than anything else.

You can implement something simpler by just having a single `byte[]` array which you discard and reallocate if you find the need for a larger array.  This way there is no concern about clearing a pool which could have a bunch of smaller buffers.

Also better place to store this scratch memory is in `ContentManager` as it is reachable by `ContentReader.ContentManager`.  And `ContentManager` does not allow for multiple threads to load content at the same time... so no extra lock is necessary.

MonoGame.Framework/Graphics/Effect/Effect.cs
```diff
@@ -69,9 +69,11 @@ protected Effect(Effect cloneSource)
             Clone(cloneSource);
 		}
 
-        public Effect (GraphicsDevice graphicsDevice, byte[] effectCode)
+        public Effect (GraphicsDevice graphicsDevice, byte[] effectCode, int index = 0, int count = -1)
```
tomspilman: I was a little worried about this at first, but really it is one of the few places where we don't have a buffer range option.

I would just break it into two constructors:

```
public Effect (GraphicsDevice graphicsDevice, byte[] effectCode)
  : this(graphicsDevice, effectCode, 0, effectCode.Length)
{
}

public Effect (GraphicsDevice graphicsDevice, byte[] effectCode, int offset, int count)
: this(graphicsDevice)
{
```
```

File: MonoGame.Framework/Graphics/Effect/Effect.cs [hunk 4]
```diff
@@ -122,15 +124,14 @@ public Effect (GraphicsDevice graphicsDevice, byte[] effectCode)
             Clone(cloneSource);
         }
 
-        private MGFXHeader ReadHeader(byte[] effectCode)
+        private MGFXHeader ReadHeader(byte[] effectCode, int index = 0)
```
tomspilman: This is an internal API...  I would remove the default parameter value and make everyone deal with the new argument.

tomspilman:
@nkast - You will need to rebase or merge the latest changes from develop.

MonoGame.Framework/Content/ContentReaders/EffectReader.cs
nkast: Effect.Read() was very compact, I had to rewrite a lot of lines and some changes related to https://github.com/mono/MonoGame/pull/3011 leaked into this. I better remove the check for existingInstance in this PR, let that issue evaluated separately.

nkast:
@tomspilman Here is the first version, very similar to what you describe. I don't use threads either but I consider this scenario before submit it. So, how about DX12? Anyway, buildLog says  PSMobile don't have SortedSet, so I can revert to the code below and rename it to `GetScratchBuffer(...)` (much better name) and add a public MinimumScratchBufferSize so people can customize the initial buffer. 

``
        static byte[] _cache;
        static byte[] GetCachedBuffer(int size)
        {
            size = Math.Max(size, 2097152); //min _cached size
            if(_cache == null)
                _cache = new byte[size];
            if(_cache.Length < size)
                _cache = new byte[size];
            return _cache;
        }
``

tomspilman:
> So, how about DX12? 

DX12 won't change how content is loaded.  One content manager can only load one content object at a time.

> can revert to the code below and rename it to `GetScratchBuffer(...)` (much better name)

I agree...  that code seems better.

Also where do you plan to put this?  I still recommend making it not static and making it an internal API in `ContentManager`.


> add a public `MinimumScratchBufferSize` so people can customize the initial buffer

I would just make it 1MB by default and not bother making it public.  Most all hardware can spare 1MB for scratch memory during loading.

nkast:
Another place could be `ContentReader`.
Have a `byte[] ReadDataBlock(bool useScratchBuffer)` and move there the `int dataSize = input.ReadInt32();` and `input.Read(data, 0, dataSize);` which gets repeated constantly. 
That could only work if we had a single static scratchBuffer. Otherwise we can pass a IScratchProvider interface.

In the case of #3011 we can also have a `.SkipDataBlock(...)` to move the position since we don't really need to read data.

tomspilman:
> Another place could be `ContentReader`

A new `ContentReader` is allocated for each content load, so it would have much less benefit there.


> That could only work if we had a single static scratchBuffer

While it seems silly... someone could use multiple `ContentManager`s in different threads at the same time.  For this reason the scratch buffer needs to either be per content manager or static with locks.

> Otherwise we can pass a `IScratchProvider` interface.

That seems unnecessary to me.

nkast:
"One content manager can only load one content object at a time."

"While it seems silly... someone could use multiple ContentManagers in different threads at the same time. For this reason the scratch buffer needs to either be per content manager or static with locks."

Ok, that was helpful. I will get the changes early this week.

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -668,5 +669,13 @@ public IServiceProvider ServiceProvider
 				return this.serviceProvider;
 			}
 		}
+
+        internal byte[] GetScratchBuffer(int size)
```
tomspilman: Eventually we could consider making this public for those writing custom content readers.  We would have to add docs to it explaining its usage and benefits.

tomspilman:
>  I found that this can reduce load/resume times down to 70%.

@nkast - What platform was this on by the way?

nkast:
@tomspilman That was on WP8.0/Release build. I tried to replicate the test but for some reason the same code now takes around 32sec to run under the profiler. no idea...
On Windows/WinRT is less -around90%, and the GC drop to 11 from 14.
contentreaderscache

tomspilman:
This got stalled out on my side.

@nkast - So what is the benefit of this optimization now?  You originally claimed a huge optimization from this, but it seems like you are not seeing that now?

nkast:
@tomspilman I can't repeat the test on WP8 because now the profiler add a lot of overhead, doesn't necessary mean that the gain is not there. On my desktop I see less gain but then again it has faster CPU and memory. 
This is the kind of opt that can only be tested on real world apps/games. The way I did it was place an exit() the moment my game transition to the Menu from the loading screen, and measure the overall loading time (+loading intro, bg music, stage graph, menus, settings). I could try to isolate the moment game resources are initialized and wrap them around a Stopwatch. What we really need here is feedback from others. Every game will see different gains from this depending on how many/big resources it use.

After a quick research, the reason memory allocations are slow is related to C# initializing every byte to zero.
http://stackoverflow.com/questions/22885616/could-allocating-a-byte-array-be-performance-critical
Even WCF uses a BufferManager, available only to WP8, dinno about mono.
https://msdn.microsoft.com/en-us/library/system.servicemodel.channels.buffermanager(v=vs.100).aspx

nkast:
@tomspilman After a second tought, ReloadAssets() would be the best place to place the stopWatch and measure what we really gain from this.

nkast:
I put a stopwatch in ContentManager.ReloadGraphicsContent() and run a Debug build with Optimization on. Tested on Lumia 620/WP8.1, project: HardHat Zone. Resume 3 times, B is with ScratchBuffer, C is with the following line commented out so we get a new buffer every time. Values are seconds.
'//if (scratchBuffer == null || scratchBuffer.Length < size)'

The results:
scratchbuffer

Yes, it's more than x3 times faster!  
It's more than initializing to zeros here, maybe searching for a large enough buffer in .net heap is more 
expensive than I though. In fact, I can't believe my findings, better to have more people verify this and get results on iOS and Android as well. Amazing!

KonajuGames:
@tomspilman , thoughts on this one now?

tomspilman:
My only worry is it getting used wrong and someone keeping the instance of the scratch buffer beyond the `ContentTypeReader.Read` method.  I can't really think of a way to prevent that or quickly detect it.

nkast:
> My only worry is it getting used wrong and someone keeping the instance of the scratch buffer beyond the ContentTypeReader.Read method. I can't really think of a way to prevent that or quickly detect it.

I don't think it's possible to detect that. I understand this optimization is an extra headache for maintenance. I could mark the method as `unsafe` to give clues that there's something to be extra careful about here, but this will probably hurt the code readability.

So, should I close this PR?
Do you think the new `Effect(GraphicsDevice graphicsDevice, byte[] effectCode, int index, int count)` constructor would be helpful to submit in a separate PR?

Perhaps in the future we could have new `.SetData(...)` methods that read from Stream which wouldn't  allocate big memory buffers.

tomspilman:
I noticed another two locations where something like this could be used:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentManager.cs#L518

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentManager.cs#L463

Both just create scratch memory for decoding.  Could be it needs something specific for those decoding cases.

nkast:
The LZ4 is already solved in another PR where I pass the LZ4Stream directly to ContentReader bypassing any buffer & memoryStream.  LZ4Stream internally use a 64k buffer and decompress the input in blocks.
For LZX we need to A) use another scratch buffer  since the one we have will be used by the ContentReader, B) revert back to MemoryPool which support multiple buffers or C) Implement a LZXStream similar to LZ4Stream.

Option C is consisted with the rest of the code and sound like an interesting challenge but might not worth the time and effort since LZX is for legacy XNA files.

KonajuGames:
> might not worth the time and effort since LZX is for legacy XNA files.

Agreed.  I don't think anyone will be using assets compressed with the XNA content pipeline any more.

KakCAT:
> might not worth the time and effort since LZX is for legacy XNA files.
> 
> Agreed. I don't think anyone will be using assets compressed with the XNA content pipeline any more.


I am using assets compressed with the XNA content pipeline :) I suppose I can use the monogame content pipeline for future projects but having to port legacy projects to the new content pipeline may be quite annoying.

If you do such change would it be possible to postpose it after the 3.5 release? i.e. 3.5 being the last release being compatible with XNA xnb files.

tomspilman:
> The LZ4 is already solved in another PR where I pass the LZ4Stream directly to ContentReader bypassing any buffer & memoryStream. 

Ah... yes.  That is https://github.com/mono/MonoGame/pull/4522.


>  I don't think anyone will be using assets compressed with the XNA content pipeline any more.

Right... few are, so less reason to spend time optimizing that path.

@KakCAT - No one is proposing dropping that support...  just if we plan time optimizing that path further.

tomspilman:
FYI...  I just tested this PR on Vita and it was a good improvement to garbage generation on our current ports.

KakCAT:
that's great to hear, sorry for the interruption.

nkast:
> FYI... I just tested this PR on Vita and it was a good improvement to garbage generation on our current ports.

:thumbsup:

KonajuGames:
I'm going to merge this.  Thanks @nkast

PR #3520, opened 2015-02-12
Create default SamplerState objects (AnisotropicClamp, etc.) per GraphicsDevice

tgjones:
Continues from #3472, #3511, and #3516, which did the same thing for `BlendState`, `RasterizerState`, and `DepthStencilState`, respectively.

Next, I'll do the one I actually needed: moving `Effect.EffectCache` into `GraphicsDevice`.


tomspilman:
> Next, I'll do the one I actually needed: moving  Effect.EffectCache  into  GraphicsDevice .

You are kicking ass getting all this done...  it was way overdue.

tomspilman:
Also don't forget to delete the `ObjectFactoryWithReset` class if it is no longer used.

tomspilman:
Don't think this failure is you.  I will check out the build box in the morning and re-run the build.

tgjones:
> it was way overdue.

Cool, happy to help! It's been a good route to learning my way around the code.

I think we'll also be able to get the OpenGL test suite up and running now. I briefly tried it, and while some tests still fail, they're failing for proper reasons now.

> Don't think this failure is you.

Actually I think this is me - these are genuine failures. Looks like I forgot to run the whole test suite before submitting. I'll push a fix.

tomspilman:
This great... merging!

prollin:
@tgjones just out of curiosity, I am not sure to understand the role of `_actualSamplerStates` and `_samplerState*` in the `SamplerStateCollection` class.
Just want to make sure I understand your changes before i pulled it in our production branch :)

tomspilman:
>  I am not sure to understand the role of  `_actualSamplerStates` and `_samplerState*`  in the  `SamplerStateCollection` class.

This solves the problem of having `GraphicsDevice` specific state in the default state statics:

  - `SamplerState.LinearWrap`
  - `SamplerState.LinearClamp`
  - `SamplerState.PointWrap`
  - `SamplerState.PointClamp`
  - etc.

When you set one of those states we replace it with an internal one that is specific to the current `GraphicsDevice`.  This way there are no GPU resources ever assigned to the default `SamplerState` statics.

To do this @tgjones added the real default `SamplerState`s as internals to the `GraphicsDevice` instance.  We also then need two arrays in `SamplerStateCollection`... the real states assigned to the GPU and the ones passed in which we need for the getter to work correctly.

We solve this before with this `ObjectFactoryWithReset` class, but it didn't allow for multiple `GraphicsDevice` objects at once.  This is all much much cleaner.

tgjones:
@prollin I've also applied the same type of change to `BlendState`, `DepthStencilState`, and `RasterizerState`. There are instance variables in `GraphicsDevice` for each of the static state objects (`BlendState.Opaque`, etc.).

Note that *user-created* state objects will still need to be created once per `GraphicsDevice` - I've only special-cased the built-in default state objects.

I'm actually not sure whether XNA allows you to attach a state object to a different `GraphicsDevice` and continue using it, but MG doesn't currently have any mechanism to re-initialize device-specific state when the device the state object is bound to changes.

tomspilman:
> I'm actually not sure whether XNA allows you to attach a state object to a different `GraphicsDevice` and continue using it

First remember than in a DX11 world the render states are specific to the graphics device.

So there are two scenarios.

First the case where you have two different `GraphicsDevice` instances at the same time.  I guess this could happen in a case of system with two graphics cards in it.  Could you ping pong the same render state from one and then another?  Yes...  but it means recreating the render state every time it switches devices.

The second case is when you dispose one `GraphicsDevice` then you create another and start using it.  Can you reuse render state objects from the previous device with the new one?  No...  the render state objects are a `IDisposable` `GraphicsResource`.  When the `GraphicsDevice` is disposed all the render states are disposed... and you shouldn't be reusing a disposed state.

tgjones:
> First the is where you have two different GraphicsDevice instances at the same time.

This is the scenario I was wondering about - I don't know what XNA does. Not that it's something I particularly care about - one `GraphicsDevice` at a time is enough for me.

> When the GraphicsDevice is disposed all the render states are disposed

Actually, do we know if XNA even disposes all the associated graphics resources when you dispose the device? Although I think it's helpful that MG does do that.

tomspilman:
> do we know if XNA even disposes all the associated graphics resources when you dispose the device?

I'm not certain as I haven't tested it.  I guess if the resource was never bound to the device it won't dispose it.

PR #3522, opened 2015-02-12
XNA test fixes

polsys:
* Following up on #3420 comments on XNA tests, on my computer the visual tests started working after I added a `AbsorbQuitMessage` call to `MockGame.EndRun`. For reasons beyond my understanding and interest, `AbsorbQuitMessage` never got called previously with `MockGame`.
* The `Vector2` tests compared the floating point values for equality, and that caused two test failures on XNA. I added a comparer based on existing bounding box comparer and updated the tests to use it where reasonable.

While testing this, the test runner got stuck multiple times at [this line](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Windows/WinFormsGameWindow.cs#L419). I am confident it is caused by the changes in #3474. The `PeekMessage` call intended to remove WM_QUIT messages could return, for example, WM_MOUSEMOVE messages (which is why reproducing this requires user interaction) and the 100ms wait only makes it more probable. Even more, thanks to the `do...while` loop behavior, this wait is guaranteed to happen at least once. As detailed [here](http://blogs.msdn.com/b/oldnewthing/archive/2005/11/04/489028.aspx), WM_QUIT is not a posted but a generated message, so the loop as of now does not really help. Removing the WM_QUIT hack altogether seemed to work for me, but I did not include it here yet... I would like to have someone else test it too, in case I am mistaken.

### DIFF
--- a/Test/Framework/GameTest.cs
+++ b/Test/Framework/GameTest.cs
@@ -308,6 +308,14 @@ namespace MonoGame.Tests {
 				DrawCount = 0;
 			}
 
+            protected override void EndRun()
+            {
+                base.EndRun();
+#if XNA
+                AbsorbQuitMessage();
+#endif
+            }
+
 			protected override void Update (GameTime gameTime)
 			{
 				base.Update (gameTime);
--- a/Test/Framework/TestGameBase.cs
+++ b/Test/Framework/TestGameBase.cs
@@ -323,7 +323,7 @@ namespace MonoGame.Tests {
 
 		const uint WM_QUIT = 0x12;
 
-		private static void AbsorbQuitMessage ()
+		protected static void AbsorbQuitMessage ()
 		{
 			NativeMessage msg;
 			if (!PeekMessage (out msg, IntPtr.Zero, 0, 0, 0))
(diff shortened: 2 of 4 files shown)


tomspilman:
@tgjones - Thoughts on this?

tgjones:
* The `Vector2` test fixes are great - I'd been meaning to fix those too.
* I have done some tests, and can confirm that this `AbsorbQuitMessage` change does fix the problem with running XNA tests, at least for me. (I've tried to debug it to find out why `AbsorbQuitMessage` wasn't being called before this change, but haven't got anywhere. I notice that it *was* called when you run a single test, but when you run all the tests together, it wasn't called.)

So these changes look good to me.

polsys:
I'll create a separate PR for fixing the test runner getting stuck, as it is not related to the XNA tests.

tomspilman:
@polsys - Were you going to modify this PR further or is it done?

polsys:
I'm done with this, now the XNA tests work for me.

tomspilman:
Ok... merging!

PR #3523, opened 2015-02-12
Docs Introduction

cra0zy:
Notes:
<ul>
<li>I just added FAQ with only that 1 question from the previous docs. If you can remember some question that has been consistent, do tell me and I'll add it.</li>
<li>I filled in system requirements page with the info I knew, there are still parts missing</li>
<li>Linux installation instructions are installation instructions that will work for MonoGame 3.3 and up</li>
</ul>


tomspilman:
Seems like an ok start... merging!

PR #3525, opened 2015-02-13
Thread safe static variable accessing for windows

christallire:
PR separated from #3436


tomspilman:
Thanks @christallire !

PR #3526, opened 2015-02-13
Move EffectCache to GraphicsDevice

### LINKED ISSUE #3449: Request: remove effects from Effect.EffectCache when effect is disposed

tgjones:
Fixes #3449.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 
 
@@ -90,6 +91,11 @@ namespace Microsoft.Xna.Framework.Graphics
 
         private readonly ConstantBufferCollection _vertexConstantBuffers = new ConstantBufferCollection(ShaderStage.Vertex, 16);
         private readonly ConstantBufferCollection _pixelConstantBuffers = new ConstantBufferCollection(ShaderStage.Pixel, 16);
+
+        /// <summary>
+        /// The cache of effects from unique byte streams.
+        /// </summary>
+        internal Dictionary<int, Effect> EffectCache;
 
 		// TODO Graphics Device events need implementing
 		public event EventHandler<EventArgs> DeviceLost;
@@ -211,6 +217,8 @@ namespace Microsoft.Xna.Framework.Graphics
             _rasterizerStateCullNone = RasterizerState.CullNone.Clone();
 
             RasterizerState = RasterizerState.CullCounterClockwise;
+
+            EffectCache = new Dictionary<int, Effect>();
         }
 
         ~GraphicsDevice()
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -413,6 +421,9 @@ protected virtual void Dispose(bool disposing)
                     // Dispose of all remaining graphics resources before disposing of the graphics device
                     GraphicsResource.DisposeAll();
 
+                    // Clear the effect cache.
+                    EffectCache.Clear();
```
tomspilman: I guess the idea here is that `GraphicsResource.DisposeAll()` is taking care of calling `Effect.Dispose()`?

tgjones: Yes - I stepped through in the debugger, and the `Effect` I was testing with was disposed in `GraphicsResource.DisposeAll()`.

tomspilman: Cool...  so the only issue remaining now is that `GraphicsResource.DisposeAll()` will dispose resources owned by a second `GraphicsDevice`.  Same drill as all these other changes... the static collection holding all the `GraphicsResource`s really needs to be owned by the `GraphicsDevice` instance.

tgjones: Yep, I have a commit with exactly that fix sitting in a local branch already. I'll push it up as soon as this one is merged - just don't like to have too many PRs open at one time, it becomes difficult to manage.

PR #3528, opened 2015-02-13
Move GraphicsResource list to GraphicsDevice

tgjones:
This changes the existing behaviour somewhat: before this change, the `GraphicsResource` **constructor** added objects to the static `GraphicsResource.resources` list. After this change, objects are only added to the `GraphicsDevice._resources` list when their `GraphicsDevice` property is set.

Most classes that derive from `GraphicsResource` set the `GraphicsDevice` in their constructor. So there's no actual change there. AFAIK, the only classes that do something different are the state classes - `BlendState`, `DepthStencilState`, `RasterizerState`, `SamplerState` - and `VertexDeclaration`. These classes have their `GraphicsDevice` properties set when they are first bound to a device. So they won't be added to the "to dispose" list until they are bound to a device - but I think that's fine, because there's also nothing to dispose of until they are bound to a device.

In the case of `VertexDeclaration`, it has a similar **potential** issue to the one I've just fixed for the static state objects. But I *think* that's okay, because we don't actually create any device-specific objects in `VertexDeclaration` objects themselves, as far as I can tell.

Anyway, it's enough of a change that I'd appreciate one or two people having a good look at this.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsResource.cs
+++ b/MonoGame.Framework/Graphics/GraphicsResource.cs
@@ -39,21 +39,13 @@
 #endregion License
 
 using System;
-using System.Collections.Generic;
+using System.Diagnostics;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
     public abstract class GraphicsResource : IDisposable
     {
         bool disposed;
-        
-        // Resources may be added to and removed from the list from many threads.
-        static object resourcesLock = new object();
-
-        // Use WeakReference for the global resources list as we do not know when a resource
-        // may be disposed and collected. We do not want to prevent a resource from being
-        // collected by holding a strong reference to it in this list.
-        static List<WeakReference> resources = new List<WeakReference>();
 
         // The GraphicsDevice property should only be accessed in Dispose(bool) if the disposing
         // parameter is true. If disposing is false, the GraphicsDevice may or may not be
@@ -64,11 +56,7 @@ namespace Microsoft.Xna.Framework.Graphics
 
         internal GraphicsResource()
         {
-            lock (resourcesLock)
-            {
-                _selfReference = new WeakReference(this);
-                resources.Add(_selfReference);
-            }
+            
         }
 
         ~GraphicsResource()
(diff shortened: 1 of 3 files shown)


Tools/2MGFX/MonoGame.Framework/GraphicsDevice.cs
```diff
@@ -1,7 +1,18 @@
...
     {
+        internal void AddResourceReference(WeakReference resourceReference)
+        {
+            
+        }
+
+        internal void RemoveResourceReference(WeakReference resourceReference)
+        {
+            
+        }
```
tomspilman: Sorry... I hope to clean up that mess in 2MGFX soon.

tgjones: I haven't dug into it to see why we need dummy classes, but I'm sure there was a good reason at the time :) Do people still use 2MGFX as a standalone executable? I guess most people just it in the content pipeline, where it's compiled in to the main content pipeline assembly.

tomspilman:
> After this change, objects are only added to the  GraphicsDevice._resources  list when their  GraphicsDevice  property is set.

This is probably simpler as it covers both when the resource is bound during the constructor or in the case of render states one first use.

MonoGame.Framework/Graphics/GraphicsResource.cs
```diff
@@ -174,7 +127,23 @@ public GraphicsDevice GraphicsDevice
...
             internal set
             {
+                Debug.Assert(value != null);
+
+                if (graphicsDevice == value)
+                    return;
+
+                // VertexDeclaration objects can be bound to multiple GraphicsDevice objects
+                // during their lifetime. But only one GraphicsDevice should retain ownership.
+                if (graphicsDevice != null)
```
tomspilman: This seems fine to me.  It is good that we're protecting against the case of a resource changing graphics devices if it ever happens.  It still is the responsibility of the resource to manage switching of graphics devices if it supports or allows it.

Tools/2MGFX/MonoGame.Framework/GraphicsDevice.cs
tomspilman: > I haven't dug into it to see why we need dummy classes

We need them because 2MGFX uses some features of `MonoGame.Framework`...  nothing platform specific... just some enums and the render state classes.

If I had referenced the `MonoGame.Framework` assembly at the time 2MGFX would require assemblies for Lidgren, Direct Input, and other junk it never ever needs.  By just sharing source code from the framework and adding a dummy class for `GraphicsDevice` I avoided that dependency.

Now Lidgren junk is separate from the main `MonoGame.Framework`, so things are better.  However there was discussion of breaking up `MonoGame.Framework` further, but I don't know if that will happen or is a good idea.

> Do people still use 2MGFX as a standalone executable?

It exists because that is how I developed the stock effects.  If you are loading an effect from the `Effect(byte[[)` constructor you need an effect not packaged in XNB form.  The sock effects are embedded resources in the 2MGFX format.

tomspilman:
> But I think that's okay, because we don't actually create any device-specific objects in  VertexDeclaration  objects themselves, as far as I can tell.

That seems correct to me.

tomspilman:
@KonajuGames - Give this a review and if you are happy with this then merge it.

KonajuGames:
I like it.  It should greatly help with multiple GraphicsDevice instances.  Merging.

PR #3529, opened 2015-02-13
Implement BoundingFrustum-BoundingFrustum intersection methods

tgjones:
Implements a couple of previously-commented-out methods, with a test.

### DIFF
--- a/MonoGame.Framework/BoundingFrustum.cs
+++ b/MonoGame.Framework/BoundingFrustum.cs
@@ -128,15 +128,33 @@ namespace Microsoft.Xna.Framework
             result = intersects ? ContainmentType.Intersects : ContainmentType.Contains;
         }
 
-        /*
         public ContainmentType Contains(BoundingFrustum frustum)
         {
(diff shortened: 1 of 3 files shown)


tgjones:
Actually please don't merge this yet, I think the implementation might be wrong.

tomspilman:
Ok.

tgjones:
That's more like it - if the tests pass, this is ready to merge. Much of this could be inlined, but I think slower-than-it-could-be is better than not-implemented-at-all.

This is a useful intersection test for shadow mapping - if the spotlight frustum or any of the 6 point light frustums don't intersect the camera frustum, you can skip that shadow map. At least, that's what I'm using it for.

tomspilman:
> Much of this could be inlined, but I think slower-than-it-could-be is better than not-implemented-at-all.

Yes...  this is just a starting point.

Ideally we inline and optimize for the frustum vs frustum case at some point.  Lots of little things we could do better in there.

tomspilman:
@tgjones - Merged!

PR #3533, opened 2015-02-14
Handle null shared resources in IntermediateSerializer

tgjones:
and throw better exception when shared resource is not found during deserialization.


tomspilman:
Thanks!

PR #3536, opened 2015-02-15
Use struct enumerator in EffectPassCollection

### LINKED ISSUE #3498: EffectPassCollection.Count: getting tons of garbage
Hi,

I'm getting a big amount of garbage when using a "'classic' foreach (var eff in  effect.CurrentTechnique.Passes)" loop.

I'm aware that this kind of loops can get garbage, so I decided to do a simple for, getting "Count" before the loop. Weirdly, it seems the garbage is coming from "Count" :

(memory profiler output)
```
    Allocating Method	Instances	Inclusive Size (Bytes)	Exclusive Size (Bytes)	Allocating Module
    System.SZArrayHelper.GetEnumerator<M>()	48347	966940	966940	mscorlib.ni.dll
    Microsoft.Xna.Framework.Graphics.EffectPassCollection. System.Collections.Generic.IEnumerable<Microsoft.Xna.Framework.Graphics.EffectPass>.GetEnumerator()	48347	966940	0	MonoGame.Framework.DLL
    System.Linq.Enumerable.Count<M>(System.Collections.Generic.IEnumerable`1<M>)	48347	966940	0	System.Core.ni.dll
    my function call here
```
I'm drawing a big amount of objects, so the GC hits every 15 seconds or so.

My solution so far is 'caching' the effect.CurrentTechnique.Passes.Count() value for every effect at the start, and use those cached values. This gets rid of the GC for several minutes. However this is a quite dirty hack.

Isn't there a way to get rid of that garbage generation?

Platform is WP8, but I don't know if this is common to all platforms.

EDIT: what I don't understand is why that function could generate garbage. Looking at the code, it seems it's just an Array.Length function...

tgjones:
Fixes #3498.

Instead of duplicating the `List<T>.Enumerator` struct, I just re-used it.

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/EffectPassCollection.cs
+++ b/MonoGame.Framework/Graphics/Effect/EffectPassCollection.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
@@ -44,6 +45,11 @@ namespace Microsoft.Xna.Framework.Graphics
             get { return _passes.Length; }
         }
 
+        public Enumerator GetEnumerator()
+        {
+            return new Enumerator(_passes);
+        }
+            
         IEnumerator<EffectPass> IEnumerable<EffectPass>.GetEnumerator()
         {
             return ((IEnumerable<EffectPass>)_passes).GetEnumerator();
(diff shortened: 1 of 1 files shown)


tomspilman:
Actually I think we are better off implementing a struct enumerator here.

If you look in the history you'll see we used to have a `List<EffectPass>` here before.  

  https://github.com/mono/MonoGame/pull/1660

The changes were prompted by wanting to reduce the memory footprint and garbage generated by the `Effect` classes.  We found that our usage of `List<>` wasted lots of memory...   by removing them we saved over 25MB of memory with 10,000 `BasicEffect` instances.

I think a good idea here which would immediately come in handy in other places is a simple template based struct enumerator.  Throw it into the utilities and use it when we need a garbage free enumerator for array types.

tgjones:
Okay, cool. I'll do that.

tgjones:
Although it will have to be a public type - are we okay with exposing something in the Utilities namespace?

tomspilman:
Hum...  good point.

It is probably ok...  whatever we use will need to be public in some way.

Why doesn't array have a struct enumerator in the first place.  Seems like something the Mscorlib should solve.

tgjones:
My guess would be it's because Array pre-dates generics, and they didn't want to change the existing behaviour.

If it's not too much code (I'll have to look, I can't remember exactly), it might be nicer to have an Enumerator struct defined wherever we need one. Nicer for the API, not for MG developers.

tgjones:
I have re-implemented this using an `Enumerator` struct defined inside `EffectPassCollection`. Let me know if a generic struct is preferred - but since .NET uses nested `Enumerator` types in `List<T>` etc., it's nice to be consistent with that.

tomspilman:
> My guess would be it's because Array pre-dates generics, and they didn't want to change the existing behaviour.

Yeah...   lame.  If only if they would break a little compatibility here just to make this very core feature more efficient.  Maybe now with the open sourcing and the new coreclr push they could do that.

> Let me know if a generic struct is preferred

Naw...  it is fine for now.

If later we find it to be some sort of maintenance problem we can reconsider.

PR #3537, opened 2015-02-15
PNG codec

disportium:
First shot at platform-independent implementation of the PNG image standard.  Uses source from the BSD-licensed [ZLIB.NET](http://www.componentace.com/zlib_.NET.htm) for the DEFLATE algorithm.  Issue mono#2424

`public Texture2D Read(Stream inputStream, GraphicsDevice graphicsDevice)` (PngReader.cs)
`public Stream Write(Texture2D texture2D)` (PngWriter.cs)

Some limitations at this stage:

* Bit depths that are not multiples of 8 are unsupported (most PNG's are 8-bit per sample though)
* The ZLIB.NET DEFLATE algorithm fails on a very small number of PNG's (like [this](http://www.deviantart.com/art/Royal-Entrance-180978838) one for instance)
* Some PNG's that are technically malformed (invalid pixel data array length) will crash the reader, even though other PNG decoders (like Picasa) will decode them without complaint.
* The PNG writer ignores alpha values at the moment.
* Should be thread-safe but haven't tested it yet.

These issues will be addressed in the future.


tomspilman:
@disportium - This is a great first step... thanks!

Since we currently have a few platform specific implementations of  `Texture2D.SaveAsPng` that support alpha... we can switch this over until at least that is fixed.

For now getting this checked in as just a work in progress is enough.

MonoGame.Framework/Utilities/Png/PngCommon.cs
```diff
@@ -0,0 +1,685 @@
...
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using System.Collections;
+using System.Text;
+using Microsoft.Xna.Framework;
+
+namespace MonoGame.Utilities.Png
+{
+    public class Palette
```
tomspilman: Can you go thru and make all these internal objects `internal` and not `public`?

MonoGame.Framework/Utilities/Png/PngReader.cs
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        public Texture2D Read(Stream inputStream, GraphicsDevice graphicsDevice)
+        {
+            byte[] signature = new byte[8];
+            inputStream.Read(signature, 0, 8);
+
+            if (signature.EqualsByElement(HeaderChunk.PngSignature) == false)
+            {
+                throw new Exception("File does not have PNG signature.");
```
tomspilman: Ideally we should have a separate `PngReader.IsPng(Stream)` method so that `Texture2D.FromStream` can eventually test for PNG first.

MonoGame.Framework/Utilities/Png/PngWriter.cs
```diff
@@ -0,0 +1,208 @@
...
+        public PngWriter()
+        {
+            // TODO: currently ignoring alpha values while writing PNG images
+            colorType = ColorType.Rgb;
+        }
+
+        public Stream Write(Texture2D texture2D)
+        {
+            textureData = new Color[texture2D.Width * texture2D.Height];
+            texture2D.GetData<Color>(textureData);
```
tomspilman: @tgjones @KonajuGames  - `Texture2D.GetData<>()` doesn't do any sort of color reformatting does it?  Like if the texture is a `Bgra4444` then a call to `Texture2D.GetData<Color>()` will result in a bad color array correct?

I suspect this will have to be more complex eventually to deal with different formats.

tgjones: @tomspilman You're right - there's no format conversion in Texture2D.GetData - and although Direct3D supports conversion between some specific formats using CopyResource, it's quite limited (and I don't know about OpenGL).

KonajuGames: Texture.GetData<>() returns the raw texture data.  The Graphics.PackedVector classes are used to convert between pixel formats, but this is usually done on the game side where required.

tomspilman: > Graphics.PackedVector classes are used to convert between pixel formats

Ahh...  that is true.  So in theory we would use that to cast from the hardware format to `Color`.  Something like...

```
var packedData = new Bgra4444[texture2D.Width * texture2D.Height]; 
texture2D.GetData<Bgra4444>(packedData);

var colorData = new Color[texture2D.Width * texture2D.Height];
for (var i=0; i < packedData.Length; i++)
   colorData[i] = new Color(packedData[i].ToVector4());
```
Obviously a bit more complex as we need to support all the different packed formats.... assuming here XNA does as well.

MonoGame.Framework/Utilities/Png/PngReader.cs
disportium: Right - added such a method.

MonoGame.Framework/Utilities/Png/PngCommon.cs
disportium: All public access modifiers in PngCommon are now internal.

MonoGame.Framework/Utilities/Png/PngWriter.cs
disportium: Ah OK, thanks.  Will cater for this in the next version.

tomspilman:
@KonajuGames - Do you have any concern having this `MonoGame.Utilities.Png` and `MonoGame.Utilities.ZLib` public at the moment?  Should we look to hide these things?

KonajuGames:
Would the `System.IO.Compression.DeflateStream` class work with this, rather than providing our own zlib?

disportium:
@KonajuGames For some reason I am under the impression that `System.IO.Compression.DeflateStream` cannot be trusted on all platforms, but thinking about it now there's no good reason for that suspicion.  I'll test tonight.

disportium:
So the situation regarding compression is as follows:  DEFLATE is an algorithm, of which there are two main implementations (formats), GZip and Zlib.  The DEFLATE algoritm is defined in [RFC 1951](http://www.faqs.org/rfcs/rfc1951.html), GZip is defined in [RFC 1952](http://www.faqs.org/rfcs/rfc1952.html) and Zlib in [RFC 1950](http://www.faqs.org/rfcs/rfc1950.html).  

According to the PNG standard:  "Deflate-compressed datastreams within PNG are stored in the zlib format."  `System.IO.Compression.DeflateStream` only implements the GZip format, and my tests confirm that it fails on PNG payloads (according to the `DeflateStream` [documentation](https://msdn.microsoft.com/en-us/library/system.io.compression.deflatestream(v=vs.110).aspx): "The DeflateStream class uses the same compression algorithm as the gzip data format used by the GZipStream class.")

I have come across the [DotNetZip](http://dotnetzip.codeplex.com/) library which provides a ZlibStream class (it carries the MS-PL license).  I have replaced the previous `ZStream` class with the `ZlibStream` class from the DotNetZip project.  Besides the code being a bit less gnarly it also has the benefit of no longer crashing on some PNG files, as mentioned in the first comment of this pull request.

I've tried to minimize the amount of public members of the `ZlibStream` and related classes.

tomspilman:
> Besides the code being a bit less gnarly it also has the benefit of no longer crashing on some PNG files

That is fantastic to hear!  Once you get this final build error worked out we can merge this.

mgarstenauer:
> I have come across the DotNetZip library which provides a ZlibStream class (it carries the MS-PL license).

Note that the ZLIB code in DotNetZip is under a different license (not MS-PL). See http://dotnetzip.codeplex.com/SourceControl/latest#License.zlib.txt.
As far as I can see nothing critical. A permissive license, but the copyright notice and the disclaimer should be included and added to the MonoGame documentation.

disportium:
Right, it's building again - there were some problems with the Windows Phone platform not supporting the ISO-8859-1 encoding natively.

tomspilman:
Ok... going to merge this!  Thanks @disportium !

In your next PR I hope you hook these up to the `Texture2D.FromStream` and `Texture2D.SaveAsPng` methods.

Issue #3043, opened 2014-10-01
ContentCompiler::GetTypeWriter, issue with Generic types


trentmillar:
I found this logic may be unnecessary or will need overhauling unless I'm missing something:

            else if (type.IsGenericType)
            {
                var inputTypeDef = type.GetGenericTypeDefinition();

                Type chosen = null;
                foreach (var kvp in typeWriterMap)
                {
                    var args = kvp.Key.GetGenericArguments();

                    if (args.Length == 0)
                        continue;

                    if (!args[0].IsGenericType)
                        continue;

                    // Compare generic type definition
                    var keyTypeDef = args[0].GetGenericTypeDefinition();
                    if (inputTypeDef == keyTypeDef)
                    {
                        chosen = kvp.Value;
                        break;
                    }
                }

                try
                {
                    var concreteType = type.GetGenericArguments();
                    result = (ContentTypeWriter)Activator.CreateInstance(chosen.MakeGenericType(concreteType));

                    // save it for next time.
                    typeWriterMap.Add(contentTypeWriterType, result.GetType());
                }
                catch (Exception)
                {
                    throw new InvalidContentException(String.Format("Could not find ContentTypeWriter for type '{0}'", type.Name));
                }
            }

I found that custom generic types (not IList<T>, Dictionary<T>, etc.) would not be added to the `typeWriterMap` and would therefore not act as a ContentTypeWriter.

I had a peek into the XNA decompiled 4.0 libraries, specifically `ContentTypeWriterFactory::CreateWriter(Type)` which I believe is used in the replacement of MonoGame's `ContentCompiler::GetTypeWriter(Type)`, and I found that `Type.IsGeneric` isn't handled differently than vanilla types. 

So.....I commented out the `else if(type.IsGenericType)`. This funneled these Generic types into the `else` and handles them the same way XNA would, which is to wrap the type in the `ReflectiveWriter<T>`. I'm not exactly sure why or what the expected behavior is, but I can tell you now the XML content in the RPG Starter Kit is processed successful into proper XNBs.

You may want to take a look.

tgjones:
I believe this bug was fixed by 93c28c9181103b9a687a626cb3bbbde760e3cad2.

To make sure, I've submitted #3538 with a test specifically for this scenario.

PR #3538, opened 2015-02-15
Add serialization test for custom generic types

tgjones:
Verifies that #3043 is no longer an issue.


tomspilman:
This is a great way to verify that an old issue is fixed... write a unit test!

Issue #3125, opened 2014-10-28
Desired: Exception when VertexDeclaration stride does not match template parameter size in DrawUser(Indexed)Primitives


DissidentDan:
In XNA, when you use a VertexDeclaration whose stride is smaller than the size of the vertex data type supplied to DrawUserIndexedPrimitives (the IVertexType template parameter), an exception is thrown. However, there's no exception in MonoGame, making errors more difficult to debug. It should be pretty easy to add an exception to catch this early rather than digging through lots of lower-level code to figure out why you crashed.

Thanks!

tgjones:
@DissidentDan I've just written some tests for this, and while there is definitely some missing parameter validation, the particular case you mention doesn't seem to be covered by XNA.

i.e. this is allowed by XNA, where `VertexPositionColor.VertexDeclaration` has a smaller stride than `VertexPositionColorTexture`:

``` csharp
var vertexData = new[]
{
    new VertexPositionColorTexture(Vector3.Zero, Color.White, Vector2.Zero),
    new VertexPositionColorTexture(Vector3.Zero, Color.White, Vector2.Zero),
    new VertexPositionColorTexture(Vector3.Zero, Color.White, Vector2.Zero)
};
var indexData = new short[] { 0, 1, 2 };

Game.GraphicsDevice.DrawUserIndexedPrimitives(
    PrimitiveType.TriangleList, vertexData, 0, 3,
    indexData, 0, 1,
    VertexPositionColor.VertexDeclaration);
```

tgjones:
(unless you meant the exception that is thrown when the vertex shader input structure doesn't match the vertex declaration, in which case that error message was improved by #3487.)

Jjagg:
[This check](https://github.com/MonoGame/MonoGame/blob/ba5c722a399fa4e192859a7dd17377b20ae4d327/MonoGame.Framework/Graphics/GraphicsDevice.cs#L1157-L1158) should also be added [here](https://github.com/MonoGame/MonoGame/blob/ba5c722a399fa4e192859a7dd17377b20ae4d327/MonoGame.Framework/Graphics/GraphicsDevice.cs#L1051).

```
if (vertexDeclaration.VertexStride < ReflectionHelpers.SizeOf<T>.Get())
                throw new ArgumentOutOfRangeException("vertexDeclaration", "Vertex stride of vertexDeclaration should be at least as big as the stride of the actual vertices.");
```

[...]

PR #3542, opened 2015-02-17
GraphicsDevice Draw call parameter validation

tgjones:
Adds parameter validation for:

* `DrawPrimitives`
* `DrawIndexedPrimitives`
* `DrawUserPrimitives`
* `DrawUserIndexedPrimitives`

I don't know if this is controversial - we have parameter validation elsewhere in MG, but until now the `Draw*` methods didn't have it. Obviously there is a (tiny) performance overhead. As @tomspilman has said elsewhere, it's a pity that VS doesn't make it easy to switch between debug / release assemblies, which would be the best of both worlds.

I don't think `Debug.Assert` is a viable replacement, since (I assume) very few developers build MG from source (or at least, that will be true once there are more frequent MG releases), and so very few developers are using the debug assemblies.

**EDIT** Forgot to say that #3125 was the trigger for this PR. However, from what I've found, XNA doesn't actually throw an exception when the `VertexDeclaration` size doesn't match the vertex data size in `DrawUser[Indexed]Primitives`. That's verified in a unit test.

### DIFF
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -126,6 +126,7 @@
     <Compile Include="Framework\Visual\BlendStateTest.cs" />
     <Compile Include="Framework\Visual\DepthStencilStateTest.cs" />
     <Compile Include="Framework\Visual\EffectTest.cs" />
+    <Compile Include="Framework\Visual\GraphicsDeviceTest.cs" />
     <Compile Include="Framework\Visual\RasterizerStateTest.cs" />
     <Compile Include="Framework\Visual\SamplerStateTest.cs" />
     <Compile Include="Framework\Visual\ScissorRectangleTest.cs" />
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -1260,7 +1260,7 @@ namespace Microsoft.Xna.Framework.Graphics
             }
         }
 
-        private void PlatformDrawUserIndexedPrimitives<T>(PrimitiveType primitiveType, T[] vertexData, int vertexOffset, int numVertices, int[] indexData, int indexOffset, int primitiveCount, VertexDeclaration vertexDeclaration) where T : struct, IVertexType
+        private void PlatformDrawUserIndexedPrimitives<T>(PrimitiveType primitiveType, T[] vertexData, int vertexOffset, int numVertices, int[] indexData, int indexOffset, int primitiveCount, VertexDeclaration vertexDeclaration) where T : struct
         {
             var indexCount = GetElementCountArray(primitiveType, primitiveCount);
             var startVertex = SetUserVertexBuffer(vertexData, vertexOffset, numVertices, vertexDeclaration);
--- a/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
@@ -911,7 +911,7 @@ namespace Microsoft.Xna.Framework.Graphics
             vbHandle.Free();
         }
 
-        private void PlatformDrawUserIndexedPrimitives<T>(PrimitiveType primitiveType, T[] vertexData, int vertexOffset, int numVertices, int[] indexData, int indexOffset, int primitiveCount, VertexDeclaration vertexDeclaration) where T : struct, IVertexType
+        private void PlatformDrawUserIndexedPrimitives<T>(PrimitiveType primitiveType, T[] vertexData, int vertexOffset, int numVertices, int[] indexData, int indexOffset, int primitiveCount, VertexDeclaration vertexDeclaration) where T : struct
         {
             ApplyState(true);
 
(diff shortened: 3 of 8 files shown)


tomspilman:
@prollin - What do you think?

It seems like in your case the issue was specifically with `CurrentThread` being slow and that a few conditionals were bad for performance.

> and so very few developers are using the debug assemblies.

In theory... since we generate our own C# templates... we could use conditionals to reference debug assemblies in a project.  However it doesn't fix existing projects out there.

Another thought is introducing a 3rd project config:

  - Debug - Un-optimized with extra internal validation tests.
  - Release - Optimized with minimal validation.
  - Shipping - Optimized with no validation.

This way to take advantage of the best performance you have to opt-in.

tgjones:
> we could use conditionals to reference debug assemblies

NuGet packages wouldn't work well with that.

> Another thought is introducing a 3rd project config

The main problem with that, IMO, is one of education - it's yet another thing to trip developers up, another topic to be constantly asked on the forums, etc.

Yet another idea, although I don't love it, is to have a `static` field somewhere, `DisableParameterValidation`. Then it would be an easy opt-in. Obviously, you'd still pay the cost of that single conditional for every call. And it has the same problem as a 3rd project config - it's unusual, and unexpected.

tomspilman:
> NuGet packages wouldn't work well with that.

 Lame...  I so dislike NuGet already.


> Yet another idea, although I don't love it, is to have a `static` field somewhere,  `DisableParameterValidation`.

So we limit the overhead at runtime for validation...  although we would really need to measure the performance difference to be sure it is helping and not hurting.

Still I agree it kinda sucks.

tgjones:
> Still I agree it kinda sucks.

Yep. My vote is for keeping runtime validation in all build configs. I realise "If it was good enough for XNA..." isn't *necessarily* a convincing argument, but I don't see many people complaining about XNA performance.

tomspilman:
> I don't see many people complaining about XNA performance.

Well at least not on PC.  On X360 and WP7 there were complaints.  Mainly because of the weaker CPUs and weaker C# runtime/GC on those platforms... the same problem we face with mobile and Mono.

I think it is worth discussing this more, but I agree that we should try to match behavior for now as best as possible.

prollin:
This type of parameter validation has never been an issue so far; they never show up as bottlenecks in my tests (on mobile). We can have them to mimic XNA behavior for now, at least until we have clear evidence that they are an issue.

That being said, it might be better to inline all of them in each function instead of factorizing the validation in a static method. That method will most likely never get inlined (due to the exception throwing) and will have performance cost. Also, while it adds a bit of duplicated code, it makes it clear what is the parameter contract for each method.

tomspilman:
> That being said, it might be better to inline all of them in each function instead of factorizing the validation in a static method.

Yeah... as much as I loathe duplicate code, I value performance more.  The draw is such a low level call that I think it could use the extra boost from avoiding a function call when it is reasonable.

Also what is the current state of "forcing inline" in C#?  Do mobile platforms obey some of the newer attributes?

prollin:
> Also what is the current state of "forcing inline" in C#? Do mobile platforms obey some of the newer attributes?

`[MethodImpl(MethodImplOptions.AggressiveInlining)]` should be supported on mobile but since it remains a *hint*, there is no guaranty the method will be inlined.

It seems safer to simply inline manually for the low level MonoGame methods to ensure optimal performance on all supported platform.

tomspilman:
>  should be supported on mobile but since it remains a hint, there is no guaranty the method will be inlined

Funny how in C++ `inline` is also a hint, but everyone trusts it to actually inline.

Why is C# so screwed up in this regard?

prollin:
> Why is C# so screwed up in this regard?

C++ is no different here; the compiler is free to inline or not a method regardless of the keyword being present or not. It might just be that C++ is older and that C++ developers are a bit more performance-conscious in general.

For future reference, [here](http://blogs.microsoft.co.il/sasha/2012/01/20/aggressive-inlining-in-the-clr-45-jit/) is a good summary describing when a method is inlined or not. On Mono, `MethodImplOptions.AggressiveInlining` removes the size limit.
According to the author, method with exception handing wouldn't be inlined.

Back to this PR, @tgjones, I would simply inline the `Validate*` method you created in each `Draw* method`; other than that it looks good.

tomspilman:
> C++ is no different here; the compiler is free to inline or not a method regardless of the keyword being present or not.

True... but more often than not if you added the `inline` keyword C++ would normally always inline.  There were exceptions, but there seem to be fewer in C++.

> I would simply inline the  Validate*  method you created in each  Draw* method ; other than that it looks good.

Unless someone wants to a big investigation into forced inlining...  I agree.

tgjones:
> Back to this PR, @tgjones, I would simply inline the Validate* method you created in each Draw* method; other than that it looks good.

Cool, I'll do that.

Actually only `DrawUserIndexedPrimitives` calls out to a separate validation method - the other 3 `Draw*` methods already do the validation inline. (`DrawUserIndexedPrimitives` is special because there are two entry points into the platform-specific code; the other methods only have one.)

tgjones:
If the tests pass, then this should be ready to merge.

tomspilman:
Sweet!

tomspilman:
Ok... merging!

PR #3544, opened 2015-02-18
Fix serialization and deserialization of empty lists

tgjones:
(Hopefully) fixes an issue that @DDReaper encountered.


tomspilman:
I really would be impossible to fix any of these issues without the unit tests.  I don't know how we have been able to live without them for so long.

tomspilman:
Good to merge?

tgjones:
>  I don't know how we have been able to live without them for so long.

These trivial-but-annoying XML issues make me wish that MS would at least open-source their XNA test suite, even if not the rest of XNA...

> Good to merge?

Yep, I'm done.

PR #3545, opened 2015-02-18
Cancel Windows 8 KeyboardInput on right thread

Nezz:


tomspilman:
Looks good... merging!

PR #3546, opened 2015-02-18
Android orientation lock

Nezz:
Simple and useful feature. Many Android devices are very sensitive when it comes to rotation. If the user shakes the device the rotation might change, so this is very useful if you use accelerometer in your game. Just lock the rotation while the game is playing and unlock once the game is paused/stopped.

### DIFF
--- a/MonoGame.Framework/Android/AndroidGameActivity.cs
+++ b/MonoGame.Framework/Android/AndroidGameActivity.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Xna.Framework
 
         public bool AutoPauseAndResumeMediaPlayer = true;
         public bool RenderOnUIThread = true; 
+        public bool PreventRotation { get; set; }
 
 		/// <summary>
 		/// OnCreate called when the activity is launched from cold or after the app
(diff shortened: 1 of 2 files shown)


KonajuGames:
I think I prefer the terminology used in the comment, and in most feature lists for this feature: Lock Rotation.  Could the property be renamed to `LockRotation` rather than `PreventRotation`?

Nezz:
LockOrientation?

danzel:
Shouldn't you just update ```GraphicsDeviceManager.SupportedOrientation``` when you want to lock rotation/orientation?

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/GraphicsDeviceManager.cs#L544
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Android/AndroidGameWindow.cs#L103

tomspilman:
> Shouldn't you just update  GraphicsDeviceManager.SupportedOrientation  when you want to lock rotation/orientation?

Yeah...  I thought that is how it worked on XNA on WP7/8.

Can we not continue to support this with that API?

KonajuGames:
I'd forgotten about that.  Read the current orientation and set SupportedOrientation to that value.

tomspilman:
> Read the current orientation and set SupportedOrientation to that value.

Yeah...  that is a little clumsy now that you say it that way.

Maybe a new "LockOrientation" which works in conjunction with that is a good addition?

Nezz:
>Shouldn't you just update GraphicsDeviceManager.SupportedOrientation when you want to lock rotation/orientation?

I will have a look at that. Don't merge in the meantime.

Issue #3540, opened 2015-02-16
Unclean exits on Linux


emarcotte:
I'm running fedora 21 and every time I exit my game something is becoming unhappy and exiting unclean. When I run my program I get output indicating that X wasn't cleaned up properly, as well:

```
[emarcotte@localhost test-crash]$ mono bin/Linux/AnyCPU/Debug/test-crash.exe && echo "Exited cleanly";
Tear down
X Error of failed request:  BadCursor (invalid Cursor parameter)
  Major opcode of failed request:  95 (X_FreeCursor)
  Resource id in failed request:  0x3600005
  Serial number of failed request:  164
  Current serial number in output stream:  166
AL lib: (EE) alc_cleanup: 1 device not closed
[emarcotte@localhost test-crash]$ 
```

I have a link to a sample project reproducing this below, but here'a a paste of the complete source that can reproduce this:

```c#
namespace TestCrash
{
    using System;
    using Microsoft.Xna.Framework;

    public class Crasher : Game
    {
        GraphicsDeviceManager graphics;

        public Crasher()
        {
            graphics = new GraphicsDeviceManager(this);
            Content.RootDirectory = "Content";
        }

        protected override void Update(GameTime gameTime)
        {
            Exit();
        }
    }

    public static class Program
    {
        [STAThread]
        static void Main()
        {
            using (var game = new Crasher()) {
                game.Run();
                Console.WriteLine("Tear down");
            }
            Console.WriteLine("Bye!");
        }
    }
}
```

I traced it as far as the `Platform.Dispose()` call in Game.cs. It's never returning out of `OpenTKGamePlatform.Dispose()`. I'm planning on continuing to debug this but if someone has any quick pointers or thoughts on what I might be missing I'd appreciate it.

(See attached sample code: https://www.dropbox.com/s/18tukij2ff4kckx/exit-crasher.tar.bz2?dl=0)

cra0zy:
I can confirm that this bug does occur on my Linux machines as well.

tomspilman:
Maybe if @thefiddler has some time he can chime in here as he maintains OpenTK.  Short of that we don't have many people helping on the Linux side of things, so anything you can find out @emarcotte would be greatly appreciated.

emarcotte:
I dropped in the latest release version of OpenTK's dlls and things and the problem goes away. Is the thirdparty/dependencies repo out of sync with the latest release perhaps?

[...]

PR #3547, opened 2015-02-18
Fix for linux exit errors 2, 3, 4 from #3540

cra0zy:
Fix for linux exit errors 2, 3, 4 from #3540

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsResource.cs
+++ b/MonoGame.Framework/Graphics/GraphicsResource.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Xna.Framework.Graphics
         ~GraphicsResource()
         {
             // Pass false so the managed objects are not released
-            DisposeExit();
+            Dispose(false);
         }
 
         /// <summary>
@@ -115,14 +115,6 @@ namespace Microsoft.Xna.Framework.Graphics
                 disposed = true;
             }
         }
-
-		/// <summary>
-		/// The method that derived classes should override to implement disposing of managed and native resources in case if Dispose will not work when the game is exiting.
-		/// </summary>
-		protected virtual void DisposeExit()
-		{
-			Dispose (false);
-		}
 
 		public event EventHandler<EventArgs> Disposing;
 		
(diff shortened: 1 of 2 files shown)


KonajuGames:
With this type of change, we need to ensure that Dispose is correctly called and on the correct thread.  If Dispose is not called, the GC will call the finalizer which disposes of unmanaged resources on the GC thread.  That said, the previous code to run on the UI thread probably wouldn't have worked if called by the finalizer.

prollin:
I am not sure we need `GL.IsShader()`; those state query functions should be avoided since they could stall the GPU; Not a huge deal during resource creation/deletion but it can't hurt to remove superfluous code. MonoGame should already know it is a shader since it created it internally.

tomspilman:
> I am not sure we need `GL.IsShader()`

What is it even doing there?  How would our shader handle not be a shader?

We call `GL.IsShader()` in that file too.


> those state query functions should be avoided since they could stall the GPU

Which is extra bad if you are trying to load resources during gameplay.

cra0zy:
So if I understood everything correctly I should:
1) Remove GL.IsShadder, unneeded check
2) Setup 2 dispose events, one for normal disposing(with threads if necessary) and one thats used when the game is exiting

tomspilman:
> So if I understood everything correctly I should:

I'll leave that up to @prollin  and @KonajuGames to decide. :)

cra0zy:
@prollin @KonajuGames Is this ok now?

KonajuGames:
I'm not too keen on the addition of another virtual method in GraphicsResource to do a slightly different  Dispose path.  We could merge this as a temporary fix for the exit issues, but I would like to revisit it soon.

cra0zy:
What do you thing of doing something like: 
> Dispose(bool disposing, bool calledbyfinilizer)

Would that be better?

KonajuGames:
That's what the `disposing` parameter is for.  If `true`, it was called by the app via `Dispose()`.  If `false` it was called by the finalizer (so don't access other managed resources because they may or may not exist).

cra0zy:
@KonajuGames I think I found a solution that you are gonna like

tomspilman:
@cra0zy - So is this a case where no one called `Dispose()` on the shader before shutdown?

KonajuGames:
That looks a lot better.  You confirmed that works for you?

cra0zy:
Works great for me, 15 runs, none showed errors 2,3,4 on exit.

tomspilman:
Seems ok to me... your call to merge @KonajuGames .

KonajuGames:
Merging.  Thanks @cra0zy

Issue #1476, opened 2013-02-27
Test Framework Improvements


tomspilman:
The current way tests are setup is cumbersome for development:

 - It is hard to find a test for a class as it is so distant from the location where the code is.
 - The tests are not included in the platform solution files... most developers probably have no clue they exist.
 - The tests are in separate submodule making it impossible to commit a fix+test together.

I suggest we consider a few things:

 1. ~~Consider moving at least the basic unit tests into the MonoGame.Framework folder itself:  Ray.cs and Ray.Test.cs.  This probably would only be beneficial if using wildcards in the test project.~~  DONE

 2. ~~Move the Tests folder into the main repository removing the old submodule.  Move the test art ("Tests/Assets" folder) into a new submodule.  Move the binaries in "Tests/Libs" into the existing thirdparty libs.~~ DONE

 3. ~~Get the tests into the automated build ASAP.  At the very least on Windows~~ and Mac.  If any tests fail we should fix them or disable them.

 4. Start making unit tests a requirement for all changes to easily testable classes like math.

 5. ~~Add unit tests for the 2MGFX tool.~~

tgjones:
I've started to have a look at the WindowsGL tests too. Lots of them hit this assertion:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/States/SamplerState.OpenGL.cs#L36

If you ignore the assertion, the tests continue and pass. Is that a known issue?

tomspilman:
> Lots of them hit this assertion

That is the issue where the static render states like `SamplerState` and `BlendState` don't get reset correctly after the `GraphicsDevice` shuts down.

I think pushing thru on your work in https://github.com/mono/MonoGame/pull/3472 should move is towards fixing this issue.

tgjones:
Oops - I forgot that `SamplerState` is one of the ones affected by that :)

(Talking of which - I'm holding fire until there's a decision in #3472. If we decide to have fewer reference images, then I'll modify my last commit.)

tomspilman:
This is pretty much done...  anything that remains should be opened as new issues.

PR #3548, opened 2015-02-18
Added test cases for empty XML datatypes

DDReaper:
Updated deserialisation to support empty XML elements for all datatypes

E.g. 
```xml
<Vector2/>
<float/>
```
 and 
```xml
<Vector2></Vector2>
<float></float>
```

Tested all data type variants that have intermediate serialisers

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/CharSerializer.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/CharSerializer.cs
@@ -17,21 +17,20 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
 
         protected internal override char Deserialize(string[] inputs, ref int index)
         {
-            if (inputs.Length > 0)
-            {
-                var str = inputs[index++];
-                if (str.Length == 1)
-                    return XmlConvert.ToChar(str);
+            if (inputs.Length == 0)
+                return new char();
 
-                // Try parsing it as a UTF code.
-                int val;
-                if (int.TryParse(str, out val))
-                    return char.ConvertFromUtf32(val)[0];
+            var str = inputs[index++];
+            if (str.Length == 1)
+                return XmlConvert.ToChar(str);
 
-                // Last ditch effort to decode it as XML escape value.
-                return XmlConvert.ToChar(XmlConvert.DecodeName(str));
-            }
-            return new char();
+            // Try parsing it as a UTF code.
+            int val;
+            if (int.TryParse(str, out val))
+                return char.ConvertFromUtf32(val)[0];
+
+            // Last ditch effort to decode it as XML escape value.
+            return XmlConvert.ToChar(XmlConvert.DecodeName(str));
         }
 
         protected internal override void Serialize(char value, List<string> results)
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/CharSerializer.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/CharSerializer.cs
@@ -17,17 +17,21 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
 
         protected internal override char Deserialize(string[] inputs, ref int index)
         {
-            var str = inputs[index++];
-            if (str.Length == 1)
-                return XmlConvert.ToChar(str);
+            if (inputs.Length > 0)
+            {
+                var str = inputs[index++];
+                if (str.Length == 1)
+                    return XmlConvert.ToChar(str);
 
-            // Try parsing it as a UTF code.
-            int val;
-            if (int.TryParse(str, out val))
-                return char.ConvertFromUtf32(val)[0];
+                // Try parsing it as a UTF code.
+                int val;
+                if (int.TryParse(str, out val))
+                    return char.ConvertFromUtf32(val)[0];
 
-            // Last ditch effort to decode it as XML escape value.
-            return XmlConvert.ToChar(XmlConvert.DecodeName(str));
+                // Last ditch effort to decode it as XML escape value.
+                return XmlConvert.ToChar(XmlConvert.DecodeName(str));
+            }
+            return new char();
         }
 
         protected internal override void Serialize(char value, List<string> results)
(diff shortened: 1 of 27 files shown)


MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/CharSerializer.cs
```diff
@@ -17,17 +17,21 @@ class CharSerializer : ElementSerializer<char>
 
         protected internal override char Deserialize(string[] inputs, ref int index)
         {
-            var str = inputs[index++];
-            if (str.Length == 1)
-                return XmlConvert.ToChar(str);
+            if (inputs.Length > 0)
```
tomspilman: The PR would be much cleaner if you would have simply added:
```
if (inputs.Length == 0)
  return new char();
```
Instead of wrapping the existing code in an `if` block.

DDReaper: Do you want me to go back and change it @tomspilman ?
I'm fine either way

Test/ContentPipeline/IntermediateDeserializerTest.cs
```diff
@@ -537,5 +537,16 @@ public void GenericTypes()
...
         }
+
+        [Test]
+        public void DeserialiseDefaults()
+        {
+            DeserializeCompileAndLoad<DeserialiseDefaults>("25_DeserialiseDefaults.xml", deserialiseDefaults =>
+            {
+                Assert.AreEqual(Vector2.Zero, deserialiseDefaults.EmptyVector2);
+                Assert.AreEqual(Vector2.Zero, deserialiseDefaults.Empty2Vector2);
+
```
tomspilman: Are we not missing a ton of asserts here?

tomspilman:
@DDReaper - Thanks for the cleanup!

Test/ContentPipeline/IntermediateDeserializerTest.cs
DDReaper: Shouldn't need them but I agree, should asset everything.
The main point of the test is if the actual XML can be de-serialised.
But I'll add them now.

tomspilman: Yeah... we do asserts not only to find current bugs, but to catch future ones.

Like if someone decides to return the wrong default for `Matrix`.

tomspilman:
@tgjones - Give this a look and if you are happy then i'll merge it!

tgjones:
Hmm, actually this doesn't match XNA's behaviour. @DDReaper, if you run these tests under XNA (by manually adding the MonoGame.Tests.XNA project to your solution), they will all fail, except for `EmptyString`.

I had thought after our conversation yesterday, that XNA distinguished between `<Vector2 />` and `<Vector2></Vector2>`, but I guess it doesn't. It doesn't seem to allow either form of empty element for maths types. Are you sure the XML files you're testing with worked in the XNA content pipeline?

(@tomspilman this brings up the related-but-slightly-off-topic question of "when are we going to add the XNA test project to Protobuild / TeamCity"? Now that #3522 is merged, the XNA tests run fine, for me anyway. Probably #1476 is a better place for that discussion, though.)

DDReaper:
Yes @tgjones I can confirm that these asset files work under XNA.

The project I am working on is to port an existing published XNA title to MonoGame.
I'd rather not have to go changing the content XML unless there is a good reason to.
Pretty sure this was existing behaviour in XNA from way back when, so not sure what mg.XNA.Tests is checking. Where would I find that?

In regards to <vector2/> & <vector2></vector2> that was simply behaviour I noted that the xml serialization handled differently even both should mean the same thing

tgjones:
`MonoGame.Tests.XNA.csproj` is in the `Tests` folder. I use ReSharper, and it sometimes gets confused about whether it's running tests from `MonoGame.Tests.XNA` or `MonoGame.Tests.Windows`, but you can usually persuade it to test the right thing eventually.

tomspilman:
> when are we going to add the XNA test project to Protobuild / TeamCity

@tgjones - I'll look into it a bit now and see what I can do.  My two concerns...  being able to tell what platform the test failed on easily... how much will it impact build times.

DDReaper:
@tomspilman If the XNA.Tests project is used, it should have it's own protobuild definition anyway. but I agree, as it's using the same test classes it will be had to tell them apart unless they are part of a separate test run.

With regards to this PR and after talking to @tgjones I've decided to pull it.

Reason being the correct way to support this is to use:
```xml
    <vector2 Null="true" />
```

Even though XNA did support having empty math types (even though the XNA tests say otherwise, which is odd) it should not be a supported path.  If it is supposed to be null, it should be set null (as shown above.

I won't delete the branch on my repo in case someone comes up with a damn good reason why it should be supported.

tomspilman:
> Even though XNA did support having empty math types 

Is this just something you know from memory?  Or is there any example of this working out there?


> in case someone comes up with a damn good reason why it should be supported.

I don't mind supporting extra features when they won't screw up behavior on existing XNA compatible XML.  It just depends on how useful it is.

DDReaper:
> Is this just something you know from memory? Or is there any example of this working out there?

The reason these XML issues were raised @tomspilman was because I picked up another existing XNA project from a team to port over to Monogame.
Their content highlighted some gaps in MG content at present.

@tgjones has resolved most of the rest. This last remaining one I can work around using the above XML fix.

Last issue is the .TGA file. However as it's one file I may just replace it.

KonajuGames:
What is `<vector2 />` or `<vector2 Null="True" />` supposed to represent?

DDReaper:
An element in XML with no value (or use default)

if you template XML files for a certain class, some properties may be null (of the data type supports it) or set to use default. Also makes the XML smaller so you don't have to do, for example:
```xml
<matrix> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 </matrix>
```
or in the test case I found in the project I was porting for someone
```xml
<vector2></vector2>
```
However it seems using the following works already without changing MG
```XML
<vector2 Null="True" />
```
It's just not as short or compressed.

tomspilman:
> in the test case I found in the project I was porting for someone
So does...
```
<vector2></vector2>
```
... work under XNA?

tgjones:
> However it seems using the following works already without changing MG
> `<vector2 Null="True" />`

@DDReaper Is the C# property you're deserializing into a nullable `Vector2?` or just `Vector2`.

Although even then, I've just tested these cases, and XNA (at least the version of XNA I have installed) doesn't allow any of them:

* Deserializing `<vector2 />` into `Vector2`
* Deserializing `<vector2></vector2>` into `Vector2`
* Deserializing `<vector2 />` into `Vector2?`
* Deserializing `<vector2></vector2>` into `Vector2?`

So far, the only cases I've been able to get working with XNA are the expected ones:

* Deserializing `<vector2>0 0</vector2>` into `Vector2`
* Deserializing `<vector2 Null="true" />` into `Vector2?`

PR #3550, opened 2015-02-19
Run XNA Unit Tests

tomspilman:
This PR enables building and running the XNA unit tests on Windows.

### DIFF
--- a/default.build
+++ b/default.build
@@ -62,6 +62,7 @@
       <exec program="Protobuild" commandline="-generate WindowsGL" />
       <exec program="msbuild " commandline="MonoGame.Framework.WindowsGL.sln /t:Clean /p:Configuration=Release" />
       <exec program="msbuild " commandline="MonoGame.Framework.WindowsGL.sln /t:Build /p:Configuration=Release" />
+      <exec program="msbuild " commandline="Test/MonoGame.Tests.XNA.csproj /t:Build /p:Configuration=Release" />
     </if>
     <if test="${os == 'Unix'}">
       <exec program="mono" commandline="Protobuild.exe -generate Windows" />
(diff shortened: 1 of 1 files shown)


tomspilman:
@tgjones - Here we go...

tomspilman:
So the first issue was it was trying to use a x64 nunit test runner with the strictly x86 XNA unit tests.

tomspilman:
I vaguely remember this being a permissions issue... so I disabled spawning a new process for each assembly when unit testing.  Crossing fingers.

tgjones:
Looking good! (apart from the failing tests)

Any ideas what might be causing the tests to fail? `Unspecified error` isn't exactly descriptive - maybe we need to temporarily use the debug build or something so we can get the full error message?

tomspilman:
> Any ideas what might be causing the tests to fail?  

Something is screwed up on the build server.  I will be revisiting this soon.

tomspilman:
I'm going to revisit this now that we've replaced that old build server and have a much faster build.

Tzenchor:
This may help to set up the build server without installing visual studio.
If Windows version is Windows 8 or newer may require [this workaround](http://blogs.msdn.com/b/astebner/archive/2012/02/29/10274694.aspx).
To install XNA avoiding Visual Studio vinculations follow [this guide](http://blogs.msdn.com/b/astebner/archive/2009/06/12/9740290.aspx).
To build the project use the "XNA Game Studio Command Prompt" and from the "test" folder run "msbuild MonoGame.Tests.XNA.csproj".
This worked a few minutes ago in this computer with Windows 10 Pro.

tomspilman:
Thanks @Tzenchor ...  i'll give it a try this weekend.

Tzenchor:
One thing I found today that may be useful. To build a XNA content project the command is
`msbuild /p:XNAContentPipelineTargetPlatform=Windows /p:XNAContentPipelineTargetProfile=Reach` if the command prompt is in the folder that contains the `*.contentproj` file

tomspilman:
Still fighting getting XNA running on the build server headless.

It is close...  but the first graphical test it will pop up a dialog saying it couldn't find a graphics device.  It I log into the server using remote desktop and "Ok" the dialog it proceeds to run all the graphical tests just fine.

What I suspect currently is that the stupid WM_QUIT bug is causing this:

https://github.com/mono/MonoGame/blob/develop/Test/Framework/TestGameBase.cs#L180

PR #3551, opened 2015-02-19
New docs and tests

Shqrdx:
Docs and tests for

- FillMode
- GraphicsAdapter.DriverType
- GraphicsDeviceStatus
- GraphicsProfile
- IndexElementSize

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.cs
@@ -19,13 +19,25 @@ namespace Microsoft.Xna.Framework.Graphics
 {
     public sealed class GraphicsAdapter : IDisposable
     {
-
+        /// <summary>
+        /// Defines the driver type for graphics adapter.
+        /// </summary>
         public enum DriverType
         {
+            /// <summary>
+            /// Hardware device been used for rendering. Maximum speed and performance.
+            /// </summary>
             Hardware,
+            /// <summary>
+            /// Emulates the hardware device on CPU. Slowly, only for testing.
+            /// </summary>
             Reference,
+            /// <summary>
+            /// Useful when <see cref="DriverType.Hardware"/> acceleration does not work.
+            /// </summary>
             FastSoftware
         }
+       
         private static ReadOnlyCollection<GraphicsAdapter> _adapters;
 
         private DisplayModeCollection _supportedDisplayModes;
@@ -52,7 +64,7 @@ namespace Microsoft.Xna.Framework.Graphics
         public void Dispose()
         {
         }
-
+            
         public DisplayMode CurrentDisplayMode
         {
             get
@@ -132,14 +144,11 @@ namespace Microsoft.Xna.Framework.Graphics
         }
 
         /// <summary>
-        /// Used to request creation of a specific kind of driver
-        /// Hardware is preferred most of times for performance. This is the default value.
-        /// Reference is useful for testing.
-        /// FastSoftware is useful when Hardware acceleration does not work.
+        /// Used to request creation of a specific kind of driver.
         /// </summary>
         /// <remarks>
         /// These values only work on DirectX platforms and must be defined before the graphics device
-        /// is created.
+        /// is created. <see cref="DriverType.Hardware"/> by default.
         /// </remarks>
         public static DriverType UseDriverType { get; set; }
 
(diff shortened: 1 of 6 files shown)


MonoGame.Framework/Graphics/GraphicsAdapter.cs
```diff
@@ -19,13 +19,25 @@ namespace Microsoft.Xna.Framework.Graphics
...
+            /// Hardware device been used for rendering. Maximum speed and performance.
+            /// </summary>
             Hardware,
+            /// <summary>
+            /// Emulates the hardware device on CPU. Slowly, only for testing.
+            /// </summary>
             Reference,
+            /// <summary>
+            /// Useful when <see cref="DriverType.Hardware"/> acceleration does not work.
+            /// </summary>
```
tomspilman: We should note that `Reference` and `FastSoftware` only work on DirectX platforms at the moment.  Not sure if we're better off listing specific platforms it works on like "Windows desktop and Windows Store".

Don't think those work on Windows Phone... someone needs to check.

MonoGame.Framework/Graphics/GraphicsProfile.cs
```diff
@@ -1,49 +1,18 @@
...
+// file 'LICENSE.txt', which is part of this source code package.
+
 namespace Microsoft.Xna.Framework.Graphics
 {
 	public enum GraphicsProfile
 	{
-        Reach =  0, // 	Use a limited set of graphic features and capabilities, allowing the game to support the widest variety of devices, including all Windows-based computers and Windows Phone.
-        HiDef =  1, // 	Use the largest available set of graphic features and capabilities to target devices, such as an Xbox 360 console and a Windows-based computer, that have more enhanced graphic capabilities.        
+        /// <summary>
+        /// Use a limited set of graphic features and capabilities, allowing the game to support the widest variety of devices, including all Windows-based computers and Windows Phone.
```
tomspilman: We shouldn't talk about specific platforms here, the "Windows-based computers and Windows Phone" part.  Instead we should say it works on all MonoGame target platforms.

MonoGame.Framework/Graphics/GraphicsProfile.cs
```diff
@@ -1,49 +1,18 @@
...
+        /// </summary>
+        Reach,
+        /// <summary>
+        /// Use the largest available set of graphic features and capabilities to target devices, such as an Xbox 360 console and a Windows-based computer, that have more enhanced graphic capabilities.
```
tomspilman: Again... we shouldn't mention Xbox or Windows here.

MonoGame.Framework/Graphics/States/FillMode.cs
```diff
@@ -1,8 +1,21 @@
...
+    /// Defines options for filling the primitive.
+    /// </summary>
     public enum FillMode
     {
+        /// <summary>
+        /// Draw solid faces for each primitive.
+        /// </summary>
         Solid,
+        /// <summary>
+        /// Draw lines connecting the vertices that define a primitive face.
```
tomspilman: This line is identical to the XNA docs on MSDN.  We cannot copy Microsoft's docs...  this is the same as copying their code... it is a copyright violation.

Shqrdx: Suggest your version.

MonoGame.Framework/Graphics/GraphicsAdapter.cs
Shqrdx: The only direct usage of this enum is Adapter.UseDriverType(and that described there), I doubt this is good idea to write it in enum description but if you wants..

MonoGame.Framework/Graphics/States/FillMode.cs
Shqrdx: @tomspilman Ok, i changed it myself.

MonoGame.Framework/Graphics/GraphicsAdapter.cs
Shqrdx: I changed DriverType description itself - added "Usable only on DirectX platforms for now." than all these members are meaning to support only on DirectX.

Shqrdx:
@tomspilman How it is now ?

tomspilman:
Good enough... merging!

PR #3552, opened 2015-02-19
Few MathHelper improvements

Shqrdx:
Changes:
- Min and Max existed method performance improvement
- New Min and Max overloads which accepts integer instead float, Clamp already have same overload so why not adding this ?
- Tests for Min and Max


tomspilman:
If you do....
```
var v = MathHelper.Max(0.1f, 0);
```
Does this confuse the compiler or does it correctly pick the floating point version?

Shqrdx:
> Does this confuse the compiler or does it correctly pick the floating point version?

Nope, this is correct.

tomspilman:
Ok...  this might cause an error in the XNA tests, but we'll see about that soon.

Issue #3549, opened 2015-02-19
Model Textures not displaying (just grey), some switched in latest build


tattries:
Many weird problems with textures not working on models  after installing recent build (platform is windows desktop). Everything worked fine two weeks ago. Wondering if it has something to do with a recent change in monogame regarding setting texture parameters to null. (3351) .

tgjones:
Looking at #3351 with fresh eyes, I'm thinking there's a bug when texture parameters aren't updated between calls to `EffectPass.Apply`, and you're switching between multiple effects. I'll add a test case for that.

tattries:
Additional note, the problem did start with build 3.0.0.02114.  Build 3.0.0.02113 works fine.






Sent from Windows Mail





From: Tim Jones
Sent: ‎Thursday‎, ‎February‎ ‎19‎, ‎2015 ‎12‎:‎16‎ ‎AM
To: mono/MonoGame
Cc: tattries





Looking at #3351 with fresh eyes, I'm thinking there's a bug when texture parameters aren't updated between calls to EffectPass.Apply, and you're switching between multiple effects. I'll add a test case for that.

tgjones:
I've setup some more tests to tease out XNA's behaviour around this, and I have a potential fix. I'll submit a PR now.

PR #3554, opened 2015-02-20
Fix logic for applying texture effect parameters

tgjones:
This essentially undoes #3385, and applies a much simpler fix.

The first 2 commits just add tests; the last of the 3 commits implements the fixes.

A couple of things to note:

* The `stateKey`-based system I added in #3385 to decide whether to apply textures from effect parameters wasn't correct. If you switched between multiple effects, but didn't explicitly call `textureParameter.SetValue` each time, textures wouldn't be re-applied. I think this probably caused #3549.
* If you are using a custom effect with `SpriteBatch`, and that custom effect uses a couple of textures, then XNA always sets the texture you specify in `spriteBatch.Draw` into the first texture slot, even if you have explicitly set that texture to something else in your effect parameters. This was the part I didn't understand before - but I've added a test for it now. Because of this, I can remove the `stateKey` stuff from `EffectPass`.

So this PR changes the logic to: **always use the textures specified in effect parameters (`null` or otherwise), but if you're using `SpriteBatch` then ensure `Textures[0]` is set to the texture you passed to `spriteBatch.Draw`, regardless of what the effect parameter values are.**

Once again, I'd appreciate feedback on this one. Especially after I got it wrong last time :) But if it looks good, then we should merge it as soon as possible, given that it (hopefully) fixes quite a big bug.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatcher.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatcher.cs
@@ -247,7 +247,7 @@ namespace Microsoft.Xna.Framework.Graphics
                     var shouldFlush = !ReferenceEquals(item.Texture, tex);
                     if (shouldFlush)
                     {
-                        FlushVertexArray(startIndex, index, effect);
+                        FlushVertexArray(startIndex, index, effect, tex);
 
                         tex = item.Texture;
                         startIndex = index = 0;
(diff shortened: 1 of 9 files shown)


tomspilman:
Something got stuck on the build server... restarting the build.

MonoGame.Framework/Graphics/SpriteBatcher.cs
```diff
@@ -292,8 +292,14 @@ private void FlushVertexArray(int start, int end, Effect effect)
                 var passes = effect.CurrentTechnique.Passes;
                 foreach (var pass in passes)
                 {
+                    // Whatever happens in pass.Apply, make sure the texture being drawn
+                    // ends up in Textures[0].
+                    var texture = _device.Textures[0];
```
tomspilman: Can you instead pass the current texture thru to `FlushVertexArray` so you have it here to begin with?

tomspilman:
The build machine seems to be having issues right now...  i'll have to reboot it in the morning.  Sorry for the delay on this.

tomspilman:
@tgjones - You have a bug in the WinGL build of the tests...

> Framework\Visual\SpriteBatchTest.cs(8,22): error CS0234: The type or namespace name 'ContentPipeline' does not exist in the namespace 'MonoGame.Tests' (are you missing an assembly reference?) [C:\TeamCity\buildAgent\work\f26e5a04a66fcbdb\Test\MonoGame.Tests.WindowsGL.csproj]

I think you need to go here...

  https://github.com/mono/MonoGame/blob/develop/Build/Projects/MonoGame.Tests.definition#L148

... and make sure WinGL is a platform for the effect utility.

tgjones:
`AssetTestUtility` depends on types from the content pipeline assembly, which isn't part of the WindowsGL solution because it depends on the WindowsDX version of the framework. I'm not sure what the correct solution is here?

tomspilman:
> I'm not sure what the correct solution is here?

I think because of the fact this is a critical fix...  the immediate solution is to make this test Windows DX only for now.

We should setup an issue to discuss how or if we need to support the pipeline on WinGL.

tgjones:
@tomspilman I think this is ready to go, if it looks okay to you.

tomspilman:
Merged!  This shows how important it is to have the right unit tests...  we can easily prove the wrong thing.

PR #3558, opened 2015-02-22
Implement TextureCube.GetData for DirectX

tgjones:
Most of the time you probably don't need this method, but when you do, you do...

(In my engine / editor, I'm rendering a skybox (either from a `TextureCube` or a procedural shader) into a `RenderTargetCube`, then using `TextureCube.GetData` to extract the rendered data and convert it to spherical harmonics, and finally using the spherical harmonics to render a directional ambient term that matches up nicely with the skybox.)


tomspilman:
Seems ok for a new feature...  we'll see if anyone runs into trouble with it.

PR #3560, opened 2015-02-22
React to resolution changes on Android (like immersive mode)

Nezz:
Previously we did not react to resolution changes at all


tomspilman:
Merging!

PR #3561, opened 2015-02-22
Handle backbutton hold correctly on Android

Nezz:
Don't fire back pressed again and again if user holds the button on Android


tomspilman:
Thanks!

PR #3564, opened 2015-02-23
Triangle strip GetElementCountArray fix

disportium:
The element count calculation for a triangle strip should be primitiveCount + 2.  A triangle is only drawn once three vertices are specified; thereafter an additional triangle is drawn for every additional vertex.

1 triangle    =>  3 vertices
2 triangles  =>  4 vertices
3 triangles  =>  5 vertices
...

thus vertex count = triangle count + 2

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -942,7 +942,6 @@ namespace Microsoft.Xna.Framework.Graphics
 
         private static int GetElementCountArray(PrimitiveType primitiveType, int primitiveCount)
         {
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks...  the math there was unnecessarily complex.

disportium:
In retrospect I see that the first operator was actually a plus: `3 + (primitiveCount - 1);`.  Thought it was `3 * (primitiveCount - 1);`

Makes this PR a bit more trivial than I thought!

KonajuGames:
I wouldn't be surprised if the compiler collapsed `3 + (primitiveCount -1)` down to `primitiveCount + 2` anyway. :)

PR #3567, opened 2015-02-24
Fixed resolution problems on Android when resuming from portrait

Nezz:
When the game is resumed from a portrait orientation it may receive a portrait surface at first. If the game does not support portrait we should ignore it because we will receive the landscape surface a moment later.

### DIFF
--- a/MonoGame.Framework/Android/MonoGameAndroidGameView.cs
+++ b/MonoGame.Framework/Android/MonoGameAndroidGameView.cs
@@ -94,6 +94,11 @@ namespace Microsoft.Xna.Framework
                 }
             }
 
+            // When the game is resumed from a portrait orientation it may receive a portrait surface at first.
+            // If the game does not support portrait we should ignore it because we will receive the landscape surface a moment later.
+            if (width < height && (_game.graphicsDeviceManager.SupportedOrientations & DisplayOrientation.Portrait) == 0)
+                return;
+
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Android/MonoGameAndroidGameView.cs
```diff
@@ -94,6 +94,11 @@ void ISurfaceHolderCallback.SurfaceChanged(ISurfaceHolder holder, Android.Graphi
                 }
             }
 
+            // When the game is resumed from a portrait orientation it may receive a portrait surface at first.
+            // If the game does not support portrait we should ignore it because we will receive the landscape surface a moment later.
+            if (width < height && (_game.graphicsDeviceManager.SupportedOrientations & DisplayOrientation.Portrait) == 0)
```
tomspilman: Can the opposite happen?  Getting a landscape surface change event first, but the game does not support landscape?

Nezz: I haven't managed to make that happen. Phones have a rotation lock to portrait, tablets have no rotation lock. Maybe with a tablet that has its homescreen locked to landscape.

tomspilman: Probably best to wait till we see a case or report of a problem then.

PR #3568, opened 2015-02-24
Fixed WinRT MediaLibrary cache

Nezz:
The bug was that we read the cache from the file, then wrote into it, but if the new file is shorter than the old one then the cache file did not become shorter and there were some leftovers which failed to read.


tomspilman:
Looks good... merging!

PR #3579, opened 2015-02-27
Added monodevelop like add file dialog to Gtk Pipeline

cra0zy:
addfiledialog

Notes:
<ul>
<li>Fixed undoing "add file" action</li>
<li>When adding a file this dialog will show up</li>
<li>Windows Pipeline will act like usual</li>
<li>Undo/Redo can undo adding a file, but will not delete copied files, I leave that to the user.</li>
</ul>


tomspilman:
@cra0zy - I fixed your image link.

tomspilman:
> Undo/Redo can undo adding a file, but will not delete copied files, I leave that to the user.

That seems reasonable.

Tools/Pipeline/Windows/MainView.cs
```diff
@@ -901,5 +901,12 @@ public void ItemExistanceChanged(IProjectItem item)
                 }
             }
         }
+
+        public bool CopyOrLink(string file, bool exists, out int def, out bool applyforall)
+        {
+            def = 2;
+            applyforall = true;
+            return true;
```
tomspilman: @cra0zy - So this is where someone would hook in the dialog for Windows?

What is `def`?  Can you rename that to something more obvious?

tomspilman:
I think we had an issue for this somewhere... we should find and link it.

Tools/Pipeline/Windows/MainView.cs
cra0zy: I'll rename it, def is default

cra0zy:
Thanks for fixing the image link!!!

Tools/Pipeline/Windows/MainView.cs
tomspilman: Default what?  Default result?  Default button?

cra0zy:
@tomspilman I think now it's much better

tomspilman:
Yep... much clearer.

tomspilman:
Looks good... going to merge this.  Just need someone to follow up with the implementation for WinForms.

PR #3587, opened 2015-03-03
Updating all the Project templates to have the same MG styled Game page template

DDReaper:
Updated the Game.cs in all templates to use @theZMan 's fantastic new MG theme :D

All templates now have the same start point, with only a few platform variances for resolution or platform specific requirements (like Ouya)

### DIFF
--- a/ProjectTemplates/MonoDevelop/MonoDevelop.MonoGame/MonoDevelop.MonoGame/templates/MonoGameApplication/Game1.cs
+++ b/ProjectTemplates/MonoDevelop/MonoDevelop.MonoGame/MonoDevelop.MonoGame/templates/MonoGameApplication/Game1.cs
@@ -72,7 +72,12 @@ namespace ${Namespace}
         /// <param name="gameTime">Provides a snapshot of timing values.</param>
         protected override void Draw(GameTime gameTime)
         {
-           	graphics.GraphicsDevice.Clear(Color.CornflowerBlue);
+			//Welcome to MonoGame - Feel free to remove this line or replace with "GraphicsDevice.Clear(Color.Black);"
+            GraphicsDevice.Clear(
+               new Color(
+                   Vector3.Lerp(Color.CornflowerBlue.ToVector3(),
+                                new Vector3(0.9059f, 0.2353f, 0f),
+                                MathHelper.Clamp((float)gameTime.TotalGameTime.TotalSeconds / 2f, 0f, 1f))));
 		
             //TODO: Add your drawing code here
             
--- a/ProjectTemplates/VisualStudio2010/Android/Game1.cs
+++ b/ProjectTemplates/VisualStudio2010/Android/Game1.cs
@@ -78,7 +78,12 @@ namespace $safeprojectname$
         /// <param name="gameTime">Provides a snapshot of timing values.</param>
         protected override void Draw(GameTime gameTime)
         {
-            GraphicsDevice.Clear(Color.CornflowerBlue);
+			//Welcome to MonoGame - Feel free to remove this line or replace with "GraphicsDevice.Clear(Color.Black);"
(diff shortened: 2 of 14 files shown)


tomspilman:
@KonajuGames  @dellis1972 - How do you feel about this lerping color in the templates?

dellis1972:
It looks great, I like the fade from the xna cornflower to the mg color. 
That said will it be confusing for people new to XNA/MG.... 

Side note  : We should probably add a Color.MonoGameRed/Orange :P

Shqrdx:
@DDReaper You could use Color.Lerp instead Vector3.Lerp - it is more compact and clever. If my addition is accepted you could use Color.MonoGame instead "new Vector3(0.9059f, 0.2353f, 0f)".

DDReaper:
Was talking with Dom and a better name for MonoGame orange would be "Clockwork" 
What do you think guys?

Shqrdx:
> Was talking with Dom

You mean Tom ?

DDReaper:
Rofl, no, I meant Dominique Louis 😆 @softsavage

theZMan:
Hilarious but Color.Clockwork doesn't really follow the standards of the other color names so I think it wold be more confusing than MonoGameOrange

tomspilman:
As funny as you might think it is we don't need "Clockwork Orange" in MonoGame.

tomspilman:
So I checked out the effect today.  I personally feel like just making it MonoGameOrange is more than enough, but i'll let @KonajuGames make the call on that.

Regardless of what is decided we should use the new `Color.MonoGameOrange` constant as suggested in https://github.com/mono/MonoGame/issues/3584.

DDReaper:
Ok, if #3674 is merged, I'll update the templates to use that style instead

tomspilman:
@DDReaper - You can maybe try rebasing to squash the commits into a single one if that is your concern here.  I've never done it myself, but I know others have.

tomspilman:
So as a test a set this up today...

logo

It is just a modification of the spinning cube code from the existing Windows Universal templates.  I generated a 3D model of the extruded MonoGame logo and embedded the index and vertex data right into the code.   I also switched it to be implemented via a `GameComponent`...
```
        {
            Components.Add(new MonoGameLogo(this));
            base.Initialize();
        }
```
This leaves the `Game.cs` clean and the model can be removed via a single line comment when starting a new project.

It also serves to remind people MonoGame isn't just about 2D games.

Thoughts?

KonajuGames:
Looks great. All self contained which is good. Perhaps add a comment above
the creation of the component to comment/remove it when staying the project.

dellis1972:
Can we put that logo component in the framework :P

Shqrdx:
Yeah with cube,sphere and teapot please : )

tomspilman:
> Can we put that logo component in the framework :P

Yeah... we could I guess.  At least then it is always available for demos and such.

> Yeah with cube,sphere and teapot please : )

My original idea was to make this component into a VS template item.  If we wanted to provide premade primitives for rendering maybe we offer that way.

theZMan:
While its cool to have it in a API adding it to the template is just code people have to delete every time... total waste IMO.

tomspilman:
> code people have to delete every time... total waste IMO.

I have to delete that unused `SpriteBatch` in XNA's template every time... still hasn't impacted my productivity.  :)

And when literally the removal is:

```
// Components.Add(new MonoGameLogo(this));
```

Then there really isn't a lot to complain about I don't think.

DDReaper:
Sounds like there is no need for the color changing startup code now then. I'll delete that PR

DDReaper:
So did you push your change @tomspilman ?
If so I can drop this PR and recreate it with the updated template if need be, let me know.

PR #3588, opened 2015-03-03
iOS KeyboardInput upgrade

Nezz:
Replaced the old and ugly full screen keyboard input with a modal that matches the style of the messagebox. It is a lot shorter in code.


tomspilman:
Thanks... merging!

PR #3600, opened 2015-03-09
SpriteBatch.DrawCount property

Shqrdx:
This is addition to SpriteBatch which is ulong value which will increase every DrawInternal call. Useful for debugging purposes and performance increasing. The value is unchecked so it is defended from overflow exception.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		Rectangle _tempRect = new Rectangle (0,0,0,0);
 		Vector2 _texCoordTL = new Vector2 (0,0);
 		Vector2 _texCoordBR = new Vector2 (0,0);
-        ulong _drawCount;
+        uint _drawCount;
 
         /// <summary>
         /// Gets the internal amount of draw calls which happened between <see cref="SpriteBatch.Begin"/> and <see cref="SpriteBatch.End"/>.
@@ -38,7 +38,7 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <remarks>
         /// String drawing will increase this value by the amount of characters in text.
         /// </remarks>
-        public ulong DrawCount { get { return _drawCount; } }
+        public uint DrawCount { get { return _drawCount; } }
 
         /// <summary>
         /// Creates a new instance of <see cref="SpriteBatch"/> class.
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -410,7 +410,7 @@ namespace Microsoft.Xna.Framework.Graphics
 			
 		    unchecked
 		    {
-                ++_drawCount;
+                _drawCount++;
 		    }
 			if (autoFlush)
 			{
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -1,3 +1,7 @@
+// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
 using System;
 using System.Text;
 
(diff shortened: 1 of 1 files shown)


Shqrdx:
I change it to uint because it is enough for any sort of drawing.

Shqrdx:
@tomspilman Your thoughts ?

KonajuGames:
This one appears to be the number of calls to Draw for a Begin/End pair.  Is that correct?

I think it is good to have some metrics rather than everyone adding their own set of similar metrics in each project, and is something I have given some thought to.  I was thinking of perhaps adding a `Metrics` static class that could be a collection of various metrics from across the framework.

Vault16Software:
Is SpriteBatch.DrawCount is the amount of draw calls between SpriteBatch.Begin() and SpriteBatch.End() in a single frame?

If so, that is really cool. I had my own implementation for that, but it was messy.

Shqrdx:
> This one appears to be the number of calls to Draw for a Begin/End pair. Is that correct? 

Yes, futhermore it can be used either inside or outside Begin/End for counting draw calls and even with Immediate flag. The result will be nullified when next Begin called.

Shqrdx:
> Is SpriteBatch.DrawCount is the amount of draw calls between SpriteBatch.Begin() and SpriteBatch.End() in a single frame?

Yes, in single frame. You can easy see output result like 
Window.Title = yourSpriteBatch.DrawCount.ToString();

Vault16Software:
That is really cool. Thanks ^_^

Shqrdx:
@KonajuGames > I was thinking of perhaps adding a Metrics static class that could be a collection of various metrics from across the framework.

This must be huge work and SpriteBatch is independent object.. What kind of data this class intended to use ?

Shqrdx:
@tomspilman We need your word.

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -394,6 +408,10 @@ void CheckValid(SpriteFont spriteFont, StringBuilder text)
 					_texCoordTL, 
 					_texCoordBR);			
 			
+		    unchecked
+		    {
+                ++_drawCount;
```
danzel: The whitespace used here is a bit of a mess.
Maybe ```_drawCount++``` instead? (There shouldn't be any difference between the two and it is nicer to read).

Sorry just nitpicking really :)

Shqrdx: the ++x notation is more native for processors than x++. I readed it in the Marc Gregoire book Professional C++. If @tomspilman or @KonajuGames says I change it like you said.

Shqrdx: However, maybe you right... this is nice looking and nobody knows how it is looking in C# assembly code.

KonajuGames: In this situation, it will make no difference in execution flow or performance either way.  In some situations in C++ it can be slightly better performance to have pre-increment (post-increment needs to make a copy of the value before incrementing and returning the copy), but for types such as integer it doesn't really matter.

Shqrdx: @KonajuGames Ok, I changed it to more nice looking code.

Shqrdx:
@tomspilman Please accept

prollin:
> I think it is good to have some metrics rather than everyone adding their own set of similar metrics in each project

That is the thing though, a lot of developer might prefer it done differently.

While this type of change is harmless and potentially useful for some ppl, it opens the door to more extra complexity that might not be needed by developers who prefer their own solution.
As I already mentioned a few times, a separate repository containing optional MonoGame extensions would certainly be the best place for this kind of addition (for extension methods, useful components and game services).

tomspilman:
This isn't even a real "draw" count in the GPU sense.  This is the number of sprite quads sent into batching.  The batching might result in any number of actual `GraphicsDevice.DrawXXX` calls... less when batched together, the same under immediate mode, or even more if the `Effect` has multiple passes.

I agree that having some simple and light weight metrics at the C# level can be useful.  Maybe we should discuss that and come up with what is needed and refactor this as part of it.

Shqrdx:
@tomspilman Should I change doc for new property, cancel this PR or leave it opened ?

tomspilman:
I think we should close it and maybe open a new issue like:  Implement Render Metrics.  

We can then all jump in and discuss a path forward for this idea because I think it would be useful to developers, but it needs a little more thought as to how to approach it.

Shqrdx:
Here we go - https://github.com/mono/MonoGame/issues/3624

PR #3601, opened 2015-03-09
Use PngWriter in Texture2D.PlatformSaveAsPng method on Windows, Mac and iOS

disportium:
Not sure whether to place this in the Texture2D.SaveAsPng method or the Texture2D.OpenGL.PlatformSaveAsPng method.  Opted for the latter.

### DIFF
--- a/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
@@ -5,6 +5,7 @@
 using System;
 using System.IO;
 using System.Runtime.InteropServices;
+using MonoGame.Utilities.Png;
 
 #if WINDOWS_PHONE
 using System.Threading;
@@ -268,9 +269,8 @@ namespace Microsoft.Xna.Framework.Graphics
 #if WINDOWS_STOREAPP
             SaveAsImage(BitmapEncoder.PngEncoderId, stream, width, height);
 #else
-            // TODO: We need to find a simple stand alone
-            // PNG encoder if we want to support this.
-            throw new NotImplementedException();
+            var pngWriter = new PngWriter();
+            pngWriter.Write(this, stream);
 #endif
         }
 
--- a/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
@@ -266,12 +266,8 @@ namespace Microsoft.Xna.Framework.Graphics
 
         private void PlatformSaveAsPng(Stream stream, int width, int height)
         {
-#if WINDOWS_STOREAPP
-            SaveAsImage(BitmapEncoder.PngEncoderId, stream, width, height);
-#else
             var pngWriter = new PngWriter();
             pngWriter.Write(this, stream);
-#endif
         }
 
 #if WINDOWS_STOREAPP
--- a/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
@@ -6,6 +6,7 @@ using System;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Drawing;
+using MonoGame.Utilities.Png;
 
 #if MONOMAC
 using MonoMac.AppKit;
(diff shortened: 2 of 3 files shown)


tomspilman:
@disportium 

So first why for iOS only to begin with?  Just the platform you are testing at the moment?

Also maybe you should start to implement support for multiple source formats like I outlined here:  https://github.com/mono/MonoGame/pull/3537#discussion_r24734078

disportium:
@tomspilman That's right, I'm only testing iOS at the moment, but I have access to Windows and Mac as well, so I'll test on those platforms and modify the compiler directives accordingly.

Will take a look at the multiple source formats issue.

disportium:
Texture surface format support has been added, but only for those surface formats that have classes in the Graphics.PackedVector namespace (Dxt1 and Rg32 don't have such classes, for instance).

PngWriter can also write images with alpha now.

The PlatformSaveAsPng methods of Texture2D.OpenGL and Texture2D.DirectX now use PngWriter for Windows, Mac and iOS.

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -268,9 +269,8 @@ private void PlatformSaveAsPng(Stream stream, int width, int height)
 #if WINDOWS_STOREAPP
             SaveAsImage(BitmapEncoder.PngEncoderId, stream, width, height);
```
tomspilman: If it works on one DX platform I suspect it should work on all.  So you can remove the `#if WINDOWS_STOREAPP` here.

MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
```diff
@@ -489,8 +490,9 @@ private void PlatformSaveAsJpeg(Stream stream, int width, int height)
 
         private void PlatformSaveAsPng(Stream stream, int width, int height)
         {
-#if MONOMAC || WINDOWS
-            SaveAsImage(stream, width, height, ImageFormat.Png);
+#if MONOMAC || WINDOWS || IOS
+            var pngWriter = new PngWriter();
+            pngWriter.Write(this, stream);
 #elif ANDROID
             SaveAsImage(stream, width, height, Bitmap.CompressFormat.Png);
```
tomspilman: What is missing for using this on Android?  Testing?

tomspilman:
Looking good!

> Dxt1 and Rg32 don't have such classes

`Rg32` is just an oversight.  I've added an issue for someone to implement it: https://github.com/mono/MonoGame/issues/3605

DXT i'm not sure if even XNA supports saving those.  I suppose we should test and see to know.  MonoGame does have a DXT decompressor here:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/DxtUtil.cs#L48

So it would be possible to support if we wanted.

MonoGame.Framework/Utilities/Png/PngWriter.cs
```diff
@@ -206,5 +203,98 @@ private int CalculateBytesPerPixel()
                     return -1;
             }
         }
+
+        private void GetColorData(Texture2D texture2D)
```
tomspilman: Note that we would want to lift this method out into a common location eventually.  So for example we could share it with the JPEG writer too.

MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
disportium: Yes, in theory it should work. Don't have access to any Android devices
though.

MonoGame.Framework/Utilities/Png/PngWriter.cs
disportium: Agreed, it's too general to be confined to `PngWriter`.  `Texture2D` seems like a good place for it eventually.

disportium:
Removed WINDOWS_STOREAPP directive in Texture2D.DirectX as requested.

tomspilman:
Thanks.... merging!

tomspilman:
@disportium - FYI.  The remaining packed vector types have been added as of https://github.com/mono/MonoGame/pull/3622.

disportium:
@tomspilman Cool - available in PR #3631

PR #3603, opened 2015-03-10
+Basic Drag&Drop functionality for MGCB

hede:
Have a look please.
Definitely not 100%.
Folders are not implemented for example.
But I didn't know how to handle them (e.g. not allow them at all?, go and add everything recursively? ..)

Anyway, this might be a basic starting point I think.

### DIFF
--- a/Tools/Pipeline/Common/IController.cs
+++ b/Tools/Pipeline/Common/IController.cs
@@ -102,7 +102,9 @@ namespace MonoGame.Tools.Pipeline
 
         void Include(string initialDirectory);
 
-        void Exclude(IEnumerable<ContentItem> items);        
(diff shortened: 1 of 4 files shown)


tomspilman:
Maybe there is some overlap here with the work @cra0zy did in https://github.com/mono/MonoGame/pull/3579 ?

cra0zy:
It does overlap with both #3579 and #3604. Also the reason I created those PRs was so I could setup a proper Drag & Drop support that can link/copy both files and folders.

PS. check #3579 branch name

hede:
Heh, yeah, looks that way.
Ok, I'll wait for cra0zy's changes to get merged and re-submit a PR with the WinForms events hooked to his PipelineController stuff.

PR #3606, opened 2015-03-11
BoundingFrustum Docs

cra0zy:


tomspilman:
Sorry, but I have to reject this PR.

The docs are word for word copies of the XNA docs for `BoundingFrustum`.  We cannot take docs from copyrighted sources the same way we cannot take code from copyrighted sources.

On a positive note this just gives us an opportunity to write better docs.  It isn't like the XNA docs were fantastic to begin with.

cra0zy:
Did not know that, anyway I was using BoundingFrustum for training with writing comments in c# since I am mostly working alone and never write them.

tomspilman:
No problem... you are not the first to miss that detail. :)

PR #3610, opened 2015-03-12
Linux Installer Update

cra0zy:
Things done:
<ul>
<li>Added automatic dependency installer, currently only for Ubuntu 14.04 and 14.10</li>
<li>Added better mime type support</li>
</ul>


tomspilman:
Thanks... merging!

PR #3612, opened 2015-03-15
Release 3.3

tomspilman:
This is a merge of the latest develop changes into master for the 3.3 release.


cra0zy:
@tomspilman Since #3361 isn't finished I haven't setup linux monogame installer to install the xs addin. My question is do you want me to create a modified version of the linux installer that includes the addin for 3.3 release(no PRs, I will send you the modified script or the compiled run file, depending on what you want)?

tomspilman:
>  do you want me to create a modified version of the linux installer that includes the addin for 3.3 release

No... I would say don't worry for now.  Focus on getting things ready for the next release which we should be able to do soon.  3.3 will just not have Linux or Mac installers.

cra0zy:
Ok, well anyway it would be nice if you could post about Linux stuff as well for MonoGame 3.3, like you did for Mac. Compiled MonoGame 3.3 stuff for Linux if you need it: https://mega.co.nz/#F!KJpWCCxK!-3eBWbra0p_bs1yNzaA0PA

tomspilman:
What is needed to get the Linux installer built on the build server?  

Can we not add this step to the installers build...

https://github.com/mono/MonoGame/blob/develop/Installers/default.build

?

cra0zy:
It needs to be built using either Linux, Mac or Cygwin, and all you have to do is cd to Installers/Linux/ and  execute generatelinuxinstaller.sh. The script is assuming that Pipeline Tool has already been compiled for Release.

tomspilman:
> The script is assuming that Pipeline Tool has already been compiled for Release.

What else it is assuming is built?  Or does this installer only contain the pipeline tool?

cra0zy:
The installer only installs the Pipeline Tool, so thats the only thing it needs built. It can also install all the needed dependencies(ie. libsdl1.2-mixer, monodevelop) that are required for developing monogame applications, it's optional, and currently only on Ubuntu 14.04/14.10.

PR #3613, opened 2015-03-16
Fixed touch capability on Windows 8

Nezz:
For some reason TouchPresent was zero on my Surface Pro 3 with the keyboard and a bluetooth mouse. Checking individual pointer devices works.


tomspilman:
I guess if you disconnected the keyboard and/or mouse it suddenly worked?

Nezz:
Yup.

PR #3615, opened 2015-03-16
Updated Assembly version number for dev release (post 3.3 release)

DDReaper:
Noticed (a little too late) that the assembly version on the branch wasn't updated on release.

Verified that Official release did version the dll's correctly, so this is just a tidy up

Updated to 3.3.1.0


tomspilman:
Those versions really don't matter.  The TeamCity build versions the builds completely on its own.

In fact I would like to change them to 0.0.0.0.  Then at least you know when someone was building from source on their own.

DDReaper:
Ok, I'll revise the pr and update it

tomspilman:
Thanks!

PR #3616, opened 2015-03-16
Post release fixed for NuSpec packages

DDReaper:
While publishing the 3.3 NuGets, a few tweaks were needed to the NuSpec definitions.

All tested with released 3.3 packages


tomspilman:
Ok... merging!

PR #3617, opened 2015-03-17
Removed Old Content Pipeline

cra0zy:
MonoGame 3.3 got released so I really wanted to sync up dependencies. Hopefully I did everything right for  #3390, tho I'll probably still have to fix some error that will occur.

PR #3626, opened 2015-03-18
Implement BorderColor texture address mode for Direct3D and OpenGL

### LINKED ISSUE #1952: Add support for border color clamp texture addressing
As mentioned here (https://monogame.codeplex.com/discussions/454253), it would be nice to have a border color clamp texture address mode.  XNA 3.0 had it, but XNA 4.0 got rid of it.  From the reading I did, it seemed they ditched it because the Windows Mobile platform didn't support it.  It would certainly be nice to have it back.

tgjones:
Fixes #1952.

@tomspilman wrote a helpful checklist in #1952, which I just went through and implemented:

* Add TextureAddressMode.Border enum value (at the end of the list).
* Add SamplerState.BorderColor property.
* Fix both the DX SamplerState.GetState() and the GL SamplerState.Activate() methods to generate a correct platform specific border color state.
* Add BorderColor sampler state parsing to 2MGFX.
* Increment the MGFX version, fix writer to write BorderColor sampler state into the file.
* Fix MGFX runtime code to read BorderColor sampler state.


tgjones:
Once #3617 is merged, I'll rebase this PR, and then the tests should pass.

tomspilman:
@tgjones - Note #3617 is merged now.

tgjones:
D'oh! Turns out there was more involved in bumping the MGFX version than I realised :)

tomspilman:
I guess we eventually should consider changing the pre-built effect XNBs in the unit tests to use the new helper for building them at runtime.  It would at least remove that extra work for people.

tomspilman:
Also...  what should we do when we hit a platform/hardware that doesn't support this feature?

Is this a documentation thing...  we need to alert people that the feature might not be available?  Should we throw and exception on use?  Should we tie this to some new feature levels?

tomspilman:
Looks like your other PR conflicted this one.  It needs a rebase.

Also the DX11 docs seem to indicate border color works on all PC GPUs pretty much.  So I expect this feature to work on all desktop systems.  It will be mobile and console where we may have troubles.  Let's leave that for discussion as part of https://github.com/mono/MonoGame/issues/1941.

tgjones:
> changing the pre-built effect XNBs in the unit tests to use the new helper for building them at runtime.

Yes - and on a related note, we also need to somehow allow the OpenGL tests to use the content pipeline.

tomspilman:
> we also need to somehow allow the OpenGL tests to use the content pipeline

Oh... yeah totally forgot about that issue.  It is a big one we should deal with first.

tgjones:
If the tests pass, this is ready to merge as far as I'm concerned.

tomspilman:
Looks great... merging!

PR #3628, opened 2015-03-18
Implement comparison samplers for Direct3D and OpenGL

tgjones:
Adds `SamplerState.ComparisonFunction`, and hooks it up for Direct3D and OpenGL.

I wasn't sure whether to call the new property `ComparisonFunction` or `CompareFunction` - I decided to go with the former, because that's how both Direct3D and OpenGL refer to it, even if it doesn't match the XNA / MG enum name.

Replaces #2088 (I'd have liked to cherry-pick those commits, but it was easier just to start from scratch).

This seems like a good use-case for `GraphicsProfile` - comparison samplers aren't supported by OpenGL ES 2.0 (without an extension), nor, I believe, Direct3D 11 feature level 9.x. We could throw an runtime exception if you try to use a comparison sampler. But I guess that is for the future - MG doesn't really use graphics profiles much yet, and whether it should or not is a whole separate discussion.

(This will conflict with my other PR, #3626 - I'll just do a rebase on whichever one isn't merged first.)

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsExtensions.cs
+++ b/MonoGame.Framework/Graphics/GraphicsExtensions.cs
@@ -438,6 +438,30 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		}
 
+        public static DepthFunction GetDepthFunction(this CompareFunction compare)
+        {
+            switch (compare)
+            {
+                default:
+                case CompareFunction.Always:
+                    return DepthFunction.Always;
+                case CompareFunction.Equal:
+                    return DepthFunction.Equal;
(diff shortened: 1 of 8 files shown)


tomspilman:
This seems safe.  From what I can tell from DX11 this seems to be a feature that works on all PC GPU hardware which is basically the same as all Mac and Linux hardware.  

The question will be if we run into mobile hardware or weird console hardware that doesn't.  What do we do then?

Probably an issue to leave to https://github.com/mono/MonoGame/issues/1941.

tomspilman:
Oh... totally forgot.  We do need some unit test that ensures this feature works.

PR #3629, opened 2015-03-18
MediaPlayer.WMS now changes stream volume

### LINKED ISSUE #2780: MediaPlayer.Volume affects all sounds
It seems like whenever I try to use MediaPlayer.Volume to change the volume of a background song it affects all sounds (sounds playing via SoundEffect). This is on Windows 7 using an unstable build I just downloaded from TeamCity. The stable 3.2 build does this too. My code works fine using XNA.

KonajuGames:
The previous iteration changed the application volume controller which affected all sound output including sound effects. It now uses the stream volume controller so it only affects the streaming media and sound effects remain unaffected.

Thanks to MarkoBL for the fix.

Fixes #2780


MonoGame.Framework/Media/MediaPlayer.WMS.cs
```diff
@@ -124,29 +137,39 @@ private static void PlatformPlaySong(Song song)
...
-            _volumeController.MasterVolume = _volume;
+            // Get the volume interface.
+            IntPtr volumeObj;
+            try
+            {
+                MediaFactory.GetService(_session, MRPolicyVolumeService, AudioStreamVolumeGuid, out volumeObj);
+            }
+            catch
+            {
+                MediaFactory.GetService(_session, MRPolicyVolumeService, AudioStreamVolumeGuid, out volumeObj);
```
tomspilman: Is this happening here again?  Where we need to get the service twice as it sometimes fails?

KonajuGames: It could be. @MarkoBL, is that correct?

tomspilman: @MarkoBL ?

MarkoBL:
I can't follow you exactly, what you're talking about. I'm not an XAudio expert, I only changed the ```SimpleAudioVolume``` to ```AudioStreamVolume```to fix the inital problem.

tomspilman:
@MarkoBL - The code you posted here:

 https://github.com/mono/MonoGame/issues/2780#issuecomment-53998018

Included this...
```
            try
            {
                MediaFactory.GetService(_session, MRPolicyVolumeService, AudioStreamVolumeGuid, out volumeObj);
            }
            catch
            {
                MediaFactory.GetService(_session, MRPolicyVolumeService, AudioStreamVolumeGuid, out volumeObj);
            }  
```
Did you add this try/catch and retry code because you saw this fail?  Or was this already in the code?

MarkoBL:
This was already in the code.

tomspilman:
@MarkoBL - Thanks for clearing that up!

MonoGame.Framework/Media/MediaPlayer.WMS.cs
tomspilman: So the code we already have that is similar to this yet improved:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Media/VideoPlayer.WMS.cs#L126

Just a different GUID, but same sort of issue.

@KonajuGames - Should we maybe call this from here too?

Spectrallic:
Is this already merged? If not, any idea when it will be?

MonoGame.Framework/Media/MediaPlayer.WMS.cs
tomspilman: @KonajuGames - Bump... for when you get back home. :)

tomspilman:
@Spectrallic - Soon... waiting on @KonajuGames for something.

MonoGame.Framework/Media/MediaPlayer.WMS.cs
tomspilman: @KonajuGames - Bump!

KonajuGames: I can add that in.

KonajuGames:
MediaPlayer and VideoPlayer are now consistent and correct when controlling volumes.

tomspilman:
Great... merging!

PR #3630, opened 2015-03-18
Pipeline now accepts uppercase extensions

KonajuGames:
Uppercase extensions foiled the importer auto-detect. File extension is now lower-cased for the detection process.

### DIFF
--- a/Tools/Pipeline/Common/PipelineTypes.cs
+++ b/Tools/Pipeline/Common/PipelineTypes.cs
@@ -355,9 +355,10 @@ namespace MonoGame.Tools.Pipeline
                 return null;
             }
 
+            var lowerFileExt = fileExtension.ToLowerInvariant();
             foreach (var i in Importers)
             {
-                if (i.FileExtensions.Contains(fileExtension.ToLowerInvariant()))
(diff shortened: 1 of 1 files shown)


Tools/Pipeline/Common/PipelineTypes.cs
```diff
@@ -357,7 +357,7 @@ public static ImporterTypeDescription FindImporter(string name, string fileExten
 
             foreach (var i in Importers)
             {
-                if (i.FileExtensions.Contains(fileExtension))
+                if (i.FileExtensions.Contains(fileExtension.ToLowerInvariant()))
```
tomspilman: How do we know `FileExtensions` is always lower case?  Do we force them to lowercase somewhere?

KonajuGames: Ahh, didn't think of that. I'll fix it up tonight.

tomspilman:
Looks good to me... merging.

tomspilman:
@KonajuGames - This reminds me that we probably should be starting to find ways to unit test even this sort of pipeline code.

PR #3622, opened 2015-03-17
Added few missing PackedVectors and Tests for them.

cra0zy:
#3605 @flibitijibibo is the one who wrote PackedVector classes, I've only modified documentation and added tests for them.

PR #3631, opened 2015-03-18
Added additional texture formats to PngWriter.GetColorData

disportium:
Alpha8, Bgra5551, Rg32, Rgba64 and Rgba1010102, as per #3622


tomspilman:
Thanks... merging!

PR #3407, opened 2015-01-16
Replaced sdl1.2 with opentk for gamepad and joystick

cra0zy:
I replaced all the sdl 1.2 code with opentk equivalent for joystick and gamepad, tho there is no way to get joystick name using opentk AFAIK. This PR should probably get tested a bit before merging.

PR #3119, opened 2014-10-25
Conversion of MonoMacGameView to OpenTK equivalent.

groundh0g:
OpenTK for `MONOMAC` platform. Now MacOS is a peer to `WINDOWS` and `LINUX` platforms. Common code for desktop platforms.

I think it's ready for review / testing. RE: #2154.

**NOTES:**

* OpenTK controller mappings can replace SDL on other desktop platforms. Didn't consider that in scope for this PR.
* Media / Video needs some love. I need to compare how the other desktop platforms interface to `GameWindow` and try to emulate.
* Three intentional `#warning DISABLED DURING MONOMAC OPENTK CONVERSION` messages were added.
 * One in `OpenTKGameWindow.cs` to bypass icon extraction from assembly
 * One in `MacOS/GamerServices/Guide.cs` for `Window.ClearKeyCacheState()` which may not be needed now that we're using a different game loop.
 * One in `Media/VideoPlayer.MacOS.cs` for `_game.Window.AddSubview(_currentVideo.MovieView);`. That's the one that need some love.

PR #3632, opened 2015-03-18
Replaced sdl1.2 with OpenTK for Joystick and Gamepad

cra0zy:
#3407 looked too ugly... and I still have to learn how to rebase. Also this PR provides Gamepad class from #3119, but a little bit modified.


tomspilman:
What platforms has this been tested on?

Also after we merge this, can you follow up with a second PR organizing:

  - `Axis`
  - `Capabilities`
  - `Input`
  - `PadConfig`
  - `Joystick`
  - `Stick`

These types seem to be internals to the OpenTK `GamePad` implementation.  I don't think we want them public.  We also probably want to move them into some other namespace like `Microsoft.Xna.Framework.Input.OpenTK` or something.

Build/Projects/FrameworkReferences.definition
```diff
@@ -26,6 +26,9 @@
       <Binary
         Name="Tao.Sdl"
         Path="ThirdParty/GamepadConfig/Tao.Sdl.dll" />
```
tomspilman: What is "Tao.Sdl.dll" used for here now?

cra0zy: sdl mixer, audio

cra0zy:
> What platforms has this been tested on?

I tested it on Linux and WindowsGL, and I know it works on Mac because of #3119.

> Also after we merge this, can you follow up with a second PR organizing:
> 
> Axis
> Capabilities
> Input
> PadConfig
> Joystick
> Stick
> These types seem to be internals to the OpenTK GamePad implementation. I don't think we want them public. We also probably want to move them into some other namespace like Microsoft.Xna.Framework.Input.OpenTK or something.

Gamepad uses OpenTK.Gamepad class, it doesn't use any of those classes. I thought it would be nice to leave those Joystick classes around, gives options for some lower level joystick implementation.

cra0zy:
Now that I think about it, maybe I should remove Joystick classes completely, GamePad.OpenTK doesn't use them, and Xna never had them.

Build/Projects/FrameworkReferences.definition
tomspilman: Ah... so we are waiting on https://github.com/mono/MonoGame/issues/3625.

tomspilman:
> Gamepad uses OpenTK.Gamepad class, it doesn't use any of those classes.

Well how about we unhook them from "MonoGame.Framework.definition" in this PR then?  This way they won't show up in the docs and won't confuse people for now.

Separately we can start some discussion on if we want to support a new API beyond GamePad for more complex input devices like many axis joysticks.

cra0zy:
Sounds good to me.

tomspilman:
Merging... thanks!

PR #3633, opened 2015-03-18
Linux installer Update

cra0zy:
Things done:
<ul>
<li>Added gtk linux so libs</li>
<li>Added libstdc++6 to dependency list</li>
<li>Added a check for libstdc++6 that checks if it supports GLIBCXX_3.4.20, if it doesn't it will offer to install custom version of it</li>
</ul>


tomspilman:
If this is all ready i'll merge it.

cra0zy:
It's ready.

PR #3636, opened 2015-03-19
Graphics interfaces docs

Shqrdx:
New docs for all Xna.Framework.Graphics interfaces.

I will release "inheritdocs" fixes for effects in next PR's..

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/IEffectFog.cs
+++ b/MonoGame.Framework/Graphics/Effect/IEffectFog.cs
@@ -1,44 +1,6 @@
-// #region License
-// /*
-// Microsoft Public License (Ms-PL)
-// MonoGame - Copyright © 2009 The MonoGame Team
-// 
-// All rights reserved.
-// 
-// This license governs use of the accompanying software. If you use the software, you accept this license. If you do not
-// accept the license, do not use the software.
-// 
-// 1. Definitions
-// The terms "reproduce," "reproduction," "derivative works," and "distribution" have the same meaning here as under 
-// U.S. copyright law.
-// 
-// A "contribution" is the original software, or any additions or changes to the software.
-// A "contributor" is any person that distributes its contribution under this license.
-// "Licensed patents" are a contributor's patent claims that read directly on its contribution.
-// 
-// 2. Grant of Rights
-// (A) Copyright Grant- Subject to the terms of this license, including the license conditions and limitations in section 3, 
-// each contributor grants you a non-exclusive, worldwide, royalty-free copyright license to reproduce its contribution, prepare derivative works of its contribution, and distribute its contribution or any derivative works that you create.
-// (B) Patent Grant- Subject to the terms of this license, including the license conditions and limitations in section 3, 
-// each contributor grants you a non-exclusive, worldwide, royalty-free license under its licensed patents to make, have made, use, sell, offer for sale, import, and/or otherwise dispose of its contribution in the software or derivative works of the contribution in the software.
-// 
-// 3. Conditions and Limitations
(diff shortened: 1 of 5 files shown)


MonoGame.Framework/Graphics/Effect/IEffectFog.cs
```diff
@@ -48,32 +10,31 @@ namespace Microsoft.Xna.Framework.Graphics
 	public interface IEffectFog
 	{
         /// <summary>
-        /// The floating point fog color.
+        /// Gets or set the fog color in floating point format.
```
tomspilman: Lets not use "gets or sets".  The programmer knows what a property is and that it can get or set.  That is not useful information to them.

KonajuGames: It is useful when you get a read-only property.  This is common in Microsoft documentation where a property might be described as "Gets" or "Gets and sets" to differentiate between read-only and read/write properties.

MonoGame.Framework/Graphics/Effect/IEffectLights.cs
```diff
@@ -73,9 +35,8 @@ public interface IEffectLights
 		bool LightingEnabled { get; set; }
 
         /// <summary>
-        /// Initializes the lights to the standard key/fill/back lighting rig.
+        /// Initializes the default lights.
```
tomspilman: I'm ok with making the interface doc a little more general...  it is possible some new `Effect` overloads this with some other default lighting rig.

However...  the information there is 100% accurate as to what the function does with the MonoGame stock effects (see http://blogs.msdn.com/b/shawnhar/archive/2007/04/09/the-standard-lighting-rig.aspx).

If you want to remove the information here then you need to add it back here:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/BasicEffect.cs#L406

Since the inherited doc from the interface will now be less accurate.

MonoGame.Framework/Graphics/Effect/IEffectFog.cs
tomspilman: > It is useful when you get a read-only property.

I'm fine with calling out a read only property saying "Returns" or "Gets" in those cases.  But why should we repeat the words "gets or sets" on most every other property that isn't read only?

Note that the doc tool and VS intellisense use an icon to show the accessibility of the getter/setter on properties.

tomspilman:
> I will release "inheritdocs" fixes for effects

What do you mean @Shqrdx ?

MonoGame.Framework/Graphics/Effect/IEffectFog.cs
KonajuGames: That's fine.  The extra "gets or sets" will be redundant because the vast majority of properties are read/write.

Shqrdx:
@tomspilman I mean I places inherit docs from these interfaces to subsequent classees.

MonoGame.Framework/Graphics/Effect/IEffectFog.cs
Shqrdx: The API docs needs unification. Wast majority of classes uses this notation, so I put it to the interface.

tomspilman: > Wast majority of classes uses this notation

A search shows that only 61 classes/structs/interfaces use "Gets or sets" in them.

tomspilman:
@Shqrdx - Can you please explain when you close a PR.

Shqrdx:
> Can you please explain when you close a PR.

I remake it without get/set's ?

tomspilman:
That would be great.

PR #3638, opened 2015-03-19
Loop length in XNB now matches XNA

### LINKED ISSUE #3507: WinDX SoundEffect/SoundEffectInstance stop playing after half of duration. (last MGdev)
I have this issue only on the WinDX (windows) version of my project, WinGL is playing fine.

Tried to use "windows8"/"windows",  "Mono wave"/"XNA wave", medium/best quality content build, it makes no difference.
Tracked for everything wrong in my code side, like a resume/stop, but nothing.

Could it be a SharpDX.Xaudio issue?

Thanks for any help.
Greg.

KonajuGames:
The XNB File Format document specifies the loop start and loop length are in bytes, but they are in samples.
Loop start and length are now populated in AudioContent when loading the sample data.
The loop start and loop length are no longer treated as bytes in XAudio playback.
XNA and MonoGame generated XNB files are now only different by one bit (1ms in duration).

Fixes #3507

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
@@ -281,6 +281,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Audio
                     channelCount,
                     format,
                     sampleRate);
+
+                // Loop start and length in number of samples. Defaults to entire sound
+                loopStart = 0;
+                loopLength = data.Count / ((bitsPerSample / 8) * channelCount);
             }
             finally
             {
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -154,11 +154,6 @@ namespace Microsoft.Xna.Framework.Audio
             _format = format;
             _dataStream = dataStream;
 
-            // Convert the loop points from bytes to samples.
-            var bytesPerSample = 2 * _format.Channels;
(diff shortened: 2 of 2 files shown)


MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
```diff
@@ -281,6 +281,10 @@ public void ConvertFormat(ConversionFormat formatType, ConversionQuality quality
                     channelCount,
                     format,
                     sampleRate);
+
+                // Loop start and length in number of samples. Defaults to entire sound
+                loopStart = 0;
+                loopLength = data.Count / ((bitsPerSample / 8) * channelCount);
```
tomspilman: Where was `loopLength` being set before?  Shouldn't we be removing that code?

KonajuGames: It wasn't being set at all.  It was defaulting to zero and never touched.

tomspilman: Then what was `loopLength /= bytesPerSample` in the runtime side doing?  Was `loopLength` 0 there too?

KonajuGames: It was returning zero, which I assume XAudio takes as meaning the whole file.  That is why the MonoGame-processed xnb appeared to work.

tomspilman: Ok...  I see now.  Merging!

PR #3641, opened 2015-03-20
Network class library added to docs generation.

Shqrdx:
I think it is a good idea.


tomspilman:
Yeah I think that would be ok too.

tomspilman:
I'm not sure what is going on there... i'll investigate a bit.

Shqrdx:
I better close this PR. You can implement it yourself easy and I dont want bother myself with such internal and hard things.

PR #3642, opened 2015-03-20
BoundingFrustum docs

Shqrdx:
Fully documented BoundingFrustum class.
-All public things are now documented
-All private fields are automatically renamed to private form with "_"(using resharper so it absolutly correct)
-Added Intersects overloads with rays(I dont know really how to get result differs from null and 0.0 so I leave this question opened). I also implemented few checks in Bounding.cs.
-PlaneCount is now opened const integer because why not ?
-ToString rewritted to more easy form.

### DIFF
--- a/MonoGame.Framework/BoundingFrustum.cs
+++ b/MonoGame.Framework/BoundingFrustum.cs
@@ -8,7 +8,7 @@ using System.Diagnostics;
 namespace Microsoft.Xna.Framework
 {
     /// <summary>
-    /// Defines a frustum for intersection operations.
+    /// Defines a viewing frustum for intersection operations.
     /// </summary>
     [DebuggerDisplay("{DebugDisplayString,nq}")]
     public class BoundingFrustum : IEquatable<BoundingFrustum>
@@ -24,12 +24,12 @@ namespace Microsoft.Xna.Framework
         #region Public Fields
 
         /// <summary>
-        /// Plane count(6).
+        /// The number of planes in the frustum.
         /// </summary>
         public const int PlaneCount = 6;
 
         /// <summary>
-        /// Corner count(8).
+        /// The number of corner points in the frustum.
         /// </summary>
         public const int CornerCount = 8;
 
@@ -38,7 +38,7 @@ namespace Microsoft.Xna.Framework
         #region Properties
 
         /// <summary>
-        /// Gets or sets the <see cref="Matrix"/> of this frustum.
+        /// Gets or sets the <see cref="Matrix"/> of the frustum.
         /// </summary>
         public Matrix Matrix
         {
@@ -52,7 +52,7 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
-        /// Gets the near plane of this frustum.
+        /// Gets the near plane of the frustum.
         /// </summary>
         public Plane Near
         {
@@ -60,7 +60,7 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
-        /// Gets the far plane of this frustum.
+        /// Gets the far plane of the frustum.
         /// </summary>
         public Plane Far
         {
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/BoundingFrustum.cs
```diff
@@ -4,103 +4,178 @@
 
 using System;
 using System.Diagnostics;
-using System.Text;
 
 namespace Microsoft.Xna.Framework
 {
+    /// <summary>
+    /// Defines a frustum for intersection operations.
```
tomspilman: I would try to get a little better definition about what a frustum is here and how it is build from a matrix.  Not something huge, but something a little more.

MonoGame.Framework/BoundingFrustum.cs
```diff
@@ -4,103 +4,178 @@
...
     public class BoundingFrustum : IEquatable<BoundingFrustum>
     {
         #region Private Fields
 
-        private Matrix matrix;
-        private readonly Vector3[] corners = new Vector3[CornerCount];
-        private readonly Plane[] planes = new Plane[PlaneCount];
+        private Matrix _matrix;
+        private readonly Vector3[] _corners = new Vector3[CornerCount];
+        private readonly Plane[] _planes = new Plane[PlaneCount];
```
tomspilman: I know how tempting it is to do these cleanups...  I really hate members without the leading `_` myself.  However you *really* should avoid doing this.  We do not want "cleanup" of code.

Go ahead and leave this in now, but please avoid it in the future.

MonoGame.Framework/BoundingFrustum.cs
```diff
@@ -4,103 +4,178 @@
...
-        #endregion Private Fields
+        #endregion
 
         #region Public Fields
-        public const int CornerCount = 8;
-        #endregion
 
-        #region Public Constructors
+        /// <summary>
+        /// Plane count(6).
```
tomspilman: ```The number of planes in the frustum.```

Remove the number too.

MonoGame.Framework/BoundingFrustum.cs
```diff
@@ -4,103 +4,178 @@
...
-        public BoundingFrustum(Matrix value)
-        {
-            this.matrix = value;
-            this.CreatePlanes();
-            this.CreateCorners();
-        }
-
-        #endregion Public Constructors
+        /// <summary>
+        /// Corner count(8).
```
tomspilman: ```The number of corner points in the frustum.```

Remove the number too.

MonoGame.Framework/BoundingFrustum.cs
```diff
@@ -4,103 +4,178 @@
...
+                this._matrix = value;
                 this.CreatePlanes();    // FIXME: The odds are the planes will be used a lot more often than the matrix
-            	this.CreateCorners();   // is updated, so this should help performance. I hope ;)
-			}
+                this.CreateCorners();   // is updated, so this should help performance. I hope ;)
+            }
         }
 
+        /// <summary>
+        /// Gets the near plane of this frustum.
```
tomspilman: use `of the frustum`.

MonoGame.Framework/BoundingFrustum.cs
Shqrdx: Please define it and I implement it

MonoGame.Framework/BoundingFrustum.cs
Shqrdx: Ok, I forgot it a bit, so sorry... You will call automatically cleanup system ?

Shqrdx:
@tomspilman It is ok for merging now ?

Shqrdx:
@tomspilman Please recheck.

tomspilman:
Merging.

PR #3643, opened 2015-03-20
AAC streams should have m4a extension

KonajuGames:
mp4 is the container format and is usually used for files containing video. m4a is used for files containing audio only.


tomspilman:
Ok... merging!

PR #3646, opened 2015-03-21
KeyboardInput and MessageBox converted to static form

Shqrdx:
These classes intended to be static. They are partial but it is not a problem for combine static and partial in one place.


tomspilman:
@Nezz - This seem right to you?

KonajuGames:
That makes sense.  It prevents someone from trying to create an instance of that class, which is pointless if all members are static.

Nezz:
Yes, this is correct.

tomspilman:
Thanks guys!

PR #3647, opened 2015-03-21
Updated improvements.md

Shqrdx:
Small update for help docs.

### DIFF
--- a/Documentation/improvements.md
+++ b/Documentation/improvements.md
@@ -6,9 +6,17 @@ The new API features which are available in the current development release and 
 
 [GameWindow.IsBorderless](http://www.monogame.net/documentation/?page=P_Microsoft_Xna_Framework_GameWindow_IsBorderless) - boolean that makes border of your window invisible. You cannot move or resize your game window after this boolean setted to true. Works only on desktop platforms.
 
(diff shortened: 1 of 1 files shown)


Shqrdx:
@tomspilman Is it good for merging ?

Shqrdx:
@tomspilman @KonajuGames Could this be merged ?

tomspilman:
I'll merge this, but I think we should reconsider having it.  

First it is 99% the same information as what is in the reference docs for each method.  It is hard enough to have one good set of docs... we don't need copies of the same info.

Second it explains dumb things.  We don't need an explanation of how `Point.ToVector2()` works...  it is self explanatory.

Third as MonoGame adds more and more features beyond XNA this document becomes huge.  It would be better to find a way to mark non-XNA APIs automatically from the doc tool.

IMO... we should copy any useful information back into the reference docs and kill this page.

Shqrdx:
Yeah, after your explanation me too think it is useless..

danzel:
We could have a custom attribute on such methods that the doc generator could pick up?

tomspilman:
> We could have a custom attribute on such methods that the doc generator could pick up?

That could be possible, but i'd prefer if whatever we did didn't add something at runtime to the type.

Maybe a custom element in the doc xml is possible?  Line `<not-xna-compatible/>` or something?

PR #3654, opened 2015-03-23
Implement texture arrays for Direct3D

### LINKED ISSUE #3653: Consider implementing / exposing texture arrays
In Direct3D, there's an `ArraySize` property for textures, which I think would be really useful to expose:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.DirectX.cs#L363

```
desc.ArraySize = 1;
```

OpenGL has [array textures](https://www.opengl.org/wiki/Array_Texture), which look like the same thing (although not supported by OpenGL ES 2.0).

I think this could be a backwards-compatible change: if you want to `SetData` on one of these, you'd use the `level` parameter. You'd calculate the value for the `level` parameter using something similar to Direct3D's [D3D11CalcSubresource](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476081%28v=vs.85%29.aspx), which we'd need to expose as a new API.

If this looks like something that would be of interest, then I'd be happy to work on a PR. I just wanted to first see if it matches MG's goals.

(My use-case: for cascaded shadow maps, it's nice to use a Texture2D array. The alternative is to use a single texture containing all the shadow maps, with offsets to sample from the correct cascade. It's certainly do-able without texture arrays, but more elegant with them.)

tgjones:
Currently only implemented for Direct3D.

A prerequisite for OpenGL support is MojoShader support for texture array samplers in HLSL.

Fixes #3653.


MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -933,6 +933,18 @@ private void PlatformSetViewport(ref Viewport value)
...
+                throw new InvalidOperationException("Texture arrays are not supported on this graphics device");
+
+            if (renderTarget == null)
+                SetRenderTarget(null);
+            else
+                SetRenderTargets(new RenderTargetBinding(renderTarget, arraySlice));
+        }
+
+        // Only implemented for DirectX right now, so not in GraphicsDevice.cs
         public void SetRenderTarget(RenderTarget3D renderTarget, int arraySlice)
```
tomspilman: I forget.... what is the difference here in `RenderTarget3D` and `RenderTarget2D`?  Are we basically making the same types now?

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -18,6 +18,7 @@ internal protected enum SurfaceType
 
 		internal int width;
 		internal int height;
+        internal int ArraySize;
```
tomspilman: Make this `_arraySize` here.  As eventually after we reformat code that is what it will be.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
tgjones: This is from [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476906%28v=vs.85%29.aspx):

> When a 3D texture mipmap slice is bound as a render target output (with a render-target view), the 3D texture behaves identically to a 2D texture array with n slices. The particular render slice is chosen from the geometry-shader stage, by declaring a scalar component of output data as the SV_RenderTargetArrayIndex system-value.
There is no concept of a 3D texture array; therefore a 3D texture subresource is a single mipmap level.

A `RenderTarget3D`, and a `RenderTarget2D` with `arraySize > 1`, are actually different things, and stored differently. But when you bind them as render targets, they're treated as the same thing.

tomspilman: Interesting... so they are different things.  I got concerned there for a second.

MonoGame.Framework/Graphics/Texture2D.cs
tgjones: I made it pascal-case because it's `internal`, not `private` - it is referenced from `RenderTargetBinding`. (And it's a field, not a property, for efficiency.)

Should I still rename it?

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
tgjones: Actually... `TextureCube`, and `Texture2D` with `arraySize > 1`, **are** basically the same thing. The only difference, in Direct3D anyway, is in the `BindFlags`. (And same goes for `RenderTargetCube` and `RenderTexture2D`).

tomspilman: > Actually...  TextureCube , and  Texture2D  with  arraySize > 1 , are basically the same thing.

Except I think `TextureCube` is sampled by a normal and filters across the different faces where as `Texture2D` does not.  Right?

MonoGame.Framework/Graphics/Texture2D.cs
tomspilman: Interestingly the MS C# style guide used for the new .NET core project doesn't mention internals:

https://github.com/dotnet/corefx/wiki/Coding-style

So looking thru their code base I was able to find one example case:

https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Process/src/System/Diagnostics/ThreadInfo.cs#L15

So I would say yeah... rename it.

tomspilman: Interesting that even in their repo it varies:

https://github.com/dotnet/corefx/blob/master/src/System.Linq.Parallel/src/System/Linq/Parallel/Utils/Shared.cs#L18

https://github.com/dotnet/corefx/blob/master/src/System.Xml.XmlDocument/src/System/Xml/Dom/XmlChildEnumerator.cs#L10

So who knows what is "right" here.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
tgjones: Yep, in HLSL, there's certainly a TextureCube type. But on the D3D side, there's no TextureCube type - only a Texture2D with arraySize == 6 and BindFlags == BindFlags.TextureCube.

That said, I don't think it's a problem for MG to have both. The fact that one is a special case of the other is a D3D (and maybe actual hardware?) implementation detail. It's not the same in OpenGL. I don't know about Vulkan / D3D12, but we can cross that bridge later.

For now, I think having TextureCube / RenderTargetCube classes matches what users expect.

tomspilman: > I think having TextureCube / RenderTargetCube classes matches what users expect.

I agree.

MonoGame.Framework/Graphics/Texture2D.cs
tgjones: Ha, 3 possibilities! I'm fairly sure I remember seeing internal field == Pascal case mentioned in an MS style guide some years ago, but I have no source to back me up. It's what ReSharper suggests... if I'm allowed to use that as a source :)

tomspilman: Sure... leave it for now since things are unclear there.

tomspilman: You can still leave this as is, but I got MS to clarify the rules for internals in their style guide:

https://github.com/dotnet/corefx/wiki/Coding-style

> 3.We use  _camelCase  for internal and private members and use  readonly  where possible. Prefix instance fields with  _ , static fields with  s_  and thread static fields with  t_ .

tgjones: That was quick :) I don't think much of the prefixes for static and thread-static fields, that feels quite old-school - but I guess MS get to decide...

tgjones:
Let me know if there's anything more you'd me to do here. No problem if you want to let it brew for a couple of days to give people time to have a look at it.

KonajuGames:
As usual you've done a great job with it, @tgjones.  I think the comments so far have covered everything I would have brought up.

tgjones:
I just fixed a bug that would have been rather embarrassing if it had been merged as it was. I added a test for that case. Then I fixed an existing bug (exposed by the new test) in `Texture2D.DirectX.cs`, in `GetData`, where it returned the wrong data for `level > 0`.

tgjones:
@tomspilman This is ready to merge, from my point of view. I've tested it in a couple of projects.

tomspilman:
Yeah...  i'm good with this.  Lets merge!

Issue #3649, opened 2015-03-22
ModelMesh.Tag ignored in ContentManager.Load<Model>()


lllAlexlll:
Hello,

***I have started porting my existing game engine from Xna to MonoGame recently. The process goes well and MonoGame looks very good but I have faced some issues. I will create 1 bug per issue.

MonoGame 3.3.0.2238
<b>Issue 1: ModelMesh.Tag ignored in ContentManager.Load<Model>()</b>
I have custom model processor which uses Model.Tag, ModelMesh.Tag and ModelMeshPart.Tag properties to store custom data. In ContentMagager.Load<Model>(), Model.Tag and ModelMeshPart.Tag are loaded just fine but ModelMesh.Tag is simply ignored. Currently I use simple workaround storing ModelMesh.Tag data in the ModelMeshPart.Tag but honestly I do not see any point to ignore MedelMesh.Tag property and I think that is a bug.

There are 
modelMeshPart.Tag = reader.ReadObject<object>();
and
model.Tag = reader.ReadObject<object>();
assignments in the
ModelReader : ContentTypeReader
class but there is no such assignment for the ModelMesh.Tag property.

Thank you!

tomspilman:
Yep...  just need to fix this one line:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentReaders/ModelReader.cs#L153

Thanks for reporting this @lllAlexlll !

AndersonReyes:
This would be a super simple fix for me to start contributing. Has it been fixed?

KonajuGames:
It hasn't yet.  Go for it.​

lllAlexlll:
I confirm that the issue is fixed in the latest build.

tomspilman:
Thanks for confirming @lllAlexlll !

PR #3659, opened 2015-03-26
ModelMesh.Tag fix

AndersonReyes:
Fixed issue where the ModelMesh.Tag instance was being ignored issue #3649

### DIFF
--- a/MonoGame.Framework/Content/ContentReaders/ModelReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/ModelReader.cs
@@ -149,9 +149,6 @@ namespace Microsoft.Xna.Framework.Content
                 var parentBoneIndex = ReadBoneReference(reader, boneCount);
 				var boundingSphere = reader.ReadBoundingSphere();
 
-                // Tag
-                reader.ReadObject<object>();
-
                 // Read the mesh part data.
                 int partCount = reader.ReadInt32();
                 //Debug.WriteLine("Mesh part count: {0}", partCount);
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Content/ContentReaders/ModelReader.cs
```diff
@@ -149,9 +149,6 @@ protected internal override Model Read(ContentReader reader, Model existingInsta
                 var parentBoneIndex = ReadBoneReference(reader, boneCount);
 				var boundingSphere = reader.ReadBoundingSphere();
 
-                // Tag
-                reader.ReadObject<object>();
-
```
tomspilman: This looks broken to me.  You can't move the `ReadObject` call...  that has to remain where it is to obey the file format for models.

Did you test it with your own game?

AndersonReyes: But there is no mesh from ModelMesh class instance before that which is what the code has to write ain't it?

tomspilman:
```
var meshTag = reader.ReadObject<object>();
```

Then assign it below.

AndersonReyes: then assigned it below like mesh.Tag = meshTag?

tomspilman: Yes.

lllAlexlll: Tested. I confirm that the issue is fixed now.

Issue #3644, opened 2015-03-20
Audio compilation broken :(


cra0zy:
So I already know what caused this issue, it's this:
https://github.com/mono/MonoGame/commit/0632ec611217985c2c4738b2f8f1777a056f0583

And to be even more precise, it's this line of code:
> loopLength = data.Count / ((bitsPerSample / 8) * channelCount);

The error that occurs is this:
> Build started 3/20/2015 8:40:13 PM

> /home/harry/Documents/songtest/Game01.mp3
/home/harry/Documents/songtest/Game01.mp3: error: Processor 'SongProcessor' had unexpected failure!
System.NullReferenceException: Object reference not set to an instance of an object
  at Microsoft.Xna.Framework.Content.Pipeline.Audio.AudioContent.ConvertFormat (ConversionFormat formatType, ConversionQuality quality, System.String saveToFile) [0x00000] in <filename unknown>:0 
  at Microsoft.Xna.Framework.Content.Pipeline.Processors.SongProcessor.Process (Microsoft.Xna.Framework.Content.Pipeline.Audio.AudioContent input, Microsoft.Xna.Framework.Content.Pipeline.ContentProcessorContext context) [0x00000] in <filename unknown>:0 
  at Microsoft.Xna.Framework.Content.Pipeline.ContentProcessor`2[Microsoft.Xna.Framework.Content.Pipeline.Audio.AudioContent,Microsoft.Xna.Framework.Content.Pipeline.Processors.SongContent].Microsoft.Xna.Framework.Content.Pipeline.IContentProcessor.Process (System.Object input, Microsoft.Xna.Framework.Content.Pipeline.ContentProcessorContext context) [0x00000] in <filename unknown>:0 
  at MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.ProcessContent (MonoGame.Framework.Content.Pipeline.Builder.PipelineBuildEvent pipelineEvent) [0x00000] in <filename unknown>:0 
Build 0 succeeded, 1 failed.

> Time elapsed 00:00:01.05.

KonajuGames:
How is data null there? That contains the PCM data.

I'll check it out later today.

cra0zy:
From the PR above:
//this line of code breaks audio compilation and therefor it has been commented
//2 things that this line of code does wrong are:
//1) data can be null, set by line 222
//2) bitsPerSample can be 0 ???
//loopLength = data.Count / ((bitsPerSample / 8) * channelCount);

[...]

PR #3665, opened 2015-03-26
Don't calculate loopLength if a null or zero

### LINKED ISSUE #3663: [Pipeline,WIN] Latest Pipeline seems to be unable to convert Mp3 to SoundEffects
C:/Repos/DoomAndDestiny/XNA4/Content/Audio/SE/A_Team.mp3: error: Processor 'SoundEffectProcessor' had unexpected failure!
System.DivideByZeroException: Attempted to divide by zero.
   at Microsoft.Xna.Framework.Content.Pipeline.Audio.AudioContent.ConvertFormat(ConversionFormat formatType, ConversionQuality quality, String saveToFile) in c:\GitHub2\MonoGame\MonoGame.Framework.Content.Pipeline\Audio\AudioContent.cs:line 287
   at Microsoft.Xna.Framework.Content.Pipeline.Processors.SoundEffectProcessor.Process(AudioContent input, ContentProcessorContext context) in c:\GitHub2\MonoGame\MonoGame.Framework.Content.Pipeline\Processors\SoundEffectProcessor.cs:line 59
   at Microsoft.Xna.Framework.Content.Pipeline.ContentProcessor`2.Microsoft.Xna.Framework.Content.Pipeline.IContentProcessor.Process(Object input, ContentProcessorContext context) in c:\GitHub2\MonoGame\MonoGame.Framework.Content.Pipeline\ContentProcessor.cs:line 60
   at MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.ProcessContent(PipelineBuildEvent pipelineEvent) in c:\GitHub2\MonoGame\MonoGame.Framework.Content.Pipeline\Builder\PipelineManager.cs:line 619
Build 0 succeeded, 1 failed.

KonajuGames:
is in any member used to calculate it.  Audio compilation does need a revision based on what I found.

Fixes #3663, #3644

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
@@ -284,7 +284,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Audio
 
                 // Loop start and length in number of samples. Defaults to entire sound
                 loopStart = 0;
-                loopLength = data.Count / ((bitsPerSample / 8) * channelCount);
+                if (data != null && bitsPerSample > 0 && channelCount > 0)
+                    loopLength = data.Count / ((bitsPerSample / 8) * channelCount);
+                else
+                    loopLength = 0;
             }
             finally
             {


MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
```diff
@@ -284,7 +284,10 @@ public void ConvertFormat(ConversionFormat formatType, ConversionQuality quality
 
                 // Loop start and length in number of samples. Defaults to entire sound
                 loopStart = 0;
-                loopLength = data.Count / ((bitsPerSample / 8) * channelCount);
+                if (data != null && bitsPerSample > 0 && channelCount > 0)
+                    loopLength = data.Count / ((bitsPerSample / 8) * channelCount);
+                else
+                    loopLength = 0;
```
tomspilman: Under what conditions can all this data...  the array, the sample bits, and the channel count.... be null?  Are these not required to correctly generate audio content?

KonajuGames: This PR will get the audio content pipeline working for all inputs again.  Loop length will be set to zero, but the XAudio path is the only one that uses loopLength, and it would seem that a loopLength of zero doesn't adversely affect XAudio sound playback.

The audio content pipeline needs some rework.  The importers currently just copy the input file into a buffer. This input file is then used to generate the AudioFormat for the output (via ffprobe), which is wrong.  Simply using the output file in ffprobe doesn't work either because of the PCM format we convert to and ffprobe can't read it.

I also think that compressed input should be decompressed to PCM in the importer, and then PCM is converted to the output format in the processor.  So a few things that I think should be looked at further, but for now this will get people building audio content again.

tomspilman: I see... so this is just a work around for now.

PR #3666, opened 2015-03-26
Replace MGFX DirectX preprocessor With CppNet

### LINKED ISSUE #1972: Replace MGFX DirectX Preprocessor With Portable One
Currently the MGFX tools depend on the Microsoft DirectX Preprocessor API for resolving `#include`s, `#define`s, and such before the shader is compiled.

https://github.com/mono/MonoGame/blob/develop/Tools/2MGFX/ShaderInfo.cs#L400

We need to replace this preprocessor with one of our own that is portable to Mac and Linux and does not have a dependency on DirectX.

tgjones:
Fixes #1972.

It would be good to have as many people as possible try this out in their own projects, before this is merged. Assuming the basic integration with CppNet holds up, we also need to make sure that CppNet itself is up to the task.


tomspilman:
@tgjones - Did you see the new preprocessor emitting `#line` statements where it changes the original file?

tgjones:
> Did you see the new preprocessor emitting #line statements 

Yes, I did (I explicitly enabled the CppNet option to emit #line statements).

tomspilman:
Fantastic... those are totally required to get good debug line info when we have warnings/errors.

tgjones:
Yep... the test I added has an `#include`, and I tried causing a syntax error and seeing the resulting line number - it looked correct to me.

Shqrdx:
@tgjones You do a fantastic job. Please continue and without loosing optimizm !

tomspilman:
@tgjones - Oops... I think merging https://github.com/mono/MonoGame/pull/3679 conflicted this PR.

tgjones:
@tomspilman Thanks for the heads-up - I've rebased this PR.

[[This](https://twitter.com/roastedamoeba/status/582845492129742848) just happened, so my response times might be a little higher than average for the next few weeks... :-)]

danzel:
Congrats @tgjones !

tomspilman:
This seems pretty safe to merge.  Worst case some weird preprocessor case fails and we won't find that until we start using it full time.

tgjones:
Thanks @danzel :)

> This seems pretty safe to merge.

My concerns are mainly around error reporting, and making sure the line numbers for errors match up correctly. It *should* be okay, but only time will tell.

PR #3667, opened 2015-03-26
Added Ubuntu 15.04 to dependencies

cra0zy:


KonajuGames:
Do you have to add each version of Ubuntu separately? Can it be >= 14 so you don't have to modify it each time they release a new version?

cra0zy:
I could, but it may fail on newer releases, I only add platforms on which I actually try running it

cra0zy:
If you still want me to setup >= just say it, and I'll add it.

KonajuGames:
@tomspilman what are your thoughts?  I'm leaning towards making it a >= check for Ubuntu version.  If any future version breaks then it can be dealt with at that time.

tomspilman:
I don't know enough about Linux to say.  Whatever you guys think is best.

cra0zy:
Silly me, yesterday I forgot to push changes, I just noticed

KonajuGames:
Thanks for that @cra0zy

KonajuGames:
Merged.

PR #3645, opened 2015-03-20
Bunch of small improvements and fixes for Gtk Pipeline

cra0zy:
Things done:
<ul>
<li>Changed "Monogame Pipeline" to "MonoGame Pipeline"</li>
<li>Fixed a bug where the Pipeline had a small chance of crashing upon calling buid/rebuild/clean</li>
<li>Output TextView now autoscrolls during compilation</li>
<li>Double clicking item in treeview now opens the default application for that item</li>
<li>Text dialog now checks entered characters to make sure they are characters than can be used for creating a directory</li>
<li>Some terminology improvements (#3292):</li>
</ul>
file
folder
main

PS. Next PR I will submit will bring WinForms Pipeline up to date

PR #3669, opened 2015-03-26
Gtk Pipeline Tool, Fixed context menu and retitled the tool to MonoGame Pipeline

cra0zy:
Notes:
<ul>
<li>In #3645 I changed the title of the tool using monodevelop's gui editor which actually only changed it in gui.stetic instead of MonoGame.Tools.Pipeline.MainWindow.cs</li>
<li>Context menu submenus were broken because of the Gtk bug (<a href="http://stackoverflow.com/questions/5221326/submenu-item-does-not-call-function-with-working-solution">stackoverflow link</a>) so I fixed that up</li>
<li>Fixed context menu separator positions</li>
</ul>


tomspilman:
Thanks @cra0zy !

PR #3670, opened 2015-03-27
Installer fix for templates on Windows

Shqrdx:
This is correction fix for installer. Before this, installer always mark templates even if appropriate visual studio's(or its folder in Documents) is not existed in the system.

image

after this change it should be(on my computer where only VS2013 existed) like this

image


tomspilman:
Looks great... merging!

Shqrdx:
And it works !

PR #3671, opened 2015-03-27
Bunch of small improvements for Gtk Pipeline

cra0zy:
Notes:
<ul>
<li>Project treeview base gets expanded by default</li>
<li>Added a new id system so I can add icons and something @tomspilman will be proud of</li>
<li>Fixed a Gtk bug which I will not explain, very complicated for explaining, tho it could only occur if compiled pipeline with system gtk libs</li>
<li>All items in project treeview now get sorted</li>
<li>Missing indicator will appear on folders in case if one of it's subitems has a missing indicator</li>
</ul>


tomspilman:
I'm proud of everything you do @cra0zy . :)

cra0zy:
Thanks :)

PR #3674, opened 2015-03-28
Added MonoGame color.

Shqrdx:
I think it is not bad addition for Color.

If this addition is accepted - the https://github.com/mono/MonoGame/pull/3587 could be rewritted to more compact form.


KonajuGames:
I think it would be a nice addition, but I also thin it should be `MonoGameOrange` so it is more definite it is the orange standard colour we use.

Shqrdx:
@KonajuGames Done

tomspilman:
Works for me... merged!

PR #3679, opened 2015-03-29
Update MojoShader dependency

### LINKED ISSUE #3658: Possible OpenGL depth range / clipping bug
It's very possible that I'm missing something obvious.

But it seems to me that MonoGame does not account for the difference in NDC space between Direct3D and OpenGL.

In Direct3D, the Z-coordinate in NDC ranges from 0 to 1, while in OpenGL it ranges from -1 to 1. So in Direct3D, anything in NDC space with `z < 0` will be clipped, while in OpenGL anything in NDC space with `z < -1` will be clipped.

The `Matrix.CreateOrthographic*` and `Matrix.CreatePerspective` methods use Direct3D conventions, so the NDC depth range created by these methods is from 0 to 1.

As far as I can see, that means that MonoGame's OpenGL renderer only ever uses half of the available depth range. Worse, near plane clipping doesn't work properly, because objects that *should* be behind the near clip plane, and therefore clipped, in fact aren't, because the near clip plane ends up (potentially) behind the camera. (That's how I noticed this bug - something that should have been clipped wasn't being clipped.)

The only mention I can find of this issue is over on FNA, here: flibitijibibo/FNA#282. @flibitijibibo mentioned a possible solution: `glClipControl`, and implemented it for FNA [here](https://github.com/flibitijibibo/FNA/commit/99019befdc4eb6d2de31706d0e89b7532affce3e), but it's only available in recent versions of OpenGL. You can also modify the `Matrix` methods that create projection matrices, but obviously that's not a general solution. @flibitijibibo has also added a note to the FNA docs about the issue, [here](https://github.com/flibitijibibo/FNA/wiki/2:-Building-with-FNA#depth-clipping-accuracy).

But I feel like I must be missing something, because surely this would have been noticed before now?

*(Note: MonoGame uses `posFixup` to account for the difference between Direct3D and OpenGL in the 0 location for the screen-space Y-coordinate. That's related to this, but it doesn't solve the NDC Z-coordinate problem.)*

tgjones:
Fixes #3658.


tomspilman:
Thanks... merging!

PR #3680, opened 2015-03-29
MGFX Render State Fixes

tomspilman:
This PR adds some missing render states into the MGFX parsing (stencil, scissor test, zfunc).

It also fixes the MGFX defaults for sampler states to match XNA.

Fixes https://github.com/mono/MonoGame/issues/2641.

### DIFF
--- a/Tools/2MGFX/MGFX.tpg
+++ b/Tools/2MGFX/MGFX.tpg
@@ -75,11 +75,42 @@ EndOfFile -> @"^$";
 [IgnoreCase] ColorWriteEnable -> @"ColorWriteEnable";
 [IgnoreCase] ZEnable -> @"ZEnable";
 [IgnoreCase] ZWriteEnable -> @"ZWriteEnable";
+[IgnoreCase] ZFunc-> @"ZFunc";
 [IgnoreCase] DepthBias -> @"DepthBias";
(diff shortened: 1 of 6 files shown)


tomspilman:
@tgjones - I want to add some sort of unit test for this.  What do you think is the best approach here?

tgjones:
> What do you think is the best approach here?

I really like using reference (i.e. `.out`) files for testing compilers - but that requires having a serializable intermediate or output format. We don't really have that for the effect parser, unless we make `ShaderInfo` / `TechniqueInfo` / `PassInfo` serializable using the MG XML serializer. But let's not do that if it's only needed for testing purposes.

So I think I'd add `ParserTests.cs`, and add some tests that call `new Parser(...).Eval()`, and then just do asserts against the resulting `ShaderInfo` object. It's quite a time-consuming way to write tests, especially for a large shader, but without a convenient serializable - and human-readable - output format from our shader parser, it's probably the best option.

Incidentally, in my CppNet PR, I added a preprocessor test to `EffectProcessorTest`, but I don't think that's the right place for it - we should have separate `PreprocessorTests` and `ParserTests`.

dellis1972:
I've tested this with some local .fx files which use those states, it works fine. +1 for merging

tomspilman:
Thanks @dellis1972 !

PR #3681, opened 2015-03-30
New constructors for Rectangle

Shqrdx:
This constructor uses two Point's and helps developers to write their 2d games more clearly and faster. At least I found this helpful.

update : added Rectangle(Vector2,Vector2)  and Rectangle(float,float,float,float) constructors.

### DIFF
--- a/MonoGame.Framework/Rectangle.cs
+++ b/MonoGame.Framework/Rectangle.cs
@@ -187,6 +187,20 @@ namespace Microsoft.Xna.Framework
             this.Height = height;
         }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


cra0zy:
Could you add float float float float and Vector2 Vector2 constructors? I think those would be really useful.

Shqrdx:
Of course

Shqrdx:
@tomspilman Be sure - Last commit does not cause ambiguity, but it would not cause confusion for new developers ? What do you think ?

Shqrdx:
@tomspilman Is it ok for merging ?

tomspilman:
Thanks!

PR #3682, opened 2015-03-30
New constuctors for Point

Shqrdx:
Point(value) - Vector2 already have same thing so why not ?

update : new constructors with float type instead integer.

### DIFF
--- a/MonoGame.Framework/Point.cs
+++ b/MonoGame.Framework/Point.cs
@@ -77,6 +77,16 @@ namespace Microsoft.Xna.Framework
             this.Y = y;
         }
 
+        /// <summary>
+        /// Creates a new instance of <see cref="Point"/> struct, with the specified position.
+        /// </summary>
+        /// <param name="value">The x and y coordinates in 2d-space.</param>
+        public Point(int value)
+        {
+            this.X = value;
+            this.Y = value;
+        }
+
         #endregion
 
         #region Operators
--- a/MonoGame.Framework/Point.cs
+++ b/MonoGame.Framework/Point.cs
@@ -87,6 +87,27 @@ namespace Microsoft.Xna.Framework
             this.Y = value;
         }
 
+        /// <summary>
+        /// Creates a new instance of <see cref="Point"/> struct, with the specified position.
+        /// </summary>
+        /// <param name="x">The x coordinate in 2d-space.</param>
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Point.cs
```diff
@@ -77,6 +77,37 @@ public Point(int x, int y)
...
+            this.X = value;
+            this.Y = value;
+        }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="Point"/> struct, with the specified position.
+        /// </summary>
+        /// <param name="x">The x coordinate in 2d-space.</param>
+        /// <param name="y">The y coordinate in 2d-space.</param>
+        public Point(float x, float y)
```
tomspilman: @KonajuGames - Do you think this is a good idea?

danzel: I don't really like this secret rounding

Shqrdx: Ok, so if I delete float overload's from Rectangle and Point, than my PR can be merged ?

tomspilman: > so if I remove float overload's from Rectangle and Point, than my PR can be merged ?

Yeah.... I say remove them.

theZMan: I don't see any value of the single int constructor either - how useful is that really? It seems to make no sense.

Shqrdx: It can be used with my new Rectangle constructor - https://github.com/mono/MonoGame/pull/3681

theZMan: I'm sure it can... but it makes no sense

new Point(14);
is like 
new Color(12);
or
new Point("11");
what does it mean?

We shouldn't just add methods unless there's a really, really great reason for doing do.

Shqrdx: The method itself is not necessary, I agree - because you can always write your own extension method. But for constructors it is the only way to decrease amount of space needed to creating instance of class/struct.

> Though wow Vector2 does indeed have such a useless constructor

It is exist for specific reasons. For example - you can create a surface with same width and height with this overload in one click.

tomspilman: I don't mind mimicking the single value ctor in `Point` to match `Vector2`.  It is a shortcut that could eliminate a bug and makes it clear that the coder indented `X` and `Y` to match.

PR #3683, opened 2015-03-30
Add Android GamePad Support

vostok4:
Use the framework laid out by the OUYA support to implement full GamePad
support for native Android.


tomspilman:
Looks like you missed adding `GamePad.Android.cs` to the framework definition file used for project generation:

https://github.com/mono/MonoGame/blob/develop/Build/Projects/MonoGame.Framework.definition#L794

tomspilman:
@dellis1972 @Nezz - What do you guys think of this?

dellis1972:
We should be able to share pretty much ALL the code between the OUYA and Android in this case. The only exception being the software Back button processing which the OUYA does not support.

tomspilman:
@dellis1972 - I see differences in this implementation and the OUYA implementation.  I don't know exactly how to merge those differences.

vostok4:
@tomspilman I'll fix the project definition, picked certain files from many diffs. Regarding OUYA and Android, yes, I think there will be ways to fix this if you don't mind some ifdef's in the code. I can take a stab at it next week.

dellis1972:
I took a look at the diff between android and ouya, probably easiest to just have an android implementation for now.

vostok4:
Not entirely sure why TC is failing on hasKeys(), it is there since API=9: http://developer.android.com/reference/android/view/InputDevice.html#hasKeys(int...)

Shqrdx:
@vostok4 Will you continue to develop and fixing this PR ?
@tomspilman Is it neccessary PR or standart GamePad and Joystick is enough even for android ? I dont heard before that android can uses gamepads.. except for maybe things like NVIDIA Shield...

tomspilman:
I know little about Android.  @KonajuGames would be a better judge of what is needed in this PR.

vostok4:
Just to bring this thread back up, I will try to close out the support for TC not failing this weekend.

Android GamePad support is definitely required, with devices like NVIDIA SHIELD TV showing off some good capabilities under the TV and already having a lot of games, native gamepad support would be great.

KonajuGames:
I recently ported a game from OUYA to NVIDIA Shield with very few changes to support the gamepad on the SHIELD handheld.  It was intended for the SHIELD console, but I never got to try it myself as they aren't available in my part of the world.

vostok4:
I will reopen this with a proper branch on my end, I'm just trying to get back to a happy state with my repo.

PR #3684, opened 2015-03-31
OpenGL platform version of Mouse.GetState(GameWindow) only works for pri...

christallire:
...mary window


tomspilman:
Looks good... merging!

PR #3688, opened 2015-03-31
[Gtk Pipeline Tool] Added new reference editor

cra0zy:
refs
One question I have for this is if person right clicks "References", and select "Delete", should I remove all references, should I not allow delete, something else?


tomspilman:
While it looks nice and I hate to reject good contributions...  I was *really* planning to keep these fake folders and stuff out of the tree view.

@KonajuGames   @dellis1972  ??

cra0zy:
I'm gonna wait to see what the rest thinks for this change, but for the reference edit dialog, I was planing to set it up anyway so do expect to see that tomorrow.

PS. Funny thing, I also named those folders "fake folders"

tomspilman:
> PS. Funny thing, I also named those folders "fake folders"

Hah.  It would be fine if this was a programmers tool, but we want to make this more artist/designer friendly.  Using folders and files to define something that isn't on disk is extremely confusing to non-programmers.

cra0zy:
I'm gonna close this, I will keep the branch around if anyone decides they want this type of ref editor.

PR #3691, opened 2015-04-03
[Gtk Pipeline Tool] Reference Editor Dialog

cra0zy:
Things done:
<ul>
<li>Added a new reference dialog that works similar to pre MonoDevelop 5.7 reference dialog.</li>
<li>Fixed YesNoCancel dialog question mark icon</li>
</ul>

refdialog

Do tell me if you want some different design for the ref dialog, but please leave the filechooser in there, it really speeds up reference selection.

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -46,6 +46,13 @@ namespace MonoGame.Tools.Pipeline
 
         public bool LaunchDebugger { get; set; }
 
+        public string ProjectLocation {
+            get
+            {
+                return _project.Location;
+            }
+        }
+
         public bool ProjectOpen { get; private set; }
 
         public bool ProjectDirty { get; set; }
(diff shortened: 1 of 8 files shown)


tomspilman:
> but please leave the filechooser in there, it really speeds up reference selection.

If you want to do that on Mac/Linux I guess that is ok.

But on Windows we should use the standard file chooser launched from an "Add Reference" button on the dialog.

tomspilman:
Why is the title on this new dialog "Pipeline"?  Shouldn't the title be like "References" or "Add/Remove References" or "Reference Editor"?

cra0zy:
> On Windows we should use the standard file chooser launched from an "Add Reference" button on the dialog.
> 
> If you want to do something different on Mac/Linux I guess that is ok.

Sounds good to me :D

> Why is the title on this new dialog "Pipeline"? Shouldn't the title be like "References" or "Add/Remove References" or "Reference Editor"?

You are right. The title of the dialog was null(default value when the title is not set), so it used application name as title. The reference dialog wasn't the only dialog that was missing a title so I fixed few other dialogs as well.

cra0zy:
@tomspilman Can this get merged?

tomspilman:
Merged!

cra0zy:
Yey, Thanks

PR #3693, opened 2015-04-03
Update GraphicsDevice.OpenGL.cs

StealthKill:
I changed the PlatformSetup() function a bit. Works now With my Editor which uses GLControl and my Game.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
@@ -88,11 +88,15 @@ namespace Microsoft.Xna.Framework.Graphics
             }
         }
 
-        private void PlatformSetup()
+       private void PlatformSetup()
         {
 #if WINDOWS || LINUX || ANGLE
             GraphicsMode mode = GraphicsMode.Default;
-            var wnd = (Game.Instance.Window as OpenTKGameWindow).Window.WindowInfo;
+
+            var wnd = OpenTK.Platform.Utilities.CreateWindowsWindowInfo(PresentationParameters.DeviceWindowHandle);
+
+            if (Game.Instance != null)
+                wnd = (Game.Instance.Window as OpenTKGameWindow).Window.WindowInfo;
 
             #if GLES
             // Create an OpenGL ES 2.0 context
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
```diff
@@ -88,11 +88,15 @@ internal void SetVertexAttributeArray(bool[] attrs)
...
         }
 
-        private void PlatformSetup()
+       private void PlatformSetup()
         {
 #if WINDOWS || LINUX || ANGLE
             GraphicsMode mode = GraphicsMode.Default;
-            var wnd = (Game.Instance.Window as OpenTKGameWindow).Window.WindowInfo;
+
+            var wnd = OpenTK.Platform.Utilities.CreateWindowsWindowInfo(PresentationParameters.DeviceWindowHandle);
```
tomspilman: @thefiddler @cra0zy - Is this safe to do for even the default window?

If so the `if (Game.Instance)` case below isn't needed then.

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
```diff
@@ -118,32 +122,43 @@ private void PlatformSetup()
                     0;
 
                 var samples = 0;
-                if (Game.Instance.graphicsDeviceManager.PreferMultiSampling)
+                if (Game.Instance != null)
                 {
-                    // Use a default of 4x samples if PreferMultiSampling is enabled
-                    // without explicitly setting the desired MultiSampleCount.
-                    if (PresentationParameters.MultiSampleCount == 0)
+                    if (Game.Instance.graphicsDeviceManager.PreferMultiSampling)
```
tomspilman: Do we really need this?  Isn't `PresentationParameters.MultiSampleCount == 0` essentially mean we don't want multisample?

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
```diff
@@ -118,32 +122,43 @@ private void PlatformSetup()
...
                     {
-                        PresentationParameters.MultiSampleCount = 4;
-                    }
+                        // Use a default of 4x samples if PreferMultiSampling is enabled
+                        // without explicitly setting the desired MultiSampleCount.
+                        if (PresentationParameters.MultiSampleCount == 0)
+                        {
+                            PresentationParameters.MultiSampleCount = 4;
```
tomspilman: This logic should be done within `GraphicsDeviceManager` if that is the case.  At this point the code should just obey the `MultiSampleCount` setting.

tomspilman:
This has some right ideas, but we can probably do it better.  Our goal should be to purge any reference to `Game` or `Game.Instance` from `PlatformSetup().

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
cra0zy: I don't know how safe it is with instancing, I would have to look at the code, but from memory, I don't think that this will work on linux or mac(there are methods like CreateX11WindowInfo and CreateMacWindowInfo for them).

Doing something like the code bellow would make much more sence:
IWindowInfo wnd;
if(Game.Instance.Window as OpenTKGameWindow != null)
    wnd = (Game.Instance.Window as OpenTKGameWindow).Window.WindowInfo;
else
    wnd = OpenTK.Platform.Utilities.CreateWindowsWindowInfo(PresentationParameters.DeviceWindowHandle);

tomspilman: @cra0zy - The point would be to not use `Game.Instance`.  We should be able to get what we need from `PresentationParameters.DeviceWindowHandle`.  `GraphicsDevice` needs to be independent from `Game`.

cra0zy: Oh, sorry, didn't read the bellow comments. That does sound like a good plan, I'll try and setup a PR for it tomorrow after school. Also later on, what do you think of setting up MonoGame WinForm Control and maybe Gtk Widget projects in the MonoGame repository?

tomspilman: > what do you think of setting up MonoGame WinForm Control and maybe Gtk Widget projects in the MonoGame repository?

If you mean setting up like a VS template project...  sure that could be useful.

christallire:
indeed, i've working for some prototype earlier in my fork.
@tomspilman could look this commit (https://github.com/christallire/MonoGame/commit/6485a9facab7728d0a9ab56cf55365a4056c5876) and comment is that you want?
i didn't made PR because it's not cleaned yet, but if you like this PR, i'll definately go for it.

tomspilman:
>  comment is that you want?

It is better...  but still don't want to see `GameWindow` added in there.  My issue is that we shouldn't be sneaking in types to `GraphicsDevice` via globals or special constructors.  

The design of XNA seems pretty clear...  `PresentationParameters` is how the "window" is passed to the `GraphicsDevice`.  This is how it works under DirectX...  it should be how this works for OpenGL too.

Ideally we would push some pointer type into `PresentationParameters.DeviceWindowHandle`.  This allows us to keep the XNA API design, but lets the platform specific code write/read the type it needs.

ArchaicOokami:
When you create a forms control with GLControl, it does contain a function that grabs the forms handle, its in WindowInfo.

ArchaicOokami:
@tomspilman to be more specific, everyone attempting to make a form control in GL is using OpenTK.GLControl (either ripping its direct source or using the dll) which at its current state doesnt actually require ANY modification. It contains a class which when your using GLControl it ALWAYS has which is WindowInfo. Window Info contains the Form Handle that you are requiring to target the window data... Outside of that, well no one is exactly going to get GLStuff to run directly in a form without recreating something similar to what is in that control... so if that has been the only thing you are really missing to getting to ensure the GL stuff works(like why you blocked it off) you should concider peaking into GLControl a bit more...

I have everything i believe you really need(+ an extra hack to get ANY GL support in a winform with current monogame though currently only works 1 window because of the GraphicsDevice limited to GameWindow) to make a single WinForms control that works DX and GL on my github post here:

https://github.com/ArchaicOokami/MonoGame.Framework.Control

Do note: I do use the MakeCurrent(WindowInfo) function in there, as it doesnt seem to "directly" pass to GraphicsDevice, however I do remember doing something that seemed i could call directly to the graphicsdevice instead... but the base GLControl does handle porting/rendering the stuff on its own without direct calls to GraphicsDevice.. The only problem right now(as i assume its from using a GraphicsDevice created on the GameWindow object) is it will only specificly handle the "Draws" from the first control that was created, but the clear info/colors work on all of them.

Correct me if im wrong!

mrhelmut:
This PR is more relevant since OpenTK has been replaced by SDL.

@tomspilman @KonajuGames

tomspilman:
Thanks @mrhelmut !

PR #3695, opened 2015-04-04
Custom Effects page documentation improvement

Itaros:
Adds information about how aggressive compiler can be in regards to :register(s0) if pixel shader is not fully defined. I think it is quite important information to be in tips as it was not highlighted in MVA course and due to silent nature of error creates confusion.

### DIFF
--- a/Documentation/custom_effects.md
+++ b/Documentation/custom_effects.md
@@ -46,6 +46,7 @@ These are some tips for writing or converting effects for use with MonoGame.
 * Note that on GL platforms default values on Effect parameters do not work.  Either set the parameter from code or use a real constant like a #define.
 * Do not name your sampler `Sampler` - it will not compile.
 * The effect compiler is aggressive about removing unused paramters, be sure the parameters you are setting are actually used.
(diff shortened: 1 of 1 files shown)


tomspilman:
The doc change here isn't very clear... i'm unsure what you are trying to say.

We already say:  `Make sure the pixel shaders inputs exactly match the vertex shader outputs`

We also say:  `The effect compiler is aggressive about removing unused parameters`

I think we could use a sample of a custom `SpriteBatch` effect which is often what trips people up.

Itaros:
About example. That will definitely solve the problem, but I wonder where it should be put. And considering those points about shader writing: ps inputs and vs outputs are not exactly helpful as simplified example of default vs output(when specifying only ps in effect.fx) is not provided in docs(source versions in repository are macros-heavy and are tricky to read) and not exactly helpful. That I was trying to achieve: to provide how inputs have to be

tomspilman:
>  but I wonder where it should be put.

Just add a `### Samples` section below the tips section.  Just keep the sample extremely simple and things should be fine.

Just submit a new PR with this change.

PR #3697, opened 2015-04-04
Drop an .MGCB file to open

nkast:


tomspilman:
This seems like an ok shortcut to me.

@cra0zy - This something we can add to Mac/Linux too?

cra0zy:
There is a bug with Gtk Textview, it doesn't receive any drag and drop events.... Setting up drag and drop for the main window would allow the .mgcb file to be dropped anywhere besides on the textview. 

Then again, there is even worst bug on MacOS, Gtk doesn't receive any drag drop events at all....

And there is even another bug, doing drag drop events to Gtk# app on Linux using Nautilus(default Ubuntu file manager) causes the drag and drop animation to randomly start appearing...

If you still want me to set it up, just say so, I've already create a git branch on my PC for it, I just haven't setup a PR because of the bugzzzzz.

tomspilman:
> There is a bug 

Wow... so lots of drag and drop bugs with GTK#.  I didn't know.

I guess we can just wait until those are fixed.  It isn't like drag and drop features are required to make things work.

cra0zy:
Gtk 3 was released in 2011, gtk# still uses gtk 2 bindings.... it might take a while....

tomspilman:
Is drag and drop just not a popular thing in the Mac/Linux world?  Don't see how this sort of feature could be left broken for so long.

cra0zy:
Drag and drop is used on Unix as much as on Windows. It's just that gtk# is really bad at it. Also, gtk# works on windows as well.

cra0zy:
I can't believe I forgot one widget that could be so useful for this... the Event Box.... Well I have a workaround for the textview bug :D. I'll setup a PR right away.

cra0zy:
False hope... even with eventbox drag & drop still doesn't work...

tomspilman:
Oh well...  all we can do is complain to the GTK# project I guess.

cra0zy:
Yesterday I found out that there were Gtk# 3 bindings here: https://github.com/mono/gtk-sharp/releases, tried them out, put the textview bug is still preset. Tho I did already port the pipeline tool on my PC to Gtk# 3, I really like the theme integration :)

tomspilman:
> tried them out, but the textview bug is still preset

 Has anyone reported this issue to them?

>  I did already port the pipeline tool on my PC to Gtk# 3

 Maybe post some experimental build up for people to test out.  Any pros/cons for updating to it?

cra0zy:
> Has anyone reported this issue to them?

First one not sure, I have to look into it a bit more(to see if it's gtk+ or gtk# bug) before reporting it
Mac bug: https://bugzilla.xamarin.com/show_bug.cgi?id=14724
Nautilus bug, this bug is way to random, and the only thing it does is it has a chance of showing the drag animation again few minutes time after you have already done it.

> Maybe post some experimental build up for people to test out. Any pros/cons for updating to it?

Yea, I will setup an experimental build of pipeline tool, but first I want to test some stuff on Mac (if I can get VM to work). Pros: Appearance and better integration with Gnome, New widgets(which are only in 3.X versions) Cons: Some methods had their name changed and a little bit different behavior for expanding widgets(some things had expand set true by default, now everything has it set to false)

PR #3698, opened 2015-04-04
Rework of Xamarin Studio/MonoDevelop Addin

dellis1972:
Includes new templates and new Shared Projects support
Includes Pipeline Tool support
Removes old ProjectTemplates

This is just a Test PR for now to check for any build issues.

Also worth noting that once this is in , the old MonoDevelopContent addin will no longer be needed and will also be removed probably as part of the PR as well. This is because we don't want to support the old xna content projects moving forward , only the .mgcb ones which this new addin will take care of.


Tools/Pipeline/Program.cs
```diff
@@ -45,11 +45,18 @@ static void Main(string [] args)
 			win.Show (); 
 			var model = new PipelineProject();
 			new PipelineController(win, model);  
+			#if LINUX
```
tomspilman: I think this should include windows too.

tomspilman:
@dellis1972 - It doesn't seem like this is hooked up to the build script?  Is that intentional?

Also something has conflicted this PR.

Tools/Pipeline/Program.cs
cra0zy: Gtk Pipeline is only built on Mac and Linux so there is no reason for there to include Windows, plus with how the file https://github.com/dellis1972/MonoGame/blob/Newaddin/Tools/Pipeline/Program.cs looks, that part of code is completely skipped on WIndows.

tomspilman: Yeah...  i'm wrong.   Github cut off line 41 here which clearly shows this is all within a `#if LINUX || MONOMAC` block.

dellis1972:
@tomspilman fixed the conflict and added the addin to build (for mac only)

cra0zy:
@tomspilman Will this get merged soon?

tomspilman:
@dellis1972 - This good to merge then?

Also do I need to add anything new to the build artifacts for the new addin?

cra0zy:
> Also do I need to add anything new to the build artifacts for the new addin?

If I understood the code correctly, there should be MonoGame.pkg generated in "Installers" folder.

PS. Also once this get's merged I have an idea on how to setup build server to automatically build Linux installer.

dellis1972:
@tomspilman  sorry for the delay tom, this notification got lost in a sea of emails. 

OK, there is a new MonoGame.pkg file in the Installers folder that needs artifacting ;) Also the Pipeline.MacOS.pkg file has moved into the Insstallers directory. Those are the only two that need to be published.

Note the MonoGame.Pkg is the beginnings of a full installer for OSX. It includes the Pipeline.pkg and the Addins, it will in the future also include the content targets file and setup a Framework in /Library/Frameworks/MonoGame that will be a future PR though. 

I still want to have the Pipeline.MacOS.pkg file available for designers etc. Although thinking about it I/we really should look at publishing that on the Store (via Xamarin.Mac) so designers DONT need to install mono.

cra0zy:
Would it be ok if I uploaded MonoGame pipeline to Ubuntu software center?

dellis1972:
@cra0zy I don't know what Tom/Steve thinks, but personally I think if its going to published on Stores (ubuntu/mac) it will need to be from an official MG account. But uploading is a good idea.

tomspilman:
> Would it be ok if I uploaded MonoGame pipeline to Ubuntu software center?

I have no clue how any of that works.

It isn't like the pipeline tool is useful on its own without the rest of MonoGame.  It seems like it would just not be all that useful to most people on that store.

Also keeping it up to date is yet another thing for us to do.

IMO we're better off distributing it ourselves.  @KonajuGames ?

cra0zy:
> It isn't like the pipeline tool is useful on its own without the rest of MonoGame. It seems like it would just not be all that useful to most people on that store.

I thought of publishing both monogame pipeline and monogame itself on store. I would still split it into 2 packages (monodevelop-monogame and monogame-pipeline).

> Also keeping it up to date is yet another thing for us to do.

I forgot to think about that, and no in a way you think. Ubuntu is released every 6 months and in order for new packages don't brake anything package upgrading is only possible before the Ubuntu release (package updates would only be possible every 6 months I think).

This doesn't account for setting up a ppa, which would actually be much smarter to do.

cra0zy:
@dellis1972 Is the addin finished/good for merging?

tomspilman:
@dellis1972 - This is conflicted again. :(

dellis1972:
@tomspilman ok I cleaned up the commit history and squashed it all down to 1 commit (so it doesn't look horrible) It should build ok, and if it does its good to go.

tomspilman:
Ok merging!  @dellis1972 - Fyi... I want to get a 3.4 release out during build... at least right after the new Win10 SDK ships.

cra0zy:
Can you give me a day of time to setup linux installation for the new addin?

tomspilman:
@cra0zy -  Sure... We have until next week some time.

dellis1972:
@tomspilman  awesome. next week sounds good. 

btw I have some local change that let the pipeline targets file work on Mac! :) Will try to get that sorted and update the mac templates to use that.

cra0zy:
@dellis1972 Did you upload the adding to addins.monodevelop.com?

dellis1972:
@cra0zy a version is in the beta channel. will upload a new one over the weekend, we'll publish to stable as soon as the next release is out

cra0zy:
Ok, in that case I will not setup linux installer for the addin(because it would be pointless).

Also you get this error:
addinerror
but after restarting monodevelop the addin works fine.

tomspilman:
>  Did you upload the addin to addins.monodevelop.com?

Can we automate that in some way @dellis1972 ?  I will not remember or think to do these special updates to various stores/systems.

tomspilman:
@dellis1972 - Also looking at cra0zy's screen shot... can we fix that green puzzle piece icon with a MonoGame one?

cra0zy:
> Also looking at cra0zy's screen shot... can we fix that green puzzle piece icon with a MonoGame one?

All MonoDevelop extensions have that icon, it's not something that's changeable...

tomspilman:
> All MonoDevelop extensions have that icon

 Weird...  oh well.

dellis1972:
@tomspilman I might be able to automate it, i'll look into it.

PR #3705, opened 2015-04-05
New docs & members for Matrix

Shqrdx:
Changes:
-Fully documented Matrix
-New constructor which accepts 4 Vector4 for Matrix
-New properties Scale and Rotation for Matrix
-New overload for Matrix.CreateOrthographicOffCenter and Matrix.CreatePerspectiveOffCenter which accepts Rectangle as viewing volume
-operator+ optimization
-Added MatrixTest class


Shqrdx:
I dont see where is the error...

Shqrdx:
Phew.. it was just a hypo error

Shqrdx:
@tomspilman Looks like random bug inside visual test. Needs your help.

Shqrdx:
@tomspilman Please fix & review

Shqrdx:
@tomspilman Looks like it is ready for merging ! Please review.

MonoGame.Framework/Matrix.cs
```diff
@@ -7,13 +7,35 @@
 using System.Runtime.Serialization;
 
 namespace Microsoft.Xna.Framework
-{
+{ 
+    /// <summary>
+    /// Represents the right-handed 4x4 float-based matrix, which can store translation,scale and rotation information.
```
tomspilman: `float-based` is weird.... just say `floating point`.
Put a space between `translation,scale`.

MonoGame.Framework/Matrix.cs
```diff
@@ -7,13 +7,35 @@
...
+    /// </summary>
     [DataContract]
     [DebuggerDisplay("{DebugDisplayString,nq}")]
     public struct Matrix : IEquatable<Matrix>
     {
         #region Public Constructors
-        
+
+        /// <summary>
+        /// Creates a new instance of <see cref="Matrix"/> struct.
```
tomspilman: We don't need to reference `Matrix` within the `Matrix`.   They are already reading the matrix docs.

Just say:  `Constructs a new matrix.`.

MonoGame.Framework/Matrix.cs
```diff
@@ -162,10 +259,13 @@ public struct Matrix : IEquatable<Matrix>
...
 		                                            0f, 0f, 1f, 0f, 
 		                                            0f, 0f, 0f, 1f);
-        #endregion Private Members
+        #endregion
 
         #region Public Properties
-        
+
+        /// <summary>
+        /// Gets or sets the backward vector of this matrix.
```
tomspilman: We discussed this one before...  no `Gets or sets`...  that is obvious from it being a property.

Also no `of this matrix` as of course it is of this matrix... what other matrix would it be?

If we want to add more words... lets add useful info like:  

`The backward vector formed from the third row M31, M32, M33 elements.`.

MonoGame.Framework/Matrix.cs
```diff
@@ -468,7 +637,12 @@ public static void Add(ref Matrix matrix1, ref Matrix matrix2, out Matrix result
 
         }
 
-
+        /// <summary>
+        /// Creates a new <see cref="Matrix"/> which contains the rotation moment around specified axis.
+        /// </summary>
+        /// <param name="axis">The axis of rotation.</param>
+        /// <param name="angle">The angle of rotation.</param>
```
tomspilman: Whenever we have an angle parameter we need to specify if it is in radians or degrees.  A very common bug that occurs when the docs are not specific.

MonoGame.Framework/Matrix.cs
```diff
@@ -545,14 +732,32 @@ public static void CreateFromQuaternion(ref Quaternion quaternion, out Matrix re
...
-		
+
+        /// <summary>
+        /// Creates a new rotation <see cref="Matrix"/> from the specified yaw,pitch and roll values.
+        /// </summary>
+        /// <param name="yaw">The yaw rotation value.</param>
+        /// <param name="pitch">The pitch rotation value.</param>
+        /// <param name="roll">The roll rotation value.</param>
+        /// <returns>The rotation <see cref="Matrix"/>.</returns>
+        /// <remarks>For more information about yaw,pitch and roll visit http://en.wikipedia.org/wiki/Euler_angles.
```
tomspilman: Space between `yaw,pitch`.

MonoGame.Framework/Matrix.cs
```diff
@@ -545,14 +732,32 @@ public static void CreateFromQuaternion(ref Quaternion quaternion, out Matrix re
...
 		    result.M44 = 1f;
         }
-		
+
+        /// <summary>
+        /// Creates a new rotation <see cref="Matrix"/> from the specified yaw,pitch and roll values.
+        /// </summary>
+        /// <param name="yaw">The yaw rotation value.</param>
+        /// <param name="pitch">The pitch rotation value.</param>
+        /// <param name="roll">The roll rotation value.</param>
```
tomspilman: Yaw, pitch, and roll are all rotation angles... are they supposed to be in radians or degrees?

MonoGame.Framework/Matrix.cs
```diff
@@ -715,7 +1008,16 @@ public static void CreatePerspectiveFieldOfView(float fieldOfView, float aspectR
...
+        /// <summary>
+        /// Creates a new projection <see cref="Matrix"/> for customized perspective view.
+        /// </summary>
+        /// <param name="left">Left coordinate of the viewing field.</param>
+        /// <param name="right">Right coordinate of the viewing field.</param>
+        /// <param name="bottom">Bottom coordinate of the viewing field.</param>
+        /// <param name="top">Top coordinate of the viewing field.</param>
+        /// <param name="nearPlaneDistance">Near plane of the viewing field.</param>
+        /// <param name="farPlaneDistance">Far plane of the viewing field.</param>
+        /// <returns>The new <see cref="Matrix"/> for customized perspective view.</returns>
```
tomspilman: The `of the viewing field` repetition here bothers me....  it feels like filler and we need more useful information.

@KonajuGames ?

MonoGame.Framework/Matrix.cs
Shqrdx: Other documented components such a Vector2,Rectangle,BoundingFrustum,Viewport etc... used this notation. Originally it is from SharpDX.

MonoGame.Framework/Matrix.cs
```diff
@@ -1444,50 +1936,84 @@ public static void Multiply(ref Matrix matrix1, ref Matrix matrix2, out Matrix r
...
 
+        /// <summary>
+        /// Copy the values of specified <see cref="Matrix"/> to the float array.
+        /// </summary>
+        /// <param name="matrix">The source <see cref="Matrix"/>.</param>
+        /// <returns>The array which matrix values will be stored.</returns>
+        /// <remarks>
+        /// Required for OpenGL 2.0 projection matrix stuff.
+        /// </remarks>
+        public static float[] ToFloatArray(Matrix matrix)
```
tomspilman: Is this still used in the code base?

MonoGame.Framework/Matrix.cs
```diff
@@ -1444,50 +1936,84 @@ public static void Multiply(ref Matrix matrix1, ref Matrix matrix2, out Matrix r
...
+									matrix.M11, matrix.M12, matrix.M13, matrix.M14,
+									matrix.M21, matrix.M22, matrix.M23, matrix.M24,
+									matrix.M31, matrix.M32, matrix.M33, matrix.M34,
+									matrix.M41, matrix.M42, matrix.M43, matrix.M44
+								};
+            return matarray;
+        }
 
+        /// <summary>
+        /// Creates a new <see cref="Matrix"/> that contains the specified matrix negation.
```
tomspilman: `Returns a matrix with the all values negated.`

MonoGame.Framework/Matrix.cs
```diff
@@ -1559,7 +2114,12 @@ public static void Negate(ref Matrix matrix, out Matrix result)
 		    return matrix1;
         }
 
-
+        /// <summary>
+        /// Divides the components of a <see cref="Matrix"/> by a scalar.
```
tomspilman: Use `elements` instead of `components`.

MonoGame.Framework/Matrix.cs
```diff
@@ -1582,7 +2142,12 @@ public static void Negate(ref Matrix matrix, out Matrix result)
 		    return matrix;
         }
 
-
+        /// <summary>
+        /// Compares whether two <see cref="Matrix"/> instances are equal.
```
tomspilman: It would be worth noting we check for exact equality without any tolerance.

MonoGame.Framework/Matrix.cs
Shqrdx: It is still been used in EffectParameter class(from WindowsDX project).

MonoGame.Framework/Matrix.cs
```diff
@@ -1800,7 +2410,11 @@ public override string ToString()
                 + " {M41:" + M41 + " M42:" + M42 + " M43:" + M43 + " M44:" + M44 + "}";
         }
 
-
+        /// <summary>
+        /// Creates a new <see cref="Matrix"/> which contains swapped rows and columns values from the specified <see cref="Matrix"/>.
```
tomspilman: `Swap the rows and columns of the matrix`

MonoGame.Framework/Matrix.cs
Shqrdx: Should I just remove it ?

MonoGame.Framework/Matrix.cs
Shqrdx: Ok, I delete it

tomspilman: > Other documented components such a Vector2,Rectangle,BoundingFrustum,Viewport etc... used this notation.

Doesn't mean it is a good thing.  We should strive for better docs than XNA or SharpDX...  they are not anything to be proud of.

Shqrdx: Ok, In next PR's I remove this from their constructors.

MonoGame.Framework/Matrix.cs
Shqrdx: Done !

Shqrdx:
Eh, I suggest completly disable those "visual tests" because they are buggy and useless.

tgjones:
> suggest completly disable those "visual tests"

I think investigating why the tests fail would be a more useful approach ;-) The tests are a really useful way to know if new pull requests break existing code.

Shqrdx:
At least ignoring results from them(without disabling) I think is a good idea for now.

tomspilman:
> At least ignoring results from them(without disabling)

We can't just ignore them either.  We need them to fail the build.

I'll re-run your changes and it should work.

Shqrdx:
@tomspilman Please make a decision about this PR.

MonoGame.Framework/Matrix.cs
tomspilman: @KonajuGames - Bump!

KonajuGames: The repetition itself doesn't really bother me, but in this case it is not really the correct terminology.  `left`, `right`, `bottom` and `top` are the bounds of the x and y axes at the near plane.  `Viewing field` is not a term used in 3D rendering.  Without copying XNA docs exactly, something like the following would be better.

        /// <param name="left">Lower x-value at the near plane.</param>
        /// <param name="right">Upper x-value at the near plane.</param>

and

        /// <param name="nearPlaneDistance">Depth of the near plane.</param>

Shqrdx:
@tomspilman I changed it how @KonajuGames said - is it ok now ?

Shqrdx:
@tomspilman recheck pls

tomspilman:
Ok... merging.  Thanks!

PR #3706, opened 2015-04-05
Add RasterizerState.DepthClipEnable

tgjones:
with implementation for Direct3D and OpenGL, and a visual test (which passes on both Direct3D and OpenGL).

On Direct3D, this corresponds to `RasterizerState.DepthClipEnable`. On OpenGL, this corresponds to calling `GL.Enable(DEPTH_CLAMP)`.

On Direct3D, this is not supported by level 9.x devices. On OpenGL, this is enabled by the ARB_depth_clamp extension. These restrictions are accounted for in `GraphicsCapabilities`.

(By the way, it was when running this visual test on OpenGL that I noticed that depth clipping wasn't working properly, even without these changes, which is what prompted the MojoShader fix.)


tgjones:
I forgot to mention why this is useful. One usage is to "pancake" depths when rendering shadow maps, so objects behind the light's near clip plane will still be rendered - which means you don't need to calculate a precise near clip plane distance. That's what I'm using it for.

(If this PR is accepted, then I will be able to publish a complete cascaded shadow maps sample that I've been working on - basically a MonoGame port of MJP's [shadows sample](https://mynameismjp.wordpress.com/2013/09/10/shadow-maps/) - which brings together some of these new features: texture arrays, comparison samplers, and depth clamping.)

tgjones:
@tomspilman @konajugames any thoughts on whether this would be a good fit for MonoGame?

tgjones:
@tomspilman @KonajuGames Sorry to be annoying, but have you had a chance to look at this? I've got my shadows sample all ready to go, but it won't work (for anybody other than me) until this PR is merged. (No problem if you decide not to merge it, I can apply a workaround, although of course it would be great if it is merged.)

Here's the sample:
https://github.com/tgjones/monogame-samples/tree/master/CascadedShadowMaps

tomspilman:
Sorry... it just has been crazy busy for me lately.

> any thoughts on whether this would be a good fit for MonoGame?

I think so.  Same concerns over making sure we eventually work out a "feature level" for this.  But generally I think it would be good to support.

I'll look over the PR in more detail tomorrow.

tgjones:
> Same concerns over making sure we eventually work out a "feature level" for this

I did the same as for texture arrays and comparison samplers - I used `GraphicsCapabilities` to set a "supported" flag. On Direct3D, it's only enabled for the `HiDef` profile, and on OpenGL, it's enabled if the required extension is present. Which basically, I think, means it's enabled for desktop platforms only.

> I'll look over the PR in more detail tomorrow.

Thanks!

tomspilman:
Ok... merging!

tgjones:
Great, thank you!

PR #3707, opened 2015-04-05
[DX] Fix rounding error at SetClientBounds()

nkast:
_clientBounds was 800x479 on a L620 device and 799x479 on the emulator.
After the fix _clientBounds is set to 800x480.


tomspilman:
Seems reasonable... merging!

PR #3648, opened 2015-03-21
Updated WinForms Pipeline with Add File Dialog, Folder Support and Better Context Menu Terminology

cra0zy:
After you merge this, you can close #3292 and #3291

PR #3708, opened 2015-04-06
Include assets by drag and drop

nkast:
Include assets by dropping on the Treeview window or the ContentProject
TreeNode.
Items that are outside the base directory are not supported.
Dropping items on a Folder node is not supported.
Dropping folders just add that folder to the project. does not add the
files from that folder.


cra0zy:
You are completely skipping over add file/folder dialogs added in #3648. I think when you drag and drop to pipeline a Copy/Link option should apear,

nkast:
Right now I just include files that are already under the content directory.
The intention of the user is clear, no need to redirect this through IView. That's why I had to add `Include(items,folders)` which mirrors the `Exclude(items, folders)`. 

In case of files/folders outside the base directory we could have Copy as default behavior and change to Link when CTRL is pressed and indicate that by toggling the mouse pointer (DragDropEffects.Link / DragDropEffects.Copy). That is the defacto behavior of drag&drop in windows.

cra0zy:
Oh, I didn't notice that you set the restriction to only files within the folder of .mgcb, makes sense now. 

For the CTRL thing, I think it sounds like an interesting idea, I like it. I'll setup a PR for Gtk Pipeline to mirror your changes.

nkast:
I see linking as a bad practice and copying is a file browser operation and Pipeline is clearly not a file browser. Although I'm not planning at the moment to implement Copy/Link of items outside the project directory here are some thoughts:

Definitions:
`contentRoot` = the directory of .mgcb 
`contentFolder` is any folder under `contentRoot` 
`OutItem` is any folder or file outside `contentRoot` 

The files of drop operator can be of any source, a user can select multiple files from Explorer tree or any other program, there is no guaranty that they are all rooted at the same folder.
Dropping anything into a file has no meaning, it's a no operation.
Dropping anything into a `contentFolder` for linking has no meaning, it will get translated relative to `contentRoot`. Therefore dropping on `contentFolder` is always a Copy operation, create a new copy and include whether its an 'OutItem' or another `contentFolder`.
Dropping a `contentFolder` into the `contentRoot` is always a Link operation, copying  is equivalent to Copy into itself + Link.
Dropping an `OutItem` into the `contentRoot` could be either Copy or Link.

Therefore we can have Copy as a default and change to Link with ALT when dropping into the 'contentRoot'. For any `contentFolder` it would be the same. It's just that Copy should be implemented as an IAction (remove the files when undo).

tomspilman:
> I see linking as a bad practice and copying is a file browser operation and Pipeline is clearly not a file browser.

Well you may feel it is a bad practice, but XNA supports linked content and MonoGame will support it too.


> we could have Copy as default behavior and change to Link when CTRL is pressed 

I think unless the file is within the folder, we should always pop up the dialog and give the user a choice.

Tools/Pipeline/Windows/MainView.cs
```diff
@@ -1041,11 +1041,56 @@ private void MainView_DragEnter(object sender, DragEventArgs e)
 
         private void MainView_DragDrop(object sender, DragEventArgs e)
         {
-            string filename = GetDropFile(e.Data, ".mgcb");
+            string filename = GetDropFile(e.Data, ".MGCB");
```
KonajuGames: Why this change to the case of the string literal for the extension?  `GetDropFile()` does a case-insensitive comparison anyway, so this change is redundant.  Lower-case file extensions have been the norm for the past 20 years, since MS-DOS was put to pasture.

nkast: fixed

cra0zy:
With #3916 there is now DragDrop method that is used for files dragged and dropped inside treeview... later today or tomorrow I'll setup Drag & Drop for WinForms pipeline.

cra0zy:
compared to Gtk treeview, WinForms treeview sucks... had to manually implement row highlighting... :|

nkast:
ok @cra0zy. Feel free to use anything you might find useful from my branch.

cra0zy:
#3920 already does it so this can be closed.

PR #3709, opened 2015-04-06
Allow MonoGame to be embedded inside GTK# applications (when using OpenTK)

hach-que:
This change allows MonoGame to be embedded inside GTK# applications when using OpenTK (so WindowsGL, Linux and Mac w/ Angle).  This is so that I can do this inside MonoDevelop:

protogameeditor

This is done by implementing `IEmbedContext` in the main application and passing it to the game when creating it.

An example standalone implementation of the embed context (with just a normal `GLWidget` in a standard GTK# application) is:

```csharp
using System;
using Gtk;

namespace ProtogameEditor
{
    public class EditorHostEmbedContext : IEmbedContext
    {
        private Gtk.Widget _glWidget;

        public OpenTK.Graphics.IGraphicsContext GraphicsContext
        {
            get;
            set;
        }

        public OpenTK.Platform.IWindowInfo WindowInfo
        {
            get;
            set;
        }

        public Gtk.Widget GLWidget
        {
            get
            {
                return this._glWidget;
            }
            set
            {
                if (this._glWidget != null)
                {
                    this._glWidget.ConfigureEvent -= this.Resize;
                }

                this._glWidget = value;

                if (this._glWidget != null)
                {
                    this._glWidget.ConfigureEvent += this.Resize;
                }
            }
        }

        public System.Drawing.Point PointToClient(System.Drawing.Point point)
        {
            int x, y;
            if (this.GLWidget.GdkWindow == null)
            {
                return new System.Drawing.Point(0, 0);
            }

            this.GLWidget.GdkWindow.GetOrigin(out x, out y);
            return new System.Drawing.Point(
                point.X - x,
                point.Y - y);
        }

        public Rectangle GetClientBounds()
        {
            return new Rectangle(
                0,
                0,
                this.GLWidget.Allocation.Width,
                this.GLWidget.Allocation.Height);
        }

        public event System.EventHandler OnResize;

        public void Resize(object sender, ConfigureEventArgs e)
        {
            if (this.OnResize != null)
            {
                this.OnResize(this, new System.EventArgs());
            }
        }
    }
}
```

and then using it in an application like this (the dependency injection kernel is just passing in the EditorHostEmbedContext as the first argument to the game constructor):

```csharp
using Gtk;
using Protogame;
using Ninject;
using ProtogameEditor;

namespace ProtogameEditorHost
{
    public static class Program
    {
        public static void Main(string[] args)
        {
            Application.Init();

            var window = new HostWindow();
            window.Show();

            var editor = new IDEEditor();

            var kernel = editor.CreateEditorKernel<EditorHostEmbedContext>();

            var context = kernel.Get<IEmbedContext>();

            EditorGame game = null;

            while (true)
            {
                Application.RunIteration();

                if (window.GLWidget.WindowInfo != null && game == null)
                {
                    ((EditorHostEmbedContext)context).GraphicsContext = window.GLWidget.GraphicsContext;
                    ((EditorHostEmbedContext)context).WindowInfo = window.GLWidget.WindowInfo;
                    ((EditorHostEmbedContext)context).GLWidget = window.GLWidget;

                    game = new EditorGame(kernel);
                }

                if (game != null)
                {
                    game.RunOneFrame();
                }
            }
        }
    }
}
```

### DIFF
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -69,7 +69,11 @@ namespace Microsoft.Xna.Framework
             _components = new GameComponentCollection();
             _content = new ContentManager(_services);
 
+#if (WINDOWS && OPENGL) || LINUX || ANGLE
             Platform = GamePlatform.Create(this, embedContext);
+#else
+            Platform = GamePlatform.Create(this, null);
+#endif
             Platform.Activated += OnActivated;
             Platform.Deactivated += OnDeactivated;
             _services.AddService(typeof(GamePlatform), Platform);
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -55,8 +55,12 @@ namespace Microsoft.Xna.Framework
 
 
         private bool _suppressDraw;
-        
+
+#if (WINDOWS && OPENGL) || LINUX || ANGLE
+        public Game(IEmbedContext embedContext = null)
+#else
         public Game()
+#endif
         {
             _instance = this;
 
@@ -65,7 +69,7 @@ namespace Microsoft.Xna.Framework
             _components = new GameComponentCollection();
             _content = new ContentManager(_services);
 
-            Platform = GamePlatform.Create(this);
+            Platform = GamePlatform.Create(this, embedContext);
             Platform.Activated += OnActivated;
             Platform.Deactivated += OnDeactivated;
             _services.AddService(typeof(GamePlatform), Platform);
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -1074,9 +1074,7 @@
     <Compile Include="Input\KeyboardUtil.OpenTK.cs">
       <Platforms>Angle,Linux,WindowsGL</Platforms>
     </Compile>
-    <Compile Include="Desktop\IEmbedContext.cs">
-      <Platforms>Angle,Linux,WindowsGL</Platforms>
-    </Compile>
+    <Compile Include="Desktop\IEmbedContext.cs" />
 
 
     <!-- iOS Platform -->
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -1074,6 +1074,9 @@
     <Compile Include="Input\KeyboardUtil.OpenTK.cs">
       <Platforms>Angle,Linux,WindowsGL</Platforms>
     </Compile>
+    <Compile Include="Desktop\IEmbedContext.cs">
+      <Platforms>Angle,Linux,WindowsGL</Platforms>
+    </Compile>
 
 
     <!-- iOS Platform -->
--- a/MonoGame.Framework/Desktop/IEmbedContext.cs
+++ b/MonoGame.Framework/Desktop/IEmbedContext.cs
@@ -1,0 +1,18 @@
+﻿namespace Microsoft.Xna.Framework
+{
+    public interface IEmbedContext
+    {
+#if (WINDOWS && OPENGL) || LINUX || ANGLE
+        OpenTK.Graphics.IGraphicsContext GraphicsContext { get; }
+
+        OpenTK.Platform.IWindowInfo WindowInfo { get; }
+
+        System.Drawing.Point PointToClient(System.Drawing.Point point);
+
+        Rectangle GetClientBounds();
+
+        event System.EventHandler OnResize;
+#endif
+    }
+}
+
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -91,11 +91,11 @@ namespace Microsoft.Xna.Framework
         private Toolkit toolkit;
         private int isExiting; // int, so we can use Interlocked.Increment
         
-		public OpenTKGamePlatform(Game game)
+        public OpenTKGamePlatform(Game game, IEmbedContext embedContext)
             : base(game)
         {
             toolkit = Toolkit.Init();
-            _view = new OpenTKGameWindow(game);
+            _view = new OpenTKGameWindow(game, embedContext);
             this.Window = _view;
 
 			// Setup our OpenALSoundController to handle our SoundBuffer pools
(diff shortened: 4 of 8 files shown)


MonoGame.Framework/Game.cs
```diff
@@ -65,7 +69,7 @@ public Game()
             _components = new GameComponentCollection();
             _content = new ContentManager(_services);
 
-            Platform = GamePlatform.Create(this);
+            Platform = GamePlatform.Create(this, embedContext);
```
cra0zy: In case of non (WINDOWS && OPENGL) || LINUX || ANGLE platforms, embedContext is not defined here.

tomspilman:
So this issue comes up every 2 or 3 months now.  I want to support embedding, but I want to avoid adding new APIs when XNA already supports this usage scenario.

What is wrong with the embedding technique already defined by XNA?

 http://xbox.create.msdn.com/en-US/education/catalog/sample/winforms_series_1

hach-que:
I'm pretty sure that uses window handles, which are a Windows-specific concept (at least in the way they are implemented in C#).

In addition, I'm in a position where I can't have MonoGame owning the GTK# control, because I need to be able to reload the AppDomain that contains the game, so that I can reload the editor when it gets rebuilt or modified in the IDE.  There's no way to marshal a GTK# control over an AppDomain (because it's neither serializable nor implements MarshalByRefObject).

However OpenTK's interfaces of `IGraphicsContext` and `IWindowInfo` are both marshallable (because I can write a proxy which implements the interfaces and performs the underlying calls as needed).

tomspilman:
> I'm pretty sure that uses window handles, which 
> are a Windows-specific concept 

It uses something it calls a window handle...

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.presentationparameters.devicewindowhandle.aspx

... but it is an `IntPtr`, so it can be anything we want it to be depending on the platform.

> I can't have MonoGame owning the GTK# control

Good... MonoGame like XNA doesn't want to own the control.  It just wants to be given a control to draw to.  See the sample I linked.

hach-que:
But under OpenTK, you need to provide access to both `IGraphicsContext` and `IWindowInfo`.. an IntPtr is useless because that isn't the variable type used inside the OpenTK platform.  You also need to provide callbacks to translate mouse position and get the client bounds.

Under Windows / XNA, that IntPtr is the native window handle, which Windows then takes care of propagating events and positioning of rendering automatically.  That doesn't apply under MacOS / Linux where the operating system can't or won't do that (at least under Linux there's multiple toolkits like GTK and Qt, so you can't just nest one window inside another and have everything automatically work).

tomspilman:
> But under OpenTK, you need to provide access to both  IGraphicsContext  and  IWindowInfo 

Then maybe this is where we need to provide one object that gives us access to both specifically for Mac/Linux platforms?


> You also need to provide callbacks to translate mouse position

XNA does not do that... we shouldn't either.  It is the responsibility of the app to process mouse events in whatever method makes sense for that application.  XNA/MonoGame is just handling the graphics.

hach-que:
> Then maybe this is where we need to provide one object that gives us access to both specifically for Mac/Linux platforms?

This is what `IEmbedContext` does.  The idea is that interface has different properties on it depending on what platform MonoGame is built for (so Windows / DirectX you probably could just have a window handle).

We could maybe try and put it in the service locator instead of a constructor argument, but at the moment there are no user callbacks between the service locator construction and the game platform construction, and you need to give the developer an opportunity to register the `IEmbedContext` in the service locator before the game platform is constructed if we want to go down that path.

> XNA does not do that... we shouldn't either. It is the responsibility of the app to process mouse events in whatever method makes sense for that application. XNA/MonoGame is just handling the graphics.

If you don't do that, then the results from `Mouse.GetState()` and `Keyboard.GetState()` will be wrong.  For someone that wants to handle input at the application layer, they can just make the callbacks null, but for me, I'm loading actual game code into an AppDomain inside the IDE (so that entities can behave as they would in the game at runtime), so I need the input APIs to report the right data back.

tomspilman:
> This is what `IEmbedContext` does.

Except you are passing this to `Game` and not `PresentationParameters`.  I see now you are looking for a different feature.

XNA and MonoGame support `GraphicsDevice` embedding.  You can create your own `GraphicsDevice` and have it render to your own control.  You provide your own update/tick/contentload/etc.

You want to embed a whole `Game` instance... a feature XNA did not support and I see little need for in MonoGame.  The existing XNA  `GraphicsDevice` embedding technique allows for the same functionality with a little more effort from the app doing the embedding.


> If you don't do that, then the results from `Mouse.GetState()` and 
> `Keyboard.GetState()` will be wrong.

I can't say I've tested this on XNA, but I wouldn't be surprised if they were wrong there too.

cra0zy:
I like what @hach-que is doing with this PR. gtk# sucks at processing Mouse/Keyboard input so allowing the game instance to be embedded and process events actually solves that problem. Also I think this is more useful than just embedding GraphicsDevice.

tomspilman:
>  I think this is much more useful than just embedding GraphicsDevice.

There are lots of features it would be useful for MonoGame to support for one person or another.  It isn't the only thing we have to consider.

What you are asking for is supporting a second way to embed MonoGame.  One that adds more complexity to several classes within MonoGame that we will have to continue to support in the future.

Since we already have support for embedding... one used by lots of XNA tools and users.... it makes me hesitant to add another.

hach-que:
Currently there is no way to embed MonoGame on non-Windows platforms right?  It's only supported for Windows / DirectX.  At the very least, for OpenTK construction, you can't just set a property on the PresentationParameters, because that's too late in the process (the graphics context and window info have to come from the embedding source, which means that it needs to be around during the game platform construction).

This functionality is a superset of the embedding functionality that currently exists; with this implementation you can use just the graphics APIs of MonoGame, but you can also use it to handle input if you want or need to.

Now of course, in my specific use case, if the embedding functionality was graphics-only, I _can_ work around this.  In Protogame, all of the input APIs are abstracted away from the `Mouse` / `Keyboard` APIs offered by MonoGame, so I can transparently redirect input for Protogame-based games.  But of course this means that the editing tools I'm building can only work for Protogame-based games, and won't be available for games written against MonoGame's APIs, because I have no way of hooking input up to work with arbitrary game structures.  I'd much rather have these tools work for both Protogame and MonoGame games, because that's more valuable for game developers.

nightst4r:
I have struggled getting the XNA way to work correctly in MG before, maybe the more recent versions work better.  #2856 Makes it easier for me by allowing me to specify the IntPtr and thus fixing the keyboard and mouse states, but I agree having the game embedable would allow people to create fantastic GUI tools for their games. Esp if it could be made to work on Mac and Linux as well.  The lack of easy ways to embed XNA was, for me, its biggest weakness and unfortunately MG seems to also struggle with it. 

But I do recall that Keyboard and mouse states were correct when XNA was embedded in a winform.

hach-que:
I'm closing this PR, but the changes for this branch will remain at https://github.com/hach-que/MonoGame/tree/embed-monogame.  For a continuously rebased and stable version of this patch on MonoGame, you can use https://github.com/RedpointGames/MonoGame.

PR #3712, opened 2015-04-07
Installer VS2015 Support

tomspilman:
This adds VS2015 support to the installer.  Mainly it just copies the templates for VS2010 desktop projects and VS2013 universal apps.

The next step is full Win10 support.

### DIFF
--- a/Installers/Windows/MonoGame.nsi
+++ b/Installers/Windows/MonoGame.nsi
@@ -279,6 +279,20 @@ Section "Visual Studio 2013 Templates" VS2013
 
 SectionEnd
 
+Section "Visual Studio 2015 Templates" VS2015
+
+  IfFileExists `$DOCUMENTS\Visual Studio 2015\Templates\ProjectTemplates\Visual C#\*.*` InstallTemplates CannotInstallTemplates
+  InstallTemplates:
+    SetOutPath "$DOCUMENTS\Visual Studio 2015\Templates\ProjectTemplates\Visual C#\MonoGame"
+    File /r '..\..\ProjectTemplates\VisualStudio2013\*.zip'
+    File /r '..\..\ProjectTemplates\VisualStudio2010\*.zip'
+    GOTO EndTemplates
+  CannotInstallTemplates:
+    DetailPrint "Visual Studio 2015 not found"
+  EndTemplates:
+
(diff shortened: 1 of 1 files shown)


nkast:
"The next step is full Win10 support."
8-)

Net.Native sounds promising but Windows 10 apps looks a lot like an upgraded .appx  from windows 8.1. You have to push a single app for all targets. I Can't find any info on new features in packaging. 
Here is the win8.1 packaging. if anyone has found info on Win10 changes please share.
https://msdn.microsoft.com/en-us/library/windows/apps/bg182885.aspx

So far you can package assets based on featurelevel,language and DPI, not on actual screensize or target platform (phone/desktop/xbox).
DPI is good only for apps to make icons look crisp but virtually same size, it doesn't suport games where I want to scale a background to fill the screen without pushing the same fat asset to low res phones where disk app size is important. A different case is a 1080 phone (HighDPI) and a 28'' 1080 desktop (mediumDPI), I had to either push a 1920x1080 background on all devices (including 800x480 phones) or end up with a 800x480 background on desktop.

tomspilman:
> You have to push a single app for all targets.

MS has said you don't have to use one app for all targets.  They recommend it, but you don't have to.  In fact C++ coders can't because they have to build EXEs for x86, x64, and ARM.  So some sort of feature exists for this, but I don't know the details beyond that.

Shqrdx:
I guess windows 8 support could be safely omitted in order to 8.1 and 10 implemented and simplification purpose.

tomspilman:
> I guess windows 8 support could be safely omitted for 8.1 and 10 implementation 

Actually I tested and had issues with both the Win 8 and Win 8.1 templates on VS2015.  So i'm thinking I just won't include those.  If you need to write a Windows/Phone 8 or 8.1 project just use VS2013.

We'll be adding Win10 templates as soon as I have the support complete.

Shqrdx:
@tomspilman And (sorry for tiresomeness) could you review my PR ?https://github.com/mono/MonoGame/pull/3705

dellis1972:
@tomspilman this looks ok. I can merge if you like?

tomspilman:
Ok... merging this with just the old VS2010 desktop project templates.

PR #3717, opened 2015-04-08
Killing improvements.md

Shqrdx:
https://github.com/mono/MonoGame/pull/3647 - Tom, you said it is useless file, so this branch just removes it from MonoGame.


tomspilman:
Did you check to see if the useful information in this doc was included in the existing reference docs?

Shqrdx:
The only useful information here is examples, but most new things is intuitive and even kid can understood how to use those..

Issue #3700, opened 2015-04-04
Issue spritebach and custom shaders (potentially other shader usage) [WinDX] (example code)


pfferrao:
It appears to be an bug with setting a texture on sampler, if an rendertarget swap occurs.

```C#
using Microsoft.Xna.Framework;
using Microsoft.Xna.Framework.Graphics;
using Microsoft.Xna.Framework.Input;

namespace DTBug {
    /// <summary>
    /// This is the main type for your game
    /// </summary>
    public class Game1 : Game {
        GraphicsDeviceManager graphics;
        SpriteBatch spriteBatch;
        Texture2D texture_0, texture_1;
        Vector2 sk_ratio, sk_jitter;
        RenderTarget2D rt;
        Effect effect;

        public Game1()
            : base() {
            graphics = new GraphicsDeviceManager(this);
            Content.RootDirectory = "Content";
        }

        /// <summary>
        /// Allows the game to perform any initialization it needs to before starting to run.
        /// This is where it can query for any required services and load any non-graphic
        /// related content.  Calling base.Initialize will enumerate through any components
        /// and initialize them as well.
        /// </summary>
        protected override void Initialize() {
            // TODO: Add your initialization logic here

            base.Initialize();
        }

        /// <summary>
        /// LoadContent will be called once per game and is the place to load
        /// all of your content.
        /// </summary>
        protected override void LoadContent() {
            // Create a new SpriteBatch, which can be used to draw textures.
            spriteBatch = new SpriteBatch(GraphicsDevice);

            // TODO: use this.Content to load your game content here
            texture_0 = Content.Load<Texture2D>(@"mono");
            texture_1 = Content.Load<Texture2D>(@"sketch");
            effect = Content.Load<Effect>(@"title");

            rt = new RenderTarget2D(GraphicsDevice, graphics.PreferredBackBufferWidth, graphics.PreferredBackBufferHeight, true, graphics.PreferredBackBufferFormat, graphics.PreferredDepthStencilFormat);

            sk_ratio = new Vector2(
                texture_0.Width / (float)texture_1.Width,
                texture_0.Height / (float)texture_1.Height);
        }

        /// <summary>
        /// UnloadContent will be called once per game and is the place to unload
        /// all content.
        /// </summary>
        protected override void UnloadContent() {
            // TODO: Unload any non ContentManager content here
        }

        /// <summary>
        /// Allows the game to run logic such as updating the world,
        /// checking for collisions, gathering input, and playing audio.
[...]

PR #3720, opened 2015-04-10
Fixing sampler texture being unbound when using render targets

### LINKED ISSUE #3710: [WinDX] Problem with shader textures/samplers and rendertarget swaps
Following in the issue #3700, this bug seems to affect all textures/samplers set by a shader, this example shows the problem using the stock DualTextureEffect:

```C#
using Microsoft.Xna.Framework.Input;

namespace DTBug {
    /// <summary>
    /// This is the main type for your game
    /// </summary>
    public class Game1 : Game {

        SpriteBatch spriteBatch;
        Texture2D texture_0, texture_1;
        RenderTarget2D rt;

        VertexBuffer v_buffer;
        DualTextureEffect dt_effect;

        public struct VertexDualTexture : IVertexType {

            Vector3 position;
            Vector2 tex1, tex2;

            public Vector3 Position { get { return position; } set { position = value; } }
            public Vector2 TextureCoordinate { get { return tex1; } set { tex1 = value; } }
            public Vector2 TextureCoordinate2 { get { return tex2; } set { tex2 = value; } }

            public readonly static VertexDeclaration dec = new VertexDeclaration
            (
                new VertexElement(0, VertexElementFormat.Vector3, VertexElementUsage.Position, 0),
                new VertexElement(12, VertexElementFormat.Vector2, VertexElementUsage.TextureCoordinate, 0),
                new VertexElement(20, VertexElementFormat.Vector2, VertexElementUsage.TextureCoordinate, 1)
            );

            VertexDeclaration IVertexType.VertexDeclaration { get { return dec; } }
        }


        VertexDualTexture[] quad = new[] {
            new VertexDualTexture() { Position = new Vector3(120, 120 + 256, 0), TextureCoordinate = new Vector2(0, 1), TextureCoordinate2 = new Vector2(0, 1) },
            new VertexDualTexture() { Position = new Vector3(120, 120, 0), TextureCoordinate = new Vector2(0, 0), TextureCoordinate2 = new Vector2(0, 0) },
            new VertexDualTexture() { Position = new Vector3(120 + 256, 120 + 256, 0), TextureCoordinate = new Vector2(1, 1), TextureCoordinate2 = new Vector2(1, 1) },
            new VertexDualTexture() { Position = new Vector3(120 + 256, 120, 0), TextureCoordinate = new Vector2(1, 0), TextureCoordinate2 = new Vector2(1, 0) }
        };

        public Game1()
            : base() {
        }

        /// <summary>
        /// Allows the game to perform any initialization it needs to before starting to run.
        /// This is where it can query for any required services and load any non-graphic
        /// related content.  Calling base.Initialize will enumerate through any components
        /// and initialize them as well.
        /// </summary>
        protected override void Initialize() {
            // TODO: Add your initialization logic here

            base.Initialize();
        }

        /// <summary>
        /// LoadContent will be called once per game and is the place to load
        /// all of your content.
        /// </summary>
        protected override void LoadContent() {
            // Create a new SpriteBatch, which can be used to draw textures.

            // TODO: use this.Content to load your game content here
            texture_0 = Content.Load<Texture2D>(@"mono");
            texture_1 = Content.Load<Texture2D>(@"sketch");

            dt_effect = new DualTextureEffect(GraphicsDevice) {
                Texture = texture_0,
                Texture2 = texture_1,
                FogEnabled = false,
                VertexColorEnabled = false
            };


            v_buffer = new VertexBuffer(GraphicsDevice, typeof(VertexDualTexture), 4, BufferUsage.None);
            rt = new RenderTarget2D(GraphicsDevice, 20, 20);
        }

        /// <summary>
        /// UnloadContent will be called once per game and is the place to unload
        /// all content.
        /// </summary>
        protected override void UnloadContent() {
            // TODO: Unload any non ContentManager content here
        }

        /// <summary>
        /// Allows the game to run logic such as updating the world,
        /// checking for collisions, gathering input, and playing audio.
        /// </summary>
        /// <param name="gameTime">Provides a snapshot of timing values.</param>
        protected override void Update(GameTime gameTime) {
            if (GamePad.GetState(PlayerIndex.One).Buttons.Back == ButtonState.Pressed || Keyboard.GetState().IsKeyDown(Keys.Escape))
                Exit();

            // TODO: Add your update logic here

            base.Update(gameTime);
        }

        /// <summary>
        /// This is called when the game should draw itself.
        /// </summary>
        /// <param name="gameTime">Provides a snapshot of timing values.</param>
        protected override void Draw(GameTime gameTime) {

            GraphicsDevice.SetRenderTarget(rt);

            GraphicsDevice.SetRenderTarget(null);

            var viewport = GraphicsDevice.Viewport;
            Matrix projection = Matrix.CreateOrthographicOffCenter(0, viewport.Width, viewport.Height, 0, 0, 1);
            Matrix halfPixelOffset = Matrix.CreateTranslation(-0.5f, -0.5f, 0);


            v_buffer.SetData<VertexDualTexture>(quad);
            GraphicsDevice.SetVertexBuffer(v_buffer);
            GraphicsDevice.BlendState = BlendState.AlphaBlend;

            dt_effect.World = Matrix.Identity;
            dt_effect.View = Matrix.Identity;
            dt_effect.Projection = halfPixelOffset * projection;

            foreach (var pass in dt_effect.CurrentTechnique.Passes) {
                pass.Apply();
                GraphicsDevice.DrawPrimitives(PrimitiveType.TriangleStrip, 0, 12);
            }

            base.Draw(gameTime);
        }
    }
}
```

The expected result would be:
expected_1

Instead I'm getting an blue window, running this example on WinGL yields the expected results. Removing the rentertarget swap or adding also "fixes" the problem:
```C#
GraphicsDevice.Textures[1] = null;
```

mrhelmut:
Following discussion on #3710, here's a patch made of @pfferrao & @tomspilman suggestion.

Fixes #3710 & #3700

### DIFF
--- a/MonoGame.Framework/Graphics/TextureCollection.DirectX.cs
+++ b/MonoGame.Framework/Graphics/TextureCollection.DirectX.cs
@@ -12,8 +12,14 @@ namespace Microsoft.Xna.Framework.Graphics
 
         internal void ClearTargets(GraphicsDevice device, RenderTargetBinding[] targets)
         {
-            ClearTargets(targets, device._d3dContext.VertexShader);
-            ClearTargets(targets, device._d3dContext.PixelShader);
+            if (_applyToVertexStage)
+            {
+                ClearTargets(targets, device._d3dContext.VertexShader);
+            }
+            else
+            {
+                ClearTargets(targets, device._d3dContext.PixelShader);
+            }
         }
 
         private void ClearTargets(RenderTargetBinding[] targets, SharpDX.Direct3D11.CommonShaderStage shaderStage)
@@ -30,6 +36,9 @@ namespace Microsoft.Xna.Framework.Graphics
             // Make one pass across all the texture slots.
             for (var i = 0; i < _textures.Length; i++)
             {
+                if (_textures[i] == null)
+                    continue;
+
                 if (_textures[i] != target0 &&
                     _textures[i] != target1 &&
                     _textures[i] != target2 &&


MonoGame.Framework/Graphics/TextureCollection.DirectX.cs
```diff
@@ -12,8 +12,14 @@ void PlatformInit()
...
-            ClearTargets(targets, device._d3dContext.VertexShader);
-            ClearTargets(targets, device._d3dContext.PixelShader);
+            if (_applyToVertexStage)
+            {
+                ClearTargets(targets, device._d3dContext.VertexShader);
+            }
+            else
+            {
+                ClearTargets(targets, device._d3dContext.PixelShader);
+            }
```
tomspilman: @tgjones - I wonder if we're better off just passing the DX pixel shader or vertex shader context on construction of the TextureCollection.  At least that cuts out some of the extra branches at runtime.

tomspilman:
Assuming all the unit tests pass i'm ok with merging this.

However it would be much better to add a unit test that exhibited the previous broken behavior.

mrhelmut:
I totally agree, the dirty flag has some quite intricate implications.

MonoGame.Framework/Graphics/TextureCollection.DirectX.cs
tgjones: I agree, that would be cleaner. `PlatformInit` would be the place to do it, but the device context isn't available there, because at the time the `TextureCollection` is created, the D3D device context hasn't been created yet. So it would require some refactoring.

tgjones:
I tried to help out by writing a test for this, but I got stuck. You'd need to access the underlying D3D context to find out what textures are *actually* bound, instead of what textures MonoGame thinks are bound, in order to write a test that verifies the fix. There isn't any public API in MonoGame to let you do that.

mrhelmut:
Yes, that's what I encountered too, but maybe we can workaround this by doing a graphical test. We could do this with a shader that combine s0 and s1 colors,  if the result is only s0 base color, that would mean that s1 got unbound (sampling a null texture always returns 0).

Also, the bug was showing up only after the first frame (which was correctly rendered), so the test should let a few frames to be rendered prior to test if anything is unbound.

What do you think?

tgjones:
That sounds sensible. MG's test suite actually supports rendering multiple frames for a single visual test, using `RunMultiFrameTest (captureCount: 5)`, for example.

And some of the more recent tests use `AssetTestUtility.CompileEffect` to compile effects on the fly. (The older tests use precompiled `.xnb` effects, but they're more difficult to work with.)

tomspilman:
Going to go ahead and merge this, but we really should follow up with a test.

PR #3721, opened 2015-04-11
Templates fixes

Shqrdx:
Changes for templates :
-Dots for Game title.
-UnloadContent now have more informative description - users can be more confused with word "all content" rather than "game-specific content".
-Removed useless call of base().
-Removed useless usings.


tomspilman:
Thanks!

Issue #3614, opened 2015-03-16
Buttons.LeftThumbstickUp incorrectly returns true when left stick pushed down


CrashLotus:
On Windows 7 with latest development build.  Xbox360-compatible joystick in USB port.  Buttons.LeftThunbstickUp returns true when the left stick is pushed DOWN(back) instead of when it is pushed up(forward).

tomspilman:
Isn't `Buttons.LeftThunbstickUp` an enum value?  How is it returning true?  Can you show us what API you are calling?

cra0zy:
Are you using OpenGL or DirectX version?

CrashLotus:
It is DirectX, although I don't see why that would matter.
That was rather imprecise of me.  I am referring to GamePadState.IsButtonDown().
As I test it further, I realize that both LeftThumbstickUp and LeftThumbstickDown are returning true when the left stick is pushed either up or down.  Here's a code snippet to illustrate the issue:

			System.Diagnostics.Debug.WriteLine(".");
			GamePadState padState = GamePad.GetState(PlayerIndex.One);
			if (padState.IsButtonDown(Buttons.LeftThumbstickUp))
				System.Diagnostics.Debug.WriteLine("Up");
			if (padState.IsButtonDown(Buttons.LeftThumbstickDown))
				System.Diagnostics.Debug.WriteLine("Down");
			if (padState.IsButtonDown(Buttons.LeftThumbstickLeft))
				System.Diagnostics.Debug.WriteLine("Left");
			if (padState.IsButtonDown(Buttons.LeftThumbstickRight))
				System.Diagnostics.Debug.WriteLine("Right");

Left and Right seem to work properly (although I think a larger dead zone would be appropriate).

cra0zy:
DirectX uses Xinput, while  OpenGL uses Sdl1.2, it matters...

tomspilman:
I've heard more reports of this issue occurring.

KakCAT:
Hi,

I think the problem is at AddThumbstickButtons in GamePad.XInput.cs . 
Switching to

            if (thumbY < -deadZone)
                result |= thumbStickDown;
            else if (thumbY > deadZone)
                result |= thumbStickUp;

seems to fix the problem. Could somebody confirm that negative Ys are down and positive Ys are up?

mrhelmut:
Hi,

I can confirm that XInput Y's are down and that @KakCAT code snipet fixes this issue.

tomspilman:
[...]

PR #3724, opened 2015-04-12
Fix for #3614

cra0zy:
Fix for #3614


tomspilman:
Thanks @cra0zy @KakCAT and @mrhelmut  !

Issue #3048, opened 2014-10-05
Windows Phone 8: SoundEffects loaded via ContentManager causing memory leak after resume from suspense


BDash2014:
My game is wp8 with latest nuget monogame/sharpdx.
First I though I have some bug in my implementation of libgdx (as game is port from libgdx to monogame), but I end up with ultra simple example:
new empty monogame project where I just load about 30 SoundEffects via ContentManager in LoadContent method and that is. Then after suspend/resume, memory increases by several MB and again and again every suspend/resume until out of memory exception...
Sound effects are waves converted to xnb format. I have no problem with other resources but with SoundEffect only.

tomspilman:
We've removed Win8/8.1 support in https://github.com/MonoGame/MonoGame/pull/5809.

Issue #1771, opened 2013-06-05
 memory leak when Resume game


MatriQ:
I think in method ContentManger.ReloadGraphicsContent memory leak when resume game. every times lost 20M.

snaiperskaya96:
Noticed that @nkast pushed lots of commits which are supposed to fix the leak, is it fixed or?

nkast:
All those commits you see above were rebase/sync with the develop branch.
As for WP8 it's no longer supported, see #5336.

You can find WP8 binaries with all my the fixes & optimizations here:
https://github.com/tainicom/Aether.Extras/tree/master/Samples/Libraries/MonoGame%203.5.1.9000/WindowsPhone8.0

Source code:
https://github.com/tainicom/MonoGame/commits/tnc_3_5_1_9000

snaiperskaya96:
Ah i see! Thanks

cra0zy:
This has been fixed, and the Windows 8.X platform support has also been dropped. Closing...

PR #3725, opened 2015-04-12
[WP8] fix memory leak when Resume game

nkast:
ContentManager & ContentReaders create new instances during
AssetReload(). The problem was with SoundEffectReader,
and with Shared resources like VertexBuffer, IndexBuffer,
Effect.

http://community.monogame.net/t/sharpdx-crashing-the-app-after-having-resumed-a-few-times/813/5
#1771
#3048

### DIFF
--- a/MonoGame.Framework/Content/ContentReaders/VertexBufferReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/VertexBufferReader.cs
@@ -11,13 +11,16 @@ namespace Microsoft.Xna.Framework.Content
     {
         protected internal override VertexBuffer Read(ContentReader input, VertexBuffer existingInstance)
         {
+			var buffer = existingInstance;
+
             var declaration = input.ReadRawObject<VertexDeclaration>();
             var vertexCount = (int)input.ReadUInt32();
             int dataSize = vertexCount * declaration.VertexStride;
             byte[] data = input.ContentManager.GetScratchBuffer(dataSize);
             input.Read(data, 0, dataSize);
 
-            var buffer = new VertexBuffer(input.GraphicsDevice, declaration, vertexCount, BufferUsage.None);
+            if(buffer == null)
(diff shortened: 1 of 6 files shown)


nkast:
Squash & rebase with latest develop branch.
What this does is keep  track of SharedResources in the ContentManager the same way it keeps track of loadedAssets. 
On reload it feed the references to `existingInstance` parameter. 
Also add checks for existingInstance and avoid to create new instances.

tomspilman:
@nkast - Don't worry about this build failure here...  i'm using your PR to test something on the build server.

MonoGame.Framework/Content/ContentReaders/VertexBufferReader.cs
```diff
@@ -34,11 +11,15 @@ class VertexBufferReader : ContentTypeReader<VertexBuffer>
     {
         protected internal override VertexBuffer Read(ContentReader input, VertexBuffer existingInstance)
         {
+			var buffer = existingInstance;
+
             var declaration = input.ReadRawObject<VertexDeclaration>();
-            var vertexCount = (int)input.ReadUInt32();
+            var vertexCount = input.ReadInt32();
```
KonajuGames: The field is a UInt32 (XNB file format document was available on MSDN but disappeared when create.msdn.com was retired), so it should be read as such.  That VertexBuffer takes an int for the vertexCount parameter is an inconsistency, but that is what XNA compatibility is all about.

nkast: You are right. post a fix for every file in the PR where is is an issues.

KakCAT:
Supposing 3.5 is probably the latest to support WP8 as said in #4340 , could please be this issue added to the 3.5 milestone?

Thanks!

KonajuGames:
I'm good with this.  @tomspilman ?

nkast:
Rebase with latest develop branch.

PR #3736, opened 2015-04-16
Gtk3 Pipeline

cra0zy:
Because I needed 2nd project for Gtk3 Pipeline I created Pipeline.Gtk3.definition, so if there is a better way to do this I would really like to know.


tomspilman:
So the Gtk3 project is only for Linux and I guess Gtk2 is Mac?  Or do we have 2 pipeline tools for Linux now?

cra0zy:
Gtk 3 is Linux only for now(saw no point in setting it up for Mac since Gtk only looks native on Linux), Gtk 2 is Mac and Linux. 

I was thinking on completely replacing Gtk 2 with Gtk 3, but I decided to keep both until Gtk Sharp 3 gets officially released(https://github.com/mono/gtk-sharp/releases). Also I was kind of inspired by how Unreal Engine keeps 4 gui systems for Linux(Gtk 2, Gtk 3, QT4, QT5).

tomspilman:
>  I decided to keep both until Gtk Sharp 3 gets officially released

Makes sense.

> Also I was kind of inspired by how Unreal Engine keeps 4 gui systems for Linux(Gtk 2, Gtk 3, QT4, QT5).

We really should drop Gtk2 as soon as we can.  Epic has the manpower/money to support things like that... we do not.

cra0zy:
If you want I can completely replace Gtk2 with Gtk3 pipeline for Linux in this PR, tho I can't replace it for Mac, I don't have a Mac so I wouldn't be able to test it out (in theory all that would need to be done is to add Mac lib locations in the config files found here: https://github.com/Mono-Game/MonoGame.Dependencies/tree/master/Gtk3)

tomspilman:
> If you want I can completely replace Gtk2 with Gtk3 pipeline for Linux in this PR,

If that will work for most Linux distros then I think we should go for it.

> I can't replace it for Mac

 Maybe @dellis1972 can investigate this at some point.

cra0zy:
Now Linux uses Gtk3 pipeline, and mac uses Gtk2 pipeline.

tomspilman:
Cool... thanks @cra0zy !

cra0zy:
Good news, I got OS X working in virtualbox :D

tomspilman:
> I got OS X working in virtualbox :D

Wow... I gave up last time I tried that.  It seemed impossible.

theZMan:
Me too - never mind the legalities the process seemed to be a random sequence of commands sourced from 20 different strange forums.. Is there a clean way of doing it now?

cra0zy:
This worked for me: https://youtu.be/CrpOjgBFMF0

cra0zy:
@dellis1972 If you get a chance, could you compile gtk 3 from: https://download.gnome.org/sources/gtk-sharp/2.99/gtk-sharp-2.99.3.tar.xz on OS X, I am not getting anywhere with it. Also the link is from: https://github.com/mono/gtk-sharp, release 2.99.3

PR #3737, opened 2015-04-16
If content compression fails, write uncompressed

### LINKED ISSUE #3664: [Pipeline,WIN] Compress=true crashes on small files (sample inside)
As in the following sample, a small wav file will crash the entire compile tree if Compress is set to TRUE

https://www.dropbox.com/s/lo9c33iuj8m9pet/CrashTest_Compress_SmallFile.zip?dl=0

KonajuGames:
LZ4 can sometimes fail with certain input byte streams. If this happens, write out an uncompressed file.

This is up here for testing.  Trying to work out how to write a test suite for this case.

Fixes #3664


tomspilman:
@KonajuGames - This good to merge?  The feature is rarely used as is and the best testing it will probably get is if we merge and get feedback from users.

tomspilman:
@KonajuGames - This good to merge?

KonajuGames:
Yeah.  Give it a test in the real world.

PR #3746, opened 2015-04-20
Winodows 10 UAP Support Pt.1

tomspilman:
This is the first pass at Windows 10 UAP support.

 - Added latest SharpDX for UAP builds.
 - Added Win10 support to Protobuild.
 - Fixed DX path to use SwapChainPanel.

The latest SharpDX has some changes that caused more `#if` statements.  We can in theory use the new SharpDX assemblies on all Windows Store/Phone apps which is what we should try to do in the future.

I need to submit an official patch to the Protobuild project so we don't have to maintain a custom xslt files to support Win10.

The build server has no Win10 support, so it is not building the Win10 binaries yet and we cannot include them in the installer until we do.  I'll look to solve this next.

~~~Also I will write up a VS project template soon.~~~


KonajuGames:
Brilliant.  I'll give this a test tonight.

tomspilman:
@KonajuGames - Thanks...  that would help a lot!

tomspilman:
Added a VS2015 template project for MonoGame Windows 10 UAPs.

tomspilman:
Things I suspect could be broken in this:

 - Rotation of the device might crash.
 - Could be that limiting the game to landscape only is broken.
 - Input and audio are totally untested.
 - We might not behave correctly on high DPI devices.
 - Haven't tested on a Windows 10 phone device.

If anyone tests any of this please let me know.  My access to Win10 systems to test is pretty limited right now.

KonajuGames:
I'll be grabbing a Nokia 630 today and putting Windows 10 preview on it.

azchohfi:
I'll give it a try ;) I'm using Windows 10 on my Surface Pro 3 and on my Lumia 1520.

azchohfi:
Btw, any specific test project I should test?

tomspilman:
> Btw, any specific test project I should test?

Nothing specific... anything you can test will help.

Note to generate the projects under protobuild: `protobuild -generate WindowsUAP`

azchohfi:
Worked pretty well on the SP3 and on my Lumia 1520.
I've only tried as simple project from the template and only using a simple SpriteFont to draw some text moving on the screen.
The only "problem" I've found was when running on the SP3 using a mouse, but only because there is a AppBar and the mouse is not visible, which seems like a glitch, but it's only because the mouse is not visible.
Also, the AppBar's button has no text and it also looks weird, but I don't think that matters because it's only a sample template.

tomspilman:
> but only because there is a AppBar and the mouse is not visible,

Try setting `IsMouseVisible` in the initialization maybe?

> only because there is a AppBar

The AppBar came from the Windows 10 UAP C++ DirectX XAML sample... I just left it in.  Maybe I should remove it from the template.

azchohfi:
Yes, that should work, but I just ignored it because I think it should be removed. It doesn't add nothing useful but the fact that knowing that you can mix your drawing with XAML, but i think that should be pretty obvious now.

mrhelmut:
Sounds great, would that work on any UAP platform? (e.g. consoles)
If so, I'd be glad to run some tests.

tomspilman:
> Sounds great, would that work on any UAP platform? (e.g. consoles)

UAP support for XB1 isn't yet release, but we fully expect this to work on there when it does.

KakCAT:
Ported one of my games to UAP platform without major monogame problems so far :) Only issue seems to be a touch offsetting in my Lumia 630 because of the initial rotation (the game is portrait) but I don't know if the problem is in my end.

I'll try to get a decent W10 dev rig and test ports extensively (I'm reluctant to upgrade my main rig to beta versions and the test W10 I have is an old dual core w/ 2GB, so it's quite slow)

tomspilman:
Thanks @KakCAT .  Any testing you can do against real games would help tremendously.

> Only issue seems to be a touch offsetting in my Lumia 630 because of the initial rotation

That could be buggy... we need more testing in that area.


> I'm reluctant to upgrade my main rig to beta versions and the test W10

I'm in the same boat...  just have my laptop using Win10.

PR #3747, opened 2015-04-21
Windows 10 Support Pt.2

tomspilman:
This PR cleans up the template project a bit removing some references to SharpDX and removing the unnecessary AppBar XAML.

We also added Win10 building into the build script.  It won't run until we get the Win10 build agent running.

### DIFF
--- a/ProjectTemplates/VisualStudio2015/WindowsUAP/Application.csproj
+++ b/ProjectTemplates/VisualStudio2015/WindowsUAP/Application.csproj
@@ -116,27 +116,6 @@
     <Reference Include="MonoGame.Framework">
        <HintPath>$(MSBuildProgramFiles32)\MonoGame\v3.0\Assemblies\WindowsUAP\MonoGame.Framework.dll</HintPath>
     </Reference>
-    <Reference Include="SharpDX">
-       <HintPath>$(MSBuildProgramFiles32)\MonoGame\v3.0\Assemblies\WindowsUAP\SharpDX.dll</HintPath>
-    </Reference>
-    <Reference Include="SharpDX.XInput">
-       <HintPath>$(MSBuildProgramFiles32)\MonoGame\v3.0\Assemblies\WindowsUAP\SharpDX.XInput.dll</HintPath>
-    </Reference>
-    <Reference Include="SharpDX.Direct2D1">
-       <HintPath>$(MSBuildProgramFiles32)\MonoGame\v3.0\Assemblies\WindowsUAP\SharpDX.Direct2D1.dll</HintPath>
-    </Reference>
-    <Reference Include="SharpDX.Direct3D11">
-       <HintPath>$(MSBuildProgramFiles32)\MonoGame\v3.0\Assemblies\WindowsUAP\SharpDX.Direct3D11.dll</HintPath>
-    </Reference>
-    <Reference Include="SharpDX.DXGI">
-       <HintPath>$(MSBuildProgramFiles32)\MonoGame\v3.0\Assemblies\WindowsUAP\SharpDX.DXGI.dll</HintPath>
-    </Reference>
-    <Reference Include="SharpDX.MediaFoundation">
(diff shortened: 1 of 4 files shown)


KonajuGames:
Merging.

KakCAT:
Hi, 

I'm having a problem creating the game window but I don't know if it's a bug or a feature :)

In example, put in the game constructor:

			graphics.PreferredBackBufferWidth=400;
			graphics.PreferredBackBufferHeight=240;

The game resulting window is not 400x240 but much bigger (I don't know if the window should appear as a small window, or Windows 10 "forces" you to use the resolution of the current monitor)

The thing is that resizing the window makes a weird effect (only a small region is painted instead of the full screen). (I know it's a terrible explanation, but the problem is obvious if you test it with the template project)

Thanks!

p.s. do you want the feedback/bugs in this thread or you prefer opening an issue for each piece?

KakCAT:
if (GamePad.GetState(PlayerIndex.One).Buttons.Back == ButtonState.Pressed) backPressed=true;

throws an exception in the Phone (both device and emulator)


    An exception of type 'System.DllNotFoundException' occurred in SharpDX.XInput.DLL but was not handled in user code
    Additional information: Unable to load DLL 'xinput1_4.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)


The exception is not thrown running on Windows 10.

tomspilman:
> do you want the feedback/bugs in this thread or you prefer opening an issue for each piece?

For now here is fine...  I will look into them.

> The game resulting window is not 400x240 but much bigger 

 I've had other reports of window sizing issues...  i'll investigate.


>  Unable to load DLL 'xinput1_4.dll'

 So either Win10 doesn't support XInput at all or just that specific version.  I suspect it doesn't support it at all and we need to account for that at runtime.

theZMan:
Check if XInput1_4.dll exists... I've seen that error on 100% clean installs of MonoGame before but I thought we had included it with the distribution. Maybe its not there in a source build??

XInput should work fine on Windows 10 - but maybe not with UAP apps

...or Maybe Not.. see Chucks Post here http://blogs.msdn.com/b/chuckw/archive/2015/03/30/visual-studio-tools-for-windows-10-preview.aspx
*Note that currently DirectX Tool Kit's GamePad class is a 'null' device for universal Windows apps as both the XInput compatibility library and the new gamepad API for universal Windows apps are not yet in the preview Windows 10 SDK.*

KakCAT:
I'm getting some strange values debugging in the Phone:

Device is Lumia 630 (854x480 with 'on screen buttons' -> 800x480)

My touch locations come between 0,0 and 800,480
GraphicsDevice.Viewport.Bounds is 0,0,533,320
GraphicsDevice.DisplayMode is 0,0,800,600    

initial Window.Current.CoreWindow.Bounds passed to XamlGame.Create is 0,0,533.3334,320   

I suspect that the "destination rendertarget" is not 800x480, not even close. (Graphics debugging is crashing my vs2015, so I draw a red vertical line every 40 pixels. Got an screenshot and separation is around 63 pixels)

KakCAT:
Hi,

about resolution, CoreWindow.Bounds is in DIP pixels and not in physical pixels. 
https://msdn.microsoft.com/en-us/library/windows/apps/xaml/windows.ui.core.corewindow.bounds.aspx

The client bounds should be multiplied by DisplayInformation.RawPixelsPerViewPixel to get the correct values. However I did that in my monogame copy but I still get the blurry low resolution display, maybe there are some other places this should be applied?

tomspilman:
> The client bounds should be multiplied by DisplayInformation.RawPixelsPerViewPixel 

Yeah... some MS people have informed me of this.  I'll look to get the code updated this weekend.

Also I learned an upcoming SDK adds a sort of `TryResizeWindow` function.  So we might be able to make UAP on the desktop behave just like XNA resizing the window to the preferred back buffer size.

PR #3751, opened 2015-04-22
Windows 10 Support Pt. 3

tomspilman:
This PR fixes the Windows installer to install the Windows 10 UAP assemblies.

I setup a temporary Win10 build agent to take care of the Win10 build... the main Windows build then copies those artifacts across when building the installer.

Also cleaned some old stuff in the installer from the old XNA content project support.


KonajuGames:
Merging.

PR #3762, opened 2015-04-23
Do not call MakeCurrent if context is already current on iOS

prollin:
`GraphicsContext.MakeCurrent` is not free and doesn't test if the context is already current.

This change also removes the `Obsolete` attribute on the `MakeCurrent()` method: the comment is enough to convey what needs to be done here and doesn't generate superfluous warnings (the warning generated by such an attribute is misleading in this case).

Note: on Android, MonoGame does call `GraphicsContext.IsCurrent()` however there is a bug in the OpenTK implementation that will make it return false in all cases. A workaround is explained [here](http://stackoverflow.com/questions/3091937/call-to-check-if-a-current-eglcontext-exists-in-android) but requires to fix OpenTK. It might be possible to go around OpenTK though.

### DIFF
--- a/MonoGame.Framework/iOS/iOSGameView.cs
+++ b/MonoGame.Framework/iOS/iOSGameView.cs
@@ -184,7 +184,7 @@ namespace Microsoft.Xna.Framework {
 				_glapi = new Gles11Api ();
 			}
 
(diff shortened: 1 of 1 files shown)


tomspilman:
Seems reasonable to me... merging!

PR #3765, opened 2015-04-25
Fixed exception when loading Windows 8 media library

Nezz:
On devices where library folders are missing or are damanged in some way


tomspilman:
Makes sense... will merge once it passes the build.

PR #3768, opened 2015-04-25
Model docs & fixes(remake)

Shqrdx:
remake of https://github.com/mono/MonoGame/pull/3738

### DIFF
--- a/MonoGame.Framework/Graphics/Model.cs
+++ b/MonoGame.Framework/Graphics/Model.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
 using System.Collections.Generic;
 
 namespace Microsoft.Xna.Framework.Graphics
--- a/MonoGame.Framework/Graphics/Model.cs
+++ b/MonoGame.Framework/Graphics/Model.cs
@@ -1,44 +1,49 @@
-﻿using Microsoft.Xna.Framework;
-using Microsoft.Xna.Framework.Content;
-using System;
-using System.Diagnostics;
+﻿using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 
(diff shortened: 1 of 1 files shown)


Shqrdx:
@tomspilman Accept ?

MonoGame.Framework/Graphics/Model.cs
```diff
@@ -1,44 +1,49 @@
...
+        /// </summary>
+        public object Tag { get; set; }
+
+		internal Model()
 		{
 
 		}
 
-		public Model(GraphicsDevice graphicsDevice, List<ModelBone> bones, List<ModelMesh> meshes)
+        internal Model(GraphicsDevice graphicsDevice, List<ModelBone> bones, List<ModelMesh> meshes)
```
tomspilman: I think it is worth keeping this API public.  There is really nothing that needs to be private or hidden here.

@KonajuGames ?

Shqrdx: In XNA there is no public constructor. I really doubt that someone needs to construct it manually from parts. These constuctors in general used only by content pipeline.

tomspilman: I am fully aware of how they are generally used and that it isn't public in XNA.  There is zero harm in making this public anyway and allows someone to build a model from code which is reasonable in some cases.

Shqrdx: Ok, than I open it and adds docs..

Shqrdx:
@tomspilman Should I also opens BuildHierarchy method ?

Shqrdx:
I guess not, because users already created model with this constructor with defined bones and this method is just confusing.

Shqrdx:
@tomspilman Accept ?

MonoGame.Framework/Graphics/Model.cs
Shqrdx: (after test) Yes this constructor works and you can really build model from code.

Shqrdx:
@tomspilman Recheck pls

tomspilman:
Ok...  i'm good with this.  Merging!

PR #3771, opened 2015-04-27
Added few random generation functions.

Shqrdx:
I think it is not bad addition. Many developers uses this common random generation algorithm and they are forced to write it itself or using other libraries. So my PR fixed this problem.

### DIFF
--- a/MonoGame.Framework/MathHelper.cs
+++ b/MonoGame.Framework/MathHelper.cs
@@ -214,7 +214,31 @@ namespace Microsoft.Xna.Framework
         {
             return value1 < value2 ? value1 : value2;
         }
-        
+
+        /// <summary>
+        /// Generates the pseudo-random <c>float</c> number within a range.
+        /// </summary>
+        /// <param name="random">Random number generator.</param>
+        /// <param name="min">Minimum value.</param>
+        /// <param name="max">Maximum value.</param>
+        /// <returns>Pseudo-random <c>float</c> number.</returns>
+        public static float GenFloat(Random random, float min, float max)
+        {
+            return (float)(min + random.NextDouble() * (max - min));
+        }
+
+        /// <summary>
+        /// Generates the pseudo-random <c>double</c> number within a range.
+        /// </summary>
+        /// <param name="random">Random number generator.</param>
+        /// <param name="min">Minimum value.</param>
+        /// <param name="max">Maximum value.</param>
+        /// <returns>Pseudo-random <c>double</c> number.</returns>
+        public static double GenDouble(Random random, double min, double max)
+        {
+            return (min + random.NextDouble() * (max - min));
+        }
+
         /// <summary>
         /// Interpolates between two values using a cubic equation.
         /// </summary>


KonajuGames:
Rather than being in MathHelpers, I think they would belong as an extension method.

```
public static class RandomExtensions
{
    public static double NextDouble(this Random random, double min, double max)
    {
        return min + NextDouble() * (max - min);
    }
}
```

then you simply call that method on your Random instance

```
    var value = random.NextDouble(10.0, 20.0);
```

Shqrdx:
Yes, but this can conflict with other frameworks on top of the MonoGame.

KonajuGames:
It might if they provide the same extension method.  I've never tried to build something that had the same extension method defined more than once.  But then, if those other libraries provide the same extension method, why not use that?

I still think this belongs as an extension method because it operates solely on the Random class and is extending the functionality of the Random class.

tomspilman:
> Yes, but this can conflict with other frameworks on top of the MonoGame.

Then that suggests we shouldn't be adding these extensions at all.  Which was my first thought seeing this PR.

That said there is something we should consider.

The main problem I've run into is that the `Random` on different versions of C# while using the same seeds do not result in the same sequence.  This is mainly between Mono and .NET.  

We've also found the stock .NET `Random` is pretty slow.  Even when doing integer randomization it under the hood uses doubles:

https://github.com/dotnet/coreclr/blob/cbf46fb0b6a0b209ed1caf4a680910b383e68cba/src/inc/random.h#L231

Worse MS can't really change the existing `Random` to fix these problems as it could break existing apps.

This all suggests adding our own `Random` that is aimed at game development needs.  Mainly a fast replacement for `Random` with support for all the data types and ranges.  We would also provide a helper template method to randomly shuffle elements in a list.

Shqrdx:
@tomspilman Thats would be cool ! Than.. I closes this PR ?

tomspilman:
I don't know.... @KonajuGames  @dellis1972  thoughts?

KonajuGames:
> The main problem I've run into is that the Random on different versions of C# while using the same seeds do not result in the same sequence.

We used our own RNG in console games for that reason, among others.

I think a custom RNG in MonoGame built for efficiency does have merit.

Shqrdx:
> It might if they provide the same extension method.
They are.

tomspilman:
>  a custom RNG in MonoGame built for efficiency does have merit.

Agreed... the goals would be:

 - High efficiency.
 - Great distribution.
 - Deterministic when you need it.
 - Can be totally random when you don't.
 - Provide complete set of min/max range functions.

Anything I missed?

Shqrdx:
I heard about special assembler instruction in modern processors for random generation, it would be good and extremly fast for use(of course it requires special checking before call).

Shqrdx:
I found it - http://en.wikipedia.org/wiki/RdRand

KonajuGames:
We don't have access to that through .NET (not easily anyway), and it is specific to newer Intel processors.  We would need a fully managed solution so it can be deterministic (return the same sequence) on all platforms regardless of processor or architecture.

tomspilman:
> I heard about special assembler instruction in modern processors for random generation

This is the least important part of `Random` in games.  If you are starting on this thinking about cryptographic randoms and hardware randoms you are already on the wrong path.

Focus on a good random generator in pure C# that is fast and has good distrubution.  You can always feed it a `Guid` as a seed which mostly solves making it a true random.

Shqrdx:
Yeah, true.

Shqrdx:
I suggest to create discussion about new class for random generation instead.

tomspilman:
@Shqrdx - Done https://github.com/mono/MonoGame/issues/3772

chrisgwilliams:
Mersenne Twister makes for an excellent fast random, with a long period and
good distribution.

Thanks,

Chris Williams

PR #3773, opened 2015-04-27
Added GraphicsMetrics struct

Shqrdx:
So I do not forgot about https://github.com/mono/MonoGame/issues/3624 and finally released it.
Test code...

	{

		// x, y, z
		// where x = batch count
		// y = draw call count 
		// z = primitives count

		var start1 = GraphicsDevice.GraphicsMetrics; // 0, 0, 0

		spriteBatch.Begin();

		spriteBatch.Draw(smile,new Vector2(0,0),Color.White);

		spriteBatch.Draw(smile, new Vector2(0, 0), Color.White);

		spriteBatch.End();

		var start2 = GraphicsDevice.GraphicsMetrics - start1; // 2, 1, 4

		spriteBatch.Begin();

		spriteBatch.Draw(smile, new Vector2(0, 0), Color.White);

		spriteBatch.Draw(smile, new Vector2(0, 0), Color.White);

		spriteBatch.Draw(smile, new Vector2(0, 0), Color.White);

		spriteBatch.End();

		var start3 = GraphicsDevice.GraphicsMetrics - start2; // 3, 1, 6

		var all = GraphicsDevice.GraphicsMetrics; // 5, 2, 10

		base.Draw(gameTime);
	}


Shqrdx:
@tomspilman Needs your help.

tomspilman:
This looks pretty good overall...  I'll add some inline notes.

@KonajuGames @Nezz @dellis1972 - Thoughts?

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -159,6 +159,20 @@ public GraphicsAdapter Adapter
             private set;
         }
 
+        internal Metrics _metrics;
+
+        /// <summary>
+        /// The rendering data for debugging and profiling. 
+        /// </summary>
+        public Metrics Metrics { get { return _metrics; } }
```
tomspilman: No reason why we couldn't allow someone to set a new set of metrics.  This allows you to store some metrics, reset and do some other work, and restore them later to continue.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -159,6 +159,20 @@ public GraphicsAdapter Adapter
...
+        /// The rendering data for debugging and profiling. 
+        /// </summary>
+        public Metrics Metrics { get { return _metrics; } }
+
+        internal bool _enableMetrics = true;
+
+        /// <summary>
+        /// Enables receiving metrics data. Disabling it is a small performance hint. <c>true</c> by default.
+        /// </summary>
+        public bool EnableMetrics { get { return _enableMetrics; } set { _enableMetrics = value; } }
```
tomspilman: I wonder if it is worth disabling them at all.  We're talking about very little overhead right now.

MonoGame.Framework/Graphics/Metrics.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// The count of draw calls.
+        /// </summary>
+        public ulong DrawCount { get { return _drawCount; } }
+
+        /// <summary>
+        /// The count of rendered primitives.
+        /// </summary>
+        public ulong PrimitivesCount { get { return _primitivesCount; } }
+        
+        internal void Reset()
```
tomspilman: Don't see why reset couldn't be public.  It can come in handy to end users.

MonoGame.Framework/Graphics/Metrics.cs
```diff
@@ -0,0 +1,56 @@
...
+            _primitivesCount = 0;
+        }
+
+        /// <summary>
+        /// Subtracts a <see cref="Metrics"/> from a <see cref="Metrics"/>.
+        /// </summary>
+        /// <param name="value1">Source <see cref="Metrics"/> on the left of the sub sign.</param>
+        /// <param name="value2">Source <see cref="Metrics"/> on the right of the sub sign.</param>
+        /// <returns>Result of the metrics subtraction.</returns>
+        public static Metrics operator -(Metrics value1, Metrics value2)
```
tomspilman: I would include an `operator +` as well.  This lets you measure costs of various subsystems then add them up later to a total.

KonajuGames:
It looks like a good start, though I would suggest some categories inside Metrics.  There will be graphics metrics now, and in the future we may also add audio, input and other categories.  Would it be better to have Metrics.Graphics, Metrics.Audio, Metrics.Input, etc?  Or leave it as a flat structure?

MonoGame.Framework/Graphics/SpriteBatcher.cs
```diff
@@ -224,7 +224,16 @@ public void DrawBatch(SpriteSortMode sortMode, Effect effect)
...
             // Determine how many iterations through the drawing code we need to make
             int batchIndex = 0;
-            int batchCount = _batchItemList.Count;
+            int batchCount = _batchItemList.Count;
+
+            if (_device._enableMetrics)
+            {
+                unchecked
+                {
+                    _device._metrics._batchCount += (ulong)batchCount;
```
tomspilman: I think the terminology here is weird....  the `batchCount` here really means sprite count.  You are really counting number of sprites being rendered (with text being one sprite per-character).

Is this what we really want to convey?  Should it be `Metrics.SpriteCount`?

KonajuGames: That does indeed look like it should be `SpriteCount`.

tomspilman:
@KonajuGames 

> Would it be better to have Metrics.Graphics, Metrics.Audio, Metrics.Input, etc?

I think I see what you mean.  I think maybe the struct needs to be renamed to `GraphicsMetrics`.  This way we could have `AudioMetrics` or others if we want later.

tomspilman:
@Shqrdx - If you rebase or merge the latest changes from develop the build error you hit should go away.

MonoGame.Framework/Graphics/SpriteBatcher.cs
tomspilman: So you would then use `SpriteCount` and `DrawCount` to understand how many sprites are being drawn and in how many draws that results in.  If you draw count starts to equal your sprite count you know that maybe you've got a small batch problem.

Or should we and an explicit `SpriteDrawCount` for counting draws from `SpriteBatch` vs draw calls from non-`SpriteBatch` stuff?

Shqrdx:
Ok, lets begin...

MonoGame.Framework/Graphics/Metrics.cs
Shqrdx: I tested - because Metrics is a struct and property of GraphicsDevice a call from Draw such as

> GraphicsDevice.Metrics.Reset() does not have effect

tomspilman: True...  `GraphicsDevice.Metrics.Reset()` would have no effect.  So lets remove it and instead use this technique:
```
GraphicsDevice.Metrics = new GraphicsMetrics();
```
Which works in all cases.

MonoGame.Framework/Graphics/SpriteBatcher.cs
Shqrdx: SpriteCount is enought in my opinion.

MonoGame.Framework/Graphics/GraphicsDevice.cs
Shqrdx: You mean add set accessor to this ?

tomspilman: Yes a setter.

tomspilman:
@KonajuGames - So is there more we want to track that would be useful on the graphics side?  

 - Render target swap count
 - Clear counts
 - Texture change counts
 - Shader change counts

Of all of these render target swaps and shader changes are generally the most expensive state changes you can do during rendering.

MonoGame.Framework/Graphics/GraphicsMetrics.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// Resetted every frame after draw at <see cref="GraphicsDevice.Present"/>.
+    /// </summary>
+    public struct GraphicsMetrics
+    {
+        internal ulong _spriteCount;
+        internal ulong _drawCount;
+        internal ulong _primitivesCount;
+
+        /// <summary>
+        /// The count of batches sended to <see cref="SpriteBatch"/>.
```
tomspilman: `The count of sprites and text characters rendered via SpriteBatch.`

Shqrdx:
@tomspilman All issues you noted, I fixed.

KonajuGames:
> So is there more we want to track that would be useful on the graphics side?

I think those cover the initial metric needs.  We can add more later if they are found to be useful.

Shqrdx:
@tomspilman Looks like everything is ok.. Accept ?

MonoGame.Framework/Graphics/GraphicsMetrics.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </summary>
+        public ulong DrawCount { get { return _drawCount; } }
+
+        /// <summary>
+        /// The count of rendered primitives.
+        /// </summary>
+        public ulong PrimitivesCount { get { return _primitivesCount; } }
+
+        /// <summary>
+        /// Subtracts a <see cref="GraphicsMetrics"/> from a <see cref="GraphicsMetrics"/>.
```
tomspilman: ```Returns the difference between two sets of metrics.```

MonoGame.Framework/Graphics/GraphicsMetrics.cs
```diff
@@ -0,0 +1,65 @@
...
+            return new GraphicsMetrics()
+            {
+                _spriteCount = value1._spriteCount - value2._spriteCount,
+                _drawCount = value1._drawCount - value2._drawCount,
+                _primitivesCount = value1._primitivesCount - value2._primitivesCount
+            };
+        }
+
+        /// <summary>
+        /// Adds two metrics.
```
tomspilman: ```Returns the combination of two sets of metrics.```

MonoGame.Framework/Graphics/GraphicsMetrics.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </summary>
+        public ulong PrimitivesCount { get { return _primitivesCount; } }
+
+        /// <summary>
+        /// Subtracts a <see cref="GraphicsMetrics"/> from a <see cref="GraphicsMetrics"/>.
+        /// </summary>
+        /// <param name="value1">Source <see cref="GraphicsMetrics"/> on the left of the sub sign.</param>
+        /// <param name="value2">Source <see cref="GraphicsMetrics"/> on the right of the sub sign.</param>
+        /// <returns>Result of the metrics subtraction.</returns>
+        public static GraphicsMetrics operator -(GraphicsMetrics value1, GraphicsMetrics value2)
```
tomspilman: @KonajuGames - Are we maybe better off having like a `Subtract` or `Difference` method instead of an `operator -` here?  Same for the `operator +` below?

tomspilman:
@Shqrdx - Lets leave this open for a day or two to give others some time to comment.  It is important enough that we don't need to rush it.

Shqrdx:
@tomspilman Ok

Shqrdx:
One more question - should I rename PrimitvesCount to PrimitiveCount(without "s") or it is fine ?

tomspilman:
> should I rename PrimitvesCount to PrimitiveCount(without "s")

Yes...  you should.  I totally missed that.

dellis1972:
Do we need a way to disable metrics? Maybe not for this PR but at some point when there is a bit more overhead. Disabling might be more important for mobile platforms given they are usually lower power devices.

Shqrdx:
> Do we need a way to disable metrics?

I initially pushed this change but Tom said it is unnecessary.

KonajuGames:
> Disabling might be more important for mobile platforms given they are usually lower power devices.
It might cost almost as much to do the if condition and have it fail than to update the metric anyway. This rule also applies to shaders where it is often quicker to do the work and throw it away than check if you should have not done the work in the first place.

tomspilman:
> Do we need a way to disable metrics?

I currently don't think so...  and probably won't need to at all even if we add more metrics to it.  We're talking like 2 adds per-draw...   that is so little compared to everything else even on mobile that I highly doubt it.


>  but Tom said it is unnecessary.

That isn't accurate... I said:

> > I wonder if it is worth disabling them at all. We're talking about very little overhead right now.

I was expecting a discussion not removal.

Shqrdx:
Sorry for misunderstood, it is easy for returning anyway.

MonoGame.Framework/Graphics/GraphicsMetrics.cs
```diff
@@ -0,0 +1,65 @@
...
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+namespace Microsoft.Xna.Framework.Graphics
+{
+    /// <summary>
+    /// The rendering information from <see cref="GraphicsDevice"/>. 
+    /// Useful for debugging and profiling purposes. 
+    /// Resetted every frame after draw at <see cref="GraphicsDevice.Present"/>.
+    /// </summary>
```
tomspilman: I think you should change the docs here.   Here you should say something like:
```
A snapshot of rendering statistics from `GraphicsDevice.Metrics` to be used for runtime debugging and profiling.
```
The other part...
```
The metrics are reset every frame after draw within <see cref="GraphicsDevice.Present"/>. 
```
Should be moved to the docs for `GraphicsDevice.Metrics` as that is where that information matters.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -159,6 +159,13 @@ public GraphicsAdapter Adapter
             private set;
         }
 
+        internal GraphicsMetrics _graphicsMetrics;
+
+        /// <summary>
+        /// The rendering information for debugging and profiling. 
+        /// </summary>
+        public GraphicsMetrics GraphicsMetrics { get { return _graphicsMetrics; } set { _graphicsMetrics = value; } }
```
tomspilman: Maybe this should be just `GraphicsDevice.Metrics`.  Also the docs could be better... see my comments below.

KonajuGames: So the typename stays the same. The property should be renamed to `Metrics`.

tomspilman: Right!

Shqrdx:
Damn, better I close it and remake.

tomspilman:
You really need to read the build log when you get an error.  You would have seen this had nothing to do with you.

Shqrdx:
Yes, I readed this. Probably it says - rebase. I rebased it and get another error.

PR #3776, opened 2015-04-28
Docs fixes

Shqrdx:
Batch of doc fixes.
Changes:
 -All private fields for properties are now also readonly.
 -Unification to all constructors
 -Unification to transform methods such as ToVector2
-Properties docs are rewrited in form without get and set
-Rewrited Color to more readable form.

### DIFF
--- a/MonoGame.Framework/BoundingFrustum.cs
+++ b/MonoGame.Framework/BoundingFrustum.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Xna.Framework
         #region Properties
 
         /// <summary>
-        /// Gets or sets the <see cref="Matrix"/> of the frustum.
+        /// The transformation <see cref="Matrix"/> of this frustum.
         /// </summary>
         public Matrix Matrix
         {
@@ -123,7 +123,7 @@ namespace Microsoft.Xna.Framework
         #region Constructors
 
         /// <summary>
-        /// Creates a new instance of <see cref="BoundingFrustum"/> class.
+        /// Constructs a bounding frustum.
         /// </summary>
         /// <param name="value">Combined matrix which usually is (View * Projection).</param>
         public BoundingFrustum(Matrix value)
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -16,6 +16,20 @@ namespace Microsoft.Xna.Framework
     [DebuggerDisplay("{DebugDisplayString,nq}")]
     public struct Color : IEquatable<Color>
     {
+        #region Private Fields
+
+        /// <summary>
+        /// ARGB.
+        /// </summary>
+        private uint _packedValue;
+
+        #endregion
+
+        #region Constructors
+
+        /// <summary>
+        /// Private static initializator for colors.
+        /// </summary>
         static Color()
         {
             TransparentBlack = new Color(0);
@@ -162,9 +176,7 @@ namespace Microsoft.Xna.Framework
             Yellow = new Color(0xff00ffff);
             YellowGreen = new Color(0xff32cd9a);
         }
-	// ARGB
-        private uint _packedValue;
-	  
+
         private Color(uint packedValue)
         {
             _packedValue = packedValue;
@@ -174,8 +186,8 @@ namespace Microsoft.Xna.Framework
 			//_packedValue = (packedValue & 0xff00ff00) | ((packedValue & 0x000000ff) << 16) | ((packedValue & 0x00ff0000) >> 16);
         }
 
-	/// <summary>
-        /// Creates a new instance of <see cref="Color"/> struct.
+        /// <summary>
+        /// Constructs a color.
         /// </summary>
         /// <param name="color">A <see cref="Vector4"/> representing color.</param>
         public Color(Vector4 color)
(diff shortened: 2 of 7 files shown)


MonoGame.Framework/BoundingFrustum.cs
```diff
@@ -123,7 +123,7 @@ internal string DebugDisplayString
         #region Constructors
 
         /// <summary>
-        /// Creates a new instance of <see cref="BoundingFrustum"/> class.
+        /// Constructs a bounding frustum.
```
tomspilman: So this may sound like i'm picking on you @Shqrdx or I'm just being mean...  that is not my intention.  I'm just *very* passionate about getting good docs into MonoGame.

This is an example of bad documentation.  It does nothing but repeat the obvious...  you know it is a constructor... you know it is a bounding frustum.  Yes in some classes there is not more to say...  as the constructor does nothing useful.. but not in this case.  

Let's actually explain what the constructor is doing:
```
/// Constructs the frustum by extracting the view planes from a matrix.
```
This actually provides information that isn't included in the function signature and will inform new users as to exactly how this class works.

MonoGame.Framework/Color.cs
```diff
@@ -16,6 +16,20 @@ namespace Microsoft.Xna.Framework
...
+        /// </summary>
+        private uint _packedValue;
+
+        #endregion
+
+        #region Constructors
+
+        /// <summary>
+        /// Private static initializator for colors.
+        /// </summary>
```
tomspilman: Lets not start documenting private members.  They don't show up in the documentation and we have plenty of public APIs that need good docs as it is...  lets focus on making them better instead.

MonoGame.Framework/Color.cs
```diff
@@ -174,8 +186,8 @@ private Color(uint packedValue)
 			//_packedValue = (packedValue & 0xff00ff00) | ((packedValue & 0x000000ff) << 16) | ((packedValue & 0x00ff0000) >> 16);
         }
 
-	/// <summary>
-        /// Creates a new instance of <see cref="Color"/> struct.
+        /// <summary>
+        /// Constructs a color.
```
tomspilman: How about...
```
Constructs an RGBA color from the XYZW unit length components of a vector.
```
At least then we are telling users the implied order (X is R, Y is G, etc.) and that they are assumed to be 0 to 1.

This same idea can be applied to the other constructors here.

MonoGame.Framework/Color.cs
```diff
@@ -188,8 +200,8 @@ public Color(Vector4 color)
             A = (byte)MathHelper.Clamp(color.W * 255, Byte.MinValue, Byte.MaxValue);
         }
 
-	/// <summary>
-        /// Creates a new instance of <see cref="Color"/> struct.
+        /// <summary>
+        /// Constructs a color.
```
tomspilman: How about mentioning what happens with the alpha?

MonoGame.Framework/Color.cs
```diff
@@ -202,8 +214,8 @@ public Color(Vector3 color)
             A = 255;
         }
 	
-	/// <summary>
-        /// Creates a new instance of <see cref="Color"/> struct.
+	    /// <summary>
+        /// Constructs a color.
         /// </summary>
         /// <param name="color">A <see cref="Color"/> for RGB values of new <see cref="Color"/> instance.</param>
         /// <param name="alpha">Alpha component value.</param>
```
tomspilman: How about mentioning the expected range for the alpha component?

MonoGame.Framework/Color.cs
```diff
@@ -296,127 +308,98 @@ public Color(float r, float g, float b, float alpha)
...
+
+        #region Public Properties
+
         /// <summary>
-        /// Gets or sets the blue component.
+        /// The alpha component of this color.
         /// </summary>
         [DataMember]
-        public byte B
+        public byte A
```
tomspilman: This is why we avoid re-sorting code in a file.  It makes this diff much more difficult to understand.  You should avoid it in the future.

MonoGame.Framework/Color.cs
```diff
@@ -296,127 +308,98 @@ public Color(float r, float g, float b, float alpha)
...
                 unchecked
                 {
-                    return (byte) (this._packedValue >> 16);
+                    return (byte)(this._packedValue >> 24);
                 }
             }
             set
             {
-                this._packedValue = (this._packedValue & 0xff00ffff) | ((uint)value << 16);
+                this._packedValue = (this._packedValue & 0x00ffffff) | ((uint)value << 24);
```
tomspilman: This is why we avoid re-sorting code in a file. I have to now check that you didn't change functionality here because at first glance it looks like you changed how packing works.

It makes this diff much more difficult to understand. You should avoid it in the future.

Shqrdx:
My bad,my bad. I better recreate this PR to more compact form.

MonoGame.Framework/BoundingFrustum.cs
Shqrdx: I understand, I was thinking - lets write same constuctors as Matrix to all documented classes for unification. And I was wrong.

tomspilman: > lets write same constuctors as Matrix to all documented classes for unification

Right...  we have to remember the primary goal is to communicate how things work.

PR #3777, opened 2015-04-28
[Gtk Pipeline] About dialog updated

cra0zy:
Things done:
<ul>
<li>Replaced AboutDialog with Gtk preset About Dialog, looks much nicer :D</li>
<li>Corrected Name</li>
<li>There is a website button in the about dialog instead of just opening the webpage when About submenu is clicked</li>
</ul>


Tools/Pipeline/Common/AssemblyAttributes.cs
```diff
@@ -81,6 +81,14 @@ public static string AssemblyCompany
 				return ((AssemblyCompanyAttribute)attributes[0]).Company;
 			}
 		}
+
+        public static string[] Developers
+        {
+            get
+            {
+                return new string[] { "cra0zy (Harry)", "dellis1972 (Dean Ellis)", "dneelyep (Daniel Neel)", "emarcotte (Eugene Marcotte)", "JamesLupiani (James Lupiani)", "jamesford42 (James Ford)", "KonajuGames (Steve 'Sly' Williams)", "mgarstenauer (Martin Garstenauer)", "nightst4r (Mark)", "nkast (Nikos Kastellanos)", "polsys (Petri Laarne)", "RayBatts (Ray Batts)", "Shqrdx (Yuri Roubinsky)", "thiagoabreu (Thiago 'Jedi' Abreu)", "tomspilman (Tom Spilman)" };
```
tomspilman: @KonajuGames - Do we want to allow this?

IMO we don't.  If we allow this then why not a credit for any contribution made?  What about every old commit in history?  Where does this all go?  When does it end?

tomspilman: @KonajuGames @dellis1972 - Do we want to allow this?

IMO we don't. If we allow this then why not a credit for any contribution made? What about every old commit in history? Where does this all go? When does it end?

cra0zy: It's only for the Pipeline Tool, and I went over every file of the Pipeline Tool and took out contributors... Since you don't like it, I'll remove it immediately.

tomspilman: > Since you don't like it, I'll remove it immediately.

Wait... lets see what others think first.

KonajuGames: Agreed. No other tool in MonoGame lists individual contributors. It should
just mention MonoGame and direct them to the website for more information.
If someone really wants to find out who worked on it, they can go to source
control history.

tomspilman: Or we can redirect people to the http://www.monogame.net/about or contributors page:

 https://github.com/mono/MonoGame/graphs/contributors

Of which @cra0zy is already the 13th largest contributor to MonoGame in the few short months he has been helping with the Pipeline tool and Linux... thanks @cra0zy!

Actually I need to refresh the MonoGame About page anyway...  I will be sure to link the GitHub contributors page to it.

cra0zy: Woho!!

dellis1972: I like the sentiment. But if we are going to list people who have
contributed perhaps we should just setup a contributors page/section on the
main monogame.net site and list everyone who has contributed over the
years. We can easily generate the data for a page with something like

git log --format='%aN <%aE>' | awk '{arr[$0]++} END{for (i in arr){print
arr[i], i;}}' | sort -rn


 Although we would probably want to only print the username / Name rather
than email addressed (this is what this command does :) )

tomspilman:
Thanks @cra0zy .  Going to merge this now.

I am however getting concerned that the GTK version is getting to be very different from the WinForm version of the Pipeline tool.  We need to get help to bring them into alignment feature wise soon.

cra0zy:
I am keeping tabs on what changes I have to apply to WinForms Pipeline, that are in Gtk Pipeline. Currently those are:
- Rename
- New Reference Editor Dialog

If you want me to do anything extra to WinForms Pipeline just say it, I have 1 more PR for Gtk Pipeline and that I'll setup PRs for WinForms Pipeline.

Issue #3739, opened 2015-04-17
Replace copy.sh with .build file


cra0zy:
So with the last PR I separated generation part of the linux installer so that part of the installer can be done on the build server. 

The idea is as following, replace the file: https://github.com/mono/MonoGame/blob/develop/Installers/Linux/copy.sh with ".build" file.

Currently copy.sh file is used to copy all the needed files to the Installers/Linux/Installer/, and this folder could be just added to an archive and be given to the user.

tomspilman:
> so that part of the installer can be done on the build server. 

Why part of and not all?  Where will the rest be done?  What files need to be moved around (we have to setup the build server to pass the right file around)?

cra0zy:
> Why part of and not all?

Because this way you can use windows build server to create linux installer. After copy script is done you have a folder that contains:
Data (Folder)
Makeself (Folder)
generate.sh (File)
generate_deb.sh (File)

generate.sh can be run from linux or mac (or cygwin) and will generate linux .run package.
generate_deb has to be run from linux and will generate .deb package for Ubuntu and Debian.

tomspilman:
> Because this way you can use windows build server to create linux installer.

But we can't really...  it only creates a part of the installer.  We still need a Linux box to do the rest.

IMO we should solve the core problem... we need to add a Linux box to the build system.


> generate_deb.sh has to be run from linux and will generate .deb 
> package for Ubuntu and Debian.

Cab `generate.sh` also be run from Ubuntu?  This means all we need is one Ubuntu box then correct?

cra0zy:
> But we can't really... it only creates a part of the installer. We still need a Linux box to do the rest.
> 
> IMO we should solve the core problem... we need to add a Linux box to the build system.

The rest (running one of the generate scripts) takes less than 30 seconds, and when I separated the generation, I was thinking that this could be used for nightly builds, and for stable builds we could just run the generate scripts from any Linux manually.

Adding a Linux build server would also be a way of doing it. In which case copy.sh could be just connected to the .build file, and not have to be replaced.

> Cab generate.sh also be run from Ubuntu? This means all we need is one Ubuntu box then correct?

Ubuntu is Linux.▼▼

> generate.sh can be run from linux or mac (or cygwin) and will generate linux .run package.

PR #3778, opened 2015-04-28
Connected Linux installers to nant scripts

cra0zy:
#3739


cra0zy:
@tomspilman I've setup the script so that if it's run from Windows/Mac, you get file: 
```
Installers/Linux/LinuxInstaller.zip
```
Which contains generate scripts for installers. This should probably get connected to artifacts.

If you run it from Linux you get:
```
Installers/Linux/monogame-linux.run
Installers/Linux/monogame-pipeline.deb
```

cra0zy:
Going threw the build log it seems it worked. This is complete than.

tomspilman:
Looks reasonable... i'm merging now.

PR #3779, opened 2015-04-29
Docs fixes

Shqrdx:
Small docs fixes.

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -174,8 +174,8 @@ namespace Microsoft.Xna.Framework
 			//_packedValue = (packedValue & 0xff00ff00) | ((packedValue & 0x000000ff) << 16) | ((packedValue & 0x00ff0000) >> 16);
         }
 
-	/// <summary>
-        /// Creates a new instance of <see cref="Color"/> struct.
+        /// <summary>
+        /// Constructs an RGBA color from the XYZW unit length components of a vector.
         /// </summary>
         /// <param name="color">A <see cref="Vector4"/> representing color.</param>
         public Color(Vector4 color)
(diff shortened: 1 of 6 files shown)


Shqrdx:
@tomspilman Needs your checking.

MonoGame.Framework/Graphics/Viewport.cs
```diff
@@ -158,7 +158,7 @@ public Rectangle TitleSafeArea
 		}
 
         /// <summary>
-        /// Creates a new instance of <see cref="Viewport"/> struct.
+        /// Constructs a viewport.
```
tomspilman: How about mentioning that the `MinDepth` will be 0 and `MaxDepth` will be 1.

MonoGame.Framework/Point.cs
```diff
@@ -78,7 +78,7 @@ public Point(int x, int y)
         }
 
         /// <summary>
-        /// Creates a new instance of <see cref="Point"/> struct, with the specified position.
+        /// Constructs a point from one value.
```
tomspilman: `Constructs a point with X and Y set to the same value.`

MonoGame.Framework/Vector2.cs
```diff
@@ -108,7 +108,7 @@ public Vector2(float x, float y)
         }
 
         /// <summary>
-        /// Creates a new instance of <see cref="Vector2"/> struct, with the specified position.
+        /// Constructs a 2d vector from a scalar.
```
tomspilman: Make this similar to the one for the `Point` class.

MonoGame.Framework/Color.cs
```diff
@@ -216,9 +216,9 @@ public Color(Color color, int alpha)
             B = color.B;
             A = (byte)MathHelper.Clamp(alpha, Byte.MinValue, Byte.MaxValue);
         }
-	
-	/// <summary>
-        /// Creates a new instance of <see cref="Color"/> struct.
+
+        /// <summary>
+        /// Constructs an RGBA color from color and alpha value(0.0-1.0).
```
tomspilman: Instead of putting these in parenthesis... just make it say this in the description:
```
Constructs an RGBA color from a Color and an alpha value.
```
Then in the description for the `alpha` parameter do:
```
The alpha value which will be clamped 0 to 1.
```
It would be better to do something like this for the other cases below.

Shqrdx:
Is it good now ? however I changed your suggestion about Color constructor a bit...

tomspilman:
Works for me... merging!

PR #3785, opened 2015-04-29
Fix bug using SpriteBatch layerDepth on OpenGL

### LINKED ISSUE #3750: Sprites not rendering with rebuilt effects
Hi,

After a long while not pulling any changes from github, I just updated my linux system with the latest of the develop branch. After doing so, none of my sprites render at all in a simple 2d game.

Doing a bisect from v3.3 to the latest seems to indicate that commit c089aa8b3ba93b10b09800d8fc9f272d202336a3
 from https://github.com/mono/MonoGame/pull/3706 is the first one where they all disappeared.

A simple game class that reproduces the issue is as follows... requires a texture2d named texture/player...

````csharp
using Microsoft.Xna.Framework.Input;

namespace DeathDeathEvolution {
    public class DDE : Game {
        Texture2D image;
        SpriteBatch spriteBatch;

        public DDE() {
        }

        protected override void LoadContent() {
            image = Content.Load<Texture2D>("texture/player");
        }

        protected override void Update(GameTime gameTime) {
            if (Keyboard.GetState().IsKeyDown(Keys.Escape)) {
                Exit();
                return;
            }
            base.Update(gameTime);
        }

        protected override void Draw(GameTime gameTime) {

            base.Draw(gameTime);

            spriteBatch.Begin(
                sortMode: SpriteSortMode.BackToFront,
                samplerState: SamplerState.PointClamp,
                blendState: BlendState.AlphaBlend
            );

            spriteBatch.Draw(
                image,
                new Vector2(100, 100),
                null,
                null,
                null,
                0,
                null,
                Color.White,
                SpriteEffects.None,
                1
            );
            spriteBatch.End();
        }
    }
}
````

When running this game I don't get any (abnormal) output, just the previously reported linux warnings from opentk which are on another issue:

````
X Error of failed request:  BadCursor (invalid Cursor parameter)
  Major opcode of failed request:  95 (X_FreeCursor)
  Resource id in failed request:  0x2e00005
  Serial number of failed request:  226
  Current serial number in output stream:  228
AL lib: (EE) alc_cleanup: 1 device not closed
````
I'm not sure what other information is relevant here. Let me know what else I can provide.

### PR DESCRIPTION
**Pull request by tgjones**
Fixes #3750.

OpenGL previously needed this depth fixup, but now that there's a depth fixup at the shader level, this fixup in SpriteBatch isn't needed, and was in fact wrong.

OpenGL can just use the same transform matrix as other platforms.

### CODE REVIEW

**Comment by tgjones** [comment]
@emarcotte Please could you test that this fixes the issue for you?

**Comment by tomspilman** [comment]
I may have to merge this without @emarcotte verifying things.

**Comment by tgjones** [comment]
@tomspilman It's up to you whether you want to merge this. I'm pretty sure it fixes the issue, but I'm also a little nervous about making this kind of change immediately before a release.

File: MonoGame.Framework/Graphics/SpriteBatch.cs [hunk 1]
```diff
@@ -125,11 +125,9 @@ void Setup()
 			var vp = gd.Viewport;
 
 		    Matrix projection;
-#if PSM || DIRECTX
             Matrix.CreateOrthographicOffCenter(0, vp.Width, vp.Height, 0, -1, 0, out projection);
```
tomspilman: So basically DX and GL both needed the same matrix construction?

tgjones:
I guess it depends on how much we trust our SpriteBatch tests :) With this fix, they all pass on OpenGL except for the one that depends on compiling an effect, which isn't yet supported from the OpenGL test suite.

tomspilman:
> I guess it depends on how much we trust our SpriteBatch tests 

I trust them pretty well...  but I worry about gaps in coverage.

Like I don't think we have a test that uses the `Matrix` parameter for the `SpriteBatch` to manipulate the final sprite transform.  That seems like maybe the sort of thing that could be screwed up by this projection change... maybe.

MonoGame.Framework/Graphics/SpriteBatch.cs
tgjones: Yes - but only since I made that change to fixup the depth values in GLSL, such that depths output from GLSL match the Direct3D convention. As soon as I made that change, it made this fixup in SpriteBatch both redundant and incorrect.

tomspilman: >  As soon as I made that change, it made this fixup in SpriteBatch both redundant and incorrect.

Ahh... I get it now.

tgjones:
> Like I don't think we have a test that uses the Matrix parameter

Actually we do have a test for supplying a transform matrix:
https://github.com/mono/MonoGame/blob/develop/Test/Framework/Visual/SpriteBatchTest.cs#L205

But you're right in principle, I'm sure there are many gaps in coverage.

MonoGame.Framework/Graphics/SpriteBatch.cs
tgjones: (and actually the `Draw_with_SpriteSortMode` tests would have told us that, if the OpenGL test suite was up and running... but there are so many failing tests in there right now, that I didn't spot that one.)

tomspilman:
Well... gonna take the gamble.  Merging!  

Thanks for the quick work @tgjones !

emarcotte:
I can test in an hour or so. Just getting out of work :(

emarcotte:
@tgjones @tomspilman Looks good from my tests!

tomspilman:
Fantastic... thanks for verifying it for us @emarcotte !

PR #3789, opened 2015-04-30
Cross-platform Random number generator

WardBenjamin:
This fulfills https://github.com/mono/MonoGame/issues/3772, which requires a random number generator that is consistent over platforms. It contains only one utility function as of yet, `ulong MathHelper.Random.GenerateSeed()`.


WardBenjamin:
Oh shoot, what's the problem? I merged into develop this time.

tomspilman:
Looking... one sec.

tomspilman:
You haven't updated your develop branch in your fork.  You need to merge in the latest upstream changes from develop into your develop branch in your fork.  Then your fork is up to date with the main repo.

I know how to do this via command line if you need instructions.

WardBenjamin:
Hmm, I thought I did that.

WardBenjamin:
Figured it out, I just forgot to push. Fix'd.

WardBenjamin:
> I know how to do this via command line if you need instructions.

You'd probably be surprised at how well I can use the command line. I am fairly adept at bash as well as git commands specifically. *cough merging orphan branch into master because team member was being stupid*

tomspilman:
Git is a weird thing...  I've been using it daily for 2 years now and still feel like I don't know what is happening.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,254 @@
...
+using System.Text;
+
+namespace Microsoft.Xna.Framework
+{
+    public static partial class MathHelper
+    {
+        /// <summary>
+        /// Class for generating random numbers consistantly when using Mono or Microsoft compilers
+        /// </summary>
+        public class Random
```
tomspilman: Before we merge we should probably have one more discussion on naming here.

`Microsoft.Xna.Framework.Random` isn't an option because it would collide with `System.Random` all the time.

`Microsoft.Xna.Framework.GameRandom` is a little long and the prefix doesn't feel all that descriptive, but allows you to use `GameRandom` everywhere easily.  Maybe there is a better `FooRandom` name to be found?

`Microsoft.Xna.Framework.MathHelpers.Random` lets you use `MathHelpers.Random` which is maybe more descriptive... random is math not game.  Howerver calling any static method on it is ugly...  `var seed = MathHelpers.Random.GenerateSeed()`.

Anyway... felt like it was worth one more discussion.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,254 @@
...
+            {
+                return (int)Remap(UInt64ToInt32(XORShift64Star()), int.MinValue, int.MaxValue, 0, int.MaxValue);
+            }
+
+            /// <summary>
+            /// Returns a non-negative random integer of type Int32 (int) that is less than the specified maximum.
+            /// </summary>
+            /// <param name="max">Maximum possible value</param>
+            /// <returns>Random integer of type Int32 (int)</returns>
+            public int Next(int max)
```
tomspilman: Maybe it is worth changing these to `NextInt`.  For example this code:
```
var v = myRandom.Next(100);
```
Is that `Next(byte)` or `Next(int)` or `Next(long)`?  Of course experienced devs know it will be `int`, but it does make it harder for developers to understand the code at a glance.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,254 @@
...
+            public ulong Seed
+            {
+                get
+                {
+                    return this._seed;
+                }
+                set
+                {
+                    this._seed = value;
+                    this._lastValue = value;
```
tomspilman: Tell me if I understand this correctly...

`_seed` is the starting seed value... where the random started.
`_lastValue` is the seed for the next generated random.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,254 @@
...
+
+            #region Private methods
+
+            private ulong XORShift64Star()
+            {
+                _lastValue ^= _lastValue >> 12;
+                _lastValue ^= _lastValue << 25;
+                _lastValue ^= _lastValue >> 27;
+                return unchecked(_lastValue * 2685821657736338717UL);
+            }
```
KonajuGames: Just use `_seed` here instead of `_lastValue`.  We don't need to keep the original seed value.  Think of `_seed` as the seed for the next random value to be generated.  If we wanted to know what the original seed value was, we would have generated it and given it to the constructor.

MonoGame.Framework/Random.cs
KonajuGames: Per my previous comment, `_lastValue` is unnecessary and should be removed.  Is there a need for the getter on the `Seed` property?  When have you ever needed to read the seed?

tomspilman: Reading the seed is very important to multiplayer games and sometimes saving game state.

If you need to roll back to a previous game state to replay a missed multiplayer input packet... you have to be able to predictably restore the randoms to the exact state then re-simulate back to the current frame.  So in this case you would store the random seed for each frame to be able to roll back later.

Some games might want to preserve the random in its current state at saving.  With being able to get the seed you can easily restore it and continue that run.  For example maybe a game that is based on generating the starting play state based on a seed itself.

KonajuGames: > Reading the seed is very important to multiplayer games and sometimes saving game state.

But you would want to be reading the seed for the next get of a random number, not the original seed when you created the random number generator.  Which is another reason to remove `_lastValue` and use `_seed` in `XORShift64Star()`.

tomspilman: > But you would want to be reading the seed for the next get of a random number

Right.  The cases I specified before all want the current state of the randomizer so it can restart from there if need be.  (oh... replays is another case where you might store the random state) 

Having the starting seed could maybe be useful in some cases, but it is easy for the developer to just store that himself.

So I agree with you.

Shqrdx:
Next instead NextInt, NextLong - This is good and do not cause ambiguity and confusing to end users ?

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,254 @@
...
+            {
+                return (short)Remap(UInt64ToInt16(XORShift64Star()), short.MinValue, short.MaxValue, min, max);
+            }
+
+            #endregion
+
+            #region Int32 Next methods
+
+            /// <summary>
+            /// Returns a non-negative random integer of type Int32 (int).
```
Shqrdx: Wait.. why non-negative. ? the integer could be negative..

WardBenjamin: This is non-negative because it follows the standards (but expanded) from the normal MS Random class according to MSDN.

MonoGame.Framework/Random.cs
WardBenjamin: Yeah, I agree that this should be changed. Doing this now.

MonoGame.Framework/Random.cs
WardBenjamin: Alright, this can be changed as well. You could argue that having the original seed able to be read is good, but it's not an issue.

MonoGame.Framework/Random.cs
tomspilman: @Shqrdx brings up a good point here.  I don't see why we should disallow:
```
var v2 = random.NextInt(-100, 100);
```
Really for any of the types... we should allow negatives.

WardBenjamin: Yeah, they are allowed in the ones with minimum specified already. Only the ones with no min. specified default to zero, mimicking the System.Random behavior.

tomspilman: Oh... duh.  Sorry my brain is fried today. :)

MonoGame.Framework/Random.cs
tomspilman: I wouldn't have a problem with having both like a `Random.Seed` (being the starting seed) and `Random.State` (being the seed for the next random generation).  @KonajuGames ?

WardBenjamin:
I have completed all of the requested changes except for the State one. Waiting for @KonajuGames's response.

WardBenjamin:
@tomspilman approval?

KonajuGames:
I can live with Seed and State.

tomspilman:
> Logically, if the state is changed, than the seed should be as well,

That is very logical to someone that read the code...  it is just nothing from the outside suggests that.  It makes me think we need something more like `StartSeed` and `CurrentSeed` but that sounds so clumsy.

@KonajuGames ?

WardBenjamin:
Yeah, I agree. I'll think about it and check out KonajuGames's response when he does respond.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,404 @@
...
+
+            /// <summary>
+            /// Returns a random single-precision floating point number that is within a specified range.
+            /// </summary>
+            /// <param name="min">Minimum possible value</param>
+            /// <param name="max">Maximum possible value</param>
+            /// <returns>Random single-precision floating point number</returns>
+            public float NextFloat(float min, float max)
+            {
+                return (float)NextDouble((double)min, (double)max);
```
Shqrdx: Float is more common than double, especially in MonoGame and it deserves their own implementation.. @tomspilman ? Remap with float overload could be more good than raw conversion from double ?

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,404 @@
...
+            #region Floating point values
+
+            private float UInt64ToFloat(ulong uint64)
+            {
+                return (float)UInt64ToDouble(uint64);
+            }
+
+            private double UInt64ToDouble(ulong uint64)
+            {
+                return Convert.ToDouble(UInt64ToInt64(uint64));
```
tomspilman: I wonder if using like a field offset trick is faster than calling the conversion:
```
StructLayout(LayoutKind.Explicit)]
struct Convert
{
  [FieldOffset(0)]
  ulong uint64;
  [FieldOffset(0)]
  double result;
};

var convert = new Convert();
convert.uint64 = uint64;
return convert.result;
```

MonoGame.Framework/Random.cs
tomspilman: I can say for sure as i'm not expert in random algorithms.  This implementation at its core generates an unsigned 64bit random... so no matter what to get a float some rescaling has to occur.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,404 @@
...
+                {
+                    return (*(long*)&uint64) + long.MinValue;
+                }
+            }
+
+            #endregion
+
+            #region Floating point values
+
+            private float UInt64ToFloat(ulong uint64)
```
tomspilman: This looks to be unused.

MonoGame.Framework/Random.cs
theZMan: Wouldn't this give a totally different random distribution. Distribution of an Int64 is very different to an even distribution of a exponent/mantissa pair.

Plus conversion from int64 to double is a single assembly language instruction IIRC this can't be worth the time to even talk about it.. .certainly not unless someone has profiled it as being a problem

MonoGame.Framework/Random.cs
WardBenjamin: UInt64ToInt64 is used inside of all of the integer Next() functions to filter the output of the xorshift to be an integer in the correct range so that it can be resized to a different integer type.

MonoGame.Framework/Random.cs
WardBenjamin: Yeah, I actually don't know what the distribution for this would be, I just added it because it needed to be there. What change would you suggest? Should I generate two numbers, one for the exponent and one for the mantissa?

MonoGame.Framework/Random.cs
Shqrdx: He means UInt64ToFloat...

MonoGame.Framework/Random.cs
theZMan: What am I missing here... NextDouble returns a number in the range 0.....1 right?

So the conversion to double is simply uint64/(double)uint64.Max

and the float version of that is (float)(uint64/(double)uint64.Max)

Maybe I need to look at the whole codebase before I comment.

WardBenjamin: I actually believe that from the testing that I remember, the conversion here directly converts the long to the same value as a decimal, so it would end up between long.MinValue and long.MaxValue while still being a value of type double.

MonoGame.Framework/Random.cs
tomspilman: Yeah.... I mean this `UInt64ToFloat` function us unused.

WardBenjamin:
Hm. While doing some testing with this, it seems that the Guid constructor returns a GUID of zero, unlike we thought.

capture

`Guid.NewGuid()` does work as expected. I will update this pull request accordingly.

WardBenjamin:
Alright. After these last few commits, the core RNG is at a state that I consider complete, though many of the previously mentioned utility functions (shuffle, etc) are not present yet. I would appreciate a review and (maybe?) a merge if the review is positive. @tomspilman

Shqrdx:
Please, add dot at the end of the sentences. Than it will look like other documented code.

WardBenjamin:
@Shqrdx yeah, I can do that right now.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,404 @@
...
+                    return this._state;
+                }
+            }
+
+            #endregion
+
+            #region Constructors
+
+            /// <summary>
+            /// Initializes a new instance of the random number generator using a new GUID as a default seed value
```
Shqrdx: Change this constructor docs to this version.

> Constructs a random number generator using a new GUID as a default seed value.

same for other constructor.

tomspilman: We don't need to call out GUID in the docs.  Please remove mention of it.

This is an internal detail we could change based on platform or other needs that doesn't need to be public.

Shqrdx:
You not done it. After first constructor code still dont have dots. And change constructor docs to this version 

> Constructs a random number generator.

WardBenjamin:
@Shqrdx could you give me an example of this? As far as I can tell all of the methods have periods at the end of the documentation in the most recent version, and they are all that are after the constructor.

ala53:
Generic suggestion:

What about adding static `NextUInt, NextFloat, etc` methods to the RNG. That way, you can simply go `MathHelper.Random.NextFloat()` or if you need determinism: `(new MathHelper.Random(290990)).NextFloat()`. Seems like it would be useful and not much more work.

WardBenjamin:
@ala53 That's another issue that I would bring up with @tomspilman. I have no issues with a private static RNG instance that these methods could be called upon transparently, but it seems to lack control. It's also not the same as the MS Random API, which could be a bit confusing.

Shqrdx:
@tomspilman I guess this PR is good and could be final jewel in the crown of the fresh MonoGame 3.5.

WardBenjamin:
It seems fine to me @Shqrdx. We'll see what Senor Spilman's response is to that.

WardBenjamin:
@tomspilman this hasn't been active for a while, are you going to merge as it is now, or do you need something else?

Shqrdx:
I asked about this and skinning PR in twitter - they keep silence...

WardBenjamin:
@Shqrdx I'll give them a short message as well. @tomspilman said he would get to it but then never did, I don't think. He's catching up on the pull requests though I think.

WardBenjamin:
Bump.

@KonajuGames any word on this? I'd like to get this taken care of soon.

WardBenjamin:
?

WardBenjamin:
It's been over two months since this pull request has been opened. What can I do to get it merged @tomspilman @KonajuGames?

cra0zy:
If I may suggest something, rebase to 1 commit

WardBenjamin:
@cra0zy Can do!

@tomspilman can we see about getting this merged?

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,386 @@
...
+using System.Text;
+
+namespace Microsoft.Xna.Framework
+{
+    public static partial class MathHelper
+    {
+        /// <summary>
+        /// Class for generating random numbers consistently when using Mono or Microsoft compilers.
+        /// </summary>
+        public class Random
```
tomspilman: @KonajuGames @dellis1972 - Do we really want this class to be nested like this?  I'm asking here one final time as we will be sort of stuck with the API once we merge.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,386 @@
...
+            #region Static methods
+
+            /// <summary>
+            /// Utility method for generating a seed for the random number generator.
+            /// </summary>
+            [CLSCompliant(false)]
+            public static ulong GenerateSeed()
+            {
+                byte[] guid = Guid.NewGuid().ToByteArray();
+                return (BitConverter.ToUInt64(guid, 0) ^ BitConverter.ToUInt64(guid, 7));
```
tomspilman: Is there a way to do this without generating garbage from the array?

MonoGame.Framework/Random.cs
dellis1972: @tomspilman @WardBenjamin quick question.. what is the original source of this Random class? or was it black box implemented? if it was copied from somewhere we need to update the header.

As for nesting... it might be useful for it NOT to be nested but I do like the 

```new MathHelper.Random();```

as it makes it clear its not a System.Random

File: MonoGame.Framework/Random.cs [hunk 12]
```diff
@@ -0,0 +1,386 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Xna.Framework
+{
+    public static partial class MathHelper
+    {
+        /// <summary>
+        /// Class for generating random numbers consistently when using Mono or Microsoft compilers.
```
tomspilman: We need a much better description of this class.

For example there is no mention of LCG or link to information about what that is.  This is important as the API assumes you know how LCG works.

Also don't mention specific compilers...  just let people know that the results are consistent across platforms when using the same inputs.

MonoGame.Framework/Random.cs
KonajuGames: I'm not in favour of nested classes in a public API.  I would rather see this in the `MonoGame.Utilities` namespace rather than nested in XNA's `MathHelper` class.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,386 @@
...
+
+            #endregion
+
+            #region Properties
+
+            /// <summary>
+            /// Original seed for the random number generator. Assigning to this will 
+            /// reset the RNG to this state as well as assigning the seed.
+            /// </summary>
+            [CLSCompliant(false)]
```
tomspilman: How important is CLSCompliant @KonajuGames ?

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,386 @@
...
+                get
+                {
+                    return this._state;
+                }
+            }
+
+            /// <summary>
+            /// Non-negative number identifying the sequence of random numbers produced by the generator.
+            /// </summary>
+            /// <devdoc>The sequence number determines the increment of the internal LCG. Only uneven,
```
tomspilman: What is `<devdoc>`?

MonoGame.Framework/Random.cs
KonajuGames: Unfortunately the `ToByteArray()` method is the only way aside from `ToString()` to get the value out of the Guid.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,386 @@
...
+
+            /// <summary>
+            /// Initializes a new instance of the random number generator using the specified seed and sequence.
+            /// </summary>
+            /// <param name="seed">Seed to be used when computing new random numbers</param>
+            /// <param name="sequence">Non-negative number identifying the sequence of random numbers.</param>
+            [CLSCompliant(false)]
+            public Random(ulong seed, long sequence)
+            {
+                this.Seed = seed;
```
tomspilman: If `sequence` should be non-negative should we be asserting or throwing on a bad input here?

MonoGame.Framework/Random.cs
WardBenjamin: @dellis1972 we don't need a header, this was black box implemented using the PCG-XSH-RS algorithm.

MonoGame.Framework/Random.cs
KonajuGames: We get compiler warnings if a public API uses unsigned types unless the `CLSCompliant` attribute is applied to it.

MonoGame.Framework/Random.cs
```diff
@@ -0,0 +1,386 @@
...
+            /// </summary>
+            /// <param name="min">Inclusive minimum value.</param>
+            /// <param name="max">Exclusive maximum value.</param>
+            /// <returns>Random integer of type Int32 (int)</returns>
+            public int Next(int min, int max)
+            {
+                // Cast to uint in case the total range is greater than int.MaxValue
+                // Force overflow on negative numbers, so they remain less than positive ones
+                uint _min = (uint)min + 0x80000000;
+                uint _max = (uint)max + 0x80000000;
```
tomspilman: What if the user passes an inverted min/max values?  Does this still produce a reasonable random?  Should we add a debug only assert to catch errors like this?

MonoGame.Framework/Random.cs
KonajuGames: The source file should have the standard MonoGame header block.  See https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Utilities/Hash.cs#L1 for an example.

hach-que: +1 for not having it as a nested class.  If it's a nested class there's no
way of pushing it into the global namespace with a "using" block AFAIK.

Test/Framework/RandomTest.cs
```diff
@@ -0,0 +1,104 @@
...
+            rng.Seed = 4568919932995229531;
+            Assert.AreEqual(0, rng.NextLong());
+
+            rng.Seed = 4568919932995229531;
+            Assert.AreEqual(-500000, rng.NextLong(-500000, 0));
+
+            rng.Seed = 4568919932995229531;
+            Assert.AreEqual(0.0, rng.NextDouble(), 0.0);
+        }
+    }
```
tomspilman: We should add a test here that compares known inputs and outputs from the original PCG-XSH-RS algorithm.  This should be at least a dozen samples or more.

This would verify that the implementation is correct and will have the same statistical results.

MonoGame.Framework/Random.cs
cra0zy: Just want to add note that for Guid, `ToByteArray()` and `ToString()` return different values.

cra0zy:
@WardBenjamin Do you plan on updating this PR anytime soon? If not, mind if I take over?

WardBenjamin:
@cra0zy go ahead! I've been extremely busy with work/school recently and haven't had time to do much programming, let alone bringing this up again. Thanks!

cra0zy:
@WardBenjamin Are the tests you added here tested against original PCG-XSH-RS algorithm? Can you by any chance give me a link to the algorithm?

cra0zy:
Never mind, found it.

cra0zy:
So the PCG-XSH-RS algorithm can be found at https://github.com/imneme/pcg-c

cra0zy:
So.. I was not able to reproduce the same numbers... maybe I am doing something wrong...

eSPiYa:
Any update about this?

tomspilman:
@harry-cpp @mrhelmut do you feel this is worth merging and improving?

mrhelmut:
I think this is a good addition, PCG algorithms are much more stable and efficient for games than the default .NET Random. And it allows to have multi-stream generators, which is cool to have dedicated generators and saving their state (without having to rely on reflection like System.Random).

I'm okay with merging it. Alternatively, I have a smaller implementation of PCG32 around that I can contribute (this is the one we're shipping our games with): https://gist.github.com/mrhelmut/259bf2ecbd0357f2911b390a136292ba

SimonDarksideJ:
Closing in favour of https://github.com/MonoGame/MonoGame/pull/8861

Issue #3798, opened 2015-04-30
SpriteBatch can't be initialized on constructor method.


consty:
Creating the SpriteBatch in the constructor method right after the GraphicsDeviceManager throws an exception.

```C#
public GameScreen()
{
	graphics = new GraphicsDeviceManager(this);
	spritebatch = new SpriteBatch(GraphicsDevice);
}
```

Does MonoGame allows this? Then a workaround must be found to correct this (e.g. update the GraphicsDeviceManager earlier so the object state is available).

Otherwise if it's not allowed then a meaningful exception message could be written so user can be informed about the situation.
(i.e. Please initialize SpriteBatch only on LoadContent method)

Thanks.

P.S. File of interest is here:
MonoGame.Framework\Graphics\SpriteBatch.cs Line 39

KonajuGames:
The SpriteBatch, like any other graphics resource, needs the GraphicsDevice
to be created. This is done in the call to base.Initialise().  This is why
all of the project templates create the default SpriteBatch in the
LoadContent() method.

Yes, it should throw a better exception. There is no intention to allow it
to work.

theZMan:
@consty is this something that worked in XNA?

consty:
@theZMan  Sorry I have not tried XNA ever, but as @KonajuGames said a better exception message could be written.

theZMan:
Thanks... I ask because the MonoGame standard for what should work is 'did it work that way in XNA'.

tomspilman:
We should at least change that from a `ArgumentException` to a `ArgumentNullException`... but there is nothing more to do here.  If someone can submit a fix for that we'll merge it and close this issue.

nanexcool:
Hope the above pull request is correct. I've tested the behavior on XNA and it indeed throws an ArgumentNullException. I copied the exact exception message given so it should work the same as XNA now.

consty:
Well, that seems to be fine. Thanks to all. :)

Shqrdx:
@tomspilman This issue seems to be fixed.. Close ?

KonajuGames:
Fixed in #3801. Closing.

PR #3801, opened 2015-05-01
Update SpriteBatch.cs

nanexcool:
Fixes error message in #3798 when trying to construct a SpriteBatch with a null GraphicsDevice. I've tested this on XNA and the behavior is correct. I've copied the exact exception message as well.


tomspilman:
Note this PR will fail to build... i'm in the middle of some work on the build server.  So just ignore that for now.

tomspilman:
Thanks for the fix @nanexcool !

PR #3802, opened 2015-05-01
NuGet Automation

tomspilman:
Some fixes to NuGet files to get the automation working.

Also took a stab at Windows UAP nuspec.

Fixes https://github.com/mono/MonoGame/issues/2843

### DIFF
--- a/NuGetPackages/MonoGame.Framework.Android.nuspec
+++ b/NuGetPackages/MonoGame.Framework.Android.nuspec
@@ -19,6 +19,7 @@
     </metadata>
     <files>
         <file src="Android\AnyCPU\Release\MonoGame.Framework.dll" target="lib\MonoAndroid\MonoGame.Framework.dll" />
+        <file src="Android\AnyCPU\Release\MonoGame.Framework.xml" target="lib\MonoAndroid\MonoGame.Framework.xml" />
         <file src="..\..\NuGetPackages\readme\MonoGame.Framework\readme.txt" target="readme.txt" />
     </files>
 </package>
--- a/NuGetPackages/MonoGame.Framework.Linux.nuspec
+++ b/NuGetPackages/MonoGame.Framework.Linux.nuspec
@@ -19,6 +19,7 @@
     </metadata>
     <files>
         <file src="Linux\AnyCPU\Release\MonoGame.Framework.dll" target="lib\net40\MonoGame.Framework.dll" />
+        <file src="Linux\AnyCPU\Release\MonoGame.Framework.xml" target="lib\net40\MonoGame.Framework.xml" />
         <file src="Linux\AnyCPU\Release\OpenTK.dll" target="lib\net40\OpenTK.dll" />
         <file src="Linux\AnyCPU\Release\OpenTK.dll.config" target="content\net40\OpenTK.dll.config" />
         <file src="Linux\AnyCPU\Release\Tao.Sdl.dll" target="lib\net40\Tao.Sdl.dll" />
--- a/NuGetPackages/MonoGame.Framework.MacOS.nuspec
+++ b/NuGetPackages/MonoGame.Framework.MacOS.nuspec
@@ -20,6 +20,7 @@
     </metadata>
     <files>
         <file src="MacOS\AnyCPU\Release\MonoGame.Framework.dll" target="lib\net40\MonoGame.Framework.dll" />
+        <file src="MacOS\AnyCPU\Release\MonoGame.Framework.xml" target="lib\net40\MonoGame.Framework.xml" />
         <file src="MacOS\AnyCPU\Release\Tao.Sdl.dll" target="lib\net40\Tao.Sdl.dll" />
         <file src="MacOS\AnyCPU\Release\OpenTK.dll" target="lib\net40\OpenTK.dll" />
         <file src="MacOS\AnyCPU\Release\MonoGame.Framework.dll" target="lib\net45\MonoGame.Framework.dll" />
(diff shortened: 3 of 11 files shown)


tomspilman:
Seems to be working for develop builds if you use this feed:

http://teamcity.monogame.net/guestAuth/app/nuget/v1/FeedService.svc/

Can others test?

tomspilman:
This works except it currently will build for every PR.  I need to restructure the build configuration to only do the NuGet packaging for merges to `develop`.

tomspilman:
@DDReaper - What would be the right "target name" for Windows 10 UAP assemblies?

DDReaper:
UAP doesn't have a specific target type at the moment, it's not scheduled to be added until NuGet V3.0

I've not managed to get a dev box with Win10 and VS2015 running yet.  Can you ask whoever built the UAP template, to see what NuGet packages they can see and use that as a target for now and we can update it later.

I'd recommend one of the following targets and test to see what works in a UAP project:
* win
* NET
* NetCore

tomspilman:
> Can you ask whoever built the UAP template, to see what NuGet packages they can see and use that as a target for now and we can update it later.

I built the template...  I poked around the packages it setup and I see lots of different ones in use.  It seems to be all over the place as I guess some of these are not tied to a particular platform as much as we are.

That said when I open a new Windows 10 UAP project from the built in template, then look at our TeamCity feed, the only one that was visible was the Windows 8 one `netcore`.  So I guess i'll use that one want make sure it shows up as "Windows 10 UAP" to be sure no one gets confused.

Make sense @DDReaper ?

tomspilman:
@DDReaper - And another question.  Maybe for the development builds we should be appending `-develop` to the version?  This will make it clear it isn't the official release and also mark it as prerelease in the NuGet UIs right?

DDReaper:
The Win 10 UAP statement makes sense.

As to the Official vs development builds.  I agree completely.

If the NuGet version has anything other than numbers, it's marked as a pre-release instead of a stable.  For example:

* 3.4  = stable
* 3.4.1-develop = prerelease

So that is a good pattern to follow

tgjones:
It's worth mentioning that every new package on NuGet needs to have a unique version number - so we need to increment the version number for every prerelease build. That's usually not what you want for continuous or nightly builds, so you can also append a version number to the append of the prerelease identifier, like this:

* 3.4.1-develop1

However, last time I checked, NuGet sorts the prerelease identifiers alphabetically, not numerically, so `develop10` will be considered a lower version than `develop2`. So you end up starting with, say, 1000, and incrementing from there:

* 3.4.1-develop1000
* 3.4.1-develop1001
* 3.4.1-develop1002

DDReaper:
I wouldn't recommend that numbering system with Nuget.
Far better to have
* 3.4.1.1-develop
* 3.4.1.2-develop
* 3.4.1.3-develop

The suffix to denote it's a prerelease shouldn't be used for numbering in my experience

tomspilman:
We use a 'major.minor.0.buildcounter' version number... so the version will always go up, even in develop builds.

So I don't think there will be a problem.

tomspilman:
@DDReaper - Another question... shouldn't we also be installing the assembly's XML files to provide reference docs for intellisense?

DDReaper:
Could do if you wish. Just add them to the files def

tgjones:
The versioning scheme I outlined is probably the most common on NuGet for prerelease packages, but of course we can do it whichever way we choose. The thing about using a build number as part of the main version is you then can't increment the minor version number until you do a stable release. Whereas continuous builds are logically connected to the *next* minor or major version number, not the current one.

tomspilman:
@tgjones - I'm not sure what problem you are trying to correct with your suggestion.  I see no issues with how we're doing things, but maybe i'm wrong.

We have a pretty simple system:

 `major.minor.0.buildcounter`

We manually control the major and minor numbers via the TeamCity server.  Currently the develop branch builds 3.5.  Once we ship 3.5 in a month or two I will increment it to 3.6 on the develop branch.  One day we'll feel like it is time to move up the major number to 4.

We currently don't use the 3rd number in the version.  I've considered using it for various things (platforms, git hash, alpha/beta, etc), but so far there hasn't been anything useful.  IMO we don't have a use for it, but since C# expects a 4 place version we keep a zero in there.

The 4th number is the build counter.  It increments with every build... this includes all PR builds including failed builds.  The only time this is reset is right after we ship a release...  when I increment the major and/or minor version I reset the build counter to zero.  

So this is why the 3.4 release has the full version of 3.4.0.456.  There were 465 builds on the build server since 3.3 was released.  And 3.3 was 3.3.0.2293 because we didn't ship a release for almost a year.  Currently we're at 3.5.0.54 ...  this is because the build counter was reset to 0 and there have been 45 builds since we shipped.

This is how I decided we should do it after tons of debate with little consensus.

tgjones:
The only problem I was trying to solve - and admittedly, it is a minor one - is that with the current system, the first stable release on NuGet for a new major or minor version number will have a build number in the hundreds.

But I'll stop here - you've already got a system that works.

tomspilman:
> the first stable release on NuGet for each new major or minor version number will have a build number in the hundreds.

Yeah... that is true.  Still to me it seems right.  A new release starts when the current release ships...  it makes sense to have larger numbers on the final release than it somehow getting reset to zero for the release.

For example look at all the versions from various releases of Windows:

  http://www.gaijin.at/en/lstwinver.php

Not a pretty sub-100 build number in the bunch.

tomspilman:
Ok... this is all setup.

I also added the link to the develop NuGet feed to the download page:

  http://www.monogame.net/downloads/

PR #3804, opened 2015-05-01
[Gtk Pipeline] Bunch of Improvements

cra0zy:
I've been reading Gtk documentation for the entire day... and I've learned a lot:
- Gdk.Color, Gtk.ColorSelection and Gtk.ColorSelectionDialog are all deprecated starting from gtk 3.4 (currently gtk is at version 3.16, every six months minor version increases by 2), the replacement for them were Gdk.Argb(existed since 3.0), Gtk.ColorChooser and Gtk.ColorChooserDialog (both exist since 3.4), and since the last 2 didn't exist in gtk-sharp I added a wrapper for them
- Fixed some annoying warnings that would appear in the Application Output about not calling the Dialog with the specified parent window
- Replaced both ColorPickerDialog and YesNoCancel dialog with Gtk native implementation of them
- Removed MainMenu.xib, useless file left from xwt
- Moved IgeMacIntegration to Gtk folder
- Added a Global file which just gives some global info about Gtk Version and running Desktop Environment


cra0zy:
I know what's wrong.. I'll fix it.

tomspilman:
Great... merging!

PR #3805, opened 2015-05-02
Added templates for Effect to Pipeline

Shqrdx:
This is basic shader templates for both OpenGL and DirectX. At the end it looks like this

image


tomspilman:
This is a great idea!

The main suggestion is that we should have just an "Effect" template and a "Sprite Effect" template.

The Effect template should have both a vertex and pixel shader and use `#if` statements to allow it to support both DirectX and OpenGL target platforms.

The Sprite Effect template should be setup with just a pixel shader that matches the correct input arguments to work with `SpriteBatch`.

Second we could add some extra helpful info in these templates.  

For example you could comment explaining how input vertex structure matches the vertex buffer format and that in this sample it expects `VertexPositionColor`.  Could also maybe add some additional input vertex structures for other stock vertex formats.

Tools/Pipeline/Templates/DXEffect.fx
```diff
@@ -0,0 +1,39 @@
...
+	float4 Position : SV_POSITION;
+	float4 Color : COLOR0;
+};
+
+VertexShaderOutput MainVS(in VertexShaderInput input)
+{
+	VertexShaderOutput output = (VertexShaderOutput)0;
+	output.Position = mul(input.Position, World);
+	output.Position = mul(output.Position, View);
+	output.Position = mul(output.Position, Projection);
```
tomspilman: We should probably use `WorldViewProjection` as a single matrix here.  We want to give people a basic example, but also we don't want to lead them down the road of bad habbits.  You can explain in the comment the performance benefit of pre-combining the world, view, and projection on the CPU and not doing the same multiplication on every vertex.

Shqrdx:
@tomspilman All issues are fixed.

Tools/Pipeline/Templates/Effect.fx
```diff
@@ -0,0 +1,53 @@
+﻿
+#define DIRECTX // comment this line for compiling to OpenGL platforms
```
tomspilman: You don't have to do that.  MGFX pre-defines some macros depending on the target platform.

https://github.com/mono/MonoGame/blob/develop/Tools/2MGFX/ShaderInfo.cs#L38

On DirectX platforms:  `HLSL`, `SM4`
On OpenGL platforms:  `GLSL`, `OPENGL`

Also `DEBUG` is defined if building the shader for debugging.

Now I will admit we probably should clean those up.  `DIRECTX` and `OPENGL` are better core macros than what we have.

Still you should be able to use `#if OPENGL` only and get the right changes in place.

Tools/Pipeline/Templates/Effect.fx
```diff
@@ -0,0 +1,53 @@
...
+
+matrix WorldViewProjection;
+
+struct VertexShaderInput
+{
+#ifdef DIRECTX
+	float4 Position : SV_POSITION;
+#else
+	float4 Position : POSITION;
+#endif
```
tomspilman: I suggest maybe doing this a little cleaner:

```
#if OPENGL
#define SV_POSITION POSITION
#define VS_SHADERMODEL vs_3_0
#define PS_SHADERMODEL ps_3_0
#else
#define VS_SHADERMODEL vs_4_0
#define PS_SHADERMODEL ps_4_0
#endif

This way you don't need extra `#ifs` all over the file.

File: Tools/Pipeline/Templates/Effect.fx [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+	return input.Color;
+}
+
+technique BasicColorDrawing
+{
+	pass P0
+	{
+#ifdef DIRECTX
+		VertexShader = compile vs_4_0 MainVS();
+		PixelShader = compile ps_4_0 MainPS();
```
tomspilman: On DirectX platforms we might want to default to the most basic shader model `vs_4_0_level_9_1` and `ps_4_0_level_9_1`.  This way the effect works on all DirectX hardware by default.

Shqrdx:
All issues are fixed.

tomspilman:
Looks good to me... but i'll let @KonajuGames chime in first.

KonajuGames:
This is a good thing, and it looks to have been done well.  If I get picky, the indentation in the setting of the macros in Effect.fx and SpriteEffect.fx differ.  The standard shaders use an include file called Macros.fxh that placed these macros into one place.  My coding brain thinks that we've duplicated them at least once, then perhaps they should go into a common include.  My other brain thinks that we won't be adding any more effect templates, so leave them as they are.

tomspilman:
>  perhaps they should go into a common include

Except our simple template system doesn't support it and really these are meant more as examples for people just starting out then anything else.

Shqrdx:
>  the indentation in the setting of the macros in Effect.fx and SpriteEffect.fx differ

Fixed

tomspilman:
@Shqrdx - I assume you've tested that both these templates work correct?  That they compile and work under both GL and DX platforms?

Shqrdx:
@tomspilman Yes, of course

Shqrdx:
So, its working...
 
image

image

KonajuGames:
Good stuff.

KonajuGames:
I'm happy to merge this.  Thanks @Shqrdx

Shqrdx:
Thanks !

PR #3809, opened 2015-05-02
[Win Pipeline] Added new reference editor dialog

cra0zy:
Mimicking #3691

WinForms Dialog looks like this:
untitled


tomspilman:
This looks great... thanks @cra0zy !

PR #3812, opened 2015-05-04
[Win Pipeline] Menu/context menu improvements

cra0zy:
Stuff done:
- Context Menu: "Open File Location" should only be visible when one item is selected
- Context Menu: "Add Menu" should only be visible if one item is selected and it's not a file
- Context Menu: "Open File" should only be visible if one item is selected and it's a file
- Context Menu: "Rename" should be visible only if one item is selected
- Menu: "Add" menu should only be enabled if one or less items are selected
- Menu: "Delete" menu should only be enabled if one or more items are selected
- Menu: "Rename" menu should only be enabled if exactly one item is selected

### DIFF
--- a/Tools/Pipeline/Windows/MainView.Designer.cs
+++ b/Tools/Pipeline/Windows/MainView.Designer.cs
@@ -65,6 +65,7 @@ namespace MonoGame.Tools.Pipeline
             this.existingItemToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.existingFolderToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.toolStripSeparator1 = new System.Windows.Forms.ToolStripSeparator();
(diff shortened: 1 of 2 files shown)


tomspilman:
I need a little more info here than just "Menu/context menu improvements".  What exactly changed?

cra0zy:
Just disabled/enabled some items in menu/context menu. eg.
Menu Delete should only be enabled if at least 1 item is selected
Menu Rename should be enabled if there is exactly 1 item selected
Context Menu open file should not be enabled if there are multiple files selected

I can list all the changes but the code in this case literally speaks for itself. If you want me to list everything do say and I'll do it.

tomspilman:
We always need details.  

I don't have the time to figure out what you are trying to do.  Someone looking at this commit a year from now doesn't want to figure out what you are trying to do.

You need your commit messages to say what you are doing and/or the PR.

cra0zy:
Updated description of the PR

tomspilman:
Thanks... merging!

PR #3815, opened 2015-05-05
Vector3 docs

Shqrdx:
Changes:
-Docs for Vector3
-Readonly added to properties fields & changed methods which used it with ref keyword.
-Some parameters renamed
-Unification changes between Vector2 and Vector3
-Transform/TransformNormal now emits exceptions as XNA, instead writting output to Debug
-Added some optimization & test


tomspilman:
Merging!

Issue #3816, opened 2015-05-05
[MGCB]Adding a reference and then saving is not possible


KakCAT:
Hi,

just add a reference to a MGCB project. Then, using File -> Save is grayed. (probably not "dirtying" the project)

cra0zy:
@tomspilman This can be closed, fixed with #3818

PR #3818, opened 2015-05-05
Fix for #3816 plus few extra

cra0zy:
Changes done:
- Win Pipeline: Fixed "Exists" visible in project properties, this parameter is used just to display missing indicator and should not be visible
- Win Pipeline: Fix for #3816
- Gtk Pipeline: Fix for not being able to save after modifying any project properties
- Gtk Pipeline: Fixed project base not automatically expanding when opening and there being only 1 item

Side note, I can open WinForms pipeline from linux, tho it looks awful :| good for fixing bugs at least.
untitled3


tomspilman:
> Side note, I can open WinForms pipeline from linux, tho it looks awful :| good for fixing bugs at least.

Yeah... we noticed that a long time ago.  We really will never support that officially as it is pretty bad.

tomspilman:
Looks fine to me... merging!

PR #3820, opened 2015-05-06
Code Coverage

tomspilman:
Looking to enable code coverage in running our unit tests.

To do that i have to first switch to running unit tests under debug builds.


tomspilman:
So goal reached as we now have code coverage when running the unit tests:

http://teamcity.monogame.net/viewLog.html?buildId=6989&buildTypeId=MonoGame_DevelopWin&tab=coverage_dotnet&guest=1

Bad news is that running the unit tests went from 6 minutes to almost 20 minutes.  This means our build takes 30 minutes now.

@KonajuGames - Thoughts?

KonajuGames:
That's a huge hit.  Is most of that time on the single Windows host?  It's not running the release unit tests as well as the debug unit tests?  I guess to parallelize more we would need another Windows host.

tomspilman:
>  It's not running the release unit tests as well as the debug unit tests?

It is only running the debug unit tests.


> Is most of that time on the single Windows host?

Yes...  but there isn't much to parallelize here.

The best we could do would be starting the Mac/Linux/Windows builds and unit tests all at once in parallel.  Then have a final step that does docs, installers, nuget, etc.  That might gain us 2 or 3 minutes at most.

I think we need to wait and re-evaluate once we have new build servers setup.  With luck that performance boost will negate a lot of this test running time.

tgjones:
Here's an idea, although this may be something you've experimented with in the past, and decided against...

If you change [this line](https://github.com/mono/MonoGame/blob/develop/Test/Framework/Visual/VisualTestGame.cs#L91):

``` csharp
GraphicsAdapter.UseReferenceDevice = true;
```

to this:

``` csharp
GraphicsAdapter.UseDriverType = GraphicsAdapter.DriverType.FastSoftware;
```

it provides a speedup when running the Direct3D tests. With this change, on my machine, the tests in MonoGame.Tests.Windows take **1 minute 3 seconds** to run, instead of **1 minute 18 seconds** with the reference device. Given that the build server takes a lot longer than that to run the tests, the potential speedup is proportionally more.

I know that, unlike the reference device, WARP isn't guaranteed to follow the Direct3D spec precisely, but... it might be good enough? At least it's still entirely implemented in software, so you won't get differences due to drivers.

tomspilman:
> Bad news is that running the unit tests went from 6 minutes to almost 20 minutes. This means our build takes 30 minutes now.

Ok on the new build server we went from 1m50s for the unit tests to 5m.  This is a bigger slowdown than I expected, but likely worth it to get the code coverage results we are getting:

http://teamcity.monogame.net/viewLog.html?buildId=8280&buildTypeId=MonoGame_TestWindows&tab=coverage_dotnet

We can see there is 95% coverage of `RasterizerState` and 0% coverage for `AlphaTestEffect`.  This seems like worth while information to track and keep an eye on.

Issue #3625, opened 2015-03-18
Replace sdl mixer 1.2/openal with...


cra0zy:
So Linux uses sdl1.2 mixer which is deprecated and it should be replaced... 
WindoiwsGL uses OpenAL that loads an entire song instead of just buffering it...
So what would be a good replacement that is cross platform(Windows / Linux / Mac)?

tomspilman:
Not sure.

Ideally we would have a solution in an existing library like OpenTK or have access to some platform specific APIs (like maybe use WMS on Windows).

The other idea would be to write our own decode and stream solution in pure C#.

KonajuGames:
There is [NVorbis](http://nvorbis.codeplex.com/), a pure managed Ogg decoder, as used by Renaud Bedard in his [OggStream](http://theinstructionlimit.com/ogg-streaming-using-opentk-and-nvorbis) class.

flibitijibibo:
I eventually moved to a DynamicSoundEffectInstance, streaming with direct Vorbisfile functions:

https://github.com/flibitijibibo/FNA/blob/master/src/Media/Xiph/Song.cs
https://github.com/flibitijibibo/Vorbisfile-CS

The thread's there since we don't update instances on an audio thread, but it's technically optional. The one bug with the non-threaded mode is that MediaPlayer.OnSongFinishedPlaying is not called, since any place I put it caused lots of fun with infinite loops.

Like with the PackedVector stuff you can redo any license headers as needed, and switch out libraries for something like NVorbis if Vorbisfile# isn't preferred.

tomspilman:
> eventually moved to a DynamicSoundEffectInstance, streaming with direct Vorbisfile functions

That is actually what I was about to suggest.

For the target platforms... Windows, Mac, Linux... we have plenty of CPU threads to throw at decoding and streaming audio.  And the work isn't very heavy in the first place.

This will force us to fix/implement `DynamicSoundEffectInstance` which we've not done so far.

I think this is a good plan.

cra0zy:
So I thought all were ok with using Vorbis-CS since no one had any objections about it... I was wrong.

Anyway let's do a poll, and type in what you think is better:
1) NVobris
2) Vobris-CS

I vote 1.

KonajuGames:
As noted above, I would prefer a managed solution such as NVorbis.

tomspilman:
> 1) NVobris

 Pros:
   - No extra libraries to manage.
   - Works on any platform we need it.
[...]

PR #3826, opened 2015-05-07
[WIP] Base implementation of NVorbis for Linux and WinGL

cra0zy:
Addressing #3625

I am not changing anything for Mac because I don't have a Mac and I really don't want to break anything.

I have tested it on Linux, and I am gonna use files that teamcity builds to test it on Windows.

OggStream.cs is modified version of http://theinstructionlimit.com/ogg-streaming-using-opentk-and-nvorbis, just added needed stuff like on finish action and position/length.

### DIFF
--- a/MonoGame.Framework/Media/Song.NVorbis.cs
+++ b/MonoGame.Framework/Media/Song.NVorbis.cs
@@ -4,19 +4,29 @@
 
 using System;
 using System.IO;
-using Microsoft.Xna.Framework.Audio;
-using Tao.Sdl;
+using MonoGame.Utilities;
+using OpenTK.Audio;
 
 namespace Microsoft.Xna.Framework.Media
 {
     public sealed partial class Song : IEquatable<Song>, IDisposable
     {
-        private IntPtr _audioData;
-        private int _volume = 128; // in SDL units from 0 to 128
+        private AudioContext acontext;
+        private OggStreamer oggstreamer;
+
+        private OggStream stream;
+        private float _volume = 1f;
 
         private void PlatformInitialize(string fileName)
         {
-            _audioData = Tao.Sdl.SdlMixer.Mix_LoadMUS(fileName);
+            acontext = new AudioContext();
+            oggstreamer = new OggStreamer();
+            
+            stream = new OggStream(fileName);
+            stream.Prepare();
+            stream.AddOnFinishedAction(OnFinishedPlaying);
+
+            _duration = stream.GetLength();
         }
         
         internal void SetEventHandler(FinishedPlayingHandler handler) { }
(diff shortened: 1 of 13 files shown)


cra0zy:
It seems that I have to either modify OpenAL implementation or implement NVorbis for soundeffect/soundeffectinstance and I choose the later one.

tomspilman:
> It seems that I have to either modify OpenAL implementation or implement NVorbis for soundeffect/soundeffectinstance 

I don't understand... why?

cra0zy:
> I don't understand... why?

Creates conflict on Windows(on Linux it surprisingly works perfect), it has to do with NVorbis also using openal to play sounds... in other words, creates alien sounds on Windows!!! literally!!!

tomspilman:
> Creates conflict on Windows, corrupts sound

Do you know exactly why?  Ideally we wouldn't need to make big changes to the `SoundEffect` system.

cra0zy:
Well I narrowed it down to initialization code of: https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs, I don't know exactly why, tho I don't know that much about OpenAL either. I'll read up on some OpenAL documentation and hopefully I'll find something.

tomspilman:
Maybe @dellis1972 can help out here too.  He is working on a project that might have some overlapping goals with what you are doing here.

MonoGame.Framework/Media/Song.NVorbis.cs
```diff
@@ -4,19 +4,29 @@
...
+using OpenTK.Audio;
 
 namespace Microsoft.Xna.Framework.Media
 {
     public sealed partial class Song : IEquatable<Song>, IDisposable
     {
-        private IntPtr _audioData;
-        private int _volume = 128; // in SDL units from 0 to 128
+        private AudioContext acontext;
+        private OggStreamer oggstreamer;
```
cra0zy: coding note 1, the above 2 variables are here only for testing. I was gonna move them to OpenTKGamePlatform since only 1 instance of them is needed for playing stuff.

MonoGame.Framework/Utilities/OggStream.cs
```diff
@@ -0,0 +1,504 @@
...
+                Reader.Dispose();
+                Reader = null;
+            }
+            Ready = false;
+        }
+    }
+
+    public class OggStreamer : IDisposable
+    {
+        Action finishedPlaying = null;
```
cra0zy: coding note 2, I was going to put List<Action> at the end so that multiple song instances would call the corresponding correct methods.

cra0zy:
> Maybe @dellis1972 can help out here too. He is working on a project that might have some overlapping goals with what you are doing here.

I added few notes in case @dellis1972 has time to look.

MonoGame.Framework/Media/Song.NVorbis.cs
tomspilman: How about `MediaPlayer` instead since it is a bit more closely related to music/audio?

cra0zy: This is so funny... I look at your comment, I look at my code, and I remember, I am creating audiocontext here, but soundcontroller creates it as well... I think I may know what I need to do :)

> How about MediaPlayer instead since it is a bit more closely related to music/audio?

Sure, any place that is instanced once is fine. I was going to put it in OpenTKPlatform only because soundcontroller was created there.

dellis1972:
Looks like the Mac platform is currently using the Song.Default  so that means loading the .xnb into a SoundEffect
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Media/Song.Default.cs#L20
which in turn uses AudioFileStream
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs#L81
So my concern with changing that other to NVobis is we would loose support for the native mac formats like m4a etc. 

That said this is perfect for my "side" proect :)

tomspilman:
>  we would loose support for the native mac formats like m4a etc. 

But for your "side project" you would have to loose all platform specific formats.  The only way it would work is if the format is fully cross-platform like Vorbis.  So that seems like it isn't a problem at all.

cra0zy:
I'm not changing anything for Mac, only for WinGL and Linux.

dellis1972:
@cra0zy I'll give this a test on Mac and see if it works :)

dellis1972:
NVorbis works perfectly on Mac :) I just built the Linux project and ran it against one of my test apps which uses an ogg for music.

tomspilman:
> NVorbis works perfectly on Mac :) 

Cool... so now all that is left is fixing this issue he had on Windows and resubmitting this PR.

cra0zy:
Found the line that's creating the problem: https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs#L197

KonajuGames:
This will help me get DynamicSoundEffectInstance implemented for OpenAL. I
have the XAudio2 version done, but OpenAL and the implementation in
MonoGame was not clear to me how to add dynamic sound into it. This is the
only thing delaying the addition of DynamicSoundEffectInstance for all
platforms.

In a branch for another project, I have used NVorbis to get Ogg Vorbis
support on Windows through DynamicSoundEffectInstance on XAudio2 (inside
MediaPlayer) because Vorbis provides seamless looping where other
compressed streaming formats do not.

tomspilman:
>  This is the only thing delaying the addition of DynamicSoundEffectInstance for all
platforms.

Would it be better if this Vorbis song stuff used `DynamicSoundEffectInstance` instead of talking to OpenAL directly?

KonajuGames:
That would certainly be ideal. I'll push up what I have done for
DynamicSoundEffectInstance on XAudio2 later today.

cra0zy:
Well as I said above, I am creating multiple audiocontext and that's what was causing the problem(allowed on unix but not on windows it seems), I'll have this PR fixed up in about 15 mins. :)

PR #3829, opened 2015-05-08
Pipeline icon from UNC file

nkast:
Move `ExtractAssociatedIcon(...)` to `Util` and try to extract icon for UNC file paths.


cra0zy:
I don't thnk adding #if code to Common folder is a good idea, it's suposed to be code that can be used by all platforms.

nkast:
You are right, that was rushed. Any suggestions on how we should handle this? Add an Interop.cs in /Windows perhaps? Or should I drop this alltoghether?  
I am trying to build from  a shared folder in a VM Win10.  Right now I just managed to get the MainWindow load but can't build the project and probably other things are broken as well. Just assigning the shared folder to a drive letter was enough and I can now move on.

nkast:
Anyway, It's impossible to debug/deploy from  a shared folder, so I'll just checkout a local copy and keep working from there.

cra0zy:
> Any suggestions on how we should handle this? Add an Interop.cs in /Windows perhaps?

That sounds ok, Gtk Pipeline also has Wrapper stuff for Mac and Linux: 
https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Gtk/Gtk3Integration.cs
https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Gtk/IgeMacIntegration.cs

PR #3833, opened 2015-05-11
Added FrameworkResources

Shqrdx:
Changes :
 -New class FrameworkResources( its internal class from XNA which contains constant strings and other resources for use in multiply places in code).
 -Added FrameworkResources.ResourceCreationWhenDeviceIsNull to all places when graphics resource is created and when graphicsDevice is null in exception message.

### DIFF
--- a/MonoGame.Framework/FrameworkResources.cs
+++ b/MonoGame.Framework/FrameworkResources.cs
@@ -1,0 +1,15 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+namespace Microsoft.Xna.Framework
+{
+    internal static class FrameworkResources
+    {
+        #region Error strings
+
+        internal const string ResourceCreationWhenDeviceIsNull = "The GraphicsDevice must not be null when creating new resources.";
+
+        #endregion
+    }
+}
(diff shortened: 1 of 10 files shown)


MonoGame.Framework/FrameworkResources.cs
```diff
@@ -0,0 +1,15 @@
...
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+namespace Microsoft.Xna.Framework
+{
+    internal static class FrameworkResources
+    {
+        #region Error strings
+
+        internal const string ResourceCreationWhenDeviceIsNull = "The GraphicsDevice must not be null when creating new resources.";
```
tomspilman: @KonajuGames - What do you think about starting to follow this technique for message strings?

KonajuGames: I've thought about it before, and is something we should do.  Thanks @Shqrdx for getting it started.

tomspilman: I guess the main benefit is removing the slight differences in related error messages and being able to localize it eventually if we wanted to.

KonajuGames: Those are the two primary reasons.

Shqrdx:
Rebase completed. Ready for merging... I hope )

tomspilman:
Ok... merging!

PR #3834, opened 2015-05-11
Better window border hangling for OpenTK

cra0zy:


MonoGame.Framework/Desktop/OpenTKGameWindow.cs
```diff
@@ -286,24 +270,11 @@ private void UpdateWindowState()
...
-                    desired = WindowBorder.Hidden;
-                else
-#if LINUX
-                    desired = WindowBorder.Resizable;
-#else
-                    desired = _isResizable ? WindowBorder.Resizable : WindowBorder.Fixed;
-#endif
-                if (desired != window.WindowBorder && window.WindowState != WindowState.Fullscreen)
-                    window.WindowBorder = desired;
+                updateborder = 2;
```
dellis1972: can we have a comment about why this needs to be 2? It would help dev's in the future know why it was done this way :) Can either be in code , or in the commit message (or both :) )

cra0zy: Sure I'll add it, one sec.

The reason why it's 2 is we need a very small delay between resizing the window and setting the borders to avoid Linux bug.

PR #3837, opened 2015-05-12
Merged WinGL and Linux OpenTKPlatform and OpenTKWindow

cra0zy:
Stuff done:
- added DESKTOPGL tag to WinGL and Linux
- setup mouse leave behavior
- added soundcontroller dispose so that the game would close
- removed few peaces of unneeded code


tomspilman:
Seems good to me... merging!

PR #3838, opened 2015-05-12
Renormalize tangent frame after transformation.

mgarstenauer:
Tangent frames (normal, tangent, binormal vectors) need to be re-normalized after transformation. This is necessary, for example, when a scale factor is applied to a model in the content pipeline.


tomspilman:
@mgarstenauer Thanks!

PR #3839, opened 2015-05-12
Added DesktopGL as content platform and marked WinGL and Linux as obsolete

cra0zy:
I've also added for the Gtk Pipeline to display [!] if the platform is obsolete.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
+++ b/MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
@@ -40,11 +40,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         Android,
 
         /// <summary>
-        /// Linux-based PCs
+        /// All desktop versions using OpenGL.
         /// (MonoGame)
         /// </summary>
-        [Obsolete("This platform is obsolete, use DesktopGL instead")]
-        Linux,
+        DesktopGL,
 
         /// <summary>
         /// Apple Mac OSX-based devices (iMac, MacBook, MacBook Air, etc)
@@ -98,12 +97,13 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         /// (MonoGame)
         /// </summary>
         [Obsolete("This platform is obsolete, use DesktopGL instead")]
-        WindowsGL,
+        WindowsGL = DesktopGL,
 
         /// <summary>
-        /// All desktop versions using OpenGL.
+        /// Linux-based PCs
         /// (MonoGame)
         /// </summary>
-        DesktopGL
+        [Obsolete("This platform is obsolete, use DesktopGL instead")]
+        Linux = DesktopGL
     }
 }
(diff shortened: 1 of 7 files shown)


MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
```diff
@@ -42,6 +43,7 @@ public enum TargetPlatform
         /// Linux-based PCs
         /// (MonoGame)
         /// </summary>
+        [Obsolete("This platform is obsolete, use DesktopGL instead")]
         Linux,
```
tomspilman: Can we do like...
```
Linux = DesktopGL,
```
It would then at least take the same path as the desktop GL code.

MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
```diff
@@ -95,6 +97,13 @@ public enum TargetPlatform
...
         /// (MonoGame)
         /// </summary>
+        [Obsolete("This platform is obsolete, use DesktopGL instead")]
         WindowsGL,
+
+        /// <summary>
+        /// All desktop versions using OpenGL.
+        /// (MonoGame)
+        /// </summary>
+        DesktopGL
```
tomspilman: @KonajuGames  @dellis1972  - Are we sure this is the name we want here?

MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
cra0zy: will do

cra0zy: not allowed in c#, throws error

tomspilman: > not allowed in c#, throws error

A compiler error or a runtime error?

cra0zy: compiler

cra0zy: It was getting caused because DesktopGL is bellow Linux, moving it upwards fixes it.

tomspilman: Ahh... yeah that would do it.  It was confused for a second because I was sure that worked.

MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
cra0zy: At first I also asked is there anything else I can put in there, but there really is no good alternative to DesktopGL. Can't put OpenGL because other platforms use it. Can't put OpenTK either for the same reason.

Shqrdx:
MacOS too needs to be obsolete ? Because it is desktop and opengl.

dellis1972:
It will be but the first task is to merge Linux and WindowsGL. we'll get to
mac next :)

cra0zy:
@dellis1972 Read @tomspilman 2nd code comment from here.

cra0zy:
@tomspilman Is the build server ok?

tomspilman:
> @tomspilman Is the build server ok?

Yea... just backed up with a lot of builds to finish.  The build currently takes like 20 minutes or so.  You get 3 or 4 builds queued up and you're waiting over an hour.

We're planning on upgrading them soon and cutting down our build times.

cra0zy:
Oh, now I see, #3840 basically broke it.

tomspilman:
Merging... thanks!

PR #3841, opened 2015-05-12
Fixes for FBX importer

mgarstenauer:
Here are a few improvements for the FBX importer (OpenAssetImporter).

- The bone weights were only collected up to indicesCount instead of vertexCount. Missing bone weights weren't handled.
- The content pipeline currently only handles indexed triangle lists. But Assimp may return points and lines for degenerate triangles. These degenerate triangles need to be removed.
- The importer now supports skinned model where the root bone has a transformation.
An example was given here: https://github.com/mono/MonoGame/issues/3672.


tomspilman:
Thanks @mgarstenauer !

PR #3842, opened 2015-05-12
[Pipeline] Set Default Effect

### LINKED ISSUE #3057: Content Pipeline is not setting the correct Effect for SkinnedModels
Effects on a model processed with the SkinnedModel pipeline example are set to BasicEffect when they should be set to SkinnedEffect

The skinned content processor has this code 
```
        /// <summary>
        /// Force all the materials to use our skinned model effect.
        /// </summary>
        [DefaultValue(MaterialProcessorDefaultEffect.SkinnedEffect)]
        public override MaterialProcessorDefaultEffect DefaultEffect
        {
            get { return MaterialProcessorDefaultEffect.SkinnedEffect; }
            set { }
        }
```
However this appears to have no effect on the XNB output. In fact I can't find any references to the `MaterialProcessorDefaultEffect` in the MonoGame codebase.

Any ideas where this is normally used to set the correct material?

nkast:
Fix #3057 as proposed by @tgjones with some more fixes for bugs that
surfaced.

-Set WeightsPerVertex = 4 for new SkinnedEffect, the default (0) throws
an error when deserialized.
-Fix Build All Textures loop.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/MaterialProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/MaterialProcessor.cs
@@ -148,11 +148,15 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
             if (input is BasicMaterialContent && DefaultEffect != MaterialProcessorDefaultEffect.BasicEffect)
             {
                 var newMaterial = CreateDefaultMaterial(DefaultEffect);
+                
                 // Preserve material properties.
+                newMaterial.Name = input.Name;
+                newMaterial.Identity = input.Identity;
                 foreach (var item in input.OpaqueData)
                     newMaterial.OpaqueData.Add(item.Key, item.Value);
                 foreach (var item in input.Textures)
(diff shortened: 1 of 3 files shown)


tomspilman:
@theZMan @mgarstenauer - This seem reasonable to you guys?

nkast:
Until now the material was always BasicEffect and execution never reach the multi-texture build loop. 
The loop throws an exception because we change the Dictionary while we access the Enumerator.

SkinnedEffect throw an error when desterilized an Effect with WeightsPerVertex=0. I set the value to 4. Probably we could set this to a lower value on the Skinned processor.
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/SkinnedEffect.cs#L328

mgarstenauer:
The rest looks fine to me. I think it is the right solution.

nkast:
@mgarstenauer .and `input.Name` I suppose, right?

mgarstenauer:
> and input.Name I suppose, right?
Right!

tomspilman:
Thanks @nkast  and @mgarstenauer !

PR #3843, opened 2015-05-12
SkinnedModelProcessor

nkast:
This is the SkinnedModel Sample with minimum changes.

After that I will submit a concrete AnimationClipWriter/AnimationClipReader.

### DIFF
--- a/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
+++ b/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
@@ -156,6 +156,7 @@
     <Compile Include="Processors\ModelMeshPartContentCollection.cs" />
     <Compile Include="Processors\ModelProcessor.cs" />
     <Compile Include="Processors\PassThroughProcessor.cs" />
+    <Compile Include="Processors\SkinnedModelProcessor.cs" />
     <Compile Include="Processors\SongContent.cs" />
     <Compile Include="Processors\SongProcessor.cs" />
     <Compile Include="Processors\SoundEffectContent.cs" />
--- a/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
+++ b/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
@@ -293,6 +293,7 @@
 
     <!-- Microsoft.Xna.Framework.Content.Pipeline.Serialization.Compiler -->
     <Compile Include="Serialization\Compiler\AlphaTestEffectWriter.cs" />
+    <Compile Include="Serialization\Compiler\AnimationClipWriter.cs" />
     <Compile Include="Serialization\Compiler\ArrayWriter.cs" />
     <Compile Include="Serialization\Compiler\BasicEffectWriter.cs" />
     <Compile Include="Serialization\Compiler\BooleanWriter.cs" />
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -322,6 +322,7 @@
     <Compile Include="Content\ContentManager.cs" />
     <Compile Include="Content\ContentReader.cs" />
     <Compile Include="Content\ContentReaders\AlphaTestEffectReader.cs" />
+    <Compile Include="Content\ContentReaders\AnimationClipReader.cs" />
     <Compile Include="Content\ContentReaders\ArrayReader.cs" />
     <Compile Include="Content\ContentReaders\BasicEffectReader.cs" />
     <Compile Include="Content\ContentReaders\BooleanReader.cs" />
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -574,7 +574,11 @@
     <Compile Include="Graphics\SpriteBatchItem.cs" />
     <Compile Include="Graphics\SpriteEffects.cs" />
     <Compile Include="Graphics\SpriteFont.cs" />
-    <Compile Include="Graphics\SpriteSortMode.cs" />
+    <Compile Include="Graphics\SpriteSortMode.cs" />    
+    <Compile Include="Graphics\SkinnedAnimation\AnimationClip.cs" />
+    <Compile Include="Graphics\SkinnedAnimation\AnimationPlayer.cs" />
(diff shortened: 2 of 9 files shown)


Shqrdx:
@tomspilman Is the SkinnedAnimation namespace should needed to be ? Instead to raw push to Microsoft.XNA.Framework.Graphics namespace ?

nkast:
@Shqrdx That's where I place the new files at the beginning, but there are a lot of classes, mostly data containers. I am not even sure if `.Graphics` is the right place, the namespace gets too long.

Shqrdx:
@nkast Ok.. I agree with you, that namespace is required if we want to extend MonoGame in good manner.

SolitudeEntertainment:
Call me a noob (cus I kinda am to GitHub), how do I get a build of this?

nkast:
@SolitudeEntertainment You can find the build at TeamCity but I suggest to work with a custom importer until this becomes part of `/develop` officially.
First, we need https://github.com/mono/MonoGame/pull/3842 to handle `SkinnedEffect` correctly in MG pipeline. For now, you have to create your own `SkinnedEffect` when you import animations with the pipeline, whether you use a custom importer or this PR.
Second, you have to flat out keyframes when you export your animations. Currently there are incompatibilities on how curved animations are handled in MG and XNA.

The best solution right now is to use an old XNB from XNA in case this is a port. If this is a new project and you can port your importer to MG pipeline you could grab #3842 and report back if there are any problems. Go to "Build 3.5.0.133"  , "Artifacts" tab, and download "Windows/MonoGameSetup.exe".

SolitudeEntertainment:
Wow thanks for the response!

Ok So I will need to get SkinnedEffect working... That has probably been the step I have been missing all along (my XNA .xnb files were not working and I had assumed they needed to be compiled in MG). You are correct that it is a port, (http://solitudegames.com), we are just trying to get this moved to monogame before we put out our Demo. I will hammer away at this. When I get it working I will add it to my fork and post a big tutorial (this is already started from the work weve done so far) on it with all the files needed... I really feel this is a big thing holding monogame back. I realize that people's games will normally need custom pipeline parts, but to have something standardized to get started with would be huge...

Do you know of a step by step for adding a Skinned effect Importer? The skinned effect seems to be in the Framework, Im guessing it just is not 100% yet?

For slightly more info, we use a Custom SkinnedEffect, but it is heavily reliant on the Stock SkinnedEffect. 

Again thank you for the response, I will get working on it!

nkast:
@SolitudeEntertainment If you load an .XNB from XNA you shouldn't have a problem. #3842 affects only the MG content pipeline.

SolitudeEntertainment:
Ok, I will run more tests on my old XNB files, and see whats up.

KonajuGames:
I had SkinnedEffect working properly in my port of the Skinned Animation Sample back in 2012, with both CPU and GPU animation working.  I haven't loaded it up recently, but it might have something to help.
http://konaju.com/downloads/CPUSkinningSample_4_0.zip

SolitudeEntertainment:
Thanks!

SolitudeEntertainment:
@KonajuGames , Ok so your sample compiles great. But my models all throw that "asset as a non-content file!" on me. My SkinnedModelProcesser is pretty heavily edited, so I went to the stock XNA example and compiled my models there and they work as expected. From there I took the .xnb files from the stock XNA example and added them to your project and I still get the same error...

More info, all .xnb files were set to build action=content, and "Copy if newer". I tried a couple of models and they all came back with the same problem. 

Any help would be much appreciated.

nkast:
@SolitudeEntertainment Is there any more info on the inner exception?

SolitudeEntertainment:
Hey nkast, I decided to throw my hands up in the air on this and write a completely custom script, which I have working. I will post a quick tutorial on this right after I get my game Demo out (a week or so). It will still require a skinnedModelProcessor, which I will also post when I get it working for Monogame (mine currently only works for XNA). 

So for my purposes I have it working, and even better than before.

tomspilman:
@KonajuGames  - You know more about this code than I do.  Leave this to you to review and merge.

standtall007:
Hi. Does the SkinnedModelProcessor Work properly yet? meaning is there any chance i can integrate it with relative easiness into my game engine? i have tried to search for a way where i can import an .fbx mesh (it is now converted to .xnb) which has no animations in it at all, and use another .fbx file which only has the bones animation(s) in it... is this possible? i am quite new to monogame but i really love every bit of it :)

nkast:
@standtall007 It is possible. 
In one project we had two fbx,  a low poly + bones + animations and a second high poly + bones.
We were able to switch between the two models and use the AnimationClip from the first model.
But the bones were 1-1 match between the two models, I am not sure how far you can push this.

We need an AnimationClip  processor at some point...
to load generic animations (var walkClip = content.Load<AnimationClip>(walkAnim) )  and apply them to models with similar skeleton.

standtall007:
@nkast Thanks for the reply :) i did also mean 1:1 scale on the bones (it's an fbx asset from Unity i used!) but it is  awesome that Monogame is capable of such Things ;-) btw does SkinnedEffect Work properly or is it still being developed on? i haven't noticed if it Works properly yet, but maybe it's just me :P

zoltanferenczfi:
@nkast Can I ask when will the SkinnedModelProcessor available in the MonoGame Pipeline Tool? I have a lot of FBX models with animations and right now I can only put them into my game without animations (selecting Model in FBX importer in the tool). Thanks for your answer!

adggithub:
Is it possible to have an example of use of a skinned model?

nkast:
This still doesn't work, obviously it needs more work to be done. I will close it so people don't get the wrong idea that this is something that is coming soon.

kosinus123:
I had the same issue with dude.fbx, I fixed it by adding a key-frame at the start of the animation
I think the exporter skips whatever is missing a key-frame.

nkast:
@kosinus123 

> I had the same issue with dude.fbx, I fixed it by adding a key-frame at the start of the animation
> I think the exporter skips whatever is missing a key-frame.

I am not sure what that means. Is it a thing with assimp or with the importer? 

Can you share your .fbx so we could do a comparison with the old model?  
Also what's the process to upgrade the .fbx, did you have to use the autodesk converter at any point?

PR #3844, opened 2015-05-12
Setup storage device for DesktopGL

cra0zy:


MonoGame.Framework/Storage/StorageContainer.cs
```diff
@@ -118,9 +119,15 @@ internal StorageContainer(StorageDevice device, string name, PlayerIndex? player
...
 			var saved = "";
-#elif LINUX || MONOMAC
+#elif MONOMAC
             // We already have a SaveData folder on Mac/Linux.
             var saved = StorageDevice.StorageRoot;
+#elif DESKTOPGL
+            string saved = "";
+            if(CurrentPlatform.OS == OS.Linux)
+                saved = StorageDevice.StorageRoot;
+            else
```
tomspilman: How about we do something like this...
```
else if (CurrentPlatform.OS == OS.Windows)
{}
else
   throw new Exception("Unexpected platform!");
```
... or something so we get a good runtime error when the platform detection fails to work.

@KonajuGames @dellis1972  ?

KonajuGames: That makes sense to me.

dellis1972: Good call

-----Original Message-----

cra0zy: Done as you've asked.

tomspilman:
Thanks... merging!

PR #3848, opened 2015-05-13
[Pipeline] Suppress unnecessary log messages

mgarstenauer:
This PR removes unnecessary/misleading "Skipping ..." messages from the build output.

Currently when an asset is referenced multiple times (e.g. a texture referenced by several models), it is built once and skipped in all subsequent build steps. The log message "Skipping TextureXyz" is printed multiple times in the build output.

This message adds unnecessary noise to the output. (In my case it adds ~2000 lines "Skipping TextureXyz" to the build output.)

The "Skipping ..." messages are also misleading: To see if an asset was built or skipped, the user has to search for the first appearance of the file in the build output. All subsequent messages are misleading.


tomspilman:
Seems simple enough...  thanks!

PR #3857, opened 2015-05-16
Vector4 docs

Shqrdx:
Fully documented Vector4 struct,
+few fixes to other things
+tests
+added missing Transform overloads

### DIFF
--- a/MonoGame.Framework/Vector3.cs
+++ b/MonoGame.Framework/Vector3.cs
@@ -886,7 +886,7 @@ namespace Microsoft.Xna.Framework
         #region Transform
 
         /// <summary>
-        /// Creates a new <see cref="Vector3"/> that contains a transformation of vector(position.X,position.Y,position.Z,1) by the specified <see cref="Matrix"/>.
+        /// Creates a new <see cref="Vector3"/> that contains a transformation of 3d-vector by the specified <see cref="Matrix"/>.
         /// </summary>
         /// <param name="position">Source <see cref="Vector3"/>.</param>
         /// <param name="matrix">The transformation <see cref="Matrix"/>.</param>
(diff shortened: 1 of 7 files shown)


nkast:
This PR brought up something strange with how `Length()` , `LengthSquared()` and static `Normalize(ref Vector4 value, out Vector4 result)` are implemented. Since zero is now `readonly`, you had to replace the `DistanceSquared(ref this, ref zero, out result)` with the slower `DistanceSquared(this, zero).` 
The use of `DistanceSquared(...)` was inefficient in the first place. Same problem with `Vector3` while `Vector2` has those inline.

Shqrdx:
@nkast Thanks, I fix it. please said me if I incorrect.

nkast:
Actually I was talking about inlining Length(),  LengthSquared() and  Normalize(ref Vector4 value, out Vector4 result). My plan was to submit the changes after this PR was merged. 
I thing the rule here is to get 'ref' methods optimized and the slower counterparts call the ref methods.
But the inline of `Normilize()` here is good, this roundtrip of `this` is too much and one can argue that a member method is not the same as the 'static' method.

Shqrdx:
@nkast Ok. I left these changes for you.
@tomspilman This is good PR or I should change it a bit ?

Shqrdx:
@tomspilman @KonajuGames Review and merge ?

Shqrdx:
@tomspilman Please react

MonoGame.Framework/Vector3.cs
```diff
@@ -392,9 +392,9 @@ public static void Distance(ref Vector3 value1, ref Vector3 value2, out float re
         /// <returns>The squared distance between two vectors.</returns>
         public static float DistanceSquared(Vector3 value1, Vector3 value2)
```
tomspilman: We need a unit test for this as there isn't one and we need to be sure this didn't break anything.

MonoGame.Framework/Vector3.cs
```diff
@@ -751,8 +751,9 @@ public void Normalize()
         /// <returns>Unit vector.</returns>
         public static Vector3 Normalize(Vector3 value)
```
tomspilman: We need a unit test for this too.

MonoGame.Framework/Vector3.cs
tomspilman: Or if @nkast plans to re-write this then I can merge as is.

Shqrdx:
So I made these tests.. it is ok ?

tomspilman:
Ok... merging!

PR #3860, opened 2015-05-18
[Gtk Pipeline] Fixed null crash when viewing project properties

cra0zy:
Because of https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/TargetPlatform.cs#L100 when WindowsGL or Linux were read they were automatically repinted to DesktopGL, and then with the old Dictionary creating code it was trying to add multiple DesktopGL keys which in turn caused a crash. This fixes it.

Also I can't check if WindowsGL and Linux are obsolete anymore because of the same reason.


KonajuGames:
Thanks @cra0zy . Merging.

cra0zy:
I just noticed I said null crash instead of duplicate keys crash in commit text... my bad.

PR #3861, opened 2015-05-18
[Gtk Pipeline] Changed Save As dialog no button to say no instead of close without saving

cra0zy:
Save as dialog can be called when the project is not saved and user clicks on build., when I modified it last time I forgot about it.


KonajuGames:
Thanks @cra0zy . Merging.

PR #3863, opened 2015-05-18
Content Reader/Writter fixes

cra0zy:


tomspilman:
So first question.  Will content built prior to the changes for DesktopGL still load correctly?

cra0zy:
> So first question. Will content built prior to the changes for DesktopGL still load correctly?

Now they will still be loadable.

tomspilman:
Thanks... merging!

cra0zy:
> This has to be 'g', no?

Yes, you are right.

PR #3869, opened 2015-05-22
Fix ReflectiveReader for properties with private setters

Aranda:
This fixes loading types that include properties like this:

    public SomeProperty { get; private set; }

### DIFF
--- a/MonoGame.Framework/Utilities/ReflectionHelpers.cs
+++ b/MonoGame.Framework/Utilities/ReflectionHelpers.cs
@@ -113,6 +113,10 @@ namespace Microsoft.Xna.Framework.Utilities
             if (getMethod == null || !getMethod.IsPublic)
                 return false;
 
+            var setMethod = GetPropertySetMethod(property);
+            if (setMethod == null || !setMethod.IsPublic)
+                return false;
+
             return true;
         }
 


KonajuGames:
Are properties with private setters permitted in XNA, or is this an extension of functionality?  You might need to edit that test to allow private setters.

Aranda:
I believe XNA ignores these properties. After all, you need to be able to read AND write a property to serialize and deserialize it. This change fixes loading some of our XNB files that were compiled from XML using normal XNA pipeline.

I'll take a look at that test. I can't seem to find a sln file to load and run the tests manually though.

KonajuGames:
The tests are in the Windows solution generated in the root of the MonoGame tree.

Aranda:
Well it looks like my assumptions were wrong. Vanilla XNA and MonoGame produce the same results when compiling, loading and saving (via IntermediateSerializer) the `GetterOnlyProperties` class and it's MonoGame test data.

I'm currently using MonoGame Windows 8.1, so perhaps it's some subtle difference in the WINRT reflective reader stuff. I'll keep looking for now but will close this PR.

nkast:
I don't know about private setters but I remember that private properties like `[DataMember] private SomeProperty { get; set; } are ok on desktop but will fail on WP7 because of strict security.

PR #3871, opened 2015-05-23
Patch to the UAP/UWP NuGet to resolve XR.xml deployment issue

DDReaper:
XR.xml files have to be deployed to a sub folder of the package name

### DIFF
--- a/NuGetPackages/MonoGame.Framework.WindowsUAP.nuspec
+++ b/NuGetPackages/MonoGame.Framework.WindowsUAP.nuspec
@@ -25,7 +25,7 @@
         <file src="WindowsUAP\AnyCPU\Release\MonoGame.Framework.dll" target="lib\netcore\MonoGame.Framework.dll" />
         <file src="WindowsUAP\AnyCPU\Release\MonoGame.Framework.xml" target="lib\netcore\MonoGame.Framework.xml" />
         <file src="WindowsUAP\AnyCPU\Release\MonoGame.Framework.pri" target="lib\netcore\MonoGame.Framework.pri" />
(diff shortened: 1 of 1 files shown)


azchohfi:
I was about to report this! Thanks!

tomspilman:
Note this isn't how things are when you build MonoGame from source for Win10.  The `MonoGame.Framework.xr.xml` is in the same folder as all the rest of the assemblies.

DDReaper:
I know. It's a NuGet thing.
Same in Win8.1/wp8.1
Also the same with the hidden Themes\generic.xr/xaml


Oh the joys of packaging and deployment

DDReaper:
Bump @tomspilman can we get this merged and the UAP NuGet package rebuilt and published as the current UAP/UWP NuGet package doesn't work.

tomspilman:
Going to merge.

> the UAP NuGet package rebuilt and published 

We're not updating packages published to the NuGet site again until 3.5 ships.  If you want something earlier you can hook into the develop feed for NuGet packages here:

http://teamcity.monogame.net/guestAuth/app/nuget/v1/FeedService.svc/

DDReaper:
Understood @tomspilman.
But I'd recommend a patch release for just the UAP package as the current one is faulty.  Or put a statement out to not use it via the blog and inform adopters to just use the installer.

There is a lot of movement on the Win10 front and a lot of people looking for win10 frameworks/solutions.  MG got a leg up by being one of the first to support it, wouldn't want this to hamper it

tomspilman:
We have 30 downloads of the WindowsUAP package on NuGet... the least downloaded packages we have in fact.

Also I know of a bunch of WindowsUAP issues that also need fixes.

I think it can wait for another couple of weeks for a better 3.5 release.

PR #3874, opened 2015-05-25
Open audio content files in read mode

### LINKED ISSUE #3873: Exception Processing ReadOnly WAV
The code here...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs#L301

Needs to be fixed to open that stream in read only mode.

KonajuGames:
Using FileStream to open files, FileAccess.Read must be specified so it can handle files that are flagged as read-only. These are found in source control systems such as Perforce, among others.

Fixes #3873


tomspilman:
Thanks @KonajuGames !

PR #3878, opened 2015-05-25
Fix and document Game.cs

faisalr:
Fix and document Game.cs


tomspilman:
If you want to do this I suggest take things in smaller steps.  It is much easier for us to verify and accept.

PR #3887, opened 2015-05-26
Make Linux launcher integrate with application

cra0zy:
Do not attempt to understand this...StartupNotify does not mean what it says...


KonajuGames:
I wasn't going to attempt to understand it.  It is like an alien language to me.  I'm going to assume it's all good.  Merging...

PR #3888, opened 2015-05-26
[Gtk Pipeline] Fix null crash when clicking on Cancel in SaveAs Dialog

cra0zy:


KonajuGames:
Thanks @cra0zy . Merging...

Issue #3831, opened 2015-05-10
Merge WinGL and Linux MonoGame assemblies into one


cra0zy:
WindowsGL and Linux use almost the exact same code, except for Song.cs(which will for both be replaced by NVorvis) and few #ifs in OpenTKPlatform.cs.

Advantage would be that this should allow one compilation for a game that can be run from both Windows and Linux.

WardBenjamin:
I personally would be all for this because it would mean I would only need to compile once. The issue that I see with it, however, is that any code written in the future would have to be written around this rather than being platform-dependant, as some of it is at the moment. I know we are working towards less of that, but we are continually introducing more and more of it as we add new features, but before the cross-platform options are found.

tomspilman:
>  The issue that I see with it, however, is that any code written in the future would have to be written around this rather than being platform-dependent

I think that is and acceptable limitation.  Plus there is still the option to execute different code at runtime based on the detected platform we're on.

WardBenjamin:
Yeah, that's true and I was thinking of pointing that out. We just would have to avoid the compile-time dependency differences.

tomspilman:
> We just would have to avoid the compile-time dependency differences.

You also have `DllImport` which lets you late link to some native APIs after platform detection.

WardBenjamin:
Fair enough, I wasn't arguing in the first place. :)

tomspilman:
> I wasn't arguing in the first place. :)

And neither am I.  Just reminding people of the flexibility we do have.

Shqrdx:
I like it, but all #define LINUX should be removed for that ?

tomspilman:
I expect all `#if LINUX` and `#if MONOMAC` stuff would be removed.  It would be replaced by `#if OPENGL` or `#if DESKTOPGL` or something like that I expect.

cra0zy:
So here is the list of stuff that needs to be done, in case someone wants to help, all of these tasks should be in separate PRs:
- get to the same code base for Monogame.Framework.Net
- fix OpenTK bug that's not allowing window to be resized when border is fixed on Linux
- test and make sure that mouse leave behavior is same on Linux and WinGL
- replace all #if LINUX || (WINDOWS && OPENGL) with a new tag, I think OPENTK would look nicely

cra0zy:
> fix OpenTK bug that's not allowing window to be resized when border is fixed on Linux

I am taking this task.

cra0zy:
[...]

PR #3891, opened 2015-05-28
DesktopGL

cra0zy:
Stuff done(for now):
 - Made DesktopGL solution and project
 - Removed WindowsGL and Linux projects

Linux solution is still generated, it contains content pipeline stuff.


Build/Projects/MonoGame.Framework.Net.definition
```diff
@@ -21,6 +21,7 @@
 	<Platform Name="WindowsPhone">C6DA906E-9E4A-428B-8870-07A39B172F83</Platform>
     <Platform Name="WindowsPhone81">423e2b58-4fbb-4a5d-8a86-1188f2889523</Platform>
   </ProjectGuids>
+	<Platform Name="DesktopGL">1B738549-0449-4504-B978-7BFC347E3F32</Platform>
```
tomspilman: Shouldn't this be within the `<ProjectGuids>` tag?

default.build
```diff
@@ -36,11 +36,11 @@
     <call target="build_windowsphone81" />
     <call target="build_windows10" />
     <call target="build_psm" />
-    <call target="build_web" />
```
tomspilman: What was failing about this?  This builds currently on the build server, so any error here is with your changes in this PR or a local problem on your machine.

Build/Projects/MonoGame.Framework.Net.definition
cra0zy: Yes, yes it should

tomspilman:
So IMO... we can't merge this until WindowsGL and Linux are removed as platforms.  The whole point of this being to reduce platforms and not increase them.

Also I would suggest we wait until Mac is merged with DesktopGL, but I worry this will take months because of the OpenTK issues and lack of time.

cra0zy:
> So IMO... we can't merge this until WindowsGL and Linux are removed as platforms. The whole point of this being to reduce platforms and not increase them.

I was going to remove WindowsGL and Linux in next PR so that build server artifacts would be alright... 

> Also I would suggest we wait until Mac is merged with DesktopGL, but I worry this will take months because of the OpenTK issues and lack of time.

Funny thing about Mac.... the current DesktopGL binaries should already work on it... but I don't have a Mac and can't test it.

tomspilman:
> I was going to remove WindowsGL and Linux in next PR so that build server artifacts would be alright.

Ok...  we can live with that temporarily I think.


> Funny thing about Mac.... the current DesktopGL binaries should already work on it... but I don't have a Mac and can't test it.

@dellis1972 - This true?

default.build
cra0zy: JSIL fails to compile, build server already has it compiled so it never does it again.

dellis1972:
I can test the DesktopGL on mac, as mentioned in #3831 I'd like to give it a test first before we drop the mac runtime completely.

dellis1972:
a quick test it worked. I do get the following crash on closing the app though.

```[ERROR] FATAL UNHANDLED EXCEPTION: System.NullReferenceException: Object reference not set to an instance of an object
  at Microsoft.Xna.Framework.Audio.OpenALSoundController.Dispose (Boolean disposing) [0x00019] in <filename unknown>:0 
  at Microsoft.Xna.Framework.Audio.OpenALSoundController.Dispose () [0x00000] in <filename unknown>:0 
  at Microsoft.Xna.Framework.Audio.OpenALSoundController.DestroyInstance () [0x00007] in <filename unknown>:0 
  at Microsoft.Xna.Framework.Audio.SoundEffect.PlatformShutdown () [0x00000] in <filename unknown>:0 
  at Microsoft.Xna.Framework.Game.Dispose (Boolean disposing) [0x000e6] in <filename unknown>:0 
  at Microsoft.Xna.Framework.Game.Dispose () [0x00000] in <filename unknown>:0 
  at MGDesktopGL.MainClass.Main (System.String[] args) [0x0001b] in /Users/dean/Projects/MGDesktopGL/Program.cs:30 ```

**Comment by dellis1972** [comment]
still need to test graphics loading, sound and music (and shaders)

**Comment by dellis1972** [comment]
On Mac.
Textures, Sound Effect and Music works
GamePad doesn't :'(

**Comment by dellis1972** [comment]
ok got the gamepad working... turns out I needed some "unofficial" drivers for osx :P

**Comment by cra0zy** [comment]
There are issues on gamepad not working on os x at https://github.com/mono/MonoGame/issues/3790 and https://github.com/opentk/opentk/issues/257 so congrats on actually making it work!!!

I'll look into soundcontroller error once I'm done with merging.

**Comment by cra0zy** [comment]
So... everything but XS/MonoDevelop templates has been updated, and I can't update those because it uses older nuget versions for everything, not the nuget provided in the PR.

We should probably make build script download nuget packages from develop builds in: https://github.com/mono/MonoGame/blob/develop/IDE/MonoDevelop/default.build#L38

**Comment by cra0zy** [comment]
I'm done, do tell if you want me to change anything.

**Comment by dellis1972** [comment]
@cra0zy in order to use the built templates we'll need to re-organise the build process. Currently the Nugets are build last, the addin is build as part of the installer step on the mac.. We'll need to split that up a bit in order to get it to download the nugets from the build system. 

I can fix up the templates

File: Build/Projects/MGCB.definition [hunk 3]
```diff
@@ -10,6 +10,8 @@
     <Reference Include="System.Xml" />
     <Reference Include="MonoGame.Framework" />
     <Reference Include="MonoGame.Framework.Content.Pipeline" />
+
+    <Reference Include="MGCB.References" />
```
tomspilman: @hach-que - Is there a better way to do this in Protobuild?  We are trying to reference the new DesktopGL version of MonoGame.Framework from the Linux build of the MGCB tool.  Since there is no more Linux build of MonoGame.Framework is there some way to tell Protobuild to always use the DesktopGL version?

hach-que: You definitely don't want to remove Linux as a platform in Protobuild; this will break anyone depending on MonoGame.Framework through Protobuild (including me).

Instead you want to leave the standard platforms (Linux, Windows, etc.) and tell them to enable the relevant services instead.

Basically, you want to have a standard set of platforms (which represent the active different environments, even if the code itself does not have to be different), and use services specific to your project (MonoGame) to select common or different sections of code based on what is needed.

tomspilman: > Instead you want to leave the standard platforms (Linux, Windows, etc.) and
> tell them to enable the relevant services instead.

The goal here is to remove the Mac, WindowsGL, and Linux platforms and have one DesktopGL platform.  This means one set of projects and one set of final binaries that work on all three platforms.

What we don't want is to have multiple projects and multiple binaries that all have 100% the same code.

This seems like a reasonable path that should be allowed when the platforms are so similar.  In some ways Android and OUYA could be done like this in the future too.

hach-que:
I'm glad I got CC'd into this.  Basically, you don't want to remove the standard platform names that Protobuild provides; anything including MonoGame as a module will break if this is merged (and projects using Protobuild won't be able to use MonoGame on Linux at all).

The platforms in Protobuild should be considered somewhat standard; their purpose is to represent the different variations in environments and devices that are relevant for running code.  Since Linux and Windows are different operating systems and do provide different APIs, at a Protobuild level they should still be separate, even if building Windows and Linux result in the same assembly produced.  (In this sense, ever having a WindowsGL platform was a mistake, but that happened before services were implemented).

You'll want to designate OpenGL / DirectX / etc. support using services instead.  I did some initial work on this which got merged, but it's not yet at a point where you can do `Protobuild.exe --generate Windows --enable MonoGame.Framework/OpenGLGraphics`, which should be the ultimate goal.

(This would mean that projects consuming MonoGame can safely put `<Uses Name="MonoGame.Framework/OpenGLGraphics" />` and always have their games built with OpenGL, even if they just do `Protobuild.exe --generate Windows`)

Build/Projects/MGCB.definition
hach-que: Refer to my other comment, but basically by removing platforms you prevent the games consuming MonoGame from having platform specific code because they are forced to use DesktopGL as their platform.

tomspilman:
> Basically, you don't want to remove the standard platform names that Protobuild provides

We want one set of projects and one set of binaries which work across these 3 platforms.  We've done all the work on the code side of things...  the only hurdle remaining is Protobuild at the moment.

To me what we want to do is conceptually simple.  We want to tell Protobuild to reference "DesktopGL" wherever "Linux" or "Mac" was referenced before.  This seems like a simple alias or redirect of some sort.

hach-que:
Here's a diagram of why you don't want to remove platforms, you want to consolidate them with services instead:

monogame protobuild

Basically for shipping MonoGame binaries for you'll just want to build with `Protobuild.exe --generate Windows --enable MonoGame.Framework/OpenGLGraphics` and then distribute those binaries as DesktopGL.

> This seems like a simple alias or redirect of some sort.

This is effectively what services do; they separate the internal, project-specific structure (the fact that 3 platforms share identical code) from the platforms themselves.  Basically everywhere in this commit that does `<Plaforms>...,DesktopGL</Platforms>` you should do `<Services>OpenGLGraphics</Services>` instead.  

If you really want to have DesktopGL as an actual, additional platform, then I'd recommend setting it up the way WindowsGL is done (but it's really unnecessary if you're using services).

tomspilman:
> Here's a diagram

I don't see why we could not support redirecting Linux references to DesktopGL:

Game Platform | MonoGame Platform
------------ | -------------
DesktopGL | DesktopGL
Mac | DesktopGL
Linux | DesktopGL

If Protobuild allowed a Linux game project to reference the DesktopGL version of MonoGame there would be no problems.  This is something you can do right now without Protobuild. so I don't see why Protobuild could not support this.

This is also something that will come up with Windows 10 Universal apps.  It is one set of projects which covers 4 or 5 platforms depending on how you look at it.


> This is effectively what services do; they separate the internal, project-specific structure 

We understand what services are... it isn't that we're looking for.

We literally want one .csproj and one .dll for desktop GL platforms and we want that reflected in our project structure.

Having 3 platforms that all generate the same code and final binary seems silly to me.

hach-que:
> I don't see why we could not support redirecting Linux references to DesktopGL:

We could support it, but it would effectively be duplicating what services do in a less flexible way, and not get us any closer to allowing game projects to express a dependency on OpenGL.  The developer would still need to manually do `Protobuild.exe --generate DesktopGL` to get OpenGL on Windows, rather than being able to state in their project definition that they want the OpenGL graphics service.

> We literally want one .csproj and one .dll for desktop GL platforms and we want that reflected in our project structure.

Yup, just remove the Linux and Mac projects from the default generations and from all `<Platforms>` tags, add DesktopGL, make it depend on `OpenGLGraphics` service and you'll get one .csproj and one .dll to ship for DesktopGL.  Using services doesn't require you to still have any Mac or Linux platform entries in your project definitions anywhere (apart from the `<Dependencies>` section in `MonoGame.Framework.definition`, but this is the only location).

I'll send through an example of what I mean shortly.

tomspilman:
> and not get us any closer to allowing game projects to express a dependency on OpenGL

I know this is your goal for Protobuild and we don't want to ruin that, but it hasn't been a goal for MonoGame.  We have always just used Protobuild just as a project generator.

> The developer would still need to manually do  Protobuild.exe --generate DesktopGL  to get OpenGL on Windows, rather than being able to state in their project definition that they want the OpenGL graphics service.

I don't see why.  We still define a service internally...  if they want to reference that I would think they could right?

hach-que:
> I don't see why. We still define a service internally... if they want to reference that I would think they could right?

Right now it doesn't work, because it's only partially been completed.  This PR is effectively doing the same work that's required to finish that work, but being done in a way that doesn't result in it actually working.

hach-que:
Also, there are a lot of files in MonoGame.Framework.definition whose inclusion / exclusion from platforms seems fairly arbitrary.  `Audio\AudioLoader.cs` for example uses OpenTK APIs and is included for Angle, MacOS, DesktopGL and Android, while `..\ThirdParty\Dependencies\OpenTK.dll.config` is only copied for MacOS and Linux; shouldn't the latter include Angle as well?

cra0zy:
> ..\ThirdParty\Dependencies\OpenTK.dll.config

Thats just a config file, on Windows it doesn't do anything(since the default names are already set in the dll), on Linux and Mac the default names are set to Windows dll names so it just references some native libs that replace those dll files. Therefor enabled on DesktopGL.

@hach-que In the diagram you said games cannot include Mac/Linux specific code. That's not true, they could still create multiple DesktopGL projects for each platform, if they wanted, very simple solution.

@tomspilman @hach-que Currently I let Linux solution still be there, but my idea is to move Linux, Mac and Windows content tools into their own solutions. This would make sense since neither Mac nor Linux solutions will remain. Also for the MonoGame version I was going to check if we could use PCL version of MonoGame.

hach-que:
> @hach-que In the diagram you said games cannot include Mac/Linux specific code, but they could still create multiple DesktopGL projects for each platform, if they wanted, very simple solution.

That completely defeats the purpose of using Protobuild.  It's purpose is that you don't have to duplicate project configuration for different platforms.

cra0zy:
My idea still keeps one project configuration:
 - Geberate and build DesktopGL
 - Generate Windows and Linux solutions that both point to DesktopGL dll file

I don't see how that's hard to do. If you were planing to connect both solutions to DesktopGL project, than I see a problem, but that's something that Protobuild would need to fix.

hach-que:
> Generate Windows and Linux solutions that both point to DesktopGL dll file

Protobuild supports using libraries in source format, not just binary format.  This allows game developers to debug and diagnose issues in all of the libraries they are using.  Requiring game developers to use MonoGame in a binary format would be a serious regression.

> If you were planing to connect both solutions to DesktopGL project, than I see a problem, but that's something that Protobuild would need to fix.

Platforms are consistent across modules.  Re-wiring platforms like that is not supported (that is what services are for).

cra0zy:
So your idea would be to keep seperate solutions for everything... IMO that's not a good idea. 

The only advantage would be that people using Protobuild and MonoGame together who need platform specific code and keep the MonoGame solution referenced as project could do it...

Now that an if if if case scenario... keeping 1 solution for Windows Linux Mac makes much more sence from MonoGame standpoint.

hach-que:
No?  There is only one "solution" in MonoGame, which is the Protobuild project definition files.  The C# project and MSBuild solution files are generated files; they aren't permanently there and they shouldn't be considered to be anything other than temporary files on disk.

cra0zy:
@hack-que now thats just mean... you know what I meant...

hach-que:
That's not being mean, that's just stating how the project generator works.  The temporary presence of solution files on disk doesn't matter in terms of configuration.

What's important is how Protobuild is invoked, and the result of the build once it's complete.

tomspilman:
@hach-que 

I know you are trying to keep MonoGame inline with your vision for Protobuild and your Protogame projects.  I don't want to break that which is why I am glad we've gotten you into this conversation now before we merged anything.

Still you have to realize that it is pointless to trying to convince us that having multiple VS solutions and projects which all generate the exact same assemblies is a good thing.  You know as well as I do that it isn't an ideal situation.

So lets start talking about how to support the feature we need.  This will just make Protobuild a more flexible tool for all users.

dellis1972:
@cra0zy can you test ToggleFullScreen on linux (With DesktopGL).. It doesn't seem to do anything on Mac OS atm. Might be an OpenTK issue though..

cra0zy:
Works fore some time, and than it gets stuck... also with newest opentk it doesn't work at all when borders are fixed.

cra0zy:
@hach-que Is there any way to set Warning level of Project with Protobuild?

hach-que:
You can disable selected C# warnings in a project by adding the NoWarn property: https://protobuild.readthedocs.org/en/latest/project_properties.html#nowarn

Is that what you mean?

KonajuGames:
We wish to enable Warnings as Errors.​

hach-que:
We'll need to add a property in Protobuild to pass that through the C# project underneath.  If someone can find me the XML name of the property to add to the C# project, I can add support for it in Protobuild.

KonajuGames:
`<TreatWarningsAsErrors>true</TreatWarningsAsErrors>`

cra0zy:
> You can disable selected C# warnings in a project by adding the NoWarn property

NoWarn disables specific warnings, what I am looking for is to add:
```xml
<WarningLevel>0</WarningLevel>
```
This would disable all warnings. It would be useful to add this to Lidgren network project, since it's a fork and we don't care about it's warnings.

hach-que:
I've added support for both those properties in the latest version of Protobuild (just download `Protobuild.exe` from the repository).

cra0zy:
Sweet!!!

cra0zy:
@tomspilman @hach-que How about we add DesktopGL as a service, make everything DesktopGL generates be connected to that service instead of the DesktopGL platform, and leave WindowsGL/Linux/Mac as disabled platforms that are only connected to DesaktopGL service? We would have 1 platform(DesktopGL) for our changes, and there would still be a way to getnerate assemblies per platform from Protobuild.

tomspilman:
> and leave WindowsGL/Linux/Mac as disabled platforms that are only connected to DesaktopGL service?

Depends on what a "disabled platform" is.  If the user can do `Protobuild -generate` or `Protobuild -generate Linux` and a MonoGame.Framework.Linux.csproj pops out that would be bad IMO.

My concern is confusion.

The user sees DesktopGL, WindowsGL, Linux, and Mac as platforms.  From outward appearances it would make sense that all three are different...  but we know they are not and would be 100% the same.

I still see no need to have any projects that are 100% the same with identical resulting assemblies.

This is similar to the issues we had previously with MonoGame where Windows, Mac, and Linux had 3 copies of the same code in different folders.  Unnecessary complexity leads to confusion and mistakes.

hach-que:
> @tomspilman @hach-que How about we add DesktopGL as a service and a platform, make everything DesktopGL generates be connected to that service instead of the DesktopGL platform, and leave WindowsGL/Linux/Mac as disabled platforms that are only connected to DesaktopGL service? We would have 1 platform(DesktopGL) for our changes, and there would still be a way to generate assemblies per platform from Protobuild.

This is what my PR does. 

> Depends on what a "disabled platform" is. If the user can do Protobuild -generate or Protobuild -generate Linux and a MonoGame.Framework.Linux.csproj pops out that would be bad IMO.

Just remove them from the default list of platforms, while still allowing it in the list of supported platforms. Don't think of the generated files on disk as projects, because that's too permanent of a concept. There'd be no technical difference between allowing --generate Linux and removing the suffix on the generated project names. 

On a further note and after thinking about it technically more, the platform aliasing will practically never happen; almost the entire Protobuild code base (C# and XSLT) assumes there is one single, static platform for the entire execution of Protobuild. Supporting that would require a significant rewrite as it would affect almost all of the code, and considering services are already a superset of the desired functionality, it is unlikely to happen ever (even if someone submitted a PR to do it, I'd be extremely hesitant of the maintainance costs).

cra0zy:
> This is what my PR does.

No, similar, but not the same.

cra0zy:
New idea

We add DesktopGL as service, and we make everything DesktopGL uses be referenced by that service instead of platform, and not keep WindowsGL or Linux as platforms.@hach-que can than fork MonoGame and create a branch that implements Windows/Linux/Mac that use DesktopGL service without much trouble.

hach-que:
> No, similar, but not the same.

It's exactly the same as you originally described. 

Making someone fork the project to add "MacOS,Linux" to the SupportedPlatforms tag in ModuleInfo.xml with absolutely no other changes, just so that they can build their game on Mac or Linux is silly in my opinion.

tomspilman:
>  just so that they can build their game on Mac or Linux

This is the disconnect here.

Nothing we are proposing to do keeps you from referencing the MonoGame.Framework.DesktopGL.dll from your own Linux or Mac projects.

However if you are using Protobuild to generate your game project with MonoGame as a submodule you would have these issues.

To me this says Protobuild is missing functionality.


> is silly in my opinion.

As silly as having three projects with identical code and assemblies? ;)

hach-que:
> As silly as having three projects with identical code and assemblies? ;)

This isn't silly at all.  There are a bunch of libraries that already do this.

Right now, as it stands, if I do:

`mono Protobuild.exe --generate Windows` and then build the project

I will get a different assembly at the same location than if I do this:

`mono Protobuild.exe --generate Windows --enable DesktopGL` and then build the project

What we're talking about here is the inverse of this situation, where you have two assemblies at two different locations that have the same code.  There is no way that is less silly than have two assemblies at the same location with different code.  But this is the way things work.  The only way to make things work uniformly across multiple projects owned by multiple people using multiple libraries is to use a standard set of platforms which control the actual C# project XML structure for MSBuild, and a dependency tree of services for code and runtime functionality for different platforms.

-

> The user sees DesktopGL, WindowsGL, Linux, and Mac as platforms. From outward appearances it would make sense that all these are different... but we know they are not and would be 100% the same.

I don't think they do.  When I submit a game to itch.io, I'm not asked for a ZIP file for the "DesktopGL" platform, I'm asked for a ZIP file for the Linux platform.  I think it's pretty clear that Windows, Mac and Linux are platforms, and "DesktopGL" is not.

-

> Nothing we are proposing to do keeps you from referencing the MonoGame.Framework.DesktopGL.dll from your own Linux or Mac projects. Create a new project reference the assembly... done. Or just use our templates after installing the MonoGame SDK. This all works.

None of these users are impacted by leaving the 11 characters "MacOS,Linux" in ModuleInfo.xml.  They aren't using MonoGame through Protobuild and will never see it.

-

What you're trying to do here is ensure that there's only one project file and assembly on disk for all the possible permutations of invoking Protobuild inside MonoGame.  To that end, we could do any of the following:

We could strip the platform names and prefixes from all projects and replace them with an MD5 hash of the enabled services and configuration values.  This would truly give you a unique "code to C# project name" mapping.  But it would also be completely unfriendly to the user.

We could also generate the C# projects into `/tmp` or `C:\Temp`; this would remove the project files from the repository entirely and users would also not have know the black-box internals of how projects are constructed.

Another alternative is (in addition to the previous option), add a `--build` parameter which invokes msbuild, xbuild or mdtool for the user, and then they don't need to know how the solutions and projects are generated for MSBuild at all.

All of these solutions go ways into hiding how projects are constructed and built from the user.  But in this case the users of MonoGame are developers, and I don't think it's fair to write them off as being confused by the presence of C# project files that exist temporarily on disk, or by the presence of assembly files which are _entirely automatically referenced for them_.  Not to mention hiding the project files in these ways makes it significantly harder to debug issues when references are not correct.

tomspilman:
Thanks @cra0zy  and @hach-que .

I'll try to find some time this week to work out a solution.

cra0zy:
@hach-que For Warning Level
I built the newest protobuild and added:
```xml
<WarningLevel>0</WarningLevel>
```
to: https://github.com/mono/MonoGame/blob/develop/Build/Projects/Lidgren.Network.definition#L31
but it didn't do anything. Project still generated with Warning Level 4.

hach-que:
You don't need to build Protobuild. Just download Protobuild.exe from the
root of the repository and use that.

cra0zy:
> You don't need to build Protobuild. Just download Protobuild.exe from the
> root of the repository and use that.

Still didn't generate WarningLevel 0.

hach-que:
There's a functional test for WarningLevel at
https://github.com/hach-que/Protobuild/tree/master/Protobuild.FunctionalTests/TestData/WarningLevelProperty.
If you place the latest Protobuild executable in that folder and run it,
you should see the warning level appear in the generated C# project.

tomspilman:
@hach-que @cra0zy - Likely the change to support WarningLevel requires changes to the xslt files which we've customized here:

https://github.com/mono/MonoGame/tree/develop/Build

Since our xslt doesn't include the changes the warning level setting isn't working.

@cra0zy - You will need to export the new xslt files and merge the change for the warning level feature manually.  But be careful...  the xslt files are sensitive to even simple whitespace changes.

hach-que:
That would definitely be the reason for things not working.

cra0zy:
so... is there anything I can do more to this PR?

cra0zy:
I am gonna close this, I've setup a PR that modifies Visual Studio project templates here: #4135

PR #3893, opened 2015-05-29
Content Writer/Reader for ClipAnimation.

nkast:
A `ClipAnimation` usually contains a large collection of `Keyframes`. We can improve loading times significantly by no longer depend on `ReflectiveWriter<Keyframe>`.

The resulting XNB is smaller since there is no need to prefix each `Keyframe` record with Reader Type. We save a byte on each record.


nkast:
Moved to SkinnedModelProcessor (3843)

PR #3894, opened 2015-05-29
[Gtk Pipeline] Set default buttons for dialogs

cra0zy:
I'm so much of a perfectionist.... this is just a small tweak to highlight the default button for the dialogs.


KonajuGames:
I have no problem with that much attention to detail. Is a good thing and
improves the overall experience.

tomspilman:
Yeah...  I would call this doing things right. ;)

PR #3896, opened 2015-05-31
DesktopGL w/ services

hach-que:
I'll squash all those fix commits (I just want to see if this builds successfully).

This is an alternate version of #3891


tomspilman:
So i'm not totally following what you did here, but I will try to find some time to pull down your changes and experiment with it.

hach-que:
So what I've done is:

Anything that had a Platforms tag referring to DesktopGL got moved to a service instead.  Some files have sets of platforms that don't make any sense to me, and don't appear to have any platform-specific code in them, so I'm not sure what feature they're actually included / excluded on.  Thus there's more services than are probably needed (some of these files can probably be included regardless of feature set).

As a side effect of this, a few of the media implementations such as video playback gained services similar to how OpenGL / DirectX is split.  This means that rather than listing a bunch of platforms on each file, the files relevant for various media implementations are just `<Services>DefaultMediaPlayer</Service>`.

Ideally there wouldn't be any files in MonoGame that have a `<Platforms>` tag on them; all of these files are just using APIs where those APIs are only available on a subset of platforms.  Rather than explicitly listing the platforms on each file that might use an API, you declare that "the API is available" by defining a service, and tell the file to be included if the service is available.  In the `<Dependencies>` section, you then have `<Uses>` to enable the service on the platforms that actually allow it.

hach-que:
This hasn't been acted on in ages, so I'm closing this PR.  The branch with changes is still available at https://github.com/hach-que/MonoGame@desktopgl.

PR #3898, opened 2015-06-01
NVorbis Dispose fix

### LINKED ISSUE #3897: Windows OpenGL: NRE Crash when exiting application
There is a NRE being generated when exiting a Windows OpenGL application.

NRE happening here: https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs#L310

The lines causing the NRE are here : https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs#L96-L104

What is happening is that when _device = Alc.OpenDevice(string.Empty); has error being created it returns with a false in the catch https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs#L103 

which leaves the _oggstreamer variable null which then throws an NRE in the dispose: https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs#L310

cra0zy:
Fixes #3897

### DIFF
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -304,11 +304,13 @@ namespace Microsoft.Xna.Framework.Audio
                 if (disposing)
                 {
                     if (_bSoundAvailable)
+                    {
                         CleanUpOpenAL();
-
 #if DESKTOPGL
-                    _oggstreamer.Dispose();
+                        if(_oggstreamer != null)
+                            _oggstreamer.Dispose();
 #endif
+                    }
                 }
                 _isDisposed = true;
             }


KonajuGames:
There is no other case where _oggStreamer might be null?  Would it be safer to do the usual check for null before disposing then setting it to null?

cra0zy:
There is no other way it would be null, _bSoundAvailable is only set if the initialization process was successful(https://github.com/cra0zy/MonoGame/blob/nvdisposefix/MonoGame.Framework/Audio/OpenALSoundController.cs#L67), but I don't mind adding null check... one sec

KonajuGames:
Thanks @cra0zy . Merging.

PR #3901, opened 2015-06-01
Warning Cleanup Part 1 of ???¿?

cra0zy:
I've split up each warning fix into it's own commit for easier checking. Do say something if you want it rebased into one commit.


KonajuGames:
I think these look good, and I don't mind them being in separate commits.

Thanks @cra0zy . Merging.

PR #3905, opened 2015-06-02
Fix for #3094

Shqrdx:
Looks like MacGamePlatform misses the Present overload and metrics will not recreated. I dont have Mac so I dont know perfectly will it works or not.

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -283,8 +283,6 @@ namespace Microsoft.Xna.Framework
 
         public override void Present()
         {
-            base.Present();
-
             var device = Game.GraphicsDevice;
             if (device != null)
                 device.Present();
--- a/MonoGame.Framework/MacOS/MacGamePlatform.cs
+++ b/MonoGame.Framework/MacOS/MacGamePlatform.cs
@@ -385,7 +385,7 @@ namespace Microsoft.Xna.Framework
         {
             var device = Game.GraphicsDevice;
             if (device != null)
-                device.Present();
+                device._graphicsMetrics = new GraphicsMetrics();
         }
 
         #endregion
--- a/MonoGame.Framework/MacOS/MacGamePlatform.cs
+++ b/MonoGame.Framework/MacOS/MacGamePlatform.cs
@@ -380,6 +380,13 @@ namespace Microsoft.Xna.Framework
         {
             _gameWindow.ResetElapsedTime();
         }
+		
+		public override void Present()
+        {
+            var device = Game.GraphicsDevice;
+            if (device != null)
+                device.Present();
+        }
 
         #endregion
 


Shqrdx:
I mean this - https://github.com/mono/MonoGame/issues/3904 sorry for confusion

Shqrdx:
I am not sure if calling full Present would be good to other parts of rendering so I change it to raw recreation of metrics.

MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
```diff
@@ -283,8 +283,6 @@ public override void Log(string Message)
 
         public override void Present()
         {
-            base.Present();
```
tomspilman: Why was this removed?

MonoGame.Framework/MacOS/MacGamePlatform.cs
```diff
@@ -380,6 +380,13 @@ public override void ResetElapsedTime()
         {
             _gameWindow.ResetElapsedTime();
         }
+		
+		public override void Present()
+        {
+            var device = Game.GraphicsDevice;
+            if (device != null)
+                device._graphicsMetrics = new GraphicsMetrics();
```
tomspilman: This is pretty gross... but we plan to kill the `MacGamePlatform` in favor of `OpenTKGamePlatform` soon anyway.

MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
Shqrdx: Because it is a call of empty method. I recommend either mark Present in GamePlatform abstract or place other content from Present of OpenTKGamePlatform to GamePlatform itself.

tomspilman: Yeah... it should be made abstract and all platforms should implement it.

MonoGame.Framework/MacOS/MacGamePlatform.cs
Shqrdx: Than this issue resolve without my PR, because OpenTKGamePlatform already have Present.

Shqrdx: As I said 
>I am not sure if calling full Present would be good to other parts of rendering on Mac so I change it to raw recreation of metrics.

So its gross, yeah..

KonajuGames: Can you explain what you mean by this?  Why would it not be good?

tomspilman: I think all it means is that the Mac code didn't do it before... so changing it might break something.

Shqrdx: > I think all it means is that the Mac code didn't do it before... so changing it might break something.

Exactly

Shqrdx:
@tomspilman @KonajuGames I guess it is not good for merging ?

KonajuGames:
I'll have a closer look when I get home.

kjpou1:
Sorry guys.  Will make the time to verify this tonight.

kjpou1:
@Shqrdx @tomspilman @KonajuGames Verified that this is working.  Thanks.

KonajuGames:
Thanks @kjpou1 and @Shqrdx . Merging.

PR #3907, opened 2015-06-03
Reverting 9f6ab17

cra0zy:
Reverting https://github.com/cra0zy/MonoGame/commit/9f6ab1788c0a9c7119ea7006da1434bd35524a5b

Sometimes I forget why I do some stuff. I've setup copying of nvtt libs because even if config file specifies the nvtt lib, it can't specify the other libs(nvcore, nvimage, nvmath).


KonajuGames:
Merging.

PR #3908, opened 2015-06-03
[Gtk Pipeline] Better text strict list

cra0zy:
Talking about when creating new folder/new item.

Stuff done:
 - Made not allowed list of characters instead of the allowed one
 - Set platform specific lists
 - Made error text have less alpha in Gtk 3.8 or later (around the same alpha as https://github.com/gnome-design-team/gnome-mockups/blob/master/nautilus/nautilus-next/create-folder-wires.png)


KonajuGames:
Thanks @cra0zy . Merging.

PR #3910, opened 2015-06-03
[Gtk Pipeline] Implemented treeview drag & drop :)

cra0zy:
http://youtu.be/PzUwg7jpDEk


KonajuGames:
Looks good.  Will have to try to do the same on WinForms.

Merging.

Issue #3911, opened 2015-06-03
Unclear how to build a spritefont when referencing documentation


WardBenjamin:
Now that the Pipeline tool has been fully released, someone should add directions for compiling SpriteFonts, etc to the documentation, as it is very unclear. I tried selecting "new spritefont" then building, but it's missing what I assume to be a TTF file. Could this happen, or at a minimum instructions be posted here?

WardBenjamin:
Referencing https://github.com/mono/MonoGame/issues/1717, I found that I was correct. This should still be added to the documentation, with any updates.

cra0zy:
I don't see how the directions are unclear... it's the exact same process as it is on Windows... exact same font types... exact same everything...

cra0zy:
Also, the default font is Arial, which is a Windows font, so unless you have mscorefonts installed, or you have arial in the same folder as the sprite font, it's gonna throw up an error...

WardBenjamin:
I've revised this to the case of when referencing documentation instead of on Linux specifically.

cra0zy:
You do realize that the ttf file required is specified in the .spritefont file, just open it in gedit/notepad and change it... you may want to read up on how XNA works/ed

cra0zy:
Side note, if you want to add stuff to documentation, toy can add it here: https://github.com/mono/MonoGame/tree/develop/Documentation

WardBenjamin:
Yeah, I know how to build spritefont's, I've been using XNA and now Monogame for a while. I'll definitely do some mods to the docs though.

KonajuGames:
One of the reasons Arial was defined as the default on Windows was because it is a standard font that is always available.  A license agreement between Microsoft and Apple makes Arial (and other MS standard fonts) available on Mac OS X as well.  This brings me to the question of what is a standard font on Linux?  From some research, it appears that the mscorefonts package is quite common, so could we leave the default as Arial and document that the default requires the mscorefonts package on Linux?  This could also be added as a comment in the XML file directly above the typeface declaration.

tomspilman:
@WardBenjamin - What is the error message you get on Linux when building a spritefont with Arial?

Maybe our first step here is making that error message clear, so the user understands what is happening and where to begin fixing it.

WardBenjamin:
The only error message is that it cannot find the Arial font. Screenshot in a sec, got to boot up my other PC (I'm using my laptop right now).

cra0zy:
[...]

PR #3913, opened 2015-06-04
Added mscorefonts as dependencies

cra0zy:
#3911


KonajuGames:
I think this is a reasonable addition.  By all accounts the mscorefonts packages is already installed on many Linux systems anyway.

Merging.

PR #3914, opened 2015-06-04
[Gtk Pipeline] Fixed minimum size for Gtk3

cra0zy:
Gtk 2 already has minimum size set.


KonajuGames:
Was that intentionally left out of GTK3?  Nevertheless, thanks for the fix.  Merging.

cra0zy:
> Was that intentionally left out of GTK3?

GTK2 has an automatic minimum size setting where it chooses the minimum size based off of minimum size of all widgets, GTK3 doesn't do this. GTK2 and GTK3 work different in some cases.

If you are asking if I left this out intentionally before, than the answer is no, I didn't even notice.

If you are asking if GNOME intentionally changed this in GTK3 than the answer is yes.

KonajuGames:
The question was whether GNOME left it out.​

PR #3916, opened 2015-06-05
[Gtk Pipeline] Implemented Drag & Drop from file manager

cra0zy:
Stuff done:
 - implemented drag & drop from file manager
 - merged addfiledialog and addfolderdialo into additemdialog
 - fixed not being allowed to include files "Hello" and "hello" at the same time on Unix


KonajuGames:
We need to remember to give the WinForms version some similar love.

Thanks @cra0zy . Merging.

PR #3920, opened 2015-06-09
[Win Pipeline] Implemented Drag & Drop from file manager

### LINKED ISSUE #3157: [Pipeline tool] Drag and drop
It would be nice to be able to add files and directories using drag and drop to the content project, just like in Visual Studio with the XNA content projects. Adding lots of directories is hard work without this as you can only add files, not directories.

cra0zy:
Resolves #3157 & #3708


simoneddeland:
Would be great to see this!

KonajuGames:
Merged.  Thanks @cra0zy . We've got a bit behind on PRs.

PR #3922, opened 2015-06-10
[SkinnedModelProcessor] struct Keyframe

nkast:
Convert Keyframe to struct.
Instead of allocating a large number of Keyframes and add each one on a
List<Keyframe> we now allocate a single array of structs.
The idea is to keep a low heap complexity (number of live objects &
number of object references) for the GC.

Also this will improve loading times and required memory by a small
factor.


KonajuGames:
I remember doing those same optimizations such as changing Keyframe to struct and creating lists at the expected size back in my XNA days.

nkast:
@KonajuGames Good to hear that you approve those changes.
I think we can compress the matrix a bit here, store just the 3x4 bits with actual information.

tomspilman:
> I think we can compress the matrix a bit here, store just the 3x4 bits with actual information.

We should share some of our skinned model code.  We stored key frames as Quaternion rot, Vector3 scale, Vector 3 pos.  That is 10 floats vs. 12 for a 4x3 matrix.

I'll post some code later.

tomspilman:
We used a `Pose` struct instead of a Matrix in the Keyframe:

https://gist.github.com/tomspilman/4f1907e15004a3d14378

This type supports quick lerp/slerp and conversion to a matrix for skinning.

nkast:
@tomspilman This is great. I mention the Matrix opt because it's easy, an hour job top.
I see there is a cost to Pose.ToMatrix()  but blending/interpolating poses cost less than blending matrixes, plus you get quaternion smooth lerp, correct?

How should I proceed from here? should I add more branches/PR like I did now? Can I have some feedback on the initial PR (ex: Namespaces). 
Because this is something that can get a lot enhancements over time I propose to add a Version to AnimationClip record in the .xnb and make Keyframe internal. So we can at least warn people that they have to rebuild their assets when we change things.

I am looking for ways to support different keyframe records, for example, 
-Interpolated or flat keyframes.
-omit some channels, for example omit scale channel or have uniform scale (float) instead of  Vector3.
-float or  quantized RL/TC

nkast:
`AnimationContent` doesn't expose the original STR so I think it's best to wait until we expand the 'Microsoft.Xna.Framework.Content.Pipeline' with new info instead of creating Pose by decomposing the transform Matrix.

nkast:
Moved to SkinnedModelProcessor (3843)

PR #3923, opened 2015-06-10
Xml compile and load fails: Empty list of custom type containing an enum

Aranda:
Failing test added to 09_Collections.xml. Test fails with `ContentLoadException : Content reader could not be found for CustomEnum type.`

The issue was that when compiling the xml, the custom class gets a `ContentTypeWriter`, but it's `ReflectiveSerializer` is never created because there are no items in the list, which means the `ContentTypeWriter`s for the custom class members are not created. At load time, the reader is created but the element readers can not be found (even though they were not needed!)

[EDIT]
I've found a better way to resolve this. Now, all element and generic arg type writers get added to the `ContentTypeWriter` at the time that their parent writer is added. This is done by the new `internal virtual void OnAddedToContentWriter(ContentWriter)` method. The new virtual method was necessary because using `Initialize` was no good as the `ContentWriter` reference is needed and using `Write` is no good because it won't always be called (empty lists etc).

Also managed to remove this comment :)

                 // TODO: This is kinda messy.. seems like there could
                 // be a better way for generics and arrays to register
                 // their inner types with the typeWriterMap


tomspilman:
@tgjones - Is this similar to your issue in https://github.com/mono/MonoGame/pull/3951, but for the content compiler side of things?

tgjones:
@tomspilman No, this looks like a different issue. Prior to #3951, MonoGame's `IntermediateSerializer` supported this:

``` csharp
public class MyContentItem
{
    public List<string> Strings { get; set; }
}
```

but didn't support this:

``` csharp
public class MyList : List<string>
{

}

public class MyContentItem
{
    public MyList Strings { get; set; }
}
```

After #3951, it supports both. So it was specifically about serialization of classes that inherit from generic collection types.

This issue, on the other hand, looks like it's about the content compiler not correctly handling empty lists of custom types.

Aranda:
Yes, #3951 is definitely a different issue, and thankfully the change sets don't conflict.

I could probably have explained *this* issue more simply: `ContentWriter` needs to have all possible `ContentTypeWriter`s registered so it can include the type name for their matching `ContentTypReader`s at the start of the Xnb. Without this change, some writers (particularly `ReflectiveWriter`) don't register their child `ContentTypeWriter` until `Write` is called. However in the case of empty lists and dictionaries, `Write` is never called which leads to missing `ContentTypeReader` exceptions when loading the Xnb.

Aranda:
Also to explain the change in `ReflectiveWriter`: Previously it filtered the properties and fields in `Write`. This change filters them up-front in `Initialize` using two helpers: `IsValidField` and `IsValidProperty`. In `OnAddedToContentWriter` it calls `output.GetTypeWriter` to register all child `ContentTypeWriter` types (and recursively register *their* child writers too).

KonajuGames:
I'm happy to merge this.  It passes the tests, and covers a case reported on the forums today.  I am taking it on good faith though as I don't have anything that uses ReflectiveWriter or Reader, so I'm trusting the unit tests.

@tomspilman, you ok with this?

tomspilman:
Yeah... this is fairly safe.  If we find and issue later we'll have to fix the tests.

KonajuGames:
Ok. Merging.

PR #3929, opened 2015-06-12
Dxt non-multiple-of-four textures

jamesford42:
Being more explicit about Mo4 requirements for dxt textures.

### DIFF
--- a/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
@@ -403,7 +403,22 @@ namespace Microsoft.Xna.Framework.Graphics
             }
             if (_shared)
                 desc.OptionFlags |= SharpDX.Direct3D11.ResourceOptionFlags.Shared;
-
+            
+            if (_format == SurfaceFormat.Dxt1 ||
+                _format == SurfaceFormat.Dxt5)
+            {
+                if (width % 4 != 0 || height % 4 != 1)
+                {
+                    // Note that the sharpdx call immediately below this will in fact fail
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -403,7 +403,20 @@ internal override SharpDX.Direct3D11.Resource CreateTexture()
...
+                _format == SurfaceFormat.Dxt5)
+            {
+                if (width % 4 != 0 || height % 4 != 0)
+                {
+                    // Note that the sharpdx call immediately below this will in fact fail
+                    // in this case. However it fails with an extremely unhelpful message. 
+                    // The message we write here matches the one given by XNA.
+                    throw new System.ArgumentException("DXT compressed texture sizes must be multiples of four.");
+                }
+            }
```
tomspilman: @tgjones @KonajuGames 

Maybe this should be higher up in the common code?  I would think even if using DXT on OpenGL or Android we would still need the same validation.  Correct?

KonajuGames:
I've got some changes to the [content pipeline for all supported texture compression types](https://github.com/KonajuGames/MonoGame/tree/etc1) that already contains these tests, but doesn't have the test on loading the texture.  There's also the Srgb DXT formats as well that should be handled.

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
KonajuGames: Yes.  It is not just a SharpDX issue.

KonajuGames:
Closing.  Should be all handled by the ETC1 branch that was merged a while back.

PR #3933, opened 2015-06-14
Interpolate curved animation

nkast:
`AnimationContent` behavior closer to XNA. This makes possible to work with custom importers ported from/to XNA without changing the importer or exporting again the FBX/animation.

XNA generates 60 keyframes per frame for every bone for the full duration of the animation.
For those keyframes it seems that it samples/lerp the animation. Keyframes in the FBX that don't fall on the right intervals are dropped.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
@@ -805,115 +805,114 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
                     }
                 }
 
-                // Get all unique keyframe times. (Assuming that no two key frames
-                // have the same time, which is usually a safe assumption.)
-                var times = scaleKeys.Select(k => k.Time)
-                                     .Union(rotationKeys.Select(k => k.Time))
-                                     .Union(translationKeys.Select(k => k.Time))
-                                     .OrderBy(t => t)
-                                     .ToList();
+                int scaleIndex = 0;
+                int rotationIndex = 0;
+                int translationIndex = 0;
 
-                Debug.Assert(times.Count == times.Distinct().Count(), "Sequences combined with Union() should not have duplicates.");
-
-                int prevScaleIndex = -1;
-                int prevRotationIndex = -1;
-                int prevTranslationIndex = -1;
-                double prevScaleTime = 0.0;
-                double prevRotationTime = 0.0;
-                double prevTranslationTime = 0.0;
-                Vector3? prevScale = null;
-                Quaternion? prevRotation = null;
-                Vector3? prevTranslation = null;
-
-                foreach (var time in times)
+                // Interpolate frames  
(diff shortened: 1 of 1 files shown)


mgarstenauer:
Can you split this into two PRs? 'fix TimeSpan rounding' can be merged. But 'interpolate animation at fixed 1/60seconds' is more controversial and needs some more discussion.

nkast:
@mgarstenauer Sure, https://github.com/mono/MonoGame/pull/3934

mgarstenauer:
@nkast Thx!

tomspilman:
@mgarstenauer @nkast - Do we still feel this is a good fix?

How would we unit test this to be sure things are working correctly?

mgarstenauer:
I am strictly against resampling the animation in the model importer. In my opinion, the model importer should import the animation "as is" without changing the data. The model processor can always resample the animations at 60 fps if necessary. But the model processor can't reconstruct the original key frames.

tomspilman:
> I am strictly against resampling the animation in the model importer.

Oh... yeah I didn't notice that again here.  That is bad... import should just import and not modify the data in most all cases.

> But the model processor can't reconstruct the original key frames.

So fundamentally any resampling within the pipeline tool will be flawed unless we are like removing half the samples or something.

@nkast - Any response to this?

mgarstenauer:
> So fundamentally any resampling within the pipeline tool will be flawed unless we are like removing half the samples or something.

@tomspilman Sorry, I didn't get what you mean. Can you rephrase?

I am fine with any changes/resampling done by the model processor. But the model importer should leave the data as is. If we could add properties to a content importer, we could simply make the behavior optional and everything would be fine. Unfortunately, XNA does not allow to add properties to content importers. We can only add properties for content processors. :(

tomspilman:
> Sorry, I didn't get what you mean. Can you rephrase?

I meant that any resampling we do in the processor would be limited to the data we have available.  The data we have has already been possibly resampled during export from Max or Maya or whatever tool the model was built with.  So it is a little like decompressing and recompressing a JPG... data is lost.

However maybe this isn't such a big deal for animation data in particular.

> XNA does not allow to add properties to content importers. 

We should take this discussion to another issue, but we could totally support properties on importers if  we wanted for MonoGame.

mgarstenauer:
> However maybe this isn't such a big deal for animation data in particular.

Just for illustration, here are two examples:

- Imagine a sliding door that slides open over 1 seconds. It has a 2 keyframes.
--> Now, after import it has 60 key frames. But at runtime, I only need 2 keyframes.
- Imagine an animated character. The skeleton consists of chains of bones.
--> Any errors I introduce are cummulative. A small rotational error up the chain can have a large error at the end of the chain (e.g. "foot sliding" or hand missing a target).

nkast:
This PR was to make mg compatible with XNA proccesors. 

I agree that XNA importer does more than it should. That was the point with #4017. Advanced processors could use the new api to get the raw animation data.

PR #3934, opened 2015-06-14
Fix TimeSpan rounding in OpenAssetImporter.cs

nkast:
Current code will round the seconds, for ex. result to .017s instead of
0.016666s


tomspilman:
Thanks!

Issue #3941, opened 2015-06-16
SpriteFont.MeasureString incorrect height


nkast:
For a  bitmap Font (Texture) imported both by XNA importer  and MG/MGCB  and the following test code:
```
SpriteFont font;
string ch = " ";
font = content.Load<SpriteFont>("Pages\\Menus\\DialogFont");
System.Diagnostics.Debug.WriteLine(".LineSpacing = " + font.LineSpacing);
System.Diagnostics.Debug.WriteLine(".MeasureString("+ch+") = " + font.MeasureString(ch));
font.LineSpacing += 10;
System.Diagnostics.Debug.WriteLine(".LineSpacing += 10;");
System.Diagnostics.Debug.WriteLine(".LineSpacing = " + font.LineSpacing);
System.Diagnostics.Debug.WriteLine(".MeasureString("+ch+") = " + font.MeasureString(ch));
```

![mgspritefont](https://cloud.githubusercontent.com/assets/3018589/8189345/62f7acd0-1464-11e5-9b29-07d5c1ca8b26.png)

I get the following results:

XNA (XNA xnb)
.LineSpacing = 14
.MeasureString(" ") = {X:13 **Y:18**}
.LineSpacing += 10;
.LineSpacing = 24
.MeasureString(" ") = {X:13 Y:24}
 
MonoGame (XNA xnb)
.LineSpacing = 14
.MeasureString(" ") = {X:13 **Y:18**}
.LineSpacing += 10;
.LineSpacing = 24
.MeasureString(" ") = {X:13 Y:24}

MonoGame ( MGCB xnb)
.LineSpacing = 14
.MeasureString(" ") = {X:13 **Y:14**}
.LineSpacing += 10;
.LineSpacing = 24
.MeasureString(" ") = {X:13 Y:24}

`char ="A";`
MonoGame (XNA xnb)
.LineSpacing = 14
.MeasureString("A") = {X:15 **Y:18**}
.LineSpacing += 10;
.LineSpacing = 24
.MeasureString("A") = {X:15 Y:24}

MonoGame (MGCB xnb)
.LineSpacing = 14
.MeasureString("A") = {X:15 **Y:14**}
.LineSpacing += 10;
.LineSpacing = 24
.MeasureString("A") = {X:15 Y:24}

.
The results under XNA and MG match (Y:18) when using the same .XNB but the height is wrong (Y:14) when using the .XNB imported by MGCB.
Clearly the problem here is in the Font importer.

[...]

PR #3945, opened 2015-06-18
Fix MeasureString for texture fonts

nkast:
fix issue #3941


tomspilman:
@nkast - Could we add something into the existing `FontTextureProcessor` tests to ensure things are working in the future:

https://github.com/mono/MonoGame/blob/develop/Test/ContentPipeline/FontTextureProcessorTests.cs

@KonajuGames - Any ideas?

nkast:
I could add Assert for `.MeasureString()` and `.LineSpacing`.

nkast:
Added a few checks in FontTextureProcessorTests.cs, is that good enough for making into 3.5?

tomspilman:
Thanks @nkast ... merging!

nkast:
Thanks!

PR #3951, opened 2015-06-20
Add support in IntermediateSerializer for classes inherited from generic collection types

tgjones:
This enables serialization of, for example,

```
public class MyCollection : Collection<T> {}

public class ClassToSerialize
{
    public MyCollection Children { get; set; }
}
```

where previously we only supported direct use of generic collection types:

```
public class ClassToSerialize
{
    public List<T> Children { get; set; }
}
```


tgjones:
@tomspilman what happened to the automated tests? Do they not run on every pull request anymore?

KonajuGames:
The build system appears to have stopped for some reason.

tomspilman:
The Win10 box went down earlier this week.  I started it back up, but it started upgrading to the latest Win10 preview build...  so I left it doing that today.  I'll have it back up sometime tomorrow and all these stalled PR builds will fire off.

tomspilman:
The tests all pass.. so this is good.  Thanks @tgjones !

PR #3953, opened 2015-06-22
Added support for MSBuild.Content.Builder on MacOS

dellis1972:
In order to support the Content.Builder.targets on
MacOS we firstly needed to install the Tooling and
Assemblies in the right location. In this case
/Library/Frameworks/MonoGame.framework/vX.X.X.X

Symbolic links are required to get everything working
right. In this case we need a v3.0 link in
/Library/Frameworks/MonoGame.framework but we also need
some links in the Mono.framework as well.

We also need to add an Imports element to the csproj files.
Now the current MonoDevleop/Xamarin Studio templating system
does not support doing this directly. So a new MSBuild
processing class has been added to the addin. This class
will check if the project is a "MonoGame" project (is references
MonoGame.Framework.dll or has a MonoGamePlatform element) and
inject the required data into the csproj.

The templates have been left referencing the Nuget packages
for now.

Modified the ContentBuilder targets file to Batch its items
this is to get round issues when running under xbuild on Mac.

Updated templates to use a new MonoGameInstallDirectory property
rather than $(MSBuildProgramFiles32) so that we can use the
same paths for references.

It also adds the followiing to the projects

```
<MonoGameInstallDirectory Condition="'$(OS)' != 'Unix' ">$(MSBuildProgramFiles32)</MonoGameInstallDirectory>
<MonoGameInstallDirectory Condition="'$(OS)' == 'Unix' ">$(MSBuildExtensionsPath)</MonoGameInstallDirectory>
```

So that we can resolve the directory correctly on both Windows
Mac and Linux


tomspilman:
@dellis1972 - Were you maybe making some changes to the Mac build box?  It seems to be failing to execute the installer [default.build](https://github.com/mono/MonoGame/blob/develop/Installers/default.build) for a few days now.

dellis1972:
I've not made any changes... So not sure what is causing that issue. Unless nant needs upgrading on it

-----Original Message-----

dellis1972:
@tomspilman I just upgraded the Mac Box to NAnt 0.92 (it was running 0.91) that seems to have fixed it locally. Can we re-run this job?

tomspilman:
Thanks @dellis1972 ...  that seems to have fixed it.  I still don't understand how it got broken or a bad script got checked in without us knowing.

dellis1972:
@tomspilman looks like NAnt 0.91 didn't support the "choose" task.. which is why we got the issue. However on the mac there was a 0.92 source checkout, it probably just didn't get installed :/

dellis1972:
@tomspilman @KonajuGames any feedback on this one?

WardBenjamin:
Any updates on this? @tomspilman @KonajuGames

WardBenjamin:
This needs to be merged for #4059, as it could possibly have relevant fixes. @tomspilman?

cra0zy:
@dellis1972 You have some merge conflicts

ProjectTemplates/VisualStudio2010/Android/AndroidApplication1.csproj
```diff
@@ -21,6 +21,8 @@
     <TargetFrameworkVersion>v4.2</TargetFrameworkVersion>
     <MonoGamePlatform>Android</MonoGamePlatform>
     <MonoGameContentBuilderExe></MonoGameContentBuilderExe>
+    <MonoGameInstallDirectory Condition=" '$(OS)' != 'Unix' ">$(MSBuildProgramFiles32)</MonoGameInstallDirectory>
+    <MonoGameInstallDirectory Condition=" '$(OS)' == 'Unix' ">$(MSBuildExtensionsPath)</MonoGameInstallDirectory>
```
dellis1972: @tomspilman @KonajuGames are you guys happy with this change? The problem we are solving is that $(MSBuildProgramFiles32) does not exist on mac.. if the Windows created projects are to work on a mac too we need to point it to the $(MSBuildExtensionsPath) instead.. 

Alternatively move the Assemblies folder from Program Files\MonoGame\v3.0\Assembles to Program Files\MSBuild\MonoGame\v3.0\Assemblies then we could just use

$(MSBuildExtensionsPath)\MonoGame\v3.0\Assemblies\XXXX\MonoGame.Framework.dll

dellis1972: Or add two Hint paths to the templates on for Windows one for OS X

PR #3958, opened 2015-06-23
Add Int32Reader. Fixes Song loading on iOS

danzel:
Was getting stripped when doing Full Linking.
Probably needed to load the duration from the xnb file?


danzel:
Build bot is broken by the looks?

tomspilman:
Thanks @danzel !

PR #3965, opened 2015-06-26
Array type EffectParameter(s) now show element value(s) in debugger.

jamesford42:
Eg, the following will be seen in watch window...
```[Vector Single] PointLightPosition : {{1 0 0}, {0 1 0}}```

Previously all you would see was...
```[Vector Single] PointLightPosition : (null)```


KonajuGames:
Merging. Thanks @jamesford42.

PR #3966, opened 2015-06-27
Fix incorrect base method call in ChildCollection.SetItem

tgjones:


tomspilman:
I could see how this could create all sorts of random errors in the pipeline... good catch!

Issue #3853, opened 2015-05-15
Problems with new effect preprocessor


mgarstenauer:
I have encountered several problems compiling DirectX Effects. I assume that they are caused by the new preprocessor (https://github.com/mono/MonoGame/issues/1972). Before everything worked fine.

Here are two effects for reproducing the problems:

Effect.fx
```
float4 VS(float4 pos : SV_Position) : SV_Position
{
  return pos;
}

float4 PS() : COLOR
{
  return float4(0, 0, 0, 0);
}

technique
{
  pass
  {
    VertexShader = compile vs_4_0 VS();
    PixelShader = compile ps_4_0 PS();
  }
}
```

Effect2.fx
```
#define DEF0 0
#define DEF1 0
#define DEF2 0
#include "Effect.fx"
```

Effect.fx compiles without a problem. But the following problems appear when compiling Effect2.fx:

**Problem 1:** When Effect.fx does not end with a newline (CRLF) the following error is output:
`Effect.fx(18,2): error X3000: syntax error: unexpected token '#'` 

**Problem 2:** When Effect.fx ends with a newline (CRLF) the following error is output:
`Effect.fx(19,7): error X1500: syntax error : unexpected token '-'`

When I remove one of the #define in Effect2.fx or when I change one of the lines from `#define DEF0 0` to `#define DEF0 1`, it compiles without a problem. Note that these #define don't do anything. So this seems to be a problem purely of the preprocessor.

**Problem 3:** When compiling complex effects with many include files, the line numbers at which errrors are reported are off.

tgjones:
* I'm working on a fix for **Problem 1**.
* I haven't seen **Problem 2** - can you post a minimal failing test file(s)?
* I have seen **Problem 3** - I know that multiline comments cause incorrect line numbers in error messages. I'm not sure if that's the only issue, but it's certainly an issue.

mgarstenauer:
Thanks for the response. [Here](http://1drv.ms/1ELSw5q) is my test case. Let me know if you need something else.

(Note: None of this is time critical. I have managed to work around these issues so far. But in the long term it should probably be fixed.)

tomspilman:
Thanks for reporting this @mgarstenauer !  We felt we would get some tricky issues from the switch in preprocessors... so just working thru these bugs as we trip up on them and improving our unit tests.

tomspilman:
@tgjones - Did you finish that fix?

[...]

PR #3968, opened 2015-06-28
Preprocessor error line numbers

tgjones:
Related to #3853. This PR improves the line numbers reported in preprocessor error messages, but as mentioned in [this comment](https://github.com/mono/MonoGame/issues/3853#issuecomment-116190926) they are still not correct.

Preprocessor errors now use the correct filename, rather than always reporting errors on the root file.

### DIFF
--- a/Test/Assets/Effects/PreprocessorTest.fx
+++ b/Test/Assets/Effects/PreprocessorTest.fx
@@ -10,7 +10,9 @@
 This is a C style comment.
 */
 
-#define
+#if foo(TEST)
+
+#endif
 
 #if TEST == 0
 Foo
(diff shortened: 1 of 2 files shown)


tomspilman:
@tgjones - So does this also ensure the file has a newline at the end?  I've seen that files with code at the end without a newline fail to process.

tgjones:
> So does this also ensure the file has a newline at the end?

@tomspilman no, but I have a PR ready for that, which I'll submit after this one is resolved (otherwise I'll get conflicts between the two PRs).

tomspilman:
Thanks @tgjones !

Issue #3825, opened 2015-05-07
Suggestion : Implementing SkinnedModel to MonoGame core ?


Shqrdx:
I think its good idea - many 3d games and engines using skinned models. XNA has it but as separate example - my suggestion is to include it to the MonoGame core. With AnimationPlayer, Keyframe, SkinningModelProcessor(possible SkinnedModel) as support to already existed skinned effect(and adding Skinned Effect to templates of pipeline is not a bad idea too).

theZMan:
XNA has it as a separate example because in my experience you need to tweak your importer to get it right all the time. Its hard enough to get non skinned models to load in correctly.

If this was part of MonoGame then someone would need to spend a lot of time tweaking it for every export format out there.. I'd reccomend someone ports XNAnimation http://xnanimation.codeplex.com/ and grows that library into something MonoGame can use

Shqrdx:
They all deprecated and for kwxport I need to install a 3dsmax11 ? My ask is about something basic and extendable solution for simple skin animation.

theZMan:
Indeed - someone has to do the work and since those projects can't find people to work on them how will MonoGame?

But really the list was not a suggestion for solutions for you more of a reading list for anyone reading this issue who would like to take on the task you suggested.

For anyone trying to solve this problem now the XNA sample ports easily but good luck getting consistent models imported. 

Or if you are prepared to pay and therefor get support Digital Rune has a very solid engine that handles skinned models http://www.digitalrune.com/Products/GameEngine/Graphics.aspx

Shqrdx:
Why I need to pay when I can simply use Unity or Unreal engines. They are free and much more powerful than MonoGame and Digital rune "engine". But they are expensive in term of performance and size(and harder to learn) - That why MonoGame needs a free and embedded solution for that problem. @tomspilman ? do you agreed with me ?

tomspilman:
I wouldn't have a problem with a stock `SkinnedModel` type in MonoGame to complement the `SkinnedEffect`.  The issue is still someone has to contribute it and maintain it.

KonajuGames:
It's similar to the `Model` class that has always been in XNA.  It's a simple implementation that does the basic job.  If you want anything more complex and you need to implement your own model class.  I think the standard `SkinnedModel` would work well in MonoGame under the same pretense.

tomspilman:
> I think the standard `SkinnedModel` would work well in MonoGame under the same pretense.

[...]

PR #3970, opened 2015-06-29
[ContentPipeline] Ignore Identity bone

nkast:
Animations usually have a ROOT/PIVOT bone. XNA remove this kind of bone. 
As can be seen from my tests in #3825 the same animation had 22 Bones when importing XNA animations and 23 Bones when importing MG animations.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
@@ -805,6 +805,18 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
                     }
                 }
 
+                // Ignore identity bone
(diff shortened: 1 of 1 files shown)


KonajuGames:
Are you able to add a test for this?

tomspilman:
@mgarstenauer - Any thoughts on this?

Also what would be a good unit test for this change?

mgarstenauer:
In my opinion the model importer should import the model "as is" without changing the data. If assimp says that the root bone is keyframed, then this might be relevant information. The data can always be cleaned up in the model processor.

tomspilman:
> In my opinion the model importer should import the model "as is" without changing the data.
> The data can always be cleaned up in the model processor.

@nkast - Do you agree?

Can/should this functionality be moved to our stock model processor?

nkast:
Either way is good. An Identity bone is not going to cause any problem.

Issue #3822, opened 2015-05-06
Linux game icon not changeable


cra0zy:
Anyway, Icon.ExtractAssociatedIcon is not implemented in mono:
https://github.com/mono/mono/blob/master/mcs/class/System.Drawing/System.Drawing/Icon.cs#L303
and we use that to setup the icon for WinGL and Linux:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Desktop/OpenTKGameWindow.cs#L337

One possible solution for this would be to add Icon property to Game.cs.

PR #3975, opened 2015-06-30
Linux GameWindow Icon Fix

cra0zy:
This fixes the following:
 - not being able to set window icon when on Linux / OS X (#3822)
 - not being able to use mkbundle with MonoGame, mkbundle is a way of bundling c# exe into a native Linux binary that does not require mono for running it. (
<br>old docs: http://www.mono-project.com/archived/guiderunning_mono_applications/
<br>usage with MonoGame: http://youtu.be/rU5TxvD3rzU )


KonajuGames:
Thanks @cra0zy .Merging.

PR #3982, opened 2015-07-04
[Pipeline Tool] Fixed moving action

cra0zy:
Just added a temporary list to collect all the items before moving them, because otherwise it could happen that for loop skips over some item.


cra0zy:
@tomspilman @KonajuGames Can this get merged?

KonajuGames:
This makes sense to me.  Iterating over a collection that is being modified during the iteration is always a problem.  Merging.  Thanks @cra0zy

Issue #3960, opened 2015-06-24
PipelineTool build canceled abruptly with no message


jamesford42:
I have on occasion had a build be abruptly canceled with no feedback to me that it occurred, or any error messages. In fact it looks as if it is still building, perhaps hanging within an individual processor. However that is not what is actually happening. Actually it isn't even building anymore, since the 'build' menu item is enabled and 'cancel' is disabled. I do not know how to reliably reproduce this issue but it happens to me from time to time, and building again will skip past what is already built and start on the next item, and continue onward as expected this time.

KonajuGames:
I've seen that while testing my texture content pipeline.  The message won't be passed through, or a failure message won't be listed.

cra0zy:
We could add a message when the MGCB process is ended, and in case there was no success/failed message we could add one.

jamesford42:
Yea. There must be an exception happening somewhere that is terminating mgcb.exe without the reason being routed to output.

ChewyMoon:
Any news on this? I experience this when I add all of my assets (around 40mb) to the project. I also get this error message when I build the solution in VS.

`
The command ""C:\Program Files (x86)\MSBuild\MonoGame\v3.0\Tools\MGCB.exe" /@:"E:\Coding\Projects\C#\Other\Zeroed\Zeroed\Content\Content.mgcb" /platform:Windows /outputDir:"E:\Coding\Projects\C#\Other\Zeroed\Zeroed\Content\bin\Windows" /intermediateDir:"E:\Coding\Projects\C#\Other\Zeroed\Zeroed\Content\obj\Windows" /quiet" exited with code -532462766.	Zeroed		
`

cra0zy:
I'll set up a PR for this in few hours

ChewyMoon:
I have found the issue after running MGCB manually in a command prompt, this is because something in the content project does not have an importer.

```
Unhandled Exception: System.Exception: Couldn't find a default importer for 'E:/
Coding/Projects/C#/Games/Zeroed/Zeroed/Content/Title/KenVector Future Thin.ttf'!

   at MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.ResolveImporte
rAndProcessor(String sourceFilepath, String& importerName, String& processorName
)
   at MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.BuildContent(S
tring sourceFilepath, String outputFilepath, String importerName, String process
orName, OpaqueDataDictionary processorParameters)
[...]

PR #3986, opened 2015-07-06
[Pipeline Tool] Added message on MGCB Compiler process end

cra0zy:
Fixes: #3960

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -429,7 +429,10 @@ namespace MonoGame.Tools.Pipeline
 
             _buildTask = Task.Factory.StartNew(() => DoBuild(commands));
             if (OnBuildFinished != null)
+            {
                 _buildTask.ContinueWith((e) => OnBuildFinished());
+                _buildTask.ContinueWith((e) => View.OutputAppend("Compiler Ended")); 
+            }
         }
 
         public void Clean()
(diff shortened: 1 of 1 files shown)


tomspilman:
This isn't a fix.  The fix is finding why the MGCB.exe process is leaving without reporting error/success.

cra0zy:
> This isn't a fix. The fix is finding why the MGCB.exe process is leaving without reporting error/success.

MGCB crashes with Native Stacktrace... I have a fbx model that does it, but it's not the only thing that can cause this type of crash. There isn't just one thing that can cause it.

Anyway we do need this because in case that something like this does happen, it would be useful to let people know that compiler process has ended.

Tho it doesn't really fix the main issues.

tomspilman:
> MGCB crashes with Native Stacktrace

Sounds like we are just missing a final `catch (Exception ex)` here:

https://github.com/mono/MonoGame/blob/develop/Tools/MGCB/BuildContent.cs#L307

Specifically that shouldn't happen and we should report it as an unexpected exception.

> I have a fbx model that does it

Sounds like maybe FBX processing is then maybe missing some protection specifically against native code throwing exceptions.

cra0zy:
> Sounds like we are just missing a final catch (Exception ex) here:
> 
> https://github.com/mono/MonoGame/blob/develop/Tools/MGCB/BuildContent.cs#L307
> 
> Specifically that shouldn't happen and we should report it as an unexpected exception.

Doesn't help :\, just tried it.

tomspilman:
> Doesn't help :\, just tried it.

So this is an uncatchable exception?  I forgot these were possible in C#.

So doesn't MGCB.exe return an error code at least to the pipeline tool?

cra0zy:
> So doesn't MGCB.exe return an error code at least to the pipeline tool?

Kind of.

It returns it to the debugger so it writes it out inside Pipeline Tool debugger window.

PR #3987, opened 2015-07-08
Have project selected initially after opening a project.

jamesford42:
Upon opening a content project, the Pipeline Tool will now have the project selected (its properties displayed within the property grid) rather than the last-added content item.

Note: this is a windows-view-specific fix.

### DIFF
--- a/Tools/Pipeline/Windows/MainView.cs
+++ b/Tools/Pipeline/Windows/MainView.cs
@@ -26,6 +26,7 @@ namespace MonoGame.Tools.Pipeline
 
         private bool _treeUpdating;
         private bool _treeSort;
+        private bool _projectLoading;
 
         private const string MonoGameContentProjectFileFilter = "MonoGame Content Build Files (*.mgcb)|*.mgcb";
         private const string XnaContentProjectFileFilter = "XNA Content Projects (*.contentproj)|*.contentproj";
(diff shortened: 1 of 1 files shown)


Tools/Pipeline/Windows/MainView.cs
```diff
@@ -765,6 +770,16 @@ private void UpdateMenus()
...
         }
+
+        private void OnProjectLoading()
+        {
+            _projectLoading = true;
+        }
+
+        private void OnProjectLoaded()
+        {
+            _projectLoading = true;
```
tomspilman: That looks like a bug.

Tools/Pipeline/Windows/MainView.cs
```diff
@@ -77,6 +78,9 @@ public void Attach(IController controller)
             _controller.OnProjectLoading += invokeUpdateMenus;
             _controller.OnProjectLoaded += invokeUpdateMenus;
 
+            _controller.OnProjectLoading += OnProjectLoading;
+            _controller.OnProjectLoaded += OnProjectLoaded;
```
tomspilman: Since they are really never removed.  You could do this as an inline delegate and make the intent more clear.

jamesford42: My own preference is to avoid using lambas in almost all situations. But I can change it if you want.

Tools/Pipeline/Windows/MainView.cs
```diff
@@ -409,7 +413,8 @@ public void AddTreeItem(IProjectItem item)
             node.ImageIndex = iconIdx;
             node.SelectedImageIndex = iconIdx;
 
-            _treeView.SelectedNode = node;
+            if (!_projectLoading)
+                _treeView.SelectedNode = node;
```
tomspilman: Why does this work to ensure the project is selected?  Is it because you are disabling selection of the child items during project loading?

jamesford42: Correct. When the tree root is set (the project) it is selected. Hence it is already in the state we want, we just have to disable the normal behavior of selecting subsequent items that are added during load (after load, you DO want new items added to be selected).

tomspilman:
Old... closing.

PR #3991, opened 2015-07-09
Texture import and processing

KonajuGames:
Texture import and processing is now more complete and more flexible.
* All format conversions go through 128-bit BGRA floating point format (Vector4).
* All major texture compression formats supported: PVR, DXT, ETC1, ATC.
* ETC1 falls back to BGRA4444 if the texture contains any non-opaque alpha values.
* ETC1 is must be square POT, same as PVR, as this is required on some PowerVR chipsets. This is the cause of the slight fuzziness on those images below as they were resized by the texture processor.

image


MonoGame.Framework.Content.Pipeline/Processors/TextureProcessorOutputFormat.cs
```diff
@@ -26,9 +28,28 @@ public enum TextureProcessorOutputFormat
...
+
+        /// <summary>
+        /// The input texture is compressed using PVR texture compression. Used on iOS and some Android platforms.
+        /// </summary>
+        PvrCompressed,
+
+        /// <summary>
+        /// The input texture is compressed using ATI texture compression.  Used on some Android platforms.
+        /// </summary>
+        AtcCompressed,
```
tomspilman: So why have explicit texture formats here?  Wouldn't it be better for the processor to just know if you want compression or you don't?  Let it work out that you use Etc1 or BGRA4444 on Android.

KonajuGames: You could use a specific compression type, for example if your game was
targeting Tegra devices only you will want to use DXT rather than ETC1 and
BGRA4444.

tomspilman: My concern here being people thinking they have to select the right format for each platform manually.

I guess this is mostly a UI issue in the Pipeline tool.  Making sure we descriptions for each enum value appearing in the app with guidance as to when it is appropriate to use each.

KonajuGames: That is something we can address after this is merged rather than delay this further.

dellis1972: My understanding is that if the Format is Compressed the pipeline decides what best to use depending on the platform. However users should be able to specify explicitly.

tomspilman:
Ok... merging!

KonajuGames:
Thanks Tom.

PR #3992, opened 2015-07-10
BoundingSphere additional doc fixes.

Shqrdx:
Docs fixes according to request of @tomspilman.


Shqrdx:
@tomspilman  So its good looks now ? or I should change it more ?

PR #3994, opened 2015-07-11
Fixed threading issue in OggStream

cra0zy:
Fixes threading issue in OggStream, update rate was getting set after the thread... also added an extra check just in case.


tomspilman:
Seems reasonable... merging!

PR #3996, opened 2015-07-12
Android : Housekeeping

rtabbara:
* Fixed support for creating textures with a surface format of type <code>SurfaceFormat.Bgra4444</code>. An internal format of type Rgba4 was specified, which is not supported in GL.ES20.

* Changed existing graphics capabilities checks when reading in a 2d texture to additionally apply to both iOS and Android platforms. In particular, Android devices vary widely, so there's no guarantee that all available texture compression formats will be supported.

* Fixed GraphicsAdapter's DisplayMode dimensions so that immersive support is correctly handled.

* Some other decoupling of the Game instance from internal content-reading logic, making it more consistent with the behaviour of other platforms.


tomspilman:
@KonajuGames  - Can you review this... has a lot of whitespace changes, so need to use the whitespace-less diff.

KonajuGames:
> Changed existing graphics capabilities checks when reading in a 2d texture to additionally apply to both iOS and Android platforms. In particular, Android devices vary widely, so there's no guarantee that all available texture compression formats will be supported.

This is the only minor thing I'd comment on.  This area is solely to decompress DXT at load time for platforms that don't support DXT.  This was to support XNA-created XNB files.  Now that we have content build pipeline support for all texture compression formats, this code will be removed very soon.

Aside from that, it looks good to me.  I'm happy to merge it.

PR #3997, opened 2015-07-12
Fix missing reference to FreeImageNET on Windows

emarcotte:
Doing a plain checkout of develop, I was unable to compile the content pipeline project. Applying this change fixes that.

It looks like the location of the FreeImage library was moved in https://github.com/Mono-Game/MonoGame.Dependencies/commit/5b324d9eae92925b3d7acb0084879dc44d06727b.


tomspilman:
@KonajuGames - ???

KonajuGames:
Hmm.  Yeah.  Valid fix.  Not sure how I missed that or how the build tests passed with that.  I'll merge it now.  Thanks for that.

PR #3999, opened 2015-07-13
Fix preprocessor not working correctly when included files don't end with a newline

tgjones:
Fixes problem 1 in #3853.


tomspilman:
Thanks @tgjones !

PR #4005, opened 2015-07-16
Warning Cleanup 3 of ???¿?

cra0zy:


MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -324,6 +324,13 @@ public RasterizerState RasterizerState
             }
         }
 
+        //TODO Implement
+        public Color BlendFactor
+        {
+            get;
+            set;
+        }
```
tomspilman: What is this?  How is this a warning cleanup?

MonoGame.Framework/Graphics/States/Blend.cs
```diff
@@ -58,11 +58,11 @@ public enum Blend
         /// </summary>
 		InverseDestinationAlpha,
 	    /// <summary>
-        /// Each component of the color is multiplied by a constant in the <see cref="GraphicsDevice.BlendFactor"/>.
+        /// Each component of the color is multiplied by a constant in the GraphicsDevice.BlendFactor.
 	    /// </summary>
 		BlendFactor,
         /// <summary>
-        /// Each component of the color is multiplied by a inversed constant in the <see cref="GraphicsDevice.BlendFactor"/>.
+        /// Each component of the color is multiplied by a inversed constant in the GraphicsDevice.BlendFactor.
```
tomspilman: I suspect that doc string is wrong...  it think meant to say `BlendState.BlendFactor`:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/States/BlendState.cs#L148

cra0zy:
> What is this? How is this a warning cleanup?

Because off the see ref here: https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/States/Blend.cs#L61

> I suspect that doc string is wrong... it think meant to say BlendState.BlendFactor:

1) I accidentally modified that in the first place... was testing something and forgot to return it...
2) No, Xna documentation says otherwise:
   - https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.graphicsdevice.blendfactor.aspx
   - https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.blend.aspx

tomspilman:
> No, Xna documentation says otherwise:

Hah... there are two `BlendFactor`s in XNA:

One in `BlendState`...

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.blendstate.blendfactor(v=xnagamestudio.40).aspx

... and one in `GraphicsDevice`...
https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.graphicsdevice.blendfactor(v=xnagamestudio.40).aspx

So the question is which one is which?  Or are they pointing to the same state?

tomspilman:
Ok... so this is the deal with that:

http://blogs.msdn.com/b/shawnhar/archive/2010/04/02/state-objects-in-xna-game-studio-4-0.aspx

> Assigning to GraphicsDevice.BlendState overrides the current GraphicsDevice.BlendFactor setting  with the BlendFactor value from the specified state object, so order is important. Set the state object first, and the high frequency property second.

It also points out there are two other values like this `MultiSampleMask` and `ReferenceStencil`.

We need to start separate tasks to fix these to work correctly.

Issue #3793, opened 2015-04-30
Manual Test Framework


tomspilman:
We should write a framework for manually run unit tests to augment the existing automated unit test framework.

These manual tests would be specifically for features like audio and input which are most easily validated by people.  The trick would be to make sure the test is quick to do and hard to get wrong.  Some core ideas:

  - The test app should be built and distributed as part of the automated build process.
  - The tests should execute in a fixed order.
  - You must complete one test to move to the next one.
  - Should take very little time to complete the tests... less than a few minutes.
  - The results should contain very basic machine info.
  - The tests should work on all our platforms... including phones.
  - Once all the tests are completed it should display the results, spit out a text file, and display a button to copy the results to your clipboard for easy pasting into GitHub.

The tests themselves should be crafted so that the user has to do the least work possible.  We should ask the user to do something... verify the result wasn't an accident by repeating it...  store the result of the test and move on to the next one.  We **should not** depend on the user to judge themselves if the test passed or not.

The framework for this should be a stander MonoGame `Game` with some sort of `TestScreen` class that is overloaded for different tests.  The framework would then just find them via reflection, sort them in some consistent order, and run them recording results.

The usage for these would be in validating PRs.  If a PR changes something that is covered in the manual tests we need people to run thru them on all the target platforms before approval and merge.  We would also run them once before we do a release.

Thoughts?

cra0zy:
I think this sounds like a good idea, also we would need to check e.g. if device has touchscreen before giving the device a touchscreen test.

tomspilman:
> we would need to check e.g. if device has touchscreen before giving the device a touchscreen test.

Right... it would have to be smart like that.  Which should be easy as MonoGame should be able to tell you if the feature is supported on the current machine.

craftworkgames:
This sounds like a great idea. I'm going to have a bit of spare time up my sleeve shortly, so I might get started on it.

 - What tests do you think we need to cover initially?
 - How are we going to cover each platform?
 - Who is going to run them? Or, how do we motivate people to run them and report issues?

Perhaps it should be a kind of MonoGame "demo" that ships with the installer. It could serve as a set of tests during development and a demo to show off what MonoGame can do after release.

tomspilman:
> What tests do you think we need to cover initially?

Test that require a person to operate.  Any test that can or had been done in our existing unit tests:

https://github.com/mono/MonoGame/tree/develop/Test

Should not be included in a manual testing framework... it would be redundant and useless.

[...]

PR #4007, opened 2015-07-17
Updated OpenTK Window

cra0zy:
With this I've done several things:
 - made the fullscreen not change the resolution but instead just stretch the viewport
   - this also fixes a bug when there are multiple monitors connected, on Linux it would disable the second monitor, and not correctly return their positions properly
 - made the fullscreen switchable during runtime
 - set the window to try and load the information from itself(assembly) if possible
 - Updated to the newest dependencies
   - contains a fix for Mouse Cursor being reset(moved) to the window bound after it looses focus


dellis1972:
@cra0zy Windows DX really stretches stuff like that...? Not sure that is normal XNA behaviour, and it doesn't seem quite right to me.
@tomspilman ?

cra0zy:
> Not sure that is normal XNA behaviour, and it doesn't seem quite right to me.

I'll look into it right away.

cra0zy:
I thought that the WindowsDX had the correct behavior, but I was wrong... anyway I fixed the Window behavior to stretch when in fullscreen, and for while windowed to do this:  http://youtu.be/0wWpdi8El8g (I have no idea what it's called...)

tomspilman:
Could we setup a unit test to verify the correct behavior?

cra0zy:
> Could we setup a unit test to verify the correct behavior?

I think this should go to the manual test framework #3793, we can't really test this correctly with unit tests.

cra0zy:
Don't merge this, I found 2 interesting things I want to add a fix for in here... different graphics drivers do different things....

cra0zy:
Stuff done with last commit:
 - Fixed all graphics related exit errors... the only thing that is left that is close to exit error is: "AL lib: (WW) FreeContext: (0x1ee8210) Deleting 32 Source(s)", which is just an OpenAL warning, and I'll look into it in another PR
 - Improved fullscreen / window switching
 - fixed error that can occur with amd driver

cra0zy:
@tomspilman Whats up with the build system? I looked at error that Win10 server reported and it said it couldn't find reference "System" O o

http://teamcity.monogame.net/viewLog.html?tab=buildLog&buildTypeId=MonoGame_DevelopWin10&buildId=9226#_state=71,72,73,119&focus=180

cra0zy:
Stuff done with the last commit:
 - fixed Windows OpenTK exit error 
 - fixed Fullscreen not settable by default on Windows(was using the wrong GraphicsDevice code)
 - fixed Windows Fullscreen / Window switching behavior, window state was always normal and never false, plus resize was getting called twice....All and all, OpenTK is buggy...
 - made window border on Windows to switch immediately and not wait for delay like Linux does, since the bug only affects Linux

bfloch:
Before submitting a new issue: With 3.4 I see a fullscreen behavior that is quite unexpected compared to XNA.

What I do is calculating a peferredbackbuffer dimension in the clientsizechanged event based on the Window.ClientBounds. When switching to fullscreen the Window.ClientBounds are not correctly updated in clientsizechanged, so that the preferredbackbuffer dimension stay at the last window's dimension = strechting in my case.

In XNA this works flawless. Wondering if you recent work solves this case?

cra0zy:
@bfloch 
1) Are you using MonoGame DirectX? I don't remember ever seeing stretching in MonoGame OpenGL stuff... 
2) Can you post a simple Game1.cs code that shows this bug?

Also, I closed this because I made a pointless mess of stuff I was doing in this. My plan is to split it into multiple PRs for easier code reading starting with #4046 #4054. I need to wait for #4046 to be merged before continuing submitting PRs because otherwise it will create merge conflicts.

bfloch:
@cra0zy Sorry for taking so long. Yes DirectX. Ok I got it figured. Apparently it was my mistake as I thought that ToggleFullscreen changed the Window.ClientBounds and therefore triggered an ClientSizedChanged event.
Apparently it doesn't and the only reason why XNA shows different results is because of its exclusive Fullscreen mode - therefore the small difference in behavior is not really worth mentioning.

I haven't compared to GL though.

PR #4008, opened 2015-07-17
[Gtk Pipeline] Fixed Enum loading

cra0zy:
Stuff done:
 - instead of having specific enum loading for each enum, I just added a code that works for all enums and any future enums
 - fixes not being able to set DebugMode for Effect Processor


KonajuGames:
Removing the specific enum cases is a good thing.  Merging.  Thanks @cra0zy

Issue #3954, opened 2015-06-22
MonoGame Project Creation Tool


cra0zy:
The idea is as follows: There would still be project templates in Visual Studio, except now when a template is selected we would take selected options, and do:

```
mgpct.exe type=WindowsDX location=/home/name/Documents
```

and open the project generated by the tool afterwards.

The tool should have a --no-gui option, and with that it just generates a default project with the other arguments it got.

Advantages:
- same project templates for Visual Studio and MonoDevelop/Xamarin Studio
- Visual Studio Code support
- Easily have advanced options, like if the person wants to use nuget or not
  - another idea: ability to have templates/samples and downloading them only in case a project creation is started with them as base
- would be able to have Local Copy enabled for MonoDevelop's MonoGame local references without any hacks

@dellis1972 @tomspilman @KonajuGames What do you think?

KonajuGames:
How would that work?  Can Visual Studio run an arbitrary executable when
creating a new project from a template?

I have been giving some thought to how to do something similar to XNA and
it's option to create a new project for Xbox 360 or for Windows Phone from
the Windows project.

cra0zy:
> Can Visual Studio run an arbitrary executable when creating a new project from a template?

It should be able to... we would need to override project creation wizard and run it from there.

> I have been giving some thought to how to do something similar to XNA and it's option to create a new project for Xbox 360 or for Windows Phone from the Windows project.

I've never used that feature, can it also do it in reverse?

KonajuGames:
> It should be able to... we would need to override project creation wizard and run it from there.

That would be a Visual Studio extension rather than just a template.

> I've never used that feature, can it also do it in reverse?

It was a right-click on any XNA project in the Solution Explorer, and it could create a project for another XNA-supported platform in the same directory linked to the same source files, then added that project to the solution.

cra0zy:
> It was a right-click on any XNA project in the Solution Explorer, and it could create a project for another XNA-supported platform in the same directory linked to the same source files, then added that project to the solution.

That sounds so good....also it doesn't sound that hard to create.

cra0zy:
I have decided to work on this instead of trying to help with Protobuild.Manager project... to many conflicts of opinion.

What I have so far(mid video is MD addin): https://youtu.be/vMIixCxcGgQ

@dellis1972 Do you want to join me on this?

tomspilman:
[...]

PR #4011, opened 2015-07-18
Moved IDE/MonoDevelop to ProjectTemplates/MonoDevelop

cra0zy:


tomspilman:
@dellis1972 - This make sense to you?

dellis1972:
@tomspilman hmmm not sure.. The ProjectTemplates contains only project templates, this stuff is an IDE extension it does more that templates. There is the Pipeline integration, and with the new MSBuild PR for Mac OS (https://github.com/mono/MonoGame/pull/3953) there is additional project processing. 

If we ever do a VS extension I would imagine it would go in IDE/VisualStudio/  ....

cra0zy:
I feel that VS and MonoDevelop templates/extensions should be within the same folder, maybe choosing some better name for folder than ProjectTemplates?

@dellis1972 Side note for md adding, the project templates group should probably be changed from "Miscellaneous > General" to ".NET > MonoGame"

dellis1972:
@cra0zy the reason they are under Miscellaneous->General is because of recent changes to MonoDevelop they reworked the entire template system, so the old _Category elements are invalid and the Misc->General is a catch all. PR 3953 should fix some of that.. but things are a bit fluid atm from what I can tell and things keep moving around.

on a side note, I'd rather get PR 3953 merged before we start moving stuff around if possible otherwise its going to cause a lot of conflicts my end which I'll need to pick though.

cra0zy:
> on a side note, I'd rather get PR 3953 merged before we start moving stuff around if possible otherwise its going to cause a lot of conflicts my end which I'll need to pick though.

ok, I'll close this until than... I'll also think of a better name for the folder in the meantime

KonajuGames:
@dellis1972 We need a MD extension that uses VS templates.  That would be ideal. :)

dellis1972:
@KonajuGames I have a plan on that. not sure if it will work though.. need to figure it out a bit more

cra0zy:
@KonajuGames I was thinking about that with #3954, it was influenced a bit by how libgdx creates projects...

PR #4016, opened 2015-07-19
Remove PlayStation Mobile support

KonajuGames:
Since Sony have discontinued support for PlayStation Mobile apps, this support should be removed from MonoGame.


tomspilman:
Sucks it turned out this way...  good bye PSM.

PR #4017, opened 2015-07-20
expose SRT Channels

nkast:
Expose Scale/Rotate/Translate channels to content Processors.

```
//ex: access Scale keyframes from content Processor
foreach (var keyframeContent in animationChannel.Scale)
{
        var time = keyframeContent.Key;
        var scale = keyframeContent.Value;
}
```

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/AnimationChannel.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/AnimationChannel.cs
@@ -15,6 +15,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
     {
         List<AnimationKeyframe> keyframes;
 
+        public Dictionary<TimeSpan, Vector3> Translation;
+        public Dictionary<TimeSpan, Vector3> Scale;
+        public Dictionary<TimeSpan, Quaternion> Rotation;
+
         /// <summary>
         /// Gets the number of keyframes in the collection.
         /// </summary>
@@ -54,6 +58,9 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
         public AnimationChannel()
         {
             keyframes = new List<AnimationKeyframe>();
+            Translation = new Dictionary<TimeSpan, Vector3>();
+            Scale = new Dictionary<TimeSpan,Vector3>();
+            Rotation = new Dictionary<TimeSpan,Quaternion>();
         }
 
         /// <summary>
--- a/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
@@ -805,6 +805,26 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
                     }
                 }
 
+                var translateTime = (TimeSpan.TicksPerSecond / aiAnimation.TicksPerSecond);
+                foreach (var key in scaleKeys)
+                {
+                    long ticks = (long)(key.Time * translateTime);
+                    TimeSpan time = TimeSpan.FromTicks(ticks);
+                    channel.Scale[time] = ToXna(key.Value);
+                }
+                foreach (var key in rotationKeys)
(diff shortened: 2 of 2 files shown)


KonajuGames:
What does this give us?  And why Dictionary?  This information is already in the list of keyframes, isn't it?

nkast:
It give us more flexibility to write advanced processors/players that use few keyframes/interpolation, store data in SRT format (Pose), drop SRT channels, etc  

I could write a custom XCollection class and a custom XKeyFrame struct but the only thing that would change is having strong named properties `.Time/.Scale instead of .Key/.Value`. 

Decomposing a Matrix has rounding errors, why not to read the original data?
Also, the Keyframes mixes the channels and interpolates at 60fps(XNA). You don't have the information of which one was KeyFrame for which channel.

KonajuGames:
I think Dictionary is not the best container for that.  Animation channels are linear events sorted by time.  There is no guarantee that iterating through a Dictionary will be ordered by Key or by the order that Keys are added to it.

KonajuGames:
Everything else is fine.  Just the choice of container at issue.

mgarstenauer:
I see one problem here: When the FBX model has transformation pivots the exposed Scale/Rotation/Translation is not the same as the `AnimationKeyFrame`. Since the model processor does not know anything about transformation pivots, the SRT information is kinda useless in this case.

@nkast Any ideas how to handle FBX transformation pivots?

Another question: Some model formats do not store SRT, only the transform matrix. What should model importers do that do not have SRT?

nkast:
@KonajuGames Ok, I could add a generic Keyframe<T>.

@mgarstenauer No idea. Perhaps we can expose them as well so content processors can work with them as well. Do you have a sample FBX with pivot points? 
How are they related to transform channels? can you have pivot points with no other transforms? 

For other formats the Sccale/Rotate/Tranform channels will be null and the processor will have to work with the matrices.

mgarstenauer:
> Do you have a sample FBX with pivot points?

Here is an example: [PlayerMarine](https://sgmotion.codeplex.com/SourceControl/latest#trunk/sgMotionSample/Content/Models/PlayerMarine.fbx) (You can convert the file to a newer FBX version using the [Autodesk FBX Converter](http://usa.autodesk.com/adsk/servlet/pc/item?siteID=123112&id=22694909).)

Most models I have dealt with, unfortunately, had tranformation pivots. One needs to explicitly reset the pivots (e.g. **Utilities | Reset XForm** in 3dsmax) to get rid of them.

> How are they related to transform channels? can you have pivot points with no other transforms?

Not exactly sure what you mean. assimp imports transformation pivots as additional nodes in the bone hierarchy. Example:

    root
      Spine_0_$AssimpFbx$_Translation
        Spine_0_$AssimpFbx$_PreRotation
          Spine_0_$AssimpFbx$_Rotation
            Spine_0
              ...

For more info see class **FbxPivot** in *OpenAssetImporter.cs*.

Exposing the FBX pivots to the content processors is not helpful: To get the actual transform you need to multiply all transformations (pivot nodes and Scale/Rotation/Translation). This is what the OpenAssetImporter is doing, which also means that the individual channels Scale/Rotation/Translation are lost anyway.

> For other formats the Sccale/Rotate/Tranform channels will be null and the processor will have to work with the matrices.

If the MonoGame team decides to merge the change, I recommend to clearly document this in the XML comments of `AnimationChannel`.

However, in my opinion the individual Scale/Rotation/Translation info is rather unreliable (either because the FBX model contains transformation pivots or because the format does not store individual channels) and is therefore not a good candidate to add to the official API. I rather see this handled in a custom model importer if necessary than in the official API.

tomspilman:
@nkast @mgarstenauer - Is this ready to merge?

mgarstenauer:
I don't think that this should go into the official API. See my reasoning above.

nkast:
@tomspilman I'didn't have the time to look into the sample models with root pivot points. Basically, skinning is not in my focus right now and I don't see any related work in the horizon. So, I am going to drop this.

PR #4026, opened 2015-07-21
TouchPanelCapabilities fix for Windows 10

Nezz:
TouchPanel.IsConnected was always false prior to this


tomspilman:
Thanks @Nezz !

PR #4034, opened 2015-07-27
Generating tangent frames

Vodacek:
Added support for generating tangent frames for models.


Vodacek:
@tomspilman , @KonajuGames any comment?

tomspilman:
@mgarstenauer  - This make sense to you?

mgarstenauer:
I suggest the following changes inside the `if (GenerateTangentFrames) { ... }`:
- The method `MeshHelper.CalculateTangentFrames` fails (throws exception) if a `GeometryContent` is missing normal vectors. Therefore, first check whether all `GeometryContent` in `mesh.Geometry` contain `VertexChannelNames.Normal()`. If not, then call `MeshHelper.CalculateNormals` first.
- The mesh may already contain valid tangents and binormals. Calling `MeshHelper.CalculateTangentFrames` in this case overwrites the existing tangent frames, which needs to be avoided. Therefore, check whether all `GeometryContent` in `mesh.Geometry` already contain a tangent or binormal channels and only call `MeshHelper.CalculateTangentFrames` if one is missing.

tomspilman:
Thanks @mgarstenauer ... totally agree and it is obvious now this needs some fixes before we can merge.

Vodacek:
@tomspilman done :-)

mgarstenauer:
Looks good to me!

Vodacek:
Thanks :-) Perhaps now Build bot will be also satisfied.

tomspilman:
Thanks guys!

PR #4039, opened 2015-07-31
Defining basic System.IO types is no longer required by UWP

Nezz:
Defining basic System.IO types is no longer required by UWP as it is now in System.IO.FileSystem.Primitives.

Leaving them there results in build errors in projects referencing MonoGame and using those types.


Nezz:
@tomspilman Can you please re-run the build now that the build server works?

tomspilman:
@Nezz - Same thing here... you need to merge in the latest develop branch changes.

tomspilman:
Thanks... merging!

Issue #2204, opened 2014-01-19
DX11.2/W8.1 Branch


nkast:
I would like to know if anyone is work on DX11.2 for monogame and manage to make it work with XAML/SwapChainBackgroundPanel.

So far, 
I used the "latest Dev Package (2.5.1)" (http://sharpdx.org/news/support-directx11-2) and move them into "MonoGame\ThirdParty\Libs\SharpDX\Windows 8.1 Metro".
Upgrade MonoGame and my Projects to Windows 8.1. 
Change some code in the GraphicsDevice.cs because SwapChain creation on sharpDX has changed.

from:
    _swapChain = dxgiFactory2.CreateSwapChainForComposition(_d3dDevice, ref desc, null);
into:
using (var comDevice = new ComObject(_d3dDevice))
                            _swapChain = new SwapChain1(dxgiFactory2, comDevice, ref desc, null);

When I try to create the SwapChain I get an Exception and the folowing error in the output: 

DXGI ERROR: IDXGIFactory::CreateSwapChain: DXGI_SWAP_EFFECT_FLIP_SEQUENTIAL is not supported on D3D10. D3D11 or later is required. [ MISCELLANEOUS ERROR #99: ]

Same behaviour on two systems, one with ATI DX 10.1 and the second with NVidia DX 9.1.
Both the sharpDX Sample and monogame seems to go through the same simple sequence of initilization. first create the Device, then the  SwapChain with identical parameters.

shiro900:
@nkast  Can you please give me the source code of sharp dx 2.5.1 ? how can i find it ? I posed a question asking in the monogame community too http://community.monogame.net/t/which-sharp-dx-version-are-you-using/2412 .

nkast:
@shiro900 I don't have it. You could search at sharpdx.org or ask at https://github.com/sharpdx/SharpDX 
I theory you can extract it from github history but I am not sure how.

SharpDX is not easy to build. Has a couple custom build tools. I don't expect it to build by simply add the .cproj to your solution.  
Monogame use diferent version of sharpdx for each version of Windows. see https://github.com/Mono-Game/MonoGame.Dependencies/tree/master/SharpDX

Why do you want to build it from source anyway?

shiro900:
@nkast Just out of curiosity and I thought it would be cool to have everything open source, even though im not really gonna change any code of SharpDX

danzel:
Open https://github.com/sharpdx/SharpDX
Click on the "branch: master" drop down
[...]

PR #4041, opened 2015-08-01
Trim DXGI device when suspending on UWP as required by app cert kit

Nezz:
This was not added during transition the to W8.1 (#2204). I don't know if these things are in the right place, but they make it possible to pass the app cert kit.

It would be also great to add similar code to the Windows 8 version as officially it is also required there.


tomspilman:
Great... merging this!

Issue #1954, opened 2013-08-22
Song issue seems to persist on WinDX.


Hardycore:
I saw pull #1874 had apparently fixed this on the WinDX project.  However I'm still getting the exact same error occurring on the catch call.  Same exact issue.  It'll play occasionally, but usually just throws the exception referenced in #1874.

Brought it up here.  https://monogame.codeplex.com/discussions/453847

Could someone confirm this this should be fixed, and if so, what would cause it to still occur for me?  I even went as far as to delete and re-download the whole repo just to make sure I was all up to date.  I'm hoping to release a windows title in a few weeks so I really appreciate any input on this.

MichaelDePiazzi:
FYI - This was fixed by PR #4044

Issue #2620, opened 2014-05-25
Music doesn't play correctly on Windows 7 (DirectX)


DissidentDan:
Hi, When I go to play an wma/xnb song WITHOUT the Visual Studio debugger attached, the first time I play a song, it does not play. Upon subsequent attempts to play ANY song within the same run of the game, I briefly hear the FIRST song play before it fades away (I have some code that fades out one song before fading in another).

If I run the game through the debugger, I get the exception listed below the first time I try to call MediaPlayer.Play(media);. If I then continue exeuction, the song will play normally.The exception will not be thrown after that. However, any subsequent attempts to play songs always play the first song that was played, rather than the correct one.

Additionally, something is weird about this exception, as I put a try/catch block around MediaPlayer.Play, but it does not catch the exception, so I can't report the exact call stack from this exception.

BTW, what is the underlying tech behind playing music? OpenAL, something from Microsoft?

A first chance exception of type 'SharpDX.SharpDXException' occurred in SharpDX.dll

Additional information: HRESULT: [0x80004002], Module: [General], ApiCode: [E_NOINTERFACE/No such interface supported], Message: No such interface supported

PR #4044, opened 2015-08-03
Correct fix for MediaFactory.GetService failure

MichaelDePiazzi:
This should be the correct fix for the MediaFactory.GetService failure and related volume problems* in the Windows DirectX build (See PR #3255 / Issue #2620). This PR may possibly also resolve issue #1954.

(* I was having a problem where after the first song was played, volume changes were not taking effect until the next song was started)

The problem is that GetService cannot be called until the session topology is ready. According to the SharpDX documentation (http://sharpdx.org/documentation/api/m-sharpdx-mediafoundation-mediasession-settopology), when the session is currently paused or stopped, the MediaSession.SetTopology method doesn't even take effect until the next call to Start. And because the method is asynchronous, you need to wait until you receive a "SessionTopologySet" event anyway to know that it's completed.

So I've changed the code to listen for this event. I've tested this on Windows 7 64-bit and it seems to work as expected now (i.e. no crashes on GetService, and volume changes take effect immediately).

PS. A similar fix will also likely be required for VideoPlayer.WMS as it uses the same methods, but I wanted to focus on MediaPlayer first as this is what I was having issues with and could test.

### DIFF
--- a/MonoGame.Framework/Media/MediaPlayer.WMS.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.WMS.cs
@@ -34,8 +34,16 @@ namespace Microsoft.Xna.Framework.Media
             {
                 var ev = _session.EndGetEvent(asyncResultRef);
 
-                if (ev.TypeInfo == MediaEventTypes.EndOfPresentation)
-                    OnSongFinishedPlaying(null, null);
+                switch (ev.TypeInfo)
+                {
+                    case MediaEventTypes.EndOfPresentation:
+                        OnSongFinishedPlaying(null, null);
+                        break;
+                    case MediaEventTypes.SessionTopologyStatus:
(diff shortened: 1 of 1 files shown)


tomspilman:
This is great @MichaelDePiazzi !  Love that someone has finally figure this out correctly!

Can someone give this a test with their own project first?

MichaelDePiazzi:
@tomspilman No worries. Glad I can contribute in some way!

mrhelmut:
I confirm that it works and that it is as clean as possible. Thanks for nailing that one @MichaelDePiazzi 

If it got merged, I'll update VideoPlayer.WMS accordingly and get rid of MediaPlayer.GetVolumeObj().

tomspilman:
Thanks @mrhelmut ...  I'm merging it now!

Issue #4045, opened 2015-08-03
Sound effects no longer work under Windows 7 (DirectX)


MichaelDePiazzi:
PR #4010 has caused a regression - Sound effects no longer work under Windows 7 (for the DirectX build).

Further investigation reveals that SoundEffect(.XAudio).InitializeSoundEffect was swallowing an exception about being unable to load "xaudio2_8.dll". I found a post on the SharpDX forums discussing this issue (http://sharpdx.org/forum/5-api-usage/3334-xaudio2-unable-to-load-dll-xaudio2-8-dll). Long story short - It seems you need to use the DirectX11 SharpDX libraries for Windows 7. The DirectX11.1/11.2 libraries don't seem to be fully supported under Windows 7.

I did a quick hack to prove this by copying the SharpDX 2.6.3 DirectX11 libraries in. I also had to revert the changes made to SoundEffect.XAudio.cs after I did this to get it compiling. Sound effects started working again with these changes.

I'm more than happy to make a pull request for this. But I wasn't sure whether to use the SharpDX "DirectX11-net40" or the "DirectX11-Signed-net40" libraries though...

tomspilman:
> But I wasn't sure whether to use the SharpDX "DirectX11-net40" or the "DirectX11-Signed-net40" libraries though...

@mattleibow ???

MichaelDePiazzi:
@tomspilman I took a quick look at whether the previous SharpDX libraries being used for Windows Desktop were the signed (i.e. strong named) version.

The ones committed by mattleibow are the signed version (which makes sense as the SharpDX packages from NuGet are apparently all using the signed versions, and this is where he got it from). But the previous commit of the libraries (v2.5.1) were the standard version (I believe RayBatts committed these - https://github.com/Mono-Game/MonoGame.Dependencies/pull/6 / https://github.com/Mono-Game/MonoGame.Dependencies/pull/7).

So I guess that doesn't really help answer the question on which to use. And I can't really find any pros or cons to using either in MonoGame. But my opinion is that the standard libraries should be fine as they were being used prior to this. And there doesn't seem to be any particular reason that I can find to start using the signed versions in MonoGame.

mattleibow:
@tomspilman @MichaelDePiazzi It is probably better to use the signed versions everywhere, this allows for other libraries to be signed. For example, I know the CocosSharp team signs their library, and they had problems with the standard dlls.

mattleibow:
[...]

PR #4046, opened 2015-08-03
Fixup for DesktopGL Exit errors

cra0zy:
Fixes: #3540

Stuff done:
 - Fixes BadWindow, BadCursor and OpenTK: Failed to make context current errors
   - was occurring because of window getting Disposed before the rest of the things completed like disposing of Textures(which require a valid window handle), to fix it I close the window at the end of platform.Dispose() so that everything could complete correctly before exiting.
 - Fixes AL Sources Error
   - whoever wrote the code didn't dispose of the sources at the end... nothing major
 - Fixes AL Device Not Closed error
   - was occurring because of the BadWindow and BadCursor errors were stopping dispose from completing
 - Fixed native stacktrace error
   - was occurring because of _toolkit.Dispose()... it's an error within OpenTK, but since _toolkit was an unused variable I just removed it


MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
```diff
@@ -88,13 +88,11 @@ class OpenTKGamePlatform : GamePlatform
 		private OpenALSoundController soundControllerInstance = null;
         // stored the current screen state, so we can check if it has changed.
         private bool isCurrentlyFullScreen = false;
-        private Toolkit toolkit;
         private int isExiting; // int, so we can use Interlocked.Increment
         
 		public OpenTKGamePlatform(Game game)
             : base(game)
         {
-            toolkit = Toolkit.Init();
```
tomspilman: Did this init call really do nothing?

cra0zy: For us it does nothing.

Also from what I read Toolkit is only used if we are combining OpenTK with something else, which we are not.

http://www.opentk.com/files/doc/class_open_t_k_1_1_toolkit.html
> You must call this method if you are combining OpenTK with a third-party windowing toolkit (e.g. GTK#).

tomspilman: Good catch!

cra0zy:
@tomspilman Just FYI, it seems that error 1 could still occur with AMD driver on Linux (Linux usually has 2 graphics drivers, 1 open source and 1 from graphics card manufacturer, and I usually switch between them weekly), so I just modified Dispose method of OpenTKGamePlatform.cs a bit to fix it.

cra0zy:
@tomspilman Can this be merged, I want to submit a PR that will also modify OpenTKGameWindow.cs and I don't want to create conflicts with this one.

MonoGame.Framework/Desktop/OpenTKGameWindow.cs
```diff
@@ -210,39 +214,51 @@ private void Keyboard_KeyDown(object sender, OpenTK.Input.KeyboardKeyEventArgs e
...
 
         internal void ProcessEvents()
         {
-            UpdateBorder();
-            Window.ProcessEvents();
-            UpdateWindowState();
-            HandleInput();
+            lock (window)
+            {
+                if (window == null)
```
tomspilman: What is the goal here?

If `window` is initialized after the `lock` but before the `if` check then it will do the work without locking.

cra0zy: Yea, you are right, but when I thought about it null check is actually useless there now because of some changes I did after I added it originally.

lock is there because resize event also modifies window so that those two don't conflict.

cra0zy:
@tomspilman Can this be merged?

tomspilman:
Anyone have some time to test this?  Ideally someone that uses the OpenTK path in a windowed game that can resize.

cra0zy:
This PR doesn't change any resize behavior....only fixes some stuff during game exit

paewie:
@tomspilman I can test this further (window resizing, OpenGL platforms - did you mean that with "OpenTK path"?) since I am/was experiencing those problems every time on every platform.
I will do so tomorrow or the day after. I will report what happened, if it's helping ;)
@cra0zy Great fix btw. Was getting annoying :) Hopefully this works for me :)

cra0zy:
Again... this does not change any behavior to window resizing....

paewie:
@cra0zy I read that ;) Just offering what was demanded :D

paewie:
Hasn't this fix already been merged into the development branch or did I get something wrong? Because using the development branch build I still experience the same error and some other things that should be fixed (related to their respective issue reports: #4045, fix #4057). I'm feeling like I did something wrong here -.-
Also the Content Pipeline confused me since there are three Linux build targets right now and no WindowsGL anymore (I know of #3831, also the MonoGame.Framework.dll's description is speaking of "all OpenGL Desktop Platforms", but three Linux targets?). One of the Linux targets got my content running on WindowsGL, but still "Failed to make Context 131072 current. Error:6" on exit...
So is this my fault or did I just miss something?

cra0zy:
1) This is not merged yet, it's not in the development branch
2) I noticed the 3 Linux thing bug, #3900 fixes it, kind of, the reason why it occurs is because WindowsGL and Linux enums are pointing to DesktopGL, which are all actually the same value
3) To actually test this PR use: http://teamcity.monogame.net/viewLog.html?buildId=9457&buildTypeId=MonoGame_DevelopWin&tab=artifacts&guest=1

paewie:
Okay, thanks for the info. Guess I'm not really good with GitHub...
I should try The WindowsGL binary i guess?

paewie:
Sadly, neither the WindowsGL nor the Linux MonoGame.Framework.dll did the trick for me. I can't say anything about things on Linux right now.

cra0zy:
Ups, I gave you the bad link, the correct link is: http://teamcity.monogame.net/viewLog.html?buildId=9535&buildTypeId=MonoGame_DevelopWin&tab=artifacts&guest=1

paewie:
xD Okay. But even with the new link stuff crashes.

paewie:
Oops, wait a secons...

cra0zy:
You made me paranoid so I had to switch to Windows and test for exit errors....:
ao

Anyway, just for clarification, now that you used the OpenTK from this branch, you don't have any exit errors?

paewie:
I'm testing every possible combination with the development branch right now...
But till now (using the development Framework and your OpenTK) I still run into those errors. I could try creating a test project, not trying with my current one...

paewie:
Actually I noticed some differences between Win7 and Win10 related to .NET. Since I see you are using Win10 there could be a difference there. I could try that, too...

paewie:
gce

cra0zy:
You need to be using both OpenTK and MonoGameFramework from: http://teamcity.monogame.net/viewLog.html?buildId=9535&buildTypeId=MonoGame_DevelopWin&tab=artifacts&guest=1#!6v3tsaej1vib,-936904oqgja0

Did you remember to cleanup or do rebuild instead of build?

Also can you give me teamviewer access so I can see whats wrong?

paewie:
I just deleted my whole \bin subdir... Nothing. I now double check using the files from the last link you gave me, deleted \bin again. Again: "Error 6"
We could talk about TeamViewer, but not in public xD

cra0zy:
I won't be needing teamviewer

I found out something, the error does not occur with Xamarin Studio, but does occur with Visual Studio. Don't know why yet, I'll look into it now.

paewie:
Okay. Well on Linux as far as I remember MonoDevelop doesn't care about what happens during GC. Xamarin Studio and MonoDevelop are quite similar, aren't they (sharing some base sources etc.)?

cra0zy:
Xamarin Studio is MonoDevelop with few extra extensions.

paewie:
Could it be the Cisual Studio host process? Since my binary seems to work fine by just launching it without VS

cra0zy:
@paewie Try http://teamcity.monogame.net/viewLog.html?buildId=10089&buildTypeId=MonoGame_DevelopWin&tab=artifacts

paewie:
@cra0zy Works perfectly with vshost enabled. Thank you for your great work on that :)
May I ask what the problem was?

paewie:
Damn it xD When hitting the window's exit button SOMETIMES (like every 10 or 20 exits) the OpenTK Errors 2000 or 0 (same description) occur.

cra0zy:
> Damn it xD When hitting the window's exit button SOMETIMES (like every 10 or 20 exits) the OpenTK Errors 2000 or 0 (same description) occur. More likely the 2000 one.

Look at the bright side... there aren't a lot of other exit errors now so we can actually see some exit errors which we wouldn't be able to see otherwise xD

paewie:
Nice point of view xD
But it is much more relaxed now. You don't have to tell VS that everything's fine and you know that that happens ;P
So that far it's a nice fix.

cra0zy:
Tomorrow I'll look into the new error as well as look into making MonoGameGL work correctly with 2 windows.

cra0zy:
@Frassle @thefiddler Is there any way to check if the window handle is valid using OpenTK?

Frassle:
I'd think it should be valid until disposed. Where are you planning on checking for validity?

cra0zy:
Here: https://github.com/cra0zy/MonoGame/blob/otkerrorfixups/MonoGame.Framework/Threading.cs#L203

The problem is that after the game exits, it's gonna dispose the window and than start disposing all the resources, and since the window handle is no longer valid, it's gonna crash.

Frassle:
Wait so maybe I'm confused here but either,
A) You know you've disposed it so assume its invalid
B) You need it valid so make sure you don't dispose it
?

I don't see why you want an internal poke to check the handle validity.

cra0zy:
Well... you gave me an idea:

Currently the window gets destroyed by GC, but the handle stays in Threading.cs, so theoretically speaking if I made Threading.cs keep the window variable instead of window handle I should be able to dispose of it after everything else, will try it out later.

Thanks @Frassle

Frassle:
That sounds like a much cleaner approach :+1:

cra0zy:
I found what the problem was \o/ 

It was the fact that if window.UpdateEvents isn't called, window gets auto destroyed, so just made RunLoop continue a bit longer and vala, no exit error.

 #4153

PR #4049, opened 2015-08-04
Fixed sample length for SoundEffect.FromStream (Windows DirectX)

MichaelDePiazzi:
SoundEffect.FromStream is currently using the wrong sample length (in the Windows DirectX build). It's currently using the length in bytes, when it needs to use the length in samples. The effect of this is that sound effects loaded with FromStream play for too long (and usually end up playing bits of other sound effects).

There was some code correcting for this at some point, but this was removed in PR #3638 because XNB sound effect lengths are already specified in samples.

So this PR calculates the correct length in samples only when using SoundEffect.FromStream. It won't affect loading of XNB sound effects.

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -125,10 +125,11 @@ namespace Microsoft.Xna.Framework.Audio
         {
             var soundStream = new SoundStream(s);
             var dataStream = soundStream.ToDataStream();
+            var sampleLength = (int)(dataStream.Length / ((soundStream.Format.Channels * soundStream.Format.BitsPerSample) / 8));
             CreateBuffers(  soundStream.Format,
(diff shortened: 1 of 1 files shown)


KonajuGames:
Thanks @MichaelDePiazzi . Merging.

SteveDunn:
I've just installed 3.5.0.1433 and the problem still exists

MichaelDePiazzi:
@SteveDunn This is an old pull request that has already been merged some time ago. If you are having a similar problem with the latest build of MonoGame, then please create a new issue with as many details as you can provide (e.g. description of the problem, format of the audio you're using, OS you're using, etc.). That way you've got the best chance of people seeing your issue and being able to help with it.

SteveDunn:
@MichaelDePiazzi - I'll see if I can fix it myself and create another pull request

MichaelDePiazzi:
@SteveDunn Cool, even better! :)

PR #4050, opened 2015-08-04
Only write out GL extensions to Application Output if it's a DEBUG build

cra0zy:
It really clogs up application output to have to always see all the possible GL extensions.


tomspilman:
Seems reasonable for now.  Eventually we may want to add some callback which reports graphics device diagnostic info like this so that users can easily include it in a crash log.

cra0zy:
@tomspilman Can you make me be able to login to community.monogame.net, I've reinstalled my systems few days ago and can't login now(I get a message "Your account has not been approved by a staff member yet. You will be sent an email when it is approved.").

tomspilman:
@cra0zy - Should be fixed.

cra0zy:
Thanks!!!

PR #4055, opened 2015-08-05
[Win Pipeline] Use window color for menubar if on Windows 10

cra0zy:
Just thought it looked a bit ugly, had to change it...

For the question why didn't I use "System.Environment.OSVersion" to check if the OS is Windows 10, that's easy to answer, that would give me a result of Windows 8:

> Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). 

https://msdn.microsoft.com/en-ca/library/windows/desktop/ms724832(v=vs.85).aspx

### DIFF
--- a/Build/GenerateProject.CSharp.xslt
+++ b/Build/GenerateProject.CSharp.xslt
@@ -598,6 +598,15 @@
       </xsl:when>
     </xsl:choose>
     </PropertyGroup>
+	<xsl:choose>
+      <xsl:when test="/Input/Properties/ApplicationManifest">
+        <PropertyGroup>
+            <ApplicationManifest>
+                <xsl:value-of select="/Input/Properties/ApplicationManifest" />
+            </ApplicationManifest>
+        </PropertyGroup>
+      </xsl:when>
+    </xsl:choose>
   </xsl:template>
 
   <xsl:template name="NativeBinary"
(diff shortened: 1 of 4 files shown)


KonajuGames:
If an application has a manifest for Windows 10, will it still run on 8 or 8.1?

cra0zy:
> If an application has a manifest for Windows 10, will it still run on 8 or 8.1?

It should run:

> Manifesting the .exe for Windows 8.1 or Windows 10 will not have any impact when run on previous operating systems. You can also add this to your .rc file if you already have it defined.

https://msdn.microsoft.com/en-ca/library/windows/desktop/dn481241(v=vs.85).aspx

KonajuGames:
Could we do that rather than a string check on the OS name? It feels like
the better solution to me.

tomspilman:
This sort of hack is usually why old versions of software break on newer versions of windows.  We really should not be doing OS version checks and changing code.

What are we trying to fix here?

cra0zy:
> Could we do that rather than a string check on the OS name? It feels like
> the better solution to me.

Didn't find anything in Protobuild about setting application manifests....

cra0zy:
> What are we trying to fix here?

https://msdn.microsoft.com/en-ca/library/windows/desktop/dn481241(v=vs.85).aspx

cra0zy:
@KonajuGames I've done what you've asked

KonajuGames:
Merging. Thanks @cra0zy.

PR #4058, opened 2015-08-06
Added Content Pipeline Extension Project

### LINKED ISSUE #3164: Add Content Pipeline Extension Template for VS
We should add a VS template for custom content pipeline extensions (processors, importers, etc).  This would be similar to the XNA "Content Pipeline Extension Library" template except aimed at the new MonoGame content pipeline.

mattiascibien:
Added a project for creating a content pipeline extension, refrences assemblies from the MGCB directory. I could not provide a Icon since I cannot draw at all.

Fixes #3164.


ProjectTemplates/VisualStudio2010/ContentPipelineExtension/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("$safeprojectname$")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Hewlett-Packard")]
+[assembly: AssemblyProduct("$safeprojectname$")]
+[assembly: AssemblyCopyright("Copyright © Hewlett-Packard 2015")]
```
tomspilman: Can you fix this file....  it should look more like this:

https://github.com/mono/MonoGame/blob/develop/ProjectTemplates/VisualStudio2010/Windows/Properties/AssemblyInfo.cs

tomspilman:
@KonajuGames - Thoughts?  What about the "data" assembly that is shared between the game and the pipeline?  Should we have a separate MonoGame library project or add something to this template?  Or should we just ignore the issue?

ProjectTemplates/VisualStudio2010/ContentPipelineExtension/Properties/AssemblyInfo.cs
mattiascibien: No problem. Sorry if I missed it.

mattiascibien:
@tomspilman I have uploaded the requested fixes, should I squash commits?

tomspilman:
> should I squash commits?

If you want...  don't think it would hurt anything.

mattiascibien:
@tomspilman ok :) thanks

mattiascibien:
@tomspilman Should I also add a content writer inside the project?

dellis1972:
@mattiascibien the Xamarin Studio template is very similar to this you can take a look at the code 

https://github.com/mono/MonoGame/blob/develop/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/

I think its mostly the same, the only real difference is the XS on used the PCL Nuget's for references. But I guess as long as the "MonoGame" are not shipped with the Extension it will work on Mac/Linux and Windows fine.

KonajuGames:
This adds the project templates for VS2010, but does the installer also needs to be modified to copy them into the right location.  How about VS2012, VS2013 and VS2015?

dellis1972:
@KonajuGames it should pick them up other wise there would be no templates in VS2015 as there is currently only one :)

mattiascibien:
@dellis1972 do you think there will be needed more modifications in here?

dellis1972:
@mattiascibien I don't think so.. as long as the XS/VS template do the same thing it will be cool. It would be nice if this was a PCL like the XS one and referenced the Pipeline PCL nuget package, but I don't think its a major issue :)

KonajuGames:
I'll merge this then.  Thanks @mattiascibien.

mattiascibien:
Great @KonajuGames :)

PR #4066, opened 2015-08-09
Updates for UWP NuGet package

BrianPeek:
This requires NuGet 3.1 available here:

http://dist.nuget.org/win-x86-commandline/v3.1.0-beta/nuget.exe

More info on these changes here:

http://blog.nuget.org/20150729/Introducing-nuget-uwp.html

### DIFF
--- a/MonoGame.Framework/Properties/MonoGame.Framework.rd.xml
+++ b/MonoGame.Framework/Properties/MonoGame.Framework.rd.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+    This file contains Runtime Directives, specifications about types your application accesses
+    through reflection and other dynamic code patterns. Runtime Directives are used to control the
+    .NET Native optimizer and ensure that it does not remove code accessed by your library. If your
+    library does not do any reflection, then you generally do not need to edit this file. However,
+    if your library reflects over types, especially types passed to it or derived from its types,
+    then you should write Runtime Directives.
+
+    The most common use of reflection in libraries is to discover information about types passed
+    to the library. Runtime Directives have three ways to express requirements on types passed to
+    your library.
+
+    1.  Parameter, GenericParameter, TypeParameter, TypeEnumerableParameter
+        Use these directives to reflect over types passed as a parameter.
+
+    2.  SubTypes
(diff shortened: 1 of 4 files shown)


tomspilman:
> This requires NuGet 3.1 available here:

When you say it requires NuGet 3.1 does this mean it affect older versions of VS?  I'm thinking about the issue with the `project.json` here.  Will installing NuGet 3.1 make VS2013 have the same issue?

BrianPeek:
No install here, the link is just to the command line nuget.exe which is used to create the packages.  But that leads to...how are you building the NuGet packages today?  Are you running the nuget.exe directly or as an msbuild step or ?

tomspilman:
> link is just to the command line nuget.exe which is used to create the packages

Oh... so this is a package creation thing.

> Are you running the nuget.exe directly or as an msbuild step or ?

It is being run from the build server script directly.  It is currently using nuget 2.8.5 ...  I can make it use a particular version like 3.1 if I need to.

> build the UWP NuGet package using this exe instead of the old one

Would it hurt to build the non-UWP packages with 3.1 too?

BrianPeek:
It **shouldn't** hurt to use the new exe to build the other packages.  It supports both old and new style.  But, note that I haven't tried it.  :)

Nezz:
I don't think that we need an essentially empty rd.xml. If you build from the sources it will create a warning if there are two of them in the same solution. That might be an early VS2015 bug though.

BrianPeek:
@Nezz - I'm not seeing the warning...can you give me some more details?

Nezz:
I'm currently out of the country but I will look it up by the end of the week (unless I forget about it) :)

BrianPeek:
Updated this to remove SharpDX.XInput as it's no longer used in UAP-land. Requires NuGet.exe v3.2 to build (the latest version)

tomspilman:
@BrianPeek - Do you mind rebasing this to the latest?

BrianPeek:
I'm buried but can try to take a look this week. That said, I think everything here was actually fixed in other PRs...we discussed the XInput thing a few weeks ago. I'm not sure there's anything else to do here.

BrianPeek:
Yup, everything here is already in the main tree.  Going to close.

tomspilman:
Thanks @BrianPeek !

PR #4067, opened 2015-08-10
Fixed WP8 Texture2D.SaveAsJpeg with different resolution

Nezz:
We have been using the wrong height and width for almost two years now. If your target size did not match the original size, the method would fail.


tomspilman:
Makes sense.  I'll merge this... however this is really the sort of thing we should have unit tests for.

PR #4068, opened 2015-08-11
Windows 10 BaseStream avoidance

Nezz:
We should no longer call BinaryReader/Writer.BaseStream as it has a huge overhead on UWP


Nezz:
@tomspilman Can you please trigger the build bot for this one?

PR #4075, opened 2015-08-13
Fix for VideoPlayer.WMS volume controller initialisation

mrhelmut:
This is a follow up to #4044

It does an async initialisation of the AudioStreamVolume and cleans up unused methods.


tomspilman:
Thanks @mrhelmut !

PR #4077, opened 2015-08-13
Added new GamePad.UWP that uses Windows.Gaming.Input

cra0zy:


tomspilman:
Can you rename this `GamePad.UWP.cs` ?  Microsoft dropped the "UAP" terminology in favor of "UWP"... Universal Windows App.

MonoGame.Framework/Input/GamePad.UAP.cs
```diff
@@ -0,0 +1,117 @@
...
+                HasLeftYThumbStick = true,
+                HasRightXThumbStick = true,
+                HasRightYThumbStick = true,
+                HasLeftVibrationMotor = true,
+                HasRightVibrationMotor = true,
+                HasVoiceSupport = (gamepad.Headset != null && !string.IsNullOrEmpty(gamepad.Headset.CaptureDeviceId)),
+                HasBigButton = false //we can't detect the big button from Windows.Gaming.Input.Gamepad, so it's always false
+            };
+
+            throw new NotImplementedException();
```
tomspilman: This looks like it should be removed.

cra0zy:
https://www.youtube.com/watch?v=Kd71zO6XAC8

MonoGame.Framework/Input/GamePad.UAP.cs
cra0zy: ups, good point

cra0zy:
ups found 1 more dumb thing I did at the end... one sec....

Nezz:
@cra0zy Can you please remove the reference to XInput here?
https://github.com/cra0zy/MonoGame/blob/win10gamepad/Build/Projects/FrameworkReferences.definition#L207-L209

Nezz:
We also need to deal with the deadzone in `GamePadThumbSticks.cs`. I couldn't find any reference for deadzone, so it's probably best to use the defaults.

Nezz:
With the changes I mentioned, it is no longer crashing on Windows Phone. We probably even have GamePad support on mobile:

```
'SongArc.exe' (Win32): Loaded 'C:\Windows\system32\Windows.Gaming.Input.dll'.
```

theZMan:
@Nezz - is this the sort of deadzone reference you were looking for ? http://blogs.msdn.com/b/shawnhar/archive/2007/03/28/gamepads-suck.aspx

Nezz:
@theZMan I was looking for something like this in the new API:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Input/GamePadThumbSticks.cs#L137-L138

cra0zy:
> @cra0zy Can you please remove the reference to XInput here?
> https://github.com/cra0zy/MonoGame/blob/win10gamepad/Build/Projects/FrameworkReferences.definition#L207-L209

done

Nezz:
Thanks @cra0zy, works fine. We can merge this once the build finishes :)

Nezz:
@tomspilman I think this is ready to merge. @mrhelmut tested it too.

tomspilman:
Thanks guys... merging!

KakCAT:
Hi!

is it possible that this commit broke the "back" button in Windows 10 mobile? I can't make it work with the standard back button detection code but a build from August 1st detects it without problems.

thanks!

KakCAT:
tracked the problem a little bit more:

The code I use to detect the back button is:

    if (GamePad.GetState(PlayerIndex.One).Buttons.Back == ButtonState.Pressed) 
        doThings();
               


Problem seems to be that in function

private static GamePadState PlatformGetState(int index, GamePadDeadZone deadZoneMode)

goes through

            if (index >= WGI.Gamepad.Gamepads.Count)
                return GamePadState.Default;


because I have no gamepad connected on the phone. Is this a bug or I'm querying the back button in a wrong way?

mrhelmut:
Ah! Good catch, it's an issue.
```Back``` should be inserted in ```GamePadState.Default``` when there's no gamepad. I'll try push a PR by tomorrow.

Nezz:
Nice catch.  We've been wondering what went wrong.

mrhelmut:
I've pushed a PR to get around this issue. #4102

PR #4078, opened 2015-08-14
[WIP] Mesh builder

### LINKED ISSUE #4070: Implement MeshBuilder
We need to implement the `Microsoft.Xna.Framework.Content.Pipeline.Graphics.MeshBuilder` class:

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.content.pipeline.graphics.meshbuilder.aspx

The implantation should include unit tests to verify it is working correctly.

mattiascibien:
A Work in Progress Implementation of mesh builder. Closes #4070. Opening as a WIP to seek help in implementation and discussion.

EDIT: at the end I am going to squash everything in one commit but for the sake of learning I am going to keep this as is until then.

### DIFF
--- a/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
+++ b/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
@@ -111,6 +111,7 @@
     <Compile Include="Graphics\IndexCollection.cs" />
     <Compile Include="Graphics\IndirectPositionCollection.cs" />
     <Compile Include="Graphics\MaterialContent.cs" />
+    <Compile Include="Graphics\MeshBuilder.cs" />
     <Compile Include="Graphics\MeshContent.cs" />
     <Compile Include="Graphics\MeshHelper.cs" />
     <Compile Include="Graphics\MipmapChain.cs" />
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -168,6 +168,9 @@
     <Compile Include="ContentPipeline\IntermediateSerializerTest.cs">
         <Platforms>Windows,MacOS,Linux</Platforms>
     </Compile>
+    <Compile Include="ContentPipeline\MeshBuilderTests.cs">
+        <Platforms>Windows,MacOS,Linux</Platforms>
+    </Compile>
     <Compile Include="ContentPipeline\ModelProcessorTests.cs">
         <Platforms>Windows,MacOS,Linux</Platforms>
     </Compile>
--- a/MonoGame.Framework.Content.Pipeline/Graphics/MeshBuilder.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/MeshBuilder.cs
@@ -1,0 +1,132 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
+{
+    public sealed class MeshBuilder
+    {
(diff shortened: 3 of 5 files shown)


tomspilman:
This looks like a good start @mattiascibien .  I would suggest setting up a unit test next as it will help tremendously in ensuring we get the same results as XNA.

mattiascibien:
@tomspilman that was one of the thing I was wondering on how to implement (I am not such an expert and the few unit tests I made were only for Business Applications) but anyway I think that the direction I am taking with the code seems actually the correct way to do. Is the code for the Open Asset Importer a good start?

tomspilman:
> Is the code for the Open Asset Importer a good start?

I would suggest starting with the code I wrote for the `ModelProcessor` tests:

https://github.com/mono/MonoGame/blob/develop/Test/ContentPipeline/ModelProcessorTests.cs

In your case you instead would be using the `MeshBuilder` to generate a `MeshContent` object then asserting that the `MeshContent` is what you expected it to be.


> is there any reason why opaquedata is readonly? 

If you mean this...

https://github.com/mono/MonoGame/blob/1756c713b4aede3db5568f69012daf68be2d8c6a/MonoGame.Framework.Content.Pipeline/ContentItem.cs#L30

It is because that is how it is defined in XNA:

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.content.pipeline.contentitem.opaquedata.aspx

Still you don't need to change the dictionary there... you just need to add information to it which is totally allowed.

tomspilman:
@mattiascibien - Please merge the latest develop updates into this branch.

mattiascibien:
@tomspilman Will do. Thanks

mattiascibien:
@tomspilman started to test this out. I guess that an implemetation for `MergeDuplicateVertices` and `MergeDuplicatePositions` is needed before continuing with this.

tomspilman:
You can maybe get away without it as long as your tests are careful to avoid cases where verts/positions would be merged.  The tests will at least pass for now.

Still we do need implementations and tests for both of those.

mattiascibien:
@tomspilman I tried the GeneratedGeometry Sample for AppHub but for what it seems I do not understand how to keep the whole thing regarding GeometryContent. It seems that only one GeometryContent will be added in the end. I originally thought that a GeometryContent was going to be created every time I called setMaterial or set opaquedata. I cannot undestand which assumptions can be made at the bottom level of the thing.

tomspilman:
This is what the unit tests help with.  Write the tests first and run them against XNA (we have an XNA project for the unit tests here:  https://github.com/mono/MonoGame/tree/develop/Test).  Then you know what the results should be for different inputs.

mattiascibien:
@tomspilman Check my updated comment before. I will have a look on hot to make some unit test for that. Thanks.

mattiascibien:
@tomspilman Something different i found between monogame and XNA while running regarding the absolute transform. In monogame it is a Matrix full of zeroes while in XNA it seems to be an Identity.

mattiascibien:
Wrong button, Sorry

mattiascibien:
@tomspilman making some progress here. I am also using the generated geometry sample to check this.

tomspilman:
@mattiascibien - What is the status of this PR?

mattiascibien:
@tomspilman sorry. I did not had much time to work on it :(

DDReaper:
Seems this item has been superseded by #5408   and should be closed.

mattiascibien:
@DDReaper no problems, I will close it

Jjagg:
Thanks for the effort @mattiascibien, this was a great starting point :)

PR #4079, opened 2015-08-15
Linux Installer Update

cra0zy:
Stuff Done:
 - renamed packages to monogame-sdk
 - included monodevelop addin
 - added a check to see if the any previous version of MonoGame is installed


tomspilman:
Ok... merging!

PR #4083, opened 2015-08-16
Don't limit the amount of GamePads that can be used

cra0zy:
I am pretty sure that XInput is the only thing that actually has the limit of 4 gamepads... no point in limiting the rest because of it...

### DIFF
--- a/MonoGame.Framework/Input/GamePad.XInput.cs
+++ b/MonoGame.Framework/Input/GamePad.XInput.cs
@@ -26,6 +26,10 @@ namespace Microsoft.Xna.Framework.Input
 
         private static GamePadCapabilities PlatformGetCapabilities(int index)
         {
+            // Make sure the player index is in range.
+            if (index < (int)PlayerIndex.One || index > (int)PlayerIndex.Four)
+                throw new InvalidOperationException();
+
             // If the device was disconneced then wait for 
             // the timeout to elapsed before we test it again.
             if (!_connected[index] && _timeout[index] > DateTime.UtcNow.Ticks)
@@ -147,6 +151,10 @@ namespace Microsoft.Xna.Framework.Input
 
         private static GamePadState PlatformGetState(int index, GamePadDeadZone deadZoneMode)
         {
+            // Make sure the player index is in range.
+            if (index < (int)PlayerIndex.One || index > (int)PlayerIndex.Four)
+                throw new InvalidOperationException();
+
             // If the device was disconneced then wait for 
             // the timeout to elapsed before we test it again.
             if (!_connected[index] && _timeout[index] > DateTime.UtcNow.Ticks)
@@ -298,6 +306,10 @@ namespace Microsoft.Xna.Framework.Input
 
         private static bool PlatformSetVibration(int index, float leftMotor, float rightMotor)
         {
+            // Make sure the player index is in range.
+            if (index < (int)PlayerIndex.One || index > (int)PlayerIndex.Four)
+                throw new InvalidOperationException();
+
 #if DIRECTX11_1
             if (!_connected[index])
                 return false;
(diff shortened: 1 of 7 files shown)


tomspilman:
@KonajuGames ?

So i think most platforms support more than 4 now... so it wouldn't be bad to support more.

The question is what do we do with the `PlayerIndex` enum?  Do we extend it to a new default maximum of 8 or something?

KonajuGames:
The `PlayerIndex` enum was always a bit awkward to use.  I always used it as an integer anyway and cast to the enum type when required.  I think this is a good extension to the API and there is no need to extend the enum to suit.

cra0zy:
@tomspilman Something is not right with the build server, #4078 crashed with the same error while processing Web assembly just like here.

tomspilman:
I am aware of it... i'll have it fixed soon.

mattiascibien:
@tomspilman Was something made by me with #4078?

tomspilman:
@cra0zy - If you merge the latest develop changes into this branch things should build now.

cra0zy:
done

cra0zy:
A friend of mine needs this so.... can this get merged any time soon?

endimilojkoski:
Hello guys. First, thanks for the awesome job that you are doing with Mono Game!

I really need to know if GamePad support for 8 controllers will be available.
I am using Joystick for 8 controllers support but it has issues.
1. So far I had 2 cases where the game would crash and if I don't allow controllers support in game it works. I can't reproduce this but the problem is with Joystick indexes and IsConnected property.
<b>I have found by remote testing that whatever index you test for Joystick it says it is Connected even with no controllers connected!</b>
2. Dpad access through Hats[] is not working on Linux and Mac, only on Windows. Mac and Linux use buttons around index 10 for Dpad.

My game Raining Blobs is almost ready for Steam release and it can be played by up to 8 players at once so controllers support is one of the most important parts.

I am fine with using the developer branch, I just need to know how to act around this asap.

tomspilman:
@cra0zy - Do you mind making a change to this?

How about we add `GamePad.PlatformGetMaxIndex()` and let individual platforms overload this?  This way all the control index validation can live in the common `GamePad` code like it did before:
```
if (index < (int)PlayerIndex.One || index > PlatformGetMaxIndex())  
   throw new InvalidOperationException();  
```

cra0zy:
Done, I was not sure how many Gamepads OUYA can use so I set it to 4 since that was the default.

MonoGame.Framework/Input/GamePad.XInput.cs
```diff
@@ -298,6 +311,10 @@ private static GamePadState PlatformGetState(int index, GamePadDeadZone deadZone
 
         private static bool PlatformSetVibration(int index, float leftMotor, float rightMotor)
         {
+            // Make sure the player index is in range.
+            if (index < (int)PlayerIndex.One || index > (int)PlayerIndex.Four)
+                throw new InvalidOperationException();
+
```
tomspilman: Can you remove the extra bounds check here in the `GamePad.XInput.cs` file?  Now that the public interface does the test this is just redundant code.

cra0zy: It has been done, or it has been undone, depending on how you look at it.

tomspilman:
Thanks @cra0zy !

endimilojkoski:
This works great! No more game crashing issues as it was the case with Joystick on some PCs. Tried with 5 different gamepads and everything works great. Will let you know if I find anything.
Thanks.

Helikin:
I think the max number of supported gamepads is off by one. For example on XInput:

    private static int PlatformGetMaxIndex()
    {
         return 4;
    }

It is used like this:

    if (index < 0 || index > PlatformGetMaxIndex())
        throw new InvalidOperationException();

That means indices 0, 1, 2, 3, 4 are valid and XInput supports 5 gamepads?

PR #4089, opened 2015-08-18
Remove PSMoble From Docs

tomspilman:
Remove depreciated PSMoble from doc generation which was missed in https://github.com/mono/MonoGame/pull/4016.


Shqrdx:
The psm classes still exist in the docs on site..

Issue #3884, opened 2015-05-25
three isses with SharpDX/MediaPlayer (MonoGame 3.4.0.456)


BenMakesGames:
[edit] after playing around more, I discovered that the problems I was having locally are due to inconsistent application of Volume settings. This is a DIFFERENT sort of bug, and I uncovered an additional bug after further testing. I have edited the post accordingly.

---

I installed CrashReporter.NET into my project recently, and have been getting several reports related to MediaPlayer. I'm unable to reproduce these locally, but I've noticed some MediaPlayer issues of my own, and I have to wonder if they could be related.

First, the issues I have locally:

I am using MonoGame 3.4.0.456. My machine runs 64-bit Windows 7.

1. The first time a song is played, I receive "A first chance exception of type 'SharpDX.SharpDXException' occurred in SharpDX.dll". The game continues running as normal for me, however I wonder if this is the cause for some of the crashes I've had reported via CrashReporter.NET (below).

2. I set MediaPlayer.Volume every time a song is played, however the first time a particular song is played, this Volume is not respected; the song plays at 100% Volume. Subsequent plays of the same song respect the Volume I've set, but every time a song is played that has never been played before, the Volume is set back to full. (MonoGame seems to have some memory of Volume on a per-song basis, which fails to be applied the first time??) When a song loops - via MediaPlayer.IsRepeating = true - this seems to count as a second playthrough, and the previously-set Volume suddenly applies itself! I tried moving the code that sets MediaPlayer.Volume to both before and after the MediaPlayer.Play(...) call, but in both places, these symptoms remain.

3. Sometimes, a song simply refuses to play. This seems to happen randomly, to any song. There is no message in the console, or any other indication that something went wrong, and MediaPlayer.Status reports that the song is Playing.

4. I created a sound/song test menu within my game to facilitate easily testing the above.

Second: two MediaPlayer-related issues have been reported via CrashReporter.NET, but I have not been able to reproduce them locally. Here are the stack traces; I've truncated them at the point where it gets into my code, since I imagine that's not useful, but I can get the full stack traces including those, if it WOULD be helpful:

This report has only come from Windows 7 machines. I can't necessarily tell how many are UNIQUE reports, but I saw two different versions of .NET (4.0.30319.34209 and 4.0.30319.18063), so there are at least two unique versions (unless one player attempted to upgrade .NET to solve the issue?)

```
[...]

PR #4057, opened 2015-08-05
Reverting audio changes (#4010) due to bug on Windows 7 (#4045)

mattleibow:
There seems to be a problem when using DX11.1 and 11.2 on Windows 7:
http://sharpdx.org/forum/5-api-usage/3334-xaudio2-unable-to-load-dll-xaudio2-8-dll

This commit reverts the requirement for the newer version of DX, and just uses DX11.
This should now work on both Windows 7 and Windows 8.

The dependencies uses my revision, but before it is pulled, I'll change it.

See issue: https://github.com/mono/MonoGame/issues/4045

PR #4090, opened 2015-08-18
Fix for songs sometimes not playing (Windows DirectX)

MichaelDePiazzi:
I've had some issues with the Windows DirectX build where songs sometimes didn't play as follows:

* On my Win7 desktop, the problem only happened occasionally and was difficult to reproduce. When it did happen, the song would start playing but then stop very shortly after.
* On my Win10 laptop, the problem was much easier to reproduce. Basically I'd play a song, stop it, play another song, stop it, and then play the first song again (which would then not start playing at all).

I believe issues #1954 and #3884 are related (and likely will be resolved by this PR and/or my previous PR #4044).

This was quite a pain to resolve (I spent the past 2 days working on it)... Eventually it mostly came down to cross threading issues, not waiting for asynchronous methods to complete when required, and unnecessary/incorrect use of some MediaSession functions.

Anyway, it should hopefully be a lot cleaner and more robust now. I have not been able to reproduce the issues I was having, so I believe they're resolved now.

VideoPlayer.WMS will most likely need similar fixes (since it uses the same APIs), but I don't have anything that uses it and can't afford the time to investigate further. If there are similar issues though, hopefully someone else can just look at what I've done here to resolve them.

### DIFF
--- a/MonoGame.Framework/Media/MediaPlayer.WMS.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.WMS.cs
@@ -17,8 +17,11 @@ namespace Microsoft.Xna.Framework.Media
         private static MediaSession _session;
         private static AudioStreamVolume _volumeController;
         private static PresentationClock _clock;
-        private static Song _newSong;
+        private static Song _nextSong;
         private static Song _currentSong;
+
+        private enum SessionState { Stopped, Stopping, Started, Paused, Ended }
+        private static SessionState _sessionState = SessionState.Stopped;
 
         private static Guid AudioStreamVolumeGuid;
 
@@ -57,6 +60,7 @@ namespace Microsoft.Xna.Framework.Media
             switch (ev.TypeInfo)
             {
                 case MediaEventTypes.EndOfPresentation:
+                    _sessionState = SessionState.Ended;
                     OnSongFinishedPlaying(null, null);
                     break;
                 case MediaEventTypes.SessionTopologyStatus:
@@ -121,7 +125,7 @@ namespace Microsoft.Xna.Framework.Media
 
         private static TimeSpan PlatformGetPlayPosition()
         {
-            if (State == MediaState.Stopped)
+            if ((_sessionState == SessionState.Stopped) || (_sessionState == SessionState.Stopping))
                 return TimeSpan.Zero;
             try
             {
@@ -170,30 +174,47 @@ namespace Microsoft.Xna.Framework.Media
 
         private static void PlatformPause()
         {
+            if (_sessionState != SessionState.Started)
+                return;
+            _sessionState = SessionState.Paused;
             _session.Pause();
         }
 
         private static void PlatformPlaySong(Song song)
         {
             if (_currentSong == song)
+                ReplayCurrentSong(song);
+            else
+                PlayNewSong(song);
+        }
+
+        private static void ReplayCurrentSong(Song song)
+        {
+            if (_sessionState == SessionState.Stopping)
             {
-                _session.Start(null, PositionBeginning);
+                // The song will be started after the SessionStopped event is received
+                _nextSong = song;
                 return;
             }
 
-            if (State != MediaState.Stopped)
+            StartSession(PositionBeginning);
+        }
+
+        private static void PlayNewSong(Song song)
+        {
+            if (_sessionState != SessionState.Stopped)
             {
                 // The session needs to be stopped to reset the play position
                 // The new song will be started after the SessionStopped event is received
-                _newSong = song;
-                _session.Stop();
+                _nextSong = song;
+                PlatformStop();
                 return;
             }
 
-            StartSong(song);
+            StartNewSong(song);
         }
 
-        private static void StartSong(Song song)
+        private static void StartNewSong(Song song)
         {
             if (_volumeController != null)
             {
@@ -205,10 +226,16 @@ namespace Microsoft.Xna.Framework.Media
 
             _session.SetTopology(SessionSetTopologyFlags.Immediate, song.Topology);
 
-            _session.Start(null, PositionBeginning);
+            StartSession(PositionBeginning);
 
             // The volume service won't be available until the session topology
             // is ready, so we now need to wait for the event indicating this
+        }
+
+        private static void StartSession(Variant startPosition)
+        {
+            _sessionState = SessionState.Started;
+            _session.Start(null, startPosition);
         }
 
         private static void OnTopologyReady()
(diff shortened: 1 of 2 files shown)


mrhelmut:
Hey there,

I tested your fixes, while the changes in ```MediaPlayer.WMS``` are fine, the ones in ```Song.WMS``` seem to fail. Only the first ```Song``` plays and any subsequent ```MediaPlayer.Play``` with a new ```Song``` fails to start (no exception thrown, it just doesn't play).

It seems that ```_mediaSource.Shutdown();``` in ```PlatformDispose``` causes this problem. Looks like it also shutdown the media session but what's strange is that it should throw a MF_E_SHUTDOWN.

See: https://github.com/MichaelDePiazzi/MonoGame/commit/e4380434da187e2964b838065b0c20a00d474bf0#diff-a42f1080f1cd56ff95300144ebbb9009R92

P.S.: I'm stopping and unloading the previous Song before loading and playing the next one, so the bug may not appear if you keep everything loaded inbetween plays.

MichaelDePiazzi:
@mrhelmut Thanks, good find! I reproduced the problem you described. I didn't see it because I'm loading all my content up front and only unloading it on app exit.

I probably over-reached a little by making those changes to Song.WMS since technically they weren't actually related to any of the issues I was having. So to resolve this for now, I've just changed the MediaSource handling back to what it was before (i.e. no call to Shutdown).

Although this probably needs a closer look at some point since the documentation (https://msdn.microsoft.com/en-us/library/windows/desktop/ms703110%28v=vs.85%29.aspx) mentions that the call to Shutdown is required to avoid potential memory leaks (for both MediaSource and MediaSession). But this will probably require some bigger changes to handle this correctly.

mrhelmut:
You are right. The Shutdown doc is a bit confusing, I haven't figured out when to call it exactly and why shutting down a media source would shut the media session too (and if the session still need shutdown to be called after that or not).

mrhelmut:
Hey @MichaelDePiazzi , did you made any progress? I noticed that I have the exact same issue (point 3 of #3884).

For some reason, some Songs won't play. I noticed that it happens only when the previous song has completely played,  then if the next has already played, it may not replay again. I'm not sure and I'm having difficulties tracking this, but I think it's related with the clock not being properly initialized after a song has been stopped naturally (when that occurs, MediaPlayer.PlayPosition seems off from the start instead of being 0).

I have tested your fix against this particular issue and it didn't fixed it.

MichaelDePiazzi:
Hey @mrhelmut. No, as far as I was aware this was all working now. I resolved the issue you had previously by getting rid of the call to "MediaSource.Shutdown" I had introduced. I haven't experienced the issue since (or heard any reports of it from my play testers).

If you're referring to the proper usage of MediaSource/MediaSession.Shutdown, this was something I was hoping to look at after this PR gets merged (as it will likely require bigger changes to do).

I'm not sure why you're still having issues with this, but I do know from first hand experience that I had a lot of trouble getting it working on my PCs here. The MediaFoundation API seems to be very sensitive and one thing out of place seems to stop it from working without any errors.

I did also have some issues with the clock while working on this PR when playing a different song. I found that it wouldn't reset unless I stopped the session, waited for the "SessionStopped" event, and then started it with the new song.

But anyway, I'll try to follow the same steps you took (ie. wait for a song to finish playing, and then play a different song) and see if I can reproduce the issue you're having.

MichaelDePiazzi:
@mrhelmut PS. Just rebased this PR on develop to resolve the conflicts with your PR #4109

MichaelDePiazzi:
@mrhelmut PPS. Just a side thought - I wonder if you're issues are possibly related to issue #4045 / PR #4057? I've manually included this PR in my build and have been doing my testing with it included...

MichaelDePiazzi:
@mrhelmut Ok, I've at least managed to reproduce the issue by playing 2 very short songs one after the other continuously. Eventually it does stop playing for me like you've found. So I'll try and mess around with this some more today and see if I can find a solution...

MichaelDePiazzi:
@mrhelmut I think I've fixed the issue. What it mainly came down to was that after a song finished playing, for some reason the play position needed to be reset or the next song may not start playing. I'm not sure why this is, but this seemed to get it working reliably. I'm now also keeping track of the actual session state which should hopefully help to avoid other potential edge cases which could throw things out of whack.

Let me know if this fixes things for you or not.

mrhelmut:
Will test that today!

mrhelmut:
Know what @MichaelDePiazzi ? Looks like you nailed it! :)
We extensively tested this PR and there was no problem at all.

Awesome work, it really wasn't trivial to sort this out. This should fix a lot of (if not all) issues related to MediaFoundation.

I think it is time to summon @tomspilman ? ^^

MichaelDePiazzi:
@mrhelmut Awesome, glad it's working now! :) Thanks for testing this!

mrhelmut:
Damn, I fear that there is yet another (new) problem... With this PR, if I play a Song, unload it, reload it and play it again, the thread freezes on this line: https://github.com/mono/MonoGame/pull/4090/files#diff-a42f1080f1cd56ff95300144ebbb9009R31

and doesn't return.

I'm having a closer look at this right now.

P.S.: all my loadings are done in a separate thread.

mrhelmut:
Found it. Apparently executing the state change callback on the UI thread make this to happen. If I remove that part everything is fine.

Do you think it really is a problem to not do so?

MichaelDePiazzi:
@mrhelmut Hmmm... can maybe get rid of it... Funnily enough, I actually added this to solve a similar issue though...

One of my original solutions to this problem involved creating a new session every time a song was played, and then completely closing, shutting down, and disposing of everything when it was stopped. But I was getting a dead-lock when playing another song - specifically on "SourceResolver.CreateObjectFromURL". I found a forum post talking about a similar issue (https://social.msdn.microsoft.com/Forums/en-US/90f8464d-041d-4f14-bb5d-a5de8098309c/source-resolver-createobjectfromurl-hangs-when-called-from-bytestreamhandler?forum=mediafoundationdevelopment) and realised this synchronous method was blocking the thread that the event was raised on. So I raised the event on the UI thread instead and this solved it.

Unfortunately I had to abandon this solution because I got strange playback issues which I couldn't explain or resolve. But I kept the part which raised the events on the UI thread as it made sense to avoid any other potential cross-thread issues.

So we can maybe get rid of it now since I'm no longer trying to load a song in the SessionStopped event... But the underlying issue is still there, so it may still happen in other circumstances... To be honest, I'm no expert when it comes to multi-threading and I'm not really sure about this.

I assume everything still works for you with the UI thread code removed? If so, I guess this is the most practical solution at hand right now so we can at least get these fixes merged. (We can then think about shutting down everything correctly and avoiding memory leaks next)

PS. The forum post mentions using the async version of the method (i.e. BeginCreateObjectFromUrl) to get around this issue. Unfortunately, it is not exposed in SharpDX (see https://github.com/sharpdx/SharpDX/issues/502). However, someone posted a workaround in that same issue which uses reflection to get access to it. So this might be a more viable solution to look at later.

MichaelDePiazzi:
@mrhelmut I just tested this myself with the UI thread stuff reverted, and it all seems to work fine still for me. So I've just pushed this change reverting it. Hopefully this is enough to get this across the line now...

mrhelmut:
Yes, without the UI thread it works correctly. :)

I think what you tried to do was the right path, the discussion in #4121 tends to that too, but it looks like it won't be possible to avoid all deadlocks unless the Play method gets fully async, which would break XNA compat'.

Anyway, this PR fixes so many MediaPlayer issues that it's worth merging IMO, even without all the async stuff fully implemented. What do you think @tomspilman ?

MichaelDePiazzi:
@KonajuGames @tomspilman Hey guys, I believe this should be ready to merge now. @mrhelmut has thoroughly tested my changes and it all seems to be working reliably. Definitely there is further work to be done in this area to avoid potential memory leaks (see issue #4121), but I feel this is best tackled in a separate PR as it will likely require bigger changes. This PR at least fixes the existing song playback issues and gets things a step closer. What are your thoughts?

KonajuGames:
Thanks for the extensive testing on this, @mrhelmut and @MichaelDePiazzi .  I'm happy to merge this now.

MichaelDePiazzi:
Thanks @KonajuGames

MichaelDePiazzi:
@KonajuGames PS. Issues #1954 and #3884 can now probably be closed as this PR (and my previous PR #4044) resolves them.

KonajuGames:
Thanks @MichaelDePiazzi

mrhelmut:
Super great. :)

PR #4092, opened 2015-08-19
Add Windows 8.1 project generation to Protobuild definitions.

kjpou1:
This will create project and solutions for Windows 8.1 version of MonoGame so that it can be loaded in Visual Studio 2015.

### DIFF
--- a/Build/GenerateProject.CSharp.xslt
+++ b/Build/GenerateProject.CSharp.xslt
@@ -277,6 +277,10 @@
           <xsl:when test="/Input/Generation/Platform = 'Windows8'">
             <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
           </xsl:when>
+          <xsl:when test="/Input/Generation/Platform = 'Windows81'">
+            <TargetPlatformVersion>8.1</TargetPlatformVersion>
+            <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
+          </xsl:when>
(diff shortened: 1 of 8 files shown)


kjpou1:
Woops. That should have read "This will now load in Visual Studio 2015"

nkast:
What's the status of this PR? 

Every time I generate the projects from `MonoGame.Framework.definition` I have to retarget Windows8 to Windows8.1. Should this be fix by this PR? 
Windows8.0 is no longer supported by MG, isn't that right?

nkast:
@hach-que 
Protobuild doesn't seem to support Windows81.
How can we force the &gt;TargetPlatformVersion&lt;8.1&gt;/TargetPlatformVersion&lt; tag in the `MonoGame.Framework.Windows8.csproj` ?

hach-que:
Is this different to the `WindowsUniversal` platform?

nkast:
Yes, it has to be similar to WindowsPhone81

https://github.com/Protobuild/Protobuild/blob/1d096deb50efeb8138d5a251f75a19620faca55c/Protobuild.Internal/BuildResources/GenerateProject.CSharp.xslt#L123-L132

hach-que:
My suggestion would be to submit a pull request to add support for Windows81.  It sounds a bit weird to target Windows 8.1 specifically though (and not Windows 10 or Windows 8).

tomspilman:
We no longer support WinRT for 8.0/8.1 and https://github.com/MonoGame/MonoGame/issues/6207.

PR #4094, opened 2015-08-19
Add credits to the original author of OggStreamer

cra0zy:
Just to mention, it was modified from the original version, don't know if it's important.

### DIFF
--- a/MonoGame.Framework/Utilities/OggStream.cs
+++ b/MonoGame.Framework/Utilities/OggStream.cs
@@ -1,4 +1,12 @@
-﻿using System;
+﻿// This code originated from:
+//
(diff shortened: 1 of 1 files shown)


tomspilman:
> it was modified from the original version, don't know if it's important.

Yeah...  I don't know how that works with licensing.  Like there is code in MonoGame than originally came from SilverSprite and Mono.XNA, but it has been refactored and fixed so many times that it resembles nothing of the original.  So is it still the original code?  Does that license apply?

For me... I always like to attribute the original code regardless of how much we modified it.  Then our modifications of that code fall under MonoGame's license.  This is why there are some files that have both the MonoGame license header and the originals.

But i'm no lawyer...  i'm just doing what seems fair and right.

Issue #4093, opened 2015-08-19
XAudio crashes if a SoundEffectInstance is playing while disposing the SoundEffect


mrhelmut:
I noticed that XAudio (and possible other implementations) would crash if a ```SoundEffectInstance``` is playing while its parent ```SoundEffect``` is being disposed or unloaded via ```ContentManager.Unload()```.

I have a temp fix (see https://github.com/mrhelmut/MonoGame/tree/SoundEffectDisposeFix) that stops any playing/paused pooled ```SoundEffectInstance``` before the disposal of their ```SoundEffect```, and returns them to the pool. But that doesn't concern non-pooled instances.

What do you guys think about this? ( @tomspilman @KonajuGames ?) Should all pooled and non-pooled instances of a given ```SoundEffect``` be stopped? Or do one consider that non-pooled instances (via ```SoundEffect.CreateInstance```) life-cycle is up to the developer?

Let me know what you think and if I should push that fix or rework it.

MichaelDePiazzi:
@mrhelmut Is this issue resolved now that your PR has been merged?

mrhelmut:
Yes! Closing!

PR #4095, opened 2015-08-22
Fix for XAudio crashing if any instance is playing while parent is disposed

mrhelmut:
This is a fix proposal for #4093

If a ```SoundEffect``` is disposed while a ```SoundEffectInstance``` of that effect is playing, it results in an XAudio crash (and possibly other audio implementations).

This fix stops all playing instances upon an effect disposal (and return them to the pool).

It doesn't concern ```SoundEffectInstance``` created through ```SoundEffect.CreateInstance()```, assuming that these instances life-cycle management are up to the developer. (How does XNA behave regarding these? I don't remember having this kind of issue back then.)

If this fix gets merged, ```SoundEffectInstancePool.Update()``` could be a bit cleaned, since the check of disposed effects would no more be useful.

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffectInstancePool.cs
+++ b/MonoGame.Framework/Audio/SoundEffectInstancePool.cs
@@ -154,6 +154,28 @@ namespace Microsoft.Xna.Framework.Audio
             }
         }
 
+        /// <summary>
+        /// Iterates the list of playing instances, stop them and return them to the pool if they are instances of the given SoundEffect.
+        /// </summary>
+        /// <param name="effect">The SoundEffect</param>
+        internal static void StopPooledInstances(SoundEffect effect)
+        {
+            SoundEffectInstance inst = null;
+
+            for (var x = 0; x < _playingInstances.Count; x++)
+            {
+                inst = _playingInstances[x];
+                if (inst.State != SoundState.Stopped && inst._effect == effect)
+                {
+                    inst.Stop();
+                    Add(inst);
+                    continue;
+                }
+
+                x++;
+            }
+        }
+
         internal static void UpdateMasterVolume()
         {
             foreach (var inst in _playingInstances)
--- a/MonoGame.Framework/Audio/SoundEffectInstancePool.cs
+++ b/MonoGame.Framework/Audio/SoundEffectInstancePool.cs
@@ -140,15 +140,6 @@ namespace Microsoft.Xna.Framework.Audio
                     Add(inst);
                     continue;
                 }
-                else if (inst._effect.IsDisposed)
-                {
-                    Add(inst);
-                    // Instances created through SoundEffect.CreateInstance need to be disposed when
-                    // their owner SoundEffect is disposed.
-                    if (!inst._isPooled)
-                        inst.Dispose();
-                    continue;
-                }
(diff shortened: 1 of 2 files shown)


tomspilman:
We currently have this code in there:

https://github.com/mono/MonoGame/blob/4208ef8686733b4dc7cf2e90f58e3cc56e31a23b/MonoGame.Framework/Audio/SoundEffectInstancePool.cs#L143

Which I guess is an attempt to stop an instance if its effect is disposed... both pooled and non-pooled.  Not sure how successful that code would be really.

Have we tested XNA behavior on this?

mrhelmut:
It wasn't successful because the Update method is always called after the disposal, hence instances are not stopped before (not preventing crashes to happen). The clean up I mentioned was about removing this code, because it would never be triggered if we pool back disposed instances separately.

I don't know how XNA would behave regarding playing instances when effects are disposed, but I never had any crash with similar code paths, so my wild guess would be that it cleans when disposed.

tomspilman:
> I never had any crash with similar code paths

More specifically...  the question is does the sound effect instance stop playing if the SoundEffect is disposed?  Because the other path... where the sound effect finishes playing before it is released would also be possible.

mrhelmut:
I see, I'll test if the sound keep playing in XNA and come back here.

mrhelmut:
I've made a simple test: playing a 5sec sound through ```SoundEffect.Play()``` and called ```ContentManager.Unload()``` 2sec after the sound started.

XNA stops the sound, MonoGame (XAudio) crashes.

After heavy testing of the proposed fix, I noticed that XAudio still crashes on very rare occasions. I think this is due to some async process in XAudio which makes instance stopping to not being executed before the disposal. So I still have to track that down before it is 100% efficient.

mrhelmut:
@tomspilman May I have your thoughts?

mrhelmut:
Sorry for the poke, would you mind reviewing this PR ? @tomspilman @KonajuGames 
*shy face*

MonoGame.Framework/Audio/SoundEffectInstancePool.cs
```diff
@@ -140,13 +140,26 @@ internal static void Update()
...
+
+        /// <summary>
+        /// Iterates the list of playing instances, stop them and return them to the pool if they are instances of the given SoundEffect.
+        /// </summary>
+        /// <param name="effect">The SoundEffect</param>
+        internal static void StopPooledInstances(SoundEffect effect)
+        {
+            SoundEffectInstance inst = null;
+
+            for (var x = 0; x < _playingInstances.Count; x++)
```
KonajuGames: This won't iterate correctly.  The x++ here along with the x++ at the end of the loop statement will skip every second playing instance.  If an instance is found and removed via the Add(inst) call, the next two instances will be skipped.  Remove the x++ from the for statement and it will work correctly.

mrhelmut: Woops! That was very unfortunate. PR updated.

mrhelmut:
PR updated. :) @KonajuGames

KonajuGames:
Merging this now.  Thanks @mrhelmut

PR #4098, opened 2015-08-24
Added DrawIndexedPrimitives overload

Shqrdx:
I marked this method as obsolete, hints are not used either in opengl and dx11. I added new overload which does not accept hints.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -808,40 +808,15 @@ namespace Microsoft.Xna.Framework.Graphics
         /// </summary>
         /// <param name="primitiveType">The type of primitives in the index buffer.</param>
         /// <param name="baseVertex">Used to offset the vertex range indexed from the vertex buffer.</param>
-        /// <param name="unused">This is unused and remains here only for XNA API compatibility.</param>
-        /// <param name="unused2">This is unused and remains here only for XNA API compatibility.</param>
+        /// <param name="minVertexIndex">This is unused and remains here only for XNA API compatibility.</param>
+        /// <param name="numVertices">This is unused and remains here only for XNA API compatibility.</param>
         /// <param name="startIndex">The index within the index buffer to start drawing from.</param>
         /// <param name="primitiveCount">The number of primitives to render from the index buffer.</param>
         /// <remarks>Note that minVertexIndex and numVertices are unused in MonoGame and will be ignored.</remarks>
         [Obsolete("Use DrawIndexedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount) instead. In future versions this method can be removed.")]
-        public void DrawIndexedPrimitives(PrimitiveType primitiveType, int baseVertex, int unused, int unused2, int startIndex, int primitiveCount)
+        public void DrawIndexedPrimitives(PrimitiveType primitiveType, int baseVertex, int minVertexIndex, int numVertices, int startIndex, int primitiveCount)
         {
-            if (_vertexShader == null)
-                throw new InvalidOperationException("Vertex shader must be set before calling DrawIndexedPrimitives.");
-
-            if (_vertexBuffer == null)
-                throw new InvalidOperationException("Vertex buffer must be set before calling DrawIndexedPrimitives.");
-
-            if (_indexBuffer == null)
-                throw new InvalidOperationException("Index buffer must be set before calling DrawIndexedPrimitives.");
-
-            if (primitiveCount <= 0)
-                throw new ArgumentOutOfRangeException("primitiveCount");
-
-            // NOTE: minVertexIndex and numVertices are only hints of the
-            // range of vertex data which will be indexed.
-            //
-            // They will only be used if the graphics API can use
-            // this range hint to optimize rendering.
-
-           
-            unchecked
-            {
-                _graphicsMetrics._drawCount++;
-                _graphicsMetrics._primitiveCount += (ulong) primitiveCount;
-            }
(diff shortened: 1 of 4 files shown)


Shqrdx:
@tomspilman your thoughts ?

Shqrdx:
@tomspilman @KonajuGames I still waiting your decision..

Test/Framework/Visual/GraphicsDeviceTest.cs
```diff
@@ -111,6 +111,66 @@ public void DrawIndexedPrimitivesParameterValidation()
...
+
+                // XNA doesn't do upfront parameter validation on the Assert.DoesNotThrow tests,
+                // but it *sometimes* fails later with an AccessViolationException, so we can't actually
+                // run these tests as part of the XNA test suite.
+
+                // baseVertex too small / large.
+#if !XNA
+                Assert.DoesNotThrow(() => Game.GraphicsDevice.DrawIndexedPrimitives(PrimitiveType.TriangleList, -1, 0, 1));
+                Assert.DoesNotThrow(() => Game.GraphicsDevice.DrawIndexedPrimitives(PrimitiveType.TriangleList, 3, 0, 1));
+#endif
```
tomspilman: Does XNA throw exceptions in these cases?  Why are we omitting XNA in this test?

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -843,6 +844,29 @@ public void DrawIndexedPrimitives(PrimitiveType primitiveType, int baseVertex, i
...
+                throw new ArgumentOutOfRangeException("primitiveCount");
+
+            unchecked
+            {
+                _graphicsMetrics._drawCount++;
+                _graphicsMetrics._primitiveCount += (ulong)primitiveCount;
+            }
+
+            PlatformDrawIndexedPrimitives(primitiveType, baseVertex, startIndex, primitiveCount);
+        }
```
tomspilman: I'm fine with adding a new overload which removes the confusing and unused parameter.

We should go a step further and change the original method and rename `minVertexIndex` to `unused` with the documentation description of  `This is unused and remains here only for XNA API compatibility`.

We will need to keep the original method as long as we want to maintain XNA API compatibility, but internally the code should just forward the call to the new method.

Still lets see what @KonajuGames thinks before proceeding on any of this.

Test/Framework/Visual/GraphicsDeviceTest.cs
Shqrdx: I dont know, I just copy/paste old test to new.

MonoGame.Framework/Graphics/GraphicsDevice.cs
Shqrdx: numVertices is unused too.. so I change it to unused2

KonajuGames: The `minVertexIndex` and `numVertices` parameters were used in XNA, but were ignored by Nvidia drivers and required in some cases by ATI drivers.  This was the cause of some primitives not drawing on ATI cards only.

Test/Framework/Visual/GraphicsDeviceTest.cs
Shqrdx: Looks like unused check. However this test will fail if PlatformDrawIndexedPrimitives throws exception. Thats the only reason to keep it up I guess.

MonoGame.Framework/Graphics/GraphicsDevice.cs
KonajuGames: I also wouldn't change the parameter name.  Just change the documentation to mention that it is not used by MonoGame and will be ignored.

Shqrdx:
How about change method body to 
<b>
public void DrawIndexedPrimitives(PrimitiveType primitiveType, int baseVertex, int minVertexIndex, int numVertices, int startIndex, int primitiveCount)
{
   DrawIndexedPrimitives(primitiveType, baseVertex, startIndex, primitiveCount);
}
</b>

MonoGame.Framework/Graphics/GraphicsDevice.cs
tomspilman: > I also wouldn't change the parameter name.

I guess I suggested to change the parameter name so that no one gets sucked into trying to figure out what their `minVertexIndex` is before finally noticing it is unused.

I'm fine doing it either way.

tomspilman:
> How about change method body to 

That is fine with me.

KonajuGames:
> How about change method body to 

Make that small change and it should be good to go.

Shqrdx:
This is fine and ready for merge ?

KonajuGames:
Yeah. I'm happy with that now.  Merging.  Thanks @Shqrdx

Test/Framework/Visual/GraphicsDeviceTest.cs
tgjones: The reason those lines are excluded from the XNA test is detailed in the comment immediately above:

``` csharp
// XNA doesn't do upfront parameter validation on the Assert.DoesNotThrow tests,
// but it *sometimes* fails later with an AccessViolationException, so we can't actually
// run these tests as part of the XNA test suite.
```

KakCAT:
Hi!

Is it correct to mark this function as obsolete? I mean that afaik this function is the only way to DrawIndexedPrimitives in XNA, and all the existing XNA code ported to MonoGame will generate this obsolete warning.

I have to go back and forth XNA/MonoGame code, so if I want to get rid of the warning, I have to use an #if MONOGAME , which defeats the purpose of backwards compatibility.

PR #4102, opened 2015-08-26
Fix for back button state being ignored if no gamepad connected on UWP

mrhelmut:
This PR fixes the issue raised by #4077 regarding back button states when no gamepad is connected with an UWP.

### DIFF
--- a/MonoGame.Framework/Input/GamePad.UWP.cs
+++ b/MonoGame.Framework/Input/GamePad.UWP.cs
@@ -58,8 +58,8 @@ namespace Microsoft.Xna.Framework.Input
 
         private static GamePadState PlatformGetState(int index, GamePadDeadZone deadZoneMode)
         {
-            if (index >= WGI.Gamepad.Gamepads.Count && index == 0)
-                return GetDefaultState();
+            if (index >= WGI.Gamepad.Gamepads.Count)
+                return (index == 0 ? GetDefaultState() : GamePadState.Default);
 
             var state = WGI.Gamepad.Gamepads[index].GetCurrentReading();
 
--- a/MonoGame.Framework/Input/GamePad.UWP.cs
+++ b/MonoGame.Framework/Input/GamePad.UWP.cs
@@ -49,10 +49,17 @@ namespace Microsoft.Xna.Framework.Input
             };
         }
 
+        private static GamePadState GetDefaultState()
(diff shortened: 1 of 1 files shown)


cra0zy:
I see one problem in this, if the back button is pressed on device, it will act as if all 4 gamepads pressed it. You may want to add if(index == 0), don't know how XNA did it...

Anyway I think that mobile back button shouldn't be in gamepad at all, it's not a gamepad and it just makes the confusion if the gamepad had back button pressed or the device had back button pressed. I think that adding the back button to Keyboard would make a lot more sense considering that it's one per device and phones can have buttons like search and menu which could also go there,

@tomspilman @KonajuGames How would you feel about moving the back button detection to Keyboard (considering it would differ from XNA)?

tomspilman:
> How would you feel about moving the back button detection to Keyboard 

I don't think it belongs in the keyboard either.  To me this is something that should be separate all together specifically for phones with dedicated programmable buttons like WP has with back.

cra0zy:
> I don't think it belongs in the keyboard either. To me this is something that should be separate all together specifically for phones with dedicated programmable buttons like WP has with back.

Yea, somewhere else is fine, I just didn't like it in Gamepad, would make programming stuff with Gamepad for UWP a bit weird. Just a small question, what do you mean by "dedicated programmable buttons like WP has with back"? 


Side note, the reason I thought it belonged in Keyboard is because all the buttons besides the Back button(Home, Search, Menu...) can also occur on Keyboard.

mrhelmut:
I agree that it doesn't quite belong to GamePad.
Back in the XNA days, the Back button on Windows Phone was out of its scope.

What about TouchPanel / TouchPanelState?

KakCAT:
Hi,

moving away from pads would break XNA WP7 compatibility, as far as I know the standard way to check for the back button in WP7 XNA is cheking the first player pad. 

Maybe the pad should be kept but you could introduce an alternate way to check for it which is more meaningful.

mrhelmut:
I've update the PR to limit that behavior only to PlayerIndex.One.

Should this be merged to keep XNA compat'?

If so, I'll open an issue right after to keep track of this and continue this discussion.

KonajuGames:
I'm going to merge this as it does keep XNA compatibility.

PR #4103, opened 2015-08-26
Don't use alcMacOSXMixerOutputRate on iOS as it is not included in iOS9.

### LINKED ISSUE #4076: alcMacOSXMixerOutputRate issue on iOS9
ctor of OpenALSoundController crashes on iOS9 device/simulator with 
> System.EntryPointNotFoundException: alcMacOSXMixerOutputRate

Here is the self-contained example:
https://github.com/xamarin/monotouch-samples/tree/master/BouncingGameCompleteiOS

Looks like there is no symbol for *alcMacOSXMixerOutputRate* on iOS9. Sample works fine on iOS8.

Also have a look at Apple's sample
https://developer.apple.com/library/mac/samplecode/OpenALExample/Listings/Scene_m.html#//apple_ref/doc/uid/DTS10003587-Scene_m-DontLinkElementID_10

```
typedef ALvoid  AL_APIENTRY (*alcMacOSXMixerOutputRateProcPtr) (const ALdouble value);
ALvoid  alcMacOSXMixerOutputRateProc(const ALdouble value)
{
    static  alcMacOSXMixerOutputRateProcPtr proc = NULL;
    
    if (proc == NULL) {
        proc = (alcMacOSXMixerOutputRateProcPtr) alcGetProcAddress(NULL, (const ALCchar*) "alcMacOSXMixerOutputRate");
    }
    
    if (proc)
        proc(value);
 
    return;
}
```

I think that we have to check existance of *alcMacOSXMixerOutputRate* symbol first before invoke alcMacOSXMixerOutputRate function

I believe that this is a crasher https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs#L95

danzel:
Fixes #4076

Not sure if we should merge this yet as iOS9 hasn't hit gold master and the symbol could potentially come back.

### DIFF
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -91,7 +91,7 @@ namespace Microsoft.Xna.Framework.Audio
         /// <returns>True if the sound controller was setup, and false if not.</returns>
         private bool OpenSoundController()
         {
-#if MONOMAC || IOS
+#if MONOMAC
 			alcMacOSXMixerOutputRate(PREFERRED_MIX_RATE);
 #endif
 


KonajuGames:
Merged.  Thanks @danzel.

PR #4097, opened 2015-08-23
Reset ScreenReceiver.ScreenLocked in OnDestroy()

hig-ag:
PROBLEM:
When launching an AndroidGameActivity from a notification the static field ScreenReceiver.ScreenLocked stays on true.
The problem is that then the game won't be initalized because of this check in AndroidGameWindow.OnUpdateFrame:

                if (!GameView.IsResuming && _game.Platform.IsActive && !ScreenReceiver.ScreenLocked)
                {
                    _game.Tick();
                } 

SOLUTION:
Reset ScreenReceiver.ScreenLocked in OnDestroy()

PR #4104, opened 2015-08-26
Fix re-launching on Android from a notification

danzel:
When we are launched from a notification ScreenLocked would still be true so things wouldn't work. Reset it early instead.

See #4097 for details. This is just a whitespace correct commit. (Actual change by @hig-ag)


tomspilman:
@KonajuGames  @nezz @dellis1972 ?

KonajuGames:
Merging. Thanks @danzel and @hig-ag.

PR #4108, opened 2015-08-29
Fix multitouch lockup on WP8.1

### LINKED ISSUE #3316: WP8.1 Multitouch slowdown
Hi! I'm trying to port my games on WP8.1 with Monogame 3.2.3 (via NuGet -pre version).

But I notice in one of my game (which is based on multitouch gameplay 4+ touches at once) that if I touch screen with 3 fingers and move them I get fps drop and if I touch it and move with 7+ finger it's drop even more and can even freeze until I go back in WP menu and back again. Even if I have no touch code at all. 

It can be reproduce in empty WP8.1 project with monogame template (with 3d cube in it), touch it with 5-10 fingers and frame will freeze. Here is the video: http://youtu.be/gSrUaaTeFFk

danzel:
Use a RunLoop thread rather than listening for ```CompositionTarget.Rendering```.
This seems to fix the lockup during touch on WP8.1.

I needed to add a lock around the orientation event otherwise we could try change orientation while drawing and we'd crash (because the run loop runs at the same time).
I don't think anywhere else needs the same lock, but I may of overlooked something.

I haven't changed the input code, so we still get 60 touches a second for every touch that is occurring. This doesn't seem to cause troubles for me, but may on bigger projects (I'm testing on a lumia 520 though, so it is pretty slow)

This will need testing on other Metro platforms too (Win8 Store app), I'll try have a crack later today, I don't expect any troubles though.

Fixes #3316

### DIFF
--- a/MonoGame.Framework/Windows8/MetroGamePlatform.cs
+++ b/MonoGame.Framework/Windows8/MetroGamePlatform.cs
@@ -12,6 +12,8 @@ using System.Diagnostics;
 using Microsoft.Xna.Framework.Graphics;
 using Microsoft.Xna.Framework.Input.Touch;
 using Windows.ApplicationModel.Activation;
+using Windows.Foundation;
+using Windows.System.Threading;
 using Windows.UI.ViewManagement;
 using Windows.UI.Xaml.Media;
 #if WINDOWS_PHONE81
@@ -115,10 +117,16 @@ namespace Microsoft.Xna.Framework
 
         public override void StartRunLoop()
         {
-            CompositionTarget.Rendering += (o, a) =>
+            ThreadPool.RunAsync(delegate(IAsyncAction action)
             {
-                MetroGameWindow.Instance.Tick();
-            };
+                while (action.Status == AsyncStatus.Started)
+                {
+                    lock (MetroGameWindow.Instance.RunLock)
+                    {
+                        MetroGameWindow.Instance.Tick();
+                    }
+                }
+            }, WorkItemPriority.High, WorkItemOptions.TimeSliced);
         }
         
         public override void Exit()
--- a/MonoGame.Framework/Windows8/MetroGameWindow.cs
+++ b/MonoGame.Framework/Windows8/MetroGameWindow.cs
@@ -40,6 +40,11 @@ namespace Microsoft.Xna.Framework
 
         internal bool IsExiting { get; set; }
 
+        /// <summary>
+        /// Used to stop orientation change events and the run loop from running at the same time
+        /// </summary>
+        internal readonly object RunLock = new object();
+
         #endregion
 
         #region Public Properties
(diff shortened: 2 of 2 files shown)


tomspilman:
> This seems to fix the lockup during touch on WP8.1.

I'm pretty hesitant to change how the entire game ticks to resolve a touch issue without some real strong evidence that we've been rendering incorrectly all this time.

We've been using `CompositionTarget.Rendering` since 2012 when WinRT first appeared.  It is how all the WinRT DirectX samples show we should be doing game updates.

These DX samples generally have touch support.  Have should examine those samples more closely to see if we can repo the lockup with them and if not what they are doing different.

danzel:
Fair enough, I'll look in to it. Where are the samples?

jakepoz:
Will test this shortly on our end regardless, just to see if it's fixed.

jakepoz:
Just tested a production game with the changes, sadly, it gets stuck loading indefinitely.

danzel:
Well that is unexpected!
Oh well thanks for testing. Will close this and check out the DX samples.

KakCAT:
Hi,

I know this is closed and have several extra problems, but the commit also fixes the rotation bug huge delay at startup on both WP8.1 and WP10  (I obviously changed MetroGamePlatform for UAPGamePlatform in WP10)

jakepoz:
Nice catch KakCAT, I will dig into the loading issue and make a bare bones project to reproduce it.

danzel:
Awesome thanks

jakepoz:
Hi Danzel, I was able to get the bug to happen in the debugger. With your patch, the first time the game tries to play background music using the MediaPlayer, a TypeInitializationException gets thrown. 

The offending line is the CoreWindow.GetForCurrentThread() call in the PlatformInitialize method in MediaPlayer.WME.cs:

```csharp

        private static void PlatformInitialize()
        {
            MediaManager.Startup(true);
            using (var factory = new MediaEngineClassFactory())
            using (var attributes = new MediaEngineAttributes { AudioCategory = AudioStreamCategory.GameMedia })
            {
                var creationFlags = MediaEngineCreateFlags.AudioOnly;

                var mediaEngine = new MediaEngine(factory, attributes, creationFlags, MediaEngineExOnPlaybackEvent);
                _mediaEngineEx = mediaEngine.QueryInterface<MediaEngineEx>();
            }

            _dispatcher = CoreWindow.GetForCurrentThread().Dispatcher;
        }
```

After that exception, the rest of the game just hangs.

danzel:
Yep that makes sense, we aren't running on the main thread any more so dispatcher will be wrong.

Try replace [this line](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Media/MediaPlayer.WME.cs#L31) with:
```
_dispatcher = CoreApplication.MainView.CoreWindow.Dispatcher;
```

Or if that doesn't work, try this as a hack.
On this line (before the ThreadPool call),
https://github.com/danzel/MonoGame/blob/4d7a1044a1f7b3e8461a19b6f07409dcf720abb1/MonoGame.Framework/Windows8/MetroGamePlatform.cs#L120
Add ```MediaPlayer.Pause();```

Will probably get us past that error and over to another :)

(I'm not at my dev PC atm so can't test these)

jakepoz:
I just temporarily removed that dispatcher completely, and I didn't see any other Monogame issues. However, a few places in our app that interact with XAML controls need to be modified to run from the core window dispatcher.

danzel:
Cool.

I've tried to run the game loop on the main thread by passing ```Synchronous``` [here](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Windows8/XamlGame.cs#L56)
So we use this run loop:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Windows8/MetroGameWindow.cs#L258

But nothing appeared on screen, I'm not sure why. (Haven't spent much time on it)

jakepoz:
It wouldn't be the worst to have the manually dispatch XAML related stuff to the proper thread. If you look at the newer Keyboard Input stuff, Nezz just grabs the dispatcher as you'd expect:

https://github.com/mono/MonoGame/blob/57936b1a66fe3257bce37c101a6d509f3bc5180e/MonoGame.Framework/Input/KeyboardInput.WinRT.cs

KakCAT:
Hi,

My WP8.1 finally has been returned to me. Is it possible @tomspilman to post a link to the samples using CompositionTarget.Rendering to test them and check if they're also having problems with the touch and move forward this issue?

All the examples I've found are using a loop instead of CompositionTarget.Rendering.

Thanks!

KakCAT:
I don't know if this is proof or not. What's certain it's that it's very very weird: #4129

PR #4109, opened 2015-08-31
Fix for NullReferenceException when trying to switch Song rapidly

mrhelmut:
Due to the async volume controller initialization in XAudio, it may happen that switching rapidly songs can trigger a null ref exception here (because the volume has not been init yet): https://github.com/mono/MonoGame/blob/03185dff502df8532600fadbe8044b3d3c978d2d/MonoGame.Framework/Media/MediaPlayer.WMS.cs#L155

This PR adds null checks to MediaPlayer and VideoPlayer to cope with the async init.


tomspilman:
Makes sense to me... merging!

PR #4110, opened 2015-08-31
IOS: reduce freezes on song looping

suilevap:
Do not call stop() when only one song is in a queue and IsRepeating is true (to avoid noticeable freezes on iOS).

### DIFF
--- a/MonoGame.Framework/Media/MediaPlayer.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.cs
@@ -215,8 +215,10 @@ namespace Microsoft.Xna.Framework.Media
 		
 		private static void NextSong(int direction)
 		{
-            Stop();
-
+            if (!IsRepeating || _queue.Count > 1)
+            {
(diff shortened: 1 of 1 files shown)


danzel:
Did you test this on other platforms?
The build error doesn't look like your fault, and this commit sounds useful if it works like it says :)

mrhelmut:
Should work fine on XAudio platforms, Songs get cleaned/stopped anyway upon new plays.

suilevap:
After some additional testing on IOS, we found that avoiding to call redundant Stop() it is not enough to avoid freezes when music loops. We tried to change music Play to wrap Song.Play to Task (specific for ios song implementation). It works better atleast in our case, but still sometimes we see small freezes when songs repeats.
Adittional difficulties that we do not know an easy way to reproduce this ussue. 
Our steps: 
play song#1 with Repeating several times - good
play song#2 with Repeating several times - soimetimes ok, sometimes freezes

PR #4114, opened 2015-09-01
Allow using all of the available SoundEffectInstancePool Instances

danzel:
We need to check there are any remaining before using ourself, not after, otherwise we count ourself against the limit.

ATM we throw the InstancePlayLimitException when calling SoundEffect.Play to play the maximum amount of instances.

Broken since e8d2069da115e7be2d11374c399a945bb28d252e


KonajuGames:
Merging.  Thanks @danzel.

PR #4122, opened 2015-09-03
Fix copy-paste error in network statistics.

Therzok:


tomspilman:
Thanks @Therzok !

Therzok:
No problem! Coverity found this in the submodule of MonoGame in CocosSharp. There are a few more issues I want to solve, so I think it would be best to share the list with you.

There would be a few possible solutions:
 * Setup coverity for the MonoGame repository (I can provide scripts to set it up.)
 * Just grab the results from the CocosSharp coverity scans (But things would get messy.)

tomspilman:
>  I think it would be best to share the list with you.

Sure...  if they are like the ones so far it is fine to just submit PRs.  You can a couple of fixes at a time in the same PR as long as none of them are crazy complex or something you suspect might be controversial.

> Setup coverity for the MonoGame repository

What would the scripts do?  Like fail the automated build if something is found?

Therzok:
The scripts would submit the build analysis results to Coverity and the scanner would do its work, kicking off emails to the people subscribed about new/solved defects. It won't affect the build in any way, just refresh the data available on scan.coverity.com 's dashboard.

There is a problem with number of builds submittable per week, as in, it's based on source code size. Bigger projects get a smaller analysis quota. Therefore the CI would have to run on a timed-ish basis.

MichaelDePiazzi:
@Therzok Looks like this same bug also still exists in the original Lidgren.Network library (https://github.com/lidgren/lidgren-network-gen3) and I'm sure they'd also appreciate a PR for this fix. Would you mind submitting a PR for this to Lidgren.Network also? Otherwise, I'd be happy to do this on your behalf.

Therzok:
@MichaelDePiazzi Whatever works for you, sure! I don't mind doing the fixes on the upstream lidgren, then backporting them here in monogame, or vice-versa.

I'll hack on a bigger PR with small commits on Saturday, the latest. Thanks for the tip!

MichaelDePiazzi:
@Therzok Cool man, I'll leave it to you then :)

tomspilman:
@Therzok - I did this...

https://scan.coverity.com/projects/mono-monogame

... I suspect that is the first step in supporting it.

It would be easy to add another task into the TeamCity server to upload some binaries to their servers once every few days or so.

Therzok:
@tomspilman Here's the script:
 * It requires coverity_email and coverity_token environment variables.
 * It requires chocolatey for easy installation of packages.
 * The machine needs to have [win32](https://scan.coverity.com/download/csharp/win-32) or [win64](https://scan.coverity.com/download/csharp/win_64) put somewhere in the PATH. You can't download them without authentication. So possibly `wget` them with authentication parameters.

```powershell
cinst curl -y
cinst 7zip.commandline -y

# Possibly do ProtoBuild, depending whether the CI does or not do it automatically.
.\Protobuild.exe

# TODO: Choose which target (config/platform) gets built, so sln name needs corrected and /p: added accordingly.
& cov-build.exe --dir cov-int "C:\Program Files (x86)\MSBuild\12.0\Bin\MSBuild.exe" /m /t:Build MonoGame.sln

$version = git rev-parse HEAD

7z a "monogame.zip" "cov-int"

# This is so we can use basic curl to post data to coverity.
Remove-Item alias:curl

curl --silent --show-error `
	--output "curl-out-$(get-date -f yyyy-MM-dd).txt" `
	--form email=$Env:coverity_email `
	--form token=$Env:coverity_token `
	--form "file=@monogame.zip" `
	--form version=$version `
	--form description="Analysis submission from $(get-date -f yyyy-MM-dd)" `
	https://scan.coverity.com/builds?project=mono%2FMonoGame
```

Therzok:
@MichaelDePiazzi I opened a PR there and added a few more issues someone with more experience and in touch with the codebase should fix: https://github.com/lidgren/lidgren-network-gen3/pull/35

Therzok:
@tomspilman can you give me higher rights on the coverity dashboard? I'll set up the initial things and push an initial scan so you can look at the defects.

PR #4123, opened 2015-09-03
Fix premature access on possibly null variable

Therzok:


KonajuGames:
That makes sense to me.  No point checking if `dictionary` is null after it has tried to access it.

Merging.  Thanks @Therzok

Issue #3405, opened 2015-01-15
MonoGame.Portable returns - now with Piranah's


DDReaper:
Right, so I've been working with @dellis1972 on the PCL generator project and it looks like it is close to being done.
It now uses a PCL generator solution based on Mono.Cecil called Piranah (which Dean has also updated to run on a Mac)
Generator branch for MG can be found here: https://github.com/DDReaper/MonoGame/tree/PiranahPCLGenerator

The generator only uses the WindowsGL build as a template, which seems sufficient for the PCL to be used in a "Bait and Switch" style PCL as a core platform.  Platform specific code should always remain in a Platform project (like iOS), where as all common "core" code should be used in the PCL.

Question is now how to provide this framework back in to the MonoGame project so it can be run in an automated fashion.  In the branch above (using Protobuild as a template) I've created a new ThirdParty folder for Piranah and put the runtime exe and Dll's in there.
There is a commandline file to generate the project manually for test at the moment (need both the WindowsGL project built and an output folder in MonoGame.Framework/bin/PCL created first)

I've also started on a new MonoGame.Portable NuGet package, which is still a work in progress.  In testing it needs more platform targets to work effectively.

So next steps?

Nezz:
What's the point of having a PCL version of MonoGame? Is it a stripped-down version that allows you to turn a subproject into a PCL even if it uses something like Vector2?

KonajuGames:
Used for the bait-and-switch technique where you use NuGet to add the PCL
version to your project and NuGet will substitute the platform-specific
version at build time.​  Useful for cross-platform solutions.

tomspilman:
> What's the point of having a PCL version of MonoGame?

I still feel like this personally. Even bait-and-switch requires the same or more csprojs to make it work.

But some people still like it which is why we are going to support it.

DDReaper:
@Nezz PCL's solve two distinct problems with multi-platform dev.

1: No file linking / copying required, it's a single project.  While both file linking / copying work fine, they do require maintenance to add new files or remove files from all projects over time.

2: Core compatibility.  Any code that is in the PCL is guaranteed to be compatible with all platforms, no need for #IF's or other code to manage all the different platforms within the PCL.  It will also restrict you from using code that will break some platforms, plus it also abstracts some basic features to ensure code is compiled specifically for a certain platform.

Further details here: http://darkgenesis.zenithmoon.com/monogame-building-portable-solutions/

[...]

Issue #3194, opened 2014-11-21
Proposal: Protobuild packages and templates for MonoGame


hach-que:
I've been working on package management functionality in Protobuild for the last month or so, and after some previous discussion, I've been developing it further to better suit MonoGame.

Protobuild's package management is designed to support both cross-platform libraries and cross-platform project templates.  For libraries, the binary packages it supports are LZMA compressed, deduplicated and can be swapped in-place for source code (and vice-versa).  For project templates, it provides an easier mechanism to start a Protobuild-based project using cross-platform libraries.

As a demonstration, [download the latest version of Protobuild](https://github.com/hach-que/Protobuild/raw/master/Protobuild.exe) into an empty directory, open a command prompt, and run:

```
Protobuild.exe --start http://protobuild.org/MonoGame/Template.2DPlatformer@develop MyNewPlatformer
```

This works on all desktop platforms (although Mac OS is untested).  For this demonstration you'll need Git in your path because there's no binary package present yet for the project template itself, and if you're running this on Mac OS or Linux, prefix the command with `mono`.  When the command completes, you should be able to open the solution in Visual Studio or Xamarin Studio and run the game.

Protobuild's package management has several advantages over using NuGet:

**Full support for cross-platform projects, in particular those that need to make native calls (such as P/Invoke)**

Protobuild designates library packages by platform; the same platforms that are used for generation, and not framework API.

**Automatic fallback to source code repositories if binary versions are not available for the platform**

Users can swap out binary packages for source versions and vice-versa transparently.  This makes it easier to diagnose issues and bugs.  To try that with the demonstration above, run:

```
Protobuild.exe --swap-to-source http://protobuild.org/MonoGame/MonoGame
Protobuild.exe --generate
```

or try the reverse:

```
Protobuild.exe --swap-to-binary http://protobuild.org/MonoGame/MonoGame
Protobuild.exe --generate
```

**Packages are LZMA compressed and deduplicated to significantly reduce their size**

For reference, the binary packages I've built are:
- Windows + Content Pipeline: 24.4Mb
- Linux + Content Pipeline: 15.7Mb
- WindowsGL: 1.5Mb

The content pipeline includes all external tools and relevant binaries (e.g. ffmpeg + ffprobe for those platforms).

**Support for module templates as packages**

As demonstrated above, this makes it very easy for people to start with MonoGame on any platform (not just Windows).  This requires no installation of MonoGame for developers to start using the engine.

[...]

PR #4128, opened 2015-09-09
Update Protobuild

hach-que:
This updates Protobuild for #3405.


hach-que:
Btw Build/GenerateProject.CSharp.xslt is getting super out-of-date and the Windows 10 support should be merged back into Protobuild sometime soon.

danzel:
I think the git submodules changes might be a bit controversial, I'm not a fan that I don't get a full checkout of the code when I check it out.

KonajuGames:
Some git clients such as TortoiseGit that I use handle submodules automatically.  How would it handle locally modified submodule references?

hach-que:
This doesn't change the behaviour of how Git works. 

Currently, you need to run `git submodule update --init --recursive` manually after running `git checkout` or `git clone` to get a full checkout of the code. This PR doesn't take away getting "a full checkout of the code when I check it out" because that's not a feature that existed to begin with. 

What it does do is mean that you don't need to initialize submodules at all, because Protobuild now does it for you, and since you need to run Protobuild right after checking out anyway, it doesn't change anything to the workflow from a user's perspective other than reducing the number of commands you need to run by 1 (while also drastically decreasing the submodule clone time because it no longer needs to download from the internet every time).

hach-que:
> How would it handle locally modified submodule references?

You mean changes in the submodule folder?  The submodule folder is marked as ignored from the perspective of the parent repository, but if you open the submodule itself in Git, then it's a repository you can commit / push as normal in. You won't get the "stage in parent repo on submodule commit" functionality that some GUIs offer, but Protobuild offers a better alternative to that anyway (see below). I've never found the auto stage to be particularly useful since you need to go back to the parent repo and still commit / push there anyway when using standard Git submodules.

Unlike Git, Protobuild can track submodules by branch name instead of commit hash; this means you don't need to update the consuming repository when you update the submodule. This means that pushing the submodule to the remote is enough to update it.

tomspilman:
I don't like it.  It is trying to solve a problem we don't have.

Note it is likely we'll drop the submodules in favor of Git's large file support in the near future.

hach-que:
It is a problem that consumers of MonoGame have though, even if upstream
doesn't experience it directly.

Git LFS would also make MonoGame not clonable by anyone who doesn't have
git LFS installed, and since it's not part of standard Git, that would have
a pretty detrimental impact on people trying to use it via the git repo.

tomspilman:
> It is a problem that consumers of MonoGame have though

This is why we offer installers, zip downloads of binaries, and Nuget packages.  We have plenty of ways to use MonoGame for people that can't manage Git.

> Git LFS would also make MonoGame not clonable by anyone who doesn't have git LFS installed

This won't be a problem... you install it and you're fixed.

hach-que:
> This is why we offer installers, zip downloads of binaries, and Nuget packages. We have plenty of ways to use MonoGame for people that can't manage Git.

None of these options provide users with source code, so they aren't even slightly comparable. 

> This won't be a problem... you install it and you're fixed.

This I don't understand. You're willing to create an external dependency on Git LFS that _users have to install on every development machine and build server_, but you won't create an external dependency for #3194 which actually solves the same problem without users having to install anything at all...?

tomspilman:
> This I don't understand.

I expect Git LFS to be extremely popular being that it will be a core feature of the GitHub site.  It also removes our need to maintain multiple repos.

hach-que:
But it doesn't negate the fact that it's still an external dependency that users have to install.  For teams that don't use GitHub to host their repositories (which is quite likely given that most developers are probably hosting their Git repositories internally / privately), they probably won't have Git LFS installed.

It also doesn't change the fact that Git LFS is still yet-to-be-supported on GitHub, and has been that way since it was announced since April.  This PR fixes the problems now, and doesn't impact your ability to use Git LFS later when it is actually supported by GitHub.

----

Another aspect to consider is that Git LFS is priced at 1GB per organisation or personal account.  Which means usage of Git LFS here will count towards the mono organisation's quota.  Once that quota is exceeded, users can only download the pointer files (not the actual binary files themselves) and no new files can be pushed until you upgrade your plan.

In addition to this, you get __1GB of bandwidth per month__.  This means every time someone clones the MonoGame repository, it's going to count towards the mono organisation's LFS bandwidth.  Once the bandwidth is exceeded, users can't download binary files again until the next month (or presumably you upgrade your plan).

More pricing information for Git LFS on GitHub is available at https://help.github.com/articles/billing-plans-for-git-large-file-storage/, but I expect that the MonoGame repository will hit the free limits very quickly with public cloning.

hach-que:
Just to check, the Dependencies repository (not including Git history) for files >16MB is 285MB, which means that 5 clones globally by any user will consume the entirety of mono's LFS quota for the month.

DDReaper:
@hach-que any update on this PR to get the version of Protobuild updated?

tomspilman:
The main sticking point here is that we have no plans to use Protobuild to manage submodules.  If this PR was remade without that it would be merged immediately.

hach-que:
Yeah, but there aren't any viable alternates as discussed (LFS isn't free
and normal Git submodules only increase in size), so I'd like to see some
serious consideration given to the topic first (other than "the problem
doesn't exist", because it does and the submodule size does impact users of
MonoGame).

tomspilman:
> Yeah, but there aren't any viable alternates as discussed 

What we are doing right now is perfectly fine...  there has been no issues with submodules.

And if in the future we want to use LFS I think we can manage $5 a month for 50GB of storage/bandwidth.

There is no problem here.

hach-que:
50GB of bandwidth would be 250 clones a month. My build server which tracks MonoGame would burn through your entire quota in just 25 commits. And that's just me. I don't think you're realizing how little 50gb is going to go when you're talking an extremely popular open source game engine. 

Large submodules are an issue for users, even if the upstream developers don't encounter it. Downloading hundreds of megabytes of dependencies every time you want to start a new project sucks.

tomspilman:
> I don't think you're realizing how little 50gb is going to go

Sure... i'm not.  I'm also not proposing to switch to LFS tomorrow or that we can only afford $5.  When we get around to that we'll see what things look like and make a decision.

> Large submodules are an issue for users, even if the upstream developers don't encounter it.

I've not heard any complaints other than from you.  I've worked on the project for 3 years now...  I've probably not made more than a couple dozen clones of the repo in that time.

If you need to make tons of new clones daily... then I suggest you are doing something wrong.

hach-que:
> If you need to make tons of new clones daily... then I suggest you are doing something wrong.

My build server makes clones for every commit.  I don't know how else you expect this to work.

In addition, whenever anyone wants to use Protogame, that also induces a clone of MonoGame.  If they're at a game jam, that means cloning hundreds of megabytes of dependencies over a limited Internet connection, even if they've used MonoGame or Protogame on that machine before.

This change makes it so that submodules will be cloned from the cache if possible.

> I've not heard any complaints other than from you. I've worked on the project for 3 years now... I've probably not made more than a couple dozen clones of the repo in that time.

Well every time I create a new project, or even need to test something, I have to clone the repo fresh if I want a new project using MonoGame, and that's because Git doesn't natively support caching submodules.  I've probably cloned this repo a dozen times in the last 2 months, and many many more times before that (not including the cloning that my build server also does).  I'd estimate I've cloned this repo hundreds of times over the course of using MonoGame, and eventually it gets tiresome to wait for it to clone the same 400MB of dependencies /all over again/.

Again, this change is only a net positive; it doesn't impact workflows or build plans, it doesn't force anyone to do something they don't already have to do (which is run Protobuild to generate projects).  The only thing that changes is that submodules are now cached, reducing clone times.  There is no downside to merging this PR, so I don't understand the hesitation.

KonajuGames:
> My build server makes clones for every commit. I don't know how else you expect this to work.

That's not what I see in usual continuous integration setups.  A single working copy is switched or branched to the commit in question, resulting in minimal transfer from the server to the working copy.

> Well every time I create a new project, or even need to test something, I have to clone the repo fresh if I want a new project using MonoGame, and that's because Git doesn't natively support caching submodules. 

This is more of a result of a design decision in Protogame rather than a limitation of git.  The usual method of MonoGame development is to use the installer, or if you want to run from source, have a single working copy that the new projects reference.  If you want to test something or make changes, branch that working copy.

tomspilman:
> My build server makes clones for every commit. I don't know how else you expect this to work.

Most build systems support doing cleans and updates instead of full checkouts on each build pass.  That is how the MonoGame automated build works and every automated build I've setup since 1999.  This feature exists specifically to avoid having to download everything again when most of it is all the same.  It is critical to game development when you often have GB of data in your repo... not just MB.


> it doesn't impact workflows or build plans

It is the exact opposite of what you suggest...  it changes *everyone's* workflow.  The way you clone is different... need to run protobuild to get your submodules.  Upstream changed the submodule reference... you need to pull then run Protobuild to fix that.  How you modify the submodule reference is now different... fix it in the protobuild script.  Even branching is affected requiring you run Protobuild to fix the submodule reference.

hach-que:
In all those scenarios you have to run Protobuild anyway to generate the projects.

Protobuild can also track a branch instead of a commit, requiring no modifications to the module configuration when updating the submodule (in that case you also don't need to make a commit in the repo using the submodule).

hach-que:
> This is more of a result of a design decision in Protogame rather than a limitation of git. 

This is a design decision of Protobuild; everything ships as source by default to enable debugging and modification (unless there's a binary package that matches). This is essential in my experience when using third party libraries that you might run into bugs with (and you need to fix those bugs ASAP).

Normally in this scenario (without Protobuild), using manually added binary references is a huge pain because you then need to manually download the source and re-add all the references. But with Protobuild you just do `Protobuild.exe --swap-to-source MonoGame` and now you have MonoGame in source form with all the references wired up correctly in every project. Basically "it just works". This of course assumes MonoGame is shipping binary packages which would be used by default, but currently only my fork of MonoGame has binary packages, because the PR to support it hasn't been merged yet.

tomspilman:
FYI.

I spoke to GitHub about LFS and costs.  They realize that how they current have it designed LFS is unusable for a public projects... as anyone on the planet could grief a project on a whim by making tons of unnecessary clones.

They say they are discussing different rules specifically for public open source projects like MonoGame.

hach-que:
Pretty much the only way LFS will work for open source projects is if the bandwidth is unlimited like repos are. 

This still doesn't negate the issue of having an additional dependency on Git LFS though (which any users of MonoGame in source form will have to install).

hach-que:
(although I guess if Git LFS is just downloading files, I could make Protobuild automatically translate Git LFS references to their downloaded versions)

hach-que:
I removed the submodule change from this PR.

DDReaper:
Any update on this @hach-que ? Seems the updated version of Protobuild is causing the build to fail.

hach-que:
Yeah, there's something going wrong when git is being called by Protobuild
(my guess is permissions on the cache). I updated Protobuild to output the
full Git output on non-zero exit code, but I haven't updated this branch
with the new version of Protobuild.

hach-que:
Looks like someone needs to agree to the XCode / iOS license on the Mac build server:

```
[11:19:11][exec] 
[11:19:11][exec] 
[11:19:11][exec] Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo.
[11:19:11][exec] 
[11:19:11][exec] 
```

I've no idea how things have been working on the build server before now, but this will need whoever owns the Mac build server to fix it.

DDReaper:
@tomspilman @dellis1972 Any ideas why the build server is failing this way?

dellis1972:
@DDReaper I updated the Mac Box the other day and accepted the agreement :) Should be all good now.

hach-que:
In that case, can someone restart the test build?

dellis1972:
@hach-que I would but I don't have access to do that.. easiest way would be to push another change.

DDReaper:
Any update @hach-que ? should only need a re-commit of a tiny change to kick this off again to test

hach-que:
I haven't had a chance to look at this. It's Melbourne International Games
Week here in Australia, so things are super hectic :P

hach-que:
This is now passing and can be merged.

Build/Module.xml
```diff
@@ -1,11 +1,15 @@
...
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Module>
+  <Name>MonoGame.Framework</Name>
+  <DefaultAction>Generate</DefaultAction>
+  <DefaultLinuxPlatforms>Angle,Linux,WindowsGL</DefaultLinuxPlatforms>
+  <DefaultMacOSPlatforms>Angle,MacOS,iOS,WindowsGL,Android</DefaultMacOSPlatforms>
+  <DefaultWindowsPlatforms>Android,Angle,Linux,Ouya,Windows8,Windows,WindowsGL,WindowsPhone,WindowsPhone81,WindowsUAP,iOS</DefaultWindowsPlatforms>
+  <SupportedPlatforms>Android,Angle,Linux,Ouya,Windows8,Windows,WindowsGL,WindowsPhone,WindowsPhone81,WindowsUAP,iOS,MacOS,Web</SupportedPlatforms>
+  <DisableSynchronisation>true</DisableSynchronisation>
+  <GenerateNuGetRepositories>true</GenerateNuGetRepositories>
```
tomspilman: What is `<GenerateNuGetRepositories>` for and why is it enabled here?

Build/Module.xml
```diff
@@ -1,11 +1,15 @@
...
+  <Packages>
+    <Package Uri="https-git://github.com/OutOfOrder/MonoKickstart.git" Folder="ThirdParty/Kickstart" GitRef="master" />
+    <Package Uri="https-git://github.com/Mono-Game/MonoGame.Dependencies.git" Folder="ThirdParty/Dependencies" GitRef="master" />
+  </Packages>
```
tomspilman: What are these packages for?  What is the effect of having these defined here?

Build/Module.xml
hach-que: This generates the "repositories.config" file if you have a "packages.config" file in any of the projects.  "true" was the default value before.  We switched away from the .NET serializer to just reading / writing XML nodes manually to avoid the serializer saving fields we didn't want when run under Mono.  This means that all of the nodes now get written out to the module configuration file.

Build/Module.xml
hach-que: This just tells Protobuild that these folders are Protobuild packages sourced directly from Git.  It has no effect other than that you can now use `Protobuild.exe --upgrade-all` to update them to the latest versions (but you'll still need to commit and merge those updates because the packages are being tracked in Git, not in Protobuild).

tomspilman:
Thanks for the explanations @hach-que.  Merging...

PR #4133, opened 2015-09-11
GraphicsMetrics update

Shqrdx:
This is update for GraphicsMetrics which resolves https://github.com/mono/MonoGame/issues/4111

I also makes all metrics values appends after actual calls(for example DrawCount will appends after PlatformDrawPrimitives and not before) - which is more nice in my opinion, and changed docs a bit.


Shqrdx:
@tomspilman Merge ? rds1983 agreed

KonajuGames:
Waiting for the build system to be rebooted.  The Mac stopped responding.

KonajuGames:
Merged. Thanks @Shqrdx.

PR #4135, opened 2015-09-12
DesktopGL templates

cra0zy:
Stuff done:
 - Removed WindowsGL templates and renamed Linux templates to MonoGame OpenGL
 - Completely removed Linux assemblies from installer, they are literally the same as WindowsGL ones
 - Made templates WinExe, this means they are graphical exe, was set to Exe which meant it was a console application


cra0zy:
@tomspilman @KonajuGames  whats up with the build server?

PR #4136, opened 2015-09-14
Support for textures with mipmaps and cubemaps from DDS files

### LINKED ISSUE #4062: TextureCube support in MonoGame pipeline
TextureCube objects do not appear to be supported by the MonoGame pipeline.  When passing a multiple face dds file into the pipeline, the result, at load time, is a Texture2D object instead of a TextureCube.  

This issue appears to have been previously discovered at http://community.monogame.net/t/loading-cubemaps-generated-with-monogame-pipeline/1025 but an issue was never recorded

KonajuGames:
General clean-up of texture processing for all faces and mipmaps contained in the imported texture.

Fixes #4062.


nkast:
Good to see this added to  3.5.
I imported the assets of a full game and the texture importer is ok. 
Did some quick tests of cube maps also. The importer preserves the mipmaps from the DDS as it should.

One scenario I want to see supported in the future is to convert compressed DDS files into Color. The OpenGL doesn't support compressed cube/3D textures.

tomspilman:
> The OpenGL doesn't support compressed cube/3D textures.

It doesn't?  Weird.

nkast:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/TextureCube.OpenGL.cs#L49-L52
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/TextureCube.OpenGL.cs#L96-L99

nkast:
Still, even if it's possible and someone implement it, it would be nice to drop older DXT compressed DDS into the pipeline GUI and have it decompress it in case of iOS and convert it to Color or PVR.

KonajuGames:
Supporting compressed cubemaps in MonoGame will simply be a case of copying the relevant bit of code to upload compressed texture data from Texture2D.cs.  I've been trying to find more information on your previous statement that GLES doesn't support DXT for cubemap, but I think it is more likely that many Android devices don't support DXT.  GLES does support [compressed cubemaps](https://www.khronos.org/opengles/sdk/docs/man/xhtml/glCompressedTexImage2D.xml).  Just need to use a compression format that the Android device supports.

Decompressing a compressed texture from one format and compressing into another format can be done, but the quality you get may not be the best.  Each compression format loses a certain band of information from the original texture, and using two compression formats on the one texture results in losing more information than each compression separately.

KonajuGames:
Merging this.  Further enhancements can be done as a separate PR.

nkast:
Great!

PR #4141, opened 2015-09-17
Added BlendFactor to GraphicsDevice

Shqrdx:
Changes:

-BlendFactor added to GraphicsDevice and acts as an easy way to access to GraphicsDevice.BlendState.BlendFactor.

-Clone method in BlendState is now public and this provides way to clone popular things like BlendState.AlphaBlend for easy prototyping

Demo screenshot
image

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Xna.Framework.Graphics
 
         private bool _isDisposed;
 
+        private Color _blendFactor;
         private BlendState _blendState;
         private BlendState _actualBlendState;
         private bool _blendStateDirty;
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -324,6 +325,33 @@ public RasterizerState RasterizerState
...
+                newBlendState.BlendFactor = value;
+
+                newBlendState.BindToGraphicsDevice(this);
+
+                _blendState = newBlendState;
+
+                _actualBlendState = newBlendState;
+
+                _blendStateDirty = true;
+            }
```
tomspilman: This PR has the right idea, but implements it wrong.

You can see here how the blend factor is a separate DirectX API from the blend state:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/States/BlendState.DirectX.cs#L46

Same with OpenGL:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/States/BlendState.OpenGL.cs#L41

We need to move the blend factor setting completely into the `GraphicsDevice` and apply it during `PlatformApplyState`.

Then the setter for `GraphicsDevice.BlendState`:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.cs#L330

... should simply do:

```
this.BlendFactor = _blendState.BlendFactor;
```
I suspect if you test this all via a unit test you'll see that setting `GraphicsDevice.BlendState` changes the value of `GraphicsDevice.BlendFactor`.

tomspilman: Also probably won't hurt to track `_blendFactorDirty` separately as often the factor is white and does not change.  This removes an extra API call for each blend state change.

Shqrdx:
Ok, then I will remake it..

PR #4143, opened 2015-09-17
Added a null check on Window disposal

mrhelmut:
We are getting a few crash reports related to ```WinFormsGamePlatform.Exit()``` throwing NullException's. I haven't identified why it is called multiple times (most probably gamers button mashing the exit button), but this little null check shouldn't harm. :)


KonajuGames:
Merging.  Thanks.

PR #4152, opened 2015-09-25
Added support for MSBuild.Content.Builder on MacOS

dellis1972:
In order to support the Content.Builder.targets on
MacOS we firstly needed to install the Tooling and
Assemblies in the right location. In this case
/Library/Frameworks/MonoGame.framework/vX.X.X.X

Symbolic links are required to get everything working
right. In this case we need a v3.0 link in
/Library/Frameworks/MonoGame.framework but we also need
some links in the Mono.framework as well. Namely

/Library/Frameworks/Mono.framework/External/MonoGame

We also need to add an Imports element to the csproj files.
Now the current MonoDevleop/Xamarin Studio templating system
does not support doing this directly. So a new MSBuild
processing class has been added to the addin. This class
will check if the project is a "MonoGame" project (is references
MonoGame.Framework.dll or has a MonoGamePlatform element) and
inject the required data into the csproj.

The templates have been left referencing the Nuget packages
for now.

Modified the ContentBuilder targets file to Batch its items
this is to get round issues when running under xbuild on Mac.

Updated templates to use a new MonoGameInstallDirectory property
rather than $(MSBuildProgramFiles32) so that we can use the
same paths for references.

It also adds the followiing to the projects

```
<MonoGameInstallDirectory Condition="'$(OS)' != 'Unix' ">$(MSBuildProgramFiles32)</MonoGameInstallDirectory>
<MonoGameInstallDirectory Condition="'$(OS)' == 'Unix' ">$(MSBuildExtensionsPath)</MonoGameInstallDirectory>
```

So that we can resolve the directory correctly on both Windows
Mac and Linux


dellis1972:
@KonajuGames @tomspilman anything I need to rework on this?

MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
```diff
@@ -25,16 +20,21 @@
 
     <PropertyGroup>
 
-      <ContentRootDirectory>Content</ContentRootDirectory>
-      <ParentOutputDir>$(ProjectDir)$(ContentRootDirectory)\bin\$(MonoGamePlatform)</ParentOutputDir>
-      <ParentIntermediateDir>$(ProjectDir)$(ContentRootDirectory)\obj\$(MonoGamePlatform)</ParentIntermediateDir>
+      <ContentRootDirectory Conditon=" '$(ContentRootDirectory)' == '' " >Content</ContentRootDirectory>
+      <ParentOutputDir Conditon=" '$(ParentOutputDir)' == '' " >$(ProjectDir)$(ContentRootDirectory)\bin\$(MonoGamePlatform)</ParentOutputDir>
+      <ParentIntermediateDir Conditon=" '$(ParentIntermediateDir)' == '' " >$(ProjectDir)$(ContentRootDirectory)\obj\$(MonoGamePlatform)</ParentIntermediateDir>
```
KonajuGames: Spelling error.  Conditon -> Condition

default.build
```diff
@@ -72,6 +72,11 @@
       <exec program="xbuild" commandline="/t:Clean /p:Configuration=Release MonoGame.Framework.Linux.sln" />
       <exec program="xbuild" commandline="/t:Build /p:Configuration=Release MonoGame.Framework.Linux.sln" />
     </if>
+    <if test="${os == 'MacOS'}">
+      <exec program="mono" commandline="Protobuild.exe -generate Linux" />
```
KonajuGames: Why is it building Linux if OS == MacOS?

MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
dellis1972: @KonajuGames DOH :) can't see the wood for the trees :P

default.build
dellis1972: I added this for my own personal testing (of the Linux build on mac) I'll take it out :)

KonajuGames:
At least one thing to fix that I spotted.  Other than that, I don't understand MonoDevelop extensions, but you know this area a lot better so I'm happy to merge it.

WardBenjamin:
So.... I've done a bit of playing with this on Linux and managed to get it to work (but then something I did broke it.... unfortunately). Is there any chance that we could do the same things that we do here, but with Linux? The MonoDevelop plugin would probably be mostly the same, and I can help out with paths for the tooling/assemblies if required.

dellis1972:
@WardBenjamin I'm sure it can work on Linux.. the main issue is putting the "tools" and assemblies in the right place.
On mac this is 
```
/Library/Frameworks/MonoGame.framework/Versions/x.x.x.x
```
The assemblies go in there in an Assemblies directory and the Tools in a Tools directory. the targets file is in the root.

with sym links in the following locations

```
/Library/Frameworks/MonoGame.framework/Versions/v3.0 -> to version
/Library/Frameworks/Mono.framework/External/xbuild/MonoGame -> /Library/Frameworks/MonoGame.framework
```

Not sure where any of this stuff should go on linux..You'll need to check where

$(MSBuildExtensionsPath)

resolves to.

WardBenjamin:
yeah, I've written a bit about it already, @dellis1972. You can find a post that Dylan Wilson (craftworkgames) kindly posted for me about it at http://dylanwilson.net.

cra0zy:
> Not sure where any of this stuff should go on linux..You'll need to check where
> 
> $(MSBuildExtensionsPath)
> 
> resolves to.

It resolves to: "/usr/lib/mono/xbuild"

dellis1972:
@KonajuGames updated. just wait for the build bot.

BTW i included the Linux build so we can include the assemblies in the Mac Installer :) This is in prep for the DesktopGL stuff.

default.build
dellis1972: Correction. We need to linux dll for the installer :) easier to build it locally

KonajuGames: That's fine.  Just making sure.

KonajuGames:
Will the tools and libraries being put in $prefix/lib/mono/xbuild be done in a separate PR?

dellis1972:
@KonajuGames I think we should do the linux installer stuff in a separate PR.

KonajuGames:
No worries.  Merging.  Thanks @dellis1972

cra0zy:
I've added MonoGame.Content.Builder.targets to Linux installer in: #4178

PR #4153, opened 2015-09-25
Fix for DesktopGL Exit errors

cra0zy:
Fixes: #3540

Stuff done:
 - Fixes BadWindow, BadCursor and OpenTK: Failed to make context current errors
   - was occurring because of window getting Disposed before the rest of the things completed like disposing of Textures(which require a valid window handle) and was also because of Run Loop stopping with processing of events for the window which would close the window and also make that handle invalid.
 - Fixes AL Sources Error
   - whoever wrote the code didn't dispose of the sources at the end... nothing major
 - Fixes AL Device Not Closed error
   - was occurring because of the BadWindow and BadCursor errors were stopping dispose from completing
 - Fixed native stacktrace error
   - was occurring because of _toolkit.Dispose()... it's an error within OpenTK, but since _toolkit was an unused variable I just removed it

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -88,13 +88,13 @@ namespace Microsoft.Xna.Framework
 		private OpenALSoundController soundControllerInstance = null;
         // stored the current screen state, so we can check if it has changed.
         private bool isCurrentlyFullScreen = false;
-        private Toolkit toolkit;
         private int isExiting; // int, so we can use Interlocked.Increment
+
+        int windowDelay = 2;
         
 		public OpenTKGamePlatform(Game game)
             : base(game)
         {
-            toolkit = Toolkit.Init();
             _view = new OpenTKGameWindow(game);
             this.Window = _view;
 
@@ -137,12 +137,23 @@ namespace Microsoft.Xna.Framework
                 // Note 2: Game.Exit() can be called asynchronously from
                 // _view.ProcessEvents() (cases #2 and #3 above), so the
                 // isExiting check must be placed *after* _view.ProcessEvents()
-                if (isExiting > 0)
+                // Note 3: We need to continue processing view events until  
+                // everything gets disposed of, otherwise it will close the window
+                // and make the window handle invalid
+                if (isExiting == 0)
+                    Game.Tick();
+                else if (windowDelay == 2)
                 {
+                    windowDelay--;
+                    Game.ExitEverything();
+                }
+                else if (windowDelay > 0)
+                    windowDelay--;
+                else
+                {
+                    _view.Dispose();
                     break;
                 }
-
-                Game.Tick();
             }
         }
 
@@ -286,27 +297,6 @@ namespace Microsoft.Xna.Framework
             var device = Game.GraphicsDevice;
             if (device != null)
                 device.Present();
-        }
-		
-        protected override void Dispose(bool disposing)
-        {
-            if (!IsDisposed)
-            {
-                if (toolkit != null)
-                {
-                    toolkit.Dispose();
-                    toolkit = null;
-                }
-
-                if (_view != null)
-                {
-                    _view.Dispose();
-                    _view = null;
-                }
-            }
-
-			base.Dispose(disposing);
-        }
-			
+        }	
     }
 }
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -395,14 +395,27 @@ namespace Microsoft.Xna.Framework
                 break;
             case GameRunBehavior.Synchronous:
                 Platform.RunLoop();
+#if !DESKTOPGL
                 EndRun();
 				DoExiting();
+#endif
                 break;
             default:
                 throw new ArgumentException(string.Format(
                     "Handling for the run behavior {0} is not implemented.", runBehavior));
             }
         }
+
+#if DESKTOPGL
+        // This code is used so that the Window could stay alive
+        // while all the resources are getting destroyed
+        internal void ExitEverything()
+        {
+            EndRun();
+            DoExiting();
+            this.Dispose();
+        }
+#endif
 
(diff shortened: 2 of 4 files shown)


MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
```diff
@@ -88,13 +88,13 @@ class OpenTKGamePlatform : GamePlatform
 		private OpenALSoundController soundControllerInstance = null;
         // stored the current screen state, so we can check if it has changed.
         private bool isCurrentlyFullScreen = false;
-        private Toolkit toolkit;
         private int isExiting; // int, so we can use Interlocked.Increment
+
+        int windowDelay = 11;
```
tomspilman: Is there really not a better way to know if the `OpenTKGameWindow` is safe to destroy than waiting for exactly 11 `ProcessEvents` calls?

MonoGame.Framework/Game.cs
```diff
@@ -404,6 +406,13 @@ public void Run(GameRunBehavior runBehavior)
             }
         }
 
+        internal void ExitEverything()
```
tomspilman: If we're gonna do this it should at least be documented why it exists and wrapped with `#if DESKTOPGL` to avoid people from thinking it is how things should get shut down.

cra0zy: Will do.

MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
cra0zy: Yea, you are right in some way, I only need a delay of 2 to get what is needed to be done.

cra0zy: Also I don't know of a better way.

tomspilman: Just seems fragile to depend on message pumps that are often inconsistent in timing and order to wait some exact number of updates.

Also why isn't OpenTK handling this for us?  It shouldn't be our job to guess when it is safe to dispose their window.

cra0zy: > Just seems fragile to depend on message pumps that are often inconsistent in timing and order to wait some exact number of updates.

Well on the bright side chances that OpenTK error appearing should also be next to never(I run my game 100 times in Windows, 0 exit errors appeared)

> Also why isn't OpenTK handling this for us? It shouldn't be our job to guess when it is safe to dispose their window.

Actually it is our job, we are disposing of a lot of different OpenGL stuff we created by using the window handle.

cra0zy: I just realized that I still have a problem on Linux somewhere...

cra0zy: Just to make myself a bit clearer, it's unrelated to this error, it's a 5th error... yey

tomspilman:
@KonajuGames - If you are good with this then lets merge it.

cra0zy:
@KonajuGames Any chance of this getting merged soonish?

http://www.urbandictionary.com/define.php?term=soonish

KonajuGames:
Sorry.  Another that got lost amongst the noise.  I'm happy with this.  I'll merge it.

@cra0zy "soonish" has long been part of my Aussie vernacular. :)

dellis1972:
@cra0zy I still get exit errors from DesktopGL on Mac

```
libc++abi.dylib: terminating with uncaught exception of type long
Stacktrace:

  at <unknown> <0xffffffff>
  at (wrapper managed-to-native) OpenTK.Audio.OpenAL.Alc.DestroyContext (intptr) <0xffffffff>
  at OpenTK.Audio.OpenAL.Alc.DestroyContext (OpenTK.ContextHandle) <0x00013>
  at Microsoft.Xna.Framework.Audio.OpenALSoundController.CleanUpOpenAL () <0x0007f>
  at Microsoft.Xna.Framework.Audio.OpenALSoundController.Dispose (bool) <0x000b7>
  at Microsoft.Xna.Framework.Audio.OpenALSoundController.Dispose () <0x0001b>
  at Microsoft.Xna.Framework.Audio.OpenALSoundController.DestroyInstance () <0x00027>
  at Microsoft.Xna.Framework.Audio.SoundEffect.PlatformShutdown () <0x0000b>
  at Microsoft.Xna.Framework.Game.Dispose (bool) <0x002bb>
  at Microsoft.Xna.Framework.Game.Dispose () <0x00019>
  at Microsoft.Xna.Framework.Game.ExitEverything () <0x00023>
  at Microsoft.Xna.Framework.OpenTKGamePlatform.RunLoop () <0x0007f>
  at Microsoft.Xna.Framework.Game.Run (Microsoft.Xna.Framework.GameRunBehavior) <0x00167>
  at Microsoft.Xna.Framework.Game.Run () <0x00023>
  at PixelBlocked.DesktopGL.Program.Main () <0x0003b>
  at (wrapper runtime-invoke) object.runtime_invoke_void (object,intptr,intptr,intptr) <0xffffffff>
```

cra0zy:
Hmm.. so this is causing the error I guess? : https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs#L284

Try destroying of _acontext instead of _context(do note that _context is just a handle, while _acontext is an object managed by OpenTK).

mrhelmut:
I'm having this error too on Mac, I'll try that tomorrow.

mrhelmut:
Doing this works without error:

```
        private void CleanUpOpenAL()
        {
            Alc.MakeContextCurrent (ContextHandle.Zero);
            if (_acontext != null)
            {
                _acontext.Dispose();
                _acontext = null;
            }
            _bSoundAvailable = false;
        }
```

I believe that ```Alc.MakeContextCurrent()``` got called twice because of ```AudioContext.Dispose()``` and that MacOS doesn't like this.

I've submitted a PR #4410

PR #4155, opened 2015-09-25
Fix For XAudio Voice Leak

### LINKED ISSUE #4146: Couple issues with XAudio2 (corruption, freeze)
Hey there,

We are getting a few reports related to sfx which may be dependent to a few sound chip manufacturers.
Here are the issues, which may be unrelated, but we highly suspect that they are:

1. **Sound corruption**: after playing for an extended time (~1 hour), ```SoundEffect``` / ```SoundEffectInstance``` get corrupted. All sounds start to crackle a lot. ```MediaPlayer``` / ```Song``` are not affected (which seems normal since ```MediaPlayer``` uses MediaFoundation and not XAudio2).
2. **Freezes / stalls**: when sounds start to crackle, it is almost certain that playing a non pooled ```SoundEffectInstance``` (i.e. through ```SoundEffect.CreateInstance()```) will totally stall the game for several seconds. ```SoundEffect.Play()``` (i.e. pooled instances) doesn't stall the game.

I'm totally puzzled about these issues and we haven't been able to reproduce them internally which makes me think that they are related to specific sound chips and possibly driver issues.

It may also be only related to looped ```SoundEffectInstance```, because they tend to make the game stall more often/quickly.

I also don't know why only non-pooled instances are making the game to stall. I would suspect an hardware instance limit, but since the pool doesn't have a limit on Windows platforms, the problem would have been exposed way before any non-pooled instance is played.

I haven't found the time to test that against OpenAL.

tomspilman:
This is a leak fix with a simple optimization for reuse of voices in XAudio.  It affected any game that relies on `SoundEffect.Play()` or XAct.  It seems like the OpenAL implementation suffers from a similar leak, but that is left to another PR.

Also there is missing optimization here that I added a `TODO` comment for.  We should be finding a `SoundEffectInstance` from the pool that is the format we need else return a new instance.  That is left for another PR.

It would be good for a few people to test this with their games before we merge this.

Thanks to @mrhelmut for the test project that exposed this issue.

Fixes #4146.

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -185,18 +185,23 @@ namespace Microsoft.Xna.Framework.Audio
 
             if (voice != null)
             {
-                // We can reuse this existing voice if the sample rate and
-                // channel count are the same which will reduce garbage generation
-                // and overhead of allocating another native resource.
-
-                var details = voice.VoiceDetails;
-
-                if (details.InputSampleRate != _format.SampleRate ||
-                    details.InputChannelCount != _format.Channels)
+                // TODO: This really shouldn't be here.  Instead we should fix the 
+                // SoundEffectInstancePool to internally to look for a compatible
+                // instance or return a new instance without a voice.
+                //
+                // For now we do the same test that the pool should be doing here.
+             
+                if (!ReferenceEquals(inst._format, _format))
                 {
-                    voice.DestroyVoice();
-                    voice.Dispose();
-                    voice = null;
+                    if (inst._format.Encoding != _format.Encoding ||
+                        inst._format.Channels != _format.Channels ||
+                        inst._format.SampleRate != _format.SampleRate ||
+                        inst._format.BitsPerSample != _format.BitsPerSample)
+                    {
+                        voice.DestroyVoice();
+                        voice.Dispose();
+                        voice = null;
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


MichaelDePiazzi:
Nice work finding this @tomspilman and @mrhelmut! :)

I've got a multiplayer session of my game scheduled for Sunday afternoon (GMT+8), so I can get out a new build to my testers and give these changes a good test then.

mrhelmut:
Thanks @tomspilman !

I've sent a test build of our game to a few players who had the issue, and I will also test this on computers we have at the office by Monday. I'll let you know if it fixes the sound corruption, but I can confirm that the leak is gone.

MichaelDePiazzi:
@tomspilman I just finished a multiplayer session with a few of my testers which lasted just over 90 minutes. These changes seemed to do the trick. I didn't see any of the stuttering or memory leaks that I got previously.

tomspilman:
Fantastic... thanks for the report @MichaelDePiazzi !

Lets be safe and see what @mrhelmut finds.  If all looks good we can merge!

mrhelmut:
Everything's clear on our side!
Our testers registered a few hours of gameplay without any sound issues or freeze, and our test computers have been running a couple hours successfully.

Thanks again.

tomspilman:
Great... @KonajuGames this is good to merge then.

@mrhelmut - The next step would be fixing `SoundEffectInstancePool.GetInstance()` to take a `SoundEffect` as an input argument.  Then adding an internal platform method to `SoundEffect` that lets the pool to test a `SoundEffectInstance` for compatibility with the effect's internal format.  For some platforms this may do nothing... but for XAudio we need to do the same test i'm doing here.

mrhelmut:
@tomspilman I'll try to implement that.

Maybe #4095 would be worth merging for 3.5 too. It's definitely going to be a nice release regarding XAudio / MediaFoundation improvements.

KonajuGames:
Merging.

PR #4157, opened 2015-09-25
ContentManager.Unload(asset), More strict memory management

HeshamMeneisi:
I made 2 changes to the content manager when I was working on a project and came across some memory problems
#1

I noticed the absence of this crucial function, Unload(asset), in the content manager so I decided to add it. This is essentially an alternative for (resource).Dispose() when the resource is loaded through a content manager.
#2

As much as I hated to, I had to use GC.Collect() in the ReadAsset and Unload functions, this seemed to be the only way to prevent crashing on devices with limited resources when attempting to "replace" current loaded data as in, for example, a theme change or moving from level to level with largely varying content. It also seems to be the only way to prevent a memory leak from occurring especially on Windows UAP because of the rather delayed collection intervals, the following graphs should clarify what I mean:

The current memory stats while I continuously switch between 2 sets of textures on UAP:

Without collect calls:
nogc

The memory simply grows endlessly in a memory leak resembling pattern. This memory is NOT reusable as this test would crash the 512mb emulator or a low end device immediately (Also reported on android by a project member). The memory is occasionally collected by the GC to a slightly lower value but this behavior would continue until the memory is out.

After adding collect calls:
agc

The issue has been reduced to those spikes, which I have found to be associated with the async nature of the WINRT file system, but so far I have not thought of a solution to this probelm. Also calling GC.Collect() once after the loading process on game side does not reduce memory usage as opposed to the call once per ReadAsset.

Note: Visual studio reformatted the document (which I have found conforming to the guidelines) so here are the lines I edited to make it easier for anyone interested in seeing the actual change.
215                  Unload(asset)
248,377,687  GC.Collect()

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -17,22 +17,22 @@ using Microsoft.Xna.Framework.Media;
 
 namespace Microsoft.Xna.Framework.Content
 {
-	public partial class ContentManager : IDisposable
-	{
+    public partial class ContentManager : IDisposable
+    {
         const byte ContentCompressedLzx = 0x80;
         const byte ContentCompressedLz4 = 0x40;
 
-		private string _rootDirectory = string.Empty;
-		private IServiceProvider serviceProvider;
-		private IGraphicsDeviceService graphicsDeviceService;
+        private string _rootDirectory = string.Empty;
+        private IServiceProvider serviceProvider;
+        private IGraphicsDeviceService graphicsDeviceService;
         private Dictionary<string, object> loadedAssets = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
-		private List<IDisposable> disposableAssets = new List<IDisposable>();
+        private List<IDisposable> disposableAssets = new List<IDisposable>();
         private bool disposed;
-		
-		private static object ContentManagerLock = new object();
+
+        private static object ContentManagerLock = new object();
         private static List<WeakReference> ContentManagers = new List<WeakReference>();
 
-	static List<char> targetPlatformIdentifiers = new List<char>()
+        static List<char> targetPlatformIdentifiers = new List<char>()
         {
             'w', // Windows (DirectX)
             'x', // Xbox360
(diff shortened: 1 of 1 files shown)


nkast:
2)
Please test  #3518 It will improve memory allocation/GC and greatly reduce the loading time of assets. I am very interested to hear your findings.

nkast:
Can we see the time bar on the profiler too?  
I suppose Calling GC.Collect() on every asset will make loading slower. This will effect mobile devices more.

KonajuGames:
Calling `GC.Collect()` so often is not the best for performance.  I do not understand how it could happen that memory use climbs constantly and only by calling GC.Collect() does it stay constant.  That goes against everything the garbage collector is supposed to do.

HeshamMeneisi:
@nkast Sure I will try it as soon as I have enough time.
The difference in time is almost negligible. Here are the times for my test texture set:
Avg time without manual collection is 110ms (min 98 - max 122)
Avg time after adding manual collection is 121ms (min 118 - max 124)

@KonajuGames 
Yes it does go against the concept of GC. It doesn't largely affect the performance but normally one should never need to actually call GC.Collect() and that was the case on my tests on Windows DX. This only happens on UAP and probably android but I have not tried it on android yet. It seems the garbage collector is not very functional for some reason. I found out it does do full reclamation but that happens after 10-20 seconds of constantly climbing memory usage and even sometimes it just never does! The GC collect happens but does nothing. Also quick enough loading/disposal of material can easily overflow low end devices before the GC does anything. One would normally expect the streams used during the loading process of a file to get claimed by GC right before the next loading operations starts but those streams just keep piling up, not even reused because the total memory usage climbs to doubles of the total files being loaded.

clee-dev:
Has there been any more looking into this? I was about to start implementing the same method until I saw this pull

nkast:
@clee-dev 
No because nobody could reproduce this bug and not enough information was given. 
Personally I dismiss the idea that GC doesn't collect data automatically and also that calling GC.Collect()  manually so frequently has no performance impact on real application.  

All the above might sound unrelated to you but remember that this PR mix 2 separate issues which makes it very difficult to get a green-light.

As for the absence of an Unload() method, the main reason I think is that it can't unload linked resources.
One example is the textures of a Model. ContentManager does not keep a link between the two, nor does it know how many assets use the same resource or if you loaded the texture individually.
And since it's not 100% guaranteed to fully unload a resource it was left out from the original API.
But it's very easy to [expand ContentManager](https://blogs.msdn.microsoft.com/shawnhar/2007/03/09/contentmanager-readasset/) to your own needs.

clee-dev:
@nkast 
Thanks for the insight. The linked resources issue is understandable, I just figured that if someone is opting to use `Unload(resource)` specifically, then they should be aware they could have issues if something later attempts to use the resource. However, if it's too iffy (could cause too many problems/confusion, etc) to bother trying to get a pull approved for the official API, then I'll go with wrapping it as you've suggested. Appreciate the timely response.

nkast:
Well, I figure that Model is not `IDisposable`. 
Also disposable assets is reasonable that should not dispose their external/shared assets.

After all , maybe it's possible to add an Unload<asset> method without this causing any serious problems.

PR #4158, opened 2015-09-26
Vector3 performance improvements

Shqrdx:
Batch of performance improvements for Vector3.

inline from "ref" versions
-Cross
-Distance
-Normalize
-Transform
-TransformNormal
-operator !=

simplification(e.g "float" to "var")
-Negate
-Normalize
-Reflect
-ToString - removed StringBuilder
-Transform
-operator /

other:
-GetHashCode - changed to XNA behaviour
-fixed bug in Vector4.GetHashCode

+new Tests for all changed things(tested with XNA)

### DIFF
--- a/MonoGame.Framework/Vector3.cs
+++ b/MonoGame.Framework/Vector3.cs
@@ -353,9 +353,13 @@ namespace Microsoft.Xna.Framework
         /// <param name="result">The cross product of two vectors as an output parameter.</param>
         public static void Cross(ref Vector3 vector1, ref Vector3 vector2, out Vector3 result)
         {
-            result.X = vector1.Y * vector2.Z - vector2.Y * vector1.Z;
-            result.Y = -(vector1.X * vector2.Z - vector2.X * vector1.Z);
-            result.Z = vector1.X * vector2.Y - vector2.X * vector1.Y;
+            float x = vector1.Y*vector2.Z - vector2.Y*vector1.Z;
+            float y = -(vector1.X*vector2.Z - vector2.X*vector1.Z);
+            float z = vector1.X*vector2.Y - vector2.X*vector1.Y;
+
+            result.X = x;
+            result.Y = y;
+            result.Z = z;
         }
 
         /// <summary>
--- a/MonoGame.Framework/Vector3.cs
+++ b/MonoGame.Framework/Vector3.cs
@@ -353,9 +353,9 @@ namespace Microsoft.Xna.Framework
         /// <param name="result">The cross product of two vectors as an output parameter.</param>
         public static void Cross(ref Vector3 vector1, ref Vector3 vector2, out Vector3 result)
         {
-            float x = vector1.Y*vector2.Z - vector2.Y*vector1.Z;
-            float y = -(vector1.X*vector2.Z - vector2.X*vector1.Z);
-            float z = vector1.X*vector2.Y - vector2.X*vector1.Y;
+            var x = vector1.Y*vector2.Z - vector2.Y*vector1.Z;
+            var y = -(vector1.X*vector2.Z - vector2.X*vector1.Z);
+            var z = vector1.X*vector2.Y - vector2.X*vector1.Y;
 
             result.X = x;
             result.Y = y;
@@ -1119,9 +1119,9 @@ namespace Microsoft.Xna.Framework
         /// <param name="result">Transformed normal as an output parameter.</param>
         public static void TransformNormal(ref Vector3 normal, ref Matrix matrix, out Vector3 result)
         {
-            var x = (normal.X*matrix.M11) + (normal.Y*matrix.M21) + (normal.Z*matrix.M31);
-            var y = (normal.X*matrix.M12) + (normal.Y*matrix.M22) + (normal.Z*matrix.M32);
-            var z = (normal.X*matrix.M13) + (normal.Y*matrix.M23) + (normal.Z*matrix.M33);
+            var x = (normal.X * matrix.M11) + (normal.Y * matrix.M21) + (normal.Z * matrix.M31);
+            var y = (normal.X * matrix.M12) + (normal.Y * matrix.M22) + (normal.Z * matrix.M32);
+            var z = (normal.X * matrix.M13) + (normal.Y * matrix.M23) + (normal.Z * matrix.M33);
             result.X = x;
             result.Y = y;
             result.Z = z;
--- a/MonoGame.Framework/Vector3.cs
+++ b/MonoGame.Framework/Vector3.cs
@@ -787,7 +787,13 @@ namespace Microsoft.Xna.Framework
         /// <returns>Reflected vector.</returns>
         public static Vector3 Reflect(Vector3 vector, Vector3 normal)
         {
+            // I is the original array
+            // N is the normal of the incident plane
+            // R = I - (2 * N * ( DotProduct[ I,N] ))
+
             Vector3 reflectedVector;
+            
+            // inline the dotProduct here instead of calling method
 
             var dotProduct = ((vector.X * normal.X) + (vector.Y * normal.Y)) + (vector.Z * normal.Z);
             reflectedVector.X = vector.X - (2.0f * normal.X) * dotProduct;
@@ -805,6 +811,12 @@ namespace Microsoft.Xna.Framework
         /// <param name="result">Reflected vector as an output parameter.</param>
         public static void Reflect(ref Vector3 vector, ref Vector3 normal, out Vector3 result)
         {
+            // I is the original array
+            // N is the normal of the incident plane
+            // R = I - (2 * N * ( DotProduct[ I,N] ))
+
+            // inline the dotProduct here instead of calling method
+
             var dotProduct = ((vector.X * normal.X) + (vector.Y * normal.Y)) + (vector.Z * normal.Z);
             result.X = vector.X - (2.0f * normal.X) * dotProduct;
             result.Y = vector.Y - (2.0f * normal.Y) * dotProduct;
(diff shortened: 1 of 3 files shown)


KonajuGames:
Did you measure the performance gains?

Shqrdx:
@KonajuGames erhm. no. I dont know what the effecient way to do it. Could you suggest ?

MonoGame.Framework/Vector3.cs
```diff
@@ -351,12 +353,9 @@ public static Vector3 Cross(Vector3 vector1, Vector3 vector2)
...
         {
-            var x = vector1.Y * vector2.Z - vector2.Y * vector1.Z;
-            var y = -(vector1.X * vector2.Z - vector2.X * vector1.Z);
-            var z = vector1.X * vector2.Y - vector2.X * vector1.Y;
-            result.X = x;
-            result.Y = y;
-            result.Z = z;
+            result.X = vector1.Y * vector2.Z - vector2.Y * vector1.Z;
+            result.Y = -(vector1.X * vector2.Z - vector2.X * vector1.Z);
+            result.Z = vector1.X * vector2.Y - vector2.X * vector1.Y;
```
KonajuGames: Using the local variables was for a reason.  The same variable may be used for vector1 or vector2 and for the result.  Your changes here will result in an incorrect cross product because later calculations will be affected by earlier calculations.  Calculate the cross product into a separate vector (our three local variables) and then assign to the result.

MonoGame.Framework/Vector3.cs
```diff
@@ -778,12 +783,9 @@ public static void Normalize(ref Vector3 value, out Vector3 result)
         /// <returns>Reflected vector.</returns>
         public static Vector3 Reflect(Vector3 vector, Vector3 normal)
         {
-            // I is the original array
-            // N is the normal of the incident plane
-            // R = I - (2 * N * ( DotProduct[ I,N] ))
             Vector3 reflectedVector;
```
KonajuGames: Explanatory comments like these shouldn't be removed.  They help to explain what is happening in the following code.

MonoGame.Framework/Vector3.cs
```diff
@@ -799,12 +801,7 @@ public static Vector3 Reflect(Vector3 vector, Vector3 normal)
         /// <param name="result">Reflected vector as an output parameter.</param>
         public static void Reflect(ref Vector3 vector, ref Vector3 normal, out Vector3 result)
         {
-            // I is the original array
-            // N is the normal of the incident plane
-            // R = I - (2 * N * ( DotProduct[ I,N] ))
-
-            // inline the dotProduct here instead of calling method
```
KonajuGames: Same here.  Please don't remove comments that serve a purpose.

MonoGame.Framework/Vector3.cs
```diff
@@ -906,12 +898,9 @@ public static Vector3 Transform(Vector3 position, Matrix matrix)
...
         {
-            var x = (position.X * matrix.M11) + (position.Y * matrix.M21) + (position.Z * matrix.M31) + matrix.M41;
-            var y = (position.X * matrix.M12) + (position.Y * matrix.M22) + (position.Z * matrix.M32) + matrix.M42;
-            var z = (position.X * matrix.M13) + (position.Y * matrix.M23) + (position.Z * matrix.M33) + matrix.M43;
-            result.X = x;
-            result.Y = y;
-            result.Z = z;
+            result.X = (position.X * matrix.M11) + (position.Y * matrix.M21) + (position.Z * matrix.M31) + matrix.M41;
+            result.Y = (position.X * matrix.M12) + (position.Y * matrix.M22) + (position.Z * matrix.M32) + matrix.M42;
+            result.Z = (position.X * matrix.M13) + (position.Y * matrix.M23) + (position.Z * matrix.M33) + matrix.M43;
```
KonajuGames: The local variables here are for the same purpose as in Cross().

MonoGame.Framework/Vector3.cs
```diff
@@ -1100,12 +1100,9 @@ public static Vector3 TransformNormal(Vector3 normal, Matrix matrix)
...
         {
-            var x = (normal.X * matrix.M11) + (normal.Y * matrix.M21) + (normal.Z * matrix.M31);
-            var y = (normal.X * matrix.M12) + (normal.Y * matrix.M22) + (normal.Z * matrix.M32);
-            var z = (normal.X * matrix.M13) + (normal.Y * matrix.M23) + (normal.Z * matrix.M33);
-            result.X = x;
-            result.Y = y;
-            result.Z = z;
+            result.X = (normal.X * matrix.M11) + (normal.Y * matrix.M21) + (normal.Z * matrix.M31);
+            result.Y = (normal.X * matrix.M12) + (normal.Y * matrix.M22) + (normal.Z * matrix.M32);
+            result.Z = (normal.X * matrix.M13) + (normal.Y * matrix.M23) + (normal.Z * matrix.M33);
```
KonajuGames: Local variables have a purpose again.

MonoGame.Framework/Vector3.cs
Shqrdx: // inline the dotProduct here instead of calling method - what is this method ?

Shqrdx:
@KonajuGames Now its ok ?

tomspilman:
It really bothers me that Cross and Transform were clearly broken with your earlier changes, but the unit tests didn't expose that as the build server passed everything.  Can we investigate why that is and fix the tests?

Shqrdx:
because for example Cross been written as "Vector3.Cross(ref v1, ref v2, out result1);" and it will fail if 
result1 is v1...

tomspilman:
We should add an additional test to catch that case.  In fact I think all the math types that have `ref/out` parameters should get these types of additional tests.

Shqrdx:
@tomspilman And there is a lot of tests are written in this manner.. Not only by me..

KonajuGames:
I would need to see tests that show performance improvements before this could be merged.  As it stands, I don't think there would be much in the way of performance gains.  The JIT or AOT compiler will likely already inline many of those calls from the pass-by-value to the pass-by-ref methods.

MonoGame.Framework/Vector3.cs
```diff
@@ -522,7 +530,7 @@ public bool Equals(Vector3 other)
         /// <returns>Hash code of this <see cref="Vector3"/>.</returns>
         public override int GetHashCode()
         {
-            return (int)(this.X + this.Y + this.Z);
+            return X.GetHashCode() + Y.GetHashCode() + Z.GetHashCode();
```
KonajuGames: Is this shown to be a better hashcode generation?

Shqrdx: the float generates different values than integer - for example hash code for 1.0f is 1065353216 whereas 1 integer generates 1 hash code.

Shqrdx: So the purpose is to make it close to XNA behavior as possible.

Shqrdx:
@KonajuGames ok I will test a Cross method... If the JIT is generates it better without I close this PR and make a new with changes to ToString and GetHashCode only,

MonoGame.Framework/Vector3.cs
KonajuGames: I don't know XNA generates hashcodes for Vectors, and no-one should know until MS release the source for it.  But simply adding the hashcodes for the separate components doesn't make for a good hashcode.  Granted, the existing version did that anyway, and it truncated floats, so it wasn't the best, but your new version suffers the same problem.  Take the following three vectors.  They will generate the same hashcode because the component hashcodes are simply added together.

{ 1.0f, 2.0f, 3.0f }
{ 3.0f, 1.0f, 2.0f }
{ 2.0f, 3.0f, 1.0f }

The component hashcodes ideally should be combined in a manner that is not commutative (plus is commutative because it doesn't matter what order the operands are in).

Shqrdx: I look into XNA dll using dotPeek.

Shqrdx:
So the new Cross 0.0109 ms for 10000000 calculations, old one is 0.0156

MonoGame.Framework/Vector3.cs
KonajuGames: No.  That is not permitted.  Any evidence of browsing into XNA assemblies will have the PR closed.

tomspilman: > I looked into XNA dll using dotPeek.

Then we have to close this PR and unfortunately can't accept any more submissions from you.

You cannot steal code or ideas from peeking at the XNA source code.

Shqrdx: Nothing has been taked. the XNA 
using "return this.X.GetHashCode() + this.Y.GetHashCode() + this.Z.GetHashCode();"
The new one is "return X.GetHashCode() + Y.GetHashCode() + Z.GetHashCode();" See the difference ?

Shqrdx: "unfortunately can't accept any more submissions from you." 
Hey are you serious ?

tomspilman: > Hey are you serious ?

Yes.

You don't seem to understand what your doing is illegal.  You cannot look at the XNA code, change it slightly and submit it to MonoGame.  Just like you cannot copy a book someone wrote... change a few words around... and call it yours.

Shqrdx: I could - thats my distinction from you.

tomspilman: > I could - thats my distinction from you.

Again... you are not understanding me.  You cannot contribute code... however you may have changed it around... that you do not own.  This is a rule that *all* open source projects have.

By telling us you are copying code from XNA we have no choice but to do this.  I don't want to ...  but I have to because it has to potential to hurt the project.

Shqrdx: forever or only temporary ?

tomspilman: > forever or only temporary ?

I really don't know...  I have to figure out what we need to do first.

tomspilman: @Shqrdx - Just a heads up on what is going on.  We're reviewing all the code you've submitted for potential violations of copyright.  After we finish this we'll let you know what we have found.

Once we get thru all that we can see how you can contribute moving forward.  I think we can work something out.

Shqrdx: And how the results ? Could I push a small PR now ?

PR #4159, opened 2015-09-26
[Gtk Pipeline] Fix possible crash when build is called

cra0zy:
It's much easier to see what I did by adding "?w=1" to url, just needed to make sure that the code was run from the GUI thread.


tomspilman:
Thanks... merging!

PR #4162, opened 2015-09-30
Sound instance pool optimization to look for matching format in the pool

mrhelmut:
This is a follow up to #4155 where @tomspilman suggested this optimization.

It checks if there is any `SoundEffectInstance` with the same format of the given `SoundEffect` in the pool to efficiently re-use instances.
This only impacts XAudio for now.

It would be cool to document this in the documentation to warn that using too many different `SoundEffect` formats may impact performance (or the opposite, that using the same format for all sounds is best).

In our project, all sounds have the same format, hence it would be nice to test this PR against more varied projects.

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
@@ -166,6 +166,17 @@ namespace Microsoft.Xna.Framework.Audio
             inst.InitializeSound();
             inst.BindDataBuffer(_data, Format, Size, (int)Rate);
         }
+		
+		/// <summary>
+        /// Test if a SoundEffectInstance is compatible (i.e. same sampling rate, number of channels, etc.) with the SoundEffect.
+        /// This method is used by the SoundEffectInstancePool to re-use instances efficiently.
+        /// </summary>
+        /// <param name="inst">The SoundEffectInstance to test</param>
+        /// <returns>True if compatible, false otherwise</returns>
+        internal bool PlatformIsInstanceCompatible(SoundEffectInstance inst)
+        {
+            return true; // OpenAL don't care about that
+        }
 
         #endregion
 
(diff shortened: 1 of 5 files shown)


tomspilman:
Maybe you should remove the code in `SoundEffect.PlatformSetupInstance` that we're not doing this same test twice.  For safety you could always add a debug assert that calls your `PlatformIsInstanceCompatible` to be sure no incompatible instances are slipped thru.

mrhelmut:
Here you go. I forgot to clean ```SoundEffect.PlatformSetupInstance```.

KonajuGames:
I'd rather the assert wasn't there.  We don't use asserts anywhere else in the code.  Also, the code still needs to handle it nicely when compiled in Release because the assert won't be there.

mrhelmut:
Should I remove this assert completely then? The purpose was to avoid the double-check in Release, or did I miss the point?

KonajuGames:
The only way it would have a voice set already is if it came from the pool, and it has to pass the compatibility test before it got selected from the pool anyway.  There is no need to check it again.

mrhelmut:
Updated. :)

tomspilman:
Anyone have time to test this in their game before we merge this?

MichaelDePiazzi:
@tomspilman I would, but the samples in my game all use the same format so it wouldn't really test this PR properly...

KonajuGames:
The disadvantage of not doing any actual game dev with MonoGame at the moment is that I don't have any good stress tests for this.  And I would likely have the same formats for all sounds anyway.

I think the best way to test this will be to get it into the develop branch and a release.

mrhelmut:
I can't thoroughly test that one either, all of our sounds are of the same format too.

If I find some time, I'll try to come up with a stress test app.

mrhelmut:
I'm checking on my opened PR. Should I look into resolving the merge conflicts and move forward with this one? Or should we close it and open an issue for optimization request?

PR #4163, opened 2015-09-30
Fix for project.json

BrianPeek:
Rename project.json to MonoGame.Framework.WindowsUAP.project.json so it only builds with the UAP project (requires NuGet 3.2 and Tools for Universal apps v1.1)


tomspilman:
@BrianPeek @kjpou1 ...  sorry... this PR totally slipped thru my inbox.

PR #4165, opened 2015-10-01
Fix ContentExtensions.GetAllProperties for Win10 when Native toolchain enabled

suilevap:
I  found that loading Xml from content files works incorrect for Win10 in release mode (or to be more precise when native toolchain is enabled).
It looks like `ContentExtensions.GetAllProperties` always return empty array due to condition `(p.GetMethod == p.GetMethod.GetRuntimeBaseDefinition())` (which is true without native toolchain, and false with) .
I am not sure what exactly should check this code, but in my case it works as expected without it.

### DIFF
--- a/MonoGame.Framework/Content/ContentExtensions.cs
+++ b/MonoGame.Framework/Content/ContentExtensions.cs
@@ -33,8 +33,7 @@ namespace Microsoft.Xna.Framework.Content
 #if WINRT
             PropertyInfo[] infos= type.GetTypeInfo().DeclaredProperties.ToArray();
             var nonStaticPropertyInfos = from p in infos
-                                         where (p.GetMethod != null) && (!p.GetMethod.IsStatic) &&
-                                         (p.GetMethod == p.GetMethod.GetRuntimeBaseDefinition())
+                                         where (p.GetMethod != null) && (!p.GetMethod.IsStatic)
                                          select p;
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Content/ContentExtensions.cs
```diff
@@ -33,8 +33,7 @@ public static PropertyInfo[] GetAllProperties(this Type type)
 #if WINRT
             PropertyInfo[] infos= type.GetTypeInfo().DeclaredProperties.ToArray();
             var nonStaticPropertyInfos = from p in infos
-                                         where (p.GetMethod != null) && (!p.GetMethod.IsStatic) &&
-                                         (p.GetMethod == p.GetMethod.GetRuntimeBaseDefinition())
```
KonajuGames: Do we know what this condition was supposed to do? @tomspilman?

tomspilman: It was trying to remove duplicate properties when one classes overrides the property from a base class.

This is used from the `ReflectiveReader/Writer`.  So some additional tests there would help make me feel better about this change.

Another idea is to wrap this change in a `#if WINDOWS_UAP` so that it only affects Win10.

tomspilman:
Closing this as the side effects are unknown.  I would be good to re-open and merge this with some unit tests to verify the changes work correctly on all platforms.

PR #4168, opened 2015-10-04
[Gtk Pipeline] Added new output view

cra0zy:
https://youtu.be/jTW6AQHe9cA


tomspilman:
Adding some sort of filtered view of the build output wouldn't be a bad idea.

The one thing I would like to see before we merge this is avoiding adding a second tab for this feature.   Two tabs implies different data and not different views on the same data.

Instead we present this feature as a filtered view mode as opposed to the raw view of the output data.  For now we can add this toggle to the `Build` menu, but ideally we would next add a button bar with common use features like 'Build', 'Rebuild', 'Clear', and this 'Filter Output' toggle.

cra0zy:
> For now we can add this toggle to the Build menu, but ideally we would next add a button bar with common use features like 'Build', 'Rebuild', 'Clear', and this 'Filter Output' toggle.

I've made a filter output toggle in menu for this PR.

As for the button bar, we may as well create a proper toolbox while we are at it. I will setup a PR for it later, here is the list of stuff I would put on it(in order):
 - New
 - Open
 - Save
 <br> Separator
 - New Item
 - New Folder
 - Add Item
 - Add Folder
 <br> Separator
 - Build
 - Rebuild
 - Clean
 - Filter Output

What do you think?

cra0zy:
Can this be merged?

tomspilman:
> I've made a filter output toggle in menu for this PR.

So it is no longer presented as a tab right?

cra0zy:
> So it is no longer presented as a tab right?

Yes, it's now a feature that's toggled in the main menu.

tomspilman:
Ok...  lets merge then.

Hopefully we can get someone to mirror this feature on the WinForms side as well.

cra0zy:
@nkast Would you by any chance be interested in mirroring this change on WinForms?

nkast:
Sure, I can look at it in a few days.

nkast:
here, https://github.com/mono/MonoGame/pull/4272

Tools/Pipeline/Gtk/Widgets/BuildOutput.cs
```diff
@@ -0,0 +1,129 @@
...
+            lock (textView1.Buffer)
+            {
+                textView1.Buffer.Text += text + "\r\n";
+            }
+
+            if (string.IsNullOrWhiteSpace(text))
+                return;
+
+            if (text.StartsWith("Build "))
+                AddItem(icon_begin_end, text.Remove(text.Length - 1), "");
```
nkast: I see you use this to remove the period from the last build and then append the elapsed time.
It also removes one character from the first build message, instead of "10:23:34 PM" it shows  "10:23:34 P".
Instead I trimed the period only when I need to appent the time, https://github.com/mono/MonoGame/pull/4272/files#diff-70e4e9aa432ee996c57064ed582baed2R62

PR #4172, opened 2015-10-07
SpriteBatcher item pool

nkast:
1) Initialize the pool with new 'SpriteBatchItem' instances in the
constructor. Avoid allocation/GC during gameplay.
2) Remove line '_freeBatchItemQueue.Enqueue(item)' from the inner loop.
Instead of a Queue we pool items in a List<> and keep an index to the
next free item. After the loop we set the index to 0 to return all the
items to the pool.

     | XNA | MG DX | MG GL
--- | ---  | ---- | ---
*Original* Average Draw(ms): | 2,46571 | 1,27025 | 1,51469
*this PR* Average Draw(ms): | N/A | 1,21511 | 1,45861
Diff (%)  | N/A | 95.7% | 96.3% 

using https://github.com/danzel/DrawTextTest

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -355,8 +355,24 @@ namespace Microsoft.Xna.Framework.Graphics
 		{
 			var item = _batcher.CreateBatchItem();
 
-			item.Depth = depth;
 			item.Texture = texture;
+
+            // set SortKey based on SpriteSortMode.
+            switch ( _sortMode )
+            {
+                // Comparison of Texture objects.
+                case SpriteSortMode.Texture:
+                    item.SortKey = texture.SortingKey;
+                    break;
+                // Comparison of Depth
+                case SpriteSortMode.FrontToBack:
+                    item.SortKey = depth;
+                    break;
+                // Comparison of Depth in reverse
+                case SpriteSortMode.BackToFront:
+                    item.SortKey = -depth;
+                    break;
+            }
 
 			if (sourceRectangle.HasValue) {
 				_tempRect = sourceRectangle.Value;
@@ -393,7 +409,8 @@ namespace Microsoft.Xna.Framework.Graphics
 					(float)Math.Cos (rotation), 
 					color, 
 					_texCoordTL, 
-					_texCoordBR);			
+					_texCoordBR,
+                    depth);
 			
 			if (autoFlush)
 			{
--- a/MonoGame.Framework/Graphics/SpriteBatchItem.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatchItem.cs
@@ -2,10 +2,10 @@ using System;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
-	internal class SpriteBatchItem
+    internal class SpriteBatchItem : IComparable<SpriteBatchItem>
 	{
 		public Texture2D Texture;
-		public float Depth;
+        public float SortKey;
 
         public VertexPositionColorTexture vertexTL;
 		public VertexPositionColorTexture vertexTR;
@@ -19,70 +19,77 @@ namespace Microsoft.Xna.Framework.Graphics
             vertexBR = new VertexPositionColorTexture();            
 		}
 		
-		public void Set ( float x, float y, float w, float h, Color color, Vector2 texCoordTL, Vector2 texCoordBR )
+		public void Set ( float x, float y, float w, float h, Color color, Vector2 texCoordTL, Vector2 texCoordBR, float depth )
 		{
 			vertexTL.Position.X = x;
             vertexTL.Position.Y = y;
-            vertexTL.Position.Z = Depth;
+            vertexTL.Position.Z = depth;
             vertexTL.Color = color;
             vertexTL.TextureCoordinate.X = texCoordTL.X;
             vertexTL.TextureCoordinate.Y = texCoordTL.Y;            
 
 			vertexTR.Position.X = x+w;
             vertexTR.Position.Y = y;
-            vertexTR.Position.Z = Depth;
+            vertexTR.Position.Z = depth;
             vertexTR.Color = color;
             vertexTR.TextureCoordinate.X = texCoordBR.X;
             vertexTR.TextureCoordinate.Y = texCoordTL.Y;
 
 			vertexBL.Position.X = x;
             vertexBL.Position.Y = y+h;
-            vertexBL.Position.Z = Depth;
+            vertexBL.Position.Z = depth;
             vertexBL.Color = color;
 			vertexBL.TextureCoordinate.X = texCoordTL.X;
             vertexBL.TextureCoordinate.Y = texCoordBR.Y;
 
 			vertexBR.Position.X = x+w;
             vertexBR.Position.Y = y+h;
-            vertexBR.Position.Z = Depth;
+            vertexBR.Position.Z = depth;
             vertexBR.Color = color;
 			vertexBR.TextureCoordinate.X = texCoordBR.X;
             vertexBR.TextureCoordinate.Y = texCoordBR.Y;
 		}
 
-		public void Set ( float x, float y, float dx, float dy, float w, float h, float sin, float cos, Color color, Vector2 texCoordTL, Vector2 texCoordBR )
+		public void Set ( float x, float y, float dx, float dy, float w, float h, float sin, float cos, Color color, Vector2 texCoordTL, Vector2 texCoordBR, float depth )
 		{
             // TODO, Should we be just assigning the Depth Value to Z?
             // According to http://blogs.msdn.com/b/shawnhar/archive/2011/01/12/spritebatch-billboards-in-a-3d-world.aspx
             // We do.
 			vertexTL.Position.X = x+dx*cos-dy*sin;
             vertexTL.Position.Y = y+dx*sin+dy*cos;
-            vertexTL.Position.Z = Depth;
+            vertexTL.Position.Z = depth;
             vertexTL.Color = color;
             vertexTL.TextureCoordinate.X = texCoordTL.X;
             vertexTL.TextureCoordinate.Y = texCoordTL.Y;
 
 			vertexTR.Position.X = x+(dx+w)*cos-dy*sin;
             vertexTR.Position.Y = y+(dx+w)*sin+dy*cos;
-            vertexTR.Position.Z = Depth;
+            vertexTR.Position.Z = depth;
             vertexTR.Color = color;
             vertexTR.TextureCoordinate.X = texCoordBR.X;
             vertexTR.TextureCoordinate.Y = texCoordTL.Y;
 
 			vertexBL.Position.X = x+dx*cos-(dy+h)*sin;
             vertexBL.Position.Y = y+dx*sin+(dy+h)*cos;
-            vertexBL.Position.Z = Depth;
+            vertexBL.Position.Z = depth;
             vertexBL.Color = color;
             vertexBL.TextureCoordinate.X = texCoordTL.X;
             vertexBL.TextureCoordinate.Y = texCoordBR.Y;
 
 			vertexBR.Position.X = x+(dx+w)*cos-(dy+h)*sin;
             vertexBR.Position.Y = y+(dx+w)*sin+(dy+h)*cos;
-            vertexBR.Position.Z = Depth;
+            vertexBR.Position.Z = depth;
             vertexBR.Color = color;
             vertexBR.TextureCoordinate.X = texCoordBR.X;
             vertexBR.TextureCoordinate.Y = texCoordBR.Y;
 		}
-	}
+
+        #region Implement IComparable
+        public int CompareTo(SpriteBatchItem other)
+        {
+            return SortKey.CompareTo(other.SortKey);
+        }
+        #endregion
+    }
 }
 
--- a/MonoGame.Framework/Graphics/SpriteBatcher.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatcher.cs
@@ -73,15 +73,10 @@ namespace Microsoft.Xna.Framework.Graphics
         /// The list of batch items to process.
         /// </summary>
 	    private readonly List<SpriteBatchItem> _batchItemList;
-
         /// <summary>
-        /// The available SpriteBatchItem queue so that we reuse these objects when we can.
+        /// Index pointer to the next available SpriteBatchItem in _batchItemList.
         /// </summary>
-        private readonly List<SpriteBatchItem> _freeBatchItemPool;
-        /// <summary>
-        /// Index pointer to the next available SpriteBatchItem in _freeBatchItemPool.
-        /// </summary>
-        private int _freeBatchItemBase;
+        private int _batchItemCount;
         
         /// <summary>
         /// IComparer instance for SpriteSortMode.Texture.
@@ -113,11 +108,10 @@ namespace Microsoft.Xna.Framework.Graphics
             _device = device;
 
 			_batchItemList = new List<SpriteBatchItem>(InitialBatchSize);
-            _freeBatchItemPool = new List<SpriteBatchItem>(InitialBatchSize);
-            _freeBatchItemBase = 0;
+            _batchItemCount = 0;
 
             for (int i = 0; i < InitialBatchSize; i++)
-                _freeBatchItemPool.Add(new SpriteBatchItem());
+                _batchItemList.Add(new SpriteBatchItem());
 
             EnsureArrayCapacity(InitialBatchSize);
 		}
@@ -129,10 +123,9 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <returns></returns>
         public SpriteBatchItem CreateBatchItem()
         {
-            if (_freeBatchItemBase >= _freeBatchItemPool.Count)
-                _freeBatchItemPool.Add(new SpriteBatchItem());
-            var item = _freeBatchItemPool[_freeBatchItemBase++];
-            _batchItemList.Add(item);
+            if (_batchItemCount >= _batchItemList.Count)
+                _batchItemList.Add(new SpriteBatchItem());
+            var item = _batchItemList[_batchItemCount++];
             return item;
         }
 
@@ -190,26 +183,26 @@ namespace Microsoft.Xna.Framework.Graphics
         public void DrawBatch(SpriteSortMode sortMode, Effect effect)
 		{
 			// nothing to do
-			if ( _batchItemList.Count == 0 )
+            if (_batchItemCount == 0)
 				return;
 			
 			// sort the batch items
 			switch ( sortMode )
 			{
 			case SpriteSortMode.Texture :
-				_batchItemList.Sort( _compareTexture );
+				_batchItemList.Sort( 0, _batchItemCount, _compareTexture );
 				break;
 			case SpriteSortMode.FrontToBack :
-				_batchItemList.Sort ( _compareDepth );
+				_batchItemList.Sort ( 0, _batchItemCount, _compareDepth );
 				break;
 			case SpriteSortMode.BackToFront :
-				_batchItemList.Sort ( _compareReverseDepth );
+				_batchItemList.Sort ( 0, _batchItemCount, _compareReverseDepth );
 				break;
 			}
 
             // Determine how many iterations through the drawing code we need to make
             int batchIndex = 0;
-            int batchCount = _batchItemList.Count;
+            int batchCount = _batchItemCount;
 
             
             unchecked
@@ -261,9 +254,8 @@ namespace Microsoft.Xna.Framework.Graphics
                 // large batches
                 batchCount -= numBatchesToProcess;
             }
-			_batchItemList.Clear();
             // return items to the pool.  
-            _freeBatchItemBase = 0;
+            _batchItemCount = 0;
 		}
 
         /// <summary>
--- a/MonoGame.Framework/Graphics/SpriteBatcher.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatcher.cs
@@ -39,6 +39,7 @@
 // #endregion License
 // 
 
+using System;
 using System.Collections.Generic;
 
 namespace Microsoft.Xna.Framework.Graphics
@@ -72,24 +73,11 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <summary>
         /// The list of batch items to process.
         /// </summary>
-	    private readonly List<SpriteBatchItem> _batchItemList;
+	    private SpriteBatchItem[] _batchItemList;
         /// <summary>
         /// Index pointer to the next available SpriteBatchItem in _batchItemList.
         /// </summary>
         private int _batchItemCount;
-        
-        /// <summary>
-        /// IComparer instance for SpriteSortMode.Texture.
-        /// </summary>
-        private static CompareTexture _compareTexture = new CompareTexture();
-        /// <summary>
-        /// IComparer instance for SpriteSortMode.FrontToBack.
-        /// </summary>
-        private static CompareDepth _compareDepth = new CompareDepth();
-        /// <summary>
-        /// IComparer instance for SpriteSortMode.BackToFront.
-        /// </summary>
-        private static CompareReverseDepth _compareReverseDepth = new CompareReverseDepth();
         
         /// <summary>
         /// The target graphics device.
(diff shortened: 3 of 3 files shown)


tomspilman:
@nkast - An idea...  do we really need both `_freeBatchItemPool` and `_batchItemList`?  Wouldn't the `_freeBatchItemBase` also define the last used element in the batch?

Seems like maybe we can just keep one list and just keep track of the item "active count"/"free index".

nkast:
Yes, that would work too. The problem was with List.Sort(), the only method that takes a index,count pair requires an IComparer<T> instance. We need 3 new nested classes to implement CompareTexture / CompareDepth / CompareReverseDepth. Perhaps we can define then as static members of SpriteBatcher. Thoughts?

KonajuGames:
> Perhaps we can define then as static members of SpriteBatcher.

That should be fine because they won't store any data and shouldn't be affected if at any time we have multiple threads accessing it.

nkast:
Ok, I will test this out.

tomspilman:
According to this...

 http://stackoverflow.com/a/930783

`IComparable<T>` is the api we should use for the best sorting performance.

tomspilman:
In that same thread they mention that `Array.Sort` is faster than list sorts.  It might be something to consider in this process.

nkast:
OK, I will use https://github.com/danzel/DrawTextTest with SpriteSortMode.Texture to see how each case behave.

tomspilman:
Sounds great!

nkast:
`List.Sort( IComparable<T>)` is indeed faster. Place a stopwatch around Sort() 
Comparison<> 
sw 1,1772
sw 0,0014
sw 0,8751
sw 0,0014
sw 0,878
sw 0,0014
sw 0,8785
sw 0,0014

IComparer
sw 1,1356
sw 0,0014
sw 0,8232
sw 0,0009
sw 0,8184
sw 0,0014
sw 0,834
sw 0,0014

nkast:
Average Draw (in milliseconds) with  SpriteSortMode.Texture

     | XNA | MG DX 
--- | ---  | ---- | ---
Develop  | 2,42024 | 2,12712 (100%)
NoQueue | N/A | 1,97407 (92.8%)
NoQueue & IComparer | N/A | 1,91679 (90.1%)

So, we got 5% improvement on SpriteBatch, 10% when used with sorting.
Am I doing that right?

tomspilman:
So I guess NoQueue is a win because we touch less memory... no copying of pointers from one list to another on each quad, no second list to clear/iterate over, etc.  We just use the cache better.

You should test `IComparable<>` vs `IComparer<>` next... I suspect you'll get an even better win.

nkast:
Missed to lines. 8-(
I Guess I should redo the last benchmark!

@tomspilman 
To use Array.Sort() we have to grow the SpriteBatchItem[] array in our code. List manage this for us now.
How  IComparable will work for 3 types of comparison?

nkast:
ok, I was setting batchCount = _batchItemList.Count; which means sorting 256 items each time. 
The correct benchmark :

     | XNA | MG DX 
--- | ---  | ---- | ---
Develop  | 2,42024 | 2,12712 (100%)
NoQueue | N/A | 1,97407 (92.8%)
NoQueue & IComparer | N/A | 1,13672 (53,4%)
 
Not bad!

tomspilman:
> To use Array.Sort() we have to grow the SpriteBatchItem[] array in our code.

Yeah...  we would have to use `Array.Resize()` ourselves which isn't too bad, but maybe there really isn't much of a win there compared to `List<>`.

> How IComparable will work for 3 types of comparison?

Oh... yeah.  I had not considered that...  maybe there is some trick.  Or maybe we can't support `IComparable` then.

MonoGame.Framework/Graphics/SpriteBatcher.cs
```diff
@@ -108,12 +123,9 @@ public SpriteBatcher (GraphicsDevice device)
...
         {
-            SpriteBatchItem item;
-            if (_freeBatchItemQueue.Count > 0)
-                item = _freeBatchItemQueue.Dequeue();
-            else
-                item = new SpriteBatchItem();
-            _batchItemList.Add(item);
+            if (_batchItemCount >= _batchItemList.Count)
+                _batchItemList.Add(new SpriteBatchItem());
+            var item = _batchItemList[_batchItemCount++];
```
tomspilman: We should probably grow the batch items better here.  Like don't add one more item... that ensures every quad causes at least one new allocation.  Instead maybe we can do something like this:
```
if (_batchItemCount >= _batchItemList.Count) 
{
   do
  {
     _batchItemList.Add(new SpriteBatchItem());
  } while (batchItemList.Count < batchItemList.Capacity)
}
```
So basically we follow whatever heuristic `List<>` observes when growing itself to add new batch items.

nkast:
I did a quick test with Array.Sort. Didn't see any difference.
I could drop the avg time down to 1.87 by increasing the array just by 64 instead of going from 256->512 which is what List<> does, but that's really a one time thing at first frame.

MonoGame.Framework/Graphics/SpriteBatcher.cs
nkast: Ok, done. 
This of course affects the benchmark results because the first frame takes longer than before.

nkast: oh, so you use do {} while to grow the list.Capacity, that looks better.

nkast:
>Oh... yeah. I had not considered that... maybe there is some trick. Or maybe we can't support  IComparable  then.
Here's an idea,
In case of `SpriteSortMode.Texture` we set `SpriteBatchItem.Depth` with the `Texture.SortingKey`.
In case of `SpriteSortMode.BackToFront` we set `SpriteBatchItem.Depth` with *minus* `depth`.

totallyeviljake:
The internal nature of SpriteBatcher and SpriteBatchItem means that you can further optimize the SpriteBatch flow. For instance, SpriteBatch always creates SpriteBatchItem for a specific sorting mode that does not change once Begin() is called. This means you can create a specialized SpriteBatcher that inline sorts the SpriteBatchItem instances and do away with the post processing Sort() calls.

tomspilman:
> Here's an idea,

That sounds like it could work.  You rename `SpriteBatchItem.Depth` to `SpriteBatchItem.SortKey`.  We then put in the correct value based on what we need to sort by.

But before all that you should see how much of a win `IComparable<T>` actually is for us.

tomspilman:
> This means you can create a specialized SpriteBatcher that inline sorts the SpriteBatchItem instances and do away with the post processing Sort() calls.

I think because of how big a batch is and how much memory you would touch on each sorted insert... sorting at the end when you are about to traverse the whole list... is likely better performance wise.

nkast:
1) Strategy pattern for SpriteBatcher, definitely, if there is a need for it.
2) The theory says that Inserting an object every now and then in the right place when you have an already sorted array is faster that inserting it at the end and then sort the list again. In our case we build a new list from scratch and then sort it using quick sort (as mentioned in the link above).
Each insertion is O(1) * n items = O(n), sort is O(n*logn), so we have O(n+n*logn) complexity.

Building a Sorted list means you have to binarySearch the index which is O(logn) provided you can randomly access the list (List & BinaryTrees). Then insert the item, O(1) for LinkedLists & BinaryTrees for n items, alas O(n)  (insert is O(n) for Lists where you have to copy/move all the items one up, for n items is O(n*n) ). We have to exclude LinkedLists because insert time is O(n).

In theory Binary trees can get as good as what we have now, O(n*logn).
In practice however traversing and maintain the tree pointer would be much slower and require more memory. A sequential representation will have much more wasted memory and many checks for valid leafs.

nkast:
> But before all that you should see how much of a win  IComparable<T>  actually is for us.
Yes, because we will add one more switch to test sorting method & negating depth. Also comparing floats might be slower than comparing integers for textures. Also, the current benchmark is unsuitable, we have to test batching small number of items. Improving an O(n*logn) method while at the same time add more instractions to a O(n) method will inevitably be a win. The question is at what value of n we get a even.

tomspilman:
>  Also comparing floats might be slower than comparing integers for textures. 

We can always make the key a `ulong` and encode float depth into an integer.

nkast:
For this benchmark I first draw a frame to force the pool to grow/initialize. 


Average Draw (in milliseconds)

   | Deffered | SortTexture
--- | ---  | ---- | ---
IComparer: | 1,12614 | 1,93389 
IComparable: | 1,05191 | 1,52211 
Diff (%)  | 93.4% | 78.7%


   | Deffered | SortTexture
--- | ---  | ---- | ---
Develop: | 1,21205 | 2,09219 
IComparable: | 1,05191 | 1,52211 
Diff (%)  | 86.8% | 72.8%

Interesting is that the new code is faster for Deffered too, I was expecting to take a hit since we added a switch in SpriteBatch (extra comparison/branching) with no benefit from the faster Sorting. Perhaps using an Array instead of a List alone gives as that 93.4%.

MonoGame.Framework/Graphics/SpriteBatchItem.cs
```diff
@@ -2,10 +2,11 @@
 
 namespace Microsoft.Xna.Framework.Graphics
 {
-	internal class SpriteBatchItem
+    internal class SpriteBatchItem : IComparable<SpriteBatchItem>
 	{
 		public Texture2D Texture;
 		public float Depth;
```
tomspilman: Do we actually use `Depth` anywhere other than in the `Set` methods below?  Can we just pass it into the vertex setup methods and not store it?

tomspilman:
> Perhaps using an Array instead of a List alone gives as that 93.4%. 

That is probably true.  Remember `List<>` internally has to do a bunch of validation of parameters and state before it even messes with the Array.  Removing all those `if (blah) throw` preconditions is likely where the win is coming from.

MonoGame.Framework/Graphics/SpriteBatchItem.cs
nkast: Yeap, I removed that just now.

nkast:
@tomspilman Are there any final small fixes you would like me to push in this one ?
ex. add full stop to comments here 
https://github.com/mono/MonoGame/pull/4172/files#diff-bf9a3678ade2b898dde0ac4d0f4a83cdR367
https://github.com/mono/MonoGame/pull/4172/files#diff-bf9a3678ade2b898dde0ac4d0f4a83cdR371
remove `index` (it's always zero)
https://github.com/nkast/MonoGame/blob/tncSpriteBatcher_ItemPool/MonoGame.Framework/Graphics/SpriteBatcher.cs#L209

tomspilman:
Maybe just double check that the unit tests cover the code paths within the batcher appropriately and we're not missing any critical test.  Other than that i'm good to merge this.

Great optimization job... everyone will benefit from this.

nkast:
I run the tests locally before submitting and aditionally swaped the values of https://github.com/nkast/MonoGame/blob/tncSpriteBatcher_ItemPool/MonoGame.Framework/Graphics/SpriteBatch.cs#L364-L374 to make sure that all three sorting tests fire.

nkast:
Great job indeed!

dellis1972:
@tomspilman @nkast should the commits for the PR be squashed into one? I know we don't normally do that but given this is a important area it might be better to have all the changes in just one commit.

nkast:
@dellis1972 I have the minimum git knowledge, just the stuff to push changes, but I could give it a try if you point me to good documentation of what you mean by that. 
#YOLO

KonajuGames:
He means git rebase.  It is a git feature that I have still not got the hang of.  It is one of those things that some people just get, and others never get.

dellis1972:
this is what I tend to do ..  
on your branch
```
  git rebase -i HEAD~X
```
where X is the number of commits you want to squash.. The -i means interactive.
When you do this you will get a screen/file edit with data like
```
  pick 01d1124 some stuff
  pick 6340aaa some more stuff
  pick ebfd367 even more stuff
```
What we do is change pick to s (or squash) for the commits you want to .. squash :)  I ALWAYS leave the top on as pick, you will get a chance to edit the commit message later in the process.
```
  pick 01d1124 some stuff
  squash 6340aaa some more stuff
  squash ebfd367 even more stuff
```
save the file and exit the editor.. you will then get a change to edit the commit message.
Once the message is saved (and exit the editor) it will then squash everything. 

Once that is all done you can 
```
  git push --force <origin/branchname>
```
to push the changes up. 

This is all entirely optional, @tomspilman will have the final word of whether this needs to be done or not. :)

Additional info can be found at http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

mrhelmut:
If you are using a git GUI such as SourceTree, you can easily do that by right clicking on a commit and selecting "rebase". You'll be able to edit commit messages and squashing them by dragging one to another (or even re-ordering them if need be) or clicking "squash with previous".

MichaelDePiazzi:
I thought this might be worth sharing with you all...

I'm doing some list sorting in my game, so this PR piqued my interest in seeing if I could also improve sorting performance. I did some benchmarks of my own, and this is what I found: *(sorting 5 million objects with randomised sort keys)*
```
Array.Sort (IComparable): 2722ms
List.Sort  (IComparable): 2718ms
Array.Sort (IComparer):   3032ms
List.Sort  (IComparer):   3020ms
Array.Sort (Comparison):  3319ms
List.Sort  (Comparison):  3287ms
```
So my tests also found that IComparable is fastest. However, I found in my tests that the sort performance of an array versus a list is practically identical.

I'm not sure if this will affect the decision to use an array over a list in this PR, but it might be worth considering if the added complexity of managing an array is worth it.

tomspilman:
> However, I found in my tests that the sort performance of an array versus a list is practically identical.

That is what i expected...  internally i'm sure the list just calls the array sort.  The real benefit of Array is removing the extra call overhead on the other methods like Add/Remove on the `List<>` itself.

> I'm not sure if this will affect the decision to use an array over a list in this PR

 We already did it... and it worked out fine.

> the added complexity of managing an array is worth it.

 There really wasn't any extra complexity... maybe 3 or 4 more lines of code overall.

MichaelDePiazzi:
@tomspilman That makes sense. All good then :)

nkast:
> I found in my tests that the sort performance of an array versus a list is practically identical.

That's what I found too on earlier test when testing IComparer, no significant different outside the margin of error. An explanation could be that `List` keeps an array internally and `List.Sort()` likely access that list the same way we do now. 
The last test with Deffered (no sorting) shows that Arrays have a bit faster access time. That's a win for us in the inner loop and in `CreateBatchItem()`.


>  I'm not sure if this will affect the decision to use an array over a list in this PR, but it might be worth considering if the added complexity of managing an array is worth it.

The only overload of `List.Sort(..)` that takes an index/count pair force as to use `IComparer<T>`, we wanted the performance benefit of `IComparable` so there were no choice here of keep using Lists instead of Arrays.

MichaelDePiazzi:
> The last test with Deffered (no sorting) shows that Arrays have a bit faster access time. That's a win for us in the inner loop and in CreateBatchItem().

Cool, I'm looking forward to this PR being merged as it should give my game a little bit more of a performance boost. Nice work on this btw :)

> The only overload of List.Sort(..) that takes an index/count pair force as to use IComparer<T>, we wanted the performance benefit of IComparable so there were no choice here to keep using Lists.

Actually you can do this by using that same overload but passing a null comparer (https://msdn.microsoft.com/en-us/library/8ce6t5ad%28v=vs.110%29.aspx)

nkast:
>Cool, I'm looking forward to this PR being merged as it should give my game a little bit more of a performance boost. Nice work on this btw :)

Good & fun collaborate work / brainstoring first of all. I wouldn't image that we could push performance that far.

> Actually you can do this by using that same overload but passing a null comparer 
That one use the default `IComparer<>` you initialized the `List` with.  
https://msdn.microsoft.com/en-us/library/234b841s(v=vs.110).aspx

nkast:
@dellis1972 I clean up the history a bit from all unrelated commits.

tomspilman:
Great job @nkast .  Merging now.

MonoGame.Framework/Graphics/SpriteBatchItem.cs
```diff
@@ -19,70 +19,77 @@ public SpriteBatchItem ()
...
             vertexBR.Position.Y = y+h;
-            vertexBR.Position.Z = Depth;
+            vertexBR.Position.Z = depth;
             vertexBR.Color = color;
 			vertexBR.TextureCoordinate.X = texCoordBR.X;
             vertexBR.TextureCoordinate.Y = texCoordBR.Y;
 		}
 
-		public void Set ( float x, float y, float dx, float dy, float w, float h, float sin, float cos, Color color, Vector2 texCoordTL, Vector2 texCoordBR )
+		public void Set ( float x, float y, float dx, float dy, float w, float h, float sin, float cos, Color color, Vector2 texCoordTL, Vector2 texCoordBR, float depth )
```
tomspilman: @nkast - Do we know under what conditions do we call this second `Set`?  Could we have just one instead?

MonoGame.Framework/Graphics/SpriteBatchItem.cs
```diff
@@ -19,70 +19,77 @@ public SpriteBatchItem ()
             vertexBR = new VertexPositionColorTexture();            
 		}
 		
-		public void Set ( float x, float y, float w, float h, Color color, Vector2 texCoordTL, Vector2 texCoordBR )
+		public void Set ( float x, float y, float w, float h, Color color, Vector2 texCoordTL, Vector2 texCoordBR, float depth )
 		{
 			vertexTL.Position.X = x;
```
tomspilman: @nkast - I wonder if it is better to store the `Set` data in the `SpriteBatchItem` in a more efficient format making this structure much smaller.... then building the `VertexPositionColorTexture` at VB packing time.

MonoGame.Framework/Graphics/SpriteBatchItem.cs
nkast: I remember that there was no path leading to it. At least on Windows and WinRT projects.
If it's ok I can send a PR to remove it.

tomspilman: > If it's ok I can send a PR to remove it.

Sure... the build server will tell us then if it is truly unused.

MonoGame.Framework/Graphics/SpriteBatchItem.cs
nkast: Most cycles in the SpriteBatcher loop are spent to copy 4 `TextureColorTexture` vertices from `SpriteBatchItem` to the `vertexArray`. It 's about 96 bytes or so total.
https://github.com/nkast/MonoGame/blob/6121e4b6d13a835206d00a951587e12d06baec6d/MonoGame.Framework/Graphics/SpriteBatcher.cs#L234-L237 
By moving the calculation, the `Set` method will only have to copy around 40 bytes into the smaller `SpriteBatchItem`. That's where my bets are.
SpriteBatchItem are currently  The `Sort` moves object references (pointers) around which are the size of a register (32/64bit), but then, there are also cache misses when it access the `SortKey`, a smaller `SpriteBatchItem` will definitely help but I can't speculate on or measure  those kind of things, only profiling will show if it's significant. Allocating all items was definitely good, unless the CLR decide to move items around... Also Sorting will shuffle references around over time. The item at array[0] will not be the first item we allocated.




Another idea to remove those 4 lines from the loop was to have `Set` write directly into the `VertexArray`! The SpriteBatchItem will hold a reference to the buffer and an incrementing index. This is straightforward for deferred mode. In case of other type of sprite sorting we can use the BatchItem index to recreate a secondary indexBuffer. That's one memory read (index) and 6 writes of `short`. Still haven't work out the details (resizing, etc).

tomspilman: > The  Sort  moves object references (pointers) around which are the size of a register (32/64bit), but then, there are also cache misses when it access the  SortKey 

I wonder if adding a structure like:
```
struct SpriteBatchSortItem
{
   int index;
   float SortKey;
}
```
... then doing the sort on an array of that structure would be better for sort performance.  We can then also change `SpriteBatchItem` to a structure as we're not sorting them directly.

tomspilman:
@nkast - I was thinking the next thing we could do to improve `SpriteBatch` performance is to start using unsafe code in the inner loop.

nkast:
On some tests I've done previously I found  pointers to be faster that accessing an array because the clr does a bound check every time you access an array through []. Using pointers in a loop will give some more perf. Good to know you are ok with pointers because I found them necessary to mix dx with OpenGL code without moving everything away from the main code file. I still can't avoid to have code in the platform file calling back some method in the main file, which I haven't seen before in MonoGame (GraphicsDevice/Texture2D/etc) and I am afraid it might make code flow hard to follow&understand. That is MethodX() calling PlatformX() calling innerX().
I have a lot going on right now, I hope to find some time soon to test all the ideas currently on the table and do proper speed tests before moving in one or the other direction.

tomspilman:
@nkast - A first pass at pointers in `DrawBatch`:

https://gist.github.com/tomspilman/2fe345c7f97aa721f0ee

... it could probably be better if we could do a single memcpy of the item `vertexTL/TR/BL/BR` verts into the array.

KonajuGames:
Pointers are OK in controlled circumstances, such as the internals of
SpriteBatch where we have a stable working solution in safe code and want
to get more performance from it.

nkast:
@tomspilman 
One thing we need to be careful is if we increment the pointer (ex.`*(vertexPtr++)`) or just add a base offset ex.(`*(vertexPtr+1)` or `*(vertexPtr+(2+i*4))`)
Normally one would do the first but I believe C# likes to write the pointer value back to memory every time. 

One more thing with the initialization of index array.
https://github.com/nkast/MonoGame/blob/6121e4b6d13a835206d00a951587e12d06baec6d/MonoGame.Framework/Graphics/SpriteBatcher.cs#L159-L165
I found it to be twice as slow under x64, probably related to casting from `int` to `short`. By using pointers I manage to get better results (but I can't find the code, probably lost it).

nkast:
@tomspilman I verified that again, incrementing the pointer will have negative effect in perf. The code keep writing it to memory and reading it back. There is no way to force better optimization similar to C++ `register` keyword.
This is true on the MSIL level, on jitted x86/x64 (4.0,4.5,4.6.1) and on ARM (WP8.1). Not tested on RyuJIT or Net.Native. 
The only way from there would be native code or rewriting the MSIL in a way that also result in better Jitted code.

nkast:
A major problem with `Spritebatch`/`DrawUserPrimitives` is the performance of  `SharpDX.Utilities.Write()` / `cpblk`.
http://xoofx.com/blog/2010/10/23/high-performance-memcpy-gotchas-in-c/
https://connect.microsoft.com/VisualStudio/feedback/details/766977/il-bytecode-method-cpblk-badly-implemented-by-x86-clr

All my test so far was done on x64 but some -if not all- MG templates generate either a x86 project or have "prefer x86" checked.

nkast:
*Correction* 
In my previous tests I failed to enable optimization. 

> I found it to be twice as slow under x64, probably related to casting from int to short.

The index array is actually faster on x64 with optimization enabled. Still using pointers is faster on both cases, I updated the numbers on the related PR. 

> The code keep writing it to memory and reading it back. There is no way to force better optimization similar to C++ register keyword.

Actually the Jitter put the pointer in EDX and execute 'inc EDX' to increment it. 
The reason the form '*(indexPtr + 1)' is faster than '*(indexPtr++)' is because it uses memory addressing arithmetic, which translates to 'mov word ptr[EDX+2],CX' which is cost-free on modern CPUs. A single 'add EDX,0Ch' at the end is better than 12 'inc EDX'.

tomspilman:
>  Hyper-V seems to affect this behavior.

You should avoid Hyper-V for any performance testing...  it does weird things at times.

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -355,9 +355,25 @@ void CheckValid(SpriteFont spriteFont, StringBuilder text)
...
+                case SpriteSortMode.FrontToBack:
+                    item.SortKey = depth;
+                    break;
+                // Comparison of Depth in reverse
+                case SpriteSortMode.BackToFront:
+                    item.SortKey = -depth;
+                    break;
+            }
+
 			if (sourceRectangle.HasValue) {
```
tomspilman: @nkast - I just noticed another small optimization here.  We can skip the `tempRect` and a bunch of math in the case where there isn't a `sourceRectangle`:

```
if (sourceRectangle.HasValue) {
	_tempRect = sourceRectangle.Value;
	_texCoordTL.X = _tempRect.X / (float)texture.Width;
	_texCoordTL.Y = _tempRect.Y / (float)texture.Height;
	_texCoordBR.X = (_tempRect.X + _tempRect.Width) / (float)texture.Width;
	_texCoordBR.Y = (_tempRect.Y + _tempRect.Height) / (float)texture.Height;
} else {
	_texCoordTL.X = 0;
	_texCoordTL.Y = 0;
	_texCoordBR.X = 1.0f;
	_texCoordBR.Y = 1.0f;
}
```

nkast: Good one. here is the PR #4825

Issue #4174, opened 2015-10-09
Latest commit has missing implementation for GamePad.UWP.PlatformGetMaxIndex()


LoveDuckie:
The most recently made commit by @tomspilman introduces an issue whereby the implementation for PlatformGetMaxIndex() is missing from the UWP partial class implementation of GamePad. Leads to compilation errors.

LoveDuckie:
FYI: https://github.com/mono/MonoGame/commit/a10b8094b0e44a8401639b843ae77cd8fff5e7a5

danzel:
This can be closed now @tomspilman

PR #4175, opened 2015-10-09
Added missing method in Gamepad.UWP.cs

cra0zy:
#4174

### DIFF
--- a/MonoGame.Framework/Input/GamePad.UWP.cs
+++ b/MonoGame.Framework/Input/GamePad.UWP.cs
@@ -11,6 +11,11 @@ namespace Microsoft.Xna.Framework.Input
     {
         internal static bool Back;
 
+        private static int PlatformGetMaxIndex()
+        {
+            return 16;
+        }
+
         private static GamePadCapabilities PlatformGetCapabilities(int index)
         {
             if (index >= WGI.Gamepad.Gamepads.Count)


tomspilman:
How is it that this missing method didn't cause a build failure before?

cra0zy:
If you look at my branch before you will notice that it was made before I made changes to UWP Gamepad and therefor this class didn't exist at that point.

tomspilman:
>  it was made before I made changes to UWP Gamepad 

Ah... that would do it.

MonoGame.Framework/Input/GamePad.UWP.cs
```diff
@@ -11,6 +11,11 @@ static partial class GamePad
     {
         internal static bool Back;
 
+        private static int PlatformGetMaxIndex()
+        {
+            return 16;
```
tomspilman: Can you return `WGI.Gamepad.Gamepads.Count` here instead.

MonoGame.Framework/Input/GamePad.UWP.cs
```diff
@@ -11,6 +11,11 @@ static partial class GamePad
...
+        }
+
         private static GamePadCapabilities PlatformGetCapabilities(int index)
         {
             if (index >= WGI.Gamepad.Gamepads.Count)
```
tomspilman: Also might be good to remove these index checks in the platform methods here.  Let the assert at the shared code level deal with the bounds check.

cra0zy: Bad idead, it would cause code like:
```c#
GamePad.GetState(PlayerIndex.Two)
```
to cause a crash when only 1 GamePad is connected.

MonoGame.Framework/Input/GamePad.UWP.cs
cra0zy: As bellow, it would cause a crash to code like:
```c#
GamePad.GetState(PlayerIndex.Two)
```
when one gamepad is connected.

tomspilman: I just confused the controller index test (done at the common GamePad.cs level) with a test for gamepad being connected or not (currently done within the extended GamePad.PlatformXXX methods).  They are sort of similar... but not really the same.

It would be nice to move the disconnected gamepad path up into the common GamePad.cs code, but that is for a separate PR.

danzel:
FYI @cra0zy if you say "Fixed #...." instead it will auto close that issue when the PR is merged.

cra0zy:
> FYI @cra0zy if you say "Fixes #...." instead it will auto close that issue when the PR is merged.

Did not know that, thanks for the info!!!

PR #4179, opened 2015-10-14
Add the MonoGame.framework.pkg to the Mac Installer.

dellis1972:
Forgot to add the Framework to the installer distribution list


KonajuGames:
Merging. Good one on picking up the omission.

Issue #4119, opened 2015-09-02
AudioSessionException on iOS 8.4


ebatiano:
When returning from background, sometimes AudioToolbox throws an exception. Here's stack trace (sorry for lack of lines and file names - data is from BugSense):

```
Exeption: AudioSessionException (Unknown error code: 561015905)
Function: AudioToolbox.AudioSession.SetActive

at AudioToolbox.AudioSession.SetActive (Boolean active) [0x00000] in <filename unknown>:0 
at Microsoft.Xna.Framework.Audio.OpenALSoundController.<OpenSoundController>m__1 (System.Object sender, System.EventArgs e) [0x00000] in <filename unknown>:0 
at AudioToolbox.AudioSession.Interruption (IntPtr userData, UInt32 state) [0x00000] in <filename unknown>:0 
at (wrapper native-to-managed) AudioToolbox.AudioSession:Interruption (intptr,uint) 
```

Eversor:
Seems fixed on ios9, i have no crashes on that platform for 5 days.

vchelaru:
While this is fixed in iOS 9, it still occurs in iOS 7 and 8, which has a large install base for a while. Any info on why this might be happening?

vchelaru:
I've found somewhat of a workaround. The Crash occurs in the AudioSession.SetActive(true) call, as indicated by the previous post. 

I played around with it a bit and found out that if I try/catch and consume the exception, the app will continue, but no more audio plays:

                AudioSession.Resumed += (sender, e) => {
                    try
                    {
                        AudioSession.SetActive(true);
                    }
                    catch(Exception exce)
                    {
                 }

Of course that's only a half-solution because it does mean the game is silent until the user relaunches the app. I continued to mess around with it and I found that if I got a crash, I could detect that and delay the re-activation of the Audio Session as follows:

                AudioSession.Resumed += (sender, e) => {
                    try
                    {
                        AudioSession.SetActive(true);
                    }
                    catch(Exception exce)
                    {
                        try
                        {
                            // maybe we can delay it?
                            System.Threading.Thread.Sleep(1000);
                            AudioSession.SetActive(true);
                        }
                        catch(Exception exce2)
                        {
                            System.Console.WriteLine(exce2.ToString());
                        }
                    }
                    Alc.MakeContextCurrent(_context);
                    Alc.ProcessContext(_context);
                };

[...]

PR #4180, opened 2015-10-15
Updated the iOS Audio Handling to fix #3652

### LINKED ISSUE #3652: iOS freeze & crash when attempting to resume following incoming call
When returning to a monogame game having received a call, the game screen displays, but updates do not resume. After about 10 seconds, the game crashes and returns to the home screen.

This occurs in the latest stable monogame build and development versions from just before this point until now.

Steps to reproduce:
1) Create a new iOS Universal app in Xamarin studio
2) Set up a standard "cornflower blue" game1 boilerplate class and update the AppDelegate and Main.cs files to run this.
3) Modify the game1 update and draw calls so that the ongoing updates can be seen (I put in a simple timer to display gameTime.TotalGameTime on the draw call)
4) Run the app (I was using an iPhone 5c running iOS 8.1)
5) Call the phone running the app
6) When the phone running the app switches to display the incoming call, hang up
7) The phone running the app will now return to the app, but the display will no longer update. After about 10 seconds, the app will crash.

dellis1972:
We were using some obsolete API's to handle
Audio Interruptions.

Can someone give this a test on a iOS 9 device?

### DIFF
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -22,6 +22,7 @@ using Android.Media;
 
 #if IOS
 using AudioToolbox;
+using AVFoundation;
 #endif
 
 namespace Microsoft.Xna.Framework.Audio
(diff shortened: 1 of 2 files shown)


danzel:
Maybe related to #4119 ?
I'll try test at the weekend.

dellis1972:
@prollin not sure... I just updated the existing code to use the new event handler.

dellis1972:
All fixed up :)

tomspilman:
Looks to be failing on Protobuild stuff...

fail

dellis1972:
@tomspilman this can still be merged into develop, no need to cherry-pick it seems :)

kjpou1:
I have tested this out and all seems to build and work.  Not sure what the protobuild error would be but wanted to confirm that it builds.

tomspilman:
Well it concerns me that possibly after merging this all other PRs will fail too.  So we need to get to the bottom of this first.

dellis1972:
@tomspilman There is nothing in this PR which would explain the error... can you kick the PR build off again? See if it works this time around?

dellis1972:
@tomspilman  actually... I'll check the branch out and rebase on the latest develop them push.. that should kick the build off again

tomspilman:
Just did... still failing.

@hach-que -  Any idea what this could be triggering this error in Protobuild?

https://cloud.githubusercontent.com/assets/103522/11605863/7fd88a3e-9ad0-11e5-9e77-c7d74e2dd7cd.PNG

dellis1972:
@tomspilman  it passed, quick merge it before the universe changes its mind :P

tomspilman:
Weird...  here we go.

hach-que:
@tomspilman I'm not sure your build server is configured to clean the working directory before or after builds, so the build artifacts from one PR might influence another.

This means that changes in PRs like #4295 might be persisting in other PRs, which is very undesirable.

You need to configure the build server to be doing `git clean -xdff` before a PR build so that everything is as it should be without any side-effect artifacts.

hach-que:
(It also means that technically it's not an issue with any individual PR, but rather that they're interfering with one another because of build artifacts being present)

tomspilman:
> I'm not sure your build server is configured to clean the working directory before or after builds

It resets all tracked files to their repo state.  It does not delete any output artifacts.

If and output artifact is breaking the build we need to fix whatever is generating that artifact.  We don't want local builds for example to fail because some process is leaking breaking garbage out.

hach-que:
It's not that, it's that I can add files or Git submodules in one PR and
those won't be removed when the next completely different PR builds.

Not cleaning artifacts is fine when the builds are descendants of one
another, but not when they're unrelated, because resetting to HEAD does not
remove added files or submodules.

PR #4181, opened 2015-10-16
Fix for WP81 template: GamePage is not calling InitializeComponent

Helikin:
When InitializeComponent() is not called, the template can still be used to create a running game but InputDialog.ResizeLayoutRoot() will throw a NullReferenceException (because the Grid in the GamePage cannot be found).

The other templates do not have this problem.


tomspilman:
Thanks @Helikin !

PR #4182, opened 2015-10-16
Fix: VertexStride not set in VertexContent if model contains only POSITION vertex channel.

Helikin:


tomspilman:
Good catch... merging!

PR #4183, opened 2015-10-16
Matrix.Invert(Matrix matrix) fix

ka1eka:
I had a weird bug with Xamarin.iOS while Debug build was running fine, but Release build had a corrupted view matrix.

### DIFF
--- a/MonoGame.Framework/Matrix.cs
+++ b/MonoGame.Framework/Matrix.cs
@@ -1691,8 +1691,9 @@ namespace Microsoft.Xna.Framework
         /// <returns>The inverted matrix.</returns>
         public static Matrix Invert(Matrix matrix)
         {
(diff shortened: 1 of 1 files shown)


tomspilman:
The temporary here shouldn't be required as the invert function internally creates temporaries for all values.  Can you tell us more about the corruption you were getting in iOS?

KonajuGames:
I think it might be a Xamarin codegen or optimization issue.  I have seen a bug on Linux with the Rectangle constructor where it wouldn't pass the height parameter correctly in certain circumstances.  The simple work-around might seem unnecessary, but a similar fix worked for my Rectangle issue.

tomspilman:
I'm going to go ahead and merge this...  the change shouldn't hurt anything and pass the same value as `ref` and `out` is a bit weird.

PR #4184, opened 2015-10-16
Fix: SpriteBatch writes wrong z value.

Helikin:
The default transformation matrix of the SpriteBatch creates wrong z values. This is only a problem if sprites are rendered with enabled depth test. 
If we switch the near and far plane of the transformation matrix, we get the same results as XNA.


nkast:
Do we have to swap the order of sorting as well?

nkast:
When drawing text/GUI (spritebatch) on top of 3D models I have to clear the depth, otherwise the 2D textures appear behind the scene. With XNA I never had to. 
Is it because until now  MG draw everything at the farPlane/maxDepth and this fix will draw Textures a the nearplane/minDepth/0?

nkast:
I tested your code and it works. No need to clear the depthbuffer anymore to draw text on top of 3D models. Windows Desktop DX & Lumia 620.

Helikin:
> Do we have to swap the order of sorting as well?

No, I have not seen a problem with sprite sorting on the CPU. Only z buffer values are affected.

> I tested your code and it works. No need to clear the depthbuffer anymore to draw text on top of 3D models.

Thanks for testing. That makes sense. Without this fix, sprites with the default layer depth (0) are drawn on the far plane (z = 1). With the fix they are drawn on the near plane (z = 0).

flibitijibibo:
I came across the same issue with Rex Rocket's sprite rendering, so this should work with OpenGL as well. The resulting Matrix should look like this:

https://github.com/flibitijibibo/FNA/blob/master/src/Graphics/SpriteBatch.cs#L1078

Helikin:
I have just verified it: The code of this PR creates the same matrix as the FNA code.

KonajuGames:
It looks like you've covered all bases quite well.  I'm happy to merge this.  Thanks @Helikin.

Issue #4033, opened 2015-07-25
DirectX and OcclusionQuery


Skeatwin:
I try to do the lens flare effect in my game (see http://xbox.create.msdn.com/en-US/education/catalog/sample/lens_flare), and that use the OcclusionQuery class.
And after long tests, I found OcclusionQuery dosn't work on DirectX platform, the result of PixelCount is always 0 and the IsComplete property is always set to False.

I hope that you can help me, thanks :smile: 

Here is a part of the code:

if (this.occlusionQueryActive == true)
            {
                // If the previous query has not yet completed, wait until it does.
                if (!this.occlusionQuery.IsComplete)
                    return;

                this.occlusionAlpha = Math.Min(occlusionQuery.PixelCount / OCCLUSION_QUERY_AREA, 1f);
            }

            // Set renderstates for drawing the occlusion query geometry. We want depth
            // tests enabled, but depth writes disabled, and we disable color writes
            // to prevent this query polygon actually showing up on the screen.
            this.GraphicsDevice.BlendState = BlendStateExtended.ColorWriteNone;
            this.GraphicsDevice.DepthStencilState = DepthStencilState.DepthRead;

            // Set up our BasicEffect to center on the current 2D light position.
            this.basicEffect.World = Matrix.CreateTranslation(lightPosition.X, lightPosition.Y, 0);
            this.basicEffect.Projection = Matrix.CreateOrthographicOffCenter(0, viewport.Width, viewport.Height, 0, 0, 1);

            this.basicEffect.CurrentTechnique.Passes[0].Apply();
            this.occlusionQuery.Begin();
            this.GraphicsDevice.DrawUserPrimitives(PrimitiveType.TriangleStrip, occlusionQueryVertices, 0, 2);
            this.occlusionQuery.End();
            this.occlusionQueryActive = true;

Vodacek:
yes, this is because OcclusionQuery is implemented only at OpenGL platforms

Skeatwin:
Thank for your reply I will waiting the implementation :)

ShyroFR:
Awesome, thanks!

Skeatwin:
Great work ! thanks Monogame ! :)

Issue #414, opened 2012-04-04
Missing OcclusionQuery class from Graphics namespace


CartBlanche:
Someone on twitter who tried to port his app to MonoGame pointed out that we don't have the OcclusionQuery class implemented....

http://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.occlusionquery.aspx

Some quick Googling last night shows the OpenGL also has that ability as well and the OpenTK site mentions it here

http://www.opentk.com/doc/chapter/2/opengl/fragment-ops/occlusion-query

Now it seems OcclusionQuery it only valid for the Desktop stack as it is part of OpenGL, but NOT OpenGL ES.

Any suggestions on how we can get it working across all our platforms? Or is it just not worth it due to over heads involved if it is not supported by the mobile hardware?

tomspilman:
We still need someone to implement this for GL platforms... at least for desktops.

Let's also require we add some unit tests for this as well.

Jjagg:
@tomspilman This was already implemented for OpenGL, which is why it isn't in the title of Helikins PR I suppose. He added tests in that PR as well :) 

This is not supported in GLES 2.0, but there is an [extension] we could possibly check. If we'd check for that and add a bool to GraphicsDeviceCapabilities if it's not available, we could add this to iOS and Android too.

tomspilman:
Thanks for letting us know @Jjagg .  This can be closed now I think.

PR #4188, opened 2015-10-19
Added OcclusionQuery implementation for DirectX.

Helikin:
This PR splits the OcclusionQuery.cs into OcclusionQuery.*Platform*.cs files and adds an implementation for DirectX.

The PR also adds unit tests. One test uses the SpriteBatch. It should currently fail because of the SpriteBatch z value bug. It should work after following PR is merged:
- Fix: SpriteBatch writes wrong z value. #4184

Previously, OcclusionQuery.cs was excluded from iOS, Android and Ouya builds. I have tested the occlusion query on iOS and Android, and the query works if the device supports OpenGL ES 3.0. I have not tested on devices without ES 3.0 support - therefore, the OcclusionQuery files are still excluded on iOS, Android and Ouya.

This PR should also fix a bug in the OpenGL implementation: 
The pixel count was retrieved with ```QueryResultAvailable``` (which returns only true/false) instead of ```QueryResult```. (The MonoMac code path did not have this problem.)

Related issues: 
- Missing OcclusionQuery class from Graphics namespace #414
- DirectX and OcclusionQuery #4033


Helikin:
With #4184 merged, the unit tests should succeed now. Is it possible to restart the build bot?

tomspilman:
@Helikin - You need to merge the latest upstream changes into your branch to include #4184.  The build server just builds the branch... it doesn't merge with develop for you.

Helikin:
> it doesn't merge with develop for you.

Ah yeah, right. I spend too little time with git. ;-)
Will do tomorrow.

tomspilman:
Thanks @Helikin ... merging!

Issue #1001, opened 2012-11-17
Rendertargets are lacking mipmapping functionality


sthoahmi:
Currently, rendertargets do not support the creation of mipmaps. 

https://github.com/mono/MonoGame/blob/develop3d/MonoGame.Framework/Graphics/GraphicsDevice.cs#L1412 

Therefore calls like: 

RenderTarget2D r = new RenderTarget2D(graphicsDevice, w, h, mip, SurfaceFormat.Color, DepthFormat.None); 

throw an exception whenever mip = true.

CartBlanche:
@tomspilman @KonajuGames @harry-cpp @Wowo51 With mobiles phones now being MUCH more powerful, should we look at adding this in across the board?

tomspilman:
Mip creation for a render target is done by the GPU, but the C# code needs to trigger it when someone requests a mip level other than 0.

It can be done on any platform... just gets more costly.

If we want to do it in a way that is reusable on all platforms:

 - When a mip other than 0 is needed.
 - Check if the mip 0 is dirty (keep track if we rendered to it).
 - If so copy mip 0 down to all other mips one at a time using a custom shader.
 
If we had compute we could accelerate it some there, but for now this is the best approach.

PR #4189, opened 2015-10-19
Added mipmap generation for render targets (DirectX).

Helikin:
Mipmaps for render targets are created in GraphicsDevice.PlatformResolveRenderTargets (similar to the already functioning mipmap generation in the OpenGL code). 

I have added a unit test which is currently set to Ignore because:
In my tests generated mipmaps are different in XNA and MonoGame. The XNA mipmaps look wrong, as if created with point sampling!? MonoGame mipmaps look correct. 
Tested mipmaps in DirectX/OpenGL and Windows/MacOS are identical but only if the texture size is a power of two. 

Related issues:
- Rendertargets are lacking mipmapping functionality #1001


Helikin:
There is one problem which I have observed:

If mipmaps are created for a cube map render target, then the mips of the last (6th) side of the cube map can contain uninitialized blocks. 

Here is a cube map visualized on a sphere. The problem are the magenta colored blocks.
image
These garbage blocks change randomly after each mip generation.

The problem was only observed with AMD GPUs (AMD R9 285, Win10). There are no problems on Intel and NVIDIA GPUs.

My first idea was that GenerateMips is called to early and some conditions have to be met before GenerateMips can be called. However, I have not found a solution yet. 

At the moment I guess this is an AMD driver problem (which would not be the first...)

tomspilman:
> In my tests generated mipmaps are different in XNA and MonoGame. 

 I suggest then adding the MonoGame image as the comparison image and leaving the test on.  Then at least we'll get a report if the feature breaks for any reason.


> The XNA mipmaps look wrong, as if created with point sampling!?

Weird... must be a card/driver specific thing.


>  the mips of the last (6th) side of the cube map can contain uninitialized blocks. 

Cubemap mipmap generation has always been sort of buggy.

I suspect the only way to ensure mipmap generation is consistent across different hardware is to generate them yourself via a custom shader.  In fact i expect that both DX12 and Vulkan do not contain a "generate mipmaps" type function.

tomspilman:
Let's try to get this merged for 3.6.

tomspilman:
@Helikin - Can you rebase this for merging?

Helikin:
Will do, later today.

Helikin:
@tomspilman I have rebased and combined commits to 1 commit.

tomspilman:
> I have rebased and combined commits to 1 commit. 

Note in the future you don't need to squash down to one commit...  

 https://github.com/blog/2141-squash-your-commits

So we can do that here from the PR merge button moving forward.

tomspilman:
Merging... thanks @Helikin !

Helikin:
> Note in the future you don't need to squash down to one commit...

Good to know. Thanks for merging!

Issue #4190, opened 2015-10-19
mgcb broken relative paths


cra0zy:
mgcb doesn't calculate relative paths correctly, in the example below it decided to completely skip over "GitHub/DarkWasTheNight/Content" part of the path 

Command:

```
mono "/usr/lib/mono/xbuild/MonoGame/v3.0/Tools/MGCB.exe" /@:"/home/harry/GitHub/DarkWasTheNight/Content/Content.mgcb" /platform:DesktopGL /outputDir:"/home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL" /intermediateDir:"/home/harry/GitHub/DarkWasTheNight/Content/obj/DesktopGL" /quiet
```

Output:

```
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Fonts/CC_Red_Alert_12.xnb
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Fonts/CC_Red_Alert_36.xnb
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Textures/blank.xnb
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Textures/stickman.xnb
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Textures/Tile0.xnb
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Textures/Tile1.xnb
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Textures/Windows_BSOD.xnb
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Fonts/LevelTitle.xnb
Cleaning /home/harry/GitHub/DarkWasTheNight/Content/bin/DesktopGL/GitHub/DarkWasTheNight/Content/Textures/stickman-64.xnb
[...]

PR #4191, opened 2015-10-19
Mgcb relative path fix when calling with /@:

cra0zy:
#4190

### DIFF
--- a/Tools/MGCB/CommandLineParser.cs
+++ b/Tools/MGCB/CommandLineParser.cs
@@ -284,6 +284,8 @@ namespace MGCB
                 if (arg.StartsWith("/@:") || arg.StartsWith("-@:"))
                 {
                     var file = arg.Substring(3);
+                    Directory.SetCurrentDirectory(Path.GetDirectoryName(file));
+
                     var commands = File.ReadAllLines(file);
                     var offset = 0;
                     lines.Insert(0, string.Format("# Begin:{0} ", file));


Tools/MGCB/CommandLineParser.cs
```diff
@@ -284,6 +284,8 @@ private IEnumerable<string> Preprocess(IEnumerable<string> args)
                 if (arg.StartsWith("/@:") || arg.StartsWith("-@:"))
                 {
                     var file = arg.Substring(3);
+                    Directory.SetCurrentDirectory(Path.GetDirectoryName(file));
```
tomspilman: I'm nervous at this sort of change.  It has the potential to break every other file operation.

First why was this not a problem before?  We've been using MGCB on Mac/Linux for months now...  at least a year on Windows.  What is different about your case and what happens to all the previously working cases?

cra0zy: Pipeline sets working directory in: https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Common/PipelineController.cs#L475 so it wasn't noticeable with it.

Plus it's not like this is changing much with this, working directory can still be overwritten with "workingDir" argument.

Also I've actually known about this for about 9 months or so... http://community.monogame.net/t/starting-directory-for-mgcb-when-loading-response-file-wrong/1608 I just forgot about it till now....

KonajuGames: I added the workingDir argument a while back so I could debug from MonoDevelop, which has no working directory option in its debug settings.

The usual mode of operation when running MGCB from the command line has been to change directory to the location of the .mgcb file, but setting the current directory from the location of the .mgcb file would make the debugging process easier in MonoDevelop (no need for the workingDir argument).

jamesford42: I think the correct fix is to expand all path's to absolute when the preprocessor expands a response file. This fixes your issue without messing with the working directory for anyone downstream. Also note that restoring the working directory after the response file could work, so long as it is done in a stack so response files with response files don't break.

PR #3276, opened 2014-12-16
Implemented GraphicsDevice.SetVertexBuffers(...)

jamesford42:
This is a re-write of https://github.com/mono/MonoGame/pull/1767/. I say re-write because GraphicsDevice.cs has changed so much since that PR that I had to handle each changed block manually.

Issue #914, opened 2012-10-29
GraphicsDevice.SetVertexBuffers API is not yet implemented


CartBlanche:
As per subject line. Just adding this here, so we know it needs looking at.

tomspilman:
This is fixed now.

Vodacek:
@tomspilman I am afraid that it is not, SetVertexBuffers is not implemented on DesktopGL platform

Issue #913, opened 2012-10-29
GraphicsDevice.DrawInstancedPrimitives API not yet implemented in GL.


CartBlanche:
As Per subject line. Just adding this here so we know it needs implementing.

gruntbatch:
Out of curiosity, should this issue be marked as closed? It looks like #4195 may have solved both this issue and #914

mgarstenauer:
#4195 only provides the DirectX implementation. OpenGL implementation is still missing. 
(--> We can remove the DirectX label.)

tomspilman:
@mgarstenauer - Done and edited the title.

PR #4195, opened 2015-10-20
Implemented DrawInstancedPrimitives(), SetVertexBuffers() and improved input layout caching for DirectX.

Helikin:
This PR adds the DirectX implementation for **DrawInstancedPrimitives** and **SetVertexBuffers**. The hashing and caching of vertex input layouts is also fixed/improved.

Before writing this code, we have reviewed the related issues for these topics:
- GraphicsDevice.DrawInstancedPrimitives API not yet implemented. #913
- GraphicsDevice.SetVertexBuffers API is not yet implemented #914
- Implemented GraphicsDevice.SetVertexBuffers(...) #3276

This implementation is used successfully in the DigitalRune Engine for mesh instancing and morph target animation. We have also added some simple unit tests.

There is one important change regarding the vertex attribute semantics **SV_Position** and **POSITION** in shaders:
The recommended semantic for the vertex input element which describes the vertex position is **POSITION**. However, the XNA stock effects use **SV_Position**. This is allowed in DirectX but it is better to use **SV_Position** ("system-value position") only for the vertex shader output and pixel shader input. 
This is important when using morph targets. Each morph target mesh has a **POSITION0** element. At runtime the morph target vertex streams are matched and combined to a single input layout with **POSITION0/1/2/3**. This is not possible with **SV_Position** because **SV_Position1/2/3** is not a valid semantic.
Therefore, we have changed the C# code and shaders to use **POSITION** for vertex inputs. 
The new code contains a fallback to handle shaders with **SV_Position**, but we recommend to only use **POSITION** for vertex shader input for all future MonoGame shaders.

The vertex input type for BLENDINDICES has been changed from int4 to uint4. This prevents SharpDX warnings when run on debug device. (The problem is that the MonoGame content pipeline outputs Byte4 for BLENDINDICES. MonoGame does not have a signed integer format that matches int4.)

This PR also fixes a minor bug in the GraphicsDevice OpenGL code: The flag **_vertexBufferDirty** was never reset to false.


Vodacek:
Good job! It will be awesome to have this merged soon.

Discosultan:
This is awesome! Thanks a ton for this PR.

tomspilman:
This looks good on my first pass thru it.

@KonajuGames  @tgjones  - Can you guys take a look thru this and report anything you see.

tgjones:
This is great - good job, @Helikin, this is a very high quality pull request.

I'm particularly happy to see my pet peeve, the `SV_Position` input semantic thing, being fixed - that's bothered me for a while :)

I haven't run the new tests against XNA, but I assume they pass?

Helikin:
@tgjones Thanks for the feedback. The majority of the code was written by @mgarstenauer. So he deserves most of the praise.

All tests pass on my test machine and as I can see on the builds server. To check for any sideeffects on the OpenGL builds, I have also run the Linux build (in Win 10 not in Linux) and the Mac build on my Mac.

KonajuGames:
You have both done well on this one.  It's a big addition that we have needed for a long time.

KonajuGames:
I'll merge this now.

Helikin:
@tomspilman @tgjones @KonajuGames Thanks for reviewing and the quick merge.

kvenkateshan:
Helikin, any chance you are also working on bringing this over to PS4? We have major features relating to instanced based rendering and are curious to know about any timelines around this feature making it over to PS4. Thanks for your help!

tomspilman:
DX instancing is a bit different than on PS4, but some of this work should help get things started there.

kvenkateshan:
Where would I find the documentation for the GL instancing calls? PS4 sdk? If there's no movement on this by end of this year, I will have to locally take a crack at it. Any info you can give me is highly appreciated.

tomspilman:
> Where would I find the documentation for the GL instancing calls?

I don't know much at all about the GL side.  For sure someone needs to address that.  But this has nothing to do with PS4.  You should look thru the PS4 sdk for instancing.

kvenkateshan:
Well, I will started digging into this and will probably open a new thread to ask for public help if I am blocked on specific bits.

tomspilman:
> ask for public help if I am blocked on specific bits.

Remember if you want to talk about PS4 specifically it can't be here in public.  It has to happen on the private repo for it.

kvenkateshan:
Got it, thanks!

pzgulyas:
One think is missing from SetVertexBuffers() implementation: When XNA is fed with a list of VertexBuffer-s, they get implicitly converted to VertexBufferBinding-s, while MonoGame doesn't do this. 

Example:
``graphicsDevice.SetVertexBuffers(vertexBuffer1, vertexBuffer2);`` works with XNA, but fails with MonoGame.

mgarstenauer:
@pzgulyas That's true. But I don't miss that feature. You should always pass your own `VertexBufferBinding[]` when calling `SetVertexBuffers` and reuse the array. Otherwise, .NET allocates a new array each time, which creates unnecessary garbage.

tomspilman:
> Otherwise, .NET allocates a new array each time, which creates unnecessary garbage

I agree.  When we can avoid adding an API that requires garbage generation we should.

I wonder if we could simply add overloads for 2, 3, and 4 VBs which internally use a scratch array.  Gives you a similar behavior without any garbage.

mgarstenauer:
Yes, new method overloads are a possible solution. However, for morphing I need up to 6 VBs (base mesh + 5 morph targets). So I would need overloads for 2, 3, ... 6 VBs, which sounds a bit much.

SeriousMartin:
Awesome! Thanks a lot!

PR #4196, opened 2015-10-20
Fixed #4192 "Copy Task does not support copying directories" UWP build issue

### LINKED ISSUE #4192: "Copy Task does not support copying directories" UWP build issue
Trying to port a project over to a new UWP project and getting some real troubles with the latest 3.5 builds (3.4 does not work anymore at all)

When trying to build the project, I am getting the following error:
"Severity	Code	Description	Project	File	Line
Error		The source file "C:\Development\Sandbox\MonoGameUWP1\MonoGameUWP1\Content\bin\WindowsStoreApp\" is actually a directory.  The "Copy" task does not support copying directories.	MonoGameUWP1"

Which seems to indicate a fault in the MGCB build process for copying the content folder.  Note, my content folder is actually blank (new project).

Also a related issue, is that the MG project is not usable when you create it fresh, as you have to close and reopen it else loads of references do not show up.

dellis1972:
Added Conditions to handle the case where there are no content files.
Also added a conditional so we don't error
if not .mgcb is provided (we do issue a warning though)

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
@@ -25,6 +25,12 @@
       <ParentIntermediateDir Condition=" '$(ParentIntermediateDir)' == '' " >$(ProjectDir)$(ContentRootDirectory)\obj\$(MonoGamePlatform)</ParentIntermediateDir>
 
       <MonoGameContentBuilderExe Condition="'$(MonoGameContentBuilderExe)' == ''">$(MSBuildThisFileDirectory)Tools\MGCB.exe</MonoGameContentBuilderExe>
+      <MonoGameContentBuilderCmd>&quot;$(MonoGameContentBuilderExe)&quot;</MonoGameContentBuilderCmd>
+      <MonoGameContentBuilderCmd Condition=" '$(OS)' != 'Windows_NT' ">mono $(MonoGameContentBuilderCmd)</MonoGameContentBuilderCmd>
+
+      <PlatformResourcePrefix Condition="'$(MonoGamePlatform)' == 'MacOSX' Or '$(MonoGamePlatform)' == 'iOS'">Resources\</PlatformResourcePrefix>
+      <PlatformResourcePrefix Condition="'$(MonoGamePlatform)' == 'Android'">Assets\</PlatformResourcePrefix>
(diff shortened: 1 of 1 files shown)


dellis1972:
@tomspilman This should fix #4192

DDReaper:
Local test works as expected. Project with empty MGCB file builds and runs successfuly

tomspilman:
@dellis1972 

I wonder if a cleaner way to do this would be to define more properties then adding more conditions to the later task.  For instance defining like:

```
<MonoGameContentBuilderCmd>&quot;$(MonoGameContentBuilderExe)&quot;</MonoGameContentBuilderCmd>
<MonoGameContentBuilderCmd Condition=" '$(OS)' != 'Windows_NT' ">mono $(MonoGameContentBuilderCmd)</MonoGameContentBuilderCmd>
```
You can then use `$(MonoGameContentBuilderCmd)` later in the script and not worry about platform.

Same with the content copy...  we can add a new property that is overloaded for these other platforms instead of duplicating the copy code 3 times.

dellis1972:
@tomspilman I can look into that no problem. Did you want to merge this now to fix the issue and I'll clean it up in a new PR?

tomspilman:
I'd rather see this fixed once and cleanly... we can wait a little bit if we have to.

dellis1972:
@tomspilman hows that? Works fine on OS X.. @DDReaper any chance you can test on Windows?

tomspilman:
@dellis1972 - I mean i'll wait for you to make the change then i'll merge it.

dellis1972:
@tomspilman I reworked the code a bit, its much cleaner now :) We still need conditionals because we cannot use a $(property) for the ItemName="Foo" attribute on the CreateItem Output.. But I think this code is much cleaner and nicer :D

tomspilman:
> We still need conditionals because we cannot use a $(property) for the ItemName="Foo" attribute on the CreateItem Output

Lame...    takes something that could be really clean and kills it.

Still much better than before.

DDReaper:
Right, will re-test again shortly.
Just finishing up the .NET UAP fix

DDReaper:
Confirmed working.
Created a new blank project with an empty MGCB file, builds and runs fine.  Much better

tomspilman:
@DDReaper - And if you delete the mgcb file and/or remove it from the project?

DDReaper:
Ohh, wasn't expecting a 2 for 1.
Yes, instead of horrible error, we now get a nice and friendly message.  Still builds without errors and runs fine.
+1Million response.

tomspilman:
Great... merging!

PR #4197, opened 2015-10-21
Removed GamePadThumbSticks.Gate property

### LINKED ISSUE #4170: Thumbstick inputs being clipped when using circular deadzone
I've been working on improving and fine-tuning controller input for my game recently, and I've noticed when using the circular deadzone setting that the thumbsticks feel a bit too touchy.

After doing some investigation, I found that the method being used in MonoGame seems to be incorrect. Here is a graph showing what it's currently doing. The colours represent the resulting thumbstick vector length: Blue = 0 (i.e. the deadzone), Green = Normal range, Yellow = 1 (i.e. full movement)
circular_deadzone_with_circular_gate
So you can see from the big thick yellow area that the values are being clipped. This results in a halfway movement of the thumbstick being equivalent to a full movement. This is not what I'd expect. And looking at Shawn Hargreaves blog article on this subject (http://blogs.msdn.com/b/shawnhar/archive/2007/03/28/gamepads-suck.aspx), the graphs he provided don't seem to show any clipping being done like this. So I'm assuming this behaviour is actually incorrect.

This is my expectation of what the graph should actually look like (which is in line with Shawn Hargreaves graph):
new_method_with_circular_gate

Unless anyone has reason to believe I'm wrong about this, I'm planning to submit a PR soon which corrects this.

Also on a related note - While looking into this, I came across the GamePadThumbSticks.Gate property. I can't seem to find this in XNA, so I assume it's been added to MonoGame for some reason? I'm not sure exactly why this setting is needed, but the default for it is GateType.Round (which normalises the thumbstick vector if it's length is greater than 1). So if I set it to GateType.None, then the above graph now looks like this:
circular_deadzone_with_no_gate
Red means that the vector length exceeds the expected maximum of 1. So basically GateType.Round was hiding these out of range values coming from the circular deadzone function.

Does anyone know why this Gate property was added to MonoGame? It seems unnecessary as the deadzone setting could be controlling this (ie. GamePadDeadZone.IndependentAxes => GateType.Square,  GamePadDeadZone.Circular => GateType.Round, GamePadDeadZone.None => GateType.None). I have a funny feeling that it might have been added to correct these out of range values, so it might be unnecessary if this was the reason.

MichaelDePiazzi:
Closes #4170

This PR removes the superfluous `GamePadThumbSticks.Gate` property as discussed in issue #4170. This property did not exist in XNA, and it's default setting (i.e. round gate type) only matched XNA behaviour when using a circular dead zone.

I've now changed clamping of thumbstick values to be in line with XNA. From my tests and observations, XNA applies circular clamping when using a circular dead zone (i.e. clamps the vector length to 1). In all other cases (including the public constructor), it applies square clamping (i.e. clamps the vector within bounds [-1,-1]..[1,1]).

I've also added some unit tests for the public constructor to test that values are being clamped as expected (which fail with the previous MonoGame code due to the incorrect default clamping).

### DIFF
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -92,6 +92,7 @@
 	<Compile Include="Framework\CurveTest.cs" />
 	<Compile Include="Framework\EnumConformingTest.cs" />
     <Compile Include="Framework\GamePadStateTest.cs" />
+    <Compile Include="Framework\GamePadThumbSticksTest.cs" />
     <Compile Include="Framework\GameTest.cs" />	
     <Compile Include="Framework\GameTest+Methods.cs" />
     <Compile Include="Framework\GameTest+Properties.cs" />
--- a/MonoGame.Framework/Input/GamePadThumbSticks.cs
+++ b/MonoGame.Framework/Input/GamePadThumbSticks.cs
@@ -45,14 +45,6 @@ namespace Microsoft.Xna.Framework.Input
 {
     public struct GamePadThumbSticks
     {
-        public enum GateType
-        {
-            None,
-            Round,
-            Square
-        };
-        public static GateType Gate = GateType.Round;
-
         Vector2 left;
         Vector2 right;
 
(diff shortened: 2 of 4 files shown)


tomspilman:
Thanks @MichaelDePiazzi !

MichaelDePiazzi:
@tomspilman Cool, thanks Tom :)

PR #4198, opened 2015-10-21
Fix: Asset was built multiple times with same settings.

Helikin:
This PR solves two problems:

- _Problem 1: "An asset is built multiple times, even though the processor parameters are identical."_
When PipelineBuildEvents are compared, the method PipelineBuildEvents.AreParametersEqual assumes that the same parameters are stored in both dictionaries. This is not true in practice.
Example: One PipelineBuildEvent stores the parameter ("ColorKeyEnabled", true). The second PipelineBuildEvent does not have a matching entry. The two PipelineBuildEvents might still be equivalent. We need to compare the parameter with the default value of the content processor.
	
- _Problem 2: "The order of the assets in the content project should not matter, but it does matter."_
Example: ModelA, ModelB, TextureC are included in the content project. Both models reference TextureC.
Currently when the order in the content project is ModelA, ModelB, TextureC, the TextureC is output as "Texture_0.xnb" and "Texture.xnb". If the order is TextureC, ModelA, ModelB, the output is only "Texture.xnb", which is in in my opinion the correct result.
To solve the problem, all asset that need to be registered as PipelineBuildEvents before building any assets. Then it is possible to correctly resolve external references.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
@@ -48,6 +48,12 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
         // with different parameters.)
         private readonly Dictionary<string, List<PipelineBuildEvent>> _pipelineBuildEvents;
 
+        // Store default values for content processor parameters. (Necessary to compare processor
+        // parameters. See PipelineBuildEvent.AreParametersEqual.)
+        //   Key = name of content processor
+        //   Value = processor parameters
+        private readonly Dictionary<string, OpaqueDataDictionary> _processorDefaultValues;
+
         public string ProjectDirectory { get; private set; }
         public string OutputDirectory { get; private set; }
         public string IntermediateDirectory { get; private set; }
@@ -88,6 +94,7 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
         public PipelineManager(string projectDir, string outputDir, string intermediateDir)
         {
             _pipelineBuildEvents = new Dictionary<string, List<PipelineBuildEvent>>();
+            _processorDefaultValues = new Dictionary<string, OpaqueDataDictionary>();
             RethrowExceptions = true;
 
             Assemblies = new List<string>();
@@ -377,6 +384,47 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
             return processor;
         }
 
+        /// <summary>
+        /// Gets the default values for the content processor parameters.
+        /// </summary>
+        /// <param name="processorName">The name of the content processor.</param>
+        /// <returns>
+        /// A dictionary containing the default value for each parameter. Returns
+        /// <see langword="null"/> if the content processor has not been created yet.
+        /// </returns>
+        public OpaqueDataDictionary GetProcessorDefaultValues(string processorName)
+        {
+            // null is not allowed as key in dictionary.
+            if (processorName == null)
+                processorName = string.Empty;
+
+            OpaqueDataDictionary defaultValues;
+            if (!_processorDefaultValues.TryGetValue(processorName, out defaultValues))
+            {
+                // Create the content processor instance and read the default values.
+                defaultValues = new OpaqueDataDictionary();
+                var processorType = GetProcessorType(processorName);
+                if (processorType != null)
+                {
+                    try
+                    {
+                        var processor = (IContentProcessor)Activator.CreateInstance(processorType);
+                        var properties = processorType.GetProperties(BindingFlags.FlattenHierarchy | BindingFlags.Public | BindingFlags.Instance);
+                        foreach (var property in properties)
+                            defaultValues.Add(property.Name, property.GetValue(processor, null));
+                    }
+                    catch
+                    {
+                        // Ignore exception. Will be handled in ProcessContent.
+                    }
+                }
+
+                _processorDefaultValues.Add(processorName, defaultValues);
+            }
+
+            return defaultValues;
+        }
+
         public DateTime GetProcessorAssemblyTimestamp(string name)
         {
             if (_processors == null)
@@ -458,6 +506,26 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
             return PipelineBuildEvent.Load(eventFilepath);
         }
 
+        public void RegisterContent(string sourceFilepath, string outputFilepath = null, string importerName = null, string processorName = null, OpaqueDataDictionary processorParameters = null)
+        {
+            sourceFilepath = PathHelper.Normalize(sourceFilepath);
+            ResolveOutputFilepath(sourceFilepath, ref outputFilepath);
+
+            ResolveImporterAndProcessor(sourceFilepath, ref importerName, ref processorName);
+
+            var contentEvent = new PipelineBuildEvent
+            {
+                SourceFile = sourceFilepath,
+                DestFile = outputFilepath,
+                Importer = importerName,
+                Processor = processorName,
+                Parameters = ValidateProcessorParameters(processorName, processorParameters),
+            };
+
+            // Register pipeline build event. (Required to correctly resolve external dependencies.)
+            TrackPipelineBuildEvent(contentEvent);
+        }
+
         public PipelineBuildEvent BuildContent(string sourceFilepath, string outputFilepath = null, string importerName = null, string processorName = null, OpaqueDataDictionary processorParameters = null)
         {
             sourceFilepath = PathHelper.Normalize(sourceFilepath);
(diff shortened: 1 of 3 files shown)


tomspilman:
> We need to compare the parameter with the default value of the content processor

Yes... we do.  Good catch!


> all asset that need to be registered as PipelineBuildEvents before building any assets. 

That seems reasonable to me.

Tools/MGCB/BuildContent.cs
```diff
@@ -284,6 +284,20 @@ public void Build(out int successCount, out int errorCount)
             errorCount = 0;
             successCount = 0;
 
+            // Before building the content, register all files to be built. (Necessary to
+            // correctly resolve external references.)
+            foreach (var c in _content)
```
tomspilman: Should we also maybe sort all the content by source files before this?

MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
```diff
@@ -458,6 +506,26 @@ private PipelineBuildEvent LoadBuildEvent(string destFile, out string eventFilep
...
+            {
+                SourceFile = sourceFilepath,
+                DestFile = outputFilepath,
+                Importer = importerName,
+                Processor = processorName,
+                Parameters = ValidateProcessorParameters(processorName, processorParameters),
+            };
+
+            // Register pipeline build event. (Required to correctly resolve external dependencies.)
+            TrackPipelineBuildEvent(contentEvent);
```
tomspilman: I suspect some processors/importers internally generate new content items to process.  How do those work out as far as consistency of processing order?

Tools/MGCB/BuildContent.cs
mgarstenauer: I think the source files should be processed in the original order, i.e. the order of the commands in the .mgcb file. 

The order of the content affects the name resolution for external references (TextureC.png -> TextureC_0.xnb, TextureC_1.xnb). The build order should be consistent between builds. When a new content is added, it is appended at the end of the .mgcb files and built last. Content that is already built is not affected (partial build).
However, when we sort the content, adding a new content changes the build order, which may change the names for external references (TextureC_0.xnb <-> TextureC_1.xnb). This *could* be a problem for partial builds. I'm not sure. I recommend leaving as is.

MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
mgarstenauer: I don't think that content importers can generate new content items.

Content processors can create new content items by creating an `ExternalReference<T>` and calling `ContentProcessorContext.BuildAsset` or `ContentProcessorContext.BuildAndLoadAsset`. This is the case that is currently causing duplicate assets and is fixed by this PR.

KonajuGames:
Merging.  Thanks @mgarstenauer.

mgarstenauer:
Thanks for merging.

nkast:
There is another bug that causes assets to build multiple times. The case is when you have a model with multiple meshes. 
The check at https://github.com/DigitalRune/MonoGame/blob/da07992b75176f033fe2808861d8e14be7f8f894/MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs#L253 fails because  parameter0 has a key `PremultiplyTextureAlpha` while defaultValues has a `PremultiplyAlpha`. 
Some issue with how ModelProcessor pass parameters to TextureProcessor perhaps. Still can spot it.

Any ideas @mgarstenauer ?

The file ball.fbx has 3 meshes all mapping to BallUV.png
mgcbmultiplemeshes

mgarstenauer:
No idea, but I will take a look.

mgarstenauer:
The bug is here: https://github.com/mono/MonoGame/blob/89f7c3f2deab37d13b3c2a89b5f602f6e69964f5/MonoGame.Framework.Content.Pipeline/Processors/MaterialProcessor.cs#L131-L132

The MaterialProcessor prepares the parameters for the TextureProcessor. The strings "PremultiplyTextureAlpha" and "ResizeTexturesToPowerOfTwo" are incorrect. The properties in TextureProcessor are called "PremultiplyAlpha" and "ResizeToPowerOfTwo".

nkast:
Thanks @mgarstenauer ! 
I tested it and it works.

I am getting 2 "Skipping P:/TestPipeline/BallUV.png", I don't know if mgcb should issue those warning in the first place since the texture comes from the same .fbx but for now is good.

I could submit a PR on this unless you want to, after all the credits is all yours.

mgarstenauer:
Since you already have it working and tested, please submit the PR.

> I am getting 2 "Skipping P:/TestPipeline/BallUV.png", I don't know if mgcb should issue those warning in the first place since the texture comes from the same .fbx but for now is good.

A while back, I submitted a PR (#3848) to suppress duplicate "Skipping ..." messages. The new PR (#4198) has effectively canceled the old change. In the new version it is no longer that simple to suppress these messages. IMHO the "Skipping ..." messages are just noise. I say we should simply remove them!

nkast:
PR #4324

tomspilman:
>  IMHO the "Skipping ..." messages are just noise. I say we should simply remove them!

Not a bad idea....  really they only exist because XNA would output the same lines when it detected a file didn't require reprocessing.

A C++ compiler doesn't emit "skipping... this cpp file is already up to date" or anything.  So maybe this is fine to remove.

If no one really objects to this... I could see it being done.

@cra0zy  @dellis1972  @KonajuGames ?

cra0zy:
No objections from me.

KonajuGames:
I agree with removing it.

nkast:
Note that the new filter view will hide issues like those.

nkast:
I'm worried that the output of building a single file might be confusing.

tomspilman:
> I'm worried that the output of building a single file might be confusing.

Yeah...  if it resulted in zero output feedback it would be weird.  What is the current output on building a single file?

nkast:
I will test it as soon as I get to my desktop. I'd like to see how the new filter view behaves as well.

nkast:
The filtered view parser can't distinguish between skipped primary textures and skipped external resources.

skippingexternalresourses


Building a single asset result to:
```
Build started 17/12/2015 12:39:50 μμ

Skipping P:/TestPipeline/BallUV.png

Build 1 succeeded, 0 failed.

Time elapsed 00:00:00.20. 
```

If we remove 'Skipping' the output will become:
```
Build started 17/12/2015 12:39:50 μμ


Build 1 succeeded, 0 failed.

Time elapsed 00:00:00.20. 
```

nkast:
The filter could save the first "Building $filename" and filter any future 'Skipping' messages with the same message.
On the MGBC a similar approach will work. There is a difference between Skipping because file was already build previously (asset is up to date) and skipping because file was already build in the current run. For example A /Rebuild shouldn't produce any 'Skipping' messages, correct?
We can keep a global list of build resources and suppress 'Skipping' warnings.

nkast:
PR #4330 , fix for multiple skipping warnings.

tomspilman:
Two things.

First I wonder if we could use indent to make clear messages coming from dependencies like the case of FBX referencing a texture.

Second maybe we should leave the messages alone...  let them be verbose.  Instead lets make the filter view better by like actually supporting filtering... by having an option to hide "skip" messages.

Thoughts?

nkast:
> First I wonder if we could use indent to make clear messages coming from dependencies like the case of FBX referencing a texture.
Sure https://github.com/mono/MonoGame/pull/4336

nkast:
> Second maybe we should leave the messages alone... let them be verbose. Instead lets make the filter view better by like actually supporting filtering... by having an option to hide "skip" messages.

The indent will make it possible to assign external resources to the parent asset and everything will be hidden inside the TreeViewItem so it's not a big problem to me. The issue is only with models that have multiple meshes. See for example what happens when GenerateTangentFrames=true.

```
Build started 18/12/2015 1:21:09 πμ

Cleaning P:/TestPipeline/bin/Windows/BallUV_0.xnb
Cleaning P:/TestPipeline/bin/Windows/Ball.xnb
P:/TestPipeline/Ball.fbx
	P:/TestPipeline/BallUV.png
	Skipping P:/TestPipeline/BallUV.png
	Skipping P:/TestPipeline/BallUV.png
Generating tangent frames.
Generating tangent frames.
Generating tangent frames.

Build 1 succeeded, 0 failed.

Time elapsed 00:00:01.08.
```

tomspilman:
> The issue is only with models that have multiple meshes and has to be handled on a per model base. See for example what happens when GenerateTangentFrames=true.

That should be fixable.  Is that `Generating tangent frames.` string not being written using the logger?  Or is your logic for indents not correctly indenting it?

IMO we need to start/end the indent within this function:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs#L622

So maybe something like this:

```
public object ProcessContent(PipelineBuildEvent pipelineEvent) 
{   
    _logger.Indent();
    try
    {
        InnerProcessContent(pipelineEvent);
    }
    finally
    {
        _logger.Unindent();
    }    
}
```

You then need to extend `PipelineBuildLogger` used here:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs#L102

To include our helper indent methods.

That should ensure all messaging during processing is indented.

nkast:
> Is that  Generating tangent frames.  string not being written using the logger? Or is your logic for indents not correctly indenting it?

#4336 uses the filename stack count in ContentBuildLogger to indent messages. The "Generating tangent frames." comes from the primary asset so it has no indent. Indeed, It doesn't look pretty,
we need to explicitly control indent as you proposed.

PR #4203, opened 2015-10-25
[project.json] Fix VS2015 compile problems.  The issue is that even w…

kjpou1:
…hen compiling non Windows Universal projects NuGet still tries to installs the .net Core 4.5 references from the project.json.  This gives the following error: Microsoft.NuGet.targets(89,5): error : Couldn't find the required information in the lock file. Make sure you have .NETCore,Version=v4.5.1 mentioned in your targets.

Moving the project.json to be project specific solves this without installing any extra tools or updates.  See (NuGet issue 1102)[https://github.com/NuGet/Home/issues/1102]

This is tested on a 5 day old VS 2015 RTM fresh install which still gives this error even after installing the windows tools 1.1 update.  The tools update did not fix the problems but this does.


kjpou1:
Closing as it is a duplicate of https://github.com/mono/MonoGame/pull/4163/

PR #4204, opened 2015-10-26
Fixes floating point precision edge case error with Lerp

hede:
According to wikipedia, the current implementation of MathHelper.Lerp suffers from a floating point precision issue.
(https://en.wikipedia.org/wiki/Linear_interpolation#Programming_language_support)

There's also a relevant SO discussion on http://stackoverflow.com/questions/4353525/floating-point-linear-interpolation#answer-23716956

Admittedly an edge case, but the fix seems to have negligible performance impact.

I couldn't run the test suite, so I created a simple console app to showcase the issue (and to see if the perf was impacted):
https://gist.github.com/hede/66914c6ebafd5a6c1c03

P.S. Why does Vector2.Lerp call MathHelper.Lerp separately on X and Y? Is it perf? We have the Vector operator overloads in place. Implementing it with vector operations might be easier to parse mentally IMHO


tomspilman:
@KonajuGames @dellis1972 

So this brings up the question...  fast or precise?

Precision errors suck...   working on something with huge scales this becomes very apparent.  Not a common use case for MonoGame, but it does happen.  Probably not as bad in lerp as in matrix transforms and the like, but still a problem.

The cost to fix that here is an additional multiply per lerp.  So 2 additional multiplies for Vector2, 3 for Vector3, etc.  This could be mitigated some for Vector2/3/4 by unrolling the lerp... but it will still be more expensive.

Is there a better way to resolve this?  Like `MathHelper.Lerp()` and `MathHelper.LerpPrecise()`?

Might be good to deduce if XNA indeed used the efficient method vs. the precise method.

KonajuGames:
I never realised there was a floating point error in this.  It makes sense to use the moire accurate version.  If we were to include both, I would suggest the other way around.  Have the default be more accurate and add a `MathHelper.FastLerp()` or similar that is less accurate.  That way it is an intentional decision to gain a slight performance benefit and accept it might be less accurate.

And yes, it would be good to determine what method XNA used.

MichaelDePiazzi:
XNA must use the efficient method as `MathHelper.Lerp(10000000000000000000f, 1f, 1f)` returns 0 instead of 1.

tomspilman:
> XNA must use the efficient method 

That is what I suspected...  which suggests we should leave `MathHelper.Lerp` using the fast method.  This is important because if not we can break subtle behaviors in games.

@KonajuGames ?

hede:
Sorry for disappearing.

I'm a bit conflicted too
From an API design perspective, it would be nice to have the precise implementation as the default.
Although, not taking a perf hit **and** not breaking existing games is pretty sweet too :)
So what's the verdict? @tomspilman @KonajuGames 
Can this be merged, or should I close this and come back with MathHelper.LerpPrecise() implementation?

WRT to the precise implementation (caveat: I know next to nothing about this stuff):
Can we SIMDfy the the operation in a xplat way?
Mono seems to have System.Numerics support.
Does that mean it will run everywhere (minus the web maybe)?
Also, I find the corefx implementation a bit confusing, if they were concerned about efficiency couldn't they have just used the SIMD path? They can check if(Vector.IsHardwareAccelerated). They do it two methods above in the same file (https://github.com/dotnet/corefx/blob/master/src/System.Numerics.Vectors/src/System/Numerics/Vector2.cs#L216)

KonajuGames:
If XNA uses the inaccurate version for Lerp, we should use that and add a LerpPrecise() method.

As for SIMD, none of them support anything other than SSE yet.  The check for `Vector.IsHardwareAccelerated` is specially handled by the latest JIT on .NET that removes the checks for hardware acceleration and compiles only the SSE version.

tomspilman:
> If XNA uses the inaccurate version for Lerp, we should use that and add a LerpPrecise() method.

I agree. Also @hede make sure you add reference docs to the `LerpPrecise()` that make it clear that it is slightly slower, but returns more accurate results for a lerp ranges of 1000.0f (or whatever appropriate estimate) or more.

dellis1972:
Just a thought would it be nice to provide

```
public static float Lerp(float value1, float value2, float amount)
public static float Lerp(float value1, float value2, float amount, bool precise = false)
```

as helper methods as well as LerpPrecise ().

tomspilman:
> Just a thought would it be nice to provide

I don't see the benefit there...  just to avoid typing 'Precise'?

KonajuGames:
There's an article we've posted before about why bool is not good as a
parameter type.

Using that approach, you would also be passing the cost of an if condition
every time you call Lerp. It's a small cost, but still avoidable.

PR #4206, opened 2015-10-27
Support for playing a song from a given start position.

danzel:
Implemented and tested on Android, iOS, WMS (Windows).

The WMS version was the trickiest, if you specify the start position immediately when you try ```_session.Start``` it doesn't work, you need to wait until the Topology is ready before you can play from anywhere that isn't the beginning. But you need to call ```_session.Start``` before it will prepare the topology. (At least in my testing)

@MichaelDePiazzi - Since you did a big bunch of WMS work in #4090 would you please be able to check over the WMS changes? Thanks!

I have a feeling I'll need to implement this for WP, WME, NVorbis(?) platforms too. Will look at that later in the week.

### DIFF
--- a/MonoGame.Framework/Media/MediaPlayer.WMS.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.WMS.cs
@@ -15,7 +15,12 @@ namespace Microsoft.Xna.Framework.Media
         private static MediaSession _session;
         private static AudioStreamVolume _volumeController;
         private static PresentationClock _clock;
+
         private static Song _nextSong;
+        private static TimeSpan? _nextSongStartPosition;
+        private static Variant? _desiredPosition;
+
+
         private static Song _currentSong;
 
         private enum SessionState { Stopped, Stopping, Started, Paused, Ended }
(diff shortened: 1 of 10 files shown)


MichaelDePiazzi:
@danzel Sure, I can take a look at your WMS changes. I'll try get to it in the next couple of days.

cra0zy:
@danzel I got you covered for NVorbis, once I get home I'll set it up.

danzel:
Awesome thanks both :)

cra0zy:
Done: 
https://github.com/cra0zy/MonoGame/tree/music-play-position
https://github.com/cra0zy/MonoGame/commit/1a7858c61188ba506a612e8cc377482eefb22a02

danzel:
Thanks @cra0zy, have pulled that in.

MonoGame.Framework/Media/MediaPlayer.WMS.cs
```diff
@@ -260,11 +275,16 @@ private static void OnSessionStopped()
...
                 else
-                    StartSession(PositionBeginning);
+                    StartSession(_nextSongStartPosition.HasValue ? PositionVariantFor(_nextSongStartPosition.Value) : PositionBeginning);
                 _nextSong = null;
             }
         }
+
+        private static Variant PositionVariantFor(TimeSpan position)
+        {
+            return new Variant { Value = (long)(10000000L * (position.TotalSeconds)) };
```
MichaelDePiazzi: The `Value` calculation can be simplified to: `return new Variant { Value = position.Ticks };`
And you can simplify things even further by moving the null check into `PositionVariantFor` instead. e.g.:
```c#
private static Variant PositionVariantFor(TimeSpan? position)
{
    return position.HasValue ? new Variant { Value = position.Value.Ticks } : PositionBeginning
}
```
That way you don't need to do this check each time you call this method.

MonoGame.Framework/Media/MediaPlayer.WMS.cs
```diff
@@ -231,6 +240,12 @@ private static void OnTopologyReady()
             _volumeController = CppObject.FromPointer<AudioStreamVolume>(volumeObjectPtr);
 
             SetChannelVolumes();
+
+            if (_desiredPosition.HasValue)
+            {
+                StartSession(_desiredPosition.Value);
+                _desiredPosition = null;
+            }
```
MichaelDePiazzi: Man, this Media Foundation library is so finicky! It doesn't work properly for me either if I try to specify the position when first starting the session. It kinda sucks that this needs to be done, but  this workaround does seem to work for me (on my Win7 machine at least).

However, I'd recommend adding a comment to the code to explain why this is being done otherwise it won't make sense to other people looking at this later.

MonoGame.Framework/Media/MediaPlayer.WMS.cs
danzel: Thanks, seems obvious looking back!

MonoGame.Framework/Media/MediaPlayer.WMS.cs
danzel: Not sure why I didn't do that. Added now.

MichaelDePiazzi:
@danzel Cool, the WMS changes look good to me now.

danzel:
Have just pushed up a WME implementation and hopefully updated the WP one enough to build, I haven't implemented it there though (I don't plan to).

The WME version probably needs a bit more testing, I've only tested on a WP8 Store app (Someone has my WP8.1 device).

Internally the WP version sometimes uses the XNA MediaPlayer (which can't support this), and sometimes it uses a ```System.Windows.Controls.MediaElement``` (which I think could support this).
Will need someone else to implement if this is important on that platform.

danzel:
I think this is all good to go now.

danzel:
Anything I can do to get this merged? We have implementations for as many platforms as I can support without major work.

KonajuGames:
Sorry.  So many emails from comments on GitHub issues/PRs and the Community forums that things might get left behind a bit if I don't stay on top of them.

Thanks for the work on this.  A combined effort.  Merging.

danzel:
Awesome thanks :)

Issue #4202, opened 2015-10-23
SpriteBatch rendering is not smooth


bousquetm:
Hi all,

I've noticed an issue with SpriteBatch rendering on WindowsGL platform (does not occur on Windows DX nor XNA 4.0). When drawing a sprite that moves at constant speed, the animation "stutters" and is not smooth, even though the program runs at 60fps. It seems that sometimes the draw is delaye or something.

Here's a very simple example to reproduce the bug (the texture red is just a 64x64 red texture):

```C#
using Microsoft.Xna.Framework;
using Microsoft.Xna.Framework.Graphics;
using Microsoft.Xna.Framework.Input;

namespace MonogameTest
{
	public class Game1 : Game
	{
		public Game1()
		{
			_graphics = new GraphicsDeviceManager(this);
			Content.RootDirectory = "Content";
			IsFixedTimeStep = true;
		}

		////////////////////////////////////////////////////////////////////////// Override Methods

		protected override void LoadContent()
		{
			_spriteBatch = new SpriteBatch(GraphicsDevice);
			_texture = Content.Load<Texture2D>("red");
			_pos = Point.Zero;
		}

		protected override void Update(GameTime gameTime)
		{
			if (Keyboard.GetState().IsKeyDown(Keys.Space))
			{
				_pos.X = 0;
			}
			_pos.X += 6;

			base.Update(gameTime);
		}

		protected override void Draw(GameTime gameTime)
		{
			GraphicsDevice.Clear(Color.CornflowerBlue);

			_spriteBatch.Begin();
			_spriteBatch.Draw(_texture, _pos.ToVector2(), Color.White);
			_spriteBatch.End();

			base.Draw(gameTime);
		}

		////////////////////////////////////////////////////////////////////////// Private Members

		private GraphicsDeviceManager _graphics;
		private SpriteBatch _spriteBatch;
		private Texture2D _texture;
		private Point _pos;

	}
}
```

I've reproduced it Monogame 3.4 and also with the last unstable version (currently 3.5.0.703). I also tried on 2 different computers, same problem with both.

Thanks in advance!

Martin

KonajuGames:
Log out the value of gameTime and see if it has any fluctuations.

bousquetm:
I logged the gameTime value in Update and Draw: no fluctuations, I'm always at 16.667ms. Just to be sure, I also removed IsFixedTimeStep = true in the constructor.

MichaelDePiazzi:
[...]

PR #4207, opened 2015-10-28
Improved accuracy of fixed time step

MichaelDePiazzi:
This is in response to issue #4202. I can't say that it will resolve the issue, but it should definitely at least improve things a little.

In summary, this PR tries to better consider the inaccuracy of "Sleep" and use it more conservatively to try and avoid overshooting the desired update time. The side effect of this is that there'll be some CPU spinning, but the CPU usage increase should be minimal.

For Windows platforms (except for WinRT), it now considers the timer resolution when sleeping. It will try to sleep as long as possible without overshooting, and then spin the CPU for the remainder. My tests have shown this to be very accurate, and in theory it should never actually overshoot due to oversleeping.

I couldn't use this trick on other platforms though. So for other platforms, it instead only tries to sleep when the next update time is still at least 1ms away, otherwise it will spin the CPU for the remainder. Also, to further reduce the chance of overshooting, it will only perform a "Sleep(1)" each time it does sleep.

Lastly, @nkast also pointed out that `Task.Delay` (used by WinRT in place of `Thread.Sleep`) was generating garbage. So I've changed it to use `Monitor.Wait` instead, which should be just as accurate without the garbage generation. I wasn't able to properly test this though with a Windows Store app.

While I was able to test all of the 3 methods on my Windows 7 machine and prove they at least work on that platform, I wasn't able to test any of this on any other platforms. WinRT in particular probably needs a good test as it's now using `Monitor.Wait` to sleep. The other platforms should be ok though as the change was much less significant, but it would be great if someone could test them also.


tomspilman:
Do you think maybe `Monitor.Wait` might be a better way to sleep across all platforms?   Less `#ifs` and all that.

MichaelDePiazzi:
> Do you think maybe Monitor.Wait might be a better way to sleep across all platforms?

I did test it, but I found that it's far less accurate than `Thread.Sleep` (at least on Windows 7). For WinRT, there doesn't seem to be an alternative to `Thread.Sleep` which is as accurate though. I also tested `Task.Delay` and `ManualResetEvent.WaitOne`, and their accuracy seemed equivalent to `Monitor.Wait`.

tomspilman:
> I found that it's far less accurate than  Thread.Sleep 

Interesting...  I would have thought that synchronization objects would be as accurate if not more.

MichaelDePiazzi:
> Interesting... I would have thought that synchronization objects would be as accurate if not more.

Here's some quick benchmarks on the range of the fluctuations in the update interval I get from different methods (my timer resolution was at 1ms while running all of this):

* `SleepForNoMoreThan` = 16.6-16.7ms
* `Thread.Sleep` = 16.5-16.9ms
* `Monitor.Wait` = 1.7-31.6ms
* `Task.Delay` = 1.6-31.5ms
* `ManualResetEvent.WaitOne` = 1.7-31.6ms

MonoGame.Framework/Utilities/TimerHelper.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Runtime.InteropServices;
```
KonajuGames: Add the standard MonoGame file header block.

MonoGame.Framework/Utilities/TimerHelper.cs
```diff
@@ -0,0 +1,37 @@
...
+            var timerResolution = GetCurrentResolution();
+            int adjustedTime = (int)(milliseconds * 10000);
+            if (timerResolution > adjustedTime)
+                return;
+            var tickCount = adjustedTime / timerResolution;
+            int adjustedSleepTime = (((tickCount - 1) * timerResolution) / 10000) + 1;
+            Thread.Sleep(adjustedSleepTime);
+        }
+    }
+}
```
KonajuGames: Add a newline at the end of the file.

KonajuGames:
If WinRT doesn't have access to Thread.Sleep or the TimerHelper class, it should spin.  Those timing results are atrocious.

MichaelDePiazzi:
@KonajuGames Whoops, my bad. I'll fix up the TimerHelper class as you've suggested.

> If WinRT doesn't have access to Thread.Sleep or the TimerHelper class, it should spin. Those timing results are atrocious.

Yeah, it's not brilliant. Of course it all averages out, but the fluctuations would probably be noticeable.

Mind you, this is just me testing this method on Win7 though. I'm not sure if it would be the same running it as a Windows Store app on Win8+. Can someone possibly try this under WinRT and see how it performs?

I also made a suggestion in issue #4202 related to spinning the CPU:

> However, XNA does seem to be spinning the CPU. Doing a comparison between MonoGame and XNA using fixed time step and no v-sync, XNA utilises 100% of a core (and for some reason runs at 61 fps?!? Maybe an XNA bug?) while MonoGame uses very little (obviously due to Sleep being used). So perhaps another solution for people who need precise timing and don't care about the CPU usage (i.e. XNA behaviour) would be to add a MonoGame specific property to enable/disable spin waiting? (e.g. Game.UseSpinWaiting with a default of false)

MichaelDePiazzi:
Also, to further illustrate the issues with timer resolution, here's some benchmarks using a 15.6ms timer resolution (I made a console test app to get around my MonoGame test app setting the resolution to 1ms):

* `SleepForNoMoreThan` = 16.6-16.7ms
* `Thread.Sleep` = 14.0-31.2ms
* `Monitor.Wait` = 15.0-31.4ms
* `Task.Delay` = 14.9-31.3ms
* `ManualResetEvent.WaitOne` = 14.8-31.4ms

You can see that `SleepForNoMoreThan` maintains accuracy with the lower resolution, whereas the other methods don't.

KonajuGames:
Those results are the same as the game randomly jumping between 30 and 60fps.  Not pretty at all.

KonajuGames:
QueryPerformanceCounter is available for Windows 8.1 Store Apps and Windows Phone 8.1.  It is used by the VS2013 DirectX templates as described [here](http://blogs.msdn.com/b/chuckw/archive/2014/12/03/understanding-game-time-revisited.aspx).

KonajuGames:
Found a great detailed investigation of all Windows time measurement methods in [The Windows Timestamp Project](http://www.windowstimestamp.com/description).

MichaelDePiazzi:
> Those results are the same as the game randomly jumping between 30 and 60fps. Not pretty at all.

Yeah, I just did a quick test with a sprite moving across the screen using the `Monitor.Wait` method, and I can see noticeable stutters.

Using CPU spinning to get accurate timing will max out a CPU core though. And this may not be desirable in all situations. e.g. as @tomspilman said:

> The sleep is there to allow desktop windowed games to free up unused CPU cycles. If not a match 3 puzzle game would consume 100% of a core mostly spin locked waiting for the timer to do the next tick.

What do you think of the additional `Game` class property I proposed to enable/disable this behaviour?

> QueryPerformanceCounter is available for Windows 8.1 Store Apps and Windows Phone 8.1. It is used by the VS2013 DirectX templates as described here.

The `Stopwatch` class also uses the high performance counter if it's available (https://msdn.microsoft.com/en-us/library/system.diagnostics.stopwatch%28v=vs.110%29.aspx), so it's just as accurate. Basically to get high accuracy timing by spinning the CPI in MonoGame right now, we'd only need to get rid of the call to "sleep".

MichaelDePiazzi:
> Found a great detailed investigation of all Windows time measurement methods in The Windows Timestamp Project.

Great doc, btw! Haven't read the whole thing, but there is some excellent information in there.

MichaelDePiazzi:
By the way, the previous benchmarks I gave had v-sync **disabled** (so that it wouldn't influence the timing). However, with it **enabled** things get a **lot better** for `Monitor.Wait` and there is **no visible stuttering**.

Here's some new benchmarks, but this time I'm showing the **standard deviation** as it better shows the typical fluctuations than the range does (I should've done this in the first place). I'm also including CPU spinning for better comparison. These benchmarks were done with the **WindowsGL** build and a **1ms timer resolution**.

    | CPU spin | SleepForNoMoreThan | Thread.Sleep | Monitor.Wait
--- | -------- | ------------------ | ------------ | ------------
V-sync **disabled** | ~0.0017ms | ~0.0046ms | ~0.0064ms | ~6.2ms
V-sync **enabled** | ~0.12ms | ~0.13ms | ~0.15ms | ~0.14ms

So you can see that there's practically no difference between the methods with v-sync enabled (at least with a 1ms timer resolution). Although, I'm sure that with a lower timer resolution (e.g. 15.6ms), `Thread.Sleep` and `Monitor.Wait` would both suffer badly. Unfortunately I can't prove this as my timer resolution is always automatically set to 1ms whenever I run a MonoGame project.

MichaelDePiazzi:
Actually, I take that back... I just realised running the DirectX build with my test project keeps the timer resolution at 15.6ms (if I close Visual Studio). Not sure why I didn't notice that earlier...

Anyway, here's some more benchmarks showing the standard deviation with the **DirectX** build and a **15.6ms timer resolution**:

    | CPU spin | SleepForNoMoreThan | Thread.Sleep | Monitor.Wait
--- | -------- | ------------------ | ------------ | ------------
V-sync **disabled** | ~0.0010ms | ~0.0034ms | ~6.2ms | ~6.2ms
V-sync **enabled** | ~0.0008ms | ~0.0031ms | ~6.1ms | ~6.2ms

So this proves that `Thread.Sleep` and `Monitor.Wait` do fall apart at a lower timer resolution, regardless of the v-sync setting. It also seems to show that DirectX has a more accurate v-sync than OpenGL.

MichaelDePiazzi:
@tomspilman @KonajuGames
So if this kind of potential fluctuation is unacceptable, what do you guys think needs to be done?

I don't think we need to worry about Windows as `SleepForNoMoreThan` definitely seems accurate enough regardless of the timer resolution (I've been using it my own Windows DirectX game for quite some time now without any reported issues). For other platforms though, the accuracy can't be guaranteed and I think we should consider spinning the CPU.

Testing XNA, it definitely maxes out a CPU core when using fixed timed step with v-sync disabled, so I assume it must be spinning the CPU. When the window is inactive though, the CPU usage drops right down. This must be `Game.InactiveSleepTime` in play (https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.game.inactivesleeptime.aspx). However, this property doesn't currently do anything in MonoGame.

When v-sync is enabled, spinning the CPU doesn't really seem to be a problem in XNA or MonoGame as waiting for the vertical retrace seems to keep the usage down.

So to try and keep everyone happy, I propose adding a new `Game` class property named `UseAccurateTiming`. For Windows, it won't do anything as it'll always just use `SleepForNoMoreThan` which is already accurate. For other platforms though, if it's set to `true` then CPU spinning will be used to ensure accurate timing. And if it's set to `false` then it will try to sleep when it can to reduce CPU usage (at the potential cost of accuracy).

I also think the default setting for this property should be `true`. I'd say this is in line with most users expectations that updates should be smooth. Plus this seems to be the XNA behaviour. And given that v-sync is enabled by default, it also means that the CPU shouldn't be getting maxed out by default anyway.

We should also Implement `Game.InactiveSleepTime` so that the CPU won't potentially continue to be maxed out when the window's not active. My assumption here is that it just sleeps this amount of time on each tick.

So what do you think? And should this additional work be part of this PR, or should it be separate?

KonajuGames:
I agree with everything you've said there.  I didn't even know the `Game.InactiveSleepTime` property existed, but it makes sense.  I would be happy for the further work to be in this PR as it isn't much more and is dependent on this.

tomspilman:
The feature was always for desktop systems where you could be windowed with vsync disabled and sucking up all the CPU for no reason.   So we can start with making the feature only apply to Windows and maybe Mac and Linux.

Making it opt in wouldn't be a problem.

MichaelDePiazzi:
@tomspilman

> So we can start with making the feature only apply to Windows and maybe Mac and Linux.

Sorry, I'm not sure so I just want to be confirm... which feature are you talking about?

My `SleepForNoMoreThan` solution (which I've already pushed) will only work on Windows. It allows sleeping while providing accurate timing. So we won't need to spin the CPU on Windows.

For all other platforms, I was proposing that we enable CPU spinning by default. But the user can opt-out by disabling the `Game.UseAccurateTiming` property and enable sleeping at the cost of accuracy (i.e. the current behaviour for all platforms).

Does this sound ok to you, or would you like it to work differently somehow?

MichaelDePiazzi:
@tomspilman Sorry, I think I was a bit tired and run down yesterday... I re-read what you said after a good nights sleep and a nice strong coffee, and it's a bit clearer to me now.

So are you proposing that we get rid of sleeping all together for non-desktop systems?

tomspilman:
@MichaelDePiazzi - Sorry crazy busy here.

> So are you proposing that we get rid of sleeping all together for non-desktop systems?

Yeah... I think so.  Things like Windows 10 is tricky because it can be either a phone, a desktop app, or a fullscreen app.  So not sure about that.

Still it is an optional feature... so maybe old school Windows desktop only is a better way to limit this.  If we find a need for it on other desktops or non-desktop systems, we can look to add it back in later.

@KonajuGames - Thoughts?

MichaelDePiazzi:
@tomspilman 

> Sorry crazy busy here.

All good, I can definitely empathise... :sweat:

> Yeah... I think so.

Ok, this sounds reasonable. I'm only really familiar with Windows desktop, so I can't comment on other platforms.

> Still it is an optional feature... so maybe old school Windows desktop only is a better way to limit this.

For Windows, I don't feel it needs to be optional as my `SleepForNoMoreThan` code is pretty accurate. So we could just scrap my proposed `UseAccurateTiming` option all together, and instead always use `SleepForNoMoreThan` for Windows desktop and CPU spinning for all other platforms. And if I then also implement the `Game.InactiveSleepTime` feature from XNA, then this will perform sleeping on each tick while the game is inactive by default.

KonajuGames:
Android and iOS use the update and draw callbacks from their OpenTK implementation.  I haven't done any testing to see how consistent the timing is on those.

For now it might be best to keep this better timer for Windows desktop only and all other platforms spin where possible to maintain best frame timing.  And implement `Game.InactiveSleepTime` to give the CPU a breather when the window is minimized.

MichaelDePiazzi:
> For now it might be best to keep this better timer for Windows desktop only and all other platforms spin where possible to maintain best frame timing. And implement Game.InactiveSleepTime to give the CPU a breather when the window is minimized.

Ok cool, I think we're all on the same page now. I'll try and get these changes pushed within the next day.

MichaelDePiazzi:
@tomspilman @KonajuGames Discussed changes pushed and build successful! Please let me know if anything else is required on this.

MichaelDePiazzi:
Whoops.. just realised I'd forgotten to update the comments which talked about sleeping. Fixed.

MichaelDePiazzi:
@bousquetm @StunGunStudios
Since you guys both had issues with this, could you possibly give this PR a quick test?
You can grab the binaries from here: http://teamcity.monogame.net/viewLog.html?buildId=11017&buildTypeId=MonoGame_DevelopWin&tab=artifacts

StunGunStudios:
@MichaelDePiazzi Sorry for the delay. I tested Windows and WindowsGL and they both seem to work fine, with no stuttering or anything. Note that the tests were done using a release build, and that they were not extensive, time-wise. Theoretically these changes shouldn't be affected by the amount of time the game has been running, though.

MichaelDePiazzi:
Cool, thanks @StunGunStudios

bousquetm:
I also tested Window and Windows GL. In both cases, it works much better than the old version! However I still have some stuttering, but it's far less noticeable than before (tested on 2 different PC).

MichaelDePiazzi:
Thanks @bousquetm. Is the stuttering you're still getting only in WindowsGL, or is it also in the DirectX version? And is this stuttering occurring within your simple test app with the sprite moving across the screen? Did you try disable v-sync to see if that got rid of it? (i.e. set `GraphicsDeviceManager.SynchronizeWithVerticalRetrace = false` in your `Game` class constructor)

MichaelDePiazzi:
@KonajuGames @tomspilman
Hey guys, I believe this should be good to merge now.

bousquetm did report that he still got some minor stuttering, but it was otherwise a big improvement from before. I have a feeling that the stuttering he saw may be related to the OpenGL v-sync implementation. This is because the benchmarks I did previously showed that OpenGL seemed to be a little less accurate when enabling v-sync (even when using CPU spinning). So I think this is maybe another issue.

If there's anything further that you feel is required for this PR, then please let me know.

MichaelDePiazzi:
I just found a minor issue in my SleepForNoMoreThan where sometimes it could sleep a little too long (by less than 1ms). I've now fixed this and simplified the logic in the process.

Here's some benchmarks (with v-sync disabled) comparing the original standard deviations with that of the fixed method:

    | CPU spin | SleepForNoMoreThan | Fixed SleepForNoMoreThan
--- | -------- | ------------------ | ------------------------
WinGL, 1ms resolution | ~0.0017ms | ~0.0046ms | ~0.0013ms
WinDX, 15.6ms resolution | ~0.0010ms | ~0.0034ms | ~0.0014ms

You can see that SleepForNoMoreThan is now practically identical to CPU spinning.

@KonajuGames @tomspilman P.S. I'm guessing you're waiting until after the 3.5 release before merging this?

MichaelDePiazzi:
Hmmm... Mac build failed on the build server for some reason... I'm guessing some things have since changed on the build server, so trying a rebase on develop...

tomspilman:
> I'm guessing some things have since changed on the build server, so trying a rebase on develop

Yep... that was exactly the issue.

MichaelDePiazzi:
@tomspilman @KonajuGames This should be good to merge. Are you good with making it part of 3.5, or did you want to wait until after it's released?

MichaelDePiazzi:
@tomspilman @KonajuGames @dellis1972 Hey guys, this has been open for over 3 months now. From my point of view, it should be all good to merge.

But please let me know if you feel anything further is required, or if you're intentionally postponing this for some reason (e.g. waiting until after 3.5 is released).

If there's anything I can do to help get this merged, then I'm more than happy to. Just let me know.

tomspilman:
@MichaelDePiazzi - I will be reviewing this over the weekend and we will get this worked out and merged for the 3.6 release.  Sorry for the long delay... we're getting better at this I swear. :)

dellis1972:
@tomspilman I believe WindowsGL is being dropped in favour of "DesktopGL". Effectively the WindowsGL and Linux projects are the same at the time (afaik).. Adding this to WindowsGL will mean that assembly will no longer work on Linux/Mac. Just thought I'd mention it.

MichaelDePiazzi:
@tomspilman No problem. Let me know if you need anything from me (e.g. in regards to the potential issue @dellis1972 has raised).

MichaelDePiazzi:
@tomspilman Just thought I'd give this another poke to see if we can get some traction on it...

@dellis1972 mentioned that adding this change to the WindowsGL build will break Linux/Mac when moving to "DesktopGL". So if I enclosed the Windows specific "TimerHelper" class with the "WINDOWS" conditional compilation constant (so that the class can still be included for other OS's without causing a compilation error), would this resolve any potential concerns?

MichaelDePiazzi:
@tomspilman @KonajuGames @dellis1972 Hey guys, any chance this can get looked at? It's been open for 10 months now. What needs to be done to move this forward?

nkast:
If I may add a couple of thoughts on this one,

Right now we just monitor the system if other processes change the resolution, which means that if the 
resolution is @16ms we mainly spin the CPU.
Also the resolution can change from 0.5ms to 16ms at any point while the game sleeps. (quite improbable of course, but in theory it could get the loop out of sync)

So, can we set the TimerResolution to the lowest possible timespan with NtSetTimerResolution?
We could either set it during initialization or only for the duration of sleep. 
The documentation from Windows Multimedia has this to say on timeBeginPeriod()/timeEndPeriod() which I suppose also uses NtSetTimerResolution.

```
Call this function immediately before using timer services, and call the timeEndPeriod function immediately after you are finished using the timer services.
...
This function affects a global Windows setting. Windows uses the lowest value (that is, highest resolution) requested by any process. Setting a higher resolution can improve the accuracy of time-out intervals in wait functions. However, it can also reduce overall system performance, because the thread scheduler switches tasks more often. High resolutions can also prevent the CPU power management system from entering power-saving modes.
```
https://msdn.microsoft.com/en-us/library/dd757624(v=vs.85).aspx

nkast:
Another issue is `Task` and whether it generates garbage. Nowadays those classes are IInspectable(CountRef) on WinRT and I'am  not sure whether they put any pressure on the GC or how they behave across different frameworks (Desktop/WinRT-8/WinRT-10/Net.Native).

But since we no longer sleep in the WINRT game loop and we don't really care about hiccups when the app is inactive we could just use Task.Sleep() if that improves readability.

MichaelDePiazzi:
@nkast I agree. I think this is a good idea. I actually recently started doing exactly this in my game because it kept the CPU usage lower (because it can sleep more and spin the CPU less). I'm still using my `SleepForNoMoreThan` method though to keep the timing as accurate as possible. Although it obviously doesn't make nearly as much of a difference when the resolution is 0.5ms, but it still helps.

I'm not sure if there is much benefit to changing the resolution just before sleeping, but I've never tried it so I can't say for sure. I'm personally just setting it once via `NtSetTimerResolution` at initialisation. No need to clean up either really as this is handled by Windows when the process ends.

I can update this PR to do this if there's still interest in it. But I'm not really sure as this PR has been dead for almost a year now. Last comments were regarding some uncertainty over how this would fit with the "DesktopGL" build, but it never progressed from there. I'll leave it in @tomspilman and @KonajuGames court to advise on how to proceed. And if this PR is no longer relevant, please feel free to close it.

nkast:
> I'm still using my SleepForNoMoreThan method though to keep the timing as accurate as possible. 

Yes, of course! We still spin for LowestSleepThreshold and we change the resolution only when we have to sleep.

> I'm not sure if there is much benefit to changing the resolution just before sleeping, [...]. I'm personally just setting it once via NtSetTimerResolution at initialisation.

(quote from above) `might prevent the power management  from entering power-saving mode` and `can also reduce overall system performance, because the thread scheduler switches tasks more often`.
A possible problem with the pseudo code I had above is when the user or 3rd party libraries also changes the resolution. Should we query the last value and restore it or there is any higher level API we can use similar to timeBeginPeriod() / timeEndPeriod()?

nkast:
@MichaelDePiazzi 
 >Last comments were regarding some uncertainty over how this would fit with the "DesktopGL" build

@KonajuGames 
>Android and iOS use the update and draw callbacks from their OpenTK implementation. I haven't done any testing to see how consistent the timing is on those.
>For now it might be best to keep this better timer for Windows desktop only and all other platforms spin where possible to maintain best frame timing. 

I think @KonajuGames means all Windows/Win32 platforms here, WINDOWS(Desktop/DX), DesktopGL (windows),  WINDOWS_STOREAPP (W8.1/WP8.1/Win10/etc).

For the other platforms that don't support NtGetResolution or similar APIs (ANDROID/IOS,etc) we can still make an estimation on a minResolution/LowestSleepThreshold, I guess near 4ms. That way it doesn't eat up all the battery on light games.

MichaelDePiazzi:
> (quote from above) might prevent the power management from entering power-saving mode and can also reduce overall system performance, because the thread scheduler switches tasks more often.

My logic for doing this was that using a higher timer resolution kept the CPU usage lower (due to less spinning). So from a power-saving point of view, this was better than using a lower timer resolution because that spun the CPU more and resulted in more CPU usage.
 
> A possible problem with changing the resolution before sleeping and then setting it back to MinimumResolution is when the user or 3rd party libraries also changes the resolution. Should we query the last value and restore it or there is any higher level API we can use similar to timeBeginPeriod() / timeEndPeriod()?

The easiest way is to just set the timer resolution to the max. That way it doesn't matter what the user specifies (it shouldn't anyway as any process can change the resolution to the max). I'm not really sure if there's much benefit to switching it each time. Switching the resolution back and forth 60 times a seconds may be a bit too much to be worthwhile. But I guess to find out for sure, the best way is to just try it.

If we were to do this, then the best way would be to use `timeBeginPeriod` / `timeEndPeriod` because it handles what you're talking about (although it only allows a max resolution of 1ms). I'm not aware of any other APIs that deal with timer resolution.

But perhaps a better solution is to only use the max timer resolution when the game window is active? When the game window is inactive, then the min timer resolution could be set instead. Also we could use a different sleep method which avoids CPU spinning and sacrifices accuracy a bit so that it can sleep more. I'm kinda already doing something like this in my dedicated server. When there's no connections to it, it uses a less accurate method which sleeps more because timing doesn't matter so much.

> Last comments were regarding some uncertainty over how this would fit with the "DesktopGL" build

This was actually in regards to the last comment from @dellis1972:

> I believe WindowsGL is being dropped in favour of "DesktopGL". Effectively the WindowsGL and Linux projects are the same at the time (afaik).. Adding this to WindowsGL will mean that assembly will no longer work on Linux/Mac. Just thought I'd mention it.

PR #4208, opened 2015-10-28
Fix NVorbis "instance not running" on game end

cra0zy:
So, long story short, in #3837 I added a dispose method which was needed because exit errors were stopping OpenALSoundController from disposing and making the game run forever, but now that #4153 got merged that's no longer the case.


KonajuGames:
Thanks @cra0zy . Merging.

PR #4209, opened 2015-10-28
Better Window Title and Icon Handling for DesktopGL

cra0zy:
Notes:
 - in some older version of mkbundle/mono, Assembly.GetEntryAssembly() would throw an error when called from application from inside mkbundle, it has been fixed since then
 - on Mono platforms "Icon.ExtractAssociatedIcon" does nothing(always loads Mono icon) so I included the MonoGame icon so it would load that as default instead of the Mono icon
   - will adding few extra KB to dll size be an issue?


KonajuGames:
Thanks @cra0zy . Merging.

Issue #3995, opened 2015-07-12
Infinite loop caused by Game.Exit doesn't allow application to end


teryror:
This bug is in the WinDX platform layer, I haven't looked at the others yet.

If you want to repro it the way I found it, it's also _really_ obscure, but the fact that it can happen at all is bad enough, right?

I was mucking around with what was left of the project template, and the call to Game.Exit ended up on the main code path. Imagine

``` C#
protected override void Update(GameTime gameTime)
{
    Exit();
    base.Update(gameTime);
}
```

That alone doesn't cause the bug though. For various reasons, I also had the [Visual Studio hosting process disabled](https://msdn.microsoft.com/en-us/library/ms185330.aspx).
These two combined will have the game window pop up, close immediately and then indefinetely continue to run. Hitting the pause button in the debugger, I got this stack trace:

```
mscorlib.dll!System.Threading.Thread.Sleep(int millisecondsTimeout)
MonoGame.Framework.dll!MonoGame.Framework.WinFormsGameWindow.RunLoop()
MonoGame.Framework.dll!Microsoft.Xna.Framework.Game.Run(Microsoft.Xna.Framework.GameRunBehavior runBehavior)
Game1.exe!Game1.Program.Main()
```

I assumed that Thread.Sleep was unlikely to be the cause of the infinite loop, so I looked at WinFormsGameWindow.cs. Looking at RunLoop(), there's only one call to Thread.Sleep, and it's inside [this loop](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Windows/WinFormsGameWindow.cs#L435-451).
Compared to the loop above it, it simply seems to lack `Application.DoEvents()` to actually pull out messages from the queue, but I have not tested this yet. I'm not sure why this seemingly never caused any issues before, but I assume calling Exit on the first update doesn't give the first loop enough time to clear the queue, and the queue would usually be empty anyway, unless the the hosting process is disabled?

But that's just conjecture. Unless one of the devs thinks it's obvious enough and fixes it themselves, I'll try and fix this myself and make a pull request later today or maybe tomorrow.

totallyeviljake:
This deadlock is likely being caused by this snipped of NUnit specific code that was added to the window RunLoop on the DX platform:

            var msg = new NativeMessage();
            do
            {
                if (msg.msg == WM_QUIT)
                    break;

                Thread.Sleep(100);
            } 
[...]

PR #4212, opened 2015-10-29
Added event for form move.

cuddlyogre:
The form move event is now implemented.

### DIFF
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -84,6 +84,7 @@ namespace Microsoft.Xna.Framework {
 
 		public event EventHandler<EventArgs> ClientSizeChanged;
 		public event EventHandler<EventArgs> OrientationChanged;
+		public event EventHandler<EventArgs> Move;
 		public event EventHandler<EventArgs> ScreenDeviceNameChanged;
 
 #if WINDOWS || WINDOWS_UAP || DESKTOPGL|| ANGLE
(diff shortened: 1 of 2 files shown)


tomspilman:
@KonajuGames @dellis1972 - Do you think we need a `GameWindow.Moved` event?

@cuddlyogre - What was the need you had which required this new feature?

cuddlyogre:
The application I am building requires that the window redraw continuously on move and resize. To bypass #3995 / #4211, I set up a second tick that runs with those events. It works perfectly on resize and after I added a move event, that worked as well.

I imagine there are more game-related uses for this as well, like an etch-a-sketch type game.

cuddlyogre:
Out of curiosity, has this been discussed any further?

tomspilman:
Bump!

@KonajuGames @dellis1972 @cra0zy - Do you think we need a `GameWindow.Moved` event?  Seems like something that is only useful on desktops.

cra0zy:
#4662 

I think this should be closed as well.

tomspilman:
Thanks for reminding me of that out @cra0zy .  We've already ruled on not wanting to expose this type of event to MonoGame.

@cuddlyogre - Take a look at the comments in https://github.com/mono/MonoGame/pull/4662 for our reasoning.

cuddlyogre:
Are there any thoughts behind solving issue #3995 so as to render my work around unnecessary? It seems counterproductive to keep a known bug in place only so tests will pass.

PR #4213, opened 2015-10-30
Dispose XAudio source voices before disposing the XAudio master voice.

jbeshir:
When tracking down some intermittent crashes on shutdown which seemed to be related to XAudio, I found that with a debug build of MonoGame and the Windows SDK installed I would see an "XAUDIO2: ERROR: The voice still has 1 senders" error output by my program on shutdown.

Looking into it further I found:

1) Reports of crashes if the mastering voice isn't destroyed properly before the device is destroyed.
2) https://msdn.microsoft.com/en-us/library/windows/desktop/hh405048(v=vs.85).aspx documenting that source voices must be destroyed before the mastering voice can be.

This PR causes sound effect instances to be disposed (which disposes their source voice) on PlatformShutdown of SoundEffect, before the MasteringVoice is disposed, and stops the XAudio error from being output on shutdown. I believe this also fixes an intermittent crash problem on shutdown but it's harder to tell.

I think so long as you didn't shut down while playing audio, you'd only have seen this problem if you were using pooled SoundEffectInstances, but otherwise it should have shown up consistently.

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -226,6 +226,8 @@ namespace Microsoft.Xna.Framework.Audio
 
         internal static void PlatformShutdown()
         {
+            SoundEffectInstancePool.Shutdown();
+
             if (MasterVoice != null)
             {
                 MasterVoice.DestroyVoice();
--- a/MonoGame.Framework/Audio/SoundEffectInstancePool.cs
+++ b/MonoGame.Framework/Audio/SoundEffectInstancePool.cs
@@ -181,5 +181,24 @@ namespace Microsoft.Xna.Framework.Audio
                 inst.Volume = inst.Volume;
             }
         }
+
+        internal static void Shutdown()
+        {
+            // We need to dispose all SoundEffectInstances before shutdown,
+            // so as to destroy all SourceVoice instances,
+            // before we can destroy our XAudio MasterVoice instance.
+            // Otherwise XAudio shutdown fails, causing intermittent crashes.
+            foreach (var inst in _playingInstances)
+            {
+                inst.Dispose();
+            }
+            _playingInstances.Clear();
+
+            foreach (var inst in _pooledInstances)
+            {
+                inst.Dispose();
+            }
+            _pooledInstances.Clear();
+        }
     }
 }


tomspilman:
This seems reasonable at first glance to me.  @KonajuGames ?

KonajuGames:
The other PR that looked for matching audio formats did the same thing too
if I remember correctly.

tomspilman:
@KonajuGames - Was that other PR merged?  Is it still open?

StunGunStudios:
@tomspilman I think @KonajuGames is talking about #4162 which hasn't been merged yet. I could be wrong, though.

KonajuGames:
That is likely it. I'm at PAX Aus at the moment.

MichaelDePiazzi:
It may have been @mrhelmut's other fix you were thinking of - PR #4095: "Fix for XAudio crashing if any instance is playing while parent is disposed" (which has already been merged)

mrhelmut:
#4095 (merged) does exactly that, but in ```SoundEffect.Dispose```, which is safer and cover the case of unloading assets at run time.

I think that this PR would have been incomplete, instances need to be stopped before being disposed (otherwise xaudio crashes).

jbeshir:
That PR stops them playing after their SoundEffect is disposed in the next Update call, but:

1) It isn't sufficient to stop them; they retain their source voice while stopped. They need to be disposed to destroy the source voice before the mastering voice is disposed, and nothing ever disposes pooled instances at present.

2) Update does not essentially get called between calls to SoundEffect.Dispose() for all instances and SoundEffect.PlatformShutdown() when disposing content on termination, so you can't rely on that Update method having run.

If nothing else is stopping sounds playing before PlatformShutdown is run, it might need a loop stopping all currently playing sound effects before disposing the source voice, mastering voices, or device as it currently is, unless disposing the device wraps things up anyway.

jbeshir:
Okay, no, I misread, it doesn't happen next Update call, that's for dropping SoundEffectInstances disposed directly, it happens immediately. The other thing, that it doesn't deal with disposing the pooled instances, it readds them to the pool, and all of those need disposing before you can dispose the mastering voice, is still an issue.

mrhelmut:
@jbeshir I have reviewed your finding and code, and you are right, there is a problem with ```SoundEffectInstance.Dispose()```. It never gets called and voices are not freed, which may cause  an xaudio crash upon game closing (MasterVoice.Shutdown).

```_playingInstances``` will already be empty through ```SoundEffectInstancePool.StopPooledInstances()``` which is called earlier by ```_content.Dispose(true)```, hence you are safe to remove the first ```foreach```.

@tomspilman @KonajuGames this PR is good to merge IMO.

KonajuGames:
Ok, I'm good with merging this.

PR #4216, opened 2015-10-31
Better OpenTK Window handling

### LINKED ISSUE #4215: [Suggestion] Set Window size in the constructor of GraphicsDeviceManager.cs
Hello

It would be cool to get the desired window size at runtime, this way there will be no more that little annoying resize effect

```cs
m_graphics = new GraphicsDeviceManager(this, WIDTH, HEIGHT);
```

### LINKED ISSUE #4344: Game.Window.ClientBounds returns incorrect location
When using XNA, `Game.Window.ClientBounds` returns the desktop location of the upper-left corner of the client area (i.e. the internal area of a window which doesn't include window borders, title bar, etc).

But in the Windows and WindowsGL builds, MonoGame is currently always returning (0, 0). There are likely more builds also affected (very likely Linux, maybe others).

I have a fix for the Windows (DirectX) build and I will submit a PR for this shortly.

However, I am not sure of the fix for WindowsGL / DesktopGL builds. I took a quick look into it, but could not see how to obtain the same position via the OpenTK INativeWindow interface. It only seems to give the desktop location of the entire window (including borders, etc). I could not find a way getting the desktop location of the internal client area.

cra0zy:
This PR does the following things:
 - makes border delay code Linux only, since that bug only affects Linux, on Windows it would create a flicker when window size would change and the border was fixed
 - moves the window set visibility code to the first time that clientbounds gets updated, this way the window will become visible with the correct size instead of becoming visible and then changing size. Fixes #4215
 - Fixes #4344 for DesktopGL side
 - Added Window centering after resize

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -171,12 +171,6 @@ namespace Microsoft.Xna.Framework
             OpenTK.DisplayDevice.Default.RestoreResolution();
         }
 
-        public override void BeforeInitialize()
-        {
-            _view.Window.Visible = true;
-            base.BeforeInitialize();
-        }
-
         public override bool BeforeUpdate(GameTime gameTime)
         {
             IsActive = _view.Window.Focused;
--- a/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
@@ -248,7 +248,15 @@ namespace Microsoft.Xna.Framework
         {
             lock (window)
             {
-                UpdateBorder();
+                if (CurrentPlatform.OS == OS.Linux)
+                {
+                    if (updateborder == 1)
+                        UpdateBorder();
+
+                    if(updateborder > 0)
+                        updateborder--;
+                }
+
                 Window.ProcessEvents();
                 UpdateWindowState();
                 HandleInput();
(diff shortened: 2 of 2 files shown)


MichaelDePiazzi:
I've tested this on Windows and all seems to working correctly.

@cra0zy I noticed in XNA the window is centred when it first appears, whereas WindowsGL doesn't do this. Is it possible to also fix this?

cra0zy:
> @cra0zy I noticed in XNA the window is centred when it first appears, whereas WindowsGL doesn't do this. Is it possible to also fix this?

It is. I'll look into fixing it in here.

cra0zy:
So @AngeloG made a PR a while back with the changes ti make the window center, but never got around to adding a window out of bounds check(#3211).

AngeloG:
Cool! :+1:

MichaelDePiazzi:
I've tested this again on Windows with the latest changes and it's all working as expected for me. Nice one! :thumbsup:

tomspilman:
Has someone tested this on Mac too?

mrhelmut:
Will do that today.

mrhelmut:
Works great on Mac and fixes the mentioned issues. @tomspilman

tomspilman:
@KonajuGames @dellis1972 -  If you guys feel this is good then lets merge it.

cra0zy:
download 1

dellis1972:
@cra0zy thanks for the prompt , have been away the last week just catching up :)

cra0zy:
Funny thing about window centering... turns out it doesn't work on Windows 8 since it already centers the windows by default... need to test it on Windows 10 now... anyone have Windows 10 so I don't have to install it?

tomspilman:
@cra0zy - If you can send me a test EXE I can try it here on my laptop right now.

cra0zy:
https://mega.nz/#!DN5SESyD!kLjxvPLe54gytnC-vFpEt_pDtN4aZ9JAqdGMf7e2fng

cra0zy:
Press A to change size

tomspilman:
@cra0zy - So it started off the edge of the screen on my laptop and was not visible.  One complication here is my laptop has one of those super high DPI displays and my resolution is 3200x1800.  I suspect that this is the reason that happened.

cra0zy:
It has the centering bug, that's what it's causing on Win 8

PR #4219, opened 2015-11-03
Save and restore the pipeline window position and size

prime31:
Minor changes to save out the window position, size and separator position and restore them when reopening the pipeline tool

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -317,6 +317,8 @@ namespace MonoGame.Tools.Pipeline
             History.Default.StartupProject = null;
             History.Default.Save();
 
+            WindowConfiguration.Default.Save ();
+
             Selection.Clear(this);
             UpdateTree();
         }
--- a/Tools/Pipeline/Common/WindowConfiguration.cs
+++ b/Tools/Pipeline/Common/WindowConfiguration.cs
@@ -1,0 +1,85 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.IO;
+using System.IO.IsolatedStorage;
+using System.Text;
+
+namespace MonoGame.Tools.Pipeline
+{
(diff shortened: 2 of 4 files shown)


tomspilman:
@cra0zy - Thoughts on this?

IMO we should be looking to store some sort of unified user settings that could also include the window position/size as well as other things like the project history.

cra0zy:
> @cra0zy - Thoughts on this?

If I remember correctly SizeAllocated was removed in Gtk3, and that is why this build is failing.

As for saving the position and size, Gtk does have a way of auto saving it, but it's for Linux only:
https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-role

And it's missing the window state check(window maximized).

> IMO we should be looking to store some sort of unified user settings that could also include the window position/size as well as other things like the project history.

Agreed. 

Side note, my midterm exams are next week so I won't be able to help much until I am done with them.

prime31:
Totally my fault for the crappy pull request. I am a super newb with MonoGame (just jumped in last week) and I couldn't find the Windows Pipeline tool solution to properly test it out over there.

cra0zy:
> Totally my fault for the bum pull request. I am a super newb with MonoGame (just jumped in last week) and I couldn't find the Windows Pipeline tool solution to properly test it out over there.

Don't worry, we all start as newbs :)

Now for some useful info:
1) To generate projects for Windows/Linux or some other platform do:
mono Protobuild.exe --generate Windows
2) Each system uses different toolkit for the Pipeline Tool:
 - Windows uses Windows Forms
 - Linux uses Gtk 3
 - Mac uses Gtk 2

Mac and Linux share the same code base and are split with different region symbols(ie. GTK2 GTK3 look at #3736 it may help), this is what caused your build to fail.

prime31:
Aha. The Protobuild.exe --generate Windows is what I was missing. That explains why I couldn't find the Windows solution. Thanks for the info!

cra0zy:
> As for saving the position and size, Gtk does have a way of auto saving it, but it's for Linux only:
> https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-role

I just realized that this only works with X11, this is bad since Mir and Wayland exist, so we shouldn't use it to save size/position.

cra0zy:
Just FYI: #4266

prime31:
Closed due to #4266

PR #4222, opened 2015-11-05
Windows DirectX: SupportedDisplayModes considers B8G8R8A8 and B5G6R5 formats

### LINKED ISSUE #4220: "SupportedDisplayModes" returns no results (Windows DirectX)
One of my testers recently tried my game on an older laptop they had. When they tried to set the video resolution though, there were none to select from. Looking into it, `GraphicsAdapter.DefaultAdapter.SupportedDisplayModes` was not returning anything at all.

After some investigation, I found the issue - the laptop only had BGR video modes (specifically `B5G6R5_UNorm`, `B8G8R8A8_UNorm`, and `B8G8R8A8_UNorm_SRgb`). But MonoGame is currently only looking for `R8G8B8A8_UNorm` modes.

Does anyone see any issue with enumerating through all the different `SharpDX.DXGI.Format` values and translating them to the relevant XNA `SurfaceFormat.Color`? (e.g. `Format.B5G6R5_UNorm` -> `SurfaceFormat.Bgr565`, `Format.B8G8R8A8_UNorm` -> `SurfaceFormat.Bgra32`, etc.)

I did notice that WindowGL is currently excluding everything but 32-bit colour modes. I'm not sure if there's a specific reason for this, but I thought it might be relevant and worth mentioning.

PS. I'm more than happy to submit a PR for this. But I just wanted to confirm how this should be resolved first.

MichaelDePiazzi:
Resolves #4220

In some cases `GraphicsAdapter.SupportedDisplayModes` would return no results because only BGR formats were available. This PR changes `SupportedDisplayModes` to also consider `B8G8R8A8_UNorm` and `B5G6R5_UNorm` formats, and more closely match XNA behaviour.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.cs
@@ -255,6 +255,15 @@ namespace Microsoft.Xna.Framework.Graphics
         }
         */
        
+#if DIRECTX && !WINDOWS_PHONE
+        private static readonly Dictionary<SharpDX.DXGI.Format, SurfaceFormat> FormatTranslations = new Dictionary<SharpDX.DXGI.Format, SurfaceFormat>
+            {
+                { SharpDX.DXGI.Format.R8G8B8A8_UNorm, SurfaceFormat.Color },
+                { SharpDX.DXGI.Format.B8G8R8A8_UNorm, SurfaceFormat.Color },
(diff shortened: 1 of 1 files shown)


MichaelDePiazzi:
@KonajuGames @tomspilman This should be good to merge now. I've since implemented this fix into the latest release of my game and it's been tested on several machines. Please let me know if there's anything else required.

KonajuGames:
This looks good to me.  There's nothing else required here I think.

tomspilman:
Thanks.... merging!

PR #4226, opened 2015-11-06
[pipeline] El Capitan did some weird stuff with the $PATH and /usr/bin

dellis1972:
Because of these changes mono can no longer be created in /usr/bin
as its read only. So the mono project have moved that exe to
/usr/local/bin.. which makes sense..

But /usr/local/bin is ONLY included in the path for the terminal/shell
and NOT for Applicaitons. So we end up in a situation where calls
to mono don't work. So the fix is to search for the locations in
the Pipeline tool AND the Build .targets.


dellis1972:
@cra0zy can you check that pathing on Linux please. Are there any additional paths we need to check?

@tomspilman @KonajuGames this one needs to be in the next release otherwise content stuff won't work on El Capitan (which I found out for myself :/ )

KonajuGames:
That change looks good to me.  I don't think there are any other paths to look for.  Why applications should have a different path to the terminal is beyond comprehension.

cra0zy:
> Are there any additional paths we need to check?

No additional paths are needed for Linux.

tomspilman:
Thanks guys... merging!

PR #4228, opened 2015-11-07
Added Mac Setup Documentation

dellis1972:


tomspilman:
Merging... thanks!

cra0zy:
Link for this documentation is broken (http://www.monogame.net/documentation/?page=Setting_Up_MonoGame):
1) Link was not added to the left panel
2) Link in the right panel doesn't work

tomspilman:
Thanks @cra0zy !

We just need to add the link into the config file:

https://github.com/mono/MonoGame/blob/develop/Documentation/config.xml#L24

PR #4239, opened 2015-11-11
Linux .run installer update

cra0zy:
Stuff done:
 - removed the option to select installation directory(it was broken and linux installers usually don't offer this anyway)
 - fixed installing monodevelop addin
 - fixed /usr/share/icons/gnome/scalable/mimetypes not existing on Fedora
 - fixed updating wrong icon cache
 - fixed not having correct permissions to write to /usr/share/applications
 - set Pipeline Tool as default for .mgcb files on both installers


cra0zy:
icon cache/.desktop files work a bit differently on Fedora, will need to look into this.

cra0zy:
this is good to go now

tomspilman:
Merging... thanks @cra0zy !

PR #4240, opened 2015-11-12
Add support for tvOS to MonoGame.  

dellis1972:
This generates a new solution file and project for tvOS which can be found in MonoGame.Framework.tvOS.

Conflicts:
	Build/Module.xml
	Build/Projects/MonoGame.Framework.definition

Thanks to the CocosSharp Team ( @kjpou1 m @rtabbara ) for their work on this.
Nant script stuff will follow once we have the tvOS stuff installed on the build machine.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project Name="MonoGame.Framework" Path="MonoGame.Framework" Type="Library" Platforms="Android,Angle,iOS,Linux,MacOS,Ouya,Windows,Windows8,WindowsGL,WindowsPhone,WindowsPhone81,WindowsUAP,Web">
-
+<Project Name="MonoGame.Framework" Path="MonoGame.Framework" Type="Library" Platforms="Android,Angle,iOS,Linux,MacOS,Ouya,Windows,Windows8,WindowsGL,WindowsPhone,WindowsPhone81,WindowsUAP,Web,tvOS">
   <!--
     Using Protobuild in your own project?  ProjectGuids are only in use here
     for backwards compatibility with existing users of MonoGame, and are not
@@ -44,6 +43,7 @@
     <CustomDefinitions>
       <Platform Name="Android">TRACE;ANDROID;GLES;OPENGL;OPENAL</Platform>
       <Platform Name="iOS">IOS;GLES;OPENGL;OPENAL</Platform>
+      <Platform Name="tvOS">IOS;TVOS;GLES;OPENGL;OPENAL</Platform>
       <Platform Name="Linux">TRACE;LINUX;OPENGL;OPENAL;DESKTOPGL</Platform>
       <Platform Name="MacOS">MONOMAC;OPENGL;OPENAL</Platform>
       <Platform Name="Ouya">TRACE;ANDROID;GLES;OPENGL;OUYA;OPENAL</Platform>
@@ -63,7 +63,7 @@
       <Platforms>Angle</Platforms>
     </Uses>
     <Uses Name="OpenGLGraphics">
-      <Platforms>iOS,WindowsGL,Ouya,Android</Platforms>
+      <Platforms>iOS,WindowsGL,Ouya,Android,tvOS</Platforms>
     </Uses>
     <Uses Name="DirectXGraphics">
       <Platforms>Windows8,WindowsPhone,WindowsPhone81,WindowsUAP</Platforms>
@@ -98,7 +98,7 @@
     <Service Name="OpenGLGraphics">
       <RemoveDefine>DIRECTX;WINDOWS_MEDIA_SESSION</RemoveDefine>
       <AddDefine>OPENGL;OPENAL</AddDefine>
-      <Platforms>iOS,MacOS,Linux,Windows,WindowsGL,Ouya,Android</Platforms>
+      <Platforms>iOS,MacOS,Linux,Windows,WindowsGL,Ouya,Android,tvOS</Platforms>
       <Conflicts>ANGLEGraphics,DirectXGraphics,WebGraphics</Conflicts>
       <Requires>OpenALAudio,_GLCompatible</Requires>
     </Service>
@@ -115,7 +115,7 @@
 
     <!-- Audio APIs -->
     <Service Name="OpenALAudio">
-      <Platforms>Angle,iOS,Linux,Windows,WindowsGL,MacOS,Android,Ouya</Platforms>
+      <Platforms>Angle,iOS,Linux,Windows,WindowsGL,MacOS,Android,Ouya,tvOS</Platforms>
       <Conflicts>XAudioAudio,WebAudio</Conflicts>
     </Service>
     <Service Name="XAudioAudio">
@@ -236,13 +236,15 @@
       <Platforms>Angle,Linux,MacOS,Windows,WindowsGL,WindowsUAP</Platforms>
     </Compile>
     <Compile Include="Threading.cs">
-      <Platforms>Android,Angle,iOS,Linux,MacOS,Ouya,WindowsGL,WindowsPhone</Platforms>
+      <Platforms>Android,Angle,iOS,Linux,MacOS,Ouya,WindowsGL,WindowsPhone,tvOS</Platforms>
     </Compile>
     <Compile Include="TitleContainer.cs" />
     <Compile Include="Vector2.cs" />
     <Compile Include="Vector3.cs" />
     <Compile Include="Vector4.cs" />
-
+    <Compile Include="IPlatformBackButton.cs">
+      <Platforms>tvOS</Platforms>
+    </Compile>
 
     <!-- Microsoft.Xna.Framework.Audio -->
     <Compile Include="Audio\AudioChannels.cs" />
@@ -253,7 +255,7 @@
     </Compile>
     <Compile Include="Audio\InstancePlayLimitException.cs" />
     <Compile Include="Audio\MSADPCMToPCM.cs">
-      <Platforms>Android,Angle,iOS,Linux,MacOS,Ouya,WindowsGL,Web</Platforms>
+      <Platforms>Android,Angle,iOS,Linux,MacOS,Ouya,WindowsGL,Web,tvOS</Platforms>
     </Compile>
     <Compile Include="Audio\NoAudioHardwareException.cs" />
     <Compile Include="Audio\OALSoundBuffer.cs">
@@ -263,7 +265,7 @@
       <Services>OpenALAudio</Services>
     </Compile>
     <Compile Include="Audio\OpenALSupport.cs">
-      <Platforms>iOS,MacOS</Platforms>
+      <Platforms>iOS,MacOS,tvOS</Platforms>
     </Compile>
     <Compile Include="Audio\SoundEffect.cs" />
     <Compile Include="Audio\SoundEffect.OpenAL.cs">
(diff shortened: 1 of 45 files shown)


tomspilman:
Great work @kjpou1 @rtabbara and @dellis1972 !

Build/Projects/MonoGame.Framework.definition
```diff
@@ -44,6 +43,7 @@
     <CustomDefinitions>
       <Platform Name="Android">TRACE;ANDROID;GLES;OPENGL;OPENAL</Platform>
       <Platform Name="iOS">IOS;GLES;OPENGL;OPENAL</Platform>
+      <Platform Name="tvOS">IOS;__TVOS__;GLES;OPENGL;OPENAL</Platform>
```
tomspilman: Why `__TVOS__` and not just `TVOS`?

rtabbara:
I was keen to help out but all thanks should go to @dellis1972, @kjpou1 and @Clancey !

By the way, as discussed in another thread, users are going to need the [Xam.iOS 9.3.x preview](https://developer.xamarin.com/guides/ios/tvos/) to get this up and running.

Build/Projects/MonoGame.Framework.definition
dellis1972: To be fair the Xamarin Platform(s) already define these constants like this 

__TVOS__
__IOS__
__ANDROID__

we should probably migrate over to using those if possible, but didn't want to do that all in this PR. It was my suggestion to stick with __TVOS__ we can change it if needed.

dellis1972:
looks like these is some fix up to do anyway for the Web Platform

Build/Projects/MonoGame.Framework.definition
dellis1972: @tomspilman so ```__TVOS__``` or ```TVOS``` ?

tomspilman: > To be fair the Xamarin Platform(s) already define these constants like this 

Sure... but if we went by what constants each platforms want to define for themselves they would all be inconsistent to each other.  IMO we either stick with our current convention or decide on a new one...  but we shouldn't just leave it to the whim of the platform we're on.

@KonajuGames ?

KonajuGames: Symbols with a preceding double underscore are generally reserved for
system implementations in the C++ world, and not standard in any C#
implementations.

It doesn't hurt to have IOS and ANDROID defined in our projects, even if
Xamarin might declare them as well (is there any documentation to that
effect?), and having TVOS is fine for those parts that need differentiation
at compile time rather than run time.

dellis1972:
@tomspilman @KonajuGames done, ```__TVOS__``` is no more :)

KakCAT:
Hi!

I just got an Apple TV today and ported a minigame without major problems. TVOS specific issues like remote control are not yet done, but MonoGame's part was flawless.

I'm linking https://github.com/mono/MonoGame/issues/4218 because I didn't realize this PR was working until a few hours ago. Some other users may want to know this!

mrhelmut:
Awesome!

I'll try to run some tests too.

@KakCAT your audio problems may be related to #4180 , which still need some tests to be merged.

KakCAT:
Hi!

What is the correct to detect the Apple Remote click?  I'm able to read the touch via TouchPanel state but I can't find the click status anywhere.

KakCAT:
Hi,

I finally found my problem with audio. It's just that Xamarin messing with MediaPlayer https://github.com/mono/MonoGame/issues/3446 .  Sound effects seem to sound well with and without #4180 fix.

On the other side, songs can't be loaded in TVOS because it's using Song.Default.cs instead of Song.IOS.cs. Song.IOS.cs brings a compiler error because MPMediaItem is not found. I don't know how to fix this, but patching the file not to need MPMediaItem makes songs loadable and playable.

hach-que:
You should merge these tvOS changes upstream into Protobuild.

dellis1972:
Updates :) 

Fixed Song :)
Added support for GamePad (ish... in progress) but its a bit difficult to test on the simulator.

dellis1972:
@KakCAT do you have a hardware device? Can you test these changes?

KakCAT:
Hi @dellis1972,

I've tested the code and gamepads are usable with minor fixes :) 

TVOS doesn't seem to assign a number to a controller (initially is set to "Unset"). The developer must assign a number when he starts using it.
https://developer.apple.com/library/tvos/documentation/GameController/Reference/GCController_Ref/index.html#//apple_ref/occ/instp/GCController/playerIndex

I think this should do:

            if (controller == null)
            {
				controller=GCController.Controllers?.FirstOrDefault (c => c.PlayerIndex == GCControllerPlayerIndex.Unset);
                if (controller==null) return new GamePadState();
                controller.PlayerIndex=ind;
				SetUpController (controller);
            }

There's no control about which pad is which like in PS3 though (i.e. if you have 2 gamepads and remote, you have no assurance Apple Remote is #0)


I suppose the correct place to call SetUpController is once the controller's been assigned a number. However this is a little problematic:

The "menu" button is supposed to work something like the back button in WP8: When you press it at the first page of your app/game, it's supposed to give control to the tvOS again. The app can't exit, it's supposed to be on stand by until further notice.
Inside the app, you can use it "as you want" (as a pause game, as a back button, ...)

However, handling the ControllerPausedHandler kills the feature of quitting the app, and you can no longer exit the game in the first page, so it'll probably bring lots of headaches on certification. There may be a function to give control back to the OS, but I've been unable to find it.

dellis1972:
@KakCAT thanks for testing.. I'm still messing with the controllers but its good to know we are making progress.

I'm working on a different Menu handling system now by hooking into PressesBegan/Ended methods on the iOSGameViewController.  ControllerPausedHandler didn't seem to work at all here.

KakCAT:
It seems there's a way to do this, but it's broken :

https://forums.developer.apple.com/thread/23803

The problem is that it will be needed by MonoGame to know if the game is in the top page of the game (where a 'back' button should rely control to the OS) or if it's inside the game (where a 'back' button must be entirely handled by the game.

what I'm doing for a quick port I'm trying to complete is using the 'pressesEnded' callback. I though it was a hack but it seems the only realiable way to do it so far.

KonajuGames:
> The problem is that it will be needed by MonoGame to know if the game is in the top page of the game (where a 'back' button should rely control to the OS) or if it's inside the game (where a 'back' button must be entirely handled by the game.

That is entirely game-dependent.  MonoGame knows nothing about the state of the game.  So unless we get the game dev to set a flag of some sort on the Gamepad class that tells it how to handle the back button on tvOS (set it when they enter the top menu, clear it when they leave the top menu), I don't see many other ways of handling it.

KakCAT:
> That is entirely game-dependent. MonoGame knows nothing about the state of the game. So unless we get the game dev to set a flag of some sort on the Gamepad class that tells it how to handle the back button on tvOS (set it when they enter the top menu, clear it when they leave the top menu), I don't see many other ways of handling it.

Yes, that's what I was implying, a standar mechanism to allow this. WP8.1 had a similar problem (I don't know if/how it was fixed) and with the current OS designs this seems to be a trend.

dellis1972:
@KakCAT I'm overriding the PressEnded and sending the "Menu" UIPress to the game controller to see the Back button.. 

Now the BIG issue is even calling base in PRessEnded means the app doesn't exit when it should. do you see the same thing? (this is on the sim btw)

KakCAT:
In the hardware, calling the base in pressEnded (and pushing menu) makes the app quit, unless ControllerPausedHandler  has been assigned.

dellis1972:
@KakCAT @KonajuGames I was thinking of adding something like 

```Microsoft.Xna.Framework.Input.GamePad.AllowAppBackgrounding = true```

which would let the dev know when MG can pass the "Menu" item press up to the system and background the app.. Not sure how we've don't this in the past for WP 8.1

**Comment by dellis1972** [comment]
Although it should probably be on the Game... or Platform?

**Comment by dellis1972** [comment]
On android we can call Finish on the app (which we do) and it will background it. but we can't do that on iOS...

**Comment by KakCAT** [comment]
I think it makes more sense out of GamePad but honestly I'm quite a bungler myself XD 

KonajuGames and @tomspilman should have a look at this.

**Comment by dellis1972** [comment]
Actually new plan.. I dislike using properties when we have a perfectly good IOC container... 
So We have an interface something like 

```IMenuHandler```

which has a method on it called say "Handled" ..
the developer implements this and registers it with the Game.Services.

For tvOS we look for this.. and if its not registered we fall back and background the game on a Menu press. if it is registered we call Handled.. if it returns true. we ignore the Menu event .. if not we pass it up the chain to tvOS and let it background the app. This safely decouples us from and hacks..

PS naming can be better..

@tomspilman @KonajuGames ?

**Comment by KonajuGames** [comment]
It would be cleaner and would allow the several platforms that require special handling of the Back/Menu button to do it properly.

**Comment by dellis1972** [comment]
Naming suggestion

```
public interface IPlatformBackButton {
   bool Handled();
}
``` 

comments?

dellis1972:
ok added a ```IPlatformBackButton``` @tomspilman @KonajuGames  let me know what you think.

KonajuGames:
I think that looks ok.

dellis1972:
ok menu system all working :) even added some docs!!

//cc @tomspilman @KonajuGames

DDReaper:
FYI @dellis1972 I've taken your TVOS platform updates for Protobuild and got them included in to Protobuild it'self.  So TVOS will be safe when we pull in the latest version

dellis1972:
@tomspilman @KonajuGames this is good to go from my point of view. I got feedback and it all seems to be working.

KakCAT:
Hi @dellis1972 ,

I'm having problems with the controllers in the real hardware, same problem described in https://github.com/mono/MonoGame/pull/4240#issuecomment-158634868

Controllers are initially unset and must be assigned a controller.PlayerIndex value before using them, otherwise the controller.PlayerIndex never match the index parameter and acts as no gamepads were found.

dellis1972:
@KakCAT so we have to assign an index... that's not what we see in the simulator.. and its a bit .. weird, but hey its apple.. In addition in the simulator we see the same controller in the controllers list twice.. once a a "remote" and once as a extendedGamepad.. do you get the same on a real device or is that just an artifact of the simulator..

dellis1972:
@KakCAT Added some code to assign the index. Can you test on hardware please :)

dellis1972:
actually scratch that. made a boo boo

KakCAT:
HI @dellis1972 ,

the Apple remote is only shown as a micro controller. The new code seems to work but I had to put the "var state= new GamePadState" in the inner loop, otherwise all controllers are assigned the same number

KakCAT:
Hi!

the latest commit also works, tested with one apple remote and one MFi controller.

I tested IPlatformBackButton which also works.

dellis1972:
@KonajuGames @tomspilman bump :) I think this is ready to go.

hach-que:
@dellis1972 You should wait until #4282 because that drops the XSLT files entirely (and has tvOS merged upstream).

hach-que:
(After #4282 you can remove your XSLT changes here entirely)

dellis1972:
@hach-que we have a bit of a deadline for tvOS and I would like this in the next release. I'm not sure if the Mac Changes will make it in time.. unless that PR is completely done?

hach-que:
The Mac PR is completely done, it just needs merging

KakCAT:
Hi!

there's a bug in the controller Index assignment in PlatformGetState / PlatformGetCapabilities

The list of controllers "GCController.Controllers" order is not deterministic.

In example:

Start game, ControllerA is connected and detected. Assigned Player Index 1.
After some time, you connect another controller ControllerB.
Sometimes (not always, but usually, I'd say 50-50) ControllerB comes first, sometimes ControllerA comes first. If ControllerB comes first in "GCController.Controllers", then the Index 1 is assigned, ending with both controllers with Index1. Controls are a bit caotic after that ;)

KakCAT:
static bool indexIsUsed (GCControllerPlayerIndex index)
        {
            foreach (var ctrl in GCController.Controllers)
                if (ctrl.PlayerIndex==index) return true;

            return false;
       }


and then checking index is not used before assigning it does the trick.

dellis1972:
@KakCAT fixed

KakCAT:
I'm not entirely sure because I can't profile on TVOS, but List<Buttons> and buttons.ToArray() is probably generating garbage. (it can be addressed in a later PR as it's probably minimal)

KonajuGames:
How is this going now?

dellis1972:
@KonajuGames I think this is good to go.. we have a bit of garbage clean up to deal with but seeing as tvOS is still in preview as far as Xamarin is concerned I think this is a good start.

KakCAT:
@KonajuGames I've submitted 2 games with this PR and no had problems so far on MonoGame's side (lots on Xamarin's side though). I failed review due to incorrect in-app usage but apparently this is the only problem Apple has found so far.

KonajuGames:
Ok.  I'll merge this then.  Thanks for everyone in getting this done.

tomspilman:
@dellis1972 - What is the status of this?

I didn't notice tvOS in the `default.build` nant script... shouldn't we be building and packaging this now?

dellis1972:
We have to wait until xamarin release the tvOS support in stable , then upgrade the mac bot before we can enable it in nant.

-----Original Message-----

Issue #4069, opened 2015-08-11
BasicEffect / Model processor Lighting issues


nilllzz:
Hello there,

I encountered a few issues with the lighting features of the BasicEffect and its relation to the Model Processor of MonoGame.
  1. The BasicEffect lighting does not seem to apply correct (tested with the EnableDefaultLighting), it just casts the same light on all parts of the model, ignorning direction. This might be directly related to 
  2. The Model Processor seems to somehow mess up normals of the model (or something else related to lighting, I am not sure). This causes light applied by the BasicEffect to look the same for the entire model and appear black.

I have an illustration of this occurence here, compared to XNA's output:
![wxpsjlt](https://cloud.githubusercontent.com/assets/2119926/9210458/b9a0efe4-407c-11e5-82c2-47c3adb501ec.png)

(Also, on that note, GraphicsDeviceManager.PreferMultiSampling does not seem to prefer multi sampling, it is activated on all models on those screenshots above, but does not seem to apply in MonoGame [yes, GraphicsDeviceManager.ApplyChanges() has been called]).

InfiniteProductions:
I have a similar issue, a model from a FBX 2014 ascii file, exported from 3DSMax 2015 works fine with XNA. same model, same code, MonoGame only render a full black model, with LightingEnabled set to true or false. Also tried with the today dev snapshot of MG, same result.

I'll add test code later.

mgarstenauer:
@InfiniteProductions I did a quick test and everything is working here. (My dev snapshot is from December.)

I see that you are referencing `H:\MonoGame-dev\Windows\AnyCPU\Debug\MonoGame.Framework.dll` in your `FBXMP-MGdev_dn45.csproj`. But the Content.mgcb is built using the standard MSBuild task `$(MSBuildExtensionsPath)\MonoGame\v3.0\MonoGame.Content.Builder.targets`.

Perhaps you are using the dev snapshot at runtime, but the content project is compiled using the old binaries in `C:\Program Files (x86)\MSBuild\MonoGame\v3.0`?

nilllzz:
According to the pull, it was resolved in November, so using your December build of MonoGame, it might just work because it got fixed.

InfiniteProductions:
@mgarstenauer : doohhh!  Is it a common issue or something specific to using two, or more, sets of MG on the same computer ? (I have a stable install, 3.x, x=4 maybe) + the dev snapshot and I wasn't aware at all of this use of MSBuild path.

[...]

PR #4244, opened 2015-11-15
Fix for content pipeline models being drawn entirely black

PikeUK:
*Disclaimer: I know nothing about 3D graphics, XNA or MonoGame (aside from simple SpriteBatch stuff) so everything below could be wrong.*

This push request is a quick fix to solve the black model problem with the content pipeline. Processing a model with MonoGame's pipeline and then drawing it using model.Draw() causes it to be drawn entirely black. However, if you process the model with XNA's pipeline and then load and Draw() the xnb in MonoGame it displays as expected. See these threads for more information on the problem:

https://github.com/mono/MonoGame/issues/4069
http://community.monogame.net/t/3d-models-appear-black/2596
http://community.monogame.net/t/problem-with-3d-objects-textures/2474
http://community.monogame.net/t/diagnosing-rendering-issues-for-3d-model-in-new-content-pipeline/1432

The problem occurs here:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs#L320

The pipeline checks if Assimp has imported a particular lighting property from the model's material, and if so, sets the relevant nullable property in a BasicMaterialContent object. If the particular lighting property is not found then the relevant BasicMaterialContent property remains null. When the BasicMaterialContent is serialized it uses GetValueOrDefault() which sets all the unfound lighting properties to the default float value of 0. Thus the resulting xnb model has Alpha and SpecularPower set to 0 causing it to be drawn entirely black:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Serialization/Compiler/BasicEffectWriter.cs

Assimp .NET seems to provide sensible defaults when getting a lighting property for which no corresponding property was found in the model's material:

https://github.com/assimp/assimp-net/blob/master/AssimpNet/Material.cs#L253

So this patch changes the pipeline to always accept the value provided by Assimp .NET, which will be either the property from the model or the default value provided. This solves the black rendering problem on the handful of models I have tried (on Windows 10) but I don't know if it is a correct or appropriate fix.

As a final note here are the final lighting values from importing a simple untextured model in various ways (unlisted properties do not differ between the three methods).

XNA Pipeline
------------
Alpha = 1
DiffuseColor = 1 1 1
SpecularColor = 1 1 1
SpecularPower = 16

Current MonoGame Pipeline
-------------------------
Alpha = 0
DiffuseColor = 0.8 0.8 0.8
SpecularColor = 0 0 0
SpecularPower = 0

Patched MonoGame Pipeline
-------------------------
Alpha = 1
DiffuseColor = 0.8 0.8 0.8
SpecularColor = 0 0 0
SpecularPower = 1

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
@@ -311,20 +311,15 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
                     material.Textures.Add("Bump", texture);
                 }
 
-                if (aiMaterial.HasColorDiffuse)
-                    material.DiffuseColor = ToXna(aiMaterial.ColorDiffuse);
+                material.DiffuseColor = ToXna(aiMaterial.ColorDiffuse);
 
-                if (aiMaterial.HasColorEmissive)
-                    material.EmissiveColor = ToXna(aiMaterial.ColorEmissive);
+                material.EmissiveColor = ToXna(aiMaterial.ColorEmissive);
 
-                if (aiMaterial.HasOpacity)
-                    material.Alpha = aiMaterial.Opacity;
+                material.Alpha = aiMaterial.Opacity;
 
-                if (aiMaterial.HasColorSpecular)
-                    material.SpecularColor = ToXna(aiMaterial.ColorSpecular);
+                material.SpecularColor = ToXna(aiMaterial.ColorSpecular);
 
-                if (aiMaterial.HasShininessStrength)
-                    material.SpecularPower = aiMaterial.ShininessStrength;
+                material.SpecularPower = aiMaterial.ShininessStrength;
 
                 _materials.Add(material);
             }
--- a/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
@@ -311,15 +311,20 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
                     material.Textures.Add("Bump", texture);
                 }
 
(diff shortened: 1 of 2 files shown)


mgarstenauer:
Good work finding the cause of the problem! But the proposed fix solves the problem only for the OpenAssetImporter. Other model importers will still produce wrong values. The problem needs to be fixed in the BasicEffectWriter where the wrong default values are written. 

    class BasicEffectWriter : BuiltInContentWriter<BasicMaterialContent>
    {
        protected internal override void Write(ContentWriter output, BasicMaterialContent value)
        {
           ...
            output.Write(value.DiffuseColor ?? new Vector3(1, 1, 1));
            output.Write(value.EmissiveColor ?? new Vector3(0, 0, 0));
            output.Write(value.SpecularColor ?? new Vector3(1, 1, 1));
            output.Write(value.SpecularPower ?? 16);
            output.Write(value.Alpha ?? 1);
            output.Write(value.VertexColorEnabled ?? false);
        }
    }

MonoGame should use the same default values as XNA. SpecularColor = 0 0 0 is not a good default because as it disables specular lighting.
Can you move the fix from OpenAssetImporter to the BasicEffectWriter and the check if this new fix works with your models? Thx.

KonajuGames:
Merging.  Thanks @PikeUK.

InfiniteProductions:
Hi everyone,

despite the fix, it looks like there is/are still some issue(s) leading to the same full black model, see my comment on issue #4069 (which include a test code).
I use a dev snapshot of MG which include the fix in BasicEffectWriter.cs file with VS2013 on windows 7 64b.

With or without lighting related parameters (ambient + one directional light), with or without specular colour and level defined, model is drawn black.
I would like to understand what is happening here and if I made a mistake or if it still the MG issue on action.

PR #4245, opened 2015-11-15
Added Mac documentation to config file

cra0zy:
#4228


tomspilman:
Looks like this fixed it... thanks!

PR #4246, opened 2015-11-16
Enable gamepad vibration for Windows DirectX

MichaelDePiazzi:
This PR enables gamepad vibration for Windows DirectX.

The code was actually already there, but only enabled for `DIRECTX11_1` builds. So I guess at some point this wasn't supported by SharpDX on Windows desktop? But they must have since added support for it as Windows Desktop is listed as a supported platform here: http://sharpdx.org/documentation/api/m-sharpdx-xinput-controller-setvibration

I've tested this on Windows 7 and it works correctly.


tomspilman:
Yeah... I've hacked this locally on one of my prototypes.  This should work fine.

@KonajuGames ?

KonajuGames:
Looks good to me.  Merging.

PR #4247, opened 2015-11-16
[Pipeline] Fix black background on file icons

nkast:


Tools/Pipeline/Windows/ContentIcons.cs
```diff
@@ -25,6 +25,7 @@ public class ContentIcons: IDisposable
         public ContentIcons()
         {
             Icons = new ImageList();
+            Icons.ColorDepth = ColorDepth.Depth32Bit;
```
tomspilman: Was the default here different depending on your PC or something?

nkast: The default value is always `Depth8Bit` both when you run under x86 or x64.
Perhaps I was compiling with x86 (=edit & continue) when I was testing that, or perhaps all my associations where good at the time, Who knows... 
I double check with the VM machine (although I updated to 1511 since then) and the latest develop build and as you can see not all icons have this problem. 
pipelinewin10vm

nkast: Ok, so now -without the ColorDepth fix- I see black background whether I build for x86 or x64!
This is getting very weird!

Setting the ColorDepth to 32bit always fix it, so no worries!

tomspilman: Very weird... but if it works then it works.  I don't see any potential side effects from this.

PR #4248, opened 2015-11-17
Remove unused method 'Set(...)' from SpriteBatchItem.

nkast:
This method doesn't calculate Rotation and origin, probably a leftover
of some earlier optimized execution path.


tomspilman:
Great... thanks!

PR #4233, opened 2015-11-08
Adds MathHelper.LerpPrecise

hede:
Following discussion on #4204, this PR implements the proposed MathHelper.LerpPrecise.

I've also fixed a few apparent typos in the tests, added an XNA compat. test for current Lerp behaviour and added the relevant tests for LerpPrecise.

PR #4250, opened 2015-11-17
Fix XNA test regressions

polsys:
The XNA test project was broken by a couple recent(-ish) PRs.

* As part of #4098 the 3D cube components were updated to use a `DrawIndexedPrimitives` overload that does not exist in XNA. Reverted the calls back to the old overload, which internally uses the new code.
* Also as part of #4098 a test for the new overload was added. Added `#if !XNA` around the test.
* The test for `MathHelper.LerpPrecise` function introduced in #4233 was included in the XNA project too. Added `#if !XNA` around the test.

As a disclaimer, I only have the XNA redist installed at the moment, so I had to exclude the Content Pipeline tests from compilation on my machine. Therefore I cannot guarantee that the project is completely buildable even with these changes, and this must be given a full test before merging.

I assume that #3550 is dead? Besides, how important are these compatibility tests anymore, as they can be broken for months and XNA has been long dead? We have a few missing features to implement and some edge cases for which crosstesting will be useful, but otherwise these seem to be neglected.

### DIFF
--- a/Test/Framework/Components/Colored3DCubeComponent.cs
+++ b/Test/Framework/Components/Colored3DCubeComponent.cs
@@ -136,7 +136,7 @@ namespace MonoGame.Tests.Components {
 			foreach (EffectPass pass in basicEffect.CurrentTechnique.Passes) {
                 pass.Apply();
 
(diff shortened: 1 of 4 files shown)


KonajuGames:
Merging this.  Thanks @polsys.

PR #4251, opened 2015-11-18
Remove DisplayMode.RefreshRate (to match XNA)

### LINKED ISSUE #4221: "DisplayMode.RefreshRate" can be incorrect (Windows DirectX)
I've come across a couple of weird refresh rates such as 59.94006Hz and 59.96894Hz. In the Windows monitor settings, they appear as 59Hz (i.e. truncated). But in MonoGame, they're being rounded to 60Hz.

Normally I guess this wouldn't be much of an issue. But on one of my PCs I have both 59Hz and 60Hz modes, so in MonoGame they are both appearing as 60Hz and look like duplicates.

So the refresh rate needs to at least be truncated to match Windows. However, given that `DisplayMode.RefreshRate` seems to be a MonoGame only property (I couldn't find it in XNA), is there any reason why we can't just make it a float so that the precise refresh rate can be retrieved?

PS. I'm more than happy to submit a PR for this. But I just wanted to confirm how this should be resolved first.

MichaelDePiazzi:
Resolves #4221

`DisplayMode.RefreshRate` was removed in XNA 4 (Shawn Hargeaves explains some of the reasons why this was removed [here](http://xboxforums.create.msdn.com/forums/p/70732/432151.aspx#432151)). This PR removes this property from MonoGame to match XNA.

### DIFF
--- a/MonoGame.Framework/Graphics/DisplayMode.cs
+++ b/MonoGame.Framework/Graphics/DisplayMode.cs
@@ -41,7 +41,6 @@ namespace Microsoft.Xna.Framework.Graphics
 
         private SurfaceFormat format;
         private int height;
-        private int refreshRate;
         private int width;
 
         #endregion Fields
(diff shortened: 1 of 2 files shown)


KonajuGames:
Merging.  Thanks @MichaelDePiazzi.

PR #4253, opened 2015-11-20
Fixed ContentManager

cra0zy:
In #3863 I set 'w' instead of 'g', 'w' is WindowsDX, 'g' is WindowsGL.


tomspilman:
Thanks @cra0zy !

PR #4254, opened 2015-11-20
[MGCB] Write averageBytesPerSecond and blockAlign for SoundEffect

nkast:
Get averageBytesPerSecond from bitRate (=bitsPerSecond) and calculate
blockAlign for PCM format.

With this fix SoundEffect XNBs can be used again from XNA. The move to
ffmpeg (https://github.com/mono/MonoGame/pull/3197) broke backward compatibility.


KonajuGames:
Merging.  Thanks @nkast.

PR #4257, opened 2015-11-21
Fedora Support

cra0zy:
Just an automatic dependency resolver for Fedora 23 or newer.

### DIFF
--- a/Installers/Linux/RUN/Dependencies/dependencies.sh
+++ b/Installers/Linux/RUN/Dependencies/dependencies.sh
@@ -21,4 +21,13 @@ then
 	fi
 fi
 
+if [ $OS = Fedora ]
+then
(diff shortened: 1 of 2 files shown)


KonajuGames:
Is Linux the most difficult platform to support?  All the different fractured versions.

Merging.  Thanks @cra0zy.

cra0zy:
> Is Linux the most difficult platform to support? All the different fractured versions.

Definitely. My logic for supporting Linux is support Ubuntu with full native .deb files, and then support everything else with .run installer(and add dependencies from time to time while testing a new distro).

> Merging. Thanks @cra0zy.

You are a bit forgetful :P

KonajuGames:
Yup.  Forgot to hit Merge.

PR #4258, opened 2015-11-21
[installer] Fix installer to include required files.

dellis1972:
The file mask was excluding any file which did NOT contain
a .
This fixes that and adds commands to fix up the permissions
of ffmpeg and ffprobe


dellis1972:
Thanks for @kjpou1 for finding this problem :)  turns out we weren't shipping the right files in the Tools folder on Mac.

KonajuGames:
Merging.  Thanks @dellis1972 and @kjpou1

PR #4259, opened 2015-11-22
[Pipeline] Remove default VertexColor Channel

nkast:
MGCB always add an extra color vertex channel to Models.
XNA 4.0 pipeline never did that with Models, at least on the two platforms I tested it with (Windows/Phone).


nkast:
MGCB always add an extra color vertex channel to Models.
XNA 4.0 pipeline never did that with Models, at least on the two platforms I tested it with (Windows/Phone).

tomspilman:
@mgarstenauer - You might know more about this than I do.  This make sense to you?

mgarstenauer:
The original code is definitely smelly: `ProcessVertexChannel` always adds a vertex color channel, but the subsequent code in `ProcessGeometryUsingMaterial` handles meshes with and without vertex color channel in?!

I agree with @nkast. I am fairly certain that XNA didn't add a default vertex color channel and I don't see a reason why there should be one. The default vertex color channel should be removed.

tomspilman:
Thanks @mgarstenauer .  Merging!

nkast:
Thanks!

PR #4268, opened 2015-11-26
Implements integer array EffectParameter setting.

Soenneker:
Implements EffectParameter.SetValue() for int[] objects.


tomspilman:
So looking at the implementations of `SetValue(float[])` and `SetValue(Vector2[])`...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/EffectParameter.cs#L789
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/EffectParameter.cs#L830

They seem to implement things in a slower manner.  I wonder if this is the "right" way or just a slow way.

I suppose since you tested your changes here that directly setting it on the data is fast and should work.  So why do we have the slow array setting method at all in here?

Any ideas @dellis1972 @tgjones @KonajuGames ?

KonajuGames:
There's no reason those other array methods must call the single element methods.  They can just set the data directly themselves.

Soenneker:
@tomspilman Perhaps the decision to set float[] and other types of arrays that way were made to avoid the #if DIRECTX directives. I have not run into any issues setting the values directly.

tomspilman:
> Perhaps the decision to set float[] and other types of arrays that way were made to avoid the #if DIRECTX directives. 

Actually it looks like that code comes from the original shader support added in 2011:

https://github.com/mono/MonoGame/commit/84476b59be6969a9b34f36a85d0dfe4dd342acdc

There was no DX at the time and this was the simple implementation that worked... even if it was slow.

So really none of them should be using that path anymore unless we absolutely have to.

Soenneker:
I added another commit that sets the values directly instead of calling SetValue().

tomspilman:
Thanks @Soenneker .

A few more things to do here if you don't mind.

First we probably should add the same invalid cast tests as we have on the other setters:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/EffectParameter.cs#L444

We should also test XNA to see what it does when you try to set a singular value on a array type.  Does it allow it?  If so do we observe the same behavior.

Finally if we've changed the setters to direct access to the `Data` then maybe the getters should be fixed to do the same.  Currently the getters use `Element`.

Soenneker:
Appreciate the reply @tomspilman

The invalid cast tests can be implemented quickly.

I won't be able to compare XNA to MonoGame as I don't really have the time, but I invite someone else of the community to do so.

Unfortunately, `Data` in `EffectParameter` is null when an array type is being set, and it's due to this line during `Effect` initialization:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/Effect.cs#L434
(elements.Count != 0 due to how the array elements are read) Any thoughts on where you want to go with this?

I feel like this has strayed a lot from the initial intent of the pull request; I just meant to add integer array functionality.

PR #4272, opened 2015-11-27
[Pipeline] Add FilterOutput view

nkast:
Add a Filtered output view similar to https://github.com/mono/MonoGame/pull/4168.

There are few hacks to hide the Header from the TabControl and also to scroll the treeview without flickering.

One difference from GTK version is that 'Cleaning' show short paths relative to project Output Directory.

Another is that I didn't merge the two views under one control , everything in Winforms is public and visible from MainForm, the original _outputView had already too many references and not sure the best way to do this in WinForms.

### DIFF
--- a/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
+++ b/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
@@ -42,6 +42,11 @@ namespace MonoGame.Tools.Pipeline.Windows.Controls
             outputUri = new Uri(pod);
         }
 
+        protected override void OnBeforeSelect(TreeViewCancelEventArgs e)
+        {
+            // disable selection for all nodes
+            e.Cancel = true;
+        }
 
         internal void AppendText(string text)
         {
--- a/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
+++ b/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
@@ -87,14 +87,6 @@ namespace MonoGame.Tools.Pipeline.Windows.Controls
                     AddSubItem(tn, text);
                     break;
                 case OutputState.BuildError:
-
-                    if (_prevFilename != outputParser.Filename)
-                    {
-                        tn = AddItem(BuildIcons.Processing, "Building " + GetRelativePath(outputParser.Filename));
-                        tn.ToolTipText = outputParser.Filename;
-                        AddSubItem(tn, outputParser.Filename);
-                    }
-
                     _lastTreeNode.ImageIndex = BuildIcons.Fail;
                     _lastTreeNode.SelectedImageIndex = BuildIcons.Fail;
                     _lastTreeNode.ToolTipText += Environment.NewLine + Environment.NewLine + outputParser.ErrorMessage;
--- a/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
+++ b/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
@@ -5,6 +5,7 @@
 using System;
 using System.IO;
 using System.Windows.Forms;
+using MonoGame.Tools.Pipeline.Common;
 
 namespace MonoGame.Tools.Pipeline.Windows.Controls
 {
@@ -15,8 +16,11 @@ namespace MonoGame.Tools.Pipeline.Windows.Controls
         [System.Runtime.InteropServices.DllImport("user32.dll", CharSet = System.Runtime.InteropServices.CharSet.Auto, SetLastError = false)]
         static extern IntPtr SendMessage(IntPtr hWnd, uint Msg, int wParam, int lParam);
 
-        BuildIcons _buildIcons;
-        TreeNode tmpIter;
+        private BuildIcons _buildIcons;
+        private TreeNode _lastTreeNode;
+
+        private OutputParser outputParser;
+        private string _prevFilename;
         
         Uri folderUri;
         Uri outputUri;
@@ -25,8 +29,8 @@ namespace MonoGame.Tools.Pipeline.Windows.Controls
         {
             this._buildIcons = new BuildIcons();
             this.ImageList = _buildIcons.Icons;
+            outputParser = new OutputParser();
         }
-        
         
         internal void SetBaseFolder(IController controller)
         {
@@ -40,6 +44,8 @@ namespace MonoGame.Tools.Pipeline.Windows.Controls
                 pod += System.IO.Path.DirectorySeparatorChar;
             pod = Path.Combine(pl, pod);
             outputUri = new Uri(pod);
+
+            outputParser.Reset();
         }
 
         protected override void OnBeforeSelect(TreeViewCancelEventArgs e)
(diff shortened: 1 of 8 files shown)


cra0zy:
Can you post a screenshot of how it looks?

nkast:
pipelinea

pipelinee


pipelinec

nkast:
@cra0zy Yeap, I hade them ready, now uploading...

cra0zy:
Sweet :+1:

Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            string pod = ((PipelineController)controller).ProjectOutputDir;
+            if (!pod.EndsWith(System.IO.Path.DirectorySeparatorChar.ToString()))
+                pod += System.IO.Path.DirectorySeparatorChar;
+            pod = Path.Combine(pl, pod);
+            outputUri = new Uri(pod);
+        }
+
+
+        internal void AppendText(string text)
```
tomspilman: We should eventually figure out a way to share the logic here with the GTK version.  So we have the same filters and behaviors between them moving forward.

tomspilman:
Looks good.

Maybe next step here is working out some icons to use here instead of the plain squares.  Probably doesn't need to be very fancy at all.  In VS you get these:

err
warn

Maybe some play on that concept.

Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
nkast: Sure, there are a few differences between Linux/windows. 
lines https://github.com/nkast/MonoGame/blob/PipelineFilterOutput/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs#L73  - Cleaning relative to outputDir,
https://github.com/nkast/MonoGame/blob/PipelineFilterOutput/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs#L54 trimming input from \n\r
https://github.com/nkast/MonoGame/blob/PipelineFilterOutput/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs#L77  use File.Exists() to detect filenames/paths

@cra0zy , any change to converge the two sources?

tomspilman: > Sure, there are a few differences between Linux/windows. 

Those differences look pretty manageable.  Seems like we just need to have some sort of simple "parser" method that calls an interface to do the platform specific work.

nkast:
We can use some familiar icons from the VS library
http://www.microsoft.com/en-us/download/details.aspx?id=35825

Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
cra0zy: > https://github.com/nkast/MonoGame/blob/PipelineFilterOutput/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs#L73 - Cleaning relative to outputDir,

I think Gtk Pipeline does the same.

> https://github.com/nkast/MonoGame/blob/PipelineFilterOutput/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs#L54 trimming input from \n\r

\r\n trimming shouldn't cause any harm on gtk pipeline

> https://github.com/nkast/MonoGame/blob/PipelineFilterOutput/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs#L77 use File.Exists() to detect filenames/paths

Your file exist logic is flawed, what if the error is about file not existing?

cra0zy:
> We can use some familiar icons from the VS library
> http://www.microsoft.com/en-us/download/details.aspx?id=35825

Sweet, we could use those for the toolbar as well.

tomspilman:
First I guess we need to be sure of the license on those icons.  Are they truly free for us to use in MonoGame?

Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
cra0zy: Solution for not using File Exist should be to check if the second character of path is ":" or if path starts with "\\", I think,

nkast:
I used them in the past a couple of times but I am not quite sure how restrictive terms they have.

cra0zy:
From the license:

3.Distribution Restrictions. You may not
 - distribute Distributable Code to run on a platform other than Microsoft operating systems,  run-time technologies or application platforms;

tomspilman:
Well that is a no then...  still we can crib ideas for icons at least.

nkast:
Oh, too bad.

Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
nkast: See the screenshot how it says, 
Cleaning GeometryA.xnb
Building GeometryA.xnb
For doing this I had to expose the project.OutputDir from pipelineController.
In your case the first line will be something like:
Cleaning \bin\Linux\GeometryA.xnb
Similarly I had a project where the output is a couple of folders down and then a few folders up, and the output was like
Cleaning ../../../Game/projCodename/projCodename.WindowsPhone/Content/GeometryA.fbx
This goes again the clean output you were going for.

nkast: Yes. I did like file.exists either. I was looking for a .IsPath/.IsFilename type if method but the closest was .IsRooted which means that using exceptions for code flow. I thought that since the name comes ftom mgcb then I would get an error before the file but really never tested it.
I guess what I need here is a Regular expression to test if the line is a file path and just that.

tomspilman: > I guess what I need here is a Regular expression to test if the line is a file path and just that.

Yeah... a nice helper to have for these sorts of things.

nkast:
Considering the issue with how to check for next build, I created two common scenarios, C.fbx is a missing file while A.fbx has it's texture 'obj.png' missing.

pipelinemissingfile

The missing texture is reported as an error of A.fbx and correctly grouped under A.fbx.
The missing C.fbx is reported in a single line as an error and incorrectly grouped under B.xml issues.

changing from `file.exists(text)` to `else if (text.Substring(1,1)==":" && !text.Contains("error")) `
has exactly the same results. The reason is that MGCB reports a missing file in a single line (filename + error).

I did a quick research on regular expressions and the situation is very complicated. All solutions out there focus on only windows platforms and I don't have the capacity to evaluate to see which one is working. Right now it wont make a difference since MGCB reports a missing file in a single line (filename + error), it will still report the output as a non-file = error message of the previous file.

If we assume we don't touch MGCB output and attack this from Pipeline/GUI alone we have to do some more advanced parsing, parse the first part of the error, extract the filename and compare it with the previous line.

tomspilman:
>  The reason is that MGCB reports a missing file in a single line (filename + error).

I'm fine with fixing that so that it always reports the file it tried to process first then any errors.  Seems like a reasonable thing to do for consistency in the output anyway.

nkast:
Then I guess I have to switch to check for [Letter+':'], at least until we have a better parser in place.

tomspilman:
> If we assume we don't touch MGCB output 

To be clear I meant that we could fix the MGCB output to make it always report the file being processed first.

nkast:
> To be clear I meant that we could fix the MGCB output to make it always report the file being processed first.
Sounds like a reasonable change.

nkast:
I just updated the Parser, solve all the current issues with MGCB output.
@cra0zy Is that something you can work with? I place 'OutputParser.cs' in the \Common\ folder.

pipelineparserexception

pipelineparsermissingfile

nkast:
@tomspilman 
https://github.com/nkast/MonoGame/blob/PipelineFilterOutput/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs#L92-L97
This does the trick for now...

cra0zy:
> @cra0zy Is that something you can work with? I placed OutputParser.cs in the \Common\ folder.

Yea, I can work with that, I'll update Gtk Pipeline as soon as this is merged.

PS. I think you could have done it with less code by setting up 2 new IView interfaces and just calling them instead of setting up state.

nkast:
About the icons, until we have something better, how about we use circles/bullets? Or simply crop them a few pixels? (square bullets)

cra0zy:
> About the icons, until we have something better, how about we use circles/bullets? Or simply crop them a few pixels? (square bullets)

Just leave them as they are for now, no need to replace something that's temporary with something that's temporary as well.

Also especially don't crop them, that would cause trouble for Gtk Pipeline.

nkast:
No cropping the size, keep them 16x16 with transparency. Make them smaller squares so they will be somewhat balanced with the fonts.

Tools/Pipeline/Windows/Controls/FilterOutputControl.cs
```diff
@@ -0,0 +1,161 @@
...
+                    AddSubItem(tn, text);
+                    break;
+                case OutputState.BuildError:
+
+                    if (_prevFilename != outputParser.Filename)
+                    {
+                        tn = AddItem(BuildIcons.Processing, "Building " + GetRelativePath(outputParser.Filename));
+                        tn.ToolTipText = outputParser.Filename;
+                        AddSubItem(tn, outputParser.Filename);
+                    }
```
nkast: we can remove this after #4366.

cra0zy: Just reminding you that you can remove them now :)

tomspilman:
@nkast - This good to merge?  Also I assume the filter view is off by default at the moment?

cra0zy:
> Also I assume the filter view is off by default at the moment (I assume this is true for Linux/Mac as well)?

It's on by default on Linux/Mac.

tomspilman:
> It's on by default on Linux/Mac.

Well then lets make it consistent and have it on by default on Windows too.  I still worry it will confuse people that were using the tool already to suddenly have this.

nkast:
|  This good to merge? Also I assume the filter view is off by default at the moment (I assume this is true for Linux/Mac as well)?

Yeap, I won't touch it again. The filter view is on by default as seen on the GTK UT video by @cra0zy.

KonajuGames:
Merging.

nkast:
That's great! Thanks @KonajuGames

cra0zy:
@nkast Found a bug in your Regex parsing on Build/Rebuild/Clean, with:
```
/home/harry/Documents/test.mp3
```
It returns:
```
home/harry/Documents/test.mp3
```
for the filename.

nkast:
The + is not needed. The idea was to skip any white space for future proofing. Feel free to remove it if you like.

 Perhaps /s would have been a better  choice.

cra0zy:
Another bug in the new filter output system:
fobug
Game01.mp3 compiled, while Game02.mp3 failed

@nkast I am currently preoccupied with some other stuff so if you have some time, please look into it.

nkast:
@cra0zy from the screenshot it looks like mp3 report errors differently than other assets. Not something to fix on the filter parser.  I am not gonna be home today but that's contently something we need to fix.

tomspilman:
The message "The source file X does not exists!" is coming from here:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs#L558

Not sure how the indent was lost for that.

nkast:
The parser doesn't look for indent but for filenames. The problem is that line 558 comes before 567. For it to work we could to move the exception (L558) after L567. In that case the output would be:
```
C:/Assets/missingAsset.ext
C:/Assets/missingAsset.ext: error: The source file 'C:/Assets/missingAsset.ext' does not exist!
```  

if that doesn't make sense for the user and missing assets should output an error immediately, we can revert the last commit (5b453c8) that handle cases like the the above.
https://github.com/nkast/MonoGame/commit/5b453c86221a721520965b17774248638f8a9983


As for the indent that's a bit more complicated, additionally to moving the check inside the try block we need replace the exception with logging the error in place and then return.

tomspilman:
> In that case the output would be:

That seems reasonable to me.  Maybe instead of moving the exception maybe just add `Logger.LogMessage("{0}", pipelineEvent.SourceFile);` within the `if` and before the `throw`.  Should be less likely to break existing behavior.

tomspilman:
@nkast - Another issue I came across with the filter view.  Sometimes you get fatal error messages during loading of the .mgcb file:

err

This message doesn't appear in the filter view.  What do we need to change here to make sure it does appear?  Is it just in the formatting of the string?

nkast:
| This message doesn't appear in the filter view. What do we need to change here to make sure it does appear? Is it just in the formatting of the string?

Not just that, the filterview expect output from MGCB and try to parse that, we have to change it to "rootedfilename: error: error message" or add a new RE match rule, like that https://github.com/nkast/MonoGame/commit/21101236e7ca78b2aee1881afab169848589e7ab

but we still need to bring back that ugly check to https://github.com/nkast/MonoGame/blob/21101236e7ca78b2aee1881afab169848589e7ab/Tools/Pipeline/Windows/Controls/FilterOutputControl.cs#L89-L97 
in order to create a new rootTree with the filename.

@cra0zy what do you think? Perhaps we can add a new Method to the control instead of AppendText(). Like .WriteWarning() or something.

@crazy Any thoughts?  
https://github.com/nkast/MonoGame/commit/21101236e7ca78b2aee1881afab169848589e7ab

PR #4273, opened 2015-11-27
Fixed index check for gamepad

cra0zy:
Thanks @Helikin for pointing it out.

#4083


Helikin:
Great, thanks for the quick fix. 

One minor note:
If the method is called **PlatformGetMaxIndex**, it is not clear from the method name alone whether the index is the inclusive or exclusive upper bound. This could be avoided if the method is named
**PlatformGetMaxNumberOfGamePads**, **PlatformGetMaxIndexExclusive** or if a comment is added.

KonajuGames:
Thanks @cra0zy and @Helikin.  Merging.

PR #4274, opened 2015-11-29
Add automated.build so I can track MonoGame upstream on my build server

hach-que:
I expect this not to initially work for all of the platforms, but I'm submitting the PR so it runs on the MG build server.

This uses the new automated build script functionality in Protobuild to generate and build for all relevant platforms.  It simplifies the `default.build` script greatly, and removes a bunch of conditional logic that is no longer necessary.

I'd like to get the `automated.build` script also running the tests, but since these run through NUnit and make use of a NAnt script, I don't want to do that until I know whether or not it'll impact the reporting of tests in TeamCity.

(I also removed the old GenerateProject / GenerateSolution files since they're no longer needed with WindowsUniversal and tvOS being merged into Protobuild).

### DIFF
--- a/automated.build
+++ b/automated.build
@@ -1,0 +1,18 @@
+#version 1
+
+set build-target Rebuild
+set build-property Configuration Release
+set execute-configuration Release
+
+if host Windows
+  set target-platforms Windows,WindowsGL,Linux,Web,Android,Ouya,Windows8,WindowsPhone,WindowsPhone81,WindowsUniversal
+endif
+if host MacOS
+  set target-platforms Linux,MacOS,iOS,Android,Ouya
+endif
+if host Linux
+  set target-platforms Windows,WindowsGL,Linux,Web
+endif
+
+generate
+build


tomspilman:
We have no need or desire for Protobuild to automate the build.  NAnt is our solution to build automation and we're sticking to it.

hach-que:
I still want to make sure the functionality works though, hence this PR to test it on the MonoGame build server.

tomspilman:
> I still want to make sure the functionality works though,

Ok.

The Win10 build depends on the Nant target `build_windows10` existing.  For your testing purposes I suggest you add it back, but have it do nothing.

No Mac it looks like Protobuild failed to find some file:

```
[07:48:17]+ /Applications/buildAgent/work/f26e5a04a66fcbdb/Protobuild.exe --generate Linux,MacOS,iOS,Android,Ouya 
[07:48:17]ERROR: ApplicationName='/Applications/buildAgent/work/f26e5a04a66fcbdb/Protobuild.exe', CommandLine='--generate Linux,MacOS,iOS,Android,Ouya ', CurrentDirectory='/Applications/buildAgent/work/f26e5a04a66fcbdb', Native error= Cannot find the specified file
[07:48:17]  at System.Diagnostics.Process.Start_noshell (System.Diagnostics.ProcessStartInfo startInfo, System.Diagnostics.Process process) [0x00000] in <filename unknown>:0
[07:48:17]  at System.Diagnostics.Process.Start_common (System.Diagnostics.ProcessStartInfo startInfo, System.Diagnostics.Process process) [0x00000] in <filename unknown>:0 
[07:48:17]  at System.Diagnostics.Process.Start (System.Diagnostics.ProcessStartInfo startInfo) [0x00000] in <filename unknown>:0 
[07:48:17]  at Protobuild.AutomatedBuildRuntimeV1.Execute (System.Object handle) [0x00000] in <filename unknown>:0 
[07:48:17]  at Protobuild.AutomatedBuildController.Execute (System.String path) [0x00000] in <filename unknown>:0 
```

hach-que:
Yeah Mac is failing because Protobuild doesn't have the executable bit set
(so Mono gives up even though the file exists)

tomspilman:
> Yeah Mac is failing because Protobuild doesn't have the executable bit set

Hum...  should I remote in and fix that on the build server?

hach-que:
Nah, I think it's something to do with the repository itself? Like on Unix
systems you can chmod to get the executable bit set so that you don't have
to prefix .NET executables with the mono command when running from another
.NET app. My workaround for this issue is to just always run chmod a+x
before attempting to run an executable on Unix systems (I just forgot to do
it here).

hach-que:
Huh, that's weird.  Your build server fails to execute Protobuild even with the executable bit set.  This definitely works on my build server, so my guess is that older versions of Mono don't do this at all on Mac (and the version of Mono on this Mac build server is way out of date).

hach-que:
It looks like the version of Xamarin Studio needs to be upgraded on the Mac build server (just in general, not specifically for this PR).  There's no `Xamarin.Mac.CSharp.targets` file, which is the newer API provided by Xamarin.

hach-que:
I'm having a bunch of issues installing nant on my Mac Mini; it won't run I believe because Xamarin has dropped the 4.0 profile (which NAnt relies on to run; it doesn't want to use 4.5).

Do you have any instructions on how to replicate your build server setup?

tomspilman:
> it won't run I believe because Xamarin has dropped the 4.0 profile (which NAnt relies on to run; it doesn't want to use 4.5).

@dellis1972 - Any ideas?

> Do you have any instructions on how to replicate your build server setup?

Not really...  it is all pretty much figured out as we went.  I know getting NANT working on the Mac was a PITA.  @dellis1972 might know more about that if I remember correctly.

hach-que:
> it won't run I believe because Xamarin has dropped the 4.0 profile (which NAnt relies on to run; it doesn't want to use 4.5).

also this is somewhat of a concern if you upgrade Xamarin on the Mac to support unified APIs, but it also could just be a configuration issue on my Mac build server (and there's some way of telling NAnt to use an older version of Mono without sticking it in the PATH and overriding the newer Xamarin version)

dellis1972:
Just for reference we did have to build Nant manually on the Mac as no packages were provided. Every time we upgrade mono we have to rebuild Nant, so we tend to only upgrade when we really need to . The next major upgrade will probably be for the tvOS build support (it currently requires an Alpha build of XS to work and we are not putting that on the server :) ).

hach-que:
Ugh yeah that sounds way more painful than I want to get into.

For now I've configured my build server to just download `automated.build` after cloning but before kicking off the build; this way I can track the upstream version of MonoGame and make sure it's working for all of the platform configurations I have (I'm testing all platforms that MonoGame supports, but I tend to update the SDKs on my build server regularly, so it'll give an indication of breakage against newer versions of Xamarin or Microsoft SDKs before the majority of users hit them).

Issue #4279, opened 2015-12-01
Strange method implementation for MathHelper.LerpPrecise


ala53:
In MonoGame.Framework/MathHelper.cs, line 196:
```
        public static object LerpPrecise(float value1, float value2, float amount)
        {
            return ((1 - amount) * value1) + (value2 * amount);
        }
```

Is there a reason this returns an object (which is both useless to API users and involves boxing/GC overhead) rather than a float?

KonajuGames:
It shouldn't. We missed that when reviewing the PR. We concentrated on what
it did compared to XNA rather than how it returned it.

ala53:
Oh, well that makes things easy. Should I do a PR?

ala53:
To be fixed in #4280

PR #4280, opened 2015-12-01
Fix LerpPrecise to return float instead of object

ala53:
In reference to #4279.


ala53:
@tomspilman Mind merging? It doesn't seem like the type of patch that has side effects needing lengthy consideration.

tomspilman:
Can we fix the whitespace issue in this PR first?

ala53:
Well, that commit broke somewhere. I'm just going to close this since #4294 fixes it as well.

Issue #4290, opened 2015-12-03
Crash in Pipeline tool on MacOS


dellis1972:
System.IO.IOException: kqueue() FileSystemWatcher has reached the maximum number of files to watch.
  at System.IO.KqueueMonitor.Start () <0x83e2640 + 0x001cb> in <filename unknown>:0 
  at System.IO.KeventWatcher.StartDispatching (System.IO.FileSystemWatcher fsw) <0x83e21b0 + 0x000ab> in <filename unknown>:0 
  at System.IO.FileSystemWatcher.Start () <0x83e2188 + 0x0001f> in <filename unknown>:0 
  at System.IO.FileSystemWatcher.set_EnableRaisingEvents (Boolean value) <0x83e2140 + 0x0002b> in <filename unknown>:0 
  at (wrapper remoting-invoke-with-check) System.IO.FileSystemWatcher:set_EnableRaisingEvents (bool)
  at MonoGame.Tools.Pipeline.PipelineController.OpenProject (System.String projectFilePath) <0x74895b0 + 0x003eb> in <filename unknown>:0 
  at MonoGame.Tools.Pipeline.MainWindow.OnShowEvent () <0x7489430 + 0x00105> in <filename unknown>:0 
  at MonoGame.Tools.Pipeline.Program.Main (System.String[] args) <0x74aed8 + 0x000b7> in <filename unknown>:0 
[ERROR] FATAL UNHANDLED EXCEPTION: System.IO.IOException: kqueue() FileSystemWatcher has reached the maximum number of files to watch.
  at System.IO.KqueueMonitor.Start () <0x83e2640 + 0x001cb> in <filename unknown>:0 
  at System.IO.KeventWatcher.StartDispatching (System.IO.FileSystemWatcher fsw) <0x83e21b0 + 0x000ab> in <filename unknown>:0 
  at System.IO.FileSystemWatcher.Start () <0x83e2188 + 0x0001f> in <filename unknown>:0 
  at System.IO.FileSystemWatcher.set_EnableRaisingEvents (Boolean value) <0x83e2140 + 0x0002b> in <filename unknown>:0 
  at (wrapper remoting-invoke-with-check) System.IO.FileSystemWatcher:set_EnableRaisingEvents (bool)
  at MonoGame.Tools.Pipeline.PipelineController.OpenProject (System.String projectFilePath) <0x74895b0 + 0x003eb> in <filename unknown>:0 
[...]

PR #4291, opened 2015-12-03
Works around #4290

dellis1972:
While this does not Fix the issue, it at least allows the Pipeline
tool to run...
We need to wait for an upstream mono fix.

Alternatives are

from the terminal

export MONO_MANAGED_WATCHER=false
open /Applications/Pipeline.app

this will allow the app to run up normally.


dellis1972:
It seems the Info.plist changes do work :D If you have an existing installation you need to let MacOS know that the Info.plist has changed using 

```
/System/Library/Frameworks/CoreServices.framework/Frameworks/LaunchServices.framework/Support/lsregister -v -f /Applications/Pipeline.app
```

This allows the Environment variables to work (it seems). I'll add that to the installer

dellis1972:
@KonajuGames this is all good . Fixes https://github.com/mono/MonoGame/issues/4308 as well.

KonajuGames:
Merged.

PR #4294, opened 2015-12-06
LerpPrecise for Color and Vector2/3/4

hede:
Adds LerpPrecise for Color, Vector2, Vector3, and Vector4
Includes a couple of minor documentation tweaks

Also has the fix for MathHelper.LerpPrecise return type, because it looks like #4280 was not merged

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -1719,6 +1719,25 @@ namespace Microsoft.Xna.Framework
                 (int)MathHelper.Lerp(value1.B, value2.B, amount),
                 (int)MathHelper.Lerp(value1.A, value2.A, amount) );
         }
+
+        /// <summary>
+        /// Performs linear interpolation of <see cref="Color"/> using <see cref="MathHelper.LerpPrecise"/> on MathHelper.
+        /// Less efficient but more precise compared to <see cref="Color.Lerp"/>.
+        /// See remarks section of <see cref="MathHelper.LerpPrecise"/> on MathHelper for more info.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


hede:
@tomspilman can you show me how to create the documentation locally please?

tomspilman:
@hede - You can see the command line to build docs here:

https://github.com/mono/MonoGame/blob/develop/default.build#L190

Basically it is this from the MonoGame checkout folder:
```
ThirdParty\Dependencies\SharpDoc\SharpDoc.exe -config Documentation\config.xml
```
It assumes all the target platforms are built already....  WindowsDX, WindowsGL, Windows 8 Store, Windows Phone 8, Linux, and Web.  You can comment out the extra platforms in the config.xml:

https://github.com/mono/MonoGame/blob/develop/Documentation/config.xml

... if that makes it easier to test locally.

Finally you can just let the build server do the testing for you.  You can browse the artifacts for your current build on the builder server like so:

http://teamcity.monogame.net/repository/download/MonoGame_DevelopWin/11957:id/Documentation.zip%21/index.html?guest=1

The `11957` is the build id that you can find when you click thru to the build results.

hede:
Thanks for the docs creation tips.

I think I'm mostly happy with the docs.
Couldn't get \<para\>....\</para\>, \<seealso\>, and full "ClassName.MethodName" (as opposed to just "MethodName" on \<see cref\> links to work :/ but apart from that, I think it's ready to go.

Maybe some of the docs for LerpPrecise methods could be in the remarks section (as opposed to the summary section) and be less in-your-face? Please let me know if you think that would be better.

KonajuGames:
I'm happy with this.  Merging.

Issue #4288, opened 2015-12-03
[Maintenance] Generate Visual Studio and Xamarin Studio templates from a common source


hach-que:
Right now MonoGame has the following project templates for various different IDEs:
- ProjectTemplates\VisualStudio2010\Android
- ProjectTemplates\VisualStudio2010\Linux
- ProjectTemplates\VisualStudio2010\OUYA
- ProjectTemplates\VisualStudio2010\Windows
- ProjectTemplates\VisualStudio2010\WindowsGL
- ProjectTemplates\VisualStudio2012\WindowsPhone
- ProjectTemplates\VisualStudio2012\WindowsStore
- ProjectTemplates\VisualStudio2012\WindowsStoreXaml
- ProjectTemplates\VisualStudio2013\WindowsPhone
- ProjectTemplates\VisualStudio2013\WindowsPhone8.1
- ProjectTemplates\VisualStudio2013\WindowsStore
- ProjectTemplates\VisualStudio2013\WindowsStoreXaml
- ProjectTemplates\VisualStudio2013\WindowsUniversal
- ProjectTemplates\VisualStudio2013\WindowsUAP
- IDE\MonoDevelop\ (and all the various template files under here)

We can probably make it easier to maintain all of these variants if we generate them - this would have the added bonus that if we want to quickly test the current version of MonoGame that we're working on as an actual app, we can just create a new project from those templates rather than having to actually install the templates into Visual Studio / Xamarin Studio first (especially since all this really is is just fancy text replacement).

cra0zy:
I had a similar idea about this with #3954 a while back.

tomspilman:
Yeah...  I think ideally we want what @cra0zy suggested in #3954.

The template generation systems in VS and XS both stink in different ways.  Neither is doing us any favors or solve any of the hard issues.

So IMO the idea of sidestepping them entirely and have it just launch our own project creation is fine.  The key still is still providing some basic hook so that you can launch our project creation from VS/XS in a way familiar to most users.

hach-que:
This task was far more focused on just unifying the VS / XS templates into some common source and generating the real templates with a shell script and some text replacements so that there isn't effectively two slightly different copies of the same thing in the repo. (i.e. not user facing at all)

[...]

PR #4282, opened 2015-12-02
[Mac] Upgrade Protobuild, Allow MacOS to target newer Xamarin APIs, etc.

hach-que:
This PR addresses multiple issues at once because they're all dependent on one another:

* **Upgrades Protobuild:** Throughout the course of this PR I had to make changes to Protobuild to support the newer APIs and fix various issues across platforms, so that is updated here.
* **Remove old XSLTs:** With the Windows Universal and tvOS platforms upstreamed into Protobuild, there's no need for MonoGame to keep it's own copies of the XSLT files any more.  This PR removes them in favour of the built-in ones.  This will assist with #4218.
* **Mac now supports the Xamarin.Mac Unified API:** It supports both the MonoMac API, XamMac API and newer Xamarin.Mac API (unified).  The older APIs are supported by #ifdef'ing namespaces with the `PLATFORM_MACOS_LEGACY` define.  This resolves #4275, but should also fix bugs like #4278.
* **WindowsUAP is now WindowsUniversal:** This brings the MonoGame platform name inline with the platform name being used in upstream.  See https://github.com/hach-que/Protobuild/pull/130 for the name change reasoning.

PR #4295, opened 2015-12-06
Add project creation tool

hach-que:
See #4288.  This adds a cross-platform project creation tool to MonoGame, intended to replace the Visual Studio and Xamarin Studio templates that are currently used.  Here is a video of a very early version working: https://youtu.be/uxnvCNX95ts

I still need to get these things done before it's mergeable (although it's in a submodule so it's just a case of updating the reference in future):
- [x] Draw a list of templates from the local folder or branding file
- [x] Instantiate the project from a template that is on-disk instead of online
- [x] Allow user to pick the project generation type (C# projects with Git submodule, C# projects with NuGet, Protobuild, etc.)
- [x] Pre-generate for all supported platforms if the template isn't using Protobuild
- [x] Apply the overlays that the user has chosen
- [x] Fix up any cross-platform support that is missing on Mac
- [x] Fix up any cross-platform support that is missing on Linux
- [x] Integrate it into the build so it can be shipped
- [x] Update the installer so that it wires it up to Visual Studio / Xamarin Studio as appropriate


hach-que:
Slowly making progress here - the tool now supports all platforms, although the integration with Xamarin Studio / MonoDevelop is a bit clunky (it launches new instances instead of re-using old ones).  We'll need @cra0zy's work on the MD add-ins to make this experience better, but I don't think it's a day 1 requirement for merging (especially since those features are less important for MonoGame).

My next steps are getting this into the build so that NSIS installs a start menu shortcut and places it in the Mac applications folder so that people can use it to create new projects.

hach-que:
This build won't pass on the build server yet until #4282 is merged, for various reasons.

tomspilman:
> My next steps are getting this into the build so that NSIS installs a start menu shortcut 

Is this what we actually want to do?

@KonajuGames @dellis1972 ?

dellis1972:
While this is nice as a standalone tool not sure it should replace the templates we have. It kind of goes against the standard IDE experience. It might be worth while looking at the new Xamarin Studio tempting system and Visual Studio IWizard

cra0zy:
> While this is nice as a standalone tool not sure it should replace the templates we have. It kind of goes against the standard IDE experience. It might be worth while looking at the new Xamarin Studio tempting system and Visual Studio IWizard

That's my goal for setting up the MD addin, if you look at https://youtu.be/IvwRahT1HTY you'll notice that it's just 4 widgets(technically 3 since the first one won't be used in the addin) and a Window Container that has next, prev, cancel options.

Now @hach-que can reuse the code for VS as well(since I used Xwt and you can get WPF stuff) once I submit the code, tho I probably won't be able to do much until saturday. I have an exam this weak.

dellis1972:
@cra0zy that looks really nice.

hach-que:
@dellis1972 The intention is to have this hooked up to the IDE project creation wizards by modifying the add-ins, but that side of things isn't ready yet. That's why this PR doesn't yet remove the old templates. While I could add that into this PR, I figured it would be better to keep it smaller now to a point when it can be merged, rather than ending up with a massive PR like the Mac upgrade. 

@tomspilman If you don't want it in the start menu for now (or you only ever want it accessible from the IDE), then I can skip over that and just work on getting it hooked into the IDE instead, but since that's a larger change it's probably better to seperate it from this PR?

hach-que:
fyi I'm working on a Visual Studio extension that will integrate this into the New Project Wizard

dellis1972:
@hach-que @cra0zy you guys both seem to be working on similar things? perhaps you should collaborate and come up with something together. I really like the UI @cra0zy put together, and love the fact that the user gets to choose if they want to use Nuget/Git etc. makes for a nice end user experience.

Ideally we want the experience in XS and VS to be exactly the same

hach-que:
I've actually found a way that automatically synchronises all of the templates defined in the Protobuild Manager into the Visual Studio New Project Wizard... it all hooks up and the user can type the new project name and select the location of their project, and then it kicks off the Protobuild Manager tool to pick the last few options (how they want to reference MonoGame etc.)

@dellis1972 This also does NuGet / etc.  you pick it when creating the project how you want to make the solution: https://www.youtube.com/watch?v=-jEqdZHXeII.  We were planning on combining our work, but I'm not sure where @cra0zy is at with their work.

hach-que:
@dellis1972 @tomspilman @KonajuGames Okay, so I have this integrated into Visual Studio now: https://youtu.be/mX6qnRUeWX0  Is this what you're looking for?

I still need to clean up the code a bit before adding it to this PR.

dellis1972:
A couple of questions

1) do we really need to download stuff from the internet? Can developers run this offline?
2) is it possible to have Linux, Android, iOS and WindowsPhone (or others) in the same solution with a shared/PCL project for the "game" code.. From my experience this is how most developers work with MonoGame now. 

I'm not sure we should force developers to have to open the tool again to switch projects or load a new solution to run up the android port. I know this is how marmalade does it but their framework requires that.

hach-que:
Now it's even fancier:

image

This entry is synchronised from the templates the tool is aware of whenever it runs.  When you press next here it triggers the project creation tool to complete the last few steps.

@dellis1972 

> 1) do we really need to download stuff from the internet? Can developers run this offline?

This is temporary.

> 2) is it possible to have Linux, Android, iOS and WindowsPhone (or others) in the same solution with a shared/PCL project for the "game" code.. From my experience this is how most developers work with MonoGame now.

I haven't written the templates for the PCL / Shared Code project drop-down options yet, but the intention is those solution will have all of the platform projects in one solution (I just have to write the XSLTs for it).

dellis1972:
ok cool. I know its a work in progress but have to ask these questions up front :) 

I'm wondering if we should offer the most common options by default i.e Shared/PCL with Nuget and all in one SLN . then offer the Protobuild/source/git stuff as an advanced option. 

I mention this because most MG developers I imagine don't use protobuild  but use the Nugets and never go near the actual MG source code. We should allow them to create a solution with as few clicks as possible. I've learnt allot about user experience in the last few years with Xamarin and that initial user experience is key to getting a good impression :)

hach-que:
By odd coincidence, I have some statistics on that :)

This is how the landscape falls as it is today:

image

However it's important to note that almost 50% of respondants said they have never heard of Protobuild in terms of managing cross-platform projects:

image

Because of that latter statistic, I'm not sure whether people actually like using Shared Code / PCL projects, or whether they don't know there's (in my biased opinion), better alternatives.

Anyway, it's totally up to you because the template JSON file will let you specify the default options (it doesn't yet though).

hach-que:
By odd coincidence, I have some statistics on that :)

This is how the landscape falls as it is today:

image

So around 40% of people who use MonoGame, use the source code, while around 27% use NuGet and another 21% use the installers (the latter two options being binary files).  So overall about a 60/40 (binary/source) split.

image

A little under 10% of the community is using Protobuild, with over 50% using either Shared Code Projects or PCLs.  *However* it's important to note that almost 50% of respondants said they have never heard of Protobuild in terms of managing cross-platform projects:

image

Because of that latter statistic, I'm not sure whether people actually like using Shared Code / PCL projects, or whether they don't know there's (in my biased opinion), better alternatives.

Anyway, it's totally up to you because the template JSON file will let you specify the default options (it doesn't yet though).

hach-que:
@tomspilman @dellis1972 @KonajuGames It looks like this build is stalled trying to deploy the extension to VS: http://teamcity.monogame.net/viewLog.html?buildTypeId=MonoGame_DevelopWin10&buildId=12394  I disabled that in the next commit but I don't have permissions to abort the build (and I don't know if it will timeout eventually)?

hach-que:
@mgbot abort

tomspilman:
@hach-que - I killed that build for you.  You cannot control It from chat... I have to do it from the TeamCity server.

hach-que:
Yeah sorry about that, I wasn't expecting the default build behavior to be "interact with a GUI tool" :/

hach-que:
@tomspilman @KonajuGames 

This is pretty much ready for review, with a few important notes:

* I haven't made the Mac / Linux installers install and run the Protobuild Manager.  While the project creation tool *does* work on these platforms, the installer just won't hook it up yet.  I won't have access to my Mac Mini until next week, and I think this PR has enough value for Windows users that it'll give us the ability to have people test out the new templating system and give feedback.  I expect wiring this up for Linux to be a little more tricky, because Linux distros tend to vary quite wildly, but I do intend on sending another PR in future to have the Linux installer set it up.
* This PR does not remove the old templates, nor does it remove the templates from the installer.  While I've tested the new template creation process, I don't want to remove them just yet.  I'll send another PR that removes the templates when I have the Mac and Linux installers setting up the new template process as well.

Let me know if this is okay, or whether you want these things addressed in this PR as well.

ProjectCreationTool/Debug/Branding.xml
```diff
@@ -0,0 +1,77 @@
...
+  -->
+  <VisualStudioAddinPackage>local-lzma://../Protobuild.IDE.VisualStudio/Windows.tar.lzma</VisualStudioAddinPackage>
+
+  <!--
+  The location of the MonoDevelop / Xamarin Studio IDE package to install.
+  You can change this URL to local-lzma://<some relative or absolute file path>
+  if you want to ship a local copy of the package and avoid downloading
+  it from the internet.
+  -->
+  <MonoDevelopAddinPackage>http://protobuild.org/hach-que/Protobuild.IDE.MonoDevelop@monodevelop-5</MonoDevelopAddinPackage>
```
hach-que: This will be replaced with a local copy when the Mac / Linux installers are hooked up in a future PR.  At the moment this will only kick in if someone builds MonoGame from source and manually uses the project creator on those platforms.

dellis1972:
This is a large commit, I think we should leave this till after the 3.5 release so we can do some testing on it and get the mac/linux side done for the next release after that.

I need to check that ALL the templates are covered including the Pipeline projects etc :)

cra0zy:
Few bad things I see:
 - After installation I tried creating 2d platformer template, it threw an error: "Error this template attempted to load component assembly 'Protobuild.IDE.VisualStudio,Wizard, Version=1.0.0.0, Culture=neutral..."
 - You are bundling samples with the SDK, they should be separate
 - You are not offering Blank template, which should be the default..
 - The colors used by the tool are the default colors, It would be much nicer if the tool used "MonoGame orange" + ((gray + white) or (white + black))... I know you'll say "Brandung.xml"
 - No Shared project support
 - Always starts with all the platforms selected
 - Platforms aren't organized in any way, at least if it was "Alphabetical" that would make sense
   - it would be nice if they were split into Desktop, Mobile, Console
 - You can unselect all platforms and still start generating the project... 
 - Generating templates takes a bit too long IMO
 - You are keeping the tool active after generating standard project in order to switch between platforms instead of just adding all the projects to one solution

hach-que:
> After installation I tried creating 2d platformer template, it threw an error: "Error this template attempted to load component assembly 'Protobuild.IDE.VisualStudio,Wizard, Version=1.0.0.0, Culture=neutral..."

What does it say in the log view (click in the bottom right)? It should tell you what Visual Studio versions it detected and what it tried to install to. The GAC install will only work in admin mode, but this is what --silent-setup does in the MG installer.

> You are bundling samples with the SDK, they should be separate. You are not offering Blank template, which should be the default.

This is just the template I had on hand for all platforms. We can put whatever templates we want in without duplicating work - that's one of the advantages of this new system. 

> The colors used by the tool are the default colors, It would be much nicer if the tool used "MonoGame orange" + ((gray + white) or (white + black))... I know you'll say "Brandung.xml" 

That's the plan. 
 
> No Shared project support

I know, I mentioned this earlier.

Will respond to the rest tomorrow.

hach-que:
> This is a large commit, I think we should leave this till after the 3.5 release so we can do some testing on it and get the mac/linux side done for the next release after that.

Yes we should.  I thought 3.5 was already out - is there an expected date for that?

> Always starts with all the platforms selected

What else do you suggest?

> Platforms aren't organized in any way, at least if it was "Alphabetical" that would make sense

This seems pretty minor - you only need to select these once when you're first making the project

> Generating templates takes a bit too long IMO

Nothing I can do about that, the generation depends on how many platforms you want to generate for

> You are keeping the tool active after generating standard project in order to switch between platforms instead of just adding all the projects to one solution

Yeah, Protobuild is one solution per platform (it doesn't make sense to have multiple platforms in one solution because the available configuration dropdown for ARM / x86 / x64 / AnyCPU / iPhone / iPhoneSimulator will change depending on what platform is being used).

tomspilman:
> > Always starts with all the platforms selected
> What else do you suggest?

Start with the host platform selected only.

> it doesn't make sense to have multiple platforms in one solution because the available configuration dropdown for ARM / x86 / x64 / AnyCPU / iPhone / iPhoneSimulator will change depending on what platform is being used

I agree.  I've never found VS well suited to managing multiple platforms in one solution.  You end up with tons of configurations and you can never tell which config is meant for which project.

That said...  some people totally want all the platforms in one solution.  Eventually someone will complain about that...  but since having multiplatform project generation is new we can live with it for now.

cra0zy:
I just remembered one important thing, remove Web as platform, if you leave it people will start submitting issues about it even tho it's not implemented yet.

> This seems pretty minor - you only need to select these once when you're first making the project

Whenever you present a user with a list, don't just randomly add stuff, sort it in some way. GUI design 101

> What else do you suggest?

Select only DesktopGL by default. This also helps with people noticing that there is now a DesktopGL platform.

> This is just the template I had on hand for all platforms. We can put whatever templates we want in without duplicating work - that's one of the advantages of this new system.

I know, what I was trying to say is this, make just a Blank template be installed with the SDK, provide a separate file for other templates, we should also use a separate repository for template, ie. https://github.com/MonoGame/MonoGame.Samples would maybe be a good place.

> What does it say in the log view (click in the bottom right)? It should tell you what Visual Studio versions it detected and what it tried to install to. The GAC install will only work in admin mode, but this is what --silent-setup does in the MG installer.

```
[info] Writing out project templates for Visual Studio 2010...
[info] Writing out project templates for Visual Studio 2012...
[info] Writing out project templates for Visual Studio 2013...
[info] Writing out project templates for Visual Studio 2015...
[info] Project templates synchronised.
[info] Visual Studio 10.0 is not installed (checked for existance of C:\Program Files (x86)\Microsoft Visual Studio 10.0).
[info] Visual Studio 11.0 is not installed (checked for existance of C:\Program Files (x86)\Microsoft Visual Studio 11.0).
[info] Will perform GAC-based installation for Visual Studio 12.0.
[info] Will perform VSIX-based installation for Visual Studio 14.0.
[info] Downloading latest IDE add-in for Visual Studio...
[info] Obtaining a copy of Protobuild...
[info] Checking to see if the cached version of Protobuild is up-to-date...
[info] Using version of Protobuild that is in the cache.
[info] Installing package from local-lzma://./Protobuild.IDE.VisualStudio.tar.lzma...
[error] Non-zero exit code from Protobuild.exe --install
[error] Unable to download latest IDE add-in for Visual Studio!
```

hach-que:
> Select only DesktopGL by default. This also helps with people noticing that there is now a DesktopGL platform.

I'm going to do what @tomspilman asked for, which is to only select the host platform by default.

> I know, what I was trying to say is this, make just a Blank template be installed with the SDK, provide a separate file for other templates, we should also use a separate repository for template, ie. https://github.com/MonoGame/MonoGame.Samples would maybe be a good place.

I'm pretty sure @tomspilman will want to avoid having templates be downloaded from the internet when used, so we need to figure out a way to download the additional templates on build and bundle them into the installer.  I don't know the best way to do this at the moment.

```
[info] Writing out project templates for Visual Studio 2010...
[info] Writing out project templates for Visual Studio 2012...
[info] Writing out project templates for Visual Studio 2013...
[info] Writing out project templates for Visual Studio 2015...
[info] Project templates synchronised.
[info] Visual Studio 10.0 is not installed (checked for existance of C:\Program Files (x86)\Microsoft Visual Studio 10.0).
[info] Visual Studio 11.0 is not installed (checked for existance of C:\Program Files (x86)\Microsoft Visual Studio 11.0).
[info] Will perform GAC-based installation for Visual Studio 12.0.
[info] Will perform VSIX-based installation for Visual Studio 14.0.
[info] Downloading latest IDE add-in for Visual Studio...
[info] Obtaining a copy of Protobuild...
[info] Checking to see if the cached version of Protobuild is up-to-date...
[info] Using version of Protobuild that is in the cache.
[info] Installing package from local-lzma://./Protobuild.IDE.VisualStudio.tar.lzma...
[error] Non-zero exit code from Protobuild.exe --install
[error] Unable to download latest IDE add-in for Visual Studio!
```

Can you run `Protobuild.exe --install local-lzma://./Protobuild.IDE.VisualStudio.tar.lzma` from the command-line (from the folder where MonoGame is installed to?)

cra0zy:
> I'm pretty sure @tomspilman will want to avoid having templates be downloaded from the internet when used, so we need to figure out a way to download the additional templates on build and bundle them into the installer. I don't know the best way to do this at the moment.

I never said anything about templates being downloaded from the internet when used, all I said is that they should all be in a separate file from the main SDK... and also that they should not live in the main MonoGame reposetory...

cra0zy:
> Can you run Protobuild.exe --install local-lzma://./Protobuild.IDE.VisualStudio.tar.lzma from the command-line (from the folder where MonoGame is installed to?)

*Protobuild.Manager.exe
No output/1 empty line.
Still same error.

hach-que:
No, you need to run `Protobuild.exe`.  This is what performs the actual installation.  You should be able to run `%appdata%\Protobuild.exe --install local-lzma://./Protobuild.IDE.VisualStudio.tar.lzma` if there isn't a Protobuild in the MonoGame installation directory.

cra0zy:
> No, you need to run Protobuild.exe. This is what performs the actual installation. You should be able to run %appdata%\Protobuild.exe --install local-lzma://./Protobuild.IDE.VisualStudio.tar.lzma if there isn't a Protobuild in the MonoGame installation directory.

```
Installing local-lzma://./Protobuild.IDE.VisualStudio.tar.lzma...
System.InvalidOperationException: No target folder was provided for package reso
lution, and the resulting package is not a global tool.
   at Protobuild.PackageManager.Resolve(ModuleInfo module, PackageRef reference,
 String platform, String templateName, Nullable`1 source, Boolean forceUpgrade)
   at Protobuild.InstallPackageCommand.Execute(Execution execution)
   at Protobuild.MainClass.Main(String[] args)
```

tomspilman:
> This also helps with people noticing that there is now a DesktopGL platform.

This makes me wonder if we could somehow provide description strings for each platform?  What "DesktopGL" or "Web" exactly mean isn't obvious... some extra info would be nice.

> they should not live in the main MonoGame repository

What is your reasoning for that @cra0zy ?

It seems to me that every time we put things in separate repositories it gets forgotten and stale.  It also makes it harder to do updates that need to be synced to changes in the framework code.

cra0zy:
> What is your reasoning for that @cra0zy ?
> 
> It seems to me that every time we put things in separate repositories it gets forgotten and stale. It also makes it harder to do updates that need to be synced to changes in the framework code.
>  Show all checks

Repository size.

hach-que:
> Repository size.

I think this is largely avoided as long as we only add PNG files, and avoid adding XNB or audio files to the repository.  PNG files seem to do quite well.

The other alternative is to have the build server build those samples with each PR / commit build of MonoGame, so we'll know immediately if they're out of date or stale, because the build will fail for any PR that breaks them.

tomspilman:
> Repository size.

That is probably the only concern.... I can't think of another.

Generally I don't think these templates should be very big to begin with... at least not much bigger than our existing template projects which should go away.

@KonajuGames  @dellis1972 ?

hach-que:
@cra0zy The newer version should fix that `local-lzma` issue.  Once it finishes building, can you download the installer, delete the `%appdata%\.monogame-manager` directroy and try again.

KonajuGames:
> I think this is largely avoided as long as we only add PNG files, and avoid adding XNB or audio files to the repository. PNG files seem to do quite well.

I think we've got enough in separate repositories as it is.  Git's submodules aren't the nicest things to deal with.  And as Tom said, any changes that cross that repository boundary become a pain.

As much as the GitHub PR view lets me see of this (not much of the 647 files in this PR), there are pre-built XNB files for several platforms.  Ideally these would be the source audio files and have the content build pipeline setup in the project to build the assets.  That is something that can be added later.  This PR is large enough as it is.

This branch needs latest from develop merged into it.

tomspilman:
@hach-que - Can the template definitions limit which platforms they can target?  Thinking about this specifically for #4409.

hach-que:
Yeah we can do that.

cra0zy:
Few things:
 - merge conflicts!!!
 - this needs to be merged before the SDL branch so there is no pointless work done with templates
 - this PR should probably update Windows/Linux/Mac all in one swoop, and remove the old template system
 - please use Gtk 3 on Linux, Gtk 2 file dialogs don't look native on Gtk 3 based desktop environments, and also don't take the dark theme correctly

cra0zy:
For anyone wondering on the difference of Gtk 2 and 3 file dialogs:

Gtk 3:


Gtk 2:

hach-que:
That still looks pretty usable to me. I'm much more concerned with compatibility with Linux distros than I am appearance, and Gtk# 2 is more widely available. Until I get a chance to review the PR that you sent through and test it on a bunch of distros to make sure everything works out of the box, we'll be sticking with Gtk# 2, because at least it works.

I'm holding off removing the old templates in this PR. This PR is already big enough, and having a bunch of file deletions just makes it harder to review.

cra0zy:
> and Gtk# 2 is more widely available.

On Windows and Mac, yes, on Linux, no, and you don't use Gtk for Windows and Mac if I remember correctly.

Also Pipeline Tool on Linux requires that Gtk# 3 installed so....

Gtk 3 was released in 2011 and since then Gtk 2 is in maintenance mode. Gtk 2 projects are either moving to Gtk 3 or QT, except MonoDevelop.

> I'm holding off removing the old templates in this PR. This PR is already big enough, and having a bunch of file deletions just makes it harder to review.

Then just do it in 2 commits, it will be just as easy to review.

hach-que:
Specifically the issue is the WebKit bindings, where the Gtk# 3 version of
them is not as widely available.

cra0zy:
> Specifically the issue is the WebKit bindings, where the Gtk# 3 version of
> them is not as widely available.

I provided you with Gtk# 3 binding(hand written) in the PR I submitted... still a bad excuse... and just on a side note, I know that I don't have Gtk 2 webkit installed (not just the bindings, the actual library, I use GNOME).

hach-que:
Yeah and I haven't had time to review it. It needs testing on OpenSUSE and
other distros, and currently I'm travelling around the U.S. I simply don't
have the capacity to perform that testing until I return to Australia.

tomspilman:
No worries... this can hold for a little longer.  I still need to pull it all down and try it out as is and figure out what the plans are for templates.

DDReaper:
Any updates / Progress on this project @hach-que @tomspilman ?  Is this an avenue still worth exploring? 
Would certainly clean up the mess of trying to maintain separate templates for different editors.

hach-que:
There's nothing waiting on my end. I've been successfully using the Protobuild Manager for my own stuff for like almost a year now.

I think (can't tell on mobile) this PR needs rebasing, but I don't want to do that work until I have a commitment that the PR is going to get tested and merged by upstream, because otherwise it'll just get out of date again.

tomspilman:
> I don't want to do that work until I have a commitment that the PR is going to get tested and merged by upstream

I've been pretty swamped with work for the last 6 months, but after GDC we will look to tackle this issue.

tomspilman:
@hach-que - So i installed Protobuild Manager directly from the Windows installer linked from here:

https://github.com/Protobuild/Protobuild.Manager

My experience so far with the tool and concerns i have about our usage of it.

The tool seems to just be named Protobuild.exe ... am i correct in assuming this means it is the same as the current Protobuild.exe we have in the root of the MG project?  Would this mean if we were to adopt it then double clicking the Protobuild.exe in the root of MG would launch this tool?  This is not something we would want to happen.

When i run Protobuild.exe it seems to create a console window then launch its GUI window.  Can that be fixed to not generate a console window first?

The tool by default seems to pull templates from some unknown online sources.  It seems that can be disabled.  We would want the templates to come from local sources in most cases with maybe an option to gather other templates from an additional online source we can define too.

My first project build took alot of time... maybe 5 minutes.  I assume it is from it downloading tons of stuff.  My second project took less time... maybe 2 minutes.  I assume we can eliminate most of that by not pulling stuff from online sources.

I seem to remember the tool is caching some stuff to its own hidden location and/or locations.  Where would i find that on my system?

Can we limit which platforms a template works with?  For instance if you try to create a `C# console project" on Windows Universal platform you'll find that doesn't work at all.  I would expect we would have some similar situations...  like a Oculus VR template not working on Android.

I'm experimenting with the tool right now... i'm sure i'll have more questions later.

tomspilman:
@hach-que - Found an issue...  if you delete the generated project from disk it seems like it confuses the tool.  It still shows the project in the list... trying to click it tries to do stuff and fails, then when you go back adds a "No Solution Found!" message to the list.

hach-que:
> @hach-que - So i installed Protobuild Manager directly from the Windows installer linked from here:
> 
> https://github.com/Protobuild/Protobuild.Manager
> 
> My experience so far with the tool and concerns i have about our usage of it.
> 
> The tool seems to just be named Protobuild.exe ... am i correct in assuming this means it is the same as the current Protobuild.exe we have in the root of the MG project? Would this mean if we were to adopt it then double clicking the Protobuild.exe in the root of MG would launch this tool? This is not something we would want to happen.

I'm a little confused about this one.  The default behaviour of Protobuild (the command line tool), in an _empty_ folder is to install and run the Protobuild Manager.  The behaviour we had like 12+ months ago where it would just create an empty module and exit was confused to a lot of people, so I changed it.

It behaves exactly as before when it's run inside a module (like the MG repo).

> When i run Protobuild.exe it seems to create a console window then launch its GUI window. Can that be fixed to not generate a console window first?

Unfortunately because of the way Windows differentiates between console and GUI apps, no.  But if you're installing the Protobuild Manager from the Windows installer, it should appear in your start menu.  Launching it from there does not create a console window.

> The tool by default seems to pull templates from some unknown online sources. It seems that can be disabled. We would want the templates to come from local sources in most cases with maybe an option to gather other templates from an additional online source we can define too.

This is part of the branding system - the one you downloaded from the Protobuild website is unbranded, which means it uses the defaults that I've set up (it uses templates from online).

The one that's in this PR is a branded version of the tool; it installs itself separately (i.e. it works side-by-side with an unbranded tool if a user has both installed).  Most importantly however, it's configured to use templates from a local source (i.e. templates that are installed onto the local system as part of the MonoGame installer).  There's no online templates configured in this PR.

> My first project build took alot of time... maybe 5 minutes. I assume it is from it downloading tons of stuff. My second project took less time... maybe 2 minutes. I assume we can eliminate most of that by not pulling stuff from online sources.

Yeah, this is because the MonoGame template I ship in the unbranded version uses https://github.com/RedpointGames/MonoGame.Platformer2D, not the templates in this PR.  The template in that repo uses MonoGame as a downloadable dependency, which it does the first time the project is created.  The templates in this PR use the version of MonoGame that's installed by the installer (it might give the user a choice of where they want to source MonoGame from?)

One of the reasons that only repository takes so long is that it `git clone`'s the `Dependencies`/`ThirdParty` and `Kickstart` Git modules, even for a binary package, just because of how the automatic packager is works by default.  If they were excluded from the binary package, I think things should download and install much faster (it's a minor bug, and needs some sort of `DoNotReferenceInPackage` flag in MonoGame's packages list).

> I seem to remember the tool is caching some stuff to its own hidden location and/or locations. Where would i find that on my system?

`%appdata%\.protobuild-manager`, `%appdata%\.protobuild-tools` and `%appdata%\.protobuild-cache`.

> Can we limit which platforms a template works with? For instance if you try to create a `C# console project" on Windows Universal platform you'll find that doesn't work at all. I would expect we would have some similar situations... like a Oculus VR template not working on Android.

Yes, I believe the local template system that's used by this PR already supports this (but if it doesn't we can change that).

I'm experimenting with the tool right now... i'm sure i'll have more questions later.

hach-que:
> Found an issue... if you delete the generated project from disk it seems like it confuses the tool. It still shows the project in the list... trying to click it tries to do stuff and fails, then when you go back adds a "No Solution Found!" message to the list.

Oh, and it doesn't remove it from the list after that message appears?  That sounds like a bug; thanks for reporting the issue.

DDReaper:
Nice to see this back under work. Is this actually going to be a thing now?

cra0zy:
*caugh* https://github.com/dotnet/templating *caugh*

tomspilman:
We'll be moving away from Protobuild moving forward.  So closing this.  The idea is still a good one, but will likely be based on new dotnet stuff.

Issue #4296, opened 2015-12-06
TextureImporter fails while importing .bmp files


Tzenchor:
TextureImporter is raising a NullReferenceException on line 236, when trying to import bitmaps with less than 8 bits per pixel.
This is happening in the develop branch. In version 3.4 it worked perfectly.

The reason is that the code is hardcoded to transform files with 8, 16, 24, or 32 bits per pixel, but does nothing with smaller sizes.

I have a fix for this ready that seems to work, at least for monochrome and 4 bits .bmp files, but I want to have a look to the rest of the code and possibly simplify it

tomspilman:
> but does nothing with smaller sizes.

Just something that wasn't accounted for.

@KonajuGames ?

KonajuGames:
Correct. Didn't account for anything less than 8-bits per pixel in BMP form.

If you have a fix, we'll be happy to have a look at it if you submit it as
a PR.

Tzenchor:
I have found another problem while refactoring.
I have created a pair of bitmaps with 16 bits per pixel to test the other cases, and created the corresponding tests.
Well, the tests are failing because the test expects to receive data with SurfaceFormat Color, but it is getting bgra5551 and bgra565 respectively. And i'll get similar results with images having the RGBF and RGBAF format (floats instead of integers).
Should I preserve this results and modify the tests, or convert all of them to 32 bits like the other bitmaps?

KonajuGames:
I don't have the tests available to me right now, but they should do what
is needed to test the functionality without losing accuracy or precision.

Tzenchor:
Not sure if anyone is aware that i did the pull request a few days ago (i'm new to git and github) as #4304

PR #4304, opened 2015-12-08
Fix for #4296

Tzenchor:
.gitignore:
 +Added rule for *.csproj.csdat files created by Visual Studio 2015, not needed with the Visual Studio extension, but useful for Git Desktop
 +Added rule for "rebrace.xml" used by the rebracer extension for Visual Studio (this extension allows the user to have one style rules per solution instead of having to change the options manually when changing projects).
TextureImporter.cs:
 *Bitmap files are allways expanded to 32bits and converted to PixelContent<Color>
 *Files with RGBF format are expanded to RGBAF and converted to PixelContent<Vector4>
 +Added check for files with correct headers but bad content
TextureImporterTests.cs:
 +Added 2 tests for bitmaps with less than 8 bits per pixel (one with 4 and another with 1)
 +Added 2 tests for bitmaps with format R5G5B5 and R5G6B5, that were a special case with the old code
 +Added test for reading files with good headers, but bad content.
 *Modified ImportStandard to receive the expected SurfaceFormat for that file
New Files in Test project:
 +Assets\Textures\LogoOnly_64px-4bits.bmp
 +Assets\Textures\LogoOnly_64px-monochrome.bmp
 +Assets\Textures\Logo555.bmp
 +Assets\Textures\Logo565.bmp
 +Assets\Textures\rgbf.tif
Protobuild definitions updated


MonoGame.Framework.Content.Pipeline/TextureImporter.cs
```diff
@@ -53,7 +53,7 @@ public class TextureImporter : ContentImporter<TextureContent>
         /// <summary>
         /// Initializes a new instance of TextureImporter.
         /// </summary>
-        public TextureImporter()
+        public TextureImporter( )
```
KonajuGames: Reformatting, especially that which is just a whitespace change, is not wanted in a PR.  It adds noise to a PR and hides the real changes.

MonoGame.Framework.Content.Pipeline/TextureImporter.cs
```diff
@@ -63,168 +63,46 @@ public TextureImporter()
...
             var redMask = FreeImage.GetRedMask(fBitmap);
             var greenMask = FreeImage.GetGreenMask(fBitmap);
             var blueMask = FreeImage.GetBlueMask(fBitmap);
 
-            // Get the bytes from the FreeImage bitmap
-            var bytes = new byte[width * height * (bpp / 8)];
-            FreeImage.ConvertToRawBits(bytes, fBitmap, pitch, bpp, redMask, greenMask, blueMask, true);
+            // Create the byte array for the data,
+            // In case of modifying the code to generate pixel data with less than 8 bpp do not remove the casts or the size or the size will be wrong
+            var bytes = new byte[(int) (width * height * ((float) bpp / 8))];
```
KonajuGames: Rather than do the operation as float and convert back to int (and run the risk of floating point inaccuracy), keep it as integer all the way through, and it's easier to read.
`var bytes = new byte[(width * height * bpp) / 8];`

MonoGame.Framework.Content.Pipeline/TextureImporter.cs
```diff
@@ -237,5 +115,58 @@ public override TextureContent Import (string filename, ContentImporterContext c
...
+                    break;
+            }
+
+            return fBitmap;
+        }
+        /// <summary>
+        /// Switches the red and blue channels
+        /// </summary>
+        /// <param name="fBitmap">image</param>
+        private static void switchRedAndBlueChannels(FIBITMAP fBitmap)
```
KonajuGames: Method names always start with an uppercase character.

Test/ContentPipeline/TextureImporterTests.cs
```diff
@@ -29,51 +29,85 @@ void ImportStandard(string filename)
...
-        public void ImportBmp()
+        public void ImportBmpMonochrome( )
         {
-            ImportStandard("Assets/Textures/LogoOnly_64px.bmp");
+            ImportStandard("Assets/Textures/LogoOnly_64px-monochrome.bmp", SurfaceFormat.Color);
         }
 
         [Test]
-        public void ImportGif()
+        public void ImportGif( )
```
KonajuGames: More unnecessary reformatting.  Also check the [Coding Guidelines](https://github.com/mono/MonoGame/wiki/Coding-Guidelines) for rules regarding spaces.

MonoGame.Framework.Content.Pipeline/TextureImporter.cs
Tzenchor: A good idea, but can lead to problems if in a future someone wants to compress the color data by reducing the bits per pixel.
If bpp is less than 8 your option returns less bytes than needed when the number of pixels does not fill a complete byte.

Tzenchor:
I have updated my branch and corrected the offending function name.
I did not change the other style mistakes to avoid further cluttering, but changed my configuration to avoid future problems.
Also i did not modify the byte array generation because of possible future problems

Tzenchor:
I have found a problem related to pull request #4136, [DdsLoader.cs](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/DdsLoader.cs) can't process a simple texture within a dds file, it will raise a NullReferenceException at line 307
Should i add a fast fix for that here, or do it in another pull?

KonajuGames:
Another pull request.

As for future formats not filling a byte, I would not be concerned about
that. Pixel formats will always fill a byte.

Tzenchor:
I have found that my code has some flaws than yours
```c#
var bytes = new byte[(int) (width * height * ((float) bpp / 8))];
```
Does exactly the same than 
```c#
var bytes = new byte[(width * height * bpp) / 8];
```
I have been investigating and found the definitive solution in
[stackoverflow](http://stackoverflow.com/a/503201)
```c#
var bytes = new byte[(width * height * bpp -1) / 8 + 1];
```
I'll make the new commits once visual studio finishes its update

KonajuGames:
I'm not convinced about that one. If we stay in the integer domain, there
is no rounding or inaccuracy error to worry about.

Tzenchor:
I now it sounds weird but works, and remains in the integer domain (in reality they are cast to long because of bpp being uint), did some tests and works in all cases even where the other functions failed. Just try all of them with this values:
Width = 5
Height = 17
bpp = 4
The first two will return 42, that is 1 byte short, while the last returns correctly 43.

KonajuGames:
Ok, I'm good with merging this now.  Thanks.

PR #4323, opened 2015-12-16
Test Fix for UWP

tomspilman:
For some reason the UWP build is broken on the develop branch... trying to figure out why this is using this PR for testing.


tomspilman:
Ok... so the source of the issue is that we never updated the Win10 build box to have the VS2015 update that enabled renaming of the `project.json` to something project specific.  This meant that https://github.com/mono/MonoGame/pull/4163 never really worked.... more on that later.

Secondarily it turns out that the `project.json` doesn't get executed when building a project via MSBuild:

https://social.msdn.microsoft.com/Forums/vstudio/en-US/3355fa3b-c902-43a1-a3ff-9500e2a259d5/msbuild-doesnt-produce-projectjsonlock-file-making-the-build-fail?forum=msbuild

I suspect at some point when we first setup UWP we manually build the project from the VS2015 GUI which caused the `project.json` to execute.  Since then we've depended on the results for that one run which has kept the automated build working fine.

I need to test that this issue still occurs under the latest VS2015 changes that allow `project.json` to be renamed.

So why was `project.json` still in the build agent months after it has been removed from the repo?  It seems that when TeamCity does git checkouts on the server and transfers differences to the build agent....  it fails to remove any files deleted/removed by a commit.  I'm currently investigating what the cleaning rules are on build agents that depend on server side git checkouts.

DDReaper:
Might explain why my Win10 platform rename PR didn't build.
However Tom, I've found it usually takes 2 builds to get the project to work as it only restores the missing .NET NuGet on build, then all references work.

tomspilman:
Ok... worked out the fixes on the build server and our build script.  Closing this an submitting a new clean PR for it.

PR #4324, opened 2015-12-16
Pass parameters from MaterialProcessor to TextureProcessor.

nkast:
"PremultiplyTextureAlpha" --> "PremultiplyAlpha"
"ResizeTexturesToPowerOfTwo" --> "ResizeToPowerOfTwo"

Thanks to @mgarstenauer for the fix.


nkast:
parametersred

tomspilman:
It would be really good to get a unit test for `MaterialProcessor` in place which could have caught this sooner.  Sort of like i wrote for the `TextureProcessor`:

Have any time @nkast ?

tomspilman:
Going to merge the fix for now...  but unit tests really are needed.

nkast:
Ok, noted.

tomspilman:
Added an issue specific to that here: https://github.com/mono/MonoGame/issues/4333.

PR #4327, opened 2015-12-17
Fix UAP/UWP Build

tomspilman:
This PR fixes the Windows UAP/UWP build which was silently broken for months now.

The main issue is that MSBuild doesn't perform the NuGet restore like running from the IDE.  So we applied the recommended work around.

Also we detected that Protobuild was smashing over the existing dependences and kickstart folders with its own git operation on the build agents.  I've disabled this Protobuild feature until that can be fixed.

More details in https://github.com/mono/MonoGame/pull/4323.


tomspilman:
@KonajuGames @dellis1972 - This is good to merge.

KonajuGames:
Merging.

tomspilman:
Thanks!

PR #4328, opened 2015-12-17
[Pipeline] Support Long filenames

### LINKED ISSUE #4186: Content pipeline showing wrong location and root node name (Windows)
When I create a new monogame project and open the MGCB file, the location of the MGCB file shown in the GUI is wrong (see attached image). The actual location of the file is what it should be, not the one shown in the pipeline GUI.

The root node is also shown with the name "CONTEN~1" instead of "Content", which I guess is related to the location name error.

I'm using swedish as my Windows language.

pipeline location

nkast:
Open files with long filenames/paths by double click.
Fix #4186


nkast:
MonoGameSetup.exe from the build server works.
Bug fixed.

tomspilman:
Looks good to me... merging!  Thanks @nkast !

PR #4330, opened 2015-12-17
[mgcb] Suppress warnings for assets allready build.

nkast:
Suppress warnings for assets allready build in the current process, otherwise 3D models with multiple textures will flood the output with 'Skipping modelTexture'.
The warning will be shown for unmodified assets when build without '/Rebuild' command.


nkast:
This depends on fix: #4324 , otherwise the pipeline will build multiple texture xnb for each mesh.

nkast:
@Helikin , @mgarstenauer does this look ok?

mgarstenauer:
@nkast I'll review it tomorrow.

mgarstenauer:
I did some tests and found that this does not work in all cases. Let's say "TextureA.png" is already built. The texture is referenced multiple times. In this case, when I build the content project I still get the "Skipping TextureA.png" every time it is referenced.

I can't think of a simple solution other than tracking all "Suppress ..." messages and checking for duplicates before printing a next message.

nkast:
I see. Thanks @mgarstenauer.

PR #4334, opened 2015-12-17
Fix for TextureImporter when importing dds files with simple texture

Tzenchor:
After #4136, texture importer couldn't handle a simple texture inside a dds file.
This pull fixes that and adds two tests.
One test imports a simple texture, and the second imports a texture with mipmaps


tomspilman:
Looks like maybe your newline settings are wrong in Git?  @dellis1972 ?

Tzenchor:
I think the problem might have been that in this computer if I type in the console "git config --system core.autocrlf" it is blank while in the other it is set to true, but in both cases it is set to true in "git config core.autocrlf"

tomspilman:
@Tzenchor - That sounds possible.  We will need to see the line endings fixed before we can consider merging this.

Tzenchor:
I forgot to add the new assets for the tests to protobuild.
Should i create a fresh pull once i fix the problem with the line endings or do it here?

tomspilman:
> Should i create a fresh pull once i fix the problem with the line endings or do it here?

Well if you know how to flatten the commits into a single one so that the history of the bad line endings is gone...  then go ahead and do it here.

Tzenchor:
I'm not a wizard outside d&d so i'll do a new one

PR #4337, opened 2015-12-18
[Direct3D] Fix setting fullscreen state while device not being ready

### LINKED ISSUE #4148: Windows DirectX: Occasionally crashes when entering fullscreen mode
A few of my testers keep reporting the following crash when launching my game:
```
Type: SharpDX.SharpDXException
Message: HRESULT: [0x887A0022], Module: [SharpDX.DXGI], ApiCode: [DXGI_ERROR_NOT_CURRENTLY_AVAILABLE/NotCurrentlyAvailable], Message: A resource is not available at the time of the call, but may become available later.

Source: SharpDX
Stack Trace:
   at SharpDX.DXGI.SwapChain.SetFullscreenState(Bool fullscreen, Output targetRef)
   at Microsoft.Xna.Framework.Graphics.GraphicsDevice.CreateSizeDependentResources(Boolean useFullscreenParameter)
   at MonoGame.Framework.WinFormsGamePlatform.EnterFullScreen()
   at Microsoft.Xna.Framework.Game.DoInitialize()
   at Microsoft.Xna.Framework.Game.Run(GameRunBehavior runBehavior)
   at FinalDays.Client.Program.Main(String[] args)
```
The crash is intermittent, and usually it works fine if they try and launch the game a second time.

Looking at the exception details, it seems to be related to entering fullscreen mode when the game is starting up.

I've personally only experienced this crash once when using fullscreen mode before. But typically I am always running in windowed mode while testing, so that's probably why I've only seen it once.

I'm planning to take a closer look into this myself sometime soon hopefully, but I was hoping that maybe someone else might have seen this before and know what the cause is.

mrhelmut:
Fixes #4148

I'm not 100% confident about this one. What should be checked:
- if it uses the correct method to compare exception's HRESULT
- if it is safe to retry setting the fullscreen state after `_swapChain.Present(syncInterval, PresentFlags.None);`

This bug is really hard to reproduce, so I'm going to roll this PR out to our game today and will check if the problem is still referenced in our crash reports DB over the week. I've never been able to observe it locally, but we're typically getting about 1 of these crashes per day, so I'll be able to monitor its occurrence.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -714,9 +714,10 @@ namespace Microsoft.Xna.Framework.Graphics
                 // This force to switch to fullscreen mode when hardware mode enabled(working in WindowsDX mode).
                 if (useFullscreenParameter)
                 {
-                    // Entering fullscreen mode may raise a SharpDXException with HRESULT 0x887A0022 if the device is not yet ready to do so.
+                    // Entering fullscreen mode may raise a SharpDXException with HRESULT 0x887A0022 [DXGI_ERROR_NOT_CURRENTLY_AVAILABLE/NotCurrentlyAvailable] if the device is not yet ready to do so.
                     // The DXGI documention states that "when this error is returned, an application can continue to run in windowed mode and try to switch to full-screen mode later"
                     // See https://msdn.microsoft.com/en-us/library/windows/desktop/bb174579(v=vs.85).aspx
+                    // Another error may be raised here, 0x887A0004 [DXGI_ERROR_UNSUPPORTED/Unsupported], upon which we disable fullscreen state
                     try
                     {
                         _swapChain.SetFullscreenState(PresentationParameters.IsFullScreen, null);
@@ -724,8 +725,12 @@ namespace Microsoft.Xna.Framework.Graphics
                     }
                     catch (SharpDXException ex)
                     {
-                        if (ex.ResultCode.Code == unchecked((int)0x887A0022))
+                        if (ex.ResultCode.Code == unchecked((int)0x887A0022)) // [DXGI_ERROR_NOT_CURRENTLY_AVAILABLE/NotCurrentlyAvailable]
                             _shouldRetrySettingFullscreen = true;
+                        else if (ex.ResultCode.Code == unchecked((int)0x887A0004)) // [DXGI_ERROR_UNSUPPORTED/Unsupported]
+                        {
+                            PresentationParameters.IsFullScreen = false;
+                        }
                         else
                             throw ex;
                     }
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -717,7 +717,7 @@ namespace Microsoft.Xna.Framework.Graphics
                     // Entering fullscreen mode may raise a SharpDXException with HRESULT 0x887A0022 [DXGI_ERROR_NOT_CURRENTLY_AVAILABLE/NotCurrentlyAvailable] if the device is not yet ready to do so.
                     // The DXGI documention states that "when this error is returned, an application can continue to run in windowed mode and try to switch to full-screen mode later"
                     // See https://msdn.microsoft.com/en-us/library/windows/desktop/bb174579(v=vs.85).aspx
-                    // Another error may be raised here, 0x887A0004 [DXGI_ERROR_UNSUPPORTED/Unsupported], upon which we disable fullscreen state
+                    // Another error may be raised here, 0x887A0004 [DXGI_ERROR_UNSUPPORTED/Unsupported], upon which we fall back to borberless in EnterFullScreen()
                     try
                     {
                         _swapChain.SetFullscreenState(PresentationParameters.IsFullScreen, null);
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -951,10 +967,22 @@ public void PlatformPresent()
...
+                        _swapChain.SetFullscreenState(PresentationParameters.IsFullScreen, null);
+                        _shouldRetrySettingFullscreen = false;
+                    }
+                }
             }
-            catch (SharpDX.SharpDXException)
+            catch (SharpDX.SharpDXException ex)
             {
+                // Exception related to the device not being ready shall be silent because it can safely retry to set the fullscreen mode later
+                if (_shouldRetrySettingFullscreen && unchecked((uint)ex.ResultCode.Code) != 0x887A0022)
```
MichaelDePiazzi: This logic is incorrect. It should be:
```c#
if (!(_shouldRetrySettingFullscreen && unchecked((uint)ex.ResultCode.Code) == 0x887A0022))
```
Or alternatively:
```c#
if (!_shouldRetrySettingFullscreen || unchecked((uint)ex.ResultCode.Code) != 0x887A0022)
```

MichaelDePiazzi:
> UPDATE: I can already tell that it still occurs. I'm most probably not comparing the HRESULT correctly.

The way you're checking HRESULT should work (assuming SharpDX is populating `ResultCode.Code` with the expected HRESULT). An alternative way of checking it would be:
```c#
ex.ResultCode.Code == unchecked((int)0x887A0022)
```
I did find a bug in your logic though. I've made a comment on your code about it.

mrhelmut:
Ah yes, that logic line was wrong, thanks.

I've just noticed that there are two different HRESULT codes involved in this issue:
- 0x887A0022 [DXGI_ERROR_NOT_CURRENTLY_AVAILABLE/NotCurrentlyAvailable]
- 0x887A0004 [DXGI_ERROR_UNSUPPORTED/Unsupported]

We received more crash reports, and I can confirm that the first one is correctly caught. The errors we were still getting were in fact only related to the second one.

I've updated this PR to correctly fix DXGI_ERROR_NOT_CURRENTLY_AVAILABLE.

What should we do regarding DXGI_ERROR_UNSUPPORTED? @tomspilman @KonajuGames 

Should we leave this uncaught, or should we do something about it, like switching back to ```HardwareModeSwitch = false;```?

Out of all our reported crashes, DXGI_ERROR_NOT_CURRENTLY_AVAILABLE represents 7.74% and DXGI_ERROR_UNSUPPORTED represents 5.16%.
I find that to be quite high in both cases and would suggest to fall back to borderless.

mrhelmut:
Quick update:

0x887A0004 [DXGI_ERROR_UNSUPPORTED/Unsupported] seems fixed, we're not getting new reports of this one.

0x887A0022 [DXGI_ERROR_NOT_CURRENTLY_AVAILABLE/NotCurrentlyAvailable] is not fixed, the exception is correctly caught, but trying to set the fullscreen state in ```PlatformPresent()``` seems to be wrong as it now raises a 0x80004005 [E_FAIL/Unspecified error] here.

mrhelmut:
Back with several hundreds sessions registered with this fix.

0x887A0004 [DXGI_ERROR_UNSUPPORTED/Unsupported] hasn't been recorded.

0x887A0022 [DXGI_ERROR_NOT_CURRENTLY_AVAILABLE/NotCurrentlyAvailable] doesn't occur anymore and players are not complaining about this anymore, but...

... it still seems to occur very rarely by throwing a 0x887A0005 [DXGI_ERROR_DEVICE_REMOVED/DeviceRemoved] in ```WinFormsGamePlatform.Present()```.

The two initial errors don't happen anymore (they were happening at least once per day before), while this new error only occurred 2 times.

I absolutely don't know if this new error is related to this PR not doing well by retrying to set the fullscreen state in ```WinFormsGamePlatform.Present()``` or if it is totally unrelated (there is a comment in ```WinFormsGamePlatform.Present()``` having a TODO marker on loss of device, so maybe it's related to this and it's just a pure coincidence that it started to happen when I started to test this fix).

Here's more information regarding DXGI_ERROR_DEVICE_REMOVED: 
> "The video card has been physically removed from the system, or a driver upgrade for the video card has occurred. The application should destroy and recreate the device. For help debugging the problem, call ID3D10Device::GetDeviceRemovedReason."

If someone can confirm that this is not related to this PR or the original issue, I would suggest to ignore it here and open a new issue.

tomspilman:
If you just search for `DXGI_ERROR_DEVICE_REMOVED` you'll see lots of games have this error reported against them.   It can occur when the graphics driver resets....  which can be triggered by all sorts of things:

  - Damaged card.
  - Overheating.
  - Bad overclocking settings.
  - Bad driver.
  - Driver bugs.

Just about every game is going to run into some of these because people have screwed up systems in the wild.

The best you can do within MonoGame is capturing this and throwing the appropriate XNA exception.... `DeviceLostException` in this case I think. 

There is also `DeviceNotResetException` which we probably should be thowing instead of retrying internally.

My understanding is that this old XNA sample...

http://xbox.create.msdn.com/en-US/education/catalog/sample/winforms_series_1

... shows how you should be handing these exceptions in the WinForm code.

mrhelmut:
Great, thank you for these information, I'll try to take a look at this.

Since my last comment, none of these errors have been reported. @tomspilman do you feel like this PR could be merged? If it's fine for you, I prefer to get these fixes in before tackling ```DXGI_ERROR_DEVICE_REMOVED```. It should already fix 99% of fullscreen issues.

mrhelmut:
This looks fairly stable. We had a bunch of new players during the holiday and no crash has been recorded.

spunky44:
Will it be possible to fix the conflicts and merge with MonoGame soon?
I am asking because we are experiencing the same issue.

mrhelmut:
Would you mind trying this commit @spunky44 ? I believe it didn't get merged because it need to be tested on a bigger scale to be sure that it works correctly.

Yet, I believe @tomspilman would prefer that this problem is handled more cleanly by tracking loss of device and resetting it when needed.

This fix is not quite clean, and I personally don't have my hands in the DX code right now.

spunky44:
I merged it, but it is crashing in fullscreen, behind the window. So I cannot see the code but have to log off. I am currently trying to get the game to run with dual monitors so I can find out more.

spunky44:
I am going to revert back, since fullscreen is no longer working. It was always brittle and I can't figure out which of the changes made it break.

mrhelmut:
Well, I believe recent changes are conflicting and make it not work correctly. I don't think that I'll have the time to take a look at it myself.

spunky44:
Could the code be made simpler with a loop that simply retries the call X times? Perhaps with a slight delay?

mrhelmut:
Given my DXGI knowledge and the fact this PR may be obsolete with the recent enhancements in the DX support, I'm closing this PR.

PR #4338, opened 2015-12-18
Fix for TextureImporter when importing dds files with simple texture

Tzenchor:
Second try after fixing the problem with line endings (see #4334).
After #4136, texture importer couldn't handle a simple texture inside a dds file.
This pull fixes that and adds two tests.
One test imports a simple texture, and the second imports a texture with mipmaps


KonajuGames:
Thanks @Tzenchor. Merging.

PR #4339, opened 2015-12-18
Initial check-in of .Core solution

DDReaper:
Moved non-platform specific classes to a central PCL project, for reuse by frameworks using basic MG types.

Next step, generate Bait and switch separate PCL for platform specific types/classes


Build/Projects/FrameworkReferences.Net.definition
```diff
@@ -7,12 +7,14 @@
     <Reference Include="System.Core" />
     <Reference Include="Xamarin.iOS" />
     <Reference Include="Lidgren.Network" />
+    <Reference Include="MonoGame.Framework.Core" />
```
hach-que: You don't need to do this here, you can just add `<Reference Include="MonoGame.Framework.Core" />` outside all of the platform-specific blocks.

Build/Projects/Framework.Content.Pipeline.References.definition
```diff
@@ -43,6 +43,7 @@
     <Binary
       Name="ATI.TextureConverter"
       Path="ThirdParty\Dependencies\ATI.TextureConverter\ATI.TextureConverter.dll" />
+    <Reference Include="MonoGame.Framework.Core" />
```
hach-que: You don't need to do this here, you can just add `<Reference Include="MonoGame.Framework.Core" />` outside all of the platform-specific blocks.

Build/Projects/FrameworkReferences.definition
```diff
@@ -7,12 +7,16 @@
     <Reference Include="System.Core" />
     <Reference Include="Xamarin.iOS" />
     <Reference Include="OpenTK-1.0" />
+    <Reference Include="MonoGame.Framework.Core" />
```
hach-que: What I said applies to all ExternalProjects ;P

Build/Projects/MonoGame.Framework.definition
```diff
@@ -1289,6 +1155,423 @@
       <Services>DirectXGraphics</Services>
     </EmbeddedResource>
 
+		<!-- PCL Platform -->
```
hach-que: If this set of files is shared between the PCL project and Linux/Web, you'll be better off using an Include Project: https://protobuild.readthedocs.org/en/latest/include_projects.html  Put all of the files (with no platform conditionals) in the include project, then reference the include project from the PCL, and reference the include project under the platform conditionals for the FrameworkReferences external project.

hach-que:
Actually it looks like the PCL isn't being applied to all platforms, so ignore what I said about moving the reference outside of the platform specific blocks.

DDReaper:
Anyone got any ideas on this error coming out of Nant?
"CSC : error CS1617: Invalid option '6' for /langversion; must be ISO-1, ISO-2, 3, 4, 5 or Default [C:\TeamCity\buildAgent\work\f26e5a04a66fcbdb\MonoGame.Framework\MonoGame.Framework.Core.Windows.csproj]"
Is this a TeamCity issue @dellis1972 / @tomspilman ?
or a ProtoBuild issue @hach-que ?

dellis1972:
it means the project is using c# 6 and it looks like only c# 5 is valid.. c# 5 is only available in the 14.0 build tools afaik

dellis1972:
so something is being generated with a c# 6 version.

DDReaper:
OK, tried @hach-que 's suggestion of using an include and it certainly makes the protobuild definitions cleaner only listing the files once.
In the project, the shared files show up in an "Included Code" folder, is this acceptable.  makes do difference to the external API.

Another observation is that a PCL for Linux is actually being generated, although not requested in the .Core definition.  What is weirder is that this is referencable in the Linux solution.  Can anyone with Linux testing validate if this reference also works for them, in which case it's another platform that can natively support it now.

hach-que:
PCLs only work on end user systems if they install the reference assemblies, which aren't installed with mono-complete, so it makes it more awkward for Linux users to use. The fact that reference assemblies aren't pre-installed on Linux (and apparently aren't available on Debian at all) is one reason I so strongly advise against them.

To disable the Core project on Linux, you need to set the Platforms attribute on the project definition.

C# 6 is the default language level, along with .NET 4.5. You'll need to set `<LangVersion>5</LangVersion>` in the properties for Core to make it build on the MG build server because MSBuild 14 isn't installed there.

cra0zy:
> PCLs only work on end user systems if they install the reference assemblies, which aren't installed with mono-complete, so it makes it more awkward for Linux users to use.

Anyone trying to install MonoGame on Ubuntu will get PCL support auto installed: https://github.com/mono/MonoGame/blob/develop/Installers/Linux/DEBIAN/control#L4

DDReaper:
Looks like it is ProtoBuild related as it's generating this in the .csproj for the PCL
    <NoWarn></NoWarn><TargetFrameworkProfile>Profile111</TargetFrameworkProfile><TargetFrameworkVersion>v4.5</TargetFrameworkVersion><MinimumVisualStudioVersion>10.0</MinimumVisualStudioVersion><LangVersion>6</LangVersion>

where in the other framework csproj's it's all separated and the LangVersion is as follows:
<LangVersion>Default</LangVersion>

Can't see anything obvious in the .definition files to control this

hach-que:
Oh yeah, set it to Default if that's what's done elsewhere (the reason we
don't actually set it to Default is because you'll be targeting different
language versions on different systems, based on whatever the latest build
tools you have is. The only reason it's necessary to do this override is
because of the MG build server, and in future it should be removed)

DDReaper:
So how in the definition files do I set it to default @hach-que

DDReaper:
On the Linux front them @cra0zy is it worth adding Linux in to the platforms directly supported by the PCL core lib?

hach-que:
Like this:
https://github.com/mono/MonoGame/blob/develop/Build/Projects/MonoGame.Framework.definition#L33

cra0zy:
> On the Linux front them @cra0zy is it worth adding Linux in to the platforms directly supported by the PCL core lib?

Yes, it is.

DDReaper:
Ok, will check this build works, then I'll also add the Core PCL to Linux as well

DDReaper:
Ok, err, help?  I'll double check locally but nothing that has changed should cause the tests to fail?

DDReaper:
OK, validated locally and most of the bitmap tests fail. Switched to the develop branch and they too fail with the same errors? How can that be?  How on earth could a build failing tests be current?
Going to resynch my develop branch and test some more, even though I started with this copy only days ago (started fresh again)

DDReaper:
Nope, at a loss.  Even from a fresh clone and submodule update, I cannot get the Bitmap / compress tests to run.
However I noticed That more recent builds have passed those tests, so unsure as to what is going on?

DDReaper:
For reference the errors mix between, Could not load content file for font's, to:

Test Name:	BitmapCompressFullNoResize
Test FullName:	MonoGame.Tests.ContentPipeline.BitmapContentTests.BitmapCompressFullNoResize
Test Source:	c:\Development\Github\MonoGameRefresh\Test\ContentPipeline\BitmapContentTests.cs : line 65
Test Outcome:	Failed
Test Duration:	0:00:00.229

Result Message:	System.BadImageFormatException : Could not load file or assembly 'PVRTexLibNET, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. An attempt was made to load a program with an incorrect format.
Result StackTrace:	
at Microsoft.Xna.Framework.Content.Pipeline.Graphics.DxtBitmapContent.TryCopyFrom(BitmapContent sourceBitmap, Rectangle sourceRegion, Rectangle destinationRegion)
at Microsoft.Xna.Framework.Content.Pipeline.Graphics.BitmapContent.Copy(BitmapContent sourceBitmap, Rectangle sourceRegion, BitmapContent destinationBitmap, Rectangle destinationRegion) in c:\Development\Github\MonoGameRefresh\MonoGame.Framework.Content.Pipeline\Graphics\BitmapContent.cs:line 132
at Microsoft.Xna.Framework.Content.Pipeline.Graphics.BitmapContent.Copy(BitmapContent sourceBitmap, BitmapContent destinationBitmap) in c:\Development\Github\MonoGameRefresh\MonoGame.Framework.Content.Pipeline\Graphics\BitmapContent.cs:line 87
at MonoGame.Tests.ContentPipeline.BitmapContentTests.BitmapCompressFullNoResize() in c:\Development\Github\MonoGameRefresh\Test\ContentPipeline\BitmapContentTests.cs:line 69

For bitmaps

KonajuGames:
Bad image format is usually from trying to run it on a 32-bit OS. The
libraries are 64-bit.

KonajuGames:
Or trying to run it as an x86 build.

DDReaper:
Mines x64 but I'll double check the build. Except it's also failing on the build server?

DDReaper:
Found root cause which was to do with reflection of types moved to the core project. Current reflection/contentpipeline fixed to MonoGame.Framework Assembly.

Will look to add more Content Type tests once this is merged.

Build/Projects/Framework.Content.Pipeline.References.definition
tomspilman: @DDReaper - Please make the change suggested here.

Build/Projects/FrameworkReferences.Net.definition
tomspilman: @DDReaper - Same.... please make this change.

Build/Projects/MGCB.definition
```diff
@@ -10,6 +10,7 @@
     <Reference Include="System.Xml" />
     <Reference Include="MonoGame.Framework" />
     <Reference Include="MonoGame.Framework.Content.Pipeline" />
+    <Reference Include="MGCBReferences" />
```
tomspilman: @DDReaper - You can remove this and the `MGCBReferences.definition` all together.  Just needs to reference `MonoGame.Framework.Core` here.

Build/Projects/FrameworkReferences.definition
```diff
@@ -25,6 +29,7 @@
         Name="OpenTK"
         Path="ThirdParty/Dependencies/OpenTK.dll" />
     </Service>
+    <Reference Include="MonoGame.Framework.Core.Include" />
```
tomspilman: Why is this here?

MonoGame.Framework/Content/ContentTypeReaderManager.cs
```diff
@@ -161,6 +161,14 @@ internal ContentTypeReader[] LoadAssetReaders(ContentReader reader)
                         readerTypeString = PrepareType(readerTypeString);
 
                         var l_readerType = Type.GetType(readerTypeString);
+
+                        if (l_readerType == null)
```
tomspilman: Can you add a comment in the code about what you are doing here and why?

Build/Projects/FrameworkReferences.definition
hach-que: That will pull in the `MonoGame.Framework.Core.Include` include project because the `Linux` platform isn't using PCLs here.

I'm personally still extremely reserved about making the Linux platform reference PCLs since this is still a huge mess across distros at the moment; the Linux platform is the reason Protobuild warns so strongly against using them, because the issues they cause can manifest themselves on end-user machines or developer machines, not build servers.  Debian for example doesn't even package the PCL reference assemblies (along with an ancient version of Mono presumably), so users need to install these manually if a PCL is being targeted.

But @cra0zy seems to think it's fine so I don't know... but if I do run into issues relating to PCLs on Linux in MonoGame I will file issues here since it's not practical to be asking end-users to manually download ZIPs and install them into /usr/lib/mono or where-ever they're meant to go.

tomspilman: > because the Linux platform isn't using PCLs here.

Oh...  well is that why @DDReaper is putting `<Reference Include="MonoGame.Framework.Core" />` into platform specific sections?

dellis1972: An end user should not need the pcl reference assemblies. Since they are just that references, i.e shims with no code in them only the public api. Those "reference assemblies" should be replaced by the real ones from the mono framework. They are only really needed on a dev machine  for building. If that doesn't work then something is truly wrong with linux ;)

-----Original Message-----

hach-que: Yes I made those comments and then corrected myself further in the thread

hach-que: I meant end users of MonoGame, i.e. The developers using it to build games.

Installing the PCL reference assemblies has come up twice in the monodevelop Gitter channel this week, and this was from people using Debian and Gentoo. Any Linux distribution that mandates the source code must be available for anything they package can't distribute the reference assemblies because they're not open source. That in turn means that on no Linux distros do you get them by installing mono-complete, you have to seperately install the reference-assemblies package. It's a lot of pain for very little benefit, especially since Protobuild makes it pretty easy to just include the files on the Linux platform and avoid any of the support issues that would arise otherwise.

DDReaper: Ok, there are two types of references used here:
```
MonoGame.Framework.Core
```
Which is the reference to the Core PCL, containing those non-platform specific classes
```
MonoGame.Framework.Core.include
```
which includes the non-platform specific files in the PCL in the host project.  Meaning this project doesn't use the PCL.  This is done for Linux and Web which don't support PCL references as yet.  I've also excluded WindowsPhone8 from the PCL as it caused a mess with the XNA references and there's no need to change a platform that will be phased out soon.

@cra0zy has indicated that Linux does support PCL references but to keep it clean, I've not included it in this change.  Will submit a separate one once this is done.

MonoGame.Framework/Content/ContentTypeReaderManager.cs
DDReaper: Fair enough, will do.

Build/Projects/MGCB.definition
DDReaper: Except the MGCB project also includes the Linux platform and as stated, Linux doesn't support the PCL (it used the files as it does now)

Unless there is another way to configure references to be platform specific or exclude Linux?

hach-que: Yeah, if you make an `ExternalProject` that will let you conditionally reference another project based on the platform.

Build/Projects/FrameworkReferences.definition
DDReaper: @dellis1972 to clear one thing up.  This change is to move non-platform specific code to it's own PCL project, this will result in all projects (except Linux/Web/WP8) to have two references per project:
* MonoGame.Framework.Core (core PCL)
* MonoGame.Framework

The Core PCL can also then be used by frameworks and projects that only need the base types from MG instead of the entire lib. (such as greatly requested by the likes of Farseer and others), this means addon frameworks will have an easier time supporting MG (and they don't need to platformise their projects)

For the Bait and switch PCL lib, that will be a separate undertaking to enable consumers to build games in a PCL project. This is different to this change, that comes later.  The added advantage of the above is that it eases this process needing less of the framework to be faked in a PCL.

Build/Projects/FrameworkReferences.Net.definition
DDReaper: As @hach-que noted later @tomspilman this comment and other related one's we recanted, because Linux, Web and WP8 are excluded from this, we can't make it global.

Build/Projects/MGCB.definition
DDReaper: So is this the best way to configure this @hach-que so that the reference is only added for Windows and Mac?  Excluding Linux

Build/Projects/FrameworkReferences.definition
cra0zy: Wouldn't it be more cleaner if you added the Linux change in here? Just saying...

DDReaper: I tried adding it and the TeamCity build process complained.  Once merged I'll fast track a Linux update so I can test it independently.

DDReaper:
Should be ready to merge now if there are no other queries.

Will work on other requirements (such as adding Linux) separately and also the generation of the bait and switch PCL for the platform specific classes.

DDReaper:
Finished completing tests and all platforms behave as expected.

Provided there are no other concerns, this should be ready to merge @tomspilman @KonajuGames

DDReaper:
Any updates @tomspilman @KonajuGames  
The PR has a risk of becoming stale and requiring more work. Is anything more needed to progress this?

tomspilman:
We need to get 3.5 out the door first.

DDReaper:
That's a real shame as it will push official PCL support back yet another release.

dsaf:
Example of a project that would exactly benefit from this:

http://community.monogame.net/t/nez-free-open-source-2d-framework/7010

https://github.com/prime31/Nez/blob/master/Nez-PCL/packages.config
(currently uses MonoGame.Framework.Portable 3.2.99.1-Beta)

@prime31

prime31:
+1 from me of course. PCL support is a must for sharing code amongst multiple projects and people.

dsaf:
@tomspilman Does this really mean pushing to 3.6 or 3.5.x with PCL still likely this year?

prime31:
Just in case it is of use I just happened across CocosSharp's way of dealing with a MonoGame PCL [here](https://github.com/mono/CocosSharp/blob/master/PCL/MonoGamePCLShared/MonoGame.FrameworkAPI.cs).

PR #4243, opened 2015-11-14
[Pipeline Tool] Use system gtk-sharp 3

cra0zy:
Stuff done:
 - added LocalCopy argument to protobuild and disabled copying of gtk-sharp3 dll files, this forces system gtk-sharp3 libraries to be used instead
 - disabled copying of gtk-sharp3 config files because of the above
 - updated installer and documentation dependencies

PR #4341, opened 2015-12-19
Updated Protobuild

cra0zy:
This fixes the Linux Pipeline Tool problem @hach-que caused with #4282

Long story short, she did not merge changes of #4243 done to xslt so LocalCopy was setting to true, and because libraries were being copied, Pipeline Tool would use local dlls instead of system ones(and this causes a crash because no config files are copied and tool tries to load Window dlls).


hach-que:
Do not merge this. I haven't merged in the PR upstream, and part of merging in upstream involves rebuilding Protobuild based on the PR changes, so the final executable that gets merged will have a different hash.

And for the record, I'm a woman; she is the correct pronoun.

hach-que:
I should clarify, technically you can merge this, but the version of Protobuild will technically be divergent from upstream since it won't ever have lived in the `master` branch.

cra0zy:
> And for the record, I'm a woman; she is the correct pronoun.

Sorry about that.

cra0zy:
Updated with Protobuild from Protobuild repo.

hach-que:
Can you update this PR with the latest version before it's merged (I pushed some more changes to Protobuild and if we're updating it then we might as well put it on the latest version)?

cra0zy:
> Can you update this PR with the latest version before it's merged (I pushed some more changes to Protobuild and if we're updating it then we might as well put it on the latest version)?

done.

cra0zy:
@tomspilman @KonajuGames Please merge this soon, MG Linux build is broken without it :(

KonajuGames:
Merged.

cra0zy:
Thank you ^^

PR #4343, opened 2015-12-21
Add PlatformGetMaxNumberOfGamePads method to Xamarin.Mac

hach-que:
Looks like someone changed the implementation of GamePads but didn't update the Xamarin.Mac implementation (or it was just because of the timing in merging the Xamarin.Mac changes in).

This updated the Xamarin.Mac input implementation to have the right methods on it.

### DIFF
--- a/MonoGame.Framework/Input/GamePad.MacOS.cs
+++ b/MonoGame.Framework/Input/GamePad.MacOS.cs
@@ -13,24 +13,24 @@ namespace Microsoft.Xna.Framework.Input
     /// </summary>
     public static partial class GamePad
     {
-        private static int PlatformGetMaxIndex()
+        private static int PlatformGetMaxNumberOfGamePads()
         {
             return 0;
         }
 
         private static GamePadCapabilities PlatformGetCapabilities(int index)
         {
-            throw new NotSupportedException("We don't support game pads under the new Xamarin.Mac API yet");
+            return new GamePadCapabilities { IsConnected = false };
         }
-
+               
         private static GamePadState PlatformGetState(int index, GamePadDeadZone deadZoneMode)
         {
-            throw new NotSupportedException("We don't support game pads under the new Xamarin.Mac API yet");
+            return new GamePadState() { IsConnected = false };
         }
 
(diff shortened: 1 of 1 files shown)


cra0zy:
I changed the name with #4273

hach-que:
@tomspilman @KonajuGames this should be merged before the 3.5 release as it's a fix for the Xamarin.Mac support

tomspilman:
I guess the build didn't fail for this because currently we're not building Xamarin.Mac....  just old MonoMac.  Correct?

If so we need to fix the build server ASAP to at the very least build that platform.

hach-que:
Correct, the fastest way to solve this is to apply for an open source Xamarin.Mac license: https://resources.xamarin.com/open-source-contributor.html and to install that license on the build server.

However, once you do so, the build server will start producing binaries for Xamarin.Mac instead of MonoMac.  There's no current way to force the build for both, and I don't really want to add another platform here because it's just an API change, not a difference in the native platform.  I was thinking about some sort of `--force-legacy-api` command-line switch but I'm not sure yet.

hach-que:
(I am however tracking upstream on my own build server at https://code.redpointgames.com.au/B10751, which is currently failing because this PR isn't merged in)

tomspilman:
> However, once you do so, the build server will start producing binaries for Xamarin.Mac instead of MonoMac.

Great... another mess to mop up.

So what is actually different when building for one or the other?  Is it just like some `BUILD_XAMARIN_MAC` sort of preprocessor define?  Or are there more differences to the CSProj?

hach-que:
There's differences in the C# project files, all of the namespaces have
changed and a few of the APIs have been dropped.

tomspilman:
> There's differences in the C# project files

So that means Prototbuild has to support some way to toggle generation for Xamarin.Mac somehow.

hach-que:
Yes, currently it just picks the latest API that you have installed.

hach-que:
@tomspilman Can we get this merged in the meantime?

tomspilman:
Sorry... Been a busy day.  Merging!

hach-que:
Thank you! ^_^

PR #4348, opened 2015-12-22
Reduced garbage generation in SpriteBatcher

MichaelDePiazzi:
I was doing some profiling of my game, and came across this:

before

What was happening was that the vertex array was only growing to exactly match the number of items in a batch, rather than growing by a percentage like the sprite batch array is. This could put a lot of load on the GC if the number of sprites being drawn was increasing over time (which was happening in my game) as the vertex array would continually be getting resized.

To resolve this, I changed the vertex array to grow in line with the sprite batch array instead. After profiling again, it now looks like this:

after

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatcher.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatcher.cs
@@ -119,6 +119,8 @@ namespace Microsoft.Xna.Framework.Graphics
                 Array.Resize(ref _batchItemList, newSize);
                 for(int i=oldSize; i<newSize; i++)
                     _batchItemList[i]=new SpriteBatchItem();
+
+                EnsureArrayCapacity(Math.Min(newSize, MaxBatchSize));
             }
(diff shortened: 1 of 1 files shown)


tomspilman:
@nkast - Thoughts?

KonajuGames:
A nice reduction in allocations there.

nkast:
Good.  A nice way to reduce garbage generation.

By the way what profiler is the screenshots from? 
Are those new allocations over the same period of time?

MichaelDePiazzi:
> By the way what profiler is the screenshots from?

CLR Profiler (https://clrprofiler.codeplex.com/)

> Are those new allocations over the same period of time?

Yes, pretty much. I did the same run through a stage in my game as consistently as I could.

tomspilman:
Ok... lets merge this!  Thanks @MichaelDePiazzi !

MichaelDePiazzi:
Cool, thanks guys :thumbsup:

Issue #4154, opened 2015-09-25
Win Pipeline Tool wrong platform selection


cra0zy:
So enmus WindowsGL and Linux point to enum DesktopGL, but the problem is that Windows PropertyGrid read the enum Linux first, and since it had the same value as the other 2 it decided to just give "Linux" 3 times in target platform selection instead of giving "WindowsGL", "Linux" and "Mac".

![untitled3](https://cloud.githubusercontent.com/assets/6773302/10102661/fee2f5de-639f-11e5-87a6-f05e934dacb3.png)

cra0zy:
@dellis1972 @KonajuGames Can you add this to 3.5 milestone?

cra0zy:
@nkast Do you by any chance have an idea on how we can fix this?

nkast:
What's the story behind this? Do we keep WindowsGL & Linux for desirialization or something?

Do we want to hide the old values or to keep show them? I tried some attributes (Displayname, editorbrowsable,etc) but nothing works. This probably need a Typeconverter.

nkast:
Wouldn't be better to remove the duplicate values and handle deserialization of older files or that would be extra work? Is this enum used on anything other than .mgcb files;

cra0zy:
> Do we want to hide the old values or to keep show them?

It doesn't matter if WindowsGL/Linux are visible there, the only thing that matters is that DesktopGL is visible.

> Wouldn't be better to remove the duplicate values and handle deserialization of older files or that would be extra work? Is this enum used on anything other than .mgcb files;

They are just used for .mgcb from what I could tell. Also I am fine with what you have suggested.

PS. The best part about this bug is that it doesn't occur on Linux/Mono Windows Forms...

nkast:
The problem is with `SortedEnumTypeConverter` 
https://github.com/mono/MonoGame/blob/0b25f56e861fb602db884c40267b99ea947bae15/Tools/Pipeline/Windows/PipelineProjectProxy.cs#L57

Apart from that, I think we should drop obsolete values from `TargetPlatform and add a `TargetPlatformTypeConverter` to deserialize older files.

nkast:
I made the two obsolete platforms hidden and moved DesktopGL last. 

Everybody seems to agree that `Enum.GetName()` always return the last Field name but officially the documentation says that we shouldn't expect consisted results. Officially this is unfixable!
It makes sense since they all have the same value, it's impossible to determine which is which at runtime.

[...]

PR #4350, opened 2015-12-22
Deserialize WindowsGL & Linux -> DesktopGL

nkast:
...and remove them from TargetPlatform enumeration.

Fix for #4154

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/PathHelper.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PathHelper.cs
@@ -65,7 +65,7 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
             switch (targetPlatform)
             {
                 case TargetPlatform.Windows:
-                case TargetPlatform.WindowsGL:
+                case TargetPlatform.DesktopGL:
                 case TargetPlatform.WindowsPhone8:
                 case TargetPlatform.WindowsStoreApp:
                     return NormalizeWindows(path);
--- a/MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
+++ b/MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
@@ -1,13 +1,17 @@
 ﻿// MonoGame - Copyright (C) The MonoGame Team
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
+
 using System;
+using System.ComponentModel;
+using System.Globalization;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline
 {
     /// <summary>
     /// Identifiers for the target platform.
     /// </summary>
+    [TypeConverter(typeof(TargetPlatformTypeConverter))]
     public enum TargetPlatform
     {
         /// <summary>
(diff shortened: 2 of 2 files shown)


MonoGame.Framework.Content.Pipeline/Builder/PathHelper.cs
```diff
@@ -65,7 +65,7 @@ public static string NormalizeOS(string path, TargetPlatform targetPlatform)
             switch (targetPlatform)
             {
                 case TargetPlatform.Windows:
-                case TargetPlatform.WindowsGL:
+                case TargetPlatform.DesktopGL:
```
cra0zy: Remove this line completely, it shouldn't have been there in the first place.

nkast: I's rather not alter the current behavior with a PR that's unrelated to PathHelper.

tomspilman:
So does this also fix the platform dropdown in the Pipeline tool?  What appears in that list now?

cra0zy:
> So does this also fix the target platform dropdown in the Pipeline tool? What appears in that list now?

So, this does the following:
 - Pipeline Tool will only display DesktopGL
 - Pipeline Tool will not display WindowsGL and Linux
 - if Pipeline Tool loads something with WindowsGL or Linux platforms it will move it to DesktopGL platform
 - if MGCB comes across WindowsGL or Linux it will compile it as if it was DesktopGL

tomspilman:
> Pipeline Tool will not display WindowsGL and Linux

So it still shows MacOS?

cra0zy:
> So it still shows MacOS?

Yes... MacOS as a platform is same as WindowsDX... both have nothing to do with DesktopGL...

tomspilman:
> Yes... MacOS as a platform is same as WindowsDX... both have nothing to do with DesktopGL...

Why is that?  Shouldn't MacOS be a part of DesktopGL now?

hach-que:
DesktopGL supports Mac, but it doesn't support it actually being a Mac OS application. i.e. DesktopGL is just a normal .NET executable on Mac, it's not a Mac .app package (nor can you publish them to the Mac app store).

mrhelmut:
You can package DesktopGL projects to .app with MonoKickstart, that works just fine if one plans to publish through self hosting / Steam, but I don't know if such .app are Mac AppStore compliant (that's the whole purpose of Xamarin Mac and the new Mac API).

Both are valid paths IMO.

cra0zy:
Well, I don't think there is anything left for me to say, you guys said it all...

tomspilman:
> Both are valid and different paths IMO.

Ahh... ok... so this is sort of like the difference in a Windows desktop app and a Windows Store app.  Technically the same platform, but just different packaging methods for those platforms.  That whole concept makes this `TargetPlatform` enum really weird to understand.

cra0zy:
> Ahh... ok... so this is sort of like the difference in a Windows desktop app and a Windows Store app. Technically the same platform, but just different packaging methods for those platforms. That whole concept makes this TargetPlatform enum really weird to understand.

We should probably rename some TargetPlatform enums:
 - Windows > WindowsDX
 - MacOS > XamMac

tomspilman:
> We should probably rename some TargetPlatform enums:

I'm hesitant to do that until we have a bigger discussion about it.  Mostly it just seems like what is presented to the user via the Pipeline tool is really confusing as it is now.

tomspilman:
Thanks @nkast and @cra0zy !

PR #4352, opened 2015-12-22
[Linux Installer] Automatically override the old addin

cra0zy:


tomspilman:
Can you explain more specifically what this is doing?

cra0zy:
> Can you explain more specifically what this is doing?

Automatically overriding the old addin... if there is a MG addin is already installed it would ask a question:
```
MonoDevelop Add-in Setup Utility
The following packages will be installed:
 - MonoGame Project Templates v3.5.0.456
The following packages need to be uninstalled:
 - MonoGame Project Templates v3.5.0.455

Are you sure you want to continue? (y/N):
```

This just makes the answer always yes.

cra0zy:
One more thing, if you are worried about what if the user doesn't want to install the new adding, then:

1) Run installer already asks if you want to install the addin: https://github.com/mono/MonoGame/blob/develop/Installers/Linux/RUN/postinstall.sh#L73

2) Deb installer can't ask the question unless you run it in terminal, and people just run it in the software center which always just picks the default answer

cra0zy:
@tomspilman Can this be merged?

tomspilman:
Thanks @cra0zy !

Issue #4305, opened 2015-12-08
Blurry Fonts from the pipeline, works right with the XNA pipeline


Tzenchor:
Hi, I have a project i did for school with XNA and now i have converted it to Monogame.
But the weird thing is that if I run the game with what the Monogame pipeline generates i get blurry texts, both with 3.4 and 3.5, but if I just copy the Content folder from the XNA version it shows right.
XNA:
![xna](https://cloud.githubusercontent.com/assets/11319966/11668345/f6917c3e-9df6-11e5-89d3-94a60786dfbe.PNG)
MonoGame 3.4
![mono3 4](https://cloud.githubusercontent.com/assets/11319966/11668383/3515ee2c-9df7-11e5-811e-ec75f1d2fecd.PNG)
MonoGame 3.5 (current)
![mono 3 5 dev](https://cloud.githubusercontent.com/assets/11319966/11668394/41fb10b8-9df7-11e5-904a-fda93730cee5.PNG)

The font is a simple calibri of size 10 with kerning enabled

KonajuGames:
Try expanding the Font Processor properties and setting the Texture Format
to Color. Font textures are compressed by default, which loses some quality.

Tzenchor:
Made the change but no difference

tomspilman:
That looks to be more than just blurry...  that looks to be completely messed up text.

Can you share the font XNB here for people to test?

Tzenchor:
Here they are in this zip. It's contents are the xnb from monogame, the one from xna and the spritefont file used for its generation
https://onedrive.live.com/redir?resid=7C0DE47C61522229!5889&authkey=!APDQX-r56qxONho&ithint=file%2czip
By the way, if I try to attach the file directly it fails, but says that zip files are supported

KonajuGames:
That's fine. Thanks for those. I can take a look tonight my time if no one
else grabs it before then.

tomspilman:
@KonajuGames  great!

KonajuGames:
It's strange.  At sizes of 12 or 14, it produces garbled results.  Anything at 15 or larger and it renders the glyphs correctly.  At first guess, it appears to be something with FreeType, but I'm still investigating.

KonajuGames:
It is indeed a bug in FreeType.
https://github.com/Robmaister/SharpFont/issues/18
[...]

PR #4357, opened 2015-12-23
Fix for small size spritefonts

Tzenchor:
Fixes issue #4305.
Changes:
Convert ImportGlyph method from static method to instance method.
Added method ExpandByte, to convert each bit in a byte to a byte.
Added processing code inside ImportGlyph to handle the case where
glyphs' bitmaps have only 1 bit per pixel

That ExpandByte method is inside SharpfontImporter just to keep it together with the rest of the changes. My opinion is to move it to MathHelper as an extension method, but i'm open to suggestions.


KonajuGames:
Ideally, we would have a `BitmapContent`-derived class that supports 1bpp bitmaps.  Then we would construct a `OneBppBitmapContent` instead of `PixelBitmapContent<byte>` for these cases and everything would still work nicely as well as being able to support importing 1bpp bitmaps for textures (if anyone ever wanted to do that).

For now though, this looks ok.  `ExpandByte` is fine to stay where it is for now.  I don't think it belongs in MathHelpers at least.  There's only one change I would suggest.  Instead of `ExpandByte` allocating a new `byte[8]` each time it is called, allocate one `byte[8]` after the check for it being a mono bitmap and pass that same array in to `ExpandByte` so it can re-use the same array.  Generates far less garbage.  Going further it could expand directly into `gpixelAlphas` using `written` and `stride` as offset and length, avoiding the `byte[8]` completely, but that's probably more complex than it needs to be for no real benefit.

Tzenchor:
Ok, i'll do the changes

Tzenchor:
Done the changes, i have opted for the option of copying directly to the `gPixelAlphas` array. One thing i have done but left commented is doing the checks for input parameters of the new function `ExpandByteAndCopy`

KonajuGames:
Code is either used or not used.  Unused code should be removed.  Commented code shouldn't be left in.  This is an internal method, and as such we can guarantee that we will pass correct values to it.  The exceptions will propagate back up to the user and have no meaning to them.

- Remove `ExpandByte()` as it is no longer used.
- Remove the commented parameter validation code.

Once that is done, I'm happy to merge this.

Tzenchor:
Changes done.

KonajuGames:
Thanks @Tzenchor. Merging.

Oj18:
Is this code now in MonoGame itself?
I still have this issue

Oj18:
Fixed it now, just had to disable anti-aliasing using custom spriteBatch.Begin()
Code:
spriteBatch.Begin(SpriteSortMode.Deferred,
      BlendState.AlphaBlend,
      SamplerState.PointClamp,
      null, null, null, null);

PR #4300, opened 2015-12-07
EffectParameter Additions / Completion

mattleibow:
I have added some changes to the `GetValue` methods to bring them up to the features of XNA. I also added a series of unit tests to ensure that I didn't break anything else.

Mostly, the current methods just catered for a single case, such as `GetValueMatrix`, which always failed if the data wasn't a 4x4 matrix. Now it supports a scalar, 2x3, 3x2, 3x4,4x3 and 4x4. (the 2x2,1x2, 2x1 seem to be missing from the other methods, so I didn't add them, but the can be)

A breaking change to the existing methods is the inclusion of a `int count` parameter to some of the `GetValueArray` methods. This is what XNA does and is a very useful feature. This size controls the size of the resulting array. For some reason, only some of these methods had the count parameter, such as `GetValueMatrixArray`.  
The breaking methods are:
- `GetValueSingleArray`
- `GetValueVector2Array`
- `GetValueVector3Array`
- `GetValueVector4Array`

These changes are the other side of #4268, I didn't touch those methods so that there will be no conflicts when merging.

NOTE: I made a whoopee somewhere and ended up with two PRs :( [#4299] Closed that one.

PR #4361, opened 2015-12-24
Implemented bool, int, quaternion, and vector array SetValue() in EffectParameter.

soenneker:
I decided to squash the commits and create a new pull request.
Previous pull request: #4268 

I've added the invalid cast tests for the various array settings.

@tomspilman:

> We should also test XNA to see what it does when you try to set a singular value on a array type. Does it allow it? If so do we observe the same behavior.

I tested a sample XNA project for this, and it does not allow it. ("Specified cast is not valid.")

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/EffectParameter.cs
+++ b/MonoGame.Framework/Graphics/Effect/EffectParameter.cs
@@ -230,8 +230,10 @@ namespace Microsoft.Xna.Framework.Graphics
                 throw new InvalidCastException();
 
             var ret = new Matrix[count];
-            for (var i = 0; i < count; i++)
-                ret[i] = Elements[i].GetValueMatrix();
+		    for (var i = 0; i < count; i++)
+		    {
+		        ret[i] = Elements[i].GetValueMatrix();
+		    }
 
 		    return ret;
 		}
@@ -274,20 +276,20 @@ namespace Microsoft.Xna.Framework.Graphics
 				}
 				return ret;
 			}
-			
-			switch(ParameterClass) 
-            {
-			case EffectParameterClass.Scalar:
-				return new Single[] { GetValueSingle () };
-            case EffectParameterClass.Vector:
-			case EffectParameterClass.Matrix:
-                    if (Data is Matrix)
-                        return Matrix.ToFloatArray((Matrix)Data);
-                    else
-                        return (float[])Data;
-			default:
-				throw new NotImplementedException();
-			}
+
+		    switch (ParameterClass)
+		    {
+		        case EffectParameterClass.Scalar:
+		            return new Single[] {GetValueSingle()};
+		        case EffectParameterClass.Vector:
+		        case EffectParameterClass.Matrix:
+		            if (Data is Matrix)
+		                return Matrix.ToFloatArray((Matrix) Data);
+		            else
+		                return (float[]) Data;
+		        default:
+		            throw new NotImplementedException();
+		    }
 		}
 
 		public string GetValueString ()
@@ -337,6 +339,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		{
             if (ParameterClass != EffectParameterClass.Vector || ParameterType != EffectParameterType.Single)
                 throw new InvalidCastException();
+
 			if (Elements != null && Elements.Count > 0)
 			{
 				Vector2[] result = new Vector2[Elements.Count];
@@ -422,12 +425,23 @@ namespace Microsoft.Xna.Framework.Graphics
             StateKey = unchecked(NextStateKey++);
 		}
 
-        /*
 		public void SetValue (bool[] value)
 		{
-			throw new NotImplementedException();
+            if (ParameterClass != EffectParameterClass.Scalar || ParameterType != EffectParameterType.Bool)
+                throw new InvalidCastException();
+
+		    for (int i = 0; i < value.Length; i++)
+		    {
+#if DIRECTX
+		        // We store the bool as an integer as that
+		        // is what the constant buffers expect.
+		        ((int[]) Data)[0] = value[i] ? 1 : 0;
+#else
+                // MojoShader encodes even booleans into a float.
+                ((float[])Data)[0] = value[i] ? 1 : 0;
+#endif
+            }
 		}
-        */
 
 		public void SetValue (int value)
 		{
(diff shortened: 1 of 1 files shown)


tomspilman:
@KonajuGames @dellis1972 - Are we good merging this without any unit tests to back it up?

InfiniteProductions:
@Soenneker: I have an issue and maybe you can give me some hints:
I have replaced the modified source file of the PR to my dev version of MG I'm using, it compile fine.

With a project which use _SetValue(int[])_ method, I just got an invalid cast exception because _ParameterType_ is '_single_' instead of '_int32_'.

the caller have a parameter: _EffectParameter textureIndicesParameter_ and call
_textureIndicesParameter.SetValue(textureIndices);_

with textureIndices declared as:
_private readonly int[] textureIndices = new int[MaximumObjectsPerBatch];_

(MG seems to be compiled without DIRECTX symbol by default (??) )
Also, I don't understand the check:
_ParameterType != EffectParameterType.Int32_

Why it is relevant into _setvalue()_ method and not the parameter the method actually get as input (_value _) ?

Thanks.

Soenneker:
Hello @InfiniteProductions, I think I understand the questions you're asking. Please let me know if I'm not.

> With a project which use SetValue(int[]) method, I just got an invalid cast exception because ParameterType is 'single' instead of 'int32'.

> Also, I don't understand the check:
ParameterType != EffectParameterType.Int32

>Why it is relevant into setvalue() method and not the parameter the method actually get as input (_value _) ?

EffectParameters are created off of your shader effect in your content, and the reason we have those checks I believe is to ensure that the EffectParameter instance created matches up with how MonoGame has interpreted your effect. It's easier to track down an issue if an exception being thrown because the type doesn't coordinate, rather than trying to debug why a shader is having problems. I'd take a look at your effect's code (check if your variable you're trying to set is indeed an int array), and then back up to the instance creation of the EffectParameter for your particular effect.

> MG seems to be compiled without DIRECTX symbol by default (??) 

If you're using MonoGame.Framework.Windows.csproj, the conditional compilation symbol "DIRECTX" is true by default.

InfiniteProductions:
You can get it [here](http://bepuphysics.codeplex.com/SourceControl/latest#BEPUphysicsDemos/BEPUphysicsDemosContent/InstancedEffect.fx), it's coming from BEPU source code, trying to make it works with the new MG3.5, this is the only error remaining. Please note that I have modified the shader a bit, but not enough I'm sure, to make it compile with MG by updating PS/VS shader model.
the parameter is defined as:

`float TextureIndices[MAX_OBJECTS];`

For the project, I use protobuild without specifying any platform (which should be current platform, according to the inline help, but it doesn't tell if it has picked DX or OpenGL)

tomspilman:
> For the project, I use protobuild without specifying any platform 

FYI.

Protobuild doesn't alter the projects in any way...  what it generates is consistent regardless of the platform you are on or the parameters passed to it.

So it only really matters what project you used...  if you used the  `MonoGame.Framework.Windows.csproj` then you were using DirectX.  If you used `MonoGame.Framework..WindowsGL.csproj` then you were using OpenGL.

InfiniteProductions:
> Protobuild doesn't alter the projects in any way... what it generates is consistent regardless of the platform you are on or the parameters passed to it.

> So it only really matters what project you used... if you used the MonoGame.Framework.Windows.csproj then you were using DirectX. If you used MonoGame.Framework..WindowsGL.csproj then you were using OpenGL.

I just use protobuild with -generate then -build, but the debugger tells me it use OpenGL (the lines of MG assembly which are used for DX are not enabled), I don't change anything in the projects. And I create a new one just yesterday after updating the source tree.

tomspilman:
> I just use protobuild with -generate then -build

Gosh I hate that feature...  it basically leaves you clueless as to what is being built which is bad IMO.

In theory it should have chosen `MonoGame.Framework.Windows` and not the GL project.  @hach-que ?

Soenneker:
@InfiniteProductions:

>the parameter is defined as:

>`float TextureIndices[MAX_OBJECTS];`

You're going to need to cast your int array to a float array and call:

`public void SetValue(Single[] value)`

if you want to set that in your effect.

InfiniteProductions:
@tomspilman : for the last entries in the console window, I see the project are:
```
Pipeline.Windows.csproj
Monogame.test.Windows.csproj
```

and `MonoGame.Framework.Windows.sln` solution file

Running it again after a clean clearly show `MonoGame.Framework.Windows.csproj` is used, I'll need more info on how to identify which API is used to be more helpful. Is it possible both are actually generated ? (then the latter overwrite the former ?)

hach-que:
@tomspilman `-build` and `-generate` default to the host platform when they aren't given an argument, which can never be WindowsGL.

hach-que:
But yeah, it's always consistent. If no argument is passed, it is the host platform, otherwise it builds the platform specified.

InfiniteProductions:
Just to avoid unnecessary headache, I have investigated a little and opening solution file generated by protobuild shows `DIRECTX` symbol is defined as it should (remember target platform is windows in my case).
But for some reason, the project which use MonoGame, when debugging (with VS2015 WD) shows `DIRECTX` code grayed out, and the `else` part looks like to be enabled, so unless there is actually something wrong here, it might just be an(other) VS issue because when debugging the project, the symbol is actually not defined (VS only consider symbols for the current project and not symbols for the debugged assembly).

Tzenchor:
@InfiniteProductions I also use VS 2015 (but Community edition), if you can point to any place where that does happen I can have a look at it? Maybe you are missing something

InfiniteProductions:
> I also use VS 2015 (but Community edition), if you can point to any place where that does happen I can have a look at it? Maybe you are missing something

I plan to try a simple library with a symbol and two different code (as simple as the classic print "A" print "B") + a project which use the reference and see what happen. I think it's the easiest way or you have to get a MonoGame project + maybe alter one MG source file and get there by a way or another and see what will happen.

MonoGame.Framework/Graphics/Effect/EffectParameter.cs
```diff
@@ -274,20 +276,20 @@ public Single[] GetValueSingleArray ()
...
-				return new Single[] { GetValueSingle () };
-            case EffectParameterClass.Vector:
-			case EffectParameterClass.Matrix:
-                    if (Data is Matrix)
-                        return Matrix.ToFloatArray((Matrix)Data);
-                    else
-                        return (float[])Data;
-			default:
-				throw new NotImplementedException();
-			}
```
aaronfranke: You may wish to split up minor things like indentation changes into its own commit separate from the functional changes.

tomspilman:
Closing this in favor of #4300.

PR #4362, opened 2015-12-27
[MGCB] Clean target handles output path that differs from source path

rtabbara:
Previously, cleaning of content was presupposing that the output content resided in the same location as the source file, which is not always the case.

As an aside, I also needed to update to the latest Dependencies in order to compile the MGCB on the Mac.

### DIFF
--- a/Tools/MGCB/BuildContent.cs
+++ b/Tools/MGCB/BuildContent.cs
@@ -272,7 +272,15 @@ namespace MGCB
                 var cleanOldContent = !inContent && !Incremental;
                 var cleanRebuiltContent = inContent && (Rebuild || Clean);
                 if (cleanRebuiltContent || cleanOldContent || targetChanged)
-                    _manager.CleanContent(sourceFile);                
+                {
+                    // Make sure we specify location of output path in case it doesn't coincide with source path
+                    // PipelineManager will handle searching for output with correct file extension, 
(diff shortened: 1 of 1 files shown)


tomspilman:
I thought that we already handled this in `CleanContent`...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs#L714

There we call `ResolveOutputFilepath` which internally builds the output path:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs#L483

Can you step thru that code and identify why it isn't correctly building the output path in your test case?

rtabbara:
@tomspilman In the instance where no output file is specified, <code>ResolveOutputFilepath</code> tries to construct the output file path by appending the path of the source file (relative to the working directory) to the output directory 

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs#L480-L483

My use case is that I want to be able to specify absolute output/inter dir paths, and clean without necessarily executing the builder in the working directory where the source files reside (there is some method to this madness).

I appreciate that this is probably an atypical use case, so I'm happy to close this off for now, but please reopen if you think it's worth looking at again.

Issue #4271, opened 2015-11-27
[MGCB] Inconsistent output when two files in a row have errors


nkast:
![mgcberror](https://cloud.githubusercontent.com/assets/3018589/11445114/491418a4-9532-11e5-9f0c-caa09723a423.png)

The .mgcb file have 3 files, GeometryA , GeometryB , GeometryC. Files B & C are corrupted.
Notice that between the error messages for B and C, a line is missing with the filename of C.
"P:/TestPipeline/GeometryC.fbx"

As a result, tools like  https://github.com/mono/MonoGame/pull/4168 might group all error lines under the file 'GeometryB.fbx'. 

Actually the GTK tool might be immune to it as it checks for lines starting with '/', haven't tested it since I haven't got a Linux box. But on windows where paths start with a letter I am trying different tricks, like File.Exists(line) as a hint to restart.

cra0zy:
This does cause trouble for Gtk Pipeline as well.

nkast:
Actually MGCB outputs the filename.
The exception message has no newline and the the filename is appended at the end of the prev line.

tomspilman:
Thanks @nkast !

PR #4366, opened 2015-12-30
[MBGC] Add a new line when write errors/exceptions

nkast:
fix for #4271


tomspilman:
Good catch... that should have been a write line all along.

PR #4370, opened 2015-12-31
Updating TargetPlatformVersion to latest SDK (November 2015)

### LINKED ISSUE #4369: Development build and WindowsUniversal project version issues
Creating an issue here and will be submitting a pull request against it shortly as the WindowsUniversal project template isn't quite right.

The problem is that it currently uses 10.0.10240.0 as the version number. This is fine if you have the old SDK installed but the current SDK is 10.0.10586.0. Without the old version installed you cannot create a new Windows 10 project. The dependency versioning is complicated and probably the cause of some issues people are having so here's what I'm seeing:

The 3.4 build uses an old TargetPlatformVersion in the .csproj template as well it's missing the project.json file (the zip file it installs is also named WindowUAP.zip and should be WindowsUniversal.zip)

The 3.5 build uses the 10240 TargetPlatformVersion and has the extra files with the correctly named zip file.

If you go to the SDK archive page and download the RTM version (https://dev.windows.com/en-us/downloads/sdk-archive) it will install version 10.0.10240 of the SDK which works with the 3.5 build (the version displayed on the screen says 10.0.26624 whcih installs to the 10.0.10240.0 folder which works with the 3.5 build). Confused?

You have to download the current SDK (https://dev.windows.com/en-us/downloads/windows-10-sdk) to get version 10586 so I'm going to submit a pull request that sets the minimum version to 10240 and the target version to 10586. This works for people that have the old SDK installed and the new one.

bsimser:
Fixes #4369
 
This sets the target version to the current SDK while maintaining backward compatibility for anyone who only has the RTM version of the SDK installed.

### DIFF
--- a/ProjectTemplates/VisualStudio2015/WindowsUniversal/Application.csproj
+++ b/ProjectTemplates/VisualStudio2015/WindowsUniversal/Application.csproj
@@ -11,7 +11,7 @@
     <AssemblyName>$projectname$</AssemblyName>
     <DefaultLanguage>$currentuiculturename$</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.10240.0</TargetPlatformVersion>
+    <TargetPlatformVersion>10.0.10586.0</TargetPlatformVersion>
     <TargetPlatformMinVersion>10.0.10240.0</TargetPlatformMinVersion>
     <MinimumVisualStudioVersion>14</MinimumVisualStudioVersion>
     <EnableDotNetNativeCompatibleProfile>true</EnableDotNetNativeCompatibleProfile>


tomspilman:
This seems like it should work.  @BrianPeek ?

BrianPeek:
Yup.  This is correct.

tomspilman:
@BrianPeek - Is there some way to say "support any newer version"?  Seems silly that we made zero code changes but had to just update this version number.

BrianPeek:
No way that I know of.  TargetPlatformVersion is the version that the app will compile against, and TargetPlatformMinVersion indicates the lowest version of the SDK that will work.  If both are set to 10240, it means that 10240 needs to be installed by the end user.  If the first is set to 10586 and the second to 10240, it means that it will compile against 10586, but it will ALSO compile against 10240.  There's no magic "work with any SDK, ever".  It's up to the templates to update their versions with each new SDK (this should happen only 1 or 2 times per year), or the user to keep old versions of the SDK installed.

tomspilman:
Thanks @BrianPeek.

Maybe it just feels like it changed  to me because of the pre-release versions.

We will stick to chasing the target version to the latest SDK.

PR #4375, opened 2016-01-02
Multithreaded friendly PVRTexture

nkast:
Depends on https://github.com/KonajuGames/PVRTexLibNET/pull/1

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/DxtBitmapContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/DxtBitmapContent.cs
@@ -106,26 +106,21 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
                     return false;
             }
 
-            try
+            // Create the texture object in the PVR library
+            var sourceData = sourceBitmap.GetPixelData();
+            var rgba32F = (PixelFormat)0x2020202061626772; // static const PixelType PVRStandard32PixelType = PixelType('r', 'g', 'b', 'a', 32, 32, 32, 32);
+            using (var pvrTexture = PVRTexture.CreateTexture(sourceData, (uint)sourceBitmap.Width, (uint)sourceBitmap.Height, 1,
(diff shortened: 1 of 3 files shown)


nkast:
The Test failure is excepted because it depends on a modified PVRTexLibNET.dll 
Future goal is to use all available cores to build content.

Test branch: https://github.com/mono/MonoGame/compare/develop...nkast:tncMultithreadContentPipeline?expand=1

tomspilman:
I'm good with this getting fixed.

> Future goal is to use all available cores to build content.

Yes... this was always my goal with implementing the XNA pipeline in MonoGame.  We want to at some point re-write the core processing loop to work with `Task` or `Parallel.For`.

nkast:
The testing branch was an afternoon play, lots of testing and work to do from here, hopefully with the help from others that know the build pipeline better. I have to say that it was very satisfying to seen the pipeline output scroll down as all my 4 cores were crunching assets!

The PVRTexture currently doesn't cause any problems so I am OK if you have to look into this PR after the 3.5 release.

tomspilman:
> I have to say that it was very satisfying to seen the pipeline output scroll down as all my 4 cores were crunching assets!

We have projects that take 40 minutes to build content single core... so we feel the pain here too.

Most importers/processors/writers should work fine in parallel.  We probably just need to review all the code for statics and look for potential threading problems.  We will need to figure out how to group the logging so that we don't get interleaved build output from multiple files.  Also need to take extra special care about dependent content and be sure it is only built once.

Still totally possible to make work... just needs a little effort behind it.

KonajuGames:
Just added some comments on the related PR.  Easily fixed issues, then I'm good with this.

nkast:
@KonajuGames I assume the changes in this PR are already part of #4496 and coupled with the related submodule update, right? Can I close this?

KonajuGames:
Yes. It no longer uses a global variable for PVRTexLib.

PR #4377, opened 2016-01-03
Change native binaries to be NativeBinary instead of None

hach-que:
This change the references of native binaries to use the `NativeBinary` reference in external projects instead of `None` with copy-on-output.  It's the same result for direct references, but it allows Protobuild to be smarter about adding copy-on-output entries when you reference a library that in turn references native binaries.


KonajuGames:
Merging.

PR #4380, opened 2016-01-03
Add GetMaximumGamePadIndex API

### LINKED ISSUE #4379: Missing public API to query maximum number of game pads
Currently the GamePad API does this internally:

```
if (index < 0 || index >= PlatformGetMaxNumberOfGamePads())
    throw new InvalidOperationException();
```

When you call GetState on an index outside the platform's maximum number of game pads, it throws an exception.  However, there's no public API for getting the value of `PlatformGetMaxNumberOfGamePads` (it is fully private), which means there's no way for a game to enumerate the list of supported gamepads without just cycling up to some number (e.g. 16) and catching exceptions.

hach-que:
Fixes #4379

Adds a `GamePad.GetMaximumGamePadIndex` API so that games can safely enumerate game pads without catching InvalidOperationException

### DIFF
--- a/MonoGame.Framework/Input/GamePad.cs
+++ b/MonoGame.Framework/Input/GamePad.cs
@@ -107,5 +107,15 @@ namespace Microsoft.Xna.Framework.Input
             
             return PlatformSetVibration(index, MathHelper.Clamp(leftMotor, 0.0f, 1.0f), MathHelper.Clamp(rightMotor, 0.0f, 1.0f));
         }
+
+        /// <summary>
+        /// The maximum number of game pads supported on this system.  Attempting to
+        /// access a gamepad index higher than this number will result in an <see cref="InvalidOperationException"/>
+        /// being thrown by the API.
+        /// </summary>
+        public static int MaximumGamePadCount
+        {
+            get { return PlatformGetMaxNumberOfGamePads(); }
+        }
     }
 }


MonoGame.Framework/Input/GamePad.cs
```diff
@@ -107,5 +107,16 @@ public static bool SetVibration(int index, float leftMotor, float rightMotor)
...
+        /// <summary>
+        /// Returns the maximum number of game pads supported on this system.  Attempting to
+        /// access a gamepad index higher than this number will result in an <see cref="InvalidOperationException"/>
+        /// being thrown by the API.
+        /// </summary>
+        /// <returns>The maximum number of game pads supported on this system.</returns>
+        public static int GetMaximumGamePadIndex()
+        {
+            return PlatformGetMaxNumberOfGamePads();
+        }
```
tomspilman: This seems to confuse what the difference in a count and an index is.  Let's just make it clear:
```
public static int MaximumGamePadCount { get; }
```
This is now clearly a count of gamepads like one would expect.

hach-que:
Updated to a property instead

KakCAT:
I think the terminology of PlatformGetMaxIndex and the likes was discarded here 

https://github.com/mono/MonoGame/pull/4273

tomspilman:
> I think the terminology of PlatformGetMaxIndex and the likes was discarded here 

Yep...  this is why I asked to change it in this PR.

tomspilman:
Thanks @hach-que !

PR #4381, opened 2016-01-03
Remove references to framework assemblies

hach-que:
These references are automatically added by Protobuild based on the platform, so adding these references in external projects is unnecessary.  For Mac projects, this is actually detrimental, because the external project specifies MonoMac, even when the project will have a Xamarin.Mac reference added by Protobuild.

This also updates Protobuild, so it fixes those other outstanding issues (like the Windows Universal SDK versioning thing).


tomspilman:
> These references are automatically added by Protobuild based on the platform

Is that something new with this version of Protobuild or something it has done for a while now?

hach-que:
It's been done ever since Protobuild supported these platforms.  Normally it doesn't cause an issue when you have the same reference, but I noticed these because there was an explicit MonoMac reference which was causing issues for Xamarin.Mac (it didn't cause compilation errors, but it did throw off IntelliSense in Xamarin Studio).

hach-que:
Amended, this is ready to merge.

tomspilman:
Thanks @hach-que !

Issue #4252, opened 2015-11-19
[WindowsGL] multiple gamepad issues


bousquetm:
Hi, it seems that there are some issues with the GamePad class on WindowsGL:
- Even though the method IsConnected() returns true, all buttons and all tumbsticks are on their default state (ButtonState.Released & Vector2.Zero).
- If I unplug the pad then restart the game, the buttons and thumbsticks work well
- But the Dpad is still not recognized (all the Dpad buttons are set to ButtonState.Released)
- The Y axis of the thumbsticks points down instead of up
- The PacketNumber always returns 0

I have observed those issues on 2 different PCs (both on Windows 7) with 2 different classic XBox 360 controllers. Also those issues do not occur in a Windows DirectX project.

edit: forgot to mention that it was tested with Monogame 3.5.0.730.

tomspilman:
@cra0zy ?

cra0zy:
What is it returning as the name for the controller?

bousquetm:
How can I get the name of the controller ? I don't see any field 'Name'.

The GamePadCapabilities shows this:
GamePadType	Unknown
HasAButton	true
HasBackButton	true
HasBButton	true
HasBigButton	true
HasDPadDownButton	true
HasDPadLeftButton	true
HasDPadRightButton	true
HasDPadUpButton	true
HasLeftShoulderButton	true
HasLeftStickButton	true
HasLeftTrigger	true
HasLeftVibrationMotor	false
HasLeftXThumbStick	true
HasLeftYThumbStick	true
HasRightShoulderButton	true
HasRightStickButton	true
HasRightTrigger	true
HasRightVibrationMotor	false
HasRightXThumbStick	true
HasRightYThumbStick	true
HasStartButton	true
HasVoiceSupport	false
HasXButton	true
HasYButton	true
IsConnected	true

cra0zy:
Sorry, my bad, forgot to mention how to do it.

Do: OpenTK.Input.GamePad.GetName(index)

bousquetm:
it returns "XInput Controller".

cra0zy:
So, I don't have time to help out with this for the next 3 days, but long story short:

OpenTK uses this database file: https://github.com/opentk/opentk/blob/develop/Source/OpenTK/Input/GamePadConfigurationDatabase.cs

Probably line 78 is causing the issue, but it could also be line 74.

Can't remember where you get GUID from.

@Frassle Any easy gamepad mapping tool for OpenTK?

[...]

PR #4382, opened 2016-01-04
Fixes DPad not being processed in IsButtonDown / Up

mrhelmut:
Fixes one of the issues reported in #4252.

Calling ```GamePadState.IsButtonDown()``` with any DPad buttons wasn't working.

### DIFF
--- a/MonoGame.Framework/Input/GamePad.OpenTK.cs
+++ b/MonoGame.Framework/Input/GamePad.OpenTK.cs
@@ -118,10 +118,6 @@ namespace Microsoft.Xna.Framework.Input
                 (stateTK.Buttons.Start == OpenTK.Input.ButtonState.Pressed ? Buttons.Start : 0) |
                 (stateTK.Buttons.X == OpenTK.Input.ButtonState.Pressed ? Buttons.X : 0) |
                 (stateTK.Buttons.Y == OpenTK.Input.ButtonState.Pressed ? Buttons.Y : 0) |
-                (stateTK.DPad.IsUp ? Buttons.DPadUp : 0) |
-                (stateTK.DPad.IsDown ? Buttons.DPadDown : 0) |
-                (stateTK.DPad.IsLeft ? Buttons.DPadLeft : 0) |
-                (stateTK.DPad.IsRight ? Buttons.DPadRight : 0) |
                 0;
             var buttons = new GamePadButtons(buttonStates);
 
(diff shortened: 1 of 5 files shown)


tomspilman:
Any clue where this was broken or how long this has been like this?  Seems weird that it was missed for this long.

mrhelmut:
Actually, I believe that the XInput implementation works fine because the ```GamePadState.Buttons``` based on ```SharpDX.XInput.State.Gamepad.Buttons``` already includes every buttons' states. It isn't the case on OpenTK. I haven't verified other implementations, but I felt like moving it here would address all at once.

Let me know if you'd like that I move it to OpenTK code instead.

tomspilman:
> I believe that the XInput implementation works fine because the `GamePadState.Buttons` based on `SharpDX.XInput.State.Gamepad.Buttons` already includes every buttons' states.

Would that mean that the state in `GamePadState.Buttons` is incorrect on some platforms?  Some including dpad and others not?

Seems like we should focus on making that consistent across platforms.

cra0zy:
This worked 2~3 months ago, I know cause I was using dpad with a game on Linux. Have no idea why it got broken.

mrhelmut:
> Would that mean that the state in ```GamePadState.Buttons``` is incorrect on some platforms? Some including dpad and others not?

Not really, they are correct, it's just that their internal field ```GamePadButtons.buttons``` contains more states than what is publicly accessible. Some platforms don't set those internal states, or partially, and ```GamePadState.GetVirtualButtons()``` assumes that the internal field ```GamePadButtons.buttons``` has every buttons' states. ```GamePadState.GetVirtualButtons()``` is virtually useless with XInput, but other implementations need it.
OpenTK, Web, UWP & tvOS don't have DPad states and / or virtual ThumbStick states in their internal ```GamePadButtons.buttons```.

There are two approaches:
- setting all states in the internal ```GamePadButtons.buttons``` for every platform in their respective ```PlatformGetState()``` and remove ```GamePadState.GetVirtualButtons()```
- use ```GamePadState.GetVirtualButtons()``` to expose non public fields of ```GamePadState.Buttons```

The latter looks more readable IMO, even though it does useless checks on some platforms.

(Going to squash those last two commits, I messed up a bit.)

mrhelmut:
Arg, I've resubmitted this PR (#4385). I messed up, sorry.

tomspilman:
> it's just that their internal field  GamePadButtons.buttons  contains more states than what is publicly accessible

I see what you are saying here...  `GamePadButtons` doesn't allow access to the dpad state, so it isn't incorrect even though it internally it may include that state depending on the platform.

> The latter looks more readable IMO, even though it does useless checks on some platforms.

Yeah... that is the right direction.  Most importantly it allows the code to live in a common location and not be platform specific.

PR #4385, opened 2016-01-05
Fixes DPad not being processed in IsButtonDown / Up

mrhelmut:
Resubmission of #4382, I've messed up with squashing things...


tomspilman:
Thanks @mrhelmut !

PR #4396, opened 2016-01-06
WindowsDX: Fixed "HasVibrationMotor" properties

### LINKED ISSUE #4395: GamePadCapabilities.HasLeftVibrationMotor / HasRightVibrationMotor  is wrong
Hi,

I am using the latest nuget package from the develop feed 3.5.0967.
My Xbox 360 GamePad has both vibration motors, but the GamePadCapabilities return false on the corresponding HasLeft/RightVibrationMotor properties.
However i can set the vibration for both motors and they work just fine.

Is this already fixed in the current head?

Thanks and appreciation

MichaelDePiazzi:
Fixes #4395 

The `SharpDX.XInput.CapabilityFlags.FfbSupported` (force-feedback) flag is not available in DirectX 11.0. However, it seems sufficient to just check the "motor speed" properties to determine if vibration is supported.

So this should work, but I was only able to test it with XInput gamepads which have vibration motors.

If anyone actually has an XInput controller which doesn't have vibration motors, then it would be awesome if you could test this and confirm that the `HasLeftVibrationMotor` and `HasRightVibrationMotor` properties both return the expected values.

### DIFF
--- a/MonoGame.Framework/Input/GamePad.XInput.cs
+++ b/MonoGame.Framework/Input/GamePad.XInput.cs
@@ -132,8 +132,8 @@ namespace Microsoft.Xna.Framework.Input
             ret.HasLeftVibrationMotor = hasForceFeedback && capabilities.Vibration.LeftMotorSpeed > 0;
             ret.HasRightVibrationMotor = hasForceFeedback && capabilities.Vibration.RightMotorSpeed > 0;
 #else
-            ret.HasLeftVibrationMotor = false;
-            ret.HasRightVibrationMotor = false;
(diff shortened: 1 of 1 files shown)


tomspilman:
@Martin-Schultz - Can you test these changes and make sure it fixes your issue?

Martin-Schultz:
i just tested it on my Windows 10 Machine and the issue seems fixed.

Anyways i don't have a gamepad without FFB either. I'll try to find one until the end of the week,otherwise i would consider this as fixed.

tomspilman:
I'm going to take a chance on this... it can't be more broken than without this.

MichaelDePiazzi:
Yeah, that's true. Thanks Tom.

PR #4397, opened 2016-01-06
Dont use SDL with OpenTK on Windows

cra0zy:
This will affect DesktopGL Windows & Mac

@mrhelmut @dellis1972 Can you test this?

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -80,6 +80,8 @@ using Microsoft.Xna.Framework.Input;
 using OpenTK;
 using OpenTK.Graphics;
 
+using MonoGame.Utilities;
+
 namespace Microsoft.Xna.Framework
 {
     class OpenTKGamePlatform : GamePlatform
@@ -95,6 +97,14 @@ namespace Microsoft.Xna.Framework
 		public OpenTKGamePlatform(Game game)
             : base(game)
         {
+            if (CurrentPlatform.OS == OS.Windows)
+            {
+                Toolkit.Init(new ToolkitOptions
+                    {
+                        Backend = PlatformBackend.PreferNative
+                    });
+            }
+
             _view = new OpenTKGameWindow(game);
             this.Window = _view;
 


mrhelmut:
I'll try it.
@endimilojkoski will most probably have feedback.

mrhelmut:
I have tested. It doesn't work on OS X. Gamepads are not recognized.

cra0zy:
@mrhelmut ok, so I set it Windows only

dellis1972:
I don't think we should do that.. 
At the moment the end user has the option of using SDL as a backend through OpenTK which is required to use DesktopGL on Mac (in my tests anyway) because the code to create a native Cocos window fails. 
So I think we should left OpenTK do what it does and IF the sdl2 dll is present let it use it.

dellis1972:
When I said DekstopGL on Mac I meant with MonoKickstarter rather than Mono normal

dellis1972:
@cra0zy what is the reasoning for doing this?

cra0zy:
#4252

Also look at the comment above, I set it Windows only.

mrhelmut:
@dellis1972 OpenTK gamepad mapping is wrong with SDL (Windows) because it returns different GUID than the native implementation.

@cra0zy oh, very strange, I should have messed up while testing on OS X. I'm fine with this being Windows only.

cra0zy:
> @cra0zy oh, very strange, I must have messed up while testing on OS X. I'm fine with this being Windows only. It removes a dependency and works fine.

It wasn't Windows only when you tested it, I updated it ^^

cra0zy:
Reload the page!!!

mrhelmut:
:+1:

tomspilman:
My two cents...  using OpenTK with SDL has always seemed dumb to me.  If SDL works better why are we using OpenTK at all?

dellis1972:
@tomspilman I would agree SDL seems to be more consistent across platforms (although opentk has improved recently), we had this discussion before when someone tried to add a SDL platform. We would still need a "OpenTK" compatible API for SDL so that we don't end up with #if SDL everywhere in the graphics apis :)

tomspilman:
> We would still need a "OpenTK" compatible API for SDL so that we don't end up with #if SDL everywhere in the graphics apis :)

We put faith in the new OpenTK team to fix things after it went unmaintained for years.  If OpenTK still sucks after 2 years of effort... we really should reconsider dropping it completely.  This has already hurt MonoGame and we don't need to continue the damage.

dellis1972:
>We put faith in the new OpenTK team to fix things after it went unmaintained for years. If OpenTK still
> sucks after 2 years of effort... we really should reconsider dropping it completely. This has already hurt
> MonoGame and we don't need to continue the damage.

Should we pull it out of iOS and Android too? in favour of our own PInvoke API for GL (which would be limited to what we need and nothing more).

tomspilman:
> Should we pull it out of iOS and Android too?

Gosh that old problem.  Maybe... our own pinvoke wouldn't be too hard to maintain.  Do we still have `ref` vs `out` issues?  I know we have enum mappings we do between the different OpenGL bindings.

cra0zy:
Would you like to see OpenTK replaced by SDL2 for DesktopGL?

tomspilman:
> Would you like to see OpenTK replaced by SDL2 for DesktopGL?

Maybe...   it seems even after 2 years we are patching up OpenTK to fix issues with it.  IMO if they can't make it good in 2 years then why would we think it will work in another 2 years?

If we wanted to do this we would do the work in an experimental branch and see what comes out of it.

What do you think @cra0zy ?  @KonajuGames ?

dellis1972:
@tomspilman lets run an experimental branch and see what drops out, this will be the new backend for DesktopGL :) //cc @cra0zy @KonajuGames

cra0zy:
https://github.com/flibitijibibo/FNA-MGHistory/tree/monogame-sdl2

tomspilman:
@cra0zy - We don't want to take the FNA code... we want our own path.

cra0zy:
> @cra0zy - We don't want to take the FNA code... we want our own path.

Just linking it to remind everyone of it's existence.

tomspilman:
@cra0zy @dellis1972 @KonajuGames - Let's start a separate issue for this https://github.com/mono/MonoGame/issues/4399.

As for this PR.... @cra0zy @dellis1972 ...  is it ready to merge?

cra0zy:
It is.

cra0zy:
Closing this since @mrhelmut managed to fix up OpenTK bugs.

mrhelmut:
Waiting for opentk/opentk#335 & opentk/opentk#334
If anybody know who to ping on the opentk side so we can proceed with an update. :)

WardBenjamin:
Pinging @Frassle

dellis1972:
I pinged both Frazzle and the fiddler. Hopefully someone will respond.. It
would be really handy for someone in the MG community to get merge rights
on OpenTK .. alternatively we fork it and pull our own PR's in then submit
them upstream.

PR #4398, opened 2016-01-06
Fix Windows DesktopGL Window... again

cra0zy:
Fixes 2 things:
 - Disables window centering code for Windows 8 and above
   - Windows 8 and above already has window centering
 - Enables back border delay code for Windows
   - turns out if you switch fast enough fullscreen - window - fullscreen - window... you can get borderless window which is not fullscreen...


cra0zy:
I just realized I need to check for if opentk is using sdl or normal, since sdl will do the centering as well... SO INCONSISTENT!!!!

cra0zy:
And turns out there is no way to check if you are using SDL in opentk... I'm just gonna go work on SDL 2 implementation of this :|

PR #4389, opened 2016-01-05
[Gtk Pipeline] Toolbar

cra0zy:
All icons were created by me... that explains the suckiness...

toolbar

PR #4321, opened 2015-12-14
[Gtk Pipeline] Save preferences between restarts

cra0zy:
#4266 got merge conflicts, and I broke it....

Stuff that gets saved:
 - windows size
 - window maximized
 - horizontal and vertical separator positions
 - filter output toggle
 - debug mode toggle

Notes:
 - I didn't save window position since the operating system itself handles that better, ie. multimonitor
 - I used XmlSerialization
 - Gtk already contained "Settings" so I named our settings "PipelineSettings"

PR #4401, opened 2016-01-07
[Win Pipeline] Toolbar and saving settings

cra0zy:
Implements #4389 and #4321 for Windows Pipeline Tool.

winpiptool

### DIFF
--- a/Tools/Pipeline/Windows/MainView.Designer.cs
+++ b/Tools/Pipeline/Windows/MainView.Designer.cs
@@ -38,15 +38,7 @@ namespace MonoGame.Tools.Pipeline
             System.Windows.Forms.ToolStripSeparator _toolStripSeparator3;
             System.Windows.Forms.ToolStripSeparator _toolStripSeparator1;
             System.Windows.Forms.ToolStripSeparator _toolStripSeparator2;
-            System.Windows.Forms.SplitContainer _splitTreeProps;
-            System.Windows.Forms.SplitContainer _splitEditorOutput;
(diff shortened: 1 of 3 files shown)


tomspilman:
Looks great... merging!

Jankamp:
Correct me if I'm wrong, I'm new to monogame:
I think the change on the MainView designer file broke the update of the propertygrid when selecting assets in the treeview. I'm no longer able to update the settings of an fbx because TreeViewAfterSelect doesn't fire, so the propertygrid is always showing the project settings.

I'm talking about the removal of the following line:
```this._treeView.AfterSelect += new 
System.Windows.Forms.TreeViewEventHandler(this.TreeViewAfterSelect);```

![pipelinetool](https://cloud.githubusercontent.com/assets/16737937/12374790/7af4fe1e-bca7-11e5-8846-9a10be2778cc.png)

cra0zy:
> I'm talking about the removal of the following line:
> this._treeView.AfterSelect += new System.Windows.Forms.TreeViewEventHandler(this.TreeViewAfterSelect);

You are right. That shouldn't have gotten removed at all. 'MainView.Designer.cs' file is edited by VS GUI designer and the only thing I did in it was move some components to make room for the toolbar... well from now on I think I will stick with editing stuff by hand, much more precise...

I'll submit a fix tomorrow unless someone beats me to it.

tomspilman:
The sooner the better.

We really need to think about how to setup automated testing for the Pipeline tool.

PR #4404, opened 2016-01-08
Gtk Pipeline, Use Common Filter Output Code and minor bugfix

cra0zy:
Stuff done:
 - Use common code for Gtk Pipline output filtering from #4272
   - Fix Clean / Skip Regex
 - Disable loading setting from file for Gtk Pipeline on menu update... no point, they are already in memory


tomspilman:
Thanks @cra0zy !

Issue #4342, opened 2015-12-20
Allow ContentImporters to accept streams in addition to filenames


hach-que:
In Protogame, I have a custom asset management system which works differently to the way MonoGame models it's content.  However, I still use the internal content pipeline to process assets from their raw formats to the platform-specific binary data.

The issue I'm having is that the content importers accept a filename instead of a stream or byte array.  This means that I'm either forced to copy the content importer implementation into Protogame, or write out the binary data to a temporary file on disk before importing it.

I'm proposing that we add the following signature to `ContentImporter<T>`:

```
public virtual T Import(Stream input, ContentImporterContext context)
{
  // default implementation that writes to a temporary file
}
```

Then for all the MonoGame implementations we can move most of the code into an `override` of this method, and change the `Import` that uses a filename to read the file binary data before calling the stream-based version.  By doing it this way, we ensure that we don't break anyone who is currently deriving from `ContentImporter` (because the method we're adding is virtual, and provides an implementation that is backwards-compatible, if less ideal).

PR #4406, opened 2016-01-10
Add stream-based content import

hach-que:
This adds a stream-based content importing API.  As per #4342, our asset importing system is primarily stream based, because we might be sourcing content from non-file based streams (like over the network).  Right now whenever we import content we have to take the stream (which might already be loaded from a file), write it out to disk somewhere, pass that filename into MonoGame (which might then read a stream again) to perform the content import.

By providing a stream-based content importing API, it removes the requirement to have a temporary file on-disk, and makes importing assets from in-memory or network sources easier.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/ContentImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/ContentImporter.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.IO;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline
 {
@@ -29,6 +30,31 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         /// <param name="context">Contains information for importing a game asset, such as a logger interface.</param>
         /// <returns>Resulting game asset.</returns>
         public abstract T Import(string filename, ContentImporterContext context);
+
+        /// <summary>
+        /// Called by external third-party systems when importing a game asset.  This method is used when the asset is being imported from a stream.
+        /// </summary>
+        /// <param name="input">The stream to read the asset from.</param>
+        /// <param name="virtualFilename">For importers that depend on filenames (such as file extensions), this provides a virtual filename for those importers.</param>
+        /// <param name="context">Contains information for importing a game asset, such as a logger interface.</param>
+        /// <returns>Resulting game asset.</returns>
+        public virtual T Import(Stream input, string virtualFilename, ContentImporterContext context)
+        {
+            var temp = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName() + Path.GetExtension(virtualFilename));
+            try
+            {
+                using (var writer = new FileStream(temp, FileMode.Truncate, FileAccess.Write))
+                {
+                    input.CopyTo(writer);
+                }
+
+                return Import(temp, context);
+            }
+            finally
+            {
+                File.Delete(temp);
+            }
+        }
 
         /// <summary>
         /// Called by the framework when importing a game asset. This is the method called by XNA when an asset is to be imported into an object that can be recognized by the Content Pipeline.
--- a/MonoGame.Framework.Content.Pipeline/EffectImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/EffectImporter.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System.IO;
+using System.Text;
 using Microsoft.Xna.Framework.Content.Pipeline.Graphics;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline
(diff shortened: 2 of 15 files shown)


MonoGame.Framework.Content.Pipeline/H264Importer.cs
```diff
@@ -16,5 +19,10 @@ public override VideoContent Import(string filename, ContentImporterContext cont
             var content = new VideoContent(filename);
             return content;
         }
+
+        public override VideoContent Import(Stream input, string virtualFilename, ContentImporterContext context)
+        {
+            throw new NotSupportedException("Video content must be stored on disk, not imported from a stream.");
```
hach-que: This isn't ideal; I'm assuming MonoGame expects developers to copy out the video files (and audio files) as-is when compiling content, since the importer doesn't actually import the file into any form of memory structure (VideoContent and AudioContent just store the file path).

Alternatively, we could add a `Stream` property to VideoContent and AudioContent so that the processor can read from that if present (or we read the stream and embed it into the XNB; in my case the processed data never makes it into an XNB).

KonajuGames: Incorrect assumption for AudioContent.  Audio content, including compressed, is loaded into memory before exporting in the target format.

MonoGame.Framework.Content.Pipeline/OggImporter.cs
```diff
@@ -24,5 +25,10 @@ public override AudioContent Import(string filename, ContentImporterContext cont
             var content = new AudioContent(filename, AudioFileType.Wav);
             return content;
         }
+
+        public override AudioContent Import(Stream input, string virtualFilename, ContentImporterContext context)
+        {
+            throw new NotSupportedException("OGG content must be stored on disk, not imported from a stream.");
```
KonajuGames: Being pedantic here, but the correct name is "Ogg", not "OGG".  It's not an acronym.

MonoGame.Framework.Content.Pipeline/FbxImporter.cs
```diff
@@ -21,5 +22,15 @@ public override NodeContent Import(string filename, ContentImporterContext conte
 
             return importer.Import(filename, context);
         }
+
+        public override NodeContent Import(Stream input, string virtualFilename, ContentImporterContext context)
```
KonajuGames: It's a public API, so it should have the XML documentation.  Some of the existing public APIs may be missing them, but new public APIs must have them.

MonoGame.Framework.Content.Pipeline/EffectImporter.cs
```diff
@@ -28,9 +29,24 @@ public EffectImporter()
...
         public override EffectContent Import(string filename, ContentImporterContext context)
         {
+            using (var stream = new FileStream(filename, FileMode.Open, FileAccess.Read))
+            {
+                return Import(stream, filename, context);
+            }
+        }
+
+        /// <summary>
+        /// Called by third-party systems when importing an .fx file to be used as a game asset.
```
KonajuGames: Leave out the comment about third-party systems.  It should simply be something like "Imports an effect file from a stream."  The <returns> should be "The effect content."

MonoGame.Framework.Content.Pipeline/ContentImporter.cs
```diff
@@ -31,6 +32,31 @@ protected ContentImporter()
...
+        /// <param name="input">The stream to read the asset from.</param>
+        /// <param name="virtualFilename">For importers that depend on filenames (such as file extensions), this provides a virtual filename for those importers.</param>
+        /// <param name="context">Contains information for importing a game asset, such as a logger interface.</param>
+        /// <returns>Resulting game asset.</returns>
+        public virtual T Import(Stream input, string virtualFilename, ContentImporterContext context)
+        {
+            var temp = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName() + Path.GetExtension(virtualFilename));
+            try
+            {
+                using (var writer = new FileStream(temp, FileMode.Truncate, FileAccess.Write))
```
tomspilman: This seems *really* horrible... writing the stream to disk just to load it again.

tomspilman: @KonajuGames @dellis1972 

I think I would rather see us break compatibility with XNA 4 here rather than have a method like this.

Or maybe implement a second class like `ContentImporter2` or whatever that only has the stream interface for opt-in to the new streaming feature?

KonajuGames: This isn't needed since Import(Stream) is defined for importers in this PR as well.

hach-que: I added for anyone who has custom importers and upgrades MonoGame.

hach-que:
Is there anything blocking this being merged?

nkast:
@hach-que why don't you write a custom ContentManager and override OpenStream()?

something like that :

if(assetName.StartsWith("@Stream:") )
 return _streamProvider.GetStream(assetName);
else
 return base.OpenStream(assetName);

nkast:
The workaround with the tmp file doesn't work 100%. What if an Importer in implemented entirely in `IContentImporter`? The `IContentImporter` is the fundamental type here. `ContentImporter` is the strong-type abstraction of  `IContentImporter`. 

Perhaps we could have a new interface. 
```
    public interface IStreamContentImporter
    {
        Object Import(string filename, ContentImporterContext context, Stream stream);
    }
```

Internally the stock importers could implement `IStreamContentImporter` explicitly. 
The standard file-based Import will open the file and call the stream-based Import.

Importers that are not `IStreamContentImporter` can be handled by your side with the same workaround. Save to tmp file and can the standard `IContentImporter.Import(tmpfile, ...)`

hach-que:
@tomspilman What do you think about adding another interface?  I know you're generally opposed to new interfaces...

(I particularly think it's fine if people need to implement the new method upon upgrading since I anticipate most people using the concrete class; I didn't even know about `IContentImporter` until now).

MonoGame.Framework.Content.Pipeline/H264Importer.cs
KonajuGames: It seems I was incorrect in calling it incorrect, or it has changed since that time.  Only WAV content is loaded into memory.  Compressed audio just stores the file path.

KonajuGames:
Adding new interfaces is generally not a good thing in the runtime portion where efficiency is paramount.

If that cleans up the temporary copying of a stream to file and back, I'd be in favour of it.  You should also address the requests above about the documentation in the PR.

Jjagg:
+1 for the interface solution. It's a million times better than writing a temp file and there's no performance or functional impact for the existing importers.

hach-que:
We still use this heavily in Protogame, but I'm going to do a rewrite of the asset system in the near future, which will give me a chance to move this over to the `IStreamContentImporter` suggestion.

PR #4407, opened 2016-01-10
Tidying OpenAL constants

### LINKED ISSUE #4403: OpenAL and maximum playing instances
Hello,

It looks like the maximum of playing instances on all OpenAL platforms is set to 32 (```OpenALSoundController.MAX_NUMBER_OF_SOURCES```).

However, this limit is not considered in ```SoundEffect``` which has its own ```SoundEffect.MAX_PLAYING_INSTANCES``` set to ```int.MaxValue``` (DesktopGL) or ```256``` (XamMac).

There is yet another constant in ```SoundEffectInstancePool.MAX_PLAYING_INSTANCES```, but this one is never used. The ```SoundEffectInstancePool``` size is ```SoundEffect.MAX_PLAYING_INSTANCES```.

Fix proposal:
- set ```SoundEffect.MAX_PLAYING_INSTANCES``` to ```OpenALSoundController.MAX_NUMBER_OF_SOURCES``` on all OpenAL platforms and cleaning the duplicated / unused values;

P.S.: unless explicitly documented, it is advised to set ```OpenALSoundController.MAX_NUMBER_OF_SOURCES``` to 32 (most modern hardware do 64, but for compatibility it is safer to stick to 32, and there is no way to query this limit by code).

P.S. 2: ```MediaPlayer``` uses 1 unchecked source too, which means that the limit of 32 is wrong if the real hardware limit is 32 too.

mrhelmut:
Fixes #4403

```SoundEffectInstancePool``` unused constants have been removed, and ```SoundEffect``` constants are now based upon ```OpenALSoundController``` constants (minus 1, because NVorbis uses a sound source and this should be subtracted).

The maximum of playing instances for DesktopGL has been set to 256, beause MacOS & Linux share this limit (even though Windows doesn't have a limit).

### DIFF
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -36,7 +36,27 @@ namespace Microsoft.Xna.Framework.Audio
 		//int[] buffers;
         private AlcError _lastOpenALError;
         private int[] allSourcesArray;
-        private const int MAX_NUMBER_OF_SOURCES = 32;
+#if DESKTOPGL || ANGLE
+
+        // MacOS & Linux shares a limit of 256.
+        internal const int MAX_NUMBER_OF_SOURCES = 256;
+
+#elif MONOMAC
+
+        // Reference: http://stackoverflow.com/questions/3894044/maximum-number-of-openal-sound-buffers-on-iphone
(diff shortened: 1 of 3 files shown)


tomspilman:
This seems reasonable to me.  @dellis1972 @KonajuGames ?

mrhelmut:
Closing this PR in favor of #4426

PR #4419, opened 2016-01-13
SpriteBatch performance optimisation

MichaelDePiazzi:
This is an optimisation for the common case of drawing sprites without any rotation.

I benchmarked my optimised `SpriteBatchItem.Set` method against the normal version and found it to be roughly 3 times faster.

My top-down 2D game uses a mix of unrotated sprites (e.g. tiles) and rotated sprites (e.g. characters and objects), and this optimisation gives it a fairly respectable frame rate increase (around 10%).

Obviously your results may vary though depending on the number of sprites being drawn and how many of those are rotated. But I believe this should be a worthwhile optimisation in general for most users.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -402,19 +402,33 @@ namespace Microsoft.Xna.Framework.Graphics
 				_texCoordTL.X = temp;
 			}
 
-			item.Set (destinationRectangle.X,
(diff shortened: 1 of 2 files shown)


tomspilman:
@nkast - This look good to you?

nkast:
Yes, looks great.

tomspilman:
Thanks @MichaelDePiazzi !

MichaelDePiazzi:
:thumbsup:

PR #4420, opened 2016-01-13
Moved Stock Effect Resources

tomspilman:
This PR moves the resource loading for the stock effects to a partial class that can be overloaded for new platforms.


tomspilman:
@KonajuGames @dellis1972 

The only downside I see to this implementation is that because of static initialization it now loads all the stock effect resources at once.  Before it would only load the resource when the effect in question was first used.  This means up to 125K of extra memory usage.

The alternate implementation is one partial file per-Effect class which is a little more verbose.

Do you guys think I should refactor this?

dellis1972:
@tomspilman is there no way to keep the partial.. but have them lazy loaded when the corresponding effect is created?

tomspilman:
> is there no way to keep the partial.. but have them lazy loaded when the corresponding effect is created?

I'm experimenting with that now.

First thought is that it would require the static variable to be replaced with a static method.  So one more function call on each creation of an `Effect`.

Another crazy idea is nested classes to wrap the statics...  but that looks messy to me.

tomspilman:
Well if we want to dip into .NET 4.0 features we can use the new `System.Lazy<>` type:

 https://msdn.microsoft.com/en-us/library/dd642331(v=vs.110).aspx

```
static public readonly Lazy<byte[]> AlphaTestEffect = new Lazy<byte[]>(() => LoadResource("Microsoft.Xna.Framework.Graphics.Effect.Resources.AlphaTestEffect.dx11.mgfxo"));
```

You then use `AlphaTestEffect.Value` to access the byte array which will load it on first demand.

dellis1972:
@tomspilman that is exactly what I was thinking :)

tomspilman:
Closing this for a clean PR... one day I will learn to correctly rebase. :/

PR #4421, opened 2016-01-13
Moved Stock Effect Resources

tomspilman:
This PR moves the resource loading for the stock effects to a partial class that can be overloaded for new platforms.


tomspilman:
@dellis1972 - Wrote something a bit more custom here instead of `Lazy<T>`, but gives us a similar result...  the effect bytecode isn't loaded until it is first accessed.

KonajuGames:
That sound like a good solution to me.

tomspilman:
@KonajuGames @dellis1972 - I'm done with this one.

KonajuGames:
Merging.

PR #4425, opened 2016-01-14
GraphicsDeviceManager Refactor

### LINKED ISSUE #2859: Refactor GraphicsDeviceManager
We need to apply the same platform abstractions as we have in `VertexBuffer`, `GraphicsDevice`, or `SoundEffectInstance` to the `GraphicsDeviceManager`.  

This would leave us with one core GraphicsDeviceManager.cs for the common bits and separate partial classes for each platform/api that requires differences.

These changes need to take special care not to break existing functionality.

tomspilman:
This PR is an attempt to clean and split up GraphicsDeviceManager to allow for partial class extension on new platforms.

 - Copied existing `GraphicsDeviceManager` to `GraphicsDeviceManager.Legacy.cs` to maintain behavior of old platforms.
 - New version of `GraphicsDeviceManager.cs` without any platform specific code in it.
 - The Windows, Windows8, Windows Phone, Windows Phone 8.1, Windows Universal, and Web platforms now use the new `GraphicsDeviceManager`.
 - Added some more reference docs.
 - Minor cleanups in Windows platform files.

Fixes #2859 
Fixes https://github.com/mono/MonoGame/issues/1762.

Once we switch over the OpenTK backed to support this then also https://github.com/mono/MonoGame/issues/1748 should be fixed if it isn't already.


tomspilman:
I just put this up to test my progress.

I still need to figure out how to abstract out `Initialize()` `ApplyChanges()`, `ToggleFullScreen()`, and `IsFullScreen`.

tomspilman:
@dellis1972 @KonajuGames 

I'm thinking of taking the mess of code in these methods and just putting them in some ".Legacy.cs" partial file.  We can then structure a new path for new platforms and slowly bring the existing platforms over.

Any other ideas?

KonajuGames:
A temporary Legacy.cs sounds like an OK plan to move forward.

dellis1972:
Ditto Legacy.cs sounds like the most reasonable approach. then we can move the code out slowly as things change.

tomspilman:
So sticking to the legacy file concept here, but I won't do any complex partial thing.  Instead we have two copies of `GraphicsDeviceManager `:

 - `GraphicsDeviceManager.cs` - The new code with partial method implementations for platform specific bits.
 - `GraphicsDeviceManager.Legacy.cs` - The old code for platforms that haven't been fixed to work with the new code.

Should have a new push here shortly with these changes.

Tzenchor:
i'm having a look at making an abstract device manager with the common parts and a concrete class with the specifics for each platform.
Form what i have done until now, it seems viable, and can remove every conditional compilation block, but needs some reworking to avoid duplicated code

tomspilman:
> i'm having a look at making an abstract device manager with the common parts and a concrete class with the specifics for each platform.

We preferably don't want to make abstract `GraphicsDeviceManager` class as it alters the class hierarchy from XNA.

The path I've taken here allows for `GraphicsDeviceManager` to remain identical to XNA with platform specific code in one of three places:

 - A partial class of `GraphicsDeviceManager`.
 - The `GraphicsDevice`.
 - The `GamePlatform`.

Things are looking pretty good this way.

tomspilman:
@dellis1972 @KonajuGames - I could use feedback on this now.

dellis1972:
Looks ok to me. @KonajuGames ?

tomspilman:
Note this can wait until after 3.5.

tomspilman:
Going to merge this.  It only affects Windows platforms at the moment.

PR #4410, opened 2016-01-11
Fixes OpenAL crashing upon exiting on MacOS

mrhelmut:
This fixes an exit error discussed in #4153 where ```Alc.DestroyContext()``` could be called twice and raise an error on MacOS.

@dellis1972 You may be interested in testing this.

PR #4426, opened 2016-01-14
Fixes OpenAL buffers and sources management

### LINKED ISSUE #4403: OpenAL and maximum playing instances

### LINKED ISSUE #4411: [OpenAL] Song playback crashes after a while
After about 20 minutes of playing different ```Song```, we end up having this error:

```
System.InvalidOperationException: Invalid Name
at MonoGame.Utilities.ALHelper.Check () [0x00000] in <filename unknown>:0
at MonoGame.Utilities.OggStreamer.EnsureBuffersFilled () [0x00000] in <filename unknown>:0
at System.Threading.Thread.StartInternal () [0x00000] in <filename unknown>:0
```

I don't know yet on which AL call it failed, and this can be related to MacOS only.

May be related to #4413

### LINKED ISSUE #4413: [OpenAL] Buffer limit ? OS X crash at 16
In #4403 we discussed about the maximum number of playing sources. It is stated that OS X has a limit of 256 sources and 1024 buffers.

However, I'm getting a lot of AL errors on this line: https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OALSoundBuffer.cs#L34

When debugging, there are 16 pooled ```SoundEffectInstance``` in the ```SoundEffetInstancePool``` and 0 playing instances.

This makes me question the 1024 buffers and 256 sources limits.

What is strange, is that the current source limit of 32 seems fine (```OpenALSoundController.MAX_NUMBER_OF_SOURCES```). I already played up to 32 sources simultaneously. But the buffer limit looks to be 16... Could this be possible?

I also noticed that while being pooled, ```SoundEffectInstance``` retain their ```OALSoundBuffer``` and create a new ```OALSoundBuffer``` when removed from the pool without clearing / reusing the previous one. Could this be the cause? But the number of 16 buffers still seem pretty low.

@cra0zy @dellis1972 Any idea?

mrhelmut:
While investigating #4413, I have noticed that OpenAL buffers were wrongly managed: every ```SoundEffectInstance.Play()``` a new buffer is created without freeing the previous one from the OpenAL context, and buffers are bound to one source at a time (buffer-source relationship is forced to be 1-1, instead of 1-multiple).

There is two problems with this: buffers are not reused, and they are not freed, which is causing #4413 to happen (the buffer limit is reached, and the number of 16 I mentioned is obviously wrong and was pure coincidence).

This PR does the following:
- attaching a buffer to ```SoundEffect``` instead of ```SoundEffectInstance``` so they can be reused and don't depend on one single source (works like XAudio2)
- removing source attachment from ```OALSoundBuffer``` so they are independent and only bound to a playing ```SoundEffectInstance```
- correctly disposing the buffers upon ```SoundEffect.Dispose()```

Fixes #4413
Fixes #4411
Fixes #4403 

While reviewing this PR, #4407 & #4410 should help getting rid of the remaining MacOS problems. (I'll most probably have to rebase them -- @tomspilman would you prefer that I push all these OpenAL changes in this present PR ?)

We've tested that extensively and it seems to work just fine. No leaks or runtime allocations to declare.
Some good testing from other developers would be welcome (inc. on mobile). @dellis1972 @cra0zy @endimilojkoski

UPDATE:

I have updated this PR to make ```OggStream``` use the ```OpenALSoundController``` source pool. This will make sure that there is no conflict between ```Song``` and ```SoundEffectInstance```. The source limit will now correctly be respected and the correct exception will be thrown if need be. This makes ```OpenALSoundController``` to be the one and only place where OpenAL sources are managed.

### DIFF
--- a/MonoGame.Framework/Audio/OALSoundBuffer.cs
+++ b/MonoGame.Framework/Audio/OALSoundBuffer.cs
@@ -18,9 +18,7 @@ namespace Microsoft.Xna.Framework.Audio
 		ALFormat openALFormat;
 		int dataSize;
 		int sampleRate;
-		private int _sourceId;
         bool _isDisposed;
-        internal int _pauseCount;
 
 		public OALSoundBuffer ()
 		{
@@ -91,20 +89,6 @@ namespace Microsoft.Xna.Framework.Audio
 
         }
 
-        public void Pause()
-        {
-            if (_pauseCount == 0)
-                AL.SourcePause(_sourceId);
-            ++_pauseCount;
-        }
-
-        public void Resume()
-        {
-            --_pauseCount;
-            if (_pauseCount == 0)
-                AL.SourcePlay(_sourceId);
-        }
-
 		public void Dispose()
 		{
             Dispose(true);
@@ -128,31 +112,6 @@ namespace Microsoft.Xna.Framework.Audio
                 _isDisposed = true;
             }
         }
-
-		public int SourceId
-		{
-			get {
-				return _sourceId;
-			}
-
-			set {
-				_sourceId = value;
-				if (Reserved != null)
-					Reserved(this, EventArgs.Empty);
-
-			}
-		}
-
-		public void RecycleSoundBuffer()
-		{
-			if (Recycled != null)
-				Recycled(this, EventArgs.Empty);
-		}
-
-		#region Events
-		public event EventHandler<EventArgs> Reserved;
-		public event EventHandler<EventArgs> Recycled;
-		#endregion
 	}
 }
 
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -36,7 +36,27 @@ namespace Microsoft.Xna.Framework.Audio
 		//int[] buffers;
         private AlcError _lastOpenALError;
         private int[] allSourcesArray;
-        private const int MAX_NUMBER_OF_SOURCES = 32;
+#if DESKTOPGL || ANGLE
+
+        // MacOS & Linux shares a limit of 256.
+        internal const int MAX_NUMBER_OF_SOURCES = 256;
+
+#elif MONOMAC
+
+        // Reference: http://stackoverflow.com/questions/3894044/maximum-number-of-openal-sound-buffers-on-iphone
+        internal const int MAX_NUMBER_OF_SOURCES = 256;
+
+#elif IOS
+
+        // Reference: http://stackoverflow.com/questions/3894044/maximum-number-of-openal-sound-buffers-on-iphone
+        internal const int MAX_NUMBER_OF_SOURCES = 32;
+
+#elif ANDROID
+
+        // Set to the same as OpenAL on iOS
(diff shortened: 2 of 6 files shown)


endimilojkoski:
I can confirm that SoundEffect memory leak is a huge issue that I encountered a while ago.
Running the game with sound effects would fill up memory without releasing it.

I was just starting with MonoGame and to avoid the memory leak problem I have created SoundEffectMemoryInstances as much as I need and use only those without creating new. So far this works for me pretty good but will definitely try this fix and let you know how it goes soon.

tomspilman:
> will definitely try this fix and let you know how it goes soon.

Thanks @endimilojkoski .... the sooner the better.

mrhelmut:
I have update this PR to make ```OggStream``` use the ```OpenALSoundController``` source pool. This will make sure that there is no conflict between ```Song``` and ```SoundEffectInstance```. The source limit will now correctly be respected and the correct exception will be thrown if need be. This makes ```OpenALSoundController``` to be the one and only place where OpenAL sources are managed.

I've moved #4407 here, because the source limit constant is tightly bound to this PR to work correctly.

And I've also moved #4410 here as well to have to get everything cleaned up.

dellis1972:
@mrhelmut I'm going to try this on a game I'm porting tonight and let you know how it goes.

dellis1972:
@mrhelmut fixes the issues on mac perfectly. it shuts down nicely now and still plays sound and music without issues.  I'm going to merge in the next hour unless there are any objections :D

dellis1972:
@mrhelmut  I think this broke OpenAL on Android :( SoundEffect.play does nothing.. music plays ok though. which is weird. @KonajuGames any chance you can test and see if you can repo?

mrhelmut:
```Song``` doesn't use OpenAL on Android, I wonder what I did wrong to break it. :/
If it doesn't crash (I would have expected ```ALHelper.Check()``` to expose problems) and that no ```SoundEffect``` are working, I guess that it failed to load / bind AL buffers.
Are there any "Failed to get buffer bits" in your logs?

Issue #4399, opened 2016-01-06
Replacing OpenTK


tomspilman:
This stems from a discussion started in another PR:

https://github.com/mono/MonoGame/pull/4397#issuecomment-169345797

It has been over 3 years since we initially investigated sticking with OpenTK in MonoGame.  The OpenTK team at that time promised fixes and renewed development of the library.

Since then there have been improvements, but we still suffer problems with OpenTK functionality and need per-platform work arounds at times.

Is it finally time to give up on OpenTK?

Do we proceed with our own custom binding for SDL2 and OpenGL?

What happens to platforms like iOS and Android where we depend on the Xamarin fork of OpenTK for OpenGL access?

hach-que:
Is it not possible to P/Invoke native libraries on iOS / Android? I was under the impression that worked.

If it does, we could just P/Invoke OpenGL there too.

tomspilman:
> Is it not possible to P/Invoke native libraries on iOS / Android? 

I think it is.

This would be a new path for the project.  Where in the past we depended on bindings provided by other open source projects or vendors like Xamarin.... we would be maintaining our own bindings.

It might be more work, but it puts us more in control over our destiny on these platforms.

Not this is sort of the same boat we are in on consoles where we generate our own bindings to native console APIs.

dellis1972:
it is possible to PInvoke on both Android and iOS, I can certainly help out on the Android side.

tomspilman:
Note that on iOS we may eventually want to use the new Metal API which I think is not available to C# in any way at the moment.  So there might be another binding to generate there at a later date.

hach-que:
It also puts the project in a position where we could drop the dependency on Xamarin for iOS / Android if an alternative way of running IL on those platforms arose.

tomspilman:
@hach-que - That would be possible too I guess.  Who knows where Microsoft's open source .NET stuff will lead in the coming years.

tomspilman:
So here is something to consider.

On consoles the styles of APIs are all over the place... some are very simple C APIs, some are handle based, some use lots of macros, some take structures with dozens of parameters, and some are actual C++ class libraries with inlined functions and everything.  These all pose different challenges in writing bindings.

Some of the APIs we use from MonoGame and game code are called infrequently so any cost of making the C# to native call transition is cheap.  However some need to be called 100s of times a frame... typically rendering calls.  This can make the call overhead to bound APIs costly.

So what we have done with MonoGame on consoles:

[...]

PR #4428, opened 2016-01-14
Replace OpenTK with SDL 2 for Joystick, Gamepad, GamePlatform, GameWindow, Mouse and Keyboard

cra0zy:
Notes:
 - this does brake Angle platform


tomspilman:
Looks like you missed a file:

```
Cannot copy MonoGame.Framework/Desktop/MonoGame.Framework.dll.config to MonoGame.Framework/bin/Linux/AnyCPU/Release/MonoGame.Framework.config, as the source file doesn't exist.
```

cra0zy:
Ok, this breaks Mac as well... will put Mac Joystick to Default so it can compile until OpenTK replacement is finished.

cra0zy:
Few more notes:
 - added Gamepad here since I got it done before this got merged
 - vibrations are finally working \o/ (OpenTK didn't have them implemented)
   - I was planing on adding just rumble effect but thanks so me forgetting the SDL_INIT_HAPTIC flag I ended up going over @flibitijibibo old monogame-sdl code trying to figure out what I did wrong... and there I noticed how to setup LeftRight effect... 
 - This also adds sdl2 gamepad database file from: https://github.com/gabomdq/SDL_GameControllerDB

cra0zy:
More notes:
 - Updates GameWindow and GamePlatform to SDL2 backend
 - Mouse and Keyboard input is no longer working, I still have to update those

cra0zy:
And EVEN MORE MORE notes:
 - in order to calculate window titlebar size and border I had to make mouse start in center of window, hopefully that is ok

tomspilman:
> in order to calculate window titlebar and border size I had to make mouse start in center of window

Why is that?  Are we already running into problems with the SDL backend?

MonoGame.Framework/Game.cs
```diff
@@ -19,6 +19,10 @@ namespace Microsoft.Xna.Framework
 {
     public class Game : IDisposable
     {
+#if DESKTOPGL
+        public static bool Resizable = false;
```
tomspilman: First let's not add more platform specific code into `Game`.

Second isn't this what `GameWindow.AllowUserResizing` is for?

MonoGame.Framework/SDL/SDLGameWindow.cs
```diff
@@ -0,0 +1,197 @@
...
+                int x = 0, y = 0, w, h;
+
+                SDL.SDL_GetWindowSize(Handle, out w, out h);
+
+                if (!isFullScreen)
+                {
+                    SDL.SDL_GetWindowPosition(Handle, out x, out y);
+
+                    x += BorderX;
+                    y += BorderY;
```
tomspilman: I suspect this border calculation is unnecessary.

The docs say that `SDL_GetWindowSize()` return the size the client area... not the window size.

Also this post confirms it...

http://stackoverflow.com/a/31996739

Are you certain this is needed?

cra0zy: I need border in order to return correct position, not size.

MonoGame.Framework/Input/Mouse.cs
```diff
@@ -139,20 +94,21 @@ public static MouseState GetState(GameWindow window)
...
-            var pc = ((OpenTKGameWindow)window).Window.PointToClient(new System.Drawing.Point(state.X, state.Y));
-            window.MouseState.X = pc.X;
-            window.MouseState.Y = pc.Y;
+            int x, y;
+
+            var state = SDL.SDL_GetGlobalMouseState(out x, out y);
+            var pos = PrimaryWindow.Position;
+
+            window.MouseState.X = x - pos.X - BorderX;
+            window.MouseState.Y = y - pos.Y - BorderY;
```
cra0zy: @tomspilman This is another position where I needed border size, the reason why it's needed here is so we can track global mouse.

We can track mouse while it's inside window with `SDL.SDL_GetMouseState`, but as soon as it gets out it won't be tracked.

MonoGame.Framework/Game.cs
cra0zy: SDL does not have a way of setting window resizable parameter after the window has already been created.

cra0zy:
Ok, so I decided to use `GetMouseState` for now, until I can setup a proper border detection code.

We are gonna set it up per platform :(

MonoGame.Framework/Game.cs
tomspilman: > SDL does not have a way of setting window resizable parameter after the window has already been created.

We will have to find a way around that.  At least on startup we could delay creation of the SDL window until `Game.Initialize()` which currently when the window is actually shown.

But changing that after startup... including making the window borderless...  is all stuff we need to support.

Regardless of all that we won't be adding a `Game.Resizable` property.

cra0zy: > We will have to find a way around that. At least on startup we could delay creation of the SDL window until Game.Initialize() which currently when the window is actually shown.
> 
> But changing that after startup... including making the window borderless... is all stuff we need to support.
> 
> Regardless of all that we won't be adding a Game.Resizable property.

Window creation is tied to a valid GL window in: https://github.com/cra0zy/MonoGame/blob/sdl2/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs#L100 I though doing this was more sensible for the moment.

Window can already be set borderless without problem, it's only Resizable property that is not setable after creation.

tomspilman: > was more sensible for the moment.

Maybe for the moment... but we definitely don't want to leave window creation in `GraphicsDevice`.

> it's only Resizable property that is not setable after creation.

We can maybe make due without that for now, but we should talk to the SDL team about it.  If OpenTK supported it I don't see why SDL couldn't.

cra0zy: > We can maybe make due without that for now, but we should talk to the SDL team about it. If OpenTK supported it I don't see why SDL couldn't.

OpenTK does not support it with SDL backend, that is why we kept SDL disabled for Linux in OpenTK.dll.config

> Maybe for the moment... but we definitely don't want to leave window creation in GraphicsDevice.

Window is not getting created in GraphicDevice, it's just that we need Window handle to initialize some GL stuff...

Anyway I can move window creation code in there and it should theoretically allow what you wanted with resizable parameter, do you want me to do that?

tomspilman: > OpenTK does not support it with SDL backend,

I'm taking about OpenTK ... not OpenTK with SDL.  Did OpenTK support changing the window to resizable after creation?

cra0zy: > I'm taking about OpenTK ... not OpenTK with SDL. Did/does OpenTK support changing the window to resizable after creation?

You don't get how opentk functions, you can't ask "does OpenTK support this" since it depends on the backend it runs with:
1) OpenTK has 2 backends for each platform, native and SDL
2) During initialization it will pick one (this can be forced within code)
3) For resizable, If it picked SDL it won't support it, if it picked native it will support it

All an all OpenTK can't help SDL with this. 

We can't use OpenTK to make SDL window resizable.

PS. SDL 1.2 did support changing resizable parameter after window creation.

tomspilman: > We can't use OpenTK to make SDL window resizable.

You have completely misunderstood me.

Obviously SDL can't be fixed by OpenTK...   I was just trying to figure out if the problem existed in OpenTK (with native backend).

> PS. SDL 1.2 did support changing resizable parameter after window creation.

So that settles it...  SDL2 lost this feature and it is totally possible to change the resize window state on Mac and Linux platforms after window creation.

This was all I was trying to figure out.

cra0zy:
@tomspilman I think I did what you wanted with the last commit.

cra0zy:
This is ready for merging.

WardBenjamin:
No it's not! Fix in a few minutes.

tomspilman:
Note I plan to be a lot more flexible with merges to this `replace_opentk` branch as I expect it to be all very experimental until we decide to merge it back to develop.

WardBenjamin:
@tomspilman @cra0zy  check out https://github.com/cra0zy/MonoGame/pull/1. Simple typo :P

cra0zy:
> @tomspilman @cra0zy check out cra0zy#1. Simple typo :P

So you made a PR for another PR instead of just warning me about the typo? #logic

Anyway I rebased and fixed the original commit.

amerkoleci:
I think would be great to distribuite also native SDL2 dlls for Windows, Linux and OSX.
Note that SDL 2.0.4 has support for Windows10 Universal app too, so probably can be used there too.

cra0zy:
> I think would be great to distribuite also native SDL2 dlls for Windows, Linux and OSX.

I agree with you on that one, but that update will come in a later PR that modifies project templates. If you are interested about general discussion on this topic look at #4399

> Note that SDL 2.0.4 has support for Windows10 Universal app too, so probably can be used there too.

Didn't know that, tho I wouldn't use SDL for it since we do get a higher level of control with the existing method. Plus it's much more complex using SDL on non Desktop stuff.

tomspilman:
> I wouldn't use SDL for it since we do get a higher level of control with the existing method.

Yeah... Win10 works great with existing MS APIs.  We really only need SDL for the GL desktop platforms.

cra0zy:
@tomspilman Can this be merged? I want to start working on some other parts of this(probably audio).

MonoGame.Framework/SDL/SDL2.cs
```diff
@@ -0,0 +1,405 @@
...
+        public static extern Button SDL_GetGlobalMouseState(out int x, out int y);
+
+        [DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SDL_GetMouseState")]
+        public static extern Button SDL_GetMouseState(out int x, out int y);
+
+        [DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SDL_ShowCursor")]
+        public static extern int SDL_ShowCursor(int toggle);
+
+        [DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SDL_WarpMouseInWindow")]
+        public static extern void SDL_WarpMouseInWindow(IntPtr window, int x, int y);
```
tomspilman: Looks like you missed removing the `SDL_` prefix on these imports?

cra0zy: Yup, fixed.

tomspilman:
Seems good to merge after the build finishes.  Since it is only being merged to the `replace_opentk` branch it won't break anything for the 3.5 release and is pretty low risk right now.

tomspilman:
Merged @cra0zy !

PR #4432, opened 2016-01-17
[Win Pipeline] Fix treeview events

cra0zy:
VS GUI editor decided to randomly remove all events from treeview in #4401

I went over the entire file, only treeview events were removed.


tomspilman:
Thanks @cra0zy !

PR #4436, opened 2016-01-18
Fix for OpenTK messing with the window size when both fullscreen and …

mrhelmut:
…resolution are changed.

I have noticed that if one tries to change ```GraphicsDeviceManager.IsFullscreen``` AND ```GraphicsDeviceManager.PreferredBackBufferWidth``` / Height through ```GraphicsDeviceManager.ApplyChanges()``` outside of the ```Game``` constructor, the new window size is totally wrong (the new size is basically ignored and OpenTK use a default value instead).

I have identified the cause to be OpenTK ```NativeWindow.WindowState``` to recalculate the window size when its state changes from ```Hidden``` to anything else. This overrides the setting of ```NativeWindow.ClientBounds```.

This PR simply invert the order in which ```NativeWindow.ClientBounds``` and ```NativeWindow.WindowState``` are set so that ```NativeWindow.ClientBounds``` keeps its intended value.

@cra0zy @dellis1972

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
@@ -292,16 +292,6 @@ namespace Microsoft.Xna.Framework
                     window.WindowBorder = WindowBorder.Resizable;
                 
                 updateClientBounds = false;
-
-                // when we resize, we also have to move the window to center of the screen here again
-                // to do this, we simply compare the old size to the target size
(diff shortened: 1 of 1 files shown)


cra0zy:
Please read #4398

I really got irritated with opentk...

mrhelmut:
Damn... didn't tested that against SDL2.

I think I'll just stick to the old fashioned "please restart to apply settings" for now...

cra0zy:
> I think I'll just stick to the old fashioned "please restart to apply settings" for now...

lol

PR #4438, opened 2016-01-18
ShaderProfile Refactor

tomspilman:
This PR is the first baby step towards making the content pipeline more easily extendable for new platforms.

Here I refactored the existing `enum ShaderProfile` into a class that knows how to perform various platform specific steps of the effect compiling process.

The next step here is to remove specific values like `ShaderProfile.OpenGL` and `ShaderProfile.DirectX_11' and allow for new profiles to be discovered via attribute.  This then opens up the possibility of loading extension DLLs for supporting new platforms.


tomspilman:
@dellis1972 @KonajuGames @hach-que @cra0zy 

The changes I made to `ShaderProfile` here is something I want to try to apply to `TargetPlatform`.  The main reason is to allow for new closed platforms (like PS4, XB1, PSVita, WiiU, etc) to be able to extend the tools without needing to fork and modify the code.

If done right...  in the end we can have a single install of MonoGame tools on a system instead of having multiple copies for different console platforms.

hach-que:
That sounds good to me.

cra0zy:
I like it. If I am not mistaken this also helps with getting 2MGFX on Linux and Mac, right?

tomspilman:
> If I am not mistaken this also helps with getting 2MGFX on Linux and Mac, right?

Not really...  there are bigger issues to deal with there.  In particular this is the next major step that someone has to take on:

https://github.com/mono/MonoGame/issues/2167#issuecomment-91461062

tomspilman:
Note this can wait until after 3.5.

tomspilman:
Did some refinement of this PR to make console work easier.

tomspilman:
Thought about this for a while and I think I need a slightly different approach... specifically with `ShaderProfile` and the `Enumeration<>` type.  The reason with the concern being that I want to apply the same technique to the existing `TargetPlatform` enumeration and want to make it work right.

I think trying to "fake" an enum like this is a bad idea.  It will never work correctly in 100% of the cases... especially considering if one of the values is defined in a separate assembly from the rest (something critical for extendibility).

  - Partial classes don't work across assembly boundaries.
  - We can't depend on it being castable to a simple integer.
  - Things depend on reflection (aka property grids or command line parsers) need special cases.

So my next refactor here will give up on the fake enum approach... which might end up more of a conceptual and type naming issue more than anything else.  Instead I want to focus on supporting that these can come from separate assemblies from the start and find a common way to discover and collect them.

In the case of `enum TargetPlatform` specifically I think we will leave the old type in place, but the old public methods that use them will just forward to a new method that takes a new extenable platform identifier type.

tomspilman:
Closed in favor of https://github.com/mono/MonoGame/pull/4992.

Issue #4265, opened 2015-11-25
SpriteFont.MeasureString(text) gives incorrect height


monobeginner:
Hey,

I've found bug with method MeasureString(text). I try to set text into center of rectangle but with some fonts it will be never centred, because MeasureString() gives incorrect height. Bug noticed with many fonts, for example
* Comix Loud - http://www.dafont.com/comix-loud.font
* Luckiest Guy - http://www.1001freefonts.com/luckiest_guy.font

Verified with stable 3.4 release and newest server build #3.5.0.826. Please take a look on attached screenshot.
![measurestringissue](https://cloud.githubusercontent.com/assets/16008535/11385338/0fc68c38-9318-11e5-9c78-8d3bb3f46088.png)

KonajuGames:
This is one of those bugs that gets fixed, then it's quiet for a while,
then the same bug is reported but with different fonts. One thing I have
learned is to never assume the data in a TTF is good, especially in fonts
from the free font sites.

Do you know if XNA gives the same or different results?

monobeginner:
https://www.google.com/fonts/specimen/Luckiest+Guy
It is Google web font, so theoretically TTF should be OK. Can I expect it to be fixed ?

tomspilman:
> [This](https://onedrive.live.com/redir?resid=7C0DE47C61522229!5977&authkey=!ANk48NnE23gN0s0&ithint=folder%2czip) is what i could get from my project without taking the Calibri or Trebuchet ones. Concretely it is a Segoe UI Mono font. Included in the zip are: the spritefont, the font, "XSegoeUIMono.xnb" is the XNA version, and "MSegoeUIMono.xnb" is the Monogame version.

> Also, just for comparison, i did a gif file comparing my project with XNA fonts and the Monogame ones
![font comparison](https://cloud.githubusercontent.com/assets/11319966/12433500/b11e2824-bf01-11e5-9fc6-9b5e6cb724f8.gif)

Tzenchor:
I've been tinkering with the pipeline, and the text is being drawn relative to the top of the area instead of centered.
I'll put some numbers to explain this conclusion.
While using the font "MS Trebuchet" with size 48, i get a line spacing of 74, with a maximum ascender of 60 (the height of the font above the text baseline), so each glyph is drawn at `maximum ascender - glyph ascender`, this seems correct, but because no character has that height (at least with the chars I import, but I use the spanish charset that includes 'Ñ' for example), the lower part of the area is empty.

[...]

PR #4439, opened 2016-01-19
SpriteFont.MeasureString fix

MichaelDePiazzi:
MonoGame's `SpriteFont.MeasureString` was in some cases returning a larger width than XNA was.

I've added some new tests which expose this issue, then fixed the problem in MonoGame.


KonajuGames:
I would like there to be a few fonts in the test.  It has been fixed to work with this font or that font several times over the past few years.  A range of fonts in the test would provide better coverage.

MichaelDePiazzi:
@KonajuGames No problem. Do you possibly have any XNA compiled fonts on hand? (I don't have XNA Studio installed)

MichaelDePiazzi:
Odd... the `RasterizerStateTest.VisualTestCullMode(CullCounterClockwiseFace)` test failed... No idea why though as I didn't touch anything related to that...??

Tzenchor:
@MichaelDePiazzi i have some fonts i used for a XNA project, i'll post them later, when i get on the other computer.
Basically the fonts i used were: Calibri and MS Trebuchet

MichaelDePiazzi:
Awesome, thanks @Tzenchor! I assume that's there's no licensing issues with using these fonts that we'd need to be concerned about?

KonajuGames:
> Odd... the RasterizerStateTest.VisualTestCullMode(CullCounterClockwiseFace) test failed... No idea why though as I didn't touch anything related to that...??

There are sometimes random failures in some graphics tests unrelated to the changes.  Doing another test run usually passes.

> Basically the fonts i used were: Calibri and MS Trebuchet

We cannot distribute them, even in bitmap form.  It violates the font licensing.  The XNA Redistributable Font Pack is available [here](http://xbox.create.msdn.com/en-US/education/catalog/utility/font_pack) which should be usable.

MichaelDePiazzi:
> There are sometimes random failures in some graphics tests unrelated to the changes. Doing another test run usually passes.

Ah ok, cool.

> The XNA Redistributable Font Pack is available here which should be usable.

If someone can supply me with some of these fonts compiled with XNA (i.e. as XNB files), then I'll update the tests to include them. I'd compile them myself, but I don't have XNA Studio installed...

tomspilman:
> We cannot distribute them, even in bitmap form. 

I think we're going to have to find a way to bend the rules in this case.  Like we have two fonts in https://github.com/mono/MonoGame/issues/4265 that we know are bugged...  not being able to include their XNB form for regression testing is only hurting MonoGame.

Maybe we can somehow watermark the resulting font image to keep them from being useful?

> If someone can supply me with some of these fonts compiled with XNA 

Sure...  can you tell me which ones and what settings you want on them?

MonoGame.Framework/Graphics/SpriteFont.cs
```diff
@@ -267,7 +267,7 @@ private void MeasureString(ref CharacterSource text, out Vector2 size)
                 }
                 hasCurrentGlyph = true;
 
-                var proposedWidth = offset.X + currentGlyph.WidthIncludingBearings + Spacing;
+                var proposedWidth = offset.X + currentGlyph.WidthIncludingBearings;
```
tomspilman: We should be sure to include test cases of fonts with and without `Spacing` to be certain of the results.

Tzenchor:
[This](https://onedrive.live.com/redir?resid=7C0DE47C61522229!5977&authkey=!ANk48NnE23gN0s0&ithint=folder%2czip) is what i could get from my project without taking the Calibri or Trebuchet ones. Concretely it is a Segoe UI Mono font. Included in the zip are: the spritefont, the font, "XSegoeUIMono.xnb" is the XNA version, and "MSegoeUIMono.xnb" is the Monogame version.

Also, just for comparison, i did a gif file comparing my project with XNA fonts and the Monogame ones
font comparison

MichaelDePiazzi:
> Sure... can you tell me which ones and what settings you want on them?

@tomspilman Thanks! I've created some .spritefont files using the fonts from that XNA pack. You can grab them from [here](http://plus7software.com/downloads/TestSpriteFonts.zip). If you (or anyone else) can compile these for me with XNA, that'd be awesome!

I've tried to select some of the more unusual fonts from the pack in the hopes that they'll cause more problems than a typical font. I've also tried to use some different spacing values as you've suggested. This is what I've selected (but feel free to change/add to this):
* Jing Jing, Size=20, Spacing=0
* Lindsey, Size=24, Spacing=4
* Motorwerk, Size=26, Spacing=0
* Quartz MS, Size=22, Spacing=1
* Segoe Keycaps, Size=10, Spacing=2

> I think we're going to have to find a way to bend the rules in this case. Like we have two fonts in #4265 that we know are bugged... not being able to include their XNB form for regression testing is only hurting MonoGame.

The "[Luckiest Guy](https://www.google.com/fonts/specimen/Luckiest+Guy)" font mentioned in that issue is licensed under Apache 2.0. Wouldn't that be fine to use with MonoGame? Not too sure about the other one though ([Comix Loud](http://www.dafont.com/comix-loud.font))... the license only says "free for personal use", but I'm sure the author would be fine with it as we'd be using it non-commercially.

> This is what i could get from my project without taking the Calibri or Trebuchet ones. Concretely it is a Segoe UI Mono font.

@Tzenchor Thanks, but I'm fairly sure that the "Default" font already included in the test project is actually Segoe UI Mono. Also, you've compiled this font under the "HiDef" profile and I'm not sure if that will work on the build server.

> Also, just for comparison, i did a gif file comparing my project with XNA fonts and the Monogame ones

This could possibly be differences between how the XNA and MonoGame content pipelines are rendering the fonts. But let's get these additional fonts added to the tests first. This will help to prove that the way the framework is drawing fonts is at least ok.

tomspilman:
@MichaelDePiazzi - Here are those fonts:

https://app.box.com/s/jku3yokjoove0qrn0xyqlcjmhtw2gzcj

Maybe once this issue is resolved we can look at the "Luckiest Guy" issue too.

MichaelDePiazzi:
> Here are those fonts

Thanks Tom! With these additional fonts, I managed to discover some more failing test cases. So I've pushed a couple more fixes, and also took the opportunity to simplify the code a bit and make it a little easier to read. I'm now very confident that MeasureString is behaving the same as XNA.

> Maybe once this issue is resolved we can look at the "Luckiest Guy" issue too.

After working on this, I have a feeling that this other issue is actually related to the content pipeline (which I'm not familiar with at all). But if you can compile the relevant fonts for me, then I can try and take a look at it when I have some spare time.

tomspilman:
> I'm now very confident that MeasureString is behaving the same as XNA.

This is fantastic!

> So I've pushed a couple more fixes, and also took the opportunity to simplify the code a bit and make it a little easier to read.

So did you check if any of these fixes also apply to drawing the text?

MichaelDePiazzi:
> So did you check if any of these fixes also apply to drawing the text?

Sorry, forgot to mention that. Yes, I did. Here's what I found:

* The main issue related to calculating the final width which didn't apply to DrawString.
* The secondary issue related to calculating the X offset as it iterated through characters. But this logic in DrawString was actually correct, so I'm guessing MeasureString was overlooked when some fixes to this were previously done. So to fix it, I just copied the logic from DrawString.
* The last edge-case issue was related to handling of the '\r' character which did also apply to DrawString, so I fixed it there too.

tomspilman:
Great... let's merge this thing!

@Tzenchor - You might want to check these fixes in your test case.

MichaelDePiazzi:
Cool :thumbsup:

Tzenchor:
There's no difference in text placement with this changes compared to Monogame version 3.4

tomspilman:
@Tzenchor - Your issue is likely all in the font processor then if no improvement was seen.  Let's take this discussion back to https://github.com/mono/MonoGame/issues/4265.

PR #4442, opened 2016-01-19
Prevent OpenAL calls with 0 buffers or empty buffers

### LINKED ISSUE #4440: [OpenAL] Occasionnal OggStreamer crash
I've only seen this happening on OS X, but given the nature of the problem, I'm pretty sure it can happen on any OS depending on the decoding performance.

A crash can occur in ```OggStreamer.FillBuffer()``` on the ```AL.BufferData()``` call with an ```Invalid Value``` AL error.

AL documentation about this error:
> The size parameter is not valid for the format specified, the buffer is in use, or the data is a NULL pointer. 

I haven't managed to reproduce it in a debug environment yet, but I'm pretty certain that ```size``` and ```format``` are always correct, and that ```data``` can't be NULL because ```CastBuffer()``` would throw an exception before the ```ALHelper.Check()```. Hence, it most probably mean that a buffer is still in use, though it shouldn't if ```AL.SourceUnqueueBuffers()``` returned it...

UPDATE: got it! I'm testing it before opening a PR.

FYI, it is a discrepancy of OpenAL across platforms. ```AL.BufferData()``` generates this error if the ```size``` parameter is ```0``` on OS X, while it does nothing on Windows. Same applies to ```AL.SourceQueueBuffers()```.

mrhelmut:
This fixes an error related to differences across OpenAL implementations.
>```AL.BufferData()``` generates an "invalid value" error if the ```size``` parameter is ```0``` on OS X, while it does nothing on Windows. Same applies to ```AL.SourceQueueBuffers()```.

I've simply added checks so that it doesn't crash ```Song``` anymore.

Fixes #4440

This however raises another issue: this shouldn't be happening. The original cause is ```VorbisReader.ReadSamples()``` occasionally failing to read the ogg stream by returning ```0``` and skipping a few samples or simply reset the stream (it loses track of the stream position). Which means that there is a bug in NVorbis, or something that Mono don't like in NVorbis.

This means that, even though it doesn't crash anymore, it may occasionally skip a few seconds of a ```Song```, replay a previous buffer, or reset it at worse. It also seems to break the stream initial position, make ```Song``` to start at random position after the first occurrence of this issue.

@dellis1972 Do you find acceptable to merge this until I can figure out what's wrong with NVorbis?

Even though the original ```OggStreamer``` may be a couple years old now, maybe @renaudbedard could have some thoughts about this (did you experienced similar NVorbis issues on Mono?).

### DIFF
--- a/MonoGame.Framework/Utilities/OggStream.cs
+++ b/MonoGame.Framework/Utilities/OggStream.cs
@@ -43,7 +43,7 @@ namespace MonoGame.Utilities
         internal readonly int[] alBufferIds;
 
         readonly int alFilterId;
-        readonly Stream underlyingStream;
+        readonly string oggFileName;
 
         internal VorbisReader Reader { get; private set; }
         internal bool Ready { get; private set; }
@@ -52,9 +52,9 @@ namespace MonoGame.Utilities
         public Action FinishedAction { get; private set; }
         public int BufferCount { get; private set; }
 
-        public OggStream(string filename, Action finishedAction = null, int bufferCount = DefaultBufferCount) : this(File.OpenRead(filename), finishedAction, bufferCount) { }
-        public OggStream(Stream stream, Action finishedAction = null, int bufferCount = DefaultBufferCount)
+        public OggStream(string filename, Action finishedAction = null, int bufferCount = DefaultBufferCount)
         {
+            oggFileName = filename;
             FinishedAction = finishedAction;
             BufferCount = bufferCount;
 
@@ -76,8 +76,6 @@ namespace MonoGame.Utilities
                 ALHelper.Efx.Filter(alFilterId, EfxFilterf.LowpassGain, 1);
                 LowPassHFGain = 1;
             }
-
-            underlyingStream = stream;
         }
 
         public void Prepare()
@@ -234,8 +232,6 @@ namespace MonoGame.Utilities
                     Empty();
 
                 Close();
-
-                underlyingStream.Dispose();
             }
 
             AL.Source(alSourceId, ALSourcei.Buffer, 0);
@@ -290,8 +286,7 @@ namespace MonoGame.Utilities
 
         internal void Open(bool precache = false)
         {
-            underlyingStream.Seek(0, SeekOrigin.Begin);
-            Reader = new VorbisReader(underlyingStream, false);
+            Reader = new VorbisReader(oggFileName);
 
             if (precache)
             {
@@ -400,20 +395,36 @@ namespace MonoGame.Utilities
                 return streams.Remove(stream);
         }
 
-        public bool FillBuffer(OggStream stream, int bufferId)
+        public int FillBuffer(OggStream stream, int bufferId)
         {
             int readSamples;
+            long readerPosition = 0;
             lock (readMutex)
             {
+                readerPosition = stream.Reader.DecodedPosition;
                 readSamples = stream.Reader.ReadSamples(readSampleBuffer, 0, BufferSize);
                 CastBuffer(readSampleBuffer, castBuffer, readSamples);
             }
             if (readSamples > 0)
+            {
                 AL.BufferData(bufferId, stream.Reader.Channels == 1 ? ALFormat.Mono16 : ALFormat.Stereo16, castBuffer,
-                    readSamples * sizeof (short), stream.Reader.SampleRate);
-            ALHelper.Check();
+                    readSamples * sizeof(short), stream.Reader.SampleRate);
+                ALHelper.Check();
+            }
+            else
+            {
+                // this shouldn't be happening
+                // if ReadSamples() returns 0, it means that NVorbis failed (this looks to be related to Mono, and has been reported to happen on OS X)
 
-            return readSamples != BufferSize;
+                // let's try to reset the stream
+                stream.Close();
+                stream.Open();
+                stream.Reader.DecodedPosition = readerPosition;
+
+                return 2;
+            }            
+
+            return (readSamples != BufferSize ? 1 : 0);
         }
         static void CastBuffer(float[] inBuffer, short[] outBuffer, int length)
         {
@@ -444,7 +455,7 @@ namespace MonoGame.Utilities
                             if (!streams.Contains(stream))
                                 continue;
 
-                        bool finished = false;
+                        int finished = 0;
 
                         int queued;
                         AL.GetSource(stream.alSourceId, ALGetSourcei.BuffersQueued, out queued);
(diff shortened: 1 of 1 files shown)


mrhelmut:
This PR has still a couple issues I'd like to try fixing. I can maybe limit the side effects.

renaudbedard:
I stopped using NVorbis in favor of a native code solution (VorbisFile), so I can't speak for problems with its usage with Mono. Sorry!

mrhelmut:
Thanks for dropping by, you confirmed what I've been wondering by looking at the FNA implementation.

dellis1972:
I have no problem merging this @KonajuGames ? 
Ideally fixing NVorbis would be nice as it means we don't have to drop to using native libs.

mrhelmut:
I'd like to try to enhance that a bit before it gets merged. I'll try to check if I can manage to save the streamer state and restore it if it fails at some point.

dellis1972:
No problem 

-----Original Message-----

KonajuGames:
Avoiding the native libs would be ideal, but if we can't find a fix for
NVorbis we may have to drop back to native libs.

cra0zy:
For after 3.5, since we are replacing OpenTK, and we also need to replace OpenAL, I think we might as well start using sdl_mixer for song again.

mrhelmut:
This is getting weird...

If I play a ```Song```, stop it, dispose it, reload it and play it again, it starts where the previously disposed one stopped.  o_O

I don't know what's this about, but it only happens on Mono, and it seems to totally wreak the file stream while ```VorbisReader``` assume it to be in sync from the given starting position. This may be why ```VorbisReader``` fails (it tries to read behind the stream size).

I really don't know what to do about this, even though I dispose everything, ```GC.Collect()```, and all streams are correctly recreated and set to 0 position, it starts the file stream where one of a previous ```Song``` stopped.

dellis1972:
Is the Bug in the MG Song class not resetting the progress in the stream?  or is this pure NVorbis?
Can you provide a repo project?

mrhelmut:
I don't even know if it comes from MG or NVorbis. Everything seems normal in both, but after a few minutes of gameplay (which involves disposing and loading new ```Song```), ```MediaPlayer.Play()``` (on a freshly loaded ```Song```) starts to play music at random position.

It's like the internal ```File.Stream``` to the ogg file is totally wrong, even though it is correctly set to 0 in ```OggStream.Open()```.
This fix and trying to reset the stream if NVorbis fails doesn't help at all, once the bug is here, recreating everything doesn't change anything.

I'm starting to think that it is a bug in Mono.

I'm going to give it one last try by letting NVorbis manage itself the stream lifecycle instead of ```OggStream```.

mrhelmut:
No luck...

I pushed my latest attempt. What it does:
- lets ```VorbisReader``` manage the life cycle of the ```File.Stream```
- detects if ```VorbisReader.ReadSamples()``` failed to read
- tries to dispose ```VorbisReader``` and recreate it at the same position

What seems to happen:

Context: in our game, we load ```Song``` on demand and dispose them right after they're no more used. This means that every ```Song``` played is a fresh ```OggStream``` / ```VorbisReader```.

Everything works just fine for a few minutes. At some point, a new ```Song``` is loaded and played. But it starts at a random position.

Starting from here, every new ```Song``` starts at random positions, even though ```VorbisReader``` has its internal stream and values correctly set to 0.

The side effect of this shift is that ```VorbisReader``` does its normal job, and tries to read behind where the stream really is, and this results in ```VorbisReader.ReadSamples()``` returning 0 samples.

Creating a new ```VorbisReader``` doesn't help because the shift problem is still present, and restoring its latest position only set it to where it failed and end up with a "position out of stream" exception.

dellis1972:
@mrhelmut is the Content Caching getting in the way here? We do keep a reference to resources loaded via the Content.Load method.. perhaps thats not being removed correctly when you dispose. (it should be)

mrhelmut:
Tracking ```Song.Position``` and...

```
Before OggStream.Prepare() = 00:00:00
After OggStream.Prepare() = 00:00:00.5000000
After ContentManager.Load() = 00:00:00.5000000
Before MediaPlayer.Play() = 00:00:14.5000000
Before MediaPlayer.PlaySong() = 00:00:14.5000000
Before MediaPlayer.PlatformPlaySong() = 00:00:14.5000000
Before OggStream.Play() = 00:00:14.5000000
After OggStream.Play() = 00:00:14.5000000
```

This means that the underlying ```VorbisReader```, or possibly the ```OggStream``` is modified between ```ContentManager.Load<Song>()``` and ```MediaPlayer.Play()``` !

I think it is related to the ```OggStreamer.EnsureBuffersFilled()``` thread and something going wrong with how playing streams are managed with ```HashSet<OggStream> streams```.

tomspilman:
> modified between `ContentManager.Load<Song>()` and `MediaPlayer.Play()`

That is always something that has worried me...  it is too easy to create a content type that is modified during usage thus changing it state for someone that later receives that same cached instance from `ContentManager`.

I would closely look at the state stored in `Song`... it should not contain state mutated by playback.

mrhelmut:
I don't think it's the ```ContentManager```, bugged instances are freshly created (not cached), and every states are reset / recreated.

Forcing ```VorbisReading``` position to 0 before playing it internally fixes this, but that doesn't tell us what's wrong... and worse, the stream position seems to be jumping while playing as well... and sometimes it makes ```OggStreamer``` think that the song has ended...

There are strange things everywhere, the more I dig into this, the more I'm lost...

tomspilman:
> There are strange things everywhere, the more I dig into this, the more I'm lost...

When I have problems like this I take a step back and rethink things.

Usually your head ends up full of all sorts of wild theories... things that are highly unlikely to actually be occurring.  Chasing those down is a waste of time.  

Go back and look at the simple things you dismissed early on.  It is more likely it is something obvious and simple that you missed than some an insidious compiler bug.

mrhelmut:
You're totally right, I'll slow down. :)


I found why ```Song``` starts with a different position. It's OpenAL having different behaviors on OS X.

```AL.GetSource(stream.alSourceId, ALGetSourcei.BuffersQueued, out queued);``` doesn't reset the number of processed buffers if the source is stopped. If there's at least one buffer processed before ```AL.CloseSource()``` is called, it keeps returning the same count even though it has been processed.

This makes ```OggStreamer``` to think that it should continue filling buffers. Hence, the ```Song``` is being read without being played.

Trying to figure out a cross-platform fix now... but the ```VorbisReader.ReadSamples()``` error looks to be another different issue.

mrhelmut:
Got it!

I couldn't directly workaround the OpenAL differences, so I did the following:
- make sure that only playing ```Song``` are active on the ```OggStreamer``` (newly loaded ```Song``` were queued even if not played, and the OpenAL problem made them to stream in the streamer thread while the source is stopped), this fixes the stream position issue;
- unqueue all buffers of a ```Song``` when it is stopped (this should help OpenAL to better count processed buffers);

With this problem solved, the ```VorbisReader``` reset works, which means that the initial problem is solved too. ```VorbisReader``` still very occasionally fails, but it now resets correctly and it can't be heard at all.

@dellis1972 could you test if everything is fine on your side?

dellis1972:
@mrhelmut will do and if all ok I'll merge it :)

mrhelmut:
Any news?

I would like this PR to be merged before enhancing the AL error checks.

dellis1972:
@mrhelmut I would merge it but thats not the done thing (merging your own PR's ;) ). 

@KonajuGames  ?

KonajuGames:
If it works better than before, that's a good thing.  I would like to eventually know the root cause of this, but this seems a good workaround.

dellis1972:
opps, I thought this my PR. do need to read the email properly next time :)

mrhelmut:
Thanks. That's why I'd like to enhance the error handling, I'm still working on making this the cleanest possible.

Issue #4431, opened 2016-01-17
Reporting issues that are happening in Raining Blobs


endimilojkoski:
Hello guys,

here is a list of issues that happened after the Steam release:
1. Game was not working on Arch Linux with MonoGame mono kickstart.
   I got with the player and solved the issue. Since Arch Linux have newer libraries this was the issue:
   libgif.so.4 => not found
   libpng12.so.0 => not found
   And on Arch Linux he had 
   libgif.so.7 and libpng16
   He added the older libs and the game started working!
2. After making it work on Arch Linux, full screen was not working, only very small part of the full screen was showing and the rest was transparent, you could see the desktop. When going in windowed mode it was fine until he reached 1920x1080 window mode resolution.
   Also when playing in windowed mode it is not a real window but a black full screen with the game of the windowed mode resolution size.
3. I have complaint from 2nd player that the game would not start at all on Windows with ATI GPU, still have to investigate this. A while ago one friend could not start the game at all on different ATI GPU PC.
4. Game is not working properly in full screen on 4K monitors and on Mac retina displays. The background is black and the game shows in top left 1/4 of the screen. Switching to windowed mode it works fine. I have added window mode resolutions up to 4K and it works with no issues.

Hope you find this helpful and find a quick solution @cra0zy @mrhelmut @tomspilman 
I will update you as I find more issues or solutions.

Thanks,
Endi

cra0zy:
> Game was not working on Arch Linux with MonoGame mono kickstart.
> I got with the player and solved the issue. Since Arch Linux have newer libraries this was the issue:
> libgif.so.4 => not found
> libpng12.so.0 => not found
> And on Arch Linux he had 
> libgif.so.7 and libpng16
> He added the older libs and the game started working!

Most gdi+ libs for Linux: https://github.com/MonoGame/MonoKickstart/pull/4

@mrhelmut We are missing libpng12.so.0

> After making it work on Arch Linux, full screen was not working, only very small part of the full screen was showing and the rest was transparent, you could see the desktop. When going in windowed mode it was fine until he reached 1920x1080 window mode resolution.
> Also when playing in windowed mode it is not a real window but a black full screen with the game of the windowed mode resolution size.

Ask about what graphics driver and what Desktop Enviorment he is using.

> I have complaint from 2nd player that the game would not start at all on ATI GPU, still have to investigate this. A while ago one friend could not start the game at all on different ATI GPU PC.

Windows, Linux or Mac?

[...]

PR #4449, opened 2016-01-21
Disabling HiDPI for Retina displays

mrhelmut:
Following discussions in #4431, this PR disables OpenTK HiDPI mode.

Making MonoGame HiDPI-compliant with OpenTK doesn't look to be interesting considering that it will be replaced, so this is a quick and safe workaround.

@dellis1972 @cra0zy what do you think?

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -95,6 +95,10 @@ namespace Microsoft.Xna.Framework
 		public OpenTKGamePlatform(Game game)
             : base(game)
         {
+            Toolkit.Init(new ToolkitOptions
+            {
+                EnableHighResolution = false
+            });
             _view = new OpenTKGameWindow(game);
             this.Window = _view;
 


dellis1972:
@mrhelmut so disabling it means it works on HiDPI better?

mrhelmut:
Problem with HiDPI is that OpenTK or SDL don't use point coordinate but pixels, so when a retina resolution is selected, all rendering is shrunk by the % amount of the increased DPI. This parameter tells OpenTK to create a GL viewport with pixel coordinate instead of point.

mrhelmut:
This doesn't fix everything. :/
OpenTK tries to handle the DPI, but the math doesn't look right, and disabling DPI scaling obviously doesn't fix that. It affects other OS too, that's why 4K on Windows also behave strangely.

I'm closing this for now.

endimilojkoski:
Yes, I can confirm that this does nothing on Windows 7 with 4K monitor and Mac with Retina display, just tried both.

mrhelmut:
I have investigated the problem the whole day. OpenTK handles HiDPI correctly for both Windows & Mac, but there are multiple mechanism in MonoGame that overrides the Viewport and assume it should be the same size of the backbuffer.
I'm having a hard time debugging this, since I don't have a retina display or a 4k monitor to test on.

PR #4450, opened 2016-01-21
[Pipeline] Multithreaded build

nkast:
I had some time today to work on it, anyone welcome to test, review, etc

Tested mostly with soundEffects. Textures and related assets (models, fonts) will fail at the moment  because of PVRTexture.

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -391,6 +391,8 @@ namespace MonoGame.Tools.Pipeline
         public void Build(bool rebuild)
         {
             var commands = string.Format("/@:\"{0}\" {1}", _project.OriginalPath, rebuild ? "/rebuild" : string.Empty);
+            if (PipelineSettings.Default.MultithreadMode)
+                commands += " /multithreadMode";
             BuildCommand(commands);
         }
 
@@ -444,6 +446,8 @@ namespace MonoGame.Tools.Pipeline
 
             // Run the build the command.
             var commands = string.Format("/@:\"{0}\" /rebuild /incremental", tempPath);
+            if (PipelineSettings.Default.MultithreadMode)
+                commands += " /multithreadMode";
             BuildCommand(commands);
 
             // Cleanup the temp file once we're done.
@@ -476,6 +480,8 @@ namespace MonoGame.Tools.Pipeline
             View.OutputClear();
 
             var commands = string.Format("/clean /intermediateDir:\"{0}\" /outputDir:\"{1}\"", _project.IntermediateDir, _project.OutputDir);
+            if (PipelineSettings.Default.MultithreadMode)
+                commands += " /multithreadMode";
 
             _buildTask = Task.Factory.StartNew(() => DoBuild(commands));
             _buildTask.ContinueWith((e) => View.Invoke(UpdateMenu));
--- a/Tools/Pipeline/MainWindow.cs
+++ b/Tools/Pipeline/MainWindow.cs
@@ -94,6 +94,9 @@ namespace MonoGame.Tools.Pipeline
         {
             PipelineController.Instance.OnProjectLoaded += () => projectControl.ExpandBase();
 
+            cmdMultithreadMode.Checked = PipelineSettings.Default.MultithreadMode; 
+            CmdMultithreadMode_Executed(this, EventArgs.Empty); 
+
             foreach (var control in _pads)
                 control.LoadSettings();
 
@@ -596,6 +599,11 @@ namespace MonoGame.Tools.Pipeline
         private void CmdCancelBuild_Executed(object sender, EventArgs e)
         {
             PipelineController.Instance.CancelBuild();
+        }
+
+        private void CmdMultithreadMode_Executed(object sender, EventArgs e)
+        {
+            PipelineSettings.Default.MultithreadMode = cmdMultithreadMode.Checked;
         }
 
         private void CmdHelp_Executed(object sender, EventArgs e)
--- a/Build/Projects/MGCB.definition
+++ b/Build/Projects/MGCB.definition
@@ -35,6 +35,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="CommandLineParser.cs" />
     <Compile Include="BuildContent.cs" />
+    <Compile Include="BuildAsyncState.cs" />
     <Compile Include="ConsoleLogger.cs" />
     <Compile Include="ConsoleAsyncLogger.cs" />
     <Compile Include="SourceFileCollection.cs" />
--- a/Build/Projects/MGCB.definition
+++ b/Build/Projects/MGCB.definition
@@ -36,6 +36,7 @@
     <Compile Include="CommandLineParser.cs" />
     <Compile Include="BuildContent.cs" />
     <Compile Include="ConsoleLogger.cs" />
+    <Compile Include="ConsoleAsyncLogger.cs" />
     <Compile Include="SourceFileCollection.cs" />
     <None Include="app.config" />
 
--- a/MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs
@@ -68,6 +68,9 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
         [XmlIgnore]
         public OpaqueDataDictionary Parameters { get; set; }
 
+        [XmlIgnore]
+        public ContentBuildLogger Logger { get; set; }
+
         public class Pair
         {
             public string Key { get; set; }
--- a/MonoGame.Framework.Content.Pipeline/Builder/PipelineImporterContext.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PipelineImporterContext.cs
@@ -10,14 +10,17 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
     {
         private readonly PipelineManager _manager;
 
-        public PipelineImporterContext(PipelineManager manager)
+        private readonly PipelineBuildEvent _pipelineEvent;
+
+        public PipelineImporterContext(PipelineManager manager, PipelineBuildEvent pipelineEvent)
         {
             _manager = manager;
+            _pipelineEvent = pipelineEvent;
         }
 
         public override string IntermediateDirectory { get { return _manager.IntermediateDirectory; } }
         public override string OutputDirectory { get { return _manager.OutputDirectory; } }
-        public override ContentBuildLogger Logger { get { return _manager.Logger; } }
+        public override ContentBuildLogger Logger { get { return _pipelineEvent.Logger; } }
 
         public override void AddDependency(string filename)
         {            
--- a/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs
@@ -525,7 +525,7 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
             TrackPipelineBuildEvent(contentEvent);
         }
 
-        public PipelineBuildEvent BuildContent(string sourceFilepath, string outputFilepath = null, string importerName = null, string processorName = null, OpaqueDataDictionary processorParameters = null)
+        public PipelineBuildEvent BuildContent(ContentBuildLogger logger, string sourceFilepath, string outputFilepath = null, string importerName = null, string processorName = null, OpaqueDataDictionary processorParameters = null)
         {
             sourceFilepath = PathHelper.Normalize(sourceFilepath);
             ResolveOutputFilepath(sourceFilepath, ref outputFilepath);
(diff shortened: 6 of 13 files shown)


tomspilman:
I'll dive into the details of this soon... but great to see progress on it!

tomspilman:
So i'm finding it hard to follow how you have dealt with the logging issue in this.  Maybe you can explain how this addresses things?

I see three big problems.

 1. We can't have errors/warnings/messages from several content items are mixed up together.
 2. We can't have content that takes a long time to build not output anything until it is complete.
 3. The simple fix for 1 will break 2 and vice versa.

KonajuGames:
One option is similar to Visual Studio.  Each asset gets an incrementing
number and this is used as a prefix in the output, e.g.

1> Building file.PNG
2> Building file2.PNG
1> Error processing file.PNG

The filtered view could group log output based on the prefix.

tomspilman:
> The filtered view could group log output based on the prefix.

Right... we might be able to fix things for the filter view acceptably.  IMO the filter view is not a good alternative to the regular output log... but it really could be with more effort.

nkast:
| We can't have errors/warnings/messages from several content items are mixed up together.

That's a requirement along with that we can't have output of assets out of order, output must be the same each time and match that of the current singlethread output.

| We can't have content that takes a long time to build not output anything until it is complete.

That's a problem, I plan to address that by having the logger on top of the queue output directly to the console. 
There's still the problem that task/threads run at random the pipeline might look frozen, the 'TaskCreationOptions.PreferFairness' option doesn't exactly work. We have to merge the two loops and create new tasks as the old tasks complete, keep the number of tasks down to `Environment.ProcessorCount`.

tomspilman:
>  output must be the same each time and match that of the current singlethread output.

I don't think that is a critical issue really...  if the order is different we could be fine with that.  However I just realized one big issue...  processing FBX files where the same texture is used with different processing settings:

  https://github.com/mono/MonoGame/pull/4198

IMO we could/should just change the naming standard for these dependent files to be based on some sort of consistent hash and not just an incremented number.


> I plan to address that by having the logger on top of the queue output directly to the console.

Interesting solution.

The only potential issue there is one long running task getting stuck at the top of the queue and having a bunch of other small tasks already completed, but can't log their results yet.  Could be mitigated by always selecting the task that has been running the least amount of time (or already completed) to the top first.

Still you could get unlucky and pick a task that has been running very little time so far, but ends up taking a very long time to finish.

nkast:
I pushed some more changes. Can anyone test this?
I add a switch  in `Build` menu to switch between singlecore/multicore.


@tomspilman 
> However I just realized one big issue... processing FBX files where the same texture is used with different processing settings:

I fix that in commit: https://github.com/mono/MonoGame/pull/4450/commits/79bcbfb9f3ade5d24ecf923548b56cd503371e79

nkast:
@KonajuGames 
I am geting a lot of System.NullReferenceException with DXT compressed textures.
Similar to http://community.monogame.net/t/mgcb-dxt-compression-sometimes-throws-nullpointer/7356/4

```
	Could not convert texture. System.NullReferenceException: Object reference not set to an instance of an object.
   at Nvidia.TextureTools.Compressor.nvttCompress(IntPtr compressor, IntPtr inputOptions, IntPtr compressionOptions, IntPtr outputOptions)
   at Microsoft.Xna.Framework.Content.Pipeline.Graphics.DxtBitmapContent.TryCopyFrom(BitmapContent sourceBitmap, Rectangle sourceRegion, Rectangle destinationRegion)
   at Microsoft.Xna.Framework.Content.Pipeline.Graphics.BitmapContent.Copy(BitmapContent sourceBitmap, Rectangle sourceRegion, BitmapContent destinationBitmap, Rectangle destinationRegion)
   at Microsoft.Xna.Framework.Content.Pipeline.Graphics.BitmapContent.Copy(BitmapContent sourceBitmap, BitmapContent destinationBitmap)
   at Microsoft.Xna.Framework.Content.Pipeline.Graphics.GraphicsUtil.Compress(Type targetType, TextureContent content, Boolean generateMipMaps)
   at Microsoft.Xna.Framework.Content.Pipeline.Processors.TextureProcessor.Process(TextureContent input, ContentProcessorContext context)

```

nkast:
Rebase from latest develop brunch.

KonajuGames:
I'm going to have to setup a solution with the C# for the MGCB and C++ for NVTT so I can find what this DXT crash is.

nkast:
@KonajuGames Great! 
Don't forger to enable multithread mode from the menu and test a release build.
multithread

nkast:
@KonajuGames The exception occur somewhere betwwen those lines
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Graphics/DxtBitmapContent.cs#L178-L182

It's hard to tell what is going on when the code has been optimized and with out the native c++ code.
By marking the method with `[MethodImpl(MethodImplOptions.NoOptimization)]` I don't get the error any more. I will give you some time to review the code and see if there is a bug somewhere in the nvtt wrapper.

KonajuGames:
Marking it as NoOptimization is interesting. I can get the crash in Debug
builds as well. It depends on the input texture but I haven't worked out
what about the input texture makes it go bad.

NVTT uses a modified version of libsquish for DXT compression. I was
wanting to use libsquish directly, and had it running successfully, but
certain blocks in the compressed output weren't correct. This was the only
reason I went back to NVTT.

nkast:
Is that the texture from @Ferenczfi? 
http://community.monogame.net/t/upgrading-to-monogame-3-5-1-from-3-5-0-1122-processor-textureprocessor-had-an-unexpected-failure/7432/8

If I understand correctly he gets a Null exception every time while I am only getting it on random textures when running multiple threads and very rarely on the single thread build. 

Is it possible to have two unrelated bugs here?

KonajuGames:
Could be. Have to be careful when multithreading with external libraries
that may or may not be thread safe.

MonoGame.Framework.Content.Pipeline/ContentProcessorContext.cs
```diff
@@ -24,7 +24,7 @@ public abstract class ContentProcessorContext
         /// <summary>
         /// Gets the logger interface used for status messages or warnings.
         /// </summary>
-        public abstract ContentBuildLogger Logger { get; }
```
nkast: We can't change this to `IContentBuildLogger` because it's part of the public API. Old processors throw `System.MissingMethodException`.

nkast:
@tomspilman 
Fails on DynamicSoundEffectInstance tests, for example:
```
[MonoGame.Tests.Audio.DynamicSoundEffectInstanceTest.BufferNeeded_DuringPlayback] Test(s) failed. System.NullReferenceException : Object reference not set to an instance of an object.
   at SharpDX.XAudio2.SourceVoice..ctor(XAudio2 device, WaveFormat sourceFormat, Boolean enableCallbackEvents)
   at Microsoft.Xna.Framework.Audio.DynamicSoundEffectInstance.PlatformCreate() in c:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\MonoGame.Framework\Audio\DynamicSoundEffectInstance.XAudio.cs:line 23
   at Microsoft.Xna.Framework.Audio.DynamicSoundEffectInstance..ctor(Int32 sampleRate, AudioChannels channels) in c:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\MonoGame.Framework\Audio\DynamicSoundEffectInstance.cs:line 87
   at MonoGame.Tests.Audio.DynamicSoundEffectInstanceTest.BufferNeeded_DuringPlayback() in c:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\Framework\Audio\DynamicSoundEffectInstanceTest.cs:line 30
```

tomspilman:
@nkast - Can you rebase/merge the latest develop into this branch?  That should fix the unit test issues.

nkast:
The https://github.com/castano/nvidia-texture-tools has many copies of the native wrapper in /project/[vc8|vc9|vc10|vc12]/Nvidia.TextureTools/ none of which is up to date. 
The samples suggest that the way to use the lib has change.

 The original push of TextureTools.cs at https://github.com/castano/nvidia-texture-tools/blob/42220b981edb27b10b8db46c6b510ad8d06807ce/project/vc9/Nvidia.TextureTools/TextureTools.cs looked close to what we have now so i tried to build and test it against the nvtt.dll we currently use. Some method had to be made public, uncomment others and recreate a missing method. At the end it failed to run because of some missing method!

So I still can't test what is causing the NullReferenceException and whether or not it's fixed on the latest  nvtt.

nkast:
Ok, now that the DXT issues are resolved it's time to revisit that one. 

The time it takes to build assets is around 60% -70% of the singlethreaded build, not what i would expect from a 4-core CPU. I redirected the output to another disk but that didn't make a difference so it can't be a disk bottleneck. Also I tried to increase the number of active thread but that didn't help either.

The generated .XNBs are identical.

The output is *almost* identical.
I have a custom importer that outputs messages by writing to the Console and cause messages to appear at random position. Of course I can pass the logger all the way down to that method but I am worried how widespread is that practice and whether or not we should address it. The importer is ported by the XNA Skinning sample.
It's possible to intercept Console output and cache it per thead:
http://stackoverflow.com/questions/10839434/two-thread-two-streams-for-console-out
If that works we can possibly remove `ConsoleAsyncLogger` altogether.
@tomspilman , @KonajuGames Thoughts?

I have a switch under `Build` menu to toggle between Multi/Single thread. I am thinking to remove it from `PipelineSettings` so it will always be disabled when pipeline is loaded. So this could be an experimental feature in the beginning until we collect enough feedback.

tomspilman:
> The time it takes to build assets is around 60% -70% of the singlethreaded build, not what i would expect from a 4-core CPU.

Well 4 cores, but they have to share cache and memory bandwidth... so totally possible this is as fast as it gets with your PC.

It is probably worth doing more detailed performance profiling here to be sure.


> The generated .XNBs are identical.

That is the critical part...  if for any reason it generates different XNBs then this feature is broken.


> It's possible to intercept Console output and cache it per thead:

As long as it can be done reliably on Windows, Mac, and Linux and it doesn't require people to change their content processors/importers then i'm good with it.

If not some differences in the console output would be acceptable.


> So this could be an experimental feature in the beginning until we collect enough feedback.

I agree... IMO we could release this now as is.  It would just be experimental and we should likely label it that way in the menu item.

I have several projects here that have over 1000 assets which would be a good test for this.

@KonajuGames @dellis1972 ?

KonajuGames:
The identical XNBs are a good thing.  I'm ok with merging it now too.

nkast:
I changed the mgcb command to `/multithreadMode` and the menu checkbox to `Multithread Mode (experimental)`. It would be always disabled on startup.
Also fixed a bug where every now and then buildTaskQueue was empty and threw an exception.

Tested on windows 10.

Tools/Pipeline/MainWindow.cs
```diff
@@ -66,6 +66,10 @@ public void Attach(IController controller)
 
             cmdDebugMode.Checked = PipelineSettings.Default.DebugMode;
             CmdDebugMode_Executed(this, EventArgs.Empty);
+            
+            // MultithreadMode is experimental and disabled on startup.
+            //cmdMultithreadMode.Checked = PipelineSettings.Default.MultithreadMode;
```
cra0zy: There really is no point in not allowing the setting for Multithreaded Mode to be saved between instances...

Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -49,6 +49,7 @@ public IEnumerable<ContentItemTemplate> Templates
         }
 
         public bool LaunchDebugger { get; set; }
+        public bool MultithreadMode { get; set; }
```
cra0zy: I see that you were copying the way `LaunchDebugger` was done, but the `Settings` class was added after this variable, and it would make more sense to just use the `Settings.Instance.variable` directly instead of adding an extra variable in the Controller.

Tools/Pipeline/MainWindow.cs
nkast: Well, I though it would be better since we just test it out but after using it for a few times I too don't see a point since we clearly state it's experimental.

Tools/Pipeline/Common/PipelineController.cs
nkast: Indeed. I'll take a look at `Settings` to see how it work. This implies that the setting will be saved between instances, right?

cra0zy: Yea, just use the variable in `Settings.Instance` instead, `Settings` is a class that gets serialized and saved during the application exit so you don't need to do anything special to save a variable in there.

nkast: You must mean `PipelineSettings.Default`, right?

cra0zy: Yes, sorry.

nkast: No worries, I made the changes. Thanks for the feedback.

nkast:
@KonajuGames , @tomspilman 
I squash and reorganized the commits into 8 steps.

The first 4 are required changes to make the code thread safe. I can push those as separate PRs if you like to review them one by one.

The commit that adds the actual multithreading code is at `BuildContent in parallel` 0d32980.

The last one adds the menu switch to the pipeline.

Issue #4452, opened 2016-01-22
SoundEffect.Play does nothing on Android


dellis1972:
Latest unstable release, Music seems to play ok but sounds effects seem to do nothing. 
We probably should fix this (if its an issue and not must me) before the 3.5 release.

mrhelmut:
It must be something wrong with ```OpenALSoundController```. Is there anything in the console output?

I can't debug on Android, can anyone try this branch and tell what's in the console output: https://github.com/mrhelmut/MonoGame/tree/OpenALCheck

It adds error checks on every AL calls, if there's something wrong it will either crash or print errors in the console.

tomspilman:
@mrhelmut - Could we do something like we do for OpenGL where the error checking conditional for debug builds:

https://github.com/mono/MonoGame/blob/50a348a6ce7dae4e397353b7459c37d906dce587/MonoGame.Framework/Graphics/GraphicsExtensions.cs#L910

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs#L291

mrhelmut:
Sure, I'll do a pass on all AL calls. Let me know if you prefer that I submit a PR now to help resolve this issue or if you wish to wait until it is resolved for future maintenance.

tomspilman:
> Sure, I'll do a pass on all AL calls.

Great... the sooner the better. It can't hurt in finding this and other potential bugs.

dellis1972:
@mrhelmut ok.. nothing in the log :( it probes and finds the openal end points without any issue. Just no sound. neither SoundEffect or SoundEffectInstance works. AL.SourcePlay is not returning any errors either.  @KonajuGames have we updated the openal_soft stuff for android recently?

dellis1972:
OK scratch that.AL.GenBuffers(1, out openALDataBuffer); is failing with IllegalCommand...

dellis1972:
Probably because the OpenALSoundController is not being initialised BEFORE the content is loaded. So it can't create an buffers

dellis1972:
[...]

PR #4453, opened 2016-01-22
Fixed SoundEffect.Play does nothing on Android #4452

dellis1972:
The OpenALSoundController was not being initialized before
the SoundEffects were trying to load.

Tested on Android and MacOS (DesktopGL) works fine

### DIFF
--- a/MonoGame.Framework/Android/AndroidGamePlatform.cs
+++ b/MonoGame.Framework/Android/AndroidGamePlatform.cs
@@ -11,6 +11,8 @@ namespace Microsoft.Xna.Framework
 {
     class AndroidGamePlatform : GamePlatform
     {
+        OpenALSoundController soundControllerInstance = null;
+
         public AndroidGamePlatform(Game game)
             : base(game)
         {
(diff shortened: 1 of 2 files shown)


tomspilman:
This seems like it was some sort of static initialization error?

mrhelmut:
Ah!
```OpenALSoundController``` is a singleton. I think I never encountered it because I load and play ```Song``` (which initialize the ```OggStreamer``` and therefore OpenAL) before loading ```SoundEffect```...

Maybe doing something like:
```
            if (OpenALSoundController.GetInstance == null)
                        return;
            // bind buffer
            SoundBuffer = new OALSoundBuffer();
            SoundBuffer.BindDataBuffer(_data, Format, Size, (int)Rate);
```
would avoid having a misleading "unused" parameter.

Working great. :+1:

KonajuGames:
If it is a singleton that simply requires initialization, why is it not initialized as part of the platform initialization?  Having OggStreamer initialize it or SoundEffect initialize it on first use means it could potentially be initialized on worker threads.  But then, I don't know how thread-safe OpenAL is.

dellis1972:
Moved the Initialisation to the AndroidGamePlatform constructor like the DesktopPlatform

dellis1972:
@tomspilman not sure why that test failed.. can we reset he bot?

tomspilman:
> can we reset he bot?

I reran the build.

dellis1972:
@tomspilman looks like it passed this time.. weird.

mrhelmut:
Looks perfect. :+1:

KonajuGames:
Merging.

PR #4455, opened 2016-01-23
[android] Fix Accelerometer state.

dellis1972:
The state valiable is NOT updated after the accelerometer has
been created so it always returns NotSupported..


tomspilman:
Merged!

Issue #4457, opened 2016-01-23
Pipeline Tool Issues


tomspilman:
There are a bunch of issues I noticed when using the MonoGame Pipeline tool on Windows.  Some of this might be Windows specific... some of it might apply the to GTK frontend as well.

 - We should include the open project name in the title bar (see VS).
 - It isn't obvious if a content build is active occurring.
 - We lost the build succeeded/failed message and build time at the end of the output log.
 - The control border around the filter view is different that that around the output view.
 - The output and filter views fights the user set scrollbar position.
 - Text selection in the output view doesn't work correctly.
 - The filter view should have buttons to allow you to filter stuff out... like not showing skipping messages.
 - The 'References' property value should show a `;` separated string of assembly names.
 - Undo/Redo should be visible on the toolbar.
 - Toolbar icons do not update to disabled when a build is triggered.

cra0zy:
> We should include the open project name in the title bar (see VS).

Oh... I totally forgot to implement it for WinForms Pipeline, Gtk had it for half a year... #3912

![title](https://cloud.githubusercontent.com/assets/6773302/12533781/3c742eb8-c23c-11e5-9c68-215065baa77f.png)

> some of it might apply the to GTK frontend as well.

Issues that apply to Gtk:
 - It isn't obvious if a content build is active occurring.
   - In headerbar part it is visible since the last change added changing of build/rebuild, cancel buttons #4418
   - I would like for us to add some sort of progressbar, then for Windows we could also use taskbar progressbar
 - The output and filter views fights the user set scrollbar position.
   - we should add a check to see if it's scrolled to the bottom, if not don't scroll
 - The filter view should have buttons to allow you to filter stuff out... like not showing skipping messages.
   - this should be a separate issue, and we should discuss it in more detail
 - The 'References' property value should show a ; separated string of assembly names.
 - Undo/Redo should be visible on the toolbar.

@nkast Join the fun talk :)

cra0zy:
One more thing, if I could add one more thing to the list, we should add "Open With" option.

tomspilman:
> In headerbar part it is visible since the last change 

Having some buttons change to a disabled state isn't obvious enough.  We really need some sort of explicit animated building graphic.

> I would like for us to add some sort of progressbar

Depends on what we want here.

[...]

Issue #4238, opened 2015-11-10
Use Xwt for dialogs in Gtk Pipeline


cra0zy:
I was thinking that in Pipeline tools open/save file dialogs don't look native on OS X, so why don't we just use Xwt for those dialogs.

@dellis1972 Thoughts?

cra0zy:
Actually maybe http://stackoverflow.com/questions/33341914/gtk-file-chooser-non-native-look-on-osx is better.

CallumDev:
Remember MonoMac is 32-bit only, and Xamarin.Mac requires licensing from Xamarin to use.

KonajuGames:
For all those OS X file open/save dialog options, it seems to be pick-the-least-worst.  I do think native is better for file open/save dialogs, no matter the platform, especially on platforms that have native dialogs.

tomspilman:
> I do think native is better for file open/save dialogs, no matter the platform

Agreed...  people expect file dialogs to look like what their native platform provides.

cra0zy:
I just realized something even better, we could use Xwt or Eto.Forms for all dialogs and replace dialogs in: https://github.com/mono/MonoGame/tree/develop/Tools/Pipeline/Gtk/Dialogs and https://github.com/mono/MonoGame/tree/develop/Tools/Pipeline/Windows with one set of them. 

This would have 2 advantages:
1) WinForms Pipeline and Gtk Pipeline would share the same dialogs
2) Dialogs on Mac would be able to look native

tomspilman:
>  we could use Xwt or Eto.Forms for all dialogs

Designing dialogs once and having them apply to all platforms has great benefits in code maintenance, documentation, and testing...  this is why I think this is worth considering.

That said... I am concerned about two things:

 1. The results won't feel 100% native on Windows.
 2. There will be WinForm features we want to use that we can't.

As long as things feel 100% native on Windows and we're not too limited in getting the results we want...  I can live with this.

Off a first glance O would say Eto.Forms is better because it supports WinForms, WPF, and WinRT support coming...  giving us more options in the future.

We could start with an experiment in dialogs and see how it goes.

cra0zy:
I did do a xwt port of Pipeline Tool as a proof of concept a while back: https://github.com/cra0zy/MonoGame/tree/xwt

So I have all the xwt dialogs mostly done:
https://github.com/cra0zy/MonoGame/tree/xwt/Tools/Pipeline/Xwt/Dialogs

[...]

PR #4459, opened 2016-01-24
[Gtk Pipeline] General Improvements #4457

cra0zy:
This PR fixes stuff mentioned in #4457 for Gtk Pipeline Tool.

Stuff done:
 - Added "Open With", Linux only
 - Added Undo/Redo toolbar buttons
 - Added Cancel toolbar button which appears while the build is running, and build/rebuild buttons hide
 - Updated reference dialog to mimic WinForms one
 - Updated how reference property is displayed in propertygrid
 - Made headerbar separator invisible in case build/rebuild/cancel build aren't visible
 - Allowed user scrolling of build output possible
 - Smoother scrolling
 - Fixed opening of files that contain spaces

Update no. 1:
 - Toolbar items / Headerbar items are now connected to actions, this means that changing something on action will change them as well
 - Headerbar now has a proper menu

Update no. 2:
 - Fixed pressing open and then canceling creating an empty base item
 - Separate "Exclude from Project" and "Delete" options
 - Made build output cursor invisible (not mouse cursor, text cursor, text can still be selected)


Tools/Pipeline/Gtk/MainWindow.cs
```diff
@@ -742,11 +742,26 @@ public void OnRenameActionActivated (object sender, EventArgs e)
...
         {
-            projectview1.Remove ();
-            UpdateMenus();
+            var dialog = new MessageDialog(this, DialogFlags.Modal, MessageType.Warning, ButtonsType.None, "How do you want to remove selected file/s?");
+            dialog.Title = "Delete File/s";
+            dialog.SecondaryText = "Do note that selecting \"Delete\" option will delete the file and this action will not be undoable.";
+
+            dialog.AddButton("Delete", ResponseType.Apply);
+            dialog.AddButton("Cancel", ResponseType.Cancel);
+            dialog.AddButton("Remove", ResponseType.Ok);
```
tomspilman: The alternative implementation is having a separate Delete and Remove commands sort of like VS.

I do worry that mixing them both together like this... then having a 3 button prompt... could lead to costly mistakes by users.

A remove which works without prompts and a delete that asks a "Are you sure this is not undoable"... seems like a safer approach.

@dellis1972 @KonajuGames ?

cra0zy: MonoDevelop/Xamarin Studio has this type of delete dialog:

deletedilalog

cra0zy:
I am gonna use this PR to test something.

cra0zy:
I am done with testing. One of the things I tested was Gtk Pipeline on Windows, with Gtk 3 the window looks very good, but the menus in menubar are awful, and with Gtk 2 menubar and toolbar background colors are off, which isn't that hard to override.

Gtk on Windows it's 32 bit only, so I wouldn't use it, this was more of a test to see how it looks.

cra0zy:
@tomspilman I've done what you've asked.

cra0zy:
@dellis1972 @tomspilman Can this be merged so I can start working on Win Forms side of things for #4457 ?

cra0zy:
Bummer :|

tomspilman:
Sorry... just been busy with work stuff.  I'll get to it tonight for sure.

cra0zy:
I noticed that CollectionEditorDIalog had tabs left, so I just replaced them with spaces.

cra0zy:
> Bummer :|

Agreed :P

cra0zy:
I rebases this so it builds with the new build system.

cra0zy:
@tomspilman @KonajuGames @dellis1972 Can this be merged for 3.5.1? I am tired of recompiling it each time I want to update system MG...

KonajuGames:
Added it to the 3.5.1 milestone label so it doesn't get missed.

cra0zy:
Thanks

tomspilman:
I've moved this to 3.6...  we will wrap up this 3.5 service release this week, but that shouldn't keep this from being merged.

cra0zy:
Since 3.5.1 got merged, can this get merged now? I really really really want to start working on what's discussed in #4238

tomspilman:
Thanks @cra0zy !

PR #4462, opened 2016-01-25
Cleanup FX Syntax For Shader Compiler

tomspilman:
This PR uses the existing MGFX parser token tree to replace all FX syntax with whitespace.  This removes the FX syntax the system already parsed so that it isn't passed thru to the shader compiler step.  By replacing it with whitespace it ensures we don't break any line number info in errors and debugging tools.

This is one more step towards GLSL support too (see https://github.com/mono/MonoGame/issues/2167).


tomspilman:
This too can wait for 3.5 to get finished.

tomspilman:
Merging!

PR #4464, opened 2016-01-25
Adding OpenAL error checking everywhere with more verbosity

mrhelmut:
Follow up to #4452

I also moved ```OggStream``` to the ```Audio``` namespace to have every OpenAL bits under the same one.

@tomspilman @KonajuGames


mrhelmut:
Fixing...

mrhelmut:
@tomspilman @KonajuGames Does this one comply to what we discussed in #4452? We should be able to better identify which AL calls within a method may fail.

KonajuGames:
It looks good.  I'll merge this now.

PR #4467, opened 2016-01-26
Target platform identifier cleanup

tomspilman:
- Added feature to allow for static platform initialization in `ContentManager`.
- Clearly labeled all legacy platform identifiers.
- Added warning about console platform identifiers.


KonajuGames:
Merged.

PR #4469, opened 2016-01-26
SoundEffect Fixes and Tests

tomspilman:
This PR starts to add some unit tests for `SoundEffect` while fixing bugs.  I also tried to cleanup platform initialization limiting it to one path for PCM data and another for non-PCM formats.

I cleaned up the `#if` mess in `SoundEffectReader` by moving the work into `SoundEffect` platform methods.

This needs further testing before merging.


MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
```diff
@@ -125,45 +125,44 @@ private void PlatformLoadAudioStream(Stream s)
...
-            else
-                Format = bitsPerSample == 8 ? ALFormat.Stereo8 : ALFormat.Stereo16;
-
-            _name = "";
-            _data = buffer;
-
-#endif
+            // Make our own copy of the data.
+            _data = new byte[count];
+            Array.Copy(buffer, offset, _data, 0, count);
```
tomspilman: This wasn't here before, but it seems dangerous to not copy the passed in sound buffer as the caller could alter it after the fact.  In fact I expect under XNA it does something similar copying the passed in buffer into the native code.

MonoGame.Framework/Audio/SoundEffect.cs
```diff
@@ -17,32 +17,63 @@ public sealed partial class SoundEffect : IDisposable
...
+            // Everything else is platform specific.
+            PlatformInitializeFormat(buffer, format, sampleRate, channels, blockAlignment, loopStart, loopLength);
+        }
 
         #endregion
 
         #region Public Constructors
 
-        /// <param name="buffer">Buffer containing PCM wave data.</param>
+        /// <param name="buffer">Buffer containing 16bit PCM wave data.</param>
```
tomspilman: There seemed to be some confusion in the code base about this, but the `SoundEffect` public ctors in 
XNA only support 16bit PCM.

MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
```diff
@@ -125,45 +125,44 @@ private void PlatformLoadAudioStream(Stream s)
...
+                        (short)((blockAlignment / channels) - 22));
 
-            throw new NotImplementedException();
-        }
+                    format = 1;
+                }
+            }
 
+            if (format != 1)
+                throw new NotSupportedException("Unsupported wave format!");
```
tomspilman: Does the OpenAL code on iOS/MacOSX support anything other than PCM?  The pipeline suggests it supports IMA ADPCM...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/SoundEffectProcessor.cs#L48

Or is this broken?

tomspilman:
I tested the desktop DX and GL paths which seem to be working great.  Just need to be sure I didn't break iOS or Android somehow.

MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
KonajuGames: We need to make sure we're not duplicating some of these buffers. If the buffer is owned by SoundEffect, we shouldn't need to duplicate it here. If it is the buffer passed to SoundEffect ctor, I would expect SoundEffect to copy it first and it owns that copy.

MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
KonajuGames: I've seen some blogs that show how to decode IMA ADPCM for playback through OpenAL on iOS. I believe the higher level audio functions support ADPCM, but the OpenAL layer may not.  OpenAL Soft on Android does support ADPCM (IMA and MS), but I'm not sure if it has been fully tested.

MonoGame.Framework/Audio/SoundEffect.cs
KonajuGames: It would be good to extend this for the platforms that can support higher fidelity, such as 24-bit PCM and floating point. It has been brought up in comments before.

tomspilman: > It would be good to extend this for the platforms that can support higher fidelity

Seems like that could be fit in with some additional `bitdepth` passed to these ctors and helper functions.

MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
tomspilman: I guess that means the pipeline is generating compressed audio that currently doesn't work on iOS/MacOS.  Can you verify this @dellis1972 ?

dellis1972: @tomspilman with the current pipeline tool I can play audio fine on iOS and Mac as well. No idea what its generating though

tomspilman: @dellis1972 - If you set the quality on the sound in the pipeline to medium or low does it still work?

dellis1972: I seem to remember allot of noise/static when you do that. Best works fine
though.

tomspilman: > I seem to remember allot of noise/static when you do that. Best works fine
though.

That means we have a bug there.  The pipeline is specifically generating compressed audio for medium/lowest... it makes sense you are getting static because we've made no attempt to support that compressed format in the OpenAL code.

Anyway... i'm not introducing any new bugs here then.  We need to deal with this in a separate issue.  See https://github.com/mono/MonoGame/issues/4470.

tomspilman:
Retesting this now on Windows... be good to have someone test on Mac/Linux and ideally iOS and Android.

tomspilman:
Actually new plan.

@KonajuGames - After this is merged I will follow it up with set of unit tests for `SoundEffectInstance` following the pattern @polsys defined in https://github.com/mono/MonoGame/pull/4715.  My time is better spent making a good test rather than manually testing yet another time.

tomspilman:
@mrhelmut - Maybe you should look this over once too?

mrhelmut:
@tomspilman works just fine on DesktopGL.

KonajuGames:
Merged.

PR #4471, opened 2016-01-27
Fix nullable enum serialization bug

tgjones:
Prior to this fix, serializing or deserializing a nullable enum using `IntermediateSerializer` would crash in `NullableSerializer.Initialize` because `EnumSerializer` couldn't be cast to `ElementSerializer<T>`.

This pull request simplifies `NullableSerializer<T>` by using the existing `ReadRawObject` and `WriteRawObject` methods, which remove the need for the `List<T>` workaround.


tgjones:
It would be awesome (if you think it's safe enough) if this could go into the v3.5 release :)

tomspilman:
Seems reasonable to me.  I assume you ran the XNA unit tests and verified this passes?

tgjones:
Yes - I ran the XNA tests, and they pass.

tomspilman:
@KonajuGames - If this looks ok to you i guess we could merge it.

dellis1972:
looks ok to me :)

KonajuGames:
Merging.

tgjones:
Thanks!

Issue #4468, opened 2016-01-26
Displaced spritefonts comparing monogame with xna


Tzenchor:
After some days before realizing that @tomspilman said that i should make a new issue about this in pull #4439.
After many tries comparing fonts, the problem must be in the pipeline (in my project i don't use measureString so it couldn't benefit from @MichaelDePiazzi changes). I did a test program to try different fonts and the area reported by measurestring. and i get that the fonts JingJing and Lindsey from the redistributable package produce very similar results as shown here:
![jing](https://cloud.githubusercontent.com/assets/11319966/12584591/936b1908-c448-11e5-9d56-6f6c01ccdc50.gif)
![linds](https://cloud.githubusercontent.com/assets/11319966/12584592/94f06c7e-c448-11e5-93ed-0163d155fce9.gif)
But when using "Trebuchet MS" for example there are some problems as shown here:
![trebu](https://cloud.githubusercontent.com/assets/11319966/12584732/2ddfc146-c449-11e5-9fa7-4f04700914aa.gif)
Notice the big difference in the black area (it's the reported size from measureString). And also i you zoom on the 'Ç' character you can see that it is drawn below the black box. The font has size 48, spacing 0, kerning, and character region goes from 32 to 256.

The size of each character compared to XNA is 1 pixel lower (61 to 62 for 'Ç', or 58 to 59 for 'Í') but that may come from the different rendering method, and the out of bounds drawing comes, if i have understood correctly the Freetype documentation, from the rounding of the hinting algorithm (the face reports a height of 74, with an ascender of 60 and a descender of 14, but 'Ç' has a descender of 15).

I'm thinking that the way the fonts gets proccesed in each pipeline, is different, I'll try later to check if grabbing metrics for each character can provide different results.

tomspilman:
> the fonts JingJing and Lindsey from the redistributable package produce very similar results

This then brings the question if is the difference here a bug or a feature?  Which of the XNA or MonoGame results is more correct?  We use a different font rasterizers... so there will be some difference... but is there a bug to correct here?

> and the out of bounds drawing

So this means the character's sprite is larger than the metrics used for measure string?


> I'm thinking that the way the fonts gets proccesed in each pipeline, is different,

Right... totally different font rasterizers.  It is hard to pinpoint when a difference is a bug in our processing or just a natural difference in rasterizers.

Here is an old issue where I have some of my findings last time I looked into these things: https://github.com/mono/MonoGame/issues/2741

[...]

PR #4472, opened 2016-01-28
Upgrade sharp font importer to provide fitter size for texts

Tzenchor:
This pull changes the way fonts are imported. Previously the size of the font was taken from the font autohinter, and this side has proved being wrong in some cases, see #4468 with Trebuchet MS.
Now the font gets read in its entirety to measure it.
Also i have added the SharpFont documentation

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs
@@ -9,117 +9,157 @@ using SharpFont;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 {
-	// Uses FreeType to rasterize TrueType fonts into a series of glyph bitmaps.
-	internal class SharpFontImporter : IFontImporter
-	{
-		// Properties hold the imported font data.
-		public IEnumerable<Glyph> Glyphs { get; private set; }
+    // Uses FreeType to rasterize TrueType fonts into a series of glyph bitmaps.
+    internal class SharpFontImporter : IFontImporter
+    {
+        // Properties hold the imported font data.
+        public IEnumerable<Glyph> Glyphs
+        {
+            get; private set;
+        }
 
-		public float LineSpacing { get; private set; }
+        public float LineSpacing
+        {
+            get; private set;
+        }
 
-		public int YOffsetMin { get; private set; }
+        public int YOffsetMin
+        {
+            get; private set;
+        }
 
-		// Size of the temp surface used for GDI+ rasterization.
-		const int MaxGlyphSize = 1024;
+        // Size of the temp surface used for GDI+ rasterization.
+        const int MaxGlyphSize = 1024;
 
-		Library lib = null;
+        Library lib = null;
 
-		public void Import(FontDescription options, string fontName)
-		{
-			lib = new Library();
-			// Create a bunch of GDI+ objects.
-			var face = CreateFontFace(options, fontName);
-			try
+        public void Import(FontDescription options, string fontName)
+        {
+            lib = new Library();
+            // Create a bunch of GDI+ objects.
+            var face = CreateFontFace(options, fontName);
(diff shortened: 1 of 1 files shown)


KonajuGames:
Since SharpFont is in MonoGame.Dependencies, the doc files need to be merged into that repo first, then the submodule reference is updated in MonoGame in a separate PR.

Tzenchor:
Great, i did it completely wrong.
I'll do a new pull request with only the changes to `sharpfontimporter.cs` and the documentation will come later.
If someone wants to have a look to the code put your criticisms here so i can make the changes in for the new pull request

KonajuGames:
Aside from the reformatting (which just adds noise to an otherwise simple change), it looks ok.

For future PRs, turn off the automatic reformatting.

MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs
```diff
@@ -9,117 +9,157 @@
...
+
+        // Rasterizes a single character glyph.
+        //private Glyph ImportGlyph(char character, Face face)
+        //      {
+        //          uint glyphIndex = face.GetCharIndex(character);
+        //          face.LoadGlyph(glyphIndex, LoadFlags.Default, LoadTarget.Normal);
+        //          face.Glyph.RenderGlyph(RenderMode.Normal);
+
+        //          return RetrieveGlyphData(character, face);
+        //      }
```
tomspilman: Don't leave in old code like this... go ahead and delete used code.

MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs
```diff
@@ -9,117 +9,157 @@
...
-			face.Glyph.RenderGlyph(RenderMode.Normal);
-
-			// Render the character.
+                const uint dpi = 96;
+                var face = lib.NewFace(fontName, 0);
+                var fixedSize = ((int) options.Size) << 6;
+                face.SetCharSize(0, fixedSize, dpi, dpi);
+
+                if (face.FamilyName == "Microsoft Sans Serif" && options.FontName != "Microsoft Sans Serif")
+                    throw new PipelineException(string.Format("Font {0} is not installed on this computer.", options.FontName));
```
tomspilman: @kjpou1 - Do you happen to remember what this oddly specific font name exception was for?  It was added by you in https://github.com/dellis1972/MonoGame/pull/1 .

Seems like it should be removed and/or replaced with something that works for more than just MS Sans Serif?

Tzenchor: I'm not completely sure, but I think this is done because in windows that is the default font when a font is not found or a character is not included in the requested font.
For example, in XNA, if you use any of the fonts from the redistributable font pack and include characters beyond the 128, instead of blank characters you get the ones from "Microsoft Sans Serif".

tomspilman: > but I think this is done because in windows that is the default font when a font is not found 

That is likely to be the reason... we'll see if @kjpou1 can confirm.

That said... can this even happen under SharpFont/FreeType?

Tzenchor: I think it does not happen, i'll give it a try and if it does not get in there i'll remove it

Tzenchor: Definitely if the font does not exists `FontDescriptionProcessor.cs` does not call `SharpfontImporter.Import`.
Also `FontDescriptionProcessor.Process` has one dirty "if" that is better to remove it, it begins in line 47 and ends in line 81

Tzenchor:
All said here was revised in #4475.
Closing

kjpou1:
Yes that was because the default font for windows is "Microsoft Sans Serif" and it would try to fill in the missing characters.  Exactly as @Tzenchor has said.

PR #4474, opened 2016-01-28
Updated Dependencies to bring in updated OpenTK

dellis1972:


dellis1972:
@KonajuGames @tomspilman this brings in the updated OpenTK we merged last week (from the MG fork)

cra0zy:
I have to apply a fix to OpenTK part for window centering, so if you want you can leave Updating Deps to me, I'll do the PR tomorrow.

cra0zy:
Also have to add Joystick ID.

dellis1972:
@cra0zy ok ping me when its ready )

mrhelmut:
@cra0zy do you know what you are going to do regarding window centering? I remember that you mentioned removing the internal resizing of OpenTK. It's been several days that I'm struggling to find a fix for HiDPI displays and I believe that the internal resizing should be kept because it seems to be working correctly with HiDPI.

FYI: The problem is that ```GraphicsDevice``` and ```OpenTKGameWindow``` assume that both ```Viewport``` and backbuffer size are the same, but that's not how HiDPI works on OS X. That makes ```OpenTKGameWindow.OnResize()``` to behave wrongly for HiDPI (backbuffer should be set to ```targetBounds``` and ```Viewport``` to ```window.ClientRectangle``` which OpenTK set to the correct DPI dependent size). The other problem is ```GraphicsDevice.applyRenderTarget()``` overriding the ```Viewport``` size with the backbuffer one, there should be a DPI scaling conversion there.

I'm still having a hard time figuring all of this.

cra0zy:
> I remember that you mentioned removing the internal resizing of OpenTK

You misread something.

PR #4477, opened 2016-01-29
[Pipeline Tool] Allow compiling only from 64 bit OS

cra0zy:
This has been way overdue...

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -481,6 +481,12 @@ namespace MonoGame.Tools.Pipeline
 
         private void DoBuild(string commands)
         {
+            if(!Environment.Is64BitOperatingSystem)
+            {
+                View.OutputAppend("Build failed:  Content can only be compiled from a 64 bit OS.");
+                return;
+            }
+            
             try
             {
                 // Prepare the process.


Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -481,6 +481,12 @@ private string FindMGCB()
 
         private void DoBuild(string commands)
         {
+            if(!Environment.Is64BitOperatingSystem)
+            {
+                View.OutputAppend("Build failed:  Content can only be compiled from a 64 bit OS.");
```
tomspilman: Note that this isn't true on Mac where the pipeline still runs in 32bit because of the whole MonoMac issue.  So we need to allow for that here.

cra0zy: This blocks the compilation, not the Pipeline Tool.

That is when you press build, if your system is 32 bit, you will get that message.

tomspilman: > That is when you press build, if your system is 32 bit, you will get that message.

I understand that...  but currently Mac is 32bit and works in 32bit.  We don't want to block building on Mac.

cra0zy: Ping @dellis1972

KonajuGames: It is only really Windows that is difficult to support for 32-bit.  Mac at
moment does build and run in both 32 and 64 bit due to universal binaries.
Linux could work in 32-bit but it uses separate binaries and the 32-bit
market segment is so small.  While MonoMac is still a thing, we need to
keep it usable.

cra0zy: It doesn't check if the application runs in 32 bit or 64 bit, it checks if the OS is 64 bit...

tomspilman: > it checks if the OS is 64 bit...

I see... I misread what you posted.

I guess that should work on all platforms then right?

tomspilman: Maybe this is better at the start of MGCB.exe ?  Then it will report this issue if it is run from the Pipeline tool, the command line, Visual Studio, etc.

Maybe consider adding the same line to 2MGFX.exe as well?

cra0zy: > Maybe this is better at the start of MGCB.exe ?

We are compiling it for 64 bit: https://github.com/mono/MonoGame/blob/develop/Build/Projects/MGCB.definition#L30 so its not really even runnable from 32 bit.

cra0zy: Same goes for 2MGFX: https://github.com/mono/MonoGame/blob/develop/Build/Projects/2MGFX.definition#L27

tomspilman: > We are compiling it for 64 bit

My understanding is that the MonoMac compiler/VM ignores that setting...  it only generates 32bit code.

tomspilman: Sorry... I missed your point again.

You're saying that since `<ForceArchitecture>x64</ForceArchitecture>` is defined then the MGCB.exe and 2MGFX.exe have to be 64bit and the check of `Environment.Is64BitOperatingSystem` is pointless.

We have had people report issues in the past where mojoshader or some other native 64bit binary would not load for them.... suggesting they launched MGCB.exe or 2MGFX.exe successfully.  It could be that these are all old reports and we had been building still as "Any CPU", but i'm not sure of that.

Gonna do a test here on a 32bit tablet I have here.

tomspilman: So we do get a message from Windows on 2MGFX and MGCB, but the error message you get is sort of crappy:
```
This version of 2MGFX.exe is not compatible with the version of Windows you're running.  Check your computer's system information and then contact the software publisher.
```
Running from the Pipeline tool we currently get this message:
```
Build failed:  The specified executable is not a valid application for this OS platform.
```
It would be better if all these messages were more specific like `The MonoGame content tools only work on a 64bit OS.`.

To fix this for MGCB.exe or 2MGFX.exe we would have to stop setting the `<ForceArchitecture>x64</ForceArchitecture>` and prefer our own error checking at startup.  Thoughts @KonajuGames  @dellis1972 ?

Even better is us detecting a 32bit OS in the SDK installer and warning the developer at that time.  In fact right now I just fixed the download page to warn the issue.

KonajuGames: Change 2MGFX and MGCB to AnyCPU rather than x64 would work. It would allow
the 64-bit OS test to work.

PR #4478, opened 2016-01-30
Add specific message for trying to run MGCB or 2MGFX from 32 bit OS

cra0zy:


KonajuGames:
I'm happy with that.

Tools/2MGFX/Program.cs
```diff
@@ -7,6 +7,12 @@ public static class Program
     {
         public static int Main(string[] args)
         {
+            if (!Environment.Is64BitOperatingSystem)
```
tomspilman: Would this be safer as `Environment.Is64BitProcess`? 

https://msdn.microsoft.com/en-us/library/system.environment.is64bitprocess(v=vs.110).aspx

cra0zy: How would that be safer?

tomspilman: There are several ways to force a AnyCPU assembly to run as 32bit on a 64bit OS.  This error message would be skipped in those cases.  What we actually want to check is if the process is 64bit or not.

tomspilman:
Thanks @cra0zy !  Merging,

PR #4482, opened 2016-02-01
Fix an issue where the user is unable to change the app Icon from default MonoGame one.

dellis1972:
Because we explicitly use the Embedded Icon, unless a user builds
the library and changes it they cannot change the icon of the game.
While this might be nice from a MG point of view for marketing ;)
its not really desireable.

This commit adds support for looking for an "Icon.ico" file in
the applicaiton embedded resources first, then default to the
MonoGame one if needed.


dellis1972:
@cra0zy @mrhelmut can you test this please. 

Just add an embedded resource to a project with the resource name "Icon.ico" (not by default it adds a full namespace on it.. We might need to find a way to get that so people don't need to change it manually

cra0zy:
There is a Window.Icon property if the user wants to change the icon.

dellis1972:
@cra0zy yes there is :) But that is not available on all platforms.. which means yet another #if for the user or a partial class... wouldn't it be nice for them to be able to just include an EmbeddedResource?

cra0zy:
> @cra0zy yes there is :) But that is not available on all platforms.. which means yet another #if for the user or a partial class... wouldn't it be nice for them to be able to just include an EmbeddedResource?

If I may request something, could you modify Xamarin Studio template for DesktopGL to include Icon.ico as embedded resource from the start?

dellis1972:
@crazy yup no problem :) We should probably do the VS one too

MonoGame.Framework/Desktop/OpenTKGameWindow.cs
```diff
@@ -370,8 +370,13 @@ private void Initialize(Game game)
             {
                 if (t == null && assembly != null)
                     window.Icon = Icon.ExtractAssociatedIcon(assembly.Location);
-                else
-                    window.Icon = new Icon(Assembly.GetExecutingAssembly().GetManifestResourceStream("Microsoft.Xna.Framework.monogame.ico"));
+                else {
+                    var stream = Assembly.GetEntryAssembly().GetManifestResourceStream("Icon.ico");
```
tomspilman: Does this stream need to get disposed?

dellis1972:
Embedded resource added to both XS and VS templates. 
Code updated to that users don't need to change the name of the Icon resource.

PR #4483, opened 2016-02-01
Reverted and cleaned some ogg streamer changes

mrhelmut:
Looks like #4442 fixed the remaining OggStreamer issues but the safe check I've added was unnecessary and had a little bug which made it less thread-safe.
I've reverted some changes and kept the optimization.


tomspilman:
@dellis1972 @cra0zy @KonajuGames - Leaving this to you guys... I can't say I know anything about this really.

PR #4486, opened 2016-02-01
Effect/Shader Format

tomspilman:
This PR has a bunch of fixes to reduce the unnecessary platform differences in the MGFX format and Effect/Shader code.  This includes removal and/or reduction of `#if`s in that code and incrementing the MGFX version to 8.


tomspilman:
This too can wait until after the 3.5 release.

PR #4487, opened 2016-02-01
Fixed #1946 Turns out the Image needs to be cleared first otherwise the colors bleed.

### LINKED ISSUE #1946: Loading a PNG with alpha and rendering not working correctly on Windows GL
This issue affect MonoGame on Windows GL but may affect other platforms too. It affects MonoGame 3.0, and also 3.2.

Basically loading a texture with this code (F#):

```
use file = System.IO.File.OpenRead("testTexture.png")
this.texture <- Texture2D.FromStream(this.GraphicsDevice, file)
```

and rendering with this code:

    this.spriteBatch.Begin(SpriteSortMode.Immediate, BlendState.NonPremultiplied);
    this.spriteBatch.Draw(this.texture, Vector2.Zero, Color.White)
    this.spriteBatch.End()

will reproduce the error. This code works fine and renders fine when using the XNA4 libraries, but has weird alpha blending artefacts when using MonoGame.

Here's some renderings to illustrate the issue. First the correct renderings using the XNA libraries
Image

Now the bad renderings using MonoGame:

Image

I've tried to figure out the nature of the issue, but havne;t gotten very far besides realising that it's not a simple case of `BlendState.NonPremultiplied` being ignored or something similar. For examples, on analysis I've noticed the green colour channel blends differently than the red and blue colour channels among other odd things. I've creates a large image that illustrates the issue I've noticed in what is hopefully a useful way:

http://i.imgur.com/CEUQqm0.png

Here's the project files for reproducing the bad rendering (F# project):
http://www.qfpost.com/file/d?g=EhZh52Zjc

I also made a post about this issue on stack overflow which may be worth monitoring:
http://stackoverflow.com/questions/18320633/loading-texture-form-png-in-monogame-transaprency-issue

dellis1972:


dellis1972:
Related Issue #1946

tomspilman:
@dellis1972 - I don't think this code is used anymore.  Can you verify that?

I think it has been replaced with the code in #3537.

dellis1972:
@tomspilman still used here https://github.com/mono/MonoGame/blob/50a348a6ce7dae4e397353b7459c37d906dce587/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs#L345

tomspilman:
@dellis1972 - Ahh.... I totally missed that.  I guess we never really hooked up the new `PngReader` to any of the `Texture2D.FromStream()` calls.  Is there a reason for that @disportium ?

tomspilman:
Going to merge this for now... thanks @dellis1972 !

PR #4015, opened 2015-07-19
Remove raw asset support from ContentManager

KonajuGames:
Removes the raw asset support from ContentManager and the Normalize methods from various content readers.  Also removes DXT decompression at load time as we now have content build pipeline support for all common texture compression formats.

Raw assets should be loaded through the various methods meant for this, such as Texture2D.FromStream.

PR #4489, opened 2016-02-02
Remove legacy code to load "non-content" files

### LINKED ISSUE #3483: Remove code to load non-content files in ContentManager (after v3.3)
IMO, this line in ContentManager is fairly unhelpful:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentManager.cs#L315

``` csharp
if (string.IsNullOrEmpty(assetName))
{
    throw new ContentLoadException("Could not load " + originalAssetName + " asset as a non-content file!", ex);
}
```

It just covers up the original exception (unless you look at `InnerException`, but not everyone knows to do that).

If the non-content fallback file can't be found, couldn't we just re-throw the original exception?

tgjones:
Fixes #3483.

I think it's about time we removed this code; it's nothing but confusing, unless anyone still relies on it?

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -294,59 +294,27 @@ namespace Microsoft.Xna.Framework.Content
 				}
 			}
 			
-			Stream stream = null;
-			try
-            {
-				//try load it traditionally
-				stream = OpenStream(assetName);
+			var stream = OpenStream(assetName);
 
-                // Try to load as XNB file
-                try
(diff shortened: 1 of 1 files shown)


tomspilman:
Let's wait for 3.5 to get out first then merge this.

tgjones:
Sounds good to me.

KonajuGames:
Same as my #4015. :)

tgjones:
Oops :) I'll close this one, yours is more complete.

PR #4490, opened 2016-02-02
WinDX: Fixed GDI leak in GraphicsAdapter.CurrentDisplayMode

MichaelDePiazzi:
Today I got a crash report with the following exception:
```
Type: System.ArgumentException
Message: Parameter is not valid.
Source: System.Drawing
Stack Trace:
at System.Drawing.Graphics.GetHdc()
at Microsoft.Xna.Framework.Graphics.GraphicsAdapter.get_CurrentDisplayMode()
...
```
Turns out there was a bit of a GDI leak in `GraphicsAdapter.CurrentDisplayMode`. The main problem was that a `Graphics` object was being created (via `Graphics.FromHwnd(IntPtr.Zero)`) but wasn't being disposed. This meant that the resources weren't being released until the garbage collector kicked in. So calling `CurrentDisplayMode` a lot could eventually lead to the above crash.

The fix was as simple as wrapping the `Graphics` object in a `using` so that it got disposed.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.cs
@@ -96,7 +96,10 @@ namespace Microsoft.Xna.Framework.Graphics
                 using (var graphics = System.Drawing.Graphics.FromHwnd(IntPtr.Zero))
                 {
                     var dc = graphics.GetHdc();
-                    return new DisplayMode(GetDeviceCaps(dc, HORZRES), GetDeviceCaps(dc, VERTRES), SurfaceFormat.Color);
+                    int width = GetDeviceCaps(dc, HORZRES);
+                    int height = GetDeviceCaps(dc, VERTRES);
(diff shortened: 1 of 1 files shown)


tomspilman:
Good catch... it also looks like we should be freeing the dc handle:

https://msdn.microsoft.com/en-us/library/9z5820hw(v=vs.110).aspx

Mind adding a `graphics.ReleaseHdc(dc);` in there too and giving it a test?

MichaelDePiazzi:
@tomspilman
It seems `Dispose` was cleaning up properly. But you're right, the documentation explicitly says that `ReleaseHdc` must be called after calling `GetHdc`. So I've updated the code to do this and tested it as requested.

tomspilman:
Thanks @MichaelDePiazzi ... merging!

MichaelDePiazzi:
Cool, thanks Tom :thumbsup:

CartBlanche:
Nice!!!

PR #4491, opened 2016-02-02
Removed state check to force stop playing instances

mrhelmut:
This PR removes the ```SoundEffectInstance.State``` check from the disposing mechanism of the ```SoundEffectInstancePool```. This was causing problem to the OpenAL implementation because its ```SoundEffectInstance.PlatformGetState()``` was internally updating the state, which was making the pool to think that some instances were wrongly stopped, preventing them from freeing their OpenAL source.
This rare scenario ended up in an AL crash in ```SoundEffect.Dispose()``` because the sound buffer was still bound to a ```SoundEffectInstance```.

This change doesn't affect other implementations.

Ping @dellis1972 :)


cra0zy:
You didn't setup your git correctly, run commands:
```
git config --global user.email <email connected with github>
git config --global user.name mrhelmut
```

mrhelmut:
Ah thx, my future commits should be fixed.

PR #4492, opened 2016-02-02
[SDL] Fix window not getting centered after leaving fullscreen

cra0zy:
I also made sure this works correctly with multiple displays :)


tomspilman:
Thanks... merging!

Issue #4479, opened 2016-01-31
Texture2D.FromStream cannot load tga files


trommlbomml:
Reproducable with no problems in MonoGame 3.4. Take a tga file and try to load it with Texture2D.FromStream. You will get a SharpDX exception:

HRESULT: [0x88982F50], Module: [SharpDX.WIC], ApiCode: [WINCODEC_ERR_COMPONENTNOTFOUND/Componentnotfound], Message: Component not found

I need FromStream() as I have a system to load data asynchronous and transparent from local and server, so I need byte arrays.

Any chance of fixes? My workaround is using png as this seems to work properly.

KonajuGames:
PNG and JPG are the two image formats supported by FromStream, as they are more widely supported by the host systems.  Support for TGA could be added, but it would need to be written in C# as most of the host platforms do not have native support for TGA.

trommlbomml:
Ok this is reasonable. I just fall over it because I tried to load the sponza model from crytek as a first big model loading test.

When just png and jpg are supported I would like at least to have a better error msg of type NotSupportedException if possible.

tomspilman:
We should add some unit tests for `FromStream` specifically including some failure cases.

trommlbomml:
I think even document this method that these two formats are supported only would help a lot. Unit-Test is also a good idea of course.

Tzenchor:
With the changes done in the merged pull I think this can be closed

tomspilman:
Thanks @Tzenchor !

PR #4493, opened 2016-02-02
Texture2D FromStream tests and message

Tzenchor:
First pass at adding a tests to `Texture2D.FromStream` as commented in #4479.
Enveloped `Texture2D.FromStream` in a try-catch block, inside the catch I am throwing a new NotSupportedException with the original one inside and the message "Image type not supported"
Add test class `Texture2DNonVisual` to test project with tests for filetypes: bmp, dds, gif, jpg, png, tiff and tga. All assets reused from other tests.

I have tested this in Windows with the Windows solution and adapted the tests to it (only the tga image provokes an exception), let's see if the other platforms have the same capabilities.


tomspilman:
Good first pass here @Tzenchor .

You should also add the `Texture2DNonVisualTest.cs` to the XNA project...

 https://github.com/mono/MonoGame/blob/develop/Test/MonoGame.Tests.XNA.csproj

You can then run the test under XNA and verify we get the same results as XNA does.

Tzenchor:
I have added the test class to the XNA test project, but i can't run it as i don't the tooling installed.
Also i have modified the tests to check that the created `Texture2D` at least has the same size than the original image.

I don't know if i can add anything more here apart from checking some pixels.

tomspilman:
> I have added the test class to the XNA test project, but i can't run it as i don't the tooling installed.

Ah... we need to find someone to run that for you and verify we're getting the same output with XNA.

Tzenchor:
I have just added a new test for a dds file that does not contain a texture (in this case it contains a cubemap).
Also have added documentation comments to `Texture2D.cs`, please review them for typos or inaccuracies

KonajuGames:
It would appear that XNA throws an `InvalidOperationException` rather than `NotSupportedException`.
http://xboxforums.create.msdn.com/forums/t/79565.aspx
I don't have XNA installed either, so I went hunting.

Tzenchor:
> It would appear that XNA throws an InvalidOperationException rather than NotSupportedException.

Good to know, i'll update later after finishing some changes for another pull.
Also while investigating further I found this other jewel [XNA Texture2D.FromStream documentation](https://msdn.microsoft.com/en-us/library/ff434103.aspx?f=255&MSPPError=-2147217396)
It appears that XNA can only load .gif, .png and .jpg from a Stream

KonajuGames:
Correct in that XNA only supported those three formats. Supporting more without requiring a dependency on a large image loading library isn't a problem though. Think of GIF, PNG and JPG as the minimum set, and if we can support other common formats without adding a large amount of code, then there is little reason not to.

tomspilman:
@Tzenchor - FYI...  i'm switching out to a new build server, so the errors you are seeing might not be your fault.   I'll let you know when I'm done and if I need fixes from you.

tomspilman:
@Tzenchor - Ok... the build server is working and things were fine with your changes.

Tzenchor:
This morning I run the tests against XNA and all tests passed correctly after adding the missing assets to the XNA project

KonajuGames:
Excellent. I'll merge this now.

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -27,28 +29,58 @@ public Rectangle Bounds
...
+        /// <param name="graphicsDevice"></param>
+        /// <param name="width"></param>
+        /// <param name="height"></param>
         public Texture2D(GraphicsDevice graphicsDevice, int width, int height)
             : this(graphicsDevice, width, height, false, SurfaceFormat.Color, SurfaceType.Texture, false, 1)
         {
         }
-
+        /// <summary>
+        /// Creates a new texture of a given size with a surface format and optional mipmaps
```
tomspilman: @Tzenchor - In the future remember to put periods at the end of the sentence in the reference docs.  I didn't get time to review this PR before it was merged else I would have caught this then.

PR #4495, opened 2016-02-03
Clean up some code that potentially was reflected.

dellis1972:
After an incident with a contributor and auditing their contributions we
have a strong suspicion that some bits of their code came from either a
copyrighted source or from a reflection tool.

While we do not have definitive proof of this and cannot be certain of
the code's origin, we feel it is just better for the project to remove
those chunks of code.

 The code in question is in the following functions:

   - BoundingFrustum.Intersects(ref Ray ray, out float? result)
   - Vector4.Transform(ref Vector2 value, ref Quaternion rotation, out Vector4 result)
   - Vector4.Transform(ref Vector3 value, ref Quaternion rotation, out Vector4 result)
   - Vector4.Transform(ref Vector4 value, ref Quaternion rotation, out Vector4 result)

We hope that someone in the community can contribute their own implementations
to fix these holes soon (of course without reading or reusing any of the code removed in this PR).

We will also follow up with an update to the existing CONTRIBUTING.md to help further clarify what we require from them.

### DIFF
--- a/MonoGame.Framework/BoundingFrustum.cs
+++ b/MonoGame.Framework/BoundingFrustum.cs
@@ -471,65 +471,7 @@ namespace Microsoft.Xna.Framework
                     result = 0.0f;
                     return;
                 case ContainmentType.Intersects:
-                    
-                    // TODO: Needs additional test for not 0.0 and null results.
-
-                    result = null;
-
-                    float min = float.MinValue;
-                    float max = float.MaxValue;
-                    foreach (Plane plane in this._planes)
-                    {
-                        var normal = plane.Normal;
-
-                        float result2;
-                        Vector3.Dot(ref ray.Direction, ref normal, out result2);
-
-                        float result3;
-                        Vector3.Dot(ref ray.Position, ref normal, out result3);
-
-                        result3 += plane.D;
-
-                        if ((double)Math.Abs(result2) < 9.99999974737875E-06)
-                        {
-                            if ((double)result3 > 0.0)
-                                return;
-                        }
-                        else
-                        {
-                            float result4 = -result3 / result2;
-                            if ((double)result2 < 0.0)
-                            {
-                                if ((double)result4 > (double)max)
-                                    return;
-                                if ((double)result4 > (double)min)
-                                    min = result4;
-                            }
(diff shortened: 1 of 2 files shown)


amerkoleci:
SharpDX  has implementation for thoose, we can take code from there

tomspilman:
> SharpDX has implementation for thoose, we can take code from there

Generally we would prefer to get code that isn't tired to another project's license.  In particular for how simple and little code this is, It seems unnecessary to do that.

amerkoleci:
Dude SharpDX is MIT licensed and anyway if you write it from scratch it Will be the same, we can check also System.Numerics.Vectors from corefx

tomspilman:
> Dude SharpDX is MIT licensed and anyway if you write it from scratch it Will be the same

I understand, but disagree.  We don't want our list of licenses to grow even further for something so trivial.

> we can check also System.Numerics.Vectors from corefx

Actually we've been keeping an eye out for that.  We're hoping to eventually replace all of our math implementation with `System.Numerics.Vectors`, but they lack of the `ref` optimization which is holding us back at the moment.  It is an interesting read... https://github.com/dotnet/corefx/issues/157#issuecomment-99787924.

flibitijibibo:
I just had a look at this and it seems like the Vector4 work was _already_ taken from SharpDX, based on who had what first:

https://github.com/sharpdx/SharpDX/blob/master/Source/SharpDX.Mathematics/Vector4.cs#L1005

BoundingFrustum wouldn't surprise me though, especially with lines like this one:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/BoundingFrustum.cs#L493

Seems... really specific. Not Epsilon?

tomspilman:
> Seems... really specific. Not Epsilon?

Things like that are what tipped us off and made us nervous about those particular contributions.

tomspilman:
@dellis1972 - My rewrite of the contributing doc.... https://github.com/mono/MonoGame/pull/4500.

tomspilman:
@KonajuGames - If you are good with this feel free to merge it.

KonajuGames:
Merging.

PR #4501, opened 2016-02-05
Fix mouse coordinate for DesktopGL and windowed mode

mrhelmut:
```OpenTK.Input.Mouse.GetCursorState()``` and ```OpenTK.Input.Mouse.GetState()``` both return coordinates relatively to the drawing surface. The current mouse code assume them to be relative to the display, and not the drawing surface, and tries to adjust them with the window's position. This was causing mouse coordinates to be wrong when using windowed.

This PR fixes this behavior.

However, OpenTK doesn't provide methods to get the mouse position relatively to the display, and therefore ```Microsoft.Xna.Framework.Input.Mouse.GetState(GameWindow window)``` can't return coordinates specific to a ```GameWindow``` and will always return them relatively to the active OpenTK surface.

@cra0zy @dellis1972

### DIFF
--- a/MonoGame.Framework/Input/Mouse.cs
+++ b/MonoGame.Framework/Input/Mouse.cs
@@ -140,9 +140,9 @@ namespace Microsoft.Xna.Framework.Input
 #elif DESKTOPGL || ANGLE
 
             var state = OpenTK.Input.Mouse.GetCursorState();
-            var pc = ((OpenTKGameWindow)window).Window.PointToClient(new System.Drawing.Point(state.X, state.Y));
-            window.MouseState.X = pc.X;
-            window.MouseState.Y = pc.Y;
+            
+            window.MouseState.X = state.X;
+            window.MouseState.Y = state.Y;
 
             window.MouseState.LeftButton = (ButtonState)state.LeftButton;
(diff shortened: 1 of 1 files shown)


dellis1972:
Can we change OpenTK to get Display relative ? I'm not 100% what XNA does
we should try to match that if possible.

mrhelmut:
XNA doesn't have a ```Mouse.GetState(GameWindow window)```, only ```Mouse.GetState()``` which returns coordinates relatively to the active ```GameWindow```. Considering this and OpenTK being replaced, I felt like this downside would be minor compared to what it fixes.

This PR fixes ```Mouse.GetState()``` to match XNA (it calls ```Mouse.GetState(GameWindow window)``` internally).

It's up to you, but I think I won't take a look at what OpenTK does internally. :)

dellis1972:
ok just checking :)

cra0zy:
I'm fine with this, I had to do the same thing for new SDL backed.

dellis1972:
@mrhelmut as always thanks :)

cra0zy:
I feel so blind right now, I thought the GetCursorState line was changed to GetState...

Anyway this broke mouse position on DesktopGL... well easy fix at least :\

PR #4503, opened 2016-02-05
Fix Content Building on Mac

dellis1972:
At the moment 64 bit mono is not really available,
it is still in preview. So we need to disable this
64Bit check for Mac (and Linux) until its in stable.


tomspilman:
Thanks @dellis1972 !

PR #4505, opened 2016-02-06
Fix inaccessible GetGuid for OpenTK on Linux / DesktopGL

hach-que:
On Linux / DesktopGL, this method is inaccessible under OpenTK, and results in:

```
Input/Joystick.OpenTK.cs(29,44): error CS0122: `OpenTK.Input.Joystick.GetGuid(int)' is inaccessible due to its protection level
```

when building either the Linux or WindowsGL projects.  On these platforms, default to an empty string instead.


cra0zy:
That sounds like you don't have the newer OpenTK.dll for some reason. What you've done here is not the correct fix.

cra0zy:
Also I just tested it, I can get ID without any problems.

dellis1972:
probably a submodule update needs to be done

PR #4506, opened 2016-02-06
Fix casing in MonoGame.Tests project definition

hach-que:
When building on case sensitive platforms, the build fails because the MonoGame.Tests project has a casing that differs from the filenames on disk, which results in these build errors:

```
/usr/lib/mono/xbuild/12.0/bin/Microsoft.Common.targets: error : Cannot copy /srv/leases/workingcopy-26560/repo/monogame/Test/Assets/Fonts/MotorWerk.xnb to /srv/leases/workingcopy-26560/repo/monogame/Test/bin/Windows/AnyCPU/Release/Assets/Fonts/MotorWerk.xnb, as the source file doesn't exist.
797		/usr/lib/mono/xbuild/12.0/bin/Microsoft.Common.targets: error : Cannot copy /srv/leases/workingcopy-26560/repo/monogame/Test/Assets/Textures/Logo555.bmp to /srv/leases/workingcopy-26560/repo/monogame/Test/bin/Windows/AnyCPU/Release/Assets/Textures/Logo555.bmp, as the source file doesn't exist.
798		/usr/lib/mono/xbuild/12.0/bin/Microsoft.Common.targets: error : Cannot copy /srv/leases/workingcopy-26560/repo/monogame/Test/Assets/Textures/Logo565.bmp to /srv/leases/workingcopy-26560/repo/monogame/Test/bin/Windows/AnyCPU/Release/Assets/Textures/Logo565.bmp, as the source file doesn't exist.
799	
```

Instead, set the casing in the definition to match the name of the files on disk.

### DIFF
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -754,7 +754,7 @@
     <Content Include="Assets\Fonts\Lindsey.xnb">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="Assets\Fonts\MotorWerk.xnb">
+    <Content Include="Assets\Fonts\Motorwerk.xnb">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
     <Content Include="Assets\Fonts\QuartzMS.xnb">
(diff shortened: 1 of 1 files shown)


Tzenchor:
Sorry about that, when i added those files for the tests I didn't have in mind that possibility.
If you have changed the project definitions you have to change the names in the tests in `TextureImporterTests.cs` or the tests will fail.
If you just rename the files it should work directly

KonajuGames:
Merged.

MichaelDePiazzi:
Whoops! One of those was my bad also. I'll definitely be more wary of this in future.

hach-que:
This is probably a reason why MonoGame upstream needs a Linux build server - this issue has been around for a couple of weeks I think, but neither Mac or Windows exhibit the problem because they don't have case sensitive filesystems.

tomspilman:
> This is probably a reason why MonoGame upstream needs a Linux build server 

I had plans for that, but once we started building Linux on Mac I figured it was no longer an issue.  Should I still consider having a dedicated Linux build machine?

hach-que:
Yeah. There are some issues like this one that you'll only see on Linux.

PR #4507, opened 2016-02-06
refactor of wiki page to doc page

ralfjansson:
Moved text from wiki concerning Android manifest requirements and added
to Platform specific notes for Android.

My first PR ever. I hope this works.

### DIFF
--- a/Documentation/android.md
+++ b/Documentation/android.md
@@ -36,6 +36,25 @@ On the `Application Application` page, you will find `Minimum Android version` (
 <img src="images/android_xs_minimum_framework.png"/>
 </p>
 
+
+## Android Manifest Requirements
+
+### OpenGL ES 2.0 Support
+
+MonoGame uses OpenGL ES 2.0. Google requires the following to be added to AndroidManifest.xml in order for the Market to hide the game from devices that do not have support for OpenGL ES 2.0.
+
+
+```
+<!-- Tell the system this app requires OpenGL ES 2.0. -->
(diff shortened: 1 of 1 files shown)


KonajuGames:
That worked well.  Thanks @ralfjansson.  Merging.

tomspilman:
@ralfjansson - You should probably clear out that info from the wiki page now or do something to mark it as now in the main docs.

ralfjansson:
Ok. Forgive me for asking, but that means I need to fork the wiki as it is in another Project?

tomspilman:
> but that means I need to fork the wiki as it is in another Project?

No... you can edit the wiki directly from github site.  You should see an `Edit` button to the right of the title on each page.

ralfjansson:
Ok. Got it. I changed the old wiki page as you suggested by adding a commonent on "stale information" as has been done on some other wiki pages. The change seam to have taken effect at once?

tomspilman:
> adding a commonent on "stale information"

Actually I think we remove the information entirely and just link to the new doc.  Just want no confusion from duplicate info floating out there.  The wiki retains history... so if we ever really need the old info back we can get it.

ralfjansson:
Ok. Fixed that. Now there is only a link to the new documentation on that page.

tomspilman:
Yep... perfect.

https://github.com/mono/MonoGame/wiki/Android-Manifest-Requirements

Thanks @ralfjansson !

PR #4509, opened 2016-02-06
Change the XS ProjectTemplates to use the Local Installed Assemblies.

dellis1972:
Using the Nuget packages for the templates was a hack to work around
issues with Assembly resolution in Xamarin Studios/MonoDevelop.
With the changes to the MonoGameMSBuildImports class we can now
correctly inject the HintPath to pick up the local assemblies. This
will allow develop builds to actually reference the installed assemblies
rather than an out of date nuget package.

We do this by having a new msbuild .props file

	MonoGame.Common.props

This file is responsible for finding the correct Instllation path,
all the VS templates have been updated to use this new file and the
old settings to <MonoGameInstallDirectory> have been removed.
The XS templates will also get this new import when they are first loaded.

NOTE : the Import MUST be the first item in the project. This is to the
directory is resolved BEFORE the assemblies are loaded.


KonajuGames:
> The down side is that if a user upgrades mono that link will disappear and they will need to run ... to re-create the symlink.

Is that because we install into mono's Framework directory?  I don't see that being a good thing if updating mono breaks MonoGame's install.  Why do we need our libraries in mono's install location?  Our assemblies can't live in our install location?

dellis1972:
@KonajuGames We don't install the assemblies /other stuff onto Mono but into 
`/Library/Frameworks/MonoGame.framework`
which is what we are supposed to do. We also put a symlink in
`/Library/Frameworks/Mono.framework/External/`
which is what we are supposed to do.. This allows us to resolve our Builder targets correctly :)

The problem is that on Mac and Linux $(MSProgramFiles32} is empty so we have to use ${MSBuildExtensionsPath} to attempt to resolve the monogame install directory. bearing in mine that we use `$(MSProgramFiles32}\MonoGame\v3.0\Assemblies` to resolve the assemblies on Windows. On Mac ${MSBuildExtensionsPath} points to `/Library/Frameworks/Mono.framework/Versions/Current/lib/mono/xbuild` which DOESN'T have a MonoGame symlink... we can create one but because its in the *Current* if the user upgrades, the new mono won't have it. So they will need to relink...

I just saw a bug anyway this will break Linux, so I need to rework it again. And I'll have a think about a better way to do this.

dellis1972:
@KonajuGames ok I think I have another cunning plan :) Should work on all the platforms too

KonajuGames:
> @KonajuGames ok I think I have another cunning plan :)

Does it involve turnips?

dellis1972:
@KonajuGames lol :) 

Nope , enter the new MonoGame.Common.props file to save the day and fix ALL of the problems

dellis1972:
Installers have been updated to deploy the new file as well

Installers/default.build
```diff
@@ -106,10 +107,11 @@
          </fileset>
       </copy>
       <copy todir="root/Library/Frameworks/MonoGame.framework/v${buildNumber}/Assemblies/DesktopGL">
-         <fileset basedir="../MonoGame.Framework/bin/DesktopGL/AnyCPU/Release">
+         <fileset basedir="../MonoGame.Framework/bin/Linux/AnyCPU/Release">
```
KonajuGames: Was this change from "DesktopGL" to "Linux" intended?  I thought we were changing the other way.

dellis1972: It was.. Currently we do not build a "DesktopGL" platform we only build Linux and WindowsGL. As a result we currently get no Assemblies in that directory in the installer even though we have projects trying to reference them. The Windows Installer uses the WindowsGL binaries for that folder, we should do the same and Linux=WindowsGL now.

dellis1972: also if you look at https://github.com/dellis1972/MonoGame/blob/addinfix/Installers/default.build#L169 that is exactly what the linux installer does atm

dellis1972:
@KonajuGames this is all fixed up now. 

@cra0zy  a test on linux would be helpful it should pick up the assemblies in the installation directory

cra0zy:
Does not work :(

nonuget

.csproj file:
```xml
<?xml version="1.0" encoding="utf-8"?>
<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Import Project="$(MSBuildExtensionsPath)\MonoGame\v3.0\MonoGame.Common.props" Conditon="Exists('$(MSBuildExtensionsPath)\MonoGame\v3.0\MonoGame.Common.props')" />
  <PropertyGroup>
    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
    <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
    <ProjectGuid>{010EDF28-0015-4E5E-BBF2-57F1302755E6}</ProjectGuid>
    <OutputType>Exe</OutputType>
    <RootNamespace>nonugettest</RootNamespace>
    <AssemblyName>nonugettest</AssemblyName>
    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
    <MonoGamePlatform>DesktopGL</MonoGamePlatform>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x86' ">
    <DebugSymbols>true</DebugSymbols>
    <DebugType>full</DebugType>
    <Optimize>false</Optimize>
    <OutputPath>bin\Debug</OutputPath>
    <DefineConstants>DEBUG;</DefineConstants>
    <ErrorReport>prompt</ErrorReport>
    <WarningLevel>4</WarningLevel>
    <ConsolePause>false</ConsolePause>
    <PlatformTarget>x86</PlatformTarget>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|x86' ">
    <DebugType>full</DebugType>
    <Optimize>true</Optimize>
    <OutputPath>bin\Release</OutputPath>
    <ErrorReport>prompt</ErrorReport>
    <WarningLevel>4</WarningLevel>
    <ConsolePause>false</ConsolePause>
    <PlatformTarget>x86</PlatformTarget>
  </PropertyGroup>
  <ItemGroup>
    <Reference Include="MonoGame.Framework">
      <HintPath>$(MonoGameInstallDirectory)\MonoGame\v3.0\Assemblies\DesktopGL\MonoGame.Framework.dll</HintPath>
    </Reference>
  </ItemGroup>
  <ItemGroup>
    <Compile Include="Game1.cs" />
    <Compile Include="Program.cs" />
    <Compile Include="Properties\AssemblyInfo.cs" />
  </ItemGroup>
  <ItemGroup>
    <None Include="Icon.png" />
    <MonoGameContentReference Include="Content\Content.mgcb" />
  </ItemGroup>
  <ItemGroup>
    <Content Include="OpenTK.dll.config">
    </Content>
  </ItemGroup>
  <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
  <Import Project="$(MSBuildExtensionsPath)\MonoGame\v3.0\MonoGame.Content.Builder.targets" />
</Project>
```

dellis1972:
@cra0zy the bit red x is a bug in XS/MD I think as I get the same here..
But for me it builds.. does it build for you?

dellis1972:
@cra0zy  also is this path correct https://github.com/mono/MonoGame/pull/4509/files#diff-d7f10d5b1431ea5462b796e74d4d1e66R3

cra0zy:
> @cra0zy the bit red x is a bug in XS/MD I think as I get the same here..
> But for me it builds.. does it build for you?

Oh, it does build.

I did try to setup some MG reference using similar method before, and got the red X, but I never realized that it's just MD/XS being dumb.

> @cra0zy also is this path correct https://github.com/mono/MonoGame/pull/4509/files#diff-d7f10d5b1431ea5462b796e74d4d1e66R3

It's correct.

One thing left, it doesn't reference OpenTK and NVorbis.

dellis1972:
It should just pull those in as dependencies of MonoGame.Framework.dll ...
at least it does on mac.

cra0zy:
> It should just pull those in as dependencies of MonoGame.Framework.dll ...
> at least it does on mac.

Oh yea, it does, good thing they fixed that, that was broken at one point if I remember correctly.

I give this PR a :+1:

cra0zy:
@dellis1972 You have a merge conflict.

dellis1972:
@cra0zy thanks for the heads up. all fixed :)

dellis1972:
@tomspilman @KonajuGames this should be good to

IDE/MonoDevelop/MonoDevelop.MonoGame.Android/templates/MonoGameAndroidProject.xpt.xml
```diff
@@ -22,17 +22,15 @@
...
 				AndroidResgenClass="Resource"
-				TargetFrameworkVersion="MonoAndroid,Version=v4.4" />
+				TargetFrameworkVersion="MonoAndroid,Version=v5.0" />
 			<References>
 				<Reference type="Gac" refto="System, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e" />
 				<Reference type="Gac" refto="System.Xml, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e" />
 				<Reference type="Gac" refto="System.Core, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e" />				
 				<Reference type="Gac" refto="Mono.Android" />
-				<Reference type="Gac" refto="OpenTK" />
+				<Reference type="Gac" refto="OpenTK-1.0" />
```
KonajuGames: OpenTK-1.0 now works properly on Android and iOS?  The API differences in OpenGL ES bindings between the two implementations on the different platforms were all sorted out?

dellis1972: We've been building iOS and Android frameworks against OpenTK-1.0 for a while now. The templates were out of date.

https://github.com/mono/MonoGame/blob/develop/Build/Projects/FrameworkReferences.definition#L8
https://github.com/mono/MonoGame/blob/develop/Build/Projects/FrameworkReferences.definition#L47

KonajuGames:
Merging.

PR #4510, opened 2016-02-06
Update the build script to download the Archived Android and OUYA

dellis1972:
binaries from the MacBuild bot. If it can.

We'll need to update the MacBot to archive these files.


dellis1972:
@tomspilman  we'll need to make the changes to the bots we talked about before we pull this in. Although it shouldn't fail atm.

tomspilman:
> We'll need to update the MacBot to archive these files.

We build Android and OUYA on Windows... so I guess that makes this more complicated right?

dellis1972:
@tomspilman check your email about this one. Also I found that the Mac installer is missing the following assemblies

Android
OUYA
DesktopGL

One of the other PR's fixes DesktopGL  (#4509)

dellis1972:
Closing PR

PR #4514, opened 2016-02-07
Added a Visual Studio 2015 Project Template for Windows 10 UAP in VB

vogles:
Recently a few people have been trying to get MonoGame up and running using VB. This project template provides an easy way to do that.


tomspilman:
@KonajuGames @dellis1972 - Thoughts?

I don't see a problem with including templates for VB or F# in the future as long as we can support them easily?

@hach-que - Would this eventually fit in to your project creation tool?  Or is it tied to C# specifically?

dellis1972:
I have no problem with this. We probably should look at supporting VB, F# out of the box but we need someone willing to do the templates :)

tomspilman:
> We probably should look at supporting VB, F# out of the box 

We should probably wait to do most of that work until after we move to https://github.com/mono/MonoGame/pull/4295.  If not we end up with a ton of duplicated stuff like we have now.

hach-que:
> @hach-que - Would this eventually fit in to your project creation tool? Or is it tied to C# specifically?

The project creation tool just uses Protobuild underneath to do the generation.  We'd need to add support for generating VB projects (there's already experimental support for F#), but I'm fine with that if someone wants to write the templates to do so.

tomspilman:
>  We'd need to add support for generating VB projects (there's already experimental support for F#), 

Sounds good then... seems very doable to support in the future.

When we make the change over to your tool I want to push for a lot more templates.  Other .NET languages, more complex sample templates, pipeline extension templates, etc.  I want to replace all the needless duplicate templates we have now with much more variety.

KonajuGames:
Merging this.  Thanks @vogles.

PR #4516, opened 2016-02-08
[Linux Installer] Disable copying of nvtt libs

cra0zy:
This was needed because when I compiled nvtt libs, I didn't know about rpath, and left it to default(which was my home path), @KonajuGames recompiled them so this is no longer needed.


dellis1972:
@cra0zy  looks ok. but it looks like the bot was down. can you update the branch so it runs?

cra0zy:
@dellis1972 done

tomspilman:
Sorry guys...  had to fix a few things on the build server.  I restarted the build now.

tomspilman:
This good to go @cra0zy @dellis1972 ?

cra0zy:
:+1:

Issue #4460, opened 2016-01-25
Pipeline filter view doesn't handle warnings well


tomspilman:
Sometimes tools like shader compilers will dump out `warning` messages like error messages.  Here is an example of that:

https://gist.github.com/tomspilman/f502f209178e3337bc2e

The filter view in the Pipeline tool doesn't correctly detect that as the output of one build content item.

tomspilman:
@cra0zy  @nkast ?

nkast:
I tried to replicate that by creating a new fx and add a typo.
for example 'outputt.Position = mul(input.Position, WorldViewProjection);' 

What i got was a double escaped string!
"P:\\\\TestPipeline\\\\NewEffect.fx(28,2): error X3004: undeclared identifier 'outputt'\n"
This comes from: https://github.com/mono/MonoGame/blob/develop/Tools/2MGFX/EffectObject.hlsl.cs#L53-L57
It's not possible to parse this type of filenames, any idea of how we can handle this bug?

tomspilman:
> I tried to replicate that by creating a new fx and add a typo.

Well a warning isn't the same as an error...  I will have to give you some specific code to trigger a warning in the HLSL compiler.

> What i got was a double escaped string!

Hum....  we get that string from the stdout/stderr streams.  I guess those streams are generating these double slashes.

> It's not possible to parse this type of filenames, any idea of how we can handle this bug?

I would say stop trying to parse it as a file name... just accept anything before the parenthesis and colon.

nkast:
> I would say stop trying to parse it as a file name... just accept anything before the parenthesis and colon.

Yeah, we have already relaxed filename because of win/unix. This probably already works.

cra0zy:
So problems that are present are:
 - errors and warning that fx files throw out aren't registered
 - if an error occurs it still shows fx file as success

As to what we should do for filter view, I suggest that when we parse:
```
C:\\Users\\cra0z\\Desktop\\gmjosack-xna-2d-shader-examples-07cfe1a5aafb\\ShaderTests\\ShaderTestsContent\\effects\\Bevels.fx(5,5): error X3000: unrecognized identifier 'flot4'
C:\\Users\\cra0z\\Desktop\\gmjosack-xna-2d-shader-examples-07cfe1a5aafb\\ShaderTests\\ShaderTestsContent\\effects\\Bevels.fx(5,11): error X3000: unrecognized identifier 'color'
```

we shorten it to:
```
[...]

PR #4517, opened 2016-02-08
[Pipeline] FilterView for Shader warnings/errors

nkast:
#4460


tomspilman:
@cra0zy - This look ok to you?

nkast:
@tomspilman Can you test it with your shader? All I did was simulating by feeding it your output and also testing it with a simple typo so far. 

The parser also keep the filename so I had to normalize it. Not sure If we still need that but it's good to keep it around.

@cra0zy, 
According to the internets all chars are valid in a unix filename even ':', so technically all this will fail in some cases under Linux/Mac/etc.

cra0zy:
> According to the internets all chars are valid in a unix filename even ':', so technically all this will fail in some cases under Linux/Mac/etc.

":" char is not allowed on Mac. 

Technically speaking it's not the OS that blocks the character, it's the filesystem.

I am fine with it being the way you wrote it, ":" should never be used in a filename.

cra0zy:
PS. I like using https://regex101.com/ to test and write regex.

nkast:
For example I could do without the _reFileError and just strict _reBuildAsset to match [^\n\r:] instead of [.]. The rest lines from tom's output would then be handled by the finall check at https://github.com/nkast/MonoGame/blob/PipelineFilterOutputShaderWarnings/Tools/Pipeline/Common/OutputParser.cs#L157-L162

cra0zy:
Tested it, and it works.

tomspilman:
Looks good to me guys... merging!  Thanks again @nkast  and @cra0zy !

PR #4522, opened 2016-02-09
ContentReader, use Lz4DecoderStream directly

nkast:
We pass the Lz4DecoderStream to the ContentReader instead of 2-step decompressing in MemoryStream. Improve performance since we avoid extra memory allocation and memcopy.

The only difference is that we no longer check the decompressedSize but that's something we can check better on the contentPipeline (compress & decompress to check data  integrity). However the Lz4 code has been proven to be fine so far.

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -498,17 +498,10 @@ namespace Microsoft.Xna.Framework.Content
                 }
                 else if (compressedLz4)
                 {
-                    // Decompress to a byte[] because Windows 8 doesn't support MemoryStream.GetBuffer()
-                    var buffer = new byte[decompressedSize];
-                    using (var decoderStream = new Lz4DecoderStream(stream))
-                    {
-                        if (decoderStream.Read(buffer, 0, buffer.Length) != decompressedSize)
-                        {
-                            throw new ContentLoadException("Decompression of " + originalAssetName + " failed. ");
-                        }
-                    }
-                    // Creating the MemoryStream with a byte[] shares the buffer so it doesn't allocate any more memory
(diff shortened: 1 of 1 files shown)


tomspilman:
Interesting...  lets leave this for after 3.5 is shipped just in case.

I think it is safe to assume that the `Lz4DecoderStream` is well written to internally buffer the decoding in an efficient way.  So it might be little impact to performance, but a good savings in memory usage during the loading process.

Does `Lz4DecoderStream` allow for seek?  That may be needed in the future and it would be good to know now if it is a limitation.

nkast:
| Does Lz4DecoderStream allow for seek? That may be needed in the future and it would be good to know now if it is a limitation. 

Haven't checked but I don't expect it to. An importer has to do it's own caching in that case. Importers work mostly with ContentReader and not with the stream (although the `.BaseStream` is accessible). 
Potentially Content might come from net streams (.WEB platform? live/customized content?) so I'd like for stock importers not to depend on seekable streams.

tomspilman:
Just thinking about things like partial loads of mips or automatic content reloading in the future.  We can deal with it then if it comes up.

KonajuGames:
I think it's a reasonable expectation that content readers should never
seek backwards. A seek forwards would be a read and discard if it was
necessary.

dellis1972:
If I don't get to it first can someone test this change on android (and iOS)? need to make sure we don't regress the loading times on those platforms.

KonajuGames:
I believed the compressed stream is still loaded into memory in full before
any decompressing occurs, so it shouldn't have any negative effect on load
times.

tomspilman:
FYI...  I just tested this PR on Vita and it seemed to work great.

KonajuGames:
Merging.  Thanks @nkast.

PR #4523, opened 2016-02-09
Linux Installers Upgrades

cra0zy:
Stuff done:
 - removed redundant code / simplified code
 - installer now installs Pipeline Tool in the xbuild folder
   - symlink is created to /opt/MonoGameSDK
 - fixed icon location
 - moved terminal commands to /usr/bin

### DIFF
--- a/Installers/default.build
+++ b/Installers/default.build
@@ -143,6 +143,7 @@ fi
     <!-- Linux Installer -->
     <if test="${(os == 'Unix' or os == 'MacOS') and directory::exists('../Tools/Pipeline/bin/Linux/AnyCPU/Release')}">
       
+      <!-- .deb installer -->
       <delete dir="Linux/tmp_deb" failonerror="false"/>
       <mkdir dir="Linux/tmp_deb"/>
       <copy todir="Linux/tmp_deb/DEBIAN/">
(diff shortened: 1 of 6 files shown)


cra0zy:
@dellis1972 This is ready, would you mind reviewing this?

Installers/default.build
```diff
@@ -143,19 +143,20 @@ fi
...
     <if test="${(os == 'Unix' or os == 'MacOS') and directory::exists('../Tools/Pipeline/bin/Linux/AnyCPU/Release')}">
       
+      <!-- .deb installer -->
       <delete dir="Linux/tmp_deb" failonerror="false"/>
       <mkdir dir="Linux/tmp_deb"/>
       <copy todir="Linux/tmp_deb/DEBIAN/">
         <fileset basedir="Linux/DEBIAN"/>
       </copy>
-      <copy todir="Linux/tmp_deb/opt/monogame-pipeline">
+      <copy todir="Linux/tmp_deb/usr/lib/mono/xbuild/MonoGame/v3.0/Tools">
```
dellis1972: not sure if we want to follow the same pattern as we do for the mac side of things here. We have a 
`/Library/Frameworks/MonoGame.framework` folder which has a versioned folder e.g v3.5.0.1234 and then a bunch of symlinks e.g.

v3.5.0.1234
v3.0 -> v3.5.0.1234
Current -> v3.5.0.1234

That way its possible to have multiple versions installed and switch between them by modifying the v3.0 symlink. 
Just a thought :)

cra0zy: I like the idea, but I can't do that because .deb installer wouldn't work (in order to install the new version, the previous version gets autouninstalled with it).

I could still do it for the .run installer, but then .run and .deb installer would be doing different things, and I wouldn't want to make a feature that is exclusive to one installer.

PR #4524, opened 2016-02-11
[AL] Fix for stopping Song while buffers are being unqueued

mrhelmut:
This PR fixes a rare threading issue in ```OggStreamer``` where ```processed``` or ```queued``` could get altered between these two lines if ```OggStream.Stop()``` was called at that very moment:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OggStream.cs#L459
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/OggStream.cs#L462

@dellis1972 @cra0zy 

(sounds more and more interesting to switch to SDL_Mixer)


cra0zy:
:+1:

PR #4525, opened 2016-02-11
[Pipeline Tool] Optimized opening of .mgcb files a bit

cra0zy:
Modifies 2 things:
 - replaces FileSystemWatcher with FileWatcher
   - I set FileWatcher to check ~10 files per second, do say if you want that increased
 - changes the directory it searches for templates from "&lt;mgcb dir&gt;/" to "&lt;mgcb dir&gt;/MGTemplates"
   - this fixes opening files located at for example "C:\" or "/home/"

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.FileWatcher.cs
+++ b/Tools/Pipeline/Common/PipelineController.FileWatcher.cs
@@ -1,0 +1,76 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using System.IO;
+using System.Threading;
+
+namespace MonoGame.Tools.Pipeline
+{
+    internal partial class PipelineController
+    {
+        private class FileWatcher : IDisposable
+        {
+            Thread _thread;
+
+            PipelineController _controller;
+            IView _view;
+
+            public FileWatcher(PipelineController controller, IView view)
+            {
+                _controller = controller;
+                _view = view;
+            }
+
+            public void Run()
+            {
+                Stop();
+
+                _thread = new Thread(new ThreadStart(ExistsThread));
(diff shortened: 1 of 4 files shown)


Tools/Pipeline/Common/PipelineController.FileWatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System.IO;
+using System.Threading;
+
+namespace MonoGame.Tools.Pipeline
+{
+    internal partial class PipelineController
+    {
+        private class FileWatcher
```
dellis1972: @cra0zy can you implement IDisposable and call Stop in the Dispose please. Just to make sure we shut this thread down

Tools/Pipeline/Common/PipelineController.FileWatcher.cs
```diff
@@ -0,0 +1,67 @@
...
+                Stop();
+
+                _thread = new Thread(new ThreadStart(ExistsThread));
+                _thread.Start();
+            }
+
+            public void Stop()
+            {
+                if (_thread != null)
+                    _thread.Abort();
```
dellis1972: Probably should null out the _thread here too. otherwise we might call _thread.Abort() twice

Tools/Pipeline/Common/PipelineController.FileWatcher.cs
cra0zy: Sure, will do it in a moment.

cra0zy:
@dellis1972 Updated what you have requested.

dellis1972:
@cra0zy cool, much better :) 

@tomspilman  @KonajuGames I'm ok with this. it will get rid of the buggy FileSystemWatcher in mono.

KonajuGames:
Thanks @cra0zy. Merging.

PR #4533, opened 2016-02-12
Misc Improvements

tomspilman:
A few simple improvements from work on consoles:

 - Made the half pixel offset a runtime setting.
 - Clear the assigned constant buffers when switching shaders.
 - Simple shader compiler compatibility improvement.


dellis1972:
nice :)

PR #4534, opened 2016-02-14
Prevent finalizer exception on Linux systems without libgdiplus

hach-que:
On my Linux system, libgdiplus doesn't load correctly.  When the type initialization exception for libgdiplus occurs inside "new Icon" it leaves an Icon object lying around in memory which is inaccessible.  When the GC comes along and attempts to finalize it, it calls Dispose which tries to access libgdiplus again, which causes the type initializer exception to occur inside a finalizer, which in turn causes the runtime to terminate because as per the .NET documentation, exceptions in finalizers have no recovery.

This fixes the issue by calling a method on the internal GDIPlus class, and jumping out of that code block if the type initializer exception is thrown before we get a chance to new up an Icon.

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
@@ -375,6 +375,26 @@ namespace Microsoft.Xna.Framework
             // try to load it using Mono, it will cause a crash because of this.
             try
             {
+                // If we're running on a system where libgdiplus is not available, we must
+                // force an attempted load here.  When "new Icon" runs and fails here because
+                // of the failure to load the library, the object is still allocated.  Later on
+                // when the GC attempts to clean up the Icon object, that will attempt to load
+                // the library in it's finalizer and throw an exception.  As per the .NET documentation
+                // when a finalizer throws an exception, the application is immediately terminated
+                // with no recovery.
+                if (t != null)
+                {
+                    var drawingTypes = typeof(Icon).Assembly.GetTypes();
+                    foreach (var type in drawingTypes)
+                    {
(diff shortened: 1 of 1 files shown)


tomspilman:
I'll let @dellis1972 and @cra0zy comment on the fix for Linux.

I assume this updated Protobuild is the latest with the platform extension features you added correct?

hach-que:
Ah whoops, I didn't meant to include Protobuild.  I think that's from a bad rebase.

hach-que:
@tomspilman Can we get this merged sometime soon?

tomspilman:
@dellis1972 @cra0zy ?

Is this safe to do?  Should I try to get it into 3.5.1?

hach-que:
It's probably not necessary for 3.5.1 (although it is a bugfix), but we should try to merge it for 3.6.

cra0zy:
@hach-que 3.6 will replace OpenTK with SDL so it's either this get's merged in 3.5.1, or not at all.

cra0zy:
@hach-que This code would have a bad result if someone is using mono + Windows, tho I doubt anyone is doing it....

tomspilman:
> The code would have a bad result if someone is using mono + Windows

Yeah...  likely on one is doing that.

>  it's either this get's merged in 3.5.1, or not at all.

True... I will add it to 3.5.1 if @cra0zy can confirm it is safe and has no unexpected side effects.

hach-que:
No-one should seriously be using Mono on Windows for production code anyway.

cra0zy:
@tomspilman Other then mono + Windows, I see no other problems, I am ok with this getting merged.

PR #4541, opened 2016-02-15
Wiki to doc edit2

ralfjansson:
Copied text about migrating OpenTK 1.0 from old wiki to new documentation, Platform specific notes/android


ralfjansson:
This text refers to Android, but also to ios. Perhaps we should add a new page for IOS under platform specific notes?

tomspilman:
This info still valid @dellis1972 ?

dellis1972:
Looks like the MonoGame Framework has been upgraded already to OpenTK-1.0 but the templates were not. We should fix that asap.

dellis1972:
Fixed as part of #4509

ralfjansson:
So were are we now. Do we merge my changes or drop them?

tomspilman:
Info on OpenTK isn't importaint anymore.  Closing.

PR #4547, opened 2016-02-18
[SpriteBatcher] use pointer to access _vertexArray

nkast:
| MG DX x64
--- | ----
*Original* Average Draw (ms): | 0.55931
*this PR* Average Draw (ms): | 0.50928
Diff (%)  | 91.0%

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatcher.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatcher.cs
@@ -177,7 +177,7 @@ namespace Microsoft.Xna.Framework.Graphics
         /// </summary>
         /// <param name="sortMode">The type of depth sorting desired for the rendering.</param>
         /// <param name="effect">The custom effect to apply to the drawn geometry</param>
-        public void DrawBatch(SpriteSortMode sortMode, Effect effect)
+        public unsafe void DrawBatch(SpriteSortMode sortMode, Effect effect)
 		{
 			// nothing to do
             if (_batchItemCount == 0)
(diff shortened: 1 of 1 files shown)


tomspilman:
So just a small 9% improvement....  I would have expected more knowing that original code must be doing a bunch of bounds checking on each array access.

nkast:
Here is an MSDN blog on the matter 
[(http://blogs.msdn.com/b/clrcodegeneration/archive/2009/08/13/array-bounds-check-elimination-in-the-clr.aspx]
`Here is bounds-check code generated by the CLR’s x86 JIT for an example array index expression a[i]:
IN0001: 000003      cmp     EDX, dword ptr [ECX+4]         // a in ECX, i in EDX
IN0002: 000006      jae     SHORT G_M60672_IG03            // unsigned comparison`

in our case my system generate:
`mov         r9d,dword ptr [rdx+8]  
cmp         ecx,r9d  
jae         00007FF89E7FAFCE` 

I guess we reach the limit here. Using the latest optimizations as baseline/reference I tried that other strategy, just copy the .Set() data in the SpriteBatchItem and calculate vertices in the loop. I didn't manage to get better results.
From now on we can only optimize for specific cases like Deffered Mode or for DX dynamic VertexBuffers.

nkast:
Well, that ~9,8% improvement is the average Draw time of https://github.com/danzel/DrawTextTest. If I were to put a Stopwatch around the for loop I could get more impressive numbers, similar to what I got for the indices array.

tomspilman:
Thanks @nkast !

PR #4549, opened 2016-02-18
[Pipeline Tool] Fix watcher class CPU usage

### LINKED ISSUE #4548: Pipeline tool has high CPU usage while idle
Yesterday, while testing the pull #4545 I found that the pipeline was using nearly a 20% of my 4 cores CPU.
The problem is coming from the changes in #4525.

cra0zy:
Fixes #4548

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.FileWatcher.cs
+++ b/Tools/Pipeline/Common/PipelineController.FileWatcher.cs
@@ -51,13 +51,13 @@ namespace MonoGame.Tools.Pipeline
 
                         foreach (var item in items)
                         {
+                            Thread.Sleep(100);
+
                             if (item.Exists == File.Exists(_controller.GetFullPath(item.OriginalPath)))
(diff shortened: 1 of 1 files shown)


Tzenchor:
Just tried with the artifacts from the server and it works
It's incredible that by waiting before the check instead of after makes CPU usage drop to 0%

cra0zy:
> Just tried with the artifacts from the server and it works
> It's incredible that by waiting before the check instead of after makes CPU usage drop to 0%

That is because I was dumb and replaced an if with continue https://github.com/cra0zy/MonoGame/blob/pipthread/Tools/Pipeline/Common/PipelineController.FileWatcher.cs#L57, but forgot to move the Thread.Sleep after.

Tzenchor:
I think I were dumber to not see that continue until now

KonajuGames:
Merging. Thanks @cra0zy

Issue #4538, opened 2016-02-14
FullScreen Viewport Broken on Native OpenTK on Mac


dellis1972:
When using native OpenTK (not SDL) on Mac. The GraphicsDevice Viewport is not updated correctly to reflect the screen resolution when in full screen mode. As a result it draw the game in the bottom right corner of the screen. This was working "something" changed to break this. 

@cra0zy @mrhelmut any ideas 

<img width="1440" alt="screen shot 2016-02-14 at 22 39 14" src="https://cloud.githubusercontent.com/assets/810617/13036944/6e95e5f8-d36c-11e5-893a-1975d1343307.png">

cra0zy:
What happens when you use SDL backend?

dellis1972:
@cra0zy it kinda works but the window state is weird.
When toggling between full and windowed it changes the viewport to windowed but the window ends up in the bottom left and you get a full black screen. Using the menu "Enter Full screen" works, but toggling in code has some weird behaviour.

Oddly this was working perfectly a while back (28th Jan)

<img width="1440" alt="screen shot 2016-02-14 at 23 07 55" src="https://cloud.githubusercontent.com/assets/810617/13037070/faf553a0-d36f-11e5-94bb-59ce6e8dfaf8.png">

dellis1972:
Just can't nail down the commit which caused it :/

dellis1972:
Hmm ChangeClientBounds is not being called when switching to FullScreen anymore

cra0zy:
It's probably something with OpenTKWindow, try going threw https://github.com/mono/MonoGame/commits/develop/MonoGame.Framework/Desktop/OpenTKGameWindow.cs

dellis1972:
Yup that's what I've been doing.

mrhelmut:
Are you on a Retina display? (or changed the DPI settings of your Mac?) This looks quite similar to what's happening on Retinas.

dellis1972:
@mrhelmut I am.. what confused me is it worked a few weeks ago.. and I still have the binaries of the working version.

dellis1972:
I might do a monodis on the two versions and see what changed

mrhelmut:
If it worked before, that sounds interesting because I've been trying hard to figure out this without much luck. I've only been using SDL on OS X, though.
Looks like I've started working with Retinas after that change (about a month ago).

dellis1972:
[...]

PR #4551, opened 2016-02-19
[DesktopGL] Rewrite OpenTK Game Window and Platform

cra0zy:
This should fix the following things:
 - no more hacking around borders
 - no more hacking the exit loop
 - #4538
 - makes opentk window act more stable
 - simplifies code by a lot

~~Do note that it's missing a resolution change, this is because I am looking into how to make that work correctly with multiple displays, I am just putting this out early so it can be tested and so that people can tell me what they think while I figure out this last problem.~~ Linux resolution bug is present only under X11 backend, this is something that's gonna have to be fixed within OpenTK.

@dellis1972 Can you test this on Mac?

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -1,107 +1,34 @@
-#region License
-/*
-Microsoft Public License (Ms-PL)
-MonoGame - Copyright © 2009-2011 The MonoGame Team
-
-All rights reserved.
-
-This license governs use of the accompanying software. If you use the software,
-you accept this license. If you do not accept the license, do not use the
-software.
-
-1. Definitions
-
-The terms "reproduce," "reproduction," "derivative works," and "distribution"
-have the same meaning here as under U.S. copyright law.
-
-A "contribution" is the original software, or any additions or changes to the
-software.
-
-A "contributor" is any person that distributes its contribution under this
-license.
-
-"Licensed patents" are a contributor's patent claims that read directly on its
-contribution.
-
-2. Grant of Rights
-
-(A) Copyright Grant- Subject to the terms of this license, including the
-license conditions and limitations in section 3, each contributor grants you a
-non-exclusive, worldwide, royalty-free copyright license to reproduce its
-contribution, prepare derivative works of its contribution, and distribute its
-contribution or any derivative works that you create.
-
-(B) Patent Grant- Subject to the terms of this license, including the license
-conditions and limitations in section 3, each contributor grants you a
-non-exclusive, worldwide, royalty-free license under its licensed patents to
-make, have made, use, sell, offer for sale, import, and/or otherwise dispose of
-its contribution in the software or derivative works of the contribution in the
-software.
-
-3. Conditions and Limitations
-
-(A) No Trademark License- This license does not grant you rights to use any
-contributors' name, logo, or trademarks.
-
-(B) If you bring a patent claim against any contributor over patents that you
-claim are infringed by the software, your patent license from such contributor
-to the software ends automatically.
-
-(C) If you distribute any portion of the software, you must retain all
-copyright, patent, trademark, and attribution notices that are present in the
-software.
-
-(D) If you distribute any portion of the software in source code form, you may
-do so only under this license by including a complete copy of this license with
-your distribution. If you distribute any portion of the software in compiled or
-object code form, you may only do so under a license that complies with this
-license.
-
-(E) The software is licensed "as-is." You bear the risk of using it. The
-contributors give no express warranties, guarantees or conditions. You may have
-additional consumer rights under your local laws which this license cannot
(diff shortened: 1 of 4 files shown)


dellis1972:
will try to take a look over the weekend.

tomspilman:
Is it really worth making these changes with the SDL2 move right around the corner?

We really need to get 3.5 out the door in the coming weeks.

cra0zy:
> Is it really worth making these changes with the SDL2 move right around the corner?
> 
> We really need to get 3.5 out the door in the coming weeks.

That is the reason why I decided to make this change, so 3.5 DesktopGL wouldn't have a buggy window.

This should be the last change I do to the OpenTK backend.

tomspilman:
If @mrhelmut and @dellis1972 agree this is stable and better for 3.5 then i'm good with it then.

cra0zy:
@tomspilman @mrhelmut @dellis1972 OpenTK has 2 backends, native and SDL, and with SDL backend the window resizable parameter is not changeable. Do you want it to be always on, always off, or should we do something else about it (I can't do the same thing I did for SDL)?

mrhelmut:
Sounds interesting, I'll test that by Monday.

mrhelmut:
Back at the office, I tested this on a Mac Retina and Windows 4K, and they both have problems with cropped viewport (SDL backend).

cra0zy:
@mrhelmut That bug is not related to Window/Platform code, it' related to GraphicsDeviceManager/OpenTK, and therefore I cannot fix it in here.

mrhelmut:
I know, it's quite entangled together. The current implementation somehow works with Retina/4K, your changes make the situation worst unfortunately. :(

cra0zy:
How do they make it worse?

mrhelmut:
The window size end up being twice as big as it should be on OS X with DPI scaling enabled (only 1/4 of the game is displayed when fullscreen).

dellis1972:
Is this just for the SDL2 backend or native as well?
I've not had a chance to test it yet.

cra0zy:
> The window size end up being twice as big as it should

Let me get this straight, in windowed mode, without this PR window size is correct, but with this PR window size is doubled?

If I got that right, I think I know how to fix it.

mrhelmut:
@dellis1972 I've only tested the SDL2 backend.

@cra0zy It works as intended in windowed mode (the size _is_ doubled, but that's what OS X expect it to be). Once in fullscreen, the viewport is twice bigger and goes behind the screen space.

cra0zy:
> @cra0zy It works as intended in windowed mode (the size is doubled, but that's what OS X expect it to be). Once in fullscreen, the viewport is twice bigger and goes behind the screen space.

Oh, so what you are saying is that the reverse of #4431 is happening. Well that isn't too bad.

mrhelmut:
Not quite, because the viewport was still usable with resolutions below the virtual OS X desktop size (e.g. a 2880x1200 display has a reported resolution of 1440x600, anything equal or lower than this works just fine). These changes make all resolutions to have this issue.

cra0zy:
@dellis1972 Did you get the chance to try this PR out?

dellis1972:
@cra0zy I haven't yet. The build archives are missing so we'll need to get a new build somehow (do a rebase and push?), I don't really have time to compile this myself atm.

cra0zy:
@dellis1972 Done :)

cra0zy:
I just saw what is used for retina scaling code, `float scale = window.ClientSize.Width / (float)window.Size.Width;` we should be able to use it to get the correct fullscreen rendering on Retina, tho I can't do anything about it, I have no Mac or Retina display.

dellis1972:
@cra0zy this works for me on my test game, I can flick between full/windowed and it renders correctly on Mac and windows. But I'm not using the SDL2 backend like @mrhelmut is.

mrhelmut:
The easiest way is indeed to use the returned scale when using SDL2. I think this would make this PR on par with the current implementation.

(The other Retina issues can wait for OpenTK replacement IMO.)

cra0zy:
@mrhelmut Since I don't have a Mac, can you add the missing piece of code to this PR? (I will give you push privileges to my repo)

mrhelmut:
I'll take a look. :) (I won't have access to a retina until Monday)

cra0zy:
> I'll take a look. :) (I won't have access to a retina until Monday)

Ok, I gave you push privileges to my repo.

dellis1972:
I have a Retina... can you post a sample which shows the problem? then the
code and I can fix it :) (if I get time ;) )

cra0zy:
@dellis1972 Just FYI, you still have push privileges to my repo.

dellis1972:
cool ta :)

cra0zy:
I don't know what to do with this PR...

tomspilman:
@cra0zy - Likely we should just focus on the future...  @dellis1972 is working this weekend on the GL/AL bindings.  As soon as that is ready we will merge your branch and we're free of OpenTK for good.

PR #4552, opened 2016-02-19
Fixed typo in LICENSE.txt

PAVBAN95:
Some commas were misplaced.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -9,7 +9,7 @@ software.
 
 1. Definitions
 
-The terms "reproduce," "reproduction," "derivative works," and "distribution"
+The terms "reproduce", "reproduction", "derivative works" and "distribution"
 have the same meaning here as under U.S. copyright law.
 
 A "contribution" is the original software, or any additions or changes to the


LICENSE.txt
```diff
@@ -9,7 +9,7 @@ software.
 
 1. Definitions
 
-The terms "reproduce," "reproduction," "derivative works," and "distribution"
+The terms "reproduce", "reproduction", "derivative works" and "distribution"
```
tomspilman: Interesting...  I wonder if that is a legalese thing?  Because the same typo exists in the original MsPL license:

https://opensource.org/licenses/MS-PL

KonajuGames: Are we permitted to change that, even such a small change, and still claim
it is Ms-PL?

tomspilman: No clue...  my first thought is that the risk of the license being wrong isn't worth it.

tgjones: The original text, with the commas inside the quotation marks, is technically correct:
http://www.grammarbook.com/punctuation/quotes.asp (see rule 3a)
 
As a programmer, this part of English, with its unbalanced delimiters, makes me highly uncomfortable. But that's the language we're stuck with.

tomspilman: Yeah...  thanks for looking out for this @PAVBAN95, but this is correct both grammatically and to the original license.

Issue #2659, opened 2014-06-02
Able to add a custom cursor.


kourbou:
Hello,
I tried to use a spriteBatch as a cursor but it has a short delay. In OpenTK there is an option to set a custom cursor. Maybe it could be possible to implement this directly in MonoGame for both WindowsGL and WindowsDX.

cra0zy:
Base implementation for this has been added with #4755, currently implemented for DesktopGL platform.

cra0zy:
@tomspilman This got implemented in the SDL branch, this should be closed.

PR #4555, opened 2016-02-20
WIP SDL2 Branch

cra0zy:
This is here so the SDL2 branch can be compiled with MG buildbot.


cra0zy:
> All that is needed for the new build process to work with a branch is merging of this one commit: 2482fd6

It's merged, the branch is up to date with develop https://github.com/cra0zy/MonoGame/tree/SDL2

tomspilman:
@cra0zy - This is going to fail for a bit... working on getting the new build server setup.

tomspilman:
@cra0zy - Note that NuGet packaging failed for Linux...

http://teamcity.monogame.net/viewLog.html?buildId=2427&buildTypeId=MonoGame_PackageNuGet&tab=buildLog#_state=96&focus=105&guest=1

cra0zy:
> @cra0zy - Note that NuGet packaging failed for Linux...
> 
> http://teamcity.monogame.net/viewLog.html?buildId=2427&buildTypeId=MonoGame_PackageNuGet&tab=buildLog#_state=96&focus=105&guest=1

Totally normal, templates/nuget will get updated at a later point, once the framework is fully working without OpenTK.

cra0zy:
@tomspilman I want to know your opinion on the last commit I did. In it I added a GlobalX and GlobalY variables to MouseState so users can track the global mouse state. 

The reason why I added it that was is because Sdl.Mouse.GetState cannot track the mouse if it leaves the window bounds, it will just return the last known position.

tomspilman:
> I added a GlobalX and GlobalY variables to MouseState so users can track the global mouse state. 

I really don't want to add that.

If SDL2 should be able to handle this...  we're not asking for anything exotic or anything.  It is pretty common game stuff to be able to track the mouse position outside the window too.

Have we reported this issue to the SDL2 team and requested an improvement to the API?

cra0zy:
> If SDL2 should be able to handle this... we're not asking for anything exotic or anything. It is pretty common game stuff to be able to track the mouse position outside the window too.
> 
> Have we reported this issue to the SDL2 team and requested an improvement to the API?

It's not a bug.

Sdl.Mouse.GetState tracks mouse position while the mouse is in the window, and returns x, y relative to window,

Sdl.Mouse.GetGlobalState tracks mouse globally.

We can't use the second to try and do PointToClient because there is no way to get window borders;

Any ideas?

tomspilman:
> It's not a bug.

I know it isn't a bug...  it is a missing feature.

We want global mouse state relative to the client area of the window.  SDL2 as it is today can't provide that.  We are requesting a feature.

KonajuGames:
Can MouseState X and Y be updated from SDL's global if the mouse is outside
the window? This may involve keeping track of both window and global and
the offset between them. If the window X,Y stop updating (the difference
between global and window != offset we had before) then update mouse state
from global + offset.

Just an idea, without knowing the details of how SDL works internally.

cra0zy:
> Can MouseState X and Y be updated from SDL's global if the mouse is outside
> the window? This may involve keeping track of both window and global and
> the offset between them. If the window X,Y stop updating (the difference
> between global and window != offset we had before) then update mouse state
> from global + offset.

We can't do that, it would require for the mouse to start inside the window, which means the code would need to move the mouse to the window, which is very very annoying.

tomspilman:
Anyway...  for now lets not add any extra global position to MouseState.  It will remain crippled like this until either figure out how to do this within SDL2's existing API or SDL2 adds an API to allow us to get what we need.

cra0zy:
> Anyway... for now lets not add any extra global position to MouseState. It will remain crippled like this until either figure out how to do this within SDL2's existing API or SDL2 adds an API to allow us to get what we need.

Want me to add SDL preprocessor param?

cra0zy:
I implemented #2659 in here O o...

tomspilman:
> Want me to add SDL preprocessor param?

I don't see that we need any...   SDL2 is only used under `DESKTOPGL`.

> I implemented #2659 in here O o...

Let's not do any more of this.

The goal of this branch is to get SDL2 implemented... not to implement other feature requests.  We can address those after we get SDL2 done and merged back into develop.

cra0zy:
> I don't see that we need any... SDL2 is only used under DESKTOPGL.

Sorry, didn't phrase it right, I meant want me to put GlobalX and GlobalY under the DESKTOPGL preprocessor param.

> Let's not do any more of this.
> 
> The goal of this branch is to get SDL2 implemented... not to implement other feature requests. We can address those after we get SDL2 done and merged back into develop.

Yea, sorry, I can remove it from here if you want.

tomspilman:
> I meant want me to put GlobalX and GlobalY under the DESKTOPGL preprocessor param.

No...  I just don't want this API at all.  I would rather have the clamp on the mouse cords occur until we either figure out how to fix it with the existing SDL API or we get the SDL API fixed to support what we need.

Have we bothered to ask on the SDL mailing list or forms about what we're looking for?  I shouldn't be impossible to get global mouse coordinates relative to the window client area.

> Yea, sorry, I can remove it from here if you want.

We can leave it in for now...  lets just not add any more.  This will end up being a huge PR as it is.

cra0zy:
> No... I just don't want this API at all. I would rather have the clamp on the mouse cords occur until we either figure out how to fix it with the existing SDL API or we get the SDL API fixed to support what we need.

Time to try and delete a commit :D

> Have we bothered to ask on the SDL mailing list or forms about what we're looking for? I shouldn't be impossible to get global mouse coordinates relative to the window client area.

I really don't know where to ask.

Tzenchor:
@cra0zy , I downloaded your branch yesterday night and tried to tweak the mouse state, but always, when trying to run it complains that it can't find `SDL2.dll`

cra0zy:
> @cra0zy , I downloaded your branch yesterday night and tried to tweak the mouse state, but always, when trying to run it complains that it can't find SDL2.dll

Yes, you need to copy SDL2 to the output directory of your game, it will be later included in the template stage.

You can get it from: https://www.libsdl.org/download-2.0.php

Tzenchor:
> You can get it from: https://www.libsdl.org/download-2.0.php

I tried to add a reference to it instead of copying it

cra0zy:
> I tried to add a reference to it instead of copying it

Since you seem to don't know, C# .dll files are not same as system .dll files. They do have the same extension, but they are very different. Same goes for C# .exe files.

Tzenchor:
Yes, I suspected that when Visual Studio complained. Simply I didn't thought of copying the dll directly.
And now that I think of it, you'll need the dll also to run the tests.

Tzenchor:
@cra0zy, after a pair of days fighting with a BadImageFormatException (had to change my game compilation platform from x86 to x64) I have found how to get the coordinates of the mouse relative to the window always. Simply replace the lines 95-97 with this:
```
int x, y;
int wx, wy;
var state = Sdl.Mouse.GetGlobalState(out x, out y);
Sdl.Window.GetPosition(window.Handle, out wx, out wy);
window.MouseState.X = x - wx;
window.MouseState.Y = y - wy;
```
Tested on Windows 10 and gives the correct coordinates, even when the window is below the taskbar.

Also I have to say that this branch solves the problem I was having in #4591

tomspilman:
@Tzenchor - Does that account for the window borders?  Is the 0,0 point within the window client area?

cra0zy:
> after a pair of days fighting with a BadImageFormatException (had to change my game compilation platform from x86 to x64)

Hahaha, sorry for laughing but I was having the same problem with Gtk earlier today...

> @Tzenchor - Does that account for the window borders? Is the 0,0 point within the window client area?

I didn't try, and I can't test it for the next 30 mins, but now he is getting my hopes up... tho I doubt 0,0 is in client area.

Tzenchor:
> Does that account for the window borders? Is the 0,0 point within the window client area?

Yes it does, on the left border is hard to tell if it is inside or not as it is only 1 pixel wide, but the title bar is 16-20 pixels high, so there is pretty evident

cra0zy:
> Yes it does, on the left border is hard to tell if it is inside or not as it is only 1 pixel wide, but the title bar is 16-20 pixels high, so there is pretty evident

Doesn't work on Linux :( it gives the correct window position. That does seem like a bug with SDL on Windows,

Tzenchor:
> Doesn't work on Linux :( it gives the correct window position. That does seem like a bug with SDL on Windows,

Another option:
```
int x, y;
var state = Sdl.Mouse.GetGlobalState(out x, out y);
window.MouseState.X = x - window.ClientBounds.X;
window.MouseState.Y = y - window.ClientBounds.Y;
```
Let's hope the `window.ClientBounds` is consistent in Windows and Linux

cra0zy:
> Let's hope the window.ClientBounds is consistent in Windows and Linux

Nope, that was my old code, which didn't work.

tomspilman:
> That does seem like a bug with SDL on Windows,

Gosh...  SDL seems like just as much as mess as OpenTK.

I'll be posting this question on the SDL forums and i'll report back with what I find out.

cra0zy:
> Gosh... SDL seems like just as much as mess as OpenTK.

Well it has less bugs... just saying...

> I'll be posting this question on the SDL forums and i'll report back with what I find out.

Better let me open up an issue on bugzilla about this.

cra0zy:
Someone beet us to it: https://bugzilla.libsdl.org/show_bug.cgi?id=3085

cra0zy:
Good news, the mouse bug has been fixed 2 days ago :D

tomspilman:
> Good news, the mouse bug has been fixed 2 days ago :D

Great... so that means eventually here @Tzenchor suggested method should work and solve this issue once and for all.

cra0zy:
> Great... so that means eventually here @Tzenchor suggested method should work and solve this issue once and for all.

Yes, *caugh* https://github.com/cra0zy/MonoGame/blob/cbc70058ad3119c84376d73dbf46a3e1cfe4ed34/MonoGame.Framework/Input/Mouse.cs#L97 *caugh*, turns out our methods were the same.  I'll update this PR to require the current dev version of SDL a bit later today, since the current dev version will become stable before this PR is done.

tomspilman:
> turns out our methods were the same

Sucks that this bug on Linux caused this whole mess.

mrhelmut:
```SDL_GL_GetProcAddress``` doesn't seem to be working quite well on Windows. Lots of function entry point can't be returned (e.g. glEnableVertexAttribArray, glDisableVertexAttribArray, glBindBuffer, glDrawBuffers, glUniform1i, glUniform4fv, etc.).
According to the ```SDL_GL_GetProcAddress``` [documentation](https://wiki.libsdl.org/SDL_GL_LoadLibrary), it is not reliable alone, but these are standard OpenGL 2.0 functions, so I don't really know what's happening (probably drivers not compatible with GetProcAddress, which seem to be common on windows).

dellis1972:
woot it builds!

cra0zy:
> woot it builds!

https://youtu.be/xos2MnVxe-c

cra0zy:
@dellis1972 Are we ready for merging? On a side note you merged branch into itself so we have double commits. We might want to try and rebase it.

mrhelmut:
I'll give this more test soon, I still have issues with ```OpenGL.GL.LoadEntryPoints();``` not being called after the GL context creation, and some threading problem (thread-safe content loading fails).
I'll push a PR to this branch if need be.

cra0zy:
Feel free to commit directly onto my branch, I gave you push access a while back because of a PR.

dellis1972:
I don't get why it works for me on mac / linux but u have to call LoadEntryPoints later

dellis1972:
We should probably squash this beach into one commit

dellis1972:
Branch even lol

cra0zy:
> We should probably squash this beach into one commit

Agreed, wanna do the honors?

mrhelmut:
Loading ```Texture2D``` (and possibly other types of content requiring ```BlockOnUIThread```) fails if loaded from a thread.
```BlockOnUIThread``` correctly detects the thread, but after ```Add(Action action)``` call, nothing happens and ```resetEvent.Set()``` is not reached. The action doesn't seem be to started.

Any hints @dellis1972 ?

cra0zy:
Threading works fine for me from Linux.

dellis1972:
@mrhelmut is Threading.Run being called ?

mrhelmut:
Yes, and the action is correctly executed. After debugging some more, ```resetEvent.Set()``` is correctly called, but ```resetEvent.Wait()``` doesn't catch it...

mrhelmut:
Well, let's assume this is me messing up with threads. I'm using a thread-safe content loader because my assets are streamed and I was blocking the UI thread in Draw because OpenGL doesn't like textures to be loaded while draw calls occur, but with the texture loadings being dispatched on the UI thread, I believe this not a problem anymore.

Anyway, other issues: 

- disposing render targets or textures in ```Game.UnloadContent()``` ends up with a GL error 1282 in ```FrameBufferHelper.DeleteRenderbuffer(int)``` or ```Texture.OpenGL.DeleteGLTexture()```. Disposing them before works.
- some mouse button events are not working, I believe this has something to do with the out of window check being wrong in some situation (windowed, possibly related to Hi-DPI)

cra0zy:
> some mouse button events are not working, I believe this has something to do with the out of window check being wrong in some situation (windowed, possibly related to Hi-DPI)

If you are using SDL 2.0.4 or lower mouse state got fixed in #4869

mrhelmut:
Okay, I'll check this, I need to build a 64bit version of SDL.

PR #4545, opened 2016-02-17
Upgrade Font Importer

Tzenchor:
This is a new pull request based on #4475. It does in one step what it did before in #4475.
Changes:
Fonts are now completely read to get its real sizes, providing a better
fit vertically
Corrected some typos in old comments and deleting commented calls to non
existent code
Some lines have small changes in the indentation because it was wrong or
were using tabs
Added comments for future souls trying to understand how this works

PR #4537, opened 2016-02-14
[Ready to Test] Spritefont styles

Tzenchor:
**As of 20/02/2016 this is ready to test**
I'm working on enabling the style tag of the spritefonts as talked in #4149. At this time i can find the font based on its filename or its family name if it is an installed font in Windows.
The changes at this time include:
- Add 3 private fields to `FontDescriptionProcessor` that contains the supported font extensions, possible terminations of bold font files, and the possible terminations of italic font files.
- Add method `FontDescriptionProcessor.FindFontByFileName`: Considers the fontname as the filename of the font, and tries to find the font file that better matches the name and the style (for example if the users wants "arial" bold, this finds "arialb.ttf"), the method relays heavily on LINQ and I have added comments in each step at least for now, but I don't mind removing them in future.
- Add private class `FontDescriptionProcessor.FileInfoComparer`: Required in method `FontDescriptionProcessor.FindFontByFileName`
- Update method `FontDescriptionProcessor.FindFontFileFromFontName` for Windows: Added style information to the method call and updated its logic.
- Simplified method `FontDescriptionProcessor.Process`: Reworked the method to remove an `if` embedded inside an `#if` and make the procedure simpler.
- Reorder the elements of `FontDescriptionStyle`, now Regular is 0, bold is 1, italic is 2, and by extension bold and italic is 3. Before Bold was 0, italic was 1 and regular was 2.

Both methods look for the best match for the users intentions, giving preference to the italic fonts (if requested).

The code  in `FontDescriptionProcessor.FindFontByFileName` is tested, and linked inside `FontDescriptionProcessor.Process`, but it's result is not used at this time as I want to make the other procedures needed.

I'll need guidance with the following things:
- To search in every folder using family names and to see what styles are available inside each font file i'll need to use SharpFont to read that data, but it feels wrong using it outside `SharpFontImporter`. The problem is that if I do this inside `SharpFontImporter` i'll have to move code from `FontDescriptionProcessor` to `SharpFontImporter`, i'm not sure what option is better.
- Related to the previous point, for rendering the font styles manually I'm gonna need to have information about the found font to know what treatment it needs to get the expected style, and I can do it when retrieving each glyph from the font inside `SharpFontImporter` or after that when back in `FontDescriptionProcessor`
[...]

PR #4561, opened 2016-02-23
Font Style Renderer

Tzenchor:
I'm working on methods to create the font styles manually for fonts where we can't find a font with the style included. This is complimentary to my pulls #4537 and #4545 (may have conflicts with this one).
This is the result of my first version
embolden and italic first pass
Both texts are taken from the regular arial font, and the big one is emboldened and italiced, while the second one is only emboldened. In theory the first one should have been only italiced, but there's a problem with `FontDescriptionStyle` and the `Bold` flag is always set.

For the italiced effect i'm skewing the characters about 20 degrees.
For the emboldened effect i'm drawing the bitmap 3 times with 1 pixel offset each one.

I still have to review the spacing of the characters as the italics are drawn outside it's reported size horizontally (vertically the problem is resolved in #4545 as that comes from other place)


tomspilman:
This looks like a fantastic start...  thanks for taking the initiative on this @Tzenchor !

You should probably examine the results from XNA and try to mimic as much as possible.  Of course I doubt it will ever be the same...  but getting a similar slant for italic and thickness for bold is helpful.

> For the emboldened effect i'm drawing the bitmap 3 times with 1 pixel offset each one.

 It looks a little jagged on the edges to me... is that what you are seeing too?  Is that visible in the non-bold non-italicized font?

Tzenchor:
> It looks a little jagged on the edges to me... is that what you are seeing too? Is that visible in the non-bold non-italicized font?

It is normal, for small sizes some fonts contain bitmaps instead of using outlines, and SharpFont does not do any kind of antialising on them.
I'm gonna fix the widths first and then i'll have a look at softening the jagging.

tomspilman:
> It is normal, for small sizes some fonts contain bitmaps instead of using outlines, 

Well I suggest to do your initial testing with a font with vector data.  Don't want to confuse jaggies in the font with jaggies in your italic/bold technique.

> then i'll have a look at softening the jagging.

I would say leave that for a future PR.  More important to get italic and bold support done then supporting low quality fonts.

Tzenchor:
> Well I suggest to do your initial testing with a font with vector data. Don't want to confuse jaggies in the font with jaggies in your italic/bold technique.

Have a look at this, it's also arial, but big enough to use the outlines as base instead of prerendered bitmaps.
This first image is the font without modifying it
regular
And this one it is only emboldened
bold
I think that more or less the outline is the same

tomspilman:
Yeah... that looks pretty good.  Still would want to compare what XNA does in the same case.

Tzenchor:
This is what I got from XNA, first I have tried with Arial, 
xna comparison
Comparing the the italic test I think the angle is the same, for bold the increase in thickness is also the same, 2 pixels.
But, because i can't guarantee that it is not taking the corresponding font file, I have also tried using Segoe UI Mono
xna style segoe
As can be seen, the size variations are the same.

Also, when XNA makes bold effects it also increases the thickness of horizontal lines, while my method does not

tomspilman:
@Tzenchor - Looks like maybe you added the same screenshot twice in that last post?

Also can you post a screenshot of the same test using your italic and bold techniques?  Just want to see a clear comparision of the differences.

Tzenchor:
I don't know what I did with the screenshots in the previous post.
This screenshot contains all four combinations, the left ones are XNA spritefonts and right ones are from my code with a little tweak to make it generate all styles
captura.
Comparing them XNA fonts looks less jagged because if you zoom on them you will see an aura around each glyph of half opaque pixels

tomspilman:
> right ones are from my code with a little tweak to make it generate all styles

 Aside from the jaggies mostly on the italic...  it seems you have flat spots on the tops and bottoms of lowercase characters like 'a' and 'e'.  This seems like it could be an unintended side effect of not making horizontal lines thicker.

> Comparing them XNA fonts looks less jagged because if you zoom on them you will see an aura around each glyph of half opaque pixels

Well something to note that could be affecting your comparison (making XNA better or us look worse).  

XNA always compresses their fonts with a specialized DXT3 compressor...  this could be where some of that aura is coming from.

We currently don't use a specialized compressor and if you let our SpriteFontProcessor compress the font you get a really poor result.  So make sure you disable compression when using our processor for now. 

See https://github.com/mono/MonoGame/issues/2853#issuecomment-178852801 for details.

KonajuGames:
XNA also uses the GDI to rasterize glyphs, which can provide anti-aliasing for generated styles. If we are just doing a skew for italic, then we lose the anti-aliasing that FreeType might be giving us.

tomspilman:
>  If we are just doing a skew for italic, then we lose the anti-aliasing that FreeType might be giving us.

We would either need to find a way to restore it or see if we can get FreeType to skew for us.

I seem to remember FreeType having some sort of transform you can pass to it which is applies during rasterization.  Did you see that @Tzenchor ?

KonajuGames:
A 2x2 skew matrix should work there.

Tzenchor:
> I seem to remember FreeType having some sort of transform you can pass to it which is applies during rasterization.

I have just made it, and the results are a lot better. I have rearranged the fonts for this image, top left group is my manual method, right top is using the transform method, and lower left are the XNA ones.
italic transformation
Just for the record the matrix i'm using is: `FTMatrix matrix = new FTMatrix(0x10000, (int) (0.36 * 0x10000), 0, 0x10000);`, it's a modification from the one in [freetype tutorial](http://www.freetype.org/freetype2/docs/tutorial/step2.html).
Now to try to get a better bold method

tomspilman:
> I have just made it, and the results are a lot better.

Yeah...  that is a lot better.  Looks like we're getting close here.


> I know the text width is wrong at this time will try to fix it later

Sure... one thing at a time.

Tzenchor:
A little update, the italic transformation has a little problem, for bitmap glyphs it does not work. As proof, in the next capture look at the small text at the right part, it should have been skewed, but it is clearly not.
italic problem
That can be solved with my manual skew algorithm, but that is for latter.

About the bold effect, in theory the bitmap can be emboldened through `FTBitmap.Embolden` but I can't make it work, all the time i'm getting a `FreeTypeException`.
Also I have tried to reduce the flattening at the top of the glyphs in my method by adding the glyph 1 pixel higher, but the result is even worse (can be seen in the upper screenshot)

tomspilman:
> for bitmap glyphs it does not work.

When bitmap glyphs occur?  Is it something embedded in the font?  Or just something FreeType does?

KonajuGames:
Fonts, at least TrueType, can contain bitmap glyphs for small point sizes to rasterize better.  The documentation for FreeType states that the transform works for vector glyphs only.

tomspilman:
> can contain bitmap glyphs for small point sizes to rasterize better. 

So likely the Windows font system works around this as well.  Likely the also start from the bitmap font in that scenario and do pixel transforms.  We just need a much better skew that doesn't result in jaggies.

Tzenchor:
I have upgraded my skew algorith, this time instead of moving the pixels a fixed distance i move them as the angle dictates. In the next image you will see its results. The left groups are: my old method, the XNA font and the SharpFont transformation (the bold effect is from my code). The right groups are the results of my new method applyed to vector fonts and bitmap font (Calibri in this case).
improveditalic
I have not committed this new code as I still have test code and commented fragments

About the bold effect, I can't do anything about the flattening, because that is a side effect of drawing the glyph 3 times, and if drawn 2 times the text does not feel wide enough compared to the regular text.
For small text i'll have to draw the glyph only 2 times as 3 times is too wide

KonajuGames:
The horizontal spacing is being expanded when it goes through the italic transformation.  The difference in text length between the regular and italic version is quite large compared to the XNA (GDI) version.  Even the difference in bold length is quite large.

Tzenchor:
> The horizontal spacing is being expanded when it goes through the italic transformation. The difference in text length between the regular and italic version is quite large compared to the XNA (GDI) version. Even the difference in bold length is quite large.

I know, but first thing is getting the glyph transformation right, then I can get to correct the spacing.
Keep in mind that for any transformation i have to widen the bitmap, and i have not reviewed the character sizes to correct them, because it depends on the method applyed. For example, the lower left group, that comes from the sharpfont transformation, is drawn outside the area reported by MeasureString because the transformation is not applyed to the glyph metrics. In my code, the bold effect does a little correction that halves the difference in size, but the new italic method does not do any correction at this time

Tzenchor:
Today I have been trying to improve the bold render by pondering the pixel values of each copy of the glyph, and this is the best I could get by giving the central copy double the weight than the offset copies, and ignoring the value for pixel if it was white (to avoid the diluted effect on the edges). Ignore the results for the small font.
bold ponderedx2
The outline in general is better as this preserves more detail, specially when only the bold effect is applyed, but on diagonal lines or with half painted pixels.
Unless someone has any idea to improve the bold effect, i'll return to the old version and begin fixing the characters width tomorrow

Tzenchor:
I have updated the procedure, and this is the current result with the code commited.
latest
The groups are the same than the previous post, but I have changed the text to have some reference for the limits of the text.
About what has changed apart from the character placement, the `Italice` method is the same from the previous post, but with a little fix, the bottom line was not being drawn, also with this version it is possible to modify the angle at will.
The `Embolden` is the same as original one apart from drawing the glyph 2 times if the size is 15 or less. I have tried other methods but none provided good results. If someone has any idea I don't mind trying. Also, at this time it is receiving a parameter called `finalHeight` that I forgot to remove.
I have included the modified `FontDescriptionStyle.cs` that matches the XNA values but did not include the test as it is part of #4537 (I need the file or the font will always be emboldened).

Tzenchor:
I have been looking if updating Sharpfont can help with this, and I think it is. In the following image you can see the results., ignore the black areas as this was made in another copy of the project and I have only copied what was required to combine the Sharpfont embolden with my italice method
composite
The groups are from top to bottom and left to right

- XNA, Segoe UI Mono, size 20
- XNA, Calibri, size 10
- My initial bold and italics method, Segoe UI Mono or Arial, size 20
- My last italice method combined with Sharpfont Embolden method, Segoe UI Mono, size 20
- My last italice method combined with Sharpfont Embolden method, Calibri, size 10
- Sharpfont Transform for italics with my embolden method, Segoe UI Mono, size 20. This method can't be used with bitmap fonts.

The Sharpfont embolden is making the glyphs 2 pixels wider and 1 higher for size 15 or bigger and only 1 pixel wider for smaller sizes.

If this path of work is accepted, it will require to combine a pull to `Dependencies` to update SharpFont libraries, and combine it with one here to update `SharpFontImporter.cs` because since version 3.0, Sharpfont handles the conversion between pixels and points internally.

tomspilman:
Sorry for the slow reply to this @Tzenchor ... GDC16 had me crazy busy.

I think this all looks good for the first pass at this...  really this is all a 100% better than italic and bold not working at all.  Still want to be sure the processor generates a warning message if this "simulated" bold or italic is used.

We can then look to make improvements in future PRs.

Tzenchor:
> Sorry for the slow reply to this @Tzenchor ... GDC16 had me crazy busy.

I supposed something like that, no worries.

I have implemented the log messages, because `SharpFontImporter` has no access to the logger I made the booleans visible to `FontDescriptionProcessor` and made the messages there. They look like this, and do not appear if the output filter is on.
warnings
And this is the result of the procedure, concretely the 2 top groups on the right side:
warnings result
Bold effect is 2 pixels wider and 1 higher, that means the glyph is added a total of 6 times, if it is preferred the old result of only widening the glyph I don't mind changing it,

tomspilman:
Not this should fix https://github.com/mono/MonoGame/issues/4641.

tomspilman:
If the conflicts in this issue are fixed we can consider merging it.

@harry-cpp @mrhelmut ?

mrhelmut:
Merge conflits aside, it looks like an interesting addition to me.

Tzenchor:
Just solved the conflicts, but i have not tested it.

SimonDarksideJ:
Any objections to merging this @mrhelmut ?

mrhelmut:
> Any objections to merging this @mrhelmut ?

I'm not very sure with the changes to ```SharpFontImporter.ImportGlyph()``` and what it does to the glyphs data (ABC and XAdvance) when no effect is applied. I need to check the maths, even though it looks okay with the example, some fonts might break. Need to double check.

SimonDarksideJ:
Looking to include for the 3.8.3 release @Tzenchor , can you rebase and assuming the tests pass, this looks good to go.

ThomasFOG:
I don't think this is merge-ready. There are breaking changes and it is untested with more diverse fonts.

Tzenchor:
Hi, sorry for the delay, i'm unable to upgrade this. Between i'm too rusted on working with forks i messed up the branch, also because the library to read the fonts has changed needs more modifications.
What i can i do is explain the changes and someone can redo them in the current code.

## Loading the font
When reading the font, if bold or italic is requested, check if the font already provides those styles. to check for add this util to freetypeapi.cs, had to dig in freetype to confirm the values
```
static class FT_Style_Flags
    {
        public const int FT_STYLE_FLAG_ITALIC = 0b01;
        public const int FT_STYLE_FLAG_BOLD   = 0b10;
        public static bool HasStyle(nint styles, int style)
        {
            return (styles & style) != 0;
        }
    }
```
Because, the font has the styles in a clong flags field, pass HasStyle the field CLong.value
After getting the rendered glyph, 
##Italization
during the italize method, the bitmap is skewed 20º the right, 
the output of the `tan()` gives how much each row should be moved to right.
Because the bitmap needs to be wider, `(heigth of the bitmap * tan)` rounded up to the next integer, is how much wider. Also, to avoid rendering ultra wide text, remove this amount from the horizontal advance of the glyph.
For the italization per se, copy the bitmap from left to rigth and bottom to top, for each row calculate the displacement `(tan * row index)` being 0 the bottom row. The integer part is the pixel offset for the rendered pixel, the decimal part is how much filling to put in that pixel, the rest goes to the next pixel. Always use add and clamp the values to avoid overflow, as many pixels will have parts of 2 different pixels.
Keep in mind that row 0 is special, and should be copied as is.

## Emboldening
Just make the glyph a pixel wider and for each pixel copy it to the same position and to the one on its right
The part of draw3times can be omitted as it might make the text ilegible, it was only meant for bigger fonts.

SimonDarksideJ:
Changes migrated to #8872, closing

PR #4564, opened 2016-02-24
[MD Templates] Improved Program.cs and Game1.cs

cra0zy:
Stuff done:
 - files were full of both tabs and spaces, so replaced tabs with spaces
 - there were also tabs/spaces at the end of some lines... removed those
 - game will now start windowed by default
 - Program.cs is now calling the game withing `using` block

CC. @dellis1972

### DIFF
--- a/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/Program.cs
+++ b/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/Program.cs
@@ -13,69 +13,69 @@ using UIKit;
 
 namespace ${Namespace}
 {
-	#if __IOS__ || __TVOS__
-	[Register("AppDelegate")]
-	class Program : UIApplicationDelegate
-	#else
-	static class Program
-	#endif
+    #if __IOS__ || __TVOS__
+    [Register("AppDelegate")]
+    class Program : UIApplicationDelegate
+    #else
+    static class Program
+    #endif
     {
-        private static Game1 game;
-
-		internal static void RunGame() 
-		{
-			game = new Game1();
-			game.Run();
-		}
-			
+        internal static void RunGame() 
+        {
+            using (var game = new Game1())
+            {
+                game.Run();
+            }
+        }
+            
         /// <summary>
         /// The main entry point for the application.
         /// </summary>
-		#if !MONOMAC && !__IOS__  && !__TVOS__		 
+        #if !MONOMAC && !__IOS__  && !__TVOS__
         [STAThread]
-		#endif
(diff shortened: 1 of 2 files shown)


IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/Program.cs
```diff
@@ -13,69 +13,71 @@
...
 
-		internal static void RunGame() 
-		{
-			game = new Game1();
-			game.Run();
-		}
-			
+        internal static void RunGame() 
+        {
+            using (game = new Game1())
```
dellis1972: might as well remove the static and replace it with var if we are going to use a using

cra0zy: Updated ^^

IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/Program.cs
```diff
@@ -13,69 +13,69 @@
...
-		{
-			game = new Game1();
-			game.Run();
-		}
-			
+        internal static void RunGame() 
+        {
+            using (var game = new Game1())
+            {
+                game.Run();
```
dellis1972: @cra0zy unfortunately this change breaks iOS :( Because on iOS we don't stay in the Run method until the game exits we exit immediately. As a result game is free'd and boom!

We'll need to go back to static.

cra0zy: We need separate methods for Desktop/iOS, on desktop we should be doing the above to get proper Disposing.

On the side note, I think the whole Program.cs file is a mess, and that we should be using perplatform Program.cs instead of one filled with preprocessing params... just saying...

tomspilman: > On the side note, I think the whole Program.cs file is a mess, 

Yeah...  unfortunately the startup process for a bunch of platforms is weird and different from others.

We should open an issue on this specifically and discuss how to approach it.

cra0zy: I would say like VS, each platform gets their own.

tomspilman: Well I was wondering if we could move some of the work inside the MonoGame assembly and somehow simplify the entry point for everyone.

PR #4565, opened 2016-02-25
Make BitmapContent Width and Height setters protected

### LINKED ISSUE #4560: Possible bug in PixelBitmapContent.cs
I have been working on the renderer for font styles and was trying to reuse the current instance of `BitmapContent`. The problem arised when I tried to resize the bitmap, I began getting exceptions when calling `BitmapContent.SetPixelData(byte[])` because after setting the new size, the data array still has the old size.
First I thought that he properties `Width` and `Height` should be read only, but looking into the XNA documentation both of them are read and write properties.
Then I looked into the property implementation, and it does nothing apart from changing the corresponding field.
And last I saw the data container that is defined in this case in `PixelBitmapContent`, it is a jagged array with an array for the rows that contains more arrays for the data of each row. This makes enlarging the width or the height make the data wrong, and method `GetPixelData` and `SetPixelData` prone to memory corruption.

Also looking at other siblings of `PixelBitmapContent` like `DxtBitmapContent` or `Etc1BitmapContent` use a simple byte array for the data.

tgjones:
Make BitmapContent Width and Height setters protected. Fixes #4560.

Despite what MSDN [says](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.content.pipeline.graphics.bitmapcontent.width.aspx), XNA's `BitmapContent` does not have public `Width` and `Height` setters.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/BitmapContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/BitmapContent.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
             {
                 return height;
             }
-            set
+            protected set
             {
                 if (value <= 0)
                     throw new ArgumentOutOfRangeException("height");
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks @tgjones !

PR #4568, opened 2016-02-26
Implement vertexStride in VertexBuffer.SetData for OpenGL

tgjones:
This does for OpenGL what #3356 did for DirectX.

Previously, it would only have worked when `vertexStride == sizeof(T)`. This code path remains unchanged. The new code is for when `vertexStride != sizeof(T)`.

This change is covered by existing tests in `VertexBufferTest`: `SetPosition` and `SetTextureCoordinate`. These tests previously failed for OpenGL, and they now pass.

### DIFF
--- a/MonoGame.Framework/Graphics/Vertices/VertexBuffer.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/Vertices/VertexBuffer.OpenGL.cs
@@ -157,8 +157,21 @@ namespace Microsoft.Xna.Framework.Graphics
             var dataHandle = GCHandle.Alloc(data, GCHandleType.Pinned);
             var dataPtr = (IntPtr)(dataHandle.AddrOfPinnedObject().ToInt64() + startIndex * elementSizeInBytes);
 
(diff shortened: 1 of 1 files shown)


amerkoleci:
Yes I know, the problem that you would first need to call GetBufferSubData, which is not supported under GLES then replace the content applying the stride, what about removing vertexStride parameter, I know XNA supported it but probably will lead to performance issues if used on OpenGL.

tgjones:
> what about removing vertexStride parameter, I know XNA supported it but probably will lead to performance issues if used on OpenGL.

My engine is built around it :) Of course I could change it - but without someone measuring the performance on OpenGL and finding it to be a problem, I don't see the need to remove anything. It's a useful feature - and it's not like you have to use it, you can just set all the data at once if you prefer, as everyone has done up till now.

tgjones:
> I know XNA supported it but probably will lead to performance issues if used on OpenGL.

Also worth bearing in mind that setting vertex buffer data is usually done at first load, and certainly not every frame.

amerkoleci:
Agree, but you can also have dynamic vertex buffer and set portion of vertex data (particle system) every frame, that will hurt, anyway its ok for me :)

tomspilman:
Thanks @tgjones !

Issue #4494, opened 2016-02-03
Remove VertexPosition2ColorTexture add VertexPosition


tomspilman:
I think we can remove the unused `VertexPosition2ColorTexture` type...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Vertices/VertexPosition2ColorTexture.cs

This I think is a legacy type left around from some old "optimization" in `SpriteBatch`.  I don't think anything uses it anymore.

At the same time we should consider adding a `VertexPosition` type.  This is the most basic type supported by `BasicEffect`:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/Resources/Structures.fxh#L11

And there has never been a built in type in XNA for this, but I think there should be.

@KonajuGames  @dellis1972 ?

KonajuGames:
If BasicEffect supports it, it makes sense to add it. And I see no harm in
removing that other type.

dellis1972:
:+1:

tgjones:
This can be closed now.

tomspilman:
Leave a comment

PR #4569, opened 2016-02-26
Remove VertexPosition2ColorTexture and VertexColorTexture, and add VertexPosition

tgjones:
Implements #4494


tomspilman:
This is great!

Before we merge is there any useful unit tests we can add specifically for testing vertex formats?  I guess general things like testing the equality operators work, the vertex declaration is right and has the expected stride, etc.

@dellis1972 @KonajuGames   ??

tgjones:
I've added some basic unit tests for all the vertex types.

tomspilman:
Fantastic... merging!

PR #4570, opened 2016-02-26
Remove PerformanceCounter

### LINKED ISSUE #4518: Keep or Remove PerformanceCounter
I ran into this random class in MonoGame tonight `PerformanceCounter`:

https://github.com/mono/MonoGame/blob/0b25f56e861fb602db884c40267b99ea947bae15/MonoGame.Framework/PerformanceCounter.cs

Should we remove this?  Should we keep it and improve it?

IMO the class needs to be good enough to be used within MonoGame itself if we want to keep it.  It has to be easy to turn on/off... ideally at runtime... with zero cost when off.

tgjones:
Fixes #4518


tomspilman:
Perfect... thanks @tgjones !

PR #4572, opened 2016-02-26
Run Installer Update

cra0zy:
Does 2 things:
 - Makes ffmpeg and ffprobe executable after installation
 - Removes automatic dependency installer and adds a list of packages that are needed... much better since we won't have to add per distro stuff do the installer, chances are the people were running a distro that we didn't have automatic dep installer for, plus now people will get the list of needed packages during the install process

CC. @dellis1972


Installers/Linux/RUN/postinstall.sh
```diff
@@ -16,8 +16,18 @@ fi
 DIR=$(pwd)
 IDIR="/usr/lib/mono/xbuild/MonoGame/v3.0"
 
-# Automatic dependency installer
-./Dependencies/dependencies.sh
+# Show dependency list
+echo "Please make sure the following packages are installed:"
+echo " - monodevelop"
```
dellis1972: @cra0zy don't they need this before they install? Otherwise the .mpack will not get setup?
Any idea how to deal with that.

Installers/Linux/RUN/postinstall.sh
```diff
@@ -16,8 +16,18 @@ fi
...
+echo " - libopenal-dev"
+echo " - referenceassemblies-pcl / mono-pcl"
```
dellis1972: @DDReaper we are saying we need these :) Means the .Core project can be done for Linux to IMHO ;)

Installers/Linux/RUN/postinstall.sh
cra0zy: This message appears at the start of the installer.

Installers/Linux/RUN/postinstall.sh
```diff
@@ -16,8 +16,18 @@ fi
...
+echo " - ttf-mscorefonts-installer / mscore-fonts"
+echo " - gtk-sharp3"
+read -p "Continue (Y, n): " choice2
```
cra0zy: @dellis1972 Just about the question above, they get an option to cancel the installer if they don't have packages installed.

cra0zy:
Don't merge yet, I made a mistake.

cra0zy:
Ok, fixed.

tomspilman:
This look good to go @dellis1972 ?

dellis1972:
Merging thanks :)

PR #4573, opened 2016-02-26
Fix the MonoMac Template so it runs

dellis1972:
The MonoMac project was missing a MainMenu item
which was causing the application to crash.
Has to add back the use of Nuget because MonoMac projects
to NOT use MSBuild/xbuild so it cannot use the new .props
system.

### DIFF
--- a/IDE/MonoDevelop/MonoDevelop.MonoGame.Mac/MonoDevelop.MonoGame.Mac.csproj
+++ b/IDE/MonoDevelop/MonoDevelop.MonoGame.Mac/MonoDevelop.MonoGame.Mac.csproj
@@ -80,5 +80,8 @@
     <Content Include="templates\MonoGameOSXProject.xpt.xml">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="templates\Mac\MainMenu.xib">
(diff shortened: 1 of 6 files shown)


IDE/MonoDevelop/MonoDevelop.MonoGame.Mac/templates/MonoGameOSXProject.xpt.xml
```diff
@@ -27,6 +27,9 @@
 				<Reference type="Gac" refto="XamMac" SpecificVersion="false" />
 				<Reference type="Package" refto="MonoGame.Framework" />
 			</References>
+			<Packages>
+				<Package id="MonoGame.Framework.MacOS" version="3.3.0" targetFramework="net40" />
```
tomspilman: Does that mean the 3.3 release of MonoGame?

dellis1972: Yes, Ieadly I want to be able to use the Nuget's we build during the build process. Is that possible now ? Do the Nugets get built before the Addin/Installer? Are they available as archives?

dellis1972: ideally I meant...

KonajuGames:
Merging.

PR #4577, opened 2016-02-28
Fixed Minimum framework version for iOS

dellis1972:
The Minumum framework version needs to be 5.1.1 in order to
compile the iOS app. It also needs to be that version to publish
to the maximum number of devices.


dellis1972:
@KonajuGames :)

KonajuGames:
Did you get the warning that it should be a minimum of 7.0 for incremental builds to work, and also to work with iOS 9, and that the Xamarin Studio set it to 7.0 anyway?  I think we should set it to 7.0 as a minimum to avoid this warning and because no-one in their right mind should be creating a new iOS project targeting iOS 5 anyway.

dellis1972:
@KonajuGames done . changed to 7.0

KonajuGames:
Merging.

PR #4579, opened 2016-02-28
Fixed a typo in the Xamarin Studio Addin

dellis1972:
The code that was injecting the monogame .props file
had a tyyo in the Codition property. This was causing
the build to fail on Windows.

This commit fixes that issue


dellis1972:
@KonajuGames fixed the issue mentioned in Gitter :)

KonajuGames:
Merging.

PR #4581, opened 2016-02-29
Fix Mac Installer to place the OSX assemblies in the correct location.

dellis1972:
They should have been installing to 'MacOSX' rather than 'MacOS'
because 'MacOSX' is what we use to describe the platform.


dellis1972:
@KonajuGames I think we need this one in ASAP

KonajuGames:
Merging.

Issue #4311, opened 2015-12-09
Remove Storage Classes


tomspilman:
I am proposing the full removal of the existing storage classes:

 - [StorageContainer.cs](
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Storage/StorageContainer.cs)
 - [StorageDevice.cs](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Storage/StorageDevice.cs)
 - [StorageDeviceHelper.cs](
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Storage/StorageDeviceHelper.cs)

Back in https://github.com/mono/MonoGame/issues/2389 I was suggesting redesign, but really after seeing more console SDKs and how each one uses a different method for storage (as well as different TRC requirements)...  it seems better to just drop support for this old API entirely.

Then at a later date if we can work out something that works in all cases we can talk about adding a new API.

So is anyone using this functionality?  Would it be missed at all?

mrhelmut:
I would agree.

I never used them because console SDKs / TRCs are so different and some of them don't even have a folder based storage paradigm, not to mention the difference between async and blocking calls to each SDK (hence, implementing these classes may be quite touchy on some platforms).

KonajuGames:
The Storage namespace is very much targeted towards the Xbox 360, memory cards and the Guide.  And the whole Begin..End async model needs to be updated to async/await.

I have always used IsolatedStorage for PC and phones.

mgarstenauer:
I agree - the storage classes are not needed anymore.

lolznoob:
I use them!

PR #4585, opened 2016-02-29
[MG Framework] Make CurrentPlatform public

cra0zy:


MonoGame.Framework/Utilities/OS.cs
```diff
@@ -0,0 +1,46 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+namespace MonoGame.Utilities
+{
+    /// <summary>
+    /// Defines the operating system.
+    /// </summary>
+    public enum OS
```
hach-que: I'm not sure this is the best name, `OS` is particularly short.

What's wrong with the existing platform enumeration used for the content pipeline?

cra0zy: > What's wrong with the existing platform enumeration used for the content pipeline?

It's per platform instead of per OS. This returns the current OS, and not the platform, WindowsDX, DesktopGL, UWP and Metro can all return Windows as OS.

cra0zy: Also I added this mostly so people using DesktopGL can get if they are running on Windows, Linux or Mac.

hach-que: Isn't the entire argument of DesktopGL that people shouldn't be writing platform specific code?

They should be using `#if PLATFORM_WINDOWS` and the equivalent for other platforms if they want to write platform-specific code.

> This returns the current OS, and not the platform, WindowsDX, DesktopGL, UWP and Metro can all return Windows as OS.

This feels pretty arbitrary, since the native APIs available on a phone under UWP are going to be different to the native APIs available under Win32.

cra0zy: > Isn't the entire argument of DesktopGL that people shouldn't be writing platform specific code?
> 
> They should be using #if PLATFORM_WINDOWS and the equivalent for other platforms if they want to write platform-specific code.

People might still want to be able to check current OS from code. For example I am working on a game where when an error occurs I made it load fake "Blue screen of death" if on Windows, or "Kernel panic" if on Unix. Yes, I could have used "#if PLATFORM_WINDOWS" but then my .exe wouldn't be cross platform.

hach-que: That seems pretty game-specific and I'm not sure it warrants adding yet another way to check "platform" when there's already the platform enumeration provided by MonoGame, and the .NET framework provides several ways to detect it at runtime.

cra0zy: > That seems pretty game-specific and I'm not sure it warrants adding yet another way to check

This will come in very handy when Storage stuff gets removed(#4311).

> when there's already the platform enumeration provided by MonoGame

Again, platform and OS are two completely different things...

> and the .NET framework provides several ways to detect it at runtime.

Checking if it's running on Linux or Mac is a pain. 

Also there really is no negative effect in making the class public... the code is there regardless, it will just make it easier if someone wants to access the information.

hach-que: Well it has to be maintained.

If you want to check at runtime for Windows/Mac/Linux, it's as simple as:

```
            if (Path.DirectorySeparatorChar == '/')
            {
                if (Directory.Exists("/Library"))
                {
                    return "MacOS";
                }

                return "Linux";
            }

            return "Windows";
```

anyone who needs to detect it at runtime can do it.

It just seems really arbitrary to be adding another platform detection class to MonoGame.

cra0zy: If some Linux user has a need to create a Library folder your code will fail, I don't like adding code that has a chance of failure with something so simple unless I have to.

cra0zy:
@tomspilman @dellis1972 Thoughts on this?

silasary:
I'd prefer the name to be OperatingSystem, rather than just OS. 
But otherwise yes, this should be something that the framework should include, rather than expecting devs to create hacky checks like the one provided above.

MonoGame.Framework/Utilities/OS.cs
KonajuGames: > If you want to check at runtime for Windows/Mac/Linux, it's as simple as:

That's not "simple" by any means, or reliable enough.  Because Mono gets it wrong, we're supposed to magically know the multi-line work-around that depends on an arbitrary folder existing?

Using `#if` goes against the entire point of DesktopGL using the same executable across Windows, Mac and Linux.  And I believe there will always be a need to know which platform you are running on because there will always be something different on each platform, even just some integration with the OS or similar.

hach-que: This is the check used by Protobuild, and it works fine for platform detection. There are no major Linux distributions that ship a Library folder. 

Having any platform specific code goes against the idea of DesktopGL, which is my point. The game can't actually have different references or use platform specific APIs (other than P/Invoke) in a DesktopGL executable. You can only really do cosmetic things like described above.

cra0zy: > You can only really do cosmetic things like described above.

Setting default storage folder? Keyboard input? On Windows and Linux Ctrl + C = Copy, but on Mac it's different, it's Command + C. Your point is pointless... sorry I couldn't resist saying it...

cra0zy:
> I'd prefer the name to be OperatingSystem, rather than just OS.

Done, now the enum is OperatingSystem.

cra0zy:
It took me only 11 days to notice that build failed on this PR :V

cra0zy:
I forgot why I used "OS" in the first place, there is already "OperatingSystem" under "System" namespace.... anyway reverted back to "OS" unless someone can give a better idea.

@tomspilman @KonajuGames @dellis1972 Can this be merged? what are your thoughts on this?

PR #4595, opened 2016-03-02
Fix for importing RGBA16 bitmap

Tzenchor:
Fixes #4592.
This is the simplest fix for the problem, convert the bitmap to 32 bits and continue from there.
Future bitmaps that are not contemplated in the special cases are treated as bitmaps, and converted to 32 bits bitmaps.
I have tried to preserve the higher color resolution of the original image, but `PixelBitmapContent.TryGetFormat` can't handle that option at this time, and this exposed a bug in `PixelBitmapContent.SetPixelData` it ignores the return value of `TryGetFormat`, using the default value.

In game test of the texture, it is a bit cropped at bottom because the window ended there
ingametex

This can be merged now, or I can try to enable the RGBA64 format later

PR #4596, opened 2016-03-02
Add exception to PixelBitmapContent constructor

Tzenchor:
If the call to TryGetFormat returns false inside the constructor, throw
an InvalidOperationException instead of continuing with the wrong
SurfaceFormat.
This bug was detected in #4595: If for the T parameter of `PixelBitmapContent` there is no matching `SurfaceFormat` the constructor defaults to `SurfaceFormat.Color` and continues like nothing has happened, leading to problems when using this instance


Tzenchor:
Sorry, I made the pull request to my repo before doing it right, and tried to delete the branch to delete the pull without thinking that would close this also

tomspilman:
No worries... it didn't hurt anything.

KonajuGames:
This looks good.  Merging.

PR #4597, opened 2016-03-04
Fix an issue where the Symlink for MonoGame was NOT created

dellis1972:
Fix an issue where the Symlink for MonoGame was NOT created in
   
 /Library/Frameworks/Mono.framework/External/xbuild

because the directory did not exist.


dellis1972:
@KonajuGames this came up on bitter yesterday we need this one in 3.5

dellis1972:
gitter even ;)

KonajuGames:
Can't have that folder not existing.  Merging.

Issue #4598, opened 2016-03-04
Window.Title has no effect in a UWP project


rohdester:
Setting of Window.Title in e.g. the Initialize method has no effect in a Universal Windows project.
It does work though in a Windows OpenGL project, so I reckon it has to do with the UWP SDK.

So steps to reproduce:

1. Open Visual Studio
2. Create new project choosing "MonoGame Windows 10 Universal Project"
3. In Game1.Initialize set Window.Title to "My Title";
4. Run the app, and we expect the window title to be "My Title" while it is in fact still the project title.

tomspilman:
I think at the time we ported to Win10 we didn't know the solution or it was available...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs#L242

... anyone know the right UWP call to set the window title and want to submit a PR for it?

rohdester:
I can do a PR for it.

tomspilman:
Fantastic... thanks @rohdester !

PR #4601, opened 2016-03-04
Implementation of SetTitle for UWP

rohdester:
Sets the title for the current view for Universal Windows Project apps.
It should be noted that this is meant to be the title for the current
view, and that UWP apps always will append the Display Name from the app
manifest to this title.

### DIFF
--- a/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs
+++ b/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 
 using Windows.UI.Core;
@@ -241,8 +242,7 @@ namespace Microsoft.Xna.Framework
 
         protected override void SetTitle(string title)
         {
-            var appView = ApplicationView.GetForCurrentView();
-            appView.Title = title;
+            Debug.WriteLine("WARNING: GameWindow.Title has no effect under UWP.");
         }
 
         internal void SetCursor(bool visible)
--- a/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs
+++ b/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs
@@ -241,8 +241,8 @@ namespace Microsoft.Xna.Framework
 
         protected override void SetTitle(string title)
         {
-            // NOTE: There is no concept of a window 
-            // title in a Metro application.
+            var appView = ApplicationView.GetForCurrentView();
+            appView.Title = title;
         }
 
         internal void SetCursor(bool visible)
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -42,7 +42,14 @@ namespace Microsoft.Xna.Framework {
 		public abstract string ScreenDeviceName { get; }
 
 		private string _title;
-		public string Title {
+        /// <summary>
+        /// Gets or sets the title of the game window.
+        /// </summary>
+        /// <remarks>
+        /// For Windows 8 and Windows 10 UWP this has no effect. For these platforms the title should be
+        /// set by using the DisplayName property found in the app manifest file.
+        /// </remarks>
+        public string Title {
 			get { return _title; }
 			set {
 				if (_title != value) {


tomspilman:
> this is meant to be the title for the current view, and that UWP apps always will append the Display Name from the app manifest to this title.

@rohdester - Can you give us a screenshot showing how this looks?

This issue might of been why I didn't include this in the first place.

Tzenchor:
I was also trying to accomplish this, and got to the same solution, with
'ApplicationView bar = ApplicationView.GetForCurrentView( );
            bar.Title = "kukusmu";
'
Gives
uwp title
The other solution might be making a custom title bar

tomspilman:
> The other solution might be making a custom title bar

Yeah....  this is why I didn't include it at first.  It wouldn't appear like people expected so I omitted it.

What would a custom title bar be like?  Are their limitations to this?

Tzenchor:
The idea came from [this link](https://blogs.msdn.microsoft.com/eternalcoding/2015/08/20/uwptake-the-control-of-your-title-bar/).

rohdester:
Been thinking about this, and you are probably right @tomspilman it would surprise people that the assembly name is appended, and they wont think about looking into the app manifest file.

Shall I look into using a custom title bar like @Tzenchor mentions and submit a new PR if it succeeds?

tomspilman:
> Shall I look into using a custom title bar like @Tzenchor mentions and submit a new PR if it succeeds?

Maybe...  don't want you to waste your time unless we know this is a good solution.

I worry that doing this will somehow make our UWP support incompatible with some UWP devices like phones or XB1.

Also likely this really only solves the Window title bar itself.  I suspect in the Alt+Tab menu and in the task bar your game Window still shows the app title from the manifest.  Which is also just as unexpected... worse now because it shows different things in different places.

I suspect there is really nothing we can do to fix this.  I'm fairly certain the reason MS moved this into the app manifest is to avoid applications spoofing others for malicious purposes.  So i sort of doubt this functionality is coming back under UWP... we should just learn to live with it.

@BrianPeek - Any other suggestions?

rohdester:
Good points. I guess for games setting the title in the manifest suffices for most purposes.

BrianPeek:
To my knowledge, you can do the whole ExtendViewIntoTitleBar thing, but it really just acts like an "overlay" on top of the real title bar.  So no, the title won't change in alt-tab, the taskbar, taskview, etc.  Just the window itself will have the new title drawn.  I agree with @tomspilman that it would seem a bit confusing.

If you did want to do it, though, you can use the ApiInformation.IsTypePresent to look for the proper type and only do it on Desktop so "bad things" don't happen on Mobile, IoT, etc.  More on that here:

https://blogs.msdn.microsoft.com/gianlucb/2015/10/08/uwp-windows-10-app-titlebar-and-status-bar-customization/

MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs
```diff
@@ -241,8 +241,8 @@ private void DisplayProperties_OrientationChanged(DisplayInformation dinfo, obje
 
         protected override void SetTitle(string title)
         {
-            // NOTE: There is no concept of a window 
-            // title in a Metro application.
+            var appView = ApplicationView.GetForCurrentView();
+            appView.Title = title;
```
tomspilman: How about we change this to something like:

```
Debug.WriteLine("WARNING: GameWindow.Title has no affect under UWP.")
```

... or something like this.  This way at least the user sees this in the output window when debugging and can be aware of it.

tomspilman:
> . So no, the title won't change in alt-tab, the taskbar, taskview, etc.

Thanks @BrianPeek ... that is what I figured.

@rohdester - How about we also fix the docs for `GameWindow.Title`:

http://www.monogame.net/documentation/?page=P_Microsoft_Xna_Framework_GameWindow_Title

First it could use the basic docs about what it should do (reminder... do not copy docs from the XNA doc site).  Second you can add in a `<remarks></remarks>` section sort of like this:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/GraphicsDeviceInformation.cs#L22

Which can note that the Title has no affect on Windows 8 and Windows 10 UWP platforms and to change the title via the application manifest.

Make sense?

rohdester:
@tomspilman Yup - that makes sense. I'll ad the WriteLine and look at the docs.

rohdester:
Has now been updated.
And that was for issue #4598

tomspilman:
Looks good...   thanks @rohdester !

PR #4603, opened 2016-03-05
Rgba64 support in PixelBitmapContent

KonajuGames:
- Added other packed vector types that had been omitted, though I'm not sure how to support NormalizedShort2 and 4 as there does not appear to be corresponding SurfaceFormat types.
- Updated Rgba64 importer test to expect SurfaceFormat.Rgba64.


tomspilman:
Fantastic... thanks @KonajuGames !

PR #4608, opened 2016-03-06
Fix DX template (wrong .NET Framework version)

KonajuGames:
Fix Android build SDK version.
Make all VS templates consistent.
- Same output directories (csproj and mgcb).
- Same Game1.cs (with only a few minor platform differences)


dellis1972:
@KonajuGames thanks :)

PR #4611, opened 2016-03-06
Change Shared Program.cs back to using a static to fix an issue in iOS.

dellis1972:
For MonoGame 3.6 we should revisit this kind of code to have platform
specific source files rather than this mess of #if statements. Generally
speaking developers do not share startup code.

### DIFF
--- a/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/Program.cs
+++ b/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/Program.cs
@@ -20,13 +20,16 @@ namespace ${Namespace}
     static class Program
     #endif
     {
-        internal static void RunGame() 
-        {
-            using (var game = new Game1())
-            {
-                game.Run();
-            }
-        }
+		private static Game1 game;
+
+		internal static void RunGame() 
+		{
+			game = new Game1();
+			game.Run();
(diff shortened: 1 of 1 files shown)


dellis1972:
@cra0zy I just reverted this stuff for now. I think we should revisit this after we manage to get 3.5 out of the door. Perhaps as part of the template re-write.

cra0zy:
> @cra0zy I just reverted this stuff for now. I think we should revisit this after we manage to get 3.5 out of the door. Perhaps as part of the template re-write.

Again, you need to dispose of the game on Desktop platforms, add a preprocessor param for it.

And we won't have to talk about Program.cs because Protobuild.Manager should sort it out for us.

tomspilman:
> And we won't have to talk about Program.cs because Protobuild.Manager should sort it out for us.

The generated template source code should always support all our platforms.  I don't want to see Protobuild.Manager copying over different source files for different platforms.

cra0zy:
> The generated template source code should always support all our platforms. I don't want to see Protobuild.Manager copying over different source files for different platforms.

The Game.cs should be the same, but I don't think Program.cs is that important...

tomspilman:
> but I don't think Program.cs is that important to be the same

I'd rather have it all the same if possible... and I think we could make it possible.

KonajuGames:
Merging.

PR #4614, opened 2016-03-07
[Nunit] Fixed few Unix tests

cra0zy:


Test/ContentPipeline/IntermediateDeserializerTest.cs
```diff
@@ -404,13 +404,15 @@ public void ExternalReferences()
         {
             Deserialize<ExternalReferences>("17_ExternalReferences.xml", externalReferences =>
             {
+                var pathSeparator = Path.DirectorySeparatorChar.ToString();
+
                 Assert.NotNull(externalReferences.Texture);
-                Assert.IsTrue(externalReferences.Texture.Filename.EndsWith(@"\Xml\grass.tga"));
+                Assert.IsTrue(externalReferences.Texture.Filename.EndsWith(pathSeparator + "Xml" + pathSeparator + "grass.tga"));
```
tomspilman: Can we just make these forward slashes in the xml file?  I don't think that would make the test worse and it should work on both Windows and non-Windows platforms.

cra0zy: I am confused by what you mean by that... if you mean the "17_ExternalReferences.xml" xml file, that file contains file name, not file path...

cra0zy: I think I found a nicer way of doing this, good now?

KonajuGames: I understand how this works now.  The filename stored in the object created by the deserialization gets the platform-specific path separators.  In XNA of course it used Windows path separators, but other platforms get the forward slashes.  So we need the test to check for the correct platform path separators.

Build/Projects/MonoGame.Framework.definition
```diff
@@ -136,6 +136,7 @@
       <!-- Covers compatible GL implementations, such as ANGLE and OpenGL -->
     </Service>
     <Service Name="_XNADesignProvided">
+      <AddDefine>XNADESIGNPROVIDED</AddDefine>
```
tomspilman: @cra0zy - What is the reason for this?

cra0zy: Vector classes

tomspilman: I saw that...  just wasn't sure what the problem is.

My guess is that previously we were only setting the type converters on Windows builds and with this it now appears on Mac and Linux too?

cra0zy: Yes, plus in the future if we want to add more platforms to it, all we have to do is just connect the service, no need to modify any .cs files.

tomspilman:
Thanks @cra0zy !

PR #4615, opened 2016-03-07
Correct an error in the MonoMac template.

dellis1972:
It was referencing a non-existent nuget package


tomspilman:
Thanks @dellis1972 !

Issue #4617, opened 2016-03-08
Is there any reason why Buttons doesn't contain Buttons.None?


BaseCrosser:
As the title says, is there any reason why there is no Buttons.None in MonoGame?

Greetings,
BaseCrosser

tomspilman:
I suspect because there wasn't a `Buttons.None` in XNA:

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.input.buttons.aspx

But I don't see harm in adding one.

BaseCrosser:
I created a PR for it.

daglundberg:
The pull request was closed after discussion: https://github.com/MonoGame/MonoGame/pull/4618

Does that mean this can be closed too or no?

Best

BaseCrosser:
I will close this now

PR #4618, opened 2016-03-08
Add Buttons.None to Buttons.

BaseCrosser:
Issue #4617

### DIFF
--- a/MonoGame.Framework/Input/Buttons.cs
+++ b/MonoGame.Framework/Input/Buttons.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Xna.Framework.Input
         /// <summary>
         /// No button.
         /// </summary>
-        None = 0,
+        None = 3,
         
         /// <summary>
         /// Directional pad up.
--- a/MonoGame.Framework/Input/Buttons.cs
+++ b/MonoGame.Framework/Input/Buttons.cs
@@ -13,6 +13,11 @@ namespace Microsoft.Xna.Framework.Input
     public enum Buttons
     {
         /// <summary>
+        /// No button.
+        /// </summary>
+        None = 0,
+        
+        /// <summary>
         /// Directional pad up.
         /// </summary>
         DPadUp = 1,


tomspilman:
@KonajuGames @dellis1972 - If you think this is ok... we can merge it after the 3.5 release.

cra0zy:
This PR only adds a button, it doesn't actually connect it with the code... It will always be false.

tomspilman:
My thinking was the intent was for something like:

```
if (GamePad.GetState().Buttons.Value == Buttons.None)
   ...
```

But I thing what @cra0zy is getting at is that the input API never returns the `Buttons` state anywhere.

I guess it means you can now do `new GamePadButtons(Buttons.None)` where before you had to do `new GamePadButtons(0)`.

Tzenchor:
This should work through `GamePadState.IsButtonDown`

BaseCrosser:
My intent was to have a empty value in Buttons like there is in Keys.
So that it returns ``` false ``` when you do ``` GamePad.GetState().IsButtonDown(Buttons.None) ```.
I didn't actually test this because I don't have access to my dev computer right now.
If my thinking is wrong, what exactly should be changed to make it work like I imagine?

Tzenchor:
The problem is that `GamePad.GetState().IsButtonDown` implementation is doing a bitwise `&` comparison between the pressed buttons and the parameter, but `Anything & 0` returns 0, and 0 == 0 is always true

KonajuGames:
It would be a way of removing a key binding without have an accompanying bool to say it has been bound to a button, or having to typecast `0` to a non-existent `Buttons` enum value.

KonajuGames:
> The problem is that GamePad.GetState().IsButtonDown implementation is doing a bitwise & comparison between the pressed buttons and the parameter, but Anything & 0 returns 0, and 0 == 0 is always true

It would need a special case for Buttons.None.  This gets a little messy, with potential for adding bugs.

BaseCrosser:
And what if I just change the value of None to 3. Then it wouldn't be 0 anymore and this should work.
I tested it with this code:
    if (GamePad.GetState(PlayerIndex.One).IsButtonDown((Buttons)3))
        Exit();

None of the buttons on the game pad triggered it so I assume it is working then.

Tzenchor:
A "3" are two different buttons at the same time, the ones with value 1 and 2 concretelly.
Also querying `GamePad.GetState().IsButtonDown(Buttons.None)` will always returns false unless those two buttons are pressed at the same time

BaseCrosser:
The buttons with value 1 and 2 are DpadUp and DpadDown.
Those 2 can't be pressed at the same time so Buttons.None will always return false just like I intended.
It is as @KonajuGames  says a way to remove a key binding.
That is also the reason I am proposing this.

BaseCrosser:
I will close this now so that somebody else can add it correctly.

PR #4621, opened 2016-03-09
Reordered Member to match Style and Coding guidelines

MaxPlay:
The Microsoft.XNA.Framework.Rectangle struct had the Property "Center" at the wrong location with the wrong formatting.


CartBlanche:
@MaxPlay Thanks for the PR, but as you can see it doesn't pass any of our checks and is too big (196 commits and 733 files changed) a PR to accept. As your actual change is quite small, can I suggest you either create a totally new fork and send the PR for that or you may get away with doing a rebase. A new fork of develop will probably be quickest and cleanest.

MaxPlay:
Actually I'm not that good with Git and don't really know what to do and why this extreme stuff happend. I changed one file, basically 10 lines of code and this created such a huge bloat of stuff. I try to do what you suggested, though.

MaxPlay:
This bug is nonexistant in develop. There is no need for this PR.

CartBlanche:
@MaxPlay You are in luck, we have some explanations about how to do most things in git, for anyone new to git and should also be useful in other OpenSource projects. See the info here https://github.com/mono/MonoGame/wiki/Git'ing-Started-With-Git about rebasing etc.
If you have questions just ask on our forums.

We look forward to future contributions from you.

PR #4627, opened 2016-03-13
Fix for 'Sound effects are corrupt in Windows' - Issue 4626

SteveDunn:
https://github.com/mono/MonoGame/issues/4626

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -139,33 +139,34 @@ namespace Microsoft.Xna.Framework.Audio
         {
             CreateBuffers(  new WaveFormat(sampleRate, (int)channels),
                             DataStream.Create(buffer, true, false),
-                            0, 
-                            buffer.Length);
+                            0);
         }
 
+        // steve@dunnhq.com - the loopLength isn't needed by the code but is needed for compilation on other platforms as
+        // other implementations in this partial class have this parameter and we can't change the signature in just one place.
         private void PlatformInitialize(byte[] buffer, int offset, int count, int sampleRate, AudioChannels channels, int loopStart, int loopLength)
         {
             CreateBuffers(  new WaveFormat(sampleRate, (int)channels),
                             DataStream.Create(buffer, true, false, offset),
-                            loopStart, 
-                            loopLength);
+                            loopStart);
         }
 
         private void PlatformLoadAudioStream(Stream s)
         {
             var soundStream = new SoundStream(s);
             var dataStream = soundStream.ToDataStream();
-            var sampleLength = (int)(dataStream.Length / ((soundStream.Format.Channels * soundStream.Format.BitsPerSample) / 8));
+
             CreateBuffers(  soundStream.Format,
                             dataStream,
-                            0,
-                            sampleLength);
+                            0);
         }
 
-        private void CreateBuffers(WaveFormat format, DataStream dataStream, int loopStart, int loopLength)
+        private void CreateBuffers(WaveFormat format, DataStream dataStream, int loopStart)
         {
             _format = format;
             _dataStream = dataStream;
+
+            int sampleLength = (int)(dataStream.Length / (format.Channels * format.BitsPerSample / 8));
 
             _buffer = new AudioBuffer
             {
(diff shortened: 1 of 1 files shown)


SteveDunn:
This is my first PR here.  The build failure *look* like they're related to .NET Framework differences.  The projects in the **develop** branch are targeting **.NET 4.5**, but I notice from the build logs on TC, this:

```
[00:57:12][Step 1/1] Buildfile: file:///C:/BuildAgents/MonoGameWin1/work/f26e5a04a66fcbdb/default.build
[00:57:12][Step 1/1] Target framework: Microsoft .NET Framework 4.0
[00:57:12][Step 1/1] Target(s) specified: build_windows build_web build_android build_windows8 build_windowsphone build_windowsphone81 build_windows10 

```

SteveDunn:
I found the issue:

`[00:57:19]W:		 [NAnt output]      [exec]   Audio\SoundEffect.cs(57,13): error CS1501: No overload for method 'PlatformInitialize' takes 6 arguments [C:\BuildAgents\MonoGameWin1\work\f26e5a04a66fcbdb\MonoGame.Framework\MonoGame.Framework.WindowsGL.csproj]
`
The last argument, `loopLength` is never used, so I removed it (it's now calculated).  Will fix shortly

KonajuGames:
Some WAV files can have loop markers, which is where the loopLength parameter is used, and it may be less than the length of the entire sample.  Removing it is not a fix.

SteveDunn:
But it's not used in that method...?

tomspilman:
Yeah... removing `loopLength` isn't a fix we can accept.

MonoGame.Framework/Audio/SoundEffect.XAudio.cs
```diff
@@ -139,41 +139,42 @@ private void PlatformInitialize(byte[] buffer, int sampleRate, AudioChannels cha
         {
             CreateBuffers(  new WaveFormat(sampleRate, (int)channels),
                             DataStream.Create(buffer, true, false),
-                            0, 
-                            buffer.Length);
+                            0);
```
tomspilman: For sure this code is bugged...  it is passing loop length as the buffer length which is incorrect.  This code should be correctly calculating the length in samples and passing it down.

tomspilman:
@SteveDunn - Can you tell us how you were loading this sound that played incorrectly?  Was it via `SoundEffect` constructors?  `SoundEffect.FromStream()`?

SteveDunn:
@tomspilman - I load via ```content.Load<SoundEffect>(@"SoundEffects/shoot");``` - which, internally, does reads from a stream.

As an aside, looking at ```ContentManager::Load<T>```, there's an unnecessary creation of an object on line 204 - it's created but never used.

tomspilman:
> load via  content.Load<SoundEffect>(@"SoundEffects/shoot");

Are you processing the sound thru the content pipeline?  Or are you using the fallback which reads a raw wav file?

SteveDunn:
Hi Tom, through the content pipeline.

tomspilman:
Can you share the sound file with us so we can examine it?

There is no reason why a file loading thru the pipeline should have this issue.  The bug likely is in the pipeline side.

KonajuGames:
I created a DirectX project in Windows, added a .wav to the content project, added code to play that sound effect, and everything worked as expected.  It would be good to get your .wav file that is showing the issue because I cannot seem to to replicate the issue here with a sampling of my sound effects I have available.

SteveDunn:
I'll create a test project that has the issue. Will send on later in the
week.  Cheers

tomspilman:
> Will send on later in the week

We are trying to ship MonoGame 3.5 like today or tomorrow.  If you can get us just the WAV file now that would help a lot.

Another thought I had is maybe you are using an old version of the pipeline?  If you run the Pipeline tool and look in the About box does it match the version of the MonoGame.Framework.dll you are using?

SteveDunn:
I uploaded the wav about an hour ago. Here's a link instead
https://onedrive.live.com/redir?resid=1979AFB075E84971!17691&authkey=!AB8GHQU4Chgzjms&ithint=file%2cwav

Will send on later in the week

We are trying to ship MonoGame 3.5 like today or tomorrow. If you can get
us just the WAV file now that would help a lot.

Another thought I had is maybe you are using an old version of the
pipeline? If you run the Pipeline tool and look in the About box does it
match the version of the MonoGame.Framework.dll you are using?

—
Reply to this email directly or view it on GitHub
<https://github.com/mono/MonoGame/pull/4627#issuecomment-196285718>.

KonajuGames:
#4633 fixes the content pipeline processing of 8-bit wav as source audio.

PR #4630, opened 2016-03-13
Revert "Fix mouse coordinate for DesktopGL and windowed mode"

cra0zy:
Reverts mono/MonoGame#4501

Side note, we should also get OpenTK updated, now MonoGame/opentk needs https://github.com/TD25/opentk/commit/e5ab1ae791ef07750971bc8a88e250c50cace268 from opentk/opentk.

### DIFF
--- a/MonoGame.Framework/Input/Mouse.cs
+++ b/MonoGame.Framework/Input/Mouse.cs
@@ -140,9 +140,9 @@ namespace Microsoft.Xna.Framework.Input
 #elif DESKTOPGL || ANGLE
 
             var state = OpenTK.Input.Mouse.GetCursorState();
-            
-            window.MouseState.X = state.X;
-            window.MouseState.Y = state.Y;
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Input/Mouse.cs
```diff
@@ -140,9 +140,9 @@ public static MouseState GetState(GameWindow window)
 #elif DESKTOPGL || ANGLE
 
             var state = OpenTK.Input.Mouse.GetCursorState();
-            
-            window.MouseState.X = state.X;
-            window.MouseState.Y = state.Y;
+            var pc = ((OpenTKGameWindow)window).Window.PointToClient(new System.Drawing.Point(state.X, state.Y));
```
KonajuGames: I was testing a fix on Windows, Mac and Linux.

    var clientBounds = window.ClientBounds;
    window.MouseState.X = state.X - clientBounds.X;
    window.MouseState.Y = state.Y - clientBounds.Y;

Does the same thing, and cleaner.  Tested on Windows and Linux.  Couldn't get the project to build on Mac.  Some Xamarin Studio issue I think.

PR #4631, opened 2016-03-13
[DesktopGL] Fixed getting mouse position

cra0zy:
On Linux we must make sure we are not using SDL because there is a bug with SDL on getting client bounds, it's been fixed in their develop release, but it's not in stable.

CC. @KonajuGames

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -80,6 +80,8 @@ using Microsoft.Xna.Framework.Input;
 using OpenTK;
 using OpenTK.Graphics;
 
+using MonoGame.Utilities;
+
 namespace Microsoft.Xna.Framework
 {
     class OpenTKGamePlatform : GamePlatform
@@ -95,6 +97,9 @@ namespace Microsoft.Xna.Framework
 		public OpenTKGamePlatform(Game game)
             : base(game)
         {
+            if (CurrentPlatform.OS == OS.Linux)
+                Toolkit.Init(new ToolkitOptions { Backend = PlatformBackend.PreferNative });
+
             _view = new OpenTKGameWindow(game);
             this.Window = _view;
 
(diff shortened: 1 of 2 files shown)


KonajuGames:
Merging.  Thanks @cra0zy

mrhelmut:
This has broken mouse position on Windows.


While OpenTK is still being used, I think we should add an option somewhere to allow end-users to choose which backend is used on each platforms. It looks like everybody have different problems on some platforms depending on the backend. For instance, not using SDL on Linux breaks gamepad support in our game, and exposes windowing issues (same on OS X).

KonajuGames:
Mouse coordinates are working fine in WindowsGL here.  Is there any more to elaborate on?

mrhelmut:
Using SDL2 and windowed mode, mouse coordinates are off by the window's position.

KonajuGames:
So the SDL2 backend isn't the default.  Can't be because a stock project works fine.  Now how do I set this to use SDL2 to play with it?

mrhelmut:
It is true that SDL2 returns wrong window position on all platforms, but both changing this code and forcing the native backend isn't a good move in my opinion. It breaks more things than it solves (in my personal case).

To test the SDL2 backend, SDL2.dll must be present alongside the executable.

I know that all these backend differences are annoying, but they are advantages to using both.

If you don't mind, I'd like to add a ```if (RunningOnSDL2)``` here, and add a parameter to ```Game``` like ```PreferNativeBackend``` to DesktopGL.

KonajuGames:
I've got this for mouse updates now.

            // SDL2 backend doesn't support mouse coordinates outside the window, but GetCursorState uses the window client top/left as the origin.
            // Native backend does support mouse coordinates outside the window, but GetCursorState uses the screen top/left as the origin.
            if (OpenTK.Configuration.RunningOnSdl2)
            {
                window.MouseState.X = state.X;
                window.MouseState.Y = state.Y;
            }
            else
            {
                var clientBounds = window.ClientBounds;
                window.MouseState.X = state.X - clientBounds.X;
                window.MouseState.Y = state.Y - clientBounds.Y;
            }

Thinking about how to best expose the preferred backend without going too deep.

KonajuGames:
Have a look at #4632.  It should cover these cases.  ClientBounds offset is not applied if SDL2 is being used.  Allows the preferred backend to be set prior to construction of the Game object.

PR #4632, opened 2016-03-14
Mouse coords in OpenTK again.

KonajuGames:
Do not do the client bounds offset if SDL2 is being used.
Added PlatformParameters to DesktopGL platforms where the preferred OpenTK backend can be specified.  Must be set prior to construction of the Game object.

Ref: #4631

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -108,7 +108,7 @@ namespace Microsoft.Xna.Framework
         /// <summary>
         /// The preferred backend for OpenTK to use.
         /// </summary>
-        public static Backend PreferredBackend = Backend.Default;
+        public static Backend PreferredBackend = MonoGame.Utilities.CurrentPlatform.OS == OS.Linux ? Backend.Native : Backend.Default;
     }
 
     class OpenTKGamePlatform : GamePlatform
--- a/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGamePlatform.cs
@@ -84,6 +84,33 @@ using MonoGame.Utilities;
 
 namespace Microsoft.Xna.Framework
 {
+    /// <summary>
+    /// The backend options for OpenTK.
+    /// </summary>
+    public enum Backend
+    {
+        /// <summary>
+        /// Use the default backend for the current OS. If SDL2 is found, it will be used.
+        /// </summary>
+        Default,
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


tomspilman:
As much as I hate this we have no choice for at least 3.5.

Hopefully all this dies soon in 3.6.

mrhelmut:
:+1: Thanks!

I'm sharing the same thoughts.
I'm a bit busy lately, but as soon as I can I'll help with OpenTK replacement.

cra0zy:
> As much as I hate this we have no choice for at least 3.5.

I agree with that.

The new SDL2 branch has this solved :P

cra0zy:
Actually wait a sec, https://github.com/opentk/opentk/commit/e5ab1ae791ef07750971bc8a88e250c50cace268 fixed the SDL bug....

cra0zy:
PR is up https://github.com/MonoGame/MonoGame.Dependencies/pull/73

mrhelmut:
If we're updating OpenTK and mouse support, I would be interested in seeing opentk/opentk#354 reviewed... ^^

Hopefully this is the last time we have to deal with such changes.

cra0zy:
> If we're updating OpenTK and mouse support, I would be interested in seeing opentk/opentk#354 reviewed... ^^

I thought I already merged that... well its merged now.

> Hopefully this is the last time we have to deal with such changes.

I hope so too (｡-人-｡)

Anyway I also updated the PR to include your change.

mrhelmut:
Awesome!

I can confirm that MonoGame/MonoGame.Dependencies#73 fixes the SDL2 issue and that there is no need to have conditional code specific to SDL2 in this PR. (I still think that ```PlatformParameters``` is a good thing.)

tomspilman:
@KonajuGames - So do we merge this or not?

KonajuGames:
Getting the proper fix in for SDL2 would be good. It may push back 3.5
though because it needs the new Dependencies PR for OpenTK and then this
updated.

mrhelmut:
Linux forces the native backend, I think this PR is still interesting in regards to the ```PlatformsParameters``` (once the dependency is merged).

cra0zy:
@KonajuGames When updating this keep in mind that Linux needs native backend because SDL2 has this bug: https://bugzilla.libsdl.org/show_bug.cgi?id=3085 which was resolved in devel release.

KonajuGames:
PR updated.
- Uses latest OpenTK.
- Mouse coords outside window now work properly with SDL2 backend.
- Defaults to native backend on Linux, default backend on Windows and Mac.

cra0zy:
:+1: from me

tomspilman:
Thanks @KonajuGames !

mrhelmut:
Perfect, thanks.

PR #4633, opened 2016-03-14
Support 8-bit wav as source audio 

### LINKED ISSUE #4626: Sound effects are corrupt in Windows
When playing sound effects on Windows (10), the sound effect plays but it plays for too long.

KonajuGames:
The audio metadata put in the XNB files was incorrect.  It was based on the source audio file rather than the converted file.  This was most evident when a 8-bit wav was used as source data.

Fixes #4626.


tomspilman:
Thanks @KonajuGames !

Issue #4624, opened 2016-03-12
Save Button Not Updated on New Item


tomspilman:
In the Pipeline Tool on Windows if you add a new item via the "Add Item" menu the Save button on the toolbar doesn't enable to reflect the project being dirty and needing to be saved.

neofuji:
Also via "Add New Folder", "Rename", and  "Delete".

cra0zy:
This can be closed.

PR #4642, opened 2016-03-18
Enable the "Save" button after a new item is added to a project, so t…

dneelyep:
…hat the user knows that the project needs to be saved [due to the newly-added item].

This should fix Issue #4624.

One problem is that the Save button isn't disabled after it's clicked, but I think that was an already-existing issue. To explain further:
* Add -> New Item
* Fill out details
* Save button is now enabled
* Click Save button
* Save button is not disabled, where it should be at this point, I think

Anyhow, that's probably a problem for a separate Issue.


cra0zy:
You should just add a call to UpdateMenu function instead of enabling it that way.

dneelyep:
@cra0zy Ok cool, that's done now in commit b4b07f9.

tomspilman:
Thanks @dneelyep !

PR #4651, opened 2016-03-19
Fix SoundEffect.FromStream() and buffer copy

### LINKED ISSUE #4649: [OpenAL] Question about unnecessary(?) copy of sound data in RAM
SoundEffect.OpenAL.cs have field 
```C#
internal byte[] _data;
```
And as I can see, it is not used anywhere, it's just eats double size of sound in RAM.
I was trying to run one XNA game not ported to Linux with MonoGame, and with XNA this game eats only 200mb of ram, while with MonoGame on Linux it eats almost 3GB! I removed that _data field and get 1.5GB ram usage. With no stability issues. 
So i just wanted to ask, does that filed necessary or it can be safely removed to double reduce sound data ram usage?

mrhelmut:
Removed the internal ```_data``` from AL ```SoundEffect``` to avoid keeping it allocated, and fixed a bug in ```SoundEffect.FromStream()``` where the AL buffer would not get bound.

Fixes #4649

EDIT: working on the failing checks.

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
@@ -39,6 +39,8 @@ namespace Microsoft.Xna.Framework.Audio
 
         private void PlatformLoadAudioStream(Stream s)
         {
+            byte[] buffer;
+
 #if OPENAL && !(MONOMAC || IOS)
             
             ALFormat format;
@@ -60,7 +62,7 @@ namespace Microsoft.Xna.Framework.Audio
                     stream.Position = 0;
                 }
 #endif
-                byte[] buffer = AudioLoader.Load(stream, out format, out size, out freq);
+                buffer = AudioLoader.Load(stream, out format, out size, out freq);
 #if ANDROID
             }
             finally
--- a/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
@@ -48,29 +48,9 @@ namespace Microsoft.Xna.Framework.Audio
             int freq;
 
             var stream = s;
-#if ANDROID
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
```diff
@@ -87,8 +85,8 @@ private void PlatformLoadAudioStream(Stream s)
                 afs.ParseBytes (audiodata, false);
                 Size = (int)afs.DataByteCount;
 
-                _data = new byte[afs.DataByteCount];
-                Array.Copy (audiodata, afs.DataOffset, _data, 0, afs.DataByteCount);
+                buffer = new byte[afs.DataByteCount];
```
KonajuGames: Need to declare buffer before the using statement so it is visibleto the call to BindDataBuffer later.

KonajuGames:
I was just doing the same change.  You beat me to it.

You can also remove the Android-specific copy to a MemoryStream.  It was done back when AudioLoader used Seek, but it no longer does.

KonajuGames:
Thanks for this @mrhelmut. Merging.

MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
```diff
@@ -48,29 +48,9 @@ private void PlatformLoadAudioStream(Stream s)
...
-            try
-            {
-                // If seek is not supported (usually an indicator of a stream opened into the AssetManager), then copy
-                // into a temporary MemoryStream.
-                if (!s.CanSeek)
-                {
-                    needsDispose = true;
-                    stream = new MemoryStream();
-                    s.CopyTo(stream);
-                    stream.Position = 0;
```
dellis1972: @KonajuGames I'm sure we had this on Android for a reason. 
@mrhelmut did you test that this did not break android sound support?

PR #4653, opened 2016-03-19
Display project name in the window of the content pipeline tool.

dneelyep:
This fixes part of Issue #4457.

I'm not sure if exposing PipelineController._project.ProjectName publicly is the best way to get that information - I can change it to something else if desired. It looks like it was done a bit differently in the GTK pipeline tool, as pointed out by @cra0zy here: https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Gtk/MainWindow.cs#L53

### DIFF
--- a/Tools/Pipeline/Windows/MainView.cs
+++ b/Tools/Pipeline/Windows/MainView.cs
@@ -37,7 +37,14 @@ namespace MonoGame.Tools.Pipeline
         {
             get
             {
-                return "MonoGame Pipeline - " + ((PipelineController)_controller).ProjectName;
+                if (string.IsNullOrEmpty(((PipelineController)_controller).ProjectName))
+                {
+                    return "MonoGame Pipeline";
+                }
+                else
+                {
+                    return "MonoGame Pipeline - " + ((PipelineController)_controller).ProjectName;
+                }                
             }
         }
 


dneelyep:
Actually the only two commits relevant to this PR are 85e8712 and 94d672a. I need to go back and remove them from the PR.

cra0zy:
Try and keep 1 change per 1 PR.

I know I am not the one who should talk with #4459 around, but I learned my lesson.

tomspilman:
@dneelyep - It will be ok this one time.  In the future just avoid adding more than one fix to a PR unless they are tightly related.

@cra0zy - This look ok to you?

The only improvement would be sharing the title string creation code between Gtk and Winforms, but it doesn't seem critical right now.  Maybe later if we add a `*` in the title string to mark a dirty file like VS and other editors do.

Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -938,5 +938,13 @@ public string GetFullPath(string filePath)
 
             return _project.Location + Path.DirectorySeparatorChar + filePath;
         }
+
+        public string ProjectName
```
cra0zy: I don't think we should add a ProjectName param for no reason, there are already 2 other ways to get it:
1. Check root node
2. Path.GetFilenameWithoutExtension(_controller.ProjectLocation)

Tools/Pipeline/Windows/MainView.cs
```diff
@@ -33,6 +33,21 @@ partial class MainView : Form, IView, IProjectObserver
 
         public static MainView Form { get; private set; }
 
+        private string WindowTitle
```
cra0zy: This is something used only once, it should be moved into a method.

Tools/Pipeline/Windows/MainView.cs
```diff
@@ -85,7 +103,7 @@ public void Attach(IController controller)
             };
             _controller.OnBuildFinished += invokeUpdateMenus;
             _controller.OnProjectLoading += invokeUpdateMenus;
-            _controller.OnProjectLoaded += invokeUpdateMenus;
+            _controller.OnProjectLoaded += invokeProjectLoaded;
```
cra0zy: Add "UpdateTitle" inside "UpdateMenu", it may seem wasteful, but it's not, we need it there for later (updating project dirty in title aka. *)

cra0zy:
> The only improvement would be sharing the title string creation code between Gtk and Winforms, but it doesn't seem critical right now. Maybe later if we add a * in the title string to mark a dirty file like VS and other editors do.

You know, at one point I made a bunch of code sharable between WinForms and Gtk (#3980), but due to the fact that no one was interested in it and I was just resolving merge conflicts I close it.

tomspilman:
@dneelyep - Can you address and resolve the code review notes that @cra0zy posted?

tomspilman:
@dneelyep - Can you address and resolve the code review notes that @cra0zy posted?

cra0zy:
We've moved Pipeline Tool to Eto Forms and among other things this gas been fixed.

Issue #4654, opened 2016-03-20
[Windows][3.5] Accessing WinFormsGameWindow.ClientBounds via cross-thread operation crashes


nilllzz:
Just upgraded to MonoGame 3.5 and when accessing the ClientBounds property of the WinFormsGameWindow instance "Window" of the Game class from a different thread from the main thread, it crashes with an **InvalidOperationException**.

![error](https://cloud.githubusercontent.com/assets/2119926/13902247/17b5e7c4-ee42-11e5-9e32-2d779eb554c9.PNG)

I tested this with version 3.4, and this behavior is not showing up there.

tomspilman:
The only change there is this one:

https://github.com/mono/MonoGame/commit/e74c7e57227f216a0cba7b50852c72c8bd91e520

I guess the `_form.PointToScreen()` is the culprit, but we would need to debug it to know.

Hard to say if this should be thread safe or not, but if it can be easily made to be thread safe lets do that.

nilllzz:
I tested it, `PointToScreen `seems fine, `ClientSize `appears to be the problem.
But the size is also contained in `_form.ClientRectangle`, so merging those two methods should yield the correct result while still working thread safe.

I'll set up a pull request.

nilllzz:
Hm actually, this might be a little more complex.
Turns out, this only happens in Debug mode in Visual Studio for some reason.
Running the same exe with the same libraries from the bin folder does not produce a crash.

theZMan:
That is expected behavior I believe - the debugger watches for the illegal cross thread operations no debugger = no exception.

What its telling you is that you are performing something that might, one day, on one machine turn out to cause a crash. Likely a crash you will never debug as its on one of your customers machines and will be only useful to earn you a slightly worse steam review ;-)

You need to do UI operations on the main thread see http://stackoverflow.com/questions/1523878/getting-cross-thread-operation-not-valid and many other posts. Seems like something in there counts as a winforms UI operation.

nilllzz:
Yeah, that is the cause of the exception, and the source code that causes this (the `PointToScreen` and `ClientSize` methods) have been identified.

This is changed behavior from MonoGame 3.4 though, and the question is if an exception is intended when accessing `Window.ClientBounds` via a non-GUI thread.
Prior to 3.5, the property did not throw an exception because neither `PointToScreen`, nor `ClientSize` have been used.

cra0zy:
[...]

PR #4655, opened 2016-03-20
Fixed Window.ClientBounds crashing on secondary thread

nilllzz:
See issue #4654.
Fixes commit e74c7e5.

Cross thread operations accessing the Window.ClientBounds property caused an InvalidOperationException.


nilllzz:
Nevermind, see #4654.
The issue seems to be a little more complex.

Issue #4605, opened 2016-03-05
Make GameWindow a partial class instead of abstract class


cra0zy:
Long story short:

1) We could add perplatform stuff more easily without having to resort to:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/GameWindow.cs#L26
2) Why is there an AllowUserResizing property on Android? Ok, we are keeping that for XNA compatibility, but still, does make a point, we could potentially disable stuff that doesn't work on specific platforms.

CC. @dellis1972 @KonajuGames @tomspilman

cra0zy:
Closing this since it would make PCL a bit awkward, there is a better option.

PR #4662, opened 2016-03-20
Added LocationChanged API for the GameWindow.

nilllzz:
Added OnLocationChanged event for Windows DX and OpenTK platform, which
reacts to changes in the location of the game window.

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
@@ -251,6 +251,11 @@ namespace Microsoft.Xna.Framework
             }
         }
 
+        private void OnMove(object sender, EventArgs e)
+        {
+            OnLocationChanged();
+        }
+
         internal void ProcessEvents()
         {
             lock (window)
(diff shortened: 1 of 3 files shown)


tomspilman:
I don't know about adding this.  It only is relevant for the 3 desktop platforms and I see little use for it in games.  Unless there is some critical need for window move events I think this should be closed.

@KonajuGames @dellis1972 @cra0zy ?

cra0zy:
> I don't know about adding this. It only is relevant for the 3 desktop platforms and I see little use for it in games. Unless there is some critical need for window move events I think this should be closed.

Agreed. Also #4605

KonajuGames:
Agreed.  We need to keep focus on things that affect a game, and the
changing of the window position does not affect the game.  There is no
issue with you doing this in your own fork of MonoGame, but we don't
believe it belongs in the core framework.

cra0zy:
> There is no
> issue with you doing this in your own fork of MonoGame, but we don't
> believe it belongs in the core framework.

He doesn't really need to even fork it, he just needs to check window position on "Update" method and fire up an event when it gets moved...

PR #4663, opened 2016-03-20
[Content Pipeline] Improve model importer

cra0zy:
Does 2 things:
 1. Adds a proper list of importing types (http://assimp.sourceforge.net/main_features_formats.html)
 2. ~~Removes ImporterName since ContentIdentity can resolve from filename only~~

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
@@ -14,7 +14,45 @@ using Microsoft.Xna.Framework.Content.Pipeline.Graphics;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline
 {
-    [ContentImporter(".dae", ".fbx", ".x", DisplayName = "Open Asset Import Library - MonoGame", DefaultProcessor = "ModelProcessor")]
+    [ContentImporter(
+        ".dae", // Collada
+        ".gltf", "glb", // glTF
+        ".blend", // Blender 3D
+        ".3ds", // 3ds Max 3DS
+        ".ase", // 3ds Max ASE
+        ".obj", // Wavefront Object
+        ".ifc", // Industry Foundation Classes (IFC/Step)
+        ".xgl", ".zgl", // XGL
+        ".ply", // Stanford Polygon Library
+        ".dxf", // AutoCAD DXF
+        ".lwo", // LightWave
+        ".lws", // LightWave Scene
+        ".lxo", // Modo
+        ".stl", // Stereolithography
+        ".ac", // AC3D
+        ".ms3d", // Milkshape 3D
+        ".cob", ".scn", // TrueSpace
+        ".bvh", // Biovision BVH
+        ".csm", // CharacterStudio Motion
+        ".irrmesh", // Irrlicht Mesh
+        ".irr", // Irrlicht Scene
+        ".mdl", // Quake I, 3D GameStudio (3DGS)
+        ".md2", // Quake II
+        ".md3", // Quake III Mesh
+        ".pk3", // Quake III Map/BSP
+        ".mdc", // Return to Castle Wolfenstein
+        ".md5", // Doom 3
+        ".smd", ".vta", // Valve Model 
+        ".ogex", // Open Game Engine Exchange
+        ".3d", // Unreal
+        ".b3d", // BlitzBasic 3D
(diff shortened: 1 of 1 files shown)


mgarstenauer:
> ... since ContentIdentity can resolve from filename only

Could you please explain this. I don't understand what you mean.

MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
```diff
@@ -198,7 +238,7 @@ public override NodeContent Import(string filename, ContentImporterContext conte
 			catch { }
 #endif
 
-            _identity = new ContentIdentity(filename, string.IsNullOrEmpty(ImporterName) ? GetType().Name : ImporterName);
+            _identity = new ContentIdentity(filename);
```
cra0zy: @mgarstenauer I mean this part of the code.

mgarstenauer: @cra0zy Yes, but what why remove it? For debugging it can be helpful to know which importer ("FbxImporter", "CustomFbxImporter", etc.) created the content.

cra0zy: Oh, that was it's use, I'll restore it as soon as I get back home.

cra0zy:
@mgarstenauer Updated, good now?

mgarstenauer:
@cra0zy Yes, I am fine with that.

tomspilman:
@cra0zy - I think we should leave .fbx and .x off that list.  Just so that it defaults to the `FbxImporter` and `XImporter` by default for those file types.  Also removes some confusion on which one to use.

cra0zy:
> @cra0zy - I think we should leave .fbx and .x off that list. Just so that it defaults to the FbxImporter and XImporter by default for those file types. Also removes some confusion on which one to use.

Aye aye captain... and done.

KonajuGames:
Some of the file extensions, such as .raw, are also listed by the texture importer.  In the case of .raw, it is far more commonly used for images, but are we going overboard with the listing of extensions.  The importer libraries we use support them, but should we list every extension?

cra0zy:
> Some of the file extensions, such as .raw, are also listed by the texture importer. In the case of .raw, it is far more commonly used for images, but are we going overboard with the listing of extensions. The importer libraries we use support them, but should we list every extension?

1. We are doing same thing with Texture: https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/TextureImporter.cs#L17
2. There is no negative side to adding it

Want me to remove .raw?

tomspilman:
> should we list every extension?

 If we don't list the extensions of formats we support I don't know how anyone would ever guess they are supported.  The file extension on the importer is the primary way of finding that out as it lets the Pipeline tool automatically associate the right importer to the type.

However.

If we have the same extension for two different types of files...  that can be trouble.

> Want me to remove .raw?

Yeah you should remove it...  most likely if someone is opening a .raw file it is a texture and not a model.

cra0zy:
> Yeah you should remove it... most likely if someone is opening a .raw file it is a texture and not a model.

Done.

cra0zy:
@tomspilman Something is not right with the build server. http://teamcity.monogame.net/viewLog.html?tab=buildLog&buildTypeId=MonoGame_TestWindows&buildId=7214#_focus=72

tomspilman:
@cra0zy - N

MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
```diff
@@ -14,7 +14,46 @@
...
+        ".lwo", // LightWave
+        ".lws", // LightWave Scene
+        ".lxo", // Modo
+        ".stl", // Stereolithography
+        ".ac", // AC3D
+        ".ms3d", // Milkshape 3D
+        ".cob", ".scn", // TrueSpace
+        ".bvh", // Biovision BVH
+        ".csm", // CharacterStudio Motion
+        ".xml", // Ogre XML
```
tomspilman: This is another that we may want to remove...  else every time you add an XML file to your content project it will automatically decide it is a model file.

cra0zy: Done.

tomspilman:
Thanks @cra0zy !  Merging.

PR #4664, opened 2016-03-21
[Content Pipeline] Support specialized LoopLength and LoopStart

neofuji:


tomspilman:
I don't see a problem with providing a way to override the default loop information from the sound file.  @KonajuGames - Have any thoughts on this?

However the value in "samples" might not be intuitive for someone editing this from the Pipeline tool.  Maybe the overload values should be a percentage of the sound length... which we internally convert to samples on import?

Also how can you force the `loopStart` to zero if it must be greater than zero to overload the sound value.  Maybe the defaults should be -1 instead?

Also we need to clearly document that these values overload the sound file values.

KonajuGames:
This is a tricky one. I'm not decided at this stage. Ideally the loop
points will be authored into the file using a sound editing tool, but I can
also see the benefits of quickly iterating changes (but then this would be
better done in a tool designed for the purpose where iterative testing
would be quicker still).

As for the units, I don't think percentage is a good choice because quite
often the loop points need to be exact (to the sample) to avoid clicks and
pops.  Sound authors will provide the loop points in samples or, less
commonly, time.

neofuji:
Is `System.Nullable<int>` useful for providing default values?

tomspilman:
@neofuji - It is...  but I don't know how the Pipeline tool will react.  It might appear in the property window ok, it might crash, but it likely won't properly allow someone to clear the value to null again.

@cra0zy ?

cra0zy:
@tomspilman You got it right.

tomspilman:
It is better for this to be done in an audio tool.  Closing this.

Issue #4666, opened 2016-03-22
Playing Songs stop before finishing in Windows, WindowsGL, and DesktopGL


neofuji:
When the streaming finishes, Songs will invoke the `OnSongFinishedPlaying` method.
However, the samples are still in a sound buffer, so they will not be played.

mrhelmut:
Confirmed on DesktopGL.

This is because ```OggStreamer``` invokes ```OggStream.FinishedAction``` when the last buffer is filled & queued, not when it has finished playing. This represents about 500ms of sound (depending on how fast the decoder thread works).
I'll try to take a look at this, but it's not as simple as checking the last buffer state (because there are many reasons which make the state not so reliable to detect the end of a ```Song```).

I believe other implementations are not affected, could someone confirm?

neofuji:
I tried fixing it on Windows, and it has gone well.

PR #4668, opened 2016-03-22
Fix Song (#4666) in Windows

neofuji:
#4666

### DIFF
--- a/MonoGame.Framework/Media/MediaPlayer.WMS.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.WMS.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Xna.Framework.Media
 
                 switch (ev.TypeInfo)
                 {
-                    case MediaEventTypes.EndOfPresentation:
+                    case MediaEventTypes.SessionEnded:
                         _sessionState = SessionState.Ended;
                         OnSongFinishedPlaying(null, null);
                         break;


tomspilman:
Anyone have time to test this change?  It only affects the Windows desktop DX platform.

mrhelmut:
According to the Media Foundation documentation (https://msdn.microsoft.com/en-us/library/windows/desktop/ms704703.aspx), ```MediaEventTypes.SessionEnded``` happens when the last item queued has finished playing, not after each item ends. ```OnSongFinishedPlaying``` wouldn't be triggered by ```Song``` changes if multiple ones are queued through ```MediaPlayer.Play(SongCollection collection)```.
I can't test it, though.

@neofuji Would you mind double-checking that #4666 happens on Windows (DirectX) projects and not only on DesktopGL? I would assume that Media Foundation deals with this correctly.

neofuji:
@mrhelmut #4666 certainly happens on Windows.
The `Song` queue depends on `MediaPlayer`, not on Media Foundation.
So queued Songs are passed to Media Foundation one by one.

mrhelmut:
So I'm fine with this. :+1:

KonajuGames:
Merging this.  Thanks @neofuji.

KonajuGames:
@mrhelmut, the OggStreamer error you raised in #4666 will need a separate PR if you get around to fixing it.

PR #4673, opened 2016-03-24
Allow disabling Content Pipeline tests

polsys:
First off, I'm not sure if this issue actually affects more people, but I've found myself doing these changes repeatedly. Feel free to close if this seems not worth the trouble.

I don't have the full XNA framework on my dev computer, since I'm not too keen on installing an old VS version just to get the bits. The unit tests run nicely against the XNA redistributable, save for those involving the Content Pipeline, which is only included in the full framework. The missing types cause a build failure. This PR adds support for using `NO_CONTENTPIPELINE` symbol to disable all Content Pipeline tests/utilities and get the test project building.

Also added missing license headers to the files I touched. Visual Studio changed indentation on a couple of lines, sorry for that.

### DIFF
--- a/Test/ContentPipeline/AssetTestClasses.cs
+++ b/Test/ContentPipeline/AssetTestClasses.cs
@@ -2,6 +2,8 @@
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
+#if !NO_CONTENTPIPELINE
+
 using System;
 using System.Collections;
 using Microsoft.Xna.Framework;
@@ -536,7 +538,7 @@ class StructArrayNoElements
 
 namespace MonoGame.Tests.ContentPipeline
 {
-    #region Namespaces
+#region Namespaces
     public class NamespaceHelper
     {
         public bool Value;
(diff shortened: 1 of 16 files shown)


tomspilman:
> save for those involving the Content Pipeline, which is only included in the full framework. 

You know the simple approach here would be including the XNA pipeline assemblies in our dependencies and linking directly to them instead of assuming they are installed on the system.

I know technically they shouldn't be redistributed, but i'm pretty sure MS won't mind us doing it.

@KonajuGames @dellis1972 ?

tomspilman:
@polsys - Speaking with @KonajuGames I think we're going to just include the XNA assemblies into the repo so that you don't need it installed to be able to run the XNA unit tests anymore.

Going to leave this open until I get time to submit a PR which solves this.  Give me a day or two.

polsys:
Alright, that's even better from testing point of view, if it's technically and legally feasible :+1:

polsys:
Another way to solve this would be linking to the [instructions](https://blogs.msdn.microsoft.com/astebner/2009/06/12/steps-to-install-xna-game-studio-without-using-the-setup-bootstrapper/) mentioned in #3550 somewhere in the docs. I wasn't aware of that before, haven't had time to test the partial install yet though.

polsys:
Closing this since this has become stale and there should be better ways to fix the issue.

PR #4675, opened 2016-03-25
Implement FrameworkDispatcher

### LINKED ISSUE #1494: XAudio2 might be started on a worker thread
XAudio2.StartEngine() must be called on the main thread.  The static constructor for SoundEffect is called on the first use of the SoundEffect class.  If this first use is on a worker thread, the static constructor is run in the context of the worker thread and XAudio2.StartEngine() will be called on the worker thread.

I have a pending solution for this (invoking the StartEngine call back to the main thread), but it's not properly working yet.  Raising this issue so it's not forgotten.

polsys:
Fixes #1494. Implemented `FrameworkDispatcher` and hooked the sound system into it. As a result, XAudio is always initialized on game main thread and sound functions are available also without using `Game`.

Some issues:
- The OpenAL changes need to be replicated in the SDL branch.
- Thread safety? Unit test indicates that XNA allows calling `Update()` from a separate thread, but I haven't figured out yet if it does some dispatching to the UI thread. [This MSDN page](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.frameworkdispatcher.update.aspx) talks about using a `DispatcherTimer` to call `Update()`, so I've assumed that the responsibility is on the caller.
- Only tested on Windows desktop, and didn't have any game at hand to test this with.


mrhelmut:
I'll test the OpenAL part, but that looks good to me.

MonoGame.Framework/FrameworkDispatcher.cs
```diff
@@ -2,15 +2,60 @@
...
+            DoUpdate();
+        }
+
+        private static void DoUpdate()
+        {
+            // Update sound system
+#if OPENGL
+            OpenALSoundController.GetInstance.Update();
+#endif
+            SoundEffectInstancePool.Update();
```
polsys: `OpenALSoundController.GetInstance.Update()` is also called in `SoundEffectInstancePool.Update()`. This duplicated behavior actually matches the original, since it was called in `OpenTKGamePlatform.BeforeUpdate()` and after that the pool was updated. Is there any reason to preserve this behavior, or can I remove the other call (which would probably be the one in the instance pool)?

mrhelmut: I need to check this, I remember that one of the two is more important in cleaning playing instances upon ```ContentManager.Unload()```, but I believe that removing it from ```SoundEffectInstancePool.Update()``` and calling it in ```FrameworkDispatcher``` before the pool would be ok.

mrhelmut: Just letting you know that I've removed ```OpenALSoundController.Update()``` in #4696 (if it gets merged).

polsys: That seems great, I'll update this when it gets merged. Thanks for the heads-up!

KonajuGames: #4696 has been merged now.

MonoGame.Framework/Audio/SoundEffect.XAudio.cs
```diff
@@ -76,9 +76,12 @@ internal static X3DAudio Device3D
 
         static SoundEffect()
         {
-            InitializeSoundEffect();
```
KonajuGames: If the static constructor is now empty, remove it.

polsys:
Oops, messed up the merge.

polsys:
Removed the `Update()` call and the static constructor,

I had some trouble running the tests since I'm on the go and the SDL change has moved lots of things around. My Git incompetence may have its part too...

MonoGame.Framework/Game.cs
```diff
@@ -647,13 +650,9 @@ internal void DoUpdate(GameTime gameTime)
             AssertNotDisposed();
             if (Platform.BeforeUpdate(gameTime))
             {
-                // Once per frame, we need to check currently 
-                // playing sounds to see if they've stopped,
-                // and return them back to the pool if so.
-                SoundEffectInstancePool.Update();
-
+                FrameworkDispatcher.Update();
                 DynamicSoundEffectInstanceManager.UpdatePlayingInstances();
```
tomspilman: This should go now too right?

polsys:
Fixed conflicts and moved the `DynamicSoundEffectInstance` update call to `FrameworkDispatcher`.

MonoGame.Framework/Game.cs
polsys: Yup, already pushed the fix. I've still to wake up apparently...

MonoGame.Framework/FrameworkDispatcher.cs
```diff
@@ -2,15 +2,57 @@
...
+#elif OPENGL
+            try
+            {
+                var oal = OpenALSoundController.GetInstance;
+            }
+            catch (DllNotFoundException ex)
+            {
+                throw new NoAudioHardwareException("Failed to init OpenALSoundController", ex);
+            }
+#endif
```
tomspilman: We should avoid the `#if`s here as it makes it hard to extend things to the console platforms.

Still why is this here?  Usually if one were implementing some code that didn't use `Game` yet used `SoundEffect`...  it would go something like this:

```
var sfx = SoundEffect.FromStream(stream);
sfx.Play();
FrameworkDispatcher.Update();
```

Essentially initializing things a second time no?

polsys: Now that you mention it, this is indeed a breaking change. Previously, the audio device was created in `SoundEffect` static constructor (which could be called on any thread, see #1494). Now, a call to `FrameworkDispatcher.Update()` is required. On the other hand, just about all MSDN docs for XNA say things like

> Play throws an InvalidOperationException if FrameworkDispatcher.Update has not been called at least once before making this call.

The line applies to Silverlight apps, but it was just as true when I was working on `DynamicSoundEffectInstance` tests.

I guess there could be some logic in sound effect initialization to allow both cases, if we want to preserve the old behavior.

tomspilman: > On the other hand, just about all MSDN docs for XNA say things like

But is that true only under Silverlight or does it also apply to sounds on regular XNA?

IMO we should obey whatever rules are set for a regular XNA game and not special cases for Silverlight.

> if we want to preserve the old behavior.

 I don't think we need to preserve old behavior here.  People that care will appreciate XNA compatibility over everything else here.

KonajuGames: It also applies to XNA when there is no Game object, such as when XNA is
being used to render in WinForms or similar.

tomspilman: > It also applies to XNA when there is no Game object

That is my question really... does it?  Not speculation but we need a test to verify this is true.

Could be a unit test that verifies that the right calls throw the right exceptions if `FrameworkDispatcher.Update` isn't called first.

polsys: It's a bit difficult to test since the initialization is done once per process. We would need a way to either run some tests first or to be able reinitialize things reliably. That's why I skipped it while writing the `FrameworkDispatcher` tests.

MonoGame.Framework/FrameworkDispatcher.cs
```diff
@@ -2,15 +2,57 @@
...
+        {
+            DynamicSoundEffectInstanceManager.UpdatePlayingInstances();
+            SoundEffectInstancePool.Update();
+        }
+
+        private static void Initialize()
+        {
+            // Initialize sound system
+#if DIRECTX
+            SoundEffect.InitializeSoundEffect();
```
tomspilman: So the only issue that we need to fix here is this `#if` block.  If we commit as is then we break the console repos.  We need to add a platform method into the `SoundEffect` partial class like:  `SoundEffect.PlatformInitializeSystem()` which can then be separately implemented for XAudio and OpenAL platforms.

This way consoles can correctly overload this behavior for their respective audio systems.

If you don't have the time to do this @polsys let me know.

polsys:
So while I was away GitHub somehow flagged me as a bot, hiding this PR (without notifying me)...

I should have time to fix the `#if` issue in a day or two.

tomspilman:
Thanks @polsys !

MonoGame.Framework/FrameworkDispatcher.cs
polsys: Fixed.

Jjagg:
Can this be merged @tomspilman? I'm doing something that depends on this.

KonajuGames:
Merging this.

PR #4677, opened 2016-03-26
Fix Gamepad crash when platform doesn't support the amount

cra0zy:
Instead of throwing an exception we should just return an empty state, this also makes the style of this more XNAish.

http://community.monogame.net/t/monogame-3-5-gamepad-getstate/7288

### DIFF
--- a/MonoGame.Framework/Input/GamePad.cs
+++ b/MonoGame.Framework/Input/GamePad.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Xna.Framework.Input
         public static GamePadCapabilities GetCapabilities(int index)
         {
             if (index < 0 || index >= PlatformGetMaxNumberOfGamePads())
-                throw new InvalidOperationException();
+                return new GamePadCapabilities();
 
             return PlatformGetCapabilities(index);
         }
@@ -76,7 +76,7 @@ namespace Microsoft.Xna.Framework.Input
         public static GamePadState GetState(int index, GamePadDeadZone deadZoneMode)
         {
             if (index < 0 || index >= PlatformGetMaxNumberOfGamePads())
-                throw new InvalidOperationException();
+                return GamePadState.Default;
             
             return PlatformGetState(index, deadZoneMode);
         }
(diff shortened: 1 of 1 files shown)


tomspilman:
Hopefully no one depends on the exception to detect the maximum supported gamepad.  At least we added `GamePad.MaximumGamePadCount` for that.

cra0zy:
> Hopefully no one depends on the exception to detect the maximum supported gamepad. At least we added GamePad.MaximumGamePadCount for that.

Xna didn't call a crash when the gamepad was not supported(Windows Phone), but instead just returned an empty state, right?

tomspilman:
> but instead just returned an empty state, right?

Well...  it did return a gamepad state, but just for the back button detection.

Don't know what happened if you asked for the second gamepad.

cra0zy:
> Don't know what happened if you asked for the second gamepad.

From googling around I am 95% sure it returned an empty state.

cra0zy:
@dellis1972 @KonajuGames What do you think on this?

tomspilman:
So I did a little research on this before merging it.

Under XNA a disconnected gamepad does not throw any exception.  It just returns a disconnected state.

If you are using XNA on WP7 `GamePad.GetState(PlayerIndex.Two)` also does not throw an exception.  It just reports it as disconnected.  In fact on WP7 you can call `GamePad.GetState((PlayerIndex)100)` and it just returns a disconnected state.

The only case where `GamePad.GetState()` throws an exception is on the PC when you pass an invalid enum value.  So `GamePad.GetState((PlayerIndex)100)` on the PC throws and exception about an invalid enum value.

So my conclusions from this:

This PR is fine... we shouldn't throw exceptions from the indexed versions of the `GamePad.GetState/GetCapabilities/SetVibration` when referencing more gamepads than supported on the platform.  I will merge this shortly.

However we should consider following up this PR with another for the enum versions of these functions which validates that a correct `PlayerIndex` enum is passed it and throw an argument exception otherwise.  The enum versions should be strict and behave like XNA on the desktop.

tomspilman:
Thanks @cra0zy !

Issue #4678, opened 2016-03-27
Open up Shader creation without exposing APIs


hach-que:
I'm aware that the internals of Shader and the Effect classes are likely to change in the near future.  One thing I'd like to see is the ability for games (and in particular, engines written on top of MonoGame) to be able to initialize the Shader class and Effect class without requiring them to write out shaders in the MGFX format.

In particular, I want to be able to support geometry shaders before the file format is updated to support it (I really don't like touching the file format because when it gets changed everyone has to recompile their assets).

My proposal here is to replace the BinaryReader on Shader and Effect, with an interface like `IShaderReader` and `IEffectReader`.  `IShaderReader` would look like:

``` csharp
namespace Microsoft.Xna.Framework.Graphics
{
    public interface IShaderReader
    {
        ShaderStage GetStage();

        byte[] GetBytecode();

        int GetSamplerCount();

        SamplerType GetSamplerType(int samplerIndex);

        int GetSamplerTextureSlot(int samplerIndex);

        int GetSamplerSamplerSlot(int samplerIndex);

        SamplerState GetSamplerState(int samplerIndex);

        string GetSamplerName(int samplerIndex);

        int GetSamplerParameter(int samplerIndex);

        int GetConstantBufferCount();

        int GetConstantBufferValue(int constantBufferIndex);
    }
}
```

Internally MonoGame can use a BinaryReader-based version of this for MGFX files, while other engines can load shaders and effects through their own systems.  This also raises the possibility of having backwards compatibility in the MGFX format because you can swap out the implementation of the interface based on the version number in the MGFX file.

hach-que:
I should add that this makes it extremely easy to add a new reader for GLSL shaders... you just need a way of specifying the metadata if this API change is made.

PR #4679, opened 2016-03-27
Separate out effect and shader reading from effect / shader classes

hach-que:
This addresses #4678 by allowing effects to be created without relying on the MGFX binary file format.


hach-que:
This is ready for review and merge (for 3.6 release).

hach-que:
Resubmitting it from the hach-que repository.

PR #4683, opened 2016-03-28
Fix #4680 and add tests

### LINKED ISSUE #4680: Loading PNG Images
Hi,

Yesterday I have updated MonoGame and my project to the latest version of MonoGame (3.5).
For some reason, all png images are loaded with different colors as shown at the image below (reddish color).
https://gyazo.com/cec4a7a9b12782f66f726876ab085d70

Also, this bug doesnt seem to occur if I loaded the PNG images differently: "Texture2D.FromStream();"
seems to work.. (Content.Load<>(""); does not).

Would appreciate if you guys could help me asap :)

Tzenchor:
After I had all ready I saw that there is another test class called `PackedVectorTest` that does a superficial test of `RGBA64`. In the end I opted to keep the new test class apart from that as the new tests are specific of this type.

Changelog:
Fixes #4680.
RGBA64:
- Inverted the order of the colors, from ABGR to RGBA as with other
pixel formats.

- Fix overflow problems of Pack method. The shift operator was being
applyed to ints.

TextureImporter:
- Remove the red and blue channel swapping for images with format
RGBA64.

RGBA64Test:
- Test the constructor of the class and the memory copy method used in
TextureImporter so the returned values are consistent


tomspilman:
This is great, but I would prefer if we added new tests to the existing `PackedVectorTests.Rgba64()`.  This way the weakness of the other tests is more apparent and can be addressed in a later PR.

> superficial test of `RGBA64`

More specifically what you have shown is that we're not testing for correct packing order.  We need something as simple as:

```
            // Test packing order.
            Assert.AreEqual(0xC7AD8F5C570A1EB8, new Rgba64(0.12f, 0.34f, 0.56f, 0.78f).PackedValue);
```

Which passes on XNA and fails in MonoGame currently... but should pass with your fixes here.

Not sure about the memcopy test you have... not sure what it proves.

tomspilman:
It would also be good to add the failing image to our tests in some way.  It could be as simple as processing the texture and testing a few pixels for expected values.  This seem like a good idea @KonajuGames ?

Tzenchor:
I'll address the changes later today.
There's a test with an image with this format, so i'll upgrade the test, specially now that I know where the useful data begins

Tzenchor:
Done,
The memory copy test was testing the results of what was internally done in `PixelBitmapContent.SetData()`. But after integrating it into `TextureImporter.ImportRGBA16Png()` it checks the same.

About the test case you provided, it keeps failing, probably because of some rounding differences between .Net 4.0 and 4.5 (the difference between both values is of 0x100010000). So I had to adapt the values to remove the imprecision (now the test passes on MonoGame and XNA)

tomspilman:
> probably because of some rounding differences between .Net 4.0 and 4.5 

I suspect it is how the compiler decides to store the constants under different version of VS.  

Your changes there seem to be great and are very easy to follow.  I added an issue for the remaining fix for the other tests in https://github.com/mono/MonoGame/issues/4685.

tomspilman:
Thanks @Tzenchor !

PR #4696, opened 2016-04-01
Fixing and cleaning AL source management

### LINKED ISSUE #4660: [OpenAL] Question about redunant variable and some others.
Audio/OggStream.cs
Class `OggStreamer`
Method `FillBuffer()`
```C#
readerPosition = stream.Reader.DecodedPosition;
```
Is `readerPosition` variable need for something?
Or it was used or planned to be used with something like VorbisReader pool? Like a `readMutex` also.
I mean, for using many OggStream with one VorbisReader by seeking it.
For now I'm trying to implement this and plan to make SoundEffect use not only raw PCM data, but also Vorbis data, to reduce RAM usage.
`Song` class can be played only one instance in time, but I need many songs(or simply long sounds) played in one time. And this eats a lot of RAM with using raw PCM.
What do you think?

### LINKED ISSUE #4672: [DesktopGL] Heavy Song volume fluctuation
Hello,

First, I should note that I don't know if the problem comes from the ```OggStreamer``` / NVorbis, OpenAL, or my own code.

When playing ```Song``` for an extended time (from 5 to 10 minutes, and it doesn't matter if it is the same song or a queue of different ones), it happens very often that the ```Song``` playing volume changes by itself without any call to ```MediaPlayer``` in the last couple minutes, and it can go from plain 0 to 1.

This happens on all platforms, with OpenAL-Soft or Apple/Creative's one, and only affects ```Song```, not ```SoundEffect```.

Our game can be very ```SoundEffect``` intensive, and ```Song``` are cross-faded by code. At first I thought that I was putting to much pressure on the OpenAL command buffer, but even adding locks to have a max of 10 AL calls per second doesn't change anything, and the problem can happen in moments when absolutely no ```SoundEffect``` is playing for several seconds.

Any thoughts?

### LINKED ISSUE #4686: SoundEffects randomly disappear on Android
I am using Monogame 3.5 with Xamarin 5.10.3 and noticed that sounds randomly dissapear on Android. iOS version of game has no problems. 

No related debug messages in output. 

SoundEffect.Play() or SoundEffectInstance.Play() always return true.

mrhelmut:
Fixes #4672 
Fixes #4660

This PR fixes an issue where a ```SoundEffectInstance``` could share its AL source with another ```SoundEffectInstance``` or even ```Song```. It was happening because ```OpenALSoundController``` was not thread-safe and because AL sources were not recycled soon enough by the pool once an instance stopped playing (and sometime not recycled at all).

It also simplifies ```OpenALSoundController``` (removed ```Update()``` and ```PlaySound()```) which was doing pretty much the same work as ```SoundEffectInstancePool.Update()```.

@iodiot Could you give it a test to see if it also fixes #4686 ?

@cra0zy @dellis1972 @KonajuGames Could you test it on your respective systems? I could only test on Windows & OS X.

### DIFF
--- a/MonoGame.Framework/Android/AndroidGamePlatform.cs
+++ b/MonoGame.Framework/Android/AndroidGamePlatform.cs
@@ -11,8 +11,6 @@ namespace Microsoft.Xna.Framework
 {
     class AndroidGamePlatform : GamePlatform
     {
-        OpenALSoundController soundControllerInstance = null;
-
         public AndroidGamePlatform(Game game)
             : base(game)
         {
@@ -27,7 +25,7 @@ namespace Microsoft.Xna.Framework
             MediaLibrary.Context = Game.Activity;
             try
             {
-                soundControllerInstance = OpenALSoundController.GetInstance;
+                OpenALSoundController soundControllerInstance = OpenALSoundController.GetInstance;
             }
             catch (DllNotFoundException ex)
             {
--- a/MonoGame.Framework/Audio/OALSoundBuffer.cs
+++ b/MonoGame.Framework/Audio/OALSoundBuffer.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Xna.Framework.Audio
                 alError = AL.GetError();
                 if (alError != ALError.NoError)
                 {
-                    Console.WriteLine("Failed to get buffer bits: {0}, format={1}, size={2}, sampleRate={3}", AL.GetErrorString(alError), format, size, sampleRate);
+                    Console.WriteLine("Failed to get buffer channels: {0}, format={1}, size={2}, sampleRate={3}", AL.GetErrorString(alError), format, size, sampleRate);
                     Duration = -1;
                 }
                 else
--- a/MonoGame.Framework/Audio/OggStream.cs
+++ b/MonoGame.Framework/Audio/OggStream.cs
@@ -405,10 +405,8 @@ namespace Microsoft.Xna.Framework.Audio
         public bool FillBuffer(OggStream stream, int bufferId)
         {
             int readSamples;
-            long readerPosition = 0;
             lock (readMutex)
             {
-                readerPosition = stream.Reader.DecodedPosition;
                 readSamples = stream.Reader.ReadSamples(readSampleBuffer, 0, BufferSize);
                 CastBuffer(readSampleBuffer, castBuffer, readSamples);
             }
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -88,8 +88,6 @@ namespace Microsoft.Xna.Framework.Audio
 #endif
         private List<int> availableSourcesCollection;
         private List<int> inUseSourcesCollection;
-        private List<int> playingSourcesCollection;
-        private List<int> purgeMe;
         private bool _bSoundAvailable = false;
         private Exception _SoundInitException; // Here to bubble back up to the developer
         bool _isDisposed;
@@ -112,8 +110,6 @@ namespace Microsoft.Xna.Framework.Audio
 
             availableSourcesCollection = new List<int>(allSourcesArray);
 			inUseSourcesCollection = new List<int>();
-			playingSourcesCollection = new List<int>();
-            purgeMe = new List<int>();
 		}
 
         ~OpenALSoundController()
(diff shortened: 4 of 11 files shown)


iodiot:
@mrhelmut Nope, sounds disappear with the same rate.

mrhelmut:
Thanks! Your issue is something different then.

tomspilman:
Has anyone else had time to test out these changes?

KonajuGames:
I haven't tested them yet.  This came in during my down time so I missed
it.  I'll get onto it tonight.

cra0zy:
Tested this, works fine for me. Sorry for the delay, I was quite busy.

mrhelmut:
@KonajuGames @dellis1972 Did you had some time for reviewing?

KonajuGames:
I will have a better look at this one tonight.

KonajuGames:
I'm going to merge this.  If there are still some issues, they will only show up under wider usage.

polsys:
I ran into one issue while running the tests for `FrameworkDispatcher` work. The test goes like this:
```cs
var sfx = new SoundEffect(new byte[] { 0, 0 }, 44100, AudioChannels.Mono);
sfx.Play();
// Assertions that SoundEffectInstancePool gets updated
```

It is a badly written test in that the `sfx` is never explicitly disposed (and should be fixed, while adding a test for the `SoundEffect` finalizer). While running the test suite, a finalizer exception like this is thrown:
```
Unhandled Exception: System.InvalidOperationException: Failed to delete buffer. (Reason: Invalid Name)
   at Microsoft.Xna.Framework.Audio.ALHelper.CheckError(String message, Object[] args) in C:\Git\MonoGame\MonoGame.Framework\Audio\OpenALSoundController.cs:line 44
   at Microsoft.Xna.Framework.Audio.OALSoundBuffer.Dispose(Boolean disposing) in C:\Git\MonoGame\MonoGame.Framework\Audio\OALSoundBuffer.cs:line 107
   at Microsoft.Xna.Framework.Audio.OALSoundBuffer.Dispose() in C:\Git\MonoGame\MonoGame.Framework\Audio\OALSoundBuffer.cs:line 90
   at Microsoft.Xna.Framework.Audio.SoundEffect.PlatformDispose(Boolean disposing) in C:\Git\MonoGame\MonoGame.Framework\Audio\SoundEffect.OpenAL.cs:line 162
   at Microsoft.Xna.Framework.Audio.SoundEffect.Dispose(Boolean disposing) in C:\Git\MonoGame\MonoGame.Framework\Audio\SoundEffect.cs:line 438
   at Microsoft.Xna.Framework.Audio.SoundEffect.Finalize() in C:\Git\MonoGame\MonoGame.Framework\Audio\SoundEffect.cs:line 137
```

Could this be related to these changes? Can't debug more precisely at the moment.

mrhelmut:
Hmmm, most probably. Looks like @Tzenchor has been running into it too: https://github.com/mono/MonoGame/issues/2664#issuecomment-211804118

Strange thing is that it passed the test on the PR. I'm going to verify this now.

mrhelmut:
I can't reproduce it, and I don't see anything wrong in the finalizers. What's puzzling me is that ```AL.IsBuffer(openALDataBuffer)``` returns true but deleting the buffer fails.

I am using OpenAL-Soft dlls and not the creative's one, so maybe there's something hidden from me. Going to test the original one.

KonajuGames:
It will be the same for OpenAL as it is for OpenGL.  Cannot dispose of any unmanaged OpenAL resources if Dispose was called from Finalize() because Finalize() was called from the garbage collector thread, not the main thread.

Tzenchor:
I have just tried with the openal soft 64 bits lib, and the sound tests give errors, but no crashes until later:
```
Run Started: MonoGameTests
FF.....................................................................................
.................................................................EEEEEEEEEEE
EEEEEEEEEEEEEEEEEEEEE...........................................
............................................II.................
.I..............................................F..........F.EF..FF.EEE.......III.
Excepción no controlada: .System.ArgumentNullException: 
El valor no puede ser nulo.
 en System.Threading.Monitor.Enter(Object obj)
   en Microsoft.Xna.Framework.Threading.BlockOnUIThread(Action action)
   en D:\proyectos\MonoGame\MonoGame.Framework\Threading.cs:línea 195
 en Microsoft.Xna.Framework.Graphics.OcclusionQuery.Dispose(Boolean disposing)
   en D:\proyectos\MonoGame\MonoGame.Framework\Graphics\OcclusionQuery.OpenGL.cs:línea 80
 en Microsoft.Xna.Framework.Graphics.GraphicsResource.Finalize() 
   en D:\proyectos\MonoGame\MonoGame.Framework\Graphics\GraphicsResource.cs:línea 65
```

polsys:
@Tzenchor Yep, had those errors too occasionally. I guess they are related to the SDL switchover, though.

Our finalizers definitely need some attention, they are near impossible to get right and hard to test.

PR #4706, opened 2016-04-02
[Eto Dialogs] Edit Dialog

cra0zy:
Discussed in #4238

This needs: https://github.com/MonoGame/MonoGame.Dependencies/pull/75

Replaces Windows Forms and Gtk Edit Text Dialog with Eto.Forms one.

Dialog has also been improved in following ways:
 - Filenames are checked from: Path.GetInvalidFileNameChars
 - WinForms dialog now checks filenames
 - When renaming a file that has an extension, only file name without extension will be selected

Before left, after right:

Win Forms dialog now checks filename (Gtk was already doing it):


tomspilman:
Going to test this here and make sure the UI still feels like Windows after this change.

tomspilman:
@cra0zy - So with your changed merged I have a DPI issue with the Pipeline tool.

My laptop here is one of the super high DPI ones... I have issues with a few older tools, but the Pipeline tool always looked correct.

With merging your changes suddenly the Pipeline tool window appears really tiny on my screen.

Tools/Pipeline/Common/PipelineTypes.cs
```diff
@@ -408,6 +408,9 @@ private static void ResolveAssemblies(IEnumerable<string> assemblyPaths)
                 try
 #endif
                 {
+                    if (!asm.ToString().Contains("MonoGame"))
+                        continue;
```
tomspilman: This would fail on any custom pipeline extensions that don't include "MonoGame" in the file name.

We need a better way to filter this.

cra0zy: No, it won't fail because if you expand above and bellow you'll notice that that part is used for scanning referenced assemblies to the Pipeline tool, not referenced assemblies for the .mgcb file.

https://github.com/cra0zy/MonoGame/blob/ed1/Tools/Pipeline/Common/PipelineTypes.cs#L403

cra0zy:
> @cra0zy - So with your changed merged I have a DPI issue with the Pipeline tool.

I know what I broke, should be fixed now.

Tools/Pipeline/Common/PipelineTypes.cs
tomspilman: > not referenced assemblies for the .mgcb file.

Ahh...  got it.

tomspilman:
> I know what I broke, should be fixed now.

Yep... that looks fixed.

I love the fact this reduces the code size and unifies the dialog code making them work essitially the same across platforms.  This is all good for reducing bugs, making documentation easier to write, and our hopes of writing some tests against the GUI.

Merging!

PR #4710, opened 2016-04-02
Render faster if possible on fixed time step

hach-que:
Follow up from #2651 with the changes requested.

Currently when MonoGame is using a fixed time step, it won't ever render faster than 60FPS, despite the GPUs capability to do so.

This modifies the tick so that if `RenderMultipleFramesPerStep` is on, it will continue to render frames until the time is used up (instead of sleeping).

This is useful for diagnosing whether rendering is responsible for causing slow frames by giving an accurate representation of how fast the current frame can be rendered.

### DIFF
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -447,12 +447,25 @@ namespace Microsoft.Xna.Framework
                 // NOTE: While sleep can be inaccurate in general it is 
                 // accurate enough for frame limiting purposes if some
                 // fluctuation is an acceptable result.
+                if (!graphicsDeviceManager.RenderMultipleFramesPerStep)
+                {
+                    // NOTE: While sleep can be inaccurate in general it is 
+                    // accurate enough for frame limiting purposes if some
+                    // fluctuation is an acceptable result.
 #if WINRT
-                Task.Delay(sleepTime).Wait();
+                    Task.Delay(sleepTime).Wait();
 #else
-                System.Threading.Thread.Sleep(sleepTime);
+                    System.Threading.Thread.Sleep(sleepTime);
 #endif
-                goto RetryTick;
+                    goto RetryTick;
+                }
+                else
+                {
+                    // Draw until we have used up our time.
+                    DoDraw(_gameTime);
+
+                    goto RetryTick;
+                }
             }
 
             // Do not allow any update to take longer than our maximum.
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Game.cs
```diff
@@ -447,12 +447,25 @@ public void Tick()
                 // NOTE: While sleep can be inaccurate in general it is 
                 // accurate enough for frame limiting purposes if some
                 // fluctuation is an acceptable result.
+                if (!graphicsDeviceManager.RenderMultipleFramesPerStep)
+                {
+                    // NOTE: While sleep can be inaccurate in general it is 
+                    // accurate enough for frame limiting purposes if some
+                    // fluctuation is an acceptable result.
```
tomspilman: Can you remove one of these duplicate comments?

tomspilman:
I think we can support this.

Thoughts @dellis1972 @KonajuGames ?

mgarstenauer:
Hi @hach-que, could you elaborate on how this diagnose process works in practice. Someone who just sees the property probably does not understand what it is good for.

This flag could be used - for whatever reason - to run the Update "fixed step" and the Draw "free running". However, in this case the GameTime should be incremented between Draw calls. Is this the case? (I may have missed something, but I don't think that this is the case.)

hach-que:
So normally when you're running an XNA game, it will never render faster than 60FPS even with vsync off because at most it calls Draw once per Update/Draw step.

When you're trying to analyze how performant your Draw logic is, this is pretty limiting because you can only see if you're doing worse than 60FPS; you can't actually measure the graphics performance. 

This flag allows MonoGame to call Draw as many times as possible until the step time is consumed, giving you a much more accurate reading of how fast you are drawing every 16ms. This is especially useful when you have a built-in profiler and you want to find slower frames - even when those slower frames might not be under 60FPS.

hach-que:
And no, I don't believe GameTime should be updated as the game should not change state on subsequent calls, and should result in the same exact thing being rendered (which also makes this flag a good way of detecting when you're accidentally updating state in the Draw logic).

mgarstenauer:
Thx for the explanation. In this case, I guess it is necessary to document that GameTime is not updated and that it is intended for performance analysis.

hach-que:
I haven't used this in a while, and while it is present in Protogame, it isn't really used.

The original case for this was measuring graphics performance (i.e. how high can the FPS go?) as a way of determining whether your game was just barely meeting vsync / 60FPS requirements or whether it was exceeding it greatly (since vsync and other waits could mask performance when looking in a graphics debugger).

PR #1982, opened 2013-09-08
DynamicSoundEffectInstance for OpenAL

absurdhero:
We implemented DynamicSoundEffectInstance for OpenAL on Mac and iOS along with some integration tests.

We used a partial class to encapsulate the OpenAL specifics to keep the code more readable than using #if.

There is a race condition when rapidly playing and pausing that has been difficult to debug due to various issues with Xamarin Studio on Mac. Other than that issue, tests pass on Mac OS for playing, stopping, pausing, resuming, and looping clips.

TODO:

Test on PSM
Test on iOS
Integrate the existing DirectX pull request
Move more OpenAL calls into the OALSoundBuffer class

The TODO items might not all be required to integrate this pull request. I'd like to get input on what other work should be done on this patch.

PR #4715, opened 2016-04-03
Implement DynamicSoundEffectInstance

### LINKED ISSUE #4700: Implement DynamicSoundEffectInstance and Tests
We need to once and for all implement `DynamicSoundEffectInstance` across all platforms.

There was an effort 3 years ago that we were too dumb to get merged:  https://github.com/mono/MonoGame/pull/1982

We should use the ideas from it to help implement the feature for OpenAL and give the skeleton of the feature for XAudio platforms.

Note how the unit tests were written for this.  We should do the same... yes it won't tell us if the audio sounds right, but at least it will catch crashes or state problems.

polsys:
Fixes #4700. This PR adds `DynamicSoundEffectInstance` implementations on XAudio and OpenAL and some unit tests which also run against XNA. I've tested the code on Windows and really quickly on Android. I also hooked it into the iOS platform, but I'm not able to test it at all.

Some issues:
* Thread safety. I've left out any locks etc., but they should be easy to add. I think this code will generate some 1-in-10000 race conditions in multithreaded code.
* Garbage generation. For example #1982 recycled sound buffers, whereas this implementation does not. Should be easy to measure and fix.
* Disposal. I'm able to run the tests without any finalizer/`Dispose()` crashes, but I'm confident that there are issues left.
* ~~I used `nameof`, which is a C# 6 feature. Which language version are we using? Going back to explicit parameter names is a one-minute fix, if we want to support older compiler versions.~~
* ~~OpenAL implementation uses the `DynamicSoundEffectInstanceManager` helper class, which has some issues:~~ (Now shared with all implementations.)
  * ~~It depends on `GamePlatform` calling it. Plugging it into `FrameworkDispatcher` introduced in #4675 would solve that dependency.~~ (Now called by `Game`.)
  * ~~The great SDL change will move things all around the place.~~ (Didn't thanks to above.)
  * ~~It keeps a reference to the playing sound, which has side-effects if the user neither stops nor disposes the sound.~~ (Now holds a `WeakReference`.)
* ~~I know nothing about the Mac side, so I haven't touched it beyond the OpenTK version.~~ (The common SDL path hopefully clears things)
* The tests cannot confirm that the playback actually works, but rather test `PendingBufferCount`.
* And quite possibly many others, which will hopefully be found now that this is up here!


Tzenchor:
Current version is C# 5, so `nameof` is not working. 
Looking at the build log you can also see:

> [12:07:17][NAnt output]      [exec]   Audio\DynamicSoundEffectInstance.cs(74,55): error CS0103: The name 'nameof' does not exist in the current context [C:\BuildAgents\MonoGameWin1\work\f26e5a04a66fcbdb\MonoGame.Framework\MonoGame.Framework.Windows.csproj]

polsys:
Fixed. Should have thought about it, didn't notice until I had already pretty much written the PR message.

Tzenchor:
Just if you don't know how to find the build log, these are the current errors in the mac build
> [12:30:43][NAnt output]      [exec] /Applications/buildAgent/work/f26e5a04a66fcbdb/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs(77,27) : error CS1502: The best overloaded method match for `Microsoft.Xna.Framework.Audio.OALSoundBuffer.BindDataBuffer(byte[], MonoMac.OpenAL.ALFormat, int, int)' has some invalid arguments
[12:30:43][NAnt output]      [exec] /Applications/buildAgent/work/f26e5a04a66fcbdb/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs(77,50) : error CS1503: Argument `#2' cannot convert `OpenTK.Audio.OpenAL.ALFormat' expression to type `MonoMac.OpenAL.ALFormat'
[12:30:43][NAnt output]      [exec] /Applications/buildAgent/work/f26e5a04a66fcbdb/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs(84,27) : error CS1502: The best overloaded method match for `Microsoft.Xna.Framework.Audio.OALSoundBuffer.BindDataBuffer(byte[], MonoMac.OpenAL.ALFormat, int, int)' has some invalid arguments
[12:30:43][NAnt output]      [exec] /Applications/buildAgent/work/f26e5a04a66fcbdb/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs(84,56) : error CS1503: Argument `#2' cannot convert `OpenTK.Audio.OpenAL.ALFormat' expression to type `MonoMac.OpenAL.ALFormat'

On windows these tests are failing:

>  DynamicSoundEffectInstanceTest.BufferNeeded_DuringPlayback 
 DynamicSoundEffectInstanceTest.BufferNeeded_Play_AlreadySubmitted 
 DynamicSoundEffectInstanceTest.BufferNeeded_Play_NoneSubmitted 
 DynamicSoundEffectInstanceTest.Ctor 
 DynamicSoundEffectInstanceTest.GetSampleDuration 
 DynamicSoundEffectInstanceTest.GetSampleSizeInBytes 
 DynamicSoundEffectInstanceTest.IsLooped 
 DynamicSoundEffectInstanceTest.PendingBufferCount 
 DynamicSoundEffectInstanceTest.Playback 
 DynamicSoundEffectInstanceTest.Playback_Exceptions 
 DynamicSoundEffectInstanceTest.Stop_False 
 DynamicSoundEffectInstanceTest.Stop_RemovesBuffers 
 DynamicSoundEffectInstanceTest.SubmitBuffer_ParameterValidation_ComplexOverload 
 DynamicSoundEffectInstanceTest.SubmitBuffer_ParameterValidation_SimpleOverload 

All are throwing `nullPointerException` with the same stack trace:
> Test(s) failed. System.NullReferenceException : Object reference not set to an instance of an object.
   at SharpDX.XAudio2.SourceVoice..ctor(XAudio2 device, WaveFormat sourceFormat, VoiceFlags flags, Single maxFrequencyRatio, Boolean enableCallbackEvents, EffectDescriptor[] effectDescriptors)
   at SharpDX.XAudio2.SourceVoice..ctor(XAudio2 device, WaveFormat sourceFormat, VoiceFlags flags, Single maxFrequencyRatio, Boolean enableCallbackEvents)
   at SharpDX.XAudio2.SourceVoice..ctor(XAudio2 device, WaveFormat sourceFormat, Boolean enableCallbackEvents)
   at Microsoft.Xna.Framework.Audio.DynamicSoundEffectInstance.PlatformCreate()
   at Microsoft.Xna.Framework.Audio.DynamicSoundEffectInstance..ctor(Int32 sampleRate, AudioChannels channels)
   at MonoGame.Tests.Audio.DynamicSoundEffectInstanceTest.BufferNeeded_Play_NoneSubmitted()

polsys:
So if I understand correctly, the MonoMac failure is caused by different OpenAL headers, and can be fixed by replacing `using OpenTK.Audio.OpenAL` with
```cs
#if MONOMAC && PLATFORM_MACOS_LEGACY
using MonoMac.OpenAL;
#elif OPENAL
using OpenTK.Audio.OpenAL;
#endif
```
as is done in `SoundEffect.OpenAL.cs`.

The Windows test failure is really interesting, as apparently `Device` is null. Is the build server running a Server edition of Windows with no audio service enabled?

polsys:
So now it at least compiles on all platforms, which in itself is a positive surprise :smile:

tomspilman:
>  Is the build server running a Server edition of Windows with no audio service enabled?

The build machine is Windows 10 Pro and it does have audio hardware.

However I've learned that XAudio device creation fails if there are no speakers plugged into the PC.  That is weird...  it is like DirectX failing to initialized because there is no monitor plugged in.

 http://stackoverflow.com/questions/35600946/why-oncriticalerror-is-called-in-directx-version-of-xaudio2

This isn't your problem... it is a MonoGame issue to make this work better.

In the mean time i'm going to plug in some headphones to the build machine which should solve this issue.

MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs
```diff
@@ -0,0 +1,189 @@
...
+                {
+                    var buffer = _queuedBuffers.Dequeue();
+                    buffer.Dispose();
+                }
+
+                DynamicSoundEffectInstanceManager.RemoveInstance(this);
+            }
+        }
+
+        internal void UpdateQueue()
```
tomspilman: I guess this is all because OpenAL has no event fired when a buffer is finished processing?

In XAudio2 it has a callback... but it occurs from some internal XAudio2 thread correct?

So I guess at the moment these callbacks are occuring one different threads on different platforms.  Ideally we would make this more consistent.

Test/Framework/Audio/DynamicSoundEffectInstanceTest.cs
```diff
@@ -0,0 +1,366 @@
...
+        /// <summary>
+        /// Sleeps for the specified amount of time while calling FrameworkDispatcher.Update() every 10 ms.
+        /// </summary>
+        private static void SleepWhileDispatching(int ms)
+        {
+            int cycles = ms / 10;
+            for (int i = 0; i < cycles; i++)
+            {
+#if XNA
+                FrameworkDispatcher.Update();
```
tomspilman: Did you find that this was required to make XNA work for these tests?  Or did you do it just in case?

I ask because it gives us some insight into how our implementation should work.

tomspilman:
So did a first pass review... this is a great PR @polsys !

> Thread safety. I've left out any locks etc., but they should be easy to add.

Agreed.  Let's deal with that in a separate PR after this one is merged.  Our goal being to be at least as thread safe as the XNA implementation was, but make improvements if the difficulty and cost to performance isn't too high.

> For example #1982 recycled sound buffers, whereas this implementation does not. Should be easy to measure and fix.

Yes... let's do that to in a future PR.  Don't want to bog this first pass down with "optimization" like changes.

> OpenAL implementation uses the `DynamicSoundEffectInstanceManager` helper class

Yeah I don't like it either.  I want to understand the thread the buffer callbacks occur on under XNA before we decide what to do with this.


> It keeps a reference to the playing sound

You can use a weak reference like we do for `GraphicsResource`s to solve that.

> The great SDL change will move things all around the place.

It shouldn't hurt this too bad... we don't need to worry about it in this PR.

> The tests cannot confirm that the playback actually works, but rather test `PendingBufferCount`.

Yeah...  this is the best you can do there I think.  At least we know the system doesn't crash and it is processing... but it could be generating static.  The limitations of unit testing with audio.

tomspilman:
Looks like the build server is playing audio now.

KimikoMuffin:
For the record, I can confirm that BufferNeeded fires during the main thread in vanilla XNA. Which, as far as I can tell, just checks all running instances like the OpenAL version of this implementation ...

KonajuGames:
The BufferNeeded event was triggered in the mixer thread in XNA, hence the
warnings in the documentation about keeping the event handler short and
fast.  Invoking it on the main thread would cause stalls and gaps in the
output sound as it would have to wait for the appropriate time that the
main thread could be interrupted, i.e. between calls to Update and Draw.

KimikoMuffin:
I stand corrected!

theZMan:
>
> However I've learned that XAudio device creation fails if there are no
> speakers plugged into the PC. That is weird... it is like DirectX failing
> to initialized because there is no monitor plugged in.
>
Which  means you are replicating XNA4.0 perfectly ;-) it too crashes in
this scenario.

tomspilman:
> Which  means you are replicating XNA4.0 perfectly ;-) it too crashes in
this scenario.

Seems like a place we could do a better job then. :)

tomspilman:
> The BufferNeeded event was triggered in the mixer thread in XNA

I wonder if we can move the feeding of buffers into a thread too OpenAL.  This would then allow us to replicate that behavior and remove the update calls in the main thread.  However I have no clue how the OpenAL API will behave if called from a secondary thread like that.

Thoughts @polsys ?

polsys:
> So I guess at the moment these callbacks are occuring one different threads on different platforms. Ideally we would make this more consistent.

That's right, and I don't like it either. Now, the correct way to do it is a bit harder question.

The tests call `FrameworkDispatcher.Update()` periodically because `BufferNeeded` events are not raised otherwise on XNA. There is some main thread stuff going on, since my XNA test app stops receiving the events while dragging the game window.

On the other hand, changing the OpenAL implementation to run the callbacks on a secondary thread shouldn't be too hard. From quick googling it appears that OpenAL is *quite but not fully* thread-safe.

> You can use a weak reference like we do for  GraphicsResource s to solve that.

I can look into that soon.

> Which  means you are replicating XNA4.0 perfectly ;-) it too crashes in
this scenario.

Does it crash with a `NoAudioHardwareException`? We seem to throw it for OpenAL errors but not XAudio. At least we should crash with the correct error message.

tomspilman:
> There is some main thread stuff going on, 

 Interesting...  if the callback is coming from a thread in XNA then I wonder what the main thread could be doing.

> changing the OpenAL implementation to run the callbacks on a secondary thread shouldn't be too hard.

I think it could be worth looking into.


> Does it crash with a  NoAudioHardwareException ?

Let's not get into that issue here...  lets focus on `DynamicSoundEffectInstance`.  I have an issue for this separately in https://github.com/mono/MonoGame/issues/4717.

mrhelmut:
Glad to see this PR and this discussion. ```OggStreamer``` could be cleaned a lot by using ```DynamicSoundEffectInstance``` (it already runs on a separate thread).

tomspilman:
> 'OggStreamer' could be cleaned a lot by using 'DynamicSoundEffectInstance' (it already runs on a separate thread).

I'm hoping for that to be the case.  Then 'OggStreamer' can be used on any platform without much effort.

MonoGame.Framework/Audio/SoundEffect.cs
```diff
@@ -150,6 +152,8 @@ public static int GetSampleSizeInBytes(TimeSpan duration, int sampleRate, AudioC
         {
             if (sampleRate < 8000 || sampleRate > 48000)
                 throw new ArgumentOutOfRangeException();
+            if (duration < TimeSpan.Zero)
+                throw new ArgumentOutOfRangeException();
 
             // Reference: http://social.msdn.microsoft.com/Forums/windows/en-US/5a92be69-3b4e-4d92-b1d2-141ef0a50c91/how-to-calculate-duration-of-wave-file-from-its-size?forum=winforms
```
tomspilman: FYI.  Likely to be some conflicts with https://github.com/mono/MonoGame/pull/4469 in this file depending on which is merged first.

polsys:
Changed the OpenAL implementation to have only a `WeakReference` to playing instances. The tests failed again, I assume this is because the build got assigned to the other build agent this time?

How should the threading issue be solved, then? At the moment, the OpenAL version matches the XNA behavior, at least when it comes to my quick-and-dirty test app. Raising the event on main thread has its plus sides too: user code doesn't need to be thread-safe and it can always delegate the decoding work to a thread of its own.

polsys:
Of course the logical step is to write unit tests and make sure they pass on XNA and all MonoGame implementations. Then we at least match XNA behavior, and have our version better tested (even if we decide to take another approach). In fact, my current XAudio implementation is inconsistent - calling `Play()` with less than 3 buffers raises the event on the calling thread. I'll try to get the tests up tonight.

tomspilman:
> I assume this is because the build got assigned to the other build agent this time?

Well both Windows agents run on the same PC, so it shouldn't have failed...  i'll investigate.


> How should the threading issue be solved, then?

Well there are two main concerns.

First and foremost we want to behave the same as XNA...  that includes thread expectations.  Ideally if XNA's callback comes from a non-UI thread we would do the same to be consistent.

Second we want to be consistent across platforms when ever possible.  Having the callback come from different threads on different platforms could easily lead to bugs in user code.  So whatever we decide really needs to be the same on both platforms.

So do we break the XNA correct functionality of the XAudio2 implementation and redirect the callback on the UI thread?  Or do we thread the GL buffer check and have it fire callbacks from that thread?

IMO i'd rather see the GL code threaded as it is more correct to XNA and has a bonus of offloading some work from the main thread.  It is however tricky to get that threaded code right which is a challenge/danger.

> I'll try to get the tests up tonight.

Fantastic!

polsys:
My results on XNA behavior:
* The `BufferNeeded` event handler is called on the main thread - this is now verified by tests.
  * Looking at the stack trace in the event handler confirms this. `FrameworkDispatcher.Update()` raises the events.
* All the events are raised by `FrameworkDispatcher`, even the one raised because of calling `Play()` with not enough buffers. This is now verified by a test.

So it was the OpenAL version which worked correctly by raising the events on the main thread.

I split this in three commits: one for the tests, one for minor OpenAL fixes and one for XAudio main-threading (which moves the `DynamicSoundEffectInstanceManager` to the platform-independent part and therefore removes any `GamePlatform` dependencies). Now all the versions should match XNA threading behavior.

There is still some work left - I had some issues with audio popping in my test app. I'll investigate it with fresh eyes tomorrow, but putting these changes up for discussion anyway.

polsys:
I'll fix the Web build tomorrow too.

Test/Framework/Audio/DynamicSoundEffectInstanceTest.cs
```diff
@@ -324,7 +395,7 @@ private static void SleepWhileDispatching(int ms)
 #elif DESKTOPGL
                 DynamicSoundEffectInstanceManager.UpdatePlayingInstances();
 #endif
-                System.Threading.Thread.Sleep(10);
+                Thread.Sleep(10);
```
tomspilman: Just in case you are not aware... Sleep is not always accurate.  Hopefully your tests don't depend on that.

MonoGame.Framework/Audio/DynamicSoundEffectInstance.cs
```diff
@@ -290,4 +299,59 @@ private void RaiseBufferNeeded()
 
         #endregion
     }
+
+    /// <summary>
+    /// Handles the buffer events of all DynamicSoundEffectInstance instances.
+    /// </summary>
+    internal static class DynamicSoundEffectInstanceManager
```
tomspilman: Can you pull this out into its own source file?

MonoGame.Framework/Game.cs
```diff
@@ -664,6 +664,7 @@ internal void DoUpdate(GameTime gameTime)
                 // playing sounds to see if they've stopped,
                 // and return them back to the pool if so.
                 SoundEffectInstancePool.Update();
+                DynamicSoundEffectInstanceManager.UpdatePlayingInstances();
```
tomspilman: I guess we should talk about moving both of these into `FrameworkDispatcher`?

tomspilman:
> My results on XNA behavior:

Interesting... not what I expected, but good to see we can easily replicate it and be XNA compatible.


> There is still some work left 

No problem... great work so far!

polsys:
> Sleep is not always accurate. Hopefully your tests don't depend on that.

The tests have some safety margin, but a random failure is not completely out of the question.

> Can you pull this out into its own source file?

Done. Also fixed the Web build, hopefully.

> I guess we should talk about moving both of these into  FrameworkDispatcher ?

It's the perfect place for that call, and the `SoundEffectInstancePool.Update()` call is already moved in the `FrameworkDispatcher` PR. I've kept this PR independent of #4675, but it's deliberately simple to update the other when one gets merged. In fact, this is the main reason I implemented `FrameworkDispatcher`.

I'd really appreciate if someone can test this code as it is now. My brittle test app has some glitches under XAudio, and I haven't been able to isolate the cause. It might be an oversight in the framework code - or just my old quick'n'dirty proof-of-concept breaking.

polsys:
Tested and had no problems (related to `DynamicSoundEffectInstance`, that is) with the current state of SDL switchover.

The XAudio glitching seems to happen consistently, in both single- and multi-threaded cases and on separate computers, so it definitely is an issue. It might be linked to 95c9379dc4290453113230cf54651988c515a91a (making events to occur on the main thread), but I don't know why.

dellis1972:
I just saw this PR. Awesome work @polsys

RHY3756547:
I've been trying this out, and it works great apart from a few small issues on the OpenAL side of things. I needed to add these lines to PlatformPlay() to stop my other looping sound effects from breaking the DynamicSoundEffectInstances:

```csharp
            AL.Source(SourceId, ALSourceb.Looping, IsLooped); //obviously false...
            ALHelper.CheckError("Failed to set source loop state.");
```

There's also some stutter on buffer boundaries, which I'm not getting running my game under DirectX. I'm using it for streamed mp3 playback.

polsys:
Thanks @RHY3756547! Pushed the looping fix, a great catch.

The stutter issue is very interesting, since I've only seen it on DirectX. It should be sorted out before merging, but I have no idea about its cause.

SoundGoddess:
So what all has to be done to get this working?  Seems like it's just missing some null reference checks from the test details.

polsys:
The `NullReferenceException`s are related to a missing XAudio device, which is a build server issue. The correct thing to do would be throwing a `NoAudioHardwareException`, but that should be fixed separately for all XAudio code.

What this needs is some real-world testing, and the stuttering issue should be fixed. Unfortunately I have very limited time as of now, so I cannot debug it.

tomspilman:
Working on fixing the build server to get this merged... testing now.

tomspilman:
Looking good...  but still testing.  I literally had to plug in powered external speakers to the MonoGame build box and keep them powered on.  Going to try it after the build box gets a chance to sleep and be sure it still works.

If all looks good i'll merge and hope we don't get randomly failing builds later.

theZMan:
If its the same as XNA in that you need some kind of output device you can plug in a cheap pair of earbuds (or maybe just a headphone jack itself) and it will work. No need for actual speakers.

tomspilman:
@theZMan - That is what i thought too...  but a pair of ear buds did not work before.

This is a headless Win10 box with the build runners executing in a logged in desktop session.  I disabled all audio output devices except for the speakers and made them the default to avoid any confusion with the hdmi output on the headless system.

tomspilman:
Ok...  looks like this is working now.  I'm pulling the trigger on it.

Thanks for the work @polsys !

SoundGoddess:
thank you'll for your hard work on this, good timing I was hoping to figure out how to get this up and running in MG this weekend and that is the only thing missing ~ http://www.david-gouveia.com/portfolio/creating-a-basic-synth-in-xna-part-ii/

tomspilman:
Argh... forgot about the OpenAL changes.  I'll submit a PR doing the fixes here in a sec.

tomspilman:
@polsys - If you get a second please take a look at this post and see if you can suggest as to what could be going wrong in `DynamicSoundEffectInstance`:

http://community.monogame.net/t/just-trying-to-figure-out-how-to-play-a-simple-sin-wave/7938

polsys:
That's the buffer boundary issue I've mentioned and failed to pin down. I created a new issue (#5023) to track this.

DannyGlover:
Great work on this. I'm using C# and this doesn't seem to be in the latest MonoGame release: MonoGame.Framework.WindowsDX Latest version: 3.5.1.1679

How can I go about grabbing the latest development build that contains this feature?
Thanks!

Jjagg:
You can get it from the [Downloads page](http://www.monogame.net/downloads/) on the MonoGame website under "Development Builds".

DannyGlover:
@Jjagg Thank you very much. I figured it out shortly after posting.

@polsys I'm getting the following error when trying to assign a DynamicSoundEffectInstance instance like so: 

`_instance = new DynamicSoundEffectInstance(44100, (AudioChannels)2);`

Error:

`Object reference not set to an instance of an object`

Am I doing this correctly?

polsys:
@DannyGlover That is the correct way to create an instance. I suggest you create a new issue, preferably including the exception stack trace as well. I can look into this later today.

PR #4716, opened 2016-04-03
Packed vector tests and fixes

### LINKED ISSUE #4685: Add Ordering Tests For PackedVector
Our tests for packed vector types only test against a zero and one types... it does not include tests with different values for each component.  This means the tests do not correctly detect errors in the packing order.

We should add a test for packing order to each of the PackedVectorTests types in:

https://github.com/mono/MonoGame/blob/develop/Test/Framework/PackedVectorTest.cs

Use the test in the `Rgba64` test as the example as to how to form the new tests:

https://github.com/mono/MonoGame/blob/develop/Test/Framework/PackedVectorTest.cs#L85

While we are at this the coverage of the PackedVectorTest tests is pretty poor.  We should expand these to include testing:

   - `IPackedVector.ToVector4()` across all types.
   - Equality methods across all types.
   - `HalfSingle`
   - `HalfVector2`
   - `HalfVector4`

Tzenchor:
Fixes #4685 
This should fix all differences between the Packed Vector classes between XNA and MonoGame. 
The main difference between them was the data ordering, but also that XNA was rounding the values while packaging them, so now the following test is working while previously wasn't
```
// Test packing order.
Assert.AreEqual(0xC7AD8F5C570A1EB8, new Rgba64(0.12f, 0.34f, 0.56f, 0.78f).PackedValue);
```

Apart from the fixes I also did a few upgrades, and in a few cases I replaced the tabs with spaces.

Also `Short2` or `NormalizedShort2` still contains the old license information, should I update it?

### DIFF
--- a/MonoGame.Framework/Graphics/PackedVector/Alpha8.cs
+++ b/MonoGame.Framework/Graphics/PackedVector/Alpha8.cs
@@ -120,7 +120,7 @@ namespace Microsoft.Xna.Framework.Graphics.PackedVector
 
 		private static byte Pack(float alpha)
 		{
-			return (byte) (
+			return (byte) Math.Round(
 				MathHelper.Clamp(alpha, 0, 1) * 255.0f
 			);
 		}
--- a/MonoGame.Framework/Graphics/PackedVector/Alpha8.cs
+++ b/MonoGame.Framework/Graphics/PackedVector/Alpha8.cs
@@ -6,123 +6,123 @@ using System;
 
 namespace Microsoft.Xna.Framework.Graphics.PackedVector
 {
-	/// <summary>
-	/// Packed vector type containing a single 8 bit normalized W values that is ranging from 0 to 1.
-	/// </summary>
-	public struct Alpha8 : IPackedVector<byte>, IEquatable<Alpha8>, IPackedVector
+    /// <summary>
+    /// Packed vector type containing a single 8 bit normalized W values that is ranging from 0 to 1.
+    /// </summary>
+    public struct Alpha8 : IPackedVector<byte>, IEquatable<Alpha8>, IPackedVector
     {
         private byte packedValue;
 
-		/// <summary>
-		/// Gets and sets the packed value.
-		/// </summary>
-		[CLSCompliant(false)]
-		public byte PackedValue
-		{
-			get
-			{
-				return packedValue;
-			}
-			set
-			{
-				packedValue = value;
-			}
-		}
+        /// <summary>
+        /// Gets and sets the packed value.
+        /// </summary>
+        [CLSCompliant(false)]
+        public byte PackedValue
+        {
+            get
+            {
+                return packedValue;
+            }
+            set
+            {
+                packedValue = value;
+            }
+        }
 
-		/// <summary>
-		/// Creates a new instance of Alpha8.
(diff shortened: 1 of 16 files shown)


tomspilman:
> Also `Short2` or `NormalizedShort2` still contains the old license information, should I update it?

Yeah go ahead and fix the headers if you're already doing fixes in there anyway.

tomspilman:
We still need tests for `HalfSingle`, `HalfVector2`, and `HalfVector4`, but we can wait for another PR to address that part of #4685.

mgarstenauer:
I am not sure if it is the goal of this PR to fix the rounding behavior. In any case, here is some additional information that might be helpful:

As far as I can see, you do not explicitly specify the rounding method (see [Math.Round(double, MidpointRounding)](https://msdn.microsoft.com/en-us/library/ef48waz8.aspx) and [MidpointRounding](https://msdn.microsoft.com/en-us/library/system.midpointrounding.aspx)). `Math.Round(double)` uses `MidpointRounding.ToEven` by default. The [Data Conversion Rules](https://msdn.microsoft.com/en-us/library/windows/desktop/dd607323.aspx) explicitly specify which rounding method to use:

> Round-to-zero is used during conversion to another float format. If the target is an integer or fixed point format, round-to-nearest-even is used, unless the conversion is explicitly documented as using another rounding behavior, such as round-to-nearest for FLOAT to SNORM, FLOAT to UNORM or FLOAT to SRGB.

So, if I am not mistaken, we need to use `Math.Round(x, MidpointRounding.AwayFromZero)` for all FLOAT to SNORM and FLOAT to UNORM conversions.

tomspilman:
> I am not sure if it is the goal of this PR to fix the rounding behavior.

The main goal is reproducing the results of XNA.  @Tzenchor is running the same tests on XNA and ensuring we're getting the same results in MonoGame.

That said if XNA does something in a wrong way... I totally would like to fix it.

Tzenchor:
> So, if I am not mistaken, we need to use Math.Round(x, MidpointRounding.AwayFromZero) for all FLOAT to SNORM and FLOAT to UNORM conversions.

I have only tried with `RGBA64`, `NormalizedShort4` and `NormalizedByte4`, and can say that XNA is using `MidpointRounding.ToEven`.

> We still need tests for `HalfSingle`, `HalfVector2`, and `HalfVector4`, but we can wait for another PR to address that part of #4685.

Done, but review them as I may have missed some cases

tomspilman:
This looks good to me @Tzenchor !

@KonajuGames this likely affects some of the texture pipeline stuff, but all the tests pass so for now we can assume all is well.  We should keep an eye out for any issues and remember of these changes.

Tzenchor:
> @KonajuGames this likely affects some of the texture pipeline stuff, but all the tests pass so for now we can assume all is well. We should keep an eye out for any issues and remember of these changes.

The TextureImporter is safe, its possible outputs are: Color (not a PackedVector format), RGBA64 (had problems before #4683) and Vector4 (not a PackedVector format). The rounding error only affects when using inexact values, but the TextureImporter is copying the exact ones.

Also I have just done a fast run of my game with this library and the images were as always

tomspilman:
Cool...  going to merge this then!  Thanks @Tzenchor !

PR #4720, opened 2016-04-04
Separate out effect and shader reading from effect / shader classes

hach-que:
This addresses #4678 by allowing effects to be created without relying on the MGFX binary file format.


tomspilman:
@dellis1972  @KonajuGames - Thoughts?

I don't have an issue with the general premise of this PR.  Moving the MGFX reading code out of the `Effect` class would be a good cleanup.  Also the main thrust of this PR....  providing a way for engines/tools to provide data to `Effect` that doesn't necessarily exist on disk or from the 2MGFX tool...  isn't a bad extension to the system.

My only issues are with the details of this implementation.

My main issue is that `IEffectReader` and `IShaderReader` should be changed from interfaces to abstract classes.  Interfaces should be reserved when a type must implement multiple behaviors or the same behavior is used in types that have different base types... like `IDisposable`, `IEnumerable` and `IComparable`.  Abstract classes should be preferred over interfaces in all other cases.

Second... not super critical to do...  but it would be good to come up with a better name than "Reader" for these.  Reader to me reminds me of the `BinaryReader` and other type of stream classes.  This isn't quite like that as it loads the effect into memory on construction and then gives the caller access to the data read later.

Last is there a better way to provide this functionality?  Should instead consider making more of the internals of `Effect` public to allow anyone to construct one as needed?  Like if a constructor for `Effect` existed that allowed you to provide the techniques and parameters?  Is this better or worse than this loader/reader based extension method?

hach-que:
The primary reasoning for this PR isn't really around any more - originally I wanted to have my own shader/effect format which could construct effects directly instead of MGFX.  However, that never really materialised and it's likely I'll just delete the only code that depends on this from Protogame (since the implementation was never finished).

PR #4723, opened 2016-04-04
[Eto Dialogs] Add Item Dialog

cra0zy:
Discussed in #4238

Stuff done:
 - implements AddItemDialog using Eto.Forms
   - this replaces 2 WinForms dialogs and 1 Gtk dialog with 1 Eto dialog, 2 + 1 gratis
 - makes Eto Dialogs use headerbar on Linux if available :)

Before left, after right:
del


KonajuGames:
Why is Copy disabled on the GTK version?

cra0zy:
> Why is Copy disabled on the GTK version?

File already existed in the folder. 

https://github.com/cra0zy/MonoGame/blob/ed2/Tools/Pipeline/Eto/Dialogs/AddItemDialog.cs#L28

KonajuGames:
> File already existed in the folder.

By default this would then result in two output files in the same directory with the same name (one in the directory and one linked to the directory).  Do we have a check for that?  Without #3009, there is no way for the user to fix this.

cra0zy:
> By default this would then result in two output files in the same directory with the same name (one in the directory and one linked to the directory). Do we have a check for that? Without #3009, there is no way for the user to fix this.

First of all this is what Gtk and WinForms are currently doing.

That said how about I add an error message saying "file" already exists in target directory?

KonajuGames:
> That said how about I add an error message saying "file" already exists in target directory?

That sounds good to me.

cra0zy:
> That sounds good to me.

Well here is something, linked content is really broken, adding linked content like I did above will create a link, but that doesn't mean it will put compiled stuff in the build folder.

Example:

/home/harry/Documents/test.mgcb
/home/harry/Pictures/1.png     //added as link

Result:

/home/harry/Documents/Pictures/1.xnb

Now that is really broken. I am gonna leave this PR as is (it's just copying Gtk/WinForms Pipeline), and will open up an issue about properly fixing linked content.

cra0zy:
Added #4724

Please do get this merged before someone tries to fix up how linked content works.

KonajuGames:
That makes sense.  Merging.  Thanks @cra0zy

tomspilman:
@cra0zy - Can you change "Add File Action" to just "Add File" in your next PR?

cra0zy:
> @cra0zy - Can you change "Add File Action" to just "Add File" in your next PR?

Sure.

Side note, since I am working on templates dialog, is it ok if I include templates for the following items:
 - text
 - png image
 - PSD image
 - blend model
 - fbx model

tomspilman:
I really don't see the point of those templates.  Of what benefit to the user is a PNG being added of the wrong size, color depth, and content to the project?  Same with PSD.

I could see with FBX maybe adding specific models...  like the classics: Teapot, Bunny, Dragon, etc.  This would provide temp content that could be used for testing.

Still the original goal for the templates was to provide starter files for stuff that was MonoGame specific.  We have a template for .spritefont because it is our format and no other tool creates one.

However tools easily create all the files you listed.

So what is the benefit of adding these templates?

cra0zy:
> So what is the benefit of adding these templates?

My idea was that if user just wants to create their own image, they could just create a .png image from a blank template, it is convenient for the file to be autoplaced correctly.

If you want to create a text file in the folder that you are already in, are you gonna open up notepad and then navigate again to the same folder or are you gonna right click and create new text file.

So my idea is to just provide blank templates for few basic formats that are used.

> I could see with FBX maybe adding specific models... like the classics: Teapot, Bunny, Dragon, etc. This would provide temp content that could be used for testing.

That's totally not what I was thinking... I could still see it be useful.

tomspilman:
Also it doesn't help that most of the time PNGs and FBX files are results of another file...  not the source format that people edit.  The artists in my office don't edit PNGs and FBXs.  They edit PSDs and 3dsmax files and export PNGs and FBXs from them.

IMO the only useful thing here is using it to create temp models for testing.  I see little to no benefit to the rest.  @KonajuGames ?  @dellis1972 ?

Also note the dragging and dropping a file in the tree view caused this new "Add File" to appear at 0,0.  It should at least be centered to the Pipeline window.

cra0zy:
> Also note the dragging and dropping a file in the tree view caused this new "Add File" to appear at 0,0. It should at least be centered to the Pipeline window.

It will be like that until I can implement all the dialogs, once I do I can properly setup the parrent window.

KonajuGames:
I agree with Tom here with templates.  They are best generated by the tools designed for that job, rather than create a placeholder and have to replace it.

cra0zy:
@tomspilman @KonajuGames What do you think on using combobox instead of treeview/listview:

test

Why? Well treeview/listview are good when you have a lot of information, but when there are only 4 items it looks half empty and like it's missing stuff so using something like combobox makes much more sense design wise.

theZMan:
But drop downs require 2 clicks so are really only useful when you are constrained for space.

My vote would be to continue with the list box because it mirrors the Visual Studio add-item experience and we should be consistent with that.

cra0zy:
> But drop downs require 2 clicks so are really only useful when you are constrained for space.

No, drop downs are also useful when you need to select between few values, treeviews/listviews are useful when you have to select larger quantity of values.

> My vote would be to continue with the list box because it mirrors the Visual Studio add-item experience and we should be consistent with that.

That would be kind of contradictory because it's not supposed to have actual templates for everything,(as discussed above) so it can be a bit confusing to why it has so little templates (you must say that you get the feeling like it's missing stuff).

cra0zy:
Now you can critic me:

untitled

cra0zy:
untitled

PR #4726, opened 2016-04-04
Remove support for reading raw assets through ContentManager.

KonajuGames:
Use the FromStream methods for the relevant classes to load raw assets.


tomspilman:
Ok.... here we go.

Issue #4697, opened 2016-04-01
MonoGame UWP on XBox One Fixes


tomspilman:
I ran into a few issues porting a quick test to Xbox One under UWP.

 https://www.youtube.com/watch?v=JWzwNUGr9Tk

 - It doesn't want to create a 1920x1080 back buffer....  it is like 1832x1023 or something weird.
 - There is a thick white border around the display area... might be why I am not getting a 1080p back buffer.
 - For some reason XB1 doesn't like how our static VBs and IBs are created... it fails to copy data into them, however dynamic buffers work fine.

The first two i'm told are new features that MS added to make apps seamlessly work on TVs within the safe area.   There are APIs to disable these things which I think we would want to do at the platform level...  games deal with the safe area on their own.

The last one i'm still trying to figure out.  We don't have issues with static VBs or IBs on any other UWP platform...  Windows 10, Windows 10 Mobile, or even Rasberry Pi 2/3.  So either we're doing something wrong that all the other platforms are tolerant of and let slip by...  or the XB1 Win10 OS has a bug in this area.

brianost:
To remove the safe area border and use the full size of the window on Xbox One, the ApplicationView.DesiredBoundsMode property needs to be set to ApplicationViewBoundsMode.UseCoreWindow. Otherwise the XAML framework will restrict layout to what it believes is the visible area of the window, which is just the safe area in the middle.

tomspilman:
> the ApplicationView.DesiredBoundsMode property needs to be set to ApplicationViewBoundsMode.UseCoreWindow. 

Great to know!  Thanks @brianost !

I guess we should toggle this as part of the normal "IsFullScreen" state.

brianost:
I think for the purposes of MonoGame it is safe to leave it set to UseCoreWindow all the time. Games are generally going to want to use the entire window area.

I'd keep IsFullScreen tied to the ApplicationView.IsFullScreenMode. On Xbox One that'll represent whether there is a snap app next to the game, or it is really fullscreen. ApplicationView.TryEnterFullScreenMode will attempt to kick out the snap app and make the game fullscreen if possible.

tomspilman:
> I think for the purposes of MonoGame it is safe to leave it set to UseCoreWindow all the time.

Maybe then I misunderstood what it did.  I assumed that on a Windowed UWP game that if UseCoreWindow was set we would loose the title bar on the window.  But I guess you're saying we would not.

brianost:
[...]

PR #4727, opened 2016-04-05
Fixed layout of UWP windows in VS template to ignore window chrome.

tomspilman:
This fixes the layout to use entire core window on UWP apps.  This helps in particular on Xbox One where it otherwise adds borders around your swap chain to keep the app from the safe zone area (something done manually in games).

Part of the fixes for #4697 .


KonajuGames:
Merging.

PR #4728, opened 2016-04-05
[Eto Dialogs] New Item Dialog

cra0zy:
Discussed in #4238

Replaces new item/new template dialog.

Stuff done to the dialog:
 - redesigned the layout
   - headerbar and normal version can now use the same layout
   - less pointless empty space
 - implement check for if the file exists in target directory
 - always shows the extension that's gonna be added to the name

Before left, after right:
untitled


tomspilman:
Why is the new dialog a different size on Mac/Linux than Windows?

tomspilman:
Also are header bar buttons a very common thing on these platforms?  Seems weird to me.  Don't want MonoGame tools to be the weirdo tool that does things different from all other tools.

cra0zy:
> Why is the new dialog a different size on Mac/Linux than Windows?

They are actually pretty similar in sizes, headerbar counts into client size so it eats up a bit of height (and width a small amount), look at bellow image.

> Also are header bar buttons a very common thing on these platforms?

Common on GNOME, not common for the rest, so depending on if it's running under GNOME it will decided if it will use it: https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Eto/Global.cs#L19

This is what it looks like if it's not running under GNOME:
untitled

tomspilman:
> They are actually pretty similar in sizes

The dialogs are clearly different sizes...  but I guess the smaller font size on Windows is fooling me a bit too.


> This is what it looks like if it's not running under GNOME:

Ok... as long as it is only different in places where it is expected.

cra0zy:
> The dialogs are clearly different sizes... but I guess the smaller font size on Windows is fooling me a bit too.

Different DPI more likely.

tomspilman:
True.

PR #4732, opened 2016-04-06
Added CHANGELOG.md to project.

tomspilman:
This PR adds a CHANGELOG.md file to the repo populated with our old release information gathered from the website and other sources. I also hooked this into the documentation.

Please take a look and give me feedback on the formatting I've chosen for the change log:

  https://github.com/SickheadGames/MonoGame/blob/changelog/CHANGELOG.md

Am I missing anything, should I change anything in how we format this?

The next step will be integrating a tool to automatically generate the new release information using the information in PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,335 @@
+# Change Log
+
+
+## 3.5.1 Release - 3/30/2016
+
+ - Fixed negative values when pressing up on left thumbstick on Mac.
+ - Removed exception and just return empty state when requesting an invalid GamePad index.
+ - Fixed texture processing for 64bpp textures.
+ - Fixed Texture2D.SaveAsPng on Mac.
+
+
+## 3.5 Release - 3/17/2016
+
+ - Content Pipeline Integration for Xamarin Studio and MonoDevleop on Mac and Linux.
+ - Automatic inclusion of XNBs into your final project on Mac and Linux.
+ - Improved Mac and Linux installers.
+ - Assemblies are now installed locally on Mac and Linux just like they are on Windows.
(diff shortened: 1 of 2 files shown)


tomspilman:
Should we add links to the tags and diffs for each shipped revision like this changelog does:

https://github.com/skywinder/ActionSheetPicker-3.0/blob/master/CHANGELOG.md

Does anyone really care?  Of the people that care wouldn't they already know where to look for this information anyway?

KonajuGames:
Is that easily doable?  It does give more context to the single line change
note, but I'm not going to be upset if it isn't there.

tomspilman:
> Is that easily doable?

The tool I've got will do it automatically for our new releases... so it could be easily doable.  It does however make the plain text .md file harder to read with lots of lengthy urls in places.

KonajuGames:
How much would we be reading the plain text?

createdbyx:
The changelog in my opinion is much better with links ala 
https://github.com/skywinder/ActionSheetPicker-3.0/blob/master/CHANGELOG.md

Otherwise readers can only guess at what specifically the listed changes are referring to. Having links regardless of how messy it makes the plain text is a small price to pay for being able to just click the link and go directly to the change related to the log entry.

KonajuGames:
I believe it was 100% in support for adding the links.  For future releases it will add links?

KonajuGames:
I won't hold this up on that question though.  Merging.

tomspilman:
> For future releases it will add links?

Yeah... the old releases had its change logs all built by hand.  So the only way to get links for those would be someone manually finding the links an adding them.

For new releases the change log generation tool will create the links automatically.

PR #4739, opened 2016-04-07
Fixed back buffer MSAA on DirectX platforms.

tomspilman:
This PR adds some unit tests for `GraphicsDeviceManager` and fixes back buffer MSAA to work again under DirectX.

The MSAA stuff is not totally correct yet, but this is a good start.  We really need to refactor the graphics device creation so that we can query the MSAA support before creating the back buffer.  This will come in a future PR.

Also the unit test for checking if the back buffer MSAA is working is partially disabled under MonoGame.  This will be fixed after https://github.com/mono/MonoGame/pull/2382 is merged.


KonajuGames:
Merging.

PR #4742, opened 2016-04-07
Fixed thumbstick virtual buttons to always use independent axes.

### LINKED ISSUE #4741: Thumbstick virtual buttons don't always behave correctly
I'm using the thumbstick "virtual buttons" (e.g. `Buttons.LeftThumbstickUp`) for navigation of my menu system when using a controller. I noticed that it was quite difficult to move up/down without accidentally also triggering left/right movement (and vice versa).

Looking into it further, I found it was because I was using a circular dead-zone mode. If I changed to the independent axes mode, then it was much less sensitive and easier to use. However, I need the circular dead-zone mode for player movement so this wasn't a practical solution.

When I compared this behaviour to XNA, I found that XNA seems to always use the independent axes mode when determining the "virtual buttons" state regardless of the dead-zone mode you specify. This would make sense as you need the independent axes mode to properly separate up/down from left/right movement.

So I believe the solution to this is to always use the independent axes mode when determining the thumbstick virtual buttons state. This should make MonoGame behave similarly to XNA in this respect. I'm currently working on a PR which does this and will submit it shortly.

MichaelDePiazzi:
Fixes #4741.

This PR changes how the thumbstick "virtual buttons" (e.g. `Buttons.LeftThumbstickUp`) are set so that it more closely matches XNA behaviour. Basically it now always uses the "independent axes" mode to determine the virtual buttons state regardless of the actual dead-zone mode being used.

### DIFF
--- a/MonoGame.Framework/Input/GamePad.XInput.cs
+++ b/MonoGame.Framework/Input/GamePad.XInput.cs
@@ -198,10 +198,6 @@ namespace Microsoft.Xna.Framework.Input
 
             var buttons = ConvertToButtons(
                 buttonFlags: gamepad.Buttons,
-                leftThumbX: gamepad.LeftThumbX,
-                leftThumbY: gamepad.LeftThumbY,
-                rightThumbX: gamepad.RightThumbX,
-                rightThumbY: gamepad.RightThumbY,
                 leftTrigger: gamepad.LeftTrigger,
                 rightTrigger: gamepad.RightTrigger);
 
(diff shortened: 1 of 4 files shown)


tomspilman:
@mrhelmut was the last person working in this area, so lets see what he thinks.

The only thing I would want to see added to this is some form of unit test to validate the fix and detect if we break it again in the future.

Seems like these tests could be extended to check this case:

 https://github.com/mono/MonoGame/blob/develop/Test/Framework/GamePadThumbSticksTest.cs
 https://github.com/mono/MonoGame/blob/develop/Test/Framework/GamePadStateTest.cs

MichaelDePiazzi:
@tomspilman No problem. I'll see what I can do in regards to some tests.

mrhelmut:
Makes perfect sense. :+1: 
Code is fine for all platforms.

tomspilman:
Thanks @MichaelDePiazzi !

MichaelDePiazzi:
Ok, I've added some tests which test this fix. I had to use an internal constructor of `GamePadThumbSticks` so that I could apply a dead-zone mode (which is also why these tests won't work with XNA). But I made sure that I was only asserting against the public API (note the `GetAllPressedButtons` helper method which utilises `GamePadState.IsButtonDown` to access the "virtual buttons").

Let me know if you feel that any further tests are required.

MichaelDePiazzi:
P.S. I also run these tests against the previous code to prove that there were failing test cases first.

tomspilman:
Thanks @MichaelDePiazzi ... merging!

PR #4746, opened 2016-04-09
Two MaterialProcessor properties fixed

KonajuGames:
Compiler warnings showed that two MaterialProcessor properties were not using their private backing fields, therefore the default values assigned to those fields were never used.


tomspilman:
Makes sense... merging!

PR #4748, opened 2016-04-09
[Pipeline Tool] Fix loading/saving of settings for currupt or unaccessible files

### LINKED ISSUE #4733: Make sure Pipeline tool starts even if Settings.XML is corrupted.
Looks like if the Settings.XML for the Pipeline GUI tool is screwed up the application silently fails to start.

 http://community.monogame.net/t/pipeline-exe-not-opening-resolved

We should protect against those cases and ensure the tool always starts no matter if the settings are corrupted or even if a command line option is incorrect.

cra0zy:
Fixes #4733


KonajuGames:
I'll merge this one because it handles the save as well.

PR #4749, opened 2016-04-09
[Eto Dialogs] Reference Dialog

cra0zy:
This needs: MonoGame/MonoGame.Dependencies#77

Before left, after right:
untitled


tomspilman:
I just merged the dependencies.

cra0zy:
> I just merged the dependencies.

Dependencies are updated.

tomspilman:
Is that window still resizable?

cra0zy:
> Is that window still resizable?

Yup.

tomspilman:
Ok...  i'll merge this.

I'd like to see us fix the margin between the controls and the borders on this and the other new dialogs.  It is really tight... looks wrong under Windows.

cra0zy:
> I'd like to see us fix the margin between the controls and the borders on this and the other new dialogs. It is really tight... looks wrong under Windows.

Sure, just post a screenshot of some dialog and I'll try to copy the margin.

Also I did think a head of time and made a "DialogBase" class which is responcible for positioning OK/Cancel buttons and the main Content area:
https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Eto/Dialogs/DialogBase.eto.cs so all I have to do is modify it once and it will change it for all the dialogs :)

KonajuGames:
> Sure, just post a screenshot of some dialog and I'll try to copy the margin.

The original Windows reference dialog above has the correct margins between the window edges and the controls.

PR #4751, opened 2016-04-10
[Eto Dialogs] Delete Dialog

cra0zy:
This dialog is Linux only, it will be connected to Windows side once I am finished with replacing all the dialogs.

Extra stuff done:
 - bigger spacing next to Ok/Cancel buttons
 - moved some shared logic of ReferenceDialog into controller
 - fixed newtemplate crash when the name remains unchanged

Before left, after right:
untitled

PS. If a folder is selected it will scan all the subfolders and files and also add them to the list, not just the items that are shown in the Pipeline Tool.


KonajuGames:
Having the extra checkbox to click before being able to click `Delete` is not necessary.  We're already getting them to confirm by showing this dialog.  We don't need them to confirm it twice.

The checkbox to confirm/agree is only common in a EULA during an install, and then it's only once.  Deleting files may happen many times, and requiring the extra click will just be annoying.

The treeview is good, but I don't agree with the checkbox.

cra0zy:
The only reason I added checkbox is because I am used to Ok button being on right and I managed to accidentally press the Delete button instead of Cancel button...

I'll remove it in 10 mins, working on another PR and don't want to switch between branches.

KonajuGames:
The Windows standard is OK/Cancel.  The Linux and Mac standard is Cancel/OK.  We should stick to OS standards for dialog button positioning.

cra0zy:
> The Windows standard is OK/Cancel. The Linux and Mac standard is Cancel/OK. We should stick to OS standards for dialog button positioning.

Ok, will fix both in 5 mins.

cra0zy:
@KonajuGames Checkbox removed, dialog buttons check to see if it's Unix before placement so LInux and Mac get OK on right while Windows gets it on the left.

KonajuGames:
Thanks @cra0zy. Merging.

PR #4753, opened 2016-04-10
[Eto Dialogs] About Dialog

cra0zy:
This dialog was Windows only since Gtk has a built-in dialog, I thought a bit about it and decided to use the same dialog for all systems, easier maintenance.

Disregard the version, it get's updated by the build script.

Before left, after right:
untitled


KonajuGames:
This one looks good, except I have noticed that the buttons in these Eto dialogs tend to be too close to the window borders.  The position of the OK button in the original About dialog uses the correct margins.

cra0zy:
> This one looks good, except I have noticed that the buttons in these Eto dialogs tend to be too close to the window borders. The position of the OK button in the original About dialog uses the correct margins.

#4751 fixes it.

cra0zy:
@KonajuGames Just FYI, I don't need to change anything in here for the Close button to gain spacing, the class that adds spacing is DialogBase and it's been modified in #4751 

What I am trying to say is this can be merged, the spacing was added to all the dialogs by merging ##4751

KonajuGames:
Ok, cool. Merging.

PR #4754, opened 2016-04-10
[Eto Dialogs] Path Dialog

cra0zy:
This dialog was and is Gtk only.

~~Now this dialog is missing a call to get relative path on: https://github.com/cra0zy/MonoGame/blob/ed6/Tools/Pipeline/Eto/Dialogs/PathDialog.cs#L54 because #4751 implements it. Even tho it's missing the call you can still merge this PR since I will be doing another PR for Eto forms right after all 3 Eto dialog PRs have been merged.~~

No screenshot this time :(, only spacing changed...


cra0zy:
Missing method has been added.

KonajuGames:
Merging.

Issue #4692, opened 2016-03-31
Broken MonoKickstart compatibility since 3.5.1


mrhelmut:
PR #4534 broke MonoKickstart compatibility which is now throwing this error, on both OS X and Linux:
> Fontconfig error: Cannot load config file

It looks like the method used to test libgdiplus triggers a font search in the Mono font cache, which doesn't exist with MonoKickstart (and only exist if the Mono SDK is installed, which is why it doesn't fail on a dev environment).

This most probably also affects projects packaged with Xamarin.

@hach-que @cra0zy

mrhelmut:
This is triggered by GDIPlus static initializer (https://github.com/mono/mono/blob/b7a308f660de8174b64697a422abfc7315d07b8c/mcs/class/System.Drawing/System.Drawing/gdipFunctions.cs), and all extern methods which could be used to avoid it are private.

Damn GDIPlus. ^^

cra0zy:
Well that was unexpected... I am sorry for not inviting you into the conversation :(

On the bright side SDL version won't be using that part of the code at all, so maybe it's not even worth fixing right now.

CC. @tomspilman @dellis1972

mrhelmut:
GDIPlus side effects, hard to predict (I couldn't have).

(Damn GDIPlus ^^)

The problem seems to be coming from Mono GDIPlus static initializer. I tried to search another method / code patch to get libgdiplus loaded without triggering the static initializer, but couldn't find any, beside ```new Icon(stream)```...

tomspilman:
> PR #4534 broke MonoKickstart compatibility which is now throwing this error, on both OS X and Linux:

So basically...   our 3.5 release is broken on the two platforms we want to feature.  Fantastic!  :(

>  so maybe it's not even worth fixing right now.

No we want to fix it...  I want to push out a 3.5.2 build as soon as possible to correct this.   If not the stable release will be broken till June.

cra0zy:
Can we modify mono kickstart to allow launching, that is include what's missing, this may be a better solution then releasing 3.5.2

tomspilman:
Oh... didn't realize we do not ship MonoKickstart as part of the Mac/Linux SDKs.  That could maybe work as long as it is a good fix.

cra0zy:
Haha, due to the face that mono got relicenced we no longer need mono kickstart, we can use mkbundle :D

tomspilman:
Is that true?  I guess I don't know how any of that works.

[...]

Issue #4743, opened 2016-04-07
[DesktopGL / OpenTK / SDL2] GamePad crahes with some joystick


mrhelmut:
Hello,

There is a bug in OpenTK SDL2 joystick driver (reported at opentk/opentk#377) which makes MonoGame to crash when some joystick are plugged. For instance, Saitek X55 (HOTAS joystick) is recognized as a gamepad but its value range are not compatible with a gamepad use and result in a crash when trying to ```GamePad.GetState(int index)```.

tomspilman:
So can we patch this into the 3.5 release?

I guess we need a fix in the OpenTK assembly?

mrhelmut:
Yes, it's an OpenTK issue, just opened an issue here to keep track of it and remember to eventually update the OpenTK assembly if it gets resolved soon enough.

cra0zy:
How about I split the SDL2 branch in two and we merge all the changes related to replacing Windowing/Input system?

mrhelmut:
Sounds interesting, these two are the biggest OpenTK concerns, and I think it's better to spend time helping you & dellis on this than trying to maintain the OpenTK path.

tomspilman:
Want to hear from @dellis1972 on that first.  Not sure if a partial removal of OpenTK is a good thing or not.

mrhelmut:
#4755 made this no longer relevant.

Issue #4643, opened 2016-03-18
[Mac] Window is being rendered at native resolution instead of retina scaled


prime31:
Window size is being reported at the retina scaled value but rendering is occurring at full resolution. See reference image below and description that follows:

<img width="752" alt="cursor-br" src="https://cloud.githubusercontent.com/assets/60377/13894613/22678d88-ed27-11e5-9040-2b957e921da4.png">

The image shows the mouse cursor at the bottom-right of the window. It is registering as 631x354. The column on the right is being rendered from MG as a 150px width box.

Now in reality, the box on the right is actually rendered as 75px of "retina pixels". It seems that the window/OpenGL view is not being properly setup. It should take the scale into account when it is created.  This causes a host of problems. All rendering is at 50% of the requested size and then the mouse coordinates are **not** properly scaled so they are 2x off.

<img width="752" alt="screen shot 2016-03-18 at 4 40 49 pm" src="https://cloud.githubusercontent.com/assets/60377/13894698/3800d770-ed28-11e5-97cf-36b3e73bd109.png">

The below image shows the mouse position and viewport size while the mouse is in the bottom right corner of the window.

tomspilman:
@mrhelmut might know something about this.

prime31:
Adding data to this. Screenshots and original report are from OpenTK implementation.

When SDL is used there are different bugs. Mouse position is always reported from the top-left of the entire display minus the client bounds making it always incorrect. On retina, the correct mouse position is SDL_position - clientBounds * 2. I have not yet figured out how to get the screen scale but it is paramount to getting the mouse data and the window size fixed.

KonajuGames:
> On retina, the correct mouse position is SDL_position - clientBounds * 2.

Is the root cause here that client bounds is incorrect on a Retina display?  I really don't want to hack around with "if retina multiply by 2" in lots of places.  Need to fix the root cause, not the symptoms.

prime31:
The root cause here is that client bounds is in pixels instead of points. This is easily remedied with SDL2 via [this commit](https://github.com/prime31/MonoGame/commit/e80a1f2b7b6fad4cbc50f17154381aad396063af). I do not know what OpenTK is doing when SDL2 isnt being used yet. I have no experience with OpenTK so figuring out what is happening in there is slow going.

KonajuGames:
We have an in-progress branch that removes OpenTK and just uses SDL2.  This will remove a whole range of issues we are currently experiencing.

[...]

Issue #4529, opened 2016-02-12
[SDL2/OpenTK] Mouse blocked after button down on OS X


mrhelmut:
Hello,

I'm encountering some heavy mouse stuttering on OS X.

Whenever a mouse click (only on button down) is made, SDL2 doesn't return any mouse movement events for about 500ms. There's really no movement at all, being through ```SDL_PollEvents()``` (no event triggered) or ```SDL_GetMouseState()``` (coordinates haven't changed despite movements).

All other events are fine, it's only mouse movement which are "blocked" whenever a click is made.

I thought that it was linked to some trackpad palm detection, but it happens on all Mac with any kind of mouse (and trackpads are not affected by this issue).

@dellis1972 @cra0zy, would you happen to have any clues?

dellis1972:
@mrhelmut is this using OpenTK's SDL2 back end?

mrhelmut:
Yes.

dellis1972:
I've not seen that. it sounds like something is locking the UI thread though.
I don't use SDL2 with OpenTK anymore on Mac. It seems to work ok without it since we forked OpenTK and fixed a bunch of stuff.

mrhelmut:
Gamepads don't work with the native backend on both OS X & Linux in our situation, SDL2 is the most stable backend across all OS for our game.

I'll take a look at the OpenTK window's events to see if anything could block the UI thread.

mrhelmut:
UI thread doesn't get blocked, event polling is still executing, it just doesn't return any mouse move events for ~500ms after a click.

dellis1972:
@mrhelmut sounds like an SDL2 bug.

How come the native stuff doesn't work? seems to work ok for me.

cra0zy:
No idea what is causing this. I was not able to reproduce it on Linux

mrhelmut:
It seems to be OS X only, I don't have it on Windows too.

I have several problems with the native backends:
- Windows: keyboard recognized as 2 gamepads
- OS X: gamepad not recognized, retina not working correctly, mouse position always 0 (may be retina related)
- Linux: gamepad not recognized, fullscreen issues

SDL2 works fine everywhere for me, minus this current  issue.

cra0zy:
> Linux: fullscreen issues

I forgot about that... I had a fix for it for so long... submitted in #4530

dellis1972:
@cra0zy merged :P

mrhelmut:
Fixed by #4755

PR #4755, opened 2016-04-10
SDL2 Mergable Stuff

cra0zy:
These are commits that can be merged into develop branch without damaging anything. They only replace OpenTK for Windowing/Input, leaving OpenAL/OpenGL bindings to OpenTK.


tomspilman:
Want @dellis1972 to ok this before we merge.

mrhelmut:
I'm going to test this soon. Looks great, just would like to make a couple checks on GamePad & Retina support (I mean, test for regression).

MonoGame.Framework/SDL/SDLGamePlatform.cs
```diff
@@ -0,0 +1,212 @@
...
+            Sdl.GetVersion(out sversion);
+
+            Sdl.Major = sversion.Major;
+            Sdl.Minor = sversion.Minor;
+            Sdl.Patch = sversion.Patch;
+
+            var version = 100 * Sdl.Major + 10 * Sdl.Minor + Sdl.Patch;
+
+            if (version <= 204)
+                Debug.WriteLine ("Please use SDL 2.0.5 or higher.");
```
mrhelmut: Is 2.0.5 a typo or does it mean that 2.0.5 (unstable) is the target? (because ```version <= 204``` allows 2.0.4).

MonoGame.Framework/SDL/SDL2.cs
```diff
@@ -0,0 +1,717 @@
...
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using System.Text;
+using System.Runtime.InteropServices;
+
+internal static class Sdl
+{
+    private const string NativeLibName = "SDL2.dll";
```
mrhelmut: Should SDL2.dll be included in the DesktopGL template? (Same question for openal32.dll)

I know that it makes little sense because of the cross-platform nature of DesktopGL (and that it wouldn't solve OS X / Linux dependencies), but I'm wondering this to avoid misleading errors out-of-box on Windows.

@tomspilman ?

mrhelmut:
I'm running into this on Windows (on a fresh rebuild):

```
System.AccessViolationException (Attempted to read or write protected memory. This is often an indication that other memory is corrupt.)

   at Microsoft.Xna.Framework.GamePlatform.set_IsActive(Boolean value) in d:\Projects\MonoGame\MonoGame.Framework\GamePlatform.cs:line 65
   at Microsoft.Xna.Framework.SdlGamePlatform.RunLoop() in d:\Projects\MonoGame\MonoGame.Framework\SDL\SDLGamePlatform.cs:line 129
   at Microsoft.Xna.Framework.Game.Run(GameRunBehavior runBehavior) in d:\Projects\MonoGame\MonoGame.Framework\Game.cs:line 397
   at Microsoft.Xna.Framework.Game.Run() in d:\Projects\MonoGame\MonoGame.Framework\Game.cs:line 370
```

cra0zy:
Last time I got that error I was using 32 bit SDL on a 64 bit machine and exe. Make sure you are using the correct SDL.

MonoGame.Framework/SDL/SDL2.cs
cra0zy: There are 32 and 64 bit SDL, I don't see how to include it even if we wanted.

MonoGame.Framework/SDL/SDLGamePlatform.cs
cra0zy: I put in some legacy code to make it work on lower versions of SDL.

ie. Mouse won't track global pos <= 2.0.4 because there is a bug in there

mrhelmut:
Hmm, turns out it only runs with the 64bit SDL2 now, even if I set x86 in project settings...

GamePads are not detected on Windows, I'll take a closer look at this later.

tomspilman:
@dellis1972 - Did you look at this?  Do you think it is worth merging this before the GL/AL work is done?

dellis1972:
I've not had a chance to test it. but it looks about right. I don't see the
harm in getting this in while the GL stuff is on going as long as the
branch we are working on it kept up to date

MonoGame.Framework/SDL/SDL2.cs
tomspilman: > Should SDL2.dll be included in the DesktopGL template?

Well... at the very least when you build the SDL2.dll should appear in your output bin folder.  It shouldn't be up to the user to hunt down the SDL2.dll for themselves.

We could add some custom scripting into our current `MonoGame.Content.Builder.targets` that detects if we're building for 32bit or 64bit and ensures the right SDL2 assembly is added to the build output.

This make sense?  @dellis1972 @KonajuGames ?

tomspilman:
Thanks @dellis1972 !

MonoGame.Framework/SDL/SDL2.cs
dellis1972: the .targets fixup sounds like a great idea.  

We should also include the .dll in the nuget for those people using it.

mrhelmut:
I'm going to take some time tomorrow to test it more to see what's going on with the gamepads on Windows and the 32bit support.

cra0zy:
> I'm going to take some time tomorrow to test it more to see what's going on with the gamepads 

Just want to mention that if one of your gamepads doesn't work (and there is bound to be at least one), don't complain about it here but instead wait for this to get merged, and then submit a PR that adds the gamepad config to the MonoGame.Framework/SDL/gamecontrollerdb.txt

mrhelmut:
Actually, there is no gamepad at all, not even XInput 360 gamepads.
But sure, if there is any minor issue I'll wait until it gets merged. Just want to make sure there is no regression, maybe it's just something I didn't get correctly yet, like the 32/64bit dll.

cra0zy:
> Windows and the 32bit support.

Try running it from a 32 bit OS, you can't really put a 32 bit SDL on a 64 bit OS because even SDL depends on some other stuff, right?

MonoGame.Framework/SDL/SDL2.cs
cra0zy: > the .targets fixup sounds like a great idea.

Agreed.

Can one of you modify the targets file? I ask because as much as I would love to modify it, I don't have time to learn how :|

tomspilman: > Can one of you modify the targets file? 

If you can give me a day or two i can...  just need to figure out how to detect 64bit vs 32bit.

Maybe this shouldn't hold up this PR however?

@dellis1972 @KonajuGames - Do you think this can be merged for now without the DesktopGL template adding the correct SDL2 assembly for the user?

mrhelmut: You might want to do the same for OpenAL-Soft 32/64bit dlls at the same time.

tomspilman: True... we know that is coming soon.

KonajuGames: If the dev is building for AnyCPU, how would the detection work? The game
could be run on either architecture.  This is where Mono actually has the
upper hand over Microsoft in that the dll.config file would handle this
runtime selection of library based on the user's architecture. Microsoft
don't support that.
On 15 Apr 2016 08:35, "Tom Spilman" <notifications@github.com> wrote:

In MonoGame.Framework/SDL/SDL2.cs
<https://github.com/mono/MonoGame/pull/4755#discussion_r59803592>:

> @@ -0,0 +1,717 @@
> +﻿// MonoGame - Copyright (C) The MonoGame Team
> +// This file is subject to the terms and conditions defined in
> +// file 'LICENSE.txt', which is part of this source code package.
> +
> +using System;
> +using System.Text;
> +using System.Runtime.InteropServices;
> +
> +internal static class Sdl
> +{
> +    private const string NativeLibName = "SDL2.dll";

True... we know that is coming soon.

—
You are receiving this because you were mentioned.

Reply to this email directly or view it on GitHub
<https://github.com/mono/MonoGame/pull/4755/files/b02e831910f420299c3838712a3aed80f51ed1c0#r59803592>

cra0zy: I found a good solution for deciding to load 32 or 64 bit dll. We can just keep the dlls in 2 folders, and then call: https://msdn.microsoft.com/en-us/library/ms686203.aspx to decide which folder to use.

mrhelmut: Maybe overriding the PATH environment variable would be fine?

```
            if (Environment.Is64BitProcess)
                Environment.SetEnvironmentVariable("PATH", Environment.GetEnvironmentVariable("PATH") + ";" + AppDomain.CurrentDomain.BaseDirectory + "\x64");
            else
                Environment.SetEnvironmentVariable("PATH", Environment.GetEnvironmentVariable("PATH") + ";" + AppDomain.CurrentDomain.BaseDirectory + "\x86");
```

KonajuGames: > I found a good solution for deciding to load 32 or 64 bit dll.

I like that idea.  Keeps it simple.

> Maybe overriding the PATH environment variable would be fine?

That could be a backup.  The only change I would make would be to put the app path first before the existing PATH value.

Build/Projects/MonoGame.Framework.definition
```diff
@@ -155,10 +155,15 @@
...
     </None>
     <None Include="..\ThirdParty\Dependencies\OpenTK.dll.config">
-      <Platforms>WindowsGL,Linux,MacOS</Platforms>
+      <Platforms>MacOS, Linux</Platforms>
       <Link>OpenTK.dll.config</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="SDL\MonoGame.Framework.dll.config">
+      <Platforms>Linux</Platforms>
+      <Link>MonoGame.Framework.config</Link>
```
mrhelmut: Should be MonoGame.Framework.dll.config

theZMan:
Using PATH is affecting a global system wide setting what about the edge case of someone running a 32 bit and a 64 bit mono game app - will they conflict?

KonajuGames:
> Using PATH is affecting a global system wide setting what about the edge case of someone running a 32 bit and a 64 bit mono game app - will they conflict?

Environment variables are inherited from the process that launched the current process.  Any changes to the environment variables by the current process are local to the process.

mrhelmut:
Here's a review, there are a few issues but this is mostly a huge improvement, kudos @cra0zy !

The good:
- goodbye windowing issues!
- fixes #4692 #4529 #4412 #4431
- fixes the non-XNA compliant keyboard mapping when dealing with non-qwerty keyboards
- uses the universal SDL game controller API, which is WAY more reliable than the OpenTK backend mess
- should be fixing #4743, but I have no HOTAS joystick to confirm

The "could be better":
- I have graphics device crashes when disposing render targets, I still have to figure that out
- Gamepads are not initialized before the run loop, which means that you can't query ```IsConnected``` in ```Initialize```
- High-DPI somehow works on both OS X & Windows 10 (way better than OpenTK!), but the Viewport seems wrongly initialized before ```Initiliaze``` when using fullscreen, and #4643 doesn't seem fixed. Still way more usable than the current OpenTK path! (and this could easily be addressed after this PR)

:+1:

cra0zy:
> Gamepads are not initialized before the run loop, which means that you can't query IsConnected in Initialize

This is because we need to wait for a JoystickConnected event before gamepads can be used which happens in Run loop.

> High-DPI somehow works on both OS X & Windows 10 (way better than OpenTK!), but the Viewport seems wrongly initialized before Initiliaze when using fullscreen, and #4643 doesn't seem fixed. Still way more usable than the current OpenTK path! (and this could easily be addressed after this PR)

I don't have a High-DPI monitor, so I can't test it and I didn't want to code blindly, but the solution should be to use `SDL_WINDOW_ALLOW_HIGHDPI` https://wiki.libsdl.org/SDL_WindowFlags

Build/Projects/MonoGame.Framework.definition
cra0zy: Good point, will modify it, the file is not really used because there is OpenTK.dll.config which overrides SDL as well.

tomspilman:
> This is because we need to wait for a JoystickConnected event before gamepads can be used which happens in Run loop.

We need to find a way to fix that.  It is an incompatibility with XNA as i just tested and this works:
```
        static void Main(string[] args)
        {
            var state = GamePad.GetState(PlayerIndex.One);
            if (state.IsConnected)
                Console.WriteLine("GamePad IsConnected!");

            using (Game1 game = new Game1())
                game.Run();
        }
```

cra0zy:
I still don't have time... and for what you are asking above, if you want me to set it up, I will need to create a separate thread for running SDL run loop, there is no way around it.

dellis1972:
@cra0zy can we pump the SDL message queue a few times before entering the run loop? 

@CartBlanche can you remember from your SDL days how we got round this issue?

cra0zy:
> @cra0zy can we pump the SDL message queue a few times before entering the run loop?

If a gamepad gets disconnected between init and run loop it will stay connected forever and ever and ever and ever...  tho initializing SDL run loop once before Initalize will fix it, but will not allow you to use GamePad class without Game.Run later on

tomspilman:
Well...  none of that I think keeps us from merging this.  We can deal with these holes in subsequent PRs.  If this seems ok to everyone then lets merge.

cra0zy:
So I made the SdlRunLoop activate once before Initialize so we stuff is initialized properly during Initialize.

cra0zy:
Small question for everyone about XNA, if you pull out gamepad that's on PlayerIndex.One, should the gamepad on PlayerIndex.Two go to PlayerIndex.One? I only have one Xbox controller so can't test it.

tomspilman:
> So I made the SdlRunLoop activate once before Initialize so we stuff is initialized properly during Initialize.

That is great... seems like a reasonable solution.

>  if you pull out gamepad that's on PlayerIndex.One, should the gamepad on PlayerIndex.Two go to PlayerIndex.One? 

Good question...  let me test that here really quick:

```
            Console.WriteLine("PlayerIndex.One IsConnected=" + GamePad.GetState(PlayerIndex.One).IsConnected);
            Console.WriteLine("PlayerIndex.Two IsConnected=" + GamePad.GetState(PlayerIndex.Two).IsConnected);
```
With this test if I unplug the first gamepad it prints:

```
PlayerIndex.One IsConnected=False
PlayerIndex.Two IsConnected=True
```
So...  no.

Also if I start with only one GamePad connected that one is always `PlayerIndex.One`.

So basically when the gamepad is fist connected it is assigned to the first free `PlayerIndex`.  On disconnect it just frees the slot for that one player.

tomspilman:
This ok to merge then @cra0zy ?

cra0zy:
> This ok to merge then @cra0zy ?

Ok to merge. 

I am leaving adding SDL to templates for another PR, because besides SDL we might also want to add OpenAL, plus I am gonna compile a special version of SDL for Linux from newest devel, plus patch for resizable window.

flibitijibibo:
I've written a patch to SDL that should allow for AllowUserResizing to work at runtime:

http://flibitijibibo.com/resizable.diff

If anyone wants to test this it'll help get this into mainline sooner, will aim to have this in 2.0.5.

cra0zy:
@flibitijibibo Thanks for the patch, I'll test it out later, also you could attach it to something like: https://bugzilla.libsdl.org/show_bug.cgi?id=3288

On a side note, https://bugzilla.libsdl.org/show_bug.cgi?id=3272 is also very important.

cra0zy:
@tomspilman @KonajuGames Not to be impatient, but can this get merged before something else brakes it again?

mrhelmut:
Looks good to me.

tomspilman:
Thanks guys... merging!

cra0zy:
@flibitijibibo Patch is not mergable with SDL from merculiar :*|, did you maybe intend it for SDL 2.0.4?

flibitijibibo:
The patch still applies cleanly for me:

```
hg clone http://hg.libsdl.org/SDL/
cd SDL
curl -O flibitijibibo.com/resizable.diff
hg import resizable.diff
```

cra0zy:
Does your patch by any chance conflict with: https://bugzilla-attachments.libsdl.org/attachment.cgi?id=2390

flibitijibibo:
Looks like that patch touches other files, so it should still apply cleanly. For X11 SDL_SetWindowResizable only touched video/windows, the position fix just touches events and adds a symbol to the function table.

cra0zy:
My copy of SDL from hg must have been broken, it's working fine now.

hach-que:
This change has broken MacOS.  There are two implementations of GamePad and Joystick while on MacOS now:

```
    <Compile Include="Input\GamePad.Default.cs">
      <Platforms>Android,MacOS,WindowsPhone,WindowsPhone81</Platforms>
    </Compile>
```

and

```
    <Compile Include="Input\GamePad.MacOS.cs">
      <Platforms>MacOS</Platforms>
    </Compile>
```

You need to wrap a:

```
#if !MACOS || PLATFORM_MACOS_LEGACY
```

around the code in the default implementation.

hach-que:
@tomspilman @cra0zy (see above)

cra0zy:
It makes more sense to just remove MacOS platform from GamePad.Default.cs or GamePad.MacOS,cs...

hach-que:
The MacOS version doesn't have an implementation because the new Xamarin
APIs don't support gamepads or joysticks. So you can't use MacOS for
MonoMac without breaking people's code.

That said if default is running it via SDL then you can probably just use
default. I didn't get any compilation errors relating to the default
implementation when building on Xamarin so it's probably safe.

cra0zy:
.Default is not using SDL, .SDL is for SDL.

hach-que:
Then in that case you'll need to add the conditionals back.  OpenTK on Xamarin.Mac doesn't have the Gamepad or Joystick APIs.

cra0zy:
Please look at .Default.cs and .Mac.cs files since you have no idea what you are talking about... all that needs to be done is disconnect one of them.

hach-que:
So Default just pretends as though there is one controller with no state, while Mac (correctly) returns 0 controllers?

I'm not sure the default implementation is correct.

hach-que:
The other impact of this change, if I'm understanding it correctly, is that anyone using MonoGame on MacOS under MonoMac/XamMac has now lost the ability to read game pads and joysticks?  That seems like a regression to me (Xamarin.Mac never supported it, but the older APIs did).

cra0zy:
MonoMac support is getting dropped, Xamarin.Mac support is up for debate, no regression has occured.

hach-que:
Well we'll need Xamarin.Mac support regardless of what happens.  But if dropping MonoMac/XamMac is intended for 3.6 release, then we can start stripping out all of the MonoMac specific code as well.

hach-que:
(Well, code that is in a `#if PLATFORM_MACOS_LEGACY` block anyway, the Mac code in general still needs to stay around)

PR #4757, opened 2016-04-10
[Eto Forms] Partial Main Window

cra0zy:
Moved some common stuff to Controller, and moved some MainWIndow common stuff to a partial MainWindow

PS. Next PR will fix Dialog positioning.

### DIFF
--- a/Build/Projects/Pipeline.definition
+++ b/Build/Projects/Pipeline.definition
@@ -117,6 +117,7 @@
       <DependentUpon>Global.cs</DependentUpon>
       <Platforms>MacOS</Platforms>
     </Compile>
+	<Compile Include="Eto\MainWindow.cs" />
 	<Compile Include="Eto\Styles.Linux.cs">
       <Platforms>Linux</Platforms>
     </Compile>
--- a/Tools/Pipeline/Common/IController.cs
+++ b/Tools/Pipeline/Common/IController.cs
@@ -104,15 +104,15 @@ namespace MonoGame.Tools.Pipeline
 
         void DragDrop(string initialDirectory, string[] folders, string[] files);
 
-        void Include(string initialDirectory);
+        void Include();
 
-        void IncludeFolder(string initialDirectory);
+        void IncludeFolder();
 
         void Exclude(IEnumerable<ContentItem> items, IEnumerable<string> folders, bool delete);
 
-        void NewItem(string name, string location, ContentItemTemplate template);
+        void NewItem();
 
-        void NewFolder(string name, string location);
+        void NewFolder();
 
         void Move (string[] paths, string[] newpaths, FileType[] types);
         
--- a/Tools/Pipeline/Common/IView.cs
+++ b/Tools/Pipeline/Common/IView.cs
@@ -39,6 +39,8 @@ namespace MonoGame.Tools.Pipeline
 
         bool ShowDeleteDialog(string[] folders, string[] files);
 
+        bool ShowEditDialog(string title, string text, string oldname, bool file, out string newname);
+
         void BeginTreeUpdate();
 
         void SetTreeRoot(IProjectItem item);
@@ -63,7 +65,9 @@ namespace MonoGame.Tools.Pipeline
 
         bool ChooseContentFile(string initialDirectory, out List<string> files);  
 
-        bool ChooseContentFolder(string initialDirectory, out string folder);        
+        bool ChooseContentFolder(string initialDirectory, out string folder);
+
+        bool ChooseItemTemplate(string folder, out ContentItemTemplate template, out string name);
 
         bool CopyOrLinkFile(string file, bool exists, out CopyAction action, out bool applyforall);
 
@@ -72,5 +76,9 @@ namespace MonoGame.Tools.Pipeline
         Process CreateProcess(string exe, string commands);
 
         void ItemExistanceChanged(IProjectItem item);
+
+        bool GetSelection(out FileType fileType, out string path, out string location);
+
+        bool GetSelection(out FileType[] fileType, out string[] path, out string[] location);
     }
 }
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -565,8 +565,12 @@ namespace MonoGame.Tools.Pipeline
             Include(initialDirectory, files);
         }
 
-        public void Include(string initialDirectory)
+        public void Include()
         {
+            FileType type;
+            string path, initialDirectory;
+            View.GetSelection(out type, out path, out initialDirectory);
+
             // Root the path to the project.
             if (!Path.IsPathRooted(initialDirectory))
                 initialDirectory = Path.Combine(_project.Location, initialDirectory);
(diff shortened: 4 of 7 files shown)


tomspilman:
Great... going to merge this.

After we get done with the Eto refactor we really need to get to https://github.com/mono/MonoGame/issues/4558.

cra0zy:
I wanted to chop up the replacement of MainWindow into smaller pieces, but it seems that I am gonna have to do it in one swoop.

Sorry about the next PR, it will be a bit bigger.

tomspilman:
Yeah... I could see how that would be hard to break up.  Do what you can.

cra0zy:
May I use Eto.Wpf instead of Eto.WinForms? Pipeline Tool looks much nicer IMO...

tomspilman:
> May I use Eto.Wpf instead of Eto.WinForms? Pipeline Tool looks much nicer IMO...

Maybe... I would want to see how it looks and test it on some older OSs and lower end systems.

cra0zy:
> I would want to see how it looks and test it on some older OSs and lower end systems.

I haven't really applied anything special to it so for now it looks ordinary:
untitled
untitled2
There are few kinks I need to workout, but nothing too major.

But here is the fun thing about it, we should be able to apply something like VS style, or Metro style... I got the idea from Gemini (https://github.com/tgjones/gemini).

tomspilman:
> I haven't really applied anything special to it so for now it looks ordinary:

Ordinary is fine...  we want it to look like any other application on Windows.

What bothers me is the incorrect elements in that image:

  - The non-working property grid.
  - No sizing border on the window.
  - Tree view missing lines.
  - The unnecessary black sizer grip on the top of both windows.
  - The overlapping corner sizer grip with the cancel button.

Is this because of bugs in Eto or this is all stuff Eto doesn't handle?

> we should be able to apply something like VS style, or Metro style

 I would be hesitant to do that by default... it serves no functional purpose.

cra0zy:
> Is this because of bugs in Eto or this is all stuff Eto doesn't handle?

> The non-working property grid.

I didn't implement it yet, I'll reuse the code from the Gtk one. I also didn't set border style for any controls in the main window so they have black outline. 

> No sizing border on the window.

The window is resizable from any border, it is the same border as the current WinForms Pipeline...

> Tree view missing lines.

Visual Studio is using the same component, and doesn't have any treeview lines either so that's not really a complaint... that's normal.

PS. I am using TreeGridView instead of TreeView, reason is selection rectangle.

> The unnecessary black sizer grip on the top of both windows.

Not a sizer grip, it's something VS adds for debugging to WPF windows while in debug mode.

> The overlapping corner sizer grip with the cancel button.

That is a bug in Eto, I submitted a bug report (https://github.com/picoe/Eto/issues/491), tho it's easily fixable by adding a style.

Haha, when I look at it only the last thing was an actual complaint...

tomspilman:
> I didn't implement it yet, I'll reuse the code from the Gtk one.

That is going to be my ultimate issue with it...  things like the property grid where we can't use the ones that exist for WinForms or WPF... we have to build our own "custom" one.

I'm already disliking this approach.

> The window is resizable from any border,

Resizable borders in Windows have a thick edge:

border

Sometimes it includes the corner gripper... usually depends on if the window has a status bar or is a dialog or not.

> Visual Studio is using the same component, and doesn't have any treeview lines either so that's not really a complaint

As long as this is an actual native Windows tree view and not something created to mimic it we could live with that.

> it's something VS adds for debugging to WPF windows while in debug mode

Weird... It didn't do that last I used WPF 4 years ago.

cra0zy:
> That is going to be my ultimate issue with it... things like the property grid where we can't use the ones that exist for WinForms or WPF... we have to build our own "custom" one.
> 
> I'm already disliking this approach.

Windows Forms is the only GUI that offers Property Grid out of the box.

WPF, Gtk Qt, Cocoa, UWP/Metro all don't have a property grid out of the box and need a manual implementation.

> Resizable borders in Windows have a thick edge:
> 
> 
> 
> Sometimes it includes the corner gripper... usually depends on if the window has a status bar or is a dialog or not.
> 
> Visual Studio i

This is how 3.5 Pipeline Tool looks for me:

untitled

Completely the same border :|

Also the border color comes from the system theme.

Tzenchor:
> Resizable borders in Windows have a thick edge

That is the normal border in Windows 10, at least for me

tomspilman:
> Windows Forms is the only GUI that offers Property Grid out of the box.

 You're right... I forgot about that issue in WPF.  There are various replacements that mimic WinForms/VisualStudio style property grids.

We should specifically aim to mimic that as well then.

> This is how 3.5 Pipeline Tool looks for me:

That may be a Windows 10 thing then.  As long as it renders correctly on non-Windows 10 systems then we're doing it right.

> Also the border color comes from the system theme.

Of course...  I've been writing Windows GUIs since 1994.  I fully understand how it works.

cra0zy:
I figured out the border thing, maybe... I didn't really let my system update that much (I am mostly on Linux) so I still haven't gotten to installing the November update.

Tzenchor:
> I figured out the border thing, maybe... I didn't really let my system update that much (I am mostly on Linux) so I still haven't gotten to installing the November update.

For me the border thing didn't change since release, the november update only affected the title bar, as now it shows the system color instead of being white like in your screenshots.

And for your information, look for "Snipping Tool" while using Windows, it's a great help when taking screenshots

tomspilman:
> look for "Snipping Tool" while using Windows

Yep... snipping tool is great!  I just wish I could drag and drop the image right to GitHub without needing to save it first.

cra0zy:
> And for your information, look for "Snipping Tool" while using Windows, it's a great help when taking screenshots

Ha, interesting, didn't know it was there, thanks for mentioning it. On Linux with GNOME there is a tool that does the same called "Screenshot".

One interesting difference between Windows and Linux is that print screen on Linux automatically saves the image.

Tzenchor:
> One interesting difference between Windows and Linux is that print screen on Linux automatically saves the image.

Not anymore, `Windows + Print Screen`, or `Windows + Volume Down` on tablets, to take the whole desktop and it is saved to `Pictures\Screenshoots`.

KonajuGames:
A good guide to button positioning is that the resizing handle in the bottom right corner shouldn't overlap the button (see the New File dialog in your screenshot, even though that dialog doesn't need to be resizeable).

cra0zy:
> A good guide to button positioning is that the resizing handle in the bottom right corner shouldn't overlap the button (see the New File dialog in your screenshot, even though that dialog doesn't need to be resizeable).

As I said above, resize handle is a bug in Eto, it shouldn't be visible at all.

cra0zy:
Some progress report on Eto Forms Main WIndow:
 - Window, Menubar, Toolbar, ContextMenu, HPanned/VPanned have been fully setup and are fully using Eto.Forms
 - ProjectView, PropertiesView and FilterView are wrappers around native controls and I am working on this part right now
 - Windows version is sticking with WinForms so I can reuse the code from the existing Pipeline Tool for the  controls above

cra0zy:
UPDATE NO. 2
 - I skipped over the normal treeview control, but it's done much better than the treegridview, so I got ProjectView and FilterView in Eto.Forms now
  - multiselect is gonna be a pain to do for Windows :(
 - currently looking at a way to render a custom properties view

cra0zy:
New porperty grid fully written in Eto is looking good: https://youtu.be/nZeqn9Km9eI

PR #4758, opened 2016-04-11
Fix ouya template

### LINKED ISSUE #4722: Automatic building of Content fails in monogame Ouya project (Visual Studio 2015)
I am developing on a 64-Bit Windows 10 Home machine in Visual Studio 2015 and I get the following error when I try to build my Ouya test game.

`The command ""C:\Program Files (x86)\MSBuild\MonoGame\v3.0\Tools\MGCB.exe" 
/@:"C:\OuyaTest\Content\Content.mgcb" 
/platform:Ouya 
/outputDir:"C:\OuyaTest\Content\bin\Ouya" 
/intermediateDir:"C:\OuyaTest\Content\obj\Ouya" 
/quiet" exited with code -1.`

I think that the problem is that there is no 'Ouya' platform in the content pipeline to build with.
When I open the content file in the content pipeline and try to select a platform only these show up:
* Android
* DesktopGL
* iOS
* MacOSX
* NativeClient
* Playstation4
* PlaystationMobile
* RaspberryPi
* Windows
* WindowsPhone
* WindowsPhone8
* WindowsStoreApp
* Xbox360

As you can see there is no 'Ouya' platform and the command does try to use it.
Is there maybe a way to alter the command that gets run so that I could build for the 'Android' platform instead of the 'Ouya' one.

Tzenchor:
May fix #4722, but can't guarantee it, when trying to compile throught Visual Studio is giving 2 errors because I don't have a "business license".

### DIFF
--- a/ProjectTemplates/VisualStudio2010/OUYA/Content/Content.mgcb
+++ b/ProjectTemplates/VisualStudio2010/OUYA/Content/Content.mgcb
@@ -3,7 +3,7 @@
 
 /outputDir:bin/Ouya
 /intermediateDir:obj/Ouya
-/platform:Ouya
+/platform:Android
 /config:
 /profile:Reach
 /compress:False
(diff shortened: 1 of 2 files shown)


tomspilman:
@KonajuGames - This seems correct... we removed or never had OUYA as a content platform right?

KonajuGames:
Correct. It's just Android content. Merging.

PR #4763, opened 2016-04-13
Red/Blue swap implemented in DDS loader

KonajuGames:
The test for a RB swap was incorrect for 32-bit colour textures.
Implemented the RB swap for when it was needed.
Added unit test for a full colour cubemap DDS that needs a RB swap.


KonajuGames:
Fixes http://community.monogame.net/t/awful-drawings/7335/7

tomspilman:
Looks good... merging.

Do you think this solves the original poster's issue too?

KonajuGames:
This one dealt exclusively with DDS files. The original post was with PNG
files.  I'll confirm their behaviour is correct tonight.

PR #4765, opened 2016-04-14
Revert "Prevent finalizer exception on systems without libgdiplus"

### LINKED ISSUE #4692: Broken MonoKickstart compatibility since 3.5.1

mrhelmut:
This reverts #4534 to make DesktopGL compatible with Monokickstart again. It is temporary and a real fix to this situation will be provided by #4755 (i.e. removing OpenTK / System.Drawing dependencies).

Fixes #4692

### DIFF
--- a/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
+++ b/MonoGame.Framework/Desktop/OpenTKGameWindow.cs
@@ -390,26 +390,6 @@ namespace Microsoft.Xna.Framework
             // try to load it using Mono, it will cause a crash because of this.
             try
             {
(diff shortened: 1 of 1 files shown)


KonajuGames:
Merging.  Thanks @mrhelmut.

Issue #4766, opened 2016-04-14
Make SpriteEffect public


Jjagg:
I'm doing some stuff for which I need to render sprites with custom vertices. Right now I'm using BasicEffect, but it'd be nice if I could use SpriteEffect since it has everything I need. Is there any good reason for it not being public other than that's the way it is in XNA?

prime31:
Why don't you just make your own shader and process the verts in their? I don't really see how SpriteEffect being public would allow you to do that. It doesnt actually do anything besides set the projection matrix.

Jjagg:
I don't need to process the verts in any special way, just drawing primitives (that's what i mean by custom verts, not just a texture). So sprite effect was sufficient to do the rendering and it got me wondering why it isn't public since the other built-in effects are. IMO it could be nice when drawing some basic 2d textured primitives to have the shader built-in, so you don't have to write your own.

tomspilman:
I see no issue with doing this.  @KonajuGames ?

KonajuGames:
I've got no issue with making it public for those who want use it.

PR #4776, opened 2016-04-18
Make SpriteEffect public

Jjagg:
This makes SpriteEffect public like the other built-in effects as discussed in #4766.


KonajuGames:
Thanks @Jjagg. Merging.

PR #4777, opened 2016-04-18
Allow Bgra4444 textures to be created in DirectX desktop

KonajuGames:
Fixes http://community.monogame.net/t/reading-c-ushort-in-hlsl-float-is-wrong-in-monogame-64-bit-compilation/7443

- DirectX 11.2 re-added Bgra4444 texture format after it was removed in DirectX 10 (Bgra5551 and Bgr565 were re-added in DirectX 11.1). SharpDX for Windows does not appear to have the enum value for it.  Added the special case for this so it didn't create a Color format texture by default.

Other issues found during the investigation of the above fix.

- Found a resource leak when using Texture2D.FromStream() in DirectX.  The texture created in PlatformConstruct() was overwritten by the texture created in PlatformFromStream().  Texture2D instances constructed normally already had the texture created in the call to PlatformSetData() if required.
- Removed a redundant creation of DataBox in PlatformSetData(). UpdateSubresource has an override to take the separate parameters that was called internally by the DataBox override anyway.
- Texture2D.CreateTex2DFromBitmap() should not need to be public since it is only used internally.

### DIFF
--- a/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
@@ -33,9 +33,6 @@ namespace Microsoft.Xna.Framework.Graphics
 
             _renderTarget = (type == SurfaceType.RenderTarget);
             _mipmap = mipmap;
-
-            // Create texture
-            GetTexture();
         }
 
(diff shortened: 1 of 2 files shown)


tomspilman:
Looks good... merging!

Issue #4729, opened 2016-04-05
Add a way to ckeck modifier keys


cra0zy:
It would be useful if we provided some way of checking caps/scroll/num lock status.

tomspilman:
This is how you normally do modifier keys:
```
var ctrlDown = state.IsKeyDown(Keys.LeftControl) || state.IsKeyDown(Keys.RightControl);
if(ctrlDown && state.IsKeyDown(Keys.C))
{
}
```
What is it about this technique that doesn't work?

tomspilman:
Or maybe you mean the toggled state specifically of `Keys.CapsLock`, `Keys.Scroll`, and `Keys.NumLock`?

cra0zy:
> Or maybe you mean the toggled state specifically of Keys.CapsLock, Keys.Scroll, and Keys.NumLock?

Yup.

Tzenchor:
Shouldn't this be closed?

cra0zy:
@Tzenchor You are right.

Issue #4744, opened 2016-04-07
Add Unit Tests for Input Classes


tomspilman:
We need to add some unit tests which cover the following classes:
- GamePad
- GamePadButtons
- GamePadDpad
- GamePadState
- GamePadThumbSticks
- GamePadTriggers
- Joystick
- Keyboard
- KeyboardState
- Mouse
- MouseState

Here is the coverage report as of this time:

![input](https://cloud.githubusercontent.com/assets/103522/14360776/90f50d20-fcbd-11e5-9f67-211c494f11b5.JPG)

For some of these there isn't much more to do that calling the methods to ensure they don't just outright crash, verifying default values, and method argument limits.

WaveFunctionP:
I am attempting to work on this. (Haven't done much unit testing before.)

Would you prefer separate commits for each class, or in one fell swoop?

KonajuGames:
Take on one type of input at first, say Mouse or Keyboard.  Make a group of tests for that input type and see how big the PR would be.

siudeks:
Q1: What is the status of the issue? in Monogame sources I see GamePadTest, JoystickTest, KeyboardTest, MouseTest so probably code coverage is already increased bu I don't know how much.. 
Q2: could you provide link to the current Monogame code coverage in Teamcity pls?

Jjagg:
I implemented this about as much as we can test it without actual hardware devices in #5142.

siudeks:
:) sounds good, so @Jjagg  could you close the issue? Seems to be already implemented.

Jjagg:
@KonajuGames or @tomspilman?

cra0zy:
@tomspilman This has been implemented? : https://github.com/MonoGame/MonoGame/tree/develop/Test/Framework/Input

tomspilman:
Yes it seems to be good enough... closing.

PR #4789, opened 2016-04-21
[MG Framework] Implement modifier keys API

cra0zy:
This adds an API that exposes CapsLock / NumLock states to the end user (#4729). I didn't add scroll lock because SDL doesn't expose it and looking into how to add it seemed pointless... when was the last time someone even used it?

### DIFF
--- a/MonoGame.Framework/Input/Keyboard.SDL.cs
+++ b/MonoGame.Framework/Input/Keyboard.SDL.cs
@@ -1,0 +1,26 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Xna.Framework.Input
+{
+    public static partial class Keyboard
+    {
+        static List<Keys> _keys;
+
+        private static KeyboardState PlatformGetState()
+        {
+            var modifiers = Sdl.Keyboard.GetModState();
+            return new KeyboardState(_keys,
+                                     (modifiers & Sdl.Keyboard.Keymod.CapsLock) == Sdl.Keyboard.Keymod.CapsLock,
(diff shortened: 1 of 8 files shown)


tomspilman:
Glad to see the cleanup of the `Keyboard` classes.

As long as we can correctly report these states on all platforms i'm good with the addition.  It just has to be reliable else it is a waste of time to add it.

@KonajuGames @dellis1972 ?

MonoGame.Framework/Input/Keyboard.SDL.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Microsoft.Xna.Framework.Input
+{
+    public static partial class Keyboard
+    {
+        static List<Keys> _keys;
+
+        private static KeyboardState PlatformGetState()
+        {
+            var modifiers = Sdl.Keyboard.GetModState();
+            return new KeyboardState(_keys, modifiers.HasFlag(Sdl.Keyboard.Keymod.CapsLock), modifiers.HasFlag(Sdl.Keyboard.Keymod.NumLock));
```
KonajuGames: Enum.HasFlag() is slower and can generate garbage as it does [type checks](http://referencesource.microsoft.com/#mscorlib/system/enum.cs,877).  I would prefer to see a simpler and garbage free test such as

    (modifiers & Sdl.Keyboard.Keymod.CapsLock) == Sdl.Keyboard.Keymod.CapsLock

cra0zy: Done.

tomspilman:
Some unit tests for keyboard would help with https://github.com/mono/MonoGame/issues/4744.

cra0zy:
> Some unit tests for keyboard would help with #4744.

You're evil :P

tomspilman:
> You're evil :P

Someone has to be the unit test bad guy. :)

PR #4790, opened 2016-04-21
[MG Framework] Refactor Mouse

cra0zy:
Fixes https://github.com/mono/MonoGame/issues/4719

### DIFF
--- a/MonoGame.Framework/Input/Mouse.cs
+++ b/MonoGame.Framework/Input/Mouse.cs
@@ -4,79 +4,33 @@
 
 using System;
 
-#if MONOMAC || WINDOWS
-using System.Runtime.InteropServices;
-using System.Drawing;
-#endif
-
-#if OPENGL && MONOMAC
-#if PLATFORM_MACOS_LEGACY
-using MonoMac.Foundation;
-using MonoMac.AppKit;
-#else
-using Foundation;
-using AppKit;
-using PointF = CoreGraphics.CGPoint;
-#endif
-#endif
-
 namespace Microsoft.Xna.Framework.Input
 {
     /// <summary>
     /// Allows reading position and button click information from mouse.
     /// </summary>
-    public static class Mouse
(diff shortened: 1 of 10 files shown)


MonoGame.Framework/Input/Mouse.cs
```diff
@@ -151,87 +63,18 @@ public static MouseState GetState()
...
-            var flippedPt = new PointF(screenPt.X, currentScreen.Frame.Size.Height-screenPt.Y);
-            flippedPt.Y += currentScreen.Frame.Location.Y;
-            
-            
-            CGSetLocalEventsSuppressionInterval(0.0);
-            CGWarpMouseCursorPosition(flippedPt);
-            CGSetLocalEventsSuppressionInterval(0.25);
-#elif WEB
             PrimaryWindow.MouseState.X = x;
             PrimaryWindow.MouseState.Y = y;
```
tomspilman: What is the point of setting the state here and still calling `PlatformSetPosition(x,y)` too?

tomspilman:
Looks good overall.

MonoGame.Framework/Input/Mouse.cs
cra0zy: The reason was immediate update because platformsetposition wouldn't set the state of the mouse so get state might have not registered the new position, but now that I think about it, it should get moved to per platform since some platforms don't support setting mouse position.

tomspilman:
Is there any sort of useful unit testing we can have for this class (https://github.com/mono/MonoGame/issues/4744)?  Like just calling the methods to be sure it doesn't just outright crash?

Or should we only target it for the new manual testing framework? https://github.com/mono/MonoGame/issues/3793

cra0zy:
I see no harm in adding some NUnit tests for the state.

tomspilman:
This is great... thanks @cra0zy !

PR #4793, opened 2016-04-21
Add Android gamepad support

vostok4:
Reopening a pull request to add native gamepad support for Android.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -50,10 +50,10 @@
       <Platform Name="Windows8">TRACE;NETFX_CORE;WINRT;WINDOWS_STOREAPP;DIRECTX;DIRECTX11_1;WINDOWS_MEDIA_ENGINE</Platform>
       <Platform Name="WindowsPhone">TRACE;SILVERLIGHT;WINDOWS_PHONE;WINRT;DIRECTX</Platform>
       <Platform Name="WindowsPhone81">TRACE;WINRT;WINDOWS_PHONE81;DIRECTX;WINDOWS_STOREAPP</Platform>
-      <Platform Name="WindowsUniversal">TRACE;NETFX_CORE;WINDOWS_UAP;WINRT;DIRECTX;DIRECTX11_1;WINDOWS_MEDIA_ENGINE</Platform>		
+      <Platform Name="WindowsUniversal">TRACE;NETFX_CORE;WINDOWS_UAP;WINRT;DIRECTX;DIRECTX11_1;WINDOWS_MEDIA_ENGINE</Platform>
       <Platform Name="Web">TRACE;WEB</Platform>
     </CustomDefinitions>
-	
+
     <FrameworkVersions>
 	  <Platform Name="Windows8">
 	    <Version>v4.5.1</Version>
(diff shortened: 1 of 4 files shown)


vostok4:
OK so the build error seems really odd. It is complaining that Android.View.InputDevice.HasKeys() doesn't exist, and according to the Android SDK it shouldn't, it should be *hasKeys()*, not *HasKeys()* (case):

http://developer.android.com/reference/android/view/InputDevice.html#hasKeys(int...)

According to the Xamarin Android API docs, it is a "big H":

https://developer.xamarin.com/api/member/Android.Views.InputDevice.HasKeys/p/System.Int32[]/

On TeamCity Mac, iOS and Linux it builds, but on Windows it fails, which is odd, because I am developing on Windows here, and the large H is working fine.

Anyone have any ideas what's up with TeamCity, or how should I fix this build error?

mrhelmut:
Maybe @dellis1972 have some clues.

danzel:
What API level are we building against? That API was added in V19, we're probably building against an older one.

Was trying to find where it is defined for monogame, but can't off github code search. I expect this is a protobuild option?

https://github.com/mono/MonoGame/blob/8a8f21c65c6e40688ed75b64c65bbf5a5790e40f/IDE/MonoDevelop/MonoDevelop.MonoGame.Android/templates/Android/AndroidManifest.xml

https://github.com/mono/MonoGame/blob/8a8f21c65c6e40688ed75b64c65bbf5a5790e40f/Documentation/android.md
says we target API 17.

vostok4:
Is there any accepted way to ping pull requests? I'd love to see this integrated, it would help on a couple of ports for me.

KonajuGames:
`hasKeys()` (and Xamarin's corresponding `HasKeys()`) was introduced with Android 4.4.  MonoGame is currently built to target 4.2, which is why it complains that the method is missing.

If you add the following to the `<FrameworkVersions>` node in `Build\Projects\MonoGame.Framework.definition`, it will set the Android target version for Monogame.Framework.Android.csproj to v4.4

	  <Platform Name="Android">
	    <Version>v4.4</Version>
	  </Platform>

In this case, you will also want to modify the `<TargetFrameworkVersion>` node in the Android project template in `ProjectTemplates\VisualStudio2010\Android\AndroidApplication1.csproj` to also target v4.4.

vostok4:
Thank you @KonajuGames, I've updated the PR, everything seems to be passing now (the build checks are useful! I ended up not realized there were !OUYA ifdefs, and was very heavy handed in clearing OUYA ifdefs in the Android class).

Build/Projects/MonoGame.Framework.definition
```diff
@@ -50,14 +50,17 @@
...
       <Platform Name="Web">TRACE;WEB</Platform>
     </CustomDefinitions>
-	
+
     <FrameworkVersions>
 	  <Platform Name="Windows8">
 	    <Version>v4.5.1</Version>
 	  </Platform>
+	  <Platform Name="Android">
+            <Version>v4.4</Version>
```
mrhelmut: @vostok4  This property is for setting .NET Framework target version, not Android API. Would you mind removing it?

KonajuGames: @mrhelmut Try changing the Android API version in the project settings and saving the project.  It changes that property in the csproj.

mrhelmut: My bad then (it's a bit misleading).

KonajuGames:
Thanks @vostok4. Merging.

PR #4797, opened 2016-04-23
[SDL] Better window size managment

### LINKED ISSUE #4783: [DesktopGL] Fullscreen always set to 800x600
Hello,

When entering fullscreen (using ```graphics.HardwareModeSwitch = true;```), ```SDLGameWindow``` always resets the ```Viewport``` and backbuffer size to 800x600.

This is due to SDL triggering a ```Sdl.Window.Event.Id.Resized``` (which is correct) but with ```ev.Window.Data1``` & ```ev.Window.Data2``` always returning 800 & 600.

As far as I have tested, this happens on Windows and OS X.

Everything is fine on windowed or ```graphics.HardwareModeSwitch = false;``` fullscreen modes (the event values are the intended ones).

@cra0zy any thoughts? I don't know if this is the normal ```Sdl.Window.State.Fullscreen``` behavior.

cra0zy:
Fixes #4783

SDL does not register the new size until the next Update loop so we need to keep our own copy of them.

### DIFF
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -27,12 +27,9 @@ namespace Microsoft.Xna.Framework
         {
             get
             {
-                int x = 0, y = 0, w, h;
-
-                Sdl.Window.GetSize(Handle, out w, out h);
+                int x = 0, y = 0;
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @cra0zy !

PR #4798, opened 2016-04-24
MonoGame tvOS now compiles. Fixes:

vchelaru:
- Added needed assembly references in protobuild .definition file.
 - Added code reference to GamePlatform.Mobile.cs
 - Added explicit cast to deal with nine
 - Removed thumbnail from Album.cs since backing class seems to not exist on tvOS.

### DIFF
--- a/Build/Projects/FrameworkReferences.definition
+++ b/Build/Projects/FrameworkReferences.definition
@@ -8,6 +8,9 @@
     <Reference Include="OpenTK-1.0" />
   </Platform>
   <Platform Type="tvOS">
+    <Reference Include="System" />
+    <Reference Include="System.ComponentModel.Composition” />
+    <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.Core" />
     <Reference Include="OpenTK-1.0" />
   </Platform>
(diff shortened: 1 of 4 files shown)


Build/Projects/FrameworkReferences.definition
```diff
@@ -8,6 +8,9 @@
     <Reference Include="OpenTK-1.0" />
   </Platform>
   <Platform Type="tvOS">
+    <Reference Include="System" />
+    <Reference Include="System.ComponentModel.Composition” />
```
tomspilman: You have one of those fancy quotes at the end of this `”`...  needs to be a plain quote char.

MonoGame.Framework/Input/GamePad.tvOS.cs
```diff
@@ -18,10 +18,15 @@ private static int PlatformGetMaxIndex()
             return 4;
         }
 
+        private static int PlatformGetMaxNumberOfGamePads ()
+        {
+            return PlatformGetMaxIndex ();
+        }
```
tomspilman: I think you can just remove the `PlatformGetMaxIndex()` function entirely and return 4 here.

tomspilman:
@Nezz - Thoughts on the Album.cs changes here?

Nezz:
Looks good to me as long as it compiles :)

tomspilman:
Thanks @Nezz !

dellis1972:
Thanks @vchelaru

PR #4801, opened 2016-04-25
Only use Default implementation on older Mac APIs

hach-que:
This allows Xamarin.Mac to build again without introducing any regressions in the current behaviour.  If we opted for only GamePad.Mac or GamePad.Default to be included in the project, then we'd either introduce regressions for MonoMac or regressions for Xamarin.Mac.


hach-que:
@tomspilman Can we merge this ASAP so Xamarin.Mac starts working again?

tomspilman:
Sure...  this seems like it should be fine for now.  We just need to cleanup these `#if`s after https://github.com/mono/MonoGame/issues/4693.

PR #4806, opened 2016-04-26
Add NVorbis submodule

cra0zy:
So instead of us having an external .dll dependency we could just include the entire NVorbis source code inside. This will allow for the end user to only have to reference MonoGame.Framework.dll and not have to worry about any other dlls.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "ThirdParty/Dependencies"]
 	path = ThirdParty/Dependencies
 	url = https://github.com/Mono-Game/MonoGame.Dependencies.git
+[submodule "ThirdParty/NVorbis"]
+	path = ThirdParty/NVorbis
+	url = https://github.com/ioctlLR/NVorbis.git
--- a/Build/Projects/FrameworkReferences.definition
+++ b/Build/Projects/FrameworkReferences.definition
@@ -17,9 +17,6 @@
     <Service Name="MonoGame.Framework/OpenGLGraphics">
       <Reference Include="System.Drawing" />
       <Binary
-        Name="NVorbis"
-        Path="ThirdParty/Dependencies/NVorbis/NVorbis.dll" />
-      <Binary
         Name="OpenTK"
         Path="ThirdParty/Dependencies/OpenTK.dll" />
     </Service>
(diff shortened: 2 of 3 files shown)


mrhelmut:
I'm fine with this.

angrywasp:
I'm sorry, but can you please explain how this would be preferable to just forking NVorbis? Smashing the entire source code into a single file of nearly 7000 lines doesn't seem like it would make anything easier to me

angrywasp:
Can you please explain the reasoning for this? Seems to me like a fork of NVorbis would be a lot easier if you wanted to modify the code. This method results in a file of nearly 7000 lines of code, which is just going to be a nightmare to work with. Also, if (seems unlikely) NVorbis gets any new commits, it would only further complicate maintenance on this file if we have local changes

Is there a specific bug in NVorbis that needs correcting? If not, then my suggestion would be to leave it as it is. It would be good to get rid of the dll, but i'm not convinced this is the best means to that end

cra0zy:
> Can you please explain the reasoning for this?

End user will only have to reference MonoGame.Framework.dll instead of search what other dlls that are needed.

> Seems to me like a fork of NVorbis would be a lot easier if you wanted to modify the code. This method results in a file of nearly 7000 lines of code, which is just going to be a nightmare to work with. Also, if (seems unlikely) NVorbis gets any new commits, it would only further complicate maintenance on this file if we have local changes

NVorbis.cs should not be edited by hand, instead you should:
1) Fork NVorbis reposetory
2) Make modifications
3) Submit a patch to NVorbis first
4) Use the script to generate NVorbis.cs
5) Submit new NVorbis.cs to MonoGame

Making NVorbis.cs one file is actually really good because it can be easily switched with NVobis project file in case of needed modifications.

Also do note that NVorbis is a really stable piece of sofrware so it won't be changed that often.

danzel:
We could git submodule in nvorbis rather than copy/pasting it over?

Bringing the code in by copying (well generating a copy using a script) seems a bit dirty.

cra0zy:
> We could git submodule in nvorbis rather than copy/pasting it over?
> 
> Bringing the code in by copying (well generating a copy using a script) seems a bit dirty.

Good point, will update the PR tomorrow morning.

cra0zy:
@danzel Done :)

tomspilman:
I'm ok with this I guess.  I just checked and it adds less than 2MB to the clone size of the MonoGame repo.

@KonajuGames ?

cra0zy:
This should probably get merged before DesktopGL templates update so they can be fully updated in one PR.

KonajuGames:
Merging.  Thanks @cra0zy

PR #4807, opened 2016-04-26
Added new design converters and extended the pipeline

thegouldfish:
Added Colour and Point designer converters (both used in the pipeline tool) and extended the pipelineTypes so it can be used by external tools (anything that might want to load and use the content project).

### DIFF
--- a/MonoGame.Framework/Design/PointTypeConverter.cs
+++ b/MonoGame.Framework/Design/PointTypeConverter.cs
@@ -1,0 +1,100 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Xna.Framework.Design
+{
+    public class PointTypeConverter : TypeConverter
+    {
+        public override bool CanConvertTo(ITypeDescriptorContext context, Type destinationType)
+        {
+            if (destinationType == typeof(int))
+                return true;
+            if (destinationType == typeof(Point))
+                return true;
+            if (destinationType == typeof(string))
+                return true;
+
+            return base.CanConvertTo(context, destinationType);
+        }
+
+        public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
+        {
+            try
+            {
+                throw new Exception();
+            }
+            catch (Exception ex)
+            {
+
+            }
+
+            var point = (Point)value;
+
+
+            if (destinationType == typeof(int))
+            {
+                return point.X;
+            }
+
+            if (destinationType == typeof(Point))
+            {
+                return point;
+            }
+
+
+
+            if (destinationType == typeof(string))
+            {
+                var terms = new string[2];
+                terms[0] = point.X.ToString(culture);
+                terms[1] = point.Y.ToString(culture);
+
+                return string.Join(culture.TextInfo.ListSeparator + " ", terms);
+            }
+
+            return base.ConvertTo(context, culture, value, destinationType);
+        }
+
+
+        public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
+        {
+            if (sourceType == typeof(string))
+                return true;
+
+            return base.CanConvertFrom(context, sourceType);
+        }
+
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
+        {
+            try
+            {
+                throw new Exception();
+            }
+            catch (Exception ex)
+            {
+
+            }
+
+
+            var sourceType = value.GetType();
+            var point = Point.Zero;
+
+            if (sourceType == typeof(string))
+            {
+                var str = (string)value;
+                var words = str.Split(culture.TextInfo.ListSeparator.ToCharArray());
+
+                point.X = int.Parse(words[0], culture);
+                point.Y = int.Parse(words[1], culture);
+
+                return point;
+            }
+
+            return base.ConvertFrom(context, culture, value);
+        }
+    }
+}
--- a/Tools/Pipeline/Common/PipelineTypes.cs
+++ b/Tools/Pipeline/Common/PipelineTypes.cs
@@ -145,7 +145,7 @@ namespace MonoGame.Tools.Pipeline
         }
     };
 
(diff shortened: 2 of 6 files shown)


tomspilman:
I don't know anything about these designer type converters, so i'll defer to @KonajuGames, @dellis1972 or @cra0zy on this one.

@cra0zy - How do you feel about `PipelineTypes` being made public?  Do we feel that API is acceptable for others to use for reading/writing Pipeline project files?

tomspilman:
Looking at the XNA docs...

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.design.aspx

The type converters have slightly different names in official XNA and some derive from a common `MathTypeConverter`.  Should we be looking to mimic that?

cra0zy:
> and extended the pipelineTypes so it can be used by external tools

The class is located inside Pipeline Tool, and nothing ever interacts with it, why would you want to make it public, can you give an example?

> The type converters have slightly different names in official XNA and some derive from a common MathTypeConverter. Should we be looking to mimic that?

I think we should.

thegouldfish:
RE: Pipeline,
  I have an editor that I've been working on since the XNA 3/4 days and it would load the content project in and edit it and then use it for building the data so the editor can use it with the same classes that my projects do.  This was easier when the content was a regular project but harder now its a custom file.  But I found that I could reference the new content tool and use most of its classes to load up the content but I needed to be able to access the PipelineTypes or I would get some an exception when writing it out.
This is all hooked up with an extended ContentManager that uses all this to automatically add assets to the project if they are requested but not already a part of it.

RE: Design Converters
  I can have a look at that common MathTypeConverter.

MonoGame.Framework/Design/PointTypeConverter.cs
```diff
@@ -0,0 +1,100 @@
...
+                return true;
+
+            return base.CanConvertFrom(context, sourceType);
+        }
+
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
+        {
+            try
+            {
+                throw new Exception();
```
KonajuGames: What is this try..catch block for that simply throws an exception?

MonoGame.Framework/Design/PointTypeConverter.cs
```diff
@@ -0,0 +1,100 @@
...
+                return true;
+
+            return base.CanConvertTo(context, destinationType);
+        }
+
+        public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
+        {
+            try
+            {
+                throw new Exception();
```
KonajuGames: Another try..catch block that simply throws an exception.

MonoGame.Framework/Design/PointTypeConverter.cs
```diff
@@ -0,0 +1,100 @@
+﻿using System;
```
KonajuGames: New files must have the MonoGame license header.

````
````

File: MonoGame.Framework/Design/PointTypeConverter.cs [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Globalization;
+using System.Linq;
+using System.Text;
```
KonajuGames: Only include `using` statements for namespaces actually used.  Easiest way to do this in Visual Studio is right-click on the using block, select `Organize Usings` and then select `Remove and Sort`.

Tools/Pipeline/Common/PipelineTypes.cs
```diff
@@ -426,9 +426,16 @@ private static void ResolveAssemblies(IEnumerable<string> assemblyPaths)
             {
                 try
                 {                    
-                    var a = Assembly.LoadFrom(path);
-                    var types = a.GetTypes();
-                    ProcessTypes(types);
+                    // Only load in Assemblies that haven't already been loaded in.
+                    var reflectedAss = Assembly.ReflectionOnlyLoadFrom(path);
+                    bool found = assemblies.Any(ass => ass.FullName == reflectedAss.FullName);
```
KonajuGames: Please don't abbreviate variable names where the abbreviation could be considered offensive.  Just use the full name `reflectedAssemblies` and `assembly`.

KonajuGames:
> The type converters have slightly different names in official XNA and some derive from a common MathTypeConverter. Should we be looking to mimic that?

It should use the same typenames and hierarchy.

thegouldfish:
Sorry I had those try catch blocks for testing, and it seems I left them in.
I spotted them earlier when I was renaming them, and will be removed.

I'll rename those variables, I've used those names for years with out any bother and I will get the header added.

I'll look at renaming all the type converters and none of them match the XNA names and I will get them to inherit off the MathsTypeConverter (although there might not be much in that class).

KonajuGames:
> I'll rename those variables, I've used those names for years with out any bother

Thanks.  As a highly visible public project with open access to source, we need to be careful about the language used in the code.

thegouldfish:
Sorry, Fixing it up now. (Cool you use TeamCity)

tomspilman:
@KonajuGames - What do you think about merging this?

I would like to see the changes to Vector2.cs and Vector3.cs fixed to not change the entire file content.  But I don't know how it impacts things other than that... so I leave the decision to others.

thegouldfish:
I know this pull request is as old as the hills now, but did anything from this ever get taken into monogame?

The core of it was to improve tooling with additional class designers, and to change the content pipeline so anyone could open the content project programmatically in external tooling.

If not could we just close this and I'll get latest and try again.

dellis1972:
@thegouldfish you might be able to just rebase this branch, its worth a try :)

thegouldfish:
@dellis1972 I'll look into that, I'm not much of a Git user.

Jjagg:
Join the gitter channel if you need some help :)

tomspilman:
@harry-cpp does this affect your plans for content pipeline?

If anyone wants to spearhead fixing the merge conflict we could merge this.

We will close this in a few weeks if no one takes it over.

harry-cpp:
Doesn't affect it.

SimonDarksideJ:
@thegouldfish Did you ever manage to restart this PR?

PR #4811, opened 2016-04-28
VideoPlayer.WMS improvements and features

KonajuGames:
Create video texture once and re-use it instead of creating a new texture each frame.
Implement looped video support.

Video events should be dispatched via the FrameworkDispatcher once it is implemented.


tomspilman:
Any sort of test we can work out for this?  Even just calling the APIs, checking parameter and return values are correct, etc would help improve reliability.

Note I did try to write some unit tests once for video that played a simple video with some different checkerboard colored sequences that lasted a few seconds each, then captured one frame per second to compare to reference frames.  It sort of worked to validate the video output...  but of course nothing was checking the audio.

KonajuGames:
I was thinking about what unit tests could be done.  I'll add some to this PR tonight.

KonajuGames:
VideoPlayer
- Added missing call to Dispose(false) in the finalizer.

VideoPlayer.WMS
- All MediaFoundation calls are asynchronous.  Need to use the event callback to know when to call the next MF function.  Added internal state to cater for the event system and still behave synchronous to the caller.  It gets interesting when some MF calls do not trigger an event due to the current state of the MF session.
- The session clock was not a good indicator of state as it continues after the video has finished normally.
- Looping and non-looping videos now work correctly.
- Playing a video immediately after another video no longer throws an InvalidOperationException because the video did not appear to start the session correctly (revealed in unit tests).  Solved by handling events correctly.

Tests
- Added VideoImporterTests, VideoTests and related assets.
- Corrected pathParths -> pathPaths.

KonajuGames:
Perhaps the video playback tests don't like the automated system?  Can't get much info from those logs.

tomspilman:
@KonajuGames - If you revert that last commit and re-enable the unit tests I can try to diagnose why this is failing.  We really need to solve that else we will never effectively test everything.

My guess is that it is something to do with the build machine being headless and it maybe acting like the screen is asleep.

I will have to manually run tests on the box to figure it out I suppose.

romanov:
The long awaited fix. :two_hearts:

KonajuGames:
Tom and I would like to get the unit tests running properly on our build
system, but yes it does address a few issues.

KonajuGames:
This was supposed to be a damn rebase. Git is brain dead sometimes.

PR #4813, opened 2016-04-29
Added tests for GamePad per #4744

confusingbits:
NUnit VS Adapter 2.0.0.0 executing tests is finished
========== Run test finished: 218 run (0:00:02.2515109) ==========

Not really sure if the Null is valid, but I checked anyway.

### DIFF
--- a/Test/Framework/Input/GamePadTest.cs
+++ b/Test/Framework/Input/GamePadTest.cs
@@ -1,0 +1,136 @@
+﻿using NUnit.Framework;
+using System;
+using Microsoft.Xna.Framework.Input;
+using Microsoft.Xna.Framework;
+
+namespace MonoGame.Tests
+{
+    [TestFixture]
+    public class GamePadTests
+    {
+        private Array GetPlayerIndexes() // gets a list of enum values to test
+        {
+            return Enum.GetValues(typeof(PlayerIndex));
+        }
+
+        private Array GetGamePadDeadZoneIndexes()
+        {
+            return Enum.GetValues(typeof(GamePadDeadZone));
+        }
+
+        [Test] // generate tests for each PlayerIndex enum value
+        public void GetCapabilities_ShouldNotReturnNull_GivenPlayerIndex([ValueSource("GetPlayerIndexes")] PlayerIndex pi)
+        {
+            Assert.IsNotNull(GamePad.GetCapabilities(pi));
+        }
+
+        [Test] // generate tests for each PlayerIndex enum value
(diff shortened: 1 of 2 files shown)


Tzenchor:
You have to add the class to `Build\Projects\Monogame.Tests.definition`

WaveFunctionP:
I took a gander at protobuild docs and the defintions, and I'm guessing I did that right? ¯\_(ツ)_/¯

Tzenchor:
> Looks like there is an untracked KeyboardTest.cs in the same location as well.

That must be in your local copy. I have my develop branch synced and I do not have it

> Is manually editing the definitions the proper workflow?

It is. Theoretically it is possible to tell protobuild to resync the definitions, but the time I tried it said that that action was disabled

WaveFunctionP:
Note: Here's KeyboardTest.cs in the develop branch.
https://github.com/mono/MonoGame/blob/develop/Test/Framework/Input/KeyboardTest.cs

WaveFunctionP:
Forgot to mention, if these tests look good, I will begin working on more tests for the other classes in #4744

Test/Framework/Input/GamePadTest.cs
```diff
@@ -0,0 +1,136 @@
...
+            [Range(0.0f, 1.0f, 1.0f)] float rm
+            )
+        {
+            Assert.IsInstanceOf<bool>(GamePad.SetVibration(pi, lm, rm));
+        }
+
+        [Test]
+        public void MaximumGamePadCount_ShouldNotReturnNull_WhenGet()
+        {
+            Assert.IsNotNull(GamePad.MaximumGamePadCount);
```
tomspilman: This seems like an unnecessary test... if `GamePad.MaximumGamePadCount` returns an `int` it can never be null.

Test/Framework/Input/GamePadTest.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        [Test]
+        public void MaximumGamePadCount_ShouldReturnInt_WhenGet()
+        {
+            Assert.IsInstanceOf<int>(GamePad.MaximumGamePadCount);
```
tomspilman: A better test here is ensuring that `GamePad.MaximumGamePadCount` is within a reasonable range like 0 to 16.

WaveFunctionP:
NUnit VS Adapter 2.0.0.0 executing tests is finished
========== Run test finished: 217 run (0:00:02.4923933) ==========

PR #4815, opened 2016-05-02
Multisampling and RenderTarget2D

palver123:
I have spent a lot of time debugging this bug and finally cracked it with DebugView and DirectX Debug Layer. You cannot create a depth stencil view with multisampling if you do not set the underlying texture dimension. This only affects the DirectX version of RenderTarget2D I think.


palver123:
Recently I came across the problem of RenderTarget2D and multi-sampling again. Anyone read my pull request? I know it is not a complete fix, nor I provided unit tests to verify my changes, but at least a discussion should start on what's missing.

Fixme, but this feature was supported in XNA.

PR #4817, opened 2016-05-04
ADPCM support for SoundEffect

SeriousMartin:
Fixes the calculation of loopLength for ADPCM sound files

What this pull request is missing (from my point of view) is a clear distinction between the SoundEffect constructor that supports ADPCM and the one that doesn't.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
@@ -350,10 +350,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Audio
 
                 // Loop start and length in number of samples. Defaults to entire sound
                 loopStart = 0;
-                if (data != null && bitsPerSample > 0 && channelCount > 0)
-                    loopLength = data.Count / ((bitsPerSample / 8) * channelCount);
-                else
-                    loopLength = 0;
+                loopLength = (int)Math.Floor(sampleRate * durationInSeconds);
             }
             finally
             {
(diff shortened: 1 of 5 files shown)


SeriousMartin:
My pull request is unnecessary. All needed changes have already been made with https://github.com/mono/MonoGame/commit/8a70a734c93e335c2c50683aa31744316bf6374f. I didn't realize this when merging my changes from master to develop.

SeriousMartin:
This was a bit too hasty. 
One thing still has to be done: Re-active ADPCM compression in SoundEffectProcessor. This should work now.

KonajuGames:
Before merging this, we would need to check that MS-ADPCM files can be loaded correctly on Android.  I know openal-soft has support for playing back both MS and IMA ADPCM.  We just need to make sure the support is in our code for using them correctly.

KonajuGames:
Also on desktop platforms that use OpenAL.

dellis1972:
@KonajuGames I am working on code which optionally uses openal MS_ADPCM support IF Al.IsExtensionSupported reports that it is available.

tomspilman:
@KonajuGames  @dellis1972 - So is this good to go?

Should we wait for SDL2 or merge this now?

dellis1972:
@tomspilman we can probably merge this one then update SDL2. This is does the DirectX side of the house anyway I think so there shouldn't be too much of a conflict. //cc @cra0zy

cra0zy:
I'm fine with this getting merged before SDL2.

tomspilman:
Thanks @SeriousMartin !

Issue #4820, opened 2016-05-04
Texture2D.FromStream Test


tomspilman:
This is something I noticed in XNA behavior porting a game to PS4.

If you have a texture that is filled with the color:  `R:255 G:128 B:64 A:0`.  When you load this texture via `Texture2D.FromStream` the resulting texture has the color: `R:0 G:0 B:0 A:0`.

If you have a texture that is filled with the color:  `R:255 G:128 B:64 A:1`...  `Texture2D.FromStream` returns a texture with the color: `R:255 G:128 B:64 A:1`.

So although the XNA docs for `Texture2D.FromStream` explicitly state that it returns non-premultiplied colors... it does however clear the color channels if the alpha channel is 0.

I don't know if MonoGame matches XNA in this regard or not.  We need to implement a specific unit test with a special test texture to verify the results.

Secondarily we should consider adding an overload that lets you disable this color channel clearing as it was unexpected.  I can see how it would be useful to not have it modify the loaded color data.

Tzenchor:
There is a bug. For transparent images works like XNA, but for not semi transparent it is multiplying the color channel by the alpha.
I have used this color for the test: R: 254, G: 128, B: 63, A: 129. But MonoGame is returning this one: R: 129, G: 65, B: 32, A: 129; (There might have been a problem with the test image because paint.net tells one color, and XNA another one)

tomspilman:
> but for semi transparent it is multiplying the color channel by the alpha.

You mean premultiplied alpha?  XNA doesn't do that according to the docs:

> A 2D texture whose format is SurfaceFormat.Color and contains non-premultiplied alpha data.
https://msdn.microsoft.com/en-us/library/ff434103.aspx

KonajuGames:
Our implementations do not zero RGB if A is zero.  I had no idea XNA did that.

Our implementations should not be pre-multiplying alpha.  If so, then that is a bug and should be fixed.

Tzenchor:
> You mean premultiplied alpha? XNA doesn't do that according to the docs:

My fault explaining myself. Monogame is the one premultiplying. `R: 254, G: 128, B: 63, A: 129` Is the result from XNA and `R: 129, G: 65, B: 32, A: 129` the one from Monogame with the same image.

Will do a pull later with the tests or at least i'll publish the branch to my fork.

PR #4823, opened 2016-05-05
Tests for Texture2d.FromStream

Tzenchor:
Related to #4820 
This implements the test for Textures with `Alpha=0` and `Alpha=129`, both tests work on XNA, but the second one fails on Monogame.


tomspilman:
Cool...  this proves the problem.  Now to implement the fix.

Tzenchor:
There's a problem.
Image loading depends on SharpDX, and the decoder is set to `PixelFormat.Format32bppPRGBA`, that I assume it is Premultiplied RGBA. So I tried with `PixelFormat.Format32bppRGBA`, but both tests fail.
With the transparent image it's returning `R: 255, G: 255, B: 255, A: 0` (the color is wrong but I think it is more a problem with the image editor when painting with 0 alpha than a bug on SharpDX).
With the semi transparent the result is: `R: 255, G: 128, B: 64, A: 129` that nearly matches what XNA returned `R: 254, G: 128, B: 63, A: 129`

KonajuGames:
I say we use our own PNG loader like on the other platforms. More control
over the process. This does lose JPG and GIF loading though.

tomspilman:
> I say we use our own PNG loader like on the other platforms.

I agree...  consistency across platforms is critical and we can't get that using platform specific loaders.


> This does lose JPG and GIF loading though.

Minor short term losses IMO.

Tzenchor:
> I say we use our own PNG loader like on the other platforms.

I'm supposing it is `MonoGame.Utilities.Png.PngReader`. I tried it and the results are the same as with the SharpDX loader as said in my previous post

Tzenchor:
I have just made a little test and without premultiplied alpha texture composition does not work as expected. I'm printing a brown opaque texture and a white transparent one above, and the result is nearly white.

Also I have found a bug in `Texture2D.GetData<T>(int, Rectangle?, T[], int, int)` on DirectX, if startIndex (the penultimate parameter) is not zero, nothing is copied to the array (Tested on XNA and it works).

KonajuGames:
> I have just made a little test and without premultiplied alpha texture composition does not work as expected.

You rendered the non-premultiplied texture with `BlendState.NonPremultiplied`?

Tzenchor:
> You rendered the non-premultiplied texture with BlendState.NonPremultiplied?

No, but the code is the same I had from XNA

Tzenchor:
Now that I have fixed the problem I had with the image editor i'm using an image with color:
`R: 136, G: 66, B: 16, A: 24;` SharpDX loader without premultiplied alpha is returning that color, the `PngReader` stated in a previous post returns that color, the "color selector" (or whatever it is called in english) tool in paint.net also returns that color, but in XNA I get `R: 138, G: 63, B: 21, A: 24;`. And the weirdest thing is with `R: 136, G: 66, B: 16, A: 24;`, MonoGame methods keep returning the same color, but XNA now gives an amazing `R: 255, G: 0, B: 0, A: 1;`

Another weird thing is that I have the a XNA 4.0 project where I load some textures with `FromStream` and the resulting image is the same one as with Monogame before begining this work

tomspilman:
I fixed this issue finally in https://github.com/MonoGame/MonoGame/pull/5712 along with unit tests similar to these.

PR #4825, opened 2016-05-06
[SpriteBatch] Optimize _tempRect

nkast:
Skip expensive operations when sourceRectangle is `null`.


KonajuGames:
Thanks @nkast. Merging.

Issue #4832, opened 2016-05-10
Directx Texture2D.PlatformGetData bug


Tzenchor:
Not sure if this also affects OpenGL implementation.

The bug is in `private void PlatformGetData<T>(int level, int arraySlice, Rectangle? rect, T[] data, int startIndex, int elementCount)` the `startIndex` parameter is used as in offset in the texture data instead of being an offset for the `data` parameter.

To reproduce the bug take any non transparent black texture and use this code
```cs
Texture2D t= Texture2D.FromStream("whatever it is called");
Color[] colors=new Color[2];
t.GetData(0, new Rectangle(0,0,1,1), colors,1,1);
Assert.AreNotEqual(0, colors[1].A);
```

Will do a pull with the fix and the test shortly, but I have to test this a little with different rectangle sizes

PR #4833, opened 2016-05-10
Fix PlatformGetData

Tzenchor:
Implement the tests that demonstrate the bug stated in #4832.
While readying the test found another difference compared to XNA and added another test for that: XNA throws an `ArgumentException` if the `elementCount` parameter is less than the area of the `Rectangle` from where to get the Texture data.


Tzenchor:
Done, for trying to avoid the changes to the trailing whitespaces I had to do to 2 additional commits

tomspilman:
This is great work... thanks @Tzenchor !  We'll deal with the GL path when we get unit tests running on the Mac.

PR #4836, opened 2016-05-11
[Pipeline Tool] Fix context menu separator visibility

cra0zy:
This change makes it so separator code check the cases for:
 - 2 separators one after another, that means skip placing separator
 - removes last item if it's a separator


cra0zy:
@tomspilman When you get the chance, do go over this, it fixes up separators not being visible on "Settings/Base" item.

tomspilman:
Merging!

PR #4837, opened 2016-05-11
[Pipeline Tool] Fix property grid editing dialogs blocking UI thread

cra0zy:


tomspilman:
Merging... thanks @cra0zy !

PR #4844, opened 2016-05-13
[Pipeline Tool] Few minor improvements

cra0zy:
Stuff done:
 - updates submodule
 - removes Ctrl + C "Close" shortcut
 - makes propertygrid dialog starting different depending if it's on Windows/Unix to properly resolve their drawings

### DIFF
--- a/Tools/Pipeline/Controls/PropertyGridTable.cs
+++ b/Tools/Pipeline/Controls/PropertyGridTable.cs
@@ -176,9 +176,12 @@ namespace MonoGame.Tools.Pipeline
             else
                 SetCursor(CursorType.Default);
 
+            // On windows craeting a dialog from double click will freeze
+            // the GUI thread until a click occurs so we need to call the
+            // dialog at the end of Paint event so everything gets drawn.
             if(_edit)
             {
-                if (_selectedCell != null && _selectedCell.Editable)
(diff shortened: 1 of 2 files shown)


Tools/Pipeline/Controls/PropertyGridTable.cs
```diff
@@ -176,9 +176,12 @@ private void Drawable_Paint(object sender, PaintEventArgs e)
             else
                 SetCursor(CursorType.Default);
 
+            // On windows craeting a dialog from double click will freeze
+            // the GUI thread until a click occurs so we need to call the
+            // dialog at the end of Paint event so everything gets drawn.
```
tomspilman: This sounds crazy...  that never happens under regular C# WinForms or even C++ Win32 applications.

Is this an Eto bug?

cra0zy: Not sure if it's due to Eto or Windows Forms, still have to test that out. Anyway we have a fix so it's not really a priority...

Also for Gtk I have to call the dialog from the click event because Gtk drawing calls need to complete in order to send another draw call, which means if we start the dialog inside Draw event, the application will stop redrawing itself.

tomspilman:
Thanks... merging!

PR #4850, opened 2016-05-13
[Pipeline Tool] Windows UI Improvements

cra0zy:
Stuff done:
 - menubar
   - [Only on Windows 10] copies the window titlebar color info
 - toolbar
   - hide rounded edges
   - add padding
 - project control / filter view
   - hide doted lines so it's more like VS project view
   - full row select
   - increased row height to be 20 to gain a bit of spacing

Preview:
untitled3


tomspilman:
Thanks... merging!

PR #4852, opened 2016-05-13
XAct seems to have a large range for Pitch values that "normal" XNA does.

dellis1972:
So we need to allow a wider range when dealing with values from XAct

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffectInstance.cs
+++ b/MonoGame.Framework/Audio/SoundEffectInstance.cs
@@ -50,7 +50,8 @@ namespace Microsoft.Xna.Framework.Audio
             get { return _pitch; }
             set
             {
-                if (value < -1.0f || value > 1.0f)
+                // XAct sounds effects don't have pitch limits
+                if (!_isXAct && (value < -1.0f || value > 1.0f))
                     throw new ArgumentOutOfRangeException();
 
                 _pitch = value;


tomspilman:
This is an ok fix for now.

The issue is that XNA put the limit for pitch sort of low...  only up/down one octave.  The underlying XAudio2 API can do up to 11 octaves up or down.  I am fairly sure OpenAL can do more than one octave up or down too.

Longer term we should decide if we should raise the limit for `SoundEffectInstance` in general or make some internal APIs specific for XAct.

dellis1972:
@tomspilman I found another bit which has the issue 

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstance.OpenAL.cs#L65

we should probably remove that code block since we will check when setting the Pitch property. Thoughts?

tomspilman:
Yeah...  I suggest removing that range check.  It serves no purpose as the math works regardless of the pitch passed in.

mrhelmut:
OpenAL range is ]0.0f-Any], with 1.0f being the default value. Every doubling / halving represents an octave and there's virtually no limit. Note that 0.0f is an illegal value.
Some implementations may clamp this range, which makes it harder to know what the real octave range is (but that shouldn't raise AL errors, so it should be safe to assume that it is infinite).

PR #4853, opened 2016-05-14
SetRenderTarget garbage reduction

### LINKED ISSUE #4843: Garbage reduction: GraphicsDevice.SetRenderTarget
Add a `RenderTargetBinding[1]` member to `GraphicsDevice`.  In the implementations of `GraphicsDevice.SetRenderTarget(RenderTarget2D)` and `GraphicsDevice.SetRenderTarget(RenderTargetCube, CubeMap)`, populate the array with 'new RenderTargetBinding` and the parameters passed, then pass that array to the internal call to `SetRenderTargets()`.

This eliminates the garbage generated by the compiler-generated code to handle the `params`-style parameter lists.

MichaelDePiazzi:
Resolves #4843 as per @KonajuGames advice

Also fixes a little typo in `SetRenderTarget(RenderTargetCube)` where it called `SetRenderTarget(RenderTarget2D)` instead of `SetRenderTargets` when the render target was null. It still actually worked, it just meant that it went through another unnecessary method call to get there.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -53,6 +53,7 @@ namespace Microsoft.Xna.Framework.Graphics
 
         private readonly RenderTargetBinding[] _currentRenderTargetBindings = new RenderTargetBinding[4];
         private int _currentRenderTargetCount;
+        private readonly RenderTargetBinding[] _tempRenderTargetBinding = new RenderTargetBinding[1];
 
         internal GraphicsCapabilities GraphicsCapabilities { get; private set; }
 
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -655,17 +656,27 @@ public int RenderTargetCount
...
+            }
             else
-                SetRenderTargets(new RenderTargetBinding(renderTarget, cubeMapFace));
+            {
+                _tempRenderTargetBinding[0] = new RenderTargetBinding(renderTarget, cubeMapFace);
+                SetRenderTargets(_tempRenderTargetBinding);
+            }
         }
 
 		public void SetRenderTargets(params RenderTargetBinding[] renderTargets)
```
tomspilman: So are we sure that passing a `RenderTargetBinding[]` to this keeps it from generating garbage?  It doesn't make a copy of the passed array or anything?

KonajuGames: Correct. It's an array of the correct type already so it doesn't allocate
another array behind the scenes.

MichaelDePiazzi: Sorry, I should have mentioned that I profiled this before and after and confirmed that with these changes SetRenderTarget no longer generates garbage.

tomspilman:
Ok... going to merge this.

PR #4854, opened 2016-05-14
Remove Redundant code for checking Pitch range.

dellis1972:
We check the values that Pitch is being set too in the
setter of the property. There is no need for this additional
check.


tomspilman:
Perfect... thanks @dellis1972 !

PR #4855, opened 2016-05-14
Added support for KeyDown/Up to the GameWindow.

dellis1972:
XNA was missing some Key input features which make
implementing text boxes a bit easier. One of these
new features was the TextInput event on the game
window. However this only gave us access to characters
that were pressed.

This commit adds a new KeyDown/Up events which give us
access to the actual keypress events. This will allow
for better text input control.

Users can now do

	window.KeyDown += delegate (s, e) => {
		if (e.Key == Keys.Back)
			// Handle the Back key
	};

Note this does NOT replace the Keyboard.GetState and it should
only really be used when trying to capture keyboard input for
TextInput.

### DIFF
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -100,8 +100,35 @@ namespace Microsoft.Xna.Framework {
 		/// </summary>
 		/// <remarks>
 		/// This event is only supported on the Windows DirectX, Windows OpenGL and Linux platforms.
-		/// </remarks>
+		/// It should only be used to handle input for Textboxes it does not replace Keyboard.GetState
+        /// </remarks>
 		public event EventHandler<TextInputEventArgs> TextInput;
+        /// <summary>
+        /// Use this event to get notified when a Key is Pressed.
+        /// This event should be used in conjuciton with the TextInput event
+        /// to handle draw Text Input and NOT as a replacement for 
+        /// Keyboard.GetState. The TextInput event is raised for character
+        /// keys only. In order to support keys like Backspace, Tab and Enter
+        /// you will need to use this event. 
+        /// </summary>
+        /// <remarks>
+        /// This event is only supported on the Windows DirectX, Windows OpenGL and Linux platforms.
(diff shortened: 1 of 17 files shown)


dellis1972:
@tomspilman @KonajuGames thoughts on this one?

MonoGame.Framework/GameWindow.cs
```diff
@@ -101,7 +101,9 @@ protected GameWindow()
 		/// <remarks>
 		/// This event is only supported on the Windows DirectX, Windows OpenGL and Linux platforms.
 		/// </remarks>
-		public event EventHandler<TextInputEventArgs> TextInput;
+		public event EventHandler<TextInputEventArgs> TextInput;
+        public event EventHandler<KeyInputEventArgs> KeyDown;
+        public event EventHandler<KeyInputEventArgs> KeyUp;
```
tomspilman: @dellis1972 - Could we write up some docs for these methods.

In particular we should make clear when `TextInput` should be used and when `KeyUp/Down` should be used.

tomspilman:
@KonajuGames - Any feedback on this?

We would need to add support for this to the following platforms ideally before the 3.6 release:

 - Windows desktop DX
 - Windows 10
 - PS4 (yes it supports keyboard and mouse)

Really any platform where we support real keyboard input would need to trigger these events.  Are there any I've missed above?

MonoGame.Framework/KeyInputEventArgs.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
```
KonajuGames: A new file should have the header license block.

KonajuGames:
That should be all the platforms we need to support.  Android and iOS can support keyboards as well, but they are very rarely used, and almost never for games.

dellis1972:
@tomspilman I added some Xml comments on the new Event Handlers. I'm not in a position to do/test the UWP/DX bits atm. iOS/Android can probably wait

tomspilman:
Speaking with @dellis1972 today he is going to make a few more improvements to this.  So please don't merge it yet.

dellis1972:
Messed up the rebase.. closing

PR #4858, opened 2016-05-15
Fix SDL Game Window Resize

dellis1972:
For some reason SDL seems to sometimes report multipe resize
events with the same width and height. This can cause problems
for game if they are doing things like creating render targets
in the event.

So we need to make sure we only call the event when we actually change
the width or height.

### DIFF
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -249,6 +249,12 @@ namespace Microsoft.Xna.Framework
 
         public void ClientResize(int width, int height)
         {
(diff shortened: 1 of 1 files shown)


cra0zy:
Wouldn't it be better to not make this PR considering the fix is already within the SDL2 branch? Or are there plans not to fully replace OpenTK for 3.6?

dellis1972:
I don't think it matters either way. we're going to have to update the SDL2 branch anyway. I can do it there if need be.

cra0zy:
Didn't you already add this to SDL branch?

tomspilman:
Going to merge this... @dellis1972 will deal with any issues in the SDL2 branch of need be.

Issue #1455, opened 2013-02-25
Add parameter validation to SetData() of Texture2D/Texture3D/TextureCube


Helikin:
When the SetData() method of a texture is called, the caller must provide a valid target region and data with the correct number of bytes. A data array which is too small could cause UpdateSubresource() to read data outside the valid bounds.
The SetData() method should check if the number of bytes is correct for the given surface format and target region. If an incorrect number of bytes is specified, an ArgumentException should be thrown. XNA throws: "The size of the data passed in is too large or too small for this resource."

tomspilman:
Let's add unit tests for the texture `SetData()` calls which verify we throw the same exceptions as XNA for the possible argument errors.

Tzenchor:
I have made test for `Texture2D.SetData()` with 1 and 3 parameters with these Results:

- `SetData(T[])`: In XNA it is required that the lenght of the array matches the length of the texture. In Monogame does not matter what length it is, for shorter lenghts it only writes up to what it contains, for longer lenghts it writes the whole texture but can't guarantee that it is not writing beyond the texture length at least it does not crash when I tried with a texture of 4096 pixel and a color array of 2 million pixels, with an array of 2000 pixels only those 2000 pixel are written.
- `SetData(T[], int startIndex, int elementCount)`: It is misusing the `startIndex` parameter, on XNA it makes reference to the starting position of the Color array, while on Monogame it is used for the texture data.

I'm preparing a pull with the tests for these 2 cases and do the reamining `SetData()` overloads later

PR #4862, opened 2016-05-19
Testing Texture2D.SetData

Tzenchor:
Related to #1455.
At this time I have implemented some tests for `Texture2D.SetData()` with 1 and 3 parameters, the other overloads will come later. If someone reads the code, the testCases are grouped from bottom to top as follows: XNA and Monogame passed tests, only Monogame passed tests, and failing tests.

Things pending:

- Add tests for the remaining `SetData` overloads.
- Separate the TestCases if they should pass or fail
- Fix issues found in the Monogame implementation

### DIFF
--- a/MonoGame.Framework/Graphics/Texture2D.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.cs
@@ -133,17 +133,20 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <param name="elementCount"></param>
         public void SetData<T>(int level, int arraySlice, Rectangle? rect, T[] data, int startIndex, int elementCount) where T : struct
         {
+            Rectangle resizedBounds = new Rectangle(0, 0, Math.Max(Bounds.Width >> level, 1), Math.Max(Bounds.Height >> level, 1));
+            if (level >= LevelCount)
+                throw new ArgumentException("Texture only has "+_levelCount+" levels", "level");
             if (data == null)
                 throw new ArgumentNullException("data");
-            if ((!rect.HasValue && (data.Length - startIndex < Bounds.Width * Bounds.Height)) || (rect.HasValue && (rect.Value.Height * rect.Value.Width > data.Length)))
+            if ((!rect.HasValue && (data.Length - startIndex < resizedBounds.Width * resizedBounds.Height)) || (rect.HasValue && (rect.Value.Height * rect.Value.Width > data.Length)))
                 throw new ArgumentException("data array is too small");
             if (elementCount + startIndex > data.Length)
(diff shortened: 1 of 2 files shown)


Tzenchor:
This latest push fixes the possible writing of wrong data to the texture, but it is breaking 4 Texture2D visual tests which i'll have to review to understand them.

The fix I did allow to use longer arrays than the texture but will not write beyond the texture data. Also note that while there is some validation of the parameter `elementCount`, it has no effect on the actual writing of the data to the texture, at least on windows

Tzenchor:
I forgot to add in the previous message the comparison of the test cases between XNA and Monogame.
[Tests setdata.xlsx](https://github.com/mono/MonoGame/files/283961/Tests.setdata.xlsx)
Column "A" contains the tests cases in XNA with its own order, column "F" tells if the test was a success or a failure, Column "H" the tests cases in Monogame, column "L" its results, and column "M" the corresponding XNA result.

Tzenchor:
Found the problem, it occurs in the test that are accessing or texture sublevels or when accesing to values of a texture array

Tzenchor:
Done if I have missed something.
Two assumptions I had to make:
- Each sublevel of a texture has half the size than the upper level. I have made this calculation directly inside `Texture2D.setData` where the parameters get validated but it may be useful to move it to a function.
- Each Texture in a Texture array have the same size as the first texture

KonajuGames:
These look good to me.  Thanks @Tzenchor.  Merging.

nkast:
@Tzenchor 
https://github.com/Tzenchor/MonoGame/blob/e4358c240d7654695232dedf01bf176cdc46fe10/MonoGame.Framework/Graphics/Texture2D.cs#L141-L144
IMO we shouldn't check the `data.Length`. What matters is the elementCount. MG for example uses a sratchbuffer to load textures, user might do the same. 

Also the line `data.Length - startIndex < resizedBounds.Width * resizedBounds.Height` doesn't account for surface format and typeof(T). You need to convert everything into bytes, something like '(elementCount - startIndex)*sizeof(T) < (resizedBounds.Width * resizedBounds.Height)*GetBytesPerPixel(surfaceFormat)'   If you have a Color texture and load from a byte array the actual data would be Width * Height * 4.

The reason most textures pass the test now is because MG uses a scratch buffer of 1M.

Tzenchor:
True, didn't thought about that.

@Jjagg was doing changes to this and more in a recent pull request #5121. Maybe he is aware of the issue and he can solve it inside

Jjagg:
Yeah, that's done right in my PR. I don't know how XNA handles Get/SetData for compressed texture formats. That's what still needs to happen in that PR

PR #4863, opened 2016-05-19
Fixes potential stencil state mismatch

prollin:
When resolving a multisampled rendertarget, we have to disable the stencil test to make sure everything is resolved properly.
The way it was previously done was to directly call ``GL.Disable(EnableCap.ScissorTest)`` and we set ``GraphicsDevice._depthStencilStateDirty = true`` to ensure the current stencil test value was returned to its original value during the next draw call.
Unfortunately this fails if ``GraphicsDevice._lastDepthStencilState.StencilEnable == GraphicsDevice.DepthStencilState.StencilEnable == true`` since MonoGame caches state and makes sure to not submit superfluous OpenGL calls. In that case, the stencil test would not be reset to its original state (the one before the rendertarget gets resolved).

The fix is very simple and explicit: it tests the value of the last **applied** stencil state and only disable it if necessary, and reenable it right after if it was previously enabled.

Despite being the original author of the multisampling code, I have no idea why it wasn't done this way in the first place. Apologies if this sneaky bug caused you as much pain as it did for us :)


tomspilman:
This looks good to me.... I don't see how it could break anything.  Merging!

PR #4865, opened 2016-05-20
Three fixes for basic errors

Linchy:
Some errors I encountered while trying to use the develop build today.

### DIFF
--- a/Tools/Pipeline/Controls/PropertyCells/CellRefs.cs
+++ b/Tools/Pipeline/Controls/PropertyCells/CellRefs.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using Eto.Forms;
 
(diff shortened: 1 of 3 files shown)


Tools/Pipeline/Controls/PropertyGridTable.cs
```diff
@@ -196,7 +196,7 @@ private void Drawable_Paint(object sender, PaintEventArgs e)
             if(_edit)
             {
                 if (!Global.Unix)
-                    _selectedCell.Edit(pixel1);
+                    _selectedCell?.Edit(pixel1);
```
cra0zy: The null check here is pointless, _selectedCell in this part of the code cannot be null because _edit will only get set if _selectedCell is not null.

https://github.com/Linchy/MonoGame/blob/016a5abf3e0651cb9aaa9a81d44dfd807117f180/Tools/Pipeline/Controls/PropertyGridTable.cs#L237

Tools/Pipeline/Controls/PropertyCells/CellRefs.cs
```diff
@@ -3,27 +3,31 @@
...
     {
-        public CellRefs(string category, string name, object value, EventHandler eventHandler) : base(category, name, value, eventHandler)
+        public CellRefs(string category, string name, object value, EventHandler eventHandler)
+            : base(category, name, value, eventHandler)
         {
             if (Value == null)
                 Value = new List<string>();
 
-            DisplayValue = (Value as List<string>).Count > 0 ? "Collection" : "None";
+            DisplayValue = (Value as ICollection).Count > 0 ? "Collection" : "None";
```
cra0zy: Why the change to ICollection?

Tools/Pipeline/Controls/PropertyGridTable.cs
Tzenchor: Also that is a characteristic of C# 6.0, but the build bot is using the C# 5.0 compiler

Tools/Pipeline/Controls/PropertyCells/CellRefs.cs
Linchy: I can't actually manage to replicate the problem now as I've been changing a bunch of things with my project to get stuff working again, but at some point Value was being passed in as a Collection<SkinInfo> somehow, and not having a full understanding of the source I simply changed  it to support collections and it worked.

Tools/Pipeline/Controls/PropertyGridTable.cs
Linchy: @cra0zy a screenshot of it happening: http://imgur.com/HQNNATo

This happens every time I open a `mgcb` file, then click on the root node, scroll down to references,  and double click `collection`.

cra0zy: Oh, I know what the problem is...

cra0zy: #4866

KonajuGames:
In future please keep unrelated changes in separate PRs.  I have merged #4866, so please update this PR.

Tools/Pipeline/Controls/PropertyCells/CellRefs.cs
cra0zy: Undo the List > ICollection change, it seems pointless...

Linchy:
I'll re-open a new PR.

PR #4866, opened 2016-05-20
[Pipeline Tool] Fix null reference exception with Windows propertygrid dialog launching

cra0zy:
The reason why null ref exception would occur is because when a new dialog is created it will pause the current drawing, and create a new redraw event to show the dialog, but the new redraw event will reset `_selectedCell` to `null` since mouse is out of focus, and since the old event didn't finish up `_edit` would still be `true`.


KonajuGames:
Merged.  Thanks @cra0zy

PR #4867, opened 2016-05-21
Fix for "Input string was not in a correct format."

Linchy:
Using `TryParse` solves the problem when name is just `BlendIndicies` for example, in which case we default to index 0, assuming only 1 index.


KonajuGames:
Thanks @Linchy. Merging.

PR #4869, opened 2016-05-22
[SDL] General Improvements

cra0zy:
This is a set of minor improvements, it includes:
 - fix the mouse state for < 2.0.4 SDL version
 - implement resizable function (only available via the custom patch provided by @flibitijibibo, repo for SDL is at https://github.com/FNA-XNA/SDL-mirror)
   - add a version check using that function since the current devel version is reporting 2.0.4 anyway

### DIFF
--- a/MonoGame.Framework/SDL/SDLGamePlatform.cs
+++ b/MonoGame.Framework/SDL/SDLGamePlatform.cs
@@ -39,17 +39,28 @@ namespace Microsoft.Xna.Framework
             Sdl.Minor = sversion.Minor;
             Sdl.Patch = sversion.Patch;
 
+            try
+            {
+                // HACK: The current development version of SDL
+                // returns 2.0.4, to check SDL version we simply
+                // need to try and execute a function that's only
+                // available in the newer version of it.
+                Sdl.Window.SetResizable(IntPtr.Zero, false);
+                Sdl.Patch = 5;
+            }
+            catch { }
+
             var version = 100 * Sdl.Major + 10 * Sdl.Minor + Sdl.Patch;
 
             if (version <= 204)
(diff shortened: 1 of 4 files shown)


tomspilman:
@dellis1972 - This look ok to you?

I assume the SDL team will be merging these patches sometime soon?

cra0zy:
> I assume the SDL team will be merging these patches sometime soon?

Duno when they will get merged, we still need them tho.

mrhelmut:
I'm favorable to use this fork, it contains important fixes and I don't see why they wouldn't be merged upstream, but it seems that there's no schedule for the next SDL release.

A Windows 64bit build would be very welcome.

cra0zy:
Ah, thanks for reminding me, I was gonna fix that as well.

MonoGame.Framework/SDL/SDLGamePlatform.cs
```diff
@@ -39,17 +39,24 @@ public SdlGamePlatform(Game game)
             Sdl.Minor = sversion.Minor;
             Sdl.Patch = sversion.Patch;
 
+            try
+            {
+                Sdl.Window.SetResizable(IntPtr.Zero, false);
+                Sdl.Patch = 5;
```
dellis1972: @cra0zy can we add a comment here that this is a "hack" to work around people compiling SDL2 themselves and not updating the version number. Ideally this will be removed at some point once we get a release with the fixes in.

cra0zy: Will do as soon as I get home.

cra0zy:
@mrhelmut @dellis1972 Updated.

tomspilman:
Thanks @cra0zy ... merging!

PR #4874, opened 2016-05-24
Replace OpenTK with custom bindings for DesktopGL

cra0zy:
Cleaned out version of #4555

CC. @dellis1972 @mrhelmut


cra0zy:
I couldn't get the old branch rebased/squashed, too many problems, so I remade commit by:
 1. Copying over MonoGame.Framework folder and committing as @dellis1972 
 2. Copying over Build folder and committing as myself
 3. Rebasing

mrhelmut:
I'm having difficulties testing different SDL builds. The prebuilt dlls (both 32/64) from https://www.libsdl.org/download-2.0.php are working fine, but if I build the 64bit SDL from XNA-FNA/SDL-mirror using Visual Studio  2012/2013, or grab the 32bit @flibitijibibo redist, it crashes upon ```SDL.Init()```. Is there anything I'm missing? (vc redist? should I go with MingW?)

I'm using the prebuilt 2.0.4 in the meantime to check what's happening with GL resources disposal in ```UnloadContent```. Once I sort this out, I'll have a small commit for Windows and it should be good to go.

flibitijibibo:
For reference, the fnalibs binary is built with Visual Studio 2010 - that said, SDL doesn't link to the C runtime and I've built Hg binaries for fnalibs before, so it may be a real bug - is it possible to get native traces in newer VS debuggers?

You can try building with this sln, shouldn't require anything special to build:

https://github.com/FNA-XNA/SDL-mirror/tree/master/VisualC

mrhelmut:
Never mind, a reboot and rebuild did the trick... VS2013 builds work. Most probably an unrecoverable graphics driver crash.

Resizable / window centering and mouse position are fine.

Now to that dispose issue.

dellis1972:
@mrhelmut <https://github.com/mrhelmut> its possible that it needs a
RunOnUIThread around the dispose... either that or the bindings I wrote
have a bug in the parameter list..

flibitijibibo:
The crash @mrhelmut hit has been fixed by reverting this commit, will try to get this in mainline soon https://hg.libsdl.org/SDL/rev/96f7be2c885b

mrhelmut:
Thanks @flibitijibibo !

@dellis1972 ```DeleteRenderbuffer``` and ```DeleteGLTexture``` are correctly called though ```BlockOnUIThread```. Parameters look good, calling ```RenderTarget.Dispose()``` or ```Texture.Dispose()``` before ```Game.UnloadContent()``` works.

Adding a ```CheckGLError()``` before the texture/renderbuffer deletion raise the error, so it must be a previous GL call. I'm tracking this down.

mrhelmut:
I've narrowed this down to ```SDLGameWindow.Dispose(bool)```, ```Sdl.Windows.Destroy(_handle)``` raises a GL error (code 1282, INVALID_OPERATION).

cra0zy:
> I've narrowed this down to SDLGameWindow.Dispose(bool), Sdl.Windows.Destroy(_handle) raises a GL error (code 1282, INVALID_OPERATION).

That makes sense, try commenting out Sdl.Windows.Destroy, will the window get autoclosed once all the threads are dead?

mrhelmut:
It postpones the error on ```Sdl.Quit()``` in ```SDLGamePlatform.Dispose(bool)```.

cra0zy:
> It postpones the error on Sdl.Quit() in SDLGamePlatform.Dispose(bool).

Comment it out as well.

mrhelmut:
Working, without any GL error and the window is closing correctly.

dellis1972:
@mrhelmut that on Windows/Linux or OSX?

mrhelmut:
Only tested on Windows.

cra0zy:
> Working, without any GL error and the window is closing correctly.

So thinking of possible negative side effects I'm guessing that the resolution will not get restored now? That is try setting some resolution with fullscreen and exit.

dellis1972:
It will need to be called at some point.. just after all the other bits
have been done.

cra0zy:
In case I am correct we can just exit fullscreen on dispose.

mrhelmut:
Can't this line be removed? 

https://github.com/cra0zy/MonoGame/blob/e762911d9c9ffc617baf1755902ad61b1b2a2679/MonoGame.Framework/SDL/SDLGamePlatform.cs#L110

```SDLGamePlatform.Dispose()``` would then be called at the end by ```Game.Dispose()``` to close sdl systems? (https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Game.cs#L134)

It seems to solve the problem with resource disposal, but then there's a GL error on ```Sdl.GL.DeleteContext(_context)``` in ```GraphicsContext.SDL.Dispose()```, which is called just before. Though there's no blocking ```CheckGLError``` there, I don't know how much important this error is.

**Comment by cra0zy** [comment]
> Can't this line be removed?
> 
> https://github.com/cra0zy/MonoGame/blob/e762911d9c9ffc617baf1755902ad61b1b2a2679/MonoGame.Framework/SDL/SDLGamePlatform.cs#L110

Oh yea, it can, it was there as a workaround for OpenTK...

**Comment by mrhelmut** [comment]
Then we're good to go. :) I'll push a commit with this and moving the GL entry points loading.

There's a silent GL error on ```Sdl.GL.DeleteContext(_context)```, should we care?

**Comment by dellis1972** [comment]
I think we should at least try to find out what is causing it :)

**Comment by mrhelmut** [comment]
I agree. It seems that SDL init /exit and context destroy can generate GL errors, which can be ignored if SDL_GetError() is empty, but I can't find any reliable documentation about this...
The only relevant GL calls SDL is doing here are ```wglMakeCurrent(null, null)``` and then ```wglDeleteContext``` (Win API).
The SDL/GL lifecycle seems fine.

Would you mind testing a ```CheckGLError()``` after ```Sdl.GL.DeleteContext(_context)``` on your respective platforms? It may be a side effect of the Windows API only.

**Comment by mrhelmut** [comment]
Turns out it is "normal" that glGetError returns INVALID_OPERATION after the context destruction. The Windows API will always return this if there is no context.

I'm favorable to merge this PR, awesome work.

**Comment by cra0zy** [comment]
@mrhelmut Can you submit a PR to MG.Deps for SDL? Since you already the x64 binary for Windows and the rest of them you can grab from fnalibs

**Comment by mrhelmut** [comment]
Sure, that was planned all along. I'll push SDL & OpenAL-Soft x86/x64. After that I'll add the code to link to the correct dlls and switch the AL wrapper to OpenAL-Soft (just a matter of renaming the dll).

**Comment by cra0zy** [comment]
@mrhelmut Actually submit only the Windows/Mac ones.

**Comment by cra0zy** [comment]
I was testing some stuff and the fnalibs don't work with Wayland for some reason O o. On the bright side with system SDL I got my game running on Wayland :D

**Comment by dellis1972** [comment]
I have the Universal Binary for OSX for SDL2 with the version updated to
2.0.5 I'll submit that too

**Comment by mrhelmut** [comment]
I have added a few checks to strengthen the frame buffer object extension loading in case they are absent or a GL 3.0+ context is available (in which case this extension is guaranteed to be in the core functions). There was also a EXT_frame_buffer_object function which was not loaded and it now throws a proper ```PlatformNotSupportedException``` if entry points loading fails.

**Comment by dellis1972** [comment]
@cra0zy looks like we have a merge conflict :(

**Comment by cra0zy** [comment]
> @cra0zy looks like we have a merge conflict :(

I don't see any merge conflicts, it must have been you imagination :P

**Comment by dellis1972** [comment]
@cra0zy weird.. Github showed a "cannot merge this " message last time I looked.

**Comment by cra0zy** [comment]
I was joking, I got them resolved.

**Comment by mrhelmut** [comment]
Is there anything else to do? Linux / OS X installer ? SDL2 binaries with Wayland support?

From the Windows perspective, this PR is complete.

The "Package Windows" agent on the buildbot has been disconnected. Is there anything failing? @tomspilman

**Comment by tomspilman** [comment]
> The windows agents on the buildbot have been disconnected. Is there anything failing?

Looks like a power failure cause them to go offline.  I just fixed this now.

Still looks like we have a bug in the packaging....
```
Attempting to build package from 'MonoGame.Framework.DesktopGL.nuspec'.
File not found: 'WindowsGL\AnyCPU\Release\NVorbis.dll'
```
@cra0zy - Are we using `NVorbis.dll` now?  It is in the nuspec file...

https://github.com/mono/MonoGame/blob/develop/NuGetPackages/MonoGame.Framework.DesktopGL.nuspec

... but it doesn't exist in the build output.

**Comment by cra0zy** [comment]
> @cra0zy - Are we using NVorbis.dll now? It is in the nuspec file...
> 
> https://github.com/mono/MonoGame/blob/develop/NuGetPackages/MonoGame.Framework.DesktopGL.nuspec
> 
> ... but it doesn't exist in the build output.

We are no longer using NVorbis.dll file.

**Comment by tomspilman** [comment]
> We are no longer using NVorbis.dll file.

@cra0zy - Do you mind submitting a PR that fixes these nuspecs...

https://github.com/mono/MonoGame/blob/develop/NuGetPackages/MonoGame.Framework.DesktopGL.nuspec
https://github.com/mono/MonoGame/blob/develop/NuGetPackages/MonoGame.Framework.WindowsGL.nuspec
https://github.com/mono/MonoGame/blob/develop/NuGetPackages/MonoGame.Framework.Linux.nuspec

It has also made it to the template:

https://github.com/mono/MonoGame/blob/develop/ProjectTemplates/VisualStudio2010/DesktopGL/MonoGameDesktopGLApplication.csproj

@dellis1972 - Also looks like there is reference to it in the XS addin:

https://github.com/mono/MonoGame/blob/develop/IDE/MonoDevelop/MonoDevelop.MonoGame/MonoGameMSBuildImports.cs
https://github.com/mono/MonoGame/blob/develop/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/MonoGameProject.xpt.xml

**Comment by cra0zy** [comment]
Can I just append a commit to this?

**Comment by tomspilman** [comment]
> Can I just append a commit to this?

I prefer if we didn't as the NVorbis issue is unrelated to this PR.

However if you are cleaning out OpenTK junk from the templates/nuspecs at the same time I could live with it.

**Comment by mrhelmut** [comment]
Mind that the template already removes NVorbis in this PR and may conflict.

I believe that we can clean this up here. This PR is in a pretty good shape and should be ok once the nuspecs, XS add-in & Linux/OS X installers are up-to-date.

**Comment by cra0zy** [comment]
One more thing, can I switch out the mechanism for config files. Currently config files are just added, and with nuget package their action can't be set to "copy if newer" so the user has to set it. What I would like to do is add a .targets file and copy the .config file with it, like I did for gtk-sharp nuget package: https://github.com/cra0zy/gtk-sharp3-nuget/blob/master/gtk-sharp3.targets

**Comment by cra0zy** [comment]
Now that I think about it I should do the above for native libs as well.

**Comment by mrhelmut** [comment]
It isn't, when discussing it in #4399, @tomspilman showed a preference to having them in the template, for transparency purpose.

After giving it some thoughts, I think that targets would be more robust (and would make projects portable if one copies them to a machine with a different MonoGame installation path,  if this ever get supported).

**Comment by cra0zy** [comment]
Templates don't use NuGet packages.

**Comment by mrhelmut** [comment]
My bad, thought you were talking about changing that for the template too.

**Comment by mrhelmut** [comment]
I have updated the nuspec and XS add-in on my local branch of this PR: https://github.com/mrhelmut/MonoGame/commit/787c56d7a9bc9a47393098dbfa72821d99b54da8
It also removes oalinst.exe from the Windows installer (no more needed and misleading).

Let me know if it looks fine and if I can push it here.

File: MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs [hunk 1]
```diff
@@ -178,12 +200,27 @@ private void PlatformSetup()
             GL.GetInteger(GetPName.MaxTextureSize, out _maxTextureSize);
             GraphicsExtensions.CheckGLError();
 
-            SpriteBatch.NeedsHalfPixelOffset = true;
```
dellis1972: @cra0zy Opps. We kinda NEED this :) I think we should put that back in

tomspilman: Yes... this is critical to correct rendering on all GL platforms.

mrhelmut:
Is there anything I can do to help get this PR merged?
Should I revert ```SpriteBatch.NeedsHalfPixelOffset = true;``` and push the nuget/XS add-in update (https://github.com/mrhelmut/MonoGame/commit/787c56d7a9bc9a47393098dbfa72821d99b54da8) here?

This has been a big rework that I would really like to see merged, it fixes so many issues and make things way more future-proof.

KonajuGames:
The half pixel offset will need to be reinstated.

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
cra0zy: I'm not the one that removed it, duno why you mentioned me specifically...

cra0zy:
Just FYI, it's my exam period and that's why I was/am not around, I'll have time to work on stuff in 2 days, but I'll only have around a week of time to get stuff done, so do give me any tasks you want me to do for 3.6 now.

cra0zy:
@dellis1972 This PR uses `NewSDL` branch, not the `SDL2` branch, you pushed a commit to the wrong branch.

mrhelmut:
@dellis1972 I fixed the DesktopGL nuspec but it looks like the MacOS nuspec is failing because OpenTK can't be found in the output. Any chance you could take a look at this?

dellis1972:
@mrhelmut I'll check the nuget package. 

@cra0zy @mrhelmut either of you tried to use the Resizable window? It doesn't seem to work, the window is locked at the PreferedBackBufferWidth/Height it seems

cra0zy:
> @cra0zy @mrhelmut either of you tried to use the Resizable window? It doesn't seem to work, the window is locked at the PreferedBackBufferWidth/Height it seems

I tried using it, and it worked before... something must have gotten broken in between.

dellis1972:
@cra0zy looks like it was my test project :)

endimilojkoski:
Just tried the latest version with my game and it seems to be working pretty good with added SDL2 32bit downloaded from the official site.

One issue I noticed is when I add gamepad in game it crashes:

"An unhandled exception of type 'System.ArgumentException' occurred in mscorlib.dll
Additional information: An item with the same key has already been added."

mrhelmut:
Need to check that, thanks Endi.

endimilojkoski:
Also, I sent the game to a friend with 4K monitor and the game runs with no issues in Full Screen and in Windowed Mode! He is on Windows 10.

dellis1972:
@mrhelmut @cra0zy green build guys :)

cra0zy:
I know how the gamepad crash can be fixed, will do it tomorrow morning.

tomspilman:
Ok...  @cra0zy @mrhelmut ... is this good to merge then?

cra0zy:
> Ok... @cra0zy @mrhelmut ... is this good to merge then?

It will be in 1 day.

mrhelmut:
Once the gamepad issue is sorted out (was about to look at this), it's all green on my Windows side!

mrhelmut:
I can confirm the gamepad issue, but it is unrelated to this PR. I have therefore opened #4911 and, to me, this PR is good to go.

cra0zy:
I can't cause the error, tho the above commit should fix it.

cra0zy:
Also do check if the gamepads are assigned proper ids, we may need to add one more dictionary for SDLInstanceID <=> FakeID.

mrhelmut:
> I can't cause the error, tho the above commit should fix it.

I doesn't, same error. The problem is that ```Which``` is always 0, no matter the device plugged.

cra0zy:
Have you only tested it on Windows?

cra0zy:
Also it's possible for 2 devices to get deviceid 0, deviceid is not unique to a device so it can be changed, but instanceid is unique.

Joystick event gives deviceid when device is connected, and gives instanceid when a device gets disconnected.

mrhelmut:
Yes. For as many gamepad I have plugged in, DeviceID, JoystickPtr & InstanceID are the very same. ```AddDevice(int)``` is always called with a 0 deviceId.

cra0zy:
Maybe a Windows bug then, I have 2 gamepads and I couldn't get the bug to occur.

mrhelmut:
Looks like a problem with ```Sdl.PeepEvents```, it doesn't populate ```events```. I'm trying to figure that out.

mrhelmut:
While taking a look at this, I think that we shouldn't add joysticks to gamepads if they don't pass the ```IsGameController``` test. I've had a lot of players with HOTAS devices or gaming mouses recognized as joysticks with weird capabilities which make GamePad to crash because some buttons/hats are not there. Unknown gamepads should be avoided.

cra0zy:
https://wiki.libsdl.org/SDL_ControllerDeviceEvent

mrhelmut:
Tried that, same problem (events detected, but struct not populated). Could this be a problem with the ```Event``` struct marshaling?

cra0zy:
You misunderstood why I posted the link, it was to suggest we could use it for detecting gamepads so that gaming mouse/HOTAS devices get skipped.

mrhelmut:
Problem found, it was the marshalling.

```
public static extern int PeepEvents(
			[Out()] [MarshalAs(UnmanagedType.LPArray, ArraySubType = UnmanagedType.Struct, SizeParamIndex = 1)]
				Event[] events,
			int numevents,
			EventAction action,
			EventType minType,
			EventType maxType
		);
```

I'm going to fix this, and a couple other minor things (the controller detection and the fact that gamepad indexes are inverted).

cra0zy:
It's always marshalling :D Good job!!

mrhelmut:
Here you go:
- marshalling fix
- prevented non compatible joysticks to be added to gamepad
- fixed the player index order (DeviceID returns the good order, but InstanceID is inverted, so I added a mapping to keep track of the instance)

dellis1972:
@tomspilman @KonajuGames I think we an pull the trigger on this one.

tomspilman:
Alright... great job guys.... goodbye OpenTK on desktops (still need to work on purging it from mobile and the Angle build).

mrhelmut:
Awesome work @cra0zy & @dellis1972 ! What a relief.

MichaelDePiazzi:
Nice one! Great work guys! :)

KonajuGames:
This is probably one of the biggest functional changes we've had in a while.  Definitely a team effort in getting this one together and working.  Good work.

PR #4878, opened 2016-05-25
Added support for getting "Special" Keys like Backspace.

dellis1972:
The previous implementation of TextInput did not allow for
special keys to be processed. As a result we missed those keys
which makes it harder to implement things like textboxes.
(Backspace is kinda important).

### DIFF
--- a/MonoGame.Framework/SDL/SDLGamePlatform.cs
+++ b/MonoGame.Framework/SDL/SDLGamePlatform.cs
@@ -135,11 +135,13 @@ namespace Microsoft.Xna.Framework
                     Joystick.RemoveDevice(ev.JoystickDevice.Which);                
                 else if (ev.Type == Sdl.EventType.MouseWheel)
                     Mouse.ScrollY += ev.Wheel.Y * 120;
-                else if (ev.Type == Sdl.EventType.KeyDown)
-                {
-                    var key = KeyboardUtil.ToXna(ev.Key.Keysym.Sym);
-                    if (!_keys.Contains(key))
-                        _keys.Add(key);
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/SDL/SDLGamePlatform.cs
```diff
@@ -125,6 +125,7 @@ private void SdlRunLoop()
 
                     if (!_keys.Contains(key))
                         _keys.Add(key);
+                    _view.CallTextInput ((char)ev.Key.Keysym.Sym, key);
```
tomspilman: What keeps you from getting two events for the same key press?  One from this `Sdl.EventType.KeyDown` and one from `Sdl.EventType.TextInput` below?

dellis1972: I just removed the one from TextInput

dellis1972:
@tomspilman any more comments? //cc @cra0zy @mrhelmut

cra0zy:
Very useful, when I was adding textinput I mirrored the OpenTK behavior, but this is better.

dellis1972:
@tomspilman good to go I think

tomspilman:
So speaking to @dellis1972 this change is just an intermediate improvement to what we have already.  The reality is that we don't handle text input like this very well.  If you look at the docs for SDL2's text input you can see all the cases we don't handle...

https://wiki.libsdl.org/Tutorials/TextInput

We need to rethink our approach eventually to deal with all this, but for now this is an improvement.

Will follow this up with a PR that fixes Windows DX platforms to support the new Key info too.

PR #4879, opened 2016-05-25
Fixed the DisplayMode Ordering so it orders the display modes from smallest to highest.

dellis1972:
XNA seems to report the DisplayModes in size order from small to large.
So the largest is at the end of the list. This commit makes sure that
we do the same.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.cs
@@ -326,6 +326,11 @@ namespace Microsoft.Xna.Framework.Graphics
                     adapter.Dispose();
                     dxgiFactory.Dispose();
 #endif
+                    modes.Sort (delegate (DisplayMode a, DisplayMode b) {
+                        if (a == b) return 0;
+                        if (a.Format <= b.Format && a.Width <= b.Width && a.Height <= b.Height) return -1;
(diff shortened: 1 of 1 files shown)


tomspilman:
So what I found indicates XNA first sorts by the `SurfaceFormat` then the width then height.

capture

dellis1972:
@tomspilman updated to include the Format

tomspilman:
Thanks... merging!

theZMan:
The word from 'someone who would know' is that you should not assume any sort order on this property. The order is platform/implementation/who-knows-what-other-factors dependent.

So if you require a sort order, sort it yourself.

We should probably update our docs to reflect this.

tomspilman:
Interesting...  so the result is totally dependent on drivers.

Still I think us sorting it to something consistent isn't a bad improvement and something we could easily control.  Unless not sorting and letting the underlying driver/os results appear as is has some benefit i'm not thinking of.

theZMan:
There's no downside - anyone assuming a sort order in XNA is already potentially broken based on our current observations.

But every developer is going to have their own preference.. what is most important, format, horizontal res, vertical res, total pixels?

I would say the most useful would be to present a list of resolutions in a game - how do most games sort? Biggest first/last? By overall resolution?

PR #4880, opened 2016-05-25
List only the displaymodes for the current display.

dellis1972:
In the current implementation of the SDL back end we list ALL
the displaymodes for evey display. I need to check with XNA but
it seems we should only list the modes for the current display. Otherwise
you might end up trying to set a mode which isn't supported.

Anyone with XNA able to confirm the DisplayMode list is for the "current" display only.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.cs
@@ -61,6 +61,8 @@ namespace Microsoft.Xna.Framework.Graphics
         {
             _screen = screen;
         }
+#elif DESKTOPGL
+        int _displayIndex;
 #else
         internal GraphicsAdapter()
         {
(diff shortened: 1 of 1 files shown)


tomspilman:
> Anyone with XNA able to confirm the DisplayMode list is for the "current" display only.

I'll give it a check now.

tomspilman:
So looking at the `GraphicsAdapter.Adapters` list I have one per monitor:
adapters
That to me pretty clearly indicates that the `GraphicsAdapter.SupportedDisplayModes` is clearly a list of modes supported on only that one display.

dellis1972:
cool thanks @tomspilman

dellis1972:
@tomspilman is this ok to merge?

tomspilman:
I guess this is at least an improvement on what we had... merging!

PR #4882, opened 2016-05-27
Fix GetBufferData for Index- and VertexBuffer on OpenGL

Jjagg:
Binding was wrong in IndexBuffer and both did not take startIndex and elementCount into account in all cases.


tomspilman:
@dellis1972 @cra0zy - Will this get replaced with the SDL2 branch?  Should we merge it anyway until then?

tomspilman:
I'm good with this... it will be validated by the unit tests soon.

Issue #4872, opened 2016-05-23
Use xwt for ProjectControl?


cra0zy:
As you might have noticed ProjectControl is really dirty:
 - https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Controls/ProjectControl.win.cs
 - https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Controls/ProjectControl.gtk.cs
 - https://github.com/mono/MonoGame/blob/develop/Tools/Pipeline/Controls/ProjectControl.mac.cs

This is because Eto.Forms doesn't have a good multiselect treeview control. Now instead of using the one from Eto.Forms we could use the Xwt one. This would allow us to:
 - have same code for this control between platforms
 - gain same drag and drop code, Eto.Forms doesn't support drag and drop, Xwt does

Negative side effects?
 - Pipeline Tool for Windows will need to use WPF since Xwt doesn't support Windows Forms

I really wish that I thought of this idea earlier... at least I didn't start writing the new DnD code yet.

cc. @tomspilman @KonajuGames @dellis1972

tomspilman:
So I am a little concerned about this.

We've taken a big risk in throwing away the UI we already had... ending up with something less functionally useful that we had before...  to move to Eto.  Now we're looking to add some bits of Xwt which we decided was a bad choice before.

We *really* should stick with Eto and help move it forward where we can.  We've been disruptive enough breaking features of the pipeline tool that we shouldn't do that again without an extremely good reason.

So what is the long term plan if we add Xwt specifically for the treeview feature?  Is drag and drop and an improved tree view on the roadmap for Eto?

cra0zy:
Eto.Forms has a multiselect treeview component, which is called TreeGridView. This component however is a bit outdated in design and doesn't allow dynamic updating, which means in order to add/remove just 1 element it needs to refresh all the elements.

> We've taken a big risk in throwing away the UI we already had... ending up with something less functionally useful that we had before... to move to Eto. Now we're looking to add some bits of Xwt which we decided was a bad choice before.

The only lost functionality is drag and drop and multiselect (The propertygrid that is used right now is more advanced than the one Gtk Pipeline had). Also there are more things that are properly synced right now like "Exclude from Project|" (which Win Pipeline was missing) or undoing/redoing property grid changes (which Gtk Pipeline was missing).

[...]

PR #4885, opened 2016-05-27
[Pipeline Tool] Use Wpf backend for Windows

cra0zy:
NEEDS: https://github.com/MonoGame/MonoGame.Dependencies/pull/86

Originally I was doing this for #4872 but turns out regardless of that issue, Wpf backend is better then the WinForms one. 

Advantages that I like:
 - color selection has alpha
 - treeview icons appear full color

Temporary multiselect won't work, this will be fixed with #4872 or native Wpf implementation.

Screenshot:
wpf


KonajuGames:
https://github.com/MonoGame/MonoGame.Dependencies/pull/86 is merged.

tomspilman:
I'm hesitant to all this...  but let me review the current state of the pipeline this week and i'll see about merging this.

cra0zy:
> I'm hesitant to all this... but let me review the current state of the pipeline this week and i'll see about merging this.

This change is not really a big deal, I'm just swapping Eto.WinForms for Eto.Wpf dll for the most part.

tomspilman:
> This change is not really a big deal

You change is not big...  but swapping WinForms for WPF is.  Totally different technologies, requirements, etc.

cra0zy:
Just gonna mention that I will be doing a similar thing to Mac after this PR, that is I'll swap Gtk for Xamarin.Mac which looks (I would say "way more", but Gtk does not look native on Windows/Mac) native: https://files.gitter.im/mono/MonoGame/Ox4S/nativemactool.png

cra0zy:
So... what's the status on this?

tomspilman:
@cra0zy 

So tried the latest Pipeline Tool and things are ok for now.  We still need a lot of work on the property grid, but at least we're fully in control of it and it is cross platform.

Going to test the last build on this branch and see how WPF feels.

And this branch needs a rebase before we can merge.

tomspilman:
A few little things I noticed on the WPF build...

 - The icons of disabled entries in the menus do not grey out.
 - The WPF build takes 2x the memory of the old WinForms build.
 - The References Editor issues I reported (https://github.com/mono/MonoGame/issues/4912) don't occur under WPF.
 - The bottom-right resize grip renders over the Cancel button in the References Editor.

Still none of this is a big issue.

I'm fine with switching over to WPF from WinForms.  If we happen to run into issues with it later we can always consider switching back.

@KonajuGames ?

KonajuGames:
After regarding I'm good with merging this. There are some issues to be fixed, but we they don't so functionality and we can concentrate on those after merge.

KonajuGames:
Rebasing, not regarding. The phone's swipe keyboard doesn't know that word yet.

cra0zy:
Something is wrong with Mac build server.

KonajuGames:
Yes. An upgrade to latest stable didn't go well. Broke Nant.

tomspilman:
@dellis1972 should be around soon and should be able to direct me to the right fix to the Mac build machine.

dellis1972:
@tomspilman ok I think I fixed NAnt. Hopefully I did it in a way we won't hit this again.

cra0zy:
Rebased.

tomspilman:
Thanks @cra0zy !

hach-que:
This change broke building the Windows platform under Linux.  Is support for that something that was intentionally dropped?

On Linux I get:

```
Global.Windows.cs(13,28): error CS0234: The type or namespace name `Imaging' does not exist in the namespace `System.Windows.Media'. Are you missing an assembly reference?
Global.Windows.cs(40,24): error CS0246: The type or namespace name `BitmapSource' could not be found. Are you missing an assembly reference?
```

KonajuGames:
Building the Windows toolchain on Linux is not something we test. I think you might be the only person doing so. It wasn't an intentional break, but wouldn't be something that would be cause to revert the change.

cra0zy:
Mono has a replacement for Windows Forms, but doesn't have a replacement for WPF.

hach-que:
I don't use the output of the build, I just use it to make sure MonoGame builds successfully on all platforms.

It just means updating downstream projects to not attempt to build Windows on Linux.

PR #4888, opened 2016-05-28
[SDL] Fix event struct size

cra0zy:
Huge thanks to @RHY3756547  who noticed it in http://community.monogame.net/t/sdl2-backend-corrupts-memory-after-sdl-pollevent/7704

@dellis1972 @mrhelmut This fixes the problem of running 32 bit SDL and exe on 64 bit host :)

### DIFF
--- a/MonoGame.Framework/SDL/SDL2.cs
+++ b/MonoGame.Framework/SDL/SDL2.cs
@@ -50,7 +50,7 @@ internal static class Sdl
         MouseWheel = 0x403,
     }
 
-    [StructLayout(LayoutKind.Explicit)]
+    [StructLayout(LayoutKind.Explicit, Size = 56)]
     public struct Event
     {
         [FieldOffset(0)] public EventType Type;


mrhelmut:
Ha! Nice catch!
What should we do with the dependencies in the template then? Stick with 32bit binaries for Windows and remove the bitness detection?

KonajuGames:
I believe we need to keep trying to move forward.  Supporting 32-bit only, especially when 64-bit is the larger market now and other platforms are also predominantly or becoming 64-bit, is a bit backwards.

@cra0zy, 24*8 is the correct maximum size for that struct?

cra0zy:
> @cra0zy, 24*8 is the correct maximum size for that struct?

Yes, I checked SDL_Event struct size from c and it was 24 bytes.

KonajuGames:
> Yes, I checked SDL_Event struct size from c and it was 24 bytes.

Now I'm confused.  @RHY3756547 used 56 bytes, you've used 192 bytes (24 * 8) in this PR and you say the SDL_Event struct size in C is 24 bytes.

cra0zy:
> Now I'm confused. @RHY3756547 used 56 bytes, you've used 192 bytes (24 * 8) in this PR and you say the SDL_Event struct size in C is 24 bytes.

Our C# sturct is only partially implemented, that is it doesn't have all the variables, it has a total of 7, 7 * 8 = 56, but in reality there are 24 variables: https://wiki.libsdl.org/SDL_Event

KonajuGames:
The total number of members of the SDL_Event struct in C isn't what decides the total size of the union.  It's the largest of those structs that are members of the SDL_Event union that determines the size.  Is sizeof(SDL_Event) == 192 in C?  The thing that confusing me is why Size is specified as 24 * 8 when that does not appear to be a natural definition of the total size of the union.

cra0zy:
sizeof returns size in bytes, so sizeof(SDL_Event) = 24, C# StructLayout Size is defined in bits so I just had to multiply it by 8. 

I checked the sizeof for the size of the sturct, not the wiki page, I was just explaining why he made a mistake.

KonajuGames:
The [Size](https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.structlayoutattribute.size(v=vs.110).aspx) property of StructLayout is in bytes.  Structure sizes are never specified in bits.

KonajuGames:
And your largest event structure is much larger than 24 bytes.  For example, TextEditingEvent has a 32 byte array in it.

cra0zy:
ss

I don't know anymore, I didn't look at the Size wiki page, but I did try the code with 24 and it didn't work so I assumed its in bits. Time to read up on stuff.

KonajuGames:
TextEditingEvent in SDL.cs is 52 bytes (32 byte array plus 5 integer fields).

You will see that SDL [forces the size to be 56 bytes](http://hg.libsdl.org/SDL/file/7cbfd97f1430/include/SDL_events.h#l560), which will be why 56 was chosen in the original post.  I have no idea why your C test is returning 24 bytes.

Change this PR to `Size = 56` and I'll be happy to merge it.  As it stands it is not correct.

KonajuGames:
Thanks @cra0zy.  Once it has passed tests I'll merge it.

KonajuGames:
Merging.

PR #4892, opened 2016-05-30
Move LZX decompress code from ContentManager into LzxDecoderStream

nkast:
Added LzxDecoderStream similar to Lz4DecoderStream and make ContentManager a bit cleanner.
Related to https://github.com/mono/MonoGame/pull/4522


KonajuGames:
Why did the tests not run on this?  Do you want to do an update to this branch, and hopefully that will kick off the tests.

tomspilman:
@KonajuGames - Looks like it is ok now...  I think when it came in the build machine was down or something.

KonajuGames:
Ok. Merging.

nkast:
Thanks!

PR #4894, opened 2016-05-30
Implement EffectMaterialWriter

### LINKED ISSUE #4884: ExternalReference issue when loading fbx
To test #4875 I wanted to try the [InstancedModel Sample](http://xbox.create.msdn.com/en-US/education/catalog/sample/mesh_instancing), but I ran into some issues importing a .fbx model. 

- First there was a problem in the ContentWriter about writing a non-.xnb file. This was caused by [EffectMaterialContent.Effect](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Graphics/EffectMaterialContent.cs#L14) being serialized, since it's an external reference to a non-xnb file. This was an easy fix, I solved it by adding [ContentSerializerIgnore] above it. 

- Second, there was an error thrown [here](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ContentWriter.cs#L296) because the filename of the reference did not begin with the root directory, which was caused by [NormalizeOS](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/ExternalReference.cs#L63) in ExternalReference, since it changed directory separators in the ExternalReference so the strings didn't match up. Also an easy fix! I Changed that so it does a regular normalize rather than NormalizeOS (do the separators really matter as long as they're consistent? NormalizeOS is now not used anymore).

- After that it built fine! However when loading the model it can't find the content reader for ExternalReference<CompiledEffectContent>. This is the hard one :p After digging through the code a bit, I think it should resolve to a regular ExternalReferenceReader, so it can load the file that's referenced (the reader for the effect itself is loaded, so that would work), but I got a bit confused when I saw the content readers include

        {[Microsoft.Xna.Framework.Content.Pipeline.NamedValueDictionary`1[Microsoft.Xna.Framework.Content.Pipeline.ExternalReference`1[Microsoft.Xna.Framework.Content.Pipeline.Graphics.TextureContent]] 
It seems weird to have a reader for a specific generic type like that, I thought a generic reader should read its content and delegate to the reader that can handle it. In this case I would expect a NamedValueDictionaryReader, an ExternalReferenceReader and a TextureContentReader to handle a type like this, each delegating internally to the next. So my question is if I should make the writer when asked for the runtime reader write out the full name of the type (e.g. ExternalReferenceReader`1[[sometype]]) or make [ContentTypeReaderManager.GetTypeReader](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentTypeReaderManager.cs#L59) a bit smarter to use a reader for the outer type when it encounters generics. I really wouldn't go for the first option, but I don't know if the second one is applicable in general (for all generic types). I think we could make it work though, and it would be **a lot** neater.

Jjagg:
This implements EffectMaterialWriter and does some other little fixes so we can [render a model with a custom effect](https://blogs.msdn.microsoft.com/shawnhar/2006/12/07/rendering-a-model-with-a-custom-effect/). Fixes #4884.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/EffectMaterialContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/EffectMaterialContent.cs
@@ -8,7 +8,19 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 {
     public class EffectMaterialContent : MaterialContent
     {
-        public ExternalReference<EffectContent> Effect { get; set; }
-        public ExternalReference<CompiledEffectContent> CompiledEffect { get; set; }
+        public const string EffectKey = "Effect";
+        public const string CompiledEffectKey = "CompiledEffect";
+
+        public ExternalReference<EffectContent> Effect
+        {
+            get { return GetReferenceTypeProperty<ExternalReference<EffectContent>>(EffectKey); }
+            set { SetProperty(EffectKey, value); }
+        }
+
+        public ExternalReference<CompiledEffectContent> CompiledEffect
+        {
+            get { return GetReferenceTypeProperty<ExternalReference<CompiledEffectContent>>(CompiledEffectKey); }
+            set { SetProperty(CompiledEffectKey, value); }
+        }
     }
 }
--- a/MonoGame.Framework.Content.Pipeline/Graphics/EffectMaterialContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/EffectMaterialContent.cs
@@ -8,20 +8,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 {
     public class EffectMaterialContent : MaterialContent
     {
-        public const string EffectKey = "Effect";
-        public const string CompiledEffectKey = "CompiledEffect";
-
-        [ContentSerializerIgnore]
-        public ExternalReference<EffectContent> Effect
-        {
-            get { return GetReferenceTypeProperty<ExternalReference<EffectContent>>(EffectKey); }
-            set { SetProperty(EffectKey, value); }
-        }
-
-        public ExternalReference<CompiledEffectContent> CompiledEffect
-        {
-            get { return GetReferenceTypeProperty<ExternalReference<CompiledEffectContent>>(CompiledEffectKey); }
-            set { SetProperty(CompiledEffectKey, value); }
-        }
+        public ExternalReference<EffectContent> Effect { get; set; }
+        public ExternalReference<CompiledEffectContent> CompiledEffect { get; set; }
     }
 }
--- a/MonoGame.Framework.Content.Pipeline/Graphics/EffectMaterialContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/EffectMaterialContent.cs
@@ -11,6 +11,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
         public const string EffectKey = "Effect";
         public const string CompiledEffectKey = "CompiledEffect";
 
+        [ContentSerializerIgnore]
         public ExternalReference<EffectContent> Effect
         {
             get { return GetReferenceTypeProperty<ExternalReference<EffectContent>>(EffectKey); }
--- a/MonoGame.Framework.Content.Pipeline/Processors/MaterialProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/MaterialProcessor.cs
@@ -173,8 +173,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
 
             // Build custom effects
             var effectMaterial = input as EffectMaterialContent;
-            if (effectMaterial != null)
+            if (effectMaterial != null && effectMaterial.CompiledEffect == null)
             {
+                if (effectMaterial.Effect == null)
+                    throw new PipelineException("EffectMaterialContent.Effect or EffectMaterialContent.CompiledEffect should be set for materials with a custom effect.");
                 effectMaterial.CompiledEffect = BuildEffect(effectMaterial.Effect, context);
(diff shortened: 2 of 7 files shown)


MonoGame.Framework.Content.Pipeline/Processors/MaterialProcessor.cs
```diff
@@ -173,8 +173,10 @@ public override MaterialContent Process(MaterialContent input, ContentProcessorC
 
             // Build custom effects
             var effectMaterial = input as EffectMaterialContent;
-            if (effectMaterial != null)
+            if (effectMaterial != null && effectMaterial.CompiledEffect == null)
```
Jjagg: This is not really necessary, but allows a custom processor to set a CompiledEffect directly.

ThirdParty/Dependencies
```diff
@@ -1 +1 @@
-Subproject commit 49eb2a3e2190b09433a856508d94b14b649da473
+Subproject commit a62d80be574ef74ee4196e3f7f1a40835634d857
```
tomspilman: I'm not sure about updating this to...

https://github.com/MonoGame/MonoGame.Dependencies/commit/a62d80be574ef74ee4196e3f7f1a40835634d857

@cra0zy - Is this safe without other code changes?

cra0zy: Needs #4885

ThirdParty/Kickstart
```diff
@@ -1 +1 @@
-Subproject commit 5e03d0127e514dcf8a63ae9ebd385cd392dcde52
+Subproject commit 290303cd5bedfdc802582a7cb324de1f0619ab08
```
tomspilman: @cra0zy @dellis1972 @mrhelmut  - Not sure about this update to the latest revision on https://github.com/OutOfOrder/MonoKickstart .

ThirdParty/Dependencies
tomspilman: @Jjagg - Please revert this change then... we'll deal with this in the other PR.

ThirdParty/Kickstart
cra0zy: We should update it to point to our fork: https://github.com/MonoGame/MonoKickstart

cra0zy: Then again why do we even have the kickstart submodule in the main repo?

mrhelmut: https://github.com/OutOfOrder/MonoKickstart is now outdated, https://github.com/MonoGame/MonoKickstart is the one to track.

I agree, there is no need to submodule MonoKickstart.

tomspilman: @cra0zy @mrhelmut - Let's take this conversation https://github.com/mono/MonoGame/issues/4895.

@Jjagg - Please remove this change as well.  Looks like we'll be doing something different for it.

Jjagg:
I reverted the submodule update. Is it alright like this or should I squash for clarity?

tomspilman:
@Jjagg - I can squash it during the merge...  no worries.

@mgarstenauer @nkast - Thoughts?

mgarstenauer:
@tomspilman Sorry, kinda busy for the next days. Won't have time to review any changes.

nkast:
The `ModelProcessor` clearly lets you override the effect so MG should implement it as well, but writing a custom `ModelProcessor` is overkill for the user.

For example could pipeline.exe populate all derived `Effect` types from ref assemblies and imported .fx files in the Material dropdown? It's doing something like that for custom Processors & Importers.

then mgcb should handle materials beyond the stock effects with parameters
like `processorParam:DefaultEffect=MyCustomEffect` & `processorParam:DefaultEffect=/MyEffect.fx`.



This is the first time I see that part of the pipeline and can't comment on the code. 
Also, I always load .mgfxo directly and apply the effect in code, never used .fx from the pipeline. 
Is it possible to attack a custom Effect class to a .fx file and then call `content.Load<MyCustomEffect>("MyEffect.fx")`?

Jjagg:
I agree it would be nice if this could be made easier, but that would probably complicate things a lot for the pipeline tool. Loading your own custom effect without a custom processor is 't possible without some extensions. I think we could do it if we had a generic effectreader and you could set the effect type (loaded from a referenced assembly) in a processor. That's outside the scope of this PR though. It would be better to open a new issue for that.

Jjagg:
EffectMaterialWriter doesn't actually have to override GetRuntimeReader since the base class can figure it out by just replacing "Writer" with "Reader" in the name. Built-in effects don't override it either. I'll fix this later today.

MonoGame.Framework.Content.Pipeline/Graphics/EffectMaterialContent.cs
```diff
@@ -8,19 +8,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 {
     public class EffectMaterialContent : MaterialContent
     {
-        public const string EffectKey = "Effect";
-        public const string CompiledEffectKey = "CompiledEffect";
```
nkast: `EffectKey` & and `CompiledEffectKey` are part of XNA API. we can't remove them.
"Specifies the key of the key/data pair for the compiled effect property.
This key/data pair is stored in the OpaqueData property of the material." 

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.content.pipeline.graphics.effectmaterialcontent.effectkey(v=xnagamestudio.40).aspx
https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.content.pipeline.graphics.effectmaterialcontent.compiledeffectkey(v=xnagamestudio.40).aspx

Jjagg:
Can this be merged @tomspilman? Or do you want someone else to confirm the fix first?

KonajuGames:
This PR has worked for those who have tried it, so I'm happy to merge it.

Jjagg:
Thanks!

PR #4906, opened 2016-06-06
[Pipeline Tool] PropertyGridTable Improvements

### LINKED ISSUE #4905: Pipeline Tool: rotation/scale of object cannot be modifed
FBX object (at least) rotation & scale cannot be modified using Pipeline tool UI from dev branch.

double click on property didn't pop-up a way to edit value like ColorKeyColor for instance.

(editing content.mgcb, or whatever its name is, is the only way to do it for now)

cra0zy:
Stuff done:
 - text editing now has a textbox inside property grid
 - numbers are now editable and have a textbox inside combobox (fixes #4905)
 - disabled height overriding of cellcombo for Gtk, it got cut off on some themes

PS. I snuck in a line of change to `Global.Linux.cs` since I noticed a bug with it while doing this PR


tomspilman:
Thanks @cra0zy .

PR #4907, opened 2016-06-06
More rendertarget fixes

prollin:
My last PR had a typo in it where DepthStencilState.StencilEnable was used instead of RasterizerState.ScissorTestEnable to determine if the scissor test is currently enabled.

This also fixes an issue where framebuffers used for the resolve operation are never properly removed  when deleting the corresponding multisampled rendertarget. This becomes a critical issue if you create and delete a lot of RT during the lifetime of your application (not the most common use case though).


tomspilman:
@dellis1972  @cra0zy  - How does this affect the SDL2 branch?  Will these fixes merge there?

cra0zy:
> Will these fixes merge there?

They should.

tomspilman:
Thanks @prollin .  Note the replacement of OpenTK on desktop with our own GL bindings has also occurred.

PR #4924, opened 2016-06-12
Fix Mouse updating when moving the Window

dellis1972:
When you click and hold the title bar of the window
and move it around we are still updating the X/Y
location of the Mouse. This does not happen in XNA
and it makes things look a bit weird as far as the
user is concerned.

This commit just moves the updating of the X/Y position
into the check that the mouse has Focus. This fixes the
issue since when we click on the Title bar the Mouse does
NOT have focus.


tomspilman:
So speaking to @dellis1972 this change is just an intermediate improvement to what we have already.  The reality is that we don't handle text input like this very well.  If you look at the docs for SDL2's text input you can see all the cases we don't handle...

https://wiki.libsdl.org/Tutorials/TextInput

We need to rethink our approach eventually to deal with all this, but for now this is an improvement.

Will follow this up with a PR that fixes Windows DX platforms to support the new Key info too.

PR #4925, opened 2016-06-12
Remove restriction on HardwareModeSwitch

dellis1972:
The HardwareModeSwitch property controls if the game window
goes into hard full screen or into a 'soft' fullscreen (which
is basically a fullscreen borderless form).

There is a use case where the developer might want to allow
the user the ability to switch between the two. Plus I cannot
see any hardware reason why this cannot be changed at runtime.


tomspilman:
Seems fine to me... thanks @dellis1972 !

PR #4927, opened 2016-06-12
Implement BackgroundContext loading for DesktopGL

dellis1972:
@cra0zy @prollin comments welcome :)

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsContext.SDL.cs
+++ b/MonoGame.Framework/Graphics/GraphicsContext.SDL.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using Microsoft.Xna.Framework;
 
 namespace OpenGL
 {
@@ -35,6 +36,7 @@ namespace OpenGL
                 return;
             
             SetWindowHandle(info);
+            Threading.BackgroundContext = Sdl.GL.CreateContext (_winHandle);
             _context = Sdl.GL.CreateContext(_winHandle);
 
             // GL entry points must be loaded after the GL context creation, otherwise some Windows drivers will return only GL 1.3 compatible functions
--- a/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
@@ -218,10 +218,6 @@ namespace Microsoft.Xna.Framework.Graphics
                         GL.PixelStore(PixelStoreParameter.UnpackAlignment, 4);
                     }
 
-#if !ANDROID
-                    GL.Finish();
-                    GraphicsExtensions.CheckGLError();
-#endif
                     // Restore the bound texture.
                     GL.BindTexture(TextureTarget.Texture2D, prevTexture);
                     GraphicsExtensions.CheckGLError();
--- a/MonoGame.Framework/SDL/SDLGamePlatform.cs
+++ b/MonoGame.Framework/SDL/SDLGamePlatform.cs
@@ -110,7 +110,6 @@ namespace Microsoft.Xna.Framework
 
             while (true)
             {
-                Threading.Run();
                 SdlRunLoop();
                 Game.Tick();
 
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -159,6 +159,7 @@ namespace Microsoft.Xna.Framework
             Sdl.GL.SetAttribute (Sdl.GL.Attribute.DoubleBuffer, 1);
             Sdl.GL.SetAttribute (Sdl.GL.Attribute.ContextMajorVersion, 2);
             Sdl.GL.SetAttribute (Sdl.GL.Attribute.ContextMinorVersion, 1);
+            Sdl.GL.SetAttribute (Sdl.GL.Attribute.ShareWithCurrentContext, 1);
             
             _handle = Sdl.Window.Create (title,
                 _winx - _width / 2, _winy - _height / 2,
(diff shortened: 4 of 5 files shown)


cra0zy:
:computer:  :speech_balloon: :+1: (roughly translated: everything looks good)

prollin:
@dellis1972 using glFinish vs glFlush to synchronize the 2 contexts is something I have been looking at recently. For iOS, [glFlush is supposed to be enough ](https://developer.apple.com/library/ios/qa/qa1612/_index.html) but it might be worth experimenting on various PC GPU to have optimal background gpu upload.
FYI we also removed the extra calls to GL.Finish in `Texture2D.SetData()` since this should only be called if *not* on the UI thread. Reading data back from the gpu (glReadData) automatically flushes the pipeline already.

tomspilman:
@dellis1972 @cra0zy - Should we try to get this into the 3.6 release or should we wait to merge this?

cra0zy:
3.6

endimilojkoski:
I can confirm that loading for me is MUCH faster with this.

dellis1972:
awesome :)

tomspilman:
This good to merge then? @KonajuGames  @dellis1972 @cra0zy @mrhelmut ?

cra0zy:
It is. The "glFinish vs glFlush" discussion can remain for another PR/issue.

tomspilman:
Ok...  going for it.

mrhelmut:
I have a GL question related to this.

I'm loading ```Texture2D``` with a thread while drawing other stuff. It can happen occasionally that OpenGL throws a 1282 (Invalid Operation) error upon loading a texture.

This also happened with OpenTK, which I assume was doing the same as this PR.

Is there a safe way to load textures from threads with GL? I always end up using a thread-safe content loader (which blocks draw calls until textures are loaded) to work around this. Before this PR, I've been able to remove the thread-safe loader, but I now need it again.

Should this be considered as an issue?

KonajuGames:
OpenGL drivers and multi-threading don't have a good history.
http://richg42.blogspot.com.au/2014/05/things-that-drive-me-nuts-about-opengl.html

Put simply, I believe there is no 100% safe way to multi-thread OpenGL.  It was never designed for it in the first place and not all driver developers get it right.

endimilojkoski:
I can also confirm that on initial content Loading on new Thread my game in rare cases crashes.
If I start it again it is all good.

mrhelmut:
@KonajuGames That's what I assumed. Thanks for the clarification.

@endimilojkoski If that may be of any help: http://konaju.com/?p=27
You will have a few skipped frames during loading, but at least it's safe.

KonajuGames:
> @endimilojkoski If that may be of any help: http://konaju.com/?p=27
> You will have a few skipped frames during loading, but at least it's safe.

That will help the C# side, but not the OpenGL driver side.

endimilojkoski:
Thanks guys, will think about it for next update :)
It is not a big deal, it happens in very rare cases and simple retry does the job so it is not critical :)

Scellow:
What i hate the most in games, it's when the window freeze while it load assets, it drive me crazy, if you use steam overlay and chat with someone, you are fucked up

mrhelmut:
Turns out the error still occurs frequently despite thread-safing assets loading. It wasn't happening before or with OpenTK.

mrhelmut:
After giving this some heavy testing, I think it exposes more issues than it would solve. While it indeed allows to load assets faster by not blocking them on the UI thread, if assets are loaded from a thread there is an uncontrollable risk of GL errors depending on driver vendors.

If you are fine with this, I'd like to add an option to disable background context.

dellis1972:
I think adding a way to disable it might be a good idea. You can still load
stuff on a background thread, it just will marshall the GL call over to the
main context.

tomspilman:
> It wasn't happening before or with OpenTK.

Then it shouldn't happen now with SDL2...  we are calling the same OpenGL here.  This suggests a mistake in our implementation that we really should find and fix.

> I'd like to add an option to disable background context.

IMO we have should either fix the bugs or remove the feature.

mrhelmut:
I'm not enough into OpenGL to grasp the full picture, but the more I read about GL and context threading, the more I think that we should block all assets loading/creation on the GL/UI thread.
As @KonajuGames highlighted, even if we proof the C# thread-safety, there is nothing that ensure that everything is going well on the driver side. Based on what I read, NVIDIA drivers are more susceptible to crash because of context switching and thread races and Apple drivers are more thread-friendly (which confirms my experiences, I've only seen this problem occurring on NVIDIA cards and Windows, no matter if Flush/Finish are there). While threading GL seems to be fine for streaming data to the GPU, it seems not advisable to thread data setup.

I would be favorable to a revert. What do you think?

As to why @endimilojkoski experiences content loading this longer, there must be something else to the problem because texture loading should be something quite fast.

KonajuGames:
> As to why @endimilojkoski experiences content loading this longer, there must be something else to the problem because texture loading should be something quite fast.

The reason for the slower loading performance is that it limits the loading to one GPU resource per frame, rather than potentially a few per frame when it's on a background thread that is not synchronized to the main thread.

Actually, I just had a thought, and fixing this would help a lot with this performance issue.  We block the background thread on a call to `Texture2D.SetData()`, but a texture load calls `Texture2D.SetData()` for each mipmap.  So it actually loads one mipmap per frame instead of the entire texture.  This could be easily remedied by adding a call to `Threading.BlockOnUIThread()` encapsulating the `levelCount` loop in `Texture2DReader`, `Texture3DReader` and `TextureCubeReader`.  Nested calls to `Threading.BlockOnUIThread()` aren't a problem as the inner calls will just execute immediately.

mrhelmut:
Interesting, I'll try to take a look at this, thanks.

mrhelmut:
FYI #4972

PR #4928, opened 2016-06-13
Silently catching haptic initialization failures

mrhelmut:
I have a few crash reports stating that ```Sdl.Haptic.NewEffect()``` may fail (error message = "Haptic error Unable to create effect") even though ```Sdl.Haptic.EffectSupported()``` succeed.
I can't be sure, but I believe this is due to some low-end gamepads having weird/wrong capabilities.

This PR silently catch the error and closes the haptic device to continue normally without haptic support.

cc @cra0zy @tomspilman


tomspilman:
Makes sense to me... merging!

PR #4930, opened 2016-06-13
Fixing GL calling conventions for win32

### LINKED ISSUE #4929: DesktopGL doesn't run on some 32bit Windows
We have some users which are unable to run our game on 32bit Windows systems.

The problem seems to be the OpenGL bindings. Every functions throw a ```PInvokeStackImbalance```.

@dellis1972 any clue?

mrhelmut:
Fixes #4929 

On Windows, ```CallingConvention``` for OpenGL has to be ```Winapi```, while it has to be ```Cdecl``` on Unix systems. Thankfully these are the default values on the .NET CLR and Mono running on Unix, and can therefore be safely removed.

This PR also fixes the signature of ```GL.DepthRange()```.

cc @tomspilman

### DIFF
--- a/MonoGame.Framework/Graphics/OpenGL.cs
+++ b/MonoGame.Framework/Graphics/OpenGL.cs
@@ -483,71 +483,60 @@ namespace OpenGL
         public partial class Ext
         {
             [System.Security.SuppressUnmanagedCodeSecurity ()]
-            [MonoNativeFunctionWrapper]
-            [UnmanagedFunctionPointer (CallingConvention.Cdecl)]
+            [MonoNativeFunctionWrapper]           
             public delegate void GenRenderbuffersDelegate (int count, out int buffer);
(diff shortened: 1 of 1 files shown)


dellis1972:
@mrhelmut is the UnmanagedFunction Pointer attrib not required at all?

tomspilman:
Should we apply this same fix to SDL2 and OpenAL?

mrhelmut:
@dellis1972 Yes, it uses the default values, which are the ones we want.

@tomspilman SDL2 & OpenAL-Soft are both using the Cdecl convention on all systems, these bindings are fine.

tomspilman:
Ok... so this is just because the GL DLL doesn't use Cdecl by default on Windows.  That makes sense now.

tomspilman:
Thanks @mrhelmut ... going to merge this now.

endimilojkoski:
Just tested, 32 bit only version now works fine.

PR #4931, opened 2016-06-14
TextEvent Fix for Windows/UWP

tomspilman:
This is the Windows DX and Windows UWP support for the change in https://github.com/mono/MonoGame/pull/4878.

### DIFF
--- a/MonoGame.Framework/TextInputEventArgs.cs
+++ b/MonoGame.Framework/TextInputEventArgs.cs
@@ -8,25 +8,30 @@ using Microsoft.Xna.Framework.Input;
 namespace Microsoft.Xna.Framework
 {
     /// <summary>
-    /// This class is used for the game window's TextInput event as EventArgs.
+    /// An event containing keyboard text input data.  See
+    /// <see cref="GameWindow.TextEvent"/> for details.
     /// </summary>
     public class TextInputEventArgs : EventArgs
     {
-        char character;
-        public TextInputEventArgs(char character, Keys key = Keys.None)
+        /// <summary>
+        /// Construct the event.
+        /// </summary>
+        /// <param name="character">The character value.</param>
+        /// <param name="key">The key code.</param>
+        public TextInputEventArgs(char character, Keys key)
         {
-            this.character = character;
-            this.Key = key;
+            Character = character;
+            Key = key;
         }
-        public char Character
-        {
-            get
-            {
-                return character;
-            }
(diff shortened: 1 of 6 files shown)


tomspilman:
@dellis1972 @KonajuGames 

This basically matches your change for SDL2 to the other platforms that support this event.  Still has the same flaws as the other.

tomspilman:
Looks like I need to fix the old MonoMac code path.

@dellis1972 @cra0zy - Were we going to remove that in 3.7?

MonoGame.Framework/TextInputEventArgs.cs
```diff
@@ -8,25 +8,30 @@
...
 {
     /// <summary>
-    /// This class is used for the game window's TextInput event as EventArgs.
+    /// An event containing keyboard text input data.  See
+    /// <see cref="GameWindow.TextEvent"/> for details.
     /// </summary>
     public class TextInputEventArgs : EventArgs
     {
-        char character;
-        public TextInputEventArgs(char character, Keys key = Keys.None)
```
dellis1972: @tomspilman put the default back in for "key" and the monomac build will pass.

dellis1972:
@tomspilman I want to see MonoMac gone asap, its totally redundant now that Xamarin.Mac is free.

MonoGame.Framework/TextInputEventArgs.cs
tomspilman: I am aware of that...  I didn't want to let it pass that way.  I wanted to fix the other platforms that use the event to have matching behavior to your changes in DesktopGL.

tomspilman:
> I want to see MonoMac gone asap, its totally redundant now that Xamarin.Mac is free.

It definitely has some stinky bits... like how it re-implements `GameWindow` instead of inheriting from it.

I would like the remove redundant platforms... MonoMac, Linux... now what we have SDL2.  Still i worry about breaking `Protobuild -generate MacOS` and `Protobuild -generate Linux` and how it would affect some users.

dellis1972:
@tomspilman can you rebase this PR?

tomspilman:
@dellis1972 - This cannot be fixed as is.  

After i wrote this PR we merged a bunch of other changes to Windows text input over it which broke the approach.

I'm going to close this for now and put in an issue to revisit it fresh.  See https://github.com/MonoGame/MonoGame/issues/6204

PR #4935, opened 2016-06-14
[Linux Installer] Replace += operator

### LINKED ISSUE #4933: Issue with Development monogame.run
There is a problem with the current development .run installer on Linux. The standard one for MonoGame 3.5 works fine but when I try to run the development one it continuously spams './postinstall.sh: 10: ./postinstall.sh: line+=.: not found'
into the terminal and does not install. The only way to stop it is by using Ctrl+C

I am running Ubuntu Gnome 16.04. I have installed the latest version of Mono and MonoDevelop

cra0zy:
Fix #4933

### DIFF
--- a/Installers/Linux/RUN/postinstall.sh
+++ b/Installers/Linux/RUN/postinstall.sh
@@ -1,20 +1,20 @@
 #!/bin/sh
 
 # Functions
-function echodep
+echodep()
 {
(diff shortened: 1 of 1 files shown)


KonajuGames:
Merging.  Thanks @cra0zy

PR #4937, opened 2016-06-15
Silently catching rumble init errors

mrhelmut:
Following #4928, it turns out that ```Sdl.Haptic.RumbleInit``` can raise the same error despite passing the rumble test... (e.g. with Logitech F710 gamepads).
This PR silently catches init errors and closes the haptic device to continue normally without rumble.

cc @cra0zy @tomspilman


KonajuGames:
Can't trust anything these days.  Merging.

PR #4940, opened 2016-06-15
Returns proper screen size in UAP

GrafSeismo:
This change delivers the current desktop screen size for windows universal projects (WINDOWS_UAP)

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.cs
@@ -1,11 +1,12 @@
+#region License
 // MonoGame - Copyright (C) The MonoGame Team
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
+#endregion
 
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
-
 #if MONOMAC
 #if PLATFORM_MACOS_LEGACY
 using MonoMac.AppKit;
@@ -20,6 +21,9 @@ using UIKit;
 using Android.Views;
 using Android.Runtime;
 #endif
+#if DIRECTX
+using SharpDX.DXGI;
+#endif
 
 namespace Microsoft.Xna.Framework.Graphics
 {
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Graphics/GraphicsAdapter.cs
```diff
@@ -107,6 +111,12 @@ public DisplayMode CurrentDisplayMode
                     graphics.ReleaseHdc(dc);
                     return new DisplayMode(width, height, SurfaceFormat.Color);
                 }
+#elif WINDOWS_UAP
+                var visibleBounds = ApplicationView.GetForCurrentView().VisibleBounds;
+                var rawPixelsPerViewPixel = DisplayInformation.GetForCurrentView().RawPixelsPerViewPixel;
+                return new DisplayMode((int)(visibleBounds.Width * rawPixelsPerViewPixel + 0.5f),
+                                       (int)(visibleBounds.Height * rawPixelsPerViewPixel + 0.5f),
+                                       SurfaceFormat.Color);
 #else
```
tomspilman: @GrafSeismo - Doesn't this just return the current window size?  This is supposed to return the information for the desktop.

@KonajuGames ?

theZMan: @tomspilman - see https://msdn.microsoft.com/library/windows/apps/dn264258?f=255&MSPPError=-2147217396 - this gets the dimensions of the current physical device. 

Not sure about what happens in the case of multi mon

GrafSeismo:
Well it is a trick, metro style apps take the full screen when they start up. Microsoft decided to drop exclusive access to screen - no fullscreen, no vertical sync disable, no information about native screen size, etc. It is really boring to develop games with c# now. If i had the knowledge from today when i started with XNA i would have had droped it and change to an other framework and language, maybe OS too. Metro Style + NetCore is a real wrong way, do things complicated without any need and a lot less possibilities and comfort. I am really upset of MS C# misleading the users without any native support for games programming anymore. But now i have a big engine and i do not want to drop that hard work - currently. If i would have the time i would give Rust a better try.

MonoGame.Framework/Graphics/GraphicsAdapter.cs
GrafSeismo: Well it is a trick, metro style apps take the full screen when they start up. Microsoft decided to drop exclusive access to screen - no fullscreen, no vertical sync disable, no information about native screen size, etc. It is really boring to develop games with c# now. If i had the knowledge from today when i started with XNA i would have had droped it and change to an other framework and language, maybe OS too. Metro Style + NetCore is a real wrong way, do things complicated without any need and a lot less possibilities and comfort. I am really upset of MS C# misleading the users without any native support for games programming anymore. But now i have a big engine and i do not want to drop that hard work - currently. If i would have the time i would give Rust a better try.

https://social.msdn.microsoft.com/Forums/en-US/e3c4d339-a6aa-42e0-a453-3d6eddeca5d9/get-physical-screen-size?forum=winappswithcsharp

BrianPeek: They only take up the full screen if they're not windowed.  Not sure this does anything other than what Tom says, it returns the size of the window, not the desktop.

GrafSeismo: Why it work for me under 10 than?

BrianPeek: This works *if* the app is running full-screen, but that's not guaranteed or a requirement for a Win10 UWP.  Are you forcing to full-screen with TryEnterFullScreenMode or some such?  I just took your code and ran it against an app that starts in a window and I simply get the resolution of the window, not the desktop.

GrafSeismo: metro style apps do take the full screen when they start up, exept you change that behavior first an than ask for size.

GrafSeismo: If you want the desktop screen resolution after window has loaded the only way seems to use SharpDX
```
using SharpDX.DXGI;
...
            // # of graphics card adapter
            const int numAdapter = 0;

            // # of output device (i.e. monitor)
            const int numOutput = 0;

            // Create DXGI Factory1
            var factory = new Factory1();
            var adapter = factory.GetAdapter1(numAdapter);

            // Get DXGI.Output
            var output = adapter.GetOutput(numOutput);

            // Width/Height of desktop to capture
            width = output.Description.DesktopBounds.Right;
            height = output.Description.DesktopBounds.Bottom;
```

tomspilman: That seems like the correct way to do this and should work on all DirectX platforms.

BrianPeek: 100% agreed with @tomspilman .  :) Just note you're getting the resolution of monitor 0 on card 0...one could have multiple resolutions across cards/monitors...

GrafSeismo: > When using DXGI with multiple monitors, there are two rules to follow.
> 
> The first rule applies to the creation of two or more full-screen swap chains on multiple monitors. When creating such swap chains, it is best to create all swap chains as windowed, and then to set them to full-screen. If swap chains are created in full-screen mode, the creation of a second swap chain causes a mode change to be sent to the first swap chain, which could cause termination of full-screen mode.
> 
> The second rule applies to outputs. Be watchful of outputs used when creating swap chains. With DXGI, the IDXGIOutput object controls which monitor the swap chain uses when becoming full-screen. Unlike DXGI, Direct3D 9 had no concept of outputs.

What kind of solution are you think of when this framework should support multiple resolutions across cards/monitors?

> **DXGI 1.1**
> 
> The Direct3D 11 runtime included in Windows 7 and installed onto Windows Vista (see KB971644) includes version 1.1 of DXGI. This update adds definitions for a number of new formats (particularly BGRA, 10-bit X2 bias, and Direct3D 11's BC6H and BC7 texture compression), as well as a new version of the DXGI factory and adapter interfaces (CreateDXGIFactory1, IDXGIFactory1, IDXGIAdapter1) for enumerating remote desktop connections.
> 
> When you use Direct3D 11, the runtime will use DXGI 1.1 by default when calling D3D11CreateDevice or D3D11CreateDeviceAndSwapChain with a NULL IDXGIAdapter pointer. Mixing use of DXGI 1.0 and DXGI 1.1 in the same process is not supported. Mixing DXGI object instances from different factories in the same process also is not supported. Therefore, when you use DirectX 11, any explicit use of the DXGI interfaces uses a IDXGIFactory1 created by the CreateDXGIFactory1 entry-point in “DXGI.DLL” to ensure the application is always using DXGI 1.1.
> 
> **DXGI 1.2**
> 
> The Direct3D 11.1 runtime that is included in Windows 8 also includes version 1.2 of DXGI.
> 
> DXGI 1.2 enables these features:
> -     stereo rendering
> -     16 bit-per-pixel formats
>          DXGI_FORMAT_B5G6R5_UNORM and DXGI_FORMAT_B5G5R5A1_UNORM are now fully supported
>          a new DXGI_FORMAT_B5G5R5A1_UNORM format was added
> -     video formats
> -     new DXGI interfaces

What about stereo rendering and multitasking?

GrafSeismo:
oh man, that sucks. Do you have a hint why it failed or where i do have to look at?

GrafSeismo:
Well, i hope that work now....

Tzenchor:
The problem is that autoimplemented properties must a `get` and a `set` accessors, if the `set` has to inaccesible outside its class mark it as `protected` or `private`

GrafSeismo:
well, i give it up here, too much time lost so far. Sorry.

Tzenchor:
> @Tzenchor: https://msdn.microsoft.com/en-us/library/x9fsa0sw.aspx
>>    Properties that do not implement a set accessor are read only.

>It seems you need to update your builder.

Yep, the builder is using c# 5.0 specification, not 6.0 as used by Visual Studio 2015.

PR #4941, opened 2016-06-15
Make TitleContainer resolve resource paths on Windows

jakepoz:
This is the new version of https://github.com/mono/MonoGame/pull/4132 that splits its TitleContainer code into platform specific files.

When you call the TitleContainer.OpenStream method on Windows platforms, it needs to lookup the location of the specified resource. This is because on Windows 10, resources may be stored outside of the main application folder when using an "application bundle" (now the default setting). 

This solves a problem that we had in production where our game would work fine when deploying from within Visual Studio, but it wouldn't be able to find its localized resources when bundled and deployed via a package. With this PR, localized resources are looked up using the standard Windows APIs and everything works great.

This API is available for Windows8, WindowsPhone81, and WindowsUniversal but not for WindowsPhone (8.0), so the TitleContainer was split appropriately to prevent too many #ifs.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -359,7 +359,10 @@
       <Platforms>iOS,tvOS,MacOS</Platforms>
     </Compile>
     <Compile Include="TitleContainer.WinRT.cs">
-      <Platforms>Windows8,WindowsPhone,WindowsPhone81,WindowsUniversal</Platforms>
+      <Platforms>Windows8,WindowsPhone81,WindowsUniversal</Platforms>
+    </Compile>
+    <Compile Include="TitleContainer.WP.cs">
+      <Platforms>WindowsPhone</Platforms>
     </Compile>
     <Compile Include="Vector2.cs" />
     <Compile Include="Vector3.cs" />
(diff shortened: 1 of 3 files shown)


jakepoz:
Hi, can someone please take a look at this? We would like to submit it soon :)

ThomasZeman:
Why returning a Task with result "null" in case of error and not just letting the task fail? In the current code a caller has no possibility to find out what the actual problem is. Also, I would not offer a "faked" sync version of an async method. If someone needs that s/he can easily do that by herself. That's not in the scope of a library and is hard to keep bug free in regards to for example deadlocks.

jakepoz:
Hey, great points, let me address them:

Return Null in case of error: I am preserving the behavior that was there before this change. The public level code in the main TitleContainer.cs file actually catches all other exception types and rethrows them as FileNotFoundException anyways. 

The synced PlatformOpenStream:  This method is actually private and only used by the non-platform-specific TitleContainer code. All other platforms use this pattern and make their PlatformOpenStream be synchronous, and it's not exposed outside of this class.

ThomasZeman:
Of course, sure, I've overlooked the "mimic XNA behavior" aspect. So no further review comments from my side!

KonajuGames:
Thanks @jakepoz . Merging.

PR #4944, opened 2016-06-16
Add additional checks in Texture3D.SetData

Tzenchor:
This is the second part to solve issue #1455, the first being #4862 
Adds tests for Texture3d.setdata.
Adds additional checks for the parameters, note that the Exceptions thrown does not match the ones thrown by XNA as MonoGame allows operations that are safe, but in XNA are disallowed.


KonajuGames:
Thanks @Tzenchor.  Merging.

PR #4949, opened 2016-06-17
Proofing haptic init

mrhelmut:
Despite #4937 &  #4928 ( -_-' ), there are still possible errors with the haptic initialization which may result in crashes (mostly with Logitech F710 & F310 gamepads, or X360 controllers with non-official drivers).

This PR hopefully bullet-proof the whole thing.

I can't be sure, but I suspect that errors are not cleared and re-thrown outside the try/catch upon new ```Sdl.Haptic.OpenFromJoystick``` or ```Sdl.Haptic.EffectSupported```. Hence, I removed internal calls to ```Sdl.GetError()``` to avoid exceptions and handle the errors manually.

@dellis1972 This may be linked to the errors you may have encountered on Linux.

### DIFF
--- a/MonoGame.Framework/SDL/SDL2.cs
+++ b/MonoGame.Framework/SDL/SDL2.cs
@@ -141,6 +141,9 @@ internal static class Sdl
         return pointer;
     }
 
+    [DllImport(NativeLibName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SDL_ClearError")]
+    public static extern void ClearError();
+
     [DllImport(NativeLibName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SDL_GetHint")]
     public static extern IntPtr SDL_GetHint(string name);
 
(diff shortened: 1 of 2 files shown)


dellis1972:
@mrhelmut we didn't have a gamepad/joystick attached afaik. But it might help :)

mrhelmut:
@dellis1972 Good to know, I'll investigate the marshaling then. :)

endimilojkoski:
This fixes the Mac game crash issue when staring it with X360 gamepad connected.

tomspilman:
@dellis1972 @cra0zy - This look ok to merge?

mrhelmut:
I have pushed a small SDL binding clean up, the only relevant change is that the ```EventType``` enum should be ```uint```. The other changes are there only to double-check that anything related to ```PeepEvents``` and ```PollEvents``` are fine.

KonajuGames:
@dellis1972 If you can check this on Linux, then we can merge it.

dellis1972:
I'll have to get my Mate to test it. Will do that asap. If it works I'll
merge

cra0zy:
Testing...

MonoGame.Framework/SDL/SDL2.cs
```diff
@@ -740,13 +797,14 @@ public static string GetName(IntPtr gamecontroller)
 
     public static class Haptic
     {
-        public const uint Infinity = uint.MaxValue;
+        public const uint Infinity = 4292967295U;
```
cra0zy: Why the number instead of `uint.MaxValue`?

cra0zy:
This works on Linux just fine.

MonoGame.Framework/SDL/SDL2.cs
mrhelmut: That was just a test to make sure every ranges were fine, but I admit it's kind of useless.

PR #4952, opened 2016-06-19
Fix importing of skeleton which is not a direct child of root node

sim31:
Before this fix, model with a skeleton which isn't direct child of
root node would become a direct child of root node which messed up some
animations.

In my case skeleton was a child of a mesh which was animated so it impacted the whole skeleton. Assimp model viewer supports that kind of animation so I think it would make sense for MonoGame too.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/MeshHelper.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/MeshHelper.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 
@@ -301,6 +301,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
             }
         }
 
+        
         /// <summary>
         /// Search for the root bone of the skeletion.
         /// </summary>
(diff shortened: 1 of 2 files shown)


nkast:
Can you please update this with the latest repo? Build server didn't produce the Setup/installer.

Did you checked this with XNA? I have a model that gives me the following message:
`The skeleton root, "PIVOT", has been pulled out of the scene hierarchy and made into a child of the root node. One of the skeleton root's old parents, "Person", has an animation named "Base Stack" which will not be reflected on the skeleton root.` How will this PR affect that model?

nkast:
I tested your PR and saw no change with my fbx. My problem is different, in that MG converts NodeContent into BoneContent instead of moving the skeleton root BoneContent to the root node. So, with some files I end up with 23 bones instead of 22. That's a problem because I can't share animations between those models.

nkast:
Here's what I thing XNA is doing, 
Any skeleton root is moved to the root node and the animation is copied to this BoneContent. All this happen in the processor. 

for example:
root:(NodeContent)
   mesh:(MeshContent)
   Anim:(NodeContent) + 1 animation
       SkelRoot:(BoneContent)
         //rest of the skeleton             

is transformed to :
root:(NodeContent)
   mesh:(MeshContent)
   Anim:(NodeContent) + 1 animation
   SkelRoot:(BoneContent) + 1 animation
      //rest of the skeleton             

While MG now transform it into:
root:(NodeContent)
   mesh:(MeshContent)
   Anim:(BoneContent) + 1 animation
       SkelRoot:(BoneContent)
         //rest of the skeleton

SimonDarksideJ:
@sim31 were you able to address @nkast 's queries on the subject?

sim31:
@SimonDarksideJ This was a long time ago... But I'm pretty sure I would have posted an update here if I had addressed these queries.

SimonDarksideJ:
Do you want to close this PR then @sim31 or is it still your intention to progress with it?

PR #4953, opened 2016-06-20
WinDX: Fixed crash when running from a network path

MichaelDePiazzi:
I got a crash report telling me that the `WinFormsGameWindow` constructor had thrown an `ArgumentException` when calling `Icon.ExtractAssociatedIcon`. After some investigation, I found that this same crash would occur when running my game from a UNC network path.

According to the [MSDN documenation](https://msdn.microsoft.com/en-us/library/system.drawing.icon.extractassociatedicon(v=vs.110).aspx), `Icon.ExtractAssociatedIcon` does not support UNC paths (it doesn't say why though). I found a couple of stackoverflow posts<sup>[[1](http://stackoverflow.com/questions/6819466/extractassociatedicon-returns-null)][[2](http://stackoverflow.com/questions/1842226/how-to-get-the-associated-icon-from-a-network-share-file)]</sup> suggesting to use P/Invoke to get around this. So I gave this a try and found that it did indeed work correctly for both UNC and normal file paths.

This PR uses this solution to fix the crash which would occur when running from a UNC network path.

### DIFF
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -138,10 +138,7 @@ namespace MonoGame.Framework
             _form = new WinFormsGameForm(this);
             _form.ClientSize = new Size(GraphicsDeviceManager.DefaultBackBufferWidth, GraphicsDeviceManager.DefaultBackBufferHeight);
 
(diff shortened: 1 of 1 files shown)


tomspilman:
This is great...  merging!

PR #4430, opened 2016-01-17
Documentation for packaging and distributing projects

mrhelmut:
I started working on some documentation regarding how to package desktop projects for market distribution.

This is a very early work, mostly a draft, therefore not ready to be merged.

Topics planned:
- referencing all runtime requirements for both Windows & DesktopGL projects
- using MonoKickstart
- creating OS X bundles & dmg's
- creating Linux packages
- highlighting specific requirements for Steam and most popular distribution platforms

Not planned (I lack the knowledge):
- creating an installer for Windows or DesktopGL projects

PR #4954, opened 2016-06-20
SDL Controller DB update

mrhelmut:
This is an update to the SDL controllers database to support more gamepads. Changes come from upstream (https://github.com/gabomdq/SDL_GameControllerDB).


tomspilman:
Before we update this should we investigate the Steam SDL database too?

@cra0zy - Do you have a link to that?

cra0zy:
> Before we update this should we investigate the Steam SDL database too?

We cannot use that, it's not a static database, it contains the SDL general database (which we have a copy of), and it contains any controller the user configures within Steam Big Picture Mode.

Any SDL game launched through Steam should get autohooked to this database.

mrhelmut:
I confirm, it's a subset of the database, plus any mapping found locally in the Steam install folder (config/config.vdf). Steam autohooks, it's basically how the Steam Controller works.

I'll try to add documentation about this in #4430 (which I'll have time to continue after getting around the few ongoing gamepad works).

tomspilman:
Ah... I see.  As long as MonoGame supports it then we're good.

PR #4957, opened 2016-06-21
Update SDL Platform to Enable/Disable Screensaver of it loose/gain focus.

dellis1972:
We really shouldn't be Disabling the screen saver completely.
If the user switches to another app the screen saver should be
re-enabled. As a result we also need to make sure we re-disable
it if we get focus again.

Additionally we need to make sure the Cursor is set because sometimes
on MacOS it looses the setting when switching windows

### DIFF
--- a/MonoGame.Framework/SDL/SDL2.cs
+++ b/MonoGame.Framework/SDL/SDL2.cs
@@ -149,6 +149,9 @@ internal static class Sdl
     [DllImport(NativeLibName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SDL_DisableScreenSaver")]
     public static extern void DisableScreenSaver();
 
+    [DllImport (NativeLibName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SDL_EnableScreenSaver")]
+    public static extern void EnableScreenSaver ();
+
     [DllImport(NativeLibName, CallingConvention = CallingConvention.Cdecl, EntryPoint = "SDL_GetVersion")]
     public static extern void GetVersion(out Version version);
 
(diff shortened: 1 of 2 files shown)


theZMan:
Why would we disable the screen saver at all?

Disabling the screen saver is usually only done for apps that are likely to have little user input e.g. media playback or presentation mode in Powerpoint.

I would expect a game to honor my screen saver settings.

mrhelmut:
I remember that back in the OpenTK days, we disabled screensaver too because on some platforms, playing with a gamepad was considered as being idle and the screensaver started during gameplay.

Looks good to me.

tomspilman:
> Why would we disable the screen saver at all?

On windows at least I don't believe `GamePad` input keeps the screensaver from coming up... at least not always (there is some sort of setting).  So you can go to screensaver as your in the middle of playing.

Also games can have cut scenes where there is no input happening... you certainly don't want a screensaver to appear there.

However XNA this was under the control of the developer using `Guide.IsScreenSaverEnabled`:

  https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.gamerservices.guide.isscreensaverenabled.aspx?f=255&MSPPError=-2147217396

dellis1972:
Soo whats the plan? Remove Disabling the Screen saver completely? In which case we need to provide a way to do it somehow via Guide.IsScreenSaverEnabled, or go with this change?

mrhelmut:
Using only ```Guide.IsScreenSaverEnabled``` would be the correct XNA-compliant path, I believe. But since ```Guide``` doesn't exist in the MG space (and/or isn't relevant to it), should we move to something else? ```Game.IsScreenSaverEnabled```?

KonajuGames:
I think the `IsScreenSaverEnabled` static property should be reinstated somewhere.  In the Game class does seem like a sensible place to put it.

PR #4959, opened 2016-06-23
Fixed DirectX backbuffer resize behaviour

nathanleroux:
Also added an if statement to stop the infinite recursion issue I found in WinFormsGameWindow.cs (directly related to backbuffer problem)

I don't think this will cause more issues, but please let me know if it does.

Fixes https://github.com/mono/MonoGame/issues/4922
Fixes https://github.com/mono/MonoGame/issues/4018
Fixes https://github.com/mono/MonoGame/issues/4841

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/TextureImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/TextureImporter.cs
@@ -157,8 +157,20 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
 
                 default:
                     // Bitmap and other formats are converted to 32-bit by default
+                    // Are we in the BGRA format?  If so, we need to switch R & B instead of R & G
+                    bool isBGRA =
+                        imageType == FREE_IMAGE_TYPE.FIT_BITMAP &&
+                        FreeImage.GetBlueMask(fBitmap) == 0x000000FF &&
+                        FreeImage.GetGreenMask(fBitmap) == 0x0000FF00 &&
+                        FreeImage.GetRedMask(fBitmap) == 0x00FF0000;
+
                     bgra = FreeImage.ConvertTo32Bits(fBitmap);
-                    SwitchRedAndBlueChannels(bgra);
+
+                    if (isBGRA)
+                        SwitchRedAndGreenChannels(bgra);
+                    else
+                        SwitchRedAndBlueChannels(bgra);
+                    
                     FreeImage.UnloadEx(ref fBitmap);
                     fBitmap = bgra;
                     break;
@@ -178,6 +190,19 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
             FreeImage.SetChannel(fBitmap, r, FREE_IMAGE_COLOR_CHANNEL.FICC_BLUE);
             FreeImage.UnloadEx(ref r);
             FreeImage.UnloadEx(ref b);
+        }
+        /// <summary>
+        /// Switches the red and green channels
+        /// </summary>
+        /// <param name="fBitmap">image</param>
+        private static void SwitchRedAndGreenChannels(FIBITMAP fBitmap)
+        {
+            var r = FreeImage.GetChannel(fBitmap, FREE_IMAGE_COLOR_CHANNEL.FICC_RED);
+            var g = FreeImage.GetChannel(fBitmap, FREE_IMAGE_COLOR_CHANNEL.FICC_GREEN);
+            FreeImage.SetChannel(fBitmap, g, FREE_IMAGE_COLOR_CHANNEL.FICC_RED);
+            FreeImage.SetChannel(fBitmap, r, FREE_IMAGE_COLOR_CHANNEL.FICC_GREEN);
+            FreeImage.UnloadEx(ref r);
+            FreeImage.UnloadEx(ref g);
         }
     }
 }
--- a/MonoGame.Framework.Content.Pipeline/TextureImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/TextureImporter.cs
@@ -157,20 +157,8 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
 
                 default:
                     // Bitmap and other formats are converted to 32-bit by default
-                    // Are we in the BGRA format?  If so, we need to switch R & B instead of R & G
-                    bool isBGRA =
-                        imageType == FREE_IMAGE_TYPE.FIT_BITMAP &&
-                        FreeImage.GetBlueMask(fBitmap) == 0x000000FF &&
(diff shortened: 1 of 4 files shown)


nathanleroux:
Just checked, this also reproduces the desired WindowsGL behavior on WindowsDX for me with the test mentioned in https://github.com/mono/MonoGame/issues/4018, so this should fix that too.

nathanleroux:
Whoops, didn't realize that commit would get thrown on here.  I'll fix that shortly and just have a different branch for upstream stuff so it doesn't get mixed together like that.

nathanleroux:
There we go, all set.
Still getting the hang of pull requests.

tomspilman:
Thanks @nathanleroux !

@craftworkgames - Got a second to give this a test?

tomspilman:
Anyone have time to test this a bit?

lozzajp:
@tomspilman I would be happy to. Basically get Nathans branch and build it, test out the resizing etc. in a game?

KonajuGames:
> Basically get Nathans branch and build it, test out the resizing etc. in a game?

Yes, that will work.

lozzajp:
So this is previously (MG 3.5.1 I am pretty sure)

lozzajp:
With Nathan's fix. No stretching (and if you go inwards, the RT stays there and you just see less).

However if I am in full screen mode and try to change my "scale" I get this exception: (unsure if it is the way I am doing things though).

err2

tomspilman:
> going from windowed to full screen is fine, however full screen to windowed is crashing.

If you look at the `PresentationParameters.BackBufferWidth` and height there what are they?  I suspect one is probably negative or zero which is causing the exception.

lozzajp:
breaking every time that is called I get values like:

800x480 (windowed)
1920x1080 (my desktop res)
1916x1054 (just under, I have seen these values before in places, unsure why)
1932x1092 (just OVER my res)

nathanleroux:
I'll take a look, see if I can get fullscreen to stop crashing.

nathanleroux:
@lozzajp How are you changing the scale?

lozzajp:
@nathanleroux 
graphics.PreferredBackBufferWidth = screenWidth * ScreenScale;
graphics.PreferredBackBufferHeight = screenHeight * ScreenScale;
graphics.ApplyChanges();

and for full screen graphics.ToggleFullScreen();

However in a blank project just fullscreening I am having issues.  (those numbers above are from a blank project)

nathanleroux:
I can get it to crash in the same place if I attempt to use a screen size that my graphics card rejects (such as a negative size, or a very large size, apparently it accepts backbuffers of zero size), albeit with E_INVALIDARG instead of DXGI_ERROR_INVALID_CALL.  Both do mean the same thing, however (invalid argument data).

When it crashes, what values are the PresentationParameters.Width/Height?

lozzajp:
as above:
800x480 (windowed)
1920x1080 (my desktop res)
1916x1054 (just under, I have seen these values before in places, unsure why)
1932x1092 (just OVER my res)

This is just a plain blank GL project, the only code change is adding the fullscreen toggle to a key.

nathanleroux:
I can't get it to reproduce without using absurd resolution numbers, messing with the sizing and fullscreen otherwise doesn't crash for me.

lozzajp:
Interesting, perhaps my card then.

I have tried hardcoding the numbers into the swapChain.ResizeBuffers method and it still crashes the same.

Using the exact same code just swapping the reference for the MG release I have installed it works fine.

lozzajp:
@nathanleroux I have commented out your (literally) few lines of code and still get the same behaviour.

Perhaps I am doing something really stupid building it or could the SharpDX dll be slightly different?

Is it worth us swapping builds of test games?

lozzajp:
Just tested on another machine (and also a Win 10 VM) with the same results.

Am I doing something wrong in the build process?

lozzajp:
I saw a few posts around the web about releasing the backbuffer and resources when it comes to some SharpDX swapchain stuff but didn't have a clue if it was related or not (and still don't) but I came across this:

http://stackoverflow.com/questions/29041652/error-0x887a0001-from-sharpdx-resizebuffers-on-resize-event

@nathanleroux it's a bit beyond me, but what do you think?

tomspilman:
> I saw a few posts around the web about releasing the backbuffer

This has nothing to do with releasing any resources.


@nathanleroux - If the crash doesn't occur when the user manually resizes the window by dragging it or if you toggle to fullscreen and back to windowed with a "correct supported resolution....    then I say this PR is good to go as it is.

I'm sure we do zero validation of the resolution passed to the graphics device for fullscreen or windowed to ensure it is valid.  I expect XNA handles this with an appropriate exception or something like that....we need to test that.  We can then submit a fix for invalid resolutions as a separate PR.

nathanleroux:
@lozzajp If you upload the blank project I'll take a look on my end.

@tomspilman Yeah, it just passes whatever you set PreferredBackBufferWidth/Height to right on to SharpDX

lozzajp:
@nathanleroux  I just quickly chucked it up here: https://www.dropbox.com/sh/bq3m7f9v40vsent/AADQLM4AxMdP3Mnk7ho8ba_8a?dl=0

Would be interesting if the bin build behaves correctly for you, the solution has the MG windows project referenced but I obviously didn't upload that.

nathanleroux:
@lozzajp bin build does indeed crash for me when leaving fullscreen, I'm going to poke around some more with it and see whats going on.

nathanleroux:
@lozzajp Please try d10478e5c215a75ed57911b7a778e6aecc1c770c

lozzajp:
Yep no more full screen crashes.

I will test at home later make sure no stretching is going on so it is still achieving its initial requirements.

lozzajp:
@nathanleroux All seems ok for my project. Doesn't stretch etc. No crashing. I notice a SharpDX exception on project startup but guess its handled fine and intentional.

Thanks for this! I am glad to have helped a bit.

@tomspilman good to go?

craftworkgames:
@tomspilman I finally got around to testing this with MonoGame.Extended. 

The core problem seems to be resolved which is great but I did notice that `Window.ClientSizeChanged` fires immediately after `Game.Initialize` on GL and doesn't on DX. We've been relying on this behavior, but I'm fairly confident we can work around it if you don't / can't get them consistent. No biggie.

Either way, I'm pleased this one has been resolved. It's been bugging me for a long time.

tomspilman:
>  I did notice that `Window.ClientSizeChanged` fires immediately after `Game.Initialize` on GL and doesn't on DX.

Please setup a new issue for it and explain the situation with some more details.  Ideally we match the XNA behavior here although I would suggest depending on events like that might be something we can't keep 100% compatibility with.

Thanks @nathanleroux ...  merging!

PR #4962, opened 2016-06-25
properly apply tint color to RBG channels

prime31:
Small fix to the SpriteEffect shader that multiplies the texture.rgb by the vertex color alpha

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/Resources/SpriteEffect.fx
+++ b/MonoGame.Framework/Graphics/Effect/Resources/SpriteEffect.fx
@@ -40,7 +40,9 @@ VSOutput SpriteVertexShader(	float4 position	: POSITION0,
 
 float4 SpritePixelShader(VSOutput input) : SV_Target0
 {
-    return SAMPLE_TEXTURE(Texture, input.texCoord) * input.color;
+    float4 color = SAMPLE_TEXTURE(Texture, input.texCoord) * input.color;
+	color.rgb *= input.color.a;
+	return color;
 }
 
 TECHNIQUE( SpriteBatch, SpriteVertexShader, SpritePixelShader );


MonoGame.Framework/Graphics/Effect/Resources/SpriteEffect.fx
```diff
@@ -40,7 +40,9 @@ VSOutput SpriteVertexShader(	float4 position	: POSITION0,
 
 float4 SpritePixelShader(VSOutput input) : SV_Target0
 {
-    return SAMPLE_TEXTURE(Texture, input.texCoord) * input.color;
+    float4 color = SAMPLE_TEXTURE(Texture, input.texCoord) * input.color;
+	color.rgb *= input.color.a;
+	return color;
 }
```
tomspilman: I really don't think this is correct.  The existing code comes directly from the XNA implementation of the stock effects that you can see here:

http://xbox.create.msdn.com/en-US/education/catalog/sample/stock_effects

tomspilman:
I would have expected this unit test to fail:

https://github.com/mono/MonoGame/blob/develop/Test/Framework/Visual/SpriteBatchTest.cs#L161

But it didn't which suggests something is wrong with that test.

nkast:
@prime31 
What if you use a premultiply color to draw your text? would that had the same result? use Color.FromNonPremultiplied() to convert your current color. For example [1,1,1,0.5] becomes [0.5,0.5,0.5,0.5].

@tomspilman Same though here. Perhaps we don't have a test for semitransparent texts.

prime31:
Here is an image that shows the issue that uses the default SpriteEffect and shader.

blendy-time


On the top, the same rainbow image is drawn 3 times with the Color specified next to it. That is with a CornflowerBlue clear color.

On the bottom, the same rainbow image is drawn 3 times with the Color specified next to it. That is with a Black clear color.


@nkast it doesn't matter what you do with the color before you send it to the shader. Even with a 0 alpha value the texture is still displayed. When alpha is 0 one would expect a fully transparent texture to be rendered.

prime31:
For comparison, here is what the same render looks like with the shader changes:
screen shot 2016-06-25 at 5 13 44 pm

Notice that the third instance of the image is drawn with 0 alpha and thus not visible as one would expect when passing in Color( 255, 0, 0, 0 ).

nkast:
@prime31 That's because SpriteBatch use/expect premultiplied colors. 
If you premultiply Color( 255, 0, 0, 0) you will get Color( 0, 0, 0, 0). Same result, lighter shader.

If you don't want to use premultiplied colors then simply changing the BlendState in SpriteBatch.Begin() from `AlphaBlend` (default) to `NonPremultiplied` might work but it will be slower as well.

nkast:
Draw_with_alpha_blending test fixes #4963

@prime31 look at the last image at https://github.com/mono/MonoGame/pull/4963/files Draw_with_alpha_blending__White__80_.png
The transparent image used to be overburned, the new image that use pre-multiplied Color.White is what you expect from a 30% transparency.

MonoGame.Framework/Graphics/Effect/Resources/SpriteEffect.fx
prime31: @tomspilman totally your call. I personally already replaced SpriteBatch and SpriteEffect with my own implementations. This is just one of those things that always bugged me (and that confused people coming from Unity) so I figured I would send over the commit before 3.6 drops.

People new to MonoGame would not expect passing in a 0 alpha color to SpriteBatch.Draw to display the way that it does. I have never seen any other engine or framework that would have that result. For example, here is the same image in Unity with the same tint color (255, 255, 255, 100) as the [bottom image](https://github.com/mono/MonoGame/pull/4962#issuecomment-228577049):

screen shot 2016-06-25 at 5 39 36 pm

It just seems way more logical to be able to pass in a 0 alpha color and get back a 0 alpha texture but perhaps its just me ¯\\_(ツ)_/¯

prime31:
@nkast using Color.FromNonPremultiplied is exactly what the shader change is doing. The difference being folks using MonoGame don't have to bother premultiplying Colors every time they want to draw something. It makes MonoGame mimic other game engines/frameworks (Unity, Unreal, Love2D, libGDX, and tons more I'm sure all work this way).

KonajuGames:
The shader is correct for use with pre-multiplied alpha as is used throughout XNA 4.0 and MonoGame.  As stated, Color(255, 0, 0, 0) is not a valid pre-multiplied alpha color.  To apply an alpha to a color, multiply the full Color value by the alpha value, e.g. Color.Red * 0 for your example above.

Closing this PR.

MonoGame.Framework/Graphics/Effect/Resources/SpriteEffect.fx
tomspilman: > People new to MonoGame would not expect passing in a 0 alpha color to SpriteBatch.Draw to display the way that it does.

I understand what you are saying, but MonoGame and XNA have used pre-multiplied alpha as the default for over 6 years now.  We're not going to break every shipping XNA and MonoGame title by changing things to the old and less useful "interpolative alpha" technique now.

IMO this is purely an issue of education which we've done a bad job of.  Specifically we need to fix this:

http://www.monogame.net/documentation/?page=F_Microsoft_Xna_Framework_Graphics_BlendState_AlphaBlend

To include details about the premultiplied alpha blend mode that it is and suggest using `BlendState.NonPremultiplied` if they want interpolative alpha blending.

We should also fix the docs here:

http://www.monogame.net/documentation/?page=M_Microsoft_Xna_Framework_Graphics_SpriteBatch_Begin

... to remind people that the default behavior of `SpriteBatch` is to use pre-multiplied alpha blending.

PR #4965, opened 2016-06-27
Fix preprocessor directive bug

TD25:
I think this preprocessor directive is misspelled. I wasn't getting input and buttons weren't displayed in my wp8.1 project because of it.


KonajuGames:
Good catch.  Thanks @TD25. Merging.

Issue #4947, opened 2016-06-16
(PipelineTool) Enter key does not end text editing when typing in a new value for a property.


jamesford42:
Enter key does not end text editing when typing in a new value for a property.

cra0zy:
@tomspilman This has been fixed and can be closed.

PR #4966, opened 2016-06-27
[Pipeline Tool] Make PropertyGrid Extendable

### LINKED ISSUE #4946: (PipelineTool) Can not edit (int) type properties.

cra0zy:
Custom cells for propertygrid can now be added from a dll that is referenced in a project, I'll write some documentation on this later on.

Also fixes #4946 and #4947

### DIFF
--- a/Tools/Pipeline/Controls/PropertyCells/CellBase.cs
+++ b/Tools/Pipeline/Controls/PropertyCells/CellBase.cs
@@ -8,6 +8,18 @@ using Eto.Forms;
 
 namespace MonoGame.Tools.Pipeline
 {
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, AllowMultiple = true)]
+    public class CellAttribute : Attribute
+    {
+        public string Name;
+        public Type Type;
+
+        public CellAttribute(Type type)
+        {
+            Type = type;
+        }
+    }
+
     public class CellBase
     {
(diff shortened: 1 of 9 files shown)


cra0zy:
This just makes it so end user can add a cell for editing for anything that is not a built in C# property, since those are the only ones the property grid will support.

cra0zy:
Just FYI, if you don't like the idea all I have to do is remove my changes from PipelineController.cs, the rest of the changes are still needed.

KonajuGames:
I'm ok with this.  I wouldn't expect it to get used a lot though.

cra0zy:
Any chance of this getting merged any time soon? I don't mind removing the few lines from the PipelineController.cs file as I mentioned above if it would help.

tomspilman:
> I don't mind removing the few lines from the PipelineController.cs file as I mentioned above if it would help.

How about you just comment that bit out for now leaving a `TODO` to enable it later once we have time to document and test it all.

I just don't want to introduce more complications until we get the core of the Pipeline tool back to working as well as it used to before the Eto changes.

cra0zy:
> How about you just comment that bit out for now leaving a TODO to enable it later once we have time to document and test it all.

I think just removing it seems cleaner for the code base, anyway done.

cra0zy:
The build finished.

tomspilman:
Thanks... merging!

Issue #4948, opened 2016-06-17
GamePad issues


endimilojkoski:
1. Connecting X360 Gamepad while game is running the game crashes. It tries to add deviceId=0 key which already exist and I have only 1 gamepad connected.
![image](https://cloud.githubusercontent.com/assets/13436441/16155581/5054cae6-34b0-11e6-8f86-0ba0b505bd23.png)

2. When starting the game with 1 X360 and 2 generic gamepads the game starts fine and the X360 gamepad works but the 2 generic ones are not recognised.
In this scenario I see 2 GamePads and 4 Joysticks, each have +1 unknown device. So GamePad gets only the X360 and Joystick get all gamepads connected. This is where they are assigned:
![image](https://cloud.githubusercontent.com/assets/13436441/16155687/c960f00e-34b0-11e6-801c-6b95f7fa36fd.png)

3. On my old MacBook the X360 gamepad works fine if connected before starting the game but for a friend of mine with newer MacBook the game crashes if his X360 gamepad is connected. This is already reported in issue #4932 

In 3.5 gamepads were working great through GamePad. All were recognised and I could plug them while in game and they would work with no issues.

@cra0zy @mrhelmut

mrhelmut:
1. I will take a look at this, but I don't know when yet. It's on my to do list, I will need it to work too.

2. If gamepads are not recognized, it means that they are not in the "known gamepads" database. Feel free to add them by adding their GIUD and mapping to this file: https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/SDL/gamecontrollerdb.txt

3. Should be fixed by #4949 (let me know if it works).

endimilojkoski:
Yup, #4949 fixes the Mac game crash with gamepad plugged in.
So far latest version is working fine with no critical issues on all 3 platforms.

About #2, can't you somehow include the old OpenTK definitions or this is different?
Even if I add my 2 joysticks, how can I be sure that many other joysticks that I don't have will work for the players?

endimilojkoski:
One of my gamepads that is not recognised is iBuffalo and entry already exists:

83056020000000000000504944564944,iBuffalo USB 2-axis 8-button Gamepad,a:b1,b:b0,y:b2,x:b3,start:b7,back:b6,leftshoulder:b4,rightshoulder:b5,leftx:a0,lefty:a1,platform:Windows,

[...]

PR #4968, opened 2016-06-28
Improve and simplify Joystick and GamePad detection

cra0zy:
Notes:
 - Instead of trying to use instanceid/deviceid, now we are using our own ids
  - when device gets added we just use the first empty slot
  - when device gets removed we search the dictionary and remove the device with specified instanceid once found
 - GameControllerAdded event cannot be used since it doesn't always get triggered, so JoystickAdded event is used for GamePad
 - Fixes: #4948 

CC. @mrhelmut @endimilojkoski

### DIFF
--- a/MonoGame.Framework/Input/GamePad.SDL.cs
+++ b/MonoGame.Framework/Input/GamePad.SDL.cs
@@ -21,9 +21,6 @@ namespace Microsoft.Xna.Framework.Input
 
         private static readonly Dictionary<int, GamePadInfo> Gamepads = new Dictionary<int, GamePadInfo>();
 
-        // we have to maintain this mapping because instance IDs are not ordered by player index (i.e. player lights on Xbox gamepads), but DeviceID are
-        private static readonly Dictionary<int, int> _deviceInstaceToId = new Dictionary<int, int>();
-
         private static Sdl.Haptic.Effect _hapticLeftRightEffect = new Sdl.Haptic.Effect
         {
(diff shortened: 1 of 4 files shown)


endimilojkoski:
I just tried this fix and the GamePads are working as they should now.
Great job @cra0zy !

endimilojkoski:
I also tested this on Linux and Mac and it works with no issues.
Gamepads (X360 and iBuffalo) are working fine if they are plugged before starting and also work fine if I plug them while in game.
Also working fine if I plug/unplug multiple times in game.

mrhelmut:
Awesome.

Do you think we should totally drop ```Sdl.EventType.ControllerDeviceRemoved``` as well and remove gamepads in ```Joystick.RemoveDevice()```?
I don't know if this can happen, but I see a case where a gamepad could be added before a slot has been correctly removed (and have ID mismatch with its joystick, it's no big deal though).

Good to go as-is anyway. :)

Unrelated, but this removes the dependency to ```PeepEvent``` and may solve a marshaling issue @dellis1972 encountered with it.

cra0zy:
> I don't know if this can happen, but I see a case where a gamepad could be added before a slot has been correctly removed (and have ID mismatch with its joystick, it's no big deal though).

If joystick is not a gamepad a mismatch will occur regardless... totally normal.

mrhelmut:
You're right, then go go! :)

tomspilman:
Thanks for working this out @cra0zy @mrhelmut and @endimilojkoski !

Issue #4910, opened 2016-06-07
Remove GamePadDPad(params Buttons[] buttons) Ctor


tomspilman:
We need to remove the constructor `internal GamePadDPad(params Buttons[] buttons)` and replace any usages of it with a non-param array constructor:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Input/GamePadDPad.cs#L77

This ctor only serves to generate array garbage and from what I can tell is unused in MonoGame.

06needhamt:
See PR #4970

PR #4970, opened 2016-06-28
Remove array in GamePadDPad constructor

06needhamt:
This PR fixes issue #4910

### DIFF
--- a/MonoGame.Framework/Input/GamePadDPad.cs
+++ b/MonoGame.Framework/Input/GamePadDPad.cs
@@ -42,8 +42,8 @@ using System;
 
 namespace Microsoft.Xna.Framework.Input
 {
-	public struct GamePadDPad
-	{
+    public struct GamePadDPad
+    {
         public ButtonState Down
         {
             get;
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/Input/GamePadDPad.cs
```diff
@@ -74,7 +74,7 @@ public GamePadDPad(ButtonState upValue, ButtonState downValue, ButtonState leftV
             Right = rightValue;
         }
 
-        internal GamePadDPad(params Buttons[] buttons)
+        internal GamePadDPad(Buttons[] buttons)
```
tomspilman: @06needhamt - I think I misled you a bit in https://github.com/mono/MonoGame/issues/4910.  We need to remove the array completely...  there is no need for any array here.  The function should be `internal GamePadDPad(Buttons buttons)`.

This of course means the `foreach` loop is removed with it just testing the single `buttons` type.

06needhamt:
Ah i see will change it now

06needhamt:
does the game pad state constructor need to be changed too or should it take buttons[0]?

tomspilman:
> does the GamePad State constructor need to be changed too or should it take buttons[0]?

Yes...  it should be changed too.  There is no need for any array passed or used in the `GamePadDPad` constructor.

06needhamt:
Everything is fixed now i will create a new PR so there is only one commit

06needhamt:
Im not sure what happened here i rebased the PR but it is still showing the individual commits

KonajuGames:
I've always found rebasing to be a black magic thing that I can rarely get right.  We can squash the commits when we merge the PR.

06needhamt:
ok thanks a lot this is my first contribution so i was unfamiliar witth the codebase hence so many commits. also i have errors on certain platforms when trying to build on my machine so i am unable to test all platforms

tomspilman:
This looks to be working...  merging!

Thanks @06needhamt !

nkast:
This one broke XNA compatibility.
FarseerPhysics samples was using it for Virtual Gamepad.

tomspilman:
Must be an undocumented ctor.  I'm fine with adding it back as long as the internal code doesn't use it.

Halofreak1990:
> Must be an undocumented ctor. I'm fine with adding it back as long as the internal code doesn't use it.
Just checked the XNA documentation for the [GamePadDPad](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.input.gamepaddpad.gamepaddpad(v=xnagamestudio.40).aspx) structure, and it doesn't have a constructor taking an array. Never had (a public) one, either.

nkast:
@Halofreak1990 

The PR changed the ctor of  `GamePadState` as well.
https://msdn.microsoft.com/en-us/library/bb975283.aspx

It changed the Test for GamePadState as well, which probably broke the Tests for XNA (or should have).

PR #4972, opened 2016-06-29
Threading and GL resource loading

mrhelmut:
This PR is a follow up to #4927 discussion.

It reverts the use of a background context for loading OpenGL assets (which may expose uncontrollable GL driver issues if threaded), and adds thread blocking to texture reader so that all mipmap levels are loading during a single frame (instead of having to wait for the next frame to load the next mipmap level).

@KonajuGames Does this look good to you following the discussion?

### DIFF
--- a/MonoGame.Framework/Content/ContentReaders/Texture2DReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/Texture2DReader.cs
@@ -65,99 +65,105 @@ namespace Microsoft.Xna.Framework.Content
 			}
 			
             texture = existingInstance ?? new Texture2D(reader.GraphicsDevice, width, height, levelCountOutput > 1, convertedFormat);
-			
-			for (int level = 0; level < levelCount; level++)
-			{
-				var levelDataSizeInBytes = reader.ReadInt32();
-                var levelData = reader.ContentManager.GetScratchBuffer(levelDataSizeInBytes);
-                reader.Read(levelData, 0, levelDataSizeInBytes);
-                int levelWidth = width >> level;
-                int levelHeight = height >> level;
-
-                if (level >= levelCountOutput)
-                    continue;
-
-				//Convert the image data if required
-				switch (surfaceFormat)
-				{
-					case SurfaceFormat.Dxt1:
-                    case SurfaceFormat.Dxt1SRgb:
-                    case SurfaceFormat.Dxt1a:
-                        if (!reader.GraphicsDevice.GraphicsCapabilities.SupportsDxt1 && convertedFormat == SurfaceFormat.Color)
-						    levelData = DxtUtil.DecompressDxt1(levelData, levelWidth, levelHeight);
-						break;
-					case SurfaceFormat.Dxt3:
-					case SurfaceFormat.Dxt3SRgb:
-                        if (!reader.GraphicsDevice.GraphicsCapabilities.SupportsS3tc)
-                        if (!reader.GraphicsDevice.GraphicsCapabilities.SupportsS3tc && convertedFormat == SurfaceFormat.Color)
-						    levelData = DxtUtil.DecompressDxt3(levelData, levelWidth, levelHeight);
-						break;
-					case SurfaceFormat.Dxt5:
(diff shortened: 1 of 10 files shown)


KonajuGames:
That looks good.  Did you test the performance difference?

mrhelmut:
I'm not using that many mipmap and I haven't noticed significant differences with loading textures.

I believe @endimilojkoski is using a lot of textures with mipmap, his game should be a good benchmark to compare pre-#4927 & post-#4927 with this PR. Would you mind comparing, Endi?

endimilojkoski:
I just tried it and now it is roughly 4-5 times slower.

KonajuGames:
> I just tried it and now it is roughly 4-5 times slower.

Slower than threaded, or slower than the previous invoke on main thread approach?

endimilojkoski:
I am using only 1 thread to load all my content on start up and don't use threads in my game after that.

With #4927 it became much faster. Loading time was around 2-3 seconds and now with reverting that and the latest changes it takes roughly around 10 seconds.
Now it is roughly the same as it was before implementing #4927.

KonajuGames:
It seems to be lose/lose if you want to load in the background with OpenGL.  With threaded, the low quality drivers cause random crashes.  With invoked and the synchronous nature of ContentManager (one resource at a time in sequence), it is a lot slower.

endimilojkoski:
In 3.5 with OpenTK it was also fast and now the difference in loading time is pretty big.

But you know better than me, I am just testing and letting you know what happens, not an expert on this field, this is my first game :)

mrhelmut:
That sounds very huge. I've noticed about 10% difference at worst on my side. I'll get in touch with you to see if textures can be optimized/packed for your game.

Note that the OpenTK implementation was also having driver crashes when threading texture loading.

This is the safest route, which unfortunately having the drawback to sync with frames.

endimilojkoski:
And it is a bit worse.
I just realised that the ~2s loading was from HDD and the ~10s was from SSD.
Moved it to HDD and it is additional ~2s slower.

mrhelmut:
I don't have any relevant performance hit, everything in our game is packed in about 3 textures (max 2048x2048, and no mipmap), even though the XNBs are compressed. Hence, our game is not a good benchmark in that regard.

@KonajuGames I believe we are fine, if it looks good to you. I'm discussing privately with @endimilojkoski to understand the huge difference in his game.

dellis1972:
@mrhelmut @KonajuGames what is the status of this on  ... merge or !merge ? 
Could we make it optional somehow? Via a setting

KonajuGames:
I don't know if making it an option is a wise choice.  If something is unstable, liable to crash due to driver issues, and beyond our control, how can we make that a choice?

dellis1972:
fair point. I'd say merge this one then.

endimilojkoski:
Yes, I am loading 138 textures on startup and that might be the slowdown issue. It was always fast so I never bothered with that. Started packing the textures and will let you know how it changes my loading times, now I am down to 100 from 138 and will further pack it in the upcoming updates.

KonajuGames:
It is a side effect of the current ContentManager API that forces loading to be synchronous, i.e. it doesn't start loading the next asset until the current asset is completely loaded.  We could minimize this by introducing a new ContentManager API that would allow many assets to be loaded simultaneously behind the scenes (provide a collection of filenames, then once they are loaded use the regular `Load<T>()` to get the reference to the previously loaded asset).  This should allow a few textures to be passed to the main thread on the same frame, rather than one per frame.

mrhelmut:
Should we merge this in the meantime?

mrhelmut:
Hope that you don't mind that I up this one. ^^

endimilojkoski:
I hope you are not waiting for my response :)
Do what is best, I am slowly packing my textures anyway so it should start loading faster.

KonajuGames:
We'll merge this one now (a bit slower but more reliable is better than higher performance but liable to crash) and try to improve the performance issues later.  Thanks @mrhelmut.

KonajuGames:
Good to hear.

PR #4974, opened 2016-07-01
Added support for XACT reverb and filter effects

tomspilman:
Further work on XACT support to make more correct to XNA. The major changes in this one:

 - Support for reading reverb info and applying the effect.
 - Support for reading low/band/high pass filters and applying the effect.
 - Fix cases where track variations were not being set correct.

Note that the reverb and filter effects only work on XAudio2/Windows platforms... we still need implementations of them for OpenAL and other audio backends.

As part of some of the changes some tabs to spaces were changed, so use this link to see the actual differences:

https://github.com/mono/MonoGame/pull/4974/files?w=0


tomspilman:
This could use some testing by anyone using XACT in their current title to be sure nothing got broken in these changes.

tomspilman:
@dellis1972 - So do you think these changes are good after testing them?

tomspilman:
@dellis1972 @KonajuGames - Make sure we squash-merge this when we do merge it.

dellis1972:
@tomspilman I've been looking at the OpenAL side of things.. Not had a chance to test the DirectX bits (I'm on a mac most of the time). That said I have not seen any crashes or issues with the cue changes etc.

tomspilman:
@dellis1972 - Well mostly interested in seeing that it doesn't break Cue playback behavior.

Still hoping someone else can come along and test, but if not lets look to merge this later this week anyway.

tomspilman:
@KonajuGames  @dellis1972 

Did some testing on another game... this should be good to squash-merge now.

PR #4978, opened 2016-07-04
Fixes the recent addition of OpenGL version initialization.

Cardanis:
The existing logic didn't take into account the difference in version string on OpenGL ES, causing it to always default to version 1.1 .  This adds logic specific to OpenGL ES that properly gets the version number.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
@@ -212,10 +212,10 @@ namespace Microsoft.Xna.Framework.Graphics
             {
                 string version = GL.GetString(StringName.Version);
                 string[] versionSplit = version.Split(' ');
-                if(version.Length > 2 && version[0].Equals("OpenGL") && version[1].Equals("ES"))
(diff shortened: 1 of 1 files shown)


tomspilman:
@dellis1972 @cra0zy @mrhelmut - Thoughts?

mrhelmut:
Totally fine. Thanks @Cardanis for spotting this difference.

dellis1972:
We don't support GLES 1.1 (or OpenGL 1.1 for that matter) I wonder if we should change the defaults..
Otherwise this looks fine

KonajuGames:
There is no point falling back to GL ES 1.1 because we don't support it.  The minimum for GL ES must be 2.0.

mrhelmut:
I believe the GL version is read only to check if the context is 3.0 to consider that the FBO functions are core. Behind this check, the GL version string is ignored and it falls back on trying to load FBO functions with the extension.

KonajuGames:
I'll merge this one.  The fallback to 1.1 was previously there, and if any device tries to fallback to 1.1 it has bigger issues to solve.  Thanks @Cardanis.

PR #4985, opened 2016-07-06
A few misc fixes from console work

tomspilman:
- Added `XBoxOne` and `PSVita` platforms to work around Protobuild extension limitation.
- Allow `ImmutableVertexInputLayout` to be used by other platforms.
- Fixed some copyright headers.


tomspilman:
Note need access to the private Xbox One or PSVita repositories in order to use these platforms.  The change here just makes the Protobuild system aware those platforms may exist.

KonajuGames:
Merging.  I'll add the other platform to it soon.

PR #4987, opened 2016-07-06
New SetRenderTargets() method which allows for variable target count.

tomspilman:
This is a follow up to https://github.com/mono/MonoGame/pull/4853.

I've added a new render target setting call:

```
public void SetRenderTargets(RenderTargetBinding[] renderTargets, int count)
```

This allows you to specify how many targets in the array to set which lets you more easily reuse the array.  For example:

```
const int MaxTargets = 4;
RenderTargetBinding[]  targets = new RenderTargetBinding[MaxTargets];
int targetsCount;

// Get the current targets.
targetsCount = device.RenderTargetCount;
device.GetRenderTargets(targets);

// Set some new targets.
targets[0] = someTarget;
targets[1] = otherTarget;
device.SetRenderTargets(targets, 2);
```

With this I also marked old garbage generating GetRenderTargets() and SetRenderTargets() as obsolete with a comment specifying they are there only for XNA compatibility.

Thoughts?

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -561,7 +561,7 @@ namespace Microsoft.Xna.Framework.Graphics
         internal void OnPresentationChanged()
         {
             CreateSizeDependentResources();
-            ApplyRenderTargets(null);
+            ApplyRenderTargets(null, 0);
         }
 
 #endif
@@ -834,7 +834,7 @@ namespace Microsoft.Xna.Framework.Graphics
         internal void OnPresentationChanged()
         {
             CreateSizeDependentResources();
-            ApplyRenderTargets(null);
+            ApplyRenderTargets(null, 0);
         }
 
 #endif // WINDOWS
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -51,9 +51,9 @@ namespace Microsoft.Xna.Framework.Graphics
         private IndexBuffer _indexBuffer;
         private bool _indexBufferDirty;
 
+        private readonly RenderTargetBinding[] _tempRenderTargetBinding = new RenderTargetBinding[1];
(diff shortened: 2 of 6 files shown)


KonajuGames:
I think with a new API, especially where it obsoletes an older API, should have the XML doc added so it is documented from the start.

KonajuGames:
Aside from the need for the doc, I think it is a good change.  Enforces good behaviour.

dellis1972:
+1 on Docs. otherwise LGTM

tomspilman:
Wondering now if...
```
targetsCount = device.RenderTargetCount;
device.GetRenderTargets(targets);
```
... should instead be:
```
device.GetRenderTargets(targets, out targetsCount);
```
It looks a little nicer, but more important ensures the caller is aware that only some of the results in the array are valid.

Note this is necessary because a `null` render target means the "default backbuffer" and not that a render target is not set on that slot.

tomspilman:
After some discussion with @prime31 I am now considering an alternative design:

```
RenderTargetBindingCollection  targets = new RenderTargetBindingCollection();

// Get the current targets.
device.GetRenderTargets(targets);

// Set some new targets.
targets.Clear();
targets[0] = someTarget;
targets[1] = otherTarget;
device.SetRenderTargets(targets);
```

This hides the arrays and makes things more similar to how `TextureCollection` and `SamplerStateCollection` work instead of having the user manage arrays themselves.

KonajuGames:
I like that.

tomspilman:
Give me a day or so to refactor this.

tomspilman:
Ummm.... i'm confused now.  Is setting the `null` render target (aka the back buffer) illegal to do via `SetRenderTargets(RenderTargetBinding[])`?

KonajuGames:
I can't say I've ever tried it.

KakCAT:
Did this RenderTargetBindingCollection or similar moved to another issue/PR ? 

I'm having to write ugly code with GetRenderTarget/SetRenderTarget pairs in order to avoid garbage. Most of the time I don't know how many RTs are active when backing them up so I have to keep an object with all possible RenderTargetBinding array sizes. 

I was wondering why something like this RenderTargetBindingCollection wasn't implemented and I stumbled upon this PR,

PR #4988, opened 2016-07-06
Removed unnessasary dictionary lookup for user index buffers for DirectX platforms.

tomspilman:
This should be a small performance improvement in sprite batch and other draw user primitive calls... but generally makes things less complex.

Also fixed disposal of user buffers.


KonajuGames:
Merging.

PR #4993, opened 2016-07-07
Fix issues with modifying presentation parameters with multisampling enabled

hach-que:
When you enable multisampling at game startup, it determines the maximum multisampling count allowed by the hardware and configured both the swap chain and depth stencil view with it.

However, when you `ApplyChanges` at runtime, it had a hard-coded value of 4 if multisampling was enabled.  Due to this, if you modified the presentation parameters (in my specific case it was enabling window resizing), it would cause the multisampling count of the swap chain (now 4) and the multisampling count of the depth stencil view (still 16 in my case) to get out of sync, which is not permitted in DirectX.  The end result of this was that instead of rendering working correctly, the window would be entirely black, with nothing rendered.

This moved the maximum multisampling count logic into a platform-specific method on the `GraphicsDevice`, and ensures that method is called whenever the multisampling count needs to be adjusted by MonoGame.

It also addresses a scenario where a `RenderTarget` could be created with a different multisampling count on the depth stencil view than the underlying texture resource.  This update texture to keep track of the sample description used when creating the resource, and forces the logic to use the same parameters when creating the depth stencil view, to avoid the same scenario occurring on render targets as was occurring on the swap chain.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -231,7 +231,7 @@ namespace Microsoft.Xna.Framework.Graphics
         {
 #if DEBUG
             var debugLevel = SharpDX.Direct2D1.DebugLevel.Information;
-#else 
+#else
             var debugLevel = SharpDX.Direct2D1.DebugLevel.None; 
 #endif
             // Dispose previous references.
@@ -565,6 +565,15 @@ namespace Microsoft.Xna.Framework.Graphics
         }
 
 #endif
+
+#if WINDOWS_STOREAPP || WINDOWS_UAP || WINDOWS_PHONE
+
+        internal void PlatformSetMultiSamplingToMaximum(PresentationParameters presentationParameters, out int quality)
+        {
+            quality = (int)SharpDX.Direct3D11.StandardMultisampleQualityLevels.StandardMultisamplePattern;
+        }
+
+#endif
 #if WINDOWS
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


hach-que:
@tomspilman @KonajuGames I've rebased this PR, and included proper fixes for #5083 as well.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -646,6 +651,40 @@ protected virtual void CreateDeviceResources()
...
         }
+        
+        internal void PlatformSetMultiSamplingToMaximum(PresentationParameters presentationParameters, out int quality)
+        {
+            var format = presentationParameters.BackBufferFormat == SurfaceFormat.Color ?
+                               SharpDX.DXGI.Format.B8G8R8A8_UNorm :
+                               SharpDXHelper.ToFormat(presentationParameters.BackBufferFormat);
+
+            // Check that the multisample count specified by the game is valid.
+            var msc = PresentationParameters.MultiSampleCount;
```
hach-que: This checks that the multisample count value is valid (power of 2 or zero).

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -646,6 +651,40 @@ protected virtual void CreateDeviceResources()
...
+            {
+                throw new ApplicationException(
+                    "The specified multisample count is not a power of 2 (it must be " +
+                    "a value such as 0, 1, 2, 4, 8, 16, 32, etc.).");
+            }
+
+            // Find the maximum supported level starting with the game's requested multisampling level
+            // and halving each time until reaching 0 (meaning no multisample support).
+            var qualityLevels = 0;
+            var maxLevel = msc;
```
hach-que: By setting the maximum level to the game's desired level, it resolves @KonajuGames's issue as well.

hach-que:
@tomspilman @KonajuGames Ping, can someone take a look at this?

hach-que:
Builds are now passing on this.

KonajuGames:
Merging. Thanks @hach-que

PR #4997, opened 2016-07-08
Fixes #4977

### LINKED ISSUE #4977: Problem with Multiple Mgcb's
Hello,
I am attempting to split out some content so I can use 2 seperate `ContentManager` instances.

I assumed if I referenced two seperate `mgcb` files from my project, that I could get them to output their content to different directories when I build my project.

When creating a `ContentManager` you need to give it a root directory, so I assumed I could create one for each content directory where my `mgcb` content is output:

```
 var contentOne = new ContentManager(this.Services, "Content\\ScreenManagement");
 var contentTwo = new ContentManager(this.Services, "Content\\Models");
```

The problem is, when I build the project, the assets from the multilple referenced `mgcb`s are always output directly to the `content` folder under py projects output directory. Furthermore, the 1st output is cleared when the 2nd `mgcb` is built, and so I actually only end up with the content of the 2nd `mgcb` file in by output directory.

Here is how it looks (note I added the `mgcb` files using `Add as Link`

image

It doesn't seem to matter which folder in my project that I add the reference to the `mgcb` too, when I build they always seem to output to the outputDir/Content folder for my project.

This means I don't end up with seperate root paths to create my ContentManager instances for, and also the build clearing the content output by the first mgcb when it builds the second one is obvioulsy a bit problematic!

Any ideas?

dazinator:
Modified msbuild targets file, so that when mgcb's are built, their build output is copied to a directory in the output folder that is relative to the path of the mgcb within the project. This allows the content that is output (.xnb's) to more easily be isolated, by moving mgcb's to seperate folders within the project.


dazinator:
I'm close with this, but just doing some more testing.. and it has revealed an issue where all xnb's are ending up in the same folder, i.e its not preserving the original relative directory. Looking at this now

tomspilman:
No problem @dazinator .  Keep at it and a few others will come in and start testing it too.

dazinator:
Nice - I think this should work now according to my local testing. Let me know if any problems. Cheers

dazinator:
I have noticed one issue with this remaining still..

It works fine for `mgcb`s added directly to the project. Probably because the physical file path of the mgcb file mirrors where you add it within the project.

But for `mgcb` files added as `linked` files, it doesnt matter what project folder you add those linked mgcb files too, the mgcb content is output to the projects `bin/windows/Content` folder.

For example here is my project, containing linked and non-linked mgcb files:

image

The non linked mgcb files have their content output correctly at: `\bin\windows\debug\content\` 

image

The linked `mgcb` files are all output here:

image

The linked mgcb files actually physically live next to each other so it makes sense that they would be output to the same directory, because the path within the project is not being honoured - only their relative location on disk.

tomspilman:
> because the path within the project is not being honoured - only their relative location on disk.

I would think people would expect the relative location within the project path (and not the on disk path) is likely what it should follow.

Not sure how we access that in MSBuild, but it should be in there somewhere.

dazinator:
Yeah - think I have figured out the issue..

These are the mgcb's in the project file:

```
 <MonoGameContentReference Include="..\Content\Models.mgcb">
      <Link>Content\Test\Models.mgcb</Link>
    </MonoGameContentReference>
    <MonoGameContentReference Include="..\Content\ScreenManagementContent.mgcb">
      <Link>Content\ScreenManagementContent.mgcb</Link>
    </MonoGameContentReference>
    <MonoGameContentReference Include="Content\aaa.mgcb" />
    <MonoGameContentReference Include="Content\Test\bbb.mgcb" />

```

I currently use %(RelativeDir) for those items which `Contains the path specified in the Include attribute, up to the final backslash (\)` - which works for non linked ones, but as you can see, the linked ones I actually need to use the  <Link>Content\ScreenManagementContent.mgcb</Link> value..
I'll have another go at that..

dazinator:
BOOOOM! It now works with linked files also. 
I know that usually, the word BOOM precedes something that is broken, but not in this case.. hopefully :)

MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
```diff
@@ -36,8 +36,8 @@
     <!-- We do this here so we are compatible with xbuild -->
     <ItemGroup>
       <ContentReferences Include="@(MonoGameContentReference)">
-        <ContentOutputDir>$(ProjectDir)%(RelativeDir)\bin\$(MonoGamePlatform)</ContentOutputDir>
-        <ContentIntermediateOutputDir>$(ProjectDir)%(RelativeDir)\obj\$(MonoGamePlatform)</ContentIntermediateOutputDir>
+        <ContentOutputDir>$(ProjectDir)%(RelativeDir)\bin\$(MonoGamePlatform)\%(Filename)</ContentOutputDir>
```
dazinator: It was important to set the output folder and intermeditate output folder passed to MGCB.exe as ones that are unique to the mgcb being built. As otherwise, when you have multiple mgcb's in the same directory, they all end up outputting to the same folders, and this was screwing up the results of the build.

MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
```diff
@@ -84,10 +84,13 @@
           Command="$(MonoGameContentBuilderCmd) /@:&quot;%(ContentReferences.FullPath)&quot; $(Header) /outputDir:&quot;%(ContentReferences.ContentOutputDir)&quot; /intermediateDir:&quot;%(ContentReferences.ContentIntermediateOutputDir)&quot; /quiet"
           WorkingDirectory="%(ContentReferences.RootDir)%(ContentReferences.Directory)" />
 
-   
+
+    <CreateItem Include="%(ContentReferences.ContentOutputDir)\**\*.*" Condition="'%(ContentReferences.Link)' != ''" AdditionalMetadata="ContentOutputDir=$([System.String]::Copy('%(ContentReferences.Link)').Replace('%(Filename)%(Extension)', ''))">
```
dazinator: What I am doing here, is for `ContentReferences` that are `linked` files, then it will get it's `Link` metadata property to  work out what the path of it is within the project structure. I have to get rid of the filename itself as I only care about the folder it lives in. 

For non linked `ContentReferences` - the below CreateItem just uses the RelativeDir = which should be the relative dire

If the `ContentReference` is not a linked file, then I use the RelativeDir = which is the location of it within the project.

MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
```diff
@@ -84,10 +84,13 @@
...
+      <Output TaskParameter="Include" ItemName="ExtraContent" />
+    </CreateItem>
     <CreateItem Include="%(ContentReferences.ContentOutputDir)\**\*.*" AdditionalMetadata="ContentOutputDir=%(ContentReferences.RelativeDir)">
       <Output TaskParameter="Include" ItemName="ExtraContent" />
-  
+
```
dazinator: Oops - added some whitespace here accidentally!

dazinator:
Ok now i'm done ;-)

mrhelmut:
This looks nice, I'll give a test soon.

dellis1972:
@mrhelmut did you manage to test this? It looks good to me. Would be nice to test it on Mac/Linux though.

mrhelmut:
Sorry, I'm in the Gamescom rush. :(

dazinator:
Let me know if you need me to squash the commits, or if there is anything else you need from me!

Jjagg:
We should get this tested and merged

Jjagg:
Just tested this and it's working great! I tested with the following structure with 4 .mgcb files. One in the project root, one linked in the project root, one in a Content folder and one linked in a Linked folder. Each has one spritefont file and one subfolder with another spritefont file.
```
ProjectRoot
├───Content\
│   └───Content.mgcb
├───Linked\
│   └───Linked.mgcb
├───Root.mgcb
└───LinkedRoot.mgcb
```
The result is the following as expected:
```
BinFolder
├───Content
│   ├───ContentSub\
│   │   └───ContentSubFont.xnb
│   └───ContentFont.xnb
├───Linked
│   ├───LinkedSub
│   │   └───LinkedSubFont.xnb
│   └───LinkedFont.xnb
├───RootLinkedSub
│   └───RootLinkedSubFont.xnb
├───RootSub
│   └───RootSubFont.xnb
├───RootLinkedFont.xnb
└───RootLinkedFont.xnb
```

Great job with this @dazinator! LGTM!

dazinator:
Thanks @Jjagg. I tested on Windows, but wasn't sure if this needed to be tested on any other platforms.

Jjagg:
I also tested on Windows. The change aren't really platform specific, but if needed I'll give this a go on Linux tomorrow I have the project set up now anyway.

dellis1972:
Testing under xbuild is important as it behaves differently that msbuild.
Also its possible that some thing flat won't work.

Jjagg:
Alright, so just Linux is fine?

dazinator:
Never used xbuild, or linux, so I am afraid i can't help with testing this any further :(

MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
```diff
@@ -11,41 +11,35 @@
...
     </PropertyGroup>
 
     <!-- Get all Mono Game Content References and store them in a list -->
     <!-- We do this here so we are compatible with xbuild -->
     <ItemGroup>
-      <ContentReferences Include="@(MonoGameContentReference)"/>
-     </ItemGroup>
+      <ContentReferences Include="@(MonoGameContentReference)">
+        <ContentOutputDir>$(ProjectDir)%(RelativeDir)\bin\$(MonoGamePlatform)\%(Filename)</ContentOutputDir>
+        <ContentIntermediateOutputDir>$(ProjectDir)%(RelativeDir)\obj\$(MonoGamePlatform)\%(Filename)</ContentIntermediateOutputDir>
```
Jjagg: You can drop the backslash before 'bin' and 'obj', RelativeDir already has a backslash at the end.

Jjagg: It seems RelativeDir and Filename are sometimes not correctly when using xbuild.
xbuild isn't very helpful in its error message:
`Error building target Prepare: Object reference not set to an instance of an object`

... That's it :/ I only figured out that RelativeDir and Filename are the culprits by putting some messages around it.

Another issue is that MonoDevelop is very bad at handling links. Or rather it doesn't create links at all. It just includes the file like it normally would... So you have to dig into the msbuild files to get that right.

dazinator:
Ah - thanks for giving this a test with xbuild. Sad news though.

Looks like I'll need to set up an xbuild environment and figure out how to fix that null ref issue. Not sure when I am going to have time for that at the moment though. If someone more familiar with xbuild can pick this up I'd be greatful.

The other option is that rather than getting this working for xbuild, we find a way to make this feature msbuild only, and make xbuild use the legacy build logic

dellis1972:
can someone put up a sample Content.mgcb with all the relevant link tests
(if its not been done). That way people who can test this will be working
on the same thing.

Jjagg:
Good idea, I'll get a project up on GitHub later today.

Jjagg:
> I'll get a project up on GitHub later today.

Done! https://github.com/Jjagg/MonoGameContentReferenceTest

MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
dazinator: Latest commit I have removed the additional slashes. Haven't been able to test on xbuild yet though so can't pick up the issue with RelativeDir / Filename

dazinator:
I have tried testing this with xbuild. I am new to xbuild. I dounf this issue: https://github.com/Jjagg/MonoGameContentReferenceTest/issues/3

Jjagg:
Mono 5.0 will be out soon and xbuild is getting deprecated in favor of msbuild. See http://www.mono-project.com/docs/about-mono/releases/5.0.0/#xbuild
IMO this PR can get merged.

Jjagg:
Mono 5.0 Just released and it ships with MSBuild and deprecates xbuild. Source: http://www.mono-project.com/docs/about-mono/releases/5.0.0/#msbuild

dellis1972:
can someone test using msbuild on Mac/Linux? I'm not in a position too atm its a bit of a busy time at work ;)

cra0zy:
> Mono 5.0 will be out soon and xbuild is getting deprecated in favor of msbuild. See http://www.mono-project.com/docs/about-mono/releases/5.0.0/#xbuild
> IMO this PR can get merged.

Even tho Mono 5 is out, it's not out on all distros so we should either make sure this works with xbuild or wait for a few weeks for some distros to get the package.

Jjagg:
What works with xbuild in develop works in this branch too. It's just the things this fixes for msbuild that aren't working correctly in xbuild. That's why I proposed to merge this even though msbuild is not available on all distros yet.

dazinator:
@Jjagg so in other words, this doesnt make anything worse, it just makes msbuild better? :-)

Jjagg:
Yes

dazinator:
Any news on when this might be merged? Does it need any further action on my part?

KonajuGames:
It will need to be brought up to date and pass the build checks.

dazinator:
Updated and resolved conflict. Will now need to be retested on the various platforms again.

dazinator:
This is currently up to date. Would like to get this merged if possible because I wont have time to continue to keep this in line with dev should the targets in dev diverge again!

KonajuGames:
Merging.  Thanks @dazinator

vpenades:
This PR has modified heavily MonoGame.Content.Builder.targets with an unexpected consequence:

When compiling a project with missing content.mgcb (because it's not needed) MSBuild gets stuck!, as if it's doing some sort of infinite loop.

Now I have multiple MonoGame utility assemblies that leave visual studio blocked forever when I try to compile them!

Steps to reproduce:

Create a new Monogame Windows Project
Delete content.mgcb
Compile

mrhelmut:
Ah! That is something that is going to affect us too.
@vpenades would you mind opening an issue? I'm going to take over this and fix it.

dazinator:
Apologies :-(, I didn't test this very rigorously after resolving a conflict from dev. I certainly didn't test it against a project with no mgcb's so that could have been a problem that was present in this PR all along, or it could have been introduced as a result of my bad attempt at conflict resolution.

dazinator:
One thing that would definately be nice would be some integration tests around msbuild targets :-)

KonajuGames:
There's always room for more tests.  Thanks @mrhelmut for taking a look at this.  And no worries @dazinator.  There are many situations in how things are used, it's hard to think of and cover all of them.

KonajuGames:
I submitted a potential fix for this (#5886).  Try that branch or try the [installer](http://teamcity.monogame.net/viewLog.html?buildId=51930&tab=artifacts&buildTypeId=MonoGame_PackagingWindows) generated from it.  Let me know if it fixes the issue.

cra0zy:
xbuild building is broken.... told you all :|

```
XBuild Engine Version 14.0
Mono, Version 5.0.0.0
Copyright (C) 2005-2013 Various Mono authors

Build started 8/22/2017 6:26:55 PM.
__________________________________________________
Project "/mnt/Home/Projects/GameOfSpikes/GameOfSpikes/GameOfSpikes.csproj" (default target(s)):
: error : Error building target Prepare: Object reference not set to an instance of an object
Done building project "/mnt/Home/Projects/GameOfSpikes/GameOfSpikes/GameOfSpikes.csproj".-- FAILED

Build FAILED.
Errors:

/mnt/Home/Projects/GameOfSpikes/GameOfSpikes/GameOfSpikes.csproj (default targets) ->

	: error : Error building target Prepare: Object reference not set to an instance of an object

	 0 Warning(s)
	 1 Error(s)

Time Elapsed 00:00:00.1830570
```

dazinator:
hmm.. @cra0zy when you say you told us all, are you referring to your above comment:

> Even tho Mono 5 is out, it's not out on all distros so we should either make sure this works with xbuild or wait for a few weeks for some distros to get the package.

Just wondered, because that was from May 12th. The PR was merged on 18th August. Does that not count waiting for a few weeks for some distros to get the package? 

Given that xbuild is deprecated in mono5, are we still aiming to support xbuild? Or are we forcing a breaking change on xbuild consumers to that they must switch to msbuild?

cra0zy:
> Just wondered, because that was from May 12th. The PR was merged on 18th August. Does that not count as waiting for a few weeks for some distros to get the package? :-p
> 
> Given that xbuild is deprecated in mono5, are we still aiming to support xbuild? Or are we happy forcing a breaking change on xbuild consumers with the advice that they must now switch to msbuild?

You don't get how Linux currently works, there are 2 types of distros:
 1) Rolling release distros (Arch Linux, Antergos, Manjaro...), usually get the newest version of the package few days after it releases
 2) Distros that have releases (Ubuntu, Debian, Fedora...), they get releases every 6 months (and some updates to the existing release packages, but only to the core packages)

So in short, all of the 2nd type of distros won't be getting msbuild in the core repos any time soon. Supporting xbuild for the time being would be the best option.

cra0zy:
Other side note, msbuild is not in any official Linux repos at the moment, it's a long story...

dazinator:
@cra0zy - understood, thanks for clarifying the situation.
Unless we can find a way to keep support for xbuild then, we should revert this PR.

dellis1972:
@cra0zy are you sure you are not hitting this 

```	System.NullReferenceException: Object reference not set to an instance of an object
		  at (wrapper unknown) SharpFont.Internal.GlyphSlotRec:PtrToStructure (intptr,object)
		  at (wrapper managed-to-native) System.Runtime.InteropServices.Marshal:PtrToStructure (intptr,System.Type)
		  at SharpFont.PInvokeHelper.PtrToStructure[T] (System.IntPtr reference) [0x00000] in <1ab1c07fc69249aba62188cb35238785>:0 
		  at SharpFont.GlyphSlot.set_Reference (System.IntPtr value) [0x00007] in <1ab1c07fc69249aba62188cb35238785>:0 
		  at SharpFont.GlyphSlot..ctor (System.IntPtr reference, SharpFont.Face parentFace, SharpFont.Library parentLibrary) [0x00006] in <1ab1c07fc69249aba62188cb35238785>:0 
		  at SharpFont.Face.get_Glyph () [0x0001b] in <1ab1c07fc69249aba62188cb35238785>:0 
		  at Microsoft.Xna.Framework.Content.Pipeline.Graphics.SharpFontImporter.ImportGlyph (System.Char character, SharpFont.Face face) [0x00011] in <72a8bc4ede8a4fc4bd6b25e4221ff20d>:0 
		  at Microsoft.Xna.Framework.Content.Pipeline.Graphics.SharpFontImporter.Import (Microsoft.Xna.Framework.Content.Pipeline.Graphics.FontDescription options, System.String fontName) [0x0002f] in <72a8bc4ede8a4fc4bd6b25e4221ff20d>:0 
		  at Microsoft.Xna.Framework.Content.Pipeline.Processors.FontDescriptionProcessor.ImportFont (Microsoft.Xna.Framework.Content.Pipeline.Graphics.FontDescription options, System.Single& lineSpacing, System.Int32& yOffsetMin, Microsoft.Xna.Framework.Content.Pipeline.ContentProcessorContext context, System.String fontName) [0x00052] in <72a8bc4ede8a4fc4bd6b25e4221ff20d>:0 
```

In which case that looks like a bug in the Font processor

dellis1972:
run the test from the command line with a `/v:d` to get the full diagnostic output. I get the above on the latest development branch

cra0zy:
@dellis1972 The project builds just fine with msbuild, so it's definitely an xbuild related bug. As for the log: https://gist.github.com/cra0zy/f6bce9e5fbd5a46422132cf5f949e324

dazinator:
@cra0zy that looks like the same issue I found when I tested on xbuild on windows: full diagnostic output here: https://github.com/Jjagg/MonoGameContentReferenceTest/issues/3

I am going to try and move the build logic out of the targets file into a custom msbuild task. This may take me some time, so if someone else wants to patch this some other way in the iterim (or revert) then I have no objections.

dellis1972:
@cra0zy yup, looks like two separate issues. I get the crash I listed under both xbuild and msbuild on Mac :/

dazinator:
@cra0zy I have created a PR to add support back for xbuild without losing this functionality. I would appreciate if anyone has time to have a look at: #5893

I moved the complicated logic in the targets file, into a new c# msbuild task. This ofcourse meant a new project for the new msbuild task, and I am not that familiar with the msbuild code base so I expect the PR will need to be brought in line with your build procedures / naming conventions etc etc. Any feedback welcome.

cra0zy:
Let me fix this, all it needs is one conditional and few extra lines of code... (tomorrow, it's 03:37 in my time zone, need to sleep).

dazinator:
@crazy Ok, happy for you to fix in the targets! I'll leave my PR branch there just in case we ever want to move to an msbuild task in the future. The msbuild task does have one additional "feature" which the targets currently doesn't which is that it builds each MGCB / content reference in parallel, with the idea that if you do have multiple MGCB's it should result in a quicker build time.

dellis1972:
This is not only broken in `xbuild` (the null reference exception) but its also incorrect under `msbuild` on Mac. It places the output files in a the output dir using a full path. so you end up with a folder 

`Users/<user>/<path to project>`

given people are trying to use develop I think we should just revert this change until it can be fixed. 
@tomspilman @KonajuGames thoughts?

dellis1972:
I found another bug :/ If your content.mgcb is in a shared project the output never makes it into the final app output folder.

dellis1972:
The good news is that using `CreateItem` for the `@(ContentReferences)` item group fixes the build crash.

```   
<CreateItem Include="@(MonoGameContentReference)"
			AdditionalMetadata="ContentOutputDir=$(ProjectDir)%(RelativeDir)bin\$(MonoGamePlatform)\%(Filename);ContentIntermediateOutputDir=$(ProjectDir)%(RelativeDir)obj\$(MonoGamePlatform)\%(Filename)">
		<Output TaskParameter="Include" ItemName="ContentReferences" />
</CreateItem>
```

KonajuGames:
> I found another bug :/ If your content.mgcb is in a shared project the output never makes it into the final app output folder.

Already reported in #5895 with more detail.

> given people are trying to use develop I think we should just revert this change until it can be fixed.

If it's not in develop, it doesn't get the testing in all use cases.  If they stay in PRs they don't get the widespread testing.  Yes, there are some issues with some use cases.  We need to fix those issues rather than revert the entire change.

PR #4998, opened 2016-07-09
Implement Direct Font Importer and Processor

cra0zy:
Stuff done:
- Implements direct importer / processor for font files (.ttf, .otf)
- Implements CharacterRegionsDescription object and it's TypeConverter so we can serialize/deserialize it
- Implements Char and CharacterRegionsDescription cell editors for Pipeline Tool
- A minor fix with sending object type in propertygrid

Notes:
- I left style in, even tho it doesn't do anything without #4561

Screenshots:
screenshots

### DIFF
--- a/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
+++ b/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
@@ -63,6 +63,7 @@
     <!-- MonoGame.Framework.Content.Pipeline.Builder -->
     <Compile Include="Builder\FileHelper.cs" />
     <Compile Include="Builder\Convertors\StringToColorConverter.cs" />
+    <Compile Include="Builder\Convertors\StringToCharacteRegionsConvertor.cs" />
     <Compile Include="Builder\PathHelper.cs" />
     <Compile Include="Builder\PipelineBuildEvent.cs" />
     <Compile Include="Builder\PipelineBuildLogger.cs" />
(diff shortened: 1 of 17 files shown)


tomspilman:
@cra0zy - If you split out the property grid fixes into a separate PR i'll merge that no problem.

Adding another font processor will take a bigger discussion.

cra0zy:
> @cra0zy - If you split out the property grid fixes into a separate PR i'll merge that no problem.
> 
> Adding another font processor will take a bigger discussion.

I will be busy tomorrow, and possibly the day after, so how about we have a discussion for now, if it's not resolved within that time frame I will make a separate PR for the fix.

tomspilman:
This is what I have figured out reading thru the changes.

This PR adds a new importer `FontImporter` and a new processor `FontProcessor`.

`FontImporter` works with native font files like `.ttf` and produces an empty `FontDescription` with just the filename set.  This is different from the existing XNA `FontDescriptionImporter` which reads a `.spritefont` and generates a complete `FontDescription`.

You then have `FontProcessor` which allows you to define the settings normally defined in the `.spritefont` file as a processor setting.  It then calls `FontDescriptionProcessor` to do the normal font processing.

If you use `FontImporter` with `FontProcessor` things should work.  If you use `FontImporter` with `FontDescriptionProcessor` you'll likely get some crash as the other fields for `FontDescription` are not defined.  If you use `FontDescriptionImporter` with `FontProcessor` things should work allowing you to change what was defined within the `.spritefont`.

So my first issue is this makes things confusing to users.  Which importer should they use with which processor?  The only way to know that is to take a guess, read some docs, or just trial and error.  

Worse it will be luck which processor is selected for `.spritefont` as it currently picks the first one that processes a `FontDescription` which could arbitrarily be `FontDescriptionProcessor` or `FontProcessor` depending on reflection order.

IMO the "right" thing to do is to fix MGCB and the Pipeline tool to allow for properties on importers.  This seems like a logical extension to the XNA content pipeline.  This way we only need one additional importer `FontImporter` which works with native outline font files then allows the user to set the additional `FontDescription` settings as importer properties.

@dellis1972 @KonajuGames ?

KonajuGames:
It makes sense to extend importers to allow properties.  I don't know why this wasn't included in XNA originally.  Then `FontImporter` can be given the properties in order to produce a complete `FontDescription` that can be consumed directly by the standard `FontDescriptionProcessor`.

cra0zy:
Reflection won't get confused, but I could see how having multiple font processors can be confusing to the end user.

cra0zy:
#5065

PR #4996, opened 2016-07-08
Refactored texture processing for platform extensibility

tomspilman:
This is a follow up to https://github.com/mono/MonoGame/pull/4992.

Here I've added a new `TextureProfile` and `DefaultTextureProfile` classes which do the work previously scattered across `GraphicsUtil`, `TextureProcessor`, and the font processors.

Like with the work on shader processing...  one just need to implement a new instance of `TextureProfile` to add support for compressing textures to a new format for a new platform.

Our existing texture processing unit tests should prove this change consistent with the previous system.

PR #5001, opened 2016-07-10
Refactored audio processing for platform extensibility

tomspilman:
This is a follow up to #4992 and #4996.

Here I've added a new `AudioProfile` and `DefaultAudioProfile` classes which do the work previously found in `AudioContent`.

Like with the work on shader processing and textures... one just need to implement a new instance of `AudioProfile` to add support for converting audio to a new format for a new platform.

As part of this I added a bunch of unit tests for `WavImporter` and `AudioContent` to ensure we've compatible with XNA behavior.

Note that conversion to MSADPCM doesn't result in the exact final sample rates as XNA.  For this reason I worked around those checks in the test.  We need someone else to go investigate and figure this out eventually...  but it isn't critical IMO.

Fixes https://github.com/mono/MonoGame/issues/4283.


tomspilman:
@dellis1972 @KonajuGames - Make sure we squash merge this one.

tomspilman:
Added some tests for `Mp3Importer` and `WmaImporter` which exposed an incompatibility with XNA.

tomspilman:
@SeriousMartin  @KonajuGames  - You guys were the last ones to make changes to the audio processing in the pipeline.  It would be good for you guys to look over this PR.

KonajuGames:
I like these changes.  Can't see anything that should be fixed.  Happy to squash and merge.

tomspilman:
@KonajuGames - I just merged in the latest develop changes.  After that rebuilds this should be safe to merge if we trust my unit tests.

KonajuGames:
Merging.

SeriousMartin:
Code looks really good and looking at it I couldn't find any problems.
I will have a try with our game in the next days.

tomspilman:
Thanks for reviewing this @SeriousMartin !

Might want to take a quick glance at https://github.com/mono/MonoGame/pull/5006 too.

cra0zy:
I think this broke audio compilation, I might be wrong, but nothing else seemed to have touched it. The error while compiling .ogg file with Song processor is:

```
C:/Music.ogg
C:/Music.ogg: error: Importer 'OggImporter' had unexpected failure!
Microsoft.Xna.Framework.Content.Pipeline.InvalidContentException: Failed to open file Bonus1. Ensure the file is a valid audio file and is not DRM protected. ---> System.ArgumentException: Incorrect file type!
Parameter name: audioFileType
   at Microsoft.Xna.Framework.Content.Pipeline.Audio.AudioContent..ctor(String audioFileName, AudioFileType audioFileType)
   --- End of inner exception stack trace ---
   at Microsoft.Xna.Framework.Content.Pipeline.Audio.AudioContent..ctor(String audioFileName, AudioFileType audioFileType)
   at Microsoft.Xna.Framework.Content.Pipeline.OggImporter.Import(String filename, ContentImporterContext context)
   at Microsoft.Xna.Framework.Content.Pipeline.ContentImporter`1.Microsoft.Xna.Framework.Content.Pipeline.IContentImporter.Import(String filename, ContentImporterContext context)
   at MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.ProcessContent(PipelineBuildEvent pipelineEvent)
```

tomspilman:
> error while compiling .ogg file with Song processor is:

I didn't add any sort of unit test for OGG as an input format.  The bug comes from this code:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs#L125

Which is accurate to XNA behavior, but not designed to support OGG as an input format.

I guess it would be ok to extend `AudioFileType` to have OGG...  it already has MP3 so it sort of makes sense.

If we add this unit tests verifying the import works as expected is required.

cra0zy:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/OggImporter.cs

PR #5004, opened 2016-07-12
[SDL] Fix GamePad Trigger Values

### LINKED ISSUE #4995: IsButtonDown(Buttons.RightTrigger) always returning false on Cross Platform project
As the title states, when I attempt to use the isButtonDown method of a GamePadState to check if Buttons.RightTrigger or Buttons.LeftTrigger are down it always returns false. The trigger values report accurate amounts however.

This occurs in the Cross Platform Project specifically, and does not appear to happen in any of the windows projects.

cra0zy:
Stuff done:
 - Fix starting value for GamePad triggers to not be 0.5
 - Fix trigger button detection (fixes #4995)


tomspilman:
Thanks @cra0zy ... merging!

PR #5006, opened 2016-07-13
Improved SoundEffect internal platform extendability.

tomspilman:
This PR improves the internals of `SoundEffect` construction/initialization to be more flexible for per-platform extendibility.

As part of these changes I added a bunch more unit tests for creating `SoundEffect` via constructors and and `SoundEffect.FromStream`.


KonajuGames:
I know very little about XACT, but the other changes look good.

tomspilman:
The XACT change here just moves the responsibility for loading the different sound formats to a platform partial method named `SoundEffect.PlatformInitializeXAct`.  This way we can easily support platform specific formats in XACT wave banks.

KonajuGames:
Merging. Thanks @tomspilman

cra0zy:
This PR broke sound effects on UWP, they are now completely distorted.

tomspilman:
> This PR broke sound effects on UWP, they are now completely distorted.

Can you enter a bug report with details?

I just tested against the latest MonoGame develop branch on a Windows UWP and a Windows desktop project and i didn't hear any sound distortion here.

cra0zy:
Hmm testing some more and I've found that the bug occurs when loading 2 or more sound effects. And the bug is actually merging them together into 1... real fun to listen to what is created (I was loading around ~30 at the same time so I didn't notice that they got merged).

tomspilman:
>  And the bug is actually merging them together into 1

What does this mean?  We have no concept of merging any sort of sounds.

cra0zy:
The code that should reproduce it is:
```c#
var s1 = Content.Load<SoundEffect>("s1");
var s2 = Content.Load<SoundEffect>("s2");
s1.Play();
```
The result ends up being some kind of mixture of the two sounds. I'll test it out tomorrow on an empty project just to make sure.

tomspilman:
> The result ends up being some kind of mixture of the two sounds.

I think I see the bug now.

We are using a scratch buffer during SoundEffect loading:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Content/ContentReaders/SoundEffectReader.cs#L43

Which is perfectly fine and the AL paths copy that data into the AL buffer using `alBufferData`.

I thought that the XAudio path does that too, but it seems XAudio pins and keeps the memory via `DataStream`:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffect.XAudio.cs#L164

`DataStream.Create()` being defined as:

https://github.com/sharpdx/SharpDX/blob/master/Source/SharpDX/DataStream.cs#L96

Which seems to internally pin the buffer and from reading the docs in XAudio seems to be the memory used for playback...  no copy is made.

https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.xaudio2.xaudio2_buffer(v=vs.85).aspx

This is all some guessing just looking at the code.  It would be good to verify that this is indeed the case, ideally devise some unit test that exposes the bug, then ensure we make a copy of the buffer before passing it to `DataStream` in the XAudio implementation.

cra0zy:
I have never used Xaudio so I probably shouldn't try fixing this...

PR #5008, opened 2016-07-15
Added Reverb and Filter support for XACT under OpenAL

dellis1972:


dellis1972:
I think I need to fix the MacOS Build still.

MonoGame.Framework/Audio/SoundEffectInstance.OpenAL.cs
```diff
@@ -260,14 +308,44 @@ private void PlatformSetVolume(float value)
...
+                case FilterMode.LowPass:
+                filterType = EfxFilterType.Lowpass;
+                break;
+                case FilterMode.HighPass:
+                filterType = EfxFilterType.Highpass;
+                break;
+            }
+            this.filterQ = filterQ;
+            this.frequency = frequency;
+#endif
```
tomspilman: I think you have a misunderstanding of the API here.  `PlatformSetFilter` expects immediate application of the filter settings if the sound instance is playing.

You should be able to play a long running or looping sound effect and modify both the filter and reverb settings interactively without restarting playback.

MonoGame.Framework/Audio/SoundEffectInstance.OpenAL.cs
```diff
@@ -260,14 +308,44 @@ private void PlatformSetVolume(float value)
 
         internal void PlatformSetReverbMix(float mix)
         {
+#if !MONOMAC
+            if (!OpenALSoundController.Efx.IsInitialized)
+                return;
+            reverb = mix;
+#endif
```
tomspilman: Same as with `PlatformSetFilter`... this should immediately affect a sound that is currently playing.

MonoGame.Framework/Audio/SoundEffectInstance.OpenAL.cs
```diff
@@ -260,14 +308,44 @@ private void PlatformSetVolume(float value)
...
         }
 
         internal void PlatformClearFilter()
         {
+#if !MONOMAC
+            if (!OpenALSoundController.Efx.IsInitialized)
+                return;
+
+            applyFilter = false;
+#endif
```
tomspilman: Note that this is ok not being immediate...  just needs to clear the filter state before next playback.

MonoGame.Framework/Audio/SoundEffectInstance.OpenAL.cs
dellis1972: ok so if its NOT playing we need to apply it before we do, but if we are we need to apply it immediately?

tomspilman: Yes.

dellis1972:
@tomspilman Updated :) 

If we are playing already we apply the effect, otherwise we store the values for the next time we start playing. Hope thats right.

tomspilman:
That looks like it should work... thanks @dellis1972 !

PR #5011, opened 2016-07-15
Reset effectMaterial.Effect to null so we do not try to serialize it.

dellis1972:
There was a bug when serializing the effectMaterial.
After we had compiled the effect it was still trying
to serialize the actual Effect object as well not
just the CompiledEffect.
So we null it out after compilation to make sure
we dont serialize it.


dellis1972:
not required. Was not on the latest develop branch

PR #5013, opened 2016-07-15
Change the Default texture compression for Models to Compressed

dellis1972:
The default was DXTCompressed. This does not work on
iOS. To make it easier we should default to
Compressed so that the default for the target
platform is used.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/ModelProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/ModelProcessor.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
         private bool _premultiplyTextureAlpha = true;
         private bool _premultiplyVertexColors = true;
         private float _scale = 1.0f;
-        private TextureProcessorOutputFormat _textureFormat = TextureProcessorOutputFormat.DxtCompressed;
+        private TextureProcessorOutputFormat _textureFormat = TextureProcessorOutputFormat.Compressed;
 
         #endregion
 
@@ -83,7 +83,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
 
         public virtual bool SwapWindingOrder { get; set; }
 
-		[DefaultValue(typeof(TextureProcessorOutputFormat), "DxtCompressed")]
+		[DefaultValue(typeof(TextureProcessorOutputFormat), "Compressed")]
         public virtual TextureProcessorOutputFormat TextureFormat
         {
             get { return _textureFormat; }


tomspilman:
Well IMO there are a bunch of issues here that we probably need to discuss and work out.

Here is the `TextureProcessorOutputFormat` enum....

ttps://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/TextureProcessorOutputFormat.cs#L10

Some time ago it was decided to add all these platform specific formats to the list:

  - Etc1Compressed
  - PvrCompressed
  - AtcCompressed

These and the legacy XNA `DxtCompressed` field just cause problems and should be marked obsolete:

  - They confuse the user...  what is the right thing to select?
  - Error prone as you can pass formats that don't work on your target platform.
  - The format selected might not be optimal to the content being passed in.

The code currently has to figure out if the user selected a specific format or a general format.  If the format is specific be sure that format is allowed on the platform and the content meets requirements.  If the format isn't specific then pick the right format... also then make sure content meets requirements.  Now once you get to specific formats... you can then choose the appropriate subformat.  It is sort of a complex mess.

Instead I think we should stick to the following formats:

   - Color
   - Compressed (what you get depends on platform and texture)
   - Color16Bit (could be renamed to HalfColor)
   - NoChange

For `DxtCompressed` and `Compressed` should have the same value to allow importing existing XNA .contentprojs to do the right thing.

Then we just need to examine the input data and the platform and pick the best format for that case.

Anyway... this has all bothered me for a while.

Jjagg:
As someone who has been confused by the different compression formats, I agree with Tom! Sounds like a good change to me :)

KonajuGames:
The intent was that Android builds could have different data packages for different GPUs, but in reality this doesn't happen.  Compressed on Android should just use the current defaults of ETC1 for opaque and 16-bit BGRA for non-opaque.

I think we could obsolete the unnecessary enums and remap them to the smaller set of enums on load of a .mgcb file.

tomspilman:
> Compressed on Android should just use the current defaults of ETC1 for opaque and 16-bit BGRA for non-opaque.

ETC1 for opaque and 1bit alpha right?  16bit color for everything else?

> I think we could obsolete the unnecessary enums and remap them to the smaller set of enums on load of a .mgcb file.

Agreed.  If and when we make this change we want the old formats to never appear in the format combobox again...  only the formats we support.

KonajuGames:
> ETC1 for opaque and 1bit alpha right? 16bit color for everything else?

ETC1 doesn't support any alpha, as far as I can remember, so it would be
for opaque only.

KonajuGames:
@dellis1972 You also need to change line 27.  The attribute is just a hint to the UI as to what the default value is.  The actual default is set in code still.

dellis1972:
Weird None of the replies I've sent to this thread are appearing :/

I would argue against removing the ability to use ATITIC, DXT and PVRTC directly. One of the games I worked on we tried 16 bit RGBA for compression and we ran out of GPU memory on lower end devices, using ATITC/PVRTC etc worked since the compression ratio is better that 50% which is the best you can hope for with 16bit RGBA afaik. 

Perhaps we can do a better job of defaults/docs but still give advanced users the ability to use the compression they want. If a user wants to deploy 3 apk's to the play store on for each GPU (which they can.. and its recommended [1] ) then we should not put blocks in place to make the harder.

@KonajuGames whoops , overlooked that I will fix it.

[1] https://developer.android.com/google/play/publishing/multiple-apks.html

tomspilman:
>  If a user wants to deploy 3 apk's to the play store on for each GPU (which they can.. and its recommended [1] ) then we should not put blocks in place to make the harder.

Sure...  but the way to provide this isn't asking the user to manually solve this themselves.  Currently they would need to switch all textures to ATITIC... build, save results...  then switch everything to PVRTC.... repeat again for DXT, etc.

That is a poor way to provide this feature for a popular platform for new and even experience developers.  This is why I still recommend we remove all platform specific formats from the `TextureProcessorOutputFormat` enum.

dellis1972:
> That is a poor way to provide this feature for a popular platform for new and even experience developers. This is why I still recommend we remove all platform specific formats from the TextureProcessorOutputFormat enum.

@KonajuGames  @tomspilman I agree but given the way you have to do this with android you end up building 3 .apks anyway as each needs different values in the AndroidManifest.xml.
So we need to figure out a way to do this that makes sense. 

I opened an issue for this https://github.com/mono/MonoGame/issues/5028 to track the issue. 

In the meantime I have updated the defaults on this PR so it should be good to go now

tomspilman:
Thanks @dellis1972 !

PR #5014, opened 2016-07-16
SoundEffectInstance.XAudio changed to lower memory allocation.

Zolniu:
X3DAudio Emitter and Listener are cached to prevent memory allocation
for every Apply3D call.

This is a quick and dirty "correction" so it should be refactored in future as long as you approve this approach. I think we should aim for allocation-free solutions where possible.

Fixes https://github.com/mono/MonoGame/issues/5015.

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs
@@ -93,12 +93,19 @@ namespace Microsoft.Xna.Framework.Audio
             if (_dxEmitter == null)
                 _dxEmitter = new Emitter();
 
+#if WINDOWS_UAP
+            _dxEmitter.Position = new SharpDX.Mathematics.Interop.RawVector3 { X = pos.X, Y = pos.Y, Z = pos.Z };
+            _dxEmitter.Velocity =  new SharpDX.Mathematics.Interop.RawVector3 { X = vel.X, Y = vel.Y, Z = vel.Z };
+            _dxEmitter.OrientFront = new SharpDX.Mathematics.Interop.RawVector3 { X = forward.X, Y = forward.Y, Z = forward.Z };
+            _dxEmitter.OrientTop = new SharpDX.Mathematics.Interop.RawVector3 { X = up.X, Y = up.Y, Z = up.Z };
+            
+#else
             _dxEmitter.Position = new SharpDX.Vector3(pos.X, pos.Y, pos.Z);
             _dxEmitter.Velocity = new SharpDX.Vector3(vel.X, vel.Y, vel.Z);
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs
```diff
@@ -60,6 +60,86 @@ private void PlatformApply3D(AudioListener listener, AudioEmitter emitter)
...
+            vel.Z *= -1.0f;
+
+            if (_dxEmitter == null)
+                _dxEmitter = new Emitter();
+
+            _dxEmitter.Position = new SharpDX.Vector3(pos.X, pos.Y, pos.Z);
+            _dxEmitter.Velocity = new SharpDX.Vector3(vel.X, vel.Y, vel.Z);
+            _dxEmitter.OrientFront = new SharpDX.Vector3(forward.X, forward.Y, forward.Z);
+            _dxEmitter.OrientTop = new SharpDX.Vector3(up.X, up.Y, up.Z);
+            _dxEmitter.DopplerScaler = emitter.DopplerScale;
```
tomspilman: The build failure is that you are missing a `#if WINDOWS_UAP` here.  See the original code:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Windows8/SharpDXHelper.cs#L183

MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs
```diff
@@ -60,6 +60,86 @@ private void PlatformApply3D(AudioListener listener, AudioEmitter emitter)
...
+            pos.Z *= -1.0f;
+            vel.Z *= -1.0f;
+
+            if (_dxListener == null)
+                _dxListener = new Listener();
+
+            _dxListener.Position = new SharpDX.Vector3(pos.X, pos.Y, pos.Z);
+            _dxListener.Velocity = new SharpDX.Vector3(vel.X, vel.Y, vel.Z);
+            _dxListener.OrientFront = new SharpDX.Vector3(forward.X, forward.Y, forward.Z);
+            _dxListener.OrientTop = new SharpDX.Vector3(up.X, up.Y, up.Z);
```
tomspilman: Same here....  see the original code:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Windows8/SharpDXHelper.cs#L227

tomspilman:
We certainly missed this garbage generation...  good catch @Zolniu.

Generally this seems like a good improvement.  I've added some notes to what is currently causing the build to fail.

Zolniu:
Thanks @tomspilman,
I did that change in my local version of MonoGame code quite some time ago (before UAP was a thing). I didn't realize that the logic in SharpDXHelper changed since then :)

KonajuGames:
Merging. Thanks @Zolniu.

PR #5016, opened 2016-07-17
Update coding policy for MonoDevelop / Xamarin Studio 6

cra0zy:
The file is used to autoset code formating when opening with MonoDevelop/Xamarin Studio, there were some changes with version 6.


KonajuGames:
One of the other hidden changes in XS 6.  Merging. Thanks @cra0zy

PR #5020, opened 2016-07-21
Fixed initialization of GameComponent when created within another GameComponent

nkast:
Fixes https://github.com/mono/MonoGame/issues/2943

Test file 
[ComponentTest.zip](https://github.com/mono/MonoGame/files/376393/ComponentTest.zip)
based on 
http://pastebin.com/bDG4yZUG

output: 
    Constructor - Initialize
    Initialize (pre-base.Initialize) - Initialize
    Constructor- 2 - Initialize
    Initialize (pre-base.Initialize)- 2 - Initialize
    BeginRun- 2 - Initialize
    BeginRun - Initialize
    Update- 2 - Initialize
    Update - Initialize
    Draw- 2 - Initialize
    Draw - Initialize

Note: removing a component during .Initialize() will have unpredictable results.

### DIFF
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -92,6 +92,7 @@
 	<Compile Include="Framework\CurveKeyTest.cs" />
 	<Compile Include="Framework\CurveTest.cs" />
 	<Compile Include="Framework\EnumConformingTest.cs" />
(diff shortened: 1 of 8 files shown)


tomspilman:
This looks like the right fix...  thanks @nkast.  

Still could we convert the reported test into a unit test so we can keep this from regressing?

tomspilman:
This is great @nkast !  Merging!

PR #5022, opened 2016-07-22
DynamicSoundEffectInstance is added to all platforms by default

tomspilman:
`DynamicSoundEffectInstance` was only set to be added on DirectX and OpenAL platforms.... this adds it by default on all platforms.


KonajuGames:
Merging.

PR #5024, opened 2016-07-23
Implemented GraphicsAdapter for DirectX platforms

tomspilman:
This PR cleans up `GraphicsAdapter` into a partial class allowing for other platforms to cleanly extend it.

Added unit tests to verify the DirectX implementation matches XNA behavior.

Left old implementation of `GraphicsAdapter` in legacy file that is used on the other public platforms.  We should start replacing these with correct partial implementations.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.Legacy.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.Legacy.cs
@@ -1,0 +1,386 @@
+// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+#if MONOMAC
+#if PLATFORM_MACOS_LEGACY
+using MonoMac.AppKit;
+using MonoMac.Foundation;
+#else
+using AppKit;
+using Foundation;
+#endif
+#elif IOS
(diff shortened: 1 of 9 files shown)


MonoGame.Framework/Graphics/GraphicsAdapter.Legacy.cs
```diff
@@ -0,0 +1,386 @@
...
+#endif
+
+        public DisplayModeCollection SupportedDisplayModes
+        {
+            get
+            {
+                bool displayChanged = false;
+#if DESKTOPGL
+                var displayIndex = Sdl.Display.GetWindowDisplayIndex (SdlGameWindow.Instance.Handle);
+                displayChanged = displayIndex != _displayIndex;
```
tomspilman: @dellis1972 - I tested on XNA and it doesn't seem to work this way at all.

The problem is you need to be creating one `GraphicsAdapter` per-monitor per-graphics device like the new DirectX implementation does.   Then you just gather the modes for that one adapter which only covers one monitor.

Still how do you get the right adapter for the monitor the game window is in?  Well XNA didn't seem to have a built in way to do that.  It seems the idea was to somehow match the `GraphicsAdapter.MonitorHandle` with some GDI function.

What i suggest we instead add a `DesktopBounds` to each `GraphicsAdapter` allowing users to use their window bounds to pick an appropriate graphics adapter.

tomspilman:
This should be good to merge now.

KonajuGames:
Merging.

PR #5025, opened 2016-07-23
Add NVorbis Protobuild Module

cra0zy:
Just so people don't have to run the submodule commands (tho I would still recommend them because Protobuild doesn't clone git repo the same way git command does).


tomspilman:
This confused me the other day as well.  The build was failing because the source for NVorbis was missing.  Took me a second to realize I didn't init/update the submodules.

I wished that missing submodules were more apparent just when you do git updates somehow.

I really dislike Protobuild messing with git commands, but I guess we have to live with this.

cra0zy:
If you update the submodules manually, Protobuild won't mess with them.

tomspilman:
> If you update the submodules manually Protobuild won't mess with them.

Yeah I know...  IMO it would be better if Protobuild just gave an error when a submodule is missing and did nothing.  Or prompted like "Submodule XXX is missing.  Initialize submodule XXX? (Y/N)".

PR #5030, opened 2016-07-26
Fix Ogg Importer

cra0zy:


tomspilman:
@cra0zy - Two more things we need here.

We need to fix this block to detect Ogg and return the right `AudioFileType`:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Audio/DefaultAudioProfile.cs#L168

We need to add a unit test for Ogg import just like we have for MP3 import:

https://github.com/mono/MonoGame/blob/develop/Test/ContentPipeline/Mp3ImporterTests.cs

This test will only run under MonoGame, but it will ensure no one breaks Ogg import again.

cra0zy:
Sigh..

cra0zy:
@tomspilman Is build bot down?

tomspilman:
@cra0zy - Looks like it... the Windows agents are down.  Will try to remote in and fix it.  Thanks for alerting me!

tomspilman:
@cra0zy - This doesn't look like your issue... looks like a build server issue.  Working on it.

tomspilman:
@cra0zy - Error...

```
System.IO.FileNotFoundException : Could not locate audio file rock_loop_stereo.ogg.
```
Looks like you missed adding the `rock_loop_stereo.ogg` to `MonoGame.Tests.definition`.

cra0zy:
Now I don't think it's my fault :/

cra0zy:
@tomspilman Ping

tomspilman:
@cra0zy - There is something going wrong with TeamCity gathering the results from the unit tests which are all passing right now.  Working with the TeamCity people to try to work it out.

tomspilman:
@cra0zy - Can you do me a favor?

https://github.com/mono/MonoGame/blob/develop/Test/Framework/GameComponentTest.cs

I suspect that the `GameComponentTest` class and the `InitializeOrderTest` method need to be `static`.  Can you make this change in this PR?

cra0zy:
Done.

tomspilman:
So I updated to the latest version of TeamCity and the issue we were seeing in this PR seems to have gone away.  

However it reports only 306 tests when it should be running over 700 tests.  I don't think it has anything to do with this PR, so I will merge this anyway.

PR #5043, opened 2016-07-28
Adapter DPI and content .targets fix

tomspilman:
Fixes https://github.com/mono/MonoGame/issues/5040.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
@@ -61,6 +61,8 @@
 			'$(MonoGamePlatform)' != 'WindowsPhone8' And       
 			'$(MonoGamePlatform)' != 'RaspberryPi' And       
 			'$(MonoGamePlatform)' != 'PlayStation4' And       
+			'$(MonoGamePlatform)' != 'PSVita' And       
+			'$(MonoGamePlatform)' != 'XboxOne' And       
 			'$(MonoGamePlatform)' != 'WindowsGL'" />
 
     <Error
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
@@ -15,9 +15,13 @@ namespace Microsoft.Xna.Framework.Graphics
             // Get the current desktop DPI.
             Vector2 dpiScale;
             {
+#if WINDOWS_PHONE
+                dpiScale = Vector2.One;
+#else
                 var direct2d = new SharpDX.Direct2D1.Factory(SharpDX.Direct2D1.FactoryType.SingleThreaded);
                 dpiScale = new Vector2(direct2d.DesktopDpi.Width / 96.0f, direct2d.DesktopDpi.Height / 96.0f);
                 direct2d.Dispose();
+#endif
             }
 
             var factory = new SharpDX.DXGI.Factory1();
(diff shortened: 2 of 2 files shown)


Jjagg:
`DesktopDpi.Width` and `Height` are both 1, so this does not work :/

tomspilman:
> `direct2d.DesktopDpi.Width` and `Height` are both 1, so this does not work :/

Wait... really?  That is a crazy invalid setting.

tomspilman:
>  I mean they're both 96, which makes dpiScale = { 1, 1 }

 Ok...  but are you back to your old scaled display setting?  It requires a system reboot to switch if I remember correctly?

Jjagg:
You just have to log off, but I'll try a reboot just to make sure.

Jjagg:
After the reboot I get the same results, dpi as if scaling is at 100%, yet incorrect resolution.

KonajuGames:
Perhaps you'll need to go to the Win32 API to get the desktop scale or DPI, such as GetDpiForMonitor().

Jjagg:
> Perhaps you'll need to go to the Win32 API to get the desktop scale or DPI, such as GetDpiForMonitor().

GetDpiForMonitor() did not work either. I thought this would work, since on [this](https://msdn.microsoft.com/en-us/library/windows/desktop/dn280511.aspx) page it says literally:

> MDT_EFFECTIVE_DPI
The effective DPI. This value should be used when determining the correct scale factor for scaling UI elements. This incorporates the scale factor set by the user for this specific display.

Jjagg:
There was also something weird going on with getting the right value at other resolutions, so maybe something else is going on. I'll try all possible resolutions tomorrow.

tomspilman:
Thanks for investigating @Jjagg .

Jjagg:
I did the tests for all resolutions. I also checked dpi using both `DesktopDpi` from SharpDX call and `GetDpiForMonitor`, but both returned {96, 96} in all cases :/ Reported resolution by `Output.DesktopBounds` was always exactly `ActualResolution / DpiScaling`. I thought it reported the right values for some resolutions, but Windows set the dpi scaling back to 100% for those, so at least it's consistent and it's just the dpi scaling. I'm gonna try setting dpi scaling for the system (this was per monitor) and see if that does get reported by DesktopDpi or GetDpiForMonitor.

Another note is that with 800x600 Game._graphicsDeviceManager did not get initialized and caused an exception [here](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Game.cs#L711), but I didn't look into it now.

Jjagg:
With system dpi scaling reported dpi with both methods is { 96, 96 } too.

mrhelmut:
Note that querying DPI on Windows will always return 96 (event if the system or per-monitor DPI is different) unless DPI awareness is enabled (either via ```SetProcessDpiAwareness``` or via a manifest file with ```<dpiAware>true</dpiAware>```), which is a bit annoying because one may not want DPI awareness but still query it.
I'm not sure if I'm right, but that's how far my DPI experiences went.

Jjagg:
> Note that querying DPI on Windows will always return 96 (event if the system or per-monitor DPI is different) unless DPI awareness is enabled (either via SetProcessDpiAwareness or via a manifest file with <dpiAware>true</dpiAware>), which is a bit annoying because one may not want DPI awareness but still query it.

Woops, I didn't read the full documentation page last night... This did the trick! Thanks @mrhelmut! SharpDX gets the right values as well :)

tomspilman:
Seems like what we should do is make MonoGame always DPI aware.  Isn't this more inline with how things work in the SDL2 backend for Mac retina displays?

tomspilman:
Seems like what we should do is mark MonoGame always DPI aware.  Isn't this more inline with how things work in the SDL2 backend for Mac on retina displays?

@dellis1972 @cra0zy @mrhelmut @endimilojkoski

mrhelmut:
@tomspilman I believe so. On the Windows side of things, adding a manifest to the templates will do the trick; on Linux, there's nothing to do; and on MacOS, a documentation will be needed because DPI support depends on how the .app bundle is packaged and its info.plist. I'll try to setup a PR for this.

PR #5048, opened 2016-07-30
Dse stutter

### LINKED ISSUE #5023: Buffer boundary stutter using DynamicSoundEffectInstance
See [this forum post](http://community.monogame.net/t/just-trying-to-figure-out-how-to-play-a-simple-sin-wave/7938) for a repro.

This was a known issue in the `DynamicSoundEffectInstance` PR (#4715) which I was not able to pin down. It seems to occur both on XAudio and OpenAL, but I don't know if the issue is in the common code or platform implementations. Any ideas from someone more knowledgeable about audio?

Jjagg:
Fixes #5023. Or at least the OpenAL side for now. These were seperate issues for OpenAL and XAudio after all. I'll check out the XAudio issue tomorrow, but I'm hoping it's the same mistake conceptually that causes it. If that's the case it'll be a quick fix. Should I do a seperate PR or is this one fine for XAudio?

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -84,9 +84,6 @@
     <Uses Name="NVorbis">
       <Platforms>WindowsGL,Linux</Platforms>
     </Uses>
-    <Uses Name="_FrameworkDispatcherProvided">
-      <Platforms>Windows,WindowsGL,MacOS</Platforms>
-    </Uses>
     <Recommends Name="DirectXGraphics">
       <Platforms>Windows</Platforms>
     </Recommends>
@@ -147,9 +144,6 @@
     <Service Name="_XNADesignProvided">
       <AddDefine>XNADESIGNPROVIDED</AddDefine>
       <!-- Allows the Microsoft.Xna.Framework.Design framework to be present -->
-    </Service>
-    <Service Name="_FrameworkDispatcherProvided">
-      <!-- Allows the FrameworkDispatcher class to be present -->
     </Service>
 
   </Services>
@@ -291,10 +285,8 @@
     <Compile Include="CurveTangent.cs" />
     <Compile Include="DisplayOrientation.cs" />
     <Compile Include="DrawableGameComponent.cs" />
-    <Compile Include="FrameworkDispatcher.cs">
-      <Services>_FrameworkDispatcherProvided</Services>
-    </Compile>
-	<Compile Include="FrameworkResources.cs" />
+    <Compile Include="FrameworkDispatcher.cs" />
+    <Compile Include="FrameworkResources.cs" />
     <Compile Include="GameComponentCollection.cs" />
     <Compile Include="GameComponentCollectionEventArgs.cs" />
     <Compile Include="GameComponent.cs" />
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -292,7 +292,7 @@
     <Compile Include="DisplayOrientation.cs" />
     <Compile Include="DrawableGameComponent.cs" />
     <Compile Include="FrameworkDispatcher.cs">
-      <Platforms>_FrameworkDispatcherProvided</Platforms>
+      <Services>_FrameworkDispatcherProvided</Services>
     </Compile>
 	<Compile Include="FrameworkResources.cs" />
     <Compile Include="GameComponentCollection.cs" />
--- a/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs
+++ b/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs
@@ -158,16 +158,6 @@ namespace Microsoft.Xna.Framework.Audio
                     buffer.Dispose();
                 }
             }
-
-            // Raise the event
-            if (BufferNeeded != null)
-            {
-                var eventCount = TargetPendingBufferCount - PendingBufferCount;
-                for (var i = 0; i < eventCount; i++)
-                {
-                    BufferNeeded(this, EventArgs.Empty);
-                }
-            }
         }
     }
 }
--- a/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs
+++ b/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs
@@ -105,7 +105,10 @@ namespace Microsoft.Xna.Framework.Audio
             _queuedBuffers.Enqueue(oalBuffer);
 
             // If the source has run out of buffers, restart it
-            if (_state == SoundState.Playing)
+            var sourceState = AL.GetSourceState(SourceId);
+            ALHelper.CheckError();
+
+            if (_state == SoundState.Playing && sourceState == ALSourceState.Stopped)
             {
                 AL.SourcePlay(SourceId);
                 ALHelper.CheckError("Failed to resume source playback.");
@@ -155,10 +158,6 @@ namespace Microsoft.Xna.Framework.Audio
                     buffer.Dispose();
                 }
             }
-
-            // Raise the event for each removed buffer, if needed
-            for (int i = 0; i < numBuffers; i++)
-                CheckBufferCount();
         }
     }
 }
--- a/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs
+++ b/MonoGame.Framework/Audio/DynamicSoundEffectInstance.OpenAL.cs
@@ -158,6 +158,16 @@ namespace Microsoft.Xna.Framework.Audio
                     buffer.Dispose();
                 }
             }
+
+            // Raise the event
+            if (BufferNeeded != null)
+            {
+                var eventCount = TargetPendingBufferCount - PendingBufferCount;
+                for (var i = 0; i < eventCount; i++)
+                {
+                    BufferNeeded(this, EventArgs.Empty);
+                }
+            }
         }
     }
 }
--- a/MonoGame.Framework/Audio/DynamicSoundEffectInstance.XAudio.cs
+++ b/MonoGame.Framework/Audio/DynamicSoundEffectInstance.XAudio.cs
@@ -29,6 +29,14 @@ namespace Microsoft.Xna.Framework.Audio
 
         private void PlatformPlay()
         {
+            // fill up buffers before playing
+            if (_state == SoundState.Stopped && BufferNeeded != null)
+            {
+                var eventCount = TargetPendingBufferCount - PendingBufferCount;
+                for (var i = 0; i < eventCount; i++)
+                    BufferNeeded(this, EventArgs.Empty);
+            }
+
             _voice.Start();
         }
 
@@ -93,7 +101,8 @@ namespace Microsoft.Xna.Framework.Audio
                 buffer.Stream.Dispose();
             }
 
-            CheckBufferCount();
+            if (BufferNeeded != null && PendingBufferCount < TargetPendingBufferCount)
+                BufferNeeded(this, EventArgs.Empty);
         }
     }
 }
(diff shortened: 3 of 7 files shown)


Jjagg:
Woops, branched off the wrong branch -_-

Jjagg:
Hey, looks like I fixed it :D

tomspilman:
Thanks @Jjagg

polsys:
A great catch, thanks @Jjagg! Reading the OpenAL docs now, this is obvious 😆

Regarding `CheckBufferCount`, you probably should not touch it in this PR. It is required for XNA compatibility. The
```cs
var eventCount = TargetPendingBufferCount - PendingBufferCount;
```
version raises the events *every frame*, whereas XNA fires the event exactly once per removed buffer.

Jjagg:
Not every frame right? Only when less than three buffers are queued. Though if that's not how it works in XNA I'll change it. The way it was, this didn't work though because not enough buffers were queued, but I'll see what exactly went wrong and how I can make it work

Jjagg:
@polsys In the [XNA docs](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.audio.dynamicsoundeffectinstance_members.aspx) it says the BufferNeeded event 
> occurs when the number of audio capture buffers awaiting playback is less than or equal to two.

Jjagg:
Hmm, the stuttering is gone, but now I get some clipping or distortion at the start of playing the sound (after about a second)... I tried a minimal test that uses SharpDX directly, but I still get this, maybe I'm doing it wrong... I set up a gist if anyone wants to check (https://gist.github.com/Jjagg/e37303a18d743b3bb64ea9c6af049ca0), just start a new WinForms app, replace the Form1.cs file and reference SharpDX and SharpDX.Audio2 (these dll's should be in you're MonoGame installation folder under v3.0\Tools\)

Jjagg:
Looks like (or rather sounds like :p) submitting a buffer outside the BufferNeeded event when the soundeffect is playing can cause clipping and distortion... Not sure what exactly is going wrong, but everything works when I initialize with a couple buffers and then submit a new one when the BufferNeeded event is invoked. This means that the polling method and submitting in an Update calls does not work :/ and the only way to correctly use the class as it is now is to subscribe to the BufferNeeded event before playing the sound... not sure if that suffices, but I think this is a SharpDX thing and it probably happens because of threading issues. I don't really want to go digging in that right now. 

The event is probably not invoked at the same times as in XNA right now. From the XNA docs (I don't have it installed) I think the following table sums up the inconsistensies on when the BufferNeeded event is invoked:

|                          | XNA |              XAudio              |         OpenAL        |
|:------------------------:|:---:|:--------------------------------:|:---------------------:|
| During playback (Update) | Yes | **No** (causes stutter/glitches) |          Yes          |
|    On BufferEnd event    | Yes |                Yes               | **No** (only polling) |
|          On Play         | Yes |                Yes               |  **No** (can be done) |

**tl;dr** this is not very consistent because I can't seem to submit buffers to a playing sound outside the BufferNeeded event without causing clipping/glitches meaning you can only use the event method to use this class and not the polling method

tomspilman:
@Jjagg - We will need to fix the DynamicSoundEffectInstanceTest unit tests for which are currently failing.  I can verify if they still work under XNA here.

@polsys - Thoughts on these latest changes?

Jjagg:
Neat, didn't know there were unit tests for this :) The two remaining failing tests for Windows require that BufferNeeded is invoked on the main thread and in Update...

tomspilman:
@Jjagg - Only two tests failing now.  Just to be sure I re-ran the current tests on XNA and they all pass.  So the problem is likely in the MonoGame implementation and not the unit test.

Also I think this can be fixed to just be a call to `FrameworkDispatcher.Update()` now for both MonoGame and XNA:

https://github.com/mono/MonoGame/blob/develop/Test/Framework/Audio/DynamicSoundEffectInstanceTest.cs#L393

Might also be able to remove this one and do it under MonoGame too:

https://github.com/mono/MonoGame/blob/develop/Test/Framework/Audio/DynamicSoundEffectInstanceTest.cs#L18

Build/Projects/MonoGame.Framework.definition
```diff
@@ -292,7 +292,7 @@
     <Compile Include="DisplayOrientation.cs" />
     <Compile Include="DrawableGameComponent.cs" />
     <Compile Include="FrameworkDispatcher.cs">
-      <Platforms>_FrameworkDispatcherProvided</Platforms>
+      <Services>_FrameworkDispatcherProvided</Services>
```
tomspilman: This is causing the Linux build to fail.

We should probably remove this and just make `FrameworkDispatcher.cs` supported on all platforms.

I don't see any reason now that it should not be.

tomspilman: This is starting to maybe get deeper than your fix here @Jjagg .

But this looks like a big problem for the unit tests:

  https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/FrameworkDispatcher.cs#L12

We should move the sound updates from `Game` into there:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Game.cs#L653

Then call `FrameworkDispatcher.Update()` from `Game` instead.

Jjagg: Yeah, I know that's what's supposed to happen, but that wil reintroduce the stutter :/ If we really wanna fix this we'll have to dig into SharpDX code I think. Let me just try this locally to make sure

tomspilman: > Yeah, I know that's what's supposed to happen, but that wil reintroduce the stutter 

I would still hook up `FrameworkDispatcher.Update()` at least so that the AL platforms work correctly.

It is ok for now if the XAudio path just doesn't do anything under it.

tomspilman:
@Jjagg - Something to consider.  

The `byte[]` passed into `SubmitBuffer` currently cannot be modified while that buffer is playing.  If you look closely that memory is pinned by `DataStream.Create` and the low level XAudio system plays directly from that memory.

I don't know if this is a good thing or a bad thing.

I don't know if XNA has this same behavior or not.  I don't know how to properly message this important detail to users.  If we remove the feature then it means each buffer submitted would generate garbage unless we internally keep a pool of buffers to copy into.

Thoughts?

Jjagg:
I was handling the FrameworkDispatcher.Update call and noticed WP did something like that already https://github.com/mono/MonoGame/blob/8a8f21c65c6e40688ed75b64c65bbf5a5790e40f/MonoGame.Framework/WindowsPhone/WPGamePlatform.cs#L180. Is that something seperate that I can just ignore?

>The byte[] passed into SubmitBuffer currently cannot be modified while that buffer is playing. If you look closely that memory is pinned by DataStream.Create and the low level XAudio system plays directly from that memory.

Probably better to copy it, that seems like the most logical thing. I'd guess XNA does the same thing. If we pass `pin: false` to the DataStream.Create call, doesn't that take care of it?

tomspilman:
> noticed WP did something like that already 

Yeah you can ignore that.... that is there to support some media features specific to WP.

We still need to call our own dispatcher on all platforms.


> If we pass  pin: false  to the DataStream.Create call, doesn't that take care of it?

I was just looking at that and I think I found a bug in `DataStream.Create`:

https://github.com/sharpdx/SharpDX/issues/785

So I wouldn't depend on that unless the index offset is zero.

tomspilman:
@Jjagg - FYI...  I applied the buffer copy fix to non-dynamic sound effects to fix a reported bug.  https://github.com/mono/MonoGame/pull/5053

polsys:
The `FrameworkDispatcher` changes are implemented in #4675, so these changes look out of scope here. Why is that needed, I see no functional difference in this scenario?

The two failing tests are discussed in https://github.com/mono/MonoGame/pull/4715#issuecomment-207574731. 

The OpenAL fix is really clear, but on the XAudio side this mostly seems like moving event code around. The buffer copy issue is interesting: I just gave a quick go at copying the buffer before creating the `DataStream` and it *appears* to fix the glitching (at least in my test...)

polsys:
Lots of code (like [the MSDN sample](https://msdn.microsoft.com/en-us/library/ff827591.aspx)) reuse a single `byte[]` buffer, so we must support it. Since the primary reason to do so is avoiding garbage, we shouldn't generate it either. So the buffer pool sounds like a good idea.

We just must make sure that it does not end up as a memory leak - see [A cache with a bad policy is another name for a memory leak](https://blogs.msdn.microsoft.com/oldnewthing/20060502-07/?p=31333). For example, if the user submits initially 10 buffers and then keeps the buffer count as constant 3, we must not keep 7 dangling buffers.

tomspilman:
> The  FrameworkDispatcher  changes are implemented in #4675, so these changes look out of scope here. 

Sorry... @polsys and @Jjagg .  I totally forgot about that other PR.

Should we roll back the commits in this PR or should we merge the other PR and let @Jjagg fix the conflict?

Jjagg:
> should we merge the other PR and let @Jjagg fix the conflict?

I'm fine with this.

> The buffer copy issue is interesting: I just gave a quick go at copying the buffer before creating the DataStream and it appears to fix the glitching (at least in my test...)

 When you use the BufferEnd event it doesn't matter if you overwrite it (for the sine wave I'm testing with at least) so it makes sense that that was the only thing that worked. I'll try this. 

How would we handle a buffer pool? In theorie the user could submit a whole bunch of buffers, or do the XNA docs say something about this?

tomspilman:
> How would we handle a buffer pool?

I think a typical pooling technique is all we need.

We keep a free list of all `byte[]` not currently in use... we either pull one from that list or create a new `byte[]` in `PlatformSubmitBuffer`.  We separately keep a stack of active `byte[]` which we fill when we call `PlatformSubmitBuffer`.  When `BufferEnd` is called we know the first/last item in the active stack is the one that just finished, is now unused, and can be safely returned to the free list.

If we want to be extra sure we call `SubmitSourceBuffer` we pass in some identifier of the `byte[]` as the second context parameter....  however I think the FIFO nature of the buffer processing might be all we need really.

It might be worth making the free pool global with locks to allow for the pool to be shared among multiple dynamic sound instances.

KonajuGames:
The user can submit as many buffers as they want.  They will be buffered until required.  Of course this means that the latency will be huge, but we can't protect the user from themselves.  If they want to buffer a huge amount, we will buffer that amount and deal with it the best we can.

KonajuGames:
The user can submit buffers of varying sizes.  There is nothing to stop them from doing that.  The only restriction is that the buffer length for `SubmitBuffer(byte[])` or the `count` for `SubmitBuffer(byte[], int, int)` must be multiples of the block size (one complete stereo sample or one mono sample depending on which mode was chosen). 

This makes the byte[] pool a bit more difficult, but not impossible.

Jjagg:
Copying the buffer fixes the XAudio problems :) 
Once #4675 is merged, there's really just 2 things that need to be done here

1. Fix OpenAL path by only calling Play if playback stopped when submitting a buffer
2. Fix XAudio path by making sure the buffer is copied in SubmitBuffer

@polsys's code already handled the events correctly, so other changes in this PR aren't needed. I'll set up a new PR because of the unnecessary changes here.

> The user can submit buffers of varying sizes.

Should we initialize the pool with differently sized arrays (maybe powers of two) and take the smallest one that's large enough when a buffer is requested? I don't know what are reasonable sizes for this.

tomspilman:
> Copying the buffer fixes the XAudio problems :) 

 Fantastic... thanks for verifying that.


> Should we initialize the pool with differently sized arrays (maybe powers of two)

I think you should just let the pool fill in naturally from submit buffer requests.

If the user is doing something reasonable... and submitting buffers at some consistent size then the pool will be reasonable too.  We can maybe just pad them out to something consistent if needed.

If the user is submitting buffers of random sizes like a crazy person then the pool will be mess, but that isn't our fault and we can't protect against that really.

Jjagg:
Closing in favor of #5056.

PR #5050, opened 2016-07-31
Texture2D.FromStream now correctly throws on null arguments

### LINKED ISSUE #5049: Better error for Texture2D.FromStream when graphics device is null.
I was recently working on loading textures with Texture2D.FromStream and kept getting the following error:

"This image format is not supported"

After fighting with this for about 30 minutes, I realized that the GraphicsDevice I was passing to the method as an argument was unintentionally null. The error that I received above was less than helpful when debugging this problem. Perhaps a better error could be throw when a null GraphicsDevice is provided (such as a NullReferenceException or an ArgumentException).

Jjagg:
Fixes #5049.

### DIFF
--- a/Test/Framework/Texture2DNonVisualTest.cs
+++ b/Test/Framework/Texture2DNonVisualTest.cs
@@ -82,7 +82,7 @@ namespace MonoGame.Tests.Framework
         [Test]
         public void FromStreamArgumentNullTest()
         {
-            Assert.Throws<ArgumentNullException>(() => Texture2D.FromStream(_game.GraphicsDevice, null));
+            Assert.Throws<ArgumentNullException>(() => Texture2D.FromStream(_game.GraphicsDevice, (Stream) null));
             Assert.Throws<ArgumentNullException>(() => Texture2D.FromStream(null, new MemoryStream()));
         }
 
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks @Jjagg .  Can you also add in a new unit test for these two cases here:

https://github.com/mono/MonoGame/blob/develop/Test/Framework/Texture2DNonVisualTest.cs

Test/Framework/Texture2DNonVisualTest.cs
```diff
@@ -77,6 +78,14 @@ public void FromStreamShouldFailTest(string filename)
                 Assert.Throws<InvalidOperationException>(() => _texture = Texture2D.FromStream(_game.GraphicsDevice, reader.BaseStream));
             }
         }
+
+        [Test]
+        public void FromStreamArgumentNullTest()
+        {
+            Assert.Throws<ArgumentNullException>(() => Texture2D.FromStream(_game.GraphicsDevice, null));
```
tomspilman: If you change this to `(Stream)null` it should build.

@dellis1972 - Do we need these weird overloads to `Texture2D.FromStream`?

  https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs#L371
  https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs#L377
  https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs#L412

?

tomspilman:
Thanks @Jjagg !

PR #5052, opened 2016-07-31
Content .targets fix

tomspilman:
Just adds the new platforms to the MGCB .targets file.


tomspilman:
@dellis1972 @KonajuGames - This should be good to go.

KonajuGames:
Merging.

PR #5056, opened 2016-08-01
Fix DynamicSoundEffectInstance stuttering

Jjagg:
Cleaned out and completed version of #5048.

### DIFF
--- a/MonoGame.Framework/Audio/DynamicSoundEffectInstance.XAudio.cs
+++ b/MonoGame.Framework/Audio/DynamicSoundEffectInstance.XAudio.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.Xna.Framework.Utilities;
 using SharpDX;
 using SharpDX.Multimedia;
 using SharpDX.XAudio2;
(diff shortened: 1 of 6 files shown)


MonoGame.Framework/Audio/DynamicSoundEffectInstance.XAudio.cs
```diff
@@ -91,9 +103,11 @@ private void OnBufferEnd(IntPtr obj)
             {
                 var buffer = _queuedBuffers.Dequeue();
                 buffer.Stream.Dispose();
+                _bufferPool.Return(_pooledBuffers.Dequeue());
```
Jjagg: I think we should lock the queues to get rid of the race condition here. At least here (for XAudio) since this function is run from another thread.

MonoGame.Framework/Utilities/ByteBufferPool.cs
```diff
@@ -0,0 +1,64 @@
...
+                    _freeBuffers.Add(buffer);
+                else
+                    _freeBuffers.Insert(index + 1, buffer);
+            }
+        }
+
+        // Find the smallest buffer that is larger than or equally large as size or -1 if none exist
+        private int FirstLargerThan(int size)
+        {
+            return _freeBuffers.FindIndex(b => b.Length >= size);
```
Jjagg: Binary search would be better, but I don't think it matters that much here.

KonajuGames:
Merging this.  It looks good to me, and we'll be able to get more people to test it once it is in develop.

PR #5058, opened 2016-08-02
Added TitleSafeArea

### LINKED ISSUE #4840: Fix TitleSafeArea
Looks like we've not implemented TitleSafeArea in MonoGame:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Viewport.cs#L145
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/DisplayMode.cs#L6
7

In XNA it was 20% of the screen, but see the notes Chuck Walbourn (XNA dev) that he left for us on this: https://github.com/mono/MonoGame/issues/4697#issuecomment-217779403.

We should be able to do 10% now as no one uses CRTs.

Jjagg:
Fixes #4840 as per Chuck Walbourns recommendations.

### DIFF
--- a/MonoGame.Framework/Graphics/DisplayMode.cs
+++ b/MonoGame.Framework/Graphics/DisplayMode.cs
@@ -64,7 +64,15 @@ namespace Microsoft.Xna.Framework.Graphics
         }
         
         public Rectangle TitleSafeArea {
-            get { return new Rectangle(0, 0, Width, Height); }    
+            get
+            {
+                var safeX = (width + 19) / 20;
+                var safeY = (height + 19) / 20;
+
+                var safeW = width - safeX * 2;
+                var safeH = height - safeY * 2;
+                return new Rectangle(safeX, safeY, safeW, safeH);
+            }
         }
 
         #endregion Properties
--- a/MonoGame.Framework/Graphics/Viewport.cs
+++ b/MonoGame.Framework/Graphics/Viewport.cs
@@ -146,14 +146,12 @@ namespace Microsoft.Xna.Framework.Graphics
 		{
 			get
 			{
-                // TODO : Possible change it for XBOX or PS4 
+                var safeX = (width + 19) / 20;
+                var safeY = (height + 19) / 20;
 
-                var iSafeStartX = x;
-                var iSafeStartY = y;
-                var iSafeWidth = width;
-                var iSafeHeight = height;
-
-                return new Rectangle(iSafeStartX, iSafeStartY, iSafeWidth, iSafeHeight);
+                var safeW = width - safeX * 2;
+                var safeH = height - safeY * 2;
+                return new Rectangle(safeX, safeY, safeW, safeH);
 			}
 		}
 


KonajuGames:
Thanks @Jjagg.  This looks good to me.  Merging.

mrhelmut:
:+1: 

This makes me wonder if there would be any use to add an ActionSafeArea of 5%? These are the typical areas required by most TCRs.

Jjagg:
> This makes me wonder if there would be any use to add an ActionSafeArea of 5%?

I wonder how relevant these things are nowadays. As far as I know noone has complained about TitleSafeArea not being in MG. It wouldn't hurt to have this though.

KonajuGames:
> I wonder how relevant these things are nowadays.

On desktop computers and mobile devices, there is no need for a safe area.  On consoles that display on TVs, even current HDTVs, there is still a need for a title safe area.  Much smaller than they used to be, but it is still necessary.

> This makes me wonder if there would be any use to add an ActionSafeArea of 5%? These are the typical areas required by most TCRs.

I remember Action Safe Areas being a late addition to one of the console platforms several years ago, but I think it was a recommendation rather than a requirement, and I don't think it's necessary here.

tomspilman:
We probably pulled the trigger on this too early.

Under XNA on desktop `TitleSafeArea` is 0% of the viewport/screen.  This is because the XNA build for desktops aims at desktops and desktops don't need a safe area.

XNA on 360 returns the 20% safe area.

To me this means we should only return a safe area for consoles.  We shouldn't return a safe area on desktops or mobile devices.

Jjagg:
Should I do a PlatformGetTitleSafeArea thing?

tomspilman:
> Should I do a PlatformGetTitleSafeArea thing?

Yeah I think so.

Like I could see some weird device requiring a very specific safe area and that would allow for it.

I would just add a `static Rectangle GraphicsDevice.PlatformGetTitleSafeArea(int x, int y, int width, int height)`  and let each `GraphicsDevice` overload it.

You might also add an internal helper like `GraphicsDevice.GetDefaultTitleSafeArea` that does the standard 10% calculation.

KonajuGames:
Yeah, a platform-specific implementation is a better idea.

tomspilman:
> This makes me wonder if there would be any use to add an ActionSafeArea of 5%?

@mrhelmut - So far in the console ports we've done the important part is the title safe area and 90% fixed is perfectly fine and passes cert.   So I don't think we need to complicate things further.

Jjagg:
So preprocessor directives? It seems overkill to make more partial classes of GraphicsDevice just for this

Jjagg:
But you want different values for XBox and Windows desktop right? They both use GraphicsDevice.DirectX, or am I missing something here?

tomspilman:
> They both use GraphicsDevice.DirectX, or am I missing something here?

They don't.

Jjagg:
> They don't.

Okay, then I know what to do :)

PR #5061, opened 2016-08-02
Split GetTitleSafeArea for platform specific implementation

Jjagg:
Follow up on #5058.

### DIFF
--- a/MonoGame.Framework/Graphics/Viewport.cs
+++ b/MonoGame.Framework/Graphics/Viewport.cs
@@ -144,7 +144,7 @@ namespace Microsoft.Xna.Framework.Graphics
         /// </summary>
 		public Rectangle TitleSafeArea 
 		{
-			get { return GraphicsDevice.GetTitleSafeArea(width, height); }
+			get { return GraphicsDevice.GetTitleSafeArea(x, y, width, height); }
 		}
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


MonoGame.Framework/Graphics/Viewport.cs
```diff
@@ -144,15 +144,7 @@ public Rectangle Bounds
...
-			get
-			{
-                var safeX = (width + 19) / 20;
-                var safeY = (height + 19) / 20;
-
-                var safeW = width - safeX * 2;
-                var safeH = height - safeY * 2;
-                return new Rectangle(safeX, safeY, safeW, safeH);
-			}
+			get { return GraphicsDevice.GetTitleSafeArea(width, height); }
```
tomspilman: This doesn't account for the X,Y position of the `Viewport`.  Shouldn't it?  We would have to test on X360 to be sure of the XNA behavior, but according to the docs it seems it should...

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.viewport.titlesafearea.aspx
"it reduces the title safe area of the viewport uniformly."

For example for a viewport of `500,500,1920,1080` this would return `96,54,1728,970` which is outside the original viewport.  I don't think that is what is intended.

Chuck's code also seemed to indicate XNA uses the original viewport's X/Y.

Jjagg: Jup, I'll fix this

PR #5065, opened 2016-08-03
[Pipeline Tool] General Fixes

cra0zy:
Fixes extracted from #4998

Includes:
 - fix type passing
 - adds char cell


tomspilman:
Thanks @cra0zy !

PR #5067, opened 2016-08-03
[SDL] Fixed Hardware mode switch change detection

dellis1972:
If we are already in fullscreen (soft or hard) and we only change the "HardwareModeSwitch" property we need to detect that and act accordingly.


tomspilman:
Seems ok to me ....  @cra0zy  @mrhelmut ?

cra0zy:
:+1:

mrhelmut:
Good to go! :)

PR #5068, opened 2016-08-03
[DO NOT MERGE] Idea for Pipeline Tool Tests

cra0zy:
@tomspilman This was my idea on how to handle Pipeline Tool Tests, what do you think?

PS. "FileTests" is as in "File Menu Tests"


tomspilman:
@cra0zy - I don't think this sort of testing is worth it.  In my experience mocks are hard to maintain and in the end all your testing is that the mocks work...  your application has not actually been testing.

I believe a combination of unit tests for utility and backing code plus UI functional tests are needed to properly test a GUI application.

PR #5069, opened 2016-08-03
GraphicsDeviceManager fixes

Jjagg:
This is not finished yet, but I'm sharing so people that are interested and/or worried can follow progress and advise. This is a start to getting GraphicsDevice and GraphicsDeviceManager events right and handling PresentationParameters the same as in XNA.

I may have done a seperate PR for this, but I also added a simple method to TestGameBase that initializes the game without creating a window. This can speed up a lot of test that are 'visual' but don't need to be. I already fixed a couple that just needed a GraphicsDevice, but not necessarily a window... Actually, most tests don't **really** need a window, right? Note that we can't get around that for XNA tests. Other tests that can benefit are RasterizerState and Viewport among others.

There's more testing to be done and GraphicsDevice events also needs its events fixed, but that might be better to do in a seperate PR. I did do the Disposed event though, because of how easy it was and I couldn't resist :p

This fixes at least (I'll add issues here as I find them)
- fixes #539
- fixes #4433
- fixes #3114

Issue #4704, opened 2016-04-02
Improve Unit Test Performance


tomspilman:
Since we [added code coverage](https://github.com/mono/MonoGame/pull/3820) the unit tests take 5 minutes to run where they used to take under 2 minutes.  This sounds trivial, but if we want to add any more tests and run tests for more platforms like DesktopGL then the testing needs to be as fast as possible.

We are currently running tests on a debug build of the framework so that code coverage has PDB files to work with.  Maybe if we used a debug build with optimization enabled it would improve the run time significantly.

This likely would require building the code a second time specifically for unit testing.  But building the code is literally less than 10 seconds.  It also solves to other issues:
- Smaller build artifact sizes.
- Code coverage will be able to find source code again.
- The test build steps can run in parallel with the code build steps.

This sounds like a win across the board.

hach-que:
How feasible is it to run the tests themselves in parallel (if that's not done already, but I assume it is not because it's NUnit which AFAIK doesn't support it)?

tomspilman:
> How feasible is it to run the tests themselves in parallel 

I'll have to research what NUnit supports there.  What worries me is that a lot of the tests pop up a GameWindow and do some rendering.  Having several of these running at once might cause issues.

For sure I can split things so that the Windows and WindowsGL tests run in parallel... still worried about them messing with each other.

tomspilman:
Also we plan to see if we can use something like Xamarin's Test Cloud for mobile testing.  This might involve calling in a bunch of favors at Xamarin and Microsoft to be able to do affordably.  The benefit is we can test thoroughly on difficult devices like Android.

PR #5072, opened 2016-08-04
Do not create window for tests to speed them up

Jjagg:
Part of this was originally done in #5069. This changes tests that were visual, but did not actually need that. This will speed up tests a bit and make it easier to implement new tests. Note that this is not yet finished.

For now this just does tests that do not need to draw, but do need a GraphicsDevice. While working on this I figured we can go further than that and even tests that do draw not use a window. Draw tests can render to a render target without actually running the game loop. That would make almost all tests not use a window and will improve testing performance and make it easier to implement new visual tests because you can just draw as you would in your game Draw function. Thoughts on this? Either way I'll do the non-drawing tests first.


Jjagg:
> Either way I'll do the non-drawing tests first.

Actually, it would take a lot less effort to do them all at once.
@KonajuGames @tomspilman

Jjagg:
After some experimenting, I found a way to simplify the tests that do drawing. I set up a gist with the visual test for BlendState before and after the change: https://gist.github.com/Jjagg/cf7e69c8c7f7b114648f82cad62700ac. 

Basically I'm expanding the GraphicsDeviceTestFixtureBase that is added in this PR to capture rendering and write out images and diffs like the VisualTestFixtureBase does, but without the components and events. Instead it has 3 functions that you need to use to check frames. 

- `PrepareFrameCapture` to setup a render target to draw to and do some initialization
- `SubmitFrame` to save the render target data
- `CheckFrames` to submit a frame when you've prepared one but not yet submitted it, then checks all captured frames against their respective reference images and writes out the captured image, diff or neither depending on settings (properties of the fixture).

I'm not entirely sure yet if this matters much for performance, but IMO tests are easier to read and write and we don't need to create a window. I'll do some more so I can measure performance. Even if there is no serious performance benefit, the fact that most tests won't create a window could help out with #4704. @tomspilman

KonajuGames:
I like the idea of this.  Just not having the window animation for every
time the window is created and destroyed while running the tests would be a
big improvement, and could make the PC usable when tests are running.

Jjagg:
I think the fixture is pretty much done now, but I still need to test with multiple captures (not sure if file naming is handled correctly for multiple captures). If that works, I'll seperate the logic a bit (Everything is in the fixture now) and then I'll push it here so you can check it out. I think I'll be able to get that up tonight.

Jjagg:
> tonight

I mean in a couple hours :p

Jjagg:
The GraphicsDeviceTestFixtureBase works as it should, so what's left here is changing all visual tests that don't require an actual game loop to use it.

tomspilman:
@Jjagg - The builds are now updating again... got some build errors:

```
[17:43:15]Framework\Graphics\GraphicsDeviceTestFixtureBase.cs(233,39): error CS0840: 'MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.FrameComparisonResult.Similarity.get' must declare a body because it is not marked abstract or extern. Automatically implemented properties must define both get and set accessors. [C:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\MonoGame.Tests.Windows.csproj]
[17:43:15]Framework\Graphics\GraphicsDeviceTestFixtureBase.cs(234,50): error CS0840: 'MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.FrameComparisonResult.CapturedData.get' must declare a body because it is not marked abstract or extern. Automatically implemented properties must define both get and set accessors. [C:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\MonoGame.Tests.Windows.csproj]
[17:43:15]Framework\Graphics\GraphicsDeviceTestFixtureBase.cs(235,51): error CS0840: 'MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.FrameComparisonResult.ReferenceData.get' must declare a body because it is not marked abstract or extern. Automatically implemented properties must define both get and set accessors. [C:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\MonoGame.Tests.Windows.csproj]
[17:43:15]Framework\Graphics\GraphicsDeviceTestFixtureBase.cs(236,47): error CS0840: 'MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.FrameComparisonResult.CapturedImagePath.get' must declare a body because it is not marked abstract or extern. Automatically implemented properties must define both get and set accessors. [C:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\MonoGame.Tests.Windows.csproj]
[17:43:15]Framework\Graphics\GraphicsDeviceTestFixtureBase.cs(237,48): error CS0840: 'MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.FrameComparisonResult.ReferenceImagePath.get' must declare a body because it is not marked abstract or extern. Automatically implemented properties must define both get and set accessors. [C:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\MonoGame.Tests.Windows.csproj]
[17:43:15]Framework\Graphics\GraphicsDeviceTestFixtureBase.cs(239,34): error CS0840: 'MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.FrameComparisonResult.Failed.get' must declare a body because it is not marked abstract or extern. Automatically implemented properties must define both get and set accessors. [C:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\MonoGame.Tests.Windows.csproj]
[17:43:15]Framework\Graphics\GraphicsDeviceTestFixtureBase.cs(240,37): error CS0840: 'MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.FrameComparisonResult.SaveImage.get' must declare a body because it is not marked abstract or extern. Automatically implemented properties must define both get and set accessors. [C:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\MonoGame.Tests.Windows.csproj]
[17:43:15]Framework\Graphics\GraphicsDeviceTestFixtureBase.cs(241,36): error CS0840: 'MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.FrameComparisonResult.SaveDiff.get' must declare a body because it is not marked abstract or extern. Automatically implemented properties must define both get and set accessors. [C:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\MonoGame.Tests.Windows.csproj]
```

Jjagg:
> got some build errors
Yeah, I didn't know I couldn't have auto-properties without setters.

Things left to do here:

- Fix failing SamplerState test... I have no idea what happens here but when testing CompareFunctions for the samplerstate CompareFunc.Never doesn't work. This is most likely because of a missing PlatfromBefore(Initialize|Run|Draw) or something. I've been looking at this for a while, but this is really really weird.
- OcclusionQuery test to new fixture. This is not too hard but slightly more complicated than the other tests because of time dependency. Maybe sleeping is enough for this, but simulating a game loop in the fixture might be useful in the future or for
- MiscellaneousTests to new fixture. The draw order test here is pretty complex, but I'll have a look to see  if I can get this implemented with the GD fixture. The ColorCubeTest is an easy one but I held that off because of the other one.
- Move stuff to the Graphics folder and update Protobuild definition and XNA csproj accordingly. Since there's no longer the distinction between visual and non-visual tests, it makes sense to just structure tests like MonoGame.Framework itself.
- I think I'm forgetting something :p

Jjagg:
I'm not sure how representative this is, but checking the timing on this test and the other most recent PR that ran all tests:
- this one:   1:01
- the other: 1:52

tomspilman:
That is pretty great... 2x improvement basically.  That will help as we get into more and more tests.

Jjagg:
> I think I'm forgetting something :p

I should also edit the README on visual tests!

Jjagg:
The SamplerState ComparisonFunction test still needs to be fixed, other than that this should be good. I really don't have a clue what's going wrong and how to fix it though...

Jjagg:
There are at least two tests that still create a window, both in Visual/MiscellaneousTests. One because it tests Game functionality, so simulating that would defeat the purpose of a test and the other because it's a complex test and it would take a lot of effort to make it use the new fixture

Jjagg:
The issue is here: https://github.com/MonoGame/MonoGame/blob/8a8f21c65c6e40688ed75b64c65bbf5a5790e40f/MonoGame.Framework/Graphics/States/SamplerState.DirectX.cs#L40

> ComparisonFunction != CompareFunction.Never

This should not be the check to determine if a comparison function should be used. I'll look into how this is working right now unless anyone knows?

Jjagg:
Actually a seperate property makes more sense because
1. The CompareFunction enum is used for other stuff as well
2. There is no None value in it in XNA

Jjagg:
@tomspilman What should I do with the ComparisonFunction here? If I leave it like this, CompareFunction.Never is broken. IMO the best thing we can do is add an extra property to enable comparefunction.

tomspilman:
@Jjagg 

So my thinking behind this was that the default value for `SamplerState.ComparisonFunction` is `CompareFunction.Never` so that must mean "off".

Sill very possible I was wrong in this assumption.  I'll looking closer at this now.

tomspilman:
Ok... so I think I better understand what is going on now.

We added `SamplerState.ComparisonFunction` to expose a feature from DirectX11 class GPUs.  This doesn't exist in XNA at all.

There are two parts to this issue:

  1. Why did the test break?
  2. How should we define don't compare in `SamplerState.ComparisionFunction`?

We should know why this broke...  i'll investigate that next here.


So for number 2.... looking at the DirectX docs you can see 0 means don't do comparison:

https://msdn.microsoft.com/en-us/library/windows/desktop/ff476101(v=vs.85).aspx

So under DirectX technically there is a `None`.

The problem we have is that the `CompareFunction` is shared and used in `DepthStencilState` and `AlphaTestEffect` too.  So we don't want to introduce a new value to the enum that breaks things in those existing cases.

IMO `CompareFunction.Always` (meaning always pass the test) seems like a more logical default to mean "don't do comparison tests".  This is how `DepthStencilState.StencilFunction` works:

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.depthstencilstate.stencilfunction.aspx

In contrast `DepthStencilState.DepthBufferFunction` defaults to `CompareFunction.LessEqual` and `AlphaTestEffect.AlphaFunction` defaults to `CompareFunction.Greater`.

The lesson here is that each usage considers a different value of `CompareFunction` as the default behavior... so we can do that too without introducing a new value for "none".

So either we can stick with `CompareFunction.Never` or switch to `CompareFunction.Always` as the default "don't do comparison tests" value.  However that isn't the issue here... the issue is why the unit tests broke.

tomspilman:
Ok this is where `CompareFunction.Never` as the default value came from:

https://msdn.microsoft.com/en-us/library/windows/desktop/ff476207(v=vs.85).aspx

If you scroll to the bottom you'll see that it indicates the default value for sampler states is `D3D11_COMPARISON_NEVER`.

So the only question here is why did the test break.  We don't need to change anything else.

Jjagg:
> We should know why this broke... i'll investigate that next here.

Thanks Tom. I've been looking at this, but just couldn't find what broke it. Or rather why it worked in the first place.

tomspilman:
@Jjagg - So the issue here is that the writing the rendered frame to disk is failing causing the comparison process to fail.  This is the exception that triggers the failure:

```
Unhandled exception in remote appdomain: System.Runtime.InteropServices.ExternalException (0x80004005): A generic error occurred in GDI+.
   at System.Drawing.Image.Save(String filename, ImageCodecInfo encoder, EncoderParameters encoderParams)
   at MonoGame.Tests.FramePixelData.Save(String filename) in c:\BuildAgents\MonoGameWin1\work\59a4b66548a5cef5\Test\Runner\Desktop\FramePixelData.cs:line 65
   at MonoGame.Tests.Graphics.GraphicsDeviceTestFixtureBase.<>c__DisplayClass7.<CheckFrames>b__1() in c:\BuildAgents\MonoGameWin1\work\59a4b66548a5cef5\Test\Framework\Graphics\GraphicsDeviceTestFixtureBase.cs:line 200
   at MonoGame.Tests.Utilities.ActionDaemon.DoActions() in c:\BuildAgents\MonoGameWin1\work\59a4b66548a5cef5\Test\Utilities\ActionDaemon.cs:line 56
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

Looks like you added some threading of the work in `GraphicsDeviceTestFixtureBase.CheckFrames()` where as previously there was none in `VisualTestFixtureBase.cs`.  This makes me think we have some sort of threading related bug here.

Jjagg:
Hmm, saving images and diffs was done in a seperate thread for visual tests as well, I borrowed some code from there. Maybe I made a mistake somewhere. That's not what causes this to fail when I run the test locally though. The image writes out fine, but does not render the block with CompareFunction.Never right. It just renders the texture and debugging showed that it does not set the filter of the sampler to a comparison filter. It doesn't do that for the test before the changes either, which is why I don't have a clue how this test passed before

Jjagg:
@tomspilman I set up a test project to show why I'd like to change this. https://github.com/Jjagg/SamplerState-issue
If you switch out the MG reference it should be good to run. Use the left and right arrow keys to cycle through comparison functions. When CompareFunction.Never is selected it'll start flickering between the original texture and a completely black texture (which is what we'd expect with CompareFunction.Never), because the filter is not set correctly.

tomspilman:
@Jjagg 

So this is strange...  here the `CompareFunction.Never` renders black and does not flicker at all.  This is with the latest MonoGame develop code.  I have a Radeon HD 9600 in my system here.

This suggests a possible driver issue on your system maybe?

Running the unit test here it passes with the correct results for all `CompareFunction`s.


> not using a comparison filter gives some weird results.

Are you thinking that `CompareFunction.Never` is equivalent to "not using a comparison filter"?

I went and checked again and using `CompareFunction.Never` is the default behavior in all DirectX titles for every texture not using comparisons.  If you do a general search for `D3D11_SAMPLER_DESC` in GitHub:

https://github.com/search?l=cpp&q=D3D11_SAMPLER_DESC&type=Code&utf8=%E2%9C%93

`D3D11_COMPARISON_NEVER` is really the default behavior.

Jjagg:
> So this is strange... here the CompareFunction.Never renders black and does not flicker at all. This is with the latest MonoGame develop code. I have a Radeon HD 9600 in my system here.

I think behaviour might be undefined when you use SampleCmp without a comparison filter set for the sampler you pass.

> Running the unit test here it passes with the correct results for all CompareFunctions.

It's annoying that this is graphics card dependent, but the unit test fails on the build server with the same similarity score as on my machine, so most likely the same issue occurs there.

The [SampleCmp docs](https://msdn.microsoft.com/en-us/library/windows/desktop/bb509696(v=vs.85).aspx) say on the sampler that should be passed:

S [in] A sampler-comparison state, which is the sampler state plus a comparison state (a comparison function and a **comparison filter**). See the sampler type for details and an example.

> Are you thinking that CompareFunction.Never is equivalent to "not using a comparison filter"?

Not inherently, but the way we do it, yes. That's exactly the problem.
When we set the filter [here](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/States/SamplerState.DirectX.cs#L40) we check `ComparisonFunction != CompareFunction.Never` to see if we should use a comparison filter. Meaning when ComparisonFunction is set to CompareFunction.Never, we never use a comparison filter.

tomspilman:
>  we try to do comparisons without a comparion filter

I see...  i'm following what you're saying now.

The issue is in the line you linked:
```
desc.Filter = GetFilter(Filter, ComparisonFunction != CompareFunction.Never);
```
We cannot use `CompareFunction.Never` or any other valid comparison value in `CompareFunction` to decide if `Filter.ComparisonXXXX` needs to be used or not.

So something does indeed need to change there.

I think one possible fix...  maybe the most "right" fix.... is to match what DirectX does for `D3D11_FILTER` (this is also the same in `D3D12_FILTER`):
https://msdn.microsoft.com/en-us/library/windows/desktop/ff476132(v=vs.85).aspx

```
	public enum TextureFilter
	{
		Linear,
		Point,
		Anisotropic,	
		LinearMipPoint,
		PointMipLinear,
		MinLinearMagPointMipLinear,
		MinLinearMagPointMipPoint,
		MinPointMagLinearMipLinear,
		MinPointMagLinearMipPoint,

		ComparisonLinear,
		ComparisonPoint,
		ComparisonAnisotropic,	
		ComparisonLinearMipPoint,
		ComparisonPointMipLinear,
		ComparisonMinLinearMagPointMipLinear,
		ComparisonMinLinearMagPointMipPoint,
		ComparisonMinPointMagLinearMipLinear,
		ComparisonMinPointMagLinearMipPoint
	}
```
Then users that want to use comparison filtering need to select the right filter just like you do in DirectX.  Still I find this a bit verbose when all we need is a simple flag indicating if we do or do not want comparison filtering.

Another possible fix would be making `SamplerState.ComparisonFunction` a nullable property.  Then by default the comparison function is null and unset meaning no comparisons.  It is a little weird as you don't often see nullable properties...  but it limits the changes to just `SamplerState` without affecting other parts of MonoGame.

The maybe less weird fix would be just adding a simple `bool SamplerState.ComparisonFilterEnable` property like `RasterizerState.DepthClipEnable` or `RasterizerState.ScissorTestEnable`.  Simpler than the nullable property and still only affects `SamplerState`.

A more forward looking fix would be adding a new enum:
```
	public enum TextureFilterMode
	{
		Default,
		Comparison,
		// Minimum,
		// Maximum,    
	}
```
So if we ever start to support the additional filter modes offered by GPUs we have a place to put it without having to add more properties to `SamplerState`.  Also wouldn't have to deal with conflicting filter modes from multiple booleans.  This one sort of solves the mess in `D3D11_FILTER` and might be easier to work with in OpenGL and console platforms.  I think this is the best fix we can implement here.

IMO the worst option would be adding a `CompareFunction.None`.  It would work for deciding if no texture comparison was needed, but it would affect all the other places we use `CompareFunction`.  In all other cases we would have to decide what `None` means in that context if the user sets it.

tomspilman:
@Jjagg - I think for this PR you should just disable the `VisualTestComparisonFunction` test and get this PR merged.

You can then follow up with another PR fixing our implementation of comparison filtering and re-enabling the test.

Jjagg:
> A more forward looking fix would be adding a new enum:

I added a bool locally, but this sounds better. 

> You can then follow up with another PR fixing our implementation of comparison filtering and re-enabling the test.

Will do!

Jjagg:
By the way, should we worry about the threading crash you had? This is the first time I've done multithreading in C#, but it all seemed pretty straightforward. I didn't have any issues locally and looks like the build server managed too.

tomspilman:
> By the way, should we worry about the threading crash you had? 

I didn't run into the crash myself...  that callstack was from the build server.

We'll just wait and see if the next build with that test disabled has any issues..

tomspilman:
So there is still something weird happening with these changes.

A lot of the image diffs are missing in the final artifacts.  You can compare the ones on the latest develop build:

http://teamcity.monogame.net/viewLog.html?guest=1&buildId=22167&buildTypeId=MonoGame_TestWindows&tab=artifacts

To the ones from the latest build of this PR:

http://teamcity.monogame.net/viewLog.html?guest=1&buildId=22195&buildTypeId=MonoGame_TestWindows&tab=artifacts

Ignore the `Test` folder... just look at the zips.  You can see the missing `diff-` files that are generated as the tests run.

There are no errors in the output related to this that I see.  Also checked the build server and the diff files are indeed not there meaning the test runner didn't generate them.

I suggest looking closer at the code that generates the diff images and see if maybe we're not blocking to wait until they are processed before letting the test process exit.

Jjagg:
Oh, that was intentional. I added a setting for captured images and diffs so you can set it that they only get written out when the test fails. I set it that way by default for the diffs. Sorry, forgot to mention that... Want me to change that so they always get written out again?

tomspilman:
> Oh, that was intentional

Ok...  that is fine for the diffs.  It cuts down on the artifact size some which isn't bad when the diff passes anyway.

I'm good with merging this now.

tomspilman:
FYI.  Going to squash merge this in hopes it then correctly tracks moves of files.

tomspilman:
Lame...  we lost the history connection on the moved graphics tests:

https://github.com/MonoGame/MonoGame/tree/develop/Test/Framework/Graphics

@Jjagg - In the future if you move a file, be sure to commit both the new file and the deletion together.  This way Git can detect and reconnect he history with the old file.

Jjagg:
Crap... Sorry. I'll watch out for this

KonajuGames:
I've been running the unit tests locally a lot over the past few days to
sort out a PixelBitmapContent issue.  Just before I was going to push the
changes, I pulled latest which included this change.  Ran the unit tests
again and it failed towards the end of the run with an error message about
failing the change context between COM objects after 60 seconds.  Spotted
it, ran it again and it locked up the entire PC. No solution but to kill
power. It was 2am so I stopped for the night.

I was wondering if you had seen this?  I have never seen it before in all
the times I have run the unit tests.

KonajuGames:
I usually run the unit tests in Debug configuration.

Jjagg:
@KonajuGames I didn't run into this. I always run in Debug too. Not sure what this could be...

Jjagg:
@KonajuGames This might be because of undisposed graphics resources, I thought we disposed of all of those when the GraphicsDevice they're on is disposed, but there seems to be an issue with that. I wrote a quick test with a render target and disposing game didn't dispose the RT. XNA also doesn't dispose of GraphicsResources when game is disposed, so best to always dispose unmanaged resources in the tests themselves . I'll go through the tests and dispose of all stuff that currently isn't disposed, then hopefully this is fixed. 

Wow, I wrote dispose a lot in this :p

KonajuGames:
I tried again the following day and it got through the unit tests fine.

I noticed there are a **_lot_** of undisposed resources at the end of a tests run.  In debug configuration, DirectX lists all undisposed resources and this basically slows my PC to a crawl for 20-30 seconds.

Jjagg:
I'll get a PR up ASAP to fix this

tomspilman:
@Jjagg - Remember we need to get the graphics tests to default to the reference rasterizer by default.  That is part of the reason the unit tests are failing on the develop branch currently.

KonajuGames:
#5424 fixes that so the reference rasterizer is always created, not just for the visual tests. It passes all the tests.

tomspilman:
Thanks @KonajuGames ... this works around the issue and gets things back to testing how we did before.

We still need to fix the actual bug here which is incorrectly detecting the `HiDef` profile.

PR #5075, opened 2016-08-04
Fix crash when DynamicSoundEffectInstance not disposed

### LINKED ISSUE #5064: Track DynamicSoundEffectInstances so they can be disposed
For **XAudio**, when there are voices that are not yet disposed when [MasterVoice is disposed](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffect.XAudio.cs#L351) a crash will occur as noted in [`SoundEffect.Shutdown`](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstancePool.cs#L160). This caused our unit tests to fail in #5042. 

We currently have [`SoundEffectInstancePool`](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstancePool.cs) to manage all SoundEffectInstances (and dispose of them in the Shutdown method) and [`DynamicSoundEffectInstanceManager`](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/DynamicSoundEffectInstanceManager.cs) to manage **playing** DynamicSoundEffectInstances, but not ones that are not currently playing. I suggest we let the DSEIManager handle this by having it keep a WeakReference to all instances that are not playing as well.

### LINKED ISSUE #5070: Wrong usage of weakreferences in DynamicSoundEffectInstanceManager
I already commented this in the commit, but seems to be ignored or something.

In DynamicSoundEffectInstanceManager.UpdatePlayingInstances() is an iteration over all playingInstances with a check of IsAlive and getting the target of a WeakReference. This can generate in very rare cases NullReferenceExceptions because the GC removes the instance between these lines of code.

The check should - like in all other cases of weak references I found in the source code - check Target directly to null.

Jjagg:
I had a look at the following SharpDX sample 
https://github.com/sharpdx/SharpDX-Samples/blob/master/Desktop/XAudio2/PlaySound/Program.cs#L42

and MasteringVoice is just disposed without calling `MasteringVoice.DestroyVoice`. This avoids the crash when other SFX are not disposed. I did a test for XNA and disposing a Game does not dispose of sound effects that are not yet disposed, so it's left to the user there too. If this is the right way of shutting down XAudio2 we can also remove `SoundEffect.PlatformShutdown` in `Game.Dispose`.

fixes #5064, fixes #5070

### DIFF
--- a/MonoGame.Framework/Audio/DynamicSoundEffectInstanceManager.cs
+++ b/MonoGame.Framework/Audio/DynamicSoundEffectInstanceManager.cs
@@ -47,9 +47,9 @@ namespace Microsoft.Xna.Framework.Audio
         {
             for (int i = _playingInstances.Count - 1; i >= 0; i--)
             {
-                if (_playingInstances[i].IsAlive)
+                var target = _playingInstances[i].Target as DynamicSoundEffectInstance;
+                if (target != null)
                 {
-                    var target = (DynamicSoundEffectInstance)_playingInstances[i].Target;
                     if (!target.IsDisposed)
                         target.UpdateQueue();
                 }
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -350,7 +350,6 @@ namespace Microsoft.Xna.Framework.Audio
 
             if (MasterVoice != null)
             {
-                MasterVoice.DestroyVoice();
                 MasterVoice.Dispose();
                 MasterVoice = null;
             }
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -339,8 +339,6 @@ namespace Microsoft.Xna.Framework.Audio
 
         internal static void PlatformShutdown()
         {
-            SoundEffectInstancePool.Shutdown();
-
             if (_reverbVoice != null)
             {
                 _reverbVoice.DestroyVoice();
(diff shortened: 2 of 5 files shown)


tomspilman:
> MasteringVoice is just disposed without calling  MasteringVoice.DestroyVoice 

That does make sense...  basically the voice isn't destroyed and we just decrement our ref count.

>  I did a test for XNA and disposing a Game does not dispose of sound effects that are not yet disposed, so it's left to the user there too.

Then it makes sense for us too.


> If this is the right way of shutting down XAudio2 we can also remove  SoundEffect.PlatformShutdown  in  Game.Dispose .

We still want to call it...  because we support other platforms too.  We just don't need the XAudio2 implementation to do anything when it is called.

Jjagg:
Oops, I did not mean SoundEffect.PlatformShutdown, but SoundEffectInstancePool.Shutdown @tomspilman

tomspilman:
> Oops, I did not mean 

I just mean that not all platforms operate like XAudio2...  so we have to give each platform at least a chance at doing any final cleanup for systems.

> If you want I'll add in the test too

Sue tests are always great to have.

Jjagg:
Yeah, but PlatformShutdown calls SoundEffectInstancePool.Shutdown to dispose of all SoundEffects and XNA does not dispose of soundeffects when disposing game. So that can be removed (and should be to match XNA behaviour).

Jjagg:
This should be good to merge, but DynamicSoundEffectInstance still needs to be made thread safe.

Jjagg:
@tomspilman This can be merged. I think DynamicSoundEffects can rarely crash now because of threading issues, but other than that they should be good! :)

tomspilman:
Thanks @Jjagg ... merging!

PR #5077, opened 2016-08-06
Allow #line directives to be used in .fx files

hach-que:
This allows `#line` directives to be used in effect files.

As part of Protogame, we have our own asset system which underneath passes effect code to MonoGame's content pipeline for compilation.  To do this, we write the effect code to a temporary file on disk, and tell the effect processor to compile that file.  We preemptively resolve includes on the string in memory before we pass the effect code to MonoGame, so there are no external dependencies on other files.

Right now, the CppNet preprocessor strips out any `#line` directives in the source code, even though they are recognised by the underlying shader compiler.  We want to use `#line` directives when we pass the content into MonoGame's content pipeline so that the filenames and line numbers match with our actual source files, and don't just have the temporary filename on disk (and incorrectly offset line numbers due to includes).

This change detects `#line` directives before they are passed into CppNet, and encodes them as a comment.  When reading the output from the preprocessor, they then get transformed back into `#line` directives again so that they exist for the underlying shader compiler.

### DIFF
--- a/Tools/2MGFX/Preprocessor.cs
+++ b/Tools/2MGFX/Preprocessor.cs
@@ -24,6 +24,8 @@ namespace TwoMGFX
             foreach (var define in defines)
                 pp.addMacro(define.Key, define.Value);
 
+            effectCode = effectCode.Replace("#line", "//--WORKAROUND#line");
+
             pp.addInput(new MGStringLexerSource(effectCode, true, fullPath));
 
             var result = new StringBuilder();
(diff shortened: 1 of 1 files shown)


tomspilman:
> Right now, the CppNet preprocessor strips out any `#line` directives in the source code

That seems odd...  it shouldn't be doing that.  CppNet resolves our `#includes` and generates `#line` directives ... so very strange it does this when it is passed `#line`s.

I've added an issue for it that hopefully will be addressed in the future: https://github.com/xtravar/CppNet/issues/5 

> encodes them as a comment.

This seems like a reasonable workaround.

tomspilman:
Going to go fix the build status reporting... looks like the repo move broke it.

PR #5083, opened 2016-08-06
DX11 Multisample Quality fix

KonajuGames:
DX11 Multisample Quality was being set to the maximum, even if the desired multisample quality was lower.  For example, the dev selects a multisample count of 4.  The code checks for the maximum available, which in my case was 8.  This returns a quality level of 33.  The code previously was setting multisampleDesc.Count to 4 but multisampleDesc.Quality was being set to 32 (quality level - 1 for a multisample count of 8).  This resulted in an incorrect argument HRESULT when creating the swapchain.

The change here now calculates the quality level for the selected multisample count, not the highest.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -714,7 +714,9 @@ namespace Microsoft.Xna.Framework.Graphics
                     PresentationParameters.MultiSampleCount = maxLevel;
 
                 multisampleDesc.Count = PresentationParameters.MultiSampleCount;
-                multisampleDesc.Quality = qualityLevels - 1;
+                // Get the quality level for the selected multisample count, which may be
+                // lower than the maximum found above.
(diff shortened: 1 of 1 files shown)


tomspilman:
Is there any useful unit tests for this?

Seems like all we could test is if setting MSAA settings doesn't crash?

KonajuGames:
It depends on the capabilities of the host GPU. Mine happened to support 8x
when the project only wanted 4x.

tomspilman:
Ok... merging!

hach-que:
Pretty sure this wasn't the right fix?

The reason for the maximum check was to handle scenarios where the game was asking for a multisample count *higher* than what the graphics card supports.  In that scenario, the logic was meant to go up to the multisample count that the game requested (obviously that was buggy).

However with this change it no longer does that, and just always selects the game's specified multisample count, even if the GPU can't support it.  When this happens, an exception will be thrown as per #2465.

I'm about to rebase and update #4993, and I'll include a fix for this issue in that as well since I'm having to merge conflicts on these lines changed.

Issue #4730, opened 2016-04-05
Support XNA GraphicsDevice Embedding


tomspilman:
Fix MonoGame to support the existing XNA style `GraphicsDevice` embedding as done in this tutorial:

 http://xbox.create.msdn.com/en-US/education/catalog/sample/winforms_series_1

First we need to be sure this works unchanged on Windows with DirectX.

Second we need to make it work under OpenGL on Windows with the simplest possible change to the technique.

Third we need to be sure the technique works on Mac and Linux which hopefully the fixes to support OpenGL make easier.

Once this is all in place we can investigate a method for embedding a `Game` expanding upon the techniques developed here.

PR #5084, opened 2016-08-06
VideoPlayer.WMS improvements and features (take 2)

KonajuGames:
Recreated from PR #4811

Create video texture once and re-use it instead of creating a new texture each frame.
Implement looped video support.

Video events should be dispatched via the FrameworkDispatcher once it is implemented.


romanov:
:pray:

tomspilman:
@KonajuGames - Some test failures on this one.

romanov:
Are TestMachines terminals?

KonajuGames:
There are some issues with these video unit tests on the test machine.  I believe it may be because the test machine is headless.

I also added some multiple VideoPlayer unit tests and found that there are some cross-contamination issues.  The output from the second VideoPlayer is affected by the first, or the other way around.  Yet to find a solution for that.

spunky44:
I'm new to GitHub. If I want to apply these commits to my own repository, in order to get videos playing again, is there an easier way than cloning the repository and creating a patch with the command line, then applying the patch to my own clone?

spunky44:
Any idea on when this will be finished..?

Jjagg:
> If I want to apply these commits to my own repository, in order to get videos playing again, is there an easier way than cloning the repository and creating a patch with the command line, then applying the patch to my own clone?

You can add @KonajuGames fork as a remote and pull in this branch. Then check it out or merge/cherry-pick the relevant commits into a branch of your own.

spunky44:
Thanks. Is this how to do it:

https://help.github.com/articles/merging-an-upstream-repository-into-your-fork/

KonajuGames:
There are still issues with playing multiple videos at the same time, but
if you only play one video at a time it works OK.

spunky44:
No, playing one video has poor performance and causes the game to grind to a halt.

KonajuGames:
With this PR that playback issue is fixed. It still has an issue with
playing multiple concurrent videos though.

spunky44:
Great, I want to merge it so I can release my game. Is the link I posted the right way?

KonajuGames:
That is how to merge from my branch into yours. One thing I would do
differently though is make a branch in your local repo first so you don't
merge into your local master or develop branch (to avoid messing them up if
something goes wrong or later pulls from upstream cause conflicts).

spunky44:
It looks like it's working.

romanov:
I want this fix in the distro  :clock1:

romanov:
@spunky44 can you post links to the binaries with the fix?

Test/Framework/Visual/VideoTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        [SetUp]
+        public override void SetUp()
+        {
+            base.SetUp();
+
+            Game.LoadContentWith += (sender, e) =>
+            {
+                _spriteBatch = new SpriteBatch(Game.GraphicsDevice);
+                _video = Game.Content.Load<Video>(Paths.Video("SampleVideo_360x240_5s"));
+                _player.Play(_video);
```
Jjagg: @KonajuGames Can we take another look at getting this resolved? _player is just not initialized here. 

Not necessarily for this PR, but these tests could be improved by saving some reference frames for the videos. That way this does not need to actually run a game and can use the GraphicsDeviceTestFixtureBase instead of the VisualTestFixtureBase. Checking the frames makes the tests more robust too.

MonoGame.Framework/Media/VideoPlayer.WMS.cs
```diff
@@ -59,102 +87,132 @@ private void PlatformInitialize()
...
+                {
+                    _texture.Dispose();
+                    _texture = null;
+                }
+            }
 
-            if (texData == null)
-                return null;
+            if (_texture == null)
+                _texture = new Texture2D(Game.Instance.GraphicsDevice, _currentVideo.Width, _currentVideo.Height, false, SurfaceFormat.Bgr32);
```
Jjagg: It would be nice to get these Game.Instance references out of the code at least for desktop. That makes it easier for people to embed MG (#4730) or write a custom Game class and potentially we can allow running multiple games at once. 

For this class it would be nice if it could be run without a Game instance being present too. I can think of two ways to do this, one that breaks with the XNA API and one that doesn't (but it's ugly).

1. Have VideoPlayer take a GraphicsDevice in its construction and store it. This obviously would break XNA games using the VideoPlayer, but it's just a small change and people will notice at compile time. I actually prefer this because IMO it's cleaner than option 2 and it makes sense to have a video player linked to a GraphicsDevice this way.
2. Have some global static service container that we can pull a GraphicsDevice or a GraphicsDeviceManager (and throught that a GraphicsDevice) from. GD or GDM would register on construction just like how GDM does that for Game currently. From a software design point of view I think this is ewwww.

MonoGame.Framework/Media/VideoPlayer.WMS.cs
```diff
@@ -59,102 +87,132 @@ private void PlatformInitialize()
...
+
+        bool WaitForInternalStateChange(InternalState expectedState, int milliseconds = defaultTimeoutMs)
+        {
+            var timer = System.Diagnostics.Stopwatch.StartNew();
+            while (_internalState != expectedState)
+            {
+#if WINRT
+                Task.Delay(1).Wait();
+#else
+                Thread.Sleep(1);
```
Jjagg: This seems like a bad idea. Can we add checks in other functions and let the game keep running instead of blocking everything here?

KonajuGames: XNA is inherently single threaded, meaning that the action must complete before control returns to the caller.  I don't like the delays either, but it must wait for the async action to complete before returning to the caller.

KonajuGames:
Git can be a PITA sometimes.

Jjagg:
Heh, yeah :p How are you rebasing? I always first make sure my develop is up to date and then do `git rebase develop` while on the branch I want to rebase. Fix issues as required during the rebase and when it's done `git push --force`. I never had this issue.

KonajuGames:
Tried a rebase and edit.  Thought it had worked.  Too late to fix this now.  Will have to do it tomorrow night. :rage:

Jjagg:
Sounds good!

KonajuGames:
I did `git rebase upstream/develop` and fixed each issue as it came up (mostly no actual error on `MonoGame.Tests.definition`).

MonoGame.Framework/Media/VideoPlayer.WMS.cs
```diff
@@ -59,102 +87,132 @@ private void PlatformInitialize()
...
+            System.Diagnostics.Debug.WriteLine("PlatformStopped");
+        }
+
+        bool WaitForInternalStateChange(InternalState expectedState, int milliseconds = defaultTimeoutMs)
+        {
+            var timer = System.Diagnostics.Stopwatch.StartNew();
+            while (_internalState != expectedState)
+            {
+#if WINRT
+                Task.Delay(1).Wait();
```
MichaelDePiazzi: `Task.Delay` actually generates garbage. An alternative which works just as well and doesn't generate garbage is `Monitor.Wait`. PR #4207 has an example of this (specifically commit 0a18376).

KonajuGames:
Replaced by #5437

PR #5086, opened 2016-08-06
Fix Android launch (no EFX)

KonajuGames:
The code assumed that EFX was available if it wasn't built for MONOMAC.  Replaced that with a SUPPORTS_EFX symbol that is defined if it is not MONOMAC and not GLES because the GLES platforms still use OpenTK.OpenAL.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -42,10 +42,10 @@
       <Platform Name="Android">TRACE;ANDROID;GLES;OPENGL;OPENAL</Platform>
       <Platform Name="iOS">IOS;GLES;OPENGL;OPENAL</Platform>
       <Platform Name="tvOS">IOS;TVOS;GLES;OPENGL;OPENAL</Platform>
-      <Platform Name="Linux">TRACE;LINUX;OPENGL;OPENAL;DESKTOPGL</Platform>
-      <Platform Name="MacOS">MONOMAC;OPENGL;OPENAL</Platform>
+      <Platform Name="Linux">TRACE;LINUX;OPENGL;OPENAL;DESKTOPGL;SUPPORTS_EFX</Platform>
+      <Platform Name="MacOS">MONOMAC;OPENGL;OPENAL;SUPPORTS_EFX</Platform>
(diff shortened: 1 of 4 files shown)


tomspilman:
Some build errors.

YugeshN:
Looks like the block:

    +#if !MONOMAC && !GLES
     +#define SUPPORTS_EFX
     +#endif

in OpenALController.cs needs to be moved to the top of the file since it contains a #define. That preprocessor limitation is annoying...

KonajuGames:
Too much time spent in C++ recently.

KonajuGames:
And Xamarin.Android did not complain about it when I made that change last night.

YugeshN:
Visual Studio would definitely have caught it.... that's how I learnt about that limitation in c# :-)

KonajuGames:
> Visual Studio would definitely have caught it.... that's how I learnt about that limitation in c# :-)

Microsoft's C# compiler catches it.  Xamarin.Android does not appear to catch it.

Build/Projects/MonoGame.Framework.definition
```diff
@@ -42,10 +42,10 @@
       <Platform Name="Android">TRACE;ANDROID;GLES;OPENGL;OPENAL</Platform>
       <Platform Name="iOS">IOS;GLES;OPENGL;OPENAL</Platform>
       <Platform Name="tvOS">IOS;TVOS;GLES;OPENGL;OPENAL</Platform>
-      <Platform Name="Linux">TRACE;LINUX;OPENGL;OPENAL;DESKTOPGL</Platform>
+      <Platform Name="Linux">TRACE;LINUX;OPENGL;OPENAL;DESKTOPGL;SUPPORTS_EFX</Platform>
       <Platform Name="MacOS">MONOMAC;OPENGL;OPENAL</Platform>
```
tomspilman: Shouldn't SUPPORTS_EFX be here too?

KonajuGames: EFX was previously wrapped in a !MONOMAC define.  I had SUPPORTS_EFX in that line before, but that line defines MONOMAC, and that is what caused the last build failure.

tomspilman: Oh... I see `!MONOMAC`.

PR #5090, opened 2016-08-07
Fixed Effect.OnApply to return void to match XNA

### LINKED ISSUE #1303: Effect.OnApply() return value
The function can be overridden in XNA 4.0 by
protected override void OnApply()

but in MonoGame it is
protected override bool OnApply()

Jjagg:
Fixes #1303. 

This breaks with older MG API, so that should probably be mentioned in the 3.6 release.


tomspilman:
This is great @Jjagg .... simpler and much less code than the hack I did.

PR #5091, opened 2016-08-07
Move MonoGame.Framework.Net to its own folder

### LINKED ISSUE #4794: Move MonoGame.Framework.Net to its own folder
We want to move the code for the MonoGame.Framework.Net assembly into its own top level folder like we do for ``.  This would give us three top level folders where the core assembly code is found:
- MonoGame.Framework
- MonoGame.Framework.Net
- MonoGame.Framework.Content.Pipeline 

During the moving of this code _NO_ other code changes should occur.

The list of all the code files that need to be moved can be found here:

Note that the `MonoGame.Framework\Properties\AssemblyInfo.cs` should be copied and modified for the new assembly:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Properties/AssemblyInfo.cs

Also the `Path` in the protobuild definition needs to be changed:
  https://github.com/mono/MonoGame/blob/develop/Build/Projects/MonoGame.Framework.Net.definition#L2

The installers for Mac, Linux, and Windows will need to be updated to pull from the new output bin folders.

Finally the configuration of the build server will have to change to pull from the new bin folders as well.

The idea was originally proposed in https://github.com/mono/MonoGame/issues/3055.

This also solves https://github.com/mono/MonoGame/issues/2981 and https://github.com/mono/MonoGame/issues/3480 if they haven't been resolved by fixes in XamarinStudio already.

Jjagg:
Fixes #4794. 

I don't know who usually does the installers. I did some thorough searching, but I might have missed something. Is there any way to test this?


tomspilman:
Great to see this happen...  there will likely be things to fix in the build as well on the build server.

I must run out for a bit, but i'll be back later to see what I need to fix on my end.

DDReaper:
Any update on this @tomspilman / @Jjagg ?
Noticed this still has WindowsPhone targets so will be impacted by #5336

Jjagg:
Closing because of #5378

PR #5093, opened 2016-08-09
[xact] Make sure the Cue is properly reset

dellis1972:
When using GetCue in XNA the Cue is always
returned in its initial state. There was a bug
where the Cue would be returned with a IsStopped
value of 'true' if the cue had been played before.
This did not match the expected XNA behaviour.


tomspilman:
This is great... with a unit test too... fantastic!

PR #5095, opened 2016-08-09
Set session state to proper value when resuming.

sim31:
On windows uwp platform when resuming a song, sessionState was not
reset to started, which resulted in failing to pause a song later

### DIFF
--- a/MonoGame.Framework/Media/MediaPlayer.WME.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.WME.cs
@@ -147,6 +147,7 @@ namespace Microsoft.Xna.Framework.Media
         {
             if (_sessionState == SessionState.Stopped)
                 return;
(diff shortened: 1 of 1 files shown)


tomspilman:
@TD25 - Does stop work correctly for you?

TD25:
Yes, although now I see that state is not set to stopped on stop.

tomspilman:
> Yes, although now I see that state is not set to stopped on stop.

Are we sure that isn't set by the threaded `MediaEngineExOnPlaybackEvent` callback?

TD25:
No, I attached to process set a breakpoint there and it didn't get hit. State gets set correctly in MediaPlayer.cs, that's why it works as expected I think. I should probably still set correct state in PlatformStop though.

tomspilman:
This is conflicted and for old UWP anyway.

PR #5096, opened 2016-08-09
Add manifest to templates to be dpi aware

### LINKED ISSUE #5040: SharpDX gets wrong resolution causing crash on startup
@tomspilman #5024 caused this issue (the commit before it worked fine). [`monitor.Description.DesktopBounds`](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs#L66) gets incorrect values (1536, 864) which is not a known displaymode for my graphics card, causing adapter._currentDisplayMode not to be set. `GraphicsDevice.DisplayMode` will be null and so [`GraphicsDevice.Setup`](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.cs#L218) will fail with a `NullReferenceException`.

Looks like this is an issue with SharpDX or maybe incorrect use of the API? I couldn't find this in SharpDX source, so maybe this was a bug that is fixed or API changed. I just updated my graphics drivers, but that didn't help.

### LINKED ISSUE #5092: Window is not centered on DesktopGL in HiDPI Windows environment
Confirmed on 125 % environment.

Jjagg:
@mrhelmut This fixes #5040 and should also fix #5092. A couple things:
- The manifest still needs to be set as ApplicationManifest in MD templates, but I don't know where that should go. It's just added to the project as a file now and nothing more. @dellis1972 Can you give me some pointers here?
- This only needs to be done for WindowsDX and DesktopGL right?
- Should the name be {NameOfProject}.manifest? I don't know how to replace the name of the file itself.

Sorry, I've never worked on templates before so there's a lot I don't know :/

### DIFF
--- a/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/app.manifest
+++ b/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/app.manifest
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
-  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
+  <assemblyIdentity version="1.0.0.0" name="${ProjectName}"/>
   <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
     <security>
       <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
--- a/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/app.manifest
+++ b/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/app.manifest
@@ -1,0 +1,42 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
+  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
+    <security>
+      <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
+        <requestedExecutionLevel  level="asInvoker" uiAccess="false" />
+      </requestedPrivileges>
+    </security>
+  </trustInfo>
+
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
+      <!-- A list of the Windows versions that this application has been tested on and is
+           is designed to work with. Uncomment the appropriate elements and Windows will 
+           automatically selected the most compatible environment. -->
+
+      <!-- Windows Vista -->
+      <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}" />
+
+      <!-- Windows 7 -->
+      <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}" />
+
+      <!-- Windows 8 -->
+      <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}" />
+
+      <!-- Windows 8.1 -->
+      <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}" />
+
+      <!-- Windows 10 -->
+      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
+
+    </application>
+  </compatibility>
+
+  <application xmlns="urn:schemas-microsoft-com:asm.v3">
+    <windowsSettings>
+      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
+    </windowsSettings>
+  </application>
+
+</assembly>
--- a/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/app.manifest
+++ b/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/app.manifest
@@ -35,7 +35,7 @@
 
   <application xmlns="urn:schemas-microsoft-com:asm.v3">
     <windowsSettings>
-      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
+      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true/pm</dpiAware>
     </windowsSettings>
   </application>
 
--- a/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/MonoGameProject.xpt.xml
+++ b/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/MonoGameProject.xpt.xml
@@ -28,7 +28,7 @@
 			<Files>
 				<File name="Game1.cs" src="Common/Game1.cs" />
 				<File name="Program.cs" src="Common/Program.cs" />
-				<RawFile name="app.manifest" src="Common/app.manifest" />
+				<File name="app.manifest" src="Common/app.manifest" />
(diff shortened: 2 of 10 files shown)


dellis1972:
I think this only effects DirectX based projects. I don't think OpenGL will know what to do with this. 

For MD the app.manifest would go in the common directory 'IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common'

It then needs to be added to MonoGameWindowsProject.xpt.xml 

'https://github.com/MonoGame/MonoGame/blob/develop/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/MonoGameWindowsProject.xpt.xml#L31'

And finally added to the MonoDevelop.MonoGame.addin.xml

'https://github.com/MonoGame/MonoGame/blob/develop/IDE/MonoDevelop/MonoDevelop.MonoGame/MonoDevelop.MonoGame.addin.xml#L24'

IDE/MonoDevelop/MonoDevelop.MonoGame/templates/MonoGameProject.xpt.xml
```diff
@@ -28,6 +28,7 @@
 			<Files>
 				<File name="Game1.cs" src="Common/Game1.cs" />
 				<File name="Program.cs" src="Common/Program.cs" />
+				<File name="app.manifest" src="Common/app.manifest" />
```
dellis1972: I don't think this is required as this is the Desktop GL project. Unless this manifest makes a difference on windows with OpenGL. (I'm not sure it does)

Jjagg: @mrhelmut asked for that. Not sure why exactly, but he looked into this more than I did.

Jjagg:
> And finally added to the MonoDevelop.MonoGame.addin.xml

Ah, I missed this one. But how do I set that the csproject must include?
<ApplicationManifest>app.manifest</AplicationManifest>
Thanks @dellis1972

IDE/MonoDevelop/MonoDevelop.MonoGame/templates/MonoGameProject.xpt.xml
mrhelmut: It does. :) It makes DX & GL projects behave the same in regard to DPI.

dellis1972:
does it only effect windows desktop? or Mac and Linux too? This all looks
windows specific to me.

Jjagg:
It is windows specific, but the manifest is ignored for Mac and Linux I think

Jjagg:
So how should I add `<ApplicationManifest>app.manifest</ApplicationManifest>` to the csproj in the MD templates so it's only there for DesktopGL and Windows projects?

I just tested the VS side of this and it's working :)

dellis1972:
You need to do something like

'
https://github.com/MonoGame/MonoGame/blob/develop/IDE/MonoDevelop/MonoDevelop.MonoGame/templates/MonoGameProject.xpt.xml#L37
'

note the 'BuildAction' setting.

for MonoGameProject.xpt.xml and MonoGameWindowsProject.xpt.xml

Jjagg:
The BuildAction should just be None, but the csproj needs to include the ApplicationManifest node. Is there documentation on this somewhere? I searched for it, but only found http://www.monodevelop.com/developers/articles/project-templates/ which did not help me with this :/

dellis1972:
Afaik you can't do it via a template.

I had to write code to fix up the various files [1]

[1]
https://github.com/dellis1972/MonoGame/blob/XS6/IDE/MonoDevelop/MonoDevelop.MonoGame/MonoGameMSBuildImports.cs#L132

Jjagg:
I'll take a look at that then

dellis1972:
Note that branch I have will work with Xamarin Studio 6. The current one
code in develop does not.

Jjagg:
Should I fix this in your branch or just here?

dellis1972:
my guess is we'll need to fix it in my branch. We won't be doing a release
without Xamarin Studio 6 support. We are just waiting on the Mac Build bot
being upgraded.

Jjagg:
I tested this in MD on Linux and it succesfully added the manifest. The project does not actually use the manifest at this point, but I'll leave this to @dellis1972 in his XS6 branch. I'm assuming the fact that this works with MD on Linux means it will work in XS on Mac? If that's true, this can be merged now :)

IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/app.manifest
```diff
@@ -0,0 +1,42 @@
...
+
+      <!-- Windows 10 -->
+      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
+
+    </application>
+  </compatibility>
+
+  <application xmlns="urn:schemas-microsoft-com:asm.v3">
+    <windowsSettings>
+      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
```
tomspilman: I was reading the following slides:

 video.ch9.ms/files/GDC2015-PCGamesforWindows10.pptx

They suggested all DirectX game should be set to "Per Monitor High DPI Aware".  Is this the correct setting for that?  I mean specifically the "per monitor" part.

Jjagg: No, it's system dpi aware. Should be "true/pm" for per monitor
https://msdn.microsoft.com/library/aa374191(v=VS.85).aspx
I'll get to this and the other PR's I have open in a couple days, maybe a week. I'm a bit busy atm

On 21 Aug 2016, at 00:31, "Tom Spilman" <notifications@github.com<mailto:notifications@github.com>> wrote:


In IDE/MonoDevelop/MonoDevelop.MonoGame/templates/Common/app.manifest<https://github.com/MonoGame/MonoGame/pull/5096#discussion_r75585797>:

> +
> +      
> +      <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}" />
> +
> +      
> +      <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}" />
> +
> +      
> +      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
> +
> +    </application>
> +  </compatibility>
> +
> +  <application xmlns="urn:schemas-microsoft-com:asm.v3">
> +    <windowsSettings>
> +      <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>


I was reading the following slides:

video.ch9.ms/files/GDC2015-PCGamesforWindows10.pptx

They suggested all DirectX game should be set to "Per Monitor High DPI Aware". Is this the correct setting for that? I mean specifically the "per monitor" part.

-

tomspilman: Thanks @Jjagg !

mrhelmut: Per monitor High-DPI awareness is only available since Windows 8.1.
It can only be achieved through the Win32 API using ```SetProcessDpiAwareness```, ```GetProcessDpiAwareness``` and ``` GetDpiForMonitor``` (which is only available on 8.1+). There's also a specific event message when DPI has changed (e.g. window has been moved to another monitor) ```WM_DPICHANGED```.

The manifest should work fine on 8.1+ with multi monitor configurations, with some limitations (the window will get the DPI of its starting monitor, but will keep its DPI if moved to another one).

Jjagg: From the link I poster above:
Contains "true/pm"

Windows Vista, Windows 7 and Windows 8:  The current process is system dpi aware.

Windows 8.1 and Windows 10:  The current process is per-monitor dpi aware.


Isn't that better?

KonajuGames:
Merging. Thanks @Jjagg

Issue #5066, opened 2016-08-03
apply3d soundeffect throws an exception


tchekjunior:
hi, 
there is a problem when trying to apply 3d  "SoundEffect".
the code bug with this message: An unhandled exception of type 'System.AccessViolationException' occurred in SharpDX.XAudio2.dll.

I read this discusion on the monogame forum: [http://community.monogame.net/t/solved-accessviolationexception-in-soundeffectinstance-apply3d-l-e/1980/3]

my ".wav" file in stereo. it would be very difficult to change that at the current level of my project.

Please if anyone has an idea to resolve this issue?
ps : I have lastest version on monogame on develop branch.

tchekjunior:
hi,
no body have idea how to fix that issue?

tomspilman:
Generally in most audio systems you cannot appl

tomspilman:
In many audio systems you cannot apply 3D positioning to stereo sounds...  but XNA suggests that it could by discarding the stereo data.

First thing... can you give me the full callstack of the crash?

tchekjunior:
hello Mr Spillman , thank you for your reply.
This is my "full callstack", I hope what help you understand the problem

![screen shot 08-09-16 at 04 50 pm](https://cloud.githubusercontent.com/assets/4943931/17523299/fbd9f70e-5e51-11e6-92f9-d7f08d5f46b3.PNG)

![screen shot 08-09-16 at 04 52 pm](https://cloud.githubusercontent.com/assets/4943931/17523367/3a712c76-5e52-11e6-8c7f-9e082a3a9110.PNG)

This zip contains my stereo sound .wav and .XNB
[sound pack](https://drive.google.com/file/d/0B9IHjQmx0e1fX0lYT0FFTDljblE/view?usp=sharing)

tomspilman:
Looks like a failure to call or within `X3DAudioCalculate`.
 
 https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.x3daudio.x3daudiocalculate(v=vs.85).aspx

This will take someone running this case thru the debugger and seeing specifically where the issue is coming from.  Could be we're just misusing the API here.

tchekjunior:
I do not know if it helps, but when I remove the flag "CalculateFlags.Matrix", the function will run very well, so in my opinion it is the calculated matrix that has a problem.

PR #5099, opened 2016-08-12
Fixed Apply3D() on stereo SoundEffect

tchekjunior:
hi,
I posed this problem lately [Issue 5066](https://github.com/MonoGame/MonoGame/issues/5066) I think I fix it this PR.
I read the doc SharpDX and directx, when a sound over one chanel,  value of the two variables in "emitter" must be supply : ChannelRadius; ChannelAzimuths;
I put default value zero.
https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.x3daudio.x3daudio_emitter(v=vs.85).aspx

thank you to Tom for his suggestion.
i hope to help you.


MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs
```diff
@@ -39,6 +39,12 @@ private void PlatformApply3D(AudioListener listener, AudioEmitter emitter)
             e.CurveDistanceScaler = SoundEffect.DistanceScale;
             e.DopplerScaler = SoundEffect.DopplerScale;
             e.ChannelCount = _effect._format.Channels;
+            //stereo channel
+            if(e.ChannelCount > 1 )
+            {
+                e.ChannelRadius = 0;
+                e.ChannelAzimuths = new float[] { 0f, 0f };
+            }
```
tomspilman: This looks like it could be the right path here, but we need some fixes.

First this PR should be rebuilt against a clean branch from `develop`.  There are two other non-related changes in this PR that need to be removed.

Second we should not be allocating a `float[]` over and over here.  We should allocate one static array and just assigning it here.

Third I am unsure if 0 is the correct values for the radius or azimuths.  It could be correct, it could be wrong.  We should do a little testing against XNA and make sure 0 here sounds correct to regular XNA behavior.

Other than that it seems like you're on the right track.

tchekjunior: cool, I think you are right for the PR, I'll clean it and also to the array, I did not want too impacted your work.
for the last point,"0" seems to be a good default value.
ChannelRadius is the distance from the Possition of imitter.
azimuths is value between 0 and 2pi, both value affects the least the nature of sound as in XNA.

tchekjunior:
I hope that is much better now.

tomspilman:
> sorry i can revert "Module.xml".

Please do.

tchekjunior:
ok, I do it. sorry i am beginner contributor.

tomspilman:
Since this will only affect `Apply3D()` on stereo sound effects which already did not work...  there is little danger in merging this without detailed testing.

Thanks @tchekjunior !

tchekjunior:
i welcome to help

Issue #5100, opened 2016-08-12
Calling convention incorrect for EffectsExtension in OpenGL (x86 WindowsGL)


RHY3756547:
Hey,

When trying to run a MonoGame application on the WindowsGL x86 platform, I get a PInvokeStackImbalance on an alGenFilters() call, and if I continue execution I get another one on alDeleteFilters():

![image](https://cloud.githubusercontent.com/assets/6294155/17632305/8448a6d4-60bf-11e6-8fda-201d468711a6.png)

I narrowed this down to the calling convention being incorrect for at least these two functions, but also likely all of the ones that are dynamically added via Marshal.GetDelegateForFunctionPointer. The main OpenAL functions declare their calling conventions as Cdecl, but those added via EffectsExtension() do not declare any calling convention at all.

To fix this, I just specified the calling convention on the delegates at OpenAL.cs:546.

```csharp
        /* Filter API */
        private unsafe delegate void alGenFiltersDelegate (int n, [Out] uint* filters);
        private delegate void alFilteriDelegate (uint fid, EfxFilteri param, int value);
        private delegate void alFilterfDelegate (uint fid, EfxFilterf param, float value);
        private unsafe delegate void alDeleteFiltersDelegate (int n, [In] uint* filters);
```
to
```csharp 
        /* Filter API */
        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
        private unsafe delegate void alGenFiltersDelegate (int n, [Out] uint* filters);
        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
        private delegate void alFilteriDelegate (uint fid, EfxFilteri param, int value);
        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
        private delegate void alFilterfDelegate (uint fid, EfxFilterf param, float value);
        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
        private unsafe delegate void alDeleteFiltersDelegate (int n, [In] uint* filters);
```

I don't have any other build configurations and I'm not sure if this is the correct way to fix it, so I haven't made a pull request. _Likely also need to do this with the Effect API above too,_ though my application doesn't seem to use it.

tomspilman:
@dellis1972  @cra0zy ?

dellis1972:
That looks about right.

dellis1972:
I'll do a PR for this.

Jjagg:
Fixed by #5101!

PR #5101, opened 2016-08-13
Fixed OpenAL CallingConvention on Function Pointers #5100

dellis1972:
The function pointers for OpenAL effects/filters were using
the platform default calling convention. On Mac and Linux
this is Cdecl, but on Windows it is StdCall.

OpenAL soft is built on windows using cdecl. As a result the
calls to these methods would either crash or produce unexpected
behaviour.


tomspilman:
Thanks @dellis1972 !

PR #5103, opened 2016-08-13
Reworked the SoundBank to return a new Cue each time

dellis1972:
XNA did not cache Cue's it returned a new one each time.


KonajuGames:
Merging.

nkast:
I am sure that was intentional
```
 -            // Note: In XNA this returns a new Cue instance, but that
 -            // generates garbage which is one reason to not do it.
```

Is there an actual reason why we should drop that?

tomspilman:
@nkast - Yeah it was... but at the same time the current behavior was causing bugs with incorrect playback.

I would like to revisit this in the future and try to remove this source of garbage, but i'm unsure if it can be done without a change to the XACT API.

It could be the best we can do is try to slim down the size of `Cue` and reduce references to it allowing it to maybe be a gen0 collection in most case.

PR #5110, opened 2016-08-15
Add missing XNA tests

Jjagg:
I missed some tests with the move to another folder in the XNA test project when doing #5072, this adds them back in.


KonajuGames:
Merging.  Thanks @Jjagg

PR #5112, opened 2016-08-15
Added SamplerState.TextureFilterMode to correctly support comparison filtering

Jjagg:
@tomspilman This adds a property to SamplerState so the user can explicitly set the TextureFilterMode to Compare and leaves room for more TextureFilterModes in the future. We no longer use CompareFunction.Never to indicate we don't want to use a CompareFunction. Instead it can be used like any other CompareFunction when SamplerState.FilterMode is set to TextureFilterMode.Compare. Discussion about this started in #5072

### DIFF
--- a/Build/Projects/2MGFX.definition
+++ b/Build/Projects/2MGFX.definition
@@ -106,6 +106,9 @@
     <Compile Include="..\..\MonoGame.Framework\Graphics\States\TextureFilter.cs">
       <Link>MonoGame.Framework\TextureFilter.cs</Link>
     </Compile>
+    <Compile Include="..\..\MonoGame.Framework\Graphics\States\TextureFilterMode.cs">
+      <Link>MonoGame.Framework\TextureFilterMode.cs</Link>
+    </Compile>
     <Compile Include="..\..\MonoGame.Framework\Graphics\Vertices\VertexElementUsage.cs">
       <Link>MonoGame.Framework\VertexElementUsage.cs</Link>
     </Compile>
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -814,6 +814,7 @@
     <Compile Include="Graphics\States\TargetBlendState.cs" />
     <Compile Include="Graphics\States\TextureAddressMode.cs" />
     <Compile Include="Graphics\States\TextureFilter.cs" />
+    <Compile Include="Graphics\States\TextureFilterMode.cs" />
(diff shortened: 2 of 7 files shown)


tomspilman:
This looks great @Jjagg ... solves our issue and looks like the right way to support different filter modes.

@tgjones - It would be good to get your feedback on this too since you did the original work for comparison filtering.

Jjagg:
>This looks great @Jjagg ... solves our issue and looks like the right way to support different filter modes.

Hey, you suggested it :D

tomspilman:
But you did the work... ideas are the easy part. :)

tgjones:
This looks like a good solution. I forget now why I used `CompareFunction.Never` as the `null` value. With hindsight, that's clearly wrong, so most likely I just didn't understand what `Never` actually does.

My only comment would be the name `TextureFilterMode`, which might be confused with `TextureFilter`, and I think this feature is more related to sampling than filtering. Perhaps `TextureSamplerMode`? But I don't feel strongly about it.

Jjagg:
> Perhaps TextureSamplerMode?

That sounds a lot better at least for the values we have now, but I don't know if that's true for other things we might add later on (don't know what other filters are possible)

tomspilman:
@tgjones @Jjagg 

Well the reason I suggested this was `TextureFilterMode` is because of the  `D3D11_FILTER` DirectX enumeration:

https://msdn.microsoft.com/en-us/library/windows/desktop/ff476132(v=vs.85).aspx

DirectX11/12 support the following filtering modes in addition to the default filtering behavior:

  - `D3D11_FILTER_COMPARISON_XXXXX`
  - `D3D11_FILTER_MINIMUM_XXXXX`
  - `D3D11_FILTER_MAXIMUM_XXXXX`

To me this is clearly a filtering feature as controls how the texels are combined to produce the final sample.

The "mode" bit came from OpenGL that calls this the `GL_TEXTURE_COMPARE_MODE`.  So why didn't I suggest `TextureComparisonMode`?  Because GL (and Vulkan apparently) does not support min/max filtering.  But there is an extension:

  https://www.opengl.org/registry/specs/EXT/texture_filter_minmax.txt

And it clearly indicates this is a filtering feature.

So IMO we should stick with `TextureFilterMode` as it is most accurate term for this.  I don't think it would cause much confusion and any that does arise should be handled by documentation.

Jjagg:
@tomspilman that makes sense. IMO this is good to be merged then :)

tomspilman:
Thanks @Jjagg !

PR #2382, opened 2014-04-20
Add GraphicsDevice.GetBackBufferData and implementations for GL/DirectX

danzel:
This allows taking screenshots of your running game!

At the moment the implementation only supports byte[] data and returns the data as RGBA/BGRA 32bit, it could be extended to support more formats.
The format of the pixels differs by platform. XNA and WindowsGL return it in BGRA, WindowsDX, iOS, Android returns it in RGBA.

It should be possible to get WindowsDX to return data in the same format, but this was my first time doing DirectX programming :) (Doing a cpu side swap just so user code can undo it seems wrong)
Android and iOS image code handles RGBA data, so they aren't really an issue.

Only the simplest version of this function is implemented, there are two others which have not been included:
http://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.graphicsdevice.getbackbufferdata.aspx

Tested on iOS, Android, Windows(GL/DX).

I have an example for WindowsDX, WindowsGL, iOS and Android using this here:
https://github.com/danzel/WrapTest/blob/master/WrapTest/WrapTest/Game1.cs#L199-L250
It might be nice to have this example code somewhere, any suggestions?

Fixes #1885

PR #5114, opened 2016-08-16
Implement GraphicsDevice.GetBackBufferData

Jjagg:
I hijacked #2382 from @danzel. Thanks again!
There's a test in here too, it works if I chang the default backbuffer format by removing the extra checks where this is set, I'll add a line comment there. I manually tested the OpenGL implementation and that seems to be working too :) I noticed we don't do anything with PreferredBackBufferFormat there though.
This is not finished yet, I'd like to add some tests for parameter validation.


Jjagg:
> I'll add a line comment there

Looks like I can't because it's too far away from the diff. 

The thing is at https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs#L694 we set surface format of the backbuffer to BGRA when PresentationParams have SurfaceFormat.Color in DirectX. Why? Because it is the most common format or something? 
```
var format = PresentationParameters.BackBufferFormat == SurfaceFormat.Color ?
                            SharpDX.DXGI.Format.B8G8R8A8_UNorm :
                            SharpDXHelper.ToFormat(PresentationParameters.BackBufferFormat);
```

If I remove the extra check here I don't have to do any special swapping when retrieving backbuffer data to match XNA behavior. I'm guessing this might break something, since it looks like the check is done for a good reason

Jjagg:
I only removed the check mentioned above for regular Windows DX projects (not UWP or Store App). That makes the backbuffer at least be in the expected format. Would be nice if anyone can test their Windows DX game with this to make sure nothing weird happens. Tests don't catch anything at least.

Jjagg:
Looks like I got the build passing, but there are some issues left here. I have no idea if this works for Windows Phone and GL platforms do something different than DX for now so one of the two is wrong. I think GL because DX produces the same result as XNA and I don't think they implemented this incorrectly. Will research and post back/resolve.

Jjagg:
I'm stupid. The image is flipped on OpenGL, so I reversed the data buffer rows, but I didn't make the ReadPixels call take the y parameters flipped. Posting here so I don't forget, will fix in a couple hours probs.

Jjagg:
By the way, when this is implemented the GraphicsDeviceTestBaseFixture can use this and allow rendering to the backbuffer in stead of a custom render target.

tomspilman:
> GraphicsDeviceTestBaseFixture can use this and allow rendering to the backbuffer in stead of a custom render target. 

True.... and it would test this path pretty well in the process.  I don't have a problem with changing things over to this once it is all working.

Jjagg:
On second thought maybe that's not a great idea because the Game class is not initialized in the setup right now, so there is no backbuffer. Basically it does the minimal amount of work to get a graphicsdevice to draw stuff with.

Though maybe we can extend the fixture base in case we need tests that write to the backbuffer explicitly or change render targets a lot.

Jjagg:
There's an MSAA test in GraphicsDeviceManagerTest that calls GetBackBufferData in XNA but doesn't do the check for MG, so at least that one can be improved

Jjagg:
Fixed the OpenGL implementation. At least WindowsDx and OpenGL are working with this. I'm not sure about Windows Phone and UWP.

tomspilman:
> I'm not sure about Windows Phone and UWP.

That will have to depend on manual testing for now.  My hope is that once we get unit tests running on OpenGL we can push to run unit tests of more platforms including UWP, iOS, Android, and even our console platforms.  

As you can imagine getting that setup and maintained for a little open source project like ours is tricky.

Jjagg:
Yeah, I get that. At least OpenGL should be doable for now, since that can just run on the build server as is, right? 

I almost forgot, the backbuffer format when the preferred format is Color is still set to BGRA for UWP and WP.

Jjagg:
I'm thinking it would be useful to expose the current backbuffer format. I don't think XNA does, but it's useful for this. Note that the PreferredBackBufferFormat is not necessarily the actual format. I read (I think on Shawn Hargreaves blog, but maybe just the docs) that XNA tries to find a format that is close to the one you set as preferred if that one is not available. This would probably require a lot of testing/tinkering to get right.

Come to think of it, I didn't check if XNA returns the color data in the current back buffer format or always the Color format. I'd guess the current format.

Jjagg:
> Come to think of it, I didn't check if XNA returns the color data in the current back buffer format or always the Color format. I'd guess the current format.

I tried to write a test for this, but XNA always forces the backbuffer format to Color. I didn't look into this very much, but I'm guessing it's because of hardware limitations, since in the [XNA docs](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.presentationparameters.backbufferformat.aspx) it says

> For windowed applications, the back-buffer format no longer needs to match the display-mode format because color conversion can now be done by the hardware (**if the hardware supports color conversion**). The **set of possible back-buffer formats is constrained**, but the runtime will allow any valid back-buffer format to be presented to any desktop format.

SharpDX crashes for MG when I try to create the backbuffer with any format that's not Color.

@KonajuGames Does any platform use another format than Color by default? If not this is probably not worth worrying about since 99.99% of devs will only ever use the Color format for the backbuffer.

KonajuGames:
I know Windows Phone 7 would usually force a back buffer of BGR565 because
that is what the hardware supported. Many years ago (20 or so), 16-bit high
color desktops were common, but these days I haven't seen one of those for
a long time.

tomspilman:
@Jjagg - This one is conflicted too and needs a merge/rebase.

Jjagg:
The status of this is still the same. UWP and WP are untested but DesktopGL and WinDX are working (manual test for DesktopGL and automated for DX).

Jjagg:
@tomspilman Looks to be a build bot issue. I can build fine locally (on Linux though).

tomspilman:
Yeah something is failing in the build process for Linux.  @dellis1972 @cra0zy ?

cra0zy:
Try restarting the build, considering there was a stack trace it might have been a one time build fail...

tomspilman:
> Try restarting the build, considering there was a stack trace it might have been a one time build fail

You were right @cra0zy ... a rebuild and it worked fine.

tomspilman:
@Jjagg - Looks like the tests for `GetBackBufferData` are failing.

Jjagg:
The Linux build fails in a place that's not touched in this PR, looks like something got messed up on the build bot maybe:
```
Styles.Linux.cs(115,98): error CS1061: Type `object' does not contain a definition for `Handle' and no extension method `Handle' of type `object' could be found. Are you missing an assembly reference?
```
Any ideas, @cra0zy?

cra0zy:
> Any ideas, @cra0zy?

Update `Dependencies` submodule, seems like you are requesting some outdated version of it in this PR.

Jjagg:
Oh, thanks! Probably messed up a merge  or something.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -1477,6 +1473,82 @@ private void PlatformDrawInstancedPrimitives(PrimitiveType primitiveType, int ba
             }
         }
 
+        private void PlatformGetBackBufferData<T>(Rectangle rect, T[] data, int startIndex, int count) where T : struct
```
tomspilman: @Jjagg - Would it be possible to move the generic work outside this platform call?  Like could the `Marshal.SizeOf(typeof(T))` be moved to the caller and `T[] data` be converted to `byte[]`, `byte*`, or `IntPtr`?

I'm thinking from the perspective of code bloat in this method (also in the `GetData/SetData` methods in textures, vertex, and index buffers).  This is something I've noticed in the cross compiler work I've been doing... lots of template code for stuff that is largely not generic.

Either way... this isn't critical for this PR, but I would like to see about reducing the amount of generic code we generate.

Jjagg: Hmm, I don't think that would be convenient for this part of the codebase. 
Some thoughts:
- For DirectX SharpDX provides functionality to read T's from the unmanaged memory. For OpenGL we just pass an IntPtr to the start of the managed T array and the copying/marshalling is handled by PInvoke/interop code. That's both pretty convenient and by making it more generic we'd need just as much or more code for both targets.
- If we return a byte array we'd need two copy operations, one to go from unmanaged (IntPtr) to managed memory (byte[]) and another for marshalling the byte array to a T array. Though I recall reading about the last part being done without copying with some very hacky, unsafe code. 
- In DX we can't return an IntPtr directly because of the rowPitch and because the SharpDX DataStream should be disposed inside the platform specific call. GL would be fine with an IntPtr. We do need to flip the rows of the returned data, so that would probably be done in an unsafe block to keep things cheap.

I haven't dealt with C# marshalling/interop outside of stuff I did for MG, so it's probably good to let some more experienced people look into this.

Jjagg:
I think this is good to go like this :)

Jjagg:
There was an issue when the back buffer had MSAA enabled, so I fixed that and made the MSAA test in GraphicsDeviceManagerTest use GetBackBufferData for verification like the XNA test did.
This is good to go IMO!

Jjagg:
Wow, this is getting old. Gonna try and fix this up soon so we can get it merged.

Jjagg:
This needs to be brought up to date by removing the Windows Phone parts. I'll do that ASAP, hopefully we can get this merged then :)

tomspilman:
Thanks @Jjagg ... merging now!

PR #5117, opened 2016-08-16
Optimized and added additional constructors to Color

### LINKED ISSUE #4824: Add Color(uint) and Color(byte,byte,byte,byte) Constructors
When you have to initialize a bunch of colors the current setters and constructors are inefficient.  They all do internal math to scale or clamp values to back to a byte then pack them.  Often this math isn't needed but is there just in case someone passes an out of range value.

I suggest extending the API to support two additional constructors directly taking a packed value and channel bytes:
```
public Color(uint packedValue)
{
...
}

public Color(byte red, byte green, byte blue, byte alpha)
{
...
}
```
These would be the most efficient constructors one could use.

We should also document the packed format in the reference docs on the new ctor and the packed value getter/setter.

polsys:
Fixes #4824. Added the `Color(byte, byte, byte, byte)` constructor and made `Color(uint)` public. Also added some simple unit tests for the constructors and corrected a minor typo.

In the issue, there was some discussion about a separate `Color.FromBytes()` method, but I opted for another constructor overload, since the parameters must be `byte`s in either case. I did not add a `Color(byte, byte, byte)` overload with constant alpha to avoid excessive overloads - this is a more "advanced" overload after all. This can be changed if it makes the API cleaner, of course.

As expected, the new `byte` overload is clearly (40-60%) faster than the `int` version in a quick microbenchmark.

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -162,7 +162,10 @@ namespace Microsoft.Xna.Framework
             Yellow = new Color(0xff00ffff);
             YellowGreen = new Color(0xff32cd9a);
         }
-	// ARGB
+
+        // Stored as RGBA with R in the least significant octet:
+        // |-------|-------|-------|-------
+        // A       B       G       R
         private uint _packedValue;
 	  
         /// <summary>
@@ -174,10 +177,6 @@ namespace Microsoft.Xna.Framework
         public Color(uint packedValue)
         {
             _packedValue = packedValue;
-			// ARGB
-			//_packedValue = (packedValue << 8) | ((packedValue & 0xff000000) >> 24);
-			// ABGR			
-			//_packedValue = (packedValue & 0xff00ff00) | ((packedValue & 0x000000ff) << 16) | ((packedValue & 0x00ff0000) >> 16);
         }
 
         /// <summary>
@@ -185,13 +184,8 @@ namespace Microsoft.Xna.Framework
         /// </summary>
         /// <param name="color">A <see cref="Vector4"/> representing color.</param>
         public Color(Vector4 color)
+            : this((int)(color.X * 255), (int)(color.Y * 255), (int)(color.Z * 255), (int)(color.W * 255))
         {
-            _packedValue = 0;
-			
-			R = (byte)MathHelper.Clamp(color.X * 255, Byte.MinValue, Byte.MaxValue);
-            G = (byte)MathHelper.Clamp(color.Y * 255, Byte.MinValue, Byte.MaxValue);
-            B = (byte)MathHelper.Clamp(color.Z * 255, Byte.MinValue, Byte.MaxValue);
-            A = (byte)MathHelper.Clamp(color.W * 255, Byte.MinValue, Byte.MaxValue);
         }
 
         /// <summary>
@@ -199,13 +193,8 @@ namespace Microsoft.Xna.Framework
         /// </summary>
         /// <param name="color">A <see cref="Vector3"/> representing color.</param>
         public Color(Vector3 color)
+            : this((int)(color.X * 255), (int)(color.Y * 255), (int)(color.Z * 255))
         {
-            _packedValue = 0;
-
-            R = (byte)MathHelper.Clamp(color.X * 255, Byte.MinValue, Byte.MaxValue);
-            G = (byte)MathHelper.Clamp(color.Y * 255, Byte.MinValue, Byte.MaxValue);
-            B = (byte)MathHelper.Clamp(color.Z * 255, Byte.MinValue, Byte.MaxValue);
-            A = 255;
         }
 
         /// <summary>
@@ -215,12 +204,16 @@ namespace Microsoft.Xna.Framework
         /// <param name="alpha">The alpha component value from 0 to 255.</param>
         public Color(Color color, int alpha)
         {
-            _packedValue = 0;
+            if ((alpha & 0xFFFFFF00) != 0)
+            {
+                var clampedA = (uint)MathHelper.Clamp(alpha, Byte.MinValue, Byte.MaxValue);
 
-            R = color.R;
-            G = color.G;
-            B = color.B;
-            A = (byte)MathHelper.Clamp(alpha, Byte.MinValue, Byte.MaxValue);
+                _packedValue = (color._packedValue & 0x00FFFFFF) | (clampedA << 24);
+            }
+            else
+            {
+                _packedValue = (color._packedValue & 0x00FFFFFF) | ((uint)alpha << 24);
+            }
         }
 
         /// <summary>
@@ -228,14 +221,9 @@ namespace Microsoft.Xna.Framework
         /// </summary>
         /// <param name="color">A <see cref="Color"/> for RGB values of new <see cref="Color"/> instance.</param>
         /// <param name="alpha">Alpha component value from 0.0f to 1.0f.</param>
-        public Color(Color color, float alpha)
+        public Color(Color color, float alpha):
+            this(color, (int)(alpha * 255))
         {
-            _packedValue = 0;
-
-            R = color.R;
-            G = color.G;
-            B = color.B;
-            A = (byte)MathHelper.Clamp(alpha * 255, Byte.MinValue, Byte.MaxValue);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


polsys:
Implemented the short-circuiting of clamping per the discussion. This improved the `int` overload performance a lot (almost halved the time), though the `byte` overload is still slightly (about 20% in my tests) faster. Also added some clamping tests which I had missed.

MichaelDePiazzi:
Looks good! :thumbsup:
Just a thought while these optimisations are being done - Would it maybe also be worthwhile setting `_packedvalue` in 1 hit, rather than setting each property individually (which performs additional unnecessary operations)? For example:
```c#
public Color(byte r, byte g, byte b, byte alpha)
{
    _packedValue = CalculatePackedValue(r, g, b, alpha);
}

private static uint CalculatePackedValue(byte r, byte g, byte b, byte a)
{
    return r | ((uint)g << 8) | ((uint)b << 16) | ((uint)a << 24);
}
```

MonoGame.Framework/Color.cs
```diff
@@ -273,14 +288,43 @@ public Color(int r, int g, int b)
...
+        /// Constructs an RGBA color from scalars representing red, green, blue and alpha values.
+        /// </summary>
+        /// <remarks>
+        /// This overload sets the values directly without clamping, and may therefore be faster than the other overloads.
+        /// </remarks>
+        /// <param name="r"></param>
+        /// <param name="g"></param>
+        /// <param name="b"></param>
+        /// <param name="alpha"></param>
+        public Color(byte r, byte g, byte b, byte alpha)
```
tomspilman: So isn't there an issue with both a `byte` and `float` constructors in this type?

https://blogs.msdn.microsoft.com/shawnhar/2010/03/31/color-constructors-in-xna-game-studio-4-0/

MonoGame.Framework/Color.cs
```diff
@@ -273,14 +288,43 @@ public Color(int r, int g, int b)
...
+        {
+            _packedValue = 0;
+            R = r;
+            G = g;
+            B = b;
+            A = alpha;
```
tomspilman: I recommend directly packing the value here and avoid the overhead of 4 property assignments that may or may not be inlined depending on platform and build settings.

tomspilman:
@MichaelDePiazzi - I suggest an even more direct approach:
```
public Color(byte r, byte g, byte b, byte alpha)
{
    _packedValue = r | ((uint)g << 8) | ((uint)b << 16) | ((uint)alpha<< 24);
}
```
There is little benefit in a shared packing function... it may or may not inline depending on the platform and build settings...  and it isn't like we will be changing our packing format anytime soon.

In the end bugs are reduced by the unit tests in this type... sharing more code here just makes it potentially slower for no real gain.

MichaelDePiazzi:
@tomspilman While from a clean code point of view this makes me shudder a little, from a performance point of view I have to agree with your points. And I also agree with you about unit tests. Having good test coverage is most beneficial for helping reduce bugs and regressions.

By the way, here are some benchmarks I did to get an idea of how much faster setting the packed value directly is. (Measuring time taken to create 20 million `Color`'s)

- Original `int` constructor = **3174ms**
- New optimised `int` constructor = **1802ms** _(43% less time, 76% faster)_
- New `byte` constructor = **1512ms** _(52% less time, 110% faster)_
- New optimised `int` constructor using direct set = **1203ms** _(62% less time, 164% faster)_
- New `byte` constructor using direct set = **760ms** _(76% less time, 318% faster)_

tomspilman:
> While from a clean code point of view this makes me shudder a little

I agree...  I hate seeing code duplicated like this very much, but in this sort of type as well as the vector classes performance is first priority.  The only reason I can tolerate it is knowing the unit tests will alert us to any bugs.

MichaelDePiazzi:
While we're on the subject of optimisations, the `Color` comparison operators could also potentially be improved.

Current code:
```c#
public static bool operator ==(Color a, Color b)
{
    return (a.A == b.A &&
        a.R == b.R &&
        a.G == b.G &&
        a.B == b.B);
}

public static bool operator !=(Color a, Color b)
{
    return !(a == b);
}
```
My suggestion:
```c#
public static bool operator ==(Color a, Color b)
{
    return (a.PackedValue == b.PackedValue);
}

public static bool operator !=(Color a, Color b)
{
    return (a.PackedValue != b.PackedValue);
}
```
Quick benchmark on the equals operator shows it to be ~150% faster.

tomspilman:
> the  Color  comparison operators could also potentially be improved.

Absolutely!  It is dumb of us to not be taking advantage of direct access to `_packedValue`.  We should prefer that over the `PackedValue` property and certainly the individual component properties.

MichaelDePiazzi:
> We should prefer that over the PackedValue property

I thought that also... but for some strange reason, it's slower for me when comparing `_packedValue` directly!? I have no idea why either... :confused:

My benchmarks: (Measuring time taken to do 20 million equals comparisons of equal `Color`'s)
- Original (individual property comparison) = **954ms**
- `PackedValue` compare = **447ms**
- `_packedValue` compare = **708ms**

Could it be something to do with the `[CLSCompliant(false)]` tag on the `PackedValue` property?

polsys:
I completely agree, the optimizations are worth duplicating some code. Working on it right now...

MonoGame.Framework/Color.cs
polsys: Do you mean this quote?

> When you pass integer values to a method that has byte and float overloads, C# method resolution rules choose the float version, because int -> float is considered a better conversion than int -> byte.

That is not a problem because we have the `int` overload, which is chosen by overload resolution since it has no conversions at all. The only way to get the `byte` or `float` overload is explicitly having all the parameters as the respective type (either by variable definition or casting).

tomspilman:
>  it's slower for me when comparing `_packedValue` directly!? 

I suggest checking the JIT optimized assembly code directly to see why.

MonoGame.Framework/Color.cs
tomspilman: Ah...  I see what you mean.  I wonder why the XNA team didn't just leave the `byte` ctor when adding the `int` ctor then.

KonajuGames: I think it will only use the byte ctor when the parameters are explicitly bytes.  Any operation without casts will expand to int.

tomspilman: You are suggesting that `new Color(128,128,128,128)` will always call the `int` version of the constructor.  This would mean you would have to either put the values into a temporary or explicitly cast like `new Color((byte)128,(byte)128,(byte)128,(byte)128)`.

If this is correct... and I think it is... should we instead consider ` Color.FromBytes(byte,byte,byte,byte)` for this optimized construction method?

```

```

MichaelDePiazzi:
> I suggest checking the JIT optimized assembly code directly to see why.

Hmmm... seems like the property version is being optimised to do a direct compare:
```
019A65C0  cmp         ecx,edx  
019A65C2  sete        al  
019A65C5  movzx       eax,al  
019A65C8  ret  
```
Whereas the field version is not for some reason:
```
019A65E0  push        ebp  
019A65E1  mov         ebp,esp  
019A65E3  sub         esp,8  
019A65E6  mov         dword ptr [ebp-4],ecx  
019A65E9  mov         dword ptr [ebp-8],edx  
019A65EC  mov         eax,dword ptr [ebp-4]  
019A65EF  cmp         eax,dword ptr [ebp-8]  
019A65F2  sete        al  
019A65F5  movzx       eax,al  
019A65F8  mov         esp,ebp  
019A65FA  pop         ebp  
019A65FB  ret  
```
No idea why...? :confused:

MonoGame.Framework/Color.cs
polsys: The `Color.FromBytes` version has the advantage that it will never fall back to the `int` implementation, but rather raises a compiler error. On the other hand, it adds additional API surface, and with the other optimizations here the performance advantage might actually be quite small.

KonajuGames: Test it to be sure, but I know C# likes to expand operands to int where
possible.

MichaelDePiazzi:
Update: If I run it in 64-bit mode, it optimises the field version the same as it does for the property version. So for some odd reason, in 32-bit mode the field comparison is not optimised.

polsys:
After the optimizations, the constructors run about 2-3 times faster. The `float` and vector constructors just offload work to the `int` version, and are still a lot faster. The equality operators also benefited nicely. I did use `_packedValue` for consistency until we have hard data on the JIT behavior - if that is necessary.

I also noted an inconsistency in our `FromNonPremultiplied` behavior. Added unit tests and changed to match XNA.

KonajuGames:
> The float and vector constructors just offload work to the int version

How does that work? The int version takes a range of 0-255 and the float
version takes 0.0-1.0.

polsys:
> How does that work? The int version takes a range of 0-255 and the float
version takes 0.0-1.0.

Just deduplicating the old code:

```
public Color(float r, float g, float b)
    : this((int)(r * 255), (int)(g * 255), (int)(b * 255)) 
{
}
```

tomspilman:
This good to merge?  @KonajuGames @dellis1972 @MichaelDePiazzi @polsys ?

KonajuGames:
I think so.

MichaelDePiazzi:
@tomspilman Looks good to me :+1:

tomspilman:
Great.... merging!

Issue #4939, opened 2016-06-15
Fix SpriteEffect half pixel offset


tomspilman:
Now that we've made `SpriteEffect` public we have to make sure it works.

Currently you can see it is always applying a half pixel offset... even on platforms that don't require it:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/SpriteEffect.cs#L62

It should use the `SpriteBatch.NeedsHalfPixelOffset` flag like `SpriteBatch` does:
 https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/SpriteBatch.cs#L140

While we're at it we should add in a unit test for it since we need to ensure it works now.

Verser:
Just checking, this seems to be complete but is marked as Open and HelpWanted. Should this be closed?

tomspilman:
@Verser - Thanks... we just forgot to close this.

PR #5118, opened 2016-08-16
Fix SpriteEffect half pixel offset

krzychu2510:
Fixed the issue #4939 when the NeedsHalfPixelOffset tag was not being used in the SpriteEffect class.


tomspilman:
Thanks @krzychu2510 .... this indeed is more correct.

Issue #5119, opened 2016-08-16
SetData/GetData<T>


Jjagg:
When starting to fix #1455 I thought it might be a good idea to do parameter validation for GetData and SetData in a seperate method so we do the same checks in both, don't have duplicate code here and have a better overview of it. Maybe even abstract some of the checks up to Texture itself since this code can be shared for the most part and it'll make the checks consistent for all Texture types. 

I just realized the same methods exist for Vertex/IndexBuffers, not sure if we should share all around or just do textures. Thoughts on this?

Also, there are cases the checks miss because sizeof(T) was is always expected to be equal to 4 (which is expected to be the size of the format, but that isn't always the case either) and the size of T is not explicitly checked while in XNA it is. I'm happy to fix all this once we decide on the above paragraphs.

tomspilman:
> it might be a good idea to do parameter validation for GetData and SetData in a seperate method so we do the same checks in both

We generally avoid this because it sometimes forces us to make parameter validation that doesn't exactly match XNA behavior or limits our ability to provide helpful messages.  Also we hate introducing the extra function call.

That said if SetData/GetData tests are really identical it could be something to consider.  Generally very few SetData/GetData calls are done per-frame... usually much less than draw calls, setting effect parameters, etc....  so an extra non-virtual function call would be a very small penalty if we want to reduce our code size a bit.

> not sure if we should share all around 

I'm not sure if that would be really beneficial at this time.  I expect the tests and messages are quite different.

It is unfortunate XNA wasn't designed with a base `GraphicsBuffer` type that could share some of this common stuff.  DirectX 10/11/12 all consider index/vertex/texture/constant "buffers" and share set/get data methods....  but XNA was initially designed around DX9 where this wasn't true.

Jjagg:
> Generally very few SetData/GetData calls are done per-frame...

I thought about this too. These functions are pretty slow themselves, so I think the extra overhead is negligible.

> I'm not sure if that would be really beneficial at this time. I expect the tests and messages are quite different.

Yeah, you're right... doesn't really make sense the way MG is set up. Would be weird to have some internal static method somewhere.

> It is unfortunate XNA wasn't designed with a base GraphicsBuffer type that could share some of this common stuff.

Maybe we can do something like this internally. I'll think about this for a bit. Not sure how well this maps to DX/OpenGL though and it seems like a lot of work for little gain. Still it would be nice to clean up all the duplication.

tomspilman:
[...]

PR #5121, opened 2016-08-17
Improve Texture.GetData and SetData parameter validation

### LINKED ISSUE #1455: Add parameter validation to SetData() of Texture2D/Texture3D/TextureCube

### LINKED ISSUE #4276: Texture2D.GetData buggy for DXT textures
I use MonoGame 3.4 on Windows (Windows DirectX project) with Visual Studio 2015.
I have this texture into DXT1 format:
t

When I use this code:

``` c#
        Texture2D texture;
        SpriteBatch spriteBatch;

        {

            // Load Dxt1 texture (256x256).
            texture = Content.Load<Texture2D>("t");
            byte[] data = new byte[texture.Width * texture.Height / 2]; // Dxt1 correct data size
            texture.GetData(data);
            texture.SetData(data); // Now texture is corrupted.
        }

        {
            spriteBatch.Begin();

            spriteBatch.Draw(texture, Vector2.Zero, Color.White);

            spriteBatch.End();
            base.Draw(gameTime);
        }
```

I get this result:
t2

But, the same code with the texture into Color format (ARGB8) work fine.
I have also tried with DXT3 and DXT5 format and it's the same result as DXT1: texture is corrupted.

A previous issue thread https://github.com/mono/MonoGame/issues/3590 suggest to change the type of struct to get texture data to avoid troubles, but no changes for DXT textures this time.

### LINKED ISSUE #5172: Missing TextureCube.GetData(...) method
The `TextureCube.GetData(CubeMapFace cubeMapFace, int level, Nullable<Rectangle> rect, T[] data, int startIndex, int elementCount)` method is missing (ref: https://msdn.microsoft.com/en-us/library/bb197085.aspx).

Jjagg:
This fixes the issues mentioned in #5119 and fixes #1455. Only Texture2D is thoroughly tested for now. I also implemented TextureCube.GetData overloads that were missing.

Something that not everyone might agree with is I changed back the validation of elementCount as it was in XNA. We changed this so that when elementCount >= data.Width \* data.Height the functions worked, while in XNA they check elementCount == data.Width \* data.Height. I changed this back because I think this is ambiguous. You could argue we're more flexible than XNA if we allow elementCount to be larger, but in fact elementCount itself is not used, so allowing arbitrary values might confuse the user. 

The fact that elementCount is even in there is a bit weird in the API... It's really only used for validation. Maybe when we start moving away from XNA we can break this by removing it and make it optional for a while (default value of 0 and remove the check) to ease the transition. Or just overload and obsolete the old calls. IMO it makes a lot more sense to have `GetData<T>(T[] data)`, `GetData<T>(Rectangle rect, T[] data)`, `GetData<T>(Rectangle rect, T[] data, int startIndex)` and overloads as needed for mip level, arraySlice etc.

Just a quick snippet that points out why I think elementCount is misplaced (named params for your convenience):

```
var tex = new Texture2D(graphicsDevice: gd, width: 20, height: 1);
var buffer = new Color[10];
tex.GetData(data: buffer, startIndex: 0, elementCount: 10);
```

This fails in XNA (and MG) with 

```
System.ArgumentException : The size of the data passed in is too large or too small for this resource.
```

Fixes #5172 
Fixes #4276


Jjagg:
Woops, didn't do OpenGL yet for TextureCube.PlatformGetData

Jjagg:
The failing tests are because of the change in the elementCount check. When reading a texture in [SpriteFontReader] (https://github.com/MonoGame/MonoGame/blob/7d90b229a70f1325585e15c712e9cd7db761f31e/MonoGame.Framework/Content/ContentReaders/SpriteFontReader.cs#L43) in the Texture2DReader [levelDataSizeInBytes](https://github.com/MonoGame/MonoGame/blob/ab5d3f1fa0d1012f8d3d73eb8134d03c048f44f6/MonoGame.Framework/Content/ContentReaders/Texture2DReader.cs#L74) is equal to the amount of pixels, but the surfaceformat is Dxt3 which means pixels have a bytesize of 16, hence texture size in bytes (128\*128\*16 for the test I debugged) and elementCount (128\*128\*1) don't match up. I'm not sure where the error is exactly because the texture is loaded correctly, so the number of bytes read and format should be correct.

KonajuGames:
It's 16 bytes for a 4x4 block of pixels, so a 128x128 texture occupies (128
/ 4) x (128 / 4) x 16 bytes.

Allowing a larger buffer permits re-use of memory buffers that are
potentially larger than the required size. We use scratch buffers in other
areas when loading content in order to minimise garbage generation.  This
can save a sizable amount of time and memory when loading large amounts of
content.

Jjagg:
> It's 16 bytes for a 4x4 block of pixels

Ah, now it makes sense. I thought size of a surface format was always number of bytes per pixel. Though admittedly 16 bytes per pixel would be overkill, especially for fonts :p
Doesn't it make more sense to implement the SurfaceFormay.GetSize extension method to return byte size per pixel though? Or do we actually use the fact that this is 16 for Dxt3 and similar for other formats that group pixels like this? Alternatively we can add another extension method for bytes per pixel. 

I figured out the scratchbuffer stuff, there's no issue there.

KonajuGames:
Bytes per pixel does not make sense for compressed textures. In this case,
it happens to be 1 byte per pixel. For DXT1 and some other compressed
textures it would be 0.5 bytes per pixel. Think of GetSize() as
GetBlockSize(), the size of the smallest addressable unit.

Jjagg:
> Bytes per pixel does not make sense for compressed textures.

If we want to do the element count check right we need to be able to get the number of bytes for a region of the texture. We could just drop the check of course. That would make the elementCount param unused.

tomspilman:
In general most graphics systems think of textures in "bits per pixel" and not "bytes per pixel" to solve this issue.

Jjagg:
So just add another method then? Sounds good to me :)

Jjagg:
There's something annoying going on here... 

Say you have a texture and it's in a compressed format that has 4 bits per pixel and it's size is just 1x1. Now if you call GetData on that whole texture your `elementCount * Marshal.SizeOf(typeof(T))` should be equal to `(width * height * bitsPerPixel)/8` (here we divide by eight to get to bytes instead of bits) so bytes expected and actual nr of bytes are equal. However `(1 * 1 * 4) / 8 = 0.5` which will be 0 for integer division. More generally you won't be able to get an uneven amount of pixels if the surface format has < 8 bits per pixel. I'm not sure we should allow fetching data when surface format size is less than 1 byte. It's all a bit weird when you're working with half bytes because you have to do bit shifting to fit two pixels and we also need to decompress these things (for formats we can render compressed and which are stored compressed). I'm pretty sure GetData and SetData don't work for these formats right now either way, so maybe we should just check the bitsize of the format and throw if each pixel cannot be represented in N bytes (so if `Format.GetBitsPerPixel() < 8 || Format.GetBitsPerPixel() % 8 ~= 0`) in Get/SetData. Thoughts?

KonajuGames:
For purposes of byte sizes, the dimensions are rounded up to the next
multiple of four. So a 1x1 compressed texture occupies the same memory as a
3x2 or a 4x4.

Jjagg:
@KonajuGames So what would be the expected/desired/current behavior in the case I described in my previous comment? Return more than what the user is asking for? That doesn't play nice with elementCount and is not what the user expects. Decompress and return like normal? We probably can't do this in all cases.

If we treat the size of a single stray pixel as the size of a minimal block, what happens if the user requests just one pixel and passes an array just large enough for a single pixel (less than the size of a block)? If we only allow this when the array is large enough for a block this should be properly documented IMO.

tomspilman:
> So what would be the expected/desired/current behavior 

The answer is "whatever XNA does for DXT formats"...  assuming that XNA doesn't do something completely insane.

Jjagg:
I should've known that would be the answer :p Will write some tests for this :)

Jjagg:
SharpDX is throwing when I try to create a new RT with DXT surfaceformat. Either we're doing something wrong using the API or it doesn't allow that

KonajuGames:
You can't have a compressed format as a render target. The GPU doesn't do
compression on the fly.

For tests you'll have to user some pre-compressed data. Use a small texture
processed by the content pipeline and get all mip levels. They go down to
1x1.

Jjagg:
Good to know, thanks!

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -306,7 +290,32 @@ private static void ConvertToABGR(int pixelHeight, int pixelWidth, int[] pixels)
...
+                throw new ArgumentException("Texture arrays are not supported on this graphics device", "arraySlice");
+            if (arraySlice < 0 || arraySlice >= ArraySize)
+                throw new ArgumentException("arraySlice must be smaller than the ArraySize of this texture and larger than 0.");
+            if (!textureBounds.Contains(checkedRect))
+                throw new ArgumentException("Rectangle must be inside the texture bounds", "rect");
+            if (data == null)
+                throw new ArgumentNullException("data");
+            var tSize = Marshal.SizeOf(typeof(T));
+            var fSize = Format.GetSize();
+            if (tSize > fSize || fSize % tSize != 0)
```
nkast: I think this has to be `(tSize > fSize *&&* fSize % tSize != 0)`

Jjagg: No, this is supposed to be like that. 
XNA requires tSize <= fSize AND fSize % tSize == 0. Intuitively this means that x structs must fit one pixel where x is a whole number. If we negate the conditions we get !(tSize <= fSize) OR !(fSize % tSize == 0) which simplifies to the check here. 

E.g. If your format is Color fSize is 4. Some valid T's are byte, byte2 and Color. Invalid T's are byte3 and anything larger than 4 bytes because you can't put *one* pixel in these types.

Jjagg: Can't seem to edit, but a better way to phrase the last sentence is: you can't represeng one pixel wit those types.

nkast: Oh, I see, for the moment I though that was for something else. Reading 2 pixels at once for example. I have done something similar with IndexBuffer, reading a `TriangleIndices` struct of 3 `Int16`. Does XNA throw error when you read multiples of Ts? 

Back to the original itension, Why can't I read `Byte3` from a `Color` If I am not interested in Alpha?

I will test XNA and come back. I believe XNA was very flexible with GetData/SetData.

nkast: Ok, I see. `Texture2D`  doesn't have an overload with Element Stride so you can't read/write partial pixels.

tomspilman:
@Jjagg - We should get this PR done and merged.  I've run into issues with the current validation on some projects.

Jjagg:
I have some deadlines coming up, but I can probably squeeze this in next week :)

tomspilman:
@Jjagg - No problem... I know all too well about deadlines. ;)

Jjagg:
I finally spent some time on this again. XNA rounds a passed rectangle to fit a DXT block. That is the position x and y coordinates are rounded **down** to the nearest multiple of 4 and the width and height are rounded **up** to the nearest multiple of 4. That seems like a stupid thing to depend on, but I guess we can just replicate it in MG. Or we can throw an exception if people think that's better.

Test/Framework/Graphics/Texture2DNonVisualTest.cs
```diff
@@ -465,5 +465,109 @@ public void SetData5ParameterExceptionTest(int arraySize, int startIndex, int el
...
+            var t = content.Load<Texture2D>(Paths.Texture("random_16px_dxt"));
+
+            var b = new byte[8];
+            var b2 = new byte[8];
+
+            t.GetData(0, new Rectangle(0,0,4,4), b, 0, 8);
+            t.GetData(1, new Rectangle(0,0,4,4), b2, 0, 8);
+            t.GetData(2, new Rectangle(0,0,4,4), b2, 0, 8);
+            t.GetData(3, new Rectangle(0,0,2,2), b2, 0, 8);
+            t.GetData(4, new Rectangle(0,0,1,1), b2, 0, 8);
```
Jjagg: @KonajuGames Do you know why this behaves the way it does? You can only pass a one pixel rectangle (larger will cause an exception), yet it fills up 8 bytes. Maybe internally it stores one DXT block with a solid color? This is the last mip btw since the texture is 16x16 px.

tomspilman: DXT is a block format and represents texture data in blocks of 4x4 pixels in either 64bits (DXT1 or BC1) or 128bits (DXT2/3/4/5 or BC2/3).  So even if the mip is just 1x1... that level requires 8 bytes (64bits for DXT1) of data to represent it.

Knowing that you have another test case of 128bit DXT/BC formats which should show it requiring 16 bytes of data.

Jjagg: > Knowing that you have another test case of 128bit DXT/BC formats which should show it requiring 16 bytes of data.

Do you mean I *should* have another test case? How do I create a texture with that kind of compression, that's not supported in XNA, right?

KonajuGames: Add a texture with a non-opaque alpha channel, process it in the content pipeline and it will produce a DXT5 texture.  DXT5 uses a 128bit block.

Jjagg: Will do, thanks! :)

Jjagg:
I think this is good to go like this! At least everything passed locally :)
Are all runtime compressed texture formats we support 4x4 block compression formats? Because I kind of hardcoded that in... If we need more flexibility we can have an extension method.

KonajuGames:
> Are all runtime compressed texture formats we support 4x4 block compression formats?

I believe they all are at this stage.

KonajuGames:
Mac build failed with this error.

> Graphics/TextureCube.OpenGL.cs(88,17): error CS0019: Operator `==' cannot be applied to operands of type `OpenTK.Graphics.OpenGL.PixelFormat' and `OpenTK.Graphics.OpenGL.All'

Might need a cast for this particular case.

Jjagg:
The tests show that this also fixes #4276.

Jjagg:
I'd like to change [TextureCube.SetData](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/TextureCube.cs#L67-L124) so it behaves the same as the other Texture.GetData/SetData methods if that's okay. That is the implementation will just be

```
public void SetData<T>(CubeMapFace face, int level, Rectangle? rect, T[] data, int startIndex, int elementCount) where T : struct
{
    ValidateParams(level, ref rect, data, startIndex, elementCount);
    PlatformSetData(face, level, rect, data, startIndex, elementCount);
}
```
And PlatformSetData will be changed as needed. This will conflict with the changes in #5166 by @dellis1972. If that can be merged in first, things will be a bit easier since it won't cause conflicts for this branch.

Jjagg:
This is meant to fail btw. I put in tests for TextureCube with DXT format, but no fixes yet (because of what I said in the previous comment).

Jjagg:
Since dellis is busy and #5166 is pretty small, I'll just merge it into this and finish it when I have the time if that's fine. @dellis1972 @tomspilman

Jjagg:
I merged in #5166 but I didn't implement GetData for compressed TextureCubes yet.
This should (finally) be good to go :)

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -308,7 +294,56 @@ private static void ConvertToABGR(int pixelHeight, int pixelWidth, int[] pixels)
...
+#if OPENGL
+                    // OpenGL only: The last two mip levels require the width and height to be
+                    // passed as 2x2 and 1x1, but there needs to be enough data passed to occupy
+                    // a 4x4 block.
+                    checkedRect.Width > 4 ? (checkedRect.Width + 3) & ~0x3 : checkedRect.Width,
+                    checkedRect.Height > 4 ? (checkedRect.Height + 3) & ~0x3 : checkedRect.Height);
+#else
+                    (checkedRect.Width + 3) & ~0x3, (checkedRect.Height + 3) & ~0x3);
+#endif
+                dataByteSize = checkedRect.Width * checkedRect.Height * fSize / 16;
```
Jjagg: I should change something here... Since below we check `elementCount * tSize != dataByteSize`, but for OpenGL/DX the checkedRect will be a different size for the lower mips so they require a different elementCount.

KonajuGames:
Merging.  Thanks @Jjagg

Issue #4718, opened 2016-04-03
Fix Mouse.SetPosition() docs to note platforms where it does not work.


tomspilman:
We need to add a `<remarks>` section to the docs for `Mouse.SetPosition()` to note which platforms the feature does not work under:

http://www.monogame.net/documentation/?page=M_Microsoft_Xna_Framework_Input_Mouse_SetPosition

I believe this is just the following platforms:
- Web
- Windows 8
- Windows Universal

cra0zy:
Web has it implemented, tho it's missing IsCursorVisible property due to it...

FrancescoBonizzi:
Hi! I didn't see where `Mouse.SetPosition` is not implemented, but I saw that in the `Angle` and `Windows Universal` solutions `PlatformSetCursor` are empty. 

Can I add the `<remarks>` there?

Jjagg:
Only Windows Universal should be mentioned in the comment. The documentation should be added to the shared SetCursor method, otherwise our doc system doesn't pick it up. Thanks!

cra0zy:
Ugh, UWP can be made to work, so instead of adding remarks we should, you know, get it working instead. I think I mentioned this somewhere else.

Jjagg:
Okay, closing for #5000 then.

Issue #5000, opened 2016-07-09
Mouse.SetPosition for UWP


cra0zy:
Yes, it's possible, the code is:
```c#
Window.Current.CoreWindow.PointerPosition = new Windows.Foundation.Point(100, 100);
```

It works with the standard UWP Blank template, but it doesn't work in the MG Template (It still moves the mouse, but it moves it to 0, 0)  Since I don't fully know how the UWP app works under the hood I leave this here so that someone else may add this part of the API.

Danthekilla:
There is some weirdness happening here.

I have tried playing around with the input on the latest monogame on UWP and we are having the same issue as cra0zy.

mrhelmut:
Cleaning this as UWP is getting deprecated #8406

PR #5124, opened 2016-08-19
Added remarks to Mouse.SetPosition

vaporwavre:
Notes that it does not work with Windows 8 and Windows Universal (#4718)


cra0zy:
Universal actually does posses a way of setting mouse position, see #5000

Danthekilla:
You can set the position of the mouse in windows 8 also. We did it in our game that launched 3 or so years ago. I will see if I can find the code.

cra0zy:
Closing this as this is incorrect information for documentation.

PR #5125, opened 2016-08-20
Add a null reference check on DisplayMode for #5040

hach-que:
There's currently a fix for #5040 in the form of adjusting templates, however this only assists developers creating new projects going forward.  For existing developers, they need to add or create an `app.manifest` file in their project.

We wasted about 6 hours remotely debugging this issue before we realised it was that Windows 10 configures a 125% DPI by default, and that on this particular machine that setting hadn't been changed to be 100%.  We then found #5040 which then documented the issue.

This adds a null check in `DEBUG` builds so that if DisplayMode is null, the developer gets a much more useful exception around the potential issue that might be occurring.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -213,8 +213,18 @@ namespace Microsoft.Xna.Framework.Graphics
 
         private void Setup()
         {
-			// Initialize the main viewport
-			_viewport = new Viewport (0, 0,
+#if DEBUG
+            if (DisplayMode == null)
+            {
+                throw new ApplicationException(
+                    "Unable to determine the current display mode.  This can indicate that the " +
+                    "game is not configured to be HiDPI aware under Windows 10 or later.  See " +
+                    "https://github.com/MonoGame/MonoGame/issues/5040 for more information.");
+            }
+#endif
+
+            // Initialize the main viewport
+            _viewport = new Viewport (0, 0,
 			                         DisplayMode.Width, DisplayMode.Height);
 			_viewport.MaxDepth = 1.0f;
 


Jjagg:
This is a good idea IMO

tomspilman:
Seems reasonable to me... merging!  Thanks @hach-que !

Jjagg:
I'm thinking maybe we should also warn devs when the application is not dpi aware in general since it may run on their system (if they don't have dpi scaling) but might crash for players. Maybe by just adding some documentation, but I'm not sure where this should go.

nkast:
Yes, the `MonoGame.Framework.WindowsUniversal.sln`  project.
Same for `MonoGame.Framework.Windows8.sln`.

MSDN says `ApplicationException` it's available for 
```.NET Framework
Available since 1.1
Windows Phone Silverlight
Available since 8.0```

for comparison `Exception` is available for 
```
Universal Windows Platform
Available since 8
.NET Framework
Available since 1.1
Portable Class Library
Supported in: portable .NET platforms
Silverlight
Available since 2.0
Windows Phone Silverlight
Available since 7.0
Windows Phone
Available since 8.1
```

nkast:
@hach-que @Jjagg 
Can you test if #5040 is a problem for store apps?
Should we limit the test of windows/WindowsGL or use a more generic Exception for winrt?

nkast:
> This adds a null check in DEBUG builds so that if DisplayMode is null, the developer gets a much more useful exception around the potential issue that might be occurring.

What about those that use Release builds and don't work directly with git sources?

> That is odd... @nkast why did the build server not catch this? You do mean UWP apps for the new Windows 10 store?

Because the change doesn't make it to the release builds.

hach-que:
Does the MonoGame build server not build Debug at all? Even for testing?

There's no reason to be doing this check in released versions of games. I don't use MonoGame in binary form (using the shipped assemblies), so I just assumed MonoGame supplied both Debug and Release editions for during development / at release for people using the binary versions.

(I know in Protogame we heavily make use of `#if DEBUG` to turn off runtime checks that the developer should have already addressed during development)

tomspilman:
> Does the MonoGame build server not build Debug at all? Even for testing?

Looks like the only one we have set to build under debug is the Windows DirectX build.

  https://github.com/MonoGame/MonoGame/blob/develop/default.build

Previously with the old build server... build times were a concern...  but it is not anymore.  We could fix this.

PR #5129, opened 2016-08-22
WinDX: Implemented a more reliable keyboard input method

### LINKED ISSUE #5122: [WinDX] Opening Steam overlay results in "stuck" keys
I'm currently working on Steam integration with my game and I came across an issue when opening the Steam overlay. The Shift and Tab keys (which you press to open the overlay) end up getting "stuck" in the game. I'm assuming this is because the Steam overlay intercepts all the key events, so MonoGame never sees the "release" events for these keys.

I'm having this problem with the WinDX build, but I imagine it also affects other desktop builds. (UPDATE: This actually seems to only affect MonoGame DirectX. Both XNA and MonoGame SDL2 work correctly.)

My current workaround to this issue is to clear MonoGame's internal key buffer (i.e. `Keyboard._keys`) whenever the Steam overlay is opened or closed. However given this is not publicly exposed, the only way to do this is by changing MonoGame to expose it or by using reflection. I am not sure if this is the best way of dealing with this, but it seems to work well.

Has anyone else experienced this particular problem? And if so, how did you solve it?

And if my workaround seems reasonable, would it be acceptable to add this to the MonoGame API (e.g. a `Keyboard.Flush` method which clears the buffer)? If so, I'm more than happy to submit a PR.

MichaelDePiazzi:
Fixes #5122

As discussed in issue #5122, this replaces DirectX RawInput with Windows GetKeyboardState for retrieving keyboard input.

This resolves "stuck" key issues when opening the Steam overlay, and should eliminate any other potential for stuck keys.

### DIFF
--- a/MonoGame.Framework/Input/Keyboard.Windows.cs
+++ b/MonoGame.Framework/Input/Keyboard.Windows.cs
@@ -4,14 +4,13 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Runtime.InteropServices;
 
 namespace Microsoft.Xna.Framework.Input
 {
     public static partial class Keyboard
     {
-        private static readonly Keys[] DefinedKeys;
+        private static readonly byte[] DefinedKeyCodes;
 
         private static readonly byte[] _keyState = new byte[256];
         private static readonly List<Keys> _keys = new List<Keys>(10);
@@ -19,16 +18,21 @@ namespace Microsoft.Xna.Framework.Input
         private static bool _isActive;
 
         [DllImport("user32.dll")]
-        [return: MarshalAs(UnmanagedType.Bool)]
         private static extern bool GetKeyboardState(byte[] lpKeyState);
 
-        private static readonly Predicate<Keys> IsKeyReleasedPredicate = key => !IsKeyPressed(key);
+        private static readonly Predicate<Keys> IsKeyReleasedPredicate = key => IsKeyReleased((byte)key);
 
         static Keyboard()
         {
-            DefinedKeys = Enum.GetValues(typeof(Keys)).Cast<Keys>()
-                .Where(k => ((int)k >= 1) && ((int)k <= 255))
-                .ToArray();
+            var definedKeys = Enum.GetValues(typeof(Keys));
+            var keyCodes = new List<byte>(Math.Min(definedKeys.Length, 255));
+            foreach (var key in definedKeys)
+            {
+                var keyCode = (int)key;
+                if ((keyCode >= 1) && (keyCode <= 255))
+                    keyCodes.Add((byte)keyCode);
+            }
+            DefinedKeyCodes = keyCodes.ToArray();
         }
 
         private static KeyboardState PlatformGetState()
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/Input/Keyboard.Windows.cs
```diff
@@ -4,21 +4,57 @@
...
-        static List<Keys> _keys;
+        private static readonly Keys[] DefinedKeys;
+
+        private static readonly byte[] _keyState = new byte[256];
+        private static readonly List<Keys> _keys = new List<Keys>(10);
+
+        private static bool _isActive;
+
+        [DllImport("user32.dll")]
+        [return: MarshalAs(UnmanagedType.Bool)]
```
tomspilman: You should double check to see if you really need this `MarshalAs` attribute.  Sometimes they cause extra performance overhead.

MonoGame.Framework/Input/Keyboard.Windows.cs
```diff
@@ -4,21 +4,57 @@
...
+        private static extern bool GetKeyboardState(byte[] lpKeyState);
+
+        private static readonly Predicate<Keys> IsKeyReleasedPredicate = key => !IsKeyPressed(key);
+
+        static Keyboard()
+        {
+            DefinedKeys = Enum.GetValues(typeof(Keys)).Cast<Keys>()
+                .Where(k => ((int)k >= 1) && ((int)k <= 255))
+                .ToArray();
```
tomspilman: We try to avoid all Linq... you should be able to do the same thing without it.

MonoGame.Framework/Input/Keyboard.Windows.cs
```diff
@@ -4,21 +4,57 @@
...
+using System.Runtime.InteropServices;
 
 namespace Microsoft.Xna.Framework.Input
 {
     public static partial class Keyboard
     {
-        static List<Keys> _keys;
+        private static readonly Keys[] DefinedKeys;
+
+        private static readonly byte[] _keyState = new byte[256];
```
tomspilman: This might need to be saved for a later PR...  but this uses 8x the memory it needs to.  We could easily encode all the state into 4 `uint64`s as bits.  It would likely be much more efficient to query as well.

MonoGame.Framework/Windows/WinFormsGameWindow.cs
```diff
@@ -300,48 +285,6 @@ private void OnMouseLeave(object sender, EventArgs e)
...
-                    xnaKey = (args.ScanCodeFlags & ScanCodeFlags.E0) != 0 ? XnaKey.RightControl : XnaKey.LeftControl;
-                    break;
-
-                case 0x38: // Alt
-                    xnaKey = (args.ScanCodeFlags & ScanCodeFlags.E0) != 0 ? XnaKey.RightAlt : XnaKey.LeftAlt;
-                    break;
-
-                default:
-                    xnaKey = (XnaKey)args.Key;
-                    break;
```
tomspilman: Do we know if things like left/right shift, ctrl, and alt work with the new method?

MonoGame.Framework/Input/Keyboard.Windows.cs
MichaelDePiazzi: Ok, I'll look into it.

MonoGame.Framework/Input/Keyboard.Windows.cs
MichaelDePiazzi: Yeah, I know it can be bad from a memory pressure point of view. But I figured it'd be ok since it's only called once when the static constructor is first called. But I can rewrite it without using LINQ if you prefer.

MonoGame.Framework/Input/Keyboard.Windows.cs
MichaelDePiazzi: Yeah, that's a fair point. I was going for better performance at the cost of a bit more memory usage, but didn't really stop to think if I could do it using less memory. Leave it with me, I'll improve it.

MonoGame.Framework/Windows/WinFormsGameWindow.cs
MichaelDePiazzi: Yes, they do.

MonoGame.Framework/Input/Keyboard.Windows.cs
MichaelDePiazzi: Yeah, it's not actually required for the 4-byte `BOOL` type (source: https://blogs.msdn.microsoft.com/jaredpar/2008/10/14/pinvoke-and-bool-or-should-i-say-bool/). I'll remove it.

MichaelDePiazzi:
@tomspilman I've made the changes you suggested. In regards to your comment regarding memory usage, I've reduced it by 75% (from 636 bytes to 159 bytes). I held off on going further and using a bit-mask as I couldn't think of a way to do it that would be at least equivalent performance wise...

dellis1972:
@tomspilman this this good to go?

tomspilman:
Yeah....  not much automated testing could be done here.    Let's just merge this and see how it does.

PR #5133, opened 2016-08-25
Dispose of all graphical resources in unit tests

Jjagg:
Wew, this was tedious... @KonajuGames

### DIFF
--- a/Test/Framework/Components/Colored3DCubeComponent.cs
+++ b/Test/Framework/Components/Colored3DCubeComponent.cs
@@ -111,6 +111,18 @@ namespace MonoGame.Tests.Components {
 			CreateCubeIndexBuffer ();
 		}
 
+	    public void UnloadContent()
+	    {
+            basicEffect.Dispose();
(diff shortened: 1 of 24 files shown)


tomspilman:
Thanks @Jjagg .

I wonder if there is a good way to automatically detect and assert if we miss a dispose in a future test?

Jjagg:
Actually disposing game should dispose GraphicsDeviceManager which disposes the current GraphicsDevice which I thought disposed all graphicsresources, but that doesn't always seem to happen. Maybe they don't all register with the GraphicsDevice? They register when the GraphicsDevice setter is called so they should. 

Anyway, in MG these should all be in a collection in GraphicsDevice, so we can check this if the GraphicsDeviceTestBaseFixture is used either by explicitly disposing or checking through reflection (gross). We can't do that for XNA though so I think we just need to pay attention that unmanaged resources are always explicitly disposed. It's not that big a deal if we miss a couple I think

PR #5137, opened 2016-08-28
Fix SoundEffect disposal test to not use internal constructor

Jjagg:
XNA tests didn't compile because an internal SoundEffectInstance constructor was used in the InstanceNotDisposed test. This test was also not run because it was missing the "Test" attribute. 

After adding it in, the test ran fine in XNA but failed in MG because when Game is disposed the master voice is also disposed and you don't have access to the subvoices after that. I don't think XNA disposes of the master voice when game is disposed because trying to use a SFX after disposing game will work in XNA. The 'fix' in the first commit here is crap, because it doesn't solve anything, but I only just realised that :p 

I think Game really shouln't be involved in this to keep things nicely separated. It would be nice if you could run Audio through MG without creating a Game (just using MG as an audio wrapper). So I'd like to expose functions to setup audio (automatically done when creating a game) and dispose of all audio (not called by game) so the user has more control and game disposal doesn't kill all audio (compliant with XNA behaviour). Thoughts?

### DIFF
--- a/Test/Framework/Audio/SoundEffectTest.cs
+++ b/Test/Framework/Audio/SoundEffectTest.cs
@@ -272,15 +272,12 @@ namespace MonoGame.Tests.Audio
         public void InstanceNotDisposedWhenGameDisposed()
         {
             var game = new Game();
-            var s = new SoundEffectInstance();
             var d = new DynamicSoundEffectInstance(44100, AudioChannels.Stereo);
 
             game.Dispose();
 
-            Assert.IsFalse(s.IsDisposed);
             Assert.IsFalse(d.IsDisposed);
 
-            s.Dispose();
             d.Dispose();
         }
 
--- a/Test/Framework/Audio/SoundEffectTest.cs
+++ b/Test/Framework/Audio/SoundEffectTest.cs
@@ -270,7 +270,7 @@ namespace MonoGame.Tests.Audio
         }
 
         // TODO creating/disposing a Game should not create/dispose the master voice
-        [Test]
+        [Test, Ignore]
         public void InstanceNotDisposedWhenGameDisposed()
         {
             var game = new Game();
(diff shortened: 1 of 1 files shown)


Jjagg:
I removed the breaking part of the test since it's not that important and what's left shows the issue too. At least people can then build XNA tests without having to solve this first. I opened issue #5153 to keep track of the problem this PR initially wanted to adress, described in my comment above. 

@tomspilman It might be a good idea to get this merged ASAP, because this a huge annoyance for people that run tests on latest develop (me included). I caused this btw, so sorry if anyone ran into this :p 

Should the build server try to build XNA tests to prevent things like this in the future or is that extra work for too little gain? It's easy to fix this stuff, but a bit annoying that bugs like this can make it in.

tomspilman:
> Should the build server try to build XNA tests to prevent things like this in the future 

It should... we have had issues however running the XNA tests on the build server.  See https://github.com/MonoGame/MonoGame/pull/3550 for details.

Test/Framework/Audio/SoundEffectTest.cs
```diff
@@ -272,15 +272,12 @@ public void Ctor2()
         public void InstanceNotDisposedWhenGameDisposed()
         {
             var game = new Game();
-            var s = new SoundEffectInstance();
```
tomspilman: You could create a dummy `SoundEffect` like I do here:

https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/Audio/SoundEffectTest.cs#L151

Then you can restore this part of the test.

Jjagg: Oh right, will do.

Jjagg:
@tomspilman The test will fail on MG as is because the MasterVoice will be disposed when game is disposed and after that attempting to dispose the other voices will cause an exception to be thrown as described in #5153

Test/Framework/Audio/SoundEffectTest.cs
tomspilman: @Jjagg - I think you might have forgotten to fix this?

Jjagg: I don't know exactly how MG should behave with regards to disposing the master voice, that's why I opened #5153 for discussion.
I commited the ignore today (because the test was failing) so this can be merged. This thing is really a pain because I have to edit this test before trying to compile tests in another branch.

tomspilman:
I'll merge this for now then.  We can discuss more in #5153.

PR #5138, opened 2016-08-28
DXT Compression: Keep the delegate from being re-located or collected…

nkast:
… by the garbage collector.

fix: http://community.monogame.net/t/mgcb-dxt-compression-sometimes-throws-nullpointer/7356
fix: http://community.monogame.net/t/upgrading-to-monogame-3-5-1-from-3-5-0-1122-processor-textureprocessor-had-an-unexpected-failure/7432

DXT code uses callbacks from unmanaged code. Under memory pressure those delegates could be re-located resulting to invalid pointers in nvtt.dll. One way to reproduce it is to call `GC.Collect()` right after `outputOptions.SetOutputHandler(handler.BeginImage, handler.WriteData);`  (https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Graphics/DxtBitmapContent.cs#L176).

*How to: Marshal Callbacks and Delegates By Using C++ Interop*
https://msdn.microsoft.com/en-us/library/367eeye0.aspx

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/DxtBitmapContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/DxtBitmapContent.cs
@@ -11,21 +11,35 @@ using System.Runtime.InteropServices;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 {
-    class DxtDataHandler
+    class DxtDataHandler: IDisposable
     {
         private BitmapContent _content;
         byte[] _buffer;
         int _offset;
+        
+        GCHandle delegateHandleBeginImage;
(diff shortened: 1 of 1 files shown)


KonajuGames:
Ah, I hadn't thought of that.  I'm not a fan of how the NVTT compressor works with many callbacks, and this makes sense.

tomspilman:
Great catch @nkast .... never considered that delegates could be collected like that, but it is now obvious how broken that was.

nkast:
Same here, I was fighting this 'mysterious' behavior for hours until I came across msdn documentation and everything became clear in an instance. We will never stop learning...

PR #5141, opened 2016-08-28
Added Localized Resource Processing and Loading.

### LINKED ISSUE #5136: Add Localization  Content Processors and Methods to MG
XNA had a sample for dealing with localised assets [1]. I think it might be a  good idea to add these as default processors to the MG content Pipeline so they are just built in.

So Im proposing we add the  `LocalizedFontDescription` and `LocalizedFontProcessor` code from the sample and then add a new method to the contentManager

`T LoadLocalized<T>(string assetName)`

People can then have the choice of loading using the standard Load method or using the other if that asset has a localized version.

[1] http://xbox.create.msdn.com/en-US/education/catalog/sample/localization

dellis1972:
Fixes #5136

When localizing games you sometimes need to load localized
assets as well as text. This commit includes the sample
XNA processors and ContentManager code into the core of the
MonoGame system. This means that people no longer need to
implement this stuff themselves.

Usage:

Define all of the "string" resources in standard .resx resource
files. Add these to the .spritefont file

```
   <ResourceFiles>
      <Resx>..\Strings.resx</Resx>
      <Resx>..\Strings.da.resx</Resx>
      <Resx>..\Strings.fr.resx</Resx>
      <Resx>..\Strings.ja.resx</Resx>
      <Resx>..\Strings.ko.resx</Resx>
    </ResourceFiles>
```

Change the .spritefont over to use the LocalizedFontProcessor.
Load the font as normal.

For other localized assets you can make use of the new

``````
```Content.LoadLocalized<T>```
``````

method. This will attempt to load assets which contain a culture
extension e.g

```
Texture1.fr.xnb
Texture1.en-gb.xnb
```

Note either the Name or the TwoLetterISOLanguageName can be used.

### DIFF
--- a/Test/ContentPipeline/TestProcessorContext.cs
+++ b/Test/ContentPipeline/TestProcessorContext.cs
@@ -85,6 +85,19 @@ namespace MonoGame.Tests.ContentPipeline
             if (typeof(TOutput) == typeof(MaterialContent) && typeof(TInput).IsAssignableFrom(typeof(MaterialContent)))
                 return (TOutput)((object)input);
 
+            var processor = (ContentProcessor<TInput, TOutput>)typeof(ContentProcessor<TInput, TOutput>).Assembly.CreateInstance("Microsoft.Xna.Framework.Content.Pipeline.Processors."+ processorName);
+            if (processor != null) {
+                var type = processor.GetType();
+                foreach (var kvp in processorParameters)
+                {
+                    var property = type.GetProperty(kvp.Key);
+                    if (property == null)
+                        continue;
+                    property.SetValue(processor, kvp.Value);
+                }
+                return processor.Process(input, this);
+            }
+
             throw new NotImplementedException();
         }
     }
--- a/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
+++ b/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
@@ -105,6 +105,7 @@
     <Compile Include="Graphics\Etc1BitmapContent.cs" />
     <Compile Include="Graphics\FontDescription.cs" />
     <Compile Include="Graphics\FontDescriptionStyle.cs" />
+    <Compile Include="Graphics\LocalizedFontDescription.cs" />
     <Compile Include="Graphics\GeometryContent.cs" />
     <Compile Include="Graphics\GeometryContentCollection.cs" />
     <Compile Include="Graphics\Font\Glyph.cs" />
@@ -149,6 +150,7 @@
     <Compile Include="Processors\EffectProcessor.cs" />
     <Compile Include="Processors\EffectProcessorDebugMode.cs" />
     <Compile Include="Processors\FontDescriptionProcessor.cs" />
+    <Compile Include="Processors\LocalizedFontProcessor.cs" />
     <Compile Include="Processors\FontTextureProcessor.cs" />
     <Compile Include="Processors\MaterialProcessor.cs" />
     <Compile Include="Processors\MaterialProcessorDefaultEffect.cs" />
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

**Comment by KonajuGames** [comment]
I'm happy with this as a simple localization system that should handle most localization needs that most devs have.  It also doesn't stop anyone implementing their own localization if needed.

**Comment by tomspilman** [comment]
I'm sort of on the fence on this one...  I think we should give users good guidance on how to deal with this, but unsure if this is a good solution for most users.

So I guess i'll leave the decision to you guys on this one.

However... if we're going to add it lets at unit tests for it.

**Comment by dellis1972** [comment]
well the way I see it is this is there for people to use. If they want to use it its there if they want to use something else, thats cool too. 

I'll add some unit tests this weekend and some docs I think.  Perhaps I can start a new trend of adding docs with PR's :P

**Comment by dellis1972** [comment]
I wonder if we can add support for picking the resx files to the pipeline tool...like we do for references. That would make using this easier.

**Comment by tomspilman** [comment]
> Perhaps I can start a new trend of adding docs with PR's :P

That would be fantastic...  it is specifically why we made the docs part of the repo and not some other wiki. :)

**Comment by jamesford42** [comment]
I worked on a project that was using what looks like this same sample code, some time ago. 

My only complaint is that it does not scale up very well. 

At a certain point, you end up with a single sheet thats almost too big to allocate as a texture because it contains glyphs for 8 languages. And once your string table starts getting significantly big (any game that has a lot of dialog) the processing time ends up getting almost unreasonably slow.

Still, it's a decent starting point. And I have some ideas on how both of those prior concerns could be addressed.

**Comment by KonajuGames** [comment]
> And I have some ideas on how both of those prior concerns could be addressed.

Two approaches I have used on large localizations are:

- A font per language.
- A dynamic font where glyphs are loaded from disk on demand and cached in a texture page.

**Comment by dellis1972** [comment]
I figure the best option would be to not put ALL the text into one sprite
sheet.

Split out the resx files into different sheets. Say one for western
languages, and others for eastern based ones.
Then its just a question of loading the correct font at runtime (and
placement etc).

Just an idea :)

**Comment by dellis1972** [comment]
@tomspilman @KonajuGames I added the docs. I guess now that we have a FontProcessor Unit test I need to add one for these too :)

**Comment by dellis1972** [comment]
OK . Added Unit test for the processor.. added templates for the Pipeline tool. Fixed a few bugs. Optimised the processor so that it does not duplicate characters.

**Comment by KonajuGames** [comment]
I believe this will be useful, and at least help people to think about localizing their games.  I'm going to merge it.  Thanks for the complete PR, @dellis1972

File: Test/ContentPipeline/TestProcessorContext.cs [hunk 1]
```diff
@@ -85,6 +85,19 @@ public override void AddOutputFile(string filename)
...
 
+            var processor = (ContentProcessor<TInput, TOutput>)typeof(ContentProcessor<TInput, TOutput>).Assembly.CreateInstance("Microsoft.Xna.Framework.Content.Pipeline.Processors."+ processorName);
+            if (processor != null) {
+                var type = processor.GetType();
+                foreach (var kvp in processorParameters)
+                {
+                    var property = type.GetProperty(kvp.Key);
+                    if (property == null)
+                        continue;
+                    property.SetValue(processor, kvp.Value);
```
nkast: This line broke the XNA tests.
"No overload for method 'SetValue' takes 2 arguments"

dellis1972: Weird. Do we not run the XNA tests on the Build system?

tomspilman: Unfortunately no...

https://github.com/MonoGame/MonoGame/pull/3550#issuecomment-205644961

I should try it again and see if i can make progress on it.

PR #5142, opened 2016-08-29
Improve input test coverage

Jjagg:
Better coverage for several input classes as requested in #4744


Test/Framework/Input/JoystickTest.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            var hat = new JoystickHat
+            {
+                Left = left,
+                Right = right,
+                Up = up,
+                Down = down,
+            };
+
+            Assert.AreEqual(left, hat.Left);
```
polsys: These `Joystick` tests seem to be equivalent to
```cs
var hat = new JoystickHat();
hat.Left = left;
Assert.AreEqual(left, hat.Left);
```

I think we can trust auto-setters instead of artificially increasing the line coverage.

Test/Framework/Input/MouseTest.cs
```diff
@@ -12,21 +16,56 @@ class MouseTest
...
+                Y = y,
+                ScrollWheelValue = scrollWheel,
+                LeftButton = leftButton,
+                MiddleButton = middleButton,
+                RightButton = rightButton,
+                XButton1 = xButton1,
+                XButton2 = xButton2
+            };
+            Assert.AreEqual(state.X, x);
+            Assert.AreEqual(state.Y, y);
```
polsys: Same here, though the equality and hash code check are good to have in my opinion.

Test/Framework/Input/JoystickTest.cs
Jjagg: Yes, that is equivalent, these are tests for the setters.

> artificially increasing the line coverage.

We can't really do any other automated testing than checking constructors, setters and stuff like that for classes like this, right?

tomspilman: > I think we can trust auto-setters instead of artificially increasing the line coverage.

Well these tests are not just about looking for bugs today...  they are there to detect bugs in some future changes/refactors.  We cannot guarantee that these will stay as auto props in the future.... that is why we should still test them now.

So i think this is fine to do.

Jjagg:
Oops, didn't notice there were already some tests for this. I'll merge them with these

Jjagg:
IMO once we solve the issue described in my previous comment this is good to be merged. I don't think we can test the Keyboard/Mouse/GamePad classes without actually having these devices attached to the build server. Maybe there's some tools to emulate this? I tried looking for software that can help us achieve that but only found stuff to map from one device to another, like keyboard for mouse emulation and mapping from one gamepad to another.

MichaelDePiazzi:
@Jjagg Not quite. The tests are proving that the virtual thumbstick buttons ignore the specified deadzone. Behind the scenes, they need to always use the independent axes deadzone mode. See the original issue #4741 and PR #4742 for the details.

I'm a little confused by this PR though... It says it's about improving test coverage, but you've gone and made other non-test related changes with one of the commits mentioning something about a "deadzone fix". What's this about? The changes seem to undo the fix made in my previous PR, and are why the tests proving that fix are now failing...

Jjagg:
> I'm a little confused by this PR though... It says it's about improving test coverage, but you've gone and made other non-test related changes with one of the commits mentioning something about a "deadzone fix". What's this about?

That's because I wrote a [failing test](https://github.com/MonoGame/MonoGame/pull/5142/files#diff-9889d68324a60145be6e8ba50b13d8f3R75) and wanted to fix it, but the apparent fix broke your test which is why I bumped you. 
The issue/PR you linked clears it up for me. The test I wrote is wrong as it would fail for XNA so I'll get rid of it and add a comment in the SetVirtualButtons method so noone else makes this mistake.

MichaelDePiazzi:
@Jjagg Ah, no probs. It can be tricky to add tests to existing code, especially when it may not be clear how it's meant to behave.

Also, good idea about adding a comment. In hindsight, I should've originally added a comment about the XNA behaviour as it's not really obvious.

And it does suck we can't run these tests on XNA, it would definitely help a lot! Being able to emulate input devices might be a solution, but I don't know of a practical and reliable way of doing that...

Jjagg:
> And it does suck we can't run these tests on XNA, it would definitely help a lot! Being able to emulate input devices might be a solution, but I don't know of a practical and reliable way of doing that...

I'll keep searching for this a bit.

Jjagg:
@tomspilman If you ask me, this is good to be merged. Like I said before, mostly Keyboard/Mouse/Joystick/GamePad are not well tested, but we can't really do this without hardware devices for the build bots (even then it would be an issue testing locally) or some nice way of emulating devices. The second option would be ideal, but I haven't found anything that can help us with this yet.

Test/MonoGame.Tests.XNA.csproj
```diff
@@ -143,6 +141,8 @@
     <Compile Include="Framework\Graphics\TextureCubeTest.cs" />
     <Compile Include="Framework\Graphics\VertexBufferTest.cs" />
     <Compile Include="Framework\Graphics\ViewportTest.cs" />
+    <Compile Include="Framework\Input\KeyboardTest.cs" />
+    <Compile Include="Framework\Input\MouseTest.cs" />
```
tomspilman: Missing `GamePadTest.cs`?

MonoGame.Framework/Input/GamePadThumbSticks.cs
```diff
@@ -63,7 +63,7 @@ internal GamePadThumbSticks(Vector2 leftPosition, Vector2 rightPosition, GamePad
             else
                 ApplySquareClamp();
 
-            SetVirtualButtons(leftPosition, rightPosition);
+            SetVirtualButtons(left, right);
```
tomspilman: What was this change for?

Jjagg: Woops, this is a leftover from my incorrect test/fix that's mentioned in the discussion. It doesn't really matter in practice here, but left/right are the processed values. As in they are set to 0 0 when they fall in the dead zone. IMo it's clearer if I revert this change.

Test/MonoGame.Tests.XNA.csproj
Jjagg: Jep, I left it out because none of the tests compiled for XNA (most things we use to verify tests are not public API), but with the old tests merged at least one will. I'll add this back in and #if around failing tests. Thanks for the review Tom!

MonoGame.Framework/Input/GamePadThumbSticks.cs
tomspilman: Ok.

MichaelDePiazzi: > It doesn't really matter in practice here

Actually, it does matter. The virtual buttons logic needs the raw values before any deadzone clamping has been applied. The fact that the tests aren't failing just means that I didn't add sufficient coverage for my original fix related to this. Please add the following test cases to the `ThumbStickVirtualButtonsIgnoreDeadZone` test and you should find they fail if the code uses the processed values like it currently is:
```c#
yield return new TestCaseData(
    new Vector2(0.4f, -0.4f), new Vector2(-0.4f, 0.4f), GamePadDeadZone.IndependentAxes,
    Buttons.LeftThumbstickRight | Buttons.LeftThumbstickDown | Buttons.RightThumbstickLeft | Buttons.RightThumbstickUp);

yield return new TestCaseData(
    new Vector2(0.4f, 0f), new Vector2(0f, 0.4f), GamePadDeadZone.Circular,
    Buttons.LeftThumbstickRight | Buttons.RightThumbstickUp);
```

KonajuGames:
Merging. Thansk @Jjagg

Issue #4698, opened 2016-04-01
Add Unit Tests for AlphaTestEffect, SpriteEffect, SkinnedEffect, DualTextureEffect, and 


tomspilman:

cbrantley91:
Am looking to re-familiarize myself with MonoGame and contribute where I can.  Will work something similar to your example, if that's cool.

tomspilman:
@cbrantley91 - It is beyond cool... it would be fantastic to finally get some good tests for this.

I suggest keep the first pass simple... maybe convert over my example to fit our unit test framework.  If that works well and we get that PR merged then you can look at applying the same technique to the other effect types.

Let us know if you run into any trouble.

cbrantley91:
@tomspilman 
Update on this. Your test references a BoxHelpers class that doesn't seem to exist within the MonoGame project; assumed not there for a reason (proprietary or whatnot), so started a quick re-implementation of that.  Easy, but slow; however, the call to

`GetNormalColorTextureUnitCube(GraphicsDevice device)`

tripped me up a bit.  VertexPositionNormalTexture exists, and VertexPositionColorTexture exists, but VertexPositionNormalColorTexture doesn't seem to exist.  Remove anything related to that VBO, or is BoxHelpers doing some magic?

Also figure I'd make sure BoxHelpers isn't already in existence for a reason.

tomspilman:
I'm out of the office, but I can totally give you that BoxHelpers file.  Give me a few hours.

tomspilman:
@cbrantley91 - Sorry for the slow update on this...   been super busy.

Here is the implementation of `BoxHelpers`,  `VertexPositionNormalColor`, and `VertexPositionNormalColorTexture`:

 https://gist.github.com/tomspilman/e87edf3435fbabadc62fd15f854a4e32

When I wrote it I was surprised that `BasicEffect` had modes for which a custom vertex format needed to be written that was not built into XNA/MonoGame.

@KonajuGames @dellis1972 - Should we add these vertex types into MonoGame then as official types?

KonajuGames:
Probably no harm in adding them.

dellis1972:
I agree

cbrantley91:
If you'd like I'll include it as part of the PR, though likely should fall under a different issue.

Currently my troubles have been running the tests under Linux: getting content pipeline issues causing segfaults, and include/exclude not implemented for the test runner.

Commenting out the one content pipeline issue moved me on to a load of audio errors.

[...]

PR #5144, opened 2016-08-30
Test for BasicEffect

Jjagg:
First steps in resolving #4698. Right now the MG test is failing :/ Here's the diff:

diff-drawtest

```Required similarity: 0.99
Similarity: 0.9742```

Looks like a minor offset, may be something related to the half pixel offset thingy but I don't really know what that's all about.


tomspilman:
I would make sure `GraphicsAdapter.UseReferenceDevice = true` before initializing the `GraphicsDevice`.  This ensures the tests are run under the software reference device and not the GPU hardware.  This runs slower, but it runs more correct without the differences you'll find across different GPU drivers/hardware.

We set this by default in the `VisualTestGame` which I don't think you're using here.

Jjagg:
> I would make sure GraphicsAdapter.UseReferenceDevice = true before initializing the GraphicsDevice.

This didn't help. Similarity value changed from 0.9742 to 0.9741.

tomspilman:
> This didn't help. Similarity value changed from 0.9742 to 0.9741.

Did you regenerated the `DrawTest.png` reference image with `UseReferenceDevice` enabled too?

Jjagg:
Jep, it mattered a bit more for XNA like a 0.025 similarity difference between the gpu and reference device or something like that.

tomspilman:
This seems like a bit more difference than you should be getting... so maybe there is an issue here, but it is hard to say for sure.  Things that could be causing issues:

   - Half-pixel offset.
   - Texture mips.
   - Shader optimizations.
   - Subtle matrix math differences,
   
Any other ideas @KonajuGames ?

Jjagg:
I'll compare the matrix value and try the half pixel offset. Those are easy :p

Jjagg:
> I'll compare the matrix value and try the half pixel offset.

These two are not the issue :/

tomspilman:
> These two are not the issue :/

Figured that might be the case.

Are we using the *exact* same texture XNB on both XNA and MonoGame?  That texture is uncompressed color correct?

Jjagg:
> Are we using the exact same texture XNB on both XNA and MonoGame? That texture is uncompressed color correct?

Yes, I just double-checked both these things.

tomspilman:
Then all that is left here that I can think of is differences based on shader optimiztions.  We don't expose a method to turn down the shader optimization level down...  but you can hack this into 2MGFX:

 https://github.com/MonoGame/MonoGame/blob/develop/Tools/2MGFX/EffectObject.hlsl.cs#L8

You can see the different flags for compiler options here:

http://www.nudoq.org/#!/Packages/SharpDX.Direct3D9/SharpDX.Direct3D9/ShaderFlags

I would try the different `OptimizationLevel` flags as well as maybe `IeeeStrictness` or `PartialPrecision`.

You will need to rebuild the stock effects after you make those changes using the `RebuildMGFX.bat` batch file in:

https://github.com/MonoGame/MonoGame/tree/develop/MonoGame.Framework/Graphics/Effect/Resources

Could be under XNA partial precision was preferred for stock effects or a lower optimization level that resulted in slight differences in output.

Issue #4446, opened 2016-01-20
XAudio2 Cue.Play() fail


GrafSeismo:
```
Cue.Apply3D(Listener, Emitter);
Cue.Play(); // <--- Error happen here
```

An exception of type 'SharpDX.SharpDXException' occurred in SharpDX.dll but was not handled in user code

Additional information: HRESULT: [0x88960001], Module: [SharpDX.XAudio2], ApiCode: [XAUDIO2_E_INVALID_CALL/InvalidCall], Message: Unknown

SharpDX.SharpDXException was unhandled by user code
  HResult=-2003435519
  Message=HRESULT: [0x88960001], Module: [SharpDX.XAudio2], ApiCode: [XAUDIO2_E_INVALID_CALL/InvalidCall], Message: Unknown
  Source=SharpDX
  StackTrace:
       at SharpDX.Result.CheckError()
       at SharpDX.XAudio2.XAudio2.CreateSourceVoice_(SourceVoice sourceVoiceOut, IntPtr sourceFormatRef, VoiceFlags flags, Single maxFrequencyRatio, IntPtr callbackRef, Nullable`1 sendListRef, Nullable`1 effectChainRef)
       at SharpDX.XAudio2.SourceVoice.CreateSourceVoice(XAudio2 device, WaveFormat sourceFormat, VoiceFlags flags, Single maxFrequencyRatio, IntPtr callback, EffectDescriptor[] effectDescriptors)
       at SharpDX.XAudio2.SourceVoice..ctor(XAudio2 device, WaveFormat sourceFormat, VoiceFlags flags, Single maxFrequencyRatio, Boolean enableCallbackEvents, EffectDescriptor[] effectDescriptors)
       at SharpDX.XAudio2.SourceVoice..ctor(XAudio2 device, WaveFormat sourceFormat, VoiceFlags flags, Single maxFrequencyRatio, Boolean enableCallbackEvents)
       at SharpDX.XAudio2.SourceVoice..ctor(XAudio2 device, WaveFormat sourceFormat, VoiceFlags flags, Single maxFrequencyRatio)
       at Microsoft.Xna.Framework.Audio.SoundEffect.PlatformSetupInstance(SoundEffectInstance inst)
       at Microsoft.Xna.Framework.Audio.SoundEffect.GetPooledInstance(Boolean forXAct)
[...]

PR #5150, opened 2016-09-03
[XAudio] bugfix: read adpcm blockAlign as int16

nkast:
https://msdn.microsoft.com/en-us/library/windows/desktop/dd757713(v=vs.85).aspx

https://github.com/MonoGame/MonoGame/blob/3713daf5e632395a1837f681e2d56299eb0e99b4/MonoGame.Framework/Content/ContentReaders/SoundEffectReader.cs#L24-L33

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -181,7 +181,7 @@ namespace Microsoft.Xna.Framework.Audio
             var format = BitConverter.ToInt16(header, 0);
             var channels = BitConverter.ToInt16(header, 2);
             var sampleRate = BitConverter.ToInt32(header, 4);
-            var blockAlignment = BitConverter.ToInt32(header, 12);
+            var blockAlignment = BitConverter.ToInt16(header, 12);
 
             WaveFormat waveFormat;
             if (format == 1)


nkast:
Still this didn't fix the problem with adpcm sounds. 
The pipeline always write blockAlign=0 in XNBs and sharpdx fails at https://github.com/MonoGame/MonoGame/blob/13209fb0fcf7adeabd6a1317ffd695143764a7d2/MonoGame.Framework/Audio/SoundEffect.XAudio.cs#L287
https://github.com/MonoGame/MonoGame/issues/4446

Probably we have to peak the blockalign from the generated file header from ffmpeg.
@KonajuGames ?

nkast:
For WUP apps, the SourceVoice seems to cast the value back to int16 and throws an exception. Sometimes I got dividebyZero instead of SharpDXException. 
for Desktop app it looked like SharpDX/XAudio2 was happy what whatever crazy value (0x00040000) we got. Now the desktop pltform fails as well with same error.

KonajuGames:
This is interesting that XAudio2 uses a customized version of ADPCM.
https://msdn.microsoft.com/en-us/library/windows/desktop/ee415711(v=vs.85).aspx

Looking at [other sources](https://github.com/naudio/NAudio/blob/master/NAudio/Wave/WaveFormats/AdpcmWaveFormat.cs), the blockAlign field should be the same as the SamplesPerBlock in that linked documentation.

tomspilman:
Thanks @nkast ...  even if this doesn't fix your current bug, this was a bug.  Block alignment is 2 bytes in a WAVE header.

PR #5151, opened 2016-09-03
[Pipeline Tool] Better Property Grid

cra0zy:
Changes:
 - single click to start cell editing
 - bool cells now have a checkbox control associated with them instead of double clicking to change value
 - if a cell is being edited you can click anywhere inside property grid to cancel editing it and it won't start another cell edit


tomspilman:
Thanks @cra0zy ... merging!

PR #5152, opened 2016-09-03
Removing old wiki references

tomspilman:
This PR just removes old references to the GitHub wiki which has been removed to avoid confusion by new users.


tomspilman:
@MrGrak - FYI.

Issue #5146, opened 2016-09-01
Content Pipeline Tool has so many issues


Linchy:
I'm running into so many problems today with the tool:
1. can't press the DEL key when changing 'reference' paths, it gives an error dialog.
2. I have to reference the exact dll for custom importers, meaning when I change to x86 configuration and make a change it doesn't know about it as it's still looking at bin/debug. Additionally the ref paths don't support ($OutDir) and other macros so I can't automate this??
3. Re-adding a reference doesn't register in the importer dialog, it just says '[missing] tmx importer'. But then I clean and rebuild and it works...
4. I can't have the tool open and build my projects, as VS tries to replace importer dlls but it can't as the tool has them locked.
5. I keep my mgcb file in a seperate project with just that in, because it has so many issues (an attempt to isolate it), but now any changes I make to files don't get registered. Shouldn't it go 'build main project -> build content project -> build mgcb as there were changes' ??
6. It doesn't support x86 importers. Fine. But at least tell me instead of just letting me add the dll as if it is all ok!!
7. [UI] When building things the scrollbar doesn't go to the bottom and I can't see what is currently building.
8. [UI] It highlights some of my shaders as failed to build with red, but actually they did and it is just warnings?!
9. [UI] Always asks me if I want to save before building. Can this not just be automatic? In what instance would I not want to save before building?

So many problems. I'd like to contribute fixes if possible, if you point me in the right direction to get started.

This is with the `Monogame Pipeline Tool v3.5.1.1679`

Linchy:
Another two from today:
10. [UI] Can't copy one line from the text output, it just highlights all or none.
11. [UI] The Filter Output categorises some errors incorrectly under the wrong file

tomspilman:
Thanks for reporting all this @Linchy .

@cra0zy

cra0zy:
Just gonna mention that it would be much easier if this issue was 11 separate issues, because then in each discussion I could focus on discussing about that 1 issue instead of having to use one comment to respond on 11 different things.

That said I'll get to respond to this later today... it's gonna take some time to write everything up.

cra0zy:
> can't press the DEL key when changing 'reference' paths, it gives an error dialog.

A bit more clarification please.

> I have to reference the exact dll for custom importers, meaning when I change to x86 configuration and make a change it doesn't know about it as it's still looking at bin/debug. Additionally the ref paths don't support ($OutDir) and other macros so I can't automate this??

[...]

PR #5155, opened 2016-09-04
Don't show Shader effect warnings as build errors

nkast:
fx

As discussed in https://github.com/MonoGame/MonoGame/issues/5146#issuecomment-244597511

### DIFF
--- a/Tools/Pipeline/Common/OutputParser.cs
+++ b/Tools/Pipeline/Common/OutputParser.cs
@@ -15,6 +15,7 @@ namespace MonoGame.Tools.Pipeline
         Skipping,
         BuildAsset,
         BuildError,
+        BuildWarning,
         BuildErrorContinue,
         BuildEnd,
         BuildTime,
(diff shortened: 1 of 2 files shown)


Linchy:
Might I suggest using a colour palette tool for the indicators, so you get a more cohesive look and feel to the tool. This is a good one: http://paletton.com/#uid=51q1o0kAlhyriqAu6lyKHcTTC7w

Here is one I quickly made that gives a same distance adjacent red, green and yellow: http://paletton.com/palette.php?uid=51q1o0kutnjkbvmqVt3AwihMubI

To copy into code, just click on the colour on the right panel and a popup will come up with RGB values, which you can plonk into `Color.FromArgb(140, 7, 0)`.

KonajuGames:
This will help.  Thanks @nkast

nkast:
I just found that it doesn't work with EffectProcessor because now the LogWarning() adds the ": warning : " to the output and EffecrProcessor adds the message "warning X3206: msg".
`P:\\MG Tests\\Test\\Content\\File.fx(31,12): warning : warning X3206: implicit truncation of vector type`
Probably errors have the same problem. 

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs#L211

Issue #4458, opened 2016-01-24
Add Filters To Filter View


tomspilman:
The new filter view in the Pipeline tool should have additional options to allow you to actually filter content.  
- Show/hide skipping.
- Show/hide successful content builds.
- Show/hide warnings.
- ???

Without something like this the filter view is just a more verbose version of the current output window.

PR #5156, opened 2016-09-05
[Pipeline Tool] Pads and Fixes

cra0zy:
Changes:
 - calling build automatically saves the project
 - fix property grid not resizing properly on Linux
 - fix property grid control cells not resizing when resizing propertygrid
 - fix items not expanding in treeview
 - added Pads
   - projectview, propertygrid and buildoutput controls are now pads
   - they each have their own settings, this will help with stuff like #4458
     - settings are just a simple context menu
   - property grid settings contain "sorting" (abc, group)
   - build output settings contain "filtering" and "auto scrolling"
   - main menu "Pads" just points to existing pads settings
   - removed fileter output toolbar item

screenshot from 2016-09-05 18-14-32


energyuser:
There is a bug in the current dev build that makes the Pipeline tool hang and crash. It can be reproduced like this:

-Add multiple assets to the `Content`, for example `SpriteFont`s;
-Make all of them have a different value for `TextureFormat`;
-Use Multiselect on the `SpriteFont`s and try to change `TextureFormat`'s value for all of them.

Tested on Win10 Pro. x64. Does this PR fix the bug?

cra0zy:
> There is a bug in the current dev build that makes the Pipeline tool hang and crash. It can be reproduced like this:
> 
> -Add multiple assets to the Content, for example SpriteFonts;
> -Make all of them have a different value for TextureFormat;
> -Use Multiselect on the SpriteFonts and try to change TextureFormat's value for all of them.
> 
> Tested on Win10 Pro. x64. Does this PR fix the bug?

No... I didn't even know there is a bug about that....open an issue...

tomspilman:
Thanks @cra0zy ... merging!

PR #5158, opened 2016-09-06
Refactor TouchQueue/TouchPanelState and split per platforms.

nkast:
Move event queue inside TouchPanelState. `*GamePlatform` & `*GameWindow` are no longer involved to buffer/re-route events.

The code is split into partial classes for each platform.
TouchPanelState.Default.cs does not buffer events.
~~TouchPanelState.WP.cs is similar to TouchQueue.cs, uses Queue<<T>> instead of List<<T>>.~~
TouchPanelState.WinRT.cs uses the lock-free ConcurrentQueue<<T>>.


nkast:
Inviting @danzel , @KakCAT   for comments.

danzel:
Seems ok.
I would probably just have a single queued implementation of the TouchPanelState with Queue rather than copy/pasting it across the two files.
The `lock()` version should be fine on both platforms.

nkast:
lock is the reason I worked on this in the first place, it's very expensive on clock circles. 
We lock once per frame and every event adds another lock. Also an event might block the game loop or vice versa. Even phones nowadays comes with 2 CPUs and that's a waste.

danzel:
Ok, seems good then :+1:

danzel:
Ok, seems good then :+1:

KakCAT:
Hi,

I can't help much on this :( I stopped working on WP/WinRT at the end of 2015. It's been almost a year since the last time I compiled something on those platforms.

nkast:
No problem @KakCAT, thanks!

PR #5161, opened 2016-09-07
[Pipeline Tool] General Fixes

### LINKED ISSUE #5160: Pipeline crashes when using Multiselect to change values in PropertyGrid
There is a bug in the current dev build that makes the Pipeline tool hang and crash. It can be reproduced like this:

-Add multiple assets to the `Content`, for example `SpriteFont`s;
-Make all of them have a different value for `TextureFormat` (that's where I found the bug but it likely affects the other properties as well);
-Use `Multiselect` on the `SpriteFont`s and try to change `TextureFormat`'s value for all of them.

Tested on Win10 Pro. x64.

cra0zy:
Stuff done:
 - fixed propertygrid crash on Linux/Mac due to incorrect cast of an object
   - new Eto Forms dll changed some combobox stuff so that broke it
 - fixed not being able to change property of multiple items (fixes #5160)
 - fixed closing project not clearing propertygrid
 - fixed right clicking an item on Windows not selecting the current item
 - fixed propertygrid sizing for windows sometimes showing left/right scrollbar
 - improved appearence of items in new item dialog, they now show extension of the template and an icon is no longer displayed
 - improved toolbar, now all items are always visible, with showing/hiding items there was a chance for a misclick

### DIFF
--- a/Tools/Pipeline/Controls/ProjectControl.cs
+++ b/Tools/Pipeline/Controls/ProjectControl.cs
@@ -4,6 +4,7 @@
 
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using Xwt;
 using Xwt.Drawing;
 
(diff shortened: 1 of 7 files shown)


KonajuGames:
Merging.  Thanks @cra0zy

PR #5169, opened 2016-09-12
NuGet package for UWP should reference SharpDX dlls.

azchohfi:


tomspilman:
@DDReaper @BrianPeek - I sort of remember some sort of issue with this...  can you guys verify this is correct?

azchohfi:
This is just the first fix... I'm trying to setup a CI env using Visual Studio Team Services, and I'm having a really hard time... MonoGame.Content.Builder.targets is also not available on a hosted build machine, so what would really help is if MonoGame installer (MonoGameSetup.exe) had a /silent mode. I was not able to do that with the current installer. Is there anything I'm missing?

azchohfi:
I was able to run it on silent mode using /S (needed to be upper case), but the installer requires Administrator credentials, so that's a no-go for Hosted Builds... I'll open an issue for this, since it's not directly related to this PR.

BrianPeek:
If memory serves, the explicit references aren't necessary as the SharpDX assemblies are dependencies on MonoGame.Framework.dll and get copied automatically.  We don't include the references in the Win8/Phone/DX/etc. packages, either.  @azchohfi are you seeing an error or some other functionality not working as expected?  Thanks!

KonajuGames:
From what has been said, this doesn't feel to be the "right" fix for this.

tomspilman:
Yes...  @azchohfi this isn't necessary as the UWP NuGet package already contains them.

image

PR #5174, opened 2016-09-21
avoid the case where normalizing 0,0 results in NaN,NaN and instead default to 0,0

prime31:
minor fix for Vector2.Normalize to avoid it returning NaN,NaN when passed 0,0

### DIFF
--- a/MonoGame.Framework/Vector2.cs
+++ b/MonoGame.Framework/Vector2.cs
@@ -756,9 +756,11 @@ namespace Microsoft.Xna.Framework
         /// </summary>
         public void Normalize()
         {
-            float val = 1.0f / (float)Math.Sqrt((X * X) + (Y * Y));
-            X *= val;
-            Y *= val;
+			var magnitude = (float)Math.Sqrt((X * X) + (Y * Y));
+			if(magnitude > 1E-05f)
+				this /= magnitude;
+			else
+				X = Y = 0;
         }
 
         /// <summary>


MonoGame.Framework/Vector2.cs
```diff
@@ -756,9 +756,11 @@ public static void Negate(ref Vector2 value, out Vector2 result)
         /// </summary>
         public void Normalize()
         {
-            float val = 1.0f / (float)Math.Sqrt((X * X) + (Y * Y));
-            X *= val;
-            Y *= val;
+			var magnitude = (float)Math.Sqrt((X * X) + (Y * Y));
+			if(magnitude > 1E-05f)
```
tomspilman: I know this seems logical to protect against divide by zero...  but this isn't want XNA does.  XNA returns `NaN` if you normalize a zero length vector.  So this change breaks XNA behavior and we cannot merge this as is.

That said... a lot of people make extension methods like `NormalizeSafe` that do that sort of check.

@KonajuGames @dellis1972 - Do you think we should consider adding some sort of "safe" normalization method?  What would be the fallout if it conflicts with people's existing engines that implement extension methods named `NormalizeSafe`?

prime31: Is adhering to XNA really a top priority at this point? Wouldn't making the best possible framework for people to develop with today and in the future be a more important priority? At the very least it would be nice to have a method that "does the right thing" even if it has a slightly odd name. That coupled with a method summary explaining that NaN is a possible return from the method would at least save people some time who aren't coming from XNA and/or haven't run into trying to normalize 0,0 yet.

tomspilman: > Is adhering to XNA really a top priority at this point? 

Yes... especially when it comes to math.  The last thing we want is a physics engine behaving differently on MonoGame because we decided to change the behavior that has been the same for almost 10 years now.

> At the very least it would be nice to have a method that "does the right thing" even if it has a slightly odd name. 

Which is what I suggested.

> That coupled with a method summary explaining that NaN is a possible 

Absolutely...  adding more to the reference docs is always welcome as long as it is written correctly.

prime31:
What are the file and class naming conventions and folder locations for extension methods in MG? I don't see any specifics listed in the contrib guide.

tomspilman:
> What are the file and class naming conventions and folder locations for extension methods in MG?

We wouldn't make it an extension method.  We would add it directly to `Vector2` as a new method that doesn't exist in stock XNA.

PR #5176, opened 2016-09-22
Implement GraphicsDevice.BlendFactor

### LINKED ISSUE #5175: Missing GraphicsDevice.BlendFactor
XNA has a [BlendFactor](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.graphicsdevice.blendfactor.aspx) property in GraphicsDevice that is currently missing in MonoGame. Note that there is also a [BlendFactor](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.blendstate.blendfactor.aspx) property variable in BlendState. I wrote a quick test to see how it works in XNA:

```
[Test]
public void BlendFactor()
{
    Assert.AreEqual(Color.White, gd.BlendFactor);

    var blend = new BlendState
    {
        BlendFactor = Color.Crimson
    };

    gd.BlendState = blend;
    Assert.AreEqual(Color.Crimson, gd.BlendFactor);

    gd.BlendFactor = Color.CornflowerBlue;
    Assert.AreEqual(Color.Crimson, gd.BlendState.BlendFactor);
    Assert.AreEqual(Color.CornflowerBlue, gd.BlendFactor);

    blend.Dispose();
}
```
This passes, so to get the same behavior in MG we just need to add a `public Color BlendFactor { get; set; }` and then in the BlendState setter set `BlendFactor = value.BlendFactor`. I don't know if this property should just override the BlendFactor from the BlendState, but my guess is that's how it works. XNA docs are ambiguous on this and just say both control blend factor...

Jjagg:
Fixes #5175. Details and discussion is in the issue. When handling this I noticed the `force` boolean and _last<RenderState> variables in GraphicsDevice.OpenGL. Do we need those? We already have dirty flags. Or do we have to apply state before the first call to ApplyChanges maybe?

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -1182,6 +1182,20 @@ namespace Microsoft.Xna.Framework.Graphics
             Debug.Assert(_d3dContext != null, "The d3d context is null!");
         }
 
+        private void PlatformApplyBlendFactor()
+        {
+#if WINDOWS_UAP
+			_d3dContext.OutputMerger.BlendFactor =
+				new SharpDX.Mathematics.Interop.RawColor4(
+					BlendFactor.R / 255.0f,
+					BlendFactor.G / 255.0f,
+					BlendFactor.B / 255.0f,
+					BlendFactor.A / 255.0f);
+#else
+			_d3dContext.OutputMerger.BlendFactor = BlendFactor.ToColor4();
+#endif
+        }
+
         internal void PlatformApplyState(bool applyShaders)
         {
             // NOTE: This code assumes _d3dContext has been locked by the caller.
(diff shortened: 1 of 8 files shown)


tomspilman:
Thanks @Jjagg ... merging!

tomspilman:
@Jjagg - Found out something pretty gross about how we're using SharpDX for `BlendFactor`.  See the code here:

https://github.com/sharpdx/SharpDX/blob/master/Source/SharpDX.Direct3D11/DeviceContext.OutputMergerStage.cs#L677

That is the `BlendFactor` setter we use from here:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs#L1195

How i had this implemented it can technically set the same blend state twice.  Once for the blend state part... and once for the blend factor.  This is because SharpDX added this "helper" property which under the hood reapplies the blend state with the new blend factor.

The *right* thing we should be doing at the DX platform code is to stop using the SharpDX `BlendFactor` property and the `BlendState ` property instead use the actual `SetBlendState` method which sets both at once like the api/driver/GPU expects it.

It would require some tweaks to the internal platform API, but would improve the efficiency of the DX implementation.

Jjagg:
Alright, I got some time to spare right now so I'll get to it :)

Issue #2433, opened 2014-04-27
Warning Cleanup


tomspilman:
We could use a pass cleaning up warnings in the build logs.

You just need to look at the [build log from the last successful build](http://teamcity.monogame.net/viewLog.html?buildTypeId=MonoGame_DevelopWin&buildId=lastSuccessful&buildBranch=%3Cdefault%3E&tab=buildLog&guest=1) and search for "warning".

tomspilman:
Looks like we have around 100 warnings that have built back up in the build.  If someone gets the time to clean these up it would be good to get it solved then enable the build failure on warnings.

cra0zy:
This seems... FUN?!? I'm on it.

cra0zy:
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsMetrics.cs#L19

ulong is not CLS compliant, what should be done about this? disable the warning for it? change to something else?

KonajuGames:
> ulong is not CLS compliant, what should be done about this? disable the warning for it? change to something else?

I would suggest changing them to `long`.  If the metrics ever get so high that they overflow a `long`, we have bigger problems.

WardBenjamin:
I would suggest changing them to long. If the metrics ever get so high that they overflow a long, we have bigger problems.

I agree. However, another possible option for future reference in this case is to mark it non-compliant, aka `[CLSCompliant(false)]` above the method header. Not a suggestion, but just for reference.

PR #5177, opened 2016-09-22
Fix some compiler warnings

Jjagg:
Helps out with #2433. Most of these were wrong references in comments and unused variables/parameters. I only built Windows and WindowsGL solutions. The bulk of the leftover warnings for those platforms are about a DrawIndexedPrimitives overload being obsolete in tests, but that overload is what we need to use for XNA and I didn't feel like #ifdeffing around all of them :p Then there's some warnings thrown in NVorbis and code generated by TinyPG that I didn't fix (about CLS-compliance), there's some about GraphicsDevice.BlendFactor missing which will be fixed by #5176. There are some warnings about variables not being assigned, but I left those in and some about variables not being used in tests, but I glanced over and it seemed like they were used for reflection tests.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsMetrics.cs
+++ b/MonoGame.Framework/Graphics/GraphicsMetrics.cs
@@ -9,54 +9,54 @@ namespace Microsoft.Xna.Framework.Graphics
     /// </summary>
     public struct GraphicsMetrics
     {
-        internal ulong _clearCount;
-        internal ulong _drawCount;
-        internal ulong _pixelShaderCount;
-        internal ulong _primitiveCount;
-        internal ulong _spriteCount;
-        internal ulong _targetCount;
-        internal ulong _textureCount;
-        internal ulong _vertexShaderCount;
+        internal long _clearCount;
+        internal long _drawCount;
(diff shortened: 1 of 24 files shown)


MonoGame.Framework/Graphics/GraphicsMetrics.cs
```diff
@@ -9,54 +9,54 @@ namespace Microsoft.Xna.Framework.Graphics
...
-        internal ulong _textureCount;
-        internal ulong _vertexShaderCount;
+        internal long _clearCount;
+        internal long _drawCount;
+        internal long _pixelShaderCount;
+        internal long _primitiveCount;
+        internal long _spriteCount;
+        internal long _targetCount;
+        internal long _textureCount;
+        internal long _vertexShaderCount;
```
tomspilman: I guess this was to avoid some CSL compliance warnings?

Jjagg: Yeah, @KonajuGames suggested this change in https://github.com/MonoGame/MonoGame/issues/2433#issuecomment-110202391

tomspilman:
Thanks @Jjagg !

PR #5180, opened 2016-09-23
[Pipeline Tool] Use attributes if available

cra0zy:
Stuff done:
 - properties will use `DisplayName` if available
 - properties will use `DefaultValue` for the starting value if available
 - properties will not be shown if `Browsable` attribute is set to false
   - under TODO I noticed there was an idea to use `ContentPipelineIgnore` (which does not exist) but we would just be creating a pointless attribute considering there already is an attribute for that exact purpose


tomspilman:
Thanks @cra0zy !

PR #5188, opened 2016-09-25
Fix a crash when loading ADPCM data from a WaveBank.

dellis1972:
In a previous commit the code was refactored. The adjustments
to the blockAlignent was overlooked. As a result when loading
the ADPCM data under openal the blockAlignment is incorrect.
This causes a crash.


nkast:
The blockAlignment here is Int16.
https://github.com/dellis1972/MonoGame/blob/80b9fe0a6bcc0a203168076ad340dcdab2357e90/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs#L149

Same bug as [XAudio] https://github.com/MonoGame/MonoGame/pull/5150.

tomspilman:
Thanks @dellis1972 !

PR #5189, opened 2016-09-26
DdsLoader Write Helper

tomspilman:
A basic uncompressed DDS writing for exporting content for use with different console command line tools.

This could be expanded in the future for other pipeline uses if needed.


KonajuGames:
Merging.

PR #5191, opened 2016-09-26
DDSLoader can now load 32-bit floating point textures.

KonajuGames:
To fix part of http://community.monogame.net/t/32bit-dds-cubemap-to-texturecube/8151

Will need some further refactoring to load 24-bit RGB and 16-bit luminance alpha textures.


tomspilman:
Lets be sure we get some unit tests with floating point formats in here soon.

PR #5194, opened 2016-09-27
Removed the OUYA platform

KonajuGames:


Jjagg:
https://www.youtube.com/watch?v=xTqhyHuKVKA

tomspilman:
@KonajuGames - Changes for XS6 conflicted the default.build file... can you merge/rebase?

DDReaper:
but but but, I still like my Ouya and occasionally run builds to it for fun.  Seems a shame to just remove it especially when it's just an Android variant :S

tomspilman:
> I still like my Ouya and occasionally run builds to it for fun. 

Well older versions of MonoGame are still around that support it.


> Seems a shame to just remove it especially when it's just an Android variant :S

Then can't the current Android support be used for OUYA then?  What would one loose?

mrhelmut:
> What would one loose?

OUYA SDK includes & gamepad support. Nothing that can't be easily worked around (or grabbed from previous source).

They still owe me $250 since 2014, if I ever get to see them, I'll tip them to MG. :D

To be more serious, Razer is going to release new models with new SDKs which will break previous OUYA support anyway. It's going to as fragmented as Android itself, so we're heading toward maintainability problems if we keep every Android platforms. I believe sticking to vanilla Android should be fine enough to support all kind of Android-based machines (inc. NVidia Shield, etc.).

tomspilman:
> we're heading toward maintainability problems if we keep every Android platforms

Right... we can't maintain tons of different Android hardware targets as different platforms.

It seems like the right strategy would be to default to using runtime detection of the Android hardware we're on and making any simple changes needed for that flavor of Android.

If there is another Android target that becomes *very* popular we can decide if it is worth the effort to make a special platform for it.

tomspilman:
@KonajuGames - This has a merge conflict now... can you fix that?

hach-que:
Should probably drop this from the supported list of platforms in Protobuild then...

hach-que:
There doesn't appear to be anything that the Ouya platform does that the Android platform doesn't (in terms of `GenerateProject.CSharp.xslt`), so if support is dropping here I'll remove the explicit platform name upstream as well.

If anyone needs to express an "Ouya" support concept in a Protobuild project, they can do that through [service dependencies](https://protobuild.readthedocs.io/en/latest/service_dependencies.html).

KonajuGames:
Pulled latest into this PR.  Just waiting for builds to pass.

tomspilman:
Pullling the trigger on this...  bye OUYA!

DDReaper:
@KonajuGames you missed one :D
There are still references to Ouya in MonoGame.nsi

Found it while removing WindowsPhone, my goodness these platforms were everywhere

KonajuGames:
The csproj references should disappear after re-generating the projects, but I'll check it.  The OUYA controller is still a valid controller, so it should stay in SDL's gamecontrollerdb.txt.  I'll check the nsi as well.

DDReaper:
I did a search before generating the project while I was searching for WP stuff. Had started with a cleaned repo.
Notepad++'s search in files command is so useful for things like this

mrhelmut:
This PR reverted a recent update to the dependencies submodule. I pushed #5323 to set it back to normal. :)

KonajuGames:
Git submodules strike again.

I'll merge that PR as soon as it passes the tests.

KonajuGames:
@DDReaper The references to OUYA in the Android csproj files are coming from the xslt files internal to Protobuild I believe.  The references to OUYA in those csproj files are nothing harmful or have any effect on the build.  When @hach-que removes OUYA from Protobuild, they will go away.

PR #5198, opened 2016-09-29
Fix Dependencies link

cra0zy:


tomspilman:
The build server is down right now as i'm working on the Mac.... but I will merge this anyway.

@hach-que - It is a little weird people are having trouble with this as the url `https://github.com/Mono-Game/MonoGame.Dependencies.git` normally automatically redirects to the renamed org.  Could be something Protobuild is doing that disallows the redirect to occur.

hach-que:
Hmm, we don't do anything special with Git URLs there; we just call out to the `git` executable to perform cloning or fetching.

That said, it's better just to update the URL.  Unlike Git submodules, it's not a pain when the Git URL of a package changes (`Protobuild.exe --upgrade-all` will take care of it after updating the URL).

PR #5212, opened 2016-10-06
Improve setting BlendState in DX

Jjagg:
We used the BlendState and BlendFactor setters which both called SetBlendState internally. I changed this to check if either BlendState or BlendFactor is dirty and calling SetBlendState from our code so only 1 call is done. Discussed in #5176 
cc @tomspilman


Jjagg:
I think it will build now, I don't have the Win10 SDK installed so fingers crossed :p

tomspilman:
Thanks @Jjagg ....  this seems to be better.

Issue #4387, opened 2016-01-05
[DesktopGL] Effects' MatrixTransform not set by SpriteBatch


mrhelmut:
I have noticed a small discrepancy regarding how vertex shader are applied by ```SpriteBatch```.

The following pass-through effect...

```
float4 PixelShaderFunction(float4 position : SV_Position, float4 color : COLOR0, float2 tex : TEXCOORD0) : COLOR0
{
	return tex2D(s0, tex) * color;
}

void SpriteVertexShader(inout float4 vColor : COLOR0, inout float2 texCoord : TEXCOORD0, inout float4 position : POSITION0) {
	position = mul(position, MatrixTransform);
}

technique Test
{
	pass Pass1
	{
#if SM4
		PixelShader = compile ps_4_0_level_9_1 PixelShaderFunction();
#else
		PixelShader = compile ps_3_0 PixelShaderFunction();
		VertexShader = compile vs_3_0 SpriteVertexShader();
#endif
	}
}
```

... compiles on both WindowsDX and DesktopGL, but it doesn't render on DesktopGL through ```SpriteBatch.Begin(..., Effect, ...)```.

This can be worked around by removing the vertex shader...

```
float4 PixelShaderFunction(float4 position : SV_Position, float4 color : COLOR0, float2 tex : TEXCOORD0) : COLOR0
{
	return color * tex2D(s0, tex);
}

technique Blink
{
	    pass Pass1
	    {
#if SM4
		        PixelShader = compile ps_4_0_level_9_1 PixelShaderFunction();
#else
		        PixelShader = compile ps_3_0 PixelShaderFunction();				
#endif
	    }
}
```

I believe there is something wrong with how DesktopGL sets the MatrixTransform. Maybe related to Z being inverted compared to DirectX?

tomspilman:
So you found my dirty sprite batch secret.  :)

The problem is here:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/SpriteBatch.cs#L52

We only set the "MatrixTransform" parameter on the default sprite `Effect`.  We never query the custom `Effect` passed to `Begin()` for a transform parameter or set it appropriately.

At the time I wrote the code the reason was that I didn't see a way to do it without having lots parameter lookups by string which is not fast.

Really if we just do the lookup once at `Begin()` the impact is only one lookup per-batch.

mrhelmut:
Totally understandable.

So MatrixTransform was never set on custom effects... :)

[...]

PR #5213, opened 2016-10-08
Set "MatrixTransform" effect-property in SpriteBatch.

DizzyHSlightlyVoided:
Fixes issue #4387.


Jjagg:
Did you by any chance check XNA behavior regarding this?

tomspilman:
If XNA does this same thing... automatically setting any uniform with the name "MatrixTransform" to the `SpriteBatch` matrix transform....  then this can be surely merged.  But I suspect XNA doesn't do this.  We need to setup a unit test and see what it shows us.

KimikoMuffin:
Sorry, sorry! I looked at the problem established in #4387, and completely forgot that "trying to replicate XNA's behavior" was one of the main goals. (To say nothing of the fact that I just set "MatrixTransform" without actually checking whether it's a Matrix ...) Lemme try and figure this out ...

tomspilman:
> Sorry, sorry! I looked at the problem established in #4387, took it literally

Yeah if you read the end of that thread the issue was still that we needed to see how XNA behaved and see what it does in various cases.  Like does it change "MatrixTransform" at all?  Does it change the transform if it is called something else?

> I just set "MatrixTransform" without actually checking whether it's a Matrix 

Yeah... I would think if that feature was there we would have to at least check for that.

KimikoMuffin:
OK, here's the results of my testing with XNA:

1. I literally copied the StockEffects SpriteBatch.fx (+ dependencies) into an XNA project, aaaaand ... nothing. The shaded version simply didn't appear.
2. I commented out the vertex shader, and it worked as expected.
3. I uncommented the vertex shader and manually set the "MatrixTransform" on the XNA effect, and it worked.

In other words: XNA doesn't support SpriteBatch vertex-shaders either, and MonoGame's behavior is therefore already currently in line with XNA, right down to the failure-modes as far as I can tell. ;)

So ... what do we do? Intuitively it seems to me that 1. adding vertex-shader support is adding a new feature that did not previously exist, as opposed to changing XNA's behavior or API, and 2. checking for a matrix named "MatrixTransform" is as good a way of doing this as any, and 3. if you copy SpriteBatch.fx and/or other stock effects and so on into your content project, it ought to behave like the stock effect does normally. But I can see how these three propositions are increasingly shaky, and I'd like to wait for feedback before trying to add anything else to the PR.

theZMan:
XNA supports sprite batch with custom vertex shaders... https://blogs.msdn.microsoft.com/shawnhar/2010/04/05/spritebatch-and-custom-shaders-in-xna-game-studio-4-0/ or I've not read this whole issue properly ;-)

KimikoMuffin:
Well, even that BasicEffect example required computing and setting the equivalent of MatrixTransform "manually", which is where the problem lies.

Although as far as that goes, it occurs to me that alternatively (or in addition), the "projection" local-variable could be set on a public property, and let the user sort it out themselves. The complete BasicEffect code would then become something like:

```C#
    basicEffect.World = Matrix.Identity;
    basicEffect.View = Matrix.Identity;

    spriteBatch.Begin(0, null, null, null, null, basicEffect);

    basicEffect.Projection = spriteBatch.Projection;
```

Jjagg:
IMO that's a fair compromise. Users that only do 2D stuff don't have to worry about the projection that way and they can easily use their own effect with spritebatch.

nkast:
Maybe it checks if the Effect is `IEffectMatrices`. Can you check this with the stock SpriteBatch.fx?
add IEffectMatrices to `SpriteEffect` and implement Projection,View,World properties.

KimikoMuffin:
@nkast: Let's see ... Nope, it behaves exactly the same way [as before](https://github.com/MonoGame/MonoGame/pull/5213#issuecomment-252643858): it doesn't render unless I compute and set effect.Projection myself.

nkast:
Did you override the OnApply() method?

something like 
```
        protected internal override void OnApply()
        {           
            matrixTransformParam.SetValue(this.Projection * this.View * this.World);
        }
```

KimikoMuffin:
No, the Projection property was directly setting "MatrixTransform" (yeah, I know, I know ...) Didn't change anything when I did it the proper way, either.

nkast:
Then I guess SpriteBatch doesn't override the .Projection of your Effects. It's up to you to set up the tranformMatrix. 
Actually there is  a sample about batching 3D billboards through SpriteBatch that setup a 3D projection. 
https://blogs.msdn.microsoft.com/shawnhar/2011/01/12/spritebatch-billboards-in-a-3d-world/
That wouldn't work if SpriteBatch were to override the projection.

KimikoMuffin:
Excellent point. Which leaves the possibility of either 1. just leaving it as a public Projection property for those who want to input it manually, or 2. searching only for a matrix named "MatrixTransform" (BasicEffect's equivalent is named "WorldViewProj" ...

... though another possibility just occurred to me while I was typing that: inputting the name of the "matrix transform" property as another parameter to SpriteBatch.Begin(), with a default value null.

tomspilman:
> Then I guess SpriteBatch doesn't override the .Projection of your Effects. It's up to you to set up the tranformMatrix. 

Yeah...  I was fairly certain that was the case.  Thanks to @KimikoMuffin  for verifying that was correct.

If we were to add a feature here it would have to be something the developer explicitly opts-in for.  It cannot be something that just happens because it could screw up effects ported from XNA.

> inputting the name of the "matrix transform" property as another parameter to SpriteBatch.Begin(), with a default value null.

That seems like a reasonable extension.  As long as we add it as the last argument so that it doesn't break existing code... I could support that.

Also I think having it take a `EffectParameter` and not a `string` would be better.  It promotes caching of the `EffectParameter` which is much more efficient that looking it up on every draw.

@KonajuGames @dellis1972 @cra0zy @Jjagg @nkast - Thoughts?

Jjagg:
I'm good with just exposing the projection matrix in SpriteBatch, the extra parameter isn't necessary IMO. If this does get implemented I prefer EffectParameter to a string too.

tomspilman:
> the extra parameter isn't necessary IMO

I think it is... especially if we're having the developer provide a ` EffectParameter` as it is tied to the specific `Effect` passed into `Begin`.

KimikoMuffin:
If we go for a public Projection parameter, the matrix-construction will need to be moved out of Setup() and into Begin(), for the record.

tomspilman:
> If we go for a public Projection parameter

I'm really not a fan of that... especially since `SpriteBatch` doesn't have any other properties at this point.  And while I agree methods with lots of parameters suck...  it is how `SpriteBatch` was designed and the C# named parameter feature makes it much more tolerable.

KimikoMuffin:
Gotcha; my line of reasoning was that you could do other things with it if you wanted, i.e. setting it on the BasicEffect.Projection property. But all right.

tomspilman:
> you could do other things with it if you wanted, i.e. setting it on the BasicEffect.Projection property. 

If I understand you correctly...  the user could still use it like you do with XNA now setting the projection manually on the Effect before calling `SpriteBatch`.

KimikoMuffin:
Yeah, I just thought it would be nice not to have to duplicate the system's existing efforts generating it.

tomspilman:
> duplicate the system's existing efforts generating it

I'm not following you on that one.  There shouldn't be any duplicate work happening here in any case.

KimikoMuffin:
First the SpriteBatch computes the matrix-value which is sent to _spriteEffect's MatrixTransform and possibly _effect.

Then, if the programmer wants access to the same matrix i.e. for use with BasicEffect, which overrides WorldViewProj during OnApply(), the programmer must also compute the matrix-value from scratch.

I meant that it would be convenient if the programmer had some kind of access to the ready-made value instead of having to compute the exact same value all over again. Sorry.

KimikoMuffin:
Oh yeah! What should we do for error checking? I mean, the matrix-parameter *should* belong to the effect, and it won't be meaningful if the effect is null, but SpriteBatch doesn't reportedly throw ArgumentException ... and, for that matter, I can't see how it would specifically hurt the current spritebatch. (For that matter, the only way to check if an EffectParameter belongs to a particular Effect is `if(effectParameter == effect.Parameters[effectParameter.Name])` ...)

I think at the very least, even if it doesn't throw any exceptions, it shouldn't use the matrix-parameter at all if EffectParameter.ParameterClass is not EffectParameterClass.Matrix.

(I did a quick test with `spriteBatch.Begin((SpriteSortMode)34)`, the only other immediately-obviously-wrong kind of argument: In XNA, SpriteBatch.End() subsequently throws an undocumented NotSupportedException, and in Monogame, it behaves identically to SpriteSortMode.Deferred.)

(Also-also, I'm tempted to whip up a [parameter lookup table](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/EffectParameterCollection.cs#L42) while I'm at it, but the only immediately obvious way of doing *that* would be to use a Dictionary, which is kind of a step back from the previous commit ...)

nkast:
> I meant that it would be convenient if the programmer had some kind of access to the ready-made value instead of having to compute the exact same value all over again. Sorry.

It's not always the case that the programmer would use the default OrthographicOffCenter matrix. 

> First the SpriteBatch computes the matrix-value which is sent to _spriteEffect's MatrixTransform and possibly _effect.
> Then, if the programmer wants access to the same matrix [...], the programmer must also compute the matrix-value from scratch.

If that's the case, then we have to skip the computation in the SpriteBatch when used with an Effect (if it's relatively easy to check if the effect has a Vertex Shader). 
I doubt that's a real performance issue and we can see any measurable difference in real games.

KimikoMuffin:
Fair enough, I just have this allergy to redundancy, and I'm fully aware that that's just optimizing my chickens before they hatch. ;)

nkast:
> Fair enough, I just have this allergy to redundancy, and I'm fully aware that that's just optimizing my chickens before they hatch. ;)

I know what you mean! 
The thing is that in a project like MG readability is as important as performance. Truth been told, you will see here code that will never got accepted on business oriented libraries. but you have to show that a change has clear benefits.

nkast:
Just for fun you can write an Effect extension in your code that takes a viewport and sets the params["matrixTransform"]. You can then call `myEffect.SetSpriteBatchProject(viewport);` whenever you need it.
Grab the OrthographicOffCenter code from Matrix.cs for start and then remove all redundant calculations where you have 0 and 1 or -1. Finally, add the halfpixeloffet instead of multiplying with another matrix.

KimikoMuffin:
Awesome.

Anyway, how's this look for adding a parameter?

tomspilman:
@KonajuGames - Do you have an opinion on this one?

@KimikoMuffin - The only thing missing from this is a unit test.  We have to have unit tests on new features at this point else we don't know when we've broken it.

KimikoMuffin:
Gotcha, this'll probably take a bit inasmuch as I'm kind new at unit-tests. Lemme hit up the documentation ...

tomspilman:
@KimikoMuffin - @Jjagg or @KonajuGames could possibly point you in the right direction there.  Usually there is some other unit test which can be copied and modified to fit your needs.  Much easier than figuring things out from scratch.

KimikoMuffin:
Yeah, that's the feeling I'm getting from the relevant doc.

Jjagg:
@KimikoMuffin You can pretty much copy an existing SpriteBatch test and plug in the new parameter. You'll also need a custom effect with a vertex shader There's a test in spritebatch that uses an effect. You can see how it loads the effect [here](https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/Graphics/SpriteBatchTest.cs#L281). The .fx file should be added at https://github.com/MonoGame/MonoGame/tree/develop/Test/Assets/Effects, you don't have to build it manually. Feel free to ask for pointers or help if you run into any issues. You can PM me on Gitter. I'll be a bit busy tomorrow though.

KimikoMuffin:
Yeah, I have a fairly solid idea of what to do after looking over SpriteBatchTest.cs. The one thing I'm not 100% sure about is whether I should add a new test-file, or add it to the existing SpriteBatchTest with `#if !XNA` wrapped around it.

Jjagg:
There are cases like this where we just #if it so I think that's fine for now. It's an easy refactor anyway if it should go to a new file :)

nkast:
I had to test this for a similar issue. #5259 
Maybe you can add to it if I missed something.

nkast:
> Can you add a test with a matrix passed to Begin too?

What for? There allready a test for Matrix transform with SpriteBatch.
The transform is applied only to SpriteEffect.

nkast:
Ok, I see, what you mean. It should probably go to Draw_with_matrix()
I doesn't test custom effect + matrix at the moment.

KimikoMuffin:
Ooh, yes, forgot about that!

I think I'm more inclined towards making a second test to throw into the `#if`; the problem with adding it to an existing test is that this is a new method-parameter, so it wouldn't even compile in the XNA test-project.

nkast:
@KimikoMuffin As the test shows, if you create a custom pixel shader, it will use the vertex shader of SpriteEffect and set/apply the standard transformation to it.
If you want your effect to have a custom vertex shader as well, then I what's the problem to add two extra extra lines before calling SpriteBatch.Begin()? This doesn't justify changing the current API and add a bunch of code and checks in the SpriteBatch class.

nkast:
There is however one case, and that's if you use Immediate and you want SpriteBatch to update the effect projection on every Draw, similarly to how it does for SpriteEffect.  

Here's some alternatives to the effectParameter. 
1- Have your Effect implement an interface like ISpriteProjection { Matrix Projection; }
2- Define an SpriteProjectionAttribute(string TransformParamName = "MatrixTransform") that  suggests that the Effect has a 'MatrixTransform' parameter.

nkast:
This can work the other way too. Telling SpriteBatch to shutdown applying the SpriteBatch completely.

SimonDarksideJ:
Any update on this @DizzyHSlightlyVoided ?  or can this PR be closed now?

DizzyHSlightlyVoided:
Wow, almost eight years. I'm not sure I actually still have the files for this anymore.

I think in its current form it can be closed, yeah. If it's decided that the concept itself is worthwile, a new one can be created (compatible with the actual current generation of Monogame code and standards).

SimonDarksideJ:
Well, the code/changes are still in this PR for reference, but from the discussion it seemed the changes were indeed valuable.
And sorry for the delay, the dog ate my homework :P  With the formation of the MonoGame Foundation, we are trying to accelerate things a bit more these days.

DizzyHSlightlyVoided:
Aha!

Well, I'm not 100% sure I'm in the right General Situation&trade; to take the time to fix it up *right this minute*, and I'll also need to re-familiarize myself with how MonoGame does these things. So this'll probably take a bit of time, if nobody else feels like going to the trouble.

Going just by this discussion, however:

> There is however one case, and that's if you use Immediate and you want SpriteBatch to update the effect projection on every Draw, similarly to how it does for SpriteEffect.
> 
> Here's some alternatives to the effectParameter.
> 1- Have your Effect implement an interface like ISpriteProjection { Matrix Projection; }
> 2- Define a SpriteProjectionAttribute(TransformParamName = "MatrixTransform") which suggests that the Effect has a 'MatrixTransform' parameter.

Both of these ideas have merit, I think; I would personally lean towards `ISpriteProjection` just because it's easier to check without having to delve into reflection (i.e. `if (effect is ISpriteProjection) ...`). In either case, however, we'd need to get it to play nice with the content pipeline; there'd need some kind of `BasicSpriteEffect` class or some such, which is functionally the same as `SpriteEffect` but which takes arbitrary `effectCode` data.

(Perhaps as a class which `SpriteEffect` is derived from, with all of `SpriteEffect`'s functionality moved into `BasicSpriteEffect` such that `SpriteEffect` provides only the byte code, in the name of not reusing identical code between different classes? Or would that break things somehow? I'm not actually sure how it would break things, but it would mean the class inheritence is different from both XNA and from how we're currently doing things, which seems like the kind of thing that might break things for *someone, somewhere*. Perhaps we could give `SpriteEffect` a protected constructor which accepts arbitrary byte code, and derive `BasicSpriteEffect` from that instead ... but that would mean you might get an object which is an instance of `SpriteEffect` but which behaves differently from how `SpriteEffect` currently always behaves ...)

Just some thoughts while I'm procrastinating on other things.

mrhelmut:
Not sure it's worthwhile since it's been stale for so long and got no traction.

PR #5214, opened 2016-10-08
[Pipeline Tool] General Improvements

### LINKED ISSUE #5210: Improve References display in Pipeline tool
Originally mentioned in https://github.com/MonoGame/MonoGame/issues/4457.

image

The 'Collection' text here is completely unhelpful to the user.

 - The 'References' property value should show a  ;  separated string of assembly names.

This makes it more similar to how VS shows lists of files/folders/macros:

image

image

The display could just be the assembly names and not the full path in order to show the most useful information.  The content could be read only and require the use of the editor to make changes to it.

### LINKED ISSUE #5211: Remove/Rename 'Pads' Menu in Pipeline Tool
We should remove or rename the 'Pads' menu item from the Pipeline Tool:

image

It is a duplication of information already available in other areas so it would be fine to remove it.

If we want to keep it we should name it "View" which is more common name for options relating to the different display panels within an application.

### LINKED ISSUE #5215: Pipeline Tool TextureProcessor.ColorKeyColor Broken
It seems like the `TextureProcessor.ColorKeyColor` is broken in the Pipeline Tool.

image

When i open the editor it looks like this:

image

Looking at the code the default is clearly not `#00000000`:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/TextureProcessor.cs#L23

cra0zy:
Stuff done:
 - Better reference editor, fixes #5210
 - Renamed "Pads" menu to "View" menu, fixes #5211 
 - Fix default value of properties, fixes #5215

### DIFF
--- a/Tools/Pipeline/Common/PipelineTypes.cs
+++ b/Tools/Pipeline/Common/PipelineTypes.cs
@@ -300,8 +300,6 @@ namespace MonoGame.Tools.Pipeline
                             browsable = (a as BrowsableAttribute).Browsable;
                         else if (a is DisplayNameAttribute)
                             name = (a as DisplayNameAttribute).DisplayName;
-                        else if (a is DefaultValueAttribute)
-                            defvalue = (a as DefaultValueAttribute).Value;
                     }
 
                     var p = new ProcessorTypeDescription.Property()
--- a/Tools/Pipeline/Controls/PropertyCells/CellBase.cs
+++ b/Tools/Pipeline/Controls/PropertyCells/CellBase.cs
@@ -27,6 +27,7 @@ namespace MonoGame.Tools.Pipeline
         public string DisplayValue { get; set; }
         public string Text { get; set; }
         public bool Editable { get; set; }
+        public int Height { get; set; }
         public Action OnKill;
 
         protected EventHandler _eventHandler;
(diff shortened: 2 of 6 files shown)


tomspilman:
I looks pretty good, but it seems the text size estimate is a little off on Windows:

image

See how the bottom of the `P`s are cut off?

cra0zy:
Actually all the text is shifted a bit down, there is probably an error in Eto.Forms font line height calculations on Windows.

cra0zy:
Yup, I was right: #574

nkast:
@cra0zy 

Can we make error messages red? I think it looked good.
pipelineparserexception

cra0zy:
> Can we make error messages red? I think it looked good.

There is no easy way to do it AFAIK.

nkast:
Ok. Cool. 

Also, Do we need to prefix filenames with "Building ", "Cleaning ", etc?
The filtered view feels interactive with the icons representing the current state (processing, failed, succeed). The icons are more that enough IMO.

tomspilman:
@cra0zy -  I guess we can merge this as is and hope ETO is fixed in the future?

cra0zy:
> Also, Do we need to prefix filenames with "Building ", "Cleaning ", etc?

We do, for people who are running the tool for the first time that text is more useful than trying to guess what is happening from the icon.

> @cra0zy - I guess we can merge this as is and hope ETO is fixed in the future?

Yes, I'll try to do the fix for Eto Forms tomorrow.

cra0zy:
Good news, finally had some free time so I got the bug fixed: https://github.com/picoe/Eto/pull/581

cra0zy:
This should be good for merging now.

Issue #5104, opened 2016-08-13
Fullscreen cannot mode switch


spunky44:
In XNA, it was possible to set a resolution different than the desktop resolution in full screen. The system would "mode switch" and change the resolution.
In MonoGame, this is no longer possible. The game starts as windowed instead of in full screen. However, some of my players want to be able to run at a lower resolution because they have high dpi screens, and have complained about this regression since I migrated from XNA to MonoGame

protected override void Initialize()
{
  DisplayModeCollection modes = GraphicsAdapter.DefaultAdapter.SupportedDisplayModes;

 GraphicsDeviceManager.PreferredBackBufferWidth = 1920;
 GraphicsDeviceManager.PreferredBackBufferHeight = 1080;

  if (GraphicsDeviceManager.IsFullScreen == false)
  {
      GraphicsDeviceManager.ToggleFullScreen();  
  }
  GraphicsDeviceManager.ApplyChanges();

}

Jjagg:
Check out [HardwareModeSwitch](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/GraphicsDeviceManager.cs#L337) in GraphicsDeviceManager. It determines if soft (resize window, don't draw borders) or hardware fullscreen switch is done. 

PR #5096 might also be related here, it adds a manifest to the MG templates so we are dpi aware and.


```
 if (GraphicsDeviceManager.IsFullScreen == false)
 {
     GraphicsDeviceManager.ToggleFullScreen();  
 }
```
You can just do `GraphicsDeviceManager.IsFullScreen = true;` here

spunky44:
Thanks.

I tried setting 
`GraphicsDeviceManager.HardwareModeSwitch = false;`
            
but it made no difference.

spunky44:
I am still not sure if it is possible to mode switch in MonoGame. Is it platform dependent? Note: I am only talking about mode switch at start up, not in-game.
I found some old issues discussing it, but they were also talking about in-game switching (which is something I can live without) so I'm not sure what the status is now.

spunky44:
I debugged the framework and found out that the swap chain is created without Flags being set to AllowModeSwitch. So I set the flag, but unfortunately it had no effect.

spunky44:
My solution has some problems with dual monitors it seems. There is also another case where it does not work for a single monitor.

spunky44:
ResizeBuffers would change the AllowModeSwitch flag, messing everything up.
[...]

Issue #4740, opened 2016-04-07
Windows DX: Window is not centred on startup


MichaelDePiazzi:
When using Windows DirectX, the game window is no longer centred on start up. Windows GL still works correctly.

Does anyone have any ideas on what might have broken this? If not, I will try take a look into this later when I have time.

tomspilman:
My guess is this commit maybe...

 https://github.com/mono/MonoGame/pull/4425

Specifically could be this change...

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Windows/WinFormsGameWindow.cs#L139

I don't think we ever explicitly called any API to center the form window, but maybe we should right after the `Show`.

Also it would be nice to work out a unit test to verify this isn't broken again.  Maybe we can abuse the `GameWindow.Position` and `GameWindow.ClientBounds` to somehow verify this.

tomspilman:
Actually just fired up a test i was working on and it seems the WinDX window is centered correctly for me with the default window sizing.

Are you changing the back buffer size in the ctor or during initialize?

MichaelDePiazzi:
@tomspilman I'm setting the back buffer size in the constructor.

Tzenchor:
I have done a little test in my game and it does not matter if the back buffer is set in the `Constructor` or in `Initialize`, the window position is the same. But when set on `Initialize` you can see the window resizing

MichaelDePiazzi:
@tomspilman Actually, when using the default window sizing it is centred correctly. If you change the window size, then it's not centred properly. So the centring now seems to be based from the default window size instead of the specified window size.

tomspilman:
I figured out the reason for this.

Make the following change to WinFormsGameWindow:

```
        public override Rectangle ClientBounds
        {
            get
            {
                var position = Point.Empty;
                if (_form.Visible)
                    position = _form.PointToScreen(position);
                var size = _form.ClientSize;
                return new Rectangle(position.X, position.Y, size.Width, size.Height);
            }
        }
```

If a call to get the client window came in before the form was visible then the `PointToScreen` would force the window to calculate its start position.  Later you change the window size and since the start position has already been set... it doesn't get fixed to center on show.

tomspilman:
Now that doesn't seem like the right fix... XNA will return the window position even when the window isn't visible yet.  So we really need to change our centering technique to not depend on the start position.

tomspilman:
So another broken thing in this area.

[...]

PR #5218, opened 2016-10-11
WinDX: ClientSizeChanged / backbuffer resizing bug fixes

### LINKED ISSUE #5236: Fullscreen ignoring PreferredBackBuffer values in developer branch v3.6.0.1011
Yesterday I downloaded the last installer for developer branch (v3.6.0.1011) for Windows.

I'm doing a game using DirectX Desktop project and I was using the last stable release version (3.5.1.1679) and in this version the game work as expected, now when switching to the developer's branch installer I get the following issue:

My game is forced to run with this resolution 1024x768 regardless of the Windows screen resolution. My screen resolution is 1920x1080.

Here's the graphics initialization code

```
graphics = new GraphicsDeviceManager(this)
{
    PreferredBackBufferWidth = 1024,
    PreferredBackBufferHeight = 768,
    IsFullScreen = true,
};
```

When I run the game in full screen it's ignoring the PreferredBackBuffer values and run at 1920x1080. Here's a screenshot

developbranch

Note: The vertical lines you see don't appear while playing they appear in the screenshot only.

If I comment the statement _IsFullScreen = true,_  and run the game in a window it works well but if I switch to full screen (while playing the game) I get the same result as the previous screenshot.

This wasn't happening when using the last stable release (master branch) v3.5.1.1679

Here's a screenshot in full screen with the last stable release installer

master

MichaelDePiazzi:
A couple of bugs were introduced by PR #4959 in regards to backbuffer resizing as follows:
- Fullscreen mode was forced to the native resolution. So it was no longer possible to set a different fullscreen resolution.
- If the window was minimised, the backbuffer was resized to 0x0.

The problem was that the `WinFormsGameWindow.OnClientSizeChanged` method was always resizing the backbuffer when the window size changed. This was regardless of whether or not fullscreen mode was enabled, or if the window was minimised.

This PR fixes this by simply skipping the backbuffer resize when fullscreen is enabled or if the window is minimised.

As a side effect of this, I've also cleaned up the `ClientSizeChanged` event. It was previously firing multiple times when switching to/from fullscreen mode. Now it only fires once (which is consistent with XNA behaviour).

P.S. I think this _might_ also fix issue #5104, but I'm not sure.

UPDATE: Confirmed to fix #5236

### DIFF
--- a/MonoGame.Framework/Windows/WinFormsGamePlatform.cs
+++ b/MonoGame.Framework/Windows/WinFormsGamePlatform.cs
@@ -47,6 +47,8 @@ namespace MonoGame.Framework
         {
             var gdm = Game.graphicsDeviceManager;
 
+            _window.EnableClientSizeChangedEvent(false); // Disable ClientSizeChanged event while the window is initialised
+
             _window.Initialize(gdm.PreferredBackBufferWidth, gdm.PreferredBackBufferHeight);
 
             base.BeforeInitialize();
@@ -55,6 +57,8 @@ namespace MonoGame.Framework
                 EnterFullScreen();
             else
                 ExitFullScreen();
+
+            _window.EnableClientSizeChangedEvent(true); // Re-enable (and trigger) ClientSizeChanged event
         }
 
         public override void RunLoop()
@@ -94,14 +98,13 @@ namespace MonoGame.Framework
             {
                 Game.GraphicsDevice.PresentationParameters.IsFullScreen = true;
                 Game.GraphicsDevice.SetHardwareFullscreen();
-
-                _window._form.WindowState = FormWindowState.Maximized;
             }
             else
             {
                 _window.IsBorderless = true;
-                _window._form.WindowState = FormWindowState.Maximized;
             }
+
+            _window._form.WindowState = FormWindowState.Maximized;
 
             _alreadyInWindowedMode = false;
             _alreadyInFullScreenMode = true;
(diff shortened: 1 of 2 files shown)


MichaelDePiazzi:
Looks like something's messed up... All build steps seem to have completed successfully on TeamCity, but the "Build Mac" and "Test Windows" checks have not completed here on GitHub...

MichaelDePiazzi:
I've just pushed another commit which fixes another related issue. If you switched out of fullscreen borderless window mode and changed the resolution at the same time, then the resolution change didn't apply.

As a side effect of this, I've also cleaned up the `ClientSizeChanged` event. It was previously firing multiple times when switching to/from fullscreen mode. Now it only fires once (which is consistent with XNA behaviour).

MichaelDePiazzi:
I just found another related bug. If the window was minimized, it tried to resize the backbuffer to 0x0. I've just pushed another commit fixing this.

MichaelDePiazzi:
Whoops... just realised I broke resizing when in windowed mode! Fixed.

MichaelDePiazzi:
@tomspilman @KonajuGames @dellis1972 This should be good to merge. Thoughts / comments?

Jjagg:
Still two issues when comparing behavior against XNA:

1. The resolution is correct now, but the image is stretched to native resolution while in XNA it only stretches in one direction and you get black bars on the sides (pillar-/letterboxing)
2. When exiting fullscreen the position of the window before fullscreen is not retained. Instead the window jumps to the top left of the screen

Jjagg:
I did some more tests and it seems MG does do pillar- or letterboxing like XNA when the actual resolution is not the native resolution, but it does not choose the same resolution as XNA in all cases (it chose native resolution in my previous test while XNA did not). Another difference is that in fullscreen mode XNA sets the size of the backbuffer to be the display resolution while MonoGame uses the backbuffer width and height in the presentationparameters of the graphicsdevice.

MichaelDePiazzi:
Thanks for testing this @Jjagg!

I tested an older MG release, and this fullscreen pillar/letter-boxing behaviour was still like this. So if we want to get this behaving more like XNA, then I suggest we create a new issue specifically for this.

In regards to the window centring issue, I raised issue #4740 about a similar issue some time ago. @tomspilman was going to sort this out, but I guess he got sidetracked. I'd like to take a look at it also, but it's not a major issue and I haven't had time to look into it. Tom also provided a quick patch which works well enough for my purposes, so I've just been using that for the time being.

Anyway, my main focus with this PR was fixing the regressions introduced by #4959. So I think these other issues should be dealt with separately. That way this PR can be merged as soon as possible to get these regressions resolved.

MichaelDePiazzi:
@tomspilman @KonajuGames @dellis1972: FYI - This has now been tested by @Jjagg and @vcrobe, and all seems to be good in regards to this PR.

MichaelDePiazzi:
FYI - I've deployed this fix in my game on Steam (http://store.steampowered.com/app/459830), and all seems good after a couple of weeks of testing.

dellis1972:
Looks ok to me @KonajuGames ?

KonajuGames:
Looks good to me.  Merging.

PR #5226, opened 2016-10-17
Performance improvements to SpriteBatch.DrawString()

nkast:
spritefont_opt


prollin:
I am surprised that removing ref/out is a perf win here. Just out of curiosity, does this part have a real impact when compared to the original code?
While not pretty, inlining this kind of operation is usually a win since it is not guarantied the compiler will do it for you.

nkast:
@prollin 
>  I am surprised that removing ref/out is a perf win here. Just out of curiosity, does this part have a real impact when compared to the original code?

The first few commits are just to rewrite the code in a more readable form that made it easier to simplify it. I included every step in the process as a commit so you can verify it and I could go back and review it in case the tests failed.

nkast:
@tomspilman I can't find other examples of how we comment a code block as an inlined optimization but I know it's somewhere...
Also I could replace the original code with the cleaner (but unoptimized) intermediate step from `Change mul order`.
https://github.com/nkast/MonoGame/blob/24653c2b3b1791cf5c58bda41dd2b52442f3a1c4/MonoGame.Framework/Graphics/SpriteFont.cs#L303-L311

tomspilman:
> I can't find other examples of how we comment a code block as an inlined optimization 

What you have there looks fine to me.

> Also I can replace the reference code with the cleaner (but unoptimized) intermediate step 

No need...  what you have adequately explains things.


@KonajuGames @dellis1972 @cra0zy - This look good to merge?

cra0zy:
I wouldn't leave in the old code, but that doesn't really matter.

Looks good to me.

nkast:
Somebody tripped on the speakers?

`[MonoGame.Tests.Audio.DynamicSoundEffectInstanceTest.SubmitBuffer_ParameterValidation_SimpleOverload] Test(s) failed. System.NullReferenceException : Object reference not set to an instance of an object.
   `

tomspilman:
> Somebody tripped on the speakers?

I think maybe we had power failure in the office.  It seems I need to login remotely once to get the speakers to work.  I'll fix it here in the next few hours.

nkast:
No pressure.

tomspilman:
Great work @nkast ... merging!

nkast:
Thanks!

Issue #5233, opened 2016-10-21
Pipeline Filter View slows down build process


tomspilman:
I have a content project that takes 6 seconds to build in the regular output view.

When i switch to the filter view the same build takes a crazy **2m 17 seconds**.  Nothing else has changed... the slowdown is consistent when the filter view is enabled.

I suspect how the filter view works is very very slow.... and it happens to be blocking the MGCB.exe output stream keeping it from progressing as fast as it could.

tomspilman:
@cra0zy

james0x0A:
I did some profiling on this issue a couple weeks ago and found that it spends a lot of time measuring and remeasuring the height of each element in the view whenever something is added. It actually starts out at a reasonable speed and grinds to a halt after a hundred items or so.  I wasn't able to track it to anything more specific because of deadlines, but that's the gist of it.

nkast:
How about this @cra0zy ? #5234 
I am not sure if I have to use Eto.Threading or something else.

tomspilman:
Reopening this.

First @nkast fix worked in that build itself finishes just as fast in either the plain or filtered views.

Still while the build finished...  the filtered view still takes up to 2 minutes to complete updates.  You can even see it getting slower as the more and more items are added to it.

I did a test making some obvious optimizations to how the filter view works, i even disabled the auto scroll, and i got ZERO improvement in performance.

The underlying `Xwt.TreeStore` control is horribly slow and completely cripples the filter view feature from being useful as an output window alternative.

@cra0zy - I see three paths:

 1. Replace the Xwt control with the Eto one hoping it is faster.
 2. Try to fix the Xwt control... which might not be possible.
 3. Implement our own fast custom control using Eto for rendering.

tomspilman:
@cra0zy @nkast  - Maybe we should return to having the plain output view as the default until the filter view is fixed.  It currently just makes MonoGame and the pipeline look bad.

KonajuGames:
Does Eto support a virtual treeview?  I used these back in Delphi days (90's) to populate treeviews and listviews with large numbers of items.  The control uses events to ask the application for the information to display on visible items only so less information needs to be copied into the nodes in the control.

nkast:
The control is still populated with items even if it's hidden. 
[...]

PR #5234, opened 2016-10-21
Fix for Pipeline Filter View blocking MGCB.exe output

nkast:
#5233

### DIFF
--- a/Tools/Pipeline/MainWindow.cs
+++ b/Tools/Pipeline/MainWindow.cs
@@ -7,6 +7,8 @@ using System.IO;
 using System.Collections.Generic;
 using System.Diagnostics;
 using Eto.Forms;
+using System.Windows.Threading;
+using System.Threading;
 
 namespace MonoGame.Tools.Pipeline
 {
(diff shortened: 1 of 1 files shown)


tomspilman:
@nkast - Failed on the Linux build:

```
MainWindow.cs(10,22): error CS0234: The type or namespace name `Threading' does not exist in the namespace `System.Windows'. Are you missing an assembly reference?
```

Tools/Pipeline/MainWindow.cs
```diff
@@ -197,7 +199,10 @@ public void UpdateProperties()
 
         public void OutputAppend(string text)
         {
-            Application.Instance.Invoke(() => buildOutput.WriteLine(text));
+            if(Dispatcher.FromThread(Thread.CurrentThread)==null)
+                Application.Instance.AsyncInvoke(() => buildOutput.WriteLine(text));
```
tomspilman: Would it not be ok to just always async invoke?  Is there a big performance hit if we do that?

tomspilman: Would it not be ok to just always async invoke?  Is there a big performance hit if we do that?

tomspilman: Would it not be ok to just always async invoke?  Is there a big performance hit if we do that?

tomspilman: Would it not be ok to just always async invoke?  Is there a big performance hit if we do that?

nkast: I tried to leave the current behavior unchanged just in case...
I think Eto uses Eto.Threading.Thread.IsMain but that throws an exception. 
I'll just call AsuncInvoke for now and wait for @cra0zy 's feedback on this one.

cra0zy:
👍

tomspilman:
Thanks @nkast and @cra0zy !

PR #5239, opened 2016-10-23
Populate FilterView with assets before build and map Nodes to Assets.

nkast:
***Build***

build

***Rebuild***

rebuild

### DIFF
--- a/Tools/Pipeline/Controls/BuildOutput.cs
+++ b/Tools/Pipeline/Controls/BuildOutput.cs
@@ -97,15 +97,15 @@ namespace MonoGame.Tools.Pipeline
                     PopulateAssets();
                     break;
                 case OutputState.Cleaning:                    
-                    AddItem(_output, _iconInformation, "Cleaning " + PipelineController.Instance.GetRelativePath(_output.Filename));
+                    AddItem(_output, _iconInformation, "Cleaned " + PipelineController.Instance.GetRelativePath(_output.Filename));
                     AddItem(line);
                     break;
                 case OutputState.Skipping:
-                    AddItem(_output, _iconSkip, "Skipping " + PipelineController.Instance.GetRelativePath(_output.Filename));
+                    AddItem(_output, _iconSkip, "Skipped " + PipelineController.Instance.GetRelativePath(_output.Filename));
                     AddItem(line);
                     break;
                 case OutputState.BuildAsset:
-                    AddItem(_output, _iconProcessing, "Building " + PipelineController.Instance.GetRelativePath(_output.Filename));
+                    AddItem(_output, _iconStart, "Building " + PipelineController.Instance.GetRelativePath(_output.Filename));
                     AddItem(line);
                     break;
                 case OutputState.BuildError:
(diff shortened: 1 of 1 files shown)


cra0zy:
I'm against this, I am working on a better filterview that is custom drawn (which will be done in a few days) and this will just create a bunch of pointless merge conflicts for me...

nkast:
Maybe we should swap the icons of pending (gears) / building (clock) ...
Same with the texts, keep the "Building" for the working task and restore the filename for finished assets.

nkast:
> this will just create a bunch of pointless merge conflicts for me...

Oh, come on. That touched just one file, the one that's supposed to be replaced.

cra0zy:
> Oh, come on. That touches just one file, the one that's supposed to be replaced.

Considering my PR will be ready in a few days this PR will sever almost no purpose ¯\_(ツ)_/¯

nkast:
I swapped the icons and changed the text.

Build
build2

Rebuild
rebuild2

nkast:
@cra0zy 
If that fixes the perf issues with #5233 do you really have to spent days to provide yet another TreeView over Eto?

cra0zy:
> If that fixes the perf issues with #5233 do you really have to spent days to provide yet another TreeView over Eto?

Yes, because it will be snappy fast, plus I can finally implement #4458 (problem with implementing it was the treeview refresh rate for lot of items)

Also I don't really like how you made it look like :(

nkast:
> Also I don't really like how you made it look :(

That's unfortunate because I really wanted to present the concept. The code  I don't care about, this PR is more like a prototype. As it is the FilterView now looks like a colorful version of the output view. 
I hope you take a second look at it and adopt some ideas into your control.

Jjagg:
I like the icon and text changes in this PR :) Makes more sense IMO

tomspilman:
@nkast @cra0zy - I do like the idea in this PR:

Instead of showing the same item in the list multiple times for each state it can be in...  we show each item once and just change its state as different events occur on it.

This immediately halves the amount of items in the list which isn't a bad thing for the performance of the shitty tree view we have now.   It is also nicer for the to user to read.

The only thing is @nkast current implementation does weird stuff at times with the autoscroll...  I see it flickering between the bottom and the middle of the list while it is building.  Seems best for the autoscroll to only ever scroll down and never back up the list.

Also @nkast for some reason I ended up with some "Cleaned" items still in the list in my build... so maybe it isn't working correctly yet in all cases.

So I think this isn't a bad intermediate step here... if @nkast makes some fixes I wouldn't have an issue merging it.  Yet I think the new custom control that @cra0zy is working on will be the next step.  I would just like to see it use the concept of what @nkast did here as it is a good one.

cra0zy:
> Instead of showing the same item in the list multiple times for each state it can be in... we show each item once and just change its state as different events occur on it.

There is one thing that I find nightmarish to implement and that is autoscroll :/

Also the new custom control is almost complete, I am literally doing the finishing touches:
capture

tomspilman:
> There is one thing that I find nightmarish to implement and that is autoscroll :/

 Yeah it is never fun to implement, but we do need it.  As long as it doesn't ever jump back up the list then it is generally ok.

 However I haven't yet gotten into the other issue with our current autoscroll...  when to turn it on and off.  If the user manually scrolls the list or starts text selection.... the autoscroll should stop.  If the user scrolls back to the bottom of the list (drags scrollbar, repeated page down keys, the end key) then autoscroll should be re-enabled.


> Also the new custom control is almost complete, I am literally doing the finishing touches:

Great to hear! Still I would like to see the feature @nkast developed here working with the new control at some point.

cra0zy:
PR is up #5258

nkast:
> Also @nkast for some reason I ended up with some "Cleaned" items still in the list in my build... so maybe it isn't working correctly yet in all cases. 

I am getting it too on 4 or so items.  I didn't look into that, I wanted to get some feedbacks first. 
It's tricky to map MGCB filenames into what we got internally. We have to work with relative paths, backshash vs forwardshash and do that for .XNBs too. Intimately we have to associate mgcb output with  ProjectItem and the TreeView & FilterView.

> current implementation does weird stuff at times with the autoscroll... I see it flickering between the bottom and the middle of the list while it is building. Seems best for the autoscroll to only ever scroll down and never back up the list. 

It has to scroll twice, once during cleaning and another when building. Probably those few rogue  items that miss the parsing cause it to jump to the end of the list, and back again.

tomspilman:
@nkast - Now that #5258 is up we should focus on it, but I still want to use your concept of only having one item per file instead of having one for cleaning and another for skipping/building.

nkast:
> @nkast - Now that #5258 is up we should focus on it, but I still want to use your concept of only having one item per file instead of having one for cleaning and another for skipping/building.

Sure, I stopped working on it since @cra0zy started on the new Control.

PR #5244, opened 2016-10-26
Performance improvements to SpriteBatch with custom Effect

nkast:
- Skip Setup of _spritePass when we have a custom Effect.
- Single pass custom Effects are applied During Setup(), same as
_spritePass. Effects with multiple passes are applied by
SpriteBatcher.FlushVertexArray() as before. (ImmediateMode applies the
effect on every .Draw() in that case).


tomspilman:
Most custom XNA/MonoGame custom effects used with `SpriteBatch` only implement the pixel shader.  This is why we always apply the default sprite effect first.

The only way to be able to correctly skip the default sprite effect is to *know* that the custom effect implements both a vertex and pixel shader.

nkast:
Oh, I forgot that.

PR #5246, opened 2016-10-26
Fix BoundinBox.Contains(Vector3) to match XNA

### LINKED ISSUE #5245: BoundingBox.Contains(Vector3) returns incorrect ContainmentType
The code of BoundingBox.Contains(Vector3) method checks if any of the points is on the edge of the bounding box and immediately returns `ContainmentType.Interjects` even if the given Vector3 is completely inside it.

The method should return `ContainmentType.Contains` if all of the points are inside and `ContainmentType.Interjects` if any of them are inside.

Jjagg:
Fixes #5245. XNA returns Contains when a vector is on one of the boxes faces/edges so MG should too.

### DIFF
--- a/MonoGame.Framework/BoundingBox.cs
+++ b/MonoGame.Framework/BoundingBox.cs
@@ -214,16 +214,11 @@ namespace Microsoft.Xna.Framework
                 || point.Z > this.Max.Z)
             {
                 result = ContainmentType.Disjoint;
-            }//or if point is on box because coordonate of point is lesser or equal
-            else if (point.X == this.Min.X
-                || point.X == this.Max.X
-                || point.Y == this.Min.Y
-                || point.Y == this.Max.Y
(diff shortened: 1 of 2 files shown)


james0x0A:
This matches my testing

tomspilman:
@Jjagg - You ran these unit tests on XNA too?

Jjagg:
Yes sir

KonajuGames:
The unit tests are good, but I think they miss some coverage.  There are no tests that use a vector component of 1.0.  The min bounds of the box are tested, but no tests cover the max bounds of the box.

Jjagg:
I'll add those tomorrow :)

Jjagg:
Whatever that means in an international context

tomspilman:
Thanks @Jjagg & @hifi !

PR #5247, opened 2016-10-27
Fix performance hit when Immediate is used with a custom Effect.

nkast:
SpriteBatcher.FlushVertexArray() reapplies the custom Effect on every Draw.

Frame rate drops from 80fps to 65fps in my test.
This PR apply the custom Effect the same way we do for SpriteEffect.
The only expeption is Effects with multiple passes.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -147,6 +147,16 @@ namespace Microsoft.Xna.Framework.Graphics
 
             _matrixTransform.SetValue(projection);
             _spritePass.Apply();
+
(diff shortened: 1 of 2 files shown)


nkast:
It applies to SpriteSortMode.Texture as well which calls FlushVertexArray() many times during a Batch.

tomspilman:
@nkast - How does this affect cases where one uses `SetValue` to set custom parameters on the custom `Effect` within `SpriteBatch.Begin/End()`?

nkast:
> How does this affect cases where one uses SetValue to set custom parameters on the custom Effect within SpriteBatch.Begin/End()?

Good question! 
 #5259 
diff-draw_with_viewport_changing_immediate_

PR #5248, opened 2016-10-27
Fix: IndexBufferWriter doesn't detect the correct IndexElementSize.

nkast:
IndexElementSize depends on the number of Vertices, not the number of indices.


tomspilman:
Good catch...  thanks @nkast !

PR #5253, opened 2016-10-27
Prevent duplicates in ModelMesh.Effects

### LINKED ISSUE #5252: ModelMeshPart Effect adding difference between MonoGame and XNA
I have this (very ugly and dirty) piece of code:

```
	
    // backup current effects
	i=0;
	foreach (ModelMeshPart p in mesh.MeshParts)
	{
		originalEffect[i++]=p.Effect;
		
		if (i==2) p.Effect=envEffect2;
		else p.Effect=envEffect1;
	}

	// draw
	mesh.Draw();

	// restore effects
	i=0;
	foreach (ModelMeshPart p in mesh.MeshParts)
	{
		p.Effect=originalEffect[i++];
	}
```

Basically I have a car with basic effect, and I change all the Basic Effects for Environment Effects in the menu screen to show it with environment.
After drawing, I restore the effects.

This works well with XNA, but there's a side effect in MonoGame. The number of effects inside mesh.MeshParts.Effect keeps increasing one each frame when setting p.Effect = something

It seems to me that when setting the effect, in MonoGame the effect is always added to the Effect pool, but in XNA it is only added when that effect is not present in the pool. So, using
``` 
if (!parent.Effects.Contains (_effect)) parent.Effects.Add(value);
```
in ModelMeshPart.Effect setter seems to fix the problem, although I don't know enough of MG to say if this will imply side-effects.

jamesford42:
this resolves #5252

### DIFF
--- a/MonoGame.Framework/Graphics/ModelMeshPart.cs
+++ b/MonoGame.Framework/Graphics/ModelMeshPart.cs
@@ -38,7 +38,9 @@ namespace Microsoft.Xna.Framework.Graphics
 
                 // Set the new effect.
                 _effect = value;
-                parent.Effects.Add(value);
+                
+                if (_effect != null && !parent.Effects.Contains(_effect))                
+                    parent.Effects.Add(_effect);
             }
         }
 


nkast:
There is abetter way to fix that. No need to maintain a real collection. The effectsenumerator is actually an enumerator of Mesh.MeshParts. Instead of MeshPart it returns its Effect.

tomspilman:
Thanks @jamesford42 !

PR #5264, opened 2016-10-29
[Pipeline Tool] Optimize ProjectView image loading

cra0zy:
This PR makes it so projectview icons get cached on load, so for example if it loads an image of a file with .png extension, next time it comes across a file with that extension it will use a cached version of the image instead of loading a whole new one.

This should make Pipeline Tool use a lot less memory for big projects.

### DIFF
--- a/Tools/Pipeline/Global.Windows.cs
+++ b/Tools/Pipeline/Global.Windows.cs
@@ -26,6 +26,21 @@ namespace MonoGame.Tools.Pipeline
         {
             var reg = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Windows NT\CurrentVersion");
             IsWindows10 = (reg.GetValue("ProductName") as string).StartsWith("Windows 10");
+
+            var file = ExtractIcon(0).ToBitmap();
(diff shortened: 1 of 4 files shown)


tomspilman:
Thanks @cra0zy !

tomspilman:
@cra0zy - All the folders in the project view now have red Xs on them after this PR.

Tools/Pipeline/Global.Windows.cs
```diff
@@ -26,6 +26,21 @@ private static void PlatformInit()
...
+            var folderMissing = ExtractIcon(234).ToBitmap();
+
+            _files["."] = ToEtoImage(file);
+            _fileMissing = ToEtoImage(fileMissing);
+            _folder = ToEtoImage(folder);
+            _folderMissing = ToEtoImage(folderMissing);
+
+            _xwtFiles["."] = ToXwtImage(file);
+            _xwtFileMissing = ToXwtImage(fileMissing);
+            _xwtFolder = ToXwtImage(folderMissing);
```
cra0zy: @tomspilman Ups, seems that I loaded the wrong image here.... easy fix.

cra0zy:
#5266

PR #5265, opened 2016-10-30
[Pipeline Tool] OutputParser Optimizations

cra0zy:
Stuff done:
- Reorganizes actions so that more common actions are checked first
- Replaces Regex for Skipping/Cleaning with StartsWith
- Minor coding style changes so that it matches the rest of the project
- Separate thread for filter output

### DIFF
--- a/Tools/Pipeline/Common/IController.cs
+++ b/Tools/Pipeline/Common/IController.cs
@@ -25,6 +25,8 @@ namespace MonoGame.Tools.Pipeline
 
         PipelineProject ProjectItem { get; }
 
+        List<string> BuildOutput { get; }
+
         /// <summary>
         /// True if there is a project.
         /// </summary>
--- a/Tools/Pipeline/Common/IView.cs
+++ b/Tools/Pipeline/Common/IView.cs
@@ -58,9 +58,9 @@ namespace MonoGame.Tools.Pipeline
 
         void UpdateProperties();
 
-        void OutputAppend(string text);
-
         void OutputClear();
+
+        void OutputThreadStart();
 
         bool ChooseContentFile(string initialDirectory, out List<string> files);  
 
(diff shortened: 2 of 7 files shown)


tomspilman:
Thanks @cra0zy ... this should be a ton faster.  I'll take a look now.

cra0zy:
> Thanks @cra0zy ... this should be a ton faster. I'll take a look now.

Do tell me the results, I have an idea for one extra optimization that I will try tomorrow.

tomspilman:
> Do tell me the results

With a build I see around 4 to 5 seconds slower than the time reported by MGCB.

With a clean the filter output is around 17 seconds slower than the time reported by MGCB.

So still much slower that I would like to see it be.  Is it possible that the invalidate you're doing is forcing a redraw during that call and isn't just marking the window dirty for redraw later?

You might need to run a profiler on this to figure out the real bottlenecks here.

tomspilman:
@cra0zy - Hang on.... I think I tested this wrong.

tomspilman:
@cra0zy - So I did test wrong...  it is better than what I reported before, but still not as good as it should be:

About 2 to 3 seconds slower in a clean build or a rebuild of content.

Around 9 seconds slower cleaning content.

cra0zy:
https://github.com/MonoGame/MonoGame/blob/develop/Tools/Pipeline/Common/PipelineController.cs#L508

We just need to make it not call the function directly but instead add a line to the buffer.

cra0zy:
In case you are awake and want to test out my 2nd idea of optimization I pushed the commit. It's not finished tho, but the end speed result will be that.

cra0zy:
It should be done now.

cra0zy:
Just gonna add that I think I optimized the textview build output as well.

tomspilman:
@cra0zy - This branch is conflicted... if you can fix it I can test it.

cra0zy:
Done.

cra0zy:
@tomspilman Did you get the chance to test these changes?

cra0zy:
I'm gonna temporarily close this PR, I'll resubmit it once few other parts get merged (this is so I don't have to worry about merge conflicts).

PR #5267, opened 2016-10-31
Performance improvements to SpriteBatch 

nkast:
-Skip Calculations when Origin == Zero.
-Add _texelSize in Texture2D to avoid divisions in SpriteBatch.
 Used in calculating  origin & textureCord from sourceRectangle.

Current develop (Baseline)
33.83ms (100%)
Avoid extra calculations when Origin = Zero
30.38ms (89.8%)
Add _texelSize in Texture2D. Replace DIVs with MULs.
29.34ms (86.7%)

*test*
```
 var amount = 100000;
            var screenSize = 800;
            var spriteSize = 64;
            Rectangle textRect = new Rectangle(0,0,64,64);
            for (int i = 0; i < amount; i++)
            {
                var dest = new Rectangle(random.Next(screenSize), random.Next(screenSize), spriteSize, spriteSize);
                spriteBatch.Draw(_texture, dest, textRect, Color.White, random.Next(360), Vector2.Zero, SpriteEffects.None, 1);                
            }
```

### DIFF
--- a/Installers/Linux/RUN/postinstall.sh
+++ b/Installers/Linux/RUN/postinstall.sh
@@ -9,15 +9,15 @@ echodep()
 	do
 		line="$line."
 	done
+	
+	echo -ne "$line"
     
 	if eval "$2"
 	then
-		line="$line\e[32m[Found]\e[0m"
+		echo -e "\e[32m[Found]\e[0m"
 	else
-		line="$line\e[31m[Not Found]\e[0m"
+		echo -e "\e[31m[Not Found]\e[0m"
 	fi
-	
-	echo -e "$line"
 }
 
 # Check installation priviledge
@@ -29,13 +29,32 @@ fi
 DIR=$(pwd)
 IDIR="/usr/lib/mono/xbuild/MonoGame/v3.0"
 
+# Find MonoDevelop
+MDTOOL="?????"
+
+if type "monodevelop" > /dev/null 2>&1
+then
+	if eval "monodevelop --help | grep 'MonoDevelop 6' > /dev/null 2>&1"
+	then
+		MDTOOL="mdtool"
+	fi
+fi
+
+if type "monodevelop-stable" > /dev/null 2>&1
+then
+	if eval "monodevelop-stable --help | grep 'MonoDevelop 6' > /dev/null 2>&1"
+	then
+		MDTOOL="mdtool-stable"
+	fi
+fi
+
 # Show dependency list
 echo "Dependencies:"
 echodep "mono-runtime" "type 'mono' > /dev/null 2>&1"
 echodep "gtk-sharp3" "type 'gacutil' > /dev/null 2>&1 && gacutil /l gtk-sharp | grep -q 3.0.0.0"
 echo ""
 echo "Optional Dependencies:"
-echodep "MonoDevelop 6" "monodevelop --help | grep 'MonoDevelop 6' > /dev/null 2>&1"
+echodep "MonoDevelop 6" "$MDTOOL > /dev/null 2>&1"
 echodep "Rider" "type 'rider' > /dev/null 2>&1"
 echodep "referenceassemblies-pcl / mono-pcl" "test -d /usr/lib/mono/xbuild/Microsoft/Portable"
 echodep "ttf-mscorefonts-installer / mscore-fonts" "fc-list | grep -q Arial"
@@ -96,10 +115,10 @@ then
 fi
 
 # MonoDevelop addin
-if monodevelop --help | grep 'MonoDevelop 6' > /dev/null 2>&1
+if [ "$MONODEVELOP" != "?????" ]
 then
 	echo "Installing MonoDevelop Addin..."
-	sudo -H -u $SUDO_USER bash -c "mdtool setup install -y $DIR/Main/MonoDevelop.MonoGame.mpack  > /dev/null"
+	sudo -H -u $SUDO_USER bash -c "$MDTOOL setup install -y $DIR/Main/MonoDevelop.MonoGame.mpack  > /dev/null"
 fi
 
 # Monogame Pipeline terminal commands
--- a/MonoGame.Framework.Content.Pipeline/Audio/DefaultAudioProfile.cs
+++ b/MonoGame.Framework.Content.Pipeline/Audio/DefaultAudioProfile.cs
@@ -278,14 +278,14 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Audio
             try
             {
                 string ffmpegCodecName, ffmpegMuxerName;
-                int format;
+                //int format;
                 switch (formatType)
                 {
                     case ConversionFormat.Adpcm:
                         // ADPCM Microsoft 
                         ffmpegCodecName = "adpcm_ms";
                         ffmpegMuxerName = "wav";
-                        format = 0x0002; /* WAVE_FORMAT_ADPCM */
+                        //format = 0x0002; /* WAVE_FORMAT_ADPCM */
                         break;
                     case ConversionFormat.Pcm:
                         // XNA seems to preserve the bit size of the input
@@ -297,13 +297,13 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Audio
                         else
                             ffmpegCodecName = "pcm_s16le";
                         ffmpegMuxerName = "wav";
-                        format = 0x0001; /* WAVE_FORMAT_PCM */
+                        //format = 0x0001; /* WAVE_FORMAT_PCM */
                         break;
                     case ConversionFormat.WindowsMedia:
                         // Windows Media Audio 2
                         ffmpegCodecName = "wmav2";
                         ffmpegMuxerName = "asf";
-                        format = 0x0161; /* WAVE_FORMAT_WMAUDIO2 */
+                        //format = 0x0161; /* WAVE_FORMAT_WMAUDIO2 */
                         break;
                     case ConversionFormat.Xma:
                         throw new NotSupportedException(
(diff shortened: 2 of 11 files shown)


nkast:
Add my 'standard' DrawString() test
```

            for (var i = 20; i < 480; i += 20)
                this.spriteBatch.DrawString(this.fpsCounter.spriteFont, "abcdefghijklmnopqrstuvwxyz1234567890 some words here on the end test amazing testing letters to the edge of the screen", new Vector2(0, i), Color.Black);            
```

Current develop (Baseline)
1694fps 0.48ms (100%)
This PR
1823fps 0.44ms (91.7%)

Vodacek:
Why not make TexelSize public?

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -93,6 +97,7 @@ protected Texture2D(GraphicsDevice graphicsDevice, int width, int height, bool m
             this.GraphicsDevice = graphicsDevice;
             this.width = width;
             this.height = height;
+            this._texelSize = new Vector2(1f / (float)width, 1f / (float)height);
```
tomspilman: @KonajuGames @dellis1972 - Any thoughts on this?

Since all the texture types are pretty much immutable when it comes to the size and type of texture data...  I don't find it all that bad to keep something like `TexelSize` as a helper.  It isn't so different than the sort key we added to `Texture`:

https://github.com/nkast/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture.cs#L27

The only thing I would like to see fixed if we add this is checking of the input arguments for `Texture2D`.  As the code is now you can pass 0 and get a divide by zero in this new code.  I suspect that `Texture2D` actually validates the width and height as well as the other inputs.

KonajuGames: It looks like we're relying on the graphics API to fail if the dimensions are zero, so to avoid a divide by zero we should validate width and height > 0 before using them in the Texture2D constructor.

tomspilman: > It looks like we're relying on the graphics API to fail if the dimensions are zero

Yeah but I bet XNA doesn't do that.

nkast: > As the code is now you can pass 0 and get a divide by zero in this new code. 

That was true in old code as well. The division moved from SpriteBatch.Draw() into Texture2D().
https://github.com/MonoGame/MonoGame/pull/5267/files#diff-bf9a3678ade2b898dde0ac4d0f4a83cdL391

XNA throws ArgumentOutOfRangeException
`Resource size must be greater than zero.
Parameter name: height`

I guess we need a pattern similar to https://github.com/MonoGame/MonoGame/pull/5121.
to check Texture2D/3D/Cube.

nkast: @Vodacek, 
As for making TexelSize public, what's the need for it? 
How will it work with 'ResizedToPowerOf2' textures?
Is `Vector2` (.X/.Y) good enough for a public API or should we introduce a SizeF structure (.Width/.Height)?

nkast: @tomspilman 
>The only thing I would like to see fixed if we add this is checking of the input arguments for Texture2D. As the code is now you can pass 0 and get a divide by zero in this new code. I suspect that Texture2D actually validates the width and height as well as the other inputs.

fixed in #5291

nkast:
Pushed another commit to move origin rotation out of SpriteBatchItem..Set().
Some perf benefit there as well as we do the calculations once.

MichaelDePiazzi:
@nkast - I've been messing around with optimising `SpriteBatch` a bit lately, and I came up with a few changes that you might like to include in your PR. Of course I'm happy to submit them myself also if you don't want to include them. But in that case I'll wait until this PR is merged to avoid conflicts.

https://github.com/MichaelDePiazzi/MonoGame/compare/finaldays...MichaelDePiazzi:fdSpriteBatchOpt?w=1

Summary of changes:

- Moved the logic from `SpriteBatch.DrawInternal` to `SpriteBatchItem.Set` _(goal is to make inlining in `SpriteBatch.Draw` methods simpler as per @tomspilman's proposal in #5347)_
- Updated `Draw` methods to call `SpriteBatchItem.Set` directly
- Changed `SpriteBatchItem.Set` to take texture coordinates instead of the source rectangle
- Changed `SpriteBatchItem.Set` to take large struct parameters (i.e. >32-bit) by reference
- Changed `SpriteFont` to pre-calculate glyph texture coordinates when the font is constructed **(this gave a quite good performance increase!)**
- Simplified rotation calculations in `SpriteBatchItem.Set`
- Started inlining some of the calculations in the `SpriteBatch.Draw` methods

And here's some benchmarks:

Calling `SpriteBatch.Draw` 10,000,000 times | Baseline | Optimised | Comparison
--- | --- | --- | ---
**Light**<br>(without source rectangle, origin, rotation, or flipping) | 800ms | 779ms | 2.6% less time<br>2.7% faster
**Heavy**<br>(with source rectangle, origin, rotation, and flipping) | 1715ms | 1644ms | 4.1% less time<br>4.3% faster

Calling `SpriteBatch.DrawString` 100,000 times | Baseline | Optimised | Comparison
--- | --- | --- | ---
**Light**<br>(without origin, rotation, or flipping) | 693ms | 471ms | 32.0% less time<br>47.1% faster
**Heavy**<br>(with origin, rotation, and flipping) | 1263ms | 1087ms | 13.9% less time<br>16.2% faster

nkast:
@MichaelDePiazzi 
Proposal #5347 is about inlining and then simplifying each method. I am trying to move most unrelated calculations out of .Set() for this reason. 
By moving the code from DrawInternal into Set() we are back on square one. For example, DrawString(...) need to set SortKey and sin/cos once. Also most Draw(...) methods don't need rotation, it's best to keep the two overloads. 

Which Draw() did you use for testing? Passing position (Vector2) by ref will help with Draw(Texture2D texture, Vector2 position, ...) but in the case of Draw(Texture2D texture, Rectangle dest, ...) requires to create a new `Vector2`.

When I changed texCoordTL/texCoordBR to `ref` I saw a small improvement ~2%-1% (which is within the margin of error with modern OS, I killed almost all services and did several runs to ensure there was a difference.) Later I tested https://github.com/MonoGame/MonoGame/compare/develop...nkast:tncSpriteBatchDraw with this PR and found that for the light Draw() I had to initialize _texCoordTL/BR with Vector2.Zero/One or use a member _texCoordZero/_texCoordOne which I found kind of ugly.
Perhaps I should drop the `ref` commit from this PR and re-evaluate. 
We can first test your proposal to use [MethodImpl(MethodImplOptions.AggressiveInlining)] on `.Set()` methods.
There's an oportunity to set directly `vertexTL/BR.TextureCoordinate = texCoordTL/BR` in two out of 4 cases, which some Jit frameworks might optimize better that `vertexTL/BR.TextureCoordinate.X = texCoordTL/BR.X; vertexTL/BR.TextureCoordinate.Y = texCoordTL/BR.Y` and we can't do that if texCoord is a `Vector4`.
For example commit https://github.com/MonoGame/MonoGame/pull/5267/commits/5784e8c7e9e41616048343b341c7c10019d4e100 was a bit faster.

> Changed SpriteFont to pre-calculate glyph texture coordinates when the font is constructed (this gave a quite good performance increase!)

That's impresive. I think we should apply this right after we inline DrawInternal() into DrawInto(), it's a clear win in performance.
After that we can simplify DrawInto() and finally inline it into DrawString().

MichaelDePiazzi:
> By moving the code from DrawInternal into Set() we are back on square one.

Yes, I see what you mean. I have gone down the wrong path with this.

> Which Draw() did you use for testing?

I used `Draw(texture, position, sourceRectangle, color, rotation, origin, scale, effects, layerDepth)`. And you are right, the overload with the `destinationRectangle` parameter required creating of 2 new `Vector2`'s actually (but it currently is creating a new `Vector4` anyway).

> Perhaps I should drop the ref commit from this PR and re-evaluate.

I also wondered if passing by ref was worthwhile. The performance difference was negligible (on a PC at least) and in some places it required me to copy values into local variables so I could pass them (which then negated any performance benefit).

> We can first test your proposal to use [MethodImpl(MethodImplOptions.AggressiveInlining)] on .Set() methods.

I've messed with this a little, but I haven't had much luck with it giving me a noticeable performance increase yet. I'll try spend some more time on it later though if I can.

> That's impresive. I think we should apply this right after we inline DrawInternal() into DrawInto(), it's a clear win in performance.

Cool, sounds good.

nkast:
> (but it currently is creating a new Vector4 anyway).

In DrawInternal() ? We can remove that on inlined methods.

> I've messed with this a little, but I haven't had much luck with it giving me a noticeable performance increase yet. I'll try spend some more time on it later though if I can.

I tested  `[MethodImpl(MethodImplOptions.AggressiveInlining)]` with `.Set()` on top of https://github.com/MonoGame/MonoGame/compare/develop...nkast:tncSpriteBatchDraw and got a 12.9% improvement! If it transfers well to mono, Native.Net, etc we might not need `ref` after all. We can also use it to refactor common code like `SortKey`. And also prefix with it small math methods in Vector2/3/4, Matrix, etc.

I will leave it to you if you want to post a PR.

nkast:
@tomspilman , @KonajuGames Any thoughts on this PR? 
We could use the optimizations around _texelSize before moving forward with #5347

MichaelDePiazzi:
> and got a 12.9% improvement!

That is excellent! I hadn't tried using it with `Set`, so that's great to hear that it can make a worthwhile difference!

> I will leave it to you if you want to post a PR.

Honestly, I won't have much more time to spend on this. So if you have a chance to work on it, then please feel free to submit a PR. Otherwise I'll get to it when I can.

nkast:
> Honestly, I won't have much more time to spend on this. So if you have a chance to work on it, then please feel free to submit a PR. Otherwise I'll get to it when I can.

Cool, No problem.

nkast:
Drop the commit about Texture2D.TexelSize as it is now addressed by #5453.
I still need to rebase this PR and see which changes are still relevant after #5401.

PR #5273, opened 2016-11-02
Use OpenGL flag when creating an SDL Window

james0x0A:
Allows a graphics device to be created before the main window is created by enabling a OpenGL context to be bound to the Window.   This is the primary cause of failed OpenGL unit tests under Windows.

### DIFF
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -105,10 +105,14 @@ namespace Microsoft.Xna.Framework
                 _winy = display.Y + display.Height / 2;
             }
 
+            var initflags =
+                Sdl.Window.State.OpenGL |
+                Sdl.Window.State.Hidden;
+
             // We need a dummy handle for GraphicDevice until our window gets created
             _handle = Sdl.Window.Create("", _winx, _winy,
                 GraphicsDeviceManager.DefaultBackBufferWidth, GraphicsDeviceManager.DefaultBackBufferHeight,
-                Sdl.Window.State.Hidden);
+                initflags);
         }
 
         internal void CreateWindow()


tomspilman:
@james0x0A - Not sure why the mac build failed exactly... I kicked it back off to try again.

MonoGame.Framework/SDL/SDLGameWindow.cs
```diff
@@ -105,10 +105,14 @@ public SdlGameWindow(Game game)
                 _winy = display.Y + display.Height / 2;
             }
 
+            var initflags =
+                Sdl.Window.State.OpenGL |
+                Sdl.Window.State.Hidden;
+
```
tomspilman: @dellis1972 @cra0zy - This look ok to you guys?

mrhelmut:
Caution here, it only concerns the dummy window handle created to init the ```GraphicDevice```, but it is destroyed right after and the proper window where all the rendering happen does have the GL flag. Problem is that it may affect the GL context lifecycle outside of the test framework by creating 2 GL contexts and may end up with unpredictable errors. I'm not quite sure about what this flag really does, but it needs serious testing.

cra0zy:
The PR I was gonna make after https://github.com/MonoGame/MonoGame.Dependencies/pull/92 gets merged will fix this in a proper manner. Window creation was delayed so that we could set resizable param, but SDL 2.0.5 has the dynamic resizable function so that hack is no longer needed.

james0x0A:
Seems a reasonable precaution.  I  will research into it further.

james0x0A:
Or go with @cra0zy. That would be an overall better solution.

cra0zy:
The PR is for 2.0.5 deps has been merged and new PR for updating the main repo is ready.

james0x0A:
That sounds great. closing this

PR #5275, opened 2016-11-02
Set the entry assembly when under test

### LINKED ISSUE #5274: Assembly.GetEntryAssembly() may be null when called during unit tests, causing OpenGL unit tests to fail.
Assembly.GetEntryAssembly() is called to retrieve the game executable and set the icon for the SDL Window.   

setting the icon can either be skipped on null,  or an entry assembly can be set manually while creating the Game instance as described at http://stackoverflow.com/a/21888521

james0x0A:
Fixes #5274  

Sourced from http://stackoverflow.com/a/21888521


james0x0A:
this is unreliable

PR #5281, opened 2016-11-06
UWP c# project update

DDReaper:
Updated UWP NuSpec as this didn't deploy SharpDX when updating samples
Updated Target Framework to Anniversary edition and latest .NETCore NuGet target
Added Xbox code to disable cursor

### DIFF
--- a/NuGetPackages/MonoGame.Framework.WindowsUniversal.nuspec
+++ b/NuGetPackages/MonoGame.Framework.WindowsUniversal.nuspec
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
     <metadata>
         <id>MonoGame.Framework.WindowsUniversal</id>
         <version>0.0.0.0</version>
@@ -16,9 +16,6 @@
         <releaseNotes></releaseNotes>
         <copyright>Copyright  2016</copyright>
         <language>en-US</language>
-        <references>
-            <reference file="MonoGame.Framework.dll" />
-        </references>
 	</metadata>
     <files>
         <file src="WindowsUniversal\AnyCPU\Release\Themes\generic.xbf" target="lib\netcore\MonoGame.Framework\Themes\generic.xbf" />
--- a/ProjectTemplates/VisualStudio2015/WindowsUniversal/App.xaml.cs
+++ b/ProjectTemplates/VisualStudio2015/WindowsUniversal/App.xaml.cs
@@ -35,11 +35,16 @@ namespace $safeprojectname$
     {
         this.InitializeComponent();
         this.Suspending += OnSuspending;
-        //If Deploying to Xbox, target the project at the "Windows 10 Anniversary Edition (14393) and uncomment the below code
-        //if (IsXbox())
-        //{
-        //    Application.Current.RequiresPointerMode = ApplicationRequiresPointerMode.WhenRequested;
-        //}
+
+        //API check to ensure the "RequiresPointerMode" property exists, ensuring project is running on build 14393 or later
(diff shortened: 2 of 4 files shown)


tomspilman:
@BrianPeek - These changes make sense to you for UWP/XB1?

DDReaper:
Tested and working fine on Desktop, Mobile and Xbox

BrianPeek:
I have 2 questions/comments:

1) Can you cache the IsXbox check?  I realize you only use it once, but maybe not best to compare strings if it's something that a dev would use in Update/Draw?  Or am I overthinking that?  :)

2) The RequiresPointerMode properly is new to 14393, so the app template using 10586 as a minimum isn't a good idea.  My suggestion would be to change min back to 10240 and use the ApiInformation stuffs to make sure the API exists before calling it.  Or is there some other reason it's bumped to 10586?

tomspilman:
Thanks for the feedback @BrianPeek !

DDReaper:
Updated implementation following comments: … 
1) Can you cache the IsXbox check?
Intensive check is already cached in the "deviceFamily", the "IsXbox" check is just a boolean string comparator, no need to cache it.
By using device family, you can check what the platform is being executed.

2) My suggestion would be to change min back to 10240
Rolled back platform update and commented out the "RequiresPointerMode" check.  
This API was only added in 14393, so an APIInformation call wouldn't help as that is used for differences between extensions, not new versions.  Code won't compile in previous versions

BrianPeek:
IsPropertyPresent should work for #2 above...

https://msdn.microsoft.com/en-us/library/windows/apps/windows.foundation.metadata.apiinformation.ispropertypresent.aspx

DDReaper:
That is true, it will work.  However what it won't do is stop the compiler seeing the RequiresPointerMode properties when the targetframework is lower than 14393.
So the code won't even compile. (doesn't work like a #if, unless I'm missing something) Feel free to try.
If you have an alternative, please share and I'll add it in.

As far as I'm aware, the APIInformation tests are more for cross-platform support, checking if you are running on a platform where that function will do something. Not to stop compile errors from earlier API versions.

BrianPeek:
Ah, I get what you're after.

So, in my head, the developer is (well, should be) using the latest tooling whether they're Xbox/14393/whatever or not.  If they are, and the project is setup with "Target version (TargetPlatformVersion in csproj)" of 14393 and "Min version (TargetPlatformMinVersion)" of 10240, the code compiles and IsPropertyPresent works as it should, and we cover all versions of Win10.

Maybe this is more of a question for Tom, but my opinion and basis for this reasoning is that the onus is on the developer to be running the latest dev SDK (14393 at present) when building, no matter what min Win10 version they're targeting...

DDReaper:
Yup, pretty close. Except as this api is only available from 14393, so that would also need to be the min version. If you ran it on 10240, I believe it would cause an exception.

My view us that 14393 should be the new min version as most people I'm aware of have either chosen to update or been magically updated over night to at least 14393.

14383 is the min version for Xbox anyway

DDReaper:
If we do agree to use the latest as the min (my preference) then I'll need to update the pr again.
It's still only the default (what you see when you do a new project) and users can choose to override that

BrianPeek:
It definitely won't cause an exception...

```cs
if(ApiInformation.IsPropertyPresent("Windows.UI.Xaml.Application", "RequiresPointerMode"))
    Application.Current.RequiresPointerMode = ApplicationRequiresPointerMode.Auto;
```

That code will compile using latest tools but still run on Win10 10240 and above on any device.  The "if" statement will fail on <14393, which won't run the next line on any downlevel version.

I don't agree with bumping the min version...while Win10 desktop has been updated in many cases, there are still phones stuck on older versions due to carrier policy.  Maybe not 10240, but why drop a segment of the audience that hasn't/won't/can't upgrade when there's no reason to?  I'd leave min at 10240, target at 14393 and let the dev narrow the scope/audience if they need to...that's my opinion.  :)

DDReaper:
Ok, I'll make that update to the pr shortly

DDReaper:
Nudge @dellis1972 @tomspilman @KonajuGames

PR #5284, opened 2016-11-07
Fixes Assembly.GetEntryAssembly() can be null when called from nunit

james0x0A:
#5274

### DIFF
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -117,25 +117,27 @@ namespace Microsoft.Xna.Framework
             Sdl.SetHint("SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS", "0");
             Sdl.SetHint("SDL_JOYSTICK_ALLOW_BACKGROUND_EVENTS", "1");
 
-            using (
-                var stream =
-                    Assembly.GetEntryAssembly().GetManifestResourceStream(Assembly.GetEntryAssembly().EntryPoint.DeclaringType.Namespace + ".Icon.bmp") ??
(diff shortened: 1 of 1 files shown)


tomspilman:
This seems like a reasonable way to prevent this crash in special cases like the unit tests.

(had to use the `?w=0` whitespace trick to see the diff clearly)

@KonajuGames @dellis1972 ?

tomspilman:
@Jjagg @dellis1972 @cra0zy - Thoughts?

Jjagg:
This is fine. A comment that explains why this is necessary would be nice though.

cra0zy:
This is not fine, add a null check instead of a try catch.

tomspilman:
Yeah I agree with @cra0zy ...  if a `null` check is all we need here then replace the `try` with `if (Assembly.GetEntryAssembly() != null)`.

KonajuGames:
The `catch { }` may hide other legitimate errors.

cra0zy:
Side note, if you upgrade the windows build bot it will be easier to add null checks, ie:
```c#
Assembly.GetEntryAssembly()?.GetManifestResourceStream(Assembly.GetEntryAssembly().EntryPoint.DeclaringType.Namespace + ".Icon.bmp") ??
Assembly.GetEntryAssembly()?.GetManifestResourceStream("Icon.bmp") ??
Assembly.GetExecutingAssembly()?.GetManifestResourceStream("MonoGame.bmp"))
```

C# 6 has nice `var?.method()` null check :)

tomspilman:
>C# 6 has nice var?.method() null check :)

I am aware of that... we're not going to do that yet.

tomspilman:
This was fixed by https://github.com/MonoGame/MonoGame/pull/5764.  Thanks for alerting us to the issue @james0x0A !

Issue #3977, opened 2015-07-02
sampler TextureSamplers[1]


DrSanjeep:
//FX

texture Texture0;

sampler TextureSamplers[1] = 
{
  sampler_state
  {
    Texture = (Texture0);
    
    MinFilter = Linear;
    MagFilter = Linear;
    MipFilter = Linear;
    
    AddressU = Wrap;
    AddressV = Wrap;
  }
}

//CS
effect.Parameters["Texture0"].SetValue(textures[0]);

======================
Can't get this old XNA code to work. 
It makes an effect parameter "[Object Texture2D] TextureSamplers : (null)"
called TextureSamplers that I can't SetValue at runtime 
can't pass array of textures, no SetValue overload for array of textures
XNA used to let me set the Texture0 parameter to a Texture2D
Thanks

DrSanjeep:
issue only present in shader 4_0. cranking up to shader model 5_0 is the workaround but I have to 
    effect.Parameters["TextureSamplers[0]"].SetValue(textures[0]);
    effect.Parameters["TextureSamplers[1]"].SetValue(textures[1]);
whereas in the good old days it was
    effect.Parameters["Texture0"].SetValue(textures[0]);
    effect.Parameters["Texture1"].SetValue(textures[1]);

Issue #3592, opened 2015-03-04
WP8.1 crash/freeze when IsfixedTimestep=true


KakCAT:
Hi!

WP8.1 crashes in the Tick function of Game.cs, exactly in the line 
```
Task.Delay(sleepTime).Wait();
```

more info here : http://community.monogame.net/t/wp8-1-game-freezes-sporadically/1865/13

stack trace + exception info here : http://s28.postimg.org/ra702djzh/crash.png

Surrounding the Wait call with try/catch fixes the problem but I don't think it's the right solution

Thanks!

nkast:
You can just ignore those types of exceptions. I am sure if you remove ArgumentException from the list of unhandled/thrown in the Debug-> Exceptions your program will run without problem.

KakCAT:
Hi nkast,

you're right, sorry for the noise. I was confused by the fact that initially the crash was produced when testing the touchscreen bug. Closing.

Issue #3471, opened 2015-01-28
Consider exposing low-level Shader / ConstantBuffer API


tgjones:
I see from #761 and [this comment](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/Shader/ConstantBuffer.cs#L79) that there is a long-term plan to expose a Shader API to users. I wanted to add my +1 to this idea, and get a discussion going about whether - and if so, how best - to do so.

(As background: up to v3.x, XNA did have APIs for working directly with shaders, but these APIs were removed in v4.0. You could create [VertexShader](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.vertexshader%28v=xnagamestudio.31%29.aspx) and [PixelShader](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.pixelshader%28v=xnagamestudio.31%29.aspx) objects, and there were corresponding sets of `Get*Shader*Constant` and `Set*ShaderConstant` methods on [GraphicsDevice](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.graphicsdevice_members%28v=xnagamestudio.31%29.aspx).)

The existing - `internal` - `Shader` and `ConstantBuffer` classes already look well-factored to me, and `GraphicsDevice` already has - `internal` - properties for `VertexShader` and `PixelShader`, and a `SetConstantBuffer` method.

One question I want to ask, and my apologies if this has been discussed already: are there any plans in the medium/long-term to move to a Direct3D 11-style (or more specifically, Direct3D 11 as interpreted by SharpDX) API, where there are separate classes for each pipeline stage? I can see arguments both ways, and it certainly doesn't have to be decided as part of this issue, but I wanted to raise it as it might affect the public API for shaders and constant buffers.

I don't know how controversial the idea of exposing a low-level `Shader` / `ConstantBuffer` API is - there's certainly potential for confusing new users, if you need to choose between shaders and effects.  Any thoughts on this?

Are there are any platforms currently supported by MonoGame that don't support low-level shaders, but do support effects? (I assume not, but it's worth asking.)

[...]

PR #5288, opened 2016-11-09
Don't lose textures when applying effect

hifi:
This is not necessary the correct behavior and I'm unable to test it on anything else than UltimaXNA but this XNA game expects textures are preserved when applying an effect multiple times and between frames. On MonoGame I had to reset them per effect apply.

As an additional note, the game in question has `RenderTargetUsage.PreserveContents` set for the graphics device.

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/EffectPass.cs
+++ b/MonoGame.Framework/Graphics/Effect/EffectPass.cs
@@ -122,7 +122,8 @@ namespace Microsoft.Xna.Framework.Graphics
                 var param = _effect.Parameters[sampler.parameter];
                 var texture = param.Data as Texture;
 
-                textures[sampler.textureSlot] = texture;
+                if (texture != null)
+                    textures[sampler.textureSlot] = texture;
 
                 // If there is a sampler state set it.
                 if (sampler.state != null)


nkast:
Test `EffectPassShouldSetTextureEvenIfNull` failed.
Some effects like BasicEffect apply their textures when you call .Apply().
You have to either reapply your textures after .Apply(), override the BasicEffect.OnApply(), or create a new Effect that doesn't change the textures.

hifi:
Is this an intended difference from XNA or does XNA has undefined behavior?

tomspilman:
@nkast - I think there might be more complex behavior in XNA that we're not correctly emulating in MonoGame.

@hifi - Can you provide us some example code of a case where XNA works and MonoGame does not?

hifi:
@tomspilman I will need some time to set up a XNA development environment and a small test case but I will try. I only have UltimaXNA as a reference of a game that does work on XNA and when something doesn't work on MonoGame I expect it would on XNA.

tomspilman:
>  I will need some time to set up a XNA development environment and a small test case but I will try. 

If you can just cut and paste some code that you know doesn't work someone else might be able to jump in and test it in XNA and MonoGame and examine the results.

tomspilman:
@hifi - So I tried a bunch of stuff tonight and haven't been able to reproduce this problem behavior.  From what I can tell XNA and MonoGame are working 100% the same here.

The most important thing we could get would be a sample that shows the bug in action.  Without that we will never figure this out.

hifi:
As I'm not an expert on this issue I can't really produce a small test case easily. I can only test the complete game with proposed solutions. I believe @Jjagg is on the right track with #5392.

Jjagg:
I added a test in that PR. We know exactly what the issue is, but with the D3D11 Compiler it seems we can't get the required reflection data to replicate XNA behavior... So I think just doing the null check like your fix here might be the best thing to do. Unless there's actually some value to setting a texture to null as is allowed now, but I don't think so. It's not like you can do anything with a null texture, so from a practical standpoint it won't break compatibility for xna games (even though it breaks with xna behavior a bit).

tomspilman:
> Unless there's actually some value to setting a texture to null as is allowed now

I see only two potential uses:

 - Clear the previously set texture to `null`.
 - You intended to use `null` to render black.

Neither case is super useful and I've not seen any games so far that depend on that behavior.

I think we could live with this difference without seeing any bad side effects.

@KonajuGames ?

Jjagg:
If we can expose the textureslot that a texture parameter is bound to somehow we can allow people to bypass this by doing ` GraphicsDevice.Texture[textureSlot] = null`

Jjagg:
> as a part of opening up the Shader API.

I'm talking about #3471

Also, looking for that issue I came across #3977 which is related to this problem. That is one thing we *can* fix fairly easy :) Once we reach a consensus here I'll get the fix discussed here and a fix for #3977 implemented in #5392.

KonajuGames:
> I think we could live with this difference without seeing any bad side
effects.

To be honest, I hadn't been following this thread much, so I've just got a
summary of it.

It would appear to be a difference, but trying to fix it properly is not
really possible right now. I would agree that we can live with this
difference.

tomspilman:
@Jjagg - So I guess we can go with this solution... although we need a PR that also fixes the unit test with a comment to the difference in behavior.

Jjagg:
Sounds good. I'll get this all worked out in #3592 when I have the time. I'll also think about adding some information regarding this to the docs here: http://www.monogame.net/documentation/?page=Custom_Effects

tomspilman:
Thanks @Jjagg !

Jjagg:
I was writing up something for in the docs and I came up with another solution that I much prefer over the solution we discussed here. 
I wrote it all down in a gist since it's a bit lenghty. It includes a whole part that would go in the docs to explain things a bit to users. If you guys think it's too verbose we can cut it down a bit, since there is a bit of technical information in there (which I personally would like to read in the docs btw) and we could alternatively just say "write your shaders in one of the following three ways, but NOT in this one" and link to the discussion here for interested users.

Here's the gist: https://gist.github.com/Jjagg/096c5218d21c5e2944a88dbc1b6947b3

____________________________

The case that won't work is the following:
```
Texture2D MyTexture : register(t0);
sampler MySampler : register(s0);
```
This will not create an EffectParameter named MyTexture. Note that currently it won't do that either, it will generate an EffectParameter named MySampler which is not very intuitive. XNA gets this case right, but we can just say we don't support it (because we can't). The solution here is to set the texture in the sampler_state i.e.
```
Texture2D MyTexture : register(t0);
sampler MySampler : register(s0) = sampler_state
{
    Texture = (MyTexture);
}
```
This will make the shader behave the same as in XNA. The effect will have an Effectparameter named MyTexture.

Little side note: I don't know if OpenGL will work correctly in all cases when people use the new sampler syntax and sample from different samplers (because of the complexity involved) but I know that we can make it work. MojoShader creates a seperate sampler for each sampler-texture combination in the HLSL. That means when one texture is updated the backend sometimes has to update multiple textures and the same goes for sampler states. 

cc @tomspilman 
Sorry for the long read, but IMO it's worth it :p

tomspilman:
@Jjagg - Need more time to read this thru and consider our options.

IMO I still feel the simple "don't set null textures in an Effect" solves all the issues the best and allows for older shaders to just work.  This is better than forcing users to fix shaders that functioned fine in XNA.  Especially since some devs are trying to keep both an XNA version of their game and a MG version at the same time.

However I think us giving the users advice on how best to setup textures/samplers isn't a bad idea.  We should certainly give them these recommendations in the docs.

Jjagg:
Take your time. I'm busy with exams, so I shouldn't be working on MG as much as I am right now anyway :p

> This is better than forcing users to fix shaders that functioned fine in XNA. 

They won't have to. Or at least not in cases were they don't have to now. The solution I propose in no case does worse than the current implementation in terms of XNA compatibility and does better in one.

## Summary of behavior
|                      |                 | Old Syntax | New Syntax | Sampler Only | Matching Slots |
|-------------- |----------------:|:---------------:|:--------------:|:----------------:|:------------------:|
| **XNA**        | EffectParameter | Texture         | Texture        | None             | Texture            |
| **Current**   | EffectParameter | Texture         | Texture        | Sampler          | Sampler            |
|                      |  XNA Compatible | Yes             | Yes         | ***No***           | No                 |
| **Proposed** | EffectParameter | Texture         | Texture        | None             | None               |
|                       |  XNA Compatible | Yes             | Yes        | ***Yes***          | No                 |

The issue here is occuring because of the incompatibility in the 'Sampler Only' column.

zigzag312:
+1. This PR fixed some issues for me when migrating from XNA

Jjagg:
I just got bit by this again. Gonna be implementing the fix I mentioned in my previous comment.

tomspilman:
@Jjagg @cra0zy Either of you have time to refresh this and finally fix it?

This is the fix i've been for quite a while now...

https://github.com/SickheadGames/MonoGame/commit/78d8e2743d67c03a0b0d4df5c36d3305564bd2ad

Jjagg:
@tomspilman Looking at this again I think maybe we overcomplicated the issue. IMO the `null` check as performed in this PR is a good fix. 

The issue currently is that applying an effect with a `null` texture unbinds the texture of the corresponding slot while it doesn't in XNA. In XNA however it does return `null` when querying the texture for that slot afterwards. So there's an odd distinction between what's actually bound and what's in TextureCollection. My guess is XNA immediately binds textures only if they're not `null`, but always sets the reference in the `TextureCollection` even if the texture is `null`. There's really no point for us to keep this behavior. Let's remove the test that broke and implement the fix from this PR.

tomspilman:
>  IMO the null check as performed in this PR is a good fix.

I think it is an acceptable fix for now.

I still feel we need to fix it correctly to work like XNA does, but that is a lower priority.

> Let's remove the test that broke and implement the fix from this PR.

Let's only disable the test and leave a comment back to this PR.

nkast:
EffectPass.Apply() updates GraphicsDevice.Textures from any samplers in the shader.
Later during a draw the GraphicsDevice.Textures are copied back to shaderStage.

The first step is the source of all the issues I think, GraphicsDevice.Textures should be seen as the source of textures for effects and not by coupled so close to the Graphics context. Perhaps what we need is a shadow copy of GraphicsDevice.Textures , or better yet, to set shaderStage resources directly from  GraphicsDevice.Textures  and/or Samplers.

nkast:
Here are two key points in the code that we need to merge somehow.
We have to set the texture from the sampler with the same textureSlot, if not found use the texture from  GraphicsDevice.Textures. 
Maybe we need a texture array in the PassEffect which we clean and update during   PassEffect.Apply(), and then fill in the null values from   GraphicsDevice.Textures during a Draw()/ApplyState().


https://github.com/MonoGame/MonoGame/blob/6f34eb393aa0ac005888d74c5c4c6ab5615fdc8c/MonoGame.Framework/Graphics/Effect/EffectPass.cs#L120-L130

https://github.com/MonoGame/MonoGame/blob/6f34eb393aa0ac005888d74c5c4c6ab5615fdc8c/MonoGame.Framework/Graphics/TextureCollection.DirectX.cs#L72-L93

nkast:
I see now that XNA will overwrite the GraphicsDevice.Textures from the shader if the shader defines a Texture slot. I thing we should add this case too in #5392.
Additional, MG will set a texture to null in case you attach the same texture as RenderTarget. XNA will not do that.

I don't like the idea to replace the current behavior. It can cause unexpected behavior during developing. I would expect a shader to draw black when I set one of the textures to null.

SimonDarksideJ:
Any feedback @hifi hifi from @nkast comments?

hifi:
It's been over 6 years and I haven't used MonoGame outside debugging UltimaXNA in 2016 so I can't really be of much help anymore.

If it's easier for everyone this PR could just be closed and someone can create a new one if this issue is still relevant.

SimonDarksideJ:
Thanks for the contribution @hifi

PR #5289, opened 2016-11-09
[Pipeline Tool] PathDialog Improvements

cra0zy:
Stuff done:
 - browse textbox and button now get centered vertically
 - adds a new label that states "Special constants:" above the Platform/Configuration/Config/Profile buttons
   - do tell me if you want it to say something else, I also thought about "Special symbols:" and "Special defines:"

### DIFF
--- a/Tools/Pipeline/Dialogs/PathDialog.eto.cs
+++ b/Tools/Pipeline/Dialogs/PathDialog.eto.cs
@@ -11,7 +11,7 @@ namespace MonoGame.Tools.Pipeline
     {
         DynamicLayout layout1;
         StackLayout stack1, stack2;
-        Label label1;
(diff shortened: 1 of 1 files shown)


tomspilman:
@cra0zy - VS uses the term "Macros" for these things:

image

cra0zy:
Ha, didn't realize that VS had a dialog like that... label has been changed to say "Macros:".

tomspilman:
> Ha, didn't realize that VS had a dialog like that

You get it in C++ projects which is possibly why you didn't run into it before.

cra0zy:
> You get it in C++ projects which is possibly why you didn't run into it before.

Yup, I've actually never used VS for C/C++ projects..., in my defense I am on Linux :)

PR #5290, opened 2016-11-09
[Pipeline Tool] Drag and Drop from file manager

### LINKED ISSUE #5256: Drag and drop files in the Content Pipeline tool on windows
I just downloaded the latest development build from http://www.monogame.net/downloads/ to check out how the pipeline is looking now. As I understand it, it now uses WPF instead of whatever was used previously. However, drag and drop support for adding files to a content project does not work anymore, which it used to in Monogame 3.5.

cra0zy:
Fixes #5256

### DIFF
--- a/Tools/Pipeline/MainWindow.cs
+++ b/Tools/Pipeline/MainWindow.cs
@@ -62,13 +62,11 @@ namespace MonoGame.Tools.Pipeline
         protected override void OnClosing(System.ComponentModel.CancelEventArgs e)
         {
             e.Cancel = !PipelineController.Instance.Exit();
-            base.OnClosing(e);
-        }
 
-        public override void Close()
-        {
-            Application.Instance.Quit();
-            base.Close();
+            if (!e.Cancel)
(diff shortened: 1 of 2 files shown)


Tools/Pipeline/MainWindow.cs
```diff
@@ -62,13 +62,11 @@ public MainWindow()
...
             e.Cancel = !PipelineController.Instance.Exit();
-            base.OnClosing(e);
-        }
 
-        public override void Close()
-        {
-            Application.Instance.Quit();
-            base.Close();
+            if (!e.Cancel)
+                Xwt.Application.Exit();
```
tomspilman: This change looks separate from the drag and drop... what is it fixing?

cra0zy: Xwt starts a Task with DragAndDrop operation so I needed to Exit the app from it, otherwise it wouldn't exit.

tomspilman:
Thanks... merging!

PR #5299, opened 2016-11-15
Custom texture compression for SpriteFonts

### LINKED ISSUE #2853: Monochrome Optimized DXT3 Compression For SpriteFont
The XNA content pipeline produces DXT3 compressed textures for SpriteFont.  Comparing the results from XNA to the Nvidia Texture Compressor we can see significant artifacts using normal DXT3 compression.

In fact I've seen papers talking about monochrome optimizations for DXT.... it could be possible XNA is doing this under the hood.

We need to develop our own monochrome optimized DXT3 compression specifically for SpriteFont.

KonajuGames:
- Implemented a custom DXT3 compressor for font texture pages to eliminate block compression artifacts.
- iOS uses 16-bit colour textures for compressed font texture pages.
- Android already used 16-bit colour textures for compressed font texture pages as a side effect of ETC1 support.
- Added PremultiplyAlpha property to FontDescriptionProcessor for consistency (FontTextureProcessor and TextureProcessor already have the same property).

customdxt3font

Fixes #2853

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/GraphicsUtil.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/GraphicsUtil.cs
@@ -123,10 +123,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
             return result;
         }
 
-        public static void CompressPvrtc(TextureContent content, bool generateMipMaps, bool sharpAlpha)
+        public static void CompressPvrtc(TextureContent content, bool generateMipMaps, bool isSpriteFont)
         {
             // If sharp alpha is required (for a font texture page), use 16-bit color instead of PVR
-            if (sharpAlpha)
+            if (isSpriteFont)
             {
                 CompressColor16Bit(content, generateMipMaps);
                 return;
@@ -152,7 +152,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
                 Compress(typeof(PvrtcRgba4BitmapContent), content, generateMipMaps);
         }
 
-        public static void CompressDxt(GraphicsProfile profile, TextureContent content, bool generateMipMaps, bool sharpAlpha)
+        public static void CompressDxt(GraphicsProfile profile, TextureContent content, bool generateMipMaps, bool isSpriteFont)
         {
             var face = content.Faces[0][0];
 
@@ -167,7 +167,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 
             if (alphaRange == AlphaRange.Opaque)
                 Compress(typeof(Dxt1BitmapContent), content, generateMipMaps);
-            else if (sharpAlpha)
+            else if (isSpriteFont)
                 CompressFontDXT3(content, generateMipMaps);
             else if (alphaRange == AlphaRange.Cutout)
                 Compress(typeof(Dxt3BitmapContent), content, generateMipMaps);
@@ -275,7 +275,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
         }
 
         // Compress the greyscale font texture page using a specially-formulated DXT3 mode
-        static unsafe void CompressFontDXT3(TextureContent content, bool generateMipmaps)
+        static public unsafe void CompressFontDXT3(TextureContent content, bool generateMipmaps)
         {
             if (content.Faces.Count > 1)
                 throw new PipelineException("Font textures should only have one face");
--- a/MonoGame.Framework.Content.Pipeline/Graphics/GraphicsUtil.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/GraphicsUtil.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.IO;
 using Microsoft.Xna.Framework.Graphics;
 using Microsoft.Xna.Framework.Graphics.PackedVector;
 using FreeImageAPI;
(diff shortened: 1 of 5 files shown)


MonoGame.Framework.Content.Pipeline/Graphics/GraphicsUtil.cs
```diff
@@ -263,5 +273,241 @@ static void Compress(Type targetType, TextureContent content, bool generateMipMa
                 content.ConvertBitmapType(targetType);
             }
         }
+
+        // Compress the greyscale font texture page using a specially-formulated DXT3 mode
+        static unsafe void CompressFontDXT3(TextureContent content, bool generateMipmaps)
```
tomspilman: We should make this public.  We will need to call it from the `TextureProfile`s for specific platforms like XB1, Vita, WiiU, etc.

KonajuGames: Will do.

MonoGame.Framework.Content.Pipeline/Graphics/GraphicsUtil.cs
```diff
@@ -159,7 +167,9 @@ public static void CompressDxt(GraphicsProfile profile, TextureContent content,
 
             if (alphaRange == AlphaRange.Opaque)
                 Compress(typeof(Dxt1BitmapContent), content, generateMipMaps);
-            else if (alphaRange == AlphaRange.Cutout || sharpAlpha)
+            else if (sharpAlpha)
+                CompressFontDXT3(content, generateMipMaps);
```
tomspilman: I think `sharpAlpha` is a bad way to trigger this special DXT3 compressor.  Technically it has nothing to do with the sharpness of alpha.  It is simply an optimization for perfectly monochrome (I guess actually white) textures that only vary in the alpha channel.

So I see three different approaches here:

 1. We add another `bool` to the `TextureProfile.PlatformCompressTexture`.
 2. We add a new `TextureProfile.PlatformCompressSpriteFontTexture()` and directly use it from `SpriteFont`.
 3. We internally can scan the texture data and automatically detect that this special DXT3 format is best.

I'm leaning towards 2.

But if we could do 3 right... it *might* be more useful...  or maybe it causes trouble when you don't expect it.

KonajuGames: `sharpAlpha` was added for font texture compression, and is only used by font processing, so that is the trigger for this special behaviour.  Perhaps renaming the parameter would be better.

tomspilman: So maybe something very specific like `isSpriteFont` is better?  Or maybe `isMonochrome` or `isGreyscale`?

tomspilman: I guess isSpriteFont is best as 16bit color isn't necessarily a good thing for just any greyscale texture.

KonajuGames: I first though I'd use 'isSpriteFont' because this compression only gives four levels of grey, including black and white, so it's not really useful for other greyscale textures unless they had very specific needs, i.e. a logo or similar.

tomspilman:
Was thinking it would be good to see a comparison of the DXT3 compressed sprite font and the 16bit color sprite font.

If DXT3 is 4x compression and 16bit is 4x compression... is one better than the other?  Will people porting games to different platforms run into unexpected artifacts in the font texture quality?

MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
```diff
@@ -21,12 +21,16 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
     [ContentProcessor(DisplayName = "Sprite Font Description - MonoGame")]
     public class FontDescriptionProcessor : ContentProcessor<FontDescription, SpriteFontContent>
     {
+        [DefaultValue(true)]
+        public virtual bool PremultiplyAlpha { get; set; }
```
KonajuGames: I checked both.  It works fine.

KonajuGames:
16-bit is 2:1 compression compared to DXT3's 4:1.

tomspilman:
Duh.... yeah... had mipmaps on the brain. :)

KonajuGames:
As for quality, the 16-bit texture has 16 levels of grey where the DXT3 texture has four levels of grey.  For fonts, the alpha is used for anti-aliasing only and I think you would have to be looking very closely to see any discernable difference.

MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
KonajuGames: The non-premultiplied case is why `CompressFontDXT3Block` reads and processes alpha and RGB separately.  Originally it used the alpha value for the RGB as well, but that is only valid if the texture is premultiplied.  For non-premultiplied, the RGB channels are all white and the glyph is defined by the alpha channel.

MichaelDePiazzi:
Just gave this a try and fonts look much sharper now! Nice one! 👍

tomspilman:
Ok... merging this now!

SeriousMartin:
What about colorized fonts? 
The new compressor doesn't seem to allow this and deactivating DXT compression results in more memory consumed.
I don't see why spritefonts are greyscale per defintion.
Maybe adding an "IsGreyscale" property to the FontTextureProcessor would be a solution.
What do you think?

Issue #1941, opened 2013-08-15
Expand GraphicsProfile Beyond Reach/HiDef


tomspilman:
We need to decide how to expand the existing `GraphicsProfile` system beyond the old Reach/HiDef options from XNA 4.

We have a few options:
- Go back to cap bits.
- Expand `GraphicsProfile` with more profiles.
- Add a new `FeatureLevel` enum similar to DirectX feature levels.

Cap bits enable "full flexibility" but it puts the burden all on the end user to figure out what combinations of bits are common.  I don't think anyone wants to go to cap bits... even if they say they do.  This is what sucked about DX9 and what IMO still sucks about OpenGL.

I could see expanding the existing `GraphicsProfile` to support more profiles:

```
public enum GraphicsProfile
{
    XboxOne,
    PlayStation4,                    
    HiDef,
    MidDef,
    LowDef,
    Reach,
    Limited
}
```

Some special platform specific profiles could be nice, but maybe sticking to more generic profiles is better?

The final option here is just mimicking the feature level concept from DirectX.  This is sort of what `GraphicsProfile` does albeit with only two levels.  Having it as a new separate enum just gives us some flexibility to break with the old system.

With either feature levels or profiles we will still need to look at all the different platforms MonoGame targets and all the hardware variations and make some calls as to how to decide how to categorize any particular hardware device.  We then need to clearly document these rules for developers.

tomspilman:
FYI.  DX12 adds a few more feature levels:

http://postimg.org/image/fd4kuk85v/

I keep thinking we need to adopt DirectX feature levels with maybe some better names that DX version numbers.  The defined DX feature levels are:
```
  D3D_FEATURE_LEVEL_9_1
  D3D_FEATURE_LEVEL_9_2
  D3D_FEATURE_LEVEL_9_3
  D3D_FEATURE_LEVEL_10_0
  D3D_FEATURE_LEVEL_10_1
  D3D_FEATURE_LEVEL_11_0
  D3D_FEATURE_LEVEL_11_1
  D3D_FEATURE_LEVEL_12_0
  D3D_FEATURE_LEVEL_12_1
```
Since DX runs on all the same hardware OpenGL does, DX feature levels can apply to OpenGL too.  So hardware that is 10.1 feature level will have all the OGL extensions to support those features.

Still a two tiered system where we mix caps and feature levels could work.

[...]

PR #5304, opened 2016-11-18
Use up to 8 render targets.

nkast:
Feature level 10.0 can have up to 8 render targets.
https://msdn.microsoft.com/en-us/library/windows/desktop/ff476876(v=vs.85).aspx

fix http://community.monogame.net/t/error-when-setrendertargets-more-than-4-rtbinding/8408

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -54,7 +54,7 @@ namespace Microsoft.Xna.Framework.Graphics
         private IndexBuffer _indexBuffer;
         private bool _indexBufferDirty;
 
-        private readonly RenderTargetBinding[] _currentRenderTargetBindings = new RenderTargetBinding[4];
+        private readonly RenderTargetBinding[] _currentRenderTargetBindings = new RenderTargetBinding[8];
         private int _currentRenderTargetCount;
         private readonly RenderTargetBinding[] _tempRenderTargetBinding = new RenderTargetBinding[1];
 


KonajuGames:
We should gracefully handle the cases where eight render targets are not supported.

tomspilman:
> We should gracefully handle the cases where eight render targets are not supported.

What happens right now if this code change is run on a system without 8 RT support?

KonajuGames:
It appears to submit it to the graphics API, but without checking for the capability to accept it or checking for any errors.

nkast:
If we were to emulate XNA and enforce backward compatible code and Profile limitations then we will have to limit the rendertargets to 4 for HiDef profile, and 1 for Reach profile.
Cards support 8 RTs for 10.0 or higher, 4 RTs for 9.3, 1 RT for 9.1 & 9.2.

nkast:
Right now on MG you might get 4 RTs under Reach profile, depending on whether the device was created with 9.3 or 9.1, but the user can't check for it during runtime. The worst case, is when you think your code runs with Reach profile while you develop on your desktop (9.3) but fails when you move it to an older device or phone (9.1).

nkast:
I tested the behavior of XNA and got those exceptions when trying to set RTs.

```
NotSupportedException: XNA Framework Reach profile supports a maximum of 1 simultaneous rendertargets.
NotSupportedException: XNA Framework HiDef profile supports a maximum of 4 simultaneous rendertargets.
```

IMO, validation must be addressed by #1941. Not just for RT, but more broadly.

tomspilman:
@nkast - What platforms has this been tested on?

I see this which still seems to suggest only 4 RTs are supported:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs#L71

nkast:
I did a quick test to see if I can set 5 RT, I don't have a shader that uses more than 4. 
The person who requested this feature on community haven't reply since then. I don't suppose he tested this PR. Perhaps I should just drop it.

tomspilman:
> Perhaps I should just drop it.

Yeah... lets drop this unless we actually have a need and demand for it.

Few games use more than 4 because of older hardware and performance anyway.

Fixing this correctly is more complex than just incrementing a number.  We need additional unit tests, we need to account for OpenGL, and we need to consider what happens on platforms that don't support more than 4 targets.

PR #5306, opened 2016-11-18
Add documentation for GraphicsDevice.Draw* methods

### LINKED ISSUE #5294: Flickering. DrawUserIndexedPrimitives.
Hello!

I have some troubles with DrawUserIndexedPrimitives method.

Here you can see an example of flickering. (DirectX)
https://github.com/krasnotsvetov/MonogameRenderBugExample

If I want to set offset in vertex declaration data and indices in DrawUserIndexedPrimitives i will have flickering of mesh which this method will draw.

How can i fix it?
Why can it happend?

P.S: here is video:
https://www.youtube.com/watch?v=2CBfgcWhiuw

Jjagg:
I included remarks that should clarify behavior when a vertexOffset is passed following the discussion in #5294. 

fixes #5294.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -928,6 +928,13 @@ namespace Microsoft.Xna.Framework.Graphics
             DrawIndexedPrimitives(primitiveType, baseVertex, startIndex, primitiveCount);
         }
 
+        /// <summary>
+        /// Draw geometry by indexing into the vertex buffer.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @Jjagg !

Jjagg:
I don't know if you noticed @tomspilman, I also marked the existing DrawInstancedPrimitives to be obsolete and created an overload without the 2 parameters that we don't use (but are in the XNA API), same as what we did for DrawIndexedPrimitives.

tomspilman:
Yeah that isn't a problem... that was only a thing used by older versions of DirectX and won't be coming back.

PR #5310, opened 2016-11-21
[DesktopGL] Fix debugging from Windows

cra0zy:
Documentation states that the hint name is `SDL_HINT_WINDOWS_DISABLE_THREAD_NAMING`, but looking at the source code it seems that the hint is `SDL_WINDOWS_DISABLE_THREAD_NAMING` (source code const is getting assigned a different value from it's name). 

In the end I had the time to test this.

### DIFF
--- a/MonoGame.Framework/SDL/SDLGamePlatform.cs
+++ b/MonoGame.Framework/SDL/SDLGamePlatform.cs
@@ -51,6 +51,10 @@ namespace Microsoft.Xna.Framework
             if (version <= 204)
                 Debug.WriteLine ("Please use SDL 2.0.5 or higher.");
 
+            // Needed so VS can debug the project on Windows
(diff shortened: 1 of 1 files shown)


KonajuGames:
Why is this not default behaviour in SDL itself?  Not being able to debug by default is really odd.

mrhelmut:
Maybe @flibitijibibo could share some context, if I remember well, it's one of his patch.

cra0zy:
https://wiki.libsdl.org/SDL_HINT_WINDOWS_DISABLE_THREAD_NAMING
https://github.com/MonoGame/MonoGame/issues/4399#issuecomment-250747736

mrhelmut:
Could this be merged? Right now debugging from Windows is impossible.

tomspilman:
Done!

PR #5312, opened 2016-11-21
Implement GraphicsAdapter.IsProfileSupported()

nkast:
On 3.6 you have to use SharpDX to query for the GPU feature level or target just HiDef and let the app crash if HiDef is not supported (3.5 had HiDef as default if supported).

With this PR you can check for supported profile and enable HiDef. 
ex:
```
        public Game1()
        {
            graphics.PreparingDeviceSettings += Graphics_PreparingDeviceSettings;
        }

        private void Graphics_PreparingDeviceSettings(object sender, PreparingDeviceSettingsEventArgs e)
        {
            if(e.GraphicsDeviceInformation.Adapter.IsProfileSupported(GraphicsProfile.HiDef))
                e.GraphicsDeviceInformation.GraphicsProfile = GraphicsProfile.HiDef;
        }
```

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
@@ -116,6 +116,9 @@ namespace Microsoft.Xna.Framework.Graphics
 
         protected bool PlatformIsProfileSupported(GraphicsProfile graphicsProfile)
         {
+            if(UseReferenceDevice)
+                return true;
+
             switch(graphicsProfile)
             {
                 case GraphicsProfile.Reach:
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
@@ -5,11 +5,15 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using SharpDX.Direct3D;
+using SharpDX.DXGI;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
     partial class GraphicsAdapter
     {
+        SharpDX.DXGI.Adapter1 _adapter;
+
         private static void PlatformInitializeAdapters(out ReadOnlyCollection<GraphicsAdapter> adapters)
         {
             var factory = new SharpDX.DXGI.Factory1();
@@ -31,8 +35,6 @@ namespace Microsoft.Xna.Framework.Graphics
 
                     monitor.Dispose();
                 }
-
-                device.Dispose();
             }
 
             factory.Dispose();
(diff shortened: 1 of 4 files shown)


KonajuGames:
Merging. Thanks @nkast.  We'll get the OpenGL checks for HiDef support done separately.

tomspilman:
Ok... coming back at this now i think we are incorrectly testing for HiDef.

From a core XNA developer:

> XNA 4 Note: None of these feature levels directly maps to the XNA “Reach” or “HiDef” profiles. “Reach” is very similiar to Feature Level 9.1, and “HiDef” is designed for the Shader Model 3.0 capabilities of the Xbox 360 so it’s a bit more than Feature Level 9.3 offers but not as much as Feature Level 10.0.
https://blogs.msdn.microsoft.com/chuckw/2012/06/20/direct3d-feature-levels/

So us testing for feature level 10 is incorrect and not what XNA's HiDef means.

We need to test for feature level 9.3 and manually check for some additional features to match XNA behavior correctly.

tomspilman:
@nkast @Jjagg @KonajuGames - It is critical we fix this issue I mention above in some way... even if the fix is just a temp hack.  Until this is fixed the unit tests on the develop branch will be broken causing havoc with all other PRs.

KonajuGames:
#5424 fixes the unit tests.  It was creating reference rasterizer just for the visual tests, but other non-visual tests were also creating a graphics device.

nkast:
> Maybe the extra can be found comparing these:

The only difference I see is Shader Model 3.0. SM3 was replaced by SM4 on DX10.  DX10 also removed the DX9 CAPS, so there is no direct way to test for SM version. BTW, MGFX can't build ps_3_0 either, so we might as well set HiDef = 9.3 right now.

IMO it's more important to expand profiles and start enforcing it. For example, HiDef should not allow PS_4_0 or PS_5_0 shaders.

nkast:
Another one must be "Nonpowers-of-2 unconditionally"...

Jjagg:
Is the idea to keep Reach/HiDef to map the same as XNA? Semantically it's not really HiDef anymore

Jjagg:
Maybe we should have XnaHiDef or something for compatibility. We'd have to inform people of this if we do change HiDef to be more up to modern standards.

nkast:
XNA was written on DX9 so in theory we should be able to implement HiDef on DX9+SM3 cards.
This seems ok from OpenGL which still supports SM3 but from DX point , SM3 was DOA. 
Probably the only cards that supported it were the ones released after XBOX360 and soon replaced by DX10 cards. Look at [this graph](https://www.khronos.org/opengles/). Not sure what this `10level9` is but I think the only DX9+SM3 cards are those in the orange area (it's more like a thin line in the middle). 

We can't check for SM3 support, DX10 removed the caps and introduced SM4, a superset of SM3. A possible workaround would be to load a SM3 and see if it fails, but...
We can't build SM3 (ps_3_0) from the DX11 compiler.
At this point I'm afraid that DX11 would just refuse to load a SM3 shader even if we had a trick to build one.

Issue #5280, opened 2016-11-06
'Marshal.SizeOf(Type)' is obsolete errors in MG source.


DDReaper:
Full error detail
Warning	CS0618	'Marshal.SizeOf(Type)' is obsolete: 'SizeOf(Type) may be unavailable in future releases. Instead, use SizeOf<T>(). For more info, go to http://go.microsoft.com/fwlink/?LinkID=296515'	MonoGame.Framework.WindowsUniversal	\MonoGame.Framework\Graphics\TextureCube.DirectX.cs	105	

If no one else is looking at this, I'll do a quick PR as the fix is quite easy.  Will just need to validate if the fix is applicable to all platforms.

tomspilman:
> I'll do a quick PR as the fix is quite easy. Will just need to validate if the fix is applicable to all platforms.

Sure...  just use `#if`s to fallback to `Marshal.SizeOf(Type)` on platforms that don't support it.

nkast:
Possible conflict with https://github.com/MonoGame/MonoGame/pull/5121 FYI @Jjagg 

> Sure... just use #ifs to fallback to Marshal.SizeOf(Type) on platforms that don't support it.

From the documentation:
Universal Windows Platform: Available since 8.1
.NET Framework: Available since 4.5.1
Portable Class Library: Supported in: portable .NET platforms
Windows Phone: Available since 8.1

Should we move this to ReflectionHelper class? Technically Marsal.SizeOf() has nothing to do with reflection but I see the same pattern of changes between the WINRT framework and older .NET framework. 2MGFX has a MarsalHelper class.

Jjagg:
> Possible conflict with #512

That's okay. I need to revisit that PR anyway.

DDReaper:
I'm all for moving this to the reflection helper.  it's where platform specific related functionality should go to avoid complexity in the base classes.
I'll see if I can start work on that tonight and get some testing done.

tomspilman:
> Should we move this to ReflectionHelper class? 

Yeah ideally we stick this in some helper class to centralize the `#if`d code and not have that mess duplicated everywhere.

tomspilman:
@DDReaper - Also i've noticed on some platforms `Marshal.SizeOf()` can be sort of expensive to call.

This is how i solved that:

```
using System;
using System.Runtime.InteropServices;

namespace Microsoft.Xna.Framework.Utilities
{
    internal static class SizeOf<T>
    {
        static int _sizeOf;

[...]

PR #5316, opened 2016-11-22
Marshal size of refactor

DDReaper:
PR to address #5280

All elements supporting Generics (<T>) aligned to new ReflectionHelpers.SizeOf<T> function.
SizeOf<T> tested against platforms for support. Linux / Windows falls back to old SizeOf(Type) method

IndexBuffer however has no generics, so a fall back method in ReflectionHelpers added to enable caching of Marshal.SizeOf implemented using a dictionary.

Both IndexBuffer and VertexBuffer should really support generics fully but this would be a breaking change.

### DIFF
--- a/MonoGame.Framework/Utilities/ReflectionHelpers.cs
+++ b/MonoGame.Framework/Utilities/ReflectionHelpers.cs
@@ -158,7 +158,7 @@ namespace Microsoft.Xna.Framework.Utilities
 
             static SizeOf()
             {
-#if LINUX || WINDOWS
+#if LINUX || WINDOWS || WEB
                 var type = typeof(T);
                 _sizeOf = Marshal.SizeOf(type);
 #else
--- a/MonoGame.Framework/Utilities/ReflectionHelpers.cs
+++ b/MonoGame.Framework/Utilities/ReflectionHelpers.cs
@@ -6,163 +6,183 @@ using System.Runtime.InteropServices;
 
 namespace Microsoft.Xna.Framework.Utilities
 {
-	internal static class ReflectionHelpers
-	{
-		public static bool IsValueType(Type targetType)
-		{
-			if (targetType == null)
-			{
-				throw new NullReferenceException("Must supply the targetType parameter");
-			}
+    internal static class ReflectionHelpers
+    {
+        public static bool IsValueType(Type targetType)
+        {
+            if (targetType == null)
+            {
+                throw new NullReferenceException("Must supply the targetType parameter");
+            }
 #if WINRT
-			return targetType.GetTypeInfo().IsValueType;
+            return targetType.GetTypeInfo().IsValueType;
 #else
 			return targetType.IsValueType;
 #endif
-		}
+        }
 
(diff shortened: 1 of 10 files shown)


MonoGame.Framework/Utilities/ReflectionHelpers.cs
```diff
@@ -2,60 +2,61 @@
...
 #if WINRT
-			return targetType.GetTypeInfo().IsValueType;
+            return targetType.GetTypeInfo().IsValueType;
 #else
 			return targetType.IsValueType;
 #endif
-		}
-
-		public static Type GetBaseType(Type targetType)
-		{
```
tomspilman: Let's not try to cache this case... we use it infrequently and it just introduces another non-thread safe global.

DDReaper: Updated.
Although long term this still leaves a risk for when the function is deprecated in modern .NET
IndexBuffer and VertexBuffer should really be updated with Generics, however this would be a breaking change.

DDReaper:
OK, what is the MonoGame.Framework.Web project?
When I build source it does not generate that platform?  Guessing it simply needs adding to the #IF

MonoGame.Framework/Utilities/ReflectionHelpers.cs
tomspilman: > Although long term this still leaves a risk for when the function is deprecated in modern .NET

Understood.

tomspilman:
> Guessing it simply needs adding to the #IF

Yeah assume it is like Windows... it should use the old method.

DDReaper:
Missed WindowsPhone 7, patched now
@tomspilman there is probably a case for dropping this platform now, like with Ouya.

DDReaper:
Nudge @tomspilman all checks passed now.

tomspilman:
Thanks.... merging!

tomspilman:
@DDReaper - So we have an issue with the `Marshal.SizeOf<T>()` in the console platforms.  They still require the old `Marshal.SizeOf(type)` path.

The hack fix is to just add that console platform into the existing `#if LINUX || WINDOWS || WEB || SILVERLIGHT` conditional.  However this just makes it hard to port new platforms as we have to make changes to the public repo and not just the console repo.

We need a better solution than platform `#if`s to decide what `Marshal.SizeOf` to use.  The possibilities I see right now:

 - A platform function on a partial class like we do in other places.
 - A special `MARSHAL_SIZEOF_T` that is set per platform in the Protobuild definition.

I don't think we can do a platform function on a nested class like `SizeOf`... so either we don't take that path or move `SizeOf` out of `ReflectionHelpers`.

DDReaper:
I think the "Partial" class route is probably the best way to go.  I'll make the base a partial class and create a default and Legacy sub implementation.
Then on console you can swap out the default for the console implementation.

I'll see if I can get that done this morning

DDReaper:
Submitted new PR with the refactoring.
Will just need to update the build files fro Console projects to use the .Legacy extension

PR #5319, opened 2016-11-23
Misc Pipeline Fix

tomspilman:
Two simple pipeline fixes/improvements.

Fixed case in `FontDescriptionProcessor` where we were consuming fatal exceptions instead of letting them be reported to the user as errors.

Protect against crashes when `message` or `messageArgs` is null in `ConsoleLogger.LogWarning()`.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
             // Get the platform specific texture profile.
             var texProfile = TextureProfile.ForPlatform(context.TargetPlatform);
 
-			try {
+            {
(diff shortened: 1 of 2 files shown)


MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
```diff
@@ -133,8 +133,9 @@ public FontDescriptionProcessor()
 
                 output.Texture.Faces[0].Add(face);            
 			}
-			catch(Exception ex) {
-				context.Logger.LogImportantMessage("{0}", ex.ToString());
+			catch(Exception ex)
+			{
+			    throw;
```
KonajuGames: Rather than simply rethrowing it, this clause doesn't need to be there at all.  The `Exception ex` variable is also not needed and will be issuing a compiler warning about an unused variable.

tomspilman:
@KonajuGames - Fixed.

KonajuGames:
Merging.

PR #5321, opened 2016-11-23
Remove WindowsPhone

DDReaper:
Removed WindowsPhone and WindowsPhone81 from the framework
Also removed the old Windows 8 Universal project as it just left it with "Windows8"


DDReaper:
A thought does occur whilst stripping this out. 
Windows 8 should also follow the same path, or at the very least be updated to 8.1 as the current template DOESN'T work and requires you to update it to 8.1

DDReaper:
@tomspilman @dellis1972 @KonajuGames 
looks like the build server definitions somewhere else are still looking for WindowsPhone / WindowsPhone81
Can you help?

tomspilman:
I think we should hold until after 3.6 to remove WP... but we should get this PR working.

> looks like the build server definitions somewhere else are still looking for WindowsPhone / WindowsPhone81

Yeah it is baked into the build step.

I suggest for now leave the `build_windowsphone` and `build_windowsphone81` steps in the Nant file... just empty them so they don't do anything.

DDReaper:
Fair enough. In retrospection then @tomspilman . If we're waiting until after 3.6, then we should also remove the Windows 8 platform.
It needs work to make it work again (template currently broken), so should we fix that for 3.6 or leave it and remove after?

tomspilman:
> then we should also remove the Windows 8 platform.

I don't mind leaving Win8 for a bit longer...  it is still used on over 10% of PC world wide.    We should just fix that template.

DDReaper:
OK, I'll see if I can do a new PR to fix the Windows 8 (and Universal) templates.
Will also check there isn't anything wrong with the WP templates if they are staying for 3.6

DDReaper:
Ok, 2 steps forward, 1 step back.

WindowsPhone (v8) template does have to go.  Neither VS2013 or VS2015 support it any more.
The Windows 8 Universal project only uses the 8.1 project as well which is still functional.

So Thoughts @tomspilman , it may be easier to start a new PR to only remove WP8 and leave WP81 in place.

tomspilman:
> Not removing the DLL's, just the template

Yeah we could maybe do that...  but don't we install to VS2012 which did support WP8?

DDReaper:
Unlikely as you can't submit new WP8 projects to the store anymore as far as I know

nkast:
@DDReaper 
> WindowsPhone (v8) template does have to go. Neither VS2013 or VS2015 support it any more.
They do. VS2015 dropped support for Windows 8.0 but still support WP8.0. I don't know if the store stopped accepting WP8.0 but the templates are still useful for sideloading apps. Let's drop the templates when we drop support for the platform. 

The problem is the WP8 had some issues on MG3.5  and recent changes on 3.6 broke it completely, so  either fix it or drop it.

@tomspilman 
I can submit some fixes that I have but I don't want to waste my time if they won't make it into 3.6.

Sidenote: recent hololens SDK works a lot like WP8.0, the framework provides you with the backbuffer/rendertarget (two of them actually). We might be able to share some code on how WP8.0 is implemented in the GraphicsDevice.

DDReaper:
I'm still for dropping at least the old WindowsPhone (WP8.0) project entirely within 3.6 @nkast 
Although I'd probably drop this PR and start again.

I'm still on the fence about the 8.1 projects.

nkast:
> I'm still for dropping at least the old WindowsPhone (WP8.0) project entirely within 3.6 @nkast

Whichever way this goes I am good with it.

BrianPeek:
My opinion that nobody asked for:  Ditch WP80 and Win80, Keep 8.1 and higher for both.  There are a LOT of phones that are still on 8.1 and will never get to 10.

tomspilman:
> Ditch WP80 and Win80, Keep 8.1 and higher for both

I think that is a reasonable thing to do... I would be good with making that change immediately.

Thanks @BrianPeek!

KonajuGames:
> My opinion that nobody asked for:

Always a valued opinion though. Agreed with the suggestion too.

DDReaper:
Right, well win8.0 is gone already with the fix pr I put forward, upgrading the template to win8.1.

However we can't fix the mg framework project as protobuild doesn't support win8.1 but I don't think it matters really as the dll works all the same for both.

Will sort out another pr to drop wp80 and close this one.

And @brianpeek your words of wisdom are always valid, especially when it comes to Windows stuff 😎

Installers/Windows/MonoGame.nsi
```diff
@@ -106,6 +106,11 @@ Section "MonoGame Core Components" CoreComponents ;No components page, name is n
   File '..\..\MonoGame.Framework\bin\Android\AnyCPU\Release\*.dll'
   File '..\..\MonoGame.Framework\bin\Android\AnyCPU\Release\*.xml'
   
+  ; Install OUYA Assemblies
```
nkast: Something's wrong with this branch, there are 10 references to the OUYA platform.

Installers/Windows/MonoGame.nsi
```diff
@@ -106,6 +106,11 @@ Section "MonoGame Core Components" CoreComponents ;No components page, name is n
...
+  SetOutPath '$INSTDIR\Assemblies\OUYA'
+  File '..\..\MonoGame.Framework\bin\Ouya\AnyCPU\Release\*.dll'
+  File '..\..\MonoGame.Framework\bin\Ouya\AnyCPU\Release\*.xml'
```
nkast: Something's wrong with this branch, there are 10 references to the OUYA platform.

KakCAT:
According to adduplex, this month 78% of the phones are on WP8.1 and 13% on W10 mobile (and I've always thought data is biased in favor of W10mobile according to my data) . WP8.1 is here to stay for a loooong time, due to the disaster W10Mobile represents in lots of departments. Its installed userbase is even shrinking. 

I wish some critical fixes would have been accepted before ditching WP8, but if people could live with the for 3 years, it won't be a problem :)

data source: http://blog.adduplex.com/2016/10/adduplex-windows-device-statistics-report-october-2016.html

Installers/Windows/MonoGame.nsi
DDReaper: Will check in the new PR. But I already mentioned it to the Ouya PR 🗡

PR #5332, opened 2016-11-28
Fix bug in ByteBufferPool

Jjagg:
This class is used for dynamic sound effect buffers. `_freeBuffers` are ordered by size and the lookup depends on it because it does a binary search. When implementing this I made a silly mistake that would mess up the order.


tomspilman:
Thanks @Jjagg !

Issue #4897, opened 2016-06-02
Input seriously lagging in UWP


MartinZikmund:
Keyboard and mouse input is seriously lagging in UWP games.
When you use `Keyboard.GetState().IsKeyDown( ... )` you will find out that it returns true usually even several Updates after the key is released.
Similar lag occurs for mouse state.

This is quite serious as it renders the framework unusable - or rather the games hardly playable.

Edit: I have confirmed this behavior with another developer, so it is not just an issue on my side.

Ragath:
I believe I have encountered this issue too, although in my case the keys appear to be permanently stuck until you press another key or move the mouse.

VitezslavImrysek:
Same problem here. This issue affects `TouchPanel` as well.
My guess is that the issue is somehow connected to how MonoGame interacts with `Dispatcher` (and therefore input event queue processing) as it also affects `CoreWindow` `KeyUp`/`KeyDown` events and also `GetKeyState` method.

tomspilman:
I'll give this a look on my laptop later tonight.

tomspilman:
So I made a little test app gathering keyboard state in Update and drawing the string in Draw.  I saw no input lag of any kind.

Note i'm running Windows 10 Pro with version 10.0.10586.318 and the 

@MartinZikmund @VitezslavImrysek  - What version of Windows 10 are you on?

tomspilman:
Here is my test code:

Update:
```
            var kstate = Keyboard.GetState();

            downKeys = string.Empty;
            var keys = kstate.GetPressedKeys();
            foreach (var k in keys)
                downKeys += k + ",";

            if (kstate.IsKeyDown(Keys.OemQuestion))
                downKeys += "?";
```

Draw:
```
           spriteBatch.Begin(sortMode:SpriteSortMode.Immediate);
           spriteBatch.DrawString(font, "Keys: " + downKeys, new Vector2(200, 300), Color.White);
           spriteBatch.End();
```

Danthekilla:
We are having the same issue I think.

Here is a video of the issue with my input visible from a input displaying program.

When I press Keys.Up (via Keyboard.GetState(); which eventually hooks into CoreWindow.KeyDown) I move the background forward and back. You can also see the input system moving the selection caret.

https://www.youtube.com/watch?v=vHr91Sr_W_g&feature=youtu.be

[...]

Issue #5300, opened 2016-11-16
[UWP] Mouse input blocks/delays keyboard input


MuletTheGreat:
InputEvents.cs is where the keyboard/mouse events are attached to allow MonoGame to collect input.

If the mouse is moving/clicking then keyboard input is delayed until it stops. This includes key released states, so it effects first person game very badly.

Attaching to the CoreWindow's keyboard and mouse events shows this.
Attached to the CoreIndependentInputSource for the Mouse, with the keyboard coming from the CoreWindow (standard practice) also causes this.

This did not occur in Windows 8/8.1 and only occurs in Windows 10.

tomspilman:
Fixed in #5520.

PR #5333, opened 2016-11-28
Low latency Mouse input

nkast:
Create a working thread to process Mouse input.

Fixes https://github.com/MonoGame/MonoGame/issues/4897, https://github.com/MonoGame/MonoGame/issues/5300, http://community.monogame.net/t/keyboard-input-caused-by-mouse/8203/9.

I finally found a way around this bug https://wpdev.uservoice.com/forums/110705-universal-windows-platform/suggestions/16521550-fix-coreindependentinputsource
The trick is that we left out Touch devices from CoreIndependentInputSource.

### DIFF
--- a/MonoGame.Framework/Windows8/InputEvents.cs
+++ b/MonoGame.Framework/Windows8/InputEvents.cs
@@ -218,12 +218,13 @@ namespace Microsoft.Xna.Framework
 
             var state = point.Properties;
 
-            Mouse.PrimaryWindow.MouseState.X = x;
-            Mouse.PrimaryWindow.MouseState.Y = y;
-            Mouse.PrimaryWindow.MouseState.ScrollWheelValue += state.MouseWheelDelta;
-            Mouse.PrimaryWindow.MouseState.LeftButton = state.IsLeftButtonPressed ? ButtonState.Pressed : ButtonState.Released;
-            Mouse.PrimaryWindow.MouseState.RightButton = state.IsRightButtonPressed ? ButtonState.Pressed : ButtonState.Released;
-            Mouse.PrimaryWindow.MouseState.MiddleButton = state.IsMiddleButtonPressed ? ButtonState.Pressed : ButtonState.Released;
+            Mouse.PrimaryWindow.MouseState = new MouseState(x, y, 
+                Mouse.PrimaryWindow.MouseState.ScrollWheelValue + state.MouseWheelDelta,
+                state.IsLeftButtonPressed ? ButtonState.Pressed : ButtonState.Released,
+                state.IsMiddleButtonPressed ? ButtonState.Pressed : ButtonState.Released,
+                state.IsRightButtonPressed ? ButtonState.Pressed : ButtonState.Released,
+                state.IsXButton1Pressed ? ButtonState.Pressed : ButtonState.Released,
+                state.IsXButton2Pressed ? ButtonState.Pressed : ButtonState.Released);
         }
 
         public void UpdateState()
--- a/MonoGame.Framework/Windows8/InputEvents.cs
+++ b/MonoGame.Framework/Windows8/InputEvents.cs
@@ -8,9 +8,11 @@ using Microsoft.Xna.Framework.Input;
 using Microsoft.Xna.Framework.Input.Touch;
 using Windows.Devices.Input;
 using Windows.Graphics.Display;
+using Windows.System.Threading;
 using Windows.UI.Core;
 using Windows.UI.Input;
 using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Input;
 
 namespace Microsoft.Xna.Framework
@@ -31,6 +33,33 @@ namespace Microsoft.Xna.Framework
             window.VisibilityChanged += CoreWindow_VisibilityChanged;
             window.Activated += CoreWindow_Activated;
             window.SizeChanged += CoreWindow_SizeChanged;
+
+            if (inputElement is SwapChainPanel || inputElement is SwapChainBackgroundPanel)
+            {
+                // Create a thread to precess input events.
+                var workItemHandler = new WorkItemHandler((action) =>
+                {
+                    // We subscribe only to Mouse & Pen. Touch is still processed from inputElement.
+                    // Currently PointerVisualizationSettings don't apply to CreateCoreIndependentInputSource 
+                    // and ContactFeedback is always visible when we create a coreIndependentInputSource for Touch.
+                    var inputDevices = CoreInputDeviceTypes.Mouse | CoreInputDeviceTypes.Pen;
+
+                    CoreIndependentInputSource coreIndependentInputSource;
+                    if (inputElement is SwapChainBackgroundPanel)
+                        coreIndependentInputSource = ((SwapChainBackgroundPanel)inputElement).CreateCoreIndependentInputSource(inputDevices);
+                    else
+                        coreIndependentInputSource = ((SwapChainPanel)inputElement).CreateCoreIndependentInputSource(inputDevices);
+                    
+                    // PointerVisualizationSettings don't apply to CreateCoreIndependentInputSource and ContactFeedback is 
+                    // always visible. If this bug get fixed we can add the Touch & Pen InputDeviceTypes.
+                    
+                    coreIndependentInputSource.PointerPressed += CoreWindow_PointerPressed;
+                    coreIndependentInputSource.PointerMoved += CoreWindow_PointerMoved;
+                    coreIndependentInputSource.PointerReleased += CoreWindow_PointerReleased;
+                    coreIndependentInputSource.Dispatcher.ProcessEvents(CoreProcessEventsOption.ProcessUntilQuit);
+                });
+                var m_inputLoopWorker = ThreadPool.RunAsync(workItemHandler, WorkItemPriority.High, WorkItemOptions.TimeSliced);
+            }
 
             if (inputElement != null)
             {


tomspilman:
@BrianPeek - Any thoughts on this fix?

nkast:
This PR definitely improves the situation with Keyboard events but to be fair the only thing I can guarantee is on time delivery of mouse events. :smirk:  Key events might still get lag in case of heavy load, because the game loop continuously freezes the message queue for at least 16ms and the next render message has priority over queued keyboard messages.

Danthekilla:
I have tested this PR in my games, and it only seems to have a small improvement in keyboard input.

When moving the mouse I can still get keyboard events to be delayed by up to 20 seconds and under normal play 1-2 seconds of delay is normal.

This is still 100% unshippable sadly.

Thanks for all the work you have put into this @nkast we at least understand more about the issue now. 

> because the game loop continuously freezes the message queue for at least 16ms and the next render message has priority over queued keyboard messages.
Is there a way of changing this so we can prioritize the keyboard input over the render message?

Danthekilla:
When running at 60fps, this is the distribution of mouse input event times in my project with the new mouse input logic.

The average sample time is 4ms. With a fair amount of variance.

[Mouse Input Image](http://imgur.com/82dGlqu)
mouseinput

nkast:
I wrote a simple test to track both mouse and keyboard events, but can't record a video for UWP, here is the code for testing https://gist.github.com/nkast/ffe21047c77375d1f4a86a09a4e0d69c

I found that when this happens, mouse events lag as well. Actually they stop for a while and after a couple of seconds we start receiving them with a big delay, finally the key events arrive as well. With this PR I don't see any lag, maybe I should add some Thread.Sleep to simulate a real game.

@Danthekilla What exactly does that graph show. and how does it compare with the develop branch?

> Is there a way of changing this so we can prioritize the keyboard input over the render message?

Hm, I don't know. UWP are very restrictive with Threads ,etc. The right solution is to move the render/gameloop in it's own thread. That will cause a lot of sync issues with other events, like SizeChanged,OrientationChanged,Exit, etc. The problem is known since WP8.1 came out. 
When I moved the game loop in it's own loop I found that the UI thread gets lower priority and touch events were very laggy, so I tried coreindependentinputsource and found the bug I mention above on tablet. So I just stick with WP8.0 on phones and light rendering on W8.1 until the bug get's fixed. Eventually I just gave up.

Danthekilla:
>What exactly does that graph show. and how does it compare with the develop branch?

That is the time between mousepointermoved event callbacks. Or how frequently the mouse position is getting updated via the coreindipendentinputsource thread.

nkast:
I updated the testing code to show frame number which makes it clear whether the game still draws and  when you get frames drop. Also added Sleep() in update/draw. 

When there is no load (delayUpdateMs = delayDrawMs = 0) mouse works ok but a key event triggers a lag on both mouse & keyboard. This PR fixes it.

When you increase the load (~2ms) you get constant lag on both mouse & keyboard, even with this PR. Setting fixedTimeStep to false together with this PR fixes it, but that's up to a certain limit. After some threshold mouse lags constantly.

BrianPeek:
I would take a step back and ask one question: is there a reason we use a XAML app for MonoGame instead of doing it via a CoreApplication which would eliminate the SwapChainPanel altogether?

nkast:
XAML projects are the only ones that let's you run managed code. I am not sure if that changed with w10  but it was true when MG added support for Windows 8 & Windows Phone 8.

BrianPeek:
Yeah, not true with Win10 UWP.  Using CoreApplication without the overhead of XAML might make things better.  I can try to throw together some code if/when I have some time...

Danthekilla:
>What exactly does that graph show. and how does it compare with the develop branch?

Here is a comparison of the frequency of the mouse events between the dev branch and this pull request. It is probably fairly close to the latency of the two systems also. It shows a massive improvement with the new system, even if the new system doesn't fix the keyboard input issues.

mouseinput

BrianPeek:
I threw together a very quick and dirty UWP rewrite which ditches XAML and the SwapChainPanel, going to a straight CoreApplication architecture.  I wouldn't call it complete, tested, or ready to use, but I'm curious if the people here might want to test it to see if mouse/keyboard improves with the change?

nkast:
> Here is a comparison of the frequency of the mouse events between the dev branch and this pull request. 

I assume you were keep moving the mouse while collecting those data or moved the mouse in and out across the window. 

The result however look like what I expected! Good work!

BrianPeek:
Ok, I have a branch in my own forked repo here:

[https://github.com/BrianPeek/MonoGame/tree/CoreApplication](https://github.com/BrianPeek/MonoGame/tree/CoreApplication)

There's a new UWP template (same name as the other), so you'll need to make sure that gets built and copied to the right place.  Otherwise, the remainder of the changes are in source and the protobuild stuffs should be updated to grab the new/changed files.

Very curious to see how this compares to the above, so if anyone has time to test, please give it a go!  Thanks!

Danthekilla:
>I assume you were keep moving the mouse while collecting those data

@nkast  Yeah I was moving the mouse in circles around the game window while collecting the data. It is nice to see it make such a difference, it feels a lot better in a fps game too.

nkast:
> eah, not true with Win10 UWP. 

Indeed. MSDN has a note on CoreApplication.
`Windows Phone 8
This API is supported in native apps only, except for the Properties property, which is always available.`

I guess we can replace the old code soon. The WP8.0 doesn't use SwapChainPanel/ SwapChainBackgroundPanel anyway so we can start right now. 

I wonder if we can replace it or support both types side by side..
Every so often somebody asks how to navigate between XAML pages and I image it's common to combine xaml elements with MonoGame, despite the fact that this is not portable. I think also Guide & Input keyboard was implemented in XAML. And I think some other features like Tiles were not supported on unmanaged projects, maybe that changed too.

Danthekilla:
>I imagine it's common to combine xaml elements with MonoGame

@nkast  We use XAML elements for our login UI and a few other things like the guide. But I am willing to rewrite all that if it means proper keyboard input.

Danthekilla:
@BrianPeek I am giving this a test now. I will let you know how it goes.

Danthekilla:
@BrianPeek When trying to compile your branch of monogame I get this error

`1>------ Build started: Project: MonoGame.Framework.WindowsUniversalCoreApp, Configuration: Debug Any CPU ------
1>D:\Desktop\2.0 Projects\Games\MonoGame-CoreApplication\MonoGame.Framework\MonoGame.Framework.WindowsUniversalCoreApp.csproj : XamlCompiler error WMC1006: Cannot resolve Assembly or Windows Metadata file 'System.Runtime.dll'
1>C:\Program Files (x86)\MSBuild\Microsoft\WindowsXaml\v14.0\8.2\Microsoft.Windows.UI.Xaml.Common.targets(264,5): Xaml Internal Error error WMC9999: Type universe cannot resolve assembly: System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
`

I can't see why, it seems to have all its references.

BrianPeek:
Hm. Are you loading the MonoGame.Framework.WindowsUniversal.sln solution after running protobuild?  That's building here...

Danthekilla:
All fixed, it was my mistake. I renamed the project without renaming the json file.

Danthekilla:
Based on my testing so far this seems to solve the issue 100% of the time. (and reduces general keyboard latency to boot!)

I am in the process of converting my main game to use it, its a slow process as we were using a lot more XAML than I originally thought, also I have had some issues converting my csproj so I am having to rebuild the project starting from the sample as a base. But it is getting there.

I think this should be cleaned up and possibly merged with the main branch after some more testing.

BrianPeek:
Neat!  :)  Did you notice any issues with backbuffer size, window resizing, or anything else?  I didn't dig into that code much/at all...

@tomspilman, any thoughts on how to proceed?  Do we want to keep the XAML project and add this one in addition, maybe in "experimental state" until it gets tested further?  I can do a different checkin that just adds this instead of replaces the XAML template and we can try to figure out what I screwed up...

Danthekilla:
I have just finished converting my game to use it, its about 95% working.

There are some issues regarding backbuffer size and window resizing.

As far as I can tell some of the events that would normally get called on a window resize are not getting called. And even though I requested (and the game thinks it has) a 720p backbuffer the window is 1024 by 768 (or something similar)

BrianPeek:
Yup, you're seeing about what I saw...wrong backbuffer, window sizing not working.  I'll look into it.  But, thanks for testing and verifying that the other stuff is improved!

tomspilman:
> @tomspilman, any thoughts on how to proceed?

Well I haven't been following this issue very closely.  My first question is why is broken on a XAML project to begin with?  Are we not supposed to use XAML projects in games?  That was a valid path when we started supporting UWP.

BrianPeek:
XAML is a 100% valid path.  Unity offers the option of either XAML or D3D exports.  They each have pros and cons.  Good thread here:

https://community.unity.com/t5/Windows/Your-Unity-games-and-XAML-How-many-of-you-actually-use-it/td-p/2432126

In terms of the keyboard/mouse issues mentioned way up the thread, I don't have a better way to address, so I thought I'd try this approach which eliminates the XAML/compositing/etc. overhead.  

The background thread to handle input as described above is the best way to handle [low-latency input for XAML apps](https://github.com/Microsoft/Windows-universal-samples/tree/master/Samples/LowLatencyInput), but it appears it's still not as fast as removing the XAML stack entirely, at least with MonoGame.

Danthekilla:
>The background thread to handle input as described above is the best way to handle low-latency input for XAML apps
The issue is that there is no way to get keyboard input via the low latency input methods only mouse input can be obtained. I cannot find an alternate method for keyboard input that would be valid for the store.

Danthekilla:
After some more testing I have found that my games framerate is much more stable and about 18% higher overall.

Also I think there is a bug with mouse positioning when the screen has been scaled (likely due to the issues with the backbuffer sizing)

nkast:
Well , the right way to fix XAML input and all the other issues is to move the rendering in it's own thread.
I did a few tests and it seems to fix the keyboard lag, but we will face a few issues.

Touch input must be moved to it's own thread too, a test I did more than a year ago on a phone shows that touch from main thread will lag. But coreIndependentInputSource is still buggy. We have to accept that performance on tablets will suffer until MS fixes the issue.

We can't call GetForCurrentThread(), GetForCurrentView() and neither can the user from withing the game class (Initialize/Update/Draw,etc).

We must synchronize OnSwapChainPanelSizeChanged, OnRotationChanged,etc. Draw() will fail if at the same time ApplyChanges() rebuilds rendretarget, resizes the swapChain, etc.   

#4108 solves the above problem by locking the Tick() function, which causes deadlocks when we call Dispatcher.RunAsync() and then await for the result. For example, call KeyboardInput.Show().Wait() and then resize the window. The game freeze and the messagequeue locks forever on the resize event. MG also uses Dispatcher.RunAsync() is in the MediaPlayer but there is no wait there so it's probably ok.

Developers who use XAML for Ads, UI, etc will now have to update their elements through the UIThread, CoreDispatcher must me accessible from Game.Window.

nkast:
@Danthekilla Can you try this PR along with #4207 and test it with release build?
I found that my test suffers because of Task.Delay() and when I use TimerTool app it runs ok. At least on a release build. 
https://vvvv.org/contribution/windows-system-timer-tool

Another trick that seems to improve the situation is calling Present() with DoNotWait flag 
(I had to catch a "GPU is already in use" Exception and retry in a while loop, still don't know how to test for this case).

nkast:
@Danthekilla 
>Is there a way of changing this so we can prioritize the keyboard input over the render message?

I think something like #5352 will do.

nkast:
Closed via #5520

PR #5335, opened 2016-11-28
Marshal SizeOf Refactoring

### LINKED ISSUE #5280: 'Marshal.SizeOf(Type)' is obsolete errors in MG source.

DDReaper:
Split the ReflectionHelpers out in to a partial class implementation
Update based on #5316
Fixes #5280

### DIFF
--- a/MonoGame.Framework/Utilities/ReflectionHelpers.Default.cs
+++ b/MonoGame.Framework/Utilities/ReflectionHelpers.Default.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.IO;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Xna.Framework.Utilities
+{
+    internal static partial class ReflectionHelpers
+    {
+        /// <summary>
+        /// Generics handler for Marshal.SizeOf
+        /// </summary>
+        internal static class SizeOf<T>
+        {
+            static int _sizeOf;
+
+            static SizeOf()
+            {
+                _sizeOf = Marshal.SizeOf<T>();
+            }
+
+            static public int Get()
+            {
+                return _sizeOf;
+            }
+        }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


DDReaper:
Over to you @tomspilman to check with Console builds

tomspilman:
Yeah this will work... it allows me to pick the right file on XB1/Vita/etc based on what is available there.

Thanks @DDReaper !

MonoGame.Framework/Utilities/ReflectionHelpers.Default.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Generics handler for Marshal.SizeOf
+        /// </summary>
+        internal static class SizeOf<T>
+        {
+            static int _sizeOf;
+
+            static SizeOf()
+            {
+                _sizeOf = Marshal.SizeOf<T>();
```
tomspilman: You know I was thinking that `Marshal.SizeOf<T>()` would internally do the same optimization we're doing here.  But actually it doesn't:

https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs#L256

I don't see why they even made this change really... seems pointless.

DDReaper: I did wonder why they were changing the API as it doesn't seem to offer any improvement.
However the old method is now marked as deprecated.

PR #5340, opened 2016-11-29
Fallbacks if the preferred render target formats are not supported

KonajuGames:
DirectX 11 (DXGI 1.1) does not support 16-bit surface formats Bgr565, Bgra5551 or Bgra4444 for render targets.  DirectX 11.1 (DXGI 1.2) does support these surface formats for render targets.
https://msdn.microsoft.com/en-us/library/windows/desktop/ff471325(v=vs.85).aspx

This change implements a check for the support of the requested format and falls back to `SurfaceFormat.Color` and `DepthStencilFormat.Depth24Stencil8` if the requested formats are not supported.


KonajuGames:
Referenced in forum post.
http://community.monogame.net/t/cannot-create-a-texture2d-with-a-format-of-bgra4444-directx-windows-store-windows-uwp/8260

This was fallabck behaviour from XNA we were missing.

nkast:
@KonajuGames Is it possible to move this into GraphicsAdapter and implement QueryRenderTargetFormat()/QueryBackBufferFormat() ? 

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.graphicsadapter.querybackbufferformat.aspx
https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.graphicsadapter.queryrendertargetformat.aspx

KonajuGames:
That would make sense.  I'll do that.

nkast:
Thanks! That would be helpful. 
Later we can implement `FindBestDevice()`/`RankDevices()` to choose the default Adapter based on `QueryBackBufferFormat()` and `IsProfileSupported()` #5312.

KonajuGames:
Direct3D 11's Device has CheckFormatSupport(), but this needs a device created first, and GraphicsAdapter.QueryBackBufferFormat() needs to be able to be used before creating the device.  I'll have to find some other way to check for supported formats.  Perhaps somewhere in DXGI?

nkast:
Oh, I see. I don't know about DXGI. Maybe we have to temporally create a device for each one of the adapter. In  any case it doesn't sound like a good path to take for when we create rendertargets. I think we can see that later.

KonajuGames:
This PR has now been updated to add support for `GraphicsAdapter.QueryBackBufferFormat()` and `GraphicsAdapter.QueryRenderTargetFormat()`.

KonajuGames:
I believe the reference rasterizer supports all features, so fallbacks likely won't be used.  I'll see if I can try it though.

MonoGame.Framework/Graphics/RenderTarget2D.cs
```diff
@@ -26,6 +26,8 @@ private bool SuppressEventHandlerWarningsUntilEventsAreProperlyImplemented()
 	    public RenderTarget2D(GraphicsDevice graphicsDevice, int width, int height, bool mipMap, SurfaceFormat preferredFormat, DepthFormat preferredDepthFormat, int preferredMultiSampleCount, RenderTargetUsage usage, bool shared, int arraySize)
 	        : base(graphicsDevice, width, height, mipMap, preferredFormat, SurfaceType.RenderTarget, shared, arraySize)
 	    {
+            graphicsDevice.Adapter.QueryRenderTargetFormat(graphicsDevice.GraphicsProfile, preferredFormat, preferredDepthFormat, preferredMultiSampleCount,
+                out preferredFormat, out preferredDepthFormat, out preferredMultiSampleCount);
```
nkast: QueryRenderTargetFormat() is called after the base class (Texture2D) is initialized with the preferredFormat.

nkast: We can move this code into the private Texture2D(...) and perform the fallback when SurfaceType ==  RenderTarget. 

We also need to remove the SurfaceFormat & SurfaceType parameters from Texture2D.PlatformConstruct()
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.DirectX.cs#L37-L41
When is needed, it has to use the ._format that is already initialized from Texture2D().

SimonDarksideJ:
@KonajuGames are you planning to resolve the outstanding queries/changes or can this be closed?

SimonDarksideJ:
Sadly while very useful and innovative, these approaches are soon to be superseded by the new Native Platform and DX12 update, so closing for now.

Issue #4105, opened 2015-08-27
Win10 mobile: bug with touch 


KakCAT:
The following code draws lots of rectangles to bring W10 to around 1 FPS and prints mouse taps with Red, and prints Pinches with cyan:

http://pastebin.com/B313b9RX

Run the program on Lumia 630 with latest W10mobile version, and tap with several fingers at the same time for a few seconds. Once done that, all subsequent taps are reported as Pinches  (press back on the phone to clear all the taps and see the problem better)

I know this is not a common case (running at 1 FPS with multiple fingers) but may bring evidence of other problems. I was testing #3316 in windows 10 and I randomly found this. Feel free to delete this issue if you feel it's just too picky :)

There's also a huge delay between the tap is done and the tap is notified to the app (even at ~1FPS)

BTW, what is the correct term? Windows 10 mobile? UWP mobile? Windows Phone 10?

tomspilman:
Fixed in #5520.

Issue #3948, opened 2015-06-18
Windows 8.1 Universal on Phone - Graphics Glitch on Startup


jakepoz:
This bug occurs with Windows 8.1 Universal projects and the default Monogame templates running the latest nightly build (6/18/15, as well as 3.4 release builds).

We notice that very often on app startup, there is perhaps half a second where the graphics do not take up the full screen. This happens both on the Windows Phone Emulator and on a Lumia 920 test device. We do not see it when creating Windows Phone 8.0 projects, or when running on the desktop.

You can see it very clearly in this video:
https://www.youtube.com/watch?v=OAWHxEiTdnw

As the app starts up, it first renders to the top half of the screen. After half a second, rendering starts happening like normal.

Please let me know if any other information is needed to get a repro of this issue.

tomspilman:
I've seen this too.  What is interesting is that the stock DirectX C++ sample for Win10 has the same behavior.

This makes me suspect it is a bug in the early builds of Win10 on phone.  I've mostly been waiting for Win10 to get finalized so we can make another pass fixing issues for it.

jakepoz:
Yes, I've seen this on Windows 10, but it does happen on production builds of Windows Phone 8.1 that are live today. Do you think we should look into it on WP 8.1 first, and then come back to Win10 later?

tomspilman:
Oh... totally missed the Win 8.1 there.  I can't say I know what is going on with it on that platform.

jakepoz:
Doing a bit of digging here, I believe this is related to SupportedOrientations and their initial states.

If you comment out the line shown in the sample:
```C#
graphics.SupportedOrientations = DisplayOrientation.LandscapeLeft | DisplayOrientation.LandscapeRight;
```

and you run the template project starting in portrait mode, there is no graphics glitch. However, the graphics are scaled weird, because Monogame thinks your screen is 640x384 in size yet it scales the viewport to that. If you then rotate the device to landscape, the graphics device gets reset to look proper, and then if you rotate back again to portrait, the graphics device gets reset again and the portrait mode works properly too.

I believe what's happening is that on startup, we see a duplicate initialization of the graphics device, once for some strange 640x384 size and then again for whatever the proper size is.

nkast:
This looks a lot like kakCAT video at http://community.monogame.net/t/wp8-1-game-freezes-sporadically/1865/17

[...]

Issue #3589, opened 2015-03-03
WP8.1 orientation bug at the start 


KakCAT:
Hi!

when running a WP8.1 game, the orientation is not changed until several frames have passed since the start. It is a quite weird bug, because it seems it depends on the CPU load. It looks really bad because my loading screen is in portrait (while it should be in landscape) for the whole process.

The problem can be reproduced very easily with the spinning cube example: Modify the draw function, and loop 1000 times the whole draw (except clear and base.Draw, of course).

Running the example with a single draw makes the orientation turn as soon as it is started, but with the 1000 times loop, the device takes quite a time to turn orientation.

KakCAT:
Hi,

I had to do a video for another bug which also shows this bug. Leaving here as a reference

http://youtu.be/qBe8T9Bptq8

tomspilman:
Fixed in #5520.

PR #5352, opened 2016-12-07
[UAP] Alternative RenderLoop

nkast:
Use Idle for Tick(). Let Dispatcher process all events (input/resize/rotate) before rendering.

Fixes mouse, keyboard lag [UAP]  touch, rotation lag & freeze (WP8.1). 
#4897, #5300, #3948, #3589, #3316, #4105(?).

### DIFF
--- a/MonoGame.Framework/WindowsUniversal/UAPGamePlatform.cs
+++ b/MonoGame.Framework/WindowsUniversal/UAPGamePlatform.cs
@@ -29,6 +29,8 @@ namespace Microsoft.Xna.Framework
         internal static readonly TouchQueue TouchQueue = new TouchQueue();
 
         internal static ApplicationExecutionState PreviousExecutionState { get; set; }
+        
+        bool _enableRunLoop = false;
 
         public UAPGamePlatform(Game game)
             : base(game)
@@ -99,14 +101,22 @@ namespace Microsoft.Xna.Framework
             SystemNavigationManager.GetForCurrentView().BackRequested += BackRequested;
 
             CoreApplication.Suspending += this.CoreApplication_Suspending;
+            CoreApplication.Resuming += this.CoreApplication_Resuming;
 
             Game.PreviousExecutionState = PreviousExecutionState;
         }
 
         private void CoreApplication_Suspending(object sender, SuspendingEventArgs e)
         {
+            _enableRunLoop = false;
+
             if (this.Game.GraphicsDevice != null)
                 this.Game.GraphicsDevice.Trim();
+        }
+
+        private void CoreApplication_Resuming(object sender, Object e)
+        {
+            StartRunLoop();
         }
 
         public override GameRunBehavior DefaultRunBehavior
@@ -132,11 +142,42 @@ namespace Microsoft.Xna.Framework
 
         public override void StartRunLoop()
         {
-            CompositionTarget.Rendering += (o, a) =>
+            if (!_enableRunLoop)
             {
-				UAPGameWindow.Instance.Tick();
-                GamePad.Back = false;
-            };
+                _enableRunLoop = true;
+                UAPGameWindow.Instance.CoreWindow.Dispatcher.RunIdleAsync(OnRenderFrame);
+            }
+        }
+
+        private void OnRenderFrame(IdleDispatchedHandlerArgs e)
+        {
+            if (_enableRunLoop)
+                OnRenderFrame(e.IsDispatcherIdle);
+        }
+
+        private void OnRenderFrame()
+        {
+            if (_enableRunLoop)
+            {
+                var dispatcher = UAPGameWindow.Instance.CoreWindow.Dispatcher;
+                if (dispatcher.ShouldYield(CoreDispatcherPriority.Idle))
+                    dispatcher.RunIdleAsync(OnRenderFrame);
+                else
+                    OnRenderFrame(true);
+            }
+        }
+
+        private void OnRenderFrame(bool isQueueEmpty)
+        {
+            UAPGameWindow.Instance.Tick();
+            GamePad.Back = false;
+
+            // Request next frame
+            var dispatcher = UAPGameWindow.Instance.CoreWindow.Dispatcher;
+            if (isQueueEmpty)
+                dispatcher.RunAsync(CoreDispatcherPriority.Low, OnRenderFrame);
+            else
+                dispatcher.RunIdleAsync(OnRenderFrame);
         }
         
         public override void Exit()
--- a/MonoGame.Framework/Windows8/MetroGamePlatform.cs
+++ b/MonoGame.Framework/Windows8/MetroGamePlatform.cs
@@ -12,8 +12,12 @@ using System.Diagnostics;
 using Microsoft.Xna.Framework.Graphics;
 using Microsoft.Xna.Framework.Input.Touch;
 using Windows.ApplicationModel.Activation;
+using Windows.ApplicationModel.Core;
+using Windows.ApplicationModel;
 using Windows.UI.ViewManagement;
 using Windows.UI.Xaml.Media;
+using Windows.UI.Core;
+using Windows.UI.Xaml;
 #if WINDOWS_PHONE81
 using Windows.UI.Xaml;
 #endif
@@ -27,6 +31,8 @@ namespace Microsoft.Xna.Framework
         internal static readonly TouchQueue TouchQueue = new TouchQueue();
 
         internal static ApplicationExecutionState PreviousExecutionState { get; set; }
+                
+        bool _enableRunLoop = false;
 
         public MetroGamePlatform(Game game)
             : base(game)
@@ -99,7 +105,21 @@ namespace Microsoft.Xna.Framework
                 }
             }
 
+
+            Application.Current.Suspending += this.CoreApplication_Suspending;
+            Application.Current.Resuming += this.CoreApplication_Resuming;
+
             Game.PreviousExecutionState = PreviousExecutionState;
+        }
+
+        private void CoreApplication_Suspending(object sender, SuspendingEventArgs e)
+        {
+            _enableRunLoop = false;
+        }
+
+        private void CoreApplication_Resuming(object sender, Object e)
+        {
+            StartRunLoop();
         }
 
         public override GameRunBehavior DefaultRunBehavior
@@ -114,10 +134,41 @@ namespace Microsoft.Xna.Framework
 
         public override void StartRunLoop()
         {
-            CompositionTarget.Rendering += (o, a) =>
+            if (!_enableRunLoop)
             {
-                MetroGameWindow.Instance.Tick();
-            };
+                _enableRunLoop = true;
+                MetroGameWindow.Instance.CoreWindow.Dispatcher.RunIdleAsync(OnRenderFrame);
+            }
+        }
+
+        private void OnRenderFrame(IdleDispatchedHandlerArgs e)
+        {
+            if (_enableRunLoop)
+                OnRenderFrame(e.IsDispatcherIdle);
+        }
+
+        private void OnRenderFrame()
+        {
+            if (_enableRunLoop)
+            {
+                var dispatcher = MetroGameWindow.Instance.CoreWindow.Dispatcher;
+                if(dispatcher.ShouldYield(CoreDispatcherPriority.Idle))
+                    dispatcher.RunIdleAsync(OnRenderFrame);
+                else
+                    OnRenderFrame(true);
+            }
+        }
+
+        private void OnRenderFrame(bool isQueueEmpty)
+        {
+            MetroGameWindow.Instance.Tick();
+
+            // Request next frame
+            var dispatcher = MetroGameWindow.Instance.CoreWindow.Dispatcher;
+            if (isQueueEmpty)
+                dispatcher.RunAsync(CoreDispatcherPriority.Low, OnRenderFrame);
+            else
+                dispatcher.RunIdleAsync(OnRenderFrame);
         }
         
         public override void Exit()
--- a/MonoGame.Framework/Windows8/MetroGameWindow.cs
+++ b/MonoGame.Framework/Windows8/MetroGameWindow.cs
@@ -36,6 +36,8 @@ namespace Microsoft.Xna.Framework
 
         #region Internal Properties
 
+        internal CoreWindow CoreWindow { get { return _coreWindow; } }
+
         internal Game Game { get; set; }
 
         internal bool IsExiting { get; set; }
--- a/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs
+++ b/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs
@@ -33,6 +33,8 @@ namespace Microsoft.Xna.Framework
         private InputEvents _windowEvents;
 
         #region Internal Properties
+        
+        internal CoreWindow CoreWindow { get { return _coreWindow; } }
 
         internal Game Game { get; set; }
 


nkast:
RunIdleAsync creates a new IdleDispatchedHandlerArgs object on every
frame. I changed it to RunAsync(Low), keyboard/mouse events  are still good!

We still need RunIdleAsync to start the loop, otherwise we get a blank
screen. Probably StartRunLoop() needs to be called after
xaml/SwapChainPanel OnCreated event.

MonoGame.Framework/WindowsUniversal/UAPGamePlatform.cs
```diff
@@ -132,13 +132,22 @@ public override void RunLoop()
...
         public override void StartRunLoop()
         {
-            CompositionTarget.Rendering += (o, a) =>
+            CoreWindow.GetForCurrentThread().Dispatcher.RunIdleAsync((a) =>
             {
-				UAPGameWindow.Instance.Tick();
-                GamePad.Back = false;
-            };
+                CoreWindow.GetForCurrentThread().Dispatcher.RunAsync( CoreDispatcherPriority.Low, OnRenderFrame);
+            });
```
tomspilman: This looks sort of weird... why are you dispatching once with `RunIdleAsync` to just dispatch again `RunAsync(Low)`?

Seems easier to just do:
```
public override void StartRunLoop() 
{
   OnRenderFrame();
}
```
This will do one tick to kick things off then register itself for the next.

tomspilman:
Anyone had time to test this on a game that suffered from the input lag?

Also should check if it affects your overall game performance.

Danthekilla:
>Anyone had time to test this on a game that suffered from the input lag?

@tomspilman I have just given it a go. My games keyboard inputs went from 2-20 seconds of lag to 10-100 seconds of lag with this change.

Performance was not effected in my limited testing.

tomspilman:
> My games keyboard inputs went from 2-20 seconds of lag to 10-100 seconds of lag with this change.

Are you saying the keyboard input lag is worse?  100 seconds of lag worse?

Danthekilla:
>Are you saying the keyboard input lag is worse? 100 seconds of lag worse?
Yeah exactly, much much worse.

KonajuGames:
> My games keyboard inputs went from 2-20 seconds of lag to 10-100 seconds of lag with this change.

Almost _two minutes_ of lag?

Danthekilla:
>Almost two minutes of lag?

If I constantly move the mouse around yes. I actually think that I can delay the input forever if I keep moving the mouse.

nkast:
> I have just given it a go. My games keyboard inputs went from 2-20 seconds of lag to 10-100 seconds of lag with this change.

That's strange, maybe I did something wrong when I moved to Idle->Low priority.
I rewrote that, now I schedule a Tick on Idle priority when the queue is not empty, otherwise on Low priority to eliminate the garbage of RunIdleAsync. In theory the app will consume all queued events within a frame or two. I tested it on desktop/phone/emulator, the same for WP8.1/W8.1 project and didn't notice any lag.

Danthekilla:
I will give it another go tomorrow if I get a chance. Perhaps I stuffed something up somehow.

nkast:
@Danthekilla BTW, are you using FixedTimeStep? and how is your frame rate? 
I tried to emulate most senarios with my test.

Danthekilla:
We are not using a fixed timestep, our framerate is from 50 to 60. (capped from vsync)

MonoGame.Framework/WindowsUniversal/UAPGamePlatform.cs
nkast: It's too early to call RenderFrame(). And I had to dispatch with  RunIdleAsync the first time so that XAML initialize & create events arrives before OnRenderFrame();.

BrianPeek:
I'm tired and likely missing something in what I'm reading...

Isn't this change going to only render a new frame if the message queue is empty?  So, if I'm moving the mouse (generating messages) or some such, wouldn't this not ever render a frame?

I'm sleepy and shouldn't be commenting when sleepy.  :)

tomspilman:
> Isn't this change going to only render a new frame if the message queue is empty? 

Yeah that sounds like what is happening.

What we are looking for is something like the old school windows message loops:
```
while(true)
{
    // Process all pending messages until none remain.
    while(PeekMessage(&msg, NULL, 0, 0, PM_REMOVE))
    {
        TranslateMessage(&msg);
        DispatchMessage(&msg);
    }

    // No pending messages so update/draw.
    game->Tick();
}
```
With these loops input events wouldn't get missed and the game would draw as fast as possible in old Win32/DirectX games.  

The key difference here being the tick didn't depend on its own event... which can be infinitely delayed by the message system.  My guess is that new events are continuously queued even during message processing and all end up ahead of our idle/low event...   so we never get to run.

nkast:
@BrianPeek 
> Isn't this change going to only render a new frame if the message queue is empty?
> So, if I'm moving the mouse (generating messages) or some such, wouldn't this not ever render a frame?

If messages arrive faster than we can process them, then yes. In practice I didn't found any problems with mouse input neither with touch input on a phone. For store apps we queue up mouse/touch events and process them later in the game loop before Update(), so it doesn't keep the thread for too long. Certainly it wouldn't hurt to remove any unnecessary overhead (#5158).

What you said is exactly what is happening right now to input events. `CompositionTarget.Rendering` event  fires 60 times per second with high priority. On every tick we hold the UIThread for 16.6ms, if Draw is slow and miss Vsync it's +16.6ms. By the time Tick returns, a new `Rendering` event is already dispatched in the queue, It looks that only few events can make it through this narrow window. Mouse events start to pile up with ever-growing lag and push back any keyboard events (lower priority than mouse).

For example, when I tested this PR on WP8.1/WP10 with Low priority for Tick, the rotation event never come through! Normally I have to use Idle every time but I suppose a one-frame lag rotation is acceptable  to avoid garbage.

nkast:
@tomspilman 
> What we are looking for is something like the old school windows message loops:

In store apps there is the `Dispatcher.ProcessEvents(ProcessAllIfPresent)`, but that's for CoreApplication.Run loops, you can't call it from within an event. In XAML, the UIThead runs YOU :simple_smile: 

The PR tries to emulate exactly that. Another way to see it is like the Invalidate/WM_PAINT event in MFC/WinForms. 

> The key difference here being the tick didn't depend on its own event... which can be infinitely delayed by the message system.

I don't think there's a difference between the two. You can stack in the while(PeekMessage()) as well if you have a slow Mouse handler.

nkast:
In a quick search for mouse pooling rate, someone mention the value 125Hz (8ms) but it's really up to the driver and you could set it up to 1000Hz,  although my logitech SetPoint doesn't have any setting for that.

In @Danthekilla's comparison of the frequency mouse events, it is clear that there is a separation of 4ms between mouse events. So I'd say that a rate of 250Hz is typical nowadays. 

In the worst case scenario of 1000Hz you have at least 1ms/event to keep the queue empty . Typically there would be 16 events in the queue between ticks.

tomspilman:
@nkast @BrianPeek - So where does this leave us here?

I like the idea of preserving the existing threading model where the game can be expected to be running from the UI thread.  It is what we've been doing since Windows 8 (because that is how the DirectX samples did it) and has worked.  If we can make this work under UWP that is preferred.

However if we have to change tactics here because there is no other good alternative then lets figure out how to do it with the least amount of pain to the developer.

Danthekilla:
As a semi related thing I should mention that I have gotten the non xaml version of UWP monogame working quite well now, all input works correctly and it now correctly handles screen resizing and device resets.

However as of today we have decided that it won't work for us as the loss of xaml (mainly for advertising) is too big a loss. I will be giving this pull request another go tomorrow to see if it fixes up the delayed keyboard input. Because we need to move back to a xaml system sadly.

tomspilman:
@Danthekilla - So how often do you interact with the XAML side of the app from the MonoGame code?  How painful would it be if you needed to use some thread synchronization and/or dispatch to communicate between these sides?

Danthekilla:
@tomspilman Not often. We mainly use it for ad overlays which are just told to show or hide for the most part. And we also use it for our in app currency coin system that we are planning to try out. We were just going to rewrite that UI in monogame but since we haven't found any decent ad providers that don't require XAML we will still with XAML for that UI too.

Doing some simple thread synchronization and dispatching would be absolutely fine.

nkast:
I 'am confident about that one, just waiting @Danthekilla's confirmation.

One issue would be that low priority events like Keyboard will lag by one frame because of the workaround to eliminate garbage. (I wonder if `IdleDispatchedHandlerArgs` generates garbage or being a WinRT uses ref counting and doesn't trigger the GC). I can solve this by calling  `.RunIdleAsync(OnRenderFrame)` immediately when `.ShouldYield()` is true, instead of run a Tick() and schedule an idle event for next time.

BrianPeek:
@tomspilman I'm not sure where it leaves us.  My personal opinion is that spinning a thread and running the game in the thread is the proper way to do this.  While @nkast's solution may work and be 100% valid, I've never seen a XAML/D3D game take this approach and I can't say either way whether it's good or bad.  Every implementation I've seen either runs on the UI thread, or spins a thread and runs there.

nkast:
OK, I added a `CoreWindow` property in UAPGamePlatform to avoid calls to `GetForCurrentThread()` on every frame. 

Also changed the logic a bit, when ShouldYield() is true it reschedule the RenderFrame event on Idle priority and exit. All low priority events, including keyboard, will get processed before Tick(). Previously we  could miss key events on every second frame.

@Danthekilla This is good to go. Expect to see no lag.

Danthekilla:
Nope... Still lags :(

I have just spent the day testing this and making sure that I haven't broken anything and I am still able to delay keyboard input indefinitely. 

It is fine for the first 20-30 seconds of moderate input but then keyboard events start to lag. Mouse always feels smooth. If I move the mouse in little circles after holding forward (Key.W) for 5 seconds then release forward (Key.W) I don't get a Key released event until about 5 seconds after I stop moving the mouse. During this time no new key presses come though either.

I will record a video if you like.

nkast:
@Danthekilla 
> Nope... Still lags :(

I really don't know how to explain that! The only case would be if the OS doesn't send the events on time.

> I will record a video if you like.

The only thing that would help would be a test sample to reproduce this behavior. Perhaps there is a combination of issues here.

Danthekilla:
>The only thing that would help would be a test sample to reproduce this behavior.

I figured as much, just thought I would offer, sometimes bugs are easier to understand visually rather than by my description.

I will try and get a sample going that has the issue.

nkast:
@Danthekilla 

Here is the test I use. 
https://gist.github.com/nkast/ffe21047c77375d1f4a86a09a4e0d69c

Under develop I can reproduce the keyboard/Mouse lag and the touch/rotation lag on a phone.
With this PR all are fixed. My Windows version is Build 14393.576.

If you could test this it would be easier to locate the problem. I don't use the `delayDrawMs` anymore because Task.Delay() doesn't behave in a predictable way. Instead, change the nested loop in Draw() to simulate load.

Also I draw the frame number to get a feel of how often the screen is updated. Low frame rate can be deceiving and perceived as mouse lag...

Danthekilla:
Ok so I have spent most of today making a test case that has causes the issue.

I ran this code against the public 3.5 release and also against your tncAlternativeRenderLoop(ARL) branch.

Here are some things I observed (some irrelevant, I am just noting them to be thorough)

Setting SynchronizeWithVerticalRetrace = false; on 3.5 has no effect and the game still runs with vsync (understandable as support for this was only added in the anniversary update of windows 10). However on the AlternativeRenderLoop(ARL) branch I get a crash trying to create the swapchain.

My test game runs at 90fps (on my 144hz screen) on my i7 / GeForce 960 machine when using the 3.5 release and drops about 5 fps when moving the mouse. When running the ARL branch I get ~30 fps when not moving the mouse, and moving the mouse makes the framerate become very erratic averaging around 20fps and dropping as low as 5fps.

As for the input bug it still happens with the ARL branch but much less often. It only happens when the framerate is not a perfect 60. And it gets worse the longer the program is open. Eventually getting 10-20 seconds of delay on the keyboard. It feels very different to how the input lag did before.

Also when I close the app via the windows X button it never closes with the ARL branch and does with the 3.5 branch. (this is probably because I have background tasks still going I think?)

Here is the Input Lag Test project I made. 

https://1drv.ms/u/s!Ak605HL3TOgxqfhR7ZUWAVXNAzDMrA

Let me know how you go with it.

Danthekilla:
Oh and the best way to see the issue is to press 'A' and 'D' alternating every few seconds and move the mouse in circles.

nkast:
Thanks @Danthekilla. No time to see it right now, I will check tomorrow. or maybe tonight.

jakepoz:
I just ran @Danthekilla 's test project. There is indeed a large amount of input lag using just the public 3.5 Monogame. With the included Alternative Render Loop Monogame I was not able to see any render lag at all however.

I definitely need to both be typing rapidly and moving the mouse rapidly to reproduce the issue on 3.5. However, we have had a number of user reports about unresponsive controls on UWP-based games and we were never able to reproduce those (our games are mouse-only).

Danthekilla:
Our games are effectively an FPS, so we have loads of constant mouse movement and keyboard input.

Strange that you didn't get the input lag issues we have been getting. What was your framerate in 3.5 and ARL? Do you also get a ~65% framerate drop when using ARL compared to 3.5?

If your framerate is too high, tweak the RenderCount variable (line 105), I found the input bug the easiest to trigger at around 30fps.

jakepoz:
On 3.5 I get a constant 60 fps even while the mouse is lagging like crazy.
On ARL, the framerate hover around 40-50 but is variable. If I do lots of mouse and keyboard it will drop to 20 perhaps. The mouse pointer will appear choppy, but it doesn't lag behind and then catch-up like it does with 3.5.

nkast:
@Danthekilla 
I tested your samples, so here what I found.

>Setting SynchronizeWithVerticalRetrace = false; on 3.5 has no effect and the game still runs with vsync (understandable as support for this was only added in the anniversary update of windows 10). However on the AlternativeRenderLoop(ARL) branch I get a crash trying to create the swapchain.

vsync is not implemented on MG/UAP, not related to ARL.

>My test game runs at 90fps (on my 144hz screen) on my i7 / GeForce 960 machine when using the 3.5 release and drops about 5 fps when moving the mouse. When running the ARL branch I get ~30 fps when not moving the mouse, and moving the mouse makes the framerate become very erratic averaging around 20fps and dropping as low as 5fps.

I set the load to get around 45-60 fps. With ARL + mouse input It drops to 30fps but also get readings of 62fps, which is impossible! The fps counter fluctuates a lot in an erratic way.  When you reset the stopwatch the next sample gets a headstart by  (stopwatch.Elapsed - FpsSampleSpan). The way it's written makes it sensitive to timeintervals between Update(). CompositionTarget.Rendering calls Tick on very regular intervals, while ARL process all events between last Present()/VSync and next Tick(). I then used an FPS counter that samples every second and reports actual frames/second. I got stable 54-57fps on both ARL and CompositionTarget.Rendering, with and without input. So I didn't see the new method affecting actual frame rate in any way. (Obviously it spent some more time of processing events which  CompositionTarget.Rendering pospone for later).

> As for the input bug it still happens with the ARL branch but much less often. It only happens when the framerate is not a perfect 60. And it gets worse the longer the program is open. Eventually getting 10-20 seconds of delay on the keyboard. It feels very different to how the input lag did before.

I didn't see any lag with ARL and I tried it many times for several minutes. Obviously there something else causing this. Can you test it yourself on another system?

> Also when I close the app via the windows X button it never closes with the ARL branch and does with the 3.5 branch. (this is probably because I have background tasks still going I think?)

I can confirm this. For as long as the message queue is full we don't get the `.Closing` event. I'll see if we can workaround this. If things get too complicated I should move this to another class.

nkast:
> On ARL, the framerate hover around 40-50 but is variable. If I do lots of mouse and keyboard it will drop to 20 perhaps. The mouse pointer will appear choppy, but it doesn't lag behind and then catch-up like it does with 3.5.

Thanks for testing this @jakepoz .
If you worry about frame rate I suggest to re-test with another FPS counter, the current one doesn't seem to report actual FPS. If you still see a small drop in fps that's because ARL process all events at once instead of ignoring them ("The fastest code is the code that never runs").

@tomspilman 
We can improve on mouse event speed with #5158.

tomspilman:
@nkast - So looking at the latest VS2015 template project for XAML and DirectX it does this:
```
	// Register our SwapChainPanel to get independent input pointer events
	auto workItemHandler = ref new WorkItemHandler([this] (IAsyncAction ^)
	{
		// The CoreIndependentInputSource will raise pointer events for the specified device types on whichever thread it's created on.
		m_coreInput = swapChainPanel->CreateCoreIndependentInputSource(
			Windows::UI::Core::CoreInputDeviceTypes::Mouse |
			Windows::UI::Core::CoreInputDeviceTypes::Touch |
			Windows::UI::Core::CoreInputDeviceTypes::Pen
			);

		// Register for pointer events, which will be raised on the background thread.
		m_coreInput->PointerPressed += ref new TypedEventHandler<Object^, PointerEventArgs^>(this, &DirectXPage::OnPointerPressed);
		m_coreInput->PointerMoved += ref new TypedEventHandler<Object^, PointerEventArgs^>(this, &DirectXPage::OnPointerMoved);
		m_coreInput->PointerReleased += ref new TypedEventHandler<Object^, PointerEventArgs^>(this, &DirectXPage::OnPointerReleased);

		// Begin processing input messages as they're delivered.
		m_coreInput->Dispatcher->ProcessEvents(CoreProcessEventsOption::ProcessUntilQuit);
	});

	// Run task on a dedicated high priority background thread.
	m_inputLoopWorker = ThreadPool::RunAsync(workItemHandler, WorkItemPriority::High, WorkItemOptions::TimeSliced);
```
You should review the current DX11 XAML templates in VS2015 and compare it to our XAML setup.

Danthekilla:
@nkast 
Thanks for having a go of the sample. 

The FPS counter was copied from a old XNA sample I didn't notice that that had any errors. When I would set the draw count low I would get 144FPS reported consistently on my 144hz screen. You can set the sample span to one second, I just had it set to 100ms for testing.

>So I didn't see the new method affecting actual frame rate in any way. 

That's odd, my FPS went from 90 to 30 (or as low as 5fps under mouse load) when moving to ARL I will have to check my FPS counter again because I don't think it could possibly be out by that much.

>Can you test it yourself on another system?

We have tested on 2 systems, (however they do have the same hardware) I will test on a few other devices when I get back to work.

>For as long as the message queue is full we don't get the .Closing event. 

So this means the message queue is full indefinitely? My closing events never get triggered.

nkast:
@Danthekilla 

Setting ` FpsSampleSpan` to 1 sec would give better results but it also drifts with every reset.  
I removed the reset and sample the fps whenever the stopwatch increased by one second.

>>For as long as the message queue is full we don't get the .Closing event.
>So this means the message queue is full indefinitely? My closing events never get triggered.

While the message queue is what keeps the app alive I was wrong about the closing event. The closing event never fires on store apps. I need a way other than `UAPGameWindow.Instance.IsExiting` to terminate the loop. I will setup a check around Suspend/Resume events.

BrianPeek:
Hey @Danthekilla, I did a quick and dirty test of the XAML project using a simple multithreaded renderer + CoreIndependentInputSource  vs. everything running on the main thread.  As with my other test, this is not not complete/tested, and shouldn't be used for anything real, but I'd love it if you had time to test it out with your project and report back results.  You can find the project/branch [here](https://github.com/BrianPeek/MonoGame/tree/XamlThreads).

Thanks!

jakepoz:
Just had a chance to test these builds, both nkasts and BrianPeek's solutions. I used Danthekilla's test project, but I removed the artificial CPU load (comment out the ThreadPool.RunAsync(BackgroundTask) call)  

With both solutions the input lag is gone for me, and they also close down cleanly for me when exiting. If I switch to Monogame 3.5, there is terrible input lag, even without the artificial CPU load.

The artificial load pegs my CPU to 100% on all cores, and basically every application acts weird in this case. Even then, I didn't see input lag where events were delayed by seconds, just a general sluggishness of response.

BTW, we did have a $1,000 USD bounty open with this bug: https://www.bountysource.com/issues/7309032-wp8-1-multitouch-slowdown and we are of course happy to pay it out once this fix gets accepted.

Danthekilla:
@BrianPeek THIS WORKS!

Zero input lag, moderately stable framerate, full XAML support!

I have had to make some fairly large changes to monogame to dispatch in quite a few places where it was accessing the corewindow and fix up some issues where it was accessing the currentthreads window but I have a build that doesn't have the input bug and has working XAML now.

 
```
WorkItemHandler workItemHandler = action => 
{ 
	while(true) 
	{ 
		UAPGameWindow.Instance.Tick(); 
		GamePad.Back = false;	
	} 
}; 
ThreadPool.RunAsync(workItemHandler, WorkItemPriority.High, WorkItemOptions.TimeSliced); 
```


This single change in StartRunLoop() fixes the input issue 100% of the time. You just have to remember to dispatch everything when you need to talk to the UI thread.

nkast:
@BrianPeek, @Danthekilla 
 Since I already tested separate render thread [some time ago](https://github.com/nkast/MonoGame/commit/83030941ea45b3eab0a32d9d815a057d6ea86705), here are a few thing you should look at: 

* Resize/Rotate events sync. Try to resize the window for a few seconds and a sharpDX exception will be thrown. Probably you must check all events and how they interact with the GraphicsDevice.

* MG and/or the user must dispatch calls to XAML & all .GetForCurrentThread()/.GetForCurrentView() methods.

* coreIndependentInputSource bug with touch devices. I assume that every decent game disables the touch feedback. When using coreIndependentInputSource the touch feedback is always on, causing frame drops on weak tablets with low fillrate because the OS does one extra fullscreen composition. You can check this on a tablet or the emulator. We can't avoid coreIndependentInputSource, otherwise touch events from the main thread get serious lag (seen on WP8.1).  CoreIndependentInputSource is the real blocking issue here. Unless MS fixes it across the board (W8.1/WP8.1 -> W10) I wouldn't use it in my games.

BTW, @BrianPeek , @tomspilman Can we get a pure CoreApplication project template along with XAML? 
@Danthekilla were there any issues yo solved with CoreApplication  and resize?

Also @Danthekilla, did you check this PR on other machines? Any clue what is causing the keyboard lag?

BrianPeek:
I definitely didn't explore resizing/rotate/etc. or any other events with the threaded thing I threw together.  I just setup the threads to see if it would help before going down that path.  We'll need to do that if this is the solution that makes its way forward.

I wrote a CoreApplication template which others tested, but it also needs to be vetted more closely.  That's in a branch [here](https://github.com/BrianPeek/MonoGame/tree/CoreApplication).  Again, needs to be finished, but the basics are there.

KonajuGames:
I believe the threaded approach is the correct way to progress here.  Yes, anything that interacts with XAML has to be invoked, and the resize issues would need to be fixed.

@BrianPeek Is the CoreIndependentInputSource touch feedback visualization bug referenced above by @nkast known by the Windows team? If you know who to prod gently...

jakepoz:
I was doing some testing today regarding the Alternate Render Loop within real-life games that we have developed and published. First, I tested the latest changes on this PR, and it did fully resolve the input lag bug. (Our games are not CPU intensive and Dan's repro project involves maxing out all CPUs in order to see input lag issues)

One thing I noticed was related to the XAML based Ad SDK we use from Microsoft. When I built our game with a develop branch from a few days ago, the input lag is there, but if there is a heavy-duty "rich media" ad playing with animations, it tends to play at a low framerate. The game itself occasionally stutters when an ad like this loads, but after that the game plays mostly fine. On the latest tncAlternativeRenderLoop branch, the input lag is fixed, but these types of ads render at 60fps, while the game itself stutters a lot.

The CoreApplication template is not good for developers who want to monetize with any standard ad SDK which all require XAML.

Does this point to a threaded approach like @KonajuGames is suggesting?

theZMan:
That would make sense... the alternate loop puts all of the game on a separate thread leaving the main windows to handle just input..

Your Ad SDK sees that the main windows is doing almost nothing and promptly sucks up all the CPU probably now starving the bg thread. We might want to make our background thread a higher priority - that might help it beat out the ad.

Is this on win 10 or phone?

Any way to run the ad SDK on a separate thread?

jakepoz:
This is on a Win 10 UWP project for desktop. I don't think there is a way to run certain XAML controls on a separate thread.

BrianPeek:
The AlternateRenderLoop from @nkast doesn't do threads, so not sure @theZMan your thought is correct?  My potential fix is the one with threads... @jakepoz did you try that?  https://github.com/BrianPeek/MonoGame/tree/XamlThreads

theZMan:
@BrianPeek - yes I was confused - too many branches. Fingers crossed!

jakepoz:
Just tested @BrianPeek 's XamlThreads solution, and both the input lag and XAML Ad SDK are working properly. I did have to adjust our project to dispatch the XAML stuff separately, but it was a small change.

Just as @nkast said though, it doesn't work with resizing the window. During the actual resize it will lag a lot and not actually process the resize events fast enough, and will also occasionally crash.

```
SharpDX.SharpDXException: HRESULT: [0x887A0005], Module: [SharpDX.DXGI], ApiCode: [DXGI_ERROR_DEVICE_REMOVED/DeviceRemoved], Message: The GPU device instance has been suspended. Use GetDeviceRemovedReason to determine the appropriate action.

   at SharpDX.DXGI.SwapChain.ResizeBuffers(Int32 bufferCount, Int32 width, Int32 height, Format newFormat, SwapChainFlags swapChainFlags)
   at Microsoft.Xna.Framework.Graphics.GraphicsDevice.CreateSizeDependentResources()
```

BrianPeek:
@jakepoz, yes, my branch is unfinished and was just a PoC for threads.  I need to go back and make sure we process events on the correct threads...

nkast:
@jakepoz 
> On the latest tncAlternativeRenderLoop branch, the input lag is fixed, but these types of ads render at 60fps, while the game itself stutters a lot.

That's bad. Game loop runs at idle priority and ads take over the CPU. I guess it nullifies the advantage  ARL has with XAML interoperability (no dispatching). 

> Does this point to a threaded approach like @KonajuGames is suggesting?

Yes, I am afraid that the only option to bring performance back when using XAML controls/animations.
For resizing for example we can use a `_targetSize` variable and resize the window before a frame starts if we need to. `Exit` event uses something similar (`_shouldExit`) and might work out of the box with a separate thread. We have to avoid locking because of performance and potential deadlocks.

nkast:
Closed via #5520

Issue #2167, opened 2013-12-16
2MGFX does not work on non-Windows platforms


hach-que:
I'm assuming that 2MGFX is not meant to compile under Linux.  Are there any intentions to change this in the future so that compilation can be done on non-Windows platforms?

paewie:
Even if this is a quite old post I encountered this issue, too. The .fx files won't compile on a linux system, even if the pipeline itself claims to be cross-platform (and it is; for most content).
The question now: Were one year and three months enough to implement this compilation or did I get it right that it is not supported by now?

tomspilman:
It isn't implemented yet.

We have one of the first steps worked out (https://github.com/mono/MonoGame/issues/1972), but no one has had time/interest to do the work.

There is a push to get MGFX processing working on Mac which might be how this eventually gets worked on.

paewie:
Okay, thanks for the info. Guess I'm switching the OS for shader compilation then :D

tgjones:
#1972 is implemented now.

I'm not sure if we need to keep #3566 and #3270 open - they seem like dupes of this issue?

On a related note, it looks like the Paradox HLSL -> GLSL converter might be released separately under a more friendly OSS licence very soon: https://twitter.com/xoofx/status/586087334388445185

From what I've seen of it so far, it looks pretty robust. Unlike every other HLSL -> GLSL source translator that I've seen, it supports HLSL SM 4.0+ syntax. Obviously we'll have to see if it works well for MonoGame, but if it does, it could replace MojoShader, and provide a path towards shader compilation on non-Windows platforms.

There are two separate issues here, and perhaps they should be discussed separately:

1. Extend 2MGFX to allow shaders to be written in GLSL.
2. Replace MojoShader with something that does HLSL -> GLSL source translation, so that it can be done on non-Windows platforms.

tomspilman:
> 1.Extend 2MGFX to allow shaders to be written in GLSL

Right... I think this is the next step to tackle.

We would need to test convert some complex shader to GLSL, but leave all the sampler states, render states, techniques, passes in the file.  Then we need to make fixes to the MGFX parser to let it parse the info it needs (sampler state, render state, techniques, passes) while passing thru the GLSL code.

Then we need to figure out how to strip out the GLSL code from the FX stuff.

First I suspect the GLSL compiler will not like things like techniques in the file...  where as the HLSL compiler just ignores it.  

[...]

PR #5354, opened 2016-12-08
Add GLFX format support to 2MGFX

Jjagg:
This needs GLSLOptimizerSharp to be added to the dependencies repo before it can build.
I haven't added the new stuff for Mac yet in the Protobuild definitions. I did add Linux and tested both the Windows and Linux build on Windows. I have only tested a single shader at this point, the translated version of SpriteEffect:

```
uniform mat4 MatrixTransform;
uniform sampler2D Texture;


attribute vec4 position: POSITION0;
attribute vec4 color    : COLOR0;
attribute vec2 texCoord : TEXCOORD0;

varying vec4 vColor;
varying vec2 vTexCoord;

void SpriteVertexShader()
{
    gl_Position = MatrixTransform * position;
    vColor = color;
    vTexCoord = texCoord;
}

void SpritePixelShader()
{
    gl_FragColor = texture2D(Texture, vTexCoord) * vColor;
}

technique SpriteBatch 
{
    pass 
    {
#if GLES
        VertexShader = compile glsl_100 SpriteVertexShader();
        PixelShader  = compile glsl_100 SpritePixelShader();
#else
        VertexShader = compile glsl_110 SpriteVertexShader();
        PixelShader  = compile glsl_110 SpritePixelShader();
#endif
    }
};
```
The build system can only handle old GLSL (attribute, varying), some more effort will be needed to support in/out/inout.

Discussion in #2167.
Sorry for all the whitespace changes :/

Ping @tomspilman


Tools/2MGFX/SourceShaderLanguage.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Shading languages that 2MGFX supports.
+    /// </summary>
+    public enum SourceShadingLanguage
+    {
+        Unknown,
+        Hlsl,
+        Glsl
+    }
```
tomspilman: It seems to me trying to detect shader languages is not necessary and very error prone.

Instead we should depend on the shader profile passed on the 2MGFX command line (in pipeline we set the right profile based on the target platform) and let the errors alert users to a mistake.

As for the future HLSL to GLSL and other translations that will be done via some additional command line option that doesn't exist yet.

tomspilman:
> This needs GLSLOptimizerSharp to be added to the dependencies repo before it can build.

Are you going to submit a PR for that?

Jjagg:
> Are you going to submit a PR for that?

MonoGame/MonoGame.Dependencies#95

Build/Projects/2MGFXReferences.definition
```diff
@@ -13,5 +13,10 @@
       Path="ThirdParty/Dependencies/CppNet/CppNet.dll" />
     <NativeBinary Path="ThirdParty\Dependencies\MojoShader\Windows\libmojoshader_64.dll" />
   </Platform>
+  <Platform Type="Linux">
+    <Binary
+      Name="CppNet"
+      Path="ThirdParty/Dependencies/CppNet/CppNet.dll" />
+  </Platform>
```
KonajuGames: You would need to add the references to GLSLOptimizerSharp and the native binaries as well.

Jjagg: I have that done locally. Just waiting for the dependencies so I can get it resolved in one go.

KonajuGames:
In this PR, you added Linux to 2MGFX but not Mac.  In https://github.com/MonoGame/MonoGame.Dependencies/pull/95 you added glsl_optimizer.dylib for Mac, but no .so for Linux.

Jjagg:
I don't know how to do that. There were instructions to build the optimizer for Linux but it produces a static library (.a file). Can we make do with that? I know very little about how Linux handles these things.

Jjagg:
I think I got it. I'll test on Linux and then add the right file to MonoGame/MonoGame.Dependencies#95

KonajuGames:
A .a file is a Windows .lib.  It's for static linking.  A .so (shared object) file is equivalent to a Windows .dll.

Jjagg:
Mono is throwing exceptions that it can't find the entry points for the functions in the .so lib. [nm](https://en.m.wikipedia.org/wiki/Nm_(Unix)) with the -D flag does find the entry points, so I'm not sure where the issue is.

Jjagg:
I had to wrap the glsloptimizer code in a `extern "C"` block as mentioned on the [Mono PInvoke page](http://www.mono-project.com/docs/advanced/pinvoke/): `Note that a C ABI is assumed. This makes it nearly impossible to directly invoke functions that are not C ABI compatible, such as C++ library functions that are not extern "C".`

Jjagg:
I didn't test the Mac version, but since it was already in the GLSLOptimizerSharp repo I guess it was tested before. We'll need someone to give this a test on a Mac before merging though. Windows and Linux both build the shader. Of course I didn't test the actual shader that gets produced since there's no runtime support yet.

Tools/2MGFX/MGFX.tpg
```diff
@@ -161,7 +166,7 @@ Start -> (Code | Technique_Declaration | Sampler_Declaration)* EndOfFile
    return shader; 
 };
 
-Technique_Declaration -> Technique Identifier? OpenBracket Pass_Declaration+ CloseBracket 
+Technique_Declaration -> Technique Identifier? OpenBracket Pass_Declaration+ CloseBracket Semicolon
```
Jjagg: Good thing there's a test that caught this, this would've broken all effects...

tomspilman:
Thanks @Jjagg ....  i'll review this closer against a project I have here then merge things.

KonajuGames:
I'll look into testing this on Mac tonight.

KonajuGames:
Didn't get to test this on Mac last night.  I'd like to get that done before merging.

tomspilman:
@KonajuGames - Also wait for me to review this next week before we merge.  Want to be sure it fits our needs moving forward.

KonajuGames:
@Jjagg I've added a new version of libglsl_optimizer.dylib.  Made it a Universal (x86/x64) dylib and made the APIs `extern "C"`.  The mgfxo file of the SpriteEffect above is now generated by 2MGFX.  I tried to run the Tests on Mac, but have another issue there.

Jjagg:
>  I tried to run the Tests on Mac, but have another issue there.

The runtime for this is not implemented yet and it differs a bit from the way things are done with GLSL effects that MojoShader built so that won't work.

Jjagg:
I'm going to implement the runtime too. Here's why:

1. It's the only way to properly test this. I bet there's still issues and the easiest way to find those is writing GLFX shaders and testing them.
2. I'm not sure all properties are being set, this can help find out if any necessary data is missing from the effect file.
3. We need this anyway no matter what happens to the build process. Whether it's without optimization, through [XShaderCompiler](https://github.com/LukasBanana/XShaderCompiler), [glslang](https://github.com/KhronosGroup/glslang)(GLSL to SPIR-V)+[SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools)(Optimize SPIR-V)+[SPIRV-Cross](https://github.com/KhronosGroup/SPIRV-Cross)(SPIR-V to GLSL) or possibly [LunarGLASS](https://github.com/LunarG/LunarGLASS) which I only just found out about, the runtime will stay the same. It's just different for MojoShader because of a special optimization it does. So probably not a lot of changes need to be made anyway :)
4. It will finally actually allow people to fully build their game with effects on Linux or Mac (when the above issues are solved).

KonajuGames:
The runtime portion will definitely be needed before this can go much further.

Jjagg:
I can't wait until we finally drop OpenTK completely, this is annoying as hell.

Jjagg:
screenshot from 2017-01-25 03-06-19

Working! Semantics were not being set yet, doh! Next steps are porting the built-in effects and pushing a bunch of shaders through to get this properly tested.

KonajuGames:
Nice. Well done.

tomspilman:
@Jjagg - When you get a chance please rebase/merge the latest develop changes into this PR.  It has conflicts right now.

Jjagg:
Done!

Zkylar:
@JJagg, I was testing out some complied fonts on Mac and picked up issues with DrawString where the rendering of the text looked funny. Turns out it seems related to "SpriteEffect.glfx.mgfxo" in the resource is compiled with :

gl_Position = MatrixTransform * position;

When I recompiled it with 2MGFX.exe with position and MatrixTransform switched seem like DrawString starts working again:

gl_Position = position * MatrixTransform; // This fixed DrawString issue

tomspilman:
> with position and MatrixTransform switched 

That suggests the GLSL path is using the opposite column/row ordering to HLSL... which I believe is the default.  I *think* there is a `#pragma` or something you can use in your GLSL to tell it what row order to use by default.  This is usually better because it means your math can be identical to HLSL shaders.

Jjagg:
When #5410 gets merged we can probably get rid of some of the #ifs for mobile.

angrywasp:
I have been porting my shaders and trying to compile and run them in Linux and found a few things i thought i should share. Before i begin i should mention i have only been at it for about 12 hours, so i may be doing some things wrong. 

Firstly, i can confirm that the matrix multiplication, as previously discussed has to be reversed (at least on linux). I have also found that on some shaders the gl_Position.Y Axis needs to be inverted. Might be my code, since it is not necessary for all shaders.

Having a couple of problems compiling a skinned mesh shader. My original code declare a bones array as float4x3, however a uniform mat4x3 (although it is most likely going to need to be mat3x4) doesn't compile. I get the following error:

_Error parsing shader: (9,16): error: syntax error, unexpected NEW_IDENTIFIER, expecting '{'_

An array of mat4 compiles fine. however when i attempt to access the array i get the following error:

_The method or operation is not implemented._

Lastly, there seems to be some problems with multi pass shaders. I have a shader with 2 passes. The shader compiles without error, but when attempting to load it there is a problem with the mapping of parameters to constant buffers. If i split the 2 passes into a shader of their own each compiles and loads without problem. 

If anyone is interested i can provide the shaders in question. I will look into these problems further myself, but if anyone can offer any feedback on these issues, i would appreciate it.

Thanks

angrywasp:
Bit more progress made. I noticed that in _Graphics/Effect/EffectParameter.cs_ when assigning a bool or int value to a shader parameter it is being cast to float. The comments blame this on MojoShader encoding those types to float, but this GLFX implementation appears to want bool and int casts respectively. I have pushed https://github.com/angrywasp/MonoGame/commit/f60a5d87be21722c31aec55580f6152b5d3bfa64 to my fork to rectify the problem

However i am no git guru by any stretch and so i am not sure how to create a PR for this commit.

Jjagg:
> Firstly, i can confirm that the matrix multiplication, as previously discussed has to be reversed

Thanks for confirming :) 

> My original code declare a bones array as float4x3, however a uniform mat4x3 (although it is most likely going to need to be mat3x4) doesn't compile.

No #version directive is added yet, so the GLSL version defaults to 1.10. Non-square matrices were added in 1.20, so they won't be available yet. You're right to use a mat4 instead.

> An array of mat4 compiles fine. however when i attempt to access the array i get the following error:

Is that exception thrown [here](https://github.com/angrywasp/MonoGame/blob/f60a5d87be21722c31aec55580f6152b5d3bfa64/MonoGame.Framework/Graphics/Effect/EffectParameter.cs#L297)? That's the only NotImplementedException that's thrown in the EffectParameter file. Maybe the ParameterClass is messed up.

> Lastly, there seems to be some problems with multi pass shaders. I have a shader with 2 passes. The shader compiles without error, but when attempting to load it there is a problem with the mapping of parameters to constant buffers. If i split the 2 passes into a shader of their own each compiles and loads without problem.

Not sure what's going wrong here without doing some digging. Each ShaderProgram (combination of vertex and pixel shader) has its own UniformBuffers, so these should be resolved correctly. Is it that the mapping is wrong, or that there are missing mappings? Different ShaderPrograms can have different mappings for a uniform, so maybe the wrong mapping is used (not the one from the active program).

angrywasp:
OK, so after a bit of debugging i think i found the problem. The exception was being thrown [here](https://github.com/MonoGame/MonoGame/blob/develop/Tools/2MGFX/EffectObject.cs#L610)

It appeared that the Bones array is expanded, but the parameter class and type is not set on the array items. This resulted in D3DXPARAMETER_TYPE.VOID being passed to this method.

[This commit](https://github.com/angrywasp/MonoGame/commit/f02c683a77fd0065af6ef1a2db6e895165fd4c85) sets the class and type properties for the array items and the shader now compiles without error.

I have not tested the resulting effect at this point. That's a job for tomorrow.

Now, onto the issue of the constant buffers. I have attached a screenshot of the debugger output to better describe  what is happening. The ConstantBuffer array is the list on the left and the parameters are on the right
debug_output

The constant buffers at index 5-7 are repeated, i assume because they are used in both passes of the shader. the CB at index 5 is meant to be a Boolean type, but is instead now mapped to a Texture2D.

I have also attached the (most likely buggy) shader code [here](https://github.com/MonoGame/MonoGame/files/1049672/Material.txt). I will have a bit more of a look at the situation, but hopefully this gives you a bit better insight into what is going on

Cheers

Jjagg:
Nice catch! Thanks for helping out :)
I don't have time to look into this atm, but I'm really glad this is moving forward a bit!

> The constant buffers at index 5-7 are repeated, i assume because they are used in both passes of the shader.

Yes, I think that's how it should be. The uniform locations are not necessarily the same for each pass, so we need two copies of them.

> the CB at index 5 is meant to be a Boolean type, but is instead now mapped to a Texture2D.

You mean the one with DoClip? I don't understand what you mean.

angrywasp:
Yes, that's what i mean. DoClip is a boolean, but it is mapped to a Texture2D by [this code](https://github.com/angrywasp/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/Effect.cs#L281).The CB is initialized from the item at the same index in the Parameters array. Which does not appear to work as we would like in this situation.

But i have been thinking about it and i believe the fix is quite simple. I will implement a change i believe will work and get back to you. 

I also think we need to get this PR merged. I think we will get a lot more attention on this once it is merged and it will make it a bit easier for me to create some PR's on this code as well

Jjagg:
> I also think we need to get this PR merged. 

That's fine by me. This PR does not affect any existing path.
I just rebased it, should be good for merge once the build passes.

@tomspilman What do you think?

angrywasp:
Committed a small fix [here](https://github.com/angrywasp/MonoGame/commit/afd76cdef9d939368482e34db051ce6e47a3b858) to correct the param/cb mapping. rather than a loop with a 1-1 mapping i created a new method which maps the cb to parameter by name. Possibly a naive fix, but i figure the parameters should have unique names. In any case, i can now load my multipass effect without error and render objects on screen using the effect.

Good to see your branch is merge ready. Once this PR is merged i will lodge a PR with the commits i have made.

KonajuGames:
I'm happy to merge this.  The need to enable the service in Protobuild is fine for now for further development.

It's a large change, even if it is disabled by default, so we'll get @tomspilman's thoughts as well before pressing the button.

angrywasp:
So i have ported all my shaders and have confirmed a few of my preliminary findings. I (as other people have noticed) needed to reverse my matrix multiplication and i also noticed the Y axis is inverted and winding order reversed. I have read up on this and found [this](https://matthewwellings.com/blog/the-new-vulkan-coordinate-system/) article. it appears the shader code is being compiled to use the Vulkan coordinate system. The article details the exact problems i was facing with my shaders.

I have noted that if you use SetValueTranspose instead of SetValue to pass your transform matrices to the shader then the matrix multiplication is the same as it is with HLSL. So it may be a simple matter of a preprocessor directive to pass SetValue Calls to SetValueTranspose.

The article i have linked to also provides a shader fix and a correction matrix to multiply against in code to correct the coordinate system changes. This may also be worth implementing to maintain as much compatibility as possible with XNA. I can see a bunch of issues arising from this when people try to port their shader code and get unintended results.

Zkylar:
Hey guys, I'd like to add, I've  had great success with this change set. With the help of this change I was able to port everything I had done on windows to my Mac.I've now been able to port my entire pipeline to Mac with help this change. I was also able to build stuff for {iPhone/IPad } & Mac OS thanks to this change.  I've ported all my shaders from HLSL to GLSL so far on this change set and have had very little issues which I've had to work around. Issues I had to work around include the one Technique only working thing {the first one} and multiple passes not working and the transform being swapped which affect my fonts. I also had to make some adjustments to params for matrix array passing into the shader for my skinned models. but aside from that it all worked very nicely.  Thank you for the work not this It is much appreciated and cannot wait for the merge :-)

KakCAT:
Hi @Zkylar

did you made any other change to make the shaders work in iOS? I'm getting errors like "GL does not contain a definition for Uniform1fv" after all the patches have been applied. (in iOS and android)

The windows GL version works great

Zkylar:
Howdie KakCat, I had similar compile errors on the mac with Uniform1iv delegate . I had to make change to use Uniform1fv to make this eventually build and run. there was issues with Uniform2iv and Uniform4iv that I recall too, these did not seem to want to build for iOS.

tomspilman:
I've been crazy busy lately... but I need to give this a shot soon.

Zkylar:
Im seeing some strange texture & mesh behavior after the rebase of the glfx branch. digging into it briefly I think it could possibly be related to the work for instance drawing so I'm not sure if this is the right place to ask / place this.

best I can show is example of what I am seeing. 

in the previous version of the glfx code I had a test scene that looks like this, {notice the texture mapping on the floor its uniform}:

working

after the rebase was done and I built the monogame.framework.dll I get this:

A. Area A notice the texture mapping is wrong / off.
B. Area B is what the texture mapping should look like

broken

After I debug a bit I think it might be related to this part of new code which seems the have changed:

screen shot 2017-09-05 at 7 00 57 pm

when I replace it with older code that was on the branch before rebase it seems to work like it used to:

screen shot 2017-09-05 at 7 03 02 pm

I'm digging into the PlatformDrawIndexedPrimitives changes now but lot if it is still bit over my head and could do with some pointers or help.

Jjagg:
@Zkylar kan you try to apply the changes of #5912? That might help with the meshes. Not sure about the texturing issue. 

Pm me on gittee if you want to chat.

Zkylar:
That worked like a charm Jjagg,https://github.com/MonoGame/MonoGame/pull/5912 fixed the issue and texture and mesh funnies are gone :-) Thanks a Mil!!! :-)

Jjagg:
Awesome :)

RUSshy:
Any plan to make GLSL shaders compilable/usable without  that huge dependency ?

Just like in any open gl library or even SFML/SDL, it is fast and easy to compile/use shader, why in monogame it is complex, not everyone care about ps4/3ds, most games are for desktop/mobile

Jjagg:
@RUSshy What "huge dependency" do you mean? glsl-optimizer? That's there mostly for mobile. Mobile GLSL compilers don't always optimize and optimizing them beforehand can make a very large difference.

This is complex because it needs to integrate with the existing API.

RUSshy:
@Jjagg i mean 2MGFX , other libs don't need that, we can compile and use shaders easily

Jjagg:
We need reflection data for the Effect API, so we can't have GLSL shaders loaded without parsing the file or using tools that expose reflection data.

Zkylar:
@Jjagg,  Is there still a plan to get glfx changes into develop at some stage , any eta for a rebase ?

Jjagg:
@Zkylar 

>  Is there still a plan to get glfx changes into develop at some stage

This will not get merged, at least in the near future. The focus will be on letting users write their shaders in HLSL, but providing modern shader features and a proper way of translating them to GLSL.

That said, you can keep using this if you'd like. I'll rebase it :) (if I can manage, lots of conflicts 😨)

Zkylar:
Ok thanks JJagg, how will this work for mac os peeps ? we'll be writing shaders in HLSL and it will be compiled to GLSL to be used on iOS and mac ? is this going to be done in 2MGFX ?

Jjagg:
> Ok thanks JJagg, how will this work for mac os peeps ? we'll be writing shaders in HLSL and it will be compiled to GLSL to be used on iOS and mac ? is this going to be done in 2MGFX ?

Yes, the workflow will be exactly the same as it is now, except it should work on Mac/Linux and gives us more control to add features. That said, it will take a lot of work to get to that point, so for now this is still your best option.

I rebased the branch and cleaned it up in the process, hope everything is working well, let me know if you run into any issues. Feel free to set up a PR against this branch if you make any improvements :)

Zkylar:
Ok sounds good! Thanks very much JJagg! much appreciated!

Jjagg:
Closing this because this is not the path we're taking for cross-platform effect compilation. Of course anyone that wants to can still use this branch :) (though I should probably rebase it again).

TheSpydog:
@Jjagg I'm curious how the cross-platform 2MGFX tooling is coming along? Windows-only effect compiling has been a thorn in my side for a long time, so I'm very interested in the subject. :) Is there a PR/branch/discussion that I can follow?

tomspilman:
@TheSpydog - The plan has changed just based on experience and resources.

The old plan was to support multiple shader languages.  The main other one to support was GLSL.  This has proven difficult and time consuming and in the end didn't really add much new functionality.

The new plan is to push harder into just supporting HLSL and cross compile to all other languages/platforms we need.  This is how Unity does it... this is how Unreal does it... this is how we should do it.

The main blockers are replacing our existing MojoShader HLSL to GLSL cross compiler with something that doesn't require a compiled HLSL shader to start with.  This would then let us compile HLSL on Mac to target GL platforms there.

The other more long term path is using [Microsoft's new HLSL compiler](https://github.com/Microsoft/DirectXShaderCompiler) which does run on Mac and Linux and can generate both new DXIL and SPIR-V shaders.

On consoles they all pretty much support an HLSL like language or have their own HLSL to native shader language cross compiler which we currently use.  No consoles have development environments on Mac or Linux, so that isn't a concern.

PR #5356, opened 2016-12-09
[SDL] Mouse Improvements

cra0zy:
Stuff done:
 - Fixes mouse tracking outside of window
 - Optimizes several pieces of code
   - HasFlag is replaced by a bitwise and operator
   - State enum has been replaced by a set of const int since we are not using it with any enum like stuff
   - clientBounds will only be calculated if the version of SDL is new enough

### DIFF
--- a/MonoGame.Framework/SDL/SDL2.cs
+++ b/MonoGame.Framework/SDL/SDL2.cs
@@ -252,24 +252,23 @@ internal static class Sdl
             Close,
         }
 
-        [Flags]
-        public enum State
+        public static class State
         {
-            Fullscreen = 0x00000001,
-            OpenGL = 0x00000002,
-            Shown = 0x00000004,
-            Hidden = 0x00000008,
-            Borderless = 0x00000010,
-            Resizable = 0x00000020,
-            Minimized = 0x00000040,
-            Maximized = 0x00000080,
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/SDL/SDL2.cs
```diff
@@ -252,24 +252,23 @@ public enum EventId : byte
...
+            public const int Resizable = 0x00000020;
+            public const int Minimized = 0x00000040;
+            public const int Maximized = 0x00000080;
+            public const int Grabbed = 0x00000100;
+            public const int InputFocus = 0x00000200;
+            public const int MouseFocus = 0x00000400;
+            public const int FullscreenDesktop = 0x00001001;
+            public const int Foreign = 0x00000800;
+            public const int AllowHighDPI = 0x00002000;
+            public const int MouseCapture = 0x00004000;
```
KonajuGames: With an enum, the compiler will typically use the integer value directly in the places where the enum is used.  With these const int members of a class, will the compiler be doing a read of the memory containing the class members instead?  I'm just not sure this gives us any benefits other than avoiding the casts when testing for bits.

KonajuGames:
Merging.  Thanks @cra0zy

PR #5362, opened 2016-12-14
Avoid IndexOutOfRangeException when fc-match returns no matches

rgson:
This change avoids the `IndexOutOfRangeException` that's currently being thrown if `fc-match` doesn't return any matches at all, e.g. if `fc-match` (`fontconfig`) isn't installed.

The exception prevents building spritefonts even when `fc-match` isn't actually needed, such as when the font file is located in the same directory as the `.spritefont` file.

The program already handles the case when `fc-match` fails to find the correct font. I simply handle this "not installed" case the same as the "no match" case.


tomspilman:
Thanks @rgson !

@cra0zy - I think you wrote this code originally... this look good to you?

cra0zy:
This looks good to me :+1:

tomspilman:
Thanks @rgson  and @cra0zy !

PR #5366, opened 2016-12-19
[Linux Installer] Check for monodevelop-stable

cra0zy:
Currently there is no way for Linux users to get MonoDevelop 6 and MonoGame easily (unless they are running Arch Linux), so I decided to make a MD 6 general purpose Linux installer (https://github.com/cra0zy/monodevelop-run-installer) and to make sure that I don't conflict with the existing version I named commands with `-stable` suffix.

This PR simply adds `monodevelop-stable` as a possible MonoDevelop command to the installer.


KonajuGames:
Merging.

PR #5370, opened 2016-12-22
WinDX: Frame latency fix

MichaelDePiazzi:
I recently found an issue in MonoGame while investigating some frame latency issues in my game. The maximum frame latency was being set to 3 frames when v-sync was disabled.

The problem was that it seems to have been confused with the sync interval setting ([documentation on sync interval](https://msdn.microsoft.com/en-us/library/windows/desktop/bb174576(v=vs.85).aspx)). So when v-sync was disabled, the frame latency was set to 0 (which means to use the default value of 3) ([documentation on maximum frame latency](https://msdn.microsoft.com/en-us/library/windows/desktop/ff471334(v=vs.85).aspx)).

The PR fixes it so that the maximum frame latency is always 1. I've also tidied up some things to make the code clearer.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -745,8 +745,6 @@ namespace Microsoft.Xna.Framework.Graphics
                 multisampleDesc.Quality = quality;
             }
 
-            int vSyncFrameLatency = PresentationParameters.PresentationInterval.GetFrameLatency();
-
             // If the swap chain already exists... update it.
             if (_swapChain != null)
             {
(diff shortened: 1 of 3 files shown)


tomspilman:
This looks like a bug I must have introduced when I first wrote all this code or when I added vsync support years ago now.

@nkast - Could this also affect the UWP input latency issues as well?

tomspilman:
Anyone have some time and project to test this on?

@MichaelDePiazzi - Is there any sort of useful unit test we could write to verify this?

MichaelDePiazzi:
> Is there any sort of useful unit test we could write to verify this?

@tomspilman - I'm not sure... We could maybe confirm that it has been set to 1 after creating the graphics device (by reading back the DXGI device `MaximumFrameLatency` property). It'd be specific to DirectX though... Does that sound worthwhile?

tomspilman:
> It'd be specific to DirectX though

Yeah I don't think that would be all that useful.

A more correct test might be to render 3 different colored frames then take a screenshot of the game window's "front buffer".  Maybe via a Windows GDI call to see if the expected frame is visible.

We would have to port this front buffer screenshotting feature in the unit test framework to Mac/Linux and other platforms...  but it would be better than just a DX only test.

MichaelDePiazzi:
> Anyway... not expecting you to do all that. This probably can go as is.

Ok, cool. I've got a bit on my plate atm, so I don't think I would have time to tackle this. Also, I don't have a Mac or Linux setup so I'd only be able to get it working on Windows anyway.

nkast:
> @nkast - Could this also affect the UWP input latency issues as well?

We might got a lag of 2 frames between an event and seeing the result, but it wouldn't affect the time we process the events or cause any delay in the arrival of the events.

tomspilman:
Thanks @MichaelDePiazzi  and @nkast !

PR #5139, opened 2016-08-28
Fixes and XNA compliance for GraphicsDeviceManager and GraphicsDevice

Jjagg:
This replaces #5069 because I messed up something. Sorry for the inconvenience!

This is a start to getting GraphicsDevice and GraphicsDeviceManager events right and handling PresentationParameters the same as in XNA.
- fixes #539
- fixes #4433
- fixes #3114
- fixes #4740 
- fixes #5173 
- helps with #3572 (DirectX only)

PR #5371, opened 2016-12-22
WinDX: Fix to ensure window is maximised when starting in fullscreen

MichaelDePiazzi:
I came across a rare problem which seemed to only show up on older/slower computers. Sometimes when starting in borderless fullscreen (i.e. `HardwareModeSwitch = false`), the window would not get maximised.

After some investigation, I could see that it was maximised at first but was then immediately restored to it's original size after that. It turns out the problem was that `Form.Show` was being called twice - once in `WinFormsGameWindow.Initialize` (via `WinFormsGamePlatform.BeforeInitialize`), and again in `WinFormsGameWindow.RunLoop`. So to fix this, I simply changed `WinFormsGamePlatform.BeforeInitialize` to only set the form's `ClientSize`, and let `RunLoop` call `Show` afterwards.

### DIFF
--- a/MonoGame.Framework/Windows/WinFormsGamePlatform.cs
+++ b/MonoGame.Framework/Windows/WinFormsGamePlatform.cs
@@ -49,7 +49,7 @@ namespace MonoGame.Framework
 
             _window.EnableClientSizeChangedEvent(false); // Disable ClientSizeChanged event while the window is initialised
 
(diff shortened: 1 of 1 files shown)


MichaelDePiazzi:
Closed as I just noticed this was fixed by PR #5139, specifically in commit 49c3738. Not sure if this commit was intentionally fixing this specific issue, but thanks again anyway @Jjagg! :)

Jjagg:
Not really, I just noticed form.Show was being called twice. I'm glad that solved an issue :p

PR #5372, opened 2016-12-22
WinDX: Fix to find closest available mode for CurrentDisplayMode

MichaelDePiazzi:
**UPDATE 29-Dec: As per discussion, changed this to force the current display mode to the desktop resolution even if it's not an available mode.**

I've found that `GraphicsAdapter.CurrentDisplayMode` can sometimes be null in certain circumstances because the reported desktop resolution isn't one of the available modes. This PR attempts to find the next closest mode if a mode matching the desktop resolution isn't found.

An example of this happening that I saw was crash reports showing 2715x1527 as the desktop resolution, but modes only going up to 1920x1080. After some digging I found that 2715x1527 is a resolution associated with an NVIDIA feature called Dynamic Super Resolution (DSR). I tried enabling this feature on my machine, but when I did it actually showed 2715x1527 as an available mode. So I can only assume that the machines with this problem somehow had their DSR settings messed up. So at least with this PR, `CurrentDisplayMode` falls back to the next closest resolution on these machines.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
@@ -111,6 +111,9 @@ namespace Microsoft.Xna.Framework.Graphics
 
             adapter._supportedDisplayModes = new DisplayModeCollection(modes);
 
+            if (adapter._currentDisplayMode == null) //(i.e. desktop mode wasn't found in the available modes)
+                adapter._currentDisplayMode = new DisplayMode(desktopWidth, desktopHeight, SurfaceFormat.Color);
+
             return adapter;
         }
 
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.DirectX.cs
@@ -101,9 +101,9 @@ namespace Microsoft.Xna.Framework.Graphics
 
                     modes.Add(mode);
 
-                    if (mode.Width == desktopWidth && mode.Height == desktopHeight && mode.Format == SurfaceFormat.Color)
+                    if (adapter._currentDisplayMode == null)
                     {
-                        if (adapter._currentDisplayMode == null)
+                        if (mode.Width == desktopWidth && mode.Height == desktopHeight && mode.Format == SurfaceFormat.Color)
                             adapter._currentDisplayMode = mode;
                     }
                 }


tomspilman:
> This PR attempts to find the next closest mode if a mode matching the desktop resolution isn't found.

@MichaelDePiazzi - So this is your solution to the problem and not actual XNA behavior correct?

Do we know what XNA actually does in these cases?

MichaelDePiazzi:
@tomspilman - Correct. I have no idea what XNA would do in this situation. Unfortunately I was not able to reproduce the problem myself, and the crash reports I received were anonymous so I could not investigate further. However, I can confirm that after putting this fix into my game 6 weeks ago I haven't received any further crash reports about this issue.

tomspilman:
@nkast @mrhelmut @KonajuGames - Thoughts on this behavior?

MichaelDePiazzi:
If this solution isn't desirable, then I have a simpler alternative - If the current display mode wasn't found after enumerating the available modes, then just set it to a new mode using the reported desktop resolution. i.e.
```c#
if (adapter._currentDisplayMode == null)
    adapter._currentDisplayMode = new DisplayMode(desktopWidth, desktopHeight, SurfaceFormat.Color);
```

tomspilman:
> then just set it to a new mode using the reported desktop resolution

That seems to be more reasonable and safer to me.

MichaelDePiazzi:
> That seems to be more reasonable and safer to me.

Fair enough, I'll push another commit doing exactly this. The only reason I did it this way in the first place was because the crash reports I got reported the desktop resolution as being higher than the maximum available resolution (e.g. desktop is 2715x1527, max available is 1920x1080). So I figured it'd be safer to choose from the closest available resolutions rather than using one that wasn't listed.

tomspilman:
> So I figured it'd be safer to choose from the closest available resolutions rather than using one that wasn't listed.

I see your point.  Hard to say which is the best decision here.

I see it as people depending on the current mode to get the current desktop resolution.  They still need to account for available resolutions when setting things and not just assume desktop resolution always works for fullscreen.

MichaelDePiazzi:
> I see it as people depending on the current mode to get the current desktop resolution. They still need to account for available resolutions when setting things and not just assume desktop resolution always works for fullscreen.

That's an excellent point. From this point of view, it does make way more sense to use the reported desktop resolution.

MichaelDePiazzi:
Ok, made the change, rebased, and squashed. Also did a small performance tweak when setting the current display mode.

mrhelmut:
I've been forcing fullscreen desktop resolution as default resolution in our game since the beginning and never had a crash related to this in our database. It's typically safe to do this.
(I've also encountered some adapters not listing all of their supported modes, but I don't really know what are the consequences, and desktop resolution always worked fine; I don't know if I ever had users with nvidia super resolution though, I'll try to ask a few testers.)

tomspilman:
Thanks @MichaelDePiazzi .... this looks good to me.

If @mrhelmut is good with this change then I am too.

MichaelDePiazzi:
:+1:

tomspilman:
Going to merge this... thanks @MichaelDePiazzi !

PR #5376, opened 2016-12-24
WinDX: Fixed potential crash when setting MediaPlayer.Volume

### LINKED ISSUE #4298: MediaPlayer.SetChannelVolume may fail despite null check (WinDX)
Hello,

Despite [the null check](https://github.com/mono/MonoGame/blob/083d88ce8385cf6a454a372e1d5bbafbe83243cb/MonoGame.Framework/Media/MediaPlayer.WMS.cs#L150), it may happen that `MediaPlayer.SetChannelVolumes()` raise an exception because of the async nature of Media Foundation (Windows DirectX projects).

Two different (but related to the same problem) exceptions can occur.

1) [On this line](https://github.com/mono/MonoGame/blob/083d88ce8385cf6a454a372e1d5bbafbe83243cb/MonoGame.Framework/Media/MediaPlayer.WMS.cs#L154)

```
SharpDX.SharpDXException
 at SharpDX.MediaFoundation.AudioStreamVolume.GetChannelCount(Int32& dwCountRef)
 at Microsoft.Xna.Framework.Media.MediaPlayer.SetChannelVolumes()
```

2) [On this line](https://github.com/mono/MonoGame/blob/083d88ce8385cf6a454a372e1d5bbafbe83243cb/MonoGame.Framework/Media/MediaPlayer.WMS.cs#L155) or the previous one

```
System.NullReferenceException
 at Microsoft.Xna.Framework.Media.MediaPlayer.SetChannelVolumes()
```

Situation 1 happens when the `_volumeController` is not yet ready.
Situation 2 happens when the `_volumeController` gets reset past the null check (i.e. a new session started but didn't went through the whole initialization yet).
Both are related to the async nature of starting a new session.

This problem rarely occurs, over thousands of sessions, we had 4 of these crashes reported.

@MichaelDePiazzi May I have your thoughts on this one? Every async initializaiton should be fine, I'm quite puzzled.

MichaelDePiazzi:
I'm fairly confident this fixes #4298

In certain circumstances, setting `MediaPlayer.Volume` could cause a crash. I actually managed to reliably reproduce this by continually setting the volume and playing a new song every few seconds.

The cause of the crash was due to the code not properly considering that different threads could be messing with the volume controller. So if `MediaPlayer.Volume` was being set at the same time that a new song started playing, an exception could occur.

The fix was to simply use a lock when dealing with the volume controller. After adding the lock, I could no longer reproduce this crash.

### DIFF
--- a/MonoGame.Framework/Media/MediaPlayer.WMS.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.WMS.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Xna.Framework.Media
     {
         private static MediaSession _session;
         private static AudioStreamVolume _volumeController;
+        private static readonly object _volumeLock = new object();
         private static PresentationClock _clock;
 
         private static Song _nextSong;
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @MichaelDePiazzi ... this looks like a safe fix to me.

PR #5377, opened 2016-12-27
Extract wave blockalign for adpcm format from RIFF header

SeriousMartin:
This is a fix for the problem discussed here: https://github.com/MonoGame/MonoGame/issues/4446

The block alignment of ADPCM wave files is extracted from the RIFF header and overrides what is calculated in DefaultAudioProfile.ProbeFormat.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/Audio/AudioContent.cs
@@ -137,7 +137,18 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Audio
                         fs.Read(rawData, 0, rawData.Length);
                     }
 
-                    var stripped = DefaultAudioProfile.StripRiffWaveHeader(rawData);
+                    AudioFormat riffAudioFormat;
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks @SeriousMartin ...  this passes all our unit tests, so I think it is good.

I'll merge test it against our console platforms soon too.

SeriousMartin:
That was fast. Thanks!

tomspilman:
> That was fast. Thanks!

That is entirely because I personally wrote the unit tests for audio and feel pretty confident this PR didn't break any existing cases:

https://github.com/MonoGame/MonoGame/blob/develop/Test/ContentPipeline/AudioContentTests.cs#L205

If it wasn't for that I would have been concerned about a merge.  The speed of any of our merges is directly tied to if I am fairly sure it didn't break anything.

Unfortunately few people add to the tests which means this doesn't happen as much as I would like.

If I could spend 2 or 3 months just writing unit tests this project would be much easier to maintain...  but I don't have the luxury and few others contribute in that capacity.

SeriousMartin:
Well known situation...

Issue #2211, opened 2014-01-28
Network gamer id is always 255?


panthernet:
Tried to play with network in monogame 3.0.1 but it seems that player Id is always 255 making it difficult to distinct network players :( 

Property: NetworkGamer::Id

gustavolsson:
Hi,

I'm trying to port my multiplayer kart racing game to PC and just encountered this issue in the DekstopGL path of MonoGame 3.5.

Has anyone looked into the net code since this issue was opened?

With the exception of NetworkGamer.Id always being 255 for everyone except the host and the lack of SignedInGamer.IsGuest/support for multiple SignedInGamers, the implementation seems to work as it should. I've tested transitions from Lobby to Game state, players joining/leaving and sending messages. I haven't tested host migration yet.

PR #5378, opened 2016-12-27
WIP: Net namespace rewrite (old)

gustavolsson:
Hello,

I started to port my kart-racer "Seaside Racing" to PC (from XBLIG) and got the game up and running in very little time, great work everyone! However, I ran into problems with the Net namespace. In particular, issue #2211 and the realization that supporting multiple local gamers would be problematic caused me to rewrite the code from scratch.

In addition to implementing everything XNA supported (except host migration), I wrote a simple Master Server to support NetworkSessionType.PlayerMatch and NetworkSessionType.Ranked (ie. online modes).

The master server is the only thing that is not standard XNA. The server itself is just a simple console application:

```
public static void Main(string[] args)
{
    NetworkSessionMasterServer server = new NetworkSessionMasterServer();

    server.Start("e3ba2560-9fb5-4085-9c3a-b73fb4257f0e");

    while (!Console.KeyAvailable || Console.ReadKey().Key != ConsoleKey.Escape)
    {
        server.Update();

        Thread.Sleep(20);
    }

    server.Shutdown();
}
```

and the game is modified so that it knows about the master server in the following way:

```
public static void Main (string[] args)
{
    // Set master server address before creating a NetworkSession:
    NetworkSessionSettings.MasterServerAddress = "13.81.113.129";

	using (var game = new NetRumbleGame ()) {
		game.Run ();
	}
}
```

I updated the README to describe my fork but I don't know how to exclude it from this PR. Other than that, everything should be included in the PR.

So, what do you think? Here are my notes from the README:

**This Fork**

My fork of MonoGame implements the Net namespace in order to compile networked games/apps that were built for XNA on any platform. I wanted to port my kart-racing game "Seaside Racing" to MonoGame and I decided to reimplement the Net namespace instead of rewriting the networking code in the game. The main reason to use the Net namespace for a new project is that the original API is very elegant, handles player management and lobby-to-game transitions. This implementation supports multiple backends and it should be straightforward to add support for new platforms. At the moment, the default backend uses the Lidgren library for low level networking tasks and the backend will work on any platform that supports standard .NET sockets.

**Implementation details**

Since a user can choose to use any network topology for their game, the internal implementation uses a peer-to-peer topology as it is the most general case. If the user wants a client/server topology, all traffic can be routed through the designated host or any other peer. This is inline with the original XNA implementation.

The peer that starts a game using the NetworkSession.Create() method becomes the host of the game. The host of the game broadcasts its existence to the master server (if NetworkSessionType is PlayerMatch or Ranked as opposed to Local or SystemLink) and responds to discovery requests. Any peer that connects to a host using the NetworkSession.Join() method becomes an ordinary peer. In the future, host migration could make it possible for ordinary peers to become hosts, but it is not implemented yet.

The master server is responsible for introducing (ie. finding the endpoint and performing NAT punchthrough) a peer to a host of an online game. Once connected to a host, the peer is introduced to the other peers already in the game by the host. Peers only allow introductions/connections from other peers whose endpoints are specified in an allowlist and only allow changes to the allowlist from the host. Once fully connected, the signed in gamers of the peer are able to join. This procedure ensures that all peers in a game are fully connected to each other in order to support a true peer-to-peer topology.

The session state is synchronized across all peers. Each signed in gamer that enters a game gets its own unique id that is synchronized across all peers. Incoming messages from unknown gamers are delayed until the gamer in question joins the game on the local machine, which is important because user messages do not need to be ordered and may arrive before the internal gamer join message.

Overall, there are 2 caveats of the implementation that users must be aware of:

* A peer cannot be behind the same router as the master server since the master server must know the external endpoint of the peer
* ~~A peer cannot be behind the same router as the host since the host must know the external endpoint of the peer~~
* Requiring a fully connected network **might** be a problem sometimes? I haven't run into it yet though.

**Future work (Not implemented)**

* Clean up responsibility of internal messages vs Session
* Host migration
* Cheat protection


mrhelmut:
That's quite a PR!
Mind the dependencies submodule regression.

mrhelmut:
All good, thanks.

tomspilman:
@gustavolsson - This is a pretty huge change... I will try to find some time to review it along with @dellis1972 and @KonajuGames .

I want to find a way to include your work, move forward with plans @dellis1972 has had for networking, and support more than just PCs (consoles and mobile).

Jjagg:
This will make most of #5091 useless, but it's worth redoing since this PR is awesome :D 

> A peer cannot be behind the same router as the host since the host must know the external endpoint of the peer

Any way around that?

gustavolsson:
@tomspilman Sounds good! I think it should be quite straightforward to implement a new backend for platforms where Lidgren does not work. The logic (pending connections, allowlist handling, gamer management etc) is all separate from the actual network calls (that exist in the Backend subfolder).

@Jjagg Thanks! The only way around that limitation I can come up with is to make the master server responsible for introducing a new peer to _all peers_ in the game, not just the host, but that complicates things considerably..

Jjagg:
The errors are because you're using C# 6.0 syntax. You can't have auto properties with only getters :/

gustavolsson:
Aha, too bad. I can rework it when I have some free time..

dellis1972:
This PR while big is a great step to getting the networking working. 
Once the build is fixed we should definitely look at merging it.

dellis1972:
We should probably revert the README.md changes though.

dellis1972:
@gustavolsson nice :) 

I was thinking the text you had in the read me could probably go in the documentation directory instead. Also I think @tomspilman might mention the MG licence header needs to be on the new MG file (but not the Lidgren ones).

gustavolsson:
I've been working a bit more on this yesterday and today. It is not ready yet though!

The worst caveat of the PR (nr 2 from the list above, @Jjagg) should now be fixed but more testing is required. I will update the published NetRumble executable when I some more time.

Basically, I updated Lidgren to the latest version for better NAT punch code (old one only worked one way) and rewrote the generic EndPoint to use temporary GUIDs for identification (allowlist, pending peers) instead of the external IPEndPoint used before. This should solve the NAT traversal issues when some peers are behind the same router as the current host of the game. All of these changes are in the Backend folder.

Next up is to do more testing and to remove the C# 6 dependencies..

gustavolsson:
Thanks @dellis1972! Ok, I will look into that for my next session as well.

dellis1972:
I had a thought the other day. Can we add support for reading the IP/URL of
the master server from the app config file? it would mean people would not
need to recompile the app if the IP changed.

gustavolsson:
Good idea! Should be straightforward, right? Right now it's just a public property in NetworkSessionSettings. It could stay there but the default could be read from the app config file.

dellis1972:
@gustavolsson did you managed to do the config/README changes? Also looks like this branch needs rebasing

DDReaper:
Noticed this PR removes most of the Local Gamer implementations. Does the update still allow for local network play @gustavolsson  ?

dellis1972:
I didn't spot that. I'll have to have another look.

gustavolsson:
@dellis1972 I haven't had the time to work on it since I've been busy finishing up university. If I remember correctly, I was testing network play in different configurations (host and peer behind same router + one remote peer etc) using the master server. The windows firewall caused connection problems but once I realized that and disabled the firewall it worked. However, I haven't done extensive enough testing after that realization. It's tricky to debug because the reported IPs change when the router / network card realize where a computer really is. (I don't remember the exact details)

@DDReaper I completely rewrote the system (removing the old files) so I'm not sure what you mean. This PR supports local area network play.

Jjagg:
@gustavolsson Any chance of reviving this? Or would you mind if someone else hijacked this PR? Would love to see this get merged!

gustavolsson:
@Jjagg Hi and thanks for the interest! I would very much like to finish this some day, but since I started working full time I've had very little time for hobby projects.

What's left is to iron out the last bugs related to NAT punch-through (gamer introductions), but it is difficult to test properly. I also feel that the code should be re-structured a bit to make it more elegant and easier to understand (in particular, the responsibility of individual messages vs that of the NetworkSession). I would not mind if someone continued working on it as long as I got some notes about what was changed and why so I can get back into it when I have the time :)

gustavolsson:
I want to dig into this again now that I have some more free time, but don't mind me just yet :) I've realized that some things need to be cleaned up and some things need to change.

Jjagg:
@gustavolsson That's awesome! Take your time and let us know if we can help out because this is way too good to let slip away :)

gustavolsson:
Closing in favor for #6556

PR #5382, opened 2017-01-02
fixed touch handling bug

sergiokoo:
This is a small obvious fix for a little bug: when using .RemoveAt() with List<>, it should be iterated backward.


Andreyul:
No, because `break`.

tomspilman:
> This is a small obvious fix for a little bug:

@sergiokoo - Can you elaborate what the bug is?

There is nothing wrong with that loop that I can see.  It searches the loop for a particular touch then removes it and breaks out of the loop.

sergiokoo:
I am sorry, I did not notice there is a 'break' statement. However the proposed change did fix a problem that I faced and I decided to share it (too quick, my fault).
Anyway, there is a bug somewhere else because:
1. Originally, in my project I am alarmed at line 215, TouchLocation.cs:
Debug.Assert(State != TouchLocationState.Released, "We shouldn't be changing state on a released location!");
2. If I just continue, all subsequent calls to GetState() in the app will have one non-removed touch event.
3. If I apply the proposed change to reverse direction of iteration in line 112, TouchPanelState.cs, then I do not have problems (1) and (2).
4. The bug seems to be related to the fact that on entry to ApplyTouch() the parameter 'state' contains 2 touch events 'Released' and 'Pressed' in this order! Because of originally forward iteration and a 'break' statement, the line 121 "state.RemoveAt(i);" is never called and the event remains in the collection forever causing problems (1) and (2). While my fix was logically incorrect, it solved these problems by picking the 'Pressed' event first and causing .RemoveAt() to happen.
I am still new to MonoGame and not quite sure what should be the best fix here. Maybe do sort a collection first or otherwise make sure that Pressed event comes before Released? Or maybe remove the 'break' statement and do iterate (backward) through all touch events disregard their order? Guys, please check as you know the system better.

KonajuGames:
So the bug is that there are two states in the list, but the Released state is before the Pressed state?

sergiokoo:
At least such case leads to the issue but I do not know if this combination of events in such order is expected. It looks like when ApplyTouch() is called to register a new Pressed event the collection does already contain an obsolete(?) Released event from one of previous clicks (why?). In my app, I have to set both TouchPanel.EnableMouseTouchPoint and TouchPanel.EnableMouseGestures to 'true' and the problem appears only when use a mouse on PC. When running on Lumia all seems to work ok...
A simple workaround could be to use a proposed fix and remove that 'break;' statement so the matched pair of Pressed-Released will be found disregard the existence/order of events, however it does not explain why such combination was possible at the first place.

tomspilman:
> It looks like when ApplyTouch() is called to register a new Pressed event the collection does already contain an obsolete(?) Released event from one of previous clicks (why?). 

It has been a long time since I did the big refactor of this code so I don't remember.  However I think we do first need to work out if having more than one event for the same touch id should ever happen.  Because that could be the bug if that was the assumption I made originally.

sergiokoo:
I think I have found the sequence which reproduce the problem:

frame #1
1. ApplyTouch([], {Id:1,Pressed})
2. TouchPanel.GetState()
result: _touchState == [{Id:1,Moved}]

frame #2
3. ApplyTouch([{Id:1,Moved}], {Id:1,Released})
result: _touchState == [{Id:1,Released}]

frame #100
4. ApplyTouch([{Id:1,Released}], {Id:1,Pressed})
result: _touchState == [{Id:1,Released},{Id:1,Pressed}]

frame #101
5. ApplyTouch([{Id:1,Released},{Id:1,Pressed}], {Id:1,Released})
 => Debug.Assert!!!
result: _touchState == [{Id:1,Released},{Id:1,Pressed}]

After that, any new clicks will cause _touchState to grow linearly and infinitely, for example: 

frame #200
4. ApplyTouch([{Id:1,Released},{Id:1,Pressed}], {Id:1,Pressed})
result: _touchState == [{Id:1,Released},{Id:1,Pressed},{Id:1,Pressed}]

frame #201
5. ApplyTouch([{Id:1,Released},{Id:1,Pressed},{Id:1,Pressed}], {Id:1,Released})
 => Debug.Assert!!!
result: _touchState == [{Id:1,Released},{Id:1,Pressed},{Id:1,Pressed}]

In case TouchPanel.GetState() is not called in step 2 or TouchPanel.GetState() is also called somewhere between step 3 and step 4 the problem will not happen.

Is it obligatory to call TouchPanel.GetState() every frame? If yes, let's forget about my issue :) If not, let's decide how to fix the bug.

nkast:
This bug was fixed by https://github.com/MonoGame/MonoGame/pull/5641.

PR #5385, opened 2017-01-03
Fix GL VS attribute caching bug

### LINKED ISSUE #4909: HLSL losing vertex elements?
I posted this to the community forums after staring at it for several days... I think I've ruled out the possibility of it being something foolish on my end.

http://community.monogame.net/t/hlsl-losing-vertex-elements/7724/

In summary, one call to DrawUserIndexedPrimitives seems to be affecting another several calls later, despite setting a different render target and clearing the graphics device several times in between, etc. For what it's worth, they're both acting upon different techniques in the same effect, although it's not EffectPass.Apply that's causing the issue. Oddly, the vertex shader of the later call in question seems to have all of its vertex elements - except for Position - set to zero, despite the incoming vertex data.

I tried downloading the Monogame source to debug locally, but I had issues compiling it, probably related to reference/external dependencies.

Jjagg:
Fixes #4909. The bug is explained in that issue too, but I'll summarize here.

It comes down to that for OpenGL we cache vertex shader attributes based on a hash of the vertex shader, but when an attribute is not used in the current shader program the compiler will optimize it out and asking the location of that attribute will return -1. That value will be cached however and when another technique with the same vertex shader is later applied, the cached value of the location of that attribute will be fetched even though it might be used and does not actually have location -1. The fix I applied, as suggested by @ed022, is to hash based on the current program, i.e. a hash based on the hash of the vertex and pixel shader. I used the same hash that's used for caching shader programs.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
@@ -947,7 +947,8 @@ namespace Microsoft.Xna.Framework.Graphics
             var vertexDeclaration = _vertexBuffers.Get(0).VertexBuffer.VertexDeclaration;
             var vertexOffset = (IntPtr)(vertexDeclaration.VertexStride * baseVertex);
 
-            vertexDeclaration.Apply(_vertexShader, vertexOffset);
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks @Jjagg ... the bug is obvious now.

PR #5388, opened 2017-01-04
Fix failing XNA tests

Jjagg:
First off XNA does not support PixelBitmapContent<Byte> as stated in the docs: https://msdn.microsoft.com/en-us/library/bb198425.aspx#Anchor_1

Then there were some precision errors, so I added some tolerance for comparing the floats.

Lastly MG behaves different from XNA [here](https://github.com/MonoGame/MonoGame/compare/develop...Jjagg:xnaTests?expand=1#diff-d8ef415c6bf96b2f7ba5d6f89b5cb3b0R202). When copying BitmapContent of a float to BitmapContent of Color XNA only sets the red value while MG sets all channels. Should we follow XNA or stick with what we have?

### DIFF
--- a/Test/ContentPipeline/BitmapContentTests.cs
+++ b/Test/ContentPipeline/BitmapContentTests.cs
@@ -61,7 +61,10 @@ namespace MonoGame.Tests.ContentPipeline
 
             for (var y = 0; y < b2.Height; y++)
                 for (var x = 0; x < b2.Width; x++)
-                    Assert.AreEqual(color1, b2.GetPixel(x, y));
+                    if (typeof(T) == typeof(float))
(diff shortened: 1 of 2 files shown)


Jjagg:
I don't remember why I ignored the DrawOrder test for XNA, currently it passes.

tomspilman:
@KonajuGames - I'll let you check this for merge.

KonajuGames:
I'll check the expansion of float to Color.  It may have some side effect on the specialized DXT3 compression for SpriteFont (which could be fixed to work with it).

KonajuGames:
Included these changes in #5398 that fixes the conversion of float to Color.

PR #5389, opened 2017-01-04
Fixed GraphicsDevice.Present() to check for current render target.

tomspilman:
Fixed `GraphicsDevice.Present()` to correctly throw like XNA when RT is set.


KonajuGames:
Looks good. Merging.

PR #5390, opened 2017-01-04
Partialize GraphicsCapabilities

tomspilman:
I split up `GraphicsCapabilities` into partial class for overloading on console platforms.


KonajuGames:
Merging.

Issue #5261, opened 2016-10-29
EffectPass loses textures when applying shader samplers


hifi:
In the method `SetShaderSamplers` in `MonoGame.Framework/Graphics/Effect/EffectPass.cs`, device textures are overridden by sampler `param.Data` unconditionally.

I have an XNA game that assumes device textures are preserved after applying effects. Adding a null check before overriding the device textures fixes my rendering issues.

```
                if (texture != null)
                    textures[sampler.textureSlot] = texture;
```

Disclaimer: I don't know if this is the correct fix nor I can test this on Windows/XNA myself as I'm porting bling on Linux/DesktopGL.

Danthekilla:
This is the same fix that I had to make years ago. This moves monogame closer to XNA and should probably be fixed in the main branch.

tomspilman:
@hifi @Danthekilla - I have no problems making the fix if I can get a clear test case that shows the failure.

hifi:
@Danthekilla If you know more about this, can you possibly write a test case for PR #5288?

EBrown8534:
Not sure if this is related, but I pulled the sample BloomCombine.fx shader from the Neon Shooter example (https://github.com/MonoGame/MonoGame.Samples/blob/3.8.5/NeonShooter/NeonShooter.Core/Content/Shaders/BloomCombine.fx) to do some testing, and I noticed that it wasn't working.

After some experimentation, it appears that the `s1` register was not being set correctly.

I had to make the following modification:

1. Remove / comment the `sampler BaseSampler : register(s1);` line from the `BloomCombine.fx`
2. Add a new sampler for the base texture:
    ```
    Texture2D BaseTexture;
    sampler2D BaseSampler = sampler_state
    {
        Texture = <BaseTexture>;
    };
    ```
3. Update the draw for the bloom component:
    ```
    //GraphicsDevice.Textures[1] = sceneRenderTarget;
    parameters["BaseTexture"].SetValue(SceneRenderTarget);
    ```

I'm not sure if this is fully related to this issue.

PR #5392, opened 2017-01-04
Fix EffectPass.Apply setting textures to null when it shouldn't

Jjagg:
This is just a failing test that shows what happens issue #5261, no fix yet. I did some cleanup first, so you should only look at the second commit for clarity.

Here's the situation where this issue occurs:
You set the register for your samplers in your effect, but don't bind a texture to it. Then you set the textures using `GraphicsDevice.Texture[i]`. Now you apply a pass of a technique of your effect. In [SetShaderSamplers](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/EffectPass.cs#L98) the `GraphicsDevice.Texture[i]` are set to null because the texture parameters of the effect are never set and we do `textures[sampler.textureSlot] = _effect.Parameters[sampler.parameter] as Texture;`. In XNA the textures you set before applying are still there so we should somehow avoid overwriting the value in this case. I'm not exactly sure how to fix this, I think XNA handles this in a completely different way. I'm not sure how/where XNA stores these textures, but it doesn't do it in the effect parameters like MG. When debugging I noticed XNA did not have *any* parameters for the effect while MG has the 2 implicit textures (stored under the sampler names) because there are two samplers. 
That also means that in MonoGame you can do `effect.Parameters["s0"].SetValue(MyCoolTexture);` while in XNA you can only use `GraphicsDevice.Texture[0] =MyCoolTexture1` (which also works in MG).

Another kind of unrelated detail I noticed while writing tests for this is XNA throws when you try to apply a pass from an effect that is not the current effect and MG doesn't.

### DIFF
--- a/Build/Projects/MonoGame.Tests.definition
+++ b/Build/Projects/MonoGame.Tests.definition
@@ -384,6 +384,9 @@
     <Content Include="Assets\Effects\Include.fxh">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="Assets\Effects\SamplerRegistersEffect.fx">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
     <Content Include="Assets\Effects\TextureArrayEffect.fx">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
--- a/MonoGame.Framework/Graphics/Effect/EffectPass.cs
+++ b/MonoGame.Framework/Graphics/Effect/EffectPass.cs
@@ -119,6 +119,10 @@ namespace Microsoft.Xna.Framework.Graphics
         {
             foreach (var sampler in shader.Samplers)
             {
+                // if sampler.parameter is 255 there is no matching texture parameter
+                if (sampler.parameter == 255)
+                    continue;
+
                 var param = _effect.Parameters[sampler.parameter];
                 var texture = param.Data as Texture;
 
--- a/MonoGame.Framework/Graphics/Shader/Shader.cs
+++ b/MonoGame.Framework/Graphics/Shader/Shader.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		public SamplerState state;
 
         // TODO: This should be moved to EffectPass.
-        public int parameter;
+        public byte parameter;
     }
 
     internal struct VertexAttribute
--- a/Test/Assets/Effects/SamplerRegistersEffect.fx
+++ b/Test/Assets/Effects/SamplerRegistersEffect.fx
@@ -1,0 +1,70 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+#include "include.fxh"
+
+sampler s0 : register(s0);
+sampler s1 : register(s1);
+
+struct VS_INPUT
+{
+	float4 Position : POSITION0;
+	float2 TexCoord : TEXCOORD0;
+};
+
+struct PS_INPUT
+{
+	float4 Position : SV_Position;
+	float2 TexCoord : TEXCOORD0;
+};
+
+PS_INPUT VertexShaderFunction(VS_INPUT input)
+{
+	PS_INPUT output;
+    output.Position = input.Position;
+	output.TexCoord = input.TexCoord; 
+    return output;
+}
+
+float4 PixelShader_Both(PS_INPUT input) : COLOR0
+{	
+    return tex2D(s0, input.TexCoord) * tex2D(s1, input.TexCoord);
+}
+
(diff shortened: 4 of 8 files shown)


tomspilman:
> When debugging I noticed XNA did not have any parameters for the effect while MG has the 2 implicit textures (stored under the sampler names) because there are two samplers.

That sounds like the bug to me.

We shouldn't be generating effect parameters from just the sampler name in 2MGFX.  Without that parameter then the pass apply won't make any changes to the samplers.

Since our existing tests used `BasicEffect`, which has a texture and a sampler, it didn't show this behavior.

Jjagg:
> We shouldn't be generating effect parameters from just the sampler name in 2MGFX. Without that parameter then the pass apply won't make any changes to the samplers.

Should we set SamplerInfo.parameter to -1 or something in that case?

tomspilman:
> Should we set SamplerInfo.parameter to -1 or something in that case?

Yeah something like that and not generate any parameters for them in the `Effect.Parameters` collection.

tomspilman:
Note the fix for this might break some people's existing effects using this sampler name based parameters.  But I think we just have to pay that price to get things working the same as XNA.

tomspilman:
So this broke 3 other of our tests... we need to look closer at those shaders and see what we're missing.

Jjagg:
And there are no ConstantBuffers in the reflection data, so the textures aren't in there either.

tomspilman:
> In that case reflection gives the name of the sampler for the texture resource.

Actually... maybe that is the trick.  Like I suspect this is a syntax error:
```
sampler sa1 : register(s1);
Texture2D sa1 : register(t1);  // Error from duplicate name?
```

My guess is that you cannot do this normally.  So if you find a sampler and texture with the same names then it has to be one auto-generated by the HLSL compiler.

Jjagg:
The problem is the reflection data shows the name of the sampler of the corresponding register for a texture resource, not the name of the texture itself. In fact nowhere in the reflection data is the actual name of the texture.

tomspilman:
> The problem is the reflection data shows the name of the sampler of the corresponding register for a texture resource, not the name of the texture itself.

Ok...  that is a separate problem that has made me jump thru hoops in the past.  You are doing texture sampling using the old style syntax:
```
tex2D(sa1, input.TexCoord); 
```
There is no reference to the texture in that statement... just the sampler.   I suspect the HLSL compiler then assumes you will be using the matching texture register for that sampler.  However I needed the name of the texture which reflection doesn't return in this case.

This is specifically why I started parsing `sampler_state` blocks to get the `Texture` reference.  Also if you use the new HLSL syntax then it works correctly as you expect:
```
te1.Sample(sa1, texCoord)
```

Jjagg:
> This is specifically why I started parsing sampler_state blocks to get the Texture reference.

Yeah, if the texture is set in a sampler_state block we can get the right parameter name.

> Also if you use the new HLSL syntax then it works correctly as you expect:

I noticed this with the 'LooseTexture' in my example.

So to solve the issue should we parse Textures as well? Or do we just recommend the new syntax and document this stuff really well?

tomspilman:
> So to solve the issue should we parse Textures as well?

Well not sure that solves anything.  The user isn't required to set a register on either the sampler or the texture.  So with the following case:
```
Texture2D te1;
Texture2D te2;
sampler sa1;
Texture2D te3;


tex2D(sa1, input.TexCoord); 
```
We would have the name of 3 textures and one binding that only gives us the sampler name and sampler register.  We wouldn't know which texture it is...  we could assume texture registers are implied by the order in the file...  but not sure what the rules there really are.

> Or do we just recommend the new syntax and document this stuff really well?

We could certainly add to the docs here:

http://www.monogame.rocks/documentation/?page=Custom_Effects

Jjagg:
In #5288 @hifi fixed this problem by simply not setting a texture if it's null. Maybe that's actually a reasonable solution here. The test that broke is
```
        [Test]
        public void EffectPassShouldSetTextureEvenIfNull()
        {
            var texture = new Texture2D(game.GraphicsDevice, 1, 1, false, SurfaceFormat.Color);
            game.GraphicsDevice.Textures[0] = texture;

            var effect = new BasicEffect(game.GraphicsDevice);
            effect.TextureEnabled = true;
            effect.Texture = null;

            Assert.That(game.GraphicsDevice.Textures[0], Is.SameAs(texture));

            var effectPass = effect.CurrentTechnique.Passes[0];
            effectPass.Apply();

            Assert.That(game.GraphicsDevice.Textures[0], Is.Null);

            texture.Dispose();
            effect.Dispose();
        }
```

Is there actually a use case for this? If not we can just break this test and have a really easy fix for the problem by merging #5288. And we won't break MG behavior for people that used implicit textures as parameters for setting them. I'm not sure what the alternatives are. We can't match XNA 100% either way. I wonder how they did it...

I just confirmed that using the new syntax you actually get the names of the textures. Don't know what that does for OpenGL though, since samplers and textures aren't nicely seperated like in DX. So if you have one sampler sampling multiple textures behavior would at least be different for GL vs DX.

PR #5394, opened 2017-01-05
Mark XButton1/2 as internal setters.

damianday:


tomspilman:
Good catch @damianday !

KonajuGames:
Tests passed.  Merging.

PR #5398, opened 2017-01-06
Fix conversion of float to Color to match XNA

KonajuGames:
* As a consequence, font glyph packing and DXT3 font texture compression was also fixed to suit.
* Ensured that non-premultiplied SpriteFont works correctly for compressed and non-compressed font textures.
* Changed GraphicAdapter widescreen test to treat 16:10 as widescreen for MonoGame but not for XNA.

Included adapted fixes from @Jjagg #5388
* Disable `PixelBitmapContext<byte>` tests for XNA.
* `PixelBitmapContext<float>` and `PixelBitmapContent<Vector4>` tests that resize the bitmap now use an epsilon comparison.
* Add [Ignore] attribute to `MiscellaneousTests.DrawOrder_falls_back_to_order_of_addition_to_Game` for XNA.

### DIFF
--- a/Test/Framework/Visual/MiscellaneousTests.cs
+++ b/Test/Framework/Visual/MiscellaneousTests.cs
@@ -82,6 +82,9 @@ namespace MonoGame.Tests.Visual {
 	class MiscellaneousTests : VisualTestFixtureBase
     {
 		[Test]
+#if XNA
+        [Ignore]
+#endif
 		public void DrawOrder_falls_back_to_order_of_addition_to_Game ()
 		{
 			Game.PreDrawWith += (sender, e) => {
--- a/MonoGame.Framework.Content.Pipeline/Graphics/Font/GlyphPacker.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/Font/GlyphPacker.cs
@@ -59,6 +59,8 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 		static BitmapContent CopyGlyphsToOutput(List<ArrangedGlyph> glyphs, int width, int height)
 		{
             var output = new PixelBitmapContent<Color>(width, height);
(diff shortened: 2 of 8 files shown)


Test/Framework/Visual/MiscellaneousTests.cs
```diff
@@ -82,6 +82,9 @@ namespace MonoGame.Tests.Visual {
 	class MiscellaneousTests : VisualTestFixtureBase
     {
 		[Test]
+#if XNA
+        [Ignore]
+#endif
```
tomspilman: I am a little concerned about skipping this test as I don't know why we would be.  Hopefully we can work this out later.

Jjagg: I think I did this because the XNA docs give no guarantees regarding the order in which Draw is called on components when DrawOrder is the same for components. But since the test passes anyway it doesn't have to be ignored.

nkast:
> Ensured that non-premultiplied SpriteFont works correctly for compressed and non-compressed font textures.

@KonajuGames , do you remember why you added this line? 
```
            // Fill the output with non-premultiplied black
            output.ReplaceColor(Color.Transparent, Color.Black);
```
https://github.com/KonajuGames/MonoGame/blob/885bc4631b093dfe415ad6e6d8afe5f4980324a9/MonoGame.Framework.Content.Pipeline/Graphics/Font/GlyphPacker.cs#L62-L63

I think it might be the cause of issue #5491.

KonajuGames:
> I think it might be the cause of issue #5491.

I tried removing that line and the black edges still occur.  Investigating further.

KonajuGames:
No. That was it.  My test setup was using the installed content pipeline rather than the one built from source.  Verifying it doesn't adversely affect anything else.

KonajuGames:
PR #5509 submitted.

PR #5400, opened 2017-01-06
Use lock-free ConcurrentQueue<>

nkast:
This is a light version of #5158.
Improves the performance of Touch/Mouse event processing on Windows.

### DIFF
--- a/MonoGame.Framework/Input/Touch/TouchQueue.cs
+++ b/MonoGame.Framework/Input/Touch/TouchQueue.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+﻿using System.Collections.Concurrent;
 
 namespace Microsoft.Xna.Framework.Input.Touch
 {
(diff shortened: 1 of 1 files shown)


tomspilman:
@nkast - Does this fully replace #5158 ?

What kept me from merging #5158 was the general complexity of the change and the concept of making `TouchPanelState` platform specific.

nkast:
>Does this fully replace #5158 ?

Fictionally they are the same. The thing with  #5158 was that it allows for further changes like push sync code to GetState() and do move processing in the event. All the above make sense when events come from another thread, like  #5333. 

> What kept me from merging #5158 was the general complexity of the change and the concept of making TouchPanelState platform specific.

That's understandable, I will drop #5158 if you don't like it. Anyways, I probably got carried away, I'll come back to it when'if I have changes that make actual use of such changes.

tomspilman:
Thanks @nkast ...  merging!

Issue #5347, opened 2016-12-02
Specialize SpriteBatch Draw Methods


tomspilman:
While @nkast has done some great work with his low level optimizations there is still more to be done.

For example all of these functions in `SpriteBatch`:
```
void Draw(Texture2D texture, Vector2 position, Color color);
void Draw(Texture2D texture, Rectangle destinationRectangle, Color color);
void Draw (Texture2D texture, Vector2 position, Rectangle? sourceRectangle, Color color);
void Draw (Texture2D texture, Rectangle destinationRectangle, Rectangle? sourceRectangle, Color color);
```

All redirect to the same `DrawInternal` method which does a common set of optimizations.  We do this mainly to make the code easier to read and maintain as well as to void bugs.  This however comes at the expensive of performance.

Take `void Draw(Texture2D texture, Vector2 position, Color color)` for example.  This draw doesn't need rotation, doesn't need flipping of sprites, doesn't need UV calculations.  If this function created its own `SpriteBatchItem` internally it could do so much more efficiently than `DrawInternal` does now.

Same goes for the other specializations of `Draw` and `DrawFont` in `SpriteBatch`.  This matters when games are drawing 1000s for sprites per frame.

While this does mean a lot more duplicate code... it also means much more performance.  We can deal with the issue of duplicate code via more unit tests.  Performance here is more important.

nkast:
I agree with all of the above. Actually I've done a few test recently to do exactly that. I pushed two more commits into #5267 the other day, the one called `Refactor SpriteBatchItem.Set()` is towards this goal. The performance improvements were negligible, the actual goal was to move calculations out of `.Set()`. 

> While this does mean a lot more duplicate code... [...]. We can deal with the issue of duplicate code via more unit tests. 

One pattern I notices was that we can redirect all `Draw(Texture2D texture, Vector2 position, ...)` into the similar `Draw(Texture2D texture, Rectangle destinationRectangle, ...)` without too much overhead. 

Another thing that became obvious during this fun process :dizzy_face: was the need for a standardized benchmark that can run a number of tests and compare them against the baseline.

tomspilman:
> One pattern I notices was that we can redirect all `Draw(Texture2D texture, Vector2 position, ...)` into the similar `Draw(Texture2D texture, Rectangle destinationRectangle, ...)` without too much overhead.

[...]

PR #5401, opened 2017-01-06
Inline/Optimize SpriteBatch.DrawInternal() & SpriteFont.DrawInto()

nkast:
Inline SpriteBatch.DrawInternal() & SpriteFont.DrawInto().
Remove temporary Variables and unnecessary Computations.
#5347

Method|Develop|this PR|difference
------------ | ------------- | ------------- | -------------
Draw(Texture2D, Vector2, Color)|0.18652|0.14050|75.3%
Draw(Texture2D, Rectangle, Color)|0.21533|0.12461|57.9%
Draw(Texture2D, Vector2, Rectangle, Color)|0.23807|0.17247|72.4%
Draw(Texture2D, Rectangle, Rectangle, Color)|0.25178|0.15590|61.9%
Draw(Texture2D, Vector2, Rectangle?, Color, float, Vector2, Vector2, SpriteEffects, float)|0.31193|0.29355|94.1%
Draw(Texture2D, Rectangle, Rectangle?, Color, float, Vector2, SpriteEffects, float)|0.31487|0.28922|91.9%
DrawString(SpriteFont, string, Vector2, Color)|0.48899|0.37718|77.1%
DrawString(SpriteFont, string, Vector2, Color, float, Vector2, Vector2, SpriteEffects, float)|0.69743|0.40235|57.7%


MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -461,7 +479,56 @@ internal void FlushIfNeeded()
...
 		public void Draw (Texture2D texture, Vector2 position, Rectangle? sourceRectangle, Color color)
 		{
-			Draw (texture, position, sourceRectangle, color, 0f, Vector2.Zero, 1f, SpriteEffects.None, 0f);
+			CheckValid(texture);
+            
+			var item = _batcher.CreateBatchItem();
+			item.Texture = texture;
+            
+            // set SortKey based on SpriteSortMode.
+            switch (_sortMode)
```
tomspilman: Wouldn't this be easier to read (and maybe faster) as:
```
item.SortKey = _sortMode == SpriteSortMode.Texture ? texture.SortingKey : 0;
```

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -461,7 +479,56 @@ internal void FlushIfNeeded()
...
+                case SpriteSortMode.BackToFront:
+                    item.SortKey = 0;
+                    break;
+            }
+
+            Vector2 size;
+
+            if (sourceRectangle.HasValue)
+            {
+                _tempRect = sourceRectangle.Value;
```
tomspilman: I've always disliked this `_tempRect` stuff...  are we sure using this temporary really saves any performance?

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -508,10 +682,93 @@ public void Draw(Texture2D texture, Rectangle destinationRectangle, Color color)
...
+                var p = offset;                
+                p.X += currentGlyph.Cropping.X;                
+                p.Y += currentGlyph.Cropping.Y;
+                p += position;
+
+                var item = _batcher.CreateBatchItem();
+                item.Texture = spriteFont.Texture;
+            
+                // set SortKey based on SpriteSortMode.
+                switch (_sortMode)
```
tomspilman: This test can go outside the loop entirely.

tomspilman:
I checked and it seems there is some missing code coverage in the `SpriteBatch` tests.  As part of this PR we should look to fix that to ensure we don't break any functionality.

MonoGame.Framework/Graphics/SpriteBatch.cs
nkast: Missed that one...

MonoGame.Framework/Graphics/SpriteBatch.cs
nkast: Missed that one...

MonoGame.Framework/Graphics/SpriteBatch.cs
nkast: I don't think so. I am going to remove it and use a local temp.
Although  if I do the same for _texCoordTL & _texCoordBR we might get a small hit because C# always initialize variables to their default value.

nkast:
> I checked and it seems there is some missing code coverage in the SpriteBatch tests. As part of this PR we should look to fix that to ensure we don't break any functionality.

Yeah, the drawString() with StringBuilder I imagine. How should we tackle this one?Create a copy of every test in SpriteFontTest.cs?  `public void Rotated_sb(), public void Scaled_sb()` etc?

Those methods are pretty similar,  should I redirect them to `Internal DrawString(SpriteFont, CharacterSource   ,...)` or is it too early to talk about recombining code?

tomspilman:
> How should we tackle this one?Create a copy of every test in SpriteFontTest.cs? 

Yeah... copy the most relevant tests and modify it to be sure each draw method is tested.

> Those methods are pretty similar, should I redirect them

Combining code means we're back to sacrificing a little performance.  In particular the non-rotated versions of the functions which are much much cheaper.

The `StringBuilder` one in particular is bad to share with the `string` versions because the whole goal of it is to avoid an allocation of converting from `StringBuilder` to `string`.

I think we need to depend on the unit tests to ensure we don't make any mistakes in the duplicated code to get the most performance we can out of it.

tomspilman:
@nkast - Also noticed looking at the `DrawString` tests they don't seem to verify that newlines or carriage returns work correctly.  We should use some of the same strings from the measure string tests:

https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/Graphics/SpriteFontTest.cs#L124

To verify that text drawing is working correctly.

nkast:
@tomspilman Ready.

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -253,23 +252,86 @@ void CheckValid(SpriteFont spriteFont, StringBuilder text)
...
+                    item.SortKey = -layerDepth;
+                    break;
+            }
+                        
+            origin = origin * scale;
+            
+            float w, h;
+            if (sourceRectangle.HasValue)
+            {
+				var srcRect = sourceRectangle.Value;
```
MichaelDePiazzi: Since this is all about micro-optimisations, I thought I'd mention that `GetValueOrDefault()` is actually slightly faster than `Value` - https://ericlippert.com/2012/12/20/nullable-micro-optimizations-part-one/

KonajuGames:
Were @tomspilman's comments on Jan 11 addressed?

KonajuGames:
I would like to merge this, but want to make sure the raised comments were addressed first.

nkast:
@KonajuGames 
> Were @tomspilman's comments on Jan 11 addressed?

Commit [DrawString tests ad80453](https://github.com/MonoGame/MonoGame/pull/5401/commits/ad80453829c50e70c5a089166153964b12301543) improve code coverage of Draw()/DrawString(), including newlines.
ex: https://github.com/nkast/MonoGame/blob/ad80453829c50e70c5a089166153964b12301543/Test/Framework/Graphics/SpriteFontTest.cs#L495-L503

@KonajuGames , @tomspilman 
Would you like to move the Tests on another PR or is it good this way?

KonajuGames:
I think the tests here are fine as they are related. Merging. Thanks @nkast

nkast:
Thanks @KonajuGames

tomspilman:
@nkast @MichaelDePiazzi 

So reviewing the current state of `SpriteBatch` I think the next optimization we could do would be changing `SpriteBatchItem` to a structure.  This would require `public unsafe SpriteBatchItem* CreateBatchItem()`.  It would also mean sorting would change a bit...  we would sort an index array that is then used to read from the `_batchItemList` in the correct order.

The main benefit of all this is that `_batchItemList` is a contiguous chunk of memory improving performance via cache coherency.

The other optimization I already did for PS4... just have to work out how to support it in the public repo.  Basically using a pre-initialized `IndexBuffer` along with `DrawUserIndexedPrimitives`.  This requires a new overload of `DrawUserIndexedPrimitives` that doesn't take an index array.  The main benefit of this is avoiding the memory copy of index data to the GPU on each draw.

Any other thoughts?

nkast:
> I think the next optimization we could do would be changing SpriteBatchItem to a structure. 
> The main benefit of all this is that _batchItemList is a contiguous chunk of memory improving performance via cache coherency.

That would be preferred. Having thousand of small objects is a waste of memory, put a pressure on the GC to iterate them and also are not packed close together in memory. Still I don't know how we can implement this, especially sorting. My previous attempts didn't yield good results, I have to try this again. 
We already discussed the possibility where CreateBatchItem() returns a struct that's just an index to an array and a pointer to that array.
Some more thoughts on this: When we draw with `Sorting`, we already lose the benefit of cache coherency. When we draw with `Deferred` it's probably faster to write directly into the Vertex array. SpriteBatcher & SpriteBatchItem already provide a good abstraction/isolation, so we could try a strategy pattern with abstract classes or Interfaces and have concrete SpriteBatchers specialized for each `SortMode`.

> Basically using a pre-initialized IndexBuffer along with DrawUserIndexedPrimitives. This requires a new overload of DrawUserIndexedPrimitives that doesn't take an index array. 

I've done something similar for DX, instead of changing DrawUserIndexedPrimitives() I have a VertexBuffer inside SpriteBatcher, the code is similar to what DrawUserIndexedPrimitives() is doing internally. (Additionally, I map the buffer and write directly to the mapped memory, saving one memcopy, and also batch the entire buffer at once instead of flushing on every texture change).
The problem is that on OpenGL the DrawUserIndexedPrimitives path is faster that going through a VertexBuffer & IndexBuffer. My attempts to unify those using partial classes and PlatformX() methods results in code that's more complicated than I would like and not that good in performance.

Some more ideas for SpriteBatch:
-Keep Glyphs in an array and use an index to lookup and access Glyph data instead of Dictionary<char, Glyph> and .TryGetValue(c, out currentGlyph). This will save us the time spend to copy the `Glyph` struct from the dictionary and probably make Dictionary lookup faster. Other data structures might perform better in that scenario (SortedList<char,int>, HashTable, or a Binary search operating directly on the array).
-Cache texture coordinates TL/BR in Glyph.
-Batch up to 16 textures in one Flush, combined with a new technique in SpriteEffect. This one makes Draw ~x5.5 times faster on DX and ~x7 times faster on GL when you draw different textures with Deferred. Other SortModes including SortMode.Texture will see major boost as well. The catch is that I think it will  perform terribly on GPU with SM2.0 and that it will override all texture slots after .End(). Also I don't know what happens when you push overlapping primitives with the same depth, does GPU's streams  somehow take into account the order of each primitives on IndexBuffer?

tomspilman:
>  Still I don't know how we can implement this, especially sorting. 

You don't sort the `SpriteBatchItem` array.  You keep a separate array which contains both an `int` index and the sort key.  That should be a small struct of just 8 bytes.  That is the thing you sort.  You then use that array to index into the `SpriteBatchItem` array.

> CreateBatchItem() returns a struct that's just an index to an array and a pointer to that array.

Actually `CreateBatchItem()` should just return an unsafe pointer to the item to fill in.  This makes the code easier to follow and write and also eliminates extra bounds checking on the array.

> When we draw with `Sorting`, we already lose the benefit of cache coherency.

Not at all.  The struct array is way more cache friendly even if you are jumping around in that array.

Compared to an array of object pointers... you have the pointer array in one location in memory and each object possibly different places in memory.  Each could be so far away it takes one cache hit per item.  The second your game releases memory you make gaps which the allocator could try to fill when you allocate a new `SpriteBatchItem`.  Each could be dozens of MBs away from each other.  It is total chaos when it comes to the cache.

The it vastly better with the struct array.

> When we draw with `Deferred` it's probably faster to write directly into the Vertex buffer.

Except we would have to flush if we run out of space in the vertex array... breaking from XNA behavior.

IMO just making things cache friendly is enough here.

willmotil:
> -Keep Glyphs in an array and use an index to lookup and access Glyph data instead of Dictionary<char, Glyph> and .TryGetValue(c, out currentGlyph).

I thought about trying to do this but how does this work under localization remember chars are 2 bytes not 1.  An 'e' in English is i doubt the same as 'e' in japanese '絵'. To say how are you going to assign it as a index to be looked up directly without a big array ? 

Can you simply say var charindex = byte(text[i]); then glyph[charindex] or use a pointer to it. 
What if the greek 'e' ends up having its defining byte in the higher order byte and the low order is just 0. 

Im not familiar enough with the standard to say that all letter e's are always in the lower byte but maybe they are. Just saying this is something that gave me pause when i was entertaining the thought.

KonajuGames:
.NET uses 16-bit chars throughout. This covers the first 65536 character
codes in the Unicode range. 'e' is one character code. Japanese 'e' is a
different character code, as is the Greek 'e'.

Using the character code to index an array is a bad idea from the view of
memory use. Are you going to have an array with 65536 elements? And of
those elements, how many are going to be used? You would need at least as
many elements as the highest character code you gave a glyph for, but this
could have hundreds or thousands of empty array slots before it.

willmotil:
I dunno but i think dictionary in this case is already mapping via hashcode from the underlying chars base.object so i think in this case dictionary lookups would in fact be O(1) if i said that right.

theZMan:
>.NET uses 16-bit chars throughout. This covers the first 65536 character
>codes in the Unicode range.

.Net strings are UTF-16 which doesn't mean a 16 bit char. Some characters cound take 2 16 bit values to represent. See https://en.wikipedia.org/wiki/UTF-16#Description for the painful truth. Its unlikely many people will use encodings outside the BMP (https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane) but you can't assume 65536 characters.

KonajuGames:
I wasn't certain if it was UTF-16 or UCS-2. UTF-16 does indeed allow
multi-word characters. It is not often that character codes above the BMP
are used, and if they were our SpriteFont, being based on XNA's SpriteFont,
would need to change to support the multi-word characters. The XNB format
for SpriteFont only supports the BMP.

willmotil:
Ya with that chart looking at it with a glance, basically affirms the fear were speaking of.
From 0f to 0f in hex is the very first line it's all you get with 8 bits which is var charindex = byte(text[i]);
What about the rest that dictionary handles as it is now.

MichaelDePiazzi:
> I think the next optimization we could do would be changing SpriteBatchItem to a structure.

This definitely sounds worth trying.

> -Keep Glyphs in an array and use an index to lookup and access Glyph data instead of Dictionary<char, Glyph> and .TryGetValue(c, out currentGlyph). This will save us the time spend to copy the Glyph struct from the dictionary and probably make Dictionary lookup faster. Other data structures might perform better in that scenario (SortedList<char,int>, HashTable, or a Binary search operating directly on the array).
> -Cache texture coordinates TL/BR in Glyph.

I think spending some more time on optimising DrawString is definitely worthwhile. Text drawing can still be one of the more costly areas in my game. I've personally been meaning to spend some time on this for a while. So if I come up with something, I'll be sure to submit a PR.

theZMan:
>Other data structures might perform better in that scenario (SortedList<char,int>, 
>HashTable, or a Binary search operating directly on the array).
Dictionary<T> is the generic replacement for HashTable - I hope we never consider using that. I dont know why the framework didn't deprecate it years ago.

There's no way that a SortedList or any kind of binary search is going to be faster than a hash table. A good hash function give close to O(1) time for accessing and since we know the input set at load time we should be able to preallocate the dictionary efficiently too (though it happens at Load so who cares ;-)) I suspect using Char as the Key is also super efficient - the default hash function on that should be trivial.

Really we need to profile here rather than just guessing what needs doing.

willmotil:
I think their might even be a question of is it worth it to ditch dictionary because in my humble opinion it would have to be a big gain for losing functionality as well as a possible breaking change to existing apps.

theZMan:
The prime number 'magic' is deciding how many buckets to use based on the number of items you want to store in the collection. That's how it tries to keep the number of items per bucket low.

Otherwise there is no magic, when there is a collision it does a quick loop to find the right one. That's still 'close to' O(1) unless your hashing algorithm is putting far too many in 1 bucket. If you hash function is return 0 then you will find all your accesses become O(N) because everything is in 1 bucket.

And I don't mean use profiling to show that it speeds up I mean use profiling to show what the current hot spots are for a realistic usage in a real game. It pretty easy to make fake examples that will show issues in specific places and then massive gains by fixing those issues but then it doesn't translate into real world gains because it wasn't a real problem.

willmotil:
Well in this case specifically were talking about text and the glyphs of a font so im sure if you directly map to the low byte you can get a slight speed up in specific or general cases in a real application that simply has a lot of text. The only point im trying to make, is that there is a price for dropping dictionary. Not only in possibly breaking changes but also for future improvements or functionality as in the case of, emoji's, word wrapping to rectangle areas, scientific notation. That's just off the top of my head. The alternative to this with direct array mapping is to use a very large array per font. Which would probably end up taking you back to dictionary.

PR #5403, opened 2017-01-08
Optional parameters for spriteBatch.DrawString

evgenidb:
Made some of the parameters in DrawString optional, similar to the Draw method. Now you can use named parameters, so you don't have to enter all arguments if you want only to rotate the string or set the layer.

SpriteFont, text, position, and color are still required though.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -535,7 +535,7 @@ namespace Microsoft.Xna.Framework.Graphics
 			var scaleVec = new Vector2(scale, scale);
             var source = new SpriteFont.CharacterSource(text);
             spriteFont.DrawInto(this, ref source, position, color, rotation, origin, scaleVec, effects, layerDepth);
-		}
+        }
 
         /// <summary>
         /// Submit a text string of sprites for drawing in the current batch.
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @evgenidb !

@nkast - Was just thinking... did we ever measure if optional parameters like this affect performance?

KonajuGames:
I believe the compiler simply expands it to the full method call with all parameters.  There's no extra runtime overhead for these calls, just the cost of passing all parameters.

evgenidb:
There would be a performance drop for sure because of the two ifs and the unboxing of origin and scale. But the performance penalty should be smaller than the Draw with optional parameters.

Also it seems that it will be fine if I both the original version and the one with default parameters are present, so I'll update the pull request. I deleted the old one because I thought that VS will get confused but it doesn't: it uses the original version and goes into the default parameters method only if it really needs to.

tomspilman:
> unboxing of origin and scale

Actually nullables are not boxed... they use the `Nullable<>` generic to avoid boxing.

nkast:
>> @nkast - Was just thinking... did we ever measure if optional parameters like this affect performance?

> I believe the compiler simply expands it to the full method call with all parameters. There's no extra runtime overhead for these calls, just the cost of passing all parameters.

Never measure it but I imagine it works like @KonajuGames described it.

@evgenidb I understand you follow the spirit of SpriteBatch.Begin(...) where null = (use default value) but all those checks add overhead. Why don't we change the current DrawString() into
```DrawString(SpriteFont spriteFont, string text, Vector2 position, Color color, float rotation =0, Vector2 origin = Vector2.Zero, Vector2 scale = Vector2.One, SpriteEffects effects = SpriteEffects.None, float layerDepth = 0)```

XNA 4 was released about the same time as C# 4.0 and didn't make use of optional arguments. The trick with nulls works for Begin(...) and states (classes) but doesn't work so good with Draw/DrawString and structs/values.

evgenidb:
@nkast Unfortunately `Vector2 origin = Vector2.Zero` won't work because `Vector2.Zero` is not a compile time constant (e.g. null, bool, int, float, etc, but not structs or classes). The only other way I found it to work is `Vector2 origin = new Vector2()`, but I'm not sure this is better. It is even worse for `Vector2 scale`: you can't make the unit vector (1, 1).

PR #5405, opened 2017-01-09
Typos in MD addin excludes

Jjagg:
You'd think MD would complain about this.


KonajuGames:
Not surprised.  XML is usually designed to ignore nodes it doesn't recognize.

tomspilman:
Thanks @Jjagg !

Issue #2707, opened 2014-06-11
Missing classes in Microsoft.Framework.Design


jamesford42:
For reference:
http://msdn.microsoft.com/en-us/library/Microsoft.Xna.Framework.Design(v=xnagamestudio.40).aspx

We currently have Vector2/3/4TypeConverter (though whether or not they behave exactly like XNA has not been thoroughly verified).

There is also a weird StringToColorConverter currently in the Microsoft.Xna.Framework.Content.Pipeline.Builder.Converters namespace which should 'probably' be the Framework.Design.ColorConverter.

flibitijibibo:
Spent the day writing in the rest of the TypeConverters. The result:

https://github.com/flibitijibibo/FNA/tree/master/src/Design

Some general notes can be found here:

https://github.com/flibitijibibo/FNA/issues/195#issuecomment-159064606

One MonoGame-specific note to add is that, apparently, the TypeConverters do not actually work for anything but strings. I dunno why all this is here just for strings, but in XNA4 that's all they do... hence the base MathTypeHelper class which does just about everything generically except for the actual conversion operations.

If you pull in these classes you'll probably want to reintroduce stuff like the VectorConversions so that it's actually useful for the content pipeline and other similar projects.

PR #5408, opened 2017-01-09
Implement MeshBuilder v2

Jjagg:
Hijacked this from @mattiascibien in #4078

While this improves the implementation and adds a bunch of tests, it seems there are still issues.
I tried the XNA [Generated Geometry Sample](http://xbox.create.msdn.com/en-US/education/catalog/sample/generated_geometry) and got nothing for the terrain yet :(

Note: In f12a882 I added the struct constraint to VertexChannel<T>. That breaks with the XNA API, but I think it makes more sense and it prevents issues when trying to use reference types here (see the commit description, XNA can throw a NullReferenceException). Since it's in a seperate commit it would be easy to revert if people don't agree with me.

EDIT: I reverted this change for now because OpenAssimpImporter uses a VertexChannel with a reference type


MonoGame.Framework.Content.Pipeline/Graphics/MeshHelper.cs
```diff
@@ -373,16 +374,75 @@ public static IList<BoneContent> FlattenSkeleton(BoneContent skeleton)
...
+                    for (var index = 0; index < geometry.Indices.Count; index++)
+                    {
+                        if (geometry.Indices[index] == i)
+                            geometry.Indices[index] = j;
+                    }
+
+                    verts.RemoveAt(i);
+                    break;
+                }
+            }
```
Jjagg: This is waaaaaaay too slow. Complexity is O(V^2 * P) were V is the number of vertices and P is the number of unique positions. Since we only check exact equality we can probably hash and make it linear :D Let me try that!

Jjagg: We'd still have to go throught the indices, but we can sort those beforehand to get access in our main loop in linear time.

Jjagg: We can do better for MergeDuplicatePositions too if we use some kind of space partitioning solution.

Jjagg:
I tried the xnb built with XNA and it doesn't show up in MonoGame either... So that doesn't say much about the MeshBuilder, but it shows a problem when rendering models.

MonoGame.Framework.Content.Pipeline/Graphics/MeshHelper.cs
nkast: > This is waaaaaaay too slow. Complexity is O(V^2 * P) were V is the number of vertices and P is the number of unique positions.

Shouldn't this check be in AddVertex() and return the merged index immediately?

> We can do better for MergeDuplicatePositions too if we use some kind of space partitioning solution. 

That would be interesting!
I could dig some old code to find if anything can  be used here.

nkast:
So, the XNA's XNB doesn't render on MG? 
Is that a bug with MG?

MonoGame.Framework.Content.Pipeline/Graphics/MeshHelper.cs
Jjagg: > Shouldn't this check be in AddVertex() and return the merged index immediately?

XNA docs [say](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.content.pipeline.graphics.meshbuilder.aspx#Anchor_1) merging vertices happens in FinishMesh. The docs [also say](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.content.pipeline.graphics.meshbuilder.mergeduplicatepositions.aspx#Anchor_1) merging positions is checked in CreatePositions, but I tested that and it's not. I already have a version that hashes the vertices implemented, just testing a bit and then I'll push it.
They're not merged until the FinishMesh call too.

>  I could dig some old code to find if anything can be used here.

That would be great!

Jjagg:
Yeah. I think it's in the rendering, the model looks the same when I check its properties in the debugger a bit and they seem to be the same in MG and XNA (but I didn't check thoroughly yet).

Jjagg:
Sorry that the IndexUpdateList code is so hard to read. It isn't easy coming up with good variable names  when you're working with indices into arrays that contain indices into array of values which are the vertex indices.

MonoGame.Framework.Content.Pipeline/Graphics/MeshHelper.cs
nkast: It's much easier to check and remove duplicate positions right in the CreatePosition(). Who knows, they have to implement it in MeshHelper anyways, so why having duplicate code.
As for merging vertices, does it discard repeated triangles or something? I really don't get it.

> That would be great!

I don't have anything that works with tolerance/distance though... give me some time.

Jjagg:
This is hard to reason about :p There are still bugs in the implementation...

Jjagg:
I think the implementation is alright now, but I have an issue when building the Generated Geometry content with the pipeline tool. When it's done building the content it seems to not have noticed that the content is built and keeps waiting for it. When I break the stack trace is all native code, so I don't know how to debug this. I'll see if using MGCB directly works, but my guess is it will.

Jjagg:
The issue with the mesh not being visible might be because of our Vector4Converter not behaving correctly... :/ The heightmap data is lost in the conversion because the Vector4 -> float conversion takes the alpha channel (which is just 1 everywhere). I'll fix that and see if things render correctly.

I noticed XNA actually has a Vector4Converter in the API and then stumbled on #2707. I'm just going to fix the issue at hand right now. Though we should look at getting the whole Xna.Framework.Design namespace implemented and tested.

KonajuGames:
> The issue with the mesh not being visible might be because of our Vector4Converter not behaving correctly... :/ The heightmap data is lost in the conversion because the Vector4 -> float conversion takes the alpha channel (which is just 1 everywhere).

I keep forgetting the whole Design namespace is there.  For my texture content processing, I had implemented a [Vector4Converter](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Utilities/Vector4Converter.cs) in the content pipeline (that uses the red channel like XNA for float conversions).  This would likely be replaced entirely by the more heavyweight TypeConverters in the Design namespace.

Jjagg:
image
Yes!! That actually fixed it! :D This is great! We can conclude a couple things from this:

1. MeshBuilder is working! We should probably add some more tests for MeshHelper.MergeVertices and try to optimize MergePositions, but performance is acceptable right now and it seems to be working (file size tells me vertices got merged as much as in XNA and things look the same for the GeneratedGeometry sampler at least).
2. PixelBitmapContent conversion is not tested as good as I thought it was! If we implement the converters from the design namespace and properly test those things should be fine. We have some tests in place for PixelBitmapContent, but they apparently don't cover this.
3. We don't properly load XNA built model xnb's. Not sure how big an issue this is, but we should at least let users know if they run into issues they should rebuild with the MG pipeline. And of course it would be nice to get this fixed...

Jjagg:
> For my texture content processing, I had implemented my own Vector4Converter in the content pipeline (that uses the red channel like XNA for float conversions).

Oh, right. This is that issue we ran into the other day. I guess it is covered by tests. The [fix](https://github.com/MonoGame/MonoGame/pull/5398) was merged two days ago so I don't have that locally yet.

KonajuGames:
> PixelBitmapContent conversion is not tested as good as I thought it was!

I'm keen to know more on this.  I thought I had it correct, but perhaps some channels are mixed up, but somehow give the correct results in our tests?

Jjagg:
Rebased to take care of the Vector4Converter issue. This is ready for review!

Jjagg:
The Pipeline Tool does **not** get stuck when I just launch the produced executable instead of launching it from Visual Studio, so that's not something we should worry about.

KonajuGames:
> The Pipeline Tool does not get stuck when I just launch the produced executable instead of launching it from Visual Studio, so that's not something we should worry about.

Is this the Pipeline tool waiting for the MGCB process to end, or the Pipeline tool not terminating properly when closing it?  I've seen the second case (the window closes but the process does not terminate, so Visual Studio does not come out of debugging mode).

Jjagg:
I can't seem to reproduce it anymore... Maybe I was just a bit impatient and didn't wait long enough for the build to finish (it takes a couple seconds).

KonajuGames:
Merging this. Thanks @Jjagg

PR #5410, opened 2017-01-14
Remove OpenTK from Android, iOS, tvOS and MacOS

dellis1972:
This branch removes the last uses of OpenTK. 

A quick Poll. Should be switch the MacOS platform over to use the SDL2 stuff we did?


cra0zy:
> A quick Poll. Should be switch the MacOS platform over to use the SDL2 stuff we did?

I vote yes.

Jjagg:
> Should be switch the MacOS platform over to use the SDL2 stuff we did?

MacOS is using Xamarin.Mac for managing the window, right? So switching to SDL would mean one less windowing backend to maintain?

mrhelmut:
DesktopGL is totally fine on MacOS to me.

dellis1972:
@prollin This might be of interest to you. Could use some feedback if thats ok :)

dellis1972:
Current problem. Textures/Fonts are not showing on iOS. If anyone can spot the problem let me know :)

dellis1972:
Think I found it :)

dellis1972:
ok MacOS Still fails. But iOS and Android should be ready for testing is anyone has time.

cra0zy:
Good news: I got this and a sample android project compiled from Linux and installed on my device.
Bad news: It crashes on start :(

I am not sure if it's crashing because I compiled it from Linux or because this PR was bad.

Jjagg:
Can we get this merged in a new branch maybe? That makes it easier track progress and easier for multiple people to work on it.

Jjagg:
We need to double check the entry points. For example none of the GL ES versions support any variant of glGetTexImage, but loading that is not in a try catch block in OpenGL.LoadEntryPoints
https://github.com/dellis1972/MonoGame/blob/d87e3794028af52a7500c4e5319bf2d456e83a22/MonoGame.Framework/Graphics/OpenGL.cs#L1224

MonoGame.Framework/Android/MonoGameAndroidGameView.cs
```diff
@@ -4,346 +4,919 @@
...
+
+            curUpdateTime = DateTime.Now;
+            if (prevUpdateTime.Ticks != 0) {
+                var t = (curUpdateTime - prevUpdateTime).TotalMilliseconds;
+                updateEventArgs.Time = t < 0 ? 0 : t;
+            }
+            try {
+                UpdateFrameInternal (updateEventArgs);
+            } catch (Content.ContentLoadException) {
+                // ignore it..
```
dellis1972: @Jjagg I updated the code to throw the exceptions on the UIThread of the Activity. That in theory should help.

dellis1972:
@Jjagg if you want to help on this one , you can add my fork as a 'upstream' and pull into a branch in your repo, then submit PR's to my fork.

Jjagg:
I'll do that. Though I can only do Android :/

dellis1972:
what the heck.. now I have 127 commits!

tomspilman:
what-did-you-do

KonajuGames:
Same thing happened to me with my VideoPlayer PR.  Ended recreating it from scratch.  Which you likely won't want to do.

dellis1972:
I really don't want to re-create from scratch.. I guess I could squash all my commits into one and cherry-pick the whole lot to a new branch

dellis1972:
HA! Fixed it , F$%K you git, I own you! :P

KonajuGames:
How?! For future reference when it happens again.

MonoGame.Framework/Audio/OpenAL.cs
```diff
@@ -175,7 +175,13 @@ public enum EfxEffectType
 
     public class AL
     {
+#if IOS || MONOMAC
+        public const string NativeLibName = "/System/Library/Frameworks/OpenAL.framework/OpenAL";
+#elif ANDROID
+        public const string NativeLibName = "libopenal32.so";
+#else
         public const string NativeLibName = "soft_oal.dll";
+#endif
```
dellis1972: we should probably split these out into partial classes. Get rid of these pesky #if blocks

dellis1972:
@KonajuGames I "think' I did a 

'git pull origin <branch>'

which got it all confused and resulted in a Merge conflict for something that was fixed already (in the previous commit). As a result I had to commit an empty commit to fix the merge conflict. After that I had 127 commits. 

I fixed it by doing a hard reset to the commit I knew was ok. 

'git reset --hard <hash>'

then doing a force push to my branch

'git push --force origin <branch>'

generally you want to avoid force pushing (unless your name is Vader) but in some cases its just necessary :)

Jjagg:
> generally you want to avoid force pushing

It's fine if you know no one else is interacting with your branch :) I do it all the time for rebasing and for amending commits that make me look stupid!

cra0zy:
@dellis1972 Why couldn't you do a `git rebase upstream/develop`?

dellis1972:
@cra0zy I did , but the wasn't what caused the problem.

jakepoz:
I notice that when compiling the Android version using the latest Android 6.0 Xamarin there is an ambigous type compile error on

```
        public virtual Size Size {
vs
        public virtual Android.Util.Size Size {
```

thats in MonogameAndroidGameView

jakepoz:
I just built a popular game of our with this PR, sending it to our beta testers this weekend and should hear back by Monday.

@dellis1972 Is it expected that I had to wrap a piece of Android UI code called from our game with a RunOnUiThread that was not needed earlier?

dellis1972:
That is expected. The old system created the GL context on the UI thread,
and ran the game timer on a background thread. This meant that we had to
use a SyncContext to Post the Draw and Update calls to the UI thread. This
causes a 20-200ms delay per frame.. not good.

The new system runs everything on a background thread. So if you do any UI
interaction you will need to use RunOnUiThread.

jakepoz:
Got it :)

Ok, the biggest bug we see is that rotation doesn't work right for games that support both Portrait and Landscape mode. 

image

dellis1972:
@jakepoz a small repo project would be nice if you have the time? also which device is that on? I know rotation is a pain especially on kindle devices (because amazon..:/ )

dellis1972:
Also as mentioned on another thread. We cannot remove the MacOS stuff in favour of SDL2. Because if we want to support window embedding we need to have a native system. SDL2 does not support embedding into other windows afaik :(

cra0zy:
https://wiki.libsdl.org/SDL_CreateWindowFrom

dellis1972:
@cra0zy interesting. We'd need to figure out how to use that on all the desktop platforms. Then figure out an API which would allow people to easily use it. as well as possibly figure out a way to use it on Mobile platforms too (not using SDL but say Xamarin.Forms)

jakepoz:
[RotationExampleBug.zip](https://github.com/MonoGame/MonoGame/files/753187/RotationExampleBug.zip)

@dellis1972 Attached is an example project.

Both old and new monogame work fine on landscape:


This pull request produces this result when rotated


Normally we expect this

dellis1972:
@jakepoz this is really odd. I repo'd the issue. But all the Viewport, Backbuffer width/height properties are totally correct and switch between portrait and landscape. 
Must be something to do with aspect ration or scaling or something

jakepoz:
@dellis1972 Hey, we are going to fork this PR on our end and try to fix this rotation bug. Getting rid of OpenTK will also let us apply our fix for some android sound bugs: https://github.com/MonoGame/MonoGame/pull/5443

dellis1972:
ok, if you do fix it, submit a PR to my branch and I can merge it in.

roponator:
We put in the fixes for the rotation render bug & the flaky sound bug.

The problem is that something went seriously sideways with these git branches: if we pull the branch directly from the fork of @dellis1972 it won't event compile. But if we pull the MonoGame master repository 'opentkandroid' branch it works, as if dellis1972's fixes got merged into the main monogame repo.

@dellis1972 is there something you can do about this? I tried doing a pull request on your opentkandroid branch but it says there are more than 180 files changed (in reality only ~6 are), as all old files from previous commits are bad.

Here is my branch:
https://github.com/roponator/MonoGame/tree/opentkandroid

dellis1972:
hey, great job. I'll cherry-pick those commits over to my branch and test
this weekend.

roponator:
@dellis1972 don't test it yet, there are some more bugs in it, we're working on them.

roponator:
@dellis1972 alright so we did a bunch of fixes to this thing & tested it.

Most of the MonoGameAndroidGameView.cs file has been rewritten. The reason for this is that the initial version was failing on some phones so I had to introduce some fixes but the code got really messy when all edge cases were handled. I rewrote the thing so that it's based on a state machine as certain phones go haywire and would trigger "pause" twice in a row without a "resume" between on aws device farm fuzz tests and we have to be sure that pause/resume stuff is fully processed on the worker thread before we call pause/resume again.

This thing now works for both rendering on UI and on the worker thread (you'll obviously need to dispatch UI modifying calls now is using worker thread).

There is one phone that constantly fails the test when using the worker thread for rendering, some motorola droid razr from 2012, it is as if the worker thread causes so much cpu pressure that the UI thread cannot do anything, not even read input or trigger pause/resume. Note that we did not test it hands-on on this phone but only on aws device farm which really stresses it. The phone works for a while but then the UI thread stops responding.

We chose to ignore this one phone since it's old and we tested this on 30+ other phones where it works without any problems.

Here's the link to this repo as this pull request made git go crazy: https://github.com/roponator/MonoGame/tree/opentkandroid

dellis1972:
Hey @roponator

Glad you have it working. Its gonna take a while to figure out those
changes. There are quite a few.
I'm a little concerned about the ```while (true) ``` loops that are in the
code. They will keep the thread spinning while waiting,
not sure that will be the most efficient way to wait for an event.

The previous code used Monitor.Pulse/Wait to allow the threads to sleep
until events occurred.

The OpenAL changes are a bit confusing. not sure how just creating a
wrapper which called the DLLImport fixes the issues.

I'll do a more detailed review asap and see if I can port those changes
over to my branch.

roponator:
There aren't any while loops?? Except for the main game loop. Which file are you looking at?

About the sound bug: here is more info: https://github.com/MonoGame/MonoGame/pull/5443
A tl;dr version is that functions with multiple float params passed by value cause a bug, so pointers to float must be used.

dellis1972:
ah, sorry. GitHub was not showing the right correct diffs.

jakepoz:
Thanks for taking a look @dellis1972 we are hoping to get this solved once
and for all :)

cra0zy:
I can confirm that @roponator branch runs on my device.

dellis1972:
@jakepoz @roponator I have pulled in your changes. You might want to test it out to make sure I've not missed anything.

roponator:
@dellis1972 there is a bug that happened only once I can't reproduce it but I know whats going on: touch panel gets accessed from the worker and UI thread and if can cause an error. So I'm gonna fix that, gonna do a PR on your branch if git won't go berserk again.

@cra0zy thanks for testing it :)

roponator:
This problem will only happen when the game uses the worker thread to run the main loop, not the UI thread.

The problem is a bit annoying: the entire TouchPanelState class would have to be modified (its not Android specific which is why I'm not happy making any big changes to it).

So the bug happened because the UI thread was inserting new events into the TouchPanelState but the game thread (worker) was reading them, without a lock.

This part was easily fixed (queue up input from UI thread and play it back inside main loop).

The problem is that some other functions of TouchPanelState (like display size) get called from the UI thread, so we must do some locking otherwise we can end up with some strange "input isn't working wtf bugs" due to scaling.

As I can see it we end up with 2 possible mutualy exclusive scenarios:
1) lock the hell out of the TouchPanelState, but we get some slight overhead due to locks.
2) assume the user will read touch panel state only on the game main loop (not in his UI thread eg on some UI controll callback). Reading from the UI thread can cause this bug.

I'd pick 1). Any suggestions?

dellis1972:
@roponator interesting. We default to using the UI thread (I think) so  for now  we can leave that bug and fix it in a different PR. I'd rather not include any more changes in this one since its big enough already. Also the TouchPanel is not an isolated bit of code so we'd really need to test that out on all the platforms. 

@tomspilman @KonajuGames what do you think?

roponator:
I think we should create a specialized TouchPanelState for android, one that derives from the original TouchPanelState and overrides all methods.

This way the original is left intact + all android bugs can be fixed.
The downside is that we are mixing "game layer" with "platform layer".

What do you think?

tomspilman:
> I think we should create a specialized TouchPanelState for android

We should avoid this at all costs.  The only reason touch works as well and as consistent as it does today is because I unified it.  Splitting one platform off will inevitably cause that platform to be bugged and inconsistent with the rest.

> lock the hell out of the TouchPanelState, 

I certainly prefer this to any thought of making touch processing platform specific.

However I would argue that there would be almost zero impact to performance.  Very few locks will be needed as there are very few entrypoints to `TouchPanelState` and since touch processing is very infrequent compared to other aspects of games.

So if someone wants to go ahead with this... open a new PR.

KonajuGames:
I would document that the TouchPanel can only be accessed by the developer
from the main thread. Internally we make it thread safe for those functions
that are used by the UI thread.

roponator:
Ok.

@tomspilman is the reason to open a new PR historical clearity instead of putting it in this PR (since that PR won't make sense without this PR)?

I'll get on it.

tomspilman:
> is the reason to open a new PR historical clearity instead of putting it in this PR 

It isn't for history in any way.

It is to avoid gigantic PRs that are difficult to test and understand....  especially we discourage combining a bunch of unrelated issues into a single large PR.


>  (since that PR won't make sense without this PR)?

This PR it doesn't seem to do anything with regards to touch input on Android or any other platform that I can see... am I wrong?

I wouldn't be surprised if the issue you ran into would happen without this PR as well.

roponator:
You're right. The issue would happen also in the current branch if android would run in the worker thread.

Ok, will do a separate PR.

roponator:
I created a PR which adds locking here: #5638

nkast:
We already solved the issue with async touch events on Windows Store by queueing the events #3205, #3205 and then use ConcurrentQueue<> to make it lock free #5400.  
@roponator I quess you can replicate this solution for Android by adding a `List<TouchEvent> _queue;` in AndroidGameWindow.cs.

Although I will propose the following change https://github.com/MonoGame/MonoGame/compare/develop...nkast:tncTouchPanelSplit2
which is an update to #5158.

There are a couple of arguments for making TouchPanel platform specific, 

- Eliminate duplicate code. In #5158 we moved common code from `MetroGamePlatform.cs`,`MetroGameWindow.cs`,`UAPGamePlatform.cs`,`UAPGameWindow.cs` into `TouchPanelState.Async.cs`. For Android we simply switch from `TouchPanelState.Sync.cs` to `TouchPanelState.Async.cs`. 
- Improve quality & performance. We can gradually move the syncing a few steps below all the way down to `.GetState()` which means we can assign real time timestamps, have smoother Gesture processing independent from the framerate, and free CPU for the game loop since all the event processing will be done in the input thread. 
 - Eventually if sync is done right without performance issues we can drop the TouchPanelState.Sync.cs behavior from all platforms.

jakepoz:
I agree with the general idea to move the touch locking fix to a new PR. 

@dellis1972 Do you need any help from us before merging this? Then we can talk about the other fixes with a clear mind :)

roponator:
Alright fixed my git repo, here is the PR for locking touch panel state: #5646

tomspilman:
> Although I will propose the following change https://github.com/MonoGame/MonoGame/compare/develop...nkast:tncTouchPanelSplit2
which is an update to #5158.

@nkast - Why not just make the async behavior the default behavior for all platforms?  I doubt the performance is much different either way.

And would this entirely solve the issue?  Like certainly it would now allow you to query touch/gesture state safely from multiple threads?  Is that not partially what we're trying to fix here?  @roponator ?

roponator:
I'm gonna look at nkasts thingy in detail tomorrow. 

But yes, this is what my PR would solve without splitting it. I lock everything except the GameWindow  (I assume it will never ever change, am I wrong?).

nkast:
I don't think there's a performance issue, sure ConcurrentQueue<> has some overhead compared to Queue<> but it's nowhere close to lock(). What I am worried about is that `tncTouchPanelSplit2` still needs to call 'Window.TouchPanelState.ProcessQueued();` in XGAmePlatform.BeforeUpdate(GameTime gameTime), a change that will affect all platforms. 
In a final solution the sync barrier has to move from the ProcessQueued() into TouchPanel.GetState(), TouchPanel.IsGestureAvailable,TouchPanel.ReadGesture (basically to all methods & properties).

I guess the least controversial approach for a fix right now would be to replicate what we currently do in WinStoreApps, (#3205, #3205, and #5400)

jakepoz:
I think we should get this merged and then open a discussion about the Touch Panel concurrency separately. We just had to ship a game with some nasty workarounds for flakey Sound Effects that would be otherwise fixed by this PR.

cra0zy:
@dellis1972 Your branch still does not work for me, while https://github.com/roponator/MonoGame/tree/opentkandroid works just fine.

dellis1972:
@cra0zy, in what way does it not work? more details please :)

All the required code has been copied across, so I'm not sure what is
missing.

cra0zy:
> @cra0zy, in what way does it not work? more details please :)

A simple app that only has 1 spritefont and draws "Hello World" just black screens.

dellis1972:
Sounds like the content is not loading correctly. That is usually what
causes that on Android. Check the casing of your content, android is case
sensitive.

dellis1972:
that might not be it though, since the there branch works... hmm.

cra0zy:
> android is case sensitive.

Really... you know that I am using Linux as my daily driver...

Fun fact: Did you know that NTFS is case sensitive?

> that might not be it though, since the there branch works... hmm.

Same application, works when compiled with the other branch, does not work with this one.

roponator:
@cra0zy can you upload your version or share the repo link so I can check it out?

cra0zy:
> @cra0zy can you upload your version or share the repo link so I can check it out?

I think you misread something, your branch works just fine, @dellis1972 branch does not work.

That being said if you are asking about the app that I am running, it's just loading 1 spritefont and drawing "Hello World" with it, I can't share the project files because they are a mess and will not compile on Windows/Mac.

cra0zy:
https://gist.github.com/cra0zy/12d685f44d312611aeaf951418f27340

roponator:
@cra0zy I tested your demo ( + created a spriteFont resource) but it works for me, blue screen + text draws on Android.

Could you be using some older version of dellises PR?

cra0zy:
> Could you be using some older version of dellises PR?

Nope.

roponator:
@cra0zy  Hm... which phone did you test it on?

I uploaded my solution here so you can check if it also fails for you: https://github.com/roponator/blackscreen_bug.git

cra0zy:
> @cra0zy Hm... which phone did you test it on?

My phone is s7580 with Lineage OS 14.1.

> I uploaded my solution here so you can check if it also fails for you: https://github.com/roponator/blackscreen_bug.git

Yup, black screens for me.

dellis1972:
@cra0zy can you gist or pastebin the adb logcat output.  it might have some
hints as to why its not working.

cra0zy:
https://gist.github.com/cra0zy/225e236bfe79e5e87d6e763dc33dc257

cra0zy:
I would help with debugging, but the debuger is the only thing I still haven't gotten working with Xamarin Android from Linux....

dellis1972:
Hmm the context is not being created

https://gist.github.com/cra0zy/225e236bfe79e5e87d6e763dc33dc257#file-logcatbin-L428

dellis1972:
That is an invalid enum issue. weird.

cra0zy:
Well I managed to get a black screen with @roponator branch as well, the project that was compiled was much bigger so finding out what caused it is gonna be hard. I can't share the project.

Also the important part of the logcat is: `libEGL  : called unimplemented OpenGL ES API`, I do know few things that can cause that error so I may try to fix it if I am not too lazy :P

roponator:
@cra0zy if you could find out at least which file/line throws that "unimplemented openGL..." error I will try looking into it as much as it's possible without a build.

dellis1972:
@roponator @cra0zy I have finally found the time to rebase this on the latest develop. It could do with some testing though.

cra0zy:
@dellis1972 I think the best option is to merge this, its gonna be easier to just solve any bug that arise in the develop branch.

jakepoz:
Yay, glad to see the end of Open TK :)

mrhelmut:
I'm going to give this PR a test in the next few days on iOS and Android.

dellis1972:
@mrhelmut awesome thanks.

roponator:
We at Random Salad Games did some testing, but only for android.

We ran tests on AWS device farm for 3 of our most popular apps, each on ~13 devices and everything passed OK :)

tomspilman:
> We ran tests on AWS device farm for 3 of our most popular apps, each on ~13 devices and everything passed OK :)

That is likely the heaviest test we've ever gotten on a PR....  thanks @roponator !

dellis1972:
OK.. Rebased .. lets hope we don't have to do that again :P

dellis1972:
broke something :/

cra0zy:
Can't wait for this to get merged :)

KonajuGames:
> broke something :/

Did it get fixed with that follow-up commit?

dellis1972:
@KonajuGames it did. There were a bunch of compilation errors which the build machine caught. They are fixed now

KonajuGames:
Given the mass testing it got just recently and that it is now up-to-date, do you think it is ready to be merged?

roponator:
I'd say go for it, its a big change but its gotta be done, besides the users can revert to old monogame if they find a massive bug.

We at Random Salad Games depend heavily on the latest builds (from dev branch releases) so if something goes sideways I'll be looking into it :)

mrhelmut:
I'm also up to the task.

dellis1972:
well seems we have some people ready to fix all the stuff I broke :P lets
do it :D

cra0zy:
This branch makes it so I can properly run MG Android apps from Linux so I'll also be able to help out with Android stuff :)

KonajuGames:
Boom. Done.

cra0zy:
Wuhu \o/

nkast:
@dellis1972  `len` is set but never used.

https://github.com/MonoGame/MonoGame/blob/b914f314a22dae8530ecfd5232adc8ceec55782f/MonoGame.Framework/Graphics/Shader/ConstantBuffer.cs#L87-L93

dellis1972:
@nkast doh..

PR #5412, opened 2017-01-15
WrapAngle optimisation & tests

MichaelDePiazzi:
This is a performance optimisation for the `MathHelper.WrapAngle` method, plus a heap of tests proving that it returns the same results as XNA.

The same result can be calculated using the standard modulo operator which is much faster than the `Math.IEEERemainder` operation. While the 2 operations do give different results (e.g. `10f % MathHelper.TwoPi = 3.716815f` and `Math.IEEERemainder(10f, MathHelper.TwoPi) = -2.566371f`), these are accounted for by the additional limit checks. But don't take my word for it, the tests I've added prove that.

Here are my benchmarks of the original method versus this new one using modulo. The new method takes about 64% less time! But I also added a check to skip the calculations if the value is already within range, and in this case it takes 92% less time than the original!

Method | Mean | StdDev | StdErr | Min | Q1 | Median | Q3 | Max | P95 | Op/s | Scaled | Scaled-StdDev
--- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | ---
Original | 51.2411 ns | 0.1013 ns | 0.0281 ns | 51.1090 ns | 51.1818 ns | 51.2110 ns | 51.2752 ns | 51.4786 ns | 51.4391 ns | 19515601.45 | 1.00 | 0.00
New (outside range) | 18.7181 ns | 0.1280 ns | 0.0330 ns | 18.6689 ns | 18.6694 ns | 18.6699 ns | 18.6705 ns | 19.0591 ns | 19.0216 ns | 53424196.85 | 0.36 | 0.00
New (within range) | 4.1722 ns | 0.0004 ns | 0.0001 ns | 4.1715 ns | 4.1721 ns | 4.1722 ns | 4.1724 ns | 4.1729 ns | 4.1727 ns | 239682313.82 | 0.08 | 0.00

### DIFF
--- a/MonoGame.Framework/MathHelper.cs
+++ b/MonoGame.Framework/MathHelper.cs
@@ -295,22 +295,17 @@ namespace Microsoft.Xna.Framework
         /// </summary>
         /// <param name="angle">The angle to reduce, in radians.</param>
         /// <returns>The new angle, in radians.</returns>
-	public static float WrapAngle(float angle)
-	{
-            angle = (float)Math.IEEERemainder((double)angle, 6.2831854820251465);
-	    if (angle <= -3.14159274f)
-	    {
-		angle += 6.28318548f;
-	    }
-	    else
-	    {
-		if (angle > 3.14159274f)
-		{
-		   angle -= 6.28318548f;
-		}
-	    }
-	    return angle;
-	}
+        public static float WrapAngle(float angle)
+        {
+            if ((angle > -Pi) && (angle <= Pi))
+                return angle;
+            angle %= TwoPi;
+            if (angle <= -Pi)
+                return angle + TwoPi;
+            if (angle > Pi)
+                return angle - TwoPi;
+            return angle;
+        }
 
  	/// <summary>
         /// Determines if value is powered by two.
--- a/Test/Framework/MathHelperTest.cs
+++ b/Test/Framework/MathHelperTest.cs
@@ -83,5 +83,75 @@ namespace MonoGame.Tests.Framework
             Assert.AreEqual(5, MathHelper.Max(-5, 5));
             Assert.AreEqual(5, MathHelper.Max(5, -5));
         }
+
+        [TestCase(MathHelper.PiOver4, 0.7853982f)]
+        [TestCase(MathHelper.PiOver2, 1.5707964f)]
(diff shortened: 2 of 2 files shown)


theZMan:
Nice - I think with a slight tweak you can lose one of the branches - not sure if it will give you any more gains though. See http://stackoverflow.com/questions/11498169/dealing-with-angle-wrap-in-c-code

theZMan:
This one is cool too - no branches at all http://stackoverflow.com/questions/1628386/normalise-orientation-between-0-and-360 not sure if it would give the same results as XNA or not...

MichaelDePiazzi:
Thanks @theZMan! Unfortunately, neither of these alternative methods gives the same results as XNA.

The first one does seem to work, but it gives slightly different results to XNA mainly due to it having different limits. It's limits are `-Pi <= angle < Pi`, whereas XNA is `-Pi < angle <= Pi`. There's probably also some floating point precision differences due to it modifying the value before getting the remainder. I didn't benchmark it, but I'm guessing the performance difference would be negligible anyway.

The second one I'm assuming you meant the second answer (http://stackoverflow.com/a/2021986)? It definitely seems promising, but unfortunately it has the same issues as the first one (i.e. different limits to XNA). It did inspire me to try this alternative method though (which does give the same results as XNA):
```c#
public static float WrapAngle(float angle)
{
    if ((angle > -Pi) && (angle <= Pi))
        return angle;
    const double divisor = TwoPi;
    const double divisorReciprocal = 1.0 / divisor;
    angle = (float)(angle - (int)(angle * divisorReciprocal) * divisor);
    if (angle <= -Pi)
        return angle + TwoPi;
    if (angle > Pi)
        return angle - TwoPi;
    return angle;
}
```
The performance was basically identical to using modulo though.

MichaelDePiazzi:
@tomspilman @KonajuGames @dellis1972 - Hey guys, this should be good to merge. I've added pretty extensive unit tests for it, so I'm confident it's working correctly. Thoughts?

KonajuGames:
I'm happy with this.  Lots of unit tests is good, and better performance too. Merging.

MichaelDePiazzi:
Thanks @KonajuGames

PR #5414, opened 2017-01-15
Improve Point and Rectangle HashCode

Jjagg:
These hash codes were symmetric. That means `new Point(x, y).GetHashCode() == new Point(y, x).GetHashCode()` would evaluate to true. For Rectangle it's worse since any permutation of the 4 values will result in the same hashcode. I didn't check any other structs, but there are probably more hashcodes implemented like this. Though since these are mutable structs people shouldn't rely on the hashcode too much anyway. 

New algorithm came from http://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-an-overridden-system-object-gethashcode/263416#263416 by StackOverflow/C# god Jon Skeet.

### DIFF
--- a/MonoGame.Framework/Point.cs
+++ b/MonoGame.Framework/Point.cs
@@ -187,7 +187,14 @@ namespace Microsoft.Xna.Framework
         /// <returns>Hash code of this <see cref="Point"/>.</returns>
         public override int GetHashCode()
         {
-            return X ^ Y;
+            unchecked
+            {
+                var hash = 17;
+                hash = hash * 23 + X.GetHashCode();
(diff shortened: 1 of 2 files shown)


KonajuGames:
Assuming you meant (x,y) and (y,x) in that example.

It should be trivial to add some simple unit tests for these.

Jjagg:
> Assuming you meant (x,y) and (y,x) in that example.

Woops, yes I'll edit that. I don't think unit testing hashcodes holds much value. I mean we could test if symmetric values don't collide, but it's not like anyone will ever change this back to just xorring X and Y.

KonajuGames:
Merging. Thanks @Jjagg

PR #5419, opened 2017-01-16
Some minor additions to #5139

### LINKED ISSUE #3995: Infinite loop caused by Game.Exit doesn't allow application to end

### LINKED ISSUE #5416: Exit() might not close game's process
Calling Exit() too soon seems to prevent the game from completely closing, as the process still lives on.

Tested on MonoGame 3.5 and 3.6.

```

// This example simply adds a red rectangle to the screen
// then updates it's position along the X axis each frame.
namespace Game1
{
	public class Game1 : Game
	{

		public Game1()
		{
		}

		{
			Exit();
		}
	}
}
```

If you run this w/o debugging ( does not happen w/ debugging oddly ), the window quickly appears and closes, but the process will still exist in the Task Manager indefinitely until you end the task.

Jjagg:
Fixes a test that failed in XNA for now. We don't handle a lost device at all at this point :/
Also made the PeekMessage filter only let through WM_QUIT messages since that's all that were interested in.

With the PeekMessage filter #5139 fixed two more issues (autoclosing here): fixes #3995 and fixes #5416

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -571,9 +571,6 @@ namespace Microsoft.Xna.Framework.Graphics
 
             if (DeviceReset != null)
                 DeviceReset(this, EventArgs.Empty);
-
-            if (DeviceLost != null)
-                DeviceLost(this, EventArgs.Empty);
         }
 
         public void Reset(PresentationParameters presentationParameters)
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -354,17 +354,10 @@ namespace MonoGame.Framework
             // This is critical for some NUnit runners which
             // typically will run all the tests on the same
             // process/thread.
-#if DEBUG
-            var msg = new NativeMessage();
-            do
+            NativeMessage msg;
+            while (PeekMessage(out msg, IntPtr.Zero, WM_QUIT, WM_QUIT, 1))
             {
-                if (msg.msg == WM_QUIT)
-                    break;
-
-                Thread.Sleep(100);
-            } 
-            while (PeekMessage(out msg, IntPtr.Zero, 0, 1 << 5, 1));
-#endif
+            }
         }
 
         public void CenterForm()
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -354,10 +354,15 @@ namespace MonoGame.Framework
             // This is critical for some NUnit runners which
             // typically will run all the tests on the same
             // process/thread.
-            NativeMessage msg;
-            while (PeekMessage(out msg, IntPtr.Zero, WM_QUIT, WM_QUIT, 1))
+            var msg = new NativeMessage();
+            do
             {
-            }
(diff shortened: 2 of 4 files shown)


Jjagg:
The failing tests don't seem related to these changes. Build bot issues possible?

tomspilman:
> The failing tests don't seem related to these changes. Build bot issues possible?

Actually it is a NUnit/WinForms issue with the messages issued at shutdown of Game after a test.  You can see the same sort of errors 3 years ago when I originally got the unit tests running:

 https://github.com/MonoGame/MonoGame/pull/3052

The code I wrote at the end of `RunLoop()` was specifically to ensure all remaining messages are processed before the unit test exits.  On the next test when a new WinForm window is initialized it uses the same message pump created for the current NUnit process...  hence it will process messages from the last test if they are still in the message queue.  For example a message like WM_CLOSE...  which if I remember correctly is what is happening here causing the graphics device initialization to fail.

Ideally there would be a way to force recreation of the message pump on a new test or force clearing all messages from the message pump on a new test...  I didn't see a good way to do that.  This is why I had code there waiting until all the messages before WM_QUIT are processed.

Jjagg:
Ah, I thought only the WM_QUIT message needed to be consumed. I'll revert that change then.

KonajuGames:
I wonder why the tests didn't appear to run.

Jjagg:
It seems tests don't run when I don't actually make a change. I thought I could trigger the build by doing an empty commit, but that doesn't seem to work...

Jjagg:
This is good to go!

KonajuGames:
Merging.

PR #5420, opened 2017-01-16
Persist GraphicsDevice across tests and limit frame size for visual tests

Jjagg:
This PR greatly improves unit testing performance for graphics tests. Local graphics testing time went from ~37 seconds to ~7.

Here are the local run times in this branch as reported by ReSharpers unit test runner:
image

The stuff that still takes long is:

1. Audio conversion tests take about 7 seconds. We can probably cut down on the used audio duration a bit if we want to improve here. There are 3 audio files used, one of about 1 second and the other two of 7 seconds.
2. Compiling the stock effects takes about 10 seconds. Not sure if it's D3DCompiler or our code, this could be looked at for better performance.
3. IntermediateDeserializer takes some time because of IO I guess. Not much to do here.
4. The framework tests take over 3 seconds because of 1 leftover test that actually creates a window and runs a game. If not for that the Framework tests would take less than a second.
5. Some Graphics test still need to create a new device or reset the existing one (in GraphicsDeviceTest and GraphicsDeviceManagerTest), that takes a bit of time but is required for proper testing. One thing we could do to speed up graphics tests is run the frame comparison in parallel (not across tests, but for a single test or even a single frame). It shouldn't be too hard to try that out, so I might give that a shot.
6. There's one other test under Visual/Misc that creates a window and runs a game. I have a plan for adding a setting to run MG headless in Game if that's acceptable. That would fix this and 4. and will allow for testing Game more easily in the future. Could also be nice for users (http://community.monogame.net/t/headless-monogame/7063).
7. DynamicSoundEffectInstance does some sleeping to wait for soundeffects to finish/progress. it looks like this can be cut down on a little bit by reducing soundeffect duration.

So there's still some optimization that can be done probably.

While persisting the GraphicsDevice across tests gives a significant performance boost there is a small downside too. There is no way to reset the bound shaders and constantbuffers through the api. Tests that depend on shaders not being bound will need to call HardReset first. That disposes the current device and creates a new one. For other tests a SoftReset (called automatically) seems to suffice. I made the GraphicsDeviceSetUpFixture flag a device reset so it knows to do a hard reset after the test so no big changes are persisted (like changing the GraphicsProfile). 

We're not using 'fresh' GraphicsDevice every test in this branch, but everything is reset properly so I don't think there are risks of mistesting with these changes.


Jjagg:
C# 6.0 syntax :/

tomspilman:
A few things bother me about these changes.

>  limit frame size for visual tests

With this happening by manually specifying a rectangle i worry that if a test writes some data outside this rectangle in error...  we miss the bug.  Or if someone forgets to update the rectangle the tests pass when they shouldn't.

It seems error prone to me.


> While persisting the GraphicsDevice across tests gives a significant performance boost there is a small downside too. There is no way to reset the bound shaders and constantbuffers through the api. 

Yeah that worries me.  It could lead to tests that seem like they are working, but are not in a real world situation.

> We're not using 'fresh' GraphicsDevice every test in this branch, but everything is reset properly so I don't think there are risks of mistesting with these changes.

So how do we know the reset we are doing is happening properly?  If a new state is added how do we detect we didn't miss resetting it?

This all sounds like a big mistake waiting to happen.

Can we somehow unit test the reset to ensure we don't miss resetting state?


> So there's still some optimization that can be done probably.

In the end...  Is saving 30s to 1m of unit testing worth the danger of bad testing?  Is going from unit tests with code coverage taking 2m to 1m really that important?

I'm really not convinced this is worth the danger it potentially presents.

Jjagg:
I agree about the GraphicsDevice. It would be safer to just recreate it. I got a bit optimistic because the tests just worked and it increased performance so much, but looking back at this change I think we better not do this. This change might've caused the issues on the build bot too...

About the frame size. I actually started implementing this because I had a local test passing while the output was clearly wrong. The issue was the test only drew a small sprite in the corner of the rendertarget, so similarity was pretty high. We could just increase required similarity in those cases. The test I ran into this was a scissorrectangle test. The clipped sprite was really small, I don't think any other tests suffer from this, but we could check this.

I'll close this.

tomspilman:
> The issue was the test only drew a small sprite in the corner of the rendertarget, so similarity was pretty high.

Well I think in general we should try to avoid doing tiny renders like that... but I know it happens.

But our similarity test has been a problem.  I've never been really happy with it.  I would love to spend more time on finding a better test for incorrect results in the captured screens.

Jjagg:
>  I would love to spend more time

And then work/life happens! :p

PR #5421, opened 2017-01-16
Command-line parameter prefixes

KonajuGames:
Command-line parameters on Windows use /
Command-line parameters on Linux and Mac use - for short parameters and -- for long parameters.

Started this because - and -- is what non-Windows users expect, and 2MGFX fails to parse parameters on non-Windows platforms with absolute paths (starting with /) as the input or output parameters.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
@@ -30,6 +30,11 @@
       <MonoExe Condition=" '$(OS)' != 'Windows_NT' And Exists ('/usr/bin/mono') ">/usr/bin/mono</MonoExe>
       <MonoExe Condition=" '$(OS)' != 'Windows_NT' And '$(MonoExe)' == '' ">mono</MonoExe>
 
+      <LongParamPrefix Condition=" '$(OS)' != 'Windows_NT' ">--</MonoExe>
+      <LongParamPrefix Condition=" '$(OS)' == 'Windows_NT' ">/</MonoExe>
+      <ShortParamPrefix Condition=" '$(OS)' != 'Windows_NT' ">-</MonoExe>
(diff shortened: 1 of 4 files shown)


tomspilman:
@KonajuGames - How is this solved in MGCB?

KonajuGames:
MGCB had been modified a while back to accept both / and -, and it doesn't have any required parameters as filenames that could start with / on non-Windows systems.  I've made it consistent in this PR.

KonajuGames:
That's an odd test failure.

    GameComponentTest.InitializeOrderTest
    Test(s) failed. Microsoft.Xna.Framework.Graphics.NoSuitableGraphicsDeviceException : Adapter 'Intel(R) HD Graphics 4600' does not support the Reach profile.

tomspilman:
> That's an odd test failure.

My guess is that this change is the culprit:

https://github.com/MonoGame/MonoGame/commit/669db5c15462fd796eb1b473ece759e2f93f374d

tomspilman:
> MGCB had been modified a while back to accept both / and -

Why isn't the solution in MGCB not a good one for 2MGFX?

KonajuGames:
Can't use / at all for parameters on Linux and Mac if some of the parameters are filenames, because filenames on Linux and Mac can legitimately start with /.  2MGFX has two required parameters that are filenames.

KonajuGames:
An alternative to consider is to alter those two required parameters in 2MGFX to be like the rest and have names, e.g. change

    2MGFX filename1 filename2 /profile:DirectX11

to

    2MGFX /in:filename1 /out:filename2 /profile:DirectX11

and

    2MGFX -in:filename1 -out:filename2 -profile:DirectX11

benknoble:
@KonajuGames I believe that is similar to the way the C# compiler (both csc and mcs) work. IMO probably the best way to deal with compatibility.

KonajuGames:
I'll do that tonight and update this PR.

KonajuGames:
Replaced by #5438.

PR #5424, opened 2017-01-19
Tests always create reference device

KonajuGames:
Any tests that create a graphics device always create a reference device, rather than just the visual tests.


tomspilman:
Note this is good, but the actual bug is we're incorrectly detecting what `HiDef` means.

Issue #5425, opened 2017-01-20
Text drawing broken in develop DesktopGL


polsys:
**Steps to reproduce:**
1. Install MonoGame for Windows using the (as of now latest) 3.6.0.1379 installer.
1. Create a new DesktopGL project using the Visual Studio template.
1. Add a new sprite font to the content project. No need to modify the file.
1. Load the font in `Game1`:
```cs
protected override void LoadContent()
{
    spriteBatch = new SpriteBatch(GraphicsDevice);
    font = Content.Load<SpriteFont>("Font");
}
```
5. Draw some text using it:
```cs
protected override void Draw(GameTime gameTime)
{
    GraphicsDevice.Clear(Color.CornflowerBlue);
    spriteBatch.Begin();
    spriteBatch.DrawString(font, "Hello world", new Vector2(16, 16), Color.White);
    spriteBatch.End();

    base.Draw(gameTime);
}
```

No text appears. However, using the 3.6.0.1365 installer and later framework DLL builds, I narrowed it down to 1365 being the last working version and 1366 the first failing. This build contains #5121 and #5139, so either has caused the regression (could be related to @mrhelmut's comment in #5121?).

polsys:
On a further note, definitely related to compressed textures. Setting the font texture format to `Color` solves the issue.

Jjagg:
Thanks for the heads up. I'll see if I can fix this.

Jjagg:
I figured out the issue mrhelmut ran into and will set up a fix in a bit. I'm pretty sure it's not related to this issue though.

Jjagg:
So I thought it wasn't the same issue because we had wrong bindings so it should crash the applications, but I can't reproduce in my branch with the fix #5426

mrhelmut:
The issue in #5121 triggers a crash upon ```Content.Load```, hence if compressed fonts don't crash but don't render, the problem may be elsewhere. I'll try to test #5426 today and see if compressed fonts work.

PR #5426, opened 2017-01-20
[OpenGL] Fix SetData on compressed textures

Jjagg:
We had wrong bindings for CompressedTexSubImage2D and because of the changes in #5121 all SetData calls used that binding and crashed because of it.


Jjagg:
Woops, I only fixed the DesktopGL build!

polsys:
Can confirm this fixes #5425 (at least on Windows). Thank you for the quick response! I think this should be merged ASAP since develop is pretty broken without it.

KonajuGames:
Cool. Thanks for testing.  Merging.

PR #5427, opened 2017-01-21
Fix two bugs in OpenAL implementation

polsys:
* An oversight by me caused `Stop()` to be executed on an already disposed sound effect, which is a throwing operation with `DynamicSoundEffectInstance` (XNA compatibility).
* The native DLL search path was set in `SDLGamePlatform` constructor, but the audio tests are run first. Therefore `soft_oal.dll` was not found, which caused test failures and an epic finalizer crash, rendering the test project unrunnable as is. I added the path check to `OpenALSoundController` so that there is no load order dependency anymore.


KonajuGames:
Thanks for this @polsys.  Merging.

Issue #5428, opened 2017-01-21
Getting the OpenGL tests working


polsys:
This is the list of issues and test failures I had running the OpenGL tests on Windows. I haven't looked more closely into most of these, so I'd appreciate if someone more knowledgeable with GL taked a look.

- [ ] `soft_oal.dll` not found, and a mistake in `DynamicSoundEffectInstance` disposal. Fixed in #5427.
- [ ] The `BlendState` visual test fails with a considerable diff
- [ ] Some `Texture2D.Get/SetData()` tests fail with OpenGL error `GL_INVALID_OPERATION`
- [ ] Effect compilation is not implemented, which causes all the effect tests and a `SamplerState` test to fail. (`SpriteBatchTest` tries to find the effects from the `DirectX` folder.)
- [ ] Some `SoundEffect.FromStream()` tests fail
- [ ] `Texture3D` is not implemented but the tests are executed
- [ ] `GameTest+Misc.ExitHappensAtEndOfTick` fails
- [ ] One visual `RasterizerState` depth bias test fails. I'm not sure if the diff is totally correct here, though.

I applied the #5426 fix for the compressed texture regression, so those failures are omitted. That regression would have been caught if the tests were runnable.

mrhelmut:
> Some ```Texture2D.Get/SetData()``` tests fail with OpenGL error ```GL_INVALID_OPERATION```

Should be fixed with #5426

I'll try to look into some of these points. Great roadmap!

Jjagg:
> Did you test GraphicsDeviceManager

GDM and GD tests were not included in the GL project. I get 134 failing tests now. I'll see if I can fix some of them :)

polsys:
The tests are runnable and with #5429 there aren't that many failures left, so things are looking really good!

I stumbled into one unrelated issue with the generated HTML report: it misses some results. These seem to be test classes that have both ignored and passing tests, but no failures. However, they exist in the XML file. For instance, with #5429, the `Texture2DTest` results are missing, but the XML indicates the expected mixture of passing and ignored tests. Since failures are not hidden, this is a minor issue, but it makes looking for differences harder. Have you run into this?

Jjagg:
I run the tests wit ReSharper so I have not.

Jjagg:
#5429 is passing the build! I ignored most of the unresolved issues above except BlendState because of its significance. IMO that's one we should resolve in the short-term

PR #5429, opened 2017-01-21
Get OpenGL tests running and fix the failures

### LINKED ISSUE #1449: Texture2D GetData does not work when called from another thread
Hi, im loading and manipulating a texture. When i calling my method from the main thread, its working. when i call the same method from a thredpool thread, Every R, G, B and A value is 0. For better understanding, i have included my test method (im using the latest source from the develop branch and the WindowsGL project):

```
    public void ManipulateTextureTest (object o)
    {
        Texture2D t = World.LoadTexture(Path.Combine("Textures", "Map", "grasstree"));

        Texture2D texture = new Texture2D(World.Graphic, t.Width, t.Height);

        Color[] colors = new Color[t.Width * t.Height];
        t.GetData(colors);

        foreach (Color c in colors)
        {
            if (c.R != 0 || c.G != 0 || c.B != 0 || c.A != 0)
            {
                Console.WriteLine("a real pixel is there");
            }
        }

        for (int i = 0; i < colors.Length; i++)
        {
            if (colors[i].A < 255)
            {
                colors[i].A = 0;
                colors[i].B = 0;
                colors[i].G = 0;
                colors[i].R = 0;
            }
        }
        texture.SetData(colors);
    }
```

### LINKED ISSUE #4224: Texture2D.GetData<T>(T[], int, int) incorrect behaviour
When specifying a start index and element count in Texture2D.GetData, it's checking to see if the data array passed to it is of a length equal to startIndex + elementCount instead of comparing it to just elementCount.  This, of course, means that to get only the last pixel, you would need to pass an array equal to the size of the texture, instead of one with a length of 1.

### LINKED ISSUE #4708: Texture2D GetData returns Transparent colors when called from different thread
When called from a different thread the GetData returns a transparent colors [0,0,0,0]

I tried from main thread and it returns correct colors

OS: Windows 10
GPU: Nvidia GTX 960

Platform: DesktopGL

### LINKED ISSUE #5473: Texture2D.SaveAsImage only writes 1/4th of texture data (DesktopGL)
When creating a cursor,  MouseCursor.FromTexture2D calls SaveAsImage.  The resulting stream contains only 1/4th of the texture data, and the rest of the stream is zeroed.  A previous version compiled 3 months ago did not have this issue.

### LINKED ISSUE #5474: Texture2D.GetData<T> does not return the correct number of elements when  size of T is not the same as the pixel format. (OpenGL)
This is the cause of #5473

Jjagg:
I already got it down from 134 to 54 failing tests. It's just SoundEffect.FromStream and some graphics tests that are failing now. 

We're getting there :) #5428 

Ping @polsys

- fixes #4708, fixes #1449 by throwing if GetData is not called from the main thread (SetData is allowed, but queued and executed synchronously).
- fixes #4224, this might have already been fixed before this branch actually
- fixes #5474, fixes #5473


Jjagg:
I plan to fix the remaining failures in this PR too, so please don't merge yet

Jjagg:
Four failing graphics tests left and then there's the audio loading tests for a total of 13 tests that are still failing.

Build/Projects/MonoGame.Tests.definition
```diff
@@ -33,6 +33,33 @@
   </Properties>
 
   <Files>
+    <!-- Dependencies -->
+    <None Include="..\ThirdParty\Dependencies\SDL\Windows\x64\SDL2.dll">
+      <Platforms>WindowsGL</Platforms>
+      <Link>SDL2.dll</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="..\ThirdParty\Dependencies\openal-soft\Windows\x64\soft_oal.dll">
```
polsys: Visual Studio already copies the libraries into the `x64`/`x86` folders thanks to the framework reference, but this doesn't hurt even then.

polsys:
This looks really nice so far, great work! 👍 I tested these on Windows and can confirm your findings (I had some trouble with the HTML report, but I'll discuss that in #5428).

I think the prebuilt effects are a reasonable tradeoff, though it would be nice to test the complete effect pipeline from source code to screen. The tests could be refactored to use the compilation pipeline on DirectX and load prebuilt XNBs on OpenGL. That way we wouldn't lose any coverage.

Jjagg:
> I think the prebuilt effects are a reasonable tradeoff

I didn't think of it that way. My reasoning was that we have seperate tests for the processor and the tests that are prebuilt are for testing other stuff. It would be easy to make a change so they always get processed for DX though. I'd like to know what @KonajuGames thinks about this.

Jjagg:
I haven't done much audio stuff before so it would be nice if someone else could look at that. If no one does I'll try and get it fixed anyway and learn something along the way :)

polsys:
I'm thinking of a case where the processor succeeds but produces invalid output, or something similar. However, it's easy enough to rebuild the effects, so that's a pretty minor issue in my opinion. I'd be happy to go with this as is.

Jjagg:
You're right though. It's probably a good idea to process on the go until we write tests for stock effects.

polsys:
I looked into the audio issue. It seems to be nothing more than a missing format check. [This calculation](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs#L64) should be replaced with
```cs
var bytesPerSecond = freq;
if (format == ALFormat.Mono16 || format == ALFormat.Stereo8)
    bytesPerSecond *= 2;
else if (format == ALFormat.Stereo16)
    bytesPerSecond *= 4;

duration = TimeSpan.FromSeconds((float)size / bytesPerSecond);
```

Jjagg:
Awesome, thanks @polsys. For some reason AL.GenBuffers was failing when I had my earphones plugged in, causing a bunch of audio tests to fail (should we/do we catch this in an actual game?), but that indeed fixed the issues :)

Jjagg:
There's a couple tests that are ignored for now and need more investigation. There's one failing test left and it's the visual blendstate test. I'll fix the build for now so this can be merged. Once we get #5284 and this one in and we figure out the blendstate issues @tomspilman can perhaps look into getting GL tests running on the build server :)

polsys:
I agree it's better to not make this PR any bigger. I'll update the issue to list the ignored tests. Very nice work! 👏

It would be nice if somebody checked this out on Mac, but the Windows side is now good.

Jjagg:
Oops, stuff got reformatted with that last commit.

polsys:
Actually, not all okay. I just looked at the `MiscellaneousTests.DrawOrder_falls_back...` test diff on GL, and it doesn't look right. Instead of glyphs, I have colored quads on frames 3 to 5. The default similarity threshold is too low to catch it failing.

Jjagg:
Building! :D 

This is ready for review and merge! I think you'll be happy with this @KonajuGames @tomspilman :)

Build/Projects/MonoGame.Tests.definition
```diff
@@ -5,7 +5,7 @@
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml" />
-	<Reference Include="System.Web" />
+    <Reference Include="System.Web" />
```
tomspilman: I am wondering what this reference is for... any idea?

MonoGame.Framework/Graphics/Vertices/IndexBuffer.OpenGL.cs
```diff
@@ -142,7 +142,7 @@ void GenerateIfRequired()
                 GraphicsExtensions.CheckGLError();
             }
             
-            GL.BufferSubData(BufferTarget.ElementArrayBuffer, (IntPtr)offsetInBytes, (IntPtr)sizeInBytes, dataPtr);
+            GL.BufferSubData(BufferTarget.ElementArrayBuffer, (IntPtr) offsetInBytes, (IntPtr) sizeInBytes, dataPtr);
```
tomspilman: Can we revert this whitespace change?  Just makes it hard to see real diffs.

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.FramebufferHelper.cs
```diff
@@ -34,48 +34,88 @@ namespace Microsoft.Xna.Framework.Graphics
...
-            internal const All AllFramebufferIncompleteMultisampleApple = (All)0x8D56;
-            internal const All AllMaxSamplesApple = (All)0x8D57;
-            internal const All AllReadFramebufferApple = (All)0x8CA8;
-            internal const All AllDrawFramebufferApple = (All)0x8CA9;
-            internal const All AllRenderBufferSamplesApple = (All)0x8CAB;
+            internal const All AllFramebufferIncompleteMultisampleApple = (All) 0x8D56;
+            internal const All AllMaxSamplesApple = (All) 0x8D57;
+            internal const All AllReadFramebufferApple = (All) 0x8CA8;
+            internal const All AllDrawFramebufferApple = (All) 0x8CA9;
+            internal const All AllRenderBufferSamplesApple = (All) 0x8CAB;
```
tomspilman: Can we revert these whitespace changes here?  It makes is hard to actually see the diffs.

Test/Framework/Graphics/SpriteBatchTest.cs
```diff
@@ -296,7 +294,7 @@ public void DrawWithTexture()
         [Test]
         public void DrawWithCustomEffectAndTwoTextures()
         {
-            var customSpriteEffect = AssetTestUtility.CompileEffect(gd, "CustomSpriteBatchEffect.fx");
+            var customSpriteEffect = content.Load<Effect>(Paths.Effect("CustomSpriteBatchEffect"));
```
tomspilman: @KonajuGames - How do you feel about this change?

I would say the alternative here is to only pre-compile the shaders for OpenGL and let the tests still compile the shaders at test time for platforms where we can.  This give us a little more test coverage.

That said we do need a good way to support platforms where the effects (and other content) have to be compiled offline.  This is required if we ever want to run some of these tests on mobile devices or consoles.  So there are clearly parts of this we want to keep.

Build/Projects/MonoGame.Tests.definition
Jjagg: We can probably get rid of that.

Test/Framework/Graphics/SpriteBatchTest.cs
Jjagg: @polsys and I also talked about this a little. How about we change these to AssetTestUtility.LoadEffecr and in there do

#if DIRECTX
    // compile and return effect
#else 
    // return a prebuilt effect
#endif

KonajuGames: When you get the runtime going and tests are running on Windows, I think it should do the same as DirectX and build the effect.  For now, I think the above should be ok.

Jjagg:
> ContentPipeline\IntermediateDeserializerTest.cs(453,30): error CS0103: The name 'HttpUtility' does not exist in the current context

I guess that's what the System.Web reference is for

MonoGame.Framework/Graphics/Shader/Shader.OpenGL.cs
```diff
@@ -128,11 +128,14 @@ protected override void Dispose(bool disposing)
             if (!IsDisposed && _shaderHandle != -1)
             {
                 Threading.BlockOnUIThread(() =>
+                {
+                    if (!IsDisposed)
```
Jjagg: Ow, I added this because the GL.DeleteShader call caused a crash sometimes because it was being called twice and I thought the dispose action might have been queued twice, but that is not the case since IsDisposed will return true after the base.Dispose call. I accidentally added this somewhere in a commit somewhere, but will revert this. Details of how and when the crash happened are here: 
https://github.com/MonoGame/MonoGame/pull/5429/commits/8fd62bb39838d33e59d45982bff40f07d883b88d#diff-d60307831e093a2b2bddf43a3ad0dd95R20

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.FramebufferHelper.cs
```diff
@@ -84,32 +119,39 @@ internal void GLBlitFramebufferApple(int srcX0, int srcY0, int srcX1, int srcY1,
...
-            internal const All AllFramebufferIncompleteMultisampleNV = (All)0x8D56;
-            internal const All AllMaxSamplesNV = (All)0x8D57;
-            internal const All AllReadFramebufferNV = (All)0x8CA8;
-            internal const All AllDrawFramebufferNV = (All)0x8CA9;
-            internal const All AllRenderBufferSamplesNV = (All)0x8CAB;
+            internal const All AllFramebufferIncompleteMultisampleNV = (All) 0x8D56;
+            internal const All AllMaxSamplesNV = (All) 0x8D57;
+            internal const All AllReadFramebufferNV = (All) 0x8CA8;
+            internal const All AllDrawFramebufferNV = (All) 0x8CA9;
+            internal const All AllRenderBufferSamplesNV = (All) 0x8CAB;
```
Jjagg: I missed these, I'll revert them too.

Test/Framework/Graphics/TextureCubeTest.cs
```diff
@@ -177,6 +186,9 @@ public void GetAndSetDataDxtDontRoundWhenOutsideBounds(string texName, int bs)
 
         [TestCase("SampleCube64DXT1Mips", 8)]
         //[TestCase("SampleCube64DXT5Mips", 16)]
+#if DESKTOPGL
+        [Ignore("GetData is not yet implemented for compressed images")]
+#endif
```
Jjagg: I should get rid of these ignores and see if it all check out since I implemented GetData for compressed textures in this too.

Test/Framework/Graphics/SpriteBatchTest.cs
tomspilman: > How about we change these to AssetTestUtility.LoadEffecr and in there do

Yeah I think having the `AssetTestUtility.LoadEffect` which internally builds the effect on desktop platforms that support effect content building or loads the effect on platforms that do not...  is the best path forward.

Test/Framework/Graphics/TextureCubeTest.cs
tomspilman: Lets just avoid cramming too much into this one PR.  It is fairly big already.

Jjagg: It's already in here, but I can pull it out if you want

tomspilman: Naw... no need to pull it out now.  Let's just avoid adding more to this.

Jjagg:
Okay, I'm done now.

Jjagg:
If you add **?w=1** after the URL when watching the diff it doesn't show any whitespace changes. Neat trick :)

Jjagg:
@tomspilman Any idea what's happening here?

>  Number of tests 3 is 897 less than the provided threshold 900 

>[run_tests] 2017-01-27 08:43:11,925 [11284] ERROR JetBrains.BuildServer.NAntLoggers.NUnit2LoggingTask.NUnit2TaskEx - NUnit launcher returns invalid exit code

Things are working locally.

tomspilman:
> Any idea what's happening here?

A few weeks ago I put a check on the server to ensure we don't accidentally break something and less tests are executed than expected (something that happened before).  Only 3 tests were run all successfully and the other 800+ were not run.  Hence my min tests check worked... we caught it.

I'm going to force it to re-run the build to see if it happens again.  However my guess is that https://github.com/MonoGame/MonoGame/pull/5419 broke things.  Any time we mess with the WM_QUIT hack we often break the unit test running.

Jjagg:
It's failing consistently at the third test. #5419 didn't change the WM_QUIT filtering, but 669db5c15462fd796eb1b473ece759e2f93f374d did. It's been a while since that was merged though.

Do tests run on release builds? Because that would mean #5419 actually enabled the message peeking.

Jjagg:
I'll remove the message filter. Let's see if that fixes it. That will reintroduce the 'infinite loop on exit' bug though...

KonajuGames:
> It's failing consistently at the third test. #5419 didn't change the WM_QUIT filtering, but 669db5c did. It's been a while since that was merged though.

It's not just the third test.  It's skipping the first 349 tests and starting on the three FrameworkDispatcher tests, then quits with error code 1 but no other logs.

This does not happen when running the tests locally.

Jjagg:
I noticed this branch doesn't even have #5419 in it, so it doesn't seem related.

KonajuGames:
The earliest I can find this 3 test error was the merge of #5423, which just made some platform methods private.  The tests in the PR passed, but the tests after the merge started failing with only 3 tests run.  #5419 was merged a day after that, and it was passing tests in the PR and then tests after merge failed.

KonajuGames:
NAnt reports an attempt to access an unloaded AppDomain.

````
[09:44:59][NAnt output] 2017-01-28 09:44:57,740 [3996] WARN  JetBrains.Container.Log4netDefaultLoggerInitializer - Unhandled exception in current domain: System.AppDomainUnloadedException: Attempted to access an unloaded AppDomain., , (null)
[09:44:59][NAnt output] 
[09:44:59][NAnt output] Unhandled Exception: System.AppDomainUnloadedException: Attempted to access an unloaded AppDomain.
````

KonajuGames:
Great work.  I'm going to merge this now as it fixes several outstanding issues.  Thanks @Jjagg

PR #5431, opened 2017-01-22
Fix unit test report missing some tests

polsys:
As mentioned in #5428, the HTML report generated by the test runner missed test classes that had both successful and ignored tests, but no failures.

The problem was that the XSL transform first displayed failed and then succeeded tests. However, the template for succeeded tests only ran when there were no failed *or ignored* tests. No path handled the successful+ignored case. There was no need at all for the two code paths, since the original code also sorted the test cases based on the outcome.

Also removed some files which seem to have existed for testing the XSL template back in 2012.


tomspilman:
Thanks @polsys!  Note we only use this when manually running the tests.  The build server has its own reporting system which didn't have this bug.

polsys:
The report has some trouble with `MiscellaneousTests` existing in two namespaces, I'll try take a look at it tomorrow. (Possibly just combining the two classes.)

Issue #5054, opened 2016-08-01
UWP SoundEffects are only played on the left speaker


cra0zy:
Title says it all, it's very easy to notice if wearing headphones. I think that this might affect other XAudio platforms as well, didn't really test it.

tomspilman:
It must be an issue in either `UpdateOutputMatrix` or `CalculatePanMatrix`:

https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs#L241

cra0zy:
SoundEffectInstance is played on both... wow this was unexpected... tho it may be heard higher on the left one.

tomspilman:
> SoundEffectInstance is played on both... wow this was unexpected

So are you saying that it wasn't only playing sounds on the left speaker?


> tho it is heard a bit higher on the left one.

I suggest verifying that it still is louder on the left by flipping your headphones around.  If it is still louder on the left with your headphones on backwards... it is your ears and not the sound effect system at fault.

I do this all the time to make sure i'm not fooling myself. :)

cra0zy:
> So are you saying that it wasn't only playing sounds on the left speaker?

SoundEffect is playing only on the left, SoundEffectInstance is playing on the both.

tomspilman:
> SoundEffect is playing only on the left, SoundEffectInstance is playing on both.

Got it.

Sounds like a failure to either get the right instance from the pool or properly resetting the instance before it is used.

vcrobe:
All my sound files are stereo and have the same format _.ogg_.

In my case almost all sounds are played on the left speaker only but some sounds are played stereo.

tomspilman:
So I think this is a bug in how our panning code is done here:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs#L265

If we change that panning code to match what I did for MonoGame on XB1:

 https://gist.github.com/tomspilman/9464d6646c9dbe059f6c2e516a18df7e

I suspect this may fix the issue and be more correct in the process.  This cpp code is based on my porting of the MonoGame C# code and from the sound code in DirectXTK.

leeadkins:
This seems simple enough for me to jump in on, but before I submit a PR, I just want to double check that this is what you had in mind:

https://github.com/leeadkins/MonoGame/commit/c5a9acc709bc4e19856b59523a8d5493f3bd4083

[...]

PR #5433, opened 2017-01-23
fix: SoundEffect.Play() was not setting volume, pitch, or pan

MichaelDePiazzi:
SoundEffect.Play() was not setting volume, pitch, or pan. This meant that if it grabbed an instance from the pool, it would use whatever volume / pitch / pan that instance previously had.

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.cs
@@ -262,13 +262,7 @@ namespace Microsoft.Xna.Framework.Audio
         /// </remarks>
         public bool Play()
         {
-            var inst = GetPooledInstance(false);
-            if (inst == null)
-                return false;
-
-            inst.Play();
-
-            return true;
+            return Play(1f, 0f, 0f);
         }
 
         /// <summary>Gets an internal SoundEffectInstance and plays it with the specified volume, pitch, and panning.</summary>


MonoGame.Framework/Audio/SoundEffect.cs
```diff
@@ -262,13 +262,7 @@ public static int GetSampleSizeInBytes(TimeSpan duration, int sampleRate, AudioC
...
         public bool Play()
         {
-            var inst = GetPooledInstance(false);
-            if (inst == null)
-                return false;
-
-            inst.Play();
-
-            return true;
+            return Play(1f, 0f, 0f);
```
tomspilman: This looks to already be handled in `SoundEffectInstancePool.GetInstance(bool)`:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstancePool.cs#L65

MichaelDePiazzi: Sorry, you are right. I though I was having problems with this, but I just tried again and it works as expected. I must've got it confused with the other problems I'm having with `CalculatePanMatrix` (issue #5054).

Do you want to keep it anyway (given it tidies up the code a bit)? Otherwise, I'll just close this.

tomspilman: Currently the `Play(float volume, float pitch, float pan)` sets these twice... once to the default and once to the actual desired values.  `Play()` avoids this double setting currently.

So if we wanted to improve things here I would like to see the volume/pitch/pan be passed down into the pool so it can reset the values one time to the right state.

MichaelDePiazzi: Good point. I took a look at this, but I think to do this well it would require a bit more work than what I have time for at the moment. For example, WinDX calls `PlatformSetupInstance` after getting the instance. This can potentially re-create the voice, which means any non-default settings must be applied after it's called. So it works now, but if I changed it to reset the values in the pool it could get reset by `PlatformSetupInstance` afterwards.

So I'll just close this PR for now, and I may look into improving this area of the code at a later point.

PR #5438, opened 2017-01-24
Simplified parameter prefix

KonajuGames:
Replaces #5421.

* Added "-" as an optional parameter prefix in addition to "/" in 2MGFX.
* Source file and output file parameters in 2MGFX now use "/SourceFile:" and "/OutputFile:" parameter names to avoid trying to differentiate between parameters and absolute paths on Linux and Mac.

### DIFF
--- a/Tools/2MGFX/CommandLineParser.cs
+++ b/Tools/2MGFX/CommandLineParser.cs
@@ -34,6 +34,7 @@ namespace TwoMGFX
         public CommandLineParser(object optionsObject)
         {
             this._optionsObject = optionsObject;
+            var prefix = Environment.OSVersion.Platform == PlatformID.Win32NT ? "/" : "-";
 
             // Reflect to find what commandline options are available.
             foreach (var field in optionsObject.GetType().GetFields())
@@ -54,9 +55,9 @@ namespace TwoMGFX
                     _optionalOptions.Add(fieldName.ToLowerInvariant(), field);
 
                     if (field.FieldType == typeof(bool))
-                        _optionalUsageHelp.Add(string.Format("/{0} {1}", fieldName, description));
+                        _optionalUsageHelp.Add(string.Format("{2}{0} {1}", fieldName, description, prefix));
                     else
-                        _optionalUsageHelp.Add(string.Format("/{0}:value {1}", fieldName, description));
+                        _optionalUsageHelp.Add(string.Format("{2}{0}:value {1}", fieldName, description, prefix));
                 }
             }
         }
@@ -86,7 +87,7 @@ namespace TwoMGFX
 
         bool ParseArgument(string arg)
         {
-            if (arg.StartsWith("/"))
+            if (arg.StartsWith("/") || arg.StartsWith("-"))
             {
                 // After the first escaped argument we can no
                 // longer read non-escaped arguments.
--- a/Tools/2MGFX/Options.cs
+++ b/Tools/2MGFX/Options.cs
@@ -6,10 +6,10 @@ namespace TwoMGFX
 {
     public class Options
     {
-        [CommandLineParser.Required]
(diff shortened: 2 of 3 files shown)


tomspilman:
> Source file and output file parameters in 2MGFX now use "/SourceFile:" and "/OutputFile:"

Is that really required for Mac and Linux?  Is this mostly an issue of having spaces in the file paths?

KonajuGames:
It's an issue that paths on Mac and Linux can start with /.  The command-line parser then cannot differentiate between a required (unnamed) parameter and an optional parameter.

KonajuGames:
Using / for parameters on Mac and Linux is very much against standards.

tomspilman:
> It's an issue that paths on Mac and Linux can start with /. The command-line parser then cannot differentiate between a required (unnamed) parameter and an optional parameter.

Ah... I see...  the issue is more our fault than a Mac or Linux limitation.  Since a path on Mac/Linux can start with `/` then the command line parser thinks it is a optional argument failing to parse it correctly and then errors out.

Is that correct?

KonajuGames:
> the command line parser thinks it is a optional argument failing to parse it correctly and then errors out.

Correct.

tomspilman:
So maybe we just don't support `/` for optional arguments on Mac/Linux... just `-`.  Since it isn't standard to use it on those platforms and it confuses the parser.  This way we can still have the first two required arguments as expected without changing how 2MGFX has worked on Windows since its inception.

On Windows we could support both `/` and `-` for now and start encouraging users to use `-` by changing the help spew and docs.  This way we eventually get to `-` being the standard for commands in 2MGFX.

Now switching to this same technique for MGCB would be a bit more painful, but we should possibly consider it eventually.

KonajuGames:
> So maybe we just don't support / for optional arguments on Mac/Linux... just -.

That's what my first PR did.  Used / on Windows and -- on Mac and Linux since the standard is to use - for short/abbreviated parameter names and -- for long-form parameter names.  I made the Pipeline tool call MGCB with the correct parameter prefix based on the platform it is running on.

KonajuGames:
We could easily move towards - and -- as a common parameter prefix on all platforms as there are many precedents for this on Windows now.

tomspilman:
> That's what my first PR did.

I guess I didn't read your description on that one clearly as you do mention the parser confusion issue there.

> We could easily move towards - and -- as a common parameter prefix on all platforms as there are many precedents for this on Windows now.

Yeah that is what I was thinking with as many unix tools appearing on Windows now.  Heck event GIT on Windows uses `-` and `--`.

Bonus for us that the .targets file can just use `-` and know it works on all platforms.

Still for now we would keep both `/` and `-` just to avoid breaking things for existing users that have built custom tools to automate their builds.

KonajuGames:
Here's the biggest change in this.  When .mgcb files are saved from the Pipeline tool, should they get written with `--`?  I'll make it so that it will always read both `/` and `--` on all platforms to avoid breaking anyone's content builds, but should files be written with `--` when saving.

tomspilman:
> but should files be written with -- when saving.

I think for now we can leave it with writing `/` to the .mgcb files for now.  Save that change until we've gotten the command line bits right.

Jjagg:
If we're gonna make this (or at least allow this to be) more Unix-like, maybe we should also consider replacing the `[prefix]Key:Value` notation by `[prefix]Key Value` as is customary, using whitespace as a separator. This could be implemented with backwards compatibility by having the MGCB options parser split keys and values using whitespace and colon as separators, then in the loop over the result of the `string.Split` call track if the last parsed value was a key that expects a value and check that the current value is what's expected (a key or value).

tomspilman:
> If we're gonna make this (or at least allow this to be) more Unix-like

Well lets not go overboard here.  I would be happy just seeing the command line issue we have with the first two required arguments solved.

Lets keep everything else to a wishlist thing later.

PR #5440, opened 2017-01-25
XAudio: Fixed output channel levels to match XNA

### LINKED ISSUE #5054: UWP SoundEffects are only played on the left speaker

MichaelDePiazzi:
There are currently a few issues relating to sound effect output channel levels in MonoGame. e.g.

- Stereo samples sometimes only playing on the left speaker
- Levels being too loud when panning is applied
- Levels being too loud when more than 2 output speakers are configured (e.g. 5.1/7.1 Surround)

This PR fixes these issues, and makes the output channel levels in line with XNA (see issue #5054 for further details).

I've also added some XAudio specific unit tests which confirm that the output matrix calculations are returning expected results.

Fixes #5054


tomspilman:
This looks like a good fix to me...  merging!

PR #5442, opened 2017-01-26
Monogame did not respect Android screen orientation lock by user, fixed.

roponator:
Added check if user locked screen orientation. If he did lock it, then screen orientation is not changed anymore.

### DIFF
--- a/IDE/MonoDevelop/MonoDevelop.MonoGame.Android/templates/Android/Activity1.cs
+++ b/IDE/MonoDevelop/MonoDevelop.MonoGame.Android/templates/Android/Activity1.cs
@@ -18,6 +18,7 @@ namespace ${Namespace}
 	           Theme = "@style/Theme.Splash",
                AlwaysRetainTaskState=true,
 	           LaunchMode=LaunchMode.SingleInstance,
+			   ScreenOrientation = ScreenOrientation.FullUser,
 	           ConfigurationChanges = ConfigChanges.Orientation | 
 	                                  ConfigChanges.KeyboardHidden | 
 	                                  ConfigChanges.Keyboard |
--- a/MonoGame.Framework/Android/AndroidGameWindow.cs
+++ b/MonoGame.Framework/Android/AndroidGameWindow.cs
@@ -292,8 +292,6 @@ namespace Microsoft.Xna.Framework
                         TouchPanelState.ReleaseAllTouches();
                     }
 
-                    Game.Activity.RequestedOrientation = requestedOrientation;
-
                     OnOrientationChanged();
                 }
             }
--- a/MonoGame.Framework/Android/OrientationListener.cs
+++ b/MonoGame.Framework/Android/OrientationListener.cs
@@ -4,6 +4,7 @@ using Android.Content;
 using Android.Content.Res;
 using Android.Hardware;
 using Android.Views;
+using Android.Provider;
 
 namespace Microsoft.Xna.Framework
 {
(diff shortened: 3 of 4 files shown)


jakepoz:
Just tested this, it's almost correct, but will fail with the following steps (Testing on Kindle Fire tablet)

1. On your device's main app launcher page, hold the device in portrait mode, and enable the Rotation lock in Portrait mode.
2. Hold your device in landscape mode, the rotation remains locked into portrait because of the Rotation lock setting.
3. Launch a Monogame project that supports both Portrait and Landscape mode.
4. The game will launch in Landscape mode, and be stuck in that setting, even though it should have locked itself into Portrait mode.

roponator:
Could check what your ScreenOrientation is set to in the Android Activity.cs file (Activity1.cs in my case).

It is supposed to be ScreenOrientation = ScreenOrientation.FullUser, can (did) you test with this value?

jakepoz:
Indeed, I have ScreenOrientation.Sensor, will try shortly with FullUser.

roponator:
There is another strange rotation bug, the rotation seems to be offseted by 90 degrees, checking it out now...

roponator:
I think I found the real reason for this issue, the previous fix was more of a workaround and is not needed anymore.

This fix just removes one line: that line overrode the FullUser (or some other) sensor mode.

roponator:
So if you can please test this one now Jake :)

jakepoz:
Hey, the latest fix while also setting the rotation mode to FullUser worked for me in a sample project, will test it in a real game shortly.

jakepoz:
Sweet, just confirmed this does fully fix the problem for us in a live game we have on the Google Play and Amazon stores.

@KonajuGames Can you please consider merging this change into the main develop branch?

KonajuGames:
Android has a confusing range of screen orientation options, and they keep adding more to it.  So should `ScreenOrientation.FullUser` be the default now?

jakepoz:
Yeah, I think FullUser should be the new default, but it will only work right with the rest of this change.

KonajuGames:
In that case, could we also get these lines changed as part of this PR.

https://github.com/MonoGame/MonoGame/blob/develop/ProjectTemplates/VisualStudio2010/Android/Activity1.cs#L14
https://github.com/MonoGame/MonoGame/blob/develop/IDE/MonoDevelop/MonoDevelop.MonoGame.Android/templates/Android/Activity1.cs#L20

This will fix new projects created from the templates.

roponator:
Hm... So if we change the default to FullSensor we may break some existing apps when they will update their monogame code if they rely on the default values for orientation, that why I'd suggest we leave it as it is. 

@KonajuGames why would you like to change this: LaunchMode=LaunchMode.SingleInstance ?

KonajuGames:
> why would you like to change this: LaunchMode=LaunchMode.SingleInstance ?

Not change that specific line.  The ScreenOrientation setting is missing from that template.

> Hm... So if we change the default to FullSensor we may break some existing apps when they will update their monogame code if they rely on the default values for orientation, that why I'd suggest we leave it as it is.

If they generated their project from the Visual Studio template, it will have a ScreenOrientation setting in the Activity already.  I don't think many projects have been generated from the MonoDevelop template, but if the developer was smart they would have added a ScreenOrientation setting anyway.

Changing these templates only affects new projects moving forward.  It does not retroactively change existing projects.

roponator:
Ok if it's in the template then it's ok.

I pushed the changes propsed by @KonajuGames (was logged in as different user when commited though).

jakepoz:
@KonajuGames Hey, any chance we can merge this in?

KonajuGames:
Merging.

PR #5444, opened 2017-01-28
Fix bugs in MeshHelper

Jjagg:
I made some mistakes when implementing this stuff in #5408. There was also a bug in CalculateNormals.
And I removed a level of indirection that was not needed in IndexUpdateList which makes it way easier to understand.

I hope it's all working with this, this is hard to write tests for. Well, more arduous than hard because there are so many options and it's a lot of work to create a mesh manually. Would be nice if we could test with a model that we know has some duplicate stuff in it and compare results against XNA.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/MeshBuilder.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/MeshBuilder.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
             _currentGeometryContent.Indices.Add(pos);
         }
 
-        public int CreateVertexChannel<T>(string usage) where T : struct
+        public int CreateVertexChannel<T>(string usage)
         {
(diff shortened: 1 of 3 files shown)


mattiascibien:
Nice catches. I guess the usage of the indices instead of the vertices in https://github.com/MonoGame/MonoGame/pull/5444/files#diff-d17bd30c75063679110cbc506e028b0fL130 was made wrong by me with my initial implementation.

Jjagg:
Nope, the issue was already in the initial implementation:
https://github.com/MonoGame/MonoGame/commit/9a03c72d3db98d92bebb9be3d5697816b1bfba75#diff-d17bd30c75063679110cbc506e028b0fR126

The implementation still isn't completely right with these fixes. I got a complex project that uses MeshBuilder from someone and it still crashes on this branch (it's working in XNA).

KonajuGames:
Merging. Thanks @Jjagg

PR #5445, opened 2017-01-28
DesktopGL was not using back buffer and depth buffer formats

### LINKED ISSUE #5441: DesktopGL stencil operations not working
I have the same project, configured with Windows and DesktopGL.

In the windows one the stencil buffer is working as expected.
In the DesktopGL one the stencil buffer is not working at all.

Monogame version a few weeks old.

Both solutions here: 
https://we.tl/03Wsr5YuLy (147 kb)

For reference:
```
_stencilWriteOnly = new DepthStencilState()
            {
                DepthBufferEnable = false,
                DepthBufferWriteEnable = false,
                CounterClockwiseStencilFunction = CompareFunction.Always,
                StencilFunction = CompareFunction.Always,
                StencilFail = StencilOperation.IncrementSaturation,
                StencilPass = StencilOperation.IncrementSaturation,
                CounterClockwiseStencilFail = StencilOperation.IncrementSaturation,
                CounterClockwiseStencilPass = StencilOperation.IncrementSaturation,
                ReferenceStencil = 0,
                StencilEnable = true,
                StencilMask = 0,
            };

        _stencilReadOnly = new DepthStencilState()
        {
            DepthBufferEnable = false,
            DepthBufferWriteEnable = false,
            CounterClockwiseStencilFunction = CompareFunction.Less,
            StencilFunction = CompareFunction.Less,
            StencilFail = StencilOperation.Keep,
            StencilPass = StencilOperation.Keep,
            CounterClockwiseStencilFail = StencilOperation.Keep,
            CounterClockwiseStencilPass = StencilOperation.Keep,
            ReferenceStencil = 0,
            StencilEnable = true
        };
```

KonajuGames:
OpenGL was being initialized with default OpenGL formats, which does not include a stencil buffer.  SDL2 requires the SDL GL attributes to be set before the window is created.  The window creation has been delayed to just prior to the graphics device being created and the setting of the SDL GL attributes was moved earlier to just prior to the window being created.  Unit tests were added for stencil clip.

Fixes #5441


Jjagg:
> SDL2 requires the SDL GL attributes to be set before the window is created.

This is a bit crappy :/ OpenGL requires you to set up the depth/stencil-buffer for the default framebuffers when creating the context. However, native API's for creating a context (like WinAPI and GLX) actually require you pass flags to specify DepthBufferFormat when creating a window instead of when creating the context. So we can't get around this restriction. (I don't get why the window needs to know about the DepthBuffer, but it's anoying as hell...). That means if the user wants to change the DepthBufferFormat we have to recreate the SDL window. The changes in this PR as is only set the DepthBufferFormat at startup. 

The implemenation would do something like GraphicsDeviceManager calling RecreateWindow on the GamePlatform.
 
This is all handled nicely in Direct3D btw, we can just dynamically bind any DepthStencilView to use as the DepthBuffer. So when the depthbufferformat changes we just create a new DepthStencilView and bind that when drawing to the backbuffer.

#5429 makes the DesktopGL platform use the new GraphicsDeviceManager, so the changes in GraphicsDeviceManager.Legacy in this PR would need to be moved to PlatformInitialize (called [here](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/GraphicsDeviceManager.cs#L368)) in a new file GraphicsDeviceManager.OpenGL when #5429 is merged. Probably easier to get #5429 merged first and then make the necessary changes here.

KonajuGames:
Agreed. It is annoying and makes changing the formats at runtime more
difficult (which this PR didn't handle). I'll have a look at those PRs you
listed.

dellis1972:
This looks ok to me, @cra0zy ?

dellis1972:
Why did the build fail?

dellis1972:
Ah its because the tests failed.

cra0zy:
Window needs to be created in the class tantalizer, otherwise you'll cause a lot of other problems...

KonajuGames:
> Window needs to be created in the class tantalizer, otherwise you'll cause a lot of other problems...

Problems such as what?  We must set the OpenGL attributes before the window is created due to how SDL2 behaves.  We don't have the required settings until after the Game constructor.

cra0zy:
> Problems such as what?

People accessing some stuff without game.Run or within the their Game constructors.

> We must set the OpenGL attributes before the window is created due to how SDL2 behaves. We don't have the required settings until after the Game constructor.

I know, I never said you shouldn't, the best course of action is, as @Jjagg said, to implement a `RecreateWindow` function and also to create a feature request on the <a href="https://bugzilla.libsdl.org/"> SDL bug tracker </a> to allow changes during window runtime.

Alternatively create a fake window at the start like I did before: https://github.com/MonoGame/MonoGame/blob/6af92b91f13cb5710755df7c74bbd8acb272d354/MonoGame.Framework/SDL/SDLGameWindow.cs#L109

Jjagg:
> implement a RecreateWindow function

I thought about this some more and this might not be a great idea because it makes GDM depend on the GameWindow... We could have the window register to an event in GDM to not make GDM depend on a window being available? I think we should try to get rid of the GDM->GameWindow dependency for the sake of embedding i.e. to let users create their own window. Though we probably need to expose some more internals to make it really flexible. Maybe even come up with a specific API for embedding.

> and also to create a feature request on the SDL bug tracker to allow changes during window runtime.

This is a limitation of the native APIs for creating a graphics context (at least wgl and XGL, not sure about others), so unfortunately not something SDL can do anything about :(

cra0zy:
> This is a limitation of the native APIs for creating a graphics context (at least wgl and XGL, not sure about others), so unfortunately not something SDL can do anything about :(

Yikes, did not know that :/

dellis1972:
@tomspilman @KonajuGames  can we kick off a build for this again. I think the bot was messed up when this PR came in

Jjagg:
Note that the SDL window actually needs to be recreated when backbuffer depthbuffer format (and some other settings) is changed after the initial window creation (see my comment above). Though this PR at least fixes the issue when depthbuffer format is set in the Game constructor.

tomspilman:
@dellis1972 - Done!

KonajuGames:
Closing in favour of #5468.  I'll add the unit tests in a separate PR.

PR #5448, opened 2017-01-31
Reverted PR #2481, double pitch.

Andreyul:
PR #2481 was a workaround for Xamarin.iOS bug, [Failed to get buffer attributes error when playing sounds](https://bugzilla.xamarin.com/show_bug.cgi?id=11074).
The bug was marked as resolved at 2017-01-30 19:54 UTC.
I tested this on iPad and could not reproduce the bug anymore.
I believe that PR #2481 can be reverted now.


KonajuGames:
Thanks @Andreyul. Merging.

PR #5451, opened 2017-02-01
Enforce GraphicsProfile.Reach + tests

nkast:


nkast:
Something strange is going on with DrawInstancedPrimitives.
Under XNA it doesn't complain about MaxPrimitiveCount (65535).
Also instancing is not supported under DX9.1 but Reach profile supports it!

Could it be that instancing is emulated on software? or Reach requirement are above the standard 9.1?

tomspilman:
> Also instancing is not supported under DX9.1 but Reach profile supports it!

Hum...  I have no clue why that is.  I wouldn't expect it to try to emulate it... but I guess it could just do an inner loop doing multiple draw calls as a fallback.

Jjagg:
I think this is moving in the wrong direction. Doesn't it make more sense to have each profile define the capabilities, allow users to granulary change those capabilities, when a device is created find an adapter that supports those capabilities (by building a GraphicsCapabilities instance by checking GL/GLES entrypoints or the D3DDevice, then comparing to the required caps) and to enforce it check the GraphicsCapability instead of the profile. I mean, the GraphicsProfile should just be a predetermined set of capabilities for easily setting some common collection of caps for a certain level of hardware, right?

Another thing is I believe we should create the highest level possible D3D device (or at least allow it for actual releases) because rendering can be faster when a higher feature level is used.

nkast:
>  I think this is moving in the wrong direction. [..]  I mean, the GraphicsProfile should just be a predetermined set of capabilities for easily setting some common collection of caps for a certain level of hardware, right?

Profile is not the same as a FeatureLevel.  DX levels ensure a set of capabilities is available. 
Profile makes this set an absolute upper limit, the programmer can't use any optional extra features supported by his card on this FeatureLevel. This ensures that a game developed on a high-end dx11 card will run on a 9.1 mobile/tablet/old laptop.
For DirectX feature level and capabilities are the same, a 'predetermined set of capabilities'. On GL we have to determine the level by the capabilities, most card follow the trend of DX feature levels so this will not be a problem. 
Although the discussion is still open on #1941, we might as well drop Profiles, let's say we throw an exception when HiDef is set and force the programmer to take a different path.

As for the way I implemented it, I could set a `MaxTexture2DSize` in graphics capabilities and check 
that instead of ` if (graphicsDevice.GraphicsProfile == GraphicsProfile.Reach && (width > 2048 || height > 2048))`. But I don't want to enforce HiDef in this PR. I don't know if we want to do that before we introduce a new profile or something.

> Another thing is I believe we should create the highest level possible D3D device (or at least allow it for > actual releases) because rendering can be faster when a higher feature level is used.

That's also suggested by one of the creators of XNA, a level 10 or 11 device will be faster than 9.3 device,  but haven't verified it. There are two paths we can follow: 
1) Create just a 9.1 device for Reach, 10.0 for HiDef, and depend on DX to fail any time the programmer uses features above that profile.
2) Create the device with the highest level available but validate all code that's incompatible with the selected profile/Feature-Level.

Until we figure out all the details of a Profile and how to enforce it I suggest to go with 1) for now. Currently Reach creates 9.1, 9.2 or 9.3.

Jjagg:
> Profile is not the same as a FeatureLevel

That's not what I meant

> I mean, the GraphicsProfile should just be a predetermined set of capabilities for easily setting some common collection of caps for a certain level of hardware, right?

I mean we should define GraphicsProfile so it is just a set of caps that map to a common hardware class (for example the caps of a d3d feature level). And that should be **all** we use GraphicsProfile for. For actually checking if the required stuff is supported, we should compare the set caps with actual device capabilities. For OpenGL this means checking that entry points are there for everything needed for achieving the set caps. For DX this means finding the lowest feature level that satisfies the set caps and checking if that feature level is supported.

This is following the idea of having GraphicsProfile for broad features and caps for more specific stuff as it was suggested in #1941.

nkast:
That will break how GraphicsProfile works. If we were to allow optional features we have to introduce a new unrestricted Profile and make the graphics capabilities public. 

IMO this will complicate things for the user. Most cards follow the DX feature sets and any optional features will get refined and become part of the next DX. The next DX might also make that feature obsolete/inaccessible as we have seen with SM3/HiDef.

Jjagg:
> That will break how GraphicsProfile works. If we were to allow optional features we have to introduce a new unrestricted Profile and make the graphics capabilities public.

Well yeah, GraphicsCapabilities would need to be made public. But how would this break current behavior?  If you just set the profile and not override the caps things will work just like before.

> IMO this will complicate things for the user.

The GraphicsProfiles are there specifically for user convenience. Like I said, you don't have to touch the caps if you just want to use a certain profile.

Here's what Tom said in #1941 (that was a while ago, so not sure if that's still his opinion):
> A two tiered system where we mix caps and feature levels could work.
We expand on the GraphicsCapabilities class we have adding new flags. Eventually we make this class public. Then at initialization MonoGame would figure out the correct GraphicsDevice feature level by examining the GraphicsCapabilities settings.

Which is what I'm arguing for here.

nkast:
> Well yeah, GraphicsCapabilities would need to be made public. But how would this break current behavior? If you just set the profile and not override the caps things will work just like before.

You mean something like a user-defined custom Profile ?

Jjagg:
No, I mean one of the GraphicsProfiles that MonoGame defines (it would include Reach and HiDef for backwards compatibility of course). So we'd define each member of the GraphicsProfile enum to map to a certain set of caps.

nkast:
On DX feature levels and capabilities are the same. There are a couple of optional features but we risk a situation similar to HiDef/SM3 if we were to include those in a profile. 

On GL we have a chicken-egg problem, should we have a per platform GraphicsCapabilities or just test for profile in Adapter.IsProfileSupported and have a common GraphicsCapabilities for all platforms? Also should we validate against profile or GraphicsCapabilities internally? That's really a design decision on how we initialise those things.

nkast:
Maybe I didn't quite understand the point you are making. What changes would you do in this PR for example?

Jjagg:
> Maybe I didn't quite understand the point you are making. What changes would you do in this PR for example?

I think I haven't been really clear about that, sorry :p Let me try to clarify. 
PS: I ended up writing down an entire explanation of how I think this system should work, so I mention a lot of things I haven't talked about before in this thread.

In the following, when I say profile I mean GraphicsProfile and when I say caps I mean GraphicsCapabilities.

In short my suggestion is to replace (in terms of functionality) what currently is GraphicsProfile in GraphicsDeviceManager by a GraphicsCapabilities property, let's call it 'RequiredCaps'. The GraphicsProfile property will stay available for convenience and backwards compatibility, but there'll be no _graphicsProfile field in GraphicsDeviceManager. It will just act as a fast way to set the RequiredCaps. The setter will do something like `_requiredCaps = MapProfileToCaps(value);`.

As for getting GraphicsProfile, I suggest we store a GraphicsProfile field in GraphicsCapabilities. Not all caps will map to a profile anymore, so I suggest we add a 'Custom' GraphicsProfile. When a profile is set it sets the corresponding properties in caps and also sets the profile GraphicsProfile field. When a setter for any property in GraphicsCapabilities is called it sets a dirty flag (to flag that it needs checking when GraphicsDeviceManager.ApplyChanges is called) and sets the profile to 'Custom'. The GraphicsProfile getter will do `return _requiredCaps.GraphicsProfile;`.

When a device is created by GraphicsDeviceManager it tries to create a device that supports the RequiredCaps. For OpenGL that means checking entry points. For DX that means getting the minimum required feature level and check if it's supported. (For a fast check DX could map the GraphicsProfile in the GraphicsCapabilities to a feature level if it is not 'Custom'). The requiredCaps passed to GraphicsDevice for its creation are set as its capabilities (so it's not its actual capabilities, but a subset of them). Internally we'd check everything against those GraphicsCapabilities of a device.


I hope that sounds like a good idea to you too :p 
For this PR that would mean all the checks would need to be done against GraphicsCapabilities.

> There are a couple of optional features but we risk a situation similar to HiDef/SM3 if we were to include those in a profile.

In #1941 it was suggested to make GraphicsProfile capabilities map less or more to D3D feature levels. If users mess with the GraphicsCapabilities it's their own responsibility to know what feature level is required and what hardware they do not support.

nkast:
> For this PR that would mean all the checks would need to be done against GraphicsCapabilities.

I would like that too but that will enforce limits on all profiles. I wanted this PR to leave HiDef unaffected. We can fix this in a future that enforces HiDef.

Jjagg:
> I would like that too but that will enforce limits on all profiles. I wanted this PR to leave HiDef unaffected. We can fix this in a future that enforces HiDef.

Okay, I get it know. Sorry for being dense. I agree, until we have a profile that supports all MG features let's not change HiDef behaviour.

nkast:
>  I agree, until we have a profile that supports all MG features let's not change HiDef behaviour.

Sure, I am expecting to get enough heat just from this PR! 
Things that previously seem to work will no longer work for Reach.

hach-que:
@nkast Instancing is supported under Reach (and XBox) in the original XNA with, and I quote, ["cunning magic (powered by unicorns)"](https://blogs.msdn.microsoft.com/shawnhar/2010/06/17/drawinstancedprimitives-in-xna-game-studio-4-0/).

nkast:
What kind of unicorns? African or european? :simple_smile:

nkast:
Well, that was helpful. 
Hargreaves answered this same question in the comments. When instancing is not supported, XNA issues a draw call for each instance.

tomspilman:
> When instancing is not supported, XNA issues a draw call for each instance.

Interesting... I wouldn't have expected that.

nkast:
> ... but I guess it could just do an inner loop doing multiple draw calls as a fallback.
> Interesting... I wouldn't have expected that.
:sunglasses:

nkast:
Is that something we can emulate? Some mobile GL platforms can not read back from Buffers, right?
XNA on WP7 didn't support it either.

KonajuGames:
OpenGL ES 2.0 does not have APIs to read back from GPU. Support for GetData
are platform-specific hacks at the moment. Unity supports getting data on
those platforms by keeping a copy in system memory.

nkast:
> OpenGL ES 2.0 does not have APIs to read back from GPU. Support for GetData are platform-specific hacks at the moment.

I guess part of the magic is to read back the transform matrices from the instanceVertexBuffer ...

nkast:
I saw a couple of reports lately about instancing failing under the Reach profile.
Since MG doesn't emulate Instancing, I added a validation check to `DrawInstancedPrimitives()`.

nkast:
Is there any chance to see this merged?

tomspilman:
> Is there any chance to see this merged?

I'm good with merging this.

I am a little concerned about how much validation is going into some functions and how we can both have support for validation to help debug issues for the developer and not waste performance when you need it.  But that is for another PR.

@Jjagg @dellis1972 @cra0zy ?

nkast:
This has to be an OpenGL bug.
test failed here https://github.com/nkast/MonoGame/blob/46da40c11e61bc847e197ac6bc26ceb931a480dc/Test/Framework/Graphics/ProfileReachTest.cs#L279-L283
with 
```
##teamcity[testStarted name='MaximumPrimitivesPerDrawCall("DrawUserIndexedPrimitives",0,65535)' captureStandardOutput='true']
[12:18:45][NAnt output]      [exec] 
[12:18:45][NAnt output]      [exec] Unhandled Exception:
[12:18:45][NAnt output]      [exec] System.NullReferenceException: Object reference not set to an instance of an object
[12:18:45][NAnt output]      [exec]   at Microsoft.Xna.Framework.Graphics.IndexBuffer.Dispose (System.Boolean disposing) [0x00015] in <6ad68231ba1a4167b285c2410144071e>:0 
[12:18:45][NAnt output]      [exec]   at Microsoft.Xna.Framework.Graphics.GraphicsResource.Finalize () [0x00002] in <6ad68231ba1a4167b285c2410144071e>:0 
[12:18:45][NAnt output]      [exec] [ERROR] FATAL UNHANDLED EXCEPTION: System.NullReferenceException: Object reference not set to an instance of an object
[12:18:45][NAnt output]      [exec]   at Microsoft.Xna.Framework.Graphics.IndexBuffer.Dispose (System.Boolean disposing) [0x00015] in <6ad68231ba1a4167b285c2410144071e>:0 
[12:18:45][NAnt output]      [exec]   at Microsoft.Xna.Framework.Graphics.GraphicsResource.Finalize () [0x00002] in <6ad68231ba1a4167b285c2410144071e>:0 
[12:18:45][NAnt output]      [exec] AL lib: (EE) alc_cleanup: 1 device not closed
```

nkast:
This is a problem with OpenGL IndexBuffer.Dispose() and the buffers use internally by GraphicDevice to draw primitives.

```
        protected override void Dispose(bool disposing)
        {
            if (!IsDisposed)
            {
                GraphicsDevice.DisposeBuffer(ibo);
            }
            base.Dispose(disposing);
        }
```

When the buffer is collected by the GC and [the destructor calls Dispose(false)](https://github.com/MonoGame/MonoGame/blob/aa98ee62c895d5dd4c27e448c6fd6d89427e8247/MonoGame.Framework/Graphics/GraphicsResource.cs#L64)  
the `GraphicsDevice` is already null.

tomspilman:
@nkast

> the `GraphicsDevice` is already null.
> It can happen when a test put pressure on the GC.

I forgot...  the finalizer nulls fields before finalizing an object?  Or is it null because the GraphicsDevice itself was disposed and finalized first?

If the `GraphicsDevice` is only null when it was finalized first then we just need a null check.

Otherwise it seems we need to rethink how this graphics resource is cleaned up.

@cra0zy @dellis1972 @Jjagg ?

nkast:
Moved the bug to https://github.com/MonoGame/MonoGame/pull/6352.

Now DesktopGL fails to create NormalizedByte2 & NormalizedByte4.  FYI, @dellis1972 .
I though those two formats where supported on cards since DX9.1/SM2.0.

tomspilman:
Bump @dellis1972  !

nkast:
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt
Here it says that texture_norm16 is 16Bits/component.

The NormalizedByte4/NormalizedByte2 is 8bits/per component.
https://github.com/MonoGame/MonoGame/commit/b814336d78ead2b44f1aa3848a6dcfaa369f7e1f#diff-3180fa168f28bfb04b185bfb620361feR667
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/PackedVector/NormalizedByte4.cs#L10

nkast:
@dellis1972 , I porsted a PR (#6372) that removes the 'supportsNormalized' check for NormalizedByte4/NormalizedByte2 textures as it was prior to  
I assume those are supported on all GL 2.0 cards as they were available on the Reach profile (DX 9.1 feature level). I don't think there is an Extension check for them.

nkast:
@dellis1972 , I posted a PR (#6372) that removes the 'supportsNormalized' check for NormalizedByte4/NormalizedByte2 textures as it was prior to  
I assume those are supported on all GL 2.0 cards as they were available on the Reach profile (DX 9.1 feature level). I don't think there is an Extension check for them.

nkast:
@dellis1972 , I posted a PR (#6372) that removes the 'supportsNormalized' check for NormalizedByte4/NormalizedByte2 textures as it was prior to #6264.
I assume those are supported on all GL 2.0 cards as they were available on the Reach profile (DX 9.1 feature level). I don't think there is an Extension check for them.

nkast:
Also, can someone look at #6352? It ready to go now.

nkast:
I still have problems with  NormalizedByte textures, creating one throws a GL error.
I am just going to treat them as not currently implemented/supported under openGL.

nkast:
This is good to go. 

@tomspilman @Jjagg @dellis1972 @cra0zy ?

PR #5452, opened 2017-02-01
Implemention of GraphicsContext

nkast:
Related to #4571.
Note that OpenGL allready had part of GraphicsContext as internal class.

As a proof of concept I moved BlendState.

in GraphicsDevice, a call to this.PlatformApplyX() becomes this.Context.ApplyX().
The code from GraphicsDevice.PlatformApplyX() is moved to GraphicsContext.PlatformApplyX().
Common code and check for Dirty flags is moved to GraphicsContext.ApplyX().

Thoughts?


Jjagg:
Nice! This is looking good! :) 
GraphicsContext will be exposed when everything is moved over, right? Might be good to document public methods/properties from the start.

KonajuGames:
> Might be good to document public methods/properties from the start

Please do. Easier to do it now than go back later.

nkast:
@Jjagg 
> GraphicsContext will be exposed when everything is moved over, right?

GraphicsDevice will still act as proxy for Context operations.
By wrapping the native Context into `GraphicsContext` we can have different Context per thread.

More on this on my previous comment https://github.com/MonoGame/MonoGame/issues/4571#issuecomment-273163376

DDReaper:
@nkast did you mean to close this PR or has it been superseded by another?

nkast:
It had many conflicts I I don't have the time to keep it up to date. Keeping all the different platforms is a real headache. It had much more work to be done, moving all the states in GraphicsContext which I was hopping to be done later on a one-by-one basis and having other to help out with it.
However I still think that MG should move all implementations under a common GraphicsContext.

PR #5456, opened 2017-02-03
SpriteBatch Begin/End. Do not Multiply with Identity when transformMatrix is null

nkast:
Begin/End|Develop|this PR|difference
------------ | ------------- | ------------- | -------------
(ms):|0.32165|0.26904|83.6%

test code: 
```
    // test Begin/End
    {
        for (int i = 0; i < 1000; i++)
        {
            spriteBatch.Begin();
            spriteBatch.End();
        }
    }
```

Begin()/End() is about +20% faster when used without a custom transformMatrix.

There were no noticable changes when used with a custom transformMatrix. We already check for null and assign a matrix (struct copy) inside Begin(). I assume nullables are just a ref copy.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Xna.Framework.Graphics
 	    readonly EffectParameter _matrixTransform;
         readonly EffectPass _spritePass;
 
-		Matrix _matrix;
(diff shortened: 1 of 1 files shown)


Jjagg:
All these optimizations you're doing for SpriteBatch are really impressive! Nice one :)

tomspilman:
Thanks @nkast !

MichaelDePiazzi:
Nice work @nkast :)

PR #5457, opened 2017-02-03
Bump Pipeline tool on MacOS to use v4.6.1.

dellis1972:
This Pr fixes a few issues. Firstly we bump Mac Pipeline tool to target v4.6.1 of mono. This is so that we don't get any build errors about System.Runtime. 
Secondly we include some missing files for the MGCB which we package with the Tool. namely the native libraries such as libnvtt.dylib. 
Finally we create a symlink from the Current mono to 4.6.2. Currently the build bot is on stable (4.6.2) and there is a hard dependency somewhere (i've not found it yet) on that mono. So the easiest solution for now is the use a symlink. 
Once we upgrade the build bot to 4.8.0 (when it goes stable) we will need to revisit this.

### DIFF
--- a/Build/Projects/Pipeline.definition
+++ b/Build/Projects/Pipeline.definition
@@ -28,6 +28,9 @@
       <Platform Name="MacOS">TRACE;MONOMAC;GTK2</Platform>
       <Platform Name="Linux">TRACE;LINUX;GTK3</Platform>
     </CustomDefinitions>
+    <FrameworkVersions>
+      <Platform Name="MacOS"><Version>v4.6.1</Version></Platform>
+    </FrameworkVersions>
   </Properties>
 
   <Files>
--- a/Build/Projects/Pipeline.definition
+++ b/Build/Projects/Pipeline.definition
@@ -31,6 +31,7 @@
     <FrameworkVersions>
       <Platform Name="MacOS"><Version>v4.6.1</Version></Platform>
     </FrameworkVersions>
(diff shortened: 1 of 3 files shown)


dellis1972:
Hmm Thought this might fix the crash on build .. it didn't :/

dellis1972:
@cra0zy @Jjagg @tomspilman @KonajuGames can some of you test the Pipeline.MacOS.png please.

The problem was that it seems our stuff has a hard dependency on mono 4.6.2.. (which I assume is on the Mac) people running mono 4.8.0 (in Beta) do not have the right symlinks in place. 

I don't know why Xamarin.Mac has a hard dependency on 4.6.2. But this should at least get this working enough for the next release.

dellis1972:
@tomspilman @KonajuGames when merging it might be good to start using the Squash and Merge option. As it squashes all commits into one in develop which looks nicer.

tomspilman:
> when merging it might be good to start using the Squash and Merge option

I selectively do that depending on how messy the PR is.  In particular if there are a lot of "fixing the build" type commits or lots of intermediate state that is useless.

If the PR is small with a few well commented commits and no "mistakes" then I tend to keep them and not squash.

dellis1972:
@KonajuGames this PR seems to work for me. It should be good to go

KonajuGames:
Merging.

cra0zy:
http://community.monogame.net/t/content-pipeline-gui-tool-3-6-on-os-x-crashes-on-build-rebuild-clean/8850

PR #5458, opened 2017-02-05
SpriteBatch optimisation: Only calculate projection matrix when viewport changes

MichaelDePiazzi:
I was inspired by PR #5456 by @nkast and came up with another related optimisation :)

This PR only re-calculates the projection matrix when the viewport changes. Given that the viewport will normally rarely change, this results in a not bad performance boost (takes about 16% less time). It's not quite as good a boost as PR #5456, but decent nonetheless :)

Benchmarks (using BenchmarkDotNet):
```c#
spriteBatch.Begin();
spriteBatch.End();
```
Benchmark | Median time | Scale<br>(Original baseline) | Scale<br>(PR #5456 baseline)
:--- | ---:|:---:|:---:
Original code (before PR #5456) | 383.2844 ns | 1.00 | 1.28
After PR #5456 | 299.4846 ns | 0.78 | 1.00
This PR | 253.0242 ns | 0.66 | **0.84**

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -28,6 +28,8 @@ namespace Microsoft.Xna.Framework.Graphics
         readonly EffectPass _spritePass;
 
 		Matrix? _matrix;
+	    private Viewport _lastViewport;
+	    private Matrix _projection;
 		Rectangle _tempRect = new Rectangle (0,0,0,0);
 		Vector2 _texCoordTL = new Vector2 (0,0);
 		Vector2 _texCoordBR = new Vector2 (0,0);
@@ -126,30 +128,36 @@ namespace Microsoft.Xna.Framework.Graphics
 			gd.RasterizerState = _rasterizerState;
 			gd.SamplerStates[0] = _samplerState;
 			
-            // Setup the default sprite effect.
 			var vp = gd.Viewport;
-
-		    Matrix projection;
-
-            // Normal 3D cameras look into the -z direction (z = 1 is in font of z = 0). The
-            // sprite batch layer depth is the opposite (z = 0 is in front of z = 1).
-            // --> We get the correct matrix with near plane 0 and far plane -1.
-            Matrix.CreateOrthographicOffCenter(0, vp.Width, vp.Height, 0, 0, -1, out projection);
-
-            // Some platforms require a half pixel offset to match DX.
-            if (NeedsHalfPixelOffset)
+            if ((vp.Width != _lastViewport.Width) || (vp.Height != _lastViewport.Height))
             {
-                projection.M41 += -0.5f * projection.M11;
-                projection.M42 += -0.5f * projection.M22;
+                // Normal 3D cameras look into the -z direction (z = 1 is in font of z = 0). The
+                // sprite batch layer depth is the opposite (z = 0 is in front of z = 1).
+                // --> We get the correct matrix with near plane 0 and far plane -1.
+                Matrix.CreateOrthographicOffCenter(0, vp.Width, vp.Height, 0, 0, -1, out _projection);
+
+                // Some platforms require a half pixel offset to match DX.
+                if (NeedsHalfPixelOffset)
+                {
+                    _projection.M41 += -0.5f * _projection.M11;
+                    _projection.M42 += -0.5f * _projection.M22;
+                }
+
+                _lastViewport = vp;
             }
 
             if (_matrix.HasValue)
             {
                 var transformMatrix = _matrix.GetValueOrDefault();
-                Matrix.Multiply(ref transformMatrix, ref projection, out projection);
+                Matrix projection;
+                Matrix.Multiply(ref transformMatrix, ref _projection, out projection);
+                _matrixTransform.SetValue(projection);
+            }
+            else
+            {
+                _matrixTransform.SetValue(_projection);
             }
 
-            _matrixTransform.SetValue(projection);
             _spritePass.Apply();
 		}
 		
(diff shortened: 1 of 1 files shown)


nkast:
Good one @MichaelDePiazzi !

We can also unroll the `CreateOrthographicOffCenter`. Because half of the inputs is 0, some divisions will always equal 1 or -1. 
BWT ,  Do we need all those doubles in `CreateOrthographicOffCenter()`?

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -126,30 +128,36 @@ void Setup()
...
-                projection.M42 += -0.5f * projection.M22;
+                _lastViewport = vp;
             }
 
             if (_matrix.HasValue)
             {
                 var transformMatrix = _matrix.GetValueOrDefault();
-                Matrix.Multiply(ref transformMatrix, ref projection, out projection);
+                Matrix projection;
+                Matrix.Multiply(ref transformMatrix, ref _projection, out projection);
```
nkast: Can you try something like `_matrixTransform.SetValue(_matrix.GetValueOrDefault() * _projection);` here?
With the last PR I introduce the tmp `transformMatrix`. Multiply(ref,ref,out) was still faster, but now that `projection` has moved down here and it's uninitialized we might not have reason to still use the re,ref,out.

nkast: Nuh, it's still slower.
But assuming that most value of the _projection/OrthographicOffCenter are either 0 or 1, we can unroll it too and remove redundant operations.

MichaelDePiazzi:
@nkast I'm not sure if there'll be much advantage to unrolling `CreateOrthographicOffCenter` as it's now called rarely anyway (only when the viewport size changes). There may be some advantage to unrolling the `Matrix.Multiply` though. I will look into them both and see what I can do in regards to unrolling. Thanks! :)

nkast:
Yes , you are right! the CreateOrthographicOffCenter is almost never called now. And actually the cases of calling Begin/End frequently are very rare. But #5456 was a simple change so why not.

MichaelDePiazzi:
@nkast Well, I did successfully unroll both the operations and it was a bit faster. `CreateOrthographicOffCenter` went from ~35ns to ~2ns, and `Matrix.Multiply` went from ~69ns to ~22ns. But `CreateOrthographicOffCenter` is now rarely called, so that one may not be worthwhile as it will make it more difficult to maintain.

But I think it might be best to submit this in a separate PR as it will make the code more complex and more difficult to review (so less likely to get merged quickly). And if you're saying there's some uncertainty with the Z values, then it's probably best to ensure that it's right before we unroll it.

MichaelDePiazzi:
P.S. This is the unrolled matrix operations I came up with fwiw:

Setup in constructor:
```c#
// Set static values from Matrix.CreateOrthographicOffCenter(0, Width, Height, 0, 0, -1)
// Other values are either 0, or dynamic
_projection.M33 = 1f;
_projection.M41 = -1f;
_projection.M42 = 1f;
_projection.M44 = 1f;
```
`SpriteBatch.End`:
```c#
// Optimised unrolled version of: Matrix.CreateOrthographicOffCenter(0, vp.Width, vp.Height, 0, 0, -1);
// Only dynamic values are re-calculated. Static values were already set in the constructor.
_projection.M11 = 2f / vp.Width;
_projection.M22 = 2f / -vp.Height;

// Some platforms require a half pixel offset to match DX.
if (NeedsHalfPixelOffset)
{
    _projection.M41 = -1f - (0.5f * _projection.M11);
    _projection.M42 = 1f - (0.5f * _projection.M22);
}
```
```c#
// Optimised unrolled version of: Matrix.Multiply(ref transformMatrix, ref _projection, out projection);
// Because there are mostly static values in the _projection matrix, several multiplies can be skipped
Matrix projection;
projection.M11 = (transformMatrix.M11 * _projection.M11) + (transformMatrix.M14 * _projection.M41);
projection.M12 = (transformMatrix.M12 * _projection.M22) + (transformMatrix.M14 * _projection.M42);
projection.M13 = transformMatrix.M13;
projection.M14 = transformMatrix.M14;
projection.M21 = (transformMatrix.M21 * _projection.M11) + (transformMatrix.M24 * _projection.M41);
projection.M22 = (transformMatrix.M22 * _projection.M22) + (transformMatrix.M24 * _projection.M42);
projection.M23 = transformMatrix.M23;
projection.M24 = transformMatrix.M24;
projection.M31 = (transformMatrix.M31 * _projection.M11) + (transformMatrix.M34 * _projection.M41);
projection.M32 = (transformMatrix.M32 * _projection.M22) + (transformMatrix.M34 * _projection.M42);
projection.M33 = transformMatrix.M33;
projection.M34 = transformMatrix.M34;
projection.M41 = (transformMatrix.M41 * _projection.M11) + (transformMatrix.M44 * _projection.M41);
projection.M42 = (transformMatrix.M42 * _projection.M22) + (transformMatrix.M44 * _projection.M42);
projection.M43 = transformMatrix.M43;
projection.M44 = transformMatrix.M44;
```

tomspilman:
> And actually the cases of calling Begin/End frequently are very rare. 

Actually I've seen plenty of games in porting that call `Begin/End` a lot more than they should.  This should be a great optimization for those titles.

> I'm not sure if there'll be much advantage to unrolling CreateOrthographicOffCenter 

I agree...  you would have to have some crazy game code that constantly changes viewport size during a frame for that optimization to matter.  I've yet to see a game ever do this... at most it is a couple of viewport changes a frame for UI stuff.

So is this ready to merge @MichaelDePiazzi @nkast ?

MichaelDePiazzi:
@tomspilman I've just pushed 1 more commit which simplifies the matrix multiplication (as @nkast originally suggested). In my benchmarks I actually found it to be a tiny bit faster, but the main benefit is that it simplifies the code. So after this builds, it should be ready to merge.

tomspilman:
Thanks @MichaelDePiazzi  and @nkast !

PR #5459, opened 2017-02-05
Add support for /define:foo=bar

dellis1972:
The documentation has an example of this and oddly the code
was in implemented. So I figured we should probably support it :)


dellis1972:
@tomspilman @KonajuGames 

For reference https://github.com/MonoGame/MonoGame/blob/develop/Documentation/mgcb.md#preprocessor-macros

KonajuGames:
Merging.  Thanks @dellis1972

Issue #4699, opened 2016-04-01
Add Unit Tests for Model Classes


tomspilman:
We need to add some unit tests which cover the following classes:

 - Model
 - ModelMeshPart
 - ModelEffectCollection
 - ModelMesh
 - ModelBone
 - ModelMeshPartCollection
 - ModelBoneCollection
 - ModelMeshCollection

A simple test that renders a static model and a skinned model should be enough to raise the code coverage on these types from 0% to around 90%.

tomspilman:
Bump!  Almost a year later and no one has taken these on for testing. :(

siudeks:
I have an experience with C# and unit testing, let me worm up with this one :)

Jjagg:
@siudeks, If you need any pointers regarding the API or any internals of MG feel free to send me a message on gitter or the community forums, glad to help out wherever :)

siudeks:
Hi guys, i'm trying to create for scratch a very simple example - 3d box using Model class. The goal is use it under unit tests.
Before reinventing the wheel - maybe you have a simple example of how to create simple 3d model with Model class programatically?

siudeks:
Ignore my prev request - I just created a 3D box in blender, exported them and checking it's Model structure to reimplement them manually for testing purpose.

siudeks:
@tomspilman new pull request ready for review.

jcawe:
Hi, is the issue solved with the commit of @siudeks ?

_Sent from my WIKO U FEEL using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

siudeks:
Guys, last time when I was working on code coverage, my personal issue was lack of tools to measure code coverage. Visual Studio does not provide cc numbers (excluding Enterprise edition). How do you test cc in project?

Jjagg:
We don't have anything set up, but would be nice to hook code coverage into the tests for local testing as well as on the build bots.

PR #5461, opened 2017-02-06
Increasing code coverage for Model.

siudeks:
Work in progress for issue #4699

### DIFF
--- a/MonoGame.Framework/Graphics/Model.cs
+++ b/MonoGame.Framework/Graphics/Model.cs
@@ -53,6 +53,15 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <param name="graphicsDevice">A valid reference to <see cref="GraphicsDevice"/>.</param>
         /// <param name="bones">The collection of bones.</param>
         /// <param name="meshes">The collection of meshes.</param>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="graphicsDevice"/> is null.
+        /// </exception>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="bones"/> is null.
+        /// </exception>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="meshes"/> is null.
+        /// </exception>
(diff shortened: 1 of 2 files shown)


tomspilman:
This is great @siudeks !

siudeks:
Goood :) I'll finish the rest of requested unit tests in my spare time. Feel free to reject / comment in any case of doubts

Jjagg:
@siudeks Could you add the ModelTest.cs file to the Protobuild definitions? That will make Protobuild include the file in the generated project files (.csproj) when it generates them. See how the others are added here: https://github.com/MonoGame/MonoGame/blob/develop/Build/Projects/MonoGame.Tests.definition#L143

It would also be nice if you could include ModelTest.cs in the XNA test project file (which is not generated by Protobuild). If you have XNA installed you can run the tests in that project file to check for XNA compatibility which is nice for more intricate stuff.

siudeks:
Will investigate this issues and do till the end of week.

siudeks:
Hey. I investigated scope of changes requested by @Jjagg. Updating Protobuild definition files is a very simple task and I see the point. More interesing is updating MonoGame.Tests.XNA with new tests.

Original XNA Model class doesn't have public constructor but MG Model class contains it. I use the ctor in my tests.  In the clash of XNA vs MG implementation I see that the best approach is to rewrite my tests to make them compatible with both: XNA and MG, so - in that particular case - means to avoid public Model constructors.

Any thougths? @Jjagg ? @tomspilman ? Did you notice this kind of compatibility issue in project? what is preferred approach?

Jjagg:
You can use preprocessor directives to ignore code that does not compile for XNA if you think it's worth testing in MonoGame. If you can write tests so they work for both xna and MG that's a plus because it tests compatibility.

#if !XNA
... // test goes here
#endif

siudeks:
Sadly speaking, for Model test(s) seems better to use !XNA because - at the first look - XNA supports loading models only from predefined assets what is far from pure unit tests ...

PR #5467, opened 2017-02-09
Do not center window after Position is set

### LINKED ISSUE #5460: Window.IsBorderless strange behavior
I just updated to the latest develop build and this happened. Here's a little test I wrote in my Game1.cs constructor:

```cs
Point scrDimensions = new Point(1366, 768);
graphics.PreferredBackBufferWidth = scrDimensions.X;
graphics.PreferredBackBufferHeight = scrDimensions.Y;
Window.IsBorderless = true;
// explicitly setting Window.Position = new Point(0, 0); did not fix the issue.
```

1366x768 is my native screen resolution. So when I boot the game I can see my taskbar overlap it at the bottom of my screen. The top of the game is cut off the screen. Is this the desired behavior? Yesterday, before I updated the game was properly set to borderless fullscreen. Tested on x64 Win10 Pro with a DirectX project. Looking at the recent merge history, could this be caused by an oversight in #5419?

Jjagg:
Fixes #5460 

When Window.Position is set the window should no longer be centered after it is resized. I also renamed the flag for clarity and moved the window centering code around a bit so it's only needed in 1 place in WinFormsGameWindow itself. 

Looking at this again made me wonder if it would be nice to allow users to reset the flag that's used here so the window is centered again after a resize. We'd change the name and move it to GameWindow for a nice API, `public bool GameWindow.CenterOnResize { get; set; }` or something like that.


cra0zy:
This is what XNA did, just tested it, need to fix DesktopGL...

KonajuGames:
I would suggest `bool GameWindow.AutoCenter { get; set; }` if we add that.

cra0zy:
Instead of just adding a new per platform property that is not accessible from PCL, can we instead implement a new way of setting these properties, like what SDL does with hints: https://wiki.libsdl.org/CategoryHints#Hints

KonajuGames:
This PR now has conflicts that need to resolved.

Jjagg:
@KonajuGames Fixed!

KonajuGames:
I've re-queued the build tests for this PR as the failed unit test appears to be unrelated.  If it fails again, could you try running the unit tests on this branch locally?

Jjagg:
Oops, made a mistake in the merge. Fixing...

Jjagg:
@KonajuGames This is good to go :)

KonajuGames:
Thanks @Jjagg. Merging.

PR #5468, opened 2017-02-10
[DesktopGL] General Fixes

cra0zy:
Stuff done:
 - Window will no longer get centered after it was moved, this mimics XNA behavior (DesktopGL version of #5467)
 - GL attributes are set before window is created (A proper CreateWindow function when compared to #5445, also I made sure to make it so it can be reused for RecreateWindow with minimal changes later on)

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -539,6 +539,9 @@
     <Compile Include="GraphicsDeviceManager.cs">
 		<Platforms>Windows,WindowsGL,Linux,Windows8,WindowsPhone81,WindowsUniversal,Web</Platforms>
     </Compile>
+    <Compile Include="GraphicsDeviceManager.SDL.cs">
+      <Platforms>WindowsGL,Linux</Platforms>
+    </Compile>
     <Compile Include="GraphicsDeviceManager.WinRT.cs">
 		<Platforms>Windows8,WindowsPhone81,WindowsUniversal</Platforms>
     </Compile>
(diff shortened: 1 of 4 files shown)


KonajuGames:
This replaces #5445?  I'll close it when confirmed.

cra0zy:
> This replaces #5445? I'll close it when confirmed.

Yours offers NUnit tests which are useful.

KonajuGames:
If those unit tests are merged into this one, I'm happy to go with this one.

Jjagg:
Just a reminder of what I said in #5445 

> #5429 makes the DesktopGL platform use the new GraphicsDeviceManager, so the changes in GraphicsDeviceManager.Legacy in this PR would need to be moved to PlatformInitialize (called here) in a new file GraphicsDeviceManager.OpenGL when #5429 is merged. Probably easier to get #5429 merged first and then make the necessary changes here.

KonajuGames:
With #5429 merged, this will need to be refactored as mentioned by @Jjagg above.

cra0zy:
@KonajuGames done.

Build/Projects/MonoGame.Framework.definition
```diff
@@ -307,6 +307,9 @@
     <Compile Include="IDrawable.cs" />
     <Compile Include="IGameComponent.cs" />
     <Compile Include="IGraphicsDeviceManager.cs" />
+    <Compile Include="IGraphicsDeviceManager.SDL.cs">
```
KonajuGames: I think that should have the "I" prefix.

KonajuGames:
The builds failed with

    CSC : error CS2001: Source file 'IGraphicsDeviceManager.SDL.cs' could not be found

The definition file lists `IGraphicsDeviceManager.SDL.cs` but the new file is `GraphicsDeviceManager.SDL.cs`.

KonajuGames:
It will take a little while for builds and tests to run.  I'll look back on this PR in the morning (almost 1am here).

cra0zy:
That is fine by me.

KonajuGames:
I'm still up.  Merging now.  I'll add the unit tests from #5445 in a separate PR.

PR #5471, opened 2017-02-11
Added support for PngWriter to export textures with single surface fo…

sami016:
…rmat using greyscale.

Adds the basic ability to export single channel textures into png format. Extremely useful for generating height maps to be reused later on.

### DIFF
--- a/MonoGame.Framework/Utilities/Png/PngWriter.cs
+++ b/MonoGame.Framework/Utilities/Png/PngWriter.cs
@@ -214,10 +214,12 @@ namespace MonoGame.Utilities.Png
                 case SurfaceFormat.Single:
                     var floatData = new float[colorDataLength];
                     texture2D.GetData<float>(floatData);
-                    
-                    for (int i = 0; i < colorDataLength; i++) {
+
+                    for (int i = 0; i < colorDataLength; i++)
+                    {
+                        float brightness = floatData[i];
                         // Export as a greyscale image.
-                        colorData[i] = new Color(floatData[i], floatData[i], floatData[i]);
+                        colorData[i] = new Color(brightness, brightness, brightness);
                     }
                     break;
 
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Utilities/Png/PngWriter.cs
```diff
@@ -211,6 +211,16 @@ private void GetColorData(Texture2D texture2D)
 
             switch (texture2D.Format)
             {
+                case SurfaceFormat.Single:
+                    var floatData = new float[colorDataLength];
+                    texture2D.GetData<float>(floatData);
+                    
+                    for (int i = 0; i < colorDataLength; i++) {
+                        // Export as a greyscale image.
+                        colorData[i] = new Color(floatData[i], floatData[i], floatData[i]);
```
KonajuGames: I don't know if the compiler is smart enough to optimize this down to one array access, but to avoid any extraneous bounds checks, read the array once into a temporary `float` and pass that float into the `Color` constructor.  Using `fixed` would make it faster still, but see if the first approach makes any improvement.

Also, our coding guidelines require the braces to be on the following line at the start of a statement block.

sami016: My latest commit addresses the problems you've pointed out. 

I've not changed to using fixed yet, just keeping it in consistent with how arrays are iterated over in the other case blocks. When I get some more time I will do some profiling and see what kind of performance increase this would yield.

sami016: I've just performed a small array manipulation benchmark summing an array of 1000000 integers, repeating this test 1000 times. The results are as follows:

normal
mean: 4.314 ms
stddev: 1.078 ms

unsafe with fixed
mean: 4.421 ms
stddev: 0.9709 ms

As you can see, there is a slight performance benefit, though I'm not sure if it's enough to justify the readability decrease. Any thoughts?

KonajuGames:
No need to go to fixed then. Thanks for testing it.

KonajuGames:
Thanks @sami016. Merging.

PR #5475, opened 2017-02-14
Get OpenGL tests running and fix the failures (recreated)

### LINKED ISSUE #1449: Texture2D GetData does not work when called from another thread

### LINKED ISSUE #4224: Texture2D.GetData<T>(T[], int, int) incorrect behaviour

### LINKED ISSUE #4708: Texture2D GetData returns Transparent colors when called from different thread

### LINKED ISSUE #5473: Texture2D.SaveAsImage only writes 1/4th of texture data (DesktopGL)

### LINKED ISSUE #5474: Texture2D.GetData<T> does not return the correct number of elements when  size of T is not the same as the pixel format. (OpenGL)

Jjagg:
#5429 for some reason was not getting builds run. It had some issues with the build server before. Hopefully just putting the changes in a new branch solves that.

- fixes #4708, fixes #1449 by throwing if GetData is not called from the main thread (SetData is allowed, but queued and executed synchronously).
- fixes #4224, this might have already been fixed before this branch actually
- fixes #5474, fixes #5473


Jjagg:
Closing because #5429 is running builds again.

Issue #5449, opened 2017-01-31
Latest dev builds: first performance drop, now NoSuitableGraphicsDeviceException?


yyr57695:
I am running MonoGame on Windows with Visual Studio 2015 Community.  I recently ported an XNA project that utilizes 3D graphics.  Under XNA it was using the HiDef profile.

Using MonoGame 3.5.1 the performance was indistinguishable from XNA, but there was a bug using the Xbox One controller.  If it was plugged in when the game was launched, it didn't work until it was unplugged, then reconnected.

I decided to try using the latest dev build to see if that bug was fixed.  I grabbed it as of a few weeks ago, version 3.6.0.1365.  At first, it didn't run at all...there were some minor changes I had to make, but the biggest issue was an error telling me I had to use 16-bit indices rather than 32-bit indices.  Huh?  I ultimately realized that it was trying to force the Reach profile, and fixed it by adding the following line in my constructor:

graphics.GraphicsProfile = GraphicsProfile.HiDef;

where "graphics" is my GraphicsDeviceManager.  This got the game running, and the controller detection bug had indeed been fixed in the new MonoGame build, but the framerate was approximately half of what it had been before.  The game looked identical, but the performance had dropped like a rock.

I was hoping that this was just a temporary bug.  I tried again with version 3.6.0.1425 just a few days ago (which is still the current version as of 1/29/2017).  But now the game won't start at all.  It builds OK, but crashes during initialization with a Microsoft.Xna.Framework.Graphics.NoSuitableGraphicsDeviceException.

I am using a 2010 MacBook sporting an integrated GeForce 320M GPU.  Yeah, an antique by today's standards, but it always ran my XNA stuff quite easily.  

Were the hardware requirements just raised?  If so, why?  If not, what could be causing this?

If there's no easy fix for this (and the performance issues), is it possible for me to Frankenstein the new Input code into MonoGame 3.5.1 and use that instead, for its superior performance?

Thanks very much for your assistance.

Jjagg:
> biggest issue was an error telling me I had to use 16-bit indices rather than 32-bit indices. Huh? I ultimately realized that it was trying to force the Reach profile

Changes have been made ik 3.6 in an effort to actually start enforcing the GraphicsProfile. Two issues here:

1. There are some cases were no decent exception messages are being thrown because of missing functionality due to GraphicsProfile limitations. (e.g. #5108, which a lot of people have run in to).
[...]

PR #5477, opened 2017-02-15
DirectX: Fixed multisample clamping logic

zigzag312:
Bigger changes are that I moved some things from `PlatformInitialize()` to `PlatformSetup()` in GraphicsDevice.DirectX.cs. I hope I didn't move too much. Someone please verify as I'm new to this project. 

The reason for this change is that `d3ddevice `is needed before so we can query device capabilities. I compared DirectX version to OpenGL version. GL extension are set inside `PlatformSetup()` and GL Extensions are in some way like d3ddevice .... they are used to query device capabilities. So I tried to mimic OpenGL version. 

Multisample count clamping logic is removed from `GraphicsDeviceManager `and added it to `GraphicsDevice.DirectX` like @Jjagg said in #5449. It should also prevent crashes when multisampling is not supported like PR #5446.


MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -130,24 +127,35 @@ private void PlatformInitialize()
...
+#endif
+#if WINDOWS_UAP
+			CreateSizeDependentResources();
+#endif
+#if WINDOWS_STOREAPP
+            CreateSizeDependentResources();
+#endif
+#if WINDOWS
+            CreateSizeDependentResources();
+#endif
```
KonajuGames: This can be cleaned up so it is a single

````
#if !WINDOWS_PHONE
        CreateSizeDependentResources();
#endif
````

**Comment by zigzag312** [comment]
Ok I will clean up those conditions and try to rebase

File: MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs [hunk 2]
```diff
@@ -827,13 +808,56 @@ internal void CreateSizeDependentResources()
...
+        
+
         internal void OnPresentationChanged()
         {
             CreateSizeDependentResources();
             ApplyRenderTargets(null);
         }
 
 #endif // WINDOWS
+        /// <returns>Higher than zero if multiSampleCount is supported.
```
Jjagg: Comment is cut off here.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -827,13 +808,56 @@ internal void CreateSizeDependentResources()
...
+        /// <returns>Higher than zero if multiSampleCount is supported. 
+        /// Zero if multiSampleCount is not supported.</returns>
+        private int GetMultiSamplingQuality(Format format, int multiSampleCount)
+        {
+            // The valid range is between zero and one less than the level returned by CheckMultisampleQualityLevels
+            // https://msdn.microsoft.com/en-us/library/windows/desktop/bb173072(v=vs.85).aspx
+            var quality = _d3dDevice.CheckMultisampleQualityLevels(format, multiSampleCount) - 1;
+            // NOTE: should we always return highest quality?
+            return Math.Max(quality, 0); // clamp minimum to 0 
+        }
```
Jjagg: Newline here.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -827,13 +808,56 @@ internal void CreateSizeDependentResources()
...
+                    msc = GraphicsCapabilities.MaxMultiSampleCount;
+
+                var quality = GetMultiSamplingQuality(format, msc);
+
+                multisampleDesc.Count = msc;
+                multisampleDesc.Quality = quality;
+            }
+
+            return multisampleDesc;
+        }
```
Jjagg: Since the clamping is platform agnostic it could be put in the shared GraphicsDevice.cs file so other platforms can reuse it. (wihtou the SampleDescription obviously)

MonoGame.Framework/GraphicsDeviceManager.cs
```diff
@@ -210,28 +210,6 @@ private GraphicsDeviceInformation DoPreparingDeviceSettings()
...
-                    msc -= (msc >> 1);
-
-                    if (GraphicsDevice != null)
-                    {
-                        // and clamp it to what the device can handle
-                        if (msc > GraphicsDevice.GraphicsCapabilities.MaxMultiSampleCount)
-                            msc = GraphicsDevice.GraphicsCapabilities.MaxMultiSampleCount;
-                    }
-                    gdi.PresentationParameters.MultiSampleCount = msc;
-                }
```
Jjagg: @zigzag312 I think you can just put this back in.

MonoGame.Framework/Graphics/GraphicsCapabilities.DirectX.cs
```diff
@@ -25,11 +25,13 @@ private void PlatformInitialize(GraphicsDevice device)
...
             SupportsDepthClamp = device.GraphicsProfile == GraphicsProfile.HiDef;
             SupportsVertexTextures = device.GraphicsProfile == GraphicsProfile.HiDef;
+
+            _maxMultiSampleCount = GetMaxMultiSampleCount(device);
         }
 
         private void PlatformInitializeAfterResources(GraphicsDevice device)
         {
-            _maxMultiSampleCount = GetMaxMultiSampleCount(device);
+
```
Jjagg: Only DX platform used this method. If you can keep this change we can remove it. It was a bit of a hack anyway.

MonoGame.Framework/GraphicsDeviceManager.cs
zigzag312: A have now added a call to clamping method inside Graphic.Reset instead. It is also called from ApplyChanges like DoPreparingDeviceSettings() and it also prevents wrong multi sample count if someone just calls Reset with new PresentationParameters.

zigzag312:
Ok I think I am getting a hang of this :)
I still need to change some things so it will behave like XNA.

KonajuGames:
> I still need to change some things so it will behave like XNA.

For sure.  Don't let the PR grow beyond scope though.  Keep it focused on the original intent and it will be easier to merge when it is complete.

zigzag312:
After doing some testing I found out some differences between MG and XNA that are not connected just to multisampling. 

If we change some parameters of `PresentationParameters `under `GraphicsDevice `and call `GraphicsDevice.Reset()` XNA will ignore those changes, but MG will apply those changes. If we call `GraphicsDevice.Reset(newPresentationParameters)` both MG and XNA will apply `newPresentationParameters`. That is why a call to `ClampMultisampleCountToSupportedByCurrentDevice()` is needed right now inside Reset() method.

With this in mind, I think this PR should be merged as is (if there aren't other issues). And if `GraphicsDevice.Reset()` is modified to behave like in XNA in the future, then `ClampMultisampleCountToSupportedByCurrentDevice()` method will be moved out of `Reset()` method to some other location.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -267,10 +267,34 @@ private void Setup()
...
+        {
+            if (PresentationParameters.MultiSampleCount > 1)
+            {
+                // Round down MultiSampleCount to the nearest power of two
+                // hack from http://stackoverflow.com/a/2681094
+                // Note: this will return an incorrect, but large value
+                // for very large numbers. That doesn't matter because
+                // the number will get clamped below anyway in this case.
+                // Also MultiSampleCount should never be larger than 32 
+                // (which is MultiSampleCountLimit)
```
Jjagg: The last two comment lines here aren't relevant anymore.

zigzag312: Removed

Jjagg:
This looks good to me. Would be nice to do some testing or add unit tests though.

> If we change some parameters of PresentationParametersunder GraphicsDeviceand call GraphicsDevice.Reset() XNA will ignore those changes, but MG will apply those changes.

Maybe the PresentationParameters getter makes a copy in XNA. IMO this isn't that big a problem.

zigzag312:
Yeah, because XNA itself doesn't handle this particularly well. I actually prefer the way MG behaves in this case.

Any idea how to do unit tests for this?

KonajuGames:
Unit tests are difficult because it depends on GPU support for
multisampling. We use the DirectX reference device for unit tests, and I
believe it would have full support for multisampling.

Jjagg:
I meant for the proper clamping, so test if you can't mess up PresentationParameters somehow. But now that you mention it tests could break if the testing device does not support multisampling. Maybe just enable tests for DX with the reference device then.

zigzag312:
Quick question. Should I split unit tests I just added? Because in each submitted test there are two (similar) tests. One for GraphicsDevice.ApplyChanges() and one for GraphicsDevice.Reset(PresentationParameters).

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -267,10 +267,39 @@ private void Setup()
             Dispose(false);
         }
 
+        void ClampMultisampleCountToSupportedByCurrentDevice()
+        {
+            PresentationParameters.MultiSampleCount =
+                GetClampedMultisampleCountToSupportedByCurrentDevice
+                    (PresentationParameters.MultiSampleCount);
+        }
+        internal int GetClampedMultisampleCountToSupportedByCurrentDevice(
```
Jjagg: @zigzag312  Can you make this private?

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -267,10 +267,39 @@ private void Setup()
             Dispose(false);
         }
 
+        void ClampMultisampleCountToSupportedByCurrentDevice()
```
Jjagg: I would inline this call. IMO it's not worth a seperate function and GraphicsDevice is cluttered enough as it is.

Jjagg:
> Should I split unit tests I just added?

I think it's fine the way it is.

The only qualm I have with this PR in the current state is with the function name: `GetClampedMultisampleCountToSupportedByCurrentDevice` :p That's really longer than it needs to be IMO I vote for `ClampMultiSampleCount`.

Great work with this @zigzag312. Someone on the forums ran into the no suitable graphics device exception http://community.monogame.net/t/run-causes-no-suitable-graphics-device-exception-failed-to-create-graphics-device-3-6-0-1625-resolved/8819/2
Let's get this merged for 3.6.1!

KonajuGames:
Excessively wordy method names should be avoided. You can safely assume it
is for the current device.

zigzag312:
Thanks! Yeah its a long name :) I'll shorten it.

MonoGame.Framework/Graphics/GraphicsDevice.cs
zigzag312: Ideally this method would be ClampPresentationParameters where we clamp all preferred values every time PresentationParameters are changed. Do we have any method like this?

MonoGame.Framework/Graphics/GraphicsDevice.cs
zigzag312: I need to check. I made some methods internal in the last refactoring so that same clamping logic is reused when creating multisampled render targets. I'm going to submit PR for that soon.

zigzag312:
I think this should be good to go now

Jjagg:
@KonajuGames @tomspilman Just a reminder that this should be in the 3.6.1 release. 
Without it DX/UWP crash when PreferMultisampling is set to true on the GraphicsDeviceManager and the GPU does not support a MS count of 32.

KonajuGames:
Merging this now.  Thanks @zigzag312

PR #5488, opened 2017-02-19
#4699 More unit tests for Model* family.

siudeks:


Test/Framework/Graphics/ModelTest.cs
```diff
@@ -74,6 +74,94 @@ public void CopyBoneTransformsTo_Exceptions()
...
+                var verticesPointers = new ushort[]
+                {
+                    0, 1, 2,
+                    0, 2, 3
+                };
+                var indexBuffer = new IndexBuffer(gd, IndexElementSize.SixteenBits, verticesPointers.Length, BufferUsage.None);
+                indexBuffer.SetData(verticesPointers);
+
+                var meshPart = new ModelMeshPart();
+                meshPart.NumVertices = vertices.Length;
```
tomspilman: I suspect this test won't compile under XNA.  This is because at some point we made the model classes have setters where XNA does not.

What do you think of this @KonajuGames @Jjagg @dellis1972 ?

KonajuGames:
It should be loading the cube from a pre-prepared .xnb file. It will help
greatly if as many tests as possible, especially in this area, can run on
XNA as well.

Jjagg:
I agree. There doesn't seem to be a way to build a model without the content pipelin in XNA. It can be done through MeshBuilder, but that still needs to be processed and there are a few bugs left in the current implementation. So an xnb seems like the way to go.

siudeks:
To sum up - do we prefer to test Model* family based on an attached xnb file ? It will make tests compatible with XNA

Jjagg:
Yes.

tomspilman:
@siudeks - Yes I agree with Jjagg.

siudeks:
Challenge accepted :)

siudeks:
Hi guys. I improved my tests to make them XNA-friendly. They are green for Windows platform, but when trying to run them in MonoGame.Tests.XNA project I have exception in **content.Load{Model}**:
_Cannot find ContentTypeReader
Microsoft.Xna.Framework.Content.ModelReader._
Any hint how to prepare xnb model to make the loadable in both Windows and XNA tests?

Jjagg:
You'll probably have to build the model using XNA's content tools. I didn't know MG modified the format, but there's other types of content where we test using different binaries (at least effects, not sure if there are other actually). So you can just include models built with the MG content pipeline for MG tests and those built by XNA for the XNA tests.

Unless we should use the exact same format XNA uses. It could just be a bug because I don't see why MG would need to make changes to the xnb format for models. @tomspilman ?

siudeks:
@Jjagg you are right - model created with XNA pipeline is friendly for both cases: XNA tests and MG tests, so problem solved ... 
Does it mean that pipeline is on way compatible XNA -> MG? Do we need to use XNA pipeline to create reference models in MG framework?

siudeks:
@tomspilman tests has been converted to XNA friendly. Please review and comment / accept

siudeks:
@tomspilman Is my pull request ok? Asking because I would like to find more from MG, but not sure if the quality of changes is accepted. Feel free to comment

tomspilman:
@siudeks - The only issue I see is the change to the dependencies submodule.  That shouldn't be in this PR... it rolls back the reference to an older revision and it doesn't seem to be relevant anyway.

@KonajuGames @dellis1972 - Can you review this too?

dellis1972:
Looks ok, the dependencies need rebasing though like @tomspilman said.

tomspilman:
@dellis1972 @KonajuGames - This PR seems like it got messed up... any easy way for @siudeks to fix it?

siudeks:
Will close the current pull request and create new one with the same changes
I'm not a git specialist so it seems to be the easiest way
Does it sound like a plan?

tomspilman:
> Does it sound like a plan?

Sure... if you are willing to do that then it would be clean.

siudeks:
will do

Issue #5492, opened 2017-02-21
[MGCB] Curve fails to build/import


zigzag312:
Curves stored in xml files, that were created using XNA Curve Editor fail to build with Monogame's Pipeline tool. Xml of the most simple `Curve `with only one `CurveKey `looks like this:
```
<?xml version="1.0" encoding="utf-8"?>
<XnaContent xmlns:Framework="Microsoft.Xna.Framework">
  <Asset Type="Framework:Curve">
    <PreLoop>Constant</PreLoop>
    <PostLoop>Constant</PostLoop>
    <Keys>0 1 0 0 Smooth</Keys>
  </Asset>
</XnaContent>
```
First problem is that `XmlImporter `in Monogame expects elements in different order. This issue is already described in #4792. I have started making a fix for this when I stumbled on a second problem: 
`CurveKeyCollection `is not getting deserialized from xml. 

Each `CurveKey `is represented by 5 values separated by space inside `<Keys>...</Keys>`. 
Order of values is: `Position`, `Value`, `TangentIn`, `TangentOut `and `Continuity`.
Each new CurveKey inside a collection is also separated by space. Two CurveKeys look like this: `<Keys>0 1 0 0 Smooth 0.5 0.5 0 0 Smooth</Keys>`

Debugging step-by-step I found the point where deserialization fails. `GenericCollectionHelper `doesn't read collection as it expects each `CurveKey `to be a separate xml element.
Here's the `Deserialize `method from `GenericCollectionHelper `:
```
 public void Deserialize(IntermediateReader input, object collection, ContentSerializerAttribute format)
        {
            var itemFormat = new ContentSerializerAttribute();
            itemFormat.ElementName = format.CollectionItemName;
            while (input.MoveToElement(format.CollectionItemName)) // this loop expect list of CurveKey elements!
                _addMethod.Invoke(collection, new[] { input.ReadObject<object>(itemFormat, _contentSerializer) });
        }
```
Does anyone know if there is another class in Monogame that would help to deserialize collection like this? And does anyone with more experience with deserialization have any idea how one would approach to solve this?

I have many curves in a project I want to migrate and changing all xml files might take me the same amount of time as fixing this code. Also there are [others ](http://community.monogame.net/t/importer-xmlimporter-had-unexpected-failure/7959)reporting the same issue.

KonajuGames:
I had no idea XNA had a curve editor tool. I've never used these curves in
XNA.

[...]

PR #5494, opened 2017-02-22
Fixed XML deserialization of Curve type

zigzag312:
Fixes issue #5492.

- Changed order of properties inside Curve class so that order in serialization/deserialization matches XNA's
- Added CurveKeyCollectionSerializer that serializes/deserializes CurveKey collection packed like XNA
- Added unit tests

### DIFF
--- a/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
+++ b/Build/Projects/MonoGame.Framework.Content.Pipeline.definition
@@ -376,7 +376,7 @@
     <Compile Include="Serialization\Intermediate\ContentTypeSerializer.cs" />
     <Compile Include="Serialization\Intermediate\ContentTypeSerializerAttribute.cs" />
     <Compile Include="Serialization\Intermediate\ContentTypeSerializerT.cs" />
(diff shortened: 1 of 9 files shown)


KonajuGames:
Changing the order of fields is not reliable because the C# compiler can re-order fields and properties if it needs to. The is an attribute to prevent re-ordering, but if the XNA class didn't have it then we shouldn't.

KonajuGames:
New source files need to be added to the Protobuild definition files in the Build directory.

zigzag312:
> New source files need to be added to the Protobuild definition files in the Build directory.
New *cs files that I added to the project?

zigzag312:
> Changing the order of fields is not reliable because the C# compiler can re-order fields and properties if it needs to. The is an attribute to prevent re-ordering, but if the XNA class didn't have it then we shouldn't.

I tried to change order of elements in xml and open it with XNA Curve Editor tool and it fails to open it. So XNA depends on order of elements inside xml. I can't find any serializer specific for Curve class inside XNA so I have no idea how they do it then.

tomspilman:
Thanks @zigzag312 ... merging!

Issue #5484, opened 2017-02-18
Request to expose Texel Width Height getter in Texture2D


willmotil:
Tom or nkast or whoever is working on this atm can we expose the Getter on those two properties.
internal float TexelWidth { get; private set; }
internal float TexelHeight { get; private set; }
to simply
public float TexelWidth { get; private set; }
public float TexelHeight { get; private set; }

This is extremely useful when exposed for all texture2d custom vertex uv element assignments.

We already have Width Height exposed so a user could keep recalculating the reciprocal for each texture himself but... It would be nice however to expose the getter on this it would be useful. Also since there is no need for a user to set the value set can remain private.

_If this is a matter of hesitation on deciding on a final name just call it by its math proper name._ 
WidthReciprocal.
HeightReciprocal.

https://www.mathsisfun.com/reciprocal.html

nkast:
> I was wondering if we could get a set of vectors in the glyph class that directly relates to the uv bounds of each glyph.

While this [originally](https://github.com/MonoGame/MonoGame/pull/5267#issuecomment-269262365) gave an impressive improvement around 30-45%, when compared with the latest develop, specifically after  #5453 I am only getting around 5%-10%. 
Test Branch [here](https://github.com/MonoGame/MonoGame/compare/develop...nkast:tnc_SpriteBatch_GlyphUV).

The issues I found so far are that on MG the Glyph class is public, so we have to be careful on how to implement this. Another is that it makes Glyph larger by 16 bytes, we should try to see if we can make it a class or access it directly from a Glyph[] array.
We must either disallow

willmotil:
Really i just wanted the texture2d reciprocal exposed since its there already and its really useful.

Its not so much that i expected a huge increase. It was a guess that a look up on the TL BR should also give a slightly minor performance boost because were already looking up the glyph for each character. Though it does fatten the glyph struct. Then again ill wager the Rectangle bounds struct contained in it alone is 4x as fat as the glyph itself. Anyways this also makes this code more readable and clear.
Maybe just store the 2 vectors TL BR since its all that's needed instead of all 4 then its 8 bytes.

nkast:
> Really i just wanted the texture2d reciprocal exposed since its there already and its really useful.

I made TexelWidth/TexelHeight into two separate properties having in mind that possibility after [a similar request](https://github.com/MonoGame/MonoGame/pull/5267#issuecomment-257422132) by @Vodacek.

[...]

Issue #5111, opened 2016-08-15
Pipeline automatic texture size reduction


tomspilman:
In some titles we've ported they are going from a console running at 1920x1080 with around 6GB of memory down to like a Vita which is basically half that resolution and much less GPU memory.  This is similar to having one game that targets PC and mobile where mobile sometimes requires smaller textures because of screen resolution and memory limits.

It would be good for the pipeline to support some feature that would automatically reduce texture sizes by 1 mip level before processing to account for these memory/resolution constrained devices.

In many games the majority of the textures would need this reduction with a few textures needing to opt-out of it.  Not sure how it would be presented to the user so that it is easy to setup.

 - Is this a per-texture setting?
 - Is this some global per-project setting?
 - Is it maybe some mixture of both?

prime31:
The way I see this being the most useful is multi-tiered cascading. Something like this:

**Texture Scale**
- global/project setting (1 by default)
- per-platform setting (if set overrides global/project setting)
- per-texture setting (if set overrides per-platform setting)

jamesford42:
It would be nice if we knew whether a property on a processor had been given a real value or if its value was actually from its defaults. (so we could know if it should instead use the 'template' value from the project, instead of its own)... be i don't think we have the information to do this.

I was like, imagining you add a special type of content item into the project, where the build action is 'template' and it doesn't have any importer, just a processor.

LithiumToast:
Posting due to the `help-wanted` tag. I found this issue when doing requirements gathering for a Content Pipeline which is independent from MonoGame. I'm calling it [Assetyard](https://github.com/Styxz/Assetyard). 

I had basically the same idea of doing "Content Templates" as @jamesford42 described

A base content template might look like:
```
{
    “name”:  “baseTemplate"
    “searchPattern": "^((?>[^\/]*\/)*)(.*)$",
    "importer": “ImporterType",
    "processor": “ProcessorType”
}
```
You can post that regular expression into something like https://regex101.com
The first capture group is the directory path, the second capture is the filename.
Then if I can “stack" content templates through inheritance:

```
 {
    “name”: “PNG Template”
    “inherits”: “baseTemplate"
    “searchPattern":  "^((?>[^\/]*\/)*)(.*png)$",
    "importer": “TextureImporter",
    "processor": “TextureProcessor”
}
```

[...]

PR #5495, opened 2017-02-22
Expose TexelWidth/TexelHeight properties in Texture2D

nkast:
#5484 , 
https://github.com/MonoGame/MonoGame/pull/5267#issuecomment-257422132

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -290,10 +290,10 @@ namespace Microsoft.Xna.Framework.Graphics
                 var srcRect = sourceRectangle.GetValueOrDefault();
                 w = srcRect.Width * scale.X;
                 h = srcRect.Height * scale.Y;
-                _texCoordTL.X = srcRect.X * texture.TexelWidth;
-                _texCoordTL.Y = srcRect.Y * texture.TexelHeight;
-                _texCoordBR.X = (srcRect.X + srcRect.Width) * texture.TexelWidth;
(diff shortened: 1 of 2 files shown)


willmotil:
oh i just tried to do this it said sync failed, i suck with github.

nkast:
It took me some time to get along with git and still use it to do just the basic stuff.
Sorry about that I didn't knew you were working too on a PR.

tomspilman:
So the thing that makes me hesitate on this is https://github.com/MonoGame/MonoGame/issues/5111.

I have a first implementation I wrote for a PSVita port.  The pipeline reduces the texture sizes at build time and stores the original texture size in XNB (no change to the format).  Then at runtime when the XNB is loaded we store them in a new property `Texture2D.OriginalWidth/Height`.  Internal to `SpriteBatch` we use these to scale the texture cords back up to the correct range.  This makes like 90% of the work of rescaling assets automatic.

Maybe we just work that out later.  Just wanted to avoid adding an API to then remove it or change it again.

Still we should wait until after 3.6 which we plan to release in a week or so.

nkast:
That's fine, let's see how this goes. Meanwhile a solution with reflection & c# extensions could work to get the internal properties.
Should I close this PR? We can see what is needed for #5484 after 3.6.

tomspilman:
> Should I close this PR?

No it is fine... we'll merge this after 3.6 is all.

willmotil:
Cool no rush its a ascetic nicety mostly. 

I wasn't really working on it i just thought no one had time, so i tried to do a quick submit but i screwed it up github said i failed to sync it whatever that means.

PR #5496, opened 2017-02-23
Added support for arbitrary defines (i.e. LIGHTS=3) passed to the Effect compiler

rds1983:


tomspilman:
This seems reasonable...  we will look at it soon and likely merge it.

tomspilman:
Actually @rds1983 .

If you could add this feature to the unit test here:

https://github.com/MonoGame/MonoGame/blob/develop/Test/ContentPipeline/EffectProcessorTests.cs#L83

By adding something like this into `ParserTest.fx`:
```
#if MACRO_DEFINE_TEST != 3
This is a syntax error
#endif
```

rds1983:
I've updated TestDefines()/DefinesTest.fx

tomspilman:
That is great!  Thanks @rds1983 !

PR #5497, opened 2017-02-23
Uwp separate game thread from ui to increase input responsiveness

roponator:
The goal of this fix is to remove input lag. It improves on @BrianPeek 's XamlThreads branch.

A new worked thread is used to run the actual monogame logic and draw loop. This new game thread and the UI thread must not access concurrently to functions in UAPGameWindow.cs, hence a lot of locks.

This means that the game cannot render/update the same time as a window message recieved (either the game or the UI thread must wait).

To make the UI of the app (UI thread) responsive  the render loop has to be disabled while an UI event arrives, otherwise the game loop has the lock for to long and UI is laggy.

Right now this fixes only the "async" game loop mode for monogame, not the "sync" one. It has been tested using an automated UI testing tool for quite some time on real game and currently seems to be working without any problems, but further tests are needed.

This also requires that any code that requires to run on the UI thread (eg UWP controls, AdSDK,...) must be dispatched within 2 layers of tasks (this is still something I'm looking into), otherwise the UI thread dispatch call freezes: 
1) Start a background thread task from the game thread (the task must not run neither on the game nor UI thread)
2) This new background task must call the UI thread dispatcher if you want to control UI elements, otherwise the UI thread dispatch call waits for the lock, but will never get it since the lock is owned by the game thread, which waits for the blocked UI dispatch call in the game thread and voila deadlock.
(any suggestions for a nicer fix are very welcome, maybe the dispatch call can operate async, WIP...).

Keyboard input locking is separated from the UI and game loop locks because it has to be more responsive (and is very simple to do so).

I didn't spend much time finding out why the messages get queued up since some d3d samples use the same approach.

Relevant older issues/pulls:
#5352 
#5333


jakepoz:
Sweet, I will be testing this more thoroughly shortly :)

jakepoz:
Kk, I tested this with @roponator and we found a fix for dispatching things to the UI thread. New PR is coming soon.

PR #5500, opened 2017-02-23
Fix typo in the exclude.addins which cause warnings when installing the Addin in XS

dellis1972:
@KonajuGames @tomspilman I found this one today, more on the way.


tomspilman:
This finished successfully... the build bot just failed to update the PR for some reason.

PR #5501, opened 2017-02-23
Fix #5498 Pipeline Tool template loading on MacOS

### LINKED ISSUE #5498: Add New Item in Mac Pipeline tool has an empty list
The "Add New Item" menu which usually shows a list for SpriteFonts and shaders is current empty on the Mac Version of the Pipeline Tool

screen shot 2017-02-23 at 17 04 55

dellis1972:
This had two problems

1) The templates were not being included in the Mac App because
the Action for the items was not Content but None.

2) The templates ended up in the Resources directory on Mac which
is where they are supposed to be. So we need to update the app
to look in that location too.

@KonajuGames @tomspilman bug number two fixed :)

### DIFF
--- a/Build/Projects/Pipeline.definition
+++ b/Build/Projects/Pipeline.definition
@@ -305,39 +305,39 @@
     </EmbeddedResource>
 
     <!-- Copy the template resources to the output -->
-	<None Include="Templates\Effect.png">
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks @dellis1972 ... merging!

PR #5505, opened 2017-02-24
Nuget portable target update

DDReaper:
Updated the targets for the two portable NuGet definitions.
Old files were only targeting .Net4 and not Portable Net targets.


tomspilman:
Thanks for fixing this @DDReaper !

Issue #5486, opened 2017-02-19
[DesktopGL] Window not updated when presentation changed


Jjagg:
In #5429 the DesktopGL platform switched from using GraphicsDeviceManager.Legacy to the newer (and more XNA-compatible) GraphicsDeviceManager (GDM). The way the window should get updated and how it works in the DX path is by overriding GamePlatform.OnPresentationChanged in the platform specific GamePlatform implementation ([WinFormsGamePlatform](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Windows/WinFormsGamePlatform.cs#L121) for DX). That function is called in GraphicsDeviceManager.ApplyChanges (which is not the best place for the call, but more on that later). 

For DX we update the window size in WinFormsGamePlatform.OnPresentationChanged. For SDLGamePlatform the function is not overriden at all so the window is not made aware of any changes. Looking at the UWP implementation it seems like it only handles full screen switches and not the window size, I'm not sure how the actual window resizing is handled for that path atm. So the fix is pretty straightforward, we just need to set the window client bounds to reflect the new backbuffer size in the OnPresentationChanged overrides for SDL (and possibly move some code around to do the same for UWP). This is a pretty big bug so it should get fixed ASAP.

So after applying that fix, GDM will be responsible for notifying the GamePlatform that the presentation changed which in turn notifies the GameWindow and updates its bounds. There's still two issues here. Well, one actual issue and one thing that's not so much an issue as an annoyance. 

The actual issue is that you can change presentation parameters without touching the GDM by calling `GraphicsDevice.Reset(PresentationParameters)`, so the GamePlatform/GameWindow will not be notified in that case. IMO we should create a new internal event called PresentationChanged that the GDM subscribes to on its GraphicsDevice and responds to by notifying the GamePlatform of the presentation changes by calling OnPresentationChanged. This event would be triggered right before the `GraphicsDevice.Reset` event, but note that we can't omit the PresentationChanged event and just respond to the Reset event for this purpose because Reset is public (so users can subscribe) and delegate invocation order is not specified in the docs so it's best not to make assumptions.

[...]

PR #5507, opened 2017-02-25
[DesktopGL] Fix window not getting updated

Jjagg:
Provides a minimal fix for the issue described in #5486. 
Since 3.6 is around the corner I decided to just make minimal changes needed to get this working again. Once this is merged I'll make further improvements for after 3.6. @dellis1972

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,9 @@ project.lock.json
 /MonoGame.Framework/MonoGame.Framework.WindowsUniversal.project.lock.json
 artifacts/
 
+# JetBrains Rider
+.idea/
+
 #Tooling
 _ReSharper*/
 *.resharper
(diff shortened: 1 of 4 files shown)


Jjagg:
Oh, hold on this might not be ready yet.

dellis1972:
@Jjagg this is working, but on MacOS using Ctrl+Cmd+F or the green maximise (which toggles full screen) does not appear to update the IsFullScreen setting. I'm sure it used to since I needed that to work in the Stardew Valley port.

dellis1972:
@Jjagg ok maybe not. Looks like its an existing bug.

polsys:
On DesktopGL, the `GraphicsDeviceManagerFixtureTest.ClientSizeChangedOnDeviceReset` test is failing. I did not trace it any further, but could it be related to this? If I understand correctly, this PR adds another path that may raise a `ClientSizeChanged` event.

There's also another failing `GraphicsDevice` test, caused by some recent commit.

PR #5508, opened 2017-02-25
Added ability to determine the underlying game platform type during the run-time.

rds1983:
It may be useful in some cases.
I.e. if you are creating Effect by providing it's bytecode and therefore would need to know whether the underlying platform is DX or GL.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -271,6 +271,7 @@
     <Compile Include="GamePlatform.Mobile.cs">
       <Platforms>Android,iOS,Web,tvOS</Platforms>
     </Compile>
+    <Compile Include="GamePlatformType.cs" />
     <Compile Include="GameRunBehavior.cs" />
     <Compile Include="GameServiceContainer.cs" />
     <Compile Include="GameTime.cs" />
--- a/MonoGame.Framework/Android/AndroidGamePlatform.cs
+++ b/MonoGame.Framework/Android/AndroidGamePlatform.cs
@@ -11,6 +11,11 @@ namespace Microsoft.Xna.Framework
 {
     class AndroidGamePlatform : GamePlatform
     {
+        public override GamePlatformType GamePlatformType
+        {
+            get { return GamePlatformType.Android; }
+        }
+
         public AndroidGamePlatform(Game game)
             : base(game)
         {
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -58,6 +58,11 @@ namespace Microsoft.Xna.Framework
 
         partial void PlatformConstruct();       
 
+        public GamePlatformType GamePlatformType
+        {
+            get { return Platform.GamePlatformType; }
+        }
+        
         public Game()
         {
             _instance = this;
--- a/MonoGame.Framework/GamePlatform.cs
+++ b/MonoGame.Framework/GamePlatform.cs
@@ -7,7 +7,6 @@ using Microsoft.Xna.Framework.Graphics;
 using Microsoft.Xna.Framework.Input;
 using Microsoft.Xna.Framework.Input.Touch;
 
-
 namespace Microsoft.Xna.Framework
 {
     abstract partial class GamePlatform : IDisposable
@@ -99,6 +98,8 @@ namespace Microsoft.Xna.Framework
                 _window = value;
             }
         }
+
+        public abstract GamePlatformType GamePlatformType { get; }
 
         #endregion
 
--- a/MonoGame.Framework/GamePlatformType.cs
+++ b/MonoGame.Framework/GamePlatformType.cs
@@ -1,0 +1,55 @@
+﻿using System;
+
+namespace Microsoft.Xna.Framework
+{
+    public partial struct GamePlatformType : IEquatable<GamePlatformType>
+    {
+        // The public platforms that MonoGame supports
+        public static readonly GamePlatformType Android = new GamePlatformType("Android");
+        public static readonly GamePlatformType iOS = new GamePlatformType("iOS");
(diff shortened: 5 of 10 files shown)


KonajuGames:
The only issue with this approach is that we have other private platforms that cannot be included in a public enumeration.  An enumeration is a fixed type and cannot be extended in a manner that doesn't expose the ones we wish to keep private.

rds1983:
hmm, will adding Custom constant solve the issue?

rds1983:
Another approach would be to remove that enum and change the GamePlatformType property type to int. And declare bunch of public constants:
public const int GamePlatformTypeAndroid = 0;
public const int GamePlatformTypeiOS = 1;
public const int GamePlatformTypeWindows = 2;
etc...

KonajuGames:
> hmm, will adding Custom constant to the enum solve the issue?

It won't help because you don't know how many there are right now, and in the future.  It's something we have previously looked at for other enumerations that reference specific platforms.

> Another approach would be to remove that enum and change the GamePlatformType property type to int. 

You lose all type safety with that approach, and separate implementations of different platforms could conflict.  These are all things that we have already considered.  It's not an easy problem to solve efficiently and flexibly while retaining type safety.

rds1983:
Maybe change it to `string PlatformName` then?
And return "Windows" (for Windows), "DesktopGL" (for WindowsGL/Linux), etc. It's still better than nothing.

KonajuGames:
For this case I think a string would suffice, but wrap it in a struct so we're not comparing strings at runtime by only storing and comparing the `int` hashcode.  Replace the `enum` with a separate partial struct `GamePlatformType` in its own source file.  It's a struct so it doesn't allocate from the heap (unless it is boxed for some reason) and it only stores a single `int`.

````
    public partial struct GamePlatformType : IEquatable<GamePlatformType>
    {
        int _id;

        // The public platforms that MonoGame supports
        public static readonly GamePlatformType Android = new GamePlatformType("Android");
        public static readonly GamePlatformType iOS = new GamePlatformType("iOS");
        public static readonly GamePlatformType DesktopGL = new GamePlatformType("DesktopGL");
        public static readonly GamePlatformType Windows = new GamePlatformType("Windows");
        public static readonly GamePlatformType WindowsPhone81 = new GamePlatformType("WindowsPhone8_1");
        public static readonly GamePlatformType Windows81 = new GamePlatformType("Windows8_1");
        public static readonly GamePlatformType WindowsUniversal = new GamePlatformType("WindowsUniversal");

        public GamePlatformType(string name)
        {
            _id = name.GetHashCode();
        }

        public override bool Equals(Object obj)
        {
            return obj is GamePlatformType && this == (GamePlatformType)obj;
        }
        {
            // The _id field is already the hashcode of the platform name
            return _id;
        }
        public static bool operator ==(GamePlatformType a, GamePlatformType b)
        {
            return a._id == b._id;
        }
        public static bool operator !=(GamePlatformType a, GamePlatformType b)
        {
            return a._id != b._id;
        }

        public bool Equals(GamePlatformType other)
        {
            return _id == other._id;
        }
    }
````

Private platforms can then extend the partial struct, e.g. `GamePlatformType.PS4.cs` would consist of

````
    public partial struct GamePlatformType
    {
        // A platform that MonoGame supports, but not in the public repository
        public static readonly GamePlatformType PS4 = new GamePlatformType("PS4");
    }
````

These extended platforms will be accessible as `GamePlatformType.PS4` in code that has access to the private repository.

I'd like to get some other comments on this approach before it is implemented though.

**Comment by rds1983** [comment]
Awesome idea! I've also added Name property to the GamePlatformType. Could be useful.
I hope that PR gets through.
As right now I have to use following hack:
```c#
_isOpenGL = (from f in typeof (GraphicsDevice).GetRuntimeFields() where f.Name == "glFramebuffer" select f).
```

cra0zy:
I did something similar before, *caugh* #4585 *caugh*, feel free to use it in here for an extra `CurrentSystem` property or similar :)

cra0zy:
@rds1983 You should make a separate branch for the System.Drawing change.

Since I'm guessing you don't know how execute the following commands:
```
git remote add upstream https://github.com/MonoGame/MonoGame.git
git fetch upstream
git checkout upstream/develop
git branch sysdraw
git checkout sysdraw
git submodule update
git chery-pick 7e1e3e1
git push -u origin sysdraw
```

And on GitHub site go to MonoGame repo main page and you will get "Want to submit a PR using branch".

cra0zy:
In case you want to reset your develop branch:
```
git fetch upstream
git checkout upstream/develop
git branch -D develop
git branch develop
git checkout develop
git push -u origin develop --force
```

And if you want to merge newest upstream changes to any branch do:
```
git fetch upstream
git rebase upstream/develop
# you might need to fix some files now, if so execute bellow commands after you are done
git add .
git rebase --continue
```

rds1983:
sigh, I lost again in the battle with mighty git.
I'll fix it in a few. my develop branch will have pr for this thread change(ability to determine platform type during the runtime), and I'll make separate branch `sysdraw`.

rds1983:
Finally this PR had been fixed.

rds1983:
Synchronized this PR will the current MG. I still think that'll be a useful feature to know type of the underlying game platform during the run-time. Especially it'll be useful for third party developers like me.

Jjagg:
Agreed, I think this would be nice in case people want platform specific behavior. IMO similar to this it would be nice to expose what graphics API is being used.

rds1983:
Should be easy task if this PR gets merged:
```c#
var isDirectX = myGame.GamePlatformType == GamePlatformType.Windows ||
  myGame.GamePlatformType == GamePlatformType.WindowsUniversal;
```

rds1983:
Closing this PR, as I wrongly used my develop branch as source. I'll create new PR with same code but different source branch.

PR #5510, opened 2017-02-25
Update installer for VS2017 project templates

KonajuGames:
Change back to checking for the existence of the Project Templates folder in My Documents because VS2017 massively changed their registry structure.  Still only works correctly if the user has admin privileges.  May do a separate non-admin installer for templates only at a later date.

### DIFF
--- a/Installers/Windows/MonoGame.nsi
+++ b/Installers/Windows/MonoGame.nsi
@@ -75,7 +75,7 @@ RequestExecutionLevel admin
 ; The stuff to install
 Section "MonoGame Core Components" CoreComponents ;No components page, name is not important
   SectionIn RO
-
+  
   ; Install the VS support files.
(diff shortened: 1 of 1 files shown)


dellis1972:
@KonajuGames looking a the docs "SetShellVarContext current" might we what we need in the installer?

http://nsis.sourceforge.net/Docs/Chapter4.html#setshellvarcontext

KonajuGames:
That doesn't work because at that time the installer is running as the
admin user, not the user who started the install. The "current user" (the
default behaviour) in this case would be the admin user. And we can't run
part of the installer as the user and another part as admin. We would have
to make two separate installers: one for templates and one for the core.
The templates installer would be run and would itself trigger the core
installer. The core installer runs as admin and the templates installer
runs as the user.

PR #5512, opened 2017-02-26
Fix texture ValidateParams exception with DXT1/3/5 decompression on Linux

hach-que:
When running on Linux without DXT decompression support (such as the Mesa OpenGL software renderer), the code path for decompressing DXT textures on load is hit.  This code currently has a bug where `levelDataSizeInBytes` is passed through to the `SetData` method, but `levelDataSizeInBytes` is not updated after DXT decompression, so the validation which occurs inside Texture2D to check that the element count is correct fails.

@KonajuGames has informed me this code path is likely going away soon, so I'll also be future proofing Protogame to turn off DXT compression for textures in the content pipeline when targeting Linux (since it's not supported everywhere).

### DIFF
--- a/MonoGame.Framework/Content/ContentReaders/Texture2DReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/Texture2DReader.cs
@@ -86,21 +86,32 @@ namespace Microsoft.Xna.Framework.Content
 					    case SurfaceFormat.Dxt1:
                         case SurfaceFormat.Dxt1SRgb:
                         case SurfaceFormat.Dxt1a:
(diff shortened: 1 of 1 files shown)


KonajuGames:
Don't make it a permanent change in Protogame.  It is only your use case of Mesa in a headless GPU-less environment that does not support DXT (due to S3TC patent fears).  Anyone playing a game on Linux will not be using Mesa and will have DXT support.

KonajuGames:
Merging. Thanks @hach-que

tomspilman:
I ran into this same bug on a console platform where we were using the original PC XNBs decompressed.  Thanks @hach-que !

PR #5516, opened 2017-02-27
Extend Null Check (Fixes #5493)

### LINKED ISSUE #5493: A potential bug in Audio/Xact/Cue.cs
I found a potential bug with the SonarC# plugin for SonarQube - on the following line _curSound could be null (because you check for null on the line before that):
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/Xact/Cue.cs#L260

All the best,
Valeri

06needhamt:
This PR Fixes the null check issue outlined in #5493


tomspilman:
We fixed this as part of https://github.com/MonoGame/MonoGame/pull/5709.

Issue #5311, opened 2016-11-21
Certain Draw overload misplaces Rectangles and ignores scale


whosdatdev:
When using the overload
`Draw(Texture2D texture, Vector2? position = null, Rectangle? destinationRectangle = null, Rectangle? sourceRectangle = null, Vector2? origin = null, float rotation = 0.0f, Vector2? scale = null, Color? color = null, SpriteEffects effects = SpriteEffects.None, float layerDepth = 0.0f)`

the scale is ignored, but only if you give a destinationRectangle (not ignored when using the Vector2 position)
I get that you are supposed to supply either Vector2 + scale OR the Rectangle, but the whole overload is really misleading imo.

Edit: just found another bug with the same overload:
This line works as expected: (not the buggy overload)
`_spriteBatch.Draw(testTex, rect, Color.White);`
while this one draws the rectangle on a completely different place: (buggy overload, should do nearly the same)
`_spriteBatch.Draw(testTex, destinationRectangle: rect, color: Color.White, origin: new Vector2(rect.Width / 2f, rect.Height / 2f));`

mrhelmut:
>I get that you are supposed to supply either Vector2 + scale OR the Rectangle, but the whole overload is really misleading imo.

You're right. If you specify a destination rectangle, it assumes that the rectangle size is already scaled to your need. Documentation could be enhanced.

whosdatdev:
Just edited since i found another bug. I also tried installing 3.4 from nuget instead, but it has the same bug.

After some debugging i'm sure that specifying an origin will fuck up the rectangle position. How can a bug like this exist since at least 3.4? -.- (DirectX version, didn't try the opengl one, but i assume it won't matter)

willmotil:
Origin relates to the rotation you can specify the origin of the rotation as a offset from the top left when using the rotation float. E.G.  new Vector( myimage.Width *.5f, myimage.Height *.5f) will move the centroid of the rotations origin point.

Jjagg:
Closing this because the issue is based on a misunderstanding of the API and confusing overload has been marked as `Obsolete`.

PR #5517, opened 2017-02-28
Mark 'shortcut' Draw() overload for future removal.

nkast:
This method can be confusing (#5311) and it's bad on performance.

Similar discussion in #5403.


tomspilman:
Thanks... merging!

nkast:
This method was marked 'Obsolete' since v3.6.
I propose to remove it completely in the next version 3.8.

PR #5520, opened 2017-02-28
[StoreApps/XAML] Run Game Loop and Input on a seperate Thread

nkast:
Fixes mouse, keyboard lag [UAP] touch, rotation lag & freeze (WP8.1).
#4897, #5300, #3948, #3589, #3316, #4105(?).

This probably still has a ton of bugs. It's hard to tell unless you review every line of MG codebase and run extensive tests. Unfortunetly, that's the nature of thread syncronization. I'd prefer to wait until after 3.6 (and a possible 3.6.1 service update) for this PR.

On the other hand #5352 is more predictable and solves 99% of the issues with just a couple of lines.

PR #5521, opened 2017-02-28
Support SwapChainPanel on W8.1 / WP8.1 and update templates to use SwapChainPanel

nkast:
I found SwapChainBackgroundPanel behaving very unpredictable as I was testing #5520. That makes sense since it comes from W8.0, It wasn't supposed to run in window mode (W10).
(I found a workaround for #5520 out of nowhere , but that was after I had this PR ready! 8-) )

- `SwapChainBackgroundPanel` & `SwapChainPanel` are wrapped into `GenericSwapChainPanel`.
- `SwapChainBackgroundPanel` is still supported on W8.1 to allow easier upgrade from 3.5, the constructor is marked as [Obsolete].
- W8.1 templates are updated to use `SwapChainPanel`.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -1396,6 +1396,9 @@
     <Compile Include="Windows8\GameFrameworkViewSource.cs">
       <Platforms>Windows8</Platforms>
     </Compile>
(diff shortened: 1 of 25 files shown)


nkast:
Made a few changes to fix CoreApplication project.

PR #5522, opened 2017-02-28
Improvements to window handling and full screen switches for desktop

Jjagg:
This makes some improvements to how the GameWindow works:

- Multisampling is supported by the window on initilization. I create a dummy window to create a temporary GraphicsContext to query for multisampling support. I got some weird behaviour though, so would be nice if someone can test that.
- The window position can be set before the native window is instantiated.
- The window title can be set before the native window is instantiated.
- Initializing in full screen mode no longer depends on the internal `applyChanges` method in game. I #iffed out DESKTOPGL for that method so we know it no longer needs it. One step closer to getting rid of it completely.
- Stop listening for SDL Resized event. If it is raised, the SizeChanged event is raised too, so listening to that is sufficient.
- The ClientSizeChanged event is only raised when a user resizes the window (also while resizing, not just when resizing ends but there's no way to distinguish the two with the SDL API), not when it is resized through a back buffer size change.
- The window position is restored after exiting full screen. This is not very stable because SDL raises a bunch of events when entering/exiting full screen and we can't distinguish them from when a user moves the window. Optionally we can check if window position in the event is (0, 0) relative to the current display and assume its a full screen transition and ignore it when it is.
- Window centering now also happens after resizing the backbuffer in stead of just when exiting full screen. This is also not very stable because of the previous issue. Sometimes we detect a window move when
it's because of a full screen switch and afterwards the window will no longer be centered.

I also added some todo's/comments for more intricate stuff like recreating the window when certain presentation parameters change.

### DIFF
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Xna.Framework {
 
 		public abstract string ScreenDeviceName { get; }
 
-		private string _title = MonoGame.Utilities.AssemblyHelper.GetDefaultWindowTitle();
+	    private string _title;
 
         /// <summary>
         /// Gets or sets the title of the game window.
@@ -79,6 +79,7 @@ namespace Microsoft.Xna.Framework {
 
         protected GameWindow()
         {
+            _title = MonoGame.Utilities.AssemblyHelper.GetDefaultWindowTitle();
             TouchPanelState = new TouchPanelState(this);
         }
 
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -37,7 +37,8 @@ namespace Microsoft.Xna.Framework {
 
 		public abstract string ScreenDeviceName { get; }
 
-		private string _title;
+		private string _title = MonoGame.Utilities.AssemblyHelper.GetDefaultWindowTitle();
+
         /// <summary>
         /// Gets or sets the title of the game window.
         /// </summary>
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -159,14 +159,5 @@ namespace Microsoft.Xna.Framework {
 		protected internal abstract void SetSupportedOrientations (DisplayOrientation orientations);
 		protected abstract void SetTitle (string title);
 
-#if DIRECTX && WINDOWS
-        public static GameWindow Create(Game game, int width, int height)
-        {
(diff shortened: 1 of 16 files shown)


MonoGame.Framework/SDL/SDLGameWindow.cs
```diff
@@ -64,6 +69,11 @@ public override IntPtr Handle
             get { return _handle; }
         }
 
+        private bool HasNativeWindow
```
cra0zy: Pointless helper property, the code behind it is simple enough that you should that code.

MonoGame.Framework/SDL/SDLGameWindow.cs
```diff
@@ -157,33 +179,32 @@ internal void CreateWindow()
...
+            _supressMoved = true;
         }
 
         ~SdlGameWindow()
         {
             Dispose(false);
         }
 
-        private static Sdl.Rectangle GetMouseDisplay()
+        private static int GetMouseDisplay()
```
cra0zy: The only reason to use this method is to get the mouse rectangle, so by returning the int you are making 1 extra wasteful call to the native API later on.

MonoGame.Framework/GameWindow.cs
```diff
@@ -37,7 +37,8 @@ public abstract class GameWindow {
 
 		public abstract string ScreenDeviceName { get; }
 
-		private string _title;
+		private string _title = MonoGame.Utilities.AssemblyHelper.GetDefaultWindowTitle();
```
cra0zy: It's better to move assigning of variables to the constructor, it's more readable to be able to see all the starting values in there instead of searching for their declarations.

MonoGame.Framework/Windows/WinFormsGameForm.cs
```diff
@@ -81,16 +82,15 @@ protected override void WndProc(ref Message m)
                 case WM_KEYDOWN:
                     switch (m.WParam.ToInt32())
                     {
+                        case 0x5C: // Right Windows Key
                         case 0x5B:  // Left Windows Key
 
-                            if (this.WindowState == FormWindowState.Maximized)
+                            if (_window.IsFullScreen)
```
Jjagg: This should only happen in hardware full screen probably.

cra0zy:
I would recommend you split this PR into multiple smaller PRs, plus you are modifying 2 platforms in here now... that is never a good sign.

Jjagg:
True, I continued in this branch because I depend on some changes. I'll see exactly what's needed and try to pull that out.

PR #5537, opened 2017-03-02
[SDL] Disable XInput so more than 4 controllers can be detected

cra0zy:
** NEEDS TESTING **

### DIFF
--- a/MonoGame.Framework/SDL/SDLGamePlatform.cs
+++ b/MonoGame.Framework/SDL/SDLGamePlatform.cs
@@ -56,6 +56,10 @@ namespace Microsoft.Xna.Framework
             if (version >= 205 && CurrentPlatform.OS == OS.Windows && Debugger.IsAttached)
                 Sdl.SetHint("SDL_WINDOWS_DISABLE_THREAD_NAMING", "1");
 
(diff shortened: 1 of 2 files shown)


mrhelmut:
I'll test this next week on Windows.

endimilojkoski:
I tested this with bunch of gamepads (2x Xbox 360 wireless, 1x Xbox 360 wired, PS4, 2x iBuffalo) and it works with no issues. Pushing this live so the players that have a ton of wireless gamepads can test :)

endimilojkoski:
Tried with 5 gamepads, combination of wired and wireless x360 and xone gamepads and only 4 are recognized so this is still not solved.

mrhelmut:
Hmm, it should support up to 16 gamepads. Wonder if that hint really is working.

endimilojkoski:
@mrhelmut just to be clear, I can use 8 gamepads but I can't use more than 4 XInput controllers like Xbox 360 and Xbox One. Do you see any workaround for this?

mrhelmut:
Well, no, it looks like XInput is still being used and limiting the Xbox gamepads to 4. The SDL hint isn't working as expected.

PR #5538, opened 2017-03-03
Support for RenderTarget2D multisampling under DirectX.

zigzag312:
Added RenderTarget2D multisampling for DirectX. Code is based of [this post](http://community.monogame.net/t/quick-overview-how-to-work-with-msaa-directx-too/8570). 

Changes in first commit are

- Multisampled resource is now created
- Added RenderTarget2D field to RenderTarget2D class. It serves as non-multisampled destination for resolving multisampled RT.
- Basic resolving logic

First commit is just to get it working with little consideration to performance. Resolving probably needs to be optimized (resolve only if needed). 

This depends on #5477


zigzag312:
Is there any way of knowing if RT has been written to? So it could be marked as dirty and resolved only in that case. 

Right now RTs are resolved when they are _unset_. It would be possible to improve performance in some cases if RT was resolved only when its data is used: when it is used as a source texture for drawing or on GetData. And of course only if its data has changed since last resolve. 

Any thoughts on this?

KonajuGames:
> Is there any way of knowing if RT has been written to? So it could be marked as dirty and resolved only in that case.

If a render target has been set, assume it is dirty.  It will lose contents when being set anyway, unless `RenderTargetUsage.PreserveContents` has been used (do we even support flag properly?).  If a developer sets and unsets a render target without drawing to it, that is a bad move on their part and the render target now has undefined contents.

MonoGame.Framework/Graphics/Texture.DirectX.cs
```diff
@@ -27,33 +28,33 @@ public IntPtr GetSharedHandle()
 
         internal abstract SharpDX.Direct3D11.Resource CreateTexture();
 
-        internal SharpDX.Direct3D11.Resource GetTexture()
+        internal virtual SharpDX.Direct3D11.Resource GetTexture()
```
KonajuGames: This method was made virtual, but I don't see anything that overrides it.  We try to keep the use of `virtual` to a minimum because it takes longer to call a virtual than a regular method.

zigzag312:
> If a render target has been set, assume it is dirty.

Then resolving logic is done too, I think.

> do we even support flag properly?

[Check here](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.cs#L793)  It seems OK to me.

MonoGame.Framework/Graphics/Texture.DirectX.cs
zigzag312: Its a leftover from debugging. I forgot to remove it.

hach-que:
Can this get merged?  We have a need for multisampled render targets in Protogame because of how our render pipeline works.

I'm going to temporarily merge this changes into our patched version of MonoGame, but it'd be great to have this fixed properly upstream.

hach-que:
(I have a rebased version of this PR that no longer has merge conflicts if anyone is interested)

zigzag312:
There are no known issues, so I think it can be merged. (I'll just rebase it)

zigzag312:
Hmm does anyone have any idea why there is a change "Submodule Dependencies updated from 3aee60 to 979abf" under "Files Changed" tab?. I didn't make any submodule dependencies for this PR.

hach-que:
I can confirm that this change works as intended.  The way I rebased it was I hard reset to the RenderTarget2D_Multisampling branch (before that merge that just occurred), then soft reset to the parent on the develop parent so that it would only contain the branches changes.  Then I committed it, and rebased that single commit, which gave me just the four files that were changed (rebasing the whole history turned out to be hard with lots of conflicts).

KonajuGames:
There has been no response to my code review comments re. `GetShaderResourceView()`.

KonajuGames:
And unnecessary changing of field names.

zigzag312:
@KonajuGames where can I see those review comments? I can only see comment about virtual GetTexture() method

zigzag312:
> And unnecessary changing of field names.
Some fields were renamed because they were changed from private to protected. Should I revert to old names with "_" prefix for protected fields?

KonajuGames:
Ahh, it appears I had to click another button to actually submit the review comments.

Yes, those fields should remain named as they were.

zigzag312:
OK, I see review comments now. I'll make the changes :)

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -30,19 +30,14 @@ namespace Microsoft.Xna.Framework.Graphics
...
     public partial class Texture2D : Texture
     {
-        private bool _shared;
-
-        private bool _renderTarget;
-        private bool _mipmap;
-
-        private SampleDescription _sampleDescription;
+        protected internal bool shared;
+        protected internal bool mipmap;
```
Jjagg: [Microsoft's design guidelines](https://msdn.microsoft.com/en-us/library/ms229057(v=vs.110).aspx) don't actually allow public or protected instance fields, so they don't have a naming convention for them. Though R# likes to PascalCase them by default. That's what I always do too. I thought the most applied convention was to only prefix private fields with an underscore.

KonajuGames: I'd be fine with those fields staying private and with the underscore prefix (as they were before this PR), and a adding new protected PascalCase property used by derived classes that returns the private field.

MonoGame.Framework/Graphics/Texture.DirectX.cs
```diff
@@ -35,25 +36,25 @@ internal SharpDX.Direct3D11.Resource GetTexture()
             return _texture;
         }
 
-        internal SharpDX.Direct3D11.ShaderResourceView GetShaderResourceView()
+        internal virtual SharpDX.Direct3D11.ShaderResourceView GetShaderResourceView()
```
zigzag312: Is it OK if I do it like this, as I changed _resourceView back to private?

```
internal SharpDX.Direct3D11.ShaderResourceView GetShaderResourceView()
{
    if (_resourceView == null)
        _resourceView = CreateShaderResourceView();

    return _resourceView;
}

protected internal virtual SharpDX.Direct3D11.ShaderResourceView
    CreateShaderResourceView()
{
    return new SharpDX.Direct3D11.ShaderResourceView(GraphicsDevice._d3dDevice, GetTexture());
}
```

zigzag312:
Any ideas why _Package Mac and Linux_ fails, as only changes made are to .DirectX.cs files?

KonajuGames:
It build all platforms regardless of what changed. We're having issues with
the Mac build machine at the moment. It's not behaving nicely.

zigzag312:
Resolved new merge conflicts. Added dispose call for _resolvedTexture. 
Can this PR be merged, so I don't waste time resolving new conflicts?

zigzag312:
I think I discovered a possible bug, so I'll need to test it before this PR is ready.

zigzag312:
I just found out that bug has been already fixed in #5766. So this is good to go IMO.

hach-que:
@tomspilman @KonajuGames Can this please be merged so I don't have to keep rebasing and squashing it on the Redpoint fork?

MonoGame.Framework/Graphics/RenderTarget2D.DirectX.cs
```diff
@@ -124,5 +129,68 @@ DepthStencilView IRenderTarget.GetDepthStencilView()
...
+
+        protected internal override SampleDescription CreateSampleDescription()
+        {
+            return this.GraphicsDevice.GetSupportedSampleDescription
+                (SharpDXHelper.ToFormat(this._format), this.MultiSampleCount);
+        }
+
+        internal void ResolveSubresource()
+        {
+            GraphicsDevice._d3dContext.ResolveSubresource(this._texture, 0, _resolvedTexture._texture, 0,
```
tomspilman: `GraphicsDevice._d3dContext` needs to be `lock`d here.

zigzag312: Should I also add null and/or IsDisposed checks for `_texture` & `_resolvedTexture`, before calling `_d3dContext.ResolveSubresource()` method? I am not sure if there can be any scenario where null/disposed can happen, but I also don't want to add unnecessary checks.

tomspilman: > Should I also add null and/or IsDisposed checks

I don't think we have to do that right now.

tomspilman:
Thanks @zigzag312 ... merging.

Funny enough I implemented this same feature for XB1 a week or two ago.

PR #5541, opened 2017-03-03
Remove WindowsGL and Linux NuGet packages

cra0zy:
Both packages are just leftovers from before DesktopGL.


tomspilman:
@dellis1972 this look right to you?

Also what about Mac and MonoMac?

cra0zy:
> Also what about Mac and MonoMac?

Mac and MonoMac stay because they are different platform from DekstopGL.

dellis1972:
can we change the WindowsGL/Linux packages to just depend (or redirect?) on the DesktopGL one.. That way people using the existing packages will get an upgrade.

cra0zy:
> can we change the WindowsGL/Linux packages to just depend (or redirect?) on the DesktopGL one.. That way people using the existing packages will get an upgrade.

And any new user will get confused by the existence of those 2 packages... I think they should be properly removed / hidden.

DDReaper:
Can we also hide the existing packages on NuGet once this is merged. To avoid confusion.

Hiding them on the NuGet management page doesn't break existing projects using them, just prevents their use in new projects.
Thoughts @cra0zy @dellis1972 @tomspilman ?

cra0zy:
> Can we also hide the existing packages on NuGet once this is merged. To avoid confusion.

Those are my thoughts exactly in the comment above.

> can we change the WindowsGL/Linux packages to just depend (or redirect?) on the DesktopGL one.. That way people using the existing packages will get an upgrade.

One extra thing that I forgot to mention, AFAIK NuGet packages need to have at least 1 c# assembly file so even if I did add a dependency it would still be problematic.

DDReaper:
> can we change the WindowsGL/Linux packages to just depend (or redirect?) on the DesktopGL one.. That way people using the existing packages will get an upgrade.

No, NuGet does not support that. To maintain referencing integrity the only way is to uninstall the old and install the new manually. It does not support cross-referencing packages as it could lead to all sorts of shenanigans

tomspilman:
> Hiding them on the NuGet management page doesn't break existing projects using them, just prevents their use in new projects.

I just did this now for the current 3.6 packages for Linux and WindowsGL.

 http://www.nuget.org/profiles/MonoGame

Is there anything else in that list that needs to be unlisted?

Should we add a warning or something to the Ouya and WindowsPhone8 packages that they are unsupported going forward?

tomspilman:
@dellis1972 @DDReaper - Another thing.

The descriptions on most of these NuGet packages suck.  It is a marketing message and not addressing stuff people would actually want to know.  Like "WTF is DesktopGL?"  and "Why use MacOS vs DesktopGL?".

We should look to fix all that.

Also the copyrights on all the packages need to be updated to something like:

> Copyright © 2009-2017 The MonoGame Team

Please call out here if you plan to make these fixes so we don't duplicate work.

tomspilman:
@dellis1972 @DDReaper - Or maybe the summary needs to be the informational bit and the description should focus on what MonoGame is.  Not sure which is best.

DDReaper:
I'd agree @tomspilman Things have moved on since I crafted the original Readme's for the NuGet's and other documentation, so it could do with an update.
Just needs that readme.txt file updating in the NuGet folder once it's agreed what that text should be.

cra0zy:
> Please call out here if you plan to make these fixes so we don't duplicate work.

I call dibs.

cra0zy:
I fixed a bunch of stuff: #5572 It will probably require changing something in build server considering NuGet compilation is not connected to nant build files.

On a minor side note, if you ever decide to switch build system from nant, check out <a href="http://cakebuild.net/">Cake</a>, it's awesome.

PR #5543, opened 2017-03-03
First shot at including the Native libs in the Nuget Packages.

dellis1972:
This adds the native libraries (SDL, openal-soft) to the DesktopGL Nuget. It also makes sure that we copy these files to the output directory.

Oh and MG is .net 45 now right? so the nuggets should reflect that.

### DIFF
--- a/NuGetPackages/MonoGame.Framework.DesktopGL.nuspec
+++ b/NuGetPackages/MonoGame.Framework.DesktopGL.nuspec
@@ -18,9 +18,20 @@
         <language>en-US</language>
     </metadata>
     <files>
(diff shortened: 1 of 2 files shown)


dellis1972:
@cra0zy snap! https://github.com/MonoGame/MonoGame/pull/5542

tomspilman:
yeah... I was about to say.

So you guys decide which is better and lets combine the best of both.

cra0zy:
@dellis1972 Can we go with mine PR on this?

dellis1972:
Personally I think the .targets file from this one is more compact, but what ever not fussed.

cra0zy:
Your .targets file is more compact, but mine is way more readable.

DDReaper:
Yeah, I'd have to agree with readability over compact in this case. Sorry @dellis1972 
So long as it "Just Works" (tm)

Issue #5515, opened 2017-02-27
Windows 10 Templates broken on VS2017


DDReaper:
VS2017 is throwing template errors when creating a new Windows 10 project.

Open GL project is working and will check others.

Will submit a PR to fix soon

Jjagg:
Gotta love maintaining templates...

DDReaper:
So I got the rest of the templates to work and install. However, the Windows 10 template is foobared.
The old template wizard isn't available anymore and $Safename$ replacements aren't working (even though they work in a fresh template export)
Considering completely regenerating the template in 2017 now as it is becoming irksome.

DDReaper:
Ok, completely confused now.  Just spotted the 3.6 release, installed and confirmed the Windows 10 template is FooBar'ed still.  However, what is strange, is that the original Template issue has gone (VS Complaining about the Wizard template version)
But it is not working as it doesn't generate the GUID's.

This does mean however, that Windows 10 on Vs2017 is not working.  Likely it will still work on 2015 but I haven't got that to test at the mo.

Is it work putting a post out on the MG blog for now @tomspilman ?

DDReaper:
OK, just been doing some intensive testing and on at least three machines, Windows 10 no longer works at all, not even in Vs2015 now.  So it seems my "fixed" templates wren't being overwritten by the installer. After a clean uninstall and re-install produces the old error.

In 2015, you get the Wizard Template Failure, whereas 2017 just says Project creation failed (you have to add the project to an existing solution to get the same error as 2015)
This was the initial issue I was trying to fix, which was easy enough to resolve by upping the WinRT Wizard extension in the project template. 

However, this didn't work for Windows 10 at all which is where I got to.  Still testing but so far I'm struggling.  @dellis1972 any update from you (after an uninstall/reinstall of MG 3.6?)

DDReaper:
Closed previous PR and will submit a new one.

In testing of the new 3.6 build, @dellis1972 rightly points out they are working if you have BOTH VS2015/VS2017 installed together (with all the correct options)

However, in testing I've found that with VS2017 alone, no matter your install options, you simply cannot create a new project.  Updating the WinRT template wizard to 15 (from 14) in the template definition does seem to fix it.

[...]

PR #5546, opened 2017-03-05
Fix to enable Windows 10 template to work on clean VS2017 only machines

DDReaper:
As noted in #5515 , Current Windows 10 template only works on machines with BOTH VS2015 & VS2017.  On machines with only VS2017, the template fails to create with a Wizard version error.

Does not affect opening existing projects.

Also only ignores unsupported templates, such as Windows 8.1/Phone

(Tried creating the fix against the 3.6 tag but it went horribly wrong, so re-sumbitting against develop_


DDReaper:
Tested on:
* VS 2015/2017 machine - Windows 10 - OK
* Existing VS2017 only machine - OK
* New VS 2017 only VM - OK

Seems fine.
Although until the 3.6 Installer is updated, it won't work. So may need post to explain (or re-release 3.6?)
Over to you @tomspilman @dellis1972

KonajuGames:
We will need to get a 3.6.1 release out soon with this and some other fixes.  Thanks for the testing of this @DDReaper

PR #5547, opened 2017-03-05
OpenAL error handling clean up

### LINKED ISSUE #5381: OpenAL initialization failing quite often
Hello,

This is a problem that I never encountered locally, but our crash database is filled with these kind of crashes and we have several players complaining about it.

```OggStream``` ctor crashes at times, with ```External component has thrown an exception```, which makes it quite hard to understand remotely and would indicates that the native side of OpenAL-Soft is failing.

It happens on a simple ```Content.Load<Song>()```. No threading involved. Here's the full stack trace:

```
   at Microsoft.Xna.Framework.Audio.OpenALSoundController.ReserveSource()
   at Microsoft.Xna.Framework.Audio.OggStream..ctor(String filename, Action finishedAction, Int32 bufferCount)
   at Microsoft.Xna.Framework.Media.Song.PlatformInitialize(String fileName)
   at Microsoft.Xna.Framework.Media.Song..ctor(String fileName, Int32 durationMS)
   at Microsoft.Xna.Framework.Content.SongReader.Read(ContentReader input, Song existingInstance)
   at Microsoft.Xna.Framework.Content.ContentTypeReader`1.Read(ContentReader input, Object existingInstance)
   at Microsoft.Xna.Framework.Content.ContentReader.InnerReadObject[T](T existingInstance)
   at Microsoft.Xna.Framework.Content.ContentReader.ReadObject[T]()
   at Microsoft.Xna.Framework.Content.ContentReader.ReadAsset[T]()
   at Microsoft.Xna.Framework.Content.ContentManager.ReadAsset[T](String assetName, Action`1 recordDisposableObject)
   at Microsoft.Xna.Framework.Content.ContentManager.Load[T](String assetName)
```

The thing is that ```OpenALSoundController.ReserveSource()``` doesn't make any call to OpenAL, so the problem may be elsewhere.

Note that by the time this code is reached, we already successfully initialized the device and loaded a few ```SoundEffect``` before loading the ```Song```, which indicates that the AL device and buffers initializations are doing fine until song loading.

If anyone has inputs regarding this issue, that would be very welcome. I haven't been able to reproduce it.

mrhelmut:
This PR cleans up the error handling in the OpenAL device/context initialization by making it critical and throwing ```NoAudioHardwareException``` exceptions at any kind of error and adding the AL error code to its message.

It was previously not critical and was continuing execution with no AL device, which resulted in several difficult to debug situations (because the original AL error was lost, and raising non trivial error messages or unrelated exceptions). This now allows to get more info from remote crash reporters.

It also removes some initialization checks because initialization will now halt on all errors.

This is a first step to get better stack traces in an attempt to resolve #5381.

@dellis1972 @cra0zy @KonajuGames

### DIFF
--- a/MonoGame.Framework/Audio/OALSoundBuffer.cs
+++ b/MonoGame.Framework/Audio/OALSoundBuffer.cs
@@ -32,15 +32,8 @@ namespace Microsoft.Xna.Framework.Audio
 
 		public OALSoundBuffer ()
 		{
-            try
-            {
-                AL.GenBuffers(1, out openALDataBuffer);
-                ALHelper.CheckError("Failed to generate OpenAL data buffer.");
-            }
-            catch (DllNotFoundException e)
-            {
-                throw new NoAudioHardwareException("OpenAL drivers could not be found.", e);
-            }
+            AL.GenBuffers(1, out openALDataBuffer);
+            ALHelper.CheckError("Failed to generate OpenAL data buffer.");
 		}
 
         ~OALSoundBuffer()
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -104,8 +104,6 @@ namespace Microsoft.Xna.Framework.Audio
 #endif
         private List<int> availableSourcesCollection;
         private List<int> inUseSourcesCollection;
-        private bool _bSoundAvailable = false;
-        private Exception _SoundInitException; // Here to bubble back up to the developer
         bool _isDisposed;
         public bool SupportsADPCM = false;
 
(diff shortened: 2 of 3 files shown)


tomspilman:
I am a little worried about non-critical runtime errors... like maybe setting a volume or pitch out of range now becoming fatal errors.  Other than that i'm all for improvements in error checking.

mrhelmut:
I don't think that this PR makes these operations fatal because initialization is now fatal, which prevents these operations to be executed at all. But maybe I've missed something.

mrhelmut:
FYI, I pushed this branch on Steam.

tomspilman:
> FYI, I pushed this branch on Steam.

@mrhelmut  - Cool...  how long has it been up without issues?

mrhelmut:
A couple of days, no crash reported. We're having a sale and visibility round over the weekend, I'll let you know how it goes with the influx of players.

mrhelmut:
Lots of new/returning players over the weekend, and we had way less blocking sound initialization errors. For the few we had (only 2 reported out of a thousand launch), this PR correctly returns ```NoAudioHardwareException``` and there is no other issue at all.

It doesn't solve #5381 but we now have a way better view on the issue, which seem less severe now than some false positive errors aren't triggered anymore.

KonajuGames:
Thanks for the testing. I'll merge this now.

PR #5528, opened 2017-03-01
Make MonoGame content builder work from one source on all the platforms

cra0zy:
**WIP Changes**

Stuff done:
 - Removed a bunch of #if in MG.Framework.Content.Pipeline
 - Make it so Windows, Linux and Mac content builder stuff are using the exact same files

Why do this?
 - All part of the master plan: #5262

CC. @DDReaper

PR #5548, opened 2017-03-06
Custom FreeImage wrapper

cra0zy:
Notes:
 - I tried to keep method / enum names the same as FreeImage.Net names
 - FreeImage documentation I used to write the wrappers: http://graphics.stanford.edu/courses/cs148-10-summer/docs/FreeImage3131.pdf

Why do this? Needed one set of FreeImage assemblies for #5528 and making a custom FreeImage.Net.dll seemed worse than doing this.


cra0zy:
So "small update", while writing the custom wrapper I noticed that the old method for `ConvertToRawBits` used an extra param, and I thought maybe it was there so it could support some old API, but it turns out that FreeImage.NET had a custom implementation for that method.

Make sure that the copyright text at the top of FreeImageAPIC.cs is correct before merging this.

PR #5514, opened 2017-02-27
Add missing xml doc to framework classes

Jjagg:
This was a relaxing experience up to some point, but then it got tedious and I decided to stop.

PR #5550, opened 2017-03-07
Develop document win forms game window

OriAlmogCode:
Documentation was sparse, most of it is now documented.

### DIFF
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -209,7 +209,6 @@ namespace MonoGame.Framework
             }
             finally
             {
-				// On success or failure, release write lock
                 _allWindowsReaderWriterLockSlim.ExitWriteLock();
             }
         }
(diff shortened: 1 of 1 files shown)


Jjagg:
I'm glad you want to help out, but IMO these are pointless. The code is pretty straightforward and the added comments don't clarify anything. E.g.
```
// Set icon and title
SetIcon();
Title = Utilities.AssemblyHelper.GetDefaultWindowTitle();
```
It's obvious that's what happens from the code. If you want to help out with documentation, xml docs on public classes and members is way more useful. (note that I have a PR up that already documents all framework classes, just so no duplicate work is done; you could use that as an example if you want to help out here #5514 ).

tomspilman:
@OriAlmogCode - Thanks for helping, but this isn't the best use of your time.

The documentation we actually require is on the public API that users will actually use.  If you browse thru our current reference documentation:

http://www.monogame.net/documentation/?page=api

You will find lots of namespaces, classes, enums, fields, methods, and properties that are undocumented.  Any help with that would be great to have.  You can find some information and tips on how to contribute to our API reference docs here:

https://github.com/MonoGame/MonoGame/tree/develop/Documentation#api-reference

OriAlmogCode:
Thanks for the direction, I'll get on that.

PR #5560, opened 2017-03-09
Develop documentation

OriAlmogCode:
I've documented most of the API to the BoundingBox class. Also restructured one method, should be a bit more clear what it does now.

### DIFF
--- a/MonoGame.Framework/BoundingBox.cs
+++ b/MonoGame.Framework/BoundingBox.cs
@@ -34,6 +34,11 @@ namespace Microsoft.Xna.Framework
 
         #region Public Constructors
 
+		/// <summary>
+		/// Constructs a BoundingBox using two corners
+		/// </summary>
+		/// <param name="min">The minimum-corner of the box.</param>
+		/// <param name="max">The maximum-corner of the box.</param>
         public BoundingBox(Vector3 min, Vector3 max)
         {
             this.Min = min;
--- a/MonoGame.Framework/BoundingBox.cs
+++ b/MonoGame.Framework/BoundingBox.cs
@@ -448,6 +448,11 @@ namespace Microsoft.Xna.Framework
             return this.Min.GetHashCode() + this.Max.GetHashCode();
         }
 
+        /// <summary>
+        /// Determines if this <see cref="BoundingBox"> intersects with another.
+        /// </summary>
+        /// <param name="box">The box to test against</param>
+        /// <returns>True if the boxes intersect, false otherwise</returns>
         public bool Intersects(BoundingBox box)
         {
             bool result;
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @OriAlmogCode ...  I added a few suggestions to your changes here.

MonoGame.Framework/BoundingBox.cs
```diff
@@ -7,6 +7,11 @@
 using System.Diagnostics;
 using System.Runtime.Serialization;
 
+
+/// <summary>
+/// Defines a box in 3D-space
+/// </summary>
+
```
tomspilman: This needs to go down to where the struct declaration... it doesn't work here.

MonoGame.Framework/BoundingBox.cs
```diff
@@ -352,38 +448,60 @@ public override int GetHashCode()
...
+        /// Determines if this <see cref="BoundingBox"/> intersects with another, placing the result in the "result" parameter
+        /// </summary>
+        /// <param name="box">The box to test against</param>
+        /// <param name="result">The boolean in which to place the result</param>
         public void Intersects(ref BoundingBox box, out bool result)
         {
             if ((this.Max.X >= box.Min.X) && (this.Min.X <= box.Max.X))
             {
-                if ((this.Max.Y < box.Min.Y) || (this.Min.Y > box.Max.Y))
+                if ((this.Max.Y >= box.Min.Y) && (this.Min.Y <= box.Max.Y))
```
tomspilman: First it is preferable not mix documentation work with changes to the code.  It makes it harder for us to check and approve pull requests.

Second we really don't want to see "code cleanup" like changes.  These are changes where you just shuffle code around but have no actual change to functionality.  This just makes maintenance of the project harder.

You can see more about what we expect in pull requests here:

https://github.com/MonoGame/MonoGame/blob/develop/CONTRIBUTING.md#quick-guidelines

So unless the changes in this function are an actual bug fix or performance improvement...  I suggest reverting these changes.

MonoGame.Framework/BoundingBox.cs
```diff
@@ -40,9 +50,17 @@ public BoundingBox(Vector3 min, Vector3 max)
...
+        /// </summary>
+        /// <param name="box">The <see cref="BoundingBox"/> to compare to.</param>
+        /// <returns>The <see cref="ContainmentType"/> relationship between the two objects.</returns>
         public ContainmentType Contains(BoundingBox box)
         {
-            //test if all corner is in the same side of a face by just checking min and max
+            // If, in any dimension: the other box's MAX is less than this box's MIN, OR
+            //                       the other box's MIN is is more than this box's MAX
+            //                    then the two must be disjoint
             if (box.Max.X < Min.X
```
tomspilman: I suggest avoiding adding these sorts of comments within functions unless you are the author of the code in question.

It is hard for us to decide if your comment is valid or not and we don't have the time to investigate it.

MonoGame.Framework/BoundingBox.cs
OriAlmogCode: Fair enough, those are understandable points. I've reverted the changes you noted!

MonoGame.Framework/BoundingBox.cs
```diff
@@ -14,6 +14,10 @@ namespace Microsoft.Xna.Framework
     public struct BoundingBox : IEquatable<BoundingBox>
     {
 
+        /// <summary>
+        /// Defines a box in 3D-space
+        /// </summary>
```
tomspilman: This is still in the wrong place.

MonoGame.Framework/BoundingBox.cs
```diff
@@ -14,6 +14,10 @@ namespace Microsoft.Xna.Framework
     public struct BoundingBox : IEquatable<BoundingBox>
```
tomspilman: Put the `<summary>` for this class above the `[DataContract]` here.

Jjagg:
I [mentioned](https://github.com/MonoGame/MonoGame/pull/5550#issuecomment-284708326) in #5550 that I already wrote xml docs for the framework classes specifically so this wouldn't happen :/ 
My changes are in #5514, BoundingBox docs included.

tomspilman:
>  I already wrote xml docs for the framework classes specifically 

Sorry @Jjagg and @OriAlmogCode ... this is my fault.  I forgot about #5514 already covering the `BoundingBox` type.

@OriAlmogCode I apologize this has been this difficult to contribute some doc improvements.  Just bad luck here in picking a type @Jjagg has an open PR for.  The good news is that there isn't another active open doc PR out there... so anything not in #5514 is open for someone to tackle.

Jjagg:
@OriAlmogCode Please don't let this discourage you. There's plenty of other classes that are still undocumented!

PR #5563, opened 2017-03-09
Add missing GL blend functions

polsys:
This fixes the `BlendState` visual test failure on OpenGL. There is still some noise in the diff, but not enough to fail the test, and it seems not to be related to the blend states.

Tested on Windows only.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsExtensions.cs
+++ b/MonoGame.Framework/Graphics/GraphicsExtensions.cs
@@ -365,10 +365,14 @@ namespace Microsoft.Xna.Framework.Graphics
 		public static BlendingFactorSrc GetBlendFactorSrc (this Blend blend)
 		{
 			switch (blend) {
+            case Blend.BlendFactor:
+                return BlendingFactorSrc.ConstantColor;
 			case Blend.DestinationAlpha:
 				return BlendingFactorSrc.DstAlpha;
 			case Blend.DestinationColor:
 				return BlendingFactorSrc.DstColor;
+            case Blend.InverseBlendFactor:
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Graphics/GraphicsExtensions.cs
```diff
@@ -364,11 +364,15 @@ public static BlendEquationMode GetBlendEquationMode (this BlendFunction functio
...
+			switch (blend) {
+            case Blend.BlendFactor:
+                return BlendingFactorSrc.ConstantColor;
 			case Blend.DestinationAlpha:
 				return BlendingFactorSrc.DstAlpha;
 			case Blend.DestinationColor:
-				return BlendingFactorSrc.DstColor;
+				return BlendingFactorSrc.DstColor;
+            case Blend.InverseBlendFactor:
+                return BlendingFactorSrc.OneMinusConstantColor;
```
tomspilman: Ugh... so `Blend.InverseBlendFactor` was falling thru to the `default` case?  Maybe this is a good reason to make the `default` case throw an exception to avoid this sort of mistake in the future?

tomspilman:
Great find @polsys !

MonoGame.Framework/Graphics/GraphicsExtensions.cs
polsys: Added an exception to the functions I touched, I'll leave better cleanup for later.

Jjagg:
Nice @polsys! This was IMO the worst of the tests that still failed on DesktopGL :)

tomspilman:
Going to merge this... thanks @polsys !

PR #5569, opened 2017-03-10
Don't swap PreferredBackBufferWidth/Height

### LINKED ISSUE #5555: Windows StoreApps PreferredBackBufferWidth/Height bug
PreferredBackBufferWidth/Height is wrong at startup when Client Width<Height.

W8.1 / WP8.1 Screenshots

w81
wp81

The cause of this is https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/GraphicsDeviceManager.cs#L69-L78
which I assume is an attempt to [Automatic Rotation](https://msdn.microsoft.com/en-us/library/ff966425.aspx).
 
On W10 the values are still wrong during Game.Initialize(). After initialization W10 receives a SizeChange event that restores the values.
Even if W8.1/WP8.1 got a resize event the same code is present in MetroGameWindow.cs (UAPGameWindow.cs doesn't swap Width/Height),
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Windows8/MetroGameWindow.cs#L157-L166

IMO it's best to remove this code entirely.

A proper implementation would be to change the Orientation when SupportedOrientations= Default and the game is fullscreen or runs in fullscreen environment (WP,Android,iOS).

Personally I think this behavior is confusing and goes a step too far. I found myself fighting against it in XNA in the past when trying to take advantage of the WP7 scaler.

nkast:
-also remove `_backBufferScale`

Fixes #5555

### DIFF
--- a/MonoGame.Framework/GraphicsDeviceManager.cs
+++ b/MonoGame.Framework/GraphicsDeviceManager.cs
@@ -63,20 +63,10 @@ namespace Microsoft.Xna.Framework
             _preferredDepthStencilFormat = DepthFormat.Depth24;
             _synchronizedWithVerticalRetrace = true;
 
-            // Assume the window client size as the default back 
-            // buffer resolution in the landscape orientation.
             var clientBounds = _game.Window.ClientBounds;
-            if (clientBounds.Width >= clientBounds.Height)
-            {
-                _preferredBackBufferWidth = clientBounds.Width;
-                _preferredBackBufferHeight = clientBounds.Height;
-            }
-            else
-            {
-                _preferredBackBufferWidth = clientBounds.Height;
-                _preferredBackBufferHeight = clientBounds.Width;
-            }
-
+            _preferredBackBufferWidth = clientBounds.Width;
+            _preferredBackBufferHeight = clientBounds.Height;
+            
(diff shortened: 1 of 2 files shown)


tomspilman:
Did you actually test the behavior of not having `_backBufferScale`?

I sort of remember when I wrote that code that I was trying to mimic the XNA WP7 behavior of having the back buffer scale to fit the window.

Or maybe the XAML presentation fixes that now?  I don't know.

nkast:
I tested it on a desktop W8.1 (resize in windowed mode) and on the WP8.1 emulator (rotate it a few times).
I didn't see any issues.

Also tested it with this code in Game() (in WP8.1 & W10 emulator with fullscreen and rotate a couple of times)
```
            graphics.PreferredBackBufferWidth /= 4;
            graphics.PreferredBackBufferHeight /= 4;
            this.Window.ClientSizeChanged += (s, e) =>
            {
                graphics.PreferredBackBufferWidth /= 4;
                graphics.PreferredBackBufferHeight /= 4;
                graphics.ApplyChanges();
            };
```
Everything seems to work as they should. (except for another issue with touch, it's offset by x4)

PR #5572, opened 2017-03-11
Update NuGet packages

cra0zy:
Stuff done:
 - Fixed author, owner, copyright, description, summary and licenseUrl properties
 - Included .targets file for each platform to supply MonoGamePlatform property
 - Allowed NuGet packages to be built from Unix systems
 - Reorganized each NuGet package file into it's own <a href="https://github.com/cra0zy/MonoGame/tree/nugetupdate/NuGet">folder</a>
 - Removed readme.txt file

The compilation command should now be:
```
nuget pack NuGet/DesktopGL/MonoGame.Framework.DesktopGL.nuspec -BasePath . -OutputDirectory <Some Output Directory>
```


tomspilman:
> Reorganized each NuGet package file into it's own folder

Please revert all the folder changes.  It isn't required and making that change screws up the build process.

cra0zy:
> Please revert all the folder changes. It isn't required and making that change screws up the build process.

I didn't do it randomly, I moved it because each platform nuget package needs a .targets file (and that was creating a mess).

Also even if I restore the positions of NuGet packages you'll still need to change the compilation command because otherwise we can't get NuGet package compilation to be compatible with compiling from both Windows and Unix systems.

tomspilman:
> I moved it because each platform nuget package needs a .targets file (and that was creating a mess).

I disagree on it being a mess.  There are just 18 files... now 18 files spread across in 10 folders.  The ratio of files to folders seems too low to call this a mess.

>  you'll still need to change the compilation command because otherwise we can't get NuGet package compilation to be compatible with compiling from both Windows and Unix systems.

The only place the build server compiles NuGet packages is on Windows.  It is done automatically by the build server's NuGet packaging feature.  Linux is not involved in how the automated build generates NuGet packages.

If someone else wants to build these packages on Linux they are free to use `nuget pack` directly themselves.

cra0zy:
> If someone else wants to build these packages on Linux they are free to use nuget pack directly themselves.

It will not work unless we use "/" for filepath (inside nuget package), and Windows will not work if we use "../" in filepath.

Also I am not doing this for Linux, but for MacOS. Why? Well if we are gonna be doing a separate tool for project generating, we are gonna need to be able to get NuGet packages easily from the Mac build bot.

cra0zy:
Anyway if you still wish the structure restored, I'll do it.

tomspilman:
> It will not work unless we use "/" for filepath (inside nuget package), and Windows will not work if we use "../" in filepath.

So the issue is the relative path bits like `..\..\` within lines like:
```
<file src="..\..\MonoGame.Framework.Content.Pipeline\bin\Portable\AnyCPU\MonoGame.Framework.Content.Pipeline.dll" />
```
Correct?  These won't work if the slashes are flipped to `../../`?

>  we are gonna need to be able to get NuGet packages easily from the Mac build bot.
The Mac build bot doesn't generate NuGet packages... all the packages are generated on Windows.

If we were to need the NuGet packages during building of the Mac or Linux installers...  they would be pulled in as artifacts from the Windows side... same as we do now to get all the assemblies.

So I think this is a non-issue.

cra0zy:
> So the issue is the relative path bits like ..\..\ within lines like:
> 
> <file src="..\..\MonoGame.Framework.Content.Pipeline\bin\Portable\AnyCPU\MonoGame.Framework.Content.Pipeline.dll" />
> Correct? These won't work if the slashes are flipped to ../../?

<table>
<tr>
<td></td>
<td>Windows</td>
<td>Unix</td>
</tr>
<tr>
<td>..\</td>
<td>OK</td>
<td>ERROR</td>
</tr>
<tr>
<td>../</td>
<td>ERROR</td>
<td>OK</td>
</tr>
<tr>
<td>\</td>
<td>OK</td>
<td>ERROR</td>
</tr>
<tr>
<td>/</td>
<td>OK</td>
<td>OK</td>
</tr>
</table>

cra0zy:
> If we were to need the NuGet packages during building of the Mac or Linux installers... they would be pulled in as artifacts from the Windows side... same as we do now to get all the assemblies.
> 
> So I think this is a non-issue.

But NuGet packages only get built AFTER the main build, because they require the assemblies, so....

Also debugging the tool from Mac/Linux will become a lot harder.

tomspilman:
Please revert all the folder changes here.

We don't have a project template generation tool that uses NuGet yet.  Once we do and actually have a problem to solve we can look to fix the NuGet base path to be the root of the repo and not `MonoGame.Framework\bin` like it is now.

cra0zy:
I think GitHub got bugged, I updated the branch as you can see: https://github.com/cra0zy/MonoGame/tree/nugetupdate , but GitHub is still showing the old version of it in here.

cra0zy:
I created a new PR, hopefully this bug is a one time thing: #5573

Issue #5109, opened 2016-08-15
Support <Link> content in MGCB and Pipeline tool


tomspilman:
In XNA you could do the following in a content project:

```
    <Compile Include="..\..\..\art\gfx\wood.PNG">
      <Link>gfx\wood.PNG</Link>
      <Name>wood</Name>
      <Importer>TextureImporter</Importer>
      <Processor>TextureProcessor</Processor>
    </Compile>
```

With this the content pipeline reads the source content from the `..\..\..\art\gfx\wood.PNG` path then writes it out to `gfx\wood.PNG` in the output folder.

When importing this `.contentproj` into the Pipeline tool we get this:

```
#begin ../../../art/gfx/wood.PNG
/importer:TextureImporter
/processor:TextureProcessor
/build:../../../art/gfx/wood.PNG
```

This produces both the wrong output path when building content, but also looks wrong in the Pipeline tool which shows folders for `..` which it should never do.

There are multiple things to do to fix this:
1. Add support for links/rename of content on processing.
2. Correctly visualize these links in the Pipeline tool.
3. Support adding linked content in the Pipeline tool.
4. Support importing linked content from `.contentproj`.

This issue is similar to https://github.com/MonoGame/MonoGame/issues/4724, https://github.com/MonoGame/MonoGame/issues/4846, and https://github.com/MonoGame/MonoGame/issues/4980.

tomspilman:
As for support in the `.MGCB` file I think we should do something like this:
```
/build:../../../art/gfx/wood.PNG;gfx/wood.PNG
```
... or maybe....
```
/build:../../../art/gfx/wood.PNG>gfx/wood.PNG
```
We then just use `string.Split()` to break the build argument into the two different paths.  The source file path and the optional link file path.

KonajuGames:
I was thinking along the lines of

/link:gfx/wood.png
/build:../../wood.png

The link parameter, if present, would be used for the treeview and
intermediate file output.

But either option would work.

tomspilman:
Yeah `/link:gfx/wood.png` is the most obvious way to do it.  For some reason I was thinking this could conflict with other processor parameters, but clearly it would not.

KonajuGames:
As long as it is before the /build parameter, it should be fine.

nkast:
[...]

PR #5574, opened 2017-03-12
Added support for "link" property in the .mgcb file (although not cal…

vchelaru:
…led that in the .mgcb file).

Build tool now sets output according to link if it exists.

### DIFF
--- a/Tools/MGCB/SourceFileCollection.cs
+++ b/Tools/MGCB/SourceFileCollection.cs
@@ -21,9 +21,14 @@ namespace MGCB
         [XmlArrayItem("File")]
         public List<string> SourceFiles { get; set; }
 
+        [XmlArrayItem("File")]
+        public List<string> DestFiles { get; set; }
+
+
         public SourceFileCollection()
         {
             SourceFiles = new List<string>();
+            DestFiles = new List<string>();
             Config = string.Empty;
         }
 
@@ -56,6 +61,13 @@ namespace MGCB
                 var inContent = SourceFiles.Any(e => string.Equals(e, sourceFile, StringComparison.InvariantCultureIgnoreCase));
                 if (!inContent)
                     SourceFiles.Add(sourceFile);
+            }
+
+            foreach (var destFile in other.DestFiles)
+            {
+                var inContent = DestFiles.Any(e => string.Equals(e, destFile, StringComparison.InvariantCultureIgnoreCase));
+                if (!inContent)
+                    DestFiles.Add(destFile);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


vchelaru:
Partly addresses https://github.com/MonoGame/MonoGame/issues/5109

jamesford42:
Note this other PR which implements the input;output and makes use of it for purposes of renaming the output asset (and i can import .contentproj items which had asset names given). 

https://github.com/MonoGame/MonoGame/pull/3009

Anyhow, just posting it as a reference really, since unless i or someone else has time to spend, it will probably stay there in PR limbo.

Tools/MGCB/SourceFileCollection.cs
```diff
@@ -21,9 +21,14 @@ public sealed class SourceFileCollection
         [XmlArrayItem("File")]
         public List<string> SourceFiles { get; set; }
 
+        [XmlArrayItem("File")]
+        public List<string> DestFiles { get; set; }
```
tomspilman: Isn't this going to serialize both source and dest files together?  Isn't that bad?

tomspilman:
This could use some testing before we merge it.

Tools/Pipeline/Common/ContentItem.cs
```diff
@@ -32,6 +32,10 @@ public class ContentItem : IProjectItem
         [Browsable(false)]
         public string OriginalPath { get; set; }
 
+        // For now making this not browsable, until the core functionality works
+        [Browsable(false)]
+        public string Link { get; set; }
```
cra0zy: Maybe naming this property `DestinationPath` would be nicer for consistency?

Also remove the comment, it's always gonna be browsable false, think of how linked content is shown in Visual Studio.

Tools/MGCB/SourceFileCollection.cs
vchelaru: No, it will serialize them as follows:
```
<SourceFiles>
   <File> asdf </File>
</SourceFiles>
<DestFiles>
   <File> asdf </File>
</DestFiles>
```

KakCAT:
Hi,

I've been doing some work in the Pipeline tool side and been testing this PR extensively. Everything works well but there's a missing part, which are linked files that are copied (not built).

The code to fix this is here:

https://github.com/MonoGame/MonoGame/commit/4d749131c6d73fb4e2d30df836c31c0773b44050

lines 356 and beyond of Tools/MGCB/BuildContent.cs 

The "link splitting" code is used 2 times in the MGCB and it is also used in the Pipeline tool, so it'd probably be wised to make a function for it, although I have no idea where it would be placed to.

cra0zy:
Linking support has been added in dev release of MonoGame.

Issue #5581, opened 2017-03-14
FileHelpers.ResolveRelativePath() removes critical characters


mrhelmut:
Hello,

It's quite rare, but ```FileHelpers.ResolveRelativePath()``` removes some critical characters in file paths, like ```#```.

We noticed this because some players install their game in folders such as ```C:/#Games``` and when ```FileHelpers.ResolveRelativePath``` tries to create an ```Uri```, it splits the path at the ```#``` and will return a path like ```C:/assetName.ext``` when trying to load ```C:/#Games/My_Game/Content/assetName.ext```, resulting in a file exception.

The problem is that ```Uri``` constructed with the ```file://``` protocol will interpret it as an URL and will remove hashes / anchors when resolving path. Removing ```file://``` and the addition of forward slash at the beginning seem to do the trick.

I'll try to come up with a fix and a test.

@tomspilman are the ```file://``` and ```filePath = ForwardSlashString + filePath;``` here for a specific purpose?

KonajuGames:
It's the `Uri` class handling the '#' character.  Might have to do some special handling of that.

People actually use `#` in path names?  I know it's a valid path character in Windows at least, but it's not something I have ever seen before.

KonajuGames:
Try this.  Change

    var src = new Uri("file://" + filePath);

to

    var src = new Uri("file://" + filePath.Replace("#", "%23"));

and

    var dst = new Uri(src, relativeFile);

to

    var dst = new Uri(src, relativeFile.Replace("#", "%23"));

`Uri` will correctly interpret `%23` as the hash character and output it in the `LocalPath` property.

KonajuGames:
That's a proof of concept, but perhaps one of the URI encoding functions would be a solution here.

mrhelmut:
It works, and I agree, we should encode this. I'll check that.

mrhelmut:
I tested all URI special characters, only ```#``` and ```&``` expose this issue.
I haven't figured out which good encoding method to use without depending on System.Web.

KonajuGames:
I just tried with `Uri.EscapeUriString` and `Uri.EscapeDataString`.  They don't work.  The first doesn't escape it, and the second escapes too much and breaks `Uri`.

I'd accept the `Replace` fix.  Let's see what @tomspilman and others have to say.

cra0zy:
https://msdn.microsoft.com/en-us/library/system.web.httpserverutility.urlencode.aspx

[...]

PR #5583, opened 2017-03-14
Fix FileHelpers.ResolveRelativePath

cra0zy:
Fixes: #5581

### DIFF
--- a/MonoGame.Framework/Utilities/FileHelpers.cs
+++ b/MonoGame.Framework/Utilities/FileHelpers.cs
@@ -4,11 +4,14 @@
 
 using System;
 using System.IO;
+using System.Text;
 
 namespace Microsoft.Xna.Framework.Utilities
 {
     internal static class FileHelpers
     {
+        private static readonly char[] UrlSafeChars = new[] { '.', '_', '-', ';', '/', '?', '\\', ':' };
+
         public static readonly char ForwardSlash = '/';
         public static readonly string ForwardSlashString = new string(ForwardSlash, 1);
         public static readonly char BackwardSlash = '\\';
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Utilities/FileHelpers.cs
```diff
@@ -44,7 +47,7 @@ public static string ResolveRelativePath(string filePath, string relativeFile)
                 filePath = ForwardSlashString + filePath;
 
             // Get a uri for filePath using the file:// schema and no host.
-            var src = new Uri("file://" + filePath);
+            var src = new Uri("file://" + UrlEncode(filePath));
 
             var dst = new Uri(src, relativeFile);
```
KonajuGames: I would suggest encoding the relative path as well.  It could contain characters in need of encoding.

cra0zy: That would create an incorrect file path.

cra0zy: Oh no, wait, I got confused, you are right.

MonoGame.Framework/Utilities/FileHelpers.cs
```diff
@@ -3,12 +3,15 @@
...
-
+using System.IO;
+using System.Text;
+
 namespace Microsoft.Xna.Framework.Utilities
 {
     internal static class FileHelpers
-    {
+    {
+        private const string UrlSafeChars = "._-;/?:@&=+&,";
```
mrhelmut: ```&``` isn't safe. :)

cra0zy: You are right, I included the query URL components which I shouldn't have, fixed now.

mrhelmut:
It works perfectly, thank you @cra0zy ! :)

dellis1972:
looks ok to me @tomspilman @KonajuGames

KonajuGames:
It generates more garbage than I would ideally like, and the magic numbers could be changed to their char equivalents, but functionally it does work.

cra0zy:
I'll see what I can do.

cra0zy:
@KonajuGames Better?

KonajuGames:
That does improve it a bit.  Once the tests have passed, I'm happy to merge it.

KonajuGames:
Merging. Thanks @cra0zy

PR #5584, opened 2017-03-14
Remove Scale and Rotation properties from Matrix

### LINKED ISSUE #5553: Scale and Rotation properties in Matrix class return incorrect values
When calling the Scale and Rotation properties on the Matrix class, the values returned are often incorrect. I have been struggling to get some code to work for some time and discovered that by replacing these calls with a single call to Matrix.Decompose the correct values are returned. For example, calling Scale seems to always return 1,0,0 but when using Decompose the correct value 1,1,1 is returned.

Jjagg:
See #5553 for discussion. Can we agree on this? Or should we keep the properties but correctly implement them (the same way the Decompose function works).

Fixes #5553

### DIFF
--- a/MonoGame.Framework/Matrix.cs
+++ b/MonoGame.Framework/Matrix.cs
@@ -357,17 +357,6 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
-        /// Rotation stored in this matrix.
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @Jjagg !

Issue #5530, opened 2017-03-02
Annoying fullscreen differences between DirectX and OpenGL


Thraka:
I'm using MonoGame 3.6 and `GraphicsDeviceManager.HardwareModeSwitch = false;` and having a hard time working out my code for how I want my game to render. Depending on if I'm using DirectX or OpenGL, I get different behaviors.

| Library | Mode | GraphicsDevice.Viewport.Bounds |
| ------- | ----- | ---------------------|
| DirectX | Original window | 640x400 |
| DirectX | To full | 640x400 |
| DirectX | Back to window | 624x361 |
| OpenGL | Original window | 640x400 |
| OpenGL | To full | 1920x1080 |
| OpenGL | Back to window | 640x400 |

As you can see, DirectX misbehaves when coming out of fullscreen, and OpenGL misbehaves when going to fullscreen.

KonajuGames:
Thanks for these bug reports. Simple repro steps will make this easier to
track down and add unit tests so it doesn't break again.

willmotil:
In the january version im still using any changing of size on the backbuffer from the initial start to a large windows size or to a smaller size causes a huge drop in fps my thought is scaleing but i haven't tested whats actually going on with the backbuffer.

aienabled:
This issue seems to be caused by WinFormsGameWindow no triggering ClientSizeChanged event on window state change.. just reported #5533.

mrhelmut:
Regarding ```HardwareModeSwitch = false``` and OpenGL, I may be wrong, but I believe that there's nothing to do about it, that's just how borderless OpenGL works with SDL2 and I don't know if there's a workaround. Maybe @cra0zy knows this better, I'm not sure.

cra0zy:
OpenGL side is working how it is supposed to.

Jjagg:
> This issue seems to be caused by WinFormsGameWindow no triggering ClientSizeChanged event on window state change.. just reported #5533.

That event is not used internally to respond to size changes, so it's not related.

Thraka:
Shouldn't the viewport always stay the same no matter what? Otherwise it will change how much of the scene is being pulled into and rendered on the backbuffer.

Jjagg:
[...]

PR #5585, opened 2017-03-15
[DX] Improve window handling

### LINKED ISSUE #5533: WinFormsGameWindow - no ClientSizeChanged event on window state change
Hi guys,

I've noticed that in MonoGame 3.6 ClientSizeChanged (for WinFormsGameWindow) is not invoked anymore when window state changes (normal->maximized and vice versa). It worked properly for us at v3.5.

As a workaround for this issue I've changed (in WinFormsGameWindow.cs):
```C#
Form.ResizeEnd += OnResizeEnd;
```
to
```C#
Form.ResizeEnd += OnResizeEnd;
Form.SizeChanged += OnResizeEnd;
```

But that might lead to some side effects...
BTW, I'm using `GraphicsDeviceManager.HardwareModeSwitch = false;`.

Regards!

Jjagg:
This tackles some of the issues with stuff like minimizing/maximizing the window and full screen switching. The largest fix is that now the back buffer size is properly updated when switching to full screen. Other fixes include correctly switching from hardware to soft full screen and back, minimizing the window when the windows key is pressed only when in hard full screen mode and raising the ClientSizeChanged event when switching window state (e.g. minimizing the window).

I also moved some responsibilities from WinFormsGamePlatform to WinFormsGameWindow because it made more sense there.

An issue that is not solved with this is changing the back buffer size while in hard full screen. That will not change the display mode while it should be able to do that. Can anyone more familiar with D3D pitch in? Do we just switch out of full screen and back in or is there a better way to do this?

There's another problem when creating a device in soft full screen (this didn't work before either). The back buffer size won't be right, which results in a blank screen (is that normal?). The cause of this problem is WinForms related. It seems the form ClientBounds are not updated immediately when changing the WindowState to Maximized when the form is not yet visible.

fixes #5533 
Also fixes the DirectX issue from #5530.
@aienabled, @Thraka

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -122,7 +122,8 @@ namespace Microsoft.Xna.Framework.Graphics
 		public event EventHandler<ResourceDestroyedEventArgs> ResourceDestroyed;
         public event EventHandler<EventArgs> Disposing;
 
-        internal event EventHandler<PresentationChangingEventArgs> PresentationChanging;
+        internal event EventHandler<PresentationEventArgs> PresentationChanging;
+        internal event EventHandler<PresentationEventArgs> PresentationChanged;
 
         private int _maxVertexBufferSlots;
         internal int MaxTextureSlots;
@@ -610,13 +611,14 @@ namespace Microsoft.Xna.Framework.Graphics
         {
             PlatformValidatePresentationParameters(PresentationParameters);
             EventHelpers.Raise(this, DeviceResetting, EventArgs.Empty);
-            EventHelpers.Raise(this, PresentationChanging, new PresentationChangingEventArgs(PresentationParameters));
+            EventHelpers.Raise(this, PresentationChanging, new PresentationEventArgs(PresentationParameters));
 
             // Update the back buffer.
             OnPresentationChanged();
             
+            EventHelpers.Raise(this, PresentationChanged, new PresentationEventArgs(PresentationParameters));
             EventHelpers.Raise(this, DeviceReset, EventArgs.Empty);
-        }
+       }
 
         public void Reset(PresentationParameters presentationParameters)
         {
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -122,7 +122,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		public event EventHandler<ResourceDestroyedEventArgs> ResourceDestroyed;
         public event EventHandler<EventArgs> Disposing;
 
-        internal event EventHandler<EventArgs> PresentationChanged;
+        internal event EventHandler<PresentationChangingEventArgs> PresentationChanging;
 
         private int _maxVertexBufferSlots;
         internal int MaxTextureSlots;
@@ -610,11 +610,11 @@ namespace Microsoft.Xna.Framework.Graphics
         {
             PlatformValidatePresentationParameters(PresentationParameters);
             EventHelpers.Raise(this, DeviceResetting, EventArgs.Empty);
+            EventHelpers.Raise(this, PresentationChanging, new PresentationChangingEventArgs(PresentationParameters));
 
             // Update the back buffer.
             OnPresentationChanged();
             
-            EventHelpers.Raise(this, PresentationChanged, EventArgs.Empty);
             EventHelpers.Raise(this, DeviceReset, EventArgs.Empty);
         }
 
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -122,7 +122,6 @@ namespace Microsoft.Xna.Framework.Graphics
 		public event EventHandler<ResourceDestroyedEventArgs> ResourceDestroyed;
         public event EventHandler<EventArgs> Disposing;
 
-        internal event EventHandler<PresentationEventArgs> PresentationChanging;
         internal event EventHandler<PresentationEventArgs> PresentationChanged;
 
         private int _maxVertexBufferSlots;
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -222,18 +222,8 @@ namespace Microsoft.Xna.Framework.Graphics
             GraphicsCapabilities = new GraphicsCapabilities();
             GraphicsCapabilities.Initialize(this);
 
-#if DIRECTX
-            if (PresentationParameters.IsFullScreen)
-            {
-                int newWidth, newHeight;
-                if (PresentationParameters.HardwareModeSwitch)
-                    GetModeSwitchedSize(out newWidth, out newHeight);
-                else
-                    GetDisplayResolution(out newWidth, out newHeight);
-
-                PresentationParameters.BackBufferWidth = newWidth;
-                PresentationParameters.BackBufferHeight = newHeight;
-            }
+#if WINDOWS || DESKTOPGL
+            CorrectBackBufferSize();
 #endif
 
             Initialize();
@@ -621,11 +611,11 @@ namespace Microsoft.Xna.Framework.Graphics
 
         partial void PlatformValidatePresentationParameters(PresentationParameters presentationParameters);
 
-        public void Reset()
-        {
 #if WINDOWS || DESKTOPGL
+        private void CorrectBackBufferSize()
+        {
             // Window size can be modified when we're going full screen, we need to take that into account
-            // now so the back buffer has the right size.
+            // so the back buffer has the right size.
             if (PresentationParameters.IsFullScreen)
             {
                 int newWidth, newHeight;
@@ -646,6 +636,13 @@ namespace Microsoft.Xna.Framework.Graphics
                 if (newHeight < PresentationParameters.BackBufferHeight)
                     PresentationParameters.BackBufferHeight = newHeight;
             }
+        }
+#endif
+
+        public void Reset()
+        {
+#if WINDOWS || DESKTOPGL
+            CorrectBackBufferSize();
 #endif
 
             PlatformValidatePresentationParameters(PresentationParameters);
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -636,6 +636,15 @@ namespace Microsoft.Xna.Framework.Graphics
                 PresentationParameters.BackBufferWidth = newWidth;
                 PresentationParameters.BackBufferHeight = newHeight;
             }
+            else
+            {
+                int newWidth, newHeight;
+                GetDisplayResolution(out newWidth, out newHeight);
+                if (newWidth < PresentationParameters.BackBufferWidth)
+                    PresentationParameters.BackBufferWidth = newWidth;
+                if (newHeight < PresentationParameters.BackBufferHeight)
+                    PresentationParameters.BackBufferHeight = newHeight;
+            }
 
             PlatformValidatePresentationParameters(PresentationParameters);
 
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -222,6 +222,21 @@ namespace Microsoft.Xna.Framework.Graphics
             Setup();
             GraphicsCapabilities = new GraphicsCapabilities();
             GraphicsCapabilities.Initialize(this);
+
+#if DIRECTX
+            if (PresentationParameters.IsFullScreen)
+            {
+                int newWidth, newHeight;
+                if (PresentationParameters.HardwareModeSwitch)
+                    GetModeSwitchedSize(out newWidth, out newHeight);
+                else
+                    GetDisplayResolution(out newWidth, out newHeight);
+
+                PresentationParameters.BackBufferWidth = newWidth;
+                PresentationParameters.BackBufferHeight = newHeight;
+            }
+#endif
+
             Initialize();
         }
 
@@ -609,9 +624,23 @@ namespace Microsoft.Xna.Framework.Graphics
 
         public void Reset()
         {
+            // Window size can be modified when we're going full screen, we need to take that into account
+            // now so the back buffer has the right size.
+            if (PresentationParameters.IsFullScreen)
+            {
+                int newWidth, newHeight;
+                if (PresentationParameters.HardwareModeSwitch)
+                    GetModeSwitchedSize(out newWidth, out newHeight);
+                else
+                    GetDisplayResolution(out newWidth, out newHeight);
+
+                PresentationParameters.BackBufferWidth = newWidth;
+                PresentationParameters.BackBufferHeight = newHeight;
+            }
+
             PlatformValidatePresentationParameters(PresentationParameters);
+
             EventHelpers.Raise(this, DeviceResetting, EventArgs.Empty);
-            EventHelpers.Raise(this, PresentationChanging, new PresentationEventArgs(PresentationParameters));
 
             // Update the back buffer.
             OnPresentationChanged();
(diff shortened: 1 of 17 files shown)


Thraka:
Let's get this released! 👍

Jjagg:
> There's another problem when creating a device in soft full screen

Worked around this by resizing backbuffer when initializing the window.

Still would like to get the resize while in full screen thing fixed before merging this.

Jjagg:
Note to self: use swapChain.ResizeTarget for resolution switch in fullscreen.
Source: https://msdn.microsoft.com/en-us/library/windows/desktop/bb205075(v=vs.85).aspx

When that works this is ready for review and hopefully merge :) 
I started writing a simple project for testing windowing functionality and presentation parameters, that should make testing things like this easier in the future.

MichaelDePiazzi:
@Jjagg How's this coming along? I recently synced my MonoGame fork with the latest code and found an issue when switching out of soft-fullscreen mode which this seems to fix. I plan to deploy this in an experimental build to my users to better test it and see if any other issues turn up.

Jjagg:
This needs just a little more work. I've been really busy and not sure when things will ease up a little. But I'll try and get this fixed asap :)

Thraka:
@Jjagg Thanks for your work on this. I'm really looking forward to it 👍 👍 👍

MichaelDePiazzi:
@Jjagg No worries, I can totally empathise with you on that! Thanks for the update :)

cra0zy:
@tomspilman Mac build bot got stuck, it has been building this PR for ~2 hours right now.

MichaelDePiazzi:
@Jjagg I just found a bug in this PR - In soft-fullscreen mode, you can no longer set a different resolution than the desktop resolution. e.g. If I try set the resolution to 1280x720, it remains at my desktop resolution of 1920x1080.

cra0zy:
> @Jjagg I just found a bug in this PR - In soft-fullscreen mode, you can no longer set a different resolution than the desktop resolution. e.g. If I try set the resolution to 1280x720, it remains at my desktop resolution of 1920x1080.

That is how it should be, it was a bug that you could change resolution in soft fullscreen mode.

MichaelDePiazzi:
> This is how it should be, it was a bug that you could change resolution in soft fullscreen mode.

Really, since when? It's worked like this for years in WinDX, and my game kinda depends on this behaviour... :worried:

cra0zy:
> Really, since when? It's worked like this for years in WinDX, and my game kinda depends on this behaviour... 😟

You can just change the viewport to get that behavior.

MichaelDePiazzi:
> You can just change the viewport to get that behavior.

@cra0zy Not sure if I'm missing something, but changing the viewport does not scale it to fullscreen like it was doing before. e.g. If I set the viewport to 1280x720, then instead of filling the entire screen like I was hoping for, it just draws to a smaller 1280x720 area in the top corner of the full 1920x1280 screen. This is basically what I'm doing:
```c#
_graphicsDeviceManager.IsFullScreen = true;
_graphicsDeviceManager.HardwareModeSwitch = false;
_graphicsDeviceManager.ApplyChanges();
_graphicsDeviceManager.GraphicsDevice.Viewport = new Viewport(0, 0, 1280, 720);
```
Is this right?

cra0zy:
You know, sometimes I forget some important stuff like the difference in what viewport is in GL and XNA... so in short there are 2 ways that I know of to scale your spritebatch game:
 1. Create scale matrix and feed it to transform component in your spritebatch begin methods
 2. Create a rendertarget2d and render everything to it, then just render it stretched (or scaled)

MichaelDePiazzi:
@cra0zy I was afraid you were going to say that. Method 1 is not practical at all for me. Method 2 could be an option, but it's gonna require a few changes and is going to add another fullscreen blit which isn't really ideal.

I'm kinda surprised being able to specify a different resolution for soft fullscreen was considered a bug. To me, this ability was a benefit of the mode and was how I'd expect it to work. Many other games allow you to specify a different resolution in soft fullscreen mode. It kinda sucks that the user would have to now emulate this behaviour under MonoGame... :/

Jjagg:
There was a bit of discussion about how to handle backbuffer/viewport resolution mismatch in #5530. Read down from https://github.com/MonoGame/MonoGame/issues/5530#issuecomment-283744878 if you're interested. tldr: scaling is more convenient, but forcing desktop resolution is more portable. No real decision has been made, so I just implemented what I thought was best.

MichaelDePiazzi:
> tldr: scaling is more convenient, but forcing desktop resolution is more portable.

Ok. I can't really comment from a portability point of view as I've only been dealing with Windows DirectX so far.

> No real decision has been made, so I just implemented what I thought was best.

Fair enough. I'll probably wait until a decision's been made then before I go changing how this works in my game. In the meantime, I'll have to maintain the previous behaviour and try fix the soft-fullscreen issues I was having separately.

Thraka:
Whichever way it works, please make sure it works the same for each backend. I don't want to have to code specific things for if it's using OpenGL or DX.

Jjagg:
Absolutely, consistency across platforms  is a high priority.

Jjagg:
WinForms/DX is now working correctly in this branch in all cases I tested, including switching resolutions at full screen and switching between hard and soft fullscreen!

Jjagg:
Anyone interested in this, please do test this and report if there are any issues left.
Here's the file I used for testing this PR: https://gist.github.com/Jjagg/2a30c9214d38bcc5ec79665ff762df6b

Note that this locks resolution in soft full screen to the desktop resolution.

I still need to test UWP before this is safe to merge.

Jjagg:
So UWP (tested on win10 desktop) is not doing things right at all... :/ It correctly resizes the backbuffer when the window is resized (even during resizing), but changing the back buffer size does not change the window size and it does not seem to do mode switched full screen at all. The behaviour in the develop branch and this branch is exactly the same though, so it's not a blocker.

So this is good to go IMO! :) 
Again, anyone please test this with MG DX to confirm it's good for merge! You can use the gist linked in my previous comment for a quick test.

When this is merged I'll tackle the windowing in DesktopGL and make sure its behaviour is the same as DX.

nkast:
Why clamp the backbuffer size to display size? Did XNA had a similar limitation? 

In fullscreen the backbuffer will be scaled automatically to the swapchan size, no mater if it's lower or higher resolution. In window mode it will look strange but nonetheless that's up to the dev.

Jjagg:
I was experimenting with this because Windows clamps the window size to the desktop resolution. That makes the backbuffer and window sizes mismatch and the back buffer will be scaled. XNA does not clamp back buffer size though, so I will remove the clamping.

zigzag312:
Good work @Jjagg. I'll test on MG DX and report back.

zigzag312:
My test results: Resizing & maximizing window now work without issues, which is great. Full screen toggle also works without issues in HW mode.

First minor issue I found is that when using WinKey+Left/Right shortcuts (in Win 10) to move game window, won't fire size changed events and everything will be squashed & stretched when window is resized and snapped to the edge. Very low priority issue IMO. 

Other issues are related to how MG handles multiple monitors (may be better fit for another PR?). MG is only using first adapter/monitor, even if game window is on another monitor. Toggling to fullscreen in SW mode on second display results in wrong resolution as we are querying CurrentDisplayMode on first display. Is SW fullscreen resolution limited to single resolution (only current desktop resolution) by design? Also querying SupportedDisplayModes when game is on second display returns wrong results (values from 1st display).

Jjagg:
Thanks @zigzag312!

> when using WinKey+Left/Right shortcuts

That shouldn't be too hard to fix. Maybe we can detect this through WinForms events, otherwise I'll take a look what Win32 events are passed.

> Other issues are related to how MG handles multiple monitors

I didn't test with multiple monitors, so it's good that you did!

>  Toggling to fullscreen in SW mode on second display results in wrong resolution as we are querying CurrentDisplayMode on first display.

> Also querying SupportedDisplayModes when game is on second display returns wrong results (values from 1st display).

These should be easy to fix. I'll fix these in this PR.

> Is SW fullscreen resolution limited to single resolution (only current desktop resolution) by design?

Yes. That is, there has not really been a final decision on this matter, but I opted for the behavior as it is implemented. Rationale for the current behavior is that having a backbuffer size different than window size is not available on all platforms and the platforms that do support it might not treat it the same way. So it makes consistency and portability across platforms hard or even impossible.

zigzag312:
> there has not really been a final decision on this matter, but I opted for the behavior as it is implemented. Rationale for the current behavior is that having a backbuffer size different than window size is not available on all platforms and the platforms that do support it might not treat it the same way.

What if MG changed desktop resolution and reverted back to original resolution on exit (on supported platforms)? That way game could run at any resolution in SW full screen while backbuffer size would remain same as window size.

Jjagg:
I don't think that's possible and I also don't think that's desirable. IMO soft full screen should not do mode switching.

I'd like to state an additional argument regarding up/downscaling in favor of forcing back buffer size to desktop resolution. There are choices MG has to make (if there even is a choice, platform constraints may force a decision) when rescaling the back buffer to fit the window that some games may want to do differently. We could either stretch or letter/pillarbox when rendering and there is a choice in filtering technique as well (most notably pixelated games will want to do this differently). By forcing desktop resolution we defer this choice to the user (but also the burden of implementing it of course). The current API allows for users to implement all of this by rendering to a RenderTarget first (which IMO is good practice), then to the back buffer.

zigzag312:
You're probably right, I haven' t tough of rendering to RT first :)

Jjagg:
> First minor issue I found is that when using WinKey+Left/Right shortcuts (in Win 10) to move game window, won't fire size changed events and everything will be squashed & stretched when window is resized and snapped to the edge.

This is a bit crap :/ There's no way to distinguish this from the event raised *while* the user resizes the window. We get a WM_SIZE, but no WM_EXITSIZEMOVE which is used to determine that resizing ends (at least in the Mono implementation of WinForms). To work around this, we could check that the size after the event is substantially different from the previous known size of the window, but that is not fool proof of course... I don't think there's actually a reliable way to fix without resizing the back buffer while the user is resizing the window as well :(

Jjagg:
> Toggling to fullscreen in SW mode on second display results in wrong resolution as we are querying CurrentDisplayMode on first display.

> Also querying SupportedDisplayModes when game is on second display returns wrong results (values from 1st display).

Both are fixed with the last commit (it's not showing up on this page for some reason though), that updates the `GraphicsAdapter` of the `GraphicsDevice` after the window is moved.

Jjagg:
We could also relatively easily handle runtime resolution changes by using `Microsoft.Win32.DisplaySettingsChanged`. I did not do that yet because XNA does not, but it can cause issues with soft fullscreen when people change desktop resolution at runtime, so it would be nice if we handled it.

Jjagg:
`SwapChain.ContainingOutput` is failing on the build server because it's headless. Not sure how to best work around that.

Jjagg:
By the way, in general there are two ways to approach the syncing between window and back buffer size when the user changes presentation parameters. 

1. We change the window size first, then get the client bounds and update the backbuffer to match, then let the window throw ClientSizeChanged event if required.
2. [what this PR implements] Predict the window size after applying the changes and use that to set the back buffer, then update the window and throw ClientSizeChanged if required. I can't really remember why I went this way, looking back at this option 1 actually seems more robust.

Thoughts?

zigzag312:
> There's no way to distinguish this from the event raised while the user resizes the window. We get a WM_SIZE, but no WM_EXITSIZEMOVE which is used to determine that resizing ends. 

I did some testing of WM messages. When resizing window with mouse we 1st get WM_ENTERSIZEMOVE, then a bunch of WM_SIZE messages and when user finishes resizing WM_EXITSIZEMOVE. If keyboard shortcuts are used we only get WM_SIZE message. 

So WM_SIZE probably could be used to detect resizing, if it is not between WM_ENTERSIZEMOVE & WM_EXITSIZEMOVE messages. Something like: private bool _isResizing is set to true on WM_ENTERSIZEMOVE and to false on WM_EXITSIZEMOVE. WM_SIZE is only ignored if _isResizing is true.

Jjagg:
Good call! I'll see if I can get that working :)

zigzag312:
> Both are fixed with the last commit (it's not showing up on this page for some reason though), that updates the GraphicsAdapter of the GraphicsDevice after the window is moved.

👍 I can confirm it that this works now on my end too.

> 2. [what this PR implements] Predict the window size after applying the changes and use that to set the back buffer, then update the window and throw ClientSizeChanged if required. I can't really remember why I went this way, looking back at this option 1 actually seems more robust.

This prediction sounds a more fragile that the 1st method, but I don't have any real knowledge on this topic so I can't really say. On the other hand if it ain't broke, why fix it :)

Jjagg:
Thanks @zigzag312 :) I guess you're right, things are working as is so it's fine. 

About detecting resolution changes, using `Microsoft.Win32.DisplaySettingsChanged` might not be a good idea because it's obviously not available in .NET Core. I'm thinking we can manually check each frame if things changed using DXGI. That should be acceptable, right? We can also check the number of connected displays, when that changes the collection of supported display modes should be updated. Anyways, like I said before I'll leave that out of this PR.

zigzag312:
There is [WM_DISPLAYCHANGE message](https://msdn.microsoft.com/en-us/library/dd145210(VS.85).aspx) that might be the same thing as using using `Microsoft.Win32.DisplaySettingsChanged`. [Here is an example](https://stackoverflow.com/questions/11983111/detect-when-new-display-is-connected/11985093#11985093) how to parse parameters of this message.

Jjagg:
Thanks! I didn't know about that :)

Jjagg:
Build is passing! Can anyone do a code review?

RefactoredGames:
Can this be merged now..?

zigzag312:
Could this be merged @tomspilman?

KonajuGames:
Merging.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -220,8 +220,13 @@ public GraphicsDevice(GraphicsAdapter adapter, GraphicsProfile graphicsProfile,
             _graphicsProfile = graphicsProfile;
             Setup();
             GraphicsCapabilities = new GraphicsCapabilities();
-            GraphicsCapabilities.Initialize(this);
-            Initialize();
+            GraphicsCapabilities.Initialize(this);
+
+#if WINDOWS
+            CorrectBackBufferSize();
+#endif
```
tomspilman: Can we do this in a better way that doesn't involve an `#if WINDOWS` here?  We are trying to reduce the `#if`s and not increase them.

Jjagg: We could do it with partial stuff, but I think it would be good if all platforms eventually implemented this method. The #if would be temporary. 

_Sent from my Motorola Moto G (5) using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

cra0zy: Why not just move the method to: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs#L134

cra0zy: Oh wait, you are also calling it on reset, yea, partial method would be the best IMO

vpenades:
This PR has modified heavily MonoGame.Content.Builder.targets with an unexpected consequence:

When compiling a project with missing content.mgcb (because it's not needed)  MSBuild gets stuck!, as if it's doing some sort of infinite loop!

Steps to reproduce:
1. Create a new Monogame Windows Project
2. Delete content.mgcb
3. Compile

KonajuGames:
@vpenades This PR did not touch the .targets file at all.  You are likely looking for #4997.

vpenades:
@KonajuGames  ah, yes, it might be, my mistake then

KonajuGames:
This PR may have a side effect shown in #5885.  Alt+tabbing from fullscreen results in the taskbar button for the app disappearing, i.e. the window closed, but the process is still running.  I'm having a quick look at this now, but if anyone else can help, that would be great too.

KonajuGames:
I've got a fix that I'll be submitting now.  When it was deactivating the process while in exclusive full-screen mode, it wasn't changing the swapchain out of exclusive mode.  It now forces it out of exclusive mode and minimizes the window.

KonajuGames:
Well, that was DirectX fixed.  GL doesn't exit exclusive mode either, or remove the window.  Fixing that now.

KonajuGames:
Fixed GL. Set the SDL hint to minimize on focus loss when going to exclusive full-screen mode, otherwise the hint is cleared.

KonajuGames:
#5887 submitted to fix DirectX and OpenGL.

vpenades:
From past experiences, I know handling windowed/fullscreen is a total nightmare.

Regarding Swap Chain buffer size matching the actual window... instead of depending on window resize events, wouldn't be more robust to check the SwapChain size against the actual window size at the beginning of every render frame?

PR #5587, opened 2017-03-15
Update the Tests definition so we can run the Linux unit tests on Mac.

dellis1972:
Fix the code in SDLWindow so we can handle the situation where we
do not have an EntryPoint assembly. This happens when we are running
unit tests.

Note this requires PR 101 for the audio tests to work.
There are a number of failures (28) that need to be resolved before this
gets enabled on the build bot.

https://github.com/MonoGame/MonoGame.Dependencies/pull/101

PR #5589, opened 2017-03-16
[DesktopGL] Don't call sizechanged when changing size from code

cra0zy:
@dellis1972 This should fix your test in #5587


dellis1972:
This looks ok. I would like to test it on a game to make sure though. It does indeed fix the unit test though :)

dellis1972:
@cra0zy slight problem with this, when starting a game in fullscreen it looks like the viewport (or client size) is not set correctly (probably because we don't raise the event).. 

I'll try to knock up a demo.

Jjagg:
That's not related to the event. That must have been an issue before the changes here too.

dellis1972:
This looks good

@mrhelmut any objections? 

@tomspilman I can hold off merging this unit 3.6.1 is done.

mrhelmut:
This is fine! Sorry for the late reply, I was off the grid.

PR #5596, opened 2017-03-18
Added Switch as a platform.

tomspilman:
This PR adds the Nintendo Switch as an official platform.  To get access to the MonoGame Switch repository you need to first [become a licensed as a Switch developer](https://developer.nintendo.com/) then contact me at `tom@sickheadgames.com`.

Also cleaned up some platform references in the docs.


KonajuGames:
Merging.

PR #5603, opened 2017-03-20
Make README bullets consistent

cra0zy:
Mostly did this to fix supported platforms list, which really didn't like the use of both type of bullets:

bullets

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,12 @@ One framework for creating powerful cross-platform games.  The spiritual success
 
 [![Join the chat at https://gitter.im/MonoGame/MonoGame](https://badges.gitter.im/MonoGame/MonoGame.svg)](https://gitter.im/MonoGame/MonoGame?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
- - [Build Status](#build-status)
- - [Supported Platforms](#supported-platforms)
- - [Support and Contributions](#support-and-contributions)
- - [Source Code](#source-code)
- - [Helpful Links](#helpful-links)
- - [License](#license)
+ * [Build Status](#build-status)
+ * [Supported Platforms](#supported-platforms)
+ * [Support and Contributions](#support-and-contributions)
+ * [Source Code](#source-code)
+ * [Helpful Links](#helpful-links)
+ * [License](#license)
  
 
 ## Build Status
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -31,19 +31,19 @@ Our [build server](http://teamcity.monogame.net/?guest=1) builds, tests, and pac
 
 We support a growing list of platforms across the desktop, mobile, and console space.  If there is a platform we don't support, please [make a request](https://github.com/MonoGame/MonoGame/issues) or [come help us](CONTRIBUTING.md) add it.
 
-- Desktop PCs
- * Windows Store Apps (8, 8.1 and 10)
```
tomspilman: Didn't we drop Windows 8 support recently?  @dellis1972 ?  Same with Windows Phone 8?

tomspilman:
Maybe we should copy this platform list to the docs too...

https://github.com/MonoGame/MonoGame/blob/develop/Documentation/what_is_monogame.md

README.md
```diff
@@ -31,19 +31,19 @@ Our [build server](http://teamcity.monogame.net/?guest=1) builds, tests, and pac
...
- * PlayStation Vita
- * Xbox One (both UWP and XDK)
+ * Desktop PCs
+   * Windows Store Apps (8, 8.1 and 10)
+   * Windows (OpenGL & DirectX)
+   * Linux (OpenGL)
+   * Mac OS X (OpenGL)
+ * Mobile Devices
+   * Android (OpenGL)
+   * iOS (OpenGL)
```
tomspilman: What about saying "iPhone/iPad/iTV" to make it clear we support all these devices?

README.md
```diff
@@ -31,19 +31,19 @@ Our [build server](http://teamcity.monogame.net/?guest=1) builds, tests, and pac
...
- * Android (OpenGL)
- * iOS (OpenGL)
- * Windows Phone (8, 8.1 and 10)
-- Consoles (for registered developers)
- * PlayStation 4
- * PlayStation Vita
- * Xbox One (both UWP and XDK)
+ * Desktop PCs
+   * Windows Store Apps (8, 8.1 and 10)
+   * Windows (OpenGL & DirectX)
```
tomspilman: Maybe say "Windows (7, 8, and 10 with OpenGL & DirectX)"?

cra0zy: Actually the current wording is better than that because we are listing platforms, not systems.

cra0zy:
@tomspilman All good now?

README.md
tomspilman: @cra0zy - I guess that is a point to make.

Do we list platforms (like iOS) or systems (like iPhone/iPad)?  In some cases it is both like PS4 is both a platform and a system.

Looking at Unity...

  https://unity3d.com/unity/multiplatform

They sort of mix and match as well...  Android and Android TV are two different platforms to them.  Same with iOS and tvOS.

What we can learn there is splitting them by category is best.

  - Desktop
  - Mobile
  - Consoles
  - Smart TVs
  - VR
 - Web

We're already starting to do that... so we're maybe on the right track here.

KonajuGames:
Merging.

README.md
DDReaper: We only dropped base Windows 8.0 / WP 8.0
Windows 8.1 / Windows Mobile 8.1 still exist

PR #5607, opened 2017-03-22
Use GetTypeInfo code on all platforms

cra0zy:
GetTypeInfo code works on all platforms... so there really is no reason not to use it on all platforms.


cra0zy:
#5732 does this better, closing.

PR #5608, opened 2017-03-22
[SDL] Use existing variable for pointing to SDL in OpenGL.SDL.cs

cra0zy:


KonajuGames:
No need for two copies of the same string, though they will get combined into one string instance by the compiler.

PR #5609, opened 2017-03-22
Fix up some weird Effects crashes.

dellis1972:
One of the samples I am porting hit this issue where the
EffectParameter.SetValue (int) for Single/Int parameters
as well as for Scalar and Vector parameters.

There was also a problem with access the Element[i] array in
EffectParameter.SetValue(Single[]). In this case it was crashing
event though the Data object was correct. This commit
removes the use of Element in favour of Data which other Setters
use.

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/EffectParameterCollection.cs
+++ b/MonoGame.Framework/Graphics/Effect/EffectParameterCollection.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Xna.Framework.Graphics
                 // TODO: Add a name to parameter lookup table.
 				foreach (var parameter in _parameters) 
                 {
-					if (parameter.Name == name) 
+					if (parameter.Name == name || parameter.Name == name+"Sampler") 
 						return parameter;
 				}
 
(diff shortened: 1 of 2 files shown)


dellis1972:
@tomspilman @KonajuGames this is a weird one. I hit these issues when porting a game with a number of complex shaders. It effected Dx and GL at the same places. If you have a better idea or want to investigate yourself let me know I can pass on the shaders.

dellis1972:
@tomspilman @KonajuGames any thoughts on this? Do we have any unit tests for Effects that run at runtime? If so I can probably port the .fx file over to a test.

MonoGame.Framework/Graphics/Effect/EffectParameterCollection.cs
```diff
@@ -42,7 +42,7 @@ public int Count
                 // TODO: Add a name to parameter lookup table.
 				foreach (var parameter in _parameters) 
                 {
-					if (parameter.Name == name) 
+					if (parameter.Name == name || parameter.Name == name+"Sampler")
```
tomspilman: This i am fairly sure is not correct behavior or at least something that should be fixed in the shader generation and not at runtime.  I would want to see some unit test to verify this.

tomspilman:
> Do we have any unit tests for Effects that run at runtime?

We have some unit tests that render with custom shaders to test them, but not really ones testing the correct functionality of `EffectParameter` setter and getter methods.  It would be a good unit test to add... at lease one covering the cases in this PR.

PR #5610, opened 2017-03-23
Don't throw an error if a channel exists already, just return it.

dellis1972:
While porting a game I came across a problem where the XNA pipeline
would quite happily process a model. In the MG pipeline the following
error was thrown.

	System.ArgumentException: Vertex channel with name Tangent0 already exists

I figure rather than throwing an error we should just return the
existing channel.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/VertexChannelCollection.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/VertexChannelCollection.cs
@@ -264,10 +264,13 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
                 throw new ArgumentOutOfRangeException("index");
             if (string.IsNullOrEmpty(name))
                 throw new ArgumentNullException("name");
-            // Don't insert a channel with the same name
-            if (IndexOf(name) >= 0)
-                throw new ArgumentException("Vertex channel with name " + name + " already exists");
-            var channel = new VertexChannel<ElementType>(name);
+            var existingIndex = IndexOf (name);
+            var exists = existingIndex >= 0;
+            var channel = exists ? (VertexChannel<ElementType>)channels[existingIndex] : new VertexChannel<ElementType>(name);
+            if (exists) {
+                channel.Items.Clear ();
+                channels.Remove (channel);
+            }
             if (channelData != null)
             {
                 // Insert the values from the enumerable into the channel


nkast:
Does XNA use the the model's Tangents or throw them away and generate new ones?

KonajuGames:
From a [forum reply](https://social.msdn.microsoft.com/Forums/en-US/2cc1637d-e11e-4c8c-99f1-9bb6e71a119a/bump-mapping-in-xna?forum=xnagamestudioexpress) by Shawn Hargreaves:

> XNA won't calculate tangents automatically, but does provide a helper function you can use to do so (in the content pipeline MeshHelper class).

It did load them from the model if it had them.

dellis1972:
ok this is wrong. we need to add the data to the existing channel I think..

tomspilman:
@dellis1972 - We really need some unit test to verify this change.

TheAppsStudio:
Hi Tom, just downloaded the development branch to test this as its blocking my project and it isnt in the dev branch I have download dev and applied the change but MGCB wont run. I have run protobuild and have the csproj files added for both Pipeline and Framework but when I run it just drops out. I do get a warning about targets saying: -
Severity	Code	Description	Project	File	Line	Suppression State
Warning		There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "SharpFont", "AMD64". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.	MonoGame.Framework.Content.Pipeline.Windows

tomspilman:
>  I run it just drops out. I do get a warning about targets saying:

@TheAppsStudio - Can you open a new issue for this.  Whatever issue you're having there it isn't caused by this PR.

tomspilman:
@dellis1972 - We really need some unit test to verify this change.

TheAppsStudio:
Thanks Tom for the reply and appologies I should of created a new issue for that. I did manage to get running in the end. I managed to compile the model and the tangent error is not an index out of range error. Once I get home I will attach the model which caused the issue and also start having a look as to why the index is out of bounds and report back to you. Let me know if I can help in any other way

TheAppsStudio:
Hi Tom, I replied to this last night but think the email didnt send. I found the issue and if you want a file to test with use any from the XNA Racing game starter kit. Basically the index needs setting to existing index currently it trys to add at the previous index but this has become invalid because the list of channels has one removed. Sorry this is a but of a waffle I will dig out the email tonight and put the full text and code example tonight when I get home

tomspilman:
@dellis1972 - This is a bad change... I tried it on XNA myself and found XNA will throw an exception on duplicate channels.  The real bug must be something else causing duplicate channels to occur.

Issue #2110, opened 2013-10-27
SharpDX.RawInput.dll


MarvinCZ:
Hello guys I just install monogame, create new project and run and i get tis error:
Could not load type 'SharpDX.Win32­.MessageFilter­Hook' from assembly 'SharpDX, Version=2.5.0.0, Culture=neutral, PublicKeyToken=nu­ll'
Can you tell me what to do?
btw XNA works fine in this way

nkast:
Same issue with sharpDX 2.6.3 and latest MG develop 3.6.1089
http://community.monogame.net/t/could-not-load-messagefilterhook/1311/2

robintheilade:
I realize this issue is old but I have an issue that looks like the same.

![image](https://cloud.githubusercontent.com/assets/1859029/26207377/a0867f98-3be7-11e7-894d-5541204380a4.png)

I just installed MonoGame 3.6.01625 and I use Visual Studio 2015.

I can confirm that by adding a reference to SharpDX.RawInput.dll the exception does not reappear.

Is there a reason the reference is not part of the Windows template?

nkast:
@robintheilade Which platform is that? Windows desktop ?
Is that a new project from the VS2015 template or an old project? Do you have any libraries that also reference Monogame and or SharpDX?

robintheilade:
@nkast, I'm on Windows 10 Creators Edition.
The project was created from scratch using the Windows template without the use of nuget packages. Just to verify, the csproj file contains this element "&lt;MonoGamePlatform&gt;Windows&gt;/MonoGamePlatform&gt;".
I have not modified the project in any way after it creation so no other libraries are loaded into the project.

I've extracted the template zip file (Windows.zip) and the MonoGameWindowsApplication.csproj also does not reference the missing assembly. The reference is also missing the [the current github version](https://github.com/MonoGame/MonoGame/blob/1c09add4a7407cf14690614d5ec36666d6c2a8ea/ProjectTemplates/VisualStudio2010/Windows/MonoGameWindowsApplication.csproj).

robintheilade:
Just checked a project I have been working on for a long time. It worked when I used MonoGame 3.5 and back then it was not referencing SharpDX.RawInput. That said the assembly was found in the bin folder so maybe I copied it there myself. If I delete it, it does not show up again if I recompile.

robintheilade:
Another update, maybe it will help.

[...]

PR #5611, opened 2017-03-23
Remove reference to SharpDX from project templates

nkast:
A MonoGame project don't need to reference SharpDX directly.


nkast:
NOTE: we might need to keep `SharpDX.XInput.dll` ref on some platforms and add it to the rest of the templates if necessary. #2110

tomspilman:
I don't remember why we directly referenced those in there.  It doesn't make sense to me now.

tomspilman:
Thanks @nkast !

Issue #5612, opened 2017-03-23
[W10/CoreApplication] Game loop doesn't run.


nkast:
Probably the bug surface due to the new way of event processing on StoreApps but the root of the problem I believe it to be #2084. The app stays forever in IsActive=false and the game loop doesn't run. 

That's a break from XNA behavior.
Update/Draw should always be called and it's up to the game dev to check Game.IsActive and skip physics , pause the game or whatever.

I suppose we should revert #2084. If there are issues with Suspend and Store certification those should be handled independently of lost of focus & minimize.

DDReaper:
I agree, should revert #2084 as I'm not sure that is resolving the issue what was reported and seems to have unintended consequences

nkast:
closed by #5613

PR #5613, opened 2017-03-23
Revert "added event handing when window nor active to allow runtime to suspend app"

nkast:
~~This reverts commit   ##76661cf~~.

fix for #5612


DDReaper:
You might want to refresh this PR too @nkast seems there have been other changes to this class since your PR

DDReaper:
Checking source, this fix has already been implemented @nkast

tomspilman:
Yeah so this my bug... i wrote that original ` if (Platform.IsActive)` code in the first push of UWP support.  My guess is i was following a UWP sample at the time and it did some sort of active check and i mimic'd that.

Merging... thanks @nkast !

PR #5614, opened 2017-03-23
Texture2D mipmap generation and population fixes

KonajuGames:
- Mipmaps are now generated correctly down to 1x1 in the content pipeline.
- Mipmaps are generated in the content pipeline in one location before it gets to the TextureProfile, avoiding duplication of functionality.
- Mipmaps are generated before compression, not during.
- Texture2D.OpenGL construction no longer generates mipmaps, but allocates mipmaps for population by SetData.
- Added more texture processing tests for non-square, power-of-two and non-square, non-power-of-two textures with mipmaps.
- Added Texture2D PlatformSetData overload for optimization of setting of an entire mip level.

Not tested on Android or iOS yet as I'm having Xamarin license issues.

### DIFF
--- a/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
@@ -614,11 +614,11 @@ namespace Microsoft.Xna.Framework.Graphics
                 {
                     if (_levelCount > 0)
                     {
-                        GL.TexParameter(TextureTarget.Texture2D, TextureParameterNameTextureMaxLevel, _levelCount - 1);
+                        GL.TexParameter(TextureTarget.Texture2D, SamplerState.TextureParameterNameTextureMaxLevel, _levelCount - 1);
                     }
                     else
                     {
-                        GL.TexParameter(TextureTarget.Texture2D, TextureParameterNameTextureMaxLevel, 1000);
+                        GL.TexParameter(TextureTarget.Texture2D, SamplerState.TextureParameterNameTextureMaxLevel, 1000);
                     }
                     GraphicsExtensions.CheckGLError();
                 }
--- a/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
@@ -70,73 +70,61 @@ namespace Microsoft.Xna.Framework.Graphics
         private void PlatformConstruct(int width, int height, bool mipmap, SurfaceFormat format, SurfaceType type, bool shared)
         {
             this.glTarget = TextureTarget.Texture2D;
+            format.GetGLFormat(GraphicsDevice, out glInternalFormat, out glFormat, out glType);
+            // GL texture is generated on first access in PlatformSetData
+        }
 
+        private void PlatformSetData<T>(int level, T[] data, int startIndex, int elementCount) where T : struct
+        {
+            int w, h;
+            GetSizeForLevel(Width, Height, level, out w, out h);
             Threading.BlockOnUIThread(() =>
             {
-                // Store the current bound texture.
-                var prevTexture = GraphicsExtensions.GetBoundTexture2D();
-
-                GenerateGLTextureIfRequired();
-
-                format.GetGLFormat(GraphicsDevice, out glInternalFormat, out glFormat, out glType);
-
-                if (glFormat == (PixelFormat)GLPixelFormat.CompressedTextureFormats)
+                var elementSizeInByte = ReflectionHelpers.SizeOf<T>.Get();
+                var dataHandle = GCHandle.Alloc(data, GCHandleType.Pinned);
+                // Use try..finally to make sure dataHandle is freed in case of an error
+                try
                 {
-                    var imageSize = 0;
-                    switch (format)
+                    var startBytes = startIndex * elementSizeInByte;
+                    var dataPtr = new IntPtr(dataHandle.AddrOfPinnedObject().ToInt64() + startBytes);
+                    // Store the current bound texture.
+                    var prevTexture = GraphicsExtensions.GetBoundTexture2D();
+
+                    if (prevTexture != glTexture)
                     {
-                        case SurfaceFormat.RgbPvrtc2Bpp:
-                        case SurfaceFormat.RgbaPvrtc2Bpp:
-                            imageSize = (Math.Max(this.width, 16) * Math.Max(this.height, 8) * 2 + 7) / 8;
-                            break;
-                        case SurfaceFormat.RgbPvrtc4Bpp:
-                        case SurfaceFormat.RgbaPvrtc4Bpp:
-                            imageSize = (Math.Max(this.width, 8) * Math.Max(this.height, 8) * 4 + 7) / 8;
-                            break;
-                        case SurfaceFormat.Dxt1:
-                        case SurfaceFormat.Dxt1a:
-                        case SurfaceFormat.Dxt1SRgb:
-                        case SurfaceFormat.Dxt3:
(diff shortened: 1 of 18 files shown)


KonajuGames:
I'm going to have to get this license issue sorted out before trying to fix Android and Mac build issues.

Test/ContentPipeline/TextureProcessorTests.cs
```diff
@@ -115,6 +115,100 @@ public void Mipmap()
...
+            var output = processor.Process(input, context);
+
+            Assert.NotNull(output);
+            Assert.AreEqual(1, output.Faces.Count);
+
+            var outChain = output.Faces[0];
+            Assert.AreEqual(5, outChain.Count);
+
+            var width = 23;
+            var height = 5;
```
tomspilman: Maybe use these to initialize the `PixelBitmapContent` above to avoid a mistake later.

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -170,7 +173,7 @@ public int Height
         /// <param name="elementCount"></param>
 		public void SetData<T>(T[] data, int startIndex, int elementCount) where T : struct
         {
-            this.SetData(0, null, data, startIndex, elementCount);
+            PlatformSetData(0, data, startIndex, elementCount);
```
tomspilman: Were the validation checks not in this method ever?

MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -159,7 +159,10 @@ public int Height
         /// <param name="elementCount"></param>
         public void SetData<T>(int level, Rectangle? rect, T[] data, int startIndex, int elementCount) where T : struct 
         {
-            this.SetData(level, 0, rect, data, startIndex, elementCount);
+            if (rect == null)
+                PlatformSetData(level, data, startIndex, elementCount);
```
tomspilman: I think we need to do the validation checks in this branch.

MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
```diff
@@ -70,73 +70,61 @@ public partial class Texture2D : Texture
         private void PlatformConstruct(int width, int height, bool mipmap, SurfaceFormat format, SurfaceType type, bool shared)
         {
             this.glTarget = TextureTarget.Texture2D;
+            format.GetGLFormat(GraphicsDevice, out glInternalFormat, out glFormat, out glType);
+            // GL texture is generated on first access in PlatformSetData
```
tomspilman: I'm unsure about this.  On DirectX after this method the texture fully exists and can be used to render with.  If you rendered with a setting to force rendering with a mipmap beyond the top level... it would work.  If you ran any DirectX debugger tool... you would see a valid constructed texture with a full mip chain.

Can we say the same about this for OpenGL?

KonajuGames: Rendering with a texture that had never had any data set on it? It's that something that we should/would want to support?

tomspilman: Actually a more definitive test.  Is video memory allocated or not?

On DirectX after `new Texture2D(...)` is called video memory is allocated.

I suspect in this case no video memory has been allocated at all.

Looking at the GL docs on `GL.TexImage2D` it says:

https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml
> data may be a null pointer. In this case, texture memory is allocated to accommodate a texture of width width and height height.

This suggests to me that not calling `GL.TexImage2D` means no texture memory has been allocated at all.  I've seen some suggestions that `GL.TexImage2D` can error out with `GL_OUT_OF_MEMORY`, but others suggest that it won't or at least doesn't happen on some platforms.

Also I would think that the texture memory needed for a texture would be allocated as one whole allocation for all mip levels.  Calling `GL.TexImage2D` once for each mip level would be a weird way to make a single allocation like that.  This suggests `GL.TexImage2D` doesn't allocate any memory and instead it is allocated sometime later when it has all the information needed.


Gosh GL is such a mess.

dellis1972:
@KonajuGames is this for 3.6.1?

KonajuGames:
The build failure makes no sense, and doesn't happen here.

```
Android\GamerServices\MonoGameGamerServicesHelper.cs(35,19): error CS0012: The type 'OpenTK.Graphics.ES20.TextureParameterName' is defined in an assembly that is not referenced. You must add a reference to assembly 'OpenTK-1.0, Version=1.0.0.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065'.
```

The line in question is a field declaration in a class.

```
        Texture2D signInProgress;
```

tomspilman:
@KonajuGames - I'll try updating to the latest Xamarin Studio on the build bot.

MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
tomspilman: @KonajuGames @dellis1972 @cra0zy 

I still feel like at the very least `GenerateGLTextureIfRequired();` should be called here.  Having texture creation not occur until `SetData` (or even `GetData`) is inconsistent with XNA behavior and even how MG works under DirectX and all our console platforms.

For example texture creation can fail...  is it better for this exception from the GL error to come from `SetData` or when `new Texture2D` is called?

tomspilman:
@KonajuGames - I'm pretty sure on XNA it is valid to set the texture data for all mip levels in a single `SetData`.  Does the GL path support this?  Do we have a unit test that covers this case of `SetData`?

KonajuGames:
I don't see anything in the XNA documentation or elsewhere about setting all mipmap levels in one call.  We don't have a test to cover that.  I'm certain it would throw an exception because the array is too big for the top mip level.  I'll give it a try, assuming we would just concatenate the data for mip levels.

MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
KonajuGames: I'll add the call to `GenerateTextureIfRequired` to the constructor.

KonajuGames:
> @KonajuGames - I'm pretty sure on XNA it is valid to set the texture data for all mip levels in a single SetData call. Does the GL path support this? Do we have a unit test that covers this case of SetData?

XNA throws an ArgumentException, saying the data is too large or too small for this resource.

MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
KonajuGames: `GenerateTextureIfRequired` has been added to `PlatformConstruct`.  Running the unit tests on OpenGL, it gets a `GL_INVALID_VALUE` on the first call to `Texture2D.SetData()` on a simple 64x64 Color texture with no mipmaps.  Spent a while trying to track it down without luck.  Should check if it is happening on develop branch.

KonajuGames: Turns out the calls to `glTexImage2D` and `glCompressedTexImage2D` with null data pointers at texture creation are necessary, otherwise `SetData()` fails with `INVALID_VALUE` and it complains about the texture format.

tomspilman:

mrhelmut:
This PR broke DesktopGL ability to load ```Texture2D``` from a thread.

It results in a GL error (code 1282) here: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs#L645

@KonajuGames

tomspilman:
@KonajuGames - Will you have time to look at this soon?

Issue #358, opened 2012-03-06
Antialiasing Support (GraphicsDevice.PreferMultiSampling)


ghost:
All platforms need to have antialiasing (re-?)enabled.  Pointers [here](http://www.gandogames.com/2010/07/tutorial-using-anti-aliasing-msaa-in-the-iphone/) and [here](https://devforums.apple.com/thread/45850).

aienabled:
I'm using DirectX11 and changing `GraphicsDevice.PresentationParameters.MultiSampleCount` during runtime has no effect (`Apply()` is called).
However, if I set this in MonoGame source code (method `GraphicsDeviceManager.Initialize()`) before creating the graphics device, it works fine.
So the issue is only with applying the multi-sampling after initialization.

GuilhermeAmorim93:
I'm using DirectX11 too,
But changing it during Game.Initialize() with `GraphicsDevice.PresentationParameters.MultiSampleCount = 8; GraphicsDeviceManager.Apply();`
causes nothing to be rendered. 

Any value other than 1 causes this to happen.

Noemata:
Latest build still has this issue.

SeriousMartin:
This problem still persists with the current version (https://github.com/MonoGame/MonoGame/commit/dd2e6cb79ecf6cb553582c402635d13faf9e9596 with #5477 merged). Any idea what could cause this?

zigzag312:
@SeriousMartin On what platform and are you using render targets? #5477 fixes only one small part of AA issues.

Jjagg:
@SeriousMartin You mean for a DirectX project? If so can you check the value of `MaxMultiSampleCount` in `GraphicsDevice.GraphicsCapabilities` for your `GraphicsDevice` using a debugger. Are you setting `GraphicsDeviceManager.PreferMultiSampling = true` (either in your constructor or by calling `ApplyChanges` afterwards) or using the `OnPreparingDeviceSettings` event? Note that the method mentioned above that directly changes `GraphicsDevice.PresentationParameters.MultiSampleCount` does not and should not work.
The three correct ways for modifying multisample count - and `PresentationParameters` in general - are the following:

- Setting `GraphicsDeviceManager.PreferMultiSampling` will set multisample count to the maximum supported by the GPU
[...]

PR #5617, opened 2017-03-24
Fixed DirectX back buffer update when multisampling changes

zigzag312:
Fixes issue of multisampling not working & nothing being drawn when multisampling is changed mid-game. https://github.com/MonoGame/MonoGame/issues/358#issuecomment-288793036 https://github.com/MonoGame/MonoGame/issues/358#issuecomment-288967105

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -729,7 +729,10 @@ namespace Microsoft.Xna.Framework.Graphics
                 PresentationParameters.MultiSampleCount);
 
             // If the swap chain already exists... update it.
-            if (_swapChain != null)
+            if (_swapChain != null
+                // check if multisampling hasn't changed
+                && _swapChain.Description.SampleDescription.Count == multisampleDesc.Count
+                && _swapChain.Description.SampleDescription.Quality == multisampleDesc.Quality)
             {
                 _swapChain.ResizeBuffers(2,
                                         PresentationParameters.BackBufferWidth,
@@ -741,6 +744,10 @@ namespace Microsoft.Xna.Framework.Graphics
             // Otherwise, create a new swap chain.
             else
             {
+                // Dispose of old swap chain if exists
+                if (_swapChain != null)
+                    _swapChain.Dispose();
+
                 // SwapChain description
                 var desc = new SharpDX.DXGI.SwapChainDescription()
                 {
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -744,9 +744,15 @@ namespace Microsoft.Xna.Framework.Graphics
             // Otherwise, create a new swap chain.
             else
             {
+                var isWindowed = true;//TODO: should we use PresentationParameters here?
                 // Dispose of old swap chain if exists
                 if (_swapChain != null)
+                {
+                    isWindowed = !_swapChain.IsFullScreen;
+                    // Before releasing a swap chain, first switch to windowed mode
+                    _swapChain.SetFullscreenState(false, null);
                     _swapChain.Dispose();
+                }
 
                 // SwapChain description
                 var desc = new SharpDX.DXGI.SwapChainDescription()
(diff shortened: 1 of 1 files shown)


Jjagg:
Looks like issues on the Mac build bot
```
 Error while applying patch: Failed to chmod: /Applications/buildAgent/work/59a4b66548a5cef5/Installers/MacOS/Scripts/Pipeline/postinstall 
```

zigzag312:
Why does it fail? Issues with build server?

Jjagg:
Yeah, definitely not because of changes in this PR. @tomspilman Any idea what's up?

tomspilman:
Yeah something weird happened on the build server... not sure what.  I re-ran the build and it is working now.

Jjagg:
The changes here cause issues when in full screen mode. Changing ms count in soft full screen doesn't do anything and in mode switched full screen it causes a crash. I should mention I didn't test this branch directly, but applied the changes to my branch from #5585. Maybe things are better in this branch, but this and #5585 should play nice together so I'll see if I can fix my PR so it works correctly in combination with the changes here.

I can confirm that this does work when not in full screen.

zigzag312:
Good catch. Previously _swapChain was never recreated, but I haven't found any method similar to SwapChain.ResizeBuffers(...) that would update multisampling count on swap chain. So old swap chain is disposed and recreated from scratch, but I am not sure it should be done this way. 

It crashes when it tries to create new swap chain.
`_swapChain = new SwapChain(dxgiFactory, dxgiDevice, desc);`

zigzag312:
Issues when in full screen mode should be fixed now.

zigzag312:
Can someone review this:  https://github.com/zigzag312/MonoGame/blob/61c6aca577c7b0d8115fd01214a429cb7e4bdb19/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs#L747

Jjagg:
Yes, you can use PresentationParameters because the value in there is what we want the full screen state to be. So should be `isWindowed = pp.IsFullScreen && pp.HardwareModeSwitch`. I vaguely recall initializing the swap chain in full screen causing some issues though, so maybe it's better to initialize it windowed and then call `SetHardwareFullscreen`.

zigzag312:
@Jjagg Do you have any suggestion where to put SetHardwareFullscreen?

It seems you remember correctly :) I found following remarks regarding creating full screen swap chain:

- If a Windows Store app calls CreateSwapChain with full screen specified, CreateSwapChain fails.
- Because the target output can't be chosen explicitly when the swap chain is created, we recommend not to create a full-screen swap chain. This can reduce presentation performance if the swap chain size and the output window size do not match. Here are two ways to ensure that the sizes match:
  - Create a windowed swap chain and then set it full-screen using IDXGISwapChain::SetFullscreenState.
  - Save a pointer to the swap chain immediately after creation, and use it to get the output window size during a WM_SIZE event. Then resize the swap chain buffers (with IDXGISwapChain::ResizeBuffers) during the transition from windowed to full-screen.

[IDXGIFactory::CreateSwapChain method (Windows)](https://msdn.microsoft.com/en-us/library/windows/desktop/bb174537(v=vs.85).aspx#NOTES_FOR_WINDOWS_STORE_APPS)

Jjagg:
> Do you have any suggestion where to put SetHardwareFullscreen?

It's already defined here: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs#L676

zigzag312:
Changed to always create a windowed swap chain and then set it full-screen if needed. 

It only changes new swap chain to full screen, if old swap chain was already full screen. So it doesn't conflict with EnterFullScreen() method that needs to save Form location before entering full screen.

Jjagg:
@KonajuGames Could you take a look at getting this merged? I can confirm this fixes changing MSAA at runtime.

tomspilman:
@zigzag312 @Jjagg - Could you guys look at...

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/SwapChainRenderTarget.cs

... and see if it needs a similar fix?

Also would be nice one day for that to work on OpenGL too.

tomspilman:
Thanks @zigzag312 ... merging!

zigzag312:
@tomspilman SwapChainRenderTarget is missing clamping on preferredMultiSampleCount. But I don't see other issues like here, as that swapchain never changes once created and is always windowed. I can make PR for clamping.

Issue #5131, opened 2016-08-22
Windows Phone touch bug


BartekPiura:
Hi,

after some time after game launch, in windows phone 8.0 app, when i get touches by TouchPanel.GetState(), i get extra touches, in place where i don't touch

do you know how to fix that?

tomspilman:
We've removed Win8/8.1 support in https://github.com/MonoGame/MonoGame/pull/5809.

Issue #3172, opened 2014-11-14
TouchCollection Not Clearing


adeeb1:
When I tap on the screen with a few fingers and simultaneously click with a mouse, a TouchLocation with the Moved status gets stuck in the TouchCollection and is never cleared out. All subsequent touches are still registered, but index 0 of the TouchCollection continues to return that same TouchLocation with the Moved status, and the time it was registered remains the same.

This was originally reported at the following URL: http://community.monogame.net/t/touchcollection-not-clearing/1425

A test project can be downloaded from the following URL: http://www.mediafire.com/download/lljjv28fi1fo77a/TouchCollection_TestProject.zip

Simply open up the project and run it. Whenever you tap, you will hear a sound. If you swipe, you will hear a different sound.

Reproduction Steps:

Tap and swipe on the screen like a madman with one hand while clicking on the screen using a mouse on the other hand. After a while, you should stop hearing the touch sounds. This means that you have successfully reproduced the issue. Some debug information is displayed in the Output window of Visual Studio.

tomspilman:
We've removed Win8/8.1 support in https://github.com/MonoGame/MonoGame/pull/5809.

PR #5620, opened 2017-03-24
Remove forgotten touches

zigzag312:
Quickfix for a problem I encontered. Sometimes old touch events weren't removed and got stuck indefinitely. This crude fix removes orphaned touches that weren't updated for at least two frames.

I was testing on windows with EnableMouseTouchPoint & EnableMouseGestures.

### DIFF
--- a/MonoGame.Framework/Input/Touch/TouchPanelState.cs
+++ b/MonoGame.Framework/Input/Touch/TouchPanelState.cs
@@ -48,8 +48,15 @@ namespace Microsoft.Xna.Framework.Input.Touch
             get { return _currentTimestamp; }
             set
             {
-                _previousTimestamp = _currentTimestamp;
-                _currentTimestamp = value;
+                if (_currentTimestamp != value)
+                {
+                    if (_currentTimestamp > _previousTimestamp)
+                        _previousTimestamp = _currentTimestamp;
+                    _currentTimestamp = value;
+                    // handle time rewinding
+                    if (_currentTimestamp < _previousTimestamp)
+                        _previousTimestamp = _currentTimestamp;
+                }
             }
         }
         private static TimeSpan _currentTimestamp;
--- a/MonoGame.Framework/Input/Touch/TouchPanelState.cs
+++ b/MonoGame.Framework/Input/Touch/TouchPanelState.cs
@@ -43,7 +43,17 @@ namespace Microsoft.Xna.Framework.Input.Touch
         /// <summary>
         /// The current timestamp that we use for setting the timestamp of new TouchLocations
         /// </summary>
-        internal static TimeSpan CurrentTimestamp { get; set; }
+        internal static TimeSpan CurrentTimestamp
+        {
+            get { return _currentTimestamp; }
+            set
+            {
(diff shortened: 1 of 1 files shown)


nkast:
Similar to #5382. 
The cause of the bug is that Mouse uses ID=1 when emulates touch and for some reason this let's multiple touch point in the queue.

zigzag312:
One source of the problem may be [here](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Windows/WinFormsGameWindow.cs#L247). 
```
// Don't process touch state if we're not active 
// and the mouse is within the client area.
if (!_platform.IsActive || !withinClient)
    return;
```
To reproduce bug click inside the window and while holding button down, move your mouse outside the window. Current touch handling system expects that Released or Invalid event will be **always** called. Maybe there are also other instances where that doesn't happen. Therefore this fix removes forgotten touches as a last resort. I think it shouldn't break anything.

nkast:
I also found #5518 which seems to fix the bug you describe.

jamesford42:
Btw, if its not clear from the name of my PR: it does fix this issue, and was written for that intention.

zigzag312:
I'm glad to see a proper fix for that case!

One thing that still worries me, is that the implementation of touch handling in MG seems to be fragile by design. It expects that every OS with any combination of HW, drivers will always, under all circumstances correctly report touch released or invalid events. Otherwise touch will get stuck. This PR is intended like safety net for that design.

I haven't done much testing yet, but there seems to be other issues of touch getting stuck #1360, #2103, #3172, #5131...

nkast:
> will always, under all circumstances correctly report touch released or invalid events. 

Can you give an example of an OS/API where this is a problem?

zigzag312:
No, but the question should IMO be, is the behavior MG expects, guarantied by the specs for all supported platforms under all circumstances? 

I haven't done any real research on this topic myself, but is there any harm done by this PR? It only checks for an invalid state that should never happen. Alternative to removing orphaned touch when detected, would be to throw an exception so that touch errors are caught faster by developers.

jamesford42:
Isn't there some _touchId dictionary and gesture state stuff that will become out of sync? At first glance i'm not sure this is really an appropriate place to be removing items. Normally all changes are driven through calls to AddEvent, i think.

nkast:
No. it can't be guarantied.
My question would be, how can you decide if a touchpoint is forgotten/invalid?

It's a good idea to have a failsafe to recover for the worst case, but [it doesn't work in this case](https://en.wikipedia.org/wiki/Halting_problem).  

For example, does Hold gesture work with this PR? You can't guaranty that you will get any `Moved` event between a `Press` and a `Release`.
Also does it work with TimeStep=false, if Update()/GetState() is called 600 times per second?

zigzag312:
I see. I incorrectly assumed MG is pooling touch state each frame. My bad. I apologize. I'll close this then.

nkast:
for what? No need to. I understand that you wanted this fixed asap both for you and for others. Touch has one or 2 bugs right now, IMO the focus should be on fixing the cause of those bugs.

PR #5624, opened 2017-03-27
Inverted the Platform(s) for TextureCubeReader, Texture3DReader and TextureReader to use ExcludePlatforms

dellis1972:


tomspilman:
Thanks @dellis1972 !

PR #5625, opened 2017-03-27
[Gamepad] Add Id and DisplayName properties

cra0zy:
Notes:
 - We need to use string for Id and not guid because not all platforms will return a guid.
 - ~~I like Id more than Identifier.. do tell me if you want it changed.~~
   - It's named `Identifier` now

CC. @tomspilman

### DIFF
--- a/MonoGame.Framework/Input/GamePad.SDL.cs
+++ b/MonoGame.Framework/Input/GamePad.SDL.cs
@@ -122,40 +122,90 @@ namespace Microsoft.Xna.Framework.Input
             if (!Gamepads.ContainsKey(index))
                 return new GamePadCapabilities();
 
-            if (Sdl.GameController.GetName(Gamepads[index].Device) == "Unknown Gamepad")
-                return new GamePadCapabilities
-                {
-                    IsConnected = true
-                };
+            var gamecontroller = Gamepads[index].Device;
+            var caps = new GamePadCapabilities();
+            var mapping = Sdl.GameController.GetMapping(gamecontroller).Split(',');
 
-            return new GamePadCapabilities
+            caps.IsConnected = true;
+            caps.DisplayName = Sdl.GameController.GetName(gamecontroller);
+            caps.Identifier = Sdl.Joystick.GetGUID(Sdl.GameController.GetJoystick(gamecontroller)).ToString();
+            caps.HasLeftVibrationMotor = caps.HasRightVibrationMotor = (Gamepads[index].HapticType != 0);
+
+            foreach (var map in mapping)
             {
-                IsConnected = true,
-                HasAButton = true,
-                HasBButton = true,
-                HasXButton = true,
-                HasYButton = true,
-                HasBackButton = true,
-                HasStartButton = true,
-                HasDPadDownButton = true,
-                HasDPadLeftButton = true,
-                HasDPadRightButton = true,
-                HasDPadUpButton = true,
-                HasLeftShoulderButton = true,
-                HasRightShoulderButton = true,
-                HasLeftStickButton = true,
-                HasRightStickButton = true,
-                HasLeftTrigger = true,
-                HasRightTrigger = true,
-                HasLeftXThumbStick = true,
-                HasLeftYThumbStick = true,
-                HasRightXThumbStick = true,
-                HasRightYThumbStick = true,
-                HasLeftVibrationMotor = true,
-                HasRightVibrationMotor = true,
-                HasVoiceSupport = true,
-                HasBigButton = true
-            };
+                var split = map.Split(':');
+                if (split.Length != 2)
+                    continue;
+
+                switch (split[0])
+                {
+                    case "a":
+                        caps.HasAButton = true;
+                        break;
+                    case "b":
+                        caps.HasBButton = true;
+                        break;
+                    case "x":
+                        caps.HasXButton = true;
+                        break;
+                    case "y":
+                        caps.HasYButton = true;
+                        break;
+                    case "back":
+                        caps.HasBackButton = true;
+                        break;
+                    case "guide":
+                        caps.HasBigButton = true;
+                        break;
+                    case "start":
+                        caps.HasStartButton = true;
+                        break;
+                    case "dpleft":
+                        caps.HasDPadLeftButton = true;
+                        break;
+                    case "dpdown":
+                        caps.HasDPadDownButton = true;
+                        break;
+                    case "dpright":
+                        caps.HasDPadRightButton = true;
+                        break;
+                    case "dpup":
+                        caps.HasDPadUpButton = true;
+                        break;
+                    case "leftshoulder":
+                        caps.HasLeftShoulderButton = true;
+                        break;
+                    case "lefttrigger":
+                        caps.HasLeftTrigger = true;
+                        break;
+                    case "rightshoulder":
+                        caps.HasRightShoulderButton = true;
+                        break;
+                    case "righttrigger":
+                        caps.HasRightTrigger = true;
+                        break;
+                    case "leftstick":
+                        caps.HasLeftStickButton = true;
+                        break;
+                    case "rightstick":
+                        caps.HasRightStickButton = true;
+                        break;
+                    case "leftx":
+                        caps.HasLeftXThumbStick = true;
+                        break;
+                    case "lefty":
+                        caps.HasLeftYThumbStick = true;
+                        break;
+                    case "rightx":
+                        caps.HasRightXThumbStick = true;
+                        break;
+                    case "righty":
+                        caps.HasRightYThumbStick = true;
+                        break;
+                }
+            }
+
+            return caps;
         }
 
         private static float GetFromSdlAxis(int axis)
--- a/MonoGame.Framework/Input/GamePadCapabilities.cs
+++ b/MonoGame.Framework/Input/GamePadCapabilities.cs
@@ -16,6 +16,22 @@ namespace Microsoft.Xna.Framework.Input
         public bool IsConnected { get; internal set; }
 
         /// <summary>
+        /// Gets the gamepad display name.
+        /// 
+        /// This property is not available in XNA.
+        /// </summary>
+        /// <value>String representing the display name of the gamepad.</value>
+        public string DisplayName { get; internal set; }
+
+        /// <summary>
+        /// Gets the unique identifier of the gamepad.
+        /// 
+        /// This property is not available in XNA.
+        /// </summary>
+        /// <value>String representing the unique identifier of the gamepad.</value>
+        public string Identifier { get; internal set; }
+
+        /// <summary>
         /// Gets a value indicating whether the controller has the button A.
         /// </summary>
         /// <value><c>true</c> if it has the button A; otherwise, <c>false</c>.</value>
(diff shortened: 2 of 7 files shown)


MonoGame.Framework/Input/GamePadCapabilities.cs
```diff
@@ -46,6 +46,10 @@ public struct GamePadCapabilities
     {
         public bool IsConnected { get; internal set; }
 
+        public string DisplayName { get; internal set; }
+
+        public string Id { get; set; }
```
tomspilman: I still prefer `Identifier`.

@KonajuGames @dellis1972 @mrhelmut @Jjagg - Any better ideas for this?

tomspilman: Also why is the setter public?  It likely shouldn't be.

cra0zy: > Also why is the setter public? It likely shouldn't be.

Error on my part.

tomspilman:
@KonajuGames @dellis1972 @mrhelmut @Jjagg 

Thoughts on this.  This allows for PC games in particular to display a reasonable controller display name as well as maybe customize behavior/icons based on the connected controller.

The next step would be new APIs for controller specific features.

MonoGame.Framework/Input/GamePadCapabilities.cs
dellis1972: I like Identifier.

cra0zy:
I renamed `Id` to `Identifier` for both Gamepad and Joystick.

MonoGame.Framework/Input/GamePadCapabilities.cs
tomspilman: Another thing... lets add some docs to these two new properties.

They should clearly state that this is a MonoGame extension.  Also that these strings can be `null` if the platform doesn't support identification of controllers (which is true of all our platforms other than SDL today).

tomspilman:
@cra0zy -

> Input\GamePad.Web.cs(35,48): error CS1061: 'Microsoft.Xna.Framework.Input.JoystickCapabilities' does not contain a definition for 'Id' and no extension method 'Id' accepting a first argument of type 'Microsoft.Xna.Framework.Input.JoystickCapabilities' could be found (are you missing a using directive or an assembly reference?)

cra0zy:
I know.

MonoGame.Framework/Input/GamePadCapabilities.cs
cra0zy: On consoles you can set gamepad ids to 1, 2, 3, 4, they don't need to be real ids.

cra0zy:
Well I just now realized that `GamePadCapabilities` has a `GamePadType`... 

@tomspilman Should I close or fix this?

tomspilman:
> Well I just now realized that `GamePadCapabilities` has a `GamePadType`...

I was aware of that... it is for a general type of controller and not specific controllers:

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.input.gamepadtype.aspx

cra0zy:
> So we would not be adding the entire list of SDL controllers to this enum or anything.

We just need to add PS4Controller enum....

On related side note, look at what is UWP doing with other gamepad types: https://docs.microsoft.com/en-us/uwp/api/windows.gaming.input

cra0zy:
PS. I added documentation for the rest of GamePadCapabilities class.

MonoGame.Framework/Input/GamePadCapabilities.cs
```diff
@@ -1,100 +1,180 @@
...
+        /// <summary>
+        /// Gets a value indicating whether the gamepad has voice input support.
+        /// </summary>
+        /// <value><c>true</c> if the gamepad has voice input support; otherwise, <c>false</c>.</value>
+		public bool HasVoiceSupport { get; internal set; }
+
+        /// <summary>
+        /// Gets a value indicating the type of gamepad.
+        /// </summary>
+        /// <value>A <<see cref="GamePadType"/> indicating the type of gamepad.</value>
```
Jjagg: Typo with the angular brackets here.

cra0zy: Fixed.

MonoGame.Framework/Input/GamePad.SDL.cs
```diff
@@ -122,39 +122,40 @@ private static GamePadCapabilities PlatformGetCapabilities(int index)
...
+                HasLeftTrigger = hasbutton,
+                HasRightTrigger = hasbutton,
+                HasLeftXThumbStick = hasbutton,
+                HasLeftYThumbStick = hasbutton,
+                HasRightXThumbStick = hasbutton,
+                HasRightYThumbStick = hasbutton,
+                HasLeftVibrationMotor = hasbutton,
+                HasRightVibrationMotor = hasbutton,
+                HasVoiceSupport = hasbutton,
+                HasBigButton = hasbutton
```
tomspilman: @cra0zy - Could we in a different PR make improvements here and actually query from SDL what buttons a gamepad has available?

MonoGame.Framework/Input/GamePadCapabilities.cs
```diff
@@ -1,100 +1,180 @@
...
+        /// <summary>
+        /// Gets a value indicating whether the gamepad is connected.
+        /// </summary>
+        /// <value><c>true</c> if the gamepad is connected; otherwise, <c>false</c>.</value>
+        public bool IsConnected { get; internal set; }
+
+        /// <summary>
+        /// Gets the gamepad display name.
+        /// </summary>
+        /// <value>String representing the display name of the gamepad..</value>
```
tomspilman: Double period at the end of the value description.

@KonajuGames @dellis1972 - Should we point out in the docs that this is a MonoGame extension to XNA?  If so how do we do this in some common?

dellis1972: perhaps add a tag to the documentation comments

```<api>XNA Compatible</api>```

```<api>MonoGame Only</api>```

Or something like that.

Comment (cra0zy) [inline MonoGame.Framework/Input/GamePadCapabilities.cs]: Note to self, don't write documentation at 3 in the morning.

**Comment by cra0zy** [comment]
I like how this started as a 30 line change, and now it's a 300 line change.

Anyway SDL now properly detects gamepad capabilities.

**Comment by tomspilman** [comment]
> I like how this started as a 30 line change, and now it's a 300 line change.

I did say "in a different PR" to avoid the complication.  Now we need some testing of this to ensure it works.  @mrhelmut ?

**Comment by cra0zy** [comment]
@tomspilman @dellis1972 What's up with Mac build server?

File: MonoGame.Framework/Input/GamePadCapabilities.cs (hunk 4 above)
Comment (tomspilman) [inline MonoGame.Framework/Input/GamePadCapabilities.cs]: @dellis1972 - Is `<api>` a valid doc tag?  If it doesn't appear in VS intellisense or in our docs online then it seems sort of useless.

> `<api>XNA Compatible</api>`

Well for sure we're not going to mark the XNA compatible members and types... as that is most everything.  We would just mark the things that are "MonoGame only" as that makes things much easier for us.

Could be something as simple as "MonoGame Only" at the end of the `<summary>` line like:

```
/// <summary> 
/// Gets the gamepad display name.  MonoGame only.
/// </summary> 
```
Since the summary is all that will appear in VS when writing code.

@KonajuGames ?

**Comment by tomspilman** [comment]
> What's up with Mac build server?

Not sure .... it has been doing this for the last few days.  I'll force it to clean sources again and rebuild.

File: MonoGame.Framework/Input/GamePadCapabilities.cs (hunk 4 above)
Comment (cra0zy) [inline MonoGame.Framework/Input/GamePadCapabilities.cs]: IMO that piece of information isn't important enough to be in summary.

Also we have several other APIs that weren't in XNA, and we are not marking them in any way.

Comment (tomspilman) [inline MonoGame.Framework/Input/GamePadCapabilities.cs]: > IMO that piece of information isn't important enough to be in summary.

I don't love having it in the summary...  but it is the only place to put it that makes it visible to you as you write code.

> Also we have several other APIs that weren't in XNA, and we are not marking them in any way.

I am fully aware... and I bring this up because I think it is a problem.  As someone that has ported around a dozen XNA/MG titles to consoles... most of which use XNA on Windows desktop releases... knowing if you are coding to an API that isn't cross platform with XNA is useful.

**Comment by dellis1972** [comment]
How about an Attribute which issues a warning that that api is not
compatible with XNA?

**Comment by tomspilman** [comment]
> How about an Attribute
That could be ok...  is there an attribute for that that VS/XS picks up?

**Comment by dellis1972** [comment]
The only one is [Obsolete("text")] that the compiler will turn into a
warning..
Alternatively a

#warning This  property is not available in XNA

might work.

**Comment by dellis1972** [comment]
problem is developers ignore warnings ;)

**Comment by cra0zy** [comment]
> The only one is [Obsolete("text")] that the compiler will turn into a
> warning..
> Alternatively a
> 
> #warning This  property is not available in XNA
> 
> might work.

1c8f3fd1255191e2981c97e9529dd238_hell-no-meme-gif-13jpg-hell-no-meme_500-366

**Comment by tomspilman** [comment]
Yeah... `ObsoleteAttribute` is inappropriate for this case and `#warning` is easily ignored or just messy if you are aware and intended to use that API.  Just a wart at the end of the `<summary>` in the docs seems best so far.

**Comment by cra0zy** [comment]
@tomspilman Can you force rerun of the test on this PR? Mac build server had an issue for a few days when this PR was submitted.

**Comment by cra0zy** [comment]
If someone can get around to merging this and #5601 that would be awesome. I need it since I want to add equal methods/operators and few extra constructors to `Joystick`, and also do the UWP implementation now that the new UWP SDK is out \o/.

File: MonoGame.Framework/SDL/SDL2.cs [hunk 5]
```diff
@@ -15,18 +15,17 @@ internal static class Sdl
...
+    private static string GetString(IntPtr handle)
     {
         if (handle == IntPtr.Zero)
             return "";
 
-        var ptr = (byte*) handle;
-        while (*ptr != 0)
-            ptr++;
+        int len = 0;
+        while (Marshal.ReadByte(handle, len) != 0)
```
tomspilman: Was there a bug here this is fixing?

cra0zy: No bug, I just removed `unsafe` keyword.

dellis1972: Has this been tested on Mac, Linux and Windows? I have a vague recollection that I tried this technique (or something similar) and it didn't work on one of the platforms. I could be wrong though.

cra0zy: It works just fine, I submitted the same code to Eto Forms several months ago and there have been no complaints so far... 

Why wouldn't it work?

cra0zy: Side note, if you want I'll remove this if it'll help merge the PR sooner.

tomspilman: @cra0zy - Yes please remove it.

We have enough problems with reliability and stability of the code base to not start changing things without good reasons.  There is nothing wrong with `unsafe` code.

cra0zy: Done.

tomspilman:
Thanks @cra0zy !

Issue #5567, opened 2017-03-10
Replacing System.Drawing for DesktopGL


cra0zy:
Most parts of System.Drawing are already gone, and now the last file that needs to be fixed would be: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs This needs to be done so we can get .Net Core supporting more easily when the time comes.

For replacing System.Drawing I would recommend going with either of these:
 - <a href="https://github.com/JimBobSquarePants/ImageSharp">ImageSharp</a> - it's fully managed so no need for any dependencies
 - ~~<a href="https://www.libsdl.org/projects/SDL_image/">SDL_image</a> - if we are going native we might as well pick a format that SDL understands~~
   - ~~bonus point: we get .ico format, useful for loading window icon~~

CC. @tomspilman @KonajuGames @dellis1972

tomspilman:
Well first... what are we using `System.Drawing` for?  Seems like for `Texture2D.PlatformFromStream()` and `Texture2D.SaveAsPng/JPEG()` mainly...  is there more that we need that saving/loading image data?

I see a better alternative which would take a little work from us.

We already had @disportium contribute PNG reading/writing here:

 https://github.com/MonoGame/MonoGame/tree/develop/MonoGame.Framework/Utilities/Png

The performance is a bit poor on consoles/mobile because of the design (lots of little array allocations, no use of pointers, etc).  I would love to see that improved and used on all platforms as it gives us a cross-platform portable PNG encoding/decoding.

It would be ideal to do the same with JPEG as there are reference encoders/decoders out there to work from.

cra0zy:
There is also: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Input/MouseCursor.SDL.cs#L40

And it needs to be a bmp format for SDL to be able to read it.

cra0zy:
Also it would be nice if this piece of code could load .ico files: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/SDL/SDLGameWindow.cs#L127

tomspilman:
> And it needs to be a bmp format for SDL to be able to read it.
> Also it would be nice if this piece of code could load .ico files: 

So this means we need:

  - Png reader/writer (done... but could use improvement)
  - Jpeg reader/writer
  - Bmp writer (and reader would be nice)
  - Ico reader.

Correct?

[...]

Issue #5529, opened 2017-03-01
SpriteBatch AlphaBlend doesn't look the same on OpenGL vs DirectX


Thraka:
Using the new 3.6 nuget packages, alphablending sprite (that is a raw PNG file, not sent through the pipeline) calls doesn't look right. Here is my begin

    Global.SpriteBatch.Begin(SpriteSortMode.Deferred, BlendState.AlphaBlend, SamplerState.PointClamp, DepthStencilState.DepthRead, RasterizerState.CullNone);

And here are the results.

![image](https://cloud.githubusercontent.com/assets/2672110/23485127/f18bad6e-fe8e-11e6-8948-cbb64d77145d.png)

To me it looks like on DirectX, it's figuring out the premultiply correctly while opengl doesn't see it and therefore doesn't use it.

tomspilman:
@dellis1972 @cra0zy @Jjagg - I suspect the png loading is doing something different on the DX platform vs. OGL.  Any ideas?

Thraka:
For reference, that is two sprites, a yellow box and a purple circle drawn on top.

willmotil:
With spritefonts on gl i can tell you premultiplied vs non-premultiplied set from graphics device does work even when the image is not set to pre-multiplied in the pipeline.

The final color on the right obviously has less blue and red intensity remember that alpha is transparency and transparency's ultimate effect for color itself of a object is to reduce it in the final result.
Where with pre-multiply there is no rgb reduction calculated as it has been pre-multplied by the alpha.
So while this appears to be the case are you sure it's not your settings i think you must prove this case with settings you have toggled and cases or a working example.

I normally don't use premultiplied at all its like taking away your ability to manipulate data you might need to thus giving yourself a headache.

Thraka:
I'm not using any pipeline, i'm loading texture2d directly from a png. To make the difference in pictures I just remove the WindowsDX monogame nuget package and add the DesktopGL or WindowsGL package.

No code changes between them.

KonajuGames:
There may be some platform-specific differences in the processing of the
PNG after it has been loaded. We need to make sure these are the same
across all platforms.

PR #5630, opened 2017-03-29
#5567: Replacing System.Drawing for DesktopGL

rds1983:


cra0zy:
MacOS/MONOMAC platform is not a DesktopGL platform, you partially modified it in here and that is why the build is failing.

rds1983:
Thanks for the help so far.
Is there a way to see the actual error text?
Not just "Process exited with code 1".

rds1983:
Nevermind, found error in the build log.

dellis1972:
@rds1983  I noticed the RGBToBGR function is gone. Did you make sure that it was not needed anymore? You need to test loading from .xnb and .png files and ensure transparent images are rendering correctly. 
The reason we had this is because OpenGL expects the image in BGR format (if I remember correctly).

rds1983:
ImageEx.cs was relying on System.Drawing. That why it had to go.
It was used only in the DesktopGL's version of Texture2D.FromStream. .xnb loading wasnt using it, so it shouldn't be affected.
I've attached a small sample that loads 2 textures using Texture2D.FromStream and then renders 2nd texture, which is transparent, on top of the first. It seems to work the same way it had before this PR(when it used Bitmap.FromStream for loading).

also the sample has commented section which could be used to test out SaveAsPng and SaveAsJpg.

[MonoGameStb.Test.zip](https://github.com/MonoGame/MonoGame/files/881775/MonoGameStb.Test.zip)

tomspilman:
First... thanks for all your work here @rds1983 ...  big issues like this only get solved with people like you contributing your time and efforts to MonoGame.

@KonajuGames @dellis1972 @rds1983 - So if this new image code is fully managed and works on all platforms...  is it a waste of effort to just add it to DesktopGL only?  When should we purge all the `PlatformFromStream` methods and move this code into the common `FromStream` method?

Second when should we purge the previous PNG loading/saving code?

  https://github.com/MonoGame/MonoGame/tree/develop/MonoGame.Framework/Utilities/Png

cra0zy:
> So if this new image code is fully managed and works on all platforms... is it a waste of effort to just add it to DesktopGL only? When should we purge all the PlatformFromStream methods and move this code into the common FromStream method?

IMO it's better to modify just DesktopGL in this PR so we can test it out a bit before we use it for the rest  of the platforms. It also makes it far easier to merge since it's modifying just one platform.

rds1983:
Yes, the new code works on all platforms and even on .NET Core. I think it is good idea to use it on more platforms as it'll provide consisency of Texture2D.FromStream result across different platforms.
Right now, the result is not consisent. I.e. WindowsDX premultiplies alpha, while DesktopGL doesnt.
I think there is even [an issue](https://github.com/MonoGame/MonoGame/issues/5529) about that.
On the other hand new image code is about 300kb. So it'll definitely make MonoGame size bigger. That could be an issue for the mobile platforms.

I think removing old Png code is good idea too. Instead following classes could be made public:
[ImageReaderFromStream](https://github.com/MonoGame/MonoGame/pull/5630/files#diff-ca91ee5e1600a7737bc6ffca45d62e83)
[ImageWriterToStream](https://github.com/MonoGame/MonoGame/pull/5630/files#diff-4e2ffaf2d7fc51115e741f1cc731bc27)
Which will allow users to read images in GIF(even animated)/PNG/JPG/BMP/TGA formats. And write to PNG/JPG/TGA and BMP formats.
Though class names could be reviewed. As well as their public interfaces.

dellis1972:
Size will only be an issue on Mobile if people use that code path.
Otherwise the linker will strip out the unused code.

tomspilman:
> IMO it's better to modify just DesktopGL in this PR so we can test it out a bit before we use it for the rest of the platforms. 

Yeah that makes sense to me.

My only concern is we would be merging this code blind... we have unit tests for `Texture2D.FromStream` as well as `SaveToPng`.  We just don't have the unit tests for DesktopGL hooked up to the build server yet.

So someone would have to manually run the DesktopGL tests to see if this at least passes what tests we have now.

rds1983:
I didnt change SaveToPng. It still uses Utilities/Png.
FromStream and SaveToJpg had been changed.

tomspilman:
> I didnt change SaveToPng. It still uses Utilities/Png.

I would switch that over too so we can test it all under DesktopGL?

@KonajuGames @dellis1972 -What do you guys think?

cra0zy:
> I would switch that over too so we can test it all under DesktopGL?

I agree.

rds1983:
> I would switch that over too so we can test it all under DesktopGL?
I have fixed it. If anyone has objections it would be very it roll that change back.

Also, I've reviewed entire changed code again and found another place that had been affected: MouseCursor.SDL.cs, method PlatformFromTexture2D.
It uses the new code to save the image to bmp. I hope there's a corresponding test.

rds1983:
>> I would switch that over too so we can test it all under DesktopGL?
> I have fixed it. If anyone has objections it would be very it roll that change back.

It wasn't that simple. As all platforms but android use Utilities/Png for SaveAsPng.
Means the new code imaging code needs to be added for all platforms (right now, it is added for WindowsGL,Linux,MacOS platforms). While it'll be used only for SaveAsJpg/SaveAsPng. 
Should I proceed?

cra0zy:
> It wasn't that simple. As all platforms but android use Utilities/Png for SaveAsPng.
> Means the new code imaging code needs to be added for all platforms (right now, it is added for WindowsGL,Linux,MacOS platforms). While it'll be used only for SaveAsJpg/SaveAsPng.
> Should I proceed?

I don't get what problem are you having?

Build/Projects/MonoGame.Framework.definition
```diff
@@ -1206,6 +1203,40 @@
     <Compile Include="Utilities\Png\PngReader.cs" />
     <Compile Include="Utilities\Png\PngWriter.cs" />
     <Compile Include="Utilities\ByteBufferPool.cs" />
+    <Compile Include="Utilities\Imaging\Image.cs">
+        <Platforms>WindowsGL,Linux,MacOS</Platforms>
```
cra0zy: MacOS shouldn't be here.

rds1983:
Nevermind. I've realized that the solution is to make the new code be called in SaveAsPng for WindowsGL,Linux,MacOS platforms and Utilities/Png for other platforms.

MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
```diff
@@ -495,13 +479,12 @@ private void PlatformSaveAsPng(Stream stream, int width, int height)
...
 #else
-            var pngWriter = new PngWriter();
-            pngWriter.Write(this, stream);
+            SaveAsImage(stream, width, height, ImageWriterType.Png);
 #endif
         }
 
 #if DESKTOPGL || MONOMAC
-        internal void SaveAsImage(Stream stream, int width, int height, ImageFormat format)
+        internal void SaveAsImage(Stream stream, int width, int height, ImageWriterType format)
```
cra0zy: You modified a piece of code that is used by both DESKTOPGL and MONOMAC, correct thing would be to add `#elif DESKTOPGL` and add code bellow.

rds1983:
> I don't get what problem are you having?
Nevermind, I've made it so new code is used solely for the DesktopGL(Texture2D.FromStream, Texture2D.SaveAsJpg and Texture2D.SaveAsPng). While other platforms keep their old code.
Thanks for your feedback.

rds1983:
btw, small offtop question: why doesnt have MacOS platform DESKTOPGL macro?

cra0zy:
> btw, small offtop question: why doesnt have MacOS platform DESKTOPGL macro?

MACOS platform is using Xamarin.Mac and native mac windowing and input, DESKTOPGL platform is using SDL. You can kind of compare that to WindowsDX and DesktopGL.

rds1983:
I see, thanks.

rds1983:
While commit is still there, I've commited some code. It's mostly refactoring and making generated code more cleaner and easier to read. But it is also some bug fixes.

tomspilman:
@KonajuGames @dellis1972 - We need to make a final pass on this and decide on merging it.

It would only affect GL for now, but ideally after some performance testing and running it thru our unit tests we can then use it across all platforms.

mrhelmut:
I gave this PR a test.

The good: it works and is faster than the original ```System.Drawing``` (and TONS faster than the Mono equivalent).

The bad: the alpha channel is wrong, I have all my alpha values having a ghostly color while it should be 100% transparent. I can't post any example right now and haven't looked into it yet (probably a channel swap).

tomspilman:
Thanks @mrhelmut .

If you get a chance to provide @rds1983 an image that loads incorrect i'm sure we can figure out this issue.

mrhelmut:
This should give wrong results when ```Texture2D.FromSteam()``` and rendering with ```SpriteBatch```: https://www.dropbox.com/s/8wgj3zs7fmc1un3/crop.png?dl=0

Maybe something to with premultiplied alpha?

rds1983:
Hi @mrhelmut.
It seems that image doesnt have the alpha channel at all:
crop_desc

cra0zy:
Looking at the image from Linux, it definitely has alpha.

tomspilman:
@rds1983 - Ywah @cra0zy is right... it is a 32bit image.  Did you maybe right click "Save Image" on the preview image instead of downloading the file from DropBox?

rds1983:
My bad, sorry. Now I downloaded it instead of just saving and alpha was there.

Anyway, the issue is related to the alpha-premultiplying when the texture is loaded from Texture2D.FromStream.
Different MG backends have different approaches there. I.e. WindowsDX backend does alpha-premultiplying. While DesktopGL doesnt. As this PR as aimed to replace DesktopGL, the alpha-premultiplying is not applied.
However you can still render the image correctly, using BlendState.NonPremultiplied.

mrhelmut:
I believe that DesktopGL premultiplies alpha as well, because it renders correctly with ```BlendState.AlphaBlend``` on DesktopGL.

tomspilman:
> Different MG backends have different approaches there.

Well the one I know is 100% correct is the Windows DirectX backend as I fixed it a few weeks ago.

 https://github.com/MonoGame/MonoGame/pull/5712

I recommend looking at that one and the docs on `Texture2D.FromStream` which I fixed.

It doesn't matter what any other backend does... Windows DirectX is correct and all others should be fixed to match it.

>  WindowsDX backend does alpha-premultiplying. 

It used to... which was wrong. It no longer does.

Correct XNA behavior is to not premultiply in `Texture2D.FromStream`.

However XNA does black out the color of any pixel with a zero in alpha.  You can see my implementation of that under DirectX here:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.DirectX.cs#L364

This behavior is why I think some people think it premultiplies...  it doesn't... it just does this.

rds1983:
Thanks for the detailed clarification.
I've copied code that blacks out zero-alpha pixels to the new image loading code(Texture2D.OpenGL.cs) and now it renders @mrhelmut's image correctly using default SpriteBatch(which uses BlendState.AlphaBlend).

rds1983:
A error(mac_build) I haven't seen before came out:
Utilities/ReflectionHelpers.Default.cs(19,35): error CS0308: The non-generic method 'Marshal.SizeOf(object)' cannot be used with type arguments

tomspilman:
@rds1983 - Not your fault.  Rebase or merge the latest changes from the develop branch.

tomspilman:
@rds1983 - Something weird is going on it shows a conflict now and 160 file changes.  Maybe it is how you rebased/merged?

rds1983:
I took latest develop, switched to sysdraw(this PR branch) and did 'git rebase develop', made sure it compiles and runs, then did 'git push origin sysdraw --force'. Something went wrong.

rds1983:
I forgot to sync mine develop with upstream's develop. It had been fixed now.

mrhelmut:
Looks good to me. 👍 
Begone System.Drawing!

tomspilman:
Great.... thanks @rds1983 and @mrhelmut !

Next step...

We should get the unit tests running under GL (https://github.com/MonoGame/MonoGame/pull/5587) so we can test these changes further.

Also replacing the other implementations with this one is fairly high priority.  My only concern is how fast this new implementation is vs native code.  I know on Switch i'm using a copy of STB for one project to avoid the current super slow C# png loader.  I hope to test this new loader on Switch to prove its performance.

PR #5631, opened 2017-03-29
Fix scroll wheel events on Windows Universal

jakepoz:
There was one place in the InputEvents.cs file where we should have signed up to get PointerWheelChanged events, but the call to do this was missing. This fixes that problem and allows scroll wheel events to work on Monogame for Windows Universal.


tomspilman:
Thanks @jakepoz !

PR #5635, opened 2017-04-02
[Linux Installer] Fix getting username that should run mdtool

### LINKED ISSUE #5634: [Linux Installer] Fix MonoDevelop addin installation
The installer is launching mdtool with root privileges which can cause problems if monodevelop was never opened beforehand (since it will create some files which require root access).

@tomspilman @KonajuGames @dellis1972 Add this to 3.6.1 release milestone, it's very important.

cra0zy:
Fixes #5634


tomspilman:
Thanks @cra0zy !

PR #5636, opened 2017-04-03
Fix the paths so they work on android-n+.

dellis1972:
Android N changed the way native libraries are used. You can
no longer directly load from /system/lib. Instead it seems we
need to just provide the library name and let the OS do its
thing.


tomspilman:
Merging

Issue #5047, opened 2016-07-29
EnableMouseTouchPoint - Doesn't handle loss of focus well


jamesford42:
Put a breakpoint somewhere in your code that triggers when a press occurs. Then click. You will now have a touch location stuck forever in 'moved' and future clicks also don't seem to register at all anymore.

Ziriax:
I'm also bitten by this, Monogame Windows DX platform.

zigzag312:
Should be fixed by #5518

Ziriax:
I'm not sure how XNA handled mouse events when the mouse button is down and the cursor moved out of the window, but in almost every desktop program, when the mouse button goes down, you captures the mouse, so that you keep receiving mouse events when the mouse moves outside the window. You also listen to the capturelost event to abort the drag.

tomspilman:
> when the mouse button goes down, you capture the mouse,

Yes... this is my understanding as well.  But we haven't really tested how this works under XNA.

Ziriax:
What is the general consensus from diverging from the XNA behavior? Monogame is an awesome engine, but isn't good old XNA holding it back? I mean the standard behavior should match XNA, but it would be nice to override that somehow. Of course one can always fork and tweak :)

tomspilman:
> What is the general consensus from diverging from the XNA behavior?

We stick to XNA behavior where it makes sense to do so.  If there is a better behavior we consider how to add it without breaking XNA games ported via MonoGame.

nkast:
Emulating Touch points and Gestures with mouse is a feature of Monogame but it should emulate TouchPanel as closely as possible without returning states that are unexpected/invalid for TouchPanel. 
https://github.com/MonoGame/MonoGame/pull/5518#issuecomment-289355427

nkast:
https://github.com/MonoGame/MonoGame/pull/5641 releases the TouchLocation when Mouse moves outside the Client or loses loses focus. 

I didn't implemented the second behavior,  that is to send a Pressed whenever mouse enters the Client and re-establish a TouchPoint, but at least this will fix the  duplicated Moved TouchLocations.

nkast:
https://github.com/MonoGame/MonoGame/pull/5643 is another solution that follows the proposal of @Ziriax. https://github.com/MonoGame/MonoGame/pull/5518#issuecomment-291661859

I think I can rewrite #5641 to the same pattern. The only change would be the condition to transition to  !MouseTouchLocation state. 
```
[...]

PR #5641, opened 2017-04-04
Fixed mouse touch event to release when mouse moves outside the client area or we loses focus

nkast:
#5047


tomspilman:
Thanks @nkast !

PR #5642, opened 2017-04-04
Allow loading from user's local fonts library

richkzad:
By default, fonts install to a user's local ~/Library/Fonts library. Now, the Content Pipeline will also check this directory when trying to process a font description.

It seems like it'd make sense to check this local directory after the game directory, but before the computer's global font library.

(It had taken me a while to figure out why I couldn't get my font to work.)

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
 
 #if WINDOWS || LINUX
 #if WINDOWS
-			fontName = FindFontFileFromFontName(fontName, fontDirectory);
+			fontName = FindFontFileFromFontName(fontName, FontsDirectory);
 #elif LINUX
             fontName = FindFontFileFromFontName(fontName, input.Style.ToString());
 #endif
--- a/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
@@ -246,11 +246,6 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
 			return glyphs.ToArray();
 		}
 
-		List<string> GetFontDirectories()
-		{
-
-		}
-
 #if WINDOWS
 		string FindFontFileFromFontName (string fontName, string fontDirectory)
 		{
--- a/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
@@ -27,6 +27,34 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
         [DefaultValue(typeof(TextureProcessorOutputFormat), "Compressed")]
         public virtual TextureProcessorOutputFormat TextureFormat { get; set; }
 
+		string FontsDirectory
+		{
+			get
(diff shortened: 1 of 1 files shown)


richkzad:
I guess it's probably better off for a project to include the source font file in the `Content` directory anyways. This way, anybody who downloads the project can rebuild the mgcb file right away.

But I think it'd be expected that MonoGame checks the local font library if it also checks the global one.

Jjagg:
Thanks for the contribution :)

I agree this would be nice to do. I do think, since this is Mac specific it would be nice if it worked like the Windows/Linux font folders. I.e. if there was an implementation of `FindFontFileFromFontName` for Mac like for [the other platforms](https://github.com/richkzad/MonoGame/blob/c1b44c8a1d59a07c64976ada88c43d67991583bf/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs#L249), then add an #if for Mac and call the function [here](https://github.com/richkzad/MonoGame/blob/c1b44c8a1d59a07c64976ada88c43d67991583bf/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs#L50).

richkzad:
Agreed -- I was surprised that the `/Library/Fonts` directory was always added to the list! I guess I falsely assumed MonoGame contributors figured it was fine.

Jjagg:
Oh, I didn't notice that. I don't think there's any reason for adding `/Library/Fonts` for any other platform than Mac. @cra0zy?

cra0zy:
> I.e. if there was an implementation of FindFontFileFromFontName for Mac like for the other platforms, then add an #if for Mac and call the function here.

Incorrect, those functions are trying to load font from system specific way, what was done here should be correct, see how the windows font folder gets added: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs#L60

#5528 does clean a lot of stuff up, but that isn't gonna get merged any time soon.

> Oh, I didn't notice that. I don't think there's any reason for adding /Library/Fonts for any other platform than Mac. @cra0zy?

The directory does not exist on Windows/Linux so even if it's added to the list it will do nothing... tho it probably should have had it's own `#ifdef` like Windows.

cra0zy:
No offence, but creating properties for just loading stuff in 1 function is a horrible solution... what was already there was good, it just needed an `#elif`

richkzad:
Like the commit said, I'm fresh to C#, so no offense taken!

However, I think it's worth moving the directory logic out of the `Process` method, since it makes the method easier to read. Would a method be better instead of a property?

Jjagg:
> Incorrect, those functions are trying to load font from system specific way

Oh, right. It just needs to check in the `~/Library/Fonts` folder, my bad.

cra0zy:
> However, I think it's worth moving the directory logic out of the Process method, since it makes the method easier to read.

That directory logic is actually very easy to read... it's just adding extra folders. The `#if` things do make code harder to read but once removed look at: https://github.com/cra0zy/MonoGame/blob/20eeeba115a905bcadaa8cbb857e3fcd93228770/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs#L46

Also you are creating a pointless (very very small) performance difference :|

richkzad:
You're right, it's nicer not to have those properties I added. What helped was at least bringing the font directory logic in one place.

richkzad:
Sigh. I realize the last commit I did was wrong, fixing it.

richkzad:
I can always go back to the original commit :) It was a simple change. Worst thing that would happen is it checks a folder and it fails. Wasting an extra directory check once for every spritefont in your package when building would have negligible performance implications?

richkzad:
Any update? I'm happy to make updates if necessary to get it through.

If you don't think it's necessary, I can just close it.

PR #5643, opened 2017-04-05
Process MouseTouchLocation state.

### LINKED ISSUE #5047: EnableMouseTouchPoint - Doesn't handle loss of focus well

nkast:
Ensure that TouchLocationState.Released will always follow a TouchLocationState.Pressed.
TouchLocationState.Moved will be send between.

fix #5047


tomspilman:
Fixed via #5641.

PR #5652, opened 2017-04-10
#3471: Exposing ConstantBuffer API

rds1983:
Following the discussion of the corresponding ticket #3471, the goal of this PR is to propose solution exactly as Tom suggested. Namely add this to ConstantBuffer(keeping everything else private/internal):
```c#
        public int Size
        {
            get { return _buffer.Length; }
        }

        public void SetParameter(int offset, byte[] data)
        {
            Buffer.BlockCopy(data, 0, _buffer, offset, data.Length);
        }
```

I think open Shader/ConstantBuffer API is awesome feature.


rds1983:
Found this PR #5354
And decide to close mine.

PR #5654, opened 2017-04-12
Implement GetHashCode on Vertex types

robert-wallis:
This may help someone's perf if they are using a vertices as a key in a dictionary, because lookup usually goes from "O(1)" to O(n) when keys collide.  I wasn't doing that, I just saw the TODO and thought it would be a quick fix.

### DIFF
--- a/MonoGame.Framework/Graphics/Vertices/VertexPositionNormalTexture.cs
+++ b/MonoGame.Framework/Graphics/Vertices/VertexPositionNormalTexture.cs
@@ -25,8 +25,7 @@ namespace Microsoft.Xna.Framework.Graphics
         }
         public override int GetHashCode()
         {
-            // TODO: FIc gethashcode
-            return 0;
+            return Position.GetHashCode() ^ Normal.GetHashCode() ^ TextureCoordinate.GetHashCode();
         }
 
         public override string ToString()
--- a/MonoGame.Framework/Graphics/Vertices/VertexPositionNormalTexture.cs
+++ b/MonoGame.Framework/Graphics/Vertices/VertexPositionNormalTexture.cs
@@ -26,7 +26,8 @@ namespace Microsoft.Xna.Framework.Graphics
 
         public override int GetHashCode()
         {
-            unchecked {
+            unchecked
+            {
                 var hashCode = Position.GetHashCode();
                 hashCode = (hashCode * 397) ^ Normal.GetHashCode();
                 hashCode = (hashCode * 397) ^ TextureCoordinate.GetHashCode();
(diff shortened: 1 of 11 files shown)


Halofreak1990:
Won't Xor'ing property values lead to unnecessary collisions for varying property values?

tomspilman:
@KonajuGames @dellis1972 @Jjagg @cra0zy  - I'm not good with hash code stuff... can you guys review this?

cra0zy:
> Won't Xor'ing property values lead to unnecessary collisions for varying property values?

Hashcodes are not unique, ie. 2 different object can have the same hashcode.

> @KonajuGames @dellis1972 @Jjagg @cra0zy - I'm not good with hash code stuff... can you guys review this?

I think this is fine.

Jjagg:
Yeah, just xoring won't get the best hashcodes because it's symmetric. I always let resharper generate hashcode implementations, those are good. If I remember correctly it does 
```
unchecked
{
    var hash = x.GetHashCode();
    hash = (hash * 397) ^ y.GetHashCode();
    hash = (hash * 397) ^ z.GetHashCode();
    return hash;
}
```
for a class with x, y and z properties that should affect the hashcode.

I should note that the Vector* structs have bad hashcode implementations, so that should also be fixed up if we want decent hashcodes.

cra0zy:
Multiplication by a prime number before XORing definitely helps with making sure that similar values do not get the same result.

robert-wallis:
Thanks @jjagg I updated this PR with Resharper's GetHashCode() generation.

Also I added some tests to `Vector*Test.cs` to make sure the hashes are working.  And found that the old Vector4 hash had a typo so it was missing Z.

I think the `Equals` functions need updating with the resharper generated code too.  But that should be in a different PR since those might be more controversial.

Jjagg:
Awesome! Thanks @robert-wallis :)
IMO this is good to be merged noe

MonoGame.Framework/Graphics/Vertices/VertexPositionNormalTexture.cs
```diff
@@ -23,10 +23,15 @@ VertexDeclaration IVertexType.VertexDeclaration
                 return VertexDeclaration;
             }
         }
+
         public override int GetHashCode()
         {
-            // TODO: FIc gethashcode
-            return 0;
+            unchecked {
```
cra0zy: MonoGame code is following Microsoft C# standard so you shouldn't use inline braces, instead move them to a new line:
```c#
unchecked
{
```

robert-wallis:
Thanks @cra0zy I updated the styling to use a bracket on a new line for the unchecked blocks.

Does anyone know the Resharper setting for Microsoft C# style?  Is it BSD style?

tomspilman:
Thanks @robert-wallis ... this is a great PR.

PR #5655, opened 2017-04-13
[ContentManager] Free scratchBuffer managed memory on Dispose()

nkast:


tomspilman:
This doesn't hurt anything and gives the GC a little help.

PR #5665, opened 2017-04-21
Remove GraphicsDevice.GetHighestSupportedGraphicsProfile()

nkast:
Platforms that use the `GraphicsDeviceManager.Legacy.cs` (GL?) will
default to Reach profile. Just like the DX platforms since 3.6.


tomspilman:
@nkast - I'm good with this change... XNA indeed defaults to Reach and doesn't guess for you.

I put up an issue for what is now missing...  picking the correct "device" based on the profile. See https://github.com/MonoGame/MonoGame/issues/5669 .

PR #5670, opened 2017-04-24
Implement GetHashCode and ToString methods for Joystick

cra0zy:
Extra thing I did was change the axis type from float to int since it should have been that in the first place.

### DIFF
--- a/MonoGame.Framework/Input/JoystickHat.cs
+++ b/MonoGame.Framework/Input/JoystickHat.cs
@@ -48,6 +48,36 @@ namespace Microsoft.Xna.Framework.Input
             get;
             internal set;
         }
+
+        /// <summary>
+        /// Serves as a hash function for a <see cref="T:Microsoft.Xna.Framework.Input.JoystickHat"/> object.
+        /// </summary>
+        /// <returns>A hash code for this instance that is suitable for use in hashing algorithms and data structures such as a
+        /// hash table.</returns>
+        public override int GetHashCode()
+        {
+            var hash = 0;
+
+            if (Left == ButtonState.Pressed)
+                hash |= (1 << 3);
+            if (Up == ButtonState.Pressed)
+                hash |= (1 << 2);
+            if (Right == ButtonState.Pressed)
+                hash |= (1 << 1);
+            if (Down == ButtonState.Pressed)
+                hash |= (1 << 0);
+
+            return hash;
+        }
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


MonoGame.Framework/Input/JoystickState.cs
```diff
@@ -19,7 +21,7 @@ public struct JoystickState
         /// Gets the joystick axis values.
         /// </summary>
         /// <value>An array list of floats that indicate axis values.</value>
```
tomspilman: Can you look over the docs here a little closer... seems like it still references the axis as float.

MonoGame.Framework/Input/JoystickHat.cs
```diff
@@ -48,6 +48,36 @@ public ButtonState Up
...
+            return hash;
+        }
+
+        /// <summary>
+        /// Returns a <see cref="T:System.String"/> that represents the current <see cref="T:Microsoft.Xna.Framework.Input.JoystickHat"/>.
+        /// </summary>
+        /// <returns>A <see cref="T:System.String"/> that represents the current <see cref="T:Microsoft.Xna.Framework.Input.JoystickHat"/>.</returns>
+        public override string ToString()
+        {
+            return "" + (int)Left + (int)Up + (int)Right + (int)Down;
```
tomspilman: Is the intent here to print something like `"PressedReleasedPressedPressed"` without any other formatting?

Maybe add to the docs the order of the results from the string?  Like "Returns the current state in Left, Up, Right, Down order".

cra0zy: > Is the intent here to print something like "PressedReleasedPressedPressed" without any other formatting?

The idea is to make ToString as short as possible and as easy to read(Pressed/Released would make text move while 0 and 1 keep the text stationary between value changes).

> Maybe add to the docs the order of the results from the string? Like "Returns the current state in Left, Up, Right, Down order".

Sure, that order is the default order Microsoft uses for Xaml stuff.

tomspilman: > make text move while 0 and 1 keep the text stationary between value changes

Oh... duh... I missed the `(int)`.

Anyway...  my point was there is no spacing or formatting between those values.  If that was intended then fine.

I would point out that maybe this should all be defined clearly in the docs.  Someone getting `0100` from a `ToString()` would be baffled by it and would need to read the code to figure it out.

cra0zy: What do you think on maybe making it return an angle?

cra0zy: Actually scratch that idea, angles would make things confusing since hats do not always get mapped properly.

cra0zy:
@tomspilman Fixed.

tomspilman:
Thanks @cra0zy !

PR #5671, opened 2017-04-25
Cleanup gamepad stuff

cra0zy:
Stuff done:
 - Added missing documentation
 - Fixed existing documentation
 - Fixed code formating
 - Merged variables with their properties where appropriate
 - Implemented missing ToString/GetHashcode methods
 - Fixed Equals method for GamePadState (it was using the default one, and for structs AFAIK it uses reflection which is slower than if you were to implement it manually)
 - Made some internal properties private and fixed up the tests that were using them... instead of the internal constructors....


tomspilman:
Thanks @cra0zy !

PR #5672, opened 2017-04-25
Pipeline process not terminating on exit fix

### LINKED ISSUE #5439: Pipeline tool process stays active after exiting
Hello,

When closing the Pipeline tool (at least on Windows), the window closes correctly but the process remains active and takes 100% of a CPU core usage.

zigzag312:
Fixes #5439

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.FileWatcher.cs
+++ b/Tools/Pipeline/Common/PipelineController.FileWatcher.cs
@@ -42,7 +42,8 @@ namespace MonoGame.Tools.Pipeline
 
             private void ExistsThread()
             {
-                while (true)
+                var loop = true;
+                while (loop)
                 {
                     // Can't lock without major code modifications
                     try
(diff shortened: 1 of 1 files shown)


Tools/Pipeline/Common/PipelineController.FileWatcher.cs
```diff
@@ -60,6 +61,10 @@ private void ExistsThread()
                             _view.Invoke(() => _view.UpdateTreeItem(item));
                         }
                     }
+                    catch (ThreadAbortException ex)
+                    {
+                        loop = false;
```
cra0zy: You should just call `return` instead of adding unnecessary variable.

zigzag312: I've simplified this as you suggested

tomspilman:
Thanks @zigzag312 !

PR #5673, opened 2017-04-25
Fixed Gamepad DPad on Android

cra0zy:
DPad was only implemented in case it was read as 4 buttons (which is rare), but was not implemented in case it was read as an axis.


endimilojkoski:
A while back I asked if someone can take a look at this on Android.
I have tried with Xbox 360, Xbox One, PS4 and iBuffalo gamepads on Android and DPad was not working on any of those and this is big issue for dpad only gamepads like iBuffalo.

Great job @cra0zy, can't wait to give it a try!

tomspilman:
Thanks @cra0zy !

Issue #5674, opened 2017-04-26
SamplerState.AnisotropicClamp throws 'SharpDX.SharpDXException 


nkast:
Setting SamplerState to AnisotropicClamp throws exception: 

```
SharpDX.SharpDXException was unhandled
  HResult=-2147024809
  Message=HRESULT: [0x80070057], Module: [General], ApiCode: [E_INVALIDARG/Invalid Arguments], Message: The parameter is incorrect.

  Source=SharpDX
  StackTrace:
       at SharpDX.Result.CheckError()
       at SharpDX.Direct3D11.Device.CreateSamplerState(SamplerStateDescription& samplerDescRef, SamplerState samplerStateOut)
       at SharpDX.Direct3D11.SamplerState..ctor(Device device, SamplerStateDescription description)
       at Microsoft.Xna.Framework.Graphics.SamplerState.GetState(GraphicsDevice device) in p:\MonoGame\MonoGame.Framework\Graphics\States\SamplerState.DirectX.cs:line 53
       at Microsoft.Xna.Framework.Graphics.SamplerStateCollection.PlatformSetSamplers(GraphicsDevice device) in p:\MonoGame\MonoGame.Framework\Graphics\SamplerStateCollection.DirectX.cs:line 54
       at Microsoft.Xna.Framework.Graphics.GraphicsDevice.PlatformApplyState(Boolean applyShaders) in p:\MonoGame\MonoGame.Framework\Graphics\GraphicsDevice.DirectX.cs:line 1298
       at Microsoft.Xna.Framework.Graphics.GraphicsDevice.ApplyState(Boolean applyShaders) in p:\MonoGame\MonoGame.Framework\Graphics\GraphicsDevice.cs:line 477
       at Microsoft.Xna.Framework.Graphics.GraphicsDevice.PlatformDrawUserPrimitives[T](PrimitiveType primitiveType, T[] vertexData, Int32 vertexOffset, VertexDeclaration vertexDeclaration, Int32 vertexCount) in p:\MonoGame\MonoGame.Framework\Graphics\GraphicsDevice.DirectX.cs:line 1408
       at Microsoft.Xna.Framework.Graphics.GraphicsDevice.DrawUserPrimitives[T](PrimitiveType primitiveType, T[] vertexData, Int32 vertexOffset, Int32 primitiveCount, VertexDeclaration vertexDeclaration) in p:\MonoGame\MonoGame.Framework\Graphics\GraphicsDevice.cs:line 1045
[...]

PR #5676, opened 2017-04-26
Use GraphicsCapabilities.MaxTextureAnisotropy on SamplerState

nkast:
-Move property MaxTextureAnisotropy from GraphicsCapabilities.OpenGL.cs
to GraphicsCapabilities.cs
-Enforce MaxAnisotropy limit when we create SamplerState [DX].
GL already does something similar:
https://github.com/MonoGame/MonoGame/blob/7c3d6870a38f8a5e479e64d935d692f2610e1cda/MonoGame.Framework/Graphics/States/SamplerState.OpenGL.cs#L72

Feature level 9.1 Max Anisotropy = 2.
Feature level 9.2 and up, Max Anisotropy = 16.

Fixes: #5674


tomspilman:
This makes sense... thanks @nkast .

It is also unfortunately hard to test without dedicated hardware.  We'll just have to live without that for now I guess.

PR #5678, opened 2017-04-29
Added Joystick.IsSupported property

cra0zy:


MonoGame.Framework/Input/Joystick.Default.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.Xna.Framework.Input
 {
     static partial class Joystick
     {
+        private static bool PlatformIsSupported = false;
```
tomspilman: @cra0zy - How about making these `const` or `readonly` so that it is clear this shouldn't be changed at runtime?

cra0zy: Done.

PR #5685, opened 2017-05-01
Improve GamePadThumbSticks code

cra0zy:
Stuff done:
 - Replaced constructor code with calling another constructor
 - Moved methods that were getting called just once to constructor
 - Moved constants to the top of the class

### DIFF
--- a/MonoGame.Framework/Input/GamePadThumbSticks.cs
+++ b/MonoGame.Framework/Input/GamePadThumbSticks.cs
@@ -9,7 +9,20 @@ namespace Microsoft.Xna.Framework.Input
     /// </summary>
     public struct GamePadThumbSticks
     {
-        private Vector2 _left, _right;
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Input/GamePadThumbSticks.cs
```diff
@@ -9,6 +9,20 @@ namespace Microsoft.Xna.Framework.Input
...
+#else
+        // Default & SDL Xbox 360 Controller dead zones
+        // Based on the XInput constants
+        private const float leftThumbDeadZone = 0.24f;
+        private const float rightThumbDeadZone = 0.265f;
+#endif
+
+        internal readonly Buttons VirtualButtons;
+
         private Vector2 _left, _right;
```
tomspilman: Could these now be `readonly` since they are set from the ctors?

cra0zy: Done.

tomspilman:
Thanks @cra0zy !

PR #5686, opened 2017-05-01
[Pipeline Tool] Use built-in Eto dialog stuff

cra0zy:
Stuff done:
 - Dialogs now use the integrated `Dialog<T>` method for returning result
 - Dialogs now autoposition buttons using Etos Positive / Negative buttons properties
 - Dialogs on Mac are now attached to the parent window
 - Removed some dialog per platform hacks


tomspilman:
Thanks @cra0zy !

PR #5689, opened 2017-05-02
Make SpriteBatch.End throw when Begin not called

### LINKED ISSUE #5688: SpriteBatch.End() doesn't check for Begin()
As per the documentation `SpriteBatch.End()` should throw an `InvalidOperationException` exception if `Begin()` has not been called.

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.spritebatch.end.aspx

Somehow this got lost in some refactor.

We need to add this check back into the codebase.

Also we need to add to our unit tests a check to ensure `SpriteBatch.Begin()` and `SpriteBatch.End()` throw the correct exceptions.

Jjagg:
fixes #5688 
@tomspilman


tomspilman:
Thanks @Jjagg !

PR #5691, opened 2017-05-02
[Android] Set GamePad.IsConnected to false if no gamepad is connected

### LINKED ISSUE #1170: GamePad.IsConnected is true on android but false on WP7
The following code returns false on a Windows Phone 7 device:

```
        if (GamePad.GetState(PlayerIndex.One).IsConnected)
```

The same code reports true on Android.

On WP7 all player indices report false. Yet, the back button press test works on WP7, just as it does on Android. I don't know about other targets yet.

cra0zy:
Mimicking what XNA does. Fixes #1170


tomspilman:
Thanks @cra0zy !

PR #5692, opened 2017-05-02
Added minimum size to the Pipeline Tool window

### LINKED ISSUE #5337: [Pipeline Tool] Negative width crashes property grid table
Dragging the vertical separator in the Pipeline Tool of the left side of the window will cause a crash due to `Drawable.Width` not accepting negative values. Easy fix is to use 0 if the value is negative, but maybe it makes more sense to have a minimum width (larger than 0). 
If you drag the seperator all the way to the right it can go beyond the visible window, making it impossible to make the build view visible again without resizing the window. 
@cra0zy

cra0zy:
Fixes #5337


tomspilman:
Thanks @cra0zy !

PR #5694, opened 2017-05-03
[Pipeline Tool] Optimize output filtering

harry-cpp:


tomspilman:
I'd like to get some manual testing on this before merging.

I'm actually very hesitant to merge any changes to the pipeline tool until we have some test framework in place for it.  It seems for every change we introduce new bugs that we don't find for months.

cra0zy:
> I'm actually very hesitant to merge any changes to the pipeline tool until we have some test framework in place for it.

I actually have an idea on what we could do for the testing, tho it will only work for Windows Pipeline Tool: http://docs.telerik.com/teststudio/testing-framework/write-tests-in-code/silverlight-wpf-automation-wtc/wpf-ui-automation

> It seems for every change we introduce new bugs that we don't find for months.

Its mostly the big Eto Forms switch that caused bugs.

tomspilman:
> I actually have an idea on what we could do for the testing

We can't have a test solution that only works on Windows.

The only choice I think we have is to roll our own solution that does specifically what we need.

> Its mostly the big Eto Forms switch that caused bugs.

I wish that was true.  Every time we merge something new into the Pipeline tool it seems to cause more bugs than it fixes.  Case in point https://github.com/MonoGame/MonoGame/issues/5697.

Until we resolve this testing issue we're not merging any new features into the Pipeline tool or changing anything.  I just don't trust any commit to the Pipeline tool anymore.

harry-cpp:
Gonna close this as I have no intention of fixing it up.

Also just gonna add that there is a major flaw with the pipeline tool where building using pipeline tool can take a lot longer than building from the command line, especially when building smaller assets as any output sent to the pipeline tool will pause MGCB until Pipeline Tool redraws stuff. (build taking less than 1s can take 10s+ because of this)

PR #5704, opened 2017-05-09
Implemented caching of staging resources used to copy data from a Texture2D under DirectX

ablanchard17:


tomspilman:
@ablanchard17 - Thanks for looking into this... it was needed for a while!

The reason some seeming random tests failed is because `Texture2D.GetData` is used as part of the unit tests and it is failing to return the correct data.  In fact it seems like it returns the data from the first GetData call on each subsequent call which is incorrect for `RenderTarget2D` which uses the same code path.

Maybe there is some flag or call to "invalidate" the data in the staging texture or release it before getting it again?

Also going to add a few notes on some other issues i noticed.

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -37,6 +37,8 @@ public partial class Texture2D : Texture
 
         private SampleDescription _sampleDescription;
 
+        private SharpDX.Direct3D11.Texture2D cachedStagingTexture;
```
tomspilman: Please use `_cachedStagingTexture`.

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -86,76 +88,79 @@ private void PlatformConstruct(int width, int height, bool mipmap, SurfaceFormat
...
+                desc.CpuAccessFlags = CpuAccessFlags.Read;
+                desc.SampleDescription.Count = 1;
+                desc.SampleDescription.Quality = 0;
+                desc.Usage = ResourceUsage.Staging;
+                desc.OptionFlags = ResourceOptionFlags.None;
+
+                // Save sampling description.
+                _sampleDescription = desc.SampleDescription;
+
+                cachedStagingTexture = new SharpDX.Direct3D11.Texture2D(GraphicsDevice._d3dDevice, desc);
```
tomspilman: The `cachedStagingTexture` needs to be disposed when the `Texture2D` is disposed.  You will need to override `Dispose` from `Texture`...

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture.DirectX.cs#L52

tomspilman:
@nkast @KonajuGames - Should we be concerned that keeping this staging texture uses more system memory for the lifetime of the texture?  If your game for some reason called `GetData` on a few hundred large textures it could really increase your memory usage?

Maybe for the next PR we can consider caching of these staging textures at the graphics device level and reuse them between textures?

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
ablanchard17: Thanks for the review. Yeah I was confused about the tests failing, as the staging texture was being created the same as when it wasn't cached. After looking at the tests I came to a similar conclusion, that the staging texture should be cleared out after the call to PlatformGetData

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -86,76 +88,79 @@ private void PlatformConstruct(int width, int height, bool mipmap, SurfaceFormat
...
+                var elementsInRow = rect.Width;
+                var rows = rect.Height;
+                var region = new ResourceRegion(rect.Left, rect.Top, 0, rect.Right, rect.Bottom, 1);
+                d3dContext.CopySubresourceRegion(GetTexture(), subresourceIndex, region, cachedStagingTexture, 0);
+
+                // Copy the data to the array.
+                DataStream stream = null;
+                try
+                {
+                    var databox = d3dContext.MapSubresource(cachedStagingTexture, 0, MapMode.Read, MapFlags.None, out stream);
```
tomspilman: @ablanchard17 - I wonder if this is the issue... don't we need to unmap this?  It didn't matter before because the staging texture was disposed after.  I bet this is the issue.

ablanchard17: I bet that's the issue. I looked into disposing the staging texture when the main texture is disposed. It looks like I will have to write a function to do this and then call it from Texture.DirectX.cs :: Dispose? or am I missing something

ablanchard17: Yeah, before when we were making a new Tex2D it wouldn't matter if Unmap was never called on it it looks like. But now that we are caching, we have to unmap properly or else shit will get wacky

tomspilman: You are missing something.

Just override `Dispose` in `Texture2D.DirectX.cs` , have it dispose the cached staging texture then let it call the base to deal with the rest.  Basically just copy the dispose from `Texture.DirectX.cs` over to this file and modify it.

ablanchard17: Sounds good, will do.

tomspilman:
Sweet... that looks to have fixed it @ablanchard17 !

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -86,80 +88,92 @@ private void PlatformConstruct(int width, int height, bool mipmap, SurfaceFormat
...
+                    SharpDX.Utilities.Dispose( ref stream);
+
+                    d3dContext.UnmapSubresource(_cachedStagingTexture, 0);
                 }
             }
         }
 
+        protected override void Dispose(bool disposing)
+        {
+            SharpDX.Utilities.Dispose(ref _cachedStagingTexture);
```
tomspilman: I think you need to put the `if (disposing)` here as well to implement the disposing pattern correctly.

@KonajuGames ?

KonajuGames:
Yes. Because the cached object is a managed object, it should only be
explicitly disposed if the disposing parameter is true, i.e. not called
from the finalizer.

nkast:
@tomspilman Yes, having a pool of some sort will keep the memory demand low but also improve the performance. If you call `GetData()` on a number of textures, each one will create a stage resource the first time.
Is it possible to Copy a region to a larger staging resource and then copy that in one go to a byte[]?
I mean without having to loop through every row.

@ablanchard17
I Believe most of the CPU cycles are speed of moving the data. Creating a GPU texture/buffer is not as slow as reserving Managed memory (GC pressure, initialize to Zeros, etc). 
Do you have comparison numbers between Develop and this PR?

ablanchard17:
I will set up and run some tests tonight.

ablanchard17:
I was also thinking about the proposed implementation of pooling these staging resources at the GraphicsDevice level. I'm not sure if that is the best way. It seems to me that in many use-cases, It would be unlikely that there would be a pre-existing staging texture that is an exact match for a given target. In which case the overhead of maintaining and checking the pool might go to waste.

Maybe we should think about instead adding some kind of "DoNotCacheStagingResource" flag that can be specified when a developer doesn't want to have a staging resource make and keep a cached staging resource for the rest of it's life after a call to a function that would require the use of one of these staging resources.

ablanchard17:
[TextureReadTests.pdf](https://github.com/MonoGame/MonoGame/files/998924/TextureReadTests.pdf)

@nkast Here are some tests I ran.

tomspilman:
@nkast @ablanchard17 - If either of you want to follow up with some sharing of staging resources across textures I think that is the next thing we could do to reduce the overhead of these things.

Merging.

PR #5705, opened 2017-05-09
Implemented cached staging buffer for D3D vertex buffer.

ablanchard17:
The creation of the staging buffer was moved into it's own function because it previously appeared twice in the file. This makes for DRYer code.


tomspilman:
Thanks @ablanchard17 !

Issue #2661, opened 2014-06-03
Implement XactImporter and XactProcessor


tomspilman:
We need to implement a real XactImporter and XactProcessor under the new MonoGame pipeline.

In Microsoft XNA these don't actually exist... under the hold these are piped to the command line XACT tools.

For MonoGame these should be a real importer/processor combo.  They should really parse XAP (XACT Audio Project) files and generate binaries for the platform.  This gives us the ability to replace the compressed audio with one optimized to the targeted platform.

The XAP format is easily readable in any text editor and should be easy to parse into a simple set of XapContent classes inside of the XactImporter.

The XactProcessor would convert the wave and sound data in the XapContent from uncompressed PCM to compressed format.

The XactWriter would write out the various binary files using our existing runtime Xact loading code as a guide to the format.

kvenkateshan:
Wondering how you guys are able to run xgs, xsb, and xwb successfully? I am able to initialize the audio engine successfully but the output is just noise. Interestingly enough, the timing of the pauses in the sound files played through exact is just fine.

flibitijibibo:
You can have a look at FNA's implementation, if you want:

https://github.com/flibitijibibo/FNA/tree/master/src/Audio

A lot of it's not tremendously different from MonoGame (there's only so much data out there), but maybe there's something I've forgotten.

kvenkateshan:
Discovered the issue. Had to remove using any compression presets on our wave banks. Just setting them to <none> makes everything sound just fine now.

PR #5709, opened 2017-05-13
RPC curves are now updated before Cue is played

tomspilman:
This PR fixes XAct to ensure RPC curve values are updated before a cue us played.

### DIFF
--- a/MonoGame.Framework/Audio/Xact/Cue.cs
+++ b/MonoGame.Framework/Audio/Xact/Cue.cs
@@ -135,7 +135,9 @@ namespace Microsoft.Xna.Framework.Audio
             var index = XactHelpers.Random.Next(_sounds.Length);
             _curSound = _sounds[index];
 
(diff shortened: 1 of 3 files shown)


KonajuGames:
Looks good. Merging.

DDReaper:
SO do we support Xact fully now through the content pipeline or is there still work to do?

tomspilman:
@DDReaper 

> SO do we support Xact fully now through the content pipeline or is there still work to do?

There is still work to do.

I outlined exactly what needs to happen a few years ago, but no one as taken the project on.  See https://github.com/MonoGame/MonoGame/issues/2661.

The flip side is FMOD works great with MonoGame and is a "professional" solution that works really well on desktops and consoles.

Someone has to jump in here if they want to keep XACT alive as a viable free alternative.

DDReaper:
Is the FMOD solution documented anywhere?  First I've heard/seen of it

PR #5712, opened 2017-05-16
Fixes to Texture2D.FromStream on Windows DirectX

tomspilman:
This PR fixes two bugs in Texture2D.FromStream() under Windows DirectX.

First is that it was premultiplying the color data on load which XNA does not do.

Second XNA blacks out the color data on any pixel with zero alpha.

I added two unit tests to verify this fix.

Fixes https://github.com/MonoGame/MonoGame/issues/5529.
Fixes https://github.com/MonoGame/MonoGame/issues/4820.


KonajuGames:
So it was really just one letter difference between premultiplied and non-premultiplied in one parameter to `fconv.Initialize()`.   Looks good to me.  Merging.

tomspilman:
Thanks @KonajuGames !

Yep....  That and the black out of pixels makes this behave the same as XNA.

PR #5713, opened 2017-05-16
Reusable function for raising events

gsfreema:
Before raising an event, you must first check the handler to see if it is null in order to check that an object is subscribed to handle the event.  This is best done if the event handler is stored in a local variable just in case the last subscriber unsubscribes in a different thread between the time the event handler is checked and the handler is actually invoked.  This reusable function makes it easy to raise events by storing a local copy of the handler and checking for null before invoking it.  Many classes already had a helper function for this.  This change refactors those into a single function.

There are some instances where more complex logic was being done after the event handler was checked.  In those scenarios, the code was mostly left alone except for the introduction of a local handler variable.


tomspilman:
@cra0zy @dellis1972 @KonajuGames - You guys good with this change?

Reduces the code size a little and maybe is a little safer.

KonajuGames:
It's still not safe though.  If the event was unsubscribed between the null check and the invoke, we shouldn't be calling the event handler anyway.  The intention of the unsubscribe was that the event handler should no longer be called, and bad things may happen if we do go ahead and call it.

I found this [article](https://blog.oz-code.com/net-events-thread-safe-and-reliability/) about it that suggests always adding a empty delegate.  The event handler will never be null, and therefore never need the null check.

cra0zy:
I agree with @KonajuGames, not really much to say.

tomspilman:
> I found this article about it that suggests always adding a empty delegate.

That was a pretty great article.  Yeah I agree the empty delegate design is best.  It is even less code than this PR and even safer.

@gsfreema - Do you mind redoing this PR with this technique instead?

gsfreema:
This is totally a valid concern that I thought might come up.  My line of thinking was that this change mostly followed the existing code, but it had the slight benefit of potentially calling an unsubscribed delegate with unknown side-effects vs. crashing with a NullReferenceException.  I guess the unknown side-effects could be quite concerning though.  Keyword "unknown" :)

I'm happy to rework the code to use an empty delegate.  I'm not sure if what the memory/performance costs of having that empty delegate are though.  It seems like raising an event would always try to call the empty delegate even if it is a no-op, and if I remember correctly, event handlers store a linked list of delegates.  That being said, I can get started with making changes.  The article mentions little overhead, though that is often relative.

tomspilman:
> The article mentions little overhead, though that is often relative.

Exactly.  None of these delegates are called more than maybe once or twice in a frame.  I could understand a performance concern if the delegate was called 100s of times per frame, but not in any of these cases.

Just safer, simpler, and cleaner code with empty delegates.

gsfreema:
After tracking down many events in the MonoGame framework, I really started questioning if this empty delegate pattern is truly offering the thread safety we are expecting.  In reality, it suffers from the same race condition concerns of invoking a stale event.  This can happen because the delegate can start its invoke call, then gets halted by another thread that removes its handler, then the delegate continues its invoke call.  Delegates are immutable so if it starts its invoke call, it will continue with the same invocation list.  This problem would show itself even more if there were many event handlers.  By the time it got to the latter handlers, who knows what other threads have done by then.  Here is a good article written by someone who used to work on the C# compiler and language design team.

https://blogs.msdn.microsoft.com/ericlippert/2009/04/29/events-and-races/

He suggests that there are two separate race conditions.  One is the null check, and the other is stale event handlers.  In summary, he suggests that it's the responsibility of the code that raises the event to not throw a NullReferenceException, and it's the responsibility of the handler to gracefully detect if the handler is stale and bail out.

Let me know what you all think.  I'm leaning more towards the temporary variable check again.  I have also come to the realization that this PR is missing a lot of EventHandlers so I could add to this PR or create a new one if you all would like if you decide this is the way to go.

KonajuGames:
That article explains it a lot better, and there is no effective difference between the temporary variable and the empty delegate.  They both solve the first problem, and the second problem is up to the event handler to deal with.  So I'm happy with the approach in this PR, amnd I'm good with you adding more of the same event handling changes to this PR.  I would also like some code documentation added to EventHelper.Raise explaining the purpose of it and why it is used.

tomspilman:
@gsfreema - I'm good with that too.  Go for it.

zigzag312:
Interesting discussion! So adding an empty delegate doesn't solve second type of race condition either (possibility of event being called after being unsubscribed from another thread). 

I have to ask what is the reason we are limited to old C# version? C# 6.0 null propagation (MyEvent?.Invoke(this, args)) does basically the same thing as that helper function.

gsfreema:
@zigzag312 That is a good question that I don't know the answer to.  I recently asked which version of C# was being used to develop MonoGame on the forums the other day so I could determine if the null-conditional operator could be used.  The reasons for still using C#5 that I could come up with are that you either don't want to require contributors to update to VS2015 (which released the Roslyn and C#6 compiler) or that the machines building MonoGame would need to be updated and can't for some reason.

KonajuGames:
The particular version of C# usually comes down to the highest common version that all supported platforms can use.  At this stage I believe PS4 may be the lowest version, depending on which particular version of Mono it now uses.

There is also no rush to use all new features, especially syntactic features as these often generate code behind the scenes, and quite often this generated code can allocate memory that we could otherwise avoid.  The null coalescing operator may well end up being converted to a simple `if` statement by the compiler.  I haven't seen any code using the operator to determine the case yet.

nkast:
I have allready used this patern on some places, for example here:
https://github.com/MonoGame/MonoGame/pull/5520/commits/c36898cce44bc20606dddbf06196cca210e9ccbf#diff-c36ddd5c033ea9ae2a05b30e38dfc0a6

I think the method already makes a copy of the Handler since you pass it as a parameter but I'am not sure what happens if the method gets optimized/inlined by the compiler. I guess it's best if we make a copy of the handler anyway.

I don't think the empty delegate trick will make the code cleaner, for almost every `if(handler == null)` we have to assign a delegate in the constructor.

MonoGame.Framework/DrawableGameComponent.cs
```diff
@@ -25,8 +25,7 @@ public int DrawOrder
                 if (_drawOrder != value)
                 {
                     _drawOrder = value;
-                    if (DrawOrderChanged != null)
-                        DrawOrderChanged(this, null);
+                    EventHelper.Raise(this, DrawOrderChanged, null);
                     OnDrawOrderChanged(this, null);
```
nkast: Here, instead of `EventHelper.Raise(this, DrawOrderChanged, null);` we should raise the event from within `OnDrawOrderChanged(this, null);`.

The dev should be able to override `OnDrawOrderChanged()` instead of listening to `DrawOrderChanged` and also discard the event from rising  if he chooses so (by not calling `base.OnDrawOrderChanged()`). 

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.drawablegamecomponent.ondraworderchanged.aspx

Same goes for `OnVisibleChanged()`

gsfreema: Good call.  The docs clearly state that OnDrawOrderChanged raises the event.  I'll try to find more of these instances as I do my more thorough pass through to make sure events are raised using this pattern.  It might be a little while before I get around to it, but I'll keep this in mind.

MonoGame.Framework/DrawableGameComponent.cs
```diff
@@ -70,8 +66,14 @@ public override void Initialize()
...
+        protected virtual void OnVisibleChanged(object sender, EventArgs args)
+        {
+            EventHelpers.Raise(this, VisibleChanged, args);
+        }
 
-        protected virtual void OnDrawOrderChanged(object sender, EventArgs args) { }
+        protected virtual void OnDrawOrderChanged(object sender, EventArgs args)
+        {
+            EventHelpers.Raise(this, DrawOrderChanged, args);
+        }
```
tomspilman: This and the change to `OnVisibleChanged` is a good bug fix too.  As the docs state:

> Called when the DrawOrder property changes. Raises the DrawOrderChanged event. 
https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.drawablegamecomponent.ondraworderchanged.aspx

> Called when the Visible property changes. Raises the VisibleChanged event. 
https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.drawablegamecomponent.onvisiblechanged.aspx

We were incorrectly not calling the event from the virtual before.

tomspilman:
Thanks @gsfreema !

PR #5718, opened 2017-05-19
Support DistanceScale and DopplerFactor under OpenAL

### LINKED ISSUE #5717: SoundEffect.DistanceScale does not work under OpenAL
The `SoundEffect.DistanceScale` global is correctly set under XAudio platforms:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs#L41

But it is totally unused under OpenAL:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstance.OpenAL.cs#L72

This breaks games causing sounds to be too loud or totally silent.

Jjagg:
Fixes #5717 

I haven't tested this yet.

### DIFF
--- a/MonoGame.Framework/Audio/OpenAL.cs
+++ b/MonoGame.Framework/Audio/OpenAL.cs
@@ -62,6 +62,7 @@ namespace OpenAL
     {
         Pitch = 0x1003,
         Gain = 0x100A,
+        ReferenceDistance = 0x1020
     }
 
     public enum ALGetSourcei
@@ -354,7 +355,7 @@ namespace OpenAL
         public static extern void GetSource (int sourceId, ALGetSourcei i, out int state);
 
         public static ALSourceState GetSourceState(int sourceId) {
-            int state = (int)ALSourceState.Stopped;
+            int state;
             GetSource (sourceId, ALGetSourcei.SourceState, out state);
             return (ALSourceState)state;
         }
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks @Jjagg !

mrhelmut:
Good to go from my perspective! Thanks @Jjagg 

Looking at the X3DAudio documentation, should ```SoundEffect.DistanceScale``` and ```SoundEffect.DopplerScale``` be clamped? According to the doc, they should have a max value of ```FLT_MAX```, which isn't checked.

AL seems to clamp values itself, but maybe X3DAudio does the same?

Jjagg:
Isn't FLT_MAX just float.MaxValue?

KonajuGames:
> According to the doc, they should have a max value of FLT_MAX, which isn't checked.

FLT_MAX is the maximum value a float can have, so you can't clamp to it.

mrhelmut:
My bad. ^^ Everything's fine!

Jjagg:
I just noticed `AudioEmitter` has its own `DopplerScale`. The XAudio path doesn't use it and just sets the XAudio emitter Doppler scale to `SoundEffect.DopplerScale` (see https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffectInstance.XAudio.cs#L42). We probably need to multiply `SoundEffect.DopplerScale` with `AudioEmitter.DopplerScale` to get the same effect as in XNA, I should test that. I'm not 100% sure we can achieve this effect in OpenAL because there is only a global Doppler scale and I'm not sure if changing it affects sounds that are already playing too. If it does we can't have emitter specific Doppler scaling.

Jjagg:
In OpenAL we don't have a sourceId until we play the sound. 
In `SetIsLooped`, `SetPan`, `SetPitch` etc. we check if there is a SourceId and ignore the call if there isn't. ApplyFilter and the current implementation of Apply3D use the SourceId, but don't check if it has actually been assigned, which can cause OpenAL to throw.
Checking if we have a SourceId (only assigned when playing a sfx) is not the right solution though. The following will work with XAudio but will be ignored in OpenAL:
```
var instance = sfx.CreateInstance();
instance.SetPitch(pitch);
instance.Play();
```
Switching the second and third line will work in OpenAL, though I find first configuring the sfx and then playing it way more intuitive.
I don't know why instances don't immediately get a source id, but seems like the easy solution to just fetch one on creation of an instance. Are there any downsides to that? If we can't do that we can store the values for all Set... and Apply... calls and only actually apply them when playing a sound.

tomspilman:
> I just noticed `AudioEmitter` has its own `DopplerScale`.

Interesting... I totally missed that too.  Well for XAudio it should be fixable.  I guess it is a multiply... we would have to test and see.

> I'm not 100% sure we can achieve this effect in OpenAL

I would be surprised if the API couldn't do it...  but I guess we would have to live with that.

Jjagg:
I'll open issues for my last two comments. I'll fix those in separate PRs. I tested the changes in this PR and by ear it sounds the same as DX. So IMO this is good to be merged.

PR #5719, opened 2017-05-19
Set AllowUserResizing as virtual

cra0zy:
I was gonna submit some new features to the `GameWindow`, but then I noticed how broken it is... there are a lot of things I will need to improve before I can implement anything properly. The last PR I submit regarding the `GameWindow` stuff will contain documentation and `Obsolete` attributes (and probably some code formatting fixes).

### DIFF
--- a/MonoGame.Framework/Android/AndroidGameWindow.cs
+++ b/MonoGame.Framework/Android/AndroidGameWindow.cs
@@ -181,18 +181,6 @@ namespace Microsoft.Xna.Framework
             }
         }
 
-        public override bool AllowUserResizing 
-        {
-            get 
-            {
-                return false;
-            }
-            set 
-            {
-                // Do nothing; Ignore rather than raising an exception
-            }
-        }
-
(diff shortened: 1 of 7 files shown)


Jjagg:
> I was gonna submit some new features to the GameWindow, but then I noticed how broken it is... there are a lot of things I will need to improve before I can implement anything properly.

Maybe you can take some of the SDL windowing stuff from #5522.

nkast:
I thought this PR makes sense. On platforms that doesn't allow full-screen it doesn't hurt to let the value be either true or false. But MSDN says that "On the Xbox 360, this property is always false." so I guess the current implementation is on a par with XNA.

By the way, nowadays Windows Store Apps let the user resize the window. Disabling user resizing is not implemented and I don't know if it's possible to do so in Store Apps, so maybe `AllowUserResizing` should return `true`?

cra0zy:
I was gonna make a new PR and combine several smaller ones into it.

PR #5720, opened 2017-05-19
Remove redundant _allowAltF4 variable

cra0zy:


MonoGame.Framework/GameWindow.cs
```diff
@@ -16,12 +16,10 @@ public abstract class GameWindow {
 
 		public abstract Rectangle ClientBounds { get; }
 
-	    internal bool _allowAltF4 = true;
-
         /// <summary>
         /// Gets or sets a bool that enables usage of Alt+F4 for window closing on desktop platforms. Value is true by default.
         /// </summary>
-        public virtual bool AllowAltF4 { get { return _allowAltF4; } set { _allowAltF4 = value; } }
+        public virtual bool AllowAltF4 { get; set; }
```
tomspilman: So this defaults to false now?

cra0zy: Yes.

KonajuGames: It wasn't redundant because it allowed setting the default value to `true`.

cra0zy: You can set default property values, I accidentally forgot to do it.

KonajuGames: For automatic properties, that's a recent C# feature.  If it's to move the setting of the default value from a field to the constructor, what benefit does that give us?

cra0zy: One less variable to think about, `_allowAltF4` is really only used for the property and nothing else.

nkast: We can rename it to DisableAltF4 and don't worry about setting it on every constractor while also keep it at minimal code as in this PR without the extra field.

nkast: Although it would be nice to keep AllowAltF4 for at least a couple of releases.

`[Obsolete("Use DisallowAltF4")] public virtual bool AllowAltF4 { get { return !DisallowAltF4; } set { DisallowAltF4 = !value; } }`
+

PR #5723, opened 2017-05-19
Remove SharpDX.RawInput.dll reference from DirectX graphics backend

nkast:
SharpDX.RawInput is not used but referencing it can cause issues like #2110.


DDReaper:
If we are going to merge this one, I believe should we then also remove the RawInput references from the Project templates and NuGet's

nkast:
@DDReaper I fixed the conflict in `.nuspec`. It should be good to merge.

tomspilman:
Thanks @nkast !

PR #5727, opened 2017-05-22
Use Mono properties for pipeline platform prefix

darfink:
With the current `MonoGame.Content.Builder.targets` implementation, the build breaks as soon as any of the platform specific resource prefixes are modified:

- `IPhoneResourcePrefix`
- `MonoAndroidAssetsPrefix`
- `MonoMacResourcePrefix`

This is because the directory (`PlatformResourcePrefix`) is hardcoded.

It is also not customisable from the project that imports it, since the conditional logic always sets the property on the affected platforms (iOS, Android, macOS).

I do not believe a fallback value is required since the variables are set by default on each platform.
See [Xamarin.Mac.Common.props](https://github.com/xamarin/xamarin-macios/blob/master/msbuild/Xamarin.Mac.Tasks/Xamarin.Mac.Common.props) and cousins, but my assumption may be wrong.

In case a dependency on these properties are preferred to be avoided, a customisable user property is the least that should be provided.


tomspilman:
@dellis1972 @DDReaper - This make sense to you guys?

DDReaper:
Not sure I can comment on this one, seems to make sense.  @dellis1972 @CartBlanche as Xamarin guys, do you have a view?

tomspilman:
Thanks @darfink !

PR #5728, opened 2017-05-22
Fixed leaks that affected shutting down and recreating GraphicsDevice under DirectX

KonajuGames:
Running tests has dropped from a continuous climb to 5GB memory to a more consistent use that peaks at 195MB.

- The swapchain was not being disposed in GraphicsDevice.DirectX for Windows.
- Added PlatformDispose() to BlendState.
- SpriteBatchTest was not calling base.TearDown().
- GraphicsDeviceTestFixtureBase calls game.DoInitialize() rather than CreateDevice() directly on the IGraphicsDeviceManager. This allows the call to game.Dispose() to properly clean up the GraphicsDeviceManager.
- VertexBuffer ctor was assigning the GraphicsDevice to the VertexDeclaration, but this was not used anywhere. Most common VertexDeclarations are a static global resource, thus it was holding a reference to the first GraphicsDevice.

### DIFF
--- a/MonoGame.Framework/Graphics/States/BlendState.cs
+++ b/MonoGame.Framework/Graphics/States/BlendState.cs
@@ -243,15 +243,16 @@ namespace Microsoft.Xna.Framework.Graphics
 	        return new BlendState(this);
 	    }
 
+        partial void PlatformDispose();
+
         protected override void Dispose(bool disposing)
         {
             if (!IsDisposed)
             {
                 for (int i = 0; i < _targetBlendState.Length; ++i)
                     _targetBlendState[i] = null;
-#if DIRECTX
+
                 PlatformDispose();
-#endif
             }
             base.Dispose(disposing);
         }
--- a/MonoGame.Framework/Graphics/States/BlendState.cs
+++ b/MonoGame.Framework/Graphics/States/BlendState.cs
@@ -242,6 +242,19 @@ namespace Microsoft.Xna.Framework.Graphics
 	    {
 	        return new BlendState(this);
 	    }
-	}
+
+        protected override void Dispose(bool disposing)
+        {
+            if (!IsDisposed)
+            {
+                for (int i = 0; i < _targetBlendState.Length; ++i)
+                    _targetBlendState[i] = null;
+#if DIRECTX
+                PlatformDispose();
+#endif
+            }
+            base.Dispose(disposing);
+        }
+    }
 }
 
--- a/MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
+++ b/MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
@@ -26,6 +26,10 @@ namespace Microsoft.Xna.Framework.Graphics
             this.VertexCount = vertexCount;
             this.BufferUsage = bufferUsage;
 
+            // Make sure the graphics device is assigned in the vertex declaration.
+            if (vertexDeclaration.GraphicsDevice != graphicsDevice)
+                vertexDeclaration.GraphicsDevice = graphicsDevice;
+
             _isDynamic = dynamic;
 
             PlatformConstruct();
--- a/MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
+++ b/MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
@@ -26,10 +26,6 @@ namespace Microsoft.Xna.Framework.Graphics
             this.VertexCount = vertexCount;
             this.BufferUsage = bufferUsage;
 
-            // Make sure the graphics device is assigned in the vertex declaration.
-            if (vertexDeclaration.GraphicsDevice != graphicsDevice)
-                vertexDeclaration.GraphicsDevice = graphicsDevice;
-
             _isDynamic = dynamic;
 
             PlatformConstruct();
--- a/Test/MonoGame.Tests.XNA.csproj
+++ b/Test/MonoGame.Tests.XNA.csproj
@@ -235,10 +235,16 @@
       <Link>Assets\Effects\Stock\Structures.fxh</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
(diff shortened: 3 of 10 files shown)


MonoGame.Framework/Graphics/States/BlendState.cs
```diff
@@ -241,7 +241,20 @@ static BlendState()
...
+	    }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (!IsDisposed)
+            {
+                for (int i = 0; i < _targetBlendState.Length; ++i)
+                    _targetBlendState[i] = null;
+#if DIRECTX
+                PlatformDispose();
```
tomspilman: Maybe you can declare this as a partial function... so that if it doesn't exist then it is ok?  Better than an `#if` I think.

MonoGame.Framework/Graphics/Vertices/VertexBuffer.cs
```diff
@@ -26,10 +26,6 @@ protected VertexBuffer(GraphicsDevice graphicsDevice, VertexDeclaration vertexDe
             this.VertexCount = vertexCount;
             this.BufferUsage = bufferUsage;
 
-            // Make sure the graphics device is assigned in the vertex declaration.
-            if (vertexDeclaration.GraphicsDevice != graphicsDevice)
-                vertexDeclaration.GraphicsDevice = graphicsDevice;
```
tomspilman: We could live with this maybe...  although I expect XNA does set it.

I wonder how many people have as a hack assumed they could fetch the graphics device via `VertexDeclaration.GraphicsDevice`.

tomspilman: XNA even uses these as a `static`...

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.graphics.vertexpositioncolor.vertexdeclaration.aspx

So not sure hope they ensure it is cleaned up and reused when a GraphicsDevice is disposed and recreated.

KonajuGames: Did some tests with XNA.  The `VertexDeclaration.GraphicsDevice` property remains `null` until a `VertexBuffer` is created using the vertex declaration.  After the `GraphicsDevice` is disposed, the vertex declaration retains the reference to the disposed `GraphicsDevice` object.  The disposed `GraphicsDevice` object is replaced with the new `GraphicsDevice` object when subsequently used again with a new `VertexBuffer`.

tomspilman: @KonajuGames - So that sounds exactly how we had things working here right?

Really the worst thing this could do would be keeping the old disposed `GraphicsDevice` class in memory.  Since it was disposed it should have already released all native resources so this reference shouldn't keep anything big in memory.

I think these two lines can be restored without any leaks appearing as long as we're disposing everything else correctly.

KonajuGames: Those lines were restored in the last commit. Now that GraphicsDevice properly disposes of everything, it is not an issue.

nkast:
Lately I had problem to run the XNA tests, probably because it run out of memory.
I tested this PR with the XNA tests but unfortunately nothing has changed.
Memory keeps growing and after a while the tests starts failing, at the end the app crash.

screenshot 32

Just to make it clear, This PR didn't make XNA tests worst, the bug/leak is already in Develop, but since it fixes a couple of leaks in SpriteBatchTest & GraphicsDeviceTestFixtureBase it's the reason I tests it against the XNA tests project.

KonajuGames:
The major fix here is in our GraphicsDevice.DirectX where it now disposes the swapchain.  This will have no effect on the XNA tests.

I have had the XNA tests fails because it has run out of resources as well.  Haven't tracked that one down yet.

Test/MonoGame.Tests.XNA.csproj
```diff
@@ -861,7 +927,9 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
tomspilman: Do you happen to know what this is?

Test/MonoGame.Tests.XNA.csproj
```diff
@@ -827,30 +872,51 @@
...
     </None>
+    <None Include="Assets\Fonts\QuartzMS.spritefont">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
     <None Include="Assets\Fonts\QuartzMS.xnb">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="Assets\Fonts\SegoeKeycaps.spritefont">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
tomspilman: I double checked.... this file isn't used at runtime in the tests which is why it wasn't included in the project before.  Do we want to start including everything now?  This will mean the XNA project and the MonoGame Test project won't match.

While some of these are omissions in the XNA project... I suspect some are like this and not used at runtime.

tomspilman:

MonoGame.Framework/Graphics/States/BlendState.cs
```diff
@@ -241,7 +241,21 @@ static BlendState()
...
+            if (!IsDisposed)
+            {
+                for (int i = 0; i < _targetBlendState.Length; ++i)
+                    _targetBlendState[i] = null;
+
+                PlatformDispose();
+            }
+            base.Dispose(disposing);
+        }
+    }
```
tomspilman: @KonajuGames - Could you apply this same `PlatformDispose` technique to all the other render states?  Just a little weird that the others just override `Dispose(bool)` now.  Ran into this on console.

KonajuGames: Sure.

PR #5732, opened 2017-05-25
Updated reflection to be backward compatible with .Net 4.0

### LINKED ISSUE #5716: Type.GetTypeInfo() is Net 4.5+
I just noticed that ```Type.GetTypeInfo()``` is an extension method only present since .Net 4.5.

This change breaks the compatibility with .Net 4.0: https://github.com/MonoGame/MonoGame/commit/e3261f71c780c7b8294a42a9e3f36efd447f439d#diff-51a9663b2b82c89eb208960d5902e904R38

Having a .Net 4.0 compatibility is interesting for Windows XP support.

According to the Steam hardware survey, Windows XP still represents about 1% of players. http://store.steampowered.com/hwsurvey/

Is this something we should care about or should we move forward?

I would be fine with moving forward, since Microsoft doesn't support the OS anymore and that the OpenGL 3.0 requirement for DesktopGL represents a way bigger barrier (most XP machines don't even support it anyway).

@tomspilman @KonajuGames @cra0zy

mrhelmut:
This is a follow up / rework of #5607 (cc @cra0zy for being the initial author) by using a custom protobuild transform to add conditional constant for detecting .Net 4.0 (it defines ```NET4``` if the target framework version is "v4.0").

Fixes #5716

@tomspilman @KonajuGames

### DIFF
--- a/Build/AdditionalProjectTransforms.xslt
+++ b/Build/AdditionalProjectTransforms.xslt
@@ -1,5 +1,5 @@
 <xsl:if test="/Input/Properties/DefineNET4DetectionConstant = 'True'">
   <PropertyGroup>
-    <DefineConstants Condition="$([System.Version]::Parse('$(TargetFrameworkVersion.Substring(1))').CompareTo($([System.Version]::Parse('4.0'))))   &gt;= 0">$(DefineConstants);NET4</DefineConstants>
+    <DefineConstants Condition=" '$(TargetFrameworkVersion)' == 'v4.0' ">NET4</DefineConstants>
   </PropertyGroup>
 </xsl:if>
--- a/Build/AdditionalProjectTransforms.xslt
+++ b/Build/AdditionalProjectTransforms.xslt
@@ -1,5 +1,5 @@
 <xsl:if test="/Input/Properties/DefineNET4DetectionConstant = 'True'">
   <PropertyGroup>
-    <DefineConstants Condition="$([System.Version]::Parse('$(TargetFrameworkVersion.Substring(1))').CompareTo($([System.Version]::Parse('4.5'))))   &gt;= 0">$(DefineConstants);NET45</DefineConstants>
+    <DefineConstants Condition=" '$(TargetFrameworkVersion)' != 'v4.0' And '$(TargetFrameworkVersion)' != 'v3.5' And '$(TargetFrameworkVersion)' != 'v3.0' And '$(TargetFrameworkVersion)' != 'v2.0' ">$(DefineConstants);NET45</DefineConstants>
   </PropertyGroup>
 </xsl:if>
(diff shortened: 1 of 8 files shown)


Build/AdditionalProjectTransforms.xslt
```diff
@@ -0,0 +1,5 @@
+<xsl:if test="/Input/Properties/DefineNET4DetectionConstant = 'True'">
+  <PropertyGroup>
+    <DefineConstants Condition=" '$(TargetFrameworkVersion)' == 'v4.0' ">NET4</DefineConstants>
```
tomspilman: What about below 4.0? ... should this be `<=` here?  Or would this be better as `>= 'v4.5'` and `NET45` as the define?

mrhelmut: Yes, sounds better. Inverting this and detecting 4.5+ instead should be a better strategy.
I'll update the PR tomorrow.

mrhelmut: I've updated the PR but it fails loading the project on Xamarin platforms. It looks like it cannot calls ```System.Version.Parse()``` and/or ```System.Version.CompareTo()``` within the csproj.
I believe it has something to do with XBuild not supporting this (or did we switch to msbuild and it's something else?).

mrhelmut:
@tomspilman I've managed to get a NET45+ condition working on both Visual Studio and Xamarin. Not quite generic but it works.

mrhelmut:
@tomspilman hope you don't mind a ping. :)

Build/AdditionalProjectTransforms.xslt
hach-que: Not that I care too much, but surely the number of platforms that are built on .NET 4.0 is pretty limited now... is there a reason to define `NET45` with this conditional magic instead of just appending it to the platforms that are .NET 4.5 and higher?

mrhelmut: Mostly for backward compatibility with Windows XP. It's no big deal (~2% of our players are still on XP), but this issue/fix has a limited scope, so I believe it doesn't harm to have that in.

Build/AdditionalProjectTransforms.xslt
```diff
@@ -0,0 +1,5 @@
+<xsl:if test="/Input/Properties/DefineNET4DetectionConstant = 'True'">
+  <PropertyGroup>
+    <DefineConstants Condition=" '$(TargetFrameworkVersion)' != 'v4.0' And '$(TargetFrameworkVersion)' != 'v3.5' And '$(TargetFrameworkVersion)' != 'v3.0' And '$(TargetFrameworkVersion)' != 'v2.0' ">$(DefineConstants);NET45</DefineConstants>
+  </PropertyGroup>
```
tomspilman: I guess there is no better way to test this other than explicit versions.  @dellis1972 @KonajuGames ?

mrhelmut: As far as I have looked into this, if you want a compatibility with XBuild, no.

XBuild can't use semantics like ```$([System.Version]::Parse('$(TargetFrameworkVersion.Substring(1))').CompareTo($([System.Version]::Parse('4.5'))))``` which would have been totally generic.

But since MonoGame doesn't build on .Net 3.5 and below, we could simplify this condition to just be ```'$(TargetFrameworkVersion)' != 'v4.0'```.

tomspilman:
Thanks @mrhelmut

Issue #5562, opened 2017-03-09
Replace NVorbis with SDL Mixer


cra0zy:
I am gonna replace NVorbis with SDL Mixer so it's easier to port to .net core down the line. Does anyone have any objections (song content pipeline will stay the same)?

CC. @tomspilman @dellis1972 @KonajuGames @mrhelmut

tomspilman:
Some questions first.

What platforms will this affect?

Is SDL Mixer part of our existing SDL or does it require extra native binaries?

Bigger topic...  should we try to standardize on Vorbis for music on all platforms?  Or is there a good benefit to continue to support platform specific formats/decoders/players for music?

cra0zy:
> What platforms will this affect?

DesktopGL.

> Is SDL Mixer part of our existing SDL or does it require extra native binaries?

Extra binary required.

tomspilman:
> DesktopGL.

I don't have a big issue with this if it is DesktopGL only for now.

The extra binary sucks... yet another 32bit/64bit mess to deal with.

@dellis1972 @KonajuGames ???

dellis1972:
I'm good with this  (as long as it works ;P )

dellis1972:
@cra0zy next task after this (if you are up for it) video support on
deksktopgl? ;)

Jjagg:
I saw something about @flibitijibibo getting proper video support in FNA. Maybe we can do something similar.

cra0zy:
I don't think I'll try implementing video support any time soon, if one of you wants to do it just add a comment to #3983 so no duplication of wok occurs.

flibitijibibo:
FWIW, FNA just uses a DynamicSoundEffectInstance with Vorbisfile#:

https://github.com/FNA-XNA/FNA/blob/master/src/Media/Xiph/Song.cs

This is getting replaced with FACT though; along with replacing the XACT runtime I'm also replacing the XNA Sound implementation and folding a Song implementation into it, using stb_vorbis. So this is really only helpful if you want to tie Song into your own DynamicSoundEffectInstance, which has a couple side effects (MasterVolume being the nastiest one).

The VideoPlayer stuff is separate, since TheoraPlay (and soon Theorafile) pull Vorbis data separately from Vorbisfile. If you grab anything from FNA VideoPlayer, wait until I replace the current version; TheoraPlay _loves_ eating up RAM, on top of spinning a pretty expensive thread. (It's kind of a miracle that it works, but it does!)

[...]

PR #5733, opened 2017-05-25
Updating NVorbis fork

mrhelmut:
Since upstream NVorbis doesn't seem to be maintained anymore and that trying to push small fixes there have failed for months, I would like to propose to redirect to a local fork while #5562 is being worked on.

I know that this is something that we want to really avoid, but since we our games are regularly running into NVorbis issues that we fixed locally, and since it would be temporary until #5562 is done, I thought  worthwhile to propose this.

Let me know what you think. I would totally understand if you disagree.

In the event of an approval, I'll give admin rights to this fork to the relevant MG maintainers.

@KonajuGames @cra0zy @tomspilman @dellis1972

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,4 +6,4 @@
 	url = https://github.com/Mono-Game/MonoGame.Dependencies.git
 [submodule "ThirdParty/NVorbis"]
 	path = ThirdParty/NVorbis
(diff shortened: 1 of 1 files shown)


tomspilman:
I'm good with this in the short term.  Worst case we move back to the official fork.

Issue #5706, opened 2017-05-09
Mouse horizontal scroll wheel value


aloisdeniel:
We currently have access to cumulative vertical scroll wheel value through `Mouse.GetState().ScrollWheelValue`, but the horizontal one isn't available. It could be useful to have access to it on macOS where such a gesture is quite common.

Jjagg:
@tomspilman This was fixed in #5735.

PR #5735, opened 2017-05-25
Horizontal scroll

gsfreema:
Provides horizontal scroll through MouseState as requested in #5706.  This implementation handles WinForm, Win8/Universal, SDL, and MacOS.


Jjagg:
Wow, this is a very nice PR, support for 5 platforms and tests! Awesome :)

MonoGame.Framework/Input/MouseState.cs
```diff
@@ -29,25 +29,62 @@ public struct MouseState
...
+            _y = y;
+            _scrollWheelValue = scrollWheel;
+            _leftButton = leftButton;
+            _middleButton = middleButton;
+            _rightButton = rightButton;
             _xButton1 = xButton1;
-            _xButton2 = xButton2;
-		}
+            _xButton2 = xButton2;
+            HorizontalScrollWheelValue = horizontalScrollWheel;
```
KonajuGames: Remain consistent with all the other properties and implement as an explicit backing field.

MonoGame.Framework/Windows/WinFormsGameForm.cs
```diff
@@ -32,6 +32,7 @@ public static System.Drawing.Point GetPointerLocation(this Message msg)
     internal class WinFormsGameForm : Form
     {
         GameWindow _window;
+        public const int WM_MOUSEHWHEEL = 0x020E;
```
KonajuGames: Spelling.  WHEEL, not HWEEL.

MonoGame.Framework/Windows/WinFormsGameForm.cs
```diff
@@ -137,5 +153,15 @@ protected override void WndProc(ref Message m)
...
         }
+
+        private static ushort HiWord(IntPtr word)
+        {
+            return (ushort)(((ulong)word >> 16) & 0xffff);
+        }
+
+        private static short GetWheelDeltaWParam(IntPtr wParam)
+        {
+            return (short)HiWord(wParam);
```
KonajuGames: If it is only used once, and it is a simple cast of another return value, just put this inline in the call site.  Adding it as a separate method is overkill.

MonoGame.Framework/Input/MouseState.cs
```diff
@@ -29,25 +29,62 @@ public struct MouseState
...
-			_rightButton = rightButton;
+        public MouseState(
+            int x,
+            int y,
+            int scrollWheel,
+            ButtonState leftButton,
+            ButtonState middleButton,
+            ButtonState rightButton,
+            ButtonState xButton1,
+            ButtonState xButton2) : this(
```
KonajuGames: Because this is a simple struct, and structs have special rules regarding all fields must be set, I would duplicate the setting of the fields in this constructor rather than calling the other constructor using `this`.

MonoGame.Framework/Windows/WinFormsGameForm.cs
gsfreema: I'm not a fan of the constant, but I was trying to use the same macro name that Microsoft uses.  It's not that I don't spell out WHEEL, it's that there is an H before it that spells HWHEEL.  I know that I did a double take just looking for the proper message macro for a horizontal mouse wheel.
https://msdn.microsoft.com/en-us/library/windows/desktop/ms645614(v=vs.85).aspx

MonoGame.Framework/Windows/WinFormsGameForm.cs
gsfreema: This is another place where I was trying to match common macros that people might try to search for online.  I can remove this function and inline the call though.

Jjagg:
Bump @KonajuGames

KonajuGames:
Thanks for the reminder.  Merging. Thanks @gsfreema

PR #5437, opened 2017-01-24
Recreation of VideoPlayer.WMS update

KonajuGames:
Recreated from #5084 due to GIT craziness.

PR #5737, opened 2017-05-26
Improvements to VideoPlayer for Desktop DirectX.

afrodynamics:
A game project I'm working on requires the video player. I tried playing a 1080p mp4 file in Debug and Release but in either case the playback was unacceptable. Lots of choppiness, and in some cases it was so slow that the program became inoperable and would not respond. For reference, my dev machine has a GTX 660 with 4GB of VRAM.

I went for an easy fix here, per the comment that was left in `GetTexture()`, and just cached a Texture2D. This  pretty much resolved my playback issues, although I am certain there's more to optimize here. Any and all feedback here would be greatly appreciated!

**Changes**

- DX VideoPlayer now caches a Texture2D instead of new-ing one for each call to GetTexture()
- Fixed null pointer exception if Stop() is called twice in a row ([XNA](https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.media.videoplayer.stop.aspx) documentation seems to suggest that doing this would fail silently.)

### DIFF
--- a/MonoGame.Framework/Media/VideoPlayer.WMS.cs
+++ b/MonoGame.Framework/Media/VideoPlayer.WMS.cs
@@ -153,7 +153,10 @@ namespace Microsoft.Xna.Framework.Media
                 _volumeController.Dispose();
                 _volumeController = null;
             }
-            _clock.Dispose();
+            if (_clock != null)
+            {
+                _clock.Dispose();
+            }
             _clock = null;
         }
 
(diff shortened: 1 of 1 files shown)


Jjagg:
Yeah, the video player in develop branch is not great. #5437 also keeps a texture and contains other fixes/improvements. @KonajuGames ?

tomspilman:
@KonajuGames - I am leaning towards merging this then letting you work out the conflicts on your old PR.  

This PR is simple yet fixes a rather big issue with the current player.  Your PR has some additional improvements and a unit test... so I would hate to loose it.

Make sense?

KonajuGames:
Sure.

tomspilman:
Thanks @KonajuGames !

and thanks @afrodynamics !

romanov:
@afrodynamics thank you! Finally i can use development build

PR #5741, opened 2017-05-29
Exposed classes & functions to allow better threaded texture loading

roponator:
We implemented better threaded texture loading in our games to speed up loading.

The current monogame threaded approach does not allow for good threading as it mixes loading stuff from files and graphics api dependent texture creation.

There are 2 ways we can get our approach to work with future versions of monogame, either:

a) (this PR) expose some monogame classes so we/other users can override them

b) Add one or two functions to ContentTypeReader, ContentReader, ContentManager & Texture2DReader which take an additional callback parameter and reorder some things so that it's more task/threading friendly.

An alternative would be to hack it using reflection but that is something we don't wish to do.
Method a) would allow future specialization for developers who are in need of it.


KonajuGames:
I am not in favour of wholesale exposure of everything in these classes.  We need to be be smarter about the design of this than laying bare all of the internals.  I understand the requirements of threaded loading and the need to separate the I/O tasks from the GPU tasks.  I have some thoughts on this that I will be proposing in an implementation soon.

tomspilman:
> I am not in favour of wholesale exposure of everything in these classes.

Yeah I am in agreement.  @dellis1972 @cra0zy ?

Jjagg:
Closing because this is not the path we want to take for threaded content loading. @KonajuGames Any chance you remember what you were planning and could share the general idea?

PR #5743, opened 2017-05-30
keyboard fix when keyboard stops working when window focus is lost so…

roponator:
For UWP: Sometimes when the window loses focus keyboard keyDown/up/character recieved events stop working.

The added calls to IsInputEnabled = true fix it.

These values are always true though if we print them, so calling them seems to reset something internaly.

Using "CoreWindow.GetForCurrentThread()" instead of "window" in InputEvents ctor does not fix it.

Does anyone else have any idea why this happen?


jakepoz:
We've been experiencing this in our UWP games, and @roponator 's fix does solve the problem. You have to be in your application, then leave focus by clicking into another window, then clicking back into your application. Usually on the first time this focus switch happens, we lose keyboard input. If you alt-tab and back again then it works.

jakepoz:
Hey @KonajuGames or @tomspilman Can you please take a look at this? We are experiencing this bug and want to get a fix in without having to maintain a separate branch of Monogame.

KonajuGames:
I'm good with this.  Odd fix, but it would seem to waking up something internal by simply requesting its state.  Merging.

Issue #3794, opened 2015-04-30
Upgrade SharpDX to the latest version.


Chaosus:
I found MonoGame uses old version of SharpDX(2.5.1). Should we start using 2.6.3 or wait for 3.0 ?

KonajuGames:
The Windows 10 support uses the latest version, and the plan is to migrate the rest over to that version soon.

Shqrdx:
Ah, good to know. I close this issue when its happened.

tomspilman:
So we use SharpDX 3.0 but there are even newer versions.  We need someone to fix the issues around supporting the latest SharpDX releases and test it on all the SharpDX platforms.

DDReaper:
As discovered while doing the NuGet SharpDX dependancies, I did find that versions later than Sharp DX 3.0 do not work at all.

Nothing ends up being drawn, so I'm guessing something has fundamentally changed.

If we are retargeting, we should aim for the latest 4.x release (my two pence worth)

Any ideas what could be causing this drawing breakdown? (can easily test with the new NuGet packages and simply update the SharpDX NuGet dependency to a later level.

mrhelmut:
Both @Jjagg & @vpenades have done some work in that regard, and seem to have the runtime working on SharpDX 4.0.

Jjagg:
I upgraded to 3.1 back then, just a few adjustments needed to get the runtime working. I ran into issues with the HLSL compiler though, which is why I didn't set up a PR. I lost track of it afterwards.

afrodynamics:
Upgrading SharpDX is a great idea, and I think it would go a long way to resolving some issues we've had with alt-tab and "device lost/reset" events in our game.

If you take a look at PlatformPresent() you'll notice that the Present() call on the swap chain is void, despite MSDN saying it returns an [HRESULT code](https://msdn.microsoft.com/library/windows/desktop/bb509553). We are currently not handling any of the potential error codes here, and probably should be. The graphics device events (DeviceLost, DeviceReset, etc) are still unimplemented, and I believe upgrading sharp DX would help here, and probably in other places too.

This all might be just a natural part of the upgrade, but I figure I should mention it here.

vpenades:
I've just did a proof of concept:

- Checked out latest Monogame Dev. - working on Windows Desktop Platform
- Removed SharpDx 2.6.1 references and replaced them with SharpDx 4.0.1 Nuget Packages
- Amended code:
[...]

PR #5748, opened 2017-06-01
Use SharpDX NuGet packages from our NuGet packages

DDReaper:
The NuGet's now inherit SharpDX direct from NuGet instead of including the dll's directly.

Of note, I had to use the Version 3.0.0 packages (same as what MonoGame uses now), anything higher and it simply didn't work.  No errors, it just didn't draw anything so I suspect we have an issue somewhere

### DIFF
--- a/NuGetPackages/MonoGame.Framework.Windows8.nuspec
+++ b/NuGetPackages/MonoGame.Framework.Windows8.nuspec
@@ -16,9 +16,17 @@ This package provides you with MonoGame Framework that works on Windows 8 as a n
     <summary>This package provides you with MonoGame Framework that works on Windows 8 as a native store app.</summary>
     <copyright>Copyright © 2009-2017 The MonoGame Team</copyright>
     <language>en-US</language>
-    <references>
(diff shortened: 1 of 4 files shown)


DDReaper:
@tomspilman any chance we can get a "build" dev feed for NuGet setup?
I've tested all these locally and they work fine, however, because the build system doesn't publish these until they are merged, it's tricky to validate the "built" version is working.

Thoughts?

tomspilman:
> @tomspilman any chance we can get a "build" dev feed for NuGet setup?

Maybe... I would have to research how to do that.  Like the PR builds we wouldn't want to keep those NuGets and artifacts around forever.  We would want them to get cleaned up fairly quickly after a build/merge.

tomspilman:
> I had to use the Version 3.0.0 packages (same as what MonoGame uses now), anything higher and it simply didn't work.

This is good for now.  We should focus on https://github.com/MonoGame/MonoGame/issues/3794 next.

DDReaper:
Agreed, need to sort out why SharpDX breaks after 3.0

PR #5750, opened 2017-06-02
Better support for WAV audio formats in content pipeline and FromStream

KonajuGames:
Add support for MS-ADPCM, IMA/ADPCM, 8-bit, 16-bit, 24-bit, float in both content pipeline and SoundEffect.FromStream().


KonajuGames:
Rebased on develop with SharpDX 3.1.1.

KonajuGames:
I'd like to get this merged, and I think the only way it will get real coverage is to put it in the develop branch and into real projects.

tomspilman:
Sorry for the delay @KonajuGames !

nkast:
I have a 32-float .wav that fails with `HRESULT: [0x88960001], Module: [SharpDX.XAudio2], ApiCode: [XAUDIO2_E_INVALID_CALL/InvalidCall], Message: Unknown` when I try to create an instance.

It works when quality is BEST but fails when I set quality to Medium (ADPCM).
Here is the sample project.

https://1drv.ms/u/s!AgIeh29TX5UhkPUbl79UeBJ1Nmo-Zw

KonajuGames:
Thanks for the report, @nkast. I'm trying to grab that sample project now, but OneDrive is not playing ball.  I'll try to repro with the 32-bit float WAV I have here.

KonajuGames:
I can repro here, so working on a fix now.

KonajuGames:
Ok, so the actual error is

    XAudio2: ADPCM wSamplesPerBlock must be 32, 64, 128, 256 or 512

It would appear that XAudio2 cannot handle all ADPCM files generated by ffmpeg, and I'm having trouble finding any parameters to tell ffmpeg to generate the correct files.

nkast:
Thanks for looking at that.
I see that WaveFormatAdpcm.SamplesPerBlock is calculated automatically, so tried to see what input is valid. 
```
            for(int i = 0; i<1024 * 16; i++)
            {                
                try
                {
                    var format1 = new WaveFormatAdpcm(33075, 2, i);
                    var voice1 = new SourceVoice(Device, format1, VoiceFlags.UseFilter, XAudio2.MaximumFrequencyRatio);
                    System.Diagnostics.Debug.WriteLine(i);
                }
                catch (Exception) { }
            }
```
blockAlign 44 -> SamplesPerBlock = 32
blockAlign 76 -> SamplesPerBlock = 64
blockAlign 140 -> SamplesPerBlock = 128
blockAlign 268 -> SamplesPerBlock = 256
blockAlign 524 -> SamplesPerBlock = 512
SamplesPerBlock is blockAlign-12.
It's possible to overwrite the values of WaveFormatAdpcm.SamplesPerBlock but at this point I am not sure if that's right.
I found this [sample](https://code.msdn.microsoft.com/windowsdesktop/DXUTDirectXTK-Simple-Win32-9cf797e9/sourcecode?fileId=130257&pathId=1691502359) at code.msdn that gives the following check:
```
            int nHeaderBytes = 7 /*MSADPCM_HEADER_LENGTH*/ * wfx->nChannels; 
            int nBitsPerFrame = 4 /*MSADPCM_BITS_PER_SAMPLE*/ * wfx->nChannels; 
            int nPcmFramesPerBlock = (wfx->nBlockAlign - nHeaderBytes) * 8 / nBitsPerFrame + 2; 
 
            if ( wfadpcm->wSamplesPerBlock != nPcmFramesPerBlock ) 
            { 
                DebugTrace( "ERROR: Wave format ADPCM %u-channel with nBlockAlign = %u must have wSamplesPerBlock = %u (%u)\n", 
                            wfx->nChannels, wfx->nBlockAlign, nPcmFramesPerBlock, wfadpcm->wSamplesPerBlock ); 
                return false; 
            } 
```  
this suggests that `SamplesPerBlock = (BlockAlign*2 - 28)` which does not fit exactly with  what I found above.
Maybe the right way to create a WaveFormatAdpcm is by adjusting the blockAlign parameter so the  SamplesPerBlock end up with a value of {32,64,128,256,512}, something like `new WaveFormatAdpcm(33075, 2, blockalign / channel  +/- something...);`

KonajuGames:
We have the equations in our code to convert block alignment to samples per
block, and it is not as simple as -12 or anything like that. The block
alignment comes from the WAV format header, and it is ffmpeg that is
generating this. For ADPCM files, ffmpeg has a hard-coded size that is
beyond that which XAudio2 accepts. My next task will be to write my own
ADPCM encoder as there does not appear to be any other portable solution
available to us at this stage. I spent a long time looking.

nkast:
I just tested the same file from an XNA xnb, and It loads just fine. The xnb contains those data:
format = 2 (adpcm)
channels = 2
sampleRate = 21979
blockAlignment = ***140***
sampleBits = 4


So I guess there's nothing broken in the SoundEffect.XAudio. We should start from here and work backwards. 
This looks like an issue with the SoundEffectWriter, it probably should write (blockHeaderSize+blockDataSize) in the blockAlign field. 

once it produces correct XNBs we can fix the gl platforms to handle blockAlign correctly.

nkast:
BTW, I think in ffmpeg the blocksize is controlled by setting the quality, since blockSize in adpcm affects directly the compression/quality.

KonajuGames:
ffmpeg has a #define BLOCK_SIZE, so it is not configurable. The block size
needs to be taken into account when encoding the ADPCM file as it affects
the initial values of the predictor.

nkast:
I tried a few parameter with no luck and then gave up. I didn't find the #define BLOCK_SIZE. good to now.

Issue #5749, opened 2017-06-01
Guidance for contributing to documentation


HopefulToad:
I've been a user of MonoGame for some time, and I've liked the idea of contributing to the project. The problem is, I know very little about the inner workings of the framework and I doubt I would be able to help with any of the coding issues.

One area I probably _could_ help is in documentation. Before I start submitting PRs for documentation changes out of nowhere, though, I want to discuss it so that I don't end up wasting my time or anyone else's. I apologize if this is the wrong place to bring it up (since this isn't exactly an issue with MonoGame, but rather _contributing_ to MonoGame).

A few things that might be problematic to me helping with documentation:

- My level of understanding of MonoGame is much more that of an end-user than a contributor. In other words, I'm familiar with the API, but not the implementation.
- I'm a Windows user only, and almost all of my experience with MonoGame is with working on desktop DirectX projects. I'm unfamiliar with UWP and I'm still using Windows 8.1, and I haven't really tried using OpenGL seriously.
- My knowledge is mostly limited to 2D game development.
- I've never actually published a video game, but I have been working on projects off and on for two years.

With all those caveats, should I even bother? I do think I have enough knowledge to at least contribute a little.

For documentation guidelines, I've seen #2520, which is somewhat old and it was never really confirmed in that discussion what the final outline would be. There's also the README in the documentation folder, but that just has a TODO under the style guide for manuals, guides, and tutorials.

If I could help with documentation, any guidance on how to get started would be appreciated.

tomspilman:
Hi @HopefulToad .

> I'm familiar with the API, but not the implementation.

I think then the best thing you could possibly do is help convert an existing XNA tutorial to MonoGame and add it to our docs.  It would go a long way towards helping people with starting with MonoGame and also give people some reference on how to do things.

Riemer from [http://www.riemers.net](http://www.riemers.net) has offered to let us use his tutorial in MonoGame documentation.  It would be sort of a big job... first moving it over with correct markup and then doing some improvements to it to meet the current state of MonoGame.

If you are interested let me know and we can start an issue on it.

HopefulToad:
@tomspilman That sounds good to me.

> first moving it over with correct markup and then doing some improvements to it to meet the current state of MonoGame

So does that mean the plan should be to first move the entirety of each tutorial over verbatim, and then go back and fix things up to make it MonoGame specific? Or to make changes to sections as I go?

tomspilman:
[...]

PR #5753, opened 2017-06-04
Riemer's tutorial port

HopefulToad:
As per #5749, this is the beginning of porting Riemer's XNA tutorials over to the MonoGame documentation. For now, this is a verbatim port, which means some parts don't apply to MonoGame and will have to be changed later.

For comparison, see the home page at <http://riemers.net/>.

For those curious where we were given permission to use the tutorials, see [this thread](http://community.monogame.net/t/riemers-tutorials-monogame-net/7762) over at the MonoGame website. I do think it would probably be a good idea to contact Riemer again just to make sure it's acceptable for us to use the tutorial content in this way.

I made a few assumptions in this PR about how the tutorial might be formatted and placed in the Documentation section of the MonoGame website, so if any of the assumptions were made in error, please let me know.

Also, would it be a good idea to structure the .md files for the tutorial port within folders within the Documentation folder? If I just dump them all directly in the Documentation folder, it's going to become overcrowded fast.

@tomspilman @KonajuGames


DDReaper:
Good, that's one less series I need to archive on the new XNAGameStudio Archive :D

tomspilman:
@HopefulToad - Great first push...  I just have a few notes.

> I do think it would probably be a good idea to contact Riemer again 

I'll take care of that.

> If I just dump them all directly in the Documentation folder, it's going to become overcrowded fast.

Agreed...  I suggest putting all of this... including the images... under a folder called `Riemer`.  Adding further subfolders under there for images and maybe one for each main tutorial series would be ok too.

Once you do that then maybe start working on pulling the first "2D Series 1: Shooters" tutorial over.

HopefulToad:
> Agreed... I suggest putting all of this... including the images... under a folder called Riemer. Adding further subfolders under there for images and maybe one for each main tutorial series would be ok too.

OK, I've moved everything to the Riemer subfolder, which has one subfolder for each tutorial and an images folder that also has one subfolder for each tutorial.

When working on the tutorial, should I submit a new PR for each page of the tutorial, or combine them into larger PRs?

tomspilman:
@HopefulToad - If you want to see what this looks like you just need to open the index.html in the Documentation.zip in the artifacts for the Documentation build on TeamCity.  You do that here:

http://teamcity.monogame.net/viewLog.html?buildId=46051&buildTypeId=MonoGame_GenerateDocumentation&tab=artifacts&guest=1

Here is a direct link:

http://teamcity.monogame.net/repository/download/MonoGame_GenerateDocumentation/46051:id/Documentation.zip%21/index.html?page=main

> an images folder that also has one subfolder for each tutorial.

Note the images don't see to be linking right.  I expect like most HTML the folder links are all relative to the root.

> should I submit a new PR for each page of the tutorial, or combine them into larger PRs?

I think we should focus on getting all of that first "2D Series 1: Shooters" tutorials moved over.  It would be fine to just do that from this PR.

Then I think we should look to MonoGame-ize that series completely.  Including fixing instructions on how to create the project and edit content.

After that we can review where we're at and tackle the other 3D tutorials.

HopefulToad:
> Note the images don't see to be linking right. I expect like most HTML the folder links are all relative to the root.

@tomspilman  Weird... The images aren't showing up in the TeamCity build, but I see them when I view the riemers_tutorial_intro.md file on Github. When I had the references to them as Riemer/Images/[image name] they weren't showing up on Github, but then I changed them to just Images/[image name] because the tutorial intro was already in the Riemer subfolder, and then the images appeared.

I couldn't find the images anywhere in the Documentation.zip. Also, there was no "Riemer" subfolder in the html folder within the Documentation.zip, either. But I downloaded the Documentation.zip from the TeamCity build that occurred before I migrated the files to subfolders, and the images are there in that one. Something must have went wrong when I moved everything around.

Can you confirm that the riemers_tutorial_intro.md file on Github shows the images? I don't really know what's wrong.

Also, when I try to view the tutorial intro from the index.html, I get a weird bug where it doesn't let me scroll all the way down to the end of the file.

HopefulToad:
What's also weird is that the link to the intro in the tutorials.md file works in MonoGame, but is broken in the TeamCity documentation build. It works in the older, non-subfoldered TeamCity build, though.

Then there's these lines in the config.xml file, which may be why the images in the most recent TeamCity build are nowhere to be found:

    
    <resource>images</resource>

DDReaper:
Except, the point of the archived @tomspilman is to also update them to me more MonoGame specific. The samples there are only temporary until we can get them updated.
Same with Shawn's articles I upgraded recently, although those only needed very minor changes, the articles themselves were fine.

These should live in the MonoGame docs eventually, or be submoduled in to the project.

cra0zy:
> @DDReaper - I suggest to still archive it. We plan to modify these to fit our needs for MonoGame, so it won't be like the original after we're done here.

I'll just point out out one small hole with creating any type of samples/examples, we have literally broke compatibility with every release. We should probably fix up the templates and project format first, so we never have to break it again.

HopefulToad:
@tomspilman OK, so I moved the images over to the "images" folder that already existed, in a subfolder called "Riemer". Now the images appear in the TeamCity documentation build, but not on Github, so I guess that's an improvement?

When linking to the images, I now use images/Riemer/[image name]. This works in the Documentation build, because it moves the riemers_tutorial_intro.md file to the base "html" folder instead of keeping it nested in a "Riemer" folder. On Github, however, it doesn't work because the base folder is the "Riemer" folder, meaning it's trying to find an image at Riemer/images/Riemer/[image name], which doesn't exist.

So I can either...
<ol>
<li>Keep the tutorial files nested in the Riemer folder, and either
<ol><li>Have images appear correctly in Github, but not in the documentation build (obviously not a solution) or</li>
<li>Have images appear correctly in the documentation, but not on Github (not ideal when trying to port the tutorials over)</li></ol>
</li>
<li>Put all the Riemer tutorial files directly in the documentation folder</li>
</ol>

This all seems to be a byproduct of the way the HTML files are generated from the markdown files. They're all placed in the same "html" folder in the Documentation build, even if the markdown files were originally in nested folders, which results in links to the images being broken, unless correct for that when linking to the images.

HopefulToad:
I'm going to assume that the best option is to revert to putting all tutorial files directly in the Documentation folder, so that the tutorial files don't appear broken on Github and also aren't actually broken on the MonoGame website.

HopefulToad:
@tomspilman The initial port of the 2D series is complete. While translating it into Markdown, I corrected a few obvious typos, but I noticed quite a few more errors in grammar, spelling, and content that I'd like to go through and correct. Should I do that in a second PR, or this one?

HopefulToad:
I let this project fall by the wayside a while ago when no one answered the questions I had... I'm not sure this is relevant anymore. I'm going to go ahead and close this.

PR #5755, opened 2017-06-05
Added the Portable reference to the platform NuGets to resolve #5724

### LINKED ISSUE #5724: MonoGame.Framework.Portable is broken
Going by the description, the MonoGame.Framework.Portable nuget package should be a bait and switch PCL, however it does not have the same API surface as the platform specific ones, and as a matter of fact is not a bait and switch PCL at all (look at MonoGame\NuGetPackages\MonoGame.Framework.Portable.nuspec and the only target is the PCL platform moniker - nuget packages targeting multiple platforms should be structured [like this](https://docs.microsoft.com/en-us/nuget/guides/create-cross-platform-packages)).
This includes the latest prerelease version as well.

To test, create a monogame project (I am using the Monogame UWP template), and a PCL project, then reference the PCL from the main project.

Add the MonoGame.Framework.Portable package to the PCL project and try to build: you'll get a conflict because both the PCL nuget package and the UWP try to copy their different versions of Monogame.Framework.dll to the same location.

Try adding the PCL nuget to both projects instead of the platform specific one and MonoGame.Framework is not present anymore.

This means the only way of sharing code in a game targeting different platforms is to use a shared project like in the template, which is not at all good practice.

DDReaper:
Added the Portable reference to the platform NuGets to resolve #5724

This ensures projects using the newer .csproj formats won't get confused. 

Users will now only add the same NuGet to both PCL and Platform in cases where this affects.

### DIFF
--- a/NuGetPackages/MonoGame.Framework.Android.nuspec
+++ b/NuGetPackages/MonoGame.Framework.Android.nuspec
@@ -23,5 +23,9 @@ This package provides you with MonoGame Framework that works on Android.</descri
     <!-- C# Assemblies -->
     <file src="Android\AnyCPU\Release\MonoGame.Framework.dll" target="lib\MonoAndroid\MonoGame.Framework.dll" />
     <file src="Android\AnyCPU\Release\MonoGame.Framework.xml" target="lib\MonoAndroid\MonoGame.Framework.xml" />
+
+    <!-- Portable Libs -->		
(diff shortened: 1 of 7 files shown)


tomspilman:
@dellis1972 @KonajuGames - Leaving this to you guys to decide.  I know nothing about NuGet and if this is a good idea or not.

tomspilman:
Going to hold this until after 3.7.

DDReaper:
Replaced with #5799

PR #5758, opened 2017-06-07
Added new Circle class with tests

DDReaper:
After reviewing some old archives, came across Kris Steele‏ @KrisWD40 Circle class code, Sourced from http://krissteele.net/blogdetails.aspx?id=251

So with his permission, I thought this would be a nice addition to our intersection classes, like Rectangle and so on.  Granted it's pimped up a bit to include all the other standard comparators and intersects like Rectangle. 

Also included a new Test class to validate it as well.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -324,6 +324,7 @@
     <Compile Include="Quaternion.cs" />
     <Compile Include="Ray.cs" />
     <Compile Include="Rectangle.cs" />
+    <Compile Include="Circle.cs" />
     <Compile Include="ReusableItemList.cs" />
     <Compile Include="SharedGraphicsDeviceManager.cs">
       <Platforms>Windows8,WindowsPhone81</Platforms>
--- a/MonoGame.Framework/Circle.cs
+++ b/MonoGame.Framework/Circle.cs
@@ -1,0 +1,441 @@
+// MIT License - Copyright (C) The Mono.Xna Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Xna.Framework
+{
+    /// <summary> 
+    /// Represents a 2D circle. 
+    /// Developed by Kris Steele‏ @KrisWD40, Sourced from http://krissteele.net/blogdetails.aspx?id=251
+    /// </summary> 
+    public struct Circle
+    {
+        #region Private Fields
+        private Vector2 direction;
+        private float distanceSquared;
+
+        #endregion
+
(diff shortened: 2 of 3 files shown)


Jjagg:
I remember a quality PR for implementing a Circle before that got rejected (after some discussion). It was merged into MG.Extended instead. #3419

cra0zy:
I always wanted for that PR to get merged... I mean it's a circle...

LithiumToast:
See [CircleF](https://github.com/craftworkgames/MonoGame.Extended/blob/develop/Source/MonoGame.Extended/CircleF.cs) in MonoGame.Extended. If developers don't want to use the whole library they can just copy and paste it into their project as long as they meet the requirements of the MIT license.

cra0zy:
Just gonna add that I love your F types in MG Extended, ie. RectangleF, makes some stuff a lot easier :)

LithiumToast:
@cra0zy Absolutely. The `F` suffix is to indicate that the structure uses single-precision floating-point. To keep consistency with legacy XNA which for example has `Rectangle` which uses integers, `Circle` should probably use integers as well. The pattern, i.m.o. is not the best, but it is consistent with legacy XNA.

DDReaper:
It did originally use ints, but changed as it have me some issues with the math.
Can change it back if that would make this more acceptable.

Only put it in as it seemed cool from Kris's original circle class.  Was just a fun experiment and didn't take to much time.

So it's this worth it? Ints, floats or otherwise?

LithiumToast:
Wether to have integers, floats, etc for bounding volume shapes is going to depend on some factors. One such factor to consider is _robustness_. For example, when intermediate calculations become larger than can be represented by a floating-point or an integer variable the intermediate result will be invalid. If such problems are not detected, the final result of the computation is likely to be incorrect. 

Thinking of ways of how people are going to use the code however is somewhat difficult when writing library code; it's quite likely not everyone's game is going to be the same or have the same requirements. At the same time, I would dare to say integers and floats are the most common representations of numbers that game developers use or need, but there is always the possibility that some developer may wish or need to use something different. Challenges of these types become a common theme when writing code for MonoGame.Extended.

If you guys can all agree on what kind of geometric primitives people would want to use and for what reasons I think it would worthwhile to have those primitives in MonoGame instead of MonoGame.Extended.

DDReaper:
I can see your point.  When researching I found a variance even within MonoGame itself.  Rectangles using INT's whereas Quaternions using floats for example.
Although I can also see advantages of having base INT versions (e.g. Circle) and an alternate FLOAT based version (e.g. fCircle).  Whether these are in MG, split between MG and MGE or all in MGE is another debate.

My view is such things are CORE structs / classes that get used by all games. So having them core in MG makes more sense.

In doing Ziggyware archiving today, there were several features which are also core, like AABB functions / classes for instance, which are also key.

PR #5759, opened 2017-06-07
Account for non-4x4 block texture compression formats

### LINKED ISSUE #5598: Fix Texture2D.ValidateParams() for Compressed Formats
The current implementation of `Texture2D.ValidateParams()` is broken for compressed formats (likely for all... but in particular PVRTC).  This is a critical issue that needs to be resolve ASAP as it makes MonoGame useless for most game development.

It doesn't consider bits per pixel or that there is a min size for some mip levels.  You can see my findings here:

https://github.com/MonoGame/MonoGame/issues/5557#issuecomment-287682593
https://github.com/MonoGame/MonoGame/issues/5557#issuecomment-287683524

We need to add some specific unit tests for setting/getting data on compressed formats including data on every mip level.

### LINKED ISSUE #5600: Fix or remove `checkedRect` from Texture2D Set/GetData
The `checkedRect` change in Texture2D Set/GetData is broken for compressed texture formats:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.cs#L149

It doesn't consider that for compressed formats there is a minimum mipmap size that must be enforced.  So it passes the incorrect rectangle down to methods for setting the texture data.

IMO we sure purge this  `checkedRect` business entirely from the code base.

In particular the removal of calling SetData with a `null` rectangle is likely a performance hit.

I am willing to bet that for example `GL.TexSubImage2D` is slower than `GL.TexImage2D` when setting texture data.  Same with the methods under DirectX when the full texture level was being set instead of a sub-rectangle of the texture surface.

IMO we need to refactor all this to restore the old behavior for `null` rectangles and fix the checking when a rectangle is passed to consider compression needs.

KonajuGames:
PVR 2bpp formats use 8x4 blocks rather than the more common 4x4 blocks.

Tested on Windows and WindowsGL (DXT).  Needs testing on iOS where PVR is used.

Fixes #5600
Fixes #5598

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsExtensions.cs
+++ b/MonoGame.Framework/Graphics/GraphicsExtensions.cs
@@ -818,18 +818,18 @@ namespace Microsoft.Xna.Framework.Graphics
                 case SurfaceFormat.Dxt1a:
                 case SurfaceFormat.RgbPvrtc2Bpp:
                 case SurfaceFormat.RgbaPvrtc2Bpp:
+                case SurfaceFormat.RgbPvrtc4Bpp:
+                case SurfaceFormat.RgbaPvrtc4Bpp:
                 case SurfaceFormat.RgbEtc1:
-                    // One texel in DXT1, PVRTC 2bpp and ETC1 is a minimum 4x4 block, which is 8 bytes
+                    // One texel in DXT1, PVRTC (2bpp and 4bpp) and ETC1 is a minimum 4x4 block (8x4 for PVRTC 2bpp), which is 8 bytes
                     return 8;
                 case SurfaceFormat.Dxt3:
                 case SurfaceFormat.Dxt3SRgb:
                 case SurfaceFormat.Dxt5:
                 case SurfaceFormat.Dxt5SRgb:
-                case SurfaceFormat.RgbPvrtc4Bpp:
-                case SurfaceFormat.RgbaPvrtc4Bpp:
                 case SurfaceFormat.RgbaAtcExplicitAlpha:
                 case SurfaceFormat.RgbaAtcInterpolatedAlpha:
-                    // One texel in DXT3, DXT5 and PVRTC 4bpp is a minimum 4x4 block, which is 16 bytes
+                    // One texel in DXT3 and DXT5 is a minimum 4x4 block, which is 16 bytes
                     return 16;
                 case SurfaceFormat.Alpha8:
                     return 1;
@@ -905,6 +905,37 @@ namespace Microsoft.Xna.Framework.Graphics
             return 0;
         }
 
+        public static void GetBlockSize(this SurfaceFormat surfaceFormat, out int width, out int height)
+        {
+            switch (surfaceFormat)
+            {
+                case SurfaceFormat.RgbPvrtc2Bpp:
+                case SurfaceFormat.RgbaPvrtc2Bpp:
+                    width = 8;
+                    height = 4;
+                    break;
+                case SurfaceFormat.Dxt1:
+                case SurfaceFormat.Dxt1SRgb:
+                case SurfaceFormat.Dxt1a:
+                case SurfaceFormat.Dxt3:
+                case SurfaceFormat.Dxt3SRgb:
+                case SurfaceFormat.Dxt5:
+                case SurfaceFormat.Dxt5SRgb:
+                case SurfaceFormat.RgbPvrtc4Bpp:
+                case SurfaceFormat.RgbaPvrtc4Bpp:
+                case SurfaceFormat.RgbEtc1:
+                case SurfaceFormat.RgbaAtcExplicitAlpha:
+                case SurfaceFormat.RgbaAtcInterpolatedAlpha:
+                    width = 4;
+                    height = 4;
+                    break;
+                default:
+                    width = 1;
+                    height = 1;
+                    break;
+            }
+        }
+
 #if OPENGL
 
         public static int GetBoundTexture2D()
@@ -949,7 +980,7 @@ namespace Microsoft.Xna.Framework.Graphics
             }
         }
 #endif
-    }
+            }
 
     internal class MonoGameGLException : Exception
     {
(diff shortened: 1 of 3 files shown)


tomspilman:
Can anyone give this a try on iOS?  @Jjagg @dellis1972 ?

Jjagg:
I can't, don't have a Mac.

dellis1972:
How do you generate a 8x4 PVR texture using the content pipeline?

KonajuGames:
You would have to write a custom content pipeline extension that uses the 2bpp PVRTC BitmapContent types.  Though I have just discovered a bug in `PvrtcRgb2BitmapContent` from looking at it.  It returns the wrong `SurfaceFormat` in `TryGetFormat`.

KonajuGames:
#5772 fixes `PvrtcRgb2BitmapContent.TryGetFormat()`

KonajuGames:
Rebased to develop with SharpDX 3.1.1.

KakCAT:
Hi, with iOS and a 256x256 texture, compressed and mipmapped, I'm getting the following error:

    Microsoft.Xna.Framework.Graphics.MonoGameGLException: GL.GetError() returned InvalidValue

at

    GL.CompressedTexImage2D(TextureTarget.Texture2D, level, glInternalFormat, w, h, 0, wBlocks * hBlocks * blockSize, IntPtr.Zero);

the exception happens when level is 6, w&h is 4, glInternalFormat is 35840

Jjagg:
@KonajuGames The issue is here: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs#L89-L90

```
if (glFormat == (PixelFormat)GLPixelFormat.CompressedTextureFormats)
{
    int blockSize = format.GetSize();
    int wBlocks = (w + 3) / 4;
    int hBlocks = (h + 3) / 4;
    GL.CompressedTexImage2D(TextureTarget.Texture2D, level, glInternalFormat, w, h, 0, wBlocks * hBlocks * blockSize, IntPtr.Zero);
}
```
This code should also check the actual block size instead of assuming 4x4.

Jjagg:
@KakCAT Are you building from source? If you are, can you try changing the failing code to 
```
int blockSize = format.GetSize();
int wBlocks = (w + 7) / 8; // iOS compression uses a block width of 8
int hBlocks = (h + 3) / 4;
GL.CompressedTexImage2D(TextureTarget.Texture2D, level, glInternalFormat, w, h, 0, wBlocks * hBlocks * blockSize, IntPtr.Zero);
```
and check if that fixes it?

KakCAT:
Hi @Jjagg 

I tried the fix but unfortunately it's not fixing the problem. Also, with the fix now the exception happens in level=0 instead of level=6

KakCAT:
I don't know if this is relevant or not, but the problem only appears when using Compression + MipMap.

Color+NoMipMap, Color+MipMap and Compression+NoMipMap work.

KonajuGames:
That is a case I missed. Our content pipeline by default doesn't actually
produce any textures that don't use 4x4 blocks, so the issue is something
else. iOS does use 4x4 blocks for 4bpp PVRTC. 8x4 is used for 2bpp PVRTC.

I have a new Mac now so hopefully the issue shows up in the iOS simulator.

KakCAT:
I just checked that it also fails in the iOS simulator.

KonajuGames:
Fixed it.  PVRTC image size cannot be calculated the same way as every other compressed texture format.

Also rebased on latest develop branch.

KakCAT:
I think I've applied all the changes but I'm getting the following error:

    elementCount is not the right size, elementCount * sizeof(T) is 32, but data size is 8.

It usually complained in the format "is X, but data size is X/2" . Is it possible that the content builder's PVRTC code must also be modified?

KonajuGames:
The content builder code was fine.  The validation of texture parameters was missing the special case for PVRTC data size calculations.

tomspilman:
@KonajuGames - Any sort of unit tests we can add to this to ensure it is fixed and doesn't occur again?

mrhelmut:
I've tested this PR on iOS with compressed texture and it now works. This is a much needed work. :)

dellis1972:
@KakCAT if you can confirm this works I'll merge it

KonajuGames:
> Any sort of unit tests we can add to this to ensure it is fixed and doesn't occur again?

We could, but only for the texture compression supported by the platform the tests are running on.

KakCAT:
I've tested all combinations of texture compression and mip mapping for iOS and all of them are working now. 

However I'm getting InvalidEnum in android when using a compressed texture here:

https://github.com/KonajuGames/MonoGame/blob/c2b7f71472a064267b24911412944a339779959c/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs#L216

(comes from the GL.CompressedTexSubImage2D)

Color+NoMipMap and Color+MipMap works well. Others won't.

note: previously in a conversation with jjagg I told him I was getting a black texture, however I forgot to notify it. It was probably the same exception I'm having here but I probably had it masked behind a try/catch in production code.
In other words, the problem is almost sure not from the fix, but was there before it.

dellis1972:
@KakCAT android compression textures are very device specific. So if you have a ATI chip PVRTC, DXT won't work.. only Color. Color16 and ATITC (or ATI) will work . PVRTC will work on devices which have a PowerVR chip.. and DXT Nvidia.

dellis1972:
Also I think only ETC1/2 is supported on emulators.. but I could be wrong about that.

KakCAT:
> @KakCAT android compression textures are very device specific. So if you have a ATI chip PVRTC, DXT won't work.. only Color. Color16 and ATITC (or ATI) will work . PVRTC will work on devices which have a PowerVR chip.. and DXT Nvidia.

You're right, this is probably the reason. In other projects I had Etc1Compressed but in the test project I just had "Compressed" because I shared it for iOS and Android. It's probably not choosing Etc1.

KonajuGames:
> in the test project I just had "Compressed" because I shared it for iOS and Android. It's probably not choosing Etc1.

`Compressed` will select the best compression for the target platform.  Built for iOS, it will produce PVRTC and for Android it will choose ETC1 for opaque content and RGBA4444 for content with any non-opaque alpha.  If the content is built for iOS then the XNBs shared with iOS, then they will be using PVRTC.

PR #5763, opened 2017-06-11
throw NoSuitableGraphicsDeviceException when OpenGL version not found

raizam:
Throw NoSuitableGraphicsDeviceException when GL.GetString(StringName.Version) returns null.
Currently throws a NullReferenceException.

This may happen in case of drivers problems.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
@@ -300,6 +300,10 @@ namespace Microsoft.Xna.Framework.Graphics
             try
             {
                 string version = GL.GetString(StringName.Version);
+
+                if (string.IsNullOrEmpty(version))
+                    throw new NoSuitableGraphicsDeviceException("Unable to retrieve OpenGL version");
+
                 string[] versionSplit = version.Split(' ');
                 if(versionSplit.Length > 2 && versionSplit[0].Equals("OpenGL") && versionSplit[1].Equals("ES"))
                 {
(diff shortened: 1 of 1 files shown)


cra0zy:
Better check if the context is valid instead of version.

raizam:
There are already checks done from CheckErrors() function: [GraphicsExtensions.cs#L920](https://github.com/MonoGame/MonoGame/blob/988bfb7be3b0818c6efa58cdbd2ddd642ce8bfb2/MonoGame.Framework/Graphics/GraphicsExtensions.cs#L920)
But looks like GL.GetError / GL.GetErrorCode() is not enough to validate the context in the case of the driver error we saw earlier on gitter.

prollin:
> So I guess these checks aren't done from the nuget package version. Not sure that's a good idea

This is because GetError() can cause a pipeline stall (like any other GetX() function) so you probably don't want this on release builds.

mrhelmut:
Yes, this is not a problem on release builds and anything that is a GL fail "should" correctly throw a NoSuitableGraphicsDeviceException. If there is a GL compatibility issue, this will most likely be triggered when loading the entry points during MonoGame initialization.

The two major incompatibility issues I've seen are: unsupported GL version or missing extension (triggering NoSuitableGraphicsDeviceException); can't build / attach shader program (which throws NoSuitableGraphicsDeviceException as well).

raizam:
Thank you all for your insights, I don't know much about these stuff.

The purpose of this pull request is to provide an error message when `GL.GetString(StringName.Version);` returns null. This happened to @MarkuBu on gitter, it was just a NVidia a driver issue on linux, but troubleshooting wasn't easy just with a NullReferenceException.

Also something to mention: He compiled from the sources, and added a breakpoint to that line, hence he was running Monogame in debug mode. The fact he could hit that line means previous calls of CheckGLErrors() after context creation didn't return an error.

tomspilman:
@cra0zy @dellis1972 @KonajuGames - Are the changes here a good one or not?

KonajuGames:
These changes are good.  No idea why a driver would return no version, but with OpenGL drivers anything is possible.

Merging.

cra0zy:
> These changes are good. No idea why a driver would return no version, but with OpenGL drivers anything is possible.

Well these changes are... maybe good if the driver has bad implementation? Anyway the problem @raizam had was that his driver failed to create gl context, and we are never checking if the gl context is valid, and in case the gl context is invalid the call to get gl version would return null.

In short there is still a big whole in the code when it comes to checking what the actual error is... and now when we fail to create context we will get invalid gl version.

PR #5771, opened 2017-06-13
Add PlatformDispose partial methods to States

KonajuGames:
Added a partial PlatformDispose method to DepthStencilState, RasterizerState and SamplerState to make them consistent with the recent changes to BlendState.


tomspilman:
Thanks @KonajuGames !

PR #5774, opened 2017-06-15
Fix enum that doesn't exist in OpenGL ES 2.0

### LINKED ISSUE #5773: Can't create texture in iOS
Hi,

in iOS (tested in simulator and ipod 6 10.3.2) creating a texture with i.e.

`var tex=new Texture2D (GraphicsDevice,32,32)
`

generates 

`Microsoft.Xna.Framework.Graphics.MonoGameGLException: GL.GetError() returned InvalidEnum`

at

`GL.TexParameter(TextureTarget.Texture2D, (TextureParameterName)0x813C, 0);
`

Has anybody found the same problem since the #5614 update?

KonajuGames:
GL_TEXTURE_BASE_LEVEL is not defined for OpenGL ES 2.0, so don't try to use it. A value of zero is assumed to be the default.

Fixes #5773


KonajuGames:
This is a fix for iOS and Android, so I'm going to merge it.

PR #5779, opened 2017-06-16
Add GraphicsDebug class for retrieving DirectX debug messages

hach-que:
This adds an initial GraphicsDebug class, which is surfaced as a `GraphicsDebug` property on the GraphicsDevice.  For debug builds, it allows developers to retrieve DirectX debugging messages directly in their game.

For a very long time now, Protobuild has by default generated projects with "Enable unmanaged code debugging" turned on, so that when you run and debug MonoGame games in Visual Studio, you can see the DirectX log messages in the Visual Studio output window.  These messages are essential to determining things like why a shader doesn't work, and for knowing when you're passing invalid parameters to APIs. 

However, this is a trade-off: when you enable unmanaged code debugging, Edit and Continue no longer works, which is a very useful part of the Visual Studio debugging experience. If you turn off unmanaged code debugging, you can't see the DirectX error messages.

By surfacing this API, we can turn unmanaged code debugging off by default in Protobuild, and developers can retrieve the DirectX messages directly in their game.  This allows Edit and Continue to work, while still being able to see critical DirectX warnings and errors about API usage.

This API only works in debug builds (because only in debug builds does MonoGame create a DirectX device with debug layers), and only works on DirectX 11 and higher (and I believe Windows 8 and later).

In future, I want to add APIs to this call that allow setting names on buffers and textures, so that these names (annotations as DirectX calls them) appear in graphics profiling and debugging tools.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDebug.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDebug.DirectX.cs
@@ -8,6 +8,7 @@ namespace Microsoft.Xna.Framework.Graphics
         private readonly GraphicsDevice _device;
         private readonly InfoQueue _infoQueue;
         private readonly Queue<GraphicsDebugMessage> _cachedMessages;
+        private bool _hasPushedFilters = false;
 
         public GraphicsDebug(GraphicsDevice device)
         {
@@ -15,8 +16,12 @@ namespace Microsoft.Xna.Framework.Graphics
             _infoQueue = _device._d3dDevice.QueryInterfaceOrNull<InfoQueue>();
             _cachedMessages = new Queue<GraphicsDebugMessage>();
 
-            _infoQueue.PushEmptyRetrievalFilter();
-            _infoQueue.PushEmptyStorageFilter();
+            if (_infoQueue != null)
+            {
+                _infoQueue.PushEmptyRetrievalFilter();
+                _infoQueue.PushEmptyStorageFilter();
+                _hasPushedFilters = true;
+            }
         }
 
         private bool PlatformTryDequeueMessage(out GraphicsDebugMessage message)
@@ -25,6 +30,13 @@ namespace Microsoft.Xna.Framework.Graphics
             {
                 message = null;
                 return false;
+            }
+
+            if (!_hasPushedFilters)
+            {
+                _infoQueue.PushEmptyRetrievalFilter();
+                _infoQueue.PushEmptyStorageFilter();
+                _hasPushedFilters = true;
             }
 
             if (_cachedMessages.Count > 0)
--- a/MonoGame.Framework/Graphics/GraphicsDebug.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDebug.DirectX.cs
@@ -1,0 +1,66 @@
+﻿using SharpDX.Direct3D11;
+using System.Collections.Generic;
+
+namespace Microsoft.Xna.Framework.Graphics
+{
+    public partial class GraphicsDebug
+    {
+        private readonly GraphicsDevice _device;
+        private readonly InfoQueue _infoQueue;
+        private readonly Queue<GraphicsDebugMessage> _cachedMessages;
+
+        public GraphicsDebug(GraphicsDevice device)
+        {
+            _device = device;
+            _infoQueue = _device._d3dDevice.QueryInterfaceOrNull<InfoQueue>();
+            _cachedMessages = new Queue<GraphicsDebugMessage>();
+
+            _infoQueue.PushEmptyRetrievalFilter();
+            _infoQueue.PushEmptyStorageFilter();
+        }
+
+        private bool PlatformTryDequeueMessage(out GraphicsDebugMessage message)
+        {
+            if (_infoQueue == null)
+            {
+                message = null;
+                return false;
+            }
+
+            if (_cachedMessages.Count > 0)
+            {
+                message = _cachedMessages.Dequeue();
+                return true;
+            }
+
+            if (_infoQueue.NumStoredMessagesAllowedByRetrievalFilter > 0)
+            {
+                // Grab all current messages and put them in the cached messages queue.
+                for (var i = 0; i < _infoQueue.NumStoredMessagesAllowedByRetrievalFilter; i++)
+                {
+                    var dxMessage = _infoQueue.GetMessage(i);
+                    _cachedMessages.Enqueue(new GraphicsDebugMessage
+                    {
+                        Message = dxMessage.Description,
+                        Id = (int)dxMessage.Id,
+                        IdName = dxMessage.Id.ToString(),
+                        Severity = dxMessage.Severity.ToString(),
+                        Category = dxMessage.Category.ToString()
+                    });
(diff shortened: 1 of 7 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -175,6 +175,10 @@ public GraphicsAdapter Adapter
         /// The metrics are reset every frame after draw within <see cref="GraphicsDevice.Present"/>. 
         /// </summary>
         public GraphicsMetrics Metrics { get { return _graphicsMetrics; } set { _graphicsMetrics = value; } }
+
+        internal GraphicsDebug _graphicsDebug;
```
Jjagg: This can be private.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -175,6 +175,10 @@ public GraphicsAdapter Adapter
...
+
+        public GraphicsDebug GraphicsDebug { get { return _graphicsDebug; } set { _graphicsDebug = value; } }
```
Jjagg: Can you document this?

MonoGame.Framework/Graphics/GraphicsDebug.DirectX.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            _device = device;
+            _infoQueue = _device._d3dDevice.QueryInterfaceOrNull<InfoQueue>();
+            _cachedMessages = new Queue<GraphicsDebugMessage>();
+
+            _infoQueue.PushEmptyRetrievalFilter();
+            _infoQueue.PushEmptyStorageFilter();
+        }
+
+        public bool TryDequeueMessage(out GraphicsDebugMessage message)
```
Jjagg: Maybe define this function in the GraphicsDebug class and make this `PlatformTryDequeueMessage` like we do for other methods with platform specific features. Or include a file with a default implementation of this function that always returns false and include it for other platforms. That way this won't break compilation when users switch targets.

hach-que:
@KonajuGames @tomspilman Any chance of this getting merged?

KonajuGames:
@hach-que Could you add the standard MonoGame header block to each new file?  Once that is done, I'm happy to merge it.

hach-que:
Done.

KonajuGames:
Thanks.  Just wait for the build and tests to run.

KonajuGames:
Merging. Thanks @hach-que

MonoGame.Framework/Graphics/GraphicsDebug.DirectX.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        private readonly GraphicsDevice _device;
+        private readonly InfoQueue _infoQueue;
+        private readonly Queue<GraphicsDebugMessage> _cachedMessages;
+        private bool _hasPushedFilters = false;
+
+        public GraphicsDebug(GraphicsDevice device)
+        {
+            _device = device;
+            _infoQueue = _device._d3dDevice.QueryInterfaceOrNull<InfoQueue>();
```
tomspilman: Doesn't this need to be `Disposed` on device destruction?  Will it otherwise keep the graphics device from being release?

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -175,6 +175,13 @@ public GraphicsAdapter Adapter
         /// The metrics are reset every frame after draw within <see cref="GraphicsDevice.Present"/>. 
         /// </summary>
         public GraphicsMetrics Metrics { get { return _graphicsMetrics; } set { _graphicsMetrics = value; } }
+
+        private GraphicsDebug _graphicsDebug;
+
+        /// <summary>
+        /// Access debugging APIs for the graphics subsystem.
+        /// </summary>
+        public GraphicsDebug GraphicsDebug { get { return _graphicsDebug; } set { _graphicsDebug = value; } }
```
tomspilman: Why is there a public setter here?  Who and why would someone replace this if it is created by the `GraphicsDevice` itself?

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -671,6 +671,9 @@ protected virtual void CreateDeviceResources()
 
             // Get Direct3D 11.1 context
             _d3dContext = _d3dDevice.ImmediateContext.QueryInterface<SharpDX.Direct3D11.DeviceContext>();
+            
+            // Create a new instance of GraphicsDebug because we support it on Windows platforms.
+            _graphicsDebug = new GraphicsDebug(this);
```
tomspilman: This should be moved to the common non-platform specific code if we really plan to support it across all platforms.  It is better than the developer having to check for `null` on `GraphicsDebug`.

tomspilman:
So what other platforms can this messaging feature work on?  I am aware only of 2 really... DirectX and Vulkan.  This is all part of debug spew in the VS output window on most console platforms.

MonoGame.Framework/Graphics/GraphicsDebug.DirectX.cs
hach-que: I don't believe so, we only have a managed handle to the graphics device here (unless the InfoQueue will hold the device open, but I haven't seen that in my own usage of the feature).

MonoGame.Framework/Graphics/GraphicsDevice.cs
hach-que: I just copied what GraphicsMetrics did for consistency - is there a reason why GraphicsMetrics has a public setter?

hach-que:
You only get the debug spew in VS if you enabled unmanaged debugging, which as mentioned in the PR prevents Edit and Continue from working.

Quick Googling suggests that you can debug messages in OpenGL as well: https://www.khronos.org/opengl/wiki/Debug_Output, though I haven't tried it.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
hach-que: Yeah that should probably be moved

MonoGame.Framework/Graphics/GraphicsDebug.DirectX.cs
KonajuGames: This is something I missed.  Interfaces obtained from the D3D device typically need to be disposed.

hach-que: Ah okay, I figured that QueryInterfaceOrNull was just some sort of casting-to-interface magic, not handle acquisition.

MonoGame.Framework/Graphics/GraphicsDevice.cs
tomspilman: > is there a reason why GraphicsMetrics has a public setter?

We wanted the user to be able to copy the metric data, reset it, then restore the old copy later.  This would let you measure metrics of individual subsystems while not loosing the metrics before the system ran.

PR #5780, opened 2017-06-16
Check if framebuffer is complete before calling GL.Clear under OpenGL

hach-que:
This fixes an issue where on macOS the GL.Clear called immediately after the GraphicsDevice is created throws a FrameBufferExt error, because the backbuffer isn't yet ready for writing.  Adding this check prevents GL.Clear from being called on an incomplete framebuffer.

I checked during game execution, and once the game is rendering this check passes and GL.Clear is called as expected.


KonajuGames:
Thanks @hach-que. Merging.

PR #5785, opened 2017-06-20
Fix for android 7 nougat for black screen when split-screen is activated

roponator:
The new split-screen feature in Android 7 caused black screen when activated.

The reason for it was that monogame activity was restarted, so just adding the layoutChange orientation handler fixes this.


mrhelmut:
This is relevant and can be merged.

@KonajuGames @tomspilman

KonajuGames:
Merging. Thanks @roponator

PR #5786, opened 2017-06-21
Allow DirectX debug layers to be enabled via environment variable

hach-que:
When using a release build of MonoGame (as is the case whenever using a prebuilt binary), there's no way of enabling the DirectX debug layers, even on systems with those debug layers installed.  To support this scenario, this updates device creation so that if the MONOGAME_ENABLE_DIRECTX_DEBUGGING environment variable is set to "true", then MonoGame will attempt to enable the debug layers if they are available.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -607,8 +607,20 @@ namespace Microsoft.Xna.Framework.Graphics
             // Windows requires BGRA support out of DX.
             var creationFlags = SharpDX.Direct3D11.DeviceCreationFlags.BgraSupport;
 #if DEBUG
-            creationFlags |= SharpDX.Direct3D11.DeviceCreationFlags.Debug;
+            var enableDebugLayers = true;
+#else 
+            var enableDebugLayers = false;
 #endif
+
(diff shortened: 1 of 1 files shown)


KonajuGames:
Thanks @hach-que. Merging.

tomspilman:
I think we should refactor this.

First it doesn't seem to be a good idea to make this something that end users can easily enable on a released game.

Second we already have a pattern to follow for this from XNA `static bool GraphicsAdapter.UseReferenceDevice`.

IMO a better solution here is adding `static bool GraphicsAdapter.DebugLayer`.  It can default to `true` on debug builds and `false` on release... still gives the developer full control in a release build of MonoGame.  Also as a bonus this technique works on platforms where you cannot just set an environment variable like consoles.

Also note there is more than one DX device creation function in `GraphicsDevice.DirectX.cs`.  One is for UWP/Metro and the other is for desktop.  This PR only changed the desktop creation code.

hach-que:
The reason I did it like this is I wanted to be able to easily turn on this option for games hosted in the editor by launching the game hosting process with the environment variable set, without developers having to explicitly add support for it in release builds.

I did look at trying to use a (albeit non-static) property to change the behaviour, but because the code happens in the constructor of the graphics device, that wasn't practical.

I'll see if I can wrangle up a version that uses a static property today though.

PR #5791, opened 2017-06-23
Replace environment variable with UseDebugLayers static property

hach-que:
See feedback in #5786 after it was merged.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
@@ -263,8 +263,20 @@ namespace Microsoft.Xna.Framework.Graphics
             // Windows requires BGRA support out of DX.
             var creationFlags = SharpDX.Direct3D11.DeviceCreationFlags.BgraSupport;
 #if DEBUG
-            creationFlags |= SharpDX.Direct3D11.DeviceCreationFlags.Debug;
+            var enableDebugLayers = true;
+#else 
+            var enableDebugLayers = false;
 #endif
+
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -281,19 +293,16 @@ protected virtual void CreateDeviceResources()
...
+        
             try 
             {
-#endif
                 // Create the Direct3D device.
                 using (var defaultDevice = new SharpDX.Direct3D11.Device(driverType, creationFlags, featureLevels.ToArray()))
                     _d3dDevice = defaultDevice.QueryInterface<SharpDX.Direct3D11.Device1>();
 
                 // Necessary to enable video playback
                 var multithread = _d3dDevice.QueryInterface<SharpDX.Direct3D.DeviceMultithread>();
```
hach-que: Does anyone know why this code is inside the try block?  I didn't add it, but it seems like it should be outside the try/catch in case the device creation fails the first time.

hach-que:
Can this please be reviewed / merged?

tomspilman:
Sorry... been stuck in console world the last few days.

MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs
tomspilman: Yeah this seems bad to have here... it would be fine to move it outside the try/catch.

tomspilman:
Thanks @hach-que ... merging this.  If you want to fix that video code within the `try/catch` just do it in a new PR.

PR #5792, opened 2017-06-23
Refactor GraphicsDebug according to feedback

hach-que:
See #5779. @tomspilman

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDebug.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDebug.cs
@@ -2,10 +2,35 @@
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
+using System;
+
 namespace Microsoft.Xna.Framework.Graphics
 {
-    public partial class GraphicsDebug
+    public partial class GraphicsDebug : IDisposable
     {
+        private readonly GraphicsDevice _device;
+        private bool _isDisposed = false;
+
+        /// <summary>
+        /// Constructs a new instance of <see cref="GraphicsDebug"/>, which provides debugging APIs
+        /// for the underlying graphics hardware.
+        /// </summary>
+        /// <param name="device">The associated graphics device.</param>
+        public GraphicsDebug(GraphicsDevice device)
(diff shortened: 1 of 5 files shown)


hach-que:
Can this please be reviewed / merged?

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -181,7 +181,7 @@ public GraphicsAdapter Adapter
         /// <summary>
         /// Access debugging APIs for the graphics subsystem.
         /// </summary>
-        public GraphicsDebug GraphicsDebug { get { return _graphicsDebug; } set { _graphicsDebug = value; } }
+        public GraphicsDebug GraphicsDebug { get { return _graphicsDebug; } }
```
tomspilman: We should add either to these docs or somewhere that this feature only works on some platforms.  Might even call out DX and GL platforms specifically since that is all that is supported.

MonoGame.Framework/Graphics/GraphicsDebug.cs
```diff
@@ -2,10 +2,35 @@
...
     {
+        private readonly GraphicsDevice _device;
+        private bool _isDisposed = false;
+
+        /// <summary>
+        /// Constructs a new instance of <see cref="GraphicsDebug"/>, which provides debugging APIs
+        /// for the underlying graphics hardware.
+        /// </summary>
+        /// <param name="device">The associated graphics device.</param>
+        public GraphicsDebug(GraphicsDevice device)
```
tomspilman: Should this constructor be public at all?  Since it is only created within `GraphicsDevice`, it could be `internal` keeping users from thinking they have to allocate one themselves.  Also in that case I wouldn't even document it as it is a private implementation detail.

KonajuGames:
@hach-que Can you address the two issues @tomspilman raised above?  Then I'll merge.

SimonDarksideJ:
Any update @hach-que so we can get this merged?

hach-que:
I haven't been using MonoGame for several years at this point, so I won't be revisiting this pull request.

PR #5793, opened 2017-06-23
Support GraphicsDebug in OpenGL

hach-que:
This adds support for retrieving graphics debug messages on OpenGL platforms using the new GraphicsDebug API.  This PR depends on both #5791 and #5792 being merged first.

There's some changes in OpenGL platform behaviour with this PR, and one breaking change as outlined below:

* The OpenGL implementation already captured messages on Debug builds of MonoGame, and emitted them using `System.Diagnostics.Debug`.  I updated the enablement of message capturing to align with DirectX; that is, enabled on Debug builds with the option to be enabled on Release builds by setting `UseDebugLayers` to true. This necessitated moving some code out of `GL`'s static constructor into the PlatformSetup for OpenGL.
* The OpenGL implementation appeared to call a static `OnError` callback in the `GL` class in response to messages being captured.  I've changed the code so this is no longer called, and added an `[Obsolete(..., true)]` attribute which will cause usages of this to emit a compilation error.  It wasn't used in MonoGame itself, and the nature of it's placement in the `GL` class makes it entirely platform specific, so I chose to forcefully deprecate it in favour of `GraphicsDebug`.  This can be changed to a soft deprecation (making the code still call it and changing the Obsolete to a warning) if desired.
* `GraphicsDebugMessage` got some new properties that match the data that OpenGL provides in it's debug messages, since this differs from the kind of information that DirectX provides.  I believe it's better to provide the data in fields that match the underlying subsystem, rather than trying to map things across and losing accuracy in the messages reported.

To test that this works, I ran the tests and set a breakpoint on the message capture function:

image

It's not really possible to write an automated test, since different systems may emit different debug messages or not emit messages at all.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -628,8 +628,11 @@
     <Compile Include="Graphics\GraphicsDebug.DirectX.cs">
       <Services>DirectXGraphics</Services>
     </Compile>
+    <Compile Include="Graphics\GraphicsDebug.OpenGL.cs">
+      <Services>OpenGLGraphics</Services>
+    </Compile>
     <Compile Include="Graphics\GraphicsDebug.Default.cs">
-      <Services>OpenGLGraphics,WebGraphics,ANGLEGraphics</Services>
+      <Services>WebGraphics,ANGLEGraphics</Services>
     </Compile>
     <Compile Include="Graphics\GraphicsDebugMessage.cs" />
     <Compile Include="Graphics\GraphicsDevice.cs" />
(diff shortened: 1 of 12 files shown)


hach-que:
I've just #ifdef'd around the new code to only apply to DesktopGL (Windows/Linux) for now, until OpenTK is removed from other platforms.  Once OpenTK is gone from those platforms, we should just be able to remove the #ifdef's and everything will work on remaining platforms.  It didn't seem worth it to add support for the OpenTK APIs for the sake of this PR working on macOS right now, since #5410 will kill them off anyway.

KonajuGames:
Waiting on response to feedback to #5792/

SimonDarksideJ:
Pending response, is this still valid @hach-que ?

hach-que:
I haven't been using MonoGame for several years at this point, so I won't be revisiting this pull request.

SimonDarksideJ:
Sorry to hear that @hach-que , I was always an avid supporter/followedrof your work.
Hope we can entice you back some day.

PR #5795, opened 2017-06-26
Fix for OpenGL texture construction not working from a thread

mrhelmut:
Hello there,

This is a fix for a problem introduced in #5614 which prevented ```Texture2D``` to be loaded from a thread on OpenGL platforms.

```PlatformConstruct``` needs ```Threading.BlockOnUIThread``` to make sure that no GL operations are done from other threads.

@KonajuGames @tomspilman


tomspilman:
Thanks @mrhelmut ... merging!

PR #5798, opened 2017-07-04
Allow disabling DirectX debug layers in Debug builds

alex-sherman:
GraphicsAdapter.UseDebugLayers has been added to allow enabling DirectX debug layers in Release builds, however I also which to be able to _disable_ them in Debug builds for performance reasons. This PR makes GraphicsAdapter.UseDebugLayers the sole authority on whether debug flags are sent to the d3dDevice in its creation flags. Here is my reasoning:

1. Having debug layers enabled causes draw calls to take orders of magnitude longer, it would be nice to be able to configure which aspects of the application we are debugging.
2. Additionally this makes the flag slightly less confusing, it will now always indicate whether debug layers are being used or not, rather than it being a combination of this flag and the DEBUG compiler flag.

My Visual Studio also complained about conflicting line endings and I included those changes which I can remove if you prefer.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.cs
@@ -82,12 +82,6 @@ namespace Microsoft.Xna.Framework.Graphics
         /// Used to request the graphics device should be created with debugging
         /// features enabled.
         /// </summary>
-        /// <remarks>
-        /// Debugging is enabled by default in Debug builds of MonoGame, while it
-        /// is disabled in Release builds. This option can only be used to enable
-        /// debugging in Release builds (i.e. setting it to false will not disable
-        /// debugging in Debug builds).
-        /// </remarks>
         public static bool UseDebugLayers { get; set; }
 
         public string Description { get; private set; }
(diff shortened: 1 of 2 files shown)


KonajuGames:
The only issue here is that Debug builds will no longer have the debug layer by default.  `GraphicsAdaptor.UseDebugLayers` should default to `true` for Debug builds and `false` otherwise.

alex-sherman:
If that's the functionality everyone wants I can update the PR.

However, my argument against that is it's very confusing to someone like myself who compiles open source projects with Debug and expects the results of the compilation to be relatively similar. A factor of 10 performance difference for a feature I didn't know I was enabling is very confusing. Compiling as DEBUG, in my opinion, should only degrade performance when necessary, like conditional compiling some timing calls or something. This is configurable outside of the compiler so it seems dirty to change behavior with a compiler flag.

Are a lot of people relying on this default behavior in DEBUG builds?

KonajuGames:
Fair enough, but I do find a factor of 10 performance difference to be quite beyond reasonable.  60fps compared to 6fps?  Is that what you are seeing?  If so I think there may be other issues at play with your system.  I will merge this anyway because it is a performance win for development, and the debug layer can be re-enabled when required.

PR #5799, opened 2017-07-05
Added new NuGet's for Platform type builds:

DDReaper:
* MonoGame.Framework.DX.nuspec
* MonoGame.Framework.OpenGL.nuspec

Tested locally, need to confirm with TeamCity build @tomspilman 
Does the build server need updating to identify the new NuSpec definitions?

### DIFF
--- a/NuGetPackages/MonoGame.Framework.DX.nuspec
+++ b/NuGetPackages/MonoGame.Framework.DX.nuspec
@@ -1,0 +1,84 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+    <metadata>
+        <id>MonoGame.Framework.DX</id>
+        <version>0.0.0.0</version>
+        <title>MonoGame.Framework.DX</title>
+        <authors>MonoGame Team</authors>
+        <owners>MonoGame Team</owners>
+        <licenseUrl>https://github.com/mono/MonoGame/blob/develop/LICENSE.txt</licenseUrl>
+        <projectUrl>http://monogame.net/</projectUrl>
+        <iconUrl>https://pbs.twimg.com/profile_images/487954549441691649/O3KsHAsb_400x400.png</iconUrl>
+        <requireLicenseAcceptance>false</requireLicenseAcceptance>
+        <description>MonoGame is an open source implementation of the Microsoft XNA 4.x Framework. The goal is to make it easy for XNA developers to create cross-platform games with extremely high code reuse.
+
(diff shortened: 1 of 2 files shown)


DDReaper:
@tomspilman  can we get this merged so we can check the packaged NuGet's, or make them available for download?

cra0zy:
Your formatting is all over the place (uses both tabs and spaces).

Why do we need these nugets?

DDReaper:
If you feel the formatting is important, I'll update that @cra0zy 

As for why, all will become clear with #5724 , discussion about how to fix portable projects with the new csproj formats and aim to provide a much simpler way for newcomers to adopt MG through NuGet.  It was preferred to have a single NuGet but that just wasn't possible, so the compromise is 2.

Agreed with @tomspilman that these will be additional for now with the possible view to reduce the NuGet's to just these two based on community feedback

cra0zy:
> If you feel the formatting is important, I'll update that @cra0zy

It's important.

cra0zy:
Your formatting is still off, somewhere you use 4 spaces intends, and somewhere 2 spaces intends. I would go with 2 spaces for xml based file types (but that is just my preference).

DDReaper:
Ran it through VSCode XML formatting, all uniform now.
Although, these files were based off the existing NuGet's, so the same issues must exist in those :S

DDReaper:
Any word yet on merging this @tomspilman @cra0zy ?

cra0zy:
@DDReaper I don't have any sort of merge rights, and even if I did I would only ever merge fixes.

Jjagg:
Even if it's temporary, having more redundant NuGets will likely confuse users. Maybe it can be clarified that this is an alternative to the platform specific NuGets in the description, or if this structure is fine we can just drop the others right now.

DDReaper:
As we can only currently test NuGet's properly until they are merged, not much we can do about that.
My aim was to use these to test as replacements as they solve issues with using the PCL version on platforms using project.json or the new .csproj format, as they changed how bait and switch worked.

In one discussion @tomspilman suggested adding these new platform one's and if they work out, then remove the others, if not remove these.

However, some users are reporting issues with using PCL's on only 2017 with the latest templates

KonajuGames:
I'm not familiar with the management of nuget packages and specs, but if the previous agreement was to get these in and see how they go, I'm happy to merge, and you can issue another PR later when a final decision is established.

DDReaper:
That I'm fine with @KonajuGames , as currently it's the only way to get the properly generated package from TeamCity until Tom can enable the direct download of the compiled Nuget's like the setup packages for PR's

KonajuGames:
Merged.

PR #5801, opened 2017-07-05
Fixed a crash when a newer mono is installed.

dellis1972:
If a user has a different version of mono installed on a Mac
the Pipeline tool crashes when building. This is because the
code page were are trying to use does not exist. So if we
cannot get that one we default to UTF8.

This PR also fixes another bug where if the build fails without
ANY items being processed the GUI crashes.

Also we upgrade to using msbuild for building the Mac platform.
This fixes a runtime issue when a different version of mono
is installed.


KonajuGames:
Merging.

PR #5805, opened 2017-07-10
Upgrade to SharpDX 4.0.1

vpenades:
This is an upgrade of Monogame from SharpDX 2.6.3 to 4.0.1  ... it relies on a fork that includes that SharpDX 4.0.1 assemblies, so the Build/module.xml has been modified to point to the fork

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/OutOfOrder/MonoKickstart.git
 [submodule "ThirdParty/Dependencies"]
 	path = ThirdParty/Dependencies
-	url = https://github.com/Mono-Game/MonoGame.Dependencies.git
+	url = https://github.com/vpenades/MonoGame.Dependencies.git
 [submodule "ThirdParty/NVorbis"]
 	path = ThirdParty/NVorbis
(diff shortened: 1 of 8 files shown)


DDReaper:
@vpenades we may be able to reopen - redo this PR now that #5809 is almost ready to merge.
Would then be good to get SharpDX upgraded to 4.x before 3.7 to pave the way for a DX12 upgrade after :D 

Maybe @Jjagg could help with that since he did the 3.x upgrade

vpenades:
@DDReaper Wow!, from what've I've seen in that PR, it took way more time to remove win8 than what's going to take to bump SharpDX to 4.0.1

Given that @Jjagg has way more experience bumping SharpDx, I would lend this to him. Specially because we have to deal with extracting the SharpDx from the nugets and placing them in the dependencies repo.

The only issue I recall from SharpDx 4.0.1 was the content builder failing to throw the expected exception when compiling an invalid shader, and returning null instead, which made the tests fail. It's something that can be easily sorted out.

Jjagg:
Yeah, that's fine :) I'll handle it.

PR #5810, opened 2017-07-11
Fixed back button problems in UWP

atomgott:
The back button on Windows 10 Mobile was not always reliable because GamePad.Back could be changed by the UI thread during a tick. Sometimes the button would do nothing because the game checked GamePadState before GamePad.Back was changed, which was reset to false in the end of the tick.

### DIFF
--- a/MonoGame.Framework/WindowsUniversal/UAPGamePlatform.cs
+++ b/MonoGame.Framework/WindowsUniversal/UAPGamePlatform.cs
@@ -31,6 +31,8 @@ namespace Microsoft.Xna.Framework
 
         internal static ApplicationExecutionState PreviousExecutionState { get; set; }
 
+        private static bool _backPressed = false;
+
         public UAPGamePlatform(Game game)
             : base(game)
         {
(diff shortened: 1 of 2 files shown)


nkast:
It sounds right to me.

@atomgott , 
While you are at it, is it possible to move the code inside `UAPGameWindow.ProcessWindowEvents()` ? https://github.com/MonoGame/MonoGame/blob/589bcf4820a9dd2f3810068cf6e893c9fae5e762/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs#L340-L364

This way we will have all Event Sync code nice and tide in one place.

Jjagg:
@atomgott Don't worry about the build failure, it's not because of this PR :)

atomgott:
Oh ok :)
I'm sorry about all the whitespace changes, I think that was Visual Studio trying to help

Jjagg:
You can see the diff without the whitespace changes by adding `?w=1` to the URL like so: https://github.com/MonoGame/MonoGame/pull/5810/files?w=1

nkast:
Great job @atomgott 

I just have to say that there's still a race condition although a very slim possibility of occurring.
thread#0 executes L314: GamePad.Back = _backPressed;
thread#1 executes L307: _backPressed = true;
thread#0 executes L315: _backPressed = false;
I thing it's easy to resolve by to calling `UpdateBackButton()` only when `(backPressed == true)`.

PR #5812, opened 2017-07-13
Add back the 'm' Legacy XNA content identifier

nkast:
The ''m' identifier for XNB platform was removed from ContentManager along with WP8.0 support.
'm' was originally the identifier for XNA WP7.0. 
MG still supports XNA content for backward compatibility and Testing.

Issue: http://community.monogame.net/t/asset-does-not-appear-to-be-a-valid-xnb-file-did-you-process-your-content-for-windows/9464

FYI, @DDReaper , @KonajuGames

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -35,8 +35,9 @@ namespace Microsoft.Xna.Framework.Content
 
         private static readonly List<char> targetPlatformIdentifiers = new List<char>()
         {
-            'w', // Windows (DirectX)
-            'x', // Xbox360
+            'w', // Windows (XNA & DirectX)
+            'x', // Xbox360 (XNA)
+            'm', // WindowsPhone7.0 (XNA)
             'i', // iOS
             'a', // Android
             'd', // DesktopGL


DDReaper:
I take it this doesn't cause any other ancillary content issues? Just that some older xna xnb's wouldn't be recognised, if i ready this right.

Also, was this there previously and removed for some reason? Given we don't support wp7 any more.

nkast:
'w','m','x' was there originally to load XNBs from XNA. 
'w' also marks MG for Windows/Desktop.
'm' was then used by MG for WindowsPhone8.0 as well.

We can either add back the 'm' since we already have 'x' there, mostly as a reminder not to use those for other platforms. Or drop the 'x' as well since we don't use it.

IMO MG shouldn't re-use the 'x','m' identifier for not-microsoft platforms.

theZMan:
WOuld be useful for some of the text from this thread to end up as a
comment too so that it doesn't happen in the future.

On Thu, Jul 13, 2017 at 5:53 AM, Nikos Kastellanos <notifications@github.com

tomspilman:
So on the same subject, but definitely for another PR...  it is dumb that formats that have no platform differences have different platform identifiers.

For instance and uncompressed Color texture has no differences across any of our platforms.  Yet if you diff the XNBs they all have this one byte difference for platform identifier.  And if you try to reuse that XNB on another platform it fails.

Seems like a better system would be to have a generic platform identifier by default... only use a specific platform identifier if the XNB actually contains platform specific formats.

KonajuGames:
> And if you try to reuse that XNB on another platform it fails.

We don't currently enforce that.  The only requirement we currently have is that the target platform identifier in the XNB exists in the set of target platform identifiers the `ContentManager` knows about.

tomspilman:
Thanks @nkast !

PR #5813, opened 2017-07-13
Add back the 'u' Legacy Ouya content identifier

nkast:


DDReaper:
Any particular reason to bring this back, given we dropped ouya a few releases ago?

KonajuGames:
This one is not needed.  No-one built content specifically for the Ouya.

nkast:
That PR was to be consisted and follows the same practice as we did with 'p','g','l'.
I don't know the exact reasoning behind it but I image it's to avoid potential conflicts.

PR #5814, opened 2017-07-13
[SDL] Fixes for initial window title and bounds

Jjagg:
Setting window title or position before the window was created had no
effect on DesktopGL. This fixes that issue. I was reminded of this by a forum post: http://community.monogame.net/t/cant-set-window-title-in-game1-constructor/9465

This is a small subset of the fixes from #5522 (which was closed to split up the changes). Some of the other changes made in that PR depend on #5585.

### DIFF
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -41,6 +41,8 @@ namespace Microsoft.Xna.Framework
         {
             get
             {
+                if (_handle == IntPtr.Zero)
+                    return new Point(_winx, _winy);
                 int x = 0, y = 0;
 
                 if (!IsFullScreen)
@@ -50,8 +52,13 @@ namespace Microsoft.Xna.Framework
             }
             set
             {
-                Sdl.Window.SetPosition(Handle, value.X, value.Y);
-                _wasMoved = true;
+                _winx = value.X;
+                _winy = value.Y;
+                if (Handle != IntPtr.Zero)
+                {
+                    Sdl.Window.SetPosition(Handle, value.X, value.Y);
+                    _wasMoved = true;
+                }
             }
         }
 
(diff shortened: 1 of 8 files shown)


Jjagg:
@cra0zy you probably want to review this ;)

MonoGame.Framework/SDL/SDLGameWindow.cs
```diff
@@ -28,18 +29,15 @@ public override bool AllowUserResizing
 
         public override Rectangle ClientBounds
         {
-            get
-            {
-                int x = 0, y = 0;
-                Sdl.Window.GetPosition(Handle, out x, out y);
-                return new Rectangle(x, y, _width, _height);
-            }
+            get { return new Rectangle(_winx, _winy, _width, _height); }
```
cra0zy: @Jjagg What if the user moves the window?

cra0zy:
Do we even need to do this? You shouldn't even be trying to set platform stuff before `Initialize` anyway...

Jjagg:
I think people commonly set the window title in their Game1 constructor. And it works in XNA and MG DX so for consistency and portability it's good if DesktopGL works the same.

cra0zy:
> I think people commonly set the window title in their Game1 constructor. And it works in XNA and MG DX so for consistency and portability it's good if DesktopGL works the same.

I don't think it's a good idea to guarantee any platform interaction behavior before `Initialize`, I do understand that compatibility is important, but for some things I feel like it really needs to be stopped.

Jjagg:
I disagree, so I guess we need other people to weigh in.

cra0zy:
Side note, if it is decided that this is needed, we need to add a proper `RecreateWindow` method instead of the temporary variables you created before init. This is to solve not being able to change some graphics related stuff under OpenGL.

Jjagg:
Yeah, I did some groundwork for that in #5585. Specifically we need to handle back buffer format, depth buffer size and MSAA.

cra0zy:
My point is that this change needs to be tied to implementing `RecreateWindow` method, because if we do that `Handle` will never be `IntPtr.Zero`, and code will be way more simplified than what you are currently doing.

gustavolsson:
I've been trying to understand why one of my projects kept crashing on macOS and it turned out to be related to setting Window.Title in the game constructor. Without having looked into it further, it seems to be some kind of race condition but it is only deterministic on a per-build basis. (Both projects set the title in constructor but only one of them crashes on macOS) The problematic project runs fine on Windows and macOS if I remove the line setting the title from the constructor.

If you do not intend to make this possible I suggest throwing an exception so that it cannot cause problems in the future :)

PR #5816, opened 2017-07-15
Implement Mouse.WindowHandle under Windows DirectX

nkast:
-Assign Mouse.Window through .WindowHandle

-Move assign of Mouse.Window/WindowHandle
from WinformsGamePlatform ->WinformsGameWindow.

-Make Mouse.Window private (from internal)

 

-Any Control can me assigned to Mouse.WindowHandle. (Mouse.Window Changed from Form -> Control)
Mouse is still not usable from MG embeded in custom WinForms apps since all event handling and mousestate still live in WinFormsGameWindow.

### DIFF
--- a/MonoGame.Framework/Input/Mouse.Windows.cs
+++ b/MonoGame.Framework/Input/Mouse.Windows.cs
@@ -14,11 +14,16 @@ namespace Microsoft.Xna.Framework.Input
         [return: MarshalAsAttribute(System.Runtime.InteropServices.UnmanagedType.Bool)]
         private static extern bool SetCursorPos(int X, int Y);
 
-        internal static Form Window;
+        private static Control _window;
 
-        private static IntPtr PlatformGetHandle()
+        private static IntPtr PlatformGetWindowHandle()
         {
-            return PrimaryWindow.Handle;
+            return _window.Handle;
+        }
+
+        private static void PlatformSetWindowHandle(IntPtr windowHandle)
+        {
+            _window = Control.FromHandle(windowHandle);
         }
 
         private static MouseState PlatformGetState(GameWindow window)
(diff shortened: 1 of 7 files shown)


cra0zy:
It's `WindowsDX`, not `DesktopDX` :P

MonoGame.Framework/Input/Mouse.Windows.cs
```diff
@@ -31,7 +36,7 @@ private static void PlatformSetPosition(int x, int y)
             PrimaryWindow.MouseState.X = x;
             PrimaryWindow.MouseState.Y = y;
             
-            var pt = Window.PointToScreen(new System.Drawing.Point(x, y));
+            var pt = _window.PointToScreen(new System.Drawing.Point(x, y));
```
cra0zy: Because `_window` is a static variable, this would cause problems with creating multiple instances of `Game`.

nkast: Yes, Mouse class is static too, you can only listen to one Window at the time.

nkast: Sure, we can make Mouse into a Singleton and all static members redirect to `Mouse.Instance`.
If the Dev wants he could create a new Mouse() and assign any window to it.

cra0zy: I was thinking more along lines allow `Mouse.SetPosition` to also take `GameWindow window` as an argument.

tomspilman: > I was thinking more along lines allow Mouse.SetPosition to also take GameWindow window as an argument.

I don't think we need to do this yet...  let's wait for someone to need it first.

Like I think I am the only person to ever use the multiple `GameWindow` feature.  I specifically added it for this project:

https://www.youtube.com/watch?v=yz6kkqHGEhM   #BuiltWithMonoGame

Specifically because we have games with 3 monitors and touch screens on two of them.

PR #5817, opened 2017-07-16
[SDL] Set default window position to system primary display

mrhelmut:
Hello,

This PR sets the default window position to the primary display set in the system settings.
This is the default XNA behavior.
It was previously setting the position to the display where the mouse cursor is (and this is still used as a fallback in case the primary display isn't found, though I doubt this would fail).

Note that there is no function in SDL to query which display is the primary one. SDL assumes that the display with coordinate 0,0 is the primary one.

Tested on Windows & MacOS.

@cra0zy @KonajuGames @dellis1972

### DIFF
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -102,11 +102,26 @@ namespace Microsoft.Xna.Framework
             _width = GraphicsDeviceManager.DefaultBackBufferWidth;
             _height = GraphicsDeviceManager.DefaultBackBufferHeight;
 
-            if (Sdl.Patch >= 4)
+            // look for the primary display
+            var rect = new Sdl.Rectangle();
+            var displayCount = Sdl.Display.GetNumVideoDisplays();
+            for (var i = 0; i < displayCount; i++)
             {
-                var display = GetMouseDisplay();
-                _winx = display.X + display.Width / 2;
-                _winy = display.Y + display.Height / 2;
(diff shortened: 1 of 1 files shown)


KonajuGames:
Is that documented behaviour of SDL (the primary display starts at 0, 0)?

mrhelmut:
Yes, in ```SDL_GetDisplayBounds``` documentation: https://wiki.libsdl.org/SDL_GetDisplayBounds

> Use this function to get the desktop area represented by a display, with the primary display located at 0,0.

KonajuGames:
Excellent.  Thanks @mrhelmut

cra0zy:
Apps on Linux should start on the display the mouse is on.

cra0zy:
Side note, on Wayland you don't even get control to set window position (similar to UWP on Windows on that regard).

mrhelmut:
@cra0zy Should we detect Linux and set the display to the mouse position then? Let me know if I can update this. Would this do it?

```
if (MonoGame.Utilities.CurrentPlatform.OS == MonoGame.Utilities.OS.Linux)
            {
                if (Sdl.Patch >= 4)
                {
                    var display = GetMouseDisplay();
                    _winx = display.X + display.Width / 2;
                    _winy = display.Y + display.Height / 2;
                }
            }
```

cra0zy:
Don't worry, I'll submit a fix in a few minutes, I just wish I got to comment on this before it got merged.

cra0zy:
Tada: #5818

PR #5819, opened 2017-07-16
UWP Templates use target version selected in wizard

atomgott:
When creating a Windows 10 Universal project, a window for selecting the target and minimum SDK versions pops up. Currently the selection is ignored. You can not create a new project if only version 10.0.15063.0 is installed because the template requires 10.0.14393.0 and it will fail with errors before the project has even been created.
I took the important lines from the Empty App template.


DDReaper:
Looks good to me

tomspilman:
Thanks @atomgott ... merging!

PR #5820, opened 2017-07-17
Do not allow the Pipeline tool to delete files outside the content folder

KonajuGames:
Linking files outside the project folder would include them as relative paths. Deleting these linked files, once confirmed, would delete entire folder trees outside the project folder. This unfortunately has destroyed projects of users.

This change will not show files with a path containing ".." in the delete confirmation dialog, as we will never be deleting them. Then it also checks for the file path containing ".." before deleting the file or folder.

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -866,9 +866,14 @@ namespace MonoGame.Tools.Pipeline
 
         public void Exclude(bool delete)
         {
-            if (delete && !View.ShowDeleteDialog(SelectedItems))
+            // We don't want to show a delete confirmation for any items outside the project folder
+            var filteredItems = new List<IProjectItem>(SelectedItems.Where(i => !i.OriginalPath.Contains("..")));
+
+            if (filteredItems.Count > 0 && delete && !View.ShowDeleteDialog(filteredItems))
                 return;
 
+            // Still need to pass all items to the Exclude action so it can remove them from the view.
+            // Filtering is done internally so it only deletes files in the project folder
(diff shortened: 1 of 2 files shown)


KonajuGames:
@cra0zy

Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -866,9 +866,14 @@ private List<string> GetDirectories(string folder)
 
         public void Exclude(bool delete)
         {
-            if (delete && !View.ShowDeleteDialog(SelectedItems))
+            // We don't want to show a delete confirmation for any items outside the project folder
+            var filteredItems = new List<IProjectItem>(SelectedItems.Where(i => !i.OriginalPath.Contains("..")));
+
+            if (filteredItems.Count > 0 && delete && !View.ShowDeleteDialog(filteredItems))
```
cra0zy: I think you have a bug in here, in case you only try to delete folder named `,,`, it will return here without excluding it from the Project View tree.

cra0zy: We might want to just not include a delete option for ".." folders all together.

KonajuGames: I did test that exact case and it works fine. It only prevents the delete operation from occurring, not the removal from the tree.

I kept it general because you could have the ".." selected along with other files/folders.

cra0zy: oh wait, I got confused, yea you are right...

cra0zy:
I am good with this change.

tomspilman:
Thanks @KonajuGames and @cra0zy !

PR #5821, opened 2017-07-18
Fix pre-emptive song finish in OggStreamer

KonajuGames:
The OggStreamer class was triggering the FinishedAction when the last buffers had been queued, but these buffers had not been played yet, resulting in the pre-emptive termination of the song.  This is now changed to trigger the FinishedAction when those last buffers have been consumed.


tomspilman:
Thanks... merging!

PR #5823, opened 2017-07-19
Monobranch

WingGT1:
Just a link fix, though apparently this will be automatically merged? Either way, still getting my feet under me with github, so this is good practice.


Jjagg:
Thanks for the contribution! :) 

> apparently this will be automatically merged

Not automatically, pull requests need to be reviewed first. But this is of course good to go like this.

KonajuGames:
Thanks @WingGT1. Merging.

Issue #5777, opened 2017-06-15
Getting "Method not found: 'Void SharpDX.XAudio2.XAudio2..ctor" on SoundEffect.Initialise - WindowsDX


DDReaper:
Texting on a clean Windows 10 machine and getting an odd SharpDX2 XAudio error but only on the WindowsDX platform.

UWP, Windows 8.1 compile fine.

On WindowsDX, the following error appears on calling SoundEffect.Initialise:

> 		Message	"Method not found: 'Void SharpDX.XAudio2.XAudio2..ctor(SharpDX.XAudio2.XAudio2Flags, SharpDX.XAudio2.ProcessorSpecifier)'."	string

MichaelDePiazzi:
This is probably related to issue #3240. From what I found, the required XAudio2 library is missing on clean machines: https://github.com/MonoGame/MonoGame/issues/3240#issuecomment-266639197

Jjagg:
This will be fixed by upgrading SharpDX (see #3794). Win10 ships with XAudio2.8 and 2.9 [says MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ee415802%28v=vs.85%29.aspx).
From the [SharpDX 3.0 changelog](https://github.com/sharpdx/SharpDX/releases/tag/v3.0.0):

> XInput and XAudio are now providing a backward compatible interface between the various versions. For Desktop, it means that the runtime will try to detect the correct versions for XInput (9.10, 1.3, 1.4) or XAudio (2.7 or 2.8). For Store Apps, the latest one will be used.

Issue #3240, opened 2014-12-08
Missing XINput.dll on clean machine


theZMan:
Tom and I installed MonoGame on a totally clean windows installation for the MVA video.

The desktop template doesn't work because any reference to the GamePad API throws this error
"xinput1_3.dll not found". We're not the only one to notice it was mentioned in this article too http://www.codemag.com/Article/1411081

XInput is not a part of the default DirectX install so it doesn't ship with windows. You need to run the DirectX web installer to fix it http://www.microsoft.com/en-us/download/details.aspx?id=35 (or install XNA which is why most of us never see the error). We did note that the windows store template does not suffer from this issue suggesting that the GamePad API is not wrapping XInput on windows store apps.

Solution: The MonoGame installer should use DirectX setup to make sure that XInput1_3.dll (and any other optional DX components) are installed. 

Or maybe SharpDX needs to link to a later version? Chucks block talks about a 1_4 and a 9_1 version on windows 8 http://blogs.msdn.com/b/chuckw/archive/2012/04/26/xinput-and-windows-8-consumer-preview.aspx

xoofx:
Sorry, completely forgot about this post. 

In Paradox, we are redistributing a minimal set of DirectX June 2010 runtime through the installer, but we haven't work for the client side redistribution (a client developing a game that would like to redistribute it as well).

I don't know the details about the difference between the API 9.1.0 and 1.3/1.4, and I agree that this is currently quite annoying that we can't easily target these APIs from a all-in-one assembly. I don't know if we can have an hybrid system where we could support 1.3 or 9.1.0 if it is installed, and 1.4 on Win8 assemblies. 

Unfortunately, the Windows 8 AppStore restrictions with DllImport is super annoying and forbid to achieve an ideal scenario where we could target all system DLL from the same SharpDX assembly (not allowing to LoadLibrary / GetProcAddress for system dlls but only allowing static DllImport and only allowing authorized imports, so if there are un-authorized imports - that you don't use, the App Certification will fail....etc.)

Anyway, feel free to log an issue on SharpDX though I really don't have any time to work on this for now... but if someone is interested to dig into this, he will be welcome!

polsys:
I investigated the DirectX Runtime installer option a bit...

The [runtime installer files](http://www.microsoft.com/en-us/download/details.aspx?id=8109) from Microsoft contain all the components since 2005. Of those packages, these files are required to install XInput 1.3 (about half a megabyte compressed):
```
[...]

PR #5825, opened 2017-07-19
Upgrade SharpDX to 3.1.1

### LINKED ISSUE #5742: SoundEffectInstance doesn't play
I created a new MonoGame.Windows (SharpDx) project,  loaded "blast_mono.wav" into a SounfEffect, and when trying to play the sound, I only get silence.

Same example in WindowsGL works correctly.

After some research, I found that SharpDX.XAudio2 [constructor](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffect.XAudio.cs#L119) was throwing an exception with the following message:

`HRESULT: [0x80040154], Module: [Unknown], ApiCode: [Unknown/Unknown], Message: Unregistered Class`

I tried to find a solution using the SharpDX currently used by monogame without success.

Then I followed the next steps:

I upgraded MonoGame.Framework.Windows to SharpDx 4.0
Upon compiling it gave 3 or 4 errors very easy to fix (RawBool instead of Bool and a missing argument in Reberb)
After compiling and running again, it gives more initialization exceptions, but this time in the [MasteringVoice](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffect.XAudio.cs#L134) , apparently, it doesn't like being passed a deviceID = 0
I removed the last three arguments, and left the constructor as `new MasteringVoice(Device);`

Compiled again, and it worked!, sound was playing!

So in order to fix the problem, it's a matter of upgrading to SharpDX 4.0  and adjusting the MasteringVoice constructor.

Btw, I tried removing deviceId argument from MasteringVolume with SharpDX 2.6.3 but it didn't work. Given that SharpDX 4.0 is a significant milestone and it seems there's no big hurdles upgrading monogame to SharpDX, I would ask to consider upgrading soon.

Jjagg:
This should resolve #5742, #3240 and #5777 because these newer SharpDX versions check for for multiple version of XInput and XAudio instead of depending on old versions that are no longer shipped with Windows 10. 

From [SharpDX 3.0 changelog](https://github.com/sharpdx/SharpDX/releases/tag/v3.0.0):

> XInput and XAudio are now providing a backward compatible interface between the various versions. For Desktop, it means that the runtime will try to detect the correct versions for XInput (9.10, 1.3, 1.4) or XAudio (2.7 or 2.8). For Store Apps, the latest one will be used.

The changes in this PR are exactly the same as needed to upgrade to SharpDX 4, so that will be pretty much painless after this :) 

I cleaned up a lot of preprocessor stuff because the API is more unified now that all Windows platforms use the same SharpDX version. That needed to happen anyway when we upgrade to SharpDX 4, so I thought I might as well include it in here.

The [change in 2MGFX](https://github.com/MonoGame/MonoGame/compare/develop...Jjagg:dx3.1.1?expand=1#diff-19f639315aec8ef8a47c2e29acf039f1R47) is because `SharpDX.D3DCompiler.ShaderBytecode.Compile` does not correctly set `CompilationResult.HasErrors`, but we can detect failure by checking if the returned `ByteCode` is null.

### DIFF
--- a/Build/Projects/2MGFXReferences.definition
+++ b/Build/Projects/2MGFXReferences.definition
@@ -4,10 +4,10 @@
   <Platform Type="Windows">
     <Binary
       Name="SharpDX"
-      Path="ThirdParty/Dependencies/SharpDX/SharpDX.dll" />
+      Path="ThirdParty/Dependencies/SharpDX/Desktop/SharpDX.dll" />
     <Binary
       Name="SharpDX.D3DCompiler"
-      Path="ThirdParty/Dependencies/SharpDX/SharpDX.D3DCompiler.dll" />
+      Path="ThirdParty/Dependencies/SharpDX/Desktop/SharpDX.D3DCompiler.dll" />
     <Binary
       Name="CppNet"
       Path="ThirdParty/Dependencies/CppNet/CppNet.dll" />
--- a/Build/Projects/2MGFXReferences.definition
+++ b/Build/Projects/2MGFXReferences.definition
@@ -4,10 +4,10 @@
   <Platform Type="Windows">
     <Binary
       Name="SharpDX"
-      Path="ThirdParty/Dependencies/SharpDX/Windows/SharpDX.dll" />
+      Path="ThirdParty/Dependencies/SharpDX/SharpDX.dll" />
     <Binary
       Name="SharpDX.D3DCompiler"
-      Path="ThirdParty/Dependencies/SharpDX/Windows/SharpDX.D3DCompiler.dll" />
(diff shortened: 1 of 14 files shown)


KonajuGames:
Thanks @Jjagg.  This is a good first step that will solve several issues.

vpenades:
@Jjagg yes!, I've just gave a try to MonoGame SDK 3.7.0.647 , and the audio works now!

DDReaper:
With this merged, should probably update the desktop NuGet to also use 3.1.1 now. Was downgraded to 2.6.3 due to issues initially.

PR #5829, opened 2017-07-20
OpenGL Mouse.SetCursor now works with alpha correctly

KonajuGames:
The previous implementation was saving the texture to a BMP in-memory, then loading that BMP into a SDL surface.  This process was losing the alpha channel.  It has been reimplemented to eliminate the BMP intermediate step and copy data direct from the Texture2D to the SDL surface.

Fixes #5822

PR #5831, opened 2017-07-21
Implement Mouse.SetCursor() for Windows

KonajuGames:
The `Mouse.SetCursor(MouseCursor)` and `MouseCursor.FromTexture2D(Texture2D, int, int)` methods were previously unimplemented for the Windows DirectX platform.


Jjagg:
I tested this and it's working great :) 
I noticed `PlatformSetCursor` is public, but looks like that's already addressed in #5829!

KonajuGames:
Yes, I noticed it was public when doing the earlier PR. Thanks for testing
this.

tomspilman:
Thanks... merging!

PR #5837, opened 2017-07-23
[DX] Dispose cached texture in WMS VideoPlayer

Jjagg:
Texture was not disposed when a new one was created. 
Also a minor code style edit.


KonajuGames:
Merging to fix the immediate issue. Thanks @Jjagg

PR #5841, opened 2017-07-26
[DesktopGL] Fix Joystick enum (fixes #5729)

### LINKED ISSUE #5729: SDL Joystick Hat Enum defines incorrect values
I have come across an issue where DPad/Hat switch values are not reported correctly in a DesktopGL project on both Windows and Linux when using the Joystick class. I believe I have tracked the issue down to incorrect values being defined in the Hat enum [here](https://github.com/MonoGame/MonoGame/blob/1c09add4a7407cf14690614d5ec36666d6c2a8ea/MonoGame.Framework/SDL/SDL2.cs#L634).

I think the correct values should be:
```csharp
[Flags]
public enum Hat : byte
{
    Centered = 0,
    Up = 1,
    Right = 2,
    Down = 4,
    Left = 8,
}
```

cra0zy:


tomspilman:
Thanks @cra0zy !

PR #5842, opened 2017-07-26
[Pipeline Tool] Fix double clicking not opening the file

cra0zy:
Double click was opening context menu instead of opening the file.


KonajuGames:
Thanks @cra0zy. Merging.

PR #5849, opened 2017-07-28
Improve shader error/warning parsing in Pipeline Tool

### LINKED ISSUE #5846: PipelineTool - Not showing error line when output filtering on
Hi !
I've noticed with the last build (as of 3.7.678) and maybe the previous one or 2, the line and position of the warnings/errors are no more available when using the filtered output.
I have to switch to "text mode" to find the line of the warning/error:
image

nkast:
The new shader compiler produce errors/warnings that include the columnEnd. 
ex.  (26,9-25)

The RegEx was updated to recognize 
single lines, ex: `(26)`
line & column, ex: `(26,9)`
line & column begin & column end, ex: `(26,9-25)`

The RegEx is more strict than before and will fail on invalid strings, ex: `(26,)`  `(,9)`. `(,,,,,)`

Fix: #5846 

effectwarning

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs
@@ -161,7 +161,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
             var errorsAndWarningArray = shaderErrorsAndWarnings.Split(new[] {"\n", "\r", Environment.NewLine},
                                                                       StringSplitOptions.RemoveEmptyEntries);
 
-            var errorOrWarning = new Regex(@"(.*)\(([0-9,]*)\)\s*:\s*(.*)", RegexOptions.Compiled);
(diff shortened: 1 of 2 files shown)


tomspilman:
> The new shader compiler produce errors/warnings that include the columnEnd.

I guess I don't know if how we have things setup if a older shader compiler could be used on some systems?  Would the new parsing still work if the output is from an older compiler?

nkast:
> Would the new parsing still work if the output is from an older compiler?

No problem with that. In some cases the new compiler still produce the old style of warnings with a line and a column. ex. (26,9)

Actually I was testing to see if it reports errors that span on multiple lines, something that would look something like (26,9-28,5) but what I got instead was the old (line,column). 
For testing that, I simply split the line with the warning into multiple lines
```
    float3 tst = new float4(1, 2,
                        3,
                        4);
```

tomspilman:
Thanks @nkast !

nkast:
I noticed that in the output view I am getting a message like this

```
C:/testVSOutput/Content/File.fx
C:/testVSOutput/Content/File.fx: C:\\testVSOutput\\Content\\File.fx(26,9): warning X3206: implicit truncation of vector type
```

This looks wrong. Maybe I missed something in the parsing, however I noticed the filename has double quotes. This comes from SharpDX CompilationResult.Message.

PR #5850, opened 2017-07-29
Improved disposal of OpenGL resources

KonajuGames:
GL resources (textures, buffers, effects, shaders, queries, framebuffers) are now queued by the GraphicsDevice and deleted after the Present on the next frame.  This gives them time to be released by any rendering operations and allows the GC to clean up the wrapper objects.
SDL GL contexts are treated separately since there won't be a GraphicsDevice around to clean them up.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsContext.SDL.cs
+++ b/MonoGame.Framework/Graphics/GraphicsContext.SDL.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using Microsoft.Xna.Framework.Graphics;
 
 namespace OpenGL
 {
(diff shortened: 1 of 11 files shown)


tomspilman:
This seems ok to me...  better than the `BlockOnUIThread` stuff I think.

@cra0zy @dellis1972 - Got time to review this a bit?

cra0zy:
:+1: from me

tomspilman:
Thanks @KonajuGames and @cra0zy !

Jjagg:
@KonajuGames This broke Android. `_programCache` is only assigned in DESKTOPGL, but used on other platforms too, causing a NullReferenceException in PlatformInitialize. Someone reported this on the forum: http://community.monogame.net/t/nullreferenceexception-during-bootup-of-brand-new-android-project-visual-studio-2017/9562
I think [assigning the program cache](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs#L269) can just be moved outside the `#if DESKTOPGL` block.

KonajuGames:
Ok. I'll fix it up tonight unless another PR is made for it before then
(it's the start of Monday here now).

KonajuGames:
PR #5877 submitted to fix the crash.

PR #5852, opened 2017-07-30
Resolve conflict between #5816 and #5831

nkast:
Resolve conflict between #5816 and #5831


nkast:
@KonajuGames , @tomspilman 
The build server fails because of this bug.

tomspilman:
Thanks @nkast !

PR #5853, opened 2017-07-30
[Pipeline Tool] Make launchdebugger a per project property (fixes #5768)

### LINKED ISSUE #5768: Pipeline.exe (MGCB.exe GUI tool) reports invalid option with /launchdebugger
From the [MGCB.exe documentation](http://www.monogame.net/documentation/?page=MGCB), the command line tool supports a `/launchdebugger` option parameter to allows debugging of custom pipeline. However, when I add this option in the `Content.mgcb`:

```
#----------------------------- Global Properties ----------------------------#

/outputDir:bin/$(Platform)
/intermediateDir:obj/$(Platform)
/platform:DesktopGL
/config:
/profile:Reach
/compress:False
/launchdebugger

[...]
```

The pipeline tool reports the following issue `Content.mgcb: Unknown option 'launchdebugger'` and refuses to load the content file.

Currently, I circumvented the problem by adding the option `/launchdebugger` to the `MonoGame.Content.Builder.targets` file, after the `/quiet` option, but it would be cool to support the option directly in the pipeline tool.

cra0zy:
~This will fail to compile on build bot because of #5852~

### DIFF
--- a/Tools/Pipeline/Common/IController.cs
+++ b/Tools/Pipeline/Common/IController.cs
@@ -41,11 +41,6 @@ namespace MonoGame.Tools.Pipeline
         bool ProjectBuilding { get; }
 
         /// <summary>
-        /// Passes /launchdebugger option when launching MGCB.
-        /// </summary>
-        bool LaunchDebugger { get; set; }
-
-        /// <summary>
         /// The view this controller is attached to.
         /// </summary>
         IView View { get; }
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -48,8 +48,6 @@ namespace MonoGame.Tools.Pipeline
             get { return _templateItems; }
         }
 
-        public bool LaunchDebugger { get; set; }
-
         public PipelineProject ProjectItem
         {
             get
@@ -393,8 +391,6 @@ namespace MonoGame.Tools.Pipeline
         public void Build(bool rebuild)
         {
             var commands = string.Format("/@:\"{0}\" {1}", _project.OriginalPath, rebuild ? "/rebuild" : string.Empty);
-            if (LaunchDebugger)
-                commands += " /launchdebugger";
             BuildCommand(commands);
         }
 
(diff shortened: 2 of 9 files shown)


KonajuGames:
Thanks @cra0zy. Merging.

tomspilman:
This was a bad change...  I plan to revert it.

cra0zy:
@tomspilman Why is it a bad change?

cra0zy:
On a related side note I was gonna make a suggestion to add an actual debug button... but no one is around and I felt to uninspired to wait for someone to notice the PR :\

tomspilman:
It is a bad change because it removed this feature...

image

We didn't need to remove it in order to resolve https://github.com/MonoGame/MonoGame/issues/5768.

cra0zy:
> It is a bad change because it removed this feature...

It's just moved under project properties.

cra0zy:
Also give me time till tomorrow, I'm gonna setup a PR that's better than just reverting this.

tomspilman:
> It's just moved under project properties.

It just shouldn't be there.  The feature was for debugging... not something you permanently enable in a content project then check into source control.

The right fix is for the pipeline tool to simply ignore the `/launchdebugger ` option and load the file as normal.

> I'm gonna setup a PR that's better than just reverting this.

I'll be looking for it tomorrow.  But if it still allows the Pipeline tool to save the `/launchdebugger` to the .mgcb file i will not approve it.

cra0zy:
> But if it still allows the Pipeline tool to save the /launchdebugger to the .mgcb file i will not approve it.

#5768

tomspilman:
His issue was with opening the file...  that we should fix.

It is arguable if we should preserve his hand modified .mgcb and include the `/launchdebugger` flag in saves from the pipeline tool.  I think it should not like any other non-standard hand editing that the pipeline tool doesn't preserve.

Still we still should never have the pipeline tool add the `/launchdebugger` flag itself to a .mgcb project file.   '/launchdebugger` was never intended to be part of the .mgcb file format... it was only meant to be used on the MGCB.exe command line for triggering debugging.

PR #5854, opened 2017-07-30
[Linux Installer] Hide unneeded output when adding new icon

cra0zy:


KonajuGames:
Thanks @cra0zy. Merging.

PR #5859, opened 2017-08-01
Fix crash on multi-editing bool values in Pipeline Tool

### LINKED ISSUE #5858: Pipeline crashes when setting a boolean processor parameter for 2+ items with mixed values
The MonoGame Pipeline (version 3.6.0.1625) reliably crashes when you select two or more items that have different values for the same processor parameter, then try to set that processor parameter for both simultaneously. It crashes with a NullReferenceException, but I can't get more information than that without digging into MonoGame (which I don't want to do right now).

**Steps to reproduce**
Make a new project.
Add 2 or more images.
Set *Color Key Enabled* to false for one (one is false and one is true).
Select both.
Try to set *Color Key Enabled* by clicking on it. The Pipeline should crash.

**Notes**
This bug exists for any boolean processor parameter.
image

cra0zy:
Fixes #5858.

### DIFF
--- a/Tools/Pipeline/Controls/PropertyCells/CellBool.cs
+++ b/Tools/Pipeline/Controls/PropertyCells/CellBool.cs
@@ -24,19 +24,20 @@ namespace MonoGame.Tools.Pipeline
 
             var checkbox = new CheckBox();
             checkbox.Tag = this;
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @cra0zy !

PR #5860, opened 2017-08-02
Added the x86 and x64 binaries to the DesktopGL Installation.

dellis1972:
The x86/x64 Binaries were missing for DesktopGL on Mac (and I assume Linux)

### DIFF
--- a/Installers/default.build
+++ b/Installers/default.build
@@ -80,6 +80,8 @@
       <mkdir dir="root/Library/Frameworks/MonoGame.framework/v${buildNumber}/Assemblies/MacOSX" />
       <mkdir dir="root/Library/Frameworks/MonoGame.framework/v${buildNumber}/Assemblies/Linux" />
       <mkdir dir="root/Library/Frameworks/MonoGame.framework/v${buildNumber}/Assemblies/DesktopGL" />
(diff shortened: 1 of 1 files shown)


Installers/default.build
```diff
@@ -173,6 +185,16 @@ fi
                 <include name="*.*"/>
          </fileset>
       </copy>
+      <copy todir="Linux/tmp_run/MonoGameSDK/Assemblies/DesktopGL/x86">
```
cra0zy: The line above (`<copy todir="Linux/tmp_run/MonoGameSDK/Assemblies/DesktopGL">`) already copies the native files for Linux installer.

DDReaper:
Can you double check if these are needed for the Mac NuGet as well, if they are not there already?

cra0zy:
> Can you double check if these are needed for the Mac NuGet as well, if they are not there already?

They are not needed for the Mac NuGet and I included them in the DesktopGL nuget. Also this PR does nothing for the NuGet...

KonajuGames:
What's the status of this?  Is it good to go?

DDReaper:
Looks good to me and brings it inline with the NuGet releases

cra0zy:
It's not good, Linux changes are incorrect as stated above.

dellis1972:
@KonajuGames I removed the bits @cra0zy as red to remove. This is good to go

KonajuGames:
Ok, I'm going to merge this now.

PR #5862, opened 2017-08-06
Allow switching vsync at runtime in DesktopGL

Jjagg:
The title says it all. Changing PresentInterval then calling ApplyChanges previously did not work. The value was only checked when a new GraphicsDevice was created.

I added a test to check if graphics state is cleared on reset in XNA. It's not, which matched current MG behavior, so no further changes required there.


KonajuGames:
Thanks @Jjagg. Merging.

PR #5863, opened 2017-08-06
Fix broken XNA tests because of horizontal scroll wheel

Jjagg:


KonajuGames:
Thanks @Jjagg. Merging.

PR #5864, opened 2017-08-11
Made Game class partial to allow for per-platform extensions

tomspilman:
This makes the `Game` class a partial class allowing for per-platform extensions.

A good followup would be to pull all the `#if` code out of `Game` and implement them in a platform specific partial class.

For example this feature is used on console platforms where some per-game features are exposed as partial members only when targeting that platform.


KonajuGames:
This is a good start.  Thanks @tomspilman

PR #5869, opened 2017-08-12
Fix OpenGL pixel alignment

### LINKED ISSUE #5865: TextureFormat Color16Bit does not render odd-width textures correctly (Cross Platform Desktop)
Create a Cross Platform Desktop Project.
Create a texture with an odd number pixels width.
Set TextureFormat in Monogame Pipeline Tool to Color16Bit.
Draw the texture in Monogame and it offsets to create a diagonal slanted texture.

The image in the example is 257x256 pixels.

image

Jjagg:
Fixes #5865. OpenGL packing/unpacking alignment was incorrect. This PR makes MG set the correct alignment when getting/setting data.

Note that alignment needed to be capped at 8 bytes, because OpenGL only allows an alignment of 1, 2, 4 and 8 bytes, but some MG formats have an alignment of 16 bytes.


tomspilman:
Thanks @Jjagg !

PR #5872, opened 2017-08-14
Cleanup SpriteFont

nkast:


nkast:
In the second commit I call MeasureString(string) and MeasureString(StringBuilder) instead of a direct call to MeasureString(ref CharacterSource).

Perhaps I should remove this commit for now and/or mark MeasureString() with `AggressiveInlining`.

MonoGame.Framework/Graphics/SpriteFont.cs
```diff
@@ -205,11 +200,10 @@ internal void MeasureString(ref CharacterSource text, out Vector2 size)
                     finalLineHeight = currentGlyph.Cropping.Height;
             }
 
-            size.X = width;
-            size.Y = offset.Y + finalLineHeight;
+            return new Vector2(width, offset.Y + finalLineHeight); 
 		}
 
-        internal struct CharacterSource 
+        private struct CharacterSource
```
tomspilman: @nkast - Do you think we can gain some performance by specializing this helper for `StringBuilder` vs `string` (to avoid the `if (string==null)` tests and/or using some fixed pointers?

tomspilman:
> Perhaps I should remove this commit for now and/or mark `MeasureString()` with AggressiveInlining.

I don't think `AggressiveInlining` is guaranteed in any way...  either we prefer the cleaner code here or keep manually inlining things.

One extra call per-`DrawString` isn't really all that bad IMO.  I doubt this change would even show up in a profile really.

Also note this PR is conflicted from your other font optimization PR merge.

MonoGame.Framework/Graphics/SpriteFont.cs
nkast: I couldn't find a way get to the internal arrays. The only alternatives I can think is an abstract/virtual `char this [int index]` -hopefully a virtual method it will perform better in this case- or inlining the methods as we have done in SpriteBatch.DrawString().

willmotil:
In c# strings ToString() implicitly casts to a stringbuilder as well as to a string. This is partially the reason i say that charactersource is simply pointless. Unless it really is magical, if it is id like to see some proof and explaination of how it's not just redundant and useless.

nkast:
@willmotil 
I am not sure how this will work with out generating garbage. Can you give an example?

willmotil:
Well think about it what does char source offer that stringbuilder doesn't ? indexing ?
you can index a stringbuilder. You can set a string to a stringbuilder from a methods parameter.
It offers some supposed magical work around to not generate garbage ? It does no such thing.
That's why i wrote that mgstringbuilder class because to make monogame generate garbage as it is,,, is easy. I only have to tostring some numeric variable change it and drawstring it over and over repeatedly.

I don't have a simplified example on hand but here is a snippet from some old tests. with a gif.

http://i936.photobucket.com/albums/ad207/xlightwavex/programing%20and%20concepts/textdemoII01_zpst2rr5abo.gif

        public void DrawTextWordWrap
            (
            SpriteFont tsf, StringBuilder text, Rectangle destination,
            float leftMargin, float rightMargin, float breakWordLetterLength,
            Color color, float rotation, Vector2 origin, Vector2 scale,float depth
            )
        {
            //unchecked{
            // i really should precalculate all the scaling in the glyphs
            //
            // precalculated values
            var lineHeight = (float)tsf.LineSpacing * scale.Y;
            var spacingX = tsf.Spacing;
            float bigWordSize = Math.Max(breakWordLetterLength, 2f);
            Vector2 offset = new Vector2(leftMargin - 1, leftMargin * .5f + 3f);
            var currentGlyph = defaultGlyph;
            var firstGlyphOfLine = false;
            Vector2 drawDestPos = new Vector2(destination.X, destination.Y);
            Rectangle drawDestRect = new Rectangle();
            Vector2 q = Vector2.Zero;
            if (rotation != 0)
            {
                q = new Vector2((float)(Math.Sin(rotation)), (float)(Math.Cos(rotation)));
            }
            // Allow for special commands
            int specialCommand = 0;

            word.NewWord(rightMargin);

            //DrawBasicSquare(spriteBatch, destination, 2, Color.Green, rotation, false);

            // begin looping thru text
            for (var i = 0; i < text.Length; i++)
            {

                var c = text[i];
                //var c = char.ConvertToUtf32(text[i], text[i +1]);

                if (c == '\r')
                    continue;
                if (c == '\n')
                {
                    offset.Y += lineHeight; //Math.Max(currentGlyph.Cropping.Height, lineHeight);
                    firstGlyphOfLine = true;
                    continue;
                }

                // forum style commands
                if (c == '[')
                {
                    // Disallow a bogus command to error out a seperate command
                    // This is probably the best we can hope for vs user error
                    // A little bit-math to trim it down.
                    if (specialCommand < 0) { specialCommand = 0; }
                    if (text[i + 1] == '/')
                    {
                        // remove command bit flag, flip off if on
                        var comchar = text[i + 2];
                        //
                        if (comchar == 'b')
                            if((specialCommand & 4) > 0)
                                specialCommand = specialCommand ^ 4; // specialCommand -= 4; 
                        if (comchar == 'i')
                            if ((specialCommand & 2) > 0)
                                specialCommand = specialCommand ^ 2; // specialCommand -= 2;
                        if (comchar == 'c')
                            if ((specialCommand & 1) > 0)
                                specialCommand = specialCommand ^ 1; // specialCommand -= 1;
                        i += 3;
                        continue;
                    }
                    else
                    {
                        if (text[i + 2] == ']')
                        {
                            // set command bit flip on if off
                            var comchar = text[i + 1];
                            //
                            if (comchar == 'b')
                                if((specialCommand & 4) == 0)
                                    specialCommand = specialCommand | 4;//specialCommand += 4;
                            if (comchar == 'i')
                                if ((specialCommand & 2) == 0)
                                    specialCommand = specialCommand | 2;//specialCommand += 2;
                            if (comchar == 'c')
                                if ((specialCommand & 1) == 0)
                                    specialCommand = specialCommand | 1;//specialCommand += 1;
                            i += 2;
                            continue;
                        }
                    }
                }

                // Newlines clear this, so it must procced our newline offset check.
                var lastCharLinearPosX = offset.X;

                if (_glyphs.ContainsKey(c))
                    currentGlyph = _glyphs[c];
                else
                    currentGlyph = defaultGlyph;

                if (firstGlyphOfLine == false)
                    offset.X += spacingX + currentGlyph.LeftSideBearing * scale.X;
                else
                    offset.X = leftMargin;

                firstGlyphOfLine = false;

                // We work upon the idea that our buffer will be able to know when the next character will be a word break.
                bool isNextCharAWordSeperator = true;
                if (i + 1 < text.Length)
                {
                    var next = (char)text[(i + 1)];
                    if (!char.IsSeparator(next) && !char.IsWhiteSpace(next) && next != '\n')
                    {
                        isNextCharAWordSeperator = false;
                    }
                }

                Vector2 localDrawXY = new Vector2(currentGlyph.Cropping.X, currentGlyph.Cropping.Y) * scale + offset + drawDestPos;
                Vector2 localDrawWH = new Vector2(currentGlyph.BoundsInTexture.Width * scale.X, currentGlyph.BoundsInTexture.Height * scale.Y);
                float finaloffsetX = (currentGlyph.Width + currentGlyph.RightSideBearing) * scale.X + offset.X;
                float stride = finaloffsetX - lastCharLinearPosX;
                offset.X = finaloffsetX;

                drawDestRect = new Rectangle
                (
                    (int)(localDrawXY.X),
                    (int)(localDrawXY.Y),
                    (int)(localDrawWH.X),
                    (int)(localDrawWH.Y)
                );

                if (rotation != 0)
                {
                    Vector2 r = localDrawXY - drawDestPos - origin;
                    drawDestRect.X = (int)(r.X * q.Y - r.Y * q.X + drawDestPos.X + origin.X);
                    drawDestRect.Y = (int)(r.X * q.X + r.Y * q.Y + drawDestPos.Y + origin.Y);
                }

                //____________________________________________________________
                // Word buffer operations
                //____________________________________________________________

                // records current word draw rectangles the text char index and sums strides
                word.SetDrawnLetterGlyph(drawDestRect, currentGlyph.BoundsInTexture, stride, i, specialCommand);

                // inbound outbound calculations

                // This flags a special case were a single word is bigger then a entire lines width
                // word.GetWordWidth() returns the sum of strides tallyed.
                // word.CurrentWordsAddedSinceLastNewWord retruns a set letter count since the last NewWord call was issued
                bool wordtoobig = (word.GetWordWidth + drawDestRect.Width + rightMargin) >= destination.Width || word.CurrentWordsAddedSinceLastNewWord > bigWordSize;

                // It doesn't tell us when we should break that word.
                // This check is used to delimit words that are toobig
                // We also dump words that are getting toobig for a line
                float currentBoundryPosX = localDrawXY.X + stride + localDrawWH.X + rightMargin;
                bool withinWidthBoundry = currentBoundryPosX < destination.Right;

                // height is straightforward
                bool withinHeightBoundry = offset.Y + lineHeight < destination.Height;

                //
                // inbounds y
                //
                if (withinHeightBoundry)
                {
                    //
                    // is inbound x
                    //
                    if (withinWidthBoundry || isNextCharAWordSeperator || wordtoobig)
                    {
                        if ((isNextCharAWordSeperator && word.currentWordsLetterIndex >= 0) || wordtoobig)
                        {
                            //
                            // draw a word
                            //
                            for (int j = 0; j < word.CurrentWordsLetterLength; j++)
                            {

                                //
                                // Command operations probably should be stored then read back from a word
                                //
                                var test = word.specialCommand[j];
                                var tbadd = 0;
                                if ((test & 4) > 0)
                                {
                                    test -= 4;
                                    tbadd = 1;
                                }
                                var tlr_add = 0f;
                                if ((test & 2) > 0)
                                {
                                    test -= 2;
                                    tlr_add = 4f;
                                }
                                Color fcolor = color;
                                if ((test & 1) > 0)
                                {
                                    test -= 1;
                                    fcolor = Color.Red;
                                }

                                // prep vertices
                                var rotorigin = new Vector2(word.destDrawRectangle[j].X, word.destDrawRectangle[j].Y);
                                Vector2 lt = new Vector2(word.destDrawRectangle[j].X + tlr_add - tbadd, word.destDrawRectangle[j].Y - tbadd) - rotorigin;
                                Vector2 lb = new Vector2(word.destDrawRectangle[j].X - tbadd, word.destDrawRectangle[j].Bottom + tbadd) - rotorigin;
                                Vector2 rt = new Vector2(word.destDrawRectangle[j].Right + tlr_add + tbadd, word.destDrawRectangle[j].Y - tbadd) - rotorigin;
                                Vector2 rb = new Vector2(word.destDrawRectangle[j].Right + tbadd, word.destDrawRectangle[j].Bottom + tbadd) - rotorigin;

                                // rotate
                                if (rotation != 0)
                                {
                                    lt = new Vector2(lt.X * q.Y - lt.Y * q.X, lt.X * q.X + lt.Y * q.Y);
                                    lb = new Vector2(lb.X * q.Y - lb.Y * q.X, lb.X * q.X + lb.Y * q.Y);
                                    rt = new Vector2(rt.X * q.Y - rt.Y * q.X, rt.X * q.X + rt.Y * q.Y);
                                    rb = new Vector2(rb.X * q.Y - rb.Y * q.X, rb.X * q.X + rb.Y * q.Y);
                                }

                                // translate
                                lt += rotorigin;
                                lb += rotorigin;
                                rt += rotorigin;
                                rb += rotorigin;

                                // project
                                var LT = new Vector3(lt.X * cw - 1f, lt.Y * -ch + 1f, depth);
                                var LB = new Vector3(lb.X * cw - 1f, lb.Y * -ch + 1f, depth);
                                var RT = new Vector3(rt.X * cw - 1f, rt.Y * -ch + 1f, depth);
                                var RB = new Vector3(rb.X * cw - 1f, rb.Y * -ch + 1f, depth);

                                // uv coordinates in screen space
                                float uvL = (float)word.textureSourceRect[j].Left * cu;
                                float uvR = (float)word.textureSourceRect[j].Right * cu;
                                float uvT = (float)word.textureSourceRect[j].Top * cv;
                                float uvB = (float)word.textureSourceRect[j].Bottom * cv;
                                Vector2 uv0 = new Vector2(uvL, uvT);
                                Vector2 uv1 = new Vector2(uvL, uvB);
                                Vector2 uv2 = new Vector2(uvR, uvT);
                                Vector2 uv3 = new Vector2(uvR, uvB);

                                // create the vertice quad
                                spriteVertices[vi_pointer + 0] = new VertexPositionColorTexture(LT, fcolor, uv0);
                                spriteVertices[vi_pointer + 1] = new VertexPositionColorTexture(LB, fcolor, uv1);
                                spriteVertices[vi_pointer + 2] = new VertexPositionColorTexture(RT, fcolor, uv2);
                                spriteVertices[vi_pointer + 3] = new VertexPositionColorTexture(RB, fcolor, uv3);

                                // create the indexs
                                //
                                // LT 0   2 RT
                                //    |  /|     Triangle 1 is 0 1 2  ccw
                                //    | / |     Triangle 2 is 2 1 3  ccw
                                // LB 1   3 RB
                                triangleIndexs[ti_pointer + 0] = 0 + vi_pointer;
                                triangleIndexs[ti_pointer + 1] = 1 + vi_pointer;
                                triangleIndexs[ti_pointer + 2] = 2 + vi_pointer;
                                triangleIndexs[ti_pointer + 3] = 2 + vi_pointer;
                                triangleIndexs[ti_pointer + 4] = 1 + vi_pointer;
                                triangleIndexs[ti_pointer + 5] = 3 + vi_pointer;

                                /*
                                // this looks like a wonderful idea but im not getting good results maybe a few fps differnce
                                */

                                //fixed (VertexPositionColorTexture* vertexArrayFixedPtr = spriteVertices)
                                //{
                                //    var vertexArrayPtr = vertexArrayFixedPtr;
                                //    vertexArrayPtr += vi_pointer;

                                //    // create the vertice quad
                                //    *(vertexArrayPtr + 0) = new VertexPositionColorTexture(LT, color, uv0);
                                //    *(vertexArrayPtr + 1) = new VertexPositionColorTexture(LB, color, uv1);
                                //    *(vertexArrayPtr + 2) = new VertexPositionColorTexture(RT, color, uv2);
                                //    *(vertexArrayPtr + 3) = new VertexPositionColorTexture(RB, color, uv3);
                                //}

                                //fixed (int* indiceArrayFixedPtr = triangleIndexs)
                                //{
                                //    var indiceArrayPtr = indiceArrayFixedPtr;
                                //    indiceArrayPtr += ti_pointer;

                                //    // create the indexs
                                //    //
                                //    // LT 0   2 RT
                                //    //    |  /|     Triangle 1 is 0 1 2  ccw
                                //    //    | / |     Triangle 2 is 2 1 3  ccw
                                //    // LB 1   3 RB
                                //    *(indiceArrayPtr + 0) = 0 + vi_pointer;
                                //    *(indiceArrayPtr + 1) = 1 + vi_pointer;
                                //    *(indiceArrayPtr + 2) = 2 + vi_pointer;
                                //    *(indiceArrayPtr + 3) = 2 + vi_pointer;
                                //    *(indiceArrayPtr + 4) = 1 + vi_pointer;
                                //    *(indiceArrayPtr + 5) = 3 + vi_pointer;
                                //}

                                // increment the counts and indexs
                                currentTriangles += 2;
                                vi_pointer += 4;
                                ti_pointer += 6;

                                // check capacity
                                if (currentTriangles >= triangleCapacity - 2)
                                {
                                    IncreaseCapacity();
                                }
                                // record fast draw buffer data here.
                                //
                            }
                            word.NewWord(leftMargin);
                        }
                    }
                    else // is not inbounds x
                    {
                        // This is basically a crude raw rewind to deal with really big words that overflow a line
                        // Back up the drawing index to the start of the word on the next line and recalculate x
                        if (word.currentWordsLetterIndex >= 0)
                        {
                            i = word.GetRestartIndexAndResetBuffer(i, leftMargin);
                        }
                        offset.X = 2f;
                        offset.Y += lineHeight;
                        firstGlyphOfLine = true;
                    }

                }
                else
                {
                    i = text.Length;
                }
            }
            //}
        }


        /// <summary>
        /// A ProtoType initial drawable word buffer 
        /// Facilitates the ability to wrap words withing a text area.
        /// Uses rewinding feedback to deal with overflow.
        /// </summary>
        private class DrawableWordTracker
        {
            // What well do is buffer characters till we hit a space.
            // if the word is still in bounds well empty the buffer.
            // if not well act like we hit a new line.
            // then rewind the index to the begining of the word and clear the buffer.

            // this is important as a secondary tallying value 
            // for long words that must be deblt with.
            // as it affects non special cases short words
            public int letterAddedSinceLastNewWord = 0;
            // this is a letter counter and index pointer used to rewind the loop
            public int currentWordsLetterIndex = -1;
            // tallys x the position sum of the current letters in the buffer
            public float currentWordsLettersStrideOffsetWidthSum = 0;
            public List<Rectangle> destDrawRectangle = new List<Rectangle>();
            public List<Rectangle> textureSourceRect = new List<Rectangle>();
            public List<int> specialCommand = new List<int>();
            public List<int> charRecordedIndex = new List<int>();

            /// <summary>
            /// Note if currentCharStringIndex recieves a value to small everthing will go haywire. 
            /// it needs to predict a string that is too large before the width line check condition occurse
            /// </summary>
            public void SetDrawnLetterGlyph(Rectangle dest, Rectangle uvsource, float addedCharsPositionalOffsetWidth, int currentCharStringIndex, int spCommand)
            {
                letterAddedSinceLastNewWord++;
                currentWordsLetterIndex++;
                currentWordsLettersStrideOffsetWidthSum += addedCharsPositionalOffsetWidth;
                if (destDrawRectangle.Count < currentWordsLetterIndex + 1)
                {
                    destDrawRectangle.Add(dest);
                    textureSourceRect.Add(uvsource);
                    charRecordedIndex.Add(currentCharStringIndex);
                    specialCommand.Add(spCommand);
                }
                else
                {
                    destDrawRectangle[currentWordsLetterIndex] = dest;
                    textureSourceRect[currentWordsLetterIndex] = uvsource;
                    charRecordedIndex[currentWordsLetterIndex] = currentCharStringIndex;
                    specialCommand[currentWordsLetterIndex] = spCommand;
                }
            }

            /// <summary>
            /// This occurs when a word exceds the right side boundry.
            /// A line down should procced this call.
            /// </summary>
            public int GetRestartIndexAndResetBuffer(int currentcharindex, float leftmargin)
            {
                int result = currentcharindex;
                if (charRecordedIndex.Count > 0 && currentWordsLetterIndex > -1) { result = (int)(charRecordedIndex[0]); }//-1; }
                currentWordsLetterIndex = 0;
                currentWordsLettersStrideOffsetWidthSum = leftmargin; // margin
                return result;
            }

            public void NewWord(float leftmargin)
            {
                letterAddedSinceLastNewWord = 0;
                currentWordsLetterIndex = -1;
                currentWordsLettersStrideOffsetWidthSum = leftmargin;
            }

            /// <summary>
            /// this is used to check for big words
            /// </summary>
            public int CurrentWordsAddedSinceLastNewWord
            {
                get { return letterAddedSinceLastNewWord + 1; }
            }
            public int CurrentWordsLetterLength
            {
                get { return currentWordsLetterIndex + 1; }
            }
            public float GetWordWidth
            {
                get { return currentWordsLettersStrideOffsetWidthSum; }
            }
        }

KonajuGames:
What is the purpose of this PR?

> In c# strings ToString() implicitly casts to a stringbuilder as well as to a string.

I have read that statement several times and still don't understand what it means.

nkast:
> What is the purpose of this PR?

1) Replace this code 
```
 Vector2 size; // The framework will initialize the struct will zeros.
 spriteFont.MeasureString(ref source, out size); 
```
with this:
```
var size = spriteFont.MeasureString(ref source);
```
We don't get an advantage by having an 'out' method in cases where we must first initialize the parameter.

2) Hide SpriteFont.CharacterSource from the rest of the code (SpriteBatch) and isolate it inside SpriteFont. This will hurt the performance a bit. I guess I can drop this commit for now and revisit it if we decide to remove CharacterSource completely. (by implementing MeasureString() twice for string/StringBuilder)

Scellow:
@willmotil why FPS fluctuates so much ?

willmotil:
>     In c# strings ToString() implicitly casts to a stringbuilder as well as to a string.
> 
> I have read that statement several times and still don't understand what it means

I probably should of said that a StringBuilder can accept a string and internally turn it to a StringBuilder without creating garbage i suppose that's not really implicit but nothing about strings really are.

>         // A StringBuilder is internally represented as a linked list of blocks each of which holds
>         // a chunk of the string.  It turns out string as a whole can also be represented as just a chunk, 
>         // so that is what we do.
https://referencesource.microsoft.com/#mscorlib/system/text/stringbuilder.cs,adf60ee46ebd299f
you will even find stringbuilder in the string class.
https://referencesource.microsoft.com/#mscorlib/system/string.cs,8281103e6f23cb5c

> @willmotil why FPS fluctuates so much ?
I don't remember. It was a concept test app from long ago this is just a old pic. 

I pulled this out to show two things. 

One that expending any effort on charsource is pointless as it is redundant itself.

Two that measure string is just awful period this was just one idea i had to get around it though this is probably the slowest of the methods i wrote as its just a prototype, it has one Virtue even here.

It formats text to a rectangle using a buffer which could had i continued and finished it. Simply save every letter and its draw location for blasting it straight to the screen on every subsequent call. It basically measures as it draws here though so i could see it as i worked on it.

nkast:
@willmotil 
Creating a `new StringBuilder(String:text)` will generate garbage because  `StringBuilder`  it's a class.
I have to insist on a working sample that shows how to remove CharacterSource, or better yet a complete PR.

willmotil:
> Creating a new StringBuilder(String:text) will generate garbage because StringBuilder is a class.
String isn't a struct either, 

If you don't believe me take jon skeet and mark gravels word for it.
https://stackoverflow.com/questions/14704361/c-string-class-vs-string-struct.

Considered this ....

-CharSource is a struct that takes a string or stringbuilder. 
-Internally this struct holds two class objects a string and a stringbuilder. 
-It's role is to return either a string or stringbuilders character index.

> if (_string != null)
>      return _string[index];
> return _builder[index];

_The above is litterally it's sole function._
-This struct can be created by a drawstring or by one of two user accessed MeasureString calls. 
Keep in mind that this struct has two class instances within it.

Now consider the case within Drawstring were a CharSource struct is created... 
var **source** = new SpriteFont.CharacterSource(**text**);
When the method returns **source** goes out of scope and can be marked for sweeping by the gc,. 
Think about what this means....
This means that both object instances that are readonly will also go out of scope, such that both should now be marked for sweeping by the gc. 

Now if this were strictly true. 

> "Creating a new StringBuilder(String:text) will generate garbage because StringBuilder is a class."

As well if we believe that both Jon Skeet and Mark Gravel the guys who make c# are right that "_string is not a struct_".

 then... 

CharSource should would be a nightmare for the gc as it constantly would be sweeping a instance that was going out of scope 60 times per second or more ( by holding either a string object or stringbuilder instance within it). The reason it's not a nightmare, is not because its magical, its because it doesn't do anything. The fact its being passed as a reference is mute that doesn't matter either.

> I have to insist on a working sample that shows how to remove CharacterSource, or better yet a complete PR.

I could post a working example if you like it would be more like a bunch of simple edits.
-Delete the entire CharSource struct.
-Since nearly every method that uses CharSource either accepts directly a string or stringbuilder 
_(ill get to the measure strings methods in a second)._

-Remove each silly line such as this one in SpriteBatch.DrawString, at line 784
**var source = new SpriteFont.CharacterSource(text);**  
-Replace it with.
-var source = text; 

The measure strings. 
-Delete the the entire MeasureString overload that takes a stringbuilder.
-Change the MeasureString overload that takes a string to.
```
        private StringBuilder sb = new StringBuilder(4096);
        public Vector2 MeasureString(string s)
        {
            sb.Length = 0;
            sb.Append(s);
            Vector2 size;
            MeasureString(sb, out size);
            Return size;
        }
```
Change the parameters of MeasureString that accepts a CharSource struct 
from...
`internal unsafe void MeasureString(ref CharacterSource text, out Vector2 size){`
to
`internal unsafe void MeasureString(StringBuilder text, out Vector2 size){`

I don't think there is anything else to do, it's really... just there....

Jjagg:
@willmotil CharacterSource is a struct, so it's allocated on the stack. It takes either a String or StringBuilder, which comes from outside MG so it is **already allocated by the user** and won't be GC'ed when the CharacterSource that contains it goes out of scope. The other member is not used and never assigned, so no instance is created to be GC'ed. It's just a zero-allocation abstraction over String/StringBuilder. 

Sure, it's not strictly necessary, but IMO it makes for much nicer code by avoiding the duplication.

tomspilman:
First `CharacterSource` is a struct and has no impact on the GC.  It is just a helper to abstract the source of the characters and letting us share one implementation of `MeasureString`.   `CharacterSource` is better because it avoids a copy of a string and doesn't affect threading like a member `StringBuilder` would.

Eventually `CharacterSource` will be replaced by the new `ReadOnlySpan<T>` type which does what it does but in a much more efficient way.  See https://github.com/dotnet/corefxlab/blob/master/docs/specs/span.md

PR #5874, opened 2017-08-15
Improved performance of SpriteFont.MeasureString() & SpriteBatch.DrawString()

nkast:
* Glyph lookup
Replace Dictionary<char, Glyph> with a Glyph[] array and a custom CharacterRegion struct. We use binary search to find the CharacterRegion and from it it's easy to map a char in the Glyph[] array. Perfarmance is still good even with many Regions. I assume charachers in XNB SpriteFont are ordered.
I borrowed the Binary Search algorythm from Utilities.ByteBufferPool, thanks @Jjagg!. 

* Glyph access
Instead of getting a copy of a Glyph, we access it directly from the array. We use a fixed pointer of the array to avoid bound checks.

MeasureString(String) +52.4%
DrawString(SpriteFont, Vector2, Color) +24.4%

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -652,9 +652,9 @@ namespace Microsoft.Xna.Framework.Graphics
             float sortKey = (_sortMode == SpriteSortMode.Texture) ? spriteFont.Texture.SortingKey : 0;
 
             // Get the default glyph here once.
-            SpriteFont.Glyph? defaultGlyph = null;
+            int defaultGlyphIndex = -1;
             if (spriteFont.DefaultCharacter.HasValue)
-                defaultGlyph = spriteFont.TryGetGlyph(spriteFont.DefaultCharacter.Value);
+                spriteFont.TryGetGlyphIndex(spriteFont.DefaultCharacter.Value, out defaultGlyphIndex);
             
             var offset = Vector2.Zero;
             var firstGlyphOfLine = true;
@@ -674,7 +674,8 @@ namespace Microsoft.Xna.Framework.Graphics
                     continue;
                 }
  
-                var currentGlyph = spriteFont.GetGlyphOrDefault(c, defaultGlyph);
+                var currentGlyphIndex = spriteFont.GetGlyphIndexOrDefault(c, defaultGlyphIndex);
+                var currentGlyph = spriteFont.Glyphs[currentGlyphIndex];
 
                 // The first character on a line might have a negative left side bearing.
                 // In this scenario, SpriteBatch/SpriteFont normally offset the text to the right,
@@ -822,9 +823,9 @@ namespace Microsoft.Xna.Framework.Graphics
             }
 
             // Get the default glyph here once.
-            SpriteFont.Glyph? defaultGlyph = null;
+            int defaultGlyphIndex = -1;
             if (spriteFont.DefaultCharacter.HasValue)
-                defaultGlyph = spriteFont.TryGetGlyph(spriteFont.DefaultCharacter.Value);
+                spriteFont.TryGetGlyphIndex(spriteFont.DefaultCharacter.Value, out defaultGlyphIndex);
             
             var offset = Vector2.Zero;
             var firstGlyphOfLine = true;
@@ -844,7 +845,8 @@ namespace Microsoft.Xna.Framework.Graphics
                     continue;
                 }
 
-                var currentGlyph = spriteFont.GetGlyphOrDefault(c, defaultGlyph);
+                var currentGlyphIndex = spriteFont.GetGlyphIndexOrDefault(c, defaultGlyphIndex);
+                var currentGlyph = spriteFont.Glyphs[currentGlyphIndex];
 
                 // The first character on a line might have a negative left side bearing.
                 // In this scenario, SpriteBatch/SpriteFont normally offset the text to the right,
@@ -941,9 +943,9 @@ namespace Microsoft.Xna.Framework.Graphics
             float sortKey =  (_sortMode == SpriteSortMode.Texture) ? spriteFont.Texture.SortingKey : 0;
 
             // Get the default glyph here once.
-            SpriteFont.Glyph? defaultGlyph = null;
+            int defaultGlyphIndex = -1;
             if (spriteFont.DefaultCharacter.HasValue)
-                defaultGlyph = spriteFont.TryGetGlyph(spriteFont.DefaultCharacter.Value);
+                spriteFont.TryGetGlyphIndex(spriteFont.DefaultCharacter.Value, out defaultGlyphIndex);
             
             var offset = Vector2.Zero;
             var firstGlyphOfLine = true;
@@ -963,7 +965,8 @@ namespace Microsoft.Xna.Framework.Graphics
                     continue;
                 }
 
-                var currentGlyph = spriteFont.GetGlyphOrDefault(c, defaultGlyph);
+                var currentGlyphIndex = spriteFont.GetGlyphIndexOrDefault(c, defaultGlyphIndex);
+                var currentGlyph = spriteFont.Glyphs[currentGlyphIndex];
 
                 // The first character on a line might have a negative left side bearing.
                 // In this scenario, SpriteBatch/SpriteFont normally offset the text to the right,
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -674,52 +674,47 @@ public void DrawString (SpriteFont spriteFont, string text, Vector2 position, Co
...
-                if (!spriteFont.Glyphs.TryGetValue(c, out currentGlyph))
-                {
-                    if (!defaultGlyph.HasValue)
-                        throw new ArgumentException(SpriteFont.Errors.TextContainsUnresolvableCharacters, "text");
-
-                    currentGlyph = defaultGlyph.Value;
-                }
+ 
+                var currentGlyphIndex = spriteFont.GetGlyphIndexOrDefault(c, defaultGlyphIndex);
+                var pCurrentGlyph = pGlyphs + currentGlyphIndex;
```
tomspilman: Are we sure here that `currentGlyphIndex` can never let us index outside the `SpriteFont.Glyphs` array?

MonoGame.Framework/Graphics/SpriteFont.cs
```diff
@@ -175,40 +198,88 @@ internal void MeasureString(ref CharacterSource text, out Vector2 size)
...
+                    return m;
+                }
+            }
+
+            return -1;
+        }
+
+        internal bool TryGetGlyphIndex(char c, out int index)
+        {   
+            var regionIdx = GetRegionIndex(c);
```
tomspilman: Might as well inline this...  no one else calls it.  Plus then you can consider using an unsafe pointer to access `_regions`.

MonoGame.Framework/Graphics/SpriteFont.cs
```diff
@@ -83,7 +106,7 @@ public int GetHashCode(char b)
         /// <remarks>Can be used to calculate character bounds when implementing custom SpriteFont rendering.</remarks>
         public Dictionary<char, Glyph> GetGlyphs()
```
tomspilman: Random thought here.

Maybe we should start to depreciate this method and add a new type: `GlyphDictionary`.  It would provide a read only `Dictionary` like API, but uses your fast region lookup internally.  This can then be used directly by `SpriteFont` and `SpriteBatch` as well as user's custom font measuring/rendering.

MonoGame.Framework/Graphics/SpriteFont.cs
```diff
@@ -50,7 +52,9 @@ public int GetHashCode(char b)
 			Spacing = spacing;
 			DefaultCharacter = defaultCharacter;
 
-			_glyphs = new Dictionary<char, Glyph>(characters.Count, CharComparer.Default);
+            _glyphsDictionary = new Dictionary<char, Glyph>(characters.Count, CharComparer.Default);
+            _glyphs = new Glyph[characters.Count];
+            var regions = new Stack<CharacterRegion>();
 
 			for (var i = 0; i < characters.Count; i++)
```
tomspilman: This should have been all done in the content pipleline and not runtime.  We should eventually research if we can increment the version on `SpriteFont` files and pack this data better for runtime use.

MonoGame.Framework/Graphics/SpriteBatch.cs
nkast: I don't believe so. However it wont hurt to put a single check here just to be safe.

MonoGame.Framework/Graphics/SpriteFont.cs
nkast: For now I could generate the Dictionary<char, Glyph> on the fly in GetGlyphs(). `_glyphsDictionary` is not used anywhere else.

nkast:
One more thing...
We can cache the `defaultGlyphIndex` when `DefaultCharacter` property changes instead of repeatedly pre-cache it on every DrawString() &  MeasureString().

MonoGame.Framework/Graphics/SpriteFont.cs
tomspilman: > For now I could generate the Dictionary<char, Glyph> on the fly 

That is not a bad idea... make it a cost at request time for only people that use it.

MonoGame.Framework/Graphics/SpriteBatch.cs
tomspilman: Could maybe be a debug only check so that the normal case is faster.

tomspilman:
> We can cache the `defaultGlyphIndex` when `DefaultCharacter` property changes

That is a good idea too.

nkast:
MeasureString() Benchmark of 3 fonts with a single region of 64, 256 & 1024 characters.
Values are in mseconds.
tnc_spritefont

Blue line is `Develop`. The current implementation uses a Dictionary (=hashtable, O(1) ), therefore it scales pretty well.
Green line is this PR. The number of characters is irrelevant since they are all in a single region. Mapping from Char -> Index is O(1).
Red line is  this PR modified to generate 1 region per character. It demonstrates how it scales with 64, 256 & 1024 regions. Finding the correct region is a O(logn).

Jjagg:
Gotta love the graphs :D 
Very impressive gains from this PR!

tomspilman:
This is great... thanks  @nkast !

I think as a follow up adding a `internal int SpriteFont.DefaultCharacterIndex` which is updated from the `SpriteFont.DefaultCharacter` setter would make things a tiny bit faster and simpler to understand.

nkast:
Thanks Tom,

We can fallback to a Dictionrary lookup when a spriteFont has more than 256 or 512 regions.
Am I over-thinking this? It's very unlikely than anyone will end up with that many character regions.

tomspilman:
> It's very unlikely than anyone will end up with that many character regions.

Yeah that is what I am thinking... I doubt any font using in a shipping game would end up with that many regions.

SeriousMartin:
Please have a look at my comment: https://github.com/MonoGame/MonoGame/commit/a345b71073f1c8dd6b222e521fe2a8324797c81a#r25503962

For me the index was outside the bounds of the array when I created the spritefont based on a custom texture.

Replacing
```
            Array.Reverse(_regions)
```
in Line 91 of SpriteFont.cs with
```
            Array.Sort(_regions, (a, b) =>
            {
                return a.Start.CompareTo(b.Start);
            });
```

Resolves the issue.

Jjagg:
> I assume charachers in XNB SpriteFont are ordered.

@nkast I think it makes sense to just sort the regions in the constructor like @SeriousMartin proposes.

nkast:
I guess this will work but it will also result in poor performance. If the characters are out of order but I suppose it creates a lot of small regions, which will kill the lookup performance.

@SeriousMartin 
How can it be that the charachers are unorder? The font processor/importer sorts the characters before generating the XNB. The "[Compiled (XNB) Content Format](http://xbox.create.msdn.com/en-US/sample/xnb_format)" is also clear about this fact.
"Character map: Unicode character code points corresponding to each rectangle in the glyph and cropping list *(guaranteed sorted into ascending order)*"

So maybe we are missing something in the Texture-Font processor?

SeriousMartin:
@nkast Performance is an important point and I agree - fixing FontTextureProcessor will be the best solution. We probably only need to add sorting to the ExtractGlyph method https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/FontTextureProcessor.cs#L85

I suggest to also add a unit test that imports a texture font that is not in order.

nkast:
The FontTextureProcessor starts at FirstCharacher and then increase the characters by one. I still don't see anything in the code that could have done that. Can you share the font-texture and the .mgcb file?

SeriousMartin:
I currently can't test this, because I only have an old MonoGame version and the build server is not running.

> The FontTextureProcessor starts at FirstCharacher and then increase the characters by one. 
Can you point to a code line? From my understanding the processor goes through all the pixels and adds the regions from top left to bottom right (https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/FontTextureProcessor.cs#L41). Glyphs seem to have the same order.

nkast:
#6125 will catch any invalid SpriteFont.

SeriousMartin:
@nkast Had another look at FontTextureProcessor and finally realized I'm using a custom implementation - embarrassing...

Adding the check is a perfect solution and will make sure other devs don't run into this problem.

Thansk a lot!

Issue #5883, opened 2017-08-18
MSBuild gets stuck when compiling a project with missing Content.mgcb


vpenades:
[This PR](https://github.com/MonoGame/MonoGame/pull/4997) has modified heavily MonoGame.Content.Builder.targets with an unexpected consequence:

When compiling a project with missing content.mgcb (because it's not needed) MSBuild gets stuck!, as if it's doing some sort of infinite loop.

Now I have multiple MonoGame utility assemblies that leave visual studio blocked forever when I try to compile them!

Steps to reproduce:

Create a new Monogame Windows Project
Delete content.mgcb
Compile

KonajuGames:
I submitted a potential fix for this (#5886).  Try that branch or try the [installer](http://teamcity.monogame.net/viewLog.html?buildId=51930&tab=artifacts&buildTypeId=MonoGame_PackagingWindows) generated from it.  Let me know if it fixes the issue.

vpenades:
@KonajuGames yes, just checked and it seems the problem has been solved.

Thanks!

tomspilman:
Thanks for verifying the fix @vpenades !

PR #5886, opened 2017-08-21
Fix for build hang with no mgcb file in project

KonajuGames:
If the project contains no mgcb file the targets file appears to cause the build to hang. What it actually seems to be doing is trying to include the entire drive recursively as content because `ContentReferences.ContentOutputDir` may be empty (from what I understand of the targets file syntax)


Jjagg:
Should fix #5883.

dazinator:
> What it actually seems to be doing is trying to include the entire drive recursively

Wow. Thanks for fixing this. I think this is officially the best / worst bug I have ever accidentally introduced via a PR. On the scale of epic failures it's a little way short of deleting the entire drive - but not a million miles away :-)

DDReaper:
Nice quick fix @KonajuGames

tomspilman:

PR #5887, opened 2017-08-22
Fixes to switching away from exclusive full-screen

### LINKED ISSUE #5885: Fullscreen + Alt Tab still an issue in latest MonoGame 3.7.0.753 Development build
I put together a video that shows the issue is still persisting. I saw that the latest dev build had some fixes committed 4 days ago for this issue, but it appears to persist at least on my box. 

VIDEO: https://youtu.be/1hT_YOruTdk

Machine:
1) Windows 10, 64-bit, Latest Update Pack installed as of 8/19/2017
2) NVIDIA

Steps taken to reproduce :
1) Uninstalled Monogame v3.7.0.744
2) Installed Monogame v3.7.0.753
3) Cleaned the projects
4) Rebuilt
5) Run the game in FULLSCREEN mode
6) ALT-TAB out to the desktop

Result:
1) Game icon is not visible on the task bar.
2) The game is still running as a background process (interestingly enough, OBS was still able to capture it!)
3) Holding down ALT-TAB also shows that the game process is not accessible
4) I have to End Task from the task manager to shut the game down.

Thoughts?

KonajuGames:
Fixes DirectX switching away from exclusive full-screen mode by forcing the swapchain out of exclusive mode and minimizing the window.
Fixes OpenGL switching away from exclusive full-screen mode by enabling the SDL hint to minimize video on loss of focus when changing to exclusive full-screen mode, otherwise the hint is left disabled.

Fixes #5885

### DIFF
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -218,8 +218,11 @@ namespace Microsoft.Xna.Framework
                 Sdl.Window.SetFullscreen(Handle, (_willBeFullScreen) ? fullscreenFlag : 0);
                 _hardwareSwitch = _game.graphicsDeviceManager.HardwareModeSwitch;
             }
-            // If going to exclusive full-screen mode, force the window to minimize on focus loss
-            Sdl.SetHint("SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS", _willBeFullScreen && _hardwareSwitch ? "1" : "0");
+            // If going to exclusive full-screen mode, force the window to minimize on focus loss (Windows only)
+            if (CurrentPlatform.OS == OS.Windows)
+            {
+                Sdl.SetHint("SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS", _willBeFullScreen && _hardwareSwitch ? "1" : "0");
+            }
 
             if (!_willBeFullScreen || _game.graphicsDeviceManager.HardwareModeSwitch)
             {
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -100,7 +100,7 @@ namespace Microsoft.Xna.Framework
 
             _width = GraphicsDeviceManager.DefaultBackBufferWidth;
             _height = GraphicsDeviceManager.DefaultBackBufferHeight;
-            
+
             Sdl.SetHint("SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS", "0");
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/SDL/SDLGameWindow.cs
```diff
@@ -218,6 +218,8 @@ public override void EndScreenDeviceChange(string screenDeviceName, int clientWi
                 Sdl.Window.SetFullscreen(Handle, (_willBeFullScreen) ? fullscreenFlag : 0);
                 _hardwareSwitch = _game.graphicsDeviceManager.HardwareModeSwitch;
             }
+            // If going to exclusive full-screen mode, force the window to minimize on focus loss
+            Sdl.SetHint("SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS", _willBeFullScreen && _hardwareSwitch ? "1" : "0");
```
cra0zy: This will cause problems on multimonitor setups, especially on Linux.

KonajuGames: It was tested on my multi-monitor setup here on Windows and works as it should.  Previously it did not remove the window at all, leaving that monitor useless.  If it causes issues on multi-monitor setups on Linux, then they can be fixed later because this is a major issue on Windows right now.

cra0zy: Add a conditional to only apply that to Windows, on Linux I can just manually minimize the game and I get the resolution back to the original.

tomspilman: > then they can be fixed later

I agree.

> Add a conditional to only apply that to Windows,

If we do this use the runtime conditional here, so that one binary that works on Win/Mac/Linux still works.

cra0zy: > If we do this use the runtime conditional here, so that one binary that works on Win/Mac/Linux still works.

Talking about:
```c#
if (CurrentPlatform.OS == OS.Windows)
{
   // Do Stuff
}
```

cra0zy: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Utilities/CurrentPlatform.cs

KonajuGames: How does Mac handle this? Do we have to make an exemption for Mac as well?

KonajuGames:
Exemption for Linux and Mac added.

MonoGame.Framework/SDL/SDLGameWindow.cs
cra0zy: IDK how Mac handles it. someone should probably test it out.

tomspilman:

PR #5893, opened 2017-08-23
Fixes #5892 - Fix xbuild.

### LINKED ISSUE #5892: fix xbuild
#4977 broke xbuild.

We need to either revert it, or get xbuild working. One day we can drop xbuild as will be replaced by msbuild, but that day isn't yet today apparantly.

dazinator:
I moved the complicated msbuild logic from the targets file, into a seperate msbuild task (c#), because:
1. xbuild can't handle it.
2. This opens a path to create some test coverage of the build logic through testing the msbuild task with nunit / xunit etc.

To test:

1. Build the `MonoGame.Framework.MsBuildTasks` project, and then put the `MonoGame.Framework.MsBuildTasks` dll in `C:\Program Files (x86)\MSBuild\MonoGame\v3.0\Tools\` folder.
2. Overwrite the `MonoGame.Content.Builder.targets` file in `C:\Program Files (x86)\MSBuild\MonoGame\v3.0\` with the version from this PR.

3. Create a new monogame project. 
4. Add some `mgcb`s. You can experiment by adding linked and non linked mgcb's, to different directories within the project.
5. Build the project.
6. Check the output directory for all the content.

## Debugging the msbuild task

2. In your sample monogame project, edit the csproj file, and add a `TasksAssemblyFolderPath` property which points to the bin/debug folder that contains the task assembly. For my sample project which I created in the monogame repo directory, it looks like this:
```
 <PropertyGroup>
    <TasksAssemblyFolderPath>$(MSBuildThisFileDirectory)..\MonoGame.Framework.MsBuildTasks\bin\Debug\</TasksAssemblyFolderPath>

```

Edit the project properties for the MsBuildTasks project, so that it builds the sample project using msbuild / xbuild when you run it. For me this looks like:

image

Now you can run the  msbuild tasks project, and set a break point in the msbuild task to debug.

## Other notes:

I build each mgcb in parallel as I thought would be better for performance. Seems to work well.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
@@ -16,17 +16,17 @@
     <DisableFastUpToDateCheck>true</DisableFastUpToDateCheck>
   </PropertyGroup>
 
-  <Target Name="Prepare">
-    <PropertyGroup>
-      <!-- El Capitan Support -->
-      <MonoExe Condition=" '$(OS)' != 'Windows_NT' And Exists ('/Library/Frameworks/Mono.framework/Versions/Current/bin/mono') ">/Library/Frameworks/Mono.framework/Versions/Current/bin/mono</MonoExe>
-      <MonoExe Condition=" '$(OS)' != 'Windows_NT' And Exists ('/usr/local/bin/mono') ">/usr/local/bin/mono</MonoExe>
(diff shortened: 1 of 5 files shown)


dellis1972:
I know @tomspilman has tried hard to avoid using MSBuild Tasks if possible. @tomspilman  thoughts on this?

One of the things I do is modify the `Header` item to change the command line arguments to pass in additional custom options (like Android texture compression method).  It saves me from having a bunch of different Content.mgcb files. If we do go that route we should include support for adding custom arguments.

cra0zy:
This has already been fixed in develop branch.

dazinator:
Excellent, glad to hear it :-)

PR #5896, opened 2017-08-26
Move the DesktopGL platform over to use the GraphicsAdapter partial class [dont merge]

dellis1972:
In the continuing effort to update/remote old code in favour of
partial classes, this commit adds a GraphicsAdapter.SDL for the
DesktopGL platform.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -598,13 +598,16 @@
     </Compile>
     <Compile Include="Graphics\Effect\SpriteEffect.cs" />
     <Compile Include="Graphics\GraphicsAdapter.cs">
-      <ExcludePlatforms>Android,Angle,iOS,Linux,MacOS,WindowsGL,tvOS,Web</ExcludePlatforms>
+      <ExcludePlatforms>Android,Angle,iOS,MacOS,tvOS,Web</ExcludePlatforms>
     </Compile>
     <Compile Include="Graphics\GraphicsAdapter.DirectX.cs">
       <Services>DirectXGraphics</Services>
     </Compile>
+    <Compile Include="Graphics\GraphicsAdapter.SDL.cs">
+      <Platforms>Linux,WindowsGL</Platforms>
+    </Compile>
     <Compile Include="Graphics\GraphicsAdapter.Legacy.cs">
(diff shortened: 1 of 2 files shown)


dellis1972:
@cra0zy I think this will need testing on a multi monitor setup. I'm not sure what the behaviour is supposed to be.. the old code we used to always grab the CurrentDisplayMode for the current window regardless of if that was on the current adapter. But then again we only ever had one adapter..

dellis1972:
@tomspilman regarding my previous comment, it would be useful to know what the expected behaviour should be regarding GraphicsAdapter.CurrentDisplayMode in a multi monitor environment. and what happens when you move the screen from one to the other.

tomspilman:
> it would be useful to know what the expected behaviour should be regarding GraphicsAdapter.CurrentDisplayMode in a multi monitor environment

I have a three monitor setup here... what tests do you think I should run under XNA?

My guess is that `CurrentDisplayMode` returns the information on the primary monitor only.

KonajuGames:
The doc for GraphicsDeviceManager.FindBestDevice talks about a "current
adapter". I'm assuming that is the primary adapter since there is no other
concept of a current adapter anywhere else in the API.

dellis1972:
@tomspilman I guess the test would be to run up and XNA app on one monitor, then move it to another monitor and see if any of the Adapter settings change.

tomspilman:
@dellis1972 - So my findings.  I have a single ATI card with 3 monitors hooked up.  

After the game starts the `GraphicsAdapter.DefaultAdapter` does not change at all no matter where i move the game window.

I noticed that the default adapter is always whatever the "main display" is as defined in the Windows display settings.   Also the game Window always opens on that default display.

And even if i changed the default display in the Windows display settings while the game is running, the `GraphicsAdapter.DefaultAdapter` was always the one picked at game startup.  It never changes after start.

Hope this helps.

KonajuGames:
I can do a test here soon with XNA, but my guess is that
GraphicsAdapter.DefaultAdapter won't change when you move the window from
one screen to another. The GraphicsDevice.Adapter value might though.

willmotil:
I suggest you run those tests in full screen and change the display modes. Current adapter ussually matches the default adapter however you can list all the available adapters in xna.

PR #5902, opened 2017-08-28
Get the default glyph index once in DefaultCharacter {get}

nkast:


MonoGame.Framework/Graphics/SpriteFont.cs
```diff
@@ -117,7 +119,19 @@ public int GetHashCode(char b)
...
-		public char? DefaultCharacter { get; set; }
+		public char? DefaultCharacter
+        {
+            get { return _defaultCharacter; }
+            set
+            {
+                _defaultCharacter = value;
+                // Get the default glyph index here once.
+                if (_defaultCharacter.HasValue)
+                    TryGetGlyphIndex(_defaultCharacter.Value, out _defaultGlyphIndex);
```
tomspilman: What happens here if the user assigns a character not in the font?

nkast: TryGetGlyphIndex() will set _defaultGlyphIndex = -1, that's all. 
No Exception is thrown until GetGlyphIndexOrDefault() tries to draw the default character.
I tried to preserve the existing behavior and it pass the existing tests.
Perhaps XNA throws a differed message in that case, I 'll have to check ...

nkast: OK, XNA throws an ArgumentException immediately when you set DefaultCharacter.
We have to add a test for this change too, I will leave this for later.

tomspilman:
This is great... thanks @nkast !

PR #5912, opened 2017-08-31
[GL] Check vbo in BufferBindingInfo

Jjagg:
The check in ApplyAttribs would block setting up vertex channels when the vbo changed and all other settings were the same as what was last applied.

Someone reported this issue on the forum: http://community.monogame.net/t/repeated-rendering-weird-behaviours-after-upgrading-to-3-7/9643/3

### DIFF
--- a/Test/Framework/Graphics/GraphicsDeviceTest.cs
+++ b/Test/Framework/Graphics/GraphicsDeviceTest.cs
@@ -9,6 +9,7 @@ using Microsoft.Xna.Framework.Content;
 using Microsoft.Xna.Framework.Graphics;
 using MonoGame.Tests.ContentPipeline;
 using NUnit.Framework;
+using OpenGL;
 
 namespace MonoGame.Tests.Graphics
 {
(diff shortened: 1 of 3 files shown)


Jjagg:
I tested the users project to confirm this PR fixes the issue.

KakCAT:
I can confirm it works, it fixed several "stray polygons" in iOS project of mine where indexed primitives where used.

tomspilman:
@Jjagg - Any sort of unit test we can add to trigger the bug and verify it never comes back?

Jjagg:
Yeah, I'll do that.

Jjagg:
@tomspilman I added a test :)

Jjagg:
@tomspilman or @ KonajuGames can this be merged please?

Test/Framework/Graphics/GraphicsDeviceTest.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Xna.Framework.Graphics;
 using MonoGame.Tests.ContentPipeline;
 using NUnit.Framework;
+using OpenGL;
```
tomspilman: I wonder why this isn't breaking the Windows DirectX build.  We wouldn't expect this namespace to exist outside of `DESKTOPGL` right?

Jjagg: IWindowInfo and IGraphicsContext are in the OpenGL namespace and they're included (but not used) in MG DX. If you want I'll exclude them from MG DX in this PR.

tomspilman: I see... well for now we don't need to change this PR then.

Ideally we open a new PR and first rename that namespace from `OpenGL` to `MonoGame.OpenGL` to make it clear it is a MG thing and not some other asssembly.  Then it would also be good for those to be only added to the appropriate OGL platforms.

tomspilman: Also noticed this...

image

Probably none of that should be part of the public API... but it is as it shows up in our documentation now.

That should likely all be cleaned up and made private/internal.

tomspilman:
Thanks @Jjagg ... merging!

Test/Framework/Graphics/GraphicsDeviceTest.cs
tomspilman: @Jjagg - @KonajuGames handled this...  https://github.com/MonoGame/MonoGame/pull/5927

PR #5921, opened 2017-09-03
Pool ContentManager scratch buffers

### LINKED ISSUE #5341: v3.5.1+ ContentManager.Unload() doesn't seem to free up memory
After upgrading from 3.4 to 3.5.1 (and then 3.6) I noticed that calling ContentManager.Unload() doesn't lessen the amount of memory being used by the game, as seen in the Diagnostic Tools of VS2015 and the Resource Monitor on Windows. Loading, unloading, then reloading the same content would originally keep the memory usage about the same, but now it continues to increase during every Load() while not decreasing during Unload(). 

I tested this further by installing v3.4 again and using the same project and code. Loading and unloading the same content as before now has the desired effect where the memory usage increases during Load() and decreases after Unload(), as expected.

Jjagg:
As discussed in #5341. (edit for autoclose: fixes #5341)

I don't think this is right as is. Consider the case where a user loads incrementally larger assets. For each asset a new scratch buffer will be created and returned to the pool, staying in memory until game exit.

Maybe it would be better to use a slightly modified version of ByteBufferPool that deletes the smallest buffer from the pool whenever Get is called and there are no buffers that are large enough. That way the pool size won't grow beyond the number of threads that use a buffer from it simultaneously (most often 1).

cc @nkast

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -28,10 +28,10 @@ namespace Microsoft.Xna.Framework.Content
         private Dictionary<string, object> loadedAssets = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
 		private List<IDisposable> disposableAssets = new List<IDisposable>();
         private bool disposed;
-        private byte[] scratchBuffer;
 
 		private static object ContentManagerLock = new object();
         private static List<WeakReference> ContentManagers = new List<WeakReference>();
+        private static ByteBufferPool _scratchBufferPool = new ByteBufferPool();
 
         private static readonly List<char> targetPlatformIdentifiers = new List<char>()
         {
@@ -191,7 +191,6 @@ namespace Microsoft.Xna.Framework.Content
                     Unload();
                 }
 
-                scratchBuffer = null;
 				disposed = true;
 			}
 		}
@@ -482,11 +481,14 @@ namespace Microsoft.Xna.Framework.Content
 		}
 
         internal byte[] GetScratchBuffer(int size)
-        {            
+        {
             size = Math.Max(size, 1024 * 1024);
-            if (scratchBuffer == null || scratchBuffer.Length < size)
-                scratchBuffer = new byte[size];
-            return scratchBuffer;
+            return _scratchBufferPool.Get(size);
         }
-	}
+
+        internal void ReturnScratchBuffer(byte[] buffer)
+        {
+            _scratchBufferPool.Return(buffer);
+        }
+    }
 }
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -31,7 +31,8 @@ namespace Microsoft.Xna.Framework.Content
 
 		private static object ContentManagerLock = new object();
         private static List<WeakReference> ContentManagers = new List<WeakReference>();
-        private static ByteBufferPool _scratchBufferPool = new ByteBufferPool();
+
+        internal static readonly ByteBufferPool ScratchBufferPool = new ByteBufferPool(1024 * 1024, Environment.ProcessorCount);
 
         private static readonly List<char> targetPlatformIdentifiers = new List<char>()
         {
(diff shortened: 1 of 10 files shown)


nkast:
I agree 100%. 

Here is a bufferPool I'am testing and works exactly  like that.
https://github.com/nkast/MonoGame/blob/beac8d7e7ae87877d1d282ecd3208594ae00c2cb/MonoGame.Framework/Content/ContentManager.cs#L490-L532  

The only danger is if the user hit it with dozens or hundreds of threads at once. Should we use a Semaphore or something there?

Jjagg:
You mean to allow a maximum of N threads to use a buffer from the pool at once? We could do that, it would certainly make this more foolproof :)

nkast:
yes, it's an extreme case of course. 
or we can drop returned buffers when the pool already has reached a maximum number.

Jjagg:
I prefer the second option! I'll close this since you already implemented the better pool.

nkast:
I kind of stuck on this issue and then forgot about it.

nkast:
no,  I'am ok with it.

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -476,11 +475,14 @@ public IServiceProvider ServiceProvider
...
             size = Math.Max(size, 1024 * 1024);
-            if (scratchBuffer == null || scratchBuffer.Length < size)
-                scratchBuffer = new byte[size];
-            return scratchBuffer;
+            return _scratchBufferPool.Get(size);
+        }
+
+        internal void ReturnScratchBuffer(byte[] buffer)
+        {
+            _scratchBufferPool.Return(buffer);
```
tomspilman: How about just making `_scratchBufferPool` `internal` and let people call it directly?  Just to avoid these redirect methods.  Just make `ByteBufferPool` ctor take an optional minimum size.

nkast: I am not sure if it's best to modify ByteBufferPool or use a second pool and let the two take different roads for contentManager & soundEffects. In every case I prefer having a layer that allows us to be more flexible with the internal implementation. Maybe having a IBufferPool with Get()/Return() or something like that.

nkast:
@Jjagg 
My branch is not ready to submit, I just took a pool class I had, paste it in ContentManager.cs and modify it to get the 1 buffer per thread as you described.  Ideally, I was hopping to have 1 buffer per Core - max.

Jjagg:
@nkast We can modify BBP to have a maximum number of items in the pool and a minimal buffer size. Maximum pool size for ContentManager can be set to Environment.ProcessorCount. That way we only need one implementation.

Jjagg:
`ContentManager.ScratchBufferPool` is now exposed so ContentReaders can use it directly. I added a minimum buffer size and maximum pool size option and set the max pool size for the scratchbuffers to `Environment.ProcessorCount`.

Anything left to do? This seems like a good solution to me :)

Jjagg:
I modified the ByteBufferPool to remove a buffer when it does not have a large enough buffer for a `Get` call as explained in my first post. This ensures the pool does not grow larger than the number of buffers in use simultaneously.

@nkast, @tomspilman Sounds good?

nkast:
> This ensures the pool does not grow larger than the number of buffers in use simultaneously.

Is that desirable for SoundEffectInstances ? 
maybe we can adjust that with a _minBuffers.

Other than that those changes will work very well for the ContentManager.

Jjagg:
I think it is. We use the ByteBufferPool for DynamicSoundEffect instances. The buffers there are not immediately returned like the ScratchBufferPool buffers. So the pool will grow to the maximum number of DynamicSoundEffect instances in use at the same time and no larger.

Jjagg:
Bump! IMO this is good to go like this :)

Jjagg:
Bumping because I think someone might've ran into the same issue as #5341 (which this fixes): http://community.monogame.net/t/contentmanger-not-unloading-properly/10550

Jjagg:
@cra0zy This is good to go!

PR #5927, opened 2017-09-07
Types in OpenGL and OpenAL namespaces now internal

KonajuGames:
OpenGL namespace renamed to MonoGame.OpenGL. All symbols within changed from public to internal.
OpenAL namespace renamed to MonoGame.OpenAL. All symbols within changed from public to internal.

### DIFF
--- a/MonoGame.Framework/Audio/AudioLoader.cs
+++ b/MonoGame.Framework/Audio/AudioLoader.cs
@@ -4,7 +4,7 @@
 
 using System;
 using System.IO;
-using OpenAL;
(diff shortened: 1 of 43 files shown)


tomspilman:
@dellis1972 @cra0zy - To be sure... there was no plan here that we needed to keep this part of the public API correct?

cra0zy:
> @dellis1972 @cra0zy - To be sure... there was no plan here that we needed to keep this part of the public API correct?

None whatsoever.

tomspilman:
Thanks for cleaning this up @KonajuGames !

PR #5897, opened 2017-08-27
Fix a crash when building content in xbuild 

dellis1972:
PR #4997 Introduced a crash in xbuild. This is because of a bug in xbuild where `<ItemGroup>` with Metadata are not supported. Instead
you need to use `CreateItem`.

This commit fixes that issue.

There are also other issues which need to be resolved with #4997.

That said this commit also adds a new unit test for the content system.
There is a sample project which we build as part of the unit test.
We check the output to make sure it is in the right place. This test
should be expaned on to include tests for Shared Projects and other
senarios that users use. It should allow use to catch issues such as
what happened in #4997 at the point the PR is opened.

PR #5938, opened 2017-09-14
Add new build task projects to .gitignore

Jjagg:
They are generated by Protobuild and I noticed they weren't ignored yet.
Projects where added in #5897.


tomspilman:
Thanks @Jjagg !

PR #5940, opened 2017-09-14
Fix #5939. Install MonoGame.Build.Tasks in the right folder

### LINKED ISSUE #5939: MonoGame.Build.Tasks.dll not put into the right folder at installation ?
Got error when updated to monogame latest build (the problem started after one of the builds folowing the 828):
"Severity	Code	Description	Project	File	Line	Suppression State
Error	MSB4062	The "MonoGame.Build.Tasks.CollectContentReferences" task could not be loaded from the assembly C:\Program Files (x86)\MSBuild\MonoGame\v3.0\MonoGame.Build.Tasks.dll. Could not load file or assembly 'file:///C:\Program Files (x86)\MSBuild\MonoGame\v3.0\MonoGame.Build.Tasks.dll' or one of its dependencies. Le fichier spécifié est introuvable. Confirm that the declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask"

cf http://community.monogame.net/t/trouble-getting-started-with-monogame-monogame-build-tasks-collectcontentreferences-solved/9704

Solution seems to manually move the file to the parent folder.

When I do this, the error has vanished (that's now normal), but I seem to be getting the same problem as #5904 <- #5903, building a project takes ages. The only way for me to stop it is to kill MSbuild.exe in the taskmanager

Changes between the last build  i know working (828) vs the latest:
image

dellis1972:
Fixes #5939

### DIFF
--- a/Installers/Windows/MonoGame.nsi
+++ b/Installers/Windows/MonoGame.nsi
@@ -80,6 +80,7 @@ Section "MonoGame Core Components" CoreComponents ;No components page, name is n
   SetOutPath ${MSBuildInstallDir}
   File '..\..\MonoGame.Framework.Content.Pipeline\MonoGame.Content.Builder.targets'
   File '..\..\MonoGame.Framework.Content.Pipeline\MonoGame.Common.props'
+  File '..\..\Tools\Pipeline\bin\Windows\AnyCPU\Release\MonoGame.Build.Tasks.dll'
 
   ; Install the MonoGame tools to a single shared folder.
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks @dellis1972 !

Soraiko:
Where do I find this DLL ?????
It's nowhere. If there's a "right" folder, then where's the "wrong" one? Can I get the DLL PLEASE ?

Issue #5929, opened 2017-09-08
Provide SHA-256 checksums for installers


remmah:
The macOS MonoGame installer is unsigned and has to be downloaded in the clear from monogame.net. Would it be possible to provide the SHA-256 digest for each installer on a TLS-enabled site? Perhaps in the notes for each release tag here on GitHub?

tomspilman:
I can't say i know anything about doing this, but willing to do it as long as it can be 100% automated.

remmah:
Probably the easiest way is to run `[sha256sum](https://help.ubuntu.com/community/HowToSHA256SUM)` on each installer after it's built and log the output to a place where whoever writes the release notes can securely access that output. I'm new to MonoGame so I'm not familiar with the build process and whether it offers an opportunity to run a script after the installers are built.

An alternative to providing SHA digests is to just make the installer downloads accessible via HTTPS (either on GitHub or another site with a valid TLS cert). Depending on how monogame.net is hosted, it may be possible to set up an automatically-renewing TLS cert from [Let'sEncrypt](https://letsencrypt.org/).

Either way, the concern is making sure that the bits people get when downloading the installers are the same as what the MonoGame team uploads, with nothing modified or corrupted in-transit.

harry-cpp:
Installers have been replaced by templates/nugets.

PR #5941, opened 2017-09-14
Add sha256 files for the Mac and Linux installers.

dellis1972:
Fix up the android build to it will work with the latest SDK on Mac.

### DIFF
--- a/Installers/default.build
+++ b/Installers/default.build
@@ -34,6 +34,20 @@
         <exec program="makensis" workingdir="Windows" basedir="C:\Program Files (x86)\NSIS">
 		      <arg value="MonoGame.nsi"/>
         </exec>
+        <exec program="" workingdir="Windows" basedir="">
+              <arg value="" />
+        </exec>
+        <echo append="false" file="Windows/header.nsh">
+@ECHO OFF
+FOR /f "tokens=*" %%i IN ('@certutil -hashfile %1 %2 ^| find /v "hash of file" ^| find /v "CertUtil"') DO SET r=%%i
+SET r=%r: =%
+ECHO %r%  %1 > %3
+        </echo>
+        <exec program="hash.bat" workingdir="Windows">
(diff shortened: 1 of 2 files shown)


tomspilman:
@dellis1972 - Is this related to https://github.com/MonoGame/MonoGame/issues/5929 ?

dellis1972:
yes

dellis1972:
@tomspilman this will produce

On Windows
`Installers/Windows/MonoGameSetup.exe.sha`

On Mac
`Installers/MonoGame.pkg.sha`
`Installers/Pipeline.MacOS.pkg.sha`
`Installers/monogame-sdk.run.sha`

Installers/default.build
```diff
@@ -34,6 +34,17 @@
         <exec program="makensis" workingdir="Windows" basedir="C:\Program Files (x86)\NSIS">
 		      <arg value="MonoGame.nsi"/>
         </exec>
+        <echo append="false" file="Windows/hash.bat">
+@ECHO OFF
+FOR /f "tokens=*" %%i IN ('@certutil -hashfile %1 %2 ^| find /v "hash of file" ^| find /v "CertUtil"') DO SET r=%%i
+SET r=%r: =%
+ECHO %r%  %1 > %3
+        </echo>
```
tomspilman: Seems like a bit of a dirty trick to do this here.  Why not just create a bat file and add it to the repo?

tomspilman:
@dellis1972 - Are these SHA files still useful once we start signing the Windows installer?

I guess i just post the link to the SHA along side the installer?

Installers/default.build
dellis1972: I can no problem.

dellis1972:
Once we start signing the windows installer probably not (on windows). We should also sign the Mac one too if we can setup an apple account.

Yup we can publish the .sha link next to the link for the installers.

PR #5942, opened 2017-09-14
Add AVPlayer Video Support for MacOS.

dellis1972:
Adds support for playing Video on the MacOS platform using AVFoundation rather than QTKit. 
Also adds new api for getting the native Window Info from SDL.

### DIFF
--- a/MonoGame.Framework/Media/Video.MacOS.cs
+++ b/MonoGame.Framework/Media/Video.MacOS.cs
@@ -6,8 +6,8 @@ using System;
 using System.IO;
 
 using ObjCRuntime;
-using QTKit;
 using Foundation;
+using AVFoundation;
 
 namespace Microsoft.Xna.Framework.Media
 {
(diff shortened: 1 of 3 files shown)


tomspilman:
@dellis1972 - How does this work with DesktopGL?  Do we have a unified video player that works on all platforms or is this a functionality hole we have in DesktopGL we need to fix?

cra0zy:
This will not work with DesktopGL, that being said it's not that hard to get video playback working on DesktopGL using FNA approach.

dellis1972:
@tomspilman yup this is MacOS platform only. But our Mac platform now uses an SDL backend, so this is needed to get the video working. DesktopGL will need another solution, ffmpeg, theora or some kind of factory system which uses the default video player for the platform. Hopefully one that does not require a custom built C++/C dll.

tomspilman:
> Hopefully one that does not require a custom built C++/C dll.

I think that might be unavoidable,  Last time i investigated this Theora was still the only open source video decoding library to use.

Issue #5866, opened 2017-08-12
Missing libpng12 for texture compression on Linux


Jjagg:
Running tests on Linux I got DLLNotFoundException for nvtt (NVidia Texture Tools) because it depends on libpng 1.2. I had libpng installed, but not 1.2. Installing it on Arch Linux from the official repositories was easy and fixed the issue. We should probably include libpng 1.2 as a dependency.

cc @cra0zy

KonajuGames:
Which is nuts because we don't use any PNG functionality in NVTT. I would
really like to get a more specialised DXT encoder that doesn't have the
bloat or inefficiencies of NVTT.

Jjagg:
@rds1983, maybe you are interested in porting [stb_dxt](https://github.com/nothings/stb/blob/master/stb_dxt.h) using Sichem like you have done for stb_image?

That would be a great alternative, giving us a fast, fully managed encoder.

KonajuGames:
The thing to watch with other implementations is the quality of the
encoding. The choice of the high and low color values for each block is
paramount to getting good quality out of DXT.

Jjagg:
Alternatively we could port the relevant parts from NVTT since it's MIT licensed.

rds1983:
Hey, @Jjagg.
Yes, I am interested. stb_vorbis turned out too hard for me(3 dimensional arrays dont like being converted). But stb_dxt is easy one.
So it's already in the StbSharp in those two files:
https://github.com/rds1983/StbSharp/blob/master/StbSharp/Stb.Dxt.Generated.cs
https://github.com/rds1983/StbSharp/blob/master/StbSharp/Stb.Dxt.cs

It should be easy to include that code into MG as StbSharp environment is already there: https://github.com/MonoGame/MonoGame/tree/develop/MonoGame.Framework/Utilities/Imaging

Besides stb_dxt I've also converted method stbgl__compress from there: https://github.com/nothings/stb/blob/9d9f75eb682dd98b34de08bb5c489c6c561c9fa6/tests/caveview/stb_gl.h#L527

As it's only sample of using stb_dxt I've found.
Now my version of that method is there:
https://github.com/rds1983/StbSharp/blob/7787428f684469082fea70e883c6c63c4198ddfb/StbSharp/Stb.Dxt.cs#L42

I've used it to test that StbSharp's DXT compressor makes same result byte-wise as original stb_dxt.
Good luck!

cra0zy:
[...]

PR #5952, opened 2017-09-21
Support Mac Unit Tests

tomspilman:
This PR is a fix to support running Mac unit tests.


tomspilman:
ugh... still something wonky with our develop branch adding these old commits that don't apply.

Still doesn't break this PR.

tomspilman:
Woot... tests are running on the Mac

Jjagg:
Awesome! Lots of failures though. But most are audio.

tomspilman:
@KonajuGames @dellis1972 

Ok... audio tests passing, but we have a few others that still fail.

http://teamcity.monogame.net/viewLog.html?buildId=55098&tab=buildResultsDiv&buildTypeId=MonoGame_TestMac&guest=1

Let me know what needs to be done there and i'll fix them.  Or maybe we should work around them for now?

Jjagg:
> should i be seeing the GameWindow popup on these tests

Nope, there's maybe 1 or 2 tests that actually create a game window, but IIRC they're ignored for DesktopGL.

@dellis1972 The content reference test is failing. It would be nice if it printed some more information. Maybe it can just print the Task Log if the test failed? This test also failed for me locally on Windows with MSBuild. I haven't really looked in to it yet, but anyway seems like it's not 100% right yet. 

About the IntermediateSerializer Colors test: it seems like it tries to parse the whiteline between two tags as content. Since it works on Windows, maybe it's an obscure Mono bug or something?

All other failures are about compression. They are failing because of DllNotFound for NVidia Texture Tools (nvtt). Could be the same issue as I had in #5866 or some other missing dependency. We really don't need nvtt btw (alternative with 0 deps in #5866).

So that's not too bad :D

tomspilman:
So i think i'll just disable these last few failed tests and let someone else come in and fix them in a different PR.

tomspilman:
So the current problem is that the unit tests hang at the end.  I suspect it is crashing, but i cannot be sure.  Needs more investigation.

tomspilman:
@cra0zy @Jjagg @dellis1972 @KonajuGames -  This PR is ready to merge.

Note all other PRs will need to merge/rebase once this is committed else they will not pass the Mac unit tests.

Currently this is only running 36 of our 1200 tests on Mac because we're stuck on an old version of NUnit.  As soon as i work out the configuration error on the Mac build box that keeps the newer version of NUnit from working i can update the build server and all the tests will run.

Still for now lets move forward on this PR.

Test/ContentPipeline/BitmapContentTests.cs
```diff
@@ -174,6 +174,9 @@ public void BitmapConvertFullNoResize()
         }
 
         [Test]
+#if DESKTOPGL
+        [Ignore("NVidia native library could not be loaded on Mac build server.")]
```
dellis1972: @tomspilman what is the problem here? I had these working before.

tomspilman: Don't know... exactly like i said it wasn't loading the library on the Mac build server.

However we have bigger issues to solve first as i cannot run the latest NUnit on the Mac keeping us from running 95% of our tests.

Jjagg: These should work after a rebase/merge because we switched to the managed implementation in #6102.

tomspilman: @Jjagg - But we're switching back to NVTT soon.  https://github.com/MonoGame/MonoGame/pull/6102#issuecomment-368181625

tomspilman:
Bump... lets get this merged and i will work on fixing the mac build server to run all our tests this week.

PR #5955, opened 2017-09-23
Clean up .definition and remove unused files

Jjagg:
Makes the Protobuild .definition files make more use of services rather than specific platforms. Also removes the native MacOS files no longer used since #5410 because Mac is now DesktopGL. 

I'm not sure how to handle the SDL and OpenAL dll's. I would like to include them for all platforms when the DesktopGL service is enabled (so builds are x-platform), but only for their respective platform if only SDLWindowing or OpenALAudio is enabled. I don't think that's possible, right? @hach-que 
It doesn't really matter anyway I suppose since MG isn't designed to let users mix 'n match different backends (e.g. SDL windowing with MG's XAudio audio implementation on Windows), but it would be pretty cool if that was possible.

Probably a good idea to remove al MONOMAC stuff after this.


hach-que:
You can use `<Service>` inside external projects where you would use `<Platform>`, you just need to refer to the full name. I *think* the syntax is something like `<Service Name="MonoGame.Framework/OpenALService">`.

hach-que:
Details here: http://protobuild.readthedocs.io/en/latest/service_dependencies.html#including-external-references

Jjagg:
Oh, right. At this point the binaries are simply included as files instead of as NativeBinary entries. Is it possible to set a custom output path for native binaries (for x64, x86 folders)? @hach-que

hach-que:
Yes. I believe Protogame or the Protogame Editor might have an example in
it's definition files? It's past midnight here though, so I don't want to
go digging for it now.

Jjagg:
Thanks, it was the 'Anchor' attrib I was looking for. Had to upgrade Protobuild to get it to work.

tomspilman:
> Also removes the native MacOS files no longer used since #5410 because Mac is now DesktopGL.

 Is this correct @dellis1972 ?


> but it would be pretty cool if that was possible.

It would be nice if we could support mixing and matching systems, but i'd rather see that effort applied to things like automated testing which currently benefits us much more.

tomspilman:
@Jjagg @hach-que 

So the error on the mac build is this:

```
[14:09:39]"/Applications/buildAgent/work/59a4b66548a5cef5/MonoGame.Framework.MacOS.sln" (Build target) (1) ->
[14:09:39]"/Applications/buildAgent/work/59a4b66548a5cef5/Tools/MGCB/MGCB.MacOS.csproj" (default target) (3) ->
[14:09:39]"/Applications/buildAgent/work/59a4b66548a5cef5/MonoGame.Framework/MonoGame.Framework.MacOS.csproj" (default target) (4:2) ->
[14:09:39](_CopyOutOfDateSourceItemsToOutputDirectory target) -> 
[14:09:39]  /Library/Frameworks/Mono.framework/Versions/5.2.0/lib/mono/msbuild/15.0/bin/Microsoft.Common.CurrentVersion.targets(4422,5): error MSB3025: The source file "/Applications/buildAgent/work/59a4b66548a5cef5" is actually a directory.  The "Copy" task does not support copying directories. [/Applications/buildAgent/work/59a4b66548a5cef5/MonoGame.Framework/MonoGame.Framework.MacOS.csproj]
[14:09:39]  /Library/Frameworks/Mono.framework/Versions/5.2.0/lib/mono/msbuild/15.0/bin/Microsoft.Common.CurrentVersion.targets(4422,5): error MSB3025: The source file "/Applications/buildAgent/work/59a4b66548a5cef5" is actually a directory.  The "Copy" task does not support copying directories. [/Applications/buildAgent/work/59a4b66548a5cef5/MonoGame.Framework/MonoGame.Framework.MacOS.csproj]
```
I forced a new clean build to be sure this is a real project definition error.

It seems like some copy operation is incorrectly formed.  Looking at the mac project on the build server sure enough you can find this:

```
    <None Include="..\..\59a4b66548a5cef5">
      <Link></Link>
      <NativeBinary>True</NativeBinary>
    </None>
    <None Include="..\..\59a4b66548a5cef5">
      <Link></Link>
      <NativeBinary>True</NativeBinary>
    </None>
```

Not sure what change in this PR is causing this to be generated, but this is the bug.

dellis1972:
@Jjagg the MacOS is a hybrid system now. While it does use the SDL windowing system we still use the MacOS native stuff for things like MediaPlayer, Video, VideoPlayer, GamerSerivices etc. So we should not be removing everything.

hach-que:
That sounds like the NativeBinary properties might not be set right. It
shouldn't affect anything, but try back slashes?

https://github.com/RedpointGames/Protogame.Editor/blob/master/Build/Projects/Grpc.Native.definition

Jjagg:
@dellis1972 Oh wow, I got that completely wrong. I'll change that.

tomspilman:
@Jjagg  @dellis1972  @cra0zy - Are we good with these changes now?

Jjagg:
Changes seem good, but build bot didn't test them properly. We should get them tested first.

tomspilman:
I'll force it to rebuild now.

Build/Projects/FrameworkReferences.definition
```diff
@@ -15,6 +15,16 @@
   <Platform Type="Linux">
     <Reference Include="System" />
     <Reference Include="System.Runtime.Serialization" />
+
+    <Service Name="MonoGame.Framework/OpenALAudio">
+      <NativeBinary Path="ThirdParty/Dependencies/openal-soft/Linux/x86/libopenal.so.1" Anchor="x86" />
+      <NativeBinary Path="ThirdParty/Dependencies/openal-soft/Linux/x64/libopenal.so.1" Anchor="x64" />
+    </Service>
+
+    <Service Name="MonoGame.Framework/SDLWindowing">
```
cra0zy: Redundant inclusion? since bellow you specify this stuff in `<Service Name="MonoGame.Framework/DesktopGL">`.

Also no point in defining service for platform references since it only has 1 option and is always gonna be true.

Build/Projects/MonoGame.Framework.definition
```diff
@@ -35,14 +35,14 @@
...
     <CustomDefinitions>
-      <Platform Name="Android">TRACE;ANDROID;GLES;OPENGL;OPENAL</Platform>
-      <Platform Name="iOS">IOS;GLES;OPENGL;OPENAL</Platform>
-      <Platform Name="tvOS">IOS;TVOS;GLES;OPENGL;OPENAL</Platform>
-      <Platform Name="Linux">TRACE;LINUX;OPENGL;OPENAL;DESKTOPGL;SUPPORTS_EFX</Platform>
-      <Platform Name="MacOS">MONOMAC;OPENGL;OPENAL;DESKTOPGL;SUPPORTS_EFX</Platform>
-      <Platform Name="WindowsGL">TRACE;WINDOWS;DESKTOPGL;SUPPORTS_EFX</Platform>
-      <Platform Name="Windows">TRACE;WINDOWS;DIRECTX;WINDOWS_MEDIA_SESSION</Platform>
-      <Platform Name="WindowsUniversal">TRACE;NETFX_CORE;WINDOWS_UAP;DIRECTX;DIRECTX11_1;WINDOWS_MEDIA_ENGINE</Platform>
+      <Platform Name="Android">TRACE;ANDROID;GLES</Platform>
```
cra0zy: A bit of an opinion on this, but it makes sense for ie. Android to have `OPENGL` define, I wouldn't remove any defines, they might need to be used later and they aren't really hurting anybody.

KonajuGames: Agreed.  Common OpenGL code will be under the `OPENGL` define.  OpenGL ES-specific code will be under the `GLES` define.

Build/Projects/FrameworkReferences.definition
Jjagg: Oh, right. I forgot to remove this.

Build/Projects/MonoGame.Framework.definition
Jjagg: I did not remove the define. Android uses the OpenGLGraphics service which defines the OPENGL symbol.

Jjagg:
Closing this because we're gonna drop Protobuild soon.

Issue #5845, opened 2017-07-27
Setting hlsl array elements


Kloda:
It doesn't work to set individual array elements like this:

`effectParameter.Elements[x].SetValue(value);`

If I understand the code correctly, the ConstantBuffer.Update method checks StateKey to see if a parameter needs updating, and it does so for root parameters only, not caring if an array's elements have been set.

PR #5956, opened 2017-09-24
Enable updating of individual hlsl array elements.

yousernaym:
Fix for #5845.
Checks a parameter's `StateKey` in `ConstantBuffer.SetData`, which is called for every array element.


SimonDarksideJ (changes requested):
Can you make this change an overloaded method so as not to break compatibility for existing users please

SimonDarksideJ:
Tests are passing @yousernaym , can you address the feedback above to change this to an overload to ensure it is not a breaking change

SimonDarksideJ:
> > Can you make this change an overloaded method so as not to break compatibility for existing users please
> 
> Are you sure that is required? It's a private method.

It is changing the signature of the existing method, with no default this would break anything currently referencing this method.  The safer course of action is to have an overload with the changes that can be called if need be.

yousernaym:
Sorry @SimonDarksideJ , I didn't see your request before. But not sure I understand the problem. As was pointed out, it's a private method and the code compiles, so what could be referencing it? Leaving the old method will allow for the bug to be re-introduced.

MonoGame.Framework/Graphics/Shader/ConstantBuffer.cs
```diff
@@ -61,8 +61,12 @@ internal void Clear()
             PlatformClear();
         }
 
-        private void SetData(int offset, int rows, int columns, object data)
```
tomspilman, line 64: @SimonDarksideJ @yousernaym 

So there is nothing wrong here.  It is a `private` method internal to MG.  We don't need to create a new method that will never get called.

SimonDarksideJ (approved):
Based on feedback, this is approved

tomspilman:
This change still requires a bit more thought and consideration before merging.

Give me a day or two to think about a possible better solution.

tomspilman:
Didn't forget about this... just been a busy week.

SimonDarksideJ:
Pushing back to 3.8.6 to focus on Vulkan

SimonDarksideJ:
Moving to `3.8.7` to include as part of the [Shader System refactor](#9188)

- Link #9188

PR #5957, opened 2017-09-24
Support DDS cubemaps that don't follow the rules

KonajuGames:
DDS cubemaps should have the DDSCAPS_COMPLEX flag set, but not all DDS writers do this. NormalizeCubeMap.dds from the XNA Racer sample is one of those non-compliant DDS files.


tomspilman:
Thanks @KonajuGames !

PR #5958, opened 2017-09-24
Reset framebuffer after Texture2D.GetData on GL ES

timgott:
GetData left the framebuffer bound to the texture.

Although it is rather unlikely to happen, I think there can still be a problem if a rendertarget was bound before, but I don't know how to fix that.

### DIFF
--- a/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.OpenGL.cs
@@ -212,6 +212,9 @@ namespace Microsoft.Xna.Framework.Graphics
             GL.ReadPixels(rect.X, rect.Y, rect.Width, rect.Height, this.glFormat, this.glType, data);
             GraphicsExtensions.CheckGLError();
             GraphicsDevice.DisposeFramebuffer(framebufferId);
+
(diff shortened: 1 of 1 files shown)


tomspilman:
@KonajuGames @dellis1972 - Not sure what is the right thing here.

I agree that `GetData` should not have any side effects.  It should restore any state it changes.


@atomgott - What would a test case that triggers this bug look like?  Can you add a unit test which exposes the bug so we can develop a good fix?

atomgott:
```
Texture2D tex = new Texture2D(GraphicsDevice, 1, 1);
tex.SetData(new[] { Color.Red });

Color[] data1 = new Color[1];
tex.GetData(data1);

GraphicsDevice.Clear(Color.Green);

Color[] data2 = new Color[1];
tex.GetData(data2);
```
Now `data1[0] == data2[0]` evaluates to true on Windows, but not for Android on current develop:
```
data1[0] {R:255 G:0 B:0 A:255}
data2[0] {R:0 G:128 B:0 A:255}
```

tomspilman:
> This could be implemented as a unit test.

That is great @atomgott !

@KonajuGames @dellis1972 @Jjagg ?

tomspilman:
This is conflicted and we have no unit test.  If this is fixed we can reopen it.

PR #5961, opened 2017-09-25
Multidimensional array content pipeline fix.

### LINKED ISSUE #5960: Multidimensional arrays do not work as intended when contained in a different type.
```csharp
class Foo
{
     object[,] Bars1 { get; set; }
     object[,,] Bars2 { get; set; }
}
```
Does not serialize properly, even if it did, it would not find the Reader at runtime.

Ragath:
Adds tests for the use-case and resolves #5960


dellis1972:
Looks ok to me, would like @tomspilman and @KonajuGames to give it a look over though

tomspilman:
Thanks @Ragath !

PR #5965, opened 2017-09-27
Fixed GamePad.SetVibration crash

MichaelDePiazzi:
In the Windows DirectX build, it's possible for `GamePad.SetVibration` to throw a SharpDX "device not connected" exception if the gamepad was disconnected when `SetVibration` was called.

To reproduce:

1. Create a test app which calls `SetVibration` in every update
2. Plug in a gamepad and run the app
3. Unplug the gamepad
4. Most times the app should throw the exception. If not, try plugging/unplugging the gamepad a few more times.

This was mainly because the `_connected` variable was only being updated by the `GetCapabilities` and `GetState` methods. Because of this, it was also possible for `SetVibration` to do nothing if a gamepad was connected after startup without these methods being called first.

This PR fixes these issues by using the same disconnect timeout logic that the `GetCapabilities` and `GetState` methods use. It also specifically catches the "device not connected" exception as it's still possible for it to happen (just much less often).

I have tested this first in an experimental branch of [my game](http://store.steampowered.com/app/459830/Final_Days/) for 1 week and now in the public branch for the last 3 weeks without any further issues being reported.

### DIFF
--- a/MonoGame.Framework/Input/GamePad.XInput.cs
+++ b/MonoGame.Framework/Input/GamePad.XInput.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Xna.Framework.Input
         {
             // If the device was disconneced then wait for 
             // the timeout to elapsed before we test it again.
-            if (!_connected[index] && _timeout[index] > DateTime.UtcNow.Ticks)
+            if (!_connected[index] && !HasDisconnectedTimeoutElapsed(index))
                 return new GamePadCapabilities();
       
             // Check to see if the device is connected.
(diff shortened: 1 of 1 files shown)


afrodynamics:
I made a similar change to our fork a few months ago to resolve this issue and totally forgot about it until I saw this. We've had this in [our game](http://www.toothandtailgame.com/) since June and it's been working fine. Your fix is more comprehensive though, and should definitely be merged.

PocketwatchGames/MonoGame@009579ad1539c3f73ecb41f7ee0a6f10c3ecb8d6

tomspilman:
Thanks @MichaelDePiazzi !

PR #5974, opened 2017-09-29
Throw a ContentLoadError on the UI thread if we hit one.

dellis1972:
Currently if we hit a content problem on android the error
is swallowed. We need to throw it on the UI thread so that
it will get caught by the app exception handler.


dellis1972:
@mrhelmut this might help with your black screen bug.

mrhelmut:
Going to test this. :)

mrhelmut:
It doesn't, I'm going to take a closer at what could at what could be delaying rendering.

This PR is good to go though!

KonajuGames:
Merging.

Issue #5935, opened 2017-09-13
[Android] Viewport size issues


mrhelmut:
Since #5410, I am running into ```GraphicsDevice.Viewport``` issues when using fullscreen or immersive mode.

It seems that the viewport is too small and offset'd by the size of the action bar.

This is because the initial ```Viewport``` size is based on ```CurrentDisplayMode```, which uses ```AndroidGameView.GameView.Width``` / ```Height```, while ```PreferredBackbufferWidth``` / ```Height``` gets their value from ```ClientBounds``` (set from ```Context.Resources.DisplayMetrics```).

On my device, ```AndroidGameView.GameView``` is 600x1024 while ```Context.Resources.DisplayMetrics``` is 600x976.

I believe this is due to ```MonoAndroidGameView``` being created after the ```Context``` is queried for the display size and then returns values for non-fullscreen mode. Potential fix would be to initialize ```ClientBounds``` after the view is created and using its size instead of the display metrics.

Thoughts @dellis1972 @roponator ?

dellis1972:
no sure what the best approach would be in this case. I'd have to look at the code again.

dellis1972:
@mrhelmut I have a local change which creates the `_clientBounds` using the GameView values and don't see any bad effects in my testing so far.

mrhelmut:
@dellis1972 Would you mind pointing me to this branch? I'll give it a try.  :)

dellis1972:
@mrhelmut https://github.com/MonoGame/MonoGame/pull/5975

dellis1972:
@mrhelmut how are you setting immersive mode? I can get the navigation bar to disappear but no matter what I do the Action bar is always present. and `ActionBar` on the activity is null.

dellis1972:
@mrhelmut never mind. It was me lol

ApmeM:
Hi, sorry for commenting to old issue.

Is there a workaround I can use for 3.6 version (As I can see this fix is not released).
On my Meizu m3s it hide almost half of screen.
And another problem here is that mouse is still full screen, only viewport is broken an it leads to mispointing when clicking on it.

PR #5975, opened 2017-09-29
Calculate the Client Bounds a bit later.

dellis1972:
Fixes https://github.com/MonoGame/MonoGame/issues/5935

### DIFF
--- a/MonoGame.Framework/Android/AndroidGameWindow.cs
+++ b/MonoGame.Framework/Android/AndroidGameWindow.cs
@@ -5,6 +5,7 @@
 using System;
 using Android.Content;
 using Android.Content.PM;
+using Android.OS;
 using Android.Views;
 using Microsoft.Xna.Framework.Input.Touch;
 using MonoGame.OpenGL;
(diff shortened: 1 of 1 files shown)


mrhelmut:
Strangely, ```GameView.Width``` & ```Height``` are both 0 at this moment on my device.

dellis1972:
@mrhelmut hmm. weird :( I wonder why that is. Which device is it? I might try to get a sample running in Test cloud.

mrhelmut:
It's a cheap Qilive tablet running 4.4, but it also happens on Nexus 7 (2013, running 6.0).

```GameView.Width``` is 0 up until here (and it doesn't seem to help knowing why): https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Android/MonoGameAndroidGameView.cs#L321

It probably doesn't point to the real problem and the view initialization is probably hidden by the threads work.

dellis1972:
@mrhelmut well the ClientBounds will get changed again when the SurfaceHolder is created in the SurfaceView. So I guess is will be 0,0 to start with and then gets changed later in the initialisation sequence. It does not seem to cause any problems for me.. but that said it might cause problems else where.

mrhelmut:
I'll try to verify what's going on, right now this PR completely wrecks the rendering.

dellis1972:
@mrhelmut I changed the way we do this.. This should be better now.

Ragath:
fixes #5997

mrhelmut:
Working fine! Thanks @dellis1972

KonajuGames:
Are we supposed to get notifications from the review request that was done 4 days ago?

Merging.  Thanks @dellis1972

Issue #5967, opened 2017-09-27
Pipeline tool crashes when DisplayName is missing


Ragath:
Null reference  exception when DisplayName is missing on ContentImporters, XNA falls back to the typename, just like the ContentProcessors currently do.

dellis1972:
@Ragath this should fix it in theory https://github.com/MonoGame/MonoGame/pull/5976.
Do you have a stack trace?

PR #5976, opened 2017-09-29
Fixed the Pipeline tool to handle Importers without a DisplayName

dellis1972:
Fixes https://github.com/MonoGame/MonoGame/issues/5967


tomspilman:
Thanks @dellis1972 !

Issue #5947, opened 2017-09-17
ScissorRectangle + Viewport renders the wrong part of the screen.


vchelaru:
**Bug**
If GraphicsDevice.ScissorRectangle is combined with GraphicsDevice.Viewport, the resulting area of the screen which is drawn does not match the expected intersection. Furthermore, the portion of the screen does not match the portion of the screen drawn when the code is executed in Microsoft XNA 4.0 (PC).

**Code**
Here is code to reproduce the bug:

```
public class Game1 : Microsoft.Xna.Framework.Game
{
    GraphicsDeviceManager graphics;

    SpriteBatch spriteBatch;
    Texture2D textureFile;

    public Game1() : base ()
    {
        graphics = new GraphicsDeviceManager(this);

        graphics.PreferredBackBufferWidth = 400;
        graphics.PreferredBackBufferHeight = 400;
    }

    protected override void Initialize()
    {
        textureFile = Content.Load<Texture2D>("textureFile");
        spriteBatch = new SpriteBatch(graphics.GraphicsDevice);
        base.Initialize();
    }

    protected override void Draw(GameTime gameTime)
    {
        GraphicsDevice.Clear(Color.Black);
        graphics.GraphicsDevice.Viewport = new Viewport(100, 100, 200, 200);
        var scissorTestEnabled = new RasterizerState();
        scissorTestEnabled.CullMode = CullMode.None;
        scissorTestEnabled.ScissorTestEnable = true;

        graphics.GraphicsDevice.ScissorRectangle = new Rectangle(0, 0, 200, 200);

        spriteBatch.Begin(SpriteSortMode.Immediate, BlendState.AlphaBlend, 
            SamplerState.PointClamp, DepthStencilState.Default, scissorTestEnabled);

        spriteBatch.Draw(textureFile, new Rectangle(0, 0, 400, 400), Color.White);
        spriteBatch.End();

        base.Draw(gameTime);
    }
}
```

**Results**
MonoGame Desktop GL (rendered red portion is too far down):
![image](https://i.imgur.com/GptY6dB.png)

XNA (correctly rendered):
![image](https://i.imgur.com/DR8Myuq.png)

Image Breakdown (XNA, correct rendering):
![image](https://i.imgur.com/hMsaeG0.png)

**Version Info**
This was tested on MonoGame Desktop GL 3.6 and 3.7 (build pulled at the beginning of August). Also, this was tested on Android 3.6.

tomspilman:
[...]

PR #5977, opened 2017-09-29
Fixed the Scissor rect calculation on DesktopGL and OpenGL platforms

dellis1972:
Fixes https://github.com/MonoGame/MonoGame/issues/5947

Scissor rectangle seems to be calculated always from the window, not from the current viewport.


tomspilman:
@vchelaru - This looks correct to you?  Does it fix your bug?

vchelaru:
Yes, the code and test look good. I don't know if I'll get to test it anytime soon though, because I've implemented a temporary workaround and moved on to a different project. 

I'll add a todo for myself in my trello board to check it out, but if it passes the test I say 👍

tomspilman:
Thanks @dellis1972 and @vchelaru !

vchelaru:
Apologies for the late reply but I ran tests on my end and it seems to have fixed the problem for me too.

PR #5979, opened 2017-09-29
Fix Duplicate/Weird Menus on MacOS

dellis1972:
Eto Forms seems to default `IncludeSystemItems` to `All`.
As a result the deafult system menus like `File`, `Edit`, `Help`
etc are populated by default on MacOS.

Even though we set the property later they are not cleaned out.
So we end up with duplicates and messy menus.

So we subclass the `MenuBar` and set the property to our required
default first.

### DIFF
--- a/Tools/Pipeline/MainWindow.eto.cs
+++ b/Tools/Pipeline/MainWindow.eto.cs
@@ -11,6 +11,21 @@ namespace MonoGame.Tools.Pipeline
 {
     partial class MainWindow
     {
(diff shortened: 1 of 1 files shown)


dellis1972:
@cra0zy can your review this please :)

dellis1972:
@KonajuGames @tomspilman when you get a moment :) This fixes a rather unsightly problem on Mac.

KonajuGames:
Merging.

PR #5985, opened 2017-10-03
Included mgcb autocomplete for bash

cra0zy:
Most of the autocomplete stuff done, the only thing left is to setup a bridge between bash and MGCB so that processors/importers/processorParams can also get autocompleted, but I'll leave that for a future PR.

Preview:
autocomplete

CC. @dellis1972 (pssss, I think the file should work on Mac as well)


tomspilman:
Thanks @cra0zy !

PR #5993, opened 2017-10-07
[OpenGL] Rework Android Framebuffer Support

dellis1972:
OpenTK removal also removed some support for frame buffers
on older devices. The old code was a bit of a mess with
mutliple FrameBuffer classes each doing the same thing.

This commit reworks the code to have one OpenGL FramebufferHelper
and moves all the loading of OpenGL extensions into the OpenGL.cs
class where they belong.

As a result GraphicsCapabilites was reworked to use the `Extensions`
property on the `GL` class. We also needed to rework when the
OpenGL entry points are loaded so ensure we have a current context.
If we don't we cannot get the extension strings from opengl since
it required a current context.

### DIFF
--- a/MonoGame.Framework/Android/MonoGameAndroidGameView.cs
+++ b/MonoGame.Framework/Android/MonoGameAndroidGameView.cs
@@ -68,6 +68,7 @@ namespace Microsoft.Xna.Framework
         private readonly Game _game;
 
         const int EglContextClientVersion = 0x3098;
+        const int EglContextCMinorVersion = 0x30fb;
 
         // Events that are triggered on the game thread
         public static event EventHandler OnPauseGameThread;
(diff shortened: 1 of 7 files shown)


dellis1972:
@mrhelmut this might fix some of the black screen issues you are having on android. I have improved Render Target support as well as the selection of the display config for android.

dellis1972:
Oh it will also create a GLES 3.x context if it can and fallback to GLES 2.0 if it has too.

nkast:
>  I have improved Render Target support as well as the selection of the display config for android.
> Oh it will also create a GLES 3.x context if it can and fallback to GLES 2.0 if it has too.

Does this mean support of MRT on Android?

dellis1972:
@nkast

I assume you want to be able to write to two targets from a shader? i.e a Color and Depth buffer, or a Normal buffer? If so that might be GLES 3.x only. I know GLES 2.0 does not support it unless via extensions. 

The texture formats are quite also limited on GLES. float textures are not really supported afaik, again maybe in GLES 3.x

mrhelmut:
It doesn't fix my black screen issue but it's working fine and is worth merging!

KonajuGames:
Ok. Merging.  Thanks @dellis1972

PR #5995, opened 2017-10-08
Use short for parameter count in effect binary

### LINKED ISSUE #5994: HLSL array size limited to 255
MG uses a byte in the binary effect format to store the number of parameters in an effect (written out [here](https://github.com/MonoGame/MonoGame/blob/78e193e/Tools/2MGFX/EffectObject.writer.cs#L143)). In the effect writer, the function to write parameters is called recursively for an array, treating each element as a parameter. Because of this the number of elements in that array is also written out as a byte. The number of elements is internally stored as an integer and only casted to byte (without checking for overflow) when the effect is written out. So when users have arrays of > 255 elements, the build will succeed, but things will go wrong when the effect is read back in at runtime.

I locally patched this to write out and read in the parameter count as an integer and that fixed this. If that's alright I'll set up a PR. I know I have to upgrade the version in [2MGFX](https://github.com/MonoGame/MonoGame/blob/78e193e/Tools/2MGFX/EffectObject.writer.cs#L15) and the [runtime](https://github.com/MonoGame/MonoGame/blob/78e193e288672361260d82136ef2622a590c8e55/MonoGame.Framework/Graphics/Effect/Effect.cs#L28), and rebuild stock effects afterwards.

EDIT: It seems parameter/element count is limited to 2^16, so using a short would be better.

Jjagg:
A byte was used for the parameter count previously, this limited arrays to 255 elements. Fixes #5994.


Tools/2MGFX/EffectObject.writer.cs
```diff
@@ -140,7 +140,7 @@ public void Write(BinaryWriter writer, Options options)
 
         private static void WriteParameters(BinaryWriter writer, d3dx_parameter[] parameters, int count)
         {
-            writer.Write((byte)count);
+            writer.Write((short) count);
```
nkast: How about ` writer.Write7BitEncodedInt(count)`?

nkast: Should we update those two just in case?
https://github.com/Jjagg/MonoGame/blob/cb24746589835f0d4e2b7c041b6b0c8659979737/Tools/2MGFX/EffectObject.writer.cs#L159-L160

nkast:
Should we update those two just in case?
https://github.com/Jjagg/MonoGame/blob/cb24746589835f0d4e2b7c041b6b0c8659979737/Tools/2MGFX/EffectObject.writer.cs#L159-L160

Tools/2MGFX/EffectObject.writer.cs
Jjagg: Yeah, I'll change that. It might save a couple bytes since most of the time count will fit in one byte. Thanks for the heads up, I didn't know about that function :)

Jjagg:
That's not necessary, row and column are never more than 4 (we could even encode them in a single byte actually).

nkast:
Oh, I see. you are right.

Jjagg:
`Read7BitEncodedInt` and `Write7BitEncodedInt` are protected, so I had to create subclasses to use them in a proper way. Not sure why they're protected, I couldn't find a good explanation anywhere. Anyway, just subclassing seemed like a good solution to me. This is good to go as far as I'm concerned.

nkast:
Interesting, I didn't know that. The 7BitEncodedInt is used in the ContentWriter for some parts of XNB.
In the XNB Format documentation:
 
This format matches the .NET BinaryReader.Read7BitEncodedInt method. It is a variable size encoding of a 32 bit integer value. C implementation:
`
int Read7BitEncodedInt()
{
    int result = 0;
    int bitsRead = 0;
    int value;

    do
    {
        value = ReadByte();
        result |= (value & 0x7f) << bitsRead;
        bitsRead += 7;
    }
    while (value & 0x80);

    return result;
}
`

Jjagg:
Oh, cool! I could just implement it like that if that's preferable.

nkast:
Well, it doesn't give the implementation of Write7BitEncodedInt(), something that you need to figure out. 
Anyway will work IMO.

Jjagg:
Yeah, I think this is fine as is. MG also uses the 7 bit encoded int methods in the ContentReader and ContentWriter classes, but they extend BinaryReader and BinaryWriter, so they also use the .NET implementation.

Jjagg:
@tomspilman @cra0zy This is still good to go :) Heads up, it updates the MGFX version. That's important for release notes.

mrhelmut:
Will the mgfxo be up-to-date, no rebase needed? (I haven't been following much the repo history lately, just a safe check. :) )

Jjagg:
@mrhelmut Yes, the MGFX version has not been updated by another PR (it's still 8 in develop) which means the binary format has not changed since this branch was created. I recompiled all effects in my last commit, so the .xnb files are up to date :)

PumpkinPudding:
https://github.com/Jjagg/MonoGame/blob/0c416a94b9e4b256be0c8912ad6f49218a67be3f/MonoGame.Framework/Graphics/Effect/Effect.cs#L244

var sizeInBytes = (int)reader.ReadUInt16();

Kosmonaut3d:
A checked out version of this will still crash when using arrays with size > 1541

ShaderCode (crash)
#define SIZE 1542
float indexList[SIZE ];

ShaderCode (will work)
#define SIZE 1541
float indexList[SIZE ];

sevenstreak:
Looks like Jjagg's commits are effective refactoring. I'd support pushing to base.

Jjagg:
Yes, I don't remember looking into the 1541 limit for arrays mentioned above, but this still at least improves the limit. This PR is old, but still good to go.

Jjagg:
@harry-cpp This one is good to go!
Need to put in the release notes for 3.8 that people should recompile their effects after upgrading because the format changed.

PR #5996, opened 2017-10-08
Fix property content serialization detection when using a property named `Item`.

joncloud:
* Adjusts indexer property detection to check for Index Parameters instead of get/set method's parameters.
* Moves test compliation in order to share between tests.
* Creates new tests to verify changes.


joncloud:
I did notice that `ReflectionHelpers.GetPropertySetMethod` appeared to be no longer in use. I wasn't sure if I should remove it or not.  Also `ReflectionHelpers.GetPropertyGetMethod` is only being used within the `ReflectionHelpers` type.

joncloud:
I should be able to look into the failing build checks today.  I will need to setup locally to diagnose the failures.

joncloud:
Resolved the build errors by rolling back C# syntax to a more compatible version.

tomspilman:
@joncloud - Thanks...  i'll make sure this gets reviewed and merged here shortly.

tomspilman:
For a future PR it might be better for `ArrayContentCompilerTest.cs` and `XmlContentCompilerTest.cs` to be merged and moved into a new `ContentCompilerTest.cs` as that is what these both are.  

Don't see a huge need to split this all up.

joncloud:
Got it - Do you want me to merge the two together on this PR, or setup another PR to do that?

tomspilman:
@joncloud - If you have the time and could do that in this PR it would be great.  If not it can wait for another PR.

joncloud:
OK I should be able to get the two merged on this PR. I just need to setup my test environment.

tomspilman:
Thanks @joncloud !  Merging!

PR #6003, opened 2017-10-12
Compressed texture fallback formats

### LINKED ISSUE #5983: [Android] Pipeline failing to check format
Hello,

When trying to compress textures for an Android target, I noticed that depending on the source image format, the pipeline doesn't check for the correct requirement.

Using ETC1 compression, and depending on the PNG encoding, a 1920x1080 texture may build fine, or fail with a "must be power of 2" error.

Either the pipeline fails to check the requirements, or the error message is misleading and something else happens while processing the source file.

Here are example files: https://www.dropbox.com/s/l7ly2vpcom7ebql/selfies.zip?dl=0

Both are PNG 24bit, one works, the other fails, the only difference being the internal DPI of the files.

KonajuGames:
This PR fixes some cases where ETC1 fallback wasn't working as expected, and expands the fallback support to allow NPOT and non-square textures to fall back to 16-bit formats instead of failing.

More unit tests added for texture compression.

If no rect is passed to Texture2D.SetData() that takes an optional rect, it passes the call to the full texture PlatformSetData instead for performance reasons.

~~Remove a try..catch on Android that was hiding ContentLoadException thrown during content load.~~

Fixes #5983

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/DefaultTextureProfile.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/DefaultTextureProfile.cs
@@ -118,23 +118,23 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
             switch (format)
             {
                 case TextureProcessorOutputFormat.AtcCompressed:
-                    GraphicsUtil.CompressAti(content);
+                    GraphicsUtil.CompressAti(context, content, isSpriteFont);
                     break;
 
                 case TextureProcessorOutputFormat.Color16Bit:
-                    GraphicsUtil.CompressColor16Bit(content);
+                    GraphicsUtil.CompressColor16Bit(context, content);
                     break;
 
                 case TextureProcessorOutputFormat.DxtCompressed:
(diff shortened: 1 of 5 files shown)


MonoGame.Framework/Android/MonoGameAndroidGameView.cs
```diff
@@ -656,14 +656,7 @@ void UpdateAndRenderFrame()
...
 
-            try
-            {
-                UpdateFrameInternal(updateEventArgs);
-            }
-            catch (Content.ContentLoadException)
-            {
-                // ignore it..
-            }
+            UpdateFrameInternal(updateEventArgs);
```
dellis1972: This is handled on https://github.com/MonoGame/MonoGame/pull/5974

dellis1972: If you just let the exception raise and we are rendering on a separate thread, the app will not stop. So we really need to throw this on the UI thread (if needed)

KonajuGames:
Merged #5974 and rebased this PR.

PR #6004, opened 2017-10-12
Optimizations to Length and Normalize in Vector3 and Vector4

nkast:
Length() and LengthSquared() was calcualated through
DistanceSquared(this, zero).

### DIFF
--- a/MonoGame.Framework/Vector3.cs
+++ b/MonoGame.Framework/Vector3.cs
@@ -568,7 +568,7 @@ namespace Microsoft.Xna.Framework
         /// <returns>The length of this <see cref="Vector3"/>.</returns>
         public float Length()
         {
-            float result = DistanceSquared(this, zero);
+            float result = this.LengthSquared();
             return (float)Math.Sqrt(result);
         }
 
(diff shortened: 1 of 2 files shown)


nkast:
Also note that in Vector2 the sqLength calculation is inlined for Length() & Normalize()
and Normalize() code is sightly different between Vector3 & Vector4.

MonoGame.Framework/Vector3.cs
```diff
@@ -568,7 +568,7 @@ public static void Hermite(ref Vector3 value1, ref Vector3 tangent1, ref Vector3
         /// <returns>The length of this <see cref="Vector3"/>.</returns>
         public float Length()
         {
-            float result = DistanceSquared(this, zero);
+            float result = this.LengthSquared();
```
tomspilman: I say go ahead and inline the length squared calculation here.

MonoGame.Framework/Vector3.cs
```diff
@@ -791,7 +791,7 @@ public void Normalize()
         /// <returns>Unit vector.</returns>
         public static Vector3 Normalize(Vector3 value)
         {
-            float factor = Distance(value, zero);
+            float factor = value.Length();
```
tomspilman: Go ahead and inline these too like in Vector2.

tomspilman:
Good catch @nkast .  This performance bug has been in MG since the first commit to GitHub 7 years ago.

tomspilman:
Thanks @nkast !

PR #6005, opened 2017-10-13
 Simplified Plane.Normalize()

nkast:


tomspilman:
@nkast - Thanks...  I eye balled these changes and they seem to do about the same as before.

However looking at our unit tests we don't test `Plane` very well at all...

 https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/PlaneTest.cs

If you get some time a new PR with a few more unit tests will help us ensure we're correct and completely XNA compatible.  Like i wonder what happens if the plane has a zero length normal during normalize?

nkast:
The PlaneTest already uses Normalize as part of other tests. I already run the tests to see if I needed to add some ε but surprisingly it returns exactly the same values.  but maybe we should spit then and each method get it's own test.
As for a plane with zero normal I did't think about it. similar methods like Vector3.Normalize() return NaN. After all that won't be a valid Plane to begin with.

First thing is to add a test for Matrix.CreateReflection() which uses Plane.Normalize(). I am very skeptical if that's correct.
I don't really see what's the use of this method really! It would be more useful to multiply D with the Length, which results in the same plane. However the tests, DirectXMath and a couple of third party game engines implement it this way.

tomspilman:
> As for a plane with zero normal I did't think about it. 

Right... that is my question.  Does normalizing a plane with zero length return NaN on XNA?  I don't know if it does or if it doesn't.  I just see that we're missing that test case.

PR #6006, opened 2017-10-13
Matrix.CreateReflection() normalizes/alters the *input* value

nkast:
The normalized value is not used in the Method after that.
In the beginning it creates a normalized copy of the plane.


tomspilman:
Yep... good catch.  Merging!

PR #6008, opened 2017-10-13
Use StbSharp for all Texture2D.FromStream

### LINKED ISSUE #6001: Fix Texture2D.FromStream support for 1bit PNGs
Ran across a 1bpp PNG image that our existing `Texture2D.FromStream` would not load.

We should add this format to the unit tests and ensure it works.

You can download this white 1bpp texture here:  

  https://user-images.githubusercontent.com/103522/31480817-aafc0bde-aee5-11e7-9cb5-77491b4d3ae2.png

KonajuGames:
StbSharp is now used for all implementations of Texture2D.FromStream(GraphicsDevice, Stream), avoiding discrepancies with OS-provided image loading routines.

As a result of this, DirectX platforms no longer support TIF or DDS through FromStream().

Added unit tests for loading 1-, 8-, 24- and 32-bit PNGs.

Fixes #6001


rds1983:
Also, all images will be converted to 32-bit on loading as Stb is called with Imaging.STBI_rgb_alpha. It could be avoided by using Imaging.STBI__default, then images will be loaded with their original channels per pixel. But then a conversion code will be required to convert image to the corresponding SurfaceType.

tomspilman:
> Also, all images will be converted to 32-bit on loading 

That should be correct.  If i remember correctly  XNA only returned `Color` images from `FromStream` regardless of the input format of the image.

KonajuGames:
That is correct. I tested XNA with 1-, 8-, 24- and 32-bit PNGs and it
returns each with a texture with SurfaceFormat.Color, so the current
behaviour is correct. I added the optimisation that it checks for zero
alpha only on images that had less than four channels in the source image.

rds1983:
I see. Thanks for the explanation.

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -205,26 +206,36 @@ private int CalculateSubresourceIndex(int arraySlice, int level)
...
-            // http://stackoverflow.com/questions/9602102/loading-textures-with-sharpdx-in-metro 
-            Texture2D toReturn = null;
-            SharpDX.WIC.BitmapDecoder decoder;
+            // The data returned is always four channel BGRA
+            var data = reader.Read(stream, out width, out height, out channels, Imaging.STBI_rgb_alpha);
 
-            using (var bitmap = LoadBitmap(stream, out decoder))
-            using (decoder)
+            // XNA blacks out any pixels with an alpha of zero.
+            if (channels == 4)
```
tomspilman: This should be unnecessary as we always read 4 channels?

tomspilman: Oh wait... is this the original number for channels before conversion to RGBA?  If so i see this is just here to avoid doing the black out if the data was 24bit and didn't have alpha to begin with.

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
```diff
@@ -205,26 +206,36 @@ private int CalculateSubresourceIndex(int arraySlice, int level)
...
+                fixed (byte* b = &data[0])
+                {
+                    for (var i = 0; i < data.Length; i += 4)
+                    {
+                        if (b[i + 3] == 0)
+                        {
+                            b[i + 0] = 0;
+                            b[i + 1] = 0;
+                            b[i + 2] = 0;
+                        }
```
tomspilman: Do we know if this is necessary with STB?  We should determine if it internally does this already to avoid the extra work.

amerkoleci:
Agree, why adding extra dependency if we can use WIC on windows and the corresponding platform code?

KonajuGames:
> Agree, why adding extra dependency if we can use WIC on windows and the corresponding platform code?

That's what we are trying to avoid.  We want one managed codebase that works on all platforms.

MonoGame.Framework/Graphics/Texture2D.DirectX.cs
KonajuGames: XNA was unique in this regard by partially premultiplying alpha (only where A = 0).  STB returns the data as in the original PNG.  In my test case, this was RGB = 0xFFFFFF and A = 0.

afrodynamics:
Would love to see this merged! Most of our textures are loaded via FromStream directly from PNGs, but it seems some driver vendors do not support this and [crash instead](https://cdn.discordapp.com/attachments/134810918113509376/376772075957780480/IMG_20171105_173723.jpg). The less we can depend on driver vendors the better, IMO.

rds1983:
Sorry, my PR #6131 created conflict with this one.
Hovewer should be easy to resolve as it affected only MonoGame.Framework.definition(I've removed TinyJpegSharp from MG).

dellis1972:
@KonajuGames can you rebase this?

KonajuGames:
This PR is now rebased and passing again.
@dellis1972 @tomspilman

dellis1972:
looks good, @tomspilman

tomspilman:
Thanks @KonajuGames !

Issue #5998, opened 2017-10-10
UWP / Xbox games throw an exception & won't launch in newer 3.7.0.xxx builds


YugeshN:
I have a cross platform game I am working on (OpenGL, Android and XB1). It's been working fine on versions up to 3.7.0.716

However, versions 3.7.0.966 and above (not sure how much earlier than 966 the problem was introduced... but I had installed 966 a while back and hadn't tested the game on XB1 until recently) now cause the game to crash on XB1 when launching. 

I also tested the most recent build (3.7.0.990) and the problem is still there. The specific stack trace I get it:
    
    Microsoft.Xna.Framework.Graphics.NoSuitableGraphicsDeviceException occurred
      HResult=0x80131500
      Message=Failed to create graphics device!
      Source=<Cannot evaluate the exception source>
      StackTrace:
       at Microsoft.Xna.Framework.GraphicsDeviceManager.CreateDevice()
       at Microsoft.Xna.Framework.Game.DoInitialize()
       at Microsoft.Xna.Framework.Game.Run(GameRunBehavior runBehavior)
       at MonoGame.Framework.XamlGame`1.Create(String launchParameters, CoreWindow window, SwapChainPanel swapChainPanel)
       at MyGame_UWP.GamePage..ctor() in D:\Dev\csharp\game\mygame\mygame-
    uwp\GamePage.xaml.cs:line 30
       at MyGame_UWP.MyGame_UWP_XamlTypeInfo.XamlTypeInfoProvider.Activate_0_GamePage() in D:\Dev\csharp\game\mygame\mygame-uwp\obj\x64\Debug\XamlTypeInfo.g.cs:line 202
       at MyGame_UWP.MyGame_UWP_XamlTypeInfo.XamlUserType.ActivateInstance() in 
    D:\Dev\csharp\game\mygame\mygame-uwp\obj\x64\Debug\XamlTypeInfo.g.cs:line 411

    Inner Exception 1:
    DllNotFoundException: Unable to load DLL 'ole32.dll': The specified module could not be found. 
    (Exception from HRESULT: 0x8007007E)
    

Note, if I re-install 3.7.0.716 and rebuild, it works just fine on XB1, so the problem is clearly introduced in one of the versions after 716.

Note, this is the windows version of the Monogame SDK with Visual Studio Community 2017.

nkast:
Here is the list of recent builds. Can you test a couple of those?
[...]

PR #6010, opened 2017-10-14
Use desktop and UWP specific assemblies where necessary

Jjagg:
Depends on MonoGame/MonoGame.Dependencies#115. 

Because we used the .NET Standard assemblies, some conditionally compiled code that we need was not included (see the issue in .Dependencies for details). This makes MG use the assemblies compiled specifically for either UWP or desktop where needed.

This should fix issue #5998

### DIFF
--- a/Build/Projects/2MGFXReferences.definition
+++ b/Build/Projects/2MGFXReferences.definition
@@ -4,7 +4,7 @@
   <Platform Type="Windows">
     <Binary
       Name="SharpDX"
-      Path="ThirdParty/Dependencies/SharpDX/SharpDX.dll" />
+      Path="ThirdParty/Dependencies/SharpDX/net40/SharpDX.dll" />
     <Binary
       Name="SharpDX.D3DCompiler"
       Path="ThirdParty/Dependencies/SharpDX/SharpDX.D3DCompiler.dll" />
--- a/Build/Projects/Framework.Content.Pipeline.References.definition
+++ b/Build/Projects/Framework.Content.Pipeline.References.definition
@@ -81,7 +81,7 @@
   <Platform Type="Windows">
     <Binary
       Name="SharpDX"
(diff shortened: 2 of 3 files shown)


Jjagg:
Not sure what the error is all about. Things are good locally :/

Jjagg:
Oh, Windows DX still targets .NET 4.0 and I included the SharpDX assemblies that target .NET 4.5. I think that's what causes the error.

Jjagg:
MonoGame/MonoGame.Dependencies#115 and this one are good to go! @tomspilman

tomspilman:
@KonajuGames @dellis1972 @DDReaper - Can you guys review this?  I want to be sure we got things right this time.

Jjagg:
I thought that Windows DX templates still targeted .NET 4.0 (I think that was for Windows XP support), but that's not the case anymore, they target .NET 4.5 as well now. 

.NET Standard 1.1 can only be used in projects targeting .NET Framework 4.5 and up, so it won't be possible to build for 4.0 anymore. We can use all 4.0 assemblies from SharpDX targeting 4.0 for now if that's an issue, but SharpDX is no longer building future versions for .NET 4.0 IIRC so this may be harder to do in the future if we want to keep it up longer.

Jjagg:
@DDReaper @dellis1972 @KonajuGames Can anyone review/test this? Xbox UWP is currently broken and this should fix it.

MichaelDePiazzi:
Just tested this on both Windows 7 and 10 (v1709) and all is working as expected for me. I tested this using my game and confirmed that sound effects, music, and controllers are all working. I can confirm that this fixes the issue I reported (#6083).

Jjagg:
@dellis1972 Could you merge this too. It's good to go since you merged MonoGame/MonoGame.Dependencies#115.

PR #6011, opened 2017-10-15
Fix Pipeline Tool compilation with msbuild

cra0zy:
Turns out that if you have 2 embedded resources that depend on one .cs file, and you compile that, msbuild deletes the 2nd embedded resource and makes the link point to the first one, definitely sounds like a bug to me.


cra0zy:
@dellis1972 @KonajuGames @tomspilman Ping.

dellis1972:
@cra0zy I've seen something like this before. Its happened  when I use a symlink to link the MonoGame Directory to another location. When building in the normal folder its ok, when in the symlinked folder it removes the EmbeddedResource prefix. Not sure if this is related though.

cra0zy:
@dellis1972 @KonajuGames @tomspilman Re-ping... its just a resource name change.

cra0zy:
@cra0zy sad :(

cra0zy:
Yess, thank you @dellis1972

PR #6013, opened 2017-10-18
Expose ContentIdentity of the source asset inside ContentProcessorContext.

### LINKED ISSUE #6014: Getting the ContentIdentity for an asset over to the ContentProcessor takes a lot more work than needed.
Currently you need to pass it along with the importer output, this requires creating a new type, usually based on ContentItem and specify a separate runtime type. If the type in question comes from a third party library, which is fairly common in importers, you also need to re-implement that type and transfer the data over in order to allow for the potential of processor-less content builds. If the content pipeline could transfer the identity of the source asset separately from the output of the importer, all this work would become unnecessary.

Ragath:
Exposes the ContentIdentity of the imported asset-file to reduce the need for special types in importers. The information was already available in the ProcessorContext, this just brings it up to the surface so extension authors can benefit. The main usecase is when adding external references or building dependencies during content processing. Resolves #6014

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/PipelineProcessorContext.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PipelineProcessorContext.cs
@@ -34,6 +34,8 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
 
         public override ContentBuildLogger Logger { get { return _manager.Logger; } }
 
+        public override ContentIdentity SourceIdentity { get { return new ContentIdentity(_pipelineEvent.SourceFile); } }
+
         public override void AddDependency(string filename)
         {
             _pipelineEvent.Dependencies.AddUnique(filename);
--- a/MonoGame.Framework.Content.Pipeline/ContentProcessorContext.cs
+++ b/MonoGame.Framework.Content.Pipeline/ContentProcessorContext.cs
@@ -27,6 +27,11 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         public abstract ContentBuildLogger Logger { get; }
 
         /// <summary>
+        /// Gets the ContentIdentity representing the source asset imported.
+        /// </summary>
+        public abstract ContentIdentity SourceIdentity { get; }
+
+        /// <summary>
         /// Gets the output path of the content processor.
         /// </summary>
         public abstract string OutputDirectory { get; }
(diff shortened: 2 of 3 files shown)


Ragath:
Here's an example of the type of workaround needed today, that this would deprecate https://github.com/Ragath/TileRenderer/blob/master/TileRenderer.Pipeline/TiledMapImporter.cs

dellis1972:
@KonajuGames @tomspilman thoughts?

tomspilman:
So what you want to be able to do is get the filename of the source file from a custom content processor without the need to write a custom importer?  Am i correct?

Technically this is fine...  `PipelineProcessorContext` is created once per content item, so it won't hurt any progress on threading the pipeline.

It does sort of violate the design of the pipeline to separate importers and processors.  However i've yet to see any huge flexibility benefits from the importer/processor separation.  99% of the time an importer is matched to the same processor.  The only exceptions are in some XML processing and with `SoundEffect` vs `Song` processing.

So i don't think this hurts anything.

@KonajuGames ?

Ragath:
@tomspilman it does indeed increase the surface area of the contract between the importer and processor.

Currently it only covers the Type imported. But this is also a weak coupling, so its more a case of "if the importer does indeed load content from the source asset provided, the provided identity can be used to identify the source file". Then again, that contract is already established by the architecture, an importer expects an existing input file and thus the "source asset" could never really be anything other than the value provided to the filename parameter of the import method.

I could see this posing a problem if we where to change the architecture to let multiple importers participate in the import step when building of one *.xnb(aka. providing multiple inputs to the processor). But I dont see us moving in that direction and even if we did, it would involve massive changes to the pipeline as a whole, making this a drop in the bucket.

tomspilman:
@Ragath - I didn't see an answer to my first question.  All you want is the ability to get the filename of the importer source file from a custom content processor correct?

Ragath:
@tomspilman yes, thats correct, the goal is to access the source path from the processor and the context seemed like the ideal place to put it.

tomspilman:
@KonajuGames - I'm ok with this if you are.

It could maybe use a unit test to ensure it is working correctly, but we sort of lack unit tests for the `PipelineProcessorContext` all together right now.

Ragath:
PipelineProcessorContext is a bit tricky to test, since its just a container of state and doesn't really do much. What we could test is the content builder and ensure it fills the pipeline events(which the context wraps) with the correct values, but that seems a bit out of scope for this PR.

tomspilman:
>  but that seems a bit out of scope for this PR.

Exactly.

KonajuGames:
Merging. Thanks @Ragath

PR #6019, opened 2017-10-23
Added buffered KeyDown & KeyUp events

Jure-BB:
I know this is not par of XNA's API, but since we already have TextInput I also added KeyDown & KeyUp events. Like TextInput they are buffered (no missed keys if framerate is low). This two events can be for example used to move cursor inside text box with arrow keys or with 'Home' and 'End' keys.


MonoGame.Framework/Input/KeysHelper.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
```
Jjagg: Can you add the license text here?

PR #6021, opened 2017-10-25
Generate portable dll powershell script

Jure-BB:
There isn't a lot of info available on how can we generate a portable MG dll from source ourselves. 

This Powershell script calls Piranha.exe with necessary command to generate portable MG dll.


Jure-BB:
No longer relevant as Piranha is getting removed.

PR #6032, opened 2017-10-30
Dont override the ContentOutput directories if the user has defined them.

dellis1972:
We currently always default the Content output directories to
values we define. But we should get the metadata from the
items and use that if its present.


dellis1972:
@lateralusX :)

KonajuGames:
I'm good with that.

PR #6033, opened 2017-10-30
[Pipeline Tool] Implementing Auto Build feature

raizam:
I had to modify the pipeline tool in order to build assets automaticaly on change.

To enable this feature, a checkbox has to be checked in the build menu:
autobuild

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.FileWatcher.cs
+++ b/Tools/Pipeline/Common/PipelineController.FileWatcher.cs
@@ -157,14 +157,14 @@ namespace MonoGame.Tools.Pipeline
                                 if (modifiedItem != null && modifiedItem.Item1.Exists == false)
                                 {
                                     modifiedItem.Item1.Exists = true;
-                                    _view.Invoke(() => _view.UpdateTreeItem(modifiedItem.Item1));
+                                     _view.UpdateTreeItem(modifiedItem.Item1);
                                 }
                             }
 
                             var finalList = modifiedItems.Distinct().ToArray();
 
                             if (finalList.Length > 0)
-                                _view.Invoke(() => _controller.BuildItems(finalList, false));
+                                _controller.BuildItems(finalList, false);
                         }
                     });
 


Jjagg:
So this will trigger a build every time you modify an asset? I wouldn't use this feature, but maybe others see value in adding it.

raizam:
@Jjagg Why wouldn't you?
I wanted to automate things for productivity. I've also made a custom ContentManager which reloads .xnb when they are updated. Now when I modify a spritesheet from photoshop, it updates straight to the running game, so I can see it animated.
I believe these kind of features should be provided out of the box.

Jjagg:
Oh, that's pretty cool! I tend to update assets in batches, but I get where this is coming from now. Would be neat to have this stuff in the core for quick iteration!

tomspilman:
Just a couple of concerns/comments.

What happens with multi-file content like models?  You're in the middle of saving a model out of 3dsmax and the auto build kicks off after the FBX is written but before the textures are saved.  So that fails in some gross way i guess?

I was already fairly worried about the scalability of the "Exists" code that was there... this seems possibly worse.  Would it not be better to create one `FileSystemWatcher` for the entire project folder then check changes against the content items only when they change?

@KonajuGames @dellis1972 @cra0zy - Thoughts?

raizam:
Hi @tomspilman thank you for taking the time to look at this.

> What happens with multi-file content like models?

In this situation there will me issues indeed.
If multiple assets are saved within that rate, it will trigger a single build, otherwise it will call ```pipelineController.Build(false)``` twice without checking if the previous build is finished, and this will cause an assert: [assert](https://github.com/raizam/MonoGame/blob/32706ae1e416b8f42d621d6223fe7b4f39862d8a/Tools/Pipeline/Common/PipelineController.cs#L455) in PipelineController.cs

So I first need to check a build task isn't occuring before launching another one.
The second thing that should be done, is to build files individualy rather than launching a project build.
Would this be a good approach?

> Would it not be better to create one FileSystemWatcher for the entire project folder then check changes against the content items only when they change?

My first attempt was to use a single FileSystemWatcher with IncludeSubdirectories = true, however in my case the folder containing the assets is located outside the Content directory. So I finally decided to use one FileSystemWatcher per directory, to handle linked assets.

raizam:
I have just commited fixes.
It now builds modified assets only, and not the entire Content project.
If assets are modified during a build, they are queued and built when the previous build is done.
I believe this is getting reliable

nkast:
mgcb will create some `*.mgcontent` files in /obj/. Those are use for content that is build multiple times with different processors and/or parameters (usually external references, like textures in models). mgcb  will assign different names to those assets , like tex.xnb, tex_0.xnb, tex_1.xnb, etc.
They also store the timestamp of the source file (last modified time?) for partial builds. 

I think the build process can use those files to sync build across multiple processes. 
Each build task should open the .mgcontent in exclusive mode and release it *after* the build is done.
If they can't get an exclusive write handle on the file they should wait or skip to the next task.
The tricky part is that it has to be able to read all files (tex.mgcontent,tex0.mgcontent,tex1.mgcontent) until it finds the right one. I am not sure about the details or if it's possible at all , it's just an idea.

raizam:
@nkast so the idea is to be able to sync these "child" assets too?

>mgcb will assign different names to those assets , like tex.xnb, tex_0.xnb, tex_1.xnb, etc.

I've tried to build an .obj with textures, in this case the original names of the textures are kept.
If there are other processors that assigns random names it is not a desirable behavior IMO.
Looking at the .mgcontent of the model, there is a node `<BuildAsset>` containing the xnb path of the child assets, but not the path of the original textures, I think this should be added.

cra0zy:
http://teamcity.monogame.net/repository/download/MonoGame_PackagingWindows/latest.lastSuccessful/MonoGameSetup.exe?guest=1

cra0zy:
> I was already fairly worried about the scalability of the "Exists" code that was there... this seems possibly worse. Would it not be better to create one FileSystemWatcher for the entire project folder then check changes against the content items only when they change?

Actually creating one FileSystemWatcher would be a problem if the folder where .mgcb is located is too big, or in case where one of the subfolders does not give you permissions to look at it... we had those problems btw, thats why there is that crappy File.Exists implementation.

nkast:
#4198 has a detailed explanation on how all this work.

cra0zy:
On a side note if we are gonna make it so Pipeline can autobuild in the background, we should make it so it can be closed to system tray.

nkast:
It's unknown what will happen when two pipeline/mgcb try to build the same project, and that's true with the current branch. Probably that's a different discussion unrelated to this PR.

raizam:
So you guys didn't like that Exists thread? me neither, so I've removed it.
It's now using a Timer, fired every 500ms to update the filewatcher list, and trigger a build if required.
Now the FileWatchers are also responsible of checking for removed files.

@cra0zy I've looked how to use the system tray using Eto, and I realized you wrote that part =)

raizam:
I'm currently stuck with this PR, because I haven't created a specific branch for this, and I'd like to make new changes in other areas. Unless this gets accepted soon, I'm going to close this and make a new PR some another time if requested
@nkast @cra0zy @tomspilman

raizam:
PR https://github.com/MonoGame/MonoGame/pull/6041 replaces this PR

PR #6038, opened 2017-11-02
Windows 10 CoreWindow service registry

DDReaper:
Small PR to add the reference to the CoreWindow to the service library.  Replicating behaviour used in iOS/Android

Needed for MR/HL support

In Windows HoloGraphic, this is used to define the holographic space needed for the environment.  The constructor requires access to the CoreWindow to define bounds.

### DIFF
--- a/MonoGame.Framework/WindowsUniversal/UAPFrameworkView.cs
+++ b/MonoGame.Framework/WindowsUniversal/UAPFrameworkView.cs
@@ -82,6 +82,9 @@ namespace Microsoft.Xna.Framework
         {
             // Initialize the singleton window.
             UAPGameWindow.Instance.Initialize(window, null, UAPGamePlatform.TouchQueue);
+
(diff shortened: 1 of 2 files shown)


nkast:
It's possible to get the main window for everywhere from the static `CoreApplication.MainView.CoreWindow`.

DDReaper:
Really, I had tried that but it didn't work.  Will investigate further.
Although this abstraction means it is also more available using the Service Registry, like with other platforms

KonajuGames:
This looks fine to me. Thanks @DDReaper

KonajuGames:
@DDReaper Looks like an issue caused by this, detailed in #6072.

DDReaper:
Ouch, will have to investigate.  Unless on UWP, the game platform is initialised multiple times.
Will submit a PR to ensure the service isn't registered if already existing.

Odd this didn't come up in testing or affects other platforms which do the same thing.

PR #6041, opened 2017-11-04
[PipelineTool] Auto update feature

raizam:
After a lot of fighting with Git, here is the PR.
This PR replaces #6033

Added Autoupdate option and removed the Exists thread loop and replaced with the use of FileSystemWatchers.

### DIFF
--- a/Tools/Pipeline/Common/ContentItem.cs
+++ b/Tools/Pipeline/Common/ContentItem.cs
@@ -6,6 +6,10 @@ using System.ComponentModel;
 using System.Globalization;
 using Microsoft.Xna.Framework.Content.Pipeline;
 using Microsoft.Xna.Framework.Content.Pipeline.Builder.Convertors;
+using MonoGame.Framework.Content.Pipeline.Builder;
+using System.IO;
+using System.Linq;
+using System.Collections.Generic;
 
 namespace MonoGame.Tools.Pipeline
 {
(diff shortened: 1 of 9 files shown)


Ragath:
Importers and Processors can add external dependencies discovered in an asset. In order for this to work as expected, you'd have to check the dependencies from the last build and detect changes in the referenced dlls.

raizam:
Tracking subassets would be nice to have indeed.

From my understanding, I need to load .mgcontent files to retrieve the dependencies, and I need to use [`PipelineBuildEvent Load(string filePath)` ](https://github.com/MonoGame/MonoGame/blob/7c3d6870a38f8a5e479e64d935d692f2610e1cda/MonoGame.Framework.Content.Pipeline/Builder/PipelineBuildEvent.cs#L118)  in order to load them, recreate the ContentItem and be able to trigger a build directly these items.

Ragath:
No what you want to do is track the items and then trigger a build on the assets that have them as a dependency

raizam:
That's more simple actualy. sounds good to me.

raizam:
I'm not going to implement subassets handling for now, as I don't need it, this PR does however make significant changes by removing the Exist thread, replacing it by FileSystemWatchers.

Can anyone please review this? @tomspilman @KonajuGames @cra0zy @Jjagg

raizam:
External references are now supported as well

PR #6042, opened 2017-11-04
Implementing ContentManager.UnloadAsset()

raizam:
Currently it is not possible to unload specific assets from the ContentManager.
This PR adds the feature. This was discussed with @nkast


mrhelmut:
It lacks some documention on the return boolean, otherwise very welcome to me.

raizam:
As requested I've added comments on the returning boolean.
@tomspilman @KonajuGames  Is this a feature that would likely be accepted?

PR #6053, opened 2017-11-15
Fix to resolve Services registry issue for Core Application project

### LINKED ISSUE #6052: CoreApplication project cannot start as Services Registry not initialised
No idea why this didn't show up in my initial testing, but the "Application" template now fails due to PR #6038 
The XAML template is unaffected.

DDReaper:
Fixes #6052

Needed to move CoreWindow registration to later in the start process


DDReaper:
Thanks for the quick response @dellis1972

PR #6055, opened 2017-11-15
[MGCB] Parse filenames as arguments

cra0zy:
Allows parsing of filenames (and depending if its a .mgcb file it will either treat it like a response file or build it) like so:

**Build single content file:**
```
mgcb test.png
```

**Build response file:**
```
mgcb Content.mgcb
```

PS. Also fixes the error of trying to process response file with "--@:", and the fact that "=" was not a valid separator for them.

### DIFF
--- a/Installers/Linux/Main/mgcb.1
+++ b/Installers/Linux/Main/mgcb.1
@@ -15,6 +15,10 @@ is a command line tool for building XNB content on Windows, Mac, and Linux deskt
 Typically it is executed by the Pipeline GUI tool when editing content or indirectly from Visual Studio or MonoDevelop during the build process of a MonoGame project. Alternatively you can use it yourself from the command line for specialized build pipelines or for debugging content processing.
 
 .TP
+.BR \fIFILENAME\fR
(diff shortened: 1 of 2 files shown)


Installers/Linux/Main/mgcb.1
```diff
@@ -15,6 +15,10 @@ is a command line tool for building XNB content on Windows, Mac, and Linux deskt
 Typically it is executed by the Pipeline GUI tool when editing content or indirectly from Visual Studio or MonoDevelop during the build process of a MonoGame project. Alternatively you can use it yourself from the command line for specialized build pipelines or for debugging content processing.
 
 .TP
+.BR \fIFILENAME\fR
+Automatically selects either build or response file option for the selected file, see bellow for more detail on what these options are.
```
Jjagg: below*

cra0zy: No, bellow is the correct spelling.

Jjagg: https://writingexplained.org/below-or-bellow-difference

KonajuGames:
Merging

PR #6060, opened 2017-11-16
[DX] Center window on startup (fix)

### LINKED ISSUE #6051: [WinDX] Window is not centred on startup
When using Windows DirectX, the game window is no longer centred on start up again. Windows GL still works correctly.

This was previously fixed by @Jjagg in PR #5576, but something must have broken it again since then... :/

I'm super busy at the moment, but I will try to investigate this further at some point if no one else can.

Jjagg:
Fixes #6051.

@MichaelDePiazzi 
I set the ClientSize directly, but should have used the `ChangeClientSize` method that changes client size without triggering `OnResizeEnd`. So I rebroke this myself.

### DIFF
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -138,7 +138,7 @@ namespace MonoGame.Framework
             Game = platform.Game;
 
             Form = new WinFormsGameForm(this);
-            Form.ClientSize = new Size(GraphicsDeviceManager.DefaultBackBufferWidth, GraphicsDeviceManager.DefaultBackBufferHeight);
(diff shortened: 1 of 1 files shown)


MichaelDePiazzi:
Nice, thanks @Jjagg! I'll give this a test tomorrow.

MichaelDePiazzi:
Cool, I just tested this and all is working as expected now :)

Here's what I tested and confirmed to be working:

- The window is always centred on the primary monitor at startup (launched from both my primary and secondary monitors)
- If you haven't moved the window yet, it remains centred when changing resolutions
- Once you've moved the window, the position doesn't change if you change the resolution

Jjagg:
@KonajuGames This is good to go :)

Jjagg:
Bump @tomspilman @KonajuGames, can this be merged?

PR #6064, opened 2017-11-21
[ContentManager] Turned some private fields to protected for external extensibility

raizam:
Currently it is not really possible to extend ContentManager without changing the Monogame sources themselves, there are a number of private members that are needed in order to bring real customization.


raizam:
Also this PR would make PR https://github.com/MonoGame/MonoGame/pull/6042 not so important, as users would be able to implement assets unloading externaly

PR #6065, opened 2017-11-21
[Documentation] Fixup some stuff with Getting Started

cra0zy:


Documentation/getting_started/1_creating_a_new_project.md
```diff
@@ -1,3 +1,3 @@
-This tutorial assumes that you have configured MonoGame correctly on your system. If you didn't, please read the [Setting Up MonoGame](setting_up_monogame.md).
+This tutorial assumes that you have configured MonoGame correctly on your system, if you didn't, please read the [Setting Up MonoGame](setting_up_monogame.md).
```
Jjagg: if you **haven't**

Documentation/getting_started/1_creating_a_new_project_vs.md
```diff
@@ -2,7 +2,7 @@ Start up Visual Studio and select **New Project...** in the upper left corner.
 
 ![New Solution](images/getting_started/1_new_soulution_vs.png)
 
-Now you should see a "New Project" dialog pop up. From here select **Templates > Visual C# > MonoGame** category, then select **MonoGame Cross Platform Desktop Project**. Next type in the name that you wish to give your project. Do note that you should not use space character for it. For this tutorial, it will be named **ExampleGame**. After you've entered the name, click on the **Browse** button next to location text field, and select where you wish to save your project. Finally click **OK** to create a new project.
+Now you should see a "New Project" dialog pop up, from here select **Templates > Visual C# > MonoGame** category, and then select **MonoGame Cross Platform Desktop Project**. Next type in the name that you wish to give your project, for this tutorial lets just use **ExampleGame** (do note that you should not use space character for it). After you've entered the name, click on the **Browse** button next to location text field, and select where you wish to save your project. Finally click **OK** to create a new project.
```
Jjagg: lets -> let's
you should not use a space character for it -> awkward phrasing
button next to **the** location text field

Documentation/getting_started/2_understanding_the_code.md
```diff
@@ -72,19 +82,22 @@ protected override void Update(GameTime gameTime)
...
 
     //TODO: Add your drawing code here
-    
+
     base.Draw(gameTime);
 }
 ```
-Similar to Update method, it is also called multiple times per second. The reason why Update and Draw methods are split is so that even if you have to do some heavy calculations(ie. path finding), you won't slow down your game fps.
+
+Similar to Update method, it is also called multiple times per second. Unlike the update method, this method is used for drawing stuff onto the screen.
```
Jjagg: Similar to **the** Update method
Unlike the update method -> I'd drop this part

Documentation/getting_started/4_adding_basic_code.md
```diff
@@ -29,21 +32,27 @@ Now run the game.
...
-        ballTexture.Height / 2)
+    ballTexture,
+    ballPosition,
+    null,
+    Color.White,
+    0f,
+    new Vector2(ballTexture.Width / 2, ballTexture.Height / 2),
+    Vector2.One,
+    SpriteEffects.None,
+    0f
```
Jjagg: I think it's good to actually be explicit with the parameter names here. It clarifies the meaning of the call.

Documentation/getting_started/4_adding_basic_code.md
```diff
@@ -29,21 +32,27 @@ Now run the game.
...
 );
 ```
 
 With this we are setting the origin to the center of the image. Now the image will get drawn to the center of the screen.
 
 ![Draw Ball 2](images/getting_started/4_ball_center.png)
 
-Next let's setup some movement. Find the **Update** method and add:
+Next lets setup some movement. Find the **Update** method and add:
```
Jjagg: Should be let's (contraction of let us)

Documentation/getting_started/4_adding_basic_code.md
```diff
@@ -63,19 +72,22 @@ if(kstate.IsKeyDown(Keys.Right))
 base.Update(gameTime);
 ```
 
-Let's discuss the code a bit.
+Lets discuss the code a bit.
```
Jjagg: Same here

Documentation/getting_started/4_adding_basic_code.md
cra0zy: No, that method is deprecated.

Jjagg: Yeah, I don't suggest reverting the change, but I suggest adding the parameter names.

Documentation/getting_started/3_adding_content.md
```diff
@@ -15,11 +15,12 @@ You should now see a MonoGame Pipeline Tool window open up. In case it didn't ge
...
- - **Edit > Add > Existing Item...** menu button
- - **right-click > Add > Existing Item...** context menu
- 
-In our case lets user the **Add Existing Item** toolbar button.
+
+- **Add Existing Item** toolbar button
+- **Edit > Add > Existing Item...** menu button
+- **right-click > Add > Existing Item...** context menu
+
+In our case let's user the **Add Existing Item** toolbar button.
```
Jjagg: let's **use**

Documentation/getting_started/4_adding_basic_code.md
cra0zy: For now I'll just keep this.

Documentation/getting_started/3_adding_content.md
cra0zy: @Jjagg I'll just add you as admin on my fork so you can modify instead of doing this the long way around.

dellis1972:
@cra0zy this good to go?

cra0zy:
Good to go

PR #6068, opened 2017-11-22
[WindowsDX] fix NullReferenceException in Mouse.WindowHandle

nkast:
When WindowHandle/_window is not set it should return `IntPtr.Zero`.
Currently  it throws a NullReferenceException.

Related to #5816


tomspilman:
Closing this since it is already included in #6081

PR #6070, opened 2017-11-22
[MGCB] Allow build command to build directories and add exclue command

cra0zy:
Stuff done:
- build command now accepts directories
- exclude command - extra regex rules for excluding files from build command in case of a directory

PS. Have separate PRs incoming for updating man page and unit testing.

### DIFF
--- a/Tools/MGCB/BuildContent.cs
+++ b/Tools/MGCB/BuildContent.cs
@@ -9,7 +9,7 @@ using System.Linq;
 using Microsoft.Xna.Framework.Content.Pipeline;
 using Microsoft.Xna.Framework.Graphics;
 using MonoGame.Framework.Content.Pipeline.Builder;
-
+using System.Text.RegularExpressions;
 
 namespace MGCB
 {
@@ -38,6 +38,16 @@ namespace MGCB
         {
             get { throw new InvalidOperationException(); }
             set { throw new InvalidOperationException(); }
+        }
+
+        [CommandLineParameter(
+            Name = "exclude",
+            Flag = "e",
+            ValueName = "regex",
+            Description = "Extra regex rules for excluding files getting built when specifying a directory.")]
+        public void Exclude(string regex)
+        {
+            _excludeRules.Add(new Regex(regex, RegexOptions.Compiled));
         }
 
         [CommandLineParameter(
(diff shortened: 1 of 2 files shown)


dellis1972:
@cra0zy a unit test for this would be really nice. You should be able to add a new test to [1]


[1] https://github.com/MonoGame/MonoGame/blob/develop/Test/ContentPipeline/BuilderTargetsTest.cs

cra0zy:
> @cra0zy a unit test for this would be really nice. You should be able to add a new test to [1]

As mentioned above, I have a separate PR for unit tests incoming, its doing a lot of extra stuff so I decided to keep it separate.

Tools/MGCB/BuildContent.cs
```diff
@@ -40,6 +40,30 @@ public List<string> ResponseFiles
             set { throw new InvalidOperationException(); }
         }
 
+        [CommandLineParameter(
+            Name = "autoinclude",
+            Flag = "a",
+            ValueName = "directory",
+            Description = "Autoincludes all the files for the specified directory.")]
+        // This property only exists for documentation.
```
tomspilman: I don't know exactly when this started occurring... but we need to stop.

Implement the logic into the setter.

tomspilman:
What is the problem these commands are solving?

cra0zy:
> What is the problem these commands are solving?

Making .mgcb files work like the new csproj format files, that is auto pick all the content in the directory and compile it.

This also allows you to specify a directory for a compilation instead of having to go over file by file.

cra0zy:
@tomspilman The logic has been reworked, good now?

cra0zy:
Now that I thought about it I have realized that it would be much better to just add an option for build command to pick up directories instead of adding a separate flag for it.

Jjagg:
I think this is a valuable feature to add. People complain often that including content in MG takes too many steps. While the full features of the Pipeline Tool and the customizability are very useful for full games, for stuff like game jams or prototyping it makes sense to just drop stuff into a directory and be able to use it. This makes that possible.

dellis1972:
@tomspilman I'm ok with this. any objections?

cra0zy:
@dellis1972 None have complained about this, so how about a merge :) I really want to have this in 3.7

tomspilman:
I just really dislike this direction.

We're adding "do all" commands to purposefully simple procedural command file API instead of making the GUI tool more powerful.  This is going to lead to a mess of command API that will be un-editable in the GUI tool.

> People complain often that including content in MG takes too many steps. 

I don't think people are dying for a special command in the .mgcb file to do that.  Heck i can do this now in a single DOS command:
```
for %f in (*.png *.bmp *.jpg) do MGCB.exe /incremental /build:%f
```

I would rather see great "drag and drop" and "add files" support in the Pipeline tool to solve that:

 - Ask the user if they want to copy or link the files.
 - Handle multiple files and directories appropriately.
 - If the file already exists in the project then skip it or ask to replace it.
 - Fully support the operation in undo/redo both within the tool and files on disk.

tomspilman:
Another good thing we can do is provide different code recipes or snippets in the docs on how to automate your content processing.  

It could be things like my DOS example above.  A PowerShell script.  A MsBuild target.  Even some C# code to generate a command file and launch and monitor MGCB.  Everyone likes building their content in different ways as we have seen.

But i still feel that we have to provide a visual tool that works and is easy to use.

cra0zy:
> We're adding "do all" commands to purposefully simple procedural command file API instead of making the GUI tool more powerful. This is going to lead to a mess of command API that will be un-editable in the GUI tool.

The point of these commands are to make custom .mgcb files that are not editable from the Pipeline Tool but instead easily maintained by hand.

> I would rather see great "drag and drop" and "add files" support in the Pipeline tool to solve that:
> 
> Ask the user if they want to copy or link the files.
> Handle multiple files and directories appropriately.
> If the file already exists in the project then skip it or ask to replace it.
> Fully support the operation in undo/redo both within the tool and files on disk.
> Yeah the stuff above is harder to do that adding a a command, but that is my point.

That does not solve the problem this is trying to solve.

dellis1972:
I think this is a good change. It doesn't break the existing functionality
and adds more flexibility.

cra0zy:
Just a heads up, I closed this and MGCB nuget PRs since I feel like I'm the only one who cares about having those 2 things, and I don't really mind maintaining them for myself.

PR #5980, opened 2017-09-30
MGCB Improvements

cra0zy:
All the existing commands still work, these are just some qol changes for us Unix users.

Stuff done:
 - Adds flags
   - do tell me if you want some flags changed / added
 - Fixes Unix param argument to be "--" instead of "-"
 - Makes it so separators for arguments can both be ":" and "="
 - Improves help display (mostly mimicking Unix style here)
   - If the command description is too long there is extra text warping code added
 - Includes help command

Help menu on Windows:
```
MonoGame Content Builder
Builds optimized game content for MonoGame projects.

Usage: MGCB <Options>

Options:
  /@, /@:<responseFile>             Read a text response file with additional command line options and switches.
  /b, /build:<sourceFile>           Build the content source file using the previously set switches and options.
  /c, /clean                        Delete all previously built content and intermediate files.
      /compress                     Compress the XNB files for smaller file sizes.
      /config:<string>              The optional build config string from the build system.
      /copy:<sourceFile>            Copy the content source file verbatim to the output directory.
  /h, /help                         Displays this help.
  /i, /importer:<className>         Defines the class name of the content importer for reading source content.
  /I, /incremental                  Skip cleaning files not included in the current build.
  /n, /intermediateDir:<path>       The directory where all intermediate files are written.
  /d, /launchdebugger               Wait for debugger to attach before building content.
  /o, /outputDir:<path>             The directory where all content is written.
  /t, /platform:<targetPlatform>    Set the target platform for this build.  Defaults to Windows desktop DirectX.
  /p, /processor:<className>        Defines the class name of the content processor for processing imported content.
  /P, /processorParam:<name=value>  Defines a parameter name and value to set on a content processor.
  /g, /profile:<graphicsProfile>    Set the target graphics profile for this build.  Defaults to HiDef.
  /q, /quiet                        Only output content build errors.
  /r, /rebuild                      Forces a full rebuild of all content.
  /f, /reference:<assembly>         Adds an assembly reference for resolving content importers, processors, and writers.
  /w, /workingDir:<directoryPath>   The working directory where all source content is located.
```

Help menu on Unix:
```
MonoGame Content Builder
Builds optimized game content for MonoGame projects.

Usage: MGCB <Options>

Options:
  -@, --@=RESPONSEFILE              Read a text response file with additional command line options and switches.
  -b, --build=SOURCEFILE            Build the content source file using the previously set switches and options.
  -c, --clean                       Delete all previously built content and intermediate files.
[...]

PR #6071, opened 2017-11-22
[MGCB] Fix flags not beeing processed

cra0zy:


dellis1972:
@cra0zy do you have an example of what isn't working in this instance? It would help in the future if we need to look back at the history. 
My team lead always says "Write a commit comment like you are the one who is going too maintain it in the future.. but you've lost your memory" :)

cra0zy:
Flags that do not have arguments weren't being processed at all, ie:
```
mgcb -r
mgcb \r
mgcb -rc
```

etc.

tomspilman:
@cra0zy - So you are saying that `/rebuild` does not work right now?

Was it https://github.com/MonoGame/MonoGame/pull/5980 that broke this?

cra0zy:
> @cra0zy - So you are saying that /rebuild does not work right now?
> 
> Was it #5980 that broke this?

No, `/r` does not work, `/rebuild` works fine.

tomspilman:
Thanks @cra0zy .

cra0zy:
Feel free to merge this, this is a PR after all :P

PR #6056, opened 2017-11-15
[MGCB] Set working dir for all the commands under response file

cra0zy:
Better version of: #4191

Fixes: #4190

PR #6076, opened 2017-11-24
Revert "[MGCB] Set working dir for all the commands under response file"

KonajuGames:
Reverts MonoGame/MonoGame#6056


KonajuGames:
Removing until a better solution is worked out.

PR #6081, opened 2017-11-26
[DesktopDX] Handle Mouse input/state inside Mouse class

### LINKED ISSUE #6009: setting Input.Mouse.WindowHandle should make Mouse events work WITHOUT needing Game class.
Looking at old XNA-using code, under Windows, the way to make Microsoft.Xna.Framework.Input.Mouse provide mouse events relative to a given window, was to do this:

    Mouse.WindowHandle = ..the window's IntPtr...

Then calling `FrameworkDispatcher.Update()` each game loop should be sufficient to update internal information, that is retrieved via `Mouse.GetState()`.
In MonoGame, this has no effect - a default mouse state of 0's is always returned,
unless one is using a Game class (which creates a MonoGame GameWindow).

As a result, any MonoGame client that is using custom game loop that does not create a Game class, has to replace their XNA Mouse code with some non-XNA solution.

nkast:
fixes #6009 

Breaking change:
Mouse.Position will be updated on every call of GetState(). 
Previously on DesktopDX the mouse state was updated once per Update().
The new behavior is similar to XNA and some other MG platforms (WindowsUniversal & OpenGL I think).

### DIFF
--- a/MonoGame.Framework/Input/Mouse.Default.cs
+++ b/MonoGame.Framework/Input/Mouse.Default.cs
@@ -22,6 +22,14 @@ namespace Microsoft.Xna.Framework.Input
             return window.MouseState;
         }
 
+        private static MouseState PlatformGetState()
+        {
+            if (PrimaryWindow != null)
+                return PlatformGetState(PrimaryWindow);
+
+            return _defaultState;
+        }
+
         private static void PlatformSetPosition(int x, int y)
         {
             PrimaryWindow.MouseState.X = x;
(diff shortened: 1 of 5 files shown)


nkast:
We can use native methods here to avoid garbage like #6168

tomspilman:
@nkast - FYI... i'm testing this to merge.

dellis1972:
@nkast @tomspilman this ok to merge?

nkast:
@dellis1972 
From me it's ok. I am waiting @tomspilman 's feedback.

PR #6085, opened 2017-11-27
Add new Plane constructor

Paphos:
I'm quite new to GitHub, so I don't really know if I can add a constructor to the Plane struct as it will change the API. Though I think it can be useful to define a plane that way, which is more straightforward than the a,b,c,d values and doesn't require to have three points.


Jjagg:
Hey @Paphos, thanks for contributing!

I think it's alright to add this constructor. 
The implementation looks good too.

Jjagg:
Oh, actually @Paphos could you document the ctor (only the one you added)? In the same style as I did [here](https://github.com/MonoGame/MonoGame/pull/5514/files#diff-1817d7da95370542b00a2e9e7b2c17bbR64). That PR caused some issues with building documentation, so it's not merged yet, but it documents most of the public API that is left undocumented.

Jjagg:
Thanks :) This is good to go IMO!

dellis1972:
@Jjagg @Paphos just a thought, should we note in the docs (in the method) that this is not an XNA method? I know no one would read it but at least it would be documented.

cra0zy:
> @Jjagg @Paphos just a thought, should we note in the docs (in the method) that this is not an XNA method? I know no one would read it but at least it would be documented.

We have added non XNA methods and properties and even whole classes before, and have never documented that they are not part of XNA.

dellis1972:
@cra0zy I know.. perhaps we should be? Its just a though/suggestion.

cra0zy:
Wouldn't it be better to add an Attribute for that instead of adding it to the docs?

Jjagg:
IMO it's not necessary, but I don't really care either way.

Issue #6094, opened 2017-12-02
Tuple deconstruction for common types?


portal-chan:
Apologies if this was already brought up/discussed already but I couldn't find anything relevant.

Could the `Deconstruct` method be implemented for common types in the framework to allow code like this?

```csharp
var (x, y) = vec2;
var (r, g, b) = color;
// etc.
```

Juhlinus:
As of right now the common types are `structs`, which if I recall do not support deconstruction.

portal-chan:
Well I'm not home and all online compilers apparently lost tuple support overnight so I can't check for certain but isn't `System.ValueTuple` THE main tuple `struct` which supports deconstruction? It even works as an extension method so I don't see why `struct`s wouldn't support it.

Juhlinus:
Yeah, you're right, that was my bad. I've pushed an extension file on my fork of MonoGame.

cra0zy:
This has been fixed in #6096, closing...

PR #6096, opened 2017-12-03
Added type deconstruction

Juhlinus:
As per #6094.

I don't know if this is something that has been considered at all, but I figured I'd add it either way if someone is interested. 😃


KonajuGames:
I had no idea this was a new feature of C# and we had to implement support for it. The spelling error in the filename would need to be fixed first.

Juhlinus:
@KonajuGames Tuple deconstruction is not really something you need to implement support, but if you want it for custom types then you need to implement support like it.

I renamed the file.

Jjagg:
Yeah, this would be nice to add :) Thanks @Juhlinus!

Some comments:
- Can you add a License header (check any other classfile)
- This should go in the MonoGame namespace IMO.
- Could do with an r, g, b, a deconstructor for Color
- An entry should be added to the .definition file that Protobuild uses so this classfile is included in the generated projects. [here](https://github.com/MonoGame/MonoGame/blob/develop/Build/Projects/MonoGame.Framework.definition#L319)'s an example entry. The new entry would be `<Compile Include="Utilities\TupleDeconstruction.cs" />` to add it for all platforms.

Juhlinus:
@Jjagg I updated with the things you mentioned. 👍

Jjagg:
Great! 

> Maybe the file shouldn't be called "TupleDeconstruction"?

I'd go with `Deconstructions.cs` and call the class `Deconstructions` as well.

Juhlinus:
Sigh, I'm way too quick with for my own good. 😞

cra0zy:
Wouldn't it be better for deconstructors to live next to their structs instead of a being in a separate file?

Juhlinus:
It would eliminate the need for adding an extra using directive to your game in the case of wanting to use destructors. Other than that I don't think as a developer you'll notice much of a difference. I guess it boils down to how you guys want it to be implemented.

cra0zy:
> It would eliminate the need for adding an extra using directive to your game in the case of wanting to use destructors.

Actually it would have no difference since you can just add separate namespace if you wanted.

Side note, you are using the wrong namespace, MonoGame still uses `Microsoft.Xna.Framework` as the default, not `MonoGame.Framework`.

Juhlinus:
So, should I move the extensions to their respective files? (And update the namespace)

cra0zy:
> So, should I move the extensions to their respective files? (And update the namespace)

I vote for yes, then again I am the one who suggested it so... @Jjagg you agree?

Jjagg:
It does make sense to have the deconstruction methods in the Xna namespace so users don't need an extra using, so I'm fine with that. I think it's nice to have the deconstruction methods in a separate file, but I don't really mind either way. Sorry for the confusion @Juhlinus :p

cra0zy:
@Jjagg Since I think you didn't notice, deconstruct methods don't need to be extension methods, they can (and should) be just extra methods in the struct, example:
```c#
class Point
{
    public int X { get; }
    public int Y { get; }
 
    public Point(int x, int y)
    {
        X = x;
        Y = y;
    }
 
    public void Deconstruct(out int x, out int y)
    {
        x = X;
        y = Y;
    }
}
```

KonajuGames:
> deconstruct methods don't need to be extension methods

This is a very valid point.  They should only be extension methods for types that you do not own.  In this case, we own the types, so the Deconstruct methods should be in the types they are for.

Jjagg:
> Since I think you didn't notice, deconstruct methods don't need to be extension methods

I did know this. I just think it'd be nice to have this specific functionality implemented in one place. But I really don't have a strong opinion on this.

Juhlinus:
Should I add the methods to the "Public methods" region of each file, should I make a new region for just Deconstructors, or should I just put it at the bottom without any region?

cra0zy:
> Should I add the methods to the "Public methods" region of each file, should I make a new region for just Deconstructors, or should I just put it at the bottom without any region?

Add them to the public region.

Juhlinus:
There we go! Anything else while I'm at it? 😉

Juhlinus:
It looks like it's failing because the version isn't C# on the build server (?). I tried to fix it with the packages.config, as you can see. I'm not very experienced with build servers, so it was basically a shot in the dark.

Jjagg:
The build server uses C#5, so you can't use tuple syntax. You could test the Deconstruct methods by calling them directly e.g. `vec2.Deconstruct(out x, out y)`. That way people can still build MG with an older version of VS, but we ensure that tuple deconstruction will work too. You won't need the ValueTuple package either this way.

Juhlinus:
There we go, build succeed.

Jjagg:
For Ray it makes more sense to me to have position first, then direction. Not sure if that's a universal thing :p 

The tests aren't great, because they use a lot of duplicate values, so the order isn't checked. Though I don't think this really needs tests. It's easy to tell from the implementation that everything is correct :)

tomspilman:
@KonajuGames  @dellis1972 @Juhlinus - This good to merge now?

nkast:
@Jjagg 
>> For Ray it makes more sense to me to have position first, then direction. Not sure if that's a universal thing :p

I agree. It should match the order of the Constructor when possible.
`public Ray(Vector3 position, Vector3 direction)`

KonajuGames:
It may be pedantic, but I agree that `Ray` should have the same parameter order as the constructor.

tomspilman:
@nkast @KonajuGames - If i go ahead and merge this can one of you go back and fix `Ray`?

Don't want to leave this hanging longer if @Juhlinus isn't around.

KonajuGames:
I'll merge it now and fix that `Ray` type.

KonajuGames:
PR #6198 added.

PR #6097, opened 2017-12-03
Binary search for curve evaluation

yousernaym:
In `Curve.Evaluate`, instead of looping through the keys to find those surrounding the evaluation point, I tried a binary search. It reduced the vertex buffer creation time for a large scene from a few seconds to around one second. For debug builds it went from minutes to a few seconds.

Since the underlying List container `Curve._keys._keys` (ie. `CurveKeyCollection._keys`) is private I had to make a public method `CurveKeyCollection.IndexAtPosition` in which I do the search.

### DIFF
--- a/MonoGame.Framework/CurveKeyCollection.cs
+++ b/MonoGame.Framework/CurveKeyCollection.cs
@@ -182,15 +182,14 @@ namespace Microsoft.Xna.Framework
         /// Searches for the key with the lowest position greater than or equal to the specified position.
         /// </summary>
         /// <param name="position">Position to search for.</param>
-        /// <returns>The lowest index of the matching keys or, if the last element's position is less than the specified position, <c>Count</c>. </returns>
+        /// <returns>The zero-based index of the first matching position if there is a match; otherwise, a negative number that is the bitwise complement of the index of the next element that is larger than <c>position</c> or, if there is no larger element, the bitwise complement of <c>Count</c>.</returns>
         public int IndexAtPosition(float position)
         {
             int index = _keys.BinarySearch(new CurveKey(position, 0));
-            
-            //If position doesn't exist in list, return the key with the closest larger position
+
             if (index < 0)
-                return ~index;
-            
+                return index;
+
             //If several matching keys exist, return the first one
             while (index - 1 >= 0 && _keys[index - 1].Position == position)
                 index--;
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/CurveKeyCollection.cs
```diff
@@ -20,6 +20,7 @@ public class CurveKeyCollection : ICollection<CurveKey>
         #region Private Fields
 
         private readonly List<CurveKey> _keys;
+        private readonly IComparer<CurveKey> _keyPositionComparer = Comparer<CurveKey>.Create((x, y) => x.Position.CompareTo(y.Position));
```
nkast: CurveKey already implements IComparable<CurveKey> which is faster. There's no need to use a IComparer delegate.
https://github.com/MonoGame/MonoGame/pull/4172#issuecomment-147288961

MonoGame.Framework/CurveKeyCollection.cs
```diff
@@ -178,6 +179,25 @@ public int IndexOf(CurveKey item)
             return _keys.IndexOf(item);
         }
 
+        /// <summary>
+        /// Searches for the key with the lowest position greater than or equal to the specified position.
+        /// </summary>
+        /// <param name="position">Position to search for.</param>
+        /// <returns>The lowest index of the matching keys or, if the last element's position is less than the specified position, <c>Count</c>. </returns>
+        public int IndexAtPosition(float position)
```
nkast: This looks like a useful extension.

MonoGame.Framework/CurveKeyCollection.cs
```diff
@@ -178,6 +179,25 @@ public int IndexOf(CurveKey item)
...
+        {
+            int index = _keys.BinarySearch(new CurveKey(position, 0), _keyPositionComparer);
+            
+            //If position doesn't exist in list, return the key with the closest larger position
+            if (index < 0)
+                return ~index;
```
nkast: Why not preserve the info for the developer? (result >=0 for an exact match or the negative of the closest match).
This check can be moved in `GetCurvePosition(float position)`

MonoGame.Framework/CurveKeyCollection.cs
Kloda: Yeah I see now that the position comparison is already implemented. :)

SimonDarksideJ:
@yousernaym can you find the time to address @nkast suggestions / corrections so we can get this merged?

PR #6102, opened 2017-12-11
c# Replacement for Nvidia Texture Tool (fixes #6093)

### LINKED ISSUE #6093: MGCB - NVidia.TextureTools Exception when processing a texture
Monogame `3.7.0.966`, then tried with `3.7.0.1126` and still the same problem.

I am randomly getting this exception when processing assets with MGCB:

```
System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   in Nvidia.TextureTools.Compressor.nvttCompress(IntPtr compressor, IntPtr inputOptions, IntPtr compressionOptions, IntPtr outputOptions)
   in Microsoft.Xna.Framework.Content.Pipeline.Graphics.DxtBitmapContent.TryCopyFrom(BitmapContent sourceBitmap, Rectangle sourceRegion, Rectangle destinationRegion)
   in Microsoft.Xna.Framework.Content.Pipeline.Graphics.BitmapContent.Copy(BitmapContent sourceBitmap, Rectangle sourceRegion, BitmapContent destinationBitmap, Rectangle destinationRegion)
   in Microsoft.Xna.Framework.Content.Pipeline.Graphics.BitmapContent.Copy(BitmapContent sourceBitmap, BitmapContent destinationBitmap)
   in Microsoft.Xna.Framework.Content.Pipeline.Graphics.TextureContent.ConvertBitmapType(Type newBitmapType)
   in Microsoft.Xna.Framework.Content.Pipeline.Graphics.TextureProfile.ConvertTexture(ContentProcessorContext context, TextureContent content, TextureProcessorOutputFormat format, Boolean isSpriteFont)
```

It happens when processing textures of 3D models. what's worse is that it doesn't happens always.... for now the only solution is to try to recompile until it works. But it's a nightmare because we have our project in TeamCity, so we have to manually queue new rebuilds until it compiles.

vpenades:
This is a fix for https://github.com/MonoGame/MonoGame/issues/6093 , since the NVidia Texture Tool is already 10 years old, I was told that it was due to be replaced by a c# implementation.

So I've just rolled in a c# implementation of [LibSquish](https://github.com/svn2github/libsquish) the original C++ library also used by NVidia Texture Tool.

My C# library can also be found at [TextureSquish](https://github.com/vpenades/TextureSquish) , where you can also find unit tests.

The implementation code used by MonoGame is located in Graphics\DtxTexture , and it has been modified to use Microsoft.XNA.Framework.VectorX  instead of System.Numeric.Vectors, which are faster since they're SIMD accelerated, but I didn't wanted (nor I know how to) add a nuget package reference to the project.

Compression quality is set to maximum by default, this is the same behaviour as NVTT

After trying this version with our project, all the content assets compile again.

I've left the NVidia Texture Tool dependency, since I don't know if it's being used by other projects.


SeriousMartin:
I had a look at this the other day. 

I'd suggest to use your C# library at [TextureSquish ](https://github.com/vpenades/TextureSquish) instead of adding the complete source code.

It can simply be added to the dependencies. System.Numerics.Vectors comes with a portable version for .net45 that we can use for windows systems (downloadable from https://www.nuget.org/packages/System.Numerics.Vectors/4.4.0). Mono should also support it (http://www.mono-project.com/news/2016/12/20/system-numeric-vectors/).

The nvidia texture tool dependency can be removed.

@Jjagg @vpenades 
What do you think?

vpenades:
@SeriousMartin 

Hi, I guess you mean copying TextureSquish and System.Numerics.Vectors DLLs to the Third Party dependencies directory?

TextureSquish is a NetStandard project, as such, it will never pull the Portable binary implementation of S.N.Vectors. For example, if I use TextureSquish on a NetFX project, S.N.Vectors will be NetFX only.

In order to produce a TextureSquish.dll that depends on Portable S.N.Vectors.dll flavour, I would have to craft a project specifically for this... dunno if simply replacing one DLL by another would work.

Even that is not ideal, because S.N.Vectors uses Bait & Switch so it can, at compile time, deploy a target platform optimized assembly of S.N.Vectors.  It's possible the Portable binary of S.N.Vectors is not SIMD accelerated to ease cross platform support.

In general, cherry picking the Portable/NetStandard DLL of a nuget package that also contains platform specific implementation has the side effect of loosing platform specific optimizations.

My opinion is that we either leave things as they are, or we agree to embrace Nuget package referencing the way it's intended to be used, I think anything else is going against the grain.

SeriousMartin:
From my point of view adding the portable lib is better than copying the code and using the monogame vector classes. As far as I know nuget packages only work for windows universal and not for the "old" windows projects.

@tomspilman 
Would be good to have someone from the lead devs decide on this. I would really like to have this merged - one way or the other.

vpenades:
@SeriousMartin nuget packages can be referenced by old windows projects, but the problem is that monogame projects are generated by protobuild, using a protobuild project declaration [like this one](https://github.com/MonoGame/MonoGame/blob/develop/Build/Projects/Framework.Content.Pipeline.References.definition), which is a bit outdated and, as far as I know (I might be wrong) doesn't support referencing nuget packages, and even if it does, I believe it's against current development rules.

But I disagree on adding System.Numerics.Vectors as a plain DLL, Nuget packages are not designed to be extracted to take individual DLLs, because you loose the original intent of the author; it might happen that the DLL that looks like the most "cross platform" is actually a Bait DLL that is only there to trick the IDEs to grab intellisense and definitions, but they actually do nothing if you execute them.

If anything, I could move TextureSquish code from the content pipeline project to its own project within the monogame solution, and keep using monogame vectors, to avoid referencing System.Numerics.Vectors.

I am also awaiting for this PR, since our content pipeline is currently running with this patched fork, so if moving the code to its own project (within monogame) is acceptable by everybody, I'll do it ASAP

dellis1972:
@tomspilman thoughts on this ?

dellis1972:
build

tomspilman:
@vpenades @dellis1972 

I guess this is a good step to avoid the DLL issue (#6093).  My only big concern is how fast it can compress compared to the native libraries.  Any tests of this @vpenades ?


Still this is not quite what we want in the end regardless of performance,

MonoGame's texture processing is done using a `Vector4` 32bit per component format in order to preserve the best quality regardless of input format.  With the Nvidia Texture Tool we had to reduce it to 8bit per component to pass it to the tool... which is essentially a compression step before the compression step adding artifacts to the result.

Our goal was to move to a DXT library that can process 32bit per component floating point color.

One such library is Microsoft's now open source [DirectXTex](https://github.com/Microsoft/DirectXTex) tool.  Which internally works from floating point source data preserving the best possible quality while compressing the content:

 https://github.com/Microsoft/DirectXTex/blob/master/DirectXTex/BC.h#L50

vpenades:
@tomspilman Performance tests are located [here](https://github.com/vpenades/TextureSquish/blob/master/Sources/Epsylon.TextureSquish.UnitTests/UnitTest1.cs#L44) and, in my machine, the results show that the c# code is between 2 and 3 times slower than the nvidia c++ path, which is not that bad, given that NVTT uses all sort of performance tricks, including CUDA.

Both NVTT and TextureSquish use 32bit Vector4 at the core, but they do it for performance reasons, not for quality reasons. Consider that all the DxT formats encode color in 16bit RGB 565, so most probably, increasing the quality of the input texture is not going to improve the quality of the output texture, at least noticeably. Anyway, it's something that could be implemented on TextureSquish.

I've did a fast overview of [Microsoft's DirectXTex](https://github.com/Microsoft/DirectXTex) , and as expected, it's deeply tied to DirectX SDK, so the chances to run this tool on Linux and Mac are non existant... unless you want to use this tool in windows only environment (which is fine by me since I only develop on windows, but I'm sure others don't think the same)

In the end, I don't care too much which compression library is used, as long as it doesn't crash. NVTT is being actively updated, but not the c# wrapper, in fact, I couldn't even find the source code of the NVTT wrapper being used by monogame.

So far, and given my limited knoledge, time and resources I have, this is the best solution I can provide. I would really preffer that NVTT to be fixed, since it seems a very official tool, and it keeps receiving improvements... but until somebody picks this issue, I'll keep using this fork, since it's the only thing that allows us to compile in our build server. @nkast has commented in #6093 that the issue is something that happened in the past, and somehow, it has resurfaced again. Maybe we could give a last chance for someone to look into fixing the NVTT tool, and if suceeds, I will be glad to remove my PR.

Jjagg:
The performance difference on my machine is about x8 :/

vpenades:
@Jjagg that's when doing texture compression with MGCB, or when trying the unit tests of TextureSquish?

The performance difference can be a consequence of monogame implementation of the code using Microsoft.XNA.Framework.Vector3/4 instead of System.Numerics.Vector3/4.

They're API exchangeable, but the latter uses SIMD.

I defined the vector types as `using `directives, so it's very easy to switch between them, but before using System.Numeric.Vector3/4 in MGCB we would need to reference the nuget package in the project, and I don't know how to do it. (for monogame).

Jjagg:
I ran the unit tests in TextureSquish.

vpenades:
I've just ran the PerformanceTest1 in my machine, Release Mode, AnyCPU, these are the results:


ClusterFit Iterative time: 00:00:24.2470572 (c#)
ClusterFit Alt time: 00:00:27.7154585 (c#)
NVidia time: 00:00:11.1190175 (NVTT)

That's compressing a 1600x1600 texture  100 times.

@Jjagg Could you provide the results of your test?

KonajuGames:
I had previously written a C# binding for libsquish and modified it to accept floating point texture data.  It was all ready to go except for a compression error in two 4x4 blocks in my test texture.  I much preferred the simple API of libsquish over the complexity of NVTT.

The performance of TextureSquish could likely be improved by judicious use of `unsafe` and `fixed` in the processing sections.

vpenades:
@KonajuGames could you provide that texture so I can add it to my unit tests?

Jjagg:
> @Jjagg Could you provide the results of your test?

My bad, I was testing in Debug. Results are pretty much similar to yours. I added StbSharp to test that out and it performs slightly better than nvtt. I did not check the quality however. I also don't think stb_dxt supports Dxt3. cc @rds1983 

ClusterFit Iterative time: 00:00:31.0922236
ClusterFit Alt time: 00:00:37.0291069
Stb time: 00:00:11.3990140
NVidia time: 00:00:13.4052532

tomspilman:
> the results show that the c# code is between 2 and 3 times slower than the nvidia c++ path,

Agreed... that is pretty good.  And as @KonajuGames suggested a little unsafe array use could likely improve this a lot.

I don't see that as a good reason to not use this.

> Consider that all the DxT formats encode color in 16bit RGB 565, so most probably, increasing the quality of the input texture is not going to improve the quality of the output texture, 

Actually it does with higher quality source content that is a 16bit per channel or more.  You'll notice that MS's implementation does a specialized conversion to 8bit per channel internally not just a simple scaling like we're currently doing.

A concrete example is this note about normal map compression in [this Cryteck paper](http://advances.realtimerendering.com/s2010/Kaplanyan-CryEngine3%28SIGGRAPH%202010%20Advanced%20RealTime%20Rendering%20Course%29.pdf):

image

The point is that quantization to 8bit is a form of compression...  we should leave the compression to the library and attempt to do a poor job of it ourselves.

Anyway this is a nit pick i've had for years now... it isn't the important thing here.


@KonajuGames @dellis1972 @nkast - So what are your thoughts?

I think we can merge this and move on as long as it works... 2x to 3x hit is big for only the biggest games with the largest textures, most others would never notice.  @vpenades might also get a win from some `unsafe` optimizations.  Finally we can better improve the performance by finally fixing the pipeline to thread (@nkast has a PR that we should revisit).

However... i don't want to merge this if we're just going to replace it in a few months with something else.  So now is the time to decide if this is the best path forward for us or not.

dellis1972:
Well if this works I vote we merge. I'm getting tired of `nvtt` breaking all the time (for Mac anyway). Makes us look bad.

Jjagg:
I'm interested to further check whether StbSharp.Dxt is a good solution, since it performs so well. But if we move to a managed implementation, I'm happy either way.

tomspilman:
> I'm interested to further check whether StbSharp.Dxt is a good solution

It seems to me that [stb_dxt](https://github.com/nothings/stb/blob/master/stb_dxt.h) is really just optimized for compression speed and isn't meant to be a high quality DXT compressor?  Is there any comparison of results between stb_dxt and libsquish?

tomspilman:
Also i guess none of the options we've discussed support the new DXT BC6 or BC7 formats correct?

MonoGame.Framework.Content.Pipeline/Graphics/DxtTexture/Bitmap.cs
```diff
@@ -0,0 +1,320 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace TextureSquish
+{
+    public class Bitmap
```
tomspilman: I think we should remove the `public` off here if we decide to merge this... as we don't want to make this compressor a part of our public API.

Jjagg:
> Is there any comparison of results between stb_dxt and libsquish?

Not that I know of, so i'd like to compare them.

> Also i guess none of the options we've discussed support the new DXT BC6 or BC7 formats correct?

stb_dxt does not. It does support BC4 and BC5, but not BC2 (DXT3).

MonoGame.Framework.Content.Pipeline/Graphics/DxtTexture/Bitmap.cs
```diff
@@ -0,0 +1,320 @@
+﻿using System;
```
tomspilman: @KonajuGames @dellis1972 - What should we do about a copyright header here?

I want to give credit to @vpenades work here and his library, but don't want to complicate the licensing of MG.

Jjagg: Maybe including the source files directly is not the best option here. Why not include it as a thirdparty lib. We could get started on #6144 with this. Seems like a great idea to me to use NuGet for managed dependencies. We could optionally have the installer pre-download deps to the global nuget cache so users don't have to worry about an internet connection.

tomspilman:
> Maybe including the source files directly is not the best option here

I sort of agree.

> Why not include it as a thirdparty lib. 

Mostly because updating our dependencies is a huge pain in the ass and we constantly have issues with it.  Who built it?  What version was it when they built it?  How do i find the source code to debug it? etc.

>  Seems like a great idea to me to use NuGet for managed dependencies

That i have been the grumpy old guy that has kept us from doing that.  I just like having access to the actual code/binaries used in our builds and versioned along with MG.

Lets move that discussion to that issue.

vpenades:
@Jjagg I would like to add your test code your StbSharp code to the UnitTests, from there I can add quality comparison tests.

@tomspilman Unfortunately, TextureSquish, as the original LibSquish, only supports the good old Dxt1, Dxt3 and Dxt5, the new BC6 and BC7 are not supported, for that we would have to stick with NVTT.

nkast:
@vpenades Where do you see  TextureSquish going in the future?
I see you plan to move a couple of features from nvtt, I suppose we have plans for BC6 & BC7 as well?

vpenades:
@nkast In the short/middle term, yes, I would like to see if I can port the BC6 and 7 to c# and include them in the code. Also, I will probably publish it as a nuget package.

In the very long term, I will probably request the guys of ImageSharp to pick it for integration when they begin supporting more file formats. I'm sure they'll do a better maintenance/improvement job than me. But I don't see this happening until 2019 or even later.

rds1983:
Btw, there's an alternative to both squid and stb_dxt, it is DDSHelper from SharpDX Toolkit:
https://github.com/sharpdx/Toolkit/blob/master/Source/Toolkit/SharpDX.Toolkit.Graphics/DDSHelper.cs
It is used in the Xenko Engine.

rds1983:
Nevermind. I mixed up DXT with DDS.

vpenades:
@rds1983 yes, DDS is the file format that uses DxT encoding, so the encoding bits are the right ones.

But DDSHelper, as the whole SharpDX, is Windows only, and we need an encoding solution for Linux and Apple too, that's why we discarded other solutions.

vpenades:
@tomspilman So, what's the conclusion on this?

I know it's not an ideal solution, but given that what we have is crashes and nobody was able to offer a better solution...

I'm open to do more changes if required, but I would preffer to merge and move on.

dellis1972:
@tomspilman  I just tested this on Mac and its working ok.

SeriousMartin:
I'd also like to see this merged. For our project it's better to have an increased compile time than having to do 10-100 tries unitl it's fully compiled. Especially when doing build automation...

dellis1972:
@tomspilman @KonajuGames I am going  to merge this. In my opinion having something that works is more important than having something that is fast. Without this we are blocked on Mac and probably linux. I've tried to get nvtt working on Mac without any luck. So if someone else can figure out how to fix nvtt we can revert this.

tomspilman:
So getting some numbers on performance in a real world situation.

We have some game models with around 80 textures most of them are 1024s but we have quite a bit of 2048s and 4096s.  This is very little content compared to most 3D titles which is why the numbers are alarming:

 - Native NVidia Texture Tool:  2m 37sec
 - Managed LibSquish Debug: 31m 32sec
 - Managed LibSquish Release: 25m 13sec

So this is exactly the reason i was concerned about this PR.  The performance is a big step backwards for anyone doing a moderate amounts of texture processing.  It would be impossible on most large games.

@KonajuGames mentioned adding some `unsafe` code into our TextureSquish implementation.  I'll experiment with that here for a little bit and see what i get.

If that doesn't make things at least 4x faster then we'll have to find another solution as this is not permissible as it is.

tomspilman:
> I'll experiment with that here for a little bit and see what i get.

So first pass at just converting some array indexing into unsafe pointers didn't produce much of a win.  It is still running right now, but my guess is 2 or 3 minutes shaved off of 25 minutes.

Looking at the TextureSquish implementation it really isn't written for performance.  It allocates a bunch of classes on each block compressed and it puts branches like `if (IsDxt1)` deep inside of loops.

I suspect this needs a fairly big refactor to make it perform well.

vpenades:
@tomspilman These performance numbers are odd... we had some performance tests that showed managed texturesquish is 2,4 times slower than nvidia, but your numbers show more than 10 times slower.

I'm going to add more performance tests to see if there's an unidentified problem.

Indeed, the implementation is not optimal, it's essentially an almost direct port of the C++ version and there's room for optimization. I couldn't go much deeper because we were in a hurry to fix our build pipeline, since nVidia tool kept crashing on us.

Maybe the nvidia texture tool can be reintroduced and, in MGCB, to have a switch to choose between both compression mechanisms:

|Tool |Reliability |Performance |Cross Platform|
|-|:-:|:-:|:-:|
|nVidia Texture Tool|No|Fast|With Issues|
|TextureSquish|Yes|Slow|Yes|

So people that has issues with the nVidia texture tool could have an alternative method to compress textures.

Regarding improving performance of TextureSquish, I was planning to use the newly introduced Span<T>, but then I remembered monogame is still limited to c# 5 features, so yes, removing branching and adding unsafes is the way to go.

vpenades:
@tomspilman I've strenghted the performance benchmark tests and I've identified an image in my test collection that takes unusually longer to compress than the other images; I suspect the compressors are very sensitive to the texture content.

I plan to do some profiling to see if I can identify a bottleneck that might be affecting the performance compression.

One potential issue is that over set quality to absolute maximum by default. When Iterative cluster fit is selected, the compressors can retry several times to try to find a better block, LibSquish author stated that it took 1.6 times on average to find the best fit. But it seems it can be much worse it seems.

So a fast solution for now would be to use RangeFit for debug mode, and IterativeClusterFit on release

tomspilman:
> I suspect the compressors are very sensitive to the texture content.

Yes they are.  My guess is most of your test textures were simple with lots of solid color.  You should include more complex textures like this at large sizes like 2048x2048:

image

Which is the typical sort of thing you'll find in 3D games with a realistic visual style.  Also textures with lots of transparency like tree fronds:

image

Also consider normal maps like this:

image

Which are also very common in 3D games.

> I plan to do some profiling 

That would be great.

For now we're not merging the latest MG in our project until this is resolved.


> use RangeFit for debug mode, and IterativeClusterFit on release

I think that is a bad idea.  It means you get worse quality in debug builds which i would think would be unexpected and will directly lead to questions here.  We should stick to the same quality in both modes... we just need big improvements in the compressor.

MonoGame.Framework.Content.Pipeline/Graphics/DxtTexture/Bitmap.cs
```diff
@@ -0,0 +1,320 @@
...
+
+            CompressImage(this, blocks, flags);
+
+            return blocks;
+        }        
+        
+        private static void CompressImage(Bitmap srcImage, Byte[] blocks, CompressionMode flags)
+        {
+            // fix any bad flags
+            flags = flags.FixFlags();
```
tomspilman: This is really poor...  for every block in a 2048x2048 (that would be 262,144 blocks) it fixes the flags?  This should be done outside the image loop.

MonoGame.Framework.Content.Pipeline/Graphics/DxtTexture/BlockWindow.cs
```diff
@@ -0,0 +1,612 @@
...
+        /// </remarks>
+        /// <param name="rgba">The rgba values of the 16 source pixels.</param>
+        /// <param name="mask">The valid pixel mask.</param>
+        /// <param name="flags">Compression flags.</param>
+        public void CompressMasked(Byte[] rgba, int mask, CompressionMode flags)
+        {
+            System.Diagnostics.Debug.Assert(rgba != null && rgba.Length == 64, "rgba");
+
+            // fix any bad flags
+            flags = flags.FixFlags();
```
tomspilman: This is really poor... for every block in a 2048x2048 (that would be 262,144 blocks) it fixes the flags? 

This should be done once per image compressed... not for each block.

MonoGame.Framework.Content.Pipeline/Graphics/DxtTexture/BlockWindow.cs
```diff
@@ -0,0 +1,612 @@
...
+
+namespace TextureSquish
+{
+    sealed class BlockWindow
+    {
+        public BlockWindow(Byte[] array, CompressionMode flags)
+        {
+            _Array = array;
+
+            flags = flags.FixFlags();
```
tomspilman: You should store the flags here on the BlockWindow so a) you never have to pass it again in the Compress calls, b) you don't have to fix them on every block.

MonoGame.Framework.Content.Pipeline/Graphics/DxtTexture/BlockWindow.cs
```diff
@@ -0,0 +1,612 @@
...
+
+            _BytesPerBlock = ((flags & CompressionMode.Dxt1) != 0) ? 8 : 16;
+            _ColourOffset = (flags & (CompressionMode.Dxt3 | CompressionMode.Dxt5)) != 0 ? 8 : 0;
+        }
+
+        private readonly Byte[] _Array;
```
tomspilman: This should be stored as an unsafe pointer...  you never use the length anyway,

MonoGame.Framework.Content.Pipeline/Graphics/DxtTexture/Fits/SingleColourFit.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            // initialise the best error
+            m_besterror = int.MaxValue;
+        }
+
+
+        protected override void Compress3(BlockWindow block)
+        {
+            // build the table of lookups
+            var lookups = new SingleColourLookup[][] { SingleColourLookupTables.lookup_5_3, SingleColourLookupTables.lookup_6_3, SingleColourLookupTables.lookup_5_3 };
```
tomspilman: Things like this which are constants should be allocated one time.

tomspilman:
Not going to comment on every problem in the code here... don't have the time and i would have tons of comments.

I think the main problem is this version is written like this:

  Foreach Block -> Decide on compression to do -> Do compression

And it should be written like this:

 Decide on compression to do -> Foreach Block -> Do compression

I am aware of how DXT works and that some decisions have to occur per-block...  but how this is written now all decisions, allocations, and calculations are done per block which is very bad for performance.

vpenades:
@tomspilman I agree with all your comments. Keep in mind that this is an almost direct port of the original c++ code,  which was done in a hurry to have a solution for the nvidia texture tool crash.

Part of the problem is that I was naive with the performance tests, and I thought the performance loss was acceptable. 

Since then I've been expending some time trying to figure out how the whole compression works and how to improve the compression performance. The conclusion is that a complete refactor would be required, but that would take time quite some time.

In the mean time I've been testing different optimizations with the current code base, I'll definitely add your examples to the test cases.

In the long run... Well, I don't mind if nVidia texture tool comes back, because let's face it, there's no way in hell C# is going to beat a CUDA based compressor... As long as it does it with reliability and no crashes!

I've been trying to use STBsharp to compress, it's more limited in features and I still have to check how it compares with quality with the other compressors... 

There's definitely a lot of work still to be done, I'm sorry it's affecting your work, but as others said, slow is still better than random crashes.

dellis1972:
if nvtt comes back then I guess we fork it and setup a travis-ci job to
make sure we can build the linux and mac (and windows) native libs and put
some unit tests in for the C# wrapper.
One of the problems is we need 64bit native libs.. we don't get those on
Mac by default and we have no idea exactly which version of the libs we
should be using.
This is what was causing all the problems. the C# wrapper was out of date
and nvtt seems to have moved on.

tomspilman:
> if nvtt comes back then I guess we fork it 

Doing some research I think this is the only reasonable solution for now.  No other library has fewer dependencies and officially supports Mac and Linux.

https://github.com/castano/nvidia-texture-tools

I think we should fork and make changes to ensure we static link runtime libraries to remove that extra dependency.

Also we should consider if we actually need a C# binding.  We could make a new minimal C# binding exposing just the API we need for the pipeline.  The risk here is bugs in the bindings or memory overruns within our process.

However we could also call the command line tool to do the work for us.  This means no C# binding bugs or memory issues.  We then just have to handle error cases and read back of the compressed content back from a temporary DDS file which isn't too hard.

@KonajuGames @dellis1972 ?

tomspilman:
Actually they have an official wrapper:

https://github.com/MonoGame/nvidia-texture-tools/blob/master/project/vc12/Nvidia.TextureTools/TextureTools.cs

So just need to use it.

KonajuGames:
I believe one of the issues with the supplied C# bindings is that they were
not always kept up to date with changes in the C API.

It is certainly possible to write out the TextureContent to a temporary PNG
file and use the command line tool to convert it to a DDS file. Then use
the DDS loader I wrote a while back to read it in. It would also avoid any
marshalling issues in the bindings.

tomspilman:
> I believe one of the issues with the supplied C# bindings is that they were
not always kept up to date with changes in the C API.

I have a fork now and i'm testing the current bindings. 

> It is certainly possible to write out the TextureContent to a temporary PNG
file and use the command line tool to convert it to a DDS file. 

The only thing that sort of sucks in this case is your previous refactor of the texture pipeline to compress one mip level at a time.  This means in a big texture 12 different PNGs written out, 12 executions of the nvcompressor.exe, and 12 different DDS files loaded back in.  

While i like the idea of using the command line calling it 12 times for one texture seems sort of messy.

KonajuGames:
True that would be messy. The other side would be making a special case for
mipmap generation for DXT, which could be its own issue.

tomspilman:
> The other side would be making a special case for mipmap generation for DXT, 

Well if we did anything we would go back to letting the TextureProfile decide compression for the whole TextureContent instead of a per-mip thing.  I really don't remember why we had to change that in the first place.

KonajuGames:
If you copy any pixel BitmapContent to any compressed BitmapContent type,
it compresses that single bitmap to the target format. This is currently
consistently implemented across all supported compression types. We could
special case DXT compression in TextureContent.Convert, or perhaps it would
have to be higher up than that to skip the regular mipmap generation and do
it in one call to NVTT and create the target BitmapContent instances
separately. If anyone imports an uncompressed DDS with mipmaps or has their
own texture processor that produces special mipmaps, this special case for
DXT compression would need to be bypassed and the current processing still
take place.

vpenades:
@tomspilman make sure to test NVTT under a multithreaded scenario, because it works fine when you call it in a single thread.

I've opened an issue a while ago about the NVTT wrapper here: https://github.com/castano/nvidia-texture-tools/issues/272

@nkast identified that the problem with NVTT was related to processing multiple textures simultaneously under MGCB.

Also, I would consider keeping a c# implementation around, and have the possibility to configure MGCB to allow selecting the compressor, just as a safe fallback mechanism.

📱

nkast:
@vpenades 
>> @nkast identified that the problem with NVTT was related to processing multiple textures simultaneously under MGCB.

Not true. 
NVTT compression worked fine with my Multi-threading branch.  There was an issue with NVTT in the past which was fixed. 

Anyways, the current develop is single threaded so this shouldn't be a concern.

vpenades:
@nkast then I'm lost about which was the issue that caused NVTT to crash 😦

nkast:
The last fix had to do with an unpinned delegate. Multi-threading cause it to crash once every 5-6 assets in a single build, while normally the bug rarely appear, less that once per day. I never show a crash since the fix. I run about 300 images after your report but I got nothing.

Here is the last build I have in case you want to test it yourself (3.7.0.0733). 
https://1drv.ms/u/s!AgIeh29TX5UhkPUgz7OUTGM3dm8d0w
You have to run MGCB with /multithreadMode or enable it from the options in the GUI.

SeriousMartin:
@tomspilman @vpenades 
I'm a little concerned about the highly increased build duration. Will the new manged compression be improved or replaced again? I couldn't find an open issue for this one and would like to know where this is going.

tomspilman:
@SeriousMartin - I am very concerned and i'm working on the issue.

I've [forked NVTT into the MG org](https://github.com/MonoGame/nvidia-texture-tools).  This fork is modified to build against the static VC++ runtime on Windows so there will be no issues with missing DLLs.

I also [setup an automated build](http://teamcity.monogame.net/project.html?projectId=Dependencies&guest=1) for it.  This ensures we don't have to remember how to build the thing or mess things up.  It should be as easy as merging fixes from upstream or our own bug fixes and the automated build spits out the correct files.

My current task is to setup the Mac and Linux versions building the same way.

After that i can restore NVTT into the pipeline for good.

I still believe a pure C# approach can be used for texture compression and be nearly as fast as the C+ code.  It just has to be designed to be fast and not like an academic project on object oriented code.  I just don't have the time to go prove that when existing libs that handle the formats we need exist.

It is my mistake this got merged and not just fixing the issue... static linking against the VC++ runtime.

vpenades:
Any improvement I do on TextureSquish is being done at https://github.com/vpenades/TextureSquish , so far, I replaced my naive benchmark by a proper benchmark framework using BenchmarkDotNet , and indeed the compilation times are much higher than I initially expected which renders the library pretty much unusable for large projects.

Off the grid, I've been doing some basic performance improvement tests, without success; Indeed the main performance issue is how the whole compression algorythm is architectured (because it's a 1-to-1 copy of the c++ impl), and a complete refurbish of the code is required to boost performance.

Another thing I want to try before considering a complete rewrite is to check the quality of STBSharp is comparable with TextureSquish and nVidia, because it's a very fast compressor, but I fear it's at the cost of quality.

Problem is that I'm currently in crush time at my day job, so I can't give any timeline.

@tomspilman Everybody is trying to help in any way they can... at the time, using the c# implementation seemed the best solution given the available resources, and the number of developers that were complaining about the NVTT crashes. I'm sorry if this caused you any headache, but we really needed a solution (even a slow solution) at the time.

If you're going to fork NVTT and put it back into MonoGame, something I can do is to incorporate it to the unit tests of my TextureSquish.

tomspilman:
> the main performance issue is how the whole compression algorythm is architectured

Right... that is the core issue as i see it.  Too many branches, too much temp arrays being allocated per block, too many functions calling other functions in tight inter loops.

I recommend [examining Microsoft's implementation of BC compression](https://github.com/Microsoft/DirectXTex/blob/master/DirectXTex/DirectXTexCompress.cpp#L70) for ideas on a better structure.


> check the quality of STBSharp

That would be worth it i think.

> I'm currently in crush time at my day job, so I can't give any timeline.

No worries.  I'll be glad to switch to a pure C# solution in the future.


> I'm sorry if this caused you any headache,

For sure i am not complaining about anything you did @vpenades .  You were just trying to help.

I just should have known better and detected performance problem way before we merged this.

> something I can do is to incorporate it to the unit tests of my TextureSquish.

I was thinking about that too.

Even if we had a very simple performance test of 3 or 4 different real world textures across the different formats.  We could assert on their compression times being within an expected max tolerance and detect when we've broken that.

We just should avoid making these tests take too much time.

SeriousMartin:
@tomspilman @vpenades I am very happy to hear that. Thanks for the quick update!
As soon as the new (old) implementation is merged I will have a try with Imagine Earth (our game).

PR #6108, opened 2017-12-13
[2MGFX] Split ShaderInfo

nkast:
ShaderInfo keeps only the data that are related to the parsed node.
Everything else -including ShaderInfo- are moved to ShaderResult.

The idea is to seperate the parser from everything else in order to swap
it with the HLSL parser by @TGJones.
As a next step we can move everything related to the parser and tree
nodes to another folder and namespace (TwoMGFX.TPGParser).


tomspilman:
Not sure why this failed... looks like it couldn't find any graphics adapters.

Going to retry the build.

tomspilman:
Ok... the build box was all working weird.  Updated and rebooted everything and now it works.

tomspilman:
Thanks @nkast !

PR #6118, opened 2017-12-24
Delete wrong use RuntimeHelpers.OffsetToStringData

jkotas:


MonoGame.Framework/Graphics/OpenGL.cs
```diff
@@ -1522,7 +1522,7 @@ protected unsafe static IntPtr MarshalStringToPtr (string str)
             if (intPtr == IntPtr.Zero) {
                 throw new OutOfMemoryException ();
             }
-            fixed (char* chars = str + RuntimeHelpers.OffsetToStringData / 2) {
+            fixed (char* chars = str) {
```
jkotas: The original code was actually doing something like this:

`fixed (char* chars = String.Concat(str, (RuntimeHelpers.OffsetToStringData / 2).ToString()))`

tomspilman:
NOTE: This is related to the discussion in https://github.com/MonoGame/MonoGame/pull/6132.

tomspilman:
@dellis1972 - So this change builds correctly and seems like the correct code to me.  Why should we not merge it?

tomspilman:
This same fix has already been applied:

 https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Graphics/OpenGL.cs#L1574

Closing this.

jkotas:
> This same fix has already been applied:

@tomspilman I do not see the fix in the current tip. The bug is still there.

PR #6125, opened 2017-12-29
Check for  Invalid SpriteFonts

nkast:


MonoGame.Framework/Graphics/SpriteFont.cs
```diff
@@ -73,17 +73,21 @@ public int GetHashCode(char b)
...
+                else if(characters[i] == (regions.Peek().End+1))
                 {
                     var currentRegion = regions.Pop();
                     // include character in currentRegion
                     currentRegion.End++;
                     regions.Push(currentRegion);
+                }
+                else // characters[i] < (regions.Peek().End+1)
+                {
+                    throw new InvalidOperationException("Invalid SpriteFont. Character map must be in ascending order.");
```
tomspilman: I assume this is not something XNA does correct?

nkast: I don't know what's the behavior of XNA in this case. It says that the pipeline is "guaranteed" to generate sorted spritefonts but it's unknown if the runtime fails one way or another otherwise.

@SeriousMartin, Can you share the .XNB that failed?

SeriousMartin: https://github.com/MonoGame/MonoGame/pull/5874#issuecomment-354729476 I have a custom implementation of FontTextureProcessor so this is not a problem of the core MonoGame/XNA classes.

The check you added was very helpful to find and understand the problem in my FontTextureProcessor implementation - so I would suggest to merge it into the system, although XNA likely doesn't have it.

tomspilman:
This pairs well with https://github.com/MonoGame/MonoGame/pull/6126 .

Merging

PR #6126, opened 2017-12-30
Made SpriteFont constructor public

rds1983:
Please, consider making the SpriteFont constructor public. So it could be created in the run-time, ignoring Content Pipeline.


Jjagg:
If it's gonna be made public it should be documented.

rds1983:
I've documented the constructor and also have opened SpriteFont.Glyphs(why not?).

Jjagg:
@tomspilman Any idea what's going wrong with the build bots?

rds1983:
build succeed

tomspilman:
Thanks @rds1983 !

PR #6131, opened 2018-01-05
Updated to the latest version of StbSharp

### LINKED ISSUE #6114: MonoGame.Utilities's Operations.Malloc is not threadsafe, and so Texture2D.FromStream isn't
Been suffering from random failures to load pictures into textures, which ultimately were tracked down to exceptions while trying to index the dictionary used by Operations.Malloc.

On inspecting https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Utilities/Imaging/Operations.cs it became clear that there isn't any protection against concurrent writes to the dictionary tracking allocations, and so this method would be unsafe to call from multiple threads at once, but Texture2D.FromStream assumes that the image loading can be done from any thread and doesn't dispatch to the UI thread until actually creating the texture, leading to multiple calls into malloc at once during texture loading. As a temporary fix I've modified my fork to do the image loading on the UI thread too- this is not ideal but prevents crashes.

Replication is somewhat difficult- you have to spawn multiple threads and try to load textures on them, and sometimes it will crash. I could put together a sample project if needed, although I think the problem is fairly visible to inspection- there's an object called "_lock" in Operations.cs, but nothing ever uses it, and in the absence of any locking concurrent writes to a dictionary are incorrect.

rds1983:
This PR updates MG's version of StbSharp to the latest one. 
Hopefully it would fix #6114 with loading textures from multiple threads. I've replaced Dictionary with ConcurrentDictionary and updated my testing framework, so about 800 textures were simultaneously processed from 10 different threads. There were no problems.
Also latest Stb can write JPGs. So TinyJpegSharp had been removed.


dellis1972:
build

vpenades:
@dellis1972 Could this PR indirectly fix #6093 ?

dellis1972:
I don't think so.

rds1983:
As far as I know, StbSharp is only used in the Texture2D.FromStream. And only in the DesktopGL version of MG, as #6008 hadn't been merged yet.

PR #6132, opened 2018-01-06
Android crash fix: unsafe fixed statement string->char* causes crash on Android

roponator:
**DO NOT MERGE IN YET, NEEDS MORE TESTS!!!**
**DO NOT MERGE IN YET, NEEDS MORE TESTS!!!**

Issue:  the app crashes if a function is called which contains a statement like this:
string str = ...
fixed(char* c = str) 

If we convert the string to unicode bytes and use fixed(byte* b=strBytes) it works.

EDIT: **this issue appeared when OpenTK was removed,** an older version of monogame which still has OpenTK (3.7.0.512) works fine.
OpenTK code seemed to have used some native OpenGL calls to do this.

EDIT 2: I get a crash also if I executed fixed(char* c = str) in our game code (so outside of the MonoGame lib).

**Potential issue**: this RuntimeHelpers.OffsetToStringData is not applied now: 
char* c = str + RuntimeHelpers.OffsetToStringData / 2
Any idea about this? It works for a quick test game (stuff renders, will need to test more).

@dellis1972 If I remember correctly you were the one who did the OpenTK removal and rewrite? Do you have any idea about this?

I'm using the latest development branch.

The issue is very strange because the app crashes not when this 'fixed' statement is encountered, but when the function which contains this 'fixed' statement is called:

```
unsafe void DoUnsafe()
{
  // neither code nor debugger gets to here! 
   System.Diagnostics.Debug.WriteLine("testy"); // this does not print!
    string str="bla";
   fixed(char* c = str)
   {
   }
}

void Issue()
{
  DoUnsafe(); // crash happens here!
}
```

Some observations: 
- sometimes I get a "index out of range" exception, but does not happen always. I assume its a side effect of the strange crash, because most of the time the app shuts down hard, no c# exception is thrown most of the time.

I have confirmed this crash on 2 phones: Huawie G620S-L01 and Nexus 5X.

The fix works on both phones.

Can anyone else reproduce this?

Top of stack trace:

```
01-06 22:41:26.830 9578-9578/com.game E/mono-rt: Stacktrace:
01-06 22:41:26.830 9578-9578/com.game E/mono-rt:   at <unknown> <0xffffffff>
01-06 22:41:26.830 9578-9578/com.game E/mono-rt:   at MonoGame.OpenGL.GL.MarshalStringArrayToPtr (string[]) [0x00052] in <aac7c7476e4b413091b00e98f9c80047>:0
01-06 22:41:26.830 9578-9578/com.game E/mono-rt:   at MonoGame.OpenGL.GL.ShaderSource (int,string) [0x00008] in <aac7c7476e4b413091b00e98f9c80047>:0
01-06 22:41:26.830 9578-9578/com.game E/mono-rt:   at Microsoft.Xna.Framework.Graphics.Shader.GetShaderHandle () [0x00047] in <aac7c7476e4b413091b00e98f9c80047>:0
```

Android project compilation:
Compiling with Android 7.1
Min Android version: 4.4
Target Android version: 4.4

Phone 1: Huawei G620S-L01
OS: Custom rom LineageOS 14.1
Android OS version: 7.1.2

Phone 2: Nexus 5X
Stock rom.
Android 8.0.0

### DIFF
--- a/MonoGame.Framework/Graphics/OpenGL.cs
+++ b/MonoGame.Framework/Graphics/OpenGL.cs
@@ -1522,9 +1522,13 @@ namespace MonoGame.OpenGL
             if (intPtr == IntPtr.Zero) {
                 throw new OutOfMemoryException ();
             }
-            fixed (char* chars = str + RuntimeHelpers.OffsetToStringData / 2) {
-                int bytes = Encoding.ASCII.GetBytes (chars, str.Length, (byte*)((void*)intPtr), num);
-                Marshal.WriteByte (intPtr, bytes, 0);
+
+            // Must NOT use fixed(char* c = str) because it crashes on some device for some unknown reason!
+            byte[] strSafeBytes = Encoding.Unicode.GetBytes(str);
+            fixed (byte* strRawBytes = strSafeBytes)
+            {
+                int bytes = Encoding.ASCII.GetBytes((char*)strRawBytes, str.Length, (byte*)((void*)intPtr), num);
+                Marshal.WriteByte(intPtr, bytes, 0);
                 return intPtr;
             }
         }


dellis1972:
This feels like a Xamarin.Android issue rather than ours.I've not come across this before 

Sent from my Windows 10 phone

From: roponator
Sent: 06 January 2018 22:54
To: MonoGame/MonoGame
Cc: Dean Ellis; Mention
Subject: [MonoGame/MonoGame] Android crash fix: unsafe fixed statementstring->char* causes crash on Android (#6132)

DO NOT MERGE IN YET, NEEDS MORE TESTS!!!
DO NOT MERGE IN YET, NEEDS MORE TESTS!!!
Issue: the app crashes if a function is called which contains a statement like this:
string str = ...
fixed(char* c = str)
If we convert the string to unicode bytes and use fixed(byte* b=strBytes) it works.
EDIT: this issue appeared when OpenTK was removed, an older version of monogame which still has OpenTK (3.7.0.512) works fine.
OpenTK code seemed to have used some native OpenGL calls to do this.
Potential issue: this RuntimeHelpers.OffsetToStringData is not applied now:
char* c = str + RuntimeHelpers.OffsetToStringData / 2
Any idea about this? It works for a quick test game (stuff renders, will need to test more).
@dellis1972 If I remember correctly you were the one who did the OpenTK removal and rewrite? Do you have any idea about this?
I'm using the latest development branch.
The issue is very strange because the app crashes not when this 'fixed' statement is encountered, but when the function which contains this 'fixed' statement is called:
unsafe void DoUnsafe()
{
  // neither code nor debugger gets to here! 
   System.Diagnostics.Debug.WriteLine("testy"); // this does not print!
    string str="bla";
   fixed(char* c = str)
   {
   }
}

void Issue()
{
  DoUnsafe(); // crash happens here!
}
Some observations:
• sometimes I get a "index out of range" exception, but does not happen always. I assume its a side effect of the strange crash, because most of the time the app shuts down hard, no c# exception is thrown most of the time.
I have confirmed this crash on 2 phones: Huawie G620S-L01 and Nexus 5X.
The fix works on both phones.
Can anyone else reproduce this?
Top of stack trace:
01-06 22:41:26.830 9578-9578/com.game E/mono-rt: Stacktrace:
01-06 22:41:26.830 9578-9578/com.game E/mono-rt:   at <unknown> <0xffffffff>
01-06 22:41:26.830 9578-9578/com.game E/mono-rt:   at MonoGame.OpenGL.GL.MarshalStringArrayToPtr (string[]) [0x00052] in <aac7c7476e4b413091b00e98f9c80047>:0
01-06 22:41:26.830 9578-9578/com.game E/mono-rt:   at MonoGame.OpenGL.GL.ShaderSource (int,string) [0x00008] in <aac7c7476e4b413091b00e98f9c80047>:0
01-06 22:41:26.830 9578-9578/com.game E/mono-rt:   at Microsoft.Xna.Framework.Graphics.Shader.GetShaderHandle () [0x00047] in <aac7c7476e4b413091b00e98f9c80047>:0
Android project compilation:
Compiling with Android 7.1
Min Android version: 4.4
Target Android version: 4.4
Phone 1: Huawei G620S-L01
OS: Custom rom LineageOS 14.1
Android OS version: 7.1.2
Phone 2: Nexus 5X
Stock rom.
Android 8.0.0

You can view, comment on, or merge this pull request online at:
  https://github.com/MonoGame/MonoGame/pull/6132
Commit Summary
• Fix because the unsafe "fixed" statement crashes if using a char* type like this:
File Changes
• M MonoGame.Framework/Graphics/OpenGL.cs (10) 
Patch Links:
• https://github.com/MonoGame/MonoGame/pull/6132.patch
• https://github.com/MonoGame/MonoGame/pull/6132.diff

jkotas:
I have noticed this bug too. I have not seen the crash, just things going wrong under debugger.

https://github.com/MonoGame/MonoGame/pull/6118 is a simpler more performant fix.

roponator:
@jkotas Unfortunately your fix at #6118 crashes for me: if I use fixed(char* c = str) it will crash.

I get a crash even if I executed fixed(char* c = str) in our game (so outside of the MonoGame lib).

jkotas:
> if I executed `fixed(char* c = str)` in our game code (so outside of the MonoGame lib

This is most likely a buffer overrun or similar memory corruption in your game code. The extra allocation in your proposed fix makes it not crash because of it disrupts the memory layout. The bug is still there hidden and it will cause crashes somewhere else sooner or later.

dellis1972:
I think this is related to https://bugzilla.xamarin.com/show_bug.cgi?id=60625

roponator:
@jkotas @dellis1972  I just confirmed that even creating a new Android Monogame project crashes, so it's not our game.

Here is the repo so you can try it out (assuming it crashes on your phones also) (crash is triggered in Activity1.cs file):
https://github.com/roponator/monogame_xamarin_crash_fixedStatement

roponator:
So based on the comments on those threads for the Mono issue linked by dellis the fix seems to be in Mono 5.8 already, but not in the VS xamarin build (I tried the preview version of Visual Studio but it still crashes).

nkast:
@roponator Can you try to force a 32bit assembly? (no x86_64 or arm64-v8a in Advanced properties)

@dellis1972 
The fix seems to have to do with reading the string address as 32bit and then failed to expand it to 64bit address. 
Maybe we could get the address of (object)string as (void*) and then add the RuntimeHelpers.OffsetToStringData to get to the first character.

dellis1972:
this apparently is down to a difference in the way roslyn is now emitting `conv.u` instead of `conv.i` for `fixed`. The Xamarin.Android runtime was not updated to work with this change, although the next preview should go out. 
So what ever "fix" we do it needs to work on older versions as well as newer ones.

dellis1972:
build

tomspilman:
@dellis1972 - Is this one blocked on me updating Xamarin Android on the build machine?

dellis1972:
I think we should update the build machines. It should work without changes
after that. The problem I believe is the IL that is emitted by the
compilers changed resulting in it being incompatible.. That said I might
have miss-understood

dellis1972:
@roponator the build bots have been updated with the latest stable releases of Xamarin. Can you retest with the development version and see if you still get the crash and if this is required?

SimonDarksideJ:
@roponator do you have the time to retest the functionality on the latest build bots?

roponator:
@SimonDarksideJ 

Sorry no, I don't have access to the stuff for it anymore.

SimonDarksideJ:
Sorry to hear that @roponator (although the change is IN this PR), if change your mind the closed PR will still remain here.  Thanks for your contribution.

PR #6135, opened 2018-01-08
New Template System using Nuget

dellis1972:
This includes a new Addin for VSForMac.


dellis1972:
This will probably fail as I am better the build servers don't have the correct versions of MSBuild or Mono on them.

cra0zy:
@dellis1972 Long story short, if you are updating stuff on build bots, please include dotnet 2.0 as well, need it for the custom msbuild task for MGCB nuget.

cra0zy:
(dotnet msbuild is linked against .NET Core, while standard msbuild is lined against .NET Framework)

ProjectTemplates/DotNetTemplate/content/BlankApp/BlankApp.Android/BlankApp.Android.csproj
```diff
@@ -0,0 +1,104 @@
...
+    <AndroidResource Include="Resources\Drawable\Icon.png" />
+    <AndroidResource Include="Resources\Drawable\Splash.png" />
+    <AndroidResource Include="Resources\Values\Styles.xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <MonoGameContentReference Include="Content\Content.mgcb" />
+  <ItemGroup>
+    <None Include="Properties\AndroidManifest.xml" />
+  </ItemGroup>
+  <Import Project="$(MSBuildExtensionsPath)\Novell\Novell.MonoDroid.CSharp.targets" />
```
cra0zy: This should point to the new Xamarin targets.

cra0zy:
I thought you were gonna go for the project wizard, but truthfully I like this more, I just didn't wanna mention it because of the work you put in the project wizard thing... then again since they are now all split up its probably way easier to setup a special project wizard.

Some quick review:
- Using stuff as `monogame-desktopgl-blank` is not really short, for the one template I created, I went with `mgdesktopgl`, short and to the point
  - short names do not use `-` symbol, ie. `webconfig`, `globaljson`
- Android templates are broken
- Descriptions are missing
- `groupIdentity` tag is used so you can group same templates that target different programming languages, not the way you made it here
- Icons are broken/missing
- Pipeline tool file association is broken and icon is missing

Some notes for @tomspilman and @KonajuGames : 
- We should switch to using nugets in the templates
  - The installer can just preload nugets into the global cache, this way no Internet is required
- We should also switch to using MGCB from nuget, this would preserve backwards compatibility for the projects going forward
  - Creating this nuget does not require any changes to MGCB, it just requires 4 extra nuget packages and an extra task, I have all the changes already done on my fork: https://github.com/cra0zy/MonoGame/tree/corebuilder

dellis1972:
Thanks for the feedback 😊ill try and get those fixed asap

Sent from my Windows 10 phone

IDE/VisualStudioForMac/PipelineDisplayBinding.cs
```diff
@@ -0,0 +1,64 @@
...
+			if (Environment.OSVersion.Platform == PlatformID.Win32NT) {
+				process.StartInfo.FileName = Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.ProgramFilesX86), @"MSBuild\MonoGame\v3.0\Tools", "Pipeline.exe");
+				process.StartInfo.Arguments = string.Format ("\"{0}\"", filename);
+			} else {
+				if (Directory.Exists ("/Applications/Pipeline.app")) {
+					process.StartInfo.FileName = "open";
+					process.StartInfo.EnvironmentVariables.Add ("MONOGAME_PIPELINE_PROJECT", Path.GetFullPath (filename));
+					process.StartInfo.Arguments = string.Format ("-b com.monogame.pipeline --args \"{0}\"", Path.GetFullPath (filename));
+				} else {
+					// figure out linix
```
cra0zy: You should also take into account that MD might be running in flatpak mode on Linux. Currently it does not run/debug applications outside the sandbox, however if they ever got around to implementing that, it could be totally usable. You can see how to launch an application from outside the sandbox here: https://github.com/mono/monodevelop-flatpak/blob/master/patches/monodevelop/sandbox_escape.patch#L207

PS. Yes, MD from flatpak cannot access the system MGCB, however with my nuget the compilation works just fine.

cra0zy:
@dellis1972 Can you split out .NET Core templates into its own PR?

dellis1972:
This PR doesn't have any .Net Core templates.. or do you means the Nuget
ones?
not sure that will help since that project uses the new project format so
we still need the bots upgraded.

tomspilman:
> we still need the bots upgraded.

@dellis1972  - Almost done with that now... 20 minutes.

cra0zy:
> This PR doesn't have any .Net Core templates.. or do you means the Nuget
> ones?
> not sure that will help since that project uses the new project format so
> we still need the bots upgraded.

Sorry, I meant nuget templates, the reason why I'm asking for a split is since its gonna be way easier getting those merged first, and then upgrading the MD addin in here.

Plus there are several things I wanted to contribute to the nuget templates to make them fully usable from the command line (since I will more than likely use them that way), and I don't wanna have to modify the MD addin (I can't compile it from Linux to test it out).

cra0zy:
You still need to fix shortname for the templates.

dellis1972:
@tomspilman @KonajuGames  ok this should be good to go now.

tomspilman:
Thanks @dellis1972 !

PR #6139, opened 2018-01-10
Fixed VertexContent.RemoveRange

tomspilman:
This fixes a bug in `VertexContent.RemoveRange()` and adds some unit tests verifying it works.

Also fixed the XNA Test project which wasn't building.

### DIFF
--- a/Test/ContentPipeline/TestProcessorContext.cs
+++ b/Test/ContentPipeline/TestProcessorContext.cs
@@ -49,10 +49,12 @@ namespace MonoGame.Tests.ContentPipeline
             get { throw new NotImplementedException(); }
         }
 
+#if !XNA
         public override ContentIdentity SourceIdentity
         {
             get { throw new NotImplementedException(); }
         }
+#endif
 
         public override TargetPlatform TargetPlatform
         {
--- a/Test/MonoGame.Tests.XNA.csproj
+++ b/Test/MonoGame.Tests.XNA.csproj
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>MonoGame.Tests</RootNamespace>
     <AssemblyName>MonoGame.Tests.XNA</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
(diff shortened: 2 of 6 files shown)


dellis1972:
Merging

Test/MonoGame.Tests.XNA.csproj
```diff
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>MonoGame.Tests</RootNamespace>
     <AssemblyName>MonoGame.Tests.XNA</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
nkast: I am not sure this works. 
2010 is the last VS that is officially supported by XNA and VS2010 can work with .Net framework up to 4.0.

dellis1972: I would love to know the market share for things like VS2010/12/13 etc. It would be nice to know when we can drop support for those IDE's.

Test/ContentPipeline/TestProcessorContext.cs
```diff
@@ -49,10 +49,12 @@ public override OpaqueDataDictionary Parameters
             get { throw new NotImplementedException(); }
         }
 
+#if !XNA
         public override ContentIdentity SourceIdentity
```
nkast: We should remove the SourceIdentity from ContentProcessorContext and leave it as part of the PipelineProcessorContext. The user can get the source/filename with `((PipelineProcessorContext)context).SourceIdentity`.

In general whenever we extent on the XNA API this way we should do so with adding new interfaces.
like so: `public ContentIdentity IFilebasedContentProcessorContext.SourceIdentity { get; }`

Test/MonoGame.Tests.XNA.csproj
nkast: This is about the XNA tests. 
MG dropped end-user support of VS2010 years ago when upgrading from .net 4.0 to 4.5.

Test/ContentPipeline/TestProcessorContext.cs
Ragath: i guess i could see the point in moving it to the other class, while leaving it public to avoid changing the usage pattern/visibility. But hiding api extensions designed to make the framework easier to use behind otherwise unexposed interfaces defeats the purpose.

Test/MonoGame.Tests.XNA.csproj
tomspilman: I've been using this to run XNA on VS2023/15...

https://mxa.codeplex.com/releases

It looks weird, but works really really well.

The reason for the change is we have some .NET 4.5 only code in the unit tests now.  So it was either remove that code or update the project.

Test/ContentPipeline/TestProcessorContext.cs
tomspilman: We really don't need to make simple extensions like this one that difficult and ugly to use. It only affects someone if they are using it and shared code between MG and XNA...  Like we are for unit tests.

I don't think this needs to change here.

However we could go back and make 'SourceIdentity' virtual with a default implementation instead of abstract.

Test/MonoGame.Tests.XNA.csproj
nkast: Ok , I see. I would bet you had one of those.

PR #6141, opened 2018-01-11
add SingleReader and EffectReader to auto-spidered reader list

mgambrell:
Why don't we incorporate EffectReader by default? I know a lot of games don't need effects, but every single one that doesnt not need effects, does.
SingleReader is tiny, and an obvious counterpart to ArrayReader<float> (I needed it for loading movie files)

NOTE: I made this edit on github. So I don't know whyfor the huge amount of noise. But it should be safe.


tomspilman:
We need someone to clean the whitespace changes up in this PR...  but it is good otherwise.

tomspilman:
@dellis1972 @cra0zy @Jjagg - Anyone have time to redo this PR without all the whitespace changes?

cra0zy:
On it.

PR #6146, opened 2018-01-13
Added O(1) EffectParameter lookups by name

guFalcon:
This adds a dictionary to the immutable EffectParameters class to achieve O(1) lookups by name.
Adds a test that depends on the parameters permanently being on the same index which shouldn't be a problem except when adding parameters.

Added this because it spares you the wrapping of all your shaders because of performance reasons if you have more that one parameter. No need for the dirty-flags any more (at least in trivial shaders where the variables are independent).
Hope that helps.


dellis1972:
@tomspilman @KonajuGames  thoughts?

MonoGame.Framework/Graphics/Effect/EffectParameterCollection.cs
```diff
@@ -7,10 +7,15 @@ public class EffectParameterCollection : IEnumerable<EffectParameter>
         internal static readonly EffectParameterCollection Empty = new EffectParameterCollection(new EffectParameter[0]);
 
         private readonly EffectParameter[] _parameters;
+        private readonly Dictionary<string, EffectParameter> _parameterDictionary = new Dictionary<string, EffectParameter>();
```
tomspilman: If we change this to a `Dictionary<string, int>` we can share the dictionary across all instances of same cloned `EffectParameterCollection` reducing memory usage and generated garbage.

I care because each instance of a `BasicEffect` still eats up 6K each... which is pretty fat overall.  So i want to avoid adding to that if at all possible while still improving performance.

guFalcon: Hm. I understand your concerns.
Didn't occur to me.

IMO the idea with the shared `Dictionary<string,int>` is great. You're trading an O(1) lookup for a 2*O(1) lookup which still is great.

The binary search idea would change the order of `EffectParameter[]` and that would break a lot of implementations wouldn't it?
And still; How would that speed up lookups using the name? You cannot do a binary search on a member of the `Effect`, can you? So you probably have to persist the name somewhere.

Another idea based on yours:
On load build a second array of names corresponding to the order of the `EffectParameter[]`.
Sort it and do binary search, use that index to retrieve it from the `EffectParameter[]`.
Then share that array across cloned instances.
No pointer to `Effect`, no `HashMap` overhead, same lookup time as for the `Dictionary<string,int>`.

Of course one could argue that, if you want a fast lookup, then use a `Dictionary`. It knows what's best for you ;)

Sorry if I didn't get something. It's my first time tinkering with MonoGame code (internal that is).

tomspilman: > The binary search idea would change the order of EffectParameter[] and that would break a lot of implementations wouldn't it?

I don't think it really breaks anything...  you really shouldn't be depending on the order of parameters in the collection.  I know for sure i cannot depend on each console platform returning the same order of parameters as PC does.

At least if they were alphabetical it would be consistent on all platforms.  

The overhead of sorting them at load time should be minor and could be moved into content compile time later when we next bump the MGFX format number.

>  You cannot do a binary search on a member of the Effect, can you?

Yeah you might be right about that.

The shared `Dictionary<string,int>` then sounds like the safest bet with the smallest overhead and best performance.

guFalcon: The reason I brought that up was this here:

I just recently saw that you cannot depend on the name either.
Happened with a `BaseTextureSampler`
image

The problem here was that in GL environments the shader's parameter was called "BaseTextureSampler+BaseTexture" and in DX environments "BaseTexture" like you would expect...

In that case I don't have a choice but to use the index.
Discussed it with @Jjagg and he told me that it's probably safest to rely on the index :)

guFalcon: About the Dictionary... I think I'm gonna ponder this bunch of new parameters (memory-footprint ... pah ;) never heard of such a thing.) for another day or two if you don't mind.
Maybe I can come up with something leaner.

tomspilman: `"BaseTextureSampler+BaseTexture"` should not be happening... we intend for the name to be the same across platforms.  It is something i'm always careful of when porting to a new platform.

We should investigate this issue further.

> I think I'm gonna ponder this bunch of new parameters

That is fine with me.

guFalcon: I'll do a report tomorrow on the name-thing. I have the example here (it's the BloomEffectRenderer solution I shared on nuget). So maybe it could be the awkward way of compiling the shaders when embedding them in the DLL (like MG does).
You want the report on github here, in a new issue or on the MG boards?

tomspilman: @guFalcon - The report should go here on GitHub.  Feel free to link any other relevant information from the MG boards is there is some.

guFalcon: Will do.
Since it's 12:20 am here in Austria, have a nice evening. cu

nkast: >> Another idea based on yours:
>> On load build a second array of names corresponding to the order of the EffectParameter[].
>> Sort it and do binary search, use that index to retrieve it from the EffectParameter[].

SortedDictionary or SortedList works somewhat like that.
It could be faster that Dictionary for few items.

guFalcon: As promised, the issue: #6172

Jjagg:
> Discussed it with @Jjagg and he told me that it's probably safest to rely on the index :)

I reread our discussion and there was a little communication bug here :p I meant the index of the texture in `GraphicsDevice.Textures`, not the effect parameter index.

guFalcon:
So this really looks like it's worth fixing. We cannot rely on the order of the `EffectParameter[]` so we shouldn't use the index, but the 'named' way to access them is linear search. Correct?
Hope to find some time in the evening to come up with something useful.

guFalcon:
Ok.
I pondered a few solutions (Dictionary, parallel arrays, KeyedCollection...)
I considered removing the array altogether and replacing it with the dictionary, but in the end I used the lookup-`Dictionary<string,int>` like already proposed by tom because it saves mem by keeping the lookup separated of the values via a second index-lookup. So we can share that table when cloning.
Hope it's ok. Keep commenting if you have other ideas (I'm in no hurry ;) ).

Test/Framework/Graphics/EffectTest.cs
```diff
@@ -100,5 +101,22 @@ public void EffectPassShouldOverrideTextureIfNotExplicitlySet()
...
+            var effect = new BasicEffect(game.GraphicsDevice);
+            effect.TextureEnabled = true;
+            effect.Texture = null;
+            effect.Parameters["DiffuseColor"].SetValue(Color.HotPink.ToVector3());
+            effect.Parameters["FogColor"].SetValue(Color.Honeydew.ToVector3());
+
+            Assert.That(effect.Parameters[0].GetValueVector3().Equals(Color.HotPink.ToVector3()));
+            Assert.That(effect.Parameters[14].GetValueVector3().Equals(Color.Honeydew.ToVector3()));
+            
+            effect.Dispose();
```
tomspilman: I would add a texture to this test as that is the other class of thing in parameters.

MonoGame.Framework/Graphics/Effect/EffectParameterCollection.cs
```diff
@@ -7,10 +7,21 @@ public class EffectParameterCollection : IEnumerable<EffectParameter>
         internal static readonly EffectParameterCollection Empty = new EffectParameterCollection(new EffectParameter[0]);
 
         private readonly EffectParameter[] _parameters;
+        private readonly Dictionary<string, int> _lookupDictionary = new Dictionary<string, int>();
 
         internal EffectParameterCollection(EffectParameter[] parameters)
         {
             _parameters = parameters;
+            for (int i = 0; i < _parameters.Length; i++)
```
tomspilman: Does `Dictionary` have a constructor that lets you reserve capacity?  This would be better than letting it grow as you add new keys.

Test/Framework/Graphics/EffectTest.cs
guFalcon: Didn't need it for the test, so I deleted it.
Reverting.

MonoGame.Framework/Graphics/Effect/EffectParameterCollection.cs
guFalcon: Yup. Right.
On it.

MonoGame.Framework/Graphics/Effect/EffectParameterCollection.cs
```diff
@@ -7,10 +7,21 @@ public class EffectParameterCollection : IEnumerable<EffectParameter>
         internal static readonly EffectParameterCollection Empty = new EffectParameterCollection(new EffectParameter[0]);
 
         private readonly EffectParameter[] _parameters;
+        private readonly Dictionary<string, int> _lookupDictionary = new Dictionary<string, int>();
```
tomspilman: Do you mind changing this to `_indexLookup` or something that doesn't have `Dictionary` in it?

guFalcon: np.

guFalcon:
Created new PR #6191

PR #6151, opened 2018-01-15
[Pipeline Tool] Restore old debug mode behavior

cra0zy:
Notes:
 - If the debug flag is in .mgcb, it will be preserved, however Pipeline Tool does no longer allow of adding/editing it.

CC. @tomspilman


MarcVador:
The menu item I was looking for a while :)

cra0zy:
@MarcVador  It was just moved under project properties.

tomspilman:
Thanks @cra0zy !

PR #6152, opened 2018-01-16
DXT Compression: Keep the delegate from being relocated

nkast:
Related to #5138


tomspilman:
@nkast - This looks to not work... it throws `System.ArgumentException : Object contains non-primitive or non-blittable data.`.

nkast:
Ok, now I see why it was that way in the first place. We need to try something else.

Issue #6157, opened 2018-01-19
Improve OpenAssetImporter


tomspilman:
Currently the `OpenAssetImporter` is crippled to only expose features that the old XNA `FbxImporter` and `XImporter` supported.  So for example we only expose 4 different texture types in `MaterialContent`:

  - Texture (aka Diffuse)
  - Transparency
  - Specular
  - Bump

However AssImp supports like 12 different types of defined texture maps and even custom ones.

Same sort of thing exists for material properties where we only expose a few... but some useful ones like 'IsTwoSided` is not exposed.

I think we have two paths:

 A) Just add the extra stuff as `OpaqueData` and not worry.
 B) Set a `XnaCompatible` flag to `OpenAssetImporter` when called from the `FbxImporter`.

If you use the `OpenAssetImporter` or `FbxImporter` let me know what you think.

tomspilman:
See https://github.com/MonoGame/MonoGame/pull/6158

Jjagg:
We should also update assimp to the latest stable release (#5397). MG is very far behind.

tomspilman:
> We should also update assimp to the latest stable release (#5397). MG is very far behind.

Agreed.  I tested with it, but waited to update it in the next PR.

The goal being to use this new `FbxImporter` unit tests to help detect any issues from changing it.

vpenades:
@Jjagg I agree on updating Assimp to the latest version, but it seems they're not keeping the NET wrapper up to date; the last official NET wrapper is still the one for Assimp 3, and it's +4 years old. I heard from time to time there's a new NET wrapper in the works, but it's not even in their official repository.

PR #6158, opened 2018-01-19
OpenAssetImporter Improvements and FbxImporter Unit Tests

tomspilman:
This PR is related to https://github.com/MonoGame/MonoGame/issues/6157.

I extended the material support in `OpenAssetImporter` to export all the material features exposed by AssImp.  This breaks anyone that uses `OpenAssetImporter` directly... but hopefully for the better.

`FbxImporter` and `XImporter` now pass a flag to `OpenAssetImporter` asking for XNA compatibility returning just the material properties expected under XNA 4.0.  I added a unit test for `FbxImporter` to help prove that nothing has changed or is broken.

It is debatable if this is necessary or if we're ok passing additional non-XNA material properties from `FbxImporter` and `XImporter` .


tomspilman:
Note this PR adds the old `dude.fbx` to the unit tests.  

image

I think this is acceptable as it was distributed widely as a sample and is pretty useful to us in testing XNA compatibility.

Test/ContentPipeline/FbxImporterTests.cs
```diff
@@ -0,0 +1,249 @@
...
+        // via the Open Asset Importer.
+        //
+        //  - This is annoying for users porting old samples.
+        //  - Conversion could loose data from the original FBX.
+        //  - Do we care to resolve this in the future?
+        //
+#if XNA
+        const string DudeFbx = "Assets/Models/Dude/dude.fbx";
+#else
+        const string DudeFbx = "Assets/Models/Dude/dude_2011.fbx";
```
tomspilman: Note this sucks... but is the only way to test old XNA FBXs.

tomspilman:
Note i tried testing the `NodeContent` from the `FBXImporter` via XML serialization, but ran into two issues:

 - MG's `NodeContent` crashes the `IntermediateSerializer`... XNA's works.
 - There are lots of little differences in the `NodeContent` so can't just do a text compare.

The only solution seemed like the verbose unit testing code above for now.

MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
```diff
@@ -366,6 +384,82 @@ private void ImportMaterials()
...
+                    Identity = _identity,
+                };
+
+                var slots = aiMaterial.GetAllMaterialTextures();
+                foreach (var tex in slots)
+                {
+                    var name = tex.TextureType.ToString();
+                    if (tex.TextureIndex != 0)
+                        name += tex.TextureIndex;
+                    material.Textures.Add(name, ImportTextureContentRef(tex));
```
tomspilman: We're using AssImp's `TextureType` as the naming convention for textures here.

MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
```diff
@@ -366,6 +384,82 @@ private void ImportMaterials()
...
+                if (aiMaterial.HasShadingMode)
+                    material.OpaqueData.Add("ShadingMode", aiMaterial.ShadingMode.ToString());
+                if (aiMaterial.HasShininess)
+                    material.OpaqueData.Add("Shininess", aiMaterial.Shininess);
+                if (aiMaterial.HasShininessStrength)
+                    material.OpaqueData.Add("ShininessStrength", aiMaterial.ShininessStrength);
+                if (aiMaterial.HasTwoSided)
+                    material.OpaqueData.Add("TwoSided", aiMaterial.IsTwoSided);
+                if (aiMaterial.HasWireFrame)
+                    material.OpaqueData.Add("WireFrame", aiMaterial.IsWireFrameEnabled);
```
tomspilman: We're also using AssImp's naming convention for material properties... `ColorDiffuse` vs `DiffuseColor`.

MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
```diff
@@ -366,6 +384,82 @@ private void ImportMaterials()
...
+        {
+            var texture = new ExternalReference<TextureContent>(textureSlot.FilePath, _identity);
+            texture.OpaqueData.Add("TextureCoordinate", string.Format("TextureCoordinate{0}", textureSlot.UVIndex));
+
+            if (!_xnaCompatible)
+            {
+                texture.OpaqueData.Add("Operation", textureSlot.Operation.ToString());
+                texture.OpaqueData.Add("WrapModeU", textureSlot.WrapModeU.ToString());
+                texture.OpaqueData.Add("WrapModeV", textureSlot.WrapModeV.ToString());
+                texture.OpaqueData.Add("Mapping", textureSlot.Mapping.ToString());
```
tomspilman: Note we're also adding AssImp's texture properties which could come in useful.

MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
tomspilman: Still thinking about this...   AssImp's convention comes from their material property keys like `AI_MATKEY_COLOR_DIFFUSE`.  

Thinking now that we're better off with a more readable naming than a C++ naming convention designed for editor autocompletion..

tomspilman:
Note this is still a WIP.  Do not merge.

MonoGame.Framework.Content.Pipeline/OpenAssetImporter.cs
tomspilman: AssImp's material docs:

 http://assimp.sourceforge.net/lib_html/materials.html
 http://assimp.sourceforge.net/lib_html/material_8h.html

Jjagg: IMO we don't have to stick with assimp naming. 99.9% of the people using this won't be familiar with the assimp API anyway. We just have to make sure to document the mapping in case people want to reference the assimp docs.

tomspilman: Yeah i think i will make the naming more XNA like and spend some time writing a doc page on the details of the `OpenAssetImporter`.

tomspilman:
This is good to merge now...  assuming there are no objections.

KonajuGames:
This is a good move forward.  Merging.

dxclancejz:
Is there anyway to fix the OBJ Waveform importer? The FBX importer does not replace it. That only works if the model was exported from a model program that also uses FBX and that makes using OBJ pointless. Some apps such as MagicaVoxel do no support FBX, so it does not work.

Issue #6036, opened 2017-10-31
[SharpDX] Problems with DXT render targets


Sizaar:
Creating a render target with SurfaceFormat.Dxt1/3/5 or SurfaceFormat.Dxt1/3/5SRgb results in an SharpDX exception "The parameter is incorrect", here is the stack trace:
```
   at SharpDX.Result.CheckError()
   at SharpDX.Direct3D11.Device.CreateTexture2D(Texture2DDescription& descRef, DataBox[] initialDataRef, Texture2D texture2DOut)
   at Microsoft.Xna.Framework.Graphics.TextureCube.CreateTexture()
   at Microsoft.Xna.Framework.Graphics.TextureCube..ctor(GraphicsDevice graphicsDevice, Int32 size, 
Boolean mipMap, SurfaceFormat format, Boolean renderTarget)
   at Microsoft.Xna.Framework.Graphics.RenderTargetCube..ctor(GraphicsDevice graphicsDevice, Int32 size, 
Boolean mipMap, SurfaceFormat preferredFormat, DepthFormat preferredDepthFormat, Int32 preferredMultiSampleCount, RenderTargetUsage usage)
```

tomspilman:
My guess is XNA doesn't allow this either, but likely throws a nicer exception when you do.

GPUs cannot render to a compressed format.  So the error is correct.

Sizaar:
Actually XNA allows this, checked with a DXT1 Albedo buffer and it works.

tomspilman:
We will have to setup a unit test against XNA and figure out how they are doing it.  GPUs don't render to block compressed formats.  All i can think of is XNA added some code under the hood to render to a regular surface the compress it.

Jjagg:
I think XNA falls back to a supported format. The parameter is called `prefferedFormat`. XNA has other places where it uses 'preferred' as a prefix to mean that it will try to use the given setting but falls back to something else if needed (like in GraphicsDeviceManager).

tomspilman:
> I think XNA falls back to a supported format. 

Interesting... thanks for that info @Jjagg .

We should probably setup a unit test for this and ensure we behave like XNA so that people are not surprised by an error here where XNA just works.

vpenades:
Graphics Cards, in general, don't support DXT (or any other compression format) to be used as render targets.

I think the right solution would be to throw a "not supported exception" whenever you setup such render target.

KonajuGames:
[...]

PR #6160, opened 2018-01-22
Implement GraphicsAdapter.QueryRenderTargetFormat()

nkast:
Fallback of unsuported and profile depended SurfaceFormat. (Dxt, Single, drBlendable, ...)

Not implemeneted: 
Optional/Device depended formats. (566, 4444, ...)
Depth Formats.
MultiSampleCount.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.Legacy.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.Legacy.cs
@@ -146,7 +146,17 @@ namespace Microsoft.Xna.Framework.Graphics
         /// </remarks>
         public static DriverType UseDriverType { get; set; }
 
-        /*
+        /// <summary>
+        /// Queries for support of the requested render target format on the adaptor.
+        /// </summary>
+        /// <param name="graphicsProfile">The graphics profile.</param>
+        /// <param name="format">The requested surface format.</param>
+        /// <param name="depthFormat">The requested depth stencil format.</param>
+        /// <param name="multiSampleCount">The requested multisample count.</param>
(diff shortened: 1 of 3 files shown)


dellis1972:
This works on GL based systems too right?

nkast:
Yes, this works on all platforms in a way compatible with XNA. So far I doesn't depend on the device.

After it's merged we can use it to fix #6036 and also pass the tests of https://github.com/MonoGame/MonoGame/pull/5451/commits/b26f9270aa21745576ac8f0ab0d5c0fee33476b9.

nkast:
If we are going to use this QueryRenderTargetFormat() in the RenderTarget/Texture construction it will change the behavior depending on the Profile.  For example, `HdrBlendable` will become `Color` under `Reach` profile. I think I should move some of this code in #5451.

The way this is going work on the Texture construction:  
```
bool isSupported = device.adapter.QueryRenderTargetFormat(device.Profile, prefferedSurfaceFormat, out surfaceFormat);
if(!isSupported)
{
if(this is IRenderTarget)
 surfaceFormat = prefferedSurfaceFormat;
else
 throw new Exception(device.Profile + "does not support surface "+ prefferedSurfaceFormat);
}
```
That won't be a good error message because some formats (DXT, NormalizedByte2) are not supported as RenderTargets in any profile. 
So I think I will post a fix for #6036 that does not really on QueryRenderTargetFormat() but does some checks inline and throws a better error.

nkast:
Ok, I have a new PR #6170 that deals with #6036 without the Profile fallback.

nkast:
#6170 is merged. I moved the rest to #5451

PR #6167, opened 2018-01-26
Fix loading of DXT compressed surfaces that are not multiples of 4.

tomspilman:
It looks that in the performance fix in #6003 broke `SetData` for surfaces that are not a multiple of 4.  Specificlly i think this optimization broke it back in October:

https://github.com/MonoGame/MonoGame/commit/bd22d49e9cc8db4d1731f602c53f20276332da82#diff-b1287b3ea3e4f14da31f425817f5920eR178

This caused DXT compressed textures to have black colored mip levels at 2x2 and 1x1 or odd sizes that were not multiple of 4.

I fixed this in the `SetData` under DirectX.  In theory it could be broken on OpenGL and other platforms that use DXT.

I added a unit test to catch this bug in the future and can help us find the issue on other platforms.


KonajuGames:
Thanks @tomspilman. Merging.

PR #6170, opened 2018-01-30
Fallback SurfaceFormat for RenderTargets

### LINKED ISSUE #6036: [SharpDX] Problems with DXT render targets

nkast:
fixes #6036

This PR partially implements Adapter.QueryRenderTargetFormat().

The use of QuerySelectedFormat() is not perfect. We need to rethink how RenderTargets call the base Texture constructor.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsAdapter.Legacy.cs
+++ b/MonoGame.Framework/Graphics/GraphicsAdapter.Legacy.cs
@@ -170,6 +170,19 @@ namespace Microsoft.Xna.Framework.Graphics
             selectedDepthFormat = depthFormat;
             selectedMultiSampleCount = multiSampleCount;
 
+            // fallback for unsupported renderTarget surface formats.
+            if (selectedFormat == SurfaceFormat.Alpha8 ||
+                selectedFormat == SurfaceFormat.NormalizedByte2 ||
+                selectedFormat == SurfaceFormat.NormalizedByte4 ||
+                selectedFormat == SurfaceFormat.Dxt1 ||
+                selectedFormat == SurfaceFormat.Dxt3 ||
+                selectedFormat == SurfaceFormat.Dxt5 ||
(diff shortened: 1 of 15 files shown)


dellis1972:
@nkast should we be using the data from the OpenGL extensions to decide if we support certain formats on the GL platforms?

nkast:
For now this PR deals only with compressed formats (DXT,Alpha8,NormalizedByte) that are not supported as renderTargets. There is no need to query the adapter because GPUs don't compress textures on the fly. There resulting format should always be Color.

After that I will look into implementing the rest.
* Newer formats depend on the Profile. For example a combination of Reach + Single will return Color (XNA). I will add this in my Reach PR once we have this PR merged.
* Optional formats like the 16bit bgr565 and bgr4444 will require to query the adapter via OpenGL extensions or by creating a temp Device under DirectX. Or we can phase those formats out and return Color.
* preferredMultiSampleCount is the second case where we need to query the adapter.

tomspilman:
Thanks @nkast !

PR #6183, opened 2018-02-12
FIX: SpriteFont.TryGetGlyphIndex could return an unavailable glyph

MichaelDePiazzi:
`SpriteFont.TryGetGlyphIndex` had a bug in it which meant it could mistakenly succeed for an unavailable glyph and give an invalid index. This could result in unexpected behaviour when drawing a string which contained unavailable glyphs. Because other dependant code then uses pointers to reference the glyph, it could potentially have some nasty side effects.

This PR fixes this bug, as well as adding some test cases which expose it (the test cases which include characters before 0x20 / 32 fail with the previous code).

(this issue was originally introduced in PR #5874, cc @nkast)

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteFont.cs
+++ b/MonoGame.Framework/Graphics/SpriteFont.cs
@@ -260,12 +260,7 @@ namespace Microsoft.Xna.Framework.Graphics
                     }
                     else if (pRegions[m].Start > c)
                     {
-                        r = m;
-                        if (l == r)
-                        {
-                            regionIdx = l;
-                            break;
-                        }
+                        r = m - 1;
                     }
                     else
                     {
--- a/Test/Framework/Graphics/SpriteFontTest.cs
+++ b/Test/Framework/Graphics/SpriteFontTest.cs
@@ -569,21 +569,32 @@ But the answer was still '42'.
             CheckFrames();
 		}
 
-		[Test]
-		public void Throws_when_drawing_unavailable_characters ()
+        [TestCase("The rain in España stays mainly in the plain - now in français")]
+        [TestCase("\x1f")]
+        [TestCase("\x7f")]
+        public void Throws_when_drawing_unavailable_characters(string text)
 		{
-			const string text = "The rain in España stays mainly in the plain - now in français";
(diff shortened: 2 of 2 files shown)


nkast:
I missed that one. 
Thanks @MichaelDePiazzi

MichaelDePiazzi:
@tomspilman The build server doesn't seem to be kicking off for some reason...?

nkast:
Here's something interesting ...
When I unrolled the loop and swap the checks I saw a +12% performance (AMD Athlon II).
I suspect this makes better use of branch prediction. Simply unrolling without the swap doesn't do much.

```
                while (l <= r)
                {
                    var m = (l + r) >> 1;
                    Debug.Assert(m >= 0 && m < _regions.Length, "Index was outside the bounds of the array.");
                    if (pRegions[m].End < c)
                        l = m + 1;
                    else if (pRegions[m].Start > c)
                        r = m - 1;
                    else
                    {
                        regionIdx = m;
                        break;
                    }

                    // unroll once and flip left/right check
                    if(l > r)
                        break;

                    m = (l + r) >> 1;                    
                    Debug.Assert(m >= 0 && m < _regions.Length, "Index was outside the bounds of the array.");
                    if (pRegions[m].Start > c)
                        r = m - 1;
                    else if (pRegions[m].End < c)
                        l = m + 1;
                    else
                    {
                        regionIdx = m;
                        break;
                    }
                }
```

tomspilman:
Thanks @MichaelDePiazzi !

MichaelDePiazzi:
@nkast Definitely interesting... If you'd like to try prove if it's related to branch prediction, then you may want to check out [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet). There have been a couple of new features introduced which allow you to measure branch prediction failures. I've been meaning to mess around with them, but haven't had a chance yet. Here are a couple of articles though which talk about it more and show what you can do:

- http://adamsitnik.com/Hardware-Counters-Diagnoser/
- http://adamsitnik.com/Disassembly-Diagnoser/

The disassembly diagnoser in particular looks very cool as you can combine it with the hardware counters to show misprediction rates for each instruction!

PR #6185, opened 2018-02-13
Android | Added more key maps and stopped claiming to handle unhandled keys.

Romans-I-XVI:
Unhandled keys were being claimed has having been handled. This made it
impossible to catch the key inputs elsewhere (via the Xamarin OnKey*
override methods)

Also added some more standard key mappings.

### DIFF
--- a/MonoGame.Framework/Android/Input/Keyboard.cs
+++ b/MonoGame.Framework/Android/Input/Keyboard.cs
@@ -88,7 +88,7 @@ namespace Microsoft.Xna.Framework.Input
             maps[Keycode.DpadRight] = Keys.Right;
             maps[Keycode.DpadUp] = Keys.Up;
             maps[Keycode.DpadDown] = Keys.Down;
-			maps[Keycode.DpadCenter] = Keys.Enter;
+            maps[Keycode.DpadCenter] = Keys.Enter;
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Android/Input/Keyboard.cs
```diff
@@ -120,18 +124,60 @@ internal static void KeyUp(Keycode keyCode)
 			maps[Keycode.W] = Keys.W;
 			maps[Keycode.X] = Keys.X;
             maps[Keycode.Y] = Keys.Y;
-            maps[Keycode.C] = Keys.Z;
-			maps[Keycode.Back] = Keys.Escape;
+            maps[Keycode.Z] = Keys.Z;
+            maps[Keycode.Space] = Keys.Space;
+			maps[Keycode.Escape] = Keys.Escape;
```
dellis1972: formatting looks a bit off here? spaces vs tabs?

cra0zy: formatting is off in that file in general

Romans-I-XVI:
It was actually some of the original code that was using tabs instead of spaces. Actually the entire file has the occasional smattering of tabs instead of spaces, but I just changed the ones in the maps[] section.

PR #6189, opened 2018-02-14
Fix Pipeline tool to work regardless of Mono changes.[DONT MERGE]

dellis1972:
One of the problems we have with the Pipeline tool
at the moment on Mac is this. When running the MGCB.exe
using system `mono` we can end up in a situation where
the wrong versions of the system assemblies and mscorlib
are loaded. As a result some calls to the system will
fail because entry points are missing or are invalid.

This is an attempt to fix that by bundling the `mono`
we are built against inside the Pipeline.app package.
In theory this should help.

I tried setting `MONO_PATH` and various other mono related
flags but nothing worked.

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -515,6 +515,8 @@ namespace MonoGame.Tools.Pipeline
                 _buildProcess.StartInfo.StandardOutputEncoding = encoding;
                 _buildProcess.OutputDataReceived += (sender, args) => View.OutputAppend(args.Data);
 
+                View.OutputAppend(string.Format("Working Directory: {0} ", _buildProcess.StartInfo.WorkingDirectory));
+                View.OutputAppend(string.Format("Running: {0} {1}", _buildProcess.StartInfo.FileName, _buildProcess.StartInfo.Arguments));
                 // Fire off the process.
                 _buildProcess.Start();
                 _buildProcess.BeginOutputReadLine();
--- a/Build/Projects/Pipeline.definition
+++ b/Build/Projects/Pipeline.definition
@@ -358,6 +358,16 @@
       <Platforms>MacOS</Platforms>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="..\..\ThirdParty\Dependencies\FreeImage.NET\MacOS\libfreeimage.dylib">
+      <Platforms>MacOS</Platforms>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Link>libfreeimage.dylib</Link>
+    </Content>
+    <Content Include="/Library/Frameworks/Mono.framework/Commands/mono">
+      <Platforms>MacOS</Platforms>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Link>mono</Link>
+    </Content>
   </Files>
 
 </Project>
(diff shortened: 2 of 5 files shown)


Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -515,6 +515,8 @@ private void DoBuild(string commands)
                 _buildProcess.StartInfo.StandardOutputEncoding = encoding;
                 _buildProcess.OutputDataReceived += (sender, args) => View.OutputAppend(args.Data);
 
+                View.OutputAppend(string.Format("Working Directory: {0} ", _buildProcess.StartInfo.WorkingDirectory));
+                View.OutputAppend(string.Format("Running: {0} {1}", _buildProcess.StartInfo.FileName, _buildProcess.StartInfo.Arguments));
```
tomspilman: Is there a concern the working directory and launch EXE could be bad?

dellis1972: nope, I thought it might be useful information to report.. I can take that out

tomspilman:
Could you not get the command line which launched the Pipeline tool itself to find Mono?

dellis1972:
@tomspilman I want to test the package we get before we merge this , to see if it fixes the issue I have.

dellis1972:
@tomspilman the problem isn't finding mono, but getting mono to load the assemblies from its location and not the `Pipeline.app` bundle. (Which is what its doing now). When you end up with if that happens is mono complains that the version of mono doesn't match. The compile works most of the time, but the only way to fix it is to get a matching mono. Since our build bots are not always on the very latest , users running the latest mono will get these problems.

tomspilman:
> getting mono to load the assemblies from its location and not the Pipeline.app bundle. 

Oh... so the Pipeline.app bundle includes Mono assemblies and doesn't use them from the installed system.  I guess that is different about Mono on Mac/Linux over C# on Windows which uses the installed version.

> When you end up with if that happens is mono complains that the version of mono doesn't match. 

But if we use the Mono that launched Pipeline then it has to always match no?

dellis1972:
This is the error you get 

`
cant resolve internal call to "System.Net.Sockets.Socket::Send_internal(intptr,byte[],int,int,System.Net.Sockets.SocketFlags,int&,bool)" (tested without signature also)

Your mono runtime and class libraries are out of sync.
The out of sync library is: /Applications/Pipeline.app/Contents/MonoBundle/System.dll

When you update one from git you need to update, compile and install
the other too.
Do not report this as a bug unless you're sure you have updated correctly:
you probably have a broken mono install.
If you see other errors or faults after this message they are probably related
and you need to fix your mono install first.
`

dellis1972:
this fix didn't work anyway. So I guess I need to rethink.

dellis1972:
@tomspilman its when we call `MGCB.exe` that the problem starts, not in the Pipeline tool itself.
We run `/Library/Frameworks/Mono.framework/Versions/Current/bin/mono <path to mgcb.exe>` to run the compile. But that mono seems to prefer to load the assemblies like System.dll and mscorlib.dll from `/Applications/Pipeline.app/Contents/MonoBundle/System.dll` and that causes the problem. 

Like I said most of the time its ok.. but every now and then boom :/

cra0zy:
Can't you just delete assemblies from MonoBundle and always use system ones?

dellis1972:
no because the `Pipeline` executable in Contents/MacOS will be compiled against the `mscorlib` version in `MonoBundle`. If we remove them the Pipeline tool probably won't run, unless the user has the exact version of `mscorlib` installed. The only part of the Pipeline.app which needs mono installed are the calls to MGCB.exe , otherwise its stand alone. I'll keep looking at this.

tomspilman:
Can we bundle mono with MGCB then?

Issue #6184, opened 2018-02-13
Can't load multiple effects from a same folder


MarcVador:
Hi !
Since this morning, after having installer the build 1255,
 I have not been able to load 2 effects from the same folder, it throws an ArgumentException, saying "An element with the same key has already been added". Ex:
If I do:
```
_SSLFA_DownsampleHx16Effect = content.Load<Effect>("Shaders/PostProcess/AnamorphicLensFlares/downsample-horizontal-x16");
_SSLFA_AnamorphicBlurEffect = content.Load<Effect>("Shaders/PostProcess/AnamorphicLensFlares/anamorphic-blur");
```
It throws on the second effect, as if it stopped before the last /, keeping only the folder as the asset's key.

If I load just one of the 2 effects, it loads without problem. So I've tried with my other effects and as soon I have a folder with more than one effect, it throws the exception.

MarcVador:
After some investigation by using directly the sources of Monogame with my game, I found the exact line of code:

 It's in when adding to _indexLookup, it adds an empty string:

```
internal EffectParameterCollection(EffectParameter[] parameters)
        {
            _parameters = parameters;
            _indexLookup = new Dictionary<string, int>(_parameters.Length);
            for (int i = 0; i < _parameters.Length; i++)
            {
                _indexLookup.Add(_parameters[i].Name, i);
            }
        }
```

Exception:
>  à System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
>    à System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
>    à Microsoft.Xna.Framework.Graphics.EffectParameterCollection..ctor(EffectParameter[] parameters) dans H:\_____MG\____MonogameSource\MonoGame\MonoGame.Framework\Graphics\Effect\EffectParameterCollection.cs:ligne 18
>    à Microsoft.Xna.Framework.Graphics.Effect.ReadParameters(BinaryReader reader) dans H:\_____MG\____MonogameSource\MonoGame\MonoGame.Framework\Graphics\Effect\Effect.cs:ligne 455
>    à Microsoft.Xna.Framework.Graphics.Effect.ReadParameters(BinaryReader reader) dans H:\_____MG\____MonogameSource\MonoGame\MonoGame.Framework\Graphics\Effect\Effect.cs:ligne 403
[...]

PR #6191, opened 2018-02-15
Fix proposal for added O(1) EffectParameter lookups by name #6146

guFalcon:
Since @Alkher pointed out in #6184 that `EffectParameters` also get allocated for arrays (and maybe even lines of array declarations), I prepared this fix.

IMO it would be better to not add those to the `EffectParameterCollection` in the first place, I lack the experience to see the full consequences of such a change (and the lack of tests doesn't do any good either).
So with this fix everyone would be able to use shaders containing (constant) arrays again.


Jjagg:
Let's get this merged! :)

tomspilman:
I'll merge this for now... but we should find the root cause of this issue and fix it.

guFalcon:
@Alkher tested this and it seems to work for now.
I'll take a look at it tomorrow. Maybe I can do the 'real' fix.
It'll be new code for me, but you all are a great source of programming wisdom and a great help, so I feel confident. :smile:

PR #6197, opened 2018-02-20
 Fix Pipeline tool to work regardless of Mono changes

dellis1972:
One of the problems we have with the Pipeline tool
at the moment on Mac is this. When running the MGCB.exe
using system `mono` we can end up in a situation where
the wrong versions of the system assemblies and mscorlib
are loaded. As a result some calls to the system will
fail because entry points are missing or are invalid.

This is an attempt to fix that by bundling the `mono`
we are built against inside the Pipeline.app package.
In theory this should help. We also look for the system
`mono` (on a Mac) before anything else. 
We also change the CurrentDirectory when we start MGCB to
ensure that we don't try to load any assemblies from the
Pipeline.app/Contents/MonoBundle directory. This was based
on advice from the Xamarin.Mac team.

### DIFF
--- a/Tools/Pipeline/MainWindow.cs
+++ b/Tools/Pipeline/MainWindow.cs
@@ -28,7 +28,8 @@ namespace MonoGame.Tools.Pipeline
         private string[] monoLocations = {
             "/usr/bin/mono",
             "/usr/local/bin/mono",
-            "/Library/Frameworks/Mono.framework/Versions/Current/bin/mono"
+            "/Library/Frameworks/Mono.framework/Versions/Current/bin/mono",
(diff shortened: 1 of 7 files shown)


dellis1972:
It does.

Tools/Pipeline/MainWindow.cs
```diff
@@ -332,6 +336,8 @@ public Process CreateProcess(string exe, string commands)
                     OutputAppend("Could not find mono. Please install the latest version from http://www.mono-project.com");
                 }
 
+                OutputAppend("Found mono at " + monoLoc);
```
cra0zy: Ah, a misuse of the build output.

dellis1972:
right now this should finally be good to go.

tomspilman:
Thanks @dellis1972 !

PR #6198, opened 2018-02-20
Switch order of parameters in Ray.Deconstruct

KonajuGames:
The parameters in the type deconstructors should match the order of the parameters in the constructors.

Completes #6096

### DIFF
--- a/MonoGame.Framework/Ray.cs
+++ b/MonoGame.Framework/Ray.cs
@@ -259,12 +259,12 @@ namespace Microsoft.Xna.Framework
         /// <summary>
         /// Deconstruction method for <see cref="Ray"/>.
         /// </summary>
-        /// <param name="direction"></param>
-        /// <param name="position"></param>
-        public void Deconstruct(out Vector3 direction, out Vector3 position)
+        /// <param name="position">Receives the start position of the ray.</param>
+        /// <param name="direction">Receives the direction of the ray.</param>
+        public void Deconstruct(out Vector3 position, out Vector3 direction)
         {
-            direction = Direction;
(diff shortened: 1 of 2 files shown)


dellis1972:
@KonajuGames looks like the unit tests failed.

`[MonoGameTests.exe] MonoGame.Tests.Framework.RayTest.Deconstruct
[09:40:03]
[MonoGame.Tests.Framework.RayTest.Deconstruct] Test(s) failed.   Expected: {X:0 Y:0 Z:1}
  But was:  {X:1 Y:0 Z:0}
[09:40:03]
[MonoGame.Tests.Framework.RayTest.Deconstruct] Test(s) failed.   Expected: {X:0 Y:0 Z:1}
  But was:  {X:1 Y:0 Z:0}

   at NUnit.Framework.Assert.That(Object actual, IResolveConstraint expression, String message, Object[] args)
   at MonoGame.Tests.Framework.RayTest.Deconstruct() in c:\BuildAgents\MonoGameWin2\work\59a4b66548a5cef5\Test\Framework\RayTest.cs:line 59
[09:40:06][Step 1/1] Failed tests detected`

KonajuGames:
That's why we have unit tests. :) Will fix when I get home.

Jjagg:
> That's why we have unit tests.

Though it's the test that should be fixed in this case :p

KonajuGames:
> Though it's the test that should be fixed in this case :p

Well, yes, of course.  I overlooked the test and hadn't run them locally first.  Fix incoming...

dellis1972:
Merged :)

PR #6200, opened 2018-02-20
Fix MRT support when using MSAA

prollin:
When using MSAA, the first rendertarget was attached to all color attachements.


Jjagg:
@prollin You need a cast to `IRenderTarget` because `GetFramebufferTarget` and `GLTexture` are explicit interface implementations.

Jjagg:
LGTM!

dellis1972:
Thanks @prollin , nice to see commits form you again :D

Issue #5257, opened 2016-10-28
Pipeline throws InvalidOperationException: 'This operation is only valid on generic types.'


nkast:
The exception comes from this line:
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ContentCompiler.cs#L55

This is the code that cause it, it works fine with XNA.
```
    [ContentTypeWriter]
    public class InheritableWriter : ContentTypeWriter
    {        
        public InheritableWriter() : base(typeof(MyContent)) { }
        public InheritableWriter(Type targetType) : base(targetType) { }

        protected override void Write(ContentWriter output, object value)
        {
            var content = (MyContent)value;
            // write content...
        }
    }
```

tomspilman:
I'll setup a unit tests which replicates this bug this week and see if i can fix it.

nkast:
One of the things I was trying to do was to extend Texture2D and it's processor/ContentTypeWriter. 

```
public class Sprite : Texture2D
{
    int OriginalTextureWidth;
    int OriginalTextureHeight;
}

And then load as a sprite or a texture:
```
var txA = content.Load<Texture2D>("spriteA");
var sprite1 = (sprite)txA;
var sprite2 = content.Load<Sprite>("spriteA");
```
The above is possible under XNA.

However just building and referencing the simple class `InheritableWriter ` in an ,mgcb will fail during the initialization. I could find a framework to run an .mgcb files from tests.

PR #6206, opened 2018-02-23
Added ability to load images(png/jpg/bmp/gif/tga) into the memory

rds1983:
This PR will allow a developer to do following:
```c#
int width, height;
byte[] data;
using(var stream = File.Open("myImage.jpg"))
{
  Texture2D.RawImageFromStream(stream, out width, out height, out data);
}

// Various processing
// ...

// Create Texture2D from it
var texture = new Texture2D(GraphicsDevice, width, height);
texture.SetData(data);

```
I think it is helpful feature. As one can easily do various processing to image(i.e. premultiply alpha) before creating Texture2D from it.

Of course, same could be archieved using Texture2D.GetData. But it would require more code. Also I am unsure that all platforms support it.

### DIFF
--- a/MonoGame.Framework/Graphics/Image2D.cs
+++ b/MonoGame.Framework/Graphics/Image2D.cs
@@ -3,16 +3,16 @@ using Microsoft.Xna.Framework.Graphics;
 using StbImageSharp;
 using System;
 using System.IO;
-using System.Runtime.InteropServices;
 
 namespace MonoGame.Framework.Graphics
 {
     /// <summary>
     /// 2D Image in the <see cref="SurfaceFormat.Color"/> format
     /// </summary>
-    public class Image2D
+    public class Image2D : IDisposable
     {
-        private readonly byte[] _data;
+        private readonly byte[] _pixelData;
+        private bool _isDisposed;
 
         /// <summary>
         /// Width of the image
@@ -25,11 +25,11 @@ namespace MonoGame.Framework.Graphics
         public int Height { get; private set; }
 
         /// <summary>
-        /// Data of the image in the RGBA format
+        /// Data of the image in the RGBA format.
         /// </summary>
-        public byte[] Data
+        public byte[] PixelData
         {
-            get { return _data; }
+            get { return _pixelData; }
         }
 
         /// <summary>
@@ -43,100 +43,83 @@ namespace MonoGame.Framework.Graphics
             get
             {
                 var basePos = (y * Width + x) * 4;
-                return new Color(_data[basePos],
-                    _data[basePos + 1],
-                    _data[basePos + 2],
-                    _data[basePos + 3]);
+                return new Color(_pixelData[basePos],
+                    _pixelData[basePos + 1],
+                    _pixelData[basePos + 2],
+                    _pixelData[basePos + 3]);
             }
 
             set
             {
                 var basePos = (y * Width + x) * 4;
-                _data[basePos] = value.R;
-                _data[basePos + 1] = value.G;
-                _data[basePos + 2] = value.B;
-                _data[basePos + 3] = value.A;
+                _pixelData[basePos] = value.R;
+                _pixelData[basePos + 1] = value.G;
+                _pixelData[basePos + 2] = value.B;
+                _pixelData[basePos + 3] = value.A;
             }
         }
 
-
         /// <summary>
-        /// Constructs a <see cref="Image2D"/> of the specified size
+        /// Constructs a <see cref="Image2D"/> of the specified size,
         /// </summary>
         /// <param name="width"></param>
         /// <param name="height"></param>
         public Image2D(int width, int height)
         {
             if (width <= 0)
-            {
                 throw new ArgumentOutOfRangeException("width");
-            }
-
             if (height <= 0)
-            {
                 throw new ArgumentOutOfRangeException("height");
-            }
 
             Width = width;
             Height = height;
-
-            _data = new byte[width * height * 4];
+            _pixelData = new byte[width * height * 4];
         }
 
         /// <summary>
-        /// Constructs a <see cref="Image2D"/> of the specified size and containing specified data
+        /// Constructs a <see cref="Image2D"/> of the specified size and containing specified data.
         /// </summary>
+        /// <param name="pixelData"></param>
         /// <param name="width"></param>
         /// <param name="height"></param>
-        /// <param name="data"></param>
-        public Image2D(int width, int height, byte[] data)
+        public Image2D(byte[] pixelData, int width, int height)
         {
-            if (width <= 0)
-            {
-                throw new ArgumentOutOfRangeException("width");
-            }
-
-            if (height <= 0)
-            {
-                throw new ArgumentOutOfRangeException("height");
-            }
-
-            if (data == null)
-            {
+            if (pixelData == null)
                 throw new ArgumentNullException("data");
-            }
+            if (width <= 0)
+                throw new ArgumentOutOfRangeException("width");
+            if (height <= 0)
+                throw new ArgumentOutOfRangeException("height");
 
-            var length = width * height * 4;
-            if (data.Length != length)
+            int length = width * height * 4;
+            if (pixelData.Length != length)
             {
                 throw new ArgumentException(string.Format("Inconsistent data length: expected={0}, provided={1}", length,
-                    data.Length));
+                    pixelData.Length));
             }
 
             Width = width;
             Height = height;
-            _data = data;
+            _pixelData = pixelData;
         }
 
         /// <summary>
-        /// Creates Texture2D from this image
+        /// Creates Texture2D from this image.
         /// </summary>
-        /// <param name="graphicsDevice"></param>
-        /// <returns></returns>
         public Texture2D CreateTexture2D(GraphicsDevice graphicsDevice)
         {
             var texture = new Texture2D(graphicsDevice, Width, Height, false, SurfaceFormat.Color);
-            texture.SetData(_data);
+            texture.SetData(_pixelData);
             return texture;
         }
 
         /// <summary>
-        /// Loads an image in PNG, JPG, BMP, GIF or TGA format from the stream
+        /// Loads an image in PNG, JPG, BMP, GIF or TGA format from the stream.
         /// </summary>
-        /// <param name="stream"></param>
-        /// <returns></returns>
         public unsafe static Image2D FromStream(Stream stream)
         {
+            // TODO: reduce allocations here by not using seeking at all
+
             ImageResult result;
             if (stream.CanSeek)
             {
@@ -152,8 +135,26 @@ namespace MonoGame.Framework.Graphics
                     result = ImageResult.FromStream(ms, ColorComponents.RedGreenBlueAlpha);
                 }
             }
+            return new Image2D(result.Data, result.Width, result.Height);
+        }
 
-            return new Image2D(result.Width, result.Height, result.Data);
+        protected virtual void Dispose(bool disposing)
+        {
+            // Dispose is empty for now but may be utilized in the future.
+
+            if (!_isDisposed)
+            {
+                if (disposing)
+                {
+                }
+                _isDisposed = true;
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(disposing: true);
+            GC.SuppressFinalize(this);
         }
     }
 }
--- a/MonoGame.Framework/Graphics/Image2D.cs
+++ b/MonoGame.Framework/Graphics/Image2D.cs
@@ -118,6 +118,12 @@ namespace MonoGame.Framework.Graphics
         /// </summary>
         public unsafe static Image2D FromStream(Stream stream)
         {
+            // Rewind stream if it is at end
+            if (stream.CanSeek && stream.Length == stream.Position)
+            {
+                stream.Seek(0, SeekOrigin.Begin);
+            }
+
             // TODO: reduce allocations here by not using seeking at all
 
             ImageResult result;
--- a/MonoGame.Framework/Graphics/Image2D.cs
+++ b/MonoGame.Framework/Graphics/Image2D.cs
@@ -137,12 +137,6 @@ namespace MonoGame.Framework.Graphics
         /// <returns></returns>
         public unsafe static Image2D FromStream(Stream stream)
         {
-            // Rewind stream if it is at end
-            if (stream.CanSeek && stream.Length == stream.Position)
-            {
-                stream.Seek(0, SeekOrigin.Begin);
-            }
-
             ImageResult result;
             if (stream.CanSeek)
             {
--- a/MonoGame.Framework/Graphics/Image2D.cs
+++ b/MonoGame.Framework/Graphics/Image2D.cs
@@ -1,6 +1,6 @@
 ﻿using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
-using MonoGame.Utilities;
+using StbImageSharp;
 using System;
 using System.IO;
 
@@ -126,14 +126,62 @@ namespace MonoGame.Framework.Graphics
         /// </summary>
         /// <param name="stream"></param>
         /// <returns></returns>
-        public static Image2D FromStream(Stream stream)
+        public unsafe static Image2D FromStream(Stream stream)
         {
-            int x, y;
-            int comp;
+            byte[] bytes;
 
-            var data = ImageReader.ReadColor(stream, out x, out y, out comp);
+            // Rewind stream if it is at end
+            if (stream.CanSeek && stream.Length == stream.Position)
+            {
+                stream.Seek(0, SeekOrigin.Begin);
+            }
 
-            return new Image2D(x, y, data);
+            // Copy it's data to memory
+            // As some platforms dont provide full stream functionality and thus streams can't be read as it is
+            using (var ms = new MemoryStream())
+            {
+                stream.CopyTo(ms);
+                bytes = ms.ToArray();
+            }
+
+            byte* result = null;
+            try
+            {
+                // Load image
+                int x, y, comp;
+                fixed (byte* b = bytes)
+                {
+                    result = StbImage.stbi_load_from_memory(b, bytes.Length, &x, &y, &comp, StbImage.STBI_rgb_alpha);
+                }
+
+                if (result == null)
+                {
+                    throw new InvalidOperationException(StbImage.LastError);
+                }
+
+                // Convert to managed color array
+                byte* src = result;
+                var data = new Color[x * y];
+                fixed (Color* dest = data)
+                {
+                    for (var i = 0; i < data.Length; ++i)
+                    {
+                        dest[i].R = *src++;
+                        dest[i].G = *src++;
+                        dest[i].B = *src++;
+                        dest[i].A = *src++;
+                    }
+                }
+
+                return new Image2D(x, y, data);
+            }
+            finally
+            {
+                if (result != null)
+                {
+                    CRuntime.free(result);
+                }
+            }
         }
     }
 }
--- a/MonoGame.Framework/Graphics/Image2D.cs
+++ b/MonoGame.Framework/Graphics/Image2D.cs
@@ -12,7 +12,7 @@ namespace MonoGame.Framework.Graphics
     /// </summary>
     public class Image2D
     {
-        private readonly Color[] _data;
+        private readonly byte[] _data;
 
         /// <summary>
         /// Width of the image
@@ -25,9 +25,9 @@ namespace MonoGame.Framework.Graphics
         public int Height { get; private set; }
 
         /// <summary>
-        /// Data of the image
+        /// Data of the image in the RGBA format
         /// </summary>
-        public Color[] Data
+        public byte[] Data
         {
             get { return _data; }
         }
@@ -42,12 +42,20 @@ namespace MonoGame.Framework.Graphics
         {
             get
             {
-                return _data[y * Width + x];
+                var basePos = (y * Width + x) * 4;
+                return new Color(_data[basePos],
+                    _data[basePos + 1],
+                    _data[basePos + 2],
+                    _data[basePos + 3]);
             }
 
             set
             {
-                _data[y * Width + x] = value;
+                var basePos = (y * Width + x) * 4;
+                _data[basePos] = value.R;
+                _data[basePos + 1] = value.G;
+                _data[basePos + 2] = value.B;
+                _data[basePos + 3] = value.A;
             }
         }
 
@@ -72,7 +80,7 @@ namespace MonoGame.Framework.Graphics
             Width = width;
             Height = height;
 
-            _data = new Color[width * height];
+            _data = new byte[width * height * 4];
         }
 
         /// <summary>
@@ -81,7 +89,7 @@ namespace MonoGame.Framework.Graphics
         /// <param name="width"></param>
         /// <param name="height"></param>
         /// <param name="data"></param>
-        public Image2D(int width, int height, Color[] data)
+        public Image2D(int width, int height, byte[] data)
         {
             if (width <= 0)
             {
@@ -98,7 +106,7 @@ namespace MonoGame.Framework.Graphics
                 throw new ArgumentNullException("data");
             }
 
-            var length = width * height;
+            var length = width * height * 4;
             if (data.Length != length)
             {
                 throw new ArgumentException(string.Format("Inconsistent data length: expected={0}, provided={1}", length,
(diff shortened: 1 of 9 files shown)


Jjagg:
I like the feature, but not the API design. I don't think this functionality should be in Texture2D. IMO it makes more sense to expose the ImageReader itself (albeit with slightly modified naming to follow standard C# convention.

> Of course, same could be archieved using Texture2D.GetData. But it would require more code. Also I am unsure that all platforms support it.

From a performance perspective this would also be a lot worse, because data is sent to VRAM, pulled back to RAM, modified and reuploaded to VRAM.

Jjagg:
> Btw, I am also thinking, maybe return Color[] instead of byte[]?

Agreed!

rds1983:
Now it returns Color[]

Jjagg:
Do we want this in the `MonoGame.Utilities` namespace? Other than that this looks good to me!

Jjagg:
I think after this (not in this PR) it also makes sense to expose the ImageWriter. A nice use case is saving screenshots.

```
void SaveScreenshot(GraphicsDevice gd)
{
    var pp = gd.PresentationParameters;
    var bbd = new byte[pp.BackBufferWidth * pp.BackBufferHeight * 4];
    gd.GetBackBufferData(bbd);
    using (var stream = File.OpenWrite("screenshot.png"))
    using (var writer = new ImageWriter())
        writer.Write(bbd, pp.BackBufferWidth, pp.BackBufferHeight, 4, ImageWriterFormat.Png, stream);
}
```

Currently you'd need an external image writer or create a texture first (send the image data back to the GPU) and use one of the `Texture2D.SaveAs*` methods.

tomspilman:
@dellis1972  @KonajuGames - What do you guys think of this?

I am a little nervous adding a new API like this as official as it starts locking us into this API/implementation.

Still for sure we should once and for all decide how we want to mark new non-XNA APIs.  

Also we should not merge this until we're done with 3.7 release.

cra0zy:
I don't think its a good idea to add this API to MonoGame, it would make way more sense for the end user to reference something like ImageSharp if they want to load a raw image to ram and do processing on it.

Jjagg:
My reasoning is that we have all the required classes included in the assembly already, they're just not exposed. Requiring users to include a lib that does the same thing seems unnecessary to me if we can just expose what we're using internally. But yeah, if this does get merged we really need to make clear somehow that it isn't XNA API.

cra0zy:
Well, what I dislike about this is its API design, it allows you to load raw image data, but not raw Texture2D data (I mean from the .xnb files directly). I think it would make a lot of sense to create something like `RawTexture2D` class that is meant for loading textures into ram instead of vram.

cra0zy:
Here is my API proposition:
```c#
using System;
using System.IO;

// Its in MonoGame namespace so it tells you
// from the start that it was not part of XNA
namespace MonoGame.Framework.Graphics
{
    public class RawTexture2D : Texture
    {
        private byte[] _data;

        public static RawTexture2D FromStream(Stream stream);

        public Texture2D ToTexture2D();

        public void SetData<T>();

        public T GetData<T>();
    }
}
```

tomspilman:
> Here is my API proposition

Interesting.  So it is still a `Texture` object so a `RawTexture2D` is still a "real texture" and can be used for rendering if you want.

Thinking of it like this you're essentially proposing a texture that keeps a system memory copy of the loaded data and does lazy creation of the GPU texture resources.

A good thing with this design is it makes this feature usable with `ContentManager`.  So you do `ContentManager.Load<RawTexture2D>("mytexture")` if you want a system memory copy of the texture and `ContentManager.Load<Texture2D>("mytexture")` if you want the texture on the GPU only.  This is certainly cleaner than other ideas we've had for this functionality.

I would suggest if we took this approach we inherit from  `public class RawTexture2D : Texture2D` and just set some flags within `Texture2D` to keep a system memory copy and lazy initialize the texture resources.  This way we reuse all the existing `Texture2D` logic and nothing internal to MG has to change.

I'm not sold on the name `Raw` here...  but i think the idea could be the right one for us.

tomspilman:
> If we had it as an Image class (not deriving from Texture2D)

Yeah we could do that too.  The question is how important is it that we can use this object interchangeably with textures?  I don't think this is required... the user is just trying to get pixel data back.  You can easily create a new texture from that data if you like.  

However this would be a nice feature for real textures.  Consider the situation i see in lots of pixel art games.  They load a texture from XNB... they colorize it... then use it to render.  This has to send the original data to the GPU, fetch it back from the GPU to the CPU, the player modifies it then sends it to the GPU again.  If `Texture2D` did lazy load to the GPU and kept a system memory copy all this would be way more efficient with just a single line change from `Texture2D` to `RawTexture2D` or whatever.

One possible problem now that i think of it is how internally `ContentManager` caches.  It assumes only one `object` is associated to a particular path passed to it.  So if you did like `ContentManager.Load<Image2D>("mytexture")` and then `ContentManager.Load<Texture2D>("mytexture")` it would likely do something weird/bad.  This is the side effect of the idea of one XNB could resolve to multiple asset types, but i am pretty sure we could handle it by salting the internal key with the type name.

> `SystemTexture2D` is the only other one that comes to mind,

I think we need to think about this some more.  The name here is critical to explaining the idea well to users.

> Otherwise we also do `SystemTexture3D`, `SystemTextureCube` as well.

Well unless this new `Image` type internally supported 3D and Cube image data we would still need multiple types.

rds1983:
We also need to take into account existence of various SurfaceFormats.
Should Image2D support arbitrary SurfaceFormat or only SurfaceFormat.Color?
Also what should happen if `texture.SetData(image2d)` when texture and image2d have different SurfaceFormat?

I propose adding following class:
```c#
namespace MonoGame.Framework.Graphics
{
    public class ColorBuffer
    {
        private Color[] _data;
        private int _width, _height;

        public Color this[int index]
        {
          get { return _data[index]; }
          set { _data[index] = value; }
        }

        public static ColorBuffer FromStream(Stream stream);

        public Texture2D ToTexture2D(GraphicsDevice device);
    }
}

It would be loadable from ContentManager. However it would throw an exception if incoming data has different SurfaceFormat than SurfaceFormat.Color or mipmap level count different than 1.

Also I dont think it would be an issue of someone calling `Content<ColorBuffer>("myTexture")` then `Content<Texture2D>("myTexture")`. As if someone chooses loading through ColorBuffer than he's likely to create texture manually.

**Comment by dellis1972** [comment]
Just an idea for a different approach, how about we extend `Texture2D.FromStream` to take additional parameters. eg.

```
enum TextureFromStreamOptions {
     Default = 0,
     PreMultiplyAlpha = 1,
// add more
}

public static Texture2D FromStream (
         GraphicsDevice graphicsDevice,
         Stream stream,
         TextureFromStreamOptions options = TextureFromStreamOptions.Default,
         Func<Color[], Color[]> postProcessing = null,
)
```

the `postProcessing` function gets called if not null. Passes in the Color[] of the texture and returns the Color[] after processing.

**Comment by nkast** [comment]
Maybe we don't need the Options. Instead the implementation of Default & PreMultiplyAlpha can be public, 

code samples
```
var tx1 = Texture2D.FromStream(gd, pngstream1,  TexturePostProcessing.PreMultiplyAlpha);
var tx1 = Texture2D.FromStream(gd, pngstream1,  TexturePostProcessing.GrayScale);
var tx2 = Texture2D.FromStream(gd, pngstream2,  this.MyCustomppFunc);
```

**Comment by cra0zy** [comment]
@dellis1972 The problem than becomes, how do you load raw data for an .xnb file.

**Comment by KonajuGames** [comment]
We already have `Texture2D.GetData` that can get the data for a file loaded from .xnb file, considering we're talking about 2D textures specifically.  I think this discussion ventured off the original topic by suggesting loading XNB data for `Texture2D` directly into system memory.

**Comment by cra0zy** [comment]
> We already have Texture2D.GetData that can get the data for a file loaded from .xnb file, considering we're talking about 2D textures specifically. I think this discussion ventured off the original topic by suggesting loading XNB data for Texture2D directly into system memory.

You can also `Texture2d.FromStream` and than use `GetData` to load image into memory, we are talking about bypassing the need to push data to the GPU so I don't think its ventured off the original topic at all.

**Comment by rds1983** [comment]
Actually a property `Func<Color[], Color[]> ColorTexture2DPostProcessing { get;set; }` could be added to the ContentManager class.
Which would be applied if incoming texture is SurfaceFormat.Color.
So adding alpha premultiplying to xnbs code would be:
```c#
Content.ColorTexture2DPostProcessing = TexturePostProcessing.PreMultiplyAlpha;
```

Though that solution seems a bit ugly.

KonajuGames:
> Though that solution seems a bit ugly.

Asset type-specific properties or parameters in `ContentManager` is not where we want to go.

nkast:
For .xnb you can write a processor to load data. 

However you can't inherit from texture and do something like `class RawTexture2D:Texture2D`.
Because of a bug in the pipeline (#5257). In XNA you could overwrite the ContentTypeWriter and it's RuntimeReader/RuntimeType but in MG it doesn't work, so you are not able to call var rtx = content.Load<RawTexture2D>("rawTexture2D"). The closest you can get to this is with composition `class RawTexture2D { public Texture2D Tex; }`.

You most likely will do most of the processing in the processor. by the way I'd like to see an extension of the pipeline with filters. Now we have `file.ext  -[import]-> extContent -[proccess]-> .xnb`
which force us to duplicate most of the code in a new processor.
as long as the output content type matches the input content type we can allow this:
`file.ext  -[import]-> ExtContent -[filter]-> ExtContent -[proccess]-> .xnb`

cra0zy:
> However you can't inherit from texture and do something like class RawTexture2D:Texture2D

The only thing that would need to be done is to modify the Texture2DReader.

nkast:
Sure, with a few `if`s  here and there you can add a stock reader. 
I was thinking more about letting the user add his own types and processors without having to duplicate the entire Texture3DWriter/Reader.

cra0zy:
Now that is actually off topic.

nkast:
@cra0zy 
> Now that is actually off topic.

whatever pal.
Have a nice day.

cra0zy:
Eh, I'm sorry, I didn't mean to offend you.

rds1983:
btw, concerning extended Texture2D.FromStream.
As far as I understand that feature would require closed platforms to be updated too, wouldn't it?

rds1983:
Following our discussion I've changed API of the new feature by adding new parameter to the Texture2D.FromStream. Now it looks like following:
```c#
        public static Texture2D FromStream(GraphicsDevice graphicsDevice, Stream stream, Func<Color, Color> postProcessing = null)
```

If postProcessing is not null then it is used before data send to GPU.
Maybe this will work.

Jjagg:
@rds1983 I really don't like this solution. You can only do simple pixel transformations using that API. You really want to have all pixels and the dimensions of the texture accessible in your transformation function. But IMO then it makes more sense to have a separate class for in-memory images that users can use and operate on rather than to allow passing a transformation function to `Texture2D.FromStream`. 

From the proposed ideas, I prefer introducing an `Image` class (or `Image2D`) that holds the pixel data (exposed so users can mutate), the dimensions and the pixel format. It should have a `FromStream` method and a `ToTexture2D` method that takes a `GraphicsDevice` and creates a `Texture2D`. Nothing has to change to `Texture2D` itself.

rds1983:
Well, I like Image2D solution.
However it raises some questions.
Should it support arbitrary SurfaceFormat or only SurfaceFormat.Color?
Because in first case Image2D.FromStream should take SurfaceFormat as parameter and it would be a pain to convert stb output to all possible SurfaceFormats.
If Image2d supports only SurfaceFormat.Color then it would much easier. Though maybe Image2d isn't correct name in such case(as Image word supposes support for arbitrary pixel formats). I'd go with ColorBuffer2d class having following api:
```c#
    public class ColorBuffer2d
    {
        private Color[] _data;
        private int _width, _height;

        public int Width { get { return _width; } }
        public int Height { get { return _height; } }
   
        public Color this[int index]
        {
          get { return _data[index]; }
          set { _data[index] = value; }
        }

        public static ColorBuffer2d FromStream(Stream stream);

        public Texture2D ToTexture2D(GraphicsDevice device);
    }
```

Also, what namespace should proposed class reside? Microsoft.Xna.Framework.Graphics?

Jjagg:
Oh right, since stb outputs RGBA32 we should support just the `Color` surface format. I still prefer `Image2D` over `ColorBuffer` though. IMO it describes the class better because to me a buffer is just something that holds data (so not 2D). I like your proposed API. I'd add an indexer that takes an x and y coordinate.

> Also, what namespace should proposed class reside? Microsoft.Xna.Framework.Graphics?

It should start with MonoGame because it's not XNA API. `MonoGame.Framework.Graphics` maybe?

rds1983:
I added Image2D class.
Hopefully this will work.

rds1983:
Synchronized this PR with the latest code base.

Jjagg:
@rds1983 I'm gonna pull out your last commit, because it's breaking the SDK-style csproj UWP build on develop. (we don't build it yet on the build bot, so this slipped through)

rds1983:
Ok, I'll revert that commit then.

Jjagg:
Make sure you rebase after #6843 goes through :)

rds1983:
Sure.

rds1983:
Rebased.

rds1983:
Is there any chance this PR will be merged in 3.8?

Jjagg:
I'm good with this addition, but I'd like @tomspilman to approve before merging.

harry-cpp:
Can we name the class something better than `Image2D`, I feel the end users might get confused and try to use it incorrectly due to its name.

rds1983:
@harry-cpp,

I think users might get confused by the fact Image2D resides in Microsoft.Xna.Graphics namespace. Maybe it makes sense to move it under MonoGame.Utilies. It'll lessen the confusion and also explicitly mark this class as extension to XNA.

harry-cpp:
You can place it inside `MonoGame.Framework.Graphics` namespace.

That being said, a class name should represent what the class does and `Image2D` does not do a good job at that.

rds1983:
@harry-cpp,
Can you propose better name?

harry-cpp:
`MemoryTexture2D`?

Also I just re read the old comments and realized, there is something I suggested before which would be very very useful and that is making the class a child of a texture class. Why is this useful? Very simple, you could do stuff like:

```c#
var texture = Content.Load<MemoryTexture2D>("asd");
```

tomspilman:
Hey @rds1983 @Jjagg @harry-cpp .

> I'm good with this addition, but I'd like @tomspilman to approve before merging.

I agree we need a way for people to load raw image data from files or content.

I also agree we need a way to transform this data into a GPU texture object.

I think this solves both of these things in a reasonable way.


> Can you propose better name?

I think it is important to not call this a texture.

For XNA "Texture" specifically means a GPU object.  This new class isn't a GPU object (at least how the code is written now) and shouldn't use the name texture.

Image isn't a bad name really in this context i think.

> making the class a child of a texture class.

This is a reasonable suggestion.  I guess it comes down to what this class is solving.

I think there are cases where someone wants to load possibly compressed image data and manipulate it some game specific way.  I also think there are times when they do this and don't need a GPU texture object.  For example using an image to define where procedural content is placed in a game world.

So something like `MemoryTexture2D`... which inherits from `Texture`... sort of assumes all image data is used for texturing on the GPU... which i don't think is the case.  **Unless** the texture creation is lazy... meaning it won't allocate the texture until someone tries to assigning it to the `GraphicsDevice`. 

The other path to get to what @harry-cpp suggested is to allow this...
```
var image = Content.Load<Image2D>("asd");
var texture = Content.Load<Texture2D>("asd");
```
This would allow the same "data" can be loaded in different ways as long as we have readers that can accomplish this.  Internally the cache in `ContentManager` would have to consider these different resources... which i think it easily doable.  However how much use would we get out of this beyond textures?  Any other scenarios you can think of?

Lets discuss this a bit more... but i think we're close here.

TechnologicalPizza:
Not to interrupt this PR but I've put some work into an [imaging library](https://github.com/TechnologicalPizza/MonoGame/tree/develop/MonoGame.Imaging). I've also posted an [issue](https://github.com/MonoGame/MonoGame/issues/6886) regarding imaging (it shows off some API and describes how the library works internally). 
It does use `Span<T>` practically everywhere so making it into a PR won't really work. 
Maybe you'll find some use of it regardless.

I started making that library when `ImageSharp` declared their license change. I really liked their API so my library should be pretty familiar API-wise.

harry-cpp:
> I started making that library when `ImageSharp` declared their license change.

What license did they change to?

harry-cpp:
Oh, no, I just read it... damn...

rds1983:
Frankly I think this PR is good as it is.
Yes, it lacks any Content Pipeline support.
But it could be added by separate PRs.

LaserYGD:
Will this be merged?

Jjagg:
I'm fine with merging this. 
@tomspilman @harry-cpp

MonoGame.Framework/Graphics/Image2D.cs
```diff
@@ -0,0 +1,187 @@
...
+            texture.SetData(_data);
+            return texture;
+        }
+
+        /// <summary>
+        /// Loads an image in PNG, JPG, BMP, GIF or TGA format from the stream
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <returns></returns>
+        public unsafe static Image2D FromStream(Stream stream)
```
harry-cpp, line 129: Wait a second, you are reading a stream, than copying a stream to a memory stream, than copying memory stream to a byte array, than copying byte array to a color array.

rds1983:
> Still missing the content reader.

Well, frankly, I am too lazy to add Image2DImporter/Processor/Reader. Moreover I am pretty sure that it wouldn't be used by anyone. Why load pure image through Content Pipeline, when you can call Image2D.FromStream?

So I suggest to simply clearly indicate that it's not supported by the Content Pipeline yet in the Image2D summary. Also maybe create new issue ("Image2D Content Pipeline Support") if this PR gets merged.

Alternative approach is to change the new feature's API to something like:
```c#
public static class Texture2DExt
{
  public static void ImageFromStream(Stream stream, out int width, out int height, out byte[] data);
}
```
That's how FNA addresses this problem.

MonoGame.Framework/Graphics/Image2D.cs
rds1983: Well, the comment in the method code clearly indicates why stream is being copied to a memory stream.
Some platforms(like Xamarin) FileStream lack seek functionality, which caused #6507.
So in order to fix once and for all, I decided to copy whole stream into array of bytes and load it from there.

rds1983:
Did some optimizing and now Image2D.FromStream makes way less allocations than before.

As for Image2D without ContentReader, the possible solution is to remove Image2D class and switch to API like this:
```c#
public static class DrawingUtils
{
  public static void ImageFromStream(Stream stream, out int width, out int height, out byte[] data);
}
```

If it is acceptable solution, I'd implement it. @harry-cpp

harry-cpp:
> As for Image2D without ContentReader

You got that around, we want a ContentReader for it.

MonoGame.Framework/Graphics/Image2D.cs
```diff
@@ -0,0 +1,165 @@
...
+                result = ImageResult.FromStream(stream, ColorComponents.RedGreenBlueAlpha);
+            }
+            else
+            {
+                // If stream doesnt provide seek functionaly, use MemoryStream instead
+                using (var ms = new MemoryStream())
+                {
+                    stream.CopyTo(ms);
+                    ms.Seek(0, SeekOrigin.Begin);
+                    result = ImageResult.FromStream(ms, ColorComponents.RedGreenBlueAlpha);
```
harry-cpp, line 158: What is the idea behind making a memory stream, does StbImageSharp need to seek on the stream to read stuff?

rds1983: Yes. It is used in method stbi__skip.

harry-cpp: But, why don't you just modify stbi__skip to ask if the stream is seekable, if it is, move X bytes, otherwise, read X bytes.

(assuming it never goes backwards)

rds1983: It goes backwards. I just looked code again and there's a method stbi__rewind that rewinds stream to the beginning.

harry-cpp: Hmm, looking at the source of StbImageSharp, the only thing that actually needs backwards seeking is for figuring out the image type: https://github.com/StbSharp/StbImageSharp/blob/88b9c906dd5769d7684b6ef9227af881c00eeb9c/src/StbImage.Generated.Common.cs#L75

I'm fine with your new changes, however it would be very good if you could cash the few bytes from the header so you don't need a seeable stream.

TechnologicalPizza:
I used a handy stream wrapper to solve backwards-seeking; [PrefixedStream](https://github.com/TechnologicalPizza/MonoGame/blob/develop/MonoGame.Utilities/IO/PrefixedStream.cs).
You create an instance, giving it the amount to read upon construction. You can then use `GetPrefix()`.

The only hard thing is that every Stb decoder needs to provide a `HeaderSize` value, then you have to choose the largest `HeaderSize` of all the formats and pass that to the `PrefixedStream` constructor.

Here's a (probably not so useful) [showcase of logic](https://github.com/TechnologicalPizza/MonoGame/blob/930f900163b6bdfd8d17a7c7a3a9a7a16e25b1ec/MonoGame.Imaging/Containers/Image/Image.EnumerateFrames.cs#L21) (Cloning the source and looking around should show a very clear picture)

rds1983:
From the glance, it seems to me that PrefixedStream uses features that arent supported by netstandard2.0/net452(which are current MG target platforms).

TechnologicalPizza:
Making it compatible should only involve removing async code and replacing `Memory<byte>` with `byte[]`

rds1983:
Well, it's good to know there is a way to avoid copying file contents to memory if a stream can't seek(which is true for Xamarin assets stream unfortunately). Though implementing that way would definitely take some effort.

TechnologicalPizza:
My fork of Stb and the MonoGame.Imaging library support non-seekable streams, but I've also very heavily changed the Stb code to my liking (progress reporting, progressive result, vectorization, more OOP), so we can't just copy changes over. 
This works great for unseekable streams like network streams and many other. 
What I'm trying to say is that implementing this is absolutely possible.

Maybe we should finalize this API and then do another PR for optimizations? I wouldn't mind adding support for this (as I've got some experience after making that imaging library), the problem is that MonoGame would need Stb forks as format detection and seeking there have to be tweaked for this to work.

rds1983:
@TechnologicalPizza,
Be aware that this PR might never get merged, since it seems that @harry-cpp expects me to write a ContentReader for Image2d. Which is never gonna happen, since - mildly speaking - I dislike the Content Pipeline.

So if you really want, you may go ahead and make your own PR, which will be improved version of this one: with ContentReader and custom fork of StbImageSharp that supports reading from non-seekable streams.

TechnologicalPizza:
A problem I see with adding content readers that can load Texture2D and Image2D is that Image2D would have to support decoding compressed textures, right? It would be pretty user-unfriendly to throw a NotSupportedException when trying to load a compressed texture with this API.

harry-cpp:
@TechnologicalPizza There is a huge difference between loading a texture2d and image2d, texture2d needs to send the data to the gpu, and if the texture is compressed in some format, it expects the gpu to understand the format.

On the other hand image2d never goes to the gpu, its meant for you to work with the image so it can just decompress the image on the cpu if its compressed (tho we could also not allow compression of images if you want to load them with the content reader).

> Be aware that this PR might never get merged, since it seems that @harry-cpp expects me to write a ContentReader for Image2d. Which is never gonna happen, since - mildly speaking - I dislike the Content Pipeline and not gonna contribute in it.

Honestly, the content reader I was expecting you to write is just for the color format, which is ~10 lines long :/

harry-cpp:
Oh, just looking at the tex2d reader, and it already has conversion code, so you can pretty much just copy over the tex2dreader things into an image2dreader. Yes this thing counts as more lines of code than 10 XD.

TechnologicalPizza:
But there is only a DxtUtil for Dxt1/3/5, no other formats like Etc and Pvrtc.

TechnologicalPizza:
Yikes, that was not the kind of rebase I was looking for.
Regardless, if there was to be a Image2DReader, how would it be able to read an existing Texture2D XNB?
I don't see how you could possibly have multiple readers for one type,
unless i also have to add processors to the pipeline which somewhat defeats the purpose of this PR.

rds1983:
Fixed the rebase. Now it shows only 8 changes files, instead of 116.

rds1983:
Closing since https://github.com/MonoGame/MonoGame/pull/7369 addresses the issue.

PR #6213, opened 2018-02-24
Removed Kickstart submodule

cra0zy:


tomspilman:
@dellis1972 - Do we not use this?

dellis1972:
Not directly in this Repo. It was mainly there for people to link too. But I think that is done in the docs now.

Issue #6221, opened 2018-03-02
AccessViolationException in MGCB.exe when trying to import simple model from Blender


andrekoehler:
I created a simple model in Blender 2.79a: Create cube; set simple material; subdivide one face into a 3x3 grid; extrude grid segment in the middle of it.

![blender](https://user-images.githubusercontent.com/22099714/36904623-7368ada6-1e31-11e8-939a-ed7e05b06380.PNG)

When I start the MonoGame Pipeline Tool, create a new MonoGame Content Build Project, add the Blender file with the Cube and press rebuild, it does NOT generate any output but does not show any error message either:
![screenshot](https://user-images.githubusercontent.com/22099714/36904279-93ba5222-1e30-11e8-97d1-0e729c1e3f2c.PNG)

When I create a new "MonoGame Cross Platform Desktop Project" in Visual Studio 2017, add the same Blender model file to the Content.mgcb file and Rebuild the C# project, I get the following build error:

> 1>------ Rebuild All started: Project: PipelineCrash, Configuration: Debug Any CPU ------
> 1>  C:/Users/Andre/Desktop/PipelineCrash/Content/Box.blend
> 1>  
**> 1>  Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.**
> 1>     at Assimp.Unmanaged.AssimpLibrary.ImportFile(String file, PostProcessSteps flags, IntPtr fileIO, IntPtr propStore)
> 1>     at Assimp.AssimpContext.ImportFile(String file, PostProcessSteps postProcessFlags)
> 1>     at Microsoft.Xna.Framework.Content.Pipeline.OpenAssetImporter.Import(String filename, ContentImporterContext context)
> 1>     at Microsoft.Xna.Framework.Content.Pipeline.ContentImporter`1.Microsoft.Xna.Framework.Content.Pipeline.IContentImporter.Import(String filename, ContentImporterContext context)
> 1>     at MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.ProcessContent(PipelineBuildEvent pipelineEvent)
> 1>     at MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.BuildContent(PipelineBuildEvent pipelineEvent, PipelineBuildEvent cachedEvent, String eventFilepath)
[...]

Issue #5736, opened 2017-05-25
Monogame Pipeline Tool -- freetype6.dll obsolete dependancy on VC++ 2012 Redistributable


mdrejhon:
My project was building before but now with a fresh Windows install, there is a problem cropping up:

I have a fresh Windows 10 Creator's Update install with a fresh Visual Studio 2017 install + MonoGame 3.6

I am getting the same type of error as:
https://github.com/MonoGame/MonoGame/issues/4504

**System.DllNotFoundException: Unable to load DLL 'freetype6.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)**

Even though the file freetype6.dll is there (in the same folder as MGCB.exe)

Several solutions (in 2 places online) say this error is solved by installing the VC++ 2012 Redistributable. 
 However, Microsoft has permanently removed VC++ Redistributable 2012 from their website, which eliminates my ability to fix this error.

It seems that it appears there is a problem with MGCB building with SpriteFonts on a clean Windows 10 Creators Update install (reformat from scratch) with a clean Visual Studio 2017 install (Community Edition).   This system is quite clean of VC++ 2012 redistributable files as a result (few software packages installed except latest dev tools).  For a long time, a hell lot of developer tools installed the 2012 redistributables. so it did not matter, but now it's 2017, and it appears, for the first time, none of my software packages installed the 2012 redistributables .....and it's also gone from Microsoft's website. As a result, forces me into a Catch-22.

When I try to download [Microsoft VC++ 2012 Redistributable](http://www.microsoft.com/en/download/details.aspx?id=30679):
Or from any microsoft links via Google, Microsoft says:

------------
![image](https://cloud.githubusercontent.com/assets/1774867/26470040/373f3268-416a-11e7-933e-62bdb0135a04.png)
-------------

Am I the first 'fresh' Windows 10 Creator Update install (reformat) + VS2017-only install to attempt to build with SpriteFonts?

mdrejhon:
I have tried increasing number of packages to install, and this is how far I am now in Visual Studio 2017 Installer:

![image](https://cloud.githubusercontent.com/assets/1774867/26470989/b8784e16-416d-11e7-9c16-2a4becd64860.png)

It did indeed reduce two MGCB.exe errors when I installed the VC++ toolsets, but the freetype6.dll error remains when I try to build with SpriteFont content.

Jjagg:
Related to #4485.

mdrejhon:
Eureka!
[...]

PR #6222, opened 2018-03-03
Updated Assimp to latest version

tomspilman:
This PR is for updating our version of Assimp to the latest version.

To do this we forked Assimp into the MG org:
 
  https://github.com/MonoGame/assimp

And we build it on our build server:

  http://teamcity.monogame.net/viewType.html?buildTypeId=Dependencies_AssImpWindows&guest=1

One important detail for this version is that it is built against the static VC++ runtime which means no extra VC++ DLLs are required to be on the system.  This keeps things [like this](https://github.com/MonoGame/MonoGame/issues/5736) from happening.

Note this PR depends on https://github.com/MonoGame/MonoGame.Dependencies/pull/120 being merged first.

Fixes https://github.com/MonoGame/MonoGame/issues/6221.


tomspilman:
Ok this is good to go now too.

I'll work on getting Mac built on the build server this week and figure out our part list for a dedicated Linux build box.

PR #6223, opened 2018-03-03
 [Pipeline Tool] VS like linking support

cra0zy:
This uses commits from #5574 for the MGCB support.

A lot of code wasn't designed to handle linking, so I had to rewrite quite a bit of it. Stuff done:
- Added VS like linking support
  - This required for me to rewrite the old include files/folders methods and actions, and update a bunch more code
  - NewAction has been merged with IncludeAction since it was way too similar
- Fixed LocalizedFontProcessor
- Removed "file watcher" stuff, it never worked well, and we can just change the text to red instead of the icon later on once someone gets around to fixing it

Important notes:
- Rename for now only changes the destination link instead of touching the files
- ~~Windows uses embedded icons for the linked files, would really like to draw the link on top of the proper icon, but my main PC is dead~~


tomspilman:
@cra0zy - I'll grab this stuff tomorrow and review it.

What does the .mgcb file look like in these link cases?

Can you update the doc on MGCB to mention how links work:  https://github.com/MonoGame/MonoGame/blob/develop/Documentation/mgcb.md

Can you tell from the Pipeline IDE that the file is linked and from where?

When you delete a linked file i assume it never really deletes the original correct?

cra0zy:
> What does the .mgcb file look like in these link cases?

It uses mgcb stuff from: #5574

> Can you tell from the Pipeline IDE that the file is linked and from where?

You get an icon indicating that its a link, and you can also see its actual location in the property grid:

screenshot from 2018-03-02 15-27-10

> When you delete a linked file i assume it never really deletes the original correct?

We have separate Delete/Excude actions.

tomspilman:
> We have separate Delete/Excude actions.

Lets make sure `Delete` is never an option for linked files.  We have had enough problems with deleting stuff and people getting mad.

In fact... we need to fix Delete to always use the recycle bin on desktops just in case.

cra0zy:
> Lets make sure Delete is never an option for linked files. We have had enough problems with deleting stuff and people getting mad.

Sure, I was actually gonna make "Delete" and "Exclude" options combined into a "Delete" option, which would show a delete dialog which has the "Exclude" option highlighted. That way its very hard to accidentally delete something.

> In fact... we need to fix Delete to always use the recycle bin on desktops just in case.

That requires some per platform code, C# does not expose that type of API AFAIK

tomspilman:
> Sure, I was actually gonna make "Delete" and "Exclude" options combined into a "Delete" option

That would be fine for a new PR after this one.  Still i feel we really should never ever allow delete on a linked file.   It is sort of like when you make a shortcut on your desktop deleting it doesn't delete the original.  Or deleting a symbolic link doesn't delete the original.  

Deleting links to content outside of the project path should't really delete the file.

> That requires some per platform code, 

Yeah it likely will involve some native calls...  but totally worth the effort for the few people that loose tons of work accidentally.

cra0zy:
I avoided touching the `ExcludeAction` in this PR, I would recommend that I make changes to it in a separate PR.

tomspilman:
@cra0zy - Some some issues i found testing the pipeline tool with this PR.

 - You cannot remove linked files... "Exclude from Project" does nothing.
 - Undo seems broken with linked files.
 - If i right click in a folder to add a linked file the liked file always appears in the root.
 - The icon for a linked file is a cyan dot?
 - Rename on a linked file produces a second link instead of updating the original link then doesn't save correctly.
 - Don't save `/build:particles/dust.png;particles/dust.png` on files that are not really links.
 - "Open containing directory" on a linked file does not open the folder the real file is in.
 - "Open with" doesn't seem to work on linked files.

This is just my first quick pass at things.  Once you get these fixed i'll do another and see if i can spot more bugs.

This is really what bothers me with the pipeline tool the most.  Unless i took the time to manually tests these features we would not have heard about these for weeks.  We really must find a better way to test these things.

tomspilman:
Another issue...  if you add a file after creating a new project it will show the "Add File" dialog.  It looks like doesn't recognize the project path until the project is saved (although it knows the path).

cra0zy:
> This is really what bothers me with the pipeline tool the most. Unless i took the time to manually tests these features we would not have heard about these for weeks. We really must find a better way to test these things.

I did have a suggestion on how to setup the tests for the Pipeline Tool, but since no one showed any interest, I did not continue it: https://github.com/MonoGame/MonoGame/pull/5068

cra0zy:
Seems like GetFullPath or GetRelativePath does not work the same between Windows and Linux, need to look into that once a part for my main PC arrives so I can fix it (which should be any day now).

tomspilman:
> Seems like GetFullPath or GetRelativePath does not work the same between Windows and Linux,

When you fix that add them to the unit tests here:

https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/UtilitiesTest.cs

cra0zy:
@tomspilman Its ready for another spin.

> Another issue... if you add a file within the project folder right after creating a new project it will show the "Add File" dialog. It looks like doesn't recognize the project path until the project is saved (although it knows the path).

This is a minor change for consistency. 

> Also the "Add File" dialog pops up in a weird location. It like all other popup dialogs should appear at the center of the Pipeline Tool window.

Hmm, this should probably be a change for a separate PR.

dellis1972:
> Tom said
> Can you update the doc on MGCB to mention how links work: 
> https://github.com/MonoGame/MonoGame/blob/develop/Documentation/mgcb.md

@cra0zy can we update the docs with some info on how to use linking from a response file in case users are hand editing the .mgcb?

cra0zy:
> @cra0zy can we update the docs with some info on how to use linking from a response file in case users are hand editing the .mgcb?

I'll wait for someone to test out the pipeline, if its all good, I'll add a separate commit for the docs so that those changes can be viewed separately.

dellis1972:
@cra0zy I'm seeing a weird problem where the tree does not show up. It happens when you open an .mgcb which has only one item in it and its in a subfolder.

screen shot 2018-03-10 at 15 54 56

cra0zy:
@dellis1972 Does the project work properly with devel Pipeline Tool? (just want to make sure it hadn't been broken due to Eto update)

tomspilman:
Found another issue... this one is pretty major:

 - Adding files that are within the content project directory still pops up the `Add File` dialog saying " it is outside the target directory.

cra0zy:
> Adding files that are within the content project directory still pops up the Add File dialog saying " it is outside the target directory.

I need to change the wording on that dialog, but that is normal.

cra0zy:
(that is, its an intentional change for consistency with file handling)

tomspilman:
> that is, its an intentional change for consistency with file handling

To me this just seems like we've made adding content more tedious.  How is this a feature to the user?

IMO if the file is within the content directory it should never ask to link or skip.  It should just add it as it did before.

cra0zy:
> IMO if the file is within the content directory it should never ask to link or skip. It should just add it as it did before.

Actually it did half the action before, it was never consistent on if the dialog would be shown.

tomspilman:
> Actually it did half the action before, it was never consistent on if the dialog would be shown.

Then lets make it consistent now.  If the content can be added to the project without linking or copying then don't show the dialog.

cra0zy:
@tomspilman The 3 changes you requested are in a new commit so you can see the changes. I'll get the docs fixed now.

Documentation/mgcb.md
```diff
@@ -110,9 +110,9 @@ Note all defined processor parameters are cleared when the `/processor` is set.
 
 ### Build Content File
 ```
-/build:<content_filepath>
+/build:<content_filepath>;*<destination_filepath>
```
tomspilman: Let's just write this twice to make the syntax clear (`*` does not suggest optional to me):
```
/build:<content_filepath>
/build:<content_filepath>;<destination_filepath>
```

cra0zy: Done.

tomspilman:
@cra0zy - There is one change we should still restore here.

When adding content to the "root" of the project, if the content is within any subfolder of the project, it should add it to the project within that subfolder and not ofter to copy or link it to the root.

This is how it worked previously which is how many tart add content.   The extra dialog now is just in the way... especially since almost no one wants to copy or link in this situation.

How it is currently it means you have to cancel, then add the folder, then add the file to that folder.

At some point later i would like to work out something like a second menu item and/or holding Shift or Ctrl when adding the content always brings up the dialog to give you options.

cra0zy:
> When adding content to the "root" of the project, if the content is within any subfolder of the project, it should add it to the project within that subfolder and not ofter to copy or link it to the root.

No it should not, it would break linking the files in the same folder as root, and it also does not make a lot of sense from a UI standpoint.

tomspilman:
> it would break linking the files in the same folder as root,

Can you explain what you mean?  It seems you are suggesting people will link a file within the content folder to another location within the content folder?  How is that something that is critical to users?

The old behavior was better and more important.

> it also does not make a lot of sense from UI standpoint.

I disagree.  How you have it right now is not how it was before and makes the experience worse.  

Some video to make sure we're both talking about the same thing:

 https://youtu.be/qDKJAgRlcxc 

 - I create a new content project in an existing folder with some content in it.
 - I try to add the content in a subfolder and get the add files dialog.
 - None of the options allow me to do what i want to do... so it cancel out.
 - I try to add the directory and again i get link/copy dialog.
 - That too doesn't do what i want (note dialog always appearing half off the window).
 - Finally i create the folder which already exists on disk manually.
 - I then add the file to the folder.

All this used to be one step before this PR.

cra0zy:
> It seems you are suggesting people will link a file within the content folder to another location within the content folder? How is that something that is critical to users?

Well we should support that. Also think about drag and drop operations.

How about this, we check out what VS and MD do and we copy that?

cra0zy:
I just remembered, VS has a separate "Add as a link" under the button.. hmm...

cra0zy:
Hmm, maybe upgrade the add dialog:
screenshot from 2018-03-11 14-38-48

tomspilman:
> Well we should support that.

I agree we could and maybe should support crazy linking situations. I just do not agree we should make the crazy case the default behavior.

> Also think about drag and drop operations.

We don't currently support drag and drop.  I think we should focus on our immediate needs first.


> I just remembered, VS has a separate "Add as a link" under the button

It does... i suggested earlier today we can add something like it in a later PR.


> Hmm, maybe upgrade the add dialog (this is what MD does):

Yeah maybe... but that dialog should not appear in my video above.  What i was trying to do should be the default behavior without any extra steps.

If we want to add a way to always shows the dialog...  like a second menu item or have the existing menu item change text/behavior if hit with Shift or Ctrl pressed... that would be fine.

But still... the most common features should always be the one with the least steps.

cra0zy:
> It does... i suggested earlier today we can add something like it in a later PR.

I would have to write separate open file dialogs for each platform to support something like that, and rework a bunch more code :\ 

On the other hand setting up the MD file dialog would require me to add around 20 more lines of code and there would be no confusion in what the action would do.

tomspilman:
> I would have to write separate open file dialogs

I do not want us to do that and never intended to suggest we do that.

Please put this discussion on hold until we decide to implement it.  This PR is difficult enough without having deep discussions on future features as well.

> showing the dialog is the least confusing thing here.

No one is confused.  I am simply asking to keep the same behavior which we have had for many years now and every existing user of the Pipeline Tool expects.  A behavior no one has complained about or asked us to change.  Preserving the existing behavior is not confusing to any user.

Please let's keep the existing behavior and not make things more difficult for existing users.

cra0zy:
*sigh*, fine, I'll do it, but I really dislike giving users less options...

Almost feel like Nautilus devs who actually removed features for the sake of "simplicity": https://www.omgubuntu.co.uk/2012/07/is-the-new-nautilus-a-step-in-the-direction-poll

tomspilman:
> I really dislike giving users less options

I can tell...  but you really should try to trust me.

If you are thinking that i am trying to remove features or i do not care about adding features you're completely mistaken.  I simply am trying to move forward quickly and logically without causing problems for our users.

First we get this PR is merged without disrupting users with a new popup in a situation where they will almost never need it.

After than we'll find a way so that users that want to see the popup can trigger it when they add files in any situation.  I don't know exactly how we'll do that yet and i was trying to avoid complicating this PR and make the merge take longer by trying to figure it out now.

Then after that we can look at drag and drop.

In none of these cases am i trying to remove features or keep from developing new features.  Quite the opposite i want to add much more, but i have a responsibility to users to not to make previously simple tasks harder for them.

I hope this makes my intentions clear.

tomspilman:
I'll give this another test in the morning and i think we'll be ready to merge it.

nkast:
I tried to use this for renaming content, much like the 'Name' property on XNA. (feature request #3009)

I tried the following which successfully generated `/bin/fold/tx.jpg`. (however the filterview did not show the file). 
`/build:fold/txHD.jpg;fold/tx.jpg`

I also tried the following which generated the file in the root '`/bin/tx.jpg`'
`/build:fold/tx64.jpg;tx.jpg`
I would expect this line to rename the asset but send it to `/bin/fold/`. If I wanted it moved to the root I'd write something like  
`/build:fold/tx64.jpg;/tx.jpg` or `/build:fold/tx64.jpg;../tx.jpg`. 
That's a minor detail, If it's making things too complicated I can live with the first option.

cra0zy:
The line is not there to rename the file, its there to relink the file.

tomspilman:
> The line is not there to rename the file, its there to relink the file

I think we should continue to develop it to allow for rename in the future.

cra0zy:
> I think we should continue to develop it to allow for rename in the future.

But than you can't relink, which is the whole point of this PR.

nkast:
So this works only by accident? there is no plan to implement  #3009 officially?

cra0zy:
> So this works only by accident? there is no plan to implement #3009 officially?

No, this does provide asset renaming, you just have to specify the whole output path, not just the name.

nkast:
Ok, I guess I 'll have to keep those files outside , let's say in '../HDAssets/fold/tx.jpg' and add them as links. This will work for me too.

If I keep the 'linked' items inside the content folder or specify a different name the result is unspecified - not supported. correct?

tomspilman:
@cra0zy - I'll give this a test tomorrow so we can finally merge this.

cra0zy:
@tomspilman Since I don't see us getting 3.7 released any time soon, maybe we should try to get this merged?

PR #6246, opened 2018-03-11
Align content on Windows (fixes #6245)

### LINKED ISSUE #6245: Fix misaligned treeview/output in Pipeline Tool
This one is very minor, but is still driving me nuts every time i open the Pipeline Tool:

image

This is on a Windows 10 machine with the latest develop release.  One of the other of these windows needs to be moved to fix this.

Also it would be even better if the borders had matching colors to them.

cra0zy:


tomspilman:
Thanks @cra0zy !

PR #6254, opened 2018-03-16
Fix memory leak for abandoned (finalized) SoundEffect instance

aienabled:
Hi guys,

I'm loading OGG files into `byte[]` and creating `SoundEffect` instances from that.
I just noticed that `SoundEffect` implementation for XAudio leaks memory if I don't call `Dispose()` explicitly. This pull requests fixes the issue for me.

`DataStream` instance doesn't release the memory blob in its finalization so it seems we have to call `Dispose()` explicitly for it in `SoundEffect.XAudio.cs`.

I simply don't know whether I still have references on the `SoundEffect` (again, it's not loaded from Content) and prefer to delegate this to GC. I wrap it in a weak reference just in case I will need the instance and it's still alive - so I can skip loading OGG file again and it's a pretty simple yet effective cache implementation.

Regards!

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.XAudio.cs
@@ -346,11 +346,9 @@ namespace Microsoft.Xna.Framework.Audio
 
         private void PlatformDispose(bool disposing)
         {
-            if (disposing)
-            {
-                if (_dataStream != null)
-                    _dataStream.Dispose();
-            }
+            if (_dataStream != null)
+                _dataStream.Dispose();
+            
             _dataStream = null;
         }
 


KonajuGames:
That would not be the correct fix.  The `if (disposing)` condition is so the `DataStream` instance is correctly disposed when `Dispose()` is called on the `SoundEffect` and so we do not explicitly dispose the `DataStream` instance if the `SoundEffect` has been abandoned and cleaned up by the garbage collector.  In the case of the `SoundEffect` being cleaned up the garbage collector, we cannot know if the `DataStream` has been already collected or not, so we must not explicitly reference it in `SoundEffect.Dispose()`.

[`DataStream.Dispose()`](https://github.com/sharpdx/SharpDX/blob/master/Source/SharpDX/DataStream.cs#L211) does appear to release the blob correctly.  It follows the same rules because `Blob` is a managed object and can only be safely explicitly disposed if `DataStream.Dispose()` was explicitly called.  If the blob instance in `DataStream` is not disposed when the `SoundEffect` is disposed, then some other managed object has a reference to the `Blob` instance from the `DataStream` instance.

KonajuGames:
In addition, abandoning `IDisposable` resources to be collected by the garbage collector is generally bad practice.  They implement the dispose pattern in order for you to dispose of them when you are finished with them.  The call to `Dispose(false)` is a backup in the situation where the object was accidentally left behind.  Calling `Dispose()` explicitly when you are finished with the instance is the safest way to deal with object instances that contain references to unmanaged resources.

aienabled:
@KonajuGames, thank you for the quick feedback!
Yes, I'm aware of the `Dispose()` pattern and that it must not dispose managed objects during the finalizer call as they should take care of unmanaged references themselves, and managed references are too dangerous to dispose this way for multiple reasons.

However, the `DataStream` class (which is located in SharpDX) doesn't have a finalizer (base class `System.IO.Stream` doesn't implement a default finalizer) - I think it was never intended to be used the way it's used in MonoGame (the streams in C# are generally expected to be used with `using` or `try-finally` patterns). It's not a problem with the `Blob` instance it hold (as it has a finalizer and disposed correctly), it's a problem with `_gCHandle` field it hold. It created as `Pinned` and so the reference on the bytes array is pinned to Memory and will be released only when the `Free` method of handle struct is called. However, it's never called if you don't call the `Dispose` method for `DataStream` explicitly as it doesn't have a finalizer. Adding finalizer to `DataStream` sounds like a good idea and I just reported it https://github.com/sharpdx/SharpDX/issues/996

So, my proposed fix is not technically correct and you're right about it. However, until (and if) SharpDX implement the proposed fix and MonoGame will be updated to include the fix, in my case (when `SoundEffect` is created from byte[] array and it's not disposed explicitly) it's the only way to resolve the memory leaks (which appears in Memory Profiler as "pinned reference" (GCHandle created with `GCHandleMode.Pinned`)) as `DataStream` can't release unmanaged references itself due to lack of finalizer.
So I will keep it as a workaround for my game and it might be a good idea to consider including this fix later based on the feedback to the reported issue in SharpDX repository.

aienabled:
> In addition, abandoning IDisposable resources to be collected by the garbage collector is generally bad practice.

Sure, but due to nature of some games (or game engines built on top of MonoGame, as in my case) it's simply not viable to explicitly control disposal of heavy assets such as textures and sounds - the game doesn't have a strict level division with known assets to be used. Consider an open world game - all the assets are loaded on demand and released automatically when the game is low on memory. In some cases assets such as item icons or sounds might be suddenly not required for a current frame, but a second later they might required again and some kind of cache is helpful. However, a cache with tracking of the actual usage of assets (via reference counting) is not viable for multiple reasons (and sometimes is even impossible, for example when using a middleware such as NoesisGUI and it doesn't have any API to monitor usage of the provided images). Relying on the weak references (to which the "strong references" are converted if an asset is not accessed from the cache for some reasonable time) is much easier to implement and much less error-prone solution (in the same way as we prefer to work with C# instead of a language without GC).

I'm aware that relying on finalizers is not a great idea generally, but as I'm using that only for cached content/assets and I've verified that finalizers of according classes are implemented in a proper way (except the SoundEffect disposal as I've reported here), it's fine and actually works great.
BTW, if you're curious about the game - it's in a free open alpha now http://cryofall.com/ you can read more about the engine here 

Regards!

KonajuGames:
Ahh yes, `Stream` does not implement the finalizer but depends on derived classes to implement it. Not sure I agree with that, but that is how `Stream` is.  But it does not change the fact that `SoundEffect` cannot call `Dispose` on the `DataStream` during its finalizer because the stream instance may have already been collected.  The rule still stands to never reference another managed object from your finalizer. The only managed object you can trust is still there is the one currently being finalized.

aienabled:
@KonajuGames, right, it seems SharpDX developer forgot that `System.IO.Stream` doesn't has a finalizer - it's really easy to miss that - as it does have `Dispose(bool disposing)` virtual method and most developers expect that it will be called automatically with `false` argument in case of finalization.

I agree - the proposed fix is not correct. This is a hotfix which is valid just for my game (and to a few rare MonoGame developers who create their sound effects from `byte[]` and don't dispose them explicitly).
I'm closing this pull request and will continue monitoring my issue report in the SharpDX repository.

aienabled:
Good news. The fix is merged into SharpDX.

PR #5838, opened 2017-07-24
[DX] MSAA RenderTarget GetData and SetData

Jjagg:
With these changes GetData works on MSAA rendertargets.
Fixes #5834.

I made it so the resolved texture is stored in the `_texture` field and the MSAA texture is stored in the `_msTexture` field. That made more sense, since code interacting with the texture usually requires the resolved texture. Because of that change I could clean up the code some more, to make it easier to understand and more maintainable. I hope that's fine.

I also added a test that calls SetData and GetData. Note that SetData is done on the resolved texture as explained in https://github.com/MonoGame/MonoGame/issues/5834#issuecomment-317408529.

Minor other stuff (mostly related):
- Fixed visibility of protected members (made them internal)
- Improved initialization of `_sampleDescription` (it was all over the place)
- Made it so `MultiSampleCount` is the actual MS count on the texture, not what's passed as the preferred MS count (we should probably also do that for GL)
- Fixed XNA test compilation

PR #6256, opened 2018-03-17
[DesktopDX] Fix MSAA for SwapChainRenderTarget

### LINKED ISSUE #6233: Null reference exception in RenderTarget2D when setting SwapChainRenderTarget with MSAA
- MonoGame 3.7.0.1376

- Windows

- WindowsDX

I am using MonoGame embedded in WinForms. To properly handle resizing of the window, I'm using a SwapChainRenderTarget. All works fine as long as I don't use MSAA (preferredMultiSampleCount = 0).

When I set preferredMultiSampleCount to a higher value (8 in my case), this works fine when I set this as a render target the first time. However, setting a new render target to the graphics device (SwapChainRenderTarget or a normal RenderTarget2D), a null reference occurs in the RenderTarget2D class, method ResolveSubResource(). This is because the _resolvedTexture field of a SwapChainRenderTarget is always null. The _renderTargetViews of a SwapChainRenderTarget is constructed in a different way compared to the RenderTarget2D class. This means the CreateTexture override of the RenderTarget2D is never called resulting in the _resolvedTexture always remaining null.

In the PlatformResolveRenderTargets method of GraphicsDevice the ResolveSubResource method is called for all render targets that are of type RenderTarget2D which will fail for SwapChainRenderTargets.
I'm not sure how this should be fixed; should a SwapChainRenderTarget have a _resolvedTexture when MSAA is used, or should the ResolveSubResource method never be called in these cases?

harmwind:
MSAA was already implemented for RenderTarget2D, but gives problems for the derived SwapChainRenderTarget since the resolved texture is not created. This PR fixes that.

fixes #6233


tomspilman:
@Jjagg @KonajuGames do we have a unit test for masa resolves or SwapChainRenderTarget?

Jjagg:
Is SwapChainRT used in DX projects? We have a few backbuffer msaa tests in [GraphicsDeviceManagerTest](https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/Graphics/GraphicsDeviceManagerTest.cs).

MonoGame.Framework/Graphics/RenderTarget2D.DirectX.cs
```diff
@@ -155,8 +155,14 @@ internal override Resource CreateTexture()
...
             // MSAA RT needs another non-MSAA texture where it is resolved
             if (SampleDescription.Count > 1)
-            {
-                _resolvedTexture = new RenderTarget2D(
+                _resolvedTexture = CreateResolvedTexture();
+
+            return rt;
+        }
+
+        protected RenderTarget2D CreateResolvedTexture()
```
nkast: This better be internal, because not it's accessible to the developer through a derived class.  
(unfortunately C#  doesn't support a `protected AND internal` modifier)

Additionally this could set directly the `_resolvedTexture` with a check/assert to avoid secondary calls.
_resolvedTexture can be reverted back to `private'.

nkast:
I tested this PR and it throws a SharpDX exception. I found that i had to do with multisampleDesc.Quality.
The only values that works is if I set `Quality = 0`. (AMD R9 270)
https://github.com/MonoGame/MonoGame/pull/6256/files#diff-0ad544a19b53a35ad9ffb664ddfb3e03R73
`multisampleDesc.Quality = (int)StandardMultisampleQualityLevels.StandardMultisamplePattern;`

for comparison, GraphicsDevice uses _d3dDevice.CheckMultisampleQualityLevels()
https://github.com/MonoGame/MonoGame/blob/a02b4f874b7a1f842122e9bb91375750ea06ba9e/MonoGame.Framework/Graphics/GraphicsDevice.DirectX.cs#L867-L873

tomspilman:
> I tested this PR and it throws a SharpDX exception. 

@nkast is that gpu specific or something missed in our unit tests?

nkast:
This also fails when I set an unsupported multisampleCount. For example a value of 16 fails in my GPU.
Lower values works (1,2,4,8).
The method should fallback to the highest supported values instead of throwing exceptions.
That's an issue with RenderTargets in general. The GraphicsDevice is using the` GetClampedMultisampleCount()` method to fallback to a supported multisample level.

nkast:
@tomspilman I've only tested this on my desktop but I know those things differ from card to card.

MonoGame.Framework/Graphics/RenderTarget2D.DirectX.cs
harmwind: Instead of internal, you could use `private protected` (supported since C# 7.2) which translates exactly to `protected AND internal`

harmwind:
I have tested the code (with and without MSAA and mixing RenderTarget2D's and SwapChainRenderTargets) on a NVidia Quadro card and an Intel HD620 without problems. I don't have access to AMD cards.

nkast:
@harmwind 
Can you test something if it works with your GPU? 

I replaced those two lines: 
```
            var multisampleDesc = new SampleDescription(1, 0);
            if (preferredMultiSampleCount > 1)
            {
                multisampleDesc.Count = preferredMultiSampleCount;
                multisampleDesc.Quality = (int)StandardMultisampleQualityLevels.StandardMultisamplePattern;
            }
```   

with

```
    var multisampleDesc = graphicsDevice.GetSupportedSampleDescription(dxgiFormat, preferredMultiSampleCount);
```

That works on my AMD card without problems. It is same way we create the SampleDescription for backbuffer and RenderTargets.

harmwind:
@nkast 
I have tested your change succesfully. I see this is indeed similar as how the RenderTargets are created. I added the change to this PR. Note however that this issue already existed before this PR, the MultiSampleDescription was always created like this for SwapChainRenderTargets.

nkast:
Great! thanks @harmwind 

I know that the issue was there from before. Nobody had the change to test MSAA for SwapChainRenderTarget until your PR.

nkast:
I noticed a strange behavior after enabling multi-sampling. The window seems to paint only a multiple of width/height. The control is hosted in WPF. This might as well be a bug from my part or the WPF+WinFormsHost.

swapchainrendertargetmsaa

Any ideas?
 
Using the newer SwapChain1.Present1() method seems to fix it, although I still get this artifact at the bottom of the control when resizing the Height.

SwapChain1 fix : https://github.com/nkast/MonoGame/commit/5cfd69f99ee0e1e4d8927620f22482649120ffa4

nkast:
@harmwind 
I found that the cause of this strange artifact was the resolve method. My approach was to not resolve _texture. _texture is used only when you bind the renderTarget as a source, in this case we can ignore it.
https://github.com/nkast/MonoGame/commit/eb49d13aee0aae7ece96b2cf07759820aa955e11

Surprisingly this had a significant improvement in performance, since we skip the unnecessary copy from  backbuffer to _texture.

nkast:
Hey, @harmwind  That was fast! 
Did you tested it in your app? In my app the difference in performance was quite noticeable.

One negative result of this change is that the SwapChainRenderTarget can't be used as a source. But that was never the purpose of SwapChainRenderTarget anyways.

harmwind:
Hi Nikos,

Just added your changes and did a rebuild. Monogame is a nuget package for our own software, so didn’t get to testing it yet. Will update now and let you know soon.
I accidentally already did a push, was a bit too early with that, but can roll back if it doesn’t work out.


Best regards,
Harm Wind

harmwind:
Hi Nikos,

Just tried it in our app (Windows + DX) and everything works fine. My develop PC seems to be ‘too good’ to test any performance improvements. Even when drawing 50 million triangles with MSAA at 8, it is still capped at 60FPS. However, our users have lower end PC’s so I hope to see performance improvements there, I can test this next week and will let you know.

For now, I will leave this change in. We don’t use the SwapChainRenderTarget as source, so this change should be useful anyway.

BR,
Harm

harmwind:
@nkast 

Today I tried your change on some lower end PC's (laptop with i5 and HD620 and tablet with i5 and HD520). Especially the HD520 is a real low end graphics card, but unfortunately I don't see any performance improvements.
Are there any prerequisites to see the improvement?

nkast:
In my system it was related on the window size and the preferredMultiSampleCount.
especially at large values, 8 or 16.

nkast:
If we don't have a resolve texture I think that might be a serious limitation.
GetData() will not work, (in a n editor you might want to grab the image and stream it to a 2nd device for example).

The real problem here is that textures are resolved as soon as you replace them with a SetRenderTarget().
The CPU has to wait for the GPU to finish all drawing on the backbuffer.
That's a problem on many senarios with large targets (post procesing, Deffered rendering, editor with multiple SwapChainRenderTargets) or many smaller ones (shadow maps). 

@Jjagg, I know you are looking into a fix for GetData() on MSAA RenderTargets.
Is it possible to mark a rendertarget as invalid [here](https://github.com/MonoGame/MonoGame/blob/3e92fc7bff7b4c32edf1314e143a4b7de4fab161/MonoGame.Framework/Graphics/GraphicsDevice.cs#L810) inside ApplyRenderTargets(), and call .ResolveSubresource() only by the time we need to access to the resolved texture? (GetData(...), SetRenderTarget(...), etc.)

MonoGame.Framework/Graphics/RenderTarget2D.DirectX.cs
Halofreak1990: What weird dialect of C# doesn't support protected internal? It's been supported for a long time now. The only exception is struct members, because a struct cannot be inherited from.

harmwind: nkast specifically stated that the MonoGame developers like to keep most methods internal and don't let classes outside the assembly derive from them. That's why he mentioned the method should be changed from protected to internal. **protected internal** in C# means protected OR internal. That's the reason for my reply that for methods like these, private protected could be used, but I was not sure if these new language features were already used at the time.
Doesn't matter anymore, it was changed to internal (in line with many of the methods in MonoGame).

Halofreak1990: @harmwind Ah, I see what you're getting at now. Seems I misread @nkast's reply

Jjagg:
I overhauled the way MSAA works on textures in  #5838. Really sucks that didn't go through due to issues with Piranha :/

I don't really remember what exactly that does and can't easily check rn because I'm mobile, but I don't think that changed behavior around resolving the texture. Your solution sound like a good one @nkast :)

SimonDarksideJ:
@harmwind are you able to resolve the merge conflicts and retest on the latest to see if we can get this change merged?
Should be easier now Piranah is gone.

harmwind:
Since 2018 we used a fork in which we applied several fixes.
Recently we migrated to .NET 8 (from .NET Framework) and started using Monogame 3.8.1.305 and all MSAA related issues seems to be resolved already. I believe several fixes were done by Jjagg over the years and I think my pull request is obsolete by now.

PR #6257, opened 2018-03-18
Updated Windows Universal Min SDK Versions.

dellis1972:
This should in theory allow us to update the Nuget
version on the build bots.
Also adds a missing UWP template for the Nuget Template system


dellis1972:
@tomspilman once this one is merged we can probably update Nuget to 4.5.1, in that case we can problem then build the Nuget Templates for Package.

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/BlankApp/BlankApp.UWP/BlankApp.UWP.csproj
```diff
@@ -12,8 +12,8 @@
     <AssemblyName>BlankApp.UWP</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.15063.0</TargetPlatformVersion>
-    <TargetPlatformMinVersion>10.0.10586.0</TargetPlatformMinVersion>
+    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.10586.0</TargetPlatformVersion>
```
SimonDarksideJ: This version should probably target 16299, or at a minimum 14393 as the TargetPlatformVersion (as 10596 isn't supported any more)

ProjectTemplates/VisualStudio2015/WindowsUniversal10.VB/WindowsUniversal.VB.vbproj
```diff
@@ -11,7 +11,7 @@
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
     <TargetPlatformVersion>10.0.10586.0</TargetPlatformVersion>
```
SimonDarksideJ: Same as above, either target 16299 or 14393 as in the other project files below.

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/BlankApp/BlankApp.UWP/BlankApp.UWP.csproj
tomspilman: @SimonDarksideJ - i suppose the downside to moving to 16299 is users are forced to update to the newest SDK?  It seems like to use it you have to be on VS2017 which might be annoying to existing users.

14393 seems better in that regard and users can always update their project to move forward.

I think @SimonDarksideJ knows more about this then me... so i would go with his recommendation here.

SimonDarksideJ: Not really, this simply states the requested version is MS's currently supported OS version.
But the minimum is the lowest level current API that is supported.
We do need to keep a track of SDK's that MS actively supports but since MG doesn't actually use many native API's we should be covered internally, the consumer can set their versions to whatever they want afterwards.

dellis1972:
@MonoGame/maintainers this should be good to go :)

tomspilman:
Thanks @dellis1972 !

PR #6258, opened 2018-03-18
Make sure that the window titlebar is within screen bounds on DesktopGL

cra0zy:
SDL 2.0.5 has SDL_GetWindowBordersSize function, however on some platforms it did nothing, once we update SDL this will work properly on all the platforms.

Yes, this is safe to merge.

### DIFF
--- a/MonoGame.Framework/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/SDL/SDLGameWindow.cs
@@ -236,8 +236,11 @@ namespace Microsoft.Xna.Framework
                 _height = displayRect.Height;
             }
 
-            var centerX = Math.Max(prevBounds.X + ((prevBounds.Width - clientWidth) / 2), 0);
-            var centerY = Math.Max(prevBounds.Y + ((prevBounds.Height - clientHeight) / 2), 0);
+            int ignore, minx = 0, miny = 0;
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/SDL/SDLGameWindow.cs
```diff
@@ -236,8 +236,11 @@ public override void EndScreenDeviceChange(string screenDeviceName, int clientWi
                 _height = displayRect.Height;
             }
 
-            var centerX = Math.Max(prevBounds.X + ((prevBounds.Width - clientWidth) / 2), 0);
-            var centerY = Math.Max(prevBounds.Y + ((prevBounds.Height - clientHeight) / 2), 0);
+            int ignore, minx = 0, miny = 0;
+            Sdl.Window.GetBorderSize(_handle, out miny, out minx, out ignore, out ignore);
+
+            var centerX = Math.Max(prevBounds.X + ((prevBounds.Width - clientWidth) / 2), minx);
+            var centerY = Math.Max(prevBounds.Y + ((prevBounds.Height - clientHeight) / 2), miny);
```
tomspilman: So if i read this right... you're just adjusting the center so that the title is visible and the bottom of the window is offscreen.  The window size isn't changing at all.  Correct?

cra0zy: I just changed the minimum possible value from `0` to top and left border sizes.

PR #6261, opened 2018-03-19
Create a Vsix for Visual Studio 2017/5 (DO NOT MERGE)

dellis1972:
This will allow users to discover MonoGame via the Visual Studio Extensions and Updates system.

ToDo
- [ ] Install/Use the new Templates
- [ ] Try to figure out a way to install into Program Files (x86) rather than in the Edition MSBuild Dir. (Not sure this is possible)


cra0zy:
> Try to figure out a way to install into Program Files (x86) rather than in the Edition MSBuild Dir. (Not sure this is possible)

If we use nugets instead of local references, we could just preload them into the nuget cache, which is easier.

dellis1972:
>> Try to figure out a way to install into Program Files (x86) rather than in the Edition MSBuild Dir. (Not sure this is possible)

>If we use nugets instead of local references, we could just preload them into the nuget cache, which is easier (or also add custom <RestoreSources>).

@cra0zy I still need to deploy the assemblies for backward compatibility even if the templates start using Nuget. The work around so far is to have the vsix put them in the same place as the tooling and have a custom .props file which sets `$(MonoGameInstallDirectory)`. That seems to work so far.

cra0zy:
> backward compatibility

There are new native deps for DesktopGL and Mac platforms.. we are breaking it again either way 
¯\_(ツ)_/¯.

Issue #6260, opened 2018-03-19
Float textures for OpenGL ES 3.0


Jjagg:
MonoGame currently does not support `SurfaceFormat.Single` for GL ES 3.0. I opened this issue in response to a topic on the forums: http://community.monogame.net/t/android-shadowmap-and-render-target-sufaceformat-lack-of-float-support/10459

In OpenGL ES 2.0 float textures are supported through the [`OES_texture_float`](https://www.khronos.org/registry/OpenGL/extensions/OES/OES_texture_float.txt) extension, but only in the alpha channel. Since `SurfaceFormat.Single` is defined as a float in the red channel we can't support it in GL ES 2.0.

In OpenGL ES 3.0 things work just like in OpenGL, so things should just work if MG did exactly the same thing as in DesktopGL. Currently an `UnsupportedOperationException` is thrown. The user reporting this issue in the forum topic mentioned the following line causes the exception:
```
new RenderTarget2D(GraphicsDevice, 2048, 2048, false, SurfaceFormat.Single, DepthFormat.Depth24);
```

We should investigate where this exception is thrown so we can prevent throwing it under OpenGL ES 3.0. The docs should mention that `SurfaceFormat.Single` is not supported for OpenGL ES 2.0.

This does raise the question if we should actually support OpenGL ES 3.0+ only features on mobile, since it's not portable. In this case it's zero-effort though, so I'm all for it :)

tomspilman:
> but only in the alpha channel. Since SurfaceFormat.Single is defined as a float in the red channel we can't support it in GL ES 2.0.

Are we sure if you sample the texture and do `color.r` it doesn't return the same value as `color.a`?

Still supporting it in just GL 3.x would be fine as well.  I think `Single` is only actually supported in HiDef in XNA... so it sort of makes sense.

> This does raise the question if we should actually support OpenGL ES 3.0+ only features on mobile, since it's not portable.

If ES 3.0+ is more like XNA HiDef then we should support it and set the correct graphics profile to alert the user.

Jjagg:
> In OpenGL ES 2.0 float textures are supported through the OES_texture_float extension, but only in the alpha channel.

[...]

PR #6264, opened 2018-03-21
Add Support for Floating Point Textures on Mobile Platforms.

dellis1972:
We currently restrict Android and iOS to NEVER support
`SurfaceFormat` values of `SurfaceFormat.Single` etc.
Ever if the device supports it either directly or via
and OpenGL extension.

This commit reworks the GraphicsCapabilities and GraphicsExtensions
to use the queried extensions to figure out if we support the
`SurfaceFormat` requested. If it doesn't then we throw a
`NotSupportedException` as we current do. If we do however then
we will run as normal.

Also fixed up a few hardcoded `PixelFormat` values so they have
proper enumerations.


dellis1972:
@KonajuGames @Jjagg  had some problems with detecting `SupportsHalfFloatTextures`. Not 100% sure what extensions are needed. On desktop GL should we say `SupportsHalfFloatTextures` and `SupportsFloatTextures` are always `true`?

KonajuGames:
> On desktop GL should we say SupportsHalfFloatTextures and SupportsFloatTextures are always true?

I think it's safe to say they're supported on pretty much every card we support on desktop now.

MonoGame.Framework/Graphics/GraphicsExtensions.cs
```diff
@@ -597,136 +620,118 @@ internal static int GetSwapInterval(this PresentInterval interval)
...
             case SurfaceFormat.NormalizedByte2:
+                if (!supportsFloat)
+                    goto case InvalidFormat;
                 glInternalFormat = PixelInternalFormat.Rg8i;
                 glFormat = PixelFormat.Rg;
                 glType = PixelType.Byte;
                 break;
 
             case SurfaceFormat.NormalizedByte4:
+                if (!supportsFloat)
```
KonajuGames: Is this correct for these formats?  Are they dependent on float support?

dellis1972: ah those might be an over sight on my part.

dellis1972:
@Jjagg I need to dig out my old GLES 2.0 devices and see what they support. Or boot up the OUYA since that had a pretty good GLES 2.0 chip in it. But this works on my Pixel which is GLES 3.x and seems too work on my iPad too which was a nice surprise :D

Jjagg:
@dellis1972 Ah, the good old OUYA :p My previous comment is about OpenGL 2.0 though, not OpenGL ES. 

For GL ES 2.0 things are a bit more involved because it does not support R32F (even with extensions), so maybe that's best left for another PR. We can implement it albeit slightly different from how it's implemented for regular OpenGL and GLES 3.0+. I commented on this in #6260.  Basically we can use Luminance instead of red only, which means in GLSL sampling will give (F, F, F, 1) instead of (F, 0, 0, 1). Another slight difference is the pixel type value that needs to be passed when creating a texture is different from the one in regular OpenGL and OpenGL ES 3.0+: `HALF_FLOAT_OES = 0x8D61` vs [`HALF_FLOAT = 0x140B`](https://github.com/MonoGame/MonoGame/blob/b914f314a22dae8530ecfd5232adc8ceec55782f/MonoGame.Framework/Graphics/OpenGL.cs#L388).
Here's the relevant GLES 2.0 extension: https://www.khronos.org/registry/OpenGL/extensions/OES/OES_texture_float.txt

dellis1972:
So if we are glMajorVersion < 3 we should check for GL_ARB_texture_float and GL_ARB_half_float_pixel when running full GL right ? 
I would love to get rid of all the `#if GLES` stuff in this GraphicsExtensions.cs in favour of runtime checks.

Jjagg:
> So if we are glMajorVersion < 3 we should check for GL_ARB_texture_float and GL_ARB_half_float_pixel when running full GL right ?

Yes, that seems right to me :) 

>  I would love to get rid of all the #if GLES stuff in this GraphicsExtensions.cs in favour of runtime checks.

Agreed, that'd be nice.

YTN0:
@dellis1972 Any ETA on when this PR will be merged? Are you still waiting on sign-off from @Jjagg ?

dellis1972:
@Jjagg added those checks and fixed up instancing too :P

MonoGame.Framework/Graphics/GraphicsCapabilities.cs
```diff
@@ -69,13 +69,35 @@ internal void Initialize(GraphicsDevice device)
...
+        /// If the requested format is not supported an <code>NotSupportedException</code>
+        /// will be thrown.
+        /// </summary>
+        internal bool SupportsFloatTextures { get; private set; }
+
+        /// <summary>
+        /// True, if the underlying platform supports half floating point textures. 
+        /// For Direct3D platforms this is always <code>true</code>.
+        /// For OpenGL Desktop platforms it is always <code>true</code>.
+        /// For OpenGL Mobile platforms is requires `GL_EXT_color_buffer_half_float`.
```
Jjagg: typo 'is' -> 'it'

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
```diff
@@ -1173,11 +1171,8 @@ private void PlatformDrawPrimitives(PrimitiveType primitiveType, int vertexStart
 
         private void PlatformDrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount)
         {
-#if GLES || MONOMAC
-            throw new PlatformNotSupportedException("Instanced geometry drawing is not supported yet for GLES or MONOMAC.");
-#else
```
Jjagg: @dellis1972 You forgot to remove the matching #endif

dellis1972:
Doh

Sent from my Windows 10 phone

dellis1972:
looks like I fixed that

MonoGame.Framework/Graphics/GraphicsDevice.OpenGL.cs
dellis1972: @Jjagg all fixed up :)

MonoGame.Framework/Graphics/GraphicsExtensions.cs
```diff
@@ -597,131 +621,113 @@ internal static int GetSwapInterval(this PresentInterval interval)
...
+                if (!supportsHalfFloat)
+                    goto case InvalidFormat;
                 glInternalFormat = PixelInternalFormat.Rgba16f;
                 glFormat = PixelFormat.Rgba;
                 glType = PixelType.HalfFloat;
                 break;
 
             case SurfaceFormat.HalfSingle:
+                if (!supportsHalfFloat)
+                    goto case InvalidFormat;
```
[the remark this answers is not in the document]
dellis1972: @Jjagg not sure what is the best way to support this since it depends on the runtime  GL version. GraphicsExtensions is for the most part does not care about gl version. 

My first thought is to export a HalfSingleInternalPixelFormat property which we set appropriately on the GraphicsCapabilities. But that doesn't feel right.

MonoGame.Framework/Graphics/GraphicsCapabilities.OpenGL.cs
```diff
@@ -92,8 +92,14 @@ private void PlatformInitialize(GraphicsDevice device)
             // sRGB
 #if GLES
             SupportsSRgb = GL.Extensions.Contains("GL_EXT_sRGB");
+            SupportsFloatTextures = GL.BoundApi == GL.RenderApi.ES && (device.glMajorVersion >= 3 || GL.Extensions.Contains("GL_EXT_color_buffer_float"));
+            SupportsHalfFloatTextures = GL.BoundApi == GL.RenderApi.ES && (device.glMajorVersion >= 3 || GL.Extensions.Contains("GL_EXT_color_buffer_half_float"));
+            SupportsNormalized = GL.BoundApi == GL.RenderApi.ES && (device.glMajorVersion >= 3 || GL.Extensions.Contains("GL_EXT_texture_norm16"));
```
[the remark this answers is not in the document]
dellis1972: I will fix this, it will make sure we only ever export Normalized bits when that condition is true.

MonoGame.Framework/Graphics/GraphicsExtensions.cs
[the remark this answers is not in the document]
Jjagg: I think it's fine to check `GL.BoundAPI` and `GraphicsDevice.glMajorVersion`. I agree stuff like this shouldn't be in `GraphicsCapabilities`.

YTN0:
@Jjagg  Any ETA on when this will be merged into develop? I am not sure how the process normally works once a pull request has been approved by reviewers.

Thanks.

MattiKloepper:
I tried to use SurfaceFormat.Single an my Samsung Galaxy Tab today and got a NotSupportedException.


Monogame Build: V3.7.0.1614

These values are from an App "OpenGL ES Extensions":
Android 4.2.2
OpenGL ES 2.0 build 1.9@2166536
Renderer: PowerVR SGX 540
Extensions (among others): GL_OES_depth24, GL_OES_texture_float, GL_OES_texture_half_float

MattiKloepper:
I have mixed up Rendertargets and textures. So please ignore my above comment.

PR #6266, opened 2018-03-22
[WinDX] FIX: Soft-fullscreen resolution may be incorrect

MichaelDePiazzi:
For the Windows DirectX build, the resolution used for soft-fullscreen may be incorrect if the window position was programmatically changed (i.e. via `GameWindow.Position`).

This is because the adapter was not refreshed after changing the position. So if the window moves to a different display with a different resolution, then soft-fullscreen will end up using the resolution of the previous display.

This PR fixes this by refreshing the adapter after `GameWindow.Position` is set.

(cc @Jjagg because I know you are familiar with this particular area of code)


Jjagg:
I missed that, thanks @MichaelDePiazzi!

MichaelDePiazzi:
Cheers @Jjagg :)

PR #6267, opened 2018-03-22
AssemblyHelper: Protect GetCustomAttribute in try/catch

flibitijibibo:
This simply adds a try/catch to the GetCustomAttribute call used to get the Assembly's Title attribute.

This fix is helpful for runtime environments where this function isn't available or implemented for this particular type of attribute, other than that this should not affect any other environment.


cra0zy:
Looks good to me, thanks @flibitijibibo, merging!!!

PR #6269, opened 2018-03-23
SharpDX COM object memory leak fix

aienabled:
BitmapDecoder.GetFrame() returns a BitmapFrameDecode object which is a COM object and must be disposed.

### DIFF
--- a/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
+++ b/MonoGame.Framework/Graphics/Texture2D.DirectX.cs
@@ -362,12 +362,16 @@ namespace Microsoft.Xna.Framework.Graphics
 
             var fconv = new FormatConverter(imgfactory);
 
-            fconv.Initialize(
-                decoder.GetFrame(0),
-                PixelFormat.Format32bppRGBA,
-                BitmapDitherType.None, null,
-                0.0, BitmapPaletteType.Custom);
-
+            using (var frame = decoder.GetFrame(0))
+            {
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @aienabled !

MichaelDePiazzi:
I know I'm a bit late because this has already been merged. But I actually noticed the other day that this code was recently made redundant by PR #6008. So the entire `Texture2D.DirectX.LoadBitmap` method can actually just be deleted as it's no longer used.

aienabled:
@MichaelDePiazzi,
you're right, I totally missed this as I've made this method public in my MonoGame fork and using it to load bitmaps to create native Texture2D from them. So it seems to be a good idea for me to copy this method into my game.

I'm not using StbSharp as it's about 2-3 times slower than loading with SharpDX.

MichaelDePiazzi:
> I'm not using StbSharp as it's about 2-4 times slower than loading with SharpDX (even if I reuse the byte buffers).

I didn't realise this... I use `Texture2D.FromStream` a bit, so I better check if this will be a problem for me (I only just recently merged this switch to StbSharp into my fork). So also thanks! :)

MichaelDePiazzi:
@aienabled Wow! Just looked at the performance impact of this in my game, and yeah, StbSharp is MUCH slower and does a LOT of allocations. I've just switched back to using WIC for loading textures in my game. Thanks again for the heads up!

Issue #5656, opened 2017-04-14
Freeze while clicking/moving the title bar


Scellow:
Hello

Currently each time you click or hold the mouse button or drag the title bar the window will freeze

I don't think that should happen, any idea ?

Jjagg:
That's expected behaviour. The same happens in XNA.

Scellow:
It should be treated as a bug imo, this is really annoying specially in networked games

Jjagg:
I agree it would be nice to not block the game. I'll play around a bit to try and fix it. Currently doing some stuff for DX, so that'll be first. @tomspilman Do you agree we should treat this as a bug?

tomspilman:
This sounds similar to the window resizing discussion we had before.  In XNA the window updates were disabled during resize or window dragging.  I think we agreed if we wanted to have the window continue to update during resize (or dragging in this case) that we want to make it something the developer opts-in for.

Jjagg:
This is different because the game logic is completely blocked in this case, so update isn't even called. The issue is that [`Application.DoEvents`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Windows/WinFormsGameWindow.cs#L340) blocks while the user is resizing or dragging the window. We could use a timer to make sure `Game.Tick` is still called during these operations. I would prefer this to happen by default, but if it's opt-in that's fine too.
It would be good to at least have some way to keep the game going, because like @Scellow said this can be an issue for some types of games.

tomspilman:
> This is different because the game logic is completely blocked in this case

You sure Update isn't also blocked in XNA on window dragging?

Jjagg:
I phrased that wrong. Yeah it is, but this is a different thing. This is about the run loop blocking when dragging/resizing while the other discussion was about resizing the backbuffer during window resizing. There's an argument for not resizing the backbuffer every frame, but is there for blocking the run loop during resize/draw? There might be conflicts if you try to change window size in your game logic during the dragging operation, but that can be worked around. Other than that I don't see any downsides.

cra0zy:
On Mac/Linux the window contents does not get frozen while you drag the window.

Jjagg:
That might make it hard to get right on Windows under SDL :/
Or maybe not, since it's Win32 in the backend we could maybe just do something like
```
if (event is resize start && platform is windows)
{
    Start Win32 timer like in DX projects
}
if (event is resize end && platform is windows)
{
    Kill Win32 timer
}
[...]

PR #6273, opened 2018-03-29
Reduce WinForms garbage.

nkast:
reducewinformsgarbage

Those are ~400 WM_MOUSE messages. `Application.DoEvents()` allocates about 5 times more memory than `Application.Run()` on each Message.
*(25).vsp* = Application.DoEvents()
*(23).vsp* = Application.Run()

### DIFF
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -410,27 +410,9 @@ namespace MonoGame.Framework
 
         internal void RunLoop()
         {
-            // https://bugzilla.novell.com/show_bug.cgi?id=487896
-            // Since there's existing bug from implementation with mono WinForms since 09'
-            // Application.Idle is not working as intended
-            // So we're just going to emulate Application.Run just like Microsoft implementation
-            Form.Show();
-
-            var nativeMsg = new NativeMessage();
-            while (Form != null && Form.IsDisposed == false)
-            {
-                if (PeekMessage(out nativeMsg, IntPtr.Zero, 0, 0, 0))
-                {
-                    Application.DoEvents();
-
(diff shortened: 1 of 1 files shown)


nkast:
Interesting! Test Mac failed. 
does Mac platform use `WinFormsGameWindow.cs`?

Jjagg:
The NUnit test runner crashed. cc @tomspilman

tomspilman:
@nkast - Is your branch updated to the latest develop changes?

tomspilman:
Yeah looks like you have not updated since Feb.  There were fixes and changes for unit tests to work correctly.

nkast:
>Yeah looks like you have not updated since Feb. 

very true. 
I am on it.

tomspilman:
@dellis1972 - Do you think it matters if we support Mono on Windows?

@Jjagg - Also this might in some way help with https://github.com/MonoGame/MonoGame/issues/5656 ?

cra0zy:
> @dellis1972 - Do you think it matters if we support Mono on Windows?

I don't think we should, especially since .NET Core is a thing.

Jjagg:
> Also this might in some way help with #5656 ?

I looked at the code and also thought it might fix that issue. @nkast does this branch call the update method while dragging the window?

nkast:
Unfortunately no, the #5656 is default behavior of Windows.

MichaelDePiazzi:
FYI - I've been testing this change with my game for a week or so now, and all seems to be working as expected. And I've just released an experimental build to my users to test it further. I don't expect any problems though, it seems pretty solid.

MichaelDePiazzi:
FYI - After about a month of testing this in an experimental branch of my game, I have not come across any problems with this. I've also recently taken my game update out of of experimental and released it to everyone else.

Jjagg:
That's good enough for me! :) 
Thanks @nkast!

PR #6283, opened 2018-04-15
Fix unclosed comment and typos in issue template

Jjagg:


cra0zy:
@Jjagg Your clock is mispositioned, I guess you have both a Windows and a Linux system on the device, to fix it simply set Linux time tracking to use utc: `timedatectl set-local-rtc 1`

Jjagg:
@cra0zy Yeah, I dual boot and DST causes issues every once in a while. Thanks for the tip.

PR #6295, opened 2018-04-21
MGCB Nuget

cra0zy:
Notes:
- Made the installers use the old targets file that does not use custom C# tasks
  - I took the old targets file from 3.6 tagged branch
  - This is to ensure compatibility for older projects, this targets should not be modified anymore, but instead at some point just removed
  - Also removed the hardcoded `MonoGamePlatform` error, no point in trying to check if the platform is valid if the script itself doesn't know what platforms are valid
- Added new MGCB Nugets:
  - MonoGame.Content.Builder.Platform contains only the per platform MGCB executable and should not be referenced manually
  - MonoGame.Content.Builder contains targets and tasks that will automatically download the per platform MGCB if needed
- MonoGame.Content.Builder.csproj is compiled with `msbuild /t:Pack`, this not only build the project, but also packages it inside a .nuget file
  - another important note is that this project is targeting both `netstandard2.0` and `net46`, this is to ensure the compatibility with both(or all 3 that is): `dotnet build`, `msbuild`, `xbuild`
- Modified public variable names to make them consistent and short
- Added docs for the MGCB nuget usage
- Disabled the targets test temporarily (it needs upgrading once the project file builds)
- Nuget packages are next to the project .csproj instead of the NugetPackages folder, as this is what .NET tools are moving towards (packages generated from .csprojs) 
- Build bot needs upgrading for this to be connected to it


MonoGame.Framework.Content.Pipeline/MonoGame.Content.Builder.targets
```diff
@@ -50,12 +56,11 @@
 			'$(MonoGamePlatform)' != 'MacOSX' And       
 			'$(MonoGamePlatform)' != 'WindowsStoreApp' And       
 			'$(MonoGamePlatform)' != 'NativeClient' And       
+			'$(MonoGamePlatform)' != 'Ouya' And       
 			'$(MonoGamePlatform)' != 'PlayStationMobile' And       
 			'$(MonoGamePlatform)' != 'WindowsPhone8' And       
 			'$(MonoGamePlatform)' != 'RaspberryPi' And       
 			'$(MonoGamePlatform)' != 'PlayStation4' And       
-			'$(MonoGamePlatform)' != 'PSVita' And       
-			'$(MonoGamePlatform)' != 'XboxOne' And
```
tomspilman: Don't remove these... MonoGame supports them.

cra0zy:
> Reverting back to 3.6 targets meet we remove functionality introduced in [1] which is what required the msbuild tasks assembly to get it working on Mac.

Yea, the point is to not have any new functionality if you are using the "GAC" version.

dellis1972:
> Yea, the point is to not have any new functionality if you are using the "GAC" version.

What is the reasoning behind this? I don't remember seeing a discussion about removing local targets and forcing people to use nuget? Did I miss something? I remember a discussion about if we should remove the local assemblies, but though it was agreed to keep those in for backward compatability.

If we plan to deprecate the msbuild tasks/targets being installed locally we should do so in a way that does not break existing users. If we pull the plug or revert back that will result in people hitting issues and raising issues/bugs etc. Bear in mind we generally tell people to use `develop` if something is not working in the `stable`, so people are probably already using the new functionality which #4997 (and the subsequent fixes) introduced.

I would add a warning to the existing targets that they will be removed in a future release. The warning  should point to docs telling users how to upgrade to the new system, since it will no doubt require csproj changes.

cra0zy:
> I remember a discussion about if we should remove the local assemblies, but though it was agreed to keep those in for backward compatability.

Yea... If we forced people to use NuGet assemblies, but local MGCB it would cause a lot of chaos. Telling people to use both NuGet assemblies and MGCB is perfect for backwards compatibility and .csproj files will not have to be dirtied with "custom" mgcb targets path or `MonoGamePlatform` stuff since all of that will be provided by the nugets.

> If we plan to deprecate the msbuild tasks/targets being installed locally we should do so in a way that does not break existing users. If we pull the plug or revert back that will result in people hitting issues and raising issues/bugs etc. Bear in mind we generally tell people to use develop if something is not working in the stable, so people are probably already using the new functionality which #4997 (and the subsequent fixes) introduced.

We are not breaking the stable branch, which is the important part.

Breaking develop branch is something to be expected, the whole point of develop branches is for people to test out WIP stuff that is not finalized.

> I would add a warning to the existing targets that they will be removed in a future release. The warning should point to docs telling users how to upgrade to the new system, since it will no doubt require csproj changes.

Sure, was gonna write one up on how to switch to NuGets, since as I mentioned, we have dirtied up .csproj files with custom props and stuff.

jnoyola:
I believe there's a more robust way to accomplish the task of GetMGCBTask. I would be wary of checking anything in that has nuget.org hardcoded.

I do it with conditional PackageReferences. I think this approach has a few advantages:

1. It's a more standard reference instead of manually downloading the nuget
2. The OS is controlled by build variables instead of checking for clues like the Applications folder

I have [PackagePicker.props](https://hollow-games.visualstudio.com/_git/MonoGame?path=%2FNuGetPackages%2Fbuild%2FPackagePicker.props&version=GBdevelop) that relies on the structure of the nuget package (`path/to/<Package>/<Version>/build/<Package>.props`) to extract the package name and version to choose the corresponding package on the current OS. My [Content.Builder](https://hollow-games.visualstudio.com/_git/MonoGame?path=%2FNuGetPackages%2FMonoGame.Content.Builder.nuspec&version=GBdevelop) and [Content.Pipeline](https://hollow-games.visualstudio.com/_git/MonoGame?path=%2FNuGetPackages%2FMonoGame.Framework.Content.Pipeline.nuspec&version=GBdevelop) packages then use this to choose the correct OS package.

cra0zy:
> I do it with conditional PackageReferences

The problem with those is that they do not work properly with C# IDEs. At least that was a problem an year ago, maybe newer IDEs go it fixed (do note that we still need to support older VS IDEs starting with 2013, as well as older versions of msbuild and xbuild).

jnoyola:
> note that we still need to support older VS IDEs starting with 2013, as well as older versions of msbuild and xbuild

Fair point. But then what’s the story for supporting people who build these from source and want the same ability to swap packages based on OS? (Keeping in mind some people use local nuget feeds, in particular, seems to be the more difficult piece here off the top of my head.)

cra0zy:
> But then what’s the story for supporting people who build these from source and want the same ability to swap packages based on OS? (Keeping in mind some people use local nuget feeds, in particular, seems to be the more difficult piece here off the top of my head.)

They can already do that... heck I even made the targets file able to take a custom MGCB location so it works with console versions of MG.

jnoyola:
I’m not sure I understand. The build task is responsible for downloading the development-OS-specific version of the package, not the targets file, right? If someone wants to point to their own package instead of the official one from nuget.org, it looks to me like they’d have to change the build task to find the zip in their local nuget feed/folder.

cra0zy:
> Edit: I see, maybe you’re referring to the path to MGCB.exe itself, but that would mean if you want to use a custom version then you don’t get full nuget support and you have to point it to the binaries already unpacked, unless I’m mistaken.

Yup, adding support for user mg.content.pipeline platform nuget packages is quite a redundant thing.

jnoyola:
I don’t see how it’s redundant. I always test nuget packages from a local feed before I publish them, but if no one else has this concern then I can continue to use my custom method.

cra0zy:
> I don’t see how it’s redundant. As an open source project we should support people customizing these packages.

I mean, if you are making a custom Pipeline nuget package, you'll also want to make a custom core package as well, no point in trying to hook up to an existing nuget package in that case.

jnoyola:
I think it's misleading to publish these as nuget packages though, because we're using nuget here more as file storage than offering full dynamic nuget support.

cra0zy:
I honestly have no opinion on that.

If you have a solution that works on both MonoDevelop and old versions of Visual Studio (and Rider), I wouldn't mind modifying my PR and NuGets to use it.

jnoyola:
I think I finally found something we can agree on. [Here](https://hollow-games.visualstudio.com/_git/MonoGame?path=%2FMonoGame.Framework.Content.Pipeline%2FMonoGame.Content.Builder.Base.targets&version=GBdevelop) is my new targets file (you can ignore the CreatePipelineShortcut stuff at the bottom). It basically combines our ideas.

1. Add dependency on `NuGet.CommandLine` to the base package.
2. Use `nuget install` to download the package into the tools folder based on OS and version. This obviates the need to hardcode feeds, unzip, etc. while also avoiding a nested `PackageReference`.
3. Call the normal BuildContent targets.
   1. I was hoping to be able to `Import` the targets from the platform-specific package, but `Import` will fail since it can't initially resolve the targets file. (A common workaround is to `Import` with a `Condition` on the file's existence, but then this always fails the first build before the file exists.)
   2. However `UsingTask` is much more forgiving and doesn't resolve the file until the task is actually called, which is ensured with `DependsOnTargets`.

I believe this is compatible with all the required platforms, since most of it is stuff we already do. I also left the old targets in the platform-specific Content.Builder packages so they can still be used individually if desired.

cra0zy:
Hmm, I think I misunderstood you regarding what you wanted, here was my thought process on what you wanted:
- make MG.Framework.Content.Pipleine.Platform packages which have the main content pipeline dll file as the thing they reference when added manually
  - so referencing them manually is useful for people who want to compile content during runtime
- make the main nuget reference the specific platform nuget during build, which will get used as content pipeline for the build

jnoyola:
I think that sounds right. Does that not work? I’ll write it out too in case something’s getting missed.

- make MG.Content.Builder.Common.targets with common functionality (calling CollectPackageReferences, exec MGCB, etc.)
- make MG.Content.Builder.Platform
    - contains build tasks, MGCB.exe
    - make MG.Content.Builder.Platform.targets that imports MG.Content.Builder.Common
- make MG.Content.Builder
    - depends on nuget cli
    - make MG.Content.Builder.targets that imports MG.Content.Builder.Common.targets (the wrapper has the extra functionality like calling nuget install and locating MGCB.exe within the platform package)

cra0zy:
> * make MG.Content.Builder.Platform.targets that imports MG.Content.Builder.Common

Why do you want targets files in per platform nugets? That sounds very redundant.

SimonDarksideJ:
Great to see some movement on this PR as it is needed.

jnoyola:
> Why do you want targets files in per platform nugets? That sounds very redundant.

I do want them to be usable as standalone nuget packages, as you said. But if they don't need anything beyond what's in the Common targets then that's fine, just use Common directly. I thought they might need an additional wrapper but perhaps not.

cra0zy:
Oh, I think you misunderstood me with that, there are reasons where you want to reference MonoGame.Framework.Content.Pipeline, like for manually compiling content during runtime, my idea is to have the 3 .Content.Pipeline.Platform packages be used for that purpose while the main package references them and uses them.

jnoyola:
Right, but even if the general use case is to have the main package reference and use them, I think there's still benefit and very little downside in allowing the Platform packages to be used directly if desired.

cra0zy:
Very much against that way of referencing directly as it would allow for incorrect use of the nuget.

Also closing this PR as I'll remake it once Protobuild removal is done.

jnoyola:
I disagree that it's incorrect use. It might not be _your_ favored use of the nuget, but I see no reason to publish Platform packages that intentional don't work on their own even though they have all the necessary binaries to compile content on a single platform.

Also I'm not sure when Protobuild removal will be completed, but I just want to throw it out there that what I'm describing is certainly possible with Protobuild and nuspecs, as that's what I've already done in my own fork.

cra0zy:
> I disagree that it's incorrect use. It might not be _your_ favored use of the nuget or even the _recommended_ use of the nuget, but I see no reason to publish Platform packages that intentionally don't work on their own even though they have all the necessary binaries to compile content on a single platform.

What I'm describing does make them work on their own.

> Also I'm not sure when Protobuild removal will be completed, but I just want to throw it out there that what I'm describing is certainly possible with Protobuild and nuspecs, as that's what I've already done in my own fork.

Its going on right now and I'm focusing my efforts on it, thats why I'm closing this as I'm gonna have to modify stuff afterwards anyway and I might as well leave this for later.

Issue #6293, opened 2018-04-20
Instanced geometry not visible / working on OpenGL when drawing non-instanced geometry with it


YTN0:
Drawing instanced geometry on it's own using OpenGL seems to work fine. However, when also drawing other geometry with it e.g. a model, things start getting weird. In my primary program, I have weird issues with the final image on OpenGL.

On DirectX, things look and work as they should i.e. instanced geometry and non-instanced are fine together.

I put together a test program using the same particle code that was submitted on another bug.

In the test program, if you only display the instanced particles, everything is drawn as expected. As soon as you also display the model, the instanced geometry / particles are no longer drawn / visible. Note, the current code will display the model but not show the particles. Edit and comment out the drawing of the model to see the particles being drawn. 

I *suspect* it has something to do with graphics state and / or buffers, however this is beyond my depth of knowledge.

CC: @Jjagg since you worked on OpenGL instancing.

[ParticlesGL.zip](https://github.com/MonoGame/MonoGame/files/1930946/ParticlesGL.zip)

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

YTN0:
Forgot to mention... if you create an Android project and run the same code (OpenGL ES 3.0+ device), you'll have the exact same problem.

If you create a DirectX project and run the exact same code, it all works correctly i.e. you'll see both the particles and the spaceship model.

YTN0:
Anyone have any ideas what the issue might be here? I have tried everything I can to get this working correctly on GL, but I've hit a brick wall. This works every time on DirectX without any problems, but there's nothing but frustration trying to get this working on GL / GL ES....

PumpkinPudding:
put 
`GraphicsDevice._attribsDirty = true; `
to the end of
 ` private void PlatformDrawIndexedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount)`
  and
  `private void PlatformDrawPrimitives(PrimitiveType primitiveType, int vertexStart, int vertexCount)`

GraphicsDevice.OpenGL.cs

Jjagg:
Awesome, thanks for looking into this @PumpkinPudding! If you think that's the right fix, please do set up a pull request :) 

If you're not familiar with Git, DM me on Gitter and I'll walk you through it when I have the time.

YTN0:
[...]

PR #6300, opened 2018-04-25
Fixing issue #6293 - Instanced geometry not visible / working on OpenGL when drawing non-instanced geometry with it

YTN0:
This fix resolves the issues where drawing instanced geometry with non-instanced geometry results in only the non-instanced drawing being visible i.e. the instanced geometry is not drawn.

See issue #6293

Thanks to @pumkinpudding for pointing out how to resolve this.


YTN0:
Sorry...need to redo this PR... still getting the hang of this forked repo stuff....

PR #6302, opened 2018-04-26
Fixing issue #6293 - Instanced geometry not visible / working on Open…

YTN0:
This fix resolves the issues where drawing instanced geometry with non-instanced geometry results in only the non-instanced drawing being visible i.e. the instanced geometry is not drawn.

See issue #6293

Thanks to @pumkinpudding for pointing out how to resolve this.


dellis1972:
not sure this is the best approach. since it means the attibutes will be marked as dirty after each call to `PlatformDrawIndexedPrimitives` or `PlatformDrawUserPrimitives`. I suspect this will be very inefficient.
@Jjagg ?

Jjagg:
The flag is set in VertexDeclaration.OpenGL.cs when a VertexFeclaration is applied. That's pretty poor code style on my part in retrospect. 

Anyways, this doesn't look like the right fix, since it's not the draw call that requires us to bind new attributes. I'll give this a look this weekend to try and figure out where setting the flag is missing.

dellis1972:
it seems most of methods calls `VertexDeclaration.Apply`. The two not working call `ApplyAttrib` which is in `GraphicsDevice.OpenGL.cs`. Which does similar but not quite the same code as `VertexDeclaration.Apply`. Which is weird.

YTN0:
@Jjagg @dellis1972 Note, I filed a new issue #6306 which might be related to these problems... but isn't fixed by this particular PR / change.

YTN0:
@Jjagg Any ETA on when this fix will be included? I still see this issue in the latest 3.8.0.x development build...

Thanks.

Jjagg:
I completely lost track of this :/ Again though, this is not the right fix and it hurts performance. To fix this right, we should make sure applying attributes happens only in one place (implementation is in two places rn) and the dirty flag is always set correctly.

YTN0:
Hi Jesse,No worries.I filed a couple bugs which are affecting OpenGL and OpenGL ES (Android) and may be related or similar....I am hoping you can find some time to check these out and get in the correct fix for these 😉Thanks!On Jan 6, 2019 12:04 AM, Jesse <notifications@github.com> wrote:I completely lost track of this :/ Again though, this is not the right fix and it hurts performance. To fix this right, we should make sure applying attributes happens only in one place (implementation is in two places rn) and the dirty flag is always set correctly.

YTN0:
I am no longer using Monogame and don't have access to this code anymore.

Not sure if @Jjagg can help here.

SimonDarksideJ:
The code is in the PR @YTN0 , but if you've moved on that is fine and we wish you well on your future gamedev adventures.

@Jjagg , is this something worth taking forward in YTNO's absence?

squarebananas:
@SimonDarksideJ Looks like a duplicate of an issue I fixed with #7406. Also I agree with the posts here from 6 years ago that this isn't the right fix in any case.

SimonDarksideJ:
Closing on the advice of community

PR #6312, opened 2018-05-09
Update gamecontrollerdb.txt

Jjagg:
Grabbed the latest from https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt

@cra0zy


cra0zy:
Can you fix Line encoding so the changes can be seen properly?

Also should we maybe switch to using a submodule instead of a file? Would make it easier to upgrade later on?

Jesuszilla:
I concur with the idea of making it a submodule. Just requires regular checks to make sure the submodule has been updated!

Jjagg:
> should we maybe switch to using a submodule instead of a file

Good idea, I'll update this :)

Jesuszilla:
Any updates on if/when this will be pulled into the dev branch? I'd really love to try out 3.7 and I'm waiting on this before I try it!

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "ThirdParty/NVorbis"]
 	path = ThirdParty/NVorbis
 url=https://github.com/MrHelmut/NVorbis.git
```
mrhelmut: This makes me think that we should probably fork our own https://github.com/ioctlLR/NVorbis or point back to it.

The critical PR that was pending and resulted in pointing to my fork ultimately got merged upstream. NVorbis haven't been updated since then.

PR #6314, opened 2018-05-10
[Math] Improvements to Point, added Point3 class

aaronfranke:
I've added default Point structs, similar to those found in the Vector classes, which makes using Point easier. I've also added Point3 which is the 3D version of Point.

In the future we could consider renaming Point to Point2, although this would be a breaking change.

### DIFF
--- a/MonoGame.Framework/Point3.cs
+++ b/MonoGame.Framework/Point3.cs
@@ -1,0 +1,754 @@
+// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Xna.Framework
+{
+    /// <summary>
+    /// Describes a 3D point on an integer grid.
+    /// </summary>
+    [DataContract]
+    [DebuggerDisplay("{DebugDisplayString,nq}")]
+    public struct Point3 : IEquatable<Point3>
+    {
+        #region Private Fields
+
+        private static readonly Point3 zero = new Point3(0, 0, 0);
+        private static readonly Point3 one = new Point3(1, 1, 1);
+        private static readonly Point3 negOne = new Point3(-1, -1, -1);
+        private static readonly Point3 unitX = new Point3(1, 0, 0);
+        private static readonly Point3 unitY = new Point3(0, 1, 0);
+        private static readonly Point3 unitZ = new Point3(0, 0, 1);
+        private static readonly Point3 up = new Point3(0, 1, 0);
+        private static readonly Point3 down = new Point3(0, -1, 0);
+        private static readonly Point3 right = new Point3(1, 0, 0);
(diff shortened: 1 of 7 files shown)


aaronfranke:
What is exit code 1? I can see that it's failing to build but I don't know why.

dellis1972:
`Point.cs(20,50): error CS0246: The type or namespace name 'Point3' could not be found (are you missing a using directive or an assembly reference?) `

dellis1972:
If you click on the Build Log tab when you go to team city, you can see the full log :)

aaronfranke:
I fixed the issue, it should be good now.

MonoGame.Framework/Point3.cs
```diff
@@ -0,0 +1,346 @@
+// MIT License - Copyright (C) The Mono.Xna Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
```
tomspilman: Go ahead and change this to the MonoGame copyright since it is new code:
```
```

tomspilman:
@dellis1972 - I think if we merge new math types like this we should require unit tests for them.  Agreed?

aaronfranke:
@tomspilman Good point. Any particular tests you can think of? Should we test addition/subtraction/multiplication/creation/deconstructing/conversion? Or is that too many?

There is also a file called `Test/Framework/PointTest.cs` which is not referenced in `Build/Projects/MonoGame.Tests.definition`, although it only contains a single test (deconstructing).

tomspilman:
> Any particular tests you can think of? 

We would want to fully test all the APIs in the new class.

I would start with the example of the `Vector3` tests and adapt them to this new type.

aaronfranke:
Point3Test.cs(14,13): error CS0246: The type or namespace name 'Point3' could not be found

I'm confused. Point3.cs is in `namespace Microsoft.Xna.Framework` and Point3Test.cs contains `using Microsoft.Xna.Framework;`, why can't `Point3Test` find `Point3`?

dellis1972:
@aaronfranke you need to add the new PointX.cs files to MonoGame.Framework.definition for them to be included in the assemblies.

[1]https://github.com/MonoGame/MonoGame/blob/develop/Build/Projects/MonoGame.Framework.definition#L321

aaronfranke:
@tomspilman Let me know if these tests are good. I took many of the relevant tests from `Vector3Test.cs`. I think I got it all figured out, thanks to help from dellis.

Also worth noting that in addition to testing `Point3`, I've added tests for the existing `Point` struct.

aaronfranke:
Will this be merged? I'll ping @Jjagg since he seems to be the one merging PRs.

Jjagg:
The changes look really good to me. Thanks @aaronfranke!
We don't really have a clear cut policy for adding new types though and this has been stuck in limbo for a long time, so I'll let Tom handle this.

aaronfranke:
If anyone is interested, they are welcome to salvage this PR. However, I won't be touching this anymore, as I don't use MonoGame and never will.

PR #6318, opened 2018-05-13
Added Round, Ceiling, Floor functions to Vector{2,3,4}

portal-chan:
Needed these a couple of times. All of the overloads are based on the `Normalize` functions.

### DIFF
--- a/MonoGame.Framework/Vector3.cs
+++ b/MonoGame.Framework/Vector3.cs
@@ -303,7 +303,7 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
-        /// Round this <see cref="Vector3"/> to a vector towards the positive infinity.
+        /// Round the members of this <see cref="Vector3"/> towards positive infinity.
         /// </summary>
         public void Ceiling()
         {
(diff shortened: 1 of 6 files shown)


dellis1972:
This is a really good PR :) includes docs and tests nice job :D.

@MonoGame/maintainers are we ok with adding these to the framework?

Jjagg:
I like these. The 'rounded towards positive/negative infinity' wording is a bit weird IMO. Other than that this looks good!

MonoGame.Framework/Vector3.cs
```diff
@@ -853,6 +923,41 @@ public static void Reflect(ref Vector3 vector, ref Vector3 normal, out Vector3 r
...
+            X = (float)Math.Round(X);
+            Y = (float)Math.Round(Y);
+            Z = (float)Math.Round(Z);
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="Vector3"/> that contains rounded values from another vector.
+        /// </summary>
+        /// <param name="value">Source <see cref="Vector3"/>.</param>
+        /// <returns>Unit vector.</returns>
```
tomspilman: This seems misleading as the return value description.  "Unit vector" typically means "a vector that has a magnitude of one".

MonoGame.Framework/Vector3.cs
```diff
@@ -853,6 +923,41 @@ public static void Reflect(ref Vector3 vector, ref Vector3 normal, out Vector3 r
...
+        /// </summary>
+        public void Round()
+        {
+            X = (float)Math.Round(X);
+            Y = (float)Math.Round(Y);
+            Z = (float)Math.Round(Z);
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="Vector3"/> that contains rounded values from another vector.
```
tomspilman: This should mention how this function works like the docs in `Round()` above.

tomspilman:
> @MonoGame/maintainers are we ok with adding these to the framework?

Yeah i'm good with adding them.  I added some comments as some of the documentation seemed wrong.

Jjagg:
I fixed the doc comments. I'm merging this! Thanks for the contribution @portal-chan :)

PR #6331, opened 2018-06-05
Fix crash in Windows when mgcb file is in root directory.

stromkos:
Changes path to include "" if path ends in ":".

In Windows the root directory is returned as "C:"; which crashes the file open dialog when passed in as default directory. This code appends the "" yielding "C:" which is valid in all contexts.

Should not break anything, unless someone names a directory or file that ends in ":"

Application note: The Pipeline tool also requires an "obj " directory to exist prior to building.

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -657,6 +657,10 @@ namespace MonoGame.Tools.Pipeline
         {
             var path = GetFullPath(GetCurrentPath());
 
+            // if path is root of drive in windows (C:) return C:\ instead.
+            //  Should no break anything unless file/directory names end with :
+            if (path[path.Length - 1] == ':') path += "\\";
+
             List<string> files;
             if (!View.ChooseContentFile(path, out files))
                 return;


Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -657,6 +657,10 @@ public void Include()
         {
             var path = GetFullPath(GetCurrentPath());
 
+            // if path is root of drive in windows (C:) return C:\ instead.
+            //  Should no break anything unless file/directory names end with :
+            if (path[path.Length - 1] == ':') path += "\\";
```
cra0zy: This is not a good solution for fixing this bug, you are basically fixing one bug and creating another. 

Tho from the sounds of the bug it should get fixed on the Eto.Forms side as well.

MichaelDePiazzi:
You've got this a bit mixed up I think. On Windows, `C:` is not specifically the root directory of the C drive. Only `C:\` is the root directory.

Using a drive letter by itself is not very common, but it just means the current directory on that drive. For example, if you use `C:Game` then it will only point to `C:\Game` if the current directory of `C:` is the root directory. If the current directory of `C:` is `C:\Dev`, then it will be pointing to `C:\Game\Dev`. It's actually shorthand for `C:.\Game` (not `C:\Game`).

Tools/Pipeline/Common/PipelineController.cs
stromkos: Both Mono and .Net running on Windows throw a NotSupported Exception(https://msdn.microsoft.com/en-us/library/system.io.path.getfullpath(v=vs.100).aspx) if there is a ":" anywhere in the string that is not a drive specifier.

Please point me to the accepted Platform check for Windows.  I will add that to the code then the bug created will not exist.

The form is properly handling the exception.

cra0zy: For a proper check you need to check if the Length of the path is 2 and first character is not `/`. Also the check should be moved inside `GetFullPath` method.

Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -1065,6 +1065,10 @@ private void LoadTemplates(string path)
 
         public string GetFullPath(string filePath)
         {
+            // if the path is root of a drive in windows, which normally returns C:, return C:\ instead.
+            if (filePath.Length==2 && filePath[0].ToString().ToUpper()[0] >='A' && filePath[0].ToString().ToUpper()[0] <='Z' && Path.DirectorySeparatorChar=='\\' && filePath[filePath.Length - 1] == ':')
```
cra0zy: You don't need to check if the first character is within upper letter range, you just need to check if its not `/` so we ensure that this is not a Unix path. Also your code right now does not work with `c:`.

stromkos:
That should be correct now. Sorry about the merges, git is requiring a pull after I commit the changes locally, before I can push.

Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -1066,7 +1066,7 @@ private void LoadTemplates(string path)
         public string GetFullPath(string filePath)
         {
             // if the path is root of a drive in windows, which normally returns C:, return C:\ instead.
-            if (filePath.Length==2 && filePath[0].ToString().ToUpper()[0] >='A' && filePath[0].ToString().ToUpper()[0] <='Z' && Path.DirectorySeparatorChar=='\\' && filePath[filePath.Length - 1] == ':')
+            if (filePath.Length==2 && filePath[0] !='/' && Path.DirectorySeparatorChar=='\\' && filePath[filePath.Length - 1] == ':')
```
cra0zy: Fix formating and there is no reason to check `Path.DirectorySeparatorChar`.

cra0zy: Wait, I'll just write you the line myself, it will be easier, need to wash my hands.

cra0zy:
```c#
if (filePath.Length == 2 && filePath[0] != '/')
```

stromkos: It may be possible for filePath to be a relative path (see line 1076). To prevent a match for "\a" and "aa", I check the second char for ":"in the last commit.

cra0zy: Good point, move the code under `if (Path.IsPathRooted(filePath))`, forgot to mention it before.

stromkos: The problem with moving it under IsPathRooted is the case of the Pipleline executable in the root of the drive would fail to open a project.

By leaving it where it is and adding the colon check, there are no cases excepted and has minimal overhead.

cra0zy: > The problem with moving it under IsPathRooted is the case of the Pipleline executable in the root of the drive would fail to open a project.

Why would it fail?

cra0zy: Tho do note that I'm also talking about modifying the methods `if (project == null)` thingy as well, the build bot is down so once its back online I'll just fix this properly.

cra0zy: https://pastebin.com/aQwbEm3B

stromkos: > Why would it fail?
The `_project==null` does not guarantee an absolute path.

cra0zy: It does actually, if the project is not open, relative paths can't be resolved.

stromkos: Windows doesn't seem to care about the number of backslashes beyond the beginning of the path, which it interprets multiples as a UNC path.
We should be good to go with that last commit.

Thank you for working with me on this.

tomspilman:
Thanks @stromkos ... merging!

PR #6335, opened 2018-06-10
Support unicode in window title under DesktopGL

Jjagg:
This issue was reported on the forums: http://community.monogame.net/t/encoding-problem/10700 

Encodes window title with UTF-8 so SDL handles it right.
https://wiki.libsdl.org/SDL_SetWindowTitle

Before:
image

After:
image

### DIFF
--- a/MonoGame.Framework/SDL/SDL2.cs
+++ b/MonoGame.Framework/SDL/SDL2.cs
@@ -29,7 +29,7 @@ internal static class Sdl
             ret = FuncLoader.LoadLibrary(Path.Combine(assemblyLocation, "x86/libSDL2-2.0.so.0"));
         else if (CurrentPlatform.OS == OS.MacOSX)
             ret = FuncLoader.LoadLibrary(Path.Combine(assemblyLocation, "libSDL2-2.0.0.dylib"));
-        
+
         // Load system library
         if (ret == IntPtr.Zero)
         {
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/SDL/SDL2.cs
```diff
@@ -431,7 +431,7 @@ public static void SetFullscreen(IntPtr window, int flags)
         public static d_sdl_setwindowsize SetSize = FuncLoader.LoadFunction<d_sdl_setwindowsize>(NativeLibrary, "SDL_SetWindowSize");
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-        public delegate void d_sdl_setwindowtitle(IntPtr window, string title);
+        public delegate void d_sdl_setwindowtitle(IntPtr window, ref byte value);
```
cra0zy: You should create a helper function which accepts string in SDL2.cs and not modify any code in SDLGameWindow.

Also is `in` allowed instead of `ref`?

MichaelDePiazzi: @Jjagg @cra0zy Not sure if you're aware, but there's also [directional attributes](https://msdn.microsoft.com/library/241ac5b5-928e-4969-8f58-1dbc048f9ea2) which can be used to optimise marshaling. So even though you can't use `in`, you can still use [`InAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.inattribute?view=netframework-4.5). I know it works with `extern`, but I'm not sure when using delegates as I haven't tried that before. But it should work the same way I'd assume. e.g.
```c#
[UnmanagedFunctionPointer(CallingConvention.Cdecl)]
private delegate void d_sdl_setwindowtitle(IntPtr window, [In] byte[] value);
```
```c#
SDL_SetWindowTitle(handle, Encoding.UTF8.GetBytes(title));
```
The code is fine though. This is not a nit-pick, just more of an FYI. Hope it's useful!

Jjagg: Yeah, we should probably do a run through for all delegates and add `In` or `Out` where applicable (it does work the same). I didn't add it because we don't use them anywhere.

cra0zy: `
Directional attributes are optional. You apply them to method parameters when you want to alter the default behavior of the marshaler. If you omit directional attributes from a method parameter, the marshaler determines the directional flow based on the type of the parameter (value or reference) and its modifiers, if any.
`

tomspilman:
Merging!

Issue #6338, opened 2018-06-13
XML doc request


stromkos:
Would it be possible to include XML schema in the MonoGame documentation for the serializable classes(including Reflection and Helpers)?

While the helper classes would need to be manually updated (include a comment to ensure changes are reflected in the documentation), the reflective serializer can utilize xsd.exe to generate the proper schema and wrap it with the appropriate tags.

I have concluded that attempting to solve #4792, the only solutions utilizing XmlReader.Create() instead of XmlTextReader, forces any solution that exists to be in exponential time.  

While I could mitigate the problems, by eliminating already read elements, I do not feel like my approach is the appropriate solution, given my solution only works for basic cases.
I am opting for option "B" to the original problem. Give the user the needed feedback to correct the situation. Instead of only informing the user the the expected element is missing;  Give the user the expected element AND a reference to the expected schema.
Even if the proper schema is not available; This would inform the user of the out-of-order elements and the opportunity to correct them, instead of only informing of the missing element.

I hope to provide a PR with this issue, to inform the user the element is out of order(already in the works), 
I also wish to return a reference the appropriate schema to be utilized when implementing the class(Hence the Issue).

Jjagg:
I'm closing this. I don't think the bookmarking solution is a good approach instead we should reverse control as suggested in #4792 and look up fields/properties as the XML elements are parsed.

PR #6339, opened 2018-06-13
Indicate which elements are out of order. Rather than indicating only…

stromkos:
… missing element.

It would be better to point the user to a documentation page indicating proper XML Schema for additional info. See #6338.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
@@ -174,7 +174,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate
                             continue;
                         
                         // We failed to find a required element.
-                        throw new InvalidContentException(string.Format("The Xml element `{0}` is required!", info.Attribute.ElementName));
+                        throw new InvalidContentException(string.Format("The Xml element `{0}` is required, But found '{1}' instead, try swappping elements as they are out of order", info.Attribute.ElementName, input.Xml.Name));
                     }
                 }
 


stromkos:
This update gives feedback to the user, when an invalid XML element is encountered; and it reports the
expected element , as well as, the encountered element.
This will inform the user an out of order element was encountered and which element was expected.

nkast:
There are other reasons this error can occur. The element could be renamed, or missing.
Let's not propose a solution which is in 1/3 of the cases.

Jjagg:
> Let's not propose a solution which is in 1/3 of the cases.

Agreed. Though it might be helpful to note that the elements are ordered without suggesting it's what's wrong.

stromkos:
I want to give the user more information more information to correct the error. If it is renamed or deleted this PR will notify the user of the discrepancy. I have reworded the error.  While this may not cover all of the scenarios at which this error may occur, at least it assists the user by notifying them of the expected element.  I was at a loss when I received this message and would like to make sure future users do not have the same issue.

stromkos:
@nkast In your view 2/3s of the cases are currently excluded from the existing wording *without explanation*.  I have posted an update that includes my updates and the existing case.  I have purposefully excluded the rename case as it has zero application in this discussion(no one would expect a malformed XML file to process), although if it did happen, the message would indicate it as a found/expected pair, indicating to the user the corrective action(n a major improvement from the current situation).

nkast:
>> @nkast In your view 2/3s of the cases are currently excluded from the existing wording without explanation. 

On the contrary, I would prefer a simple message with only the facts about the failure. Not an extended list of all possible causes/fixes. 
The problem I saw with the first wording was that it might be confusing in some cases.  Anyways, the new wording is not bad.

nkast:
Is it possible to get Line numbers from the XmlReader (IXmlLineInfo interface)?

Jjagg:
IIRC you need to explicitly cast to IXmlLineInfo as it's implemented as an explicit interface implementation.

MonoGame.Framework.Content.Pipeline/Serialization/Intermediate/ReflectiveSerializer.cs
```diff
@@ -174,7 +174,7 @@ protected internal override object Deserialize(IntermediateReader input, Content
                             continue;
                         
                         // We failed to find a required element.
-                        throw new InvalidContentException(string.Format("The Xml element `{0}` is required!", info.Attribute.ElementName));
+                        throw new InvalidContentException(string.Format("The Xml element `{0}` is required, But found '{1}' instead, try swappping elements as they may be out of order, or required elements may be missing altogether", info.Attribute.ElementName, input.Xml.Name));
```
tomspilman: Some grammar fixes:

```
The Xml element `{0}` is required, but element `{1}` was found.  Try changing the element order or adding missing elements.
```

stromkos: Do you want the line number as well?
```
The Xml element `{0}` is required, but element `{1}` was found at line {2}. Try changing the element order or adding missing elements.", info.Attribute.ElementName, input.Xml.Name,((IXmlLineInfo)input.Xml).LineNumber);
```
Or shorter wording with position:
```
The Xml element `{0}` is required, but element `{1}` was found at {2},{3}. Try changing the element order or adding missing elements.", info.Attribute.ElementName, input.Xml.Name,((IXmlLineInfo)input.Xml).LineNumber, ((IXmlLineInfo)input.Xml).LinePosition);
```

stromkos:
Does anyone object to the addition of the Xml Namespace(it is tied to at least functions of the class,which justifies the using statement)?(if not, I will fully qualify the IXmlLineInfo as appropriate)
If everyone approves of the new wording @tomspilman, @nkast, @Jjagg, @cra0zy .  I am always open to suggestions.  
Thank You.

PR #6344, opened 2018-06-24
Added a BeginCalled property to SpriteBatch to tell when Begin was called

tdeeb:
I added this simple property in SpriteBatch to check if Begin was called. I always thought it was a bit off that SpriteBatch throws an exception saying Begin was already called while there is no way to check for it.

I'm thinking of this similarly to data structures, such as a Stack, that throw InvalidOperationExceptions. In those structures, you can check for the precondition. I think it makes sense for SpriteBatch to be capable of doing so as well.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -38,6 +38,11 @@ namespace Microsoft.Xna.Framework.Graphics
         internal static bool NeedsHalfPixelOffset;
 
         /// <summary>
+		/// Tells whether Begin has already been called on this SpriteBatch.
+		/// </summary>
+		public bool BeginCalled { get { return _beginCalled; } }
+        
+        /// <summary>
         /// Constructs a <see cref="SpriteBatch"/>.
         /// </summary>
         /// <param name="graphicsDevice">The <see cref="GraphicsDevice"/>, which will be used for sprite rendering.</param>


stefanrbk:
A `SpriteBatch.Begin` really should only be called when starting a procedure, followed by`SpriteBatch.End` when you are done. How would this be used?

Jjagg:
@tdeeb I'm not sure this is useful. `SpriteBatch.Begin` sets state for rendering. When calling `Draw` you care about that state, so you should be well aware of how and where the `Begin` call happened anyway.

In case of your example there is no state, so it's not a good analogy IMO.

tdeeb:
That makes sense; I just found it off that you can't check for the state. I'll keep that in mind going forwards.

stefanrbk:
TBH, you also shouldn't have to check the state. If you have multiple groups of things to draw with different Draw settings, like in the game I'm building, you never need to check. 
Psudocode of what I do: 
```
DrawMapLower();
SpriteBatch.Begin(spriteSettings);
DrawSprites();
SpriteBatch.End();
DrawMapUpper();
SpriteBatch.Begin(hudSettings);
DrawHud();
SpriteBatch.End();
```

tomspilman:
I can see how `SpriteBatch.BeginCalled` could be useful, but i agree with @Jjagg that the state really matters.  So this is likely just useful for debugging if anything.

cra0zy:
I don't see any use in it, and since @Jjagg already commented above, closing...

Beyley:
in my rendering code i could benefit from this for error checking, i give some functions (that are implemented from a base class) raw access to the SpriteBatch in an ended state and i *expect* it to start and end it, but if it doesn't, that would cause a bit of a hard bug to track down in the case that someone forgets to put an End to the spritebatch before passing it back, so it'd be nice to do a quick little check `if(BeginCalled) throw new BlaBlaBlaException()` to give the dev working on it a bit more of a useful error message, along with giving it in an actually useful place. The "try to `Begin` and catch exception" (not clean at all), "write a wrapper for SpriteBatch" (seems overkill for something so simple) and "use reflection" (too slow) are all things that seem like kinda a mess in the long run, so possibly having this PR opened for reevaluation would make me very grateful

PR #6346, opened 2018-06-24
Use FuncLoader for OpenAL

cra0zy:
This could use a bit of testing on Android/iOS.

CC. @mrhelmut @dellis1972

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -1249,6 +1249,12 @@
     <Compile Include="Utilities\FuncLoader.Desktop.cs">
       <Platforms>Angle,Linux,WindowsGL,MacOS</Platforms>
     </Compile>
+    <Compile Include="Utilities\FuncLoader.Android.cs">
+      <Platforms>Android</Platforms>
+    </Compile>
+    <Compile Include="Utilities\FuncLoader.iOS.cs">
+      <Platforms>iOS,tvOS</Platforms>
+    </Compile>
     <Compile Include="SDL\SDLGamePlatform.cs">
       <Platforms>Angle,Linux,WindowsGL,MacOS</Platforms>
     </Compile>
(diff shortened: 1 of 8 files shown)


mrhelmut:
Awesome. I can't test this on mobile, but I'll give this a shot on Windows.

dellis1972:
@cra0zy interestingly I was in Redmond last week and I was told they are adding support for `dllmap` to .net core (apparently). I've yet to find the appropriate commit though.

cra0zy:
> @cra0zy interestingly I was in Redmond last week and I was told they are adding support for dllmap to .net core (apparently). I've yet to find the appropriate commit though.

I guess possible support in .NET Core 3.0 at the earliest, for now this allows us to properly support .NET Core 2.0.

tomspilman:
Anyone get a chance to test this on iOS or Android yet?  @dellis1972 ?

mrhelmut:
All good for me on Windows, Mac OS, and Linux.

The only issue I have with this is with MonoKickstart dll discovery. It works fine, but let's say that I have the Steamworks API there, MonoKickstart will still look for those libraries in the ```lib``` and ```lib64``` folders, not the ```x86``` and ```x64``` ones (so for any external libraries, you end up with two different dll path that are both required).

Good thing is that the ```MonoGame.Framework.dll.config``` map is indeed useless now.

It's no big deal, but it'd cool to have either MonoKickstart fixed or Linux/Mac OS path (though it'd be different from Windows, so I'd vote for MonoKickstart).

cra0zy:
Actually the code of this PR right now first tries to load `x86` and `x64` folders, however if it does not find any libraries in them in falls back to trying to load them from the system specified location, and MonoKickstart is overriding the PATH variable so it will use the ones from MonoKickstart.

In other words you can choose where you want the libs to be, you don't need 2 copies, it checks both locations.

cra0zy:
(also you could have just fixed the .dll.config file with the previous versions to use same paths as kickstart, or modified the kickstart start script to use same locations as stable MG, its pretty straightforward)

mrhelmut:
I'll figure this out within MonoKickstart, it's no big deal.

Good to go for me! Thank you!

cra0zy:
@tomspilman @dellis1972 Can this get merged?

tomspilman:
Ok... lets give it a try.  Worst case we get a bug report we broke Android.

PR #6348, opened 2018-06-27
Support text compositions

ClEngine:
@Jjagg Support text compositions


ClEngine:
The first time I participated in such a code submission, I don't know what the error here means. Where am I writing something wrong?

PR #6349, opened 2018-06-27
Remove unnecessary links from Content.Pipeline to 2MGFX

nkast:


nkast:
Mac and linux tests failed.
```
Microsoft.Xna.Framework.Content.Pipeline.InvalidContentException : Windows effects are not supported.
   at Microsoft.Xna.Framework.Content.Pipeline.Processors.EffectProcessor.Process(EffectContent input, ContentProcessorContext context) in c:\BuildAgents\MonoGameWin2\work\f7381a85a626990\MonoGame.Framework.Content.Pipeline\Processors\EffectProcessor.cs:line 61
```

?

tomspilman:
What as the idea here @nkast ?

Unless i am mistaken the pipeline assembly directly includes 2MGFX because it calls it that way.  It doesn't call it via the command line... although maybe it should?

nkast:
At some point the importer calls 2MGFX indirectly through the command line, but the writer/reader also need some data structures. It's best if we could remove all those dependencies to the 2MGFX.  
For start I removed a couple of linked files that didn't seem necessary, all the projects compile but then the tests failed for some reason.

tomspilman:
> At some point the importer calls 2MGFX indirectly through the command line,

I don't think you have that right...

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs#L54


> It's best if we could remove all those dependencies to the 2MGFX.

 I think it could be nicer to make `2MGFX.exe` be called from the pipeline, but it will complicate things for others using our custom pipeline now.  They need to be sure `2MGFX.exe` is included with their project as before they only needed `MonoGame.Framework.Content.Pipeline.dll`.

Also 2MGFX.exe would have to be extended to support some features the pipeline needs via the command line instead of how it does it now by calling the code directly.

I don't think this is high priority, but would be nice to clean up in the future.

tomspilman:
>  something like that

I honestly don't remember.  But it does sound like a mess to clean up.

PR #6208, opened 2018-02-23
[Piepline Tool] Upgrade Eto

cra0zy:
**NEEDS:** https://github.com/MonoGame/MonoGame.Dependencies/pull/118

This PR just upgrades Eto code for 2.4.1. Most notable stuff are:
- Eto Forms now supports drag and drop, so goodbye Xwt treeview
  - This does not implement drag and drop functionality since I am reworking part of its logic so that we can support linking, I just need this upgrade so I don't have to write Xwt code
- Eto is using my own fork of GtkSharp which requires no native glue libraries and does not use dllmaps for calling native assemblies, this means that the Linux installer now has 1 less dependency to worry about
  - This also allows me to ditch a lot of the wrapper code for Gtk stuff

PR #6350, opened 2018-06-28
Updated MGCB and Pipeline EXEs to use Net Framework 4.7.1 instead of 4.5.0

vpenades:
Right now, MGCB and the Pipeline command line tool target Net Framework 4.5.0 on windows. That means that it is not possible to create content processor plugins which depend on assemblies with higher version, or net standard assemblies.

In order to solve the issue, I've raised the framework version from 4.5.0 to 4.7.1

This change only affects:
- MGCB.exe
- Pipeline.exe

Also, I've only changed it for Windows, since I don't know which are the latest version of MONO on mac and linux.... I've left the properties commented so others might fill in the appropiate values if needed.

### DIFF
--- a/Build/Projects/MGCB.definition
+++ b/Build/Projects/MGCB.definition
@@ -16,13 +16,13 @@
 
     <FrameworkVersions>
       <Platform Name="Windows">
-        <Version>v4.6.2</Version>
+        <Version>v4.5</Version>
       </Platform>      
       <Platform Name="MacOS">
-        <Version>v4.6.2</Version>
+        <Version>v4.5</Version>
       </Platform>
       <Platform Name="Linux">
-        <Version>v4.6.2</Version>
+        <Version>v4.5</Version>
       </Platform>      
     </FrameworkVersions>
     
--- a/Build/Projects/MGCB.definition
+++ b/Build/Projects/MGCB.definition
@@ -17,15 +17,13 @@
     <FrameworkVersions>
       <Platform Name="Windows">
         <Version>v4.7.1</Version>
-      </Platform>
-      <!--
+      </Platform>      
       <Platform Name="MacOS">
-        <Version>4.0</Version>
+        <Version>v4.7.1</Version>
       </Platform>
       <Platform Name="Linux">
-        <Version>4.0</Version>
-      </Platform>
-      -->
+        <Version>v4.7.1</Version>
+      </Platform>      
     </FrameworkVersions>
     
     <LangVersion>Default</LangVersion>
--- a/Build/Projects/MGCB.definition
+++ b/Build/Projects/MGCB.definition
@@ -16,13 +16,13 @@
 
     <FrameworkVersions>
       <Platform Name="Windows">
-        <Version>v4.5</Version>
+        <Version>v4.7.1</Version>
       </Platform>      
       <Platform Name="MacOS">
-        <Version>v4.5</Version>
+        <Version>v4.7.1</Version>
       </Platform>
       <Platform Name="Linux">
-        <Version>v4.5</Version>
+        <Version>v4.7.1</Version>
       </Platform>      
     </FrameworkVersions>
     
--- a/Build/Projects/MGCB.definition
+++ b/Build/Projects/MGCB.definition
@@ -13,6 +13,20 @@
   </References>
 
   <Properties>
+
+    <FrameworkVersions>
+      <Platform Name="Windows">
+        <Version>v4.7.1</Version>
+      </Platform>
+      <!--
+      <Platform Name="MacOS">
+        <Version>4.0</Version>
+      </Platform>
+      <Platform Name="Linux">
+        <Version>4.0</Version>
+      </Platform>
+      -->
+    </FrameworkVersions>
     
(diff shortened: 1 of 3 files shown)


Build/Projects/MGCB.definition
```diff
@@ -13,6 +13,20 @@
...
 
   <Properties>
+
+    <FrameworkVersions>
+      <Platform Name="Windows">
+        <Version>v4.7.1</Version>
+      </Platform>
+      <!--
+      <Platform Name="MacOS">
+        <Version>4.0</Version>
```
dellis1972: why are we using v`4.0` on macOS/Linux? Linux should certainly be able to use v4.7.1.

dellis1972: You should be able to use `v4.7.1` on all the platforms.

vpenades: macOS and Linux are not using 4.0, the lines are commented, since I don't know if the latest version of windows net framework matches those used by Mono.

I can give a try, though.

vpenades:
It seems TeamCity fails whenever I set a target framework higher than v4.5 ... in my computer it builds fine.

Is TeamCity using VS2017 MSBuild?

cra0zy:
The MG build bot does not have the new enough version of mono to build it for the Linux/Mac side the Windows side should get built tho. I know cause I wanted to bump the versions of content tools to 4.7.1.

vpenades:
@cra0zy the Windows version does not build either if you target 4.7.1, I suspect it's because the build server is using an old version of MSBuild

nkast:
that's the first error of Windows build.

```
[03:18:07][NAnt output]      [exec] (GetReferenceAssemblyPaths target) -> 
[03:18:07][NAnt output]      [exec]   C:\Program Files (x86)\MSBuild\12.0\bin\Microsoft.Common.CurrentVersion.targets(991,5): warning MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.7.1" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend. 
```

tomspilman:
I guess i need to update the build server to use VS2017 for this to work correct?

Will this hurt any of the other builds we're leaving on 4.5 for now?

nkast:
on the `lowering to net4.6.2` the error was from the tests. 
The test project has to be upgraded to 4.6.2.

nkast:
> I guess i need to update the build server to use VS2017 for this to work correct?

You can try to install the 4.71 DevPack from here. That's the link that VS2013/2015 gave me to install newer frameworks.
https://www.microsoft.com/net/download/visual-studio-sdks

tomspilman:
> You can try to install the 4.71 DevPack from here.

I'll give that a shot now.   @vpenades  can you re-break this PR by moving it back to 4.7,1?

vpenades:
@tomspilman trying again... , I'll also update the test

tomspilman:
Thanks @vpenades ...  i'll get this fixed now.

tomspilman:
@dellis1972 - Looks like installing the 4.71 DevPack as @nkast suggested worked for the Windows box.

What do i need to do for the Mac box?  Is there a similar runtime install or do i have to upgrade XS/VS there?

vpenades:
@tomspilman about the Mac/Linux versions, I really don't how how they're related to the windows version.

If Mac/Linux use Mono, [I guess you could try updating the mono framework](http://www.mono-project.com/download/stable/). , But since its MSBuild the one that's compiling for all the cases, I really don't know if there's some mapping under the hood.

Maybe we could leave Mac/Linux at v4.5 for now.

vpenades:
@cra0zy @tomspilman Well... so far, my PR solves the issue of how to tell Protobuild the framework version, and sets the latest known framework version for Windows. I don't think I can move further on my own, since I don't have access to a Linux/Mac devices, nor I know which would be the right frameworks to download there.

tomspilman:
@dellis1972 ?

dellis1972:
4.7.1 should be ok for both linux and Mac. We might need to install a newer `mono` on the build machine though. // cc @tomspilman

cra0zy:
I totally forgot to try using msbuild instead of xbuild during #6208, so need to update the Pipeline Tool as well.

Anyway @vpenades , once Deans PR gets merged, I'll fix this, there are few extra steps needed.

vpenades:
@cra0zy Oookey, let me know when it's ready.

cra0zy:
Was gonna bump the .NET Framework requirement for tools to 4.7.1 however turns out that build bot is still broken :(

@tomspilman @dellis1972 currently I see that build bot has mono 5.8, however the current stable version is using 5.12 and the VS for Mac comes with 5.10. We should bump it up.

https://www.mono-project.com/download/stable/
https://www.mono-project.com/download/vs/

Also closing this as I've taken over this task, move any further discussions to: #6403

PR #6355, opened 2018-07-01
Updated to the latest StbSharp

rds1983:
This PR updates StbSharp to the latest version.
It's mostly refactoring update: no new functionality had been added, but existed code became less chaotic.
Now every StbSharp module lies in the corresponding class(StbImage and StbImageWrite in case of MG). Also all of the C runtime emulation had been moved in the corresponding class(CRuntime.cs).
Also I've renamed "Imaging" subfolder to "StbSharp", as previous name caused some confusion.


cra0zy:
@dellis1972 @Jjagg Should we switch to a submodule for this?

Jjagg:
I'm in favor of using a submodule for this. Files would be included just like how we do it for NVorbis.

rds1983:
Closing this as StbSharp is likely to be included as submodule.
Maybe later I'll provide such PR.

PR #6357, opened 2018-07-02
Remove the use of xbuild from the default.build script

dellis1972:


tomspilman:
Cool... thanks @dellis1972 !  Merging!

tomspilman:
@dellis1972  - Does this also mean we now exclusively depend on VS on Mac now?

dellis1972:
msbuild is independent of XS or VSForMac, but it does mean people need a
recent mono to build. but to be honest its been that way for a while now.

PR #6358, opened 2018-07-02
Visual Basic FULL Template creation. READ DESC.

SpiceyWolf:
A fore-warning, I am not totally familiar with the new format so I do not know how to actually put this in a template format to boot up with VS for testing, im just relying on the factor that i got VS to seemingly load a few that have been converted without errors...

As well - I hand converted everything (From text editors) as I AM experienced doing, only using an online converter for minor assistence - I am unsure if any small things may have been missed, though Visual Studio code editor didn't report any more errors on all files I have modified and I am convinced I havent missed any files. csproj files and everything were modified internally through notepad to boot up with VB properly.

I have no ability to test alot of this stuff in general even 'if' I have the templates as I have a really poopy machine right now and the only thing its even capable of running for sure is the DesktopGL project which seems to be fine to me.

I do not know MonoDevelop/Xamarin well enough to know for sure if VB.net compatibility exists for the mobiles and to what degree if so provided all the syntax is correct either so chances of those not working at all.

Despite these factors I will be getting a new pc soon capable of running all the software for sure, in a month or 2, and should anyone have trouble with any of the templates I fully plan to be responsible for figuring out if its a mistake on their end, a mistake with my templates (With me correcting them officially), or if its a problem with VB.net itself that makes such usage impossible.

(Side Note: I did in fact update the default.build file as well to include the VB files with the C# ones it contains.)


Jjagg:
Thanks for this PR. Sorry we didn't respond to this.

I'm gonna close this because there's some changes happening to templates and we're gonna use SDK style project files so we can't merge this PR like this. It's also easier to get these templates merged 1 by 1 so it's easy to review. Still some of the files here can definitely be used for VB templates.

@cra0zy Some of this will be useful when creating VB versions of the new templates.

cra0zy:
Don't worry, I know exactly how to "port" templates thanks to these: https://github.com/GtkSharp/GtkSharp/tree/develop/Source/Templates

PR #6361, opened 2018-07-04
[SDL] Reduce some garbage with TextEvent (and do a bit of overoptimization)

cra0zy:


MonoGame.Framework/SDL/SDLGamePlatform.cs
```diff
@@ -27,12 +27,14 @@ public override GameRunBehavior DefaultRunBehavior
 
         private int _isExiting;
         private SdlGameWindow _view;
+        private TextInputEventArgs _textInputEventArgs;
```
tomspilman: This makes me wonder... does `TextInputEventArgs` need to be a class?  Can it not be a struct?

MonoGame.Framework/SDL/SDLGamePlatform.cs
```diff
@@ -150,23 +156,32 @@ private void SdlRunLoop()
                 }
                 else if (ev.Type == Sdl.EventType.TextInput)
                 {
-                    int len = 0;
-                    string text = String.Empty;
-                    unsafe
+                    for (int i = 0; i < 32; i++)
```
tomspilman: What is `32` here?  isn't `ev.Text.Text.Length` better?

MonoGame.Framework/SDL/SDLGamePlatform.cs
nkast: Reuse of an EventArgs is problematic. After you fire an Event the eventArgs belongs to the user, he would not expect previous EventArgs to change on the fly.
This requires a change in design, using a delegate or registering an ITextInputObserver interface. 
The best we can for now is to skip the creation of TextInputEventArgs  if the user didn't subscribe to the event.

cra0zy: Yea, we get it, thats why Tom was suggesting making it a struct...

tomspilman: @nkast - i always want to see you involved.  You make good suggestions and have good ideas. :)

cra0zy: I try to use my casual speech when replying to stuff that is higher knowledge programming related, I mean no offense.

cra0zy: Now that I thought about the code, you should never try to keep EventArgs in the first place, since if anyone/anything else is subscribed to that event, they also get the exact same copy and can modify it.

tomspilman: I agree no one should.  But still feel the right fix is to convert `TextInputEventArgs` to a struct.

cra0zy: I don't like it, but I've done it.

MonoGame.Framework/GameWindow.cs
```diff
@@ -144,9 +144,10 @@ protected void OnScreenDeviceNameChanged ()
 		}
 
 #if WINDOWS || WINDOWS_UAP || DESKTOPGL || ANGLE
-		protected void OnTextInput(object sender, TextInputEventArgs e)
-		{
-            EventHelpers.Raise(this, TextInput, e);
+        internal void OnTextInput(object sender, TextInputEventArgs e)
+		{
+            if (TextInput != null)
```
gsfreema: Now that TextInputEventArgs is a struct, I can see not wanting to pass it to another method to raise it because that would be another unnecessary copy.  If you're going to raise events manually, the best practice is to store event in a local variable before checking for null and invoking it.  The reason is because it's possible (while highly unlikely) that another thread could be activated between the check for null and the actual invocation that will remove the last subscriber to the event.  This would lead to a hard to duplicate NullReferenceException.

If you're allowed to use C#6 language features, then the simple way around this is to do the following.
`TextInput?.Invoke(sender, e)`.

If C#6 is not an option, then using a local variable is the way to go.

Side note:  Nice catch here.  Looks like there was a hidden bug where the sender should have been passed to the EventHelpers.Raise method.

Issue #6305, opened 2018-04-27
I need help asap


staeri1:
Hello. I am new to programming and not very skilled. The problem I have faced is that when I was programming a game using visual studios and monogame i was supposed to add two .png to my monogame pipeline but somehow I accidentally added all items on my desktop. After i closed visual studios I noticed that all my files on my desktop where missing and that the game i was making was unable to start. I have tried looking for the files but I can’t seem to find them. Please help me find the missing files as they include most of my school work.

Thanks

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

Apostolique:
Did you delete the files from your content pipeline instead of using "exclude from project"?

mrhelmut:
Like Apostolique noted, if you used the "delete" function instead of "exclude from project", you just lost your files for good.

You might recover some of them with software such as Recuva.

Also, you'll find more help on the forum instead of here (which is dedicated to bug reports, not support): http://community.monogame.net/

Good luck with the recovery.

Apostolique:
This is one of the reason I find the content pipeline dangerous. I feel like delete shouldn't be an option at all.

tdeeb:
Furthermore, VS's delete function recognizes Recycle Bin settings and puts deleted files and folders into the Recycle Bin if specified. The Content Pipeline removes files regardless of your settings.

mrhelmut:
Yeah, it should definitely be able to put files in the recycle bin instead of deleting them.

cra0zy:
Delete option has been removed from the Pipeline Tool, so closing...

PR #6365, opened 2018-07-10
Make delete button red, and mention that items will be deleted from disk

RUSshy:


RUSshy:
I think adding a warning icon will be better

Apostolique:
tbh, the feature to delete content from disk should be removed. More discussion here: #6305.

cra0zy:
My linking PR is removing delete support.

RUSshy:
oh ok so i can close it!

Issue #4805, opened 2016-04-26
SpriteFont does not support emoji characters


dtaylorus:
The content pipeline tool fails with the following exception when a SpriteFont contains characters from the Unicode emoji character range (surrogate characters):

    Importer 'FontDescriptionImporter' had unexpected failure!
    System.FormatException: String must be exactly one character long.
       at System.Xml.XmlConvert.ToChar(String s)
       at Microsoft.Xna.Framework.Content.Pipeline.Serialization.Intermediate.CharSerializer.Deserialize(String[] inputs, Int32& index)

The failing SpriteFont file is attached (built on Windows).

[EmojiSpritefont.zip](https://github.com/mono/MonoGame/files/236848/EmojiSpritefont.zip)

tomspilman:
Interesting...  we certainly should support this if possible.  Just needs a little debugging I think.

KonajuGames:
The .NET Framework natively supports UTF-16, but the Char type is a 16-bit integer.  Therefore anything that requires a multi-character sequence, such the emoji character range, is not currently supported.  This includes the SpriteFont internal file format.  It could be done, but it gets messy and breaks compatibility.  Even using UTF-32, the emoji characters are above the legal range for a single UCS-4 character (0x0 - 0x10FFFF per ISO 10646).

FreeType supports unsigned 32-bit integer for the character code.  How to get it to support it properly I'm not sure.  Maybe see if it accepts character codes beyond the legal range.

tomspilman:
> Therefore anything that requires a multi-character sequence, such the emoji character range, is not currently supported.

Right... that is a bigger fix than I thought then.

Well the first thing is to maybe handle this more gracefully...  reporting a reasonable error like "We do not support multi-character glyphs"  or something?

KonajuGames:
@dtaylorus Where did you get those character ranges from?  They are above the Unicode 8.0 character range (up to 0x10FFFF).  The Unicode emoji page lists a range of character codes, ranging from simple one character codes to sequences containing up to seven or eight character codes (still within the range listed above).  These are not surrogate pairs in the normal sense however, as they are individual glyphs overlaid on each other.

KonajuGames:
http://unicode.org/emoji/charts/full-emoji-list.html

dtaylorus:
@KonajuGames The spritefont file contains decimal values for the coding range. Hex values are 0x1F300 - 0x1F5FF, which should be a valid character range.

KonajuGames:
[...]

PR #6366, opened 2018-07-11
Initial support for Emoji.

stromkos:
WIP. 
Allows all single codepoint Unicode to work within Spritefonts.

Does not break Xnb compatibility, forwards or backwards, when using "regular characters".

See issue #4805

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegion.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegion.cs
@@ -8,11 +8,11 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 	[TypeConverter(typeof(CharacterRegionTypeConverter))]
 	public struct CharacterRegion
 	{
-	    public char Start;
-	    public char End;
+	    public CharEx Start;
+	    public CharEx End;
 
 		// Enumerates all characters within the region.        
-	    public IEnumerable<Char> Characters()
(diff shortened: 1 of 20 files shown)


stromkos:
Obviously, I have no clue as to how I should add a new file to the project. If someone would please inform me of the correct procedure, the tests will pass. 
Thank You.

mrhelmut:
To add files to projects, you have to add them to the Protobuild project definitions here: https://github.com/MonoGame/MonoGame/tree/develop/Build/Projects

For instance, to add ```CharExWriter.cs``` to the content pipeline project, you have to add ```<Compile Include="Processors\CharExWriter.cs" />``` to this definition file: https://github.com/MonoGame/MonoGame/blob/develop/Build/Projects/MonoGame.Framework.Content.Pipeline.definition#L375

stromkos:
@mrhelmut, Thanks for the information.
Through unit testing, I was able to correct all of the bugs in CharEx.cs.
The sub-types continue to be an issue.  The existing system is not flexible enough to encode the proper sub-type in Lists, Collections, Arrays and Nullable types... Without hard-coding type replacements, which I am not sure would be the appropriate approach.  This change would exclude backward compatibility(new XNB files cannot be read by older versions) and should be slated as such for future version release(ex, 3.7), all future compatibility(i.e. previously generated files will always be accepted).
@tomspilman, Is this the correct approach?

tomspilman:
This is conflicted and debatable if we need this feature.

PR #6370, opened 2018-07-13
Fixing many events that are being raised with incorrect senders.

gsfreema:
I believe that these are all of the places where the incorrect `this` was being raised instead of `sender`.  The one exception is the `TextInput` event which will soon be fixed by #6361.

### DIFF
--- a/MonoGame.Framework/DrawableGameComponent.cs
+++ b/MonoGame.Framework/DrawableGameComponent.cs
@@ -68,12 +68,12 @@ namespace Microsoft.Xna.Framework
 
         protected virtual void OnVisibleChanged(object sender, EventArgs args)
         {
-            EventHelpers.Raise(this, VisibleChanged, args);
+            EventHelpers.Raise(sender, VisibleChanged, args);
         }
 
         protected virtual void OnDrawOrderChanged(object sender, EventArgs args)
         {
-            EventHelpers.Raise(this, DrawOrderChanged, args);
+            EventHelpers.Raise(sender, DrawOrderChanged, args);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


nkast:
I think all the `OnEvent(object , EventArgs )` should be changed to `OnEvent(object , EventArgs )`.
[Handling and Raising Events](https://docs.microsoft.com/en-us/dotnet/standard/events/index)

nkast:
I think all the `OnEvent(object , EventArgs )` methods should be changed to `OnEvent(EventArgs )`.
[Handling and Raising Events](https://docs.microsoft.com/en-us/dotnet/standard/events/index)

gsfreema:
These are all protected virtual methods that look to have included sender in the XNA Framework.  Changing them would be a breaking change and differ from XNA's API.

https://docs.microsoft.com/en-us/previous-versions/windows/xna/bb198314(v%3dxnagamestudio.41)

It is interesting how these `OnEvent` signatures differ from standard practices.  I'm not sure if they were intending for you to pass the object that caused something like the game to exit or the visibility of an object to change.

nkast:
That's strange! Anyway, it makes little difference.

gsfreema:
Maybe, but I can't imagine why Microsoft would have accepted a sender in their API and then ignore it.  It seems like that would quickly lead to issues being file by people overriding some of these classes, calling the `OnEvent` methods, and noticing that the sender passed in wasn't being used for the raised event.

Halofreak1990:
Odd that the Xna On**** functions don't follow the general .NET convention of such functions. On**** functions are provided to allow derived classes to raise events on the base class, if needed, or to provide custom handling before firing the event. On**** shouldn't require a `sender` argument, since the base class already provides `this` when invoking the event.

PR #6372, opened 2018-07-17
[DesktopGL] NormalizedByte2/4 texture format

nkast:
This PR removes the check for 'supportsNormalized' for
NormalizedByte2/4.

NormalizedByte2/4 are signed 8bits per component formats.
'supportsNormalized' queries the GL_EXT_texture_norm16 which is a later
16bits per component format.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsExtensions.cs
+++ b/MonoGame.Framework/Graphics/GraphicsExtensions.cs
@@ -654,16 +654,12 @@ namespace Microsoft.Xna.Framework.Graphics
                 break;
 
             case SurfaceFormat.NormalizedByte2:
-                if (!supportsNormalized)
-                    goto case InvalidFormat;
                 glInternalFormat = PixelInternalFormat.Rg8i;
                 glFormat = PixelFormat.Rg;
                 glType = PixelType.Byte;
                 break;
 
             case SurfaceFormat.NormalizedByte4:
-                if (!supportsNormalized)
-                    goto case InvalidFormat;
                 glInternalFormat = PixelInternalFormat.Rgba8i;
                 glFormat = PixelFormat.Rgba;
                 glType = PixelType.Byte;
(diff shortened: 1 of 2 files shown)


Jjagg:
Hmm, we maybe need to do another check for these formats since they're core in GL 3.0 (higher than what MG supports). Not sure if there's an extension for lower. Maybe we should just check if GL major version is >= 3?

I'll merge this for now since the current check is wrong anyways.

nkast:
Thanks @Jjagg 
I found this one: https://www.khronos.org/opengl/wiki/Image_Format#Texture_only
that lists formats which should always be supported by GL. 
NormalizedByte2/4 has to be the [ RGBA, RG | signed normalized | 8 ] in the table. 
Is that correct? 

That's only for textures, whether or not they are supported as rendertargets is another story. 
Currently MG convert them to Color, as did XNA. However that's not specified by a profile or feature level and might be different from card to card. If we can somehow test if they are supported as render targets we should do so.

nkast:
NormalizedByte texture still not working. Creating a texture with NormalizedByte4/2 throws an openGL error. It could be that we are not creating them with the right parameters. 
`Rgba8i` is supposed to be an integer format. For comparison FNA is using the RGBA8 value.
Same goes for RG32 (Rg16ui). We need to fix that and add visual tests to ensure that the values are mapped correctly.

https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Constants

https://github.com/FNA-XNA/FNA/blob/e79740b52daee92a532186cd67f527225756aaa3/src/FNAPlatform/ModernGLDevice.cs#L3573-L3574

Jjagg:
I checked for support of RGBA8i and RG8i because that's what is used in the code. But you're right that these are incorrect. Those formats are used for integer textures and those aren't supported in MonoGame. 

> NormalizedByte2/4 has to be the [ RGBA, RG | signed normalized | 8 ] in the table.
Is that correct?

Yes, that's correct. The wiki however assumes OpenGL 3.3 support however (as stated in the [style guide](https://www.khronos.org/opengl/wiki/Help:Style_Guide)), so the page listing the required supported texture formats is not fully applicable for us. Checking the specs for support for Rgba8 and Rg8 shows:
- Rgba8: Core in [OpenGL 2.1](https://www.khronos.org/registry/OpenGL/specs/gl/glspec21.pdf)
- Rg8: Core in OpenGL 3.0, [extension](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_rg.txt) for OpenGL 2.1

PR #6374, opened 2018-07-18
Added a garbageless GetGlyphs overload

tdeeb:
This commit adds a `GetGlyphs` overload that takes in a `Dictionary<char, Glyph>` and adds all glyphs in the SpriteFont to it. This lets you reuse a dictionary if desired.

Use-case: I have a dialogue system in [this](https://github.com/tdeeb/PaperMarioBattleSystem/blob/master/PaperMarioBattleSystem/PaperMarioBattleSystem/Classes/Dialogue/DialogueBubble.cs#L225) project that renders each character separately based on parsed text routines, which include adjusting the position, color, and scale of each character. When the font is changed, I have a usable dictionary that I can clear and repopulate with the new font's glyphs, but I am forced to discard it. This new overload allows you to keep an existing dictionary and still obtain the glyphs from the font.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteFont.cs
+++ b/MonoGame.Framework/Graphics/SpriteFont.cs
@@ -125,7 +125,8 @@ namespace Microsoft.Xna.Framework.Graphics
         public Dictionary<char, Glyph> GetGlyphs()
         {
             var glyphsDictionary = new Dictionary<char, Glyph>(_glyphs.Length, CharComparer.Default);
-            GetGlyphs(glyphsDictionary);
+            foreach (var glyph in _glyphs)
+                glyphsDictionary.Add(glyph.Character, glyph);
             return glyphsDictionary;
         }
 
--- a/MonoGame.Framework/Graphics/SpriteFont.cs
+++ b/MonoGame.Framework/Graphics/SpriteFont.cs
@@ -125,9 +125,18 @@ namespace Microsoft.Xna.Framework.Graphics
         public Dictionary<char, Glyph> GetGlyphs()
         {
             var glyphsDictionary = new Dictionary<char, Glyph>(_glyphs.Length, CharComparer.Default);
-            foreach(var glyph in _glyphs)
-                glyphsDictionary.Add(glyph.Character, glyph);
+            GetGlyphs(glyphsDictionary);
             return glyphsDictionary;
+        }
+
+        /// <summary>
+        /// Adds the glyphs in this SpriteFont to an existing dictionary.
+        /// </summary>
+        /// <param name="glyphDict">An existing dictionary with a key of type char and a value of type Glyph.</param>
+        public void GetGlyphs(Dictionary<char, Glyph> glyphDict)
+        {
+            foreach (var glyph in _glyphs)
+                glyphDict.Add(glyph.Character, glyph);
         }
 
 		/// <summary>


Jjagg:
Pretty narrow usage for this overload I think, but I'm fine with this change.

cc @tomspilman @cra0zy @dellis1972 This adds public API.

stromkos:
I would suggest adding this method in your user code as:

    public void GetGlyphs(Dictionary<char, Glyph> glyphDict, SpriteFont sf)
        {
            foreach (var glyph in sf.Glyphs)
                glyphDict.Add(glyph.Character, glyph);
         }

And not modify SpriteFont.cs at all. 


Additional Note:
Lines 128(129) should be left alone.  Adding overhead by calling method that contains a single statement that is unlikely to change would be a waste of resources.

Jjagg:
@stromkos The whole point of this change is to prevent generating garbage. The existing method creates a new dictionary and there's currently no way to get the glyphs without that allocation.

tdeeb:
It looks like [this PR](https://github.com/MonoGame/MonoGame/pull/6126) merged back in February added the public `Glyphs` property, which can be used to do what @stromkos suggested.

Jjagg:
You're right, I misread! I think that property slipped through in the PR you linked, because we probably don't want it to be possible to modify the glyphs of a `SpriteFont`.

tdeeb:
Since that is the case, there are now two ways of getting the glyphs in a `SpriteFont`:
1. The `GetGlyphs()` method, which returns a new `Dictionary<char, Glyph>`
2. The `Glyphs` property, which returns a reference to the internal `Glyph[]` array.

Which way should we go with?

stromkos:
@Jjagg, 
Does it matter if someone decides to change to Glyph array? That would be their mistake(or intentional).  The size of the array cannot change, guaranteeing errors are user induced and will be reported in user code, not in the framework.  During tests that I carried out, duplicating glyphs in the array did not throw an error(it should have given the GetGlyphs() method returns a dictionary(which cannot have duplicate keys??))
Exposing the Glyph array has more benefits, see OP, than drawbacks.


@tdeeb, 
Please wait for a response from Jjagg.
If he agrees, please close this PR, and follow the code in my original post.
 
I have an issue with your micro-optimizations, since both operations(clear and new) act on the heap, the only savings is in the pointer on the stack. Stack operations are very fast compared to heap allocations, but both methods require heap allocation, UNLESS the sizes are the same or less, otherwise it is faster to allocate a new buffer. 
Unless you are on a memory constrained platform, the existing methods will serve you well.

tdeeb:
@stromkos I'm unsure what your last paragraph is referring to. Is it the array, the dictionary, my code or this PR?

Jjagg:
Really the only reason glyphs are exposed as a dictionary is because it's XNA API and we supported it previously so we should keep supporting it. But the internal implementation uses an array since #5874, so `GetGlyphs` doesn't really make sense anymore. IMO we should drop `GetGlyphs` in 4.0 (when we break XNA compatibility). 

My suggestion is to wrap the glyphs array with `Array.AsReadOnly` and expose that instead of the glyphs array directly (through the `Glyphs` property). That way it's obvious you shouldn't change the array while providing all the benefits of exposing the array directly.

Tagging @tomspilman @cra0zy @dellis1972 again because of public API change.

stromkos:
@tdeeb, 
The post questions your reasoning behind the optimization in the first place.  I am not discounting it as to its validity, only questioning the motivation behind it. I am a firm believer in combating [Wirth's law
](https://en.wikipedia.org/wiki/Wirth%27s_law), but the existing GetGlyphs() method, should give you all of the information you require(at an O(N) cost(2N when your are done + 2n heap allocations)),  but my comment was directed at the need(or lack thereof) for such a solution.

tdeeb:
@stromkos I admittingly gave a poor example usage; my primary concern was being able to obtain the glyphs without generating garbage, which I feel is very useful to have, as it gives more control to the developers in managing their memory. It was my bad on submitting an example that was not using an up-to-date version of MonoGame with the publicly accessible `Glyphs` array. Even without the array, games heavy on text can use other techniques, such as caching the dictionaries for each font.

stromkos:
@Jjagg 
The GetGlyphs() is used in SpriteBatch.Draw() methods.  It makes sense(until you consider multi-character emojis, but that is beside the point) to leave the existing API as is.
What is the objection to a read-only public array(where each item can be modified) glyphs?    I have used this in my emoji font run-time Android demo.
Without this access I would be restricted to reflection/CECIL(SLOW).

Jjagg:
> The GetGlyphs() is used in SpriteBatch.Draw() methods. It makes sense(until you consider multi-character emojis, but that is beside the point) to leave the existing API as is.

`GetGlyphs()` is not used in `SpriteBatch.Draw()`. Before the PR I linked we did use a Dictionary, but in that PR we switched to binary searching the containing 'region' of subsequent characters and computing the index in the `_glyphs` array from the character and the starting index of the region in the `_glyphs` array. That's faster because often you only have one region except for some localizations or special characters (the latin character set). 
Also, ligatures and emoji sequences should be parsed in the step before finding the actual glyphs, so that's not relevant here. In this step we only care about getting the glyph given the UTF-32 character.

> What is the objection to a read-only public array(where each item can be modified) glyphs? I have used this in my emoji font run-time Android demo.

The objection is that its mutable. 
Since the array is only as large as the number of glyphs in the SpriteFont, I assume you have to copy them to a new array anyway if you want to add glyphs. And MonoGame will support emoji out of the box in the future. IMO there's really no benefit to exposing the mutable array and it might trip up less experienced programmers or devs expecting we return a copy (which is what GetGlyphs did when we still used a Dictionary internally).

stromkos:
@tdeeb, 
Sorry, my previous comment was cut short.  Either way your collection **may** suffer from garbage collection/reallocation. 
My point was that the number of spritefonts, versus the memory/cpu used for each does not warrant a new method in the framework.

stromkos:
@Jjagg,
I understand the array is mutable.
There is no adding of glyphs. 
    -- IMO there's really no benefit to exposing the mutable array and it might trip up less experienced 
    programmers or devs expecting we return a copy.
While I agree with this sentiment, if the less experienced programmer or dev? What harm could they cause?  
Beyond proper coding techniques, I do not see the harm.

stromkos:
This pr should be closed.  As it stands the requirements are fulfilled  by the existing code.

Jjagg:
@LithiumToast Like I said in my previous comment, IMO we should expose the glyphs array that's used internally as a read only collection. If you - as a user - want a dictionary you can build it yourself. Since the internal representation was changed to an array it's impossible to get a dictionary without creating one (be it in MG or user code).

Jjagg:
Superseded by #6579.

Issue #6386, opened 2018-07-26
Crash  at startup after failing to load OpenGL functions on Windows


james0x0A:
Since the method of loading GL functions on windows was changed to use the mechanism in kernel32 via FuncLoader (#6376),  not all GL functions are properly loaded. This causes the graphics device to throw creation.

OpenGL proscribes using wglGetProcAddress from the OpenGL library. 

[OpenGL Documentation](https://www.khronos.org/opengl/wiki/Load_OpenGL_Functions)
[SDL Implementation](https://www.libsdl.org/tmp/SDL/src/video/windows/SDL_windowsopengl.c)

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:
- DesktopGL

tomspilman:
@cra0zy ?

mrhelmut:
Do you happen to know which functions?
I haven't been encountering any issue so far, but I'm not using the full scope.

cra0zy:
I already know where the problem lies, will have a PR fix for this ready later today.

cra0zy:
PR up, if someone has the time to test it that would be awesome.

PR #6387, opened 2018-07-27
[DesktopGL] Use wglGetProcAddress on Windows for loading OpenGL functions

cra0zy:
I'll have time to test this out later, however if one of you has the time, comment the results bellow.

Fixes: #6386

### DIFF
--- a/MonoGame.Framework/SDL/SDL2.cs
+++ b/MonoGame.Framework/SDL/SDL2.cs
@@ -591,6 +591,10 @@ internal static class Sdl
         }
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+        public delegate IntPtr d_sdl_gl_getprocaddress(string proc);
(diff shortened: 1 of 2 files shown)


james0x0A:
Looks like this doesn't quite work.  I think a call to `wglMakeCurrent` may be necessary after the context is created so that the appropriate extensions are enabled when when wglGetProcAddress is called.

james0x0A:
I've done some testing and `wglMakeCurrent` is not absolutely necessary to get running, so I am not sure why SDL calls it that way. The problem seems to be the way wglGetProcAddress  is loaded.  Using `Windows.GetProcAddress` and marshalling a delegate works.

cra0zy:
Yea, I found something fishy about what SDL is doing as well, for DesktopGL under Windows we'll have the `SDL_GetProcAddress` to load up the opengl functions.

Loading OpenGL on Unix platforms is so much easier.

cra0zy:
This has been moved to: #6389

PR #6389, opened 2018-07-28
Fixup and improve FuncLoader code

cra0zy:


MonoGame.Framework/Graphics/OpenGL.MacOS.cs
```diff
@@ -7,16 +7,23 @@
...
 using AppKit;
 using System.Security;
+using MonoGame.Utilities;
 
 namespace OpenGL
 {
     public partial class GL
-	{
+    {
+        public static IntPtr Library = FuncLoader.LoadLibrary("/System/Library/Frameworks/OpenGLES.framework/OpenGLES");
```
dellis1972: Should we only be using OpenGLES on Desktop Mac?

dellis1972: I think this is a cut and paste typo.. it should be "/System/Library/Frameworks/OpenGLES.framework/OpenGL".

cra0zy: https://github.com/MonoGame/MonoGame/pull/6390

PR #6392, opened 2018-07-29
[DesktopGL] Add Window.FileDropped event

harry-cpp:
Adds an event for when file is dropped on the window.

### DIFF
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -88,6 +88,7 @@ namespace Microsoft.Xna.Framework {
 		public event EventHandler<EventArgs> ClientSizeChanged;
 		public event EventHandler<EventArgs> OrientationChanged;
 		public event EventHandler<EventArgs> ScreenDeviceNameChanged;
+        public event EventHandler<string> FileDropped;
 
 #if WINDOWS || WINDOWS_UAP || DESKTOPGL|| ANGLE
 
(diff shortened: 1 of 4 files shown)


dellis1972:
@cra0zy I think tom will probably want a bit more detail of what this does and why its needed in the PR description :)

cra0zy:
I thought it was self explanatory since it was Window.FileDropped, anyway added description.

tomspilman:
I see `GameWindow` as something designed for games.  This to me does not fit that purpose.

I especially don't like when features like this are only supported on one platform.  To even consider this we need it supported on UWP and WindowsDX.

Finally if we are going to add something like this it needs documentation.

tomspilman:
> honestly us support it is not that big of a deal.

We have so many half implemented things already i just don't want to add to them.

It makes MG unreliable to users.

> Gonna add `#if DESKTOPGL`

That wasn't my point.  My point was we need a plan to implement this on UWP and WinDX.  If there is no plan to do it then i don't want the feature.

cra0zy:
> That wasn't my point. My point was we need a plan to implement this on UWP and WinDX. If there is no plan to do it then i don't want the feature.

Fine fine, I'll do it.

Man I hate implementing stuff on Windows.

tomspilman:
> Man I hate implementing stuff on Windows.

I know... i hate implementing things on Mac/Linux/iOS/Android, but we gotta do what we gotta do. :)

jessefreeman:
This was a huge help for my fantasy console. Hope it makes it into the main branch!

tomspilman:
@harry-cpp  do you want to drop this or fix this?

harry-cpp:
I'll fix it up.

SimonDarksideJ:
@harry-cpp were you able to fix up this PR, or can it be closed in the modern world

Issue #5126, opened 2016-08-21
Content Statistics


tomspilman:
It would be very useful for the `MGCB` tool to gather content statistics.  The statistics would be generated when content building occurs and would be persisted to an XML file.

Example of statistics we could gather would be:

    - Number of models in the project.
    - MBs of compressed `SoundEffect` s.
    - Percentage of uncompressed vs compressed `Texture2D`s.
    - The disk space required for all content.
    - The approximate memory usage in game of all content.
    - The amount of textures in a particular folder.

These statistics then could be parsed and presented in the Pipeline tool UI in some useful manner.

Ideally most of the information could be gathered from the builder without the help of individual processors.  Types of content, what folders they exist in, number of files, disk space required, etc can be all gathered from the outside.  Things like if the content is compressed or not or the runtime memory usage would require either some cooperation from the processor to work efficiently.  We could easily add additional methods to the `ContentProcessorContext` or the base `ContentItem` to help record the necessary information.

This feature would make diagnosing content size issues easier and more accurate than the guessing and runtime measurements we have to do now.

Ragath:
> The approximate memory usage in game of all content.
The only way to do this somewhat accurately would be to either have the ContentReader expose the information or write the tooling separately and associate it with the ContentReaders available. This does pose another issue though, because there could be multiple candidates for the ContentReader used to load a specific asset depending on how they and the writer get implemented.

tomspilman:
So working on this a bit tonight.

What i have so far:

 - Write a `.mgstats` file in the content intermediate directory.
 - We persist this file between builds so that you always have the correct stats.
 - The stats file is a simple CSV format data dump.
 - We store build times in addition to file sizes and in game memory estimates.
 - We will eventually build some sort of visualizer into the Pipeline tool.

tomspilman:
@cra0zy - I'll likely leave the visualization work to you on this.

I think the right thing to do here would be have a new "Stats" button in the Pipeline tool that brings up a window with various views on the stat data.   Letting you sort the data in various ways.  See per-type info like "how much space are all textures taking" or "how many non-compressed textures do we have and how much space they take".  Being able to sort all textures by disk size or build times.  Things like that.

For now i'm visualizing it in an Excel for my own needs.

tomspilman:
Interesting data so far...
![image](https://user-images.githubusercontent.com/103522/43621257-21e6fd1e-969c-11e8-8ead-66928dfbba80.png)
[...]

Issue #6243, opened 2018-03-11
Improve Pipeline Tool Filter View


tomspilman:
Some suggestions to improve the filter view of the Pipeline Tool.

Remove drop down arrows on "Cleaning" and "Build" lines that do not contain useful info.  All of these lines:

![image](https://user-images.githubusercontent.com/103522/37252319-48d5bc60-24ed-11e8-8688-b0f23bd4a296.png)

And all of these lines:

![image](https://user-images.githubusercontent.com/103522/37252359-d95a6150-24ed-11e8-8d1f-d40de62b6448.png)

Contain useless info in them and we should not show expansion arrows on them.  All those arrows with useless info hidden under it suggest to the user that all the info is useless.  Instead if we only show the arrow when there is useful info... like error messages or warnings... then the arrow becomes a feature and not "visual noise".

The logic could be as simple as if the content didn't generate an error and there is only one captured line of output for the item... then don't render the arrow or allow expansion.

Second it seems we're losing some info in the filter view:

![image](https://user-images.githubusercontent.com/103522/37252394-5bc972fc-24ee-11e8-9994-fb029686ba0d.png)

![image](https://user-images.githubusercontent.com/103522/37252406-85d990c2-24ee-11e8-9af2-5eec041867f8.png)

Here the warning message was lost in the filter view and not visible to the user.  We really need to capture these lines and show them else this view becomes irrelevant as it hides useful information.

Finally we also need an "Expand All" "Colapse All" feature for this view.  Could just be added via a new context menu when right clicking on the view.  Could come in useful for other things in the future too.

cra0zy:
I would say this is connected to: https://github.com/MonoGame/MonoGame/issues/5804

We don't need just to get better info from the Pipeline Tool, but also from the MGCB. I really like what @nkast did with https://github.com/MonoGame/MonoGame/pull/5163

tomspilman:
> We don't need just to get better info from the Pipeline Tool, but also from the MGCB

That would be nice... but we cannot assume perfect formatting from MGCB.

We often depend on 3rd-party tools that have their own ways to format errors.  Us trying to "fix" the error/warning output of all shader compilers, texture compressors, etc to fit our own rules is just going to cause more errors.

I think we have to embrace the fact all we can control a few lines that MGCB outputs:

 1. The "Clean" line.
 2. The "Skipping" line.
 3. The "Build" line.
 4. The "Error" line.

[...]

PR #6401, opened 2018-08-04
MGCB now generates content building statistics.

tomspilman:
This PR adds the feature where `MGCB.exe` will generate a CSV file `.mgstats` to the intermediate folder during content building.  This stats file contains per-content item information which can be used to diagnose a ton of different content issues.

Currently we record the following stats:

  - Source File
  - Dest File
  - Processor Type
   - Content Type
   - Source File Size
   - Dest File Size
   - Build Seconds

We can expand this with more information in the future if needed.

The implementation takes care to cache the stats and incrementally update them on rebuilds of selected content.

This is just the first part of this feature.  It will be followed up with a stand alone viewer that can be launched from the Pipeline tool to analyze and visualize the generated stats in a more useful format.

See https://github.com/MonoGame/MonoGame/issues/5126.

### DIFF
--- a/Tools/MGCB/BuildContent.cs
+++ b/Tools/MGCB/BuildContent.cs
@@ -308,7 +308,7 @@ namespace MGCB
             var previousContent = SourceFileCollection.Read(contentFile);
 
             // If the target changed in any way then we need to force
-            // a fuull rebuild even under incremental builds.
+            // a full rebuild even under incremental builds.
             var targetChanged = previousContent.Config != Config ||
                                 previousContent.Platform != Platform ||
                                 previousContent.Profile != Profile;
@@ -331,6 +331,8 @@ namespace MGCB
                 if (cleanRebuiltContent || cleanOldContent || targetChanged)
                     _manager.CleanContent(sourceFile, destFile);                
             }
+
+            // TODO: Should we be cleaning copy items?  I think maybe we should.
 
             var newContent = new SourceFileCollection
             {
@@ -403,7 +405,10 @@ namespace MGCB
             // If this is an incremental build we merge the list
             // of previous content with the new list.
             if (Incremental && !targetChanged)
+            {
                 newContent.Merge(previousContent);
+                _manager.ContentStats.MergePreviousStats();
+            }
 
             // Delete the old file and write the new content 
             // list if we have any to serialize.
@@ -430,16 +435,24 @@ namespace MGCB
                     // Only copy if the source file is newer than the destination.
                     // We may want to provide an option for overriding this, but for
                     // nearly all cases this is the desired behavior.
-                    if (File.Exists(dest))
+                    if (File.Exists(dest) && !Rebuild)
                     {
                         var srcTime = File.GetLastWriteTimeUtc(c.SourceFile);
                         var dstTime = File.GetLastWriteTimeUtc(dest);
                         if (srcTime <= dstTime)
                         {
-                            Console.WriteLine("Skipping {0}", c.SourceFile);
+                            if (string.IsNullOrEmpty(c.Link))
+                                Console.WriteLine("Skipping {0}", c.SourceFile);
+                            else
+                                Console.WriteLine("Skipping {0} => {1}", c.SourceFile, c.Link);
+
+                            // Copy the stats from the previous stats collection.
+                            _manager.ContentStats.CopyPreviousStats(c.SourceFile);
                             continue;
                         }
                     }
+
+                    var startTime = DateTime.UtcNow;
 
                     // Create the destination directory if it doesn't already exist.
                     var destPath = Path.GetDirectoryName(dest);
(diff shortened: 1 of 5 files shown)


Tools/MGCB/BuildContent.cs
```diff
@@ -332,6 +332,8 @@ public void Build(out int successCount, out int errorCount)
                     _manager.CleanContent(sourceFile, destFile);                
             }
 
+            // TODO: Should we be cleaning copy items?  I think maybe we should.
```
tomspilman: @cra0zy @nkast @Jjagg - I noticed we haven't been cleaning copy items.  It seems to me like we should.

Should i go implement this in a new PR?

Tools/MGCB/BuildContent.cs
```diff
@@ -430,17 +435,25 @@ public void Build(out int successCount, out int errorCount)
                     // Only copy if the source file is newer than the destination.
                     // We may want to provide an option for overriding this, but for
                     // nearly all cases this is the desired behavior.
-                    if (File.Exists(dest))
+                    if (File.Exists(dest) && !Rebuild)
```
tomspilman: @cra0zy @nkast @Jjagg 

It made sense to me to make a "Rebuild" command always re-copy the content for copy items.

Agree?

Tools/MGCB/BuildContent.cs
```diff
@@ -430,17 +435,25 @@ public void Build(out int successCount, out int errorCount)
...
                     {
                         var srcTime = File.GetLastWriteTimeUtc(c.SourceFile);
                         var dstTime = File.GetLastWriteTimeUtc(dest);
                         if (srcTime <= dstTime)
                         {
-                            Console.WriteLine("Skipping {0}", c.SourceFile);
+                            if (string.IsNullOrEmpty(c.Link))
+                                Console.WriteLine("Skipping {0}", c.SourceFile);
+                            else
+                                Console.WriteLine("Skipping {0} => {1}", c.SourceFile, c.Link);
```
tomspilman: @cra0zy @nkast @Jjagg 

Since the change to support copy items with a link...  it seemed appropriate to show this by showing the link.  How i have it here the logging looks like:
```
C:\Souce\File\Input\File.png => Dest\Folder\File.png
```
The `=>` seemed like good shorthand to mean a linked copy.

Thoughts?

Tools/MGCB/BuildContent.cs
```diff
@@ -453,6 +466,16 @@ public void Build(out int successCount, out int errorCount)
                     fileAttr = fileAttr & (~FileAttributes.ReadOnly);
                     File.SetAttributes(dest, fileAttr);
 
+                    var buildTime = DateTime.UtcNow - startTime;
+
+                    if (string.IsNullOrEmpty(c.Link))
+                        Console.WriteLine("{0}", c.SourceFile);
+                    else
+                        Console.WriteLine("{0} => {1}", c.SourceFile, c.Link);
```
tomspilman: @cra0zy @nkast @Jjagg

We were never logging copy items which meant they were invisible to the build process before.

I used the same logging as built items... but maybe we should log these differently?
```
Copy C:\Souce\File\Input\File.png
Copy C:\Souce\File\Input\Other.png =>Dest\Folder\Other.png
```
I can even fix the output view to show them differently if this makes sense.

Tools/MGCB/BuildContent.cs
nkast: Indeed. Furthermore when you deleted then by hand and rebuild it will not output the filenames of the copied items.

nkast:
just an observation...
ContentStats is very similar to the data saved in `.mgcontent`.

tomspilman:
@nkast - I considered that, but felt it was better separate from the .mgcontent stuff.  Different needs and may contain much different data in the future.

Tools/MGCB/BuildContent.cs
tomspilman: I'll implement "Clean" for copy items in a new PR then.

cra0zy:
I'll go over the changes later today/tomorrow.

Tools/MGCB/BuildContent.cs
Jjagg: Agreed, nice idea.

Tools/MGCB/BuildContent.cs
Jjagg: Also agreed :) 
The Pipeline Tool needs to support parsing this and the link output, I guess it will cause a crash in this branch (not sure how the parsing is implemented).

cra0zy: Hmm, for now I would recommend skipping fixing the Pipeline Tool/MGCB output, and instead we should go over MGCB with proper discussion: https://github.com/MonoGame/MonoGame/issues/6243

tomspilman: >  I guess it will cause a crash in this branch

@Jjagg - It doesn't.  It appears like a normal "Building" line.

> for now I would recommend skipping fixing the Pipeline Tool/MGCB output

@cra0zy - Sure...  i'm all up for a bigger discussion on the issue.  

For now it is better that we output something instead of nothing, so i would prefer to leave what i have in this PR until we decide on something better.

cra0zy: > For now it is better that we output something instead of nothing, so i would prefer to leave what i have in this PR until we decide on something better.

Fix the Pipeline Tools output parser than at least.

tomspilman: > Fix the Pipeline Tools output parser than at least.

Fix it to do what?  It doesn't crash or anything... it seems ok as it is.

cra0zy: Thats fine if it works, but I thought it would not resolve relative paths or not show the linked path or something like that.

tomspilman: > I thought it would not resolve relative paths or not show the linked path or something like that.

It works.  Just says "Building XXXX => XXXX" which isn't technically correct... but works for now.

PR #6402, opened 2018-08-04
Restore EnsureExecutable

cra0zy:
Removed it because it was crashing in case of not having permissions to modify the executable, this is no longer happening and we are gonna need this if we ever wanna package up tools inside nugets (nuget packages don't preserve permissions for files on Unix platforms).

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
+++ b/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
@@ -56,6 +56,8 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
                 RedirectStandardInput = true,
             };
 
+            EnsureExecutable(fullPath);
+
             using (var process = new Process())
             {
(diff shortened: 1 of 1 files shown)


Jjagg:
We could also use [.NET Core Global Tools](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools) to deploy our tools through NuGet.

cra0zy:
> We could also use .NET Core Global Tools to deploy our tools through NuGet.

Interesting, tho this does not change the way we should distribute MGCB, we still want MGCB to be tied to the NuGet itself so that the project is self contained when it comes to compilation.

PS. This change is just me reverting my previous commit, so you can merge it.

cra0zy:
PPS. That does solve Pipeline Tool distribution so there are nice positives with that, we should be able to get installers to be really bone back thanks to it.

Jjagg:
Looks good, merging :)

PR #6403, opened 2018-08-04
Bump tools to .NET Framework 4.7.1

cra0zy:
This is needed so developers can write content extensions targeting .NET Standard.


Build/Projects/MonoGame.Framework.Content.Pipeline.definition
```diff
@@ -45,6 +45,10 @@
       <Platform Name="Windows">TRACE;WINDOWS</Platform>
     </CustomDefinitions>
 
+    <FrameworkVersions>
+      <Version>v4.7.1</Version>
```
nkast: Do we need to upgrade Framework.Content.Pipeline too?
This will require all existing third party importers to upgrade.

Will this change affect the generated PCL or it will still be 4.5?

cra0zy: > Do we need to upgrade Framework.Content.Pipeline too?

Yes, thats literally the one thing that needs to be upgraded for this to work, all the other changes are because of it.

> This will require all existing third party importers to upgrade.

No, it will not, they should be linking against the portable package which means they will now be able to anything from .NET 4.5 to PCL to .NET Standard for their content extensions.

> Will this change affect the generated PCL or it will still be 4.5?

It's a PCL, it does not have a target framework (well has a target PCL bit that does not tell you much), this does not change it.

nkast: > Yes, thats literally the one thing that needs to be upgraded for this to work, all the other changes are because of it.

I don't think that's true. The  Content.Pipeline is the last in the hierarchy of dlls, but anyways, 
if you build against the PCL this won't matter.

cra0zy: > I don't think that's true. The Content.Pipeline is the last in the hierarchy of dlls, but anyways,
> if you build against the PCL this won't matter.

Are you sure you are able to load up an assembly that uses lower version of .NET Framework that loads up an external assembly which uses a higher version of .NET Framework in case the main app is using the higher version?

Anyway, even if you can, we need to bump its version because anybody making tools for it needs to know what version of .NET Framework they should target.

nkast: Yes, an .EXE that runs on 4.7.1 can load all libraries that were build for the a previous version. Each version of the Dot.Net framework is a superset of the previous version.

cra0zy: I do understand that, however in this case we are talking about using reflection with a library which is compiled using older .NET Framework to access one which was compiled using newer .NET Framework, but now that I think about it, you are probably right.

On a related note, there are some exceptions to the "can load all libraries that were build for a previous version" like the `RuntimeInformation.IsOSPlatform` API, but these are rare.

vpenades: @cra0zy even if the library has been compiled with an older net framework, that old library is running with netfx471, so the reflection calls are fine. The NET version of an assembly is used only for validation, everything ends running with the framework defined by the EXE.

vpenades: What's blocking this PR?  I thougth it was going to be a dead easy PR...

cra0zy: Mac build bot is the problem, it needs updating.

tomspilman: I'll try to do this today.  It just needs the latest stuff i guess?

cra0zy: Yea.

cra0zy:
Protobuild removal is in progress, and this will be resolved during this month.

Jjagg:
CP and tools are upgraded in #6824.

PR #6270, opened 2018-03-24
Update iOS/tvOS to use OpenAL Soft [WIP]

dellis1972:
Needs [1]. 
Will no doubt conflict with https://github.com/MonoGame/MonoGame/pull/6257 since it needs`Build/GenerateProject.CSharp.xslt` changes.

[1] https://github.com/MonoGame/MonoGame.Dependencies/pull/124

PR #6407, opened 2018-08-06
Update iOS/tvOS to use OpenAL Soft

tomspilman:
This is an updated version of https://github.com/MonoGame/MonoGame/pull/6270.


tomspilman:
Still testing this to be sure it works... will report back shortly.

PR #6418, opened 2018-08-24
[WinDX] FIX: Implement missing X1 & X2 mouse buttons

MichaelDePiazzi:
This is a nice easy one - In the WinDX build, the X1 and X2 mouse buttons weren't being set.


cra0zy:
Thanks @MichaelDePiazzi , merging!

Issue #6424, opened 2018-08-30
16bit per channel PNG images are incorrectly imported by the content-pipeline tool


roy-t:
I believe PNG images that have 16 bits per channel are incorrectly processed by the content-pipeline tool. 

I've created an extremely minimal reproduction here: https://github.com/roy-t/monogame_16bitpp_png_bug

Below you see two textures rendered with the spritebatch, with default settings. The left sprite is a texture generated from an 8 bit per pixel PNG file, and is displayed correctly. The right sprite is a texture generated from the same PNG file, this time stored with 16 bits per pixel. As you can see the colors are completely different.

![bug](https://user-images.githubusercontent.com/623634/44863637-f371a180-ac7d-11e8-8532-a2b4835fd206.png)

#### What version of MonoGame does the bug occur on:
- MonoGame 3.6.0.1625

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:
- Windows, DirectX

roy-t:
I think the bug is in the code below.

Although the code does check the bytes per pixel to see how far to jump to get to the next color channel. It assumes a single color is one byte. If I'd patch this code to down-sample from however many bits to a single byte, would that be accept as a PR?

https://github.com/MonoGame/MonoGame/blob/6f34eb393aa0ac005888d74c5c4c6ab5615fdc8c/MonoGame.Framework/Utilities/Png/PngReader.cs#L245
```
{
    switch (colorType)
    {
        // snip
        case ColorType.Rgb:
            for (int x = 0; x < width; x++)
            {
                int offset = 1 + (x * bytesPerPixel);
                
                int red = defilteredScanline[offset];
                int green = defilteredScanline[offset + bytesPerSample];
                int blue = defilteredScanline[offset + 2 * bytesPerSample];

                data[(y * width) + x] = new Color(red, green, blue);
            }

            break;
    // snip
    }
}

```

tomspilman:
Thanks for the detailed report @roy-t !

I don't know this code as well... we would totally accept a PR to fix it along with a unit test to verify the fix.

roy-t:
I first found the PngReader by searching for PNG, but not it seems that, at least on Windows, its not referenced. Using the unit tests I came to this file:

https://github.com/MonoGame/MonoGame/blob/6f34eb393aa0ac005888d74c5c4c6ab5615fdc8c/MonoGame.Framework.Content.Pipeline/TextureImporter.cs#L139

[...]

PR #6425, opened 2018-09-02
Fix 6424, do not swap color channels for BGR image formats

roy-t:
This fixes issue https://github.com/MonoGame/MonoGame/issues/6424

I've changed the importing of both `FIT_RGB16` and `FIT_RGBF` images to respect that in this case the red and blue color channels are already in the correct order. 

I've based the correct order on the FreeImage source code, extensive tests with PNG images, and the fact that `FIT_RGBA16` and `FIT_RGBAF` images were already imported without the color channel swapping.

This PR includes tests for 24, 32, 48, and 64 bit PNG images to verify that both BGR and RGB images are imported correctly.

Unfortunately I couldn't add a test case for 96 and 128 bit floating point images. It seems FreeImage.net is not able to read or save 96/128bit TIFFs and PPM. It could also not load these images when they are created with ImageMagick or ShaderMap4. It seems this is a know issue. (See the comments in the `ImportImageWithBadContent` test).

_By submitting this PR I agree to distribute this work under the MonoGame license and copyright._


Jjagg:
This is great @roy-t! Thanks! It seems the build not didn't run, but if it passes this is good to go :) 

@tomspilman Seems like build bot issues.

CartBlanche:
Love it!! It has been a while since I've looked at an MG PR. Great to see one of this quality.

roy-t:
@Jjag anything I can do to trigger a build? 

Thanks for the kind words everyone 😁

Jjagg:
@roy-t Nope, nothing you can do! Build bots are down apparently.

Jjagg:
Thanks @roy-t! Merging :)

PR #6429, opened 2018-09-14
Fixed SoundEffectInstance.IsLooped XNA correctness.

tomspilman:
Adding some unit tests to verify that `SoundEffectInstance.IsLooped ` works like XNA.

### DIFF
--- a/Test/Framework/Audio/SoundEffectTest.cs
+++ b/Test/Framework/Audio/SoundEffectTest.cs
@@ -9,6 +9,7 @@ using Microsoft.Xna.Framework.Audio;
 using NUnit.Framework;
 using Microsoft.Xna.Framework.Content;
 using Microsoft.Xna.Framework.Graphics;
+using System.Threading;
 
 namespace MonoGame.Tests.Audio
 {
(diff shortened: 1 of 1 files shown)


tomspilman:
FYI.  This is going to fail... will fix the MG bugs next.

Jjagg:
I guess it could be required in case you play a soundeffect and only later know it should loop based on some game state. Or maybe more likely a sound effect that loops until some condition is true, but you don't want it to stop playing in the middle of a loop. 

I don't see why it should be forbidden to change loop state after playing apart from mimicking XNA behavior. In terms of XNA -> MG compatibility it won't matter either if we support changing loop (because XNA cose can't change loop after playing, so nothing changes) and MG -> XNA compatibility hasn't been a priority in the past.

So basically, not allowing the loop change is an unnecessary restriction that might prevent some people from easily getting the behavior they want and current behavior doesn't cause XNA -> MG compatibility issues. So IMO we should not follow XNA here.

Halofreak1990:
@Jjagg, if you really need to change the loop state, is it _that_ much of a hassle to simply create a new SoundEffectInstance and discard the old one?

Ragath:
I don't see an issue with the OpenAL platform being able of doing more, but I think XAudio should then be failing with an exception like XNA instead of failing silently.

tomspilman:
> I don't see an issue with the OpenAL platform being able of doing more

The problem is when you then port that game to a platform that does less and you get a crash and have to scramble to find a work around.  People don't like those differences when porting.

This is why it is best to keep the functionality the same across platforms as much as possible.  It is fine to add platform specific features... but it has to be something very clearly platform specific and something a dev has to explicitly opt-in to.

I still feel my plan is the best way forward.

Issue #6255, opened 2018-03-16
TooHighMultiSampleCountClampedToMaxSupported fails on some machines


tomspilman:
I find on some of my machines the unit test `TooHighMultiSampleCountClampedToMaxSupported ` fails:

https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/Graphics/GraphicsDeviceManagerTest.cs#L516

It came from https://github.com/MonoGame/MonoGame/pull/5477 by @Jure-BB,

Specifically i get errors on this line:

```
            // Reference device supports 32 samples
            Assert.AreEqual(gdm.GraphicsDevice.PresentationParameters.MultiSampleCount, 32);
```
(note the assert also should be fixed as `expected` and `actual` parameters are reversed)

In my case my reference device gets 16 and not 32.

So either:

 a) we're not setting the reference device properly.
 b) reference device can have different max multisample counts on different systems.
 c) some other weird bug.

Someone needs to do a little research on this and fix this test.

Assuming A is not true... my guess is fixing it to verify that the resulting `MultiSampleCount` is < the set value might be a more robust test and covers what we're looking to verify anyway.

Also ideally we could also enable it to run on OpenGL and XNA now and not just Windows DX platforms.

Jure-BB:
> In my case my reference device gets 16 and not 32.

I thought reference device should support every Direct3D feature at max levels. I'm not 100% sure. https://msdn.microsoft.com/en-us/library/windows/desktop/ff476328(v=vs.85).aspx

Could you provide what are the relevant differences between systems (OS version, SDK version etc)? It seems Windows SDK has to be installed for reference device to work. 

> note the assert also should be fixed as expected and actual parameters are reversed

My mistake :)

> Assuming A is not true... my guess is fixing it to verify that the resulting MultiSampleCount is < the set value might be a more robust test and covers what we're looking to verify anyway.

That won't catch bugs that disable multisampling (e.g. force it to always be zero). In that case I think we should also verify that resulting MultiSampleCount is greater than 0. Otherwise I think it would be a good solution, especially if B is true.

tomspilman:
>  It seems Windows SDK has to be installed for reference device to work.

Well the creation code is requesting the reference device and i get a device back.  Don't know how to tell if i got the reference device or not.

> OS version, SDK version etc

Windows 10 Pro 1709.  I didn't have the Win 10 SDK installed... trying that next.

tomspilman:
[...]

PR #6446, opened 2018-09-27
Fix a test when it runs on some machines

siudeks:
According to #6255 test 'TooHighMultiSampleCountClampedToMaxSupported' fails on some machines. I just noticed it doesn't work on my own PC as well, so the PR contains some changes more flexible in terms of the machine where test runs

Now test uses GraphicsDevice settings instead of hardcoded value


Jjagg:
Thanks @siudeks!

PR #6449, opened 2018-09-28
[Pipeline Tool] Fix crash when canceling choose folder dialog

cra0zy:


Jjagg:
Thanks @cra0zy! Merging!

PR #6450, opened 2018-09-28
[Pipeline Tool] Fix dependency loading (fixes #6443)

### LINKED ISSUE #6443: MonoGame Pipeline Tool Custom Processor bug
I have an issue with MonoGame Pipeline Tool 3.7.0.1708 where my custom processor project is no longer accepted by the pipeline tool and it gives a "Invalid / Missing Processor" where it used to be assigned.
Doing some code digging, it looks like it occurred due to commit #6202, where the following code is throwing an exception (which is swallowed) in the file PipelineTypes.cs:
```
  var a = Assembly.Load(File.ReadAllBytes(path));      // This line is the problem
  var types = a.GetTypes();                            // This throws an exception
```
Looking at the previous code, it used to be `var a = Assembly.LoadFrom(path);`

Fishing [in stackoverflow](https://stackoverflow.com/questions/16597138/loading-an-assembly-by-bytes-loses-the-location) it seems there are subtle differences between `Assembly.Load` and ` Assembly.LoadFrom`

I don't fully understand what's going on, however changing this line back makes the pipeline recognize my project again.

FYI, the exception that got thrown is:
`System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.`

- MonoGame Pipeline Tool 3.7.0.1708

- Windows

EDIT:
Looking a bit more carefully at #6202, I see the point is so the custom pipeline dll isn't locked by the pipeline tool, which is why Assembly.Load is used instead of Assembly.LoadFrom.
It seems the issue with Assembly.Load(byte[]) is that it can't locate the dependent libraries because it no longer knows its location of itself so the app only looks in the pipeline app's directory?
In any case, changing the code to include AppDomain.CurrentDomain.AssemblyResolve & loading dependencies:
```csharp
Dictionary<string, Assembly> asmMap = new Dictionary<string, Assembly>();
var mainAsm = AppDomain.CurrentDomain.GetAssemblies();
ResolveEventHandler resolveHandler = (object sender, ResolveEventArgs args) =>
{
    asmMap.TryGetValue(args.Name, out var assembly);
    return assembly;
};
AppDomain.CurrentDomain.AssemblyResolve += resolveHandler;

foreach (var path in assemblyPaths)
{
    try
    {
        var a = Assembly.Load(File.ReadAllBytes(path));
        asmMap[a.FullName] = a;
        foreach (var childAsmName in a.GetReferencedAssemblies())
        {
            var aa = mainAsm.FirstOrDefault(x => x.FullName == childAsmName.FullName);
            if (aa != null)
            {
                continue;
            }

            var chPath = Path.Combine(Path.GetDirectoryName(path), childAsmName.Name + ".dll");
            if (!File.Exists(chPath))
            {
                continue;
            }
            var a2 = Assembly.Load(File.ReadAllBytes(chPath));
            asmMap[a2.FullName] = a2;
        }
        var types = a.GetTypes();
        ProcessTypes(types);

        var watch = new FileSystemWatcher();
        watch.Path = Path.GetDirectoryName(path);
        watch.EnableRaisingEvents = true;
        watch.Filter = Path.GetFileName(path);
        watch.Changed += (sender, e) =>
        {
            if (Path.GetFileName(path) == e.Name)
                PipelineController.Instance.OnReferencesModified();
        };
        watch.Created += (sender, e) =>
        {
            if (Path.GetFileName(path) == e.Name)
                PipelineController.Instance.OnReferencesModified();
        };

        _watchers.Add(watch);
    }
    catch (Exception e)
    {
        //Logger.LogWarning(null, null, "Failed to load assembly '{0}': {1}", assemblyPath, e.Message);
        // The assembly failed to load... nothing
        // we can do but ignore it.
        continue;
    }
}

AppDomain.CurrentDomain.AssemblyResolve -= resolveHandler;
```
I'm not an expert in this area so I don't know if assuming dependent dlls being in the same folder as the desired assembly is enough (it is sufficient for my project...).

Furthermore, PropertyGridTable.Update() needed to be changed from
`drawable.Invalidate();`
to
`Application.Instance.Invoke(() => drawable.Invalidate());`
because when the file change detection occurs, it doesn't seem to be on the main UI thread and crashes the pipeline app.

cra0zy:


Jjagg:
Thanks @cra0zy!

PR #6456, opened 2018-09-30
Fix Gamepad index tracking under UWP

### LINKED ISSUE #6421: GamePad.GetState(int): Not preserving controller order on disconnect
I am compiling a sample project in the latest version of MonoGame with plans to port over an existing XNA project. In polling the controllers that are active, I have the following code (MaxPlayers defined as 8)

    // Update previous gamepad states and store the latest state
    for(int padIndex = 0; padIndex < Global.MaxPlayers; ++padIndex)
    {
      m_gamePadPrevious[padIndex] = m_gamePadCurrent[padIndex];
      m_gamePadCurrent[padIndex] = GamePad.GetState(padIndex);

      if (m_gamePadCurrent[padIndex].Buttons.A == ButtonState.Pressed)
      {
        Debug.WriteLine("Controller {0} A button pressed", padIndex);
      }
    }

This code, however, doesn't behave the same way as XNA.
XNA - Controller index 2 disconnects - Controller 2 IsConnected = false, all controllers before and after this index stay connected with IsConnected = true
MonoGame - Controller 2 disconnects - Controller 2 IsConnected = true, controllers after this index move down in the array, Controller 4 shows IsConnected = false instead.

If any controllers disconnect, the result shows the last controller as disconnected, even if any other player's controller disconnects.  This causes a huge problem with players controlling other characters.  I've tested this with 3 XBox 360 controllers and 2 XBox One controllers.

Is there an easy way to preserve the behavior that XNA provides or a code sample I can use to detect the properly assigned controller index that disconnected? I'm not seeing anything in GamePad or GamePadState that is obvious. I have 4 XBox 360 and 4 XBox One controllers to use for testing.

EDIT: I've tested this on the latest development build of MonoGame (3.7.0.1681) and the problem still occurs.

Zorcher

- MonoGame 3.6 and MonoGame 3.7

- Windows

- WindowsUWP

cra0zy:
Fixes #6421

### DIFF
--- a/MonoGame.Framework/Input/GamePad.UWP.cs
+++ b/MonoGame.Framework/Input/GamePad.UWP.cs
@@ -3,6 +3,9 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
 using WGI = Windows.Gaming.Input;
 
 namespace Microsoft.Xna.Framework.Input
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @cra0zy !

PR #6463, opened 2018-10-02
Add KeyCode in TextInputEventArgs on Windows DX

### LINKED ISSUE #6447: TextInputEventArgs.Keys is empty
TextInputEventArgs doesn't have the Keys property filled. This also affects both UAP and MacOS from looking at the codebase.

https://github.com/MonoGame/MonoGame/search?q=OnTextInput&unscoped_q=OnTextInput

https://github.com/MonoGame/MonoGame/blob/6ce62db0566114b5ed1adb655c7a13d9e1371bff/MonoGame.Framework/Windows/WinFormsGameWindow.cs#L328

Suggested code change:

```
using Keys = Microsoft.Xna.Framework.Input.Keys;

OnTextInput(sender, new TextInputEventArgs(e.KeyChar, (Keys)e.KeyChar));
```

- MonoGame 3.7

- Windows 10

- WinDX

Jjagg:
Fixes the Windows DX part of #6447. 

I should note that this solution isn't always correct as you cannot translate a character to a key unambiguously. For example pressing the divide key on my numpad gives a forward slash character which is translated back to OemQuestion because it can also be used to type a forward slash. Similarly all other numpad keys are never reported.

The alternative is to store the key pressed in WinForms' KeyDown event as was done in #4931, but that isn't foolproof either and more hacky to implement.

### DIFF
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -325,11 +325,11 @@ namespace MonoGame.Framework
         }
 
         [DllImport("user32.dll")]
-        private static extern short VkKeyScan(char ch);
+        private static extern short VkKeyScanEx(char ch, IntPtr dwhkl);
 
         private void OnKeyPress(object sender, KeyPressEventArgs e)
         {
-            var key = (Keys) (VkKeyScan(e.KeyChar) & 0xff);
+            var key = (Keys) (VkKeyScanEx(e.KeyChar, InputLanguage.CurrentInputLanguage.Handle) & 0xff);
             OnTextInput(sender, new TextInputEventArgs(e.KeyChar, key));
         }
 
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Windows/WinFormsGameWindow.cs
```diff
@@ -323,9 +324,13 @@ private void OnMouseLeave(object sender, EventArgs e)
             }
         }
 
+        [DllImport("user32.dll")]
+        private static extern short VkKeyScanEx(char ch, IntPtr dwhkl);
+
         private void OnKeyPress(object sender, KeyPressEventArgs e)
         {
-            OnTextInput(sender, new TextInputEventArgs(e.KeyChar));
+            var key = (Keys) (VkKeyScanEx(e.KeyChar, InputLanguage.CurrentInputLanguage.Handle) & 0xff);
```
cra0zy: Wait, this works O o, never realized that XNA Keys enum was based off of this one.

Anyway, are you sure there is no better way of doing this like abusing other events to get more accurate information like I did for UWP?

Jjagg: I mentioned an alternative in my original post:

> The alternative is to store the key pressed in WinForms' KeyDown event as was done in #4931 , but that isn't foolproof either and more hacky to implement.

We just have to choose either one. Neither solution is great, but both are okay in most cases.

cra0zy: Oh, I missed that, well we could have both implemented and see what the users complain about and figure out the better way that way (I mean different implementations for UWP and DX versions).

Jjagg: Heh, that's actually pretty clever :D

PR #6469, opened 2018-10-04
Added docs on setting up MonoGame on Windows

simoneddeland:


cra0zy:
This is great! Thanks @simoneddeland!

simoneddeland:
@mrhelmut I was not aware, I might make a new PR adding that info then!

cra0zy:
@mrhelmut @simoneddeland I would much rather fix the installer than fix the doc on that one.

simoneddeland:
@cra0zy OK I'll skip it then =)

mrhelmut:
I agree!

PR #6475, opened 2018-10-04
[Templates] Better NetStandard template

cra0zy:
Usage of the `MonoGame.Framework.DesktopGL.Core` is not a mistake, the PCL package throws a bunch of warnings, as its not a .NET Standard package.

On a related note we won't need to maintain a separate PCL package in the future as any package can be used for bait and switch purpose.


Jjagg:
That's 1!

Issue #5454, opened 2017-02-01
[MGCB] Shader: compile array is not supported?


LonghronShen:
@tomspilman Here this shader code compile failed using the latest 3.6.0.1465 content pipeline:
```C
texture Texture;
uniform const sampler TextureSampler : register(s0) = sampler_state
{
	Texture = (Texture);
	MipFilter = Linear;
	MinFilter = Linear;
	MagFilter = Linear;
};

uniform const int UseSphere;
texture Sphere;
uniform const sampler SphereSampler : register(s1) = sampler_state
{
	Texture = (Sphere);
	MipFilter = Linear;
	MinFilter = Linear;
	MagFilter = Linear;
};
uniform const bool UseToon;
texture ToonTex;
uniform const sampler ToonTexSampler : register(s2) = sampler_state
{
	Texture = (ToonTex);
	MipFilter = Linear;
	MinFilter = Linear;
	MagFilter = Linear;
};

uniform shared const float3	EyePosition;		// in world space

uniform const float3	DiffuseColor	: register(c5) = 1;
uniform const float		Alpha			: register(c6) = 1;
uniform const float3	EmissiveColor	: register(c7) = 0;
uniform const float3	SpecularColor	: register(c8) = 1;
uniform const float		SpecularPower	: register(c9) = 16;
uniform const bool		Edge=true;

uniform const float3	AmbientLightColor;
uniform const float3	DirLight0Direction;

uniform const float4x4	World;	
uniform shared const float4x4	View;
uniform shared const float4x4	Projection;

struct ColorPair
{
	float3 Diffuse;
	float3 Specular;
	float2 ToonTex;
};

struct CommonVSOutput
{
	float4	Pos_ws;
	float4	Pos_ps;
	float4	Diffuse;
	float3	Specular;
	float2 ToonTexCoord;
	float2 SphereCoord;
};


//-----------------------------------------------------------------------------
// Shader I/O structures
// Nm: Normal
// Tx: Texture
// Vc: Vertex color
//
// Nm Tx Vc
//  1  0  0 VSInputNm
//  1  0  1 VSInputNmVc
//  1  1  0 VSInputNmTx
//  1  1  1 VSInputNmTxVc


//-----------------------------------------------------------------------------
// Vertex shader inputs
//-----------------------------------------------------------------------------


struct VSInputNm
{
	float4	Position	: POSITION;
	float3	Normal		: NORMAL;
};

struct VSInputNmVc
{
[...]

PR #6482, opened 2018-10-05
Update custom effect docs

Jjagg:
Adds some information regarding shader models, defined symbols and fixes some typos. I really should've put this in a long time ago considering how many times I wrote this down in the forum.
Also add a note that preshaders are not supported (#5454).

### DIFF
--- a/Documentation/custom_effects.md
+++ b/Documentation/custom_effects.md
@@ -73,13 +73,3 @@ These are some tips for writing or converting effects for use with MonoGame.
 * The effect compiler is aggressive about removing unused parameters, be sure the parameters you are setting are actually used.
 * Preshaders are not supported.
 * If you think you've found a bug porting a shader [please let us know](https://github.com/MonoGame/MonoGame/issues).
-
-# Roadmap
-There is still work to be done for better support of custom effects and shaders in MonoGame:
-
-* Support GLSL in FX files.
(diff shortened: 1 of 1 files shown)


Documentation/custom_effects.md
```diff
@@ -33,23 +34,52 @@ To use a custom effect with MonoGame you must do one of the following (not both)
...
-* If you think you've found a bug porting a shader [please let us know](https://github.com/mono/MonoGame/issues).
+* The effect compiler is aggressive about removing unused parameters, be sure the parameters you are setting are actually used.
+* Preshaders are not supported.
+* If you think you've found a bug porting a shader [please let us know](https://github.com/MonoGame/MonoGame/issues).
 
 # Roadmap
 There is still work to be done for better support of custom effects and shaders in MonoGame:
 
 * Support GLSL in FX files.
   * Write a new preprocessor replacing [D3DPreprocess](http://msdn.microsoft.com/en-us/library/windows/desktop/dd607332.aspx).
```
tomspilman: We need to remove the roadmap.  It is crazy out of date.

Jjagg: Yeah, that's true. Just remove it altogether?

tomspilman: Yeah... i don't think it serves any purpose here.

Jjagg: Yeah, actually none of these are still planned.

PR #6484, opened 2018-10-06
Custom /define preprocessor parameter taken into account

sqrMin1:
MonoGame.Content.Builder.targets

Custom /define parameter only working when they are set right after the MGCB.exe command. 
E.g: MGCB.exe /define:Test=No /@:ContentReference.txt

To make this work I simply moved the $(MonoGameMGCBAdditionalArguments) property right after the $(MonoGameContentBuilderCmd) property in the "RunContentBuilder" target.

This helps when dynamically modifiyng the content respond file (.mgcb) and creating specific content in own build scenarios.

Originally occured here:
http://community.monogame.net/t/how-to-ignore-some-content-files-for-debug-builds/10946/7?u=sqrmin1

Additionally I placed the $(Header) property right behind $(MonoGameMGCBAdditionalArguments) to make the console output better readable (the following /outputDir and /intermediateDir can be pretty long). I tested this and it works.


Jjagg:
Awesome! Thanks @sqrMin1!

PR #6485, opened 2018-10-06
[Templates] Update pipeline extension template

cra0zy:
Notes:
 - For now we use .NET Framework 4.5 as the default target, tho you can use net standard if you want by adding `-f netstandard2.0` as argument


Jjagg:
Building with netstandard2.0:
`....Projects\MyAwesomeGame\MyAwesomeGame.csproj` : warning NU1701: 
Package 'MonoGame.Framework.Content.Pipeline.Portable 3.7.0.1708' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETStandard,Version=v2.0'. This package may not be fully compatible with your project.

Is this because the PCL target isn't compatible with netstandard2.0?

cra0zy:
> Is this because the PCL target isn't compatible with netstandard2.0?

Yes, but it can be safely ignored.

Jjagg:
Thanks!

PR #6486, opened 2018-10-07
[Templates] Update DesktopGL template

cra0zy:
.NET Core app by default in this case, you can pass `-f net45` if you need .NET Framework (and exchange the nuget package). Later on, we will merge the nuget package so one package can be used from both .NET Core and .NET Framework so package replacement step won't exist.


Jjagg:
Thanks!

PR #6489, opened 2018-10-08
Only handle MG window and exit when it's closed

### LINKED ISSUE #6432: Game will not close using window close button (Unique Situation)
I am having an issue with the window/game freezing when clicking the window's X close button.

Hold onto your hats. I am going to try to explain this the best that I can and I am sure that the issue has to do with the way I am doing this and using **MonoGame**.  I just don't know how to fix it.

I am creating a game engine that has a **plugin type architecture**. The whole idea is that my engine will be able to load implementations of various parts of my engine, by loading a DLL library at runtime that contains those implementations.

Doing this will let me (or other developers) actually take only certain parts of my engine and swap out the implementation with another one as long as they adhere to the interfaces provided. This shared library that contains all the interfaces is in its own library.  The implementation library is another separate project.

This implementation library would be the **plugin** for the underlying core of the engine.

So the way it would work would be this.  Developer John Doe would create a C# windows project running .NET Framework 4.5.  This project is the **game** and would have a reference to my engine library for use.  John Doe would then create a C# library project using .NET Framework 4.5 under the same solution.  For this library project he would reference my **engine core** library that contains all of the interfaces that he has to adhere to for my game engine to use the new implementation library he is creating.

This implementation library is loaded during runtime by my game engine.  Throughout my engine I use a static class to pull a particular object implementation from the assembly that was loaded and inject it where its needed.

---

Example Of Usage:

I create the windows app, reference the game engine.  Create another library project called **MyImplementationA** and I reference my engine core library.  In the implementation library I just created I would create a class called **TextureA** that implements the **ITexture** interface that comes from the core library. I write my **implementation** for **TextureA**.

Now I create another implementation library project called **MyImplementationB** and reference my engine core library just like above.  But this time I create a class called **TextureB** that implements the **ITexture** interface and write its implementation.

Now I have 2 assemblies.  **MyImplementationA.dll** implements **TextureA** and **MyImplementationB.dll** implements **TextureB**.  Now at this point I can just simply load either library up and I get that particular functionality.  

Doing this means that if **DevA** wants to use **MonoGame** to do some of the work for the engine under the hood, 
and **DevB** wants to use **SDL**, it will all operate the same.  Or they can use something else if they wanted.

Now everything I explained above works like a charm.  Its pretty awesome and I am very happy with the results.

But.....my implementation that I am using for testing my game engine is MonoGame.  This means in the implementation library it uses the **Game** class.  I am only assuming that this is the cause because I technically am loading the **MonoGame** **Game** during runtime from another assembly.  Could this be the issue?  I know that it must be obviously running in the same process as my game engine, because my engine is the one loading the assembly so that means that **MonoGame** under the hood would be running in the same process as my engine.

One thing I tried to do is have my engine get access to the **Game.Exit()** method so that way I could try to exit the game and close the window in code.  This does not work.  It just throws a **System.NullReferenceException** at the line of code where **Game.Run()** is used.  This tells me that at this point **Game** has been disposed of, garbage collected, and doesn't exist anymore.  So I guess it is **"Exiting"**, but not closing the window.  And with that I think the reason the window doesn't close is because it isn't **"attached"** to **MonoGame** anymore.

I know this situation is unique and there are probably a lot of unknowns, but at this point I don't know what to do about the issue and really don't want this to prevent me from using **MonoGame** or any other libraries.

If you need more information or have some questions for me to help you understand, don't hesitate to ask.

Any help would greatly be appreciated!!

---

- Windows 10

- MonoGame 3.6

- DesktopGL

Jjagg:
Checks the window id so we only handle window events for the window we created. Also sets the exit flag when that window is closed. Fixes #6432.

### DIFF
--- a/MonoGame.Framework/SDL/SDL2.cs
+++ b/MonoGame.Framework/SDL/SDL2.cs
@@ -381,6 +381,10 @@ internal static class Sdl
         public static d_sdl_destroywindow Destroy = FuncLoader.LoadFunction<d_sdl_destroywindow>(NativeLibrary, "SDL_DestroyWindow");
 
         [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+        public delegate uint d_sdl_getwindowid(IntPtr window);
+        public static d_sdl_getwindowid GetWindowId = FuncLoader.LoadFunction<d_sdl_getwindowid>(NativeLibrary, "SDL_GetWindowID");
+
+        [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
         private delegate int d_sdl_getwindowdisplayindex(IntPtr window);
         private static d_sdl_getwindowdisplayindex SDL_GetWindowDisplayIndex = FuncLoader.LoadFunction<d_sdl_getwindowdisplayindex>(NativeLibrary, "SDL_GetWindowDisplayIndex");
 
--- a/MonoGame.Framework/SDL/SDLGamePlatform.cs
+++ b/MonoGame.Framework/SDL/SDLGamePlatform.cs
@@ -171,14 +171,19 @@ namespace Microsoft.Xna.Framework
                 }
                 else if (ev.Type == Sdl.EventType.WindowEvent)
                 {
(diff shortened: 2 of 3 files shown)


Jjagg:
@cra0zy This one's good to go IMO.

PR #5809, opened 2017-07-11
Remove Windows 8.1 and Windows Phone 8.1 support

Jjagg:
As per #5807 :) 

I checked installers, NuGets, templates and Protobuild files and removed files that are not used by other platforms. I still need to check preprocessor directives. Anything else?

PR #6491, opened 2018-10-08
Remove unused WP files

Jjagg:
I noticed we had a couple unused files left from WP. They aren't used anymore since #5809.


cra0zy:
All this removing of useless data is really gonna be helpful for Protobuild removal, nice work 👍

PR #6493, opened 2018-10-08
[Templates] Delete PCL template

cra0zy:
.NET Standard template should be used instead of PCL one, PCL is dead, as such we should not advocate its usage.


tomspilman:
FYI.  I use the bait-and-switch "portable" version of MG and the Pipeline assembly all the time.   It is extremely useful in setting up content pipeline extensions and 3rd-party C# libraries to work with various MonoGame versions when you're targeting 3 or 4 console platforms.

We cannot kill that without a *perfect* replacement solution.  Perfect meaning it has to work 100% with existing C# libraries that are not .NET Standard without rewriting them or even fixing parts of them.  It has to be a drop in replacement for our existing portable assembly without need to change the code.

I say this as someone that doesn't want to see their workload increase fixing C# libraries to work for .NET standard when there is absolutely no need to do so.

cra0zy:
@tomspilman I'm just removing a template so that people don't use a deprecated API, not proposing we drop support for PCL yet.

On a related note, .NET Standard is just an updated version of PCL, it has the exact same use cases except it might have a higher required framework version, and a lot more C# API access: https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support

cra0zy:
@tomspilman @Jjagg All good now for this to be merged?

Jjagg:
Yes, I'm fine with this. I'll let Tom merge this because of the discussion :)

PR #6494, opened 2018-10-09
[Templates] Update Android template

cra0zy:
Important notes:
- By default sets API 21 (Android 5.0) as the minimal version, there is nothing stoping you from switching to a lower API if you need it.
- No splash screen for extra speed, also its not really a good policy to use just one image for a splash screen, and also there are multiple ways to make splash screens for Android depending on what you want so... no splash screen.


stromkos:
Could you include the API changes into the documentation page for consistency? https://github.com/MonoGame/MonoGame/blob/develop/Documentation/android.md

cra0zy:
> Could you include the API changes into the documentation page for consistency? https://github.com/MonoGame/MonoGame/blob/develop/Documentation/android.md

Hmm... looking at that page, that page is totally misleading.

Jjagg:
Thanks @cra0zy!

PR #6498, opened 2018-10-11
Fix custom effect doc markdown

Jjagg:
This didn't render right on the website:

image


Jjagg:
@tomspilman Looks like this didn't get pushed to the website: http://www.monogame.net/documentation/?page=Custom_Effects

Issue #6499, opened 2018-10-11
Runtime exception "Shader Compilation Failed" on iOS and Android, works on Windows


dtaylorus:
I am successfully using a shader on Windows and now trying to port it to iOS and Android.

The shader compiles just fine via the content pipeline for iOS and Android.

However, it fails with the following exception at runtime on iOS and Android:

> System.InvalidOperationException: Shader Compilation Failed occurred

The HLSL version of the shader is:

````
#if OPENGL
	#define SV_POSITION POSITION
	#define VS_SHADERMODEL vs_3_0
	#define PS_SHADERMODEL ps_3_0
#else
	#define VS_SHADERMODEL vs_4_0_level_9_3
	#define PS_SHADERMODEL ps_4_0_level_9_3
#endif

float2 TextureSize;
float PxRange;

sampler TextureSampler : register(s0);

float4 SamplePS(float4 Position : SV_POSITION, float4 Color : COLOR0, float2 TexCoord : TEXCOORD0) : SV_TARGET0
{
	float2 msdfUnit = PxRange / TextureSize;
	float3 samp = tex2D(TextureSampler, TexCoord).rgb;
	float sigDist = max(min(samp.r, samp.g), min(max(samp.r, samp.g), samp.b)) - 0.5f;
	sigDist *= dot(msdfUnit, 0.5f / fwidth(TexCoord));
	float opacity = clamp(sigDist + 0.5f, 0.0f, 1.0f);
	return Color * opacity;
}

technique Technique1
{
	pass Pass1
	{
		PixelShader = compile PS_SHADERMODEL SamplePS();		
	}
};
````

The GLSL created by the content pipeline is:

````
#ifdef GL_ES
precision mediump float;
precision mediump int;
#endif

uniform vec4 ps_uniforms_vec4[2];
const vec4 ps_c2 = vec4(-0.5, 0.5, 0.0, 0.0);
vec4 ps_r0;
vec4 ps_r1;
vec4 ps_r2;
#define ps_c0 ps_uniforms_vec4[0]
#define ps_c1 ps_uniforms_vec4[1]
uniform sampler2D ps_s0;
varying vec4 vFrontColor;
#define ps_v0 vFrontColor
#define ps_oC0 gl_FragColor
varying vec4 vTexCoord0;
#define ps_v1 vTexCoord0

void main()
{
	ps_r0.xy = dFdx(ps_v1.xy);
	ps_r0.zw = dFdy(ps_v1.xy);
[...]

PR #6501, opened 2018-10-12
Enable standard derivatives extension for GLSL shaders

dtaylorus:
Fix for issue #6499: Runtime exception "Shader Compilation Failed" on iOS and Android, works on Windows.

### DIFF
--- a/MonoGame.Framework/iOS/iOSGameViewController.cs
+++ b/MonoGame.Framework/iOS/iOSGameViewController.cs
@@ -145,6 +145,21 @@ namespace Microsoft.Xna.Framework
 
         #endregion
 
+        #region iOS 11 or newer
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


Jjagg:
@dtaylorus In the future it would be nice if you could work with a [feature branch](https://guides.github.com/introduction/flow/) so you can keep your develop branch up to date, keep PRs to a logical set of changes and including your changes in your projects before they're merged. 

That said, IMO both of the changes in this branch are good to go. @tomspilman Is it fine for you to just merge these like this?

Fixes #6499 and #6553.

tomspilman:
I'm good with this.  Thanks for submitting it @dtaylorus !

dtaylorus:
Thank you! And thanks for the info on feature branches. I'll use that for any future changes.

Issue #6507, opened 2018-10-19
MonoGame Android 3.7.0.1708 : Texture2D.FromStream() now refuses to load certain PNG files


shimfish:
PNG files that worked fine with MonoGame 3.6, no longer load using Texture2D.FromStream with MonoGame Android 3.7.

Example image attached:
![tv](https://user-images.githubusercontent.com/43265983/47209790-e7f74880-d399-11e8-80f6-7c294e76e9ce.png)

Fails with stack trace:

`{System.InvalidOperationException: This image format is not supported ---> System.NotSupportedException: Specified method is not supported.
  at Android.Runtime.InputStreamInvoker.Seek (System.Int64 offset, System.IO.SeekOrigin origin) [0x00000] in <f22a0e61648d492d93a9f4b449a72df2>:0 
  at MonoGame.Utilities.ImageReader.SkipCallback (System.Void* user, System.Int32 i) [0x00000] in <7deb70aebe4a49d3a23e40fc033ccbf1>:0 
  at MonoGame.Utilities.Imaging.stbi__skip (MonoGame.Utilities.Imaging+stbi__context s, System.Int32 n) [0x00040] in <7deb70aebe4a49d3a23e40fc033ccbf1>:0 
  at MonoGame.Utilities.Imaging.stbi__parse_png_file (MonoGame.Utilities.Imaging+stbi__png z, System.Int32 scan, System.Int32 req_comp) [0x0074a] in <7deb70aebe4a49d3a23e40fc033ccbf1>:0 
  at MonoGame.Utilities.Imaging.stbi__do_png (MonoGame.Utilities.Imaging+stbi__png p, System.Int32* x, System.Int32* y, System.Int32* n, System.Int32 req_comp, MonoGame.Utilities.Imaging+stbi__result_info* ri) [0x00022] in <7deb70aebe4a49d3a23e40fc033ccbf1>:0 
  at MonoGame.Utilities.Imaging.stbi__png_load (MonoGame.Utilities.Imaging+stbi__context s, System.Int32* x, System.Int32* y, System.Int32* comp, System.Int32 req_comp, MonoGame.Utilities.Imaging+stbi__result_info* ri) [0x0000d] in <7deb70aebe4a49d3a23e40fc033ccbf1>:0 
[...]

PR #6508, opened 2018-10-21
Fix for #6507

rds1983:
As some platforms doesnt support Stream.Seek(see #6507), I've made it so the image data is loaded into the memory and then parsed from there.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/TextureImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/TextureImporter.cs
@@ -194,11 +194,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         {
             var output = new Texture2DContent { Identity = new ContentIdentity(filename) };
 
-            var reader = new ImageReader();
             int width, height, comp;
             byte[] data = null;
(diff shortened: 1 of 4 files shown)


MonoGame.Framework.Content.Pipeline/TextureImporter.cs
```diff
@@ -198,7 +198,7 @@ TextureContent LoadImage(string filename)
             int width, height, comp;
             byte[] data = null;
             using (var stream = File.Open(filename, FileMode.Open, FileAccess.Read, FileShare.Read))
-                data = reader.Read(stream, out width, out height, out comp, Imaging.STBI_rgb_alpha);
+                data = ImageReader.Read(stream, out width, out height, out comp, Imaging.STBI_rgb_alpha);
```
Jjagg: You don't need the reader instance anymore when loading an image in the changed files.  Sorry, I missed that before.

rds1983: Thanks! I've made ImageReader static and removed all places it had been instantiated.

Jjagg:
Awesome, thanks @rds1983!

PR #6510, opened 2018-10-23
Update Rectangle.cs

willmotil:
Location Size Returnable or Setable as a Vector2. and Center Returnable as a Vector2.

This is proposed as a QOL change.
 
For example converting a location to a vector2 for use in a Drawstring or back is both extra typing and more calls. 
Considering just how many times this can happen in a game as well as the extra time (probably within the framework as well) i think it would be nice to have.


Jjagg:
I'm closing this because `Point.ToVector2` exists.

Issue #6515, opened 2018-10-30
[OpenAL] Wrong exception type when binaries are missing


mrhelmut:
Hey there,

Even if none of the ```Audio``` namespace is used, MonoGame initialize the static ```OpenAL``` bindings and tries to load the library and its entry points, resulting in a ```EntryPointNotFoundException``` if the OpenAL binaries are missing.

This is an issue when using third party sound engines and not shipping the OpenAL binaries.

Please also note that this is broken: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/SDL/SDLGamePlatform.cs#L72

The ```DllNotFoundException``` is no more triggered because the entry points loading have changed and are now throwing ```EntryPointNotFoundException```. And even if ```GetNativeLibrary()``` would throw the correct exception type, this code won't be triggered because it happens in the static initializer and no more within the ```OpenALSoundController``` constructor.

Related to #6107 and #4717

I think we should move forward with making the ```Audio``` namespace and its dependencies fully optional.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- All

#### What MonoGame platform are you using:

- DesktopGL

mrhelmut:
I'm working on refactoring the entry points loading.

Jjagg:
> I think we should move forward with making the Audio namespace and its dependencies fully optional.

Yes, I agree. People should easily be able to use a more elaborate third-party solution for audio if they need to.

mrhelmut:
I've setup #6516 to repair OpenAL entry points, and I'm looking into making the audio namespace optional right now.

The main thing is to avoid calling ``` SoundEffect.InitializeSoundEffect()``` within ```FrameworkDispatcher``` and skipping on the audio stuff in ```DoUpdate()```. For this, we need something like a "NoAudio" parameter to pass to ```FrameworkDispatcher```.

What would be the best strategy? Adding a static property to ```FrameworkDispatcher``` so that it can be set before calling the ```Game``` ctor? Or adding a ```Game()``` overload? Or something else?

cc @tomspilman for the design decision.

cra0zy:
This is actually a duplicate issue, there has been an issue about not crashing when openal is missing and instead just not playing sounds (also when audio device is missing).

mrhelmut:
[...]

PR #6516, opened 2018-10-30
Refactoring AL entry points loading

mrhelmut:
This is a first pass at addressing #6515. It refactors the entry points so that ```DllNotFoundException``` and ```NoAudioHardwareException``` are correctly thrown upon OpenAL initialization failures.

cc @cra0zy @Jjagg 

If this gets merged, I'll work on another PR to make the ```Audio``` namespace optional.

### DIFF
--- a/MonoGame.Framework/Audio/OpenAL.cs
+++ b/MonoGame.Framework/Audio/OpenAL.cs
@@ -191,7 +191,72 @@ namespace MonoGame.OpenAL
 
     internal class AL
     {
-        public static IntPtr NativeLibrary = GetNativeLibrary();
+        internal static void LoadEntryPoints()
+        {
+            IntPtr nativeLibrary = GetNativeLibrary();
+
+            if (nativeLibrary == IntPtr.Zero)
+            {
+                throw new DllNotFoundException("Couldn't load the OpenAL library, it appears to be missing.");
+            }
+
+            Enable = FuncLoader.LoadFunction<d_alenable>(nativeLibrary, "alEnable");
+            alBufferData = FuncLoader.LoadFunction<d_albufferdata>(nativeLibrary, "alBufferData");
+            alDeleteBuffers = FuncLoader.LoadFunction<d_aldeletebuffers>(nativeLibrary, "alDeleteBuffers");
+            Bufferi = FuncLoader.LoadFunction<d_albufferi>(nativeLibrary, "alBufferi");
+            GetBufferi = FuncLoader.LoadFunction<d_algetbufferi>(nativeLibrary, "alGetBufferi");
+            Bufferiv = FuncLoader.LoadFunction<d_albufferiv>(nativeLibrary, "alBufferiv");
+            alGenBuffers = FuncLoader.LoadFunction<d_algenbuffers>(nativeLibrary, "alGenBuffers");
+            alGenSources = FuncLoader.LoadFunction<d_algensources>(nativeLibrary, "alGenSources");
+            GetError = FuncLoader.LoadFunction<d_algeterror>(nativeLibrary, "alGetError");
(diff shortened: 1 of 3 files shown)


cra0zy:
> This is a first pass at addressing #6515. It refactors the entry points so that `DllNotFoundException` and `NoAudioHardwareException` are correctly thrown upon OpenAL initialization failures.

We are trying to do the opposite, no exception should be thrown if audio stuff is missing and instead everything should continue as normal.

mrhelmut:
That's what I did here: https://github.com/MonoGame/MonoGame/pull/6516/files#diff-8c76ca43b718d6fa4dfb764513677375R200

I moved every static initialization to make sure that nothing gets statically initialized and throw an exception outside of the try/catch loop.

cra0zy:
> I moved every static initialization to make sure that nothing gets statically initialized and throw an exception outside of the try/catch loop.

It never throws an exception... If it does not get initialized, the default value is set, the only thing that could throw an exception at that point is actually trying to call the delegate.

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Utilities/FuncLoader.Desktop.cs#L48

mrhelmut:
> It never throws an exception... If it does not get initialized, the default value is set, the only thing that could throw an exception at that point is actually trying to call the delegate.

Oh ok! Though we still need to get them out of the static initializer so that we are sure that the native library pointer gets loaded before the delegates.

> On one console if MonoGame initialized the audio system itself and you happen to use FMOD it would not work correctly. In that project i had to hack the MG audio system to not initialize on startup to avoid the problem.

That's what I figured looking at the consoles. This should be workable.

cra0zy:
> Oh ok! Though we still need to get them out of the static initializer so that we are sure that the native library pointer gets loaded before the delegates.

It always does as `NativeLibrary` is getting set on its declaration: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenAL.cs#L194

You don't need to move any initialization code.

mrhelmut:
We do, if we want to make sure that ```DllNotFoundException``` and ```NoAudioHardwareException``` are correctly thrown. With static initialization it can get initialized anywhere and most importantly, outside the scope of these statements: 

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/OpenALSoundController.cs#L152

https://github.com/MonoGame/MonoGame/blob/6f34eb393aa0ac005888d74c5c4c6ab5615fdc8c/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs#L245

Which makes those try/catch blocks useless, because any exception is uncatchable due to its static ctor nature (and a missing library is currently not a silent error), which makes the detection of an audio initialization failures impossible (or at least, not consistent at all).

We need to make sure that things are done in an ordered and timed manner. We also need to have a control over whether or not the library is getting loaded if we want to avoid it be loaded while using third party audio engines.

tomspilman:
> We need to make sure that things are done in an ordered and timed manner. We also need to have a control over whether or not the library is getting loaded if we want to avoid it be loaded while using third party audio engines.

100% agree.

cra0zy:
@mrhelmut you can just ask if the `NativeLibrary` is `IntPtr.Zero` at any time to know if it got loaded.

tdeeb:
Have there been any updates regarding this?

Jjagg:
@tdeeb Anything that happens is public. If it's not here it hasn't happened.

mrhelmut:
I've been off to other stuff, but I'm planning to go back to this and make audio optional. But since there are multiple audio backend to deal with, it's most probably going to take some time before anything gets merged.

cra0zy:
@mrhelmut Not sure if you got what changes I requested, anyway want me to make a new PR with those changes?

mrhelmut:
@cra0zy I'm going to revamp this PR later, but feel free if they are changes you'd like to make, I'll branch from there.

PR #6518, opened 2018-11-01
Fix for Texture2D.FromStream when MemoryStream is passed to it.

rds1983:
A problem had been reported:
http://community.monogame.net/t/texture2d-fromstream-no-longer-works/11067

I was able to reproduce it. It seems there is a problem when MemoryStream is passed to Texture2D.FromStream.
This PR fixes it.


rds1983:
nvm, I was passing MemoryStream with position set to its end.

PR #6519, opened 2018-11-01
Added rewind to the start of the stream in the Texture2D.FromStream

rds1983:
Right now, following code doesnt work in the MG:
```c#
            using (var memoryStream = new MemoryStream())
            {
                using (var fileStream = File.OpenRead("d:\\temp\\default_uiskin.png"))
                {
                    fileStream.CopyTo(memoryStream);
                }

                _texture = Texture2D.FromStream(GraphicsDevice, memoryStream);
            }
```
Because memoryStream points to the end just before Texture2D.FromStream call.
While same code works well in the XNA.
This PR makes MG work same as XNA.

### DIFF
--- a/Test/Framework/Graphics/Texture2DNonVisualTest.cs
+++ b/Test/Framework/Graphics/Texture2DNonVisualTest.cs
@@ -103,6 +103,35 @@ namespace MonoGame.Tests.Graphics
         }
 
         [TestCase]
+        public void FromStreamAtTheEnd()
+        {
+            // Check whether texture can be loaded if a stream being at its end
+            using (var memoryStream = new MemoryStream())
+            {
+                using (var fileStream = File.OpenRead("Assets/Textures/red_128.png"))
+                {
+                    fileStream.CopyTo(memoryStream);
+                }
+                using (var texture = Texture2D.FromStream(gd, memoryStream))
+                {
+                    Assert.AreEqual(8, texture.Width);
+                    Assert.AreEqual(8, texture.Height);
+                    Assert.AreEqual(1, texture.LevelCount);
+                    var pngData = new Color[8 * 8];
+                    texture.GetData(pngData);
+
+                    for (var i = 0; i < pngData.Length; i++)
+                    {
+                        Assert.AreEqual(255, pngData[i].R);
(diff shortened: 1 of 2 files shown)


rds1983:
Actually, dont merge this PR yet. As it breaks mobile versions because they lack stream.Seek(see #6507). Maybe there are other ways of rewinding stream to the beginning?

rds1983:
Oh, I remembered about Stream.CanSeek.
So now this PR is good to be merged.
It'll make code demonstrated in the original post work. But only on backends that support Stream.Seek.

rds1983:
@cra0zy, Firstly, it's highly unlikely that stream passed to Texture2D.FromStream will contain anything than the image.

Secondly, it's much more likely that users will forget to rewind stream from time to time and will get "This image format is not supported" Exception.
Actually it even happened today: http://community.monogame.net/t/texture2d-fromstream-no-longer-works/11067/9?u=rds1983

Thirdly XNA rewinds stream(though its logic is more complicated as I can judge from creating FakeStream and tracing Read & Seek calls).
Also, btw, MG 3.6 DesktopGL rewinded stream too.

cra0zy:
Again, Stream related functions should not perform backwards seeking, its very much an unconventional behavior. 

Honestly the error you are describing is like reading to MemoryStream, and than creating a FileStream without seeking it and wondering why the code failed, its to be expected.

PS. If you really want to copy XNA behavior, I'm guessing you need to figure out if the stream is at the end, and only try to seek if that is the case.

rds1983:
@cra0zy yeah, you're right. I've updated the code so rewinding happens only if stream is at end.

Test/Framework/Graphics/Texture2DNonVisualTest.cs
```diff
@@ -102,6 +102,35 @@ public void FromStreamNotPremultiplied()
             }
         }
 
+        [TestCase]
+        public void FromMemoryStream()
```
cra0zy: Rename the method for the test to something that more accurately represents the test, FromStreamWithoutSeeking

rds1983: Good catch. I've renamed it to FromStreamAtTheEnd.

PR #6520, opened 2018-11-04
Replace BackRequested with HardwareButtons 

### LINKED ISSUE #6404: "B" button triggers "BACK" button on UWP (xbox controller)
UWP builds of MonoGame report that "BACK" is being pressed when the "B" button is pressed on an X1 controller.

This doesn't seem to be an issue for any other platforms, and seems to be somewhat intentional (UWP implements a generic "Back" event which gets mapped to the "B" button at the OS level, I believe).

More details here: http://community.monogame.net/t/xbox-one-back-vs-b-button-possible-bug/8862/6

I think that the issue is on line 312 of https://github.com/MonoGame/MonoGame/blob/157bdd9b3313be47eeebb882dfc8038f05b1c807/MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs

```
        private void BackRequested(object sender, BackRequestedEventArgs e)
        {
            // We need to manually hide the keyboard input UI when the back button is pressed
            if (KeyboardInput.IsVisible)
                KeyboardInput.Cancel(null);
            else
                _backPressed = true;

            e.Handled = true;
        }
```

You are registering for the "BackRequested" event (see line 133 of the same file) and treating that to mean that the back button has been pressed, but that is mapped to the "B" button on PC.

Committed here: https://github.com/MonoGame/MonoGame/pull/5810 by @atomgott (who may have more insight into this issue).

- MonoGame 3.6

- Windows 10

- WindowsUWP

nkast:
Fixes #6404

Uses mobile extensions to check for back button on phones. Tested on the Mobile emulator.

### DIFF
--- a/Build/GenerateProject.CSharp.xslt
+++ b/Build/GenerateProject.CSharp.xslt
@@ -816,6 +816,21 @@
     </Reference>
   </xsl:template>
   
+  <xsl:template name="ReferenceToSDKExtension"
+  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <xsl:param name="sdkReference" />
+    <xsl:param name="source_project" />
+
+    <SDKReference>
+      <xsl:attribute name="Include">
+        <xsl:value-of select="$sdkReference/@Include" />, Version=<xsl:value-of select="user:DetectWindows10InstalledSDK()" />
+      </xsl:attribute>
+      <Name>
+        <xsl:value-of select="$sdkReference/@Name" />
+      </Name>
+    </SDKReference>
+  </xsl:template>
+  
   <xsl:template name="ReferenceToGAC"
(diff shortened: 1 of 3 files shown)


Jjagg:
@nkast Did you manage to test this on desktop to make sure it doesn't break on non-mobile? If so I'll merge this :)

nkast:
I tested this on Windows 10 to make sure that an xbox controller doesn't fire the Back button,
and on the Windows 10 mobile emulator to make sure that we are still getting the hardware back button.

I took the code from this article that describes the use of ApiInformation for extensions.
https://docs.microsoft.com/en-us/windows/uwp/porting/w8x-to-uwp-porting-to-a-uwp-project

Jjagg:
Awesome, thanks!

PR #6535, opened 2018-11-13
Improved accuracy of fixed time step (recreation of PR #4207)

### LINKED ISSUE #6526: Gameloop with fixed frame rate 60 FPS produce unstable timing
Gameloop with fixed frame rate 60 FPS produce unstable timing.
The problem exist very clearly and should also affect low GPU load.
Screen's with lower refresh rate might not encounter this problem.

I have identified two problems with the gameloop timing.

1. The gameloop drift and is just below 60 fps. This will make at some point one frame loss (same frame drawn twice)
2. When idle and running Thread.Sleep() timing is cast to (int) making the wait little to short.
Because it then do RetryTick: it could make another Thread.Sleep() that will produce a significant long frame because Thread.Sleep() is not exactly and often can't produce very short delays.

**Comment:**
I suggest that the multiple Update() for long framerates and IsRunningSlowly could be removed or possible made as optional on/off. It is very confusing and hard to understand that there could be multiple Updates() for one Draw(). (I have comment out the code).

**Solution:**
1. Make the timing a little faster to produce slightly more than 60 FPS.
Decrease the sleep wait with one ms. This makes the game at some point produce one extra frame.
2. After Thread.Sleep() don't goto RetryTick() move forward instead.

**Code:**
```
            // Calculate the accumulated elapsed time.
            var currentTicks = _gameTimer.Elapsed.Ticks;
            _accumulatedElapsedTime = TimeSpan.FromTicks(currentTicks - _previousTicks);

            // If were in the fixed timestep mode and not enough time has elapsed
            // to perform an update we sleep off the the remaining time to save battery
            // life and/or release CPU time to other threads and processes.
            if (IsFixedTimeStep && _accumulatedElapsedTime < TargetElapsedTime)
            {
                var sleepTime = (int)(TargetElapsedTime - _accumulatedElapsedTime).TotalMilliseconds;
                if (sleepTime > 0) {
                    sleepTime--;
                }

                // NOTE: While sleep can be inaccurate in general it is 
                // accurate enough for frame limiting purposes if some
                // fluctuation is an acceptable result.
#if WINRT
                Task.Delay(sleepTime).Wait();
#else
                System.Threading.Thread.Sleep(sleepTime);
#endif
                currentTicks = _gameTimer.Elapsed.Ticks;
                _accumulatedElapsedTime = TimeSpan.FromTicks(currentTicks - _previousTicks);
            }
            _previousTicks = currentTicks;
….
            //_accumulatedElapsedTime = TimeSpan.Zero;


```<!-- System stats -->

- MonoGame 3.6 (with some of 3.7 updates)

- Windows 10 , Screen with 60 Hz refresh rate

- WindowsDX

### PR DESCRIPTION
**Pull request by MichaelDePiazzi**
This is a recreation of PR #4207 to get it all clean and up to date so it can be merged (as per @Jjagg's request). It resolves #6526 (and #4202 - the original issue which was mistakenly closed).

There is a lot of detailed information, discussion, test results, and benchmarks in the original PR #4207 and issue #4202.

But to quickly re-cap -- The main problem with the accuracy of fixed time step is the `Sleep` method. It is not accurate and is dependant on the current resolution of the system timer (which is influenced by many things).

If the current timer resolution can be measured, it is possible to still use `Sleep` and compensate for any inaccuracy with it. But if the timer resolution can't be measured, then the only alternative which maintains accuracy is to spin the CPU.

This PR does the following:

- On Windows desktop platforms, it will sleep for as long as possible without overshooting the desired time. It will then spin the CPU for the remaining time.
- On other platforms (where I couldn't measure the timer resolution), it will spin the CPU to maintain accuracy. (This also seems to be the same behaviour as XNA.)
- Implements the [`Game.InactiveSleepTime`](https://docs.microsoft.com/en-us/previous-versions/windows/xna/bb197440(v%3dxnagamestudio.42)) feature from XNA so that sleeping can be performed when the game window is inactive. This helps alleviate the higher CPU usage resulting from CPU spinning.
- Gets rid of the garbage generation on Windows UAP platforms which came from the use of `Task.Delay`.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -1190,6 +1190,9 @@
     <Compile Include="Utilities\ReflectionHelpers.Legacy.cs" >
       <Platforms>Angle,Linux,MacOS,Windows,WindowsGL,Web</Platforms>
     </Compile>
+    <Compile Include="Utilities\TimerHelper.cs">
+      <Platforms>Windows,WindowsGL</Platforms>
+    </Compile>
     <Compile Include="Utilities\Lz4Stream\Lz4DecoderStream.cs" />
     <Compile Include="Utilities\LzxStream\LzxDecoderStream.cs" />
     <Compile Include="Utilities\ZLibStream\ZlibStream.cs" />
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -411,6 +411,9 @@ namespace Microsoft.Xna.Framework
         private Stopwatch _gameTimer;
         private long _previousTicks = 0;
         private int _updateFrameLag;
+#if WINDOWS_UAP
+        private readonly object _locker = new object();
+#endif
 
         public void Tick()
         {
@@ -420,6 +423,16 @@ namespace Microsoft.Xna.Framework
             // modes across multiple devices and platforms.
 
         RetryTick:
+
+            if (!IsActive && (InactiveSleepTime.TotalMilliseconds >= 1.0))
+            {
+#if WINDOWS_UAP
+                lock (_locker)
+                    System.Threading.Monitor.Wait(_locker, (int)InactiveSleepTime.TotalMilliseconds);
+#else
+                System.Threading.Thread.Sleep((int)InactiveSleepTime.TotalMilliseconds);
+#endif
+            }
 
             // Advance the accumulated elapsed time.
             var currentTicks = _gameTimer.Elapsed.Ticks;
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -426,21 +426,14 @@ namespace Microsoft.Xna.Framework
             _accumulatedElapsedTime += TimeSpan.FromTicks(currentTicks - _previousTicks);
             _previousTicks = currentTicks;
 
-            // If we're in the fixed timestep mode and not enough time has elapsed
-            // to perform an update we sleep off the the remaining time to save battery
-            // life and/or release CPU time to other threads and processes.
             if (IsFixedTimeStep && _accumulatedElapsedTime < TargetElapsedTime)
             {
-                var sleepTime = (int)(TargetElapsedTime - _accumulatedElapsedTime).TotalMilliseconds;
-
-                // NOTE: While sleep can be inaccurate in general it is 
-                // accurate enough for frame limiting purposes if some
-                // fluctuation is an acceptable result.
-#if WINDOWS_UAP
-                Task.Delay(sleepTime).Wait();
-#else
-                System.Threading.Thread.Sleep(sleepTime);
+#if WINDOWS
+                // Sleep for as long as possible without overshooting the update time
+                var sleepTime = (TargetElapsedTime - _accumulatedElapsedTime).TotalMilliseconds;
+                Utilities.TimerHelper.SleepForNoMoreThan(sleepTime);
 #endif
+                // Keep looping until it's time to perform the next update
                 goto RetryTick;
             }
 
--- a/MonoGame.Framework/Utilities/TimerHelper.cs
+++ b/MonoGame.Framework/Utilities/TimerHelper.cs
@@ -1,0 +1,47 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace Microsoft.Xna.Framework.Utilities
+{
+    internal static class TimerHelper
+    {
+        [DllImport("ntdll.dll", SetLastError = true)]
+        private static extern int NtQueryTimerResolution(out uint MinimumResolution, out uint MaximumResolution, out uint CurrentResolution);
+
+        private static readonly double LowestSleepThreshold;
+
+        static TimerHelper()
+        {
+            uint min, max, current;
+            NtQueryTimerResolution(out min, out max, out current);
+            LowestSleepThreshold = 1.0 + (max / 10000.0);
+        }
+
+        /// <summary>
+        /// Returns the current timer resolution in milliseconds
+        /// </summary>
+        public static double GetCurrentResolution()
+        {
+            uint min, max, current;
+            NtQueryTimerResolution(out min, out max, out current);
+            return current / 10000.0;
+        }
+
+        /// <summary>
+        /// Sleeps as long as possible without exceeding the specified period
+        /// </summary>
+        public static void SleepForNoMoreThan(double milliseconds)
+        {
+            // Assumption is that Thread.Sleep(t) will sleep for at least (t), and at most (t + timerResolution)
+            if (milliseconds < LowestSleepThreshold)
+                return;
+            var sleepTime = (int)(milliseconds - GetCurrentResolution());
+            if (sleepTime > 0)
+                Thread.Sleep(sleepTime);
+        }
+    }
+}

### CODE REVIEW

**Comment by MichaelDePiazzi** [comment]
@Jjagg This should pretty much do it I think! Please let me know if anything else is required to get this merged.

File: Build/Projects/MonoGame.Framework.definition [hunk 1]
```diff
@@ -1190,6 +1190,9 @@
     <Compile Include="Utilities\ReflectionHelpers.Legacy.cs" >
       <Platforms>Angle,Linux,MacOS,Windows,WindowsGL,Web</Platforms>
     </Compile>
+    <Compile Include="Utilities\TimerHelper.cs">
+      <Platforms>Windows,WindowsGL</Platforms>
```
Jjagg: It's only WindowsDX that needs this. WindowsGL does not define the WINDOWS symbol so it won't use the Timer helper.

MichaelDePiazzi: Are you sure? The WindowsGL project definitely has this symbol defined for me:
image
Also from the Protobuild `MonoGame.Framework.definition`:
```
    <CustomDefinitions>
      ...
      <Platform Name="WindowsGL">TRACE;WINDOWS;DESKTOPGL;SUPPORTS_EFX</Platform>
      ...
    </CustomDefinitions>
```
Do you still want me to remove this?

MichaelDePiazzi:
> In your original PR you discussed having a settable property that would determine whether or not the CPU spins or the old, inaccurate Thread.Sleep method is used. Maybe that could still be an option for those concerned about the increased power usage resulting from the change.

Yes, but it was decided against after the discussion. Let's focus on getting this PR merged because it does solve the frame timing problem. After this has been merged, if the CPU usage does actually come up as a legitimate issue later, then we can potentially revisit an option like this.

Build/Projects/MonoGame.Framework.definition
Jjagg: Sorry, you're right. I didn't check. I thought it wasn't there in the Protobuild definition file because we want DesktopGL to have cross platform builds (though Mac uses some platform specific stuff already). It's fine as is for now :) I'll see about removing the symbol definition in a later PR, because there might be some complications.

Jjagg:
Thanks @MichaelDePiazzi! Let's merge this and see if anything comes up regarding increased CPU usage.

MichaelDePiazzi:
Woo!! After the amount of work that has gone into this, I'm pretty stoked to finally see this merged! Thanks for helping to get this across the line @Jjagg! :D

Jjagg:
It's a step forward IMO. Frame time issues come up every once in a while so I'm happy to merge a proposed solution to the issue :)

Build/Projects/MonoGame.Framework.definition
mrhelmut: No biggy, but ```WindowsGL``` should be removed here (it's not an issue that ```TimerHelper``` is included, as long as it's not used in DesktopGL, because ```ntdll.dll``` won't exist on Mac and Linux).

MichaelDePiazzi: Yeah, @Jjagg raised this also in his review. But he said it's fine as is for now, and he'd make some changes later.

tbh, I'm a little confused with this "DesktopGL" build (I only use the DirectX build). So to clarify - Is the "WindowsGL" project definition actually the "DesktopGL" build, which is also used on Mac and Linux? So if that's the case, then what are the "Linux" and "MacOS" project definitions used for?

Jjagg: Protobuild needs a definition for each platform. Even though the DesktopGL builds for Windows and Linux where the same (or rather should be) they both need definitions. But the way we structure the .definition files is pretty bad because it uses platforms instead of services (e.g. DesktopGL).

MichaelDePiazzi: Thanks @Jjagg, I think I understand now. So ultimately, there will be one build for "DesktopGL" which will work on all 3 platforms. But right now, there are still actually 3 separate builds. And this is why "WindowsGL" running this code is not a problem right now - because it's currently only used for the Windows build. But it is an issue in working towards having one build for all 3 platforms. Have I got this right?

Jjagg: Yes, kind of. I'm not sure what the plan for Mac is, but AFAIK before this PR the Windows and Linux builds were identical and now a Windows build won't run on Linux or Mac.

MichaelDePiazzi: Ok. Well I understand what needs to be done to fix this up now, so just say the word if you'd like me to submit a PR.

mrhelmut: Yeah, DesktopGL = WindowsGL = Linux. They used to be very different but they are now all fused and are the same.
Regarding Mac, both the "GL" target and the "Mac" target are almost identical, except that the "Mac" one has a platform specific video player implementation like Jjagg mentioned.

(Confusing, yes. I guess that at some point we should work toward having a unique video player for GL and fuse all those platforms.)

And sorry for not reading thoroughly Jjagg's review!

cra0zy: Yup, disable this code for WindowsGL, that is submit a PR for it.

MichaelDePiazzi: Thanks for the clarification guys! I'll get a PR submitted shortly to fix this up.

srodrigo:
Hi @MichaelDePiazzi, I just noticed yesterday some quite noticeable stuttering on 3.7 on Mac, so I found this issue :) In case this helps (it might not be related), I've printed out the elapsed time in both `Game.Update` and `Game.Draw`, and `Update` is totally consistent (0.0167s), but `Draw` is not (it prints `0.03334` some times. Also, my laptop is quite old (MBP 2011), I don't know if this could make an impact as well, although my game is extremely simple, and I haven't tried but I imagine just moving a texture around the screen would suffer from the same problem.

I'd like to try the `development` branch and see if your PR fixes this on Mac.

mrhelmut:
@srodrigo It won't, the fix is Windows only, for now.

MichaelDePiazzi:
> It won't, the fix is Windows only, for now.

This is not correct. This fix is for all platforms. The only difference with Windows (DirectX) is that it still tries to sleep when possible (because it can compensate for the inaccuracy). While other platforms will now only spin the CPU to maintain timing accuracy.

So @srodrigo, definitely give the development branch a try. This PR should resolve your stuttering issue (if it is in fact related to this fixed time step timing issue, which I suspect it is). Please let us know how it goes for you.

srodrigo:
Thanks @MichaelDePiazzi 

Unfortunately, I've tried the last build (3.8.0.132) but looks like I have the same problem. I've tried with a fresh project, which now interestingly it's Mac only, not Desktop (I wouldn't expect this to make it any worse). I have just one texture that moves through the screen, to isolate the problem, and still the same. I include the code just in case:

```c#
[...]
namespace Stuttering.MacOS
{
    public class Game1 : Game
    {
        SpriteBatch spriteBatch;

        Texture2D texture;
        Point enemy = new Point(0, 0);
        
        const int WIDTH = 2 * 640;
        const int HEIGHT = 2 * 360;
        
        public Game1()
        {
            graphics = new GraphicsDeviceManager(this)
            {
                PreferredBackBufferWidth = WIDTH,
                PreferredBackBufferHeight = HEIGHT
            };
        }

        [...]
        
        {
            
            texture = Content.Load<Texture2D>("enemy-1");
        }
        
        {
            if (GamePad.GetState(PlayerIndex.One).Buttons.Back == ButtonState.Pressed || Keyboard.GetState().IsKeyDown(Keys.Escape))
                Exit();

            enemy += new Point((int)(200.0f * gameTime.ElapsedGameTime.TotalSeconds), 0);

            base.Update(gameTime);
        }

        {
            Debug.WriteLine("Draw - gametime total seconds: " + gameTime.ElapsedGameTime.TotalSeconds);
            graphics.GraphicsDevice.Clear(Color.CornflowerBlue);

            spriteBatch.Begin();
            spriteBatch.Draw(texture, enemy.ToVector2(), Color.White);
            spriteBatch.End();

            base.Draw(gameTime);
        }
    }
}
```

The movement is related to the elapsed time, so this should be correct?

If I remove the `Debug` line, it works better, but still stutters. I haven't tried drawing the FPS to avoid slowing it down with Debug lines, but as the stutter is quite noticeable I didn't think that was useful. I attach in a separate file (to avoid cluttering this message) with the output of the elapsed seconds.
[stutter.txt](https://github.com/MonoGame/MonoGame/files/2588797/stutter.txt)

MichaelDePiazzi:
@srodrigo Bummer that it didn't work. There is nothing immediately obvious to me in this. So can you please open a new issue for this? That way we can discuss further and maybe run a few tests to find the cause.

srodrigo:
@MichaelDePiazzi I just noticed that I if I unplug my external screen, the problem is much less noticeable. But unfortunately, still happens every 8-10 seconds :( I might just need a new computer, as apparently my Intel HD Graphics 3000 can't handle an external screen and a game running at the same time.

I'll open a new issue if you think this shouldn't be related to an ancient graphics card.

MichaelDePiazzi:
@srodrigo Yes, please still open a new issue for this. I feel like something so basic should work smoothly even on older hardware. There are a couple of things I'd like to get you to try, but I'd like to move the discussion out of this closed PR first. And please be sure to include all the details you have provided here (for the benefit of everyone else).

mrhelmut:
My bad for misunderstanding the PR. :)

MichaelDePiazzi:
@mrhelmut No worries mate, all good! :)

damian-666:
on windows in c# you can do     [System.Runtime.InteropServices.DllImport("winmm.dll", EntryPoint = "timeBeginPeriod")]
        public static extern uint timeBeginPeriod(uint uMilliseconds);

you can set it to 1ms.  now i get 300 fps with piling  CCD ,  or  whatever , at least untill physics is maxed out..which on mine is before draw maxes out.

     i just hope if u turn on fixed time rate it wont spin cpu since im not using update,  i guess i could turn off and have vsync on tho..


this is a system wide setting  but chrome does it.  ( thats how some people figure it out, chrome was googling the question and issue went away lol)
with defaults its 16y ms..
simply meausre Sleep(1) and resuts vary from 1 to 16 ms... or 1/60 sec.. useless.  
        
    for battery on windows laptop    set it to 5 or something close enough and you can get a smooth framerate with producer /consumer. with physics on bk thread, and draw the  last good frame w vsync.    on  droid/ linux i donno how to set timers to be of any use and dont use any wait handles because they are depending on ths i thin.. 
    
    ill try the new spinning idea, but right now androidtimers and threads arei much slower so i max it out dont bother. on windows i get 4000 fps if i dont.  i dont do anyting on Updat anywaye i just draw what i got.

PR #6544, opened 2018-11-16
Fix VS template installation when C# folder is missing

Jjagg:
This has been reported a bunch on the forum. I don't think there's an open issue for it. Saw another report on the forum just now and decided to look into it. Turns out it's a really easy fix 😅

The issue was that if the `ProjectTemplates\Visual C#` directory was missing, the templates could not be installed.


tomspilman:
So what does it mean for the `ProjectTemplates\Visual C#` folder to not exist?  Does it mean VS is not installed correctly?

Jjagg:
No, in VS2017 the default templates are installed in a separate location and the Visual C# folder is sometimes not created (for whatever reason). Would be better if we use registry keys to check for VS installation of course, there's still a few open issues regarding the installer not finding VS installations.

tomspilman:
Ok this works for me... merging.

PR #6549, opened 2018-11-17
[Pipeline Tool] Clear the current selections when excluding items in the pipeline UI

tdeeb:
This PR clears the selection in the pipeline tool after excluding items. This prevents users from changing settings on items no longer in the project.


cra0zy:
Looks great, thanks @tdeeb !

PR #6556, opened 2018-11-22
WIP: Net namespace rewrite (new client/server implementation)

gustavolsson:
This is a rewrite of #5378 that implements a client-server based network topology (on the backend) as opposed to the old completely peer-to-peer implementation. This simplifies the code, will make it easier to implement anti-cheating functionality and works around a critical flaw of the old implementation*. I decided to create a new branch based on a more recent develop commit since most of the old code is obsolete now anyway.

This branch is not ready for merging yet. I'm in the process of releasing my [old XBLIG multiplayer game](https://gustavolsson.com/) on itch and since it uses most of the XNA networking functionality I'm quite confident that the behavior is correct once the game works like it did on Xbox. (Some unit-tests would be nice though)

Notable improvements over the current (develop branch) net implementation:
* Correct gamer id management (crucial)
* Support for multiple local gamers
* Master server console application to connect online gamers (I run mine on google cloud)

*Note on the critical flaw of the old implementation: The old implementation would force the master server to introduce all players to all other players in a session OR any session in which a peer was behind the same router as the host would be unjoinable for peers outside of their local network.


gustavolsson:
Is there a reason TeamCity is using `Microsoft (R) Build Engine version 12`? I'm trying to use the C# 6 feature `$ string interpolation` as a shorthand for `String.Format` but it breaks the TeamCity build.

Jjagg:
@tomspilman Maybe a good time to finally upgrade to C#6 (or even 7?). I think it's more annoying than useful for anyone at this point. Also, I think this came up before, but would be nice to enforce the supported language version in the csproj files (it is now set to 'Default') so users get the compiler errors locally rather than on the build bot. It's a minor change in the Protobuild definition files, so I'll set up a PR if that's alright.

Jjagg:
@mrhelmut I wasn't aware of runtime dependencies in C# 6. Do you know if the pitfalls are documented somewhere? Also, wouldn't the change still be fine if MG itself is compiled for .NET 4.5 because the compiler would report errors.

mrhelmut:
I don't know for sure, I think Tom would have to jump in for the details.

It's a mix of target framework and compiler version. I believe that the incompatibilities are mostly related to the VS 2017 compiler.
I'd say 4.6.2+ (C#7) is touchy, and 4.6 (C#6) might be fine if built with VS 2015.

gustavolsson:
I suspected consoles were to blame, if I remember correctly you generate native code from the CIL for them? In that case I definitely see how switching compiler could trip something up :)

C#6 is not a blocker for this PR so I will just rewrite the affected code!

mrhelmut:
That's it. :)

But it's true that having to avoid some syntax is kind of annoying in some situations.

gustavolsson:
Phew, everything is green now :)

What about the `GamerServices` namespace, does anyone rely on this nowadays? It seems to be quite badly maintained and include a lot of hard-coded stuff (for example `SignedInGamer.PlayerIndex` in DesktopGL). I'm a bit afraid to touch it as I don't have the build setup for other platforms.

Ideally, I would like to support signing in multiple users using `Guide.ShowSignIn(int paneCount, bool onlineOnly)` (showing some simple UI in which a controller/keyboard button can be pushed to connect a controller with a SignedInGamer) but I'm currently leaning towards building my own game-specific Input-wrapper and ignoring the `SignedInGamer.PlayerIndex` property.

mrhelmut:
I guess it's going to be ditched at some point, it's not usable. Making a complete implementation of ```GamerServices``` sounds like a touchy challenge given the API paradigm differences for each platform.

mrhelmut:
Better ask @tomspilman for design propositions. :)

My personal take on this would be that, unless we can provide a fully cross-platform sign-in system, it's most probably something to consider adding to another library such as MonoGame.Extended?
Public ctor for ```SignedInGamer``` looks like a relevant solution in the absence of a sign-in system.

tomspilman:
> unless we can provide a fully cross-platform sign-in system,

I don't think anyone can unless they 100% know inside and out all our platforms including the NDA'd console platforms.

This is why i've never considered much of the Net or Guide APIs all that useful.  That said the low level sending packets back and forth bits make sense.  In all platforms there is a:

   - Server/Host
   - Clients
   - Packets
   - Reliable/unreliable send/receive
   - Players
   - Guests
   - Friends

The problem is all the details from here out vary.

You can pretty much consider any "signin" to be asynchronous as depending on platform you may have one or more interactive popups to get a valid signin or for it to just fail.

> Also, if I decide to create a "choose controller/sign-in screen" for my game I might as well see if I can make it generic (and skinnable, even?) and include it in the GamerServices namespace for DesktopGL.

We wouldn't want that.

It will not work for anything other than PC platforms.  It won't work on any other of our platforms.

In fact anything that implements a UI is generally not portable and not something we want to include.

gustavolsson:
> > unless we can provide a fully cross-platform sign-in system,
> 
> I don't think anyone can unless they 100% know inside and out all our platforms including the NDA'd console platforms.

I agree, seems difficult.

> This is why i've never considered much of the Net or Guide APIs all that useful. That said the low level sending packets back and forth bits make sense. In all platforms there is a:
> 
>     * Server/Host
>     * Clients
>     * Packets
>     * Reliable/unreliable send/receive
>     * Players
>     * Guests
>     * Friends
> 
> The problem is all the details from here out vary.

Do you think this this PR should work on all platforms before being merged, or is it sufficient with Windows/macOS/Linux?

In my old re-implementation, a generic "backend" interface abstracted the Lidgren calls (which in turn abstracted the system socket calls). For this implementation I just use the Lidgren calls directly as the abstraction layer becomes too similar to what it abstracts when only one "backend" exists. That limits the reach of this PR to platforms on which Lidgren runs. One can of course introduce an abstraction layer to this branch, but the one designing that must have access to all platforms..

> You can pretty much consider any "signin" to be asynchronous as depending on platform you may have one or more interactive popups to get a valid signin or for it to just fail.

A public `SignedInGamer` constructor in conjunction with `NetworkSession.AddLocalGamer()` would support this.

> > Also, if I decide to create a "choose controller/sign-in screen" for my game I might as well see if I can make it generic (and skinnable, even?) and include it in the GamerServices namespace for DesktopGL.
> 
> We wouldn't want that.
> 
> It will not work for anything other than PC platforms. It won't work on any other of our platforms.
> 
> In fact anything that implements a UI is generally not portable and not something we want to include.

Alright, I'll not spend time on that then :)

gustavolsson:
> @gustavolsson Is this ready for review?

I don't consider the PR done yet, but I don't think it will change considerably before it is. I'm working on my game part time so there hasn't been any progress lately but the plan is to release it in a month or two. I suspect this code will change a bit during this time and a bit after release as well.

What about reviewing and potentially merging this, say, next week after I've had the time to do a proper full session test with my friends and then create a follow-up PR with the improvements?

Jjagg:
There's a lot of changes and complexity of networking code is high and AFAIK none of the active maintainers actually knows much about the structure of the networking namespace of MG (at least I don't know the codebase **at all**).

I think testing it is honestly the best way to verify the quality of the changes. And the best way to test this is with an actual game. So I'll hold off on merging this and wait for updates from you :) If you think it's kind of functional we can merge and let people know they can try it so we can get better testing coverage and feedback. Thanks for contributing back these improvements @gustavolsson :)

gustavolsson:
> There's a lot of changes and complexity of networking code is high and AFAIK none of the active maintainers actually knows much about the structure of the networking namespace of MG (at least I don't know the codebase **at all**).

This is a complete rewrite so a reviewer does not need to know the current implementation.

> I think testing it is honestly the best way to verify the quality of the changes. And the best way to test this is with an actual game. So I'll hold off on merging this and wait for updates from you :) If you think it's kind of functional we can merge and let people know they can try it so we can get better testing coverage and feedback. Thanks for contributing back these improvements @gustavolsson :)

Sounds good :) I think it's ready to be tested in other games now, there are not any important parts missing. Before we click the merge button however, I would like to run a large scale test with friends on the same network and over the internet simultaneously with and without multiple local gamers. I'll keep you updated :)

harry-cpp:
MonoGame.Framework.Net has been moved to https://github.com/MonoGame/MonoGame.Framework.Net so closing this PR.

Issue #6561, opened 2018-11-25
EXT_framebuffer No Longer Available


jcsnider:
I've been creating a mostly SpriteBatch 2d game using the DesktopGL template. After recently upgrading my game from MonoGame 3.6 to 3.7 I started receiving reports that players who could play my game using MonoGame 3.6 but could no longer play using MonoGame 3.7.

One of my players has an old Intel Core 2 Duo with integrated graphics. Framebuffer support was provided via the EXT_framebuffer object. As of this PR (#5993) his machine no longer loads the BindFramebufferDelegate, GenFramebufferDelegate, or any of the other EXT functions which eventually leads to a NoSuitableGraphicsDeviceException when trying to call those functions from PlatformApplyRenderTargets in GraphicsDevice.OpenGL.cs.

By bringing back the old EXT loading code, and code in the Framebufferhelper.cs file I was able to compile a MonoGame dll from the develop branch that he could use. Obviously [these changes](https://github.com/jcsnider/MonoGame/commit/b4c057571d7ef0c82c07d3dbed8e853c8aa0896a) are not suitable for a PR but I was able to prove to myself that these players can still use MonoGame 3.7+ with very minor modifications.

I guess I'm wondering if it was intended to drop support for computers that required the EXT functions (ie 'glBindFramebufferEXT') or if this is a valid issue. If it's valid I can devote more time and attempt to find a real solution for these players that won't break other devices.

Thanks!
JC

CPU: Intel Core 2 Duo 3.00ghz
Memory: 5gb
IGPU: Intel Q45/Q43 Express Chipset
OpenGL Version: 2.1 + EXT_Framebuffer

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows 7 x64

#### What MonoGame platform are you using:
- DesktopGL

mrhelmut:
What you did in your branch is most likely to do the opposite on some systems: break frame buffer extension for OpenGL version where these extensions are core features.

I found the issue though, the problem is here: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/OpenGL.cs#L1388

The ```GL_EXT_framebuffer_object``` check and loading should be outside the ```if (GL.RenderbufferStorageMultisample == null)``` statement.
Right now, the framebuffer object extension gets loaded only if ```glRenderbufferStorageMultisample``` didn't get loaded.

I'm going to push a fix (and clean up a few things).

[...]

PR #6563, opened 2018-11-26
Fixes GL.RenderbufferStorage for devices that use the EXT entry points

jcsnider:
Fixes GL.RenderbufferStorage function for devices that use the EXT entry points. Those devices crash when creating 0 sample render targets due to this function not currently being mapped.

Ref: https://github.com/MonoGame/MonoGame/issues/6561#issuecomment-441478319


Jjagg:
Thanks for the fix @jcsnider!

Issue #6538, opened 2018-11-14
Pipeline tool crashes on startup


svend06:
I'm on macOS 10.14 and use MonoGame 3.7.1708. Error was not there from the start, but now everytime i open the pipeline tool it crashes. Reinstallation, reboot and manually installing pipline with Pipeline.MacOS.pkg doesn't help. This is the error message:

```
System.Reflection.ReflectionTypeLoadException: Exception of type 'System.Reflection.ReflectionTypeLoadException' was thrown.
  at (wrapper managed-to-native) System.Reflection.Assembly.GetTypes(System.Reflection.Assembly,bool)
  at System.Reflection.Assembly.GetTypes () [0x00000] in <ede5cf6930dc4b9a8366f15f6c89bbd0>:0 
  at MonoGame.Tools.Pipeline.PipelineTypes.ResolveAssemblies (System.Collections.Generic.IEnumerable`1[T] assemblyPaths) [0x00045] in <a6bd9c13308c443098175b5c4a6dd6f0>:0 
  at MonoGame.Tools.Pipeline.PipelineTypes.Load (MonoGame.Tools.Pipeline.PipelineProject project) [0x00090] in <a6bd9c13308c443098175b5c4a6dd6f0>:0 
  at MonoGame.Tools.Pipeline.PipelineController.ResolveTypes () [0x00000] in <a6bd9c13308c443098175b5c4a6dd6f0>:0 
  at MonoGame.Tools.Pipeline.PipelineController.OnReferencesModified () [0x00007] in <a6bd9c13308c443098175b5c4a6dd6f0>:0 
  at MonoGame.Tools.Pipeline.CellRefs.Edit (Eto.Forms.PixelLayout control) [0x00042] in <a6bd9c13308c443098175b5c4a6dd6f0>:0 
  at MonoGame.Tools.Pipeline.PropertyGridTable.<Drawable_MouseUp>b__26_0 () [0x00000] in <a6bd9c13308c443098175b5c4a6dd6f0>:0 
  at MonoGame.Tools.Pipeline.PropertyGridTable.Drawable_MouseUp (System.Object sender, Eto.Forms.MouseEventArgs e) [0x0003c] in <a6bd9c13308c443098175b5c4a6dd6f0>:0 
  at Eto.PropertyStore.TriggerEvent[T] (System.Object key, System.Object sender, T args) [0x0000e] in <6db481985cca4510b36c8dbd12852b82>:0 
[...]

PR #6566, opened 2018-11-28
Fix for issue where plugins that reference types from other assemblies cause Pipeline to display the "Invalid / Missing importer" error

timesscar:
Fix for issue where plugins that reference types from other assemblies cause Pipeline to crash and/or display the "Invalid / Missing importer" error

### DIFF
--- a/Tools/Pipeline/Common/PipelineTypes.cs
+++ b/Tools/Pipeline/Common/PipelineTypes.cs
@@ -522,7 +522,7 @@ namespace MonoGame.Tools.Pipeline
                 try
 #endif
                 {
-                    if (!asm.ToString().Contains("MonoGame"))
+                    if (!asm.ToString().Contains("MonoGame") || _pluginAssemblies.ContainsKey(asm.FullName))
                         continue;
 
                     var types = asm.GetTypes();
--- a/Tools/Pipeline/Common/PipelineTypes.cs
+++ b/Tools/Pipeline/Common/PipelineTypes.cs
@@ -367,6 +367,8 @@ namespace MonoGame.Tools.Pipeline
             _processors = null;
             Processors = null;
 
+            _pluginAssemblies = new HashSet<string>();
+
             ImportersStandardValuesCollection = null;
             ProcessorsStandardValuesCollection = null;
         }
@@ -464,7 +466,10 @@ namespace MonoGame.Tools.Pipeline
                     var referencedAssemblies = a.GetReferencedAssemblies();
 
                     foreach (var assembly in referencedAssemblies)
-                        _pluginAssemblies.Add(assembly.FullName);
+                    {
+                        if(!_pluginAssemblies.Contains(assembly.FullName))
+                            _pluginAssemblies.Add(assembly.FullName);
+                    }
 
                     var types = a.GetTypes();
                     ProcessTypes(types);
--- a/Tools/Pipeline/Common/PipelineTypes.cs
+++ b/Tools/Pipeline/Common/PipelineTypes.cs
@@ -171,7 +171,9 @@ namespace MonoGame.Tools.Pipeline
         private static List<ImporterInfo> _importers;
         private static List<ProcessorInfo> _processors;
         private static List<FileSystemWatcher> _watchers;
-        private static HashSet<string> _pluginAssemblies = new HashSet<string>();
+
+        // Dictionary of encountered plugin assemblies and their paths
+        private static Dictionary<string, string> _pluginAssemblies = new Dictionary<string, string>();
         private static string _currentAssemblyDirectory;
 
         public static ImporterTypeDescription[] Importers { get; private set; }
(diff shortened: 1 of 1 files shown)


Jjagg:
@timesscar Can you provide an explanation of what issue this fixes and how?

timesscar:
Hello,  it addresses the issue in https://github.com/MonoGame/MonoGame/issues/6538 which while marked as fixed in dev, I was still able to reproduce the issue.  The exception is thrown at https://github.com/MonoGame/MonoGame/blob/4301eb548b8d4b97fcf8c92df4453a25376d4317/Tools/Pipeline/Common/PipelineTypes.cs#L458 when using a pipeline plugin that references other assemblies.  The technique used here won't work in this case because the assemblies end up in the app domain but dependencies are not resolvable because they are not in the same directory as pipeline.  The technique in https://github.com/MonoGame/MonoGame/blob/4301eb548b8d4b97fcf8c92df4453a25376d4317/Tools/Pipeline/Common/PipelineTypes.cs#L479 is the correct one in this case.  This PR just splits the reflection between appdomain assemblies and plugin assemblies into two functions for clarity and causes the assemblies loaded from plugins to be ignored when loading from the appdomain.  

Note, I do need to add an iteration on this to check to make sure the assembly hasn't been seen first before adding it to the hashset.

Jjagg:
cc @cra0zy

cra0zy:
Hmm, this should have been fixed with overriding the assembly resolver event: https://github.com/MonoGame/MonoGame/blob/develop/Tools/Pipeline/Common/PipelineTypes.cs#L209
(the event above only gets called when the reference cannot be resolved).

Can you provide an example content project that fails with MG 3.8?

timesscar:
Is there a way I can send you the bits?  You'd need my assemblies too.  Alternatively, I can invite you to my VSTS repo and you can build it directly.

timesscar:
Sorry for the iteration spam, but I reworked the resolver to fit into the pattern you already had better and fix an issue I encountered while testing.

cra0zy:
If I understood you correctly, assemblies get loaded correct the first time, but after that they are returned in `new List<Assembly>(AppDomain.CurrentDomain.GetAssemblies())` and so the variable does not get set properly for resolving the 2nd time around?

timesscar:
Yes, that's correct.  The issue I encountered that warranted the further updates was that the N level deep references cause the same issue.  The latest iteration caches the requesting assembly so the children of the children (of the children etc) should be OK too.

timesscar:
I thought about that, but there's extra complexity reloading the other types you need, not just the plugins.  I thought caching the dll locations would be a reasonable middle ground.

scott-lin:
What's the status of this issue? I believe I'm running into it.

I wrote a custom pipeline extension, and when I went to test it, I cannot select my importer / processor type in the tool's drop down because they do not show up. 

Here's what I see instead:

image

I'm using version 3.8.0.423 of Monogame, which should be the latest development available to download at the moment.

Happy to share source / binaries upon request.

cra0zy:
Assembly loading has been reverted to load from filePath, the issue has been resolved in devel.

scott-lin:
Hmm, perhaps I'm doing it wrong, but I don't see this behavior as resolved in the latest version of `develop` branch. Perhaps you can point out what I'm doing wrong.

Steps I took to check:

1. git clone https://github.com/MonoGame/MonoGame.git --branch develop
2. git submodule update --init
3. Run protobuild.exe
4. Opened and built Windows sln
5. Navigated to `Monogame\Tools\Pipeline\bin\Windows\AnyCPU\Debug` and ran `Pipeline.exe`
6. Opened my content project with reference to my custom pipeline extension dll
7. Verified the Importer and Processor drop downs say "Invalid / Missing ...." for my asset

Happy to share source, binaries and content project upon request.

scott-lin:
I attached a debugger and stepped through the Monogame Pipeline code. It is throwing an exception when loading my custom pipeline extension assembly at [PipelineTypes.cs line 451](https://github.com/MonoGame/MonoGame/blob/a50501f5facdbb4e7fdc3d5c22677b8f83d90dc6/Tools/Pipeline/Common/PipelineTypes.cs#L451).

```
System.IO.FileLoadException: Could not load file or assembly 'D:\/Projects\/Game Development\/Monogame.SpriteBox\/PipelineExtension\/bin\/Debug\/Monogame.SpriteBox.PipelineExtension.dll' or one of its dependencies. The given assembly name or codebase was invalid. (Exception from HRESULT: 0x80131047)
File name: 'D:\/Projects\/Game Development\/Monogame.SpriteBox\/PipelineExtension\/bin\/Debug\/Monogame.SpriteBox.PipelineExtension.dll'
   at System.Reflection.AssemblyName.nInit(RuntimeAssembly& assembly, Boolean forIntrospection, Boolean raiseResolveEvent)
   at System.Reflection.RuntimeAssembly.CreateAssemblyName(String assemblyString, Boolean forIntrospection, RuntimeAssembly& assemblyFromResolveEvent)
   at System.Reflection.RuntimeAssembly.InternalLoad(String assemblyString, Evidence assemblySecurity, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean forIntrospection)
   at System.Reflection.RuntimeAssembly.InternalLoad(String assemblyString, Evidence assemblySecurity, StackCrawlMark& stackMark, Boolean forIntrospection)
   at System.Reflection.Assembly.Load(String assemblyString)
   at MonoGame.Tools.Pipeline.PipelineTypes.ResolveAssemblies(IEnumerable`1 assemblyPaths) in D:\Projects\Game Development\MonoGame\Tools\Pipeline\Common\PipelineTypes.cs:line 450
```

I changed the line to use `Assembly.LoadFrom( path )` instead, and the exception no longer occurred. I'm not sure why `Assembly.Load` is being used here. The .Net Framework documentation below suggests the current usage is incorrect in Monogame.

[Assembly.Load(string)](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.load?view=netframework-4.8#System_Reflection_Assembly_Load_System_String_): Loads an assembly given the long form of its name.
[Assembly.LoadFrom(string)](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=netframework-4.8#System_Reflection_Assembly_LoadFrom_System_String_): Loads an assembly given its file name or path.

Since Monogame is using file paths, the latter API seems correct to use.

cra0zy:
Ups, I feel dumb, wanna submit a PR or should I?

Also thanks for pointing out the mistake.

scott-lin:
Hopefully I created the PR correctly. First time working in this repo. https://github.com/MonoGame/MonoGame/pull/6770

SimonDarksideJ:
What was the resulting status with this PR @harry-cpp ?  Is the fix still valid from @timesscar or has it been superseded with recent changes?

SimonDarksideJ:
Parking this PR and likely will close with the introduction of the new Content Project implementation once added

SimonDarksideJ:
Seems this PR is redundant as stated in
https://github.com/MonoGame/MonoGame/pull/6566#issuecomment-505204863

Issue #6027, opened 2017-10-26
GamePadState miss a public constructor. Breaks compatibility with XNA.


nkast:
The API was broken in PR #4970. 

`public GamePadState(Vector2, Vector2 , float , float , params Buttons[])`

Jjagg:
Fixed in #6572.

nkast:
![gamepadstate](https://user-images.githubusercontent.com/3018589/50505233-28023800-0a7b-11e9-9718-cbab730e2dfc.png)

PR #6572, opened 2018-12-07
#6027 Add old interface and unit test for GamePadState

griseus:
I tried to solve a issue that was tagged as "Good First Issue". I would be glad to get feedback.

### DIFF
--- a/MonoGame.Framework/Input/GamePadState.cs
+++ b/MonoGame.Framework/Input/GamePadState.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Xna.Framework.Input
     /// <summary>
     /// Represents specific information about the state of the controller,
     /// including the current state of buttons and sticks.
-    /// 
+    ///
     /// This is implemented as a partial struct to allow for individual platforms
     /// to offer additional data without separate state queries to GamePad.
     /// </summary>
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/Input/GamePadState.cs
```diff
@@ -82,7 +96,7 @@ public GamePadState(GamePadThumbSticks thumbSticks, GamePadTriggers triggers, Ga
         /// <param name="leftTrigger">Left trigger value. This value is clamped between 0.0 and 1.0.</param>
         /// <param name="rightTrigger">Right trigger value. This value is clamped between 0.0 and 1.0.</param>
         /// <param name="buttons"> Array or parameter list of Buttons to initialize as pressed.</param>
-        public GamePadState(Vector2 leftThumbStick, Vector2 rightThumbStick, float leftTrigger, float rightTrigger, Buttons buttons)
+        public GamePadState(Vector2 leftThumbStick, Vector2 rightThumbStick, float leftTrigger, float rightTrigger, params Buttons[] buttons)
```
Jjagg: Please remove the `params` here. While it is mentioned like this in #6027 the XNA constructor did not have a `params` argument (just an array) and we want to prevent the implicit allocation.

griseus: Updated the pull request. Thanks for the feedback!

Jjagg:
Thanks @griseus :)

MonoGame.Framework/Input/GamePadState.cs
nkast: The parameter needs to be `params Buttons[] buttons`. 
In XNA you could create a GamePadState() with any number of buttons parameters. 
e.g. `new GamePadState(leftTS, rightTS, leftT, rightT, Buttons.DPadDown, DPadRight.Buttons)`

Issue #6522, opened 2018-11-04
Loading SDL fails when a # is in the directory path


SpacialCircumstances:
Starting a Monogame game fails with the exception: "Failed to load SDL library." when starting the game from a folder where the path contains a "#" character.

This is caused by this line: https://github.com/MonoGame/MonoGame/blob/acada356bbd95b15cd02a99419764e307a3f3672/MonoGame.Framework/SDL/SDL2.cs#L21
where the path to the sdl library is determined.

I´ve got my Monogame project in a folder called E:\C#\MyGame
`(new Uri(typeof(Sdl).Assembly.CodeBase)).LocalPath` will return "E:\C" as a path, cutting the "#" and everything after it. Therefore, the search directory is determined to be "E:\" and the sdl library can not be found.

To reproduce, simply create a project using the default template in a folder that has a "#" in its name.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7 (Release from the VS installer), but the issue is present in the current development build

#### What operating system are you using:
- Windows 10

#### What MonoGame platform are you using:
- DesktopGL

omniboystudio:
Thanks! I had the bad idea of naming my folder C# and had the same issue. Thank you, thank you! Now works great

tomspilman:
@cra0zy - could we fix this?

Jjagg:
@tomspilman We need to escape the Uri: https://docs.microsoft.com/en-us/dotnet/api/system.uri.escapeuristring?view=netframework-4.7.2
Then unescape afterwards. 

Seems a bit of a roundabout way to do it though, there's likely a nicer solution.

cra0zy:
Yea, this one is easy to fix.

griseus:
Anything prevents us from using Assembly.Location? Seems to do the trick: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location?view=netframework-4.7.2

cra0zy:
@griseus Nothing prevents us, feel free to submit a PR.

Jjagg:
Fixed in #6573.

PR #6573, opened 2018-12-08
Remove Uri in SDL2

griseus:
This should fix the bug #6522


Jjagg:
Awesome @griseus! Thanks for helping out :)

PR #6577, opened 2018-12-08
Remove trailing NUL chars in registry font name

### LINKED ISSUE #4061: MGCB can't handle malformed Fonts in the Windows Registry
I tried to add a simple font to my project and got errors while building it.
I tracked the issue down to the following [line](https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs#L208)

After analyzing my registry, i saw that some of my entries have invalid data. All of them have completely uppercase Names, as you can see in the image.

The data in these registry entries look like <code>ARIAL.TTF\00\0\0\0\0\0</code> 
As craftworkgames pointed out, these could be some kind of line endings. 

I don't know what made (corrupted?) the data look like this, but i just updated my machine to Windows 10.

Replacing the invalid characters could solve this, because these fonts exist in my Fonts folder and other programs can access it.

corrupted

Here is the corresponding [Stackoverflow Question](http://stackoverflow.com/questions/31886866/invalid-path-while-processing-font-with-mgcb/31887410)

The stack trace:

<pre><code>
default.spritefont: error: Processor 'FontDescriptionProcessor' had unexpected failure! System.ArgumentException: Illegales Zeichen im Pfad.
bei System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional) bei System.IO.Path.IsPathRooted(String path)
bei Microsoft.Xna.Framework.Content.Pipeline.Processors.FontDescriptionProcessor.FindFontFileFromFontName(String fontName, String fontDirectory)
bei Microsoft.Xna.Framework.Content.Pipeline.Processors.FontDescriptionProcessor.Process(FontDescription input, ContentProcessorContext context)
bei Microsoft.Xna.Framework.Content.Pipeline.ContentProcessor`2.Microsoft.Xna.Framework.Content.Pipeline.IContentProcessor.Process(Object input, ContentProcessorContext context)
bei MonoGame.Framework.Content.Pipeline.Builder.PipelineManager.ProcessContent(PipelineBuildEvent pipelineEvent)</code></pre>

How to reproduce:
- Look inside your registry at SOFTWARE\Microsoft\Windows NT\CurrentVersion\Fonts
- Search for a completely uppercase Entry
- Try to build with this font.

I'm sorry for my bad English. When i forgot something, please let me know.

Jjagg:
Fixes #4061.

Removes all characters from the first NUL character onward in font names retrieved from the Windows registry.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
@@ -233,6 +233,13 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
                     if (font.StartsWith(name, StringComparison.OrdinalIgnoreCase))
                     {
                         var fontPath = key.GetValue(font).ToString();
+
+                        // The registry value might have trailing NUL characters
+                        // See https://github.com/MonoGame/MonoGame/issues/4061
+                        var nulIndex = fontPath.IndexOf('\0');
+                        if (nulIndex != -1)
+                            fontPath = fontPath.Substring(0, nulIndex);
+
                         return Path.IsPathRooted(fontPath) ? fontPath : Path.Combine(fontDirectory, fontPath);
                     }
                 }


Jjagg:
I think the way I did it is more robust in case other characters end up after a NUL.

cra0zy:
> I think the way I did it is more robust in case other characters end up after a NUL.

I was just thinking of removing any null characters from the string, as I see you never want to have a null character in it and your approach only removes 1 null character from the beginning if found. Now if you want to remove just trailing null chars you can also use `.Trim`, or `.TrimStart` if you just want to remove them from the beginning.

Do say something if you want to keep your current approach, I don't mind merging it.

Jjagg:
Ah, yes I worded this poorly. I intended to trim everything from the first NUL character onward, not just trailing NUL characters. So I'd like to keep this as is.

cra0zy:
> Ah, yes I worded this poorly. I intended to trim everything from the first NUL character onward, not just trailing NUL characters. So I'd like to keep this as is.

Oh, that makes sense now, I didn't read up your code but the title, and saw an indexOf... mething!

PR #6586, opened 2018-12-10
Add FromFile to Texture2D and SoundEffect

### LINKED ISSUE #5761: Missing Texture2D.FromFile ?
In upgrading the ZiggyWare blog posts, I came across the good old Texture2D.FromFile method.

I know this was a XNA 1 artifact but this seems a much easier function for users not using the content pipeline to use that the current Texture2D.FromStream

Any reasons why we couldn't add such a method ?

This would simply wrap the existing FromStream function to actually load the texture from a stream safely and return the resultant Texture2D.

Jjagg:
Fixes #5761.

These are very simple methods that don't take maintenance and offer some convenience to users I think nice to have these. Should these be conditionally compiled only for DesktopGL and WindowsDX?

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.cs
@@ -197,7 +197,6 @@ namespace Microsoft.Xna.Framework.Audio
         /// </item>
         /// </list>
         /// </remarks>
- 
         public static SoundEffect FromFile(string path)
         {
             if (path == null)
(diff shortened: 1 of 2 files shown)


Jjagg:
@cra0zy Should we add these?

cra0zy:
Originally I was not gonna merge anything that is a feature request, but since Tom said we can merge the small features, I guess I don't see why we should hold off on this.

nkast:
Shouldn't this use TileContainer to get a stream?

PR #6589, opened 2018-12-10
GraphicsDevice.DirectX: expose the D3D device context

filnet:
introduces a public readonly ContextHandle property that exposes the D3D device context
similar to the existing Handle property that expose the D3D device.

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/EffectPass.cs
+++ b/MonoGame.Framework/Graphics/Effect/EffectPass.cs
@@ -105,6 +105,10 @@ namespace Microsoft.Xna.Framework.Graphics
                     device.SetConstantBuffer(ShaderStage.Pixel, c, cb);
                 }
             }
+            else
+            {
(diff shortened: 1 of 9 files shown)


Jjagg:
Hey @filnet! Thanks for this contribution.

I think it's useful to somehow expose native handles like for the graphics device and native window handles, but we should think about how to expose them. Seperate handles like this is one way to do it. Another is to provide a `PlatformData` property in the `GraphicsDevice` and `GameWindow` classes (and wherever this may be applicable) that has subclasses for each platform type and contains the data relevant for that platform. The benefit of the second approach is that users don't need conditional compilation and can just check the type of the `PlatformData` instance.

That said I'm fine with this change for now, because we already expose the device handle in a similar way. We may want to change the API in the future.

cc @tomspilman

Jjagg:
We should have a discussion about exposing native handles. I'm closing this for now.

PR #6594, opened 2018-12-14
[WinDX] Avoid freezing while moving/resizing window

MichaelDePiazzi:
As discussed in issue #5656, this solution works around the game freezing while the window is being moved or resized.

This logic has been in the public release of my game for several months now and there doesn't seem to be any issues with it. It has also proven to be very useful for video playback as the timing for that seems to mess up if updates fall behind.

Note that update timing won't be accurate while this is engaged. But the important thing is that updates are still happening and not falling behind.

_Note that this doesn't resolve the related issue (#5656) as it only fixes the problem on the Windows DirectX build. The DesktopGL build will need a different fix._

### DIFF
--- a/MonoGame.Framework/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Windows/WinFormsGameWindow.cs
@@ -43,6 +43,9 @@ namespace MonoGame.Framework
         // true if window position was moved either through code or by dragging/resizing the form
         private bool _wasMoved;
 
+        private bool _isResizeTickEnabled;
+        private readonly System.Timers.Timer _resizeTickTimer;
+
         #region Internal Properties
 
         internal Game Game { get; private set; }
@@ -156,9 +159,13 @@ namespace MonoGame.Framework
             Form.MouseEnter += OnMouseEnter;
             Form.MouseLeave += OnMouseLeave;            
 
+            _resizeTickTimer = new System.Timers.Timer(1) { SynchronizingObject = Form, AutoReset = false };
+            _resizeTickTimer.Elapsed += OnResizeTick;
+
             Form.Activated += OnActivated;
             Form.Deactivate += OnDeactivate;
             Form.Resize += OnResize;
+            Form.ResizeBegin += OnResizeBegin;
             Form.ResizeEnd += OnResizeEnd;
 
             Form.KeyPress += OnKeyPress;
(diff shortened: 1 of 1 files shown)


MichaelDePiazzi:
P.S. Credit goes to @squarebananas for originally coming up with this workaround 👍

Jjagg:
> Note that update timing won't be accurate while this is engaged. But the important thing is that updates are still happening and not falling behind.

This is because of the 1 ms timer duration? Do you think we could improve this? I'm good with merging this as is, though.

MichaelDePiazzi:
> This is because of the 1 ms timer duration?

No, it's because `System.Timers.Timer` only has a ~15.6ms (64Hz) resolution. The 1ms interval is just me telling it to fire as soon as it can. _([Apparently](https://stackoverflow.com/a/52521911) the .NET runtime timer queue is fixed at 64Hz and it ignores the system timer resolution)_

> Do you think we could improve this?

Yes, probably. I did experiment with a different method which ran a tighter loop on a separate thread. I still ended up throwing a `Sleep(1)` in it though to stop it from being too intensive. It's a bit hacky and could be improved, but here's the guts of it fwiw:
```c#
private async void OnResizeBegin(object sender, EventArgs e)
{
    _isResizeTickEnabled = true;
    while (_isResizeTickEnabled && (Form != null) && !Form.IsDisposed)
    {
        UpdateWindows();
        Game.Tick();
        await Task.Run(() => Thread.Sleep(1));
    }
}

private void OnResizeEnd(object sender, EventArgs eventArgs)
{
    _isResizeTickEnabled = false;
...
```
But I didn't end up going with it because there are allocations from `Task.Run`. Plus I was a bit worried that there'd be some cross-thread issues. And I didn't want to complicate the code further to try deal with that. Also I just couldn't afford to spend any more time on it.

So especially given this is an edge case, it wasn't worth it in my opinion and a little potential jitter while the user was moving/resizing the window was acceptable.

> I'm good with merging this as is, though.

Given that this logic has been well tested and is working well, I feel that this would be best. And if someone wants to take a crack at improving it later, then they can submit another PR.

Jjagg:
> So especially given this is an edge case, it wasn't worth it in my opinion and a little potential jitter while the user was moving/resizing the window was acceptable.

I agree with this. At least this is a good improvement from current behavior. Thanks @MichaelDePiazzi! Merging :)

PR #6606, opened 2018-12-28
No alloc to marshal UTF-8 to string

Jjagg:
There's no need to create an intermediate array when marshalling UTF-8 to a .NET string.

### DIFF
--- a/MonoGame.Framework/Utilities/InteropHelpers.cs
+++ b/MonoGame.Framework/Utilities/InteropHelpers.cs
@@ -18,18 +18,13 @@ namespace MonoGame.Utilities
             if (handle == IntPtr.Zero)
                 return string.Empty;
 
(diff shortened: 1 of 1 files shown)


tomspilman:
@Jjagg - I vaguely remember having some weird issue that required the allocation here.  @cra0zy ?

Jjagg:
The used overload for GetBytes was only introduced in .NET Framework 4.6. We'll have to postpone this until we move to .NET Standard.
- [Docs for 4.5.2](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=netframework-4.5.2)
- [Docs for 4.6](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=netframework-4.6)

Issue #6534, opened 2018-11-11
Cannot jump to different position in audio


ninjamida:
MonoGame either lacks, or at least its documentation lacks any reference to (as far as I can tell), any method to set the position (time-wise) in an audio file smoothly.

An EXTREMELY common use case for this in gaming is to have music that has an intro that's played once, followed by a looping segment.

The closest thing to a workaround, other than using a third-party audio library (generally involving hefty costs and sacrifices to platform support) or splitting a single file into multiple ones (which will still likely have a gap in playback anyway), is to play it as a Song via MediaPlayer, then when the end of the loop is reached, run MediaPlayer.Play(Song, MediaPlayer.Position - LoopLength) <--- didn't bother to include typecasts etc in this sample code. Unfortunately, this results in a VERY noticable silence gap.

Furthermore, an approach to deal with this that specifically DID work on XNA, does not work in MonoGame - using a WAV file via XACT that has loop points specified. I attempted to create such a file, and indeed it loops perfectly in XACT itself, but MonoGame code completely ignores this looping.

I can understand that reproducing every detail of XACT correctly may be low priority, but MonoGame should provide some means via which we can at least set the playback position of a Song or SoundEffectInstance or the like, and jump to it smoothly - if it's up to the user to actually use that feature to handle looping or whatever, that's understandable, but the total lack of any way to control the position at all (other than the MediaPlayer workaround mentioned above) seems like a major oversight.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

ninjamida:
> As for the silence when using the Play overload for MediaPlayer, can you try setting MediaPlayer.PlayPosition instead and see if that works?

MediaPlayer.PlayPosition is a read-only property.

Jjagg:
> MediaPlayer.PlayPosition is a read-only property.

Oops, yes seeking with that property is only supported on iOS and Android, sorry. 

> I was ultimately able to achieve what I wanted to using a DynamicSoundEffectInstance and manual loading of WAV file data

Yeah, we definitely don't want it to be that hard. We should use the OpenAL Soft extension to support loop points as Tom mentioned in [a comment](https://github.com/MonoGame/MonoGame/issues/6369#issuecomment-406380240) on a related issue.

ninjamida:
[...]

Issue #5186, opened 2016-09-25
OpenAL ignores loop points


KimikoMuffin:
As of the latest commit, if you initialize a SoundEffect with loop points in projects with OpenAL-based audio (specifically WindowsGL, I can't really test it with other systems), the loop points are straight-up ignored. I know that in 3.5.1 it threw a NotSupportedException, which seems to be a more "correct" way of dealing with that kind of thing ...

tomspilman:
We need to get a test sound with loop points that we can add to our unit tests and ensure we don't break this again.

KimikoMuffin:
Threw this together: [Bach Musette Loop.zip](https://github.com/MonoGame/MonoGame/files/491715/Bach.Musette.Loop.zip), it's Bach's Musette In D Major with an intro-bit added. Loop start: 88200 samples, loop length: 352800 samples for the full thing, 176400 if you're in a hurry.

Issue #6369, opened 2018-07-13
SoundEffect loopStart and loopLength do nothing


Jesuszilla:
I'm trying to loop a PCM stream (decoded from MP3 using MP3Sharp) and the loopStart and loopLength parameters do not loop the track after being decoded.

```
bgm = new SoundEffect(decoded.ToArray(), 0, (int)decoded.Count, mp3Dec.Frequency, (AudioChannels)mp3Dec.ChannelCount, loopStart, loopEnd - loopStart);
```

Instead, it just goes right past the loop point and plays the rest of the track. I know this because it gets to the end of the song, then loops around, rather than using the specified loop points.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

tdeeb:
OpenAL SoundEffects don't actually use these values for some reason. Most of them route to [PlatformInitializePcm](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs#L46), which ignores them completely. This contrasts to XAudio's constructors which call the [CreateBuffers](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Audio/SoundEffect.XAudio.cs#L245) method, which does look to utilize them in some form.

What would it take to fix this in OpenAL?

Jesuszilla:
OpenAL does have alSourceQueueBuffers, which I believe is what DynamicSoundEffectInstance uses underneath. It could do what I do for my MP3/OGG looping (due to this not working) and split the PCM data on the sample boundaries and constantly feed it the loop region. It needs to clear the previously played buffers to prevent memory leaks, though.

tomspilman:
There must be a OpenAL feature for this.   It really shouldn't require streaming of a sound asset.

Jesuszilla:
The way I see it, there's three ways to go about it:

1. Set a timer to go off exactly at the loop start period, then start playing the loop portion (inaccurate; tried similar approach in C# and there were still early fires or delays, though maybe there's a more accurate, lower-level timer that can be used).
2. Constantly poll the state of the alSource. This would result in a busy loop, which is quite undesirable. May not be 100% accurate, either.

[...]

PR #6607, opened 2018-12-28
Support SoundEffect loop points on platforms using OpenAL

kimimaru4000:
This PR uses the `loopStart` and `loopLength` parameters in SoundEffect constructors and passes them through `alBufferiv` using the `AL_LOOP_POINTS_SOFT` extension to set the loop points in a sound.

Related issues: #6534, #5186

I tested the example clip provided in #5186, but the XNB read the loopStart as 0. I'm not sure how to properly test this feature.

### DIFF
--- a/MonoGame.Framework/Audio/OALSoundBuffer.cs
+++ b/MonoGame.Framework/Audio/OALSoundBuffer.cs
@@ -56,7 +56,7 @@ namespace Microsoft.Xna.Framework.Audio
                 ALHelper.CheckError("Failed to fill buffer.");
             }
 
-            if (loopStart >= 0 && loopLength > 0)
+            if (OpenALSoundController.GetInstance.SupportsLoopPoints && loopStart >= 0 && loopLength > 0)
             {
                 //Loop end is loopStart + loopLength
                 int[] loopData = new int[2] { loopStart, loopStart + loopLength };
--- a/MonoGame.Framework/Audio/OALSoundBuffer.cs
+++ b/MonoGame.Framework/Audio/OALSoundBuffer.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Xna.Framework.Audio
 			set;
 		}
 
-        public void BindDataBuffer(byte[] dataBuffer, ALFormat format, int size, int sampleRate, int sampleAlignment = 0)
+        public void BindDataBuffer(byte[] dataBuffer, ALFormat format, int size, int sampleRate, int sampleAlignment = 0, int loopStart = 0, int loopLength = 0)
         {
             if ((format == ALFormat.MonoMSAdpcm || format == ALFormat.StereoMSAdpcm) && !OpenALSoundController.GetInstance.SupportsAdpcm)
                 throw new InvalidOperationException("MS-ADPCM is not supported by this OpenAL driver");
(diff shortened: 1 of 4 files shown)


mrhelmut:
Oh, interesting, looks like the MacOS tests are not using OpenAL-Soft but instead the system OpenAL implementation? (which most obviously wouldn't have the extension)

Weird, since the tests should be using the OpenAL-Soft dylib from here: https://github.com/MonoGame/MonoGame.Dependencies/tree/master/openal-soft/MacOS/Universal

Give me a few days to test if this runs correctly on MacOS with the correct OpenAL-Soft binary. If it passes, it's a test framework issue.

tdeeb:
@mrhelmut Any updates on the macOS test?

mrhelmut:
I haven't been able to test, but it should work with the correct OpenAL-Soft dylib. Yet, the Mac tests are designed to work with the system library which isn't a compliant implementation.

I see a couple issues:
- the test project has a dllmap to indicate to Mono that it should use the libopenal.1.dylib (which is cool), but the file isn't added to the project so it never gets use; ```MonoGame.Tests.XNA.csproj``` should be updated to have the Mac and Linux binaries to be outputed to their bin folder;
- nice to have: an extension check to the OpenAL initiatlization so that we can skip the ```ALBufferi.LoopSoftPointsExt``` call if the extension is absent

I guess that adding an extension check would be cool either ways.

Jjagg:
> MonoGame.Tests.XNA.csproj should be updated to have the Mac and Linux binaries to be outputed to their bin folder; 

That's not the right project file. The other test projects are generated by Protobuild. The extension check would be really nice.

tdeeb:
I added the extension check, but the macOS test still failed. It's saying that it has the extension but fails to set the loop points; am I missing something?

mrhelmut:
The tests are actually working for most formats, but are failing for IMA/ADPCM.
This format is only tested on OpenAL platforms, and I believe that loop points will fail on any platform with IMA/ADPCM involved (it's not a Mac specific issue).

We should set ```SupportsLoopPoints``` to false when this format is used. It looks to be a limitation of OpenAL-Soft.

It may also fail for other formats that the test framework can't reach yet: MS-ADPCM, and float format are untested so far.

A good reminder that we should also update OpenAL-Soft, these limitations may have been fixed.

MonoGame.Framework/Audio/OALSoundBuffer.cs
```diff
@@ -56,6 +56,29 @@ public void BindDataBuffer(byte[] dataBuffer, ALFormat format, int size, int sam
...
+            if (format == ALFormat.MonoIma4 || format == ALFormat.StereoIma4
+                || format == ALFormat.MonoMSAdpcm || format == ALFormat.StereoMSAdpcm)
+            {
+                formatLoopEnabled = false;
+            }
+
+            if (OpenALSoundController.GetInstance.SupportsLoopPoints && formatLoopEnabled && loopStart >= 0 && loopLength > 0)
+            {
+                //Loop end is loopStart + loopLength
+                int[] loopData = new int[2] { loopStart, loopStart + loopLength };
```
Jjagg: @tdeeb Could you do a `stackalloc` here? The function will need to be marked `unsafe`, but that's fine.

tdeeb:
I'm still unsure how to test this feature. @Jesuszilla I saw the issue you filed, #6369. Would you be able to provide a test clip or test this locally and see if it works as expected?

Jesuszilla:
> 
> 
> I'm still unsure how to test this feature. @Jesuszilla I saw the issue you filed a while back, #6369. Would you be able to provide a test clip or test this locally and see if it works as expected?

Certainly! Is this in the latest NuGet package, or will I have to install from elsewhere?

tdeeb:
The artifacts on the [TeamCity](http://teamcity.monogame.net/viewLog.html?buildId=77947&buildTypeId=MonoGame_DevelopWin&tab=artifacts#!1kn3m7xif5nhs) page can be downloaded. If you prefer to build from source, you can build from the `OALLoopPoints` branch. Thank you!

SimonDarksideJ:
Any chance you can resolve the merge conflicts and re-test this PR @kimimaru4000 , If you have any issues getting the test artifacts, let me know.

mrhelmut:
Not sure if it's worth the effort with OpenAL being planned to be replaced by FAudio.

SimonDarksideJ:
Agreed, with the change in audio platforms, this will likely no longer be required, closing.

Issue #6241, opened 2018-03-11
MGCB doesn't clean empty project


tomspilman:
This is a minor issue... but one we probably should support correctly.

 - Create a new empty Pipeline project.
 - Add one content item to it.
 - Build the content.
 - Undo adding the content item leaving an empty project.
 - Build content.

At the end you would expect the intermediate `.mgcontent` files and the `.xnb` to be cleaned, but they are not and still exist.

It suggests to me MGCB.exe is taking an "early out" when there are no files and fails to clean the previous content.

tdeeb:
I'm looking into this, and the issue lies in `Program.cs` of MGCB with the line `if (content.HasWork)`, which maps out to:
```
public bool HasWork
{
    get { return _content.Count > 0 || _copyItems.Count > 0 || Clean; }    
}
```

Commenting out the `if` statement makes it work, but I'm unsure of any other issues this may cause since I'm new to MGCB.

tomspilman:
@tdeeb 

Looks like @cra0zy added the `HasWork` check in https://github.com/MonoGame/MonoGame/pull/5980.  Maybe he can enlighten us as to why he added it as it wasn't mentioned in the PR.

IMO... we should remove the `if (content.HasWork)` check as it doesn't seem all that useful to me.

PR #6615, opened 2019-01-02
Fixed MGCB Build not cleaning empty projects

tdeeb:
This PR fixes an issue where the Build action for empty pipeline projects outputs nothing and doesn't clean them. It also adds an error message if the debugger attempts to run on a Mono platform it's not implemented on.

Related issue: #6241


Jjagg:
Thanks @tdeeb :)

Issue #6613, opened 2019-01-01
Basic Effect is Nothing


TheFatTurtle:
#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

TheFatTurtle:
When I try to work with models and create a New ModelMeshPart and want to give it as effect( MeshPart.Effect = Basic Effect) I get the message MeshPart.Effect Is Nothing
I can't change the Effect.

Jjagg:
We'll need more information for this, because I don't understand what's up.

@TheFatTurtle Could you set up a project that reproduces this error?

TheFatTurtle:
i made a project.
in this project i initialized a new model with the name ModelA
to make a model i need a modelpart.
so i defined modelpartA as new modelpart
i defined basiceffectA as basiceffect
i tried to set the effect of modelparta  to basiceffectA:
                 modelparta.effect = basiceffectA 
this is were i got the message: "NullReferenceExcepion: meshpart.effect is nothing"

Jjagg:
Building a model at runtime is not supported. The `ModelMeshPart` constructor should not be public. There's no way to add meshes to a model at runtime.

If I got this right, the exception occurs because `ModelMeshPart` assumes its parent is not null, which of course it is if you create one using the default constructor.

TheFatTurtle:
Okay. Thanks for answering so quick

PR #6620, opened 2019-01-04
Make ModelMeshPart constructor internal

Jjagg:
This constructor is not exposed in XNA and creating a ModelMeshPart without attaching it to a model causes issues, see #6613. 

I also removed the non-doc comments that were copied from XNA docs.

### DIFF
--- a/MonoGame.Framework/Graphics/ModelMeshPart.cs
+++ b/MonoGame.Framework/Graphics/ModelMeshPart.cs
@@ -1,13 +1,11 @@
-﻿using System;
+using System;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
 	public sealed class ModelMeshPart
 	{
-		// Summary:
-		//     Gets or sets the material Effect for this mesh part. Reference page contains
-		//     code sample.
         private Effect _effect;
+
         public Effect Effect 
         {
             get 
@@ -44,39 +42,18 @@ namespace Microsoft.Xna.Framework.Graphics
             }
         }
 
-		//
-		// Summary:
-		//     Gets the index buffer for this mesh part.
 		public IndexBuffer IndexBuffer { get; set; }
 
-		//
-		// Summary:
-		//     Gets the number of vertices used during a draw call.
 		public int NumVertices { get; set; }
 
-		//
-		// Summary:
-		//     Gets the number of primitives to render.
 		public int PrimitiveCount { get; set; }
 
-		//
-		// Summary:
-		//     Gets the location in the index array at which to start reading vertices.
 		public int StartIndex { get; set; }
 
-		//
-		// Summary:
-		//     Gets or sets an object identifying this model mesh part.
 		public object Tag { get; set; }
 
-		//
-		// Summary:
-		//     Gets the vertex buffer for this mesh part.
 		public VertexBuffer VertexBuffer { get; set; }
 
-		//
-		// Summary:
-		//     Gets the offset (in vertices) from the top of vertex buffer.
 		public int VertexOffset { get; set; }
 
 		internal int VertexBufferIndex { get; set; }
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @Jjagg !

robintheilade:
The ModelMeshPart is actually added to a model at runtime, see
https://github.com/MonoGame/MonoGame/blob/889c9f53e57318f7eb49f9b764267043a08a1c5a/MonoGame.Framework/Content/ContentReaders/ModelReader.cs#L141

With the ModelMeshPart constructor being internal it is no longer possible to make a custom ModelReader without using reflection to instantiate the ModelMeshPart.

tomspilman:
@Jjagg - Can you investigate the right fix here?

vpenades:
@Jjagg It seems making the constructor internal breaks my glTF loader library, since it relies on creating ModelMeshPart objects at runtime to construct an animatable model.

I understand the reasoning behind making the constructor internal, but then, I would need some kind of workaround, at least until monogame if finally able to handle glTF files from the content pipeline side.

willmotil:
As you know dx xtk is walborn and hargraves. So ... details here.

https://github.com/Microsoft/DirectXTK/wiki/ModelMeshPart

**Initialization**
ModelMeshPart instances are typically created by a Model loader along with the ModelMesh instances that contain the submeshes.

**Data**
All members of ModelMeshPart are public to facilitate writing custom model loaders and sophisticated rendering solutions. Care should be taken ,,

This information is used by ModelMeshPart::CreateInputLayout for setting up custom effect overrides and ModelMeshPart::ModifyEffect when changing the effect instance used by the submesh.


> Simplicity vs Simplexity.    ~ i first heard this term from reading a blog by Anders Hejlsberg.
> 
> Simply dis-allowing a person who doesn't know what he is doing to make a mesh part make the constructor internal. 
> This means a very complicated way of forcing someone who does know what he is doing to go thru hoops.

All this extra because its not public? 
Obviously its not going to work if you don't add it to a mesh part. So we are making it complicated to use restrictive and aggrivating so someone doesn't try to use something that they don't know how to use because it wont work if they try to use it wrong ?    new ModelMeshPart();

willmotil:
@Jjagg  can we roll this change back for now?

There isn't much use in having a proper xna identical matchup to the old xna model class when.

1) It's been public for years it wasn't a big enough problem in all that time to notice.
2) It's outdated anyways.
3) I doubt many people are knowledgeable enough to do this work on the glb gtlf format and it's needed.
4) That constructor can be made internal again later.

Jjagg:
Yeah, that's fine by me. If anyone can set up a PR I'll merge it. 
Should add to the docs how to properly use ModelMeshPart though, because the reason it was deleted in the first place was because you need to initialize it properly to avoid exceptions.

vpenades:
@Jjagg Although I've already added an alternative solution to the gltf loader, as @willmotil says it's a hack, since I'm essentially replicating monogame's code in my library.

I'm not against making it private in the future, but unless there's an alternative, it's best to be pragmatic.

An additional suggestion to keeping the constructor public for a while would be to add the [Obsolete] attribute.

I'm in crunch time right now, so if anyone else could....

Jjagg:
Yes, of course. It's not going to be made private again without an alternative. Agree with marking it Obsolete.

limbusdev:
So whats the best workaround now? I am assembling a Model with a custom model importer and this broke my code. Any good hints :) ?

robintheilade:
> So whats the best workaround now? I am assembling a Model with a custom model importer and this broke my code. Any good hints :) ?

Might not be the best workaround, but it is easy: create an instance using reflection for now and undo it when the change is released.

The only other workaround I can come up with right now is to use a custom build of Monogame, but it ain't that easy.

Personally I would go for the reflection workround.

limbusdev:
Thanks for answering. So, maybe the better question is: What is the correct way to write a custom model loader?

Issue #6619, opened 2019-01-03
DesktopGL emulates DX9 half pixel offset


roy-t:
#### What operating system are you using:
- Windows, DesktopGL

There is a rendering difference between the DirectX and OpenGL versions of MonoGame when using the BasicEffect. Because the half pixel offset for OpenGL is not taken into account. While this is taken into account for the SpriteBatch:
https://github.com/MonoGame/MonoGame/blob/faea8a6a89504673e2bb7e435f0da8cc513d8c30/Mono
Game.Framework/Graphics/GraphicsDevice.OpenGL.cs#L272

This means that anything rendered with the BasicEffect shader is slightly blurry on OpenGL. For example see these two screenshots:

OpenGL:
![text_a](https://user-images.githubusercontent.com/623634/50658255-11873f80-0f99-11e9-8789-2a56e4bdbc39.PNG)

DirectX:
![text_c](https://user-images.githubusercontent.com/623634/50658257-14823000-0f99-11e9-9809-9ce415e9e2d8.PNG)

I think we should give the user some way to know that the half pixel offset should be set by themselves or change the OpenGL shader for the BasicEffect to fix this.

What do you think? I can make a PR once I've gathered more input :).

mrhelmut:
We could use preprocessing directives to put OpenGL specific stuff in .fx files with ```#ifdef OPENGL``` and rebuild ```BasicEffect.ogl.mgfxo```?

(ImGUI FTW)

Jjagg:
Hold up, there is already a half pixel offset being applied for OpenGL and not for DX. The current behavior is completely wrong, but there's an explanation for how it got like this, from my understanding:

DX9 is the weird one out. All other graphics APIs do it 'right'. XNA didn't. It applied a half pixel offset for SpriteBatch. So MG technically needs the half-pixel offset on all platforms that do **not** use DX9 if we want to be XNA compatible since XNA uses DX9. The DX backend doesn't have the offset applied I think because it evolved from DX9 and it was forgotten, though it should to be XNA compatible (looking back this is part of why #5144 failed). XNA applies half a pixel offset in SpriteBatch so DX behavior was consistent there which we prove in testing. For OpenGL the half pixel offset is applied to every effect that goes through MojoShader. Somewhere along the line someone must've thought that OpenGL spritebatch rendering wasn't consistent with DX, due to that half pixel offset for every effect, so a reverse half pixel offset to spritebatch was added.

[...]

PR #6621, opened 2019-01-04
Add UseStandardPixelAddressing option

Jjagg:
This implements an option to enable/disable half pixel offset as discussed in #6619. 

Currently this does not yet actually apply the half pixel offset for MG DX (behavior isn't changed). That's a bit more intricate to do because we have to modify 2MGFX to support the offset in user-written vertex shaders.

This does implement the option for OpenGL platforms. Default behavior did not change, but the `GraphicsDeviceManager.UseStandardPixelAddressing` flag can be set to `true` to render like the DX backend does (different from XNA).

### DIFF
--- a/Test/Framework/Graphics/GraphicsDeviceTestFixtureBase.cs
+++ b/Test/Framework/Graphics/GraphicsDeviceTestFixtureBase.cs
@@ -51,6 +51,11 @@ namespace MonoGame.Tests.Graphics
         [SetUp]
         public virtual void SetUp()
         {
+#if !XNA
+            // We enable the half-pixel offset for XNA compatibility
+            GraphicsDeviceManager.UseStandardPixelAddressing = false;
+#endif
+
             game = new TestGameBase();
             gdm = new GraphicsDeviceManager(game);
             // some visual tests require a HiDef profile
--- a/Documentation/config.xml
+++ b/Documentation/config.xml
@@ -58,6 +58,8 @@
         <topic id="tvOS" name="tvOS" filename="tvOS.md"/>
         <topic id="UWP" name="UWP" filename="UWP.md"/>
     </topic>
+
+    <topic id="Porting_from_XNA" name="Porting an XNA game to MG" filename="porting_xna.md">
     
     <topic id="External_Links" name="External Links" filename="links.md"/>
 
--- a/Documentation/config.xml
+++ b/Documentation/config.xml
@@ -59,7 +59,7 @@
         <topic id="UWP" name="UWP" filename="UWP.md"/>
     </topic>
 
-    <topic id="Porting_from_XNA" name="Porting an XNA game to MG" filename="porting_xna.md">
+    <topic id="Porting_from_XNA" name="Porting an XNA game to MG" filename="porting_xna.md"/>
     
     <topic id="External_Links" name="External Links" filename="links.md"/>
 
--- a/Documentation/porting_xna.md
+++ b/Documentation/porting_xna.md
@@ -25,6 +25,11 @@ referred to as the half-pixel offset.
 MonoGame supports replicating XNA behavior (currently only on OpenGL platforms) by setting
 the `UseStandardPixelAddressing` flag in `GraphicsDeviceManager` to `false`. This flag is
 set to `true` by default to encourage users to use the modern style of pixel addressing.
+DirectX platforms will ignore setting the `UseStandardPixelAddressing` flag and will
+always render with a half pixel offset compared to XNA. This is usually not noticeable.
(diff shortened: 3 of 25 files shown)


Jjagg:
#6623 should get merged before this one.

Jjagg:
This is ready for review @cra0zy @tomspilman @roy-t @mrhelmut. 
This PR changes nothing for DX platforms and by default nothing on GL platforms. However users can make rendering on GL platforms consistent with DX by setting the `UseStandardPixelAddressing` flag to `true`.

Jjagg:
@cra0zy So what do you propose then as an alternative? We want people to set the flag to true.

cra0zy:
> @cra0zy So what do you propose then as an alternative? We want people to set the flag to true.

Why not set it to true by default and write a documentation page about it?

Apostolique:
Having it set to true as default would also be my choice. MonoGame aims to be consistent across platforms. This is a step in the right direction.

cra0zy:
In the meantime you can modify the templates back and write the doc page.

mrhelmut:
Since it doesn't break SpriteBatch and makes everything on par with DirectX, I'm totally fine with breaking the default value.

Jjagg:
Changes made, this is good to go if you ask me.

Test/Framework/Graphics/GraphicsDeviceTestFixtureBase.cs
```diff
@@ -51,6 +51,11 @@ internal class GraphicsDeviceTestFixtureBase
         [SetUp]
         public virtual void SetUp()
         {
+#if !XNA
+            // We enable the half-pixel offset for XNA compatibility
+            GraphicsDeviceManager.UseStandardPixelAddressing = false;
```
nkast: if SpriteBatch rendering is not affected by the flag and produces the same results, then we don't need that. Shouldn't we instead test MG with the default setting?

Jjagg: The tests that failed were not SpriteBatch tests. I don't really know why they did fail though.

 But we do want to set this flag in the tests for when we add tests for the built in effects (if we ever get around to that). Of course DX should support manipulating half pixel offset at runtime for it to work.

Jjagg:
@tomspilman Need your approval for this one because we changed the default to be non-XNA compatible (also added a doc page for XNA compatibility that explains the flag).

Apostolique:
I'd really like this to get merged. It makes MonoGame be more consistent between OpenGL and DirectX.

Jjagg:
For DX we should do the same as GL. That is, inject some code into shaders and set a flag/value in the shader that controls the offset based on the UseStandardPixelAddressing flag. 

Here's an interesting blog post that fixes DX9 shaders by injecting shader bytecode (the reverse of what we want to do, but the approach is the same): https://aras-p.info/blog/2016/04/08/solving-dx9-half-pixel-offset/.

Jjagg:
@cra0zy Could you merge this?

cra0zy:
Of course :)

Jjagg:
Thanks!

nkast:
I don't think that GraphicsDeviceManager is a good place for UseStandardPixelAddressing. It should go under GraphicsDevice. We should avoid a dependency from GraphicsDevice to the GraphicsDeviceManager.

GraphicsDeviceManager is just an implementation of IGraphicsDeviceManager/IGraphicsDeviceService, there are cases where you have a GraphicsDevice without a GraphicsDeviceManager/Game. e.g. in an editor or for GPU data processing.

Jjagg:
It doesn't really matter if you don't have a GraphicsDeviceManager because the property is static. But I'm fine with it in GraphicsDevice as well. It was put in GDM because that's where you set the configuration before creating the GraphicsDevice. If we do move if, we should definitely do it before a release.

nkast:
>> It was put in GDM because that's where you set the configuration before creating the GraphicsDevice. 

I see. A more complete implementation would be with a  'GraphicsDeviceManager.PrefferedUseStandardPixelAddressing', pass the value to the construction Parameters (which is how I would set it in my my Game-less editor), and set a fixed/readonly value for GraphicsDevice.UseStandardPixelAddressing for the entire life of GraphicsDevice. 

With this setup, we can have, 
 * Each instance of  GraphicsDevice has it's own setting.
 * The creator of GraphicsDevice is in control, 3rd party libraries can't change/force their preferred settings.
 * We can pre-calculate most of projection matrix values in SpriteEffect.

In the sort term thou , I can move it to GraphicsDevice and make it non-static.

PR #6623, opened 2019-01-05
Support TransformMatrix in SpriteEffect

Jjagg:
This exposes an optional transformation matrix in `SpriteEffect` so it can actually be used by `SpriteBatch`.
The transformation matrix works exactly like the matrix that can be passed to `SpriteBatch.Begin`. This also fixes `SpriteEffect` depth to work like `SpriteBatch` instead of reversed and improves caching for `SpriteEffect` as was done in `SpriteBatch`. 

This makes `SpriteBatch` a little simpler, makes `SpriteEffect` more useful and puts half pixel offset logic for sprites in one place (which was the driver for me, did this when working on #6621).

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -23,13 +23,9 @@ namespace Microsoft.Xna.Framework.Graphics
 		Effect _effect;
         bool _beginCalled;
 
-		Effect _spriteEffect;
-	    readonly EffectParameter _matrixTransform;
+		SpriteEffect _spriteEffect;
         readonly EffectPass _spritePass;
 
-		Matrix? _matrix;
-	    private Viewport _lastViewport;
-	    private Matrix _projection;
 		Rectangle _tempRect = new Rectangle (0,0,0,0);
 		Vector2 _texCoordTL = new Vector2 (0,0);
 		Vector2 _texCoordBR = new Vector2 (0,0);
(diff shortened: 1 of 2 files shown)


Jjagg:
@tomspilman @cra0zy Can anyone take a look at this.
Need this merged for #6621.

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -130,29 +124,6 @@ void Setup()
 			gd.DepthStencilState = _depthStencilState;
 			gd.RasterizerState = _rasterizerState;
 			gd.SamplerStates[0] = _samplerState;
-			
-			var vp = gd.Viewport;
```
nkast: We should always apply the _matrixTransform on the _spriteEffect, see: https://github.com/MonoGame/MonoGame/pull/5244#issuecomment-256257160

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -51,9 +47,7 @@ public SpriteBatch (GraphicsDevice graphicsDevice)
 
 			this.GraphicsDevice = graphicsDevice;
 
-            // Use a custom SpriteEffect so we can control the transformation matrix
-            _spriteEffect = new Effect(graphicsDevice, EffectResource.SpriteEffect.Bytecode);
```
nkast: This is on intentional. We create an Effect from  SpriteEffect.Bytecode to skip the SpriteEffect.OnApply(),
because we already calculate and apply _matrixTransform  inside Setup().

MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -130,29 +124,6 @@ void Setup()
...
-
-                _lastViewport = vp;
-            }
-
-            if (_matrix.HasValue)
-                _matrixTransform.SetValue(_matrix.GetValueOrDefault() * _projection);
-            else
-                _matrixTransform.SetValue(_projection);
 
             _spritePass.Apply();
```
nkast: does _spritePass.Apply() call SpriteEffect.OnApply()?
Why don't we call  _spriteEffect.Apply() here?

nkast: I though there was an Effect.Apply() but there isn't. 
This PR looks ok.

Jjagg: Thanks!

cra0zy:
As its a feature change I can't merge it, leaving it to @tomspilman

Jjagg:
Ping @tomspilman

Jjagg:
Awesome, thanks Tom!

Issue #6107, opened 2017-12-13
Lack of OpenAL DLLs shouldn't be fatal.


s-m-k:
I don't want to use OpenAL audio at all, so I don't think I should be required to distribute the libraries related to it.

It would be better to make me choose whether I want to initialize the MonoGame audio (and load the libs) or not. It's done in the runtime anyway, so this functionality should be easy to add (e.g. static `Game.InitializeAudio = false;` before creating the Game instance), especially that my use case isn't really isolated. I believe that many gamedevs choose FMOD or Wwise over solutions native to engines/frameworks simply because their sound designers are used to these tools.

I get that it might be a minor issue to most people, but there's no need to be forced to distribute two audio frameworks simultaneously (which is one of many reasons I don't really like Unity) and it seems to be easy to avoid in case of MonoGame. Give us more freedom :)

Jjagg:
I thought MG already silently failed when OpenAL initialization throws an exception. Can you post a stack trace? 

Should we add an API for this as suggested or just silently fail?

s-m-k:
It did in 3.6, but I decided to try 3.7.

Stack trace:
`Microsoft.Xna.Framework.Audio.NoAudioHardwareException occurred
  HResult=0x80004005
  Message=Failed to init OpenALSoundController
  Source=MonoGame.Framework
  StackTrace:
   at Microsoft.Xna.Framework.SdlGamePlatform..ctor(Game game)
   at Microsoft.Xna.Framework.Game..ctor()
   at Monster.EngineSystem..ctor() in E:\LocalProjects\Game210617\Game210617\_Main\EngineSystem.cs:line 71
   at Monster.Program.Main() in E:\LocalProjects\Game210617\Game210617\_Main\Program.cs:line 18`

I was looking at these sources before creating this issue and it seems that SdlGamePlatform does this:

https://github.com/MonoGame/MonoGame/blob/1a9466ef46b98d9b4ea5818190eaa33df1c5a6df/MonoGame.Framework/SDL/SDLGamePlatform.cs#L77

tdeeb:
I'm getting this same error in MonoGame 3.7.0.1129, but I'm using OpenAL. If the user doesn't have sound enabled, I agree that we should be able to handle whether we want the game to run or not. Has there been any progress on this?

andrekoehler:
[...]

Issue #4717, opened 2016-04-03
Startup crash when speakers are not plugged in


tomspilman:
I've learned that XAudio device creation fails if there are no speakers plugged into the PC. That is weird... it is like DirectX failing to initialized because there is no monitor plugged in.

http://stackoverflow.com/questions/35600946/why-oncriticalerror-is-called-in-directx-version-of-xaudio2

Not sure if the OpenAL path suffers from this as well.

MonoGame should be able to internalize this and keep it from being an issue the game developer has to deal with.

lozzajp:
Reading 4715; this issue replicates what XNA does but we want it better?

Is this at runtime when a game starts, in either XNA or MonoGame will crash if it can't find hardware to create an audio device from?

theZMan:
If we changed this behavior then all the existing code that relies on XNA throwing an exception wouldn't work the same ;-)

patridge:
I have this error happening out of the `Microsoft.Xna.Framework.Game` constructor when I call `base()`, so I'm trying to figure out how to even add code that relies on this exception. If it is throwing an exception in the base constructor every time, how would you rely on that exception in a way that isn't horrible for the user? How do you spin up anything useful for the user without an audio device connected? Do you just try-catch around the game constructor and show an alert before ultimately exiting to desktop?

Halofreak1990:
Odd place for an exception like this, and not how XNA does things;. XNA will throw a NoAudioHardwareException only when you actually try to use the audio subsystem, either by initializing the XACT classes or by attempting to play audio using the SoundEffect or Song classes. As long as you don't do that, you _should_ be safe.

I think that the MonoGame code should be changed to lazily initialize the audio subsystem when first used.

patridge:
Even if the audio system was lazily initialized, it doesn't seem like the game should necessarily have to care about there not being audio hardware. Even if a game tries to play audio on a system with no way to do so, it should just carry on like it worked (potentially offering a way to detect the no-hardware situation for user-facing troubleshooting messaging). It feels like there shouldn't really be a NoAudioHardwareException as much as a HasAudioHardware property somewhere.

Halofreak1990:
@patridge 
> Even if a game tries to play audio on a system with no way to do so, it should just carry on like it worked (potentially offering a way to detect the no-hardware situation for user-facing troubleshooting messaging).

That's what the exception is for ;)
You catch it, set a flag to not use audio from that point forward, and you're good.

[...]

PR #6629, opened 2019-01-06
Optional sound system

### LINKED ISSUE #6515: [OpenAL] Wrong exception type when binaries are missing

mrhelmut:
Hey there,

This PR should make ```Microsoft.Xna.Framework.Media``` and ```Microsoft.Xna.Framework.Audio``` optional so that no sound system is initialized if none of those namespace are used. This should make MonoGame friendly to third party sound engine like FMOD or Wwise.

It addresses:
- #6515 exception types are now consistent (it's either ```DllNotFoundException``` or ```NoAudioHardwareException```)
- #6107 if no ```SoundEffect``` ```Song``` ```MediaPlayer``` ```Microphone``` or ```DynamicSoundEffectInstance``` are used, OpenAL won't get loaded and the dll can be removed from release packages
- #4717 this error can now be caught through ```SoundEffect.Initialize()``` and developers can make it so it is not fatal

As suggested previously, sound engine initialization is now done whenever ```SoundEffect``` ```Song``` ```MediaPlayer``` ```Microphone``` or ```DynamicSoundEffectInstance``` are constructed.
It is however possible to manually initialize the ```SoundEffect``` API by calling ```SoundEffect.Initialize()```. This allows developers to get any initialization errors and act accordingly (e.g. disabling sound, or make it fatal).

Fixes #6515 #6107 #4717

@Jjagg @cra0zy 
@tomspilman if this goes through, I'll try to get the console repos' compliant.

### DIFF
--- a/MonoGame.Framework/Audio/Microphone.cs
+++ b/MonoGame.Framework/Audio/Microphone.cs
@@ -115,6 +115,7 @@ namespace Microsoft.Xna.Framework.Audio
         {
             get
             {
+                SoundEffect.Initialize();                
                 if (_allMicrophones == null)
                     _allMicrophones = new List<Microphone>();
                 return new ReadOnlyCollection<Microphone>(_allMicrophones);
--- a/MonoGame.Framework/Audio/Microphone.cs
+++ b/MonoGame.Framework/Audio/Microphone.cs
@@ -129,7 +129,7 @@ namespace Microsoft.Xna.Framework.Audio
         /// </summary>
         public static Microphone Default
         {
-            get { return _default; }
+            get { SoundEffect.Initialize(); return _default; }
         }       
 
         #endregion
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -286,29 +286,36 @@ namespace Microsoft.Xna.Framework.Audio
             return false;
         }
 
-		public static OpenALSoundController GetInstance
+        public static void EnsureInitialized()
+        {
+            if (_instance == null)
+            {
+                try
+                {
+                    _instance = new OpenALSoundController();
+                }
+                catch (DllNotFoundException)
+                {
+                    throw;
+                }
+                catch (NoAudioHardwareException)
+                {
+                    throw;
+                }
+                catch (Exception ex)
+                {
+                    throw (new NoAudioHardwareException("Failed to init OpenALSoundController", ex));
+                }
+            }
+        }
+
+
+        public static OpenALSoundController Instance
         {
 			get
             {
                 if (_instance == null)
-                {
-                    try
-                    {
-                        _instance = new OpenALSoundController();
-                    }
-                    catch (DllNotFoundException)
-                    {
-                        throw;
-                    }
-                    catch (NoAudioHardwareException)
-                    {
-                        throw;
-                    }
-                    catch (Exception ex)
-                    {
-                        throw (new NoAudioHardwareException("Failed to init OpenALSoundController", ex));
-                    }
-                }
+                    throw new NoAudioHardwareException("OpenAL context has failed to initialize. Call SoundEffect.Initialize() before sound operation to get more specific errors.");
 				return _instance;
 			}
 		}
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -291,8 +291,17 @@ namespace Microsoft.Xna.Framework.Audio
         {
 			get
             {
-				if (_instance == null)
-					_instance = new OpenALSoundController();
+                if (_instance == null)
+                {
+                    try
+                    {
+                        _instance = new OpenALSoundController();
+                    }
+                    catch (Exception ex)
+                    {
+                        throw (new NoAudioHardwareException("Failed to init OpenALSoundController", ex));
+                    }
+                }
 				return _instance;
 			}
 		}
--- a/MonoGame.Framework/Audio/OpenALSoundController.cs
+++ b/MonoGame.Framework/Audio/OpenALSoundController.cs
@@ -108,6 +108,9 @@ namespace Microsoft.Xna.Framework.Audio
         /// </summary>
 		private OpenALSoundController()
         {
+            if (AL.NativeLibrary == IntPtr.Zero)
+                throw new DllNotFoundException("Couldn't initialize OpenAL because the native binaries couldn't be found.");
+
             if (!OpenSoundController())
             {
                 throw new NoAudioHardwareException("OpenAL device could not be initialized, see console output for details.");
@@ -148,10 +151,6 @@ namespace Microsoft.Xna.Framework.Audio
             {
                 _device = Alc.OpenDevice(string.Empty);
                 EffectsExtension.device = _device;
-            }
-            catch (DllNotFoundException ex)
-            {
-                throw ex;
             }
             catch (Exception ex)
             {
@@ -296,6 +295,14 @@ namespace Microsoft.Xna.Framework.Audio
                     try
                     {
                         _instance = new OpenALSoundController();
+                    }
+                    catch (DllNotFoundException)
+                    {
+                        throw;
+                    }
+                    catch (NoAudioHardwareException)
+                    {
+                        throw;
                     }
                     catch (Exception ex)
                     {
--- a/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
@@ -235,17 +235,9 @@ namespace Microsoft.Xna.Framework.Audio
 
 #endregion
 
-        internal static void InitializeSoundEffect()
+        internal static void PlatformInitialize()
         {
-            try
-            {
-                // Getting the instance for the first time initializes OpenAL
-                var oal = OpenALSoundController.GetInstance;
-            }
-            catch (DllNotFoundException ex)
-            {
-                throw new NoAudioHardwareException("Failed to init OpenALSoundController", ex);
-            }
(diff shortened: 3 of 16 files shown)


MonoGame.Framework/Audio/OpenALSoundController.cs
```diff
@@ -291,8 +290,25 @@ public static OpenALSoundController GetInstance
         {
 			get
             {
-				if (_instance == null)
-					_instance = new OpenALSoundController();
+                if (_instance == null)
+                {
+                    try
```
Jjagg: This `GetInstance` property doesn't really make sense to me, at least with this naming. Should be a method IMO.
Also would make more sense to be able to call an initialize function instead of this to initialize OpenAL. Can you extract the code here into an `EnsureInitialized` function:

```
public static void EnsureInitialized()
{
    if (_instance ~= null)
        return;
    try
    {
    ...
}

public static OpenALSoundController GetInstance()
{
    EnsureInitialized();
   return _instance;
}
```

MonoGame.Framework/Audio/SoundEffect.OpenAL.cs
```diff
@@ -235,17 +235,9 @@ private void PlatformDispose(bool disposing)
...
-            try
-            {
-                // Getting the instance for the first time initializes OpenAL
-                var oal = OpenALSoundController.GetInstance;
-            }
-            catch (DllNotFoundException ex)
-            {
-                throw new NoAudioHardwareException("Failed to init OpenALSoundController", ex);
-            }
+            var inst = OpenALSoundController.GetInstance;
```
Jjagg: This should call OpenALSoundController.EnsureInitialized().

MonoGame.Framework/Audio/SoundEffect.cs
```diff
@@ -63,7 +71,11 @@ internal SoundEffect(byte[] header, byte[] buffer, int bufferSize, int durationM
 
         // Only used from XACT WaveBank.
         internal SoundEffect(MiniFormatTag codec, byte[] buffer, int channels, int sampleRate, int blockAlignment, int loopStart, int loopLength)
-        {
+        {
+            Initialize();
+            if (_systemState != SoundSystemState.Initialized)
```
Jjagg: These should maybe throw?

MonoGame.Framework/Media/Song.NVorbis.cs
```diff
@@ -16,6 +16,9 @@ public sealed partial class Song : IEquatable<Song>, IDisposable
 
         private void PlatformInitialize(string fileName)
         {
+            // init OpenAL if need be
+            var inst = OpenALSoundController.GetInstance;
```
Jjagg: Should call OpenALSoundController.EnsureInitialized() too.

MonoGame.Framework/Audio/OpenALSoundController.cs
mrhelmut: Yeah, this GetInstance has been there for a long time and has been used to initialize stuff at several places...

I think we should turn it into a property and make it throw if it is called before the initialization, so that the code the more readable with no hidden initialization due to a property accessor.

mrhelmut:
Thanks! I'm going to refactor the controller instance management, it has been a bit messy for quite too long.

MonoGame.Framework/Audio/OpenALSoundController.cs
Jjagg: > I think we should turn it into a property and make it throw if it is called before the initialization, so that the code is more readable with no hidden initialization due to a property accessor.

 Even better

MonoGame.Framework/Audio/SoundEffect.cs
mrhelmut: I believe there are not even useful since Initialize() throws.

mrhelmut:
@Jjagg I think I got things clean up. Let me know how it goes for you.

mrhelmut:
Ready for review.

Jjagg:
Looks good to me! Gonna hold off a little bit in case anyone else wants to go over this, but will merge if not.

MonoGame.Framework/Audio/Microphone.cs
```diff
@@ -128,7 +129,7 @@ public static ReadOnlyCollection<Microphone> All
         /// </summary>
         public static Microphone Default
         {
-            get { return _default; }
```
cra0zy: Write code like this in the static initializer instead of static property.

MonoGame.Framework/Audio/Microphone.cs
```diff
@@ -128,7 +129,7 @@ public static ReadOnlyCollection<Microphone> All
...
+            get { SoundEffect.Initialize(); return _default; }
```
cra0zy: Write code like this in the static initializer instead of static property.

cra0zy: Just so there is no confusion, ie:

```c#
static Microphone()
{
    SoundEffect.Initialize();
}
```

cra0zy: You could also use static constructors in few other places.

Jjagg: I think it's better to check if initialization was done properly on every call, rather than do it only once. It makes behavior more predictable. Static initializers complicate control flow because you can't be sure when they'll be called.

cra0zy: They get called during first access to their respective class.

Jjagg: Yes, I mean we don't control when it's called.

mrhelmut: The thing is, ```Default``` and ```All`` are the microphone entry points, but you have to initialize the sound API to populate the microphones, and the Microphone class is static by design...
Unless we change the design of the Microphone class, I don't think we can do anything else than having this in a static field. I'm not a fan of static classes because this kind of design problems.

Alternative proposition: not initializing the sound API at all in the Microphone class, and throw if All/Default are accessed while not initialized. Forcing users to either call ```SoundEffect.Initialize()``` or to use the auto initialization when loading content.

I've seen static classes initializing without obvious calls to them, so I'd feel more comfortable with not having any library loading happening inside static ctor/fields.

For the same reason, I don't quite like ```OpenAL.NativeLibrary``` being initialized statically. I think I'm going to change that too.

Jjagg: We should not change the API by forcing explicit initialization. If we could have a fresh API, I'd definitely want to make it explicit though.

cra0zy: How about just checking if Audio systems can be initialized on game run?

> For the same reason, I don't quite like OpenAL.NativeLibrary being initialized statically. I think I'm going to change that too.

Please don't, those are just wrappers. Think of them as `DllImport` statements :P

mrhelmut: Sure, but they do load the dll if the static ctor kicks in, which we want to avoid to not mess up with third party sound engine.

But if we can't get the sound api to initialize in a non static way for Microphone, there's no point in removing the static initialization from OpenAL anyway because it'll still be around within Microphone.

mrhelmut:
I think we have to take a stance regarding static initialization.

Option 1) Leave it be, and 100% trust the runtime to not initialize static classes unless they are used. Move Microphone initialization to a static ctor.

Option 2) Don't trust the runtime. Move all OpenAL function loading outside of the static initialization.

I'm in favor of 2), so that we have a control over when and how OpenAL gets loaded.

MonoGame.Framework/Audio/Microphone.cs
mrhelmut: I am more suspicious of Microphone than I am of OpenAL. I'd prefer to not add a static constructor to Microphone.

If we'd like to be 100% safe, I'd suggest to move static initialization out of OpenAL func loading and manually load them (to have a perfect control on the "when" and "how").

mrhelmut:
After giving some thoughts about static initializers, I believe that we are 100% fine with ```OpenAL``` statically loading the binary. I assume that people not willing to use OpenAL will remove the dll from their distribution, which doesn't trigger any error if OpenAL tries to load anyway (unless ```SoundEffect.Initialize()``` is called, but then it's the expected behavior).

I am however not in favor of having ```Microphone``` use a static initializer to initialize the sound API. As of now everything would be fine, but we have no control over what could reference Microphone and possibly trigger its initializer. For safety, I prefer to have the initialization upon accessing the microphones list.

@Jjagg @cra0zy What do you think? If it's ok to you, then it should be good to go.

Jjagg:
That's fine by me.

mrhelmut:
@cra0zy Could you share your thoughts about the current status of this PR? It would be cool to move forward.

I've been running on this PR for quite time now, using FMOD.

cra0zy:
Looks fine to me, sorry for the delay.

Jjagg:
Thanks @mrhelmut! Merging this :D

PR #6647, opened 2019-01-22
Fixing RenderTarget2D.GetData() when MSAA is used

palver123:
We use MonoGame and cannot update to the latest version because of a bug in RenderTarget2D.GetData().

Texture2D.**PlatformGetData**() used a staging texture in DirectX to obtain pixel data, which was constructed based on the underlying SharpDX texture. In DirectX 11 (and probably older) it is forbidden to use a staging texture with multisampling according to the [documentation](https://docs.microsoft.com/en-us/windows/desktop/api/d3d11/ne-d3d11-d3d11_usage):

>  ...Therefore, staging resources can't be a depth-stencil buffer or a multisampled render target.

Older versions of SharpDX did not enforce this (in MonoGame 3.6 you could get away with it). At the time of writing this PR, SharpDX 4.0.1 is used which throws an exception.

I have modified the code to create the staging texture **without MSAA**. However, this was not enough. If we want to copy data from the multisampled render target to the non-multisampled staging texture, we [have to use](https://docs.microsoft.com/en-us/windows/desktop/api/d3d11/nf-d3d11-id3d11devicecontext-copyresource) _ID3D11DeviceContext::ResolveSubresource()_ instead of _CopySubresourceRegion()_.

I included unit tests which failed before my modifications and passed after it.


palver123:
So what's up with my fix? Is it going to be merged? It seems to build alright on every platform. Only affects the DirectX platform and improves stability on that particular platform. Any reasons not to merge it?

We cannot upgrade our MonoGame NuGet in our projects until then.

harry-cpp:
A PR that does the same thing has been merged: https://github.com/MonoGame/MonoGame/pull/5838

Closing this one.

PR #6650, opened 2019-01-24
Move platform-specific files for easy new csproj setup

### LINKED ISSUE #6207: Remove Protobuild
There are so many better ways of organizing the repo without using Protobuild:
- using shared projects
- using the new csproj format + few wildcards
- using the old csproj format + few wildcards

For the first pass we would just need to generate all the .csproj files, this way we won't be changing anything for the build system, and afterwards we could slowly do modifications over time with smaller changes.

Alternatively we could make it only generate the MonoGame.Framework and MonoGame.Framework.Net projects, while everything else is made of static projects.

Jjagg:
~This is a WIP branch that removes Protobuild and replaces it with SDK-style .csproj files (where possible).
I'm setting up this PR for feedback and to share progress. If anyone wants to help out by porting platforms, please let me know.~

~It will be pretty painful to get this merged because it moves around a lot of files, so it will cause merge conflicts.~

- [x] ~WindowsDX~
- [x] ~DesktopGL~
- [ ] ~DesktopGL Mac~
- [x] ~UWP~
- [x] ~Android~
- [x] ~iOS~
- [ ] ~Web~

EDIT: We're changing the strategy here. This pull request will only move the files so the platform-specific files are separated from shared files. We'll set up seperate PR's for new csproj files for each platform to make the transition more manageable.


cra0zy:
I just found out that you can actually click on the checkbox items instead of having to edit them... impressive github, I like it :)

cra0zy:
> It will be pretty painful to get this merged because it moves around a lot of files, so it will cause merge conflicts.

It won't actually, moving files won't make any git conflicts.

Jjagg:
Seems like Xamarin still doesn't officially support SDK style projects for Xamarin.Mac, Xamarin.iOS and Xamarin.Android. I used [MSBuild.Sdk.Extras](https://github.com/onovotny/MSBuildSdkExtras), but I don't have a Mac, so I didn't test it. It probably needs some tinkering to build properly. I'll use the same project for mobile.

Jjagg:
I added project files for everything I listed except web. I've only tested WindowsDX and DesktopGL (I'm on Linux and don't have Android working), so I'll need some help to test and fix up the other project types (if necessary).

Jjagg:
I managed to build Android, but have not tested it.

cra0zy:
> I'm on Linux

My man :)

dsaf:
@Jjagg how would you test Android / UWP? Is it a matter of running some unit/integration tests?

Jjagg:
@dsaf no, we just have to check that all classes are included, so not necessarily by running tests. If it builds it's already very likely things are good :) 
One way to check is to compare the included files with the Protobuild project.

dsaf:
> @dsaf no, we just have to check that all classes are included, so not necessarily by running tests. If it builds it's already very likely things are good :)
> One way to check is to compare the included files with the Protobuild project.

@Jjagg 
I am not familiar enough with Protobuild to do any comparison so I tried just building.
I might not be super useful here to be honest. My experience is with creating projects via Visual Studio (which gives me a nice platform-specific project icon etc.).

So saying that, this is my experience building projects so far:

1. Is there a reason not to have an ```.sln``` file?

2. Is not including some of the ```..\ThirdParty\Dependencies...``` stuff by design / due to limitations (as opposed to NuGet or keeping in same repo)?

3. Android project. I also had to install Android 9.0 - Pie SDK. When it comes to Android does it make sense to always target the latest available OS? I am asking because I don't know.

image

4. DesktopDX project.

image

5. DesktopGL project. Failed because it probably needs additional sources downloaded? Why not a DLL/NuGet?

```1>CSC : error CS2001: Source file '... ..\ThirdParty\NVorbis\NVorbis\BufferedReadStream.cs' could not be found.```

6. iOS project. Not familiar with these. Failed because of this:

image

image

...tried this one randomly and it started compiling further it seems:

image

...now gives a lot of similar type missing errors (might be my setup - never tried Xamarin iOS):

image

7. UWP project. I think some paths are not fully right:

This one seems to be in ```Plaftorm```:
image

The name is seems different here:

image

image

Also added this one - not sure if was a right thing to do:

```
    <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
      <Version>6.2.3</Version>
    </PackageReference>
```

...otherwise it was giving this: 
image

8. Aren't JSON projects deprecated now?

image

I kind of rushed through things, hopefully it will be at least a bit useful and not misleading.

cra0zy:
@dsaf most of the questions you asked are really redundant, just wait for the PR to be ready for testing...

> 1. Is there a reason not to have an `.sln` file?

This is WIP, first part is replacing the project files, solution files are easily generated but there is no need until project files are replaced.

> 2. Is not including some of the `..\ThirdParty\Dependencies...` stuff by design / due to limitations (as opposed to NuGet or keeping in same repo)?

We want it that way so that those deps can be modified when needed, and also included in the main .dll file.

> 3. Android project. I also had to install Android 9.0 - Pie SDK. When it comes to Android does it make sense to always target the latest available OS? I am asking because I don't know.

There is a difference between target SDK and min SDK.

> 4. DesktopDX project.

WIP!

> 5. DesktopGL project. Failed because it probably needs additional sources downloaded? Why not a DLL/NuGet?

You did not clone the submodules but only the repository itself.

> 6. iOS project. Not familiar with these. Failed because of this:
> 7. UWP project. I think some paths are not fully right:

iOS and UWP projects are not even marked as ready.

dsaf:
@cra0zy

> @dsaf most of the questions you asked are really redundant, just wait for the PR to be ready for testing...

I only spent time doing this because @Jjagg asked for testers in Gitter and nobody responded...

cra0zy:
I stand corrected then

Jjagg:
@cra0zy I asked for help because I can't test iOS and Mac and don't have the UWP workload installed. 

Thanks for helping out @dsaf!

So yeah, for DesktopGL you need to clone the submodules.
For iOS, the MSBuild.Sdk.Extras repo had xamarin.ios in the readme for the target, but I also had to change the target for Android. I remember getting that same DataContract error as well, but don't remember how I fixed it. Will check it out later, I'm currently on holiday.

Regarding UWP, good catch on the AssemblyInfo and the theme. The json projects are old, I don't think they're used, they just were not removed yet.

Jjagg:
@dsaf for DataContract we need to add a reference to System.Runtime.Serialization, check my last commit :)

dsaf:
@cra0zy @Jjagg thank you for replies, understood! I will give it another go once the PR progresses further!

Jjagg:
UWP is building :) 

@dsaf I made some changes that might fix the iOS build. If you happen to have some time, could you try iOS again? If it doesn't work please send me a message on Gitter so we can go back and forth on it faster.

dsaf:
> @dsaf I made some changes that might fix the iOS build. If you happen to have some time, could you try iOS again? If it doesn't work please send me a message on Gitter so we can go back and forth on it faster.

It works!

image

Jjagg:
@cra0zy what should we do with web here? Also, could you please take a look at the Mac build (assuming you still run OSX)?

cra0zy:
> what should we do with web here?

Burn it down! No, seriously, that project file is useless to my Bridge.NET version. Once you are done with this PR I'll look into merging the web version back.

> Also, could you please take a look at the Mac build (assuming you still run OSX)?

I don't own a single Apple device, tho I do have access to a mac machine at my work right now.

Jjagg:
Okay, so let's skip Web for now then.
So to get all non-console platforms at least building that only leaves the Mac DesktopGL variant :)

Jure-BB:
Any new development? Can't believe no one has a Mac to test this 😮

cra0zy:
@Jjagg Look at all the fun merge conflicts you have :P

cra0zy:
@Jjagg say when its ready.

Jjagg:
Will do. It might be already, but I'm gonna run a game to make sure.

Jjagg:
@cra0zy Good to go! :D

cra0zy:
Finally, we are moving :)

cra0zy:
in retrospect I probably should have pressed rebase and merge... oh well.

nkast:
The following files should also be moved to /Platform as openGL targets.
IMO they should be renamed as IGLWindowInfo and IGLGraphicsContext.

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/IWindowInfo.cs
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/IGraphicsContext.cs

nkast:
I see [here](https://koukia.ca/rename-or-move-files-in-git-e7259bf5a0b7) that the correct way to move/rename files with 'git mv' to avoid conflicts. Is that right?

cra0zy:
It can detect it automatically, only exception being you go over the user/system specified git rename limit.

cra0zy:
(aka. rename too many files)

Issue #6646, opened 2019-01-22
[Android] Xbox One Wireless controller triggers are not working properly


Eversor:
Xbox One Bluetooth Wireless controller has triggers on different axis than the default android controllers.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows 10

#### What MonoGame platform are you using:
- Android

PR #6651, opened 2019-01-25
[Android] Fixed support to XBox One Wireless gamepad.

Eversor:
Solves issue : https://github.com/MonoGame/MonoGame/issues/6646
Fixed support to XBox One Wireless gamepad. Triggers are now connected to the Gas and Brake axes as required.

(cherry picked from commit 77ca5b3156514905e7f75a9a1c04b0f1d5712588)

### DIFF
--- a/MonoGame.Framework/Input/GamePad.Android.cs
+++ b/MonoGame.Framework/Input/GamePad.Android.cs
@@ -258,8 +258,17 @@ namespace Microsoft.Xna.Framework.Input
 
             gamePad._leftStick = new Vector2(e.GetAxisValue(Axis.X), -e.GetAxisValue(Axis.Y));
             gamePad._rightStick = new Vector2(e.GetAxisValue(Axis.Z), -e.GetAxisValue(Axis.Rz));
-            gamePad._leftTrigger = e.GetAxisValue(Axis.Ltrigger);
-            gamePad._rightTrigger = e.GetAxisValue(Axis.Rtrigger);
+
+            if (e.Device.Name == "Xbox Wireless Controller")
+            {
+                gamePad._leftTrigger = e.GetAxisValue(Axis.Brake);
+                gamePad._rightTrigger = e.GetAxisValue(Axis.Gas);
+            }
+            else
+            {
+                gamePad._leftTrigger = e.GetAxisValue(Axis.Ltrigger);
+                gamePad._rightTrigger = e.GetAxisValue(Axis.Rtrigger);
+            }
 
             if(!gamePad.DPadButtons)
             {
--- a/MonoGame.Framework/Input/GamePad.Android.cs
+++ b/MonoGame.Framework/Input/GamePad.Android.cs
@@ -2,6 +2,7 @@
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
+using System;
 using Android.Views;
 
 namespace Microsoft.Xna.Framework.Input
@@ -20,6 +21,16 @@ namespace Microsoft.Xna.Framework.Input
 
         public readonly GamePadCapabilities _capabilities;
 
+        public bool _hasCustomMapping = false;
+
+        public Axis _leftStickXCustomAxis = Axis.X;
+        public Axis _leftStickYCustomAxis = Axis.Y;
+        public Axis _rightStickXCustomAxis = Axis.Z;
+        public Axis _rightStickYCustomAxis = Axis.Rz;
+
+        public Axis _leftTriggerCustomAxis = Axis.Ltrigger;
+        public Axis _rightTriggerCustomAxis = Axis.Rtrigger;
+
         public AndroidGamePad(InputDevice device)
         {
             _device = device;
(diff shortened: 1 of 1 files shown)


filnet:
Seems to address only Android (if true maybe it should say so in the commit log).
What about other platforms ?

MonoGame.Framework/Input/GamePad.Android.cs
```diff
@@ -258,8 +258,17 @@ internal static bool OnGenericMotionEvent(MotionEvent e)
 
             gamePad._leftStick = new Vector2(e.GetAxisValue(Axis.X), -e.GetAxisValue(Axis.Y));
             gamePad._rightStick = new Vector2(e.GetAxisValue(Axis.Z), -e.GetAxisValue(Axis.Rz));
-            gamePad._leftTrigger = e.GetAxisValue(Axis.Ltrigger);
-            gamePad._rightTrigger = e.GetAxisValue(Axis.Rtrigger);
+
+            if (e.Device.Name == "Xbox Wireless Controller")
```
filnet: this test could be done once (somewhere...) and the Axis to use would be stored in a class variable to be used here.

Eversor: Absolutely correct, could be optimized via a configuration that allows more known gamepads to be recognized. Still this works better than the current implementation, that doesn't work with one of the most wide-spread controllers to date.

MonoGame.Framework/Input/GamePad.Android.cs
```diff
@@ -20,13 +21,32 @@ internal class AndroidGamePad
...
+        Axis _dPadYCustomAxis = Axis.HatY;
+
         public AndroidGamePad(InputDevice device)
         {
             _device = device;
             _deviceId = device.Id;
             _descriptor = device.Descriptor;
             _isConnected = true;
 
+            if (device.Name == "Xbox Wireless Controller")
```
filnet: The name is always in English ?

MonoGame.Framework/Input/GamePad.Android.cs
```diff
@@ -256,19 +320,26 @@ internal static bool OnGenericMotionEvent(MotionEvent e)
             if (e.Action != MotionEventActions.Move)
                 return false;
 
+            if (gamePad._hasCustomMapping)
+            {
+                gamePad.ProcessMotionEvent(e);
```
filnet: why not just update the code below to use the Axis variables you introduced ?

Eversor: because the code below is faster as it deals with constants. Theorethically the whole performance-wise cost of the fix, for gamepads that didn't need the fix, is a string comparison at creation of the pad and a boolean check per connected pad every frame.

mrhelmut:
Other platforms don't use that code and should be working fine with official drivers.

SimonDarksideJ:
Can you resolve the merge conflicts on this PR @Eversor so we can retest with the latest MonoGame?

Eversor:
Hello @SimonDarksideJ i'll try, but my git is a little rusty.

SimonDarksideJ:
If you need any assistance, then just shout.  If need be, you can give me contributor access to your fork and I should be able to resolve it for you.

Eversor:
I gave you access just in case, but i'll see if i can lighten the load and fix this myself... In short what i have to do is update to the latest Monogame commit, fix the conflicts and make a new pull request?

SimonDarksideJ:
- Pull from the latest develop
- Resolve any conflicts
- Push back to your branch

No need for another PR, as this one is already linked to your fork/branch

mrhelmut:
Is this still needed at all? As far as I can tell, we have Xbox One controllers working fine on Android over here without this PR.

SimonDarksideJ:
That is a fair point @mrhelmut but I've not had a chance to test.  @Eversor , did you want to check with the latest MonoGame libraries to see if this is still an issue.

You can probably test with the Platformer2D sample on the MonoGame.Samples repo which supports gamepad input.
https://github.com/MonoGame/MonoGame.Samples/tree/3.8.1/Platformer2D

Eversor:
Didn't have time yesterday but i'll try tonight. the pull request just had a few conflicts so i can try to build with and without the fix. @mrhelmut afaik the platforming sample doesn't use the triggers but i should be able to easily create a demo to test all controller buttons and axes.

Eversor:
I managed to recreate the same issue with the latest version. The IsButtonDown for Buttons.LeftTrigger and Buttons.RightTrigger is not being called correctly. I was trying to test my pr in this situation exactly but cannot link my Monogame projects because i get this error "The MonoGamePlatform property was not defined in the project".
I used to link the Monogame csproj to the main game in 3.5 and before like this...

image

SimonDarksideJ:
Sounds like your csproj using an old template, or the MG packages you are using are older than 3.8.0.

Try creating a new MGAndroid DotNet project (making sure to update the target framework to `net8.0-android`) and retesting.

Also make sure to uninstall the MGCB global tool to ensure the local version is used:
https://docs.monogame.net/articles/migrate_38.html

SimonDarksideJ:
Any update @Eversor ?  Make sure to also test with the latest 3.8.2 release

SimonDarksideJ:
OK, we spun up a test project using the latest 'Develop' branch and tested all the GamePad inputs.
The issue advertised does not appear to be present in the current state of development, however extended testing has identified other issues which need to be resolved.
https://github.com/SimonDarksideJ/MGGamePadTest

Will review any changes needed against this PR, but closing this PR as the issue being fixed is not present

Issue #6279, opened 2018-04-13
A question about widely misused finalizers and dispose


quixoticaxis:
I've been delving in the code and I've found a lot of classes that implement Dispose pattern not correctly.

1. Some of them try to emulate destructor semantics with a finalizer (which is considered an anti-pattern) to interact with global singletons. Although it is ugly and unreadable and a few classes do it in an unsafe way, (for example, `Game` with `ANDROID` defined, `GamePlatform`, etc.), I can understand their existence.
2. Some classes implement finalizes that do nothing useful at all. They are either empty or delegate their functions to some helper methods that do nothing meaningful for finalization (`VideoContent`, `SoundEffect`, `SoundEffectInstance`, `AudioEngine`, `SoundBank`, `WaveBank`, `ContentManager`, etc.) 
3. `GameComponent` defines completely meaningless finalizer that can only call the logic defined in his children, which is not needed because the children should provide their own finalizers if they work with unmanaged resources
4. Some classes seem to rely on the order of finalization (they try to dispose other entities through references and it is only meaningful if the entities have not been finalized) which is undetermined.

To cut the long story short, I've found few classes with finalizers that actually work with `GCHanldes` or unmanaged pointers.

Can someone explain me the reason why? Is it a natural artefact of the long going development (as I see in the history, some of the finalizers were removed/deprecated once, but came back later) or is their any greater meaning to them?

quixoticaxis:
@Jjagg could you please provide a little guidance to me? As i started refactoring finalizers, I'm obviously in need of running tests. Is `MonoGameTests.exe` a confessional way of running tests? Also what result should I expect?
I've got
```
Test Run Summary
  Overall result: Warning
  Test Count: 1292, Passed: 1272, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 20
    Skipped Tests - Ignored: 20, Explicit: 0, Other: 0
```
Are there any other tests I should run?

Jjagg:
@quixoticaxis Yes, those are the tests you should run. If you set up a pull requests the build bot will run the tests as well, on Windows and on Mac, so you don't *have* to run them locally.

quixoticaxis:
Created a PR #6656

PR #6656, opened 2019-02-03
Removes finalizers from multiple classes

quixoticaxis:
Discussed [here](https://github.com/MonoGame/MonoGame/issues/6279).

I removed finalizers which can be proven (or guessed with a high probability) to do little to nothing useful. The ones I left intact are the ones that are linked to some magic logic and global state handling. I hope this PR would make the code of this awesome framework more manageable.

At the time of writing the tests for this PR are passing at least as good as original "develop" branch on my local machine (1 error in `Debug` mode, 1 error + 1 failure in `Release` mode).

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/VideoContent.cs
+++ b/MonoGame.Framework.Content.Pipeline/VideoContent.cs
@@ -110,7 +110,6 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         public void Dispose()
         {
             Dispose(true);
-            GC.SuppressFinalize(this);
         }
 
         protected virtual void Dispose(bool disposing)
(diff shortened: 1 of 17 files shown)


nkast:
The finalizes are needed to implement the dispose pattern.
https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

quixoticaxis:
@nkast From the link you provided: "The dispose pattern is used only for objects that access unmanaged resources". And *all* of these classes do *not* access unmanaged resources.

nkast:
Then you should remove the IDispose first before removing the finalizers.

quixoticaxis:
@nkast you probably may want to read the article to the end or use Visual Studio to auto-implement "Dispose pattern" - you'll see that it would be implemented without the finalizer (and there even would be a comment that explains why)

quixoticaxis:
After re-evaluating the class hierarchies I decided to close this PR in order to re-open it (or another one) later. Removing finalization suppression everywhere was a mistake on my part as some of the changed classes are probably meant to serve as base classes for custom user hierarchies.

Issue #6164, opened 2018-01-25
Unloading individual content


tdeeb:
MonoGame currently does not support unloading content individually, which is a feature I feel can be useful for a variety of reasons. I'll list several use-cases:

1. Playing a cutscene that has one new asset that needs to be loaded just for that cutscene.
2. Unloading a single shader that no longer needs to be used.
3. Wanting to unload some assets without unloading every asset.

These can be done with multiple ContentManagers, but that seems overkill if you have one or two assets that you'd need.

I suggest we can either add an overload to Unload with the following signature:

`public virtual void Unload(string assetName)`

or add a new method:

`public virtual void UnloadAsset(string assetName)`

Thoughts?

tdeeb:
I'm going to look more into this. @Jjagg I vaguely recall a while back you said that models may be problematic because this method would have to also unload anything related to them, such as textures. What if we make it clear that it's unloading only a single asset and not anything else?

Jjagg:
I'm fine with adding the method. It's something that comes up every once in a while. 

This has been a discussion point before though, so I'd like other people to weigh in.

@tomspilman @cra0zy

mrhelmut:
Definitely something that would prove useful. I keep hearing from devs who re-implement their own content manager because of this.

cra0zy:
Agreed.

PR #6663, opened 2019-02-16
Added selective ContentManager asset unloading

tdeeb:
This PR adds two new methods to ContentManager: 

1. `UnloadAsset` - Unloads a single asset and disposes it if it's disposable
2. `UnloadAssets` - Takes in an `IList<string>` of asset names to unload and disposes the associated assets if they're disposable

Relevant issue: https://github.com/MonoGame/MonoGame/issues/6164

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -471,6 +471,58 @@ namespace Microsoft.Xna.Framework.Content
 		    loadedAssets.Clear();
 		}
 
+        /// <summary>
+        /// Unloads a single asset.
+        /// </summary>
+        /// <param name="assetName">The name of the asset to unload. This cannot be null.</param>
+        public virtual void UnloadAsset(string assetName)
+        {
+            if (string.IsNullOrEmpty(assetName))
+            {
+                throw new ArgumentNullException("assetName");
+            }
+            if (disposed)
+            {
+                throw new ObjectDisposedException("ContentManager");
+            }
+
+            //Check if the asset exists
+            object asset;
+            if (loadedAssets.TryGetValue(assetName, out asset))
+            {
+                //Check if it's disposable and remove it from the disposable list if so
+                var disposable = asset as IDisposable;
+                if (disposable != null)
+                {
+                    disposable.Dispose();
+                    disposableAssets.Remove(disposable);
+                }
+
+                loadedAssets.Remove(assetName);
+            }
+        }
+
+        /// <summary>
+        /// Unloads a set of assets.
+        /// </summary>
+        /// <param name="assetNames">The names of the assets to unload.</param>
+        public virtual void UnloadAssets(IList<string> assetNames)
+        {
+            if (assetNames == null)
+            {
+                throw new ArgumentNullException("assetNames");
+            }
+            if (disposed)
+            {
+                throw new ObjectDisposedException("ContentManager");
+            }
+
+            for (int i = 0; i < assetNames.Count; i++)
+            {
+                UnloadAsset(assetNames[i]);
+            }
+        }
+
 		public string RootDirectory
 		{
 			get


nkast:
@tdeeb 

-What would happen if I unload a Model ?
(will it free up the resources? VertexBuffers,Effects, and textures?)

-What would happen if I continuously  unload and reload the same model Model?  
(when the player navigates back and forth the menus for example)

(the same applies to any custom assets that isn't disposable but contains native resources, not just Models)

tdeeb:
-I'm not the most knowledgeable about models, but I just went through the code and it doesn't look to unload any of those. The currently existing `Unload` method won't do so either. The user has to free these up manually.

-It would keep creating new Effects, etc. and likely cause memory leaks, should the user fail to manually free up the additional resource the model uses.

-If custom assets contain native resources, they should be implementing `IDisposable`. See [here](https://docs.microsoft.com/en-us/dotnet/api/system.idisposable?view=netframework-4.7.2). Failure to do so gives them no standardized mechanism of releasing them and they'll have to find another way to properly unload everything, either by manually doing it or overriding these methods.

Jjagg:
> The currently existing Unload method won't do so either. The user has to free these up manually.

No, the content manager used to load the model will contain all disposable resources the model requires like effects and textures, so they're disposed when the content manager is unloaded.

> If custom assets contain native resources, they should be implementing IDisposable

They should after this change, but just like Model assets that can only be created through the content pipeline could depend on the unloading behavior instead of implementing IDisposable.

mrhelmut:
I think that unloading models should only unload the vertex buffer to give the control to the developer to handle the effect/texture on their own if need be (thinking about shared effect/texture). Which is what this PR does.

tdeeb:
What modifications need to be made to get this merged?

BigBenEco:
As far as I understand the C# language, I don't think this code is safe to use.

While it will remove the item from memory, it does not remove any pointers to it, which could cause null point exceptions. 

This is because the way the content manager works, any time a part of the code wants to load the asset, (say for example, grass in a game, an object containing location and texture) it checks to see if it already has the asset. If yes, it just returns the pointer to the same asset. (pointer, pass by reference, asset is a generic, generics are classes, classes are only ever handled by reference). So say, I have lots of grass objects, and I go into a house, so I unload a scene, we have to make sure at no point does the grass object act on or off of the texture after it gets loaded, every object owning a pointer to the asset must remove that connection or the parent object must be deleted (disconnected and forced garbage collected, maybe even defragged by a second forced collection), or by some other trick make sure the texture is not used ever again, so that you can then unload the asset safely knowing that no other objects have fantom memories of the asset. Generally, the idea is not idiot-proof, that doesn't make the code bad, but it does make it an issue.

Basically, you have to be smart and safe with the unloading.
You could modify the content manager such that it keeps track of all connections, and removes them when unloading, but that does not help with if an object wants to try to access the asset again for some reason in the code, unloading is generally not safe.

A typical workaround that is kinda safe, it seems many users make a unique content manager for each unique scene, which they tell to unload or rather, dispose, at the end of the scene. But that still requires users to design scene so that all the objects are destroyed as well. The only issue with this, however, is now you have to reload all assets, none carry over. A bit more load time at the risk of memory conservation. though I would also garbage-collect and defragment at this time as well.

jnoyola:
Not sure if this has been on anyone’s mind lately, but I’m also curious about what this needs or what decisions need to be made to be able to merge.

> While it will remove the item from memory, it does not remove any pointers to it, which could cause null point exceptions.

I don’t think this should be a concern, because unloading the entire ContentManager has the same problem. It’s up to the developer to manage either way.

mrhelmut:
It's up to the developer to manage the lifecycle of objects and the content manager is no substitute to this.

I believe that this PR is as good as it can get, though it would be nice to update the documentation to highlight the fact that unloading a ```Model``` will not unload its associated content (e.g. textures, shaders) because they might be shared.

jnoyola:
> unloading a Model will not unload its associated content (e.g. textures, shaders)

What are the options for manually unloading these? I haven't looked through how associated content is loaded, but would it be possible to unload these with an additional call to `ContentManager.UnloadAsset("texture")` despite it not having been loaded like that? Or is the only way to unload the entire `ContentManager`?

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -471,6 +471,58 @@ public virtual void Unload()
...
+
+                loadedAssets.Remove(assetName);
+            }
+        }
+
+        /// <summary>
+        /// Unloads a set of assets.
+        /// </summary>
+        /// <param name="assetNames">The names of the assets to unload.</param>
+        public virtual void UnloadAssets(IList<string> assetNames)
```
jnoyola, line 509: For usability, can this be `params string[] assetNames` instead?

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -471,6 +471,58 @@ public virtual void Unload()
 		    loadedAssets.Clear();
 		}
 
+        /// <summary>
+        /// Unloads a single asset.
+        /// </summary>
+        /// <param name="assetName">The name of the asset to unload. This cannot be null.</param>
+        public virtual void UnloadAsset(string assetName)
```
jnoyola, line 478: Does overloading `Unload` make more sense than `UnloadAsset`, since the load method is `Load`, not `LoadAsset`?

mrhelmut:
It depends on the setup, but basically yes. Either resources are internal to a model (and are therefore freed with the model) or they are external (and are not freed but can be unloaded through the content pipeline by naming them manually).

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -471,6 +471,58 @@ public virtual void Unload()
...
+            //Check if the asset exists
+            object asset;
+            if (loadedAssets.TryGetValue(assetName, out asset))
+            {
+                //Check if it's disposable and remove it from the disposable list if so
+                var disposable = asset as IDisposable;
+                if (disposable != null)
+                {
+                    disposable.Dispose();
+                    disposableAssets.Remove(disposable);
```
jnoyola, line 498: I'm not sure why `disposableAssets` is a list, since it seems like it's only used as a set. Should it be changed to a set for perf here?

mrhelmut (approved):

mrhelmut:
@Jjagg @harry-cpp May I request a review to move forward? It's good to go to me, and could be improved later on demand.

Jjagg:
I'm fine with merging this as is.
@harry-cpp?

PR #6664, opened 2019-02-16
Reduce garbage from TextInput

### LINKED ISSUE #6360: Each keystroke produces some garbage
#### Symptoms:
Each keystroke produces garbage. If you hold the key, the garbage will seem to be generated each time a repeated keystroke is reported by the OS.

#### How to reproduce:
Download the attached .NET Core 2.1 project that contains some diagnostic code: https://phobia.games/experiments/Game1Core.zip

1. Run the project.
2. It should report no allocations per frame.
3. Press (and hold) any alphanumeric key, the program should report some allocations.

#### Possible cause:
https://github.com/MonoGame/MonoGame/blob/b1a49e4ae60eaf62169d08c791d4f9a913b495b3/MonoGame.Framework/SDL/SDLGameWindow.cs#L302

That line creates an object each time a system's key stroke event is handled, regardless of whether the programmer actually utilizes the `Window.TextInput` event or not.

- MonoGame 3.7

- Windows 10

- DesktopGL

mrhelmut:
Hey there,

This is a proposal to start addressing #6360.

It removes the string manipulations and decodes each UTF8 characters individually and sends them to the event handler.
It also skips the event processing if there is no event handler attached to ```GameWindow.TextInput```.

This should:
- remove all garbage if ```TextInput``` isn't used when pressing keys (which I believe is the most important fix here)
- greatly reduce garbage when actually using ```TextInput```

There is still some garbage generated from this line when attaching an handler to ```TextInput```:
https://github.com/MonoGame/MonoGame/blob/b1a49e4ae60eaf62169d08c791d4f9a913b495b3/MonoGame.Framework/SDL/SDLGameWindow.cs#L302

We will have to figure out (in another PR?) what to do with event args types.

Also, it should be noted that this PR comes with a limitation (note that the original code also has this limitation): it doesn't support UTF8 characters that are encoded on more than 2 bytes (because ```TextInputEventArgs``` is limited by the ```char``` type). Chinese, Japanese, and Korean won't be supported.

This should be addressed in another PR I guess (at the same time as addressing the event args type?).

Partly fixes #6360

### DIFF
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -103,7 +103,7 @@ namespace Microsoft.Xna.Framework {
 		/// </remarks>
 		public event EventHandler<TextInputEventArgs> TextInput;
 
-        public bool IsTextInputHandled { get { return TextInput != null; } }
+        internal bool IsTextInputHandled { get { return TextInput != null; } }
 #endif
 
 		#endregion Events
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -102,6 +102,8 @@ namespace Microsoft.Xna.Framework {
 		/// This event is only supported on the Windows DirectX, Windows OpenGL and Linux platforms.
 		/// </remarks>
 		public event EventHandler<TextInputEventArgs> TextInput;
+
+        public bool IsTextInputHandled { get { return TextInput != null; } }
 #endif
 
 		#endregion Events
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/GameWindow.cs
```diff
@@ -102,6 +102,8 @@ protected GameWindow()
 		/// This event is only supported on the Windows DirectX, Windows OpenGL and Linux platforms.
 		/// </remarks>
 		public event EventHandler<TextInputEventArgs> TextInput;
+
+        public bool IsTextInputHandled { get { return TextInput != null; } }
```
cra0zy: Shouldn't this be internal?

mrhelmut: You are right, I've updated this.

cra0zy:
Thanks @mrhelmut , merging!

PR #6666, opened 2019-02-17
Fix #6336

### LINKED ISSUE #6336: Error with BoundingSphere.Intersects(BoundingBox)
Created a Sphere with `BoundingSphere Bound = new BoundingSphere(new Vector3(0, 5, 0), 1);`
and a Box with `BoundingBox BoxBound = new BoundingBox(-Vector3.One, Vector3.One);`.
When I decrease the Y-value of the Bound, `Bound.Intersects(BoxBound)` will never return True when the intersect.

`Bound.Contains(BoxBound) == ContainmentType.Intersects` on the otherhand will return true.

- MonoGame 3.7

- Windows

- DesktopGL

lithiumtoast:
See #6336. I used C# 7 deconstructor here. If C# 7 features are not yet allowed I can modify the solution.

### DIFF
--- a/MonoGame.Framework/BoundingBox.cs
+++ b/MonoGame.Framework/BoundingBox.cs
@@ -385,13 +385,13 @@ namespace Microsoft.Xna.Framework
         public bool Intersects(BoundingSphere sphere)
         {
             var squareDistance = 0.0f;
-            var (x, y, z) = sphere.Center;
-            if (x < Min.X) squareDistance += (Min.X - x) * (Min.X - x);
-            if (x > Max.X) squareDistance += (x - Max.X) * (x - Max.X);
-            if (y < Min.Y) squareDistance += (Min.Y - y) * (Min.Y - y);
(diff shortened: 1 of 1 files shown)


lithiumtoast:
@Jjagg There is no test I can find that uses the `Intersects` method. Also why is this build failing?

mrhelmut:
You can check the build/test log on the teamcity details for each checks.

It fails because C# 7 isn't within the MonoGame requirements. You'll have to break down the ```var (x, y, z)``` statement.

Jjagg:
@lithiumtoast Thanks for this!

There is no test for it yet. Related tests are in [Bounding.cs](https://github.com/MonoGame/MonoGame/blob/develop/Test/Framework/Bounding.cs).

lithiumtoast:
Should I add test then to Bounding.cs?

Jjagg:
Yes, please do :)

Jjagg:
Gonna merge this. Would still be nice to get a test in though. Thanks @lithiumtoast!

PR #6669, opened 2019-02-17
[Templates] Hide console for DesktopGL

cra0zy:


Jjagg:
Hmm, not sure we should add NSubsys to the templates. Maybe add a doc page about it instead? Console can be nice for debugging and without an explanation we will probably get some people that don't know how to un-remove the console annoyed by this.

cra0zy:
> Hmm, not sure we should add NSubsys to the templates. Maybe add a doc page about it instead? Console can be nice for debugging and without an explanation we will probably get some people that don't know how to un-remove the console annoyed by this.

This only patches the final executable with `dotnet publish` to be a Windows executable and not a Console executable.

cra0zy:
@Jjagg ping

PR #6671, opened 2019-02-22
Add HKCU traversal for font descriptions

KingofBeast:
This PR makes FontDescriptionProcessor also traverse HKCU if the font isn't found in HKLM

Windows 1809 changed font installation behavior. Now fonts are installed by default in the local user's AppData and subsequently registered in HKCU instead of HKLM. This was made presumably so that non-administrator accounts may install custom fonts:

New context menu options

Registry


tomspilman:
This makes sense to me.

KingofBeast:
Anything keeping this from being merged?

tomspilman:
@Jjagg  or @cra0zy ... would be good for one of you to take a look at it before merging.

cra0zy:
Looks good to me, merging!

PR #6675, opened 2019-03-01
[MacOS] Update nuspec to point to correct libSDL2-2.0.0.dylib

CartBlanche:


cra0zy:
Project file definition needs to be fixed to select the correct native lib, not the nuspec file.

CartBlanche:
OK looking into it.

CartBlanche:
@cra0zy The MacOS project looks to be pointing to the right file
`..\ThirdParty\Dependencies\SDL\MacOS\Universal\libSDL2-2.0.0.dylib` 
How can I be 100% sure?

CartBlanche:
@cra0zy As stated on #Maintainers The issue isn't that it is being copied into the top level directory (this works), the issue is that it **isn't** copying those 2 files into the `Contents/MonoBundle` directory which exists inside the MacOS App bundle. Does that make sense?

cra0zy:
> the issue is that it **isn't** copying those 2 files into the `Contents/MonoBundle` directory which exists inside the MacOS App bundle. Does that make sense?

How is this fixing it in that case, you are just referencing the DesktopGL output files?

CartBlanche:
> How is this fixing it in that case, you are just referencing the DesktopGL output files?

Well I'm trying to centralise 1 set of lib* files, so that we don't end up in the situation where we might be using different versions of the same lib. So this PR is primarily to fix that issue. I'm not 100% sure this would indeed fix the 2nd issue, which is the copying issue. That's why I asked the question above. Atm, from what I can tell is **no** lib* files are copied into `Contents\MonoBundle` directory during a MacOS build with the 3.7.1 nugets. If it is indeed a build issue, then that would need to be investigated and another PR raised.

cra0zy:
> Well I'm trying to centralise 1 set of lib* files, so that we don't end up in the situation where we might be using different versions of the same lib.

Nuget packaging should not be using an output library of another project file, you can reference the original source library if you want, but we won't be adding 2 version of the same lib anyways...

> I'm not 100% sure this would indeed fix the 2nd issue, which is the copying issue.

If its not fixing it we can always manually copy the library like what we are doing for DesktopGL nuget.

CartBlanche:
> If its not fixing it we can always manually copy the library like what we are doing for DesktopGL nuget.

@cra0zy Could you point me to where this is being done (is it the .targets file?) and I'll test it locally for MacOS builds. I'm new to messing around with builds, so any pointers would be helpful.

cra0zy:
Yes, its a targets file: https://github.com/MonoGame/MonoGame/blob/develop/NuGetPackages/build/DesktopGL/MonoGame.Framework.DesktopGL.targets

harry-cpp:
Gonna close this as its outdated.

harry-cpp:
Also deleted the branch as it was created on the main repo and not a fork.

PR #6677, opened 2019-03-05
Fix for loading native libraries on Amazon Fire HD 8

tomspilman:
This fixes loading native libraries on some Android devices.

In my case the Amazon Fire HD 8 tablet was not loading `libopenal32.so` without this change to explicitly load from the library path.

I also added a little logging to make it easier to spot issues when working from the Android console log.

### DIFF
--- a/MonoGame.Framework/Utilities/FuncLoader.Android.cs
+++ b/MonoGame.Framework/Utilities/FuncLoader.Android.cs
@@ -1,4 +1,6 @@
+using Android.App;
 using System;
+using System.IO;
 using System.Runtime.InteropServices;
 
 namespace MonoGame.Utilities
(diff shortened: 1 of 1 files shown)


tomspilman:
@dellis1972 - Any opinions on this?

MonoGame.Framework/Utilities/FuncLoader.Android.cs
```diff
@@ -15,7 +17,27 @@ internal class FuncLoader
 
         public static IntPtr LoadLibrary(string libname)
         {
-            return dlopen(libname, RTLD_LAZY);
+            // Let the OS search for the library by default.
+            IntPtr lib = dlopen(libname, RTLD_LAZY);
+            if (lib != IntPtr.Zero)
+            {
+                Console.WriteLine("FuncLoader.LoadLibrary {0}", libname);
```
cra0zy: We should replace this with `Debug.WriteLine`.

tomspilman: Will that work in shipping games?  I ask because ideally if a player or service provides a log from shipped device we would want to know if these fail to load.

cra0zy: Hmm, I though you wanted this as debug information. I guess this is fine than.

PR #6678, opened 2019-03-05
Fix #6676

### LINKED ISSUE #6676: Content Pipeline Tool crashes when adding an assembly that references another assembly
- MonoGame 3.7, 3.7.1, 3.8 (development branch)

- Windows

- DesktopGL

I have an assembly that uses Newtonsoft.Json for a custom importer. When I attempt to add the assembly as a reference in the content pipeline tool, I get a crash with the following exception:

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.IO.FileNotFoundException: Could not load file or assembly 'Newtonsoft.Json, Version=12.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The system cannot find the file specified.
```

If I edit the Content.mgcb file manually to add the reference, I simply get an error window that says, "Content.mgcb: Failed to open the project due to an unknown error."

This previously worked under the same conditions with MonoGame 3.6. I'm currently using a relative path to my assembly. The NewtonSoft.Json assembly is in the same folder. I have also tried absolute paths.

BrettStory:
See #6676. I noticed that _currentAssemblyDirectory is used explicitly for the CurrentDomain_AssemblyResolve event, but unfortunately that event does not get called until after the current assembly is marked null. I think some recent changes to how the pipeline tool loads assemblies offloaded the resolution of referenced assemblies until one of their processors is constructed through Activator.CreateInstance(...). An issue I see with this going forward is that referenced assemblies will only be loaded correctly if they exist in the same directory as the referencing assembly. If anyone can think of a more robust solution, I'm all ears, but I'm also not super up to date with assembly resolution logic.

I apologize if I performed any steps in the issue creation or pull request process, I haven't really contributed to a repository that is not my own before.

### DIFF
--- a/Tools/Pipeline/Common/PipelineTypes.cs
+++ b/Tools/Pipeline/Common/PipelineTypes.cs
@@ -166,6 +166,7 @@ namespace MonoGame.Tools.Pipeline
         {
             public ContentProcessorAttribute Attribute;
             public Type Type;
+            public string Path;
         }
 
         private static List<ImporterInfo> _importers;
@@ -305,7 +306,9 @@ namespace MonoGame.Tools.Pipeline
             cur = 0;
             foreach (var item in _processors)
             {
+                _currentAssemblyDirectory = Path.GetDirectoryName(item.Path);
                 var obj = Activator.CreateInstance(item.Type);
+                _currentAssemblyDirectory = null;
                 var typeProperties = item.Type.GetProperties(bindings);
                 var properties = new List<ProcessorTypeDescription.Property>();
                 foreach (var i in typeProperties)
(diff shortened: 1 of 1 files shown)


BrettStory:
There was a bit of a discussion about this pull request in Discord. There was a bit of an agreement that instead of contiuously patching this broken feature until its fixed, it might be worth reverting to before commit 9d7d5b78 and only implementing the dynamic reference handling when it works for all scenarios, since it is has been breaking a fairly normal use case since its implementation. I believe @cra0zy was there for said conversation.

PR #6682, opened 2019-03-18
Adding capacity parameter to SpriteBatch ctor

mrhelmut:
Hey there,

I'd like to propose the addition of a parameter to the ```SpriteBatch``` constructor to have a control over its initial batch items capacity. The intention is to give a finer control of the memory allocations and garbage generation to developers (e.g. increasing the batch size right away to avoid any runtime array resizing).

This PR shouldn't break anything and should remain backward compatible.

@Jjagg @cra0zy @tomspilman

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -37,8 +37,9 @@ namespace Microsoft.Xna.Framework.Graphics
         /// Constructs a <see cref="SpriteBatch"/>.
         /// </summary>
         /// <param name="graphicsDevice">The <see cref="GraphicsDevice"/>, which will be used for sprite rendering.</param>
+        /// <param name="capacity">The initial capacity of the internal array holding batch items (the value will be rounded to the next multiple of 64).</param>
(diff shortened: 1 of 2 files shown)


Jjagg:
I like this. Thanks @mrhelmut, merging!

cra0zy:
This is an API change, we shouldn't merge stuff like this without @tomspilman or others consent.

Jjagg:
It's not breaking, straightforward and very minor so I figured it was fine, but yeah.

tomspilman:
This is ok...  Very easy thing to add that shouldn't cause any troubles on console.

cra0zy:
I sure do hope somebody gets the reference XD

AimedNuu:
Hi, it actually broke some compatibility with EmptyKeys GUI framework. Bad thing is that its source code is not open. But with this version of code and EmptyKeysUI_MonoGame version 3.1.0 from nuget I get
`Method not found: 'Void Microsoft.Xna.Framework.Graphics.SpriteBatch..ctor(Microsoft.Xna.Framework.Graphics.GraphicsDevice)'."} | System.MissingMethodException`

My assumption is that method signature within MonoGame.Framework.dll changes enough to brake compatibility at least for EmptyKeys but it might be the same case for more libraries out there

PR #6686, opened 2019-03-23
Revert #6202

cra0zy:
Creates way more problems than it solves, so reverting it until someone can implement a better version.


cra0zy:
@Jjagg Can you take a look at this?

Jjagg:
Yeah, merging!

PR #6687, opened 2019-03-23
[Templates] Update shared library template

cra0zy:


cra0zy:
@Jjagg Can you take a look at this?

Jjagg:
Will test this later today!

Jjagg:
The Content.mgcb file has Windows set as platform. Shouldn't that use the Platform property?

cra0zy:
> The Content.mgcb file has Windows set as platform. Shouldn't that use the Platform property? Other than that this is good!

No, the platform during compilation gets set from the command line, the platform set in Content.mgcb is irrelevant.

cra0zy:
> Shouldn't that use the Platform property?

Also not sure how you would do that.

Jjagg:
> No, the platform during compilation gets set from the command line, the platform set in Content.mgcb is irrelevant.

Ah, right. Merging!

PR #6697, opened 2019-04-03
Catch ArgumentException in MGCB Pipeline GUI

minimalism:
using CurrentCulture "en-SE" yields OEMCodePage 1, which, when passed in to Encoding.GetEncoding on PipelineController.cs ln 504, throws ArgumentException.
This causes all content building via the MGCB GUI to fail silently.

It's not clear to me why this method sometimes generates ArgumentException instead of NotSupportedException, but the docs ( https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getencoding?view=netframework-4.7.2 ) recommend catching both.

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -499,9 +499,16 @@ namespace MonoGame.Tools.Pipeline
         private void DoBuild(string commands)
         {
             Encoding encoding;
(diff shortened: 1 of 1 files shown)


minimalism:
Looks like this was fixed here https://github.com/MonoGame/MonoGame/commit/cf6394d6446c32e147149932a5559d151166bbce

Issue #6703, opened 2019-04-06
PngWriter - ZlibStream dropping data


juanjp600:
I have an image that gets truncated by the ZlibStream class when I save it to a stream using Texture2D.SaveAsPng.

Input:
![img](https://user-images.githubusercontent.com/8607119/55673800-2b2f7000-5883-11e9-9771-dd91e6dd1e14.png)

Result after saving to a FileStream using Texture2D.SaveAsPng (note the missing bottom portion):
![result](https://user-images.githubusercontent.com/8607119/55673809-4e5a1f80-5883-11e9-8463-53531069a5ac.png)

On top of that, attempting to load this image in MonoGame afterwards will cause the program to hang because of failed decompression.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:
- WindowsDX

Jjagg:
I think the reported size in the png does not match the pixel data in the file. The imagedata for the bottom portion is missing. Some image viewers cannot succesfully read the file, others will use a zero color to display the missing data. Depending on how the image is loaded and rendered in the used image viewer, you might not see that the data is missing.

However, opening it in `feh` (it uses libpng) gives me the following:
![image](https://user-images.githubusercontent.com/14875382/62947078-2c991800-bde2-11e9-955f-ccc93fcdb300.png)

It also reports: `libpng error: Not enough image data`

juanjp600:
The first image on the issue should have the complete pixel data: https://user-images.githubusercontent.com/8607119/55673800-2b2f7000-5883-11e9-9771-dd91e6dd1e14.png

The second image is what MonoGame outputs when using Texture2D.SaveAsPng, the changes made in #6704 allow it to correctly write all the data.

Jjagg:
Ah, i got the image from the linked issue in the barotrauma repo. I'll give this another shot.

rds1983:
@Jjagg,
I dont want my opinion to look like promotion of StbImageWriteSharp(which I maintain) and yet I think it would make sense to completely remove PngWriter and replace it with StbImageWriteSharp.

Jjagg:
I agree.

Jjagg:
@juanjp600 this should be resolved with #6867. The development build should be updated in a bit. Please give it a try if you can :)

juanjp600:
Tested, it seems to be working correctly now :)

Jjagg:
[...]

PR #6704, opened 2019-04-06
Fix compression in PngWriter

juanjp600:
See #6703 

Note: this fix is a bit hacky, it exposes an unused private method to force ZlibStream to finish deflating.


juanjp600:
Better fix introduced in #6867

Issue #6690, opened 2019-03-27
MathHelper.ToRadians precision error


MaKiPL:
MathHelper.ToRadians function currently returns **float**, but it should return **double**. 
Giving an example of 270 degrees the result will differ by float and double:
float:      4.712389
double: 4.71238898038469
As we can see the float ceils the number. The issue appears when we will be trying to calculate cosinus from these radian results- using System.Math.Cos from 270 degrees the results vary by the sign:
float->      1.9615310081256202e-08
double-> -1.8369701987210297e-16

as we can see the value reported from float is positive, while double one is negative. This yields errors in Matrix calculations for example in RotationMatrix

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7 (Development build)

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

tomspilman:
I don't have an issue with adding a new `ToRadiansPrecise` or `ToRadiansD` to the API for people to opt-in to.  We could then use it internally in places to improve precision when needed.

PR #6705, opened 2019-04-06
Add more precise MathHelperD class to complement MathHelper

kimimaru4000:
This PR adds `ToRadiansD` and `ToDegreesD` methods to `MathHelper` that return more precise results. It also replaces `ToRadians` calls in `ModelProcessor` with `ToRadiansD`.

Related issue: https://github.com/MonoGame/MonoGame/issues/6690


MonoGame.Framework/MathHelper.cs
```diff
@@ -266,29 +266,55 @@ public static float SmoothStep(float value1, float value2, float amount)
...
         /// <returns>The angle in radians.</returns>
         /// <remarks>
-        /// This method uses double precission internally,
+        /// This method uses double precision internally,
         /// though it returns single float
         /// Factor = pi / 180
         /// </remarks>
         public static float ToRadians(float degrees)
         { 
             return (float)(degrees * 0.017453292519943295769236907684886);
```
Apostolique: Is there a good reason to not directly use a float here and get rid of the casting? Along with the `ToDegrees(float)` above?

tdeeb: I'm not entirely sure, but with the more precise variants available, we should be safe to remove the casts.

KakCAT:
Net Core 2 introduced System.MathF to complement System.Math, which imho is great idea as I always end up casting to float everything that comes out of System.Math, and you can easy replace Math with MathF via "using" to easily go back and forth between both precisions.

I'm wondering if it'd be wiser to create MathHelperD to complement MathHelper and add ToRadians there, instead of adding ToRadiansD to MathHelper.

tomspilman:
> I'm wondering if it'd be wiser to create MathHelperD to complement MathHelper and add ToRadians there, instead of adding ToRadiansD to MathHelper.

Interesting idea.

@cra0zy @Jjagg @dellis1972 ?

Apostolique:
I like that idea @KakCAT

cra0zy:
I have no opinion on it, both sound fine to me.

tdeeb:
The latest commit has MathHelperD and MathHelperDTest. I kept the int Clamp overloads for consistency. In MathHelperDTest, I removed the XNA compatibility tests since XNA does not have a similar class and the double precision would likely result in them failing.

tdeeb:
The builds must've failed because the new file isn't being included. What's the proper way to add a new file to the build?

mrhelmut:
They must be added to Protobuild project definitions.

Main library: https://github.com/MonoGame/MonoGame/blob/develop/Build/Projects/MonoGame.Tests.definition

Tests: https://github.com/MonoGame/MonoGame/blob/develop/Build/Projects/MonoGame.Framework.Net.definition

MonoGame.Framework/MathHelper.cs
```diff
@@ -266,13 +266,13 @@ public static float SmoothStep(float value1, float value2, float amount)
         /// <param name="radians">The angle in radians.</param>
         /// <returns>The angle in degrees.</returns>
         /// <remarks>
-        /// This method uses double precission internally,
+        /// This method uses double precision internally,
```
mrhelmut: For backward compatibility, I think we should leave the orignal MathHelper as-is and that internal double precision.

MonoGame.Framework.Content.Pipeline/Processors/ModelProcessor.cs
```diff
@@ -99,9 +99,9 @@ public override ModelContent Process(NodeContent input, ContentProcessorContext
             // Perform the processor transforms.
             if (RotationX != 0.0f || RotationY != 0.0f || RotationZ != 0.0f || Scale != 1.0f)
             {
-                var rotX = Matrix.CreateRotationX(MathHelper.ToRadians(RotationX));
-                var rotY = Matrix.CreateRotationY(MathHelper.ToRadians(RotationY));
-                var rotZ = Matrix.CreateRotationZ(MathHelper.ToRadians(RotationZ));
+                var rotX = Matrix.CreateRotationX((float)MathHelperD.ToRadians(RotationX));
```
mrhelmut: Model processor changes should be in another PR. Also, I don't quite get the relevance of casting a double to float here while the original MathHelper does the same thing internally?

MonoGame.Framework/MathHelper.cs
```diff
@@ -266,13 +266,13 @@ public static float SmoothStep(float value1, float value2, float amount)
...
         /// though it returns single float
         /// Factor = 180 / pi
         /// </remarks>
         public static float ToDegrees(float radians)
         { 
-            return (float)(radians * 57.295779513082320876798154814105);
```
cra0zy: `57.295779513082320876798154814105` is not the same as `57.295779513082320876798154814105f`, the later is less precise. Return the original code.

MonoGame.Framework/MathHelper.cs
```diff
@@ -281,13 +281,13 @@ public static float ToDegrees(float radians)
...
         /// <returns>The angle in radians.</returns>
         /// <remarks>
-        /// This method uses double precission internally,
+        /// This method uses double precision internally,
         /// though it returns single float
         /// Factor = pi / 180
         /// </remarks>
         public static float ToRadians(float degrees)
         { 
-            return (float)(degrees * 0.017453292519943295769236907684886);
```
cra0zy: Same as above comment.

MonoGame.Framework/MathHelper.cs
cra0zy: For reference on what actually happens to constant numbers: https://dotnetfiddle.net/aBz4fx

tdeeb:
Pushed requested changes, but the MathHelperD tests are failing. While I can fix the expected Pi value tests, I'm not entirely sure on the best values to use for the WrapAngle tests.

tdeeb:
@cra0zy The requested changes have been added and tests pass.

tdeeb:
Any additional feedback on this?

BenMakesGames:
With new C# goodness, MonoGame can "auto-implement" many of its custom mathy methods for many numeric types at once; no need to do `F`/`D` suffixes in C# ever again.

For example:

```
public static T Hermite<T>(T value1, T tangent1, T value2, T tangent2, T amount)
    where T: IFloatingPoint<T>, IMultiplyOperators<T, T, T>
{
    if (amount == T.Zero)
        return value1;

    if (amount == T.One)
        return value2;

    var two = T.One + T.One;
    var three = two + T.One;

    return (two * value1 - two * value2 + tangent2 + tangent1) * amount * amount * amount +
           (two * value2 - three * value1 - two * tangent1 - tangent2) * amount * amount +
           tangent1 * amount +
           value1;
}
```

skyebird189:
This is for .NET 7 only. Since current MG (rightfully so) targets LTS .NET 6, the new generic math is not an option.

tomspilman:
We don't need this.  In the future MG may move to double as the primary math type (like Unreal does).

PR #6709, opened 2019-04-09
Fixes reflection support on SpriteBatch ctor

### LINKED ISSUE #6708: Backwards compatibility problem caused by #6682
https://github.com/MonoGame/MonoGame/pull/6682
Hi, it actually broke some compatibility with EmptyKeys GUI framework. Bad thing is that its source code is not open. But with this version of code and EmptyKeysUI_MonoGame version 3.1.0 from nuget I get
`Method not found: 'Void Microsoft.Xna.Framework.Graphics.SpriteBatch..ctor(Microsoft.Xna.Framework.Graphics.GraphicsDevice)'."} | System.MissingMethodException`

This happens when I call MonoGameEngine constructor which initialize EmptyKeys GUI stuff and creates SpriteBrach instance(s) internally.
My assumption is that method signature within MonoGame.Framework.dll changes enough to break compatibility at least for EmptyKeys but it might be the same case for more libraries out there.

mrhelmut:
Fixes #6708

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -36,10 +36,19 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <summary>
         /// Constructs a <see cref="SpriteBatch"/>.
         /// </summary>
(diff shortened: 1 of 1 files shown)


tomspilman:
Thanks @mrhelmut !

PR #6712, opened 2019-04-13
We are able to type CJK characters in MonoGame now.

ryancheung:
This commit may not be feature-complete for IME support, but workable at least, and it need to test on different OS platform and versions.

Need more support on **Text Composition** and **Candidate List** 

Test environment: 
1. DirectX - Win10, IME works perfectly.
2. SDL- Windows, IME works, but no candidate list windows showing up, which is a bug in SDL2 https://bugzilla.libsdl.org/show_bug.cgi?id=3421


ryancheung:
I’m closing this PR since there will be a more refined PR in the future.

PR #6729, opened 2019-05-11
Fix for #6506

### LINKED ISSUE #6506: MonoGame Android 3.7.0.1708 : Texture2D.GetData corrupts the Texture2D object being read
I've been using CocosSharp successfully with Monogame 3.6. However, when upgrading to 3.7 (or the latest 3.8 dev build), the following code is now corrupting loaded texture data

`var b = new Color[rt.Texture.PixelsHigh * rt.Texture.PixelsWide];
                rt.Texture.XNATexture.GetData<Color>(b); `
                

It appears, rather than just reading out the data, to completely overwrite the texture with a section from a corner of the current screen.

To emphasise: this isn't just a case of it returning the wrong data. The texture being read becomes *corrupted and overwritten* entirely.

It seems the actual data being returned is correct. However, calling SetData on the texture straight afterwards has no effect.

- MonoGame.Framework.Android 3.7.0.1708

- MacOS Mojave 10.14

- Android

rds1983:
Fixes #6506.


cra0zy:
Ha, well that was a silly mistake in code... thanks @rds1983 , merging!

Issue #6728, opened 2019-05-04
GraphicsDevice.SetVertexBuffers generating garbage every frame due to params option


YTN0:
I came across this issue when I decided to profile my (in progress) game and discovered I was generating a ton of VertexBufferBinding[] arrays during the Draw() call.

It appears as though the API GraphicsDevice.SetVertexBuffers(params VertexBufferBinding[] vertexBuffers) is generating garbage every frame due to the params option. 

An example of using this API would be instancing viz. 

SetVertexBuffers(vertexBinding, instanceBinding);

The params option generates a new array on every call which then needs to be garbage collected.

There is currently a workaround i.e. creating a static array and passing that to the API e.g.

VertexBufferBinding[] bindings = new VertexBufferBinding[2];
bindings[0] = blah;
bindings[1] = blah;

<later in the Draw call>
SetVertexBuffers(bindings);

However, it's pretty easy to miss this issue, so it would probably be best to clean up the APIs to remove the params option and have a couple different API signatures e.g. one that takes a statically defined array and one that takes in both 1 VertexBufferBinding and 2 VertexBufferBindings, since those would be the most common use cases.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7.1

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL, WindowsDX

tomspilman:
>  remove the params option 

I think we have to keep that as i believe it was part of the original XNA API and we need it for compatibility.

However...

>  one that takes a statically defined array and one that takes in both 1 VertexBufferBinding and 2 VertexBufferBindings, 

We can do this and keep the params version for fallback if for some reason you happen to use more than 2 or 3 bindings.

Halofreak1990:
@tomspilman if we're keeping the params version, the first method suggested by @YTN0 becomes unnecessary; a params method already takes an array, and you can prevent the garbage generation by simply creating the array ahead of time. That leaves us with just the second overload to implement.

willmotil:
Are you sure i seem to remember params calls all work by object[] if i remember right unless it has changed. So essentially its boxing unboxing. I think when you pass a type even if it is global it can get boxed in the case of params casting anything to object i think gets boxed. I haven't ever had to use params i doubt i ever will i avoid it like i avoid arraylists. Like the plauge.

YTN0:
I am pretty sure. I did some googling on params and that is definitely the consensus. My profiling also indicated the same thing.

[...]

PR #6730, opened 2019-05-13
Fix for #6728 - GraphicsDevice.SetVertexBuffers(params VertexBufferBinding[] vertexBuffers) generating garbage on every frame

YTN0:
Fix for issue #6728 viz. GraphicsDevice.SetVertexBuffers(params VertexBufferBinding[] vertexBuffers) generating garbage on every frame due to params option.

I created a few new overloaded methods which uses a statically created VertexBufferBinding[] array. The original params version is still there, so it should not impact any legacy code.

My testing showed that this work, however, the Visual Studio IntelliSense documentation still showed the original params version when hovering over the method (not sure why). Profiling does show that the garbage is now gone with this change.

I don't believe I have broken anything, but someone who knows this stuff better probably wants to give this a more detailed check.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -55,6 +55,8 @@ namespace Microsoft.Xna.Framework.Graphics
         private IndexBuffer _indexBuffer;
         private bool _indexBufferDirty;
 
+        private VertexBufferBinding[] _vertexBufferBindings;
+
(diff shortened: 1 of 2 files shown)


YTN0:
Apologies for the minor hiccup.. I was working on another fix (for #6732) and had pushed that to my repo, but it got included into this PR (I hate that git does that). I forced my repo back to the previous commit so all should be good again. 

I will create a separate branch to work on #6732. This PR should be back to it's original (good) state.

YTN0:
@tomspilman  Am closing this PR and submitted a new one which is based off a branch so I can work on some other changes without conflicting.

PR #6734, opened 2019-05-18
Add a SpriteFont.MeasureString() overload accepting a ReadOnlySpan<char>

JoJoJet:
* Add the CharacterSourceSpan ref struct.

    * Wraps either a ReadOnlySpan<char> or a StringBuilder.

    * Changed SpriteFont.MeasureString() to use CharacterSourceSpan.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteFont.cs
+++ b/MonoGame.Framework/Graphics/SpriteFont.cs
@@ -177,11 +177,26 @@ namespace Microsoft.Xna.Framework.Graphics
 		/// this font.</returns>
 		public Vector2 MeasureString(string text)
 		{
-			var source = new CharacterSource(text);
+			var source = new CharacterSourceSpan(text.AsSpan());
 			Vector2 size;
 			MeasureString(ref source, out size);
 			return size;
 		}
+
+        /// <summary>
+        /// Returns the size of a Span of <see cref="char"/>s when
+        /// treated as a <see cref="string"/> and rendered in this font.
+        /// </summary>
+        /// <param name="text">The text to measure.</param>
+        /// <returns>The sizes, in pixels, of 'text' when rendered in
+        /// this font<./returns>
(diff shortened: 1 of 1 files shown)


Jjagg:
This is nice to have, but we'd need to upgrade to at least C# 7.2 (or 7.3 even, I'm not 100% sure) to get this building. I think that would cause issues for consoles. We could consider conditionally including things like this for desktop, but diverging APIs make porting to consoles harder, so I don't think we want to do that.

Cc @tomspilman

nkast:
@JoJoJet 
You can add a method that accepts an Enumerator<char>.
This will work universally for all types of char collections, including  ReadOnlySpan<char>.

Actually the first step would be to convert CharacterSource into an enumerator.

JoJoJet:
> @JoJoJet
> You can add a method that accepts an Enumerator<char>.
> This will work universally for all types of char collections, including ReadOnlySpan<char>.
> 
> Actually the first step would be to convert CharacterSource into an enumerator.

I suppose that would work, but I wanted to change as little of the code as possible. I don't understand `unsafe` code very well, so I didn't want to mess with the internal `MeasureString()` method.

willmotil:
The CharSource struct is useless you could remove it altogether and pass your string or stringbuilder right thru and operate on it directly and nothing would change. I still haven't seen a reason for it to exist.

You can also just copy measure string and keep it around as a utility function measure string shouldn't be accessing any pointers as it doesn't need to do any drawing just run off the measurement calculation.

Jjagg:
@willmotil you are 90% wrong. CharacterSource is internal and did not exist (publicly at least) in XNA. And just a reminder that none of the MG code came from XNA.

CharacterSource exists so we don't need duplicate implementations for string and StringBuilder.

Jjagg:
I'm going to close this for now. I still hope we can support higher version of C# in the future, but it definitely won't happen short term due to how native compilation works for consoles.

PR #6735, opened 2019-05-19
Fix for #6728. GraphicsDevice.SetVertexBuffers generating garbage every frame

YTN0:
Per #6728 ... This fixes the garbage generate by the params version of SetVertexBuffers().

@tomspilman  Replacing the previous PR I submitted for the same issue.

I moved the changes for this fix to a branch so I can work on some other changes without conflicting with these. The changes are exactly the same, and I tested and verified them again.

When you have some time, please re-target this to the appropriate release.

Thanks!

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -55,6 +55,8 @@ namespace Microsoft.Xna.Framework.Graphics
         private IndexBuffer _indexBuffer;
         private bool _indexBufferDirty;
 
+        private VertexBufferBinding[] _vertexBufferBindings;
+
         private readonly RenderTargetBinding[] _currentRenderTargetBindings = new RenderTargetBinding[4];
         private int _currentRenderTargetCount;
         private readonly RenderTargetBinding[] _tempRenderTargetBinding = new RenderTargetBinding[1];
@@ -324,6 +326,7 @@ namespace Microsoft.Xna.Framework.Graphics
 
             // Force set the buffers and shaders on next ApplyState() call
             _vertexBuffers = new VertexBufferBindings(_maxVertexBufferSlots);
+            _vertexBufferBindings = new VertexBufferBinding[_maxVertexBufferSlots];
             _vertexBuffersDirty = true;
             _indexBufferDirty = true;
             _vertexShaderDirty = true;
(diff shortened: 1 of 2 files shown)


YTN0:
@tom @Jjagg  @cra0zy Any chance we can get this targeted to 3.8? 

Much appreciated!

YTN0:
@tomspilman Could we get this assigned to 3.8?

YTN0:
@Jjagg @tomspilman Any thoughts on getting this PR into 3.8?

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -890,6 +893,54 @@ public void SetVertexBuffer(VertexBuffer vertexBuffer, int vertexOffset)
...
+            }
+        }
+
+        public void SetVertexBuffers(VertexBufferBinding vertexBuffer)
+        {
+            _vertexBufferBindings[0] = vertexBuffer;
+            _SetVertexBuffers(_vertexBufferBindings, 1);
+        }
+
+        public void SetVertexBuffers(VertexBufferBinding vertexBufferA, VertexBufferBinding vertexBufferB)
```
tomspilman, line 921: `vertexBuffer1`, `vertexBuffer2`, etc is a little more common to see than A,B,C.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -890,6 +893,54 @@ public void SetVertexBuffer(VertexBuffer vertexBuffer, int vertexOffset)
                                    ? _vertexBuffers.Clear()
                                    : _vertexBuffers.Set(vertexBuffer, vertexOffset);
         }
+
+        private void _SetVertexBuffers(VertexBufferBinding[] vertexBuffers, int bindingCount)
+        {
+            if (vertexBuffers == null || vertexBuffers.Length == 0 || bindingCount == 0)
```
tomspilman, line 899: Do you really need this path?  Only your overloaded `SetVertexBuffers` call this and it always passes a buffer and always has a count and length.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -890,6 +893,54 @@ public void SetVertexBuffer(VertexBuffer vertexBuffer, int vertexOffset)
...
+
+        private void _SetVertexBuffers(VertexBufferBinding[] vertexBuffers, int bindingCount)
+        {
+            if (vertexBuffers == null || vertexBuffers.Length == 0 || bindingCount == 0)
+            {
+                _vertexBuffersDirty |= _vertexBuffers.Clear();
+            }
+            else
+            {
+                if (vertexBuffers.Length > _maxVertexBufferSlots || bindingCount > _maxVertexBufferSlots)
```
tomspilman, line 905: Same...  only our internal overloads call this so it is never wrong.  Doesn't seem this is needed.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -890,6 +893,54 @@ public void SetVertexBuffer(VertexBuffer vertexBuffer, int vertexOffset)
...
+                {
+                    var message = string.Format(CultureInfo.InvariantCulture, "Max number of vertex buffers is {0}.", _maxVertexBufferSlots);
+                    throw new ArgumentOutOfRangeException("vertexBuffers", message);
+                }
+
+                _vertexBuffersDirty |= _vertexBuffers.Set(vertexBuffers, bindingCount);
```
tomspilman, line 911: Actually...  i think you don't need `_SetVertexBuffers` at all here.  This line....

`_vertexBuffersDirty |= _vertexBuffers.Set(vertexBuffers, bindingCount);`

Is the only bit you need.  Might was well put it directly in the overloaded `SetVertexBuffers` calls.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -890,6 +893,54 @@ public void SetVertexBuffer(VertexBuffer vertexBuffer, int vertexOffset)
...
+            }
+        }
+
+        public void SetVertexBuffers(VertexBufferBinding vertexBuffer)
+        {
+            _vertexBufferBindings[0] = vertexBuffer;
+            _SetVertexBuffers(_vertexBufferBindings, 1);
```
tomspilman, line 918: You should just put...

```
_vertexBuffersDirty |= _vertexBuffers.Set(_vertexBufferBindings, 1);
```

here!

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -890,6 +893,54 @@ public void SetVertexBuffer(VertexBuffer vertexBuffer, int vertexOffset)
...
+            _SetVertexBuffers(_vertexBufferBindings, 3);
+        }
+
+        public void SetVertexBuffers(VertexBufferBinding vertexBufferA, VertexBufferBinding vertexBufferB, VertexBufferBinding vertexBufferC, VertexBufferBinding vertexBufferD)
+        {
+            _vertexBufferBindings[0] = vertexBufferA;
+            _vertexBufferBindings[1] = vertexBufferB;
+            _vertexBufferBindings[2] = vertexBufferC;
+            _vertexBufferBindings[3] = vertexBufferD;
+            _SetVertexBuffers(_vertexBufferBindings, 4);
```
tomspilman, line 942: Same... just put...

`_vertexBuffersDirty |= _vertexBuffers.Set(_vertexBufferBindings, 4);`

... here.

tomspilman:
I'm good with merging this after some fixes.  @harry-cpp @Jjagg ?

Jjagg:
I prefer not having API like this for params in general (because users can pass an array anyway), but I don’t mind very much so I’m fine with accepting the changes.

tomspilman:
> I prefer not having API like this for params in general (because users can pass an array anyway)

You're saying because someone could pass their own array to `SetVertexBuffers(params VertexBufferBinding[] vertexBuffers)` ?

I can see what you mean.  However the params one is easily misused by new users and these overloads avoid the possibility of misuse.  And they are very simple:

```
        public void SetVertexBuffers(VertexBufferBinding vertexBuffer1, VertexBufferBinding vertexBuffer2)
        {
            _vertexBufferBindings[0] = vertexBuffer1;
            _vertexBufferBindings[1] = vertexBuffer2;
            _vertexBuffersDirty |= _vertexBuffers.Set(_vertexBufferBindings, 2);
        }
```

C# itself does this sort of thing under the hood in fact with `string.Concat` having 1, 2, and 3 argument overloads to optimize the simpler more common cases.

Jjagg:
Yeah, having this improved in the language/runtime would be a lot nicer.
There are some open issues related to params and garbage:

- https://github.com/dotnet/csharplang/issues/179
- https://github.com/dotnet/csharplang/issues/535

tomspilman:
We would merge this if it were fixed.  I added a bunch of feedback for changes on this PR back when it was submitted, but nothing was fixed.  

So closing this for now, but i would welcome a fix.

Issue #6732, opened 2019-05-16
Exposing the DrawInstancedPrimitives() instance base index


YTN0:
Does anyone have any objections to exposing (as an optional parameter) the instance base index in the call:

`DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount)`

Currently the instance base index is hardcoded to 0 inside of the private method `PlatformDrawInstancedPrimitives.`

My proposal is to modify DrawInstancedPrimitives to have the following signature:

`DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount, int instanceIndex = 0)`

This will preserve existing usage and behavior but allow you to optionally set a different base index for the instances.

Reason being is that I have a circular buffer of instance data and I want to draw different sections of it without having to also have separate and multiple SetData calls. With this change I can set the data once and have separate draw calls drawing from different sections of the set instance data.

#### What version of MonoGame does the bug occur on:
- All

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL, WindowsDX

willmotil:
Yes.

The question is though is why was it hardcoded to zero was that just done because in a hurry then forgotten ? or was it done for a reason ?

YTN0:
@willmotil  FYI, the hardcoded 0 base was only on the DirectX side of things. I am not sure the original reasoning.... but based on what I learnt from making the code changes, I would speculate that it's because the DirectX API allowed you to set a base instance index, but OpenGL had a separate API for that, and that API is only available from 4.2 and above.

Because earlier versions of the API didn't support setting the base instance index, it was probably simpler to just hardcode it to 0 for DirectX.

PR #6737, opened 2019-05-20
Fix for #6732 - Exposes the instancing base index

YTN0:
Per #6732 ... This change exposes the option to set the instance base index for instancing. This works perfectly on DirectX, and somewhat works on OpenGL. On OpenGL, this is still impacted by the instancing bug #6293. 

Note: For OpenGL, if not setting the base instance index (which is the default behavior), it will continue using the original OpenGL API. If setting the base, this will use the OpenGL API glDrawElementsInstancedBaseInstance which is only available in OpenGL 4.2 and higher.

FYI @Jjagg


YTN0:
FYI, protobuild isn't generating the web project, so I missed that in my initial checks. I did update that after the build failed... however it would be good to have the VS project for web also generated by protobuild so it isn't missed in any fixes / updates.

YTN0:
Anyone have any thoughts on this PR? I am hoping this can get into 3.8....

YTN0:
@Jjagg Would love to get some input on this... I am currently using a modified version of the Monogame dll with this patch, but would love to get it into the core platform. It's fully backward compatible.

Jjagg:
Hey @YTN0! Feel free to ping me more often if I don't respond for a while :) 

This is great to have IMO!
However [glDrawElementsInstancedBaseIndex](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDrawElementsInstancedBaseInstance.xhtml) requires OpenGL 4.2 which is a lot higher than the minimum for MG. If we're gonna add this it needs to be added to GraphicsCapabilities and we need to check if the GPU supports the feature when a base index is passed. We could implement the fallback ourselves when the feature is not supported, but either way we should have it added to GraphicsCapabilities.

YTN0:
@Jjagg Will take a look at GraphicsCapabilities and see if I can figure that stuff out.

Thanks.

YTN0:
@Jjagg Added support and check using GraphicsCapabilities. Let me know if it looks good and is what you had in mind.

YTN0:
@jjag Ugh... I rebased and looks like its now showing all the commits from the rebase. Not sure how to resolve that.

YTN0:
@Jjagg will close this PR and resubmit a new clean one based on the new develop changes.

PR #6741, opened 2019-05-20
fixes #6740

### LINKED ISSUE #6740: Android crashes on launch and resume (OpenAL related)
OnResumeGameThread is null when starting and OnPauseGameThread is null when resuming.
In both cases a null-ref is thrown.

- MonoGame 3.8

- Windows

- Android

Ragath:
Not sure if this just hides the symptoms of a bigger issue, but it does get android builds running again.
(did not have any audio in my tests).


Jjagg:
It's a good check to have either way. Thanks @Ragath, merging!

Issue #1880, opened 2013-07-22
Add Garbage Free BoundingBox.CreateFromPoints()


tomspilman:
The current `BoundingBox.CreateFromPoints()` looks like this...

```
public static BoundingBox CreateFromPoints(IEnumerable<Vector3> points)
{
```

While this is convenient because it can work with anything it.... it generates garbage because of the generic iterator loop.

We should add a List<Vector3> specialization of this function because list iteration is garbage free:

```
public static BoundingBox CreateFromPoints(List<Vector3> points)
```

YTN0:
@tomspilman  I recently ran into this same issue when profiling my code. For now, I just created a local clone of the BoundingBox.CreateFromPoints() API which takes a Vector3[] as input.

Is there a reason we don't just add a couple overloads to BoundingBox.CreateFromPoints() that takes in an array and List (as well as keep the original)?

i.e.

```
BoundingBox.CreateFromPoints(Vector3[] points);
and
BoundingBox.CreateFromPoints(List<Vector3> points);

```

willmotil:
Well if anyone tried to use this in a dynamic way they would be in for pure pain.

I used evil jukes test code and came out with 20 x the total memory footprint to 
![ejsgarbage test](https://user-images.githubusercontent.com/7495196/58312540-7ec92f00-7dd9-11e9-98a3-f31c81718eab.png)
monogame itself in collections over 1 minute.

nkast:
>> Back to the original topic, would IList<> as an overload be sufficient to
call the non-allocating enumerator rather than the specific List<>? Using
IList would potentially allow more containers to use it.​

>> Unfortunately no... only List or Vector3[] would work. This is one of this crappy cases where the best thing is messy.

With an IList<> you are not limited to use `foreach` which creates an enumerator.
There's the option to use a `for` loop and acess the list through an index.

We can plan ahead and add the full signature we are usually end up with when we work with lists. 
The one that has an index and a count. 

`public static BoundingBox CreateFromPoints(IList<Vector3> points, int index = 0, int count = -1)`

FYI, @YTN0

PR #6743, opened 2019-05-22
Fix for #1880. Added a couple overloads for a garbage-less BoundingBox.CreateFromPoints()

YTN0:
See issue #1880

### DIFF
--- a/MonoGame.Framework/BoundingBox.cs
+++ b/MonoGame.Framework/BoundingBox.cs
@@ -229,17 +229,61 @@ namespace Microsoft.Xna.Framework
         /// Create a bounding box from the given list of points.
         /// </summary>
         /// <param name="points">The array of Vector3 instances defining the point cloud to bound</param>
+        /// <param name="index">The base index to start iterating from</param>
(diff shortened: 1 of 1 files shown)


YTN0:
@cra0zy  Done!

cra0zy:
I meant doing the following:
```
public static BoundingBox CreateFromPoints(List<Vector3> points, int index = 0, int count = -1)
{
    if ( count == -1)
        count = points.Count;

    // Add code here
}
```

cra0zy:
(feel free to do it for the existing CreateFromPoints method as well)

YTN0:
@cra0zy  Done... again!

I didn't modify the original / existing API since it takes an IEnumerable<Vector3> and would require keeping a counter in the loop and skipping iterations until you hit the index and then counting from there. 

I'd rather not mess with the original API too much.

cra0zy:
> I'd rather not mess with the original API too much.

Definitely agree with that, didn't remember what the original API was using.

cra0zy:
This looks great, thanks @YTN0, merging!

Issue #6746, opened 2019-05-24
Float * Color vs Color * Float


kibblebits:
![image](https://user-images.githubusercontent.com/989840/58361656-49b2f000-7e5e-11e9-8a2e-29393ab0471d.png)

Compiler error when trying to compute a float * Color. It works fine when doing Color * float.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7.1.189

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

PR #6747, opened 2019-05-25
Added an additional multiply operator overload to Color with the scale first

tdeeb:
This PR adds a new overloaded multiply operator to `Color` so the scale can be specified first as a convenience. For instance, this allows the following when previously it would not compile: `Color color = 0.5f * Color.White`. This PR also uses this operator in the Color multiplication tests.

Relevant issue: https://github.com/MonoGame/MonoGame/issues/6746


cra0zy:
Looks good to me, thanks @tdeeb, merging...

PR #6749, opened 2019-05-28
[SDL] Fixes OnTextInput

mrhelmut:
Hey there,

This one is a bit hairy.

I noticed that my previous attempt to remove garbage from ```OnTextInput``` resulted in an encoding error. SDL returns UTF8-encoded multibyte characters while the C# ```char``` type is UTF16-encoded. The error was that my code assumed that ```char``` was UTF-8 encoded.

This PR fixes this by converting the UTF8 multibyte character properly to ```char```, while still trying to avoid garbage generation.

The hairy part is within ```UTF8ToUnicode```. Let me know if you'd like this to be refactored elsewhere.

Everything's now working properly within the supported ```char``` range.

@cra0zy @Jjagg


cra0zy:
I think the function is fine where it is, as I doubt it would be needed from somewhere else.

This looks fine to me and as such, merging...

PR #6760, opened 2019-06-19
Track GamePad packet number and general optimizations (fixes #5971)

### LINKED ISSUE #5971: GamePadState.PacketNumber stays 0 even when input is working.
Only issue I could find that might pertain is [https://github.com/MonoGame/MonoGame/issues/4002](#4002)

I've set up breakpoints on GamePadState.GetButtonPressed(Buttons.A) and it returns true but PacketNumber never updates.

cra0zy:
Stuff done:
- Replaced `if` with `switch` statement
- Added translation dictionary that translates from SDLs instanceId to our index
- Added Gamepad `PacketNumber` tracking

### DIFF
--- a/MonoGame.Framework/Platform/Input/GamePad.SDL.cs
+++ b/MonoGame.Framework/Platform/Input/GamePad.SDL.cs
@@ -5,8 +5,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
-using System.Reflection;
 using MonoGame.Utilities;
 
 namespace Microsoft.Xna.Framework.Input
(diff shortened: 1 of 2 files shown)


Jjagg:
Nothing changed functionally with this in SDLGamePlatform, right?

cra0zy:
Thee only change is:
```
case Sdl.EventType.ControllerButtonUp:
case Sdl.EventType.ControllerButtonDown:
case Sdl.EventType.ControllerAxisMotion:
    GamePad.UpdatePacketInfo(ev.ControllerDevice.Which, ev.ControllerDevice.TimeStamp);
    break;
```

Jjagg:
@cra0zy I should have merged this before, there are conflicts now. If you can rebase, I'll merge it in :)

cra0zy:
I was just about to fix it, I guess there is no longer a need :)

nkast:
>> Nothing changed functionally with this in SDLGamePlatform, right?

I don't know if it's significant but I found two changes in WindowEvent.
-The EventId.FocusLost is no longer checked,  IsActive is never set to false.
-Previously the code would check if  (WindowID == _view.Id) before doing any changes.

Jjagg:
You're right, both of those are important and need to be fixed.

cra0zy:
Window id check technically is neither important nor needed.

Jjagg:
It was implemented to fix #6432. It's a small effort to re-add the check.

cra0zy:
What was needed is:
```
else if (ev.Window.EventID == Sdl.Window.EventId.Close)
```

Checking window id on every event is redundant, if anybody wanted to add a custom SDL window they will need to modify MG source to add custom event handling either way.

Jjagg:
Oops, yes you're right. Nevermind the window id check then. That leaves us with handling the FocusLost event.

PR #6761, opened 2019-06-19
Add buffered input

Xaeroxe:
This is #6019 but updated to address the merge conflict.


Xaeroxe:
There appears to be some problems I didn't account for before. I'm going to close this while I work those out.

PR #6762, opened 2019-06-20
Buffered keys

Xaeroxe:
This is #6019 but rebased on the latest develop.

I'd like to get this reviewed and merged before it goes stale again. Buffered input is really important for my use case, and I imagine many others as well.


MonoGame.Framework/SDL/SDLGamePlatform.cs
```diff
@@ -129,13 +129,17 @@ private void SdlRunLoop()
                     if (!_keys.Contains(key))
                         _keys.Add(key);
                     char character = (char)ev.Key.Keysym.Sym;
+                    if (KeysHelper.IsKey((int)key))
```
cra0zy: KeyboardUtil.ToXna already returns a valid key enum so this will always return true.

MonoGame.Framework/WindowsUniversal/UAPGameWindow.cs
```diff
@@ -216,6 +217,27 @@ private void UpdateSize()
...
 
+        private void Dispatcher_AcceleratorKeyActivated(CoreDispatcher sender, AcceleratorKeyEventArgs args)
+        {
+            // NOTE: Dispatcher event is used becuase KeyDown event doesn't handle Alt key
+            var key = InputEvents.KeyTranslate(args.VirtualKey, args.KeyStatus);
+            switch (args.EventType)
+            {
+                case CoreAcceleratorKeyEventType.KeyDown:
+                case CoreAcceleratorKeyEventType.SystemKeyDown:
+                    if (KeysHelper.IsKey((int)key))
```
cra0zy: InputEvents.KeyTranslate should always return a valid key, so this should always be true.

Xaeroxe:
Thank you for your quick guidance @cra0zy . I'll work on applying your feedback!

Xaeroxe:
Ready for re-review

Xaeroxe:
CI isn't triggering on my commits any more. Is there something I need to do that I'm not aware of?

Xaeroxe:
Apparently that thing was "Don't include submodule updates in your commits accidentally"

cra0zy:
Since this is an API addition that is of bigger caliber, I can't merge it, but you get my thumbs of approval.

CC. @tomspilman @Jjagg

Xaeroxe:
Github was showing merge conflicts so I rebased, hence the force push.

MonoGame.Framework/Input/KeysHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Xna.Framework.Input
+{
+    internal static class KeysHelper
+    {
+        static Dictionary<int, Keys> _map;
```
Jjagg: You can do this with a HashSet instead of a Dictionary.

MonoGame.Framework/InputKeyEventArgs.cs
```diff
@@ -0,0 +1,19 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using Microsoft.Xna.Framework.Input;
+
+namespace Microsoft.Xna.Framework
+{
+    public class InputKeyEventArgs : EventArgs
```
Jjagg: You added two event classes (probably a conflict got incorrectly resolved).

MonoGame.Framework/KeyEventArgs.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Xna.Framework.Input;
+
+namespace Microsoft.Xna.Framework
+{
+    /// <summary>
+    /// This class is used for the game window's KeyDown and KeyUp event as EventArgs.
+    /// </summary>
+    public class KeyInputEventArgs : EventArgs
+    {
+        public KeyInputEventArgs(Keys key)
```
Jjagg: Can you add documentation to the public members of the class you're keeping?

MonoGame.Framework/Platform/SDL/SDLGameWindow.cs
```diff
@@ -302,7 +302,20 @@ public void ClientResize(int width, int height)
 
         public void CallTextInput(char c, Keys key = Keys.None)
         {
-            OnTextInput(this, new TextInputEventArgs(c, key));
+            if(IsTextInputHandled)
```
Jjagg: You don't need these checks. `EventHelpers.Raise` does the null check.
I suggest we make the added EventArgs structs so there's no allocation here.

MonoGame.Framework/GameWindow.cs
```diff
@@ -146,13 +159,21 @@ protected void OnScreenDeviceNameChanged ()
...
-		}
+		protected void OnTextInput(object sender, TextInputEventArgs e)
+		{
+            EventHelpers.Raise(this, TextInput, e);
+		}
+	    protected void OnKeyDown(object sender, InputKeyEventArgs e)
+	    {
+	        EventHelpers.Raise(this, KeyDown, e);
+	    }
+	    protected void OnKeyUp(object sender, InputKeyEventArgs e)
```
Jjagg: Maybe just expose these as internal, so we don't need the additional helper methods in the GameWindow subclasses that just call this.

MonoGame.Framework/Platform/Windows/WinFormsGameWindow.cs
```diff
@@ -343,6 +343,38 @@ private void OnKeyPress(object sender, KeyPressEventArgs e)
...
+                OnKeyDown(this, new InputKeyEventArgs(key));
+        }
+
+        internal void CallOnKeyUp(Microsoft.Xna.Framework.Input.Keys key)
+        {
+            if(IsKeyUpHandled)
+                OnKeyUp(this, new InputKeyEventArgs(key));
+        }
+
+        internal static Microsoft.Xna.Framework.Input.Keys KeyEventTranslate(KeyEventArgs e)
```
Jjagg: This isn't used.

MonoGame.Framework/Platform/SDL/SDLGameWindow.cs
cra0zy: Its not a null check, its there to avoid garbage.

Xaeroxe: InputKeyEventArgs inherits from EventArgs, which is a requirement for the event machinery. So, I can't convert this to a structure.

Jjagg: It's not a requirement. Usually event arguments inherit from it in case consumers don't care about the argument.

Xaeroxe: Here's what happens when I remove that inheritance

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0311	The type 'Microsoft.Xna.Framework.InputKeyEventArgs' cannot be used as type parameter 'TEventArgs' in the generic type or method 'EventHelpers.Raise<TEventArgs>(object, EventHandler<TEventArgs>, TEventArgs)'. There is no implicit reference conversion from 'Microsoft.Xna.Framework.InputKeyEventArgs' to 'System.EventArgs'.	MonoGame.Framework.Windows	C:\Users\JKiesel.CISDEV\MonoGame\MonoGame.Framework\GameWindow.cs	172	Active
```

Xaeroxe: Ok, I think I found a way to get the best of both worlds. I'm bypassing the EventHelper and invoking the event directly with my own null check.

Jjagg: Ah, right. The EventHelper that's used has a generic type constraint so type arguments must inherit from EventArgs. We should get rid of that so we can use structs IMO.

Xaeroxe: @Jjagg just so we have clear expectations, is that a change you want me to make here?

Jjagg: Yes, please do :)

Xaeroxe:
@Jjagg Ready for review!

Jjagg:
This look good! I didn't test it functionally, but if someone can confirm it works for each platform I'm happy to merge it! Awesome work @Xaeroxe :) 

@cra0zy We'll need to add the new files to the .csproj files as well.

Xaeroxe:
I can promise this works on Windows DirectX, as I ported my game to use my fork of MonoGame here. I'll test Linux, Windows GL, and Windows Universal this evening. I cannot test OSX due to not having hardware for it.

Jjagg:
If DesktopGL works on 1 OS, we can assume it works on others. Else there would be a terrible terrible bug in SDL :p

Jjagg:
cc @tomspilman 
You'd maybe want to sign off on the API.

Xaeroxe:
Behavior validated! I've tested this on Desktop GL and Windows Universal. It worked in both places.

Jjagg:
@Xaeroxe I updated this because we changed some things in how the directories are organized to migrate to .NET Core. I also added the new files to the relevant projects. They'll still need to be added to the MacOS project (#6809).

cc @tomspilman Could you chip in if you're good with merging these changes.
tl;dr version is it adds `KeyUp` and `KeyDown` events for desktop platforms.

PR #6766, opened 2019-06-23
Test Default behaviour

nkast:
This PR tests how MonoGame behaves under the default settings (aka = the next official MG release)
https://github.com/MonoGame/MonoGame/pull/6621#discussion_r251419377

### DIFF
--- a/Test/Framework/Graphics/GraphicsDeviceTestFixtureBase.cs
+++ b/Test/Framework/Graphics/GraphicsDeviceTestFixtureBase.cs
@@ -51,11 +51,6 @@ namespace MonoGame.Tests.Graphics
         [SetUp]
         public virtual void SetUp()
         {
-#if !XNA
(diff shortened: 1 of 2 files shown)


nkast:
Mac test fails on: 
Draw_with_viewport_changing(Deferred)  
Draw_with_viewport_changing(Immediate) 

...with the message: '1 of 1 frames failed the similarity test.'

nkast:
FYI @Jjagg

Jjagg:
The test should not apply the half pixel offset manually and should leave the flag at false. That should get everything passing. Regardless of the value of the standard addressing flag or the platform, SpriteBatch will render without the DX9 style half-pixel offset which is consistent with XNA (because they fixed the offset for SpriteBatch). IMO that better shows the intent, because the flag should be set to false for XNA compatibility and our tests check for XNA compatibility.

Does that sound sensible to you?

PR #6770, opened 2019-06-24
[Pipeline Tool]: Fix assembly load API used for custom user references

scott-lin:
Assemblies are currently loaded using the wrong API. The first is the current API being used, and the latter is the correct one for file paths.

[Assembly.Load(string)](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.load?view=netframework-4.8#System_Reflection_Assembly_Load_System_String_): Loads an assembly given the long form of its name.
[Assembly.LoadFrom(string)](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=netframework-4.8#System_Reflection_Assembly_LoadFrom_System_String_): Loads an assembly given its file name or path.

### DIFF
--- a/Tools/Pipeline/Common/PipelineTypes.cs
+++ b/Tools/Pipeline/Common/PipelineTypes.cs
@@ -448,7 +448,7 @@ namespace MonoGame.Tools.Pipeline
             {
                 try
                 {
-                    var a = Assembly.Load(path);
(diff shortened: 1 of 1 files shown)


cra0zy:
Thanks @scott-lin ! Merging!

PR #6771, opened 2019-06-25
Include Template for  Visual Studio 2019

facilityweb:
A simple part of all templates, the first one is for Windows project.


cra0zy:
We plan on switching to using `dotnet templating` templates going forward: 
- https://github.com/dotnet/templating
- https://github.com/MonoGame/MonoGame/tree/develop/ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content

They can be used instead of traditional templates with VS/MonoDevelop (requres VS 2017 tho), and we get Rider support out of the box as well.

Jjagg:
@cra0zy How can dotnet templates be used from VS? I thought they still didn't have proper integration? But I'd be very happy if we could drop VS templates. (The only reason left to install MG would be the Pipeline tool, but that's not a discussion for here)

cra0zy:
> How can dotnet templates be used from VS? I thought they still didn't have proper integration?

Nope, they can be integrated, you just have to write an extension that references them, same as for MonoDevelop.

Rider picks up any dotnet templates installed in the global cache.

cra0zy:
PS. We talked about this a long time ago, MonoDevelop templates have already been switched:
- https://github.com/MonoGame/MonoGame/blob/develop/IDE/VisualStudioForMac/Properties/Manifest.addin.xml#L17

Installers/Windows/MonoGame.nsi
```diff
@@ -272,6 +288,7 @@ LangString VS2012Desc ${LANG_ENGLISH} "Install the project templates for Visual
 LangString VS2013Desc ${LANG_ENGLISH} "Install the project templates for Visual Studio 2013"
 LangString VS2015Desc ${LANG_ENGLISH} "Install the project templates for Visual Studio 2015"
 LangString VS2017Desc ${LANG_ENGLISH} "Install the project templates for Visual Studio 2017"
+LangString VS2017Desc ${LANG_ENGLISH} "Install the project templates for Visual Studio 2019"
```
evertonfragoso: Shouldn't it be `VS2019Desc` instead of `VS2017Desc`?

facilityweb: Done

harry-cpp:
#7076 fixes this.

amursky:
Yaay! 2019 templates in 2020! :)

PR #6775, opened 2019-06-25
Move MonoGame.Framework.Net to seperate directory

### LINKED ISSUE #4794: Move MonoGame.Framework.Net to its own folder

Jjagg:
Does what it says on the tin. I hope the build passes 🤞 

Fixes #4794.


cra0zy:
You need to fix file renaming XD

Jjagg:
Well, that didn't work. I think this might have to do with line endings.

Jjagg:
Yeah, I accidentally removed carriage returns. IMO that's not an issue and we would be better off enforcing Unix style line endings in a .gitattributes file.

cra0zy:
We can make a separate PR for converting line encodings, in this PR please don't do that, its gonna make a nightmare of git history.

Jjagg:
Oh, I didn't know it messes up history. Everything showed up as renames locally. But will fix tomorrow.

Jjagg:
Good to go!

Issue #6773, opened 2019-06-25
Replace nant scripts with CAKE scripts


harry-cpp:
CAKE is literally made with C# project compilation in mind, as such its much easier to call stuff like `dotnet restore` or `msbuild pack` without having to worry if you properly setuped the location of where the stuff is found on each platform.

CAKE scripts are just C# scripts with some predefined methods, any C# function can be called from them because of this. Its also much easier to read the build script logic as its just C# code.

https://cakebuild.net

CC. @tomspilman @Jjagg @dellis1972 @KonajuGames

Jjagg:
IMO the biggest benefit is that Cake can be installed as a .NET Core global tool: 
- global tools doc page: https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools
- Cake tool nuget: https://www.nuget.org/packages/Cake.Tool/

So it's easy to run the build scripts anywhere.
Installation is as simple as running `dotnet tool install -g Cake.Tool`.

cra0zy:
Actually CAKE is self contained, it will automatically download itself from the build scripts and place itself into tools folder. You literally just need to have C# runtime installed.

hazevich:
@Jjagg you don't even need `dotnet tool` just run it through powershell and it will install relevant packages itself

tomspilman:
I'm good with all this... NANT is ancient.

As long as the new script breaks up the build in a similar way to how we had it in NANT we can fix up the build server to use it instead.

I assume i will have to install a Cake binary on the build box to make this work or find a Cake Runner for TeamCity.

Jjagg:
Cake can be installed by running the bootstrap scripts that are included already or by installing it as a dotnet global tool:

Run the bootstrap scripts (downloads CAKE and runs the build script):
```
./build.sh # Unix
build.ps1 # Windows
```

dotnet global tool:
```
$ dotnet tool install -g Cake.Tool
$ dotnet cake
```

cra0zy:
> I assume i will have to install a Cake binary on the build box to make this work or find a Cake Runner for TeamCity.

No need, just make it run the bootstrap script.

cra0zy:
We do need build bot C# stuff updated tho for the new projects.

tomspilman:
@cra0zy  @Jjagg @mrhelmut 

So how should we proceed with this?

I don't want to break the automated build for weeks... but we need a way to test automated builds using Cake.

Any ideas?

cra0zy:
[...]

PR #6772, opened 2019-06-25
Setup DesktopGL project

cra0zy:
Notes:
- Restored the csproj ignore to .gitignore, we should force csproj files until move is done
- Added DesktopGL project
  - This targets both `.netstandard2.0` and `.net451`, and this means that the nuget will be referencable by older projects as well
- Project file is a nuget package file at the same time :)
- I'm cheating because this isn't connected to the build bot, and is not breaking any stuff as our project name was never .DesktopGL.csproj

How to build nuget:
```sh
dotnet restore MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
msbuild /t:Pack /p:Version=3.8.0.3 MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
```

CC. @Jjagg

PR #6776, opened 2019-06-25
Add CAKE build script

cra0zy:
Adds the build script as per #6773.

Just run either `build.sh` or `build.ps1` depending on the platform.

Stuff will slowly be added to the build script as they are moved from Protobuild.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,0 +1,77 @@
+#tool nuget:?package=NUnit.ConsoleRunner&version=3.4.0
+
+//////////////////////////////////////////////////////////////////////
+// ARGUMENTS
+//////////////////////////////////////////////////////////////////////
+
+var target = Argument("build-target", "Default");
+var version = Argument("build-version", "1.0.0.0");
+var configuration = Argument("build-configuration", "Release");
+
+//////////////////////////////////////////////////////////////////////
+// PREPARATION
+//////////////////////////////////////////////////////////////////////
+
+MSBuildSettings mspacksettings;
+DotNetCoreMSBuildSettings dnbuildsettings;
+DotNetCorePackSettings dnpacksettings;
+
+//////////////////////////////////////////////////////////////////////
+// TASKS
+//////////////////////////////////////////////////////////////////////
+
+Task("Prep")
+    .Does(() =>
+{
+    mspacksettings = new MSBuildSettings();
+    mspacksettings.Verbosity = Verbosity.Minimal;
+    mspacksettings.Configuration = configuration;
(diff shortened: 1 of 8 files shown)


Jjagg:
Could you use DotNetCorePack() instead of MSBuild() for packing? It's more consistent and will remove the MSBuild dependency.
https://cakebuild.net/api/Cake.Common.Tools.DotNetCore/DotNetCoreAliases/0F7A518E

jnoyola:
> Could you use DotNetCorePack() instead of MSBuild() for packing?

Just reading through. He mentioned the following in https://github.com/MonoGame/MonoGame/pull/6772#issuecomment-505391103, so I'm assuming it needs to be MSBuild().
> dotnet msbuild does not support standard .net framework compilation on non Windows platforms

Jjagg:
Ah, thanks. I missed that comment.

jnoyola:
Cake looks pretty powerful overall, but it looks like right now it'll be used here for restoring, building, and packing. Is this difficult to orchestrate cross-platform with just dotnet and msbuild, or do we expect to see the full benefit of Cake as the complexity increases in the future?

cra0zy:
Unit testing, template packing, installer building...

Jjagg:
The MSBuild call fails on Windows:
```
\MonoGame\MonoGame.Framework\MonoGame.Framework.DesktopGL.csproj(1,1): error MSB4041: 
  The default XML namespace of the project must be the MSBuild XML namespace.
  If the project is authored in the MSBuild 2003 format, please add 
  xmlns="http://schemas.microsoft.com/developer/msbuild/2003"  to the <Project> element.
  If the project has been authored in the old 1.0 or 1.2 format, please convert it to MSBuild 2003 format.
An error occurred when executing task 'BuildDesktopGL'.
```

build.cake
```diff
@@ -0,0 +1,77 @@
...
+    dnpacksettings.Configuration = configuration;
+});
+
+Task("BuildDesktopGL")
+    .IsDependentOn("Prep")
+    .Does(() =>
+{
+    DotNetCoreRestore("MonoGame.Framework.DesktopGL.sln");
+
+    if (IsRunningOnWindows())
```
Jjagg: Maybe a short comment to explain why this is necessary?
Also would be nice to make a function for this to make it less verbose:
```
void PackProject(string project)
{
    DotNetCoreRestore(project);

    if (IsRunningOnWindows())
        DotNetCorePack(project, dnpacksettings);
    else
        MSBuild(project, mspacksettings);
}
```

Jjagg:
Looks good, merging!

PR #6777, opened 2019-06-26
Add .gitattributes and normalize line endings

### LINKED ISSUE #4335: Adding A .gitattributes File
Should we add a .gitattributes file to the project?

https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings

Looks like it could help with line endings differences?

Jjagg:
Fixes #4335.


Jjagg:
@cra0zy Good to go!

cra0zy:
Yea, this totally messes up with git history, I am against this move, sorry.

cra0zy:
So if we are gonna do this in any point in time, I think the following set of stuff needs to be done:
- Rebase all commits to 1 commit
- Ensure that account signing of the commits is not your own but instead the mgbot account
- Run the following tool on the entire repo as well: https://github.com/dotnet/codeformatter
- Do it after Protobuild removal

PR #6778, opened 2019-06-26
Fix SpriteBatch test that manually applied half pixel offset

Jjagg:
Supersedes #6766.
This might need the reference image to be updated.

### DIFF
--- a/Test/Framework/Graphics/SpriteBatchTest.cs
+++ b/Test/Framework/Graphics/SpriteBatchTest.cs
@@ -456,13 +456,8 @@ namespace MonoGame.Tests.Graphics {
             _spriteBatch.End();
 
             // Test viewport/effect BasicEffect (Vertex & Pixel shader)
-#if DIRECTX
-            Matrix halfPixelOffset = Matrix.Identity;
-#else            
-            Matrix halfPixelOffset = Matrix.CreateTranslation(-0.5f, -0.5f, 0);
-#endif
 
-            _effect.Projection = halfPixelOffset * Matrix.CreateOrthographicOffCenter(0, gd.Viewport.Width, gd.Viewport.Height, 0, 0, 1);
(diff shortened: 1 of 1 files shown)


nkast:
In XNA you need to add a halfpixel to the Projection when you use a custom Effect with SpriteBatch.

http://www.shawnhargreaves.com/blog/spritebatch-billboards-in-a-3d-world.html

```
    basicEffect.Projection = Matrix.CreateTranslation(-0.5f, -0.5f, 0) * 
    Matrix.CreateOrthographicOffCenter(0, viewport.Width, viewport.Height, 0, 0, 1);
```

Jjagg:
Ohh, I completely misunderstood! Okay, that makes sense. Then the test is good as it is since we don't support the reverse half pixel offset in DX.

nkast:
Until now it would add halfpixel for (!DIRECTX), aka (XNA || GL).
What changed is that GL now behaves like DIRECTX and doesn't need the halfpixel.
Or does it? That really depend's on the value of UseStandardPixelAddressing.

nkast:
>> Or does it? That really depend's on the value of UseStandardPixelAddressing.

The [second commit](https://github.com/MonoGame/MonoGame/pull/6766/commits/f96dd80d245e9af0b9e704cb6b11cfe957ba5445) is generic and will work either way.

Jjagg:
> Until now it would add halfpixel for (!DIRECTX), aka (XNA || GL).

This is correct because we set `UseStandardPixelAddressing` to `false` for testing to make rendering with custom effects consistent with XNA, but DIRECTX does not yet have any way to inject code into user shaders to apply the reverse half pixel offset. So OpenGL platforms behave like XNA when `UseStandardPixelAddressing` is `false`, but our DirectX backend does not yet support this feature.

If you want to make explicit that this depends on `UseStandardPixelAddressing` the semantically most correct way would be your commit. I would merge a PR that only applies your second commit :) 
Sorry for the confusion!

nkast:
ok. I am on it.

PR #6780, opened 2019-06-26
Move UseStandardPixelAddressing

nkast:
- [x] Move UseStandardPixelAddressing to GraphicsDevice
- [x] Make UseStandardPixelAddressing non-static
- [x] Initialize UseStandardPixelAddressing from GraphicsDevice constructor and make readonly
- [x] Add PreferStandardPixelAddressing to GraphicsDeviceManager


nkast:
FYI @Jjagg 

First stage is ready and you can merge it as is if you like.

Next a decision need to made on how to set the value initially.
Instead of GraphicsDeviceManager.PrefferedUseStandardPixelAddressing, we can add it to GraphicsDeviceInformation. T o change the default value you would have to subscribe to PreparingDeviceSettings event.  e.g.

```
    graphics.PreparingDeviceSettings += (sender, e) => 
    {
        e.GraphicsDeviceInformation.UseStandardPixelAddressing = false;
    }
```

GraphicsDeviceInformation is usually for parameters that don't change, like the selected Adapter, Profile, etc.

Jjagg:
The flag doesn't fit in PresentationParameters. So we'd have to go with a ctor overload.

That said, I'm questioning if making the flag a GD property is the right thing to do...
I don't think there's a scenario where you want to have multiple graphics devices with a different setting for half pixel offset. Setting the flag statically is easier for users.

nkast:
If you are ok with just the first commit I can remove the second commit , and keep on discussing this.

Here's what I think, 
I would prefer the flag to be readonly and be set once during the constraction. This can only be done if the flag is an instance property. The reasons I want it readonly is to (1) not allow 3rd party libraries to change it and force their preference to the entire project. Libraries should only read it and adapt to the dev preferences. (2) being able to [pre-calculate](https://github.com/nkast/MonoGame/commit/03b3539aee23810b91206e7c143ea0e916e5e11d#diff-2586cbffd4a24e6c7439988a0b360cf3) the projection matrix, instead of calculating it every time just case it's the flag has been changed between calls. 

If it's not readonly then you have to treat it as any other state. just as you have to set the BlendState, etc before a draw call, you have to set the flag just in case something else changed it.
If the flag is static and the dev uses multiple GDs, possibly on different threads, then this wont work, even if the dev didn't indent to use mixted configs a 3rd library might do change it.

Jjagg:
Yeah, I do agree it would be nice to have it readonly and GraphicsDevice is the place that makes most sense to put it. I guess my issue is more with his XNA designed GraphicsDevice initialization. But that's a bad reason to not implement this flag the right way. So let's do it like you suggest :)

nkast:
That's the only thing I don't like about adding it, the GD constructor gets too long. Maybe we need to copy how GraphicsDeviceInformation works and have a public ctor like 
'GraphicsDevice(DeviceParameters deviceParameters , PresentationParameters presentationParameters)'
where we put all initial values (adapter, graphicsProfile, UseStandardPixelAddressing) in a single struct or class. Then we can add, remove or mark obsolete any initial parameters with like impact to the users that don't use those paramereters. 

Anayways, I can look into it in the next couple of days.

nkast:
On a second though I didn't put the flag on GraphicsDeviceInformation because that object is linked to the adapter/adapters. Instead I put it on the GraphicsDeviceManager as PreferStandardPixelAddressing.

Jjagg:
Could you modify the `GraphicsDeviceManager` doc for the flag to specify that the value is passed on to a `GraphicsDevice` when it is created and that the value in `GraphicsDeviceManager` does not control the effect directly?

nkast:
@Jjagg I modified the documentation to reflect the new changes.

One more thing to consider... 
I unusually try to have boolean variables default to 'false'. That's because boolean in c# are initialized to 'false' by default, thou leads to less code and I don't have to worry about ensuring that the variable is always initialized when refactoring. 
In this case the 'PreferStandardPixelAddressing' can be negated to 'PreferNonStandardPixelAddressing' (or 'PreferLegacyPixelAddressing' or 'PreferHalfPixelAddressing').

Jjagg:
That's a good point. `PreferStandardPixelAddressing` was chosen because it was opt-in to get non-XNA compatible behavior, but in the process the default changed to be not XNA compatible. I think it makes sense to rename the property. 

I think the best option is `UseHalfPixelOffset`/`PreferHalfPixelOffset`. It says exactly what MG does, add half a pixel offset.

Apostolique:
I vote for `UseHalfPixelOffset`. Prefer to me doesn't sound as authoritative and leads me to believe there are cases where the property wouldn't get honored.

Jjagg:
I meant `PreferHalfPixelOffset` for the property in `GraphicsDeviceManager` and `UseHalfPixelOffset` for the corresponding property in `GraphicsDevice`.

Jjagg:
The setting the user provide will in fact not be honored right now by the DX implementation because we don't support half pixel offset there yet.

Jjagg:
This is great @nkast, thanks!

PR #6786, opened 2019-06-27
Make IWindowInfo internal

Jjagg:


Jjagg:
Oops, looks like this is exposed on iOS so users can make their context current easily. I'll close this for now.

Issue #5959, opened 2017-09-25
Incorrect resolution values being returned on my new 4K monitor


AgisisTheCoder:
OK, so this is a fun mysterious one. I bought a new 4K monitor and am having very weird results as far as what XNA says my resolution is.

**In Windowed Mode:**
Windows OS says my display resolution is 3840 x 2160
Our game, however, says my OS resolution is 2560x1440!
The game says the client window is 2560x1377
A screenshot taken of the windowed mode game comes out at 3840x2160

**I am using these two values to get the information from in the code:**
game.Window.ClientBounds.Height
GraphicsAdapter.DefaultAdapter.CurrentDisplayMode.Height

I've shared a screenshot containing all of the windows in question and the resolutions. The interesting thing is my co-developer has a 4K monitor as well, same resolution, but does not have this issue in windowed mode. It all comes out correct. To top it off, if you go into full screen mode, the values are all exactly right.

![screenresolutionhell](https://user-images.githubusercontent.com/31192691/30791405-c9f04940-a177-11e7-98d7-d3c6ba10df53.jpg)

AgisisTheCoder:
Even this is giving me the wrong resolution. 
System.Windows.Forms.Screen.PrimaryScreen.Bounds;

Something wacky is afoot. I don't think it's monogame so not sure if this is even appropriate now.

tomspilman:
> Something wacky is afoot.

By default MG and XNA titles are not "dpi aware".  You have to opt in by a Win32 call and/or the app manifest:

https://msdn.microsoft.com/en-us/library/windows/desktop/aa374191(v=vs.85).aspx


@KonajuGames @dellis1972 @Jjagg @cra0zy - Should we just force MG to always be DPI aware?  4k systems are becoming more common and likely games just need to deal with it.

Jjagg:
IMO the manifest in the templates is enough.

mrhelmut:
The manifest should be enough, but I'm going to give this some test anyway.

I've been using the manifest in the template to work with DPI awareness and it worked quite well so far. But recently we started to received reports of DPI awareness being broken on Windows 10 while we didn't update the game. I don't know if it's a problem specific to Windows 10 Creator Update (or a more recent update) or a problem on the user side. I need to investigate.

AgisisTheCoder:
[...]

PR #6790, opened 2019-06-27
Update WindowsDX template

Jjagg:
I added dpiAwareness to the manifest as per #5959. But I think we better set DPI awareness with a Win32 call so we can get rid of the manifest. Though that's not something for this PR.


ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Application.WindowsDX.CSharp/MGNamespace.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
jnoyola: Didn't you say in the forums that Windows projects should target .NET Framework for smaller binaries?

cra0zy: ILLink.Tasks nuget solves that problem quite nicely.

jnoyola: Ooo that sounds useful! I'm assuming that still needs to be added here unless I missed it.

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Application.WindowsDX.CSharp/MGNamespace.csproj
```diff
@@ -0,0 +1,26 @@
...
+  <ItemGroup>
+    <MonoGameContentReference Include="Content\Content.mgcb" Visible="false" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MonoGame.Content.Builder" Version="3.7.0.9" />
+    <PackageReference Include="MonoGame.Framework.WindowsDX" Version="3.7.*" />
+    <PackageReference Include="NSubsys" Version="1.0.0">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
```
jnoyola: Minor, but I'm not sure if IncludeAssets and PrivateAssets are necessary here. I've used NSubsys without them and it removed the console without copying anything to the output directories.

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Application.WindowsDX.CSharp/MGNamespace.csproj
Jjagg: Oops, this should target .NET 4.5 because that's what WindowsDX targets (because WinForms isn't available in stable .NET Core). And also the binary size will still be smaller than a self-contained build with .NET Core. But ILLink is really nice.

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Application.WindowsDX.CSharp/MGNamespace.csproj
Jjagg: Ah, I'll try that. Thanks for the review @jnoyola :)

PR #6792, opened 2019-06-27
FUNDING.yml file with PayPal link

CartBlanche:
Step 1 of hopefully getting some more visible funding.

I'll update this to a proper subscription button. It may end up on the home page.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,13 @@
+# These are supported funding model platforms
+
+# github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
+# patreon: # Replace with a single Patreon username
+# open_collective: # Replace with a single Open Collective username
+# ko_fi: # Replace with a single Ko-fi username
+# tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+# community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
+# liberapay: # Replace with a single Liberapay username
+# issuehunt: # Replace with a single IssueHunt username
+# otechie: # Replace with a single Otechie username
+# custom: [![paypal](https://www.paypalobjects.com/en_US/i/btn/btn_donateCC_LG.gif)](https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=XHSP3K7VTENNW)
+custom: https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=XHSP3K7VTENNW


tomspilman:
@CartBlanche - So is this how the GitHub thing works?  You have to add a .yml file?

CartBlanche:
@tomspilman Yes adding a .yml is the first step. Then we need to apply as "sponsor-able" devs, which currently has a waiting list. Then once the sponsored devs are added, we can add those github names to the .yml file. So right now it is just the "donate" link, until the above criteria is met.

cra0zy:
I guess I should probably sign up for the thingy in case it can be useful.

Jjagg:
Sounds good, merging!

Jjagg:
If anyone donates via the PayPal link, where does the money go @CartBlanche? Is there a MG account of sorts?

tomspilman:
Cool...

image

RUSshy:
Some transparency would be nice indeed, where the money goes? it looks like it goes into an individual person, how is it handled?

Also looks like the link doesn't match the recommendation in the comment, there should be an icon ?

rejurime:
Talking about this issue, I would like to (re)ask (http://community.monogame.net/t/patreon-test/11385) if you have the idea of using the [MonoGame Patreon](https://www.patreon.com/MonoGame/). Several who have it offer simple rewards. Github sponsor supports Patreon y the yml file.

CartBlanche:
@Jjagg There is currently no separate MG account. All donations from the MonoGame.net/donate page and this link point to my PayPal account. It has been that way since the year we first started the donations. It was the simplest way to set it up at the time. Whenever I get notification about a donation I forward this onto Tom, so he can update the Donation page with the donator's details. I don't use the PayPal account other than for the odd E-Bay purchase (which I transfer my own money into to pay off) so all the money in the account is currently MG's. Only the original team members (4 of us) know the MG account balance. Our major outgoing right now is paying GitHub payment every month. Every so often we spend money on upgrading server/build hardware. Should a platform/console require us to buy dev kits, then the money we have will go towards that, if necessary.

Our choices going forward are completing the Patreon page (needs dedicated time and a separate MG bank account and possible incorporation as an entity) or getting a few of us Sponsored. Github dev Sponsors has a waiting list.

Jjagg:
Maybe the balance and spending should be public. I think funding transparency goes a long way to incentivising people to donate because they know what's needed.

CartBlanche:
@Jjagg I think this is a good idea. I'll chat to @tomspilman and others about it and we'll see what we can do to move that side forward.

Jjagg:
There's a pretty cool website for this: https://opencollective.com
See for example ImageSharp: https://opencollective.com/imagesharp
However, I think it's annoying that people need to create an account to donate and the platform takes 5% of donations (same as the minimum amount on Patreon).

Apostolique:
I don't know if the service is limited to streamers, but Streamlabs takes 0% fees. https://streamlabs.com/

The money can be deposited in a Paypal account. Streamlabs offers way to create leaderboards or show other information using their API: https://dev.streamlabs.com/

PR #6793, opened 2019-06-28
Fix MSBuild location for build bot in NANT script

Jjagg:
Build bot should use VS2019 MSBuild. I assumed the default location, hope that's correct. @tomspilman 
The new CAKE script uses [vswhere](https://github.com/microsoft/vswhere) to find msbuild, so that should will also work on user machines if they have VS2019 in different locations.

### DIFF
--- a/default.build
+++ b/default.build
@@ -4,8 +4,7 @@
   <description>The MonoGame automated build script.</description>
 
   <property name="os" value="${operating-system::get-platform(environment::get-operating-system())}" />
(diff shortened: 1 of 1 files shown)


Jjagg:
Well, this didn't work...
Locally I fixed the UWP build error just by building the UWP solution from VS. I got the same error as the build bot locally when trying to build from the developer command prompt.

Jjagg:
Closing for now, I'll investigate the build bot issue.

PR #6798, opened 2019-06-28
Run MakeNSIS from Cake for the Windows installer

Jjagg:
Cake script to build the installer for Windows.


Jjagg:
Right now the installer still tries to use the paths for the old projects, which the CAKE script does not build. Once we fully switch to the new projects, the paths in the installer need to be updated.

build.cake
```diff
@@ -5,14 +5,22 @@
 // ARGUMENTS
 //////////////////////////////////////////////////////////////////////
 
+const string DefaultVersion = "1.0.0.0";
```
cra0zy: Why the random const?

build.cake
```diff
@@ -5,14 +5,22 @@
...
+
 var target = Argument("build-target", "Default");
-var version = Argument("build-version", "1.0.0.0");
+var version = Argument("build-version", DefaultVersion);
 var configuration = Argument("build-configuration", "Release");
 
+if (version == DefaultVersion)
+    Warning($"No version specified. Using default version '{DefaultVersion}'.");
```
cra0zy: Redundant.

build.cake
```diff
@@ -5,14 +5,22 @@
...
+
 //////////////////////////////////////////////////////////////////////
 // PREPARATION
 //////////////////////////////////////////////////////////////////////
 
+var majorVersion = "3.0";
+var buildNumber = EnvironmentVariable("BUILD_NUMBER");
```
cra0zy: Add a way to specify these using arguments.

build.cake
Jjagg: Because it's the default and I compare to it below so to prevent typos.

build.cake
Jjagg: Why?

build.cake
Jjagg: I just copied how they were set in the old build script.
I think we should wait with fixing this until we set up proper versioning (I want to set up NerdBank.GitVersioning).

build.cake
cra0zy: You can achive the same with the argument.

build.cake
cra0zy: Because if you don't specify `X`, either one of 2 things will happen, it will fail, or it will use the default value.

I find it rather redundant to say "using the default version".

build.cake
```diff
@@ -137,16 +140,84 @@ Task("BuildContentPipeline")
...
+{
+});
+
+Task("PackMac")
+    .IsDependentOn("BuildAll")
+    .Does(() =>
+{
+});
+
+Task("PackInstallers")
```
cra0zy: Move this bellow task targets.

build.cake
```diff
@@ -137,16 +140,84 @@ Task("BuildContentPipeline")
...
-// TASK TARGETS
-//////////////////////////////////////////////////////////////////////
+Task("BuildTools")
+    .IsDependentOn("Prep")
+    .Does(() =>
+{
+});
 
-Task("Default")
+Task("BuildAll")
```
cra0zy: Move this bellow Task Targets.

PR #6800, opened 2019-06-29
CAKE General Improvements

cra0zy:
Stuff done:
- Fixed Cake plugins still using tools folder
- Fixed Xamarin.Android warning regarding Content tag
- Build fails on Windows if MSBuild is not found
- Xamarin Android is built from Linux if its installed
- Simplified future msbuild path finding by using Criteria and a bit of magic! (we don't need a variable for each platform now)

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.AndroidCore.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.AndroidCore.csproj
@@ -72,7 +72,7 @@ This package provides you with MonoGame Framework that works on Android.</Descri
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="MonoGame.Framework.AndroidCore.targets" PackagePath="build" />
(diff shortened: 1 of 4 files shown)


MonoGame.Framework/MonoGame.Framework.AndroidCore.csproj
```diff
@@ -72,7 +72,7 @@ This package provides you with MonoGame Framework that works on Android.</Descri
   </ItemGroup>
 
   <ItemGroup>
-    <Content Include="MonoGame.Framework.AndroidCore.targets" PackagePath="build" />
```
cra0zy: I need an extra modification here, once done with my current changes I'll add it.

Jjagg:
> we don't need a variable for each platform now

I did this in case you have e.g. a VS2019 install with Xamarin workload and VS2017 with UWP workload.
Criteria could initialize the variable, then just return if it's `null` or not. But if you prefer keeping it like this, that's fine with me too.

cra0zy:
I prefer it like this tbh.

PR #6809, opened 2019-07-11
Add macOS sln and csproj

Jjagg:
This adds a csproj for the macOS project. I did not yet test this because I don't own a Mac.
The project needs to be explicitly restored and built with MSBuild.
I did not yet add the build to the CAKE script.


Jjagg:
Closing this because we're dropping the Mac-specific platform. See #6811 for details.
DesktopGL will get a video player so it can fully replace the macos platform and we don't have to maintain this platform anymore :raised_hands:

PR #6810, opened 2019-07-11
Add iOS csproj

Jjagg:
I did not yet add a solution file and did not add building iOSCore to the CAKE script. Let's make sure it builds first (which I did not test yet).


cra0zy:
I'm gonna merge this as for any changes I want to make, I'll have to do it on top of this.

PR #6815, opened 2019-07-16
Use Resume when Play is called on Paused SFX instance

### LINKED ISSUE #6814: OpenALSoundController doesn't clear available sources that were paused during their lifetime
It seems that any looping SoundEffectInstances that have been paused at any point, after being stopped, will never be cleared from the list availableSourcesCollection in OpenALSoundController.cs , causing a InstancePlayLimitException after some time.  I managed to reduce it down to a test case, if you comment out the pausing it will no longer crash.  This means we are unable to pause sound effect instances. 

```
public void AudioLeakTest()
        {
            if (loop1 != null)
            {
                loop1.Stop(true);
                loop1.Dispose();
                loop1 = null;
            }

            if (loop2 != null)
            {
                loop2.Stop(true);
                loop2.Dispose();
                loop2 = null;
            }

            if (stop)
            {
                return;
            }

            loop1 = Audio.PlaySound("sfx_boulder_launch_00", true);
            loop1.Pause();
            loop1.Play();

            loop2 = Audio.PlaySound("sfx_boulder_launch_00", true);
            loop2.Pause();
            loop2.Play();

            Utils.Delay(0.5f, () =>
            {

                if (loop1 != null)
                {
                    loop1.Stop(true);
                    loop1.Dispose();
                    loop1 = null;
                }

                if (loop2 != null)
                {
                    loop2.Stop(true);
                    loop2.Dispose();
                    loop2 = null;
                }

                loop1 = Audio.PlaySound("sfx_boulder_launch_00", true);
                loop1.Pause();
                loop1.Play();

                loop2 = Audio.PlaySound("sfx_boulder_launch_00", true);
                loop2.Pause();
                loop2.Play();

                Utils.Delay(0.5f, () =>
                {
                    AudioLeakTest();
                });
            });

        }
```

- MonoGame 3.7

- Windows, Android, iOS

- DesktopGL, Android, iOS

Jjagg:
Fixes #6814.

cc @Owlzy @tdeeb

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffectInstance.cs
+++ b/MonoGame.Framework/Audio/SoundEffectInstance.cs
@@ -146,6 +146,12 @@ namespace Microsoft.Xna.Framework.Audio
             if (State == SoundState.Playing)
                 return;
 
+            if (State == SoundState.Paused)
+            {
+                Resume();
+                return;
+            }
+
             // We don't need to check if we're at the instance play limit
             // if we're resuming from a paused state.
             if (State != SoundState.Paused)


tdeeb:
@Jjagg This changes the behavior. Previously it would play the sound from the start, but now it resumes. WindowsDX projects also play sounds from the start.

In my opinion, we should stop the sound first if it's paused or reuse the source it already has reserved.

Jjagg:
Ah, I wasn't aware of that. That's not what the docs indicate is the intent though, so I'm not sure what we should do rn. Maybe check XNA behavior? To me resuming makes more sense.

Jjagg:
[XNA docs](https://docs.microsoft.com/en-us/previous-versions/windows/xna/dd282361%28v%3dxnagamestudio.41%29)clearly state the sound continues at the paused position so I think we should make this change.

tdeeb:
Sounds good! WindowsDX (and other platforms?) will need to be changed as well down the road.

Jjagg:
This changes behaviour in the shared source file, so it's already good for all platforms with this PR :)

Jjagg:
@cra0zy this is good to go.

PR #6817, opened 2019-07-17
Mark Color.TransparentBlack Obsolete

### LINKED ISSUE #6805: Color.TransparentBlack
XNA 3.0 had two transparent values,  TransparentBlack & TransparentWhite. When XNA 4.0 switch to use premultiplied colors by default, they drop both properties and replaced them with Color.Transparent. 

MG still defines Color.TransparentBlack which has the same value as Color.Transparent. 
I propose to mark TransparentBlack as 'Obsolete' for now and then remove it on a later version.

- MonoGame 3.7,.1 3.8 (dev)

- Windows

- WindowsDX, WindowsUWP, Android

nkast:
Closes #6805


Jjagg:
Thanks @nkast!

PR #6824, opened 2019-07-22
Upgrade Tools target and add SDK-style test projects

Jjagg:
This makes MonoGame.Framework.Content.Pipeline support the .NET Core runtime so MGCB and 2MGFX can target .NET Core 2.1 and can be distributed as a dotnet global tool.
Depends on MonoGame/MonoGame.Dependencies#131.

Due to a string of dependencies this also does the following:
- Fixes managed dependencies and native libraries so they can be used with the .NET Core runtime (which does not support dll.config files)
- Makes the Protobuild generated tools projects and content pipeline target .NET Framework 4.7.2
- Adds a reference to the .NET Framework reference assemblies for those projects on Linux and Mac (Mono does not yet support 4.7.2).
- Upgrades Protobuild (never too late, right?)
- References NUnitLite as a nuget package in the test projects.
- Adds SDK-style projects for Tests (these are WIP, but we might as well merge them in the current state)
- Makes non-Protobuild Content.Pipeline project target netstandard2.0
- Some misc other fixes for .NET Core


Jjagg:
Seems like the build fails because it's still using the old dependencies.
@tomspilman I think the build bot didn't update the dependencies submodule.

Jjagg:
@tomspilman The build bot is using the wrong version of MSBuild. It's using the old one at `C:\Program Files (x86)\MSBuild\12.0\bin`. Can you uninstall that so it uses the new one?

Jjagg:
> The build bot is using the wrong version of MSBuild. It's using the old one at C:\Program Files (x86)\MSBuild\12.0\bin. Can you uninstall that so it uses the new one?

^^^ @tomspilman ?

Jjagg:
@tomspilman?

Jjagg:
I don't know what's up, UWP is building succesfully locally...

Jjagg:
Seems like an issue with freetype native lib :/ 

```
Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at SharpFont.FT.FT_Render_Glyph(IntPtr slot, RenderMode render_mode)
   at SharpFont.GlyphSlot.RenderGlyph(RenderMode mode)
   at Microsoft.Xna.Framework.Content.Pipeline.Graphics.SharpFontImporter.ImportGlyph(Char character, Face face) in C:\Users\jesse\projects\MonoGame\MonoGame.Framework.Content.Pipeline\Graphics\Font\SharpFontImporter.cs:line 94
```

Jjagg:
Only 4 failing tests left. Seems the issue is with Android and iOS compression. Might be that something got messed up in the PVRTexLib upgrade.

Jjagg:
Oops, need to update deps first, don't merge yet!

Jjagg:
This is ready for review! Don't know why tests don't show up here, but they're passing :) 
@cra0zy

Jjagg:
Hold up, forgot to do 2MGFX.

Build/Projects/PipelineReferences.definition
```diff
@@ -2,6 +2,7 @@
 <ExternalProject Name="PipelineReferences">
 
   <Platform Type="Linux">
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0-preview.2" />
```
cra0zy: Only needed if targeting .NET 4.7.0 or lower.

Jjagg: We target 4.7.2 now, so you can't target lower. Mono does not yet support 4.7.2, that's why the reference assemblies are required.

cra0zy: > We target 4.7.2 now, so you can't target lower.

Yes, with 4.7.2 the package is redundant.

> Mono does not yet support 4.7.2, that's why the reference assemblies are required.

It does support it, it supported it for a long time.

Jjagg: I'm on Ubuntu 18.04, which does not have a new enough version of mono in it's package repo. That means you can't build without the reference assemblies.

MonoGame.Framework.Content.Pipeline/ExternalTool.cs
```diff
@@ -168,8 +168,7 @@ private static void EnsureExecutable(string path)
 
             try
             {
-
-                var p = Process.Start("chmod", "u+x '" + path + "'");
+                var p = Process.Start("chmod", "u+x " + path);
```
cra0zy: If path contains spaces this will fail.

Jjagg: Hmm, I don't remember why I did that. Will fix!

Jjagg: Fixed!

Tools/2MGFX/2MGFX.csproj
```diff
@@ -0,0 +1,201 @@
...
+    <Reference Include="..\..\ThirdParty\Dependencies\SharpDX\SharpDX.D3DCompiler.dll" />
+    <Reference Include="..\..\ThirdParty\Dependencies\CppNet\CppNet.dll" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="..\..\MonoGame.Framework.Content.Pipeline\Processors\MGFX\CommandLineParser.cs" />
```
cra0zy: This should reference MG / MG.Content.Pipeline, console stuff are special so...

Test/MonoGame.Tests.WindowsDX.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <PackageReference Include="NUnit3TestAdapter" Version="3.11.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
+    <PackageReference Include="System.Drawing.Common" Version="4.5.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\MonoGame.Framework\MonoGame.Framework.WindowsDX.csproj" />
+    <ProjectReference Include="..\MonoGame.Framework.Content.Pipeline\MonoGame.Framework.Content.Pipeline.csproj" />
+  </ItemGroup>
+
```
cra0zy: bad line encoding.

Issue #6825, opened 2019-07-23
PipelineController catches only NotSupportedException upon initialization of encoding


tchanturia:
Currently `PipelineController` upon `DoBuild` tries to set `encoding` within a `try/catch` clause. The clause only catches `NotSupportedException` and initializes `encoding` with `UTF8` in case of the exception.
https://github.com/MonoGame/MonoGame/blob/56694bbb9f0b92a5d85014982d9e423b63267049/Tools/Pipeline/Common/PipelineController.cs#L501-L506

According to msdn `Encoding.GetEncoding` may also throw `ArgumentException` which means that provided `codepage` is not supported too. Not catching this exception leads to the pipeline tool sometimes being unresponsive on build/rebuild/clean actions:

![pipelinebug](https://user-images.githubusercontent.com/35450004/61712255-35a04780-ad56-11e9-8669-6a1afb7a6261.gif)

#### What version of MonoGame does the bug occur on:
- latest develop build

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- WindowsDX

Jjagg:
Fixed in #2826.

PR #6826, opened 2019-07-23
Fix for 6825. Catch ArgumentException when trying to get encoding upon PipelineController.DoBuild

hazevich:
Related to issue #6825

### DIFF
--- a/Tools/Pipeline/Common/PipelineController.cs
+++ b/Tools/Pipeline/Common/PipelineController.cs
@@ -501,7 +501,7 @@ namespace MonoGame.Tools.Pipeline
             Encoding encoding;
             try {
                 encoding = Encoding.GetEncoding(CultureInfo.CurrentCulture.TextInfo.OEMCodePage);
-            } catch (NotSupportedException) {
+            } catch (Exception ex) when (ex is NotSupportedException || ex is ArgumentException) {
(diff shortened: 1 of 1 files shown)


Tools/Pipeline/Common/PipelineController.cs
```diff
@@ -501,7 +501,7 @@ private void DoBuild(string commands)
             Encoding encoding;
             try {
                 encoding = Encoding.GetEncoding(CultureInfo.CurrentCulture.TextInfo.OEMCodePage);
-            } catch (NotSupportedException) {
+            } catch (Exception ex) when (ex is NotSupportedException || ex is ArgumentException) {
```
tdeeb: MonoGame is using an older version of C# (C# 5?), so this will have to handle both exceptions in separate catch blocks.

tchanturia: Yeah, I asked @Jjagg if there are any plans to update the compiler, also given that there is movement towards .net standard. But I am also fine going the old ways.

hazevich: I removed the usage of the pattern matching.
Also I've found out that I was using two different github accounts. My apologies for the confusion.

Jjagg:
Thanks @hazevich!

Issue #6820, opened 2019-07-18
Joystick.GetState generates lots of garbage


tdeeb:
[`Joystick.GetState`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Input/Joystick.SDL.cs#L78) allocates new arrays for each `JoystickState`, which can lead to lots of garbage over time considering input is usually obtained multiple times a frame for controllers. I can think of a few options for improving this:
* Add a cached default state that is simply returned when we need to return a [default JoystickState](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Input/Joystick.SDL.cs#L80). This avoids allocating multiple zero-length arrays.
* Add some type of garbage-less overload that takes in user-defined arrays to fill the information with. The user can get the `JoystickCapabilities` first to see how many axes, buttons, and hats the joystick has - the existing code already does this, as a matter of fact. One thing to consider would be how to cleanly present this API.

#### What version of MonoGame does the bug occur on:
- Develop

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

nkast:
>> Add a cached default state that is simply returned when we need to return a default JoystickState. This avoids allocating multiple zero-length arrays.

We have done the same thing for TouchPanelState #2902

>> Add some type of garbage-less overload that takes in user-defined arrays 
Here's [another idea](https://github.com/MonoGame/MonoGame/issues/2876#issuecomment-51115871). 
As a matter of fact I had implemented a garbage free touch collection for use in Windows phone and I am still using the same code on Android. It uses a fixed array for up to 4 touch positions. For more that 4 it fallback to an array.

tdeeb:
More garbage/performance improvements:

- Reduce a dictionary lookup in [`PlatformGetCapabilities`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Input/Joystick.SDL.cs#L55) via `TryGetValue`.
- Replace [`HasFlag` checks](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Input/Joystick.SDL.cs#L107) with a direct bitwise check to avoid boxing and various security checks. For example: `(hatstate & Sdl.Joystick.Hat.Up) != 0`

tdeeb:
[...]

PR #6829, opened 2019-07-27
Performance and garbage optimizations to Joystick.SDL

tdeeb:
Relevant issue: https://github.com/MonoGame/MonoGame/issues/6820

Summary of changes in the initial commit:
- Added a default JoystickState that's returned if the joystick isn't found to bypass allocating several zero-length arrays
- Replaced HasFlag with bitwise operators to avoid boxing
- Reduced dictionary lookups if the joystick exists by replacing ContainsKey followed by a dictionary indexer with TryGetValue

I'd like to figure out a way to let users bypass allocating the arrays involved in creating each `JoystickState`, as they can generate a lot of garbage over time.

### DIFF
--- a/MonoGame.Framework/Platform/Input/Joystick.SDL.cs
+++ b/MonoGame.Framework/Platform/Input/Joystick.SDL.cs
@@ -9,6 +9,14 @@ namespace Microsoft.Xna.Framework.Input
 {
     static partial class Joystick
     {
+        private static JoystickState _defaultJoystickState = new JoystickState
+        {
+            IsConnected = false,
+            Axes = new int[0],
+            Buttons = new ButtonState[0],
+            Hats = new JoystickHat[0]
+        };
+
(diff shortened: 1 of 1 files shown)


Jjagg:
You can use `Array<T>.Empty` instead of `new T[0]`. No need to cache the default instance. Otherwise, this looks good :)

tdeeb:
I'm still using Protobuild to generate the project files so I'm seeing the project as .NET Framework 4.5. Should I be running the CAKE script?

Jjagg:
Oh wow, I thought it existed way back. Nevermind then. Merging! Thanks @tdeeb

tdeeb:
@Jjagg I appreciate the merge; this is definitely an improvement over before. I should have mentioned that it wasn't complete yet since I wanted to discuss if we can make any further improvements when a joystick does exist - for example, an API to allow the user to supply their own arrays, or an alternative that can help reduce or eliminate garbage if the user wants to.

Jjagg:
We can discuss it further in #6820 :)

PR #6834, opened 2019-08-02
Further garbage optimizations to Joystick

tdeeb:
Relevant issue: https://github.com/MonoGame/MonoGame/issues/6820

Summary of changes:

- Added a Joystick.GetState overload taking in a `ref JoystickState` that is updated, and implemented the overload on all publicly accessible platforms
- Moved `_defaultJoystickState` to the shared file so all platforms can use it
- Made `Joystick.Web` use the default joystick's zero-length arrays in several instances to avoid allocations

I do not see a project for Web, so I'm not sure if the changes to Web compiles.

### DIFF
--- a/MonoGame.Framework/Input/Joystick.cs
+++ b/MonoGame.Framework/Input/Joystick.cs
@@ -10,6 +10,17 @@ namespace Microsoft.Xna.Framework.Input
     public static partial class Joystick
     {
         /// <summary>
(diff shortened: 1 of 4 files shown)


tdeeb:
@Jjagg Can you review these changes on your convenience?

Jjagg:
Thanks @tdeeb, looks good!

Issue #6837, opened 2019-08-03
GraphicsDevice.EffectCache is not threadsafe


dmanning23:
I have a project where it is loading content in different threads. Very occasionally, it will crash when loading Effect objects:

```
System.ArgumentException
  HResult=0x80070057
  Message=An item with the same key has already been added.
  Source=mscorlib
  StackTrace:
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Microsoft.Xna.Framework.Graphics.Effect..ctor(GraphicsDevice graphicsDevice, Byte[] effectCode, Int32 index, Int32 count)
   at Microsoft.Xna.Framework.Content.ContentTypeReader`1.Read(ContentReader input, Object existingInstance)
   at Microsoft.Xna.Framework.Content.ContentReader.InnerReadObject[T](T existingInstance)
   at Microsoft.Xna.Framework.Content.ContentReader.ReadObject[T]()
   at Microsoft.Xna.Framework.Content.ContentReader.ReadAsset[T]()
   at Microsoft.Xna.Framework.Content.ContentManager.ReadAsset[T](String assetName, Action`1 recordDisposableObject)
   at Microsoft.Xna.Framework.Content.ContentManager.Load[T](String assetName)
   at BloomBuddy.Bloom.LoadContent(SpriteBatch spriteBatch, ContentManager content, GraphicsDeviceManager device)
...
```

Chased it down to this line of code:

https://github.com/MonoGame/MonoGame/blob/ac217ce1f6de1d9764f84983d5c043c5d44aa9aa/MonoGame.Framework/Graphics/Effect/Effect.cs#L110

Effects are being cached in the GraphicsDevice object, but there is no thread protection around that cache. There are two solutions I can see:

1. Add a lock around access to the EffectCache
2. Move the EffectCache into ContentManager. That would be the logical place to cache it.

Super appreciate all the work yall are putting into MonoGame.
Cheers!

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

PR #6838, opened 2019-08-05
6837: Moved Effect caching logic into GraphicsDevice, added thread lock

dmanning23:
Fixes the thread safety issue described in:
https://github.com/MonoGame/MonoGame/issues/6837

- Moved the logic for caching Effect objects into the GraphicsDevice
- Added a thread lock around access to the EffectCache

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -526,7 +526,7 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <param name="effectKey"></param>
         /// <param name="headerSize"></param>
         /// <returns></returns>
-        internal Effect GetEffect(GraphicsDevice graphicsDevice, byte[] effectCode, int index, int count, int effectKey, int headerSize)
+        internal Effect GetEffect(byte[] effectCode, int index, int count, int effectKey, int headerSize)
         {
             Effect cloneSource;
 
@@ -539,11 +539,11 @@ namespace Microsoft.Xna.Framework.Graphics
                     using (var reader = new BinaryReader(stream))
                     {
                         // Create one.
-                        cloneSource = new Effect(graphicsDevice);
+                        cloneSource = new Effect(this);
                         cloneSource.ReadEffect(reader);
 
                         // Cache the effect for later in its original unmodified state.
-                        graphicsDevice.EffectCache.Add(effectKey, cloneSource);
+                        EffectCache.Add(effectKey, cloneSource);
                     }
                 }
             }
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -8,7 +8,7 @@ using System.Diagnostics;
 using System.Globalization;
 using MonoGame.Utilities;
 using System.Runtime.InteropServices;
-
+using System.IO;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
@@ -118,9 +118,14 @@ namespace Microsoft.Xna.Framework.Graphics
         private readonly ConstantBufferCollection _pixelConstantBuffers = new ConstantBufferCollection(ShaderStage.Pixel, 16);
 
         /// <summary>
+        /// Effects can be added to the EffectCache from multiple threads.
+        /// </summary>
+        private readonly object _effectCacheLock = new object();
+
+        /// <summary>
         /// The cache of effects from unique byte streams.
         /// </summary>
-        internal Dictionary<int, Effect> EffectCache;
+        private Dictionary<int, Effect> EffectCache;
 
         // Resources may be added to and removed from the list from many threads.
         private readonly object _resourcesLock = new object();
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -116,11 +116,16 @@ private bool PixelShaderDirty
 
         private readonly ConstantBufferCollection _vertexConstantBuffers = new ConstantBufferCollection(ShaderStage.Vertex, 16);
         private readonly ConstantBufferCollection _pixelConstantBuffers = new ConstantBufferCollection(ShaderStage.Pixel, 16);
+
+        /// <summary>
+        /// Effects can be added to the EffectCache from multiple threads.
+        /// </summary>
+        private readonly object _effectCacheLock = new object();
```
nkast: You could lock the EffectCache object directly. There is no need for an extra lock object.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -509,6 +514,41 @@ public DepthStencilState DepthStencilState
...
+        /// Get an effect and store it in the cache if necessary.
+        /// </summary>
+        /// <param name="graphicsDevice"></param>
+        /// <param name="effectCode"></param>
+        /// <param name="index"></param>
+        /// <param name="count"></param>
+        /// <param name="effectKey"></param>
+        /// <param name="headerSize"></param>
+        /// <returns></returns>
+        internal Effect GetEffect(GraphicsDevice graphicsDevice, byte[] effectCode, int index, int count, int effectKey, int headerSize)
```
nkast: passing graphicsDevice is no longer needed. All references to graphicsDevice can be replaced with 'this.'

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -509,6 +514,41 @@ public DepthStencilState DepthStencilState
...
+        {
+            Effect cloneSource;
+
+            lock (_effectCacheLock)
+            {
+                //First look for it in the cache.
+                if (!EffectCache.TryGetValue(effectKey, out cloneSource))
```
nkast: Here we check this.EffectCache ...

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -509,6 +514,41 @@ public DepthStencilState DepthStencilState
...
+                {
+                    using (var stream = new MemoryStream(effectCode, index + headerSize, count - headerSize, false))
+                    using (var reader = new BinaryReader(stream))
+                    {
+                        // Create one.
+                        cloneSource = new Effect(graphicsDevice);
+                        cloneSource.ReadEffect(reader);
+
+                        // Cache the effect for later in its original unmodified state.
+                        graphicsDevice.EffectCache.Add(effectKey, cloneSource);
```
nkast: ... while here we add the effectKey to graphicsDevice.EffectCache, which can be a different object.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -116,11 +116,16 @@ private bool PixelShaderDirty
...
+        /// <summary>
+        /// Effects can be added to the EffectCache from multiple threads.
+        /// </summary>
+        private readonly object _effectCacheLock = new object();
 
         /// <summary>
         /// The cache of effects from unique byte streams.
         /// </summary>
-        internal Dictionary<int, Effect> EffectCache;
+        private Dictionary<int, Effect> EffectCache;
```
Jjagg: Why not use `ConcurrentDictionary` with [`GetOrAdd`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=netframework-4.8#System_Collections_Concurrent_ConcurrentDictionary_2_GetOrAdd__1__0_System_Func__0___0__1____0_)? Let the runtime handle things :)

tomspilman: I don't mind the lock approach... simple, easy to debug, not dependent on .NET lib features.

Moving it to GraphicsDevice *i think* mirrors some of the other cleanups we have done in other caches right?

dmanning23: Yeah I copied the same locking mechanism to match the coding standard used elsewhere in the codebase:

https://github.com/MonoGame/MonoGame/blob/bdc5951e8ffe15d61253d9aa6b01f5cb371be52c/MonoGame.Framework/Graphics/GraphicsDevice.cs#L131

I tried to keep these changes as safe as possible, since this code worked as-is for 99% of use cases. None of this is actually new code, it is just moved around and behind thread safety :)

MonoGame.Framework/Graphics/GraphicsDevice.cs
dmanning23: That's an excellent point, I'll go change that.

SimonDarksideJ:
Any chance you can resolve the merge issues @dmanning23 so this can be retested?

dmanning23:
You can go ahead and close this, I'm pretty sure the merge conflicts were because someone else fixed this bug in another PR.

SimonDarksideJ:
Thanks for the contribution @dmanning23

PR #6841, opened 2019-08-09
Uwp sdk

Jjagg:
cc @cra0zy

Getting UWP to build in #6824 was driving me insane, so let's migrate this. It'll also be a good starting point to complete the migration and actually drop the old projects.
~Before merging I would like to get Guide into the main assembly so we can drop the Net project.~

EDIT: Actually, I'm of the opinion we should drop Guide for now and offer a new API in the MonoGame.Framework assembly. What do you think @cra0zy?
If all is green, this can be merged if you ask me!

This migration will change how NuGet packages are generated. Instead of the nuspec file, MSBuild will handle NuGets by using `msbuild /t:pack`. But the CAKE build script will handle that :) 

(side note: we should look into setting up a package feed - I'd like for it to be here on GitHub - and have the CAKE build script publish the development nugets on there)


cra0zy:
@Jjagg Conflicts!

Jjagg:
@cra0zy Fixed!

PR #6844, opened 2019-08-09
Disable Piranha portable build

Jjagg:
This drops the PCL build in favor of the DesktopGL .NET Standard build.
We need to document how to write a lib for all platforms.


Jure-BB:
So, MG is finally doping Piranha? That means I can close #6021?

tomspilman:
So how does this affect things like the Piranha build of the content pipeline assembly?

That is fantastic when writing custom pipeline extensions and not needing to link against a specific platform's version of MG.  When your extension is loaded in the pipeline tool or MGCB it just works.

Will that still be true?

Jjagg:
There is now a Content.Pipeline project that is cross platform. We don't need bait and switch anymore for the Content.Pipeline assembly. 
The cross platform Content.Pipeline project references DesktopGL, but that can be changed to another runtime if necessary with bait-and-switch.

cra0zy:
And also the MG.Content.Pipeline main nuget can be used as bait and switch.

PR #6846, opened 2019-08-10
Added ability to determine the underlying game platform type during the run-time

rds1983:
This is same as #5508 but done from dedicated branch.


Jjagg:
Please don't do these formatting changes to trigger the build bot. The build bots run fine, but the integration just doesn't show properly on GitHub. You can see the builds if you go to the teamcity server.

I didn't review this when it was originally proposed so here's my feedback.

I'm not a fan of the name `GamePlatformType`, I'd rather we call `MonoGamePlatform` or just `GamePlatform`. I also think we shouldn't expose the platform through `Game`, because it's basically hardcoded into the assembly since we compile MG for a specific platform, so IMO it should be a static property. I'd put a getter in the `GamePlatformType` as `GamePlatFormType.Current` or maybe a method `GamePlatformType.Get()` to differentiate it from the defined platforms (which are properties). The implementation then calls (if it's a property) or is (if it's a method) a partial method that returns the platform.

The console platforms need to be added to the list, because otherwise you'd still need conditional compilation as a user to compile which defeats the purpose of this addition because if you're going to conditionally compile based on your target platform you already know your target platform.

Then finally I don't like the hashcode comparison. Equal hashcodes doesn't mean the strings are equal. I know the probability is virtually zero of collisions here, but it's still bad practice. I'd rather we'd just do string comparison. If you really want the optimization, we should make `GamePlatformType` a class, let the partial classes set the current platform and check for reference equality instead of value equality.

The constructor for `GamePlatformType` should be private.

Could you add docs to the public members as well? I think that's especially important for API that isn't in XNA (I also think we should have examples for all new API, but that's not for this PR).

rds1983:
> Please don't do these formatting changes to trigger the build bot. The build bots run fine, but the integration just doesn't show properly on GitHub. You can see the builds if you go to the teamcity server.
Thing is sometimes build bots dont run.
I.e. this PR has two commits that was made within one hour.
However build bot ran only after 2nd commit:
image

Jjagg:
Oh, you're right! I take back what I said :p

rds1983:
I agree solution with GamePlatformType struct isn't elegant one. Thus I've come with new one, which involves enum(MonoGamePlatformType) with all possible platforms and partial class MonoGamePlatform.

Jjagg:
Oh yeah, actually an enum is fine because the code is only useful if the platform value is present in all builds :)

MonoGame.Framework/MonogamePlatform.cs
```diff
@@ -0,0 +1,36 @@
...
+        Android,
+        iOS,
+        tvOS,
+        DesktopGL,
+        Windows,
+        WindowsUniversal,
+        Web,
+        PSVita,
+        XBoxOne,
+        PlayStation4
```
Jjagg: Can you add Switch as well?

Jjagg:
Hmm, I would like the enum to be called `MonoGamePlatform` because it's the target platform, not the *type* of the target platform, but not sure what the name of the exposing class should be then... Maybe it should go in an existing class? What's your opinion @rds1983? cc @cra0zy for your opinion as well.

rds1983:
@Jjagg 
I'd keep it separate class and name it **PlatformInfo**. And I'd rename Current property to MonoGamePlatform too.
As in future this PlatformInfo could gain more functionality. Like reporting type of underlying OS. Or graphical system(DirectX or OpenGL). Etc.

Jjagg:
I think that's a great idea! :)

MonoGame.Framework/MonogamePlatform.cs
rds1983: Fixed

rds1983:
After some thought I figured even more elegant solution that changed only 3 files.

rds1983:
I've did small refactoring, so now only 3 files affected by this PR instead of 18.

MonoGame.Framework/PlatformInfo.Platform.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+#endif
+
+#if WINDOWS_UAP
+    partial class PlatformInfo
+    {
+        private static MonogamePlatform PlatformGetCurrent()
+        {
+            return MonogamePlatform.WindowsUniversal;
```
Jjagg: You can have only the return statement in the `#if` `#endif` block i.e.
```
partial class PlatformInfo
{
    private static MonoGamePlatform PlatformGetCurrent()
    {
#if DESKTOPGL
        return MonoGamePlatform.DesktopGL;
#endif
        // etc....
    }
}
```

rds1983: I did it that way on purpose, so closed platforms could make their own implementation of PlatformGetCurrent().
Though if I'll be provided closed platforms defines, I'd do it the way you propose.

Jjagg: Right... Let's keep it like this for now.

rds1983: After more thinking I moved all that stuff under one class declaration. So now code looks like this:
```c#
#if ANDROID
        private static MonogamePlatform PlatformGetCurrent()
        {
            return MonogamePlatform.Android;
        }
#endif

#if DESKTOPGL
        private static MonogamePlatform PlatformGetCurrent()
        {
            return MonogamePlatform.DesktopGL;
        }
#endif
```

rds1983:
Btw, regarding API that isnt in XNA?
How about move it all in the dedicated namespace? Say **MonoGame.Framework.Extension**?
This way it wouldn't be required to explicity say that every such API isn't in XNA.

Jjagg:
I did not actually mean it should be documented that it's not XNA API, I meant it should be document *because* it's not XNA API. Otherwise there's no way for people to discover it, that's why I also mentioned examples. That said, I do think it's a good idea to put new stuff in the MG namespace (not .Extension), but we'll have to discuss that.

rds1983:
I see.

tdeeb:
I would like to see this merged. Is documentation all that's remaining?

Jjagg:
I was waiting for a response from @tomspilman about the defined symbols for closed platforms. But I'll merge this for now. We can always change it later.

cra0zy:
I request the following changes:

- Move this under `MonoGame.Framework.Utilities` namespace
- Rename `MonogamePlatform` to `MonoGamePlatform`
- Do the following for defines, make as little duplicate code as possible:
```
private static MonogamePlatform PlatformGetCurrent()
{
#if ANDROID
    return MonogamePlatform.Android;
#elif X
    return MonogamePlatform.X;
#endif
}
```

cc. @Jjagg @rds1983

Jjagg:
I agree with 1 and 2.

> Do the following for defines, make as little duplicate code as possible:

This is because of closed platforms. The way it is now, closed platforms can have a seperate file with the `PlatformGetCurrent` implementation.

cra0zy:
Its just platform defines of the closed platforms that we need, which are already public btw.

cra0zy:
No point in this being a partial class.

Jjagg:
>  which are already public btw

I was not aware of this. We're are they documented?

rds1983:
I've applied requested changes to the PR #6872

PR #6855, opened 2019-08-11
Made it so Texture2D in any format could be saved as Png/Jpeg when using MonoGame.DesktopGL backend

rds1983:
Right now, MonoGame.DesktopGL's Texture2D.SaveAsPng and Texture2D.SaveAsJpeg  work through StbImageWriteSharp and assume that saved texture has SurfaceFormat.Color.
Maybe that is what causes the issue: http://community.monogame.net/t/saveaspng-missing-1-4-of-image

This PR would make it save textures in any format.


MonoGame.Framework/Graphics/Texture2D.cs
```diff
@@ -422,5 +423,166 @@ private static void ConvertToABGR(int pixelHeight, int pixelWidth, int[] pixels)
...
+                    GetData(colorData);
+                    break;
+
+                case SurfaceFormat.Alpha8:
+                    var alpha8Data = new Alpha8[colorDataLength];
+                    GetData(alpha8Data);
+
+                    for (int i = 0; i < colorDataLength; i++)
+                    {
+                        colorData[i] = new Color(((IPackedVector)alpha8Data[i]).ToVector4());
```
Jjagg: Use new `Color(0, 0, 0, alpha8Data.PackedValue)` so you don't need ToVector4, the interface cast causes boxing. Similar fix for the other IPackedVectors.

It would be better if all the IPackedVectors had a `ToVector4` function that isn't an explicit interface implementation, so we can call the method directly on the struct to avoid boxing and the conversions here can just do `xData[i].ToVector4()`.

Jjagg: The explicit interface implementation and normal ToVector4 method can exist side by side btw.

Jjagg: Would be nice if these types are all readonly as well, so no shadow copying happens, but we can't do that because we're on an old C# version :/

rds1983: We can make ToVector4 public method, than it'll cover the interface method too.
That is what my new commit.

Jjagg:
Great! Thanks @rds1983!

rds1983:
Just a small notice that issue with png writing http://community.monogame.net/t/saveaspng-missing-1-4-of-image was related not to StbImageWriteSharp, but to PngWriter(which works in MonoGame.WindowsDX backend).
#6704 fixes that.

Issue #6827, opened 2019-07-23
Resource leak in DrawableGameComponent; UnloadContent is never called.


vpenades:
Checking the code, I noticed that `DrawableGameComponent` implements `LoadContent` and `UnloadContent` virtual methods.

- `LoadContent` is called from Initialize();
- `UnloadContent` is not called from anywhere, never.

I think that, at the very least, `DrawableGameComponent` should overload `GameComponent.Dispose(bool disposing)` and call `UnloadContent`.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7.1.189

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- WindowsDX

PR #6858, opened 2019-08-12
Override Dispose in DrawableGameComponent to unload content

tdeeb:
This PR fixes a memory leak in `DrawableGameComponent` by overriding `Dispose(bool)` and unloading its content.

Relevant issue: https://github.com/MonoGame/MonoGame/issues/6827
@vpenades @Jjagg


Jjagg:
Thanks!

PR #6859, opened 2019-08-13
Swap red and blue when setting mouse cursor texture

### LINKED ISSUE #6754: R and B channels swapped on SetCursor.FromTexture2D in WindowsDX
When setting a mouse cursor from Texture2D, the red and blue channels are swapped.  I see there was a previous PR that has already been merged awhile ago that addresses this for GL, but the same bug exists in DX as well.

Monogame 3.7.1 (latest version).

Jjagg:
Fixes #6754.

I haven't tested this yet.


Jjagg:
@cra0zy I confirmed this fixes the issue :)

Issue #3983, opened 2015-07-04
Video for MonoGame DesktopGL platforms


cra0zy:
Video class is still missing for WinGL / Linux. What do you thing would be the best thing to be used to play videos?

Some options I know of:
 - TheoraPlay-CS
 - GStreamer
 - Vlc.Net

CC. @tomspilman @flibitijibibo @KonajuGames @dellis1972

flibitijibibo:
The current FNA VideoPlayer is written in a way where it doesn't necessarily depend on any specific platform backends, but it _does_ depend on the IGLDevice. These files pretty much stand on their own otherwise:

https://github.com/flibitijibibo/FNA/blob/master/src/Media/Xiph/Video.cs
https://github.com/flibitijibibo/FNA/blob/master/src/Media/Xiph/VideoPlayer.cs
https://github.com/flibitijibibo/FNA/blob/master/src/Graphics/Effect/YUVToRGBA/YUVToRGBAEffect.fx

GL_pushState and GL_popState are where things get a little messy.

Shqrdx:
I think devs wants something universal and crossplatform for video playback.

KonajuGames:
> I think devs wants something universal and crossplatform for video playback.

Not necessarily across every platform, but cross-platform where it makes sense.  Similar to music where we generate M4A/AAC where the platform supports the playback, WMA on Windows DirectX platforms and Vorbis elsewhere.  We can use Theora on some platforms, and other formats where Theora is not suitable such as those platforms that have media playback but do not support Theora.  We need to be careful about the choice of video encoding we use in the content pipeline because of licensing.  This is why we do not generate MP3 files in the content pipeline.  MP3 files incur a licensing fee if the game ships more than 5,000 copies.  We are not going to encourage people to violate the licensing on those encoded files.

tomspilman:
This is something we should look to fix for the 3.6 release.

@cra0zy - It seems like Theora is our best bet for a cross-platform format for desktop platforms?

cra0zy:
> This is something we should look to fix for the 3.6 release.

SDL branch is blocking this.

> @cra0zy - It seems like Theora is our best bet for a cross-platform format for desktops?

Yea, one good side tho is that the original Theoraplay file is actually quite small (~800 lines) so we may want to think about translating it to C# instead of writing bindings, this way we could integrate it with NVoribs.

@mrhelmut We are having fun without you so I though to invite you :)

tomspilman:
[...]

PR #6860, opened 2019-08-14
DesktopGL Theora videoplayer

Jjagg:
See #3983.

This is WIP, I think. I haven't tested at least, but best-case scenario it already works.
Also added float buffer submitting support to OpenAL and used it in OggStreamer. Theorafile also has a vorbis implementation (native lib), so now we have two implementations. We should fix that after we get this merged by replacing NVorbis.

Thanks @flibitijibibo ;)


mrhelmut:
Cool! That was clearly missing.

This depends on the native ```theorafile``` binary as well, right? (it looks to be missing)
We'll have to make sure that it's not a requirement and that games shipping without videos don't need to ship that too (like OpenAL).

It might be interesting to consider using ```Theorafile-CS``` from source like ```NVorbis``` so that there's not another assembly to ship as well.

Jjagg:
I have the libs added locally, will test before submitting a PR to Dependencies. Adding from source is a good idea :) I'll copy over the source file instead of adding as a submodule so people don't clone the whole repo when building from source.

Jjagg:
Notes to self:
- Need to verify number of audio channels in the audio file is 1 or 2
- Need to make sure video format is supported for target platform (or transcode, but I don't think we should do that)

Jjagg:
Gonna close this for now because it needs a lot of work. We could go with a more modern codec for our video player implementation, like AV1.

PR #6863, opened 2019-08-17
Fix path to ExternalTool for content pipeline on .NET Core

Jjagg:
From what I understand, single quotes in command arguments are escaped on Unix in .NET Core to be consistent with Windows.
See https://github.com/dotnet/corefx/issues/38483 and mono issue referred to in there.

@cra0zy


Jjagg:
This doesn't work like this, there's places where we nest double quotes that need to be fixed.

PR #6864, opened 2019-08-19
Etc2 texture formats.

ryancheung:


ryancheung:
I really need the ETC2 texture formats. 😃

MonoGame.Framework/Platform/Graphics/OpenGL.iOS.cs
```diff
@@ -35,7 +35,7 @@ public class GraphicsContext : IGraphicsContext
     {
         public GraphicsContext ()
         {
-            Context = new EAGLContext (EAGLRenderingAPI.OpenGLES2);
+            Context = new EAGLContext (EAGLRenderingAPI.OpenGLES3);
```
Jjagg: Documentation is sparse, so I don't know what happens here if the device does not support GLES3. I assume it will either throw or return null. I don't see an easy way to fallback to GLES2 in the API. I think we need to create a context and use `glGetString` to check the supported version if we want to continue to support GLES2.

Not sure if we *should* still support GLES2 on iOS though. A quick search indicates iPhone supported GLES3 in iOS 7+ since iPhone 5S/iPad Air.

ryancheung: Yeah, iOS devices have  good upgrade rate to latest version.

tomspilman: Yeah i think for now it would be best to fallback to GLES2 if possible.  It would suck to hurt install base of existing users unless we really have to.

MonoGame.Framework/Platform/Graphics/OpenGL.iOS.cs
```diff
@@ -35,7 +35,13 @@ public class GraphicsContext : IGraphicsContext
...
         public GraphicsContext ()
         {
-            Context = new EAGLContext (EAGLRenderingAPI.OpenGLES2);
+            nuint major, minor;
+            EAGLContext.EAGLGetVersion(out major, out minor);
+
+            if (major >= 3)
+                Context = new EAGLContext (EAGLRenderingAPI.OpenGLES3);
+            else
+                Context = new EAGLContext (EAGLRenderingAPI.OpenGLES2);
```
ryancheung: I just added the fallback check. Could you be sure this is capable? @tomspilman @Jjagg

ryancheung:
Commit Edit: Init EGALContext with GLES3 if ios version is 7 or above.

ryancheung:
After upgrade my iPhone iOS to ios13+, bad things start to be seen:

1. `EAGLContext.EAGLGetVersion(out major, out minor);` returns major 1 and minor 0, before it's major 3.
2. Init EAGLContent with GLES3 and game crashing on starting, raising GL.GetError 1286.

ryancheung:
@Jjagg Could u consider merging this PR?

Jjagg:
Yeah, this is great! Thanks @ryancheung!
We'll should look into supporting etc2 in the content pipeline as well and making it the default compression for textures on mobile.

PR #6867, opened 2019-08-20
Replaced PngWriter with StbImageWriteSharp

rds1983:
This is follow up to #6703 
I did some refactoring and moved all StbSharp stuff to Texture2D.StbSharp.cs and removed code that seems to be obsolete.
Also I've made it so StbImageWriteSharp is used instead of NotImplementedException.
Some platform specific image writers are still used: for Android and UWP.

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/Texture2D.StbSharp.cs
+++ b/MonoGame.Framework/Platform/Graphics/Texture2D.StbSharp.cs
@@ -1,0 +1,105 @@
+// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using StbImageSharp;
+using StbImageWriteSharp;
+using System;
+using System.IO;
+
+namespace Microsoft.Xna.Framework.Graphics
+{
+    public partial class Texture2D
+    {
+        private unsafe static Texture2D PlatformFromStream(GraphicsDevice graphicsDevice, Stream stream)
+        {
+            byte[] bytes;
+
+            // Rewind stream if it is at end
+            if (stream.CanSeek && stream.Length == stream.Position)
+            {
+                stream.Seek(0, SeekOrigin.Begin);
+            }
+
+            // Copy it's data to memory
+            // As some platforms dont provide full stream functionality and thus streams can't be read as it is
+            using (var ms = new MemoryStream())
+            {
+                stream.CopyTo(ms);
+                bytes = ms.ToArray();
+            }
+
+            // The data returned is always four channel BGRA
+            var result = ImageResult.FromMemory(bytes, StbImageSharp.ColorComponents.RedGreenBlueAlpha);
+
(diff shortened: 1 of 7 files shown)


Jjagg:
That was fast :D 
Awesome, thanks @rds1983! I'll review this later.

Jjagg:
@rds1983 Why did you undo the last commit? Instead you should also not include the Texture2D.StbSharp file for UWP. Or did you also want to let UWP use StbImageWriteSharp?

rds1983:
I did that because UWP uses both StbImageWriteSharp and platform dependent code:
```C#
        private void PlatformSaveAsJpeg(Stream stream, int width, int height)
        {
#if WINDOWS_UAP
            SaveAsImage(Windows.Graphics.Imaging.BitmapEncoder.JpegEncoderId, stream, width, height);
#else
            SaveAsImage(stream, width, height, ImageWriterFormat.Jpg);
#endif
        }

        private void PlatformSaveAsPng(Stream stream, int width, int height)
        {
            SaveAsImage(stream, width, height, ImageWriterFormat.Png);
        }
```

Though I agree that it would be nice to have one codepath and will move PlatformSaveAsJpeg to StbImageWriteSharp only this evening.

Speaking of which.
Right android is the last backend that uses platform dependent image writing:
```C#
        private void PlatformSaveAsJpeg(Stream stream, int width, int height)
        {
#if !ANDROID
            SaveAsImage(stream, width, height, ImageWriterFormat.Jpg);
#else
            SaveAsImage(stream, width, height, Bitmap.CompressFormat.Jpeg);
#endif
        }

        private void PlatformSaveAsPng(Stream stream, int width, int height)
        {
#if !ANDROID
            SaveAsImage(stream, width, height, ImageWriterFormat.Png);
#else
            SaveAsImage(stream, width, height, Bitmap.CompressFormat.Png);
#endif
        }
```
How about move it to StbImageWriteSharp too?

Jjagg:
I'm good with moving everything to StbImageWriteSharp :)

rds1983:
Ok. I've moved everything on StbImageWriteSharp.
I didnt delete PngLibs as they might be used in the closed platforms.

Jjagg:
@tomspilman are the png classes used in any closed platforms?

Jjagg:
I'll merge this and open an issue for removing the png classes.
Thanks again, @rds1983! This is great work! :)

MonoGame.Framework/Platform/Graphics/Texture2D.StbSharp.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            // Copy it's data to memory
+            // As some platforms dont provide full stream functionality and thus streams can't be read as it is
+            using (var ms = new MemoryStream())
+            {
+                stream.CopyTo(ms);
+                bytes = ms.ToArray();
+            }
+
+            // The data returned is always four channel BGRA
```
harry-cpp: @rds1983  This returns RGBA format it seems XD

Issue #6453, opened 2018-09-28
graphicsDevice.PlatformDispose throws System.NullReferenceException


EnemyArea:
After upgrade to monogame 3.7, I've got an NullReferenceException on calling graphicsDevice.Dispose, if monogame is used in a wpf application. Thats because, there is no _swapChain. The line is the first line @ PlatformDispose

"if (_swapChain.IsFullScreen)" <- Here should be checked whether _swapChain is not null

Before 3.7 that error not exists. I have testet on the latest dev branch on Windows 10 Prof.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows 10

#### What MonoGame platform are you using:

- WindowsDX

craftworkgames:
I ran into the same issue. 

The problem should be reproduceable in my [MonoGame.WpfCore project](https://github.com/craftworkgames/MonoGame.WpfCore).

The exception occurs when the application is shutting down and the `GraphicsDevice` is disposed on [line 18](https://github.com/craftworkgames/MonoGame.WpfCore/blob/master/MonoGame.WpfCore/MonoGameControls/MonoGameGraphicsDeviceService.cs#L18). Although it doesn't seem to happen all the time (maybe some sort of race condition or maybe the debugger is detached before it happens, I'm not sure).

Here's the stack trace I got:

```
   at SharpDX.DXGI.SwapChain.GetFullscreenState(RawBool& fullscreenRef, Output& targetOut)
   at SharpDX.DXGI.SwapChain.get_IsFullScreen()
   at Microsoft.Xna.Framework.Graphics.GraphicsDevice.PlatformDispose()
   at Microsoft.Xna.Framework.Graphics.GraphicsDevice.Dispose(Boolean disposing)
   at Microsoft.Xna.Framework.Graphics.GraphicsDevice.Dispose()
   at SpriteFactory.MonoGameControls.MonoGameGraphicsDeviceService.Dispose()
   at SpriteFactory.MonoGameControls.MonoGameContentControl.CleanUp()
   at SpriteFactory.MonoGameControls.MonoGameContentControl.Finalize()
```

The problem is quite clear after stepping into the code. In the `PlatformDispose` method the `_swapChain` variable is used to change the fullscreen state without checking it for `null`.

```
      if (this._swapChain.IsFullScreen)
        this._swapChain.SetFullscreenState((RawBool) false, (Output) null);
```

[...]

PR #6869, opened 2019-08-21
Fix issue #6453 - SwapChain is null in GraphicsDevice.PlatformDispose

craftworkgames:
This is a very small fix to resolve issue #6453 and prevent `GraphicsDevice.PlatformDispose` throwing `NullReferenceException` in WPF applications.


Jjagg:
Thanks!

PR #6871, opened 2019-08-22
iOS core build with Keyboard.GetState method

ryancheung:
Currently, nuget package of iOS core build with cake has no `Keyboard.GetState` method. And this break thirdparty librarie that are platform agnostic and depend on that method, which stops project depending on these libraries compiling(Keyboard.GetState method not found).

Here is a instance: https://github.com/craftworkgames/MonoGame.Extended/blob/develop/Source/MonoGame.Extended.Input/InputListeners/KeyboardListener.cs#L40

The keyboard class was intentionally removed. I don't why. It could sit there and wouldn't affect anything.


Jjagg:
I'm sorry, I completely missed this PR. You're right, this is a mistake made with the migration, Android is also missing the keyboard class. Thank you for the PR!

ryancheung:
Thank you for merging.

PR #6877, opened 2019-09-03
Work on SongCollection

### LINKED ISSUE #3352: Make SongCollection constructor public
Monogame 3.2 allows construction and modification to class SongCollection. This is now changed in latest dev. Here is code that works in 3.2:

  Songs =  new Microsoft.Xna.Framework.Media.SongCollection();

Songs.Add((Song)item.Value);

Latest dev build 1/2/2015 throws this error:

Error   2   The type 'Microsoft.Xna.Framework.Media.SongCollection' has no constructors defined

a2937:
Fixes  #3352

Hopefully the documentation is acceptable. I used the base documentation as reference. At the very least the song collection has a public constructor.

### DIFF
--- a/MonoGame.Framework/Media/SongCollection.cs
+++ b/MonoGame.Framework/Media/SongCollection.cs
@@ -8,14 +8,23 @@ using System.Collections.Generic;
 
 namespace Microsoft.Xna.Framework.Media
 {
-	public class SongCollection : ICollection<Song>, IEnumerable<Song>, IEnumerable, IDisposable
-	{
-		private bool isReadOnly = false;
-		private List<Song> innerlist = new List<Song>();
 
-        internal SongCollection()
+    /// <summary>
+    /// Class that represents a collection of songs.
+    /// </summary>
+    public class SongCollection :  ICollection<Song>, IEnumerable<Song>, IEnumerable, IDisposable
+    {
+        private bool disposed;
+        private bool isReadOnly = false;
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Media/SongCollection.cs
```diff
@@ -5,67 +5,115 @@
...
+
+        /// <summary>Adds an item to the <see cref="SongCollection" />.</summary>
+        /// <param name="item">The object to add to the <see cref="SongCollection" />.</param>
+        /// <exception cref="ArgumentNullException">The song is null.</exception>
+        public void Add(Song item)
         {
 
             if (item == null)
-                throw new ArgumentNullException();
+                throw new ArgumentNullException(nameof(item));
```
tdeeb: The build bot uses C# 5 for console compatibility, so this will have to be replaced with "item."

cra0zy: Yup.

nkast:
SongCollection is a special type of collection that returns the songs found on a device.
It's implementation might be specific to a platform and users shouldn't use it as a generic collection of songs. Actually it shouldn't had public Add()/Remove() methods either.

https://docs.microsoft.com/en-us/previous-versions/windows/xna/dd282398(v%3Dxnagamestudio.41)

For custom playlists you should create a Collection<Song> or a List<Song>, etc.

PR #6880, opened 2019-09-11
Fixing project definitions

mrhelmut:
Hi :)

Just fixing Linux x86 and the solution file to properly build the framework and tests for DesktopGL.

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
@@ -100,6 +100,17 @@ This package provides you with MonoGame Framework that uses OpenGL for rendering
       <PackagePath>runtimes\win-x64\native</PackagePath>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+	
+	<Content Include="..\ThirdParty\Dependencies\openal-soft\Linux\x86\libopenal.so.1">
+      <Link>x86\libopenal.so.1</Link>
+      <PackagePath>runtimes\linux-x86\native</PackagePath>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="..\ThirdParty\Dependencies\SDL\Linux\x86\libSDL2-2.0.so.0">
+      <Link>x86\libSDL2-2.0.so.0</Link>
+      <PackagePath>runtimes\linux-x86\native</PackagePath>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
 
     <Content Include="..\ThirdParty\Dependencies\openal-soft\Linux\x64\libopenal.so.1">
       <Link>x64\libopenal.so.1</Link>
--- a/MonoGame.Framework.DesktopGL.sln
+++ b/MonoGame.Framework.DesktopGL.sln
@@ -7,6 +7,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MonoGame.Framework.DesktopG
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MonoGame.Tests.DesktopGL", "Test\MonoGame.Tests.DesktopGL.csproj", "{D80F2C27-F1E6-4395-95E2-D6EAB308C9DF}"
 EndProject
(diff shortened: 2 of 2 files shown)


MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
```diff
@@ -100,6 +100,17 @@ This package provides you with MonoGame Framework that uses OpenGL for rendering
       <PackagePath>runtimes\win-x64\native</PackagePath>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+	
+	<Content Include="..\ThirdParty\Dependencies\openal-soft\Linux\x86\libopenal.so.1">
```
Jjagg: Can you use spaces here instead of a tab?

cra0zy:
Wait wait wait:
- I made the DesktopGL solution just for it, I was planning on adding a separate 3 solutions for Tools (when I do the Pipeline Tool stuff)
- .NET Core actually does not support x86 on Linux, that's why I did not include the Linux libs

mrhelmut:
Alright for Linux.
Though I had to add the Pipeline reference to the solution because VS2017 didn't let me build the tests without it.

cra0zy:
> Alright for Linux, but what about non-Core users?

You need to fixup the .targets file to get it working, tho honestly I would just drop us supporting x86 libs for Linux.

> Though I had to add the Pipeline reference to the solution because VS2017 didn't let me build the tests without it.

Hmm, I assume some tests target the content pipeline stuff, those should be moved to the new Tools related solutions.

mrhelmut:
We should move those tests later. For now I'm trying to make things work as-is.

There's multiple issues with references here.

```MonoGame.Framework.DesktopGL.sln``` needs a reference to ```MonoGame.Framework.Content.Pipeline.csproj``` for tests to work.

```MonoGame.Framework.WindowsDX.sln``` needs a reference to ```MonoGame.Framework.Content.Pipeline.csproj``` for tests to work.

The problem is that ```MonoGame.Framework.Content.Pipeline.csproj``` references ```MonoGame.Framework.DesktopGL.csproj``` and therefore can't work for ```WindowsDX```.

Anyway to resolve these dependencies without having dedicated csproj for ```MonoGame.Framework.Content.Pipeline```?

mrhelmut:
Also, is there any reason why WindowsDX targets ```net471``` and DesktopGL ```netstandard2.0;net451```?

cra0zy:
> We should move those tests later. For now I'm trying to make things work as-is.

These new projects are not meant to be hacky workarounds. Also these are still WIP, and not really meant to be used just yet...

> There's multiple issues with references here.
> 
> `MonoGame.Framework.DesktopGL.sln` needs a reference to `MonoGame.Framework.Content.Pipeline.csproj` for tests to work.
> 
> `MonoGame.Framework.WindowsDX.sln` needs a reference to `MonoGame.Framework.Content.Pipeline.csproj` for tests to work.

Only because tests also test the content pipeline as well as the framework itself, once the tests are split this won't be an issue.

> Also, is there any reason why WindowsDX targets `net471` and DesktopGL `netstandard2.0;net451`?

You can't target .netstandard2.0 with the WindowsDX project, it uses WinForms. As to why it targets net471 and not net451, don't remember, @Jjagg probably does.

mrhelmut:
> Also these are still WIP, and not really meant to be used just yet...

What need to be done? I'd like to help moving things faster but as-is I don't know what those ```.sln``` are meant for because they don't work.

So I'll start splitting the tests.

Jjagg:
WindowsDX should target net451 (or something old like that that's installed by default on old Windows OS's), I don't remember why it targets 471. .NET Core 3.0 is around the corner and we'll add that as a target for WindowsDX when it's out because it supports WinForms.

mrhelmut:
Okay, I'll leave of this be for now. I'll rebase the linux x86 things only.

Jjagg:
@mrhelmut I'm curious if you have any metrics on Linux 32-bit vs 64-bit downloads. Distros are dropping 32-bit support one by one, and like cra0zy said, even .net Core doesn't support 32-bit Linux. So I wonder how much it's actually still used.

mrhelmut:
According to Steam stats:
- Linux 32bit < 0.01% (likely ~0%)
- macOS 32bit < 0.01% (likely ~0%)
- Windows 32bit ~ 1.20% (which is more than 1 million users, so still relevant I'd say)

Well, right, let's drop 32bit support for both Linux and macOS.
Ignore my resubmission (pfeww, that was a productive attempt to contribute ahah).

Jjagg:
Hey, now we know for sure :)

PR #6890, opened 2019-10-03
[iOS][Android] Fix Texture2D reference

jnoyola:
#### Description
Add `Texture2D.StbSharp.cs` to iOSCore and AndroidCore.

#### Explanation
The cake builds for iOSCore and AndroidCore are failing for me on macOS-10.14 with the following errors:
```
Graphics/Texture2D.cs(328,13): error CS0103: The name 'PlatformSaveAsJpeg' does not exist in the current context
Graphics/Texture2D.cs(339,13): error CS0103: The name 'PlatformSaveAsPng' does not exist in the current context
```

- It looks like #6867 intended to add `Texture2D.StbSharp.cs` to all platforms, but only modified desktop platforms
- In develop right now, iOSCore and AndroidCore only use `Texture2D.OpenGL.cs`, but #6867 removed `PlatformSaveAsJpeg` and `PlatformSaveAsPng` from that

After these changes, iOSCore and AndroidCore build fine for me.


Jjagg:
Thanks!

Issue #6879, opened 2019-09-11
[Todo list] Future-proofing MonoGame


mrhelmut:
Hey there!

Many efforts have been started to future-proof MonoGame and make it a modern .Net framework but it seems that there is no comprehensive centralized point where to understand how things should move forward and their current state.

Hence, I'm trying to maintain this issue to keep track of the different tasks. Please feel free to point me to any untracked tasks or progress.

I guess that we will agree on the following main objective: **making MonoGame fully .Net Standard 2.0 compliant in its entirety** (all targets, all of its tools, and its build/test/package process), while maintaining a backward compatibility where it's relevant.

This would mean: being able to consume MonoGame as a ```dotnet``` (Core) framework, and as a full .Net framework.

So here comes the long list. @cra0zy @Jjagg @tomspilman please let me know if this list makes sense to you and if there's anything more to say about the current state of the repository.

**Moving to .Net Standard 2.0 (and .Net Core compatibility):**
- [x] Removing Protobuild & switching to SDK-style project format #6207 #6768 #4975 #6604
  - [x] MonoGame.Framework
    - [x] Windows DirectX
    - [x] Windows UWP
    - [x] DesktopGL
    - [x] Android
    - [x] iOS
  - [x] Archive MonoGame.Framework.Net to a dedicated repo (#6900) -> https://github.com/MonoGame/MonoGame.Framework.Net
  - [x] MonoGame.Framework.Content.Pipeline
    - [x] Windows
    - [x] macOS
    - [x] Linux
  - [x] Tools/2MGFX
    - [x] Windows/macOS/Linux
  - [x] Tools/MGCB
    - [x] Windows/macOS/Linux
  - [x] Tools/Pipeline
    - [x] Windows/macOS/Linux
  - [x] Remove Protobuild and any file/folder related to it
- [x] Making all tools Dotnet Tools and packaging them as nugets (and make the templates to use them)
  - [x] MGCB #6905 #6930
  - [x] 2MGFX #6930
  - [x] Pipeline #6922
- [x] Making Core templates (that uses global tools) and publishing them as nuget for ```dotnet new -i``` #6930
  - [x] Windows DirectX
  - [x] ~Windows UWP~ (not .net core compatible)
  - [x] DesktopGL
  - [x] ~Android~ (not .net core compatible)
  - [x] ~iOS~ (not .net core compatible)
[...]

PR #6901, opened 2019-10-12
Enforce C#5 in MonoGame.Framework

Jjagg:
See #6879 

@mrhelmut Do we need this for MonoGame.Framework.Content.Pipeline as well?


cra0zy:
> Do we need this for MonoGame.Framework.Content.Pipeline as well?

We don't.

mrhelmut:
cra0zy is right. This looks cool!

tomspilman:
Looks good to me @Jjagg !

Issue #6898, opened 2019-10-09
Use dotnet tools during build


jnoyola:
This is a question about the design and usage of the recent dotnet tools: 2MGFX and MGCB. I haven't found much discussion of this, but please point me in the right direction if I missed a thread. Apologies for the verbosity.

#### Question

Do we expect users to manually install dotnet tools before being able to build their game, or should they get the functionality automatically from the template or a PackageReference?

I'm under the impression that MonoGame is moving toward a model where a dev can clone a game repo and build without requiring any manual installation.

#### Problem

Dotnet tools seem to be intended more as manual helpers, rather than for use in automated build steps, for the following reasons:

1. Dotnet tools are not automatically restored on build. `dotnet restore` is run on build, but still not `dotnet tool restore`. This means tools need to be manually and explicitly installed or restored.
2. Dotnet tools cannot be transitive dependencies, by design, as opposed to package references.  You won't be able to reference a MonoGame.Content.Builder package that in turn references the MGCB tool. This means that any project intending to use MGCB will either need the tool installed globally, or listed in a manifest for local tool restore.

#### Solutions

Given those constraints, I see the following options:

1. Publish a MonoGame.Content.Builder package that expects the MGCB tool to be present.
    - The user can then choose to install the tool globally, or check in a local tool manifest and manually restore from it.
    - **Concern:** The user must manually install or restore the tool.
2. Publish a MonoGame.Content.Builder package that depends on the MGCB tool, but automatically runs `dotnet tool restore`.
    - The user can still choose to install the tool globally or check in a local tool manifest. In the case of a local manifest, though, it will be automatically restored as part of build. I'm not sure if this has perf implications or why it's not already automatic.
    - **Concern:** The repo must contain a local manifest, or the user must manually install the tool.
3. Keep MonoGame.Content.Builder as is.
    - The package has full control over the MGCB dependency, requiring no manual action.
    - **Concern:** This is somewhat non-standard and doesn't use the dotnet tools at all.

Jjagg:
I think the best way to go is to have the MSBuild task that's run by MonoGame.Content.Builder try to find an installed mgcb and install it if no installed version is found. I'm thinking it should try things in the following order:
- Check for MSBuild property `MGCBCommand` to override the command that's used to invoke mgcb (can be used to specify local executable or for additional build steps)
- Try locally installed mgcb tool, check for manifest if a local version can be restored
- Try globally installed mgcb tool
[...]

PR #6905, opened 2019-10-14
Add MonoGame.Content.Builder package

jnoyola:
### Description

A single NuGet package that combines the jobs of the old `MonoGame.Content.Builder.targets` and `MonoGame.Build.Tasks` steps while using the new MGCB dotnet tool. Addresses #6898.

1. Get the package version.
    - I considered several different approaches (reading the nuspec, generating a props file post-build), but decided on using the `$(Version)` property to pack a props file that inspects its own path. This option has the fewest moving parts, and relies only on pieces that we explicitly place (unlike the nuspec).
2. Run `dotnet tool list --local` to check for a locally installed MGCB tool with the same version.
3. Restore an existing MGCB tool of the correct version, or (re-)install the MGCB tool.
4. Collect content references.
    - Translation and simplification of the same task from `MonoGame.Build.Tasks`.
    - I opted to not use the existing `MonoGame.Build.Tasks` for several reasons. `dotnet pack` doesn't support packing files from `ProjectReferences` and the workaround doesn't work with `IncludeBuildOutput` disabled, and it didn't seem worth adding extra manual steps to copy the files or publishing a separate `MonoGame.Build.Tasks` NuGet package. Additionally, I believe MonoGame.Build.Tasks was originally written to accomodate xbuild, but that's no longer necessary, so the same steps can easily be performed in targets.
5. Build content.
6. Include content.

### Extensibility

All of the following MSBuild properties can be customized.

New
- `DotnetCommand`: dotnet command. Default: `dotnet`
- `MGCBCommand` MGCB command. Default: `$(DotnetCommand) tool run mgcb`
- `MGCBPackageId`: MGCB dotnet tool package ID. Default: `dotnet-mgcb`
- `MGCBVersion`: MGCB dotnet tool version. Default: `MonoGame.Content.Builder` version

Pre-existing
- `MonoGameMGCBAdditionalArguments`: additional arguments to pass to MGCB. Default: `/quiet`
- `PlatformResourcePrefix`: platform-specific prefix for included content paths. Default: depends on platform and prefixes defined by Xamarin

### Validation

I tested the following scenarios with WindowsDX, using a local package of `dotnet-mgcb`. All were successful in running MGCB, including content, and loading content in a game.

- Having no `dotnet-mgcb` installed
    - `MonoGame.Content.Builder` creates a manifest and installs the correct version of `dotnet-mgcb`
- Having the correct version of `dotnet-mgcb` installed in the project directory or parent directory
    - `MonoGame.Content.Builder` restores the installed version of `dotnet-mgcb`
- Having an old version of `dotnet-mgcb` installed in the project directory
    - `MonoGame.Content.Builder` uninstalls and reinstalls the correct version of `dotnet-mgcb`
- Having an old version of `dotnet-mgcb` installed in the parent directory
    - `MonoGame.Content.Builder` creates a manifest and installs the correct version of `dotnet-mgcb` in the project directory

### DIFF
--- a/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
+++ b/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net452</TargetFrameworks>
     <IncludeBuildOutput>false</IncludeBuildOutput>
   </PropertyGroup>
 
--- a/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
+++ b/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
@@ -1,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="MonoGame.Content.Builder.targets" PackagePath="build" />
+  </ItemGroup>
+
+</Project>
--- a/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
+++ b/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
@@ -6,7 +6,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Content Include="MonoGame.Content.Builder.props" PackagePath="build" />
     <Content Include="MonoGame.Content.Builder.targets" PackagePath="build" />
+    <Content Include="version\Version.props" PackagePath="build\version\$(Version)" />
   </ItemGroup>
 
 </Project>
--- a/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
+++ b/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net452</TargetFrameworks>
     <IncludeBuildOutput>false</IncludeBuildOutput>
-    <NoWarn>NU5128</NoWarn>
+    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
+++ b/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
@@ -8,7 +8,7 @@
   <ItemGroup>
     <Content Include="MonoGame.Content.Builder.props" PackagePath="build" />
     <Content Include="MonoGame.Content.Builder.targets" PackagePath="build" />
-    <Content Include="version\Version.props" PackagePath="build\version\$(Version)" />
+    <Content Include="version\Version.props" PackagePath="build\version\$(PackageVersion)" />
   </ItemGroup>
 
 </Project>
--- a/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
+++ b/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net452</TargetFrameworks>
     <IncludeBuildOutput>false</IncludeBuildOutput>
+    <NoWarn>NU5128</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.targets
+++ b/Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.targets
@@ -1,7 +1,5 @@
 ﻿<Project>
 
-  <UsingTask TaskName="MonoGame.Build.Tasks.CollectContentReferences" AssemblyFile="MonoGame.Build.Tasks.dll" />
-
   <!-- Add MonoGameContentReference to item type selection in Visual Studio -->
   <ItemGroup>
     <AvailableItemName Include="MonoGameContentReference" />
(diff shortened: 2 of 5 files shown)


Jjagg:
Awesome! This looks really good :) 
I'll give it a try once I find some time.

cc @cra0zy 

> decided on using the $(Version) property to pack a props file that inspects its own path

I'm not the biggest fan of this approach, it seems like a bit of a hack. Though it's not that important IMO.

jnoyola:
> it seems like a bit of a hack

In my opinion all approaches seemed like a hack, but this one seemed the best because again we control all the pieces. We explicitly set the props file to be included as e.g. `build\version\2.0.0\Version.props` and it knows that its parent folder will always be the version number.

I originally planned to use the `XmlPeek` task on `..\monogame.content.builder.nuspec` but that seemed just as hacky, a little more complicated of a task, and wouldn't work if it was run outside of a NuGet package context.

Let me know if you have other suggestions.

mrhelmut:
Awesome! I'll take a look at this when I'll have some spare time, but it sounds promising.

SimonDarksideJ:
Agreed, really opens up the netcore tooling for a much needed area

Tools/MonoGame.Content.Builder/version/Version.props
```diff
@@ -0,0 +1,5 @@
+﻿<Project>
+  <PropertyGroup>
+    <MonoGameContentBuilderVersion>$([System.IO.Path]::GetFileName($([System.IO.Path]::GetDirectoryName($(MSBuildThisFileFullPath)))))</MonoGameContentBuilderVersion>
```
Jjagg: `$([System.IO.Path]::GetDirectoryName($(MSBuildThisFileFullPath)))` => You can use $(MSBuildThisFileDirectory).

Jjagg:
```suggestion
    <MonoGameContentBuilderVersion>$([System.IO.Path]::GetFileName($(MSBuildThisFileDirectory)))</MonoGameContentBuilderVersion>
```

Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net452</TargetFrameworks>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
```
Jjagg: We can disable the warning for having no assembly for the TFM, see https://github.com/nuget/home/issues/8583.
```suggestion
    <IncludeBuildOutput>false</IncludeBuildOutput>
    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
```

Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
```diff
@@ -0,0 +1,14 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="MonoGame.Content.Builder.props" PackagePath="build" />
+    <Content Include="MonoGame.Content.Builder.targets" PackagePath="build" />
+    <Content Include="version\Version.props" PackagePath="build\version\$(Version)" />
```
Jjagg: Nuget uses PackageVersion for the version. If not specified it will take on the value of the Version property, but it can be set directly, so let's use PackageVersion instead of Version here.

```suggestion
    <Content Include="version\Version.props" PackagePath="build\version\$(PackageVersion)" />
```

Jjagg:
As an enhancement we could auto include any files with the mgcb extension unless some flag is set, similar to how SDK projects work.

```
<PropertyGroup>
  <EnableMGCBItems Condition=" '$(EnableMGCBItems)' == ''">true</EnableMGCBItems>
</PropertyGroup>

<ItemGroup Condition="'$(EnableMGCBItems)' == 'true'>
    <MonoGameContentReference Include="**/*.mgcb" />
</ItemGroup>
```

Tools/MonoGame.Content.Builder/version/Version.props
jnoyola: That's what I tried first, but `$(MSBuildThisFileDirectory)` has a trailing backslash, which unfortunately means `GetFileName()` won't work as you expect unless you also trim the trailing separator. [This answer](https://stackoverflow.com/a/57691321) and some of the comments on this post discuss this. What I have now is the most robust solution I could find.

jnoyola: Unfortunately this won't work; discussed in previous comment.

Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj
jnoyola: Added.

cra0zy:
@Jjagg I am in process of writing something custom for the content builder, among the other stuff, so I'm leaving this to you. I really don't wanna touch MGCB/Pipeline Tool anymore.

jnoyola:
I found a discrepancy in the build output if the MGCB item uses a link. Fixing...

jnoyola:
> I found a discrepancy in the build output if the MGCB item uses a link. Fixing...

Ok, fixed. Back to a good state now.

Jjagg:
@jnoyola I'm having some trouble testing this. What's your setup? I packed dotnet-mgcb and MonoGame.Content.Builder and added them to a local feed. Then I set up a console project with an mgcb file and added a reference to the content builder. That should be enough to run mgcb, right?

jnoyola:
> I'm having some trouble testing this.

I think I broke something :/
In it's current state, it looks like it only works if the mgcb file uses a link. I'll push a fix soon.

jnoyola:
The issue is that I tried fixing a potential bug in the current `CollectContentReferences.cs`, but it caused an inconsistency.

When setting `ContentFolder`, the original `CollectContentReferences.cs` uses `RelativeDir`. Presumably the intent is to capture cases where your MonoGameContentReference is `Subfolder\Content.mgcb`. However if your MonoGameContentReference is `..\Content.mgcb`, then your content will be placed at `..\`, which is _outside_ the output directory. So I opted to use `FullPath` instead. But then if `RelativeDir` is empty I still take the parent folder of the `FullPath`.

I'm not sure how this would be handled ideally, but I'll fix it to keep the current behavior, since no one's run into this issue. If this causes any pain, it should be fixed separately.

Jjagg:
RelativeDir is used to get the directory of the MGCB file relative to the project root because that's where the content should go. In case the file is linked, the value of `Link` is used because that's kind of the virtual path to the file inside the project directory. I haven't tested it, but the logic seems fine to me. Unless Link is not required for files outside the project directory, if that's possible we'd need to make an exception and put the content in the root of the output.

jnoyola:
Right, but this use of `RelativeDir` only makes sense if the MGCB file is in a subfolder, not a parent folder. For example, concatenating `$(OutputPath)\%(RelativeDir)\Content` could cause content to be placed in `netcoreapp2.0\..\Content\`, which is outside the build output directory.

In any case, it's fixed to match behavior of the previous MonoGame.Content.Builder package now. I tested the following cases:
- Using automatic `EnableMGCBItems` with an mgcb in project folder
- Using automatic `EnableMGCBItems` with an mgcb in subfolder
- Manually including mgcb in project folder
- Manually including mgcb in subfolder
- Manually including mgcb in parent folder (this is the case that seems broken, but at least consistent)

Jjagg:
I think if the file is outside the project folder and no link is set, we should either
- write a warning and skip the file or
- put content in the root directory.

We can detect this by checking if RelativeDir contains `..`. I don't think Link can be outside the project folder, but I'm not sure about that.

We could additionally let users override the content output folder by setting ContentFolder directly in the include. That's a bit more straightforward than setting Link for the MGCB file I think.

jnoyola:
Back from vacation.

I like the idea of allowing overriding the ContentFolder directly. Then the file structure of the output is disconnected from the file structure in Solution Explorer in VS. I know this is an extreme edge case, but folder names _can_ actually contain `..`, so maybe just a warning and continue.

To maintain compatibility, I can do this to preserve priority:

1. Use `ContentFolder` if present.
2. Use `Link` if present.
3. Use `RelativeDir` otherwise.
4. Throw warning if final value contains `..`

mrhelmut:
Alright, I tested this package on our current big project (which implies linked .mgcb files outside of the project root) and it works wonder.

mrhelmut:
Ah, forgot to note that it would be cool to add build/packaging to the Cake script.

jnoyola:
Everything should be addressed now with those commits.

mrhelmut:
Would you mind using ```PackProject``` in the cake script? Otherwise it's really cool! Clearly something that was greatly required.

jnoyola:
Cool, didn't realize we had that. Done.

mrhelmut:
I've extensively tested it and it's a go for me. Nice job @jnoyola !

Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.targets
```diff
@@ -0,0 +1,302 @@
...
+        <FullDir>$([System.String]::Copy(%(FullDir)).Replace('\','/'))</FullDir>
+        <ContentFolder Condition="'%(ContentFolder)' == '' AND '%(Link)' != ''">$([System.IO.Path]::GetDirectoryName(%(Link)))</ContentFolder>
+        <ContentFolder Condition="'%(ContentFolder)' == '' AND '%(Link)' == '' AND '%(RelativeDir)' != ''">$([System.IO.Path]::GetFileName($([System.IO.Path]::GetDirectoryName(%(RelativeDir)))))</ContentFolder>
+      </ContentReference>
+
+      <!-- Assemble final metadata. -->
+      <ContentReference>
+        <ContentDir>%(ContentFolder)/</ContentDir>
+        <ContentOutputDir>%(FullDir)bin/$(MonoGamePlatform)/%(OutputFolder)</ContentOutputDir>
+        <ContentIntermediateOutputDir>%(FullDir)obj/$(MonoGamePlatform)/%(OutputFolder)</ContentIntermediateOutputDir>
```
Jjagg: I'm good with merging this PR as is, but we should look into fixing the directories that are used. The built content should also be put in the intermediate output folder, instead of bin. I'd suggest something like the following:
- `%(IntermediateOutputPath)/mgcontent/$(MonoGamePlatform)/$(OutputFolder)` for the MGCB intermediate folder.
- `%(IntermediateOutputPath)/xnb/$(MonoGamePlatform)/$(OutputFolder)` for the XNB's.

Jjagg:
This is awesome! Great job @jnoyola!
I'll merge this tomorrow if nothing comes up :)

Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.targets
jnoyola: Fair point. I wanted to try to keep this a straight translation, with fixes separate. Also, do we ever plan to use the bin/obj values specified in the mgcb?

Jjagg:
Merging! We'll get better testing by making this public, so let's move forward.

jnoyola:
Thanks! If anyone's still following this thread, I urge you to weigh in on #6922 🙂

PR #6907, opened 2019-10-17
Make SongCollection constructor public #3352

FrancescoBonizzi:


tomspilman:
This doesn't seem to do anything useful now.

PR #6913, opened 2019-10-25
Decouple Content.Pipeline from Graphics

nkast:
Removes the depedency of GraphicsDevice from base class ContentReader and the cached IGraphicsDeviceService from ContentManager.

Readers of GraphicsResources request the GraphicsDevice directly from the ServiceProvider.

### DIFF
--- a/Build/Projects/MonoGame.Framework.definition
+++ b/Build/Projects/MonoGame.Framework.definition
@@ -420,6 +420,7 @@
     <Compile Include="Content\ContentLoadException.cs" />
     <Compile Include="Content\ContentManager.cs" />
     <Compile Include="Content\ContentReader.cs" />
+    <Compile Include="Content\ContentReaderExtensions.cs" />
     <Compile Include="Content\ContentReaders\AlphaTestEffectReader.cs" />
     <Compile Include="Content\ContentReaders\ArrayReader.cs" />
     <Compile Include="Content\ContentReaders\BasicEffectReader.cs" />
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -25,7 +25,6 @@ namespace Microsoft.Xna.Framework.Content
 
 		private string _rootDirectory = string.Empty;
 		private IServiceProvider serviceProvider;
-		private IGraphicsDeviceService graphicsDeviceService;
         private Dictionary<string, object> loadedAssets = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
 		private List<IDisposable> disposableAssets = new List<IDisposable>();
         private bool disposed;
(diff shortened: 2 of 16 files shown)


Jjagg:
We could have a public method to get the GraphicsDevice through the service on ContentManager. Just so we don't duplicate the code everywhere. That's also nice for extensions to use. Maybe also a function on ContentReader that delegates to its ContentManager so readers can just do `input.GetGraphicsDevice()`.

nkast:
All it does now now is to remove the caching of IGraphicsDeviceService from ContentManager. It's not really what I was going for since the ContentReader still dependence on the ContentReader.

Jjagg:
Oh, okay I see your point now. By using the DI container from ContentManager directly it doesn't have to know about IGraphicsDeviceService or GraphicsDevice. Do you see any benefits in having ContentManager in the shared assembly?

nkast:
One example are the [ContentReaders of FarseerPhysics](https://github.com/tainicom/Aether.Physics2D/tree/master/Physics2D/ContentReaders). The importers would triangulate an image or an .svg and export it either as Polygons - a set of raw Vector2s- or directly to a Body. Other scenarios are when one has [game world data serialized](url) from XML via the pipeline. [[1]](http://community.monogame.net/t/solved-content-load-xml-with-a-custom-class-in-the-xml/10438) [[2]](http://community.monogame.net/t/xmlserializer-wont-deserialize-vector2/10391/9) [[3]](https://stackoverflow.com/questions/36070484/monogame-xml-serialization-with-content-pipeline)

nkast:
One more thing, XNA is throwing a ContentLoadException when there is no IGraphicsDeviceService.
MonoGame is throwing InvalidOperationException.

Screenshot (70)

Jjagg:
So the point is that libraries that don't require anything from the platform-specific assembly can have custom content readers and still only depend on the shared assembly?

nkast:
Yes, assuming that the  shared assembly is used to run multiplayer game physics/updates on a head-less server. It's nice to have Vector2 and Vector2Reader on the same library and the ability to serialize simple content the same way you do on the game.

Jjagg:
Are you okay with the extension method approach?

nkast:
You mean something like 'GraphicsDevice GetGraphicsDevice(this ContentReader)' ? 
Sounds good. Where should this extension go?

Jjagg:
Hmm, I'd like for it to be public. Following the usual class name for extension methods we could add a new class `ContentReaderExtensions` that implements the `GetGraphicsDevice` method.

nkast:
I moved it in ContentExtensions.cs that has namespace 'Microsoft.Xna.Framework.Content'.
Is that ok or should I add ContentReaderExtensions.cs?

Jjagg:
Please add the new class and make it public, that way third-party readers can use it as well. The ContentExtensions class is internal.

nkast:
It's done.

Jjagg:
Great, thanks!

PR #6914, opened 2019-10-25
Use PackageReference for SharpDX

### LINKED ISSUE #6912: [WindowsDX] SharpDX missing from nupkg
I built the WindowsDX nuget package from the develop branch.
When I run a game, I get this error when MonoGame.Framework.dll tries to load SharpDX:
```Could not load file or assembly 'SharpDX, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b4dcf0f35e5521f1' or one of its dependencies. The system cannot find the file specified.```

When I look inside the nuget package (both from cake's `build.ps1` and `dotnet pack`), the SharpDX DLLs are indeed missing. They're present in the MonoGame.Framework WindowsDX build output, but don't get pulled into the nuget, presumably because of the type of `Reference`. Do we still need the local DLLs instead of using `PackageReference`s like WindowsUniversal does? If so, we need to fix this by packing them into the nupkg, right?

- develop

- Windows

- WindowsDX

### PR DESCRIPTION
**Pull request by Jjagg**
Fixes #6912.

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.WindowsDX.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.WindowsDX.csproj
@@ -71,13 +71,13 @@ This package provides you with MonoGame Framework that uses DirectX for renderin
   </PropertyGroup>
 
   <ItemGroup>
-    <Reference Include="..\ThirdParty\Dependencies\SharpDX\net40\SharpDX.dll" />
-    <Reference Include="..\ThirdParty\Dependencies\SharpDX\SharpDX.Direct2D1.dll" />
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jnoyola** [comment]
Verified this works for me.

File: MonoGame.Framework/MonoGame.Framework.WindowsDX.csproj [hunk 1]
```diff
@@ -70,14 +70,18 @@ This package provides you with MonoGame Framework that uses DirectX for renderin
     <UseWindowsForms>true</UseWindowsForms>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <SharpDXVersion>4.0.1</SharpDXVersion>
```
cra0zy: Lets not do this, I can see IDEs having trouble with this.

PR #6916, opened 2019-10-25
Use local image for NuGets icon

Jjagg:
PackageIconUrl is [deprecated](https://aka.ms/deprecateIconUrl).


Directory.Build.props
```diff
@@ -1,10 +1,14 @@
 <Project>
   <PropertyGroup>
-    <PackageIconUrl>https://pbs.twimg.com/profile_images/487954549441691649/O3KsHAsb_400x400.png</PackageIconUrl>
+    <PackageIcon>icon.png</PackageIcon>
```
cra0zy: Lets use PascalCase for filenames as all the other files in repo are that way. Also maybe move it in MonoGame.Framework directory?

Jjagg: Tools use this as well.

PR #6920, opened 2019-10-27
[CAKE] Disable WindowsDX compilation on non-Windows platforms

cra0zy:
Since we added netcoreapp3.0 as a target framework, this can no longer be compiled from non Windows operating systems.


Jjagg:
Why would it not work on Linux/Mac?

PR #6924, opened 2019-10-30
FileHelpers: Fix Uri getting confused by double slashes

flibitijibibo:
While working on a Proton title I ran into an issue with ReadExternalReference, where the file path and external reference path looked something like this:

```
Models//zolt
..\effects\instancerDeep
```

With the current function, you end up getting `Models/effects/instancerDeep`, which is wrong. Strangely enough, Uri is sensitive to empty directory names, so to resolve this you have to strip out the double slashes manually. It also has to be looped because you could get something really awful like `Something\\\\Or\\Other`, and Replace only does one pass per call.


Jjagg:
> With the current function, you end up getting `Models/effects/instancerDeep`

You probably meant `Models//effects/instancerDeep`?

Thanks Ethan! :)

PR #6930, opened 2019-11-05
Cake and template updates

mrhelmut:
Hello!

This PR continues #6879 

- Adding MGCB and 2MGFX to the cake build (not incuding Pipeline tool for now)
- Fixing the templates pack project to work and adding it to the cake build (it now produces a nuget that can be consumed by ```dotnet new -i``` to install the templates)
- Fixing some templates that were not setting the ```MonoGamePlatform```
- Making the templates to use the main repo nugets instead of @cra0zy ones for net core (assuming that we'll have to place them on the MonoGame nuget account along with the Builder #6905 package)

Once #6905 is merged, I believe that we only have to upload the nugets to have a proper pipeline

@cra0zy @Jjagg


mrhelmut:
All changes done.

mrhelmut:
I've updated this PR with a huge revamp of all the templates (and removing the obsolete ones) to make sure that they work with #6905.

This should make MonoGame 100% nuget based and fully compliant with the ```dotnet``` CLI. The installer is definitely obsolete.

I've tested 2 scenarios over all the targets:
- using Visual Studio 2017 and targeting .Net 4.5.2 (all target but DesktopGL Core and WindowsDX Core)
- using Visual Studio 2019 and targeting .Net Core 3 (DesktopGL Core and WindowsDX Core)

I have everything working out-of-the-box and building content (including .Net Core self contained single exe stuff).

Using MonoGame should be as simple as typing ```dotnet new -i MonoGame.Templates.CSharp``` and creating a new project with ```dotnet new```.


The only issue remaining is a target file being misplaced for UWP templates, but I'd like to address it in a separate PR.

Once #6905 and this are merged, we will be able to move on with cleaning up the old nugets config, the old templates, the installer, and protobuild (then comes the test and wiring things to teamcity).

@Jjagg @cra0zy

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/MonoGame.Templates.CSharp.csproj
```diff
@@ -1,26 +1,24 @@
...
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+	<PackageType>Template</PackageType>
+	<Title>MonoGame project templates</Title>
     <Description>A set of C# templates written for the MonoGame framework.</Description>
     <PackageId>MonoGame.Templates.CSharp</PackageId>
-    <PackageVersion>0.0.1</PackageVersion>
     <Authors>@MonoGameTeam</Authors>
+	
+	<TargetFrameworks>netstandard2.0;net452</TargetFrameworks>
```
Jjagg: We can just have netstandard here, can't we?

Jjagg:
Awesome! We're getting there :) 

Couple things:
- I don't think we need to keep the non-SDK style project templates around for DGL and WDX. You can target NETFX fine with the SDK style projects. 
- You've used tabs instead of spaces in the csproj files
- We shouldn't set PackageVersion for nuget in the csproj. If we do we have to keep it up to date each time we bump the MG version. The version should be overwritten in the packaging step.

mrhelmut:
I believe the non-SDK style projects can come handy for developers stuck at VS2015.

Damn those tabs, happens every time. I'll fix that.

I totally agree for the package version. I haven't figured out how to ask Cake to do that, and the team city environment variable doesn't seem to work. We definitely need a mechanism to generate the proper package version from Cake, and there should be a default version number when building from source locally so that testing out local nugets don't require to "downgrade to 1.0.0 packages" (like it is right now).

mrhelmut:
@Jjagg I've reverted the ```PackageVersion``` and removed the tabulations.

Meanwhile I figured how Cake works. To set the version when building, one must use ```build.ps1 -build-version="3.8.0.0"```.

And I can confirm that the ```BUILD_NUMBER``` environment variable needed by Team City is not applied to nugets (it's only used for the installer signature).

So I patched the Cake script to use ```BUILD_NUMBER``` everywhere if it is set.
If the ```build-version``` is set, it uses it over ```BUILD_NUMBER``` env var.

This should cover all use cases.

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/MonoGame.Templates.CSharp.csproj
```diff
@@ -1,26 +1,23 @@
...
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+    <PackageType>Template</PackageType>
+    <Title>MonoGame project templates</Title>
     <Description>A set of C# templates written for the MonoGame framework.</Description>
     <PackageId>MonoGame.Templates.CSharp</PackageId>
-    <PackageVersion>0.0.1</PackageVersion>
     <Authors>@MonoGameTeam</Authors>
+
+    <TargetFrameworks>netstandard2.0;net452</TargetFrameworks>
```
Jjagg:
```suggestion
    <TargetFramework>netstandard2.0</TargetFramework>
```

This can just target netstandard2.0.

Jjagg:
> I believe the non-SDK style projects can come handy for developers stuck at VS2015 (it requires the dotnet CLI though, so that's kind of a wacky setup, I must concede. Up to you!).

I'd like to see us move to modern tooling as much as possible. It's all free and available, so I think we should encourage people to get the newest stuff. Like you said, we already require dotnet CLI.
I guess the workflow that this breaks is building console and SDK projects all at once, because console projects require VS2015 and SDK projects require VS2017+. Do you think that's a valid workflow that we should support? I can't really speak on it because I don't know how people use MG on consoles. I would prefer to drop the old project templates.

This is looking great, thanks for helping out!

mrhelmut:
@Jjagg Don't worry about the consoles, they will need their own non-SDK projects anyway (shouldn't be much work). The on-going refactor on the main repository doesn't impact consoles as long as a C#5 policy is enforced (and hopefully this should be addressed at some point).

I removed the duplicates. :)

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Application.UWP.CoreApp.CSharp/project.json
```diff
@@ -0,0 +1,16 @@
+﻿{
+  "dependencies": {
+    "Microsoft.NETCore.UniversalWindowsPlatform": "5.2.2"
```
Jjagg: I think this file can be removed, because the csproj has a PackageReference set. The Include for this file in the csproj should also be removed.

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Application.WindowsDX.CSharp/MGNamespace.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
+    <MonoGamePlatform>Windows</MonoGamePlatform>
```
Jjagg: `MonoGamePlatform` does not necessarily have to be set in the templates, because the MonoGame.Framework package comes with a props file that sets it for you. It might be nice to quickly set up a project with a P2P reference for source debugging though, and it doesn't do any harm having this line so let's leave it in.

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Library.Shared.CSharp/.template.config/template.json
```diff
@@ -18,5 +18,5 @@
     ],
     "sourceName": "MGNamespace",
     "preferNameDirectory": "true",
-    "description": "A MonoGame shared library project."
+    "description": "A MonoGame shared library project that can use to share code and content across all targets."
```
Jjagg: Typo

```suggestion
    "description": "A MonoGame shared library project that can be used to share code and content across all targets."
```

Jjagg:
@mrhelmut Gave this one last full review and it's looking great! Just a few notes from me, and because I merged the Content.Builder PR there are merge conflicts in the CAKE build script. If you can fix the merge conflicts I'll merge this :)

ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Application.UWP.CoreApp.CSharp/project.json
mrhelmut: Good catch!

mrhelmut:
@Jjagg I'm going to apply your suggestions and squash this into another PR to rebase.

PR #6569, opened 2018-12-01
Added 𝜏 field that represents 2π

Lautenschlager-id:

PR #6937, opened 2019-11-08
Add τ constant

Lautenschlager-id:
Bringing back the idea of using τ as 2π, as we discussed a bit about it last year in [#6569](https://github.com/MonoGame/MonoGame/pull/6569)


Jjagg:
Thanks! :)

PR #6944, opened 2019-11-18
Update 4_adding_basic_code.md

jwigert:
Probably very obvious to C# developers, but as a complete newbie I was confused by the build error.


cra0zy:
I would rather replace the min/max with if checks.

jwigert:
Thanks for the feedback. I've created a new commit based on your suggestion.

PR #6949, opened 2019-11-26
Fix: a crash on OSX and CoreRT related to mishandling unmanaged function pointers

s-m-k:
Related issue and explanation: https://github.com/dotnet/corert/issues/6479#issuecomment-558433440


cra0zy:
Thanks @s-m-k, merging!

PR #6950, opened 2019-11-27
Make Cake script can build individual nuget packages.

windperson:
Make cake build script be able to build individual library, such as when you only want Android part, in PowerShell:

```
.\Protobuild.exe --generate Android
.\build.ps1 --target="BuildAndroid"
```

To build project template nuget:

```
.\build.ps1 --target="PackVSTemplates"
```

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -162,6 +162,7 @@ Task("BuildTools")
 });
 
 Task("PackVSTemplates")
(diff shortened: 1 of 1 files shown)


harry-cpp:
They already can build individual targets, you just have to use `build-target` param, so closing this.

PR #6951, opened 2019-11-27
[Android] fix GamePad.Back emulation

nkast:
The current implementation will report Buttons.Back as Pressed and immediately return to Released on the next frame. 
The state is changed repeatedly from frame to frame for as long as the user keeps the back button pressed.

this results is various problems:
-Reports multiple Pressed/Released for a single press.
-It's not possible to detect  a HOLD gesture.
-Writing crossplatform code is problematic, checking for both gamepad Buttons.Back and Keys.Back releases will result in registering multiple events. (Windows Store uses the Keys.Back which is also what Android is reporting natively when the back button is pressed. GamePad's Buttons.Back is used on Desktop but also on Android with bluetooth gamepads).


Jjagg:
Yeah, this is better. Thanks!

PR #6952, opened 2019-11-27
[Android] Fixes android APIs fallback

nkast:
http://community.monogame.net/t/choosing-the-right-android-versions/12020


Jjagg:
Thanks!

PR #6953, opened 2019-11-27
Throw InvalidContentException with error line number/position in ContentIdentity

nkast:
Uses the existing NewInvalidContentException()
method of ReflectiveSerializer.
That method takes care to report the error linenumber/position by passing it in the ContentIdenity of InvalidContentException.

This results in a consisted error message from mgcb tool.


nkast:
I removed the line number/pos from the msg as I though it would be just a repetition, 
but I noticed that the filtered view on pipeline doesn't show the line number , position.

That an issue with pipeline but for now I can change the msg to include those as it did before.

nkast:
pipeline0
pipeline1

Jjagg:
Yeah, let's keep the line info in the message for now.

nkast:
ok, the error now reads as 
"C:/Users/nkast/Desktop/testPipeline/File.spritefont(14,31): The Xml element `Size` is required, but element `` was found at line 14:31. Try changing the element order or adding missing elements."

Jjagg:
Great, thanks!

PR #6954, opened 2019-11-27
[Pipeline] Rethrow InvalidContentException for Importers instead of wraping it in a new PipelineException. 

nkast:
Similar to what we are doing for processors.

https://github.com/MonoGame/MonoGame/blob/477dbd28c5af48ca7b2c910109da4e92a5c0e7fc/MonoGame.Framework.Content.Pipeline/Builder/PipelineManager.cs#L715-L726

Currently an InvalidContentException from an importer will be wrapped in a new PipelineException with the error msg replaced by 'Importer '{0}' had unexpected failure!' while this should happed only for unknown errors/exceptions.


Jjagg:
Thanks @nkast!

nkast:
This shows the changes from this PR and the changed error message from #6953 .

pipeline0

pipeline1

PR #6955, opened 2019-11-27
Move the implementation of BoundingSphere. Intersects(BoundingSphere) in the fast 'ref' method

nkast:
This moves the implementation body in the fast 'ref' method, as is the case with most other methods that have a 'ref' variant.


Jjagg:
Oops, nice catch!

Issue #6362, opened 2018-07-04
Installing and Uninstalling MonoGame MacOS 


bigmonte:
PROBLEMS:

INSTALLATION / UNINSTALLATION (3.6)

The uninstallation command has some syntax error when its called, and doesn't removes at all the MonoGame Framework, letting this output:

orgot package 'com.monogame.pipeline' on '/'.
/usr/local/bin/monogame-uninstall: line 11: syntax error near unexpected token `fi'
/usr/local/bin/monogame-uninstall: line 11: `fi'
bash-3.2$ pwd

In the End of installation pipeline isn't in Application's folder. 
Have to build the MonoGame. I don't find any 3.7 installer yet.
Edit: to clarify, tested in dev build (6 july), and the same issue in both things

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- MacOS

#### What MonoGame platform are you using:

- MonoGame 3.7 Framework, etc (own build via VS for Mac)
- Pipeline (MonoGame 3.6) - Successfully installed in different partition (not the same)

tomspilman:
@dellis1972 @cra0zy ?

Lus1t4nUm:
Since last update that I did (High Sierra), my MacOS had some strange behaviours. Because as far I remember this didn't happen to me before that update . So had to uninstall all stuff and reinstall most all stuff.  TOday I ended up avoiding using a VM to open the Pipeline, installing the Pipeline 3.6 in a empty partition that I have in my computer. The strange behaviour that I issued this may not occour in fresh installations ( I didn't tested).

 But the reality is that should have a clean installation that makes the properly checks. http://i64.tinypic.com/o8zzhu.png This is was what I get in my successfull installation (At least creating files and don't crashing).  

I opened the "Library/Receipts" in my Mac and it sugested me a different structure, maybe that was the cause that, the App didn't came to Application folder?

Lus1t4nUm:
**Update**: The installation method is running good now (tested with development build at this date) 👍 👍  

Backing to the uninstall... I think this thing is hardcoded, but I propose updating the uninstall script for Mac. To something like this ( I didnt test this draft, but if something can fail i bet its the second if)

```
#!/bin/bash
sudo rm -rf '/Applicatons/Pipeline.app/'
# sudo pkgutil --forget com.monogame.pipeline (i do not have this)
if [ -d '/Applications/Visual Studio.app' ]
then
[...]

PR #6956, opened 2019-11-28
fix syntax error on uninstall script issue #6362

educrod:
Hi,

There are an syntax error on this script and the following code will fix it.

Best Regards,


Jjagg:
Thanks :)

Issue #6262, opened 2018-03-20
SetPosition not always working


KakCAT:
I think Mouse.SetPosition sometimes does not move the cursor. This issue is similar to other issues (like https://github.com/MonoGame/MonoGame/issues/6210 , but it's closed now)

Please note that this bug is not the same as the issue where MouseState is not updated until the next frame. It's not that the values are not updated. The cursor is not moved.

I've set up a 'working' example of the bug: https://github.com/KakCAT/mouseBugTest

Move the mouse gently left to right provoking the recentering of the mouse. After some time (it's random, usually 10-20 seconds in my rig) the exception inside lastFrameMouseWasCentered is thrown.

#### What version of MonoGame does the bug occur on:
- Tested on latest build (3.7.1429)

#### What operating system are you using:
- Windows 10

#### What MonoGame platform are you using:

- WindowsDX

svantechgit:
More test on this issue done. My previous workaround didn't work on all situation. Worked fine on my computer but not another.

The problem with SetPosition() (Mouse) is several.
1. We assume that GetCursorPos() should return same value after SetCursorPos() but it's not the case.
Instead we should think of it as two ques in/out to the OS. One that reads the position and one that updates the position (magic happens behind our control)

2. The MouseState is updated always before the main Update() is done. Even if we not using a mouse. Monogame should instead always read the position when we call Mouse.GetState().

3. The position might have changed when we call Mouse.GetState() and SetPosition() updates the internal mouse X,Y position. But should not. 

How to fix this issue is to move update when GetState is called and ignore updating X/Y position when SetPosition() is done. Make it clear that GetState and SetPostion not always follow each other if we use them fast. It would also make sense to make the Setters in MouseState public.
 
How do we then check mouse movement when centering the mouse every frame.
Read mouse position GetCursorPos()/GetState() and update the Diff X/Y from center position then
Use SetCursorPos()/SetPosition() to center the mouse again. If we do this on a regular rate the Diff X/Y work like expected. Suggested changes must be done. 
I do think this can be done for windows without breaking monogame compatibility.

PR #6958, opened 2019-12-05
Reduce chance of losing mouse events due to Mouse.SetPosition()

bartwe:
Reduce chance of losing mouse events or reading stale data due to Mouse.SetPosition's potential async delay.

It does this by waiting to read mouse state till after the Wait for the start of the framelimiter, (avoids some input lag)
Then determining the delta of mouse movement since the previous frame, then applying that delta to the requested SetPosition (this avoids losing some of the mouse movement)
This also ensures the least time is lost between GetCursorPos and SetCursorPos and the most time is between the previous SetCursorPos and the next GetCursorPos

This is not a complete fix as SetCursorPos can have a non-zero delay before having an effect on GetCursorPos which can lose some mouse input or in extreme cases cause the recenter to not happen by the next frame of the game causing phantom input as per https://github.com/MonoGame/MonoGame/issues/6262#issuecomment-417878184

image


mrhelmut:
Thanks Bart!
My only concern right now is that it breaks non-Windows platforms. I would suggest to find a way to pass the window handle to the mouse code without breaking ```PlatformSetPosition``` in the ```Mouse``` class.

bartwe:
I am wondering about the UWP/WinRT, it is not clear to me what parts they do and don't share with winforms in this context.

bartwe:
Check fail is due to 'nuget ICSharpCode.SharpZipLib' ? i haven't touched anything close to that..

mrhelmut:
Thanks for the update. Fine to me. I'll leave this to other eyes regarding ```TickWait()``` and ```TickRun```.

bartwe:
Hi, would this be more mergable without the TickWait/Run bit, i can remove that i guess, would make it less effective

mrhelmut:
Summoning @Jjagg and @cra0zy to have their thoughts.

Jjagg:
For the input lag, IMO it makes a lot more sense to 

1. Process events
2. Update
3. Wait

in that order rather than waiting after processing events. I would like to change that for ALL platforms, but it might have little side effects that can break behavior that some users rely on. So I don't think we should merge this casually and might want to keep that change for a new major version (4.0).

> Reduce chance of losing mouse events or reading stale data

By stale data you mean getting the cursor position does not immediately reflect the set cursor position? By losing events you mean mouse movement that happens during the wait in the game loop is lost because the cursor has not actually moved yet? 
We could use a message mask with `GetMessage`/`PeekMessage` to process the cursor position message right after calling SetPosition. In combination with reordering wait and update as stated above I think that should fix the issue.

bartwe:
I don't think waiting after the update makes sense seeing as MG doesn't have enough information on when vsync happens, and we want to minimize the time between sampling input and having the resulting gamestate rendered to the screen.

@Jjagg the core problem is that when SetCursorPos goes into effect can have a delay on windows,(as the mouse is a resource owned by the kernel, this delay can be more than 16ms in extreme cases) losing mouse movement that happened between the previous GetCursorPos and the moment the SetCursorPos goes into effect.
Placing them next to eachother minimizes this loss but is no guarantee, that requires switching to rawevents (mrgrak made an implementation of this i believe)

bartwe:
How can i restart this check, the failure is unrelated to the change.

tomspilman:
@bartwe - there have been changes to `develop` branch that the build server requires.  You will need to merge the latest develop into this branch to make it successfully kick off a build.

SimonDarksideJ:
@bartwe can you retest this PR with the latest develop branch and verify if the issue still exists.  Were you able to resolve @Jjagg and @mrhelmut  queries?

SimonDarksideJ:
Any update @bartwe or can this be closed now?

bartwe:
Feel free to close

PR #6959, opened 2019-12-07
DesktopGL fix: GraphicsAdapter.Description always returning empty string

s-m-k:
Now it uses `glGetString(GL_RENDERER);` to retrieve the information about the video hardware used. Especially useful for diagnosing certain hardware-specific problems.

The fix is specific to DesktopGL platforms.


mrhelmut:
Cool! This one might come handy.
Maybe instead of ```String.Empty``` in case of failure, we could put something like "Unknown OpenGL device"?
Anyway, it's a go for me.

s-m-k:
OK, if you have a reasonable idea on how to reduce the #ifs without spaghettifying everything else, then go ahead. So far every other solution either requires a) a lot of attention (= easy to make a BS mistake) in order to separate the functionality from the legacy file, or b) making changes that aren't really compliant with the "general purpose" version of the class, so not that safe to tackle either (or safe only within platform-specific parts of the code), plus that would undoubtedly render the legacy class even harder to do away with in the future.

Personally, I'd stay with the #if in this case, as it doesn't really change the original (lack of) maintainability of this file in any way, and doesn't really contribute to the overall spaghetti that much - it's at least self-contained within the file which is already planned to be removed, the "outside world" doesn't see the nasty details. Plus, it's fairly robust the way it is.

> Maybe instead of String.Empty in case of failure, we could put something like "Unknown OpenGL device"?

Good idea. Better yet, let's put an explanation of what you need to do to get it right: `Unknown adapter. Please initialize the graphics device first.`

mrhelmut:
```GraphicsAdapter.Legacy``` is no more used for DirectX platforms, so I guess it could be renamed ```GraphicsAdapter.OpenGL``` and be cleaned up of all the useless DirectX code and reduce some of the preprocessing since it would be fully OpenGL platforms.

Though it should be fine to address this in another PR if you'd like to keep this one focused on the original issue.

s-m-k:
I guess one step at a time is the way to go. I'm done with it for now anyway.

harry-cpp:
Build bot is working now, just pull/rebase from develop to let the build pass.

Jjagg:
Merging, thanks @s-m-k :)

Issue #6850, opened 2019-08-10
Extract platform-agnostic classes from MG.Framework into separate assembly


Jjagg:
This is essentially meant to restart the discussions from #2500.

I suggest we extract some foundational platform-agnostic classes into a seperate assembly called MonoGame.Core. 

The content pipeline and tools by proxy reference a platform specific assembly right now. In practice this doesn't matter because the parts of MonoGame.Framework that are used by the content pipeline are the same across all platforms (it does not use any device resources like gpu, windowing or audio system), but that's also why it does not make sense to build the CP with a reference to a platform-specific assembly. 

Mostly this restructuring serves to make things more sensible. We go from this:
```
DesktopGL    WindowsDX
    ^            ^
    | ref        | also
    |            | works
    CP -----------
```

To this:
```
      -------->  Core  <------
      |           ^          |
      |           |          |
DesktopGL    WindowsDX       CP
```

It also makes things less complicated for users that want to write a reusable library that only uses stuff in Core or a CP extension. The Core project can multitarget netstandard1.1 (for .NET Framework 4.5) and netstandard2.0. 

Additional misc stuff this improves:
- Doesn't include MonoGame.Framework runtime dependencies in Content.Pipeline and dependents.
- Shared stuff isn't tested seperately for each platform.

@tomspilman Will this work for consoles? What .NET do they target?

vpenades:
Candidate types to be moved to the core would be:

Maths types: Vector2, Vector3, Matrix, Quaternion...
Bounding types: BoundingBox, Rectangle, etc.
Curve types.
Color.
GameTime.
ReusableItemList.
Probably many types from Input namespace.

and... Interfaces!

As I stated in the past, the advantages of having a shared entry point assembly for all platforms is that it could help in simplifying a number of use cases.

For example, right now, many physics engines (Box2D, Farseer) targeting monogame and reusing monogame's maths, need to deploy multiple packages, one per monogame's target package. This is extremely inefficient, since physics are inherently platform agnostic.

Another use case is, for people like me that likes to use the MVC pattern, having the game logic in a separated assembly that is platform agnostic is a plus.

Also, any MonoGame extension capable of targeting only the Core would be future proof. Right now, even the simplest of monogame extension becomes outdated very fast as new platform specific packages are updated or introduced.

jnoyola:
[...]

PR #6961, opened 2019-12-08
Running the tests on .Net Core and Visual 2017/2019

mrhelmut:
Hello there,

I'm starting to work on making the tests to run on both ```dotnet``` and Visual Studio, and also having the solution files to work (they aren't right now, and are missing MGCB, 2MGFX and Pipeline).

I'm testing on Visual 2017+2019 and .Net Core 3.1.

**What this PR does:**
- Fixes the solution files to load ```MonoGame.Framework.Content.Pipeline.csproj``` for the test to build and run
- Adds MGCB and 2MGFX to the solutions (Pipeline doesn't have a SDK-Style csproj yet to be included)
- Fixes the tests to work (WIP, see status section)

**Issue with ```MonoGame.Framework.Content.Pipeline.csproj```:**

Right now, the ```.sln``` don't work because they are missing ```MonoGame.Framework.Content.Pipeline.csproj```, which itself is an issue because it references ```DesktopGL``` (and therefore prevents the ```WindowsDX``` tests and tools to build).

We can workaround this by adding conditions to the base framework reference in ```MonoGame.Framework.Content.Pipeline.csproj```:

```
  <ItemGroup Condition="'$(SolutionName)' != 'MonoGame.Framework.WindowsDX'">
    <ProjectReference Include="..\MonoGame.Framework\MonoGame.Framework.DesktopGL.csproj">
    </ProjectReference>
  </ItemGroup>

  <ItemGroup Condition="'$(SolutionName)' == 'MonoGame.Framework.WindowsDX'">
    <ProjectReference Include="..\MonoGame.Framework\MonoGame.Framework.WindowsDX.csproj">
    </ProjectReference>
  </ItemGroup>
```

This allows everything to build and work properly from the sln files for both ```WindowsDX``` and ```DesktopGL```.

However, this is kind of hacky and requires to build/run the tests from sln (e.g. ```dotnet test .\MonoGame.Framework.WindowsDX.sln``` or loading the sln in VS).
Building/running the tests directly (e.g. ```dotnet build .\Tests\MonoGame.Tests.WindowsDX.csproj```) won't work because of how the content pipeline references the base framework.

How do we want to handle this? @Jjagg @cra0zy @tomspilman 
Does working from solution files good enough when willing to run tests? Or should the test projects work on their own without solution files (and if so, how should we handle the problematic reference)?

**Remaining errors while running tests:**
- WindowsDX: stock effects are absent from the test folder
- WindowsDX: nvtt can't be loaded because it has the wrong bitness
- WindowsDX: SharpFont tries to load freetype dll with the wrong bitness
- WindowsDX: freeimage.dll has the wrong bitness and can't be loaded
- WindowsDX: Assimp32.dll can't be found (likely a bitness issue too)
- WindowsDX & DesktopGL: running the tests from Visual Studio or from ```dotnet test``` doesn't run the exact same set of tests (1282 vs 1298 on WindowsDX; 1141 vs 1246 on DesktopGL)
- WindowsDX: tests can't find assets when run from Visual Studio (they work fine from ```dotnet new```)

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
@@ -73,10 +73,20 @@ This package provides you with the content pipeline for Windows, Mac and Linux w
     </Reference>
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(SolutionName)' != 'MonoGame.Framework.WindowsDX'">
     <ProjectReference Include="..\MonoGame.Framework\MonoGame.Framework.DesktopGL.csproj">
       <PrivateAssets>All</PrivateAssets>
     </ProjectReference>
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(SolutionName)' == 'MonoGame.Framework.WindowsDX'">
+    <ProjectReference Include="..\MonoGame.Framework\MonoGame.Framework.WindowsDX.csproj">
+      <PrivateAssets>All</PrivateAssets>
+    </ProjectReference>
+  </ItemGroup>
+
+
+  <ItemGroup>
     <PackageReference Include="Microsoft.Win32.Registry" Version="4.5.0" />
   </ItemGroup>
 
--- a/MonoGame.Framework.DesktopGL.sln
+++ b/MonoGame.Framework.DesktopGL.sln
@@ -1,11 +1,17 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26124.0
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29519.181
 MinimumVisualStudioVersion = 15.0.26124.0
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MonoGame.Framework.DesktopGL", "MonoGame.Framework\MonoGame.Framework.DesktopGL.csproj", "{C01379FE-63B0-4DDC-AE9C-9C916CCE5F45}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MonoGame.Framework.DesktopGL", "MonoGame.Framework\MonoGame.Framework.DesktopGL.csproj", "{C01379FE-63B0-4DDC-AE9C-9C916CCE5F45}"
 EndProject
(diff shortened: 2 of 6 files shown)


Jjagg:
I'm fine with anything that's functional for now. When we get around to #6850 this can be a lot nicer.

I don't think the CP or Pipeline Tool should be a part of the solution files then, because they are not connected to the runtime libs for the platform you're building for. E.g. the Windows Pipeline Tool technically has nothing to do with the WindowsDX runtime.

mrhelmut:
We do get request from people who'd like to have the tools in the solution files to build and debug everything from Visual Studio (even though the Cake script is around, some users prefer to fire up an IDE).

I fully agree that the assemblies are too much bound to their platform. The content pipeline is the perfect example of this issue.

For the time being, I'll check what's wrong with the tests. These look like minor issues. What's puzzling me however, is why VS and ```dotnet test``` don't run the same set of tests.

mrhelmut:
Okay, so ```dotnet test``` actually doesn't guarantee the bitness and there is currently no way to force it to run in 64bit (ref https://github.com/microsoft/vstest/issues/1296). The ```-- RunConfiguration.TargetPlatform=x64``` trick doesn't work as of .Net Core 3.1 / VS2019.

So I guess we'll have to reintroduce ```AssImp32.dll``` and have ```FreeImage.Net``` to support ```AnyCPU```.

Jjagg:
I think it's fine if we do not support dotnet test for noe. We haven't the runner set up so we can use dotnet run to generate a test report. Currently the test project is also executed so running dotnet run doesn't change things very much.

mrhelmut:
Alright. From the runner perspective, the one and only issue is that the stock effects got lost in translation. Working on that.

mrhelmut:
I got everything working, but one test: ```BuildSimpleProject```

The main issue here is that ```Assets\Projects\BuildSimpleProject.csproj``` uses the old fashioned ```MonoGame.Content.Builder.targets``` to build contents.

How do we want to handle this?

Ideally this should test ```MonoGame.Content.Builder``` and not ```MonoGame.Build.Tasks```, but this implies that we build MGCB and ```MonoGame.Content.Builder``` to be able to reference them in that project.

Jjagg:
I don't think that's an issue we have to fix right away. We can ignore or remove the old test for now.

mrhelmut:
@Jjagg Fair enough.

Here we go, on .Net Core 3.1.

```dotnet run -p .\Test\MonoGame.Tests.WindowsDX.csproj```:
```
Test Run Summary
  Overall result: Warning
  Test Count: 1298, Passed: 1277, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 21
    Skipped Tests - Ignored: 21, Explicit: 0, Other: 0
```

```dotnet run -p .\Test\MonoGame.Tests.DesktopGL.csproj```:
```
Test Run Summary
  Overall result: Warning
  Test Count: 1246, Passed: 1173, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 73
    Skipped Tests - Ignored: 73, Explicit: 0, Other: 0
```

Jjagg:
Awesome!
Did you try on Windows only? I'll give this a go tonight as well before merging. Thanks! :)

mrhelmut:
Currently the tests fail on Linux and macOS because all native dependencies of the content pipeline can't be loaded (e.g. nvtt, assimp...), this is because ```dotnet``` can't read ```*.dll.config``` and would require all those dependencies to be loaded from code like SDL and OpenAL with ```FuncLoader```.

Jjagg:
> When passing a library name to a DllImport, it will for .so files on Linux or .dylib on macOS.

Yeah, let's use this for now. Gonna merge this. Thanks :)

Issue #6962, opened 2019-12-10
DesktopGL doesn't run on 32bit .Net Core / dropping 32bit support?


mrhelmut:
Hello there,

Just noticed that DesktopGL projects don't run with ```win-x86``` identifier on .Net Core (tested on 3.0 and 3.1). It results in an access violation on a native level.

As far as it goes, it seems related to SDL.

The exact same project works perfectly on .Net 4.5 while using "prefer 32bit" targeting.

This makes me think that it's an issue on the .Net Core side. Possibly linked to how the native functions are marshalled.

**The real question is**: should we drop 32bit targets once and for all?

Few arguments:
- ```MonoGame.Framework.Content.Pipeline``` is already 64bit only
- .Net Core is 64bit only on macOS and Linux
- As of November 2019, Steam players using 32bit systems represent 0.96% (and declining)
- It could simplify the repository quite a bit

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

KakCAT:
I think the last @s-m-k  messages in gitter talk about this

https://gitter.im/MonoGame/MonoGame

PR #6963, opened 2019-12-10
Fix: wrong calling convention for OpenGL crashing on Windows x86

s-m-k:
I've changed `CallingConvention.Cdecl` to `CallingConvention.Winapi`, which should fix https://github.com/MonoGame/MonoGame/issues/6962:

Docs:
> This member [Winapi] is not actually a calling convention, but instead uses the default platform calling convention. For example, on Windows the default is StdCall and on Windows CE.NET it is Cdecl.

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/OpenGL.cs
+++ b/MonoGame.Framework/Platform/Graphics/OpenGL.cs
@@ -557,289 +557,290 @@ namespace MonoGame.OpenGL
         }
 
         internal static RenderApi BoundApi = RenderApi.GL;
+        const CallingConvention callingConvention = CallingConvention.StdCall;
 
         [System.Security.SuppressUnmanagedCodeSecurity()]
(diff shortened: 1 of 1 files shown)


mrhelmut:
Great, I'll be testing that later today!

MonoGame.Framework/Platform/Graphics/OpenGL.cs
```diff
@@ -557,289 +557,290 @@ internal enum RenderApi
         }
 
         internal static RenderApi BoundApi = RenderApi.GL;
+        const CallingConvention callingConvention = CallingConvention.Winapi;
```
cra0zy: Add private before const, always specify visibility.

s-m-k: Done.

mrhelmut:
I confirm that this fixes the 32bit issue with .Net Core. Thanks!

PR #6973, opened 2019-12-26
MouseState Improvements

harry-cpp:
Stuff done:
- OptimizedMouseState memory usage by combining buttons into a single byte (from 36 bytes down to 17 bytes, or 20 if we include the extra padding that will get added)
- Fixed up the coding format of some parts

### DIFF
--- a/MonoGame.Framework/Input/MouseState.cs
+++ b/MonoGame.Framework/Input/MouseState.cs
@@ -9,14 +9,17 @@ namespace Microsoft.Xna.Framework.Input
     /// </summary>
     public struct MouseState
     {
-        int _x, _y;
-        int _scrollWheelValue;
-        ButtonState _leftButton;
-        ButtonState _rightButton;
-        ButtonState _middleButton;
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Input/MouseState.cs
```diff
@@ -137,122 +140,184 @@ public override int GetHashCode()
...
         /// </summary>
-		public ButtonState MiddleButton { 
-			get {
-				return _middleButton;
-			}
-			internal set { _middleButton = value; }			
-		}
+		public ButtonState MiddleButton
+        {
+            get
```
CommanderMark: A few of the property declarations are using tabs instead of spaces for indenting.

harry-cpp: The line endings changed so it marked the line as changed, however it was already using tabs.

To be more precise half of some files are still using tabs, maybe I should just force format the entire file.

tomspilman:
@Jjagg - This seems great to me.

PR #6980, opened 2020-01-09
Fix XNA tests project

nkast:


nkast:
Can we get the the CI server to try and build the XNA tests project?

Ideally it should run run the test, although I understand that this would be a waste of time most of the times. at the very least somebody should run and check the XNA tests manually every time a PR makes changes in the tests.

Jjagg:
Tom worked on getting XNA tests running on the build boys, but ran into some issues. I don't remember what exactly.

PR #6981, opened 2020-01-09
Update Android SDK template

nkast:
This updates the old SDK template to be functional from VS2017 and later.
-Remove reference to OpenTK. MG now has it's own GL bindings and OpenTK is no longer required.
-Add Manifest file. The manifest file is not generated automatically, VS2017 fails to build with an error and the user had to open properties, go to manifest, and click 'Generate Manifest file'.
-Include the 64-bit ABIs by default. Google Play store requires both ARM/x86 and 32/64 builds.


harry-cpp:
We plan on removing the standalone templates and using .NET Core templates in Visual Studio in the future.

Up to date Android template is at:
https://github.com/MonoGame/MonoGame/tree/develop/ProjectTemplates/DotNetTemplate/MonoGame.Templates.CSharp/content/MonoGame.Application.Android.CSharp

PR #6988, opened 2020-01-12
[Documentation] Add setting up a new project docs

harry-cpp:


Documentation/setting_up_project/setting_up_project_vscode.md
```diff
@@ -0,0 +1,137 @@
...
+mkdir Pong
+cd Pong
+```
+
+Now that we have our directory, let us create our shared game library project which will possess all the code:
+
+```sh
+dotnet new sln
+dotnet new mgshared -n Pong
+dotnet sln add Pong/Pong.csproj
```
denningk: I tried this and I got an error. The new project only has a `.shproj`, no `.csproj`

harry-cpp: Ups, that command was not not needed with shared project, fixed.

tomspilman:
Conflicted and our docs have changed since then.

PR #6992, opened 2020-01-18
Remove submodule information from Protobuild definition

harry-cpp:


harry-cpp:
One of those rare moments where I think its fine to merge my own pull request, well time to get other pull requests sorted out.

PR #6996, opened 2020-01-21
Offset should not be removed from length, length is the the length of…

### LINKED ISSUE #6995: SoundEffect.XAudio .ToDataStream() incorrectly handles offset
The offset is incorrectly subtracted from length when building a sound effect.

https://github.com/MonoGame/MonoGame/blob/a50501f5facdbb4e7fdc3d5c22677b8f83d90dc6/MonoGame.Framework/Platform/Audio/SoundEffect.XAudio.cs

code should be:

        private static DataStream ToDataStream(int offset, byte[] buffer, int length)
        {
            var data = new byte[length];
            Buffer.BlockCopy(buffer, offset, data, 0, length);

            return DataStream.Create(data, true, false);
        }

- MonoGame 3.7.1

- Windows

- WindowsDX

bartwe:
… data to use, which is separate from the source buffer length

fixes #6995

(cherry picked from commit 7fc6bb3758cf14b423987b7e99e93fef170598cb)

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.cs
@@ -164,7 +164,7 @@ namespace Microsoft.Xna.Framework.Audio
                 throw new ArgumentException("Ensure that the buffer length is non-zero.", "buffer");
 
             var blockAlign = (int)channels * 2;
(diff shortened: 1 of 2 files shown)


Jjagg:
Thanks @bartwe!

PR #6999, opened 2020-01-24
Fix load sdl2 in MacOS with dotnet core

ryancheung:
Fix load sdl2 and openal in macos.

### DIFF
--- a/MonoGame.Framework/Platform/Audio/OpenAL.cs
+++ b/MonoGame.Framework/Platform/Audio/OpenAL.cs
@@ -234,6 +234,15 @@ namespace MonoGame.OpenAL
                 if (ret == IntPtr.Zero)
                     ret = FuncLoader.LoadLibrary(Path.Combine(assemblyLocation, "runtimes", rid, "native/soft_oal.dll"));
             }
+            else if (CurrentPlatform.OS == OS.MacOSX)
+            {
+                if (ret == IntPtr.Zero)
+                    ret = FuncLoader.LoadLibrary(Path.Combine(assemblyLocation, "../../runtimes", "osx", "native/libopenal.1.dylib"));
+
+                if (ret == IntPtr.Zero)
+                    ret = FuncLoader.LoadLibrary(Path.Combine(assemblyLocation, "runtimes", "osx", "native/libopenal.1.dylib"));
+            }
+
 #elif ANDROID
             ret = FuncLoader.LoadLibrary("libopenal32.so");
 
--- a/MonoGame.Framework/Platform/SDL/SDL2.cs
+++ b/MonoGame.Framework/Platform/SDL/SDL2.cs
@@ -53,6 +53,14 @@ internal static class Sdl
             if (ret == IntPtr.Zero)
                 ret = FuncLoader.LoadLibrary(Path.Combine(assemblyLocation, "runtimes", rid, "native/SDL2.dll"));
         }
+        else if (CurrentPlatform.OS == OS.MacOSX)
+        {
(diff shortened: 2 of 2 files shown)


mrhelmut:
Could you elaborate on the motivation behind this change?

Native libraries shouldn't be in any ```runtimes``` folder once an app is built or ```dotnet publish``` with .Net Core. Native libs on macOS should be in the root of your output folder.
This change will likely not work in production environments because the runtimes folder doesn't even exist there.

I believe that your project setup is missing the deployment directives (```MonoGame.Framework.DesktopGL.targets```).

ryancheung:
My game project generates the native libraries structures like the following screenshot in macOS:

4911579863239_ pic_hd

Native libs of SDL2 and openal were located in `runtimes/osx/native` folder. 

I have to copy the libs to  `the root of output directory` or the game will crash of failed to load native sdl or openal.

BTW, it actually puts all the native libs in `runtimes` directory.

ryancheung:
The framework had already handle this case for dotnet core in windows, the code is above my edit, but not handled in macOS.

mrhelmut:
Hmmm, that's really weird, it's like .Net Core for Mac doesn't process the targets file. Puzzling.

This change shouldn't be required. We should probably investigate if the targets works on mac.

How are you referencing ```MonoGame.Framework.DesktopGL```? Did you build the nuget and referenced the nuget or did you added a reference to the ```csproj``` directly?

ryancheung:
I reference `MonoGame.Framework.DesktopGL` nuget package in the project. It should be the PackageReference.

mrhelmut:
So I confirm that there is something wrong and that the targets are not being applied.

Which version of .Net Core are you using?
Which ```TargetFramework``` does your project target?

harry-cpp:
Try referencing `MonoGame.Framework.DesktopGL.Core` nuget package.

ryancheung:
I'm actually `referencing MonoGame.Framework.DesktopGL.Core` package, sorry im not make it clear before.

ryancheung:
I'm switching my games to dotnet core.

mrhelmut:
Pretty weird. Could this be a VS for Mac issue?

What's sure is that the issue is with the targets file, and not with library loading.

harry-cpp:
@mrhelmut You have it a bit wrong there, its not about the targets file.

On .NET Core it should automatically load libraries that are in rids folder when debugging and automatically copy them when publishing, as such, our targets file is disabled on .NET Core (` Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'`).

Maybe they changed something with the newest .NET Core release regarding this.

mrhelmut:
On my .Net Core 3.1 setup (targeting ```netcoreapp3```), native libraries are outputed in the root/x86/x64 folder. There is no ```runtimes``` folder, being debug or publish. Which is why I was surprised about the proposed changes in the first place.

To me, the ```Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'``` condition has always been the intended behavior for .Net Core (and it seems to be a valid condition on Windows when building/running with .Net Core 3 & 3.1). And it works just fine in any scenario.

Beside this, it can't harm to merge these changes anyway. It's just very puzzling that the targets is applied on Windows, and not on Linux or macOS.

harry-cpp:
I've gone over the loading of SDL/OpenAL and fixed it up quite a bit in #7063, it also contains a fix for this, sorry for not getting around to merging this sooner and letting the PR get stale.

As for some good news, I've uploaded a new version of DesktopGL.Core nuget package with the needed fix.

PR #7008, opened 2020-02-03
Added joystick display name to the Joystick API

tdeeb:
This commit fetches the readable name of the joystick with [`SDL_JoystickName`](https://wiki.libsdl.org/SDL_JoystickName), which is stored in a new `DisplayName` property of `JoystickCapabilities`. This enables more parity with the GamePad API, which features a DisplayName.

Tested with two rather obscure joysticks:
- [vJoy](https://github.com/shauleiz/vJoy), a virtual controller
- Raphnet GC/N64 to USB v3.4 adapter

Both joysticks reported the correct display names.


tdeeb:
I changed them to LF; is this what they should be?

tdeeb:
@harry-cpp How's this looking now?

harry-cpp:
Looks good to me.

harry-cpp:
@tomspilman @Jjagg Since its an API change I'll leave it up to one of you for a merge. Unless you take longer than a week to notice this.

MonoGame.Framework/Platform/Input/Joystick.Default.cs
```diff
@@ -15,6 +15,7 @@ private static JoystickCapabilities PlatformGetCapabilities(int index)
             return new JoystickCapabilities()
             {
                 IsConnected = false,
+                DisplayName = "",
```
tomspilman: Use `string.Empty` here.

MonoGame.Framework/Platform/Input/Joystick.SDL.cs
```diff
@@ -57,6 +57,7 @@ private static JoystickCapabilities PlatformGetCapabilities(int index)
                 return new JoystickCapabilities
                 {
                     IsConnected = false,
+                    DisplayName = "",
```
tomspilman: Use `string.Empty` here.

MonoGame.Framework/Platform/Input/Joystick.Web.cs
```diff
@@ -37,6 +37,7 @@ private static JoystickCapabilities PlatformGetCapabilities(int index)
             return new JoystickCapabilities()
             {
                 IsConnected = connected,
+                DisplayName = "",
```
tomspilman: Use `string.Empty` here.

tomspilman:
This seems generally ok.

Note this name will never be localized... so it's usefulness may be a bit limited by that.  We shouldn't try to remedy that issue... it isn't our problem to deal with.

MonoGame.Framework/Platform/Input/Joystick.Web.cs
harry-cpp: Fun note, `string.Empty` was only useful for .NET 2 or so, its long since been fixed to be the same as using "", so right now its only about which one you prefer to use.

harry-cpp:
@tdeeb Your git seems misconfigured so the commits are not tied to your GitHub account. I don't mind merging it however I assume you'd like the commits to properly credit you. (just set git user.name and user.mail and redo the commits).

tdeeb:
That's intentional; I don't mind!

harry-cpp:
Merging, thanks @tdeeb !

PR #7014, opened 2020-02-08
Update Pipeline Tool project files

harry-cpp:
Notes:
- Pipeline Tool no longer copies all the MGCB stuff and Content Pipeline native deps as it does not need them to be operational, it just needs to find them in specified paths, this should make packaging Pipeline Tool for dotnet a bit easier
  - MGCB will no longer be built automatically when running the Pipeline Tool because of the above
- Versions are marked with Gtk and Wpf, Gtk version is connected to the DesktopGL project and can be run from any system, including Windows, Wpf version is connected to the WindowsDX project and only works on Windows 
- Keyboard shortcuts are disabled for the Gtk version

### DIFF
--- a/Build/Projects/Pipeline.definition
+++ b/Build/Projects/Pipeline.definition
@@ -27,7 +27,7 @@
     <CustomDefinitions>
       <Platform Name="Windows">TRACE;WINDOWS</Platform>
       <Platform Name="MacOS">TRACE;MONOMAC;GTK2</Platform>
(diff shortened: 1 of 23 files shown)


tomspilman:
@Jjagg - This seems fine to me.

harry-cpp:
This is still marked WIP (work in progress), I still need to go over it for the final cleanup.

harry-cpp:
Now its ready.

Jjagg:
Awesome, merging!

PR #7016, opened 2020-02-10
Fix #6732 - Support base index for instancing

### LINKED ISSUE #6732: Exposing the DrawInstancedPrimitives() instance base index

YTN0:
Per #6732 ... This change exposes the option to set the instance base index for instancing. This works perfectly on DirectX, and somewhat works on OpenGL. On OpenGL, this is still impacted by the instancing bug #6293.

Note: For OpenGL, if not setting the base instance index (which is the default behavior), it will continue using the original OpenGL API. If setting the base, this will use the OpenGL API glDrawElementsInstancedBaseInstance which is only available in OpenGL 4.2 and higher.

Also added a capabilities check to GraphicsCapabilities.cs. If the glDrawElementsInstancedBaseInstance is not available (i.e. < OpenGL 4.2), this defaults to the default behavior and calls the original method without attempting to set the base index.

FYI @Jjagg  Let me know if this looks good.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -1310,12 +1310,13 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <param name="startIndex">The index within the index buffer to start drawing from.</param>
         /// <param name="primitiveCount">The number of primitives in a single instance.</param>
         /// <param name="instanceCount">The number of instances to render.</param>
+        /// <param name="baseInstance">Used to offset the instance range indexed from the instance buffer.</param>
         /// <remarks>Note that minVertexIndex and numVertices are unused in MonoGame and will be ignored.</remarks>
         [Obsolete("Use DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount) instead. In future versions this method can be removed.")]
         public void DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int minVertexIndex,
-                                            int numVertices, int startIndex, int primitiveCount, int instanceCount)
+                                            int numVertices, int startIndex, int primitiveCount, int instanceCount, int baseInstance = 0)
         {
-            DrawInstancedPrimitives(primitiveType, baseVertex, startIndex, primitiveCount, instanceCount);
+            DrawInstancedPrimitives(primitiveType, baseVertex, startIndex, primitiveCount, instanceCount, baseInstance);
         }
 
         /// <summary>
@@ -1326,8 +1327,9 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <param name="startIndex">The index within the index buffer to start drawing from.</param>
         /// <param name="primitiveCount">The number of primitives in a single instance.</param>
         /// <param name="instanceCount">The number of instances to render.</param>
+        /// <param name="baseInstance">Used to offset the instance range indexed from the instance buffer.</param>
         /// <remarks>Draw geometry with data from multiple bound vertex streams at different frequencies.</remarks>
-        public void DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount)
+        public void DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount, int baseInstance)
         {
             if (_vertexShader == null)
                 throw new InvalidOperationException("Vertex shader must be set before calling DrawInstancedPrimitives.");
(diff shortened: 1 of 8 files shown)


MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -1310,12 +1310,13 @@ public void DrawPrimitives(PrimitiveType primitiveType, int vertexStart, int pri
         /// <param name="startIndex">The index within the index buffer to start drawing from.</param>
         /// <param name="primitiveCount">The number of primitives in a single instance.</param>
         /// <param name="instanceCount">The number of instances to render.</param>
+        /// <param name="baseInstance">Used to offset the instance range indexed from the instance buffer.</param>
         /// <remarks>Note that minVertexIndex and numVertices are unused in MonoGame and will be ignored.</remarks>
         [Obsolete("Use DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount) instead. In future versions this method can be removed.")]
         public void DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int minVertexIndex,
-                                            int numVertices, int startIndex, int primitiveCount, int instanceCount)
+                                            int numVertices, int startIndex, int primitiveCount, int instanceCount, int baseInstance = 0)
```
nkast: This one change the signature of DrawInstancedPrimitives(). It's an API change and will break existing code.
 The new functionality should be implemented as a new method.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -1326,8 +1327,9 @@ public void DrawPrimitives(PrimitiveType primitiveType, int vertexStart, int pri
         /// <param name="startIndex">The index within the index buffer to start drawing from.</param>
         /// <param name="primitiveCount">The number of primitives in a single instance.</param>
         /// <param name="instanceCount">The number of instances to render.</param>
+        /// <param name="baseInstance">Used to offset the instance range indexed from the instance buffer.</param>
         /// <remarks>Draw geometry with data from multiple bound vertex streams at different frequencies.</remarks>
-        public void DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount)
+        public void DrawInstancedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount, int instanceCount, int baseInstance = 0)
```
nkast: This one change the signature of DrawInstancedPrimitives(). It's an API change and will break existing code.
The new functionality should be implemented as a new method.

MonoGame.Framework/Graphics/GraphicsDevice.cs
YTN0: @nkast Resolved by maintaining original API and adding overloads.

MonoGame.Framework/Graphics/GraphicsDevice.cs
YTN0: @nkast Resolved by maintaining original API and adding overloads.

YTN0:
@nkast Modified to retain original APIs and added overloaded versions.

nkast:
Thanks @YTN0

Here is a recent issue that describes in more details the problems with default parameters.
https://github.com/MonoGame/MonoGame/pull/6682

Jjagg:
We're getting there @YTN0! Great work so far! 

2 more things
- I don't think we should add the obsolete overload
- when a base index is passed and it is not supported, I think we should throw an exception. I don't know what type of exception we usually throw in cases like this, but I'll check when I get to my laptop!

YTN0:
@Jjagg Made the changes you suggested. Removed the obsolete override, and also changed the code to throw a PlatformNotSupportedException if trying to use base index instancing when < OpenGL 4.2 / the API is not supported.

Jjagg:
Awesome, thanks @YTN0! Gonna merge this one :)

I'm thinking it would be good if we had a single page in the docs that listed all graphics capabilities and what platform native caps they match with for the DX and GL backend. And we should also add the graphics capabilities requirements to the xml docs for the specific methods IMO. But for now this is great!

nkast:
The pattern we use throughout the API is that `base`/`index` go first before `count`. I think we should change those new methods to follow the same pattern to be consistent before the final 3.8 release.

PR #7017, opened 2020-02-10
Added LastJoystickIndex to the Joystick API

tdeeb:
This PR adds a `JoystickCount` property to query the number of joysticks connected to the system.

There was previously no direct way of knowing how many joysticks were connected. The only way to truly find out is call `Joystick.GetCapabilities` on an increasing index until `IsConnected` is false for an index. This is not very effective nor performant because on some platforms each `GetCapabilities` call generates unavoidable garbage to obtain the GUID string of the joystick.

This PR improves things by supplying a concrete number developers can reference with minimal overhead.

@harry-cpp @Jjagg

### DIFF
--- a/MonoGame.Framework/Input/Joystick.cs
+++ b/MonoGame.Framework/Input/Joystick.cs
@@ -30,6 +30,14 @@ namespace Microsoft.Xna.Framework.Input
         }
 
         /// <summary>
+        /// Gets a value indicating how many joysticks are connected to the system.
+        /// </summary>
+        public static int JoystickCount
+        {
+            get { return PlatformJoystickCount; }
+        }
+
+        /// <summary>
         /// Gets the capabilites of the joystick.
         /// </summary>
         /// <param name="index">Index of the joystick you want to access.</param>
--- a/MonoGame.Framework/Input/Joystick.cs
+++ b/MonoGame.Framework/Input/Joystick.cs
@@ -35,9 +35,9 @@ namespace Microsoft.Xna.Framework.Input
         /// As such, this value may be larger than 0 even if only one joystick is connected.
         /// </para>
         /// </summary>
-        public static int LastJoystickIndex
+        public static int LastConnectedIndex
         {
-            get { return PlatformLastJoystickIndex; }
+            get { return PlatformLastConnectedIndex; }
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


harry-cpp:
There is one problem with the API, and that is the following situation:
- Connect 2 gamepads
- Disconnect first one
- Count will return 1, but index 0 will be disconnected and index 1 will be connected

On a related note, Joystick API is due for an upgrade.

tdeeb:
How's that?

harry-cpp:
Well, now the valid name should be LastJoystickIndex. Also you can make it O(1) by incrementing / decrementing a counter on add/remove device methods :)

PS. For web don't implement anything, the web backend that's in the main repo is useless the real thing lies in my fork.

tdeeb:
Sorry, I'm not following. The original implementation returned the joystick count, as in how many joysticks were actually connected. Now it looks like we should be going for returning the last index based on my prior commit and the comments. Is that correct?

Unless I'm missing something, incrementing/decrementing a counter would be equivalent to returning the number of joysticks connected. If we're going for index, take this example:
 - Four joysticks are connected, with indices 0-3 used. The last index is 3.
 - Joysticks at indices 0 and 2 are disconnected. The last index is still 3.
 - Joystick at index 3 is disconnected. The last valid index should be 1.

Let me know if this is how to proceed going forwards. Thanks for the note about web; I'll add a stub in place so the build bot doesn't complain.

harry-cpp:
Didn't mean literal increment/decrement:
- On Add event, if the new index is higher than the `LastJoystickIndex`, than use it
- On Remove event, if the index is `LastJoystickIndex`, recalculate LastJoystickIndex

harry-cpp:
I am fine with this, I'll leave it to someone else to merge. The only thing is `LastJoystickIndex` maybe is not the best naming convention, but I don't have a better suggestion.

Jjagg:
@tdeeb I believe it would make more sense to have the default implementation return -1 for the index to indicate no joysticks are connected (or at least none are available through the API). I think the name could perhaps be LastConnectedJoystickIndex because it's a little more obvious what joystick index it gives you. Just 'last' is more ambiguous. Though that does make it a little verbose. We could drop Joystick because it's scoped on the Joystick class so 'LastConnectedIndex'. 
Opinion, @cra0zy?

Other than that this is great :)

harry-cpp:
Agreed.

tdeeb:
@Jjagg I implemented the suggestions in the latest commit. How's it looking now?

Jjagg:
That's great, thanks @tdeeb! Merging :)

PR #7020, opened 2020-02-12
Google Stadia Support

tomspilman:
These are the initial bits needed in the public repo for Google Stadia support.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
@@ -262,4 +262,6 @@ This package provides you with the content pipeline for Windows, Mac and Linux w
     </Compile>
   </ItemGroup>
 
+  <Import Project="..\Stadia\Pipeline.targets" Condition="exists('..\Stadia\Pipeline.targets')"/>
+
 </Project>
--- a/MonoGame.Framework/Graphics/Shader/Shader.cs
+++ b/MonoGame.Framework/Graphics/Shader/Shader.cs
@@ -113,7 +113,7 @@ namespace Microsoft.Xna.Framework.Graphics
                 Attributes[a].location = reader.ReadInt16();
             }
 
-            PlatformConstruct(isVertexShader, shaderBytecode);
+            PlatformConstruct(Stage, shaderBytecode);
         }
 
         internal protected override void GraphicsDeviceResetting()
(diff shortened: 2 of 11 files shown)


MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
```diff
@@ -262,4 +262,6 @@ This package provides you with the content pipeline for Windows, Mac and Linux w
     </Compile>
   </ItemGroup>
 
+  <Import Project="..\Stadia\Pipeline.targets" Condition="exists('..\Stadia\Pipeline.targets')"/>
```
tomspilman: This seems like the right way to go for all future console support.  Lets us modify the project without it being visible to the public.

MonoGame.Framework/Graphics/Shader/Shader.cs
```diff
@@ -113,7 +113,7 @@ internal Shader(GraphicsDevice device, BinaryReader reader)
                 Attributes[a].location = reader.ReadInt16();
             }
 
-            PlatformConstruct(isVertexShader, shaderBytecode);
+            PlatformConstruct(Stage, shaderBytecode);
```
tomspilman: This was a convenient change which seemed good for future support of other shader stages.

MonoGame.Framework/Graphics/SurfaceFormat.cs
```diff
@@ -36,7 +36,7 @@ public enum SurfaceFormat
         /// <summary>
         /// DXT5. Texture format with compression. Surface dimensions must be a multiple 4.
         /// </summary>
-        Dxt5 = 6,
+        Dxt5,
```
tomspilman: No functional or numeric change to the enum values...   this was just a mistake that wasn't necessary.  Makes interop with native code more clear.

MonoGame.Framework/Utilities/PlatformInfo.cs
```diff
@@ -38,6 +38,8 @@ public static MonoGamePlatform MonoGamePlatform
                 return MonoGamePlatform.PlayStation4;
 #elif PSVITA
                 return MonoGamePlatform.PSVita;
+#elif STADIA
+                return MonoGamePlatform.Stadia;
```
tomspilman: @Jjagg @harry-cpp - May be good to clean up these #ifs and replace it with partial class stuff?

harry-cpp: This is one of those rare moments where I think #if is what we actually want.

tomspilman:
@Jjagg @harry-cpp - Please review and merge this as soon as you can.

MonoGame.Framework/Utilities/PlatformInfo.cs
Jjagg: Yeah, we chose to do it like this. Code is fairly recent.

Jjagg:
Looks good, merging!

Issue #7018, opened 2020-02-12
SDL GamePad memory leak


tdeeb:
The [documentation for SDL_GameControllerMapping](https://wiki.libsdl.org/SDL_GameControllerMapping) explicitly states that the returned string must be freed with SDL_free(). [MonoGame's implementation](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/SDL/SDL2.cs#L992) gets gamepad mappings and converts the native strings to .NET strings, but the associated `IntPtr` is never freed. This happens every time [PlatformGetCapabilities is called](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Input/GamePad.SDL.cs#L160) on a connected gamepad.

A simple fix would be to pass the IntPtr into SDL_free in GetMapping right after obtaining the .NET string but before returning it.

#### What version of MonoGame does the bug occur on:
- Develop

#### What operating system are you using:
- Windows, Linux

#### What MonoGame platform are you using:

- DesktopGL

harry-cpp:
Ha, here is a weird thing tho, other function documentations that use string don't mention `SDL_free`, and also there is no wiki page about that function.

That being said the correct fix is to move https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Utilities/InteropHelpers.cs back to the SDL class, and just call SDL_free in it.

tdeeb:
That's because it appers only a few functions return a pointer that the caller owns. Another example is `SDL_GetPrefPath`, which the documentation also notes needs to be freed. Looking at SDL's source for `SDL_GameControllerMappingForIndex` (which `SDL_GameControllerMapping` calls): 
```
pMappingString = SDL_malloc(needed);
if (!pMappingString) {
    SDL_OutOfMemory();
    return NULL;
}
SDL_snprintf(pMappingString, needed, "%s,%s,%s", pchGUID, mapping->name, mapping->mapping);
return pMappingString;
```
Memory is allocated that we need to sometimes free depending on the call. Freeing it for all calls will cause errors because then we'd be freeing memory that doesn't need to be freed, so I think for this fix it just needs to go in that one method.

harry-cpp:
Isn't it for all functions that return string type?

tdeeb:
The wiki doesn't say it's required for all the functions; I'm not sure how accurate that is for sure. I can take a closer look at the SDL source later and confirm this.

[...]

PR #7021, opened 2020-02-12
Fixed memory leak in SDL_GameControllerMapping by freeing the returned string

tdeeb:
As discussed in https://github.com/MonoGame/MonoGame/issues/7018, a memory leak occurs whenever obtaining a `GamePad`'s mapping from SDL, which is retrieved whenever `GamePad.GetCapabilities` is called.

This PR fixes the memory leak by freeing the unmanaged string SDL allocates when obtaining the mapping before returning the managed mapping string.

@harry-cpp


Jjagg:
Great, thanks @tdeeb!

Issue #7012, opened 2020-02-07
macOS not looking in "Frameworks" for .dylibs in .app bundles


tdeeb:
When loading libraries such as SDL on macOS, MonoGame looks in the directory of the Microsoft.Xna.Framework assembly. This means the .dylib files must be in the same folder as the .DLL files (or executable if using mkbundle) for MonoGame to load them, but this shouldn't be relied on for .app bundles.

According to the [official documentation](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW19), the SDL and OpenAL dylibs would be better suited in the "Frameworks" folder of an app bundle so the game uses the packaged version of the dylibs instead of a shared version that may be installed. Loading the libraries in the same folder as the assembly may work now, but there's no telling if this will change in future versions of macOS .app bundles, not to mention certification for the App Store.

The [OpenAL](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Audio/OpenAL.cs#L213) and [SDL](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/SDL/SDL2.cs#L32) library loaders can be changed to look for the files in "../Frameworks/" first if the platform is OSX then default to the current directory if not found, or vice versa. Basically, something similar to what [TitleContainer already does](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/TitleContainer.Desktop.cs#L19).

#### What version of MonoGame does the bug occur on:
- Develop

#### What operating system are you using:
- macOS

#### What MonoGame platform are you using:

- DesktopGL

harry-cpp:
Fixed in #7022

PR #7022, opened 2020-02-13
Look for native libraries in Frameworks folder for macOS .app bundles

tdeeb:
This PR makes macOS also look in the "Frameworks" folder of a .app bundle when loading native libraries. The "Frameworks" folder is the preferred location for libraries of .app bundles based on official documentation.

Related issue: https://github.com/MonoGame/MonoGame/issues/7012


harry-cpp:
Looking good, thanks, mergining!

Issue #7025, opened 2020-02-13
Add Link to Samples in Readme


chrisevans9629:
I think it would be nice to have a list of samples in the readme to make it easy for people to find examples of monogame apps.  Maybe it could look something like this?

## Monogame Samples & Demos
You can find monogame samples at the following locations
- [Official Monogame Samples](https://github.com/MonoGame/MonoGame.Samples)
- [Monogame.Extended Samples](https://github.com/craftworkgames/MonoGame.Extended/tree/develop/Source/Demos)
- [CartBlanche's Samples](https://github.com/CartBlanche/MonoGame-Samples)

### Bonus

It would also be nice to add screenshots for each of the samples/demos so people can see what they are.

### Thoughts?

I think it's important to have samples for new comers.  Maybe this would be good to have on the website as well?  Are there other collection of samples that I missed?  Overall, I want to make it easier for people to start using monogame.

If you all are okay with this, I'd be happy to make a pull request to add this.

tomspilman:
@chrisevans9629 - How about you add a new page to the docs...

![image](https://user-images.githubusercontent.com/103522/74472671-efccdd00-4e67-11ea-8e58-ce8788992cab.png)

https://github.com/MonoGame/MonoGame/tree/develop/Documentation

Then in the readme add a link in this section to the doc page...

![image](https://user-images.githubusercontent.com/103522/74472446-91076380-4e67-11ea-967d-a3c0dc6d2e99.png)

chrisevans9629:
That makes the most sense to me.  I'll make a pull request for it.

chrisevans9629:
This has been completed by #7027

PR #7027, opened 2020-02-13
Add samples to docs

chrisevans9629:
## Changes
I added samples to the website documentation so that it is easier for users to find examples.

I added links to each collection of samples and explained what they contained

Related to #7025


Jjagg:
Thanks @chrisevans9629 :)

PR #7029, opened 2020-02-15
New csproj file for Mac Pipeline Tool + minnor fixes

harry-cpp:
Ping @Jjagg


Jjagg:
We could have the eto version defined in a Directory.Build.props file, so you don't have to edit all the csproj files when upgrading.

PR #7031, opened 2020-02-15
More fixes for MonoGame iOS template (fixes #5640)

### LINKED ISSUE #5640: Include a default LaunchScreen.storyboard in the iOS template
When adding a MonoGame iOS Application to a project, and you run your game on an iPhone 7 Plus (and likely other devices), it doesn't render fullscreen. It looks like the whole thing is being scaled up inappropriately.

A good fix is to add a storyboard to serve as a launch screen. Then iOS realizes the app supports the larger resolutions, and displays you in fullscreen as you likely intend.

It'd be great if MonoGame included one by default, or at least some default such that iOS will render the game fullscreen and the device's native resolution.

harry-cpp:


Jjagg:
This looks good. I can't test this easily, so 'll trust that this works as intended.

PR #7040, opened 2020-02-25
Protobuild Removal

harry-cpp:
For now we will need 4 steps in the build system:

Windows build:
```ps1
./build.ps1 -ScriptArgs '-build-version="1.001.01"', '-build-target="Pack"'
```

Windows test:
```ps1
./build.ps1 -ScriptArgs '-build-version="1.001.01"', '-build-target="Test"'
```

Mac build:
```sh
./build.sh --build-number=1.001.01 --build-target=Pack
```

Mac test:
```sh
./build.sh --build-version=1.001.01 --build-target=Test
```

Extra notes:
- Version can be specified using `BUILD_NUMBER` env variable instead of `build-version`
- We can install `CAKE` dotnet tool on the build server instead of calling the wrapper scripts (dotnet tool install --global Cake.Tool --version 0.37.0)

Post CAKE stuff:
- Artifact publishing
- MGCB Editor dotnet tool
- Installer that just calls few dotnet instal commands?
- VSIX template automatic upload
- VS Mac template automatic upload
- Visual Studio templates using dotnet templates

PR #7041, opened 2020-02-26
TeamCity Settings

tomspilman:
This is the initial commit of current TeamCity build settings.

This allows us to version the TeamCity build script changes along with the code.

Note the TeamCity server has read-only access to the MG repo.  This disallows us from using the two-way synchronization feature to let us make changes from the TC UI.  However it is a bit safer IMO that we only manually commit these when we need changes.

Next step is to customize these the support the Cake via #7040.


tomspilman:
Woot... this is working now!

PR #7047, opened 2020-03-02
Fix initialization crash on iOS GLES 2.0 devices

dtaylorus:
Fall back to GLES 2.0 after first attempting to initialize GLES 3.0 rather than assuming the GLES version based on the iOS version as per the [Apple developer documentation](https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/WorkingwithOpenGLESContexts/WorkingwithOpenGLESContexts.html) (see Listing 2-1).

Tested on an iPad 4 running iOS 9.3.5 (GLES 2.0) and an iPad 6 running iOS 13.3.1 (GLES 3.0).

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/OpenGL.iOS.cs
+++ b/MonoGame.Framework/Platform/Graphics/OpenGL.iOS.cs
@@ -35,12 +35,15 @@ namespace MonoGame.OpenGL
     {
         public GraphicsContext ()
         {
-            var ios7OrAbove = UIKit.UIDevice.CurrentDevice.CheckSystemVersion(7, 0);
-
-            if (ios7OrAbove)
+            try
+            {
                 Context = new EAGLContext(EAGLRenderingAPI.OpenGLES3);
-            else
+            }
+            catch
+            {
+                // Fall back to GLES 2.0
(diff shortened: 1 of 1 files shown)


dtaylorus:
The issue fixed by this change is described in #7051

ryancheung:
That is weird,. Apple said iOS 7 or above support gles 3, but actually it's not true?

dtaylorus:
If you check out Apple's [Device Compatibility Reference](https://developer.apple.com/library/archive/documentation/DeviceInformation/Reference/iOSDeviceCompatibility/DeviceCompatibilityMatrix/DeviceCompatibilityMatrix.html), you can see that some devices like the iPhone 5 and the iPad 3 don't support GLES 3, however, they support up to iOS 9.

I also have a physical iPad 3 device running iOS 9 that I test on and that's how I found the issue. It definitely doesn't support GLES 3 but works fine with my fix to fall back to GLES 2.

harry-cpp:
This seems fine to me, was gonna suggest avoiding try / catch, but there really does not seem to be a way to detect if the device supports GLES3 from C#, unless we want to manually invoke the create method or hardcode GLES2 devices, which I would rather not do, so merging!

Thanks @dtaylorus!

PR #7052, opened 2020-03-04
Added missing STBSHARP_INTERNAL definitions

rds1983:
STBSHARP_INTERNAL definition makes StbSharp classes(which are included in the MG as source code) internal. 
Lack of this definition keeps them public.


Jjagg:
Thanks!

PR #7069, opened 2020-03-24
Enable executable generation for macOS by default

harry-cpp:
UseAppHost is enabled by default on Windows / Linux, and disabled by default on macOS.


mrhelmut:
Can this be a .Net Core issue? UseAppHost is supposedly defaulting to true since 3.0.

Jjagg:
> UseAppHost is supposedly defaulting to true since 3.0.

This link says otherwise: https://docs.microsoft.com/en-us/dotnet/core/install/macos-notarization-issues#apphost-is-disabled-by-default

harry-cpp:
Before anyone asks, Pipeline Tool generates a .app file, it does not need this enabled.

mrhelmut:
> > UseAppHost is supposedly defaulting to true since 3.0.
> 
> This link says otherwise: https://docs.microsoft.com/en-us/dotnet/core/install/macos-notarization-issues#apphost-is-disabled-by-default

Haha, that's so misleading with this not pointing to it: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#useapphost

PR #7080, opened 2020-03-29
Update SpriteFont.cs

willmotil:
removed redundant lines 
finalLineHeight is pre-calculated and doesn't need to be set inside the loop.


willmotil:
Don't merge this needs to be tested, didn't see the line at the end, not sure why that is there.

PR #7082, opened 2020-03-30
Update SpriteFont.cs

willmotil:
Removed inconsistency's with drawstring. The lines removed account for height cropping that spritebatch doesn't account for which would result in, a accumulating error between what is drawn and what measure string would report for the total height of the measured text. 
Im doubtful if it is correct to second guess the line height returned by a font if spritebatch did account for it.

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteFont.cs
+++ b/MonoGame.Framework/Graphics/SpriteFont.cs
@@ -222,8 +222,6 @@ namespace Microsoft.Xna.Framework.Graphics
 
                 if (c == '\n')
                 {
-                    finalLineHeight = LineSpacing;
-
                     offset.X = 0;
                     offset.Y += LineSpacing;
                     firstGlyphOfLine = true;
@@ -251,9 +249,6 @@ namespace Microsoft.Xna.Framework.Graphics
                     width = proposedWidth;
 
                 offset.X += pCurrentGlyph->RightSideBearing;
-
-                if (pCurrentGlyph->Cropping.Height > finalLineHeight)
-                    finalLineHeight = pCurrentGlyph->Cropping.Height;
             }
 
             size.X = width;


willmotil:
The failure seems to link back to SpriteEffects.Vertical which measure string has no internal or parameter relation to SpriteEffects which then cannot influence the result. Kinda makes sense the test is probably applied to both and the pull is precisely because measure string is accounting for something drawstring does not in either of is main overloads.

Jjagg:
So I think the intention of the current implementation is that for the height, linespacing is returned unless glyphs in the last line of text are bigger, in which case the bottom of the lowest glyph is returned. However without digging very deep it looks to me like the code does not take the Y offset of the glyph into account.

It seems like you don't think that should be MeasureStrings intention in the first place. With your change MeasureString would always return `LineSpacing * numberOfLines`. You can easily calculate that yourself because `LineSpacing` is exposed by `SpriteFont`. I don't know what XNA does exactly, but apparently (from reading a random stackoverflow comment) it didn't do a tight crop measurement of glyphs, instead returning at least the `LineSpacing` so I don't know if we're wrong there. There's some more discussion here: #3454.

This is also not behavior we can just change without changing glyph layout for existing games.
MeasureString behavior has been like this for over 8 years: 9a7c9c8992a48efd46db16d586d7c094bedede5b.

mrhelmut:
As far as I know, XNA always returns ```LineSpacing``` for the height of any given character in the font (like most font renderers do). Otherwise line spacing would vary depending on each string, which is something you'd like to avoid when rendering text consistently and in a typographically correct way.

Also, it should be noted that a glyph crop is just an optimization to fit them in a texture.
Font characters may have measurements that include actual margins because that's how those characters should typographically behave and be measured.

I don't think there is something wrong with the YOffset either. It is normal that some characters may go beyond their font line height while not impacting the actual line height. It sounds counter-intuitive, but that's how typography may work.

In my book, ```SpriteFont``` is correct.

If you need to get the "actual drawing surface size", I guess that ```SpriteFont``` is lacking something like ```Rectangle MeasureActualDrawingSurface(string text, Vector2 position)```.

The BMFont documention is a simple a good entry point into understanding how fonts work and typography: https://www.angelcode.com/products/bmfont/doc/render_text.html

Jjagg:
@mrhelmut If you look at the current implementation it does not always return `LineSpacing` for the height. I'm aware of how `LineSpacing` is used (correctly) to render text and how it is not affected by rendered glyphs. If you look at the implementation we have, it also does not take glyph size into account, except for the last line. I think the intention here is to return a bounding box that fully encapsulates the measured text, even if some bottom glyphs extend downward farther than expected (into the next line of text if there would be one). I'd argue that's useful to position text (e.g. in a box in UI).

willmotil:
The problem is the measurement between spriteFont measure string and spriteBatch draw string is inconsistent but only on this condition were the cropping below the line height might occur.

Which as mr helmut said would probably be wrong and i thought so too.

The effect of this would mean the measurement of measure string would not match the visual output of spritebatch when the condition occurred on the lines proposed for removal.

This as jjag said this is breaking but consider it is breaking only on that condition and then in that case already broken compared to the visual output.

Aka one is right and one is wrong on that condition.
But they cant both be right because draw string would ignore that extra height were measure string is taking it into account. So this is actually broken only in (just that case) otherwise those lines don't do anything in normal operation which is probably 99% of the time.

(_I actually tried to find a font a font were this condition occured and after the first 10 windows fonts i decided that this is probably something that occurs if at all for special characters. I mean if you guys know of font were this happens it would be better to test it but as i said i couldn't find one._)

**Pick your poison** 

1) consider it acceptable (_it could potentially be a tough bug to track down later on and potentially someone could decide to make the fix to spritebatch draw string instead. It would be proven to be broken in that case as well._)
2) Or change spritebatch to match measure string (That is visually breaking when it happens)
3) Or change measure string (which is inconsistent to what is visually shown when it happens)

Jjagg:
I think at this point backwards compatibility is more important than returning the most expected thing. if people upgrade and their text is slightly off I think we should avoid unless there's a very good reason for it. But we should better document `MeasureString` so people can use LineSpacing if that's what they need. 

So yes, I think this PR can be closed and we should not change this behavior. Happy to hear other opinions/arguments though. 

cc @harry-cpp @tomspilman

PR #7085, opened 2020-03-30
Discovery of MGCB language agnostic.

### LINKED ISSUE #7084: MonoGame.Content.Builder.Task works on english systems only
I have integrated MonoGame.Content.Builder.Task into my project in order to have my Content built by MGCB during build. On my german system this throws an exception during build:

`Der Ausdruck ""Paket-ID      Version      Befehle      Manifest".Substring(-1, 15)" kann nicht ausgewertet werden. StartIndex darf nicht kleiner als Null sein.` referencing the file `C:\Users\matth\.nuget\packages\monogame.content.builder.task\3.8.0.1179-develop\build\MonoGame.Content.Builder.Task.targets`

Diving into the targets file I discovered that the output of `dotnet tool list -local` is parsed to discover the installed version of MGCB. The problem here is that the parsing depends on the english column names `Package Id`, `Version`, `Commands` and `Manifest`while the columns are named `Paket-ID`, `Version`, `Befehle` and `Manifest` on my german system.

I tried to mitigate the problem by changing the culture for the current thread, but that does not effect the command's output.

Instead of depending on the exact string one could instead use the column position, for e.g. change

```
    <!-- Get header indices, because the Manifest path may contain spaces or semicolons, which means we cannot simply split. -->
    <PropertyGroup>
      <DotnetToolHeader>$(ConsoleOutput.Split(';')[0])</DotnetToolHeader>

      <DotnetToolPackageIdIndex>$(DotnetToolHeader.IndexOf('Package Id'))</DotnetToolPackageIdIndex>
      <DotnetToolVersionIndex>$(DotnetToolHeader.IndexOf('Version'))</DotnetToolVersionIndex>
      <DotnetToolCommandsIndex>$(DotnetToolHeader.IndexOf('Commands'))</DotnetToolCommandsIndex>
      <DotnetToolManifestIndex>$(DotnetToolHeader.IndexOf('Manifest'))</DotnetToolManifestIndex>

      <DotnetToolPackageIdLength>$([MSBuild]::Subtract($(DotnetToolVersionIndex), $(DotnetToolPackageIdIndex)))</DotnetToolPackageIdLength>
      <DotnetToolVersionLength>$([MSBuild]::Subtract($(DotnetToolCommandsIndex), $(DotnetToolVersionIndex)))</DotnetToolVersionLength>
      <DotnetToolCommandsLength>$([MSBuild]::Subtract($(DotnetToolManifestIndex), $(DotnetToolCommandsIndex)))</DotnetToolCommandsLength>
    </PropertyGroup>
```

to

```
    <!-- Get header indices, because the Manifest path may contain spaces or semicolons, which means we cannot simply split. -->
    <PropertyGroup>
      <DotnetToolHeader>$(ConsoleOutput.Split(';')[0])</DotnetToolHeader>
      <DotnetToolHeaderPackageIDColumnName>$(DotnetToolHeader.Split(' ', StringSplitOptions.RemoveEmptyEntries)[0])</DotnetToolHeaderPackageIDColumnName>
      <DotnetToolHeaderVersionColumnName>$(DotnetToolHeader.Split(' ', StringSplitOptions.RemoveEmptyEntries)[1])</DotnetToolHeaderVersionColumnName>
      <DotnetToolHeaderCommandsColumnName>$(DotnetToolHeader.Split(' ', StringSplitOptions.RemoveEmptyEntries)[2])</DotnetToolHeaderCommandsColumnName>
      <DotnetToolHeaderManifestColumnName>$(DotnetToolHeader.Split(' ', StringSplitOptions.RemoveEmptyEntries)[3])</DotnetToolHeaderManifestColumnName>

      <DotnetToolPackageIdIndex>$(DotnetToolHeader.IndexOf($(DotnetToolHeaderPackageIDColumnName)))</DotnetToolPackageIdIndex>
      <DotnetToolVersionIndex>$(DotnetToolHeader.IndexOf($(DotnetToolHeaderVersionColumnName)))</DotnetToolVersionIndex>
      <DotnetToolCommandsIndex>$(DotnetToolHeader.IndexOf($(DotnetToolHeaderCommandsColumnName)))</DotnetToolCommandsIndex>
      <DotnetToolManifestIndex>$(DotnetToolHeader.IndexOf($(DotnetToolHeaderManifestColumnName)))</DotnetToolManifestIndex>

      <DotnetToolPackageIdLength>$([MSBuild]::Subtract($(DotnetToolVersionIndex), $(DotnetToolPackageIdIndex)))</DotnetToolPackageIdLength>
      <DotnetToolVersionLength>$([MSBuild]::Subtract($(DotnetToolCommandsIndex), $(DotnetToolVersionIndex)))</DotnetToolVersionLength>
      <DotnetToolCommandsLength>$([MSBuild]::Subtract($(DotnetToolManifestIndex), $(DotnetToolCommandsIndex)))</DotnetToolCommandsLength>
    </PropertyGroup>
```

That will make the parsing language agnostic. Just tested it ok on my german system.

- MonoGame 3.8.0.1179

- Windows

- WindowsDX

MattiKloepper:
Fixes #7084 

Changed parsing of `dotnet tool list -local` to depend on column order instead off on column titles. Column titles are language specific, so that the code threw on non-english systems.
Tested new version on english and german systems.

### DIFF
--- a/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.targets
+++ b/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.targets
@@ -61,10 +61,10 @@
          is formatted as a `PrintableTable` (https://github.com/dotnet/cli/blob/master/src/dotnet/PrintableTable.cs).
          `PrintableTable` defines six consecutive spaces as its column delimiter (public const string ColumnDelimiter = "      ";)  -->
     <PropertyGroup>
-      <DotnetToolHeaderPackageIDColumnName>$(ConsoleOutput.Split("      ")[0])</DotnetToolHeaderPackageIDColumnName>
-      <DotnetToolHeaderVersionColumnName>$(ConsoleOutput.Split("      ")[1])</DotnetToolHeaderVersionColumnName>
-      <DotnetToolHeaderCommandsColumnName>$(ConsoleOutput.Split("      ")[2])</DotnetToolHeaderCommandsColumnName>
-      <DotnetToolHeaderManifestColumnName>$(ConsoleOutput.Split("      ")[3])</DotnetToolHeaderManifestColumnName>
+      <DotnetToolHeaderPackageIDColumnName>$(ConsoleOutput.Split("      ", StringSplitOptions.RemoveEmptyEntries)[0])</DotnetToolHeaderPackageIDColumnName>
+      <DotnetToolHeaderVersionColumnName>$(ConsoleOutput.Split("      ", StringSplitOptions.RemoveEmptyEntries)[1])</DotnetToolHeaderVersionColumnName>
+      <DotnetToolHeaderCommandsColumnName>$(ConsoleOutput.Split("      ", StringSplitOptions.RemoveEmptyEntries)[2])</DotnetToolHeaderCommandsColumnName>
+      <DotnetToolHeaderManifestColumnName>$(ConsoleOutput.Split("      ", StringSplitOptions.RemoveEmptyEntries)[3])</DotnetToolHeaderManifestColumnName>
 
       <DotnetToolPackageIdIndex>$(DotnetToolHeader.IndexOf($(DotnetToolHeaderPackageIDColumnName)))</DotnetToolPackageIdIndex>
       <DotnetToolVersionIndex>$(DotnetToolHeader.IndexOf($(DotnetToolHeaderVersionColumnName)))</DotnetToolVersionIndex>
--- a/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.targets
+++ b/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.targets
@@ -59,11 +59,15 @@
     <!-- Get header indices, because the Manifest path may contain spaces or semicolons, which means we cannot simply split. -->
     <PropertyGroup>
       <DotnetToolHeader>$(ConsoleOutput.Split(';')[0])</DotnetToolHeader>
+      <DotnetToolHeaderPackageIDColumnName>$(DotnetToolHeader.Split(' ', StringSplitOptions.RemoveEmptyEntries)[0])</DotnetToolHeaderPackageIDColumnName>
+      <DotnetToolHeaderVersionColumnName>$(DotnetToolHeader.Split(' ', StringSplitOptions.RemoveEmptyEntries)[1])</DotnetToolHeaderVersionColumnName>
+      <DotnetToolHeaderCommandsColumnName>$(DotnetToolHeader.Split(' ', StringSplitOptions.RemoveEmptyEntries)[2])</DotnetToolHeaderCommandsColumnName>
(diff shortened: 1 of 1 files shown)


nkast:
The comment line above says that you can't use split(), because the output might contain extra spaces.

you can try setting StdOutEncoding to something like ANSI in the EXEC  command.
https://docs.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2019
https://books.google.gr/books?id=WKNCAwAAQBAJ&pg=PT370&lpg=PT370&dq=targets+Exec+StdOutEncoding&source=bl&ots=44pJhPB8LT&sig=ACfU3U2jOiMO9nzwQHBhGsR-XJ-XPf9eyw&hl=en&sa=X&redir_esc=y#v=onepage&q=targets%20Exec%20StdOutEncoding&f=false

MattiKloepper:
You are right, my code will break for column names with spaces :-/
I will update the PR to use the `Format-List -Property *` command. That should give output that is much safer to parse.

MattiKloepper:
Formatting the output with `Format-List` does not work. Digged up the official column delimiter in the sources and use that for parsing now.

Jjagg:
I think it would be nice to have a more robust way of handling this, the 6 spaces seem like an implementation detail that shouldn't be depended on. It would be better if we could parse the .config/dotnet-tools.json file instead.
But for now once this is done I'll merge it, it's better than what we have currently. Thanks for your effort on this @MattiKloepper, feel free to give me a heads up when you think it's ready :)

MattiKloepper:
@Jjagg Should be all fixed now. Tests just finished. Local testing in my project is ok, too.

If someone has a sleek idea on how to parse the json from within the .targets file I'll be happy to give it a try.

Jjagg:
I think it'd be a lot easier to program the task [in C#](https://docs.microsoft.com/en-us/visualstudio/msbuild/task-writing?view=vs-2019) than to do the whole thing in MSBuild.

MattiKloepper:
Never tried that. But I might have a read through the relevant documentation tonight.

Jjagg:
That's great @MattiKloepper, thanks!

Jjagg:
@MattiKloepper Some more info on manifest discovery.
The way tool manifest discovery works is it checks the current folder for a `./.config/dotnet-tools.json` file, then a `./dotnet-tools.json` file and then every parent folder for those files. It stops discovery when it finds a manifest, so it only picks up tools installed in the first manifest it finds.

For instance on Windows from my `D:\Projects\` folder the search paths are the following:
```
D:\Projects\.config\dotnet-tools.json
D:\Projects\dotnet-tools.json
D:\.config\dotnet-tools.json
D:\dotnet-tools.json
```

You can have it list search paths by executing `dotnet -d tool restore` from a folder where it will not find any tools.

MatthiasKloepperQC:
> I think it'd be a lot easier to program the task in C# than to do the whole thing in MSBuild.

@Jjagg Now I finally get to have a look at this again, after private things kept me from programming for quite a while. Two questions for this:
- Do you have any preference on where to store the source / assembly for that custom build task?
- Is there any preferred way to parse JSON within the Monogame project? I'm quite reluctant to add any additional dependencies (e.g. Newtonsoft) but would rather try to set this up with System.Text.Json.

harry-cpp:
MGCB has been bundled with MGCB Task so we no longer need to manage the dotnet tool.

MatthiasKloepperQC:
That even better :-)

PR #7086, opened 2020-04-02
Extract MG.Framework.Core assembly

### LINKED ISSUE #6850: Extract platform-agnostic classes from MG.Framework into separate assembly

Jjagg:
Resolves #6850. 

In short this moves a bunch of classes that are shared to a separate assembly. That means the content pipeline assembly and some user libraries can depend on the core assembly instead of the full runtime and no longer have to run on a different runtime than they were compiled against.

WIP because graphics state classes have to be changed to only hold graphics state and not the implementation for setting them.


mrhelmut:
Awesome!

My personal preference would be to have the Core assembly being shipped with each target nugets, or better (from my personal perspective) having the Core source being a shared project so that targets are standalone from the Core assembly nuget.
My only concerns here are MonoGame's footprint (self-contained .Net Core apps are already quite a hell when it comes to files footprint) and dependencies (I really like that MG is a monolithic assembly).

I don't know if this is a shared standpoint, so just checking if I'm the weirdo here. :)
Whatever works is fine.

Jjagg:
> My personal preference would be to have the Core assembly being shipped with each target nugets, or better (from my personal perspective) having the Core source being a shared project so that targets are standalone from the Core assembly nuget.

A shared project does not offer the same benefits as a seperate assembly. While it does let us remove the Content Pipeline -> runtime dependency, it does not help users do the same. User libraries that only require the Core classes either need to either
- build from source to reference the shared project - and get access to all internal classes; shared projects are not intended for public consumption
- reference a platform runtime - which we don't want and that was part of the motivation to split off a Core assembly in the first place
- reference the CP assembly - which only makes sense if they actually require the CP so this is not applicable for most libraries.

> My only concerns here are MonoGame's footprint (self-contained .Net Core apps are already quite a hell when it comes to files footprint) and dependencies (I really like that MG is a monolithic assembly).

I think neither of those arguments are really a practical concern, but feel free to elaborate if you disagree. Splitting off the assembly does have tangible benefits and I don't think we can achieve those in any other way.

mrhelmut:
Ah, I didn't mean to not distribute the Core assembly independently, I do see the advantage here.
I just meant to keep the targets monolithic to avoid cascading nuget deps.

Jjagg:
Yeah, I got it after reading your reply in #7089.

vpenades:
Right now, it is nearly impossible to improve monogame's ecosystem because any third party library developer extending monogame needs to deploy a nuget package for every monogame platform flavour. This is not friendly at all and I believe it is greatly hurting monogame's expansion.

I believe it is absolutely critical for the future of Monogame, to improve on this aspect, and allow "one-nuget-for-all-platforms" library extensions.

Certainly, it is currently limited to essentially math libraries, so at least physics libraries will be able to deploy a single nuget package, instead of one for every target platform.

But I do hope that in the future, to find ways so it can be applied to other aspects of monogame, like graphics, audio, etc.

mrhelmut:
That's the whole idea of making the Core of MonoGame platform agnostic. This and referencing it in tools.

More work will be done later to further improve on that and maybe removing the need to per-platform C# assemblies (tomspilman has been busy working on this but that's likely for a future iteration of MG).

vpenades:
I've seen the core library includes Microsoft.NETFramework.ReferenceAssemblies ... I guess this is neccesary for some reason but, it's also required for net standard?

I mean, maybe adding that package could be conditional, so it is only reference when using net452.

Jjagg:
It's there so you can simply do `dotnet build` with only the .NET Core SDK installed. It's not necessary for the netstandard2.0 build. Since it's a build-time dependency I didn't really care about conditionally including it.

Jjagg:
Hmm, not sure why UWP build is failing. I upgraded the minimum target version for the SDK to the Falls Creator update which should support .NET Standard 2.0. Things build locally for me. Maybe the build bot does not have the right SDK installed?

```
error NU1201:
Project MonoGame.Framework.Core is not compatible with uap10.0 (UAP,Version=v10.0).
Project MonoGame.Framework.Core supports:
  - net452 (.NETFramework,Version=v4.5.2)
  - netstandard2.0 (.NETStandard,Version=v2.0)
```

mrhelmut:
Targeting uap with netstandard needs the VS2017 15.4 toolchain (or more recent), with the WindowsSDK payload. Dunno what's on the build bot, but just leaving this here for reference.

Jjagg:
Yeah it has those things. I think this is a MSBuild.Sdk.Extras issue :/ 
https://github.com/novotnyllc/MSBuildSdkExtras/issues/200

mrhelmut:
Ah :/
Not quite convenient workaround: adding uap target to the Core assembly?

vpenades:
@Jjagg regarding UAP, maybe the solution is to switch from NetStandard2.0 to NetStandard1.5

Given that most of the core in core is Net452 compliant, NetStandard1.5 might probably cover UAP

Jjagg:
Fixed the UWP issue by specifying the target in the TFM. MSBuild.Sdk.Extras could probably handle this.

At least some of the test failures (maybe all) are happening because xnb files specify the fully qualified type name of generic types and the assembly of a bunch of types just changed. E.g. Rectangle is now in `MonoGame.Framework.Core` assembly instead of `MonoGame.Framework` assembly. I'm not sure there's another way to handle this issue than to just try both assembly names, but I'll look into it.

tomspilman:
@Jjagg - First thing that concerns me....

https://github.com/Jjagg/MonoGame/commits/core-ass/MonoGame.Framework.Core/Point.cs

Not sure why but it seems like we're losing all commit history on these moved files.  We really should figure out why this is happening and make sure that doesn't happen in the final commit.  It makes history very hard to track down later.

MonoGame.Framework.Content.Pipeline/Effect/GraphicsState/SamplerStateData.cs
```diff
@@ -0,0 +1,24 @@
...
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
+
+namespace MonoGame.Effect
+{
+    public class SamplerStateData
```
tomspilman, line 11: Why is this one `public`?

tomspilman:
@Jjagg - So i get your idea now... just have a duplicate internal class just used for data.  So we have for example `RasterizerStateData` in the Pipeline assembly and `RasterizerState` in the Framework assembly.

That does solve our internal problem, but doesn't work for others making pipeline extensions.  So for example someone already has a pipeline extension that uses `RasterizerState` they will have continue to use it... worse they have to link against a platform specific version of Framework.

Ideally we could move `RasterizerState` to Core and make it contain no platform specific code in it, but doesn't seem like that is possible.  While we could hide the  The dependency on `GraphicsResource` and then `GraphicsDevice` seem like things we cannot work around.

So maybe the only other thing we could do is move `RasterizerStateData` to Core and make it public.  This makes it official and lets people use it instead of `RasterizerState`.  We could even let you construct a `RasterizerState` from a `RasterizerStateData`.  The down side... we have two classes to keep in sync to each other, duplicate documentation of members, etc.

Welcome any other ideas here.  That said we could try to not solve this and just keep the `RasterizerStateData` internal for now.

MonoGame.Framework.Content.Pipeline/Utilities/RuntimeTypeStrings.cs
```diff
@@ -0,0 +1,20 @@
+// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+
+namespace MonoGame.Framework.Content.Pipeline
+{
+    public static class RuntimeTypeStrings
```
tomspilman, line 9: It may be useful for this to be `public` for pipeline extensions to be written against.

If so we should probably also document it.

MonoGame.Framework.DesktopGL.sln
```diff
@@ -41,6 +43,18 @@ Global
...
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Debug|x64.Build.0 = Debug|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Debug|x86.Build.0 = Debug|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Release|Any CPU.Build.0 = Release|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Release|x64.ActiveCfg = Release|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Release|x64.Build.0 = Release|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Release|x86.ActiveCfg = Release|Any CPU
+		{462504E8-562E-4A21-A34C-8990A10681E7}.Release|x86.Build.0 = Release|Any CPU
```
tomspilman, line 57: @Jjagg @harry-cpp 

Why do we have x86 and x64 targets at all?  Shouldn't we just have Any CPU?

tomspilman:
@Jjagg @harry-cpp @mrhelmut -  Going to bring this up one last time... promise.

Why is [MonoGame.Framework.Portable](https://www.nuget.org/packages/MonoGame.Framework.Portable) not a better solution to all this?

Nothing has to move around.  We keep a single monolith assembly so people only need one dependency.  Everything in MG is now platform independent... you can write a game library that uses `VertexBuffer` and that library can be used on any platform.

I see nothing but positives for `MonoGame.Framework.Portable` and nothing but negatives to `MonoGame.Framework.Core`.

Why is it that we cannot continue with that direction?  Why is this better?

Jjagg:
@tomspilman The Core assembly and the portable assembly serve different purposes. I actually believe we should have both.

- The core assembly is a library that can be referenced by projects that do not require the full runtime with platform specific bits. Most importantly the Content Pipeline assembly.
- The portable assembly is for libraries that need the full runtime but want to target more than one platform.

While the portable assembly can be used in the first case it does not make a lot of sense. 
It still requires one of the MG platform implementation runtime libs to run the code. What platform should they pick? The code should run on any of them regardless, just arbitrarily pick one?
Unless the consuming lib pulls some tricks, referencing a runtime lib means it will also include its native dependencies (e.g. sdl). Users could delete those libs manually - or with some custom build steps. But in the portable lib all API is available, including API that requires a platform-specific implementation. So who's to say it doesn't require those native libraries? 

I also think we should have a 'portable' assembly again. The improved way to do this is with [reference assemblies](https://docs.microsoft.com/en-us/dotnet/standard/assembly/reference-assemblies). They work the same as the Piranha skeleton build, i.e. only expose public API without any implementation. Only now there's official support for it.

Right now - as @harry-cpp has mentioned in several threads - you can reference any runtime assembly and switch it out at runtime. Essentially that's the bait-and-switch trick used with the PCL assembly too. But this is confusing for library authors. On top of that the API surface is not exactly the same for all platforms (which is an issue we should fix), so some APIs might fail at runtime. With a reference assembly and documentation that explains the process we have a logical system for library authors to build cross-platform libs.

Jjagg:
> Not sure why but it seems like we're losing all commit history on these moved files.

We're not. Check out the git blame. Git detected that these were just file moves and you can follow the history. It's just that GitHub does not do that (yet) for the history tab: https://github.com/isaacs/github/issues/900

MonoGame.Framework.Content.Pipeline/Effect/GraphicsState/SamplerStateData.cs
Jjagg: Oops, that's a mistake.

Jjagg:
> So for example someone already has a pipeline extension that uses RasterizerState they will have continue to use it... worse they have to link against a platform specific version of Framework.

Yes, but I don't know if anyone really uses the graphics state classes in extension libs. They aren't exposed anywhere by the CP, so I can't really think of a case where you'd need to use them.

MonoGame.Framework.Content.Pipeline/Utilities/RuntimeTypeStrings.cs
Jjagg: I intended for this to be internal. It might be useful to make it public for CP extensions that build xnbs that should be read with built-in readers.

vpenades:
> That does solve our internal problem, but doesn't work for others making pipeline extensions. So for example someone already has a pipeline extension that uses `RasterizerState` they will have continue to use it... worse they have to link against a platform specific version of Framework.

@Jjagg I'm not sure what's the issue, but if you're having issues of moving internal classes to core, or moving public classes that depend on internal classes that are also used in the framework, the solution would be as easy as to include this in core:

[assembly:InternalsVisibleTo("Monogame.Framework.WindowsDX")]
[assembly:InternalsVisibleTo("Monogame.Framework.DesktopGL")]
[assembly:InternalsVisibleTo("Monogame.Framework.XXXX")]

That would make the internal clases of Core visible to the platform specific classes.

> Why is [MonoGame.Framework.Portable](https://www.nuget.org/packages/MonoGame.Framework.Portable) not a better solution to all this?

Personally, I always had troubles using the portable package. The theory is that I could create a Monogame.FancyEffects using the portable package. And when using Monogame.FancyEffects into a project that, let's say, uses Monogame.Framework.DesktopGL,   the DesktopGL DLL would automatically replace/override the Portable DLL.

But this never worked for me. whenever I ran an application in such configuration, I always ended with DLL conflicts.

Keep in mind that since the moment it was available, I always created projects with the new SDK style, which supports transitive dependencies, so at some point I figured out that both the portable DLL and the DesktopGL were being copied to the target directory at the same time, which ended one DLL overwriting the other at random.

So I suspect the portable package is only able to work with old verbose csproj style that is not transitive and requires you to reference packages explicity on every project, regardless if in a referenced project is also included. That's why it worked for people using the templates (which are old style csproj).

This all comes by personal experince, but it makes sense: the portable library is not playing well with transitive dependencies of the new SDK style projects.

harry-cpp:
> Not sure why but it seems like we're losing all commit history on these moved files. We really should figure out why this is happening and make sure that doesn't happen in the final commit. It makes history very hard to track down later.

Nah, we are not loosing history, GitHub just sucks at displaying it. ie. here is Threading.cs:
Screenshot from 2020-04-04 11-59-10

> Portable Portable Portable

We already support it, just add a reference in your .NET Standard project like so, and the portable assembly will be replaced in the final project:

```xml
<ItemGroup>
    <PackageReference Include="MonoGame.Framework.DesktopGL" Version="3.8.0.X-develop">
    </PackageReference>
</ItemGroup>
```

I also wrote the docs for properly setting up a new project with .NET standard library or Shared project: https://github.com/MonoGame/MonoGame/blob/7f20e9da6d077b839ce1e32730e4e4f4fbbb803d/Documentation/setting_up_project/setting_up_project_vscode.md

harry-cpp:
(The docs are in this PR https://github.com/MonoGame/MonoGame/pull/6988)

tomspilman:
> Git detected that these were just file moves and you can follow the history. It's just that GitHub does not do that (yet) for the history tab: isaacs/github#900

@Jjagg 

Ugh... that is still a bug in GitHub?  I wonder how much people's time has been wasted with this.


> While the portable assembly can be used in the first case it does not make a lot of sense.
It still requires one of the MG platform implementation runtime libs to run the code. What platform should they pick?

They should pick MonoGame.Framework.Portable.  

The portable assembly doesn't have to be stripped of all code.  It could still be a fully usable assembly on its own.  Just the platform bits... graphics, audio, input implementations... are not implemented.  No native dependencies of any kind.

Portable really still seems like a bad name.  Portable was never really a usable assembly.  I think the correct name would make this concept make more  sense.


> The improved way to do this is with reference assemblies.

Interesting...  that is good that they are acknowledging this is an issue.  And i agree if we have a `.Core` we should still keep `.Portable` and move to reference assemblies.


> On top of that the API surface is not exactly the same for all platforms (which is an issue we should fix), so some APIs might fail at runtime. 

Right... that is an issue we need to address.  The only way forward i can see is one API that is documented to tell you platforms which may not support it.  But we should likely keep this discussion to another thread.


> We already support it, just add a reference in your .NET Standard project like so, and the portable assembly will be replaced in the final project:

@harry-cpp 

Oh...  i see what you mean there.  By using `<PrivateAssets>` you keep the compiler from copying all the native dependencies to the project.  Basically turning whatever platform MonoGame.Framework assembly we reference into something that can run on any platform.  This is essentially what i am proposing instead of `Core` but in a more proper way by having in essence a null/headless/unimplemented platform.

>  the portable library is not playing well with transitive dependencies of the new SDK style projects.

@vpenades 

Do you know if this is correct @Jjagg @harry-cpp ?

Seems to me that using  `<PrivateAssets>` in a `<PackageReference>` or `<Private>` in a `<ProjectReference>` would control that.

harry-cpp:
> the portable library is not playing well with transitive dependencies of the new SDK style projects.

That is correct, however if you add `<PrivateAssets>All</PrivateAssets>` it works fine as it does not copy over the managed library or the dependencies, however because of that, and because of .NET Standard, there is no longer a reason to have a Portable version.

tomspilman:
> however if you add `<PrivateAssets>All</PrivateAssets>` it works fine as it does not copy over the managed library or the dependencies

It sucks that they broke bait and switch like this.  It was the best part of it all.


> however because of that, and because of .NET Standard, there is no longer a reason to have a Portable version.

To be clear i know the PCL was replaced with .NET Standard.  PCL isn't the part i care about.


While Core solves issues for referencing math types for instance...  it doesn't solve the issue of game libraries referencing VertexBuffer for example.  Core doesn't fully solve our problem.  We still have to combine it with "reference assemblies" to get to a full solution.

If for instance we had `MonoGame.Framework.Whatever` that had stubs for all graphics/audio/input native bits but otherwise all the other code... it would solve the issues of Core and solve the issues that "reference assemblies" do at the same time.  I believe it would solve all our issues without making things more complex for us as maintainers or for users of MG.

The broken bait and switch in new SDK style projects makes it worse now.  There has to be instructions on how to manually edit csprojs to fix things.  But maybe that isn't so bad.

Anyway... wanted to go thru this one last time before we commit to the additional complexity to the project.

Jjagg:
To summarize and so we can reference the issues and possible solutions:

There's two questions we'd like to answer:
- What should libraries reference if they need to be usable with multiple platform specific runtime assemblies (bait-and-switch)? (1)
- What should libraries or executables reference if they don't need any platform specific bits? (2)

And there's 3 suggested solutions:
- Use DesktopGL runtime assembly for both (a)
- Create 'Portable' assembly with implementation for both (b)
- Create reference assembly without implementation for (1) and split off core assembly for (2) (c)

vpenades:
> It sucks that they broke bait and switch like this. It was the best part of it all. They should have left the default behavior to be the same as before.

It depends on what are you working on, I guess..... the solution where we work has +300 projects, about 300.000 lines of code, no kidding. The new SDK style projects and transitive dependencies has been a blessing.

> What should libraries reference if they need to be usable with multiple platform specific runtime assemblies (bait-and-switch)?

I would like to propose yet another possible solution, which is to have some device abstractions in Core, which is what the libraries would use.

I know that abstract classes has been discussed, and rejected in the past, but my proposal is a bit different, instead of having abstract base classes, it would use interfaces. For example, SpriteBatch is typically constructed like this:

`SpriteBatch batch = new SpriteBatch(graphicsDevice);`

And this way of creating a SpriteBatch would stay, so whoever targets a specific platform will keep using the current API.

For whoever wants to be platform agnostic, SpriteBatch would implement ISpriteBatch interface, defined in core, and in order to create it you would do:

`ISpriteBatch batch = graphicsDevice.CreateCompatibleSpriteBatch();`

So GraphicsDevice would become a factory of sorts, that way we would have new methods and interfaces, but keeping compatibility with XNA, and games targeting one specific platform would keep using `new SpriteBatch(graphicsDevice);` construction method, so they would not incur in performance issues due to virtual calls.

I don't know if this is possible for all the platform specific classes, it's just an idea that sticked 😸

tomspilman:
@vpenades 

>  have some device abstractions in Core, which is what the libraries would use.

Actually we are working towards something like that.  If things go to plan we may eventually have one C# MonoGame.Framework only which works for all platforms.  The only thing that would change is the native libraries it references.

But still too early to know about that.


> `ISpriteBatch batch = graphicsDevice.CreateCompatibleSpriteBatch();`

Yeah that will never happen.  Interfaces are significantly slower than regular method calls.  Notice how XNA has almost no interfaces.  This will never change as by design interface calls require more work to be done.

And while an interface may not matter on a PC when doing a few dozen draws.  When targeting a Switch or a Vita like device and doing a few thousand draws it really matters.

tomspilman:
> And there's 3 suggested solutions:

@Jjagg 

Yeah that is a good summation of the questions and solutions.

To me as a fan of the simpler single assembly approach (b) seems like a better path.  It is a single solution to both question (1) and (2) and gives us the ability to define how things should behave,

While (a) works i think the assembly named `DestopGL` is confusing and most users will assume it isn't the right solution for either question (1) and (2).

With (b) we could also extend it do other useful things.  Like we could make it work 100% as sort of a "headless server".  You could load content, create a game object, allocate things, draw...  but it does nothing.  It could internally log things for other useful reasons.  Not that any of this is required... just that (b) is more flexible.

To me the big possible problem with (b) is that with the changes with the new SDK style projects and transitive dependencies makes bait and switch behavior tricky for users to understand and deal with.  Last thing we want is spending years answering questions on why someone's game/library isn't working right to explain how to fix it over and over again.

Jjagg:
> While (a) works i think the assembly named DestopGL is confusing and most users will assume it isn't the right solution for either question (1) and (2).

Definitely agree with this. I think it's weird and unintuitive for users to use DesktopGL in scenarios (1) or (2).

For me there are two important reasons to prefer (c) to (b). 
- The reason you mentioned @tomspilman. Library authors need to take extra steps in scenario (1) to make sure no assemblies/assets are included. Also people consuming user-made libraries can run them with the 'portable' assembly but will get runtime errors when unsupported methods are used. With a reference assembly (i.e. no implementation) users get an exception when the assembly is loaded.
- For scenario (2) the issue is that all API is accessible, but only part of it works. With a core assembly only API that works is available.

In my eyes problems (1) and (2) don't have a single solution that fits both. That's why I prefer different solutions for the different scenarios.

tomspilman:
This was never settled and moot at this point.

PR #7087, opened 2020-04-02
Manually import base MSBuild props file from root

Jjagg:
MSBuild stops discovery when it finds a Directory.Build.props file, so the one in the root folder of the repo did not get picked up.


Jjagg:
Actually might be nicer to keep a .props file in the root but manually import it from the other projects.

harry-cpp:
Don't templates and tools also need to pick it up?

Jjagg:
Tools import it in Tools/Directory.Build.props. Will fix templates!

PR #7088, opened 2020-04-02
Set up versioning with NBGV

Jjagg:
This sets up NerdBank.GitVersioning to automate version numbering.
Major and Minor version number are set in `version.json`. Patch number is automatically determined by git commit depth (number of commits since the last major/minor version bump). For development builds the git commit hash will be appended (e.g. `3.8.42-gc0ffee`), for public releases it will not.
`version.json` also determines how public releases happen. I left it at the default which does a public release
- from the master branch
- when a commit is tagged with `v{major}.{minor}`.

You can force a public release by setting the `PublicRelease` MSBuild switch to `true` though I didn't expose this in the CAKE script.

More info about public releases with NBGV [here](https://github.com/dotnet/Nerdbank.GitVersioning/blob/master/doc/public_vs_stable.md).

Note that this will change our versioning scheme. We only incremented the patch version for public releases (3.7.0, 3.7.1, 3.7.2), but with this the patch version will be commit depth.


Jjagg:
Need to fix the TeamCity build numbering

mrhelmut:
Is there an interest to append the commit hash? Nuget convention seems to be appending the package type, e.g. "-nightly", "-alpha", or "-develop".

Jjagg:
The git commit info is added for non-public releases, so you don't often see it on nuget.org. We'd only publish develop releases to our own feed with the git commit version.

With the commit ID in the version we can easily find the commit a development nuget package is built for.

And, from the NBGV readme:

> Prioritize absolute build reproducibility. Every single commit can be built and produce a unique version.

Also nice to know:

> This project is supported by the .NET Foundation.

PR #7093, opened 2020-04-07
Fix MGCB.Task edge cases

jnoyola:
Fixing various little bugs I found after more testing.

* Fix String.Split to split on string delimiter instead of multiple character delimiters
* Quote copied strings in case they contain commas
* Apply MGCB platform after parameters file


Jjagg:
Awesome, thanks!

PR #7095, opened 2020-04-09
Fixing openal audio duplicating sourceids

### LINKED ISSUE #7094: [DesktopGL] Issuing duplicate source ids for sound effects, dynamic sound effects, and songs
Because [availableSourcesCollection is a List(int) within OpenALSoundController.cs](https://github.com/MonoGame/MonoGame/blob/3b0df35ce1c73435f48953ccd2bb1a8b72c507be/MonoGame.Framework/Platform/Audio/OpenALSoundController.cs#L98) multiple integers of the same value can be added to it.

There are possible instances where ReleaseSource(int sourceId) function can be is called for the same sourceId.

If two sound effect instances end up with the same source id there are invalid operation exceptions within openal leading to sounds not playing, dynamic sound effects looping, and other problems.

There are a few instances where this has occurred in our projects.

**Instance A:**

- When DynamicSoundEffects are disposed [PlatformDispose() and base.Dispose()](https://github.com/MonoGame/MonoGame/blob/3b0df35ce1c73435f48953ccd2bb1a8b72c507be/MonoGame.Framework/Audio/DynamicSoundEffectInstance.cs#L278) are called.

- [PlatformDispose() releases the source by calling RecycleSource instead of FreeSource so HasSourceId is never set to false](https://github.com/MonoGame/MonoGame/blob/3b0df35ce1c73435f48953ccd2bb1a8b72c507be/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs#L114)

- base.Dispose() calls [base.PlatformDispose()](https://github.com/MonoGame/MonoGame/blob/3b0df35ce1c73435f48953ccd2bb1a8b72c507be/MonoGame.Framework/Audio/SoundEffectInstance.cs#L214) which calls [FreeSource()](https://github.com/MonoGame/MonoGame/blob/3b0df35ce1c73435f48953ccd2bb1a8b72c507be/MonoGame.Framework/Platform/Audio/SoundEffectInstance.OpenAL.cs#L358) and [releases the source again](https://github.com/MonoGame/MonoGame/blob/3b0df35ce1c73435f48953ccd2bb1a8b72c507be/MonoGame.Framework/Platform/Audio/SoundEffectInstance.OpenAL.cs#L193).

**Solution A:** This should be as easy as using controller.FreeSource() instead of [controller.RecycleSource()](https://github.com/MonoGame/MonoGame/blob/3b0df35ce1c73435f48953ccd2bb1a8b72c507be/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs#L114), and adding a conditional check for HasSourceId == true beforehand so it cannot be disposed more than once.

--------------------------------

**Instance B:**
Unfortunately that is not the only case where this can happen. There is also the possibility that GC tries to dispose SoundEffects while SoundEffectInstancePool.Update() is stopping finished sounds. Both of those instances running at the same time result in the available openAl sources list being filled with duplicate ids.

**Solution for B:** Wrap the FreeSource() functions within mutex locks so that we can be sure that the source hasn't been freed and the source isn't currently being freed before finally recycling the source.

--------------------------------

Monogame DesktopGL v3.7.1.189 also occurs on [current](https://github.com/MonoGame/MonoGame/commit/3b0df35ce1c73435f48953ccd2bb1a8b72c507be) dev branch.

Windows 10 (also applies to OSX Catalina, and assuming Linux OSes)

- DesktopGL

jcsnider:
Resolves #7094.

Ensures that no two threads try to release openal source ids at the same time.

Ensures that openal source id aren't released twice by making sure HasSourceId is true beforehand.

### DIFF
--- a/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs
+++ b/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs
@@ -106,12 +106,12 @@ namespace Microsoft.Xna.Framework.Audio
         {
             // Stop the source and bind null buffer so that it can be recycled
             AL.GetError();
-            if (AL.IsSource(SourceId))
+            if (HasSourceId && AL.IsSource(SourceId))
             {
                 AL.SourceStop(SourceId);
                 AL.Source(SourceId, ALSourcei.Buffer, 0);
                 ALHelper.CheckError("Failed to stop the source.");
-                controller.RecycleSource(SourceId);
+                controller.FreeSource(this);
             }
             
             if (disposing)
--- a/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs
+++ b/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs
@@ -106,14 +106,18 @@ namespace Microsoft.Xna.Framework.Audio
         {
             // Stop the source and bind null buffer so that it can be recycled
             AL.GetError();
-            if (HasSourceId && AL.IsSource(SourceId))
+
+            lock (sourceMutex)
             {
-                AL.SourceStop(SourceId);
-                AL.Source(SourceId, ALSourcei.Buffer, 0);
-                ALHelper.CheckError("Failed to stop the source.");
-                controller.FreeSource(this);
+                if (HasSourceId && AL.IsSource(SourceId))
+                {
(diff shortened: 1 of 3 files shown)


Jjagg:
Thanks for your report and this PR! This is great :) 

I think the dispose code is hard to follow currently, that's probably part of what caused this issue in the first place. I'll see if I can refactor this a little to make this part of the codebase more manageable.

Anyway, this looks good. Merging!

Issue #6415, opened 2018-08-18
Provide a way to prevent closing of Monogame Window


monsieurmax:
Hello,

I'm currently writing a level editor in Monogame 3.7 Desktop GL.

I'd like to intercept window exiting to be able to let the user save his work or cancel exiting.
It is a pretty common feature, maybe not useful for games, but for other tools.

Right now, we can only perform some actions while the game is exiting, but not cancel it.

What would be nice is to have a method to overload that returns true or false, depending on wether the game is authorized to exit.

Thanks in advance.

Jjagg:
We could have a flag in the `Game.Exiting` event argument that you can set to cancel exiting. That's simple to implement and a non-breaking way to implement this feature. Sorry for the extremely late response!

tomspilman:
If we can do it properly and support at least all desktop platforms i guess it is ok.

My concern is that you cannot keep a mobile or console game from shutting down.  So this is exclusively a PC feature really.

Jjagg:
Yes, we'll just be able to implement this for desktop. I guess this falls in the same bucket as #6392. 

For synchronous run loop we simply wrap [this code](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Game.cs#L397-L401) in a while loop with the stop condition that the new flag blocking game exit isn't set.

I'll have to look into async run loop.

jcsnider:
@Jjagg: I created a PR for this above (#7096) ... My changes only impact the synchronous loops because it doesn't look like async run loops are supported for the [DesktopGL](https://github.com/MonoGame/MonoGame/blob/5a75f402acf32fbed9db56a118e64ad7bac1a567/MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs#L258) and [WindowsDX](https://github.com/MonoGame/MonoGame/blob/5a75f402acf32fbed9db56a118e64ad7bac1a567/MonoGame.Framework/Platform/Windows/WinFormsGamePlatform.cs#L68) platforms.

As I mentioned in the PR I'm not sure if that's exactly how we should go about this, but I'm very happy to discuss and modify in order to get this to an acceptable state. Thanks!

topnik-code:
I am in the same situation: I would like to intercept window closing under Desktop GL. 🙂

Are there any news on this issue? I would also be happy to get any tip regarding a workaround! 😸

PR #7096, opened 2020-04-10
[WinDx && DesktopGL] Method to prevent closing of Monogame Window

jcsnider:
PR for issue #6415 

This edit creates a new Game.cs event that will fire before the application would exit on all platforms. The BeforeExit() event arguments contains a .Cancel property that, when set to true on WindowsDX and DesktopGL platforms will keep running the Platform.RunLoop() code effectively keeping players in-game.

In our use case this will be great to inform players online that they are about to exit while in combat, and that their character will remain in-game until they are deemed safe or die.

Tested and working. 

If you all would like this to be done differently I'm happy to discuss and revise in order to get this merged in. Thanks!

### DIFF
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Xna.Framework
 
         private bool _shouldExit;
         private bool _suppressDraw;
-        private bool _exitAborted = true;
+        private bool _processExit;
 
         partial void PlatformConstruct();       
 
@@ -397,10 +397,10 @@ namespace Microsoft.Xna.Framework
                 break;
             case GameRunBehavior.Synchronous:
                 //True by default, we don't want to exit on launch
-                while (_exitAborted)
+                while (!_processExit)
                 {
-                    //We don't want to continue looping unless (on desktop platforms) the exit is terminated
-                    _exitAborted = false;
+                    //Next time an exit is requested we want to break out of this loop unless we're on a desktop platform and the exit is aborted in the OnBeforeExit event.
+                    _processExit = true;
 
                     // XNA runs one Update even before showing the window
                     DoUpdate(new GameTime());
@@ -726,7 +726,7 @@ namespace Microsoft.Xna.Framework
 
 #if (WINDOWS && DIRECTX) || DESKTOPGL
             //Only allow aborting of the exit process on desktop platforms where it is supported.
-            _exitAborted = args.Cancel;
+            _processExit = !args.Cancel;
 #endif
         }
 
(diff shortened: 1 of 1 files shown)


mrhelmut:
Hey there,

I feel like the first ```DoUpdate``` and ```EndRun``` should be out of the loop.

```
            case GameRunBehavior.Synchronous:
                // XNA runs one Update even before showing the window
                DoUpdate(new GameTime());

                // True by default, we don't want to exit on launch
                while (!_processExit)
                {
                    // Next time an exit is requested we want to break out of this loop unless we're on a desktop platform and the exit is aborted in the OnBeforeExit event.
                    _processExit = true;

                    Platform.RunLoop();

                    // Raise the before exiting event and provide an opportunity for the exit to be prevented
                    DoBeforeExit();
                }

                EndRun();

                DoExiting();
                break;
```

jcsnider:
@mrhelmut thanks for looking this over. I've committed those changes 👍

Jjagg:
Thanks @jcsnider! A couple notes:

I mentioned in #6415 that I would add the argument for cancellation to the Exiting event. I think that would be nicer. Technically changing the event type is a breaking change, but in practice it won't affect implementations. And I don't like the clutter of adding another event. The name of the event `Exiting` already implies it happens before the actual game exit, consistent with e.g. `GraphicsDevice.DeviceResetting` (as opposed to `GraphicsDevice.DeviceReset` which is called after the reset).

We need to make sure platforms do their setup outside the `RunLoop` method. I just did a quick check and WinDX does not do that: https://github.com/MonoGame/MonoGame/blob/e8ec2db4fff49c3dfb0d33343b3402363c0b3f4d/MonoGame.Framework/Platform/Windows/WinFormsGameWindow.cs#L448
That code should be moved to `WinFormsGameWindow.Initialize`. I haven't checked the SDL implementation.

`Platform.Exit` is called inside [Tick](https://github.com/MonoGame/MonoGame/blob/e8ec2db4fff49c3dfb0d33343b3402363c0b3f4d/MonoGame.Framework/Game.cs#L537). That should be pulled out and called when exit is confirmed outside the loop in the run method.

jcsnider:
I wasn't sure if modifying the existing Exiting event would be acceptable or not given that it could be breaking. 

I will take a stab at those edits tomorrow and see how far I can get. 

Thanks again guys.

jcsnider:
The `Platform.Exit` code for SDL only results escaping the RunLoop by incrementing an integer variable `_isExiting`. For DirectX `Platform.Exit()` actually has to dispose of the game window in order to escape the blocking `Application.Run()` command.

I am not sure how I'd end up firing an event in order to prevent a close while the DirectX side was still running, as disposing and recreating the Directx window isn't an option.

On top of that, I'm unsure how these changes would impact other platforms, which we definitely don't want to break.. so unfortunately I think this might be beyond what I am able to work out and contribute for the time being.

I am going to go ahead and close this PR and I might take another stab at it later. Thanks so much for the input, advise, and entertaining this solution which wouldn't have worked at all outside of DesktopGL XD

Jjagg:
No worries, thanks for looking into this some more :)

PR #7105, opened 2020-04-12
Support canceling game exit

### LINKED ISSUE #6415: Provide a way to prevent closing of Monogame Window

Jjagg:
Felt inspired by #7096. Fixes #6415.

Note that this breaks user code that overrides `OnExiting` because the signature changed. I think however that that's very minor and this is better than cluttering the API with another way to do it.

@tomspilman @mrhelmut @harry-cpp I made the assumption that all platforms use `Tick` to run their game loop. Note that invocation of the `Exiting` event, the `EndRun` call and the `UnloadContent` call now only happen from `Tick`. So if closed platforms have another way to run the game loop they need to call those methods where they exit the game. If so the whole [`_shouldExit` block](https://github.com/MonoGame/MonoGame/compare/develop...Jjagg:exit-cancel?expand=1#diff-c9a47e02910a8951188f9eb0ac3615e8R522-R534) should be extracted into a separate method.

@jcsnider Please try this if you're interested :)

### DIFF
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -362,6 +362,8 @@ namespace Microsoft.Xna.Framework
 
         #region Events
 
+        private readonly ExitingEventArgs _exitingEventArgs = new ExitingEventArgs();
+
         /// <summary>
         /// Raised when the game gains focus.
         /// </summary>
@@ -380,7 +382,7 @@ namespace Microsoft.Xna.Framework
         /// <summary>
         /// Raised when this game is exiting.
         /// </summary>
-        public event EventHandler<EventArgs> Exiting;
+        public event EventHandler<ExitingEventArgs> Exiting;
 
 #if WINDOWS_UAP
         [CLSCompliant(false)]
@@ -495,8 +497,6 @@ namespace Microsoft.Xna.Framework
                 DoUpdate(new GameTime());
 
                 Platform.RunLoop();
-                EndRun();
-				DoExiting();
                 break;
             default:
                 throw new ArgumentException(string.Format(
@@ -631,8 +631,17 @@ namespace Microsoft.Xna.Framework
 
             if (_shouldExit)
             {
-                Platform.Exit();
-                _shouldExit = false; //prevents perpetual exiting on platforms supporting resume.
+                _exitingEventArgs.Cancel = false;
+                OnExiting(this, _exitingEventArgs);
+
+                if (!_exitingEventArgs.Cancel)
+                {
+                    Platform.Exit();
+                    EndRun();
+                    UnloadContent();
+                }
+
+                _shouldExit = false;
             }
         }
 
@@ -743,7 +752,7 @@ namespace Microsoft.Xna.Framework
         /// </summary>
         /// <param name="sender">This <see cref="Game"/>.</param>
         /// <param name="args">The arguments to the <see cref="Exiting"/> event.</param>
-        protected virtual void OnExiting(object sender, EventArgs args)
+        protected virtual void OnExiting(object sender, ExitingEventArgs args)
         {
             EventHelpers.Raise(sender, Exiting, args);
         }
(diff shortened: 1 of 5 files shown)


jcsnider:
Wow, this is awesome, and much simpler than what I had gotten myself into for my second iteration haha.

Not sure if I'll find time today given the Holiday but I will definitely see if I can test the DesktopGL systems (Win/Mac/Linux) and DirectX tomorrow. 

Really appreciate it!

jcsnider:
Sorry it took me so long to test and get back to you.

It seems like your changes are tied into the `Game.Exit()` function. Which sets `_shouldExit = true`. Those changes work when I manually call the `Game.Exit()` function.

Unfortunately when closing the DesktopGL window `Game.Exit()` doesn't seem to execute, so the window closes without firing the `OnExiting` event.

For WindowsDX the window seems to be minimized, and I can't get it to actually appear in the foreground (not sure if that's something broken on my end or not, this is a clean project). Same behavior though as far as close button never calling the `OnExiting` event.

Jjagg:
Oops, yes that's quite the oversight on my part. I just added a commit that should fix this, but I haven't tested it at all yet.

MonoGame.Framework/Game.cs
```diff
@@ -303,10 +303,12 @@ internal bool Initialized
 
         #region Events
 
+        private readonly ExitingEventArgs _exitingEventArgs = new ExitingEventArgs();
```
tomspilman, line 306: I assume this is some sort of memory optimization?

Seems unnecessary and complicates the PR.  It isn't like we get dozens of exit events each frame.

MonoGame.Framework/Game.cs
```diff
@@ -303,10 +303,12 @@ internal bool Initialized
 
         #region Events
 
+        private readonly ExitingEventArgs _exitingEventArgs = new ExitingEventArgs();
+
         public event EventHandler<EventArgs> Activated;
         public event EventHandler<EventArgs> Deactivated;
         public event EventHandler<EventArgs> Disposed;
-        public event EventHandler<EventArgs> Exiting;
+        public event EventHandler<ExitingEventArgs> Exiting;
```
tomspilman, line 311: My concern here is that some MG titles are still using XNA on Windows.

This makes it inconvenient for them to work around this change.

IMO we should keep this as `EventArgs` and make it the responsibility of the implementer of `OnExiting` to cast to `ExitingEventArgs` if they want to use that feature.

I would use this opportunity to add documentation here to this event.

harry-cpp: This does not break anything, I assume you are worried about the following, but the event params do get autocast and it does compile:

```cs
// Inside MonoGame:
public class ExitingEventArgs() : EventArgs;
public static event EventHandler<ExitingEventArgs> Exiting;

// Inside the XNA based game:
public static void Main()
{
    Exiting += MyCustomExitEvent;
}

private static void MyCustomExitEvent(object sender, EventArgs args) { }
```

MonoGame.Framework/Game.cs
Jjagg: Yes, just to avoid allocating. I agree it's unnecessary in this case.

MonoGame.Framework/Game.cs
tomspilman: Ahh... i had not thought of that @harry-cpp .

tomspilman:
@Jjagg this is conflicted... if you fix it we can merge it.

MonoGame.Framework/Game.cs
Jjagg: Note that this does break user code that overrides OnExiting in a Game implementation because the signature changed.

Jjagg:
@harry-cpp @tomspilman this is good to go, but note that it does break user code that overrides OnExiting in a Game implementation because the signature changed.

SimonDarksideJ (approved):

PR #7107, opened 2020-04-13
Add script for automatic wine environment setup for mgfxc

harry-cpp:
Was gonna add this script a week or 2 later with the docs, but I thought I might as well add it now so I can point people to use it if they want to.


Tools/MonoGame.Effect.Compiler/mgfxc_wine_setup.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+# This script is used to setup the needed Wine enviorment
```
Jjagg, line 2: enviorment

harry-cpp: 😅

PR #7108, opened 2020-04-15
Fix incompatibility in cake script when multiple versions of .Net Core are installed on Windows 10

### LINKED ISSUE #7092: Pipeline tool will not open on Windows 10 (MG build 3.8.0.1082)
I could no longer open the MonoGame content pipeline executable after updating to the most recent development build of MonoGame.

Using Windows 10.

Steps to reproduce:

- Download MonoGame build 3.8.0.1082 installer and installer
- Attempt to open the MonoGame pipeline tool. Blue circle spins for fraction of second and nothing else occurs. Restart doesn't help. Uninstalling and reinstalling MonoGame does not help.
- Download MonoGame 3.7.1 (last official release)
- MonoGame pipeline tool runs properly again

I'm willing to pull extra system details if useful in investigation. Fortunately, anyone can just revert to the last official release to continue working.

Brett208:
When both .Net Core 3.1 & 2.1 are both installed, cake script aborts during BuildTools portion with the following message:

```
C:\Program Files\dotnet\sdk\2.1.512\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(137,5): error NETSDK1045: The current .NET SDK does not support targeting .NET Core 3.1.  Either target .NET Core 2.1 or lower, or use a version of the .NET SDK that supports .NET Core 3.1. [\MonoGame\Tools\MonoGame.Content.Builder\MonoGame.Content.Builder.csproj]
An error occurred when executing task 'BuildTools'.
Error: One or more errors occurred. (MSBuild: Process returned an error (exit code 1).)
        MSBuild: Process returned an error (exit code 1).
```

Closes #7092 

Tested locally on Windows 10 64Bit

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -164,6 +164,7 @@ Task("BuildContentPipeline")
 
 Task("BuildTools")
     .IsDependentOn("Prep")
+    .WithCriteria(() => GetMSBuildWith("Microsoft.VisualStudio.Component.Windows10SDK.18362"))
     .Does(() =>
 {
     DotNetCoreRestore("Tools/MonoGame.Content.Builder/MonoGame.Content.Builder.csproj");


Jjagg:
I don't think this is the right fix.
It looks like this will only work on Windows and if you have the specified Win 10 SDK and .net core installed in a Visual Studio installation.

cc @harry-cpp

harry-cpp:
GetMSBuildWith only does stuff on Windows, there is an If(Windows) thingy inside that method.

mrhelmut:
Still, I believe that many devs don't have the Win 10 SDK and don't need to (it's mostly a thing that you have when targeting UWP).
Also, relying on a precise version makes it likely to break or to not being the proper version that users targeting UWP are actually using.

Jjagg:
Wouldn't this work if we used the .NET Core settings from the cake build script to build the tools? .NET Core CLI should be able to pick a matching SDK version.

Brett208:
Is anyone else able to reproduce the bug by installing .Net Core 3.1 & 2.1 on their machine and running 'dotnet cake'?

@Jjagg, I'm not familiar with cake, so not sure how to implement what you are suggesting.

PR #7109, opened 2020-04-15
Fixed deconstruction of color for bytes and floats

Chapnik13:
Deconstruction of color class returned floats instead of bytes with range of 0-255,
If tried to pass them to constructor of color class it would create white color because of the mismatch with the float/byte constructor.

I modified the class so not it will have 2 deconstructors, one for floats and one for bytes (included with and without alpha channel).


Jjagg:
Great, thanks!

Issue #6347, opened 2018-06-27
Support text compositions


Jjagg:
MonoGame does not support getting IMM compositions. We should expose an API so games can render these compositions instead of having the OS show an ugly window for it.

We should discuss the API before talking about the implementation. The API should support starting and stopping composited input and getting the current composition (possibly using an event similar to `TextInput`). `TextInput` should be invoked when a composition is finished.

### API Suggestion

```
/// <summary>
/// Enable the system IMM service to support composited character input.
/// </summary>
/// <seealso cref="EndTextComposition" />
/// <seealso cref="TextComposition" />
void StartTextComposition();

/// <summary>
/// Stop the system IMM service.
/// </summary>
/// <seealso cref="StartTextComposition" />
/// <seealso cref="TextComposition" />
void StopTextComposition();

/// <summary>
/// Invoked when the IMM service is enabled and a character composition is changed.
/// </summary>
/// <seealso cref="StartTextComposition" />
/// <seealso cref="EndTextComposition" />
event EventHandler<TextCompositionEventArgs> TextComposition;

class TextCompositionEventArgs : EventArgs
{
    public string CompositedText { get; }
    public int CursorPosition { get; }
}
```

### Links

Forum post: http://community.monogame.net/t/i-cant-use-ime-for-chinese-input/10737
Microsoft docs: https://docs.microsoft.com/en-us/windows/desktop/DxTechArts/using-an-input-method-editor-in-a-game
SDL docs: https://wiki.libsdl.org/Tutorials/TextInput

### Implementation notes

- The API should be exposed in [`GameWindow`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/GameWindow.cs)
- The Windows implementation needs to handle certain window messages in the [`WndProc`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Windows/WinFormsGameForm.cs#L72)
- The SDL implementation (DesktopGL) handles events in [`SDLGamePlatform`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/SDL/SDLGamePlatform.cs#L116).

Jjagg:
cc @tomspilman @cra0zy @dellis1972 because this adds API.

mrhelmut:
All consoles/mobiles have their own native IME. Most of the (console) ports have some wrapper support for it, but nothing that would be candidate for a high level cross-platform API.

[...]

PR #7117, opened 2020-04-17
Update docs for 3.8

Jjagg:
This adds docs for
- creating a project
- list of target platforms
- migrating to 3.8 for existing games (I didn't add this to the table of contents)

I also updated tool naming in the docs:
- 2MGFX -> MGFXC
- Pipeline Tool -> MGCB Editor

I did not make changes to the existing getting started guide in this PR. I'm working on redoing it completely, partly based on MrGrak's [Getting Started Guide](https://github.com/MrGrak/Monogame-Getting-Started).

PR #7118, opened 2020-04-17
Add text composition support

ryancheung:
This is the PR mentioned before in the following issue: https://github.com/MonoGame/MonoGame/issues/6347#issuecomment-604898284

A `IImeService` is added MonoGame.Framework/Input folder. This service is going to provide text input or composition APIs for SDL2, WindowsDX, Android and iOS platforms. More platforms like (UAP, PlayStation 4, Nintendo Switch etc.) could be added in the future by someone, if supported.

Platform specific IImeService implementations to be implemented:

- WinFormsImeHandler.cs
- SdlImeHandler.cs
- AndroidImeHandler.cs
- iOSImeHandler.cs

Notes on the Android/iOS implementation: The implementation of ImeHandler in Android or iOS is basiclly by creating a native invisible textbox (Android is EditText, iOS is UIBackwardsTextField) and then listening the text change events on them.


MonoGame.Framework/Platform/Windows/Input/ImeCompositionAttributes.cs
```diff
@@ -0,0 +1,37 @@
+namespace Microsoft.Xna.Framework.Windows.Input
+{
+    /// <summary>
+    /// Winforms IME composition character attributes
+    /// </summary>
+    public enum ImeCompositionAttributes
```
Jjagg, line 6:
```suggestion
    internal enum ImeCompositionAttributes
```

MonoGame.Framework/Input/IImeService.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Stop the system IMM service.
+        /// </summary>
+        void StopTextInput();
+
+        /// <summary>
+        /// Show the Ime Candidate window renderer by the OS.
+        /// In SDL2 platform it's always true since there's no API to get candidate texts in SDL2.
+        /// In WindowsDX platform, it's possible to choose show candiate window renderer by the OS or render it yourself.
```
Jjagg, line 29:
```suggestion
        /// Show the Ime Candidate window rendered by the OS.
        /// Set to <c>false</c> if you want to render the IME candidate list yourself.
        /// On DesktopGL the candidate list is not exposed and this will always return <c>true</c>.
```

Jjagg: Is this only about the candidate list or does the OS window also show the current composition?
We should make that clear in the docs.
If it's only the candidate list we should probably rename the property to `RenderPlatformCandidateList` or something like that.

MonoGame.Framework/Input/IImeService.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// </summary>
+        bool ShowOSImeWindow { get; set; }
+
+        /// <summary>
+        /// Position Y of virtual keyboard, for mobile platforms has virtual keyboard.
```
Jjagg, line 34:
```suggestion
        /// Height of the virtual keyboard, for mobile platforms that have a virtual keyboard.
```

Hmm, we have the static [`KeyboardInput`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Input/KeyboardInput.cs) class that also exposes API to show a virtual keyboard for mobile.
This API seems like it would fit in there better than the IME service.

MonoGame.Framework/Input/IImeService.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// </summary>
+        int VirtualKeyboardHeight { get; }
+
+        /// <summary>
+        /// Update the Ime service. For instance, stop the text input when clicking on the other
+        /// screen area than the virtual keyboard poped up.
+        /// </summary>
+        void Update(GameTime gameTime);
```
Jjagg, line 42: We could update the `ImeService` internally so users don't have to. That would simplify usage a little.

MonoGame.Framework/Input/IImeService.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// Return true if text input enabled else false;
```
Jjagg, line 45:
```suggestion
        /// Returns true if text input is enabled, else returns false.
```

MonoGame.Framework/Input/IImeService.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// </summary>
+        bool IsTextInputActive { get; }
+
+        /// <summary>
+        /// Set the position of Candiate Window rendered by the OS, supported platforms includes SDL and WindowsDX.
+        /// In SDL2 there's no API to get the candiate list. We have to use the candidate windows provided by the OS.
+        /// But, in windows SDL2 has bug that not showing the candidate window of the OS, so SDL2
```
Jjagg, line 52: Can you provide the link to the bug report if there is one?

MonoGame.Framework/Input/IImeService.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// platform was suggested not to use in Windows, use WindowsDX instead.
```
Jjagg, line 53:
```suggestion
        /// Set the position of the candidate window rendered by the OS.
        /// Let the OS render the candidate window by setting <see cref="ShowOSImeWindow"/> to <c>true</c>.
        ///
        /// This API is supported on DesktopGL and WindowsDX.
```

We should mention the DesktopGL Windows bug in the class documentation. You can't get a candidate list and can't make the OS render the candidate list, so effectively the whole ImeService is useless due to those things on Windows with DesktopGL.

MonoGame.Framework/Input/IImeService.cs
```diff
@@ -0,0 +1,67 @@
...
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+
+namespace Microsoft.Xna.Framework
+{
+    /// <summary>
+    /// Interface to handle text composition.
+    /// </summary>
+    public interface IImeService
```
Jjagg, line 12: Shouldn't we call this `IImmService`?

MonoGame.Framework/TextCompositionEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+            CursorPosition = cursorPosition;
+
+            CandidateList = candidateList;
+            CandidatePageStart = candidatePageStart;
+            CandidatePageSize = candidatePageSize;
+            CandidateSelection = candidateSelection;
+        }
+
+        /// <summary>
+        /// The full string as it's composited by the IMM.
```
Jjagg, line 28:
```suggestion
        /// The full string as it's composed by the IMM.
```

MonoGame.Framework/TextCompositionEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// </summary>    
+        public readonly string CompositionString;
+
+        /// <summary>
+        /// The position of the cursor inside the composited string.
```
Jjagg, line 33:
```suggestion
        /// The position of the cursor inside the composed string.
```

MonoGame.Framework/TextCompositionEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// </summary>    
+        public readonly int CursorPosition;
+
+        /// <summary>
+        /// The candidate text list for the composition.
+        /// Note candidate properties are only valid in  WindowsDX(or UAP) platform.
```
Jjagg, line 39:
```suggestion
        /// This property is only supported on WindowsDX and WindowsUniversal.
```

MonoGame.Framework/TextCompositionEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// It could be empty if compsition string does not generates candidates.
```
Jjagg, line 40:
```suggestion
        /// If the composition string does not generate candidates this array is empty.
```

MonoGame.Framework/TextCompositionEventArgs.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// First candidate index of current page.
+        /// </summary>
+        public readonly uint CandidatePageStart;
+
+        /// <summary>
+        /// How many candidates should display per page.
+        /// </summary>
+        public readonly uint CandidatePageSize;
```
Jjagg, line 52: The OS provides this information? It seems like this is a renderer decision.

ryancheung:
Thanks for the reviews, it's midnight here, I'll update the PR tomorrow.

MonoGame.Framework/Input/IImeService.cs
ryancheung: No, it's not just about the candidate list, the OS renders nothing about IME in SDL2, So in SDL2, we have to render both text composition and candidate list by ourself.

MonoGame.Framework/Input/IImeService.cs
ryancheung: The mobile api in `KeyboardInput` does different things like this ImeService, it creates a popup native Android/iOS view and put a textfield control in it. But ImeService is trying to provide a unified interface to do populate text input events, and sometimes we need to know the virtual keyboard height.

MonoGame.Framework/Input/IImeService.cs
ryancheung: I agree this.

MonoGame.Framework/Input/IImeService.cs
ryancheung: Of course https://bugzilla.libsdl.org/show_bug.cgi?id=3421

MonoGame.Framework/Input/IImeService.cs
ryancheung: Sure. We can call that IImmService.

MonoGame.Framework/TextCompositionEventArgs.cs
ryancheung: Yeah. candidate list are rendered with pagination, since one text composition may generate lots of candidates.

ryancheung:
@Jjagg https://github.com/MonoGame/MonoGame/blob/089c37bdb1b0e7c28a73e3aca676588aaac51c68/MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs#L212 This line seems to shows text input event not support multi-bytes unicode characters like emojji, i'm just not sure about that. If it's true, we should switch the retrieve text code to something like `var compositionString = System.Text.Encoding.Unicode.GetString(ev.Text.Text, 32);`.

MonoGame.Framework/Input/IImeService.cs
ryancheung: @Jjagg If update `ImmService` internally, maybe make it a abstract class that could be better than an interface? Since interface cannot have non-public members.

ryancheung:
We can do some testing to check if the text composition and input events populate correctly now.

Jjagg:
> These lines seem to shows text input event not support multi-bytes unicode characters

Hmm, right. TextInputEventArgs only holds a char. We could add a `string Text` property but that would create garbage. It sucks that we can't use Span... We could reuse a char buffer then. I do think we should avoid garbage here.

IMO the best solution is to invoke TextInput multiple times in case of multi-codepoint input. However I also think one invocation should not contain only 1 char of a surrogate pair. So we should add a second character field to the TextInput event so we can raise the event one codepoint at a time.

Jjagg:
There is some more discussion about the char-based API and full Unicode support in #4805.

ryancheung:
@Jjagg I wasn't saying that we need SpriteFont to support Emoji, rendering Emoji is thing that some library like freetype2 does. 
But i mean the non-garbage text input code supports emojis characters, which is two surrogate per one emoji character. I wonder if the non-garbage text input code would drop unicode surrogates.

ryancheung:
And I think "TextInputEventArgs only holds a char."" is fine. The underline WinForm message populate char event just like this. 

For Emoji, if user composed a emoji character, there would be two `TextInputEventArgs` populated, each on is a surrogate. This should be fine also.

As GC, I think we can't avoid generating garbages here too. Since `TextCompositionEventArgs`  must holds a string `CompositionString`. It's just not feasible to populate one `TextCompositionEventArgs` for each char in CompositionString, since rendering composition string is more complicated than receiving a text input.

Jjagg:
I only brought up the emoji issue because there's a similar discussion about API design where we currently expose char, but we want to support full Unicode. TextInputEventArgs is similar. We need to support full Unicode strings, preferably without allocations.

But just invoking TextInput multiple times for strings that take more than 1 UTF-16 character is fine. Or at least it is not necessary to change this behavior.

ryancheung:
@Jjagg I added a demo project to test the text composition feature https://github.com/ryancheung/MonoGameTextCompositionSamples

Text composition in WindowsDX should work out of the box at the moment .

ryancheung:
@Jjagg DesktopGL should works also, you can see a screenshot in the Sample project.

Jjagg:
@ryancheung I'm testing your sample project on WindowsDX and DesktopGL currently.  First off, good job! It's really cool to be able to render IME natively in the game window :) 
There are 2 important things to discuss I think:
- Do we need the `IImmService.TextInput` event? SDL uses the same event for direct TextInput and IME results, I think that makes the most sense for us too (and it would be hard to do it differently on SDL). Shouldn't we just invoke `GameWindow.TextInput` for both direct input and composition results?
- `StartTextInput` and `StopTextInput` are not consistent on DesktopGL and WindowsDX. We need to decide what behavior we want here. On SDL we can only disable or enable *all* text input. On Windows the current implementation always takes text input, but only handles IME events and shows an IME after calling `StartTextInput`. I think it's a good idea to have an API like SDL where either we don't handle text input at all or we do handle it **and** enable IME.
  It would be best to be backwards compatible and raise `GameWindow.TextInput` without the user first enabling text input events though. So we can have `GameWindow.StartTextInput` and `GameWindow.StopTextInput` methods, but enable text input by default. Because the default IImmService settings make the application responsible for rendering the IME, at least users in IME-enabled locales won't get an IME window even if the game does not disable text input. It would still be better to disable text input when the game will not handle it, so perhaps we can disable text input in the project templates. Alternatively we could have a breaking change and disable text input by default...

Also, in my tests on Windows the OS-rendered IME window position was not set correctly, does that work for you? It renders at the bottom right of my desktop.

ryancheung:
> * Do we need the `IImmService.TextInput` event? SDL uses the same event for direct TextInput and IME results, I think that makes the most sense for us too (and it would be hard to do it differently on SDL). Shouldn't we just invoke `GameWindow.TextInput` for both direct input and composition results? That means also adding `GameWindow.TextInput` for platforms where it currently does not exist.

I agree also, for backward compability.

> * `StartTextInput` and `StopTextInput` are not consistent on DesktopGL and WindowsDX. We need to decide what behavior we want here. On SDL we can only disable or enable _all_ text input. On Windows the current implementation always takes text input, but only handles IME events and shows an IME after calling `StartTextInput`. I think it's a good idea to have an API like SDL where either we don't handle text input at all or we do handle it **and** enable IME.

I forgot disable handle WM_CHAR message when TextInput is not active. Make it behave as SDL is easy.

> * It would be best to be backwards compatible and raise `GameWindow.TextInput` without the user first enabling text input events though. So we can have `GameWindow.StartTextInput` and `GameWindow.StopTextInput` methods, but enable text input by default. Because the default IImmService settings make the application responsible for rendering the IME, at least users in IME-enabled locales won't get an IME window even if the game does not disable text input. It would still be better to disable text input when the game will not handle it, so perhaps we can disable text input in the project templates. Alternatively we could have a breaking change and disable text input by default...

This seems good, but the only one drawback I have put is: For CJK locales users, as IME is enabled by default, when there's no textbox control focused and user pressed keys like `Ctrl+Space` to  accidendtly enable the IME text composition, then every subsequent key presses would do text compsition with the IME and raise text composition or text input events in the background, this may not affect the gameplay, but just a little weird.
I prefer to disable text input by default. When user put a focus on a textbox control, enable text input, when not textobx control focused, disable text input.

> Also, in my tests on Windows the OS-rendered IME window position was not set correctly, does that work for you? It renders at the bottom right of my desktop.

I got this also. Set position of  the OS-rendered IME window will work or not dependent on what IME software used. It seems to be bugged in the OS IME service, or I'll try to fix that issue with some research on the code in Chromium.

ryancheung:
I want raise another question also: how to update the `ImmService`, create a unified interface by making it a abstract class, or just let it be a interface and update the initialized IME handler instance seperately?

Jjagg:
> For backwar compability, I prefer to retain ImmService.TextInput and relay to GameWindow.TextInput event. Since TextComposition and TextInput are too related to seperate them in two place, I think.

I think we have 3 reasonable options here:
1. Put all new API including TextInput in IImmService and forward events to GameWindow.TextInput.
2. Put IME-specific API in IImmService, but have only GameWindow.TextInput and put Start/StopTextInput in GameWindow too.
3. Expose all IME API on GameWindow.

I think having the same API in two places is confusing and is generally a bad idea. I do think it'd be nice to be able to pass around IImmHandler in user code to handle all text input + compositions though and I agree that it would make sense to have TextInput and TextComposition in the same place.

That said I like option 2 the best. Even if we could redo the API without worrying about breaking changes, I think I'd prefer to have TextInput on GameWindow, because a lot of users won't care about IME configuration and only want to handle text input. Many people don't even know what IME is and if we put TextInput in ImmHandler we're forcing them to learn about it. But maybe that's not a bad idea?

I don't think either of us likes the last option, but I'll put it out there for completeness. Adding all this new API to GameWindow clutters it up way too much.

> This seems good, but the only one drawback I have put is: For CJK locales users, as IME is enabled by default, when there's no textbox control focused and user pressed keys like Ctrl+Space to accidendtly enable the IME text composition, then every subsequent key presses would do text compsition with the IME and raise text composition or text input events in the background, this may not affect the gameplay, but just a little weird.

Yes exactly, this is already happening right now in DesktopGL and it's not great.

> I prefer to disable text input by default. When user put a focus on a textbox control, enable text input, when not textobx control focused, disable text input.

I agree. We'll need some other people to weigh in because this is a breaking change that has the potential to cause a lot of frustration for users. If we do go ahead with this change we should mention it in the 3.8 migration docs (#7117). Also if we do break this it might justify moving TextInput to ImmHandler so we can notify users of the new API with a Deprecated warning saying they have to first enable text input on the ImmHandler.

@tomspilman @harry-cpp @mrhelmut Please read this comment and weigh in on the API and potential breaking change of disabling text input by default.

> I want raise another question also: how to update the ImmService, create a unified interface by making it a abstract class, or just let it be a interface and update the initialized IME handler instance seperately?

Since different platforms IME's behave differently it might make most sense to let the GameWindow implementations handle updating IME's seperately. That gives us a little more control over when Update is called for different platforms and what information is passed. I haven't looked at the implementations so I don't know if that's really necessary, but it is more flexible.

ryancheung:
> Since different platforms IME's behave differently it might make most sense to let the GameWindow implementations handle updating IME's seperately. That gives us a little more control over when Update is called for different platforms and what information is passed. I haven't looked at the implementations so I don't know if that's really necessary, but it is more flexible.

I agree too, it seems more reasonable to have the target platform update their ImmHandler, or not. Actually most of the platform's ImmHandler do not need to be updated.

ryancheung:
> BTW, the sample project are ready to F5-to-run without resolving the package reference, i just reference the packages to a public github package registry.

I'm sorry this doesn't work, github does not allow add Personal Access Token in public repos.

ryancheung:
After testing in macOS, It's true that the TextInput in SDL doesn't support emoji characters, it just drop surrogates. So for supporting emojis, we have to switch to `System.Text.Encoding.Unicode.GetString`, which means we can't avoid to generate Garbage for TextInput, and also for TextComposition.
So in my opinion, we have to disable TextInput/TextComposition by default. Let them to be manually enabled/disabled when required. At least it won't keep generating garbages in every game loop, but just a few game loop when TextInput is enabled.

 BTW, I agree your option 2 also. Non-CJK users need not care about the ImmService or TextComposition.

ryancheung:
I stick to the Option 1, because it is not good to put `Start/StopTextInput` in GameWindow and others in ImmService, the internal implementation of  `Start/StopTextInput` are different from platforms. It is good to just let `Start/StopTextInput` methods stay in ImmService and different platform implementation, like `WinFormImeHander` or `SdlImeHandler`, implement the methods differently. The code is  going to be ugly if we put  `Start/StopTextInput` methods in `GameWindow`.

Jjagg:
> It is good to just let Start/StopTextInput methods stay in ImmService

We can have this and still expose the API on `GameWindow`. I think it's important people that don't want to worry about IME don't have to touch the IMM service instance.

ryancheung:
> > It is good to just let Start/StopTextInput methods stay in ImmService
> 
> We can have this and still expose the API on `GameWindow`. I think it's important people that don't want to worry about IME don't have to touch the IMM service instance.

This way should be easy. I'll add these two method into `GameWindow`.

ryancheung:
I think this PR is about ready. Bugs were fixed, ios text composition events added.
@Jjagg If you think it's fine I'm going to rebase this branch to develop and reopen a final PR.

ryancheung:
@Jjagg We may be able to fix ime with TSF, which is to enable IME by TSF, instead of the IMM way. This is an example: https://github.com/Windmill-City/InputFix It's a mod for Stardew Valley to fix IME issue in Windows 10.

Jjagg:
Oh great! Weird that I didn't come across that API at all when reading about Windows IME support. The msctf.dll comes with the Windows installation, right?

ryancheung:
Yep, after Windows XP(XP has enable TSF manually), Windows has TSF(msctf.dll) by default (even Win 8 TSF was forced so IMM32 based IME would not work, but IMM32 based IME software works in Win 10).

[Here](https://gamedev.stackexchange.com/questions/54436/can-not-enable-ime-to-capture-chinese-characters-in-sharpdx) is a link shows IME not working in sharpdx in Win8 with a non TSF-based IME software.

There is doc link about TSF https://docs.microsoft.com/en-us/windows/win32/tsf/text-services-framework . But it's has rare information. And there's no even single doc page about how to use TFS in Winforms C#(Microsoft don't give a shit about TSF support in Winforms, but they seems integrate TSF in WPF, see: [here](https://referencesource.microsoft.com/#PresentationCore/Core/CSharp/System/Windows/Input/TextServicesContext.cs)). So, to support TFS, we have to do it in the native cpp side. This is how Stardew Valley mod [InputFix](https://github.com/Windmill-City/InputFix) works.

In cpp world, many software do TSF and Imm32 very well, like Chromium, Firefox, etc. And here is the chromium implementation: https://chromium.googlesource.com/chromium/src/+/6373124a3638f6c9f77c7914f78b226163973c68/ui/base/ime/win/ You'll see chromium has `imm32_manager.cc` and `tsf_bridge.cc` to handle TSF IME and IMM32 IME. Except that Chromium or Firefox doesn't render candidate list itself, it just render composition string, set os candidate window position, but I think it's not hard to hide os candidate window and render it manually for fullscreen game.

Here is picture about how TSF and imm32 works with each other:
ime-imm-tsf
In theory, TSF should be compatibale with IMM32, which means enable IME with TSF and detect text composition and result with IMM32 API should work. But not sure if would be actually work without bug.

ryancheung:
So I have an idea to support TSF and IMM32 in MonoGame:
1. Write a native cpp library with the IME support code borrowed from chromium.
2. Write  a C# wrapper to enable/disable IME and receiving IME events, like text composition/result from the native library.
3. Or if enabling TSF also enable IMM32 with no bug, we can handle IME events in `WndProc` in C# side.

Jjagg:
@ryancheung I would prefer not having a native library for this.

I've managed to generate bindings for TSF with [SharpGenTools](https://github.com/SharpGenTools/SharpGenTools). This way we can include a managed assembly or include the binding code directly. I'll push the bindings to a repo when I've done some testing :)

ryancheung:
That sounds great. It would be much simpler to use managed library directly.

ryancheung:
I found ime code in Chromium has lots of docs that describe the TSF and IMM32 API, like [What is TSFTextStore ](https://chromium.googlesource.com/chromium/src/+/6373124a3638f6c9f77c7914f78b226163973c68/ui/base/ime/win/tsf_text_store.h#22), [IMM32](https://chromium.googlesource.com/chromium/src/+/6373124a3638f6c9f77c7914f78b226163973c68/ui/base/ime/win/imm32_manager.h#24), and [How TSF_Bridge makes sure be consistent with IMM32 behavior](https://chromium.googlesource.com/chromium/src/+/6373124a3638f6c9f77c7914f78b226163973c68/ui/base/ime/win/tsf_bridge.h#18)
I think we can consider Chromium as a reference, or mimic its implementation.

ryancheung:
Just found another TSF usage here: https://github.com/microsoft/terminal/tree/master/src/tsf

I've tested that both TSF-based Input Method(Microsoft PinYin) and purely MM32-based Input Method(Sougou PinYin) work every well in Microsoft Terminal.

ryancheung:
WPF core has everything of TSF, e.g. : 

- [UnsafeNativeMethodsTextServices.cs](https://github.com/dotnet/wpf/blob/ac9d1b7a6b0ee7c44fd2875a1174b820b3940619/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/UnsafeNativeMethodsTextServices.cs)
- [DefaultTextStore.cs](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/DefaultTextStore.cs)

But it's not for WinForms.  Pity...

Jjagg:
Great resources! I also used [this](https://web.archive.org/web/20100211233018/http://msdn.microsoft.com/en-us/magazine/cc187238.aspx) blog post to get started with TSF.

Here's my repo with (very incomplete) TSF bindings and the start of an implementation: https://github.com/Jjagg/TsfSharp
It uses [OpenWindow](https://github.com/Jjagg/openwindow) to manage the window, so you'll need to clone that. [Here](https://github.com/Jjagg/OpenWindow/blob/master/src/OpenWindow/Backends/Windows/Win32WindowingService.cs)'s the WndProc method. Though it might be easier to directly call Win32 to create a window.

ryancheung:
Good jobs. I saw many microsoft doc links in Chromium code comments, but most of them are dead. It's good that web archive fix that.

Today I've digged a little into TSF. Through debugging the demo in the Stardew Valley mod `InputFix` repo, I get to known that the ime composition string and result populated by [ITextStoreACP ](https://docs.microsoft.com/en-us/windows/win32/api/textstor/nn-textstor-itextstoreacp) interface. Basically composition string or result text was sent by calling [ITextStoreACP::SetText](https://docs.microsoft.com/en-us/windows/win32/api/textstor/nf-textstor-itextstoreacp-settext) by TSF. It seems TSF provides extra advanced text processing feature(text selection, underline character positions to indicate the composition string) for implementing a TextBox, unlike IMM32, which just give us a cursor position. 

As for the compability or interop of TSF and IMM32, it seems they works independently. Text composition and result events generated in TSF would not be replayed to IMM32, and also events in IMM32 won't affect TSF (I could be wrong). So we may have to implemented TSF and IMM32 separetely. As for IMM32, we can use the current implementation in `WndProc ` method, and for TSF, we could send the text composition and result text from TSF to `WndProc` with `SendMessage` Win32 API.

Some concerns about TSF:
In TSF, text composition and result text are both sent by call the same method in `ITextStoreACP` interface, which is
```
HRESULT SetText(
  DWORD         dwFlags,
  LONG          acpStart,
  LONG          acpEnd,
  const WCHAR   *pchText,
  ULONG         cch,
  TS_TEXTCHANGE *pChange
);
```
Currently I don't know how to distinguish them. TSF expect us to replace the composition string with the `acpStart` and `acpEnd` character positions when the composition generates a result text.

Jjagg:
Check out the TextStore implementation in my repo, there's methods for detecting composition start and end.

ryancheung:
I found something useful again:

[Here](https://bugs.chromium.org/p/chromium/issues/detail?id=963558) is how Chromium support TSF-based and IMM32-based IMEs between different Windows verions.
And [the code](https://github.com/chromium/chromium/blob/ef2500c53aa97a5433371707b78d99f706974a3b/ui/base/ime/init/input_method_factory.cc#L59) to switch input method  based on Windows version.

[Here](https://github.com/chromium/chromium/blob/6efa1184771ace08f3e2162b0255c93526d1750d/ui/base/ime/win/imm32_manager.cc#L492) is a method to detect if the active IME is IMM32-based one or anything else including TSF-based IMEs. We can utilize this method to switch implementations between TSF and IMM32. As in MonoGame, we can use IMM32 implementation in Windows 7 and below, use IMM32 also in Windows 8+ if the active IME is IMM32-based, use TSF for the rest cases.

And I managed to fetch composition and result string with TSF, demo [repo](https://github.com/ryancheung/WindowsForms.IMEHelper.Core). The TSF was implemented in a cpp native library, which is the one in [InputFix] repo (https://github.com/Windmill-City/InputFix), but with my modification.

And to retrieve Candidate List, I believe we can use the [ITfCandidateList](https://docs.microsoft.com/en-us/windows/win32/api/ctffunc/nn-ctffunc-itfcandidatelist) (and its related) intefaces.

ryancheung:
I just managed to [Set the OS Candidate window position](https://github.com/ryancheung/WindowsForms.IMEHelper.Core/blob/tsf/WindowsForms.IMEHelper.Core/IMENativeWindow.cs#L290). But still found no way to hide the OS Candidate list window. Tried return NULL in [ITfContextView](https://docs.microsoft.com/en-us/windows/win32/api/msctf/nf-msctf-itfcontextview-getwnd) but no lucky.

Jjagg:
Good progress! It sounds like it's easiest to do the implementation in your repo and then port it to the managed bindings?

ryancheung:
Yep, we can just implement TSF in the repo right now. It's much easier to just add modification to make TSF work than completely implement TSF ourself. After that was done, we can mimic TSF implementation(just few interfaces are enough) of Chromium and rewrite it with managed bindings.

ryancheung:
@Jjagg I managed to make a runnable IME wrapper around TSF by extracting the TSF code form WPF core. [Here](https://github.com/ryancheung/ImeSharp) is the repo.

ryancheung:
@Jjagg After almost 10days work. I managed to implement a workable IME integration: https://github.com/ryancheung/ImeSharp
Next I'm gonna port ImeSharp to this PR.

Jjagg:
Awesome @ryancheung! Really excited to get this integrated in MonoGame :D

ryancheung:
Hi, I think [ImeSharp](https://github.com/ryancheung/ImeSharp) is ready, with all the feature required for MonoGame. Could you weigh in  please? 
I think there're two options to integrate this lib to MonoGame:
1. Add ImeSharp as a Thirdparty depedency in to MonoGame.
2. Reimplement ImeSharp in MonoGame's repo

I prefer option 1 since I think it's good enough and I intend to make it shared library that can be used in other framwork/engines, like Unity (Unity's IME support is also limited). But I let you weigh in to deside.

After ImeSharp integration is done. I'll continue do update `WinFormsImeHandler.cs` to make use of the ImeSharp API.

Jjagg:
I'm fine with including it as a lib, but pinging some other people to weigh in.
If others prefer including source code directly we can do it the same way we include StbSharp and NVorbis: include the repo as a submodule and include the source code with a wildcard include (only for WindowsDX in this case).

@tomspilman
@harry-cop
@mrhelmut

PR #7119, opened 2020-04-17
Set MG nuget versions in templates to match template build version

Jjagg:
This adds a task that sets the `PackageReference` version for the MonoGame NuGet references in the template projects.


Jjagg:
@harry-cpp This is good to go

Issue #7115, opened 2020-04-16
Add Sourcelink support to nuget packages


CloudStu:
Would it be possible to add sourcelink support to the MonoGame nuget packages? With sourcelink, one could have the MonoGame source code downloaded automatically from github when stepping into MonoGame code in a supported IDE like Visual Studio.

https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink
https://github.com/dotnet/sourcelink

tomspilman:
Seems like a great idea!  @Jjagg  @harry-cpp ?

Jjagg:
Yeah, we should do that.

PR #7125, opened 2020-04-22
Add SourceLink support to MonoGame.Framework

Jjagg:
Requested in #7115. 

I have not yet tested this. I think it'll only work for DesktopGL and WindowsDX because by default non-SDK-style projects generate full PDB symbols which is not compatible with SourceLink.

This includes the pdb in the NuGet packages.
Builds for NuGet should NOT include the pdb, instead a seperate snupkg should be built and published to the NuGet symbol server to distribute the symbols.
See https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg.

A snupkg can be generated by setting
```
<IncludeSymbols>true</IncludeSymbols>
<SymbolPackageFormat>snupkg</SymbolPackageFormat>
 ```

How do we best handle this? I think it'd be nice to let the build script take an argument that indicates if the build should be published to the develop feed or to nuget.org. It can then set an MSBuild property based on that and we can conditionally choose to include symbols or to build a .snukpg.

### DIFF
--- a/MonoGame.Framework/Directory.Build.props
+++ b/MonoGame.Framework/Directory.Build.props
@@ -14,5 +14,13 @@
     <LangVersion>5</LangVersion>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
+  </ItemGroup>
 </Project>
 


tomspilman:
@harry-cpp @mrhelmut  Is this worth merging?

https://github.com/dotnet/sourcelink

tomspilman:
No one responded to this... so i am closing it.

Apostolique:
I add sourcelink to my libraries. It does make debugging be much nicer since I can step through even the library code.

lordzsolt:
@harry-cpp @mrhelmut Can we get a response on this Pull Request?

It's significantly better DevX to be able to jump to source code directly from the editor, as opposed to looking at decompiled code, or having Monogame GH page open in the browser.

There was already an existing PR, that ended up being closed because of lack of response:
https://github.com/MonoGame/MonoGame/pull/7863

AristurtleDev:
> It's also very annoying that all the documentation comments are missing:

In context of the missing documentation, this was fixed in #8049, so next release (and current `develop` nugets) have the documentation xml correctly packed.

In the mean time, I also provided a workaround for those who want it with the current NuGets
https://github.com/AristurtleDev/monogame-xml-documentation

AristurtleDev:
MonoGame was updated to dotnet 8 in the `develop` branch with PR #8089 

Per SourceLink documentation at https://github.com/dotnet/sourcelink?tab=readme-ov-file#using-source-link-in-net-projects

> Starting with .NET 8, Source Link for the following source control providers is included in the .NET SDK and enabled by default:
> -  [GitHub](http://github.com/) or [GitHub Enterprise](https://enterprise.github.com/home)
> - [Azure Repos](https://azure.microsoft.com/en-us/services/devops/repos) git repositories (formerly known as Visual Studio Team Services)
> - [GitLab](https://gitlab.com/) 12.0+ (for older versions see [GitLab settings](https://github.com/dotnet/sourcelink?tab=readme-ov-file#gitlab))
> - [Bitbucket](https://bitbucket.org/) 4.7+ (for older versions see [Bitbucket settings](https://github.com/dotnet/sourcelink?tab=readme-ov-file#bitbucket))

So we no longer need to add this package as it's already included in the SDK

SimonDarksideJ:
Closing as provided by .NET 8

PR #7130, opened 2020-04-23
Move content building before PrepareForBuild so it works for iOS

harry-cpp:
iOS uses BundleResource action which is calculated before the BuildDependsOn action (like our targets is), it worked by sheer luck before...


ryancheung:
I've just tested this PR and it works now, the issue is fixed.

PR #7131, opened 2020-04-24
Adding Github Sponsorship

CartBlanche:


tomspilman:
@Jjagg @harry-cpp @dellis1972 

So the idea here would be to have individual sponsor links various project maintainers?

This may be a good first step?  Add @Jjagg and @harry-cpp if they can get approval to be in the sponsors program.

Sponsor pages are like a mini patreon:  https://github.com/sponsors/ThePhD

Ideally the next step is getting MonoGame to be a sponsor like this.

CartBlanche:
Should we go with just 1 for now and then add people as they get approved?

tomspilman:
We now have the Foundation for this.

PR #6908, opened 2019-10-17
Can't hide mouse cursor #6180

FrancescoBonizzi:

Issue #6180, opened 2018-02-10
Can't hide mouse cursor


mikea314:
Setting IsMouseVisible = false doesn't hide the mouse cursor

MonoGame.Framework.WindowsUniversal
3.7.0.1248-develop

cra0zy:
Hmm, I can't seem to reproduce this one.

mikea314:
Thanks for taking a look cra0zy.
It's still happening for me on the latest Dev build.
Initially, the mouse cursor is hidden, but the moment I move the mouse it re-appears.

cra0zy:
Can you make a minimal example that reproduces the issue?

mikea314:
Will do

mikea314:
Here's a link to the download.
https://www.dropbox.com/s/ke3vgkh6zs0ch8n/MouseTest.MG.zip?dl=1

cra0zy:
Oh, I see whats causing the bug...

cra0zy:
I don't have time to dig into this right now, but the bug is getting caused by the Xaml template, however it is not getting caused by the Core template.

mikea314:
No hassles, thanks for having a look

Jjagg:
Fixed in #6180.

PR #7134, opened 2020-04-27
Revert "Can't hide mouse cursor #6180 (#6908)" because of win32 threading crash

mrhelmut:
Hello :)

#6908 introduced a threading race condition that makes win32 API calls to crash (e.g. loading assets).

Also, ```IsMouseVisible = false;``` is working without this PR as of Windows 10 SDK 1903 (and targeting Fall Creator Update at minimum), which is the intended baseline of MonoGame.

So I guess that we're good as-is and this issue doesn't need to be addressed for older targets.

Cheers!

### DIFF
--- a/MonoGame.Framework/Platform/WindowsUniversal/UAPGameWindow.cs
+++ b/MonoGame.Framework/Platform/WindowsUniversal/UAPGameWindow.cs
@@ -334,7 +334,7 @@ namespace Microsoft.Xna.Framework
                 return;
 
             var asyncResult = _coreWindow.Dispatcher.RunIdleAsync( (e) =>
- {
+            {
                 if (visible)
                     _coreWindow.PointerCursor = new CoreCursor(CoreCursorType.Arrow, 0);
                 else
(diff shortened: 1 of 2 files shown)


harry-cpp:
Have you tried with xaml template?

mrhelmut:
No, I have only tested CoreApp. I'll check that later.

If ```IsMouseVisible``` doesn't work with XAML, I'll look into fixing the race condition.

harry-cpp:
> If `IsMouseVisible` doesn't work with XAML

That was the bug the original PR was trying to solve.

mrhelmut:
Yeah, I just went through the original issue.

It maybe fixed the cursor in XAML, but it made CoreApp to just not work anymore because ```_coreIndependentInputSource.Dispatcher.TryRunAsync``` is called when it shouldn't.

mrhelmut:
@harry-cpp Since resolving the actual race condition isn't required on core app for mouse cursor visibility to work, I've simply re-reverted the PR and added a null check for core app to just work.

MonoGame.Framework/Platform/WindowsUniversal/UAPGameWindow.cs
```diff
@@ -334,7 +334,7 @@ internal void SetCursor(bool visible)
                 return;
 
             var asyncResult = _coreWindow.Dispatcher.RunIdleAsync( (e) =>
-            {
+ {
```
harry-cpp, line 337: @mrhelmut code formatting

PR #7141, opened 2020-05-10
Fixed a bug where The "New Item" dialogue would have no templates listed

TheBoneJarmer:
When I pull the latest commits from the develop branch and run or debug the MonoGame.Content.Builder.Editor.Windows project, I was not able to add new items to my content file using the NewItemDialogue dialogue as the templates could not be found. This resulted in a NullReferenceException when I tried closing the dialogue without having anything selected.

I am not certain what the original piece of code is supposed to do, hence I commented it out instead of removing it. Any comments on that matter would be highly appreciated.

### DIFF
--- a/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
@@ -112,13 +112,9 @@ namespace MonoGame.Tools.Pipeline
             {
                 LoadTemplates(macPath);
             }
-            else if (Directory.Exists(windowsAndLinuxPath))
-            {
-                LoadTemplates(windowsAndLinuxPath);
-            }
             else
             {
-                throw new DirectoryNotFoundException("'Templates' or 'Resources' folder not found");
+                LoadTemplates(windowsAndLinuxPath);
             }
 
             UpdateMenu();
--- a/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
@@ -105,19 +105,22 @@ namespace MonoGame.Tools.Pipeline
 
             _templateItems = new List<ContentItemTemplate>();
             var root = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var macPath = Path.Combine(root, "..", "Resources");
+            var windowsAndLinuxPath = Path.Combine(root, "Templates");
 
-            // I do not understand why this piece of code is here. Is there a possibility where the "Templates" folder is actually called "Resources"?
-            // What I do know is that I don't have a folder called "Resources" in this commit, so to prevent the path to be renamed to "Resources" I commented it out instead of removing it
-            //if (Directory.Exists(Path.Combine (root, "..", "Resources", "Templates")))
-            //{
-            //    root = Path.Combine(root, "..", "Resources");
-            //}
+            if (Directory.Exists(macPath))
+            {
+                LoadTemplates(macPath);
+            }
+            else if (Directory.Exists(windowsAndLinuxPath))
+            {
+                LoadTemplates(windowsAndLinuxPath);
+            }
+            else
+            {
+                throw new DirectoryNotFoundException("'Templates' or 'Resources' folder not found");
+            }
 
-            // It seems obvious what this is supposed to do, but the resulting path, provided as parameter, does not exist in my case
-            //LoadTemplates(Path.Combine(root, "../../../Templates"));
-
-            // This way the templates folder can be found in my case
-            LoadTemplates(Path.Combine(root, "Templates"));
             UpdateMenu();
 
             view.UpdateRecentList(PipelineSettings.Default.ProjectHistory);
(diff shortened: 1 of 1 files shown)


harry-cpp:
The original if is there for the Mac mgcb-editor, as its a .App app that has Templates copied in a directory inside it.

TheBoneJarmer:
I see, so theoretically that could mean this code wont work on mac, right?

harry-cpp:
Yup. You need that if code above.

Jjagg:
@harry-cpp  The `../../../Templates` check is for Mac? Shouldn't it use `../../..` as the root then, instead of `../../../Templates`? It seems inconsistent with the other case where `../Resources` is used as the root.

It seems the Templates folder is copied to output root in the new packaging scheme ([link](https://github.com/MonoGame/MonoGame/blob/develop/Tools/MonoGame.Content.Builder.Editor/MonoGame.Content.Builder.Editor.targets#L35-L37)), so this PR makes it look in the right place. Do we still need the `../../../Templates` path?

harry-cpp:
Templates are broken on Mac with or without this PR then.

TheBoneJarmer:
Since I own no Mac I cannot test any code changes I make for it. So if you guys are ok with I would like to suggest something. What if you create a feature or bugfix branch from develop and have me update the PR to merge to that branch instead of develop. That way you can alter my changes to fit a Mac folder structure.

Jjagg:
Okay, so if it exists we need `root/Templates` and otherwise it's `root/../../../Templates` ? They're never put in `root/Resources`?

@jnoyola Do you now if this is correct?

harry-cpp:
All templates are in .app/Contents/Resources, just checked.

jnoyola:
I’ll have to take a closer look later today. It’s quite possible I broke templates. I can also test on Mac tonight.

jnoyola:
So it actually does currently work on Mac, but due to the `../../../Templates` path finding the main Templates folder, not the Resources check finding it. So the templates currently exist in 2 places, thanks to [Eto packing all `Contents` in the `Resources` folder](https://github.com/picoe/Eto/blob/develop/src/Eto.Mac/build/Mac.targets#L232-L234):

* `/Templates/`
* `/mgcb-editor-mac.app/Contents/Resources/`

So Windows and Linux (with dlls in `/`) should use `Path.Combine(root, "Templates")` like this PR has it.
And Mac (with dll in  `/mgcb-editor-mac.app/Contents/MacOS`) should check and use `Path.Combine(root, "..", "Resources")` if it's present.

So @TheBoneJarmer could you please remove the commented part and update it to check `Path.Combine(root, "..", "Resources")` and use that if it exists, and otherwise use `Path.Combine(root, "Templates")` as you have it?

harry-cpp:
> @harry-cpp Also side question: Is Eto intentionally ignoring the folder structure to place them in `Resources` and not `Resources/Templates`? That seems like a bug in my opinion.

We would need to check what Xamarin Mac does to know.

TheBoneJarmer:
Hi @jnoyola 

I just pushed my commit. Added an exception to it as well in case none of the folders could be found since the application was crashing when trying to add a new item. This way at least the user knows why. :)

Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
```diff
@@ -105,11 +105,22 @@ private PipelineController(IView view)
...
+            }
+            else if (Directory.Exists(windowsAndLinuxPath))
             {
-                root = Path.Combine(root, "..", "Resources");
+                LoadTemplates(windowsAndLinuxPath);
             }
-            LoadTemplates(Path.Combine(root, "../../../Templates"));
+            else
+            {
+                throw new DirectoryNotFoundException("'Templates' or 'Resources' folder not found");
```
harry-cpp, line 121: We don't want to crash the application if the file templates are missing XD

TheBoneJarmer: Ok, fair enough. Will remove it.

TheBoneJarmer:
Hi @harry-cpp 

I modified my code again. Removed the exception and changed it to a if/else statement.

jnoyola:
Looks good to me!

Being nitpicky, I'd say maybe `windowsAndLinuxPath` should just be `defaultPath` or something, because if you're splitting the code by OS then you should really be using `#if`, which is potentially even better, since we already compile assemblies separately for each platform. But that's super minor and I'm not a maintainer anyway 😛

TheBoneJarmer:
> EDIT: regarding not crashing the app if templates aren't available, I believe this just restores the original functionality where it seems to crash when you hit Cancel if they're not present. It's not the purpose of this PR, but we could go one step further to make sure that original crash doesn't occur if the templates do ever get misplaced again.

I agree, for that reason I added the exception. But if it is not being caught somewhere it crashes anyway so I understand @harry-cpp too. But I agree with your idea for a check. Perhaps adding an exception anyway and let the GUI show a message box? Just sprouting ideas here. One could also say that no action has to be taken as the application has to expect the folder being there. If not, something went wrong with either the installation, the installer or the release and that would be a bug.

jnoyola:
Maybe I didn't explain it well, but I'm in agreement with Harry. We don't want it to crash so we shouldn't throw an exception. I'm saying what would be even better would be to go the extra step in that direction and also prevent the current exception from being thrown. But you're right that this should only occur in the event of something going wrong with the installation, so maybe it's not a concern.

TheBoneJarmer:
Hi @harry-cpp 

It has been a while since I pushed my last commit. Ever since I have not heard back. As this is my first pull request ever to a repo on GitHub, is there anything I need to check/click/activate in order for you to continue?

harry-cpp:
Sorry for the delay, everything looks good, merging!

PR #7144, opened 2020-05-13
Add Web platform information to the content builder

harry-cpp:
Both my WebAssembly and Bridge.NET versions can use all the content DesktopGL spits out, except for music, which needs to be in mp3 format for the most wide range support. This just adds the needed bits for the content part.


harry-cpp:
@Jjagg merge please :)

PR #7148, opened 2020-05-19
Effect Compiler Fixes

harry-cpp:
- Shader build tests enabled for both Windows and macOS
- EffectProcessor is now using source file instead of bytecode
- Restored deleted code for effect error collecting in EffectProcessor
- Fixed selecting target platform for shaders

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs
@@ -50,45 +50,107 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
         public override CompiledEffectContent Process(EffectContent input, ContentProcessorContext context)
         {
             var mgfxc = Path.Combine(Path.GetDirectoryName(typeof(EffectProcessor).Assembly.Location), "mgfxc.dll");
-            var sourceFile = Path.GetTempFileName();
+            var sourceFile = input.Identity.SourceFilename;
             var destFile = Path.GetTempFileName();
-            var target = context.TargetPlatform.ToString().Contains("Windows");
-            
-            File.WriteAllText(sourceFile, input.EffectCode);
-
-            var proc = new Process();
-            proc.StartInfo.FileName = "dotnet";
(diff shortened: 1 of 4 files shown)


MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs
```diff
@@ -50,45 +50,112 @@ public EffectProcessor()
...
-            var success = proc.ExitCode == 0;
+            var success = ExternalTool.Run("dotnet", arguments, out stdout, out stderr) == 0;
             var ret = success ? new CompiledEffectContent(File.ReadAllBytes(destFile)) : null;
 
-            File.Delete(sourceFile);
             File.Delete(destFile);
 
+            ProcessErrorsAndWarnings(!success, stderr, input, context);
+
             if (!success)
```
tomspilman, line 72: @harry-cpp - this block can be removed.  `ProcessErrorsAndWarnings` handles the failure case.

Tools/MonoGame.Effect.Compiler/Program.cs
```diff
@@ -13,7 +13,7 @@ public static int Main(string[] args)
         {
             if (!Environment.Is64BitProcess && Environment.OSVersion.Platform != PlatformID.Unix)
             {
-                Console.Error.WriteLine("The MonoGame content tools only work on a 64bit OS.");
+                Console.WriteLine("The MonoGame content tools only work on a 64bit OS.");
```
tomspilman, line 16: @harry-cpp what was the idea here?

The reason i had these going to stderr was so that errors appear in the pipeline output.  Removing this means if for whatever reason these fire when running mgfxc from the pipeline then we'll have a failure with no error message.

On Windows at least console output shows both stdout and stderr.

MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs
harry-cpp: That is the only position where `ProcessErrorsAndWarnings` is getting called from. I've also modified mgfxc to output warnings and errors to stderr, while other user related information to stdout.

Issue #7160, opened 2020-05-24
Darken message text in filtered build output


tomspilman:
The text in the filtered build output in the MGCB Editor is very hard to read...

![image](https://user-images.githubusercontent.com/103522/82759562-dcf56d00-9db3-11ea-997e-99c874424ac9.png)

We should darken this text or just make it the default text color.  The text is generally well formatted so color of the text could likely be normal.

jojotrujillo:
Hey, @tomspilman! Could I take a crack at this? I've overviewed the Contributing file and I think I could handle this issue with my skillset. :smile:

harry-cpp:
Why of course, go right ahead :)

jojotrujillo:
I forgot to use one of those keywords to automatically close an issue. For future reference, do y'all prefer to close issues manually, or is it okay to automatically close an issue on merge?

tomspilman:
@jojotrujillo - if you use the magic `Fixed #number` that works for us.

Going to close this now.

PR #7163, opened 2020-05-25
Remove Color object's (DisabledTextColor) alpha assignment

jojotrujillo:
#7160 Removing the alpha assignment darkens the text.

I had a thought to refactor since one value, `SystemColors.ControlText`, is getting assigned to two variables now, but I'm unsure about how much the `disabled` distinction plays into other methods.

What do you all think?

### DIFF
--- a/Tools/MonoGame.Content.Builder.Editor/Controls/BuildItem.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/Controls/BuildItem.cs
@@ -64,7 +64,7 @@ namespace MonoGame.Tools.Pipeline
                 if (_expanded)
                 {
                     _descriptionOffset = (_descSize - DrawInfo.TextHeight) / 2;
-                    Height = (int)(CellHeight + _descSize * _description.Count);
+                    Height = (int)(CellHeight + Margin + (_descSize * _description.Count));
 
                     foreach (var des in _description)
                     {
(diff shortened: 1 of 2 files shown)


tomspilman:
@harry-cpp thoughts?  I don't know how many other places "disabled" text appears in the UI.

IMO maybe we need to both:

  - Reduce disabled text color to `0.6` or so.
  - Consider changing the code [here](https://github.com/MonoGame/MonoGame/blob/develop/Tools/MonoGame.Content.Builder.Editor/Controls/BuildItem.cs#L111) to use the regular text color.

@jojotrujillo maybe make some adjustments and post some screenshots here?

harry-cpp:
The correct solution is to modify BuildItem.cs:111

jojotrujillo:
Thanks for the advice, y'all! I'll make the changes and post some screenshots. 😃

jojotrujillo:
build-output
The build output text has the regular text color

tomspilman:
@jojotrujillo - that for sure looks better.

For bonus points here see if you can balance the margin here:

image

See how the ones i marked on the left have a few pixel margin from the top/bottom to the content.  But not the one i marked on the right side bottom?  I assume this only occurs when it is expanded.  Should be in the same place you just changed the code.

jojotrujillo:
build-output-gap

tomspilman:
That looks much better to me!  @harry-cpp @Jjagg ?

Jjagg:
LGTM!

harry-cpp:
Looks good, thanks @jojotrujillo, merging!

PR #7164, opened 2020-05-26
Omit mention of thrown exception in GamePad.MaximumGamePadCount

tdeeb:
This PR improves the documentation for `GamePad.MaximumGamePadCount` by omitting the point about throwing an exception, as it doesn't do so on any public platforms. Will require confirmation if it actually does on private ones.


Jjagg:
Good catch, thanks!

Issue #7150, opened 2020-05-22
dotnet tool `dotnet-mgcb` to be reinstalled every time when project build.


ryancheung:
When trying to build the project every time, it tries to install the dotnet tool `dotnet-mgcb`, it connects to the configured nuget source which dramatically reduces the build speed if you have a bad network connection to the nuget server.

The following is the log when project starts build:
```
  dotnet-mgcb      3.7.200-develop      mgcb      D:\xxx\xxx\xxx.Playground.WindowsDX\.config\dotnet-tools.json
  Creating this template will make changes to existing files:
    Overwrite   .config/dotnet-tools.json

  Rerun the command and pass --force to accept and create.
C:\Users\ryanc\.nuget\packages\monogame.content.builder.task\3.7.200-develop\build\MonoGame.Content.Builder.Task.targets(105,5): warning MSB3073: Command“dotnet new tool-manifest”exited, exit code is -2147467260。 [D:\xxx\xxx\xxx.Playground.WindowsDX\xxx.Playground.WindowsDX.csproj]
```

NOTE I'm use a custom preview version number `3.7.200-develop`(which is actually the MonoGame develop branch). Maybe it's bugged on the version string detection?

@jnoyola Could you please give some favor on this? It's really hard to debug with `MonoGame.Content.Builder.Task.targets`

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL, WindowsDX, Android

jnoyola:
That’s a good point. I can jump into the detail when I’m off mobile lately, but I think the idea is that it will check what version you intend to install based on your nuget reference or a property if you override it. It will then check the manifest for that version. If it’s not present it will install it. If it is present, it _always_ does a restore because it’s hard to tell if the tool has actually been installed or just written in the manifest.

I can look into ways of improving this and check out your error later. Please share if you have any ideas :)

jnoyola:
Can you give more details about that log message? It looks like it’s trying to use 3.7.200-develop, which appears to be the same version as your nuget package. Is there an issue with the version there, or is the issue that it’s asking you to call the command with —force?

ryancheung:
There should be no issue with the version, it's just a custom version number. And I use the same version number for all MG nuget packages. 
[...]

PR #7169, opened 2020-05-28
Decouple Builder.Task from dotnet-mgcb

jnoyola:
## Summary

I know this change hasn't really been discussed, but I think it should be considered. It decouples `Builder.Task` from the `dotnet-mgcb` global tool and packs a copy of `mgcb` directly into `Builder.Task` (just as `Builder.Editor` already does). Now `dotnet-mgcb`, `Builder.Editor`, and `Builder.Task` each ship with their own copy of `mgcb`.

Now the Task is much simpler, and the dotnet global tool is still there if you want it.

**Cons:**
* `Builder.Task` doesn't automatically install the `dotnet-mgcb` tool for you if you wanted to run it manually
* Two copies of `mgcb` will be saved, taking up twice as much space (~300MB extra) if a user wants to use the Task _and_ the dotnet tool.

**Pros:**
* No version issues like #7150
* No potential issues from parsing `dotnet tool` output
* No .config folders strewn across your project when you don't want them
* Easier to understand and maintain targets file (look at how much was deleted from `MonoGame.Content.Builder.Task.targets`)
* Easier to work from source without needing to first publish the dotnet tool to a local NuGet source (see [sample below](https://github.com/MonoGame/MonoGame/pull/7169#issuecomment-635205749))

## Validation

* Tested building a game with `Builder.Task` and running the output with each of the following
  * WindowsDX
  * WindowsUniversal
  * DesktopGL
  * iOS
* Tested building a .mgcb file with each of the following dotnet local tools
  * `dotnet-mgcb`
  * `dotnet-mgcb-editor`

### DIFF
--- a/Tools/AppReference.targets
+++ b/Tools/AppReference.targets
@@ -7,7 +7,7 @@
             See https://github.com/dotnet/sdk/issues/1675#issuecomment-346135772 -->
   <Target Name="CopyAppReferences" BeforeTargets="GetCopyToOutputDirectoryItems">
 
-    <!-- Get the output file of this AppReference. -->
+    <!-- Get the output folders of all AppReferences. -->
     <MSBuild Projects="%(ProjectReference.FullPath)"
              Targets="GetTargetPath"
              Properties="Configuration=$(Configuration)"
@@ -41,9 +41,16 @@
 
     <!-- Copy this content to the build and package output. -->
     <ItemGroup>
-      <TfmSpecificPackageFile Include="%(AppReferenceContent.FullPath)">
-        <PackagePath>tools\$(TargetFramework)\any\%(AppReferenceContent.SubDir)%(RecursiveDir)</PackagePath>
-      </TfmSpecificPackageFile>
+      <ContentWithTargetPath
+        Include="%(AppReferenceContent.FullPath)"
+        TargetPath="%(AppReferenceContent.SubDir)%(RecursiveDir)%(Filename)%(Extension)"
+        CopyToOutputDirectory="PreserveNewest" />
+
+      <!-- This is required to include the files in a nupkg if the project isn't packed as a tool. -->
+      <TfmSpecificPackageFile
+        Include="%(AppReferenceContent.FullPath)"
+        PackagePath="tools\$(TargetFramework)\any\%(AppReferenceContent.SubDir)%(RecursiveDir)"
+        Condition="'$(PackAsTool)' != 'true'"/>
     </ItemGroup>
 
   </Target>
--- a/Tools/AppReference.targets
+++ b/Tools/AppReference.targets
@@ -7,7 +7,7 @@
             See https://github.com/dotnet/sdk/issues/1675#issuecomment-346135772 -->
   <Target Name="CopyAppReferences" BeforeTargets="GetCopyToOutputDirectoryItems">
 
-    <!-- Get the output folders of all AppReferences. -->
+    <!-- Get the output file of this AppReference. -->
     <MSBuild Projects="%(ProjectReference.FullPath)"
              Targets="GetTargetPath"
              Properties="Configuration=$(Configuration)"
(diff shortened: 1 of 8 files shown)


tomspilman:
@Jjagg and @harry-cpp can make the call on this.  @mrhelmut may have an opinion too.

I personally never use the task, so don't really care either way.  

But I will say that if you are working from source it would be good that this still work correctly.

jnoyola:
> But I will say that if you are working from source it would be good that this still work correctly.

That's a good point, and I think that's an additional benefit of this approach. If you decide to reference the projects from source, the previous approach would still try to install the dotnet tool from NuGet. With the new approach, you can reference everything locally like usual:

```xml
  <!-- Reference projects -->
  <ItemGroup>
    <ProjectReference Include="C:\path\to\MonoGame\MonoGame.Framework\MonoGame.Framework.DesktopGL.csproj" />
    <ProjectReference Include="C:\path\to\MonoGame\Tools\MonoGame.Content.Builder.Task\MonoGame.Content.Builder.Task.csproj" />
  </ItemGroup>

  <!-- Set properties -->
  <Import Project="C:\path\to\MonoGame\Tools\MonoGame.Content.Builder.Task\MonoGame.Content.Builder.Task.props" />
  <PropertyGroup>
    <MGCBPath>C:\path\to\MonoGame\Artifacts\MonoGame.Content.Builder.Task\$(Configuration)\netcoreapp3.1\mgcb.dll</MGCBPath>
  </PropertyGroup>

  <!-- Import targets -->
  <Import Project="C:\path\to\MonoGame\MonoGame.Framework\MonoGame.Framework.DesktopGL.targets" />
  <Import Project="C:\path\to\MonoGame\Tools\MonoGame.Content.Builder.Task\MonoGame.Content.Builder.Task.targets" />
```

Jjagg:
I think the biggest downside is users will have to have the installation of a single version of mgcb on disk twice. But this is much much much less prone to bugs and easier to maintain, so I think we can make that trade-off.

harry-cpp:
For now this solution might be better.

tomspilman:
> have the installation of a single version of mgcb on disk twice.

MGCB is fat.... it is 95MB nupkg...  260MB unpacked.  Most of that is unused code in ffmpeg...

image

All the circled bits are ffmpeg overhead.

If we customized our ffmpeg build to just have the formats and support we need...  it could be much smaller i think.

That said...  do we care about 300MB when a crazy fast 2TB SSD is $300?

jnoyola:
> users will have to have the installation of a single version of mgcb on disk twice

Absolutely. I actually intentionally left that off this list because I think I made the same argument on another PR or the original design and Tom made a similar argument about how we don’t need to worry so much about storage on dev machines :P But I’ll add it for completeness.

Jjagg:
> I personally don’t see that many people talk about calling mgcb directly.

No, but people use mgcb-editor and that also includes mgcb.

jnoyola:
> No, but people use mgcb-editor and that also includes mgcb.

If we had a good way to get the Task and mgcb-editor to share a copy of mgcb, then it _might_ be worth it, but they already don’t because we saw too many issues with it before.

jnoyola:
[Here](https://community.monogame.net/t/problem-with-fbximporter-in-3-8-0-1375-develop-build/12777/4) is another issue that seems like it could be at least somewhat mitigated by this change.

The entire problem hasn't been identified yet, but part of the issue is that different parts of the command require different working directories.
* `dotnet` needs the working directory to be the project folder so it can find the tool manifest
* The MGCB tool itself needs the working directory to be the tool directory so Assimp can find it's DLLs.

harry-cpp:
That would not get mitigated with this.

jnoyola:
It would need another small change to actually fix the working directory, but I think the point is clear that there is another class of issues caused by differing requirements for working directory.

Are you simple pointing out that this PR doesn't actually change the working directory, or are you actually saying that that issue is unrelated?

harry-cpp:
Changing the working directory is not a proper fix for that.

jnoyola:
> Changing the working directory is not a proper fix for that.

No argument from me there, but that's why I only said mitigated and not really solved. I'm just trying to chisel down the surface area of where things might go wrong. 🙂

harry-cpp:
I'm gonna merge this as there was no opposition. Thanks @jnoyola !

Random question, are you on MG discord server / do you use discord?

jnoyola:
Thanks! Not sure if we want to mention in the 3.8 preview thread that the Task installation is changing and people should delete uninstall the tool and delete those .config folders if they intend to manually use the tool.

I do use discord but I'm not on the MG server yet. I'll check it out today!

PR #7173, opened 2020-05-29
Splitting assingment in ZlibStream that cause AoT issues

diogoriba:
When generating native code in some AoT compilers, this line causes the type of the expression within the if statement to get incorrectly interpreted. splitting this line into two different statements fixes this issue without changing the expected funcionality.

Existing tests should continue to exercise the portions of this code that were already tested.


tomspilman:
Thanks @diogoriba

RUSshy:
if it is valid code, it should work with the AOT compiler

so it is a bug in the AOT compiler, please don't forget to report that bug to the AOT compiler maintainers

tomspilman:
> if it is valid code, it should work with the AOT compiler

@RUSshy - i am the AOT compiler maintainer in question.  I am aware. ;)

PR #7176, opened 2020-06-02
Changed position for contextmenu item Rebuild

### LINKED ISSUE #7157: Move "Rebuild" up in context menu
We should move up the "Rebuild" command in the MGCB Editor context menu....

image

It is almost always the command i'm looking to execute and it is closer to the bottom of the menu than the top.

I would say it belongs as the top menu entry.

AndiAn94:
Moved the Rebuild Item to a new position see Issue #7157 

rebuild

Fixes #7157


tomspilman:
That looks good to me.... @harry-cpp ?

Maybe worth considering if Add should be below it?

harry-cpp:
> Maybe worth considering if Add should be below it?

Nah.

harry-cpp:
Thanks @AndiAn94!!! Merging!!!

PR #7180, opened 2020-06-03
Fix project dependencies

harry-cpp:


harry-cpp:
@Jjagg You missed one :P

Issue #7089, opened 2020-04-03
On external online dependencies


mrhelmut:
Hello there,

Following #7086 and #7088 (which are very welcome moves), I see some nuget gymnastic going on and I thought it would be interesting to open a question over how to manage dependencies (both internal and external), just for future reference.

On some projects I've worked on, there was an architectural policy of "no online dependencies in the repository" because it was a requirement to be able to deploy new build environments in a context with limited access to the internet (some game companies have build stations cut from the internet to avoid breaches), to prevent any failures (e.g. internet outages, nuget.org outages, external dependencies deprecating or being removed), to handle third parties (e.g. private nugets), to provide a standalone repository (i.e. all dependencies are internal and build out-of-the-box), or simply to avoid starting a dependency hell.

Most of these concerns can be handled by setup'ing a nuget cache local to the repository, like having all nugets pushed to the repo with a setup file for MSBuild/Visual Studio (Nuget.Config):
```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
    <packageSources>
        <add key="local-packages" value="../offline_nugets" />
    </packageSources>
</configuration>
```

In the case of #7086, this is fine.

In the case of #7088, I don't know if Cake can fetch from a local source? (I assume it uses MSBuild for its setup and can read the config while restoring packages)

Independently of these PRs, the question I'd like to open here, is whether we should endorse external online references or not?

My personal concern is that starting to allow them exposes MonoGame little by little to adding any kind of nuget references everywhere and making the project less and less resilient to being built offline/out-of-the-box.

What are your thoughts?

@tomspilman @Jjagg @harry-cpp

mrhelmut:
> Since users have to download the MG packages anyway at some point, the requirements for an internet connection do not change by switching to NuGets.

It does change things when it comes to administrating an offline environment with a package manager. Calling your administrator every time you need a new package to be added to the offline repo can be complicated at times (especially when some cascading deps are resolved at build time and you couldn't figure that prior to calling the administrator the first time).

My question is not at all against nuget or any package manager, but it is about being cautious about somehow hidden or cascading dependencies that can hardly been figured out if anyone wants to setup an offline repo at some point.

It's not just about "downloading once".

Jjagg:
> Calling your administrator every time you need a new package to be added to the offline repo can be complicated at times (especially when some cascading deps are resolved at build time and you couldn't figure that prior to calling the administrator the first time).

So this is about trust/security rather than availability?

[...]

PR #7184, opened 2020-06-05
Use AssimpNet from nuget

harry-cpp:
There is this nifty AssimpNet implementation made by "Nicholas Woodfield": https://bitbucket.org/Starnick/assimpnet/src/master/

It comes with prepackaged native libs and is available on nuget, been meaning to use it for a while now :)

PS. He also has another nifty library wrapper for textures: https://www.nuget.org/packages/TeximpNet/ , we might wanna look into using that and replace our current FreeImage and NVTT stuff.


tomspilman:
We can likely kill our fork of it here:  https://github.com/MonoGame/assimp

Also can remove the build scripts from the Teamcity server too.

Note we may need to add more model processing unit tests in the near future to help ensure Assimp keeps working for us.

mrhelmut:
You can likely expect me to butt in about this every time :D #7089

PR #7186, opened 2020-06-06
[MGCB Editor] macOS fixes

harry-cpp:
- Fixed register/unregister on macOS
- Fixed MG Addin for VS for Mac not calling MGCB Editor properly
- Fixed MGCB Editor versioning on macOS

### DIFF
--- a/Templates/VisualStudioForMac/PipelineDisplayBinding.cs
+++ b/Templates/VisualStudioForMac/PipelineDisplayBinding.cs
@@ -25,12 +25,15 @@ namespace MonoGame.IDE.VisualStudioForMac {
 
 	}
 
-	class PipelineDesktopApplication : MonoDevelop.Ide.Desktop.DesktopApplication {
+	class PipelineDesktopApplication : MonoDevelop.Ide.Desktop.DesktopApplication
+	{
+		private readonly static string appPath = Path.Combine(Environment.GetEnvironmentVariable("HOME"), "Applications/MGCB Editor.app");
+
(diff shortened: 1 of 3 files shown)


jnoyola:
@harry-cpp Shouldn't this be done in the csproj, not the cake build script? Now packing with cake and packing directly with dotnet will produce different output here.

```js
    var versionReg = @"<key>CFBundleShortVersionString<\/key>\s*<string>([^\s]*)<\/string>";
    var plistPath = "Tools/MonoGame.Content.Builder.Editor/Info.plist";
    var newVersion = "<key>CFBundleShortVersionString</key>\n\t<string>" + version + "</string>";
    ReplaceRegexInFiles(plistPath, versionReg, newVersion, System.Text.RegularExpressions.RegexOptions.Singleline);
```

harry-cpp:
We have CAKE exactly to handle stuff like packaging, if you want to use stuff from source you reference it, if you want it packaged, there is CAKE.

That being said, feel free to move it to the csproj if you want to, there is no real downside to it.

jnoyola:
Fair enough. If this were DesktopGL or WindowsDX then I might care more to be able to `dotnet pack` for testing without getting cake involved, but I don't personally feel much motivation to fix this for Mac.

From a theoretical standpoint, I like the idea that build/package details are pushed down as far as possible, meaning that since it's possible to `dotnet pack` this package, and it's possible for `dotnet pack` to do this task, I think it should. And then cake is used more of as an orchestrator for the various build and pack steps.

PR #7190, opened 2020-06-09
Remember pipeline tool position, size and panel dimensions

### LINKED ISSUE #6430: Feature Request : Pipeline Tool remember Position and Size
The tool should remember it's previous position, size, and dimensions of all the individual parts.

Just a small quality of life improvement.

Additionally it would be nice if it remembered which folders were expanded too.

AndiAn94:
The Pipeline Tool remembers the size of the panels and the size, position of the window and wether it was maximized.
The information get stored in the `Settings.xml`. If this file doesn't exist the hard-coded values from the earlier version will be used. The informations will be stored when the user closes the window.

I've used some of the `PipelineSettings` members which weren't used. Of course I will change that if the members were indeed used somewhere.

Fixes #6430

### DIFF
--- a/Tools/MonoGame.Content.Builder.Editor/Common/PipelineSettings.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/Common/PipelineSettings.cs
@@ -19,8 +19,8 @@ namespace MonoGame.Tools.Pipeline
 
         public List<string> ProjectHistory;
         public string StartupProject;
-        public System.Drawing.Size WindowSize;
-        public System.Drawing.Point WindowPosition;
+        public int WindowWidth, WindowHeight;
+        public int WindowPositionX, WindowPositionY;
         public bool WindowMaximized;
         public int HSeparator, VSeparator;
         public bool Maximized, DebugMode, PropertyGroupSort;
@@ -43,8 +43,10 @@ namespace MonoGame.Tools.Pipeline
             FilterShowCleaned = true;
             AutoScrollBuildOutput = true;
 
-            WindowSize = new System.Drawing.Size(900, 550);
-            WindowPosition = new System.Drawing.Point(182, 182);
+            WindowWidth = 900;
+            WindowHeight = 550;
+            WindowPositionX = 182;
+            WindowPositionY = 182;
             WindowMaximized = false;
             HSeparator = 200;
             VSeparator = 230;
--- a/Tools/MonoGame.Content.Builder.Editor/Common/PipelineSettings.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/Common/PipelineSettings.cs
@@ -20,7 +20,7 @@ namespace MonoGame.Tools.Pipeline
         public List<string> ProjectHistory;
         public string StartupProject;
         public int WindowWidth, WindowHeight;
-        public int WindowPositionX, WindowPositionY;
+        public int WindowPositionX, WindowPositionY, WindowMaximizedPositionX, WindowMaximizedPositionY;
         public bool WindowMaximized;
         public int HSeparator, VSeparator;
         public bool Maximized, DebugMode, PropertyGroupSort;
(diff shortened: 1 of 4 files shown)


harry-cpp:
What did you test this on, only Windows?

Tools/MonoGame.Content.Builder.Editor/Common/PipelineSettings.cs
```diff
@@ -19,7 +19,9 @@ public class PipelineSettings
 
         public List<string> ProjectHistory;
         public string StartupProject;
-        public Microsoft.Xna.Framework.Point Size;
+        public System.Drawing.Size WindowSize;
```
tomspilman, line 22: @harry-cpp - are we using `System.Drawing` in other places... is it portable for us?

tomspilman:
One thing that should be tested to be sure works... is that if the state of the window is stored for a secondardy monitor.  Then that second monitor is disconnected and the content editor opened...  does it fix its positioning to be on the primary monitor or does it move outside the desktop where it is hard to move back?

Tools/MonoGame.Content.Builder.Editor/Common/PipelineSettings.cs
harry-cpp: Its not, it should not be used.

AndiAn94: Ok, gonna change it.

AndiAn94:
> What did you test this on, only Windows?

Yes only on Windows.

> One thing that should be tested to be sure works... is that if the state of the window is stored for a secondardy monitor. Then that second monitor is disconnected and the content editor opened... does it fix its positioning to be on the primary monitor or does it move outside the desktop where it is hard to move back?

Didn't tested it before I opened the PR, and indeed the window will be created off-screen. I will look for a solution for this. Worst-case would be either restrict the position wihtin the primary monitor or don't save the position at all.

harry-cpp:
> Yes only on Windows.

Linux and macOS might have their own quirks.

AndiAn94:
> Linux and macOS might have their own quirks.

On Linux there is a small bug I'm still working on. The displayed window is bigger than the set size (I assume because of the Gtk style) and is increasing each time the pipeline tool gets started.

Unfortunately I don't have any option to test the changes for macOS. Any idea's how we could test the changes?

> One thing that should be tested to be sure works... is that if the state of the window is stored for a secondardy monitor. Then that second monitor is disconnected and the content editor opened... does it fix its positioning to be on the primary monitor or does it move outside the desktop where it is hard to move back?

How should the window behave if it's off-screen? Moving it back to the default position or just moving it into the screen so that the complete window is visible again?

I also tested it on Linux and Gtk is handling this already. As stated above I can't test it on macOS.

harry-cpp:
Important note about Linux, you can remember the size of the window, and the maximized state, but do not remember its position, let the system handle that!

I'll test it on macOS one you are done with Windows and Linux.

AndiAn94:
> Important note about Linux, you can remember the size of the window, and the maximized state, but do not remember its position, let the system handle that!

Ubuntu 20.04 by itself does not remember the position and it seems some other distributions don't do that aswell. Should I still leave out saving the position for Linux?

harry-cpp:
Yes, it does not remember it, you should not try to remember it either.

harry-cpp:
Use `Global.IsLinux` instead of `#` stuff.

AndiAn94:
With the latest commits this Feature should be ready for Linux and Windows.

- The Pipeline Tool starts with the same size as it was closed before and also if maximized or not.
- The 3 individual parts of the window will remain the same.

Windows:
- If window is off-screen, when starting it will be set to the default position and size.

Linux (tested on Ubuntu 20.04):
- As @harry-cpp stated above, I removed the saving of the position and the Pipeline Tool will be started on the systems default position. If by any chance the window will be off-screen, Gtk handles it.

If any changes are not ok or you find some bugs, of course I will change/fix it.

PR #7198, opened 2020-06-12
Make ModelMeshPart constructor public but obsolete

limbusdev:
Turning `ModelMeshPart` constructor private/internal prevents people from writing custom model loaders. Until there is a better way to work around it, this change should be rolled back. Instead of making the constructor internal I marked it as obsolete.


Jjagg:
Hey, thanks @limbusdev! 
Can you add a sentence about why it’s dangerous to use this constructor? I explained in the PR that made this internal.

limbusdev:
Sure. As comment or in the obsolete attribute?

limbusdev:
Better?

Jjagg:
Great, thank!

PR #7199, opened 2020-06-13
Fix content failing to load due to mscorlib and System.Private.CoreLib

harry-cpp:
Depending on how we compile the content, one of these 2 core libs will be written down as the type, however mscorlib is the .NET Framework name for the library, while System.Private.CoreLib is the .NET Core name for the library so we need to check the current runtime to know which of them to use.


YTN0:
@Jjagg  Can we get this PR merged? Spritefonts are currently broken in the .net framework nugets without this.

Thanks.

Issue #5988, opened 2017-10-04
Tap on trackpad not works correctly [Mac OS X]


mgkcortyw:
Hello, guys!

I trying to migrate my iOS game to MacOS, and my code: 
'mouse.LeftButton == ButtonState.Pressed' only is executed when I click the trackpad and not when I tap it.

My MacOS is configured to "Tap to click - Tap with one finger".

Sometimes on trackpad-Tap state of left mouse button changed on 'Pressed' state, but it's very unstable (1 of 10 taps) in other cases in update method we get always 'Released' mouse state like we don't Tap ever.

Looks like MonoGame not catch very fast clicks (Tap-s). (i.e. 'Pressed' state changed very fast from 'Pressed' to 'Released', and Update method don't catch it).

Tested on last **stable MonoGame Release 3.6** and with **develop branch**. Same problem.

Thank in advance!

cra0zy:
Hmm... this might be due to how XNA input API works, ie. it might be missing the state when the touchpad is pressed.

What platform MG assembly are you using, DesktopGL?

mgkcortyw:
@cra0zy thank you for answer.

Yes, DesktopGL.

> it might be missing the state when the touchpad is pressed.

No, input works fine.

I make some research and can say that ->
Update method calls slower than Mac trackpad change state (chain) ['Released' -> 'Pressed' -> 'Released']. i.e. between Update calls mouse state changed from 'Released' to 'Released' and we can't catch that use make tap.

I check this with snippet, in my LoadContent method i call it:

        private void CheckActualMouseState()
        {
            Task.Run(() =>
            {
                while (true)
                {
                    var mouseState = Mouse.GetState();
                    if(mouseState.LeftButton == ButtonState.Pressed)
                        Debug.WriteLine(mouseState.LeftButton);
                }
            });
        }

forever while catch my trackpad taps correctly, but it's ugly solution, because new thread, and while block.

cra0zy:
> I make some research and can say that ->
> Update method calls slower than Mac trackpad change state (chain) ['Released' -> 'Pressed' -> 'Released']. i.e. between Update calls mouse state changed from 'Released' to 'Released' and we can't catch that use make tap.

That is exactly what I said... its missing the state when its pressed...

@tomspilman @KonajuGames @dellis1972 So... anybody for an event driven input API?, in all seriousness this does sound like a big problem and I'm just tagging you along for an idea on how to fix it.

[...]

PR #7206, opened 2020-06-15
[DesktopGL] Add left and right mouse button event handlers

HaikuJock:
[DesktopGL] Add left and right mouse button event handlers and trigger them from the event polling.

This allows client applications to handle sub-frame mouse events.
Addresses issue MonoGame#5988 [Mac OS X] Tap on trackpad not works correctly
This is only for the DesktopGL version as the issue has only been reported on Mac OS.
Three-finger tap for middle mouse button is not reported by SDL as a button event so no event handler has been added for the middle mouse button.

### DIFF
--- a/MonoGame.Framework/GameWindow.cs
+++ b/MonoGame.Framework/GameWindow.cs
@@ -117,6 +117,28 @@ namespace Microsoft.Xna.Framework {
 
 #endif
 
+#if DESKTOPGL
+        /// <summary>
+        /// Use this event to receive all left mouse button presses and releases.
+        /// This includes sub-frame events. Handlers are called at the beginning of a frame, before game updates.
+        /// <remarks>
+        /// This event is only supported on the DesktopGL platforms.
+        /// </remarks>
+        /// </summary>
+        public event EventHandler<ButtonState> LeftButtonInput;
+
+        /// <summary>
+        /// Use this event to receive all right mouse button presses and releases.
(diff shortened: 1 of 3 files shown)


HaikuJock:
I made this DesktopGL only because that's the only platform that there is a need for it. I'm reluctant to write something that will never be used. The usual way of obtaining mouse state with Mouse.GetState() is sufficient for most users and simpler. Adding this API to other platforms will complicate them unnecessarily. I understand this is a trade-off against API consistency.

I'm happy to change it to MouseButtonPressed and MouseButtonReleased.

I can't find any SDL MouseClickedEvent.

I felt like we discussed this in https://github.com/MonoGame/MonoGame/issues/5988. Shall we continue this conversation there?

harry-cpp:
> I felt like we discussed this in #5988. Shall we continue this conversation there?

Yes, propose the API, I'll add some comments, also ping Tom, any API change needs to go by him.

ryancheung:
Have you tried the event based inputs provided by  MonoGame.Extemded repo, if you want a mouse button event. But not sure if the tap on trackpad event has bug there.

HaikuJock:
MonoGame.Extended builds on top of MonoGame. Looking at MouseListener in that repo it works off Mouse.GetState() so I expect it will have the same bug.

HaikuJock:
Unacceptable!

PR #7211, opened 2020-06-18
[Android] Fixed surface recreation when resuming from sleep mode on FireTV

### LINKED ISSUE #7209: EglMakeCurrent fails when returning from sleep mode on FireTV devices
## Issue
So on some Amazon FireTV devices there is an issue resuming from sleep mode. The game resumes however it is only a black screen with music playing. This is currently blocking an update submission of mine. I say _some_ devices because I actually couldn't reproduce the problem at first, I ended up having to do a factory reset on my FireTV Stick 4k to get the problem to show up. So obviously there is something about this that's partially an issue with the device/os itself, however I have found that it is possible to solve the problem on our end.

## Cause
For reference, the issues I'll describe are in this [MonoGameAndroidGameView.cs](https://github.com/MonoGame/MonoGame/blob/3fde2bf96888ca7e21dc090389e59260a999401a/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs)

What I've found is that when the FireTV comes out of sleep mode it triggers certain events before monogame is ready. This method must be called __successfully__.
```C#
public void SurfaceChanged(ISurfaceHolder holder, global::Android.Graphics.Format format, int width, int height)
{
    // Set flag to recreate gl surface or rendering can be bad on orienation change or if app 
    // is closed in one orientation and re-opened in another.
    lock (_lockObject)
    {
        // can only be triggered when main loop is running, is unsafe to overwrite other states
        if (_internalState == InternalState.Running_GameThread)
        {
            _internalState = InternalState.ForceRecreateSurface;
        }
    }
}
```
[SurfaceChanged()](https://github.com/MonoGame/MonoGame/blob/3fde2bf96888ca7e21dc090389e59260a999401a/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs#L104-L117)

The method is indeed getting called, however the _internalState is still InternalState.Resuming_UIThread . Because of this nothing actually happens and the InternalState is never set to ForceRecreateSurface, which needs to happen.

Because the surface never gets recreated the method [MakeCurrent()](https://github.com/MonoGame/MonoGame/blob/3fde2bf96888ca7e21dc090389e59260a999401a/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs#L156-L165) constantly fails and spits out the error 
> eglMakeCurrent:819 error 300b (EGL_BAD_NATIVE_WINDOW)

## Solutions
The way I see it there are 2 possible solutions. 

One would be to implement a system that ensures _internalState will eventually get set to ForceRecreateSurface even if _internalState is not Running_GameThread when SurfaceChanged gets called. I assume this would involve adding some variable such as _needToForceRecreateSurface that __always__ gets set to true in SurfaceChanged and then when _internalState returns to Running_GameThread the state could be set to ForceRecreateSurface by [RunIteration()](https://github.com/MonoGame/MonoGame/blob/3fde2bf96888ca7e21dc090389e59260a999401a/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs#L568-L633)

The second solution would be to simply insert the same code from SurfaceChanged into the MakeCurrent method if the failure occurs. Here's an example of that implementation.
```C#
public virtual void MakeCurrent()
{
    EnsureUndisposed();
    if (!egl.EglMakeCurrent(eglDisplay, eglSurface,
            eglSurface, eglContext))
    {
        if (egl.EglGetError() == EGL10.EglBadNativeWindow)
        {
            lock (_lockObject)
            {
                // can only be triggered when main loop is running, is unsafe to overwrite other states
                if (_internalState == InternalState.Running_GameThread)
                {
                    _internalState = InternalState.ForceRecreateSurface;
                }
            }
        }
        System.Diagnostics.Debug.WriteLine("Error Make Current" + GetErrorAsString());
    }
}
```

I would be glad to implement either of these solutions (or a proposed 3rd option) but wanted to get feedback on the preferred fix before submitting a pull request.

- MonoGame 3.7.1
- Current develop branch build

- Windows

- Android

Romans-I-XVI:
Fixes  #7209 

There are more details about what this fixes in the issue I opened. Essentially, the timing at which certain calls occur make it so [SurfaceChanged()](https://github.com/MonoGame/MonoGame/blob/3fde2bf96888ca7e21dc090389e59260a999401a/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs#L104-L117) gets called while _internalState is still set to Resuming_UIThread. The result is a black screen when resuming from sleep mode on FireTV.

### DIFF
--- a/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs
+++ b/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs
@@ -52,6 +52,7 @@ namespace Microsoft.Xna.Framework
         volatile InternalState _internalState = InternalState.Exited_GameThread;
 
         bool androidSurfaceAvailable = false;
+        bool needToForceRecreateSurface = false;
 
         bool glSurfaceAvailable;
         bool glContextAvailable;
@@ -111,6 +112,10 @@ namespace Microsoft.Xna.Framework
                 if (_internalState == InternalState.Running_GameThread)
                 {
                     _internalState = InternalState.ForceRecreateSurface;
+                }
+                else
+                {
+                    needToForceRecreateSurface = true;
                 }
 
             }
@@ -574,6 +579,11 @@ namespace Microsoft.Xna.Framework
 
             lock (_lockObject)
             {
+                if (needToForceRecreateSurface && _internalState == InternalState.Running_GameThread)
+                {
+                    _internalState = InternalState.ForceRecreateSurface;
+                    needToForceRecreateSurface = false;
+                }
                 currentState = _internalState;
             }
 


Jjagg:
Completely missed this PR, but seems like a good fix! Thanks @Romans-I-XVI!

Issue #7195, opened 2020-06-10
Class to wrap an ID3D11Texture2D as a RenderTarget2D


YTN0:
Per the contributor guide, any new code or proposals should first be discussed here.

I would like to submit a class that wraps an
ID3D11Texture2D as a RenderTarget2D. I used the SwapChainRenderTarget as a baseline for creating this, and currently have something that works.

This will become useful for anyone interested in OpenXR, since the OpenXR runtime returns an already created ID3D11Texture2D. This differs from the way OpenVR works, since with OpenVR you create the RenderTarget and pass it the handle, as opposed to the other way around for OpenXR.

I currently have the class named D3DRenderTarget, but am open to renaming it. It resides in the same directory and namespace as SwapChainRenderTarget.

Note, this would be for Windows / DirectX only.

I would love to get this into 3.8 so I don't have to run a modified version of the Monogame dll.

Note, this is new functionality and should not impact any existing behavior..

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- WindowsDX

YTN0:
@Jjagg Can you clarify what you mean?

Are you saying modify the current RenderTarget2D and overload the constructor to take an ID3D11Texture2D handle, or something else?

YTN0:
@Jjagg If you can point me in the right direction, I can try and put something together for this.

Jjagg:
Sorry for the delay. I think the best way to expose this is with a factory method. A static method on RenderTarget2D (and maybe Texture2D) ‘FromDirect3DHandle’ that creates an instance from the existing handle, validating the handle in the method as well.

tomspilman:
> A static method on RenderTarget2D (and maybe Texture2D) ‘FromDirect3DHandle’ 

Why not a `FromHandle(IntPtr handle)`  or `FromNativeHandle` or some other name that can be supported on multiple backends?

I could see the same feature being useful on OpenGL or other graphics APIs.

YTN0:
@Jjagg @tomspilman Thanks for the feedback. I'll try to put something together that's more generic so it can be used across platforms. Will ping you guys if I have questions.

uygary:
@tomspilman would this be something we can tackle for 3.8.6 maybe?
I don't wanna shove anything in that could potentially delay 3.8.5 release, so if 3.8.6 sounds reasonable, I'd like to take a stab at it.
[...]

PR #7214, opened 2020-06-19
Fix for #7195 - Add FromNativeHandle() to RenderTarget2D

YTN0:
@Jjagg @tomspilman Let me know if this works.

Adding support for creating RenderTarget2D from native ID3D11Texture2D handle. I modeled this from the original SwapChainRenderTarget code, and it was working fine for me. However, the original code I had broke with change #5838 (and I am guessing SwapChainRenderTarget is also broken).

@Jjagg  I believe there was an error that was introduced in #5838 

You created a new SampleDescription, however, the original SampleDescription in Texture2D.DirectX wasn't being set / updated when the SurfaceType was SwapChainRenderTarget, which was causing an error when calling out to SharpDX (I believe this error will only surface when the SurfaceType is SwapChainRenderTarget, so probably would not happen otherwise).

In order to fix this, I made the change in line 210 of RenderTarget2D.DirectX.

I could be wrong.... please double check this!

@Jjagg  Update: Looks like I might have broken something with that SampleDescription change (line 210). One of the MSAA tests is failing. Would appreciate if you could give this PR a look-over and let me know how to fix the SampleDescription problem without breaking anything. This code is fairly complex, so it's hard to determine what the correct fix should be.

### DIFF
--- a/MonoGame.Framework/Graphics/RenderTarget2D.cs
+++ b/MonoGame.Framework/Graphics/RenderTarget2D.cs
@@ -23,7 +23,20 @@ namespace Microsoft.Xna.Framework.Graphics
             return ContentLost != null;
         }
 
-	    public RenderTarget2D(GraphicsDevice graphicsDevice, int width, int height, bool mipMap, SurfaceFormat preferredFormat, DepthFormat preferredDepthFormat, int preferredMultiSampleCount, RenderTargetUsage usage, bool shared, int arraySize)
+        public static RenderTarget2D FromNativeHandle(GraphicsDevice graphicsDevice, IntPtr handle, int width, int height)
+        {
+            return FromNativeHandle(graphicsDevice, handle, width, height, false, SurfaceFormat.Color, DepthFormat.Depth24, 0, RenderTargetUsage.DiscardContents);
+        }
+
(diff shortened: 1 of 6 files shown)


YTN0:
@Jjagg  I believed I fixed the SampleDescription issue (which was failing one of the MSAA tests). The tests are all passing, and my code that is using the wrapped handle works without any issues.

Please take a look when you get a chance. 

FYI...The code for this stuff is pretty complex.... hopefully someday, someone who knows this stuff really well (definitely not me 😄) could refactor this to make it a little easier to understand!

nkast:
The commit https://github.com/Jjagg/MonoGame/commit/020f4e221a14eb46b0f995a58e373d1289b4f165
seems to merge two constructor into one. 

One of those was meant to be used by the SwapChainRenderTarget, overiding the default constructor without calling PlatformConstruct() and letting SwapChainRenderTarget create it's own _texture.

Before:
https://github.com/Jjagg/MonoGame/blob/31e9409aedef7384969deb0eb2ace26d28b3bfea/MonoGame.Framework/Graphics/RenderTarget2D.cs#L72-L86

After: 
https://github.com/Jjagg/MonoGame/blob/4d866e4a6a9d27423157dc433748245cead7f18d/MonoGame.Framework/Graphics/RenderTarget2D.cs#L49-L60

nkast:
@YTN0 , note that if #7217 gets accepted, I 'd like to see _texture in Texture2D become private. 
SwapChainRenderTarget is the only one that directly set it and needs some clean up, 
IMO, it's more clear to override Texture2D.CreateTexture().

YTN0:
@Jjagg @nkast 

Deleting this PR. Will submit a new one based on the changes @nkast made that were just merged into develop.

PR #7215, opened 2020-06-20
[MGCB Editor] Fix parsing filename from arguments on macOS

### LINKED ISSUE #7203: mgcb editor can't open mgcb files when its passed as an argument on macOS
On macOS we always had a very awkward implementation of sending the filename to the mgcb editor.

harry-cpp:
Fixes #7203


Jjagg:
Awesome :)

Issue #5098, opened 2016-08-10
Fixed incorrect glyph position in FontDescriptionProcessor.cs


zsmolnar:
Hi guys,

I believe I fixed a bug in the FontDescriptionProcessor.cs. 

line 120 original:
var cropping = new Rectangle(0, (int)(glyph.YOffset - yOffsetMin), (int)glyph.XAdvance, output.VerticalLineSpacing);

fixed:
var cropping = new Rectangle((int)(glyph.XAdvance + glyph.XOffset), (int)(glyph.YOffset - yOffsetMin), (int)glyph.XAdvance, output.VerticalLineSpacing);

In our case, using https://www.fontsquirrel.com/fonts/kingthings-petrock font, the 'c' character was positioned incorrectly. See the attached image.

![image](https://cloud.githubusercontent.com/assets/1948647/17573784/973315ca-5f54-11e6-91a6-f44b5989403a.png)

Sorry I'm not submitting a PR, I'm not a git user.

dellis1972:
@zsmolnar Thanks for the bug report.

stromkos:
This is caused by the GlyphCropper.Crop() method not adjusting the kerning values for the trimmed glyph.

This problem only happens for certain fonts. From the freetype2 code comments:

> The control box encloses all the outline’s points, including Bezier control points. Though it coincides with the exact bounding box for most glyphs, it can be slightly larger in some situations (like when rotating an outline which contains Bezier outside arcs).

>Computing the control box is very fast, while getting the bounding box can take much more time as it needs to walk over all segments and arcs in the outline. To get the latter, you can use the ‘ftbbox’ component which is dedicated to this single task.

nkast:
I've made a mistake in my first commit 87dfc7d. This issue is closed by #7501, not #7502.

PR #7218, opened 2020-06-21
Fix incorrect glyph position

delightedcat:
This pull request contains a fix to properly crop a font on the vertical axis. The issue is described in further detail in #5098.


tomspilman:
This seems like it would be a pretty big bug in font processing.  Thanks @DelightedCat and @zsmolnar.

mrhelmut:
I'd like to double check this if you don't mind. Cropping XAdvance and/or XOffset doesn't sound correct to me, it is likely to crop too much on some/most fonts.
If they are rendering inaccuracies, it is likely somewhere else.

mrhelmut:
It seems to me that #7501 may have the proper fix. @delightedcat any chance you could test #7501?

delightedcat:
I just checked that one out. It seems more reliable than the patch in this PR, although I currently don't have the right setup to test the fix. Maybe someone else is in a position to test it out?

mrhelmut:
#7501 has been merged. Thank you for working on the initial proposal!

PR #7227, opened 2020-07-05
Bump dependencies

### LINKED ISSUE #7226: Statically link freetype and freeimage on Windows
Title says it all :) Lets statically link freetype and freeimage on Windows.

mrhelmut:
Bumping dependencies to use the static freetype & freeimage builds on Windows.

Fixes #7226

@harry-cpp


harry-cpp:
And I would have gotten away with it too, if it weren't for you meddling build bot XD

Something is not right with freetype lib.

mrhelmut:
Hmm, it worked locally. Do you know which version of SharpFont we use? The latest upstream uses freetype 2.5.5 but we may be off (or using a custom version?).

PR #7228, opened 2020-07-06
[UWP] Fix to SynchronizeWithVerticalRetrace = false, causing a crash on startup

### LINKED ISSUE #6020: [UWP] Disabling V-Sync crashes MG
Setting graphics.SynchronizeWithVerticalRetrace to false causes MG to crash.

BMarques:
This is a simple fix to stop from the game to crash when starting, when SynchronizeWithVerticalRetrace  = false when building for UWP, but the VSync is still enabled anyway (will probably do another Pull Request some time later to fix this as well, after I'm confortable with my (somewhat big) changes). 
Fixes #6020


BMarques:
This was my first pull request, I'm pretty sure I'm doing something wrong ... I only wanted to pull the 466b07b

BMarques:
Do I need to say something spefici or request a review so that someone can have a look at this? As I said it's my first PR, but i'm a bit confused why there's no reaction after three weeks :).

Jjagg:
No, it's all on us that we missed this PR, sorry!

I'm on mobile but I'll review this later today.

Reading some more about the flipping model, it seems like it would be a good optimization to check for buffer flipping support on WindowsDX too.

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -187,6 +187,30 @@ protected virtual void CreateDeviceResources()
...
+#if WINDOWS_UAP
+            // Check if tearing is supported
+            if (PresentationParameters.PresentationInterval == PresentInterval.Immediate)
+            {
+                RawBool allowTearing;
+                using (var dxgiFactory2 = new Factory2())
+                {
+                    unsafe
+                    {
+                        var factory5 = dxgiFactory2.QueryInterface<Factory5>();
```
Jjagg, line 199: This line should go inside the `try ... catch` block in case the interface isn't supported.

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -187,6 +187,30 @@ protected virtual void CreateDeviceResources()
                 creationFlags |= SharpDX.Direct3D11.DeviceCreationFlags.Debug;
             }
 
+#if WINDOWS_UAP
+            // Check if tearing is supported
+            if (PresentationParameters.PresentationInterval == PresentInterval.Immediate)
```
Jjagg, line 192: This check should happen in `OnPresentationChanged` because CreateDeviceResources is only called when the GD is created (I think the rest of the code here is not 100% consistent either though).

BMarques:
Thanks, I'll have a look at your comments either today or tomorrow.

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
BMarques: The reason why it's not within a try catch is because that factory interface was introduced on the Anniversary SDK (14393). MonoGame doesn't support a lower version, so I just decided not to do a try catch because it shouldn't ever throw an exception. Do you want me to do it nevertheless?

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
BMarques: I moved it as you requested, it's still working on my sample as before.

BMarques:
> No, it's all on us that we missed this PR, sorry!
> 
> I'm on mobile but I'll review this later today.
> 
> Reading some more about the flipping model, it seems like it would be a good optimization to check for buffer flipping support on WindowsDX too.

From what I saw, on WindowsDX, there are some things that are not clear. MonoGame supports Windows 7, right? The new flip models only work on Windows 8 (FlipSequential) and Windows 10 (FlipDiscard). If MonoGame had a way to know which version of Windows is running the code, then it should be possible to create a better ToSwapEffect helper method. For UWP this was easy, but not so much for Desktop.

Jjagg:
Yes, we can figure out the Windows version and verify the OS supports flipping. But that was just a thought for future improvement and not a suggestion for this pull request :)

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -317,11 +341,19 @@ internal void CreateSizeDependentResources()
             // If the swap chain already exists... update it.
             if (_swapChain != null)
             {
+                var swapChainFlags = SwapChainFlags.None;
+#if WINDOWS_UAP
+                if (PresentationParameters.PresentationInterval == PresentInterval.Immediate)
+                {
+                    swapChainFlags = SwapChainFlags.AllowTearing;
+                }    
+#endif
```
Jjagg, line 350: If you move the code here out of the if statement you can use the  swapChainFlags variable below in the initializer when creating a new swapchain and remove the duplicate code.

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -251,6 +252,29 @@ protected virtual void CreateDeviceResources()
 
         internal void CreateSizeDependentResources()
         {
+#if WINDOWS_UAP
```
Jjagg, line 255: Could you extract this into a function? That'd keep this function a little easier to read.

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -187,6 +187,7 @@ protected virtual void CreateDeviceResources()
                 creationFlags |= SharpDX.Direct3D11.DeviceCreationFlags.Debug;
             }
 
+
```
Jjagg, line 190: Looks like you accidentally added a line here.

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
BMarques: Nice catch, I didn't notice. Will improve this.

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
BMarques: Yep, cut paste code to somewhere else left a trail behind :)

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
BMarques: Done.

MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs
```diff
@@ -378,7 +358,7 @@ internal void CreateSizeDependentResources()
                 };
```
Jjagg, line 358: You can move setting Flags in the SwapChainDescription1 initializer here.
Also don't need the WINDOWS_UAP check since it's set to none for WindowsDX which is fine.

```suggestion
                    Flags = swapChainFlags,
                };
```

Jjagg:
Awesome! This is great @BMarques, thanks again 😁

BMarques:
Glad to help, it was a nice first PR experience, learnt a lot about git :).

PR #7239, opened 2020-07-19
fix path in RebuildMGFX.bat

cpt-max:
I also added a pause at the end so you can read error messages.


harry-cpp:
Thanks @cpt-max !

PR #7242, opened 2020-07-22
Hull, Domain and Geometry Shaders for WindowsDX

cpt-max:
This pull request is based on [JSandusky's MonoGame branch](https://github.com/JSandusky/MonoGame).
It adds the missing DX11 shader stages to the framework and to the content pipeline (mgfxc). 
I created this [MonoGame sample project](https://github.com/cpt-max/ShaderTest) to test the new stages.

The MGFXVersion has been raised to 10 which means shader content need to be rebuilt. 
The changes shouldn't break other platforms (theoretically), but they add some non-functional elements to the framework, like for example new PrimitiveTypes that only work with hull shaders. I could #if them out of those platforms for now, but I'd prefer to make them functional instead :)

The DX unit tests didn't work for me, every test that needs a DX device fails to create one. What am I missing? OpenGL unit tests work fine.


tomspilman:
So for sure this needs to sit at least until 3.8 is out the door... we'll get this into 3.9.

MonoGame.Framework/Graphics/Shader/ShaderStage.cs
```diff
@@ -4,5 +4,8 @@ internal enum ShaderStage
     {
         Vertex,
         Pixel,
+        Hull,
+        Domain,
+        Geometry,
```
tomspilman, line 9: We should document exactly what these are even with it being internal.  Will help when we port to new platforms and also when we move to make the Shader API public.

tomspilman:
@cpt-max - You should add a new effect processor test for these new shader stages.  These were moved here:

https://github.com/MonoGame/MonoGame/blob/develop/Tools/MonoGame.Tools.Tests/EffectProcessorTests.cs

You will need to put a `#if DIRECTX` around it keep it from running the tests on the Mac.

georg-eckert-zeiss:
@cpt-max - I had a look into your PR. Looks very promising, great work! 

I noticed that `PrimitiveType` does not contain `LineListAdj`, `LineStripAdj`, `TriangleListAdj` and `TriangleStripAdj` - on purpose?

Cheers,
Georg

cpt-max:
> I noticed that `PrimitiveType` does not contain `LineListAdj`, `LineStripAdj`, `TriangleListAdj` and `TriangleStripAdj`

I'll add those when I update this PR.

cpt-max:
@tomspilman I added the documentation and unit tests you requested. I don't think the DIRECTX symbol is defined though, so those tests don't run.
@georg-eckert-zeiss the adjacency primitive types are now available.

tomspilman:
Thanks @cpt-max .  I'll find some time to review this in more detail this week.

georg-eckert-zeiss:
> @georg-eckert-zeiss the adjacency primitive types are now available.

Great : ) Thank you. Really looking forward to this.

cpt-max:
I didn't realize at first that some files in MGFXC are supposed to be autogenerated by TinyPG, rather than changed manually. With the latest commit they are autogenerated again.

cpt-max:

PR #7246, opened 2020-07-25
Fix for #7195 - Add FromNativeHandle() to RenderTarget2D

YTN0:
@Jjagg This is an updated fix for #7195 to add support for wrapping a native ID3D11Texture2D handle. This is useful for anyone who wants to support OpenXR, since OpenXR handles the creation of the ID3D11Texture2D textures in the OpenXR runtime.

I updated this based on the new code that @nkast submitted that got merged into develop. 

Note: I tested these changes with my OpenXR code and it works without any issues.

CC: @tomspilman


YTN0:
Any chance someone could take a look at this PR?

Thanks.

PR #7247, opened 2020-07-25
Fix location of Info.plist in packaging documentation

nathanhoad:
The `Info.plist` should be *inside* the top level `Contents` folder otherwise you get a message about the app being "damaged or incomplete".


Jjagg:
Thanks!

PR #7251, opened 2020-07-25
Upgrade VSMac addin

harry-cpp:
- MGCB Editor is now properly set as a default editor
- Fixed some minor issues in MGCB Editor property grid
- VSMac addin project is connected to the mac tools solution
- Added MGCB Editor Pad


harry-cpp:
I'll get some important bits in a separate PR.

PR #7253, opened 2020-07-25
VSMac addin fixes

harry-cpp:
- Fixed MGCB Editor not set as default
- Hide MGCB Editor option if MGCB Editor is not in the system
- Add addin to the Mac Tools solution


SimonDarksideJ:
Tested and working

SimonDarksideJ:
Also confirmed that after running mgcb-editor --register on a mac with this PR now enables opening the editor from VS for mac.

Will update the docs

PR #7257, opened 2020-07-27
Renamed all instances of MonoGame.Utilities to MonoGame.Framework.Utilities

### LINKED ISSUE #7254: Two conflicting Utility API namespaces within MonoGame namespace
Whilst reviewing the API documentation, there seems to be two conflicting/overlapping namespaces in the MonoGame codebase, which could lead to confusion:

MonoGame.Framework.Utilities
MonoGame.Utilities

IMHO, there should only be a single MonoGame Utilities namespace to avoid confusion
(There is also a MonoGame.Utilities.Deflate namespace that will need aligning)

Given the breaking nature of MonoGame 3.8, it would be advisable to align these for the 3.8 release.

- MonoGame 3.8 Development

SimonDarksideJ:
As identified in the docs review, there were multiple instances of a Utilities namespace within the MonoGame framework.

To that end, all MonoGame utilities instances have been renamed to:

MonoGame.Framework.Utilities

Resolves #7254

(Clean Resubmit to avoid formatting changes/updates on the entire codebase)

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
+++ b/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
@@ -6,7 +6,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Threading;
-using MonoGame.Utilities;
+using MonoGame.Framework.Utilities;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline
 {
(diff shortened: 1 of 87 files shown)


tomspilman:
@Jjagg @harry-cpp this looks good to me.

Jjagg:
Thanks Simon!

PR #7260, opened 2020-07-28
Fix loading of content with special characters in a file name

### LINKED ISSUE #7167: Loading content fails when '#' is in the file name
Example: `var test = Content.Load<Texture2D>("tes#t");`

'#' character is lost in this method, by the Uri class:
https://github.com/MonoGame/MonoGame/blob/3e92fc7bff7b4c32edf1314e143a4b7de4fab161/MonoGame.Framework/TitleContainer.cs#L66 

MonoGame 3.8 (pre-release, build from source)
Platform: WindowsDX
OS: Windows 10

EDIT: `%20` character sequence also causes problems with Uri.

Jure-BB:
Fix content loading when there are special characters like '#' or %20 in the file name/path.

Fixes #7167

### DIFF
--- a/MonoGame.Framework/TitleContainer.cs
+++ b/MonoGame.Framework/TitleContainer.cs
@@ -4,7 +4,7 @@
 
 using System;
 using System.IO;
-using MonoGame.Framework.Utilities;
+using  MonoGame.Framework.Utilities;
 
 namespace Microsoft.Xna.Framework
 {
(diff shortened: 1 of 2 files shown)


Jure-BB:
@harry-cpp Any idea how to fix this? I've been trying to correct line endings for over an hour with no success.

harry-cpp:
Undo commit and make a new one with ignored line changes (`git reset HEAD~` will undo your last commit and keep the changes you made available to you).

Jure-BB:
New PR #7267

PR #7265, opened 2020-07-29
shortcut for getting 4 directions(right,left,up,down)

Raader:
add 4 static public properties to vector2.cs file.
its a shortcut for getting 4 directions easily.

### DIFF
--- a/MonoGame.Framework/Vector2.cs
+++ b/MonoGame.Framework/Vector2.cs
@@ -25,7 +25,11 @@ namespace Microsoft.Xna.Framework
         private static readonly Vector2 unitVector = new Vector2(1f, 1f);
         private static readonly Vector2 unitXVector = new Vector2(1f, 0f);
         private static readonly Vector2 unitYVector = new Vector2(0f, 1f);
-
+        
+        private static readonly Vector2 rightVector = new Vector2(1f,0f);
+        private static readonly Vector2 leftVector = new Vector2(-1f,0f);
+        private static readonly Vector2 upVector = new Vector2(0f,-1f);
+        private static readonly Vector2 downVector = new Vector2(0f,1f);
         #endregion
 
         #region Public Fields
@@ -61,7 +65,7 @@ namespace Microsoft.Xna.Framework
         {
             get { return unitVector; }
         }
-
+       
         /// <summary>
         /// Returns a <see cref="Vector2"/> with components 1, 0.
         /// </summary>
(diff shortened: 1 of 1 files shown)


Raader:
line encoding got broken because of github but i dont understand why this code doesnt follow microsofs coding guidelines

MonoGame.Framework/Vector2.cs
```diff
@@ -1,670 +1,702 @@
...
+
+        private static readonly Vector2 zeroVector = new Vector2(0f, 0f);
+        private static readonly Vector2 unitVector = new Vector2(1f, 1f);
+        private static readonly Vector2 unitXVector = new Vector2(1f, 0f);
+        private static readonly Vector2 unitYVector = new Vector2(0f, 1f);
+        
+        private static readonly Vector2 rightVector = new Vector2(1f,0f);
+        private static readonly Vector2 leftVector = new Vector2(-1f,0f);
+        private static readonly Vector2 upVector = new Vector2(0f,-1f);
+        private static readonly Vector2 downVector = new Vector2(0f,1f);
```
nkast, line 32: UP and DOWN have wrong value. They need to be the other way around...

harry-cpp:
This:

```
public static Vector2 down{
    get {return downVector; }
}
```

Should be:

```
public static Vector2 Down
{
    get { return downVector; }
}
```

or:

```
public static Vector2 Down => downVector;
```

You have random missing spaces, in-line brackets and a public property starting with a small letter.

Raader:
Oh all of those are because i made it on mobile using github edit  i will fix it ande make a new pr when i have access to a computer

Jjagg:
> public static Vector2 Down => downVector;

@harry-cpp Can't use this syntax due to BRUTE.

Raader:
created the fixed pr

harry-cpp:
@Jjagg We can actually, we just can't use types that are C# 6+

Jjagg:
We have the language version set to 5 to prevent issues.

https://github.com/MonoGame/MonoGame/blob/687756238f4a660448526c3bf16e3db8bda8e7e2/MonoGame.Framework/Directory.Build.props#L14

willmotil:
He should be using the existing the existing UnitX and Y instead of creating new variables.
```
        public static Vector2 Up
        {
            get
            {
                return Vector2.UnitY;
            }
        }
        public static Vector2 Down
        {
            get
            {
                return -Vector2.UnitY;
            }
        }
```

Jjagg:
> Anyways.
A structs performance from everything i've seen is directly related to its size.

> This should be rejected on the basis that its bloat to a struct.
These are better as users personal extension methods.

> They would be nice but i fear for dropping fps for a unrequired edge case, especially when this is perfect to place in a extension class.

These are static members, they don't change instances in any way.

PR #7267, opened 2020-07-29
Fix loading of content with special characters in a file name v2

### LINKED ISSUE #7167: Loading content fails when '#' is in the file name

Jure-BB:
Fix content loading when there are special characters like '#' or %20 in the file name/path.

Fixes #7167

@harry-cpp I've created new PR as previous one was causing too much trouble. Mixed line endings and VS Code don't play well together.

### DIFF
--- a/MonoGame.Framework/TitleContainer.cs
+++ b/MonoGame.Framework/TitleContainer.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Xna.Framework
 
         internal static string NormalizeRelativePath(string name)
         {
-            var uri = new Uri("file:///" + name);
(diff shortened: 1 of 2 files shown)


harry-cpp:
This looks great! Thanks @Jure-BB !

PR #7269, opened 2020-07-30
shortcut for getting 4 directions(right,left,up,down) Fixed

Raader:
updated vector2.cs file.
Added 4 public static properties
(Up,Down,Right,Left).
this a fixed version of the old pull request

### DIFF
--- a/MonoGame.Framework/Vector2.cs
+++ b/MonoGame.Framework/Vector2.cs
@@ -25,7 +25,10 @@ namespace Microsoft.Xna.Framework
         private static readonly Vector2 unitVector = new Vector2(1f, 1f);
         private static readonly Vector2 unitXVector = new Vector2(1f, 0f);
         private static readonly Vector2 unitYVector = new Vector2(0f, 1f);
-
+        private static readonly Vector2 rightVector = new Vector2(1f, 0f);
+        private static readonly Vector2 leftVector = new Vector2(-1f, 0f);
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Vector2.cs
```diff
@@ -25,7 +25,10 @@ public struct Vector2 : IEquatable<Vector2>
         private static readonly Vector2 unitVector = new Vector2(1f, 1f);
         private static readonly Vector2 unitXVector = new Vector2(1f, 0f);
         private static readonly Vector2 unitYVector = new Vector2(0f, 1f);
-
+        private static readonly Vector2 rightVector = new Vector2(1f, 0f);
+        private static readonly Vector2 leftVector = new Vector2(-1f, 0f);
+        private static readonly Vector2 upVector = new Vector2(-1f,0f);
+        private static readonly Vector2 downVector = new Vector2(1f, 0f);
```
Kwyrky, line 31: right down and up left are the same vectors?

SimonDarksideJ: Any update @Raader on resolving the identified issue?

SimonDarksideJ (changes requested):

Raader:
i stopped working on this because the maintainer didn't think it was necessary and it would just be performance loss and it can be made easily with a helper class if the user wants to.

SimonDarksideJ:
OK, can you then close this PR and withdraw it.

PR #7271, opened 2020-07-30
Updated SDK versions in universal project and universal templates

### LINKED ISSUE #7220: Update MonoGame UWP SDK target
On Visual Studio 2019, Windows SDK 18362 is the default that is installed by the Visual Studio Installer. MonoGame is still targeting Windows SDK 17763 though... Is there any specific reason for that?

image

- MonoGame 3.8.0.1365

- Windows 10

- WindowsUWP

BMarques:
Update both the csproj for the Universal project and the templates. Fixes #7220.

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.WindowsUniversal.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.WindowsUniversal.csproj
@@ -1,11 +1,11 @@
-<Project Sdk="MSBuild.Sdk.Extras">
+﻿<Project Sdk="MSBuild.Sdk.Extras">
 
   <PropertyGroup>
     <TargetFramework>uap10.0</TargetFramework>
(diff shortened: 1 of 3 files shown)


BMarques:
Thanks

mrhelmut:
Could you please also set ```TargetPlatformMinVersion``` to ```10.0.16299.0``` (Windows 10 build 1709)? It is the most compatible and safest while still covering 98.3% of the Win 10 install base.

BMarques:
@mrhelmut Yes, I'll do so later today.

BMarques:
I was wondering if it would be a good idea to update the **Microsoft.NETCore.UniversalWindowsPlatform** nuget package from 6.2.9 to 6.2.10?

In advance, I tried but it's blocked by project
image

It's not very clear what I need to update to get the newest version, since it should work with both minimum and target versions I've set on the project...

Jjagg:
Thanks @BMarques!

Issue #7232, opened 2020-07-11
[MGCB] Missing symbols conversion for references


persn:
In my .mgcb file, if I write something like this then the symbols the converted as I expect, it creates a path like `bin/Debug/DesktopGL`
```
/outputDir:bin/$(Config)/$(Platform)
```

However I would expect the same thing to happen with assembly references as well, if I do this, the the symbol `$(Config)` does not get converted
```
/reference:..\Hero6.MonoGame.Pipeline\bin\$(Config)\netstandard2.0\Hero6.MonoGame.Pipeline.dll
```

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.0.1375-develop

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

persn:
I can see that this isse is not present in the official release of MonoGame 3.8

PR #7272, opened 2020-07-30
Replace symbols for assembly references, see #7232

Taronyu:
I added a call to ReplaceSymbols when the references are processed. Hope that was all to fix this.

### DIFF
--- a/Tools/MonoGame.Content.Builder/BuildContent.cs
+++ b/Tools/MonoGame.Content.Builder/BuildContent.cs
@@ -306,7 +306,7 @@ namespace MonoGame.Content.Builder
             // so it can resolve importers, processors, writers, and types.
             foreach (var r in References)
             {
-                var assembly = r;
+                var assembly = ReplaceSymbols(r);
                 if (!Path.IsPathRooted(assembly))
                     assembly = Path.GetFullPath(Path.Combine(projectDirectory, assembly));
                 _manager.AddAssembly(assembly);


Tools/MonoGame.Content.Builder/BuildContent.cs
```diff
@@ -306,7 +306,7 @@ public void Build(out int successCount, out int errorCount)
             // so it can resolve importers, processors, writers, and types.
             foreach (var r in References)
             {
-                var assembly = r;
+                var assembly = ReplaceSymbols(r);
```
tomspilman, line 309: Seems like a neat trick.

@harry-cpp ?

harry-cpp: Yea, we are gonna have to add support to mgcb-editor for it to be able to add symbols to reference paths.

Taronyu: From the GUI you mean? As far as I understood ticket #7232 it looked like a CLI/manual call but I must admit that I'm fairly new to the project.

tomspilman: @harry-cpp can you point @Taronyu at the code for that in the editor?

harry-cpp: It does not brake anything, just adds it to the CLI, I'll fix the editor when I get the chance.

PR #7276, opened 2020-08-02
Updated images to new docs standard for Windows

SimonDarksideJ:
Updated images to new docs standard for Windows.

Additional tasks:

* Update code docs to 3.8 standards
* removed any lingering "pipeline" references
* Removed redundant pipeline images


Documentation/articles/content/adding_ttf_fonts.md
```diff
@@ -7,13 +7,11 @@ MonoGame supports more than one method of using fonts, the following is an expla
...
-1. Create the .spritefont file.
+1. Create the .spritefont file by selecting "Edit -> Add -> New Item" from the MGCB Editor menu, then select **SpriteFont Description** from the list and click **Create**.
 
-  ![Adding TTF fonts step 1](~/images/adding_ttf_fonts_step_1.PNG)
+  ![Adding TTF fonts step 2](~/images/content/adding_ttf_fonts.PNG)
 
-  ![Adding TTF fonts step 2](~/images/adding_ttf_fonts_step_2.PNG)
-
-1. Open the newly created .spritefont file in your text editor of choice, find this line and change it to your selected .ttf font.
+2. Open the newly created .spritefont file in your text editor of choice, find this line and change it to your selected .ttf font.
```
Jjagg, line 14: Please leave this at 1. The markdown renderer automatically sets the right numbers in numbered lists, so conventions often use 1 for all entries in a list so things can be reordered/added/deleted without issue. We should probably have a style guide. I use the MS doc extension for vs code which is great to standardize formatting in md files and detect issues.

You can see the list is correct in the rendered file: https://docs.monogame.net/articles/content/adding_ttf_fonts.html

SimonDarksideJ: Apologies for the delay, sorted now @Jjagg

Documentation/articles/getting_started/3_adding_content.md
```diff
@@ -22,11 +22,11 @@ Now open up your game project and look at the Solution explorer window. Expand t
 
 ![Open Content](~/images/getting_started/3_open_content.png)
 
-You should now see the Pipeline Editor window open up. If it does not open up (you see a text file open), then you can right-click on **Content.mgcb** and select **Open With**, then select **MonoGame Pipeline** in the list, click **Set as Default** and then click **OK**.
+You should now see the MGCB Editor window open up. If it does not open up (you see a text file open), then you can right-click on **Content.mgcb** and select **Open With**, then select **mgcb-editor-wpf** in the list, click **Set as Default** and then click **OK**.
```
Jjagg, line 25: Hmm, this is not a very friendly name. I think we should be able to change this, but I don't know how VS picks this.

tomspilman: You mean `mgcb-editor-wpf` ?

@harry-cpp ?

Jjagg: Yes

SimonDarksideJ: Well if you can change it, then that PR should also update the name in the docs, shouldn't affect this PR.

Should be raised as a separate issue if need be (I agree, it's a crap name, seems to be based on the .NET project name maybe?)

Jjagg: It's the assembly name and the executable name.

jnoyola: My initial thought was that each platform's assembly could be in a subfolder, allowing them all to be named `mgcb-editor`, but then all supporting assemblies, templates, etc. would have to be duplicated in each platform's subfolder. I couldn't think of a clean way to do this so I just gave them unique names so they could live in the same main folder.

Jjagg: Yes, we shouldn't change that, just need a way to set the name as it appears in VS. If it's the executable name we can create a shortcut and register that in the Windows registry. Can we reuse the link from the start menu (#7223)?

PR #7277, opened 2020-08-03
3.8 Changelog Update

tomspilman:
This is the change log for 3.8.

Note that moving forward we plan to keep the change log updated as we go.  This will require people submitting PRs to update the changelog for their changes/features.


CHANGELOG.md
```diff
@@ -1,4 +1,171 @@
-# Change Log
+﻿# Change Log
```
tomspilman, line 1: Should i fix this while i am at it?

image

Change Log or Changelog ?

Should the title here be `MonoGame Change Log` ?

Anything else we want to add to this?

CHANGELOG.md
```diff
@@ -1,4 +1,171 @@
...
+
+## 3.8 Unreleased - 2020-08-??
+
```
tomspilman, line 4: Should we add some sort of summary to the release?

CHANGELOG.md
```diff
@@ -205,7 +372,7 @@
  - Automatic inclusion of XNBs into your final project on Mac and Linux.
  - Improved Mac and Linux installers.
  - Assemblies are now installed locally on Mac and Linux just like they are on Windows.
- - New cross-platform �Desktop� project where same binary and content will work on Windows, Linux and Mac desktops.
+ - New cross-platform “Desktop” project where same binary and content will work on Windows, Linux and Mac desktops.
```
tomspilman, line 375: Gross.... i will fix these Mac quotes.

CHANGELOG.md
harry-cpp: Changelog

SimonDarksideJ:
These more readable notes are a lot better @tomspilman

Jjagg:
I think generally the changelog isn't great because commit messages or PR titles tend to be more developer-oriented while the changelog is for people using MonoGame and should not be about implementation. For example a change I did had the message
```
Use short for parameter count in effect binary.
```
while the actual problem this fixed was #5995, so I changed the message to
```
Support arrays > 255 elements in shaders.
```
in SickheadGames/MonoGame#19. That conveys what's actually interesting to MG users. Things that are not interesting for MG users should not be in the changelog IMO (except big stuff like the Protobuild migration).
I also don't think documentation changes should go in the changelog.

I don't think we should spend hours going through these entries though. This will be better from now on with a manual changelog.

tomspilman:
@Jjagg @harry-cpp in the future on doc PRs like this we should always "squash and merge".  Don't think it is useful to have a dozen random doc edit commits in the history.

Issue #7046, opened 2020-02-29
Set custom GamePad deadzone


tdeeb:
I'm working on a 2D game that doesn't require precise inputs and relies on the virtual buttons (Ex. `Buttons.LeftThumbstickLeft`) to detect player movement when using controllers. With most of my controllers, the GamePad deadzone is too sensitive and makes it easy to make unintentional directional inputs. 

Unfortunately, [the GamePad deadzone values are hardcoded constants](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Input/GamePadThumbSticks.cs#L54). Even if `GamePadDeadZone.None` is used, the GamePad code still uses these constants to decide if the virtual buttons flags are set.

I'd like additional constructors in `GamePadThumbsticks` and `GamePadState` that take in custom left and right thumbstick deadzone values so this can be easily configured. This shouldn't any break existing code.

#### What version of MonoGame does the bug occur on:
- Develop

#### What operating system are you using:
- Windows, Linux

#### What MonoGame platform are you using:

- DesktopGL

Xonatron:
Could this be solved by making ApplyDeadZone() public and setting the dead zones values to be mutable (instead of const)? The idea is to let developers set the dead zones manually with the same functionality MonoGame uses to set them internally. Might be missing something.

mrhelmut:
I think that the best would be new overloads of ```GamePad.GetState()``` that takes deadzone values for both sticks and that cascade the value.

For instance, changing this one:

```GamePadState GetState(PlayerIndex playerIndex, GamePadDeadZone leftDeadZoneMode, GamePadDeadZone rightDeadZoneMode)```

into something like this:

```GamePadState GetState(PlayerIndex playerIndex, GamePadDeadZone leftDeadZoneMode, GamePadDeadZone rightDeadZoneMode, float leftDeadZoneAmount = -1.0f, float rightDeadZoneAmount = -1.0f)```

with negative values telling MonoGame to use the default internal constant.

PR #7278, opened 2020-08-03
Added constructor in GamePadThumbSticks that allows for custom deadzones

paytonrog:
To fix issue #7046

### DIFF
--- a/MonoGame.Framework/Input/GamePadThumbSticks.cs
+++ b/MonoGame.Framework/Input/GamePadThumbSticks.cs
@@ -42,10 +42,47 @@ namespace Microsoft.Xna.Framework.Input
             get { return _right; }
         }
 
+        public GamePadThumbSticks(Vector2 leftPosition, Vector2 rightPosition, float leftDeadZoneAmount, float rightDeadZoneAmount)
+            : this(leftPosition, rightPosition, GamePadDeadZone.None, GamePadDeadZone.None, leftDeadZoneAmount, rightDeadZoneAmount)
+        {
+            
+        }
+
         public GamePadThumbSticks(Vector2 leftPosition, Vector2 rightPosition)
             : this(leftPosition, rightPosition, GamePadDeadZone.None, GamePadDeadZone.None)
         {
-            
+
+        }
+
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Input/GamePadThumbSticks.cs
```diff
@@ -42,10 +42,47 @@ public Vector2 Right
             get { return _right; }
         }
 
+        public GamePadThumbSticks(Vector2 leftPosition, Vector2 rightPosition, float leftDeadZoneAmount, float rightDeadZoneAmount)
```
Jjagg, line 45: Since the default GamePadDeadZone is None, this constructor is not useful.

MonoGame.Framework/Input/GamePadThumbSticks.cs
```diff
@@ -42,10 +42,47 @@ public Vector2 Right
...
+        }
+
         public GamePadThumbSticks(Vector2 leftPosition, Vector2 rightPosition)
             : this(leftPosition, rightPosition, GamePadDeadZone.None, GamePadDeadZone.None)
         {
-            
+
+        }
+
+        internal GamePadThumbSticks(Vector2 leftPosition, Vector2 rightPosition, GamePadDeadZone leftDeadZoneMode, GamePadDeadZone rightDeadZoneMode, float leftDeadZone, float rightDeadZone) : this()
```
Jjagg, line 57: This should be public and documented

MonoGame.Framework/Input/GamePadThumbSticks.cs
```diff
@@ -42,10 +42,47 @@ public Vector2 Right
...
+        {
+            // Apply dead zone
```
Jjagg, line 59: This code should be extracted into a method so it's not duplicated.

MonoGame.Framework/Input/GamePadThumbSticks.cs
Jjagg: The parameter name doesn't match the usage below. I prefer `xDeadZone` or `xDeadZoneSize` to `xDeadZoneAmount`.

PR #7279, opened 2020-08-06
Tiny improvements to MGCB Editor and VS for Mac addin

harry-cpp:


Stiles-X:
I really think we should try to release 3.8 now, and put new issues under 3.9 milestone.

harry-cpp:
This is a PR, not an issue.

PR #7304, opened 2020-08-12
Add quotation marks around MGCBPath

### LINKED ISSUE #7303: MonoGame.Content.Builder.Task will not execute properly if MGCBPath contains a space
The MonoGame.Content.Builder.Task will not execute properly if MGCBPath contains a space, for example if the current user has a name that includes a space. For me MGCBPath is set to `C:\Users\FirstName LastName\.nuget\packages\monogame.content.builder.task\3.8.0.1641\build\\..\tools\netcoreapp3.1\any\mgcb.dll`. This results in an invalid command in the builder task since the MGCBPath is included without quotations.

- MonoGame 3.8

- Windows

bjornenalfa:
Fixes  #7303


harry-cpp:
Thanks @bjornenalfa , merging!

PR #7312, opened 2020-08-16
Updated to the latest version of StbSharp

rds1983:
This PR does two things:
1. Updates to the latest version of StbSharp and updates csprojs accordingly. As directory structure was changed slightly in the StbSharp.

2. Adds small optimization to the Texture2D.FromStream. Now the image file is copied into the memory only if incoming stream lacks Seek functionality.


Jjagg:
Thanks @rds1983!

PR #7322, opened 2020-08-19
Updated and added links to latest MonoGame 3.8 content in tutorials.md

SimonDarksideJ:
Refresh the YT links from my site to the latest MonoGame 3.8 content into the tutorials.md docs

### DIFF
--- a/Documentation/articles/tutorials.md
+++ b/Documentation/articles/tutorials.md
@@ -26,7 +26,10 @@ Check out the awesome work of the larger MonoGame community with their own tutor
 
 ## Darkside of MonoGame video series
 
-- [Getting Started with MonoGame on Visual Studio 2017](https://www.youtube.com/watch?v=zphaylhOrm0)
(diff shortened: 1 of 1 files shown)


tomspilman:
@Jjagg you mentioned DocFX has some thing to detect bad links right?  Maybe we can make it fail the build in that case?

SimonDarksideJ:
I believe it showed up in the build logs.  But I thought I checked all the other links when last I reviewed this list.

I'll double-check.

SimonDarksideJ:
Nope @tomspilman , just tested all the links on that page. Working fine.

Some are probably not relevant anymore, but that is another matter entirely :D

Jjagg:
Docfx checks internal references, but it does not ping web URLs.

tomspilman:
I bet we could knock up some C# using a regex we could write inline to the Cake script to check urls and report errors.

Issue #3856, opened 2015-05-15
SpriteBatch.Draw() applying random draw order to sprites with same layerDepth


aramisgrey:
Situation:
Draw 2 sprites on top of each other for 500 times. Sprite 2 is smaller than sprite 1, so you always clearly see sprite 1 behind sprite 2. Make sure not to overlap the drawing for each sprite pair so you can evaluate each pair-drawing.

Notice that: in some cases you won't see sprite 2 above sprite 1 but only sprite 1.

Reason: using Draw() with SpriteSortMode = BackToFront and blendState = AlphaBlend will not always honor the injection-order when sprites are drawn out, even if they all have the same layerDepth.

Engine: MonoGame 3.4.0.456

danzel:
We use List<T>.Sort to sort the sprites to draw
https://github.com/mono/MonoGame/blob/develop/MonoGame.Framework/Graphics/SpriteBatcher.cs#L212-L223

This is an unstable sort
https://msdn.microsoft.com/en-us/library/b0zbh7b6%28v=vs.110%29.aspx

If this is really needed then we could use a stable sort, but it'll be slower, quite a bit by the looks:
http://www.codeproject.com/Articles/26048/Fastest-In-Place-Stable-Sort

Could you see how this behaves in XNA? I suspect they would have the same problem. If you want consistent ordering, don't use BackToFront or give things different depths?

KonajuGames:
If you are asking it to sort back to front and then not giving it anything
sensible to sort with, it is not guaranteed to be in the same order as you
called Draw().​  If you want sorted rendering, then you must give it
something to sort with.

Jjagg:
I don't consider this a bug, I think it's fine we use an unstable sort. But we should probably document that, so I'm leaving this open for now.

PR #7323, opened 2020-08-20
Explicitly document BackToFront and FrontToBack use an unstable sort.

goosenoises:
#3856

If two sprites have the same depth, their draw order isn't guaranteed to be preserved when sorting with `SpriteSortMode.BackToFront` or `SpriteSortMode.FrontToBack`. Using a stable sort is expensive, but as @Jjagg mentioned, wanted to document this for easier reference.

(Side note that I'm relatively new to contributing to OSS, please let me know if I'm missing anything.)


Jjagg:
Thank you! :)

PR #7326, opened 2020-08-21
Text composition support enhanced

ryancheung:
This is an new version of #7118

This is the PR mentioned before in the following issue: https://github.com/MonoGame/MonoGame/issues/6347#issuecomment-604898284

A abstract class `ImmService` is added MonoGame.Framework/Input folder. This service is going to provide text input and composition APIs for SDL2, WindowsDX, Windows Universal, Android and iOS platforms. More platforms like (PlayStation 4, Nintendo Switch etc.) could be added in the future by someone, if supported.

Platform specific ImeService implementations to be implemented:

- WinFormsImeHandler.cs
- SdlImeHandler.cs
- AndroidImeHandler.cs
- iOSImeHandler.cs
- UapImeHandler.cs

WindowsDX IME handler is implemented completely by utilizing the [ImeSharp](https://github.com/ryancheung/ImeSharp) library written by me. And I request MG maintainers to weigh in the library.

Edit: [Here](https://github.com/ryancheung/MonoGameTextCompositionSamples) is a demo of this PR.

Edit: *Windows Universal* IME support just added.


ryancheung:
@Jjagg I just intended to add CoreRT support for this PR. As this comment https://github.com/dotnet/corert/issues/4219#issuecomment-431106826 and https://github.com/dotnet/corert/issues/4219#issuecomment-587389324 noted that Com-Interop is not supported in CoreRT and it is possible to use https://github.com/SharpGenTools/SharpGenTools to make COM work in CoreRT. Could you please make a basic TSF COM API sets that ImeSharp required with SharpGenTools?

Jjagg:
I'm leaving on holiday for a couple weeks next Monday so it'll be a while before I get to that.

If you want to handle it yourself before I get back, check out the Mapping.xml file in https://github.com/Jjagg/TsfSharp.
Following the structure for the existing bindings you should be able to add what you need.

ryancheung:
OK, thanks, I'll try it.

PR #7327, opened 2020-08-22
Update mgfxc_wine_setup.sh

Kwyrky:
Make sure we don't mess up .profile / .zprofile files which could have text or commands in the last line. In this case the export command would be appended directly to the last command and not end up in its own line.


Kwyrky:
I added the `-e` flag as the change request suggested.

harry-cpp:
Thanks @Kwyrky, merging!

harry-cpp:
On a random note, did you know that a text file without a new line character at the end is technically invalid on Unix :)

Kwyrky:
Nice, thanks for merging! That is what I assumed and wrote I think in the community forums ^^

PR #7328, opened 2020-08-23
Kwyrky ensure newlines

Kwyrky:


Kwyrky:
This is a duplicate and can be closed. The issue is already handled in pull request https://github.com/MonoGame/MonoGame/pull/7327

PR #7338, opened 2020-08-29
Fix multisampling in DesktopGL

sk-zk:
This is a quick and dirty fix to make multisampling work in DesktopGL.

The problem seems to be that OpenGL doesn't let you change the multisampling params after the window has been created, so they need to be passed to SDL before that point, in `PlatformInitialize()`. There was no way for the user to do that though because setting params via `graphics.GraphicsDevice` or through the `PreparingDeviceSettings` event are too late since `PlatformInitialize()` has already run. To deal with this, I added another event that is raised right before `PlatformInitialize()` so the user can pass the multisample count at that point.

The game code for enabling multisampling would look like this:

```csharp
        public Game1()
        {
            graphics.PreferMultiSampling = true;
            graphics.BeforeDeviceCreated += Graphics_BeforeDeviceCreated;
        }

        private void Graphics_BeforeDeviceCreated(object sender, PresentationParameters e)
        {
            e.MultiSampleCount = 16;
        }
```

### DIFF
--- a/MonoGame.Framework/GraphicsDeviceManager.cs
+++ b/MonoGame.Framework/GraphicsDeviceManager.cs
@@ -32,7 +32,6 @@ namespace Microsoft.Xna.Framework
         private GraphicsProfile _graphicsProfile;
         // dirty flag for ApplyChanges
         private bool _shouldApplyChanges;
-        private int? _initMultiSampleCount;
 
         /// <summary>
         /// The default back buffer width.
@@ -97,12 +96,6 @@ namespace Microsoft.Xna.Framework
             _game.Services.AddService(typeof(IGraphicsDeviceService), this);
         }
 
-        public GraphicsDeviceManager(Game game, int multiSampleCount) : this(game)
-        {
-            _initMultiSampleCount = multiSampleCount;
-            PreferMultiSampling = true;
-        }
-
         ~GraphicsDeviceManager()
         {
             Dispose(false);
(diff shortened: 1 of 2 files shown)


harry-cpp:
Couldn't you pass the `MultiSampleCount` in constructor?

sk-zk:
You could, but then you'd have to store that in a member variable until CreateDevice is called. Idk if that's a good idea.

harry-cpp:
Everything else in `GraphicsDeviceManager.SDL.cs` works like that.

Constructor is the only thing that happens before the main window gets created and some other window related things need to be set from it. We were gonna implement the window recreation at some point, but that was/is too low on the priority list.

sk-zk:
I see. I've updated the PR

harry-cpp:
Actually you have the `PreparingDeviceSettings` event to modify the params before device creation, so you just need the gdm.sdl.cs changes, right?

harry-cpp:
Unless its a partial void with half of the method missing XD

sk-zk:
PreparingDeviceSettings happens after PlatformInitialize, that's the issue

harry-cpp:
Yea, I now see why it was so problematic, modify CreateDevice in GDM to do the following, that should do the trick:
```
var gdi = DoPreparingDeviceSettings();

if (!_initialized)
    Initialize();

CreateDevice(gdi);
```

sk-zk:
That'll do it

harry-cpp:
Big thanks @sk-zk ! Merging!

OtherStuffThatSpam:
I've been testing Monogame 3.8.1.303 and I am not able to enable multi-sampling on Windows DesktopGL Cross platform project, the same code base works on DirectX Windows, is this really fixed in the latest version of Monogame? or I am missing something?

lorissikora:
Sorry for necrobumping, but I'm experiencing the exact same issue, also on version 3.8.1.303. Has there been any development on this? @OtherStuffThatSpam did you find a solution?

orosbogdan:
Experiencing the same issue on 3.8.1.303

PR #7340, opened 2020-09-01
Fix shader error messages for OpenGL

cpt-max:
If the capacity of the StringBuilder is not set, the error messages will be cut short to the default capacity, and memory behind the buffer get's corrupted.


Jjagg:
Thanks @cpt-max! :)

PR #7344, opened 2020-09-02
Added IEffectBones interface, implemented by SkinnedEffect.

vpenades:
This PR adds a convenience interface IEffectBones, which must be implemented by all shaders supporting skinning.

For years, MonoGame has relied on the good old SkinnedEffect, but as new skinned effects are on the way, a generic way of setting the bone matrices is required, in the same way as IEffectTransform, IEffectLights, etc.

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/IEffectBones.cs
+++ b/MonoGame.Framework/Graphics/Effect/IEffectBones.cs
@@ -1,51 +1,14 @@
-﻿// #region License
-// /*
-// Microsoft Public License (Ms-PL)
-// MonoGame - Copyright © 2009 The MonoGame Team
-// 
-// All rights reserved.
-// 
-// This license governs use of the accompanying software. If you use the software, you accept this license. If you do not
-// accept the license, do not use the software.
-// 
-// 1. Definitions
-// The terms "reproduce," "reproduction," "derivative works," and "distribution" have the same meaning here as under 
-// U.S. copyright law.
-// 
-// A "contribution" is the original software, or any additions or changes to the software.
-// A "contributor" is any person that distributes its contribution under this license.
-// "Licensed patents" are a contributor's patent claims that read directly on its contribution.
-// 
-// 2. Grant of Rights
-// (A) Copyright Grant- Subject to the terms of this license, including the license conditions and limitations in section 3, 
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Graphics/Effect/IEffectBones.cs
```diff
@@ -0,0 +1,56 @@
...
+// (D) If you distribute any portion of the software in source code form, you may do so only under this license by including 
+// a complete copy of this license with your distribution. If you distribute any portion of the software in compiled or object 
+// code form, you may only do so under a license that complies with this license.
+// (E) The software is licensed "as-is." You bear the risk of using it. The contributors give no express warranties, guarantees
+// or conditions. You may have additional consumer rights under your local laws which this license cannot change. To the extent
+// permitted under your local laws, the contributors exclude the implied warranties of merchantability, fitness for a particular
+// purpose and non-infringement.
+// */
+// #endregion License
+//
```
Jjagg, line 40: Please use the newer license text:
```
```

MonoGame.Framework/Graphics/Effect/IEffectBones.cs
```diff
@@ -0,0 +1,56 @@
...
+// #endregion License
+// 
+using System;
+
+namespace Microsoft.Xna.Framework.Graphics
+{
+    /// <summary>
+    /// The common effect bones transforms parameters.
+    /// </summary>
+	public interface IEffectBones
```
Jjagg, line 48:
```suggestion
    public interface IEffectBones
```

MonoGame.Framework/Graphics/Effect/IEffectBones.cs
```diff
@@ -0,0 +1,56 @@
...
+// purpose and non-infringement.
+// */
+// #endregion License
+// 
+using System;
+
+namespace Microsoft.Xna.Framework.Graphics
+{
+    /// <summary>
+    /// The common effect bones transforms parameters.
```
Jjagg, line 46: Ah, I see you made this consistent with other effect interfaces. Those docs aren’t great though, so let’a not mimic them. Maybe “Interface for Effects that support bone transforms.”?.

Jjagg (changes requested):
Some nits

vpenades:
I've just fixed the docs and license.

Jjagg:
Great, thanks! :)

Issue #7193, opened 2020-06-10
Calling DrawIndexedPrimitives and DrawInstancedPrimitives in succession causes problems.


OpenPachinko:
It may not occur depending on the calling sequence.
The behavior seems to be different between Android and DesktopGL.

I made a program to reproduce the problem.
https://github.com/OpenPachinko/MonogameCompareBehaviors

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.0.1375-develop

#### What operating system are you using:
- MacOS
- Android

#### What MonoGame platform are you using:

- DesktopGL
- Android

OpenPachinko:
After drawing, I tried saving it to a file with RenderTarget2D.SaveAsPng().
I did not notice it visually, but the first frame is normal.
The problem seems to occur from the second frame.

I used ObjectDumper.NET to get a dump of GraphicsDevice for the first and second frames and compared them, but there was no difference.

OpenPachinko:
I modified GraphicsDevice.OpenGL.cs a little.
DesktopGL has improved a lot.
Android is still useless.

My main platform is Android,
Android is troubled because it cannot be solved by brute force...
```
        private void ApplyAttribs(Shader shader, int baseVertex)
        {
            var programHash = ShaderProgramHash;
            var bindingsChanged = false;

            for (var slot = 0; slot < _vertexBuffers.Count; slot++)
            {
                var vertexBufferBinding = _vertexBuffers.Get(slot);
                var vertexDeclaration = vertexBufferBinding.VertexBuffer.VertexDeclaration;
                var attrInfo = vertexDeclaration.GetAttributeInfo(shader, programHash);

                var vertexStride = vertexDeclaration.VertexStride;
                var offset = (IntPtr)(vertexDeclaration.VertexStride * (baseVertex + vertexBufferBinding.VertexOffset));

                //if (!_attribsDirty)
                //    if (_bufferBindingInfos[slot].VertexOffset == offset)
                //        if (ReferenceEquals(_bufferBindingInfos[slot].AttributeInfo, attrInfo))
                //            if (_bufferBindingInfos[slot].InstanceFrequency == vertexBufferBinding.InstanceFrequency)
                //                if (_bufferBindingInfos[slot].Vbo == vertexBufferBinding.VertexBuffer.vbo)
                //                    continue;
            

                bindingsChanged = true;

[...]

PR #7352, opened 2020-09-09
Hull, Domain and Geometry Shaders for DirectX and OpenGL

cpt-max:
This PR is based on my previous PR #7242 which was DirectX only. Now those new shader stages also work in OpenGL. A switch from MojoShader to ShaderConductor was neccessary to make this happen, so PR #7345 is also included.

Here you find a  [sample project](https://github.com/cpt-max/ShaderTest) containing a simple Hull, Domain and Geometry shader.
This [shader migration document](https://github.com/cpt-max/Docs/blob/master/Migrating%20shaders%20to%20ShaderConductor.md) explains what changes need to be made to old shader code. 
Here are some more details about the [shader model support in OpenGL](https://github.com/cpt-max/Docs/blob/master/Shader%20models%20for%20OpenGL.md).

Anyone wanting to try out this branch, please use #7533 instead. It fully contains this branch and is kept up to date.


cpt-max:
When I added texture sampling support for hull, domain and geometry shaders, it made sense to also do it for the vertex shader. So VTF (vertex texture fetch) will now also work in OpenGL, before it only worked in DirectX.

tomspilman:
Thanks for all this @cpt-max !

Bear with us as we need to review this carefully and get it merged after we get this 3.8.1 release out.

cpt-max:
> Bear with us as we need to review this carefully and get it merged after we get this 3.8.1 release out.

Of course, no hurry.

squarebananas:
Thanks for your work on adding this functionality. Last week I tried using this with one of my projects and it's working well so far. In particular I was after VTF in OpenGL as it will greatly help performance and also saves rewriting game code that already works on DirectX platforms.

The things I noted while implementing this with my project were as follows:

- With MojoShader at various points in shaders I had to pointlessly rearrange things in order for it to build the OpenGL shaders. With ShaderConductor these rearrangements were no longer necessary making it much easier to port from HLSL.

- When texture sampling in the vertex shader "tex2Dlod" needs to be replaced with "MyTexture.SampleLevel" function to work on OpenGL.

- In shader vertex structures a float4x4 parameter needs replacing with multiple float4 parameters and the element usage index number can't be left out. For example the XNA instancing sample (https://github.com/SimonDarksideJ/XNAGameStudio/blob/archive/Samples/InstancedModelSample_4_0/InstancedModelSample/InstancedModelSample/Content/InstancedModel.fx) used "float4x4 instanceTransform : BLENDWEIGHT" and this will no longer work. Using "float4 instanceTransform1 : BLENDWEIGHT0" and so on for each row will work however.

- With one of my effects I had 1 sampler with 6 possible textures and an if statement selects the correct texture to use from a vertex input value. This resulted in the wrong textures being sampled and textures switching when different models were in view. The workaround for this was to use a separate sampler for each texture but something might need fixing here.

- ShaderConductor no longer adds the sampler name into the effect parameter name. For example "SamplerState0+xTexture0" becomes "xTexture0". This means for both DirectX and OpenGL platforms the same effect parameter names can now be used in game code.

- For instancing with multiple vertex buffer shaders the shader.GetAttribLocation function now returns different values. This is because MojoShader ordered the second buffer's attributes before the first buffer whereas ShaderConductor seems to do the opposite. For my project this change of order happened to also fix the issue mentioned at #7193. However whether this fix works depends entirely on the shader as the example project in the issue isn't fixed by this change of order (slightly unrelated but forcing cache bindings to be ignored when a different number of buffers has been set will fix this).

- When building shaders with the MGCB Editor, if the build is successful the warnings aren't shown at all. With MojoShader the warnings are present regardless of whether the build was successful.

- When creating a new default effect using the MGCB Editor it fails to build. The template effect that needs updating is located at Tools\MonoGame.Content.Builder.Editor\Templates\Effect.fx

Let me know if you need any more information about these things.

Not exactly related but I couldn't figure out how to add the tools projects into my own project for debugging while building content. Instead I used the updated editor to build the content and then copied/pasted the files into the bin folder. If I could figure out how to do this I could look into some of the issues more. Is there a guide anywhere for referencing the tools projects directly?

Thanks again.

cpt-max:
Great feedback, thanks!

>  When texture sampling in the vertex shader "tex2Dlod" needs to be replaced with "MyTexture.SampleLevel" function to work on OpenGL.

I mentioned this in the [migration document](https://github.com/cpt-max/Docs/blob/master/Migrating%20shaders%20to%20ShaderConductor.md) in #7345, which this PR is based on.

>  In shader vertex structures a float4x4 parameter needs replacing with multiple float4 parameters and the element usage index number can't be left out. For example the XNA instancing sample (https://github.com/SimonDarksideJ/XNAGameStudio/blob/archive/Samples/InstancedModelSample_4_0/InstancedModelSample/InstancedModelSample/Content/InstancedModel.fx) used "float4x4 instanceTransform : BLENDWEIGHT" and this will no longer work. Using "float4 instanceTransform1 : BLENDWEIGHT0" and so on for each row will work however.

@minimalism mentioned this in #7345. I couldn't reproduce it, but it looks like I misunderstood. I think I tested passing a float4x4 from vertex to pixel shader, but this issue seems to be about vertex shader inputs. 

> With one of my effects I had 1 sampler with 6 possible textures and an if statement selects the correct texture to use from a vertex input value. This resulted in the wrong textures being sampled and textures switching when different models were in view. The workaround for this was to use a separate sampler for each texture but something might need fixing here.

Does this work with MojoShader? Both MojoShader and ShaderConductor create a combined sampler+texture object for every sampler-texture combination they find. If the texture is not determined at compile time (if I understand you correctly), then I don't see how either of them can succeed. Do you have a code snippet showing what you were doing exactly?

> ShaderConductor no longer adds the sampler name into the effect parameter name. For example "SamplerState0+xTexture0" becomes "xTexture0". This means for both DirectX and OpenGL platforms the same effect parameter names can now be used in game code.

Thank god for that. The texture name change is a byproduct of the sampler-texture combo I mentioned above. 

>  For instancing with multiple vertex buffer shaders the shader.GetAttribLocation function now returns different values. This is because MojoShader ordered the second buffer's attributes before the first buffer whereas ShaderConductor seems to do the opposite. For my project this change of order happened to also fix the issue mentioned at #7193. However whether this fix works depends entirely on the shader as the example project in the issue isn't fixed by this change of order (slightly unrelated but forcing cache bindings to be ignored when a different number of buffers has been set will fix this).

Does that mean you have to use different usage indices compared to MojoShader? I guess the goal should be to be consistent with DirectX, and that should be the case, right?

>  When building shaders with the MGCB Editor, if the build is successful the warnings aren't shown at all. With MojoShader the warnings are present regardless of whether the build was successful.

That's true. There's an `if (result.hasError) { //output errors }`  in ShaderData.conductor.cs that causes this. The fix should be easy. Errors and warnings have to be read unconditionally and passed to the errorsAndWarnings parameter in OpenGLShaderProfile.CreateShader(), the same way it's done for MojoShader.

>  When creating a new default effect using the MGCB Editor it fails to build. The template effect that needs updating is located at Tools\MonoGame.Content.Builder.Editor\Templates\Effect.fx

Totally didn't think about that, of course this template needs to be converted.

> Not exactly related but I couldn't figure out how to add the tools projects into my own project for debugging while building content. Instead I used the updated editor to build the content and then copied/pasted the files into the bin folder. If I could figure out how to do this I could look into some of the issues more. Is there a guide anywhere for referencing the tools projects directly?

Why can't you debug using the editor? When I debug shader compilation stuff, I either use the MGCB editor and throw a System.Diagnostics.Debugger.Launch() into the MGFXC project, or, most of the time, I launch MGFXC directly having the fx file as a command line argument (Project properies -> Debug -> Application arguments): `c:\sourceFile.fx c:\outputFile.mgfxo /Profile:OpenGL`

squarebananas:
Thanks for the information. Your suggestions with the editor and MGFXC worked so I'll try them out some more.

> Does this work with MojoShader? Both MojoShader and ShaderConductor create a combined sampler+texture object for every sampler-texture combination they find. If the texture is not determined at compile time (if I understand you correctly), then I don't see how either of them can succeed. Do you have a code snippet showing what you were doing exactly?

I've just put together a simple project which demostrates this issue here: https://github.com/squarebananas/SharedSamplerTest
With the current build of Monogame using MojoShader you should see a red and green texture on the screen. After swapping to this version with ShaderConstruct, rebuilding the effect and removing the sampler name from the effect parameters, both textures change to green when it seems they shouldn't. I will look into this some more.

> Does that mean you have to use different usage indices compared to MojoShader? I guess the goal should be to be consistent with DirectX, and that should be the case, right?

The usage indices stayed the same. The values for Shader.VertexAttribute.location change though. I noticed this had an effect on whether the existing instancing bug occurs but I'm not sure if it could change anything else.

Thanks again.

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.OpenGL.cs
```diff
@@ -12,57 +13,194 @@ namespace MonoGame.Effect
...
+                                        uniqueSamplerSlot++;
+
+                                    samplerSlotMapping.Add((sampler.samplerSlot, sampler.textureSlot), uniqueSamplerSlot);
+                                }
+                            }
+                        }
+
+                        // assign new slots in 2nd pass
+                        if (pass == 1)  
+                            sampler.samplerSlot = uniqueSamplerSlot;
```
squarebananas, line 184: This is the line causing the shared texture issue. Setting the value here doesn't update the struct value within the array. Changing this to "shader._samplers[i].samplerSlot = uniqueSamplerSlot;" will fix this.

cpt-max: Oh yeah, sampler is a struct. Thanks for fixing this!

cpt-max:
> The usage indices stayed the same. The values for Shader.VertexAttribute.location change though. I noticed this had an effect on whether the existing instancing bug occurs but I'm not sure if it could change anything else.

I would need to look into this to find out what exactly is going on here, right now I don't have time for that.

squarebananas:
> 
> 
> > The usage indices stayed the same. The values for Shader.VertexAttribute.location change though. I noticed this had an effect on whether the existing instancing bug occurs but I'm not sure if it could change anything else.
> 
> I would need to look into this to find out what exactly is going on here, right now I don't have time for that.

No problem, I was just noting a difference in the internal behavior and I don't think it will cause any issues. The only noticeable difference was due to an existing Monogame bug which will disappear once fixed and fixing that is really unrelated to this.

eylvisaker:
I also ran into the issue with the float4x4 parameters. (I have a repro [here](https://github.com/eylvisaker/DeferredRenderingDemo/tree/shadows-mgupdate)). I wouldn't mind poking around in the source to see if I could fix that, if anyone can offer an idea of what might be wrong.

cpt-max:
It's possible that this is a row-major/column-major issue. Maybe transposing the matrices before putting them in the vertex buffer fixes it.

eylvisaker:
It's a good thought, that actually is a part of the problem (More on that below).

The core issue is in [VertexDeclaration.OpenGL.cs](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Graphics/Vertices/VertexDeclaration.OpenGL.cs)]. There's a comment around line 27 that `XNA appears to ignore usages it can't find a match for, so we will do the same`, and the code following it causes it to skip setting up a vertex attribute info structure for BlendWeights1,2,3, and ultimately leads to only a single call made to `glVertexAttribPointer` for BlendWeights0. But when I look at OpenGL examples for instancing, like [this one](https://learnopengl.com/Advanced-OpenGL/Instancing) I see four calls to glVertexAttribPointer to set up each row (or col?) of the matrix. 

I coded up a really simple [fix here](https://github.com/MonoGame/MonoGame/commit/8b14f0898d1a723e8440ccf9ecb91ab7ec0ece5f). This makes an assumption that if you have e.g. BlendWeights0-3 that it should assign BlendWeights1-3 to the slots right after BlendWeights0, if it doesn't find an attribute location for them. There are probably ways for this to go horribly wrong. But it at least fixes the problem in my case.

But that then leads to another problem. There is a row/column major mismatch between D3D and OGL. D3D needs the matrix to be transposed but OpenGL does not. So either the CPU or the GPU code needs to determine which graphics API is being used to decide whether to transpose. With MojoShader I could use `#if HLSL` to do this, but it doesn't seem to work with ShaderConductor.

cpt-max:
> With MojoShader I could use `#if HLSL` to do this, but it doesn't seem to work with ShaderConductor.

Those defines should still be available. The DirectX shader profile adds the HLSL define in the AddMacros function [here](https://github.com/cpt-max/MonoGame/blob/glshaderstages/Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.DirectX.cs), and the OpenGL profile adds the GLSL and OPENGL defines [here](https://github.com/cpt-max/MonoGame/blob/glshaderstages/Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.OpenGL.cs)

eylvisaker:
Ah, you're right, I must have done something silly the first time I tried it.

I was thinking my fix for the matrix vertex parameter would be a lot more robust if it could inspect the datatype or size of the parameter before making the decision to assign unmatched parameters to vertex attribute slots. But the vertex attribute structure doesn't store any information about that, so I think it would need to be added. To pipe that information through, it looks like GetStageInputs in ShaderConductor.cs would need to get something about that from ShaderConductor and add it to the StageInput structure. I'm not really sure how to do that, it looks like we would need something like GetParameter for a stage input.

cpt-max:
It should be possible to add some extra size information to stage inputs. [Here](https://github.com/cpt-max/ShaderConductor/blob/fe4c90c4b7f5baa7b8f00a5eb65a37e9f38c9286/Source/Core/ShaderConductor.cpp#L514) is the ShaderConductor code that reflects on stage inputs.
The project I'm working on right now should be coming to an end soon, so I might be able to look into this then.

eylvisaker:
So I found two other regressions with this. 

One is with setting matrix array parameters, [simple fix here](https://github.com/eylvisaker/MonoGame/commit/bae1c39719146b8002f23a6f35f05cb04b9acc8c).. would it make sense for me to submit a PR for this to your fork, or would it make more sense to wait until this PR is merged into MonoGame and make a separate PR for that fix?

The other problem is a bit more complex. The code I'm working on uses SpriteBatch to render some things after doing some DrawPrimitives calls, using the same effect for both. It looks like what's happening is that it looks like when I compile an effect with multiple Texture2D parameters, ShaderConductor begins assigning them starting with register t1 instead of register t0. This breaks on the SpriteBatch calls because SpriteBatch always assigns its texture parameter to register t0. If I add `: register(t0)` to my texture parameter in my effect file, then my DrawPrimitives calls break. Note that all these scenarios work with Direct3D. From inspecting the state of the graphics device, it looks like the textures are being assigned to the correct registers, so my best guess is that the shader itself is not reading from the correct registers. 

When I look at the generated glsl in the Nsight debugger, the samplers are not declared with any assigned registers. I found [this issue](https://github.com/KhronosGroup/SPIRV-Cross/issues/584) which looks like it might be relevant. If I have some time this weekend I'll investigate further.

cpt-max:
> would it make sense for me to submit a PR for this to your fork

Sure, that way the fix is immediately available to everyone.

> ShaderConductor begins assigning them starting with register t1 instead of register t0

I think it wants to keep t0 open for the default texture. OpenGL projects often have a default texture permanently at t0.   

> When I look at the generated glsl in the Nsight debugger, the samplers are not declared with any assigned registers.

The register slots are only used MonoGame internally. They are really just defining the index for the GraphicsDevice.Textures array.

> If I add `: register(t0)` to my texture parameter in my effect file, then my DrawPrimitives calls break.

This should work. In which way do the DrawPrimitives calls break? 

In the current implementation there are some fundamental differences btw DirectX and OpenGL with regards to textures and samplers. OpenGL (before 3.1 or so) doesn't separate samplers and textures. The sampler parameters are attached to the texture object. This is really odd, it means when you want to sample the same texture from multiple samplers, you have to make multiple copies of the same texture object. I don't think MojoShader ever did that, so this has always been a problem. 

I'm not sure if your problem is related to this, but if it is, the solution would be the SeparateSamplerObjects OGL extension introduced in OGL 3.1. I already have a [branch](https://github.com/cpt-max/MonoGame/tree/samplerobjects) for this. I didn't merge this yet, because it might be missing a capability check, or two.

eylvisaker:
> 
> I think it wants to keep t0 open for the default texture. OpenGL projects often have a default texture permanently at t0.
> 
This seems to be at odds with how SpriteBatch works in MonoGame 3.8.

> The register slots are only used MonoGame internally. They are really just defining the index for the GraphicsDevice.Textures array.

But they also need to be used by the shader so that it reads from the correct texture index.

> > If I add `: register(t0)` to my texture parameter in my effect file, then my DrawPrimitives calls break.
> 
> This should work. In which way do the DrawPrimitives calls break?
 
They just don't draw anything at all. I've tried debugging it in Nsight, but all it tells me is that no fragments were drawn, but I'm having trouble seeing why, because it looks like the correct textures are bound, but I'm not really sure how to read the location/value fields of its outputs.

I've put together a repro here: https://github.com/eylvisaker/ShaderTester

It has four projects that all use basically the same GPU+CPU code. Two each for Direct3D and OpenGL with MonoGame 3.8 and MonoGame+ShaderConductor. The Direct3D code works the same on both versions of MonoGame, but the OpenGL code works differently. Without register assignments, the SpriteBatch+effect code works correctly in OpenGL+Mojo (but you have to do that ugly texture parameter name mangling). Using register assignments behaves funky with MG 3.8, but with this ShaderConductor branch it simply doesn't draw anything. 

> In the current implementation there are some fundamental differences btw DirectX and OpenGL with regards to textures and samplers. OpenGL (before 3.1 or so) doesn't separate samplers and textures. The sampler parameters are attached to the texture object. This is really odd, it means when you want to sample the same texture from multiple samplers, you have to make multiple copies of the same texture object. I don't think MojoShader ever did that, so this has always been a problem.
> 
> I'm not sure if your problem is related to this, but if it is, the solution would be the SeparateSamplerObjects OGL extension introduced in OGL 3.1. I already have a [branch](https://github.com/cpt-max/MonoGame/tree/samplerobjects) for this. I didn't merge this yet, because it might be missing a capability check, or two.

I'll check this out, and see if it changes anything in any of my tests. Is there anything that needs to be done in the code to use the SeparateSamplerObjects extension?

eylvisaker:
Actually, I think a better solution for SpriteBatch would be to provide a means for to bypass the assumption that it should be setting the first texture parameter. I'll create a separate PR for that. I'd still like to figure out why setting texture registers causes DrawPrimitives to do nothing for OpenGL though.

cpt-max:
> Is there anything that needs to be done in the code to use the SeparateSamplerObjects extension?

No, it will automatically create and use an OpenGL sampler object for every sampler state.

minimalism:
Thought I should mention I couldn't get this to build shaders on my minspec GPUs (an Intel integrated card and a Radeon M9 370X, the Macbook 2015 one) even though they ostensibly support OpenGL>4.0. I wasn't able to get any relevant error messages. 
I also struggled to downgrade my shaders to SM3, repeatedly running into various unimplemented features in SPIRV - for example, the % operator wouldn't build because it insisted on translating them to the unimplemented Frem function instead of just falling back to Fmod. I got the impression that SPIRV is a bit lacking in supporting legacy GL/SM, but there's been a lot of work done since then so it might be time to try again. 

Apologies for the unconstructive reply but I imagine most folks will be focusing on testing the newer shader models and GL4 features which SPIRV appears to handle well, but I think many monogame devs want to support older hardware too. I wasn't able to do so without reverting back to mojoshader for those devices, which is quite reliable when it comes to building legacy shaders. I'll give this another try with the latest ShaderConductor and SPIRV builds soon and hopefully come up with some solutions.

cpt-max:
> Thought I should mention I couldn't get this to build shaders on my minspec GPUs (an Intel HD4000 and a Radeon M9 370X, the Macbook 2015 one)

It fails at runtime, not during content building, right? I assume you get an exception in Shader.GetShaderHandle() in Shader.OpenGL.cs. You really get nothing useful from GL.GetShaderInfoLog()?

> I got the impression that SPIRV is a bit lacking in supporting legacy GL/SM

You got the right impression. With ShaderConductor you really want to use shader model 4 or higher. Shader model 2 and 3 will map to OpenGL 2, which is just not that well supported. 
I managed to get the default shaders working for OpenGL 2, so I guess it's not completely useless, but you definitely don't want to use it for extensive shader development.

> I think many monogame devs want to support older hardware too.

All you really need is OpenGL 3.3 (released in 2010), you really think there's a lot of demand for even older hardware?

cpt-max:
A little update, as I finally have some time for MonoGame stuff again. I addressed a few issues mentioned above:

- Updated to newest MonoGame version.
- Improved error and warning output.
- Merged @eylvisaker fix for matrix array parameters (sorry, I totally missed this PR when you first made it).
- Updated MGCB editor template effects to work with ShaderConductor.
- Looked into the float4x4 issue with vertex shader inputs:

The root of the float4x4 problem is this: 
The C# vertex structure contains multiple separate vectors to hold the entire matrix. If you do the same in the vertex shader (use multiple vectors instead of a single matrix) everything is fine, because now there is match between vertex elements and shader inputs. 
However, if you use a single matrix parameter for the vertex shader, it won't work, because now vertex elements and shader inputs don't match. This worked with MojoShader, because it would split the matrix parameter into vectors automatically. ShaderConductor doesn't do that, it will create a single matrix parameter:
```glsl
layout(location = 3) in mat4 in_var_BLENDWEIGHT0;
```

The easiest solution would be if ShaderConductor could flatten the matrix, just like MojoShader does. I made a [request on the SPIRV-Cross GitHub](https://github.com/KhronosGroup/SPIRV-Cross/issues/1695).

cpt-max:
I just noticed that ShaderConductor actually has a flag to switch between row-major and column-major matrices, this allowed me to remove the manual matrix transposing for OpenGL. I could slap myself for not seeing this earlier.

cpt-max:
Matrices should work now as vertex shader inputs. I had to store the matrix size in the effects file for this, so effects need to be recompiled.
 
Unfortunately I found a problem with the row-major/column-major switch I made recently (https://github.com/MonoGame/MonoGame/pull/7352/commits/7fc7d9b710b66751693cbffff47c7d9a3f08c8d1)
It only works with square matrices currently. I may have to revert this. 
SkinnedEffect.fx from the default shaders fails to compile because of this, it uses a 4x3 matrix.

cpt-max:
Apparently the square matrix limitation only applies to OpenGL 2 (shader models 2 and 3).
Support for those shader models isn't that great anyway right now, so I'm not planning to look for a solution to this. I added the limitation to the migration document, and changed the SkinnedEffect shader, so it doesn't use non-square matrices anymore.

MrScautHD:
i can not wait anymore :( it looks so amazing!

cpt-max:

PR #7361, opened 2020-09-27
DesktopGL support for Texture2DArray

minimalism:
Allocate array texture in C#: 
`List<byte[]> bytesList = Foo.GetRawBytesForEachArraySlice();
Texture2D = new Texture2D(device, Width, Height, false, SurfaceFormat.Color, bytesList.Count);
for (int i = 0; i < bytesList.Count; i++) {
    Texture2D.SetData(0, i, null, bytesList[i], 0, bytesList[i].Length);
}`

Use it in a shader:
`Texture2DArray ArrayTexture;
sampler ArraySampler = sampler_state {
	Texture = <ArrayTexture>;
};

PSOut PS(PSIn) {
	float4 texColor = ArrayTexture.Sample(ArraySampler, PSIn.TexCoords.xyz);
}`
This is based off @cpt-max PR here: https://github.com/MonoGame/MonoGame/pull/7352
There are no overlapping edits with the ShaderConductor branch so the relevant changes ( https://github.com/MonoGame/MonoGame/commit/f7b640ee360aec2a973cb82567004e17f638a6eb ) can be cherrypicked, but attempting to compile such a shader with MojoShader gives the error "Unsupported texture type for ps_3_0" and raising the shader model to 4_0 gives "Shader must be SM 3.0 or lower!"

Not Implemented:
- Compressed texture arrays. I'm not really familiar with these, are they useful?
- PlatformGetData<T> does not support the TextureArray case yet


minimalism:
Found a potential issue with this: Binding a Texture2DArray to a Texture2D slot appears to fail silently. On my hardware the glBindTexture doesn't generate an error but also doesn't seem to change any state. 

To me it seems friendlier to throw an exception if the user does this, but I'm a little out of my depth on what this would entail. I suppose we'd have to add a Texture2DArray value to the EffectParameterType enum. On the CPU side of things it's easy enough to see if a Texture2D instance represents an array texture or not.

eylvisaker:
I've been fiddling with this to get cascaded shadow maps working. I had a problem where this doesn't correctly bind the array slice when using render target array. I have a [fix here](https://github.com/eylvisaker/MonoGame/commit/5f758aa075c62fe34361fd4cbc7e454a617bc355) and a [demo here](https://github.com/eylvisaker/DeferredRenderingDemo).

SimonDarksideJ:
Any chance you can rebase your PR to resolve the merge conflicts @minimalism ?
Is this PR still valid

squarebananas:
> Any chance you can rebase your PR to resolve the merge conflicts @minimalism ? Is this PR still valid

- It looks like this (#7361) is based on #7352 and requires that to be merged first.
- #7352 is now closed, however #7533 "fully contains this branch and is kept up to date".
- A comment #7533 states "once #7361 is merged, I'll add support", in reference to Texture2DArray. 
- Currently #7533 already runs side-by-side with the existing API & build pipeline. Assets can be built using the MGCB editor to use either MojoShader or ShaderConductor, and at runtime follow the existing or the compute path accordingly.
- A while back the #7533 milestone was changed from 3.9 to 5.0. This implies #7533 which works with the API will require breaking API changes in order to work with the API.

So in short, this PR is waiting for another PR to be merged. That PR is compatible with the API but needs to be incompatible with the API so that breaking API changes can be made to make it API compatible. That PR is waiting for this PR to be merged first, but once this PR and that PR are merged, this PR can finally be merged.

Hope that clears everything up.

SimonDarksideJ:
@minimalism are you available to rebase and resolve the merge conflicts?

Only concern is that @squarebananas states this is dependent on #7352 which was withdrawn, so is this still valid or can it work without it?

cpt-max:
@minimalism explained why this is based on the ShaderConductor branch:
> There are no overlapping edits with the ShaderConductor branch so the relevant changes ( https://github.com/MonoGame/MonoGame/commit/fdfc4d98c46b2ba7c95b8e63c74b2fa7008853a1 ) can be cherrypicked, but attempting to compile such a shader with MojoShader gives the error "Unsupported texture type for ps_3_0" and raising the shader model to 4_0 gives "Shader must be SM 3.0 or lower!"

ThomasFOG:
Blocking this for the time being as it requires a wider range of shader profile support.

PR #7362, opened 2020-09-27
File drop event on DesktopGL and WindowsDX platforms

Quant1um:
I've added support for handling user file drop.
This PR adds `FileDrop` event to `GameWindow` that fires when user drops one or multiple files into a game window. 
MonoGame already has such events as `TextInput`, `KeyUp` and `KeyDown` that are not part of the original XNA API so there is no reason not to have this event implemented.


MonoGame.Framework/GameWindow.cs
```diff
@@ -155,6 +155,19 @@ protected GameWindow()
 
 #endif
 
+#if WINDOWS || DESKTOPGL || ANGLE
```
harry-cpp, line 158: Don't hide the event behind a platform define.

MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs
```diff
@@ -49,6 +53,11 @@ public SdlGamePlatform(Game game)
             if (version >= 205 && CurrentPlatform.OS == OS.Windows && Debugger.IsAttached)
                 Sdl.SetHint("SDL_WINDOWS_DISABLE_THREAD_NAMING", "1");
 
+            // https://wiki.libsdl.org/SDL_DropEvent
+            // SDL_DROPTEXT, SDL_DROPBEGIN, and SDL_DROPCOMPLETE are available since SDL 2.0.5. 
+            _multidropSupported = version >= 205;
```
harry-cpp, line 58: You don't need to check this, we already request that SDL version is 2.0.5 or higher few lines above.

harry-cpp (changes requested):

MonoGame.Framework/GameWindow.cs
```diff
@@ -242,7 +252,14 @@ internal void OnKeyUp(InputKeyEventArgs e)
 	    }
 #endif
 
-		protected internal abstract void SetSupportedOrientations (DisplayOrientation orientations);
+#if WINDOWS || DESKTOPGL || ANGLE
```
harry-cpp, line 255: There shouldn't be an #if define here.

MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs
```diff
@@ -233,6 +238,46 @@ private void SdlRunLoop()
...
                         break;
+
+                    case Sdl.EventType.DropFile:
+                        if (ev.Drop.WindowId != _view.Id)
+                            break;
+
+                        if (!_view.IsFileDropHandled)
+                            break;
+
+                        unsafe
```
harry-cpp, line 249: You've got https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Utilities/InteropHelpers.cs#L16 to help you with converting IntPtr to string.

harry-cpp (changes requested):
Sorry for the very late reply, this totally slipped my attention.

harry-cpp:
Build bot failed to build the PR, merge upstream/develop into your branch, it should fix it.

MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs
```diff
@@ -233,6 +238,37 @@ private void SdlRunLoop()
...
+                        }
+
+                        break;
+
+                    case Sdl.EventType.DropComplete:
+                        if (ev.Drop.WindowId != _view.Id)
+                            break;
+
+                        if (!_view.IsFileDropHandled)
+                            break;
```
TechnologicalPizza, line 263:
```suggestion
                            {
                                _dropList.Clear();
                                break;
                            }
```
This should also clear the drop list so you won't get stale drops.

TechnologicalPizza (changes requested):

MonoGame.Framework/GameWindow.cs
```diff
@@ -155,6 +155,16 @@ protected GameWindow()
...
 
+        /// <summary>
+        /// This event is raised when user drops a file into the game window
+        /// </summary>
+        /// <remarks>
+        /// This event is only supported on desktop platforms.
+        /// </remarks>
+        public event EventHandler<FileDropEventArgs> FileDrop;
+
+        internal bool IsFileDropHandled { get { return FileDrop != null; } }
```
harry-cpp, line 166: There is no reason for this helper property.

MonoGame.Framework/Platform/SDL/SDL2.cs
```diff
@@ -1106,4 +1111,20 @@ public static void UpdateEffect(IntPtr haptic, int effect, ref Effect data)
...
     }
+
+    public static class Drop
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        public unsafe struct Event
+        {
+            public EventType Type;
+            public uint TimeStamp;
+            public char* File;
```
harry-cpp, line 1122: You can use IntPtr here to avoid unsafe, the remaining code already uses it so we should probably stick with it for now.

MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs
```diff
@@ -233,6 +238,37 @@ private void SdlRunLoop()
                                 break;
                         }
                         break;
+
+                    case Sdl.EventType.DropFile:
+                        if (ev.Drop.WindowId != _view.Id)
+                            break;
+
+                        if (!_view.IsFileDropHandled)
```
harry-cpp, line 246: DropFile event isn't something that is gonna happen every input action or something like that, it shouldn't be necessary to protect against it if nothing is hooked up, its only gonna create problems if you try to add/remove stuff from it.

MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs
```diff
@@ -233,6 +238,37 @@ private void SdlRunLoop()
...
+                    case Sdl.EventType.DropComplete:
+                        if (ev.Drop.WindowId != _view.Id)
+                            break;
+
+                        if (!_view.IsFileDropHandled)
+                            break;
+
+                        if (_dropList.Count > 0)
+                        {
+                            _view.OnFileDrop(new FileDropEventArgs(_dropList.ToArray()));
```
harry-cpp, line 267: OnFileDrop will protect you against null check for the event, just pointing it out so you don't add a null check for the event.

harry-cpp (changes requested):

MonoGame.Framework/Platform/SDL/SDL2.cs
TechnologicalPizza: `char*` is actually confusing as the underlying data is UTF8. `IntPtr` and `byte*` are the correct types.

harry-cpp (changes requested):

harry-cpp (approved):

harry-cpp:
Everything looks great! Huge thanks @Quant1um and sorry for taking so long to get around to reviewing this again!

PR #7366, opened 2020-09-30
[Pipeline Tool] Output the location MGCB looks for a font in the error message

kimimaru4000:
This PR outputs the location MGCB looked for a font in the error message if it can't be found. This helps diagnose why it can't find a font; this is especially useful across different platforms.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/FontDescriptionProcessor.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
             }
 
             if (!File.Exists(fontFile))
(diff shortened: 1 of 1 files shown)


tdeeb:
Just to note: I have not been able to actually test this PR because I can't seem to get a custom build of the pipeline tool running. It outputs "mgcb is not in the search path" when running the custom build, and the documentation doesn't cover this.

SimonDarksideJ:
If you open the "MonoGame.Tools.xxx" sln, you can create and run the Content Builder tool and set your "startup" project to the "Editor" for your platform.  Then you can build and test.

Make sure to resolve the git conflict issues by updating from base first @kimimaru4000

SimonDarksideJ:
Migrated fix to https://github.com/MonoGame/MonoGame/pull/8844

Thanks for the contribution @kimimaru4000

PR #7367, opened 2020-10-01
Updated Matrix.CreatePerspective* to support infinite far planes.

vpenades:
As many of you are aware, `System.Numerics.Vectors` was originally designed using XNA vectors and matrices as a template, this means there's a great similarity between both APIs.

Recently, I've noticed that Microsoft updated the `System.Numerics.Vectors` `Matrix4x4.CreatePerspective` and `Matrix4x4.CreatePerspectiveFieldOfView` so they can handle positive infinity far planes.

So it felt natural to keep these methods functionality as close as possible.

The System.Numerics.Vectors code, with the new infinity functionality can be found here:

- [CreatePerspective](https://github.com/dotnet/runtime/blob/e64bc548c609455652fcd4107f1f4a2ac3084ff3/src/libraries/System.Private.CoreLib/src/System/Numerics/Matrix4x4.cs#L824)
- [CreatePerspectiveFieldOfView](https://github.com/dotnet/runtime/blob/e64bc548c609455652fcd4107f1f4a2ac3084ff3/src/libraries/System.Private.CoreLib/src/System/Numerics/Matrix4x4.cs#L860)

The funcionality of these methods has not changed for usual ranges, but now it's possible to pass `float.PositiveInfinity` as the far plane.

### DIFF
--- a/MonoGame.Framework/Matrix.cs
+++ b/MonoGame.Framework/Matrix.cs
@@ -914,7 +914,7 @@ namespace Microsoft.Xna.Framework
         /// <param name="width">Width of the viewing volume.</param>
         /// <param name="height">Height of the viewing volume.</param>
         /// <param name="nearPlaneDistance">Distance to the near plane.</param>
-        /// <param name="farPlaneDistance">Distance to the far plane.</param>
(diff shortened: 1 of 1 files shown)


Jjagg:
Thanks, this is great!

Issue #7363, opened 2020-09-28
Create a version of Texture2D.FromStream() which does not set RGB = 0 when alpha = 0


YTN0:
I was using Texture2D.FromStream() to load some textures for a terrain system I am working on. It looks like this method is doing some post loading processing by setting the RGB values to 0 where the alpha is 0. This will of course break any splatmaps which relies on these values being maintained.

It would be useful to have a version of this API which does not modify the original pixel data i.e. does not set the RGB values to 0 if alpha is 0. 

This could be just overloading the existing method with a flag, or a new method altogether.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- WindowsDX

rds1983:
Well, this PR might fix that: https://github.com/MonoGame/MonoGame/pull/6206
However it wasnt approved yet.

PR #7369, opened 2020-10-03
Added pixels processing function parameter to the Texture2D.FromStream/TextureFromFile

rds1983:
This PR adds parameter of type Action<byte[]> to Texture2D.FromStream/FromFile, which processes data in RGBA format before the data is sent to the video memory.
Also it adds two default color processors: ZeroTransparentPixels(used by default) and PremultiplyAlpha.
With this PR it would be possible to address #7363 by following code:
```c#
  var texture = Texture2D.FromStream(device, stream, null);
```
Or it would be possible to premultiply alpha:
```c#
  var texture = Texture2D.FromStream(device, stream, DefaultColorProcessors.PremultiplyAlpha);
```
Finally it would possible to write custom processor:
```c#
  // Sets every other pixel to red
  var flag = false;
  var texture = Texture2D.FromStream(gd, stream, data =>
            {
                for(var i = 0; i < data.Length; i += 4)
                {
                    if (flag)
                    {
                        data[i + 0] = 255;
                        data[i + 1] = 0;
                        data[i + 2] = 0;
                        data[i + 3] = 255;
                    }

                    flag = !flag;
                }
            });
```


SimonDarksideJ:
Bump @tomspilman / @Jjagg , candidate for 3.8.1?

bartwe:
Calling a delegate function per pixel seems like it would be pretty slow compared to passing the buffer as a whole

rds1983:
Makes sense, @bartwe. Fixed.

harry-cpp:
> Bump @tomspilman / @Jjagg , candidate for 3.8.1?

Only fixes should go into 3.8.1, no new features.

RUSshy:
I don't think such code has place on such a library

- it bloats the library
- Action is slow

SimonDarksideJ:
Ok, Can this be merged after the 3.8.1 so it is available in dev.  Maybe now that we have moved to full nuget packages, we could look to publishing "preview" packages from the dev branch.

vpenades:
I was wondering... as far as I know, "FromStream" does not generate mipmaps, right?

If that's the case, I would consider how the arguments would affect having a FromStream(stream stream, bool generateMipmaps) or something like that.

mrhelmut:
This is really interesting, we should make it move forward for 3.8.1.

harry-cpp (approved):

Jjagg (approved):

PR #7370, opened 2020-10-06
Docs/content pipeline update

SimonDarksideJ:
Added a new article to highlight the benefits of the content pipeline following a discussion.
Also cleaned up the TOC to remove some confusion

### DIFF
--- a/Documentation/articles/content/why_content_pipeline.md
+++ b/Documentation/articles/content/why_content_pipeline.md
@@ -1,0 +1,59 @@
+# Why use the Content Pipeline
+
+The MonoGame team has been putting a lot of effort into a cross-platform content pipeline, but given that for the most part we support loading native assets like .png, .mp3, .wav why bother? Well, it all boils down to a couple of words.. performance, efficiency. Let us look at an example, graphics are some of the biggest assets games use, they are also a major resource hog. Textures will probably take up most of the room in your deployment and will be taking up most of the memory on your device as well.
+
+## Textures
+
+Formats like PNG's are generally not optimized for games.
+
+For example, most GPUs have specific hardware supported compressed formats for textures. PNGs or JPEGs just get uncompressed when passed to the GPU. By using the hardware compression you often get 4x to 8x more textures space for your game as well as faster load times and smaller packages.
+
+If we have a 256×256 32 bit .png texture that we are using in our game, we don’t want to bother with all this compiling to .xnb rubbish that people do, we just want to use the texture as a raw .png file. On disk .png is very impressive in its size, that image probably only takes up 2-5 kb on disk, keeping your application package size down. Great!
+
+Now, what happens when we load this .png from storage on a device (like an iPhone). Firstly it is loaded from storage into memory and decompressed/unpacked from its compressed png format into raw bytes. This is done because the GPU on your device doesn’t know how to use a png image directly, it can only use certain types of compression. So we unpacked the image into memory, this is 262,144 bytes (256x256x4, the x4 is because we have 1 byte per channel Red, Green, Blue, and Alpha). 
+
+> Note that 262 KB  is quite a bit bigger than the compressed size. 
+
+The next thing to do is create a texture for that data because your device cannot compress on the fly (yet), it has to use that data as is. So in creating the texture we used 262kb  of graphics memory on the GPU. That doesn’t sound too bad, but if you are using larger textures say 1024×1024 then you are using 4 MB of GPU memory for that one texture. Multiply that over the number of textures in your game and you soon run out of texture memory on the GPU. Then the GPU has to swap that data out into system memory (if it supports that) or throw an error when you try to create textures that won’t fit into available memory. 
+
+So to sum up:
+
+> using **.pngs** = smaller package size & higher memory usage & less textures
+
+Now let us look at a texture that has been pre-processed using the content pipeline, because we know we are targeting iOS we know the GPU’s on those devices support using PVRTC texture compression directly. So we take our sample .png and compress it at build time using PVRTC and we end up with is a 32kb file (size depends on the texture, alpha channel, etc). Hmm, that is a lot bigger than the .png on disk, but that is not the whole story. The difference here is that there is no need to unpack/decompress it which saves on load time, also, we can create a texture from that data directly so we only use 32kb of texture memory on the GPU and not 262kb. That is a massive saving.
+
+Summing up:
+
+> **compressed textures** = larger package size (maybe) & lower memory usage & more textures
+
+Now we only looked at iOS, but the same applies to desktop environments, most desktop GPUs support DXT texture compression, so the content pipeline will produce DXT compressed textures which can be loaded and used directly. The only platform that is a pain is Android because the Android platform does not have consistent support for compressed textures at the moment and MonoGame has to decompress DXT textures on the device and use it directly. However, even Android will get compressed texture support eventually. There is also a piece of work happening where the pipeline tool will automatically pick a texture format to use, so for opaque textures, it will use ETC1 (which is supported on all android devices but doesn’t support alpha channels) but for textures with an alpha channel, it will use RGBA4444 (dithered) but also allow the user to pick from a wide variety of compression options manually such as PVRTC, ATITC, DXT/S3TC, ETC1, and RGBA4444. This will give the developer the choice of what to use/support.
+
+## Audio
+
+Now let us look at audio. All the different platforms support different audio formats, if you are handling this yourself you will need to manually convert all your files and include the right ones for each platform. A better option would be to keep one source file (be it .mp3, .wmv, etc) and then convert it to a supported format for the target platform at build time? Ok, it does make for longer build times, but at least we know the music will work. MonoGame uses ffmpeg to do the heavy lifting when converting between formats as it can pretty much convert any type to any other type which is really cool.
+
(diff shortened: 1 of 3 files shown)


Documentation/articles/content/why_content_pipeline.md
```diff
@@ -0,0 +1,59 @@
...
+
+> Summing up:
+>
+> **compressed textures** = larger package size (maybe) & lower memory usage & more textures
+
+Now we only looked at iOS, but the same applies to desktop environments, most desktop GPUs support DXT texture compression, so the content pipeline will produce DXT compressed textures which can be loaded and used directly. The only platform that is a pain is Android because the Android platform does not have consistent support for compressed textures at the moment and MonoGame has to decompress DXT textures on the device and use it directly. However, even Android will get compressed texture support eventually. There is also a piece of work happening where the pipeline tool will automatically pick a texture format to use, so for opaque textures, it will use ETC1 (which is supported on all android devices but doesn’t support alpha channels) but for textures with an alpha channel, it will use RGBA4444 (dithered) but also allow the user to pick from a wide variety of compression options manually such as PVRTC, ATITC, DXT/S3TC, ETC1, and RGBA4444. This will give the developer the choice of what to use/support.
```
tomspilman, line 29: > There is also a piece of work happening where the pipeline tool will automatically pick a texture format to use,

I am pretty certain this already works.  Use the `Compressed` setting which i think is the default now?

Documentation/articles/content/why_content_pipeline.md
```diff
@@ -0,0 +1,59 @@
...
+
+Most platforms have audio processing that is optimized to certain compressed formats. By not using them you loose performance and system memory.  Often this is when people resort to just using a format optimized for your current target platform. You could save the sound effects as ADPCM and run pretty optimally on Windows systems. The problem here is only apparent once you try to take your game to another platform where ACPCM is not optimal. You need to manually re-export all your game content in the new optimal format for that platform… and you did keep your original content uncompressed right?
+
+## Shaders
+
+This is an area that causes real pain, custom shaders. There are a number of shading languages that you can use depending on the platform you are targeting. For OpenGL based systems that is GLSL, for DirectX based systems it is HLSL, there is also CG from Nvidia.
```
tomspilman, line 39: > This is an area that causes real pain, custom shaders.

I am concerned this doc is being too informal with the reader.  They are not looking for jokes or fun... just information.

It is also not pleasant to just load the user down with "this sucks", "that sucks", "this is difficult", etc,  For example... i don't find shaders "a real pain".  They work fine.

Documentation/articles/content/why_content_pipeline.md
```diff
@@ -0,0 +1,59 @@
...
+
+> From MonoGame 3.8, the MGFX tool can build content on any platform. For MonoGame 3.7 or earlier, MGFX only runs on a windows box, this is because it needs the DirectX shader tooling to compile the HLSL before passing it to libmojoshader for conversion (for OpenGL platform targets).
+
+## Models
+
+For the most part, the model support in XNA/MonoGame is pretty good. XNA supports .x, .fbx files for 3D models.  MonoGame, thanks to the excellent assimp project, supports a much wider range of models including .3ds. However, some of these formats might produce some weirdness at render time as only .fbx has been fully tested. Also note that assimp does not support the older version .fbx format which ships with most of the original XNA samples, so you will need to [convert those to the new format manually](https://www.codeproject.com/articles/1041397/updating-old-fbx-files-for-the-modern-era).
```
tomspilman, line 47: > For the most part, the model support in XNA/MonoGame is pretty good.

This is all sounding too much like a personal blog post and less like documentation.

The OpenGL docs don't talk about "well threading sort of works".  I would remove personal anecdotes from these docs and focus on the feature set.

Documentation/articles/content/why_content_pipeline.md
SimonDarksideJ: :D, blame @dellis1972 , likely that was from his article :D 

But I'll do another pass and make it a bit more formal.

Documentation/articles/content/why_content_pipeline.md
SimonDarksideJ: It was a blog, as it was @dellis1972 , although it did have a fair bit of revision already.

SimonDarksideJ:
Content updated to a more formal language

SimonDarksideJ:
Any further comments @tomspilman or is this good to get in ?

Documentation/articles/content/why_content_pipeline.md
```diff
@@ -0,0 +1,69 @@
+# Why use the Content Pipeline
+
+The MonoGame team has been putting a lot of effort into a cross-platform content pipeline, but would you use the **Content Pipeline** when MonoGame also supports loading assets natively, like .png, .mp3, .wav? Well, it all boils down to a couple of words, performance and efficiency.
```
tomspilman, line 3: >, but **why*8 would you use

tomspilman: Not sure that Content Pipeline needs to be bolded here.

Documentation/articles/content/why_content_pipeline.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Textures
+
+Most texture formats (like PNGs) are generally not optimized for games.
```
tomspilman, line 7: > Most image formats

PNGs are not a texture formats.

Documentation/articles/content/why_content_pipeline.md
```diff
@@ -0,0 +1,69 @@
...
+
+For example, most GPUs have specific hardware supported compressed formats for textures. PNGs or JPEGs just get uncompressed when passed to the GPU. By using the hardware compression you often get 4x to 8x more textures space for your game as well as faster load times and smaller packages.
+
+If we have a 256×256 32 bit .png texture that we are using in our game, why should a developer bother with compiling to .xnb's, they just want to use the texture as a raw .png file. On disk .png is very impressive in its size, that image probably only takes up 2-5 kb on disk, keeping your application package size down. Great!
```
tomspilman, line 11: > If we have a 256×256 32 bit .png texture that we are using in our game, why should a developer bother with compiling to .xnb's, they just want to use the texture as a raw .png file.

This sentence makes no sense.  I would cut everything after the first comma... the rest is unnecessary info.

Documentation/articles/content/why_content_pipeline.md
```diff
@@ -0,0 +1,69 @@
...
+
+However, what happens when we load this .png from storage onto a device at runtime (like an iPhone).  The texture is loaded from storage into memory and decompressed/unpacked from its compressed png format into raw bytes. This is done because the GPU on your device does not know how to use a png image directly, it can only use certain types of compression. With the image loaded into memory, this takes up 262,144 bytes (256x256x4, the x4 is because we have 1 byte per channel Red, Green, Blue, and Alpha).
```
tomspilman, line 13: > (like an iPhone)

PNGs are decompressed on all platforms... no need to give an example of a platform.

tomspilman: >  This is done because the GPU on your device does not know how to use a png image directly,

Didn't we already say this in the first paragraph of this section?

tomspilman:
@SimonDarksideJ added a few more comments.  I haven't gone thru all of it yet, but will do more later.

SimonDarksideJ:
OK, but might be easier of you just use the GitHub ability to make changes to the PR directly @tomspilman in the web interface, which should cut down on the back and to.

Documentation/articles/content/why_content_pipeline.md
SimonDarksideJ: For Example, [redacted]
:D

Documentation/articles/content/why_content_pipeline.md
SimonDarksideJ: resolved points above

Documentation/articles/content/why_content_pipeline.md
SimonDarksideJ: Updated section and removed hyperbole

SimonDarksideJ:
Any further updates / corrects @tomspilman  so we can clear this PR?

Any updates @dellis1972 (especially as a lot of the content came from your blog)

dellis1972:
This looks ok. The only change I would make is to create another page to describe the steps of converting old fbx files rather than rely on external articles. But that can be done later I suppose.

dellis1972 (approved):

SimonDarksideJ:
Oi, stop stealing my traffic @dellis1972 , that's one of my most viewed pages on codeproject and my blog :P

Hedmun395:
The information about the PNGs is absolutely misleading for pixel art projects (which are probably 90% of the uses of MonoGame, if not more)

To start with, pixel art must go into the GPU uncompressed, otherwise artifacts appears and colors change.
So, the DXT compression is not possible and there's no memory savings at all in regards of GPU memory.

Next, the size of an uncompressed color image compared to PNG is huge. Depending on pixel art style and density of the atlas, it can go from x10 to x100, even more.

In order to reduce this size, MonoGame uses LZ compression (or something similar, I can't remember). I agree that PNG is not a format optimized for videogames, but neither is LZ. And at least PNG uses redundancy of previous/next lines, which offers you a better compression. Big win for PNGs in terms of disk space, which is crucial in mobile platforms.

The disadvantage of "an intermediate buffer is required" disappears also, because the LZ decompressor requires exactly the same space as a PNG. There's no advantages in CPU memory either after all, unless you want a 20KB png convert into a 2MB file.


So for pixel art, the Content Pipeline takes the same GPU memory, more CPU memory (because the compressed source file is bigger), much more disk memory, and is slightly faster because the PNG decompressor MonoGame uses is incredibly slow compared with the native version you were using a few years ago.

SimonDarksideJ:
Great feedback @Hedmun395 .

How would you recommend updating the article to point that out, bear in mind that a lot of users are not technically savvy , so the advantages / disadvantages need to be clearly understood without technical bias

PR #7376, opened 2020-10-14
Updated Existing "Getting Started" page to be the new "Downloads" page

SimonDarksideJ:
Updated Existing "Getting Started" page to be the new "Downloads" link target.

Recommend the Link on the main website be renamed to "Getting Started" instead of Downloads.

### DIFF
--- a/Documentation/articles/getting_started/0_getting_started.md
+++ b/Documentation/articles/getting_started/0_getting_started.md
@@ -1,5 +1,7 @@
 # Getting Started
 
+This section walks you through the basics of MonoGame and helps you create your first game.  We have also included some additional information related to the MonoGame project for developers.
+
 To get started, first, select the toolset and operating system you will be working with to create your first MonoGame project, then continue reading to understand the basic layout of a MonoGame project.
 
 By the end of this tutorial set, you will have a working project to build off of for your target platform and will be ready to tackle your next steps.
--- a/Documentation/articles/getting_started/0_getting_started.md
+++ b/Documentation/articles/getting_started/0_getting_started.md
@@ -1,8 +1,6 @@
 # Getting Started
 
-This section walks you through the basics of MonoGame and helps you create your first game.
-
-First, select the toolset and operating system you will be working with to create your first MonoGame project, then continue reading to understand the basic layout of a MonoGame project.
+To get started, first, select the toolset and operating system you will be working with to create your first MonoGame project, then continue reading to understand the basic layout of a MonoGame project.
 
 By the end of this tutorial set, you will have a working project to build off of for your target platform and will be ready to tackle your next steps.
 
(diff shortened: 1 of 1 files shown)


Documentation/articles/getting_started/0_getting_started.md
```diff
@@ -1,8 +1,6 @@
 # Getting Started
 
-This section walks you through the basics of MonoGame and helps you create your first game.
```
tomspilman, line 3: I hate loosing this line...  it is perfect to explain what this is for.

Documentation/articles/getting_started/0_getting_started.md
```diff
@@ -23,3 +21,23 @@ By the end of this tutorial set, you will have a working project to build off of
 - [Understanding the Code](3_understanding_the_code.md)
 - [Adding Content](4_adding_content.md)
 - [Adding Basic Code](5_adding_basic_code.md)
+
```
tomspilman, line 24: @SimonDarksideJ @harry-cpp @Jjagg 

I think this just confuses things.  

We have a very simple clear document talking about getting started with MG.

I don't see the benefit to distract the user by listing all this other stuff.  A new user... one that needs this guide... doesn't care about any of this stuff.

Documentation/articles/getting_started/0_getting_started.md
SimonDarksideJ: Ok, will revert

Documentation/articles/getting_started/0_getting_started.md
SimonDarksideJ: Except the "Downloads" page doesn't really server any purpose, so the suggestion was to merge in any additional information in to inform the user.

Thy can choose to follow or not, but at least they are informed and all the information is in one location instead of posted in several different locations.

mrhelmut: I agree that things seem quite clear as they are. 
I would however see value in a separate advanced topic page to detail how to use MonoGame from source and development nugets (which could be linked from the download page; I find the info on the download page to be a bit lacking on those topics, but I would also agree that this page needs to be small and straightforward, hence the suggestion to separate these in advanced topics).

SimonDarksideJ: In which case then @mrhelmut would it be better to update the "Downloads" page with better links to the "Getting Started" page, since installing the Setup installer is only needed on Windows.

Maybe a grid for each Operating system, something like this

image

tomspilman:
This is conflicting @SimonDarksideJ .  Should we close this?

SimonDarksideJ:
Withdrawn, to be addressed in the new MonoGame docs site

PR #7384, opened 2020-10-27
Generic Threading.BlockOnUIThread for reducing garbage allocs

### LINKED ISSUE #6884: OpenGL implementation of VertexBuffer generates garbage
Hi, I've been trying to figure out what generates garbage in VertexBuffer.SetData and I believe I have nailed it. In OpenGL implementation most of the methods are wrapped in a `Threading.BlockOnUIThread` with a lambda. The thing is that this lambda refences local variables, like here:
```c#
private void PlatformSetDataInternal<T>(int offsetInBytes, T[] data, int startIndex, int elementCount, int vertexStride, SetDataOptions options, int bufferSize, int elementSizeInBytes) where T : struct
        {
            Threading.BlockOnUIThread(() => SetBufferData(bufferSize, elementSizeInBytes, offsetInBytes, data, startIndex, elementCount, vertexStride, options));
        }
```
Even though the lambda it self may be referenced as a pointer to code, local variables referenced in this lambda have to be stored somewhere, and internally memory is allocated for them on heap (https://stackoverflow.com/a/7133064). I've tried to get rid of allocations in this case, but I believe it might be impossible. Actions are later stored in a List, so that they can be run on a main thread. Since there are different lambdas with different local variable counts and types referenced in each, it's not possible to hack it in a way that these arguments are stored in a List of some value type.

Right now I see the following possible solutions:
1) Create a class for each lambda type used in Threading.BlockOnUIThread, and keep them in an object pool.
2) Make user responsible for calling all these methods on ui thread, and just throw an exception if these methods are not called on ui thread (by a preprocessor directive or exposing additional methods)
3) Ignore this issue

- dc4b541e3 commit

- Windows
- Linux

- DesktopGL

TechnologicalPizza:
Fixes #6884

This PR is not fully implemented as I want it to be accepted first.
Initialization functions, content readers, and some device-gets don't use the new state overload yet.  
Neither is it necessary to implement these at they are called way less than device-sets e.g. `VertexBuffer.SetData`.

I have not benchmarked this code. It only strives to reduce allocations, but performance was kept in mind.

This moves around some code but I hope that it will be worth it in the long run.


TechnologicalPizza:
There seem to be no tests for threaded calls to various GL methods. I may write some tests for common methods tomorrow.

TechnologicalPizza:
Hmm, I wasn't able to create useful tests. They seem to crash in a thread-race as multiple tests call the static `Threading.Run`.

mrhelmut:
Hey @TechnologicalPizza,

I think that this is very interesting, especially in regard to reducing allocations on vertex and index buffers. Is there anything that this PR might still require or would you consider it fine enough for review?

TechnologicalPizza:
The only thing in this PR as far as I remember is to just use it more throughout the code base.
I only implemented this new way for hot paths for it to be reviewed.

mrhelmut:
I'll test it out then. Less garbage is always a great thing for per-frame operations.

mrhelmut:
I gave it a test spin, and as far as allocations go, it's really great! Pretty cool to see vertex/indice buffers working garbage-free. The threading queue is a welcome rework.

I wonder if the new ```try```/```finally``` blocks are a good thing or not. Shouldn't we let things crash in those cases? It might end up hiding issues.

@Jjagg @harry-cpp for another pair of eyes.

TechnologicalPizza:
Are you talking about the try/finally around certain drawing functions that I added? They don't hide any exceptions and GCHandle should always be released after use.

mrhelmut:
You're right, and I guess AL/GL CheckError would do the trick anyway.

This PR is good to go to me. Waiting for a second review.

mrhelmut (approved):

TechnologicalPizza:
Actually I was wondering about why `PlatformGetData()` in `Texture2D.OpenGL.cs` and `TextureCube.OpenGL.cs` use `Threading.EnsureUIThread();`  instead of `Threading.BlockOnUIThread();`.
Could it be to protect the developer from synchronization issues? 
Using `Threading.BlockOnUIThread()` outside the main thread means that `GetData()` will be called at the end of a frame, which may be undesirable.

mrhelmut:
The build bot failed but it is due to your changes, it's an issue on the bot that we're trying to figure out (it fails infrequently).

OpenGL is totally not friendly with threading and postponing any threaded operation to the end of a frame was the safest route to cover most cases (and for basic compatibility with other targets which are more resilient to threading). This indeed comes with limitations, but what was assumed was that timing dependent operations are not threaded and therefore should execute immediately.

TechnologicalPizza:
I am so confused, the build bot failed because it couldn't build the Content editor for Linux as far as I see.
And I saw that the bot failed yesterday and then succeded, but now it's failed again?
I'm 99.9% sure that this is not from my change (I only moved a code block that is under a `#if ANDROID` conditional)

mrhelmut:
It's not about your changes, it fails infrequently recently due to a recent .NET 5 issue. We're looking into it.

mrhelmut:
@TechnologicalPizza Would you mind trying to rebase your PR? This would trigger the bot to restart (we made a change which might have resolved the build issue).

TechnologicalPizza:
Ah yes, I rebased incorrectly. Hopefully that doesn't matter 😅

mrhelmut:
Ah, you were quicker than me, I deleted my message asking for a rebase because the issue is still there. #7462 might fix it though. We'll likely have to to wait for this PR to get merged before rebasing and merging this one.

mrhelmut:
@TechnologicalPizza Would you mind updating your PR so that it merges with the conflicts that just appeared? (a rebase should do it)

TechnologicalPizza:
Ok I messed up again by deleting global.json, lovely

mrhelmut:
It looks like you reverted ```global.json```, ```ThirdParty/Dependencies``` and ```ThirdParty/SDL_GameControllerDB```. If you revert them back, we should be fine.

mrhelmut:
We're clear!
Summoning @harry-cpp and @Jjagg

Jjagg:
Thanks @TechnologicalPizza, this looks great!

Generally we avoid formatting changes because it makes it harder to look through the Git history, but that's not a blocker for me, and the other changes in this PR are very valuable so I'll go ahead and merge this :)

TechnologicalPizza:
Nice. Hopefully no one complains that their Texture2D constructor creates one object of garbage 😄  
(Only hot paths like GetData and SetData on textures and buffers have allocations mitigated)

MonoGame.Framework/Platform/Threading.cs
```diff
@@ -71,41 +102,84 @@ internal static void BlockOnUIThread(Action action)
...
+            lock (_queuedActions)
+            {
+                StateActionHelper<TState>.Queue.Enqueue(queuedAction);
+                _queuedActions.Add(StateActionHelper<TState>.DequeueAction);
+            }
+
+            try
+            {
+                if (!resetEvent.Wait(MaxWaitForUIThread))
+                    throw new TimeoutException();
```
roeyskoe, line 146: I just updated MonoGame to latest develop and now I am having problems with this check, since the MaxWaitForUIThread time is too short in some cases.

Is there a specific reason for Threading class being internal and MaxWaitForUIThread a const, or would it be ok to make this class public and remove the const modifier so that the value could be changed?

Maybe just change the field static?

I can make a pull request if that change would be ok.

mrhelmut: Could you elaborate on the use case?

roeyskoe: Because of that check, some previously working code will crash, if for example in my case the update loop takes too long to finish during game startup and object creation.

I fixed my crashing by just increasing MaxWaitForUIThread to 10 seconds, and started thinking that maybe it would be good idea to allow users change the value however they wish to. Of course just by increasing the value would also solve the problem, but I see no reason why this value should be locked to something specific.

mrhelmut: Why would your startup take this much time? If you don't mind, I would like more context than just saying that it doesn't work for you because we might be missing something.

Usually, if doubling a timing value fixes an issue, it's not the right fix and it is likely that there's more to it.

roeyskoe: Well, several hours of debugging later I found out that I (maybe?) had older dependencies on my output folder, and those were causing the problems.
Redownloading  solved  my problems.

Sometimes rm -rf is actually a solution :D

Sorry about that...

mrhelmut: @roeyskoe I don't think that it was an issue from your end, I pushed an updated monogame to our testers and this crashes _a lot_. I believe that both the max wait, and the max thread count are an issue.

@TechnologicalPizza Do you have a suggestion to get rid of ```MaxWaitForUIThread``` and ```MaxPooledResetEvents``` entirely ? These can very frequently fail on simple texture loadings and I don't think that increasing them would fix anything.

I'm going to open up an issue and look into this today.

roeyskoe: Yeah, I just found out that the problem was actually garbage collector being forced to run on UI thread. 

I had one RenderTarget2D being turned into garbage and not disposed correctly. Then while creating a bunch of objects automatic GC starts, and since the Dispose function of GraphicsResource calls BlockOnUIThread, it may time out.

On my case I got it fixed it by just adding a call to _target.Dispose(), but it is not really ideal if you forget to do that.

mrhelmut: For the record #7471

Issue #7388, opened 2020-10-30
3D Audio Direction Issues


squarebananas:
The stereo mix for 3D audio seems incorrect in Monogame for both DesktopGL and WindowsDX platforms.

For testing this I have updated the XNA Audio3D sample to run on Monogame 3.8 which can be found here:
https://github.com/squarebananas/Audio3DMonogame

The audio pans between left/right correctly when facing north. However turning the camera to a west/east direction (with left/right on the keyboard) results in reversed audio.

I will create a pull request for the WindowsDX fix as that seems fairly simple.

The DesktopGL code on the otherhand already seems correct but the same issue exists. One solution that seems to work is to negate "finalPos.Z" on SoundEffectInstance.OpenAL.cs line 80, but I don't understand why this works as OpenAL supposedly already uses a right hand coordinate system. Maybe someone else can look at this.

If anyone has a surround sound setup it might be useful to check the front/back mix is also working. Also there are two additions to the original sample. Pressing A/D will roll the camera (for testing the listener in a non-up direction) and pressing F1/F2 will toggle a test tone with a faster cat for doppler effect testing (move the camera further away to hear if the pitch shift is correct).

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:
- DesktopGL, WindowsDX

squarebananas:
Since yesterday I've done more testing and identified some more issues as follows:

1) The OpenAL direction issue is because the position and velocity are projected into listener space incorrectly. Transposing the orientation matrix will fix this.

2) The doppler effect is incorrect on OpenAL platforms as the listener velocity is not used in any calculations at all. With the DesktopGL test project above pressing F1 to start the tone test and then pressing F3 to run the relative velocity test proves this (when moving the camera forwards the relative velocity of the cat is zero, but the pitch still changes). Calculating the relative velocity instead of using the emitter velocity will fix this.

3) On X3DAudio platforms stopping a looping sound with immediate set to false doesn't stop the sound effect at the end of the loop. Running the WindowsDX test project above for 20 seconds proves this as around 3 dog sound effects play on top of each other (this doesn't happen with the DesktopGL project). Calling the ExitLoop method within PlatformStop instead for looped sound effects seems to fix this.

Shall I create separate issues and push requests for some of these?

PR #7389, opened 2020-10-30
Fixes 3D Audio Direction Issues On Platforms Using X3DAudio

squarebananas:
This pull request is to fix 3D audio direction issues on platforms using X3DAudio. More details on the issue can be found here: https://github.com/MonoGame/MonoGame/issues/7388

Only the Z component of the emitter/listener forward direction should be negated to swap between coordinate systems.


Jjagg:
Thanks for doing these in separate PRs, makes it very easy to review :)

PR #7390, opened 2020-10-30
Changed uses of Math to MathF where it was used for floats

### LINKED ISSUE #7382: Math to MathF
Suggestion to change calls to Math to MathF. Saves extra casts and is likely more performant.
For example from Matrix.cs:
`float num = (float) Math.Cos((double) angle);`
Could be changed to:
`float num = MathF.Cos(angle);`

Or does Monogame still need to have .Net Framework support? MathF is in Core, Standard and 5.

initram:
Also added a MathF class that forwards to Math in case there is no MathF class provided by the framework.

Fixes #7382

Please let me know if I was too aggressive in replacement of double Math to MathF, there may be some places where there is a loss of precision, but I am not sure of the impact. I have tried to only change to single precision where it was already happening right after the call to Math, or where there was a single operation afterwards.


TechnologicalPizza:
Shouldn't this forwarding class be under the `System` namespace?

mrhelmut:
> Shouldn't this forwarding class be under the `System` namespace?

I agree.
Looking good to me otherwise.

initram:
I have now moved it to the class to the System namespace.

harry-cpp:
> Shouldn't this forwarding class be under the `System` namespace?

We shouldn't be providing classes under System namespace.

TechnologicalPizza:
This one is at least internal so wouldn't break much for users.

initram:
So what should I do then? Should I move it back to the xna namespace or leave it in the system namespace like the type that it is standing in for. As mentioned the class is an internal class, so users of the library should not be affected.

harry-cpp:
Oh, I did not realize that its not public, it doesn't matter then, leave it as is.

mrhelmut (changes requested):

MonoGame.Framework/MathF.cs
```diff
@@ -0,0 +1,71 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
+
+namespace Microsoft.Xna.Framework
+{
+#if (!NETCOREAPP && !NETSTANDARD2_1) || NETCOREAPP1_0 || NETCOREAPP1_1
```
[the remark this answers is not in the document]
initram: dotnet 5 sets both `NETCOREAPP` and `NET5_0` (I just tested this using RC2). I don't expect them to come with anymore net standards as that was the whole point of dotnet 5. The idea is that all platforms will now support dotnet 5, instead of having netstandard as a lowest common denominator / interface / spec.

I also do not really like that preprocessor check, but I could not really come up with a much better version.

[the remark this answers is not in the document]
TechnologicalPizza: https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/#for-the-impatient-tldr third paragraph;
> We won’t be releasing a new version of .NET Standard, but .NET 5 and all future versions will continue to support .NET Standard 2.1 and earlier.

Would be ideal if every MG project moved to NET5 but that is a far-away future, so it's understandable if some projects stay on netstandard20 or lower. These "legacy" projects should probably include `MathF` as a `Compile` item instead as mentioned.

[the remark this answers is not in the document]
initram: To make the if simpler i could also remove the checks for dotnet core 1.0 and 1.1 as no one will use those. They are past end of support from Microsoft. I just added them for good measure, so that it would not break if compiled with it.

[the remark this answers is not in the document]
mrhelmut: Okay, cool then!

mrhelmut (approved):

harry-cpp:
Looks good, thanks @initram

MonoGame.Framework/Audio/Xact/XactHelpers.cs
```diff
@@ -53,7 +53,7 @@ public static float ParseVolumeFromDecibels(byte decibles)
         public static float ParseVolumeFromDecibels(float decibles)
         {
             // Convert from decibles to linear volume.
-            return (float)Math.Pow(10.0, decibles / 20.0);
```
nkast, line 56: Precision lost!
Here 'decibles' (decibels?) were upgraded to double to be divided by 20.0.
And then Power was calculated with double precision.

MonoGame.Framework/Matrix.cs
```diff
@@ -984,12 +984,12 @@ public static void CreatePerspectiveFieldOfView(float fieldOfView, float aspectR
 		    if (nearPlaneDistance >= farPlaneDistance)
 		    {
 		        throw new ArgumentException("nearPlaneDistance >= farPlaneDistance");
-		    }
-
-            var yScale = 1.0f / (float)Math.Tan((double)fieldOfView * 0.5f);
```
nkast, line 989: Precision lost!
Here '0.5f' was upgraded to double, because fieldOfView was explicitly converted to double, and the multiplication was done with double precision.
Then Tan() was calculated with double precision.

MonoGame.Framework/Plane.cs
```diff
@@ -30,8 +30,8 @@ public static float ClassifyPoint(ref Vector3 point, ref Plane plane)
         public static float PerpendicularDistance(ref Vector3 point, ref Plane plane)
         {
             // dist = (ax + by + cz + d) / sqrt(a*a + b*b + c*c)
-            return (float)Math.Abs((plane.Normal.X * point.X + plane.Normal.Y * point.Y + plane.Normal.Z * point.Z)
-                                    / Math.Sqrt(plane.Normal.X * plane.Normal.X + plane.Normal.Y * plane.Normal.Y + plane.Normal.Z * plane.Normal.Z));
```
nkast, line 34: Precision lost!
The division was done with double precision because there was no cast to float in front of Math.Sqrt().

MonoGame.Framework/Quaternion.cs
```diff
@@ -748,10 +748,10 @@ public static Quaternion Slerp(Quaternion quaternion1, Quaternion quaternion2, f
 		    }
 		    else
 		    {
-		        float num5 = (float) Math.Acos((double) num4);
-		        float num6 = (float) (1.0 / Math.Sin((double) num5));
```
nkast, line 752: Precision lost!
The division (reciprocal) was calculated with double precision.

MonoGame.Framework/Quaternion.cs
```diff
@@ -786,10 +786,10 @@ public static void Slerp(ref Quaternion quaternion1, ref Quaternion quaternion2,
...
```
nkast, line 790: Precision lost!
The division (reciprocal) was calculated with double precision.

Jjagg:
It would have been better if this had not changed internal usages except where the result was directly cast. In general doing the cast as late as possible preserves more precision.

nkast:
Indeed, when the old code was in the form (float)Math.OpMethod((float)floatValue) it could be replaced with the MathF method.
The tricky part is that the compiler upgrades an operation and its result to double if one of the operants is double.

harry-cpp:
Honestly almost all Math -> MathF calls would loose a bit of precision, its not just those 4 examples above, but the base classes are all float based, and this is meant for writing games, for speed of operation, not for super precise math, so I like this change.

@tomspilman Penny for your thoughts.

tomspilman:
@harry-cpp @mrhelmut @nkast

Most of it is ok.  The changes around matrix math and quaternions is a little concerning.

I could see that maybe some user may see their game slightly change behavior under the right conditions.  Cases like doing lots of small substep iterations in a mathematical calculation or math ops between large and small scale numbers.

MonoGame has always maintained some internal double precision, so this will produce a change in results.  That said i cannot point to a specific case it will break.

It sort of reminds me when we added LerpPrecise https://github.com/MonoGame/MonoGame/pull/4233.

All this in mind.... it will make things faster for the majority of MG users doing normal things at regular scales and precision levels.  People doing large 3D scenes may hit issues, but depends on what they are doing.

mrhelmut:
I'm also for keeping the double precise for inner calculations.

Note to self: this breaks consoles, I'll have to add the MathF replacement to them.

Issue #7371, opened 2020-10-06
New 3D Model architecture with in built Animation support and PBR


vpenades:
A while ago, I published an example of how to load glTF models at runtime... since then, this project has evolved into much more, so I would like to introduce the new repository:

https://github.com/vpenades/MonoGame.AnimAndPBR

In it's current state, it's quite usable, and it already supports most of glTF and PBR features.

Since a lot has changed in a while, this issue supersedes #6932

The general plan is to have a new set of effects and model classes that can be used, _out of the box_, as replacement of the very outdated XNA's Model and BasicEffect.

As an example of how the new PBR effects look like, this is a glTF model loaded with the library:

![imagen](https://user-images.githubusercontent.com/5433822/95261239-54d7a780-082a-11eb-9b7c-b59c1c4b579a.png)

So what's this issue about?

Although a lot has been done, there's still a lot to do. In particular, a missing key point is that right now, the glTF loader only runs at runtime... _there's no content pipeline processing_.

So, first and foremost, I would like big brains around here to review the library and give feedback on how to improve it and ways to solve the current big issues it has to move it forward.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- All

#### What MonoGame platform are you using:

- All

Jjagg:
I'm really excited for this! MGs Model class is outdated by 15 years and this is a long overdue modernisation of it! (Though we'll keep the old model class of course)

Could you be more specific about what kind of review/help you need? Maybe open issues in your repo and link them in this issue.

For the content pipeline:
Off the top of my head, I don't think we want to use any of the existing content classes that exist for Model in the content pipeline.
Instead we can use data only classes that map directly to the runtime classes for the glTF model. So anything that has native resources will need a data-only class for the content pipeline.

I really would love to see this integrated! Lets make it happen! Wish I had more time to put into this!

vpenades:
First of all, I would like people that might want to be involved, to get familiar with the current basecode, so we can be on the same page. no need to hurry since this is going to be a long trip.

The repository is divided into 3 main parts:

- The PBR shaders
- The new Model classes
- The new importer architecture

Each of these sections has a number of problems of their own.

[...]

PR #7392, opened 2020-11-01
Don't recompile the same shader multiple times.

cpt-max:
As discussed in #7371 this should shorten compilation times whenever the same shader appears in multiple techniques.


tomspilman:
Looks good to me @Jjagg @harry-cpp .

mrhelmut:
Looking fine to me! Merging, thank you @cpt-max !

Issue #7391, opened 2020-10-31
Maximum number of techniques reached in a mgfxo file


vpenades:
So far I've been able to compile and consume shaders with lots of techniques... up to today, I've was able to load MGFXO files with 128 techniques

Today I've pushed to 256 techniques on a MGFXO and when loading the bytecode into an Effect, it threw this exception:

    System.IndexOutOfRangeException: 'Index was outside the bounds of the array.'

    at Microsoft.Xna.Framework.Graphics.Effect.ReadEffect(BinaryReaderEx reader)
    at Microsoft.Xna.Framework.Graphics.Effect..ctor(GraphicsDevice graphicsDevice, Byte[] effectCode, Int32 index, Int32 count)
    at Microsoft.Xna.Framework.Graphics.Effect..ctor(GraphicsDevice graphicsDevice, Byte[] effectCode)
    at Microsoft.Xna.Framework.Graphics.AnimatedEffect..ctor(GraphicsDevice device, Byte[] effectCode) in D:\Github(Owned)(LIBS)\git.MonoGame.AnimAndPBR\src\MonoGame.Framework.Graphics.PBR\Effects\AnimatedEffect.cs:line 19
    at Microsoft.Xna.Framework.Graphics.PBREffect..ctor(GraphicsDevice device, Byte[] effectCode) in D:\Github(Owned)(LIBS)\git.MonoGame.AnimAndPBR\src\MonoGame.Framework.Graphics.PBR\Effects\PBREffect.cs:line 13

So I've guess I've hit some sort of limit in the file format.

The MGFXO that fails to load can be retrieved [__here__](https://github.com/vpenades/MonoGame.AnimAndPBR/blob/master/src/MonoGame.Framework.Graphics.PBR/Effects/Shaders/Resources2/MetallicRoughnessEffect.ogl.mgfxo)

After discussing with @Jjagg it seems there's some limits in-built in the format that need to be pushed up, which would require bumping the version number of the mgfxo format.

... 

Digging into the code:
https://github.com/Jjagg/MonoGame/blob/cb24746589835f0d4e2b7c041b6b0c8659979737/Tools/2MGFX/EffectObject.writer.cs#L49

Is clamping the number of techniques to 255.... but I see plenty of (Byte) casts....  all these should be replaced by encoded7 or shorts at the very least

I would suggest having subclasses for loading different versions for backwards compatibility (if desired)

Like:

internal static class EffectReaderVer9 { ... }
[...]

PR #7397, opened 2020-11-06
Fix for MGFXO file format to increases collection size limits.

vpenades:
From #7391

Additionally, I did some refactoring, like moving the header struct and header reader to within the Effect Reader region... it did not make much sense having all that stuff scattered through the Effect class.

I tried to avoid doing changes to BinaryReaderEx and BinaryWriterEx, but it proved very difficult to support backward compatibility without modifying these classes. I checked that these classes are only used in the effects readers and writers... so to avoid future conflicts I would like to propose renaming them to MgfxBinaryReader and MgfxBinaryWriter

Although it was not part of the issue this pull request fixes, I added the MGFX signature to the tail of the ver 10 file. This is useful because with binary serialization it is damn easy to have a mismatch between the writer and the reader, in most cases it will fail with end of file exceptions, etc, but in a few cases it might be able to partially load a shader without completely filling the Effect content, and without giving errors, which will fail at runtime when the effect is used, but not when it's loaded. This tail ensures that all the bytes of the file have been properly read.

Finally, since the new reader is backwards compatible, I didn't compile the shaders, let me know if you preffer to keep the current ones, or the new ones compiled to version 10.


MonoGame.Framework/Graphics/Effect/Effect.cs
```diff
@@ -228,68 +191,120 @@ internal protected override void GraphicsDeviceResetting()
...
+            public const int MGFXVersion = 10;
+
+            /// <summary>
+            /// The minimum MonoGame Effect file format version supported (inclusive).
+            /// </summary>
+            /// <remarks>
+            /// We should avoid supporting old versions for very long if at all 
+            /// as users should be rebuilding content when packaging their game.
+            /// </remarks>
+            public const int MGFXMinVersion = 9;
```
harry-cpp, line 213: We never support older generated versions of shaders as there is no good reason to.

harry-cpp (changes requested):

MonoGame.Framework/Graphics/Effect/Effect.cs
harry-cpp: Scratch this, Tom wants it.

vpenades: Okey

vpenades: @harry-cpp I've just updated the PR, version 9 compatibility removed. Since the older shaders no longer work I had to regenerate both the core shaders and the ones used in the content effect tests.

I've ran all the tests and everything seems good, you might want to try regeneraring the MGFXO files on your own for extra safety.

MonoGame.Framework/Graphics/Effect/Effect.cs
```diff
@@ -228,68 +191,111 @@ internal protected override void GraphicsDeviceResetting()
...
-
-				// Read the parameter index values.
-				var parameters = new int[reader.ReadByte ()];
-				var offsets = new int[parameters.Length];
-				for (var i = 0; i < parameters.Length; i++) 
-                {
-					parameters [i] = (int)reader.ReadByte ();
-					offsets [i] = (int)reader.ReadUInt16 ();
-				}
+        struct MGFXHeader
```
harry-cpp, line 194: Move the struct back to the top please.

MonoGame.Framework/Utilities/BinaryReaderEx.cs
```diff
@@ -10,7 +10,7 @@ namespace MonoGame.Framework.Utilities
     public class BinaryReaderEx : BinaryReader
     {
         public BinaryReaderEx(Stream input) : base(input)
-        {
+        {
```
harry-cpp, line 13: We have no need for the changes to BinaryReaderEx / BinaryWritterEx.

MonoGame.Framework/Graphics/Effect/Effect.cs
```diff
@@ -455,6 +456,25 @@ private static EffectParameterCollection ReadParameters(BinaryReaderEx reader)
...
+        private EffectTechniqueCollection ReadTechniques(BinaryReaderEx reader)
+        {
+            var techniques = new EffectTechnique[reader.ReadCount()];
+
+            for (var t = 0; t < techniques.Length; t++)
+            {
+                var name = reader.ReadString();
+
+                var annotations = ReadAnnotations(reader);
+
```
harry-cpp, line 469: Remove the spaces between variables.

harry-cpp (changes requested):

MonoGame.Framework/Graphics/Effect/Effect.cs
```diff
@@ -228,68 +191,111 @@ internal protected override void GraphicsDeviceResetting()
...
+            /// </summary>            
+            public const int MGFXVersion = 10;            
+
+            public int Signature;
+            public int Version;
+            public int Profile;
+            public int EffectKey;
+            public int HeaderSize;
+
+            public static MGFXHeader ReadHeader(byte[] effectCode, int index)
```
harry-cpp, line 212: Why is this part of a struct now?

vpenades: It's basically the constructor for that class.

MonoGame.Framework/Utilities/BinaryReaderEx.cs
vpenades: I'll have to see how to undo all the changes from the backwards compatibility version.

MonoGame.Framework/Graphics/Effect/Effect.cs
```diff
@@ -300,27 +306,22 @@ private static EffectAnnotationCollection ReadAnnotations(BinaryReader reader)
             return new EffectAnnotationCollection(annotations);
         }
 
-        private static EffectPassCollection ReadPasses(BinaryReader reader, Effect effect, Shader[] shaders)
+        private static EffectPassCollection ReadPasses(BinaryReaderEx reader, Effect effect, Shader[] shaders)
```
harry-cpp, line 309: This can still use normal BinaryReader I assume.

MonoGame.Framework/Graphics/Effect/Effect.cs
```diff
@@ -234,62 +238,64 @@ private void ReadEffect (BinaryReaderEx reader)
...
-				}
-
-                var buffer = new ConstantBuffer(GraphicsDevice,
-				                                sizeInBytes,
-				                                parameters,
-				                                offsets,
-				                                name);
-                ConstantBuffers[c] = buffer;
-            }
+            ConstantBuffers = ReadConstantBuffers(reader, GraphicsDevice);
```
harry-cpp, line 241: Lets not do any refactor on these as well.

harry-cpp (changes requested):

harry-cpp (approved):

PR #7403, opened 2020-11-14
fix 7402: enqueue buffer ref before sending to OAL

### LINKED ISSUE #7402: DynamicSoundEffectInstance "Queue empty." crash in PlatformUpdateQueue
For DesktopGL using `DynamicSoundEffectInstance`s, `PlatformUpdateQueue()` can be called while `PlatformSubmitBuffer()` is executing in multi-threaded environments.

Normally that's not a problem, unless that happens to occur between [L92](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs#L92) and [L94](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs#L94).

L92 sends the buffer to OAL (which means OAL's buffer count will be n+1), and L94 enqueues the managed reference (so that MonoGame's buffer count will also be n+1) -- but when `PlatformUpdateQueue()` is executed in between those two lines it will blindly attempt to dequeue n+1 buffers before MonoGame's queue actually has n+1 buffers.

- MonoGame 3.6+ (since PR #4715 for issue #4700)

- Windows

- DesktopGL

lodicolo:
Fixes #7402 

Moves the `Enqueue()` call to before the buffer is sent to OAL so that all attempts to dequeue all of the buffers that OAL is aware of do not try to dequeue more buffers references than have been enqueued by `DynamicSoundEffectInstance`.

### DIFF
--- a/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs
+++ b/MonoGame.Framework/Platform/Audio/DynamicSoundEffectInstance.OpenAL.cs
@@ -89,9 +89,9 @@ namespace Microsoft.Xna.Framework.Audio
             }
 
             // Queue the buffer
+            _queuedBuffers.Enqueue(oalBuffer);
             AL.SourceQueueBuffer(SourceId, oalBuffer.OpenALDataBuffer);
             ALHelper.CheckError();
-            _queuedBuffers.Enqueue(oalBuffer);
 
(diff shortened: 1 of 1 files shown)


Jjagg:
Thanks! 😊

PR #7404, opened 2020-11-18
Fixes 3D Audio Direction Issues On Platforms Using OpenAL

squarebananas:
This pull request is to fix 3D audio direction issues on platforms using OpenAL. More details on the issue can be found here: https://github.com/MonoGame/MonoGame/issues/7388

The positions/velocities were projected into listener space incorrectly and the listener velocity was missing from the calculations.


Jjagg:
Awesome, thanks!

PR #7405, opened 2020-11-18
Fixes Stopping Looped Sounds On Platforms Using XAudio

squarebananas:
This pull request is to fix stopping looped sound effects on platforms using XAudio. More details on the issue can be found under item 3 here: https://github.com/MonoGame/MonoGame/issues/7388

Stopping a looping sound with immediate set to false didn't stop the sound effect at the end of the loop. This needed ExitLoop to be called instead.


Jjagg:
Thanks for these fixes!

Issue #6710, opened 2019-04-10
spriteBatch breaks Hardware Instancing


SaculRennorb:
Its quite simple realy, and I guess best explained by some screenshots:

this is how the rendering works if i only use `Vertex`- / `IndexBuffer`s + `DrawInstancedPrimitives`:
![instancing working fine](https://image.prntscr.com/image/it06axSgRZidsO6JzLxOcA.png)
its terrain generation, each chunk, outlined by the white lines, is one instance - everything works fine.

Then I wanted to add some debug draw - and everything went downhill.
If i remember correctly XNA used to Restore the `GraphicsDevice` `DepthStencilState` etc, sadly MG does not but whatever, not that big of a problem ... or so I thought.

If i only call `spriteBatch.Begin();` and `spriteBatch.End();` after my actual draw code it breaks some of the states like menioned before, but thats not the problem, those can easily just saved before and later restored. Atm Im doing this with the 

    GraphicsDevice.RasterizerState
    GraphicsDevice.BlendState
    GraphicsDevice.DepthStencilState
    GraphicsDevice.SamplerStates[0]

since I found those are changed in the `SpriteBatch.Setup()`. This fixes al problems, **but** as i said before:
>  If i **only** call `spriteBatch.Begin();` and `spriteBatch.End();`

This is what hapens if i actually try to use `DrawString(...)`:
![spectacular degenerate primitives](https://image.prntscr.com/image/fJznfW7ZS0iYxCvacR8b7A.png)
So yaaaaayy, it draws some strings - and it also breaks everything else.
Ive got a feeling that im missing some state change somewhere but after searching for hours, im just gona psot this issue.

One more thing; the white lines are actualy the edges of a cube, and are created with regualar draw calls, `DrawIndexedPrimitives` to be precise, and those work perfectly fine.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7
- MonoGame 3.8.0.356 (develop)

#### What operating system are you using:
- Windows 7

#### What MonoGame platform are you using:

- DesktopGL (universal VS project)

Jjagg:
> DesktopGL (universal VS project)

You mean the cross platform desktop project type, right? We refer to that as DesktopGL (though I don't think that's actually documented 😅).

[...]

PR #7406, opened 2020-11-18
Fixes Buffer Bindings Cache Issue On Platforms Using OpenGL

squarebananas:
This pull request is to fix a buffer bindings cache issue on platforms using OpenGL.

The issue (mentioned in https://github.com/MonoGame/MonoGame/issues/7193) is demostrated by the Draw project by OpenPachinko found here:
https://github.com/OpenPachinko/MonogameCompareBehaviors

The _bufferBindingInfos object within the GraphicsDevice contains a cache list of the last binding details for each vertex buffer slot set and is used to prevent unnecessary buffer bindings.

However when switching between using one and two vertex buffers (between non instanced and instanced draw events) the cached details of the second slot are dealt with as still being set even though they are potentially no longer current. This can cause issues as a non instanced draw call with a single vertex buffer can have overlapping element AttributeLocation values with the previously used second vertex buffer. Other details may also be out of sync.

For example in the above project the following element AttributeLocation values are used:

Draw Non Instanced
First Buffer Slot - Element Attribute Locations 0, 1

Draw Instanced
First Buffer Slot - Element Attribute Locations 2, 3, 4, 5
Second Buffer Slot - Element Attribute Locations 0, 1

One reason this appears to fail is because both the first and second buffer slots have element attribute locations that overlap  (0 & 1) between both draw events. As the cache list values match setting the second buffer is skipped even though these locations have been overwritten (the first frame will work but every frame after will fail). 

A fix for this is to treat cache list slots as inactive for vertex buffer slots that were not used in the last draw event.

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/GraphicsDevice.OpenGL.cs
+++ b/MonoGame.Framework/Platform/Graphics/GraphicsDevice.OpenGL.cs
@@ -116,6 +116,7 @@ namespace Microsoft.Xna.Framework.Graphics
         static readonly float[] _posFixup = new float[4];
 
         private static BufferBindingInfo[] _bufferBindingInfos;
+        private static int _activeBufferBindingInfosCount;
         private static bool[] _newEnabledVertexAttributes;
         internal static readonly List<int> _enabledVertexAttributes = new List<int>();
         internal static bool _attribsDirty;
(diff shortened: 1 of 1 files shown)


Jjagg:
Thanks, this is a great catch! We'll need to check if this was the same issue that caused #6710.

PR #7413, opened 2020-12-08
NUnit Test Attachments To preview Test images in VS.

vpenades:
This is a tiny feature of NUnit, that allows files produced by a unit test to be browsable withing visual studio.

Visual studio integration looks like this:

imagen


Jjagg:
Neat, thanks!

PR #7415, opened 2020-12-12
Fix library search path for macOS .app bundles

### LINKED ISSUE #7414: [CoreRT,macOS] Game crash when runing in app bundle but works in command line
Blank game cannot run in macOS app bundle. 

Test project: https://github.com/ryancheung/GameFoo

Steps to reproduce:
1. Clone the test project
2. Run `dotnet publish -c Release -r osx-x64 -t:BundleApp` to get a app bundle in `bin/Release/netcoreapp3.1/osx-x64/publish/`
3. Run the generated `Foo.app`
4. You'll see the crash similar with  https://gist.github.com/ryancheung/694f75e42649bbe95c30e5048527970e

- MonoGame 3.8

- macOS

- DesktopGL

ryancheung:
Fixes #7414

### DIFF
--- a/MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
+++ b/MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
@@ -65,6 +65,15 @@ namespace MonoGame.Framework.Utilities
             if (ret == IntPtr.Zero)
                 ret = LoadLibrary(Path.Combine(assemblyLocation, libname));
 
+            // Try MacOS folder for macOS .app bundles
+            if (ret == IntPtr.Zero)
+            {
+                // Correct the working directory location for macOS .app bundles
+                var currentDirectory = System.AppDomain.CurrentDomain.BaseDirectory;
+
+                ret = LoadLibrary(Path.Combine(currentDirectory, libname));
+            }
+
             // Try loading system library
             if (ret == IntPtr.Zero)
                 ret = LoadLibrary(libname);


MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
```diff
@@ -43,6 +43,8 @@ public static IntPtr LoadLibraryExt(string libname)
             // Try .NET Framework / mono locations
             if (CurrentPlatform.OS == OS.MacOSX)
             {
+                // Fix assembly location for macOS .app bundles
+                assemblyLocation = System.AppDomain.CurrentDomain.BaseDirectory;
```
harry-cpp, line 47: You shouldn't be modifying assemblyLocation, it will affect a lot of other ways of packaging stuff. Instead you would need to add one more if ret == IntPtr.Zero check.

Also noticed that you are using `Dotnet.Bundle` package, however as that thing is not open source, I don't think we should even support it.

harry-cpp (changes requested):

mrhelmut:
Hello Ryan,

the ```dotnet-bundle``` package doesn't package MonoGame games in a way that is compliant to the Mac app store requirements (they would likely fail certification) and wouldn't work on Steam even if you tweak MG's library discoverability path (because of how Steam injects its API).

I would recommend to not alter MonoGame library discoverability, and instead add a bootstrap script to your macOS and Linux packaged games to properly set the library path, like this (which is required anyway, otherwise Steam won't work properly):

```
#!/bin/bash
# .Net Core bootstrap script

# Move to script's directory
cd "`dirname "$0"`"

# Get the system architecture
UNAME=`uname`

# Setup LD path to the script current path
if [ "$UNAME" == "Darwin" ]; then
	# macOS
	export DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:./

	if [ "$STEAM_DYLD_INSERT_LIBRARIES" != "" ] && [ "$DYLD_INSERT_LIBRARIES" == "" ]; then
		export DYLD_INSERT_LIBRARIES="$STEAM_DYLD_INSERT_LIBRARIES"
	fi
else
	# Linux
	LD_LIBRARY_PATH=./
	export LD_LIBRARY_PATH
fi

./YourExecutable $@
```

This is something I need to find time to expand the documentation.

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
ryancheung: https://github.com/egramtel/dotnet-bundle this should be  the source of `Dotnet.Bundle`.
I folllowed this doc https://avaloniaui.net/docs/packing/macOS and I'm just too lazy to write the `Info.plist` for app bundle so I choose to use it to pack a app bundle.

And I believe it's not about `Dotnet.Bundle`, what it does is simply help us write `Info.plist` and copy files. I tried move *.dylib files to `.app/Content/Frameworks` directory and it's the same error. 

I'd print the `Directory.GetCurrentDirectory()` and it shows the current working directory is `/`, and then print  `assemblyLocation` is `./`, that's why search *.dylib fails.

ryancheung:
@mrhelmut Thanks for providing another way that makes library discovery working. But I think it's ok to just make library discovery working without a bootstrap script, in case someone don't need to publish to steam.

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
harry-cpp: What I'm trying to tell you is to do the following:
```cs
// Look in .app bundle
if (ret == IntPtr.Zero)	
    ret = LoadLibrary(Path.Combine(Directory.GetCurrentDirectory(), "something something"));
```

`assemblyLocation` is used outside of the if statement for a lot of other packaging methods, you should not modify it.

ryancheung: I just pushed and dont' change `assemblyLocation `.

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
```diff
@@ -43,11 +43,16 @@ public static IntPtr LoadLibraryExt(string libname)
             // Try .NET Framework / mono locations
             if (CurrentPlatform.OS == OS.MacOSX)
             {
-                ret = LoadLibrary(Path.Combine(assemblyLocation, libname));
```
harry-cpp, line 46: You must not remove this line.

harry-cpp (changes requested):

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
ryancheung: OK, wait a moment.

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
```diff
@@ -43,11 +43,16 @@ public static IntPtr LoadLibraryExt(string libname)
...
+                // Correct the working directory location for macOS .app bundles
+                // App's current working directory is actually `/`, so "./" won't work.
+                var currentDirectory = System.AppDomain.CurrentDomain.BaseDirectory;
+
+                // Look in MacOS for .app bundles
+                ret = LoadLibrary(Path.Combine(currentDirectory, libname));
```
harry-cpp, line 51: Now that I look at the entire file, this section is for .NET framework, your code needs to be added bellow.

harry-cpp (changes requested):

ryancheung:
@mrhelmut Just found even simple DllImport would fail:
```
        [DllImport("libMacUtils.dylib")]
        public static extern void setMenuBarVisible(bool visible);
```

Have you experience this before? It seems a bootstrap script is required.

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
```diff
@@ -65,6 +65,17 @@ public static IntPtr LoadLibraryExt(string libname)
...
 
+            // Try MacOS folder for macOS .app bundles
+            if (ret == IntPtr.Zero)
+            {
+                // Correct the working directory location for macOS .app bundles
+                // App's current working directory is actually `/`, so "./" won't work.
+                var currentDirectory = System.AppDomain.CurrentDomain.BaseDirectory;
+
+                // Look in MacOS for .app bundles
+                ret = LoadLibrary(Path.Combine(currentDirectory, libname));
```
harry-cpp, line 76: You are missing `if (ret == IntPtr.Zero)` check.

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
```diff
@@ -65,6 +65,17 @@ public static IntPtr LoadLibraryExt(string libname)
...
                 ret = LoadLibrary(Path.Combine(assemblyLocation, libname));
 
+            // Try MacOS folder for macOS .app bundles
+            if (ret == IntPtr.Zero)
+            {
+                // Correct the working directory location for macOS .app bundles
+                // App's current working directory is actually `/`, so "./" won't work.
+                var currentDirectory = System.AppDomain.CurrentDomain.BaseDirectory;
+
+                // Look in MacOS for .app bundles
```
harry-cpp, line 75: Redundant comment.

harry-cpp (changes requested):

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
ryancheung: I'm sure I have the check.

harry-cpp: Oh yea, you didn't check for macos with the line above, this is fine.

MonoGame.Framework/Platform/Utilities/FuncLoader.Desktop.cs
ryancheung: done.

mrhelmut:
As far as I explored this, I haven't found out a solution which seems to work properly by code with .NET Core/.NET5 because unlike Mono, it doesn't provide proper mechanisms to configure native dependencies. The bootstrap script is all I could come up with and which would work with any Mono/.NET runtime.

ryancheung:
Just tried the bootstrap script you provided. It still crash on PInvoke dylib. I'm using CoreRT.

harry-cpp:
This was taking too much time so I just made a separate PR with a proper comment message and no extra variable: https://github.com/MonoGame/MonoGame/pull/7416

I was just going to modify the files directly in this PR, but didn't have permission to push here.

ryancheung:
That's great.

PR #7417, opened 2020-12-14
Proposal for adding basic interoperability with System.Numerics.Vectors

vpenades:
This PR offers these features: 
- Adds the System.Numerics.Vectors dependency for all platforms.
- Defines implicit conversions from Numerics to XNA, and explicit conversions from XNA to Numerics.
- The conversion methods have been added to `MathNumerics.cs`, so they can be opted in/out on platforms that might not require them or where they might not be available (consoles?)

This PR was easy to do, so it's mostly a showcase on how to begin introducing numerics.vectors into XNA seamlessly.

If you find this might cause friction to projects, or having the system.numerics.vectors dependency is too much, feel free to reject this PR.


Jjagg (approved):
I like this. It's great for System.Numerics compatibility. I'd also like to see us fully migrate to System.Numerics in the future, but this at least makes it easier to use with MG.

IMO the conversion could even be implicit both ways, but I'm fine with this too.

vpenades:
> IMO the conversion could even be implicit both ways, but I'm fine with this too.

In my experience, having bidirectional implicit conversion is _dangerous_ because the moment you have "using" directives for the two types, the compiler can't resolve the ambiguity and throws errors.

The rationale is that with these conversions, a developer can choose to use XNA or Numerics types seamlessly to call MonoGame methods. The opposite conversion is less common, and to prevent compiler conflicts, it's better to keep an explicit conversion.

If anything, I would add explicit conversions from Numerics to XNA too....  something like a `static Vector3 From(System Numerics,,,);`

MonoGame.Framework/MathNumerics.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Xna.Framework
+{
+    public partial struct Vector2
```
harry-cpp, line 7: Put this inside Vector2 instead of making a partial class.

harry-cpp (changes requested):

MonoGame.Framework/MathNumerics.cs
vpenades: The reason because I've put them outside in a separate file is because I am worried about consoles not being able to include System.Numerics.Vectors dependency. If that's the case, a console project can simply remove the "MathNumerics.cs" file from compiling.

I don't how detailed knowledge of the console projects constraints, but if you know that system.numerics.vectors can be used, then you're right and these methods should be moved to their respective classes.

harry-cpp: If consoles can't support it, we won't be adding it at all, as having an important math API that does not work on consoles defeats the purpose of having it in the first place as 3rd party libraries won't be able to benefit from it.

Should be usable, thats why I'm saying to add it directly.

MonoGame.Framework/MonoGame.Framework.WindowsDX.csproj
```diff
@@ -80,6 +80,11 @@
     <PackageReference Version="4.0.1" Include="SharpDX.XInput" />
   </ItemGroup>
 
+  <!-- NetCore3.1, NetStandard2.1 and Net5 onwards already include Vectors-->
```
harry-cpp, line 83: I feel its self explanatory why a reference to System.Numerics exist, don't think we want comments like these in the .csproj. @Jjagg Thoughts?

Jjagg: Yeah, these should be removed.

vpenades: Referencing System.Numerics.Vectors has its tricks.... it only needs to be referenced below certain framework versions. Because it's in by default on higher versions. Also, Xamarin provides its own assembly, that's why in xamarin platforms the package should not be referenced in favour of the assembly reference. I don't think everybody is aware of these details, so if another contributor comes in and changes the frameworks versions without realizing that they should change the vectors reference...

harry-cpp: Build bot will stop them, don't worry :)

Not having to reference stuff when you get to new enough framework is quite normal for Microsofts system packages.

harry-cpp (approved):

mrhelmut:
FYI, this PR broke consoles. The package ```System.Numerics.Vectors``` even though supporting```net45``` is not compatible with the console runtime.

I'll submit a workaround in a moment.

tomspilman:
@mrhelmut - i need to make a big push to bring the cross-compiler up to date.  We can talk about this more later.

mrhelmut:
@tomspilman Alright, I'll wait for your update to check if it works.

tomspilman:
@mrhelmut - no don't wait on me... that is generally a bad idea.  I would work around it for now.

mrhelmut:
@tomspilman Roger that.

Would adding something like ```#if !NO_NUMERICS``` around numerics work for you? I can't really think of a cleaner way to handle this.

Or should we maybe write our own ```System.Numerics``` replacement so that we don't rely on package references?

mrhelmut:
I've pushed a proposal to #7432 with a dummy replacement.

PR #7432, opened 2021-01-12
Updated console references

mrhelmut:
Hello there,

I'm currently working on upgrading all console repositories to the latest ```develop``` structure.

I'd like to propose these changes which are more flexible to deal with specific situations.

@Jjagg @harry-cpp @tomspilman

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
@@ -220,33 +220,6 @@
     </Content>
   </ItemGroup>
 
-  <ItemGroup>
-    <Compile Include="..\Switch\Tools\MonoGame.Effect.Compiler\**\*.cs">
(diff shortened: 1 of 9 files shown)


Jjagg (approved):
No harm in doing this and if the flexibility helps, that's great :)

tomspilman:
This is great!  Go for it!

mrhelmut:
Before merging, I've updated this PR with a workaround for #7417. The idea is to have a target to use for platforms where ```System.Numerics.Vectors``` doesn't work.

What do you think?

vpenades:
@mrhelmut I understand that adding the source code directly is just a temporary fix, until the consoles cross compiler is able to deal with the system.numerics.vectors, am I right?

mrhelmut:
> @mrhelmut I understand that adding the source code directly is just a temporary fix, until the consoles cross compiler is able to deal with the system.numerics.vectors, am I right?

Yes, it's a quick workaround while we wait for console runtimes to support intrinsics.

Jjagg:
Great, thanks @mrhelmut! Merging :)

PR #7433, opened 2021-01-12
Removed residual OUYA references

mrhelmut:
Here we go @Jjagg :)


Jjagg:
Wow, a step into the future! 😂
Thanks 😉

Issue #7430, opened 2021-01-11
MGCB Incompatibilities with .NET 5.0, .NET Core 3.1


Apppppppple:
With a pipeline plugin targeting .NET Core 3.1 or higher (possibly earlier versions / different .net branches as well, but not .NET Standard 2.0) any classes or structs defined in the assembly that do not extend from an MG class will cause the plugin to break - the importers and processors don't show up in the editor, builds fail, etc.

Additionally, switching back to .NET Standard will oddly not solve the issue until the builds are cleaned.

To recreate download the zip below, build the PipelineExtension project, and open the mgcb editor to get the broken version. Uncomment `//: Game` in Breaker.cs and rebuild the project for the working version.
I know that mgcb is only built around net standard plugins but it would be nice to be able to use a newer version - I'd like be able to reference my main library in the plugin, and that one's on net5.0

This hasn't been tested on the dev branch as I don't know how to install / switch to it, but if someone could tell me how I can test it on there. It doesn't seem like any of the commits there would fix this, though.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

PR #7437, opened 2021-01-16
Upgrade tooling to .NET 5

harry-cpp:
@tomspilman Gonna need .NET 5 SDK installed on build bot for this to pass the build.

https://dotnet.microsoft.com/download/dotnet/5.0

fixes: #7430


tomspilman:
@mrhelmut - will this affect us on consoles?  I guess not since we control the .net version for the console builds?

tomspilman:
@harry-cpp - looks like maybe one project was missed?

```
  C:\Program Files\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\MonoGame.Content.Builder\MonoGame.Content.Builder.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [C:\BuildAgents\MonoGameWin1\work\f7381a85a626990\Tools\MonoGame.Content.Builder.Task\MonoGame.Content.Builder.Task.csproj]  C:\Program Files\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\MonoGame.Content.Builder\MonoGame.Content.Builder.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [C:\BuildAgents\MonoGameWin1\work\f7381a85a626990\Tools\MonoGame.Content.Builder.Task\MonoGame.Content.Builder.Task.csproj]
```

harry-cpp:
There is a bit of trouble with how we package up mgcb-editor, I'm looking into a way to resolve it.

tomspilman:
Got it... i think both the Windows and Mac boxes are updated now.   Let me know if you have any more trouble with it.

mrhelmut:
This won't affect console runtime libraries (which are NET45).

mrhelmut:
Dropping by to highlight this discovery: https://github.com/dotnet/sdk/issues/12125#issuecomment-774021450

mrhelmut:
FYI the build bot still fails on .NET because of https://github.com/dotnet/sdk/issues/12125#issuecomment-774181936

I have pushed a enforcement of Core 3.1 in #7462 so that we can keep moving until we have a better workaround or hopefully a fix to work with .NET 5 again (Microsoft has acknowledged the issue, so we can hope for a fix).

mrhelmut:
.NET Core 3.1 is now enforced through a ```global.json``` file. Whenever we have a workaround for the .NET 5 issue or a fix, we should remove it.

kreghek:
Does anybody work on tooling under NET 5? Is it possable to work around to build a Monogame project with only NET5/6?

mrhelmut:
I need to check if we can lift the 3.1 limitation. I haven't encountered the bug which makes the build bot to fail on NET5 in a while, which may indicate that it got fixed in a recent update.

@harry-cpp could you remove the ```global.json``` at the root in this PR to remove the 3.1 rule?

@tomspilman we might need to update NET5 to its latest patch on the build bot.

kreghek:
I use `highbyte/sonarscan-dotnet` Github Action which based on image with NET 6.0. There is error with MG tool. I use `Condition` in csproj and check mandatory variable SONAR_TOKEN to fix it.

```xml
<PackageReference Condition="'$(SONAR_TOKEN)' == ''" Include="MonoGame.Content.Builder.Task" Version="3.8.0.1641" />
```

May be it will help anybody until this PR is in draft.

mrhelmut:
@harry-cpp I gave it a test spin. Only issue that I've seen, is that the builder task still points to 3.1 tools:

https://github.com/MonoGame/MonoGame/blob/3debbf5247b38d5ae06d2ec241ff7bebab2824a7/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.props#L6

Otherwise we should be good to go.

After that (or as port of it), we should likely upgrade WindowsDX lib, DX/GL templates, and tests to net5.0 as well so that there are no more dependencies toward core3.1.

harry-cpp:
> After that (or as port of it), we should likely upgrade WindowsDX lib, DX/GL templates, and tests to net5.0 as well so that there are no more dependencies toward core3.1.

This was meant more of an upgrade for the tools only, so I would rather do this in a PR after this one.

mrhelmut:
> > After that (or as port of it), we should likely upgrade WindowsDX lib, DX/GL templates, and tests to net5.0 as well so that there are no more dependencies toward core3.1.
> 
> This was meant more of an upgrade for the tools only, so I would rather do this in a PR after this one.

Fair enough! I'll do that.

Merging, thanks!

PR #7440, opened 2021-01-16
[DesktopGL] Improve mouse handling outside gamewindow

harry-cpp:


harry-cpp:
Seems to work fine on all 3 desktop platforms.

PR #7442, opened 2021-01-17
Revert "Changed uses of Math to MathF where it was used for floats"

harry-cpp:
Reverts MonoGame/MonoGame#7390


mrhelmut:
The build bot, and I guess I know why. ```dotnet``` fails to create an app host because of antiviruses / Windows Defender. I've encountered the same thing locally and had to exclude projects and dotnet installation from realtime protection. We should also make sure that the build script has administrator rights.

Anyway, this PR is important. We have console users who are complaining about MathF because of #7390.

harry-cpp:
Ping @tomspilman

tomspilman:
@harry-cpp @mrhelmut - build box is fixed.

initram:
Can't console users just include the MathF replacement, like it is done when compiling for .Net framework? and why are they relying on the "devlop" branch to be stable?

tomspilman:
@initram - it isn't about `MathF` and consoles... it is about the precision loss concern.  Look at the convo here:  #7390

mrhelmut:
I'm good with this one.

PR #7443, opened 2021-01-17
Set pauseCount to zero on stop

### LINKED ISSUE #7372: Bug in SoundEffectInstance.Pause on Windows
If you Pause() a SoundEffectInstance, and then Stop() it while it's paused,  you can never pause it again.   

If you look at the code in SoundEffectInstance.OpenAL.cs you'll see that neither `PlatformStop()` nor `PlatformPlay()` reset the `pauseCount` and a `pauseCount` > 0 will prevent you from actually pausing the sound again in `PlatformPause()`, no matter how many times you stop or restart it.

- MonoGame 3.7

- Windows

DesktopGL

initram:
This should the resolve the issues with pause state getting out of sync when stopping a paused SoundEffectInstance.

I am not sure why we are keeping track of both SoundState and pauseCount, it seem rather unintuitive. When would it make sense for pause count to be any value other than 0 (not paused) and 1 (paused)? @mrhelmut  maybe you can remember? it seems you added the code some 5 years ago.

I would recommend that we just get rid of the pauseCount, but I don't know the code and the quirks well enough.

Fixes #7372

### DIFF
--- a/MonoGame.Framework/Platform/Audio/SoundEffectInstance.OpenAL.cs
+++ b/MonoGame.Framework/Platform/Audio/SoundEffectInstance.OpenAL.cs
@@ -170,6 +170,7 @@ namespace Microsoft.Xna.Framework.Audio
         private void PlatformStop(bool immediate)
         {
             FreeSource();
+            pauseCount = 0;
             SoundState = SoundState.Stopped;
         }
 


initram:
Better PR #7453 has been submitted that include a test case.

Issue #6298, opened 2018-04-24
Error in BoundingBox.Contains(BoundingFrustum)


Apostolique:
https://github.com/MonoGame/MonoGame/blob/6f4e51030120913bc1d5b21b34b07493d2eeebff/MonoGame.Framework/BoundingBox.cs#L71

The function ```BoundingBox.Contains(BoundingFrustum)``` doesn't work properly. If I have a BoundingFrustum that is outside a BoundingBox, the returned value will be ```ContainmentType.Intersects```.

If I instead call the equivalent (?) function on the ```BoundingFrustum.Contains(BoundingBox)```, I get the right result, namely ```ContainmentType.Disjoint```.
https://github.com/MonoGame/MonoGame/blob/6f4e51030120913bc1d5b21b34b07493d2eeebff/MonoGame.Framework/BoundingFrustum.cs#L179

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL, WindowsDX, NetCore

nkast:
The BoundingFrustum code is over complicated. 
The [construction says](https://github.com/MonoGame/MonoGame/blob/6f4e51030120913bc1d5b21b34b07493d2eeebff/MonoGame.Framework/BoundingFrustum.cs#L128) that it's defined as a View * Projection matrix.

Let's look at [Contains(Vector3 point)  ](https://github.com/MonoGame/MonoGame/blob/6f4e51030120913bc1d5b21b34b07493d2eeebff/MonoGame.Framework/BoundingFrustum.cs#L294-L306)
If you have a vector3 in world coordinates and you tranform it by ViewProj, you get this point in normalized device coordinates (NDC a.k.a. Clip coordinates).  You then check if the point is withing -1 to +1 (-1 to 0 for Z).

You can extract corners and planes the same way. Tranform the corner (1,1,0) from Clip coordinates with inverse ViewProj into world coordinates.
I bet you can simplify the transform by taking the matrix Left,right,Up,down, etc. (it's multiplying with 1 and 0 after all).

Same goes for Planes, Spheres, boxes. Tranform them with ViewProj and then check if it's inside the NDC.

baudii:
Has anyone solved this yet? It was first mentioned 6 years ago and I see there isn't even a test case for this scenario.

PR #7444, opened 2021-01-24
6298 Error in BoundingBox.Contains(BoundingFrustum)

Tauroctonos:
Fixed logical error in frustum containment method; see #6298

### DIFF
--- a/MonoGame.Framework/BoundingBox.cs
+++ b/MonoGame.Framework/BoundingBox.cs
@@ -111,8 +111,6 @@ namespace Microsoft.Xna.Framework
         /// </returns>
         public ContainmentType Contains(BoundingFrustum frustum)
         {
-            //TODO: bad done here need a fix. 
-            //Because question is not frustum contain box but reverse and this is not the same
             int i;
             ContainmentType contained;
             Vector3[] corners = frustum.GetCorners();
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/BoundingBox.cs
```diff
@@ -132,20 +130,20 @@ public ContainmentType Contains(BoundingFrustum frustum)
                 return ContainmentType.Intersects;
 
 
-            // If we get here, it means the first (and only) point we checked was actually contained in the frustum.
-            // So we assume that all other points will also be contained. If one of the points is disjoint, we can
-            // exit immediately saying that the result is Intersects
+            // If we get here it means the first (and only) point we checked was outside the frustum, so we assume
```
initram, line 133: The point we checked here was outside the box, not the frustum.

MonoGame.Framework/BoundingBox.cs
```diff
@@ -132,20 +130,20 @@ public ContainmentType Contains(BoundingFrustum frustum)
...
             for (; i < corners.Length; i++)
             {
                 this.Contains(ref corners[i], out contained);
-                if (contained != ContainmentType.Contains)
+                if (contained == ContainmentType.Contains)
                     return ContainmentType.Intersects;
 
             }
 
             // If we get here, then we know all the points were actually contained, therefore result is Contains
```
initram, line 145: This comment should also be updated to something like: If we get here, then we know all the points are actually outside the box, therefore result is Disjoint.

On that note, is it not possible to have an intersection between a BoundingBox and BoundingFrustum where none of their corners are contained within the other volume? The Frustum is not box shaped and is not axis aligned, so a corner could technically pass through the whole of the other volume and not be contained within while still intersecting.

Tauroctonos: I had the same thought, but if the Frustum's bigger than the Box then it can't really be contained by it. Maybe that case should return an Intersect instead though

Tauroctonos: You could have an intersection where the Fustrum's bigger than the box, but if you're calling `.Contains()` on the box while it's entirely contained in the Fustrum it makes sense to still return Intersects I think.
So, I think if all the Fustrum's points are outside the box, we can check to see if the box is contained in the Fustrum, and if we get anything but Disjoint return Intersects.

initram:
Have you considered this edge case? I would think that would result in intersect.
edge case

Here no corner is inside the volume of the other shape, so just checking corners are not enough to handle all intersection cases.

Tauroctonos:
> Have you considered this edge case? I would think that should result in intersect.
> edge case
> 
> Here no corner is inside the volume of the other shape, so just checking corners are not enough to handle all intersection cases.

tomspilman:
This needs a test case for us to merge @Tauroctonos . 

If you or someone adds one we can merge it.

SimonDarksideJ:
@Tauroctonos are you able to provide the Unit test to accompany this PR, or should it be closed?

PR #7445, opened 2021-01-25
Update SDL to 2.0.14

mrhelmut:
Bumping the dependencies to upgrade SDL 2.0.14 and get the latest fixes and controller support.

@Jjagg @harry-cpp


mrhelmut:
@harry-cpp @Jjagg The mac tests timed out, I assume it is related to SDL but I can't seem to find evidence of it in the logs. It seems that something crashes upon disposing the Game instance.

Jjagg:
I think the failed test is the first one that actually creates a `GraphicsDevice`.

Something changed in graphics initialization with the SDL update I think.

mrhelmut:
Weird, I have this binary working with our games.

Jjagg:
It might be specific to how our tests are set up somehow.

mrhelmut:
I don't see anything in [the 2.0.14 changelog of SDL](https://www.libsdl.org/tmp/SDL/WhatsNew.txt) which may hint toward something.

Maybe @harry-cpp has some insights.

harry-cpp:
Try re running the build bot?

mrhelmut:
@Jjagg @harry-cpp I relaunched the bot with a rebase, same issue. I don't know what's going on and don't see anything different in the logs.

I don't know if we're missing something from SDL 2.0.14 on macOS (I didn't find anything relevant in the changelog) or if it's related to how I built SDL. We have this binaries published to players without issues.

I presently can't run the tests, our mac will be gone for several weeks.

mrhelmut:
Okay, I've updated the mac SDL with a clean rebuild, making sure that the backcompat flags are there and all, to no avail.

Jjagg:
Must be a build bot issue :/

mrhelmut:
I don't know. I was assuming that it could be related to the infrequent bot issues with the app host failure on .NET 5, but that doesn't explain that only this PR fails every time. Something is going on with SDL. Unfortunately I can't test presently.

harry-cpp:
Cmon TestMac, you can do it buddy!

harry-cpp:
Oh no :(

mrhelmut:
We got our mac back. Dunno if it's our mac or me, but debugging tests on VS for Mac is a very unpleasant experience. I'm not able to step through them it, and once the test gets stuck, it permanently prevent VS from running new tests and I have to kill VS.
It seems to be related with the creation of the graphics device, can't be sure though.

mrhelmut:
Okay, it's ```SDL_GL_CreateContext``` which gets stuck and doesn't return. I have no idea why. Everything works outside of tests (we shipped publicly this binary, no issue so far).

mrhelmut:
I give up on this one, I have no clue. My best guess is that a change within SDL made the context creation not liking the threaded nature of NUnit. There is a way to force single threaded tests on NUnit 3.12+ which I tried, but it fails as well because it's not compatible with either .Net Core or .Net 5 (tried both) or possibly macOS.

What seems sure is that it's not an SDL build issue, nor a function pointer marshalling issue.

harry-cpp:
Have you tried running tests from terminal?

mrhelmut:
Yes, it gets stuck from both ```dotnet test``` and VS for Mac test explorer.

mrhelmut:
Here we go.

I have identified that SDL 2.0.14 introduced a problematic change in regard to threading and the OpenGL context on macOS: https://bugzilla.libsdl.org/show_bug.cgi?id=4932

So I reverted the related change set and rebuilt SDL 2.0.14: https://hg.libsdl.org/SDL/rev/19c4516de68d

I've reopened this discussion over the SDL bug tracker to check with the SDL team if we are doing something wrong or if the threading could be improved on their end.

While we wait for a possible 2.0.15, I believe that it would good to switch to 2.0.14 with this change for fixes and controllers support.

PR #7449, opened 2021-01-27
Upgrade to use Xamarin.Legacy.Sdk [WIP]

dellis1972:
Switch the Android and iOS projects over to use [Xamarin.Legacy.Sdk](https://github.com/xamarin/Xamarin.Legacy.Sdk).
This is a new set of targets which allow us to produce both legacy Xamarin TargetFramework based Assemblies, as well as .net 6 ones. Currently `MSBuild.Sdk.Extras` is not able to do this, and it makes sense to allow the nuget packages support both old and new frameworks.

Todo

- [ ] Upgrade Nuget Packages for iOS and Android to include both legacy and .net based assemblies and .net 6 ones. 
- [ ] Figure out how to upgrade the Build system to use `dotnet build` rather than calling `msbuild` for iOS and Android (This might already happen in cake)
- [ ] Install the required .net 6 previews on the build hosts so we can build those.
- [ ] Possibly upgrade other projects to support .net 6 (possibly a different PR).
- [ ] Run unit tests under .net 6 (probably in a different PR).

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.Android.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.Android.csproj
@@ -9,7 +9,6 @@
     <PackageTags>monogame;.net core;core;.net standard;standard;android</PackageTags>
     <PackageId>MonoGame.Framework.Android</PackageId>
     <AppendTargetFrameworkToOutputPath>True</AppendTargetFrameworkToOutputPath>
-    <LangVersion>7.3</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/MonoGame.Framework/MonoGame.Framework.Android.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.Android.csproj
@@ -1,7 +1,7 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Xamarin.Legacy.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0-android</TargetFramework>
+    <TargetFrameworks>monoandroid90;net6.0-android</TargetFrameworks>
     <DefineConstants>ANDROID;GLES;STBSHARP_INTERNAL</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <MGOpenGL>GLES</MGOpenGL>
@@ -9,6 +9,7 @@
     <PackageTags>monogame;.net core;core;.net standard;standard;android</PackageTags>
     <PackageId>MonoGame.Framework.Android</PackageId>
     <AppendTargetFrameworkToOutputPath>False</AppendTargetFrameworkToOutputPath>
+    <LangVersion>7.3</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
@@ -73,7 +74,7 @@
     <Compile Include="Platform\Utilities\FuncLoader.Android.cs" />
     <Compile Include="Platform\Utilities\InteropHelpers.cs" />
     <Compile Include="Platform\Utilities\ReflectionHelpers.Default.cs" />
-    
+
     <Compile Include="..\ThirdParty\StbImageSharp\src\**\*.cs" LinkBase="Utilities\StbImageSharp"/>
     <Compile Include="..\ThirdParty\StbImageWriteSharp\src\**\*.cs" LinkBase="Utilities\StbImageWriteSharp"/>
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


harry-cpp:
Is there a reason to support anything bellow .net 6 for android/ios once its out?

Jjagg:
At least while .NET 6 is in preview we should support the old stuff.

dellis1972:
Upstream MS will be supporting the old system for a while so that people have time to migrate. It should not be much effort (at the moment) to support both and package both in the nuget.

mrhelmut:
Any update on this now that NET 6 is available?

@tomspilman May we get NET 6 to be installed on the build bot?

mrhelmut:
Hey Dean, now that net6 has been merged, would you mind updating your PR?

mrhelmut:
In regard to this, would it be safe enough to you if we consider that 3.8.1 is for net6 only, and that we still have 3.8.0 around for monodroid9/xamarinios10?

MonoGame.Framework/MonoGame.Framework.Android.csproj
```diff
@@ -1,14 +1,15 @@
...
+    <TargetFrameworks>monoandroid90;net6.0-android</TargetFrameworks>
     <DefineConstants>ANDROID;GLES;STBSHARP_INTERNAL</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <MGOpenGL>GLES</MGOpenGL>
     <Description>The MonoGame runtime for Android which supports Android 4.2 and newer.</Description>
     <PackageTags>monogame;.net core;core;.net standard;standard;android</PackageTags>
     <PackageId>MonoGame.Framework.Android</PackageId>
-    <AppendTargetFrameworkToOutputPath>False</AppendTargetFrameworkToOutputPath>
+    <AppendTargetFrameworkToOutputPath>True</AppendTargetFrameworkToOutputPath>
+    <LangVersion>7.3</LangVersion>
```
mrhelmut, line 12: Is it possible target C#10 or remove LangVersion? Or this is a requirement to back target those frameworks?

MonoGame.Framework/global.json
```diff
@@ -1,5 +1,6 @@
 {
     "msbuild-sdks": {
-        "MSBuild.Sdk.Extras": "2.0.54"
+        "MSBuild.Sdk.Extras": "2.0.54",
+        "Xamarin.Legacy.Sdk": "0.1.0-alpha1"
```
mrhelmut, line 4: Shouldn't this be bumped to 0.1.2-alpha6?

MonoGame.Framework/global.json
```diff
@@ -1,5 +1,6 @@
 {
     "msbuild-sdks": {
-        "MSBuild.Sdk.Extras": "2.0.54"
+        "MSBuild.Sdk.Extras": "2.0.54",
+        "Xamarin.Legacy.Sdk": "0.1.0-alpha6"
```
mrhelmut, line 4: You missed the .2 in 0.1.2-alpha6

dellis1972: 🤦

mrhelmut:
Thanks Dean! It seems okay but the github actions are failing because it misses xamarinios10 (surprisingly this doesn't seem to be a problem for monoandroid90 which builds properly). I'll look into it.

There also seems to be a lot of conflicting namespace warnings (e.g. MathF, System.Numerics which have local net452 implementations). I don't know how impactful they are, I'll have to test this out.

mrhelmut:
I don't know if we can solve this with github actions. As I understand it, ```windows-latest``` doesn't have the VS2019 toolchain installed, we would need ```windows-2019``` for that, which in turn doesn't support the VS2022 toolchain. @harry-cpp thoughts?

dellis1972:
I'll chat with the author and see if it still works with Xamarin.iOS

MonoGame.Framework/MonoGame.Framework.Android.csproj
```diff
@@ -1,14 +1,14 @@
...
-    <TargetFramework>net6.0-android</TargetFramework>
+    <TargetFrameworks>monoandroid90;net6.0-android</TargetFrameworks>
     <DefineConstants>ANDROID;GLES;STBSHARP_INTERNAL</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <MGOpenGL>GLES</MGOpenGL>
     <Description>The MonoGame runtime for Android which supports Android 4.2 and newer.</Description>
     <PackageTags>monogame;.net core;core;.net standard;standard;android</PackageTags>
     <PackageId>MonoGame.Framework.Android</PackageId>
-    <AppendTargetFrameworkToOutputPath>False</AppendTargetFrameworkToOutputPath>
+    <AppendTargetFrameworkToOutputPath>True</AppendTargetFrameworkToOutputPath>
```
nkast, line 11: @mrhelmut 
We should make AppendTargetFramework true for all projects 
for consistency and to allow future upgrades and testing.

harry-cpp: Agreed.

tomspilman:
@dellis1972  is this still importaint?

dellis1972:
> @dellis1972 is this still important?

I think we dropped support for the legacy xamarin.* sdks in a previous release. So I guess its not needed anymore.

PR #7451, opened 2021-01-28
Changed Math to MathF

### LINKED ISSUE #7382: Math to MathF

initram:
Tried to avoid reducing precision of calculations this time.

Fixes #7382

I would like to make sure that this is verified for precision loss before merging, so that we don't end up rolling it back again. @nkast  do you have time to look this through? It is mostly the same changes as last time, except that I have avoided using MathF where it was used for double precision.

Also only DesktopDx will benefit from the changes as it is now, as that is the only platform that publishes a netcore3.1 package. (NetStandard2.0 does not have support for MathF). But we can look into that at a later point.

### DIFF
--- a/MonoGame.Framework/Audio/Xact/Cue.cs
+++ b/MonoGame.Framework/Audio/Xact/Cue.cs
@@ -255,7 +255,7 @@ namespace Microsoft.Xna.Framework.Audio
                     direction /= distance;
                 var right = Vector3.Cross(listener.Up, listener.Forward);
                 var slope = Vector3.Dot(direction, listener.Forward);
-                var angle = MathHelper.ToDegrees((float)Math.Acos(slope));
+                var angle = MathHelper.ToDegrees(MathF.Acos(slope));
                 var j = FindVariable("OrientationAngle");
                 _variables[j].SetValue(angle);
                 if (_curSound != null)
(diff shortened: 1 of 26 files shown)


initram:
Does no one have time to look at this?

nkast:
I checked every line of your previous PR, and commented on the ones where the results are affected.
If this is the same as #7390 minus the ones where precision was reduced, then I suppose it's ok.

initram:
That is exactly what it is. I could not change the original pull request as it had already been merged.

mrhelmut (approved):

mrhelmut:
We should be good to go on this one. Waiting for review.

harry-cpp:
Honestly, I was fine with: https://github.com/MonoGame/MonoGame/pull/7390

That being said, you'll always have a bit of precision loss on the normal .NET Framework side as it does do reordering of operations to give you more precise results: https://dotnetfiddle.net/HvbP1M

harry-cpp (approved):

PR #7453, opened 2021-01-29
Fixed GLDesktop PlatformStop() resolving issue #7372

### LINKED ISSUE #7372: Bug in SoundEffectInstance.Pause on Windows

JacksonDorsett:
Bug addressed: When using DesktopGL the pause counter field is not reset when stop is called.

In SoundEffectInstance.PlatformStop() on the OpenAL platform. pauseCounter is set to 0 if its value is greater than 0. 

Fixes #7372.

### DIFF
--- a/Tests/Framework/Audio/SoundEffectInstanceTest.cs
+++ b/Tests/Framework/Audio/SoundEffectInstanceTest.cs
@@ -9,6 +9,7 @@ using Microsoft.Xna.Framework.Audio;
 using NUnit.Framework;
 using Microsoft.Xna.Framework.Content;
 using Microsoft.Xna.Framework.Graphics;
+using System.Threading;
 
 namespace MonoGame.Tests.Audio
 {
@@ -26,9 +27,6 @@ namespace MonoGame.Tests.Audio
         /// is called after calling Pause(), Stop().
         /// </summary>
         [Test]
-#if !DESKTOPGL
-        [Ignore("bug is platform specific to GLDesktop")]
-#endif
         public void SoundEffectPauseStopPlay()
         {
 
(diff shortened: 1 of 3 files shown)


mrhelmut (approved):

initram:
Should the order of the calls i the test not be: play, pause, stop, play?

Tests/Framework/Audio/SoundEffectInstanceTest.cs
```diff
@@ -0,0 +1,54 @@
...
+            // Necessary to get audio initialised
+            FrameworkDispatcher.Update();
+        }
+
+        /// <summary>
+        /// Unit test for issue #7372 where the Sound effects instance does not play after Play()
+        /// is called after calling Pause(), Stop().
+        /// </summary>
+        [Test]
+#if !DESKTOPGL
```
Jjagg, line 29: No need to exclude this for other platforms, so this directive can be removed.

Jjagg (changes requested):
Thanks for the PR :)

Tests/Framework/Audio/SoundEffectInstanceTest.cs
JacksonDorsett: I originally intended to make it a test for both platforms but when I tried testing the DirectX Audio, its implementation sets the state to stop rather than pause. 
I noticed that after additional testing that the Assert.AreEqual(SoundState.Stopped, instance) is runs to quickly after calling stop. This can be temporarily fixed by making the thread sleep for 10 ms before running the test, but it may be an indicator of additional bugs.

JacksonDorsett:
The automated build may be broken since the last commit was a comment and it passed two more checks.

Jjagg:
Yes, there's some issues on the build bots unrelated to this PR.

Thanks again!
This looks good to me, though we might hold off merging until we get the build bot fixed.

Jjagg (approved):

mrhelmut:
Could you rebase your changes so that the build bot rechecks everything? Once it's all green, I'll proceed with merging it.

mrhelmut:
@JacksonDorsett mind checking this PR for a rebase?

mrhelmut:
I have manually run the test since the buildbot is down. We're good with this PR, thank you!

Issue #7458, opened 2021-02-05
Android not resuming to app after using Exit(); via backbutton


7unk10:
I investigated it and found that in AndroidGameActivity.cs if that EventHandlers.Raise is commented out, I can resume like expected. See https://community.monogame.net/t/resume-android-app-running-in-the-background/14894/23
I don't understand it any further..

#### What version of MonoGame does the bug occur on:
-monogame 3.8.0.1641

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- Android

mrhelmut:
As per Android life cycle management, it is not recommended to call ```Exit()```.

PR #7459, opened 2021-02-05
Fixes Android not resuming to app after using Exit(); via backbutton

Kwyrky:
Android not resuming to app after using Exit(); via backbutton #7458
https://github.com/MonoGame/MonoGame/issues/7458
but needs probably more investigation because maybe the commented source is rerquired for other things.


mrhelmut:
The recommended life cycle management of Android apps is to never close them programmatically and let the system handle its life cycle.
So if we follow this guideline, we should actually prevent Exit() to even work in the first place (or at least display a debug message explaining that it isn't recommanded).

PR #7462, opened 2021-02-08
Help with compatibility toward NativeAOT

### LINKED ISSUE #7460: Incompatibilities with NativeAOT/CoreRT
Hello there,

NativeAOT (formerly known as CoreRT) has introduced breaking changes in how it process the marshalling of structures (see https://github.com/dotnet/runtimelab/issues/604).

Presently, some code paths of MonoGame will result in runtime crashes.

Here are notes to make MonoGame more reliant toward native compilation:
- Replace ```Marshal.SizeOf(typeof(T))``` by ```Marshal.SizeOf<T>()``` everywhere;
- ```Enum.GetValues(typeof(Keys))``` within ```KeysHelper``` needs to be changed (so far it concerns only WindowsDX, but it's along a hot path nonetheless);
- ```ReflectionHelpers.ManagedSizeOf()``` or its use should be reworked. As-is, it makes ```DynamicIndexBuffer``` prone to crashes.

Further notes:
- ```ContentManager.ReloadGraphicsContent()``` would fail because it relies ```MethodInfo.MakeGenericMethod()``` but this an Android path only, so can be ignored (it will raise a compilation warning by NativeAOT, though);

- MonoGame 3.8

- Windows

- DesktopGL
- WindowsDX

mrhelmut:
Hello there,

This might provide a better reliance to NativeAOT even though the expected types should be safe enough already.

Fixes #7460

cc @Jjagg @harry-cpp 

EDIT: the build bot is still failing on .NET 5 because of this: https://github.com/dotnet/sdk/issues/12125#issuecomment-774181936

I have pushed a ```global.json``` to force the build bot to use .NET Core 3.1.
This will affect #7437 until .NET 5 fixes this issue.

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/GraphicsDevice.OpenGL.cs
+++ b/MonoGame.Framework/Platform/Graphics/GraphicsDevice.OpenGL.cs
@@ -1217,7 +1217,7 @@ namespace Microsoft.Xna.Framework.Graphics
         private void PlatformGetBackBufferData<T>(Rectangle? rectangle, T[] data, int startIndex, int count) where T : struct
         {
             var rect = rectangle ?? new Rectangle(0, 0, PresentationParameters.BackBufferWidth, PresentationParameters.BackBufferHeight);
-            var tSize = Marshal.SizeOf(typeof(T));
+            var tSize = Marshal.SizeOf<T>();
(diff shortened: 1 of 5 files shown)


Jjagg:
Looks good, thanks!

PR #7463, opened 2021-02-09
Fix UWP Vsync support

### LINKED ISSUE #6020: [UWP] Disabling V-Sync crashes MG

mrhelmut:
Hi there,

UWP Vsync was broken following #7228 because the swap chain wasn't initialized with the proper flag (it was a partial implementation).

This PR also makes it so UWP apps will not crash on older Windows where tearing is not supported for UWP apps (Vsync off will be ignored instead of crashing; I believe it's non existant case only happening on out of support Windows versions like Windows 7 on which the Windows Store isn't available, so I don't think it's worth throwing a ```NoSuitableGraphicsDeviceException``` here).

Fixes #6020 

cc @Jjagg @harry-cpp


Jjagg:
Awesome, thanks!

PR #7466, opened 2021-02-12
Fix type collisions and default implementations for consoles

mrhelmut:
Hello,

This is a PR to exclude the dummy ```System.Numerics.Vectors``` types to be included and generating collisions on some of the latest SDKs.


mrhelmut:
I have also moved ```WaveBank``` default implementation to the Platform folder. This will help console repo's structure.

@harry-cpp @Jjagg

PR #7472, opened 2021-02-24
Removed threading limitations

### LINKED ISSUE #7471: BlockOnUIThread crashing on low end hardware
Hello there,

#7384 introduced ```MaxWaitForUIThread``` to the ```Threading``` class.

https://github.com/TechnologicalPizza/MonoGame/blob/6176b242eb799e9b876ca7e107779a3897c202a6/MonoGame.Framework/Platform/Threading.cs#L22

We didn't see any problem in this at first because we likely run on machines which are powerful enough to not hit this wall, but after publishing this to our players, we immediately started to run into those _very_ frequently on calls as simple as loading a small texture.

I believe that we should entirely get rid of both of this limit since increasing it will likely just hide things and still be hit on some machines. Maybe we should just use them to display a warning to developers, but it should not crash.

I'll try to work on a PR for this.

cc'ing @TechnologicalPizza @roeyskoe 

- MonoGame 3.8

- All systems

- DesktopGL

mrhelmut:
Hello,

This PR removes the time threshold on ```BlockOnUIThread``` actions which was crashing depending on the action data size and hardware speed. Since we can't predict a maximum duration for every kind of operations and hardware, it is likely best to wait indefinitely for a blocked action to return (this was the behavior of MonoGame before #7384 introduced the threshold).

cc @Jjagg @harry-cpp 

Fixes #7471

### DIFF
--- a/MonoGame.Framework/Platform/Threading.cs
+++ b/MonoGame.Framework/Platform/Threading.cs
@@ -19,9 +19,6 @@ namespace Microsoft.Xna.Framework
 {
     internal class Threading
     {
-        public const int MaxWaitForUIThread = 1000; // In milliseconds
-        public const int MaxPooledResetEvents = 32; // Throw away objects when above this amount
-
         static int _mainThreadId;
 
         static Stack<ManualResetEventSlim> _resetEventPool = new Stack<ManualResetEventSlim>();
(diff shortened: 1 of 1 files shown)


Jjagg:
I agree with this and considered setting up a PR too :) 

The time limit was a breaking change that can cause very annoying issues for users. 

Crashes that happen in specific situations on *some* machines are the worst and there might be cases where waiting out certain operations is the only solution. In case of timeout, the only thing game devs could do is retry and hope the operation goes through faster, which is a bad solution.

I'm merging this, thanks @mrhelmut.

PR #7477, opened 2021-03-10
Support drawing point primitives

roy-t:
Hey everyone,

For a particle system similar to https://directtovideo.wordpress.com/2009/10/06/a-thoroughly-modern-particle-system/ I wanted to draw point primitives. To my big surprise MonoGame only supports lines and triangles.

I've made a small PR that shows how easy it to add support for point primitives. I believe I can also make it work with the OpenGL version of MonoGame. However before I proceed I want to check if this PR would be welcomed? I don't have a Mac, Linux machine or any knowledge of iOS or Android so for those parts I would need help.

Please let me know :)

PS.

Apologies for the formatting changes, it seems a few files had mixed LF and CRLF line endings and I couldn't figure out a way to not change that.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -1435,6 +1435,8 @@ namespace Microsoft.Xna.Framework.Graphics
                     return primitiveCount * 3;
                 case PrimitiveType.TriangleStrip:
                     return primitiveCount + 2;
+                case PrimitiveType.PointList:
+                    return primitiveCount;
             }
 
             throw new NotSupportedException();
--- a/MonoGame.Framework/Graphics/Vertices/PrimitiveType.cs
+++ b/MonoGame.Framework/Graphics/Vertices/PrimitiveType.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Xna.Framework.Graphics
     /// Defines how vertex data is ordered.
     /// </summary>
 	public enum PrimitiveType
-	{
+    {
         /// <summary>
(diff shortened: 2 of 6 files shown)


roy-t:
The current implementation is tested and works on the Windows/DirectX.

Here you see 1 million (instanced) point primitives
OneMillionParticles

mrhelmut:
Would you mind looking into adding the OpenGL support? It should be as straightforward. FYI, ```GLPrimitiveType.Points``` would be ```0x0```.

roy-t:
@mrhelmut sure, this should do it I think :)

mrhelmut:
Thank you. This looks good to me, though the test bot is presently down so the review will be slightly delayed.

viniciusjarina (commented):
I would fix the white space changes

Apostolique:
> 
> 
> Thank you. This looks good to me, though the test bot is presently down so the review will be slightly delayed.

This is offtopic, I see the the build bot often has issues. It should be possible to do a lot of the work using a GitHub Actions pipeline. I could set it up myself. Do you think there's any interest in that? @tomspilman @Jjagg @harry-cpp @mrhelmut

roy-t:
> I would fix the white space changes

I'll try to fix that if you think it is necessary, but I can't even touch GraphicsDevice.cs in Visual Studio without it changing some of the indentations. My suggestion would be to leave the whitespace changes so that its not a problem for the next PR.

roy-t:
Just checking in, is there anything holding up this MR? From my point of view it looks done. I haven't had comments on this in almost two months :).

mrhelmut (approved):

mrhelmut:
Good to go to me. Waiting for another pair of eyes. :)

mrhelmut:
I'm going to merge this one since it looks relatively safe. Thanks!

Issue #7457, opened 2021-02-04
PS4 Controller Init Bug


KelsamGames:
Originally posted here: https://community.monogame.net/t/duplicate-controller-bug/14906

I do not know if this is a bug with MonoGame, SDL, or something even deeper. Note: I've found that it is easiest to test this with a USB-connected PS4 controller, as Windows bluetooth seems to hold on to the PS4 bluetooth connection even after disconnecting for a while for some reason.

If anyone can shed some light on this, it'd be super helpful.

**If you start a MonoGame program with a PS4 controller connected and some other controller (tested with Xbox 360 and Xbox One controllers), instead of MG reporting the PS4 and Xbox controller, there are 2 instances of the Xbox controller** on 2 different GamePad indexes, and both report button presses to polling and such. If you disconnect the Xbox controller and reconnect, then the new one will work as expected, but leaves behind the "ghost" duplicate that does not report button presses or anything. If you connect the PS4 controller after the program starts (presumably, after MG/SDL have initialized), then everything works as expected, but if the programs starts with the PS4 controller and other controller connected you can see the bug. If you disconnect the PS4 controller, the "ghost" remains until you restart the program (presumably, until MG/SDL/both have re-Init-ed).

I wonder if this is a MonoGame bug, an SDL bug, or perhaps something even deeper. Seemingly not a driver-level bug, because Windows doesn't report multiple copies of the Xbox device, and this only happens if you Init the program with both connected at the same time.

**Tested with MonoGame 3.8 - DesktopGL on Windows 10 x64**

Here's a simple MG game class to test what I have described above (you will need a simple font content file to render the debug info - in this code it loads an Arial.xnb spritefont):

```
using Microsoft.Xna.Framework;
using Microsoft.Xna.Framework.Graphics;
using Microsoft.Xna.Framework.Input;
using System;
using System.Diagnostics;


namespace ControllerReader
{
	public class Game1 : Game
	{
		private GraphicsDeviceManager _graphics;
		private SpriteBatch _spriteBatch;

		string outputStr;
		SpriteFont font;

		public Game1()
		{
			_graphics = new GraphicsDeviceManager(this);
			Content.RootDirectory = "Content";
			IsMouseVisible = true;
			IsFixedTimeStep = false;
		}

		protected override void Initialize()
		{
			_graphics.PreferredBackBufferWidth = 1600;
			_graphics.PreferredBackBufferHeight = 900;
			_graphics.ApplyChanges();

[...]

PR #7486, opened 2021-04-02
Fixes Duplicate/Missing Gamepads On Platforms Using SDL

squarebananas:
This pull request is to fix duplicate/missing gamepads on platforms using SDL.

This relates to issue #7457

When SDL builds it's initial list of controllers internally it can shuffle the detected controllers around and reorder them while doing so. SDL doesn't wait for the initial list to be fully built but rather each controller detected fires a JoyDeviceAdded event containing it's pre-shuffled device index before being reordered.

So when Monogame later reads the event's device index values (JoystickDevice.Which) these values are now potentially incorrect and really shouldn't be used with the SDL.Joystick.Open function. Because Monogame currently does use these values it creates duplicate controllers / has missing controllers.

A fix to this is whenever a JoyDeviceAdded event occurs attempt to open all available Joysticks (Sdl.Joystick.NumJoysticks gives the number of connected Joysticks) and don't use the received and likely out of date device index at all. Calling Sdl.Joystick.Open on an already opened Joystick simply returns it's existing pointer without issue, so checking if this already exists in Monogame's Joystick dictionary is all that is needed to prevent duplicates.

This fix is working for me but it could do with more testing with different controller/joystick combinations.

### DIFF
--- a/MonoGame.Framework/Platform/Input/Joystick.SDL.cs
+++ b/MonoGame.Framework/Platform/Input/Joystick.SDL.cs
@@ -22,9 +22,7 @@ namespace Microsoft.Xna.Framework.Input
         internal static void AddDevice(int deviceId)
         {
             var jdevice = Sdl.Joystick.Open(deviceId);
-            foreach (var entry in Joysticks)
-                if (jdevice == entry.Value)
-                    return;
+            if (Joysticks.ContainsValue(jdevice)) return;
 
             var id = 0;
 
--- a/MonoGame.Framework/Platform/Input/Joystick.SDL.cs
+++ b/MonoGame.Framework/Platform/Input/Joystick.SDL.cs
@@ -12,9 +12,20 @@ namespace Microsoft.Xna.Framework.Input
         internal static Dictionary<int, IntPtr> Joysticks = new Dictionary<int, IntPtr>();
         private static int _lastConnectedIndex = -1;
 
+        internal static void AddDevices()
+        {
+            int numJoysticks = Sdl.Joystick.NumJoysticks();
+            for (int i = 0; i < numJoysticks; i++)
+                AddDevice(i);
+        }
+
         internal static void AddDevice(int deviceId)
         {
             var jdevice = Sdl.Joystick.Open(deviceId);
+            foreach (var entry in Joysticks)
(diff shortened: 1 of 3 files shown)


squarebananas:
Not sure why but now the original code is working. Even though I tried SDL 2.0.14 earlier I may have not cleaned the project properly. It seems switching between 2.0.12 and 2.0.14 sends the correct device index in the event now. Interestingly 2.0.14 for me is now giving the Xbox controller's GamePadCapabilities.DisplayName as my mobile phone name, so something still isn't quite right but the GamePadState is correct.

I'll close this request if it is confirmed in the issue post that this fix isn't necessary.

mrhelmut:
Thank you for working this out. Thinking about it, I now believe to already have encountered this issue with ghost devices (from remote desktop apps or xinput emulators) blocking legit controllers from working.

I believe that it is a welcome fix even though it might have been stabilized within SDL. Better be safe than sorry.

mrhelmut (approved):

MonoGame.Framework/Platform/Input/Joystick.SDL.cs
```diff
@@ -12,9 +12,20 @@ static partial class Joystick
...
+            for (int i = 0; i < numJoysticks; i++)
+                AddDevice(i);
+        }
+
         internal static void AddDevice(int deviceId)
         {
             var jdevice = Sdl.Joystick.Open(deviceId);
+            foreach (var entry in Joysticks)
+                if (jdevice == entry.Value)
+                    return;
```
Jjagg, line 27:
```suggestion
            if (Joysticks.ContainsValue(jdevice)) return;
```

Jjagg (approved):
I agree with @mrhelmut, just a little nit from me :)

squarebananas:
No problem and thanks.

I'm still getting used to github. Do I need to click the "Commit Suggestion" button and add a commit message or does that happen from your side?

Jjagg:
> I'm still getting used to github. Do I need to click the "Commit Suggestion" button and add a commit message or does that happen from your side?

We can both do that. I went and committed it so I could merge :)
Thanks for the PR!

PR #7496, opened 2021-04-22
Reload disposed assets

mrhelmut:
Hey there,

Currently, if we manually dispose an asset loaded from the ```ContentManager``` and try to reload it, it will return the old disposed instance instead of a fresh one.

This PR checks the state and reloads the asset if it has previously been disposed and a reload is attempted.

cc @Jjagg @harry-cpp

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -245,7 +245,14 @@ namespace Microsoft.Xna.Framework.Content
             object asset = null;
             if (loadedAssets.TryGetValue(key, out asset))
             {
-                if (asset is T)
+                // Check if we're retrieving a disposed asset
+                // and if so, let's reload it instead
+                bool assetDisposed = false;
+                var disposable = asset as IDisposable;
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -245,7 +245,14 @@ public virtual T Load<T>(string assetName)
             object asset = null;
             if (loadedAssets.TryGetValue(key, out asset))
             {
-                if (asset is T)
+                // Check if we're retrieving a disposed asset
+                // and if so, let's reload it instead
+                bool assetDisposed = false;
+                var disposable = asset as IDisposable;
+                if (disposable != null && disposable.IsDisposed)
```
TechnologicalPizza, line 252: Where does .IsDisposed come from here?

mrhelmut: Ahah, from a local extension. Got to rework it, hold on.

mrhelmut:
Nevermind, it's likely more suitable to move forward with #6663 instead.

PR #7501, opened 2021-05-14
Fix FontDescriptionProcessor kerning

### LINKED ISSUE #5098: Fixed incorrect glyph position in FontDescriptionProcessor.cs

nkast:
fixes #5098

https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-abc

https://docs.microsoft.com/en-us/dotnet/desktop/wpf/advanced/introduction-to-the-glyphrun-object-and-glyphs-element?view=netframeworkdesktop-4.8#glyph-metrics

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs
@@ -145,6 +145,8 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 			abc.A = face.Glyph.Metrics.HorizontalBearingX >> 6;
 			abc.B = face.Glyph.Metrics.Width >> 6;
 			abc.C = (face.Glyph.Metrics.HorizontalAdvance >> 6) - (abc.A + abc.B);
+			abc.A -= face.Glyph.BitmapLeft;
+			abc.B += face.Glyph.BitmapLeft;
 
 			// Construct the output Glyph object.
 			return new Glyph(character, glyphBitmap)


mrhelmut:
This might be the proper fix instead of #7218.

mrhelmut:
I have properly tested this change against different fonts with kerning and it seems to be the way to go! Thanks!

wldevries:
I believe this fix to be incorrect. Kerning when rendering text with Segoe UI is off. [BitmapLeft](
https://github.com/Robmaister/SharpFont/blob/26454a8b5b53733ff7a10de2322469b64cb75ee8/Source/SharpFont/GlyphSlot.cs#L231) is the described as:
> Gets the bitmap's left bearing expressed in integer pixels. Of course, this is only valid if the format is <see cref="GlyphFormat.Bitmap"/>.

The BitmapLeft value is similar to scale adjusted `HorizontalBearingX` and should be used instead for the A value (distance before drawing the glyph bitmap). B, the width of the drawn portion of the character glyph, is equal to the width of the bitmap. And C, the distance to add to the current position to provide white space to the right of the character glyph, is the horizontal advance minus the distance taken up by A and B.
```
abc.A = face.Glyph.BitmapLeft;
abc.B = face.Glyph.Bitmap.Width;
abc.C = (face.Glyph.Metrics.HorizontalAdvance >> 6) - (abc.A + abc.B);
```
Example rendering with Segoe UI 24px. In yellow the default 3.8.1 font processor and in white the adjustments above:
image

wldevries:
Of course after reverting this PR the problem will start acting up with the _Kingthings Petrock_ font. I believe the actual source of the problem to be in the rectangle packing that was added in 3.8.1.

In `GlyhCropper.Crop` the `XOffset`, `XAdvance` are adjusted, but those values are unused. It adjust `glyph.CharacterWidth.A`, `B and `C` as well. To make the code more clear it would probably be best to remove the unused properties on `GlyphData` to avoid problems like this in the future.

When `GlyphCropper.Crop` also adjust the ABC values in addition to the changes I mentioned above, both the Segoe UI font and the Kingthings Petrock font render well.

PR #7515, opened 2021-05-26
Remove range check in SoundEffectInstance.Pitch.set

notexplosive:
Hi! First time contributing, please be gentle!

I ran into a problem where I couldn't set the Pitch of a sound effect instance below `-1.0f` or above `1.0f` without throwing an exception. This is on purpose, according to the docs, but 1 octave up/down is really limiting!

I saw discussion of it in [this ancient PR](https://github.com/MonoGame/MonoGame/pull/4852) and it seems there's no technical reason for this limitation. The best justification I can imagine is "That's how XNA did it." (I have no evidence of this, total guess).

My proposal is pretty straightforward: we just get rid of the range check and let AL handle it. I'm not changing any of the math for backwards compatibility, plus 1.0 == 1 octave is actually pretty elegant, I just wish I could have 2.0, or 10.0!

I'd be happy to add/modify a test for this, but I wasn't able to run the unit tests because I can't get Monogame to build on my machine (attempting to build via Visual Studio 2019 using `MonoGame.Framework.DesktopGL.sln` and/or `MonoGame.Tools.Windows.sln`).


tomspilman:
This seems fairly safe to me.  Even if porting an old XNA game depended on it being range limited like it was before they surely didn't let it overflow and capture exceptions during gameplay.

The only thing i would suggest is to decide on what the rules are with range of the pitch moving forward.  Do we guarantee at least 2 octaves on all platforms?  Do we need more than 2 octaves?

mrhelmut:
While removing the limit here makes sense, we have to be careful as to which value we pass to the implementation. As of now, implementions are assuming this limit. Uncapping it might expose them to crashes.

For example, not all OpenAL implentations are equal, Android might crash below -1 and above 1,  while on PC any value should be safe (OpenAL and XAudio should be able to support near infinite shifts on PC).

This PR looks safe in regard to DesktopGL and WindowsDX but we should safe check on mobile and consoles (I'll try to safe check the console side of things).

mrhelmut:
I looked into every platforms expected range and error handling:
- DesktopGL from ```float.MinValue``` to ```float.MaxValue``` (though we can expect the tolerance to be hardware dependent)
- WindowsDX from  ```-10.0f``` to ```10.0f``` (input values are clamped)
- Android from  ```-1.0f``` to ```1.0f``` (no clamping, error to be expected)
- iOS seems undocumented, it seems fair to assume that it's like Android (no clamping, error to be expected)
- Console 1 range from  ```-4.0f``` to ```4.0f``` (unspecified)
- Console 2 range from  ```float.MinValue``` to ```2.0f``` (no clamping, error to be expected)
- Console 3 range from  ```float.MinValue``` to ```2.0f``` (no clamping, error to be expected)
- Console 4 range from ```-10.0f``` to ```10.0f``` (input values are clamped)


So in regard to this PR, we should add a clamp on OpenAL ```PlatformSetPitch()```: -1;1 for mobiles, and something reasonable for DesktopGL (just to be safe, because I don't believe that anything beyond 10 is safe throughout all hardware, and I would even expect some hardware to fail way before that).

We have nothing to do for WindowsDX since it will clamp the values to +/- 10 octaves.

For consoles, we definitely need to add a clamp (I'll look into it if this PR moves forward).

tomspilman:
@notexplosive  if you address what @mrhelmut  mentioned we can move forward with this.  Or i can close it.

notexplosive:
If I understand the feedback correctly, I think I did what @mrhelmut requested.  (sorry for the force-push, I screwed up line endings and had to un-screw up the line endings).

SimonDarksideJ:
Any update @notexplosive ?, so long as the ranges posted by @ThomasFOG are implemented, I think we are good to go here (with a rebase against development)

notexplosive:
Hey @SimonDarksideJ, here's my update:
- I have clamped the range to [-1, 1] on iOS and Android as @ThomasFOG requested. I clamp to -10, 10 on all other platforms.
    - It's up to the console implementations to clamp further if they need to, which Thomas said he'd take care of after this PR
    - I'm willing to go lower than [-10, 10] if that makes people nervous.
- It looks like we're waiting on the new dotnet runtime to allow my Clamp functions to build 
    - Maybe this already happened since Thomas's comment was from over a year ago.
- I've fixed a bug in this PR (where I set `_pitch = value` _before_ clamping `value`) and I've rebased against develop. Oops! Good thing I had so long to ruminate on this PR otherwise I'd have missed that.
- I also re-introduced the `xAct` check, if `_isXAct` is true, we don't do any clamping (this mirrors the old behavior).

ThomasFOG:
> * Maybe this already happened since Thomas's comment was from over a year ago.

Unfortunately we have a blocking issue on one specific console that makes the move impossible there. Which means that a .NET 4.5 compatibility is still desired. I would suggest working around the absence of ```System.Math.Clamp(Single, Single, Single)```.

SimonDarksideJ:
Back to you @squarebananas for the final mile

notexplosive:
On it! sorry for the delay!

SimonDarksideJ:
Looks like this is still failing the console check, you might want to run locally to identify the language issues @notexplosive

ThomasFOG:
```highestAbsoluteValue``` should be ```validePitchRange```, I guess?

SimonDarksideJ (approved):

Issue #7520, opened 2021-06-04
iOS `Texture2D` constructor from bgthread hangs


Mindfulplays:
Calling `new Texture2D` or `Texture2D.SetData` from background thread queues up a workitem in `Threading.BlockOnUIThread` however, the workitem never gets invoked.

I found out that [unlike Android/SDL platforms](https://github.com/MonoGame/MonoGame/search?q=threading.run), `Threading.Run` is not invoked in iOS (and possibly other platforms?) in [`iOSGameWindow`](https://github.com/MonoGame/MonoGame/blob/a9e5ae6befc40d7c86320ffdcfcd9d9b66f786a8/MonoGame.Framework/Platform/iOS/iOSGameWindow.cs).

I worked around this by moving the call to the UI thread.
At a high-level, is there anything that prevents moving the texture loading to happen from a worker thread as opposed to the main UI thread? This will avoid any jank even if there is a loading screen being presented.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- MacOS

#### What MonoGame platform are you using:
iOS/OpenGL

TechnologicalPizza:
OpenGL only allows operations from the main thread, so there must be some synchronization.
But this iOS issue seems like a missed bug.

Mindfulplays:
Got it, I figured as much but my knowledge is from a decade ago so was hoping things changed😄

Happy to send a fix later if this is considered an actual bug.

mrhelmut:
Yeah, trying to load OpenGL textures from a thread on iOS is extremely prone to corrupting textures or eventually generating errors.
```Threading.Run()``` missing is an oversight. If anyone has the time to setup a PR for this, that would be very welcome.

Mindfulplays:
This is my first time contributing, could you please let me know if I have done anything foolish with my PR process?
Thank you.

PR #7522, opened 2021-06-10
Fix iOS GamePlatform to run pending background tasks #7520

Mindfulplays:
In other `Platform` classes (such as `SDLGamePlatform` and `AndroidGamePlatform`, `Threading.Run()` is called during `Tick()` or `RunLoop()` (inconsistently, it may be called before or after `Game.Tick()` unfortunately).

Fixing this for `iOSGamePlatform`. `WinForms` and perhaps other platforms (Switch?) may suffer from the same issue.

Note that texture loading from background thread is discouraged however a lot of computationally intensive operations (reading/decoding the image file streams, allocating CPU buffers) could be done in the background thread and hence is a good practice to move such work alone into the bgthread until we request the actual texture from the main thread.

### DIFF
--- a/MonoGame.Framework/Platform/iOS/iOSGamePlatform.cs
+++ b/MonoGame.Framework/Platform/iOS/iOSGamePlatform.cs
@@ -225,6 +225,7 @@ namespace Microsoft.Xna.Framework
             //        directly to NSTimer.CreateRepeatingTimer.
             _viewController.View.MakeCurrent();
             Game.Tick ();
+            Threading.Run();
 
             if (!IsPlayingVideo)
             {


mrhelmut (approved):

mrhelmut:
Looks relevant to me. Just requesting another pair of eyes in case I might not be aware of any restriction which may have explained why this hasn't been done in the first place (but most likely it was just missing).

tomspilman:
This seems ok, but it clearly needs people to test it.  Only thing i noticed is that Android runs it before the `Game.Tick()`...

https://github.com/MonoGame/MonoGame/blob/a9e5ae6befc40d7c86320ffdcfcd9d9b66f786a8/MonoGame.Framework/Platform/Android/AndroidGameWindow.cs#L80

... but not sure if that is anything to be concerned about.

Mindfulplays:
I agree, in other `*Platform` classes, I have noticed the `Tick()` being before the `Threading.Run()`. So not sure if there is a 'correct way'...

Will wait for the other reviewers. I ran the tests but however there are no iOS tests. :(

Mindfulplays:
@Jjagg @harry-cpp could you please take a look?
Thanks

Mindfulplays:
@Jjagg @harry-cpp Could you please take a look? Thanks!

mrhelmut:
@Mindfulplays Have you tested yourself that loading textures from a thread works with this PR? I don't think that we have iOS testers available at the moment.

Mindfulplays:
Hi Sorry for the delay. Yes I tested this on iPhone 8/iPad Mini 2. Both texture loading from the background and 'SetData/new Texture2D' calls on the UI thread.
thanks!

Mindfulplays:
What is the right way to move this fix asking?
Is monogame maintained?, as I am not seeing much code flow these days...

mrhelmut:
Much of it is stable and bigger works are being worked on. Though there is a lack of contributors for mobile platforms, hence we presently have no time or people to test this PR (which is a sensitive enough change that we can't just trust the automated tests).

harry-cpp:
Sorry for the delay, and thanks @Mindfulplays , merging!

Mindfulplays:
Thank you @harry-cpp , appreciate it.

PR #7530, opened 2021-07-03
Event key to the first frame of key press

SeteCogumelun:
Added IsKeyDownInThisFrame() and IsKeyUpInThisFrame() method. These methods return true only in the first frame in the key press or release. Knowing that every developer needs these events, I think it would be interesting to add them.


SeteCogumelun:
Now I see that this same type of event was previously denied.

PR #7531, opened 2021-07-08
Removed PS Vita support

mrhelmut:
Hello there,

Build submissions for the PS Vita are now permanently closed. It doesn't make sense anymore to support it.

This PR removes the platform support and marks it as deprecated if set as ```MonoGamePlatform```.

I would be in favor of entirely removing it with no deprecation warning. Let me know if I should go ahead with this or not.

### DIFF
--- a/Documentation/articles/platforms/0_platforms.md
+++ b/Documentation/articles/platforms/0_platforms.md
@@ -8,7 +8,6 @@ MonoGame supports the following systems:
 - Android
 - iOS
 - PlayStation 4
-- PlayStation Vita
 - Xbox One
 - Nintendo Switch
 - Google Stadia
--- a/Documentation/articles/tools/mgcb.md
+++ b/Documentation/articles/tools/mgcb.md
@@ -82,13 +82,12 @@ Set the target platform for this build. It must be a member of the [TargetPlatfo
 * PlayStation4
 * WindowsPhone8
 * RaspberryPi
-* PSVita
 * XboxOne
 * Switch
 
 If not set, it will default to Windows.
 
-NOTE: PlayStation 4, Xbox One, PS Vita, and Switch support is only available to licensed console developers.
+NOTE: PlayStation 4, Xbox One, and Switch support is only available to licensed console developers.
 
 ### Target Graphics Profile
 
--- a/Documentation/articles/tools/mgfxc.md
+++ b/Documentation/articles/tools/mgfxc.md
@@ -42,11 +42,10 @@ The `/Profile` option defines the platform we're targeting with this effect file
 - DirectX_11
 - OpenGL
 - PlayStation4
-- PSVita
 - XboxOne
 - Switch
 
-NOTE: PlayStation 4, Xbox One, PS Vita, and Switch support is only available to licensed console developers.
+NOTE: PlayStation 4, Xbox One, and Switch support is only available to licensed console developers.
 
 ### Help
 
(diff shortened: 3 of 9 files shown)


wardPlaced:
there are sixteen million ps vitas a float around the globe according to
wikipedia... i guess you can just leave everything as is... plus, i think
the ps4 still interacts with the vita, so if people want some kind of dual
screen gameplay on playstation or just want to make a portable niche indie
title you could leave that option as is... just an idea or suggestion...

Am Do., 8. Juli 2021 um 18:32 Uhr schrieb Thomas Altenburger <
***@***.***>:

RUSshy:
i agree, that's not because it's not officially supported that people don't want to use monogame to build homebrews

+1 to keep it

tomspilman:
I have no clue about the possibilities in the homebrew space.  If MG for Via works in homebrew we could leave it all in i guess, it just won't be actively supported by us.

@mrhelmut ?

mrhelmut:
MonoGame implements the official PS Vita SDK, which is not a support for homebrew. Sony has deprecated the SDK and it is no more possible to submit builds to them.

The SDK is not public and it is not legal (per Sony's terms of use) to build homebrews with the official SDK (even though hackers might be using the official SDK unofficially, it is a use-case that we can't promote or endorse for legal reasons).

As-is, the vita mentions are useless. It is an empty shell with no working implementation.

I don't see a way to leverage the current implementation into what would be a public homebrew support since it is not legal to reuse any of it.

Adding homebrew support would be an entirely new platform and unofficial SDK to implements. Until someone look into working on such project, I believe that keeping the mentions to the PS Vita are misleading.

Jjagg:
I think the best course is to tag the last build with Vita support, mention it somewhere in the docs, and remove all Vita code from the repo. People can still fork and use the code as they like, including for homebrew stuff if that works for them.

tomspilman:
@Jjagg agreed.

mrhelmut:
@tomspilman Would you mind adding a tag to the current develop? I'll update this PR's documentation to note that the platform is deprecated and pointing to that specific tag.

wardPlaced:
thanks for the considderation and support of my humble oppinions, there are still independent vita games made and sold in low quantities as offline printed releases, I did not know sony closed off the support for vita ports/builds.

mrhelmut:
> thanks for the considderation and support of my humble oppinions, there are still independent vita games made and sold in low quantities as offline printed releases, I did not know sony closed off the support for vita ports/builds.

This is a misconception. All PS Vita cart production has ended last February. Factories are now all dismantled. The reason why games are still releasing physically is because publishers got them manufactured in advance before the February closing, and kept them unreleased to spread them throughout their agenda to keep the demand relevant.

Also, Sony has now entirely closed submission of new games and patches. It is now strictly impossible to develop or manufacture anything for the PS Vita through official means.

And again, the MonoGame implementation on PS Vita uses the official SDK and is private, which is illegal to use for non-registered Sony developers, and the MonoGame implementation cannot be made publicly available because that would breach Sony's term of use of the SDK. I also don't believe that any of it might be salvaged for anyone to work on an unofficial homebrew support.

As-is, the couple references of the PS Vita in the public MonoGame repository are meaningless and are confusing users that a homebrew support exists. I would advise to remove PS Vita references entirely, and if one day someone wants to implement an entirely new homebrew support, let them add them back.

tomspilman:
@mrhelmut 

https://github.com/MonoGame/MonoGame/releases/tag/last_psvita

mrhelmut:
Good to go!

PR #7533, opened 2021-07-15
Compute Shader

cpt-max:
This PR adds compute shader support for DirectX and OpenGL.
It's based on #7352 which already adds tessellation and geometry shaders.

The following use cases are covered (with sample projects):

- Fill up a StructuredBuffer in a compute shader, then use it directly in another shader stage ([sample project](https://github.com/cpt-max/MonoGame-Shader-Samples/tree/compute_gpu_particles))
- Fill up a StructuredBuffer in a compute shader, then download the data to the CPU ([sample project](https://github.com/cpt-max/MonoGame-Shader-Samples/tree/compute_cpu))
- Write directly to a texture from a compute shader ([2D texture sample](https://github.com/cpt-max/MonoGame-Shader-Samples/tree/compute_write_to_texture), [3D texture sample](https://github.com/cpt-max/MonoGame-Shader-Samples/tree/compute_write_to_3d_texture))
- Write directly to a vertex or index buffer from a compute shader ([sample project](https://github.com/cpt-max/MonoGame-Shader-Samples/tree/compute_write_to_vertex_buffer))
- Indirect drawing  ([simple sample](https://github.com/cpt-max/MonoGame-Shader-Samples/tree/object_culling_indirect_draw), [complex sample](https://github.com/cpt-max/MonoGame-Shader-Samples/tree/indirect_draw_instances))

[Compute Shader Guide for MonoGame](https://github.com/cpt-max/Docs/blob/master/MonoGame%20Compute%20Shader%20Guide.md)
[NuGet Packages, Platform Support and Build Requirements](https://github.com/cpt-max/Docs/blob/master/Build%20Requirements.md)

[Shader Migration Guide for OpenGL](https://github.com/cpt-max/Docs/blob/master/Migrating%20shaders%20to%20ShaderConductor.md).
[Shader Model Support for OpenGL](https://github.com/cpt-max/Docs/blob/master/Shader%20models%20for%20OpenGL.md).


cpt-max:
I added a computer shader guide to the description on top.

cpt-max:
I added support for writing to textures. A sample has been added on top, and the compute shader guide has been updated.

- Currently only Texture2D is supported, no Texture3D or TextureCube.
- Only mipmap level 0 can be written to. 
- Writing to texture arrays is not yet possible.

cpt-max:
Writing to vertex and index buffers is now also possible, see sample project and the updated compute shader guide on top.

cpt-max:
3D textures are now also accessible from a compute shaders. A sample project can be found on top, even though nothing is fundamentally different than with 2D textures.

Texture arrays are currently not working for GL, maybe once #7361 is merged, I'll add support for those.
Writing to mip levels could be added, but I want to look into indirect drawing next. If anyone is interested in mip level support, I'd like to here about it.

cpt-max:
Indirect drawing is now also possible. While most people have some idea what compute shaders can do, you may not be familiar with indirect drawing. I added a short explanation to the indirect draw chapter in the compute shader guide.

One thing to look out for with OpenGL at the moment, if you have two buffers of the exact same type
```HLSL
RWStructuredBuffer<Ship> FirendlyShips;
RWStructuredBuffer<Ship> EnemyShips;
```
the 2nd buffer will get a name change, so in C# you have to set the parameters like this
```
effect.Parameters["FirendlyShips"].SetValue(firendlyShips);
effect.Parameters["EnemyShips_1"].SetValue(enemyShips);
```
This is only for OpenGL, and should of course be solved eventually: https://github.com/KhronosGroup/SPIRV-Cross/issues/1721

JohnLouderback:
This is extremely exciting. Really hope to see this merged soon. 🙂

cpt-max:
Shader compilation for OpenGL works natively on Linux now, so Wine is not needed anymore (tested on Ubuntu 20.04). I also published NuGet packages, so you can try this out without building from source. The samples have been updated to use the custom packages.

cpt-max:
Since this PR was recently added to the 3.9 Milestone let me mention some decision that need to be made:

**What happens to MojoShader and shader model 2 and 3 support?** 
- Drop shader model 2 and 3 support completely. I wouldn't mind, but a lot of people probably will.
- Use MojoShader to compile shader model 2 and 3 shaders, use ShaderConductor for shader model 4 and 5. I like this option quite a bit, as It guarantees full backwards compatibility.
- Use ShaderConductor also for shader model 2 and 3. The problem here is that ShaderConductor doesn't work very well for those shader models. It's an open source project and the developers don't seem to care that much about outdated shader models. I don't blame them.

**On Mac and iOS OpenGL is still limited to verion 2**

This is because OGL is created using the legacy context there. 
Windows and Linux use the compatibility context which allows mixing of OGL 2, 3 and 4. Apple doesn't support the compatibility context at all. If you want shader model 4+ on Apple platforms you have to create an OGL core context. If you do that you loose OGL 2 support though. That means unlike on Windows/Linux you won't be able to mix shader models 2, 3, 4 and 5. You have to make the decision up front: shader model 2 and 3 using the legacy context, or shader model 4 and 5 using the core context.
 
This probably means that the feature level (HiDef, Reach) needs to be used for GL as well, currently only DX seems to use it. You would need to make the decision between legacy and core context based on the feature level. It's probably also time for a 3rd feature level: SuperHighDef? HiDef is for DX 10 level hardware, compute shaders need DX 11 hardware. 

Oh and of course compute shaders won't work on Apple platforms through OGL anyway, because they stopped at OGL 4.2, compute shaders are 4.3. But shader model 4 and 5 should work, once the core context is supproted.

damian-666:
looks to me like absolutely world-changing mind blowing stuff..    i keep telling scientists to stop with the python, c++, linux only code,  the direct to the metal code that i cant even compile after 5 hours.   stop scala, java , jupiter notbooks, CUDA, open CL, , and i  tell that matlab code is basically C#. i hope AMD works too with this .  wtht all those samples u make i think legacy code can be dealt with either way.    This forces harware / OS/ and chip makers to make better  and cheaper hardware,  not lazily try trapping customers  because of legacy or arcane code.  i dont need it now but hoep it can get in.. who doesnt want 100-1000x performance?  ur samples should make it so that they can convert their old shaders if breaking changes

Ravendarke8:
> 
It is world changing. Even fairly old HW supports compute shaders, honestly, as long as we are talking desktop, every mainstream GPU have full compute functionality. Majority of rendering techniques these days use compute in some way. Efficiently written forward+ / deferred+ (tiled) is impossible without compute. Every state aware particle system runs in compute these days. Hell, there are compute rasterizers that beat normal path (Nanite's rasterization) as well as "software" raytracing. MG supporting compute is absolutely crucial and from where I stand, in last six years that I can remember, this is single handedly most important PR for MG I've seen.

Ravendarke8:
> **What happens to MojoShader and shader model 2 and 3 support?**
> 
>     * Drop shader model 2 and 3 support completely. I wouldn't mind, but a lot of people probably will.
> 
>     * Use MojoShader to compile shader model 2 and 3 shaders, use ShaderConductor for shader model 4 and 5. I like this option quite a bit, as It guarantees full backwards compatibility.
> 
>     * Use ShaderConductor also for shader model 2 and 3. The problem here is that ShaderConductor doesn't work very well for those shader models. It's an open source project and the developers don't seem to care that much about outdated shader models. I don't blame them.

I would say drop SM 2 - 3 support, BUT that's more in line with MG 4.x+ plans I guess. Hence I completely agree with second option "Use MojoShader to compile shader model 2 and 3 shaders..." I think it is win win scenario for everyone..

mrhelmut:
I believe that we should be fine dropping SM 2 and 3, we need to move forward at some point. People who need SM 2 or 3 could still use MG 3.8.x.

There is no clear plan yet, but it's likely going to be hard to maintain one single version of monogame that supports each and every scenario. I would vouch for a "legacy" version (something like MG 3.8.x or 3.9.x) with the current OpenGL and Direct3D 11 backends, calling it LTS, and then move into modern stuff with a new numbering (and working with Direct3D 12, Vulkan, and Metal from here).

I also wonder if we should just skip geometry shaders entirely since they are highly discouraged/deprecated and their implementation on each hardware is often a mess (providing it even exists). Compute are much more relevant by now.

Ravendarke8:
I agree with dropping support of SM 2 / 3 in future, I do not agree with dropping GS support, they are incredibly useful in great many cases (including optimizations) and working perfectly fine on DX. From where I stand they were mostly discouraged by people coming from platforms that have poor or no support for them (if I can't have it, neither can you sort of sad mindset). However I agree that Compute is the most important.

cpt-max:
I finally updated the compute fork to the new MonoGame version (3.8.1). 
The [samples](https://github.com/cpt-max/MonoGame-Shader-Samples/tree/overview) have been updated to .NET 6 and the new compute Nugets (3.8.2). 
There's now also a Nuget for Android (MonoGame.Framework.Compute.Android) and some [Android Samples](https://github.com/cpt-max/MonoGame-Shader-Samples-Mobile).

A few more changes:
- Shader files that only contain shader model 2 and 3 shaders will now be compiled by MojoShader instead of ShaderConductor. Check the [Shader Migration Guide](https://github.com/cpt-max/Docs/blob/master/Migrating%20shaders%20to%20ShaderConductor.md) for more details and a way to override this.
- Added integer surface formats, which can be used for per-pixel atomic operations.
- Assigning default values to shader parameters in HLSL will now also work for OpenGL. This is limited to assigning simple values though, no expressions or identifiers.
- Pixel shaders can now also write to buffers and textures (UAV) like compute shaders. This is DirectX only for now. With OpenGL that's possible too, but a bit more complicated.

MrScautHD:
Does this support Windows, Linux, Mac and Android + IOS?

Beyley:
> Does this support Windows, Linux, Mac and Android + IOS?

read up in the thread

MrScautHD:
> > Does this support Windows, Linux, Mac and Android + IOS?
> 
> read up in the thread

i not saw IOS, so does it not?

cpt-max:
There's this link from my first post: [NuGet Packages, Platform Support and Build Requirements](https://github.com/cpt-max/Docs/blob/master/Build%20Requirements.md)

cpt-max:
> @cpt-max can you put up a sponsor button ill hit it.

I don't really want to do that. Once I start taking peoples money, I'll feel obligated to deliver something in return. While I have no plans to stop supporting the compute fork, it's just a spare time project for me, not business.
If you want to support MG development financially, there's a Patreon page.

cpt-max:
I published new MGCB Nugets. This fixes an issue where content wasn't built, when building from within the MGCB editor.
The new version is 3.8.2.3. This affects the Nugets that go into [dotnet-tools.json](https://github.com/cpt-max/MonoGame-Shader-Samples/blob/overview/.config/dotnet-tools.json)

cpt-max:
Solved some merge conflicts and updated to new MG version with .NET 8 support. New NuGet version is 3.8.3.

orosbogdan:
Is there any update on getting this officially merged into monogame ?

mrhelmut:
The foundation is planning to merge the move to ShaderConductor in its next goals, but the addition of other shader stages will be considered after DesktopVK and WindowsGDK, and the move NativeAOT on consoles have shipped (those are the priority at the moment).

orosbogdan:
> The foundation is planning to merge the move to ShaderConductor in its next goals, but the addition of other shader stages will be considered after DesktopVK and WindowsGDK, and the move NativeAOT on consoles have shipped (those are the priority at the moment).

I have a small question, the shader conductor github seems abandoned by microsoft, will the monogame foundation plan to support the maintainance of shaderconductor alongside monogame ?

mrhelmut:
> I have a small question, the shader conductor github seems abandoned by microsoft, will the monogame foundation plan to support the maintainance of (a fork of) shaderconductor alongside monogame ?

I likely shouldn't have mentioned ShaderConductor at all because we might not even need it.

mrhelmut:
> * Most of this code in this PR is applicable with or without ShaderConductor. Essentially this PR is to add new graphics features (4 new shader stages including compute), to allow higher shader levels and to no longer be limited by the MojoShader issues.

I've reviewed and tested this PR a while ago, I know. MojoShader needs to go.

> * ShaderConductor states that "it glues existing open source components to do the cross-compiling". The two main open source components to this are [DirectXShaderCompiler](https://github.com/Microsoft/DirectXShaderCompiler) and [SPIRV-Cross](https://github.com/KhronosGroup/SPIRV-Cross), both of which are very actively maintained. ShaderConductor just assists building and wrapping these libraries together

We really don't need ShaderConductor at all for what MonoGame needs to do. A while ago, it used to be relevant considering that we were expecting to keep DesktopGL around (so that's what we still have written on our roadmap and why I mentioned it like this, but it's more about cleaning things out than moving to another tool), but now that the future is mostly going to be about the Vulkan and DX12 implementations, we don't need anything actually, those pipeline are implemented now and running without specific tools. Mojoshader and OpenGL will just retire once the new targets are stable. We won't even need SPIRV-Cross (unless we keep OpenGL around for some reason like mobiles).

> * This way there are no breaking API changes (and this is also why I'm still not sure why this PR has been moved to the 5.0 roadmap).

Because it _has_ breaking changes in the public API. The other important thing is that we're expecting feature changes/new features to have feature-parity throughout all platforms. Who's going to implement these changes for consoles (some stages just work entirely differently there, so that makes it challenging)? Likely someone at the foundation because we don't have that many contributors with clearances for all consoles. So that makes few people, with non-unlimited time. That's mostly why this PR is in a limbo. If we were against it, we would have closed it. We could skip its implementation for consoles, but that still makes a bunch of changes to make that takes time for every private platforms.

Then comes the discussion about priorities. Adding new shader stages currently isn't a hot topic because besides the people asking for it in this thread, the vast majority of MonoGame's users don't even know what a geometry shader is. We also want to clear the current refactor work before introducing new features or breaking stuff. The current priorities are moving everything to NativeAOT and refactoring everything into the native backend (which will allow the private platforms to be less sensitive to public API changes).

I understand the frustration that some people have around this very PR (and sorry for that), and the fact that some people in the community are using this PR as the avatar of "MonoGame isn't modern" (it's part of the game, we accept those comments) but we need to stay focused on what's hot.

We'll get to this PR, and we're trying to accelerate things. That's all I can say.

squarebananas:
Thanks for the comprehensive reply, I was replying primarily to orobogdan's concern that "shader conductor github seems abandoned by Microsoft" in particular though. That's why I mentioned the "code in this PR is applicable with or without ShaderConductor" and "the ShaderConductor approach doesn't necessarily need ShaderConductor itself".

> Adding new shader stages currently isn't a hot topic because besides the people asking for it in this thread, the vast majority of MonoGame's users don't even know what a geometry shader is.

Just an observation but compute shaders do get raised in the community discord from time to time, sometimes by people considering MonoGame for their project. Also while some won't use compute shaders directly, they may still benefit indirectly through the libraries built for MonoGame which may add updated features (graphic effects, particle shaders, physics engine improvements, etc). I agree though that there are definitely other priorities first. 

> Because it has breaking changes in the public API.

Maybe I have misunderstood in that case then. In my own experience I've been able to swap directly without changing method arguments in around 10 or so games. I haven't done a full line by line review to confirm this fully though.

The approach I particularly like with this PR is the side-by-side design of not directly swapping all in one go (everything or nothing) but having the content pipeline option to use the existing system or the new experimental system on a per shader basis. It's a bit like the experimental flags in browsers where existing systems work as expected but new features can be tested/developed by anyone while retaining the option to include, rework or scrap the implementation down the line.

In any case I'm looking forward to hearing more about progress and the roadmap at the Open Hours event!

mrhelmut:
Yes, cpt-max put a lot of effort in having a "isMojoshader" path and a newer path. For what it is, this PR is good and clean, and it's really cool that it is being distributed properly too with documentation.

It's just that "it's complicated" to get it moving forward upstream due to all the platforms.

We hope that moving to the native backend will make things much smoother, with much less situations like this one (because changing the public API will have less impact on platform implementations).

PR #7535, opened 2021-07-20
Fix VSCode launch configuration for MGCB Editor on Mac

the-maverick-m:
The launch configuration on Mac currently targets the directory mgcb-editor-mac.app, where the actual directory of the program is in MGCB Editor.app. This is a quick fix in order to target the program's actual directory.


harry-cpp:
Ah, good spot, thanks @the-maverick-m !

PR #7536, opened 2021-07-21
OpenGL debug context for better error messages

cpt-max:
DesktopGL uses glDebugMessageCallback for error reporting, but doesn't use a debug context, when building in debug mode.

From https://www.khronos.org/opengl/wiki/Debug_Output:
"In Debug Contexts, debug output starts enabled. In non-debug contexts, the OpenGL implementation may not generate messages even if debug output is enabled."


mrhelmut:
Thanks!

PR #7555, opened 2021-08-28
Optimized ApplyAlpha by removing two unnecessary arithmetic operations

rds1983:
Actually maybe MathHelper.Clamp call isnt necessary either. Since in edge case, when both color and alpha are equal to 255, then the result will be 255 * 255 / 255 = 255. But I am unsure.

### DIFF
--- a/MonoGame.Framework/Graphics/DefaultColorProcessors.cs
+++ b/MonoGame.Framework/Graphics/DefaultColorProcessors.cs
@@ -6,11 +6,7 @@ namespace Microsoft.Xna.Framework.Graphics
     {
         private static byte ApplyAlpha(byte color, byte alpha)
         {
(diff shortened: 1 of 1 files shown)


rds1983:
Actually maybe MathHelper.Clamp isnt necessary either. Since in edge case, when both color and alpha are equal to 255, then the result will be 255 * 255 / 255 = 255. But I am unsure.

nkast:
You can calculate the af once per pixel if you move (alpha / 255.0f) outside the method, in the start of the loop.

rds1983:
@nkast,
True. After some thought applied that optimization too.

nkast (approved):

harry-cpp:
Looks good to me, thanks @rds1983 !

PR #7558, opened 2021-08-31
Null song parameters for MediaPlayer.Play are not allowed in XNA

james0x0A:
This PR tries to mirror xna behavior around null Songs and SongCollections. The caller should be able to catch and handle this before any internal states are changed in the player.


mrhelmut:
Safe enough, merging, thanks.

PR #7561, opened 2021-09-06
Multisampling on Android

j0nat:
MonoGameAndriodGameView checks if PreferMultiSampling is true and enables 4x multisampling.

It's not perfect but seems better than not supporting it at all.

Tested on emulator and my own phone.


mrhelmut:
This can't harm since the configuration support is tested along the way and skipped if not supported.

Issue #7562, opened 2021-09-09
XACT Cue.IsPlaying Behaviour When Paused Doesn't Match XNA


squarebananas:
For XNA projects when a Xact Cue is paused both IsPlaying and IsPaused are set to true. MonoGame currently doesn't match this behaviour and sets IsPlaying to false. In particular this can cause audio issues when porting XNA games to MonoGame.

For demonstration purposes please see the CueIsPlaying project for both XNA and MonoGame here:
[https://github.com/squarebananas/MonoGameSamplesForIssues](https://github.com/squarebananas/MonoGameSamplesForIssues)

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL, WindowsDX

PR #7563, opened 2021-09-09
Fixes XACT Cue.IsPlaying Behaviour When Paused

squarebananas:
This pull request is to fix the behaviour of XACT Cue.IsPlaying when paused. More details on the issue can be found here: #7562

For demonstration purposes please see the CueIsPlaying project for both XNA and Monogame here:

### DIFF
--- a/MonoGame.Framework/Audio/Xact/Cue.cs
+++ b/MonoGame.Framework/Audio/Xact/Cue.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Xna.Framework.Audio
             get 
             {
                 if (_curSound != null)
-                    return _curSound.Playing;
+                    return _curSound.Playing || _curSound.IsPaused;
 
                 return false;
             }
--- a/Tests/Framework/Audio/XactTest.cs
+++ b/Tests/Framework/Audio/XactTest.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.IO;
+using System.Threading;
 using Microsoft.Xna.Framework.Audio;
 using NUnit.Framework;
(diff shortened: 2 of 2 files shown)


tomspilman:
@squarebananas could our unit test for XACT be fixed to verify this works correctly?

https://github.com/MonoGame/MonoGame/blob/develop/Tests/Framework/Audio/XactTest.cs

squarebananas:
> 
> @squarebananas could our unit test for XACT be fixed to verify this works correctly?
> 
> https://github.com/MonoGame/MonoGame/blob/develop/Tests/Framework/Audio/XactTest.cs


I'll look into this more, but just quickly glancing at the SoundBankPlayCue test there is the following line:
`// TODO: Add actual playback tests!`

That might mean there are quite a few tests to be added here.

squarebananas:
I've added play/pause/resume/stop tests within the existing SoundBankGetCue and SoundBankPlayCue tests.

The SoundBankGetCue additions ensures both XNA and MonoGame are working the same.

However with SoundBankPlayCue XNA is purely fire and forget so access to the cue is lost once started. However MonoGame provides access through AudioEngine.ActiveCues, so these additions are for MonoGame only.

Also SleepWhileAudioEngineUpdates has been added to allow the audio engine to be updated before the values are tested. This seems to only be required for XNA as some requests are not processed until an audio engine update has occurred, whereas MonoGame updates immediately.

There are a number of additional tests which could be added (such as looping / non looping operation, adpcm playback, etc.) but I'll leave those for a future PR.

Jjagg:
Thanks @squarebananas! :)

Issue #5662, opened 2017-04-19
XACT crashes with a SharpDX Exception when ADPCM compression is enabled


Danthekilla:
When using XACT with monogame I expected to not be able to use xWMA (sadly) and I can live with the massive files of ADPCM for the most part with disk streaming. But when I change from PCM to ADPCM my game crashes when trying to play a Cue. With the following exception.

```
{SharpDX.SharpDXException: HRESULT: [0x88960001], Module: [SharpDX.XAudio2], ApiCode: [XAUDIO2_E_INVALID_CALL/InvalidCall], Message: Unknown
   at SharpDX.XAudio2.XAudio2.CreateSourceVoice_(SourceVoice sourceVoiceOut, IntPtr sourceFormatRef, VoiceFlags flags, Single maxFrequencyRatio, IntPtr callbackRef, Nullable`1 sendListRef, Nullable`1 effectChainRef)
   at SharpDX.XAudio2.SourceVoice.CreateSourceVoice(XAudio2 device, WaveFormat sourceFormat, VoiceFlags flags, Single maxFrequencyRatio, IntPtr callback, EffectDescriptor[] effectDescriptors)
   at SharpDX.XAudio2.SourceVoice..ctor(XAudio2 device, WaveFormat sourceFormat, VoiceFlags flags, Single maxFrequencyRatio)
   at Microsoft.Xna.Framework.Audio.SoundEffect.PlatformSetupInstance(SoundEffectInstance inst)}
```

My platform is UWP, running on the latest windows 10, fast ring.

I would expect this to be supported, and unlike xWMA it doesn't throw a not supported exception. So I assume it is either a bug or I am doing something wrong somehow.

tomspilman:
Yeah this should work... in particular on Windows.

SeriousMartin:
I ran into the same problem, updating to build 1041 from 362.

Might have something to do with the state of the device.

In addition I wonder whether this has something to do with the SamplesPerBlock value that is not between 32 and 512 as stated here: https://msdn.microsoft.com/en-us/library/microsoft.directx_sdk.xaudio2.adpcmwaveformat(v=vs.85).aspx

Here is a an example of a values that go into the SourceVoice constructor when the crash happens

![adpcmformat](https://user-images.githubusercontent.com/9325330/32597051-b94ea942-c535-11e7-898e-2a42ba6cf450.png)

Any suggestions on this one?

nkast:
@SeriousMartin 
[...]

PR #7564, opened 2021-09-09
Fixes XACT ADPCM Compression Playback

squarebananas:
This pull request is to fix XACT ADPCM compression playback. More details on the issue can be found here (the issue also includes Non-XACT playback which remains broken until FFMPEG is fixed, this only fixes XACT): #5662

The fix was originally posted by @tomspilman in 2018 here:
[https://github.com/MonoGame/MonoGame/issues/5662#issuecomment-410486541](https://github.com/MonoGame/MonoGame/issues/5662#issuecomment-410486541)

The issue relates to XAudio projects, however the same fix is required for OpenAL projects as well.

For demonstration purposes please see the AdpcmAlignment project within here:

### DIFF
--- a/MonoGame.Framework/Platform/Audio/SoundEffect.OpenAL.cs
+++ b/MonoGame.Framework/Platform/Audio/SoundEffect.OpenAL.cs
@@ -159,7 +159,7 @@ namespace Microsoft.Xna.Framework.Audio
         {
             if (codec == MiniFormatTag.Adpcm)
             {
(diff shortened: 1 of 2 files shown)


tomspilman:
Thanks for this PRs @squarebananas .

@mrhelmut i've had to do a fix like this before on some projects.  So i suspect it is correct.

mrhelmut:
Yes, this seems relevant to me! The sample has been very welcome. I've been having crashes with XACT related to this.

Issue #7565, opened 2021-09-09
DynamicSoundEffectInstance Requests Buffers Too Late


squarebananas:
When streaming sounds the TargetPendingBufferCount for a DynamicSoundEffectInstance is hard-coded to 3 and the PendingBufferCount should generally match this.

However the PendingBufferCount is never seen to be above 1. If the PendingBufferCount reaches 0 then all buffers were played before more data was submitted. This results in a choppy audio glitch.

This is currently occurring because the PendingBufferCount is being calculated incorrectly.

For demonstration purposes please see the OpenALPendingBufferCount project within this project here:
[https://github.com/squarebananas/MonoGameSamplesForIssues](https://github.com/squarebananas/MonoGameSamplesForIssues)

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

PR #7566, opened 2021-09-09
Fixes DynamicSoundEffectInstance Requesting Buffers Too Late

squarebananas:
This pull request is to fix DynamicSoundEffectInstance requesting buffers too late. More details on the issue can be found here: #7565

For demonstration purposes please see the OpenALPendingBufferCount project within here:

### DIFF
--- a/MonoGame.Framework/Audio/DynamicSoundEffectInstance.cs
+++ b/MonoGame.Framework/Audio/DynamicSoundEffectInstance.cs
@@ -284,7 +284,7 @@ namespace Microsoft.Xna.Framework.Audio
         private void CheckBufferCount()
         {
             if ((PendingBufferCount < TargetPendingBufferCount) && (_state == SoundState.Playing))
-                _buffersNeeded++;
+                _buffersNeeded = TargetPendingBufferCount - PendingBufferCount;
         }
 
         internal void UpdateQueue()


squarebananas:
I'll need to look at this again another time as the tests don't quite look correct.

According to this:
https://docs.microsoft.com/en-us/previous-versions/windows/xna/ff433726(v=xnagamestudio.40)
> The BufferNeeded event is raised in the following situations:
> - During DynamicSoundEffectInstance playback when the pending buffer count is less then or equal to 2.
> - Every time the DynamicSoundEffectInstance instance completes playing a buffer, and updates the pending buffer count.
> - When a DynamicSoundEffectInstance instance transitions from the stopped to playing state, and the pending buffer count is less than or equal to 2.

However some of the tests seem to be checking that only a single BufferNeeded event is raised per buffer used or play call. In other words only points 2 & 3 of the above, not point 1. This doesn't seem to match XNA which keeps raising the BufferNeeded event until 3 buffers are queued.

nkast:
Point 1 makes me think that the check/event happens on every tick. 
However for point 3, mg implementation sets a flag during PlatformPlay and the events is invoked during the gameloop tick.
If we were to add a queue check in update it make the check on PlatformPlay irrelevant. The documentation explicitly states 3 as separate case which makes me think that XNA would invoke the event directly from PlatformPlay(). Probably the same is true for point 2.

squarebananas:
Yeah, during normal playback some of the points seem to be asking for the same thing.

However I think the points begin to be different when SubmitBuffer is called outside of BufferNeeded and when no SubmitBuffer occurs during a BufferNeeded event.

My next step is to create a table of XNA results with the combinations of SubmitBuffer before playback, SubmitBuffer outside of BufferNeeded and no SubmitBuffer during BufferNeeded.

I'm hoping to do that in the coming weeks and revise the PR. For now the PR will fix the choppy audio glitch but shouldn't be committed yet as it will cause tests to fail.

nkast:
Something that seems to work is to make _bufferNeeded boolean and call the event once per frame.

```
  internal void UpdateQueue()
        {
            // Update the buffers
            _dstrategy.DynamicPlatformUpdateBufferQueue();

            // Raise the event
            if (_bufferNeeded || (PendingBufferCount < TargetPendingBufferCount))
            {
               var bufferNeededHandler = BufferNeeded;
               if (bufferNeededHandler != null)
               {
                   bufferNeededHandler(this, EventArgs.Empty);
               }
               _bufferNeeded = false;
            }            
        }
```

squarebananas:
I did quickly try the above, however I'm still getting failed tests with that.

For some reason I thought I had tried the sample project with XNA, however I must have only compared it to the XNA docs. Point 1 is not being done by XNA and it will not automatically queue up 3 buffers (even though it's stated in the docs). Both XNA and MonoGame have the audio glitch when submitting buffers solely on the BufferNeeded event. Pre-submitting buffers before playback can fix this, but the documentation makes no mention of this as a requirement. Sorry for the confusion there.

Also I noticed with the MonoGame documentation (https://docs.monogame.net/api/Microsoft.Xna.Framework.Audio.DynamicSoundEffectInstance.html) that the BufferNeeded event should be called "when the number of queued audio buffers is less than or equal to 2". Again this matches the XNA documentation but hasn't been implemented in this way.

So going forward this raises the question should MonoGame match the documentation and the tests be revised or match XNA behaviour and the documentation be revised?

Another thing I noted just is in the tests the Thread.Sleep calls within the SleepWhileDispatching loop can take longer than expected and this can cause intermittent fails. Adding StopWatch to exit the loop early seemed to fix this issue.

nkast:
XNA implementation isn't that good. for example State isn't updated immediately after Play/Pause/Stop.
I bet what is described in the documentation is more correct behavior and what was their their requirements/plan.
Today I also found this https://github.com/nkast/XNAGameStudio/blob/master/Documents/DynamicAudio_Article_4_0.docx
There's also this comment [here](https://github.com/MonoGame/MonoGame/blob/a9e5ae6befc40d7c86320ffdcfcd9d9b66f786a8/MonoGame.Framework/Audio/DynamicSoundEffectInstanceManager.cs#L44) "XNA posts DynamicSoundEffectInstance.BufferNeeded events always on the main thread."
I 'll try a couple more things. I will bring back the eventNeeded counter, but I have removed the code entirely from Play().

squarebananas:
Great. Nice find on the article. I need to read it some more, but there are a couple of points of interest.

```
// Start the sound, no currently pending buffers but sound will raise
// the BufferNeeded event giving us a chance to submit buffers for playback.
```

This comment implies buffers (multiple) should be sent on a BufferNeeded event, not just one.

```
void DynamicSoundInstance_BufferNeeded(object sender, EventArgs e)
{
    // Submitting the larger buffers as two smaller buffers ensures that
    // the event is raised before the buffers run out, avoiding glitches
    dynamicSoundInstance.SubmitBuffer(buffer, 0, buffer.Length/2);
    dynamicSoundInstance.SubmitBuffer(buffer, buffer.Length/2, buffer.Length/2);
}
```

This example is also submitting two buffers to "avoid glitches".

That implies it's probably working as expected and a single buffer per event will cause glitches, which is what is occurring in the sample project.

I guess the question now is should this be improved with MonoGame automatically calling BufferNeeded to properly maintain the TargetPendingBufferCount or should the documentation be improved to state that multiple buffers are required per BufferNeeded event.

nkast:
by looking at the test i see that xna fired event only on buffer completion. I prefer the documented behavior better. 
The implementation i currently have works fine. I might add a boolean and set it at Play(), then on update, and only if the user hasnt submited more than 2 buffers ,i will call the event teo additional times. just to avoid initial glitses.

nkast:
By the way, thank you for the test project. It  helped a lot.

nkast:
this will halt if the user write the following
```
        private void DynamicSoundEffect_BufferNeeded(object sender, EventArgs e)
        {
            //TODO: tomorrow
        }
```

This is what i ended up with. Will probably break the tests.
```
        internal void Update()
        {
            lock (AudioService.SyncHandle)
            {
                // Update the buffers
                DynamicPlatformUpdate();

                if (_beginPlay)
                    _buffersNeeded = Math.Max(_buffersNeeded, TargetPendingBufferCount - 1 - PendingBufferCount);

                // Raise the event
                var bufferNeededHandler = BufferNeeded;
                if (bufferNeededHandler != null)
                {
                    // raise the event for each processed buffer
                    while(_buffersNeeded-- != 0)
                        bufferNeededHandler(this, EventArgs.Empty);

                    if (State == SoundState.Playing && PendingBufferCount < TargetPendingBufferCount)
                        bufferNeededHandler(this, EventArgs.Empty);
                }

                _beginPlay = true;
                _buffersNeeded = 0;
            }
        }

```

squarebananas:
> the while loop will halt if the user write the following, unless you add a limit check

If you were referring to the while loop in my code that won't happen as it already breaks the loop if the PendingBufferCount doesn't increase.

With the code you posted I think it will fail the tests as it won't stop raising the event after SubmitBuffer isn't called. According to the document you linked:

> Once the buffer count reaches 0, DynamicSoundEffectInstance assumes that the client is not interested in immediately submitting more buffers and stops raising this event. Of course, if the game queues more buffers, the event is raised again as the queued buffers are consumed.

The tests seem to be based around this, although this seems to go against point 1 of the XNA api (https://docs.microsoft.com/en-us/previous-versions/windows/xna/ff433726(v=xnagamestudio.40)) where events keep occurring in playback. The tests could be changed though if it was intended to keep raising events regardless of whether SubmitBuffer occurs. I'm not sure if this is desirable or not though.

Just with the code I posted, to make it clearer I've reordered it and commented although it does exactly the same as before

```
if (_buffersNeeded >= 1)
{
    // At least one event is needed after the initial play call and per processed buffer
    int minimumNoOfBufferNeededEvents = (_buffersNeeded < 3) ? _buffersNeeded : 3;

    // Multiple events may be raised to reach the target buffer count so long as a SubmitBuffer occurs for each event
    int noOfBuffersNeededEventsToReachTarget = TargetPendingBufferCount - PendingBufferCount;

    for (int i = 0; i < noOfBuffersNeededEventsToReachTarget; i++)
    {
        // Raise the BufferNeeded event and check if SubmitBuffer occurred
        int lastPendingBufferCount = PendingBufferCount;
        bufferNeededHandler(this, EventArgs.Empty);
        bool submitBufferOccurred = PendingBufferCount > lastPendingBufferCount;

        // Stop raising events if no SubmitBuffer occurred during BufferNeeded and the minimum number of events has already been raised
        if (!submitBufferOccurred && ((i + 1) >= minimumNoOfBufferNeededEvents))
            break;
    }
}
```

SimonDarksideJ:
Do you feel this is ready for prime time @squarebananas ?

squarebananas:
To save you running the demo, here is proof from before and after (from repo cloned today).

https://github.com/user-attachments/assets/aa52719e-f81e-4ebb-9abe-2fa444b0f4c9

https://github.com/user-attachments/assets/a7884136-5388-4651-81b0-d60bb3fbb6f1

Slight issue with the unit tests though (these don't appear to be run on GitHub, so fail locally).

image

The current expected values are based on producing the same glitchy audio as XNA. I can change them but then anyone running the tests against XNA will probably be confused.

As discussed in the posts above, it comes down to a decision on API spec compliance vs XNA accuracy.

ThomasFOG:
> As discussed in the posts above, it comes down to a decision on API spec compliance vs XNA accuracy.

@tomspilman you're likely the most fitted to decide when it comes to sound.

tomspilman:
> @tomspilman you're likely the most fitted to decide when it comes to sound.

I trust @squarebananas logic here...  i don't fully follow the fix, but it seems reasonable.

If tests prove it works... then it works.

tomspilman (approved):

Issue #7567, opened 2021-09-09
EffectParameter SetValue Behaviour Using Int For Float Parameters Doesn't Match XNA


squarebananas:
For XNA projects when using EffectParameter SetValue it is valid to set a float parameter with an int value. However in MonoGame this throws an invalid cast exception. This can cause issues when porting XNA games to MonoGame.

For demonstration purposes place the following code into a XNA and MonoGame project (a cast error will occur in the MonoGame project).

```
BasicEffect basicEffect = new BasicEffect(GraphicsDevice);
basicEffect.Parameters["SpecularPower"].SetValue(1);
```

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL, WindowsDX

PR #7568, opened 2021-09-09
Fixes EffectParameter SetValue Using Int For Float Parameters

squarebananas:
This pull request is to fix the behaviour of EffectParameter SetValue to allow float parameters to be set with int values. More details on the issue can be found here: #7567


tomspilman:
Thanks again for these @squarebananas.

Note in the future you don't really need to create an issue and a PR both.  It is fine to include the details in the PR directly.

squarebananas:
> 
> 
> Thanks again for these @squarebananas.
> 
> Note in the future you don't really need to create an issue and a PR both. It is fine to include the details in the PR directly.

No problem, I'll just submit a PR for the next requests I'm doing. Thanks.

Jjagg:
Thanks @squarebananas!

PR #7569, opened 2021-09-09
Allow Texture2D.GetData On Non UI Threads

squarebananas:
Currently Texture2D.GetData can only be run from the UI thread for DesktopGL projects. WindowsDX / UWP projects allow Texture2D.GetData from other threads.

This pull request uses the BlockOnUIThread approach (already used for Texture construction, Buffer GetData, etc) to add this functionality.

This code change is just swapping EnsureUIThread to the following, but the PR shows 50 additions and 49 deletions due to the indent change:
```
Threading.BlockOnUIThread(() =>
{
    // Existing Code
}
```

For demonstration purposes please see the TextureGetDataThread project within here:
https://github.com/squarebananas/MonoGameSamplesForIssues


Jjagg:
The reason this throws on non-UI threads, is because people generally expect `GetData` to fill the array immediately. They might try to use the data immediately after the `GetData` call. That said, people of course might sometimes want to use GetData from a non-UI thread. But they should be made aware of the delay in that case.

I think maybe a better approach would be to:

- Add a `GetDataUnsafe` (not sure that's the best name for it) method that uses BlockOnUIThread.
- Keep EnsureUIThread in the normal GetData method, but have a parameter for custom error text. In this case letting users know about the OpenGL limitation and that they should use GetDataUnsafe to get data before the next frame, but not immediately.

We should look into inconsistencies with the buffer GetData methods; they should act in the same way.

squarebananas:
Just to understand the issue more, the concern is that Texture2D.GetData will take up to 1 frame to complete?

In particular I was loading a heightmap in a level loading thread and needed GetData for the terrain construction. This had always worked fine in XNA, Monogame WindowsDX and Monogame UWP. It appeared BlockOnUIThread had been missed here for DesktopGL projects when compared to other GetData methods which is why I submitted the PR.

Should the other GetData methods not have used the BlockOnUIThread approach?

Jjagg:
> Just to understand the issue more, the concern is that Texture2D.GetData will take up to 1 frame to complete?

Yes, exactly.

> Should the other GetData methods not have used the BlockOnUIThread approach?

In my opinion not. Anecdotally, I mostly see GetData used to fill an array that's used immediately after. Code like that will not work on OpenGL with BlockOnUIThread.

> In particular I was loading a heightmap in a level loading thread and needed GetData for the terrain construction.

Yeah, this is exactly the kind of case for which it would be useful to still allow calling GetData on OpenGL off the UI thread.

I think EnsureUIThread with an alternative method that schedules the call on the UI thread seems like a good solution, but I'm not sure that's the best design, so alternative suggestions are welcome.

cc @harry-cpp @mrhelmut

mrhelmut:
I'm not a fan of BlockOnUIThread in this context but I must admit that EnsureOnUIThread has a misleading behavior because it postpone the job to the next frame.

In my opinion, both are problematic, and I don't think there is a golden path because either mechanism will trigger different use-cases. Ultimately, I don't think that we want to block, because that would be very visible to players and unavoidable to developers, while postponing to the next frame is workable (though tricky to developers and currently undocumented).

Dunno if we could rework the OpenGL threading stuff, it's really misleading as it is but as far as I gave thoughts to it, I couldn't figure a better model than postponing to the next frame (and manually considering it in code). The biggest issue is that this behavior is undocumented and problems such as this one (or having unload/load sequences making two textures to coexist in memory while they shouldn't) can arise.

Jjagg:
You are right. Sorry for confusing things. It seems I actually imprinted a wrong idea of BlockOnUIThread in my head 4 years ago, thinking it just scheduled the task and did not wait for it to complete (which I feel dumb about because the Block part is in the method name). Sooo, yeah the whole reason we have EnsureUIThread in the first place is [me misunderstanding BlockOnUIThread](https://github.com/MonoGame/MonoGame/commit/854ea8ae194c341bfad6e502daab9dec212094db#r21384672) 😓

So with that cleared up for me, I agree GetData should use BlockOnUIThread.

Like @mrhelmut said, it would be nice to document this behavior. I think perhaps we should have a doc page with OpenGL specifics/limitations. But this PR is great as is IMO :) 

> having unload/load sequences making two textures to coexist in memory while they shouldn't

I don't follow, when would this issue occur?

stromkos:
> WindowsDX / UWP projects allow Texture2D.GetData from other threads.

Historically, XNA had a single threaded model, therefore WindowsDX and all SharpDX derived platforms, do as well.

With the SDL platforms, there is a separate rendering thread (by necessity on some platforms like Android). 

> Where it becomes noticeable is when GetData is run for a few hundred textures on a thread as it will add approximately 1 second to the thread's completion time for every 60 textures at 60 fps.

The given approximation depends on the UI thread runtime vs the game thread runtime. Not to mention that a one second delay on the main thread in Windows will cause the close window due to unresponsive task box to appear if any click or key-press is made while the main thread is waiting.

There are too many XNA tutorials offering per pixel sprite collision detection that work fine when single-threaded, but quietly and silently break (slow down to a crawl) with this change. 

The current behavior forces the programmer to consider the impact and react somewhat appropriately. Even moving the calls to `Draw()`, there is still no guarantee of data consistency across threads upon access from `Update()`. 

---

What if any effect does `IsRunningSlowly` have on a separate thread? Could this create a race condition?

---

I would suggest in 4.0, requiring all `GetData()` and `SetData()` calls must be batched.`;
Thus allowing for a single wait period for the threads to synchronize.

cc @harry-cpp @mrhelmut @Jjagg

damian-666:
the timers resolution on andriod is so bad i stopped use locks or sleep or waithandles entirely to sync physics , collisoin procossing AI and whatever, with the draw calls.   its unusable for me but was fine on windows.   so i change to a 2 frame buffer setup..  i suggest the other wayh around, whatever u need from UI thread , you can make a copy of, then access that copy and orphan the last good cpy ref on next draw update which can be sync with the refresh.    whe u are done with it it gets GCd..     its not suitable for everythig but what unseen stuff happens between frames i might not matter adn if u have any continous  collision methods it might be ok.     also  u can save a queue of  them and get a free replay feature.      i think isrunning slowly is only useful if u arent using background thread , and ur own timers and game loop.

 Pixel[] lastDrawnTexture  is simply changing a reference atomically  and no sync is needed.

stromkos:
@damian-666 
I must say that I am having trouble translating your "textese", forgive me I am old.  I get the double buffering, but you loose me on the buffer reuse factor.  With true double buffering, the existing memory allocation is reused for the next frame and not discarded for the GC especially in memory constrained environments.  

My question about `IsRunningSlowly` was on how it is implemented in multi-threaded environments.  In XNA it prevents `Draw()` from being called. This is unnecessary in a multi-threaded environment, so most likely nothing happens other than the flag being set, but if it did wait on the UI thread, this would cause a race condition.

damian-666:
If  u call getdata  on  texture think it's a fresh buffer so u cant  reuse
it ,   i don't think..maybe there's a way if u are getting the texture from
a large render frame, by setting the render buffer w unmanaged pointer ,
and if  u don't want it GCd and it's might be shown in with  one in the WPF
monogame samples of integration.   eirher way i think making a copy.is goin
g to be faster than any thread sync attempt.  sorry for the sms speak i had
Nokia phones way too long can't write normally.  I think is running slowly
might skip draw calls and if u don't do anyting in update it's never gonna
be set or it doesn't matter. i put my own timers and settimerresulution to
1 Ms or it's 16ms to stabilize frame rate.   But on Android i just let.it
run as fast as it is can and i can't measure time at a meaninful resolution
do i don't its slow enough i just let it do whatever for now.   Anyways i
spend weeks  sorting this stuff out,  hope some of the ideas saves u some
time

squarebananas:
> Historically, XNA had a single threaded model, therefore WindowsDX and all SharpDX derived platforms, do as well.

In my experience XNA allowed GetData requests without issue from other threads (I used this extensively on Xbox 360 games).

Just to make it clear as stated above other GetData functions work on DesktopGL platforms (such as VertexBuffer.GetData), it's only Texture2D.GetData that has been missed.

This PR just adds the missing BlockOnUIThread approach already in use through out the DesktopGL project.

> The given approximation depends on the UI thread runtime vs the game thread runtime. Not to mention that a one second delay on the main thread in Windows will cause the close window due to unresponsive task box to appear if any click or key-press is made while the main thread is waiting.

The UI thread will continue to run its updates each frame like normal while the thread is active. In the example the level loading thread will just take longer to complete (as stated one frame per GetData request longer). This already happens when loading textures in a loading thread and displaying a game loading animation on the UI thread.

> What if any effect does IsRunningSlowly have on a separate thread?

IsRunningSlowly is based on the concept of a frame drop occuring on the UI thread. Other threads just do the work they are given regardless of whether the UI thread is updating/drawing so IsRunningSlowly isn't really related to that. (Also IsRunningSlowly currently doesn't work correctly #7571).

stromkos:
Would it not be better to synchronize the game and UI threads to replicate the XNA single threaded model, with a proper implementation of `IsRunningSlowly`?

The goal of MonoGame versions < 4  is unity or at least parity with XNA.

stromkos:
> In my experience XNA allowed GetData requests without issue from other threads (I used this extensively on Xbox 360 games).
Yes,  the main game thread was blocked on the call, as your change purposes. 

The difference is in `XNA`  all of your calls were attended to in a long single frame (other thread delay + game loop time + execution time)

With your change is that each and every call blocks the continuation of the Game thread **for each call made**.  

My solutions:  Either provide a batch calling method(4.0) or bind the threads together, make the UI thread wait on the game thread(Update) each step, like XNA, making locking and waiting unnecessary.

squarebananas:
The purpose of the PR was just to add the missing BlockOnUIThread which is already used in 22 other similar situations. Currently without the PR it throws an exception.

There may be better approaches going forward but whatever changes are made would ideally need to be made to the other 22 instances of BlockOnUIThread. Presumably that would be an entirely separate issue/PR.

Also the Game thread shouldn't be blocked, only additionally created background threads (for example level loading threads). For DesktopGL the UI thread and Game thread are the same (which is why dragging the title bar freezes the game) unless I'm missing something?

stromkos:
> Also the Game thread shouldn't be blocked...

This depends on usage, multiple direct calls from `Game1` as shown in tutorials for XNA, some of which either directly state or imply no performance penalty(as is the case in a single threaded model), would lead to a lock release delay cycle(s) in the multi-threaded model, thereby possibly degrading the main and/or rendering threads' performance.


> unless I'm missing something?

No, I was not objecting to the PR, only suggesting possible better ways moving forward. Which I agree, that it needs to be in a separate PR. I was simply here commenting to plant the seed for later changes needed for all 23 instances in version 4, since the suggestions would break XNA compatibility in the API.

damian-666:
if you are making  a heightmap (2d image to 3d data based on color i
presume) and need to process it on after getData in batch on CPU,
could u consider
looking at the alternative like the SIMD optimized GPU image processor
ImageSharp, for loading and /or processing those assets?  heres an example
in  Silk.Net,

https://github.com/dotnet/Silk.NET/blob/main/examples/CSharp/OpenGL%20Tutorials/Tutorial%201.4%20-%20Textures/Texture.cs


if that link doesn't work: hreres the last update a few days ago
https://github.com/dotnet/Silk.NET/commit/a352fdb42c5f74bd3960094eff598f22e1908885


i feel its best to avoid the the render/ gpu thread for this because of
the syncing madness already mentioned and the cost of passing data from GPU
to cpu and back, ( if that both  applies here, as in you are using
rendertarget,.. but f just loading raster files, idk the content internals
.   Even if you are usng render targets, again Image sharp might offer a
safer and high performance alternative.   then just set the 3d data and
draw w Mg on one thread hopefully producer consumer like , lockless.

as an aside core devs should montior SILK..ive been blown away by the flux
in recent days..
they dont care about consoles, but there is a TON of cross over n
Silk..brilliant devs on it and not all on MSFT staff hte lead code reviewer
is 18 year old student from Germany, 100's of brillaint people working on
this and and .net its overwhelming... some might never come to fruition but
the net 7 looks like another big leap ahead liine Netcore 6 was.


As an aside..
 but im starting to move toward SIMD and away from gpgpu in my plans, and
.net 7 might  make make JIT live to its promise, and with hopefully more
generalized Vectorization and HW intrinsics to be much easier.   but, i
have montored ImageSharp for 3 years and they already did as best they
could its released and stable.  Like Bepu argues, the GPU drivers are
heavily nuanced, best used just for rendering and graphics,  maybe 10x
faster but not worth the trouble, and the  that SIMD code in Imageshape (
and Bebu) code is hideous but fast, and you dont have to worry about it if
it works for your case.    i just will wait for net 7 before vectorizing my
own stuff..



using Silk.NET.OpenGL;
using System;
using SixLabors.ImageSharp;
using SixLabors.ImageSharp.PixelFormats;

namespace Tutorial
{
    public class Texture : IDisposable
    {
        private uint _handle;
        private GL _gl;

        public unsafe Texture(GL gl, string path)
        {
            _gl = gl;

            _handle = _gl.GenTexture();
            Bind();

            //Loading an image using imagesharp.
            using (var img = Image.Load<Rgba32>(path))
            {
                //Reserve enough memory from the gpu for the whole image
                gl.TexImage2D(TextureTarget.Texture2D, 0,
InternalFormat.Rgba8, (uint) img.Width, (uint) img.Height, 0,
PixelFormat.Rgba, PixelType.UnsignedByte, null);

                img.ProcessPixelRows(accessor =>
                {
                    //ImageSharp 2 does not store images in contiguous
memory by default, so we must send the image row by row
                    for (int y = 0; y < accessor.Height; y++)
                    {
                        fixed (void* data = accessor.GetRowSpan(y))
                        {
                            //Loading the actual image.
                            gl.TexSubImage2D(TextureTarget.Texture2D, 0, 0,
y, (uint) accessor.Width, 1, PixelFormat.Rgba, PixelType.UnsignedByte,
data);
                        }
                    }
                });
            }

            SetParameters();

        }

mrhelmut:
> I would suggest in 4.0, requiring all `GetData()` and `SetData()` calls must be batched.`;

Considering newer APIs like Vulkan, D3D12, and consoles... this is likely what is going to happen. I don't see ourselves trying to go into complex command synchronisation stuff just for those operations (which would totally bottleneck such pipelines). Those API are not meant to wait on GPU operations to complete, and it is encouraged to have everything operating asynchronously. OpenGL is a different beast, it is meant to be entirely sequential and blocking, and is not friendly with threading at all (which is why those blocking were introduced in the first place). OpenGL is meant to disappear and at some point we will redirect DesktopGL users to the future DesktopVK implementation, which will likely assume GetData/SetData to be async.

I also believe that we should discourage the usage of those operations. They are way too often (unknowingly) abused for operations that don't need them or for problems which could be solved differently and more efficiently. There are use cases where they are useful and I'm not saying that we should remove them, but we should update the documentation to explain the nature of those operations and to think twice or thrice before using them. Maybe we should output a warning whenever they are used in debug builds.

We didn't even implemented them on consoles, and people seem okay with that because they end up finding better solutions in their absence.

I believe that we should be fine with blocking. That will expose to users how those operations can be an issue and will discourage them to use them. We could add a console output warning when running from a thread like "this operation is blocking the UI thread".

squarebananas:
> We could add a console output warning when running from a thread like "this operation is blocking the UI thread".

I could amend the PR to add console output warnings for all BlockOnUIThread calls although it's the non UI thread that is blocked not the UI thread itself.

Should I add these lines?
https://github.com/MonoGame/MonoGame/blob/158c0154ac18ed6102c65e24665c6a080ccb8ed2/MonoGame.Framework/Platform/Threading.cs
Line 141 `Debug.WriteLine("Thread blocked, waiting for action to be completed by the UI thread");`
Line 199 ` Debug.WriteLine("Pending action completed by the UI thread");`

mrhelmut:
Sounds good to me.

squarebananas:
> Sounds good to me.

Thanks, I've added those just.

MonoGame.Framework/Platform/Threading.cs
```diff
@@ -139,6 +139,7 @@ internal static void BlockOnUIThread<TState>(Action<TState> action, TState state
 
             try
             {
+                Debug.WriteLine("Thread blocked, waiting for action to be completed by the UI thread");
```
dellis1972, line 142: These will be stripped out for release builds won't they? Which means someone using the NuGet packages will not see these messages. `Trace.WriteLine` might be a better option.

tomspilman: Not sure what is the best option here, but these are our concern:

1. On release/shipping builds we don't want to slow down the game code with lots of debug checks that should not happen in a shipping game.

2. Since NuGet is our only delivery option how do we add debug/diagnostic checks that aid in development, but doesn't hurt when you ship?  Are there Debug NuGet packages?

Note this may be a bigger discussion beyond the scope of this PR.  But for games we cannot waste even a little time on debug checks that 99.99% never fire in a shipping game.

squarebananas: After a quick check just there are around ~25 other Debug.WriteLine instances with potentially useful information.

Maybe they should all go to a new MonoGame method where all such output messages can be rerouted at runtime. For Debug this could always route to Debug.WriteLine as default and for Release somewhere else only if Debugger.IsAttached is true.

User defined Streams could also be attached to get these messages through a custom Console app, localhost webpage, etc.

In addition by adding unique error codes for each type of message this could be used for custom filtering and disabling of error checking/reporting by error code or error category (audio, input, graphics, video errors).

Mindfulplays: [`Debug.WriteLine`](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Debug.cs,148) is fully optimized away ([including the interpolated string](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQAwAIpwHQBECWAhgOYB2A9gM4Au+AxpQNzLJQDMmM6AwugN7J0QzBygAWdAFkAFAEp+g4Uvylq6AB7oAvOgCszJEqUB1AE75qAUwAyKy9IAkAIgASlgDbvy/dWF0BfJ1lGdAB6UPQAFQALfEp0dzt0OPRyAAdaAFt8AC9LYHRCAHdCAE9sRSF/SvQagG1uclJgC3wmwndpJ0iAUQBlSIBJADkAcSCAXTCI4fI1YEsAMzsCwlJSoujLU0sKw2E08wA3QitMCTMLGztpLAxKeRqBJABifEX0XB6AIQBVUZqwiwAE5pA8DC9LM13koatUkP4gA===)) in non-DEBUG builds.

I added a [`GraphicsDebug.Blah`](https://github.com/Mindfulplays/MonoGame/blob/ios-metal/MonoGame.Framework/Platform/Graphics/GraphicsDebug.Default.cs#L115) in my private repo to streamline these. Maybe I can bring these in a separate PR?

stromkos: >     1. On release/shipping builds we don't want to slow down the game code with lots of debug checks that should not happen in a shipping game.

Use Assert instead of "try catch".

SimonDarksideJ:
What is the status of this PR @squarebananas ?  are we close to a resolution of the comments above?

squarebananas:
> What is the status of this PR @squarebananas ?

Just did a sync and it still fixes the issue with the linked sample project. This 3 year old PR is essentially 3 lines of code to fix an issue. Two additional lines were later added at [mrhelmut](https://github.com/mrhelmut) request for debug warnings.

> are we close to a resolution of the comments above?

There are already around ~25 other Debug.WriteLine instances within the existing code in addition to the 2 new ones. Any proposals for a new debug output system should probably be discussed in a new issue and all ~27 instances resolved in a new PR. Also [tomspilman](https://github.com/tomspilman) mentioned "Note this may be a bigger discussion beyond the scope of this PR" as well.

Hope that helps.

PR #7570, opened 2021-09-09
Fixes UWP Set Window Size Thread Access Error

squarebananas:
It should be possible to change the window size during Update by setting the preferred back buffer size and using ApplyChanges. However this results in a thread access error as Windows.Graphics.Display.DisplayInformation.RawPixelsPerViewPixel can only be accessed from the UI thread (different from the thread running the Game Tick).

For demonstration purposes please see the UWPSetWindowSize project within here:
https://github.com/squarebananas/MonoGameSamplesForIssues

This pull request fixes this by using the UI thread's dispatcher to retrieve the RawPixelsPerViewPixel value.

### DIFF
--- a/MonoGame.Framework/Platform/WindowsUniversal/UAPGameWindow.cs
+++ b/MonoGame.Framework/Platform/WindowsUniversal/UAPGameWindow.cs
@@ -273,7 +273,7 @@ namespace Microsoft.Xna.Framework
                 _viewBounds.Height == height)
                 return;
 
-            double rawPixelsPerViewPixel = 1d;
+            double rawPixelsPerViewPixel = 1.0d;
             if (CoreWindow.GetForCurrentThread() != null)
                 rawPixelsPerViewPixel = _dinfo.RawPixelsPerViewPixel;
             else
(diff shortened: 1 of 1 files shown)


squarebananas:
Not sure why this is failing the checks. First time it failed on Build ubuntu. I made a minor change (1d to 1.0d) to force the checks to rerun. The second time it's failed on SoundEffectInstanceTest even though this pull request has nothing to do with sound code.

Jjagg:
@tomspilman Could you rerun the tests for this one?

mrhelmut:
@squarebananas Did you happen to test this change on both XAML and CoreApp setup of UWP? I feel like this is the kind of change which might break one while not the other.

squarebananas:
> @squarebananas Did you happen to test this change on both XAML and CoreApp setup of UWP? I feel like this is the kind of change which might break one while not the other.

I had only tested XAML but it worked with a CoreApp project just. Also it looks like this fix is only required for XAML projects from my testing just:


XAML Project
- Uses GameRunBehavior.Asynchronous
- Game tick runs on a different thread than CoreWindow
- Throws exception (PR fixes)

CoreApp Project
- Uses GameRunBehavior.Synchronous
- Game tick runs on the same thread as CoreWindow
- No exception occurs

mrhelmut:
Thank you! merging.

PR #7574, opened 2021-09-11
Add RollForward directive, so that if user has newer version of dotnet…

CartBlanche:
… installed it will still work.

More info available here - https://natemcmaster.com/blog/2019/01/09/netcore-primitives-3/


mrhelmut:
We should likely have a rollforward directive on all of the projects. Does it still apply to net6 and future releases?

CartBlanche:
@mrhelmut I've personally not had the opportunity to test if it is still required for .NET 6+, but would think/hope that MS would continue to allow it to work with newer versions.

mrhelmut:
I took a look at the documentation and this seems valid with net6 and net7 as well.
I hope that you don't mind, I applied your suggestion to all the repository with #7789 while doing other stuff, so I'll be closing this one.

CartBlanche:
As long as it proved useful in some way, I'm happy with that.

PR #7575, opened 2021-09-17
Added a keyboard shortcut (F2) to rename a file.

### LINKED ISSUE #7233: Add more hotkeys in MGCB Editor
Small quality of life improvement.
VS2019 scheme should be pretty familiar for most C# developers

"New Item" - Ctrl+Shift+A
"Existing Item" - Shift+Alt+A
"Rename" - F2

It would be nice if it could be configured

rbwhitaker:
I was investigating another bug and tried to rename it with F2, which didn't work, because that keyboard shortcut hadn't been defined. That felt like low-hanging fruit, so I went and made this change.

This partially addresses #7233. I'd add the other two shortcuts listed there, but I'm not sure how to do it. A partial fix seemed like an improvement over no fix.

Is this something the MonoGame maintainers would be interested in pulling in?

### DIFF
--- a/Tools/MonoGame.Content.Builder.Editor/MainWindow.eto.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/MainWindow.eto.cs
@@ -194,6 +194,7 @@ namespace MonoGame.Tools.Pipeline
             cmdRename = new Command();
             cmdRename.MenuText = "Rename";
             cmdRename.Image = Global.GetEtoIcon("Commands.Rename.png");
+            cmdRename.Shortcut = Keys.F2;
 
             cmdDelete = new Command();
             cmdDelete.MenuText = "Delete";


rbwhitaker:
Hm. Issue #7233 now has a comment that says, "rbwhitaker linked a pull request that will close this issue 8 minutes ago." I don't want to actually close that issue, since it is only a partial fix. Is there some other procedural thing that should be done to split that issue up or something?

Apostolique:
It's because you used the word "resolves" right before the issue link. You can see a list of the special words: <https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue>.

rbwhitaker:
Thanks, @Apostolique. I edited my description to say "partially addresses" instead of "partially resolves". I think that still conveys what I wanted to say without the extra string attached.

Jjagg:
Great! Thanks @rbwhitaker!

PR #7584, opened 2021-10-13
Implement AddDependency in PipelineImporterContext

### LINKED ISSUE #6034: Unable to add dependency in a custom pipeline importer
When we create a custom **importer** in a pipeline extension project, we override this method :
`public abstract TOutput Process(TInput input, ContentProcessorContext context);`

However, we can't add a dependency to an external file by using `context.AddDependency(filepath)` because this method is empty :
https://github.com/MonoGame/MonoGame/blob/1c09add4a7407cf14690614d5ec36666d6c2a8ea/MonoGame.Framework.Content.Pipeline/Builder/PipelineImporterContext.cs#L22-L24

The type `ContentProcessorContext`  is abstract but the concrete type is `PipelineImporterContext` which has this empty method, and this method is not overriden anywhere else.

You may note that adding a dependency does work with a custom **processor** as the `PipelineProcessorContext` has an implementation for the method `AddDependency` :
https://github.com/MonoGame/MonoGame/blob/1c09add4a7407cf14690614d5ec36666d6c2a8ea/MonoGame.Framework.Content.Pipeline/Builder/PipelineProcessorContext.cs#L37-L40

bjornenalfa:
Fixes  #6034

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/PipelineImporterContext.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/PipelineImporterContext.cs
@@ -10,9 +10,12 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
     {
         private readonly PipelineManager _manager;
 
-        public PipelineImporterContext(PipelineManager manager)
+        private readonly PipelineBuildEvent _pipelineEvent;
+
+        public PipelineImporterContext(PipelineManager manager, PipelineBuildEvent pipelineEvent)
         {
             _manager = manager;
+            _pipelineEvent = pipelineEvent;
         }
 
         public override string IntermediateDirectory { get { return _manager.IntermediateDirectory; } }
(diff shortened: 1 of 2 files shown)


conlanpatrek:
This has been open for a while, but I needed this functionality and so I built this myself and can confirm that it works.

Thanks for the hard work. I hope it's merged in soon.

bjornenalfa:
If you need the functionality but don't want to have a custom MonoGame build you can achieve the same thing by just adding the dependencies in your ContentProcessor instead.

conlanpatrek:
> If you need the functionality but don't want to have a custom MonoGame build you can achieve the same thing by just adding the dependencies in your ContentProcessor instead.

Oh thanks for the tip! I didn't realize it was implemented in PipelineContentProcessorContexr though this makes me even more confused about the intended pipeline workflow.

It made more sense to me intuitively to flag all dependent imported files in the importer, and then flag all output files in the processor, but now... I'm less certain i understand the pipeline design.

bjornenalfa:
I agree and would like to see it fixed. From what I can tell the reason it is only implemented for the processor is because the only built-in pipeline that adds dependencies needs to do it there. Presumably it was never tested in the other steps and simply forgotten.

SimonDarksideJ (approved):

PR #7585, opened 2021-10-13
Prevent the timing of audio tests from overrunning

squarebananas:
There are currently three classes that simulate periodically updating sound effect objects by running Thread.Sleep in a loop. By placing a stopwatch around these loops it can be seen that this sometimes is quite an inaccurate method of timing.

This is likely contributing to audio tests that rely on timing randomly failing at times.

To fix this stopwatches can be used to exit the loops early, potentially preventing the tests from overrunning and failing.

Also it may be possible to reinstate the disabled BufferNeeded_DuringPlayback audio test for DesktopGL if this happens to be the cause of that. Maybe the test can be run manually a few times on the test server to see if this is the cause?


tomspilman:
This is fine... merging

Issue #7586, opened 2021-10-14
Ray class' intersects method for BoundingFrustum commented out


EclipseLunaria:
While looking through the Ray intersects method I noticed that the overloaded method for Intersects(BoundingFrustum) is commented out. Is there a particular reason for it being commented out?
```c#
 /*
        public float? Intersects(BoundingFrustum frustum)
        {
            if (frustum == null)
			{
				throw new ArgumentNullException("frustum");
			}
			
			return frustum.Intersects(this);			
        }
        */
```

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

PR #7588, opened 2021-10-15
Support Ray-Frustum Intersection

EclipseLunaria:
This PR adds support for Intersects method in Bounded Frustrum with a Ray class.

the frustum checks if the ray intersects with any of its planes, If there is an intersection it checks if the point past it using a very small epsilon value is contained within the frustum using the contains method.


JacksonDorsett:
This PR addresses issue #7586

squarebananas:
This line looks like it will be true if the ray intersects the plane and not the frustum.
`if (this.Contains(point) == ContainmentType.Contains && result == null || d < result)`

I'm guessing the brackets are missing and it should be this instead?
`if (this.Contains(point) == ContainmentType.Contains && (result == null || d < result))`

Also if you're concerned the Contains(point) test may fail due to floating point issues with the point being located close but not on the plane, then there may be alternative methods.

For example 2 triangles could be obtained by using the 4 points which form the square on the plane. These triangles could then be used with a "point within triangle" test (this would need adding). This would potentially work even if the point was slightly above/below the plane (from floating point issues) as it would be based on the point being within the angle range of the triangles.

mrhelmut:
Hi @JacksonDorsett, would you mind checking @squarebananas 's comment to move this PR forward? Thanks!

PR #7622, opened 2021-11-28
Extracted and refactored a long method 

ashi5396:


stromkos:
I object to this PR for the following reasons:

- **A long method is not inherently bad**, as long as each section is well documented.
- Separation into "single-use" methods adds unneeded overhead. **Unnecessary overhead is bad** and should be avoided in games.

My main problem with splitting all code into tiny chunks is the readability of the code to new "users" of the code. While it may make unit testing and writing easier, it muddles the waters for anyone later reading the code; too many methods/files to go through for a complete linear picture.

mrhelmut:
Closing because supposedly from a school assignment with no intention to follow up on it.

PR #7626, opened 2021-11-28
added comments and removed commented code

diwashbiswa:
I'm taking this class where I have to find the code smells using the tool and fix the issue/warning. I found multiple instances of commented-out code and comments that needed to be included at the function header.

### DIFF
--- a/MonoGame.Framework/Content/ContentReader.cs
+++ b/MonoGame.Framework/Content/ContentReader.cs
@@ -9,6 +9,9 @@ using MonoGame.Framework.Utilities;
 
 namespace Microsoft.Xna.Framework.Content
 {
+    /// <summary>
+    /// Content Reader class, inherits from Binary Reader class
+    /// </summary>
     public sealed class ContentReader : BinaryReader
     {
         private ContentManager contentManager;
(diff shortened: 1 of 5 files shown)


DimitarBogdanov:
This just says "Reads"
https://github.com/MonoGame/MonoGame/pull/7626/files#diff-33ca81a0a2fe4fc80d3f718a1750fa93a17a9090be775d5b9bfc7b0524457fd3R197

I also find some comments to be quite useless, to be honest. Why is this comment needed, for example?
https://github.com/MonoGame/MonoGame/pull/7626/files#diff-412659a32441a84a20fcdf75be5c536537d0a09fd7d80d55ddb087cab80b73ecR11

You don't need a comment for that. It's implied by the name. Besides, it's not like you're even going to be using the type name `ContentReaderExtensions` (as extensions appear under the object type you're extending), so it's likely that pretty much nobody will even get to see that class.

Other than these two things, this is a really nice commit. Especially the change in `Texture2DReader` with the parentheses - much more readable!

stromkos:
> I'm taking this class where I have to find the code smells using the tool and fix the issue/warning. 
A noble task indeed.

Most times commented code shows what was tried and did not work. This technique prevents the same errors from recurring and shows the history of the changes without referring to the diffs.

mrhelmut:
@diwashbiswa Hey, would you still be around and willing to update some of your PRs? We had a surge of school assigned PRs which are incomplete and if they were just "submit & run" assignment with no intention to get involve in the project in the long run, we will close them because they add noise to the review process.

PR #7627, opened 2021-11-29
removed unnessesary null check from Texture2D.StbSharp

JacksonDorsett:
The if statement is unnecessary since the value will always end as null


mrhelmut:
StbSharp is an external library. Fixing it in MG's repo' is likely to make it come back whenever we update the library. I'd suggest contributing to StbSharp instead.

PR #7629, opened 2021-11-29
Corrective changes to Rgba64.cs

AlexW57:


stromkos:
I am ok with everything but the removal of the redundant cast.  There is no advantage to removing it since the compiler will do so itself when not needed.

damian-666:
does this me we will have HDR support or is it just a step that way.. any samples available?  or did floating point surface formats already  support it on some cards and monitors or tvs?   anyways getting the usb to HDMI to work at 60 hz is pretty hopeless  wiht HDR for me on my surface laptop studio, but using an OLED LG TV i alwasy appreciated HDR on newer films.

stromkos:
@damian-666 ,

True HDR is a different surface type, AFAIK not implemented in MonoGame yet.  However, the major video card vendors drivers' have been known to upscale to target formats, so you may still get the full dynamic range with gradients.

damian-666:
thnka im wanting it for gradients, smoke and stuff... i get banding artifacts anywasy on HDR tv its dissapointing except bright lights and flashy swords..

mrhelmut:
Closing because this one has been from a school assignment and the author don't seem like being involved.

PR #7464, opened 2021-02-09
Optimized SpriteFont GlyphPacker

TechnologicalPizza:
Fixes #7351

PR #7637, opened 2021-12-15
Optimisations for Glyph Packing

PhantomBadger:
The Glyph Packing process for larger fonts had some heavy performance sinks, attempting to build a 65k character font took about 10 days at the least before I gave up. With my changes below it successfully built in about 2-3 hours.

The changes can be summarised as;

- Storing each glyph in a dictionary based on its index to allow for some quicker lookups later down the line
- Starts our searching for the glyph's packed location at the maximum X and Y values of previously placed glyphs that are smaller than us (If it's smaller than us and has been placed at a position, we know there's no placed before that we could squeeze into)
- After searching through a row, instead of dropping by one pixel we drop the minimum amount to reach the bottom of a glyph we intersected (If we have searched a row full of glyphs and found no spots, we know there also wont be any spots until we clear the Y value of the smallest glyph we intersected)

A lot of the performance hits came from the repeated searching and intersecting of previous glyphs using loops as we iterate over the space trying to find where to pack our target glyph, as a result these changes aimed to minimise how often we needed to search.
I'm using the packed font in one of my other projects so can attest to it's functionality remaining consistent.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/Font/GlyphPacker.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/Font/GlyphPacker.cs
@@ -3,9 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Threading.Tasks;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 {
@@ -38,17 +36,9 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 			int outputHeight = 0;
 
             // Choose positions for each glyph, one at a time.
-            // Keep a record of glyphs in a dictionary so we can look them up based on
-            // their index later for some optimisations
-            Dictionary<int, ArrangedGlyph> glyphDict = new Dictionary<int, ArrangedGlyph>();
-            for (int i = 0; i < glyphs.Count; i++)
-            {
-                glyphDict.Add(i, glyphs[i]);
-            }
-
 			for (int i = 0; i < glyphs.Count; i++)
 			{
-				PositionGlyph(glyphDict, i, outputWidth);
+				PositionGlyph(glyphs, i, outputWidth);
 
 				outputHeight = Math.Max(outputHeight, glyphs[i].Y + glyphs[i].Height);
 			}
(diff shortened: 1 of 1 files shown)


MonoGame.Framework.Content.Pipeline/Graphics/Font/GlyphPacker.cs
```diff
@@ -35,10 +37,18 @@ public static BitmapContent ArrangeGlyphs(Glyph[] sourceGlyphs, bool requirePOT,
 			int outputWidth = GuessOutputWidth(sourceGlyphs);
 			int outputHeight = 0;
 
-			// Choose positions for each glyph, one at a time.
+            // Choose positions for each glyph, one at a time.
+            // Keep a record of glyphs in a dictionary so we can look them up based on
+            // their index later for some optimisations
```
tomspilman, line 42: Hi @PhantomBadger 

I don't follow how looking it up by index in `Dictionary<int, ArrangedGlyph>` vs by index in `List<ArrangedGlyph>` changes things.

MonoGame.Framework.Content.Pipeline/Graphics/Font/GlyphPacker.cs
```diff
@@ -3,7 +3,9 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Threading.Tasks;
```
tomspilman, line 8: I don't see any new conccurent collections or tasks in this PR.  Are these `using`s needed?

PhantomBadger: Ah apologies not at all, I was originally pursuing a local optimisation involving parallel execution and must have forgotten to remove the using! I'll make another commit to rectify this shortly

MonoGame.Framework.Content.Pipeline/Graphics/Font/GlyphPacker.cs
PhantomBadger: You're completely right, I think I had some other functionality around the hash-based lookup in a previous implementation that has since been lost. 

Sorry for the confusion there, this was originally intended as a local solution to a problem but after seeing the perf increase I wanted to share it :) I'll update it appropriately in a follow-up commit.

PhantomBadger:
Added a new commit based on some feedback, apologies for the prior confusion - this was something I worked on a little bit ago and was retro-actively getting ready to make a PR out of and I must have gotten a bit mixed up!

TechnologicalPizza:
Just wanted to tune in and bump https://github.com/MonoGame/MonoGame/pull/7464, which should build absolutely massive amounts of glyphs in seconds.

PhantomBadger:
I think ditching this in favour of your change makes the most sense!

PR #7640, opened 2021-12-29
Templates: Add missing `Visible="false"` attributes for Content references

seclerp:
Seems like `.mgcb` files are **double referenced** (from `Builder.Task` and project template itself)

But not all project templates are updated with this in mind. This PR introduces missing parts.

BTW maybe it's better to introduce everywhere something like the example below (instead of `Include` + `Visible="false"`)?

```xml
<ItemGroup>
  <MonoGameContentReference Update="Content\Content.mgcb" />
</ItemGroup>
```

The issue on JetBrains YouTrack: https://youtrack.jetbrains.com/issue/RIDER-65136

Issue representation in JetBrains Rider (DesktopGL template-based project):

image


mrhelmut:
Actually they are double referenced because otherwise they should up in Visual Studio.

Unless we have a solution that pleases both VS and Rider, we'll have to arbitrate between the two (likely in favor of VS because it is the main target).

harry-cpp:
I've already fixed this with my last template update, they were double referenced from both the template and the .targets file :)

Issue #7510, opened 2021-05-22
Rendering right to left text does not work properly


ayousuf23:
When rendering a spritefont for a right-to-left language, the text appears in the wrong direction (left-to-right).

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

mrhelmut:
```SpriteFont``` currently doesn't support RTL rendering. It shouldn't be complicated to support it. If anyone wants to tackle this, I'm marking it as good first issue.

ayousuf23:
What exactly would need to be modified to support RTL rendering?

willmotil:
I imagine just a new overload of drawstring were you decrement instead of increment provided the data in spritefont has sufficient information.

mrhelmut:
I would also recommend to make the starting cursor position to be on the right side of the first character, so that it's easier to align texts.

mrhelmut:
If anyone is willing to give #7644 a test spin, that would be great. I would like to avoid doing that myself because I lack the cultural knowledge to make sure that it does things right.

BlueElectivire:
I might try that tomorrow if I could find the time (and I manage to figure out how to do that).

BlueElectivire:
So I tried to do that and I really didn't manage to test it, I couldn't figure out how to reference my working repo (Even when I removed the nuget references) and I'm not really familiar with writing in-code tests (especially not when it comes to graphics and stuff like that). If anyone is willing to try and explain it to me it'd be great.
Or if anyone would like to try that themselves and needs a quick explanation on how rtl text should look (Hebrew in my case) then I can gladly do that as well.

mrhelmut:
@BlueElectivire Ah, by "testing" I meant trying to display text with it to verify that it renders as expected for people native to a RTL language. I didn't mean to make a unit code testing. Sorry for the confusion.

Let me know how it goes and I'll be happy to merge it if it's fine.

BlueElectivire:
@mrhelmut so I tested it, turns out it didn't work.
I fixed it and committed the changes to the same branch, and now it seems to work fine.
I tested it with SpriteEffects applied as well and it works as I would expect it to.
[...]

PR #7644, opened 2022-01-09
RTL text rendering overload for SpriteBatch.DrawString

BlueElectivire:
Added new overload to the SpriteBatch.DrawString method that makes drawing right to left strings possible with a new rtl boolean parameter. Solves issue #7510.
Should work correctly, although I didn't test it because I didn't really figure out how one would go about testing that (This is my first PR in this repo).

### DIFF
--- a/MonoGame.Framework/Graphics/SpriteBatch.cs
+++ b/MonoGame.Framework/Graphics/SpriteBatch.cs
@@ -934,196 +934,100 @@ namespace Microsoft.Xna.Framework.Graphics
             var firstGlyphOfLine = true;
 
             fixed (SpriteFont.Glyph* pGlyphs = spriteFont.Glyphs)
-                if (rtl)
-                    for (var i = 0; i < text.Length; ++i)
+                for (var i = 0; i < text.Length; ++i)
+                {
+                    var c = text[i];
+
+                    if (c == '\r')
+                        continue;
+
+                    if (c == '\n')
                     {
-                        var c = text[i];
-
-                        if (c == '\r')
-                            continue;
-
-                        if (c == '\n')
-                        {
-                            offset.X = 0;
-                            offset.Y += spriteFont.LineSpacing;
-                            firstGlyphOfLine = true;
-                            continue;
-                        }
-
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Graphics/SpriteBatch.cs
```diff
@@ -847,6 +847,192 @@ public unsafe void DrawString (SpriteFont spriteFont, string text, Vector2 posit
...
+                        firstGlyphOfLine = false;
+                    }
+                    else
+                    {
+                        offset.X += spriteFont.Spacing + pCurrentGlyph->LeftSideBearing;
+                    }
+
+                    var p = offset;
+
+                    if (rtl ^ flippedHorz)
```
squarebananas, line 970: This DrawString method appears to be the same as the one above with the exception of the extra argument and this line of code. Unless I'm missing something this just adds an extra gap between the characters and makes the rendered text look broken. I don't fully understand RTL text rendering but I don't think this will achieve it.

BlueElectivire: You were right, it doesn't work. What I was trying to do is basically hijack the logic that prints horizontally flipped strings and use it to print rtl, but I misunderstood the way it works and without testing I didn't notice it didn't work as I intended.

squarebananas:
Gave the changes a try and it's working much better. The order of characters is reversed and the text alignment changes to right aligned (which I'm guessing is correct for RTL rendering).

When rendering LTR and RTL side by side though the gaps between characters still don't look quite right. In particular there are characters touching especially punctuation marks. When I swap every instance of LeftSideBearing and RightSideBearing with each other for RTL it seems to get fixed. Not sure if this happens for you also?

One other thing is all other DrawString methods have a String version and a StringBuilder version. It should just be as simple as copying/pasting the code and renaming the argument type to StringBuilder to achieve this.

BlueElectivire:
I saw what @squarebananas was talking about, and the fix he suggested worked so I implemented it, as well as the stringbuilder version (which I did just copy-paste). Looks great imo, I think you could safely merge it.

mrhelmut:
Thank you for the contribution!

PR #7655, opened 2022-02-22
Add thread locking to the UWP gamepads dictionary

squarebananas:
This pull request is to fix a potential thread issue that may occur when a gamepad is disconnected.

WGI.Gamepad.GamepadAdded and WGI.Gamepad.GamepadRemoved are adding and removing gamepads from a different thread to the game tick where PlatformGetState is likely to be called. If the timing was right a disconnected gamepad could be removed from the dictionary while also accessing it for GetState/GetCapabilities.

This can be prevented by adding a lock statement around each use of the gamepads dictionary.

### DIFF
--- a/MonoGame.Framework/Platform/Input/GamePad.UWP.cs
+++ b/MonoGame.Framework/Platform/Input/GamePad.UWP.cs
@@ -20,6 +20,8 @@ namespace Microsoft.Xna.Framework.Input
 
         private static Dictionary<int, WGI.Gamepad> _gamepads;
 
+        private static readonly object _locker = new object();
+
         static GamePad()
         {
             _gamepads = new Dictionary<int, WGI.Gamepad>();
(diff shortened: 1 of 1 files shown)


nkast:
Locking is not necessary if you replace the dictionary and linq operations with an array.
Reading and writing ints to the array is an atomic operation.
My implementation is [here](https://github.com/nkast/MonoGame/commit/78a0e9796992c84f58ae85f6696ff8554e945b56).

Of course I don't object if the maintainers chose to merged your commit right way, 
your solution is very straight forward.

damian-666:
jut a lil note, locks shouldnt not be as bad as wait handles..:    https://geeksqa.com/is-the-overhead-of-acquiring-a-lock-dependent-on-the-os-timer-resolution-when-there-is-a-context-switch        i had to stop using wait handle   on android because i cant set timer resolution on it, copy data from last update instead.

squarebananas:
> My implementation is [here](https://github.com/nkast/MonoGame/commit/78a0e9796992c84f58ae85f6696ff8554e945b56).

I gave that a try and it seems to work fine.

I was thinking the gamepad object shouldn't be used after being removed and your solution could rarely allow WGI.Gamepad.GetCurrentReading to be run after a gamepad has just been disconnected. However it turns out this works and just returns a default WGI.GamepadReading.

If someone lets me know which way to proceed I can amend the PR.

Also with that first call empty workaround I'm only getting an empty list when placing a breakpoint there but not when running normally. Not sure why that is.

nkast (approved):

nkast:
Honesty, your PR is fine. There's no reason to postpone a merge. 


About the workaround, I have no idea why it's behaving like that but I left it in just in case
it's a weird race thing that could happen every now and then.

About your other comment, I think you are saying that after reading the gamepad in this line
    var gamepad = _gamepads[index];
we might receive a GamepadRemoved event for that gamepad.
I'd say that the gamepad had been physically detached before the event had reach our code anyway.
The end result is the same whether we catch the removed event first and return a default state 
or call GetCurrentReading() on a removed gamepad.

squarebananas:
Thanks.

> About your other comment, I think you are saying that after reading the gamepad in this line
> var gamepad = _gamepads[index];
> we might receive a GamepadRemoved event for that gamepad.

Yeah that's it. I was initially thinking that calling GetCurrentReading on a removed gamepad might work like calling Play on a used SoundEffectInstance which can't be reused. After testing this I found GetCurrentReading works even if the gamepad no longer exists in the static WGI.Gamepad.Gamepads list. So I was just confirming that your method works fine.

mrhelmut:
Fine by me! Thank you!

Issue #7665, opened 2022-04-04
[DesktopGL] Memory leak on static variable (GL.Extensions)


felaugmar:
[`List<String> [Static variable MonoGame.OpenGL.GL.Extensions]`](https://github.com/MonoGame/MonoGame/blob/v3.8/MonoGame.Framework/Platform/Graphics/OpenGL.cs#L1428)

This ^ variable is used to store 'OpenGL extensions', but it is never cleared and grows forever.
I noticed that it increases when the game runs.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

mrhelmut:
Could you provide more context regarding your use case of MG? The only reason I would see for this collection to grow would be that you are creating several GraphicalContext.

felaugmar:
> Could you provide more context regarding your use case of MG? The only reason I would see for this collection to grow would be that you are creating several GraphicalContext.

Yes, I'm disposing and recreating the `Game` instance every time there is an unhandled exception, probably recreating the `GraphicalContext` as you pointed.

mrhelmut:
While the fix should be straightforward, I'd like to note that this isn't a use case within the scope of MonoGame. Restarting your game because it crashed doesn't sound like a good practice.

PR #7684, opened 2022-04-10
Fix memory leak on desktop gl static variable

adnanioricce:
From the issue #7665, there is a static variable ``GL.Extensions`` used in initialization that can grow invariably and wasn't being cleared. 
I replace the calls to the static variable to a local variable from the return result of a method. Part of the problem could be solved with a [simple check when filling ``GL.Extensions`` ](https://github.com/MonoGame/MonoGame/blob/9ce9e1b03d3e01df8b36a9afb6b3ca0d1d1ca48e/MonoGame.Framework/Platform/Graphics/OpenGL.cs#L1448). I decided to go with the route of using a local variable because it seemed a tiny change, and the memory would be cleared.

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/GraphicsCapabilities.OpenGL.cs
+++ b/MonoGame.Framework/Platform/Graphics/GraphicsCapabilities.OpenGL.cs
@@ -30,11 +30,12 @@ namespace Microsoft.Xna.Framework.Graphics
 
         private void PlatformInitialize(GraphicsDevice device)
         {
+            var extensions = GL.GetExtensions();
 #if GLES
-            SupportsNonPowerOfTwo = GL.Extensions.Contains("GL_OES_texture_npot") ||
-                GL.Extensions.Contains("GL_ARB_texture_non_power_of_two") ||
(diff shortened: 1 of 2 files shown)


mrhelmut:
Thanks for trying to enhance this. The issue with your proposition is that it's a bit over-engineered and it generates four times the amount of garbage because of using Linq's ToArray() and calling GetExtensions() twice. We're trying to avoid relying on Linq in the codebase because it is very garbage-prone.

A more simple fix would have been a one liner:
```
if (Extensions.Count == 0)
{
           string extstring = GL.GetString(StringName.Extensions);
            var error = GL.GetError();
            if (!string.IsNullOrEmpty(extstring) && error == ErrorCode.NoError)
                Extensions.AddRange(extstring.Split(' '));
}
```

Let me know if you'd be willing to update your PR.

adnanioricce:
Thanks for the response. I will revert the changes in the two commits and add this check instead.

adnanioricce:
I updated the PR, if there is something to change, please, let me know.

mrhelmut:
Thanks!

PR #7688, opened 2022-04-13
Update stuff to .NET 6

harry-cpp:
Notes:
- VS for Mac addon is no longer getting built as with VS for Mac 2022 they are no longer using Gtk for the UI, I'll be making a separate PR to upgrade it
- I've set the target framework of WindowsDX, DesktopGL, Android and iOS to only net6.0, I couldn't think of a reason to support older framework versions, especially now that everything finally got unified with it
- Templates have been updated to net6.0, but haven't been tested yet 😅 
- .NET Core 3.1 is needed for the building because of the dotnet tools that we use
- MonoGame.Content.Builder.Task is now using dotnet mgcb tool once again, this time MS provides us per project local dotnet tools, templates have been updated accordingly
- ~~Removed the C# language version lock that was set for the projects, we can use quite a bit of new language features, we just need to be careful to not use the new classes in PRs for the main framework~~
  - Only WindowsDX is still marked with LangVersion 5 to ensure shared code is using it


mrhelmut:
Should we use this PR to bump the version to 3.8.1?

> I've set the target framework of WindowsDX, DesktopGL, Android and iOS to only net6.0, I couldn't think of a reason to support older framework versions, especially now that everything finally got unified with it

Having either WindowsDX or DesktopGL having a net452 target would be useful to make sure that console versions still build. But net452 is getting deprecated by the end of the month and will likely be absent from net7 targets... Moving forward I believe that having a "reference target platform" for consoles separated from WindowsDX / DesktopGL will likely be unavoidable whenever we switch to net7+, so maybe we should just drop net452 here and I'll work on that separated target for console checks.

Templates/MonoGame.Templates.CSharp/content/MonoGame.Application.DesktopGL.CSharp/MGNamespace.csproj
```diff
@@ -1,38 +1,27 @@
 <Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
     <TargetFramework>net5.0</TargetFramework>
```
mrhelmut, line 4: net6.0

Templates/MonoGame.Templates.CSharp/content/MonoGame.Application.WindowsDX.CSharp/MGNamespace.csproj
```diff
@@ -1,30 +1,20 @@
...
-  <ItemGroup>
-    <MonoGameContentReference Include="Content\Content.mgcb" />
-  </ItemGroup>
-
-</Project>
-
+<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <TargetFramework>net5.0-windows</TargetFramework>
```
mrhelmut, line 4: net6.0-windows

mrhelmut (changes requested):
All ```DotNetCore...``` MSBuild actions are deprecated and replaced by their equivalent without ```Core```. Would you mind switching?

Missing switches:
- DesktopGL template uses net5.0
- WindowsDX template uses net5.0
- Builder.Task.props is looking for mgcb in the net5.0 path
- Builder.Launcher is using net5.0

So far everything builds, but the builder task doesn't work. It seems that mgcb is not embedded into the task nuget anymore.

mrhelmut:
> Removed the C# language version lock that was set for the projects, we can use quite a bit of new language features, we just need to be careful to not use the new classes in PRs for the main framework

I would disagree on this one. Unfortunately because of console target, there is still a strict C# 5 need here and I don't think that we can rely on human eyes to check if a language feature is safe enough for consoles.

harry-cpp:
> I would disagree on this one. Unfortunately because of console targets, there is still a strict C# 5 need here and I don't think that we can rely on human eyes to check if a language feature is safe enough for consoles.

Aren't we only limited to not using any new classes, like `Span`?

harry-cpp:
Also we you really want to, how about we just mark just one project (ie. WindowsDX) with LangVersion 5 instead of all of them, that way we can use the new fancy stuff on the platform side that is not being used by the consoles?

mrhelmut:
If we create a new "fake" platform that is a copy of either WindowsDX or DesktopGL with a net452 target and a LanguageVersion set to 5 to build with the github action, we're good. But I don't quite see the advantage because most of the code is shared anyway, it would make using newer language features limited to platform code (or having preprocessing to detect net452 everywhere).

It's not just about types like span, it's also about using APIs that aren't present in 4.5.2 and some syntactic sugars are hidden calls to assemblies absent from 4.5.2. We can't reliably track this by eyes.

harry-cpp:
> If we create a new "fake" platform that is a copy of either WindowsDX or DesktopGL with a net452 target and a LanguageVersion set to 5 to build with the github action, we're good. 

We'll get a project for that with .Native.csproj version, we can temporarily use the WindowsDX project.

> it would make using newer language features limited to platform code

That is mostly what I will need it for :D

> some syntactic sugars are hidden calls to assemblies absent from 4.5.2. We can't reliably track this by eyes.

Ah, I see.

mrhelmut:
Awesome! Thanks!

I tested DesktopGL and WindowsDX.
I'll test mobiles later but it's a good start to move forward.

evilkomododragon:
Sorry. I'm not a MonoGame developer but I was wondering why Tom Spilman, monogame lead developer according to the website, is not giving his opinions and/or advice on these decisions for the next update?
In fact, he doesn't say anything here anymore. Is Tom not working on monogame anymore?

rds1983:
Sorry for being late to the party, however - as a MG lib developer - I have important question. Or rather two questions:
1. What .NET  framework should my libs target so they could be used on all MG platforms? For now I target .net standard 2.0.
2. What MG backend should my libs reference? For now I reference MG portable 3.7, which is very-very dated. If I'll switch to MG DesktopGL 3.8.1, I'll have to bump target framework to .net 6.0. That bump might make it difficult for my libs to work on all MG platforms.
Thanks.

damian-666:
i use netstandard 2.0  ( could not go to to 2.1) and MG desktop GL as a proxy.

then I have my netcore exe as thin as possible,  targeting the various... WindowsDX or GL.   things like local storage, window sizing i do in the exe because its not been generalized..  i dont use Full screen.  I have all scalable graphics.

i took an android template i found and I adjusted to and netcore 6 and got 40% framerate on my background physics thread aat least.. im using vsync.

https://github.com/damian-666/MGShadersXPlatform

  this is my template that lets you touch  a clip shader, use a render target,  ( i couldnt figure out the ones that come with MG)  and build for all.  The shaders get rebuild every time but it works.    My code was build once , no repeat code and no linking files.   everything common in netstandard libs.  i used ideas from Nez and from lots of places. 


yuou need .net core 3.1 installed after- netcore 6 -to build content i think.  that will be fixed in new MG.

this wil get u linux , osx , and windows dx or windows gl if you want that.
the netstandard is linked to GL but its sort of a proxty , the EXE picks up the runtime the mg gpu targets.  

if mg 3.8.1 has a better way to organize solutions  im all ears.. also i think the rebuild issue untouched shaders or content is fixed.  Thers is no repeat code , you just edit the MCGB manually so that the visual tool doesnt put a target to windows or openGl.. if it did you erase it that line.    this is necessary for shaders building but not all content.  but insurse the products go in the right folders.  The shaders have #ifdef code that works on GL and DX.. at least the one im using.. I threw in some other shaders but didnt test them.


im going to waitt to see how the xamarin story unfolds before i go back to android buit i develop on a tourch surface beasue xamarin is slow , even with haxm and i end up just using windows touch, wiht mono.. build the UI, then sideload it or remote debug in droid.. its so slow im going  to waint for native ARM support in windoows, i think its might even be tehre now, in vs 2022 preview.

mrhelmut:
> Sorry for being late to the party, however - as a MG lib developer - I have important question. Or rather two questions:
> 
> 1. What .NET  framework should my libs target so they could be used on all MG platforms(including consoles)? For now I target .net standard 2.0.
> 2. What MG backend should my libs reference? For now I reference MG portable 3.7, which is very-very dated. If I'll switch to MG DesktopGL 3.8.1, I'll have to bump target framework to .net 6.0. That bump might make it difficult for my libs to work on all MG platforms.
>    Thanks.

If you need to check if your libraries build against the console targets, we made a dedicated "fake" project for that in the repository: [MonoGame.Framework.ConsoleCheck.csproj](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj). This, and targeting ```LangVersion``` 5 guarantees that projects will work as-is on consoles.

Otherwise we recommend targeting net6 because both net5 and n452 are deprecated by now. We hope to be able to have console targets running on net6 in the future (some targets already can, but not all).

Besides consoles, net6 will work everywhere for the public platforms.

rds1983:
I see, thanks for the clarification.

So if I create MyLibrary.ConsoleCheck.csproj that includes all my library sources and targets "net452"/LangVersion=5, then its successful compilation will guarantee the library will work on consoles.

That's definitely a solution. However it brings up one justified question: why not retartet MG libs back to "netstandard2.0"? Then ConsoleCheck.csproj wouldn't be required neither for MG, neither for dependented projects.

mrhelmut:
> So if I create MyLibrary.ConsoleCheck.csproj that includes all my library sources and targets "net452"/LangVersion=5, then its successful build will guarantee the library will work on consoles.

Yes, but forgot to mention that it should support ahead-of-time compilations, which comes with a few more strings attached:
- Reflection won't work
- IL emit won't work

A good test is to build with NativeAOT. If it works with NativeAOT, it'll work everywhere.

> However it brings up one justified question: why not retartet MG libs back to "netstandard2.0"(afaik, it's not too different from "net452") and set LangVersion to 5? Then ConsoleCheck.csproj wouldn't be required neither for MG, neither for dependent projects.

Only DesktopGL builds against ```netstandard2.0```. The other targets don't. ```netstandard2.0``` is compatible with ```net462``` (which won't work on consoles) but not ```net452```, so we can't use it to check compatibility. Other than that, we choose to follow .NET recommendation to target the current LTS version of .NET move forward and follow the deprecation list of .NET (which I contradictory to have consoles running on 4.5.2, but hopefully these will be upgraded to 6.0 too).

rds1983:
I see. Thanks for the thorough explanation!

PR #7690, opened 2022-04-13
Improve GamePad support on macOS and introduce support for rumble

carlfriess:
Controller support on macOS is currently a little unstable. This seems to be due to the driver implementation in SDL that usually makes use of IOKit. According to Apple this is now considered a legacy way of interacting with game controllers and use of the new [Game Controller Framework](https://developer.apple.com/documentation/gamecontroller) should be preferred. SDL includes a driver [implementation](https://github.com/libsdl-org/SDL/blob/main/src/joystick/iphoneos/SDL_mfijoystick.m) that makes use of this new framework, which also brings rumble support to macOS. Unfortunately, this drivers is only [enabled](https://github.com/libsdl-org/SDL/blob/120c76c84bbce4c1bfed4e9eb74e10678bd83120/include/SDL_config_macosx.h#L164-L167) when you build SDL2 targeting macOS >= 10.8 but the official build targets macOS 10.6.

For controller rumble, MonoGame currently relies on the SDL Haptics API. In newer versions, SDL exposes new functions that are much simpler:

- [`SDL_GameControllerRumble()`](https://github.com/libsdl-org/SDL/blob/46616af722f299b56e68e418be179aa4c1408c54/include/SDL_gamecontroller.h#L845-L863)
- [`SDL_GameControllerRumbleTriggers()`](https://github.com/libsdl-org/SDL/blob/46616af722f299b56e68e418be179aa4c1408c54/include/SDL_gamecontroller.h#L865-L888)
- [`SDL_GameControllerHasRumble()`](https://github.com/libsdl-org/SDL/blob/46616af722f299b56e68e418be179aa4c1408c54/include/SDL_gamecontroller.h#L901-L912)
- [`SDL_GameControllerHasRumbleTriggers()`](https://github.com/libsdl-org/SDL/blob/46616af722f299b56e68e418be179aa4c1408c54/include/SDL_gamecontroller.h#L914-L925)

I built SDL version 2.0.20 ([Source](https://github.com/surprised-turtle/SDL/tree/release-2.0.20-macos-10.9)) targeting macOS 10.9 (the oldest officially supported SDK) and updated MonoGame to use the new rumble API. Please let me know what you think about these changes.

I uploaded the binaries here: https://github.com/surprised-turtle/MonoGame.Dependencies/commit/06e4bb47e41430327dec597277f8f13d3eb344ab. I'd be happy to create a pull request for those as well. The windows and linux binaries are taken from FNA.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "ThirdParty/Dependencies"]
 	path = ThirdParty/Dependencies
-	url = https://github.com/surprised-turtle/MonoGame.Dependencies.git
+	url = https://github.com/MonoGame/MonoGame.Dependencies.git
 [submodule "ThirdParty/NVorbis"]
 	path = ThirdParty/NVorbis
 url=https://github.com/NVorbis/NVorbis.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "ThirdParty/Dependencies"]
 	path = ThirdParty/Dependencies
-	url = https://github.com/MonoGame/MonoGame.Dependencies.git
+	url = https://github.com/surprised-turtle/MonoGame.Dependencies.git
 [submodule "ThirdParty/NVorbis"]
 	path = ThirdParty/NVorbis
 url=https://github.com/NVorbis/NVorbis.git
--- a/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
@@ -130,8 +130,8 @@
       <PackagePath>runtimes\osx\native</PackagePath>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="..\ThirdParty\Dependencies\SDL\MacOS\Universal\libSDL2-2.0.0.dylib">
-      <Link>libSDL2-2.0.0.dylib</Link>
+    <Content Include="..\ThirdParty\Dependencies\SDL\MacOS\Universal\libSDL2.dylib">
+      <Link>libSDL2.dylib</Link>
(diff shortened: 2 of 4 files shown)


mrhelmut:
Very welcome contribution because beyond rumble support, MonoGame would also work with Apple's official support for game controllers such as bluetooth PlayStation and Xbox controllers without the need of third-party unofficial drivers.

I'll gladly merge this once we have SDL 2.0.20 merged into the dependencies.

If you wish to, I'll be happy to to merge SDL 2.0.20, providing it is built with M1 support. It seems totally safe to me to have macOS 10.9 as the oldest backward compatible version because this still covers [98.99% of current macs](https://gs.statcounter.com/macos-version-market-share/desktop/worldwide).

carlfriess:
The build currently doesn’t have M1 support. I briefly tried a universal build but it caused a weird problem with dotnet. I’ll look into this again though and report back.

mrhelmut:
Maybe @NoFr1ends can share some insights in building SDL with M1 support and backward compat down to macOS 10.9

They have an SDL 2.0.14 build pending, but I don't know if it was built with backward compat enabled:
https://github.com/MonoGame/MonoGame.Dependencies/pull/156

flibitijibibo:
I believe we build with Apple gamecontroller support, if not we definitely build with hidapi support which is the preference anyhow. M1 I haven't integrated yet, but it's possible to build with native tools as well as osxcross.

carlfriess:
Looks like my problems were simply because I was still building with the .NET 5 SDK. Universal binaries with M1 support only work on .NET >= 6.0 due to code signing issues. I've pushed the universal build and created a pull request: MonoGame/MonoGame.Dependencies#157. The build includes a small bug fix: libsdl-org/SDL#5533.

mrhelmut:
Feel free to switch back the dependencies to your commit on the official repo and I'll get this PR moving forward.

mrhelmut:
Thank you!

PR #7698, opened 2022-04-23
Console check fake project

mrhelmut:
When we moved to .NET 6, we cleaned projects and removed compatibility toward .NET 4.5.2, which was a target used to ensure that MonoGame's main repository was still compatible with the private console implementations (which requires .NET 4.5.2 and C# 5).

We removed this target in preparation to .NET 7 deprecating it and not supporting to build it.

This PR adds a backup plan to still ensure that console repositories are building by adding a "fake" target (which is a copy of DesktopGL) to build against .NET 4.5.2 with C# 5.
This fake project named ```ConsoleCheck``` will help us to know if future PRs break might consoles or not by having the github actions building it (but without packaging it as a nuget to avoid confusion).

@tomspilman thoughts?

@harry-cpp I moved the LangVersion from WindowsDX to this project.

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
@@ -1,25 +1,16 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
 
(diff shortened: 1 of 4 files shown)


harry-cpp:
Can we use a copy of WindowsDX instead of DesktopGL 😅 ?

mrhelmut:
Here you go.

PR #7765, opened 2022-04-27
Make the VS extension to open .mgcb

mrhelmut:
When the VS extension is installed, double clicking on an .mgcb file will open the MGCB Editor.

I'm not sure if this is the cleanest way to do that, but welp, it seems to work.

### DIFF
--- a/Templates/MonoGame.Templates.VSExtension/MonoGameTemplatesVSExtensionPackage.cs
+++ b/Templates/MonoGame.Templates.VSExtension/MonoGameTemplatesVSExtensionPackage.cs
@@ -28,7 +28,7 @@ namespace MonoGame.Templates.VSExtension
     /// </remarks>
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid(MonoGameTemplatesVSExtensionPackage.PackageGuidString)]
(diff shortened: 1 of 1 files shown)


Templates/MonoGame.Templates.VSExtension/MonoGameTemplatesVSExtensionPackage.cs
```diff
@@ -25,6 +28,7 @@ namespace MonoGame.Templates.VSExtension
     /// </remarks>
     [PackageRegistration(UseManagedResourcesOnly = true, AllowsBackgroundLoading = true)]
     [Guid(MonoGameTemplatesVSExtensionPackage.PackageGuidString)]
+    [ProvideEditorExtension(typeof(EditorFactory), ".mgcb", int.MaxValue)]
```
harry-cpp, line 31: If I read the docs correctly, there should be a `Default Name` property that can be used to specify the name of the editor when you right click the file.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.shell.provideeditorextensionattribute.defaultname?view=visualstudiosdk-2022

harry-cpp (changes requested):

mrhelmut:
@harry-cpp the DefaultName worked as intended, thanks.

harry-cpp (approved):

PR #7769, opened 2022-04-30
Added VS code Launch and Task base templates

stromkos:
My attempt at making the VS Code Launch and Tasks work on platforms other than Mac.

VS Code is the only editor available on Linux, capable of supporting builds for multiple platforms without manual configuration when changing build targets.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -3,24 +3,23 @@
     // Hover to view descriptions of existing attributes.
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
 
-    //The echo command is valid on all three platforms
     "version": "0.2.0",
     "configurations": [
         {
             "name": "MGCB Editor",
             "type": "coreclr",
             "request": "launch",
-            "preLaunchTask": "mgcb-editor",
-            "program": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder/Debug/mgcb",
+            "preLaunchTask": "Build mgcb-editor",
+            "program": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder/Debug/Windows/mgcb-editor-windows.exe",
             "windows": {
-                "program": "${workspaceFolder}//Artifacts/MonoGame.Content.Builder/Debug/mgcb",
+                "program": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder/Debug/Windows/mgcb-editor-windows.exe",
                 "args": [],
-                "cwd": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder.Editor/Debug",
+                "cwd": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder.Editor/Windows/Debug",
             },
             "linux": {
-                "program": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder/Debug/mgcb",
+                "program": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder.Editor/Linux/Debug/mgcb-editor-linux",
                 "args": [],
-                "cwd": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder.Editor/Debug",
+                "cwd": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder.Editor/Linux/Debug",
             },
             "osx": {
                 "program": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder.Editor/Mac/Debug/MGCB Editor.app/Contents/MacOS/mgcb-editor-mac",
@@ -29,56 +28,6 @@
             },
             "console": "internalConsole",
             "stopAtEntry": false
-        },
-        {
-            "name": "Build DesktopGL",
-            "type": "coreclr",
-            "request": "launch",
-            "preLaunchTask": "Build MonoGame Framework (DesktopGL)",
-            "program": "echo ",
-            "args": [],
-            "console": "internalConsole",
-            "stopAtEntry": false
-        },
-        {
-            "name": "Build IOS",
-            "type": "coreclr",
-            "request": "launch",
-            "preLaunchTask": "Build MonoGame Framework (IOS)",
-            "program": "echo ",
-            "args": [],
-            "console": "internalConsole",
-            "stopAtEntry": false
-        },
-        {
-            "name": "Build WindowsDX",
-            "type": "coreclr",
-            "request": "launch",
-            "preLaunchTask": "Build MonoGame Framework (WindowsDX)",
-            "program": "echo ",
-            "args": [],
-            "console": "internalConsole",
-            "stopAtEntry": false
-        },
-        {
-            "name": "Build WindowsUniversal",
-            "type": "coreclr",
-            "request": "launch",
-            "preLaunchTask": "Build MonoGame Framework (WindowsUniversal)",
-            "program": "echo ",
-            "args": [],
-            "console": "internalConsole",
-            "stopAtEntry": false
-        },
-        {
-            "name": "Build MonoGame Framework Content Pipeline",
-            "type": "coreclr",
-            "request": "launch",
-            "preLaunchTask": "Build MonoGame Framework Content Pipeline",
-            "program": "echo ",
-            "args": [],
-            "console": "internalConsole",
-            "stopAtEntry": false
-        },
+        }
     ]
 }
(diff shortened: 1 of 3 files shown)


stromkos:
This change provides a **template** for vscode projects. Output directories to run the MGCB-Editor depend on target frameworks, and must be corrected.


The rest of the 'Build' targets will build then fail with the message:

>> "The target process exited without raising a CoreCLR started event. Ensure that the target process is configured to >> use .NET Core. This may be expected if the target process did not run on .NET Core.
>> The program 'echo' has exited with code 0 (0x0)."

This message indicates a successful build. 

Since this error will only occur for those people compiling from source on VS Code using the build launch target, the impact should be minimal, since the only target previously was for the 'mac'.

`launch.json` requires an  executable program, and "echo" was the best cross platform low impact option, for building a dll.

VS Code was designed around executable code, not libraries, so to make a launch target, a cross platform executable (echo) was chosen.

A `json` file has no programmatic elements.

So the best that I can provide is a template, given the variability of output directories.

stromkos:
This PR provides a template for VS Code to allow building and running of code on platforms other than Mac.

I must stress the template nature, since each of the build-target or build-targets, framework(s) has its own directory.

I also added the build requirement of the content pipeline before building MGCB editor.

I considered adding a dependency between the content pipeline and the framework in the end I thought the extra build times were not needed in most cases.

.vscode/launch.json
```diff
@@ -2,18 +2,83 @@
...
+            "osx": {
+                "program": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder.Editor/Mac/Debug/MGCB Editor.app/Contents/MacOS/mgcb-editor-mac",
+                "args": [],
+                "cwd": "${workspaceFolder}/Artifacts/MonoGame.Content.Builder.Editor/Mac/Debug/MGCB Editor.app/Contents/MacOS",
+            },
+            "console": "internalConsole",
+            "stopAtEntry": false
+        },
+        {
+            "name": "Build DesktopGL",
```
harry-cpp, line 34: launch.json is used for launch configurations, I wouldn't keep the library projects in here.

.vscode/tasks.json
```diff
@@ -2,16 +2,143 @@
...
+                    "build",
+                    "${workspaceFolder}/Tools/MonoGame.Content.Builder.Editor/MonoGame.Content.Builder.Editor.Windows.csproj",
+                    "/property:GenerateFullPaths=true",
+                    "/consoleloggerparameters:NoSummary"
+                ]
+            },
+            "problemMatcher": "$msCompile"
+        },
+        {
+            "label": "Run mgcb-editor",
```
harry-cpp, line 119: There is already a launch configuration for mgcb-editor above, also `dotnet run` is invalid for the mgcb-editors mac project, you need to use `open -a <pathtogeneratedappfile>`

.vscode/tasks.json
```diff
@@ -2,16 +2,143 @@
...
+                "${workspaceFolder}/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj",
+                "/property:GenerateFullPaths=true"
+            ],
+            "problemMatcher": "$msCompile"
+        },
+        {
+            "label": "Build mgcb-editor",
+            "command": "dotnet",
+            "type": "process",
+            "dependsOn": ["Build MonoGame Framework Content Pipeline"],
```
harry-cpp, line 91: Don't add any `dependsOn`, `dotnet build` will already resolve the dependency while building the `.csproj` so now we are just slowing down the build process.

harry-cpp (changes requested):

.vscode/tasks.json
```diff
@@ -116,29 +109,98 @@
             "problemMatcher": "$msCompile"
         },
         {
-            "label": "Run mgcb-editor",
+            "label": "Clean MonoGame Framework (Android)",
```
harry-cpp, line 112: I don't think there is a reason to add clean task 😅

.vscode/tasks.json
stromkos: Thank you for letting me know about the `dotnet run` on Mac.

.vscode/tasks.json
stromkos: Probably my ignorance with the SDK style workflow, but I found that changes in the Content pipeline would not cause a rebuild on a framework build on Net Core 3.1(this may have been fixed since then).  

It ran with the old code and gave the message:
>> All projects are up-to-date for restore.

This is the reason for the `[DependsOn]` and `clean`.  

I had considered a VS style "rebuild" task and hiding the cleans. But AFIK, the task system does not allow purely compound tasks.

Other than the additional space on the task list, I do not see any reason not to add it.

People are free to modify these files as they see fit.  It is easier to customize via deletion than addition.

harry-cpp: Conntent.Pipeline references the MonoGame.Framework, therefore if you modify Content.Pipeline, there is no reason for MonoGame.Framework to be rebuilt.

> People are free to modify these files as they see fit. It is easier to customize via deletion than addition.

True, but I don't think we should keep anything except build tasks in there, otherwise you get a question of, should we keep rebuild, pack, publish, etc.

stromkos: Normally, I would block comment out the offending code to allow it to be re-added by the user, but the json format does not allow that.

For the above reasons, my PR stands **as is**, notwithstanding corrections to the`mgcb-editor` file names and locations.

The true :sweat_smile: is that the original and subsequent PR's did not take other OS's into account at all. I recall you beating that into me with my first PR. (**Thank You**, I have not forgotten that lesson.)

stromkos:
@Jjagg, since you merged the original files, do you have an opinion on the matter at hand?

harry-cpp:
He is not around right now, only mrhelmut and myself are.

The only person on the team that uses vs code is me, thats why there was only mgcb-editor for macOS configuration in both the tasks and launch jsons, that was the only thing I needed to work on MG.

.vscode/tasks.json
harry-cpp: > Normally, I would block comment out the offending code to allow it to be re-added by the user, but the json format does not allow that.

json format does not, but VS Codes json configuration files do allow it.

harry-cpp:
Now that I think about it, what use do you have for all the build/clean tasks?

stromkos:
> The only person on the team that uses vs code is me, 

Ok, I'll leave the other issues that I have with that statement alone, and focus on the major one: 

Since you are the only one using vs code: Does this PR have any negative bearing on your workflow?

> Now that I think about it, what use do you have for all the build/clean tasks?

I freeze the version for each project that I work on. When I start a new project, I download the latest version of the framework and apply my additions and corrections; most are font related like `Alpha8` fonts, true emoji support, runtime font replacement, color font support, 16K+ texture support, Vulkan shader support.  

Then, I strip any unused code for the project type and target at hand.  

All of this work is done on my Linux workstation, which unlike OSX has only one viable editor. 

---

The cleans are needed after the initial build runs to verify functionality, before implementing my custom C# wrappers, or replacements, for upstream libraries.

The builds allow for quick error checking without an project.

---

> The only person on the team that uses vs code is me, 

Now, since you are the only person on the team, **and by extension the entire world?**, not using Visual Studio (by choice in your case) to make modifications to the framework, this PR is completely useless and should be closed, since you are the only user of vs code in the universe.

stromkos:
@harry-cpp, My apologies for my bluntness, but many of the features in the above list would have been submitted PR's (making this one unneeded), had the environment here not been so harsh.

mrhelmut:
Having VS Code config files to build the repo with VS Code is legitimate to have on the repo. I believe that what Harry wanted to mean is that we didn't bother adding those files because it wasn't a need for repository maintenance (which is different than saying that external needs don't exist).

I believe that the only question that remains is if the clean tasks really are needed (but I guess it can't harm to have them). Otherwise welcome contribution!

stromkos:
> Yea, @mrhelmut already explained it by now sweat_smile , I was just explaining why @Jjagg merged them. Sorry if I offended you.

I can say it rubbed me the wrong way. 

> No worries, be as blunt as you want, it's the internet.

Be that as it may, my outburst was uncivil and violated "PLEASE keep a civil and respectful tone when discussing and reviewing contributions."  For that I owe you an apology, I am sorry.

> Yep, I've been thinking about it and I think we should add a `vscode.cake` script to autogenerate tasks and luanch json files(build/clean/rebuild/pack possible version) for all the projects in `MonoGame.Framework`, that way if we modify the projects we can easily regenerate the files. We should also commit the generated files by default.

That sounds great, and makes this PR unneeded. 

I would offer to help, but my knowledge of cake involves eggs and flour.  

This is the great thing about open source:

 "At the start of this PR, I was trying to save myself some keystrokes, without harming the existing code.  By my second revision that goal was accomplished. After the third round, a better path forward was announced." 

---
"The hard road is usually worth it in the end. "

PR #7771, opened 2022-04-30
Update templates

harry-cpp:
- Add ImplicitUsings to all net6.0 templates
- Use file scoped namespaces in all net6.0 templates
- Use top-level statement in Program.cs
- Add iPad support to Info.plist in iOS templates
- Add automatic dotnet tool restoring

### DIFF
--- a/MonoGame.Framework/Directory.Build.props
+++ b/MonoGame.Framework/Directory.Build.props
@@ -8,8 +8,6 @@
     <BaseOutputPath>..\Artifacts\MonoGame.Framework\$(MonoGamePlatform)</BaseOutputPath>
     <AssemblyName>MonoGame.Framework</AssemblyName>
     <RootNamespace>Microsoft.Xna.Framework</RootNamespace>
-
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
 </Project>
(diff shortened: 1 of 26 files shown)


mrhelmut:
Interesting use of implicit usings. Thanks!

Mindfulplays:
This is awesome, thank you!
I did notice that for an iOS build on a Mac M1 with .NET 6, I still don't see file-scoped namespace working:

```
Welcome to .NET 6.0!
SDK Version: 6.0.104
Microsoft (R) Build Engine version 16.10.1 for Mono
...


error CS1514: { expected
```

(Running on a Mac Mini M1 / arm64)

Is there a newer version of .NET that lets you build this successfully?
Most other things (pattern-matching / switch expression / record etc) seem to work fine though.
Thanks!

harry-cpp:
Hmm, are you using VS for Mac 2022?

mrhelmut:
Peculiar.
Running .NET SDK 6.0.202 here on Windows and I have no problem building all the templates with file-scoped namespace (being from the CLI or from VS).

Mindfulplays:
(@mrhelmut yes, on Windows, .NET 6 works fine with file-scoped namespaces).

I am using VS 2019 but installed .NET SDK / runtime from the website. I am not sure if VS 2022 is fully released yet...

mrhelmut:
It isn't fully released yet on macOS, but this isn't supposed to be bound to VS. Your NET 6 version seems quite up-to-date.

PR #7788, opened 2022-05-03
Auto versioning of VS extension

mrhelmut:
Make the cake script to inject automatically a friendly version string into the VS Extension for Windows.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -291,6 +291,18 @@ Task("PackVSTemplates")
     .WithCriteria(() => IsRunningOnWindows())
     .Does(() =>
 {
+    var shortVersion = version;
+	if (shortVersion.Contains("-")) shortVersion = shortVersion.Substring(0, shortVersion.IndexOf("-"));
+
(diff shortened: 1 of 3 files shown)


build.cake
```diff
@@ -291,6 +291,18 @@ Task("PackVSTemplates")
     .WithCriteria(() => IsRunningOnWindows())
     .Does(() =>
 {
+    var shortVersion = version;
+	if (shortVersion.Contains("-")) shortVersion = shortVersion.Substring(0, shortVersion.IndexOf("-"));
```
harry-cpp, line 295: Random tab instead of spaces 😅 

Also can be written much shorter but less optimised :P
```cs
var shortVersion = version.Split('-')[0];
```

harry-cpp, line 295: Random tab instead of spaces 😅 

Also can be written much shorter but less optimised :P
```cs
var shortVersion = version.Split('-')[0];
```

mrhelmut:
Your suggestion is much more readable. I also removed a couple other tab characters throughout the script.

PR #7789, opened 2022-05-03
Added RollForward directive

mrhelmut:
I went ahead and moved #7574 forward across all the repository so that people coming from the future won't have questions in regard to requirements.


harry-cpp:
If I understood it correctly, `RollForward` should only be added to projects that produce an executable.

https://docs.microsoft.com/en-us/dotnet/core/versions/selection#framework-dependent-apps-roll-forward

mrhelmut:
Updated!

Templates/MonoGame.Templates.CSharp/content/MonoGame.Application.Android.CSharp/MGNamespace.csproj
```diff
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net6.0-android</TargetFramework>
+    <RollForward>Major</RollForward>
```
harry-cpp, line 4: Post merge regrets, iOS and Android aren't supposed to have it, there is no point for them 😅

PR #7790, opened 2022-05-04
Update wine script for NET 6

mrhelmut:
Is this okay to you @harry-cpp ?

Couldn't we host d3dcompiler_47.dll ourselves?


harry-cpp:
> Is this okay to you @harry-cpp ?

Yep, this should be fine, I'll test it out properly once I get my new mac.

> Couldn't we host d3dcompiler_47.dll ourselves?

I just copied what winetricks script does to install it in wine, I'm not sure if we would be breaking some tos if we included it ourselves.

PR #7791, opened 2022-05-06
Cleaned up rollforward from mobiles

mrhelmut:
Here you go @harry-cpp


mrhelmut:
Actually auto-merged it, I guess it's safe enough.

harry-cpp:
I guess I was too late :P

PR #7795, opened 2022-05-09
fix Effect Constructor with Index

### LINKED ISSUE #7794: Error on creating Effect with byte array and range
*example code:*
```
{
    var mgfxo = File.ReadAllBytes(@"test.mgfxo");
    var index = 100000;
    var byteArray = new byte[index+mgfxo.Length];
    Buffer.BlockCopy(mgfxo, 0, byteArray, index, mgfxo.Length);
    var effect = new Effect(this.GraphicsDevice, byteArray, index, mgfxo.Length);
}

```
*result:* 
```
System.ArgumentOutOfRangeException: Non-negative number required. (Parameter 'count')
   at System.IO.MemoryStream..ctor(Byte[] buffer, Int32 index, Int32 count, Boolean writable, Boolean publiclyVisible)
```

*root cause:*
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Effect/Effect.cs#L131
```
header.HeaderSize = index;
```

The `index` here is absolute index of the entire buffer,
and we can't assume that the `index` argument passed in this function to always be 0.

- MonoGame 3.8

- Windows

- WindowsDX

----

The bug is there for over 8yrs but even no test case covers it:(
image

nkast:
fixes #7794

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/Effect.cs
+++ b/MonoGame.Framework/Graphics/Effect/Effect.cs
@@ -92,15 +92,21 @@ namespace Microsoft.Xna.Framework.Graphics
  
             //Read the header
             MGFXHeader header = ReadHeader(effectCode, index);
-			var effectKey = header.EffectKey;
-			int headerSize = header.HeaderSize;
+            if (header.Signature != MGFXHeader.MGFXSignature)
+                throw new Exception("This does not appear to be a MonoGame MGFX file!");
+            if (header.Profile != Shader.Profile)
+                throw new Exception("This MGFX effect was built for a different platform!");
+            if (header.Version > MGFXHeader.MGFXVersion)
+                throw new Exception("This MGFX effect seems to be for a newer release of MonoGame.");
+            if (header.Version < MGFXHeader.MGFXVersion)
+                throw new Exception("This MGFX effect is for an older release of MonoGame and needs to be rebuilt.");
 
             // First look for it in the cache.
             //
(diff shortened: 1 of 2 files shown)


BlueElectivire:
Perhaps change the exceptions to a more descriptive type of exception like System.IO.FileFormatException? Would probably make logs more accurate and helpful and it's not a big change.

nkast:
@mrhelmut , in addition to the fix, I moved the file checks outside of the ReadHeader. as I think it's best to have them in the main  method and keep the readHeader method clean and short. This makes it easier to change the code with custom formats, support back versions, etc.  let me know if, it's ok. 

@BlueElectivire that sounds reasonable. Do you think we should change all the exceptions or just the first one that checks for the file signature?

nkast:
@mrhelmut , in addition to the fix, I moved the file checks outside of the ReadHeader. as I think it's best to have them in the main  method and keep the readHeader method clean and short. This makes it easier to change the code with custom formats, support back versions, etc.  let me know if, it's ok. 

@BlueElectivire that sounds reasonable. Do you think we should change all the exceptions or just the first one that checks for the file signature?

mrhelmut:
Thanks! I'll review it soon, it seems straightforward enough.

BlueElectivire:
@nkast I suppose all of them, because they are all to do with the format of the file not being compliant with what the program is looking for and I don't think there is another exception type that describes that more accurately.

PR #7797, opened 2022-05-12
[MGCB Editor] Added file watcher for the content project file with project reload prompt

beartwigsam:
Right now, if you have the MGCB Editor open and you pull in changes to the content project file from Git, the editor won't reload those changes automatically nor will it warn you that the project file has been changed outside of the editor. If you then go ahead and make some changes and save, the changes you pulled in would be overwritten.

This PR adds a file watcher for the content project file and prompts a dialog asking if the user would like to reload the project whenever an unexpected change is detected.

image

### DIFF
--- a/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
@@ -306,6 +306,7 @@ namespace MonoGame.Tools.Pipeline
                 _projectFileWatcher.Path = dirName;
                 _projectFileWatcher.Filter = fileName;
             }
+            _projectFileWatcherIgnoreEvent = false;
 
             UpdateTree();
             View.UpdateTreeItem(_project);
--- a/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
@@ -11,8 +11,8 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Eto.Forms;
 using MonoGame.Content.Builder;
-using Application = System.Windows.Application;
 using PathHelper = MonoGame.Framework.Content.Pipeline.Builder.PathHelper;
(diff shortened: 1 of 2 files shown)


Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
```diff
@@ -11,6 +11,7 @@
 using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
+using Eto.Forms;
```
harry-cpp, line 14: Don't reference Eto.Forms from PipelineController, instead of `Application.Instance.Invoke` call `View.Invoke`.

Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
```diff
@@ -123,6 +129,26 @@ public static PipelineController Create(IView view)
             return new PipelineController(view);
         }
 
+        public void SetupProjectFileWatcher(string projectFilePath)
```
harry-cpp, line 132: You are not uninitializing FileWatcher when you close the project (`CloseProject` method).

harry-cpp (changes requested):

Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
beartwigsam: Fixed

Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
beartwigsam: Fixed

BenMakesGames:
This is rad, and solves a real & common problem, making devs' lives better.

With my messy desktop, I also sometimes open a second copy of MCGB editor via the IDE, which can also cause one of the instances to get out of sync with disk changes. I'm more diligent about checking for this now, but I'd still appreciate the tool helping me out for those times when my diligence wanes...

SimonDarksideJ (approved):

PR #7799, opened 2022-05-13
Move type encoding of Mojo shader in mgcb format

nkast:
Moves the information about Mojo internal encoding of Booleans & integers into the file format.
Moves the information about Mojo constant uniforms into the file format.

Removes the assumption that   OPENGL shaders = Mojo shaders

Removes conditional compilation in Effect & EffectParameters.

Enables future improvements (Generic EffectParameter ? ) 

Enables building stand alone effect processors with other compilers. This change is needed to implement a processor with ShaderConductor, see https://github.com/MonoGame/MonoGame/blob/e2df5a0fd76d83229c484ddd9ffd21658a50c015/MonoGame.Framework/Graphics/Effect/EffectParameter.cs#L183-L189


nkast:
@cpt-max , can you review this?

Additionally I would like to move the serialization of the effects into custom binary readers.
This way we would not have to pass parameters as arguments all the way down like I did in https://github.com/MonoGame/MonoGame/pull/7799/commits/65b1944679230f18273652aa2338184cf281639a
and also to implement Zig-Zag encoding to replace the 7bitEncoding that was removed recently.

Tools/MonoGame.Effect.Compiler/Effect/EffectObject.writer.cs
```diff
@@ -142,30 +142,33 @@ public void Write(BinaryWriter writer, Options options)
...
+                WriteParameter(writer, profile, parameters[i]);
         }
 
-        private static void WriteParameter(BinaryWriter writer, d3dx_parameter param)
+        private static void WriteParameter(BinaryWriter writer, ShaderProfile profile, d3dx_parameter param)
         {
             var class_ = ToXNAParameterClass(param.class_);
             var type = ToXNAParameterType(param.type);
+            // MojoShader encodes integers and booleans into a float.
+            var encodeAsType = (profile == ShaderProfile.OpenGL) ? EffectParameterType.Single : type;
```
cpt-max, line 157: This will need to be changed for ShaderConductor. Only MojoShader encodes everything as float. ShaderConductor uses ints too. I'm pretty sure it doesn't do bools though, they will end up as ints.

cpt-max (commented):
This should work fine to get rid of the MojoDataLayout field I introduced in EffectParameter.cs. I still need to know if an effect was compiled using MojoShader or ShaderConductor for another reason though -> uniform buffer handling. MojoShader doesn't use uniform buffers, ShaderConductor does.

Tools/MonoGame.Effect.Compiler/Effect/EffectObject.writer.cs
nkast: The way I see this going forward is to write a separate processor along side the current d3d/Mojo compiler.
Each Processor would write the appropriate data and the deserialized Effect should be able to handle all shaders with no assumptions about the platform or the complier. 

The Writer however is not a good place to override encodeAsType.  I will move this inside the  parameter, it will make thing much clearer.

nkast: Here we set encodeAsType to float inside 'ConstantBufferData.mojo.cs' for integers and booleans. 
Everywhere else 'encodeAsType' equals 'type'.

https://github.com/MonoGame/MonoGame/blob/eeedfda19baaeeec82d7359f8a6c60cb13c7fce4/Tools/MonoGame.Effect.Compiler/Effect/ConstantBufferData.mojo.cs#L75-L90

nkast:
I see what the problem is. 
Mojo encodes all parameters as one big array of floats and monogame update all of them in one go.
https://github.com/MonoGame/MonoGame/blob/a9e5ae6befc40d7c86320ffdcfcd9d9b66f786a8/MonoGame.Framework/Platform/Graphics/Shader/ConstantBuffer.OpenGL.cs#L69-L77

In one of your PRs you have three paths,
1) using the current trick with GL.Uniform4() for Mojo,
2) using uniform buffers for ShaderConductor 
2) fallback to individual uniforms if uniform buffers are not supported.
https://github.com/MonoGame/MonoGame/blob/5c64b94a42bca6013bbcf0785297517bdf647932/MonoGame.Framework/Platform/Graphics/Shader/ConstantBuffer.OpenGL.cs#L70-L101 
 
I am shifting towards to a single flag in the file, since all those seem very specific to Mojo, and all those changes seem to pollute the code with unnecessary changes.

nkast:
@cpt-max I suppose I can get you method `UpdateParametersIndividually()`  and plug it in my constant buffer PlatformApply() .

cpt-max:
> I am shifting towards to a single flag in the file, since all those seem very specific to Mojo, and all those changes seem to pollute the code with unnecessary changes.

I think that's pretty much how I did it in my PR too. I added some Mojo flag to the shader file. 

> I suppose I can get you method UpdateParametersIndividually() and plug it in my constant buffer PlatformApply()

Alright. This fallback is needed for older OpenGL versions. Since 3.1 UBO support is guaranteed. If it was up to me I would drop support for everything before OGL 3.3 pretty soon, because it's just so outdated nowadays.

PR #7800, opened 2022-05-13
Remove default Inner Exceptions

nkast:


mrhelmut:
Thanks! I have no idea why these were there in the first place.

nkast:
Odd, right? The code is gone now, so let's move to the next thing!
It's nice to see so all that activity on the project.

PR #7805, opened 2022-05-21
Automatic issue closing

mrhelmut:
This PR adds an automatic github actions that will run every day at 12am UTC.

This action will mark issues (not PRs) that have been inactive for 350 days with a notice message inviting to bump the issue.

If no activity is detected within the next 15 days after being marked as inactive, the issue will be closed automatically (and inviting to verify the validity of the issue against the latest version before opening a new issue).

If an issue gets active again, the mark is removed automatically.

It's basically a "prune inactive issues older than 1 year" action.

We can adjust the timing. As of now, it would close 591 out of 672 issues (or 450 if we set it to 2 years).
If need be, we can configure the action to never expire issues that we mark with a specific label.

### DIFF
--- a/.github/workflows/stale_issues.yml
+++ b/.github/workflows/stale_issues.yml
@@ -1,0 +1,20 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  close-issues:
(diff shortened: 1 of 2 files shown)


mrhelmut:
I will revisit this later.

Issue #7634, opened 2021-12-09
Suggestion: Add Buttons.None, to match Keys.None


JamesC01:
I'll preface this by saying that this may just be me being naive, and this might not fit with the framework, so sorry if that's the case.

I'm trying to write a re-usable input class, which has Actions, which represent the inputs.

public class Action
{
    public Keys key { get; set; } = Keys.None;
    public Buttons button { get; set; } = Buttons.None;
    public MouseButtons mousebutton { get; set; } = MouseButtons.None;
}
In some cases, you might want a keyboard only, or mouse only action. The problem is, Buttons.None doesn't exist. (MouseButtons was simple enough to make, though, so that's not a problem.) I could make Buttons nullable, and I think that would work, but doesn't feel like the most consistent choice.

I read in another issue that when polling for controller input, you'll never get the value None, which is a problem, I suppose. Adding it just for some edge cases may not be the best choice, but it also seems fairly consistent. Why would their only be Keys.None, and not Buttons.None.

JamesC01:
I'm reading my code back again, and I'm thinking that making all of them nullable might be the better choice anyway. All I'm doing is checking if the key isn't none, and then returning the state of the key. I may be wrong, but nullable feels like it would be a cleaner choice there instead. Not sure.

Apostolique:
What are you trying to do? I have this library, maybe it helps you with inputs? <https://apostolique.github.io/Apos.Input/>

JamesC01:
@Apostolique Yeah, basically exactly what I'm trying to do. I'm mainly doing this as practice, though, so I'm trying to do it myself. I've figured out a fairly clean way by making the enums nullable, though.

mrhelmut:
```Buttons``` is a flag, so you can use it to do something like this: ```public Buttons button { get; set; } = (Buttons)0; // no button set```

It's not elegant, but it works.

If you feel like it, contributing a ```Buttons.None = 0``` is straightforward and I'd be happy to merge it.

PR #7818, opened 2022-06-10
Add Buttons.None

TimerbaevRR:
#7634

### DIFF
--- a/MonoGame.Framework/Input/Buttons.cs
+++ b/MonoGame.Framework/Input/Buttons.cs
@@ -13,6 +13,11 @@ namespace Microsoft.Xna.Framework.Input
     public enum Buttons
     {
         /// <summary>
+        /// Reserved.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
         /// Directional pad up.
         /// </summary>
         DPadUp = 1,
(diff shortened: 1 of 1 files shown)


TimerbaevRR:
Fixes #7634

BaseCrosser:
About 6 years ago I also tried to add a `Buttons.None`. #4618
Doing this comes with some problems.

As was said back then the `GamePad.GetState().IsButtonDown` implementation is doing a bitwise & comparison between the pressed buttons and the parameter, but `Anything & 0` returns 0, and `0 == 0` is always true
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Input/GamePadState.cs#L138

With this implementation of adding `Buttons.None` and setting its value to 0 the result of querying `GamePad.GetState().IsButtonDown(Buttons.None)` would always return true.

TimerbaevRR:
@mrhelmut  Maybe should we revert this PR?
I will try to fix problems and create new PR

mrhelmut:
I don't feel like there is a need to revert this PR, it doesn't seem to cause any harm.

We should be fine with a ```if (button == Buttons.None) return false``` in ```IsButtonDown()```.

mrhelmut:
I went ahead because the test were failing because of this and I'm looking into restoring them: #7822

PR #7828, opened 2022-06-29
Create a release when pushing a tag

mrhelmut:
This is a proposal to automate the generation of a draft release whenever a tag is pushed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       packages: write
+      contents: write
     needs: [ build ]
     if: ${{ github.event_name == 'push' }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -93,7 +93,18 @@ jobs:
...
+      - name: Make a release
+        uses: ncipollo/release-action@v1
+        with:
+          name: 'MonoGame ${{ github.ref_name }}'
+          body: 'This is the official MonoGame ${{ github.ref_name }} release source code and Visual Studio 2022 extensions for installing templates.'
+          tag: ${{ github.ref_name }}
+          draft: true
+          allowUpdates: true
+          removeArtifacts: true
+          artifacts: "**/*.vsix,**/*.mpack"
```
harry-cpp, line 105: We should also add nuget files to the release.

harry-cpp (changes requested):

PR #7837, opened 2022-07-13
System.Drawing does not exist on android

cobalthex:
Not sure why this ever worked

### DIFF
--- a/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs
+++ b/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Xna.Framework
 
         bool disposed = false;
         ISurfaceHolder mHolder;
-        System.Drawing.Size size;
+        Point size;
 
         ManualResetEvent _waitForPausedStateProcessed = new ManualResetEvent(false);
         ManualResetEvent _waitForResumedStateProcessed = new ManualResetEvent(false);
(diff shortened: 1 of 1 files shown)


mrhelmut:
```Mono.Android``` and ```Mono.iOS``` implement a few of the ```System.Drawing``` features, such as basic types like ```Point```.

This proposed change is actually a breaking change because it is part of the public API. If games use the ```View``` service (```Game.Services.GetSevice(typeof(View))```), which is very likely on Android, the expected type for ```Size``` would change.

Beside harmonizing types within the codebase, would there be any other motivation?

mrhelmut:
Building the repository should just take a ```dotnet cake``` from the repository's folder. Though building Android individually with ```dotnet build ./MonoGame.Framework/MonoGame.Framework.Android.csproj``` should just work too.

Make sure that you have the Android workload installed (```dotnet workload install android```).

cobalthex:
building android was fine, but when trying to use 3.8.0 (public release) on a netcore6-android project, I got an assembly load failure for System.Drawing.Core v4.0.0

mrhelmut:
3.8.0 is not compatible with ```net6-android``` targets, it requires the "Xamarin" style of projects (likely Xamarin.Android 11 with VS 2019).

3.8.1 supports ```net6-android``` only.

cobalthex:
fyi, this code still exists in 3.8.1

mrhelmut:
> `Mono.Android` and `Mono.iOS` implement a few of the `System.Drawing` features, such as basic types like `Point`.
> 
> This proposed change is actually a breaking change because it is part of the public API. If games use the `View` service (`Game.Services.GetSevice(typeof(View))`), which is very likely on Android, the expected type for `Size` would change.
> 
> Beside harmonizing types within the codebase, would there be any other motivation?

Since it breaks the virtual property inherited from ```SurfaceView```, I don't think this is desired change.

You issue likely comes from somewhere else.

Issue #7848, opened 2022-07-24
Remove global usings from 3.8.1 nuget packages.


vpenades:
I've just tested the new 3.8.1 nugets, and I've found they have in built global usings for all monogame namespaces.

```csproj
<Project>

  <PropertyGroup>
    <MonoGamePlatform>DesktopGL</MonoGamePlatform>
  </PropertyGroup>

  <ItemGroup>
    <Using Include="Microsoft.Xna.Framework.Audio" />
    <Using Include="Microsoft.Xna.Framework" />
    <Using Include="Microsoft.Xna.Framework.Content" />
    <Using Include="Microsoft.Xna.Framework.Graphics" />
    <Using Include="Microsoft.Xna.Framework.Media" />
    <Using Include="Microsoft.Xna.Framework.Input" />
    <Using Include="Microsoft.Xna.Framework.Input.Touch" />
    <Using Include="MonoGame.Framework.Utilities" />
  </ItemGroup>

</Project>

```

This might look like a great idea for beginners and simple projects, but for already existing, or complex projects that might require using multiple third party libraries, it prevents managing namespaces correctly.

Consider this use case: A project using MonoGame AND BepuPhysics, which uses System.Numeric.Vectors by default.

Since Monogame has Vector3 type, and System.Numerics also has Vector3 type, And Monogame is aplying a "global using" on their types project wide. It forces a developer to type System.Numerics.Vector3 with the full name to prevent a collision with monogame's Vector3.

In any case, developers must be able to decide whether to use global usings or not, and by forcing these global usings into the developers, it makes much more difficult to resolve complex type collision scenarios.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

mrhelmut:
It might be best to remove those implicit usings. If anyone wishes to contribute, it is likely a good first contribution.

In the mean time, if anyone runs into type collisions due to this, I'd suggest disabling implicit usings by adding this to your project csproj: ```<ImplicitUsings>disable</ImplicitUsings>```

vpenades:
I can give a try with a PR

PR #7853, opened 2022-07-25
Removed explicit global usings

vpenades:
From #7848

Removed explicit global usings to let developers choose whether to use them or not, and prevent type collisions with third party libraries.

Motivation: Developers should be able to choose when and how Global Usings are defined, and not set explicitly by the library. 

Usings management is a key feature required to prevent type collisions when using third party libraries that have types with the same names. (Color, Vector3, Matrix, Rectangle, Curve, Plane, Point, Ray, etc)

Setting global usings explicitly forces developers to fully qualify the names of the types of the third party libraries project wide. Or disable global usings completely, which cannot be disabled selectively on explicit monogame globals.

### DIFF
--- a/Documentation/articles/getting_started/toc.yml
+++ b/Documentation/articles/getting_started/toc.yml
@@ -16,7 +16,7 @@
     href: 2_creating_a_new_project_netcore.md
 - name: 3. Understanding the Code
   href: 3_understanding_the_code.md
-- name: Adding Content
+- name: 4. Adding Content
   href: 4_adding_content.md
-- name: Adding Basic Code
+- name: 5. Adding Basic Code
   href: 5_adding_basic_code.md
--- a/Documentation/articles/migrate_38.md
+++ b/Documentation/articles/migrate_38.md
@@ -4,7 +4,7 @@ Migrating from 3.8.0 should be straightforward for most platforms.
 
 The major difference is that 3.8.1 now requires .NET 6 and Visual Studio 2022. You can follow the [environment setup tutorial](./getting_started/0_getting_started.md) to make sure that you are not missing any components.
 
-The MGCB Editor is no more global .NET tool and we recommend that you use the new Visual Studio 2022 extension which helps accessing it without the need of CLI commands.
+The MGCB Editor is no longer a global .NET tool and we recommend that you use the new Visual Studio 2022 extension which helps accessing it without the need of CLI commands.
 
 ## WindowsDX, DesktopGL, and UWP
 
--- a/Documentation/articles/platforms.md
+++ b/Documentation/articles/platforms.md
@@ -108,7 +108,7 @@ Building for Android requires the .NET Xamarin component to be installed. You ca
 
 The latest version of Xcode will also be required.
 
-You can an iOS game on Windows by [pairing your Visual Studio 2022 with a mac on your local network](https://docs.microsoft.com/en-us/xamarin/ios/get-started/installation/windows/connecting-to-mac/). This feature is not avaible for Rider, Visual Studio Code, or the CLI.
+You can test and deploy an iOS game on Windows by [pairing your Visual Studio 2022 with a mac on your local network](https://docs.microsoft.com/en-us/xamarin/ios/get-started/installation/windows/connecting-to-mac/). This feature is not avaible for Rider, Visual Studio Code, or the CLI.
 
 ## Other templates
 
--- a/MonoGame.Framework/MonoGame.Framework.Android.targets
+++ b/MonoGame.Framework/MonoGame.Framework.Android.targets
@@ -4,16 +4,5 @@
     <MonoGamePlatform>Android</MonoGamePlatform>
   </PropertyGroup>
 
-  <ItemGroup>
-    <Using Include="Microsoft.Xna.Framework.Audio" />
-    <Using Include="Microsoft.Xna.Framework" />
-    <Using Include="Microsoft.Xna.Framework.Content" />
-    <Using Include="Microsoft.Xna.Framework.Graphics" />
-    <Using Include="Microsoft.Xna.Framework.Media" />
-    <Using Include="Microsoft.Xna.Framework.Input" />
-    <Using Include="Microsoft.Xna.Framework.Input.Touch" />
-    <Using Include="MonoGame.Framework.Utilities" />
-  </ItemGroup>
-
 </Project>
 
--- a/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net452</TargetFramework>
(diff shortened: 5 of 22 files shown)


mrhelmut:
Thanks! This is welcome. Would you mind also updating the templates? They make heavy use of implicit usings and have ```ImplicitUsings``` enabled in their csproj as well.

vpenades:
@mrhelmut sure, I can copy the global usings to the csprojs of the templates

mrhelmut:
Either this, or removing them entirely. Up to you, both work. I don't have an opinion on which is best for newcomers.

vpenades:
> Either this, or removing them entirely. Up to you, both work. I don't have an opinion on which is best for newcomers.

I think templates are going to be used to newcomers, so it's not a bad idea to make it easy for them with the global usings

But as they progress, they should have the ability to edit/remove global usings if they require it

mrhelmut:
Your call! I'll double-down on it.

harry-cpp:
The Using directives are only used if `ImplicitUsings` tag is set to true, you can also modify which usings can included with it in the .csproj, so don't remove the lines in the .targets file.

The templates have `ImplicitUsings` set to true, I thought it would be nicer for new users.

vpenades:
> 

I use global using in my projects, but I use them for other libraries, not for MonoGame.

If the solution is to disable _all_ global usings, then it prevents developers to use global usings the way they want.

Making it easy for newcomers is not a bad idea, it's just that the global usings should go in the templates, not in the nugets. Making it easy for newcomers must not come at the cost of making it more difficult for advanced developers

harry-cpp:
> If the solution is to disable _all_ global usings, then it prevents developers to use global usings the way they want.

You have `Using Remove=` if you want to not include specific implicit usings that come with nugets.

vpenades:
@harry-cpp From my perspective, templates are going to be used mostly by newcomers, so it makes sense that all helping bits go into the templates.

Experienced users might reference nugets directly, from visual studio nuget package manager, or directly editing the csproj. Usually into projects that combine other dependencies.

When I added the new nuget packages to some of the projects and help libraries I have, I got conflicts with:

OpenCVSharp
MathNet.Numerics
BepuPhysics
System.Numerics.Vectors
System.Drawing.Common
System.Drawing.Primitives
SkiaSharp
WPF's System.Windows.Imaging
SixLabors.ImageSharp
MAUI

(most imaging libraries have their own Color, Point and Rectangle types)

So I have to add the "Remove=" on all the projects where I have monogame and any of these libraries.

Back to helping newcomers, in order to build a new complete project, you rarely use MonoGame _alone_ , so the moment a newcomer references one of the aforementioned libraries, it's probably easier for him to figure out that he has to remove the global using in _his_ csproj.... than to figure out the global usings are hidden in the nuget package, and he has to apply a "Remove=" because he cannot edit the targets in the nuget.

Personally, I believe having the global usings in the templates is the best of both worlds, so I leave this PR here to choose whether to merge it or not.

mrhelmut:
From my perspective, I'm thinking about "customer support", and I personally wouldn't want to have to explain this every time someone drops by with a type collision (because who read the documentation these days?) arguing that MonoGame doesn't work.

I'd vouch for removing global usings entirely from both the targets and the templates. As convenient as they are, the support cost implied is too high.

harry-cpp:
Maybe we make a seperate nuget `MonoGame.ImplicitUsings` for quick implicit usings?

@mrhelmut I don't have any strong feelings for one way over the other, leaving it up to you on what we should do.

vpenades:
I'll be awaiting for the decision, let me know if I leave the PR as is, or I remove the global usings, and re-introduce standard usings into the templates.

mrhelmut:
> I'll be awaiting for the decision, let me know if I leave the PR as is, or I remove the global usings, and re-introduce standard usings into the templates.

If that's not too much asking, I'd like to validate the removal of global usings and restoring usings in the templates.

mrhelmut:
Ah, while you're at it, would you mind changing ```Microsoft.NET.Sdk.WindowsDesktop``` to ```Microsoft.NET.Sdk``` in the DirectX template?

vpenades:
> Ah, while you're at it, would you mind changing `Microsoft.NET.Sdk.WindowsDesktop` to `Microsoft.NET.Sdk` in the DirectX template?

What about a new PR that fixes that, and also adds the Embedded sources to the PDB?

mrhelmut:
> Embedded sources to the PDB

I'm ok with this being in another PR, though I'd like the template fix to be merged quickly to publish a 3.8.1 hotfix.
The embedded source can likely wait 3.8.2.

vpenades:
> > Embedded sources to the PDB
> 
> I'm ok with this being in another PR, though I'd like the template fix to be merged quickly to publish a 3.8.1 hotfix. The embedded source can likely wait 3.8.2 (this needs a bit more testing).

I've also noticed WindowsUniversal uses Windows.SDK.Extras instead of Microsoft.NET.Sdk ... should I change that one too?  I think extras is no longer needed.

vpenades:
I've pushed the Windows.NET.Sdk change a moment ago.... I've left the SDK.Extras out because I think it needs further investigation

mrhelmut:
> > > Embedded sources to the PDB
> > 
> > 
> > I'm ok with this being in another PR, though I'd like the template fix to be merged quickly to publish a 3.8.1 hotfix. The embedded source can likely wait 3.8.2 (this needs a bit more testing).
> 
> I've also noticed WindowsUniversal uses Windows.SDK.Extras instead of Microsoft.NET.Sdk ... should I change that one too? I think extras is no longer needed.

Sorry, I meant from the templates, not the library projects (I would feel safer to not change that for ConsoleCheck and UWP libraries for the time being).

vpenades:
I've been following the evolution of the SDKs, and that's more or less the history:

1- In NetCore3.1 if you wanted a project to use WindowsForms or WPF, you used `Windows.NET.Sdk.WindowsDesktop`

2- When Net5.0 was introduced, `.WindowsDesktop` was no longer needed, and in order to use WindowsForms or WPF you just use `<UseWindowsForms>true</UseWindowsForms>` or `<UseWPF>true</UseWPF>`

All this is explained here:  https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/5.0/sdk-and-target-framework-change

So as I understand it, .WindowsDesktop has been deprecated and is no longer needed.

mrhelmut:
This I understand, I'd just like to avoid doing too many things (and touching more than the templates) in the PR but I concede that I did bring the topic first. Up to you but please test the libraries if they run and not just build.

We really need to get rid of the global usings in the templates, some of them simply don't build at all even on empty projects.

vpenades:
> This I understand, I'd just like to avoid doing too many things (and touching more than the templates) in the PR but I concede that I did bring the topic first. Up to you but please test the libraries if they run and not just build.
> 
> We really need to get rid of the global usings in the templates, some of them simply don't build at all even on empty projects.

OK, I'm going to revert the SDK changes and focus on fixing the global using issues on the templates ASAP... I'll leave the SDK changes for another PR for later today

vpenades:
So far, all the templates compile for me.
I was also able to run the desktop/UWP templates
I couldn't test the iOS one because I don't have the hardware.

In order to compile and run the templates directly I had to do these local changes (NOT in the PR):
- uninstall mgcb global tools from 3.8.0
- remove linux and mac packages from the dotnet-tools.json, because it complains about not being able to install them
- restore, exit visual studio, and compile again (this is not related to monogame, from time to time it happens with our own tools too)

mrhelmut:
Would you mind removing the ```<ImplicitUsings>true</ImplicitUsings>``` lines from the template csproj as well? We'll be good after that!

vpenades:
> Would you mind removing the `<ImplicitUsings>true</ImplicitUsings>` lines from the template csproj as well? We'll be good after that!

done

mrhelmut:
Great, thanks!

PR #7870, opened 2022-07-28
Add support for MaximizeWindow in DesktopGL platforms (uses SDL_MaximizeWindow)

HS-Dave:


mrhelmut:
If we're adding a method to the public API, we should implement it for all platforms to not mislead users with platform-specific public APIs.

HS-Dave:
> If we're adding a method to the public API, we should implement it for all platforms to not mislead users with platform-specific public APIs.

I agree in principle, although the work to add to other platforms is a margin higher - but there is already precedent for platform only features in this class such as IsBordless to hide the border which docs specify "Currently only supported on the WindowsDX and DesktopGL platforms." and FileDrop which obviously only works on desktop platforms.

mrhelmut:
Yes, but we should avoid making an exception to be the rule if we want the framework to be more accessible. At minimum, I'd like to suggest that we have the public API available for all platforms, and per-platform throwing NotImpletedException than having no public API at all.

damian-666:
i use very thin core front ends that do the window managemetn adn user settings stuff..  took some days or mabye ovefr a week of unexpectedly hard work. i dont think at this level much can be generalized , and even with through SDL via dllimport. 

   so i think really well structured samples are the way to go for this.    misleading APIs that arent working on everything ( even SDL ) will just lead people the wrong way.
   
   i dont do consoles yet  can do all destop and all mobile with one or two Solutions and no repeat code at all, and im thrilled by that even with 3.8 especiailly since i am using a touch PC for the androidn developement  before i bother to emulate it or remote debug or deploy it to there i can get it to work and it often just works first try.  next i want an ARM surface so i think MSFT is killing it now i hope they dont go back to old ways and i doubt they can or will.


  i had loads of trouble talking to SDL  via dllimport, so i have and Xconsole (meaning .net core console app for openGL desktops linux and mac ) with GL,  and and a thin winodw lancher for windows DX, and all my game code. in netstandard 2.0   only  way i could get windows to work was settting up delegates in the start to set up the winodws, and the windows EXE talks  to windows forms ,  getting the window handle.   theres just no way to factor out things sort of things when there is not a common factor.   i dont dare to update MG just now, but if i find i have a decent skeleton ill upload it plus the windows state stuff..  im still waiting to see how the android/ xamain / ARM windows story pans out..

 herers what im using now  https://github.com/damian-666/MGShadersXPlatform...    its amazing to me what monogame already does factor our for liek 10 platfroms already, and Unity is sorta taking desperatea measures now and its starting to piss people off,  so i dont feel its worth the bother , Mg is killing it already.   .. i wrap my windows DX to makethe windows store package which i might never bother to put in their store that noone uses so i didnt fall into the UWP early adopter hole, and the packager doesnt require special work nor it it yet stabilized since their start menu is wrecked.
 

heres what i ended up doing.  in my startup delegate code for windows anyways..


 ..using Point = Microsoft.Xna.Framework.Point;

namespace DesktopApp
{
    class Settings
    {

        static void BeginCode()
        {
            ReloadLastSaved(); 
        }

      
        static public void LoadSettings()
        {

            Form form = (Form)Control.FromHandle(CoreGame.Instance.Window.Handle);


            //TODO if offset or somethign dont use, happed once
            if (DataStore.Instance.ContainsKey(windowPosXKey))
            {
                Point pos = new Point(
                    int.Parse(DataStore.Instance[windowPosXKey]),
                    int.Parse(DataStore.Instance[windowPosYKey]));

                if (pos.X < 0 || pos.Y < 0)
                    return;

                   CoreGame.Instance.Window.Position = pos;

                CoreGame.Instance.Width = int.Parse(DataStore.Instance[windowWidthKey]);
                CoreGame.Instance.Height = int.Parse(DataStore.Instance[windowHeightKey]);

                form.Width = CoreGame.Instance.Width;
                form.Height = CoreGame.Instance.Height;
                form.Location = new System.Drawing.Point(pos.X, pos.Y);
            }

            if (DataStore.Instance[windowMaxedKey] == "True")
            {

               form.WindowState = FormWindowState.Maximized;
            }
        }

        const string windowPosXKey = "WindowPosX";
        const string windowPosYKey = "WindowPosY";
        const string windowWidthKey = "WindowWidth";
        const string windowHeightKey = "WindowHeight";
        const string windowMaxedKey = "WindowMaximized";

        public static void OnGraphicsDeviceSizeChanged()
        {
            SaveWindowState();
        }

        public static void SaveWindowState()
        {
            if (!CoreGame.Instance.Window.AllowUserResizing)
                return;

            Form form = (Form)Control.FromHandle(CoreGame.Instance.Window.Handle);
            bool isMaximized =  form.WindowState == FormWindowState.Maximized;

            if (!isMaximized)
            {

                DataStore.Instance[windowPosXKey] = form.Location.X.ToString();
                DataStore.Instance[windowPosYKey] = form.Location.Y.ToString();
                // DataStore.Instance[windowPosXKey] = CoreGame.Instance.Window.Position.X.ToString();
                // DataStore.Instance[windowPosYKey] = CoreGame.Instance.Window.Position.Y.ToString();
                DataStore.Instance[windowWidthKey] = CoreGame.Instance.Window.ClientBounds.Width.ToString();
                DataStore.Instance[windowHeightKey] = CoreGame.Instance.Window.ClientBounds.Height.ToString();
            }

            DataStore.Instance[windowMaxedKey] = isMaximized.ToString();
            DataStore.SaveToDisk();
        }



Also showing the producer / consumer pattern might help alot of people i wsa stuck for months on UWP and andrond trying to lock the physics thread so i could draw the stuff, now im lockless and its blazing and trouble free.. i get up to 4000! physics fps when i used to struggle for 200 and w/  stutter  semphores and timeouts  and all that grief.

HS-Dave:
that seems an insane amount of external work to do for what is currently 18 lines in API and 1 call in implementation. I still strongly disagree that adding a method like this that is only available for specific platforms waters down the API in any meaningful way (and no more than throwing a not implemented exception) but I would rather implement the feature for whichever platforms support it than over engineer an external window manager. All we are trying to do here is maximise an already existing window without having to manually fetch the screen dimensions and set them ourselves :)

damian-666:
It is insane amount of work for a trivial  yet important little detail.  I just released my " final solution" doesn't even work anymore.. no scaling factor , one monitor, nothing weird.  and it's being mysterious and I'm using vs 2022 preview so I'm giving up it after spendogn another two hours  trying to fix it again.    forms tells me if it's maximized but I can't save the size of it's not.  Monogame Viewport, form client, size whatever.  To see if maximized,  One might try to query the size of the screen there are allot of.people who struggle with this stupid simple thing.  I donno , be nice if it was in monogame but I can even implement it on windows now for one platform, sigh.. I'm assuming it's just preview software and tools in flux and I'll leave it for later.  I hate computers.

HS-Dave:
> > I still strongly disagree that adding a method like this that is only available for specific platforms waters down the API in any meaningful way (and no more than throwing a not implemented exception)
> 
> I should have started with that I guess. We have the plan to [unify the public API](https://github.com/MonoGame/MonoGame/issues/4571#issuecomment-1178704039) and cleanse it of platform-specific `#if`. For that project to happen, we need the public API to be the same for all platforms. We also believe that it is a better experience to have a code that compiles as-is for all platform (and proceed silently) and that doesn't oblige multi-platform projects to have `#if` too.

in context of your plan this now makes a lot more sense - I will see what I can do to support the non SDL platforms with this same feature as I feel its a pretty useful base feature, leave it with me a couple of evenings

damian-666:
ok i got it window state, maximized, and any aspect ration save/load   working in windows DX.. using Forms which is a tiny dependency .       Before ths fix is avail if its adddressed, maybe someone can save themselves some time by using the delegates at least for windows, i did need to use forms, to see the window size.     i dont need to query the desktop size but you might if you  or yhour users change that desktop size  and then the window is gone on load ..


the issue is aspect ratio:  
   I dont pass in 1280 and 720 to my Game constructor and i cant use full screen because i want it to act like a normal window...But it seems to want to put it in that aspect ratio if i tell MG window what size to be or the backbuffer size..., its wont let me set it to just any rectangle shape.     if i allow user resizing fine.. but not programmatically, or i couldnt find a way.   

 mabye forms isnt necessary but i didnt find another way.   .. the issue is the preferred back bufffer tries to maintain aspect ratio.  i dont care about that.   maybe im missing a settng to that it doesnt care about the size you pass in..  im not passing in anyingn using default Game constructor.
 
   i use vector graphics.. so i often want to make a thin or small game window that fits near my bebugger, stay there,, then run maximized, save, close, minimize and have acode and graphics visible like  before so i dont save when its maximized... . anyone with a 4k screen using the game doent need the whole screen for it either.     So if you put at least support for any aspect windows, setting  into MG via SDL and or forms,  , its gonna be useful for anyone that isnt doing pixel perfect graphics and has a normal camera that keeps the aspect ratio like most vector samples cameera like the Aether physics, and such.         herer is the fixed code n case its still relevant and if anyone finds it useful.. at least on windows... 
   
   
   herer is my save / load window state delegate:  
     
        static public void LoadSettings()
        {

            Form form = (Form)Control.FromHandle(CoreGame.Instance.Window.Handle);


            //TODO if offset or somethign dont use, happed once
            if (DataStore.Instance.ContainsKey(windowPosXKey))
            {
                Point pos = new Point(
                    int.Parse(DataStore.Instance[windowPosXKey]),
                    int.Parse(DataStore.Instance[windowPosYKey]));

                if (pos.X<0||pos.Y<0)  // TODO prolly do something more  if desktop size changed and its not on the screen..  can query all that stufff..
                    return;


                form.Location=new System.Drawing.Point(pos.X, pos.Y);
                CoreGame.Instance.Window.Position=pos;   //dont remember if this worked..  probablby does..



                var width = int.Parse(DataStore.Instance[windowWidthKey]);
                var height = int.Parse(DataStore.Instance[windowHeightKey]);


                if (width<10||height<10)
                    return;

                CoreGame.Instance.Width=width;   //this tries to set the preferred back buffer but only one aspect is acugtally set.. 
                CoreGame.Instance.Height=height;  //this will not set it , will be adjusted when i read it back..

//so i do this... 
                form.Width=width;  //                CoreGame.Instance.Width;  //this is NEEDED.  otherwise is uses the original size to try and preserve aspect ratio but I use a camera and vectors so i dont care about that.. not doing , pixel art
                form.Height=height; // CoreGame.Instance.Height

            }

            if (DataStore.Instance[windowMaxedKey]=="True")
            {

                form.WindowState=FormWindowState.Maximized;
            }
        }

        const string windowPosXKey = "WindowPosX";
        const string windowPosYKey = "WindowPosY";
        const string windowWidthKey = "WindowWidth";
        const string windowHeightKey = "WindowHeight";
        const string windowMaxedKey = "WindowMaximized";

        public static void OnGraphicsDeviceSizeChanged()
        {
            SaveWindowState();
        }

        public static void SaveWindowState()
        {
            if (!CoreGame.Instance.Window.AllowUserResizing)
                return;


            try
            {


                Form form = (Form)Control.FromHandle(CoreGame.Instance.Window.Handle);

                bool isMaximized = form.WindowState==FormWindowState.Maximized;  //this has nothing to do with full screen its a special windows statethat is ecoverable to last sized window.  

                if (!isMaximized)  //dont  bash ur debugging size settings  to maximum just because u exited maximized while testing..
                {

                    DataStore.Instance[windowPosXKey]=form.Location.X.ToString();
                    DataStore.Instance[windowPosYKey]=form.Location.Y.ToString();


                    //        DataStore.Instance[windowWidthKey]=form.ClientSize.Width.ToString();
                    //          DataStore.Instance[windowHeightKey]=form.ClientSize.Height.ToString();


                    //works fine dont need to read from  forms, iether way is fine
                    DataStore.Instance[windowWidthKey]=CoreGame.Instance.GraphicsDevice.Viewport.Width.ToString();
                  
               DataStore.Instance[windowHeightKey]=CoreGame.Instance.GraphicsDevice.Viewport.Height.ToString();

                }

                DataStore.Instance[windowMaxedKey]=isMaximized.ToString();


                DataStore.SaveToDisk();

            }


            catch (Exception exc)
            {

                Debug.WriteLine("error saving window state "+exc);
            }
        }

damian-666:
thanks.. ill remember to wait and  use this ..im sure windows handle has the info somewhere via interop and the C api.

SimonDarksideJ:
Closing due to lack of response.

This capability can be achieved by using the [GraphicsDeviceManager.ToggleFullScreen](https://docs.monogame.net/api/Microsoft.Xna.Framework.GraphicsDeviceManager.html#Microsoft_Xna_Framework_GraphicsDeviceManager_ToggleFullScreen) Method, or by implementing the approach indicated in [this article](https://industrian.net/tutorials/fullscreen-and-windowed/)

Happy to re-evaluate if the use case can be detailed more in the description of this PR

PR #7895, opened 2022-08-02
Fixed _isRunningOnNetCore bug in ContentTypeReaderManager.cs

### LINKED ISSUE #7872: Exception during XNB content loading : mismatch between "mscorlib" and "System.Private.CoreLib"
Hello!

I have migrated my projects to MonoGame 3.8.1.303 and everything is building correctly.

Unfortunately, at runtime, when I am trying to load XNB content from the ContentManager...

```csharp
ArmatureData armature = _monogameContentManager.Load<ArmatureData>(myArmatureName);
```

... I get this error :

```
Microsoft.Xna.Framework.Content.ContentLoadException: 'Could not find ContentTypeReader Type.
Please ensure the name of the Assembly that contains the Type matches the assembly in the full type name:
Microsoft.Xna.Framework.Content.ArrayReader'1[[System.Numerics.Matrix4x4, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]
(Microsoft.Xna.Framework.Content.ArrayReader'1[[System.Numerics.Matrix4x4, mscorlib]])'
```

I think the problem comes from a mismatch between `System.Private.CoreLib` and `mscorlib`. My whole solution is in .NET 6, and so are the MonoGame packages, so there shouldn't be any references to `mscorlib`, am I correct?

# Context

I am trying to load this type :

```csharp
public class ArmatureData
{
    public Matrix4x4[] BindPose { get; set; }
    public Matrix4x4[] InverseBindPose { get; set; }
    public int[] SkeletonHierarchy { get; set; }
}
```

As you can see, I am using `System.Numerics.Matrix4x4` and not the `Microsoft.Xna.Framework.Matrix` but it never caused any problems until today.

My solution has 3 projects (all in `net6.0-windows`) :
- `MyGame.Shared`, a class library in `net6.0-windows` that does not have any package or reference, and containing shared models including my `ArmatureData` class.
- `MyGame.MgcbPipelineExtension`, a class library in `net6.0-windows` used dynamically by the MGCB Builder Tool, and that references `MyGame.Shared` and these packages :
```xml
    <PackageReference Include="MonoGame.Framework.Content.Pipeline" Version="3.8.1.303" />
    <PackageReference Include="MonoGame.Framework.WindowsDX" Version="3.8.1.303" />
```
- `MyGame.App`, the main executable in `net6.0-windows` that references `MyGame.Shared` and these packages :
```xml
    <PackageReference Include="MonoGame.Framework.WindowsDX" Version="3.8.1.303" />
    <PackageReference Include="MonoGame.Content.Builder.Task" Version="3.8.1.303" />
```

# Possible solution?

I think the mismatch between `System.Private.CoreLib` and `mscorlib` comes from here :

https://github.com/MonoGame/MonoGame/blob/574fd7f870a90d4c85ea6776acda359b08874d2a/MonoGame.Framework/Content/ContentTypeReaderManager.cs#L229-L232

The detection of whether we are in .NET Core doesn't seem to work, but I am not really sure : 
https://github.com/MonoGame/MonoGame/blob/574fd7f870a90d4c85ea6776acda359b08874d2a/MonoGame.Framework/Content/ContentTypeReaderManager.cs#L24

When I am loading XNB content at runtime (in .NET 6), it looks like `_isRunningOnNetCore` is set to `false` for a strange reason, and this line wrongfully replace `System.Private.CoreLib` by `mscorlib` in the fully qualified name of the `Matrix4x4` type.

In any case, since we are now in .NET 6 only, could we just remove these lines ?
Or maybe just keep this one for compatibility purpose ?

https://github.com/MonoGame/MonoGame/blob/574fd7f870a90d4c85ea6776acda359b08874d2a/MonoGame.Framework/Content/ContentTypeReaderManager.cs#L230

I am not really familiar with these parts so I am not sure what it could break, but it seems to be the solution to my problem.

What do you think?

eosxff:
System.Private.CoreLib is an assembly, so `Type.GetType("System.Private.CoreLib")` always returned null. `typeof(object).Assembly.GetName().Name == "System.Private.CoreLib"` should fulfill the role the other line had intended to accomplish.

This should close #7872

### DIFF
--- a/MonoGame.Framework/Content/ContentTypeReaderManager.cs
+++ b/MonoGame.Framework/Content/ContentTypeReaderManager.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Xna.Framework.Content
 
         private static readonly string _assemblyName;
 
-        private static readonly bool _isRunningOnNetCore = Type.GetType("System.Private.CoreLib") != null;
+        private static readonly bool _isRunningOnNetCore = typeof(object).Assembly.GetName().Name == "System.Private.CoreLib";
 
         static ContentTypeReaderManager()
         {
@@ -51,7 +51,7 @@ namespace Microsoft.Xna.Framework.Content
             // Trick to prevent the linker removing the code, but not actually execute the code
             if (falseflag)
             {
-                // Dummy variables required for it to work on iDevices ** DO NOT DELETE ** 
+                // Dummy variables required for it to work on iDevices ** DO NOT DELETE **
                 // This forces the classes not to be optimized out when deploying to iDevices
                 var hByteReader = new ByteReader();
                 var hSByteReader = new SByteReader();
@@ -152,7 +152,7 @@ namespace Microsoft.Xna.Framework.Content
                                 catch (TargetInvocationException ex)
                                 {
                                     // If you are getting here, the Mono runtime is most likely not able to JIT the type.
-                                    // In particular, MonoTouch needs help instantiating types that are only defined in strings in Xnb files. 
+                                    // In particular, MonoTouch needs help instantiating types that are only defined in strings in Xnb files.
                                     throw new InvalidOperationException(
                                         "Failed to get default constructor for ContentTypeReader. To work around, add a creation function to ContentTypeReaderManager.AddTypeCreator() " +
                                         "with the following failed type string: " + originalReaderTypeString, ex);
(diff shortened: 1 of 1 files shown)


eosxff:
I guess my IDE also picked up on some trailing whitespace in comments and removed those as well

mrhelmut:
I suspect that ```Assembly.GetName()``` might not work with NativeAOT and will crash, it's a case where ```Assembly``` or ```Name``` might be ```null``` on that runtime.

harry-cpp:
We should probably also reverse the check, that is have the condition be by default true, otherwise if `mscorlib` is available, set it to false.

eosxff:
There are caveats with reflection, but this code works as intended in NativeAOT

eosxff:
@harry-cpp if it's not System.Private.CoreLib shouldn't it be mscorlib then? Is there a reason reversing it is better?

PR #7899, opened 2022-08-03
mgfxc_wine_setup.sh download dotnet x64 instead of dotnet x86

### LINKED ISSUE #7898: `mgfxc_wine_setup.sh` downloads the x86 dotnet exectuable instead of the x64 exectuable
Currently `mgfxc_wine_setup.sh` creates a 64-bit wine prefix than installs the 32-bit version of dotnet into `system32`. This is incorrect. The 32-bit executable should be in the `syswow64` folder, so wine freaks out when trying to execute a 32-bit executable from `system32`. MGFXC requires being on a 64-bit computer to run, so the 64-bit version of dotnet should be downloaded and installed instead. 

Before 3.8.1, the 64-bit version of dotnet 5 was installed then PR #7790 changed it from dotnet 5.0 x64 to dotnet 6.0 x86.

Likely the cause of:
 - #7865 
 - [Game build fails when shaders are used](https://community.monogame.net/t/game-build-fails-when-shaders-are-used/17817)
 - [MGFXC 8.3.1 / Wine](https://community.monogame.net/t/mgfxc-8-3-1-wine/17809)

- MonoGame 3.8.1

Tacodude7729:
One-liner fixes #7898

### DIFF
--- a/Tools/MonoGame.Effect.Compiler/mgfxc_wine_setup.sh
+++ b/Tools/MonoGame.Effect.Compiler/mgfxc_wine_setup.sh
@@ -36,7 +36,7 @@ wine64 regedit crashdialog.reg
 popd
 
 # get dotnet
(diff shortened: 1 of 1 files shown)


harry-cpp:
Awesome notice @Tacodude7729, merging!

PR #7906, opened 2022-08-11
Update tutorials.md

girlmonogame:
Hi

Added 3 books to tutorials.md


Hedmun395:
Tantrum bet for Aug, 14th 2022, 04:00 UTC

girlmonogame:
@Hedmun395 : you have been reported to github for abuse...

tomspilman:
FYI.

We've decided as a group to block [girlmonogame](https://github.com/girlmonogame) from any and all communications with the MonoGame project, team members, and users.

We have tried to be patient with them as they have reasons this behavior, but it has gotten to be too much.

The difficult thing is that they continue to create new emails and GitHub accounts and work around bans.  Nothing we can do with that other than ignore them and continue to ban them on site.

Sucks it had to go this far, but we had little choice left.

PR #7907, opened 2022-08-11
Add tutorial series for first 2D game for beginners

Forged1212:
This PR adds a tutorial series for beginners to build a 2D space shooter from start to finish.

The game project in the tutorial series addresses common features that beginners might struggle:

- Adding collision interactions
- Creating a basic user interface
- Managing objects in the game scene
- Adding different sets of content including font and music
- General code organization

The tutorial series bridges some community tutorials together as available resources. This should make newcomers feel more comfortable with jumping around to other community tutorials. It also does not use any existing libraries other than having the MonoGame extension installed.

Additionally, we added a new category called "Tutorials" and left room for other contributors to add their own.

Co-Authored with @AlexJeter17


AlexJeter17:
Along with this PR I have a question. We wanted to address our SoundEffects and Songs. We have a folder in my repository that we have the sounds and soundeffects in where they are still currently linked to. Is there somewhere you would like us to put this sound content? - https://github.com/AlexJeter17/MonoGameStarShooter

I am interested in making video tutorials about this tutorial series. I have a set up for it and if this goes through or you have any tips, I will start making them.

HCLzlh (approved):

Mindfulplays:
This is awesome, thanks for doing this. 

My only nit is the generous use of `static` everywhere... I wonder if it makes sense to help set the right standard here for those that follow a tutorial... (`static` will force all other code to be `static` too and is contagious/dangerous especially when you are starting off on a new project).

SimonDarksideJ:
Looks interesting, will set aside some time to go through it properly, but might be good to also get ported over to the MonoGame.Samples repo, along with the guide.

Kinda reminds me of the tutorial I wrote back in 2012, which was a redevelopment of an old DigiPen tutorial
https://darkgenesis.zenithmoon.com/recapthe-xna-2d-from-the-ground-up-tutorial-series.html

I do agree with @Mindfulplays comment though, tutorials need to demonstrate "Good Practice" and statics are not exactly recommended for data/state.

BenMakesGames:
> My only nit is the generous use of `static` everywhere...

> `static` [...] is contagious

I like the contagious nature of static, but only when `static` is used well, and I agree that this code is misusing `static`. `static` methods should generally be "pure": not referencing any variable that was not explicitly passed in as an argument, and ideally not modifying any objects passed in unless explicitly signaled to the caller (`out`). Changing outside state makes code harder to reason about, and more brittle (higher chance of introducing bugs when future changes are made).

Other thoughts:
* Avoid `virtual`/`abstract`, unless it's unreasonable to do so. Composition is generally preferable: https://en.wikipedia.org/wiki/Composition_over_inheritance Your `EntityManager` also suggests you're going for an ECS-style architecture (https://en.wikipedia.org/wiki/Entity_component_system), which is explicitly designed with composition in mind.
* A tiny nitpick, but `Random.Shared` is available; no need to new up your own `static` instances of `Random`

tomspilman:
@SimonDarksideJ is this something we want to fix and get merged?

SimonDarksideJ:
I'll review and look into it, thanks for the heads up @tomspilman

SimonDarksideJ:
@Forged1212 , we need to move this to the new MonoGame documentation site and update it to address the review comments above.
Would you prefer me to migrate it or can you submit a PR to the newer https://github.com/MonoGame/monogame.github.io site ?

Forged1212:
@SimonDarksideJ If you can I would not mind if you migrate it to the new repo. I however don't think I'll have time to address the feedback at the moment.

SimonDarksideJ:
No worries, happy to handle it.

SimonDarksideJ:
Can you close this PR please @Forged1212 while I migrate it over, just leave your branch alone (don't delete), although I have already cloned it.

PR #7917, opened 2022-08-22
Set the SupportedOSPlatformVersion within the Android project file

### LINKED ISSUE #7896: Warning CA1416 when buildig for Android
Hi, there is alot of CA1416 warnings in build log when I try to build my project for Android.

`warning CA1416: This call site is reachable on: 'Android' 23.0 and later. 'XXX' is only supported on: 'Android' 31.0 and later.`
where 'XXX' is some entity from MonoGame like 'GamePadButtons.Back' or 'Game.Update(GameTime)'.

I think this happens because there is no `<SupportedOSPlatformVersion>23</SupportedOSPlatformVersion>` property in MonoGame.Framework.Android.csproj.

PS: Thank you fore releasing .Net 6 version, it is really cool.

- MonoGame 3.8.1.303

- macOS 12.4 ARM

- Android

Brett208:
Set the SupportedOSPlatformVersion within the Android project file to prevent large quantities of Warning CA1416 when targeting an API earlier than 31 on Android.

Resolves #7896.

I set the SupportedOSPlatformVersion to API 23 (Marshmallow V6.0, released Aug 2015, since it was mentioned in the issue and everything compiles ok. Might be good for someone to check that this is appropriate before merging though.

Image of warnings prior to implementing this change below:
image

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.Android.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.Android.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0-android</TargetFramework>
+    <SupportedOSPlatformVersion>23</SupportedOSPlatformVersion>
     <DefineConstants>ANDROID;GLES;STBSHARP_INTERNAL</DefineConstants>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <MGOpenGL>GLES</MGOpenGL>


espinosaluis:
I still get the same warnings after setting SupportedOSPlatformVersion to 23.

Brett208:
@espinosaluis could you confirm that you are setting `SupportedOSPlatformVersion` in MonoGame.Framework.Android.csproj, compiling MonoGame itself and still receiving the warnings? 

If you are adding `SupportedOSPlatformVersion` to your game's project file only (and not the MonoGame Android project), I believe you will still get the warnings.

-Brett

mrhelmut:
That must be something new because ```SupportedOSPlatformVersion``` wasn't for libraries, only for executables.

(The build failed, but it's not related to this PR. I'm figuring it out.)

espinosaluis:
> @espinosaluis could you confirm that you are setting `SupportedOSPlatformVersion` in MonoGame.Framework.Android.csproj, compiling MonoGame itself and still receiving the warnings?
> 
> If you are adding `SupportedOSPlatformVersion` to your game's project file only (and not the MonoGame Android project), I believe you will still get the warnings.
> 
> -Brett

Brett, I'm doing this only in my game's project file. Sorry I misunderstood.

espinosaluis:
Please double check it's really API 23. In my case the warnings I get are related to 21. I don't think we want to exclude more devices than needed by adding that.

`warning CA1416: This call site is reachable on: 'Android' 21.0 and later. 'Vector2.Y' is only supported on: 'Android' 31.0 and later.`

mrhelmut:
The bare minimum supported by MonoGame is API 23 (Android 6) because it is the minimum suggested by the Google Play Store (otherwise users will get a warning when starting the app that it is possibly not compatible).

Target should be level 31 or above, as anything below will be refused by google.

Brett208:
@espinosaluis The compiler warning is just pulling the minimum API you manually set to target in your project files (it is set in the .csprog file and another XML file for I think the app store). I had set the minimum API in my project to 23 hence seeing 23 in the warning. Try changing the API you tell VSyou support to a different API and I suspect you will see the new number follow in the warning.

This PR is trying to correct this warning to only emit if you target a min version of Android less than 23 which is (I think) the minimum possible supported by MonoGame 3.8.1. currently the compiler is falsely warning that you app must target a min of 31.

Hope that helps.

espinosaluis:
> @espinosaluis The compiler warning is just pulling the minimum API you manually set to target in your project files (it is set in the .csprog file and another XML file for I think the app store). I had set the minimum API in my project to 23 hence seeing 23 in the warning. Try changing the API you tell VSyou support to a different API and I suspect you will see the new number follow in the warning.
> 
> This PR is trying to correct this warning to only emit if you target a min version of Android less than 23 which is (I think) the minimum possible supported by MonoGame 3.8.1. currently the compiler is falsely warning that you app must target a min of 31.
> 
> Hope that helps.

Thanks, Brett. That makes sense.

Brett208:
Rebased this PR to the latest development branch commit. This should enable passing all CI tests.

PR #7932, opened 2022-09-17
Improved documentation for Ubuntu

rejurime:
* The [ that was inside the url was removed.
* In minimal installations of Ubuntu curl is not installed, which is used internally in the script.


harry-cpp:
Looks good, thanks @rejurime !

Issue #6144, opened 2018-01-11
Proposal to switch from local dependency references to NuGet packages.


vpenades:
Right now, all third party dependencies required to build MonoGame SDK are located in a "dependencies" repository and so far, this architecture has worked well.

But right now, developers are moving towards NuGet packages as the preferred mechanism to reference third party dependencies. We will have different feelings whether this is a right or wrong approach, but certainly, it's a train that cannot be stopped anymore.

More and more developers expect MonoGame to be available as NuGet packages. But in order to offer proper nuget packages, MonoGame also needs to be part of the ecosystem, which means that MonoGame should consume nuget packages when applicable.

There's already a number of third party dependencies that are available as NuGet packages, in fact, some dependencies are ONLY available as NuGet packages, SharpDX being the most notorious case, as more projects in which we depend move to NuGet, this problem will only get worse.

In the early days, a lot of work was done to provide a specific cross platform solution for MonoGame, when nothing else was available, but IDEs and NET itself have finally catched up to provide better and more powerful cross platform solutions. I understand some developers might want to work the traditional way, everything with local dependencies, etc, but not supporting NuGets is not an option anymore, and trying to keep MonoGame SDK being able to build "the old fashioned way" and "with nugets" is consuming a lot of time from the lead developers.

So my proposal is to write here some sort of roadmap for a progressive transition form local dependencies to nuget packages:

1. Identify the current problems that prevent referencing nuget packages in monogame SDK.
2. Identify which local third party dependencies can be switched to nuget packages.
3. progressively replace these dependencies, one by one.

Other issues already mentioning the Nuget package issue: #5724 #5565 #5262

Jjagg:
I think this is a good idea. It makes upgrading managed dependencies a lot easier too. To upgrade SharpDX I used a script to pull the nugets. Would be way easier to just reference the NuGets directly.

Jjagg:
@tomspilman Responding to [this comment](https://github.com/MonoGame/MonoGame/pull/6102#issuecomment-357744176) from #6102.

> Mostly because updating our dependencies is a huge pain in the ass and we constantly have issues with it. 

This issue would be resolved by switching to NuGet. MG's dependency management is annoying because dependencies don't have a consistent workflow to upgrade and they live in a seperate repository. With NuGet upgrading is as simple as changing a version number in the main repo.

>  I just like having access to the actual code/binaries used in our builds

[...]

PR #7933, opened 2022-09-22
Use SharpDX and NVorbis from Nuget

rejurime:
Based on #7089 and #6144 I made changes to use SharpDX and NVorbis with Nuget.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "ThirdParty/Dependencies"]
 	path = ThirdParty/Dependencies
 	url = https://github.com/MonoGame/MonoGame.Dependencies.git
-[submodule "ThirdParty/NVorbis"]
-	path = ThirdParty/NVorbis
-url=https://github.com/NVorbis/NVorbis.git
 [submodule "ThirdParty/SDL_GameControllerDB"]
 	path = ThirdParty/SDL_GameControllerDB
 	url = https://github.com/gabomdq/SDL_GameControllerDB.git
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
@@ -50,12 +50,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Reference Include="SharpDX">
-      <HintPath>..\ThirdParty\Dependencies\SharpDX\net40\SharpDX.dll</HintPath>
-    </Reference>
-    <Reference Include="SharpDX.D3DCompiler">
-      <HintPath>..\ThirdParty\Dependencies\SharpDX\SharpDX.D3DCompiler.dll</HintPath>
-    </Reference>
     <Reference Include="CppNet">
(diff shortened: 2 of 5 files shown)


damian-666:
The lastest sharpDX is in nuget is 4.2 not 4.01..  i'm confused because the compiler told in vs2022  preview warned me i should use sharpDX.xinput 4.2 at some point, building my own app.. ( i cant remember when or why,  analyser or warning)  its  a net6.windows with  wpf,     i'm not using the develpbranch, usng the MG nuget release 3.8.1.   i'm using wpfinterop and it works now..  

 I do the same thing to my net6.0 app.. add nuget refs to the EXE  provject that update to 4.2... for all the parts xaudio2, xinput, mediafoundaton, dxgi, direct3d11, direct2d1.... then it  will replace the mg dependencies with those versions, and in the output folder.. mabye there is a roll forward way to do this...

image

rejurime:
Hi @damian-666, Although the latest version of Sharpdx is 4.2.0, what I did in this specific PR was to replace the way of using it in the project (by using Nuget). I didn't want to update the Sharpdx version to the last one since that would imply doing more tests. I also checked which were the versions that the project is currently using (as you can see here https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/MonoGame.Framework.WindowsDX.csproj#L74). My idea is to update all the dependencies that MG already has from SharpDX to its latest version in a second PR, so I can test it better.

damian-666:
Good idea. Honestly I don't remember why it gave me a warning that went away when I put the latest reference. ..the wpf interop from craftwork works either way , for me now ,with  unmerged PR sources.  But I'm not testing w Xbox sdk.   Right now it's fine  either way..but there's an older netstandard dependency  in the referenced version might be an issue..

rejurime:
Hi! I want to check if I need to change something in this PR or if there are any feedback. Regards.

rejurime:
Hi! Is there anything blocking this being merged? @mrhelmut can you help me? thanks!

mrhelmut:
I am not maintaining MG lately and am taking some distance. I'd gladly leave my seat to someone with fresh ideas but it is not my call to make.

rbwhitaker (commented):
In the OggStream file, what's the difference between the old `DecodedTime` and the new `TimePosition`? Why the change? They look like they might be conceptually the same thing, but I assumed the switch to NuGet would have put us on a version that matched what we had before, which should have the same API, and wouldn't have needed changes like this. (But I'm only asking for my own edification. I don't think this represents an actual problem.)

TechPizzaDev:
The name change of the DecodedTime property was just arbitrary cleanup.
And it is recommended to use up-to-date NVorbis because old versions are quite buggy when it comes to seeking.

rejurime:
I didn't know @mrhelmut sorry 😞.
Hi @rbwhitaker I made the change based on https://github.com/NVorbis/NVorbis/blob/519d4e2aae7d6a4d5bab552ec5c1e517e9c78855/NVorbis/VorbisReader.cs#L195.
@TechPizzaDev I also think that is best to upgrade versions, but I wanted the PR to be small and without many changes, just to switch from Dll to Nuget dependencies. My idea is to upgrade both the version of SharpDX and NVorbis in future contributions.

harry-cpp:
This looks good to me, many thanks @rejurime, merging!

PR #7936, opened 2022-10-01
Added Integer Vector Classes and Color Operators

delinx32:
Simple addition of Integer vectors for operations that require more precise accuracy, such as grid positioning where floats may not be desirable.

Also added * and / operators on Color to simplify some operations


benjitrosch (commented):
What's wrong with using `Point` for the scenario you've described?

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Point.cs

delinx32:
> What's wrong with using `Point` for the scenario you've described?
> 
> https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Point.cs

To my knowledge there are no Point3 or Point4 classes.  In addition, the vector classes add more math functionality than point.  Also, naming them "vector" gives more consistency in the code base for similar operations.

People porting code from unity might expect there to be integer vectors as well which is the primary reason I want them.

delinx32:
> Just my opinion but I'm not sure those reasons are compelling enough. Floats make sense for what a Vector describes and works with all of the graphics API's that expect floats anyways.
> 
> If you require integer Vector classes for the specific use case of porting from Unity it might be worth keeping in your own project, or you could write extension methods for Point. But I'm not sure this belongs in the main repo.
> 
> Also to your last edit, the `Point` class has a `ToVector2()` method.

Ints perform better than floats, and floats have questionable equality when comparing whole numbers which leads to extra math operations to ensure equality.  Again, there is no Point3 or Point4 class so Point isn't really an all encompassing solution.  

Godot, Unity, and Unreal all have this functionality.  I can keep maintaining my fork, but it seems like a no-brainer to add this as there's no cost and it adds a feature that people will use.

mrhelmut:
It sure adds features but it makes things misleading regarding Point vs Vector. We will likely need a different naming convention when it comes to MonoGame, which will maybe involve renaming Point to Point2 (which is a breaking change).

delinx32:
> It sure adds features but it makes things misleading regarding Point vs Vector. We will likely need a different naming convention when it comes to MonoGame, which will maybe involve renaming Point to Point2 (which is a breaking change).

I'm not sure what this has to do with Point.  Benji suggested using a Point.  This PR has nothing to do with Points.  I was "Point"ing out that Point wouldn't work because there are no Point3 or Point4 classes.  Points and Vectors are 2 different things.

Point represents a point in space.

Vector represents a magnitude and direction (as a float)

This pull request adds Vectors that are ints when you don't need the precision of floats.  This allows them to perform better and be more accurate.

I'm really kind of amazed at the pushback on this.  There is a reason that other game engines have added integer vectors.

lithiumtoast:
Hey @delinx32, I had this conversation over at MonoGame.Extended many many many times. 

I was once upon a time in agreement with you; a point and a vector are different things! You can [read my explanation here which is in your favour](https://github.com/craftworkgames/MonoGame.Extended/issues/510#issuecomment-398434410). 

However, over time I stopped caring about principles because it's more pragmatic to think of any point as the vector from the origin when coding. There is less types, less casting, less code to maintain, less less less etc. Less is more.

Ravendarke8:
Nothing better than vector you can't normalize... or vector where .w is basically worthless. Extremely niche thing that would be bloat for framework while being acceptable for engine. Hard no from me.

stromkos:
Color operators
---

I am not sure about the changes to `Color.cs` with the operators.  

The basic rules of algebra fail to hold.(The same could be said of the precision of floats, but the impact is much greater with ints as the interoperation precision limit)

There is ambiguity created by including the Alpha component in the operations.  

` Color.White * .1f` results in high transparency  instead of dark gray.

The same goes for excluding the alpha:
`Color1 / 2 + Color2 / 2` 

Produces a close to correct, 2 odd values will be off by 1, blending result, whereas:

 `(Color1 + Color2) / 2` Desaturates the lighter colors due to clamping on the addition.

If you want to use mathematical operations on colors use the

 'color.ToVector3()' or 'color.ToVector4()'  instance functions, then

Perform your operations and convert back to a color by passing the vector to the appropriate constructor.

Grid alignment and floats
----

I use grids in my code all of the time as a representation of 32x32 pixels:
to get the integer grid flattened id for a stride of 25 (800 pixels wide):

```
Vector2 Position;
//...

// for integers: >> 5 is a very fast version of / 32

int x = (int)(Position.X + .5f)  >> 5;  // .5f for rounding errors
int y = (int)(Position.Y + .5f)  >> 5;  //  use 15.9f for centering on grid
int id = x + y  * 25;
```

Alignment checks:

Errors accumulated during floating point operations increase over time.

To combat this use the following check occasionally:
Note: 1/8 pixel window for this code.
```
Vector2 intPosition = new Vector2((int) (Position.X + .0625f) , (int) (Position.Y + .0625f));
if ((Position - intPosition).LengthSq() < .015625f)
    Position = intPosition;
```

PR #7937, opened 2022-10-01
Update 3_understanding_the_code.md

simoneddeland:
Added some missing information about the code in the update and draw methods.

### DIFF
--- a/Documentation/articles/getting_started/3_understanding_the_code.md
+++ b/Documentation/articles/getting_started/3_understanding_the_code.md
@@ -107,7 +107,7 @@ protected override void Update(GameTime gameTime)
 }
 ```
 
-The **Update** method is called multiple times per second, and it is used to update your game state (checking for collisions, gathering input, playing audio, etc.).
+The **Update** method is called multiple times per second, and it is used to update your game state (checking for collisions, gathering input, playing audio, etc.). The if statement allows the game to be exited by pressing either the escape key on a keyboard or the back button on a gamepad which is especially useful if you change your game to run in fullscreen mode.
 
 ## Draw Method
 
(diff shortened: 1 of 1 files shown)


Documentation/articles/getting_started/3_understanding_the_code.md
```diff
@@ -107,7 +107,7 @@ protected override void Update(GameTime gameTime)
 }
 ```
 
-The **Update** method is called multiple times per second, and it is used to update your game state (checking for collisions, gathering input, playing audio, etc.).
+The **Update** method is called multiple times per second, and it is used to update your game state (checking for collisions, gathering input, playing audio, etc.). The if statement allows the game to be exited by pressing either the escape key on a keyboard or the back button on a gamepad which is especially useful if you change your game to run in fullscreen mode.
```
Mindfulplays, line 110: I wonder if it is better to use `Update` instead of **Update** (applies to `if`, `Clear` etc ...)?

rbwhitaker: You're referring to the `code` type of markup instead of **bold**?

Personally, I'd _much_ prefer what you're suggesting. But using bold is at least consistent with the rest of the file here. Aside from the full blocks of code, I don't see any other examples of using `this style` in this article.

Mindfulplays: Yep, I was suggesting the code markup, but consistency wins though.

SimonDarksideJ: An interesting thought, but it would mean updating ALL of the other documentation to match.

tomspilman:
@SimonDarksideJ should we merge this or not.

SimonDarksideJ:
Will review in the documentation update to the new docs site
https://github.com/SimonDarksideJ/monogame.github.io/issues/11

Issue #7488, opened 2021-04-02
Hash key not recognised on UK keyboard


AliciaRaven:
The hash / pound key '#' returns as Key.None for EnGB keyboards

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:
- DesktopGL

MLSTRM:
Bumping this.
From some research of my own I believe it's probably mapped as VK_OEM_10 (or equivalent). Confusingly I would have almost expected it to map as the VK_OEM_102 key based on it being the additional input in UK vs US layouts but that doesn't seem to be the case (as that key code is correctly mapped). At the same time it might be nice to update the docs to help decode the names of the OEM keys to UK keyboard users as currently the naming matches the US scancode mapping whereas the buttons are mapped differently for UK layouts. (i.e. OemTilde is actually the `/¬ key, and OemPipe only gives a Pipe when shift modified, otherwise is backslash).
Interestingly trying it out in the microsoft keyboard layout creator, the key does come up as VK_OEM_7 but doesn't map accordingly in MonoGame at current, I would have potentially assumed that GetKeyboardState would be after layout mapping but I guess it's just at raw scancode level.

I will try and turn this musing into a pull request at some point in the near future, but thought I would leave the notes here in case someone else is willing/able to get to it before I can.

PR #7945, opened 2022-10-20
Add SDL mapping for Key 35 (102 key backslash or UK pound key)

MLSTRM:
See #7488 
Adding this key entry as Oem8 due to lack of suitable existing mappings in the Oem keys and I wasn't sure it was right to necessarily add a new one for this use case.

The UK keyboard #/~ key maps as key code 35 in SDL due to its internal conversion to ascii, however there is no OEM Pound virtual key, and in native UK keyboard mappings this key maps to OEM 7 (which is OEM Quote in mono's mappings).
By Mapping this key to Oem8 we can allow support for this key in DesktopGL for 102 key users, however it may need to be checked/verified with someone that has a US 102 key layout keyboard to ensure the mapping comes through correctly.


stromkos:
Less reliance needs to be made on SDL key mappings. They are not cross platform and skip many non-US keys.

PR #7946, opened 2022-10-24
Corrected images with the right URLs

TejusWadbudhe:
Now images can be seen and when clicked it opens the right image


tomspilman:
Not relevant anymore.

Issue #7897, opened 2022-08-02
[Mac/iOS] Content does not seem to copy from shared project


Dokug:
After upgrading my existing solution to MonoGame v3.8.1 (from 3.8) I found that my setup with a dedicated iOS/Mac/Windows project and a shared MonoGame library does not seem to work anymore for iOS and Mac.
It seems as if the .xnb files created by mgcb do not get copied to the correct directory, as the LoadContent() method throws a FileNotFoundException() when executing my guard logic against missing files. (Basically the file Content/[...].xnb is not found.)

Am I missing a step in the upgrade guide?
I added a new iOS project (that is basically empty anyway, as the whole thing uses the shared library), which I generated from the template.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1.303

#### What operating system are you using:
- MacOS

#### What MonoGame platform are you using:

- iOS

AndreyFromGomel:
I have the same issue for iOS project, but macOS and Android projects works well for me.

kurtcz:
Same for me, iOS project is unable to find any xnb files - I am using the MG 3.8.1.303 templates and MGCB editor. The files are not getting copied to the output folder

damian-666:
what  i do in the meantime (i dont have ios licnese. so cant help) is  I just  use the mgcb editor and build everything and copy the xnb files directly and move to other bin folders that all got moved around with fresh project files i had to make, as this gets sorted out for the various build tools being used.   but have similar problems with a legacy level editor , that used .netstandard 2 and moving that to net6 really introduct sublte weirdness.  I just  use the mgcb editor and build everything and copy the xnb files directly and move to other bin folders that all got moved around with fresh project files i had to make, as this gets sorted out for the various build tools being used.   check the paths that the .net libs the new project files simpler, but you lost your old setup.      im havign a ton of problesm and dont have an ios license so i cant build and trace MG and help or seee why thsi happened.. it works for me but i used to just build and embedded xnb files like nez does, as embedded resources in the shared DLL, but then the copy and build of content, works for osx and GL and windows netcore, so i stopped dong that because i wanted shader transpiling.      ****The shared project should be part of the test suilte.****  right now even the sample is not updated.. def needs help but im swamped.

inkcomic:
> Same for me, iOS project is unable to find any xnb files - I am using the MG 3.8.1.303 templates and MGCB editor. The files are not getting copied to the output folder. Android project is able to load them just fine.

[...]

PR #7952, opened 2022-11-06
Added CollectBundleResourcesDependsOn in target to fix BundleResource…

corykroll:
#7897 
Adding CollectBundleResourcesDependsOn to .targets file to get Content to be added properly to iOS .app files.
Would love to see this go out as a hotfix for the NuGet package MonoGame.Content.Builder.Task if approved so users could utilize iOS Content pipeline using default templates out of the box.


RCGame:
This works like a charm. I built my local NuGet for time being.

PR #7955, opened 2022-11-19
Fix file drag&drop crash in WindowsDX

Metaray:
Fixed drag&drop message handler in WindowsDX calling DragQueryFile with incorrectly sized buffer.


harry-cpp:
I've merged another PR which fixed the issue: https://github.com/MonoGame/MonoGame/pull/7998 😅, great work non the less @Metaray !

PR #7961, opened 2022-11-22
Added code to .targets to fix iOS Content inclusion

corykroll:
#7897 
Fixes an issue of Content not copying into an iOS build when compiling/debugging.  This fixes issues when loading content on iOS when running `Content.Load<T>` due to the content not being there.
Would be helpful to redeploy the MonoGame.Content.Builder.Task NuGet package with this fix to allow iOS MonoGame template to use content.


harry-cpp:
Tabs instead of spaces, but I guess I can fix that manually, merging!

Issue #7868, opened 2022-07-28
SpriteFonts failing to load (iOS)


Lofionic:
SpriteFonts failing to load in iOS:

`File.spriteFont` is default SpriteFont file generated by MGCB editor.

`Content.mgcb` entry:
```
#begin File.spritefont
/importer:FontDescriptionImporter
/processor:FontDescriptionProcessor
/processorParam:PremultiplyAlpha=True
/processorParam:TextureFormat=Compressed
/build:File.spritefont
```

Call:
`SpriteFont foo = Content.Load<SpriteFont>("File");`

Error:
`"Could not find ContentTypeReader Type. Please ensure the name of the Assembly that contains the Type matches the assembly in the full type name: Microsoft.Xna.Framework.Content.ListReader`

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1.303

#### What operating system are you using:
- MacOS 12.4 / M1 / VS2022 

#### What MonoGame platform are you using:

- iOS

mrhelmut:
This is a reflection issue when using AOT compilation, you might want to disable trimming and/or assembly-level linking, or configuring the trimmer to not trim ```Microsoft.Xna.Framework.Content.ListReader``` (and likely other Reader types).

corykroll:
I've been digging through this one, and for me, the issue was in ContentTypeReaderManager.cs
When deploying to a physical iPhone for testing, this line of code is an issue:

`private static readonly bool _isRunningOnNetCore = Type.GetType("System.Private.CoreLib") != null;`

It does not work correctly.  Changing it to:

`private static readonly bool _isRunningOnNetCore = true;`

Does the trick.  This is because later on, if _isRunningOnNetCore is incorrect, mscorlib is attempted to be loaded, fails, and the app crashes.

Ultimately detecting if running on NET Core is something I'm trying to code a fix for.  I'm thinking a quick method to parse the string of RuntimeInformation.FrameworkDescription might do the trick.

mrhelmut:
Duplicate of #7872

PR #7962, opened 2022-11-22
Removed NET Core checks

corykroll:
#7868 
When deploying to physical iOS devices, errors are occurring due to code that tried to detect NET core environments or not and replace names of libraries to load.  This shouldn't be necessary anymore, as the updated templates all use NET 6 and use 'Core' libraries.  I've tested this removal on all types of projects, including WindowsUniversal, DirectX, OpenGL, Android and iOS projects, and deployed to Android and iOS physical devices and found this fix works across those scenarios.


mrhelmut:
We still need a backward compatibility with the .NET 4 runtime on consoles. A more appropriate take is on #7895

corykroll:
Understood on that - wasn't aware.  Appreciate the feedback.

Issue #7959, opened 2022-11-21
Unable to Open MonoGame.Framework.WindowsUniversal Solution or Project Reference MonoGame.Framework.WindowsUniversal Project


corykroll:
Upon trying to open the MonoGame.Framework.WindowsUniversal.sln file in Visual Studio 2022, I get an error saying 'One or more projects in the solution were not loaded correctly.'.  The Output window displays the following error:

`...\MonoGame\MonoGame.Framework\MonoGame.Framework.WindowsUniversal.csproj : error  : The project file cannot be opened. C:\Program Files\dotnet\sdk\6.0.402\Sdks\MSBuild.Sdk.Extras\Sdk not found. Check that a recent enough .NET SDK is installed and/or increase the version specified in global.json.`

This is odd since there doesn't exist a MSBuild.Sdk.Extras version 6.0.402 package on NuGet.

Additionally, when trying to link up the Universal Windows solution to the csproj file directly, I additionally get the same error.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- WindowsUWP

mrhelmut:
Cleaning this as UWP is getting deprecated #8406

PR #7963, opened 2022-11-22
Moved MSBuild Sdk version to fix Project Loading

corykroll:
#7959 
Moved version of MSBuild out of global.json file and into project.  Fixes loading of solution and also allows for reference directly from the MonoGame UWP template (where is currently doesn't load in VS 2022).


stromkos:
What effect would this PR have on the CLI and VS Code compilation?

harry-cpp:
Thanks @corykroll! Merging!

Issue #7965, opened 2022-11-26
iOS ExtendedGamepad ButtonOptions Error


corykroll:
When debugging a MonoGame iOS game, doing a GamePad.GetState crashes the game.
In GamePadState.PlatformGetState,  controller.ExtendedGamepad.ButtonOptions has a null state, and the call `controller.ExtendedGamepad.ButtonOptions.IsPressed` results in a null exception.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- iOS

PR #7966, opened 2022-11-26
Added null check for ExtendedGamepad.ButtonOptions for iOS compatibility

corykroll:
When debugging iOS MonoGame, doing a GamePad.GetState was crashing out the program due to a null state for ExtendedGamepad.ButtonOptions.  This addresses the issue by adding a null check.

Addresses issue #7965


harry-cpp:
Looks good! Merging!

mrhelmut:
This is okay for iOS, but we should mind the fact it uses C# 5+ language features and it might open the door to more of them.

It also appears that this PR did not run the build checks to verify that.

PR #7983, opened 2023-01-28
[SDL] Fixed SdlGameWindow fails to load Icon.bmp.

shin1m:
It occurs when the namespace of the entry point class is not the same as the name of the entry assembly.
This is also the case for the top-level statements, which generates an entry point class with no namespace.

Changed to use the name of the entry assembly for the resource name of Icon.bmp.

Also changed to use Type.Assembly for the fall back case, which is the recommended way when available instead of
Assembly.GetExecutingAssembly().


harry-cpp:
Looks good, thanks @shin1m, merging!

PR #7989, opened 2023-02-16
Restore Color.TransparentBlack

TeamDoodz:
The static property `Color.TransparentBlack` was removed in version 3.8.1 due to it being deprecated. This is a violation of the [semantic versioning specification](https://semver.org/):

> Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.

Although the functionality was deprecated, removing it is still a backwards-incompatible change and shouldn't occur in a minor or patch release.

This PR simply restores the property along with its `Obsolete` attribute.

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -18,6 +18,7 @@ namespace Microsoft.Xna.Framework
     {
         static Color()
         {
+            TransparentBlack = new Color(0);
             Transparent = new Color(0);
             AliceBlue = new Color(0xfffff8f0);
             AntiqueWhite = new Color(0xffd7ebfa);
@@ -429,6 +430,15 @@ namespace Microsoft.Xna.Framework
         }
 
         #region Color Bank
+        /// <summary>
+        /// TransparentBlack color (R:0,G:0,B:0,A:0).
+        /// </summary>
+        [Obsolete("Use Color.Transparent instead. In future versions this method can be removed.")]
+        public static Color TransparentBlack
+        {
+            get;
+            private set;
+        }
         
         /// <summary>
         /// Transparent color (R:0,G:0,B:0,A:0).


Apostolique:
Where did you see that MonoGame strictly adheres to semantic versioning?

Beyley:
> The static property `Color.TransparentBlack` was removed in version 3.8.1 due to it being deprecated. This is a violation of the [semantic versioning specification](https://semver.org/):
> 
> > Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.
> 
> Although the functionality was deprecated, removing it is still a backwards-incompatible change and shouldn't occur in a minor or patch release.
> 
> This PR simply restores the property along with its `Obsolete` attribute.

re-introducing it would also break semantic versioning as you are making a breaking change (adding a property) which may break some reflection code that expects that property to not be there, meaning this PR could not be shipped in a patch release either, and since this is deprecated anyway, it shouldnt still exist after a non-patch release

TeamDoodz:
> Where did you see that MonoGame strictly adheres to semantic versioning?

I guess I just assumed that since MonoGame uses three numbers for its versioning and semantic versioning is the most popular versioning scheme, that MG used semantic versioning; I may not be correct in that regard. Still, I believe that at least some attention should be paid to keeping API changes backwards-compatible.

I mainly made this PR because the removal of the property broke the Tiled content importers in `MonoGame.Extended.Content.Pipeline`, and I believe that it is a bit unfair to assume that developers have the time to migrate away from deprecated functionality only one minor version away from removal.

nkast:
> I mainly made this PR because the removal of the property broke the Tiled content importers in `MonoGame.Extended.Content.Pipeline`,

The issue has been fixed on Extended.
https://github.com/craftworkgames/MonoGame.Extended/pull/735

TeamDoodz:
I did not know about that, thanks for bringing it to my attention. I still believe that the removal of functionality should come during major releases only, but I'm not going to die on that hill; it is, after all, a single static property in question.

stromkos:
Why was this removed?  as its removal violates compatibility with XNA. This only makes sense in a non-premultiplied alpha environment.  Which may have not been a possible result in XNA, but it was offered, and should be supported until MonoGame breaks from XNA in 4.0.

tomspilman:
This was an XNA 3.0 type that was removed for XNA 4.0.  We don't plan to add it back.

PR #7990, opened 2023-02-20
Added instructions for targeting M1.

davidhesselbom:
Maybe it's obvious to everyone else, but I kept trying to build for M1 using the old `osx-x64` command since MonoGame is supposedly able to build M1 native binaries now with no extra work needed.

I don't know if anything else needs changing to publish for M1, but the example Info.plist file says

```xml
    <key>LSMinimumSystemVersion</key>
    <string>10.15</string>
```

and I take it that number should be something higher for M1 apps?

### DIFF
--- a/Documentation/articles/packaging_games.md
+++ b/Documentation/articles/packaging_games.md
@@ -28,9 +28,13 @@ We recommend using the .tar.gz archiving format to preserve the execution permis
 
 ### Build and packaging for macOS
 
-From the .NET CLI:
+From the .NET CLI (x64):
 
 `dotnet publish -c Release -r osx-x64 /p:PublishReadyToRun=false /p:TieredCompilation=false --self-contained`
+
+If targeting M1:
+
+`dotnet publish -c Release -r osx-arm64 /p:PublishReadyToRun=false /p:TieredCompilation=false --self-contained`
 
 We recommend that you distribute your game as an [application bundle](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html). Application bundles are directories with the following file structure:
 


stromkos:
Does it not build and run without these changes?

davidhesselbom:
Honestly, I haven't been able to run anything I've built on Windows on a Mac yet. Maybe it's because I don't have a certificate to sign the binaries with (I'm a hobbyist, not a professional), or maybe because I put files in the wrong folders inside the application bundle (the instructions aren't clear on where the DLLs and all the other non-content files should go).

Either way, the binary built with the `osx-arm64` flag is shown as an Apple Silicon binary, while a binary built with the `osx-x64` flag will be shown as an Intel binary. This suggests that in order to build a native M1 binary, you do need to use the `osx-arm64` flag. Not that I know whether it makes any difference for performance (or otherwise).

Documentation/articles/packaging_games.md
```diff
@@ -28,10 +28,14 @@ We recommend using the .tar.gz archiving format to preserve the execution permis
 
 ### Build and packaging for macOS
 
-From the .NET CLI:
+From the .NET CLI (x86):
```
harry-cpp, line 31: Typo, was supposed to be `x64`.

harry-cpp (changes requested):

Documentation/articles/packaging_games.md
davidhesselbom: Lol yeah, sorry!

harry-cpp:
Actually looking at the rest of instructions this isn't enough, you would just end up making an executable for either the x64 or arm64, not both. I'll look into updating the instructions later this week.

mrhelmut:
Yep, to fully support macOS to modern standard, you have to produce both an x64 and an arm64 executable and fuse them into a universal executable.

This is a step that .NET doesn't support and doesn't intend to support for the foreseeable future, so this will imply a command line call to ```lipo``` to merge them.

harry-cpp:
New fixed instructions PR is up: https://github.com/MonoGame/MonoGame/pull/8022

Issue #4149, opened 2015-09-24
SpriteFont Styles Not Working


borealwinter:
The SpriteFont does not display any styles other than regular;

cra0zy:
Can confirm, will try to look into this tomorrow.

repstyle:
I have the same issue, How can we fix it ?
VS 2013 . Monogame 3.4.

tomspilman:
If I remember correctly it is because Freetype doesn't have the features that the Windows font system employs to generate them.  If you look at my post from 2014:

https://github.com/mono/MonoGame/issues/2741#issuecomment-51572666

> The bad news is that there is still one major issue... we don't support bold and italic styles. The Windows GDI will render a bold or italic font automatically, but FreeType has no such feature... so we need to invent one.

My understanding is that the Windows font system does the following when trying to resolve for a bold or italic font style:

 - First see if the TTF includes the style itself... few fonts do this.
 - Second look for another TTF in the same family that includes that style.
 - Lastly if all else fails Windows generates the style manually in its rendering engine.

Unfortunately Freetype does not have any features like this.  If we want these to work we would need to implement these three steps to get the same behavior in our font processor.

repstyle:
Ok, i was using Verdana font (TrueType).
I have found a solution : Instead of using the style, i use "Verdana Bold" instead of "Veradana" for the font name.It's working.
So if you want to use styles for a font, in fact you will have to get a font for regular, one for bold, one for italic etc
It's the most clean and simple way to do it.

tomspilman:
> It's the most clean and simple way to do it.

I agree...  just provide the right font name is a simple approach.  However people expect that setting a style works for fonts, so I think we still need to provide it.

repstyle:
I totally agree with you.

Tzenchor:
Just wanted to comment that i'm having a look at this.
I think I have had success with the option of looking for font files in the same family as the requested ones in Windows, but I need to test it a bit further.
About the option 1 (looking if the font has the style included), i'll need to read the font, but it feels wrong to use SharpFont outside `SharpFontImporter.cs`, so i'm not sure if I should move the font finding steps from `FontDescriptionProcessor.cs` to `SharpFontImporter.cs` or use SharpFont directly in `FontDescriptionProcessor.cs`.
Option 3 (Rendering the style manually) will be interesting to make italics

wldevries:
This was fixed in the PR linked above and then #8570 caused a regression. This no longer works.

[...]

PR #7993, opened 2023-02-25
Match Spritefont style on Windows

wldevries:
Implements #4149 by parsing TrueType fonts to extract their font family and font sub family (style) names.

The content pipeline speed is not affected much, since TrueType fonts parsing is quite simple and super fast. The font family is already filtered on using information in the Windows registry.

There are many more font sub families in fonts. Calibri for instance has Light, Light Italic and Bold Italic. The Bahnschrift font in Windows has even more: Condensed, SemiCondensed, SemiBold, etc... It might make sense to treat the Style property as a string instead of an enum with fixed values.


stromkos:
I do not know of "RoyT.TrueType", but I am opposed to adding another requirement when the "issue" could be solved internally in a cross platform nature.

SimonDarksideJ:
Are you able to resolve the merge conflicts for a final review @wldevries ?

wldevries:
Sure can.

An alternative to https://github.com/roy-t/TrueType could be https://github.com/LayoutFarm/Typography which seems to have more active development and features.

SimonDarksideJ (approved):

PR #7996, opened 2023-02-26
Fix spritefont documentation in localization.md

wldevries:
Fixed two issues:
* resx files are loaded relative to the .mgcb file, not the .spritefont file
* added documentation to change the asset type to LocalizedFontDescription

### DIFF
--- a/Documentation/articles/content/localization.md
+++ b/Documentation/articles/content/localization.md
@@ -68,7 +68,7 @@ To make use of this functionality you need to tell the spritefont which resx fil
 </ResourceFiles>
 ```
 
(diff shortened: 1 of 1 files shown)


rbwhitaker:
I like this change (though I don't use the `LocalizedFontDescription` enough to confirm that all of the documentation here is correct, from just the quick glance that I've given it), but I did want to point out a typo in the revised text:

> above the .mgcb prjoect file.

That should be "project".

tomspilman:
The documentation has been moved to the website repo:

https://github.com/MonoGame/monogame.github.io/blob/main/articles/content_pipeline/localization.md

So this PR needs to be resubmitted to that repo.

PR #7999, opened 2023-03-08
Added GenerateDocumentationFile to .net 6 projects

skyebird189:
Currently, MonoGame on nuget does not seem to ship with XML comments.

I've added `GenerateDocumentationFile` to all the .NET 6 projects so that XML documentation should now be generated on push to nuget.


harry-cpp (changes requested):
Oh, I thought this was true by default, how curious :D 

This change should be applied to the Directory.build file instead of the project files:
- https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Directory.Build.props

And also to the content pipeline project as well:
- https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj

PR #8010, opened 2023-04-09
Removed blank spaces in Program.cs in two templates

Meteoroetem:
The ```Program.cs``` file in the DesktopGL and WindowsDX templates had unnecessary blank lines at the start.
They just really bothered me.


Beyley:
for future reference: you can add *more* commits to a PR, you dont have to make multiple

Issue #8005, opened 2023-03-29
`LoadContent` causing `NullReferenceException` when publishing with NativeAOT enabled


seclerp:
Fresh MonoGame project is failing to load content after publishing with `<PublishAot>true</PublishAot>` property enabled with the following error:

```csharo
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Xna.Framework.Content.ContentTypeReaderManager.LoadAssetReaders(ContentReader) + 0x938
   at Microsoft.Xna.Framework.Content.ContentReader.InitializeTypeReaders() + 0x40
   at Microsoft.Xna.Framework.Content.ContentReader.ReadAsset[T]() + 0x20
   at Microsoft.Xna.Framework.Content.ContentManager.ReadAsset[T](String, Action`1) + 0xa4
   at Microsoft.Xna.Framework.Content.ContentManager.Load[T](String assetName) + 0xa8
   at Game1.SampleGame.LoadContent() + 0x124
   at Game1.SampleGame.Initialize() + 0x60
   at Microsoft.Xna.Framework.Game.DoInitialize() + 0x60
   at Microsoft.Xna.Framework.Game.Run(GameRunBehavior) + 0xbc
   at Program.<Main>$(String[] args) + 0x48
   at Game1!<BaseAddress>+0x26bb08
```

Because I'm using M1, I'm also using 8.0 version of ILCompiler to target `osx-arm64`:

```xml
  <ItemGroup>
    <PackageReference Include="Microsoft.DotNet.ILCompiler" Version="8.0.0-*" />
    <KnownILCompilerPack Update="Microsoft.DotNet.ILCompiler"
                         TargetFramework="net7.0"
                         ILCompilerPackNamePattern="runtime.**RID**.Microsoft.DotNet.ILCompiler"
                         ILCompilerPackVersion="8.0.0-preview.3.23174.8"
                         ILCompilerRuntimeIdentifiers="linux-musl-x64;linux-x64;win-x64;linux-arm;linux-arm64;linux-musl-arm;linux-musl-arm64;osx-arm64;osx-x64;win-arm;win-arm64;win-x86"
    />
  </ItemGroup>
```

Any other compatible with the NativeAOT app is working fine with such a setup.

### System info

- System Version: macOS 13.2.1 (22D68)
- Kernel Version: Darwin 22.3.0
- Boot Volume: Macintosh HD
- Model Name: MacBook Pro
[...]

PR #8014, opened 2023-04-14
NativeAOT: Add `[ContentReadersConsumer]` attribute to enable preserving content readers while trimming

seclerp:
## About

While investigating the potential implementation for the proposal stated in https://github.com/MonoGame/MonoGame/issues/8005#issuecomment-1494415800 I found a quite appealing way of providing users with the one-liner ability to include content readers into the executable.

My proposed solution is to use a special meta-attribute with all necessary inner [`[DynamicDependency]` attributes](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?source=recommendations#dynamicdependency) specified.

## Usage

Attach attribute to a consumer assembly in the code:
```csharp
// AssemblyInfo.cs
[assembly: ContentReadersConsumer]
```

Or attach an attribute to a consumer assembly via `.csproj`:
```xml
<ItemGroup>
   <AssemblyAttribute Include="Microsoft.Xna.Framework.Content.ContentReadersAttribute"/>
</ItemGroup>
```

## Notes

It is also possible to add this (a bit boilerplate) code to the templates or embed it directly to a platform-specific NuGets. Still, I'm not sure about that as it will enforce more testing with different platforms adding a new layer of "magic" things to the background, which is not a great thing IMO.

### DIFF
--- a/MonoGame.Framework/Content/ContentReadersConsumerAttribute.cs
+++ b/MonoGame.Framework/Content/ContentReadersConsumerAttribute.cs
@@ -1,0 +1,56 @@
+using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.Xna.Framework.Content;
+
+/// <summary>
+/// Mark the target assembly as one that requires content readers assemblies at runtime.
+/// </summary>
+/// <remarks>
+/// By default, when using NativeAOT with MonoGame, trimmer will cut off content reader types from the native output, because they are used
+/// in the consumer project, but not by MonoGame itself. The attribute will tell the linker that all MonoGame-bundled content readers should be presented in the output.
+/// </remarks>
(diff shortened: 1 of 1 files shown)


seclerp:
cc @mrhelmut @kant2002

Ragath:
This would only protect the stock readers, while also severely limiting the amount of trimming projects would be able to achieve. Having a wrapper attribute like this available may not hurt, but I think a better solution is to just add the individual ones to the templates. Both to educate the consumer on their existance/purpose and to provide enough flexibility to remove the readers(and by extension their dependencies) that the project in question does not need.

mrhelmut:
I'll be closing this one for the time being as we're moving toward a different solution. See #8104 
Feel free to update the PR if you feel like it.

PR #8015, opened 2023-04-15
Fixes a bug in the property grid

rbwhitaker:
In the property grid, there is a little divider that allows you to resize the two columns. When you hover over it, you get a different cursor and can choose to drag it to either side. However, this region did not always extend to the full height of the control, leading to the occasional situation where you couldn't resize the two columns, depending on where, within the control, your mouse was.

This PR addresses this minor bug.

**Before:**
MGCBEditorDividerBug

**After:**
MGCBEditorDividerFix


harry-cpp:
Good find, thanks and merging!

Issue #8004, opened 2023-03-28
Add a description on an InvalidOperationException thrown from `Model.Draw(Matrix, Matrix, Matrix)`


rbwhitaker:
I was playing with custom effects, set up my own effect (from the template), loaded it, and then attached it to a model, then ran it. My code already had a line like `model.Draw(world, view, projection)`, and this line threw an `InvalidOperationException`. That's because my effect did not implement `IEffectMatrices`, which is a decent reason to fail. My problem is that the error message gave me little to go on. I had to dig into the source code to find the issue.

What if we changed that line in `Model.Draw(Matrix world, Matrix view, Matrix projection)` to be something more like `throw new InvalidOperationException("effect must implement IEffectMatrices to support this call");` or something?

That at least points people in a reasonable direction if they do what I did, rather than being faced with a completely opaque error. (I had initially assumed something had been disposed.)

I'd be willing to make a PR for it, as long as I knew it was something people were interested in.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

nkast:
That would be helpful.  Here's how I changed the code.

I stayed out of the way and let the CLR do it's job.
In case the cast fail, the system will throws an InvalidCastException.

```
                {
	   	    IEffectMatrices effectMatricies = (IEffectMatrices)effect;
                    effectMatricies.World = sharedDrawBoneMatrices[mesh.ParentBone.Index] * world;
                    effectMatricies.View = view;
                    effectMatricies.Projection = projection;
                }
```

rbwhitaker:
I'm not opposed to that solution, and kind of like it. It does change the API, technically. What once threw an `InvalidOperationException` would now throw an `InvalidCastException`, and it is possible (but I'd think highly unlikely) that somebody is catching the current `InvalidOperationException` and handling it otherwise. The change in exception type could break that. Strictly speaking, even changing the text, as I initially suggested, could break something if people are keying off the message to do stuff.

[...]

PR #8019, opened 2023-04-16
Adds a description to an InvalidOperationException for clarity.

rbwhitaker:
This should "fix" #8004, if that's something the maintainers wish to do something about.

I was porting some really old XNA code to MonoGame and bumped into an issue where I got an `InvalidOperationException`. I think the exception was correct, but in order to understand what was going on, I had to pull down the code and step through it to discover that the effect's type was expected to implement `IEffectMatrices`. Once I saw that, I knew how to fix it. But I had to dig quite deep to find it.

This change would hopefully prevent somebody else from having to do the same digging in the future.

On the original issue, there was a comment by @nkast suggesting another possible solution: Just do a cast and let the runtime handle things. I feel that solution would also address my problem, if we'd rather go that route. Though that might run a slightly higher risk of breaking people, since it changes the exception type.

If the maintainers would prefer that solution, or any other tweaks to this solution, let me know and I'll change it.

If we don't want to do anything with this at all, you won't hurt my feelings to just decline this PR.

### DIFF
--- a/MonoGame.Framework/Graphics/Model.cs
+++ b/MonoGame.Framework/Graphics/Model.cs
@@ -132,7 +132,7 @@ namespace Microsoft.Xna.Framework.Graphics
 					IEffectMatrices effectMatricies = effect as IEffectMatrices;
 					if (effectMatricies == null)
                     {
-						throw new InvalidOperationException($"Effect must implement {nameof(IEffectMatrices)} to support this call.");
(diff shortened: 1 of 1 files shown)


rbwhitaker:
I updated the text shown in this exception to match what XNA does, as shown in the screenshot here: https://github.com/MonoGame/MonoGame/issues/8004#issuecomment-1511236718

harry-cpp:
Non descriptive error messages are the worst :D Thanks! Merging!

PR #8028, opened 2023-05-03
Small fixes

rinatxp:
Some small fixes:
- using (obj is [type] objType) to get rid of an extra cast
- using StringBuilder in ToString() methods for more performance


rejurime:
Hi! my 2 cents, I think 21 commits to change 15 files isn't great for commit history, wouldn't it be better to do a PR with a clean history to reflect what you want to change? For example, the same commit message is repeated many times and I don't think that adds value to the story.

rinatxp:
Hi! Ty for your advice. I squashed all these commits into one

harry-cpp (changes requested):
I like small cleanup of `is` operator usage, but please revert the `StringBuilder` usage in `ToString` methods.

Also don't worry about making too many commits in a PR, the maintainers can easily squash them down during the merge process 😉

rinatxp:
No problem, I reverted StringBuilder usage

MonoGame.Framework/Vector2.cs
```diff
@@ -6,6 +6,7 @@
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
+using System.Text;
```
harry-cpp, line 9: Missed reverting this one.

mrhelmut:
These changes are C# 5+ features. We really need to restore the C# 5 check from the ConsoleCheck csproj and re-run it to make sure that none slipped in while it was disabled.

mrhelmut:
Like I feared, we already merged code that breaks this. I'll track that later to revert those changes.

SimonDarksideJ:
Any follow up @rinatxp / @mrhelmut ?

mrhelmut:
> These changes are C# 5+ features.

Still an issue as of now. Wouldn't recommend merging this for the time being.

AristurtleDev:
@SimonDarksideJ i labeled this one as `status: blocked` due to @ThomasFOG 's comments about changes pushed that need to be reverted, but also the comment on #8194 being open. That issue/bounty has been marked closed, so this needs to be looked at again.

SimonDarksideJ:
@rinatxp any chance you can re-review this change now that the NativeAoT platform has been merged?

Issue #8030, opened 2023-05-15
Going fullscreen resizes VS Code (and other Electron apps) window


Nebulaxin:
I don't know how going fullscreen can affect other windows. Other Electron apps (ex. Discord) also can be resized this way. This is probably an SDL/Electron issue, but maybe it is somehow related to MonoGame.

By the way, setting `graphics.HardwareModeSwitch` to `false` will make window maximized borderless instead of going fullscreen because of [`var fullscreenFlag = _game.graphicsDeviceManager.HardwareModeSwitch ? Sdl.Window.State.Fullscreen : Sdl.Window.State.FullscreenDesktop`](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/SDL/SDLGameWindow.cs#L227) and window behind game's one will not be resized.

https://github.com/MonoGame/MonoGame/assets/69744837/509a7730-59e7-47d9-9fb9-ed17a706b0f1

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1.303

#### What operating system are you using:
- Windows 10 22H2

#### What MonoGame platform are you using:
- DesktopGL

Nebulaxin:
The only solution I see is `graphics.HardwareModeSwitch = false` when debugging.

harry-cpp:
This is not a bug, this is just how Windows behaves when the resolution of the monitor is changed and there are windows on that screen.

As you've noticed `graphics.HardwareModeSwitch` dictates if the monitor resolution should change if its `true`, or if only the games rendering resolution should change and not touch the actual screen resolution if its `false` (aka. "Fullscreen" and "Borderless Fullscreen" modes for games).

Nebulaxin:
Seems like Electron apps are sensitive to resolution changing.

I tried reversing order of going fullscreen/resizing:
- when going fullscreen first resize window and then go fullscreen
- when going windowed first go windowed and then resize

When going fullscreen on native resolution (ex. `1920x1080` -> `1920x1080`) everything works like in fullscreen desktop, but still not the same because SDL uses `Fullscreen` state, not `FullscreenDesktop`.

When going fullscreen on non-native resolution (ex. `1920x1080` -> `1280x720`) VS Code still get resized (because this is how fullscreen works), but at least it's not that annoying.

PR #8031, opened 2023-05-17
Fix SDL Fullscreen

Nebulaxin:
Made some chages based on https://github.com/MonoGame/MonoGame/issues/8030

Because of the order of SDL calls, when going fullscreen and `GraphicsDeviceManager.HardwareModeSwitch` is true, fullscreen mode is enabled before resizing

And because of that resolution is set to window size, not `PreferredBackBufferWidth/Height`, and only then resolution is set to `PreferredBackBufferWidth/Height`.

In this fix window will switch to fullscreen mode only after resize.


Nebulaxin:
The only problem I see is that window blinks when going fullscreen and `graphics.HardwareModeSwitch` is false

Nebulaxin:
> The only problem I see is that window blinks once when going fullscreen and `graphics.HardwareModeSwitch` is false

This blinking also happens on the `develop` branch. I tried applying this fix on `master` and everything works without blinking.

I think it happens because of new SDL version.

Nebulaxin:
this is why you create new branches

PR #8032, opened 2023-05-18
Use ScanCodes instead of KeyCodes for Keyboard mapping

rastla:
Fixes the issue described in https://github.com/MonoGame/MonoGame/issues/8033
US keyboard layout *should* be unaffected by this.

DE keyboard layouts would now be able to use more keys (`^ ß ´ ü ö ä #`).
Similar for other keyboard layouts like `FR`.

Keys.W and Keys.A would now map to `Keys.Z` and `Keys.Q` in the `FR` keyboard layout.
So if you map your walking to **WASD** via the Keys enum, then players with FR layout will be able to walk with **ZQSD** - which is the preferred way.

This also means that the Keys enum is only representative for the `US` layout (it also already included only `US` keys, so that shouldn't be much of an issue I assume).

Meaning, although you are pressing `^` on the `DE` keyboard, it is detected as `Keys.OemTilde` (`~`)
Meaning, although you are pressing `ö` on the `DE` keyboard, it is detected as `Keys.OemSemicolon` (`;`)
You get the idea


stromkos:
The problem with your PR is that scan-codes vary per keyboard layout. In the DE keyboard Z = English Keyboard Q. The scancodes are the same. This is a discrepancy is normally resolved via the OS; SDL simply passes the scan-codes, ignorant of the keyboard layout.

rastla:
> The problem with your PR is that scan-codes vary per keyboard layout

To quote yourself 😄 
> The scancodes are the same

The scancodes stay the same, regardless of keyboard layout. It's the keycodes that change with the keyboard layouts.
Which is exactly what this PR is about.

>This PR breaks the existing system in 2 ways: It breaks existing keyboard mappings, and loses the ASCII mapping.

It doesn't break the US keyboard layout, which is the only one that works properly.
It kind of breaks the keyboard mappings of non-US keyboard layouts for existing projects.
It also provides a fully usable keyboard for non-US keyboard layouts, which isn't possible without this PR.

Maybe it would be better to offer the ScanCode functionality in addition to the existing KeyCode mapping?

Epicguru:
Has there been any consideration towards merging this PR? It seems like a straight upgrade, assuming that Sdl accurately reports scancodes accross desktop platforms.
Most modern game engines/libraries use this aproach, such as Unity which recently made 'physical keys' the new default (https://docs.unity3d.com/Manual/class-InputManager.html).

The way I see it (again assuming that it works as intended), it will have no negative impact on the current standard US layout, but it will provide a better default behaviour for non-US layouts.

mrhelmut:
This solution is incomplete to me. Yes, most libraries made that choice, but Unity does offer an API to get what's actually printed on the physical key depending on the user's layout. Which we don't (and it's a concern whenever you want to display key bindings in a setting menu, for example).

The initial XNA design (and SDL's virtual keys) was to try to fuse those concerns into a single bloat-free API: you get what you expect. It indeed came with a very US-centric view on the problem, but the intent was good to me and I believe that we can still expand on that.

Anyway, as-is it's a breaking change and can't be considered for the ongoing 3.8 branch.

SimonDarksideJ:
@rastla any plans to revisit this PR to make it more complete.  My overriding concern is you have completely removed some mappings just because you couldn't test them, so this is indeed incomplete.

This PR would probably be better served with having alternate "mapping modes" so the developer could "Choose" the method to be used, this not breaking the existing mappings in all existing titles.
A developer would have to intentionally switch to the new "scanmodes" option to enable it, thus preserving backwards compatibility.

Can you let us know if you intend to complete this work?

rastla:
@SimonDarksideJ 
Sure, I can refactor it to make the usage of ScanCodes instead of KeyCodes opt-in, rather than replace the current behaviour.

However, I will be unable to complete the map, because I don't have any US keyboard, nor any special keyboard with those additional keys.
From quickly going through the keys, the following keys are the ones that are missing from my PR:

```
Keys.OemBackslash
Keys.OemClear
Keys.SelectMedia
Keys.LaunchMail
Keys.BrowserSearch
Keys.BrowserHome
Keys.BrowserBack
Keys.BrowserForward
Keys.BrowserStop
Keys.BrowserRefresh
Keys.BrowserFavorites
Keys.Sleep
Keys.F13
Keys.F14
Keys.F15
Keys.F16
Keys.F17
Keys.F18
Keys.F19
Keys.F20
Keys.F21
Keys.F22
Keys.F23
Keys.F24
```


It also should be said, that with this PR, you can use more keys on non-US keyboard layouts, but they will still be reported as if the US layout is used.
This means, I press `Z` on my german keyboard, it will be reported as `Y`.
But as I said, it also means, that keys that previously just couldn't be used at all in a german keyboard layout, can now be used.
For example `^` - previously Keys.None - could not be used.
After this PR `^` will be reported as Keys.OemTilde (`~`)

Other example example `ö` - previously Keys.None - could not be used.
After this PR `ö` will be reported as Keys.OemSemicolon (`;`)


Though being able to use a mislabelled key is still better than not being able to use a key at all.

SimonDarksideJ:
Thanks for responding @rastla , but that is what the community is for.  SO if you can convert the PR to "Draft Status" and then reach out on the MG Discord for help testing, or getting original codes, you should be able to fill the gaps.

But our 100% concern is NOT breaking backwards compatibility, so the original implementation MUST be preserved, defaulted to.  Then a mode switch can change to this alternate approach.

I look forward to your update.

SimonDarksideJ:
Any further update @rastla ?  Was looking to include this in another develop release but it seems not to be ready yet.

rastla:
@SimonDarksideJ sorry, no didn't have time yet
I'm out of the country currently and will be back next week wednesday. I will look into doing that on the weekend august 10

ThomasFOG:
We should discuss this more, I'm pretty sure that we can do something better than a mode switch. I'll try to elaborate when I can, but there's likely no need for fancy stuff for things to work within the current API design.

PR #8034, opened 2023-05-20
Changed SDL fullscreen calls order

Nebulaxin:
Small fix based on #8030

Because of the order of SDL calls, when going fullscreen and `GraphicsDeviceManager.HardwareModeSwitch` is true, fullscreen mode is enabled before resizing

And because of that resolution is set to window size, not `PreferredBackBufferWidth/Height`, and only then resolution is set to `PreferredBackBufferWidth/Height`

In this fix window will switch to fullscreen mode only after resize


dellis1972:
@harry-cpp do you have thoughts on this? On the issue you reported its not a bug. That said if this improves fullscreen switching overall I have no problem with it.

harry-cpp:
In theory, it should be fine, tho this part of code is known for behaving differently on different operating systems.

dellis1972:
@Nebulaxin are you able to test this on Linux or Mac? to make sure it does not break the behaviour on those platforms?

harry-cpp:
Maybe easier to just merge and keep our eyes peeled if something crops up?

mrhelmut:
I'd prefer to test most scenarios. Not a fan of the "let's see if something breaks" approach in that regard.

SimonDarksideJ:
So long as it passes tests, we can review in the next preview

PR #8037, opened 2023-05-27
adding an implicit operator from System.Drawing.Color to Microsoft.Xna.Framework.Color

BenMakesGames:
I've occasionally wanted to use `System.Drawing.Color` colors in MonoGame (for color palettes provided by other, non-MonoGame-aware libraries, for example), and manually converting them can be a little tedious, so I figured I'd put this PR together, and see if it sticks. if this change does not mesh with MonoGame's philosophies, a thousand pardons, and feel free to just close it.

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -312,6 +312,12 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Nebulaxin:
I think it should be explicit `ToSystem` or `ToDrawing` like [Vector2.ToNumerics](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Vector2.cs#L1271)

PR #8040, opened 2023-06-09
Added checks when opening files in content editor

Nebulaxin:
Previously it would throw an exception and crash because file path doesn't exist


harry-cpp:
Looks good, thanks @Nebulaxin, merging!

Issue #8045, opened 2023-07-17
Project Templates Do Not Restore NuGets Automatically


AristurtleDev:
## Description
The current MonoGame templates, when creating a new project, do not restore NuGet packages once a project is created.  This often leads new users showing errors on missing namespaces and such until they either do a first build or manually run `dotnet restore` on the project.

This can be resolved by including a `postAction` for `dotnet restore` into the `template.json` configuration file for each template.  More information can be found at https://github.com/dotnet/templating/wiki/Post-Action-Registry

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1

#### What operating system are you using:
- All

#### What MonoGame platform are you using:

- All

AristurtleDev:
This was resolved with the merge of PR #8046

PR #8046, opened 2023-07-17
Automatically Restore NuGets When New Project Created From Templates

AristurtleDev:
## Description
This pull request edits all `template.json` configuration files to include a `postActions` section that will automatically perform a `dotnet restore` when a new project is created from the template.

This is something that is done by default automatically with Visual Studio, however it is not something done automatically by other IDEs such as VSCode.  By including this post action, it ensures that when new users create new projects based on the templates, regardless of IDE, the NuGets should be restored and they are not presented with the initial errors that the namespaces used by MonoGame cannot be found.

More information about the `postActions` can be found at https://github.com/dotnet/templating/wiki/Post-Action-Registry#restore-nuget-packages

This pull request is in reference to Issue #8045 that I opened just prior.


harry-cpp:
All this time I thought they built in the restore command into the dotnet new command :D 

This is great, thanks @AristurtleDev, merging!

Issue #8050, opened 2023-07-24
Setting `Window.Title` to `string.Emtpy` Causes Crash


AristurtleDev:
## Description
If at any point the `Window.Title` property is set to an empty string (e.g. `""` or `string.Empty`) the game will crash

## Expected Behavior
Setting `Window.Title` property to an empty string would properly set the window title to an empty string

## Action Behavior
The game silently crashes

## Steps to Reproduce
1. Create a new project using `dotnet new mgdesktopgl`
2. In the `Game1.cs` ctor, set `Window.Title = string.Empty`
3. Run the game

The game will not even launch and just silently crashes.

## Research
This issue comes from the `SDL2.SetTitle(IntPtr, string)` ([Reference](https://github.com/MonoGame/MonoGame/blob/b21463b419e55b4c898030fc22bee77dabb11210/MonoGame.Framework/Platform/SDL/SDL2.cs#L469))

```cs
public static void SetTitle(IntPtr handle, string title)
{
    var bytes = Encoding.UTF8.GetBytes(title);
    SDL_SetWindowTitle(handle, ref bytes[0]);
}
```

When the `title` parameter is an empty string, the resulting `byte[]` array is a zero length array.  Thus when calling `SDL_SetWindowTItle` an `IndexOutOfBoundsException` is thrown by the `ref byte[0]` argument being passed in.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1.303

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

AristurtleDev:
This was resolved with the merge of PR #8051

PR #8051, opened 2023-07-24
Resolves Crash When Setting `Window.Title` To An Empty String

AristurtleDev:
## Description
When setting `Window.Title` to an empty string (e.g `""` or `string.Empty`, the game will crash due to `SDL_SetWindowTitle` passing `ref byte[0]` on a `byte[]` that has a zero length, causing an `IndexOutOfBoundsException`.

This PR resolves this issue by appending a `'\0'` null terminator character to the `title` parameter, ensuring that the `byte[]` generated is non-zero length and still produces an empty string for the title.

## Reference:
This PR is in reference to the issue I submitted at Issue #8050


harry-cpp:
Looks good, merging!

Issue #7913, opened 2022-08-19
Microsoft.Xna.Framework.Game.Window.Title does not set the correct  title in it's constructor 


MauNguyenVan:
Microsoft.Xna.Framework.Game.Window.Title does not set the correct  title in it's constructor .

See my code
 ```
  internal class UnitCircle: Game
    {
        public UnitCircle()
        {
            this.Window.Title = "Unit Circle";
        }
}
```
When I run, my window does not display new title.

captainkidd5:
Move 

`Window.Title = "Unit Circle";`

into the `Initialize()`

see [https://community.monogame.net/t/cant-set-window-title-in-game1-constructor/9465](url)

Beyley:
> Move
> 
> `Window.Title = "Unit Circle";`
> 
> into the `Initialize()` method
> 
> [See this issue](https://community.monogame.net/t/cant-set-window-title-in-game1-constructor/9465)
> 
> Seems to be a bug which has been around with DesktopGL for a while

Its likely not a bug but just due to how constructor order works

PR #8052, opened 2023-07-24
Respect setting `Window.Title` in `Game1` constructor

AristurtleDev:
## Description
When setting the `Window.Title` in the `Game1` constructor, the title set is ignored.  This is because the `SDLGameWindow.CreateWindow()` method explicitly chooses to use `AssemblyHelper.GetDefaultWindowTItle()`.  During this time, the `SDLGameWindow.Title` property contains the correct window title that is set during the `Game1` constructor, but ignores this.

This PR resolves this issue by explicitly checking if `SDLGameWindow.Title` is `null`, and if so, then it will use `AssemblyHelper.GetDefaultWindowTitle()`, otherwise it will use the title that was set by the user in the `Game1` constructor as expected.

Reference:
An issue was opened previously at Issue #7913 

Note:
This PR is also related to PR #8051, however I created two separate PRs as each focuses on a separate issue to fix.

### DIFF
--- a/MonoGame.Framework/Platform/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/Platform/SDL/SDLGameWindow.cs
@@ -159,7 +159,7 @@ namespace Microsoft.Xna.Framework
             _height = GraphicsDeviceManager.DefaultBackBufferHeight;
 
             _handle = Sdl.Window.Create(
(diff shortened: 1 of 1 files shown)


harry-cpp:
Looks good to me, thanks @AristurtleDev , merging!

Issue #8054, opened 2023-08-19
Content Builder error message confuses Target Platform and Texture Format


KUNGERMOoN:
### The problem:
Error message received when compressing a texture for a platform that doesn't support the selected format confuses target platform with texture format.
For example, when building content for Windows and compressiong texture with ATITC (ATC texture format, not supported on windows) the message of the thrown error is 
> AtcCompressed platform only supports DXT texture compression
while it should be 
> Windows platform only supports DXT texture compression

### The fix:
In [DefaultTextureProfile.cs
](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/Graphics/DefaultTextureProfile.cs) on line 71 replace 
```csharp
throw new PlatformNotSupportedException(format + " platform only supports DXT texture compression");`
```
 with 
```csharp
throw new PlatformNotSupportedException(platform + " platform only supports DXT texture compression");
```
I'd have fixed it myself if I didn't already have an active Monogame fork.

### Why anyone should care about it:
For people getting started with Content Pipeline texture compression and formats might be confusing enough, so wrong and confusing error messages showing up when they try to build their content are definitively not helpful.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

mrhelmut:
Fixed in  #8055

PR #8055, opened 2023-08-19
Correct error message in `DefaultTextureProfile`

AristurtleDev:
## Description
This pull request is to correct the error message presented at https://github.com/MonoGame/MonoGame/blob/b21463b419e55b4c898030fc22bee77dabb11210/MonoGame.Framework.Content.Pipeline/Graphics/DefaultTextureProfile.cs#L71

by replacing the `format` variable with the `platform` variable.

## Reference
This is in reference to and resolves issue #8054.  Thanks to @KUNGERMOoN for opening the issue and sharing the resolve, I've created the PR in place since they were unable to according to the issue they opened.


harry-cpp:
Thanks @AristurtleDev, merging!

PR #8069, opened 2023-09-13
Updated dotnet install command

MeblIkea:
Dotnet new --install is now deprecated.
I replaced `dotnet new --install MonoGame.Templates.CSharp` with `dotnet new install MonoGame.Templates.CSharp`.


harry-cpp:
Good find. thanks @MeblIkea , merging!

PR #8071, opened 2023-09-17
Updataing Docs: Setting Up Your Development Environment Windows

AristurtleDev:
## Current Status
Ready for Review

## Description
This pull request is to update the **Setting Up Your Development Environment Windows** documentation.  Commits will be made on a change basis and should be squashed before merging.

## Change Reasons
The following are reasons specific changes have been made to this document
1. `Updated image to show current workloads where mobile now shows MAUI instead of Xamarian`
    * Visual Studio 17.7.x update changed the workload image to state that mobile development uses MAUI now instead of Xamarian.  Image was updated to reflect the current way it will look for users
2. `Moved prior version information into a more visual Alert box`
    * ~~Information regarding prior versions of Visual Studio not being compatible with MonoGame 3.8 was moved into an `Important` alert box to better better emphasize the message.~~
    * Removed verbiage and alert box per review as the information is not relavent since the extension/templates are not available in Visual Studio prior to 2022 any way, so the version is implied.
 3. `Adjusted paragraph to be more verbose and include Note alert about visual studio editions`
     * ~~Information about the visual studio editions is more of a foot note or notation for the user reading the guide than it is information included inline with the paragraph.~~
     * Alert and verbiage removed as per review since information provided here is redundant. 
4. `Adjusted workload list to use nested bullets and group by mandatory vs optional`
    * Makes the information more digestible when reading by nesting the bullet points and grouping by Mandatory vs Optional
5. `Information about needing DirectX runtime moved inside an Important alert box to visually draw attention`
    * This is an setup detail that is important for readers to see, moved it into an `Important` alert box to visually draw their eyes to it
6. `Updated Extension Manager image to reflect the current 3.8.1.303 version`
    * The Extension manager image was updated to show the current 3.8.1.303 version of the extension instead of the 3.8.0.175 version
7. `Specify MonoGame projects`
    * Updated to specify MonoGame projects specifically instead of just the generic "projects" which could imply any projects.
8. `Specify the full name of the extension and put it in bold`
    * This changes `MonoGame extension` to `MonoGame Framework C# project template extension` to reflect the actual name of the extension as shown in the Extension manager.  Name was changed to bold style for emphisis.
 9. `Updated steps to install extension to be more verbose for new users not familar with installing Visual Studio
extensions`
    * Installing extensions in Visual Studio is not a typical scenario that most users will encounter, especially new users coming from other engines/frameworks.  The steps to install extension were made more verbose with additional images included to ensure that all users understand how to install the extension, include the VSIX Installer pop-up that occurs after downloading the extension (this pop-up has been missed by users in Discord and caused confusion before)
 10. `any changed to either`
     * Grammatically, `any` should be used when there is three or more choices, `either` should be used when there are only two choices.  Since this sentence is speaking of Rider or VSCode, this is only two choices, so `either` is the more appropriate language.


Documentation/articles/getting_started/1_setting_up_your_development_environment_windows.md
```diff
@@ -2,35 +2,64 @@
 
 This section provides a step-by-step guide for setting up your development environment on Windows.
 
-MonoGame can work with most .NET compatible tools, but we recommend [Visual Studio 2022](https://visualstudio.microsoft.com/vs/) (prior versions are not supported with MonoGame 3.8.1).
+MonoGame can work with most .NET compatible tools, but we recommend [Visual Studio 2022](https://visualstudio.microsoft.com/vs/)
+
+> [!IMPORTANT]
+> Prior version of Visual Studio are not supported with MonoGame 3.8.1
```
harry-cpp, line 8: Only the templates are missing from older versions, its not like we provide some sort of support for it.

Documentation/articles/getting_started/1_setting_up_your_development_environment_windows.md
```diff
@@ -2,35 +2,64 @@
...
 
 Alternatively, you can use [JetBrains Rider](https://www.jetbrains.com/rider/) or [Visual Studio Code](https://code.visualstudio.com/).
 
 ## [Recommended] Install Visual Studio 2022
 
-Before using MonoGame, you will need to install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/) or later (any edition, including Community) with the following workloads, depending on your desired [target platform(s)](~/platforms.md):
+Before using MonoGame, you will need to install [Visual Studio 2022](https://visualstudio.microsoft.com/vs/)
+
+> [!NOTE]
+> You can install any edition of Visual Studio, including the Community edition.
```
harry-cpp, line 17: I'd say this is redundant.

harry-cpp (changes requested):

Documentation/articles/getting_started/1_setting_up_your_development_environment_windows.md
AristurtleDev: The change here was to move the information already present into an alert box.  Would you suggest removing the information about prior versions completely since older versions wouldn't have the extension/templates to install anyway?

Documentation/articles/getting_started/1_setting_up_your_development_environment_windows.md
AristurtleDev: I agree, this information is redundant and the alert could be removed.  I just took the existing verbiage and moved it into an alert box.

Documentation/articles/getting_started/1_setting_up_your_development_environment_windows.md
harry-cpp: It can also just be reworded to specify that older versions can't get the templates installed for them (and mgcb-editor launching will manually need to be added as external tool).

AristurtleDev: Alert for prior versions removed since extension/templates are not available for prior versions so 2022 version is implied.

Documentation/articles/getting_started/1_setting_up_your_development_environment_windows.md
AristurtleDev: Removed redundant information in alert box.

harry-cpp (approved):

PR #8072, opened 2023-09-18
Fix typo in MeshHelper.cs

eltociear:
recieve -> receive


harry-cpp:
Thanks for the type fix @eltociear! Merging!

PR #8075, opened 2023-09-21
Add GitHub Issue Templates

AristurtleDev:
## Description
The following PR adds GitHub Issue templates to the repository.  Currently, when a user opens a new issue, they are only given the following to submit an issue with

image

## Motivation
The motivation for this pull request it to provide a better user experience when users need to file a bug report or want to submit an issue request.  By breaking down the relevant information we need for these types of submissions, these templates provide a better user experience by presenting them as form submissions instead of one massive text box, with required and optional fields.

So how does the workflow change for users submitting a new issue with this change.

### 1. New Issues Page
First, when a user clicks the "New Issue" button, they will now be shown the following page

image

From here, they have the option of choosing to specifically submit a **Bug Report** or a **Feature Request**.  Each of these are their own templates.   Additionally, links are provided to the community forums and the discord for when they have general questions instead of bugs or feature requests

### 2. Bug Report Template
The bug report template is broken down into several fields, most of which were already asked by the default issue text.  Only now it is more organized and provides a better overall user experience when filling out the information.  New questions are being asked such as to describe the expected behavior and the intended behavior to better indicate the type of information we need from users when facing a bug.

image

The MonoGame Platform dropdown includes the following options

Screen Shot 2023-09-21 at 1 07 58 AM

### 3. Feature Request
The feature request template provides two fields where users can describe the feature they would like to request be added and/or removed, as well as provide the motivation for requesting this feature and how it would better the MonoGame user experience.

image


AristurtleDev:
One thing I haven't included in this PR is another feature of the templates.  I have have them auto assign a label when a user submits them.

So when a user submits a feature request from the feature request template, it can auto-assign the `feature request` label that's in this repo.

I can also update it to assign something like a `New` label to both bug reports and feature requests, that way maintainers can filter the issues page to see anything that's "New" that hasn't been looked at yet, then they can change the status once it's looked at and initially responded too.

If maintainers would like this feature included as well, let me know and I'll update the PR.  There currently isn't a "New" label in this repo, so that would need to be added first

SimonDarksideJ (approved):

dellis1972 (approved):

.github/ISSUE_TEMPLATE/01_bug_report.yml
```diff
@@ -0,0 +1,111 @@
...
+    options:
+      - MonoGame Android Application (mgandroid)
+      - MonoGame Cross-Platform Desktop Application (mgdesktopgl)
+      - MonoGame Windows Universal CoreApp Application (mguwpcore)
+      - MonoGame Windows Universal XAML Application (mguwpxaml)
+      - MonoGame Windows Desktop Application (mgwindowsdx)
+      - MonoGame iOS Application (mgios)
+      - MonoGame Game Library (mgnetstandard)
+      - MonoGame Content Pipeline Extension (mgpipeline)
+      - MonoGame Shared Library Project (mgshared)
```
harry-cpp, line 40: Add N/A option, sometimes its a tooling issue.

Also I don't think we need library templates in here, they aren't really platforms and could be counted under N/A.

harry-cpp (changes requested):

harry-cpp:
> So when a user submits a feature request from the feature request template, it can auto-assign the `feature request` label that's in this repo.
> 
> I can also update it to assign something like a `New` label to both bug reports and feature requests, that way maintainers can filter the issues page to see anything that's "New" that hasn't been looked at yet, then they can change the status once it's looked at and initially responded too.

I like the feature-request tag, the new tag won't be that useful tho.

If it can auto assign the platform tags, that would be very useful as well.

AristurtleDev:
@harry-cpp Unfortunately, the GitHub Issues Templates only allows assigning the tag at a global level (so the bug report could have a bug tag and the feature request template can apply the feature request tag).  It currently doesn't support dynamically setting tags based on things like drop-downs or checkboxes selected on the form.


I could definitely a GitHub Action to assign the tag based on the value, but that would be outside of the templates themselves.  I'll look into that next in another PR

SimonDarksideJ:
Actually, I found today you can set rules to default options on new issues. No need to use a github action for it anymore.
Just go to "settings", find the issues section and then setup the rules from there.
Quite a few options to and fast easier.

.github/ISSUE_TEMPLATE/01_bug_report.yml
AristurtleDev: Changes made

AristurtleDev:
@SimonDarksideJ 

I don't see anywhere to setup rules? Maybe I'm missing something?  There's the Rulesets that you can create, but these are for doing restrictions on actions and you can do that based on tags.

Other than that, the only thing I see under the Issues section in repos is the option to create templates (like I've done here) which can apply 1 or more tags based on the template, but it isn't dynamic based on choices in the template.

.github/ISSUE_TEMPLATE/01_bug_report.yml
```diff
@@ -29,15 +29,14 @@ body:
...
     options:
+      - N/A
       - MonoGame Android Application (mgandroid)
       - MonoGame Cross-Platform Desktop Application (mgdesktopgl)
       - MonoGame Windows Universal CoreApp Application (mguwpcore)
       - MonoGame Windows Universal XAML Application (mguwpxaml)
       - MonoGame Windows Desktop Application (mgwindowsdx)
       - MonoGame iOS Application (mgios)
-      - MonoGame Game Library (mgnetstandard)
       - MonoGame Content Pipeline Extension (mgpipeline)
```
harry-cpp, line 39: Missed one library template.

.github/ISSUE_TEMPLATE/01_bug_report.yml
```diff
@@ -29,15 +29,14 @@ body:
   attributes:
     label: Which MonoGame platform are you using?
     options:
+      - N/A
```
harry-cpp, line 32: N/A is usually found at the bottom of the list.

harry-cpp (changes requested):

.github/ISSUE_TEMPLATE/01_bug_report.yml
AristurtleDev: Missed it because it didn't say "library". 
(also as a side note, I use that template quite a lot for extension development, but I understand removing it from here)

.github/ISSUE_TEMPLATE/01_bug_report.yml
AristurtleDev: Moved to the bottom, thanks

Issue #8073, opened 2023-09-19
 Translating into Portuguese and Portuguese-BR (Pt-BR)


BrunoMPVieira:
**There are many game developers in Brazil. I think it's important to translate the README to PT-BR for a better understanding, considering that the premise of the project is to be "EASY" to create games.**

MrValentine7777:
Hi, 

On behalf of the MonoGame Community, Welcome.

  For the time being, individual learners may utilise free translation services, however, for a business this would not be the case.

If some members from the Brazil side of the ponds can assist in translation, this would speed up the process greatly.

MonoGame is entirely a community driven project, and unfortunately I do not speak/read/write in more than ~two languages and neither of them are PT-BR otherwise I would have started on this by now. If you can reach out to a translation group and work with them to transcribe the documentation, this would go a long way to speed up the process for all.

Thank you for showing your interest in MonoGame, we hope to see your works in the future.

- A MonoGame Community Member.

harry-cpp:
Unfortunately we can't keep docs in languages other than english up to date.

BrunoMPVieira:
Thanks for your personal attention! **Great job!**

PR #8080, opened 2023-09-29
added a readme.pt-br and translated the readme to Portuguese

saiffyros:
Hey, I've added a readme.pt-br.dm file and translated the readme.dm to Portuguese. It's in the same folder as the readme file. I kept the same formating.


AristurtleDev:
This is related to the request in issue #8073.

harry-cpp:
Thank you for your contribution, but we really don't have the capacity to maintain docs in languages other than english.

PR #8087, opened 2023-10-30
fix: query GL_MAX_SAMPLES before setting multisampling hint

pandinocoder:
### Description of Change

- adds better located exceptions to catch things like "couldn't find matching glx visual"
- creates and disposes of temporary gl-enabled window and gl context to grab GL_MAX_SAMPLES from

This specifically fixes cases where `MultiSampleCount` is set to something like 16 on a build for Desktop GL and then is run on weaker hardware like a Steam Deck where `GL_MAX_SAMPLES` is 8.

Games do not have an API to query for this information to set a valid value for `MultiSampleCount` so the error is effectively unrecoverable and extremely hard to diagnose.

The exceptions currently thrown are only symptoms of the issue, they don't expose the actual failure which in this case needed to be queried for via `SDL_GetError()`.

### Contributor Declaration


mrhelmut:
You should be able to query ```GraphicsCapabilities.MaxMultiSampleCount``` in ```Game.Initialize()``` and apply graphics device parameters from there. Does that work?

DesktopGL comes with some limitation compared to WindowsDX in the fact that some graphics capabilities can be queried only after the Game constructor due to how SDL operates things under the hood. I'm a bit on the fence to create a fake window just to query that, I feel like it might introduce more issues in the long run that fixing this use case.

lodicolo:
`GraphicsCapabilities` and `GraphicsCapabilities.MaxMultiSampleCount` are internal so it would require some code changes to expose this.

It also has to create a temporary window no matter what, because GL can't query the API without an active context, and the way MonoGame is consuming SDL and GL it requires creating an SDL window.

I think creating the full graphics device chain in managed memory is creating a lot more garbage and introducing more opportunities for bugs instead of just quickly creating and destroying an SDL window in a controlled manner and then creating the real window. This is also, to my understanding, standard practice when using SDL (and GL).

I also personally have a concern with the approach of exposing this to the end users and expecting them to deal with it because:
- `MaxSamples` value is *not* always reliable, it reports 32 on my desktop when in reality it only supports up to 16 -- my Steam Deck reports correctly. Ordinarily I would be against my own fix for this, but the PR is meant as a "best effort" to prevent a crash, and includes better messaging so the end-user doesn't have to dive into dependencies with a debugger to figure out the root cause. I think having a potentially incorrect value as the maximum acceptable for a best effort, but not acceptable for a public API.
- Seems weird to expect users to know to poll graphics capabilities and then set the value
- Correct me if I am wrong, but in DX this can be done in the constructor, so this is also a GL-specific issue. I think it goes against the spirit of MonoGame to expect users to understand the internals of MonoGame and its dependencies in order to properly fix this issue.

dellis1972:
After reading the docs on this, I think we can handle this a different way (I could be wrong though).
We can leave the code to call `SetAttribute` as it is now since it is only a `hint`. If the OpenGL driver cannot handle it is will use the maximum value it can. 
We should then use `GetAttribute` to query the actual value from the driver after the window has been created. Then update the presentationParameters.MultiSampleCount  accordingly.

Something like

```
if (presentationParameters.MultiSampleCount > 0)
{
    Sdl.GL.SetAttribute(Sdl.GL.Attribute.MultiSampleBuffers, 1);
    Sdl.GL.SetAttribute(Sdl.GL.Attribute.MultiSampleSamples, presentationParameters.MultiSampleCount);
}
((SdlGameWindow)SdlGameWindow.Instance).CreateWindow();
Sdl.GL.GetAttribute (Sdl.GL.Attribute.MultiSampleSamples, out int samples);
presentationParameters.MultiSampleCount = samples;
```

See https://wiki.libsdl.org/SDL2/SDL_GL_GetAttribute

mrhelmut:
> it will use the maximum value it can

We need to verify that it actually does that. But if it does without error, it seems relevant.

lodicolo:
> If the OpenGL driver cannot handle it is will use the maximum value it can.

Unfortunately the GL driver doesn't intelligently do this, and that causes the crash that this PR is meant to prevent.

> We can leave the code to call `SetAttribute` as it is now since it is only a `hint`.

It may be called a hint, but it's not treated as something optional. The `SetAttribute()` call with an invalid `GL_SAMPLES` value will cause the context creation that happens after it to fail. `GL_SAMPLES` (abstracted as `SDL_GL_MULTISAMPLESAMPLES`) must be set to a value supported by the driver before context creation, and changing it requires re-creating the context.

`Sdl.GL.GetAttribute()` would call `glGetIntegerv()` which cannot be used without a current valid GL context, which will fail to be created if `GL_MAX_SAMPLES` is set too high. The hidden dummy window being created would have the default `GL_MAX_SAMPLES` value of `0` which would allow it to be safely created in order to poll the `GL_MAX_SAMPLES` value in order to limit it.

Note that SDL2 does not expose `GL_MAX_SAMPLES` via `SDL_GL_GetAttribute()`, GL must be queried directly.

ThomasFOG:
There is already an hidden window created here that gets destroyed when the actual window is created (I don't know why it's created fullscreen, that seems weird but likely there for a reason):

https://github.com/MonoGame/MonoGame/blob/d5bf79ca793d1e36c4060e3a178a3f7e32c9653e/MonoGame.Framework/Platform/SDL/SDLGameWindow.cs#L133

This could be used to initialize the temporary OpenGL context by passing ``` Sdl.Window.State.OpenGL``` to it.
Remains the question of ```glGetIntegerv``` which we should avoid loading twice (but the problem is that entry point loading need an active GL context to function properly).

@pandinocoder could you look into this? Otherwise the PR will get stale and closed.

SimonDarksideJ:
Any update @pandinocoder / @ThomasFOG ?  Else this will be closed in 1 week as incomplete.
Also has merge conflicts now

SimonDarksideJ:
Last call @pandinocoder , else we will have to consider this PR stale.

pandinocoder:
@SimonDarksideJ Sorry, for whatever reason I did not get notifications (at least not on my phone) for any of the previous mentions.

I've resolved the merge conflict (my editor had removed some whitespace, and that line was then modified separately it seems).

I'm currently re-reading through the PR as it's been nearly 2 years since I last looked at it.

dellis1972:
@pandinocoder not saying you are, however its been a few years, I just wanted to point out our LLM policy https://github.com/MonoGame/MonoGame/blob/develop/CONTRIBUTING.md#large-language-models--generative-ai. 

You'll need to also add a LLM declaration to your PR, see https://github.com/MonoGame/MonoGame/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

pandinocoder:
@ThomasFOG Assuming that the other window (created during `PlatformCreate()`) is indeed safe to use, it appears to work. I do not know why it was created without `OpenGL` in the first place, but if there was a reason, I did not run into it. I pushed changes accounting for that, but I did leave a backup measure just in case the `PlatformCreate()` code is altered to remove that original window (which would break the `GL_MAX_SAMPLES` check).

Regarding the loading of `glGetIntegerv` "twice", it is more accurately being loaded once per context, and unfortunately two contexts need to be created to safely set the multisampling hints (on Linux at least).

@dellis1972 Thanks for letting me know of the policy change, I was not aware of it. I updated the PR message to include my declaration of no LLM usage. (I am happy to see such a policy in place too.)

dellis1972:
@pandinocoder 

There are also a few tests it might be worth enabling to check the functionality

https://github.com/MonoGame/MonoGame/blob/develop/Tests/Framework/Graphics/GraphicsDeviceManagerTest.cs#L438

and 

https://github.com/MonoGame/MonoGame/blob/develop/Tests/Framework/Graphics/GraphicsDeviceManagerTest.cs#L415

pandinocoder:
@dellis1972 

1. `MultiSampleCountRoundsDown()` (L415) needed some additional non-test code to be fixed, but it does work for me with the additional changes. If you think additional changes that are only tangentially related to the PR are worth adding, do you want me to push these?
2. `MSAAEnabled()` (L438) will not pass without creating a new fixture specifically for it. Do you want me to do that, or do you think it should just remain disabled for DesktopGL for now?
   - This is because MSAA needs to be enabled before window and context creation, and because of how `SetUp` is written in `GraphicsDeviceTestFixtureBase` they are created before it can be set.
   - It _**does**_ pass if I go add `gdm.PreferMultiSampling = true;` to `GraphicsDeviceTestFixtureBase`.`SetUp` on L66.

dellis1972:
@pandinocoder 

Hmm, not sure what to do in that case. We don't want this PR to get too big, but it would be nice to have a test of some sort (it doesn't have to be the existing ones, we can fix those later if need be).

pandinocoder:
I was initially hesitant to discard and recreate what was done in `SetUp`, but after noticing that the `InitializeEventCount` test also ignored the setup logic and used its own clean `Game` and `GraphicsDeviceManager` instances, I figured it was fine to do the same in `MSAAEnabled` rather than create an entirely new test fixture for what was essentially 1 partially-redundant test for 1 platform.

Most of the test was still more or less valid, so I left the test largely the same logically besides some minor code adjustments like assertion criteria (to be safe I made the DesktopGL test use MSAA 2x, and that didn't meet the gray pixel threshold so I had to lower it) and a preprocessor guard to keep the original test for non-DesktopGL platforms.

Also, this patch file has the changes needed to make `MultiSampleCountRoundsDown` pass as well: [fix-MultiSampleCountRoundsDown.patch](https://github.com/user-attachments/files/28079087/fix-MultiSampleCountRoundsDown.patch)

I _suspect_ it probably does not depend on the changes in this PR at all (but I have not confirmed that suspicion), but at the very least they are definitely not needed to fix what this PR addresses (the reason `MultiSampleCountRoundsDown` was failing was a different issue :joy: ).

I personally think that based on that, the patch for the second test belongs in a separate PR, but I will defer to your judgement on this.

pandinocoder:
I also just noticed that the Windows pipeline failed, but it does not seem like it is related to my changes?
> `warning MSB3246: Resolved file has a bad image, no metadata, or is otherwise inaccessible. PE image does not have metadata. PE image does not have metadata`

pandinocoder:
I am not entirely sure why the tests are only failing on 3 of the 4 environments, but since Windows is the only one passing I would guess that Windows is running all of the tests on the UI thread, while the other platforms do not.

The tests specifically are failing right after trying to create the window in `PlatformCreate()` because `Sdl.Window.Create(...)` fails. It did not fail prior to adding `Sdl.Window.State.OpenGL` to it, so I suspect that this has to run on the UI thread.

Since the `GraphicsDeviceManagerTests` that I actually modified are not appearing as failures in the logs so I am assuming they are passing, and both `SetUp` and `TearDown` are annotated with `[RunOnUI]` for those tests (this is not the case for failing tests while fail in `GameTest.FixtureBase.SetUp` which is only annotated with `[SetUp]`).

With the assumption that all of the tests ran and the listed failures are exhaustive, I see two ways forward:
1. The "reasonably certain" option I think is to remove `Sdl.Window.State.OpenGL` and use the temporary window solution all of the time, instead of just as an emergency fallback. This was what I recently changed after @ThomasFOG 's message in July.
2. The solution I am not entirely certain will even fix it is to add `[RunOnUI]` to all of the test fixtures that are currently failing, but in addition to not being certain this will fix it, I could see how this could incur additional test environment costs by increasing test runtime since only the UI thread can run them.

Which of the two options is preferred, or is there another path forward that I am not thinking of?

dellis1972:
The `[RunOnUI]` Attribute is there for making sure that tests run on the UI thread on MacOS, there might be some that need that now we've made changes.. so its worth a try. It probably won't fix the linux ones though as they seem to be trying to load EGL? which is odd.

pandinocoder:
Ok, to at least fix the MacOS tests I added the attribute.

I suspect that EGL is being used for context management on Linux and that is why it was being loaded. If that suspicion is correct, loading it also may not work off of the UI thread.

pandinocoder:
So looking at the new failures, the Linux ones are still failing with:
> `Microsoft.Xna.Framework.Graphics.NoSuitableGraphicsDeviceException : Could not load EGL library`

The MacOS failures are now:
> `Microsoft.Xna.Framework.Graphics.NoSuitableGraphicsDeviceException : NSWindow should only be instantiated on the main thread!`

The latter seems pretty strange given that I had added `[RunOnUI]`.

I have gone and pushed a revert of the OpenGL flag change and the `[RunOnUI]` change to see if it improves the test results.

pandinocoder:
That revert fixes the test failures on Linux, but the `NSWindow should only be instantiated on the main thread!` persists on MacOS.

I suspect that Linux was failing because the OpenGL initialization was possibly pushed too earlier, or maybe whatever is causing the MacOS failures contributed.

I do not see any failure logs for the `GraphicsDeviceManager` tests, and I do not see anything that would _exclude_ them from only MacOS, so I suspect they do run there and do pass, so it is a matter of figuring out what the difference is between those passing tests, and the tests that now fail because of the error checking/throw that I added to surface more accurate errors to developers so they do not need to spelunk through MonoGame to find platform issues like this PR is trying to fix.

I see two differences:
- `[RunOnUI]` was added on the fixture class as well, not just the methods. I did not include this because I figured the tests themselves did not need that, only `SetUp` (and possibly `TearDown`).
- I also see `[NonParallelizable]`, but given the error message complaining about not being on the main thread it does not seem as related.

`[NonParallelizable]` can also be added to methods (according to the attribute usage flags anyway), so I am going to try adding that and `[RunOnUI]` to the failing `SetUp` method (and its `TearDown` sibling).

I do not have an environment to reproduce the tests myself in unfortunately, and since these ideas are mostly just shots in the dark I am interested in any input to reduce the guesswork.

dellis1972:
So, on mac any "window" method calls need to be called on the main thread, when running nunit tests the tests are run on background threads, this is what causes the `NSWindow` issue for any test which does anything with the window. 
See https://github.com/MonoGame/MonoGame/blob/develop/Tests/Runner/Desktop/RunOnUIAttribute.cs#L26. 

Now I'm not sure if adding `RunOnUI` to setup methods will work. Because that class only seems to work on "TestCommand", I don't know if that covers Setup methods.

pandinocoder:
So after further inspection since nothing seemed to work, I realized that the `NSWindow should only be instantiated on the main thread!` message was not the exception causing the test failure (directly anyway), the exception causing the test failure was because in `TearDown()` `Dispose()` was being called on `_game` while it was `null` because of the `NSWindow should only be instantiated on the main thread!` exception being thrown in `SetUp()` causing it to never get set.

Because the NSWindow exception was indirectly triggering the test failure, I simply removed it. The only reason I had originally added exceptions was to bubble up the real platform errors during initialization, but that particular SDL window creation doesn't include GL, so throwing `NoSuitableGraphicsDeviceException` from there was not really necessary.

Also while investigating the failures I did eventually notice that my assumption that the `GraphicsDeviceManagerTest` tests were passing was in fact _not correct_ because they appear to never be executed at all. The class is not included in the workflow file at all. I do not know if that is intentional.

To reiterate from an earlier comment, putting `OpenGL` into the init flags for the `PlatformCreate()` SDL window would require the `NoSuitableGraphicsDeviceException` to be thrown there and would result in all of the Linux and MacOS tests to fail. I found no remediation for those tests without removing the exception being thrown from there, but without the exception it would not bubble up the correct error messages to the developer/end user (causing the confusion that this PR was meant to solve).

@dellis1972 Is there anything else wanted in this PR from my end?

MonoGame.Framework/Platform/GraphicsDeviceManager.SDL.cs
```diff
@@ -51,8 +53,44 @@ partial void PlatformInitialize(PresentationParameters presentationParameters)
 
             if (presentationParameters.MultiSampleCount > 0)
             {
-                Sdl.GL.SetAttribute(Sdl.GL.Attribute.MultiSampleBuffers, 1);
-                Sdl.GL.SetAttribute(Sdl.GL.Attribute.MultiSampleSamples, presentationParameters.MultiSampleCount);
+                // Store the "default" multisample count, which will be lowered if necessary to meet GL_MAX_SAMPLES.
+                var multiSampleCount = presentationParameters.MultiSampleCount;
+
+                // We should create a new temporary window to make a best-effort on grabbing GL_MAX_SAMPLES.
+                var temporaryWindowHandle = Sdl.Window.Create(
```
dellis1972, line 60: @AristurtleDev  based on your research for the native opengl , is this a good idea?

AristurtleDev: @dellis1972 I don't think so, at least not as the actual fix.  Probing it with a temp window is more of a best effort workaround.  From the native stuff I've been working on, probing like that didn't tell us what the real backbuffer could use.

On the native side, I got the MSAA caps reporting correctly, but even then that doesn't guarantee the default framebuffer would render with MSAA at that sample count.  So we'd be relying on something that doesn't really tell us whether the final window/backbuffer configuration is valid.

dellis1972: @AristurtleDev ok so if we can't query before hand, we need so way to deal with the problem where if the MultiSample setting is too high for the device it needs to be capped. otherwise you end up with black screens. So if the user requests 32 in the game, but it only supports 4 (or what ever real values would be) Then we need to cap it at 4 , rather than produce a black screen.

AristurtleDev: @dellis1972 I agree. My hesitation is more about using the temp window probe as if it is the authoritative answer.  From the work I've been doing, it can be useful as a best effort, but it doesn't necessarily tell what the final default framebuffer can create or render with.

Ideally there would be a fallback based on  the actual window/context creation path, and the probe could just be used as a hint.

pandinocoder: From what's been mentioned I'd like to offer a correction: I've never seen a black screen because of the MSAA value exceeding the limit, it was a hard crash during startup that also was opaque to the actual reason of the crash ("couldn't find matching glx visual") which makes diagnosing the issue impossible without stepping through MonoGame internals with a debugger. It being an unrecoverable crash is bad enough, but it's worse because MonoGame currently hides the true reason so developers would just be making guesses if they aren't debugging their dependencies.

And correct me if I am wrong, but it almost seems like this is being characterized as a best-effort workaround because it does not provide exhaustive information about what the backbuffer _could potentially use_, something that would matter primarily for an in-game graphics settings menu.

That's strictly _not_ what this is meant to address, it's only meant to ensure that no matter what the MSAA value is set to by the developer/defaulted to by MonoGame that, assuming the underlying driver is not lying when queried for `GL_MAX_SAMPLES`, the final GL context creation for the application does not fail and crash the application with no useful information presented to either the developer or end-user.

A more proper exhaustive solution requires dependencies that the current .NET implementation of MonoGame does not include nor expose to fetch information that it does not in its current form have access to. On Linux (the only OS I've seen this issue on so far) it would be by either querying `glxinfo` (external executable call, or native library dependency) or using the same calls as `glxinfo` does (still native). Adding a call to an external executable is naturally not something I would even suggest adding, and a native library dependency, to me, seems out of scope for this PR.

This is the only solution that I have found on Linux that does not involve the inclusion of additional native dependencies, and it has worked reliably across my SteamDeck, laptop, and desktops, on both NVIDIA and AMD GPUs (both integrated and discrete).

PR #8091, opened 2023-11-09
Fixed Game.IsActive on initial window creation

nichtsgames:
The following bug occurs upon executing a game build with MonoGame.Framework.DesktopGL on Windows:
If one quickly (before the game's window has been instantiated) changes the focus to another window by alt-tabbing or dragging another window with the mouse, the IsActive flag of the Game class will be set to true even though the game's window is not in focus. this persists until the game's window is actually in focus once.

This pull request fixed this bug. However, it does not do so in the most elegant way. This is intended more as an issue including a potential fix.

### DIFF
--- a/MonoGame.Framework/GamePlatform.cs
+++ b/MonoGame.Framework/GamePlatform.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Xna.Framework
 
         protected TimeSpan _inactiveSleepTime = TimeSpan.FromMilliseconds(20.0);
         protected bool _needsToResetElapsedTime = false;
+        bool ignoredInitialActivation = false;
         bool disposed;
         protected bool InFullScreenMode = false;
         protected bool IsDisposed { get { return disposed; } }
@@ -60,7 +61,14 @@ namespace Microsoft.Xna.Framework
             get { return _isActive; }
             internal set
             {
-                if (_isActive != value)
+                if (!ignoredInitialActivation)
+                {
+                    ignoredInitialActivation = true;
+
+                    _isActive = false;
+                    EventHelpers.Raise(this, _isActive ? Activated : Deactivated, EventArgs.Empty);
+                }
+                else if (_isActive != value)
                 {
                     _isActive = value;
                     EventHelpers.Raise(this, _isActive ? Activated : Deactivated, EventArgs.Empty);


MonoGame.Framework/GamePlatform.cs
```diff
@@ -16,6 +16,7 @@ abstract partial class GamePlatform : IDisposable
 
         protected TimeSpan _inactiveSleepTime = TimeSpan.FromMilliseconds(20.0);
         protected bool _needsToResetElapsedTime = false;
+        bool ignoredInitialActivation = false;
```
Mindfulplays, line 19: Is it better to have an `enum ActiveState { NotInitialized, Active, Inactive }` instead of these two booleans?
Also, the PR title really seems to be "Fix Game.IsActive notification on initial window creation" correct? Not really that the value is incorrect...

stromkos: > Is it better to have an `enum ActiveState { NotInitialized, Active, Inactive }` instead of these two booleans?

No.  That would bloat the code base.

My concern: 
Is there need for this change?  

In the cases that you describe, the user is fully aware the game may be starting without them.

This is an edge case that the only time that should be a problem is a game without a menu or title screen at startup. 

As this is an additional 32 bits of memory, and a one time event dispatch, I am not opposed to the change, I just question the need.

nichtsgames: > > Is it better to have an `enum ActiveState { NotInitialized, Active, Inactive }` instead of these two booleans?
> 
> No. That would bloat the code base.
> 
> My concern: Is there need for this change?
> 
> In the cases that you describe, the user is fully aware the game may be starting without them.
> 
> This is an edge case that the only time that should be a problem is a game without a menu or title screen at startup.
> 
> As this is an additional 32 bits of memory, and a one time event dispatch, I am not opposed to the change, I just question the need.

I definitely see a need for this change. The problem isn't really that the game already started.
Here is a contrived example: I want to start a game but I know that it takes 30 seconds to load up. So I start it but switch to my browser to do something else for those 30 seconds. If the game is done loading and is showing the title screen (which is hidden by the browser), you click buttons in the browser and without knowing it you also click buttons in the game.

PR #8092, opened 2023-11-15
Add high-frequency touch events to iOS GameView

Mindfulplays:
Fine-grained/high-frequency touch events are available in certain platforms upon request. On iOS, `coalesced touch events` can be obtained from the `GetCoalescedTouch` via the passed-in event.

This PR proposes a side-channel `OnCoalescedTouchEvent` that can be used in lieu or in addition to `OnTouchEvent` without any performance penalty. I have tested with and without the coalesced touch event including Apple Pencil which produces events > 60 events per second.

Also fixed minor typos here and there (going forward, I will create smaller commit PRs to avoid polluting).


Mindfulplays:
@harry-cpp @mrhelmut Could you please help review? thank you!

stromkos:
On the PR itself, there are several regression issues that need to be addressed.

Please collapse the commits.

There are thread timing issues that need to be formally addressed since the entire TouchCollection becomes asynchronous. 

Rather than passing an event through the framework for this case, set a listener, hook with callback delagate, for the event and call the handler(a static function elsewhere) in Program.cs or Activity.cs.

That was how I handled the delay issue on Android and Windows.

What are the associated costs:

https://developer.apple.com/documentation/uikit/touches_presses_and_gestures/handling_touches_in_your_view/getting_high-fidelity_input_with_coalesced_touches

>Important
>
>Coalesced touches are intended for apps that need the extra precision and can handle the associated costs. Processing coalesced touches means gathering additional data and applying it to your content. If you don’t need the extra precision, continue using the set of touch objects that UIKit passes to the methods of your views or gesture recognizers.

Mindfulplays:
> On the PR itself, there are several regression issues that need to be addressed.
> 
> Please collapse the commits.
> 
> There are thread timing issues that need to be formally addressed since the entire TouchCollection becomes asynchronous.
> 
> Rather than passing an event through the framework for this case, set a listener, hook with callback delagate, for the event and call the handler(a static function elsewhere) in Program.cs or Activity.cs.
> 
> That was how I handled the delay issue on Android and Windows.
> 
> What are the associated costs:
> 
> https://developer.apple.com/documentation/uikit/touches_presses_and_gestures/handling_touches_in_your_view/getting_high-fidelity_input_with_coalesced_touches
> 
> > Important
> > Coalesced touches are intended for apps that need the extra precision and can handle the associated costs. Processing coalesced touches means gathering additional data and applying it to your content. If you don’t need the extra precision, continue using the set of touch objects that UIKit passes to the methods of your views or gesture recognizers.


Could you add what the regressions are to the PR comments? Not seeing it...

The coalesced events (+processing) is optional (existing apps can choose to subscribe by adding `OnCoalescedTouchEvent` handler). No extra work is done unless the caller asks for it.

I am not aware of any thread timing issues as the `evt` holds on to the `TouchCollection` in memory. There is no round-trip back to the OS or any other blocking main-thread activity. I verified by both stepping through the framework code and the provided iOS documentation.

In case I am missing something, please clarify / provide specific examples.

Mindfulplays:
@harry-cpp @mrhelmut  could you please take a look? Thanks!

tomspilman:
@Mindfulplays 

I have a few issues with this PR.

One i really don't like having two event paths thru the whole code base... touch events and coalesced touch events.  Especially since it seems coalesced touch has the same data as a normal touch event.  

So maybe it just needs to be a flag on `TouchLocation`?  It goes thru the same code path, but is flagged as these special high precision touch data.

Also i don't like calling this "coalesced" touches...  it just perpetuates the bad naming from Apple.  This should be called `TouchPanel.EnableHighPrecision` or something like that.

Thoughts @mrhelmut @harry-cpp ?

Mindfulplays:
Both the comments - @tomspilman make sense, will address soon. Thanks!

Mindfulplays:
> @Mindfulplays
> 
> I have a few issues with this PR.
> 
> One i really don't like having two event paths thru the whole code base... touch events and coalesced touch events. Especially since it seems coalesced touch has the same data as a normal touch event.
> 
> So maybe it just needs to be a flag on `TouchLocation`? It goes thru the same code path, but is flagged as these special high precision touch data.

Done. 
> 
> Also i don't like calling this "coalesced" touches... it just perpetuates the bad naming from Apple. This should be called `TouchPanel.EnableHighPrecision` or something like that.
> 
Done. High-Precision vs High-Frequency, going with High-frequency.
(Spatial vs temporal) 
Although Windows used to call the umbrella touchpad improvements as Precision Touch so ymmv!

> Thoughts @mrhelmut @harry-cpp ?

MonoGame.Framework/Input/Touch/TouchLocation.cs
```diff
@@ -1,7 +1,7 @@
 #region License
 // /*
 // Microsoft Public License (Ms-PL)
-// MonoGame - Copyright � 2009-2010 The MonoGame Team
+// MonoGame - Copyright � 2009-2010 The MonoGame Team
```
tomspilman, line 4: What happened here @Mindfulplays ?

MonoGame.Framework/Input/Touch/TouchLocation.cs
```diff
@@ -253,6 +262,12 @@ internal bool UpdateState(TouchLocation touchEvent)
             return _state != _previousState || delta.LengthSquared() > 0.001f;
         }
 
+
+        public bool IsHighFrequencyEvent()
```
tomspilman, line 266: This needs reference docs.

MonoGame.Framework/Input/Touch/TouchPanel.cs
```diff
@@ -152,5 +157,16 @@ public static bool IsGestureAvailable
         {
             get { return PrimaryWindow.TouchPanelState.IsGestureAvailable; }
         }
+
+        /// <summary>
+        /// Gets or sets if high-frequency events are sent to any listeners (if the underlying OS supports it).
+        /// If <see langword="false"/>, no additional CPU usage is incurred i.e. no high-frequency events are obtained
+        /// from the underlying platform.
```
tomspilman, line 164: This should probably mention what the default setting is.

MonoGame.Framework/Input/Touch/TouchPanelState.cs
```diff
@@ -312,6 +335,11 @@ public int DisplayWidth
 
         public bool EnableMouseGestures { get; set; }
 
+        /// <summary>
+        /// Gets or sets if high-frequency touch event processing is enabled.
+        /// </summary>
+        public bool EnableHighFrequencyTouch { get; set; }
```
tomspilman, line 341: Maybe add a '@see` reference to the other doc.

tomspilman:
Hey @Mindfulplays 

There are some changes to readme.md files in this PR.  Is that in other PR somewhere?  Are we waiting for that to be merged?

Mindfulplays:
> Hey @Mindfulplays
> 
> There are some changes to readme.md files in this PR. Is that in other PR somewhere? Are we waiting for that to be merged?

Apologies, non-familiarity with gh workflows, accidental, removed.

MonoGame.Framework/Input/Touch/TouchLocation.cs
Mindfulplays: Codepage issues (apparently, VS Code says UTF-8!), just made it (C) :) note that GH diff shows/has codepage issues in the 'before' too for this file and the other ones, so fixed both.

Mindfulplays:
Please take a look, I resolved your comments.

SimonDarksideJ:
I take it @Mindfulplays there is not an Android equivalent path for Android? (as this PR only focuses on iOS)

Mindfulplays:
There is: we can get History of touch events in Android.  I will send a separate patch for that.

Similarly we have DirectManipulation in Windows but that might be a moot point as I am not sure if it's useful in the context of Windows tablets  as a market compared to mobile. (Thoughts?)

Mindfulplays:
Also am not sure if there is such a thing on Consoles (may be Switch or Steam Deck?), but I don't have access to Consoles SDKs so currently wouldn't be able to dabble in it.

SimonDarksideJ:
If it is available on Android, then it would make sense @Mindfulplays to have parity there.
As for Windows touch, tbh, I'm not sure we need to bother unless it is easy to implement as it is not a touch-first platform like iOS / Android.

As for consoles, only Switch really would be compatible (possibly steam deck) but again, they are not touch-first devices, so I do not think we need to worry about those.

tomspilman:
This is one of those features where it may only be available on a couple of platforms.

We should come up with a way to document what platforms these features work on.

Is there some sort of reference document tag we can use for this?

If not maybe some sort of common language in reference docs like:

```
This feature is available on iOS, Android.
```

We would just need to be sure to keep this up to date.

Mindfulplays:
Added Android support + doc in https://github.com/MonoGame/MonoGame/pull/8109

Issue #8074, opened 2023-09-19
The usage of layerDepth argument in SpriteBatch.Draw is not properly documented


KirylPrykhodzka:
I could not find any mention on the website that when using overloads of SpriteBatch.Draw that have layerDepth argument, layerDepth has to be between 0 and 1.
It would be nice if documentation / method summary specified it.

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

stromkos:
That is not true.  Some XNA documentation may present it that way, but monogame accepts any layer depth values.

The values are used as the Z component of the generated quad, so in a mixed 3D and Spritebatch environment, it is better to use a RenderTarget for the 2D items to render an invariant HUD.

KirylPrykhodzka:
Upon some further investigation, it looks like I indeed was not totally correct. Please see more information below.

I have created a default MonoGame project for testing and added the following code:

```
private Color[] colors = new Color[]
{
    Color.Red,
    Color.Green,
    Color.Blue,
    Color.Yellow,
    Color.Purple,
};

protected override void Draw(GameTime gameTime)
{
    GraphicsDevice.Clear(Color.CornflowerBlue);

    _spriteBatch.Begin();

    for (int i = 0; i < 5; i++)
    {
        var texture = new Texture2D(GraphicsDevice, 1, 1);
        texture.SetData(new Color[] { colors[i] });
        _spriteBatch.Draw(
            texture,
            new Rectangle(100 + i * 10, 100 + i * 10, 100, 100),
            null,
            Color.White,
            0f,
            Vector2.Zero,
            SpriteEffects.None,
            i / 10);
    }

    _spriteBatch.End();

    base.Draw(gameTime);
}
```
Notice how layerDepth (last argument in `_spriteBatch.Draw` is `i / 10` which ensures that it is between 0.0 and 0.4.
When I run this code, the following image is displayed:

![выява](https://github.com/MonoGame/MonoGame/assets/38721357/e38fe886-e6ef-4c9a-9ebd-94ef1ae48f9e)

However, if I change layerDepth from `i / 10` to `i`, the following result is drawn:
![выява](https://github.com/MonoGame/MonoGame/assets/38721357/a4c1f215-b97b-4c7a-bca7-9a8778c8841e)

So I think there is something going on with layerDepth, or I am doing something wrong. Please help me understand this

[...]

PR #8093, opened 2023-11-16
Update SpriteEffect.cs to honor the Viewport settings

stromkos:
Added position and min/max depth values from the GraphicsDevice.Viewport.
See #8074

### DIFF
--- a/MonoGame.Framework/Graphics/Effect/SpriteEffect.cs
+++ b/MonoGame.Framework/Graphics/Effect/SpriteEffect.cs
@@ -63,12 +63,14 @@ namespace Microsoft.Xna.Framework.Graphics
         protected internal override void OnApply()
         {
             var vp = GraphicsDevice.Viewport;
-            if ((vp.Width != _lastViewport.Width) || (vp.Height != _lastViewport.Height))
+            if (vp.X != _lastViewport.X || vp.Y != _lastViewport.Y || 
+                vp.Width != _lastViewport.Width || vp.Height != _lastViewport.Height || 
+                vp.MinDepth != _lastViewport.MinDepth || vp.MaxDepth != _lastViewport.MaxDepth)
             {
                 // Normal 3D cameras look into the -z direction (z = 1 is in front of z = 0). The
                 // sprite batch layer depth is the opposite (z = 0 is in front of z = 1).
-                // --> We get the correct matrix with near plane 0 and far plane -1.
-                Matrix.CreateOrthographicOffCenter(0, vp.Width, vp.Height, 0, 0, -1, out _projection);
+                // --> We get the correct matrix with near plane 0 and far plane -1 (MinDepth, -MaxDepth).
+                Matrix.CreateOrthographicOffCenter(vp.X, vp.X + vp.Width, vp.Y + vp.Height, vp.Y, vp.MinDepth, -vp.MaxDepth, out _projection);
 
                 if (GraphicsDevice.UseHalfPixelOffset)
                 {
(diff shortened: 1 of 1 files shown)


mrhelmut:
This seems correct, though it'd be best to wait for graphical unit tests to come back online before we merge just to make sure.

Mindfulplays:
Are we sure this is correct? `glViewport` (or any platform Viewport call such as Metal/DX) should apply `vp.X/Y` etc already. And hence, the original code is correct. Note that the backbuffer that the sprite effect renders onto has the correct width/height within the viewport (not the screen). So the shader correctly multiplies normalized coords with the matrix W/H.

(Viewport X/Y is generally 0,0 so it's hard to test this. But, on Android, `GraphicsDevice::ResetClientBounds` explicitly sets viewport with a x/y offset due to safe-area/cutouts - which is a common enough use case so you can verify this quickly) 

I recommend not touching position.

SimonDarksideJ:
Responses to feedback @stromkos ?

mrhelmut:
We really need to test this.

stromkos:
>I recommend not touching position.

For all current applications the defaults are fine and respected in this PR. 

This PR only **allows** changes to be reflected. Not that they would be needed for most use cases.  

The fixed values passed are from initial implementations and do not follow XNA documentation. This PR corrects that fixed-value test condition.

Peru-S:
There are two things here: Viewport position and depth. Your PR touches both. I reckon not touching the viewport position as if it ain't broke, don't fix it (and I reckon it will actually break in subtle ways - I am not sure if you tested all the corner cases esp. Android).

Passing depth seems fine. 

Perhaps move the position part into a separate PR if you feel strongly and have tested it?

stromkos:
@Peru-S 
Thank you for making me take a closer look. What should have been consistent and standardized across implementations was not at all.

The entire ViewPort implementation is completely messed up. For most purposes the value should be read-only since it is   either ignored or overwritten at seemingly random times. Yes, the events/calls can be determined at runtime...

The Width and Height values are set to the presentation parameters' values sometimes.

Game wide:

Any use of render targets ignores all values.  
To/from full screen sets defaults.


IOS ignores X,Y for all operations and sets height and width to the screen:

https://github.com/MonoGame/MonoGame/blob/ca3b1de568462654098b8a4da55fe68f23cdf79c/MonoGame.Framework/Platform/iOS/iOSGameView.cs#L195

and again here:
 
https://github.com/MonoGame/MonoGame/blob/ca3b1de568462654098b8a4da55fe68f23cdf79c/MonoGame.Framework/Platform/iOS/iOSGameView.cs#L225-L228

DirectX works with X,Y:

https://github.com/MonoGame/MonoGame/blob/ca3b1de568462654098b8a4da55fe68f23cdf79c/MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs#L570-L577

The base OpenGL is different( Y is inverted ): 

https://github.com/MonoGame/MonoGame/blob/ca3b1de568462654098b8a4da55fe68f23cdf79c/MonoGame.Framework/Platform/Graphics/GraphicsDevice.OpenGL.cs#L560

Peru-S:
> The entire ViewPort implementation is completely messed up. For most purposes the value should be read-only since it is either ignored or overwritten at seemingly random times. Yes, the events/calls can be determined at runtime...

I don't think it's 'messed up' - MonoGame has evolved to support a wide variety of platforms while still maintaining API compatibility. For some platforms such as Android/iOS the concept of a viewport/`Window` are often blurry as it's really not the same as in DesktopGL/DX. 

So I think it's actually a strength of MG that these calls are plumbed without much workarounds for these platforms.

stromkos:
@Peru-S 
Yes, I agree.  I just would have liked a consistent approach when possible, and in this case it was possible just not done.

Take textures  as an example, we can use whatever format we like, but it will be converted to an array of `Color`s then a byte array of the components, the the final conversion takes place. `Alpha8` becomes `R8` on GL platforms.

>So I think it's actually a strength of MG that these calls are plumbed without much workarounds for these platforms.

There is much less platform parity than you think. Yes these are minor things that add up. Where are Emojis??

PR #8105, opened 2023-12-15
Fix DesktopGL function loader ILC error

mrhelmut:
Updated the function loader for DesktopGL project to not crash on macOS when compiling with ```PublishAot```.


Mindfulplays:
LGTM

PR #8111, opened 2023-12-23
Fix Android song unable to find a specified URI

Mindfulplays:
Changed Android `Song` to differentiate between Assets/URI-based `DataSource`.

@tomspilman @mrhelmut

### DIFF
--- a/MonoGame.Framework/Platform/Media/Song.Android.cs
+++ b/MonoGame.Framework/Platform/Media/Song.Android.cs
@@ -77,15 +77,22 @@ namespace Microsoft.Xna.Framework.Media
 
             if (assetUri != null)
             {
+                // Check if we have a direct asset URI.
                 _androidPlayer.SetDataSource(MediaLibrary.Context, this.assetUri);
+            }
+            else if (_name.StartsWith("file://"))
+            {
+                // Otherwise, check if this is a file URI.
+                _androidPlayer.SetDataSource(_name);
             }
             else
             {
-                var afd = Game.Activity.Assets.OpenFd(_name);
-                if (afd == null)
-                    return;
-
-                _androidPlayer.SetDataSource(afd.FileDescriptor, afd.StartOffset, afd.Length);
+                // Otherwise, assume it's a file path. (This might throw if the file doesn't exist)
+                var afd = Game.Activity?.Assets?.OpenFd(_name);
+                if (afd != null)
+                {
+                	_androidPlayer.SetDataSource(afd.FileDescriptor, afd.StartOffset, afd.Length);
+                }
             }
 
 


mrhelmut:
Not a big fan of a try/catch within a try/catch and exception being silenced.

Suggestion:

```csharp
var afd = Game.Activity.Assets.OpenFd(_name);
if (afd != null)
     _androidPlayer.SetDataSource(afd.FileDescriptor, afd.StartOffset, afd.Length);
else
     _androidPlayer.SetDataSource(_name); // assuming that if anything is wrong here, the user did something wrong with their URI and we shouldn't silently handle the exception to let the user know
```

Mindfulplays:
> Not a big fan of a try/catch within a try/catch and exception being silenced.
> 
> Suggestion:
> 
> ```cs
> var afd = Game.Activity.Assets.OpenFd(_name);
> if (afd != null)
>      _androidPlayer.SetDataSource(afd.FileDescriptor, afd.StartOffset, afd.Length);
> else
>      _androidPlayer.SetDataSource(_name); // assuming that if anything is wrong here, the user did something wrong with their URI and we shouldn't silently handle the exception to let the user know
> ```

`OpenFd` throws if the file is not found (it's for a file I am `Play`ing outside the `ContentManager` system). I do agree its ugly but this lets me use MG Song with my own files (I wonder if this is useful at all to others? If not I can close this PR).

Mindfulplays:
Squashed commits to ease PR review, thanks!

dellis1972:
Can we make use of `Uri.TryCreate` to query if the argument is a Uri then call the appropriate method?

`Uri.TryCreate ("http://foo", UriKind.Absolute, out Uri result); `

Mindfulplays:
> Can we make use of `Uri.TryCreate` to query if the argument is a Uri then call the appropriate method?
> 
> `Uri.TryCreate ("http://foo", UriKind.Absolute, out Uri result); `
> 
> This code above will return `true` in this case, but `false` if you just passed say `Assets/foo.xnb`

This is an Android problem not an dotnet `Uri` problem :( Android assumes that you know that you know the file you are accessing. The Uri itself would still be considered [valid either way](https://source.dot.net/#System.Private.Uri/System/UriExt.cs,622).

Or perhaps I am missing something else?

dellis1972:
So the Pr description says this is about knowing the difference between an "Asset" and a "Uri".
The problem is `Game.Activity.Assets.OpenFd(_name);` will throw if passed a Uri rather than an "Asset". 
Things retrieved via the ContentManager do not usually have a Uri based filenames if I remember rightly. Its more like "Assets/songs/mysong.xnb" , which `Uri.TryCreate` will return `false` for. 
If that is the case we could do 

```
if (!Uri.TryCreate ("http://foo", UriKind.Absolute, out Uri result)) {
    var afd = Game.Activity.Assets.OpenFd(_name);
    if (afd != null)
       _androidPlayer.SetDataSource(afd.FileDescriptor, afd.StartOffset, afd.Length);
} else {
    _androidPlayer.SetDataSource(_name);
}
```
I could be wrong about the Uri, its been a while since took a look at that code. 

On a side note its not really good practice to swallow exceptions in a library if you can help it. There are a number of reasons why a file might not exist. The main one on Android is someone using the incorrect Filename casing as its a case sensitive file system. If we swallow the exception the developer will have no idea what happened in code, the only hint would be the Song not playing. Not everyone will read the `adb logcat` output to figure out what is happening. 

So you have some samples of the kinds of values that get passed to `_name` in this case? one for assets and the other for direct file access?

mrhelmut:
Couldn't we just have something like ```_name.StartsWith("file://")``` to differentiate content from external files?

```csharp
            if (assetUri != null)
            {
                _androidPlayer.SetDataSource(MediaLibrary.Context, this.assetUri);
            }
            else if (_name.StartsWith("file://"))
            {
                _androidPlayer.SetDataSource(_name);
            }
            else
            {
                var afd = Game.Activity.Assets.OpenFd(_name);
                if (afd == null)
                    return;

                _androidPlayer.SetDataSource(afd.FileDescriptor, afd.StartOffset, afd.Length);
            }
```

Mindfulplays:
Good idea @mrhelmut fixed, please take a look.

dellis1972 (approved):
Looks ok to me 👍

dellis1972:
@mrhelmut can we get one last look at this one please? If its good, I'll merge it.

Mindfulplays:
Gentle ping...

mrhelmut:
I'm a bit on the fence with the unnecessary changes I commented.

Mindfulplays:
FWIW, the code 'works' :) but yeah, lmk if it might break existing code.

SimonDarksideJ:
> > Not a big fan of a try/catch within a try/catch and exception being silenced.
> > Suggestion:
> > ```cs
> > var afd = Game.Activity.Assets.OpenFd(_name);
> > if (afd != null)
> >      _androidPlayer.SetDataSource(afd.FileDescriptor, afd.StartOffset, afd.Length);
> > else
> >      _androidPlayer.SetDataSource(_name); // assuming that if anything is wrong here, the user did something wrong with their URI and we shouldn't silently handle the exception to let the user know
> > ```
> 
> `OpenFd` throws if the file is not found (it's for a file I am `Play`ing outside the `ContentManager` system). I do agree its ugly but this lets me use MG Song with my own files (I wonder if this is useful at all to others? If not I can close this PR).

Possibly then, also include a `File.Exists` check to ensure the file is present and accessible.

Can you also add a Unit test around this feature (or validate with an existing test)? @Mindfulplays

Mindfulplays:
> > > Not a big fan of a try/catch within a try/catch and exception being silenced.
> > > Suggestion:
> > > ```cs
> > > var afd = Game.Activity.Assets.OpenFd(_name);
> > > if (afd != null)
> > >      _androidPlayer.SetDataSource(afd.FileDescriptor, afd.StartOffset, afd.Length);
> > > else
> > >      _androidPlayer.SetDataSource(_name); // assuming that if anything is wrong here, the user did something wrong with their URI and we shouldn't silently handle the exception to let the user know
> > > ```
> > 
> > 
> > `OpenFd` throws if the file is not found (it's for a file I am `Play`ing outside the `ContentManager` system). I do agree its ugly but this lets me use MG Song with my own files (I wonder if this is useful at all to others? If not I can close this PR).
> 
> Possibly then, also include a `File.Exists` check to ensure the file is present and accessible.

`File.Exists` does not work with file descriptor (`OpenFd` is an Android/seLinux-style API; not dotnet-oriented).

> 
> Can you also add a Unit test around this feature (or validate with an existing test)? @Mindfulplays

I don't see Android unit-tests (or I guess Robolectric/Espresso tests). Are there tests setup already?
Might be a worthwhile bounty-like way to add espresso tests that run on emulators IMO... (I have some experience doing it for my projects, but I don't have bandwidth to add to MG as it's a very time-consuming step TBH).

SimonDarksideJ (approved):
Approved, we will look into a bounty to encourage adding more unit testing

dellis1972:
@Peru-S for unit testing do you have a sample of the usage of the uri for a song ?

Peru-S:
Example:
`Song: Playing 1: URI: file://content/test/test.m4a File: Content/test/test.m4a `

For a `SomeAndroidProj.csproj` containing 
```
  <AndroidAsset Include="Assets\Content\test\**" />
```

dellis1972:
@Peru-S 

I can't seem to get this to work. 

In the apk I have ` 257347  Stored   257347   0% 01-01-1980 01:00 ca748fa5  assets/Content/Logo.m4a`

I'm trying this code

```csharp
var song2 = Song.FromUri ("foo", new System.Uri ("file://content/Logo.m4a"));
MediaPlayer.Play(song2);
```

If i copy the data into the app directory manually it then works. 

```
adb push test.droid/Content/bin/Android/Content/Logo.m4a /data/local/tmp/logo.m4a
adb shell run-as test.droid.test.droid cp /data/local/tmp/logo.m4a files/logo.m4a
```

```csharp
var song2 = Song.FromUri ("foo", new System.Uri ("file:///data/user/0/test.droid.test.droid/files/logo.m4a"));
MediaPlayer.Play(song2);
```

Any ideas?

Peru-S:
Android (Linux) is case-sensitive. Wondering if there's a lower case 'c' there?

Otherwise I will try to whip up the Interactive Tests with Android song test this week.

dellis1972:
@Peru-S I know its case-sensitive. I tried all sorts of combinations to get this to work as you described in https://github.com/MonoGame/MonoGame/pull/8111#issuecomment-2486639574. 

For context I'm trying to fix a regression in Android Song playback and this is causing me issues as I don't have a solid example of usage.

Peru-S:
I did a `Assets.List()` and dumped all the files...

```
-Asset Content/test/Music/test.m4a / file://content/test/Music/test.m4a
```

Then used:  `var song = Song.FromUri("foo", new Uri("file://content/test/Music/test.m4a");`

which should go through the fix I have in this commit. i.e. `_name` has `file://` prefix and hence should set the data-source correctly.

1. Could you try listing the files using `Assets.List("Content/")` fully to ensure it's actually there? (Sometimes, `adb push` does some "clever" stuff and borks)
2. Does it go through the `else if (_name.StartsWith("file://"))` in `Song.Android.cs`?

Trying to see where the mismatch is...

PR #8112, opened 2023-12-25
Fixing int overflow exception

devedse:
So I'm building a 3d Maze viewer that can render "big" mazes. I ran into an issue where if I stuff the whole maze into a single buffer we run into a buffer overflow exception since the startIndex * elementSize > int.MaxValue.

I've done an attempt in solving this but this needs some thorough review. To fully fix this we probably also need to modify the method header where we pass in the bufferSize, but I'm not sure if we're allowed to edit that.

image


mrhelmut:
Vertex buffers can't be this big, so no matter which checks you bypass with a long cast, things are going to crash further down the line (and in most of your changes, it will on the very next line because there's a cast to IntPtr, which is 32bit addressing).

The actual fix to this would be to split your rendering into batches of ```int.MaxValue / vertexDataSize```. That's what any renderer would do.

But that'd be problematic anyway because you're basically sending multiple 2GB vertex buffers to the GPU per frame (so 120GBps worth of vertex buffer at minimum, while [Android's recommandation is 1.5GBps](https://developer.android.com/agi/sys-trace/vertex-memory-bw#counter_analysis) -- assuming that you're targeting Android based on your screenshot).
Assuming that your mazes are entirely static, you might want to break down parts of your mazes into meshes, and make use of hardware instancing (in addition to batching your GPU uploads) to optimize the rendering and reduce the bandwidth usage. You might also want to look into frustum culling and limiting distant parts to be displayed.

harry-cpp:
> which is 32bit addressing

Just a minor correction, `IntPtr` size is architecture dependant actually, you can't really count on it being any specific size.

devedse:
Ok thanks for the explanation, I do think it could use a nice "error" / "fix" to ensure we solve the buffer overflow error as leaving this could potentially result in undefined behaviour. (e.g. the Pointer pointing to a piece of memory in the GPU that's not part of this draw call maybe?)

PR #8117, opened 2024-01-09
Vector rotation helper functions in Vector2

itsBuggingMe:
First pull request here, please don't bite!

Rotating vectors is somewhat common, and I was surprised when Monogame's way (besides creating one's own function) is through matrices. For instance, [this](https://community.monogame.net/t/solved-rotate-vector-by-random-max-degree/9032/2) uses a 3D matrix just to rotate a 2D vector. Or [here](https://gamedev.stackexchange.com/questions/6551/rotate-vector-by-matrix), where a Matrix is created just to rotate. This seems particularly wasteful for most applications, and also somewhat unintuitive. I believe that a few helper functions will make it much easier to deal with rotation.

### DIFF
--- a/MonoGame.Framework/MathHelper.cs
+++ b/MonoGame.Framework/MathHelper.cs
@@ -322,41 +322,5 @@ namespace Microsoft.Xna.Framework
 	    {
 	         return (value > 0) && ((value & (value - 1)) == 0);
 	    }
-
-        /// <summary>
-        /// Rotates a vector by the specified number of radians
-        /// </summary>
-        /// <param name="value">The vector to be rotated.</param>
-        /// <param name="radians">The amount to rotate the vector.</param>
-        /// <returns>A rotated copy of value.</returns>
-        /// <remarks>
-        /// A positive value and negative value
-        /// would rotate counterclockwise and clockwise,
-        /// respectively
-        /// </remarks>
-        public static Vector2 RotateVector(Vector2 value, float radians)
-        {
-            float cos = MathF.Cos(radians);
-            float sin = MathF.Sin(radians);
-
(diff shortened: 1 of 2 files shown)


mrhelmut:
Hello, thank you for your first contribution.

It is indeed a bit convoluted to have to create a matrix to do those possibly frequent and useful operations. I would be favorable to have them directly inside the ```Vector2``` type (both static and non static versions) as ```Rotate()``` and ```RotateAround()```.

Most users would add those as Vector2 extension methods anyway.

itsBuggingMe:
I have moved the functions to Vector2 and made non static versions. Few questions - should the non static functions return copies of themselves? Or should the non static functions keep the original vector intact and return a copy.  Should I also add versions for Vector3? Thank you!

mrhelmut:
Yes, the static versions should return copies and not modify the input values.

I don't have strong opinions in regard to Vector3 or Vector4 because I assume that these would be mostly used on 3D games which are more likely to already be knees deep into Matrix/Quaternion manipulations so there might be not much comfort gain with those types.

I leave other reviewers be judges. Otherwise good to go to me.

Brett208:
@mrhelmut Is it recommended that change like this adding a helper function include unit test(s) to cover its functionality? Asking a bit out of my own curiosity.

Happy to see the documentation spelling improvements riding along as well.

mrhelmut:
> @mrhelmut Is it recommended that change like this adding a helper function include unit test(s) to cover its functionality? Asking a bit out of my own curiosity.
> 
> Happy to see the documentation spelling improvements riding along as well.

We should aim for the best coverage, but it's not an hard requirement.

mrhelmut:
Those changes are good to me. Merging. Thanks!

Issue #7992, opened 2023-02-24
Game.IsActive is always `true` on launch, even if the window doesn't have focus


bwiklund:
The repro steps for this are quick:

Make a new game from the desktopgl template:
```dotnet new mgdesktopgl -o MyGame```

On `Game1.cs:45`, change the drawing code so we can get some visual feedback on the state of `Game.IsActive`:
```GraphicsDevice.Clear(IsActive ? Color.CornflowerBlue : Color.OrangeRed);```

Launch the game (dotnet run or a build), then quickly at tab to some other window and click on it to give it focus.

Here it is behind a chrome window, for example:

![image](https://user-images.githubusercontent.com/3586801/221213222-3b6b767f-3f38-4778-9e62-77f3f4e98035.png)

It's blue, showing that `IsActive` is true even though we have never given the window focus.

Tabbing in and then out correctly turns the window red, as `IsActive` is now correctly false.

![image](https://user-images.githubusercontent.com/3586801/221213645-cd8dd9ab-f216-40c5-bab6-7bf7444667f7.png)

I've seen several google results for this bug going back years, but none of the commonly quoted solutions actually fix the problem (checking window handles, registering our own events, etc). it would be awesome if this could be fixed at the source

@Apostolique recommended including this SDL2 file, and doing this workaround:

`https://github.com/flibitijibibo/SDL2-CS/blob/master/src/SDL2.cs`

```
SDL_WindowFlags flags = (SDL_WindowFlags)SDL_GetWindowFlags(Window.Handle);
Console.WriteLine(flags.HasFlag(SDL_WindowFlags.SDL_WINDOW_INPUT_FOCUS));
```

This fixes the bug: is it possible monogame could check this way instead? Currently it's listening to events, which means it misses the initial loss of focus before the app has fully started.

https://github.com/MonoGame/MonoGame/blob/2fa596123f834f322be19811d97fe6c20616d570/MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs#L218-L223

- MonoGame 3.8.1.303
- Windows
- DesktopGL

**Comment by bwiklund** [comment]
@Apostolique suggested this solution as a workaround:

Adding this SDL file to the project: `https://github.com/flibitijibibo/SDL2-CS/blob/master/src/SDL2.cs`

And checking focus state instead like this:
```
  public bool IsActiveFixed {
    get {
[...]

PR #8123, opened 2024-01-17
Fix active state event on startup

mrhelmut:
Need testing.

Fixing #7992

### DIFF
--- a/MonoGame.Framework/Platform/SDL/SDL2.cs
+++ b/MonoGame.Framework/Platform/SDL/SDL2.cs
@@ -392,6 +392,33 @@ internal static class Sdl
             public IntPtr window;
         }
 
+        [Flags]
+        public enum Flags : uint
+        {
+            Fullscreen = 0x00000001,
+            FullsceenDesktop = (Fullscreen | 0x00001000),
+            OpenGL = 0x00000002,
+            Vulkan = 0x10000000,
+            Metal = 0x10000000,
+            Shown = 0x00000004,
+            Hidden = 0x00000008,
+            Borderless = 0x00000010,
+            Resizable = 0x00000020,
+            Minimized = 0x00000040,
+            Maximized = 0x00000080,
+            InputGrabbed = 0x00000100,
+            InputFocus = 0x00000200,
+            MouseFocus = 0x00000400,
+            Foreign = 0x00000800,
+            AllowHighDPI = 0x00002000,
+            MouseCapture = 0x00004000,
+            AlwaysOnTop = 0x00008000,
+            SkipTaskbar = 0x00010000,
+            Utility = 0x00020000,
+            Tooltip = 0x00040000,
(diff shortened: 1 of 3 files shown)


gluser2:
> Need testing.

I patched this into my custom fork of MonoGame and it works beautifully for DesktopGL! The window comes up behind my active window (taskmgr.exe), and as soon as it grows to the correct size window, it turns red (using [bwiklund](https://github.com/bwiklund)'s !iGame.IsActive code) to show that it recognizes it's inactive.

Could a similar fix be made for WindowsDX, which appears to have the same problem? It looks like MonoGame.Framework\Platform\Windows has a WinFormsGamePlatform.cs, which may function similarly to SDL\SDLGamePlatform.cs.

gluser2:
Your new patch defaults to active state on 3 of 3 tests I did, which is what it was intended to do. Is there no way to help the WinForms event pump detect the right startup state?

mrhelmut:
@gluser2 I've applied the same to WindowsDX. Would you mind testing that too?

gluser2:
On both computers I tested*, your patch still starts an inactive game in an active state, just like MG 3.8.1 and the develop branch do. However, your patch also keeps MonoGame in an active state even when Alt+Tab is pressed or another window is drawn on top of it. I don't know why it's behaving this way, but the game seems to never enter an inactive state now.

* AMD A8-7600 x64 w/ AMD R9 255 graphics, and Qualcomm 7C ARM64 w/ Adreno 618 graphics

mrhelmut:
Is this with DesktopGL or WindowsDX?

gluser2:
The DesktopGL fix worked perfectly, but the one I reported on in my last comment was WindowsDX.

I made a new DX project that referenced the source code of your commit, a second that referenced the source of the develop branch, and a third using the standard 3.8.1.303 release NuGet (no source), all in Visual Studio 2022. They were built on the x64 machine and run on both that and the arm64. The behavior was the same on both.

OSes were Windows 10 and 11. .NET 8.0.1 and WinForms were used to draw the MG windows. The new DX projects were a created template from 3.8.1, since I didn't see one in the source files. I only modified the games to draw Red instead of CornflowerBlue while the window is inactive.

I open the 3.8.1 or develop built game and click another window like Windows Task Manager. When the game launches, it should be inactive (red). But it's still blue under the other window. After clicking on it for the first time, anytime you switch windows it turns red like it should.

The game I built on your commit (with the modified WinFormsGameWindow.cs) launches blue when inactive, like the other ones. But in contrast, no matter how often I click the window and back on Task Manager, it never turns red. It seems like it's stuck in an active (blue) state.

Do you have a different setup? How does the patch perform on your machine?

mrhelmut:
Closing because no traction and no time to check if this is even possible consistently across all implementations. If anyone wishes to takeover, please do feel free.

PR #8144, opened 2024-01-30
Remove unused compression libraries

mrhelmut:
There are 2 different implementations of ```GZipStream``` in MonoGame, and none of them is actually used since we moved to ```StbImage``` to decode PNGs.

Moreover, one of them is ```public``` while it definitely should have been ```internal```. This has pushed the library documentation to the official MonoGame documentation while it shouldn't have.

If it happens that users have used the public implementation within MonoGame, they can safely switch to [DotNetZip](https://www.nuget.org/packages/DotNetZip/) or [System.IO.Compression](https://learn.microsoft.com/fr-fr/dotnet/api/system.io.compression.gzipstream?view=net-8.0). Both of them have an API-compatible ```GZipStream``` implementation that is safer to use.


tomspilman:
Yeah i bet at the time we figured why not make it public so that others can use it without pulling in another dependency, but that came back to haunt us.

EtichBruh:
Thanks, now i can't use 1 line of code to decompress array of bytes..........

mrhelmut:
You know it's just a one-line change for you, right? Just add ```using System.IO.Compression;``` or reference ```DotNetZip``` and you won't have to change anything to your code.

PR #8145, opened 2024-01-31
Update requirements.md to new site and .net 8

JakeLegendXIII:
- Update platform link to new site.
- Update required .NET to 8 and required Visual Studio to 17.8

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -1,10 +1,10 @@
-Depending on the [platform](Documentation/articles/platforms/0_platforms.md) that you are targeting, MonoGame has different sets of requirements.
+Depending on the [platform](https://monogame.net/articles/platforms/) that you are targeting, MonoGame has different sets of requirements.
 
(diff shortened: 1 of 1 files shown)


mrhelmut:
This only applies to the develop branch and not the public release, so we'll have to hold to it until the next release.

JakeLegendXIII:
> This only applies to the develop branch and not the public release, so we'll have to hold to it until the next release.

Sounds good. The reason I had submitted it now is that requirements.md is linked from the readme in the instructions on how to build from source [here](https://github.com/MonoGame/MonoGame?tab=readme-ov-file#source-code). When trying to build develop from source the requirements are different than master or the released packages, but it obviously has more context than that so makes sense if you'd like to wait.

mrhelmut:
Ah yes, this makes sense. Sorry I read your PR too hastily and assumed it was about the website documentation and not the source.

I'll merge it now.

PR #8150, opened 2024-02-04
Update WindowsDX csproj to Microsoft.NET.Sdk

JakeLegendXIII:
Small cleanup on the project tag to match the other updated csprojs and remove the following warnings on MonoGame.Tests.WindowsDX.csproj:

- warning NETSDK1137: It is no longer necessary to use the Microsoft.NET.Sdk.WindowsDesktop SDK. Consider changing the Sdk attribute of the root Project element to 'Microsoft.NET.Sdk'.
- warning NETSDK1106: Microsoft.NET.Sdk.WindowsDesktop requires 'UseWpf' or 'UseWindowsForms' to be set to 'true'


mrhelmut:
Thanks!

Issue #8128, opened 2024-01-23
Repair WindowsDX tests


mrhelmut:
Since we moved to newer .NET runtimes and NUnit versions, WindowsDX tests are failing due to Direct3D/Winforms not working in that context. Non-graphical tests work normally.

We need to figure out what prevents them from initializing a Direct3D context and fix that.

Any help welcome.

JakeLegendXIII:
So I created a Draft PR #8151 that actually has the tests passing again, however not in an ideal or probably recommended way. Wanted to provide some information about what I've run into and tried. Assuming someone more familiar with SharpDX or the WindowsDX code in MonoGame may have some better ideas about what is going on. 

- This problem exists in the Master branch today as well. Rolling back versions of .NET, NUnit, and various other things did not seem to make much impact. The issue is still happening on the .NET 6 code even (unless this has been broken for a long time!)
- The Issue appears to be happening on GraphicsDevice.DirectX.cs when creating the Direct3D device via the SharpDX.Direct3D11.Device(driverType, creationFlags, featureLevels) constructor. Whenever you pass a driverType of Reference it throws this issue: _SharpDXException: HRESULT: [0x887A0004], Module: [SharpDX.DXGI], ApiCode: [DXGI_ERROR_UNSUPPORTED/Unsupported], Message: The specified device interface or feature level is not supported on this system._ If I simply disable the code in the TestGameBase.cs that sets UseReferenceDevice to true and make it use driverType Hardware all of the tests start passing again. Obviously the Reference device is supposed to be fully featured, and the ideal way to do testing. It bypasses a bunch of inconsistencies and unreliability of testing on different hardware, but it does not seem to be working currently.
- I did for giggles try a half baked upgrade of SharpDX to 4.2.0 and got about 50% there where the code could at least build and the tests could run. I was still seeing the fast device creation failures on tests that were not even impacted by the breaking API changes that were not fully fleshed out. Could be user error, but that did not seem to fix the issue.

I am not as familiar with how the Reference device creation worked via debugging prior to when the tests stopped working. Nor I am super familiar with SharpDX. While I can try to find some time to dig in a bit more, and get a better idea of what is happening I wanted to at least share some info. Anyone that has better ideas on how to fix this feel free to jump in or make suggestions!

mrhelmut:
Thank you for looking into this!

Ah! The reference driver is a software renderer for Direct3D, it is indeed meant to test accuracy and is very slow.

[...]

PR #8151, opened 2024-02-04
Fix the WindowsDX tests

JakeLegendXIII:
Creating as a Draft as these fixes are likely not ideal/final, and probably controversial. Assuming we can discuss here or on the issue. The Reference Driver type is not successfully being created by the call to the SharpDX.Direct3D11.Device constructor. It is throwing this error: _The specified device interface or feature level is not supported on this system._

Switching the test suite to run using the Hardware Driver Type makes all of the tests work on a variety of my Windows dev machines. However, there is one unrelated Adapter test that fails on machines using onboard graphics (but it also fails on Master on those machines today), so I added a comment to just identify that for now. It is does not actually use the TestGameBase.cs today and already executes with the Hardware driver type so it is not exactly the same issue.

This is in regards to Issue #8128 where I will be adding some research notes and further information that may hopefully lead to a better fix if this is not acceptable.

### DIFF
--- a/Tests/Framework/Graphics/GraphicsAdapterTest.cs
+++ b/Tests/Framework/Graphics/GraphicsAdapterTest.cs
@@ -53,6 +53,8 @@ namespace MonoGame.Tests.Graphics
                 Assert.IsNotNull(adapter.CurrentDisplayMode); 
                 Assert.IsNotNull(adapter.SupportedDisplayModes);
                 Assert.GreaterOrEqual(adapter.SupportedDisplayModes.Count(), 1);
+
+                // This Assert can fail on laptops or systems with onboard graphics.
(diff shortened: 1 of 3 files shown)


mrhelmut:
I've merged your other PR, so you might want to rebase this one. Sorry, I should have favor pushing that change into this PR instead, it would have avoided the rebase.

JakeLegendXIII:
Thanks @mrhelmut ! My rebase is a bit messy. Sorry about that! I could always and scrap this and re-introduce the changes in a cleaner PR if that would be better. I now have the merged commit as the one updating the csproj, but there is some other messy stuff in there.  

The requested change to switch to the WARP device instead of REF is now included. Works on my test machines!

PR #8153, opened 2024-02-04
Fix the WindowsDX tests

JakeLegendXIII:
This is a cleaner version of PR #8151 without some of my messy commits. This is to fix the issue #8128 

- Resolves the failing WindowsDX tests that are failing, because of the REF device. Using the WARP device does work for the tests and is better than my previous fix that relied on using Hardware.
- Also adds a comment to an Assert that I saw failures for on machines using onboard graphics. The GraphicsAdapterTest.cs does not use the TestGameBase.cs, and relies on the Hardware device for testing today so that is a separate intermittent/device specific issue.


mrhelmut:
Thanks!

Issue #8154, opened 2024-02-04
MGCB editor fails with a parsing error in latest "develop"


SimonDarksideJ:
### Prerequisites

- [X] I have verified this issue is present in the `develop` branch
- [X] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

3.8.1.617-develop

### Which MonoGame platform are you using?

N/A

### Operating System

Windows

### Description

In the current `develop` state, the MGCB tool fails to load a `.mgcb` configuration due to a parsing error.

In the current 3.8.1 release, the following is supported/permitted

```xml
/processorParam:ColorKeyColor={R:0 G:0 B:0 A:0}
```

However, now only the following is supported:

```xml
/processorParam:ColorKeyColor=0,0,0,0
```

So something has changed which has broken the backwards compatibility 

### Steps to Reproduce

- Create a new `.mgcb` content package using the release MGCB editor
- Add an FBX file to the content
- Save and close the MGCB editor
- Check the file contents and update the following:
    ```xml
    /processorParam:ColorKeyColor=0,0,0,0
    ```

  To
    ```xml
    /processorParam:ColorKeyColor={R:0 G:0 B:0 A:0}
    ```
  (Which is a previous XNA supported format, and also supported by the Release version of 3.8.1)

- Re-open the `.mgcb` file and confirm it is working
- Close the `.mgcb` file and reopen using the latest Develop version of the MGCB editor (3.8.1.617-develop).
- Observe the file can no longer be opened.

### Expected Behavior

The format
```xml
/processorParam:ColorKeyColor={R:0 G:0 B:0 A:0}
```

Should be supported for XNA backwards compatibility

### Resulting Behavior

MGCB editor fails to load

SimonDarksideJ:
Actually, in further testing, the 3.8 mgcb editor also fails with the following error:
> System.ArgumentException: Could not convert from string({R:0 G:0 B:0 A:0}) to Color, expected format is 'r,g,b,a'

So this is more of an XNA regression issue.

mrhelmut:
Likely an easy fix. Color parsing happens here:

[...]

PR #8167, opened 2024-02-09
Fix MGCB Color parsing error on ColorKeyColor

JakeLegendXIII:
This is a fix for #8154 . There are multiple ways to do this, and I even had a regex version... But I went with kind a simpler more old school way. Please feel free to make any suggestions or request any changes to it!

- Added code that checks for the XNA format {R:0 G:0 B:0 A:0} and parses it into a Color. If the format is the r,g,b,a format it is handled as it was before. 
- Updated the exception message to include both formats.

-Unrelated to this change specifically, but I updated the Help link in the MGCB Editor while I was in there to correctly point to the website after the latest changes.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/Convertors/StringToColorConverter.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/Convertors/StringToColorConverter.cs
@@ -40,7 +40,6 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Builder.Convertors
             if (value.GetType() == typeof(string))
             {
                 string strValue = (string)value;
-                int r, g, b, a;
 
                 // Check if the string is in the older XNA "{R:0 G:0 B:0 A:0}" format
                 if (strValue.StartsWith('{') && strValue.EndsWith('}'))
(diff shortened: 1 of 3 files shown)


mrhelmut:
It looks good to me! Thanks for grabbing it!

@SimonDarksideJ do you happen to know if the format can be a mix like ```{A:0 B:0 G:0 R:0}``` or is it always expected to be RGBA? If it's always RGBA, then this PR seems valid.

SimonDarksideJ (approved):

SimonDarksideJ (changes requested):
Can we also have some Unit Tests of possible testing both combinations with various content types?

JakeLegendXIII:
Hey @SimonDarksideJ makes sense to me. There were no tests on this file today so I ended up creating a new file. There are positive and negative test cases for each. I made sure that Cornflower Blue and MonoGame orange were two of the test colors :)

mrhelmut:
Merging, thanks for the effort!

SimonDarksideJ:
Fantastic work @JakeLegendXIII and thanks for the tests, which ensure functionally continues to work as changes are made.

PR #8175, opened 2024-02-13
[Documentation] Add missing XML documentation for Artist class

Maniekko:
## Notes
- MonoGame does not have an ArtistCollection class, so part of the documentation that mentioned it has been omitted:
`You can obtain an Artist object through the`<strike>`ArtistCollection.Item indexer and`</strike>`Album.Artist and Song.Artist properties`


Maniekko:
Oops, another commit got into this PR, will reopen it with proper commit

PR #8177, opened 2024-02-13
[Documentation]: Adds missing XML Documentation for Audio namespace

Maniekko:
This PR adds the documentation for the following classes in `Microsoft.Xna.Framework.Audio` namespace:
- `AudioEngine`
- `Cue`
- `SoundBank`
- `WaveBank`
- `MicrophoneState` (enum, located in `Microphone.cs`)

### DIFF
--- a/MonoGame.Framework/Audio/Xact/AudioEngine.cs
+++ b/MonoGame.Framework/Audio/Xact/AudioEngine.cs
@@ -381,6 +381,10 @@ namespace Microsoft.Xna.Framework.Audio
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        /// Allows this object to attempt to free resources and perform other
+        /// cleanup operations before garbage collection reclaims the object.
+        /// </summary>
         ~AudioEngine()
         {
             Dispose(false);
--- a/MonoGame.Framework/Audio/Xact/Cue.cs
+++ b/MonoGame.Framework/Audio/Xact/Cue.cs
@@ -64,6 +64,7 @@ namespace Microsoft.Xna.Framework.Audio
             }
         }
 
+        /// <summary>Returns whether the cue is stopping playback.</summary>
         public bool IsStopping
         {
             get
@@ -73,13 +74,16 @@ namespace Microsoft.Xna.Framework.Audio
             }
         }
 
+        /// <summary>Returns whether the cue is preparing to play.</summary>
         public bool IsPreparing 
         {
             get { return false; }
         }
 
+        /// <summary>Returns whether the cue is prepared to play.</summary>
         public bool IsPrepared { get; internal set; }
 
+        /// <summary>Returns whether the cue has been created.</summary>
         public bool IsCreated { get; internal set; }
 
         /// <summary>Gets the friendly name of the cue.</summary>
--- a/MonoGame.Framework/Audio/Microphone.cs
+++ b/MonoGame.Framework/Audio/Microphone.cs
@@ -13,7 +13,13 @@ namespace Microsoft.Xna.Framework.Audio
     /// </summary>
     public enum MicrophoneState
     {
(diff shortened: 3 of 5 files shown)


dellis1972 (approved):

MonoGame.Framework/Audio/Xact/AudioEngine.cs
```diff
@@ -381,6 +381,10 @@ public void Dispose()
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        /// Allows this object to attempt to free resources and perform other
+        /// cleanup operations before garbage collection reclaims the object.
+        /// </summary>
         ~AudioEngine()
```
tomspilman, line 388: @dellis1972 @SimonDarksideJ should we be documenting private/protected methods that the developer using the public APIs won't ever call or use?

SimonDarksideJ: It certainly doesn't hurt.  But as a general rule no.
We should approve this and then improve our guidance for future commits

SimonDarksideJ (approved):

MonoGame.Framework/Audio/Xact/Cue.cs
```diff
@@ -73,13 +74,16 @@ public bool IsStopping
             }
         }
 
+        /// <summary>Returns whether the cue is preparing to play.</summary>
         public bool IsPreparing 
         {
             get { return false; }
         }
 
+        /// <summary>Returns whether the cue is prepared to play.</summary>
```
tomspilman, line 83: With members like this one it would be good to document the WHEN a cue goes into the preparing state.

SimonDarksideJ: This would certainly be a good addition, although it should be the audio documentation which goes into a deeper dive regarding the lifecycle of the api.

SimonDarksideJ (changes requested):
Please provide additional detail behind the method docs, we should not just be effectively repeating what the name of a method is. the docs should assist the developer in understanding the use of a method.

Maniekko:
I reread the original XNA documentation and found some remarks for the IsPrepared member that I previously missed.
Should I also add remarks for the IsStopping and IsPreparing members, since they always return false?

dellis1972:
@Maniekko if you can do IsStopping and IsPrepareing as well that would be great. Basically if you see something that could benifit from more detail then add that in.

MonoGame.Framework/Audio/Xact/Cue.cs
```diff
@@ -81,6 +81,11 @@ public bool IsPreparing
         }
 
         /// <summary>Returns whether the cue is prepared to play.</summary>
+        /// <remarks>
+        /// This property returns <see langword="true"/> only if the cue is prepared but has not yet been played.
+        /// Cues that are played return <see langword="true"/> for <see cref="IsPlaying"/> if they are currently playing
+        /// or <see langword="true"/> for <see cref="IsStopped"/> if they are stopped.
+        /// </remarks>
```
tomspilman, line 88: Thanks for the extra effort here @Maniekko .

First i would avoid writing docs about other members in this one... like we don't need to document what IsPlaying and IsStopped does unless it directly relates to the behavior of this one.

I would instead say something like:

```
This returns <see langword="true"/> after the cue has been created, has finished preparing, and is ready to be played for the first time.
```

SimonDarksideJ:
Any further updates @Maniekko ?

Maniekko:
> Any further updates @Maniekko ?

I'm currently focused on filling the missing docs with Aristurtle before updating already filled docs. 

PR is marked as "draft" so it doesn't get merged before documentation guidelines are released

Maniekko:
Docs are now following current contribution guidelines for documentation

PR #8178, opened 2024-02-13
Fix to avoid allocation/garbage when submitting OpenAL buffers.

ParanoidCactus:
This is a minor change to prevent a new int array from being allocated every time DynamicSoundEffectInstance.SubmitBuffer() is called.

I found this by running a performance profile on a project I'm working on that uses DynamicSoundEffectInstance for the audio. It was allocating over 2000 arrays in 15 seconds.

For reference the previous version was calling the below method and always passing in 1 as the count. So this is just doing the exact same thing without allocating a new array on the heap.

```
internal unsafe static void GenBuffers(int count, out int[] buffers)
{
    buffers = new int[count];
    fixed (int* ptr = &buffers[0])
    {
        alGenBuffers(count, ptr);
    }
}
```

### DIFF
--- a/MonoGame.Framework/Platform/Audio/OALSoundBuffer.cs
+++ b/MonoGame.Framework/Platform/Audio/OALSoundBuffer.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Xna.Framework.Audio
 
 		public OALSoundBuffer()
 		{
-            AL.GenBuffers(1, out openALDataBuffer);
+            AL.GenBuffer(out openALDataBuffer);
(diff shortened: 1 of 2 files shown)


dellis1972 (approved):
This looks like a good fix :) Not allocating stuff is a good thing :D
I'll let @harry-cpp or @mrhelmut comment as well.

tomspilman:
Looks great to me!  Good job @ParanoidCactus

mrhelmut:
It's a welcome tweak, merging!

Issue #8165, opened 2024-02-06
Feature Request: Resolve Missing XML For Public Type Warnings


AristurtleDev:
## Description
This is a feature request to have the source code updated so that all publicly visible types contain XML code documentation. This serves two purposes 

1. Ensures that intellisense is provided for all public types when using MonoGame
2. Ensure that all public types have API Reference documentation generated for the website.

Currently when building MonoGame from source, there are approximately 861 warnings issues due to missing XML documentation in code.  Here is a spreadsheet for reference from a current build of the `develop` branch 

https://docs.google.com/spreadsheets/d/1CMKClYiAuwbw1M7vMwqQFRj-Loq6a7VmTpqbb1HiNTE/edit?usp=sharing

Based on discussion from #7693, Microsoft has given permission and provided access to the original XNA API reference that can be used to fill in the missing XML code documentation in the source <a href="#reference-1"><sup>[1]</sup></a> as long as the following copyright text is included.

```
© 2012 Microsoft Corporation. All rights reserved.
```

## Implementation
My suggested implementation of this is to take the XNA API reference documentation provided by Microsoft and use that to complete the missing XML code documentation in the MonoGame source code.  However, in doing this, a few questions are posed:

1. Given that Microsoft provided this documentation to the MonoGame foundation, it may not be available to distribute publicly to community members to contribute to this feature. This has been discussed already in the MonoGame discord and I believe the current recommendation is to provide a private repo and have contributing members sign an NDA <a href="#reference-2"><sup>[2]</sup></a>
2. Where should the copyright text be placed?  @tomspilman has suggested placing it in the footer of the website along with the MonoGame copyright<a href="reference-3"><sup>[3]</sup></a>. This would satisfy the requirement for the website presence but do we need to also include the copyright information in the source code as well?  For instance

```xml
/// <summary>
/// Example documentation that might have been provided by microsoft
/// <summary>
/// © 2012 Microsoft Corporation. All rights reserved.
```

These are the two questions I can think of at the moment, but I'm sure there are more.  Regardless, to ensure we remain in legal compliance with the supplied documentation provided by Microsoft, guidelines and a process will need to be established.

## Motivation
There are times when the question is asked in the community discord and forums about where to find API reference documentation that is missing <a href="#reference-4"><sup>[4]</sup></a>. Currently we have to point these questions to the MSDN site where the original XNA API reference exists.  However, there is no guarantee that Microsoft will keep these pages available. 

References:
[...]

PR #8182, opened 2024-02-14
[Documentation] Add missing XML documentation for Color

AristurtleDev:
## Description
This PR adds the missing XML documentation for the Color struct.

## Reference

## Note
The method missing the XML documentation was the `operator *` overload which is the same as the already documented operator overload, but the the params reversed to allow the `float` on the left hand side.  Used the existing documentation.


MonoGame.Framework/Color.cs
```diff
@@ -1761,6 +1761,12 @@ public static Color Multiply(Color value, float scale)
             return new Color((int)(value.R * scale), (int)(value.G * scale), (int)(value.B * scale), (int)(value.A * scale));
         }
 
+	    /// <summary>
+        /// Multiply <see cref="Color"/> by value.
+        /// </summary>
+        /// <param name="scale">Multiplicator.</param>
```
tomspilman, line 1767: I don't think the periods here are necessary unless you are writing a full sentence. 

That said we probably want to define some guidance here to keep consistency for this @SimonDarksideJ @dellis1972 ? 

Maybe we should have some "examples" of different types of docs in the documentation contributing guide?  Maybe like the coding standard showing right and wrong way to write various types of API docs?

AristurtleDev: @tomspilman sorry I updated the PR before I saw this comment.

I took the existing API XML from the multiply operator above and just reused it since it's the same thing, just swapped params.

The commit I just submitted has, I feel, a better Api documentation of the methods.  But that's personal bias 😄.  If you could look at that.

But yes having a specific guideline would be great.  The one in the coding style doc is really basic, and myself and @Maniekko are using the XNA docs @dellis1972 provided to fill in the missing holes where we can

AristurtleDev:
Switching to draft PR so reviews and feedback can be made, but waiting for guidelines to be updated for documentation before checking for merge

SimonDarksideJ (approved):

dellis1972:
@AristurtleDev thanks for the docs 👍

PR #8183, opened 2024-02-14
[Documentation]: Add XML documentation to Media namespace

Maniekko:
This PR **is quite big** and adds the documentation for `Microsoft.Xna.Framework.Media` namespace. 
**A review is probably needed**

### Media
- `Genre`
- `MediaLibrary`
- `MediaPlayer`
- `MediaQueue`
- `MediaSource`
- `MediaSourceType` (enum)
- `MediaState` (enum)

### Playlist
- `Playlist`
- `PlaylistCollection`

### Song
- `Song`
- `SongCollection`
- `Platform/Media/Song.NVorbis`

### Video
- `Video`
- `VideoPlayer`

## Notes
- `MediaSource`, `MediaSourceType` docs were heavily modified because the original XNA docs did not match with actual MG implementation (Thanks to **Aristurtle** for help)
- `SongCollection` and `PlaylistCollection` didn't have the docs for most members. Original C# docs were reused, where possible (for `ICollection.Clear()`, `ICloneable.Clone()`, `operator==` etc.)
- GitHub diff shows some lines as deleted. Not sure why, probably different line endings.

## Reference:

### DIFF
--- a/MonoGame.Framework/Media/MediaPlayer.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Xna.Framework.Media
 {
     /// <summary>
     /// Provides methods and properties to play, pause, resume, and stop songs.
-    /// <see cref="MediaPlayer"/> also exposes shuffle, repeat, volume, play position, and visualization capabilities.
+    /// MediaPlayer also exposes shuffle, repeat, volume, play position, and visualization capabilities.
     /// </summary>
     public static partial class MediaPlayer
     {
@@ -274,7 +274,7 @@ namespace Microsoft.Xna.Framework.Media
 		}
 
         /// <summary>
-        /// Stops currently playing song, moves to the previous song in the queue of playing songs and plays it.
+        /// Stops currently playing song, moves to the next song in the queue of playing songs and plays it.
         /// </summary>
         /// <remarks>
         /// <para>
--- a/MonoGame.Framework/Media/MediaPlayer.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.cs
@@ -122,8 +122,6 @@ namespace Microsoft.Xna.Framework.Media
         /// calls to <see cref="Play(Song)"/>, <see cref="Stop()"/>, <see cref="Pause()"/>,
         /// <see cref="Resume()"/>, <see cref="MoveNext()"/>, and <see cref="MovePrevious"/>
         /// might have no effect, depending on the platform.
-        /// If another application's background music is playing, your game will need to call
-        /// in order to pause the other application's background music in order to play the game's music.
         /// </remarks>
         public static bool GameHasControl
         {
--- a/MonoGame.Framework/Media/MediaPlayer.cs
+++ b/MonoGame.Framework/Media/MediaPlayer.cs
@@ -23,7 +23,8 @@ namespace Microsoft.Xna.Framework.Media
 		private static readonly MediaQueue _queue = new MediaQueue();
 
         /// <summary>
-        /// Raised when the active song changes due to active playback or due to explicit calls to the <see cref="MoveNext()"/> or <see cref="MovePrevious()"/> methods.
+        /// Raised when the active song changes due to active playback or due to explicit calls to the
+        /// <see cref="MoveNext()"/> or <see cref="MovePrevious()"/> methods.
         /// </summary>
 		public static event EventHandler<EventArgs> ActiveSongChanged;
         /// <summary>
@@ -56,7 +57,8 @@ namespace Microsoft.Xna.Framework.Media
         /// Gets or sets the repeat setting for the media player.
         /// </summary>
         /// <remarks>
-        /// When set to <see langword="true"/>, the playback queue will begin playing again after all songs in the queue have been played.
+        /// When set to <see langword="true"/>, the playback queue will begin
+        /// playing again after all songs in the queue have been played.
         /// </remarks>
         public static bool IsRepeating 
         {
@@ -68,7 +70,8 @@ namespace Microsoft.Xna.Framework.Media
         /// Gets or sets the shuffle setting for the media player.
         /// </summary>
         /// <remarks>
-        /// When set to <see langword="true"/>, songs in the playback queue are played in random order rather than from first to last.
+        /// When set to <see langword="true"/>, songs in the playback queue are
+        /// played in random order rather than from first to last.
         /// </remarks>
         public static bool IsShuffled
         {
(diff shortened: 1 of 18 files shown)


Maniekko:
For some reason VS decided to add some weird `using`s in files i modified. Will fix this when i get to home

MonoGame.Framework/Media/MediaPlayer.cs
```diff
@@ -69,17 +109,40 @@ private set
...
             }
-        }
-
+        }
+
+        /// <summary>
+        /// Determines whether the game has control of the background music.
+        /// </summary>
+        /// <remarks>
+        /// A gamer can play their own music as the background to your game by using the Xbox 360 dashboard.
```
dellis1972, line 118: We will probably want to keep an eye out for references to things like Xbox 360/Zune or Windows Phone. Most of those specific comments will no longer be needed. We've had to do this with other documentation we are importing for the tutorials side of things.

Maniekko:
Docs **that were written by me** are now following current contribution guidelines for documentation.
Existing doclines were kept untouched.

MonoGame.Framework/Media/MediaPlayer.cs
```diff
@@ -69,17 +112,41 @@ private set
...
+
+        /// <summary>
+        /// Determines whether the game has control of the background music.
+        /// </summary>
+        /// <remarks>
+        /// If a gamer is currently playing their own music as the background to your game,
+        /// calls to <see cref="Play(Song)"/>, <see cref="Stop()"/>, <see cref="Pause()"/>,
+        /// <see cref="Resume()"/>, <see cref="MoveNext()"/>, and <see cref="MovePrevious"/>
+        /// might have no effect, depending on the platform.
+        /// If another application's background music is playing, your game will need to call
```
dellis1972, line 125: I think this sentence is missing a word

`If another application's background music is playing, your game will need to call <???> in order to pause the other application's background music in order to play the game's music.`

Maniekko: It looks like it, but that's exactly how this sentence is written in XNA docs. Open XNA docs [doesn't mention this](https://learn.microsoft.com/en-us/previous-versions/windows/xna/dd254741(v=xnagamestudio.30)#remarks) at all. 
I'll probably remove this part for now

Maniekko:
I tried to resolve a merge conflict and broke the PR :/

MonoGame.Framework/Media/MediaPlayer.cs
dellis1972: I guess is was `Pause` based on the context. But if we don't know how this works on each platform its probably best not to mention it. 👍

dellis1972:
@SimonDarksideJ  can you review please.

SimonDarksideJ (changes requested):
So two overriding concerns in the latest update are as follows:

1. Don't use self-references unless it is needed to bring understanding for a class. 
E.g. The Genre class describes itself as "The Genre class provides information about a genre"
Which is fine if you know what a genre is. plus the first "Genre" creates a link to itself with the cref.
I would probably word it more like "The Genre class is used to categorise audio content from the device under a common `<see cref="Name"/>`, including the the `<see cref="Albums"/>` and `<see cref="Songs"/>` in that genre that are on the device."

    Try to keep the information informative and drive what a "thing" is without repeating it is just a thing.

2. when you use a cref, it only gives you the sub Name for the thing you are referencing, e.g. `<see cref="Album.Genre"/>` only displays as "Genre" in the generated docs, which looks weird.  See image below of the generated page for Genre
image

MonoGame.Framework/Media/SongCollection.cs
```diff
@@ -100,21 +138,54 @@ public SongCollection Clone()
...
         {
             return innerlist.IndexOf(item);
         }
 
+        /// <summary>
+        /// Removes the first occurrence of a <see cref="Song"/> from the <see cref="SongCollection"/>.
+        /// </summary>
+        /// <param name="item">The object to remove from the <see cref="SongCollection"/>.</param>
+        /// <returns>
+        /// <see langword="true"/> if item was successfully removed from the <see cref="SongCollection"/>;
```
SimonDarksideJ, line 184: I don't think we need to include SEE references for value types like "true", simply stating "Returns True if the item was successfully removed...

MonoGame.Framework/Media/SongCollection.cs
```diff
@@ -8,27 +8,49 @@
...
+        /// </summary>
         public static readonly SongCollection Empty = new SongCollection();
-
 		private bool isReadOnly = false;
 		private List<Song> innerlist = new List<Song>();
 
         internal SongCollection()
         {
-
+
```
SimonDarksideJ, line 37: Added whitespace?

SimonDarksideJ (changes requested):

MonoGame.Framework/Media/SongCollection.cs
Maniekko: "See" references with "langword" make it look good in Intellisense, and theoretically shouldn't change DocFx docs

MonoGame.Framework/Media/SongCollection.cs
Maniekko: No, at least not intentionally

MonoGame.Framework/Media/SongCollection.cs
SimonDarksideJ: Discussed this with @AristurtleDev , and while fine, it just wasn't something I've used, so I reversed my position on it :D

Maniekko:
> when you use a cref, it only gives you the sub Name for the thing you are referencing
It looks like Intellisense does show a class name, 
image
and explictly setting the class would look weird:
image

SimonDarksideJ:
> 2\. <see cref="Album.Genre"/>

@AristurtleDev suggestion seems the best, which will make its way into the guidance (when I get another minute)

Make sure to mention the display name within the "see" tag, for Example

`<see cref="Album.Genere">Album.Genre</see>`

MonoGame.Framework/Platform/Media/Song.NVorbis.cs
```diff
@@ -98,6 +98,9 @@ internal float Volume
             }
         }
 
+        /// <summary>
+        /// Gets the duration of the <see cref="Song"/>.
```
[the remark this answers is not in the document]
Maniekko: It's the only `public` member in this file. XML docline was added to remove a warning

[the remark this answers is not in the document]
SimonDarksideJ: > It's the only `public` member in this file. XML docline was added to remove a warning

Very good point!

dellis1972:
Looking good.

Maniekko:
- All `Dispose()` methods are now inheriting docs from `iDisposable.Dispose()`
- Most self-references in `cref` were removed, a normal text representation of the class is used instead.
- All references with possible name collisions are now using `<see>ClassName</see>` format
- Minor indentation fixes

MonoGame.Framework/Media/VideoPlayer.cs
```diff
@@ -12,6 +12,10 @@
 
 namespace Microsoft.Xna.Framework.Media
 {
+    /// <summary>
+    /// Provides methods and properties to play back, pause, resume, and stop video.
+    /// <see cref="VideoPlayer"/> also exposes repeat, volume, and play position information.
+    /// </summary>
     public sealed partial class VideoPlayer : IDisposable
```
SimonDarksideJ, line 19: Missing Cref links to `Video`

MonoGame.Framework/Media/Song.cs
```diff
@@ -6,15 +6,18 @@
 using System.IO;
 
 namespace Microsoft.Xna.Framework.Media
-{
+{
+    /// <summary>
+    /// Provides access to a song in the song library.
```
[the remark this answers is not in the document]
Maniekko: The docline here is from original XNA documentation, not sure if i should update it.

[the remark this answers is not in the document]
dellis1972: It probably used to work that way on Windows Phone .

I think its just song's from the SongCollection or the specified song now, we'd need to check the implementation.

SimonDarksideJ (approved):

dellis1972:
@Maniekko thanks for the contribution 👍

PR #8192, opened 2024-02-21
Update requirements.md platforms link and contributing.md tests link

JakeLegendXIII:
Two small link updates in md files.

- Updating the REQUIREMENTS.MD to correctly point to Platforms documentation on the new site (I previously updated this after the websites 11ty migration, but before the latest route changes so it is currently returning a 404, this fixes that).
- Update to CONTRIBUTING.MD to make the unit tests reference go correctly to the Tests folder instead of a 404

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -1,4 +1,4 @@
-Depending on the [platform](https://monogame.net/articles/platforms/) that you are targeting, MonoGame has different sets of requirements.
+Depending on the [platform](https://monogame.net/articles/platforms.html) that you are targeting, MonoGame has different sets of requirements.
 
 For desktop platforms
 ====================
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -1,4 +1,4 @@
-Depending on the [platform](https://monogame.net/articles/platforms.html) that you are targeting, MonoGame has different sets of requirements.
+Depending on the [platform](https://monogame.net/articles/platforms) that you are targeting, MonoGame has different sets of requirements.
 
 For desktop platforms
 ====================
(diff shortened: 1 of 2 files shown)


REQUIREMENTS.md
```diff
@@ -1,4 +1,4 @@
-Depending on the [platform](https://monogame.net/articles/platforms/) that you are targeting, MonoGame has different sets of requirements.
+Depending on the [platform](https://monogame.net/articles/platforms.html) that you are targeting, MonoGame has different sets of requirements.
```
dellis1972, line 1: Thanks for the contribution. I think this was fixed with a recent website update the non html url is now working.

JakeLegendXIII: Hey @dellis1972 when clicking the link without the .html and the / as it currently is you're taken to a 404. Removing the .html does work, but not with the current / at the end. I can remove the .html as well as the / and it will fix the current issue and correctly route with the recent changes. 

But it is currently still a broken link without this change.

dellis1972: @AristurtleDev any thoughts on this one?

dellis1972: I would be in favour of just removing the `/` unless other links use `.html`. We should be consistent at least.

AristurtleDev: @dellis1972 the initial port of the website to 11ty did in fact use the 'prettier" urls that end with `/`.  However, after implementing it this way, it was discovered that this broke all search engine results and other existing articles/documentation that point to the website, so it was reverted back during the maintainability update in PR https://github.com/MonoGame/monogame.github.io/pull/91

So @JakeLegendXIII has it correct here that it should end with `.html` and not `/`

dellis1972 (approved):

PR #8193, opened 2024-02-21
[Native] Add generator for c# types used on c++ side

harry-cpp:
To avoid us having to keep track of C# enums on the C++ side, here is a small generator that exports them from the c# side.

In the future, if there is a need, we could also make it export C# structs.


src/monogame/include/csharp_enums.h
```diff
@@ -0,0 +1,2104 @@
+//
```
tomspilman, line 1: So is `src` the top level folder decided for the native code?

What will be in this `src` folder?  Is it just the MonoGame native C++ dll?

src/monogame/include/csharp_enums.h
```diff
@@ -0,0 +1,2104 @@
...
+};
+
+enum CSContainmentType : csint
+{
+    Disjoint = 0,
+    Contains = 1,
+    Intersects = 2,
+};
+
+class ECSContainmentType
```
tomspilman, line 83: Why is this in here?

harry-cpp: Easy way to get string value for the enum as C++ has no built in way.

src/monogame/include/csharp_enums.h
harry-cpp: `src/monogame` is the top level folder for the native code.

We should eventually move our src related stuff into one folder and not keeping it all mixed up in the top level, but that can come later once the native platform is fully working.

src/monogame/include/csharp_enums.h
tomspilman: No i mean why is `ECSContainmentType` here... that isn't planned to be exposed to native code.

But i see why... your generator just dumps all enums from MG framework to this header.

Issue #8124, opened 2024-01-19
macOS and Linux setup experience


ThomasFOG:
I'm creating this issue to centralize all of the other issues to setup a MonoGame project on macOS or Linux.

It is a set of different issues degrading the setup experience on those systems. They are aggregated with the [setup label](https://github.com/MonoGame/MonoGame/labels/Setup), most of them being duplicates.

How to fix those:
- [ ] Build all native dependencies with ```osx-arm64``` support (this is currently being done by refactoring the dependencies into separate projects so that MonoGame can built it own native dependencies without relying on third-parties)
  - [x] SDL
  - [x] FreeType (needs to be wired and tested with SharpFont)
  - [x] FreeImage
  - [x] ffmpeg/ffprobe
  - [x] assimp
  - [x] Replace ATI.TextureCompression, PVRTexLib and NVTT with "basis universal"
  - [x] Hook all packages with the pipeline
- [ ] Fix ```mgfxc_wine_setup.sh```
   - [x] Change the .NET SDK link to .NET 8.0
   - [x] Update WineHelper.cs comments to .NET 8
   - [x] Check for Wine version requirement (there is a minimum version required)
   - [x] Enhance the script to support both 32 and 64 bit versions of Wine (some users have both installed and stuff got installed on the 32 bit version instead)

AristurtleDev:
@mrhelmut 
- Change the .NET SDK link to .NET 8.0: Completed in #8218 
- Update WineHelper.cs comments to .NET 8: Completed in #8219 
- Check for Wine version requirement (there is a minimum version required): Completed in #8220

AristurtleDev:
>  Enhance the script to support both 32 and 64 bit versions of Wine (some users have both installed and stuff got installed on the 32 bit version instead)

Ok so the problem seems to be that monogame assumes that the Linux user is using apt and installed wine64.

But if you install wine like normal using like winehq-staging, then you get both 32-bit and 64-bit support.

Setting the 64-bit support is just the WINE_ARCH flag.

The script would need to be changed to **not** check for the assumption of wine64 and instead just check for just plain wine, then carry on from there. But there needs to be a way to ensure that 64-bit support was also installed

Thoughts @mrhelmut @harry-cpp

AristurtleDev:
`dpkg -l | grep wine` will output

```
aristurtle@ubuntu-personal:~/ftp$ dpkg -l | grep wine
[...]

Issue #8103, opened 2023-12-06
.NET 8 Update Issues On Linux


AristurtleDev:
### Prerequisites

- [X] I have verified this issue is present in the `develop` branch
- [X] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

.NET 8 Update

### Which MonoGame platform are you using?

MonoGame Cross-Platform Desktop Application (mgdesktopgl)

### Operating System

Linux (Ubuntu 22.04.3 LTS)

### Description

Testing the current `develop` branch with .NET8 in LInux, the following issues were found and need to be corrected
1. If attempting to build FX file before setting up wine, the error message states to install `.NET6` and not `.NET8`
```
Setup instructions:
- Create 64 bit wine prefix
- Install d3dcompiler_47 using winetricks
- Install .NET 6
- Setup MGFXC_WINE_PATH environmental variable to point to a valid wine prefix
```

2.  The `DOTNET_URL` in the **mgfxc_wine_setup.sh** script incorrectly points to .NET6 SDK and not the .NET8SDK

https://github.com/MonoGame/MonoGame/blob/84a2690d4c1278365478893c9852f09103d3baa1/Tools/MonoGame.Effect.Compiler/mgfxc_wine_setup.sh#L39

It should point to `https://download.visualstudio.microsoft.com/download/pr/2b2d6133-c4f9-46dd-9ab6-86443a7f5783/340054e2ac7de2bff9eea73ec9d4995a/dotnet-sdk-8.0.100-win-x64.zip`

3. Wine needs to be on a newer version; otherwise the following error is thrown.  Currently tested successfully with Wine version 8
```
wine: Call from 0x7bc7c9e1 to unimplemented function KERNEL32.dll.GetProcessGroupAffinity, aborting
wine: Unhandled page fault on read access to 000000000000000C at address 00000000005C9834 (thread 002e), starting debugger...
```

**this is something that will need to be noted in the documentation, not something we can do in the repo**

### Steps to Reproduce

1. Setup .NET 8 SDK in Linux
2. Setup wine according to documentation at https://monogame.net/articles/getting_started/1_setting_up_your_development_environment_unix.html#optional-setup-wine-for-effect-compilation
3. Clone `develop` branch
4. Run `cake build` script
[...]

PR #8220, opened 2024-03-05
Validate Minimum Wine Version 8.0 for .NET 8

AristurtleDev:
## Description
This updates the **mgfxc_wine_setup.sh** script to validate that at minimum Wine version 8.0 is installed.

## References
- [Issue #8103: .NET 8 Update Issues On Linux ](https://github.com/MonoGame/MonoGame/issues/8103)
- [Issue #8124: macOS and Linux setup experience](https://github.com/MonoGame/MonoGame/issues/8124)

### DIFF
--- a/Tools/MonoGame.Effect.Compiler/mgfxc_wine_setup.sh
+++ b/Tools/MonoGame.Effect.Compiler/mgfxc_wine_setup.sh
@@ -15,6 +15,16 @@ then
     exit 1
 fi
 
+# wine 8 minimum needed for dotnet8
+# see https://github.com/MonoGame/MonoGame/issues/8103
+# wine --version will output 'wine-#.# (Ubuntu #.#.#)'
+# grep to find only the 'wine-#' then use sed to remove the 'wine-' prefix
+WINE_VERSION=$(wine --version 2>&1 | grep -oP 'wine-\d+' | sed 's/wine-//')
+if (( $WINE_VERSION < 8 )); then
+    echo "Wine version $WINE_VERSION is below the minimum required version (8.0)."
+    exit 1
+fi
+
 # init wine stuff
 export WINEARCH=win64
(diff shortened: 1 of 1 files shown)


Tools/MonoGame.Effect.Compiler/mgfxc_wine_setup.sh
```diff
@@ -15,6 +15,16 @@ then
     exit 1
 fi
 
+# wine 8 minimum needed for dotnet8
+# see https://github.com/MonoGame/MonoGame/issues/8103
+# wine --version will output 'wine-#.# (Ubuntu #.#.#)'
+# grep to find only the 'wine-#' then use sed to remove the 'wine-' prefix
+WINE_VERSION=$(wine --version 2>&1 | grep -oP 'wine-\d+' | sed 's/wine-//')
```
harry-cpp, line 22: Why is grep needed?

AristurtleDev: `wine --version` will output something like

```
wine-#.# (Ubuntu #.#.#)
```

`grep` is used to pattern match the `wine-#` string specifically from thr output, then `sed` to remove the `wine-` prefix which just leaves the major version number of wine detected.

Since it's an integer value and not the #.# floating point value, that makes the version check a simple if statement.

harry-cpp: `sed` does support regex, but I guess this is fine as well. 

Modify comment to:
```
# wine 8 is the minimum requirement for dotnet 8
# wine --version will output "wine-#.# (Distro #.#.#)" or "wine-#.#"
```
The rest is not needed.

AristurtleDev: @harry-cpp comment modified.

PR #8240, opened 2024-03-14
Performance: Using newer Environment.CurrentManagedThreadId for UI/Main thread check

MegaChuck64:
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1840

According to Microsoft, System.Environment.CurrentManagedThreadId is a 'compact and efficient replacement' to Thread.CurrentThread.ManagedThreadId.

 I've made this change in Platform/Threading.cs where the UI thread is checked.

### DIFF
--- a/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs
+++ b/MonoGame.Framework/Platform/Android/MonoGameAndroidGameView.cs
@@ -315,7 +315,7 @@ namespace Microsoft.Xna.Framework
 
         protected void WorkerThreadFrameDispatcher(SynchronizationContext uiThreadSyncContext)
         {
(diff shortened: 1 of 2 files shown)


Mindfulplays:
Neat! Also, `MonoGameAndroidGameView` has a couple of references...

MegaChuck64:
> Neat! Also, `MonoGameAndroidGameView` has a couple of references...

thanks, updated those!

dellis1972:
@MegaChuck64 have you been able to test this on Android and DesktopGL? 
Also although it looks like this property is available for .NET Framework 4.5 can you confirm it works there? We need to be mindful of changes which might impact the console side of things.

mrhelmut:
I don't see added value in this PR beside modernisation. To me, if it doesn't fix an existing issue, we should leave it be.

Issue #6661, opened 2019-02-13
ContentWriter bugged...


sharkist:
whenever I get the "ContentWriter.BaseStream" pointer, then the ContentWriter adding again and again the header in the file. Why ContentWrite write something in the file, i only get basestream.position?

If i do something like below:
```

        protected override void Write(ContentWriter cs_writer, TilesetContentData tsdata)
        {
            long a = cs_writer.BaseStream.Position;
            long b = cs_writer.BaseStream.Position;
            long c = cs_writer.BaseStream.Position;
            long d = cs_writer.BaseStream.Position;
            long e = cs_writer.BaseStream.Position;
            long f = cs_writer.BaseStream.Position;
            long g = cs_writer.BaseStream.Position;
        };

```
My file will have "8" XNB header files.

my workaround if i need 'basestream.position' just make first MemoryStream, and later everything copy by Write(memstream.ToArray());

[http://oi65.tinypic.com/2gx1eer.jpg](http://oi65.tinypic.com/2gx1eer.jpg)

#### What version of MonoGame does the bug occur on:
- MonoGame 3.7

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

Jjagg:
`ContentWriter` writes stuff when its `Flush` method is called. This is not correct behavior. Flush should not do the actual writing.

h3allen:
Hi! I'm a first time contributor and I'd like to take a stab at fixing this if it's still a problem!

mrhelmut:
I don't think this issue is still relevant. This behavior is inherited from ```BinaryWriter``` and doesn't seem to be a MonoGame issue. This should be tested against .NET 8 to be sure.

Jjagg:
That doesn't sound right @mrhelmut. The `Flush` implementation should complete buffered write operations, but `ContentWriter` actually writes new data to the underlying stream.
You can see from the doc comment that the implementation has a wrong perception of what `Flush` should do:

https://github.com/MonoGame/MonoGame/blob/72919dc8185e70734507b58bd9a8e03bfe9a0d96/MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ContentWriter.cs#L123

[...]

PR #8266, opened 2024-04-02
Issue #6661: Moved content writing from Flush to new method FinalizeContent

h3allen:
Solution for [Issue #6661 ](https://github.com/MonoGame/MonoGame/issues/6661)
Moves the previous code writing content in Flush() to new method FinalizeContent() and calls it in ContentCompiler method Compile() before calling Flush().


mrhelmut:
Note for maintainers: tests must be run against this PR.

MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ContentWriter.cs
```diff
@@ -172,7 +172,14 @@ public override void Flush()
...
                         compressedStream.Dispose();
                 }
-            }
+            }
+        }
+
+        /// <summary>
+        /// Clears buffer using BinaryWriter's flush method
+        /// </summary>
+        public override void Flush()
```
Mindfulplays, line 181: Do we need this anymore? If it's defaulting to the `virtual base` call anyways.

AristurtleDev: If all it's doing is calling the base, then it's not needed

h3allen: I can certainly go and delete it if it's only calling base and has no other purpose

h3allen: Deleted! Apologies for the delay.

Mindfulplays: I am not a maintainer so you would need to ping @harry-cpp / @mrhelmut to review the PR(s).

MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ContentWriter.cs
```diff
@@ -118,14 +118,14 @@ protected override void Dispose(bool disposing)
...
+        }
+
         /// <summary>
         /// All content has been written, so now finalize the header, footer and anything else that needs finalizing.
-        /// </summary>
-        public override void Flush()
-        {
-            // Write shared resources to the end of body stream
+        /// </summary>
+        public void FinalizeContent()
```
dellis1972, line 126: Does this need to be a public API? can it be `internal` ?

MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ContentWriter.cs
```diff
@@ -175,14 +175,6 @@ public void FinalizeContent()
             }
         }
 
-        /// <summary>
-        /// Clears buffer using BinaryWriter's flush method
-        /// </summary>
-        public override void Flush()
```
dellis1972: Now that I think about it, removing this will cause an ABI break, people may well be calling this methid in custom content Pipeline extensions.

dellis1972: ignore me, the base class has the Flush method.

SimonDarksideJ (changes requested):

MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ContentWriter.cs
dellis1972: Probably needs to be internal actually.

h3allen:
Apologies for the delay. Yes, I'll get to updating it ASAP!

SimonDarksideJ (approved):

PR #8274, opened 2024-04-11
Add Reference Assembly Packages.

dellis1972:
When building Library projects or Content Pipeline it would be better if we referenced a reference
assembly version of the MonoGame.Framework.dll.

We used to have a Portable library which was generated before the upgrade to .net 6.

This commit adds two new Packages

MonoGame.Framework.ReferenceAssembly
MonoGame.Framework.Content.Pipeline.ReferenceAssembly

These are generated using the `ProduceReferenceAssembly` on the DesktopGL and Content.Pipeline projects.

The main reason for adding these packages is to avoid confusion. Having a new user use the DesktopGL package for a library project made them think they needed to separate project or config for the WindowsDX package. Having reference assemblies for both the framework and Content extensions makes it more obvious that these references will not be used.

Also enabled the Android build on Linux as that is now supported by dotnet.


build/BuildContext.cs
```diff
@@ -19,7 +19,7 @@ public BuildContext(ICakeContext context) : base(context)
     {
         var repositoryUrl = context.Argument("build-repository", DefaultRepositoryUrl);
         var buildConfiguration = context.Argument("build-configuration", "Release");
-        BuildOutput = context.Argument("build-output", "artifacts");
+        BuildOutput = context.Argument("build-output", "Artifacts");
```
harry-cpp, line 22: Let the artifacts folder be called with a starting small letter was intentional on my part, I wanted to slowly resort the repo into:
- artifacts
- build
- src
- tests

etc. instead of our current setup.

harry-cpp:
> We used to have a Portable library which was generated before the upgrade to .net 6.

We dropped it as DesktopGL was able to serve as a reference assembly, so didn't see a reason to keep a dedicated package for it.

Also can't you just use `ProduceReferenceAssembly` / `ProducyOnlyReferenceAssembly` to generate it instead of the jetbrains tool? (I haven't used these compiler flags before, just know that they exist).

build/BuildContext.cs
dellis1972: I had to change this because it was inconsistent across the repo. I'll have to update all the references to use lowercase if that is the intention, Not sure how this worked with mixed case on Linux in the first place.

harry-cpp: Oh yea, I forgot we use the artifacts folder for the build output as well.

I guess I'll try to fix it up tomorrow in a separate PR.

dellis1972:
oops, I commited some code which got the android build working on linux also doh

dellis1972:
Closing this.

PR #8275, opened 2024-04-11
Multi-platform 2D Start-Kit 

CartBlanche:
So for people new to MonoGame, this 2D Start-Kit is for them to hit the ground running with a 2D game. 
Ideally pulling in things we showed in isolation, in our samples.

To that end this Start-Kit currently implements:

- [x] Screen Management system (from samples)
- [x] Mult-lingual support, and dynamically changing them
- [x] Settings to be able to toggle Full Screen, Change Languages, and Which Particle Effect to use.
- [x] Basic scripting for player movement
- [x] About screen to show 1 way to launch the default browser (more testing required)
- [x] Basic ParticleManager with 3 simple particle effects (needs tweaking and enhancing)
- [x] Saving Setting State, so it can pick them up between runs
- [x] Counter now counts up
- [x] Player now dies if they fall a particular distance
- [x] Added support for breakable tiles, a la Mario
- [x] Short Tutorial, to show gameplay goals, from MainMenu
- [x] Basic tracking of previous high score
- [x] Mouse/Touch interaction with menu items.
- [x] Mouse/Touch interaction to move player around level.
- [x] Mouse/Touch input scales across platforms, based on BaseScreenSize
- [x] Added extra Idle frames for player character (definitely need s proper artist down the line)

Level Tile Types:
| Level Character | Rendered As | Tile |
|---|---|---|
| . | Blank | |
| P | Player | |
| X | Exit | Exit |
| ~ | Platform Block | Platform |
| : | Passable Block | Passable |
| # | Impassable Block | Impassable |
| ; | Breakable Block | Breakable |
| 1, 2 or 3 | Gem Tile of different values. 1 lowest. 3 highest | Gem 1 |
| A, B, C or D | Different Enemy Types | Enemy A Enemy B Enemy C Enemy D |

Game Levels

Bounties
- [ ] Level Loading Screen and completed management.
- [ ] Convert HighScore Tracking into proper leaderboard, with Player Name input 
- [ ] Level visualiser and Editor 
- [ ] Support more languages.
- [ ] More levels?
- [ ] Release on App Stores?
- [ ] implement Power ups 
- [ ] Get artist to create player idle, walk and  whip crack animations. 
- [ ] Add functionality to support walk and whip crack animations and enemy death from whip crack hits

I think the code base needs cleaning up before merging, but also may need refactoring and input from others to make sure it's using best practices.


CartBlanche:
> Looks good, however:
> 
> * No content project?
> * No Game1.cs ?
> 
> Unless you intend to provide a guide for adding these?
> 
> I suspect you missed out the "Core" project from the MG samples @CartBlanche

Seems I did. I'll double check what happened!

Templates/MonoGame.Templates.CSharp/content/MonoGame.StartKit/___SafeGameName___.Android/___SafeGameName___.Android.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net8.0-android</TargetFramework>
+    <SupportedOSPlatformVersion>23</SupportedOSPlatformVersion>
+    <OutputType>Exe</OutputType>
+    <ApplicationId>com.companyname.___SafeGameName___.Android</ApplicationId>
```
dellis1972, line 6: probably don't need the .Android at the end here (or other platforms).

SimonDarksideJ: Then how do you tell the difference between projects when you open them all in one solution.
Or even side by side?
Unless I'm not getting what you are getting at?

SimonDarksideJ: Oh I see, just in the namespaces. In that case I agree 😂

dellis1972: Generally you should avoid using `Android` in anthing that might end up in a `namespace` as well. Since it can cause conflicts with the root `Android` namespace.

dellis1972: But in this case its the ApplicaitionId which will end up as the apk/aab name and in the store. So on the Play store we don't need to have the name of the app has Android at the end. com.company.MyGame is better than com.company.MyGame.Android.  its more consistent with other platforms.

CartBlanche: @dellis1972 I need to double check, but think we need a way to separate out each platform.
I'll remove the platform from each app equivalent Id.

dellis1972: We used to use `Droid` for project names, rather than Android. It might have been fixed. 
But the ApplicationId's definately do not need to have platform specific bits in them.

dellis1972: @CartBlanche did you see the above comment about having `Android`at the end of the namespace?

CartBlanche: @dellis1972 I had seen your comment, but I seem to have lost the changes back then. Fixed in latest commit.

CartBlanche:
@SimonDarksideJ, @dellis1972 I've updated the description on this PR, to show what I'm ultimately aiming for.
If either of you wants an exe to play with, let me know. Not sure if it's worth implementing the things in the TODO list, or put these up as bounties.

SimonDarksideJ:
Saving/loading is a good one, especially in a way that works across platforms and serializing data.

Possibly leader board. 

But all also need dedicated tutorials for the how to section too.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/___SafeGameName___.contentproj
```diff
@@ -0,0 +1,320 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
dellis1972, line 1: This is an old XNA content project?

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Android/___SafeGameName___.Android.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <ApplicationId>com.companyname.___SafeGameName___</ApplicationId>
+    <ApplicationVersion>1</ApplicationVersion>
+    <ApplicationDisplayVersion>1.0</ApplicationDisplayVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <AndroidResource Include="Resources\Drawable\Splash.png" />
+    <AndroidResource Include="Resources\Values\Styles.xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <MonoGameContentReference Include="..\___SafeGameName___.Core\Content\___SafeGameName___.mgcb">
```
dellis1972, line 15: Should this be just `..\___SafeGameName___.Core\Content\Content.mgcb`?

The reason is it effects the value used in 
`Content.RootDirectory = "Content";`

Or at least it used to.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/Fonts/Hud.spritefont
```diff
@@ -0,0 +1,60 @@
...
+of the font in your game, and to change the characters which are available to draw
+with.
+-->
+<XnaContent xmlns:Graphics="Microsoft.Xna.Framework.Content.Pipeline.Graphics">
+  <Asset Type="Graphics:FontDescription">
+
+    <!--
+    Modify this string to change the font that will be imported.
+    -->
+    <FontName>Tahoma</FontName>
```
dellis1972, line 14: Is this font available on all Desktop Platforms? If not can we get one that is or ship one with the template? (We would need rights to be able to ship a font).

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Screens/GameplayScreen.cs
```diff
@@ -0,0 +1,396 @@
...
+        if (content == null)
+            content = new ContentManager(ScreenManager.Game.Services, "Content");
+
+        spriteBatch = ScreenManager.SpriteBatch;
+
+        // Load fonts
+        hudFont = content.Load<SpriteFont>("Fonts/Hud");
+
+        // Load overlay textures
+        winOverlay = content.Load<Texture2D>("Overlays/you_win");
```
dellis1972, line 106: Should these be using the `LoadLocalized` api? If you are going to have localization you need to make sure you have localized text even if graphics. See https://docs.monogame.net/articles/getting_started/content_pipeline/localization.html#other-localized-assets for details.

Same for fonts, you might be using a different SpriteFont if you are supporting CKJ languages.

CartBlanche:
> Saving/loading is a good one, especially in a way that works across platforms and serializing data.
> 
> Possibly leader board.
> 
> But all also need dedicated tutorials for the how to section too.

@SimonDarksideJ I'm thinking the tutorial docs would come shortly afterwards as well as separate samples for say, particle manager, toast message. So that in the tutorial we can see, we took this sample and inserted it here. Etc.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/___SafeGameName___.contentproj
CartBlanche: @dellis1972 I copied it from platformer. I'll see what the project templates generate.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Android/___SafeGameName___.Android.csproj
CartBlanche: @dellis1972 it doesn't appear to, everything runs fine, and maybe shouldn't because it's talking about the root directory and not the content file within the root directory. Anyone else know??

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/Fonts/Hud.spritefont
CartBlanche: This is placeholder for now, for testing in Windows. But we definitely need one we can ship.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Screens/GameplayScreen.cs
CartBlanche: @dellis1972 Nice pick-up. I'd forgotten about these. I have an idea on how that might work with graphics.

CKJ??

dellis1972: Chinese, Korean, Japanese (CKJ). They require specialised fonts which support the character sets they need.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/___SafeGameName___.contentproj
dellis1972: We should not be including the old XNA projects. They are not needed. Nothing can build them now afaik.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Android/AndroidManifest.xml
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="___SafeGameName___.Android" android:versionCode="1" android:versionName="1.0">
+	<uses-sdk android:minSdkVersion="23" android:targetSdkVersion="34" />
```
dellis1972, line 3: you can remove the `uses-sdk` element completely. This is filled in by the `SupportedOSPlatformVersion` property in the csproj.

SimonDarksideJ:
Agreed, using this template as the sample downloadable code would be a good way to go

dellis1972:
> Agreed, using this template as the sample downloadable code would be a good way to go

Does it need to be a template then? If the sample is going to be part of a tutorial?

SimonDarksideJ:
I think it's a nice approach as it is easily installable as a project of your own name, then you can remove /add as you wish.
It is also good boilerplate for hacks.

If we take this approach, we should do a basic 3d one too

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/Fonts/Hud.spritefont
CartBlanche: @dellis1972 @SimonDarksideJ I found this dude https://fonty.condak.cz/index.php who has quite a few completely FREE fonts, as well as commercial ones. Do you think it is worth us hosting the free ones, in case he disappears, and we store/promote them (with attribution of course) as being part of MonoGame and free for devs to you?

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Android/AndroidManifest.xml
CartBlanche: Thanks, fixed.

CartBlanche:
@dellis1972, @SimonDarksideJ I was also thinking that maybe we won't need the sample, as this should hopefully become the "preferred" starting point for 2D games.

SettingsManager class etc implemented, Saving and Loading between restarts appears to work. 
I have a Desktop and Mobile SettingsStorage classes and an empty Console class. Everything probably needs more testing.

I don't know if this happens already on console, but I'd like Environment.SpecialFolder.Personal  to map similarly to the Mobile.
To me that makes sense, but not sure how the console heads view that. Either way if it did, it would simplify our SettingsStorage story.

I don't think I'll have time this weekend, as away with family, but sometime next week I think I can get the last thing, local leaderboards, implemented.

tomspilman:
> I don't know if this happens already on console, but I'd like Environment.SpecialFolder.Personal to map similarly to the Mobile.

Not all consoles use plain folders and files for "save" storage.  So mapping to a file path isn't a viable cross platform solution.  Worse some consoles require you mount and unmount storage within a fixed time limit.  Others require you to "commit" results at the end otherwise the data is reverted back.  Some have options like automatic backups.

So it is a complicated subject.

At the moment we don't have a cross platform solution that works everywhere and console save game code requires per platform tweaking or writing your own high level API to manage it.

This would be a good area for us to write something new or adapt something old.

CartBlanche:
@dellis1972 , @SimonDarksideJ : Added
- [x] Player now dies if they fall a particular distance
- [x] Added support for breakable tiles, a la Mario

It's really starting to feel a little more like a complete game now.

CartBlanche:
@dellis1972, @SimonDarksideJ Latest commit allows mouse/touch to select menu items. 
Which got me thinking (probably not a good thing), I'm not a huge fan of overlays (old school IMO). What are your thoughts on using Touch/Mouse to "lead" the player around. Eg if I touch above the player's head he jumps, if I touch X pixels ahead they walk, X+Y pixels they run??

SimonDarksideJ:
Walking left and right should be enough, join mping is for whimps 👌😅

CartBlanche:
@SimonDarksideJ Please don't rebase without letting me know 1st.

CartBlanche:
@dellis1972 @SimonDarksideJ @tomspilman, @mrhelmut  I think in terms of features/functionality for a StartKit, this has everything that I think should be needed for people to hit the ground running with a basic 2D StartKit. 
In it's current state it probably isn't a "best practise" coding wise, which I'm hoping you seasoned game developers can help out with, but telling me what could be done better. I also assuming there may be a few gameplay bugs I've missed.

I've attached a DesktopGL zip file to this comment, if you want to download and try it out. I'm hoping to do some refactoring in the next few days, based on your comments.

[2DStartKit.zip](https://github.com/user-attachments/files/17548025/2DStartKit.zip)

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/Fonts/Hud.spritefont
```diff
@@ -11,7 +11,7 @@ with.
     <!--
     Modify this string to change the font that will be imported.
     -->
-    <FontName>Tahoma</FontName>
+    <FontName>Arial</FontName>
```
squarebananas, line 14: I remember this statement used to be in the XNA archive:
>Note that fonts are licensed and not sold, and that most commercial fonts, and most fonts supplied with Microsoft products such as Windows and Office, are not licensed for redistribution with games or other applications. It is your responsibility to make sure any fonts you include with your games are properly licensed for redistribution.

The main issue is that free fonts may allow use within documents/pictures/etc, but there are usually embedding & modifying restrictions, and `SpriteFont` XNB files are essentially embedded fonts in a modified bitmap format.

There was the "Redistributable Font Pack for XNA Game Studio" (https://github.com/SimonDarksideJ/XNAGameStudio/tree/archive/Samples/RedistributableTTFs_ARCHIVE_3_1/RedistributableTTFs), but even that license isn't clear if these are usable or for "XNA Game Studio" use only.

It seems Google Fonts (https://fonts.google.com) have "Open Font License" fonts that allow embedding, not used them myself though.

CartBlanche: @squarebananas we are discussing how best to support unicode characters internally, especially now that Microsoft no longer distribute the Arial Unicode MS TTF.

Though the licences aren't 100% clear I've yet to hear of any game Devs, even as far back as XNA 3.x, being sued for using an xnb version of their font. Always worth remembering than an xnb is usually a small subset of the full TTF. So it's not usually a 1 for 1 likeness and that's why I think game Devs aren't being chased. But it's definitely worth keeping an eye on the situation.

squarebananas: Yeah, I haven't heard of any issues either. I raised it more because this sample will be within the MonoGame repository, and the MonoGame licenses imply any users will have fairly open distribution/publishing rights of the code/assets (https://github.com/MonoGame/MonoGame/blob/develop/LICENSE.txt).

dellis1972: We are not embedding the .ttf in the xnb, we are rendering to a texture. That is allowed otherwise people who use fonts for artwork or images would be paying licenses per image download not just a one off for using the .ttf in the first place. 

This is someone people forget about this when they want MG to support direct font rendering.

squarebananas: Thanks for the response. Using fonts within documents & images should be permitted, but I don't think commercial companies would let that loophole exist for other usages. Especially as various bitmap formats were used before vector based ones were introduced.

https://learn.microsoft.com/en-us/typography/fonts/font-faq
>If I convert the font into a bitmap font can I include that in my game or app?
>>No, converting Windows fonts to other formats does not change the rules around embedding or redistribution, and format conversion itself is not allowed. Many Microsoft supplied fonts are available for app and game licensing through the original font foundry or Monotype.

https://learn.microsoft.com/en-us/previous-versions/windows/xna/bb464059(v=xnagamestudio.42)
>Font licenses vary from vendor to vendor, but most don't allow redistribution of the fonts, and that includes redistribution of reproductions such as bitmaps containing the rasterized character set.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/Sprites/Player/Die.png
squarebananas, line 1: PlayerDie
This animation contains 12 frames instead of 11 and isn't 64 pixel aligned.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/Fonts/Hud.spritefont
dellis1972: Yup sorry getting mixed messages. If you want to use a font in your game then yes you need a license to use the font in your game.  
I think what I was getting at was you probably don't need to pay a per install license ever time your game is installed.

squarebananas:
(Just saw more commits happened. These comments don't take those into account).

## Issues
1) **Best Issue** - Partial control of the player is possible after selecting "Play" but before loading the level. Invisible hyper pogo stick mode can be activated by holding up/left in this state.
2) The game sometimes freezes on exiting to main menu. Appears to be a threading dead lock issue when disposing fully played sounds.
3) You can survive any high fall by pressing/holding jump before landing.
4) While jumping hitting the player's head on a block doesn't stop the vertical velocity, meaning you can head slide around.
5) The player can hover at the side of block platforms (the collision could take into account shoe pixels to prevent this).
6) Pausing doesn't pause player/enemy animations.
7) Pressing space to resume also jumps or confirms next level / gameover popup.
8) Particles aren't cleared between game screens and visibly carry over between levels.
9) The menu's active selection scaling doesn't seem to have the correct origin.
10) Full screen doesn't maintain the aspect ratio.
11) Clicking "MonoGame Site" can open ~10 instances of the website.

## Suggestions

1. Language select could be shown first on the initial boot of the game (if the "settings.json" doesn't exist yet).
2. Allow window resizing, either with black borders or utilize some of the backgrounds that can seamlessly tile.
3. Add support for mouse forward/backward navigation buttons (`MouseState.XButton1`, `MouseState.XButton2`) for menu navigation.

CartBlanche:
@squarebananas, thanks for taking the time to play with the game. I will try to get round to fixing the issues you mentioned, but can't guarantee it will be in a timely manner as sometimes other issues/fixes take precedent over the StartKit. Also worth mentioning we accept PRs, so if you have the time, branch off my branch and submit a PR and I or someone on the team will take a look at it 🙂.

squarebananas:
> Also worth mentioning we accept PRs, so if you have the time, branch off my branch and submit a PR and I or someone on the team will take a look at it 🙂.

Ok, I'll see if I can fit that in. I'm busy with a few other things at the moment though. In addition I have been trying to port a game to DesktopVK, have more issues to write up with that and potential fixes to investigate further as well.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/Fonts/Hud.spritefont
viniciusjarina: You could use Noto Sans/Open Sans that have a permissive license. AFAIK

CartBlanche: @viniciusjarina the team is investigating a font fall back system, particularly when it comes to multi-byte character sets.

viniciusjarina: I see. NotoSansCJK covers pretty much all the majos languages characters.  Latin, Cyrilic, Chinese, Japanese, Korean https://github.com/notofonts/noto-cjk that's what I have been using

CartBlanche:
Latest commits fix enemy doing walking cycle when game is paused. He now politely stands there breathing impatiently ;)  and to keep the background "alive" gems  now continue to animate like the enemy does. Just looks better than having them frozen.
Also added simple mouse wheel support, so you can use that to scroll the menu items. May need tweaks.

Templates/MonoGame.Templates.CSharp/content/MonoGame.2D.StartKit/___SafeGameName___.Core/Content/Fonts/Hud.spritefont
CartBlanche: @viniciusjarina that maybe what we need to fall back to, as `Arial Unicode MS` is no longer being distributed or updating by Microsoft.

CartBlanche:
Closing this PR as Code ported to the `MonoGame.Templates` repo and and new PR here - https://github.com/MonoGame/MonoGame.Templates/pull/3

PR #8276, opened 2024-04-12
[Documentation]: Add documentation to the Effect classes

Maniekko:
This PR adds the documentation for the following classes in the `Graphics` namespace:
- `Effect`
- `EffectPass`
- `EffectAnnotation`
- `EffectParameter`
- `EffectTechnique`
- `EffectMaterial`
- `IEffectMatrices` (interface) (added 1 hour later)

## Notes
One of the doclines in `EffectMaterial` refers to `EffectMaterialContent` class, which is only available on DirectX.

## Reference:


SimonDarksideJ (approved):

SimonDarksideJ:
Made a few corrections directly to speed up the review.

@Maniekko I've accepted this, but can we please avoid the "monkey see monkey do" style of notation in the future, for example:

```csharp
/// <summary>
/// Gets the row count of this effect annotation.
/// </summary>
public int RowCount {get; private set;}
```

To a lay person, what would this even mean, what is a row and why is the count important?  There were a few of these.   And the style guide has a specific note against just spelling out the name of a parameter

MonoGame.Framework/Graphics/Effect/EffectMaterial.cs
```diff
@@ -5,11 +5,23 @@
 using System;
 namespace Microsoft.Xna.Framework.Graphics
 {
+    /// <summary>
+    /// Contains an effect subclass which is used to load data for an EffectMaterialContent type.
+    /// </summary>
+    /// <remarks>
+    /// For most purposes, this type can be ignored, and treated exactly like a regular effect.
+    /// When an EffectMaterial type is loaded from .xnb format, its parameter values and textures
+    /// are also loaded and automatically set on the effect, in addition to the HLSL shader code.
```
dellis1972, line 14: We might want to remove the reference to HLSL here. by the time the Effect is being used the onderlying shader will be platform specific , e.g GLSL HLSL, metal etc.

dellis1972 (approved):
Some minor questions which can probably be fixed up later if need be. 

Great work.

SimonDarksideJ:
@Maniekko can you look to see if you can address @dellis1972 comments above, looking to get all these merged this week

Maniekko:
> @Maniekko can you look to see if you can address `@dellis1972` comments above, looking to get all these merged this week

Sorry, currently too busy preparing to end of semester exams.

Issue #8084, opened 2023-10-23
DesktopGL: GraphicsDevice.ScissorRectangle is set only once and isn't updated if the window gets resized


rds1983:
### Prerequisites

- [X] I have verified this issue is not present in the `develop` branch
- [X] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

MonoGame 3.8.1.303

### Which MonoGame platform are you using?

MonoGame Cross-Platform Desktop Application (mgdesktopgl)

### Operating System

Windows

### Description

GraphicsDevice.ScissorRectangle is set to wrong value if GraphicsDevice.Clear is called.
Initially ScissorRectangle is set to the window size on the initialization: https://github.com/MonoGame/MonoGame/blob/a91dcc70b681d0230e960ec8eeaec4f9ff94b605/MonoGame.Framework/Graphics/GraphicsDevice.cs#L385

And seems that it never gets updated.

### Steps to Reproduce

1. Unpack the attached sample, open in the IDE and run.
2. Resize the window.
3. Notice that the image gets cut. That happens because of the wrong Graphics.ScissorRectangle.

### Expected Behavior

GraphicsRectangle should be synchronized with the ViewPort. That's how it works for MonoGame.WindowsDX.

### Resulting Behavior

https://github.com/MonoGame/MonoGame/assets/1057289/8d2513e6-3c71-4a39-85d3-5e7365eb3a3f

### Files
[MGScissorResize.zip](https://github.com/MonoGame/MonoGame/files/13066244/MGScissorResize.zip)

Mindfulplays:
When the viewport gets [resized, doesn't it update both `Viewport` and `ScissorRectangle`](https://github.com/MonoGame/MonoGame/blob/a91dcc70b681d0230e960ec8eeaec4f9ff94b605/MonoGame.Framework/Graphics/GraphicsDevice.cs#L900) ?

rds1983:
Well, check the sample. It's quite simple and shouldn't be cut. But it is cut.

samercer2002:
Hello, as this issue seems to still be present, is it open for me to take a look at? I'd like to work on it.

PR #8284, opened 2024-04-18
Issue #8084: DesktopGL GraphicsDevice.ScissorRectangle Fix

samercer2002:
Solution for issue [#8084](https://github.com/MonoGame/MonoGame/issues/8084).

Updates ClientResize function in order to update the ScissorRectangle when the viewport is resized. This change mimics the behavior found on other platforms such as WindowsDX.

### DIFF
--- a/MonoGame.Framework/Platform/SDL/SDLGameWindow.cs
+++ b/MonoGame.Framework/Platform/SDL/SDLGameWindow.cs
@@ -299,9 +299,16 @@ namespace Microsoft.Xna.Framework
                 _game.GraphicsDevice.PresentationParameters.BackBufferHeight == height) {
                 return;
             }
+
+            Viewport prevViewport = _game.GraphicsDevice.Viewport;
+            Rectangle prevScissorRect = _game.GraphicsDevice.ScissorRectangle;
+
             _game.GraphicsDevice.PresentationParameters.BackBufferWidth = width;
             _game.GraphicsDevice.PresentationParameters.BackBufferHeight = height;
             _game.GraphicsDevice.Viewport = new Viewport(0, 0, width, height);
+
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Platform/SDL/SDLGameWindow.cs
```diff
@@ -299,10 +299,17 @@ public void ClientResize(int width, int height)
...
             }
+
+            Viewport prevViewport = _game.GraphicsDevice.Viewport;
+            Rectangle prevScissorRect = _game.GraphicsDevice.ScissorRectangle;
+
             _game.GraphicsDevice.PresentationParameters.BackBufferWidth = width;
             _game.GraphicsDevice.PresentationParameters.BackBufferHeight = height;
             _game.GraphicsDevice.Viewport = new Viewport(0, 0, width, height);
 
+            if (_game.GraphicsDevice.RasterizerState.ScissorTestEnable && prevScissorRect == prevViewport.Bounds)
```
SimonDarksideJ, line 310: A question, if the check is only on the PREVIOUS viewport bounds (and not the updated version in lines 306-308), then why is this change AFTER the viewport update?

1. If the Scissor check is based on the PREVIOUS viewport bounds, then this condition should be moved to line 305.
2. If the Scissor check is AFTER the update to the viewport bounds, should this not be validating against the NEW and not the previous bounds?

SimonDarksideJ (changes requested):

MonoGame.Framework/Platform/SDL/SDLGameWindow.cs
samercer2002: After looking over it, it should have been at line 305 as you stated. Just pushed an update.

SimonDarksideJ (approved):
Grumble, Grumble, Brackets... But otherwise ok

dellis1972 (approved):

Issue #6936, opened 2019-11-08
(macOS) Renders to bottom left quarter when involving RenderTargets


meniku:
When using the 2D rendering in MonoGame, as soon as I begin involving RenderTargets, the game renders to the bottom left quarter of the screen. I don't even need to use the render targets at all! Just switching to one and switching back to the back buffer is enough to make it break.
I tried to debug MonoGame, it seems to be somehow related to the Viewport that's being set, but I couldn't really find an appropriate fix.

I have some workaround to be able to develop at least, but that won't do for I real game release I guess ( the commented out code in the following example ).

Maybe I'm just doing something fundamentally wrong:

```csharp
using Microsoft.Xna.Framework;
using Microsoft.Xna.Framework.Graphics;

namespace Project.MacOS
{
    public class DemoGame : Game
    {
        private GraphicsDeviceManager Graphics;
        private RenderTarget2D RenderTarget;
        private Texture2D Texture;
        private SpriteBatch SpriteBatch;

        public DemoGame()
        {
            IsFixedTimeStep = true;
            Graphics = new GraphicsDeviceManager( this );
            Graphics.SynchronizeWithVerticalRetrace = false;
            Graphics.PreferredBackBufferWidth = 320;
            Graphics.PreferredBackBufferHeight = 240;
            Graphics.IsFullScreen = false;
            Graphics.ApplyChanges();
        }

        protected override void Initialize()
        {
            base.Initialize();
            RenderTarget = new RenderTarget2D( Graphics.GraphicsDevice, 32, 32 );
            Texture = new Texture2D( Graphics.GraphicsDevice, 1, 1, false, SurfaceFormat.Color );
            Texture.SetData( new[] { Color.White } );
        }

        protected override void LoadContent()
        {
            SpriteBatch = new SpriteBatch( GraphicsDevice );
            base.LoadContent();
        }

        protected override void Draw( GameTime gameTime )
        {
            Graphics.GraphicsDevice.SetRenderTarget( RenderTarget );
            Graphics.GraphicsDevice.SetRenderTarget( null );
            Graphics.GraphicsDevice.Clear( new Color( 0.0f, 0.5f, 0.5f ) );

            // Workaround to make it work on my macbook
            //int fRetinaFactor = 2;
            //Viewport oldViewport = Graphics.GraphicsDevice.Viewport;
[...]

Issue #4802, opened 2016-04-25
[DesktopGL] Retina not supported


mrhelmut:
Hello,

I tried to add Hi-DPI / Retina support to the new ```SDLGameWindow``` but it turns out it is way more complicated than just adding the ```Sdl.Window.State.AllowHiDpi``` creation flag to the window.

I lack the OS X understanding and I don't get what OS X is expecting when it comes to backbuffer, window and viewport size. Their virtual DPI system is totally different from other systems and the documentation didn't quite helped me. There are some design decisions to take as well, because setting a 2880x1800 resolution may end up to a 1440x900 viewport, which is misleading compared to what XNA users are used to and specific to OS X. Do we want this? Do we need some virtualisation?

Right now, ```GraphicsDeviceManager.HardwareModeSwitch = true``` renders a black screen on any Retina displays, while the borderless mode displays games only on 1/4th of the screen (on the lower left). This means that it is a regression compared to OpenTK (which was supporting Hi-DPI, but only correctly working with the desktop resolution, or borderless).

I wouldn't mind some help on this front, or any hints on how OS X works with Hi-DPI, OpenGL and SDL2. ( @tomspilman mind adding the "help wanted" tag?)

Note that on Windows, enabling Hi-DPI is as simple as adding a manifest to enable it. No other action is required and it works as expected. I'll push a PR to add this to the DesktopGL template.

@dellis1972 Did you have any issues with Retina on the latest SDL changes?

mrhelmut:
Current situation, with Hi-DPI **disabled**:

```graphics.HardwareModeSwitch = true;``` **doesn't work** at all on Retina displays and Windows with 4k displays. It results in either a black screen or cropped viewport depending on the resolution begin under or above the desktop virtual resolution.

```graphics.HardwareModeSwitch = false;``` **works**, but doesn't expose the selected resolution, it forces games to use the DPI scaled desktop resolution (i.e. if DPI are x2 and the desktop resolution 1920x1080, viewport and backbuffer will be 960x540 no matter the ```PreferedBackBuffer```). This is a loss of quality if selected resolution is above desktop, and a loss of performance if under desktop. That's how SDL2 works in borderless mode and there's nothing to do about this.

With Hi-DPI enabled, nothing works on any OS on Retina / Windows4K.

cra0zy:
http://stackoverflow.com/questions/18544881/sdl-2-0-retina-mac

mrhelmut:
[...]

PR #8296, opened 2024-05-03
Add HiDPI support for MacOS (Retina)

shatyuka:
### Description
Bring Retina support to MonoGame. Now we can render at full resolution instead of 1/2.

### Details/Checklist
- [x] Set `SDL_WINDOW_ALLOW_HIGHDPI` to SDL window
- [x] Get the backing scale factor. Should be `-[NSWindow backingScaleFactor]`. But `SDL_GL_GetDrawableSize() / SDL_GetWindowSize()` is acceptable
- [x] Convert between points and pixels (viewport/back buffer/mouse state)
- [x] Only affects MacOS

### What users should do
1. Set `NSHighResolutionCapable` to true in Info.plist.
```xml
<key>NSHighResolutionCapable</key>
<true/>
```

It's recommended to create a separate project for MacOS and set target framework to have the `-macos` suffix.

2. Set window size to a proper scale (like what we do in Windows).
We can get scale factor from `NSScreen.MainScreen.BackingScaleFactor.Value`.

3. Handle screen change event and set back buffer size accoradingly.
Insert the following code between `using var game = new Game()` and `game.Run()`.
```csharp
NSApplication.Init();
var window = NSApplication.SharedApplication.DangerousWindows[0];
NSNotificationCenter.DefaultCenter.AddObserver(NSApplication.DidChangeScreenParametersNotification, action, null);
NSNotificationCenter.DefaultCenter.AddObserver(NSWindow.DidChangeScreenNotification, action, window);
```

### Related
#4802 #6936


mrhelmut:
Would you mind rebasing this PR so that the build passes the checks? I'll proceed to reviewing it.

dellis1972:
@shatyuka can I assume that if a user does not put `NSHighResolutionCapable` in the info.plist, then it will behave as it currently does?

shatyuka:
> @shatyuka can I assume that if a user does not put `NSHighResolutionCapable` in the info.plist, then it will behave as it currently does?

Yes.

frauber84:
Any chance of this being included in the milestones for MonoGame 3.8.2? It would definitely improve developer experience on macOS.

mrhelmut:
I won't have the opportunity to test this thoroughly before 3.8.2 happens.

harry-cpp:
Adding the `AllowHighDPI` flag is fine, but doing any scaling inside the MG itself shouldn't happen.

SimonDarksideJ:
@harry-cpp  / @mrhelmut any chance to retest this PR before the `3.8.3` release?

CartBlanche:
@shatyuka I'll try to test this, this week.

CartBlanche:
@shatyuka I just noticed that there seems to be some duplicate code here:
https://github.com/MonoGame/MonoGame/pull/8296/files#diff-5a1b29fb9825355ba15bad94d0f296ee4a6999224f31ea0890abb77d4537963cR143-R147

and here

https://github.com/MonoGame/MonoGame/pull/8296/files#diff-5a1b29fb9825355ba15bad94d0f296ee4a6999224f31ea0890abb77d4537963cR171-R175

Is this check required in both locations?

shatyuka:
> Is this check required in both locations?

Yes, I think every window should have the `AllowHighDPI` flag.

mrhelmut:
I don't think that we should force high DPI support to every developers, it should be a switch.

Also, I think that this PR is not enough to properly handle DPI on macOS. MonoGame needs to redesign how backbuffer and window size are coupled, as-is there are a lot of cases where the rendering will break, like changing render targets, applying backbuffer-size dependent effects or having multiple monitors. macOS has a very shitty way of handling DPI when it comes to games (it might be cool for apps and all, but for games it's a just a mess). macOS will change the backbuffer size on a whim when the app is "retina aware", so you basically need to query the backbuffer size each frame if you don't want to break rendering on every occasions. For the time being, MonoGame is too strict about how the window size should be bound to the backbuffer size. Applying just a scale ratio is clearly not enough, it cascades into more stuff like the viewport, bounds, etc.

MonoGame.Framework/Platform/Input/Mouse.SDL.cs
```diff
@@ -39,14 +40,30 @@ private static MouseState PlatformGetState(GameWindow window)
...
             PrimaryWindow.MouseState.X = x;
             PrimaryWindow.MouseState.Y = y;
-            
+
+            // Get scale from primary window... Should be okay
+            if (CurrentPlatform.OS == OS.MacOSX)
+            {
+                var scale = ((SdlGameWindow)PrimaryWindow).Scale;
+                y = (int)(y * scale);
+                y = (int)(y / scale);
```
PerryAsleep, line 64: I stumbled across this PR when trying to fix MacOS HiDPI scaling in my application. Thank you for the work you've done here. Is there a typo in these two lines? Looks like the intent is to be scaling x and y.

shatyuka: oops, it should be `x = (int)(x / scale);` here?

ThomasFOG:
> I don't think that we should force high DPI support to every developers, it should be a switch.
> 
> Also, I think that this PR is not enough to properly handle DPI on macOS. MonoGame needs to redesign how backbuffer and window size are coupled, as-is there are a lot of cases where the rendering will break, like changing render targets, applying backbuffer-size dependent effects or having multiple monitors. macOS has a very shitty way of handling DPI when it comes to games (it might be cool for apps and all, but for games it's a just a mess). macOS will change the backbuffer size on a whim when the app is "retina aware", so you basically need to query the backbuffer size each frame if you don't want to break rendering on every occasions. For the time being, MonoGame is too strict about how the window size should be bound to the backbuffer size. Applying just a scale ratio is clearly not enough, it cascades into more stuff like the viewport, bounds, etc.

We should postpone and rethink DPI support for macOS. As-is it doesn't entirely address the issue and might make it worse in some cases.

SimonDarksideJ:
While a good start @shatyuka , sadly we need to close this PR in favour of an implementation that targets the new native platform and include it with the DesktopVK platform target.

PR #8299, opened 2024-05-08
Added error detection for effect content files

Nebulaxin:
MSBuild can detect warnings and errors from standard output of `<Exec>` task, which is used to build content with `mgcb`, and these detected warnings and errors will be displayed by IDE or text editor.

This is how it looks:
warning and errors


SimonDarksideJ:
Thoughts @mrhelmut / @tomspilman , this looks particularly interesting.

SimonDarksideJ (approved):

Issue #8242, opened 2024-03-15
[PAID BOUNTY][LOCKED] A shared base for public and console repos


harry-cpp:
_MonoGame needs a new base that can be used between public and private console repos._

📃**Full details**

MonoGame provides console ports of itself in private repos, all you need to do to access it is to be a registered developer for the correct platform. They are in private repos because of the console NDAs.

In order to make the work between the public and private repos less separate, we wanted to move MonoGame to a new system where there will be one C# **MonoGame.Framework.dll** assembly that can be used for both. It would containt only the C# wrapper calls to our custom **monogameframework** native library, then we could switch out the native library depending on the platform you are targeting.

🔒 Bounty status

Locked for @tomspilman to complete. 🔒

💰 **Bounty**

$1200,00 USD (One thousand and two hundred US dollars).

✋ **Requirements to apply**

_Standard requirements from [our generic bounty requirements](https://monogame.net/bounties/#requirements) as well as knowlage on C++, C++ build systems, Vulkan and SDL._

📦 **Expected delivery content**

- **MonoGame.Frameowk.Native** C# project
  - All C# Stubs need to be fully filled with wrapper calls to **monogameframework**
  - A generator needs to be writted to export all the wrappers to header files that need to be filled using reflection
  - Needs to be hooked up to the tests and passing non graphics / audio tests
- **monogameframework** native library
  - This library needs to contain exposed methods for all the platform call MonoGame does
  - All platform calls need to be fully implemented (Windowing + Input, using SDL)
  - Vulkan code needs to be moved from the private MonoGame repo
    - All graphics calls need to be fully hooked up, it does not need to be fully working
  - No audio call should be implemented, we are planing on using FAudio and a relavent bounty for it will be released at future time
- Both projects need to be fully hooked up to the CAKE Frosting build system

📅 **Delivery timeline**

As soon as possible.

tomspilman:
FYI.  As part of this i will move the sharable bits of the DX12 work @Togimaro did for https://github.com/MonoGame/MonoGame/issues/8195 into the public MG repo under this new native API.

tomspilman:
This bounty is completed.

PR #8309, opened 2024-05-20
[PAID BOUNTY] A shared base for public and console repos

tomspilman:
This is a shared native base for future platform work in MonoGame.

See https://github.com/MonoGame/MonoGame/issues/8242

The implementation is now finished and we're looking for final feedback before we merge things.

- ~~C# to C++ wrapper generator.~~
- ~~C# native wrapper.~~
- ~~Windows VK setup.~~
- ~~C++ native wrapper.~~
- ~~SDL native backend.~~
- ~~Vulkan rendering backend.~~
- ~~Input native backend.~~
- ~~WindowsVK unit tests.~~
- ~~Sound API stubs~~


MonoGame.Framework/Platform/Native/Interop.cs
```diff
@@ -0,0 +1,88 @@
...
+internal readonly struct MGG_Texture { }
+
+internal struct MGG_GraphicsDevice_Caps
+{
+    public int MaxTextureSlots;
+    public int MaxVertexTextureSlots;
+    public int MaxVertexBufferSlots;
+}
+
+internal static unsafe partial class MGG
```
tomspilman, line 36: We need to wart these with something that works for us in C# and in C++ (don't want conflicts with other native libs).

So i'm thinking:

- MGG for Graphics.
- MGA for Audio.
- MGI for Input.
- MGF for Framework?

Thoughts?

src/monogame/include/csharp_MGG.h
```diff
@@ -0,0 +1,26 @@
...
+//
+
+#pragma once
+
+#include "csharp_common.h"
+#include "csharp_enums.h"
+#include "csharp_structs.h"
+
+
+struct MGG_GraphicsDevice;
```
tomspilman, line 13: The idea here is the C++ implementation will define the internals of these pre-declared structures.

For example one for DX could look like:

```
struct MGG_GraphicsDevice
{
	ID3D12Device* _device;
	ID3D12GraphicsCommandList* _commandList;
	ID3D12CommandQueue* _commandQueue;
       // etc!
}
```

This would be how native platforms pass along its data across pinvokes.

The C# just passes pointers and the C++ is responsible for allocation and freeing of these types.

src/monogame/include/csharp_common.h
```diff
@@ -0,0 +1,11 @@
...
+
+typedef char                 cschar;
+typedef unsigned char        csbyte;
+typedef short                csshort;
+typedef unsigned short       csushort;
+typedef int                  csint;
+typedef unsigned int         csuint;
+typedef long long            cslong;
+typedef unsigned long long   csulong;
+typedef bool                 csbool;
```
tomspilman, line 11: Do you think it is worth making this for example `mgbool` instead of `csbool` ?

src/monogame/include/csharp_enums.h
```diff
@@ -0,0 +1,156 @@
+//
+// This code is auto generated, don't modify it by hand.
+// To regenerate it run: Tools/MonoGame.Generator.CTypes
+//
+
+#pragma once
+
+#include "csharp_common.h"
+
+enum CSSurfaceFormat : csint
```
tomspilman, line 10: Same here... `MGSurfaceFormat` instead of `CSSurfaceFormat` ?

Tools/MonoGame.Generator.CTypes/PInvokeWritter.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        if (type.FullName.StartsWith("System."))
+            return false;
+
+        // Our handle types don't need to be exported
+        // and these are all readonly structs.
+        //
+        // So skip these.
+        //
+        if (type.GetCustomAttribute<IsReadOnlyAttribute>() == null)
```
tomspilman, line 69: Feels sloppy depending on the `readonly` flag to identify our handle types.

Any better ideas?

src/monogame/include/csharp_common.h
harry-cpp: yes

Tools/MonoGame.Generator.CTypes/PInvokeWritter.cs
harry-cpp: My idea was to skip the types which names end in `Ptr`.

tomspilman:
> Both projects need to be fully hooked up to the CAKE Frosting build system

@harry-cpp what do i need to do for this?

Is it just building?  Or is it also a project generator for C++?

I had assumed we would maintain C++ projects manually on a per-platform basis.

MonoGame.Framework/Platform/Native/Interop.cs
```diff
@@ -0,0 +1,171 @@
...
+        int dataBytes);
+
+    #endregion
+}
+
+
+/// <summary>
+/// MonoGame image native calls.
+/// </summary>
+internal static unsafe partial class MGI
```
tomspilman, line 141: I think we should ditch StbSharp for native stb directly since we have a native library.

This API exposes reading images (jpg/png) from memory as well as writing jpg and png.

We could implement more here in the future.

tomspilman:
@harry-cpp @mrhelmut

I assume we'll have one Windows C++ project for a `monogame,native.dll` that depending on build configuration would build Vulkan or DirectX?

Or should we plan on multiple C++ Windows projects?

I assume we won't support runtime swapping of Vulkan or DX12?  Especially since we pre-build the effects for the platform.

So we will end up with `WindowsVK` and `WindowsDX` as platforms sort of like we had before with `WindowsGL`.

Also i assume WindowsGL is going away.

mrhelmut:
> I assume we'll have one Windows C++ project for a `monogame.native.dll` that depending on build configuration would build Vulkan or DirectX?

I don't have a strong opinion on this. Whatever seems smooth enough in regard to Linux, macOS, and consoles moving to that too.

> I assume we won't support runtime swapping of Vulkan or DX12? Especially since we pre-build the effects for the platform. (IMO this is a gimmick and not actually useful in shipping games)

It doesn't feel useful to me. Not on the C++ at least. Worst case we can manage that on the C# side of things with a simple switch loading either native dll at launch.

harry-cpp:
> @harry-cpp what do i need to do for this?

It should:
- build all the generator projects
- run all the generators
- ensure no file is now in modified state compared to before running the generators (aka. all the generated files were properly committed)
- build all the native projects

> Is it just building? Or is it also a project generator for C++?
> 
> I had assumed we would maintain C++ projects manually on a per-platform basis.

CAKE is just building the projects. But.

We should use some cross platform solution for building the projects as VS projects are Windows only. Some of them even support generating a VS project for you to work in. Good options are meson and cmake.

MonoGame.Framework/Graphics/Texture3D.cs
```diff
@@ -74,11 +74,7 @@ protected Texture3D (GraphicsDevice graphicsDevice, int width, int height, int d
 		{
             ValidateParams(level, left, top, right, bottom, front, back, data, startIndex, elementCount);
 
-            var width = right - left;
```
tomspilman, line 77: I figured i would fix bad platform apis when i could.

Here it was passing left, top, right, bottom, front, back....  AND width, height, depth. 

So i removed the extra data passing that not all platforms need.

MonoGame.Framework/Graphics/SamplerStateCollection.cs
```diff
@@ -22,9 +22,9 @@ public sealed partial class SamplerStateCollection
 
         private readonly SamplerState[] _samplers;
         private readonly SamplerState[] _actualSamplers;
-        private readonly bool _applyToVertexStage;
+        private readonly ShaderStage _stage;
 
-		internal SamplerStateCollection(GraphicsDevice device, int maxSamplers, bool applyToVertexStage)
+		internal SamplerStateCollection(GraphicsDevice device, int maxSamplers, ShaderStage stage)
```
tomspilman, line 27: This felt like a better design moving forward.

src/monogame/include/csharp_GameWrapper.h
```diff
@@ -0,0 +1,20 @@
...
+// file 'LICENSE.txt', which is part of this source code package.
+                
+//
+// This code is auto generated, don't modify it by hand.
+// To regenerate it run: Tools/MonoGame.Generator.CTypes
+//
+
+#pragma once
+
+#include "csharp_common.h"
```
tomspilman, line 12: The prefix `csharp_` feels wrong here.  Better ideas?  `gen_` or `api_` or something?

src/monogame/vulkan/RebuildMGFX.bat
```diff
@@ -0,0 +1,19 @@
+@echo off
```
tomspilman, line 1: Decided it was best to keep the generated binary effect files in per-platform folders instead of all in a common place.

This seems like a good direction considering console support.

MonoGame.Framework/Platform/Native/TitleContainer.Native.cs
```diff
@@ -15,6 +15,13 @@ partial class TitleContainer
...
     {
-        return Stream.Null;
+        // TODO: How do we handle this?
+        //
+        // It seems like we still need platform specific C# code here?
+        //
+        // Or do we make the native code transform the string?
+        //
+        var absolutePath = Path.Combine(Location, safeName);
+        return File.OpenRead(absolutePath);
```
tomspilman, line 25: Will be moving the file path fixup to native code.

MonoGame.Framework/Platform/Graphics/Effect/Resources/Structures.fxh
```diff
@@ -80,7 +80,7 @@ struct VSInputNmTxWeights
     float4 Position : POSITION0;
     float3 Normal   : NORMAL0;
     float2 TexCoord : TEXCOORD0;
-    uint4  Indices  : BLENDINDICES0;
```
tomspilman, line 83: This works around an issue DXC/Glsl where unsigned int 4 didn't compile.  This shouldn't matter as you never have huge numbers for blend indices.

MonoGame.Framework/Graphics/Vertices/VertexBufferBindings.cs
```diff
@@ -165,5 +165,23 @@ public VertexBufferBinding[] Get()
 
             return bindings;
         }
+
+        /// <summary>
+        /// Creates an <see cref="ImmutableVertexInputLayout"/> that can be used as a key in the
+        /// <see cref="InputLayoutCache"/>.
+        /// </summary>
+        /// <returns>The <see cref="ImmutableVertexInputLayout"/>.</returns>
+        internal ImmutableVertexInputLayout ToImmutable()
```
tomspilman, line 174: Moved this from DirectX specific code to be shared between platforms.

MonoGame.Framework/Graphics/TextureCollection.cs
```diff
@@ -13,14 +13,14 @@ public sealed partial class TextureCollection
     {
         private readonly GraphicsDevice _graphicsDevice;
         private readonly Texture[] _textures;
-        private readonly bool _applyToVertexStage;
+        private ShaderStage _stage;
         private int _dirty;
 
-        internal TextureCollection(GraphicsDevice graphicsDevice, int maxTextures, bool applyToVertexStage)
+        internal TextureCollection(GraphicsDevice graphicsDevice, int maxTextures, ShaderStage stage)
```
tomspilman, line 19: This seems like a better approach considering we may eventually have more shader stages.

src/monogame/vulkan/MGG_Vulkan.cpp
```diff
@@ -0,0 +1,973 @@
...
+	*bytecode = nullptr;
+	*size = 0;
+
+	// Get the handle of this DLL.
+	HMODULE module;
+	GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, (LPCSTR)&MGG_EffectResource_GetBytecode, &module);
+
+	LPCSTR id = nullptr;
+
+	if (strcmp(name, "AlphaTestEffect") == 0)
```
tomspilman, line 271: Should we consider making an enum for the stock effects instead of passing around string names here?

src/monogame/vulkan/MGG_Vulkan.cpp
```diff
@@ -0,0 +1,973 @@
...
+			system->adapters.push_back(adapter);
+		}
+	}
+
+	free((void*)gpus);
+}
+
+static MGG_System* System_Get()
+{
+	if (mgg_system != nullptr)
```
tomspilman, line 329: I'm debating if we need some object like `MGG_System` exposed at the C# level to manage graphics state before a graphics device is created...  like initializing the Vulkan API (or initializing DX12).

Or should we let the native code manage this internally?

src/monogame/vulkan/MGG_Vulkan.cpp
```diff
@@ -0,0 +1,973 @@
...
+	if (err != VK_SUCCESS)
+	{
+		printf("Failed to initialize volk!");
+		return nullptr;
+	}
+
+	VkApplicationInfo app_info = { VK_STRUCTURE_TYPE_APPLICATION_INFO };
+	app_info.pNext = nullptr;
+	app_info.applicationVersion = VK_MAKE_VERSION(1, 0, 0);
+	app_info.pApplicationName = "None";
```
tomspilman, line 342: Ideally we pass down an application name...  i assume this is used by drivers and utilities for optimizing/tweaking settings for specific games.

src/monogame/vulkan/MGG_Vulkan.cpp
```diff
@@ -0,0 +1,973 @@
...
+	// Is there a description somewhere for us to get?
+	static const char* description = "";
+	info->Description = (void*)description;
+
+	info->SubSystemId = 0;
+	info->MonitorHandle = 0;
+
+#ifdef _WIN32
+
+	HMONITOR primaryMonitor = MonitorFromPoint(POINT{ 0, 0 }, MONITOR_DEFAULTTOPRIMARY);
```
tomspilman, line 408: This is probably all wrong and needs to be rewritten, but it works for this first pass at Vulkan.

src/monogame/include/stl_common.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include <assert.h>
+#include <vector>
+#include <string>
+#include <queue>
+#include <map>
+
+
+template <class T>
+void remove_by_value(std::vector<T>& vector, const T& element)
```
tomspilman, line 16: For now i'm depending on STL for vectors, string, queue, maps, etc.

I have seen differences in functionality and performance across platforms over the years.

I wonder if eventually we want our own small common set of native helper collections to use across our native backends?

MonoGame.Framework/GameWindow.cs
```diff
@@ -34,13 +34,13 @@ public abstract class GameWindow
         /// </summary>
         public virtual bool AllowAltF4 { get { return _allowAltF4; } set { _allowAltF4 = value; } }
 
-#if (WINDOWS && !WINDOWS_UAP) || DESKTOPGL
```
tomspilman, line 37: Thinking we should just expose this on all platforms and let it be ignored on platforms that cannot use it.

Tools/MonoGame.Generator.CTypes/PInvokeWritter.cs
Mindfulplays: Perhaps a fully qualified attribute? Say annotate with `[MGHandleAttribute]`  which works well with the reflection here as you have the `type`.

tomspilman: Yep this worked better:

```
[MGHandle] internal readonly struct MGG_GraphicsSystem { }
```

Thanks @Mindfulplays

src/monogame/vulkan/MGG_Vulkan.cpp
harry-cpp: Yes, vulkan has fields to pass both game name and engine name so the driver can optimise it if it so chooses.

src/monogame/vulkan/MGG_Vulkan.cpp
harry-cpp: Yes.

MonoGame.Framework/GameWindow.cs
harry-cpp: Agreed.

src/monogame/include/stl_common.h
harry-cpp: I think we should only implement our own if we come across an issue with performance.

MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
```diff
@@ -102,7 +102,12 @@ public enum TargetPlatform
         /// <summary>
         /// WebAssembly and Bridge.NET
         /// </summary>
-        Web
+        Web,
+
+        /// <summary>
+        /// All desktop versions using Vulkan.
+        /// </summary>
+        DesktopVK,
```
tomspilman, line 110: @mrhelmut @harry-cpp this is the plan correct?

tomspilman:
For visibility here is the initial rendering under SDL2 + Vulkan on the new native backend:

https://github.com/MonoGame/MonoGame/assets/103522/3cf4b16b-f248-4384-947d-6d7b73dae2ca

Has bugs, but we have progress too.

MonoGame.Framework/MonoGame.Framework.Native.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
dogruis, line 1: I wonder why git is showing a diff here

tomspilman: A whitespace change of some sort.  It happens.

src/monogame/sdl/MGP_sdl.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+    while (SDL_PollEvent(&ev) == 1)
+    {
+        switch (ev.type)
+        {
+        case SDL_QUIT:
+            platform->exit = true;
+            break;
+
+        /*
```
dogruis, line 80: Just a reminder to not leave so much comments.

tomspilman: WIP...  i expect these will be restored in the next week.

MonoGame.Framework/Platform/Native/GamePlatform.Native.cs
```diff
@@ -11,54 +17,138 @@ partial class GamePlatform
...
+    }
+
+    internal static unsafe MGG_GraphicsSystem* GraphicsSystem
+    {
+        get
+        {
+            if (_system == null)
+            {
+                Environment.SetEnvironmentVariable("VK_LOADER_DEBUG", "all");
+                Environment.SetEnvironmentVariable("VK_LAYER_PATH", "C:\\VulkanSDK\\1.3.283.0\\Bin");
```
dogruis, line 45: Looks very os specific

tomspilman: Yeah I already removed it. It was some hacked code when I was trying to fix Vulkan validation.

src/monogame/vulkan/MGG_Vulkan.cpp
```diff
@@ -0,0 +1,3958 @@
...
+
+	std::map<uint32_t, MGG_BlendState*> blendStates;
+	std::map<uint32_t, MGG_RasterizerState*> rasterizerStates;
+	std::map<uint32_t, MGG_DepthStencilState*> depthStencilStates;
+
+	MGVK_TargetSet targets;
+	std::map<uint32_t, MGVK_TargetSetCache*> targetCache;
+
+
+	//
```
dogruis, line 251: Empty comment

MonoGame.Framework/Platform/Native/Platform.Interop.cs
```diff
@@ -0,0 +1,167 @@
...
+    public nint Window;
+    public uint Character;
+    public Keys Key;
+}
+
+[StructLayout(LayoutKind.Explicit)]
+internal struct MGP_Event
+{
+    [FieldOffset(0)]
+    public EventType Type;
```
harry-cpp, line 54: Add a field for Timestamp bellow Type.

src/monogame/sdl/MGP_sdl.cpp
```diff
@@ -0,0 +1,386 @@
...
+	assert(identifier != nullptr);
+
+	auto window = new MGP_Window();
+	window->platform = platform;
+	window->identifier = identifier;
+
+    // TODO: Why did this start with SDL_WINDOW_FULLSCREEN_DESKTOP in the old C# SDL?
+    // We should write coments to document odd behaviors like this.
+
+	Uint32 flags = SDL_WINDOW_HIDDEN;// | SDL_WINDOW_FULLSCREEN_DESKTOP;
```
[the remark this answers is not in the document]
harry-cpp: On DesktopGL we create the window twice, once before initialize to setup some opengl stuff, and once afterwards to recreate it with the set attributes: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/GraphicsDeviceManager.SDL.cs#L52

MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
mrhelmut: Yeah

tomspilman:
So getting this all to build how the core team wants it to was a pain...  it has complicated and delayed the PR, but i think i'm close finally.

I ditched Xmake for Premake (https://premake.github.io/) which ONLY generates project files.

Also i put the static SDL2 library into our SDL2 nuget:

  https://github.com/MonoGame/MonoGame.Library.SDL/pull/3

This solves our need for static libs without just including them in the repo.

If all this works out we can finish up this PR in the next day or two.

build/BuildFrameworksTasks/BuildDesktopVKTask.cs
```diff
@@ -0,0 +1,25 @@
...
+        // Generate the native projects.
+        var exit = context.StartProcess("premake5", new ProcessSettings { WorkingDirectory = "src/monogame", Arguments = "clean" });
+        if (exit < 0)
+            throw new Exception($"Setting Premake clean failed! {exit}");
+        exit = context.StartProcess("premake5", new ProcessSettings { WorkingDirectory = "src/monogame", Arguments = "--os=windows --verbose vs2022" });
+        if (exit < 0)
+            throw new Exception($"Setting Premake generation failed! {exit}");
+
+        // Build it.
+        exit = context.StartProcess("msbuild", new ProcessSettings { WorkingDirectory = "src/monogame", Arguments = "monogame.sln /p:Configuration=Release /p:Platform=x64" });
```
tomspilman, line 21: @harry-cpp - I'm getting the error in the workflow:

> An error occurred when executing task 'Build DesktopVK'.
Error: An error occurred trying to start process '"msbuild"' with working directory 'D:/a/MonoGame/MonoGame/src/monogame'. The system cannot find the file specified.

Any ideas why this is happening?  It seems like that folder clearly exists.

tomspilman: Or is it not finding `msbuild`?

tomspilman: The issue is that `msbuild` was not installed as part of the workflow.  Fixed it now.

tomspilman:
Ok the build process works.

Off to finish some APIs and this bounty is completed.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
```diff
@@ -0,0 +1,299 @@
...
+    TextInput,
+
+    MouseMove,
+    MouseButtonDown,
+    MouseButtonUp,
+    MouseWheel,
+
+    ControllerAdded,
+    ControllerRemoved,
+    ControllerStateChange,
```
tomspilman, line 36: @mrhelmut @harry-cpp I opted for a simple 3 event design which seems to work well.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
```diff
@@ -0,0 +1,299 @@
...
+
+[StructLayout(LayoutKind.Sequential)]
+internal struct MGP_MouseWheelEvent
+{
+    public nint Window;
+    public int Scroll;
+    public int ScrollH;
+}
+
+internal enum ControllerInput : int
```
tomspilman, line 91: @mrhelmut @harry-cpp By having one input type instead of seperate Axis and Button ones we can send just one event and not have extra unused fields or multiple structures.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
```diff
@@ -0,0 +1,299 @@
...
+{
+    public int Id;
+
+    public ControllerInput Input;
+
+    /// <summary>
+    /// For a axis input this can range from -32768 to 32767.
+    /// For a button input the value is 0 or 1.
+    /// </summary>
+    public short Value;
```
tomspilman, line 138: @mrhelmut @harry-cpp I followed the SDL standard for axis range.

All our other platforms will have to convert to it before passing the event to the C# side.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
tomspilman: Done.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
```diff
@@ -0,0 +1,299 @@
...
+
+
+[StructLayout(LayoutKind.Explicit)]
+internal struct MGP_Event
+{
+    [FieldOffset(0)]
+    public EventType Type;
+
+    [FieldOffset(4)]
+    public ulong Timestamp;
```
tomspilman, line 157: I'm using the 64bit timestamp from `SDL_GetTicks64()`  BUT it seems SDL returns 32bit time stamps in their events.  Need advice here @mrhelmut @harry-cpp .

MonoGame.Framework/Platform/Native/Platform.Interop.cs
```diff
@@ -0,0 +1,299 @@
...
+    public static partial void Window_GetPosition(MGP_Window* window, out int x, out int y);
+
+    [LibraryImport(MonoGameNativeDLL, EntryPoint = "MGP_Window_SetPosition", StringMarshalling = StringMarshalling.Utf8)]
+    public static partial void Window_SetPosition(MGP_Window* window, int x, int y);
+
+    #endregion
+
+    #region Mouse
+
+    [LibraryImport(MonoGameNativeDLL, EntryPoint = "MGP_Mouse_SetVisible", StringMarshalling = StringMarshalling.Utf8)]
```
tomspilman, line 277: Should i split out input stuff into its own interop static class?  Or place it all in `MGP` platform class?

Tools/MonoGame.Generator.CTypes/EnumWritter.cs
```diff
@@ -0,0 +1,117 @@
...
+        };
+
+        """);
+
+        // TODO: This causes issues with enums that have repeat
+        // values causing the `switch` to not work.
+        //
+        // We don't use this feature now anyway.
+        /*
+        _outputText.AppendLine($$"""
```
tomspilman, line 86: @harry-cpp - I disabled the C++ ToString because with enums like this:

```
enum ControllerInput
{
   A,
   B,
   X,
   Y,
  LastButton = Y,
}
```

It has issues. 

If you build it from enum Values some have names that are not ideal.

If you build it from enum Names the switch cases conflict and how can you tell which to return.

So i disabled this as we don't need it for now.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
tomspilman: Another question... should i expand this to support the MG `Joystick` APIs thru the same events?

What makes things a joystick vs gamepad?

I REALLY want to unify this to keep the native API simple.

@mrhelmut @harry-cpp

squarebananas: > What makes things a joystick vs gamepad?

In XNA terms, as far as I'm aware Joysticks were DirectInput devices and GamePads were XInput devices (https://learn.microsoft.com/en-us/windows/win32/xinput/xinput-and-directinput).

GamePads are typically more standardized with 2 thumbsticks, 10 buttons and 2 triggers. So a flight sim controller with 20 buttons and 8 analogue values for the stick and engine throttles would not have XInput support and wouldn't appear as a GamePad. The XNA GamePad Enums also wouldn't exist for all the extra values. Vibration is handled differently as well (SetVibration left/right instead of force feedback implementation).

Also with GamePads potentially being wireless, XInput added a connected status and handles controller assignment swapping (for example when the batteries go flat, turning on another controller can replace the original number assignment instead of adding a second controller).

In MonoGame DesktopGL terms, DirectInput joysticks that are actually gamepads can be automatically mapped by SDL as GamePads. That is why PS4 controllers (DirectInput) work on DesktopGL but not on WindowsDX.

The joystick implementations for MonoGame currently do not exist for WindowsDX and Android (maybe others) as well.

src/monogame/common/MGM.cpp
```diff
@@ -0,0 +1,92 @@
...
+	// portable to all our target platforms:
+	//
+	// libvorbis
+	// minimp3
+	// wave ??
+	// mp4 ??
+	// FLAC ??
+	//
+	// It should then spin up thread which decodes the
+	// audio and streams buffers to a native SoundEffect.
```
tomspilman, line 33: @mrhelmut @harry-cpp 

So the idea here is Song can be shared code across all platforms.

It takes care of streaming a music file from disk and pushing decoded packets into the platform specific sound effect.

squarebananas: Songs with some platforms are played via SoundEffects, however where possible they are played via the Platform's built-in media player and passed by URI. 

My understanding is this should provide integration (where possible) with not playing over other music apps (MediaPlayer.GameHasControl), playing/shuffling music from the user's own music library, allowing playback of DRM protected files in that library, getting album artwork, audio visualization data, etc.

Quite often I have seen developer's ask how to cross fade between 2 Songs or play additional layer tracks on top of each other. Song isn't really designed for this (one song for the device at any time) whereas SoundEffect fits this purpose well. At a later point it might be useful to have a PlatformBehavior setting for playback preference (by device media player or SoundEffects).

MonoGame.Framework/Media/Video.cs
```diff
@@ -83,7 +83,7 @@ void Dispose(bool disposing)
         {
             if (!_disposed)
             {
-                //PlatformDispose(disposing);
+                PlatformDispose(disposing);
```
tomspilman, line 86: Not sure why this was disabled.

I put it back...  let the platform specific code handle it internally.

MonoGame.Framework/Platform/Native/Video.Native.cs
```diff
@@ -3,13 +3,180 @@
...
+
+        // TODO: We should expose the back buffer count so that
+        // video playback and optimize the number of textures created
+        // for video frames.
+        int cachedFrameNum = 2; // Game.Instance.GraphicsDevice.BackBufferCount;
+
+        // TODO: May be worth letting the video return texture
+        // format, so it can avoid CPU conversions when possible.
+
+        _video = MGM.Video_Create(mediaFilePath, cachedFrameNum, out var width, out var height, out var fps, out var duration);
```
tomspilman, line 33: Using the same design here as with Song.

The plan is that video does all the playback work internally managing some textures and an audio stream from a playback thread.

MonoGame.Framework/Platform/Threading.cs
```diff
@@ -148,6 +150,41 @@ internal static void BlockOnUIThread<TState>(Action<TState> action, TState state
...
+
+            OnUIThread(_metaAction, action);
+        }
+
+        /// <summary>
+        /// Queues the action to occur on the UI thread and returns immediately.
+        /// </summary>
+        /// <param name="action">The action to be run on the UI thread</param>
+        /// <param name="state">The data to pass to <paramref name="action"/></param>.
+        internal static void OnUIThread<TState>(Action<TState> action, TState state = default)
```
tomspilman, line 170: @mrhelmut @harry-cpp

Note i added this so that some systems can make callbacks from worker thread that resolve on the main thread.

Specifically the callback for song ending comes from a native thread and we want the callback to occur from the main/UI thread.

src/monogame/common/MGM.cpp
tomspilman: > where possible they are played via the Platform's built-in media player and passed by URI.
> My understanding is this should provide integration (where possible)

@squarebananas  this is the problem.

"Where possible" is just two platforms.... Android and iOS.  No other console or platform has these concepts of unified music libraries.  It made sense in 2004... it makes no sense in 2024.  Is it going to integrate with Spotify?

The point is...  almost no games need these features.  Ones that do can code to Android/iOS/Spotify APIs directly.

99% of MG/XNA users just want to efficiently stream a song from content/files in the game. 

> Quite often I have seen developer's ask how to cross fade between 2 Songs or play additional layer tracks on top of each other.

I would like to see us support this functionality in the future, and we can if we don't tie ourselves to old platform media players.

This implementation is designed for that future.

> whereas SoundEffect fits this purpose well.

If you want to manually deal with streaming or punt and load the whole sound into memory.  Sure.

Most devs don't want to mess with streaming sounds.  Song i feel fits that roll.

> At a later point it might be useful to have a PlatformBehavior setting for playback preference (by device media player or SoundEffects).

I think this is would be a waste of MG development resources.  I've seen maybe 2 or 3 MG/XNA games in the last 10 years that wanted to integrate with the device media player... mostly music rhythm games.

Every other game just wants to stream music files.

src/monogame/common/MGM.cpp
```diff
@@ -0,0 +1,183 @@
...
+	// portable to all our target platforms:
+	//
+	// libtheora - free but limited quality/performance
+	// OpenH264 -
+	// ???
+	//
+	// It should then spin up thread which decodes the
+	// video/audio streams.
+
+	auto video = new MGM_Video();
```
tomspilman, line 119: Same as the design around `Song`.

In a separate bounty we need to integrate at least Theora.

But ideally we support H264 and a few other video formats.

src/monogame/common/MGM.cpp
squarebananas: Ok thanks, that all makes sense. I didn't realize newer Xbox consoles no longer have those features (used to use the Xbox 360 audio player a lot).

If the plan is to remove platform media players then there won't be any issues. Mainly I was just worried you might develop new shared Song code and then realize not all platforms currently use SoundEffects for this.

When I said "SoundEffect fits this purpose well" I was thinking in terms of the current implementations of the SoundEffect/SoundEffectInstance/DynamicSoundInstance system vs Song/MediaPlayer system. Making Song a simplified version of DynamicSoundInstance where it manages BufferNeeded events for the user sounds like a good idea.

tomspilman:
@mrhelmut  @harry-cpp - So everything is pretty much in place.

I want to get fullscreen working under SDL+VK to prove the API, but that is all that is left.

I decided against adding stubs for audio because we should have a bigger discussion on how we plan to use FAudio (in particular the XAct implementation).

I want to get final comments and get this merged by end of the week.

MonoGame.Framework/Platform/Native/Audio.Interop.cs
```diff
@@ -0,0 +1,131 @@
...
+internal readonly struct MGA_Buffer { }
+
+[MGHandle]
+internal readonly struct MGA_Voice { }
+
+
+/// <summary>
+/// MonoGame native calls for platform audio features.
+/// </summary>
+internal static unsafe partial class MGA
```
tomspilman, line 27: @mrhelmut @harry-cpp  

Because i needed it on Switch, i went ahead and created the native audio API.  This API is based on how the console APIs worked with a audio system, buffers, and voices.

Later we will have to figure out how FAudio will be integrated under this API but also move to using FAudio's Xact implementation.

src/monogame/include/csharp_GameWrapper.h
mrhelmut: ```api_``` or ```export_``` seems more fitting but ```csharp_``` is fine enough to me. Nitpicking here.

src/monogame/vulkan/MGG_Vulkan.cpp
mrhelmut: Let's figure if we need this post-merge based on user requests.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
mrhelmut: I think that we can consider gamepads anything that goes through SDL's game controller API, and anything that isn't is a joystick.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
mrhelmut: It would make sense to have them in Mouse.Native and GamePad.Native.

src/monogame/common/MGM.cpp
mrhelmut: Supporting H264/AAC seems like the best, anything else is bonus and not critical.

mrhelmut:
I don't have any critical comments on the API. To me it's good enough and we shouldn't comment it too much because it will likely move as we continue to implement platforms.

I need to test this out, but otherwise good to me.

tomspilman:
@mrhelmut @harry-cpp So i think this PR is done.

I added the sound API.  I fixed fullscreen mode switching.  Also fixed window icons.

It needs a review then we can move on to next tasks:

  - Fix/optimize the Vulkan implementation.
  - Add FAudio implementation.
  - Support Mac and Linux desktops with vulkan/sdl.
  - Move/share DX12 from Xbox to support Windows desktop too.
  - Add OpenGL native backend for legacy support.
  
Once Mac and Linux work we could consider dropping the old C# side opengl, openal, SharpDX 11, etc.

tomspilman:
So a few lingering questions i have on things i could fix up before merge:

- Are we sure we want `src` to be the top level folder for all the native code in the repo?  Why not something more informative like `native` or `native_code` or `cpp` or whatever else?

- I'm still unhappy with my acronyms for things like "MonoGame Platform" `MGP`, "MonoGame Graphics" `MGG`, "MonoGame Audio" `MGA`, etc.  We only have a few of these.  Any better ideas here?

- We didn't end up with a seperate interop API for controllers because it was considered part of the "platform" code.  This is mostly because all controller input comes thru the platform event system.  Should we change this?

build/BuildFrameworksTasks/BuildDesktopVKTask.cs
```diff
@@ -0,0 +1,34 @@
...
+    public override bool ShouldRun(BuildContext context) => context.IsRunningOnWindows();
+
+    public override void Run(BuildContext context)
+    {
+        // TODO: This should be moved to its own seperate
+        // build process so that it doesn't run on all MG builds.
+        {
+            var buildDir = "src/monogame/external/sdl2/sdl/build";
+            context.CreateDirectory(buildDir);
+            context.StartProcess("cmake", new ProcessSettings { WorkingDirectory = buildDir, Arguments = "-A x64 -D CMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded ../" });
```
dellis1972, line 17: Should we be also building for windows arm? `-A ARM` ?

mrhelmut:
> * Are we sure we want `src` to be the top level folder for all the native code in the repo?  Why not something more informative like `native` or `native_code` or `cpp` or whatever else?

I like the ```native``` or something similar wording. It feels more fitting to a C# framework. ```src``` is a wording very C/C++.

> * I'm still unhappy with my acronyms for things like "MonoGame Platform" `MGP`, "MonoGame Graphics" `MGG`, "MonoGame Audio" `MGA`, etc.  We only have a few of these.  Any better ideas here?

MG_Native_Platform, MG_Native_Graphics, MG_Native_Audio...? Dunno if it's too long to your liking. I personally like verbosity when it comes to top level naming conventions.

> * We didn't end up with a seperate interop API for controllers because it was considered part of the "platform" code.  This is mostly because all controller input comes thru the platform event system.  Should we change this?

I think it's fine to have them under platform code. It's almost always bound to it. Feels standard to me.

build/BuildFrameworksTasks/BuildDesktopVKTask.cs
mrhelmut: I'm really super skeptical with win-arm64, it has an extremely low adoption rate (below 0.01% if we look at the Steam stats), and I feel like it's just going to be extra work.

MonoGame.Framework/Platform/Native/Platform.Interop.cs
tomspilman: How the type generator works is it finds static classes in the `MonoGame.Interop.` namespace that have pinvokes.

So just making the `LibraryImport` members of the `Mouse` static class isn't enough to get them exported unless we change how these are found.  An attribute on the class maybe?

Thoughts @harry-cpp ?

Do we want/need more organization to the pinvokes than we have now?

tomspilman:
Ok found i needed `OcclusionQuery` still... working on that API.

I am skipping Microphone and Joystick still as they are not as critical.

.github/workflows/main.yml
```diff
@@ -21,12 +21,27 @@ jobs:
         with:
           dotnet-version: '8.0.x'
 
+      - name: Add msbuild to PATH
+        if: runner.os == 'Windows'
+        uses: microsoft/setup-msbuild@v1.0.2
```
dellis1972, line 26: why do we need this? can we not use `dotnet build` or `dotnet msbuild`?

tomspilman: That came from the SDL2 repo and enables building C++ projects.

If there is a better way to do it, i'm happy to change it.

@harry-cpp  ?

tomspilman:
@mrhelmut @harry-cpp @dellis1972 i think this bounty is done.  Last looks and lets get it merged and move on.

Do we need to put any information anywhere explaining that the native platform is WIP and not fully finished yet?

tomspilman:
@mrhelmut @harry-cpp @dellis1972 Going to push a few more fixes here (learning about flaws in the API from the Switch NativeAOT work).

But lets get this merged this week.  So please do your final reviews on it.

tomspilman:
@harry-cpp @mrhelmut @dellis1972

This PR and the bounty for it is done.  Let's get it merged!

dellis1972 (approved):

ThomasFOG (approved):

tomspilman:
Yay!

PR #8313, opened 2024-05-23
Re-enable 17.* wildcard for extension project

JakeLegendXIII:
Looks like Microsoft has been releasing the problem packages for the 17.10 set of VS tools in the last 24 hours (including 17.7.22 of the SDK analyzer from the original issue where it was looking for 17.7.20). The builds were failing again, because of another new package that should be there now.

Testing locally I am able to use the 17.* wildcards successfully, so basically resetting this to what it was before all of the broken packages. The GitHub action build is successful again as well.


mrhelmut:
Cool. Merging, thanks!

PR #8325, opened 2024-05-30
[Documentation] Update XML documentation for `GraphicsDevice.Legacy`

AristurtleDev:
## Description
This PR adds missing and updates existing XML documentation to the `GraphicsDevice.Legacy` class.

## Reference

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/GraphicsAdapter.Legacy.cs
+++ b/MonoGame.Framework/Platform/Graphics/GraphicsAdapter.Legacy.cs
@@ -19,6 +19,9 @@ using Android.Runtime;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
+    /// <summary>
+    /// Provides methods to retrieve and manipulate graphics adapters.
+    /// </summary>
     public sealed class GraphicsAdapter : IDisposable
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Platform/Graphics/GraphicsAdapter.Legacy.cs
```diff
@@ -59,11 +62,15 @@ internal GraphicsAdapter()
         }
 #endif
 
+        /// <inheritdoc />
```
Maniekko, line 65: I may be wrong, but empty inheritdoc on Dispose might not create a docline. There also should be `cref=IDisposable.Dispose` with it.

AristurtleDev: Only using `<inheritdoc />` correctly adds the documentation inheritance without needing to specify the `cref` attribute.

However, for clarity, if preferred to add the `cref` attribute, let me know
CC: @mrhelmut, @SimonDarksideJ

SimonDarksideJ: So long as the interface it is inheriting from is documented, then Inheritdoc is the recommended way to go.

PR #6285, opened 2018-04-15
Begin show keyboard uap xaml

madninjaskillz:
Add support for Guide.BeginShowKeyboard/Guide.EndShowKeyboard to UAP Xaml projects.

Have tried to ensure does not need any changes to existing projects, nor force a requirement for Framework.Net dll when you are not using Guide.

Style wise I have tried to keep the feel of 360 XNA whilst also using native design rules.

Supports standard and Password Boxes.

guide beginshowkeyboard2

PR #8348, opened 2024-06-05
Suppress warnings from UWP project

Nebulaxin:
UWP platform will be [dropped soon](https://github.com/MonoGame/MonoGame/pull/6285#issuecomment-2123244514), so warnings that exist only in that project can be suppressed


mrhelmut:
Fair enough.

PR #8352, opened 2024-06-06
[Documentation] Update XML documentation for `Utilities.Vector3`

theLaffa:
## Description
This PR adds missing and updates existing XML documentation to the `Utilities.Vector3` strut.


theLaffa:
Would guess this might not be required after #8350 is merged, feel free to reject if this is the case

mrhelmut:
It is indeed not required. Those types are only here for consoles targeting the legacy runtime (which is soon going to be replaced so those types will disappear with it).

PR #8353, opened 2024-06-06
[Documentation] Update XML documentation for `Utilities.Vector4`

theLaffa:
## Description
This PR adds missing and updates existing XML documentation to the `Utilities.Vector4` strut.

Would guess this might not be required after #8350 is merged, feel free to reject if this is the case


mrhelmut:
From my comment in @8353

> It is indeed not required. Those types are only here for consoles targeting the legacy runtime (which is soon going to be replaced so those types will disappear with it).

Issue #8194, opened 2024-02-22
[PAID BOUNTY] Switch the console runtime from BRUTE to NativeAOT


mrhelmut:
_The MonoGame Foundation wishes to incentivize the deprecation of the current console .NET runtime (BRUTE) in favor of switching to NativeAOT. This is a paid bounty._

📃**Full details**

The .NET console runtime of MonoGame needs to be changed to NativeAOT (publicly known as the ```PublishAot``` build option since .NET 7, see [documentation](https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=net7%2Cwindows)).

Related issue on the MonoGame issue tracker for more context: [#7873](https://github.com/MonoGame/MonoGame/issues/7873)

The MonoGame Foundation wishes to accelerate this transition by offering a paid opportunity to undertake this task.

Read this page to check how to apply and the requirements.

🟢 **Bounty status**

_Open for applications_. ⏳
To apply, please [read this page](https://monogame.net/bounties/).

💰 **Bounty**

$5.000,00 USD (Five thousands US dollars).

✋ **Requirements to apply**

_In addition to [our generic bounty requirements](https://monogame.net/bounties/#requirements), here are requirements specific to this bounty._

This bounty requires to be a registered developer with **all** of the console platforms (Sony, Microsoft, and Nintendo). We would like that this task is undertaken by a single person/entity, and you would therefore have to be cleared with all manufacturers. We need the implementation to have the same structure and user experience throughout all platforms, hence it would be more complex to manage if we were to accept per-platform applications. If need be, we can help with getting registered with either platforms.

📦 **Expected delivery content**

- All console platforms working with NativeAOT with no regression in regard to current MonoGame features (we can however expect runtime regressions since BRUTE and NativeAOT might have big differences).
- A smooth "build & run/debug" user experience, and if it can't be achieved a good documentation to get a project running with NativeAOT on consoles.

📅 **Delivery timeline**

As soon as possible.

📚 **Historical context**

Long story: [#7873](https://github.com/MonoGame/MonoGame/issues/7873)

Back when MonoGame was initially ported to game console hardware, the most recent .NET version was .NET 4.5, associated with C# 5.0. At that time, .NET was a closed source technology and there was no runtime available to run .NET applications on game consoles.

[...]

PR #8369, opened 2024-06-08
Update ConsoleCheck to .NET 8.0

Nebulaxin:
Right now shared code in MonoGame.Framework is limited to C# 5, making it impossible to use common features like string interpolation

This limitation, however, can be lifted; instead cheking against specific language version in ConsoleCheck project, it can be done for decompiled framework; decompiler will try to convert C# 6+ features to C# 5 ones (eg. string interpolation to string concatenation)


mrhelmut:
We can't update ConsoleCheck to newer .net, it's not just about the language version. It's also about the .net compiler _and_ the BCL version. It must be oldschool MSBUILD from older frameworks and limited to net452. This is because newer compilers won't produce IL code and format compatible with the legacy console runtime. This setup is the only way to make sure that the main repo doesn't break the console runtime.

But this will be irrelevant once #8194 is completed. Once the console runtime is entirely moved to NativeAOT, ConsoleCheck will be useless and removed from the repository (as well as the language and target framework limitation).

Nebulaxin:
Right now I am trying to remake commonly used features from newer languages and BCL like Span and interpolated strings

I made this pull request so almost all features from net8.0 can be used while #8194 and #8309 are currently in progress

mrhelmut:
It's really cool but I don't know if it's really worth doing since ConsoleCheck is going to disappear relatively soon and there is no impeding need to use C# 12 features in the framework codebase.

Nebulaxin:
Yes, there is no need to use C#12; this pr exists just so we can already use C#5+ features; I also made another pr with all the changes that can be made: https://github.com/Nebulaxin/MonoGame/pull/4

mrhelmut:
ConsoleCheck and UWP will be removed soon.

Side note if you intend to push your changes from your local fork: not a fan of upgrading code just for the sake of upgrading code (especially when things just work or when syntactic sugars make code harder to read/debug).

mrhelmut:
Closing this because #8194 will make ConsoleCheck irrelevant.

PR #8281, opened 2024-04-16
Applied the suggested changes for #6728.

itsBuggingMe:
Just added the suggested changes for #6728.

PR #8370, opened 2024-06-09
Applied the suggested changes for MonoGame#6728, and docs

itsBuggingMe:
Follow-up to #8281

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -84,6 +84,8 @@ namespace Microsoft.Xna.Framework.Graphics
         private IndexBuffer _indexBuffer;
         private bool _indexBufferDirty;
 
+        private VertexBufferBinding[] _vertexBufferBindings;
+
         private readonly RenderTargetBinding[] _currentRenderTargetBindings = new RenderTargetBinding[8];
         private int _currentRenderTargetCount;
         private readonly RenderTargetBinding[] _tempRenderTargetBinding = new RenderTargetBinding[1];
@@ -427,6 +429,7 @@ namespace Microsoft.Xna.Framework.Graphics
 
             // Force set the buffers and shaders on next ApplyState() call
             _vertexBuffers = new VertexBufferBindings(_maxVertexBufferSlots);
+            _vertexBufferBindings = new VertexBufferBinding[_maxVertexBufferSlots];
             _vertexBuffersDirty = true;
             _indexBufferDirty = true;
             _vertexShaderDirty = true;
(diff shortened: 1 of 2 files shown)


SimonDarksideJ:
Looks good to me, a second review @mrhelmut ?

dellis1972:
Can we get a more descriptive commit message/PR description if possible please?
I had to go through 3 different PR's to get to the reason as to why these API's are being added. I just makes it harder for people to see why something changed when going through the history.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -1089,8 +1092,49 @@ public void SetVertexBuffer(VertexBuffer vertexBuffer, int vertexOffset)
             _vertexBuffersDirty |= (vertexBuffer == null)
                                    ? _vertexBuffers.Clear()
                                    : _vertexBuffers.Set(vertexBuffer, vertexOffset);
+        }
+
+        /// <inheritdoc cref="SetVertexBuffers(VertexBufferBinding, VertexBufferBinding, VertexBufferBinding, VertexBufferBinding)"/>
+        /// <param name="vertexBuffer">The vertex buffer to set</param>
+        public void SetVertexBuffers(VertexBufferBinding vertexBuffer)
+        {
+            _vertexBufferBindings[0] = vertexBuffer;
```
Mindfulplays, line 1101: Clear the other elements? (i.e. 1, 2, 3) here and below? Also a prior call to `SetVB(a1, a2, a3)` should clean up after calling `_vertexBuffers.Set` otherwise it's not gonna be GC'ed?

MonoGame.Framework/Graphics/Vertices/VertexBufferBindings.cs
```diff
@@ -88,22 +88,33 @@ public bool Set(VertexBuffer vertexBuffer, int vertexOffset)
...
         /// Binds the the specified vertex buffers to the input slots.
         /// </summary>
         /// <param name="vertexBufferBindings">The vertex buffer bindings.</param>
+        /// <param name="bindingCount">Number of bindings from the array to set, starting from index 0.</param>
         /// <returns>
         /// <see langword="true"/> if the input layout was changed; otherwise,
         /// <see langword="false"/>.
         /// </returns>
-        public bool Set(params VertexBufferBinding[] vertexBufferBindings)
+        public bool Set(VertexBufferBinding[] vertexBufferBindings, int bindingCount)
```
Mindfulplays, line 108: May be add doc on the negative effects of passing a binding (and its effect on per-frame heap alloc).
IMO, I feel like the API user can do all this without introducing `_vertexBufferBindings[]` array at the `GraphicsDevice` level - you can maintain it yourself, can't you? But I realize you already made some `params` changes so will let others chime in here.

It's a good idea though, but could mostly live inside a doc comment as opposed to prescribing/adding complexity..

SimonDarksideJ:
After some internal conversation, could you add some Unit Tests to ensure these enhancements are covered in testing please @Mindfulplays

Mindfulplays:
> After some internal conversation, could you add some Unit Tests to ensure these enhancements are covered in testing please @Mindfulplays

I agree too, The original change + this one seems to add complexity that could really be avoided/pushed outside, but tests should help contain it imo. @itsBuggingMe

itsBuggingMe:
Is it possible to temporarily do what @Mindfulplays has suggested, and switch to span params once #8194 has completed and C# 13 is done.

itsBuggingMe:
It is not strictly necessary but would make the api more streamlined and convenient. Takes away another pitfall and one less thing for the user to think about. C# itself does this too to avoid allocations such as in `string.Concat()`.

PR #8384, opened 2024-06-14
Added pooling for Texture2D.PlatformGetData

### LINKED ISSUE #8304: Getdata<T> allocates an extra array of T[]
### Prerequisites

### MonoGame Version

3.8.1

### Operating System

Windows

### Description

Getdata<T> allocates an extra array of T[] GC runs reducing the frame rate

### Steps to Reproduce

```          Rectangle r = new Rectangle(X, Y, 1, 1);
            texture.GetData<Color>(0, r, TextureCheckbuffer, 0, 1);
```

### Minimal Example Repo

_No response_

### Expected Behavior

No extra allocations needed

### Resulting Behavior

Color[] is allocated each frame

### Files

_No response_

Nebulaxin:
This should fix #8304

### Contributor Declaration


MonoGame.Framework/Platform/Graphics/Texture2D.OpenGL.cs
```diff
@@ -251,7 +251,7 @@ private void PlatformGetData<T>(int level, int arraySlice, Rectangle rect, T[] d
                 // Note: for compressed format Format.GetSize() returns the size of a 4x4 block
                 var pixelToT = Format.GetSize() / tSizeInByte;
                 var tFullWidth = Math.Max(this.width >> level, 1) / 4 * pixelToT;
-                var temp = new T[Math.Max(this.height >> level, 1) / 4 * tFullWidth];
+                var temp = GetDataPool<T>.GetArray(Math.Max(this.height >> level, 1) / 4 * tFullWidth);
```
tomspilman, line 254: So this makes the code not thread safe...  but i suppose OpenGL never was?

@harry-cpp @mrhelmut ?

Nebulaxin: GetData and SetData should only execute on UI thread ([line 228](https://github.com/MonoGame/MonoGame/blob/b1244f114a08b103d97c8fcec3e38f27dec93f24/MonoGame.Framework/Platform/Graphics/Texture2D.OpenGL.cs#L228)), so we can use this non thread safe pooling

I also thought about using ArrayPool.Shared, but it seemed that will be an overkill for this pr

MonoGame.Framework/Platform/Graphics/Texture2D.OpenGL.cs
```diff
@@ -268,7 +268,7 @@ private void PlatformGetData<T>(int level, int arraySlice, Rectangle rect, T[] d
             {
                 // we need to convert from our format size to the size of T here
                 var tFullWidth = Math.Max(this.width >> level, 1) * Format.GetSize() / tSizeInByte;
-                var temp = new T[Math.Max(this.height >> level, 1) * tFullWidth];
+                var temp = GetDataPool<T>.GetArray(Math.Max(this.height >> level, 1) * tFullWidth);
```
Mindfulplays, line 271: Are we gonna always leave behind a stale largest texture ever allocated?
i.e. it never gets deallocated unless the size mismatches correct on a subsequent call? This could be in MBs (size of screen).

Usually in a pool, you rent and then return. I don't see the return part here. In fact, why not just use `ArrayPool` like you said - it's not overkill imo?

Nebulaxin: ArrayPool should be fine if `ArrayPool.Create()` is used with maxArrayLength = 4096 * 4096 (that should be enough) and maxArraysPerBucket = 1, since only one array will be used at a time

Nebulaxin: Also maybe `maxArrayLength` should be `int.MaxValue` instead of `4096 * 4096`, so array can be any length and still be saved to the pool

MonoGame.Framework/Platform/Graphics/Texture2D.OpenGL.cs
```diff
@@ -503,5 +506,10 @@ struct SetDataRectState<T>
             public static Action<SetDataRectState<T>> Action =
                 (s) => s.texture.PlatformSetDataBody(s.level, s.arraySlice, s.rect, s.data, s.startIndex, s.elementCount);
         }
+
+        static class GetDataPool<T> where T : struct
+        {
+            public static ArrayPool<T> pool = ArrayPool<T>.Create(int.MaxValue, 1);
```
Mindfulplays, line 512: Could this be `readonly` and named 'Pool` (capitalize?)... 
In general looks good - will need a maintainer to review this @mrhelmut ?.

Mindfulplays, line 512: Perhaps add a comment as to why `int.MaxValue` is being used? (I reckon the default 2^20 doesn't work well for our large texture use case?).

SimonDarksideJ:
Can you address the remaining review comments @Nebulaxin , as this looks good to merge with those changes

mrhelmut:
ArrayPool is not available on .NET 4. It's an issue as long as #8194 is open.

Peru-S:
Not sure if you saw @mrhelmut's comment about `ArrayPool` in BRUTE (only available when NativeAoT is). This looks good otherwise.

mrhelmut:
From a personal perspective and coding style, I don't think that we need a pool and that a static byte array that we grow if needed is likely enough.

squarebananas:
Just a few thoughts:

1) This change could also be made to the TextureCube class as well.
2) Currently GetDataPool is a static inside Texture2D. As each instance of a pool will increase overall memory usage, GetDataPool could be moved & shared for use by other applicable classes.
3) If GetData is performed on large resources of differing SurfaceFormats (if I'm understanding this correctly https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Buffers/SharedArrayPool.cs), higher overall memory usage may occur because a pool is setup for each SurfaceFormat type.

> From a personal perspective and coding style, I don't think that we need a pool and that a static byte array that we grow if needed is likely enough.

In terms of graphics planning, I believe one of the aims stated was to implement multithreaded graphics API support (such as drawing on one thread while using creating/set/get on others).

A static byte array could do this by implementing thread safe methods around it.

Looking at the `ArrayPool<T>` documentation though it says "this class is **_thread-safe_**. All members may be used by **_multiple threads concurrently_** (https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1?view=net-8.0).

So using `ArrayPool<T>` might simplify threading implementations and it also provide bucketing of similar array sizes along with a fast access tiered caching system (https://github.com/dotnet/runtime/blob/ad0c47c42e8901b9300701b237c42345b83c2bd9/src/libraries/System.Private.CoreLib/src/System/Buffers/SharedArrayPool.cs#L16).

For future consideration, I wonder if there would be any performance & memory allocation improvements if a single static `ArrayPool<byte>` was used across multiple class types?

mrhelmut:
> In terms of graphics planning, I believe one of the aims stated was to implement multithreaded graphics API support (such as drawing on one thread while using creating/set/get on others).

MonoGame isn't designed for that and currently uses graphics backends that are not friendly with that (especially OpenGL in this PR). Also, MonoGame already has thread-safety on those operations through ```Threading.BlockOnUIThread```, although this is not real multi-threading and will stop threads and run operations back to back on the UI thread and then continue (because of how limited OpenGL is in that regard).

ArrayPool doesn't add much here beside using modern API. Even if we were using it, it wouldn't resolve the non-thread-friendliness of OpenGL.

I'm pretty sure that a single scratch buffer can be used across all kind of texture and surface types.

squarebananas:
Thanks for the reply. Just to make it clearer, the first 3 numbered points relate to this PR while the rest relates to ArrayPool usage "in terms of graphics planning" considerations in general.

I do understand that MonoGame isn't currently designed for this at all and that Threading.BlockOnUIThread still performs the backend graphics calls from a single thread at the moment.

In terms of graphics planning "for future consideration" though (so not this PR or OpenGL), I was suggesting ArrayPool might still be worth considering for multithreading benefits. These might be realized when accelerating load/build times by loading graphics resources across multiple threads and when any future async methods might increase the number of temporary buffers that concurrently exist.

AristurtleDev:
The original issue this addressed #8304 was closed by the original person that opened it, and it looks like we need some clarity on the discussion here.

Specifically: should this proceed as a targeted performance enhancement or is there additional design and validations before we merged.

@Nebulaxin we would also need you to update your initial PR description to include the new statement regarding our LLM policy as shown in the updated PR templates here: https://github.com/MonoGame/MonoGame/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

```
### Contributor Declaration

```

PR #8396, opened 2024-06-23
Platform-specific launchers for MGCB Editor Build Task

Lutz89:
The current Nuget Build Task (develop) for the MGCB Editor does not create platform-specific launcher packages (e.g. mgcb-editor-windows), so the Nuget packages for Mac, Linux and Windows are missing. However, these are required for the bootstrap launcher and in the dotnet manifest files.

This leads to errors when creating the project and the editor cannot be opened.

### DIFF
--- a/build/BuildContext.cs
+++ b/build/BuildContext.cs
@@ -7,7 +7,8 @@ public enum ProjectType
     Tools,
     Templates,
     ContentPipeline,
-    MGCBEditor
+    MGCBEditor,
+    MGCBEditorLauncher
 }
 
 public class BuildContext : FrostingContext
@@ -126,6 +127,7 @@ public class BuildContext : FrostingContext
         ProjectType.Templates => $"Templates/{id}/{id}.csproj",
         ProjectType.ContentPipeline => "MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj",
         ProjectType.MGCBEditor => $"Tools/MonoGame.Content.Builder.Editor/MonoGame.Content.Builder.Editor.{id}.csproj",
+		ProjectType.MGCBEditorLauncher => $"Tools/MonoGame.Content.Builder.Editor.Launcher/MonoGame.Content.Builder.Editor.Launcher.{id}.csproj",
         _ => throw new ArgumentOutOfRangeException(nameof(type))
     };
 
--- a/build/BuildToolsTasks/BuildMGCBEditorTask.cs
+++ b/build/BuildToolsTasks/BuildMGCBEditorTask.cs
@@ -24,5 +24,6 @@ public sealed class BuildMGCBEditorTask : FrostingTask<BuildContext>
 
         context.DotNetPublish(context.GetProjectPath(ProjectType.MGCBEditor, platform), context.DotNetPublishSettings);
         context.DotNetPack(context.GetProjectPath(ProjectType.Tools, "MonoGame.Content.Builder.Editor.Launcher.Bootstrap"), context.DotNetPackSettings);
+		context.DotNetPack(context.GetProjectPath(ProjectType.MGCBEditorLauncher, platform), context.DotNetPackSettings);
     }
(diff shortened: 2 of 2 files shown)


dellis1972:
@Lutz89 thanks for the PR, do you have examples of the kinds of errors you see with this issue? It would be helpful to know and it would help others find this PR if they search for the errors.

SimonDarksideJ:
I suspect the errors mentioned @dellis1972 are more user-orientated, by forgetting to switch the platform when configuring multi-platform projects.  Although I fail to see why you would need to add another launcher command rather than just update the existing.

Is there a reason you had to create a new command for this @Lutz89 , plus can you provide more specifics how the errors you mention are discovered, just to prove/disprove my assumption.

Lutz89:
Sorry, I should add a little more context.

Before the build process was changed, there were various Nuget packages for the MGCB editor:
- dotnet-mgcb-editor
- dotnet-mgcb-editor-windows 
- dotnet-mgcb-editor-linux
- dotnet-mgcb-editor-mac 

Currently, however, only dotnet-mgcb-editor is packed and provided as a Nuget package. 

If, for example, I integrate the artifacts from the build process locally as Nuget repository or use https://nuget.pkg.github.com/Monogame/index.json as package source, the packages 
- dotnet-mgcb-editor-windows 
- dotnet-mgcb-editor-linux
- dotnet-mgcb-editor-mac
no longer exist for the current version.

However, dotnet-mgcb-editor is not a stand-alone editor, but merely a "redirect" to the specific launcher depending on the platform. However, since dotnet-mgcb-editor-windows and co are no longer provided, the editor does not start either.

So if I enter e.g. the develop version "3.8.1.977-develop" in the project and in the manifest dotnet-tools.json and recreate the project or just execute the command `dotnet tool restore`, I get the following output:
`repos\MGCB error\MGCB error>dotnet tool restore
"Version "3.8.1.977-develop" of the package "dotnet-mgcb-editor-linux"" was not found in NuGet feeds "https://api.nuget.org/v3/index.json;C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\;https://nuget.pkg.github.com/Monogame/index.json".`

For example, if I change the manifest file so that only the MGCB packages created are included, the build works at first, but when I try to start the editor, the next error appears:

For example, if I change the manifest file so that only the MGCB packages created are included, the build works at first, but when I try to start the editor, the next error appears because the "mgcb-editor" tool tries to bootstrap the specific tool like `dotnet-mgcb-editor-windows`

```
repos\MGCB Error\MGCB Error>dotnet mgcb-editor

repos\MGCB Error\MGCB Error>Execution was not possible because the specified command or file was not found.
Possible reasons for this are:
  * You have written an integrated dotnet command incorrectly.
  * You wanted to execute a .NET program, but dotnet-mgcb-editor-windows does not exist.
  * You wanted to run a global tool, but an executable file with dotnet-prefix and this name was not found in PATH.
```


Manifest Example with only existing packages:
`{
  "version": 1,
  "isRoot": true,
  "tools": {
    "dotnet-mgcb": {
      "version": "3.8.1.977-develop",
      "commands": [
        "mgcb"
      ]
    },
    "dotnet-mgcb-editor": {
      "version": "3.8.1.977-develop",
      "commands": [
        "mgcb-editor"
      ]
    }
  }
}`

You can clearly see in the logs of the Github actions that the platform-specific editor packages are not created and uploaded (see https://github.com/MonoGame/MonoGame/actions/runs/9549320147/job/26319576632)

Therefore, I have added another command which, in addition to the MGCB editor (Bootstrap), also builds the corresponding platform-specific package (e.g. mgcb-editor-windows).

Maybe there is another problem with the build process, then someone else should take a closer look at this, but this solution works and the platform-specific editor packages are built again.

Let me know if you need more information.

Thanks for your time

harry-cpp:
Ups, we did a build system update and it seems we forgot to reconnect the launcher projects, good find, merging!

PR #8401, opened 2024-06-27
Create PULL_REQUEST_TEMPLATE.md

dellis1972:
Add a new Template for Pull Requests. 
This is to hopefully encourage contributors to give descriptive messages in their PR's so that future contributors can understand the context in which changes were made.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,23 @@
+<!--
+Are you targeting develop? All PRs should target the develop branch unless otherwise noted.
+-->
+
+Fixes #
+
+<!--
+Please try to make sure that there is a bug logged for the issue being fixed if one is not present.
+Especially for issues which are complex. 
+The bug should describe the problem and how to reproduce it.
+-->
+
+### Description of Change
+
+<!-- 
+Enter description of the fix in this section.
+Please be as descriptive as possible, future contributors will need to know *why* these changes are being made.
+For inspiration review the commit/PR history in the MonoGame repository.
+-->
+
+
+
+


SimonDarksideJ (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+<!--
+Are you targeting develop? All PRs should target the develop branch unless otherwise noted.
```
AristurtleDev, line 2: Would recommend not including things like this as commented HTML sections.  They tend to bloat the PR especially when trying to write it or edit it, users will need to delete it out or accept the boilerplate that's there.

If you want to ensure something like the PRs are targeting the `develop` branch, it's might be better to make it an explicit task (checkbox) that they have to click so that they are verifying they did check it's targeting `develop`.

For instance, this is an example I use in PR templates. 

```
## Prerequisites
- [ ] I have verified that there are no existing pull requests that would overlap with this pull request.
- [ ] I have verified that I am following the guidelines as outlined in this project's contribution policy
- [ ] I Have verified that this pull request adheres to this project's code of conduct.
- [ ] I have written a descriptive title for this pull request.
- [ ] I have provided appropriate test coverage were applicable.
```

If users haven't checked all boxes, then the PR isn't even reviewed until they do. Obviously adjust to MonoGame specific requirements

dellis1972: This looks like a pretty comprehensive list and a good place to start 😄

mrhelmut: It feels a bit hardcore to me and might prevent contributions. We can likely find something a bit smoother. Like encouraging test coverage but making them optional. Like noting that PRs with tests will be reviewed in priority, and likewise for PRs that are not conflicting and respecting guidelines.

I prefer to encourage with faster reviews than to fence things off.

SimonDarksideJ (approved):

Issue #8406, opened 2024-07-03
Deprecation of the UWP platform/target


SimonDarksideJ:
### Intent

UWP is no longer a targetted/supported platform with Microsoft, and UWP submissions are no longer accepted on the Xbox Creators program for Xbox.

As per the roadmap, UWP is due to be deprecated for the 3.8.2 release.

### Motivation

Detailed above

mrhelmut:
Just noting that once UWP will be deprecated and replaced by the Direct3D 12 version of MonoGame, a few things will change in regard to publishing games on Xbox or the Microsoft Store:
- It will no longer be possible to run MonoGame apps on retail Xbox consoles using the dev mode
- It will no longer be possible to self-publish MonoGame apps on the Microsoft Store with just the Creators Program
- An ID@Xbox license will be required to publish games on the Microsoft Store (including on PC)

PR #8407, opened 2024-07-03
UWP Deprecation

SimonDarksideJ:
Feature Request #8406

- Removal of the UWP project templates
- Removal of UWP entries and implementations
- Removal of UAP entries and implementations
- Removal of WINRT entries and implementations


MonoGame.Framework/Utilities/PlatformInfo.cs
```diff
@@ -28,8 +28,6 @@ public static MonoGamePlatform MonoGamePlatform
                 return MonoGamePlatform.WebGL;
 #elif WINDOWS && DIRECTX
                 return MonoGamePlatform.Windows;
-#elif WINDOWS_UAP
-                return MonoGamePlatform.WindowsUniversal;
```
Mindfulplays, line 32: Note [MonoGamePlatform.cs](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Utilities/MonoGamePlatform.cs) still has the enum - remove that too?

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -406,12 +401,11 @@ protected virtual Stream OpenStream(string assetName)
 			{
 				throw new ContentLoadException("The content file was not found.", fileNotFound);
 			}
-#if !WINDOWS_UAP
+
```
Mindfulplays, line 404: Should the 'W' in line 45 be moved to legacy identifiers? Also the 'M' there :)

SimonDarksideJ: Thanks for the review, I did remove all uap "stuff" but will check when I'm back in the UK.

Mindfulplays:
LGTM, thank you.

Mindfulplays (approved):

Brett208:
Has anyone seen a timeline on when UWP will no longer be supported for Xbox? I didn't see an answer with a quick internet search.

We currently have 2 games published on XBOX using UWP. I just updated one this week and Microsoft was ok with the update.

If this PR is merged, anyone maintaining a UWP MonoGame project on XBOX would likely be stuck not accepting new updates to MonoGame. I don't know how much maintenance work would be required to keep it in place until UWP is further sunsetted. May be worth supporting existing published games for a while longer.

However, if removing it is needed to move MonoGame's larger priorities forward, I would understand.

Thank you.

-Brett

SimonDarksideJ:
Uwp is currently not supported for new submissions for Xbox, it was discontinued some time ago.

Existing titles can still be updated, but the sdk's and features are now fixed, so future updates to MonoGame are unlikely to be of use to them.

The only publisher path to Xbox AFAIK, is with an id@xbox agreement and the native build system, which is still available for MonoGame (and in the process of being updated for dx12 and gdk)

Hope that helps.

mrhelmut:
Please note that deprecating a platform doesn't mean that it will disappear from previous MG releases or that it would be impossible to build previous versions from source. Developers willing to keep using that target will still be able to, although with a bit more friction but that's how deprecation goes.

UWP is in a limbo state when it comes to gaming at Microsoft, there's not much info and nothing quite encouraging using it for gaming. Meanwhile, it is encouraged to move toward DirectX 12 targets. In our book this means that we shouldn't bet on UWP for anything future-proof.

UWP is not accepted anymore for new ID@Xbox titles. It still works for Creator Program titles, but those end up in a sub-store almost invisible to Xbox users, which gives little to no incentive to keep using UWP.

The reasoning of the deprecation is that there are almost no developer still using it, and the DirectX 12 targets will cover the same (and more) scenarios. So better maintaining one target that fits many, than multiple ones that fit almost none.

The one thing that the DirectX 12 target won't do, is allowing to side-load/debug apps on retail Xbox with devmode activated. It will require a dedicated Xbox devkit, and since getting into ID@Xbox is fairly accessible, it's not a big deal.

Hobbyist willing to play around with UWP and side-loading will still be able to do it with 3.8.1 or from source.

Brett208:
Thank you for the followup and detailed responses.

-Brett

dellis1972 (approved):

PR #8420, opened 2024-07-16
Upgrade Android to handle obsolete API's.

dellis1972:
As the underlying Android platform evolves they tend to obsolete many API's. We should be keeping on top of these so that the Android implementation stays current. However we cannot remove the old APi's completely due to the massive number of devices which Android supports.

So we need to make use of Runtime checks to make sure we call the correct "supported" API's on the correct versions of Android. Previously this had to be done by checking the `SdkInt` version. With .net 6 and beyond we can make use of the
`OperatingSystem.IsAndroidVersionAtLeast` utility method. This is a method which the compiler is aware of and it can hide the warnings about obsolete API's around the code it protects.

Also upgraded other obsolete API's which do not need runtime checks.

### DIFF
--- a/MonoGame.Framework/Media/Album.cs
+++ b/MonoGame.Framework/Media/Album.cs
@@ -10,6 +10,7 @@ using CoreGraphics;
 using MediaPlayer;
 using UIKit;
 #elif ANDROID
+using Android.Content;
 using Android.Graphics;
 using Android.Provider;
 #endif
--- a/MonoGame.Framework/Media/Album.cs
+++ b/MonoGame.Framework/Media/Album.cs
@@ -167,7 +167,13 @@ namespace Microsoft.Xna.Framework.Media
 #elif ANDROID
         public Bitmap GetAlbumArt(int width = 0, int height = 0)
         {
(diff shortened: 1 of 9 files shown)


MonoGame.Framework/Media/Album.cs
```diff
@@ -167,7 +168,13 @@ public UIImage GetAlbumArt(int width = 0, int height = 0)
 #elif ANDROID
         public Bitmap GetAlbumArt(int width = 0, int height = 0)
         {
-            var albumArt = MediaStore.Images.Media.GetBitmap(MediaLibrary.Context.ContentResolver, this.thumbnail);
+            Bitmap albumArt;
+            if (!OperatingSystem.IsAndroidVersionAtLeast (29)) {
```
dellis1972, line 172: I tried writing a wrapper method for `IsAndroidVersionAtLeast` which takes the `BuildVersionCodes`  enum. But doing that stopped the compiler from ignoring the obsolete warnings which is kind of the point of this whole exercise. So I decided to stick with the `OperatingSystem.IsAndroidVersionAtLeast` api.

SimonDarksideJ:
Is this PR now complete @dellis1972 ?

SimonDarksideJ (approved):

PR #8427, opened 2024-07-28
Add MouseKeys support - Windows

benda:
I noticed while using Myra that MouseKeys on Windows do not work. In other words, I can't use the numeric keypad to simulate mouse presses, which is important for usability. This PR fixes the MouseState to properly allow Myra to detect the keypress as a mouse press.

The way MouseKeys works is that Control.MouseButtons does not have the simulated mouse states, it shows the actual mouse states of no buttons pressed even though the user pressed the numeric keypad to simulate a mouse press. In  WndProc you do get WM_LBUTTONDOWN message, and the events MouseDown and MouseUp do get raised for the the simulated events. 

I am using Task.Delay to delay the mouse up event by 100 milliseconds so the simulated mouse pressed state can be queried. If that is an issue please let me know the preferred solution.

### DIFF
--- a/MonoGame.Framework/Platform/Windows/Accessibility/MouseKeysManager.cs
+++ b/MonoGame.Framework/Platform/Windows/Accessibility/MouseKeysManager.cs
@@ -1,0 +1,17 @@
+﻿using System.Runtime.InteropServices;
+
+namespace Microsoft.Xna.Framework.Windows.Accessibility
+{
+    internal class MouseKeysManager
+    {
+        internal static bool IsEnabled()
+        {
+            MouseKeys mouseKeys = new MouseKeys();
+            mouseKeys.cbSize = (uint) Marshal.SizeOf(mouseKeys);
+
+            NativeMethods.SystemParametersInfo(SPI.SPI_GETMOUSEKEYS,(uint) Marshal.SizeOf(mouseKeys), ref mouseKeys, SPIF.None);
+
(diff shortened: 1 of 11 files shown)


dellis1972:
This looks like a nice addition. Shame SDL does not seem to support this type of input afaik. So this would only be useful on the DX platform. The DesktopGL would no have it :(

benda:
Updated code to better handle user turning MouseKeys on/off while running.

MonoGame.Framework/Platform/Windows/Accessibility/MouseKeysManager.cs
realcoloride, line 1: Would it be possible to avoid creating a new MouseKeys instance everytime IsEnabled is called and store it in a static variable instead? 

(C# 12.0 code, feel free to translate to the supported C# version)
```cs
internal static MouseKeys MouseKeys = new();
```

benda: IsEnabled is only called when the initial game window is created or when a setting is changed via the Settings application. My point being it is very rarely called so I don't think it would be a performance issue (which is what I assume is your concern). Does that resolve your issue?

realcoloride:
Yes sure. Thank you for letting me know.

SimonDarksideJ (approved):

MonogameGuy:
The main issue is the consumption of the WM_LBUTTONDOWN and  WM_LBUTTONUP Messages before the native handler has a chance to process them.

PR #8436, opened 2024-08-12
Run the Basic ContentBuilder Tests on CI

dellis1972:
We are currently not running ANY tests on CI.
Lets correct this by running at least the content building and
processor tests on CI.

Certain content pipelines tests (Audio & Effects) have
been disabled because they fail or hang on CI at this time.

Current the tests run in x64 dotnet so that we can pass the 
TextureCompression tests.
This PR #8456 should fix the texture compression issues so we do not
need x64 dotnet.

The Effect compilation for Linux /Mac needs to happen
in wine. But dotnet 8 does not seem to run under wine on CI at this
time. So that will need to be looked at.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,24 +25,76 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'microsoft'
-          java-version: '11'
+          java-version: '17'
 
       - name: Disable annotations
         run: echo "::remove-matcher owner=csc::"
 
+      - name: install wine64 on linux
+        run: |
+          sudo apt install p7zip-full curl
+          sudo dpkg --add-architecture i386 
+          sudo mkdir -pm755 /etc/apt/keyrings
+          sudo wget -O /etc/apt/keyrings/winehq-archive.key https://dl.winehq.org/wine-builds/winehq.key
+          sudo wget -NP /etc/apt/sources.list.d/ https://dl.winehq.org/wine-builds/ubuntu/dists/jammy/winehq-jammy.sources
+          sudo apt update && sudo apt install --install-recommends winehq-stable
+        if: runner.os == 'Linux'
+
+      - name: Install Arial Font
+        run: |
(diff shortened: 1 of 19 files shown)


mrhelmut:
I would expect all but audio and graphical tests to run.

Just for reference:

Ideally, we would be running all of the tests on the standard Github Actions runners.
The problem is that they are headless, which means that they have no support for Direct3D or OpenGL, and the absence of audio devices makes audio tests to fail too.

We managed to have the graphical tests running with mesaGL, but the accuracy is so off that the similarity tests are failing by quite a large margin (although we don't know if it's really an accuracy issue or something else).

Audio is a real problem though. We could be using a similar virtual device software but so far we haven't found a suitable one that is easy to install on a Github runner.

Tools/MonoGame.Tools.Tests/ContentCompilerTest.cs
```diff
@@ -17,9 +17,6 @@
 
 namespace MonoGame.Tests.ContentPipeline
 {
-#if MACOS
-    [Ignore("Hanging on Mac in CI?")]
-#endif
```
mrhelmut: Dunno if it's related, but there are some tests which won't run on threads and will just block. As far as I have looked into this, there is not system to ensure that tests are running on the main thread with the current test runner (NUnit).

dellis1972:
> I would expect all but audio and graphical tests to run.
> 
> Just for reference:
> 
> Ideally, we would be running all of the tests on the standard Github Actions runners. The problem is that they are headless, which means that they have no support for Direct3D or OpenGL, and the absence of audio devices makes audio tests to fail too.
> 
> We managed to have the graphical tests running with mesaGL, but the accuracy is so off that the similarity tests are failing by quite a large margin (although we don't know if it's really an accuracy issue or something else).
> 
> Audio is a real problem though. We could be using a similar virtual device software but so far we haven't found a suitable one that is easy to install on a Github runner.

I am aware of the Audio and Graphics issues ( I was the one who suggested using mesa :P) . This PR is mostly about getting some of the content pipeline tests running since we currently have nothing.

PR #8439, opened 2024-08-15
Change to support auto-publishing nugets on dispatch trigger.

CartBlanche:


.github/workflows/main.yml
```diff
@@ -1,15 +1,29 @@
...
 
 jobs:
   build:
     name: build-${{ matrix.os }}
+
     runs-on: ${{ matrix.os }}
+
     strategy:
       matrix:
-        os: [windows-latest, macos-latest, ubuntu-latest]
```
harry-cpp, line 11: I assume macos war removed by accident.

.github/workflows/main.yml
```diff
@@ -42,37 +56,76 @@ jobs:
...
-
-      - name: Expose GitHub Runtime
-        uses: crazy-max/ghaction-github-runtime@v3
+        run: dotnet run --project build/Build.csproj -- --target=Default         
 
-      - name: Upload Artifacts
-        run: dotnet run --project build/Build.csproj -- --target=UploadArtifacts
-        env:
-          ACTIONS_RUNTIME_TOKEN: ${{ env.ACTIONS_RUNTIME_TOKEN }}
-          ACTIONS_RUNTIME_URL: "${{ env.ACTIONS_RUNTIME_URL }}"
```
harry-cpp, line 54: This is uploading the build artifacts to the build dashboard, don't remove it.

.github/workflows/main.yml
```diff
@@ -42,37 +56,76 @@ jobs:
...
   deploy:
-      name: Deploy
+      name: build-macos-latest and Optionally Publish Nugets
       needs: [ build ]
-      runs-on: ubuntu-latest
```
harry-cpp, line 58: Always use Linux agents if you can.

.github/workflows/main.yml
```diff
@@ -42,37 +56,76 @@ jobs:
...
+        
+        - if: ${{ github.ref != 'refs/heads/main' }}
+          name: Preview Build
+          run: dotnet run --project build/Build.csproj -- --target=Default --nuget-target=preview
 
-          - name: Expose GitHub Runtime
-            uses: crazy-max/ghaction-github-runtime@v3
+        - if: ${{ github.ref == 'refs/heads/main' }}
+          name: Release Build
+          run: dotnet run --project build/Build.csproj -- --target=Default
```
harry-cpp, line 115: The deploy action should do no building, its there to download artifacts from the build dashboard and upload them where its needed.

harry-cpp (changes requested):

CartBlanche:
@harry-cpp could you take a look at this deprecation error from the last CI here - https://github.com/MonoGame/MonoGame/actions/runs/10451368955

I couldn't find where we can update the upload call to v4, so thought you might know.

harry-cpp:
The warning is probably coming from CAKE, ignore it for now.

build/DeployTasks/DeployNuGetsToMarketplaceTask.cs
```diff
@@ -0,0 +1,19 @@
...
+public sealed class DeployNuGetsToMarketplace : FrostingTask<BuildContext>
+{
+    public override bool ShouldRun(BuildContext context) => context.BuildSystem().IsRunningOnGitHubActions;
+
+    public override void Run(BuildContext context)
+    {
+        var repositoryOwner = context.GitHubActions().Environment.Workflow.RepositoryOwner;
+        context.DotNetNuGetPush($"nugets/*.nupkg", new()
+        {
+            ApiKey = context.EnvironmentVariable("GITHUB_TOKEN"),
```
harry-cpp, line 15: Make a separate env variable.

.github/workflows/main.yml
```diff
@@ -76,3 +103,8 @@ jobs:
             run: dotnet run --project build/Build.csproj -- --target=DeployNuGetsToGithub
             env:
               GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+          - name: Push Marketplace Nugets
+            run: dotnet run --project build/Build.csproj -- --target=DeployNuGetsToMarketplace
+            env:
+              GITHUB_TOKEN: ${{ secrets.NUGET_API_KEY }}
```
harry-cpp, line 110: Make a deploy task in CAKE thats combined of multiple tasks like so:
- https://github.com/MonoGame/MonoGame/blob/develop/build/Tasks.cs#L4

Then add prerequirements for running both of those tasks on cake side like so:
- https://github.com/MonoGame/MonoGame/blob/develop/build/BuildFrameworksTasks/BuildAndroidTask.cs#L7

harry-cpp (changes requested):

PR #8447, opened 2024-08-19
[CI] Improve versioning and allow upload to nuget.org

harry-cpp:
Stuff done:
- Added deploy task for nuget.org
- Replaced the hardcoded version with loading version from the latest tag
- Automatically use the current tag as version if building from GitHub workloads
- Automatically make a release when tag is pushed
- Fully use the tags version when its created:
  - Non master branches can be tagged
  - Tags should be specified with 4 numbers, first 3 being our release, 4th being our attempt at releasing that one, ie. `3.8.0.1` and `3.8.0.0-rc1`

### DIFF
--- a/build/BuildContext.cs
+++ b/build/BuildContext.cs
@@ -47,8 +47,8 @@ public class BuildContext : FrostingContext
             }
             else if (workflow.RefType == GitHubActionsRefType.Tag)
             {
-                var baseVersion = workflow.RefName;
-                if (!Regex.IsMatch(baseVersion, @"v\d+.\d+.\d+"))
+                var baseVersion = workflow.RefName.Split('/')[^1];
+                if (!VersionRegex.IsMatch(baseVersion))
                     throw new Exception($"Invalid tag: {baseVersion}");
                 
                 VersionBase = baseVersion[1..];
--- a/build/BuildContext.cs
+++ b/build/BuildContext.cs
@@ -52,7 +52,7 @@ public class BuildContext : FrostingContext
                     throw new Exception($"Invalid tag: {baseVersion}");
                 
                 VersionBase = baseVersion[1..];
-                Version = $"{VersionBase}.{workflow.RunNumber}";
+                Version = VersionBase;
             }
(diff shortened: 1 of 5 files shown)


build/BuildContext.cs
```diff
@@ -13,35 +16,61 @@ public enum ProjectType
...
+
             if (workflow.Repository != "MonoGame/MonoGame")
             {
-                Version = $"{DefaultBaseVersion}.{workflow.RunNumber}-{workflow.RepositoryOwner}";
+                Version = $"{VersionBase}.{workflow.RunNumber}-{workflow.RepositoryOwner}";
             }
             else if (workflow.RefType == GitHubActionsRefType.Branch && workflow.RefName != "refs/heads/master")
             {
-                Version = $"{DefaultBaseVersion}.{workflow.RunNumber}-develop";
+                Version = $"{VersionBase}.{workflow.RunNumber}-develop";
```
CartBlanche, line 63: the `develop` shouldn't be hard-coded and needs to the parameterised, so we can swap that to `preview` when build from non `master` branch for the preview release.

CartBlanche (changes requested):
As per added comments.

build/BuildContext.cs
```diff
@@ -13,35 +16,61 @@ public enum ProjectType
...
             }
             else
             {
-                Version = $"{DefaultBaseVersion}.{workflow.RunNumber}";
+                Version = $"{VersionBase}.{workflow.RunNumber}";
             }
         }
+        else
+        {
+            Version = context.Argument("build-version", VersionBase + ".1-develop");
```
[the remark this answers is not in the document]
harry-cpp: This is just a default value if no build-version is specified.

build/BuildContext.cs
harry-cpp: I've modified the code so that if non master branch is tagged, it will pick up version from the tag fully.

CartBlanche: OK LGTM

build/BuildContext.cs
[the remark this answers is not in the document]
CartBlanche: OK.

PR #8448, opened 2024-08-20
Fix edgecase where global tool may not be prefixed with `dotnet-`

Emik03:
### Problem
When running `dotnet tool install dotnet-mgcb-editor -g`, it can sometimes install the binary as `mgcb-editor` which causes the command `dotnet mgcb-editor` to fail since it doesn't have the required `dotnet-` prefix.
```
❯ mgcb-editor

~ 
❯ Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet-mgcb-editor-linux does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
  ```

### Solution
This PR fixes that by first checking if the command without the `dotnet-` prefix resides in `PATH`, and invokes it directly if found. Otherwise, it uses the current implementation. I hope I've done everything correctly.

### Workarounds
Creating a symlink `dotnet-mgcb-editor-<OS>` that points to `mgcb-editor-<OS>` is a temporary solution.
```sh
# Windows
mklink "%DOTNET_ROOT%/tools/mgcb-editor-windows" "%DOTNET_ROOT%/tools/dotnet-mgcb-editor-windows"
```
```sh
# Mac
ln -s "$DOTNET_ROOT/tools/mgcb-editor-mac" "$DOTNET_ROOT/tools/dotnet-mgcb-editor-mac"
```
```sh
# Linux
ln -s "$DOTNET_ROOT/tools/mgcb-editor-linux" "$DOTNET_ROOT/tools/dotnet-mgcb-editor-linux"
```

### DIFF
--- a/Tools/MonoGame.Content.Builder.Editor.Launcher.Bootstrap/Program.cs
+++ b/Tools/MonoGame.Content.Builder.Editor.Launcher.Bootstrap/Program.cs
@@ -17,7 +17,18 @@ namespace MonoGame.Content.Builder.Editor.Launcher.Bootstrap
                 RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? "mgcb-editor-linux" :
                 throw new NotImplementedException("Unsupported Operating System");
 
-            Process.Start(executable, $"\"{string.Join("\" \"", args)}\"");
+            // We invoke the tool directly without the `dotnet-` if it exists.
+            // See: https://github.com/MonoGame/MonoGame/pulls/8448.
+            var isInPath = Environment
+                .GetEnvironmentVariable("PATH")
+                .Split(Path.PathSeparator)
+                .Select(x => Path.Combine(x, executable))
+                .Any(File.Exists);
+
+            if (isInPath)
+                Process.Start(executable, $"\"{string.Join("\" \"", args)}\"");
+            else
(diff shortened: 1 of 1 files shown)


harry-cpp:
The tool name is `mgcb-editor` and if `dotnet mgcb-editor` is not working, then there is an error somewhere else:
<https://github.com/MonoGame/MonoGame/blob/develop/Tools/MonoGame.Content.Builder.Editor.Launcher.Bootstrap/MonoGame.Content.Builder.Editor.Launcher.Bootstrap.csproj#L10>

The reason the nuget name is `dotnet-mgcb-editor` is because that was the requirement back when the tool was originally created, the package name will be changed with some upgrades that are coming to the content pipeline system.

Emik03:
Ah, I didn't spot that within the project file, my bad. I just assumed it was something to do with the `dotnet` binary itself stripping the `dotnet-` prefix from the tool name itself.

Emik03:
That being said, should I make a separate PR for that or shall I leave that work to someone else?

harry-cpp:
I'll handle it very soon, need one more thing to submit before hand 😄

Issue #8450, opened 2024-08-21
To do: remove Google Stadia from supported platforms


mrhelmut:
Since the platform is discontinued, we need to remove references to Google Stadia: https://github.com/search?q=repo%3AMonoGame%2FMonoGame%20Stadia&type=code

We also need to remove a couple of references in the documentation repository: https://github.com/search?q=repo%3AMonoGame%2Fdocs.monogame.github.io%20stadia&type=code

PR #8452, opened 2024-08-21
Remove references to google stadia platform (#8450)

Dehayat:
Removed references to Google Stadia as listed in #8450


mrhelmut:
Thank you!

PR #8455, opened 2024-08-21
Updating links to include getting_started subdomain

JakeLegendXIII:
Fixing some 404s when linking to docs that have the getting_started subdomain now.


SimonDarksideJ (approved):
Great catch!

Issue #8419, opened 2024-07-12
[PAID BOUNTY] Upgrade MonoGame to use BasisUniversal for cross platform Texture Compression


dellis1972:
MonoGame's texture compression needs to be updated to support arm based machines.

📃**Full details**

MonoGame currently uses many native libraries to handle texture compression (nvtt, libpvr). Unfortunately these libraries are out of date and are tied to the x86/x86_64 CPU's because they use SIMD. As a result these libraries do not work on arm based chipsets like the Apple M1/M2. 

In order to fix the developer experience we need to switch out the texture compression code to use [Basis Universal](https://github.com/BinomialLLC/basis_universal). Which is a cross platform texture compression tool. The MonoGame build of this library is [here](https://github.com/MonoGame/MonoGame.Tool.BasisUniversal).

This new tooling is a command line tool. It will generate an intermediate file from which the compressed formats can be created. We will also need to shell out to the new `basisu` tool twice. Once to generate the intermediate file for the source image and once to generate the compressed data. Note the intermediate file can be used to generate many different compression types so it might not need to be called twiece every time.

This bounty is important as it will improve the stability and usability of the content pipeline on Mac and Linux. 

🔒 Bounty status

Assigned to @cdhanna
In Progress https://github.com/MonoGame/MonoGame/pull/8456/

💰 **Bounty**

$2000,00 USD (Two thousand US dollars).

✋ **Requirements to apply**

Standard requirements from [our generic bounty requirements](https://monogame.net/bounties/#requirements).
Having access to a Mac and Linux machines to aide testing is important.

📦 **Expected delivery content**

- Update MonoGame.Framework.Content.Pipeline
   - Remove all references to Nvidia.TextureTools.dll, libnv*.so/dll/dylib
   - Remove all References to ATI.TextureConverter.dll
   - Remove all Reference to PVRTexLibNET.dll and libPVRTexLibWrapper.so
   - Add references to the new MonoGame.Tool.BasisUniversal and make sure the `basisu` tool is being deployed along side the content pipeline assembly. Similar to how ffmpeg works.
   -  Create a central utility class through which the `basisu` tool can be called to generated the compressed textures. Note this should make use of the existing `ExternalTool.Run` method.
[...]

PR #8456, opened 2024-08-22
[Paid Bounty] Upgrade MonoGame to use BasisUniversal for cross platform Texture Compression

cdhanna:
To briefly restate the paid bounty, MonoGame's texture compression tooling does not work on ARM based CPUs (like the recent Apple M) chips. The goal of the bounty was to restore functionality on ARM based CPUs. The actual implementation detail of the bounty was to swap out old texture compression tools (like PVRTexLibNET and NVTT) with a new command-line utility called [Basis Universal](https://github.com/BinomialLLC/basis_universal). This PR completes the cross-platform goal of the bounty, and introduces some implementation _beyond_ what was originally documented in the paid bounty. 

Here is a quick tldr of what is in the PR, and more details below.
- Content Pipeline Context
- ETC2 Support
- ATSTC Support
- Dotnet tool references to `mgcb-basisu` and `mgcb-crunch`
- Nuget library references `ktxLoader` and `bcnEncoder`
- Removal of PVRTex and NVTT

To begin, I ripped out the PVRTex and NVTT libraries. Immediately, the following classes (and in some cases, their sub-classes) had compile errors, 
- `DxtBitmapContent`, 
- `AtcBitmapContent`, 
- `Etc1BitmapContent`, 
- `PvrtcBitmapContent`

From those 4 base types, _several_ `SurfaceFormat` types were no longer compressible. However, at this point, I find it helpful to remember that not all `SurfaceFormat` values were actually selectable by users. In the Content Builder, the users select a `TextureProcessorOutputFormat` value which _hints_ which `SurfaceFormat` to use, but doesn't strictly enforce it. For example, the user may select `TextureProcessorOutputFormat.DxtCompressed`, but exactly which DXT `SurfaceFormat` is selected is left to the Content Builder code (specifically, the `GraphicsUtil.CompressDxt` function)

There are a cascade of `SurfaceFormat`s that broke, and unfortunately, _Basis Universal_ doesn't actually cover all of the formats. _BasisU_'s format list is [here](https://github.com/BinomialLLC/basis_universal/blob/ad9386a4a1cf2a248f7bbd45f543a7448db15267/transcoder/basisu_transcoder.h#L49). 

To cover the formats that weren't supported directly by _BasisU_, this PR also includes the external command line utility, [Crunch](https://github.com/MonoGame/MonoGame.Tool.Crunch), as well as a nuget library for [BcnEncoder](https://github.com/Nominom/BCnEncoder.NET). 

Here is a table of all the formats that broke after I removed the old texture compression tools, if they're still supported, and if so, what tool is doing the compression. 

| Format | Still Supported | Tool | Note |
| - | - | - | - |
| DXT1 | yes | BcnEncoder | | 
| DXT1a | yes | BcnEncoder | | 
| DXT3 | yes | Crunch | BcnEncoder produces bad results for textures with alpha for dxt3 only | 
| DXT5 | yes | BcnEncoder | | 
| RgbaAtcExplicitAlpha | yes | BcnEncoder | | 
| RgbaAtcInterpolatedAlpha | yes | BcnEncoder | | 
| RgbaAtcInterpolatedAlpha | yes | BcnEncoder | | 
| RgbEtc1 | yes | Crunch | | 
| ETC2 | yes | Crunch | this PR adds support for ETC2 compression, which previously was not available | 
| ASTC_4x4_Rgba | yes | BasisU | this PR adds support for ASTC compression, as well as the SurfaceFormat value | 
| RgbPvrtc4Bpp | yes | BasisU | | 
| RgbaPvrtc4Bpp | yes | BasisU | | 
| RgbPvrtc2Bpp | no | - | | 
| RgbaPvrtc2Bpp | no | - | | 

Alarm bells may be going off in your brain for the lack of `RgbPvrtc2Bpp` and `RgbaPvrtc2Bpp` support. However, the correlated subtypes of `PvrtcBitmapContent` for those 2Bpp variants, `PvrtcRgb2BitmapContent`, and `PvrtcRgba2BitmapContent` had _no_ usages in the codebase. The `TextureProcessorOutputFormat.PvrCompressed` option would always lead users to the _4Bpp_ variant.  This PR marks the 2Bpp variants with the `[Obsolete]` tag but does not delete them. _BasisU_ does not support the 2Bpp variant, but since the types were never used, I felt it was okay to deprecate them. 

It turns out that _BasisU_ is actually only used for a _relatively_ small subset of the formats. The _BcnEncoder_ library is handling several formats. The library is much faster an transcoding the textures because it is running within the same dotnet process and does not need to generate any intermediate files. The _BcnEncoder_ library is a nuget package, and comes with a few dependencies. Comparatively, the _Crunch_ and _BasisU_ tools are invoking separate processes on the machine to handle the compression, and those processes require data to be passed via file, which requires the Content Builder to do file reads&writes. In both cases, I'm using another library, _KtxLoader_, to read the compressed byte arrays. 

The file requirement is the reason for the newly added `ContextScopeFactory` in this PR. The current architecture is set up so that the compression is triggered from within a `BitmapContent`'s public API, but since those types are originally XNA types, I cannot change their public API to include any contextual information about where to generate the intermediate files. Instead of modifying the public API, I've introduce a new concept into the Content Builder, called a _Context Scope_. Essentially, every time some content is being processed, it creates a `static`-ally available view of the contextual information, which is used to infer the directory where intermediate files should be created/destroyed.  

As noted in the table above, this PR adds support for `ETC2`. I've deprecated the `TextureProcessorOutputFormat.Etc1Compressed` option, and added a more generalized `TextureProcessorOutputFormat.EtcCompressed` option. This new option will use ETC2 compression if the source texture is found to have non opaque alpha values. Otherwise, ETC1 is still used. 

Also, this PR adds support for ASTC. To support this, I created a brand new entry in the `TextureProcessorOutputFormat` enum, and the `SurfaceFormat` enum. 

Basis Universal was packaged into a custom [monogame tool](https://github.com/MonoGame/MonoGame.Tool.BasisUniversal), and published to [nuget](https://www.nuget.org/packages/mgcb-crunch). 

The original paid bounty issue can be found here, 
https://github.com/MonoGame/MonoGame/issues/8419

And I had my draft PR against my personal fork of MonoGame. A fair amount of development discussion took place on that thread, so feel free to review it.
https://github.com/cdhanna/MonoGame/pull/1

Known Issues
- MultiThreaded context issue?
   - I need to do a bit digging into the `ContextScopeFactory` and make sure it works in a possibly multi-threaded scenario. I was testing it via Unit Tests. 
- How do the users get access to the dotnet tools `mgcb-crunch` and `mgcb-basisu`?
   - For development, there is a local `.config/dotnet-tools.json` file which makes the tools available to use during Unit Tests. But its unclear to me how these tools get resolved in an actual production use case. I think there is a plan amongst the maintainers of MonoGame, but I don't fully understand it (yet). 
   
   
  ----

I've enjoyed working on this, and I've learned _a lot_. Thanks to all those who have helped out and answered my questions :)


dellis1972:
@cdhanna Thanks for your work. Will start reviewing this asap.

MonoGame.Framework.Content.Pipeline/ContextScopeFactory.cs
```diff
@@ -0,0 +1,196 @@
...
+    ///
+    /// <para>
+    /// The most recent content operation can be retrieved with the <see cref="ActiveContext"/>
+    /// property.
+    /// </para>
+    ///
+    /// </summary>
+    public static class ContextScopeFactory
+    {
+        private static List<IContentContext> _contextStack = new List<IContentContext>(1);
```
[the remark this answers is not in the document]
cdhanna: Yeah, just wanted to say I'm aware of an issue here. 
I wrote this unit test, and it is failing due to the contexts being wrong. If we are multi-threading the content pipeline, this is a huge issue. I'll get a fix committed asap. 
```csharp

[Test]
public async Task ContextTesting()
{
    var tasks = new List<Task>();
    long errorCount = 0;
    for (var i = 0; i < 100; i++)
    {
        var capturedIndex = i;
        var task = Task.Run(async () =>
        {
            try
            {
                // spawn a context to be used
                using var ctx = new NoOpContext(capturedIndex);
                ContextScopeFactory.BeginContext(ctx);

                // wait some "arbitrary" amount of time
                await Task.Delay(10);

                // get the context and make sure it is the same as the one we expect
                var retrievedCtx = ContextScopeFactory.ActiveContext;
                Assert.That(ctx, Is.EqualTo(retrievedCtx));
            }
            catch (Exception)
            {
                Interlocked.Increment(ref errorCount);
            }
        });
        tasks.Add(task);
    }

    await Task.WhenAll(tasks);
    Assert.That(errorCount, Is.EqualTo(0));
}

class NoOpContext : ContextScopeFactory.ContextScope
{
    public readonly int Id;
    public override string IntermediateDirectory { get; }
    public override ContentBuildLogger Logger { get; }
    public override ContentIdentity SourceIdentity { get; }

    public NoOpContext(int id)
    {
        Id = id;
    }

    public override string ToString() => "test-ctx-" + Id;
}
```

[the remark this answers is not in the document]
cdhanna: Okay, I've added the unit test (and a second one), and modified the `ContextScopeFactory` to use `AsyncLocal<T>`. I understand this may be a controversial choice, because `AsyncLocal<T>` can be confusing to reason about. However, in this case, I think the usage is warranted. If multiple tasks for forked to do an content-import per task, then each task _may_ now have its own context.

MonoGame.Framework.Content.Pipeline/Processors/TextureProcessorOutputFormat.cs
```diff
@@ -2,6 +2,8 @@
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
+using System;
+
 namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
 {
     /// <summary>
```
tomspilman, line 9: Maybe the comment for the `Compressed` case below should be updated?

```
 /// <summary>
        /// The SurfaceFormat value, of the input TextureContent object, is converted to an appropriate compressed format for the target platform.
        /// This can include PVRTC for iOS, DXT for desktop, Windows 8 and Windows Phone 8, and ETC1 or BGRA4444 for Android.
        /// </summary>
```

MonoGame.Framework.Content.Pipeline/Utilities/BcnHelpers.cs
```diff
@@ -0,0 +1,61 @@
...
+using KtxSharp;
+using Microsoft.Xna.Framework.Content.Pipeline.Graphics;
+
+namespace Microsoft.Xna.Framework.Content.Pipeline.Utilities;
+
+/// <summary>
+/// The <see cref="Encode"/> method uses the https://github.com/Nominom/BCnEncoder.NET/tree/master
+/// library to encode various block based compression formats
+/// </summary>
+public static class BcnUtil
```
tomspilman, line 18: Should this be internal and not public?

Do we want to make this part of the public API?

@cdhanna @harry-cpp @mrhelmut ?

dellis1972: I think it should be internal , that is the same for the other helpers which have been added.

.config/dotnet-tools.json
```diff
@@ -0,0 +1,20 @@
...
+        "mgcb-basisu"
+      ],
+      "rollForward": false
+    },
+    "mgcb-crunch": {
+      "version": "1.0.4.2",
+      "commands": [
+        "mgcb-crunch"
+      ],
+      "rollForward": false
```
dellis1972, line 17: @harry-cpp should we be rolling forward on these automatically?

MonoGame.Framework.Content.Pipeline/ContextScopeFactory.cs
```diff
@@ -0,0 +1,209 @@
...
+    /// when the operation has concluded.
+    /// </para>
+    ///
+    /// <para>
+    /// The most recent content operation can be retrieved with the <see cref="ActiveContext"/>
+    /// property.
+    /// </para>
+    ///
+    /// </summary>
+    public static class ContextScopeFactory
```
dellis1972, line 53: does this need to be public or internal (I know the tests need it , but we can use InternalsVisibleTo as we do for others) same for the `interface`

MonoGame.Framework/Graphics/SurfaceFormat.cs
```diff
@@ -158,31 +158,35 @@ public enum SurfaceFormat
...
+		/// </summary>
+        Rgba8Etc2 = 94,
+		/// <summary>
+        /// Etc2 SRGB8A8 EAC (Android/iOS withh OpenglES 3.0)
+		/// </summary>
+        SRgb8A8Etc2 = 95,
+        /// <summary>
+        /// Adaptive scalable texture compression ; 4x4 matrix using rgba channel interpretation
+        /// </summary>
+        ASTC_4x4_Rgba = 96,
```
dellis1972, line 189: This is a style/consistency choice. Looking at the other enumerations we historically do not include underscores. Should this be `Rgba4Astc` to fit in with the other tytpes? (or something similar).

dellis1972 (changes requested):
Looking good. A few minor changes but this looks ok to me. 
Did you get a chance to test this on windows or linux?

It might be worth us getting https://github.com/MonoGame/MonoGame/pull/8436/ done so we can run at least some of the tests on the bots.

cdhanna:
>Did you get a chance to test this on windows or linux?

I need to do more testing, and I plan to, but I've been using Github Actions as a test env to run on win/linux/mac/mac-m1

https://github.com/cdhanna/MonoGame/actions/runs/10570544720/job/29285179566

The m1 still has the `FreeImage` issue; and I've just been a dunce and not come up with a valid texture/config combo yet; but I will soon.

cdhanna:
Okay, I have a more comprehensive Github Actions run that checks compressing textures across several platforms, both profiles, and operating systems. 

Other than the FreeImage issue, it seems to be working. I did have a bug in my code that I will commit back shortly :+1:
https://github.com/cdhanna/MonoGame/actions/runs/10578126737

dellis1972:
> Another quirk is that I have to do a `dotnet tool restore` to make things work. Again, I'm not 100% clear on what the plan is to make those dotnet tools available, and/or if it is included in another planned work item.

@harry-cpp given we need to do a `dotnet tool restore` to get this PR to even work on develop, where is the best place to put that so it automatically runs as part of the `build/Build.csproj` ?

harry-cpp:
Its temporary so anywhere is fine.

dellis1972:
@cdhanna I woulod guess for now add a step to https://github.com/MonoGame/MonoGame/blob/develop/build/BuildToolsTasks/BuildContentPipelineTask.cs to restore the tools

cdhanna:
I've gone ahead and added some lines to the `BuildContentPipelineTask` to create the `.config/dotnet-tools.json` file for the local development. _If_ you happen to run the build script, _or_ intuit to run `dotnet tool restore`; then the `BitmapContent` tests will pass. 

At the risk of having a discussion in two places, I left a comment on the original bounty. I feel like I'm flying blind. I'm not sure how this supposed to work. https://github.com/MonoGame/MonoGame/issues/8419#issuecomment-2322671797

dellis1972 (approved):
This is looking good to me. 
I'd like to get a review from one of the other maintainers as its such a large PR, just to make sure I haven't missed anything.

MonoGame.Framework.Content.Pipeline/ExternalTool.cs
```diff
@@ -56,6 +65,11 @@ public static int Run(string command, string arguments, out string stdout, out s
                 RedirectStandardInput = true,
             };
 
+            if (!string.IsNullOrEmpty(workingDirectory))
+            {
+                processInfo.WorkingDirectory = workingDirectory;
+            }
```
tomspilman, line 71: We usually don't bracket single line conditions/loops.

mrhelmut:
I'll try to review this promptly!

tomspilman (approved):
In general it seems fine to me.

My only concerns are how does it feel performance wise with the new conversion libs?

cdhanna:
> My only concerns are how does it feel performance wise with the new conversion libs?

Not sure how to answer this yet. I'm away this weekend, but I can setup some Benchmark.Dotnet tests on my windows machine when I get back. I can say that the compression formats that require using external processes, like basisU or crunch, are _much_ slower than the in-process BcnEncoder library.

dellis1972:
@cdhanna (not in this bounty) but do you think (given the work you have done) we could write an 'C' API which would use Crunch or BasisU directly rather than going via the command line? I remember seeing basisu had a example of sorts. But it might help improve the performance if we could pass the texture data directly without having to go via the file system.

It might be worth another bounty if it helps speed things up a bit.

cdhanna:
> @cdhanna (not in this bounty) but do you think (given the work you have done) we could write an 'C' API which would use Crunch or BasisU directly rather than going via the command line? I remember seeing basisu had a example of sorts. But it might help improve the performance if we could pass the texture data directly without having to go via the file system.
> 
> It might be worth another bounty if it helps speed things up a bit.

Yeah, I'm sure it would improve stuff! I think this library is doing just that https://github.com/roy-t/SuperCompressed/blob/main/src/SuperCompressed/TranscodeFormats.cs

cdhanna:
I know there is a small format change about some brackets, and I still owe a benchmark. I'm fine with it getting merged, but no stress on my end. I'm also happy to hold it and explore more options if the community thinks that is best.

mrhelmut:
I believe that it just comes down to one of use testing it. Thank you!

dellis1972:
@cdhanna I just noticed we are missing some of the `ContextFactory` calls in https://github.com/MonoGame/MonoGame/blob/ec16df54e622ba9a2386943bcaf09e31564d8e06/Tools/MonoGame.Tools.Tests/TextureProcessorTests.cs#L74.

There are also some failing tests there. Not sure if they were failing before though.

```
Error Message:
   System.NotImplementedException : The method or operation is not implemented.
  Stack Trace:
     at Microsoft.Xna.Framework.Content.Pipeline.Graphics.TextureProfile.ConvertTexture(ContentProcessorContext context, TextureContent content, TextureProcessorOutputFormat format, Boolean isSpriteFont) in /Users/dean/Documents/Sandbox/Repos/MGcdhanna/MonoGame.Framework.Content.Pipeline/Graphics/TextureProfile.cs:line 90
   at Microsoft.Xna.Framework.Content.Pipeline.Processors.TextureProcessor.Process(TextureContent input, ContentProcessorContext context) in /Users/dean/Documents/Sandbox/Repos/MGcdhanna/MonoGame.Framework.Content.Pipeline/Processors/TextureProcessor.cs:line 118
   at MonoGame.Tests.ContentPipeline.TextureProcessorTests.CompressDefault[T](TargetPlatform platform, Color color, Int32 width, Int32 height) in /Users/dean/Documents/Sandbox/Repos/MGcdhanna/Tools/MonoGame.Tools.Tests/TextureProcessorTests.cs:line 272
   at MonoGame.Tests.ContentPipeline.TextureProcessorTests.CompressDefaultWindowsCutOut() in /Users/dean/Documents/Sandbox/Repos/MGcdhanna/Tools/MonoGame.Tools.Tests/TextureProcessorTests.cs:line 290
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

I was running 

`dotnet test Tools/MonoGame.Tools.Tests/MonoGame.Tools.Tests.csproj --blame-hang-timeout 1m -c Release`

but also 

`dotnet64 test Tools/MonoGame.Tools.Tests/MonoGame.Tools.Tests.csproj --blame-hang-timeout 1m -c Release`

where `dotnet64` is a sym link to the x64 install of dotnet

```
 ls -ls /usr/local/share/dotnet/dotnet64
0 lrwxr-xr-x  1 root  wheel  10 29 Feb  2024 /usr/local/share/dotnet/dotnet64 -> x64/dotnet
```

cdhanna:
I did some benchmarks, and as expected, this PR significantly _reduces_ the performance of compression across the board at the gain of supporting more platforms. 

As for the methodology...
I used Benchmarker on my windows machine. The machine has 32GB RAM, Win10 64 bit, 3.4Ghz 8CPUs, Nvidia GTX 1080 8GB VRAM. 

I haven't set up a full repo with the benchmark, but you can see the [benchmark script](https://gist.github.com/cdhanna/7d4a4e3ccd03e723b62b36b0fb8ead00) and the supporting [util code](https://gist.github.com/cdhanna/688d1cf81543f4d1ac267f9f45e1f292). 

ATC compression didn't work on my windows machine, and I didn't spend much time trying to figure out why. 
I ran the (slightly modified to remove dependencies on this branch) benchmark against monogame/develop, and recorded those runs with the prefix, "Old_". I then ran the benchmark with this branch (and re-enabled all the new dependencies on the contextFactory etc...), and prefixed those runs with "New_". 

I ran the benchmark in release mode. 

Here are the results, 
```
| Method                       | Mean         | Error        | StdDev       | Gen0     | Gen1   | Allocated |
|----------------------------- |-------------:|-------------:|-------------:|---------:|-------:|----------:|
| Old_Compress_Dxt1            |     199.3 us |      3.91 us |      5.60 us |  58.5938 | 0.7324 | 238.91 KB |
| New_Compress_Dxt1            |     781.7 us |      8.59 us |      7.18 us | 109.3750 | 1.9531 | 449.81 KB |
| Old_Compress_Dxt1a           |     198.2 us |      3.92 us |      6.66 us |  58.5938 | 0.2441 | 238.75 KB |
| New_Compress_Dxt1a           |     749.4 us |     12.28 us |     10.89 us | 110.3516 | 1.9531 | 449.79 KB |
| Old_Compress_Dxt3            |     207.4 us |      0.98 us |      0.91 us |  58.5938 | 0.4883 | 240.91 KB |
| New_Compress_Dxt3            | 306,989.6 us |  4,229.30 us |  3,749.17 us |        - |      - | 786.82 KB |
| Old_Compress_Dxt5            |     218.0 us |      0.81 us |      0.76 us |  58.8379 | 0.2441 |  240.8 KB |
| New_Compress_Dxt5            |     799.0 us |     12.56 us |      9.80 us | 113.2813 | 1.9531 | 465.78 KB |
| Old_Compress_Etc1            |   7,610.9 us |     73.49 us |     68.74 us |  31.2500 |      - | 136.83 KB |
| New_Compress_Etc1            | 306,160.4 us |  5,625.28 us |  6,018.98 us |        - |      - | 617.83 KB |
| New_Compress_Etc2            | 305,002.6 us |  2,774.39 us |  2,316.74 us |        - |      - | 624.04 KB |
| New_Compress_Astc            | 631,243.6 us | 10,892.19 us |  9,655.64 us |        - |      - | 971.48 KB |
| Old_Compress_Pvrtc_Rgb       |   6,193.9 us |     28.07 us |     24.88 us |  31.2500 |      - | 136.98 KB |
| New_Compress_Pvrtc_Rgb       | 635,680.6 us | 12,335.36 us | 11,538.50 us |        - |      - | 969.34 KB |
| Old_Compress_Pvrtc_Rgba      |   6,198.8 us |     26.22 us |     20.47 us |  31.2500 |      - | 136.98 KB |
| New_Compress_Pvrtc_Rgba      | 629,203.5 us |  6,344.57 us |  4,953.42 us |        - |      - |  969.6 KB |
| Old_Compress_AtcExplicit     |           NA |           NA |           NA |       NA |     NA |        NA |
| New_Compress_AtcExplicit     |     738.3 us |     14.67 us |     14.41 us |  71.2891 | 5.8594 | 288.65 KB |
| Old_Compress_AtcInterpolated |           NA |           NA |           NA |       NA |     NA |        NA |
| New_Compress_AtcInterpolated |     757.3 us |     15.02 us |     13.31 us |  73.2422 | 1.9531 | 298.59 KB |
```

The only real positive thing to say about these results are that they're even possible on the ARM CPU. But performance has taken a pretty big nose dive otherwise.

cdhanna:
@dellis1972 , thanks for the heads up about the tests. I should have run them explicitly on windows. On my mac, these failures were obscured by the `Unable to load shared library 'FreeImage' or one of its dependencies` issue. 
I've just tried to quickly update them, but they're still failing on my windows machine with a new issue, `Incorrect pixel format`. I'm out of time this morning (the kids have awoken), but I will take a look when I can today.

mrhelmut:
Thank you for the effort of running those tests. Users using texture compression with lots of textures are likely marginal MonoGame users, it's a downside I think we should accept and address later.

Future plans include moving what we can into a single native library for MGCB to improve performances, should this start to become a widespread issue. It's just cool to have those results so that we know that it is a point to keep an eye on.

dellis1972:
@cdhanna 

Thanks for the updates. It looks like the `TestProcessorContext` needs to implement the `IntermediateDirectory` and probably the `OutputDirectory` now as we are using tests which write to disk.

cdhanna:
@dellis1972 , okay, some updates.

I hooked up the context scope stuff, and configured the existing scope stuff to implement the intermediate directory. I'm able to run this command and get all tests passing on my windows machine. 
```
dotnet test Tools/MonoGame.Tools.Tests/MonoGame.Tools.Tests.csproj --blame-hang-timeout 1m -c Release
```

There were a few other issues perhaps worth calling out. 
First, I thought we'd talked about the fallback behaviour for ETC, but I couldn't find that comment-thread anywhere. Anyway, the code as written as goofing a bit. If a texture wasn't a POT texture, the ETC compression logic was _always_ falling back to bgra4444. Previously, there was an alpha based fallback. So I added this back. [Now, if the texture isn't POT, then it'll use a fallback based on the alpha.](https://github.com/MonoGame/MonoGame/pull/8456/commits/e40f383f954dfdfc4c6da3e223cf8224eb914666#diff-37a0beedf17141b267965f66061702d42451da617536f0244474d6cf907261b2L238)

A minor add on is that some tests indeed are new being converted to ETC2, not their fallbacks. 

Second, the Pvrtc mipmaps were broken. I missed the important fact that pvrtc compression has a minimum width & height, and that wasn't being handled correctly. When the really tiny mipmaps were being generated, the size coming back from the KTX reader wasn't happy with the Pvrtc. I did some research, and it [appears that the actual constraint is that the byte array be 32 long](https://github.com/BinomialLLC/basis_universal/issues/30), so I just [padded it out.](https://github.com/MonoGame/MonoGame/pull/8456/commits/e40f383f954dfdfc4c6da3e223cf8224eb914666#diff-0a6d01f0955dafa5a94554920b46641e8d7a92831859ac72b2cfcba9e57ac6d9R100)

dellis1972 (approved):
This is looking good. I'm gonna merge this now thanks for the work 😄

PR #8457, opened 2024-08-23
Code refactor Microphone/SoundEffect to modern C# code

realcoloride:
Hello,

I open this PR to refactor the following files to more modern C# code.
image

**This is my first PR here, so please feel free to review and give your feedback.**

### DIFF
--- a/MonoGame.Framework/Audio/Microphone.cs
+++ b/MonoGame.Framework/Audio/Microphone.cs
@@ -30,15 +30,8 @@ namespace Microsoft.Xna.Framework.Audio
     {
         #region Internal Constructors
 
-        internal Microphone()
-        {
-
-        }
-
-        internal Microphone(string name)
-        {
-            Name = name;
-        }
+        internal Microphone() { }
+        internal Microphone(string name) => Name = name;
 
         #endregion
 
@@ -60,59 +53,46 @@ namespace Microsoft.Xna.Framework.Audio
         /// </summary>
         public TimeSpan BufferDuration
         {
-            get { return _bufferDuration; }
+            get => _bufferDuration;
             set
             {
                 if (value.TotalMilliseconds < 100 || value.TotalMilliseconds > 1000)
                     throw new ArgumentOutOfRangeException("Buffer duration must be a value between 100 and 1000 milliseconds.");
                 if (value.TotalMilliseconds % 10 != 0)
                     throw new ArgumentOutOfRangeException("Buffer duration must be 10ms aligned (BufferDuration % 10 == 0)");
+
                 _bufferDuration = value;
             }
         }
 
-        // always true on mobile, this can't be queried on any platform (it was most probably only set to true if the headset was plugged in an XInput controller)
-#if IOS || ANDROID
-        private const bool _isHeadset = true;
-#else
-        private const bool _isHeadset = false;
-#endif
         /// <summary>
         /// Determines if the microphone is a wired headset.
         /// Note: XNA could know if a headset microphone was plugged in an Xbox 360 controller but MonoGame can't.
         /// Hence, this is always true on mobile platforms, and always false otherwise.
         /// </summary>
-        public bool IsHeadset
-        {
-            get { return _isHeadset; }
-        }
-
-        private int _sampleRate = 44100; // XNA default is 44100, don't know if it supports any other rates
+        // always true on mobile, this can't be queried on any platform (it was most probably only set to true if the headset was plugged in an XInput controller)
+#if IOS || ANDROID
+        public const bool IsHeadset = true;
+#else
+        public const bool IsHeadset = false;
+#endif
 
         /// <summary>
         /// Returns the sample rate of the captured audio.
         /// Note: default value is 44100hz
         /// </summary>
-        public int SampleRate
-        {
-            get { return _sampleRate; }
-        }
-
-        private MicrophoneState _state = MicrophoneState.Stopped;
+        public static int SampleRate => 44100; // XNA default is 44100, don't know if it supports any other rates
 
         /// <summary>
         /// Returns the state of the Microphone. 
         /// </summary>
-        public MicrophoneState State
-        {
-            get { return _state; }
-        }
+        public MicrophoneState State { get; private set; } = MicrophoneState.Stopped;
 
         #endregion
 
         #region Static Members
 
-        private static List<Microphone> _allMicrophones = null;
+        private static List<Microphone> AllMicrophones = null;
 
         /// <summary>
         /// Returns all compatible microphones.
@@ -122,9 +102,9 @@ namespace Microsoft.Xna.Framework.Audio
             get
             {
                 SoundEffect.Initialize();                
-                if (_allMicrophones == null)
-                    _allMicrophones = new List<Microphone>();
-                return new ReadOnlyCollection<Microphone>(_allMicrophones);
+                AllMicrophones ??= [];
+
+                return new(AllMicrophones);
             }
         }
 
@@ -135,8 +115,11 @@ namespace Microsoft.Xna.Framework.Audio
         /// </summary>
         public static Microphone Default
         {
-            get { SoundEffect.Initialize(); return _default; }
-        }       
+            get {
+                SoundEffect.Initialize();
+                return _default;
+            }
+        }
 
         #endregion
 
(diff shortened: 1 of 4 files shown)


Maniekko:
From [Contributing guidelines](https://github.com/MonoGame/MonoGame/blob/develop/CONTRIBUTING.md):

> DO NOT send PRs for code style changes or make code changes just for the sake of style.

mrhelmut:
Unless there is a use to refactoring PRs (e.g. performance gains, issues solved) , we tend to try to avoid them. There's a few reasons to this (among other things):
- Avoid possible regressions in code that have been just working fine for a while
- Make the git history less readable
- Possibly less readable to newcomers due to too many syntactic sugars
- Might break some private targets not running on the very latest .NET runtime

That being said, we don't want to discourage contributions and we're open to discussions. Our main incentive is to try to justify refactors with solving issues/projects.

realcoloride:
Hey there. I do believe this PR doesn't just change the style but also
improves some areas like making some parts static or use gets and sets
instead of a field and property

Maniekko:
Changing something to `static` or adding getters/setters is a breaking change, since things like serialisation are relying on it.

realcoloride:
> Changing something to `static` or adding getters/setters is a breaking change, since things like serialisation are relying on it.

This is the Microphone class we are talking about

SimonDarksideJ:
Whilst your intentions are honorable @realcoloride , these changes while making the classes more readable does not actually improve the functionality in question.  The generated IL from the compiler would be identical.

We welcome any contribution that helps to improve or resolve known issues within the Framework, however, there is no linked issue or alteration within this PR which changes behaviour (except for defaults such as `return sampleCount * 2; // 16bit adjust`

So unfortunately we need to reject this PR under that basis and we will also update the contribution guidelines to make that clearer.

If there is a specific issue you are trying to resolve, then please log an issue detailing the problem and by all means supply a PR, with test cases, demonstrating the fix.

We hope you understand this position where we do not want to introduce unnecessary risks in functionality that has not changed in over 6 years.  There is also ongoing work to replace the native backend for the majority of the platforms for native implementations.

I do hope you continue to contribute to the evolution of MonoGame, but we recommend before undertaking a considerable effort to raise an issue proposing change and avoid lost work.

realcoloride:
Hello,

> We hope you understand this position where we do not want to introduce unnecessary risks in functionality that has not changed in over 6 years. There is also ongoing work to replace the native backend for the majority of the platforms for native implementations.
> I do hope you continue to contribute to the evolution of MonoGame, but we recommend before undertaking a considerable effort to raise an issue proposing change and avoid lost work.

I understand the the guidelines you are talking about and care about this project, however, some parts of the code **are strictly** not necessary. For instance, using properties everywhere when there are things in place like `{ get; private set; }` where necessary should stay as an improvement
image

image

MonoGame also recently went through changing .NET version, including the language version. To me, this indicates it would be also better to use higher version language syntax. Like in here:
image

image

image


And here for example, I just improve the visual appeal of the code by using reverse conditions to avoid nesting:

image

image

**As you can see, no logical sense of the code was changed, so it isn't harming the codebase or any kind of output in any way.** The code does the same, but more efficiently, and uses probably less ram (since you are using less variables to store objects for instance).

I don't think that modernizing and refactoring in proper modern C# code like I show here is negative, in fact I think it should spark more positivity and educate towards the new modern and brilliant C# syntax that should also improve the way developers do things here. Switching to .NET 8.0 comes with brand new syntax advantages, and I believe they should be taken advantage of to level up of the codebase, and some parts simply should not have to be so redundant.

Mainly, you do not need to always create `_privateProperty` just to store a public field that returns a get and private set property when C# has a built in solution like `get; private set;` and other constraints.

I am not trying to strictly enforce a specific type of programming, but simply making things more optimized and more logical than they should be while respecting new and modern C# syntax (12.0/13.0) and improve code readability.

mrhelmut:
We know what better syntax may bring but you're just confirming exactly what we'd like to avoid: using syntactic sugars for the sake using of syntactic sugars.

Please re-read both of our statements. There are reasons for not opening a race to modernize every bits of a stable codebase. Your comment doesn't challenge any of our statement.

For instance, here's a few things we'd like to see challenged if we should go into a refactoring project:
- How would you ensure that merging one refactor PR doesn't open a race to modernize all of the codebase?
- How would you ensure that this doesn't bloat the git history and ease the PR review process with many non-functional PRs against fewer but more important functional PRs?
- How would you ensure that changes which seemingly don't have functional changes _really_ don't have functional changes? (i.e. how would you ensure that human reviews are flawless to avoid breaking years-long stable code?)
- How would you ensure that using the latest version of C# works on every runtime, including private console runtimes?
- How does using syntactic sugars everywhere possible makes a codebase more readable, accessible to contribute, and easier to maintain? (i.e. should we really turn a readable and stable codebase into a C++like gatekeeping project?)

We know that some people would like to see MonoGame being "more modern", but does its coding style really is about making it more appealing? Most of users just use MonoGame as a black-box and don't even know how it works and how it's made. Coding style is only a valid problem for maintainers.

realcoloride:
### Hello, to answer your questions:

> How would you ensure that merging one refactor PR doesn't open a race to modernize all of the codebase?

Technically nothing forbids anyone from making modernizations PRs later on. However this PR is not just here to modernize the code, but to improve it.

> How would you ensure that this doesn't bloat the git history and ease the PR review process with many non-functional PRs against fewer but more important functional PRs?

This should not bloat the git history as much because the changes are relatively straightforward. However, the PR review process is not managed by me, so I am not exactly sure how do you exactly go by your review process. _I don't think this is a question up for me to answer._

> How would you ensure that changes which seemingly don't have functional changes really don't have functional changes? (i.e. how would you ensure that human reviews are flawless to avoid breaking years-long stable code?)

Regarding my PR, my changes are really straightforward and do not change the functional aspect of the code or its logic. It is simply here to improve, **and I am always up to talk about the changes and reviews.** I also have replied earlier with a brief summary of the changes and how they improve the code, but I can always go in more details and elaborate.

**Though, collaboration and communication is key, which I am open to.**

> How would you ensure that using the latest version of C# works on every runtime, including private console runtimes?

MonoGame has recently transitioned to using .NET 8.0, which by default uses C# 12.0, [according to the official microsoft documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning), therefore it should not have problems with all runtimes, since all runtimes (except console?) indicate using `<TargetFramework>net8.0</TargetFramework>` in their `.csproj` file.
image

However, the private console runtimes is something that is beyond my reach, as I do not have any knowledge of the constraints required by private console runtimes. I am currently not aware of the .NET version and C# version consoles use (if they even use .NET to reliably work.) and I am open for discussion and improvement.

> How does using syntactic sugars everywhere possible makes a codebase more readable, accessible to contribute, and easier to maintain? (i.e. should we really turn a readable and stable codebase into a C++like gatekeeping project?)

**My goal is not to make "syntactic sugar" code, and definitely not unreadable either.**

The syntax I currently use makes the code if not easier to read and maintain. The changes are in priority logical and improve readability and maintability.

I have also full interest in making sure to speak every step along the way to make sure everyone understands the changes and the reasons behind me changing parts of the code. If you wish to understand how it makes it more readable, you do not have to re-read my previous message, but you can always ask me any question towards my changes and if you have feedback. 

For instance, one of my main issues with the code previously was the fact useless properties and fields were created, effectively bloating the code, or that a field could have been a public constant the entire time and effectively remove some bloat, etc..

I believe that inciting any developer, from a contributor to an enthusiast reading MonoGame's code, or even a beginner to look at the code in a way that should reflect how C# should be used. 

If we talk simply esthetical though, to me, if the C# version offers new powerful syntaxes to use that can ease development or make it better, it should be used. One of the primary examples in the commit that represents this is the following:

image

Here like you can see, the new C# syntax offers you the ability to create Collections by using `= []`, effectively making it easier _and much less to write_ & also avoid an extra `if ( x == null )` condition via using `??=`. ***Why not use it?***

> We know that some people would like to see MonoGame being "more modern", but does its coding style really is about making it more appealing? Most of users just use MonoGame as a black-box and don't even know how it works and how it's made. Coding style is only a valid problem for maintainers.

I again must state, **this PR is not just purely about style and syntax**; meaning it will benefit both maintainers and users. 

However, I must state again that I appreciate the dedication the team has for the project, and that my goal is just to make room for improvement, while always keeping feedback and collaboration in mind.

SimonDarksideJ:
I would also like to call to your attention @realcoloride another critical factor against "Modernizing" code at this stage, as I was reminded while updating the coding guidelines.

[Code Guidelines for .NET version support](https://github.com/MonoGame/MonoGame/blob/develop/CONTRIBUTING.md#code-guidelines)

Due to the current implementation and how it is consumed on some AoT platforms, we have a limitation in the language we can use currently.  This will be superseded in future releases as we move to a more "Native" implementation of classes, but as it stands, some of the language features you have used will break some platforms.

Although as you rightly pointed out, we have updated the base implementation of MonoGame to .NET 8, mainly to gain performance enhancements from the compiler, this sadly does not yet mean we can fully utilize all C#12 and .NET 8 features at this time, although it is the eventual intention.

We welcome you passion to improve the framework, but handling soo many platforms with different dependencies sometimes means we have to sacrifice newer language features to ensure support for all, at least in their current implementation.

We hope you continue your passion to help improve the framework.

realcoloride:
Hello there, I understand the constaints. However, knowing these constraints, why cant the LangVersion be tested to check what is the highest compatible LangVersion and then be specified in the `.csproj` files? In order to help with the syntax specifications.

And furthermore, how does the AoT platforms limit the C# language or .NET constraints? (what C# version could it be limited by?)

Also, how do the AoT platforms in question react to the .NET Standard (2.1)?

Cheers

PR #8466, opened 2024-08-29
Fix OpenAL SoundEffectInstance Panning Issue on DesktopGL and Mobile

### LINKED ISSUE #6543: SoundEffect panning does not split properly
From the forums: http://community.monogame.net/t/soundeffect-panning-does-not-split-properly/11092

The docs say that the pan value of `SoundEffect.Play` ranges from -1.0f to 1.0f, but audio is only playing out of the left speaker, right speaker, or both, equally. `pan = 0.1f` is being treated the same as `pan = 1.0f`. This behaved as expected when using the SharpDX library / XAudio, but not after we moved to OpenGL.

This is the function we wrote to play the audio.

```
public static void PlaySound(string name, float vol = 0.5f, float pan = 0.0f, bool jitter = false)
{
    bool ok = SFX.TryGetValue(name, out SoundEffect audio);
    if (ok)
    {
        float pitch = jitter ? ((float)Utils.Random.NextDouble() / 4.0f) : 0.0f;
        audio.Play(vol, pitch, pan);
    }
}
```

- MonoGame 3.7

- Windows 10, macOS Mojave

- DesktopGL (bug did not occur with WindowsDX)

### LINKED ISSUE #6876: SoundEffect pan property not working
Since the Apply3D isn't working at all, I tried to build a workaround with "pan". But setting pan from -1f to 1f doesn't change anything. SoundEffect appears on both speakers. WAV file is in stereo format.

```
            using (var effectInstance = soundEffectBuffer[effect].CreateInstance())
            {
                effectInstance.Volume = volume;
                effectInstance.Pitch = pitch;
                effectInstance.Pan = pan;
                effectInstance.Play();
            }
```

Also did try the suggested solution #6543

```
            using (var effectInstance = soundEffectBuffer[effect].CreateInstance())
            {
                effectInstance.Play();
                effectInstance.Volume = volume;
                effectInstance.Pitch = pitch;
                effectInstance.Pan = pan;
            }
```

without any difference.

- MonoGame 3.7.1.189

- Windows

- DesktopGL

dellis1972:
Fixes #6876
Fixes #6543

### Description of Change

So OpenAL has an outstanding issue where is does not support stereo panning (see https://github.com/kcat/openal-soft/issues/194)

They do however have an extension `AL_EXT_STEREO_ANGLES` which can produce a similar effect. So lets make use of this extension if it is supported to support panning of stereo audio.

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.Android.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.Android.csproj
@@ -13,6 +13,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="MonoGame.Library.OpenAL" Version="1.23.1.7" />
+  </ItemGroup>
+
+  <ItemGroup>
     <Compile Remove="bin\**\*" />
     <Compile Remove="obj\**\*" />
     <Compile Remove="Platform\**\*" />
@@ -76,21 +80,6 @@
 
   <ItemGroup>
     <None Include="MonoGame.Framework.Android.targets" Pack="true" PackagePath="build" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <EmbeddedNativeLibrary Include="..\ThirdParty\Dependencies\openal-soft\libs\armeabi-v7a\libopenal32.so">
-      <Link>libs\armeabi-v7a\libopenal32.so</Link>
-    </EmbeddedNativeLibrary>
-    <EmbeddedNativeLibrary Include="..\ThirdParty\Dependencies\openal-soft\libs\arm64-v8a\libopenal32.so">
-      <Link>libs\arm64-v8a\libopenal32.so</Link>
-    </EmbeddedNativeLibrary>
-    <EmbeddedNativeLibrary Include="..\ThirdParty\Dependencies\openal-soft\libs\x86\libopenal32.so">
-      <Link>libs\x86\libopenal32.so</Link>
-    </EmbeddedNativeLibrary>
-    <EmbeddedNativeLibrary Include="..\ThirdParty\Dependencies\openal-soft\libs\x86_64\libopenal32.so">
-      <Link>libs\x86_64\libopenal32.so</Link>
-    </EmbeddedNativeLibrary>
   </ItemGroup>
 
   <Import Project="Platform\OpenGL.targets" />
--- a/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
@@ -18,6 +18,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="MonoGame.Library.SDL" Version="2.26.5.5" />
+    <PackageReference Include="MonoGame.Library.OpenAL" Version="1.23.1.7" />
     <PackageReference Include="NVorbis" Version="0.10.4" />
   </ItemGroup>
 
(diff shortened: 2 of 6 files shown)


dellis1972:
Requires new versions of libopenal for Android (and probably ios) to work there.

squarebananas:
This PR almost fixes it, but it doesn't quite mix the channels correctly yet. Here is an example of ```SoundEffectInstance.Pan``` being toggled between -1 and 1 each second.

AudioPanning

WindowsDX and FAudio both handle this correctly. DesktopGL is very different in comparison. I have just about written a fix though which seems to work fine so far. I'll test it some more first and post a PR during the week.

dellis1972:
@squarebananas cool. I took my best guess as I didn't have access to a Windows box to test the DX side to compare. 
If you can get this closer that would be great 👍

mrhelmut:
Any contribution is welcome, though just noting that a move to FAudio is planned. Just so people contributing to the OpenAL implementation can gauge the energy they are willing to put into that.

dellis1972:
@squarebananas any update of an improvement? I ask because the foundation are going to try moving to a monthly release so it would be nice to get a better fix in. That said , if not we can ship it as is because it's better than nothing 😆

squarebananas:
> @squarebananas any update of an improvement? I ask because the foundation are going to try moving to a monthly release so it would be nice to get a better fix in. That said , if not we can ship it as is because it's better than nothing 😆

I'll get it sorted out tomorrow.

squarebananas:
@dellis1972 I still need to test these fixes on a few different devices, but I've submitted a draft PR in the meantime.

vchelaru:
A few notes:

I built MonoGame and tried this out in a test project DesktopGL. Two notes:

1. I had to copy over the openal.dll file to the bin folder. Not sure if that's expected
2. Left/right panning seemed to work with values -1 and +1, but a value of 0 seemed to still pan full left

squarebananas:
> I built MonoGame and tried this out in a test project DesktopGL

Did you build & test that with or without my additional fixes PR #8480? Also are the original sound files mono or stereo as they are handled quite differently internally to make panning work on OpenAL.

vchelaru:
I did the develop branch so no, it did not include that PR. 

This is the file I used - I had to zip it since wav is not supported by github. I'm not sure how to check mono vs stereo actually.

[SoundEffectFile.zip](https://github.com/user-attachments/files/17000861/SoundEffectFile.zip)

squarebananas:
> I'm not sure how to check mono vs stereo actually.

One way is copy the file to a directory with ffprobe.exe (try MonoGame\ThirdParty\Dependencies\ffmpeg\Windows\x64), open command prompt, navigate to the directory and run:
`ffprobe -i SoundEffectFile.wav -select_streams a:0`

It should output something like:
`Input #0, wav, from 'SoundEffectFile.wav':
  Duration: 00:00:00.02, bitrate: 1432 kb/s
    Stream #0:0: Audio: pcm_s16le ([1][0][0][0] / 0x0001), 44100 Hz, 2 channels, s16, 1411 kb/s`

Also it looks like my PR has now been accepted to the develop branch if you want to try again.

dellis1972:
> 1. I had to copy over the openal.dll file to the bin folder. Not sure if that's expected

The MonoGame.Library.OpenAL package should contain a `runtimes` folder which has the relevant dll/so/dylibs in it. 
If you were referencing the MonoGame.Framework.DesktopGL.csproj directly you might need to copy over the dll. 
However using the NuGet it should be picked up as a dependency just like SDL.

Can you give me more details? What was the setup in this senario?

squarebananas:
>I did the develop branch so no, it did not include that PR.

Looks like it will work after syncing with develop. Here is some pan cycling code and the audacity output before/after applying #8480 when looping your sound effect. 

`soundEffectInstance.Pan = (float)Math.Sin(gameTime.TotalGameTime.TotalSeconds);`

AudioPanning3

Issue #8195, opened 2024-02-22
[PAID BOUNTY] [LOCKED] Implement the Direct3D 12 / GDK/GDKX backend


mrhelmut:
_MonoGame needs a Direct3D 12 implementation based on the GDK/GDKX._

📃**Full details**

This implementation will be the new Windows and Xbox targets, and would replace the current WindowsDX, UWP, and Xbox targets altogether.

"D3D 11 on 12" is not allowed in this context. Therefore, we require an actual D3D12 implementation. The use of DirectXTK is allowed.

@tomspilman has worked on a partial implementation already but couldn't finish it. He's making it available to the applicant in a private repository. The state of this work requires pipeline and memory management to be done.

The Xbox (GDKX) and Windows (GDK) targets are expected to share 99% of the code. We expect that one codebase to be shared between the future WindowsDX and Xbox targets (which should be Windows Store compliant).

🔒 **Bounty status**

Locked for @Togimaro to complete. 🔒

💰 **Bounty**

$5.000,00 USD (Five thousands US dollars).

✋ **Requirements to apply**

_In addition to [our generic bounty requirements](https://monogame.net/bounties/#requirements), here are requirements specific to this bounty._

This bounty requires to be a registered console developer with Microsoft in order to access the GDKX.

📦 **Expected delivery content**

- A Windows GDK implementation of MonoGame
- An Xbox GDKX implementation of MonoGame running with NativeAOT.

📅 **Delivery timeline**

As soon as possible.

mrhelmut:
Following the foundation meeting of March 14, the applications have been exanimated and this bounty has been locked for @Togimaro to complete.

tomspilman:
This bounty is completed.

amerkoleci:
Hi,
Curious where is the source code? or is it under NDA codebase?

Thanks

tomspilman:
FYI.

https://github.com/MonoGame/MonoGame/pull/8646

PR #8474, opened 2024-09-02
Add GDK-related platforms to enums and csproj

Togimaro:
See #8195.

This PR add the GDK platforms to various enums, this is necessary to build the [GDK backend](https://github.com/MonoGame/MonoGame.XB1/pull/51).

Unfortunately this add two GDK-specific #ifdef to the code, one in Game.cs already present for DX11 and another one to WinFormsGamePlatform.cs in order to add a Prepare() call used by the DX12 backend and to avoid calling SharpDX related stuff. This could probably be removed when merging when #8242 is ready.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
+++ b/MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
@@ -91,6 +91,15 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         /// All desktop versions using Vulkan.
         /// </summary>
         DesktopVK,
+        
+        /// Windows GDK
+        /// </summary>
+        WindowsGDK,
+
+        /// <summary>
+        /// Xbox Series
+        /// </summary>
+        XboxSeries
     }
 
 
--- a/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
+++ b/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
@@ -174,5 +174,6 @@
   <Import Project="..\PlayStation4\MonoGame.Framework.Content.Pipeline.targets" Condition="exists('..\PlayStation4\MonoGame.Framework.Content.Pipeline.targets')" />
   <Import Project="..\PlayStation5\MonoGame.Framework.Content.Pipeline.targets" Condition="exists('..\PlayStation5\MonoGame.Framework.Content.Pipeline.targets')" />
   <Import Project="..\PSVita\MonoGame.Framework.Content.Pipeline.targets" Condition="exists('..\PSVita\MonoGame.Framework.Content.Pipeline.targets')" />
+  <Import Project="..\GDKX\MonoGame.Framework.Content.Pipeline.targets" Condition="exists('..\GDKX\MonoGame.Framework.Content.Pipeline.targets')" />
 
 </Project>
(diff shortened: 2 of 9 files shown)


MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
```diff
@@ -85,7 +85,22 @@ public enum TargetPlatform
...
+
+        /// <summary>
+        /// Xbox One (GDKX) 
+        /// </summary>
+        GamingXboxOne,
+
+        /// <summary>
+        /// Xbox Series
+        /// </summary>
+        GamingScarlett
```
mrhelmut, line 103: @tomspilman thoughts on naming? Otherwise good to me.

tomspilman: I would go with `GamingXboxSeries` because we've avoided code names on all the other ones.

Togimaro: > But why the `Gaming` prefix?

It was mainly because it's the name of the target in GDK:
image
I didn't give much thought when naming these but yeah it's a bit weird as a platform name

I will change that for WindowsGDK and XboxSeries and reuse XboxOne if everyone is ok with that?

mrhelmut: Sounds good to me as suggested, thanks!

mrhelmut:
Good to go to me! @tomspilman your call.

MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
SimonDarksideJ: I agree, we shouldn't use codenames even if they are in the GDK itself.  Better to set our own statement.

So I agree with WindowsGDK and XboxSeries and reuse XboxOne.  Makes more sense and it clearer to normal devs when they see it.

SimonDarksideJ (approved):

MonoGame.Framework.Content.Pipeline/TargetPlatform.cs
```diff
@@ -85,7 +85,17 @@ public enum TargetPlatform
         /// <summary>
         /// WebAssembly and Bridge.NET
         /// </summary>
-        Web
+        Web,
+
+        /// <summary>
+        /// Windows GDK
+        /// </summary>
+        DesktopGDK,
```
tomspilman, line 93: Sorry @Togimaro  i thought i had commented on this a few days ago, but couldn't find it.

The term "Desktop" here is used to indicate that platform works on all desktop PCs... 
 Windows, Mac, and Linux.  This is why we have `DesktopGL`.

`DesktopGDK` would then imply GDK works on Mac and Linux... which it does not.

So `WindowsGDK` would be the right name here as it really is Windows desktop only.

It also then matches the other platform enum `MonoGamePlatform`.

Togimaro: Fixed!

mrhelmut (approved):

SimonDarksideJ:
@Togimaro can you rebase this PR with development to resolve the conflicts, so we can get this merged asap?

tomspilman:
Oops... i was working from @Togimaro branch when i tested GDKX.  So i missed this.

tomspilman:
@Togimaro what do you mean? There is no architectural change to the project to worry about.

Togimaro:
> @Togimaro what do you mean? There is no architectural change to the project to worry about.

A lot of stuff have moved around I mean, like the `PlatformGetBytecode` in `EffectResource`, the ShaderStage passed to the Sampler/TextureCollection etc. It's trivial to port...but it still need to be changed in the other repo unfortunately

tomspilman:
If you want you can just fix this PR. I can fix the Xbox side since I'm on it already. But it's your call.  @Togimaro

Togimaro:
Should be good to merge now, a new [PR is opened in the MonoGame.XB1 repo](https://github.com/MonoGame/MonoGame.XB1/pull/52)

mrhelmut:
Thanks! Merging.

PR #8480, opened 2024-09-10
Additional fixes for SoundEffectInstance.OpenAL panning

squarebananas:
This PR provides additional fixes on top of #8466 for supporting SoundEffectInstance.OpenAL panning.

These are the points I noted while testing/fixing:

### AudioChannels.Mono Sound Sources
- During playback of mono sources, it is not possible to pan fully between left & right.
- OpenAL only applies 3D positions to mono channel sources.
- When fully panning, moving the position in only one axis will not fully stop output on the opposite speaker.
- Also the varying distance between the emitter & listener gives the perception of pan control being less uniform.
- Through testing, fully panned left/right output is observed when positioning the emitter at +/-60° at constant distance.

### AudioChannels.Stereo Sound Sources
- During playback of stereo sources, channels are not mixed correctly.
- OpenAL only applies StereoAngles to stereo channel sources.
- StereoAngles are applied counter-clockwise.
- With StereoAngles set both to 0°, normal left/right stereo output occurs.
- Through testing, full left/right output is observed when the StereoAngles are both set at +/-30°.
- If the StereoAngles extension is unsupported, stereo sources cannot be panned at all.
- As a workaround, the user could manually split each stereo file into 2 mono files and play simultaneously.

### Other Notes
- The percentage of cross channel audio mixed for each speaker may change depending on the platform/driver OpenAL implementation.
- I found this particularly noticeable with headphones using Realtek audio on Windows where it is never mixed entirely to just one side. This does not occur on XAudio2 with headphones.

### Description of Changes
- Determines if source is AudioChannels.Mono or AudioChannels.Stereo and uses 3D emitter positioning or StereoAngles for panning.
- Rotates the 3D emitter position between +/- 60° using Vector2.Rotate for constant distance through both axes.
- Sets StereoAngles both to the same value (between -30° to 30°) for full left/right panning.
- ALHelper.CheckError descriptions have been corrected or changed.
 
### Results
AudioPanning2
_Audacity levels for mono/stereo sources panned from -1 to 1 instantly (square) or using a sine wave. A tone was used for the mono tests and a drum loop for the stereo test._

I've submitted this as a draft for now, but all that's left is testing on a few different devices (only just realised mobile platforms use OpenAL as well). I could make a sample project for testing if you like.

### DIFF
--- a/MonoGame.Framework/Platform/Audio/SoundEffectInstance.OpenAL.cs
+++ b/MonoGame.Framework/Platform/Audio/SoundEffectInstance.OpenAL.cs
@@ -225,29 +225,37 @@ namespace Microsoft.Xna.Framework.Audio
 
             if (HasSourceId)
             {
-                // Pan range values: -1.0 (fully left) --- 0.0 (centered) --- 1.0 (fully right)
-                // The proportion of cross channel audio mixed for each speaker may change depending on the
-                // platform/driver OpenAL implementation
+                // Pan value  |   -1.0   |    0.0   |   +1.0   |
+                // Output     |   Left   | Centered |   Right  |
+                // - The proportion of cross channel audio mixed for each speaker may change depending on the
+                //   platform/driver OpenAL implementation
 
+                float maxPanAngle = (float)Math.PI / 3f;
                 switch (sourceChannels)
                 {
                     case AudioChannels.Mono:
-                        // Pan via 3D emitter positioning
+                        // Simulate pan via 3D emitter positioning
                         // - Rotates the emitter position +/- 60 degrees around the listener while keeping a constant distance
                         // - OpenAL only applies 3D positions to mono channel sources
-                        Vector2 pannedPosition = Vector2.Rotate(new Vector2(0, -1), _pan * ((float)Math.PI / 3f));
+                        Vector2 pannedPosition = Vector2.Rotate(new Vector2(0, -1), _pan * maxPanAngle);
                         AL.Source(SourceId, ALSource3f.Position, pannedPosition.X, 0.0f, pannedPosition.Y);
                         ALHelper.CheckError("Failed to set source position.");
                         break;
 
                     case AudioChannels.Stereo:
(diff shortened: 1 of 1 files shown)


dellis1972:
@squarebananas thank for you this PR, it looks great. Love the detail in the PR message 😃 
I'll review it properly later today, a quick scan over it looks fine thought 👍 

The OpenAL audio mix isn't really something we can control as you said. But this should get us much closer and help to fix a long standing issue 🥇

squarebananas:
I've corrected an issue I found with my stereo angle changes. While full panning was fixed, it resulted in centered (pan value of zero) incorrectly having an output similar to playing a mono version of the file.

---
### Initial angles
```
Pan value                | -1.0 |  0.0 | +1.0 |
panAngles[0] (Output L)  | +60° | +30° |   0° |
panAngles[1] (Output R)  |   0° | -30° | -60° |
```
✅ Centered - Outputs normal stereo mix.
❌ Fully panned - Sent to both outputs.

### First commit angles
```
Pan value                | -1.0 |  0.0 | +1.0 |
panAngles[0] (Output L)  | +30° |   0° | -30° |
panAngles[1] (Output R)  | +30° |   0° | -30° |
```
❌ Centered - Outputs like mono mix.
✅ Fully panned - Sent to a single output.

### Second commit angles
```
Pan value                | -1.0 |  0.0 | +1.0 |
panAngles[0] (Output L)  | +90° | +30° | -30° |
panAngles[1] (Output R)  | +30° | -30° | -90° |
```
✅ Centered - Outputs normal stereo mix.
✅ Fully panned - Sent to a single output.

---
I've finished testing DesktopGL as well as Android now. As far as I can tell, everything is working as expected and sounds closer to the XAudio2 implementation than before.

dellis1972:
@squarebananas if you think this is good to go, make it as ready for review and we can look at getting it merged 👍

dellis1972:
@squarebananas thanks for all the hard work on this 😄

PR #8489, opened 2024-09-12
Fixed Matrix.GetHashCode() implementation

itsBuggingMe:
The current `Matrix.GetHashCode()` uses addition to combine the hash codes of the float fields. This could easily lead to collisions, such as when just swapping two members of a matrix.
```csharp
//example
Matrix m1 = Matrix.Identity;
Matrix m2 = Matrix.Identity;

(m1.M11, m1.M12) = (m1.M12, m1.M11);

Debug.WriteLine(m1.GetHashCode());//these two return the same value
Debug.WriteLine(m2.GetHashCode());
Debug.WriteLine(m1 == m2);//False
```

To fix this, the only change was to use `HashCode.Combine()` to combine the hashcodes.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Matrix.cs
+++ b/MonoGame.Framework/Matrix.cs
@@ -1674,7 +1674,9 @@ namespace Microsoft.Xna.Framework
         /// <returns>Hash code of this <see cref="Matrix"/>.</returns>
         public override int GetHashCode()
         {
-            return (((((((((((((((this.M11.GetHashCode() + this.M12.GetHashCode()) + this.M13.GetHashCode()) + this.M14.GetHashCode()) + this.M21.GetHashCode()) + this.M22.GetHashCode()) + this.M23.GetHashCode()) + this.M24.GetHashCode()) + this.M31.GetHashCode()) + this.M32.GetHashCode()) + this.M33.GetHashCode()) + this.M34.GetHashCode()) + this.M41.GetHashCode()) + this.M42.GetHashCode()) + this.M43.GetHashCode()) + this.M44.GetHashCode());
+            int h1 = HashCode.Combine(M11, M12, M13, M14, M21, M22, M23, M24);
+            int h2 = HashCode.Combine(M31, M32, M33, M34, M41, M42, M43, M44);
+            return HashCode.Combine(h1, h2);
         }
 
         /// <summary>
--- a/MonoGame.Framework/Matrix.cs
+++ b/MonoGame.Framework/Matrix.cs
@@ -1670,7 +1670,42 @@ namespace Microsoft.Xna.Framework
         /// <returns>Hash code of this <see cref="Matrix"/>.</returns>
         public override int GetHashCode()
         {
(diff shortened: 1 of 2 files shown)


dellis1972:
This is using an API which is not available for the console builds (net 4.5) so we cannot merge this at this time. It will have to wait until the console builds are upgraded to use NativeAOT.

AristurtleDev:
@itsBuggingMe we would need you to update your initial PR description to include the new statement regarding our LLM policy as shown in the updated PR templates here: https://github.com/MonoGame/MonoGame/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

```
### Contributor Declaration

```

AristurtleDev:
> This is using an API which is not available for the console builds (net 4.5) so we cannot merge this at this time. It will have to wait until the console builds are upgraded to use NativeAOT.

@tomspilman is this still a blocker because of NativeAoT updated?

itsBuggingMe:
I can also rewrite the implementation to not use `Hashcode.Combine`

AristurtleDev:
> I can also rewrite the implementation to not use `Hashcode.Combine`

I think that would be fine and we would be able to merge it in without worry on the console side 

Also noticed that the branch is out of date if you can update that. I don't think they'll be any merge conflicts from updating the branch.

AristurtleDev (dismissed):
Thanks @itsBuggingMe

MonoGame.Framework/Matrix.cs
```diff
@@ -1670,7 +1670,42 @@ public override bool Equals(object obj)
...
+
+                int mix = 17;
+
+                mix = mix * 31 + hash1;
+                mix = mix * 31 + hash2;
+                mix = mix * 31 + hash3;
+                mix = mix * 31 + hash4;
+
+                return mix;
+            }
```
tomspilman, line 1708: So this is an improvement on the old code for certain.

It is also continue to be stable and not randomized (like `System.HashCode` is now) which i think is good for us.

This will change the result of Matrix hashes which i hope doesn't break anyone's code that depended on the old result.  Maybe it does... but seems like it was a mistake to depend on GetHashCode as always being a fixed value.

tomspilman (dismissed):

MonoGame.Framework/Matrix.cs
dellis1972: So we need a unit test to check the behaviour which is outlined in the PR description? To make sure we don't regress?

AristurtleDev: @tomspilman @dellis1972 is this more of a statement, or do we need @itsBuggingMe to implement tests?  Otherwise it's been approved twice now so we should merge it

tomspilman: Yeah let's get a unit test in there.  It is always better for the framework long term to have them.

AristurtleDev (approved):

PR #8493, opened 2024-09-14
Add Self Hosted Runner for Tests

dellis1972:
This enables the UI unit tests on some self hosted machines we have available. 
There were some issues regarding running UI based tests in NUnitLite via `dotnet test` on MacOS. This is because you MUST create NSWindow instances on the Main Thread. So we had to modify the Test runner to marshal the UI tests over to the main thread. 

This added a new `[RunOnUI]` attribute, which can be used to flag a test which MUST run on the UI thread. This works for all platforms. 

There were still some slight issues with the test on MacOS if running ALL the UI tests in on go they would fail. But running them in smaller batches works. I suspect that there is a bug in one of the tests which is causing this we need to track it down. 

A few tests have been disable, issues need to be opened for these so that they can be fixed. 

Also fixed Short2 and NormalizedByte4 types so they actually pass the tests.

Fixed the casing on the `Artifacts` folder so it is consistent across all platforms (Linux and Mac were using lower case in some areas). 

Fixed the Adpcm audio support so it unpacks the data correctly.

### DIFF
--- a/MonoGame.Framework/Graphics/PackedVector/NormalizedByte4.cs
+++ b/MonoGame.Framework/Graphics/PackedVector/NormalizedByte4.cs
@@ -99,10 +99,10 @@ namespace Microsoft.Xna.Framework.Graphics.PackedVector
 
         private static uint Pack(float x, float y, float z, float w)
         {
-            var byte4 = (((uint) MathF.Round(MathHelper.Clamp(x, -1.0f, 1.0f) * 127.0f)) & 0xff) << 0;
-            var byte3 = (((uint) MathF.Round(MathHelper.Clamp(y, -1.0f, 1.0f) * 127.0f)) & 0xff) << 8;
-            var byte2 = (((uint) MathF.Round(MathHelper.Clamp(z, -1.0f, 1.0f) * 127.0f)) & 0xff) << 16;
-            var byte1 = (((uint) MathF.Round(MathHelper.Clamp(w, -1.0f, 1.0f) * 127.0f)) & 0xff) << 24;
+            var byte4 = (uint)(((int) MathF.Round(MathHelper.Clamp(x, -1.0f, 1.0f) * 127.0f)) & 0xff) << 0;
+            var byte3 = (uint)(((int) MathF.Round(MathHelper.Clamp(y, -1.0f, 1.0f) * 127.0f)) & 0xff) << 8;
+            var byte2 = (uint)(((int) MathF.Round(MathHelper.Clamp(z, -1.0f, 1.0f) * 127.0f)) & 0xff) << 16;
+            var byte1 = (uint)(((int) MathF.Round(MathHelper.Clamp(w, -1.0f, 1.0f) * 127.0f)) & 0xff) << 24;
 
             return byte4 | byte3 | byte2 | byte1;
         }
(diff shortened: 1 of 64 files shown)


MonoGame.Framework/Graphics/PackedVector/NormalizedByte4.cs
```diff
@@ -99,10 +99,10 @@ public override string ToString()
...
         private static uint Pack(float x, float y, float z, float w)
         {
-            var byte4 = (((uint) MathF.Round(MathHelper.Clamp(x, -1.0f, 1.0f) * 127.0f)) & 0xff) << 0;
-            var byte3 = (((uint) MathF.Round(MathHelper.Clamp(y, -1.0f, 1.0f) * 127.0f)) & 0xff) << 8;
-            var byte2 = (((uint) MathF.Round(MathHelper.Clamp(z, -1.0f, 1.0f) * 127.0f)) & 0xff) << 16;
-            var byte1 = (((uint) MathF.Round(MathHelper.Clamp(w, -1.0f, 1.0f) * 127.0f)) & 0xff) << 24;
+            var byte4 = (uint)(((int) MathF.Round(MathHelper.Clamp(x, -1.0f, 1.0f) * 127.0f)) & 0xff) << 0;
+            var byte3 = (uint)(((int) MathF.Round(MathHelper.Clamp(y, -1.0f, 1.0f) * 127.0f)) & 0xff) << 8;
+            var byte2 = (uint)(((int) MathF.Round(MathHelper.Clamp(z, -1.0f, 1.0f) * 127.0f)) & 0xff) << 16;
+            var byte1 = (uint)(((int) MathF.Round(MathHelper.Clamp(w, -1.0f, 1.0f) * 127.0f)) & 0xff) << 24;
```
tomspilman, line 105: I guess this was broken code spotted by the unit test runs?

MonoGame.Framework/Graphics/PackedVector/Short2.cs
```diff
@@ -115,8 +115,8 @@ private static uint PackInTwo (float vectorX, float vectorY)
 			const float minNeg = ~(int)maxPos; // two's complement
 
             // clamp the value between min and max values
-            var word2 = ((uint) MathF.Round(MathHelper.Clamp(vectorX, minNeg, maxPos)) & 0xFFFF);
-            var word1 = (((uint) MathF.Round(MathHelper.Clamp(vectorY, minNeg, maxPos)) & 0xFFFF) << 0x10);
+            var word2 = (uint)((int) MathF.Round(MathHelper.Clamp(vectorX, minNeg, maxPos)) & 0xFFFF);
+            var word1 = (uint)(((int) MathF.Round(MathHelper.Clamp(vectorY, minNeg, maxPos)) & 0xFFFF) << 0x10);
```
tomspilman, line 119: I guess this was also broken code spotted by the unit test runs?

tomspilman (approved):
Generally looks good to me!

SimonDarksideJ (approved):

MonoGame.Framework/Graphics/PackedVector/NormalizedByte4.cs
dellis1972: Yes it was failing one of the unit tests.

MonoGame.Framework/Graphics/PackedVector/Short2.cs
dellis1972: Yes it was failing one of the unit tests.

PR #8505, opened 2024-09-25
Upgrade Mac Editor to net8.0-macos. Upgrade Eto.Forms

dellis1972:
Fixes https://github.com/MonoGame/MonoGame/discussions/8469

Upgrade Eto.Forms to the latest stable version.
Switch to using `net8.0-macos` for the mac editor so we can support both x64 and arm64 architectures.
Upgrade AssimpNet to 5.0.x
Switch over to using MonoGame.Library.Assimp to get macOS arm64 support.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,6 +75,11 @@ jobs:
         run: wget -qO- https://monogame.net/downloads/net9_mgfxc_wine_setup.sh | bash
         if: runner.os != 'Windows'
 
+      - uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: latest-stable
+        if: runner.os == 'macOS'
+
       - name: Install required workloads
         run:   |
               if [ "$RUNNER_OS" == "Linux" ]; then
@@ -82,7 +87,7 @@ jobs:
               elif [ "$RUNNER_OS" == "Windows" ]; then
                     dotnet.exe workload install android ios macos
               else
-                    dotnet workload install android
+                    dotnet workload install android macos ios
               fi
         shell: bash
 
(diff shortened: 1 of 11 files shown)


mrhelmut:
Just a side note: we really need to get the content pipeline dependencies sorted out because things like reference AssimpNet while excluding parts of it to replace it with our own native build feels really backward. If we provide a native lib, we should provide our own minimal C# wrapper because replacing 1 nuget with 2 frankenstien nugets sounds like more problems than we're trying to solve.

dellis1972:
> Just a side note: we really need to get the content pipeline dependencies sorted out because things like reference AssimpNet while excluding parts of it to replace it with our own native build feels really backward. If we provide a native lib, we should provide our own minimal C# wrapper because replacing 1 nuget with 2 frankenstien nugets sounds like more problems than we're trying to solve.

I agree, however if we want to meet the next release this will have to do for now. I don't have time to write a C# wrapper for assimp when I have bigger problems to solve like how to install these `dotnet tools`.

SimonDarksideJ (approved):

zapoutix:
Hello, 
just tested on M3 with Sequoia 15.1 and dotnet 8.0.402 arm64 version

image

mrhelmut:
@dellis1972 Could you elaborate on the reference of ```SixLabors.ImageSharp```?

dellis1972:
> @dellis1972 Could you elaborate on the reference of `SixLabors.ImageSharp`?

Its a dependency of BCnEncoder.Net.ImageSharp. https://www.nuget.org/packages/BCnEncoder.Net.ImageSharp#dependencies-body-tab. 

I had to add an explicit PackageReference to stop the mac app from crashing on startup.

mrhelmut:
Pfeww, we really need to sort the content pipeline dependencies out.

Good to go to me.

DanielBeeck:
@zapoutix can you please explain how you got it running?
im going crazy trying to get it running.
I have a M2

PR #8517, opened 2024-10-05
Capture Asserts in MGCB logging

tomspilman:
This adds a trace listener to the start of MGCB so that asserts and other diagnostic messages appear in the output logging.

### DIFF
--- a/Tools/MonoGame.Content.Builder/Program.cs
+++ b/Tools/MonoGame.Content.Builder/Program.cs
@@ -13,11 +13,38 @@ namespace MonoGame.Content.Builder
 {
     class Program
     {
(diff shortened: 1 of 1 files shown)


dellis1972:
Is there a way to turn this off? Diagnostics generally slow down builds? Perhaps only if `/quiet` is not used?

tomspilman:
> Is there a way to turn this off? Diagnostics generally slow down builds? Perhaps only if /quiet is not used?

This is not adding any diagnostics.

This is simply ensuring *if* any debug diagnostics are enabled... the error message doesn't get silently consumed.

How it works now... the pipeline will just stop processing content with NOTHING in the output log as to why it stopped.

PR #8527, opened 2024-10-09
Install dotnet tools locally in the Nuget Packages

dellis1972:
In commit 7a398b01 we added two new tools for compressing textures, `crunch` and `basisu`. 
The problem there is that the management of the `.config/dotnet-tool.json` file by the users was becoming an issue. 
We needed a more automatic way to install the required tooling. 

The problem is using the standard `dotnet tool install` calls requires a `.config/dotnet-tool.json` file to be present in either the current directory or a directory that is in the path ABOVE the current one. You would use the `--create-manifest-if-needed` flag to create the manifest, but that will still leave the users having to manage and upgrade the .json file every time we do a release. 

So lets get the pipeline to install the tooling itself. 
The `dotnet tool install` command has an additional argument`--tool-path` this allows us to say where we want the tool installed. Once that has happened we get a native binary launcher in `--tool-path` which allows us to launch the app directly without using the `dotnet` executable. So what this allows us to do is install the tooling locally in the directory that the content pipeline is installed. This will usually be the global `.nuget/package` directory. 

We need to keep an eye on the `DOTNET_ROOT` environment variable when installing the tooling, just in case a user (or CI) wants to use a custom dotnet install. 

Fix the calls to `basisu` and `crunch` to handle spaces in paths. Same for other area's which shell out to `mgcb`. 

Also added support for overriding which MGCB to use in the editor by looking for a `MGCBCommand` environment variable. This should be the full path to either the `mgcb` exe or the `mgcb.dll` that the user wants to use. This will allow users to change to a custom content compiler without having to change the editor.

The one downside is users will no longer be able to run `dotnet mgcb` directly in their project directory unless they install the tooling manually. 

The Editor also needs to download the `mgcb` tool. We place this in the `ApplicationData` folder rather than in the local assembly folder because we cannot guarantee that folder is writable. So in this instance a location specific to the current user seemed like the best approach.

The long term plan is to bundle all these tools into a single native library which can be called directly by the content pipeline.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
+++ b/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
@@ -23,6 +23,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         private static string CrunchVersion = "1.0.4.2";
         public static string BasisU = "mgcb-basisu";
         private static string BasisUVersion = "1.16.4.2";
+        public static string FFmpeg = "mgcb-ffmpeg";
+        private static string FFmpegVersion = "7.0.0.6";
+        public static string FFprobe = "mgcb-ffprobe";
+        private static string FFprobeVersion = "7.0.0.6";
 
         static ExternalTool()
         {
@@ -69,6 +73,8 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
                 return;
             RestoreDotnetTool("install", Crunch, CrunchVersion, path);
             RestoreDotnetTool("install", BasisU, BasisUVersion, path);
+            RestoreDotnetTool("install", FFmpeg, FFmpegVersion, path);
+            RestoreDotnetTool("install", FFprobe, FFprobeVersion, path);
             File.WriteAllText(versionFile, version);
         }
 
--- a/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
+++ b/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
@@ -24,6 +24,11 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         public static string BasisU = "mgcb-basisu";
         private static string BasisUVersion = "1.16.4.2";
 
+        static ExternalTool()
+        {
+            RestoreDotnetTools ();
+        }
+
         public static int Run(string command, string arguments)
         {
             string stdout, stderr;
@@ -34,7 +39,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
             return result;
         }
 
-        public static void RestoreDotnetTool(string command, string toolName, string toolVersion, string path)
+        static void RestoreDotnetTool(string command, string toolName, string toolVersion, string path)
         {
             Directory.CreateDirectory(path);
             var exe = CurrentPlatform.OS == OS.Windows ? "dotnet.exe" : "dotnet";
@@ -50,7 +55,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
             }
         }
 
-        public static void RestoreDotnetTools()
+        static void RestoreDotnetTools()
         {
             var version = Assembly.GetExecutingAssembly().GetName().Version.ToString();
             string path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory);
@@ -71,7 +76,6 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
         /// </summary>
         public static int RunDotnetTool(string toolName, string args, out string stdOut, out string stdErr, string stdIn=null, string workingDirectory=null)
         {
-            RestoreDotnetTools();
             var exe = FindCommand(toolName);
             var finalizedArgs =  args;
             return ExternalTool.Run(exe, finalizedArgs, out stdOut, out stdErr, stdIn, workingDirectory);
--- a/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
+++ b/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
@@ -5,6 +5,8 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Net;
+using System.Reflection;
 using System.Threading;
 using MonoGame.Framework.Utilities;
 
@@ -17,6 +19,11 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
     /// </summary>
     internal class ExternalTool
     {
+        public static string Crunch = "mgcb-crunch";
+        private static string CrunchVersion = "1.0.4.2";
+        public static string BasisU = "mgcb-basisu";
+        private static string BasisUVersion = "1.16.4.2";
+
         public static int Run(string command, string arguments)
         {
             string stdout, stderr;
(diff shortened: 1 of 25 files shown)


build/DeployTasks/UploadArtifactsTask.cs
```diff
@@ -15,10 +15,52 @@ public override async Task RunAsync(BuildContext context)
...
+       
+
         await context.GitHubActions().Commands.UploadArtifact(new DirectoryPath(context.NuGetsDirectory.FullPath), $"nuget-{os}");
         await context.GitHubActions().Commands.UploadArtifact(new DirectoryPath(System.IO.Path.Combine(context.BuildOutput, "Tests", "Tools", "Release")), $"tests-tools-{os}");
         await context.GitHubActions().Commands.UploadArtifact(new DirectoryPath(System.IO.Path.Combine(context.BuildOutput, "Tests", "DesktopGL", "Release")), $"tests-desktopgl-{os}");
         if (context.IsRunningOnWindows())
             await context.GitHubActions().Commands.UploadArtifact(new DirectoryPath(System.IO.Path.Combine(context.BuildOutput, "Tests", "WindowsDX", "Release")), $"tests-windowsdx-{os}");
     }
+
+    void DeleteToolStore (BuildContext context, string path)
```
harry-cpp, line 52: extra space here and a couple of places bellow, and above in previous files :D

Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.props
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <DotnetCommand Condition="'$(DotnetCommand)' == ''">dotnet</DotnetCommand>
     <EnableMGCBItems Condition="'$(EnableMGCBItems)' == ''">true</EnableMGCBItems>
+    <MGCBToolDirectory>$(MSBuildThisFileDirectory)dotnet-tools</MGCBToolDirectory>
```
harry-cpp, line 6: Maybe add `/` at the end to follow how the other dir variables are structured?

MonoGame.Framework.Content.Pipeline/ExternalTool.cs
```diff
@@ -27,13 +31,39 @@ public static int Run(string command, string arguments)
...
+        }
+
         /// <summary>
         /// Run a dotnet tool. The tool should be installed in a .config/dotnet-tools.json file somewhere in the project lineage.
         /// </summary>
         public static int RunDotnetTool(string toolName, string args, out string stdOut, out string stdErr, string stdIn=null, string workingDirectory=null)
         {
-            var finalizedArgs = toolName + " " + args;
-            return ExternalTool.Run("dotnet", finalizedArgs, out stdOut, out stdErr, stdIn, workingDirectory);
+            RestoreDotnetTools ();
```
harry-cpp, line 63: Running the restore / install process every time is gonna be tedious, we should cache the result of dotnet tools being restored once it was successfully done so.

MonoGame.Framework.Content.Pipeline/ExternalTool.cs
```diff
@@ -68,6 +98,11 @@ public static int Run(string command, string arguments, out string stdout, out s
             if (!string.IsNullOrEmpty(workingDirectory))
                 processInfo.WorkingDirectory = workingDirectory;
 
+            var dotnetRoot = Environment.GetEnvironmentVariable ("DOTNET_ROOT");
+            if (!string.IsNullOrEmpty(dotnetRoot)) {
```
harry-cpp, line 102: Inlined `{` instead of new lined.

harry-cpp (changes requested):

MonoGame.Framework.Content.Pipeline/ExternalTool.cs
dellis1972: We check here https://github.com/MonoGame/MonoGame/pull/8527/files#diff-3f7b501f8d90c49eb02c352c09349a0f48f4c4fe74e0f0d025746ac3615bdd51R46 to see if its installed already

harry-cpp: Oh I see, totally missed it.

Tho I don't know if we should check have to run executables to check if the tool is installed, as right now the more tools we add, the slower they will all launch as for each tool separately we would have to run `dotnet.exe` once.

dellis1972: I'll add an extra file which we write so we only install all the tools once.

build/BuildToolsTasks/BuildMGCBEditorTask.cs
```diff
@@ -8,12 +8,15 @@ public sealed class BuildMGCBEditorTask : FrostingTask<BuildContext>
...
-        context.ReplaceRegexInFiles(
-            @"<key>CFBundleShortVersionString<\/key>\s*<string>([^\s]*)<\/string>",
-            "Tools/MonoGame.Content.Builder.Editor/Info.plist",
-            $"<key>CFBundleShortVersionString</key>\n\t<string>{context.Version}</string>",
-            RegexOptions.Singleline
-        );
+        if (context.BuildSystem().IsRunningOnGitHubActions)
+        {
+            context.ReplaceRegexInFiles(
+                "Tools/MonoGame.Content.Builder.Editor/Info.plist",
```
dellis1972, line 14: This arg was in the wrong place so this never worked 🤦

Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.props
```diff
@@ -3,7 +3,11 @@
   <PropertyGroup>
     <DotnetCommand Condition="'$(DotnetCommand)' == ''">dotnet</DotnetCommand>
     <EnableMGCBItems Condition="'$(EnableMGCBItems)' == ''">true</EnableMGCBItems>
+    <!-- Allow users the ability to disble tool restoration if needed -->
+    <AutoRestoreMGCBTool Condition="'$(AutoRestoreMGCBTool)' == ''">true</AutoRestoreMGCBTool>
+    <MGCBToolDirectory>$(MSBuildThisFileDirectory)dotnet-tools/</MGCBToolDirectory>
     <MGCBCommand Condition="'$(MGCBCommand)' == ''">mgcb</MGCBCommand>
+    <MonoGameVersion>3.8.1.1-develop</MonoGameVersion>
```
dellis1972, line 10: Do we need a better name for this? It will be replaced with the actual version of the app at build time on CI.

Tools/MonoGame.Content.Builder.Editor/Common/Util.cs
```diff
@@ -46,5 +48,41 @@ public static string GetRelativePath(string filespec, string folder)
...
+        }
+
+        public static Process CreateProcess(string command, string arguments, string workingDirectory, Encoding encoding, Action<string> output)
+        {
+            var exe = command;
+            var args = arguments;
+            if (command.EndsWith (".dll")) {
+                // we are referencing the dll directly. We need to call dotnet to host.
+                exe = Global.Unix ? "dotnet" : "dotnet.exe";
+                args = $"{command} {arguments}";
```
dellis1972, line 68: do we need to quote the command here? 🤔  to handle paths with spaces

Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.props
harry-cpp: You forgot to check the Condition before setting the variable.

dellis1972: I don't think the user should be overriding this. This will be the version of the NuGet package these targets were shipped in. Unless we want users to install the targets from one version of Mg but install the tooling from another?

dellis1972: I added the check anyway.

MonoGame.Framework.Content.Pipeline/ExternalTool.cs
```diff
@@ -27,13 +34,47 @@ public static int Run(string command, string arguments)
...
+                exe = Path.Combine(dotnetRoot, exe);
+            }
+            if (Run(exe, $"tool {command} {toolName} --version {toolVersion} --tool-path .", out string _, out string _,  workingDirectory: path) != 0)
+            {
+                // install the latest
+                Run(exe, $"tool {command} {toolName} --tool-path .", out _, out _,  workingDirectory: path);
+            }
+        }
+
+        public static void RestoreDotnetTools()
```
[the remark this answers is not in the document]
dellis1972: So I moved things around a bit, the `RestoreDotnetTools` now gets called in the static constructor on first use, then not at all after that, so that should help with file/io :D. I also made these called private as they are not part of the public api. 

We already have a exception handler for tools that are not present or missing, which happens here https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework.Content.Pipeline/ExternalTool.cs#L46. 

Given the plan is to swtich out allot of this stuff into a new native content pipeline so/dll/dylib which statically links all the tools we use and presents a simple API we can leave this as is for now I think.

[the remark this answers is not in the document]
cdhanna: Cool, the static constructor sounds good.

build/DeployTasks/UploadArtifactsTask.cs
```diff
@@ -27,4 +54,24 @@ public override async Task RunAsync(BuildContext context)
             await context.GitHubActions().Commands.UploadArtifact(new FilePath(vsixFilePath), "MonoGame.Templates.VSExtension.vsix");
         }
     }
+
+    void DeleteToolStore(BuildContext context, string path)
+    {
+        if (System.IO.Directory.Exists(path)) {
```
harry-cpp, line 60: { should be on new lines.

harry-cpp (changes requested):

build/DeployTasks/UploadArtifactsTask.cs
```diff
@@ -27,4 +54,24 @@ public override async Task RunAsync(BuildContext context)
...
+            var store = System.IO.Path.Combine (path, ".store");
```
harry-cpp, line 61: Space between combine and (.

Issue #8521, opened 2024-10-07
Packaging and running tests for DesktopVK


mrhelmut:
Now that the native backend has been setup and defined with #8309, MonoGame has a Vulkan implementation running as the ```DesktopVK``` target platform.

```DesktopVK``` is currently **_experimental_** (it has a few glitches) and doesn't ship with current MonoGame releases.
It is only usable through manually building the source code.

In order to start having users testing it and contributors fixing it, we need to package this platform into a nuget and to distribute it.

```DesktopVK``` depends on [the new native backend](https://github.com/MonoGame/MonoGame/tree/develop/native/monogame) that needs to be natively built for Windows, macOS, and Linux, and be packaged as runtime binaries within the ```DesktopVK``` nuget package.

Currently, Github Actions builds everything but doesn't package and distribute. Due to the dependency toward the native backend, it might be more appropriate that the native library is moved to its own repository (which self-builds its own nuget, [like we're doing for SDL](https://github.com/MonoGame/MonoGame.Library.SDL) and other native dependencies).

We also need to run the unit tests for ```DesktopVK``` on Windows, macOS, and Linux.
Currently tests are run for ```DesktopGL``` on Windows, macOS, and Linux, while ```WindowsDX``` tests only run on Windows.

Any help on that is welcome.

mrhelmut:
The reasoning would be to split build scripts for maintenance, and fit the SDL native dep' distribution model. But if we consider that having too many nugets is starting to be backward, it is indeed more desirable to have everything under one flag.

tomspilman:
> The reasoning would be to split build scripts for maintenance, and fit the SDL native dep' distribution model.

Ok we could do that.  Our SDL native repo doesn't include the SDL source code either.   It pulls it in from submodule.  We can do the same and just build the native backend code from this other repo just for distribution needs.

All the code could still live in the main MG repo.

Does that make sense @mrhelmut @harry-cpp ?

harry-cpp:
The main reason why I started them as separate repos is that they are 3rd party deps, and there is no need to rebuild them on every commit push, or really modify them unless we want to bump the version, and you don't even need to build them to build MG from source, as MG references those repos as nugets.

I think we wanna keep our native library code and build scripts inside our main repo, we can have manual build scripts for rebuilding the native binary if its taking to long to build it on every push, no need to split it up like the 3rd party deps.

harry-cpp:
[...]

PR #8542, opened 2024-10-22
Starting on GitHub Actions for Native builds 

Layoric:
Working on options for #8521 

### Description of Change

Working through the steps outlined in #8521 for native VK building to be in GitHub Actions. I'm not overly familiar with the MonoGame build process and there is a lot of moving parts, but trying to help with the orchestration and automation. I'll keep this PR in draft, and feel free to close if its not useful.

- BuildDesktopVKTask still present to not be as disruptive
- BuildNativeDesktopVKTask - New build task for Windows,Linux and macOS
- Following existing patterns with CakeFrosting + cmake + premake
- Added new `build-native.yml` GitHub Action as an isolated step focusing on building native resources for downstream steps.

I tried to fix some of the build errors as I saw them but reverted as I'm likely missing context to how its all put together.

Linux still not building, now hitting some build errors that others might need to take a look at, so this is just the GitHub Action and the CakeFrosting task as best as I could with info I could work out. Happy to fix things with some guidance.


tomspilman:
@harry-cpp can you give this a look and give us some guidance on it?

harry-cpp:
The CAKE cs file is fine.

As for the yaml file, we haven't really discussed on how we want to build the native library, we are gonna need to optimize it so it does not happen on every build and instead only happens when .cpp / .h files get modified probably.

Tho first things first, I'll do some work tomorrow and over the weekend to get the native backend working on macOS and Linux, afterwards we can come back about hooking it up to CI.

Layoric:
Thanks @harry-cpp , I've limited the builds to `native/monogame` path, and added manual workflow trigger, and artifact retention of 90 days which is the max for public repos. There are other action scripts to consume the artifacts from other builds or can interact with GH API to pull down latest version or specific SHA. Alternatively you could use tags or other ways to more easily target specific artifact builds.

SimonDarksideJ:
This might need rebasing and re-working under the new workflows being designed
https://github.com/MonoGame/MonoGame/pull/8665
The structure there is a lot cleaner and abstract, so injecting this additional build step would be easier.

SimonDarksideJ:
Any plans to continue / update this PR @Layoric , although I suspect it has been superseded with recent build changes

Layoric:
@SimonDarksideJ Thanks, will close this PR off, don't want to disrupt things if already progress with a different solution

PR #8544, opened 2024-10-24
Fixes SPIR-V globals key lookup & parsing int/uint/float input types

squarebananas:
### Issue

Various custom effects failed to build. This was happening at the shader creation stage where the SPIR-V code is used to generate the `ConstantBuffer` along with its parameters' data types / sizes set.

### Description of Changes
- Changed `"%type__MG_Globals"` to `"%type__Globals"` according to this line in the SPIR-V output: `"type__Globals" OpName %type__Globals "type.$Globals"`.
- Initial changes for supporting parsing scalar types where only vector types worked before. For example `v2float` worked but `float` didn't. (https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_types)  
- Initial changes for supporting `uint` data type. (https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_unsigned_versus_signed_integers)


mrhelmut:
@tomspilman might want to check that out.

PR #8553, opened 2024-11-01
(README) Replace ship game example to an image

Maniekko:
One thing i forgot to check about updated README is the attached files sizes, and turns out the ship game gif is pretty massive (**46mb**). I tried to compress the gif, but i was only able to reduce it's size to 41mb, which is still too much.
The problem with this file is that it creates an ugly empty space while the file is loaded, and it is loaded **every time (!)** the main repository page is accessed. It also severely slows down the page load time.
image

*On a side note, Auto Pong example image is just 161kb*


squarebananas:
Some alternative suggestions:
1) Reduce the resolution closer to the display area = 7.5 MB
Ship7
2) Remove frames, speed up, reduce color, accidentally corrupt it & call it "style" = 2.3 MB
ship2

SimonDarksideJ (approved):

SimonDarksideJ:
Thanks @squarebananas , also replaced the source gif in the sample repo for the smaller one.  Will keep an eye on gif size int the future

SimonDarksideJ:
He He , although I do love the "corrupt" version @Maniekko 
How about we use this one for the updated version @tomspilman

Maniekko:
> He He , although I do love the "corrupt" version .Maniekko How about we use this one for the updated version .tomspilman

I think you messed up the pings 😅
Squarebananas was the one who created "corrupt" version and i created the PR

PR #8554, opened 2024-11-01
Sprite font multi byte fix

### LINKED ISSUE #3984: Bug with Monogame SpriteFont Importer
Been having a problem with displaying Japanese characters in my app ever since using the monogame pipeline.

help for charracters

 So I look and found that the spriteFonts are imported by a new importer. I put a stop point in my program before it loaded the font and swaped the Font xnb generated by the monogame importer for one made by the Xna one, this seem to solve the problem i was having. 

help for charracters4

Thought I should post this in case anyone else was having problems.

CartBlanche:
Fixes #3984

### Description of Change

Hopefully better checks to support multi-byte characters.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegion.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegion.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
         public CharacterRegion(char start, char end)
         {
             if (start > end)
-                throw new ArgumentException();
+                throw new ArgumentException($"The start character '{start}' must not be greater than the end character '{end}'.");
 
             Start = start;
             End = end;
--- a/MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegion.cs
+++ b/MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegion.cs
@@ -4,23 +4,23 @@ using System.Collections.Generic;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
 {
-	// Describes a range of consecutive characters that should be included in the font.
-	[TypeConverter(typeof(CharacterRegionTypeConverter))]
-	public struct CharacterRegion
-	{
-	    public char Start;
-	    public char End;
+    // Describes a range of consecutive characters that should be included in the font.
+    [TypeConverter(typeof(CharacterRegionTypeConverter))]
+    public struct CharacterRegion
+    {
+        public char Start;
+        public char End;
 
-		// Enumerates all characters within the region.        
-	    public IEnumerable<Char> Characters()
-	    {
-	        for (var c = Start; c <= End; c++)
-	        {
-	            yield return c;
-	        }
-	    }
+        // Enumerates all characters within the region.
+        public IEnumerable<char> Characters()
+        {
+            for (var c = Start; c <= End; c++)
+            {
+                yield return c;
+            }
+        }
 
-	    // Constructor.
+        // Constructor.
         public CharacterRegion(char start, char end)
         {
             if (start > end)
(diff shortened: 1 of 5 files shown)


MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegionTypeConverter.cs
```diff
@@ -4,68 +4,93 @@
...
+                return value[0];
+            }
+            else
+            {
+                // Check if the value is hexadecimal with the "0x" prefix
+                if (value.StartsWith("0x", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Remove "0x" prefix and parse as hexadecimal
+                    value = value.Substring(2);
+                    if (int.TryParse(value, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out int hexResult))
```
tomspilman, line 63: I would change `value` here to `value.Substring(2)`.  This keeps you from changing the value of `value` so if this fails the exception message below gets the original value of it.

MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegionTypeConverter.cs
```diff
@@ -4,68 +4,93 @@
...
+                // Single character directly specifies a codepoint.
+                return value[0];
+            }
+            else
+            {
+                // Check if the value is hexadecimal with the "0x" prefix
+                if (value.StartsWith("0x", StringComparison.OrdinalIgnoreCase))
+                {
+                    // Remove "0x" prefix and parse as hexadecimal
+                    value = value.Substring(2);
```
tomspilman, line 62: I would create a temporary here, so that the error exception below shows the original value.

CartBlanche: Great idea. Fixed.

stromkos:
The higher Unicode characters require 2 `char`  slots to encode in UTF-16.  You need to use a non-standard, other than "&#0000",  to indicate a higher codepoint that could be directly interpreted as a string from a UTF-32 codepoint.

The new `Rune` class was supposed to help.

CartBlanche:
@stromkos Could you please clarify and maybe link to the Rune class you mentioned?

MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegion.cs
```diff
@@ -4,23 +4,23 @@
...
+            {
+                yield return c;
+            }
+        }
 
-	    // Constructor.
+        // Constructor.
         public CharacterRegion(char start, char end)
         {
             if (start > end)
```
kjpou1, line 26: Maybe consider updating the exception to include a clear message. For example:

    if (start > end)
        throw new ArgumentException($"The start character '{start}' must not be greater than the end character '{end}'.");

As well as a unit test.

squarebananas:
> @stromkos Could you please clarify and maybe link to the Rune class you mentioned?

I would think `System.Text.Rune`.
https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-text-rune

Tools/MonoGame.Tools.Tests/CharacterRegionTypeConverterTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            Assert.Throws<ArgumentException>(() => converter.ConvertFrom(null, null, ""));
+            Assert.Throws<ArgumentException>(() => converter.ConvertFrom(null, null, "A--Z"));
+            Assert.Throws<ArgumentException>(() => converter.ConvertFrom(null, null, "Invalid"));
+            Assert.Throws<ArgumentException>(() => converter.ConvertFrom(null, null, "0xZZ"));
+        }
+
+        [Test]
+        public void ConvertFrom_InvalidRangeOrder_ThrowsArgumentException()
+        {
+            Assert.Throws<ArgumentException>(() => converter.ConvertFrom(null, null, "Z-A"));
```
CartBlanche, line 98: @kjpou1 I believe this unit test and the next one should catch the start > end error, for both standard and multi-byte characters.

CartBlanche:
@stromkos, @squarebananas I've not been keeping up to date with what MS has to offer Unicode wise. Thanks both for bringing Rune to my attention.

MonoGame.Framework.Content.Pipeline/Graphics/Font/CharacterRegion.cs
CartBlanche: Exception message updated. Thanks.

CartBlanche:
@stromkos, @squarebananas thanks again for your suggestions. We're discussing Rune and if we decide to use it, it will happen in a future PR, maybe along with the font fallback fix. For now this PR will get merged, as it does fix an existing parsing error.

stromkos:
The primary error is in the assumption that "value.Length == 1" implies a single char makes up the codepoint via `value[0]`.

The `value[0]` should be checked for the continuation value and amended to include the second value if needed. (Third or fourth, If the UTF-8 file loader would allow)

This change would allow BSP characters directly in the start and end fields. 

----

My suggestion of the Rune class removes the 16-bit limitation of the `char` storage.

Calling `Char.ConvertFromUtf32() ` instead of the `(char)` cast would have corrected the issue without the need for custom code.

----

This PR is yet another patch layer that never addresses the actual issue.

Issue #8555, opened 2024-11-01
HashKey collision in ShaderProgramCache


arosolino:
### Prerequisites

- [X] I have verified this issue is present in the `develop` branch
- [X] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

MonoGame 3.8.2

### Which MonoGame platform are you using?

MonoGame Cross-Platform Desktop Application (mgdesktopgl)

### Operating System

Windows

### Description

The bitwise OR method for creating a hash key between the vertex and pixel shader easily causes collisions.
https://github.com/MonoGame/MonoGame/blob/8a21b20fdfd7a2d36fd52be73d93984e7469bc01/MonoGame.Framework/Platform/Graphics/Shader/ShaderProgramCache.cs#L72

For example this is the collision I ran into. Both had the same vertex shader hash key  (-382211486), with different pixel shader hash key (1330679647, 1717683999).
```
-382211486 | 1330679647 = -277352577
-382211486 | 1717683999 = -277352577
```

Can we get a better hashing method implemented here like
```c# 
var key = 17;
key = key * 23 + vertexShader.HashKey;
key = key * 23 + pixelShader.HashKey;
```

### Steps to Reproduce

If you need the specific shaders that produce these hash codes let me know.

### Expected Behavior

The two different pixel shaders should create a different hash key when combined with the same vertex pixel shader.

### Resulting Behavior

The two different pixel shaders are producing the same hash key when combined with the same vertex pixel shader.

tomspilman:
This seems like a good change... lets get a PR in there.

arosolino:
Alright made a pull request
https://github.com/MonoGame/MonoGame/pull/8557

stromkos:
I would suggest updating the actual hash code instead of applying a fix to a single use case.

https://github.com/MonoGame/MonoGame/blob/8b35cf50783777507cd6b21828ed0109b3b07b50/MonoGame.Framework/Utilities/Hash.cs#L18

Extend the hash function and all calls to it to use a 64 bit hash.

-------

In this case the collision is known at/near build time, and easily corrected manually once identified. Any single change in either source will correct the problem.

May I suggest a `try catch` around the `Dictionary.Add` to notify the user of the collision.

tomspilman:
[...]

PR #8557, opened 2024-11-03
Improved shader cache hash key to not cause collisions

arosolino:
Fix for the following issue
https://github.com/MonoGame/MonoGame/issues/8555

I made the key used for the shader program cache use a value tuple so it can avoid potential hash key collisions.

Also made use of `TryGetValue` so it only has to check the key in the dictionary once.

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/Shader/ShaderProgramCache.cs
+++ b/MonoGame.Framework/Platform/Graphics/Shader/ShaderProgramCache.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Xna.Framework.Graphics
     /// </summary>
     internal class ShaderProgramCache : IDisposable
     {
-        private readonly Dictionary<(int, int), ShaderProgram> _programCache = new Dictionary<(int, int), ShaderProgram>();
+        private readonly Dictionary<ulong, ShaderProgram> _programCache = new Dictionary<ulong, ShaderProgram>();
         GraphicsDevice _graphicsDevice;
         bool disposed;
 
@@ -69,7 +69,7 @@ namespace Microsoft.Xna.Framework.Graphics
             // buffers here as well.  This would allow us to optimize
             // setting uniforms to only when a constant buffer changes.
 
-            var key = (vertexShader.HashKey, pixelShader.HashKey);
+            var key = ((ulong)(uint)vertexShader.HashKey << 32) | (uint)pixelShader.HashKey;
     
             if(!_programCache.TryGetValue(key, out var shaderProgram))
             {
--- a/MonoGame.Framework/Platform/Graphics/Shader/ShaderProgramCache.cs
+++ b/MonoGame.Framework/Platform/Graphics/Shader/ShaderProgramCache.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Xna.Framework.Graphics
     /// </summary>
     internal class ShaderProgramCache : IDisposable
     {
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Platform/Graphics/Shader/ShaderProgramCache.cs
```diff
@@ -37,7 +37,7 @@ public int GetUniformLocation(string name)
     /// </summary>
     internal class ShaderProgramCache : IDisposable
     {
-        private readonly Dictionary<int, ShaderProgram> _programCache = new Dictionary<int, ShaderProgram>();
+        private readonly Dictionary<(int, int), ShaderProgram> _programCache = new Dictionary<(int, int), ShaderProgram>();
```
tomspilman, line 40: I don't like the idea of changing this into a Tuple.

I would do what was suggested in the original issue report (https://github.com/MonoGame/MonoGame/issues/8555) and implement a better hash.

Or store a `ulong` and pack the two `int`s into it if we want to avoid a hash here.

arosolino: The original idea that was suggested could still potentially cause a collision.

Let's just go the ulong route and completely eliminate the chance.

Ill supply the update.

arosolino: Okay they update is in 4ad91310c88323a7b4a0602a4c9905518537afd1

tomspilman:
Looks good to me.

stromkos:
In this case the collision is known at/near build time, and easily corrected manually once identified. Any single change in either source will correct the problem.

May I suggest a `try catch` around the `Dictionary.Add` to notify the building user of the collision, instead of **doubling** the run-time memory required for each dictionary entry on all client computers.

arosolino:
> In this case the collision is known at/near build time, and easily corrected manually once identified. Any single change in either source will correct the problem.
> 
> May I suggest a `try catch` around the `Dictionary.Add` to notify the building user of the collision, instead of 1.5x the run-time memory required for each dictionary entry on all client computers.

I agree with you that small change to the source will give it a different hash key. But that small change could end up causing a collision with another shader. There is no warning what is happening, and it took me looking through the MonoGame source code to figure it out.

I don't see how using `try catch` around the `Dictionary.Add` would work. It's only being added to the dictionary if it doesn't already exist. In the case of the collision it already exists and `Dictionary.Add` will not be called. Could you provide an example of what you are thinking of?

stromkos:
> But that small change could end up causing a collision with another shader. 

That is both the danger and the purpose of a hash.  Small change == large changes in the output.

You are correct:
Line 73 : ` if (!_programCache.ContainsKey(key))` should not exist as it is only added once during normal use.

arosolino:
> You are correct: Line 73 : ` if (!_programCache.ContainsKey(key))` should not exist as it is only added once during normal use.

The `GetProgram` method is called every time the vertex/pixel shader needs to be activated. This is called every single frame not just once. Therefore removing `if (!_programCache.ContainsKey(key))` would call `Dictionary.Add` every time you use the vertex/pixel shader and break it's usage.

The increased memory changing the dictionary key to a 64-bit is minimal as you are generally only linking a small number of shaders. This PR also gets a perf boost by removing the double lookup on the dictionary by utilizing `Dictionary.TryGetValue`.

dellis1972 (approved):

PR #8560, opened 2024-11-07
Update MonoGame.Library.OpenAL to 1.23.1.10

dellis1972:
An issue was reported that our OpenAL binaries were not working on Linux. It turns out that we had not installed the developer packages for most of the supported backends. As a result we only supported ALSA.

The CI has been updated to include all the supported backends with the exception of SoundIO. This is because if you build with SoundIO support you get a hard link to the library. However it is not a common library. Having a hard link stops the .so being loaded if it is not installed.

So lets bump all the projects that use OpenAL to use the latest package.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -171,14 +171,20 @@ jobs:
         include:
           - os: windows
             platform: windows
+            shell: cmd
           - os: macos
             platform: macos
+            shell: bash
           - os: ubuntu-latest
             platform: linux
+            shell: bash
             filter: --where="Category != Audio"
           # - os: linux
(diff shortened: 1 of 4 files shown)


SimonDarksideJ (approved):

CartBlanche:
@dellis1972 this is currently failing on during the Windows test. I'm looking into it.

CartBlanche:
@dellis1972 I ran the same tests locally and don't get the process failure like on CI. Does it have to run through power shell or could it run inside a cmd terminal, just to see if that give a different result?

CartBlanche (approved):
LGTM

PR #8564, opened 2024-11-09
Fixed BoundingBox.Contains(BoundingFrustum) method #6298

### LINKED ISSUE #6298: Error in BoundingBox.Contains(BoundingFrustum)

baudii:
Fixes #6298

### Description of Change

- Refactored the BoundingBox.Contains(BoundingFrustum) method. The method now accurately returns ContainmentType.Disjoint when the frustum is entirely outside the bounding box. 

- Added targeted unit tests to validate these changes, ensuring edge cases are handled where the frustum is near, partially intersecting, or fully outside the box.


baudii:
Ok, I just noticed an issue with my implementation. I'm only testing for corners of frustum. To perform accurate calculations, I think I need to get planes array from the BoundingFrustum class. Can I add a getter for this? Or should I manually check each plane?

Also, there is 1 edge case unclear to me, where the box is entirely within the frustum. What should the method return in this case? Intersects?

squarebananas:
Here is an old reply of mine on the MonoGame forums (https://community.monogame.net/t/help-with-bounding-frustum/17082):
> My understanding is frustum intersection tests in MonoGame are a quick test that may include false positives for performance reasons. If it is vital that no false positives occur another test such as a separating axis theorem test could then be used afterwards but this is much more computationally expensive and isn’t normally used for fast object render culling.

Also thinking a bit more about this just, the expected results should probably look like this:

| **ContainmentType** | **Box.Contains(Frustum) == Frustum.Contains(Box)** |
| :---: | :---: |
| Contains   | False |
| Intersects   | True |
| Disjoint | True |

### Suggestions
1) Improve the method summary tags to explain that there are false positives for performance reasons.
2) Add precise collision methods (or optional parameters to switch between performance/precision algorithms).
3) `BoundingFrustum` with `Ray` is also broken. It might be worthwhile someone reviewing/reworking everything `BoundingFrustum` related at some point.

baudii:
@squarebananas Sorry, I don't think I follow. You mean that the issue #6298 is outdated and this problem actually requires refactoring the entire class?

squarebananas:
> I think I get it now. You mean I should keep the original method with adjusted summary indicating that it is false positive, and make another method that performs more accurate calculations?

Yeah that's it. Or an optional parameter (`BoundingFrustum.Contains(BoundingBox box, bool precisionMode = false)`)
The optional parameter would mean `Frustum.Contains(Box)` would automatically function as before that way.

Generally because `BoundingFrustum` accuracy is expensive and because its main use is fast culling, the most important thing is it doesn't have false negatives (causing visible objects not to be drawn).

First step I think though is labelling/documenting what collision methods should always be accurate and which are inaccurate for rendering purposes only (false positives allowed). Suggestions 2 & 3 are probably better for future consideration.

tomspilman:
> make another method that performs more accurate calculations?

> Or an optional parameter (BoundingFrustum.Contains(BoundingBox box, bool precisionMode = false))

Yeah i think a `ContainsPrecise()` is a better path to take than adding a `bool` to the existing method.

baudii:
@squarebananas Ok, now I follow. It's so misleading how they almost merged this here #7444
I'll add another pull request, fix the summaries

squarebananas:
> @squarebananas Ok, now I follow. It's so misleading how they almost merged this here #7444 I'll add another pull request, fix the summaries

Yeah, I agree. Thanks for that.

@SimonDarksideJ I agree with baudii's comment that PR #7444 isn't the correct solution. While occasional false positives are fine for render performance reasons (allowing a few objects just outside the frustum to be drawn), false negatives would cause visible objects to disappear or flicker. Initram's diagram (https://github.com/MonoGame/MonoGame/pull/7444#issuecomment-766643439) shows that #7444 could give false negative results. With this PR merged, I believe that other PR can now be closed.

Another thing to bear in mind, the comments added to the summary here will also apply to all/some of the other `BoundingFrustum` methods and other bounding type methods with a `BoundingFrustum` argument. A rough estimate is around 14 cases minimum and 28 maximum depending on which bounding types it affects.

baudii:
@squarebananas I edited Summary and adjusted comments to accurately describe the logic (#8566). 
Now it makes even less sense tbh 😆 

1) Method actually never return `Disjoint`. It's either `Contains` (pretty much correctly) or `Intersects` (for everything else)
2) If first point of the `BoundingFrustum` (Intersection of Near/Left/Top planes - `NLT`) is `Disjoint` and rest of them are `Contained` - it returns `Contains`. If the first point `NLT` is contained, but any other point is `Disjoint` - then it returns `Intersects`. 

I'm new to this project, so I don't really know what this mehod was designed for. Probably, some algorithm needs this information to detect wether `BoundingFrustum` is entirely inside the `BoundingBox` or not. But still, why its ok for the very first point to be `Disjoint`? I understand based on what you said that false positives are not that bad, but I feel like this false-positive isn't really neccessary.

squarebananas:
@baudii You're right that doesn't seem right. Although the fix was incorrect, the original maybe more broken than I thought as well. 

If you look back at my first post, then `Box.Contains(Frustum) == Frustum.Contains(Box)` should be the same except the `Contains` case. If all the points within a frustum are within the min/max of a box we know it contains it. The same can't be said for the opposite, as a frustum isn't axis aligned so it's min/max has gaps around it.

Bearing that in mind, the methods should have the same code except for the `Box.Contains(Frustum)` method which should additionally have the all frustum points inside check. If you look at the `Frustum.Contains(Box)` method though, it's code is entirely different.

By the way this article I found just details the challenges of efficient frustum/box collision detection in more detail: https://iquilezles.org/articles/frustumcorrect/

If you're interested in the precise but computationally expensive solution to this, look up "Separating Axis Theorem" used for discovering the existence of a hyperplane of separation between convex objects.

baudii:
@squarebananas Yeah, I read about this theorem when you first mentioned it. But I figured you don't really need it here. Since the box is always aligned with axis you can just compare min/max values with edge points of a frustum. It should be much faster. I'll do PR soon and include benchmark tests.

PR #8569, opened 2024-11-12
Throw Exception In `Content.Load` When `assetName` Is A Rooted Path.

### LINKED ISSUE #8568: `Content.Load<T>` cannot find asset file if the asset name provided is a rooted path
### Prerequisites

### MonoGame Version

MonoGame 3.8.2.1105

### Operating System

Windows

### Description

When using the `ContentManager.Load<T>` method, the `assetName` parameter is assumed to be a path relative to `Content.RootDirectory` minus the extension.  However, if you provide a value that is a rooted path (one beginning with `\\` or `/`), then the `ContnetManager` will not be able to locate the file to load and throw a `FileNotFoundException`

This is happening due to the use of `Path.Combine` in the `ContentManager.OpenStream` method 

https://github.com/MonoGame/MonoGame/blob/fb03c0a90c0a65152feca6b58f2b9d5c9e40ac31/MonoGame.Framework/Content/ContentManager.cs#L377-L383

When `Path.Combine` is used, if the second or later parameter is a rooted path, then it will take precedence over earlier parameters, causing the path generated to be incorrect.

This should be handled in the `Content.Load<T>` method, since the `assetName` parameter is expected to be a relative (non-rooted) path, then a check should be made an an exception thrown with a clearer error message so the developer understands not to provide a rooted path

### Steps to Reproduce

1. Create a new MonoGame project
2. Open the `Content.mgcb` file generated in the project with the MGCB Editor
3. In the MGCB Editor, right-click the *Content* node and choose *Add > New Item*
4. In the New File Dialog, enter the name `font` for the *Name* textbox, then choose *SpriteFont Description (.spritefont)* from the *Type* listbox, then click the *Create* button
5. Save the content project in the MGCB Editor (`CTRL + S` or choosing *File > Save* from the top menu).  
6. Close the MGCB Editor.
7. Open the *Game1.cs* file in the project
8. Locate the `LoadContent` method and add the following to it directly after the `_spriteBatch` initialization occurs

```cs
SpriteFont font = Content.Load<SpriteFont>("\\font");
```

9. Run the project in Debug mode

### Minimal Example Repo

_No response_

### Expected Behavior

The expected behavior is that the asset is loaded

### Resulting Behavior

The resulting behavior is a `ContentLoadException` with an inner exception of `FileNotFoundException` stating that it could not find the file at the path `C:\font.xnb`.

### Files

Two images attached to show that the error occurs when using `\\` and when using `/`

image  
**Figure 1: Example using `\\` at the start of the asset name**  
   
image  
**Figure 2: Example of using `/` at the start of the asset name**

AristurtleDev:
Fixes #8568 

### Description of Change
Adds a `Path.IsRootedPath` check on the `assetName` parameter, and if it is a rooted path, throws a `ContentLoadException` with a clear exception message on why the exception occurred.  

Previous behavior would have allowed a rooted path as the `assetName` parameter which would later throw a `FileNotFoundException` in the `ContentManager.OpenStream` method when attempting to open the stream.  This is due to how `Path.Combine` works, where if the second parameter is a rooted path, it ignores the first parameter.

While technically the `FileNotFoundException` is correct when the stream is attempted to open, the exception doesn't clearly describe to the developer why it occurred.  The `assetName` parameter is defined as being a relative path, so if it is a rooted path, the exception should be thrown earlier to tell the developer they used a rooted path and need to use a relative path instead.

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -345,6 +345,10 @@ namespace Microsoft.Xna.Framework.Content
             {
                 throw new ObjectDisposedException("ContentManager");
             }
+            if(Path.IsPathRooted(assetName))
+            {
+                throw new ContentLoadException("assetName '" + assetName + "' cannot be a rooted (absolute) path. Remove any leading drive letters (e.g. 'C:'), forward slashes or backslashes");
+            }
 
             T result = default(T);
 
--- a/Tests/Framework/Content/ContentManagerTest.cs
+++ b/Tests/Framework/Content/ContentManagerTest.cs
@@ -1,0 +1,21 @@
+﻿using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Content;
+using Microsoft.Xna.Framework.Graphics;
+using NUnit.Framework;
+
+namespace MonoGame.Tests.Content
+{
+    public class ContentManagerTest
+    {
+        [TestCase("C:\\image.png")]
+        [TestCase("\\image.png")]
+        [TestCase("//image.png")]
+        public void ThrowExceptionIfAssetNameIsRootedPath(string assetName)
+        {
+            GameServiceContainer services = new GameServiceContainer();
+            ContentManager content = new ContentManager(services, "Content");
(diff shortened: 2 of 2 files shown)


mrhelmut:
Thanks!

PR #8577, opened 2024-11-17
Add support for multiplying two colors

des54321:
Lets you multiply two colors component-wise via "Color * Color".
Also adds a method that lets you only multiply the alpha channel of a color.

Currently, working with non-premultiplied colors is pretty painful in Monogame and this seeks to help improve that.

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -1749,6 +1749,17 @@ namespace Microsoft.Xna.Framework
 	    {
 	        return new Color((int)(value.R * scale), (int)(value.G * scale), (int)(value.B * scale), (int)(value.A * scale));
 	    }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


mrhelmut (approved):

arosolino:
Love this idea as I already am using something similar a lot.

What do you think how improving the multiplication operator to only use integer arithmetic.

```
       public static Color operator *(Color color1, Color color2)
        {
            return new Color(
                color1.R * color2.R / 255,
                color1.G * color2.G / 255,
                color1.B * color2.B / 255,
                color1.A * color2.A / 255
            );
        }
```

des54321:
Good idea! In hindsight it doesn't make much sense to convert to float for no reason.

tomspilman (approved):

CartBlanche:
@des54321 Thanks again for your contribution!

PR #8580, opened 2024-11-19
Added vertex size check with exception to GraphicsDevice.DrawUserPrimitives()

### LINKED ISSUE #3125: Desired: Exception when VertexDeclaration stride does not match template parameter size in DrawUser(Indexed)Primitives

baudii:
Fixes #3125

### Description of Change

Added vertex size check to GraphicsDevice.DrawUserPrimitives()

Now method throws an exception when the size of a vertexDeclaration.VertexStride is 
less than ReflectionHelpers.FastSizeOf<T>() as suggested here: https://github.com/MonoGame/MonoGame/issues/3125#issuecomment-446291572

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -1280,6 +1280,9 @@ namespace Microsoft.Xna.Framework.Graphics
             if (vertexDeclaration == null)
                 throw new ArgumentNullException("vertexDeclaration");
 
+            if (vertexDeclaration.VertexStride < ReflectionHelpers.FastSizeOf<T>())
+                throw new ArgumentOutOfRangeException("vertexDeclaration", "Vertex stride of vertexDeclaration should be at least as big as the stride of the actual vertices.");
+
             PlatformDrawUserPrimitives<T>(primitiveType, vertexData, vertexOffset, vertexDeclaration, vertexCount);
 
             unchecked


baudii:
I believe the failing tests have nothing to do with my commit, right? It says that Arial font is missing somehow

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -1278,7 +1278,10 @@ public void DrawUserPrimitives<T>(PrimitiveType primitiveType, T[] vertexData, i
                 throw new ArgumentOutOfRangeException("primitiveCount");
 
             if (vertexDeclaration == null)
-                throw new ArgumentNullException("vertexDeclaration");
+                throw new ArgumentNullException("vertexDeclaration");
+
+            if (vertexDeclaration.VertexStride < ReflectionHelpers.FastSizeOf<T>())
+                throw new ArgumentOutOfRangeException("vertexDeclaration", "Vertex stride of vertexDeclaration should be at least as big as the stride of the actual vertices.");
```
tomspilman, line 1284: Is this really true?

Like i could see someone passing a `MyStruct[]` that happens to include multiple verts in it.

Is the more correct check is if the size of the `vertexData` is a multiple of the `vertexDeclaration.VertexStride` ?

@mrhelmut @harry-cpp ?

baudii: I might be unaware of something, but there are 3 methods in this class that take `VertexDeclaration` as an argument. And 2 of them does this exact check. One of them doesn't - the subject one.

Here:
https://github.com/MonoGame/MonoGame/blob/ba5c722a399fa4e192859a7dd17377b20ae4d327/MonoGame.Framework/Graphics/GraphicsDevice.cs#L1237-L1238

And here:
https://github.com/MonoGame/MonoGame/blob/ba5c722a399fa4e192859a7dd17377b20ae4d327/MonoGame.Framework/Graphics/GraphicsDevice.cs#L1157-L1158

SimonDarksideJ:
Any chance you can resolve the comments above @baudii , looking to get the 3.8.3 release out and getting as many PR's ready to go for inclusion

MonoGame.Framework/Graphics/GraphicsDevice.cs
ThomasFOG: Seems safe enough to me.

ThomasFOG:
Seems safe to me.

Issue #8484, opened 2024-09-11
[Help Wanted] Introduce Content.Load<Video>() fallback to load from file when no .xnb exists


SimonDarksideJ:
### Intent

Content.Load<Video> should have the ability to "fall back" and load video files when no corresponding `.xnb` exists.

> This will enable a single path for developers to load videos whether they choose to use the content pipeline or not and greatly simplify the documented process for loading content.

The "Fallback" would need to handle:

* Locating the specific file in question based on the file name (including handling for naming casing) - e.g. (Cutscene.mp4 vs cutscene.mp3 or CutScene.mp3 from "cutscene" as a name.
* Handling sub-paths / folders (e.g. "Video/Menu/cutscene.mp3")
* Determining the correct filetype being loaded (based on supported type for MonoGame, mp4, mov, etc
* Handling the correct functionality to use to load the file and return the `Video` definition

Support is preferred by the community to help to make this happen and unify the path for all developers.

### Motivation

In order to simplify the loading of manual content, rather than having separate code to identify, load, and manage files, the existing Content.Load methods should be enhanced with a fallback measure to engage with content of the specified time.

This ensures a simple path for all MonoGame developers regardless of which path they choose and also reduces the impact should the developer need to switch path and cut out unnecessary code changes.fallback

SimonDarksideJ:
Were you planning to take this up @titanix ?
If not, following the merge of #8582 it is planned to be taken up this month.

titanix:
Hello @SimonDarksideJ, my original intention was to tackle most or all of the remaining similar issues. However, things get really wrong in my life recently and I can't work on that anymore.

SimonDarksideJ:
Thanks for letting us know @titanix and I do wish that things turn around for you again soon.  These are certainly "trying" times at present.

Issue #8483, opened 2024-09-11
[Help Wanted] Introduce Content.Load<Song>() fallback to load from file when no .xnb exists


SimonDarksideJ:
### Intent

Content.Load<Song> should have the ability to "fall back" and load song files when no corresponding `.xnb` exists.

> This will enable a single path for developers to load songs whether they choose to use the content pipeline or not and greatly simplify the documented process for loading content.

The "Fallback" would need to handle:

* Locating the specific file in question based on the file name (including handling for naming casing) - e.g. (Background.mp3 vs background.mp3 or BackGround.mp3 from "background" as a name.
* Handling sub-paths / folders (e.g. "Audio/Menu/background.mp3")
* Determining the correct filetype being loaded (based on supported type for MonoGame, mp3, wav, etc
* Handling the correct functionality to use to load the file and return the `Song` definition

Support is preferred by the community to help to make this happen and unify the path for all developers.

### Motivation

In order to simplify the loading of manual content, rather than having separate code to identify, load, and manage files, the existing Content.Load methods should be enhanced with a fallback measure to engage with content of the specified time.

This ensures a simple path for all MonoGame developers regardless of which path they choose and also reduces the impact should the developer need to switch path and cut out unnecessary code changes.fallback

SimonDarksideJ:
Were you planning to take this up @titanix ?
If not, following the merge of #8582 it is planned to be taken up this month.

Issue #8482, opened 2024-09-11
[Help Wanted] Introduce Content.Load<SoundEffect>() fallback to load from file when no .xnb exists


SimonDarksideJ:
### Intent

Content.Load<SoundEffect> should have the ability to "fall back" `At Runtime` and load sound files when no corresponding `.xnb` exists.

> This will enable a single path for developers to load sounds whether they choose to use the content pipeline or not and greatly simplify the documented process for loading content.
>
> This is not for the MGCB tool as that is used to build `.xnb`s already, this is about **runtime**!

The "Fallback" would need to handle:

* Locating the specific file in question based on the file name (including handling for naming casing) - e.g. (Background.wav vs background.wav or BackGround.wav from "background" as a name.
* Handling sub-paths / folders (e.g. "Audio/Menu/background.wav")
* Determining the correct filetype being loaded (based on what MonoGame supports at runtime), wav, etc
* Handling the correct functionality to use to load the file and return the `SoundEffect` definition

Support is preferred by the community to help to make this happen and unify the path for all developers.

### Motivation

In order to simplify the loading of manual content, rather than having separate code to identify, load, and manage files, the existing Content.Load methods should be enhanced with a fallback measure to engage with content of the specified time.

This ensures a simple path for all MonoGame developers regardless of which path they choose and also reduces the impact should the developer need to switch path and cut out unnecessary code changes.fallback

## See ALso

- [Texture](https://github.com/MonoGame/MonoGame/issues/8481)
- [Song](https://github.com/MonoGame/MonoGame/issues/8483)
- [Video](https://github.com/MonoGame/MonoGame/issues/8484)

> Shaders and Models are excluded for the moment as they have additional dependencies

titanix:
.WAV, .OGG, .MP3, .WMA are the extensions present in the Tests/Assets/Audio folder. So I guess, that's the expected extensions to be supported by the feature? Is the order I wrote fine as the order the search for assets?

SimonDarksideJ:
Were you planning to take this up @titanix ?
If not, following the merge of #8582 it is planned to be taken up this month.

Issue #8481, opened 2024-09-11
[Help Wanted] Introduce Content.Load<Texture>() fallback to load from file when no .xnb exists


SimonDarksideJ:
### Intent

Content.Load<Texture> should have the ability to "fall back" `At Runtime` and load texture files when no corresponding `.xnb` exists.

> This will enable a single path for developers to load textures whether they choose to use the content pipeline or not and greatly simplify the documented process for loading content.
>
> This is not for the MGCB tool as that is used to build `.xnb`s already, this is about **runtime**!

The "Fallback" would need to handle:

* Locating the specific file in question based on the texture name (including handling for naming casing) - e.g. (Background.png vs background.png or BackGround.png  from "background" as a name.
* Handling sub-paths / folders (e.g. "Textures/Menu/background.png")
* Determining the correct filetype being loaded (based on supported type for MonoGame, png, jpg, etc
* Handling the correct functionality to use to load the file and return the `Texture` definition
* Support all the different / variant "Texture" types supported by MonoGame

Support is preferred by the community to help to make this happen and unify the path for all developers.

### Motivation

In order to simplify the loading of manual content, rather than having separate code to identify, load, and manage files, the existing Content.Load methods should be enhanced with a fallback measure to engage with content of the specified time.

This ensures a simple path for all MonoGame developers regardless of which path they choose and also reduces the impact should the developer need to switch path and cut out unnecessary code changes.fallback

## See ALso

- [SoundEffect](https://github.com/MonoGame/MonoGame/issues/8482)
- [Song](https://github.com/MonoGame/MonoGame/issues/8483)
- [Video](https://github.com/MonoGame/MonoGame/issues/8484)

> Shaders and Models are excluded for the moment as they have additional dependencies

squarebananas:
A few random thoughts regarding this:

- Would files still need to be included in the project by using build action "Content - Copy if newer"? If not, would files need manually copying when publishing?
- When using MGCB, if a file was accidentally missed, would the fallback system just mask the issue by loading it unprocessed anyway? Maybe ContentManager could have a fallback disable option.
- Is the idea to remove the need to reference MonoGame.Content.Builder.Task altogether? If it can be optionally referenced it could still provide useful features, even if it only ever handles copying without building behind the scenes.
- New builder task options could be added for:
  - Global build/copy override (allowing quick switching between the two methods).
  - Only copy files contained within Content.mgcb, or alternatively auto generate Content.mgcb from all files within the same folder.

[...]

PR #8582, opened 2024-11-20
Content.Load<Texture>() fallback to load from file when no .xnb exists

titanix:
The PR is a reply to the feature request https://github.com/MonoGame/MonoGame/issues/8481.

For the longest time, I had a class in my game project that looks like the PoC code posted by [squarebananas](https://github.com/squarebananas) in the discussion so I decided to tackle this issue.

### Description of Change

- Adds the capability to load Texture2D from PNG, JPG/JPEG and BMP files without using the content pipeline.
- Fixes the documentation of TitleContainer.OpenStream which was incorrect.
- Adds placeholder comment for related feature requests https://github.com/MonoGame/MonoGame/issues/8482 https://github.com/MonoGame/MonoGame/issues/8483 https://github.com/MonoGame/MonoGame/issues/8484.
- Adds an alternate internal method to ContainerTitle.FromStream that doesn't rely on exceptions for signalling errors.

### What it doesn't do

The fallback discussed in issue 8481 proposed a "smart" way to search for a file when originally not found. For instance, locating it in different subfolders, or using different casing.

I think that "advanced" behavior is distinct from the ability to load a file without the content pipeline. It is also involving more code changes, for a benefit I cannot see, as some rigor is expected anyway from the developer. Adding more magic to the system would likely make it harder to use/debug in the end.

### Testing

Manually tested against my game in DesktopGL on macOS M1.


dellis1972:
@titanix thanks for the PR. 

I think it would be useful to include a unit test for this in something like https://github.com/MonoGame/MonoGame/blob/develop/Tests/Framework/Content/ContentManagerTest.cs#L13. This way we can unit test the functionality on all of our supported platforms (which are covered by the tests anyway).

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -372,18 +374,83 @@ public virtual T Load<T>(string assetName)
...
+                    {
+                        result = reader.ReadAsset<T>();
+                        if (result is GraphicsResource)
+                        {
+                            ((GraphicsResource)result).Name = originalAssetName;
+                        }
+                    }
+                }
+            }
+            catch (ContentLoadException)
```
SimonDarksideJ, line 413: We definitely need some Unit tests as this "should" be a "Stream open Exception" and not a "ContentLoadException" which was a custom exception the framework originally threw but has been moved to line 441.

titanix: OpenStream doesn't throw anything but ContentLoadException since every kind of exception thrown in its body is catched (line 484), and ContentLoadException are thrown in other exception paths as well.

dellis1972:
@titanix mind if we make this PR draft until the tests have been added?

titanix:
Yes, you can manage the thing like you want until I add the tests.

titanix:
@dellis1972 the tests are here! Working on Windows (a good part of tests aren't working out of the box on macOS in the repo).

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -372,18 +374,83 @@ public virtual T Load<T>(string assetName)
...
             // Load the asset.
             result = ReadAsset<T>(assetName, null);
-
             loadedAssets[key] = result;
+
             return result;
 		}
 
+        /// <summary />
+		protected T ReadAsset<T>(string assetName, Action<IDisposable> recordDisposableObject)
```
tomspilman, line 383: In the future @titanix avoid re-ordering functions like this.  It makes it hard to track the actual code changes when the diff no longer lines up in history.

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -70,6 +70,7 @@ public partial class ContentManager : IDisposable
             'l', // Linux
         };
 
+        private static readonly string[] supportedTexture2DExtensions = new string[] { ".bmp", ".jpg", ".jpeg", ".png" };
```
tomspilman, line 73: I would order this:  PNG, JPG, JPEG, BMP...  just because PNG is overwhelmingly the most used format in game dev... might as well have it try it first.

BMP is almost never used, so it makes more sense for it to be last.

titanix: Make sense, I was thinking about the same but in absence of spec I used alphabetic order 😅

MonoGame.Framework/Content/ContentManager.cs
CartBlanche: @titanix could you please move it back.

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -506,6 +540,27 @@ private ContentReader GetContentReaderFromXnb(string originalAssetName, Stream s
 
             return reader;
         }
+    
+        object LoadTexture2DHeuristically(string assetName)
```
CartBlanche, line 544: This will either return `Texture2D` or null correct? If so, could this return `Texture2D` instead?

MonoGame.Framework/Content/ContentManager.cs
CartBlanche: Thanks for changing it.

MonoGame.Framework/Content/ContentManager.cs
```diff
@@ -506,6 +540,27 @@ private ContentReader GetContentReaderFromXnb(string originalAssetName, Stream s
 
             return reader;
         }
+    
+        Texture2D LoadTexture2DHeuristically(string assetName)
```
dellis1972, line 544: Any chance we can change the name of this method to something a little more straight forward?
`LoadTexture2DFromImageFile` or something like that? 

Its a small thing, however having simple method names makes things easier to maintain IMHO.

CartBlanche: Thanks!

dellis1972:
This is looking good. @titanix feel free to mark it ready for review if you think its good 😄

CartBlanche:
@titanix Thanks again for the PR! We seem to be very close :).
I just cloned your repo and built it locally and managed to repro the CI crash. Seems the crash is being caused by changing that Test to inherit from `GraphicsDeviceTestFixtureBase`. 
Just to confirm it's not just happening for me and CI, could you please load the `MonoGame.Framework.WindowsDX.sln` in your IDE and build the `MonoGame.Tests.WindowsDX` project. 
This should create a `MonoGame.Tests.dll` file in your `MonoGame\Artifacts\Tests\WindowsDX\{Configuration}` directory. 
Open up a command line there and run:
`dotnet test MonoGame.Tests.dll`

I know that from the IDE is Just Works(TM). But we really need the unit tests to pass on the CI too.

Could you please double check that this class doesn't require any additional changes, due to the new base class.

Note: Your tests may not need `new ContentManager content = new ContentManager(game.Services);` in any of them because GraphicsDeviceTestFixtureBase.SetUp() creates a `content` variable for all derived classes to use. But probably worth retesting to be sure.

CartBlanche:
@titanix The other alternative would be not to inherit from `GraphicsDeviceTestFixtureBase` and modify the tests to work similarly to `ThrowExceptionIfAssetNameIsRootedPath()`, perhaps, and see if that gets better results?

titanix:
@CartBlanche There is indeed an issue with DX tests from inheriting `GraphicsDeviceTestFixtureBase`. However, inheriting this class was the way I found to have a working GraphicsDeviceService which is required deep down the call stack to ReadAsset. I need that for the tests to be meaningful.

The test `ThrowExceptionIfAssetNameIsRootedPath` is actually not very good: it will succeed to almost whatever errors occurs even if unrelated to the non-existing of a file because exceptions are catched and rethrown as ContentLoadException. If I rewrite my tests as negative, they will succeed because the graphic device service related exception will create a ContentLoadException, but that's not really testing the new behavior being implemented.

CartBlanche:
@titanix thanks for clarification.

titanix:
@CartBlanche The issue persists when instantiating and disposing the objects I need. The issue arise from an unsafe method call `SharpDX.XAudio2.Voice.DestroyVoice` to an unmanaged method. It's documentation says that "DestroyVoice waits for the audio processing thread to be idle, so it can take a little while".

The audio tests pass when ran alone. Others tests pass as well. This is clearly an issue in the test runner environment and how it deals with resources. I tried various things unsuccessfully.

How about solving the issue by adjusting the CI? I mean, splitting the current 
`dotnet MonoGame.Tests.dll`
into
```
dotnet MonoGame.Tests.dll --filter MonoGame.Tests.Audio
dotnet MonoGame.Tests.dll --filter FullyQualifiedName!~MonoGame.Tests.Audio
```

tomspilman (approved):

CartBlanche:
@titanix Yes, could you please update the CI and split out the DirectX test to something like:

- name: Run DirectX Audio Tests
        run: dotnet test --filter Category=Audio MonoGame.Tests.dll 
        env:
          CI: true
        working-directory: tests-windowsdx
        if: runner.os == 'Windows'
        
- name: Run DirectX All Tests Except Audio
        run: dotnet test --filter Category!=Audio MonoGame.Tests.dll 
        env:
          CI: true
        working-directory: tests-windowsdx
        if: runner.os == 'Windows'
        
Please also add a comment, in the yaml, explaining why we've had to split these out, because no one will remember that down the line :)

Once this goes green, I think we'll be ready to merge.

titanix:
@CartBlanche Thanks for the code snippet! I updated the CI and hope it'll be ok this time 😌.

dellis1972 (approved):
Looks good. It might be worth raising an issue around the Audio Destroy thing, it should be something that gets fixed

CartBlanche (approved):
LGTM

CartBlanche:
@titanix thanks for your patience in helping us get this across the line. I imagine it might have been a little frustrating from your end, but the team is grateful for your contribution. Merging :)

titanix:
Thanks everyone! @CartBlanche no it's ok, the feedback was warranted and it's normal to aim to a given level of quality.

PR #8583, opened 2024-11-20
Fix Android MediaPlayer Song implementation.

dellis1972:
Commit 5e918e91 introduced a bug on Android where
song's would not place. This was because it was using the `name` of the song rather than the `filename`.

Fix up the Android implementation so that it uses the right field when accessing the audio.


CartBlanche (approved):
LGTM

PR #8590, opened 2024-11-26
Remove temp buffer alloc get data when reading full texture

MutsiMutsi:
When GetData is called on an OpenGL Texture2D a temporary buffer is allocated which causes heap allocations.
In scenarios where you request the entire texture (rect param is not given, or rect param is of full extend (0,0,width,height). We do not have to allocate a temporary buffer, and we can also skip Array.Copy in this case.

### Description of Change
The proposed implementation checks for this specific condition, and then proceeds to fill the given T[]Data directly with GetTexImage results.


MonoGame.Framework/Platform/Graphics/Texture2D.OpenGL.cs
```diff
@@ -264,6 +264,13 @@ private void PlatformGetData<T>(int level, int arraySlice, Rectangle rect, T[] d
                         Array.Copy(temp, tempStart, data, dataStart, tRectWidth);
                     }
                 }
+
+            //If we want the full texture rect == (0,0,width,height), we dont have to allocate a temp buffer to copy from.
+            else if (rect.X == 0 && rect.Y == 0 && this.width == rect.Width && this.height == rect.Height)
+            {
+                GL.GetTexImage(TextureTarget.Texture2D, level, glFormat, glType, data);
+                GraphicsExtensions.CheckGLError();
+            }
```
dellis1972, line 273: The indentation of this code does not seem to match the code around it. It makes it difficult to read.

dellis1972 (changes requested):

MonoGame.Framework/Platform/Graphics/Texture2D.OpenGL.cs
MutsiMutsi: @dellis1972 oops! Thanks for checking 👌.
I've pushed the format fix, additionally ive also removed the duplicate code of pixelToT in each branch outside of the branching logic, and also check the pixelToT == 1 to ensure that we can safely retrieve all data without further processing. We want to ensure that if we use this highly simplified condition we have the same number of bytes in the surface data and type of T data given.

squarebananas:
Tried this just and it works great.

Here are some rough figures when using GetData every frame for comparison purposes.

|  | **Frames per second** | **GC Rate (including game code)** |
| :---: | :---: | :---: |
| Before | 70 fps | 4.4 GB |
| After | 220 fps  | 32 KB |
| Result |↗️✅ | ↘️✅ |

And compared to other platforms:
|  | **Frames per second** | **GC Rate (including game code)** |
| :---: | :---: | :---: |
| DesktopGL | 220 fps | 32 KB |
| WindowsDX | 110 fps | 8 KB |
| DesktopVK | 3 fps  | 5 KB |

### Comments
- One issue I almost didn't see, the `startIndex` argument should offset the destination pointer (for example retrieving bytes into a large shared buffer containing existing resources). For now maybe just check `startIndex == 0`. It is possible to fix this in `OpenGL.GetTexImage`, however other things are unimplemented in that anyway (see below).

### Other General Notes (Not Directly PR Related)
- `Texture2D.GetData<T>` on DesktopGL doesn't implement all arguments such as `arraySlice` and `elementCount`.
- This PR wouldn't break getting back supported convertible `SurfaceFormat`s returned, but it seems it's already broken in MonoGame anyway. From a breakpoint I can see OpenGL can correctly retrieve data when using `Texture2D.GetData<PackedVector.Alpha8>` with a `SurfaceFormats.Color` texture. MonoGame DesktopGL doesn't handle the different byte sizes though. (Looks like DesktopVK won't work either because it doesn't even pass the `SurfaceFormat` requested to the native side).

MutsiMutsi:
@squarebananas thats awesome thanks a bunch!
With regards to startIndex I think adding a check for it being zero is appropriate, since this optimized read should only occur when you grab the entire texture data, so always starting from 0 reading to the end.

squarebananas:
> With regards to startIndex I think adding a check for it being zero is appropriate, since this optimized read should only occur when you grab the entire texture data, so always starting from 0 reading to the end.

That's what I was thinking, until I realised the source's start is already determined by the `Rectangle` values and then looked at this line in the existing method code:

```
var dataStart = startIndex + r * tRectWidth;
Array.Copy(temp, tempStart, data, dataStart, tRectWidth);
```

So `startIndex` relates to the destination buffer. Also the equivalent call to SharpDX for WindowsDX has a parameter summary which states _"offset: The offset in the array to write to"_ and the value sent to it is `startIndex`".

Like I mentioned it could be fixed properly but the `elementCount` isn't working either, and fixing that is kind of outside the purpose of this PR. The `nint` that would need amending though is the `pixelsPtr` value here: https://github.com/MonoGame/MonoGame/blob/4b996cf133ab446bdf17577eb0e9542735821ad4/MonoGame.Framework/Platform/Graphics/OpenGL.cs#L1663

tomspilman (approved):

PR #8593, opened 2024-12-02
Add MonoGame.Templates submodule

CartBlanche:
### Moved the Project Templates to separate repo
- [x] Remove Templates
- [x] Point to new [MonoGame.Templates repo](https://github.com/MonoGame/MonoGame.Templates/) 
- [x] Modify build the right template project


.gitmodules
```diff
@@ -22,3 +22,6 @@
 [submodule "src/monogame/external/vma"]
 	path = native/monogame/external/vma
 	url = https://github.com/LunarG/VulkanMemoryAllocator.git
+[submodule "ThirdParty/MonoGame.Templates"]
+	path = ThirdParty/MonoGame.Templates
```
dellis1972, line 26: can we not put it in ThirdParty? Its not a third part its us. 
maybe `external` or something like that.

dellis1972: ThirdParty is the same to me, if you have a better folder idea to put it in that would be good 👍

dellis1972 (approved):

PR #8607, opened 2024-12-20
[Pipeline] Use AssimpNetter nuget for now

harry-cpp:
For now, using the prebuilt binaries from the wrapper itself instead of our prebuilt binaries, until we get the time to either make a wrapper or setup the native mgcp project.


CartBlanche (approved):
Worked for me on Mac. 🎆

Issue #8104, opened 2023-12-14
Make MonoGame trimming/AOT compatible as-is


mrhelmut:
### Intent

Game developers are likely to be willing to produce apps using either the ```PublishTrimmed``` or ```PublishAot``` build options (the later being the expected console path). Right now, this requires projects using MonoGame to root the MonoGame assembly so that commonly dynamically loaded types (e.g. typically anything loaded with the Content Pipeline) by adding this to their .csproj definitions:

```xml
  <ItemGroup>
    <TrimmerRootAssembly Include="MonoGame.Framework" />
  </ItemGroup>
```

There's a couple of problems with this approach:
- Even though this can be partly fixed by adding this to project templates, it is not the intended .NET experience and is error-prone if anyone is creating a project from scratch or editing their .csproj
- This doesn't make MonoGame to be trimmable, it forces the IL compiler to _not_ trim MonoGame at tall, which is a bad fix in regard to what users might expect from trimming

We should try building MonoGame with ```<EnableTrimAnalyzer>true</EnableTrimAnalyzer>```, verify if any IL warnings related to trimming or AOT scenarios shows up, and fix them.

Once those warnings are fixed, we will be able to mark the MonoGame library as ```<IsAotCompatible>true</IsAotCompatible>``` (which also sets ```<IsTrimmable>true</IsTrimmable>```). This way users will know that MonoGame has been tailored to support trimming and AOT scenarios and won't have to add anything to their .csproj.

As a side note, users willing to have their game running on consoles should also aim at fixing any warnings generated by enabling ```<EnableTrimAnalyzer>true</EnableTrimAnalyzer>``` because console games are built using AOT compilation.

### Motivation

test

mrhelmut:
One of the warnings we should fix (this one is from DesktopGL builds):
> ILC : warning IL3000: MonoGame.Framework.Utilities.FuncLoader.LoadLibraryExt(String): 'System.Reflection.Assembly.Location.get' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.

Mindfulplays:
> (e.g. typically anything loaded with the Content Pipeline) are properly linked by adding this to their .csproj definitions:

[...]

Issue #8622, opened 2025-01-07
 Custom content and AOT


pfferrao:
### Prerequisites

- [x] I have verified this issue is present in the `develop` branch
- [X] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

MonoGame 3.8.2.1105

### Which MonoGame platform are you using?

N/A

### Operating System

Windows

### Description

Loading a custom content type (base solely on XNA types) fails under AOT.

### Steps to Reproduce

Using the following custom serializer (very similar to one of the original Texture atlas from XNA):

```
[ContentSerializerRuntimeType("target name space, assembly")]
public class NamedTextureAtlasContent {
    public Texture2DContent Texture = new Texture2DContent();
    public Dictionary<string, Rectangle> Sources = new Dictionary<string, Rectangle>();
}
```

Loaded as:

```
public class NamedTextureAtlas {
    public Texture2DTexture;
    public Dictionary<string, Rectangle> Sources;
}
```

### Expected Behavior

```
Content.Load<NamedTextureAtlas>("<name>");
```

To load the custom resource/asset

### Resulting Behavior

Under AOT:
```
Unhandled Exception: System.NotSupportedException: 'Microsoft.Xna.Framework.Content.DictionaryReader`2[System.String,Microsoft.Xna.Framework.Rectangle]' is missing native code or metadata. This can happen for code that is not compatible with trimming or AOT. Inspect and fix trimming and AOT related warnings that were generated when the app was published. For more information see https://aka.ms/nativeaot-compatibility
   at System.Reflection.Runtime.General.TypeUnifier.WithVerifiedTypeHandle(RuntimeConstructedGenericTypeInfo, RuntimeTypeInfo[]) + 0x75
   at System.Reflection.TypeNameParser.GetType(String, Func`2, Func`4, Boolean, Boolean, Boolean, String) + 0xb8
   at Microsoft.Xna.Framework.Content.ContentTypeReaderManager.LoadAssetReaders(ContentReader reader) + 0x9e8
   at Microsoft.Xna.Framework.Content.ContentReader.InitializeTypeReaders() + 0x40
```

The issue is that we have no way to provide the information that the de-serialization will need the following type
```
DictionaryReader<string, Rectangle>
```
[...]

PR #8623, opened 2025-01-08
Making MonoGame AOT/trimming compliant

ThomasFOG:
I started addressing #8104 and those changes should make MonoGame mostly safe in AOT contexts (it fixes all the errors of loading assets and removes the need to force MonoGame to not be trimmed).

Working on this actually made visible more AOT issues. I fixed the most trivial IL warnings but some of them persists. 
Also, since BRUTE hasn't fully retired yet, I'm trying to write this in a ```net452``` backward compatible way (some of the newer and better reflection stuff isn't available on older .NET Frameworks).

Any help is welcome!

Here are my notes on the remaining issues (none really are blockers, just enhancing the AOT experience and avoid misleading warnings for users):

- [x] ```ReflectiveReader```, ```DictionaryReader```, ```ArrayReader```, and ```MultiArrayReader``` need to be flagged as not compatible with AOT with the ```RequiresDynamicCode``` attribute. Those types are generic and can't be made safe for AOT. The warning inside the attribute must encourage users to rewrite their content serialization in a way that doesn't rely on reflection is they wish to use AOT.

- [ ] WindowsDX will [fail loading an embedded icon](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Windows/WinFormsGameWindow.cs#L261) because ```assembly.Location``` is always null in AOT contexts.

Those ones aren't really issues and will likely not trigger runtime errors. We could just disable warnings on those, but it would be nice to try to fix them in a clean way to avoid unforeseen consequences:

- [ ] [VectorConversion](https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Design/VectorConversion.cs) uses reflection to convert to/from ```IPackedVector```. This needs to be rewritten reflection-free. We can also assume that ```IPackedVector``` is never trimmed and just disable the warnings, but that's an easy fix and likely less error-prone to not rely on disabling warnings.

- [ ] There are a couple of instances of type instance creation that we should try to get rid of if possible:
https://github.com/MonoGame/MonoGame/blob/2e72fcfa4144ea5bc413963af8f97a2d5e902338/MonoGame.Framework/Content/ContentReaders/MultiArrayReader.cs#L36
https://github.com/MonoGame/MonoGame/blob/2e72fcfa4144ea5bc413963af8f97a2d5e902338/MonoGame.Framework/Graphics/Vertices/VertexDeclaration.cs#L214

- [ ] ```ContentTypeReaderManager``` has a warning here (might be fixable):
https://github.com/MonoGame/MonoGame/blob/2e72fcfa4144ea5bc413963af8f97a2d5e902338/MonoGame.Framework/Content/ContentTypeReaderManager.cs#L154

- [ ] ```ContentManager``` has 2 warnings here (can be disabled, this will likely never fail, and this is only useful for Android):
https://github.com/MonoGame/MonoGame/blob/2e72fcfa4144ea5bc413963af8f97a2d5e902338/MonoGame.Framework/Content/ContentManager.cs#L592

- [ ] DesktopGL has ```VertexBuffer.GetBufferData()``` which might fail in AOT depending on the vertex declaration (we can't fix this, but it would be best to flag this method as incompatible, but like other warnings, this should actually be fine):
https://github.com/MonoGame/MonoGame/blob/2e72fcfa4144ea5bc413963af8f97a2d5e902338/MonoGame.Framework/Platform/Graphics/Vertices/VertexBuffer.OpenGL.cs#L92

### DIFF
--- a/MonoGame.Framework/Content/ContentReaders/AlphaTestEffectReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/AlphaTestEffectReader.cs
@@ -6,6 +6,7 @@ using Microsoft.Xna.Framework.Graphics;
 
 namespace Microsoft.Xna.Framework.Content
 {
+    [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
     class AlphaTestEffectReader : ContentTypeReader<AlphaTestEffect>
     {
         protected internal override AlphaTestEffect Read(ContentReader input, AlphaTestEffect existingInstance)
--- a/MonoGame.Framework/Content/ContentReaders/AlphaTestEffectReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/AlphaTestEffectReader.cs
@@ -6,7 +6,9 @@ using Microsoft.Xna.Framework.Graphics;
 
 namespace Microsoft.Xna.Framework.Content
 {
+    #if !NET45
     [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
+    #endif
     class AlphaTestEffectReader : ContentTypeReader<AlphaTestEffect>
     {
         protected internal override AlphaTestEffect Read(ContentReader input, AlphaTestEffect existingInstance)
--- a/MonoGame.Framework/Content/ContentReaders/AlphaTestEffectReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/AlphaTestEffectReader.cs
@@ -6,9 +6,7 @@ using Microsoft.Xna.Framework.Graphics;
 
 namespace Microsoft.Xna.Framework.Content
 {
-    #if !NET45
     [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
-    #endif
     class AlphaTestEffectReader : ContentTypeReader<AlphaTestEffect>
     {
         protected internal override AlphaTestEffect Read(ContentReader input, AlphaTestEffect existingInstance)
--- a/MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
@@ -8,8 +8,13 @@ using MonoGame.Framework.Utilities;
 
 namespace Microsoft.Xna.Framework.Content
 {
+    /// <summary>
+    /// This type is not meant to be used directly by MonoGame users.
+    /// Its purpose is to allow to work-around AOT issues when loading assets with the ContentManager fail due to the absence of runtime-reflection support in that context (i.e. missing types due to trimming and inability to statically discover them at compile-time).
+    /// If ContentManager.Load() throws an NotSupportedExeception, the message should provide insights on how to fix it.
+    /// </summary>
     [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
-    internal class DictionaryReader<TKey, TValue> : ContentTypeReader<Dictionary<TKey, TValue>>
+    public class DictionaryReader<TKey, TValue> : ContentTypeReader<Dictionary<TKey, TValue>>
     {
         ContentTypeReader keyReader;
 		ContentTypeReader valueReader;
--- a/MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
@@ -8,7 +8,7 @@ using MonoGame.Framework.Utilities;
 
 namespace Microsoft.Xna.Framework.Content
 {
-
+    [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
     internal class DictionaryReader<TKey, TValue> : ContentTypeReader<Dictionary<TKey, TValue>>
     {
         ContentTypeReader keyReader;
--- a/MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
@@ -8,9 +8,7 @@ using MonoGame.Framework.Utilities;
 
 namespace Microsoft.Xna.Framework.Content
 {
-    #if !NET45
     [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
-    #endif
     internal class DictionaryReader<TKey, TValue> : ContentTypeReader<Dictionary<TKey, TValue>>
     {
         ContentTypeReader keyReader;
--- a/MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
+++ b/MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
@@ -8,7 +8,9 @@ using MonoGame.Framework.Utilities;
 
 namespace Microsoft.Xna.Framework.Content
 {
+    #if !NET45
     [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
+    #endif
     internal class DictionaryReader<TKey, TValue> : ContentTypeReader<Dictionary<TKey, TValue>>
     {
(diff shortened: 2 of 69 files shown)


MonoGame.Framework/Content/ContentReaders/AlphaTestEffectReader.cs
```diff
@@ -6,6 +6,9 @@
 
 namespace Microsoft.Xna.Framework.Content
 {
+    #if !NET45
+    [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
+    #endif
     class AlphaTestEffectReader : ContentTypeReader<AlphaTestEffect>
```
tomspilman, line 12: A cleaner way to do this is create a "fake" `DynamicallyAccessedMembersAttribute` for older .NET versions within our assembly:

```
#if NET45
namespace System.Diagnostics.CodeAnalysis
{
    [Flags]
    public enum DynamicallyAccessedMemberTypes
    {
    } 
    
    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Method | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
    public sealed class DynamicallyAccessedMembersAttribute : Attribute
    {
        public DynamicallyAccessedMembersAttribute(DynamicallyAccessedMemberTypes memberTypes)
        {
            MemberTypes = memberTypes;
        }

        public DynamicallyAccessedMemberTypes MemberTypes { get; }
    }
}
#endif
```

Then you don't need to put all the `#if !NET45` around all the attribute usages.

I think we did something like this before for communicating some AOT stuff specifically for Mono platforms.

ThomasFOG: Ah yes, good suggestion. Will do that.

ThomasFOG:
Actually, we need to flag ```ReflectiveReader```, ```DictionaryReader```, ```ArrayReader```, and ```MultiArrayReader``` as not compatible with AOT. Generics are where MonoGame can't do magic and will fail in most cases.

We need to add the ```RequiresDynamicCodeAttribute``` to those types with a warning in it encouraging to rewrite content so that it doesn't depend on reflection if the user wishes to use AOT.

IceReaper:
Couldnt the reflection part be replaced by a source generator which simply generates a loader class which calls the right methods depending on the input? While reflection executes on runtime, this could move the logic into compile time, hence trimming or AOT properly resolving the usages?

MonogameGuy:
@IceReaper 

> Couldnt the reflection part be replaced by a source generator which simply generates a loader class which calls the right methods depending on the input? While reflection executes on runtime, this could move the logic into compile time, hence trimming or AOT properly resolving the usages?

Yes, but that would close the door on external custom processors and require source compilation of the framework for any custom importers.  In other words it could not be built and packaged independent of the framework.

ThomasFOG:
We need to keep this problem simple and my personal view on this, is that we shouldn't overengineer something just so that runtime-reflection can work in this specific occurrence. It's a lot of effort for little gains.

My view is that we should teach to users what AOT is and that it's not a magic switch, it's a typical trade-off feature and we need to set the expectations right. Sometime hiding complexity isn't beneficial, and users will typically want to know what works and what doesn't. No just in regard to MonoGame, but in regard to the other code and libraries that they use.

For instance, if we make MonoGame totally reflection-proof, and someone is using NewtonSoft.JSON (which is notoriously crashing in AOT -- unless you're using one of those hacked forks for Unity), we can totally expect people dropping by and start complaining that MonoGame broke their JSON while this has nothing to do with MonoGame. On the opposite, if we raise warnings at compile-time when incompatible types are used and redirect to a documentation explaining why this happens and what to do, it likely has much more value for both users and maintainers.

@pfferrao mentioned a few relevant ideas in #8622, like making the generic readers public so that they could be manually added with ```ContentTypeReaderManager.AddTypeCreator()``` from user code. It's a very simple fix that is to me a good trade-off.

ThomasFOG:
Okay, so here's the deal.

I made the generic type readers public (not every readers), and I added how to fix content loading issues in the ```NotSupportedException``` thrown when ```ContentManager.Load()``` fails due to AOT. The message invites the user to use ```ContentTypeReaderManager.AddTypeCreator()``` in their ```Game``` constructor so that the compiler knows the expected type.

I don't have the time to look into the other AOT warnings, but all of them should be safe to ignore (they all load types that should be discoverable at compile-time). I haven't silenced them so that we can keep track of them should any problem arise or contribution in that regard be made. The caveat to this is that users doing an AOT compilation will see in their build log something like "Microsoft.Xna.Framework has generated IL warnings" without actually displaying the warnings anywhere. Just a minor inconvenience in the user experience that can be ignored, but it can be misleading to users trying to fix any and all IL warnings.

@pfferrao @tomspilman thoughts? (mac is failing but it seems related to the github runner initialization)

The one actual issue that persists: icons will not load on WindowsDX. We need to load the embedded resource differently without reflection.

pfferrao:
> Okay, so here's the deal.
> 
> I made the generic type readers public (not every readers), and I added how to fix content loading issues in the `NotSupportedException` thrown when `ContentManager.Load()` fails due to AOT. The message invites the user to use `ContentTypeReaderManager.AddTypeCreator()` in their `Game` constructor so that the compiler knows the expected type.
> 
> I don't have the time to look into the other AOT warnings, but all of them should be safe to ignore (they all load types that should be discoverable at compile-time). I haven't silenced them so that we can keep track of them should any problem arise or contribution in that regard be made. The caveat to this is that users doing an AOT compilation will see in their build log something like "Microsoft.Xna.Framework has generated IL warnings" without actually displaying the warnings anywhere. Just a minor inconvenience in the user experience that can be ignored, but it can be misleading to users trying to fix any and all IL warnings.
> 
> @pfferrao @tomspilman thoughts? (mac is failing but it seems related to the github runner initialization)
> 
> The one actual issue that persists: icons will not load on WindowsDX. We need to load the embedded resource differently without reflection.

Sorry for the delay, just tested these changes, it address the issues I had raised on #8622, and probably more (: 

On more suggestion would be to pre-add these to the `ContentTypeReaderManager`:
```
ContentTypeReaderManager.AddTypeCreator("System.Collections.Generic.List<System.Char>", () => new ListReader<char>());                           // for strings
ContentTypeReaderManager.AddTypeCreator("System.Collections.Generic.List<Microsoft.Xna.Framework.Vector3>", () => new ListReader<Vector3>());    // arrays of vec3
ContentTypeReaderManager.AddTypeCreator("System.Collections.Generic.List<Microsoft.Xna.Framework.Vector2>", () => new ListReader<Vector2>());    // arrays of vec2
ContentTypeReaderManager.AddTypeCreator("System.Collections.Generic.List<System.Int32>", () => new ListReader<int>());                           // arrays of integers
ContentTypeReaderManager.AddTypeCreator("System.Collections.Generic.List<System.Single>", () => new ListReader<float>());                        // arrays of floats
// same for the common arrays types that usually end-up on custom content
```
Here I don't know what are the common types that usually end-up on custom content, but at least strings and some arrays of other monogame related types are expected.

pfferrao:
> The one actual issue that persists: icons will not load on WindowsDX. We need to load the embedded resource differently without reflection.

Not sure if that's an issue at the moment, as `WindowsDX` requires `UseWindowsForms` to be true which forces us to set `PublishTrimmed` to false. And for `AOT` we need `PublishTrimmed` to be true...
```
.nuget\packages\microsoft.dotnet.ilcompiler\8.0.11\build\Microsoft.NETCore.Native.Publish.targets(55,5): Error : PublishTrimmed is implied by native compilation and cannot be disabled.
```

ThomasFOG:
Providing we can run the build against this PR, it's good to go. I'm available if need to be to get it merged for 3.8.3.

dellis1972:
@ThomasFOG can you rebase this so we can run the checks again please :D

ThomasFOG:
Rebased. ✔️ 

Let's see if it builds properly.

SimonDarksideJ (approved):

MonoGame.Framework/Content/ContentReaders/DictionaryReader.cs
```diff
@@ -7,9 +7,14 @@
...
-{
-
-    internal class DictionaryReader<TKey, TValue> : ContentTypeReader<Dictionary<TKey, TValue>>
+{
+    /// <summary>
+    /// This type is not meant to be used directly by MonoGame users.
+    /// Its purpose is to allow to work-around AOT issues when loading assets with the ContentManager fail due to the absence of runtime-reflection support in that context (i.e. missing types due to trimming and inability to statically discover them at compile-time).
+    /// If ContentManager.Load() throws an NotSupportedExeception, the message should provide insights on how to fix it.
+    /// </summary>
+    [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.All)]
```
dellis1972, line 16: @ThomasFOG  I this has been merged, but any particular reason why we are using the Full class namespace there rather than adding a `using` clause?

ThomasFOG: No particular reason. Just my personal coding style which is "if it's used only once, avoid a using".

PR #8639, opened 2025-02-08
Fix CI by using xcode 16.2 and latest .net

dellis1972:
Requires https://github.com/MonoGame/monogame.github.io/pull/198 to fix the Linux builds


SimonDarksideJ:
Still failing @dellis1972 , even with the related PR merged.  (ignore the Windows fail, I lost power during the build)

dellis1972:
@SimonDarksideJ the website had not been published so the shell script was still out of date.

PR #8643, opened 2025-02-12
Re-add StorageContainer classes.

### LINKED ISSUE #8518: Restore StorageContainer API
### Intent

We need to restore the old XNA StorageContainer APIs to provide a more generate way for developers to work with save data.

See the version last released in MG 3.5:

https://github.com/MonoGame/MonoGame/blob/v3.5/MonoGame.Framework/Storage/StorageContainer.cs

### Motivation

We need a cross platform way for developers to access save data on PCs, consoles, and mobile.

While the `StorageContainer` has some old ideas (the `AsyncCallback` system) in general it does abstract file access well enough hide the complexity of console save game access.

Some extra features and documentation could complete things and make it a good solution.

What would need to happen:

 - Restore old classes for `StorageContainer`.
 - Convert it to use the partial class platform abstraction calls.
 - Refresh the implementation for PC saves that feels good for players and devs.
 - Be sure the PC implementation plays well with PC storefronts like Steam and GoG.
 - Implement Android and iOS considering current mobile save storage systems.
 - Implement for Switch, Xbox, and PlayStation.
 - Write new documentation to guide developers and suggest best practices with saves.

This could be made into one or more bounties if we need to accelerate the development.

CartBlanche:
Fixes #8518 

### Description of Change

Re-implement StorageContainer classes and make them usable in a modern scenarios

### Contributor Declaration


MonoGame.Framework/Storage/StorageContainer.cs
```diff
@@ -0,0 +1,440 @@
+// MonoGame - Copyright (C) MonoGame Foundation, Inc
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using Microsoft.Xna.Framework;
+using System;
+using System.IO;
+
+#if WINDOWS_STOREAPP
```
tomspilman, line 9: Can we remove this now?

MonoGame.Framework/Storage/StorageContainer.cs
```diff
@@ -0,0 +1,440 @@
...
+    /// <summary>
+    /// Contains a logical collection of files used for user-data storage.
+    /// </summary>			
+    /// <remarks>MSDN documentation contains related conceptual article: http://msdn.microsoft.com/en-us/library/bb200105.aspx#ID4EDB</remarks>
+    public class StorageContainer : IDisposable
+    {
+        internal readonly string _storagePath;
+        private readonly StorageDevice _device;
+        private readonly string _name;
+        private readonly PlayerIndex? _playerIndex;
```
tomspilman, line 64: Do we want to commit the old implementation first?

Or should we gut this and/or start the new implementation?

MonoGame.Framework/Storage/StorageDevice.cs
```diff
@@ -0,0 +1,133 @@
+using Microsoft.Xna.Framework;
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Xna.Framework.Storage
+{
+    public delegate Task<StorageDevice> ShowSelectorAsync(PlayerIndex? player, int sizeInBytes, int directoryCount, CancellationToken cancellationToken = default);
+    public delegate Task<StorageContainer> OpenContainerAsync(string displayName, CancellationToken cancellationToken = default);
```
tomspilman, line 10: We should kill these delegates.

MonoGame.Framework/Storage/StorageDevice.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    StorageDeviceHelper.Path = StorageRoot;
+                    return StorageDeviceHelper.TotalSpace;
+                }
+            }
+        }
+
+        private string GetDevicePath => _storageContainer?._storagePath ?? StorageRoot;
+
+        public static event EventHandler<EventArgs> DeviceChanged;
```
tomspilman, line 75: Don't think we need this anymore either.

MonoGame.Framework/Storage/StorageDeviceHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+
+using System;
+
+namespace Microsoft.Xna.Framework.Storage
+{
+    /// <summary>
+    /// This is a helper class to obtain the native file system information.
+    /// </summary>
+    /// <remarks>Look at the Mac implementation.</remarks>
+    internal class StorageDeviceHelper
```
tomspilman, line 13: This should all be removed and be the responsibility of the native code.

CartBlanche:
@tomspilman everything there was just placeholder to have a compilable API. I'll gut things as per your suggestion.

MonoGame.Framework/Storage/StorageContainer.cs
CartBlanche: Definitely. This was from the old code.

MonoGame.Framework/Storage/StorageContainer.cs
CartBlanche: There just for the API, internal will 100% change.

MonoGame.Framework/Storage/StorageDevice.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        public Task<StorageContainer> OpenContainerAsync(string displayName, CancellationToken cancellationToken = default)
+        {
+            return Task.Run(() => Open(displayName), cancellationToken);
+        }
+
+        private StorageContainer Open(string displayName)
```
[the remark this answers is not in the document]
CartBlanche: @tomspilman If you look above, if's being called by the public API OpenContainerAsync on line 79, just above it.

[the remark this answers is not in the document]
tomspilman: Yeah i think we want to remove those `Async` ones and leave this.

Maybe it should be:

```
public StorageContainer TryOpen(string displayName);

public StorageContainer OpenOrCreate(string displayName);

public bool DeleteContainer(string displayName);
```

That gives people a way to open a container only if it exists, also to always create a new one or open a container, also it has an option to delete the container without needing to open it first.

MonoGame.Framework/Storage/StorageContainer.cs
```diff
@@ -0,0 +1,291 @@
...
+        public bool FileExists(string file)
+        {
+            if (string.IsNullOrEmpty(file))
+                throw new ArgumentNullException("Parameter file must contain a value.");
+
+            // relative so combine with our path
+            var filePath = Path.Combine(_storagePath, file);
+
+            // return A new file with read/write access.
+            return File.Exists(filePath);
```
MonogameGuy, line 203: The comment is incorrect.  Should be something like:
```
// return  A boolean indicating the file's existence.
```

MonoGame.Framework/Platform/Storage/StorageDevice.Native.cs
```diff
@@ -0,0 +1,105 @@
+﻿using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Xna.Framework.Storage
+{
+    partial class StorageDevice
```
tomspilman, line 7: For the native platform you need to add a `Storage.Interop.cs` and forward all platform calls here to it.  See 
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Native/Platform.Interop.cs

Tests/Framework/Storage/StorageTests.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        const string MY_GAME = "MyGame";
+
+        [Test]
+        public void IsConnected_ShouldReturnFalse_OnException()
+        {
+            var device = new StorageDevice(null);
+
+            // We haven't opened a container yet, so should be false
+            Assert.IsFalse(device.IsConnected);
```
tomspilman, line 23: @ThomasFOG Do you find `IsConnected` to be useful in a modern API for storage?  Is there any case now where save storage is disconnected?

Tests/Framework/Storage/StorageTests.cs
```diff
@@ -0,0 +1,64 @@
...
+            Assert.AreNotEqual(0, device.TotalSpace);
+            Assert.AreNotEqual(0, device.FreeSpace);
+        }
+
+        [Test]
+        public void OpenContainer_ShouldThrowException_OnInvalidDisplayName()
+        {
+            var device = new StorageDevice(PlayerIndex.One);
+            Assert.Throws<ArgumentNullException>(() => device.OpenContainer(null), "Value cannot be null. (Parameter 'displayName')");
+            Assert.Throws<ArgumentException>(() => device.OpenContainer(string.Empty), "The value cannot be an empty string. (Parameter 'displayName')");
```
tomspilman, line 44: So some platforms have limits on what characters are allowed for save storage names.

Should we just allow the code to fail on that platform?  Or should we set some rules on all platforms?

For example on Windows it would have to be limited by names that work as file names disallow `<>:"/\|?*` in a name.

Also should we come up with some name length limit?

MonoGame.Framework/Storage/StorageContainer.cs
```diff
@@ -0,0 +1,279 @@
...
+            var storageRoot = StorageDevice.StorageRoot;
+            var savedGames = Path.Combine(storageRoot, "SavedGames");
+#endif
+            _storagePath = Path.Combine(savedGames, _displayName);
+
+            // If we have a PlayerIndex use that, otherwise save to AllPlayers folder
+            _storagePath = _playerIndex.HasValue ? Path.Combine(_storagePath, "Player" + (int)_playerIndex.Value) : Path.Combine(_storagePath, "AllPlayers");
+
+            // Create the "device" if need be
+            if (!Directory.Exists(_storagePath))
```
tomspilman, line 83: All file operations should be moved into the platform specific partial classes as some platforms do not use the file system for save games.

MonoGame.Framework/Storage/StorageContainer.cs
CartBlanche: Thanks @MonogameGuy, nicely spotted. Fixed in latest commit.

MonoGame.Framework/Platform/Storage/StorageDevice.Native.cs
CartBlanche: I've added a placeholder `Storage.Interop.cs` file in the native sub-dir. But need to chat to you about what exactly I need to put in there. I'm still unclear how the Native stuff hangs together. Maybe if you have time tomorrow?

MonoGame.Framework/Storage/StorageDevice.cs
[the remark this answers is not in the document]
CartBlanche: @tomspilman I think we should try to keep the API sort of similar, while removing the "Selectors". Hence why I have  both OpenContainer and DeleteContainer
The async ones weren't obviously in the original, but are a minor addition, in case there are people trying to write async stuff(but happy to remove them if you 100% think no one will ever use it)
I'm not a fan of adding TryOpen or OpenOrCreate. I think these scenarios should be automagically handled internally and we can either raise an approproate exception, that users can look out for or  alternatively maybe even fire some events like OnOpen, OnError, OnClosed, so they can be notified when those things happen. Let's jump on call and chat perhaps?

MonoGame.Framework/Platform/Storage/StorageDevice.Native.cs
CartBlanche: As per Discord: With Native I've "left the methods there throwing NotImplemented for now" So may need more direction from you, as to what exactly needs to go in there, or if you have the cycles for you to implement that platform, perhaps.

MonoGame.Framework/Platform/Native/Storage.Interop.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Xna.Framework.Storage
+{
+    public sealed partial class StorageDevice
```
tomspilman, line 7: What is this for?  The `.Interop.cs` files are reserved for the C# to C++ APIs which are all `internal` types.

See https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Native/Graphics.Interop.cs

tomspilman: Really you can remove this if you don't plan to write the code here to reference it.

Tests/Framework/Storage/StorageTests.cs
ThomasFOG: No, I don't think so. All systems ensure that storage device is always available nowadays.

CartBlanche:
@ThomasFOG here's the initial implementation for Desktop etc of your in memory container methods:
https://github.com/MonoGame/MonoGame/pull/8643/files#diff-bf7d656ec104081f0afb5e82e70cb4fa9bce42ab583cc889bce6b8eb58e97735R115-R162

Is this what you envisioned or something else?

Tests/Framework/Storage/StorageTests.cs
CartBlanche: I'll remove it.

Maniekko:
Is copilot thingy even needed here?

CartBlanche:
> Is copilot thingy even needed here?

I'm using it just to have an extra pair of eyes on the code. Unlikely to use it to "approve" a PR.

CartBlanche:
@ThomasFOG Could you please have a look at the in-memory and `SaveData()` implementations, taken from your post and let me know if this is what you had in mind or if any further changes are required? Also if you think we can get this into 3.8.3?

tomspilman:
>  Also if you think we can get this into 3.8.3?

Let me take a look at it first at least to be sure we're good for native and console.

MonoGame.Framework/Platform/Native/Storage.Interop.cs
tomspilman: Yes remove this...  this isn't how the `.Interop.cs` files are used.

tomspilman:
> Also if you think we can get this into 3.8.3?

I think we should wait... this PR needs some more discussion and work first.

Tests/Framework/Storage/StorageTests.cs
CartBlanche: Done for Android, iOS, Deskop and Windows. Native left untouched, for now.

MonoGame.Framework/Storage/StorageContainer.cs
```diff
@@ -0,0 +1,475 @@
...
+    /// <summary>
+    /// Contains a logical collection of files used for user-data storage.
+    /// </summary>			
+    /// <remarks>MSDN documentation contains related conceptual article: https://learn.microsoft.com/en-us/previous-versions/windows/xna/bb199074(v=xnagamestudio.40)</remarks>
+    public partial class StorageContainer : IDisposable
+    {
+        private readonly PlayerIndex? _playerIndex;
+
+        // In memory container
+        private Dictionary<string, byte[]> _containers;
```
tomspilman, line 29: This IS the `StorageContainer`...  it contains files.

Should this not be a list of files in the container?  Not a list of containers.

MonoGame.Framework/Storage/StorageContainer.cs
```diff
@@ -0,0 +1,475 @@
...
+
+            PlatformCreateDirectory(directoryName);
+        }
+
+        /// <summary>
+        /// Creates a file in the storage-container.
+        /// </summary>
+        /// <param name="fileName">Relative path of the file to be created.</param>
+        /// <returns>Returns <see cref="Stream"/> for the created file.</returns>
+        public Stream CreateFile(string fileName)
```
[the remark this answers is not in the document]
CartBlanche: @tomspilman, @ThomasFOG Ok I thought we were keeping the file manipulation methods as is, but the default state was the cache. Does it make sense to keep Create/Delete File/Directories etc at all if ultimately it will all just go to the container?

dellis1972:
@CartBlanche is this still being worked on? //cc @tomspilman

tomspilman:
I need to find some time to review this in detail.  I will in the next week.

MonoGame.Framework/Platform/Storage/StorageDevice.Native.cs
```diff
@@ -0,0 +1,102 @@
...
+                    return osConfigDir;
+                }
+                else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                {
+                    string osConfigDir = Environment.GetEnvironmentVariable("HOME");
+                    if (string.IsNullOrEmpty(osConfigDir))
+                    {
+                        return "."; // Fallback to current directory
+                    }
+                    osConfigDir = Path.Combine(osConfigDir, "Library", "Application Support");
```
dellis1972, line 67: This code will overwrite the `HOME` value? is that expected?

CartBlanche:
@tomspilman @ThomasFOG  build is now rebased against latest `develop` for review.

MonoGame.Framework/Platform/Storage/StorageDevice.Desktop.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <returns>
+        /// A string representing the root directory path for application data storage.
+        /// </returns>
+        internal static string StorageRoot
+        {
+            get
+            {
+                if (CurrentPlatform.OS == OS.Linux)
+                {
+                    string osConfigDir = Environment.GetEnvironmentVariable("XDG_DATA_HOME");
```
dellis1972, line 50: why are we using XDG_DATA_HOME and not just HOME?

dellis1972: I suspect this is why you had to try to support the virtual display stuff.

dellis1972: why are we not using `Environment.GetFolderPath(Environment.SpecialFolder.UserProfile)` ?

dellis1972: or `LocalApplicationData` ?

tomspilman:
So @CartBlanche this PR isn't ready.  I reviewed it and it needs a bunch of work.  I will see about taking over this work in the upcoming Friday stream and see if i can get it over the line.

tomspilman:
Ok an update... dug into this on the stream today, but didn't get things "done" will have to continue the work next week.

CartBlanche:
@tomspilman I can't see your latest changes and I'v just rebased this PR

tomspilman:
@CartBlanche it's not ready yet. I'm trying to get 3.8.5 out the door, then I can come back to this.

tomspilman:
FYI.  Moved this back to draft status while i work on it!

tomspilman (changes requested):

tomspilman:
FYI.  I pushed my latest work from the Code Time stream today, but still a big WIP!

PR #8646, opened 2025-02-17
DirectX 12 Support

tomspilman:
This PR adds DirectX 12 support to the new native platform backend.

The implementation was originally created for Xbox (see https://github.com/MonoGame/MonoGame/issues/8195) but refactored to fit under the new native API.

This implementation is unified to support [Windows and Xbox GDK](https://learn.microsoft.com/en-us/gaming/gdk/) platforms.


MonoGame.Framework.Native.sln
```diff
@@ -89,6 +91,21 @@ Global
...
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Debug_VK|x64.ActiveCfg = Debug|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Debug_VK|x64.Build.0 = Debug|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Debug|Any CPU.ActiveCfg = Debug|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Debug|Any CPU.Build.0 = Debug|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Debug|x64.ActiveCfg = Debug|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Debug|x64.Build.0 = Debug|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Release_VK|Any CPU.ActiveCfg = Release|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Release_VK|Any CPU.Build.0 = Release|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Release_VK|x64.ActiveCfg = Release|x64
+		{CC33DB03-389E-8F7A-81DC-4020ED856DCF}.Release_VK|x64.Build.0 = Release|x64
```
tomspilman, line 104: I should probably fix this to not be `Release_VK` and `Debug_VK`.  I think i set that up when i original thought that we would have one vcproj with both Vulkan and DirectX support in it.

Thoughts?

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.Vulkan.cs
```diff
@@ -448,34 +446,6 @@ internal override ShaderData CreateShader(ShaderResult shaderResult, string shad
                             }                        
                         }
 
-                        int size;
```
tomspilman, line 451: I found that this section of code was unused. So i removed it.

native/monogame/directx12/CommandContext.cpp
```diff
@@ -0,0 +1,363 @@
...
+#include "directx12.h"
+
+#include "CommandContext.h"
+#include "DeviceResources.h"
+#include "GraphicsEnums.h"
+#include "Texture.h"
+#include "Sampler.h"
+#include "Heaps.h"
+
+#if defined(_GAMING_XBOX_SCARLETT)
```
tomspilman, line 14: So i decided after reviewing other open source projects it is fine for us to put Xbox code like this in public.

Microsoft does it in their open source repos all the time.  Other open source projects do as well.

https://github.com/search?q=_GAMING_XBOX&type=code

native/monogame/sdl/MGP_sdl.cpp
```diff
@@ -678,7 +682,7 @@ MGP_Window* MGP_Window_Create(
 
 	Uint32 flags = SDL_WINDOW_HIDDEN;// | SDL_WINDOW_FULLSCREEN_DESKTOP;
 
-#if MG_VULKAN
+#if defined(MG_VULKAN) || defined(MG_DIRECTX12)
 	flags |= SDL_WINDOW_VULKAN;
```
dellis1972, line 686: should we use a VULKAN window when using directX? Seems a bit odd.

MonoGame.Framework.Native.sln
```diff
@@ -13,6 +13,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MonoGame.Tests.WindowsVK",
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "desktopvk", "native\monogame\desktopvk.vcxproj", "{60D6243F-CC40-D9B5-157F-8A5B8128B70A}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "windowsdx", "native\monogame\windowsdx.vcxproj", "{CC33DB03-389E-8F7A-81DC-4020ED856DCF}"
```
tomspilman, line 16: Should the native project be called WindowsDX like i did here?  Or maybe DestkopDX to match DesktopVK?

I used WindowsDX because there is no Mac or Linux support... which was the intent of the term "desktop".

Thoughts?

native/monogame/sdl/MGP_sdl.cpp
```diff
@@ -257,6 +257,8 @@ MGMonoGamePlatform MGP_Platform_GetPlatform()
 {
 #if MG_VULKAN
     return MGMonoGamePlatform::DesktopVK;
+#elif MG_DIRECTX12
+    return MGMonoGamePlatform::Windows;
```
tomspilman, line 261: So in this case we just use SDL on desktop, so we use Windows as the platform.

But for Vulkan i had added a platform `DesktopVK`.

I think we need to discuss and rethink what is a platform.

native/monogame/sdl/MGP_sdl.cpp
tomspilman: I think this flag is required regardless of what it looks like.  I will double check it.

MonoGame.Framework.Native.sln
ThomasFOG: If they are separate project files, I think that "debug" and "release" should fit.

squarebananas:
Managed to build and run the default Game1 code yesterday. I wasn't able to build the Tools though.

I had a look again just and noticed `MonoGame.Effect.Compiler` isn't building because the `ConstantBuffer.SetSize` method and a `ConstantBuffer.AddParameter` overload method are not defined anywhere.

My guess is, as `ConstantBufferData` is a partial class, there is a missing `ConstantBufferData.DX12` file (or some other name) which hasn't been included in the project.

MonoGame.Framework.Native.sln
ThomasFOG: Maybe match the platform definitions? desktopVK, windowsGDK...

tomspilman:
> noticed `MonoGame.Effect.Compiler` isn't building

@squarebananas this should be fixed now.

MonoGame.Framework.Native.sln
tomspilman: Ok... so GDK... are we really GDK?

In this case we're just using SDL2 (window and inputs) + DirectX12 + no audio yet...  so is this considered GDK?

Or is that just plain Windows ?

squarebananas:
> @squarebananas this should be fixed now.
<details><summary>Confirmed, custom effects are now working.</summary>

https://github.com/user-attachments/assets/0330c262-dab7-4d2c-a2e0-9dc50cb4cfa2
</details>

tomspilman:
This PR is done... looking for final reviews before merging it.

PR #8658, opened 2025-03-06
Fix Vulkan CI build

CartBlanche:
Reference specific commit for the Vulkan setup, so things build. We'll need to Update this once the 1.2.1 release it official.

As per temporary suggested fix there - https://github.com/humbletim/setup-vulkan-sdk/issues/28#issuecomment-2702505769

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,10 +37,10 @@ jobs:
           java-version: '17'
 
       - name: Install Vulkan SDK
-        uses: humbletim/install-vulkan-sdk@v1.1.1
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -37,10 +37,10 @@ jobs:
           java-version: '17'
 
       - name: Install Vulkan SDK
-        uses: humbletim/install-vulkan-sdk@v1.1.1
+        uses: humbletim/setup-vulkan-sdk@523828e49cd4afabce369c39c7ee6543a2b7a735
```
tomspilman, line 40: This seems like it should be ok.  

We just need the Vulkan SDK installed on the build machine (because we don't use a dependencies repo anymore).

If this is the right fix to that then i'm good with it.

PR #8660, opened 2025-03-10
Fix: Last char missing from filename on filedrop event - WindowsDX platform

### LINKED ISSUE #8657: FileDrop in a WindowsDX project cuts off the last character in the path
### Prerequisites

### MonoGame Version

Monogame 3.8.2.1105

MonoGame Windows Desktop Application (mgwindowsdx)

### Operating System

Windows

### Description

When using Window.FileDrop event to get the path of a file dropped onto the GameWindow the expected string is the path of the file that was dropped. In Monogame.Framework.WindowsDX 3.8.1.303 the string returns as expected. In 3.8.2.1105 the string returned cuts off the last character of the path. 

Cross Platform projects that use DesktopGL function properly on both versions 3.8.1.303 and 3.8.2.1105.

### Steps to Reproduce

Create a WindowsDX desktop project with Monogame 3.8.2.1105

Create a function to handle the FileDrop event and subscribe to it.

Drop file into the game window and check the path.

Code I used that works as expected in 3.8.1.303 but fails in 3.8.2.1105

```cs
    {
        Window.FileDrop += new EventHandler<FileDropEventArgs>(FileDropped);
        base.Initialize();
    }
    void FileDropped(object sender, FileDropEventArgs e)
    {
        Window.Title = e.Files[0];
    }
```

### Minimal Example Repo

_No response_

### Expected Behavior

Example file at location "C:\1.png" would return "C:\1.png"

### Resulting Behavior

Example file at location "C:\1.png" returns "C:\1.pn"

### Files

_No response_

theLaffa:
Fixes #8657 

FileDrop in a WindowsDX project cuts off the last character in the path

### Description of Change

Adding a byte to the final call of DragQueryFile as it expects the final byte to be the destination of the null string terminator. Ie, '.png' is 4 chars, with buffSize = 4, result would be '.pn`null`'

> [out] lpszFile
> 
> Type: LPTSTR
> 
> The address of a buffer that receives the file name of a dropped file when the function returns. This file name is a null-terminated string. If this parameter is NULL, DragQueryFile returns the required size, in characters, of this buffer.

### DIFF
--- a/MonoGame.Framework/Platform/Windows/WinFormsGameForm.cs
+++ b/MonoGame.Framework/Platform/Windows/WinFormsGameForm.cs
@@ -235,7 +235,7 @@ namespace Microsoft.Xna.Framework.Windows
             {
                 uint buffSize = DragQueryFile(hdrop, i, null, int.MaxValue);
                 StringBuilder builder = new StringBuilder((int)buffSize);
-                DragQueryFile(hdrop, i, builder, buffSize);
+                DragQueryFile(hdrop, i, builder, buffSize + 1); // Extra byte for null terminator
                 files[i] = builder.ToString();
             }
 


MonoGame.Framework/Platform/Windows/WinFormsGameForm.cs
```diff
@@ -235,7 +235,7 @@ void HandleDropMessage(ref Message m)
             {
                 uint buffSize = DragQueryFile(hdrop, i, null, int.MaxValue);
                 StringBuilder builder = new StringBuilder((int)buffSize);
-                DragQueryFile(hdrop, i, builder, buffSize);
+                DragQueryFile(hdrop, i, builder, buffSize + 1); // Extra byte for null terminator
```
Zeroto, line 238: Would this not write a null character in the byte just after the buffer? The buffer itself needs to be increased 1 in size most likely

tomspilman: Yes i mean both:

```
StringBuilder builder = new StringBuilder((int)buffSize + 1);
DragQueryFile(hdrop, i, builder, buffSize + 1); // Extra byte for null terminator
```

Does this not work?

theLaffa: It does, the results are the same for either way.

Technically, the only difference is that the n+1 byte is never written to as null terminator is not parsed.

Current solution :
image

bufferSize +1:
image

Zeroto: Looking at the docs, the marshaller will add 1 to the buffer size to account for the null terminator:

"It can also be initialized to a fixed length. For example, if you initialize a StringBuilder buffer to a capacity of N, the marshaller provides a buffer of size (N+1) characters. The +1 accounts for the fact that the unmanaged string has a null terminator while StringBuilder does not."
https://learn.microsoft.com/en-us/dotnet/framework/interop/default-marshalling-for-strings

So yes, the pr code will work correctly.

However, the page also notes that using stringbuilder is not great for performance because of the allocations involved. In this case it might not matter much, but it might be better to switch to a stackalloc char[] buffer or array pool in the future instead.

https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices#string-parameters

MonogameGuy: > However, the page also notes that using stringbuilder is not great for performance because of the allocations involved. In this case it might not matter much, but it might be better to switch to a stackalloc char[] buffer or array pool in the future instead.

By giving the buffer size as an argument there is only one memory allocation as no expansion is needed.

Zeroto: > > However, the page also notes that using stringbuilder is not great for performance because of the allocations involved. In this case it might not matter much, but it might be better to switch to a stackalloc char[] buffer or array pool in the future instead.
> 
> By giving the buffer size as an argument there is only one memory allocation as no expansion is needed.

If you read this page, https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices#string-parameters, it outlines which allocations happen. Marshalling a StringBuilder will also allocate a native buffer. And ToString will also allocate a new array for the string.

Not completely sure about the 2c allocation mentioned on that page. But even without that, we are still looking at 3 allocations. If it is switched to using Arraypool, then we avoid 2 of those allocations(Initial buffer allocation and the marshalling native buffer). We can also use stackalloc. Even though it is still an allocation, it is a very cheap one because it is a stack allocation. But anyway, I don't think this is a performance cricital piece of code, so it should be fine to use a stringbuilder

tomspilman: Unless you plan to handle 100s of drag and drop operations a frame...  i don't feel the extra allocations matter too much here.

My only concern was about memory corruption here.  If the marshaler correctly handles the null and we just need to pass a +1 to the call then we're good.

dellis1972: @theLaffa any update on this? ^

theLaffa: I think the consensus was that this fix was okay as is.

String marshal knows how to handle the extra null char
It is highly unlikely that users will be expected to drop 100's of files
- String builder is set to a fixed width, reducing allocations
- Char[] variations would still require at least 2 allocations, instantiation and .ToString()

SimonDarksideJ (approved):
Approved for merge, keep watch on use for the 3.8.3 release

PR #8663, opened 2025-03-17
Refactor to make accelerometer and compass classes available in cross-platform apps.

### LINKED ISSUE #8664: Accelerometer and compass classes are not available in cross-platform apps
### Prerequisites

### MonoGame Version

MonoGame 3.8.2.*

### Operating System

Desktop, Native, WindowsDX

### Description

Currently the Accelerometer classes are only available from within iOS and Android apps. To be able to support the *.Core library model, it essentially need to be available

### Steps to Reproduce

Create a Cross Platform app, try to refer to the Accelerometer classes, so you have 1 Core codebase. 
As the Accelerometer and Compass classes were only available in iOS and Android apps, you weren't easily able to access those classes.

### Minimal Example Repo

_No response_

### Expected Behavior

We should be able centralise the Accelerometer and Compass calls. These will be ignored or throw `PlatformNotSupported` if called on Desktop, WindowsDX, but should just work (TM) on the the mobile platforms.

### Resulting Behavior

Unable to build crossplatform apps, with a shared Core library.

### Files

_No response_

CartBlanche:
Fixes #8664 

### Description of Change

Moved the SensorData classes up to the top level (still don't like that the namespaces are `Microsoft.Devices.*`), then refactored the Platform classes to remove the iOS IFDEF, following more consistent patterns in the codebase. 
Then added the classes to the non supported classes (the throw `PlatformNotSupported` exceptions if called), but this now allows game devs to use Accelerometer and Compass classes in a more cross-platform way.

Since Microsoft has pulled the WinPhone, in MG 4, we should consider renaming the namespace to `MonoGame.Devices.*`, IMHO.


MonoGame.Framework/Platform/Desktop/Devices/Sensors/Accelerometer.cs
```diff
@@ -0,0 +1,81 @@
...
+using System;
+using Microsoft.Xna.Framework;
+
+namespace Microsoft.Devices.Sensors
+{
+    /// <summary>
+    /// Provides Desktop applications access to the device's accelerometer sensor.
+    /// Currently stubs, for cross-platform development.
+    /// </summary>
+    public sealed class Accelerometer : SensorBase<AccelerometerReading>
```
dellis1972, line 14: While this is a move in the right direction, I think the pattern is to have a `partial` class which them implements partial `Platform*` methods to do the work.  e.g

```
public partial sealed class Accelerometer : SensorBase<AccelerometerReading>
{
        partial bool PlatformIsSupported();

 public static bool IsSupported
        {
            get
            {
                return PlatformIsSupported ();
            }
        }
}
```

then in a `Accelerometer.Desktop.cs` 

```
public partial sealed class Accelerometer : SensorBase<AccelerometerReading>
{
        partial bool PlatformIsSupported()
       {
             return false;
        }
}
```

This would reduce the amount of duplicate code, but also allow use to at some point implement the "Native" side. 
The question is should this be done now or cleaned up later?

@tomspilman ?

CartBlanche: @dellis1972 agreed. I just didn't want to change existing stuff just yet. But can certainly do that as part of this PR.

CartBlanche: @dellis1972 Refactor done. Partialisated ALL the things!

MonoGame.Framework/Microsoft/Devices/Sensors/Accelerometer.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        /// <summary>
+        /// Initializes the platform resources required for the accelerometer sensor.
+        /// </summary>
+        static void Initialize()
+        {
+            PlatformInitialize();
+        }
+
+        void ActivityPaused(object sender, EventArgs eventArgs)
```
dellis1972, line 43: I don't think these should be in the this file, but should be in the Android one as they are Android specific

CartBlanche: @dellis1972 I've left them in Android and removed them everywhere else. Could you please double check the Compass implementation on Android, just to be sure that's correct?

MonoGame.Framework/Microsoft/Devices/Sensors/Accelerometer.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    /// Provides Desktop applications access to the device's accelerometer sensor.
+    /// Currently stubs, for cross-platform development.
+    /// </summary>
+    public sealed partial class Accelerometer : SensorBase<AccelerometerReading>
+    {
+        /// <summary>
+        /// Gets or sets whether the device on which the application is running supports the accelerometer sensor.
+        /// </summary>
+        public static bool IsSupported => PlatformIsSupported();
```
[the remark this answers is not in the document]
ThomasFOG: It does, but quite frankly, we should really not start to disable the language check and assuming that PR reviewers will have sharp enough eye to prevent incompatible stuff to get merged. Because they won't (including me).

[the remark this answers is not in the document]
CartBlanche: @dellis1972, @ThomasFOG Just double checking that the language check hasn't been disable, right?

SimonDarksideJ (approved):
Approved, looks good but would also benefit being tested with the Startkit

CartBlanche:
@SimonDarksideJ namespace rename done and merged into this PR/branch. Testing with StartKit will happen in the next PR, where a `Accelerometer/Compass.GetState()` method will be added for consistency with other input APIs.

PR #8669, opened 2025-03-20
Add a new native API to handle loading assets

dellis1972:
So lets expose an API for loading ReadOnly assets. 
This will allow us to make use of specific native I/O API's on various platforms such as Consoles and Mobile, while keeping the native C# interop code consistent and free of #if blocks or partial classes.

### DIFF
--- a/MonoGame.Framework/Platform/Native/TitleContainer.Interop.cs
+++ b/MonoGame.Framework/Platform/Native/TitleContainer.Interop.cs
@@ -1,0 +1,115 @@
+using System;
+using System.IO;
+using MonoGame.Interop;
+using System.Runtime.InteropServices;
+
+internal unsafe class MGReadOnlyStream : Stream
+{
+    private MG_Asset* _asset;
+    private long _length;
+    private long _position;
+
+    public override bool CanRead => true;
+
+    public override bool CanSeek => true;
+
+    public override bool CanWrite => false;
+
+    public override long Length => _length;
+
+    public override long Position
+    {
+        get => _position;
+        set
+        {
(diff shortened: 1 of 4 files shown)


MonoGame.Framework/Platform/Native/TitleContainer.Interop.cs
```diff
@@ -0,0 +1,115 @@
+using System;
+using System.IO;
+using MonoGame.Interop;
+using System.Runtime.InteropServices;
+
+internal unsafe class MGReadOnlyStream : Stream
```
tomspilman, line 6: Have we done this sort of prefixing of "MG" in C# classes before?

Maybe this should be `MG_AssetStream` or `AssetStream` ?

Minor thing.

tomspilman:
Seems good overall.

Note we need to update the Switch and Xbox repos to support this new API.

MonoGame.Framework/Platform/Native/TitleContainer.Native.cs
```diff
@@ -10,14 +10,14 @@ namespace Microsoft.Xna.Framework;
...
 partial class TitleContainer
 {
+
     static partial void PlatformInit()
     {
     }
 
     private static Stream PlatformOpenStream(string safeName)
     {
         var absolutePath = MGP.Platform_MakePath(Location, safeName);
```
tomspilman, line 20: So do we need MGP.Platform_MakePath anymore?

Do we use it other places?  Is it better for `MG.OpenRead` to take the `Location, safeName` instead and build the path internally?

tomspilman (approved):

PR #8673, opened 2025-03-23
Fix Effect WineHelper to work with wine 10.0.

dellis1972:
Wine 10.0 removed the wine64 executable and now only has wine. This change detects which one is available and uses it.

It also fixes an issue where the working directory could not be set.

```
An error occurred trying to start process 'wine64' with working directory '/Users/runner/work/MonoGame/MonoGame/Artifacts/Tests/Tools/Release'.
```

We do this by providing an actual working directory which we know exists. This is the location of the WineHelper executable.


dellis1972:
@SimonDarksideJ this fixes the mac build.

PR #8680, opened 2025-03-26
SDL update to 2.32.2

ThomasFOG:
This PR updates DesktopGL to SDL 2.32.2

Please mind that before merging this PR, [the nuget](https://github.com/MonoGame/MonoGame.Library.SDL/releases/tag/2.32.2.1) needs to be pushed to nuget.org.


harry-cpp:
Modify the following line: <https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/SDL/SDL2.cs#L23>

ThomasFOG:
> Modify the following line: https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/SDL/SDL2.cs#L23

It would break for any user trying to use/switch their nuget manually. I think that we're doomed to use that filepath now (unless we're fine with breaking a nuget mid-way).

CartBlanche (approved):
LGTM

PR #8681, opened 2025-03-26
Rename Microsoft.Devices.Sensors namespace to MonoGame.Framework.Devices.Sensors

CartBlanche:
As per subject line.


dellis1972:
This namespace change will break anyone using the old namespace, when they upgrade.

SimonDarksideJ (approved):

ThomasFOG:
Good to me and I think that users impacted will be few enough to handle the eventual support requests.

PR #8688, opened 2025-03-31
Fix GamePad button mappings for Android and iOS devices.

### LINKED ISSUE #8522: The GamePad.Back may remain set true on Android after the application calls the Exit()
### Prerequisites

### MonoGame Version

MonoGame 3.8.x

MonoGame Android Application (mgandroid)

### Operating System

Windows

### Description

The **GamePad.Back** is set true in the **OnKeyDown**, and it should be reset to false in the **OnKeyUp**.
But it seems there's a race condition and sometimes the **OnKeyUp** is not received.
That makes the **GamePad.Back = true** permanently.

### Steps to Reproduce

That is reproduced on the default android template. You just have to click back a few times.

### Minimal Example Repo

_No response_

### Expected Behavior

The  **GamePad.Back** should be set to false after the application comes to foreground again.

### Resulting Behavior

The  **GamePad.Back** remains set to true after the application comes to foreground again.

### Files

_No response_

CartBlanche:
Fixes #8522

**Known Issue** 
iOS *ThumbStickButton is always null. Issue raised here - https://github.com/dotnet/macios/issues/22524

### DIFF
--- a/MonoGame.Framework/Platform/Input/GamePad.Android.cs
+++ b/MonoGame.Framework/Platform/Input/GamePad.Android.cs
@@ -151,6 +151,8 @@ namespace Microsoft.Xna.Framework.Input
             {
                 if (index == 0 && Back)
                 {
+                    Back = false;
+
                     return new GamePadState(
                         new GamePadThumbSticks(),
                         new GamePadTriggers(),
--- a/MonoGame.Framework/Platform/Input/GamePad.Android.cs
+++ b/MonoGame.Framework/Platform/Input/GamePad.Android.cs
@@ -283,7 +283,26 @@ namespace Microsoft.Xna.Framework.Input
             gamePad._leftTrigger = e.GetAxisValue(Axis.Brake);
             gamePad._rightTrigger = e.GetAxisValue(Axis.Gas);
 
-            if(!gamePad.DPadButtons)
+            // We set the Trigger buttons here, because it never fires in OnKeyDown
+            if (gamePad._leftTrigger > 0f)
+            {
+                gamePad._buttons |= Buttons.LeftTrigger;
+            }
+            else
+            {
+                gamePad._buttons &= Buttons.LeftTrigger;
+            }
+
+            if (gamePad._rightTrigger > 0f)
+            {
+                gamePad._buttons |= Buttons.RightTrigger;
+            }
+            else
+            {
+                gamePad._buttons &= Buttons.RightTrigger;
+            }
+
+            if (!gamePad.DPadButtons)
             {
                 if(e.GetAxisValue(Axis.HatX) < 0)
                 {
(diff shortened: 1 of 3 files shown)


SimonDarksideJ:
P.S. while fixing this @CartBlanche , this issue was also logged related to the Android "Back" button
https://github.com/MonoGame/MonoGame/issues/8522

ThomasFOG:
For Android back button, I remember that we historically disabled it because Android maps it by default on the system "back", which means that it was basically quitting the app. Or something like that, can't remember exactly but there was something in that line.

MonoGame.Framework/Platform/Input/GamePad.iOS.cs
```diff
@@ -151,10 +151,16 @@ private static GamePadState PlatformGetState(int index, GamePadDeadZone leftDead
                     if (controller.ExtendedGamepad.RightTrigger.IsPressed)
                         buttons |= Buttons.RightTrigger;
 
-                    if (controller.ExtendedGamepad.ButtonMenu.IsPressed)
+                    if (controller.ExtendedGamepad.ButtonMenu != null
+                    && controller.ExtendedGamepad.ButtonMenu.IsPressed)
```
ThomasFOG, line 155: If you feel like it, you may write this as ```controller.ExtendedGamepad.ButtonMenu?.IsPressed```. Since the iOS specific code doesn't have to abide to the C# 5 rule, it's fine in this occurrence and won't fail ```ConsoleCheck```.

CartBlanche: @ThomasFOG the issue we have is that the Xamarin bindings have some properties as nullable, like `controller.ExtendedGamepad.ButtonOptions?`, while others are not. Hence why I  used the above.

MonoGame.Framework/Platform/Input/GamePad.Android.cs
```diff
@@ -359,8 +380,10 @@ private static Buttons ButtonForKeyCode(Keycode keyCode)
 
                 case Keycode.ButtonStart:
                     return Buttons.Start;
-                case Keycode.Back:
+                case Keycode.ButtonSelect:
                     return Buttons.Back;
+                case Keycode.ButtonMode:
+                    return Buttons.BigButton;
```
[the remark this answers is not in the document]
CartBlanche: @squarebananas after an internal discussion, I've removed the BigButton mapping, at least for this release.

CartBlanche:
> P.S. while fixing this @CartBlanche , this issue was also logged related to the Android "Back" button #8522

I''ve added the suggested fix to this PR.

MonoGame.Framework/Platform/Input/GamePad.Android.cs
```diff
@@ -283,7 +287,26 @@ internal static bool OnGenericMotionEvent(MotionEvent e)
...
+                gamePad._buttons &= Buttons.LeftTrigger;
+            }
+
+            if (gamePad._rightTrigger > 0f)
+            {
+                gamePad._buttons |= Buttons.RightTrigger;
+            }
+            else
+            {
+                gamePad._buttons &= Buttons.RightTrigger;
```
viniciusjarina, line 306: This should be:
```
gamePad._buttons &= ~Buttons.RightTrigger;
```
Otherwise you will reset the state of the other buttons

PR #8690, opened 2025-03-31
Fix some build warnings in Debug due to malformed xmldoc comments.

ben-marsh:
Just a quick one - saw a few trivial warnings when building in debug that were due to invalid XMLDOC comments. Thought I'd fix them up.


SimonDarksideJ (approved):
Thanks for the contributions @ben-marsh !

PR #8706, opened 2025-04-09
Remove call to RestoreMGCB()

CartBlanche:
As per subject line. 

Devs should be relying on `dotnet tool restore` and we shouldn't be doing it for them.

### DIFF
--- a/MonoGame.Tools.Linux.sln
+++ b/MonoGame.Tools.Linux.sln
@@ -23,8 +23,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MonoGame.Content.Builder.Ed
 		{B3201192-46D5-40C5-A24E-493EC66D8C9B} = {B3201192-46D5-40C5-A24E-493EC66D8C9B}
 	EndProjectSection
 EndProject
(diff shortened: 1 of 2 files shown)


MonoGame.Tools.Linux.sln
```diff
@@ -23,8 +23,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MonoGame.Content.Builder.Ed
 		{B3201192-46D5-40C5-A24E-493EC66D8C9B} = {B3201192-46D5-40C5-A24E-493EC66D8C9B}
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MonoGame.Packaging.Flatpak", "Tools\MonoGame.Packaging.Flatpak\MonoGame.Packaging.Flatpak.csproj", "{2AAE746F-0EBE-48B4-8657-2B469B3E1C8F}"
```
SimonDarksideJ, line 26: I'd revert this removal and do it in a separate PR, checking with @harry-cpp as to whether it should be removed, as this is not the focus of this PR.

SimonDarksideJ (changes requested):
The remove "restoremgcb" changes are fine, but they should be the only change in this pr as the other changes are out of scope and should be in a separate pr

SimonDarksideJ (approved):

MonoGame.Tools.Linux.sln
harry-cpp: That project does not exist, its just a reference left in .sln file, its fine to be in here.

PR #8707, opened 2025-04-10
Fix GamePad issue with Android and iOS

viniciusjarina:
*  Regression on Android resetting button state.
*  Update `HasLeftStickButton` and  `HasRightStickButton` (Android)
*  Mark `HasDPadDownButton` as true since `hasKeys` always return false for directional keys
*  Update `LeftShoulder` and `RightShoulder` on iOS when falling back to `GamePad`
*  Mark `capability.IsConnected` as `true`

cc @CartBlanche

### DIFF
--- a/MonoGame.Framework/Platform/Input/GamePad.Android.cs
+++ b/MonoGame.Framework/Platform/Input/GamePad.Android.cs
@@ -92,18 +92,21 @@ namespace Microsoft.Xna.Framework.Input
             // this will need fixing
             capabilities.HasLeftXThumbStick = hasMap[4];
             capabilities.HasLeftYThumbStick = hasMap[4];
+            capabilities.HasLeftStickButton = hasMap[4];
+
             capabilities.HasRightXThumbStick = hasMap[5];
             capabilities.HasRightYThumbStick = hasMap[5];
+            capabilities.HasRightStickButton = hasMap[5];
 
             capabilities.HasLeftShoulderButton = hasMap[6];
             capabilities.HasRightShoulderButton = hasMap[7];
             capabilities.HasLeftTrigger = hasMap[8];
             capabilities.HasRightTrigger = hasMap[9];
 
-            capabilities.HasDPadDownButton = hasMap[10];
-            capabilities.HasDPadLeftButton = hasMap[11];
-            capabilities.HasDPadRightButton = hasMap[12];
-            capabilities.HasDPadUpButton = hasMap[13];
+            capabilities.HasDPadDownButton = true;
+            capabilities.HasDPadLeftButton = true;
+            capabilities.HasDPadRightButton = true;
+            capabilities.HasDPadUpButton = true;
 
             capabilities.HasStartButton = hasMap[14];
             capabilities.HasBackButton = hasMap[15];
@@ -294,7 +297,7 @@ namespace Microsoft.Xna.Framework.Input
             }
             else
             {
-                gamePad._buttons &= Buttons.LeftTrigger;
+                gamePad._buttons &= ~Buttons.LeftTrigger;
             }
 
             if (gamePad._rightTrigger > 0f)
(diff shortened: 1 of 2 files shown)


CartBlanche:
Testing it now.

MonoGame.Framework/Platform/Input/GamePad.Android.cs
```diff
@@ -92,18 +92,21 @@ private static GamePadCapabilities CapabilitiesOfDevice(InputDevice device)
...
             capabilities.HasLeftShoulderButton = hasMap[6];
             capabilities.HasRightShoulderButton = hasMap[7];
             capabilities.HasLeftTrigger = hasMap[8];
             capabilities.HasRightTrigger = hasMap[9];
 
-            capabilities.HasDPadDownButton = hasMap[10];
-            capabilities.HasDPadLeftButton = hasMap[11];
-            capabilities.HasDPadRightButton = hasMap[12];
-            capabilities.HasDPadUpButton = hasMap[13];
+            capabilities.HasDPadDownButton = true;
```
CartBlanche, line 106: Why are these being hardcoded as true?

MonoGame.Framework/Platform/Input/GamePad.Android.cs
```diff
@@ -294,7 +297,7 @@ internal static bool OnGenericMotionEvent(MotionEvent e)
             }
             else
             {
-                gamePad._buttons &= Buttons.LeftTrigger;
+                gamePad._buttons &= ~Buttons.LeftTrigger;
```
CartBlanche, line 300: Thanks missed this.

MonoGame.Framework/Platform/Input/GamePad.iOS.cs
```diff
@@ -213,6 +215,11 @@ private static GamePadState PlatformGetState(int index, GamePadDeadZone leftDead
                     if (controller.Gamepad.ButtonY.IsPressed)
                         buttons |= Buttons.Y;
 
+                    if (controller.Gamepad.LeftShoulder.IsPressed)
```
CartBlanche, line 218: I'm surprised these didn't already exist, as the button appeared to be toggling correctly in @SimonDarksideJ 's GamePad app.

MonoGame.Framework/Platform/Input/GamePad.iOS.cs
```diff
@@ -59,7 +59,7 @@ private static GamePadCapabilities GetCapabilities(GCController controller)
             //All iOS controllers have these basics
             var capabilities = new GamePadCapabilities()
             {
-                IsConnected = false,
+                IsConnected = true,
```
CartBlanche, line 62: I think this should default to false, because right now even after the controller is turned off, it still says connected.

viniciusjarina: if the code reach this point the game controller is connected otherwise wouldn't be listed in `GCController.Controllers`. Also we never update capabilities `IsConnected` to `true` only the state.
if the game controller is not connected `PlatformGetCapabilities` will return `GamePadCapabilities { IsConnected = false };`

MonoGame.Framework/Platform/Input/GamePad.iOS.cs
viniciusjarina: I noticed a long time ago. I used to have a SteelSeries that was not extended that's when I noticed it was missing.

MonoGame.Framework/Platform/Input/GamePad.Android.cs
viniciusjarina: Android bug. `hasKeys` always return false for the DPads because for Android is a motion event not a keydown

MonoGame.Framework/Platform/Input/GamePad.iOS.cs
SimonDarksideJ: Yeah, this is very curious, as testing worked for both MS and Sony GamePads on all platforms correctly.

PR #8718, opened 2025-04-14
Fix content error silencing

### LINKED ISSUE #8717: ContentManager is hiding exceptions
When ```ContentManager.Load<T>(string)``` fails, the actual problem is discarded and all that we get is an uninformative exception with a simple "Could not load [assetName] asset!" and the message/stack trace of the underlying exception is lost.

This is due to this:

https://github.com/MonoGame/MonoGame/blob/3ccbead15ac6b849b6795c94f611ae0371dfdf54/MonoGame.Framework/Content/ContentManager.cs#L458

This is a regression introduced by #8582

The try/catch block silence the loading exception and doesn't pass it on. This makes error messages unclear and it also makes AOT errors invisible and impossible to debug properly. It is important and ```ContentLoadException``` are rethrown as-is to preserve the error message and stack trace (because they indicate in their message how to solve loading issues).

Instead of this:

```csharp
            catch (ContentLoadException)
            {
                try
                {
                    // If the file is not found, we try searching a file with differents extensions
                    // based on the type of asset searched (e.g. '.bmp' and '.png' for a Texture2D)
                    if (typeof(Texture2D).IsAssignableFrom(typeof(T)))
                    {
                        result = LoadTexture2DFromImageFile(assetName);
                    }
                }
                catch
                {
                    result = null;
                }
            }

			if (result == null)
            {
				throw new ContentLoadException("Could not load " + originalAssetName + " asset!");
            }
```

It should be something like this:

```csharp
            catch (ContentLoadException)
            {
                try
                {
                    // If the file is not found, we try searching a file with differents extensions
                    // based on the type of asset searched (e.g. '.bmp' and '.png' for a Texture2D)
                    if (typeof(Texture2D).IsAssignableFrom(typeof(T)))
                    {
                        result = LoadTexture2DFromImageFile(assetName);
                    }
                }
                catch
                {
                    result = null;
                }

                if (result == null) throw;
            }
```

ThomasFOG:
Fixes #8717

### DIFF
--- a/MonoGame.Framework/Content/ContentManager.cs
+++ b/MonoGame.Framework/Content/ContentManager.cs
@@ -453,27 +453,34 @@ namespace Microsoft.Xna.Framework.Content
                     }
                 }
             }
-            catch (ContentLoadException)
+            catch (ContentLoadException ex)
             {
-                try
+                // If the file is not found, we try searching a file with differents extensions
+                // based on the type of asset searched (e.g. '.bmp' and '.png' for a Texture2D)
+                if (ex.InnerException != null &&
+                    (ex.InnerException is FileNotFoundException ||
+                    ex.InnerException is DirectoryNotFoundException))
                 {
-                    // If the file is not found, we try searching a file with differents extensions
-                    // based on the type of asset searched (e.g. '.bmp' and '.png' for a Texture2D)
-                    if (typeof(Texture2D).IsAssignableFrom(typeof(T)))
+                    // only try if an image file exist to avoid loosing the original error
+                    if (typeof(Texture2D).IsAssignableFrom(typeof(T)) &&
+                        ImageFileExists(assetName))
                     {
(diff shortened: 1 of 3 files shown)


ThomasFOG:
I've remade the logic of loading alternative image files. We need to check if a file exists first to differentiate the two possible paths of exceptions. This way stack traces are preserved and error messages are clearer.

dellis1972:
@ThomasFOG There are some failing unit tests which check the error messages.

```
Failed ThrowExceptionIfNoAssetInAnySupportedImageFormats [230 ms]
  Error Message:
     Expected: String starting with "Could not load"
  But was:  "The content file was not found."
```

ThomasFOG:
I'll update the tests.

ThomasFOG:
PR is fixed (there was a bug in the logic and I updated the tests).

SimonDarksideJ (approved):
Fix looks good

PR #8720, opened 2025-04-15
DesktopGL implementation for MessageBox.Show()

ThomasFOG:
Here's an implementation of ```MessageBox.Show()``` for DesktopGL.

All platforms should be supported now.

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
@@ -74,7 +74,7 @@
     <Compile Include="Platform\Input\KeysHelper.cs" />
     <Compile Include="Platform\Input\Mouse.SDL.cs" />
     <Compile Include="Platform\Input\MouseCursor.SDL.cs" />
-    <Compile Include="Platform\Input\MessageBox.Default.cs" />
+    <Compile Include="Platform\Input\MessageBox.SDL.cs" />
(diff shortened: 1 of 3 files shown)


CartBlanche:
@ThomasFOG does SDL correctly order the buttons depending on platform? 
eg. Windows OK on Left, Cancel on Right, 
  Linux, Mac: Cancel on Left, Ok on right.

ThomasFOG:
By default, SDL will do "right to left" (*nix-style) ordering of the list provided, on all platforms.

What I did is mimicking what Tom has done on the Native backend: reordering to "left to right" (Windows-style, like what WindowsDX is already doing) on all platforms.

If we want to address that on DesktopGL, we'll need to replicate that on the Native backend too.

CartBlanche:
@ThomasFOG I think at some point in the future, it would be good for the dialogs to appear in their "Native" order on each platform and it shouldn't be a big fix, should it? Especially for a MessageBox, that usually only has 1 button (OK) or 2 max.

SimonDarksideJ (approved):
This looks good for DesktopGL, but can we create issues/Requests for any other platforms that will also need this implementation @ThomasFOG ?

ThomasFOG:
> This looks good for DesktopGL, but can we create issues/Requests for any other platforms that will also need this implementation @ThomasFOG ?

As far as I can tell it was the only platform missing it.

I'll update it to support automatic button orders whenever I can.

ThomasFOG:
PR updated to use native button order on each system.

Issue #8096, opened 2023-11-22
Window is set to null before UnloadContent on the WindowsDX platform


Apostolique:
### Prerequisites

- [X] I have verified this issue is not present in the `develop` branch
- [X] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

MonoGame 3.8.1.303

### Which MonoGame platform are you using?

MonoGame Windows Desktop Application (mgwindowsdx)

### Operating System

Windows

### Description

On the DesktopGL platform, it's possible to access the game window during `UnloadContent`. This makes it possible to save the window size and position for example. The window is only disposed during the Platform.Dispose:

https://github.com/MonoGame/MonoGame/blob/a91dcc70b681d0230e960ec8eeaec4f9ff94b605/MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs#L329-L339

On the WindowsDX platform, the Windows is disposed earlier in Platform.Exit:

https://github.com/MonoGame/MonoGame/blob/a91dcc70b681d0230e960ec8eeaec4f9ff94b605/MonoGame.Framework/Platform/Windows/WinFormsGamePlatform.cs#L71-L77

### Steps to Reproduce

Access the game window during `UnloadContent`.

### Expected Behavior

The game window shouldn't be disposed before `UnloadContent` is done.

### Resulting Behavior

The game window position and size can't be accessed.

stromkos:
> The game window position and size can't be accessed.... The game window shouldn't be disposed before UnloadContent is done.

`UnloadContent()`  Is the place to dispose any manually allocated resources previously allocated via native `Malloc`calls.

Why would you expect the `window` to exist during `UnloadContent()`?


Store the information at load and modify.

Apostolique:
The window's values can change through the game's lifetime. For example when the window resize. It should be possible to cleanly get the window's state when the game exists instead of having to keep track of every possible changes that the window can get. This is currently already the case with OpenGL.

It makes for a nicer user experience when I restore the window at launch to the values that it had when the game last exited.

CartBlanche:
[...]

PR #8721, opened 2025-04-16
Fix Issue #8096, so UnloadContent happens before Exiting. (#8721)

CartBlanche:
As per subject.

Fixes: #8096

### DIFF
--- a/MonoGame.Framework/Platform/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Platform/Windows/WinFormsGameWindow.cs
@@ -240,7 +240,7 @@ namespace MonoGame.Framework
 
         private void OnFormClosing(object sender, FormClosingEventArgs e)
         {
-            e.Cancel = true;
+            // TODO WHY? e.Cancel = true;
             _platform.Game.Exit();
         }
 
--- a/MonoGame.Framework/Platform/Windows/WinFormsGameWindow.cs
+++ b/MonoGame.Framework/Platform/Windows/WinFormsGameWindow.cs
@@ -240,7 +240,7 @@ namespace MonoGame.Framework
 
         private void OnFormClosing(object sender, FormClosingEventArgs e)
         {
-            // TODO WHY? e.Cancel = true;
+            e.Cancel = true;
             _platform.Game.Exit();
         }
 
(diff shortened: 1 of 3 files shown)


SimonDarksideJ (approved):

SimonDarksideJ:
Looks like the issue persists @CartBlanche 
https://github.com/MonoGame/MonoGame/issues/8096#issuecomment-2816317538

ThomasFOG:
I think we shouldn't touch this and leave it be. It has the potential to regress.

Apostolique:
I think it exposed an existing issue in the WindowsDX platform. The destructor doesn't get called which is why there was a workaround for hiding the window.

MonoGame.Framework/Platform/Windows/WinFormsGameWindow.cs
```diff
@@ -240,7 +240,7 @@ private void UnregisterFromAllWindows()
 
         private void OnFormClosing(object sender, FormClosingEventArgs e)
         {
-            e.Cancel = true;
+            // TODO WHY? e.Cancel = true;
```
CartBlanche, line 243: @ThomasFOG, @SimonDarksideJ, @Jjagg Took a bit longer to track down, but this is what was stopping the Form from closing and `Dispose()` from being called. I can't see a reason why we'd want to cancel the form closing. Does anyone else know?

ThomasFOG:
@CartBlanche Do you mind reverting the changes in ```WinFormsGamePlatform```? They are unnecessary to the initial issue and prone to unexpected regressions.

CartBlanche:
@ThomasFOG I'm reluctant to revert where the Window is being destroyed, because it really should be destroyed in the destructor and not in Exit, like on other platforms. Unless you are seeing a regression?? 

How about we see what the feedback is on preview.2. We can always revert it before official 3.8.4 release, if someone hits something we've not forseen.

ThomasFOG:
I really have a problem with touching something that works and the philosophy of seeing after if something that used to work will start failing in production after that change, but let's have someone else settle on this.

CartBlanche:
@ThomasFOG I'll revert the change, so then freeing will happen in both `OnExit()` and in `Dispose()`. 
I also have an issue with something that works, but is in the wrong place. Just because it currently works, doesn't mean it is in the right place, or won't cause us issues down the line. At the very least in the future preview, we should remove what's in `OnExit()` and see what breaks for devs. That's what preview releases are for.

Issue #8701, opened 2025-04-07
MGCB Editor Issue with 3.8.3 on Linux ( and Mac )


monsieurmax:
### Prerequisites

- [x] I have verified this issue is present in the `develop` branch
- [x] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

3.8.3

### Which MonoGame platform are you using?

MonoGame Cross-Platform Desktop Application (mgdesktopgl)

### Operating System

Linux, Mac

### Description

I am unable to use the MGCB Editor in 3.8.3 on Linux.
Another user reproted the same issue on Mac

### Steps to Reproduce

Here are the steps to reproduce the issue : 
- Starting a new DesktopGL project using the templates.
- Using CLI, `dotnet tool restore` to make sure the tooling is installed for this project
- Checking `.config/dotnet-tools.json` to be set as expected
- Launching the command `dotnet mgcb-editor` result in the following error : 
`System.ComponentModel.Win32Exception (2): An error occurred trying to start process '"dotnet"' with working directory '/home/maxime/.local/share/mgcb-dotnet-tool/3.8.3.0'. No such file or directory
   at System.Diagnostics.Process.ForkAndExecProcess(ProcessStartInfo startInfo, String resolvedFilename, String[] argv, String[] envp, String cwd, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at MonoGame.Tools.Pipeline.Util.Run(String command, String arguments, String workingDirectory) in /home/runner/work/MonoGame/MonoGame/Tools/MonoGame.Content.Builder.Editor/Common/Util.cs:line 54
   at MonoGame.Tools.Pipeline.PipelineController.RestoreMGCB() in /home/runner/work/MonoGame/MonoGame/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs:line 596
[...]

PR #8729, opened 2025-04-23
[MGCB Editor] Quick fixing mgcb invocation

harry-cpp:
Should potentially Fix : https://github.com/MonoGame/MonoGame/issues/8701

### DIFF
--- a/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
+++ b/Tools/MonoGame.Content.Builder.Editor/Common/PipelineController.cs
@@ -512,11 +512,8 @@ namespace MonoGame.Tools.Pipeline
             } catch (ArgumentException) {
                 encoding = Encoding.UTF8;
             }
-            
-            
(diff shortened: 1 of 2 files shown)


CartBlanche (approved):
Tested the CI build on Windows and macOS and it runs and builds content.

CartBlanche:
@harry-cpp the only additional change I made in my PR was:

```
-            _buildProcess.OutputDataReceived += (sender, args) => output(args.Data);

+            _buildProcess.OutputDataReceived += (sender, args) =>
+            {
+                if (args.Data != null)
+                    output(args.Data);
+            };
```

Just as an extra belt and braces to avoid sending null data. Not necessarily required here, but trying to potentially avoid issues down the line.

harry-cpp:
We can't null check as if we receive a null we are still supposed to print an empty line.

The way this issue is solved is by enabling nullable reference types https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references and refactoring the code to properly handle the `?`. Random null checks can result in undesirable behaviour the same way missed null checks do.

PR #8734, opened 2025-04-25
Map BigButton on Android to Keycode.BigButton

### LINKED ISSUE #8733: Map BigButton (Xbox, PS4) on Android Gamepads
### Intent

Add the extra mapping for BigButton to Android for GamePad.GetState

Currently, the keymap is ignored but available and can be mapped to KeyCode.ButtonMode

### Motivation

BigButton is available in DesktopGL projects, bringing this in would make the behaviors closer between the platforms

theLaffa:
Fixes #8733

### Description of Change

Adds a mapping on the Android platform for gamepads to BigButton. This is internally mapped to Keycode.ButtonMode.

Tested with controllers:
Xbox One  - Registers the Xbox button
Playstation 4 - Registers PS button
Playstation 5 - Registers PS button
Switch Pro - Registers Capture button (Home button is picked up by Android system)
Gamesir x2 - Registers Xbox button

### DIFF
--- a/MonoGame.Framework/Platform/Input/GamePad.Android.cs
+++ b/MonoGame.Framework/Platform/Input/GamePad.Android.cs
@@ -110,6 +110,7 @@ namespace Microsoft.Xna.Framework.Input
 
             capabilities.HasStartButton = hasMap[14];
             capabilities.HasBackButton = hasMap[15];
+            capabilities.HasBigButton = hasMap[16];
 
             return capabilities;
         }
@@ -387,6 +388,8 @@ namespace Microsoft.Xna.Framework.Input
                     return Buttons.Start;
                 case Keycode.ButtonSelect:
                     return Buttons.Back;
+                case Keycode.ButtonMode:
+                    return Buttons.BigButton;
             }
 
             return 0;


CartBlanche:
@ThomasFOG I think this fine. It just does the mapping and doesn't affect the actual BigButton functionality. We already do this mapping in DesktopGL and have done for a while now. Thoughts?

SimonDarksideJ:
In reality, the BugButton is a reserved button on platforms that support it. It was a "fad" back in the XNA days with games like SceneIT where they had a BIG button for use.

In modern systems, this button is unusable sadly.

ThomasFOG:
Yeah, actually the "big button" is not the "xbox" button. It's the additional button that could be found on dedicated controllers such as the one for the _Scene It!_ trivia game series.

image

Like Simon noted, the "Xbox" button is reserved/unmappable on most platforms, so the design intent is to make it unavailable to avoid confusion.

We should maybe remove the big button, or mark it as deprecated / enhance the documentation about it.

theLaffa:
That makes sense, was just looking at the GL implementation and trying to keep in line with that. Happy for this to be closed off if there's no value to be taken from it.

PR #8739, opened 2025-04-28
Fix imported link paths not being nested correctly in the MGCB browser.

ben-marsh:
XNA contentproj files use Windows paths separated by backslashes. MGCB expects forward slashes to be able to nest content items correctly.

The call to PathHelper.GetRelativePath() below the lines changed use System.Uri methods, which has the effect of performing backslash -> slash conversions on the supplied source file path, so this seemed the correct location to convert the link path too.

Without this change, any contentproj that references items with a `<Link>` path will be shown at the root of the MGCB project. With this change, the items will be nested at the correct location in the browser.


SimonDarksideJ:
I pushed a patch to change the call use use the existing `PathHelper.Normalize` function rather than duplicate code, which does the same thing.

PR #8745, opened 2025-05-03
Point struct scalar multiplication/division operators

not-phoeniix:
Description of changes:
- Added `Point * int` scalar multiplication operator overload
- Added `Point / int` scalar division operator overload
- Added XML documentation for both new overloads

I figured these would prove helpful due to the existence of Vector2/3/4 float scalar operators, preventing unnecessary conversions between Points and Vector2s.


not-phoeniix:
Yeah I assumed since the similar Vector 2/3/4 counterparts don't have this exception throw I'd keep it consistent across all data types and a `System.DivideByZeroException` error would be sufficient. If y'all think an exception would be better here then let me know and I'll add it but it seems unnecessary if the vectors don't have it either.

SimonDarksideJ:
Running builds again for test, but can we include some Unit Tests @not-phoeniix to complete this PR, we are trying to ensure that anything added includes tests for future proofing.

not-phoeniix:
Howdy! I added an additional operator overload and some unit tests. Let me know if there's anything else I can do!

ThomasFOG (dismissed):

not-phoeniix:
My apologies, I accidentally pushed a commit that wasn't building on Wednesday. I just updated it and verified it builds on my system, it should pass tests now.

Dzhake:
Any chance this gets merged soon? The change is pretty small, so i'm not sure why this would be delayed until 3.8.6

SimonDarksideJ (approved):

Dzhake:
Thank you!

PR #8750, opened 2025-05-09
Add NO_AUDIO switch to build without audio

ThomasFOG:
This PR adds a ```NO_AUDIO``` constant switch to the DesktopGL, WindowsDX, Android, and iOS .csproj of MonoGame.

The reasoning for this switch is to better accommodate third-party sound engines like FMOD or Wwise by entirely disabling audio and any audio dependency that could interfere with or corrupt FMOD/Wwise (while not changing the API).

While MonoGame already has a mechanism that doesn't initialize audio unless you use it, there are still possible annoyances:
- OpenAL/NVorbis/XAudio are enforced as a dependency, even if ```PrivateAssets``` is used. Therefore users using FMOD or Wwise are still required to ship with those binaries even if they don't use the MonoGame Audio namespace at all (otherwise the app won't start);
- the non-initialization mechanism doesn't guarantee that OpenAL or XAudio aren't loaded (it entrusts static initializers for not doing so, which can't be assumed to be reliable throughout all .NET runtimes)

The ```NO_AUDIO``` switch fixes those concerns by replacing the Audio implementation with a ```NotImplemented``` version and also removes OpenAL/NVorbis/XAudio binaries/nugets dependencies.

### DIFF
--- a/MonoGame.Framework/Media/Album.cs
+++ b/MonoGame.Framework/Media/Album.cs
@@ -165,7 +165,7 @@ namespace Microsoft.Xna.Framework.Media
 
 			return this.thumbnail.ImageWithSize(new CGSize(width, height));
         }
-#elif ANDROID
+#elif ANDROID && !NO_AUDIO
         public Bitmap GetAlbumArt(int width = 0, int height = 0)
         {
             Bitmap albumArt;
@@ -197,7 +197,7 @@ namespace Microsoft.Xna.Framework.Media
         {
             return this.GetAlbumArt(220, 220);
         }
-#elif ANDROID
+#elif ANDROID && !NO_AUDIO
         public Bitmap GetThumbnail()
         {
             return this.GetAlbumArt(220, 220);
--- a/MonoGame.Framework/Media/MediaQueue.cs
+++ b/MonoGame.Framework/Media/MediaQueue.cs
@@ -118,7 +118,7 @@ namespace Microsoft.Xna.Framework.Media
 			}
 		}
 
-#if !DIRECTX
+#if !DIRECTX || NO_AUDIO
         internal void SetVolume(float volume)
         {
             int count = songs.Count;
@@ -132,7 +132,7 @@ namespace Microsoft.Xna.Framework.Media
             songs.Add(song);
         }
 
-#if !DIRECTX
+#if !DIRECTX || NO_AUDIO
         internal void Stop()
         {
             int count = songs.Count;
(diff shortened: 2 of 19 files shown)


SimonDarksideJ:
Looks good, but will need summary guides in the docs repo to describe the usage and flow of this change.

For instance, is a `#define NO_AUDIO` sufficient or would the user need to define the 'NO_AUDIO' define in their csproj?  Just to be specific.

ThomasFOG:
Indeed. If we merge it, I'll open another PR for documentation.

SimonDarksideJ (approved):

arosolino:
@ThomasFOG 

EDIT: I was wrong the code and solutions below don't work. How do I get this working :)

This is the only way I could get it working by redefining all the constants.
```
<ProjectReference Include="$(MonoGameSourceDirectory)/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj">
	<Properties>DefineConstants=OPENGL;XNADESIGNPROVIDED;DESKTOPGL;NETSTANDARD;STBSHARP_INTERNAL;NO_AUDIO</Properties>
</ProjectReference>
```

Is there another way I am not aware of?

If not maybe it should of been setup as a property?
```
<ProjectReference Include="$(MonoGameSourceDirectory)/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj">
	<Properties>NoAudio=true</Properties>
</ProjectReference>
```

Then in the MonoGame project
```
<PropertyGroup Condition="'$(NoAudio)' == 'true'">
	<DefineConstants>$(DefineConstants);NO_AUDIO</DefineConstants>
</PropertyGroup>
```

arosolino:
Ok only way I could get it working is by adding the constant directly in the OpenAL.targets

```c#
  <PropertyGroup>
    <DefineConstants>$(DefineConstants);NO_AUDIO</DefineConstants>
  </PropertyGroup>
```

Is this the intended way?

ThomasFOG:
It was made with the idea to manually add the constant into the MG's csproj.

But it can indeed be improved with something like additional params with something like this in the csproj:

```xml
<DefineConstants Condition=" '$(AdditionalConstants)' != '' ">$(DefineConstants);$(AdditionalConstants)</DefineConstants>
```

and pushing it from the CLI: ```dotnet build -p:AdditionalConstants=NO_AUDIO```

arosolino:
> It was made with the idea to manually add the constant into the MG's csproj.
> 
> But it can indeed be improved with something like additional params with something like this in the csproj:
> 
> ```
> <DefineConstants Condition=" '$(AdditionalConstants)' != '' ">$(DefineConstants);$(AdditionalConstants)</DefineConstants>
> ```
> 
> and pushing it from the CLI: `dotnet build -p:AdditionalConstants=NO_AUDIO`

Ok thanks. I don't mind manually adding it since I am maintaining a custom branch anyways.

Thanks for the help =)

PR #8753, opened 2025-05-11
Merge to master for 3.8.4.preview.2 release

CartBlanche:
REMEMBER: **DO NOT SQUASH** this PR. Change the repo setting to allow a rebase and merge and then change it back.


MonoGame.Framework/Game.cs
```diff
@@ -612,10 +612,10 @@ public void Tick()
                 OnExiting(this, exitingEventArgs);
 
                 if (!exitingEventArgs.Cancel)
-                {
+                {
+                    UnloadContent();
```
tomspilman, line 616: Do we have any unit tests that cover this change?

tomspilman (approved):

MonoGame.Framework/Game.cs
CartBlanche: @tomspilman we do not. Only way I can think we could unit test this would be to add events for Loading/Unloading contents, then we can make sure which order things are fired. Is there some other way?

Issue #8743, opened 2025-04-30
Warning: Dotnet Tool Already Installed


AristurtleDev:
### Prerequisites

- [x] I have verified this issue is present in the `develop` branch
- [x] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

Monogame 3.8.3

### Which MonoGame platform are you using?

MonoGame Cross-Platform Desktop Application (mgdesktopgl)

### Operating System

Windows 11

### Description

When executing a build of a project, the `MonoGame.Content.Builder.Task.targets` file executes the command to install the `dotnet-mgcb` tool.  If the tool is already installed, it will throw the following warning on build

```
Warning MSB3073 : The command ""dotnet" tool install  dotnet-mgcb --version 3.8.3 --tool-path ." exited with code 1.
```

Some users are reporting this thinking its an error.

### Steps to Reproduce

1. Create a new MonoGame 3.8.3 project
2. Build project

### Expected Behavior

No warning

### Resulting Behavior

Shows warning

AristurtleDev:
As an additional note to this, I'm wondering what the purpose of doing this tool install here is and why it was added.

Each template already generates the project with the `.config/dotnet-tool.json` file, and the template.configs were updated to ensure that a full restore is performed when a project is created in PR #8046 

There is also the0[ dotnet restore that is run as a task](https://github.com/MonoGame/MonoGame.Templates/blob/0b2fc382ac9430e55ff238aec42d9a343b922fe2/CSharp/content/MonoGame.Application.DesktopGL.CSharp/MGNamespace.csproj#L29) inside each project's csproj file.

The only reason to do an install behind the scenes like this in the content builder task is if the user deletes their .config directory with the tools manifest, or if they don't commit it as part of their git repo, which is i would think would user error and not something MonoGame should be responsible for.

harry-cpp:
Agreed. Stuff that needs to be done:
- Remove `dotnet tool restore` from templates
- Add `dotnet tool restore` instead of `dotnet tool install` here: 
https://github.com/MonoGame/MonoGame/blob/develop/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.targets#L93
- Only call `dotnet tool restore` once for project by creating a dummy file under Intermediate folder

[...]

PR #8755, opened 2025-05-11
Fix redundant dotnet-mgcb install warning during build

cagriclbgl:
### Summary

Fixes warning `MSB3073` caused by an unnecessary call to `dotnet tool install dotnet-mgcb` during builds, even when the tool is already installed.

### What Changed

- Modified `RestoreContentCompiler` target in `MonoGame.Content.Builder.Task.targets`
- Replaced unconditional install command with a conditional script (`check-mgcb.cmd`)
- Prevents misleading build warnings for users who already have `dotnet-mgcb` installed

### Notes

Tested with MonoGame 3.8.3 on Windows 11 using `mgdesktopgl`.  
Verified that the warning no longer appears during build and content compiles correctly.

Fixes #8355

### DIFF
--- a/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.targets
+++ b/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.targets
@@ -88,12 +88,16 @@
   <!-- Use MGCBToolAdditionalArguments to provide additional arguments to the install
    for example a path to your custom NuGet.config file.
   -->
-  <Target Name="RestoreContentCompiler" Condition=" '$(AutoRestoreMGCBTool)' == 'true' And !Exists ('$(MGCBToolDirectory)$(MGCBCommand)')">
-    <MakeDir Directories="$(MGCBToolDirectory)"/>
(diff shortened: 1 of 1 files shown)


SimonDarksideJ:
Can you re-test and confirm the issue still exists with the 3.8.4-Preview.2 release please @cagriclbgl as there was work done in this area for 3.8.4.

As far as I can see, the change simply reduces the conditions for the install, negating the need to check the directory it is installed.  The install directory was one of the things I believe was resolved.

cagriclbgl:
Thanks for the suggestion! @SimonDarksideJ 

I updated the test project to use MonoGame.Content.Builder.Task 3.8.4.2 (Preview 2) and confirmed that the MSB3073 warning related to `dotnet tool install dotnet-mgcb` no longer appears during build.

The current build failure was due to an unrelated shader compilation issue (`error X4532`) in a .fx file. So it seems the warning I was originally fixing has already been resolved in the new release 🎉

I'm happy to close the PR unless you'd like to keep the extra conditional logic for safety or compatibility.

harry-cpp:
This PR has a lot of issues:
- you did not include the cmd script in the PR
- cmd scripts cannot be included to be run by targets, they are Windows only
- the line is supposed to be removed, see: https://github.com/MonoGame/MonoGame/issues/8743#issuecomment-2842633938

Therefore closing this PR.

PR #8764, opened 2025-05-18
use a more modern implementation for CurrentPlatform

MutsiMutsi:
Fixes #

The code to check for the current platform was some ancient esoteric magic.
Surely this implementation is more understandable, succint, and perhaps more robust.

### Description of Change

Instead of voodoo we simply check RuntimeInformation.IsOSPlatform.

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
 
   <PropertyGroup>
-    <TargetFramework>net452</TargetFramework>
+    <TargetFramework>net48</TargetFramework>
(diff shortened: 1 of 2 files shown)


ThomasFOG:
ConsoleCheck needs to be ```net452```, this project ensures that [nothing breaks on consoles](https://github.com/MonoGame/MonoGame/blob/develop/CONTRIBUTING.md#code-guidelines).

It is fine to modernize other platforms, although this hits another guideline: we try to avoid modernizing code for the sake of modernizing code if the changes don't solve existing issues. This is to avoid unattended regressions (and as harmless changes may be, we can never know how and when it could cascade).

MutsiMutsi:
@ThomasFOG fair enough thanks for taking the time to check if our in such short time. I was writing a custom custom importer, which had to check Linux, came across this impl and was able to replace it with a oneliner, figured id PR it into MG but I understand your arguments, so this can be closed at your discretion!

ThomasFOG:
It's a good intent and please be welcome to contribute, we know that these rules can feel backward and we're working toward improving this.

PR #8765, opened 2025-05-19
Update Cake build to include specific runtime platforms when building tools

SimonDarksideJ:
### Description of Change

Enhancements to address nuget packaging size issues where additional runtime platforms are included by default

### DIFF
--- a/build/BuildToolsTasks/BuildMGCBEditorTask.cs
+++ b/build/BuildToolsTasks/BuildMGCBEditorTask.cs
@@ -25,6 +25,13 @@ public sealed class BuildMGCBEditorTask : FrostingTask<BuildContext>
             _ => "Linux"
         };
 
+        var runtime = context.Environment.Platform.Family switch
+        {
+            PlatformFamily.Windows => context.DotNetPublishSettings.Runtime = "win-x64",
+            PlatformFamily.OSX => context.DotNetPublishSettings.Runtime = "osx",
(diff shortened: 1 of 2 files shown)


build/BuildToolsTasks/BuildMGCBEditorTask.cs
```diff
@@ -25,6 +25,13 @@ public override void Run(BuildContext context)
             _ => "Linux"
         };
 
+        var runtime = context.Environment.Platform.Family switch
```
harry-cpp, line 28: runtime variable is unused, use a normal switch

build/BuildToolsTasks/BuildMGCBEditorTask.cs
```diff
@@ -25,6 +25,13 @@ public override void Run(BuildContext context)
...
+        {
+            PlatformFamily.Windows => context.DotNetPublishSettings.Runtime = "win-x64",
```
harry-cpp, line 30: You are modifying the global context settings, this will cause issues for any step that happens after mgcb editor to use these properties.

Also checking the family will cause issues for building the linux mgcb editor on macOS / Windows.

Instead modify the packaging for each specific mgcb-editor variant to have specified runtime if it fixes the sizing issue.

SimonDarksideJ:
The journey continues, both this path and switching to using the `RuntimeIdentifier` property in the csproj, while reducing the build are NOT including the required files in the NuGet now.

Investigation continues

PR #8767, opened 2025-05-19
Add C Header generation to MGFXC

tomspilman:
The stock effects for the new native runtimes are embedded in the native MonoGame library.

Since not all platforms have a form of embedded resources (like Windows) we need to resort to header files to include a binary stream.

This PR adds the ability to generate a C header file with the embedded binary effect:

image

Once merged we will switch to using this technique for all native backends.


Tools/MonoGame.Effect.Compiler/Effect/CommandLineParser.cs
```diff
@@ -219,7 +223,14 @@ void ShowError(string message, params object[] args)
             }
             Console.Error.WriteLine(message, args);
             Console.Error.WriteLine();
-            Console.Error.WriteLine("Usage: {0} {1}", name, string.Join(" ", _requiredUsageHelp));
+            Console.Error.WriteLine("Usage: {0} {1}", name, string.Join(" ", _requiredUsageArguments));
```
tomspilman, line 226: The changes here in the command line parser allow us to show individual descriptions for required command line arguments like so:

image

ThomasFOG:
Looks good to me.

ID4-Xytem-Apple:
Gg

PR #8769, opened 2025-05-20
[MGCB Editor] Fix sizing issue

harry-cpp:
Tested on macOS only.


SimonDarksideJ (approved):
Tested and working as expected

PR #8771, opened 2025-05-21
[MGCB Editor] Experimental options to reduce the size

harry-cpp:
This will require testing of mgcb editor on all platforms as well as testing of mgcb itself.

This does not affect the macOS mgcb-editor size tho due to self containment.

### DIFF
--- a/Tools/MonoGame.Effect.Compiler/Effect/EffectObject.writer.cs
+++ b/Tools/MonoGame.Effect.Compiler/Effect/EffectObject.writer.cs
@@ -14,6 +14,38 @@ namespace MonoGame.Effect
 
         private const string Header = "MGFX";
         internal const int Version = 10;
+        
+        static int ComputeHash(Stream stream)
+        {
+            System.Diagnostics.Debug.Assert(stream.CanSeek);
+
+            unchecked
+            {
+                const int p = 16777619;
+                var hash = (int)2166136261;
+
+                var prevPosition = stream.Position;
+                stream.Position = 0;
+
+                var data = new byte[1024];
+                int length;
+                while((length = stream.Read(data, 0, data.Length)) != 0)
+                {
+                    for (var i = 0; i < length; i++)
+                        hash = (hash ^ data[i]) * p;
+                }
+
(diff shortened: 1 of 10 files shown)


Tools/MonoGame.Effect.Compiler/Effect/EffectObject.writer.cs
```diff
@@ -14,6 +14,38 @@ internal partial class EffectObject
 
         private const string Header = "MGFX";
         internal const int Version = 10;
+        
+        static int ComputeHash(Stream stream)
+        {
```
dellis1972, line 19: @harry-cpp should these changes be here? They don't seem related

harry-cpp:
This breaks mgfxc invoking dxc, I plan on making a bounty for getting a MonoGame.Tool.Dxc after fixing up some stuff with how we handle dotnet tooling stuff.

Still, safe to merge for all of our stable platforms.

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.Vulkan.cs
```diff
@@ -222,7 +223,53 @@ internal override ShaderData CreateShader(ShaderResult shaderResult, string shad
                 }
                 toolArgs += "\"" + hlslFile + "\"";
 
-                toolResult = ExternalTool.Run("dxc", toolArgs, out stdout, out stderr);
+                var processInfo = new ProcessStartInfo
```
tomspilman, line 226: For sure move this all into a shared helper method so it can be used from other shader profiles as needed.

harry-cpp: This piece of code is temporary, I have something much better in mind for it but thats for a separate PR.

> This breaks mgfxc invoking dxc, I plan on making a bounty for getting a MonoGame.Tool.Dxc after fixing up some stuff with how we handle dotnet tooling stuff.

SimonDarksideJ (approved):

PR #8774, opened 2025-05-23
Point to master instead of main.

CartBlanche:


SimonDarksideJ (approved):
Approved.
I pity the fool who forgot our release branch was still called master 😂

PR #8780, opened 2025-05-26
Fixed bug where asset file length wasn't being written in MG_Asset_Open

cssoftware-main:
Issue: Parameter mglong& length was not being written to. Causing returned stream to have a length of zero. Causing assets to fail to load with exception: System.IO.EndOfStreamException : Unable to read beyond the end of the stream.

### DIFF
--- a/native/monogame/common/MG_Asset.cpp
+++ b/native/monogame/common/MG_Asset.cpp
@@ -21,7 +21,6 @@ mgbool MG_Asset_Open(const char* path, MG_Asset*& handle, mglong& length)
         return false;
     }
 
-    mglong prevPos = ftell(handle->file);
     if (fseek(handle->file, 0, SEEK_END) != 0)
     {
         //unable to seek file for some reason
@@ -31,7 +30,7 @@ mgbool MG_Asset_Open(const char* path, MG_Asset*& handle, mglong& length)
 
     length = ftell(handle->file);
 
-    if (fseek(handle->file, prevPos, SEEK_SET) != 0)
+    if (fseek(handle->file, 0, SEEK_SET) != 0)
     {
         //unable to seek back to file start for some reason
         delete handle;


cssoftware-main:
Looks like the pipeline might just need to be ran again for the last commit. It appears it was unable to download wine64

`E: Failed to fetch https://dl.winehq.org/wine-builds/ubuntu/dists/jammy/main/binary-i386/Packages.xz  File has unexpected size (324120 != 322320). Mirror sync in progress? [IP: 151.101.46.217 443]
   Hashes of expected file:
    - Filesize:322320 [weak]
    - SHA512:f40a5d627c3e565b1cf98e07ba9a75a7a33f5aab46e395b48a32c87e0aefaaee6d7dd8c9d80cc9ff4b8d00c0ddc20efe6376aef665081e736c7e0e0013952cf2
    - SHA256:e6062c82166add203873bc92df6c4b489d579132bd63e170f786249a9e0a1d17
    - SHA1:c1e4bd6ec80dd3b8a706d07906925efb0b4ec07e [weak]
    - MD5Sum:bf228544e15e324e346c39613539308b [weak]
   Release file created at: Sat, 17 May 2025 13:08:54 +0000
E: Some index files failed to download. They have been ignored, or old ones used instead.`

cssoftware-main:
> I've re-run the jobs.

Looks like the ubuntu dependency step is still failing with the same issue about an unexpected wine64 filesize. Have you seen this before?

cssoftware-main:
https://forum.winehq.org/viewtopic.php?p=147445#p147445

Looks like this is at least somewhat of a known issue, perhaps sporadic.

tomspilman:
@SimonDarksideJ @harry-cpp what do we need to do here to fix the Linux build process?  This keeps coming up and is unrelated to this PR specifically.

cssoftware-main:
> @SimonDarksideJ @harry-cpp what do we need to do here to fix the Linux build process? This keeps coming up and is unrelated to this PR specifically.

How many times has it happened? It might have just been the wine package having a bad day, I see other pipelines have passed since a couple days ago. perhaps 3rd time's a charm if we just re-run it.

PR #8793, opened 2025-05-30
Reorder workflows to match runs

SimonDarksideJ:
# Description of Change

Restructure of Dependencies install to align both build and test.  Each component in its own block.

This switches each dependency into it's own block of management which makes it easier to manage each section, especially for the build pass, rather than a single block of scripts.

This aligns what was happening in the Test path already which seemed more stable.


CartBlanche (changes requested):
@SimonDarksideJ could we move the ifs to the top, so it's easier to tell when this block is executing?

CartBlanche (approved):
LGTM

PR #8809, opened 2025-06-02
Extend `PlayerIndex` Enum To Support 8 Players On Compatible Platforms

AristurtleDev:
### Summary

This PR extends the `PlayerIndex` enum to support up to 8 players on platforms that can handle more than 4 controllers simultaneously.  The change uses preprocessor directives to maintain platform-specific limitations.

### Changes

- Added `PlayerIndex.Five` through `PlayerIndex.Eight` (values 4-7).
- Implemented preprocessor directives `#if !PLAYSTATION4 && !PLAYSTATION5` to exclude additional values on Playstation platforms as they only support a maximum of four controllers.


AristurtleDev:
> I remember @KelsamGames saying DesktopGL can have up to 16 controllers: https://community.monogame.net/t/does-monogame-support-more-than-4-controllers/18575/3 https://muffed.kelsam.net/docs/game/engine/input/devices/#device-connection-limits

This is because of the differences in DirectInput vs XInput. XInput API documentation states that the ID is in range 0 - 3.  When using an ID outside this range with additional XInput devices they won't return back (from my understanding).

Steam has the "Extended Xbox Feature Support" in the settings which bypasses the four gamepad limit.

However i believe the new Microsoft GDK supports up to eight, which would go with with the DirectX 12 implementation from PR #8646 

@tomspilman may be able to answer if this is the case, if so, then the only four gamepad limitation at that point is the Playstation platforms.

squarebananas:
@AristurtleDev I actually meant if MonoGame.DesktopGL can have 16 controllers, then maybe increase it to `PlayerIndex.Sixteen`. I would think that will be the case for DesktopVK as well.

As for WindowsDX though (the current one without GDK), that is hardcoded to 4 controllers in multiple places:
https://github.com/MonoGame/MonoGame/blob/4629faed86a7ccbe3300361ea796dab3a9bdaec0/MonoGame.Framework/Platform/Input/GamePad.XInput.cs#L17

KelsamGames:
Should update this as a part of this change too, and probably take a look at other platforms: https://github.com/MonoGame/MonoGame/blob/4629faed86a7ccbe3300361ea796dab3a9bdaec0/MonoGame.Framework/Platform/Input/GamePad.XInput.cs#L31

SimonDarksideJ (approved):

PR #8813, opened 2025-06-03
Better Runtime Shader Compiler Errors

tomspilman:
On all our OpenGL platforms we depend on runtime shader compiling of GLSL.

If that GLSL source (currently generated from HLSL -> MojoShader) has an error you get just "Shader Compilation Failed" exception with no further details.  This was not helpful in cases where you had complex effects with lots of different passes and techniques.

This PR adds a new public type `ShaderCompilerException` which is thrown when a OpenGL shader fails to compile at runtime. (This also can be used for runtime shader compiling of HLSL on desktop at a later date).

To have enough information to diagnose issues we now pass the original shader source file path and the entrypoint method to the serialized shader data from MGFXC.

The exception now looks like this:

image

image

The exception itself contains all the useful details... the errors, the source code, the source file, the stage and entrypoint:

image

Further the `ToString()` on the exception will output all the details:

image

This should make shader development under OpenGL much easier than before.

### DIFF
--- a/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
+++ b/MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
@@ -6,7 +6,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <UseWindowsForms>true</UseWindowsForms>
     <Description>Fake console assembly that checks if MonoGame's code is still compatible with the private console implementations and building against .NET 4.5.2 and C# 5.</Description>
-    <LangVersion>5</LangVersion>
+    <LangVersion>11</LangVersion>
     <IsAotCompatible>true</IsAotCompatible>
   </PropertyGroup>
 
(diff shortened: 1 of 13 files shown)


MonoGame.Framework/MonoGame.Framework.ConsoleCheck.csproj
```diff
@@ -6,7 +6,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <UseWindowsForms>true</UseWindowsForms>
     <Description>Fake console assembly that checks if MonoGame's code is still compatible with the private console implementations and building against .NET 4.5.2 and C# 5.</Description>
-    <LangVersion>5</LangVersion>
+    <LangVersion>11</LangVersion>
```
tomspilman, line 9: @harry-cpp and i discussed this.  It seems to work fine for just language features on BRUTE, NativeAOT, and even old Mono console runtimes.

PR #8814, opened 2025-06-03
Some small improvements to WineHelper.cs

dellis1972:
There are a few additional environment variables that can be set to help with MonoGame development on Wine.

We can expand the items in `WINEDLLOVERRIDES` so that it does not try to start `explorer.exe` when running the content pipeline. Also that it does not start `services.exe` either.

The `MVK_CONFIG_LOG_LEVEL` environment variable can be set to `0` to disable the MoltenVK logging, which can be quite verbose and distracting when running the content pipeline.

Also added an entry in the launch.json file to allow the debugging of `MGFX` in VSCode.

Added a new error message to the `WineHelper` class if wine is not found. 

- [ ]  update the docs site so the error url works.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -36,6 +36,20 @@
             "stopAtEntry": false
         },
         {
+            "name": "MGFX",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "Build: mgcb",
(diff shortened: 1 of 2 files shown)


SimonDarksideJ:
https://github.com/MonoGame/docs.monogame.github.io/pull/135 is merged @dellis1972 , is this ready to go now?

SimonDarksideJ (approved):

PR #8815, opened 2025-06-03
Remove the `RestoreContentCompiler` as it was redundant.

dellis1972:
* Remove the code which installed tools in a custom folder.
* Added a new `_RestoreMGCBTool` target. This target will restore the tools, but importantly it will put a flag in the `obj` directory to track that is has done so. This means that the target will not run on every build. 
* Moved the content build from before `Build` to before `Compile` but after `ResolveProjectReferences` , this allows any dependencies to be built first. So if you reference a content pipeline extension it will be built before the content is built.

### DIFF
--- a/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.props
+++ b/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.props
@@ -8,7 +8,6 @@
     <AutoRestoreMGCBTool Condition="'$(AutoRestoreMGCBTool)' == '' And '$(MGCBCommand)' != ''">false</AutoRestoreMGCBTool>
     <AutoRestoreMGCBTool Condition="'$(AutoRestoreMGCBTool)' == ''">true</AutoRestoreMGCBTool>
     <MGCBCommand Condition="'$(MGCBCommand)' == ''">mgcb</MGCBCommand>
-    <MonoGameVersion Condition="'$(MonoGameVersion)' == ''">3.8.4.0</MonoGameVersion>
   </PropertyGroup>
 
 </Project>
--- a/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.props
+++ b/Tools/MonoGame.Content.Builder.Task/MonoGame.Content.Builder.Task.props
@@ -7,8 +7,6 @@
     <!-- If the user is using a custom $(MGCBCommand) , disable auto restore. -->
     <AutoRestoreMGCBTool Condition="'$(AutoRestoreMGCBTool)' == '' And '$(MGCBCommand)' != ''">false</AutoRestoreMGCBTool>
     <AutoRestoreMGCBTool Condition="'$(AutoRestoreMGCBTool)' == ''">true</AutoRestoreMGCBTool>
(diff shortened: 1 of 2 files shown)


dellis1972:
@SimonDarksideJ where are we with testing on this ? I can't remember?

dellis1972:
> 7. See restore / tools issues, also Android refuses to build content in a combined build using startkit template. (even though running command manually or using mgcb works)
> 
> I got numerous issues getting projects to build or even upgrade with the latest builds. The previous build seemed to work better.

Do you have the actual error messages you get?

dellis1972:
I'll take a look, that is a bit odd

dellis1972:
So this is better, the main problem I have not is `dotnet tool restore` is called for each project. As a result sometimes NuGet trips over itself and we get a file locking issue (even on macos.....) when it tries to access the .nupkg.

dellis1972:
@SimonDarksideJ I'm made some more changes, hopefully it will help a bit.

SimonDarksideJ:
> @SimonDarksideJ I'm made some more changes, hopefully it will help a bit.

Looks like this is failing to build @dellis1972

dellis1972:
@SimonDarksideJ 

```
CSC : error CS2012: Cannot open 'D:\a\MonoGame\MonoGame\MonoGame.Framework\obj\DesktopGL\Release\MonoGame.Framework.dll' for writing -- The process cannot access the file 'D:\a\MonoGame\MonoGame\MonoGame.Framework\obj\DesktopGL\Release\MonoGame.Framework.dll' because it is being used by another process.; file may be locked by '.NET Host' 
```

SimonDarksideJ:
Running test/build again, but this seemed ready to go from the last test @dellis1972 , ready to come out of draft?

SimonDarksideJ:
OK @dellis1972 , this is working as expect now with one little "migration" issue.

After updating to the nugets included here, the projects are incompatible with the existing "Target" definition in the csproj.

```
  <Target Name="RestoreDotnetTools" BeforeTargets="CollectPackageReferences">
    <Message Text="Restoring dotnet tools (this might take a while depending on your internet speed and should only happen upon building your project for the first time, or after upgrading MonoGame, or clearing your nuget cache)" Importance="High" />
    <Exec Command="dotnet tool restore" />
  </Target>
```

If the above exists in the project, it will not compile and results in the following error (for all project types)

```
 error MSB3073: The command "dotnet tool restore" exited with code 1.
```

Is this intended?  We can implement a migration guide, but this will likely be painful, as this is essentially a breaking change for all existing projects.

dellis1972:
@SimonDarksideJ 

We as its in the csproj there isn't much I can do about it, the question is why does it error? what is the underlying error message you see? (not just the exit code 1). you might need to build it with diagnostic logging to find out.

SimonDarksideJ:
OK, confirmed, and this is something that exists already if the tools have never been restored
Resolution is to either:

- Run dotnet tool restore manually once (to cache the nuget)
- Ironically, run dotnet build TWICE, something seems to happen to cache the nuget on completion, running again then works 😛

SimonDarksideJ (approved):

SimonDarksideJ:
Approved,  I'll let you do the honors and merge this one @dellis1972

PR #8823, opened 2025-06-06
Add version to mgcb output and fix Usage name output

CartBlanche:


Tools/MonoGame.Content.Builder/Program.cs
```diff
@@ -54,10 +54,13 @@ static int Main(string[] args)
 
             var content = new BuildContent();
 
+            // Get the assembly version
+            var versionString = $"v{System.Reflection.Assembly.GetExecutingAssembly().GetName().Version}";
```
harry-cpp, line 58: do:
```cs
var versionString = System.Reflection.Assembly.GetExecutingAssembly().GetName().Version;

Title = $"MonoGame Content Builder: v{versionString}\n" +
```

No need for multiple string operations that way.

Tools/MonoGame.Content.Builder/Program.cs
```diff
@@ -54,10 +54,13 @@ static int Main(string[] args)
 
             var content = new BuildContent();
 
+            // Get the assembly version
```
harry-cpp, line 57: Comment feels redundant.

PR #8830, opened 2025-06-08
[Content.Pipeline] Add new content project system

harry-cpp:
This introduces a new content builder system. The idea is to specify the content compilation from a C# project and run it to actually compile the content.

For this example lets use the Platformer2D from the samples repo as an example. To set up the use of the new content builder system:

- Create a new C# console project
- Reference the `MonoGame.Content.Pipeline` nuget/package
- Reference the respective `MonoGame.Framework` nuget/package
- Add the following code to it.

**Program.cs:**

```csharp
using ContentProject;
using Microsoft.Xna.Framework.Content.Pipeline;
using MonoGame.Framework.Content.Pipeline.Builder;

var contentCollectionArgs = new ContentBuilderParams()
{
    Mode = ContentBuilderMode.Builder,
    WorkingDirectory = "", // path to where your content folder can be located
    SourceDirectory = "Content", // Not actually needed as this is the default, but added for reference
    OutputDirectory = "", // Defaults to "bin/Content" of the Working Directory
    Platform = TargetPlatform.DesktopGL
};
var contentCollector = new MyContentCollector();
contentCollector.Run(contentCollectionArgs); // alternativly just pass args to read from command line
```

**MyCollector.cs:**

```csharp
using Microsoft.Xna.Framework.Content.Pipeline;
using MonoGame.Framework.Content.Pipeline.Builder;

namespace ContentProject;

public class MyContentCollector : ContentBuilder
{
    public override IContentCollection GetContentCollection()
    {
        var contentCollection = new ContentCollection();
        
        // include everything in the folder
        contentCollection.Include<RegexRule>(".");

        // override .txt files to be copied
        contentCollection.IncludeCopy<RegexRule>(".txt");

        // exclude bin / obj paths
        contentCollection.Exclude<RegexRule>("bin/");
        contentCollection.Exclude<RegexRule>("obj/");
        
        return contentCollection;
    }
}

```

Afterwards run from the working directory with `Content` in it or simply override the `WorkingDirectory` and `SourceDirectory`. The content for Platformer2D should now get compiled.

This was a quick example of trying to use the system to build content. Now with that out of the way lets talk about the features the system offers:

- The content builder can be run in 2 modes, builder mode and server mode:
  - **Builder** mode is used for building all the content for the specified platform before shipping.
  - **Server** mode is used for development purposes, its goal is to allow developers to compile content as its needed instead of compiling everything ahead of time each time they compile code.
    - Later on the idea would be to upgrade this mode to allow hot reload of assets as well.
    - Right now this provides with a `ContentServer` class which can be inherited from to provide your own implementations for the server (needed for private platforms as some don't get connected to the network).
    - When launching server mode, it will launch all the available `ContentServer` types.
    - I've setup the `NetworkContentServer` class using an HTTPClient for both client and server side but this will need to be reworked when we get to hot reload.
    - To use it simply assign the `TitleContainer.ContentProvider` static property to an apropriate `IContentProvider`, ie:
        ```cs
        TitleContainer.ContentProvider = new NetworkContentProvider()
        {
            Address = "http://127.0.0.1:7771/"
        };
        ```
    - In order to avoid weird dependencies or shipping debug code, I've added a new classlib called `MonoGame.Framework.DebugAdapters` that contains the `NetworkContentProvider` class. Right now it only contains it but it can be a place to put in stuff that should not be part of MonoGame assembly that we wish to provide for debugging purposes, *caugh* imgui *caugh*.
- `ContentBuilderParams` specify the platform information for which the `ContentBuilder` should run for
  - For convinience this can also be specified by passing the cli `args` variable so that you can run the content builder with different platform parameters easilly
- The content collection for both mode is done by a `CollectContent` method as seen above by filling up a collection of type `IContentCollection`.
  - We have a special implementation for filling up this collection as its purpose is not to hold all the current filepaths but take into account any new filepaths or changes that might happen while the **Server** mode is running (see sample code above).
    - Our internal implementation is called `ContentCollection` and uses `ContentRule`s for figuring out what status the content should be compiled with by appending each rule to a list, and then passing the contents filepath through each of the rules.
    - There are 2 sets of rules for including and exluding content that we provide by default, those are `RegexRule` and `WildcardRule`
    - `WildcardRule` uses Visual Basics `Like` operator for its implementation as C# has no official one and we want consistency between platforms.
    - A custom rule can be setup by making a child class of `ContentRule`.
  - We could make an `MGCBCollection` later on that just reads all the content information from an .mgcb file to use with the new system.
- The content caching exists and is being done by `IContentCache` implementation that can be specified by the `ContentBuilder.ContentCache` property.
  - By default its enabled and set to use an internal implementation.
  - Right now the internal implementation does not track code changes of importers / processors, I'll open up an issue on how we wanna handle this as I don't like just checking for date time of assemblies as we do for the mgcb.
- Misc:
  - docs - I've documented all the publicly exposed code alredy.
  - tests - API is still in development so I thought it better to let it evolve first before adding any.
  - templates - Separate repo, I'll add some once this is merged.


MonoGame.Framework/Platform/TitleContainer.Android.cs
```diff
@@ -9,9 +9,35 @@ namespace Microsoft.Xna.Framework
 {
     partial class TitleContainer
     {
+        private static string CacheLocation { get; set; } = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), CACHE_PATH);
+
         private static Stream PlatformOpenStream(string safeName)
         {
-            return Android.App.Application.Context.Assets.Open(safeName);
+            var cachePath = Path.Combine(CacheLocation, safeName);
```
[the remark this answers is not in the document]
harry-cpp: Hold your horses for the PR description, will talk about it in there XD

MonoGame.Framework/TitleContainer.cs
```diff
@@ -3,7 +3,11 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Net.Http;
+using System.Net.Sockets;
```
tomspilman, line 9: Not all console platforms have working networking at the C# level.

We need to put any networking into platform calls.  The default implementation for PC/mobile is free to use C# networking APIs.  For console it is a different story.

MonoGame.Framework/TitleContainer.cs
```diff
@@ -13,15 +17,118 @@ namespace Microsoft.Xna.Framework
...
+        /// The location of the content server for loading assets. To use it enable <see cref="CheckContentServer"/>.
+        /// </summary>
+        /// <value><c>http://localhost:7771/</c> by default.</value>
+        public static string ContentServerAddress { get; set; }
+
+        internal static string Location { get; private set; }
+
+        private static Dictionary<string, long> ModifiedTimes { get; set; }
+
+        private static HttpClient Client { get; set; }
```
ThomasFOG, line 51: I'm wondering about the impact of referencing ```HttpClient```, it makes self-contained apps substantially bigger and untrimmable.

SimonDarksideJ: This also feels like something that ideally should ONLY run in debug mode, as you shouldn't/wouldn't use the asset server in a release build.
Which should also address the self-contained issue @ThomasFOG ?

Ideally though, rather than implementing httpclient directly, should this not use another library, that is only loaded for debug, unless that is overkill?

SimonDarksideJ:
> So how do we keep the old pipeline "builder" and the new one from getting mixed up while we're supporting both?
> 
> Seems like we have `Builder` and `PipelineBuilder` folders. Is one the old one and one the new one? Are they both used together?

I believe, from my limited understanding, simply removing the "MonoGameContentReference" should be enough.  From the discussions, the new content project is simply a referenced solution that builds alongside the main project.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
```diff
@@ -0,0 +1,390 @@
+// MonoGame - Copyright (C) MonoGame Foundation, Inc
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System.Reflection;
+using Microsoft.Xna.Framework.Content.Pipeline;
+using Microsoft.Xna.Framework.Content.Pipeline.Serialization.Compiler;
+using MonoGame.Framework.Content.Pipeline.Builder.Server;
+
+namespace MonoGame.Framework.Content.Pipeline.Builder;
```
tomspilman, line 11: `MonoGame.Framework.Content.Pipeline.Builder.ContentBuilder` sounds weird.

Should this maybe move up to `MonoGame.Framework.Content.Pipeline` and leave `MonoGame.Framework.Content.Pipeline.Builder` for the old stuff?

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderHelper.cs
```diff
@@ -0,0 +1,289 @@
...
+            }
+        }
+
+        Serializer = serilizer.Build();
+        Deserializer = deserializer.Build();
+    }
+
+    public static readonly ISerializer Serializer;
+
+    public static readonly IDeserializer Deserializer;
```
tomspilman, line 143: Does this mean the builder can only work single threaded?

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderHelper.cs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    public void WriteYaml(IEmitter emitter, object? value, Type type, ObjectSerializer serializer)
+    {
+        var color = (Color)(value ?? new Color());
+        emitter.Emit(new Scalar($"{color.R},{color.G},{color.B},{color.A}"));
+    }
+}
+
+class ContentBuilderHelper
```
tomspilman: Not sure i like having multiple classes in the same file.

harry-cpp: Those are private classes, pretty normal stuff when you don't want them exposed to autocomplete anywhere else and don't want to use a tuple.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderHelper.cs
harry-cpp: No? Not sure why that part would prevent it from working multithreaded.

As for compiling multiple content files at the same time, we should be able to set it up, the only real issue is proper logging and adding some logic to chose which files to go to which thread, which is really outside of the scope of what this is doing.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
harry-cpp: This definitely should not be top level as no importer / processor would try to access this stuff.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderHelper.cs
harry-cpp: Oh, I forgot to move the one from above into it, I see.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderHelper.cs
tomspilman: I assumed that these need to be different per asset type and being globals they would need to change per-asset.

But i'm probably misunderstanding how these are used.

SimonDarksideJ (approved):

PR #7995, opened 2023-02-26
Crop ABC values instead of truetype layout values

wldevries:
This pull request fixes a regression and a bug in the horizontal layout for SpriteFonts in MonoGame 3.8.1 and addresses issue  #7994.

* It reverts the fix that cancelled out the horizontal bearing X. 
* It fixes the cropping algorithm by adjusting the ABC values instead of the typographic values

The following screenshots shows the fonts Segoe UI, [Kingthings Petrock](https://www.fontsquirrel.com/fonts/kingthings-petrock) and Consolas, all at size 24. The first line of text is rendered using the Monogame 3.8.1. The second line is drawn with correct horizontal layout with the changes in this PR applied.

image

Issue #7994, opened 2023-02-26
Incorrect horizontal SpriteFont layout due to glyph cropping


wldevries:
Horizontal layout of SpriteFont characters is incorrect. This problem was first mentioned in #5098 and a fix was applied in https://github.com/MonoGame/MonoGame/pull/7501 causing a regression problem for other fonts. Unfortunately this fix only worked for the font mentioned in the bug. It breaks horizontal layout for fonts like _Segoe UI_ and _Consolas_.

The following screenshots shows the fonts Segoe UI, [Kingthings Petrock](https://www.fontsquirrel.com/fonts/kingthings-petrock) and Consolas, all at size 24. The first line of text is rendered using the Monogame 3.8.1. The second line is drawn with correct horizontal layout (see PR).

![image](https://user-images.githubusercontent.com/1745610/221405541-04a7d446-1012-4af9-819c-05d5df1ee2ed.png)

#### What version of MonoGame does the bug occur on:
- MonoGame 3.8.1

#### What operating system are you using:
- Windows

#### What MonoGame platform are you using:

- DesktopGL

vpenades:
We also have the same issue with Arial font:

![imagen](https://user-images.githubusercontent.com/5433822/228376412-2754c212-eadd-47e9-8c7f-de71393d40b2.png)

On top it's how it's being rendered by 3.8.1

Below is how it was originally rendered by 3.8.0

kaltinril:
I went down a rabbit hole and finally found this issue and the forum post from 2022.

This makes fonts almost useless.

MutsiMutsi:
To those that reported issues with sprites, would you be so kind to check if your specific cases are resolved in https://github.com/MonoGame/MonoGame/pull/8837 if you do not know how to build the mgcb from source. Could you kindly report your font settings, string to be rendered, and current issues with it. So that I can try it out for you instead.

MutsiMutsi:
> I do not have time for that. I submitted a PR fixing this ages ago which did not go through. See the linked PR above.

I'm sorry to hear that you're so busy, thanks for at least finding the time to reply this query! With regards to why your PR was never merged I assume it's because it introduces regression bugs and removed cropping. But it would nice if it at least could've been closed. If the balls goes rolling again I'll ask for an update on your PR.

MutsiMutsi:
[...]

PR #8837, opened 2025-06-11
Fixed the kernings in spritefont at the import.

MutsiMutsi:
Fixes #

Spritefont kernings have been broken since 3.8.0.

### Description of Change

Fixed the spritefont kernings, will need further validation.

1st is arial at size 12, built with mgcb from 3.7.1 which has nice kernings, but rather thin
2nd is arial at size 12, built with mgcb from 3.8.3 which is a bit thicker, but kernings are messed up.
3rd is arial at size 12 built with my suggested changes.

image

Let me know if you have any specific fonts to test, verifying this will be exhaustive.


MutsiMutsi:
@nkast pointed out issue https://github.com/MonoGame/MonoGame/issues/5098
And raised concerns I might break that with my fixes, which I did:
image

I will further investigate and definitely rule out the font (kingthings_petrock) itself is not broken.

MutsiMutsi:
I have reintroduced the fixes by @nkast from the aforementioned merged PR #7994 however this time only conditionally and only after the changes suggested by me. This seems to fix both cases. Let's hope there are no more issues with other fonts that I've not yet come across.

SimonDarksideJ:
Additionally, does this fix also resolve the changes mentioned in this PR?
https://github.com/MonoGame/MonoGame/pull/7995

MutsiMutsi:
@SimonDarksideJ the way I would imagine doing it is building and rendering various spritefonts to a rendertarget, then comparing this rendertarget byte for byte against a source of truth, a validated correct rendering. The issue lies with getting a source of truth, text rendering in 3.7 does not look the same as it does in 3.8, so tests will always fail, even if we have succesfully fixed kerning.

MutsiMutsi:
> Any update @MutsiMutsi on completing this so we can get it in for the first 3.8.5 preview and continue testing.

What do you reckon needs to be done to get this wrapped up, a unit test as I had described? Would the text rendering as done by this PR serve as the source of truth in that case?

MutsiMutsi:
> > > Any update @MutsiMutsi on completing this so we can get it in for the first 3.8.5 preview and continue testing.
> > 
> > 
> > What do you reckon needs to be done to get this wrapped up, a unit test as I had described? Would the text rendering as done by this PR serve as the source of truth in that case?
> 
> Yes, I believe unit tests are the only thing outstanding. Then we can evaluate further improvements later

Okidoki!

MutsiMutsi:
@SimonDarksideJ I've added a unit test for the spacings. I've confirmed it fails against the old implementation and succeeds against the new implementation. Biggest questions here are, where should we store this source of truth, right now its a hardcoded JSON string within the test. And there's the question of exposing the findfont logic, currently I've copied this from the FontDescriptionProcessor, since this is not accessible. Its mentioned as a TODO.

SimonDarksideJ:
Thanks @MutsiMutsi , however, the tests are failing on Mac.  Can you investigate please?

```text
Failed ValidateABCSpacings() [14 ms]
  Error Message:
   System.Exception : An error occured in freetype.
  Stack Trace:
     at Microsoft.Xna.Framework.Content.Pipeline.Graphics.SharpFontImporter.CheckError(Int32 error) in /Users/runner/work/MonoGame/MonoGame/MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs:line 68
   at Microsoft.Xna.Framework.Content.Pipeline.Graphics.SharpFontImporter.CreateFontFace(FT_Library* library, FontDescription options, String fontName) in /Users/runner/work/MonoGame/MonoGame/MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs:line 76
   at Microsoft.Xna.Framework.Content.Pipeline.Graphics.SharpFontImporter.Import(FontDescription options, String fontName) in /Users/runner/work/MonoGame/MonoGame/MonoGame.Framework.Content.Pipeline/Graphics/Font/SharpFontImporter.cs:line 30
   at MonoGame.Tests.ContentPipeline.SharpFontImporterTests.ValidateABCSpacings() in /Users/runner/work/MonoGame/MonoGame/Tools/MonoGame.Tools.Tests/SharpFontImporterTests.cs:line 23
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

Data collector 'Blame' message: All tests finished running, Sequence file will not be generated.
```

MutsiMutsi:
> Thanks @MutsiMutsi , however, the tests are failing on Mac. Can you investigate please?

I imagine MacOS doesn't have the arial font 😅, I have added assertions to see if path to found is found and whether that path exists, lets see what the test result comes up with. I don't think we are allowed to pack arial.ttf with the repo since that is MS licensed? Could we simply skip this test in MacOS, its in my opinion reasonable to assume there are not crossplatform differences in which field you select to populate the glyph ABC spacing fields with.

MutsiMutsi:
@SimonDarksideJ Error Message: Font: arial was not found.

SimonDarksideJ:
Yes, we do not use Arial for tests I believe as it is a Microsoft only font.  Best to include the Font.xnb you are using in the Tests/Assets folder like the other font tests @MutsiMutsi

SimonDarksideJ (approved):

PR #8839, opened 2025-06-12
Fix build error where the Effect Compiler was referening the Pipeline

dellis1972:
Fix an build error where we were using `ExternalToo.Run` from the Effect Compiler. 
Recent changes means we cannot access that method. So lets refactor the Vulkan code to use a shared `RunTool` method which can be used from DX12.


SimonDarksideJ (approved):

tomspilman:
@dellis1972 @SimonDarksideJ When did `ExternalTool.Run` get removed from MGFXC?

This broke all the console pipeline tools which depended on it.

tomspilman:
Also `ExternalTool.Run` would resolve environment vars in commands...  `RunTool` does not.

dellis1972:
Well not removed, just not referenced. Maybe a Source link would have been better

PR #8849, opened 2025-06-16
Partial fix for issue 8104 (ContentLoaders in AOT/Trimmed published executables fail to load at startup)

Lucasharskamp:
Partial fix for issue https://github.com/MonoGame/MonoGame/issues/8104
(Otherwise this change would be too large)

### Reason for submission

For practice I was trying to create published builds that are AOT compiled and trimmed - which has issues as issue 8104 noted. The issue at hand is that ContentLoaders implementations are trimmed of specific parts of their class, breaking functionality when trying to load xnb files.
2025-06-16 21_54_18-KingdomEngine - Microsoft Visual Studio

The XNB files would be at the correct location, but their loaders wouldn't function. Another is that if a ContentLoader isn't utilized, AOT would trim it, making loading it in impossible (and unnecessary, as things like _Assembly.Load_ will not happen anyway).

I've made a build where the ContentLoaders load in as required, which worked for my own project. **However**, my project is small in scope. It may be prudent to test this branch on projects with a ton of assets, to see it would break anything.

### Description of Change

In general:
1. Added XML summary documentation for methods that have been altered. (Based on what the roadmap told me is required https://docs.monogame.net/roadmap/) 
`Update to the XML API documentation, ensuring all of the API is covered. (A Massive task!)`
2. Added `[DynamicallyAccessedMembers(...)]` to specific methods where reflection is enacted for ContentLoaders.
3. For a few methods a generic type `<T>` was introduced in favor of a `Type type` parameter, because that works better with trimming/AOT compilation. (Compile time vs. runtime lookup, etc.)

In ReflectiveReader.cs
1. ReflectiveReader's `<T>` is always an implementation of ContentTypeReader and always needs specific parts of it, so I put them in the constraints/DynamicallyAccessedMembers. This is what likely fixed the bug the most.

In ContentTypeReaderManager.cs:
1. The readers are now stored by type rather than the type name for the sake of performance, and by how readers are added in code this ensures they aren't trimmed.
2. in AOT scenarios, if a loader isn't required during runtime (as no assets are utilized), the loader won't be written. This causes the reader to retrieve an empty string. if so, we can skip setting up that specific resource loader.
**EDIT:**  this may have been due to ReflectiveWriter using the Type ''blindly'', which was fixed after the results of the failed build came in. I fixed that now, and this may resolve this issue as well.

### How to reproduce 

Publish a project with trimming/AOT enabled and where assets are registered in the _.mgcb_ file. Then load those assets.

### Uncertainties / Thoughts

Here are a few uncertainties in the back of my head that one might want to take into account:
1. I do not know exactly what will happen with custom loaders. 
2. I've tested with a few images, an audio file and a spritefont. I have not tested with other resource types, custom types, or in bulk.
4. I do not know if this breaks .NET Framework 4.5 (it shouldn't?)
5. I've tested a win-x64 publish, I have no idea if this will fix the issue for Android/iOS/Linux/etc. (Although I wouldn't be surprised if it did). 
6. I've gotten successful builds for MonoGame.Tools.Windows.sln, MonoGame.Tools.Linux.sln, MonoGame.Framework.WindowsDX.sln and MonoGame.Framework.DesktopGL.sln. (But not the others as I don't have Android/iOS/etc. tooling going on)

Further:
1. I've not touched anything beyond functionality for loading content for the sake of keeping the PR simple.
3. According to a test I did, there are more than a dozen other reflection scenarios that need mediation, so I assume 8104 will be open for the foreseeable future.
4. There are calls to, for example, _MakeGenericMethod_. https://docs.monogame.net/articles/getting_started/preparing_for_consoles.html#no-runtime-compilation--il-emit However, i think that's for another day.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ReflectiveWriter.cs
+++ b/MonoGame.Framework.Content.Pipeline/Serialization/Compiler/ReflectiveWriter.cs
@@ -27,11 +27,13 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Compiler
         {
         }
 
+        /// <inheritdoc />
         public override bool CanDeserializeIntoExistingObject
         {
             get { return TargetType.IsClass; }
         }
 
+        /// <inheritdoc />
         protected override void Initialize(ContentCompiler compiler)
         {
             _compiler = compiler;
@@ -47,8 +49,8 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Serialization.Compiler
             if (typeVersion != null)
                 _typeVersion = typeVersion.TypeVersion;
 
-            _properties = TargetType.GetAllProperties().Where(IsValidProperty).ToArray();
-            _fields = TargetType.GetAllFields().Where(IsValidField).ToArray();
+            _properties = ContentExtensions.GetAllProperties<T>().Where(IsValidProperty).ToArray();
+            _fields = ContentExtensions.GetAllFields<T>().Where(IsValidField).ToArray();
         }
 
         /// <inheritdoc/>
--- a/MonoGame.Framework/Content/ContentExtensions.cs
+++ b/MonoGame.Framework/Content/ContentExtensions.cs
@@ -1,12 +1,24 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using System.Linq;
 
 namespace Microsoft.Xna.Framework.Content
 {
+    /// <summary>
+    /// Methods to retrieve information about types, such as constructors, properties, and fields.
+    /// </summary>
     internal static class ContentExtensions
     {
-        public static ConstructorInfo GetDefaultConstructor(this Type type)
+        /// <summary>
+        /// Retrieves all non-static constructors belonging to <paramref name="type"/>.
+        /// </summary>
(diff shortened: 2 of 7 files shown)


ThomasFOG:
I'm closing your PR because there is a misunderstanding about how MonoGame works with AOT:
- we can't use types as key for the constructor cache trick because this opens to assembly versions and namespace collisions
- the actual error is because the cache had a wrong type signature for ```SpriteFontReader```, I fixed this in #8850
- ```DynamicallyAccessedMemberTypes``` doesn't do what you think it does, it doesn't prevent types from being trimmed
- Reflective content is expected to fail in AOT and requires a manual ```AddTypeCreator()``` call from users. There is no way to anticipate all its usages and fix this within MonoGame itself (like other generics such as ArrayReader), DynamicallyAccessedMembers  has no effect on this
- it indeed breaks .NET 4.5 but this is a minor inconvenience as we're getting ready to remove .NET 4.5 compatibility

The reason why your PR seems to work is because you simplified the constructor cache, which makes the broken type signature to pass (but opens to type collisions). Adding DynamicallyAccessedMembers actually didn't fix this.

ThomasFOG:
Unless there are scenarios that we missed and other type signature issues like the one you encountered, I think that the AOT support is as good as it can get (without breaking stuff).

PR #8857, opened 2025-06-25
Added ToHSVL() to Color

ThomasFOG:
Proposal to add an HSV / HSL conversion to ```Color``` which can come handy for some color manipulations.

No big deal, it's just that I needed it for a project and thought backporting it.

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -1973,9 +1973,9 @@ namespace Microsoft.Xna.Framework
         /// <param name="l">Luminosity (or brightness) component from 0.0f to 1.0f.</param>
         public void ToHSVL(out float h, out float s, out float v, out float l)
         {
-            double r = c.R / 255.0;
-            double g = c.G / 255.0;
-            double b = c.B / 255.0;
+            double r = R / 255.0;
+            double g = G / 255.0;
+            double b = B / 255.0;
 
             double max = Math.Max(r, Math.Max(g, b));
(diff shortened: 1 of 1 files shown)


ThomasFOG:
If it's rejected and anybody need that, it can be just be added as an extension method:

```cshap
static class ColorExtensions
{
    public static void ToHSVL(this Color c, out double h, out double s, out double v)
    {
        h = 0.0;
        s = 0.0;

        double r = c.R / 255.0;
        double g = c.G / 255.0;
        double b = c.B / 255.0;

        double max = System.Math.Max(r, System.Math.Max(g, b));
        double min = System.Math.Min(r, System.Math.Min(g, b));
        double delta = max - min;

        // hue
        if (max == r)
            h = (60.0 * ((g - b) / delta) + 360.0) % 360.0;
        else if (max == g)
            h = (60.0 * ((b - r) / delta) + 360.0) % 360.0;
        else if (max == b)
            h = (60.0 * ((r - g) / delta) + 360.0) % 360.0;

        // saturation
        if (max != 0.0)
            s = (delta / max) * 100.0;

        // value
        v = max * 100.0;

        // luminosity
        l = (max + min) / 2.0;
    }
```

SimonDarksideJ:
> I'm ok with this. Would a `FromHSVL` be worthwhile?

This would make sense to me @ThomasFOG , if we are adding a To, we should reciprocate with a From.

We do need more real-world test case things in the framework beyond the original XNA to modernise MonoGame

tomspilman (approved):

PR #8882, opened 2025-07-08
Bump OpenAL to 1.24.3.1

dellis1972:
Bump OpenAL to 1.24.3.1.
The brings in Android 16kb alignment.
Also removes iOS iPhone from the NuGet Package, but still has the simulator. This will allow developers to build and test and release.

### DIFF
--- a/MonoGame.Framework/Platform/OpenAL.targets
+++ b/MonoGame.Framework/Platform/OpenAL.targets
@@ -5,6 +5,7 @@
 
   <!-- iOS has its own Song and Video implementation -->
   <ItemGroup Condition="$(DefineConstants.Contains('NO_AUDIO')) != 'true' And $(DefineConstants.Contains('IOS'))">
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Platform/OpenAL.targets
```diff
@@ -5,6 +5,7 @@
 
   <!-- iOS has its own Song and Video implementation -->
   <ItemGroup Condition="$(DefineConstants.Contains('NO_AUDIO')) != 'true' And $(DefineConstants.Contains('IOS'))">
+    <PackageReference Include="MonoGame.Library.OpenAL" Version="1.24.3.1" />
```
dellis1972, line 8: Added this back in as the Nuget now only includes the simulator binaries. the device binary which was causing publishing issues has been removed.

ThomasFOG:
The nuget needs to be published (we should likely add auto-publish to dependencies).

SimonDarksideJ (approved):

PR #8888, opened 2025-07-10
Dotnet Wasm aot Prototype

IrishBruse:
Currently have SDL working.

Spent some time trying to get wasm working but ive hit a roadblock something in graphicsdevice is crashing and i cant debug it as the source mapping for the breakpoint doesnt seemt to be working

image

image

Any help would be apreciated


SimonDarksideJ:
Please bear with us @IrishBruse as we are hip deep with other tasks.  We should have time after the Open Hours this month (Wednesday this week)

But bear in mind, WASM is not specifically a target we have been reaching for, which is also why KNI exists.

That being said, if we can help make this work, we will certainly consider it.

SimonDarksideJ:
I see you closed this @IrishBruse, any particular reason?
This would still be a good long term goal once Vulkan and Dx12 are stable

ThomasFOG:
I don't know if basing this work on DesktopGL is the way forward, though; since we're trying to deprecate it and clean the code base of it at some point.

Still a relevant test bed, but getting it merged will require some serious refactoring into the Native backend, and have the Native backend to target Emscripten.

PR #8889, opened 2025-07-11
Removed unused parameters in AudioEngine constructor.

Lucasharskamp:
### Description of Change

in AudioEngine.cs, the second constructor uses `TimeSpan lookAheadTime` and `string rendererId`, which are never used. I presume these were left in a long time ago, but now they can only deceive people by promising features that do not exist.

1) Removed parameters from the constructor that are not in use (causing AudioEngine to only have 1 constructor now)
2) Added a small bit of missing documentation to the method and removed documentation that is no longer needed.
3) Turned ArgumentNullException into ArgumentException (one should throw ArgumentException for empty/null strings, not ArgumentNullException which is specifically for ``null` only. See also https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception.throwifnullorempty)


MonoGame.Framework/Audio/Xact/AudioEngine.cs
```diff
@@ -50,12 +50,6 @@ internal RpcVariable[] CreateCueVariables()
         /// </summary>
         public const int ContentVersion = 39;
 
-        /// <param name="settingsFile">Path to a XACT settings file.</param>
-        public AudioEngine(string settingsFile)
```
dellis1972, line 54: not sure we can remove this public API point even if the arguments are not used. It may break the public API, so if there are compiled libraries (Nuget's) which are out there calling this method, one upgrade they will error at runtime.  

What we can do if its not used is `Obsolete` the one that does nothing, then we can remove it at a later date.
Although the plan is to integrate FAudio I think, so we might end up using these parameters in the end anyway.

dellis1972: it would be nice if this feature of look ahead could be actually implemented...

tomspilman: The best could do is mark it Obsolete and add doc comments explaining what does and doesn't work.

MonoGame.Framework/Audio/Xact/AudioEngine.cs
```diff
@@ -78,14 +72,15 @@ internal static Stream OpenStream(string filePath, bool useMemoryStream = false)
...
+        /// <summary>
+        /// Initializes a new instance of the AudioEngine class by reading out the specified XACT <paramref name="settingsFile"/>
+        /// </summary>
         /// <param name="settingsFile">Path to a XACT settings file.</param>
-        /// <param name="lookAheadTime">Determines how many milliseconds the engine will look ahead when determing when to transition to another sound.</param>
-        /// <param name="rendererId">A string that specifies the audio renderer to use.</param>
-        /// <remarks>For the best results, use a lookAheadTime of 250 milliseconds or greater.</remarks>
-        public AudioEngine(string settingsFile, TimeSpan lookAheadTime, string rendererId)
+        /// <exception cref="ArgumentException">Invoked if <paramref name="settingsFile"/> is <see cref="String.IsNullOrEmpty"/></exception>
+        public AudioEngine(string settingsFile)
```
tomspilman, line 91: Same here... we cannot remove this.

Best we can do is add into the `<remarks>` that the look ahead and render id is unused.

Lucasharskamp:
I've made a workaround @tomspilman @dellis1972 based on your feedback: in this case, the ''bad constructor'' is added as an optional constructor, to prevent an unnecessary `: this()` call if one uses the correct constructor.  (Yes, I know It won't be adding a frame to the call stack, but I just think it's cleaner to separate the wheat from the chaff)

1) Many people believe `[Obsolete(...)]` implies functionality is momentarily retained, which is not how MS intended it. (Since it says ["Marks program elements that are no longer in use."](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=net-9.0)). So I've ensured there can be no misconceptions about this by verbosely warning about it.
2) I've documented the parameters as ''not in use''; the "behavior they should be having" is irrelevant after all, and if FAudio gets integrated at a later date you may end up with another API contract entirely anyways.

**Note:** If the FAudio functionality gets added later, and ends up using the old contract, this might end up changing the functionality of the libraries in ways the library authors didn't anticipate.  So it would be prudent perhaps to add a warning about this in a future patch cycle if that is the case. This problem exists regardless of this PR.

Lucasharskamp:
Because the constructor now throws a different exception, the corresponding UnitTest has been altered to reflect this.

Lucasharskamp:
We're green! (The XML bug in the other PR bricked Android builds as I suspected), we're good to go. @tomspilman

dellis1972:
@MonoGame/maintainers thoughts on this ?

ThomasFOG (approved):

PR #8890, opened 2025-07-11
Fix XML compiler warnings.

CartBlanche:
Fix some low hanging XML compiler warnings. We have SOOO many warnings, we introduce silly ones all the time.
At some point we should try to reduce these warnings down to zero. Just so we can see real issues when we're being warned about them.

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -1965,7 +1965,7 @@ namespace Microsoft.Xna.Framework
         }
 
         /// <summary>
-        /// Calculates Hue & Saturation value from RGB, to avoid code duplication.
+        /// Calculates Hue and Saturation value from RGB, to avoid code duplication.
         /// </summary>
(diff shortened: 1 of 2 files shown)


SimonDarksideJ (approved):

Lucasharskamp:
This PR has introduced new warnings to the build, which causes errors when one wants to export the XML summaries. The change to random.cs should be reverted; it was denoted as `&lt;` to prevent the XML from being broken while displaying a '<' sign to the user (or in the online source). The change to the other file is fine.

@SimonDarksideJ 

image

SimonDarksideJ:
@CartBlanche are you able to address this post-merge issue?

CartBlanche:
@SimonDarksideJ sure

CartBlanche:
@Lucasharskamp  Fix in https://github.com/MonoGame/MonoGame/pull/8898

PR #8893, opened 2025-07-13
[Pipeline] Automatically build the native pipeline when building the main cs project

harry-cpp:


SimonDarksideJ (approved):

ThomasFOG:
I've been trying to build this locally and here's a few things to consider about this:

```xml
  <Target Name="BuildMGCPW" Condition="'$(DisableNativeBuild)' != 'True' and '$(OS)' == 'Windows_NT'" BeforeTargets="CollectPackageReferences">
    <Exec Command="premake5 vs2022" WorkingDirectory="../native/pipeline/" />
    <Exec Command="msbuild pipeline.sln /p:Configuration=$(Configuration) /p:Platform=x64" WorkingDirectory="../native/pipeline/" />
  </Target>
```

We have a local ```premake5.exe``` in ```./native/monogame```, but not in ```./native/pipeline```. We should maybe look into sharing or removing it.

This poses the question of what should be on ```PATH``` or not. Premake, MSBuild, CMake, and Make may all not be on ```PATH``` but installed along with VS. I personally don't like to have anything and everything on ```PATH``` on Windows for security concerns, but I think that we need to decide whether we should tell developers to add those to their ```PATH``` or automate their discovery with ```vswhere```. This is why I proposed https://github.com/MonoGame/MonoGame.Library.SDL/pull/10.

PR #8900, opened 2025-07-14
Fix native backends

ThomasFOG:
I'm looking into getting the native backends up and running again (just getting it to run on Windows without linkage issues).

@sepcnt I don't where you're at, and I assume that you're running into the same issues.

### DIFF
--- a/native/monogame/common/MGM.cpp
+++ b/native/monogame/common/MGM.cpp
@@ -13,11 +13,11 @@ struct MGG_Texture;
 #include <stdio.h>
 
 
-void MGM_ReadSignature(const char* filepath, MGM_SIGNATURE)
+void MGM_ReadSignature(mgbyte* filepath, MGM_SIGNATURE)
 {
 	memset(signature, 0, 16);
 
-	FILE* handle = fopen(filepath, "rb");
+	FILE* handle = fopen((const char*)filepath, "rb");
 	if (handle == nullptr)
 		return;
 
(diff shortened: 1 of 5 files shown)


tomspilman (approved):

sepcnt:
It did fix some issues, however, I am afraid that it is not an optimal solution.
Instead of changing the string type, I removed the StringInterop module totally, since we cannot do stackalloc and CoAlloc on other platforms.
You can check my branch https://github.com/sepcnt/MonoGame/tree/desktopvk
It now only left 8 unit tests across multiple platforms and remains full compatibility.

I am now running into some issues with the shader compiler and still waiting for assistance.

ThomasFOG:
We'll see where this goes! Just wanted to give you the heads up because I needed this to work in the meantime. Looking forward to all the fixes.

sepcnt:
I made a PR now: https://github.com/MonoGame/MonoGame/pull/8903

tomspilman:
> Instead of changing the string type, I removed the StringInterop module totally, since we cannot do stackalloc and CoAlloc on other platforms.

I added this specifically just 3 weeks ago.  What was the issue?

sepcnt:
> > Instead of changing the string type, I removed the StringInterop module totally, since we cannot do stackalloc and CoAlloc on other platforms.
> 
> I added this specifically just 3 weeks ago. What was the issue?

It allocates up to 4x memory and may mess up UTF8 conversion.

tomspilman:
> It allocates up to 4x memory

I don't see the concern with the extra memory.   We're not a word processor passing 1000s of words of strings.  It won't be noticed and is a small memory trade off for less CPU work passing strings at runtime.

> may mess up UTF8 conversion.

How could it do that?

Issue #8899, opened 2025-07-14
Consider providing preventative controls to help contributors in the MG repository


Lucasharskamp:
### Intent

The MonoGame libraries/tools have very specific needs and it's easy to break some dependency, even outside the solution itself.

1. It is written in  C# 5.0 because BRUTE requires it for specific scenarios
2. It needs to ensure that types can be dynamically used in a lot of scenarios
3. Changes in the API contracts can break libraries that depend on MG
4. Use of optional arguments can break ABI requirements (especially Android)
5. Specific issues that Visual Studio regards as warnings or innocent break items.

Solution: provide the user with controls to prevent them accidentally destroying something they don't understand. This can be done in several ways:

**1. Place `<LangVersion>5.0</LangVersion>` in .csproj files and/or `Directory.Build.props`**

This will ensure Visual Studio will not allow the user to write C# code with features that are for C# versions newer than C# 5.0, which includes things as simple as `nameof(x)`

**2. Use of an `.editorconfig` file (at the root folder?) with strict rules that throw errors during a build that can or will break something.**

https://learn.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options
For work, I use a _very_ strict .editorconfig file, but it works wonders;  It can be annoying, true, but it helps one deliver quality. And in the case of MG, prevent code from being used that will break functionality - whether in code style, dependent libraries, how methods or classes are utilized or for other dynamic scenarios. 

Visual Studio loads in the .editorconfig file and, where provided by the file, overrides its default setting entries with errors/warnings/diagnostics from said editorconfig file.

Here are a few rules MG would benefit from if they would throw errors: _(or warnings until the issues found are fixed in subsequent PR's, then changed into an error)_

1. [CS1570 - broken XML](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1570) In MonoGame, if a `<summary>` has malformed XML (which happened today) it bricks XML exports.
2. [CS1571 - duplicate param tag](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs1571)  (may brick XML exports)
(The same goes for CS1572, CS1573, CS1574, CS1580, CS1581, CS1582 and CS1584)
3. [CA1001 - Types that contain IDisposable items should be disposable](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/CA1001) A good way to prevent memory leaks.
[...]

PR #8903, opened 2025-07-15
Cross Platform Vulkan Native Backend Support

### LINKED ISSUE #8642: Finish off MacOS and Linux Vulkan Native Backend Support [WIP]
** This Bounty is still being drafted **

_MonoGame needs support the vulkan backend on Mac and Linux._

📃**Full details**

Context https://github.com/MonoGame/MonoGame/pull/8638.

Work has been started fixing the native Vulkan backend so that it will build and run on Mac.
We need someone to help take this [PR](https://github.com/MonoGame/MonoGame/pull/8638) over the finish line. 

The MonoGame Foundation wishes to accelerate this transition by offering a paid opportunity to undertake this task.

🔒 **Bounty status**

Open for Applications

💰 **Bounty**

$500 USD (Five hundred  US dollars).

✋ **Requirements to apply**

_Standard requirements from [our generic bounty requirements](https://monogame.net/bounties/#requirements)._

📦 **Expected delivery content**

- PreMake build script should be updated to support `gmake` generation on Mac and Linux
- The Native backend needs to be updated to 
- The BuildDesktopVK Task (https://github.com/MonoGame/MonoGame/pull/8638/files#diff-0082161fb8c9ba4cdc862150e4ea96711b13bc60c544fbf206c29fdbfeee5807) needs to be updated so that it can build Linux/Mac binaries. 
- The binaries produced should include x64 and arm64 both all the platforms. This includes windows. 

📦 **Bonus delivery content**

- The Premake build script should also work with cmake if possible. 

📅 **Delivery timeline**

As soon as possible.

sepcnt:
Fixes [https://github.com/MonoGame/MonoGame/issues/8642](https://github.com/MonoGame/MonoGame/issues/8642)

This pull request enables the native Vulkan backend for macOS and Linux and introduces significant improvements to interop safety, graphics capabilities, and build system modernization.

-----

#### Cross-Platform Vulkan Backend

The Vulkan backend can now be built and used on **macOS** and **Linux**, expanding native graphics support beyond Windows. This includes:
- macOS support via MoltenVK framework (Static Linking)
- Linux support with proper Vulkan SDK integration and SDL
- Platform-specific build configurations for all desktop platforms

#### Interop Safety & Memory Management

* **Managed String Handling:** Refactored string handling in native modules to use managed `string` types instead of unsafe `byte*` pointers, improving memory safety
* **Simplified Marshalling:** Used `[MarshalAs(UnmanagedType.LPUTF8Str)]` for improved string marshalling safety
* **Memory Management:** Enhanced error handling and memory management in graphics device and texture processing

#### Graphics & Rendering Enhancements

* **Mipmap Generation:** Added support for mipmap generation in Vulkan backend using `vkCmdBlitImage`
* **BackBuffer Operations:** Implemented `GetBackBufferData` functionality for Vulkan backend
* **Depth Operations:** Added support for `depthClipEnable` and proper depth bias scaling to match DirectX behavior
* **Buffer Management:** Enhanced buffer stride handling and flexible buffer updates with element count, stride, and total size parameters
* **Texture Support:** Extended support for compressed textures, cube map faces, and mip levels

#### Build System & Infrastructure

* **Modernized Dependencies:** Added CppNet as a submodule and removed legacy Dependencies submodule
* **Cross-Platform Builds:** Improved CMake arguments for static SDL and position-independent code
* **Platform Linking:** Fixed native linking for macOS to use proper framework references
* **Shader Compilation:** Updated shader build tasks to use header format and relative root paths
* **CI Integration:** Added DesktopVK build targets to CI pipeline

#### Improved Robustness

* **Platform-Specific Fixes:** Fixed assertion errors on Linux and macOS-specific Vulkan macro issues (Thanks @dellis1972)
* **Input Validation:** Added checks for negative values in draw operations to prevent invalid calls
* **Error Handling:** Enhanced Vulkan backend with improved error handling, object naming, and validation
* **Memory Allocation:** Improved Vulkan memory allocator integration and management

-----

### Compatibility Issues
* **SDL on macOS**: Potential threading issues with Cocoa API not invoked on main thread
* **GLIBC on Linux**: Compiled on Ubuntu 24.04, may have compatibility issues with older distributions, consider using some modern build system like `Zig` to ensure compatibility
* **MoltenVK**: Some Vulkan extensions may not be fully implemented due to Metal limitations

### Known Issues & Limitations

* **Lacks of Error Handling**: `PlatformTryDequeueMessage` has not been implemented yet, which may lead to unhandled exceptions in some cases
* **No FAudio Support**: This module has not been implemented yet
* **Shader Compilation**: The `mgcb` tool may have issues with Vulkan shader compilation
* **Platform-Specific Quirks**: Some platform-specific behaviors may differ from existing backends
* **Extension Support**: Not all Vulkan extensions may be available on all platforms due to driver/hardware limitations
* **Behavioral Discrepancy:** `ClientSizeChangedOnDeviceReset` exhibits different behavior compared to existing backends.
  * **Failing Tests:**
      * `MSAAEnabled` tests fail because `gdm.ApplyChanges()` invokes `MGG_GraphicsDevice_ResizeSwapchain`, which breaks the rendering pipeline.
      * `VertexTextureVisualTest` and `VisualTestComparisonFunction` fail due to the lack of a compiled effect for the Vulkan backend.
      * `SetDataRowPitch` fails due to an incorrect surface format.
      * `GenerateMips` is failing for reasons currently under investigation.
      * `QueryOccludedSprites` returns an incorrect result.
      * `MonoGame.Tests.GameTest+Disposal.Method_does_not_throw_after_Dispose` fails on Linux only, still under investigation.


native/monogame/vulkan/MGG_Vulkan.cpp
```diff
@@ -795,8 +914,138 @@ void MGG_GraphicsAdapter_GetInfo(MGG_GraphicsAdapter* adapter, MGG_GraphicsAdapt
 	info.CurrentDisplayMode.height = monitorInfo.rcMonitor.bottom - monitorInfo.rcMonitor.top;
 	info.CurrentDisplayMode.format = MGSurfaceFormat::Color;
 
+#elif defined(__linux__)
+	// Linux implementation using SDL2 (supports both X11 and Wayland)
+#if defined(MG_SDL2)
+	// Get the number of display modes for the primary display
+	int displayIndex = 0; // Primary display
+	int numModes = SDL_GetNumDisplayModes(displayIndex);
```
dellis1972, line 885: is it not possible to use SDL2 for both Linux and Mac (and Maybe windows?) for `MGG_GraphicsAdapter_Get` ? 
It would cut down on the need to maintain the windows/mac specific code.
//cc @harry-cpp

sepcnt: Yeah, it will be VERY tricky to deal with both Wayland and X

sepcnt: macOS and Windows counterpart however is much simple.

MonoGame.Framework/Platform/Native/Graphics.Interop.cs
```diff
@@ -148,8 +149,8 @@ internal static unsafe partial class MGG
 {
     #region Effect Resources
 
-    [DllImport(MGP.MonoGameNativeDLL, EntryPoint = "MGG_EffectResource_GetBytecode", ExactSpelling = true)]
-    public static extern void EffectResource_GetBytecode(byte* name, out byte* bytecode, out int size);
+    [DllImport(MGP.MonoGameNativeDLL, EntryPoint = "MGG_EffectResource_GetBytecode", ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+    public static extern void EffectResource_GetBytecode([MarshalAs(UnmanagedType.LPUTF8Str)] string name, out byte* bytecode, out int size);
```
tomspilman, line 153: I am unsure about these MarshalAs cases.

Do we know the performance implications of this?  In the past MarshalAs with DllImport lead to poor performance.  Has that changed?

Also i would need to test this on consoles to be sure it doesn't break us.

These are the reasons i used raw pointers to begin with.  Simple and works everywhere without questions.

MonoGame.Framework/Platform/Native/GraphicsDevice.Native.cs
```diff
@@ -2,10 +2,12 @@
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
+using MonoGame.Framework.Utilities;
+using MonoGame.Interop;
 using System;
 using System.Collections.Generic;
-using MonoGame.Interop;
-using MonoGame.Framework.Utilities;
+using System.Drawing;
```
tomspilman: What is System.Drawing here?

MonoGame.Framework/Platform/Native/GraphicsDevice.Native.cs
```diff
@@ -411,6 +420,10 @@ private int SetUserIndexBuffer<T>(T[] indexData, int indexOffset, int indexCount
     private unsafe void PlatformDrawIndexedPrimitives(PrimitiveType primitiveType, int baseVertex, int startIndex, int primitiveCount)
     {
         ApplyState(true);
+        if (baseVertex < 0)
+            baseVertex = 0;
+        if (startIndex < 0)
+            startIndex = 0;
```
tomspilman, line 425: I'm not sure this is good behavior.... to hide bugs by fixing invalid arguments.

On one hand asserts don't work as most users consume Release build Nugets... so they never see them.

Throwing on these errors means your game could fail after you ship to users because of this.

And even the check itself just erodes performance over 1000s of calls which can happen with Draw calls.

@harry-cpp @ThomasFOG ?

MonoGame.Framework/Platform/Native/Image.Interop.cs
```diff
@@ -17,11 +17,15 @@ internal static unsafe partial class MGI
     public static extern void ReadRGBA(
         byte* data,
         int dataBytes,
-        byte zeroTransparentPixels,
+        bool zeroTransparentPixels,
```
tomspilman: I did byte here on purpose... it wasn't a mistake.  

Not all platforms share the same size for bool.  It causes issues in the API generation.  Byte is 100% the same size on all platforms.

And it avoids more MarshalAs attributes which lose us performance.

MonoGame.Framework/Platform/Native/VertexInputLayout.Native.cs
```diff
@@ -65,11 +65,10 @@ public void GenerateInputElements(VertexAttribute[] inputs, out MGG_InputElement
...
 
-            var message =   "An error occurred while preparing to draw.\n\n" +
-                            "The set VertexDeclaration does not provide all the elements " +
-                            "required by the current vertex shader:\n\n\t" +
-                            string.Join(",  ", inputs.Select((x) => x.ToShaderSemantic())) + "\n\n" +
-                            "To fix the error change your VertexDeclaration or your Effect.";
+            var message = "An error occurred while preparing to draw. "
+                        + "This is probably because the current vertex declaration does not include all the elements "
+                        + "required by the current vertex shader. The current vertex declaration includes these elements: "
+                        + string.Join(", ", inputs.Select((x) => x.ToShaderSemantic())) + ".";
```
tomspilman, line 71: Please avoid changes just to fit your personal taste in formatting.  It just makes merges harder for no benefit.

Also the error message we had here was IMO more clear than this new one.

If we want to make this message better we could share the name of the Effect/vertex shader in use (which is oddly difficult to do).

MonoGame.Framework/Platform/Native/GraphicsDevice.Native.cs
sepcnt: Since Vulkan won't accept baseVertex < 0.
GraphicsDevice.DirectX.cs also do validation here, to satisfy some unit tests.
But it is also possible to move this to native.

Tests/Framework/Graphics/VertexBufferTest.cs
```diff
@@ -450,8 +450,12 @@ public void ShouldThrowHelpfulExceptionWhenVertexFormatDoesNotMatchShader()
 #else
             Assert.That(ex.Message, Is.EqualTo("An error occurred while preparing to draw. "
                 + "This is probably because the current vertex declaration does not include all the elements "
-                + "required by the current vertex shader. The current vertex declaration includes these elements: " 
+                + "required by the current vertex shader. The current vertex declaration includes these elements: "
+#if VULKAN
+                + "POSITION0."));
+#else
                 + "NORMAL0, TEXCOORD0."));
+#endif
```
tomspilman, line 458: Why is the error different here than on other platforms?

MonoGame.Framework/Platform/Native/Image.Interop.cs
sepcnt: Is the native backend also target at consoles? Then I would suppose to use a mask enum for that to enable multi native processors.

MonoGame.Framework/Platform/Native/VertexInputLayout.Native.cs
sepcnt: This is changed to pass a unit test.

sepcnt: The test limit the format of exception message.

Tests/Framework/Graphics/VertexBufferTest.cs
sepcnt: It is different from DirectX but same as OPENGL target. Due to naming convention difference between Khronos and Microsoft I guess.

native/monogame/premake5.lua
```diff
@@ -1,174 +1,120 @@
...
+    filter "system:linux"
+    pic "On"
+    filter {}
+    defines {"DLL_EXPORT"}
+    targetdir(platform_target_path)
+    targetname "monogame.native"
+    cppdialect "C++17"
+
+    files {"include/**.h", "common/**.h", "common/**.cpp"}
+    includedirs {"include", "../../external/stb"}
```
tomspilman, line 27: Please avoid reformatting things just because you like different formatting.

sepcnt: I guess we need a shared file format configuration in the repository for C#, C++ and lua

tomspilman: So this was some reformatting tool doing this? 

Is this VSCode?  @dellis1972 @harry-cpp ?  Thoughts?

sepcnt: - https://marketplace.visualstudio.com/items?itemName=sumneko.lua
- clangd
- Visual Studio itself

native/monogame/vulkan/MGG_Vulkan.cpp
ThomasFOG: But can't the SDL code also be used on mac and Windows?

MonoGame.Framework/Platform/Native/GraphicsDevice.Native.cs
ThomasFOG: There are 2 unit tests that will not like negative values. Question is: do we want to test for negative values?

The other parameters are validated and they throw exceptions, so I'm not sure if we're gaining a lot by not validating them.

Anyway, I'm fine with both: either don't validate them and remove negative values from unit tests, or validate them (although I don't see where DirectX validates them, if it does, it doesn't seem to be done by MonoGame itself).

MonoGame.Framework/Platform/Native/Image.Interop.cs
ThomasFOG: Yes, it is the whole purpose of the native API: to target any and all platforms.

That indeed means some oddities that aren't exactly the definition of "modern" or "secure". It is often more desirable to aim for raw compatibility.

MonoGame.Framework/Platform/Native/Graphics.Interop.cs
Lucasharskamp: "Has that changed?"
Answer: No; MG, because of the Brute/ABI requirements, can't use Microsoft's newer toys:
- `[LibraryImport(...)]` is .NET 7+ only (which produces source generated interop code)
- `MarshalAs` for strings still involve allocation and encoding conversion  (in most scenarios).
- .NET 5+ aggressively inlines and optimizes blittable marshalling (e.g., int, float, byte*), .NET Framework has more overhead even with blittables.
- Source generation for ComWrappers is .NET 8+
- Etc.

sepcnt: I use `MarshalAs` most for the concern of encoding conversion, these APIs almost only involves in filename. So I think it is simple enough and may not affect performance for most cases.

MonoGame.Framework/Platform/Native/Image.Interop.cs
sepcnt: Okay, I will add an enum for this:
``cpp
enum TextureProcessor
{
    ZeroTransparentPixels = 0b1,
    // etc
}
```

File: native/monogame/premake5.lua (hunk 8 above)
Comment (dellis1972) [reply native/monogame/premake5.lua:27]: we need a .editorconfig to define the formatting rules.

**Comment by sepcnt** [comment]
How about this?

**Editor Config**:

```toml
root = true

[*]
indent_style = space
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true

[*.cs]
indent_size = 4
tab_width = 4

# C# Code Style Rules (.NET 8 Best Practices)
csharp_style_namespace_declarations = file_scoped:suggestion
csharp_style_prefer_readonly_struct = true:suggestion
csharp_style_prefer_simplified_interpolation = true:suggestion
csharp_style_prefer_pattern_matching_over_as_with_null_check = true:suggestion
csharp_style_prefer_pattern_matching_over_is_with_cast_check = true:suggestion
csharp_style_prefer_conditional_expression_over_assignment = true:silent
csharp_style_prefer_conditional_expression_over_return = true:silent
csharp_style_prefer_compound_assignment = true:suggestion
csharp_style_prefer_simplified_boolean_expressions = true:suggestion
csharp_style_prefer_switch_expression = true:suggestion
csharp_style_prefer_local_over_anonymous_function = true:suggestion
csharp_style_using_directive_placement = outside_namespace:suggestion

# C# var preferences
csharp_style_var_for_built_in_types = true:suggestion
csharp_style_var_when_type_is_apparent = true:suggestion
csharp_style_var_elsewhere = false:suggestion

# C# expression-bodied members
csharp_style_expression_bodied_methods = when_on_single_line:silent
csharp_style_expression_bodied_constructors = false:silent
csharp_style_expression_bodied_operators = false:silent
csharp_style_expression_bodied_properties = true:silent
csharp_style_expression_bodied_indexers = true:silent
csharp_style_expression_bodied_accessors = true:silent

# C# braces preferences
csharp_prefer_braces = true:suggestion

# .NET Code Quality Rules (Analyzers)
dotnet_analyzer_diagnostic.severity = warning
dotnet_code_quality_analyzer.nosymbol.severity = none

# General Rules
dotnet_diagnostic.CA1014.severity = none
dotnet_diagnostic.CA1028.severity = suggestion
dotnet_diagnostic.CA1031.severity = suggestion
dotnet_diagnostic.CA1050.severity = suggestion
dotnet_diagnostic.CA1062.severity = error
dotnet_diagnostic.CA1303.severity = none
dotnet_diagnostic.CA1707.severity = warning
dotnet_diagnostic.CA1724.severity = suggestion
dotnet_diagnostic.CA1812.severity = suggestion
dotnet_diagnostic.CA1822.severity = warning
dotnet_diagnostic.CA2007.severity = warning
dotnet_diagnostic.CA2225.severity = suggestion

# Naming Rules
dotnet_naming_rule.public_members_should_be_pascal_case.symbols = public_symbols
dotnet_naming_rule.public_members_should_be_pascal_case.style = pascal_case_style
dotnet_naming_rule.public_members_should_be_pascal_case.severity = suggestion

dotnet_naming_symbols.public_symbols.applicable_kinds = property, method, field, event, delegate
dotnet_naming_symbols.public_symbols.applicable_accessibilities = public, internal, protected, protected_internal
dotnet_naming_style.pascal_case_style.capitalization = pascal_case

dotnet_naming_rule.private_fields_should_have_underscore.symbols = private_fields
dotnet_naming_rule.private_fields_should_have_underscore.style = underscore_prefix_style
dotnet_naming_rule.private_fields_should_have_underscore.severity = suggestion

dotnet_naming_symbols.private_fields.applicable_kinds = field
dotnet_naming_symbols.private_fields.applicable_accessibilities = private
dotnet_naming_style.underscore_prefix_style.required_prefix = _
dotnet_naming_style.underscore_prefix_style.capitalization = camel_case

[*.{cpp,h,hpp,c,cc,hh}]
indent_style = space
indent_size = 4

[*.lua]
indent_style = space
indent_size = 2
```

**Clang Format**:

```yaml
# .clang-format
Language: Cpp
BasedOnStyle: Microsoft
IndentWidth: 4
TabWidth: 4
UseTab: Never
ColumnLimit: 120
```

But I think it will be better to discuss it later and integrate in the CI/CD in another PR.

native/monogame/premake5.lua
Lucasharskamp: I've written yesterday about this, should also help with ensuring specific bugs don't enter the system unintentionally. (Which normally MSBuild regards as innocuous or warnings)
https://github.com/MonoGame/MonoGame/issues/8899

Feel free to add (some of them) to the .editorconfig list @sepcnt in your comment

MonoGame.Framework/Platform/Native/GraphicsDevice.Native.cs
sepcnt: This involves multiple backend so I think it is out of scope here, but worth discussing later.

harry-cpp:
Please remove any file built by RebuildMGFX.bat files from your commits, those will be getting built separately from this PR.

sepcnt:
> Please remove any file built by RebuildMGFX.bat files from your commits, those will be getting built separately from this PR.

Ok, but this will also breaks the cicd. Doesn't matter here?

harry-cpp:
I'll set it up so they are compiled by the build bot for both testing and submission by tomorrow.

.gitignore
```diff
@@ -102,4 +102,10 @@ _*
...
 /native/**/*.make
-/native/**/Makefile
\ No newline at end of file
+/native/**/Makefile
+
+# Compiled Shaders
+/native/monogame/directx12/*.mgfxo
+/native/monogame/directx12/*.mgfxo.h
+/native/monogame/vulkan/*.mgfxo
+/native/monogame/vulkan/*.mgfxo.h
```
harry-cpp: These still need to be committed to the repo, execpt the build machine will be doing it, don't add them to gitignore.

build/BuildToolsTasks/BuildContentPipelineTask.cs
```diff
@@ -7,6 +7,17 @@ public sealed class BuildContentPipelineTask : FrostingTask<BuildContext>
 {
     public override void Run(BuildContext context)
     {
+        var buildPremake = new BuildPremake();
+        buildPremake.Run(context, "Content Pipeline", "native/pipeline", "pipeline.sln");
```
harry-cpp: Native content pipeline is being built by the MonoGame.Framework.Content.Pipeline.csproj, this is not needed.

build/Utils/StaticLibCheck.cs
```diff
@@ -184,4 +184,4 @@ public static void CheckLinux(BuildContext context, string filePath)
             throw new Exception("Invalid library linkage detected!");
         }
     }
-}
+}
```
harry-cpp: Accidental commit of line removal.

native/monogame/directx12/AlphaTestEffect.dx12.mgfxo
harry-cpp, line 1: Don't commit shader mgfxo files to be removed.

Tools/MonoGame.Effect.Compiler/Effect/Options.cs
```diff
@@ -21,5 +21,8 @@ public class Options
 
         [CommandLineParser.Name("Defines", "\t - Semicolon-delimited define assignments")]
         public string Defines;
+
+        [CommandLineParser.Name("RelativeRoot", "\t - The root path to which the source file path will be made relative in the generated header.")]
+        public string RelativeRoot = string.Empty;
```
harry-cpp: Since I see this is only used for the comment, remove it and just use `Path.GetFileName(sourceFilepath)` instead.

harry-cpp (dismissed):
Can you rebase your changes off of develop instead of merging develop on top of your changes?

Tools/MonoGame.Effect.Compiler/Effect/Options.cs
sepcnt: The path may vary on different machine if not, I would suggest adding this option.

harry-cpp: There is no reason for the full filepath to be written in the comment, only the name is what we need.

sepcnt: Actually mgfxc would write `sourceFilepath` to `// Source: `

ThomasFOG:
I've tested this PR on Windows only, and as far as I can tell it looks good to me.

I'll let @tomspilman be the judge on the native API changes, and @harry-cpp on the build script and mac/linux builds.

dellis1972:
I tested this on a Intel MacOS.

A few things. 

1. I needed to download and distribute libMoltenVK.dylib in order to the test app to run 
2. I had to put the following code which adds layer in a #if DEBUG conditional
```cplusplus
enabledLayers.push_back("VK_LAYER_KHRONOS_validation");
CheckValidationLayerSupport(enabledLayers);
```
3. The app then got further but crashes with https://gist.github.com/dellis1972/5eaa4f0f4eed6866f016769feea398c1

sepcnt:
> 1. I needed to download and distribute libMoltenVK.dylib in order to the test app to run

Need to static link MoltenVK

> 2. I had to put the following code which adds layer in a #if DEBUG conditional
> 
> ```
> enabledLayers.push_back("VK_LAYER_KHRONOS_validation");
> CheckValidationLayerSupport(enabledLayers);
> ```

Some debugging extension won't work without Vulkan SDK, same for other platform.
However `MoltenVK` will never provide validation layer.

> 3. The app then got further but crashes with https://gist.github.com/dellis1972/5eaa4f0f4eed6866f016769feea398c1

`VK_EXT_CUSTOM_BORDER_COLOR` is no longer required: https://github.com/KhronosGroup/MoltenVK/issues/1987

ThomasFOG:
We indeed need to disable validation layers on release builds.

I don't know why ```VK_EXT_custom_border_color``` was used to set clamping color (@tomspilman?). So I guess that this code can be cleaned and using ```samplerInfo.borderColor``` directly:

https://github.com/MonoGame/MonoGame/blob/f36ffc37ded5ab50f699d4c638f573c4b2bc31c8/native/monogame/vulkan/MGG_Vulkan.cpp#L3024

sepcnt:
https://github.com/libsdl-org/SDL/issues/11437

This would also break test runner on macOS

dellis1972:
> libsdl-org/SDL#11437

This should be handled by the https://github.com/MonoGame/MonoGame/blob/develop/Tests/Runner/Desktop/RunOnUIAttribute.cs#L12
and https://github.com/MonoGame/MonoGame/blob/develop/Tests/Runner/Desktop/MainThreadSynchronizationContext.cs#L12.

Tests marked with `[RunOnUI()]` should have their test code marshalled to the main UI thread. 

For example https://github.com/MonoGame/MonoGame/blob/f36ffc37ded5ab50f699d4c638f573c4b2bc31c8/Tests/Framework/Graphics/Texture2DTest.cs#L19.

sepcnt:
Now `MonoGame.Tests.DesktopVK` builds in CI
https://github.com/MonoGame/MonoGame/issues/8925

sepcnt:
To ensure compatibility on Linux, `glibc` could be replaced by `musl libc`.

ThomasFOG:
> [libsdl-org/SDL#11437](https://github.com/libsdl-org/SDL/issues/11437)
> 
> This would also break test runner on macOS

Once SDL 2.32.9 is released, I'll upgrade it (we need that for win-arm64 to build as well).

sepcnt:
> We indeed need to disable validation layers on release builds.
> 
> I don't know why `VK_EXT_custom_border_color` was used to set clamping color (@tomspilman?). So I guess that this code can be cleaned and using `samplerInfo.borderColor` directly:
> 
> https://github.com/MonoGame/MonoGame/blob/f36ffc37ded5ab50f699d4c638f573c4b2bc31c8/native/monogame/vulkan/MGG_Vulkan.cpp#L3024

Metal does not support custom border color, like GLES
https://developer.apple.com/documentation/metal/mtlsamplerbordercolor

sepcnt:
The binaries build on Ubuntu 24.04 will depend on GLIC 2.38, which may affect compatibility to some degree.
Should consider using `musl-libc` and `alpine`.

ThomasFOG:
> The binaries build on Ubuntu 24.04 will depend on GLIC 2.38, which may affect compatibility to some degree. Should consider using `musl-libc` and `alpine`.

I'll open a different issue specifically for this. EDIT: #8934

sepcnt:
> > [libsdl-org/SDL#11437](https://github.com/libsdl-org/SDL/issues/11437)
> > This would also break test runner on macOS
> 
> Once SDL 2.32.9 is released, I'll upgrade it (we need that for win-arm64 to build as well).

Since the native backend has not implemented video and audio support yet, I think it worth considering switching to SDL3.

SimonDarksideJ:
> Now `MonoGame.Tests.DesktopVK` builds in CI #8925

You beat me to it, that was my next task after the NuGet testing

ThomasFOG:
@sepcnt I found the problem with the shader generation, I've submitted a PR to fix that #8941

ThomasFOG:
```MGP_Window_SetTitle``` in the SDL implementation hasn't been updated to match the changes in ```api_MGP.h``` (which will result in a runtime crash due to missing symbol).

**Comment by sepcnt** [comment]
Some recent changes may break the mgcb on Windows. Should double check that.

**Comment by CartBlanche** [comment]
@sepcnt We're having issues confirming the changes work on macOS and Linux. Even though the CI build is green, we want to confirm at the very least the Plaformer2D sample builds and run on the all  the Desktop platforms. Hence the delay in doing the final approval of this PR.

**Comment by dellis1972** [comment]
So I tried building the VK branch from source on MacOS. It built ok.

However it does not run on mac. First problem is it still needs libMoltenVK.dylib to be present. I had to copy that from the Vulkan SDK to the executable direction and do export DYLD_LIBRARY_PATH=.:$DYLD_LIBRARY_PATH to get to load it. Should we be statically linking the MoltenVK? 

Second I get a `bus error` when trying to run and this is the following crash https://gist.github.com/dellis1972/e400f3a07c75f4e7edd6544881d162e5

**Comment by sepcnt** [comment]
> So I tried building the VK branch from source on MacOS. It built ok.
> 
> However it does not run on mac. First problem is it still needs libMoltenVK.dylib to be present. I had to copy that from the Vulkan SDK to the executable direction and do export DYLD_LIBRARY_PATH=.:$DYLD_LIBRARY_PATH to get to load it. Should we be statically linking the MoltenVK?

It's weird since the latest commit should have already statically link libMoltenVK.xcframework. I will double check that.

> Second I get a `bus error` when trying to run and this is the following crash https://gist.github.com/dellis1972/e400f3a07c75f4e7edd6544881d162e5

Seems like some memory fault.

**Comment by dellis1972** [comment]
> It's weird since the latest commit should have already statically link libMoltenVK.xcframework. I will double check that.

so the problem for me was VULKAN_SDK was pointing to the `~/VulkanSDK/1.4.304.0/` directory NOT the `~/VulkanSDK/1.4.304.0/macos` directory. Fixing that statically linked. Should we error if we cannot find the `MoltenVK.xcframework` ?

> > Second I get a `bus error` when trying to run and this is the following crash https://gist.github.com/dellis1972/e400f3a07c75f4e7edd6544881d162e5
> 
> Seems like some memory fault.

I used `lldb` to step into the code, the crash was happening when calling `vkCreateInstance`. 

I had to apply this diff https://gist.github.com/dellis1972/85d716972291c6df858b5e97aa370638 to fix the issue (and add more error handling/logging).

**Comment by sepcnt** [comment]
> > It's weird since the latest commit should have already statically link libMoltenVK.xcframework. I will double check that.
> 
> so the problem for me was VULKAN_SDK was pointing to the `~/VulkanSDK/1.4.304.0/` directory NOT the `~/VulkanSDK/1.4.304.0/macos` directory. Fixing that statically linked. Should we error if we cannot find the `MoltenVK.xcframework` ?

I guess it should be ~/VulkanSDK/1.4.304.0/macos, same as the environment variable in the CI.

**Comment by sepcnt** [comment]
> I had to apply this diff https://gist.github.com/dellis1972/85d716972291c6df858b5e97aa370638 to fix the issue (and add more error handling/logging).

Thanks, I merged you solution for the macros for macOS.
As for the error messaging, I guess it will be better to integrate with `PlatformTryDequeueMessage` in another PR.

**Comment by dellis1972** [comment]
Re-tested on mac, the latest changes are working now.

**Comment by sepcnt** [comment]
Now should also works on Linux, however it failed one more tests: `MonoGame.Tests.GameTest+Disposal.Method_does_not_throw_after_Dispose`

**Comment by ThomasFOG** [comment]
I'll try to validate Linux in the upcoming days so that the PR and the bounty can be validated.

We'll figure out tests and remaining issues independently of the PR/bounty.

**Comment by ThomasFOG** [comment]
I am unable to make this to work on Linux (Ubuntu 24.04 with confirmed Vulkan 1.4 support), I'm running into a native crash when trying to run any of the samples.

Moreover, audio content currently doesn't build because profile support for DesktopVK is missing here:

https://github.com/MonoGame/MonoGame/blob/0ac4f81ba7ed44f36b7a4dbee34069b423f3d7b9/MonoGame.Framework.Content.Pipeline/Audio/DefaultAudioProfile.cs#L18

**Comment by sepcnt** [comment]
> I am unable to make this to work on Linux (Ubuntu 24.04 with confirmed Vulkan 1.4 support), I'm running into a native crash when trying to run any of the samples.

I will try to investigate. Are you using X11 or Wayland?

> Moreover, audio content currently doesn't build because profile support for DesktopVK is missing here:
> 
> https://github.com/MonoGame/MonoGame/blob/0ac4f81ba7ed44f36b7a4dbee34069b423f3d7b9/MonoGame.Framework.Content.Pipeline/Audio/DefaultAudioProfile.cs#L18

We don't have FAudio support yet. Although I have tried to implement one, it's not included in this PR.

**Comment by dellis1972** [comment]
@sepcnt  You don't need to implement FAudio, that is going to be handled by another bounty. 
@ThomasFOG  is saying that if you build a game which targets `DesktopVK` the build fails when an audio file is included. 
You just need to add `DesktopVK` to the line in [DefaultAudioProfile.cs](https://github.com/MonoGame/MonoGame/blob/0ac4f81ba7ed44f36b7a4dbee34069b423f3d7b9/MonoGame.Framework.Content.Pipeline/Audio/DefaultAudioProfile.cs#L18)

If you haven't, I would highly recommend you create yourself a small sample which would allow you to debug issues.

**Comment by sepcnt** [comment]
While running the latest build tests, I encountered an unexpected Vulkan error:
```log
Fatal: VkResult returned 0xffffff…ff9 in vulkan/MGG_Vulkan.cpp:1219
ERROR: vkGetDeviceQueue: Invalid device [VUID-vkGetDeviceQueue-device-parameter]
```
This occurred during a call to `vkGetDeviceQueue`, where the `device` parameter was reported as invalid by the Vulkan validation layers (`VUID-vkGetDeviceQueue-device-parameter`).

I recently switched to a different Mesa Vulkan driver, so the issue may be related to that change.
Mesa extension support status: https://mesamatrix.net/

Based on the extension table, I suspect it might be related to the lack of [`VK_EXT_custom_border_color`](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_EXT_custom_border_color.html) support.
I plan to make this extension optional at runtime to improve compatibility.

ThomasFOG:
It's maybe a driver specificity.

On Windows, I have a few unconsequential validation layer errors (at least with an NVIDIA driver), but other drivers might be picky with it. I found the Windows NVIDIA driver to be laxist in some areas.
On Linux, I'm on mesa / lavapipe (software vulkan) + wayland. vkinfo tells me that it supports everything we need (including VK_EXT_custom_border_color), and vkcube works. I need to check if lavapipe is capable enough.

ThomasFOG:
As far as I can tell, the crash happens _before_ the graphics device creation, likely upon initializing the SDL platform, so maybe it is SDL related.

---

When working on another issue, I found something else that will crash on many hardware (and lavapipe for sure).

Swapchain frames array is initialized here with a constant size of 2:

https://github.com/MonoGame/MonoGame/blob/0ac4f81ba7ed44f36b7a4dbee34069b423f3d7b9/native/monogame/vulkan/MGG_Vulkan.cpp#L1090

but the ```presentMode``` might require more than 2 ```minFrameCount``` here, which will crash because the array is set to 2:

https://github.com/MonoGame/MonoGame/blob/0ac4f81ba7ed44f36b7a4dbee34069b423f3d7b9/native/monogame/vulkan/MGG_Vulkan.cpp#L1366


On my Windows NVIDIA driver, the ```minFrameCount``` is always 2, which works.
But on Linux with lavapipe, it's 4, which will crash.

I'm working on fixing this one.

sepcnt:
> As far as I can tell, the crash happens _before_ the graphics device creation, likely upon initializing the SDL platform, so maybe it is SDL related.

SDL may depend on a lot of libraries and I did try to fix a issue only happened on Linux. However I only tested X11: https://github.com/MonoGame/MonoGame/pull/8903/commits/d3226d3e135bc97f2f22bffa853129810dadf91b

dellis1972:
@sepcnt I tested the latest on macos and it is still working. 

I did still have the issue where if I exported 

`export VULKAN_SDK=~/VulkanSDK/1.4.304.0/`

it would fail to link MoltenVK in the dylib and the app would crash as a result.

`export VULKAN_SDK=~/VulkanSDK/1.4.304.0/macos` 

fixed that issue. S I suspect by default it won't link MoltenVK because I doubt the exported path will include the `macos` folder. You should update the premake5.lua to include the `macos` folders when refereecing the `lib` folder. Although keep in mind that we might port this to iOS at some point so at some point we will need to use `export VULKAN_SDK=~/VulkanSDK/1.4.304.0/iOS/` to build vulkan for that platform.

Regarding linux, are you able to test and debug this on Ubuntu? Even in a VM would work. I ask because Ubuntu is our baseline linux supported OS. So we need to make sure it works there.

dellis1972:
This seems to work 

```diff
git diff native/monogame/premake5.lua
diff --git a/native/monogame/premake5.lua b/native/monogame/premake5.lua
index ec0b8428e..08ddbe53a 100644
--- a/native/monogame/premake5.lua
+++ b/native/monogame/premake5.lua
@@ -61,11 +61,11 @@ function vulkan()
 
     files {"vulkan/**.h", "vulkan/**.cpp"}
 
-    includedirs {"external/vulkan-headers/include", "external/volk", "external/vma/include",
-        path.join(vulkan_sdk, "include")}
+    includedirs {"external/vulkan-headers/include", "external/volk", "external/vma/include"}
 
     filter {"system:macosx"}
-    libdirs {path.join(vulkan_sdk, "lib/MoltenVK.xcframework/macos-arm64_x86_64")}
+    includedirs { path.join(vulkan_sdk, "macos/include")}
+    libdirs {path.join(vulkan_sdk, "macos/lib/MoltenVK.xcframework/macos-arm64_x86_64")}
     links {"MoltenVK", "IOSurface.framework", "Foundation.framework", "QuartzCore.framework", "AppKit.framework"}
     filter {}
 end
```

sepcnt:
> This seems to work
> 
> ```diff
> git diff native/monogame/premake5.lua
> diff --git a/native/monogame/premake5.lua b/native/monogame/premake5.lua
> index ec0b8428e..08ddbe53a 100644
> --- a/native/monogame/premake5.lua
> +++ b/native/monogame/premake5.lua
> @@ -61,11 +61,11 @@ function vulkan()
>  
>      files {"vulkan/**.h", "vulkan/**.cpp"}
>  
> -    includedirs {"external/vulkan-headers/include", "external/volk", "external/vma/include",
> -        path.join(vulkan_sdk, "include")}
> +    includedirs {"external/vulkan-headers/include", "external/volk", "external/vma/include"}
>  
>      filter {"system:macosx"}
> -    libdirs {path.join(vulkan_sdk, "lib/MoltenVK.xcframework/macos-arm64_x86_64")}
> +    includedirs { path.join(vulkan_sdk, "macos/include")}
> +    libdirs {path.join(vulkan_sdk, "macos/lib/MoltenVK.xcframework/macos-arm64_x86_64")}
>      links {"MoltenVK", "IOSurface.framework", "Foundation.framework", "QuartzCore.framework", "AppKit.framework"}
>      filter {}
>  end
> ```

This would break the CI/CD, I suggest you to set `VULKAN_PATH` as `~/VulkanSDK/~0/macos`

sepcnt:
To enable iOS support, we need to link against `lib/MoltenVK.xcframework/ios-arm64_x86_64`.
However, this library is currently located under the `~/VulkanSDK/<version>/macos` directory.

Before committing to MobileVK — which lacks native support for Android and iOS — we should evaluate more suitable build systems.
I recommend considering **xmake**, as it can target all our intended platforms (including consoles).

Additionally, we could leverage **Zig** to resolve glibc compatibility issues efficiently.
https://github.com/MonoGame/MonoGame/issues/8934

ThomasFOG:
We'll ignore mobiles for the vulkan backend for the moment. It is preferred to iron out the desktop version, make it battle tested, and then consider the move to mobile.

For the time being, we need to stay focus on figuring out what's happening with Linux, and getting this merged as it is.

sepcnt:
I finally figured out the issue: some dependencies are missing in the CI environment or your local build environment.

For **X11** support, install the following:
```bash
sudo apt install -y libx11-dev libxext-dev libxrandr-dev libxcursor-dev \
  libxinerama-dev libxi-dev libxss-dev libxxf86vm-dev
```

For **Wayland** support, install:
```bash
sudo apt install -y libwayland-dev libwayland-egl1-mesa libwayland-cursor0 \
  libxkbcommon-dev libegl1-mesa-dev
```

Finally, make sure that the **Vulkan validation layers** are present in your Debug build so that validation can run correctly.

ThomasFOG:
I'll check this out! Might not be until the weekend, though. I got a busy week.

sepcnt:
> As far as I can tell, the crash happens _before_ the graphics device creation, likely upon initializing the SDL platform, so maybe it is SDL related.
> 
> When working on another issue, I found something else that will crash on many hardware (and lavapipe for sure).
> 
> Swapchain frames array is initialized here with a constant size of 2:
> 
> https://github.com/MonoGame/MonoGame/blob/0ac4f81ba7ed44f36b7a4dbee34069b423f3d7b9/native/monogame/vulkan/MGG_Vulkan.cpp#L1090
> 
> but the `presentMode` might require more than 2 `minFrameCount` here, which will crash because the array is set to 2:
> 
> https://github.com/MonoGame/MonoGame/blob/0ac4f81ba7ed44f36b7a4dbee34069b423f3d7b9/native/monogame/vulkan/MGG_Vulkan.cpp#L1366
> 
> On my Windows NVIDIA driver, the `minFrameCount` is always 2, which works. But on Linux with lavapipe, it's 4, which will crash.

I fixed this in the latest commit.

ThomasFOG:
This didn't work, I had to install more dependencies to get Wayland running. I just grabbed [what SDL does on their own repository](https://github.com/libsdl-org/SDL/blob/84af0bb53afab9d1f80139f6913de3afdd4e88d3/.github/workflows/release.yml#L91). Dunno which one was really missing.

> sudo apt-get install -y \
            gnome-desktop-testing libasound2-dev libpulse-dev libaudio-dev libjack-dev libsndio-dev \\
            libusb-1.0-0-dev libx11-dev libxext-dev libxrandr-dev libxcursor-dev libxfixes-dev libxi-dev \\
            libxss-dev libwayland-dev libxkbcommon-dev libdrm-dev libgbm-dev libgl1-mesa-dev libgles2-mesa-dev \\
            libegl1-mesa-dev libdbus-1-dev libibus-1.0-dev libudev-dev fcitx-libs-dev

Good news is: it works on Ubuntu now.

Bad news is: with some heavy flickering, which I'll figure out later.

@sepcnt if you can add the missing dependencies, I'll be happy to consider this PR and bounty done.

sepcnt:
> This didn't work, I had to install more dependencies to get Wayland running. I just grabbed [what SDL does on their own repository](https://github.com/libsdl-org/SDL/blob/84af0bb53afab9d1f80139f6913de3afdd4e88d3/.github/workflows/release.yml#L91). Dunno which one was really missing.

I have applied the SDL2 version: https://github.com/libsdl-org/SDL/blob/SDL2/docs/README-linux.md
However, this would conflict with existing CI. `libjack-dev` will conflict with `wine` and remove it.

.github/workflows/main.yml
```diff
@@ -68,6 +77,52 @@ jobs:
       - name: Install Fonts
         uses: monogame/monogame-actions/install-fonts@v1
 
+      - name: Check Wine
```
harry-cpp: Wine is not used for Linux / macOS vulkan stuff, remove all of these weird wine stuff here.

sepcnt: Ok, these are just for debug purpose. I will remove them soon.

sepcnt:
This should work fine with `wine`:
```bash
sudo apt-get install -y --no-install-recommends \
            libasound2-dev libpulse-dev libaudio-dev libjack-jackd2-dev libsndio-dev libsamplerate0-dev \
            libx11-dev libxext-dev libxrandr-dev libxcursor-dev libxfixes-dev libxi-dev libxss-dev \
            libwayland-dev libxkbcommon-dev libdrm-dev libgbm-dev libgl1-mesa-dev libgles2-mesa-dev \
            libegl1-mesa-dev libdbus-1-dev libibus-1.0-dev libudev-dev fcitx-libs-dev \
            libpipewire-0.3-dev libdecor-0-dev
```

.github/workflows/main.yml
```diff
@@ -67,6 +76,19 @@ jobs:
 
       - name: Install Fonts
         uses: monogame/monogame-actions/install-fonts@v1
+        
+      - name: Protect Wine runtime & install SDL2 deps on Linux
+        if: runner.os == 'Linux'
+        run: |
+          set -e
+          sudo apt-mark hold winehq-stable wine-stable wine-stable-i386
```
harry-cpp: This should not be added either.

ThomasFOG:
Build logs show that Wayland and X11 are properly built in.

```
SDL_WAYLAND                 (Wanted: ON): ON
SDL_WAYLAND_LIBDECOR        (Wanted: ON): ON
SDL_WAYLAND_LIBDECOR_SHARED (Wanted: ON): ON
SDL_WAYLAND_QT_TOUCH        (Wanted: ON): ON
SDL_WAYLAND_SHARED          (Wanted: ON): ON
SDL_X11                     (Wanted: ON): ON
SDL_X11_SHARED              (Wanted: ON): ON
SDL_X11_XCURSOR             (Wanted: ON): ON
SDL_X11_XDBE                (Wanted: ON): ON
SDL_X11_XFIXES              (Wanted: ON): ON
SDL_X11_XINPUT              (Wanted: ON): ON
SDL_X11_XRANDR              (Wanted: ON): ON
SDL_X11_XSCRNSAVER          (Wanted: ON): ON
SDL_X11_XSHAPE              (Wanted: ON): ON
```

So I guess it comes down to @harry-cpp letting us know that everything is fine with the CI setup, and we should be good to go.

harry-cpp:
I have nothing more to add for CI.

ThomasFOG (approved):

ThomasFOG:
Then it's a go! Thanks @sepcnt ! I'll ping whoever needs to be pinged for fulfilment.

PR #8905, opened 2025-07-15
Use MSBuild task instead of Exec

ThomasFOG:
For building the pipeline assembly, we can use an MSBuild task directly instead of calling msbuild manually (which may not work if you don't have msbuild on PATH).

cc @harry-cpp


harry-cpp (dismissed):

ThomasFOG:
> D:\a\MonoGame\MonoGame\native\pipeline\monogame_native_pipeline.vcxproj : error MSB4057: The target "Build" does not exist in the project.

Hmmm, why does this work locally but not on the runner...

SimonDarksideJ:
Seems it is unable to resolve something

> D:\a\MonoGame\MonoGame\native\pipeline\monogame_native_pipeline.vcxproj : error MSB4057: The target "Build" does not exist in the project.

Should this use  DefaultTargets="Build"  ? @ThomasFOG

SimonDarksideJ:
Hmm odd, I guess it defaults to "Build" if not actually stated.

ThomasFOG:
Closing this as it seems that the ```MSBuild``` task will always pick up whatever msbuild has been called in the first place, which on the CI happens to be .NET's, and not Visual Studio's.

We'll need to assume that github runners have always Visual Studio's MSBuild on their PATH, and we'll need to tell users to do the same.

PR #8910, opened 2025-07-18
[CI] Deploy step improvements

harry-cpp:
Stuff done:
- Add DOTNET_SYSTEM_NET_SECURITY_NOREVOCATIONCHECKBYDEFAULT so that macOS doesn't randomly fail on us as Dean suggested
- Fix nuget preparation on Deploy step not properly including mgpipeline library
- Make deploy step create a nuget artefact containing the combined set of nuget packages for easier testing

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -96,7 +96,7 @@ jobs:
         run: |
           dotnet run --project build/Build.csproj -- --target=Default
         env:
-          DOTNET_SYSTEM_NET_SECURITY_NOREVOCATIONCHECKBYDEFAULT: 1
+          DOTNET_SYSTEM_NET_SECURITY_NOREVOCATIONCHECKBYDEFAULT: 1 # android compilation is failing randomly due to downloads from microsofts website during it
 
       - name: Run Tools Tests
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,6 +95,8 @@ jobs:
       - name: Build
         run: |
           dotnet run --project build/Build.csproj -- --target=Default
+        env:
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yml
```diff
@@ -95,6 +95,8 @@ jobs:
       - name: Build
         run: |
           dotnet run --project build/Build.csproj -- --target=Default
+        env:
+          DOTNET_SYSTEM_NET_SECURITY_NOREVOCATIONCHECKBYDEFAULT: 1
```
ThomasFOG: Can I be picky and ask that we add a comment explaining why we added this to workaround some SSL failures? Just so we remember what this is about in 5 years. Since it's an undocumented thing...

ThomasFOG (dismissed):

.github/workflows/main.yml
SimonDarksideJ: Done

SimonDarksideJ (dismissed):

SimonDarksideJ:
Not sure why this is showing a commit change to templates?  Maybe it needs updating from develop @harry-cpp

harry-cpp:
> Not sure why this is showing a commit change to templates? Maybe it needs updating from develop @harry-cpp

Gimme a sec, I'll fix it up.

SimonDarksideJ:
:D @harry-cpp you fixed the template checkout and removed the comment I added to address @ThomasFOG query :D

.github/workflows/main.yml
SimonDarksideJ: Needs comment again @harry-cpp

SimonDarksideJ (approved):

PR #8916, opened 2025-07-22
[CI] Hook up shader compilation to build

harry-cpp:
This is the first part to making shader compilation be done on the CI.

The 2nd part will automatically commit the files so a fresh repo clone would not require shader recompilation.

~~I've removed the RebuildMGFX.bat scripts, would you like me to make a build_shaders.bat/.sh on top level of the repo by any chance? ~~

Added scripts folder to the repo root with new versions of RebuildMGFX

~~DX12 shader build step weren't working and I can't easily get to the bottom of it from macOS / Linux :(~~

This PR is needed to unblock https://github.com/MonoGame/MonoGame/pull/8903 from getting reviewed and merged

CC. @tomspilman @ThomasFOG


Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.DirectX12.cs
```diff
@@ -80,7 +80,7 @@ internal override ShaderData CreateShader(ShaderResult shaderResult, string shad
 
                 // Compile the shader once just to get reflection info.
                 string stdout, stderr;
-                var result = ExternalTool.Run("dxc", toolArgs + "\"" + inputFile + "\"", out reflectionData, out stderr);
+                var result = Dxc.Run(toolArgs + "\"" + inputFile + "\"", out reflectionData, out stderr);
```
tomspilman: @harry-cpp are we SURE this is going to work?  

Every time we've done this change we're met with months of "my content won't build" because the binary cannot be found.

In this case we're moving from the DXC installed from Windows Kit to our own.  So it MUST exist.

tomspilman:
> DX12 shader build step weren't working and I can't easily get to the bottom of it from macOS / Linux :(

Like before this PR?  Or after this PR?


> I've removed the RebuildMGFX.bat scripts, would you like me to make a build_shaders.bat/.sh on top level of the repo by any chance?

Yes.  Ideally... it would just be part of the VS build step to generate them.  But a top level batch works too.

The reason i had them separated is 99% of the time you're not working on all platforms at once.  So waiting for all the effects to rebuild when you're just working on one platform was a waste of time.

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.DirectX12.cs
harry-cpp: It's the same tech we use for all other tooling, so yes.

harry-cpp:
> Like before this PR? Or after this PR?

Even before this, tried manually running with a freshly built dxc copy.

> Yes. Ideally... it would just be part of the VS build step to generate them. But a top level batch works too.
> 
> The reason i had them separated is 99% of the time you're not working on all platforms at once. So waiting for all the effects to rebuild when you're just working on one platform was a waste of time.

Maybe we make a scripts folder top level with useful stuff to quickly run? (build_shaders_dx11.bat, build_shaders_dx12.bat, build_shaders_all.bat, etc.)

I don't like scripts being all over the place randomly.

.github/workflows/main.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Setup DotNet on MacOS
         if: runner.environment == 'github-hosted' && runner.os == 'macos'
         run: |
-          dotnet workload install android macos ios
+          dotnet workload install macos ios
```
harry-cpp, line 36: Android builds started failing on macOS again so disabled them.

They only seem to be problematic for macOS builds, Linux and Windows are building and testing Android so not that important.

ThomasFOG:
> I've removed the RebuildMGFX.bat scripts, would you like me to make a build_shaders.bat/.sh on top level of the repo by any chance?

Ideally, I would say that each individual target should build their own stock shaders as a pre-build step... and we could even remove the pre-built shaders from the repository entirely.

But I think that in the mean time, a .bat/sh will do.

> The 2nd part will automatically commit the files so a fresh repo clone would not require shader recompilation.

I have some concerns about this, I don't quite like auto-commited stuff. What would be the rule/condition of an auto binary commit? Because if we want fresh clones to be up-to-date, that virtually means auto-commiting rebuilt shaders on every PR?

But that's not a blocking question for this PR anyway.

.github/workflows/main.yml
ThomasFOG: So disabling the .NET SSL check didn't work?

harry-cpp:
> I have some concerns about this, I don't quite like auto-commited stuff. What would be the rule/condition of an auto binary commit? Because if we want fresh clones to be up-to-date, that virtually means auto-commiting rebuilt shaders on every PR?
> 
> But that's not a blocking question for this PR anyway.

It won't be a task thats gonna be a trigger, instead if we need new shaders we would manually start it from actions panel.

.github/workflows/main.yml
harry-cpp: Nop :/

ThomasFOG:
> > I have some concerns about this, I don't quite like auto-commited stuff. What would be the rule/condition of an auto binary commit? Because if we want fresh clones to be up-to-date, that virtually means auto-commiting rebuilt shaders on every PR?
> > But that's not a blocking question for this PR anyway.
> 
> It won't be a task thats gonna be a trigger, instead if we need new shaders we would manually start it from actions panel.

Ah yes cool. Basically automating what would otherwise be making a manual PR for that. Yes, let's go.

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.DirectX12.cs
SimonDarksideJ: Are we also going to have to add a "Windows Kit" installer step to the build automation, or will the DXC install be picked up by (shudder) NuGet?

tomspilman:
> Ideally, I would say that each individual target should build their own stock shaders as a pre-build step... and we could even remove the pre-built shaders from the repository entirely.

So why not do this and never commit it?  @harry-cpp

harry-cpp:
> So why not do this and never commit it? @harry-cpp

Because its gonna slow down building waaaay too much, I ain touching that for now at least :D

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.DirectX12.cs
harry-cpp: The Windows runner already comes with Windows SDK, nothing needs to be changed.

build/Build.csproj
```diff
@@ -33,8 +33,18 @@
...
+    <PackageReference Include="Microsoft.Direct3D.DXC" Version="1.8.2505.32" />
+    <PackageReference Include="MonoGame.Tool.Dxc" Version="1.8.2505.7" />
     <PackageReference Include="NuGet.Packaging" Version="6.11.1" />
     <PackageReference Include="System.Formats.Asn1" Version="9.0.0" />
   </ItemGroup>
 
+  <Target Name="CopyNuGetToolFiles" AfterTargets="Build">
+    <ItemGroup>
+      <ToolFiles Include="$(NuGetPackageRoot)Microsoft.Direct3D.DXC\1.8.2505.32\build\native\bin\x64\*.dll" />
+	    <ToolFiles Include="$(NuGetPackageRoot)Microsoft.Direct3D.DXC\1.8.2505.32\build\native\bin\x64\*.exe" />
```
ThomasFOG, line 45: I don't like the fact that we have version numbers in the path there, it's an opportunity to forget to update this when updating the nuget...

... but I assume this will not be a problem anymore since we're not relying on those nugets directly in the future, right?

ThomasFOG (approved):

ThomasFOG:
Looks good to me.

SimonDarksideJ (approved):

build/Build.csproj
harry-cpp: Only CI is relying on it, its reliance can be removed if we properly configure the Windows SDK on it, turns out @SimonDarksideJ was right about them not coming with any Windows SDK, and thats pretty easy to do with a 3rd party github action, but I thought this was simpler and faster to do.

PR #8918, opened 2025-07-22
Fix bounds calculation on blacking out pixels

### LINKED ISSUE #8917: Blacking out pixels on native uses wrong bounds
### Prerequisites

### MonoGame Version

3.8.3+

N/A

### Operating System

Console

### Description

Loading a PNG from a file stream on Native doesn't black out all pixels that are transparent, only a subset of them.

The issue is how the byte bounds are calculated. It's using pixel-space when blacking byte-space. See the source file: MGI.cpp.

I'll be making a PR to fix this shortly.

### Steps to Reproduce

The following snippet will print the bytes from a png file. I recommend using a 3*3 image as it's big enough to show the problem easily without being too many bytes to read.

```
var texture = Texture2D.FromStream(GraphicsDevice, File.OpenRead(pathToPng));
byte[] data = new byte[texture.Width * texture.Height * 4];
texture.GetData(data);
Console.WriteLine(string.Join(',', data));
```

You will find that only a quarter of the initial bytes have been zeroed.

### Minimal Example Repo

_No response_

### Expected Behavior

Expected that RGB bytes of all pixels with A = 0 are also 0.

### Resulting Behavior

Most bytes that weren't 0 before (usually 255) haven't been blacked out. Only the first quarter is.

### Files

_No response_

Vectovox:
Fixes #8917

### Description of Change
All I've done is to make sure the code uses the byte bounds instead of the pixel bounds when blacking out the image.

## Issue
Previously, the code would just stop processing after a quarter. An example with a 3x3 image gives:

w = 3
h = 3

Such that we cover pixel space indices [0,9).

However, we would then increment the index variable by 4 as if we were in byte space, terminating after 25% of the processing.

## Bonus
I also took the liberty of making the comments more concise.


tomspilman (approved):
This is fine.

Note @sepcnt has the same fix in the Vulkan PR #8903.  So  there will be merge conflicts here.

PR #8924, opened 2025-07-24
Fix type collision

ThomasFOG:
The ```DynamicallyAccessedMemberTypes``` trick to keep a .NET compatibility was made public (my mistake) and therefore will collide with the actual .NET type. So fixing this by making it ```internal```.


Lucasharskamp:
Sanity check; but the [DynamicallyAccessedMemberTypes] is used in more than 1 assembly, right? Won't this break on .NET Framework builds?

SimonDarksideJ (approved):

PR #8928, opened 2025-07-25
Feature/cake nuget testing

SimonDarksideJ:
# Summary

Added an additional test/validation step in the build automation workflow, the primary intention being to validate if projects can be created and built prior to packages being deployed.

This builds projects using the following templates:

- DesktopGL
- WIndowsDX (only on Windows)
- Android
- iOS (only on Mac)
- Blank 2D Starter kit - larger scale project
- Full Starter kit - larger project with basic content

A report is generated with each run identifying the success state and the versions of NuGet's used during testing, sourced from the local build run.


build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
```diff
@@ -0,0 +1,597 @@
...
+    {
+        var currentPlatform = context.Environment.Platform.Family;
+        var nugetSourcePath = IOPath.GetFullPath(context.NuGetsDirectory);
+        var testsPath = context.GetOutputPath("TemplateTests");
+        var projectPath = IOPath.Combine(testsPath, ProjectFolderName);
+        var nugetSourceName = "MonoGameTestSource";
+
+        try
+        {
+            // Step 0: Clean up any previous test artifacts
```
harry-cpp: These comments feel really redundant.

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
```diff
@@ -0,0 +1,597 @@
+using System.Text.RegularExpressions;
+using System.IO;
+using IOPath = System.IO.Path;
+
+namespace BuildScripts;
+
+public abstract class TestMonoGameTemplateTaskBase : FrostingTask<BuildContext>
+{
+    // Pre-compiled regex patterns for better performance
+    private static readonly Regex PackageReferenceRegex = new(@"<PackageReference\s+Include=""(MonoGame\.[^""]*)""\s+Version=""([^""]*)""\s*/?(?:\s*/>|>)", RegexOptions.Compiled);
```
harry-cpp: Use the following regex instead:
```
<PackageReference\s+?Include=""(?<packageName>MonoGame\.[^""]+?)""\s+?Version=""(?<version>[^""]+?)"".+?>
```

Then you can access `packageName` and `version` named groups instead of accessing using an index with `[0]` etc.

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
```diff
@@ -0,0 +1,597 @@
...
+        // Update packages directly via csproj editing (more reliable than dotnet add package for prerelease versions)
+        context.Information($"   📝 Updating {monoGamePackages.Count} package(s) to version {version}...");
+        
+        var updatedContent = csprojContent;
+        var successfulUpdates = 0;
+        
+        foreach (var packageName in monoGamePackages)
+        {
+            // Pre-escape the package name once for reuse
+            var escapedPackageName = Regex.Escape(packageName);
```
harry-cpp: You have the version number from the previous step, simple string.Replace(oldVersion, newVersion) should make this much cleaner.

Alternatively if we want accuracy we can use the same regex from above, a bit modified, to do the replacement of version numbers, ie.

```
(?<partA><PackageReference\s+?Include=""(?<packageName>MonoGame\.[^""]+?)""\s+?Version="")(?<version>[^""]+?)(?<partB>"".+?>)
```

Then do partA + versionNumber + partB with regex replace.

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
```diff
@@ -0,0 +1,597 @@
...
+              ]
+            }
+        """;
+
+        // Add platform-specific editor tool
+        string platformSpecificTool = platform switch
+        {
+            PlatformFamily.Windows => $$"""
+            ,
+            "dotnet-mgcb-editor-windows": {
```
harry-cpp: Ident this and lines above one tab to the right.

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
```diff
@@ -0,0 +1,597 @@
...
+        {
+            // Source might not exist, continue
+        }
+
+        // Try using CAKE's cross-platform process execution with proper argument handling
+        try
+        {
+            var addSourceSettings = new ProcessSettings
+            {
+                Arguments = new ProcessArgumentBuilder()
```
harry-cpp, line 146: Why `ProcessArgumentBuilder` here while everywhere else a string with properly escaped paths with `"`?

SimonDarksideJ: In testing, this worked better here.  Other commands caused either failures, or untrappable errors

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
SimonDarksideJ: I'll try it, but the above works

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
SimonDarksideJ: Rejected as the pattern is because of the differences between project, the "Core" project for example uses a different pattern.  SO this handles both self enclosed and listed references

SimonDarksideJ:
Changes applied @harry-cpp , back to you.

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
SimonDarksideJ: Removed

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
SimonDarksideJ: Indented

build/TestNuGetsTasks/TestMonoGameTemplateTaskBase.cs
```diff
@@ -447,18 +447,9 @@ private int UpdateProjectFile(BuildContext context, string csprojPath, string ve
...
-            var selfClosingReplacement = $@"<PackageReference Include=""{packageName}"" Version=""{version}"" />";
-            var newContent = Regex.Replace(updatedContent, selfClosingPattern, selfClosingReplacement);
-            
-            if (newContent == updatedContent)
-            {
-                var openTagReplacement = $@"${{1}}{version}${{2}}";
-                newContent = Regex.Replace(updatedContent, openTagPattern, openTagReplacement);
-            }
+            // Handle both self-closing and open tag formats properly, targeting specific package
+            var replacementPattern = $@"(?<partA><PackageReference\s+?Include=""{escapedPackageName}""\s+?Version="")(?<version>[^""]+?)(?<partB>"".*?>)";
```
harry-cpp: Move this to `PackageReferenceRegex`, its the same regex.

harry-cpp: @SimonDarksideJ ^

SimonDarksideJ: Already done @harry-cpp 
Back to you (once I resolve this templates checkout conflict)

PR #8929, opened 2025-07-25
Fix readme footnotes

ThomasFOG:
Sorry, footnotes in the readme were broken. Although I'm not sure if I like how this looks, they appear at the bottom of the readme and not at the bottom of the section.

@SimonDarksideJ What do you think? You can check how that looks [here](https://github.com/FlyingOakGames/MonoGame/tree/footnotes?tab=readme-ov-file#supported-platforms).

### DIFF
--- a/README-packages.md
+++ b/README-packages.md
@@ -39,9 +39,11 @@ We support a growing list of platforms across the desktop, mobile, and console s
   * Xbox (GDKX & XDK)
   * Nintendo Switch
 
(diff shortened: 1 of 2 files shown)


SimonDarksideJ:
~I don't think you need the links BACK from the footnotes.  The reference link from the list items is sufficient (one way)
The icons don't look great on the footnotes themselves~
Strike that, seems to be a built in feature of the GitHub markdown

SimonDarksideJ (approved):

PR #8940, opened 2025-08-03
Add all platforms to stock shader building tests

ThomasFOG:
This PR enables the tools unit tests to build the stock effects for all platforms (it previously only tested DirectX 11).

This will likely fail the tests as I don't comprehend why ```libmojoshader_64.dll``` doesn't seem to load when building OpenGL shaders (edit: this is due to #8942).

EDIT: this depends on #8939 and will need a rebase.

I also wasn't aware that the tool tests are running on Linux, so this will fail DirectX 12 shaders no matter what. I need to disable them for testing.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
+++ b/MonoGame.Framework.Content.Pipeline/ExternalTool.cs
@@ -9,6 +9,7 @@ using System.Net;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Threading;
+using Microsoft.Win32;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline
 {
@@ -158,6 +159,10 @@ namespace Microsoft.Xna.Framework.Content.Pipeline
             // Expand any environment variables.
             command = Environment.ExpandEnvironmentVariables(command);
 
+            // Add .exe if needed
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && !command.ToLowerInvariant().EndsWith(".exe"))
+                command += ".exe";
+
             // If we have a full path just pass it through.
             if (File.Exists(command))
                 return command;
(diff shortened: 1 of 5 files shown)


SimonDarksideJ (dismissed):

ThomasFOG:
It seems that DXC can't be found on CI (but is locally when I debug that on Windows). @harry-cpp any hint?

harry-cpp:
We are not bundling DXC on Windows as we can't build it, and the one Microsoft provides has a LICENSE with a clause against redistributing it.

harry-cpp:
I've used the DXC from the nuget on the CI so it does not need to install the Windows SDK, check CAKEs shader compilation.

ThomasFOG:
> We are not bundling DXC on Windows as we can't build it, and the one Microsoft provides has a LICENSE with a clause against redistributing it.

I've disabled building the DX12 shaders, but the VK shaders should build, right? It references the external tool nuget for that if I'm not mistaking. It seems that it's the nuget that can't be loaded/found (but it should be there in MGFXC and MGCB?):

> Microsoft.Xna.Framework.Content.Pipeline.InvalidContentException : Could not load file or assembly 'MonoGame.Tool.Dxc, Version=1.8.2505.7, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.

ThomasFOG:
Okay, so this is really dumb... the ```dxc.exe``` that ships with the Windows SDK isn't built with SPIR-V codegen support. But the nuget one is. (Don't mind the PR's content, I'm experimenting to make this work.)

Which means, that our only valid option to build shaders on Windows for both DX12 and Vulkan, is [the official nuget](https://www.nuget.org/packages/Microsoft.Direct3D.DXC). (Or a mix of the Windows SDK and a custom built one, but let's exclude that option.)

I can't seem to retrieve the line in the 3 different licenses that would prevent us from redistributing it. 🤔

ThomasFOG:
This PR is irrelevant now that we are building the shaders as part of building each individual target framework. It acts like a test.

PR #8941, opened 2025-08-03
Fix Vulkan constant buffer reflection data parsing

ThomasFOG:
This fixes a regression introduced by #8544 in Vulkan constant buffers parsing.

This should repair shader building and usage (notably the stock shaders).

### DIFF
--- a/Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.Vulkan.cs
+++ b/Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.Vulkan.cs
@@ -394,9 +394,29 @@ namespace MonoGame.Effect
 
                 // TODO: Support multiple constant buffers one day!
 
+                // Count how many constant buffers there are
+                int cbCount = 0;
+                string cbName = string.Empty;
+                foreach (var descriptor in descriptors)
+                {
+                    if (descriptor.Value.type == VkDescriptorType.UNIFORM_BUFFER)
+                    {
+                        cbCount++;
+                        // consider the first hit only
+                        if (cbCount <= 1)
+                            cbName = descriptor.Key.Replace("%", "%type_"); // convert into equivalent struct name
+                    }
+                }
+
+                if (cbCount > 1)
+                {
+                    errorsAndWarnings += "Building effects for Vulkan currently doesn't support more than one constant buffer (cbuffer) structures. Please consider refactoring your HLSL code.";
+                    throw new ShaderCompilerException();
+                }
+
                 // First gather the uniforms.
                 VkStruct globals;
-                if (structs.TryGetValue("%type__MG_Globals", out globals))
+                if (structs.TryGetValue(cbName, out globals))
                 {
                     foreach (var member in globals.members.Values)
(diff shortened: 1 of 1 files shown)


Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.Vulkan.cs
```diff
@@ -396,7 +396,7 @@ internal override ShaderData CreateShader(ShaderResult shaderResult, string shad
 
                 // First gather the uniforms.
                 VkStruct globals;
-                if (structs.TryGetValue("%type__Globals", out globals))
+                if (structs.TryGetValue("%type__MG_Globals", out globals))
```
tomspilman: @ThomasFOG so this `MG_Globals` comes from this line:

https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Graphics/Effect/Resources/Macros.fxh#L14

This feels like a bad choice here.   

It means all custom shaders need to use `MG_Globals` as the default constant buffer name.

(lets ignore we don't support multiple cbuffers for now).

Should we just accept the first cbuffer found regardless of name?

ThomasFOG: I would be fine with this, and that would allow shaders not built using the Macros.fxh.

I can edit that to throw an error with explanations if more than one cb is defined.

Alternatively... I could hack something like putting all constants into a unique buffer with an automatic 4-register aligned offset...

tomspilman: @ThomasFOG 

> I can edit that to throw an error with explanations if more than one cb is defined.

An error may be best as multiple buffers likely means something is broken.


> Alternatively... I could hack something like putting all constants into a unique buffer with an automatic 4-register aligned offset...

This would mean hacking at the incoming shader HLSL file? 

Seems like a lot of work and really we need to switch to our goal of supporting multiple cbuffers natively.  So maybe lets not spend time on that.

ThomasFOG: Gotcha, I'll update the PR in a moment accordingly.

ThomasFOG:
@tomspilman I made the uniform buffer discovery generic. It tries to read the first cbuffer hit, and will fail with a message if more than one cbuffer is encountered. I've tested with different cbuffer, with and without Macros.fxh, and it seems fine.

EDIT: we have a stock effect failing, investigating... TextureCube are uniform buffers on Spir-v. Checking if I can do something about that...

ThomasFOG:
@tomspilman Fixed it, stock shaders are a go. TextureCube are single buffers, so I worked around that by only gathering uniforms when there's a corresponding struct with the expected name. That should do it.

My only concern is that I don't think TextureCube will work, they are not ```_type_2d_image``` but ```_type_cube_image``` and this descriptor is currently not parsed. But I've noticed that it's not implemented in ```MGG_Texture_Create()```, so I left that as-is.

tomspilman (approved):

PR #8948, opened 2025-08-06
[Pipeline] Drop mgfx reference

harry-cpp:
Stuff done:
- Remove mgfxc.exe reference from the Content.Pipeline and instead include all the cs code
- Use freshly built mojoshader binaries that are statically linked on all platforms
- Package up Windows dxc.exe inside MonoGame.Tool.Dxc and hook it up so we can now ship dxc on all platforms
- Add ability for mgfxc to only call the needed fxc.exe call under wine while compiling OpenGL shaders
- Add more basic effect compilation unit tests that now also builds stuff for DX12 / Vulkan / OpenGL
- Drop support of automatically relaunching mgfxc under wine
- Drop out of the box support for compiling DirectX 11 shaders on macOS / Linux


SimonDarksideJ (dismissed):
Looks fine, just a concern about Wine for Mac, but otherwise good

harry-cpp:
> Looks fine, just a concern about Wine for Mac, but otherwise good

What concerns?

SimonDarksideJ:
> > Looks fine, just a concern about Wine for Mac, but otherwise good
> 
> What concerns?

The concern we had a private chat about regarding the 
```
static bool DetectWine()
        {
            string[] wineCommands = RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ?
```

Which was clarified to state it was just for changing the order for detecting Wine.

This was clarified and confirmed.

MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
```diff
@@ -54,29 +59,24 @@
...
 
   <ItemGroup>
     <ProjectReference Include="..\MonoGame.Framework\MonoGame.Framework.DesktopGL.csproj">
       <PrivateAssets>All</PrivateAssets>
     </ProjectReference>
-    <ProjectReference Include="..\Tools\MonoGame.Effect.Compiler\MonoGame.Effect.Compiler.csproj">
-      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
-      <OutputItemType>Content</OutputItemType>
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </ProjectReference>
```
tomspilman, line 70: @harry-cpp So we had this at the end of effect compiler csproj:

https://github.com/MonoGame/MonoGame/blob/develop/Tools/MonoGame.Effect.Compiler/MonoGame.Effect.Compiler.csproj#L43

```

  <Import Project="..\..\Switch\MonoGame.Effect.Compiler.targets" Condition="exists('..\..\Switch\MonoGame.Effect.Compiler.targets')" />
  <Import Project="..\..\XBoxOne\MonoGame.Effect.Compiler.targets" Condition="exists('..\..\XBoxOne\MonoGame.Effect.Compiler.targets')" />
  <Import Project="..\..\PlayStation4\MonoGame.Effect.Compiler.targets" Condition="exists('..\..\PlayStation4\MonoGame.Effect.Compiler.targets')" />
  <Import Project="..\..\PlayStation5\MonoGame.Effect.Compiler.targets" Condition="exists('..\..\PlayStation5\MonoGame.Effect.Compiler.targets')" />
  <Import Project="..\..\GDKX\MonoGame.Effect.Compiler.targets" Condition="exists('..\..\GDKX\MonoGame.Effect.Compiler.targets')" />
```

Which provides the hooks for console to inject their own support.

How would this work now?  Ideally it would still do this and we don't have to fix all the console platforms after this merge.

MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs
```diff
@@ -2,170 +2,194 @@
...
+            // If there were any additional output files we register
+            // them so that the cleanup process can manage them.
+            foreach (var outfile in shaderResult.AdditionalOutputFiles)
+                context.AddOutputFile(outfile);
+        }
+        catch (ShaderCompilerException)
+        {
+            // This will log any warnings and errors and throw.
+            ProcessErrorsAndWarnings(true, shaderErrorsAndWarnings, input, context);
+        }
```
tomspilman, line 87: Possibly good to have some `Exception` catch here just in case something else fails within the shader compiler?

Tests/Assets/Effects/DefinesTest.fx
```diff
@@ -2,7 +2,7 @@
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
-#include "include.fxh"
+#include "Include.fxh"
```
tomspilman, line 5: @SimonDarksideJ this tip would be good to include somewhere in the effect docs/tutorials.  Just a reminder that #include in your effect file is case sensitive if you're ever planning to build on Mac or Linux.

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.cs
```diff
@@ -78,6 +79,15 @@ protected static void ParseShaderModel(string text, Regex regex, out int major,
             minor = int.Parse(match.Groups["minor"].Value, NumberStyles.Integer, CultureInfo.InvariantCulture);
         }
 
+        public static ShaderProfile GetProfileForPlatform(TargetPlatform platform) => platform switch
+        {
+            TargetPlatform.Windows => ShaderProfile.DirectX_11,
+            TargetPlatform.iOS or TargetPlatform.Android or TargetPlatform.DesktopGL or TargetPlatform.MacOSX or TargetPlatform.RaspberryPi or TargetPlatform.Web => ShaderProfile.OpenGL,
+            TargetPlatform.DesktopVK => ShaderProfile.Vulkan,
+            TargetPlatform.WindowsGDK or TargetPlatform.XboxOne or TargetPlatform.XboxSeries => new DirectX12ShaderProfile(),
+            _ => ShaderProfile.FromName(platform.ToString())
```
tomspilman, line 88: Why not depend on `ShaderProfile.FromName(platform.ToString())` for all platforms like we have in the past?

Tools/MonoGame.Tools.Tests/EffectProcessorTests.cs
```diff
@@ -83,19 +83,19 @@ public void WriteError(string file, int line, int column, string message)
         [TestCase("Assets/Effects/ParserTest.fx")]
         public void TestParser(string effectFile)
         {
-            BuildEffect(effectFile, TargetPlatform.Windows);
+            BuildEffect(effectFile, TargetPlatform.DesktopGL);
```
tomspilman, line 86: Why the change to DesktopGL from Windows for these EffectProcessor tests?

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.cs
harry-cpp: We have never had a depndency on it for the public platforms, `FromName` does not expect a platform string, it expects a `ShaderProfile.Name`.

Tools/MonoGame.Tools.Tests/EffectProcessorTests.cs
harry-cpp: So test can be run on macOS / Linux as well. I thought about using Vulkan for it but decided that DesktopGL is more stable now.

Plus all 4 public `ShaderProfile` targets have been added bellow to test.

MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
harry-cpp: My bad, I forgot to include the effect.compiler targets into .csproj, I'll fix it in a moment.

MonoGame.Framework.Content.Pipeline/Processors/EffectProcessor.cs
harry-cpp: While restoring the old `EffectProcessor` code I did notice this, but technically its only useful to catch it if we want to process it ourselves, which we don't want to do for the generic exception here.

ie. above we are doing:

```cs
catch (InvalidContentException)
{
    throw;
}
catch (Exception ex)
{
    // TODO: Extract good line numbers from mgfx parser!
    throw new InvalidContentException(ex.Message, input.Identity, ex);
}
```

Tests/Assets/Effects/DefinesTest.fx
SimonDarksideJ: Task added to docs

MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
SimonDarksideJ: Was this resolved @harry-cpp ?

harry-cpp: Yes, in the latest commit.

SimonDarksideJ (approved):
Approved by committee

PR #8949, opened 2025-08-08
Content Builder Fixes

tomspilman:
A few fixes after using the new content builder.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
@@ -30,8 +30,6 @@ public abstract class ContentBuilder
     private readonly object _contentRequestLock = new();
     private readonly Dictionary<string, ContentInfo> _content = [];
     private readonly Dictionary<string, string> _outputContent = [];
-    private uint _succeededToBuild = 0;
-    private uint _failedToBuild = 0;
 
     /// <summary>
     /// Parameters to be used by the <see cref="ContentBuilder"/> or any of its subsystems.
@@ -60,12 +58,12 @@ public abstract class ContentBuilder
     /// <summary>
     /// Returns the number of content items that failed to build.
     /// </summary>
-    public uint FailedToBuild => _failedToBuild;
+    public uint FailedToBuild { get; private set; }
 
     /// <summary>
     /// Returns the number of content items that built successfully.
     /// </summary>
-    public uint SucceededToBuild => _succeededToBuild;
+    public uint SucceededToBuild { get; private set; }
 
     /// <summary>
     /// Initiates a build of the specified asset and then writes down the result to disk..
@@ -80,12 +78,12 @@ public abstract class ContentBuilder
         try
         {
             contentFileCache = ProcessContent(relativePath, contentInfo, true).contentFileCache;
-            _succeededToBuild++;
+            SucceededToBuild++;
         }
         catch (Exception ex)
         {
             Logger.Log(LogLevel.Error, $"Countent failed to build:\n{ex}");
-            _failedToBuild++;
+            FailedToBuild++;
         }
         Logger.PopFile();
         return contentFileCache;
@@ -103,13 +101,13 @@ public abstract class ContentBuilder
         try
         {
             var content = ProcessContent(relativePath, contentInfo, false);
-            _succeededToBuild++;
+            SucceededToBuild++;
             return content;
         }
         catch (Exception ex)
         {
             Logger.Log(LogLevel.Error, $"Countent failed to build:\n{ex}");
-            _failedToBuild++;
+            FailedToBuild++;
         }
         Logger.PopFile();
         return (null, null);
@@ -322,10 +320,10 @@ public abstract class ContentBuilder
         ContentCache.FlushCache(this);
 
         Logger.PushFile("Content Builder Finished");
-        Logger.Log($"{_succeededToBuild} succeeded, {_failedToBuild} failed");
+        Logger.Log($"{SucceededToBuild} succeeded, {FailedToBuild} failed");
         Logger.PopFile();
 
-        return _failedToBuild == 0;
+        return FailedToBuild == 0;
     }
 
     private void RunServer()
--- a/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
@@ -58,6 +58,16 @@ public abstract class ContentBuilder
     public abstract IContentCollection GetContentCollection();
 
     /// <summary>
+    /// Returns the number of content items that failed to build.
+    /// </summary>
+    public uint FailedToBuild => _failedToBuild;
+
+    /// <summary>
+    /// Returns the number of content items that built successfully.
+    /// </summary>
+    public uint SucceededToBuild => _succeededToBuild;
+
+    /// <summary>
     /// Initiates a build of the specified asset and then writes down the result to disk..
     /// </summary>
     /// <param name="relativePath">A relative path to the source asset.</param>
@@ -221,13 +231,13 @@ public abstract class ContentBuilder
     /// Runs the <see cref="ContentBuilder"/> with the specified parameters.
     /// </summary>
     /// <param name="parameters">A <see cref="ContentBuilderParams"/> describing both the platform paramteres for the content compilation as well as the configuration of the <see cref="ContentBuilder"/> itself.</param>
-    public void Run(ContentBuilderParams parameters)
+    public bool Run(ContentBuilderParams parameters)
     {
         Parameters = parameters;
         if (parameters.Mode == ContentBuilderMode.None)
         {
             // This means we are just showing the help menu.
-            return;
+            return false;
         }
 
         Directory.SetCurrentDirectory(Parameters.WorkingDirectory);
@@ -261,12 +271,13 @@ public abstract class ContentBuilder
         switch (Parameters.Mode)
         {
             case ContentBuilderMode.Builder:
-                RunBuild();
-                break;
+                return RunBuild();
             case ContentBuilderMode.Server:
                 RunServer();
                 break;
         }
+
+        return true;
     }
 
     /// <summary>
@@ -294,7 +305,7 @@ public abstract class ContentBuilder
         }
     }
 
-    private void RunBuild()
+    private bool RunBuild()
     {
         foreach (var pair in _content)
         {
@@ -313,6 +324,8 @@ public abstract class ContentBuilder
         Logger.PushFile("Content Builder Finished");
         Logger.Log($"{_succeededToBuild} succeeded, {_failedToBuild} failed");
         Logger.PopFile();
+
+        return _failedToBuild == 0;
     }
 
     private void RunServer()
--- a/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
@@ -148,83 +148,6 @@ public record ContentBuilderParams
     }
 
     /// <summary>
-    /// Parses out the main entry point args into a <see cref="ContentBuilderParams"/> to be used by <see cref="ContentBuilder"/>.
-    /// </summary>
-    /// <param name="args">Arguments passed to the main entry point of the app.</param>
-    /// <returns>
-    /// <see cref="ContentBuilderParams"/> containing the parsed arguments, or an empty <see cref="ContentBuilderParams"/> if no arguments were passed.
-    /// </returns>
-    public static ContentBuilderParams Parse(params string[] args)
-    {
-        var ret = new ContentBuilderParams();
-        var defaultValues = new ContentBuilderParams();
-        var rootCommand = new RootCommand("Content builder and conntent server for MonoGame.");
-        var rootOptions = new RootOptions(rootCommand);
-
-        var buildCommand = new Command("build", "Build all the content.");
-        var skipCleanOption = new Option<bool>(
-                name: "--skip-clean",
-                description: "Should the builder skip cleaning up old content cache data after the build is finished.",
-                getDefaultValue: () => defaultValues.SkipClean);
-        buildCommand.AddOption(skipCleanOption);
-        buildCommand.SetHandler(
-            (contentBuilder, skipCleanOption) => ret = contentBuilder with { Mode = ContentBuilderMode.Builder, SkipClean = skipCleanOption },
-            rootOptions,
-            skipCleanOption);
-        rootCommand.AddCommand(buildCommand);
-
-        var serverCommand = new Command("server", "Start a content server.");
-        var sererOptions = new ServerOptions(serverCommand);
-        serverCommand.SetHandler(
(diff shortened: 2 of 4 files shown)


MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
```diff
@@ -35,42 +35,42 @@ public RootOptions(RootCommand rootCommand)
...
-                description: "The output directory relative to the workingDir.",
+                description: "The output content directory.",
                 getDefaultValue: () => defaultValues.OutputDirectory);
             outputDirectoryOption.AddAlias("-o");
             rootCommand.AddGlobalOption(outputDirectoryOption);
 
             var intermediateDirectoryOption = new Option<string>(
                 name: "--intermediate",
-                description: "The intermediate directory relative to the workingDir.",
+                description: "The intermediate content directory.",
```
tomspilman, line 52: Maybe these should be more of a description of what they are for?

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
```diff
@@ -192,6 +196,48 @@ public static ContentBuilderParams Parse(params string[] args)
         return ret;
     }
 
+    private string MakeRooted(string path)
+    {
+        var result = path;
+        if (!Path.IsPathFullyQualified(result))
+        {
+            // Combine fails for paths wtih leading slashes.
+            while ( result.StartsWith(Path.DirectorySeparatorChar) ||
```
harry-cpp: result = result.TrimStart(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
```diff
@@ -192,6 +196,48 @@ public static ContentBuilderParams Parse(params string[] args)
...
+                result = result.Substring(1);
+
+            result = Path.Combine(WorkingDirectory, result);
+        }
+
+        result = Path.GetFullPath(result);
+        return result;
+    }
+
+    private static string MakeRelative(string root, string source)
```
harry-cpp: I think this entire method can be simplified like so:
```cs
return Path.IsPathRooted(source) ? Path.GetRelativePath(root, source) : source;
```

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
```diff
@@ -313,6 +324,8 @@ private void RunBuild()
         Logger.PushFile("Content Builder Finished");
         Logger.Log($"{_succeededToBuild} succeeded, {_failedToBuild} failed");
         Logger.PopFile();
+
+        return _failedToBuild == 0;
```
harry-cpp: Should we maybe just pipe out the _failedToBuild as an exit code?

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
```diff
@@ -192,6 +196,48 @@ public static ContentBuilderParams Parse(params string[] args)
         return ret;
     }
 
+    private string MakeRooted(string path)
```
harry-cpp: private methods of a class should go to the bottom next to public methods.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
```diff
@@ -57,6 +57,16 @@ private class ContentRequest
     /// <returns>An <see cref="IContentCollection"/> that contains information about the content handling.</returns>
     public abstract IContentCollection GetContentCollection();
 
+    /// <summary>
+    /// Returns the number of content items that failed to build.
+    /// </summary>
+    public uint FailedToBuild => _failedToBuild;
```
harry-cpp: Since you are already exposing this, better to remove private members and make the properties be `{ get; private set; }`

harry-cpp (changes requested):

Tools/MonoGame.Tools.Tests/BuilderTests.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            args = ContentBuilderParams.Parse(
+                "build",
+                "-s", Path.Combine(Directory.GetCurrentDirectory(), "../Some/Folder"),
+                "-o", Path.Combine(Directory.GetCurrentDirectory(), "\\Other/Folder"),
+                "-i", Path.Combine(Directory.GetCurrentDirectory(), "/Folder")
+            );
+            Assert.AreEqual(NormalizePath("../Some/Folder"), args.SourceDirectory);
+            Assert.AreEqual(NormalizePath("Other/Folder"), args.OutputDirectory);
+            Assert.AreEqual("Folder", args.IntermediateDirectory);
```
harry-cpp, line 58: This test is invalid,  `/` is a rooted path on Unix and `\\` is a rooted path on Windows representing shared drives so it should not try to trim the beginning of those paths.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
tomspilman: i considered that.  But figured it would be more flexible if we don't assume it was used as an exit code.

Instead i return false and let the user decide how they want to handle an exit code.

So in our starter kit case i would do:

```
var builder = new Builder();
builder.Run(args);
return builder.FailedToBuild > 0 ? -1 : 0; 
```

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
tomspilman: `IsPathRooted` is the wrong thing here.  `/Folder` is a "rooted" path on Windows.... so it wouldn't resolve it to a absolute path.  Thank the unit tests for showing me that.

Also the `source` may be absolute... but it could have flipped slashes or `..` in it.  Which is why i pass it thru GetFullPath as well.

harry-cpp: We need consistent behaviour between Windows and Unix, if the parh started with `/` or `\`, its a rooted path.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
tomspilman: This is the bottom below the last public method.

All you have below this is properties.

harry-cpp: Its bellow an internal class, not bellow a public method. Methods should go bellow properties.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
tomspilman: hum....  i could see people doing `builder build -s \Assets` and expecting it to resolve to the assets folder in the current folder.

Are you saying `\Assets` is a root folder on Unix/Mac?

harry-cpp: `/Assets` is a valid rooted folder on Unix.

`\\Assets` is a valid rooted networking path on Windows.

If the user passed an incorrect path, its not our job to try to fix it.

Tools/MonoGame.Tools.Tests/BuilderTests.cs
tomspilman: Actually this test is fine.

The input intermediate folder is `Path.Combine(Directory.GetCurrentDirectory(), "/Folder")` so it isn't `/Folder`.  And the correct resulting root relative path is `Folder`. 

So i think these tests are correct.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
tomspilman: I don't think we want to enforce Unix paths on Windows (or the opposite).

By using [Path.IsPathFullyQualified](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.ispathfullyqualified?view=net-9.0#system-io-path-ispathfullyqualified(system-string)) we ensure we're not breaking any absolute paths on whatever platform we're building on.

Tools/MonoGame.Tools.Tests/BuilderTests.cs
harry-cpp: `/Folder` is a valid rooted path on unix.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
tomspilman: Sorry it is not.  See here:

https://github.com/MonoGame/MonoGame/blob/cc34472b8a874dd6e02012749e0dae43d09c25bf/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs#L199

It is below the `ContentBuilderParams.Parse` method which is the last public method in the `ContentBuilderParams` class which it is in.

If you want it after all the properties i can do that.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
harry-cpp: Don't forget about this method as well.

tomspilman: What am i forgetting?

harry-cpp: To fix it, use the code snippet I posted above.

Right now when you pass a relative path, you make it absolute, and make it relative again, which is silly.

tomspilman: Here is the code in question:

```
        // First be sure we have an absolute unambigous path first.
        if (!Path.IsPathFullyQualified(result))
            result = Path.Combine(Directory.GetCurrentDirectory(), result);

        // Now make it relative to the incoming root.
        //
        // Note this may still return an absolute path in the case
        // that these directories are on different drives.
        //
        result = Path.GetRelativePath(root, result);
```

I did it like this because there is no guarantee that `Directory.GetCurrentDirectory()` and `root` are the same path... they could be totally different drives.

What i could have wrong is the intent of the `--workingDir` argument.  

I see `--workingDir` as the root to convert all paths to during content building.

I see CWD as the root for any relative paths passed to `-s -i -o` arguments on the command line. 

Which is why i make it absolute to one then relative to the other.

harry-cpp: Its fine if you need that to fix some absolute path case, but then just add the:
```cs
if(!Path.IsPathRooted(source))
    return source;
```

To the top of the function.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
harry-cpp: Honestly, that entire Parse function should be at the bottom as well :D

I think it was originally a constructor and after I converted it to a Parse method I forgot to move it.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
harry-cpp: > What i could have wrong is the intent of the --workingDir argument.

Its meant to override the working dir from the CLI, if you have:
```
--workingDir D:\ContentFolder --src Assets
```

Then rooted src dir should be `D:\ContentFolder\Assets`

tomspilman: Ok... that doesn't make sense to me, but will change it.

Tools/MonoGame.Tools.Tests/BuilderTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            Assert.AreEqual(ContentBuilderMode.Builder, args.Mode);
+            Assert.AreEqual(TargetPlatform.DesktopGL, args.Platform);
+            Assert.AreEqual(false, args.CompressContent);
+            Assert.AreEqual(false, args.SkipClean);
+            Assert.IsTrue(Path.IsPathRooted(args.WorkingDirectory));
+            Assert.AreEqual(Directory.GetCurrentDirectory(), args.WorkingDirectory);
+            Assert.AreEqual("Content", args.SourceDirectory);
+            Assert.AreEqual(NormalizePath(Directory.GetCurrentDirectory(), "Content"), args.RootedSourceDirectory);
+            Assert.AreEqual(NormalizePath("bin/Content"), args.OutputDirectory);
+            Assert.AreEqual(NormalizePath(Directory.GetCurrentDirectory(), "bin\\Content"), args.RootedOutputDirectory);
```
tomspilman: Hey @harry-cpp .

Any ideas why this line works on Windows but fails on Mac and Linux:
```
  Failed CommandLineParserTests [153 ms]
  Error Message:
     String lengths are both 78. Strings differ at index 70.
  Expected: "...k/MonoGame/MonoGame/Artifacts/Tests/Tools/Release/bin\\Content"
  But was:  "...k/MonoGame/MonoGame/Artifacts/Tests/Tools/Release/bin/Content"
```
Is the normalize code doing `path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)` not actually reversing the slashes?

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilderParams.cs
```diff
@@ -272,4 +227,71 @@ public static ContentBuilderParams Parse(params string[] args)
...
+        if (!Path.IsPathRooted(path))
+            path = Path.Combine(WorkingDirectory, path);
+
+        return Path.GetFullPath(path);
+    }
+
+    private static string MakeRelative(string workingDir, string path)
+    {
+        if (!Path.IsPathRooted(path))
+            return path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
```
tomspilman: Seems i needed this else Unix style paths could be passed thru the pipeline on Windows.

Tools/MonoGame.Tools.Tests/BuilderTests.cs
harry-cpp: It's not, `\` is not an alt directory separator on Unix, unlike Windows passing `\` in a path will just result in a crash on Unix, the only exception being .csproj files as they have compatibility mode.

MonoGame.Framework.Content.Pipeline/PipelineBuilder/FileHelper.cs
```diff
@@ -6,8 +6,39 @@
...
+
+        /// <summary>
+        /// The required or default directory seperator for this platform.
+        /// </summary>
+        public static readonly char Separator = Path.DirectorySeparatorChar;
+
+        /// <summary>
+        /// Fixes the incoming path to have the default directory seperators for the current platform.
+        /// </summary>
+        public static string NormalizeDirectorySeparators(string path)
```
tomspilman, line 37: This is basically a copy of what we had implemented as an internal in Framework.FileHelper.

harry-cpp (approved):

Issue #8923, opened 2025-07-23
Proposal: resolving culture-issues in value type serialization.


Lucasharskamp:
**Issue**

When one serializes a number it gets serialized based on a culture. If no culture is provided, .NET applies the `CultureInfo.CurrentCulture` instead. Which means you get different behavior based on the culture of the machine;

Example:
```
var number = 123456.78;
Console.WriteLine(number.ToString()); // Uses CultureInfo.CurrentCulture
// Output: 123,456.78 (en-US)
// Output: 123.456,78 (de-DE, nl-NL, etc.)
```
(Note: it's not just decimal numbers that are displayed differently based on culture)

In a lot of value types defined in MonoGame, `ToString()` methods  or `DebugDisplayString` calls simply apply `ToString()` on their internal members without considering the culture or format provider, which can cause different behaviors on different machines, without any way for the end programmer to specify which one they specifically want to utilize. In a previous PR (for the .editorconfig) I've set CA1304/CA1305 (the corresponding issues) as warnings in the build so we can trace where such issues can rear its head.

This serialization issue applies to the following value types:
- Alpha8
- BoundingSphere 
- Byte4
- Color
- HalfSingle
- Matrix
- NormalizedByte2
- NormalizedByte4
- NormalizedShort2
- NormalizedShort4
- Point
- Quaternion
- Ray
- Short2
- Short4
- Vector2
- Vector3
- Vector4

**Proposed solution**

We have to consider regression; ''rectifying this issue'' may cause problems for existing projects where the original behavior was anticipated. Therefore:

- Apply hard-coded `CultureInfo.InvariantCulture` when serializing for `DebugDisplayString` or other non-UI related matters, since that is about technical matters anyway where you'd want the exact same value between any device.
- Provide an additional `ToString(IFormatProvider provider)` method for value types where the programmer can provide their own preferred setting. The original `ToString()` will utilize `CultureInfo.CurrentCulture`, to ensure the behavior remains the same for existing projects. (And also because it is more intuitive, given how Microsoft does it this way as well)

PR #8950, opened 2025-08-08
Specifying culture to use when serializing value types in MonoGame

Lucasharskamp:
This PR contains a few solutions to (potential) issues with serializing value types, as discussed in detail in https://github.com/MonoGame/MonoGame/issues/8923

1) ValueTypes (like `Quaternion`) now have an overload to their ToString() methods where one can provide their desired `IFormatProvider`
2) All existing usages of ToString, String.Format or other features that may have different behavior based on their culture are now specified as using `CultureInfo.InvariantCulture`, to ensure we get the same behavior regardless of whatever unknown culture rules are in use on someone's machine. 
3) StringBuilder methods where a single char was appended are now done as adding a char rather than a string of 1 char in length.

### DIFF
--- a/MonoGame.Framework/BoundingBox.cs
+++ b/MonoGame.Framework/BoundingBox.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.Runtime.Serialization;
 
 namespace Microsoft.Xna.Framework
@@ -877,7 +878,17 @@ namespace Microsoft.Xna.Framework
         /// <returns>A <see cref="String"/> representation of this <see cref="BoundingBox"/>.</returns>
         public override string ToString()
         {
-            return "{{Min:" + this.Min.ToString() + " Max:" + this.Max.ToString() + "}}";
+            return "{{Min:" + this.Min.ToString(CultureInfo.CurrentCulture) + " Max:" + this.Max.ToString(CultureInfo.CurrentCulture) + "}}";
+        }
+
+        /// <summary>
+        /// Serializes the <see cref="BoundingBox"/> to a string using the specified format provider.
+        /// </summary>
+        /// <param name="formatProvider">Format provider to utilise</param>
+        /// <returns>String representation of this <see cref="BoundingBox"/></returns>
+        public string ToString(IFormatProvider formatProvider)
+        {
+            return "{{Min:" + this.Min.ToString(formatProvider) + " Max:" + this.Max.ToString(formatProvider) + "}}";
         }
 
         /// <summary>
--- a/MonoGame.Framework/BoundingSphere.cs
+++ b/MonoGame.Framework/BoundingSphere.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.Runtime.Serialization;
 
 namespace Microsoft.Xna.Framework
@@ -40,7 +41,7 @@ namespace Microsoft.Xna.Framework
             {
                 return string.Concat(
                     "Center( ", this.Center.DebugDisplayString, " )  \r\n",
-                    "Radius( ", this.Radius.ToString(), " )"
+                    "Radius( ", this.Radius.ToString(CultureInfo.InvariantCulture), " )"
                     );
             }
         }
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -6,6 +6,7 @@ using System;
 using System.Text;
 using System.Runtime.Serialization;
 using System.Diagnostics;
+using System.Globalization;
 
 namespace Microsoft.Xna.Framework
 {
@@ -1842,10 +1843,10 @@ namespace Microsoft.Xna.Framework
             get
             {
                 return string.Concat(
-                    this.R.ToString(), "  ",
-                    this.G.ToString(), "  ",
-                    this.B.ToString(), "  ",
-                    this.A.ToString()
+                    this.R.ToString(CultureInfo.InvariantCulture), "  ",
+                    this.G.ToString(CultureInfo.InvariantCulture), "  ",
+                    this.B.ToString(CultureInfo.InvariantCulture), "  ",
+                    this.A.ToString(CultureInfo.InvariantCulture)
                 );
             }
         }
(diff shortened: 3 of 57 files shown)


ThomasFOG:
I'll continue reviewing this later, but it's a bit hard to read because there are changes of different kind which aren't useful/not desirable. It feels like you didn't try to target specific parts and just batch replaced things without checking if it's useful or breaking.

Before trying to fix any of that, we need to make sure that #8923 _is_ an issue with reproducible use cases.

Lucasharskamp:
> I'll continue reviewing this later, but it's a bit hard to read because there are changes of different kind which aren't useful/not desirable. It feels like you didn't try to target specific parts and just batch replaced things without checking if it's useful or breaking.
> 
> Before trying to fix any of that, we need to make sure that #8923 _is_ an issue with reproducible use cases.

I am having problems with that, hence why I made the PR. In my case, it's about logging diagnostic data for debugging purposes.
Here's an example:

```
CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
var foo = new Vector2(100.4f, 50.5f);
Console.WriteLine(foo.ToString()); // { X:100.4, Y:50.5 }
Console.Writeline(foo.DebugDisplayString); // 100.4 50.5
CultureInfo.CurrentCulture = new CultureInfo("de-DE"); // German. Also applies to Dutch, French, etc.
Console.WriteLine(foo.ToString()); // { X:100,4, Y:50,5 }
Console.WriteLine(foo.DebugDisplayString); // 100,5  50,2
```

Of course there are workarounds (serializing fields by hand or using stuff like Newtonsoft.Json or System.Text.Json that does this for you), but properties like `DebugDisplayString` should simply be Culture agnostic out-of-the-box, and ToString should allow the programmer to specify their desired Culture to utilise (as .NET does for most default value types)

In the other examples it's about ensuring bugs related to the machine the code runs on cannot pop up; I've learned that one shouldn't make assumptions about the culture of the machine .NET code runs on, and presume it's something exotic, which can cause different behavior to pop up nobody anticipated.  So these are more ''just to be safe''.

SimonDarksideJ:
Any progress @Lucasharskamp on defining a FAILING Unit test for https://github.com/MonoGame/MonoGame/issues/8923

Before we start on a PR to fix the issues raised by the Unit test.  We cannot fix something unless we have a testable identification of the problem.

Lucasharskamp:
``

> Any progress @Lucasharskamp on defining a FAILING Unit test for #8923
> 
> Before we start on a PR to fix the issues raised by the Unit test. We cannot fix something unless we have a testable identification of the problem.

I'm creating a UnitTest that does the following for every Value type:
- Set different distinct CultureInfo's, serialize that way, see if it serializes as expected (string comparison - using en-US, en-GB, fr-FR, de-DE, nl-NL, it-IT,  ja-JP, zh-Hans)
- Set CultureInfo explicitly when serializing
- Ensure DebugDisplayString is always equal regardless of machine/set culture.
- For ''failure'': serialize a value type 2 times (in different set cultures) and it fails because the output is not equal. 

**Edit**: It's a lot of items to go through, I'll have it finished before the 22nd.

SimonDarksideJ:
Hay @Lucasharskamp , just checking in on how the tests were going.  Would be great to get this included in the 3.8.5 timelines, but no immediate rush

ThomasFOG:
Could you remove culture changes in places where it has nothing to do with printing data?

Lucasharskamp:
> Hay @Lucasharskamp , just checking in on how the tests were going. Would be great to get this included in the 3.8.5 timelines, but no immediate rush

Done. I've only added a few test - I could add one for each type, but that would make it a "repeated method" about two dozen times over, only changing the type to alter, without adding anything new. If preferred, I could of course add those anyway.


> Could you remove culture changes in places where it has nothing to do with printing data?

Convert.ChangeType can lead to unexpected issues depending on the Culture present, as it under the hood utilises _IConvertible_.  A few generic examples of this happening are included in the new unittests (method SerializationTest.cs -> TestChangeType). I doubt existing MonoGame implementations could lead to such issues occurring, but I wouldn't be surprised either if they did; it's a "better be safe than sorry" for edge cases with rare cultures.  
See https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Convert.cs,3bcca7a9bda4114e

The other places where changes were made have similar issues.

MonogameGuy:
The purposed change could break times, dates, fonts, and monetary units. 

If you want a `ToString()`  formatted a certain way, pass the Culture as intended in your game. The base `ToString()` was not meant for end user consumption. It is primarily a debugging aid, but when used in a game, it is not expected to vary based on culture, without the programmer asking it to.

The `Date` format alone shows the danger of mixing formats in strings or constants. What is the day after 11/11/xx: 12/11/xx or 11/12/xx? 

For my games, It's about making the runtime behavior deterministic across all locales. 
Otherwise it is a debugging nightmare. 

Please add a Culture `override` method where not provided in the framework:
```
public override string ToString()
        {
            return X.ToString();
        }
 public override string ToString(CultureInfo cultureInfo)
        {
            return X.ToString("X", cultureInfo);
        }
```
Adding a culture‑aware `ToString(CultureInfo cultureInfo)` is absolutely the right thing to do when you care about determinism and locale independence. The pattern above gives you both safety (invariant output) and flexibility (culture‑specific overloads).

What may seem like an innocuous simple change can be quite harmful to programs dependent upon the existing behavior.

AristurtleDev:
@Lucasharskamp we would need you to update your initial PR description to include the new statement regarding our LLM policy as shown in the updated PR templates here: https://github.com/MonoGame/MonoGame/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

```
### Contributor Declaration

```

AristurtleDev:
One note about what @MonogameGuy said here

> If you want a `ToString()` formatted a certain way, pass the Culture as intended in your game. The base `ToString()` was not meant for end user consumption. It is primarily a debugging aid, but when used in a game, it is not expected to vary based on culture, without the programmer asking it to.

This is technically correct, the base `Object.ToString` was not meant to be culture aware. Official .NET guidelines state that if it needs to be culture sensitive, then we should be using the `IFormattable` interface on the types.  The API is similar to what was added for the various types, only it has an additional `string` param that goes with it.  See the remarks and examples here 

https://learn.microsoft.com/en-us/dotnet/api/system.iformattable.tostring?view=net-10.0#definition

SimonDarksideJ:
A few conflicts @Lucasharskamp , can you resolve for consideration in `3.8.6` and add the LLM statement please.

PR #8953, opened 2025-08-13
Add CppNet as a submodule and drop the Dependencies submodule

harry-cpp:
Stuff done:
- Add CppNet as submodule
- Remove Dependencies submodule (finally)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "ThirdParty/Dependencies"]
-	path = ThirdParty/Dependencies
-	url = https://github.com/MonoGame/MonoGame.Dependencies.git
 [submodule "ThirdParty/SDL_GameControllerDB"]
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,6 +1,3 @@
-[submodule "ThirdParty/Dependencies"]
-	path = ThirdParty/Dependencies
-	url = https://github.com/MonoGame/MonoGame.Dependencies.git
```
tomspilman, line 3: Are we sure nothing is used here anymore?  Even on mobile?

I thought we use openal-soft for Android from there.

dellis1972, line 3: 🥳 🎉

harry-cpp: Nop, `MonoGame.Library.OpenAL` nuget for them.

tomspilman (approved):

SimonDarksideJ (approved):

Issue #8944, opened 2025-08-04
[DesktopVK] Road to release


ThomasFOG:
Here are the remaining DesktopVK issues that need to be fixed before we consider publishing it publicly.

**General maintenance**

- [ ] #9001 
- [x] #9007
- [ ] #8934

**Vulkan related issues**
- [x] #9000 🔥 
- [x] ~~Lots of flickers (possibly related to using RenderTargets or switching Effects) which seem weirdly related to the variation of draw calls: if the number of draw calls is stable, there will be no flicker, but if draw calls are varying frame to frame, flickers are introduced and the more variations there are, the more the flickers~~ (fixed in #8999)
- [x] ~~Changing resolution while in exclusive fullscreen (```HardwareModeSwitch``` set to true) doesn't refresh the viewport/bounds correctly,  and exiting exclusive mode sets the window size to 800x600 (it seems that SDL reports a resize event of 800x600 when exiting fullscreen, and this seems wrong)~~ (fixed in #8999)
- [x] ~~Exclusive fullscreen seems to ignore the requested backbuffer size and seems to always be pretty small~~  (fixed in #8999)
- [x] ~~Vsync isn't respected and is always off, no matter the fullscreen mode or windowed~~  (fixed in #8999)
- [x] ~~```GraphicsDevice_Present()``` crashes randomly when resizing the window or changing fullscreen/windowed mode (which means that it can happen on app launch too)~~ (fixed in #8999)
- [x] ~~```GraphicsDevice_Destroy()``` crashes randomly when closing apps (may be the same cause as the previous point)~~ seems to be related to #9002 
- [ ] #9002 
- [ ] #9003
- [x] ~~Minimizing the window crashes (because it tries to set a 0x0 backbuffer)~~ (fixed in #8999)
- [ ] #9006
- [ ] #9005 🔥 
- [ ] #9004
- [x] ~~We need to verify which version of SPIR-V DXC generates to make sure that it is as low as possible (the Vulkan implementation is 1.0 compliant, it would be best to keep the SPIR-V version to 1.0 too for maximum compatibility)~~

**Other issues**

- [ ] Native crashes don't report anything beside an empty ```System.ExecutionEngineException```, can we do better in that regard? (non-critical, but if native crashes happen, users will likely be very disoriented)

**Doing the release**

- [ ] #8761
- [ ] #8521
- [ ] #8760

tomspilman:
>  I've run into issues which are not covered by the tests but may be considered as critical

[...]

PR #8957, opened 2025-08-17
Fix Vulkan descriptor cache

ThomasFOG:
This PR fixes the flickering/corrupt frames of the Vulkan backend (see #8944).

@tomspilman Note that I have increased the retention of unused resources to 3 frames (because 2 was not working for descriptors due to the uniforms buffer flip/flop).

### DIFF
--- a/native/monogame/vulkan/MGG_Vulkan.cpp
+++ b/native/monogame/vulkan/MGG_Vulkan.cpp
@@ -40,6 +40,47 @@
 	}																					\
 }
 
+#if defined(DEBUG)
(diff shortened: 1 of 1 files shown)


ThomasFOG:
Okay, this will conflict with #8903 because I worked from that branch and I've added ```VK_SET_OBJECT_NAME``` to avoid missing adding it back post-merge.

We should likely wait for #8903 to be merged, and I'll resolve the merge conflicts here.

ThomasFOG:
Replaced by #8999

PR #8961, opened 2025-08-21
Content Builder Fixes

tomspilman:
This has a few fixes to the new content builder process.


MonoGame.Framework.Content.Pipeline/Processors/MaterialProcessor.cs
```diff
@@ -111,7 +111,7 @@ public MaterialProcessor()
         /// <remarks>If the input to process is of type EffectMaterialContent, this function will be called to request that the EffectContent be built. The EffectProcessor is used to process the EffectContent. Subclasses of MaterialProcessor can override this function to modify the parameters used to build EffectContent. For example, a different version of this function could request a different processor for the EffectContent.</remarks>
         protected virtual ExternalReference<CompiledEffectContent> BuildEffect(ExternalReference<EffectContent> effect, ContentProcessorContext context)
         {
-            return context.BuildAsset<EffectContent, CompiledEffectContent>(effect, "EffectProcessor");
+            return context.BuildAsset<EffectContent, CompiledEffectContent>(effect, new EffectImporter(), new EffectProcessor());
```
tomspilman, line 114: @harry-cpp Since the `EffectContent` here is already loaded... why do i need to pass an importer?

The implementation seems to use the importer in caching...  but it doesn't import anything.

Is it safe to pass null?

harry-cpp: I'll look into modifying it so null importer becomes valid.

PR #8974, opened 2025-08-28
Re-enable instancing tests for OpenGL

SimonDarksideJ:
# Summary

Validating Graphics Instancing tests


ThomasFOG:
I believe you got it backward and disabled the test for WindowsDX.

SimonDarksideJ:
You're right, lol

ThomasFOG:
Looking good it seems?

tomspilman:
I cannot tell if the test ran.

It seems the instancing test is run under:

image

...but that doesn't run under Mac or Linux.

Someone needs to dig in further here to be sure this test is running... i don't think it is.

tomspilman:
Ok yea... it did not run.

`DrawInstancedPrimitivesVisualTest` is the test.  You can see it in the raw log of `tests-windows-selfhosted-latest`.  It does not appear in the MacOS and Linux logs.

SimonDarksideJ:
Investigating...   This is one of the reasons @harry-cpp and I want to migrate all the tests to Cake builds, will be easier to identify these things.  In fact it is damn hard to tell what runs what tests atm.

I'm on it.

SimonDarksideJ:
The reason it didn't run was because the YML entry was "COMMENTED OUT" because it didn't run on MacOS well.
SO instead it has NEVER been run again, even on Windows?  ALL tests in the "GraphicsDeviceTest" test class :S

SimonDarksideJ:
Thoughts @tomspilman / @dellis1972 , slightly concerned that the tests are enabled but are passing?  Although we know there is an issue.

dellis1972:
@SimonDarksideJ the test is passing because its not using the API which is broken. 
There is an overload of `DrawInstancedPrimitives` which takes an additional offset argument, that is the one which is broken.

dellis1972:
@SimonDarksideJ also that offset API doesn't work on OpenGL on mac, it never will due to limitations in the API. It requires OpenGL 4.2 and only OpenGL 4.1 is available on mac.

ThomasFOG:
And we shouldn't upgrade OpenGL just for this overload to work, we would loose a great deal of compatibility over a marginally used feature. Let's just throw a ```NotImplementedException``` (and disable its test, providing we add a test for it).

dellis1972:
> And we shouldn't upgrade OpenGL just for this overload to work, we would loose a great deal of compatibility over a marginally used feature. Let's just throw a `NotImplementedException` (and disable its test, providing we add a test for it).

We should also add the `[SupportedOSPlatform("Windows")]` attribute to the method if it only works on windows

tomspilman:
@SimonDarksideJ what is the status of this PR?

SimonDarksideJ:
We began by ensuring the platform now correctly runs tests in the area, however, it was determined that instancing was never implemented for OpenGL

dellis1972:
@tomspilman  not there was only one specific API which did not work on MacOS its the Instanced Draw which has an offset due to the version of OpenGL on mac (4.1) 
I seemed to remember you mentioning we might be able to calculate the offset ourselves on Mac?
Personally, if the tests actually pass we should merge this.

SimonDarksideJ:
My concern is @dellis1972 , that while the tests are passing, this is only because:

* it isn't actually testing
* the implementations on opengl are not actually implemented and are skipped

But of which matches the experience where instancing is working on dx but fails on opengl

AristurtleDev:
@SimonDarksideJ looks like there are conflicts that need to be resolved, and missing the LLM statement in the PR description

SimonDarksideJ:
Closing as this is not required for legacy OpenGL as it was never implemented, instead, ensuring it is active and tested in OpenGL4/Web ([9491](https://github.com/MonoGame/MonoGame/pull/9491))

PR #8975, opened 2025-08-28
Add PowerStatus API. Useful for laptops and portable gaming devices.

CartBlanche:
Feature: fleshing out unimplemented API.

### Description of Change

Adding the various back-ends to support Battery Status on our supported platforms.

### Contributor Declaration


MonoGame.Framework/Devices/Power/PowerStatus.cs
```diff
@@ -0,0 +1,11 @@
+
+
+namespace MonoGame.Framework.Devices.Power
+{
+    public sealed partial class PowerStatus
+    {
+        public string BatteryChargeStatus => PlatformBatteryChargeStatus;
+        public string PowerLineStatus => PlatformPowerLineStatus;
```
harry-cpp: These should return enum values and not string.

harry-cpp (changes requested):

MonoGame.Framework/Devices/Power/PowerStatus.cs
CartBlanche: Sure. Thanks for looking over the PR.

dellis1972:
Looks like the MonoGame.Framework.Native Side of this is missing

sepcnt:
> Looks like the MonoGame.Framework.Native Side of this is missing

It's not hard to implement these API. But have to get PR #8903 merged first.

CartBlanche:
@dellis1972 I've not generally touched the Native side as @tomspilman usually handled that. Happy to add it if he points me in the right direction. Thing is this is non breaking addition as most people haven't used this API since Windows Phone, which is where I believe it was introduced. So in theory we could add it in stages, as Native is not yet widely used. But again I can add it in here, if pointed in the right direction.

MonoGame.Framework/Platform/Windows/Devices/Power/PowerStatus.cs
```diff
@@ -0,0 +1,35 @@
+using System.Windows.Forms;
+using MonoGame.Framework.Devices.Power;
+
+public partial class PowerStatus
+{
+    private BatteryChargeStatus PlatformBatteryChargeStatus()
+    {
+        return SystemInformation.PowerStatus.BatteryChargeStatus switch
+        {
+            BatteryChargeStatus.Charging => BatteryChargeStatus.Charging,
```
AristurtleDev: I see the switch expression is typed from `SystemInformation.PowerStatus.BatteryChargeStatus`, which should disambiguate the enum. Do we still want to fully qualify the enum members here (e.g., `System.Windows.Forms.BatteryChargeStatus.*`) for readability and to avoid future confusion with `MonoGame.Framework.Devices.Power.BatteryChargeStatus`, or is that considered unnecessary noise?

CartBlanche: @AristurtleDev I think it is a good suggestion. Done.

CartBlanche:
@dellis1972 I personally don't think it is worth waiting until the full native is implemented. I still think it is worth merging while we wait for native support. At least devs have something they can use now, while the native is incoming. Once merged I'll raise an issue for the native side of things, so we don't lose track of it.

dellis1972:
So, we need unit tests to make sure this API is functioning as expected. 
I also think we do need a implementation for native, we use SDL2 for both Vulkan and DX12 so it should be similar to the current DesktopGL one that has been implemented.

Tests/Framework/Devices/Power/PowerStatusTest.cs
```diff
@@ -0,0 +1,147 @@
...
+				StubPowerInfo(testCase.State, 50);
+
+				var powerStatus = new PowerStatus();
+
+				Assert.AreEqual(testCase.Expected, powerStatus.BatteryChargeStatus);
+			}
+		}
+
+		[Test]
+		public void PowerLineStatus_MapsDesktopPowerState()
```
dellis1972: Can we use `TestCase` on these also. the cases in the test look weird

PR #8983, opened 2025-08-30
Made TPGParser Private

tomspilman:
The TPGParser was part of the public API when it shouldn't be.

Also added copyright notices to the code.


harry-cpp (dismissed):

Tools/MonoGame.Effect.Compiler/Effect/TPGParser/PassInfo.cs
```diff
@@ -1,9 +1,13 @@
...
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System;
 using System.Text.RegularExpressions;
 using Microsoft.Xna.Framework.Graphics;
 using System.Globalization;
 
-namespace MonoGame.Effect.TPGParser
+namespace MonoGame.Effect
```
tomspilman, line 10: @harry-cpp i moved these directly to the `Effect` namespace since they need to be public for `ShaderProfile` to work.

tomspilman: Should i move them out one folder too?

harry-cpp: It's an internal method, we just need to mark all the classes as internal instead of public and it should work fine without expositing the methods no?

dellis1972 (approved):

Tools/MonoGame.Effect.Compiler/Effect/TPGParser/PassInfo.cs
SimonDarksideJ: > It's an internal method,

I agree these classes should all be marked `internal`, personally, I see it as bad practice to have no accessor at all on classes/methods.  It should ALWAYS state clearly what its access pattern is, not left to assumptions.

SO if it is meant to be private, then state it as private to avoid confusion.

SimonDarksideJ (approved):
The rest of the methods without accessors "should" also be private. But approved

tomspilman:
@SimonDarksideJ i was planning to fix this to make things all internal.  I'll submit a new PR later.

PR #8988, opened 2025-09-01
[Pipeline] Refactor the content builder

harry-cpp:
- ContentCollection: Add ability for specifying multiple outputs for the same asset input
- ContentCollection: Make include / exclude only do operations on current content root
- ContentCollection: Modify outputPath function to expect a path without extension in case of build action
- Parameters: Make Parameters a class so they can both be read from CLI and afterwards modified in code more easily
- Parameters: Add ability to rebuild content
- Cache: Make content cache track content based on produced outputs instead of inputs
- Cache: Mark ContentFileCache as internal and make an IContentFileCache that the IContentCache is working with
- Cache: Track output file times in case of external modifications
- Cache: Sanitize filepaths
- Builder: Add ability for multiple dependencies of the same asset to have separate properties (more refactoring on this is coming so that we don't waste space)
- Builder: Sanitize passed pathing to ContentCollection
- Builder: Automatically append .xnb to output path in case of build action
- Builder: Remove GetDefaultOutputPath and GetDefaultCopyPath as they now no longer do any modifications

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/ContentCollectionRoot.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/ContentCollectionRoot.cs
@@ -1,0 +1,145 @@
+// MonoGame - Copyright (C) MonoGame Foundation, Inc
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+ 
+using Microsoft.Xna.Framework.Content.Pipeline;
+
+namespace MonoGame.Framework.Content.Pipeline.Builder;
+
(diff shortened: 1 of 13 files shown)


MonoGame.Framework.Content.Pipeline/Builder/ContentCollectionRoot.cs
```diff
@@ -0,0 +1,142 @@
+
```
tomspilman: Need the copyright info at the top of all these new code files.

SimonDarksideJ (approved):

PR #8995, opened 2025-09-05
Fix texture importer error message

tomspilman:
When the source texture is not found we need to return an appropriate error message.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
+++ b/MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
@@ -164,13 +164,13 @@ public abstract class ContentBuilder
 
         if (!ContentBuilderHelper.GetImporter(relativePath, contentInfo.Importer, out IContentImporter importer))
         {
-            Logger.Log(LogLevel.Warning, "Importer: Not found :(");
+            Logger.Log(LogLevel.Warning, "Importer: Not found");
             return null;
         }
         Logger.Log($"Imposter: {importer.GetType().Name}");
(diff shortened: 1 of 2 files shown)


tomspilman:
I was getting this error message before this fix:

```
2>    00:06:45.889 [E]     Microsoft.Xna.Framework.Content.Pipeline.InvalidContentException: TextureImporter failed to load 'G:/Projects/Ascent/Environment/Modular Tunnels/Long Straight/Textures/GrayGridCeilingFloor.png': no SOI
2>    00:06:45.889 [E]
2>    00:06:45.889 [E]        at Microsoft.Xna.Framework.Content.Pipeline.TextureImporter.Import(String filename, ContentImporterContext context) in D:\Projects\MonoGame\MonoGame.Framework.Content.Pipeline\TextureImporter.cs:line 84
```

The `no SOI` was coming from STB JPEG which had nothing to do with the actual error in this case.

MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
```diff
@@ -164,13 +164,13 @@ public void BuildAndWriteContent(string relativeSrcPath, ContentInfo contentInfo
 
         if (!ContentBuilderHelper.GetImporter(relativePath, contentInfo.Importer, out IContentImporter importer))
         {
-            Logger.Log(LogLevel.Warning, "Importer: Not found :(");
+            Logger.Log(LogLevel.Warning, "Importer: Not found");
```
harry-cpp, line 167: :(

harry-cpp (approved):

PR #8996, opened 2025-09-05
[Pipeline] Fix builder not detecting failed builds if a dependency fails

harry-cpp:


MonoGame.Framework.Content.Pipeline/Builder/ContentBuilder.cs
```diff
@@ -82,10 +84,20 @@ public void BuildAndWriteContent(string relativeSrcPath, ContentInfo contentInfo
         }
         catch (Exception ex)
         {
-            Logger.Log(LogLevel.Error, $"Countent failed to build:\n{ex}");
             FailedToBuild++;
```
tomspilman, line 87: Will in be weird if you had lets say 1 FBX in your content...  but it returned that 2 failed to build because a asset dependency failed?
```
Logger.Log($"{SucceededToBuild} succeeded, {FailedToBuild} failed");
```
Maybe the failure increment should only happen for the top level asset?

tomspilman: Maybe `SucceededToBuild` should also only increment when there is no parent context?

harry-cpp: From the beginning the builder tracked them together as technically multiple assets are being built, not sure what would be the most clear here, but that is a discussion for a separate PR.

tomspilman (approved):

PR #8997, opened 2025-09-06
Fix backslashes in content builder paths

tomspilman:
This PR corrects incoming backslashes to forward slashes on any paths passed in by the developer from a public API.

This is important because most Windows devs will use backslashes `\` because that is what they are used to.  This however is wrong for cross platform use.  While we recommend people use forward slashes, people will either not notice or not care.  

In particular the wildcard input of `**\*.png`  won't work on any platform but `**/*.png` will.

These changes ensures things "just work" and reduces support requests.

Also fixed a spelling mistake.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
+++ b/MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
@@ -175,7 +175,7 @@ public class ContentBuildLogger
     public virtual void PushFile(string filename)
     {
         filename = filename.Sanitize();
-        Log(filename.Sanitize());
+        Log(filename);
         _filenames.Push(filename);
     }
 
--- a/MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
+++ b/MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System.Diagnostics;
+using MonoGame.Framework.Content.Pipeline.Builder;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline;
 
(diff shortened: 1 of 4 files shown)


MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
```diff
@@ -173,7 +174,8 @@ public virtual void LogWarning(string helpLink, ContentIdentity contentIdentity,
     /// <param name="filename">Name of the file containing future messages.</param>
     public virtual void PushFile(string filename)
     {
-        Log(filename);
+        filename = filename.Sanitize();
+        Log(filename.Sanitize());
```
harry-cpp: This will have an effect on mgcb.

Also no need for double Sanitize.

tomspilman: > This will have an effect on mgcb.

@harry-cpp 

What sort of effect?

MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
```diff
@@ -173,6 +174,7 @@ public virtual void LogWarning(string helpLink, ContentIdentity contentIdentity,
     /// <param name="filename">Name of the file containing future messages.</param>
     public virtual void PushFile(string filename)
     {
+        filename = filename.Sanitize();
```
tomspilman, line 177: @harry-cpp so what is the effect on MGCB you mention?

harry-cpp: ContentBuildLogger.cs is used by both mgcb and the new builder.

tomspilman: > ContentBuildLogger.cs is used by both mgcb and the new builder.

@harry-cpp 

I get that.  So what is the problem? 

Is there a bug this causes?  Does it have some adverse effect?

harry-cpp: I don't know. I'm just pointing out that you are not affecting just the content builder output with the change.

SimonDarksideJ:
Needs conflicts resolving @tomspilman

SimonDarksideJ (approved):

PR #8999, opened 2025-09-07
DesktopVK fixes

ThomasFOG:
This PR has several fixes to DesktopVK:
- fixes rendering glitches related to the descriptor cache not working
- fixes possible issues with gamepad stick events
- fixes Vsync support (I've also disabled ```MAILBOX``` mode promotion, as I found that some drivers give false-positives about this mode: they tell you that it's supported, but won't work)
- fixes a crash when minimizing the window on Windows
- fixes exclusive-fullscreen support, changing resolution while in exclusive-fullscreen, restoring window size when going out of exclusive-fullscreen, and a unit test about resizing events
- fixes/better extension compatibility
- reverted ```minImageCount``` as it didn't fix the issue properly, see #9000 
- upgraded SDL to 2.32.10
- some light clean up and compatibility tweaks

### DIFF
--- a/MonoGame.Framework/Platform/Native/GameWindow.Native.cs
+++ b/MonoGame.Framework/Platform/Native/GameWindow.Native.cs
@@ -162,7 +162,13 @@ internal class NativeGameWindow : GameWindow
             MGP.Window_ExitFullScreen(_handle);
         }
 
-        ClientResize(pp.BackBufferWidth, pp.BackBufferHeight);
+        if (_width == pp.BackBufferWidth && _height == pp.BackBufferHeight)
+            return;
+
+        _width = pp.BackBufferWidth;
+        _height = pp.BackBufferHeight;
+
+        MGP.Window_SetClientSize(_handle, pp.BackBufferWidth, pp.BackBufferHeight);
     }
 
     public unsafe void ClientResize(int width, int height)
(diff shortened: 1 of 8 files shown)


MonoGame.Framework/Platform/Native/GameWindow.Native.cs
```diff
@@ -162,46 +162,26 @@ public unsafe void OnPresentationChanged(PresentationParameters pp)
...
-        // TODO: Implement sperate swap change logic for
-        // non-primary windows.
-
-        // TODO: We should expose a feature to allow
-        // for either the swapchain to resize to match
-        // the window size or remain fixed size and stretch.
-
-        // Only the primary window will resize the
-        // default back buffer.
-        if (!_primaryWindow)
```
tomspilman, line 195: Are we handling this somewhere else?

MonoGame.Framework/Platform/Native/GameWindow.Native.cs
```diff
@@ -162,46 +162,26 @@ public unsafe void OnPresentationChanged(PresentationParameters pp)
...
-            return;
+        MGP.Window_SetClientSize(_handle, width, height);
 
-        var manager = _platform.Game.graphicsDeviceManager;
-        if (manager.GraphicsDevice == null)
-            return;
-
-        manager.PreferredBackBufferWidth = _width;
-        manager.PreferredBackBufferHeight = _height;
-        manager.ApplyChanges();
```
tomspilman, line 204: I thought client resize resized the back buffer?  Did it not in XNA?

ThomasFOG: Problem here is that Vulkan doesn't work like that: you resize the surface only (which will automatically expire the swapchain, which will automatically recreate it with the correct backbuffer size).

We had a bunch of problems with vulkan wrongly recreating the swapchain because of unwanted backbuffer resize requests (e.g. we should not recreate the swapchain upon window resize events, just resizing the surface).

This change might be a problem for DirectX12, though.

MonoGame.Framework/Platform/Native/GameWindow.Native.cs
ThomasFOG: Here:

https://github.com/MonoGame/MonoGame/blob/00b09d89eec57e9fb99eb08db7448e2d3194c6ab/native/monogame/vulkan/MGG_Vulkan.cpp#L2094

Vulkan will expire the swapchain whenever the surface is resized. At least that's what SDL expect us to do, and doesn't like manual recreations (that's what broke exclusive-fullscreen).

tomspilman (approved):

MonoGame.Framework/Platform/Native/GameWindow.Native.cs
tomspilman: So you're saying on Vulkan the backbuffer must match the Window size?

ThomasFOG: Not quite. The window pilots the surface, which pilots the backbuffer. When the surface is modified, ```vkQueuePresentKHR()``` will let you know that it changed and that the swapchain must/is ready to be recreated.

If we recreate the swapchain manually, there will be a chain of events that will recreate 2 to 3 times the swapchain on the same frame with different, and most of time broken/wrong sizes (e.g. when exiting fullscreen, SDL sends a 800x600 event that will be processed _after_ VK has gracefully recreated the swapchain, and overriding the correct values ; another one is windows sending a 0x0 resize upon minimizing, which should be ignored and rendering just disabled).

PR #9025, opened 2025-09-20
Added mouse keys to Keys

### LINKED ISSUE #9024: Add mouse keys to Keys enum [PR included]
### Intent

The keyboard keys are already included in the `Keys` enum, which follows the unified key code standard. This standard also includes mouse buttons, so it feels natural to add the mouse buttons to the enum too. The change will unify handling of the keyboard and mouse keys.

### Motivation

Currently, describing mouse keys requires a separate enum, like `MLEM.Input.MouseButton` or hacking the `Keys` enum with type-casting integers to the enum, which may not work properly if the users are expecting a value with a variable assigned to it. This change will hopefully solve this issue, especially when well-designed users will expect updates to all enums.

Monniasza:
Fixes #9024

### Description of Change
This PR adds mouse keys to the Keys enum, to unify the keyboard and mouse keys and buttons.
The keycodes are compliant with https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes?redirectedfrom=MSDN

### DIFF
--- a/MonoGame.Framework/Input/Keys.cs
+++ b/MonoGame.Framework/Input/Keys.cs
@@ -5,7 +5,7 @@
 namespace Microsoft.Xna.Framework.Input
 {
     /// <summary>
-    /// Defines the keys on a keyboard.
+    /// Defines the keys on a keyboard and the mouse according to <a href="https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes">Microsoft doc on key codes</a>.
     /// </summary>	
 	public enum Keys
 	{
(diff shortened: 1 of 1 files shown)


ThomasFOG:
Noting that this will only work for WindowsDX.

Other desktop platforms are going to be problematic: SDL doesn't report those virtual keys as keyboard events, and makes the choice to inject them into mouse events instead. Which means that, beside WindowsDX, it looks hardly implementable for other Windows targets based on SDL (DesktopGL, DesktopVK, and WindowsDX12).

Monniasza:
@ThomasFOG Having it work will need merging the mouse events and keyboard events into a single stream or checking if the keycode is a mouse button in `Keyboard` and delegating it to the `Mouse`. In that class `GetPressedKeys` will have to be modified to also include mouse buttons. In short, mouse and keyboard inputs would have to be merged into one stream prior to dispensing via `Keyboard.GetState`.

ThomasFOG:
And do you think you can do that? We require that PR implement feature parity across targets.

SimonDarksideJ:
Please note @Monniasza it is the policy of MonoGame to ensure any enhancements MUST work across all platforms, including the new Native implementation before being considered for review.

The team have too few resources available to do things piecemeal.  Hope this is ok for you.

SimonDarksideJ:
Any update @Monniasza?
We will have to close this PR currently without change, as XNA did not have the need to define Mouse Buttons as Keys (and in reality they would be buttons in that case).
Please resubmit if you have the time to address the issues identified.

PR #9027, opened 2025-09-25
Use HashSet to track GraphicsDevice resources

riperiperi:
Improves performance of `GraphicsDevice.RemoveResourceReference` (and by extension `GraphicsResource.Dispose`) when there are a lot of resources. 

### Description of Change

`List<T>.Remove(T item)` needs to iterate every item of the list to find the one to remove, then shift items when removing from the middle. HashSet<T> has much faster deletion, at the cost of a slightly slower insert.

My specific example had over 30000 resources, and triggered large stutters when clearing a cache with around 4000 recently allocated items. It was iterating approximately 120000000 resources removing these entries, since they would be at the end of the list.

My test case was a little extreme since I was scrolling a map with a huge number of unique models that stream in+out, but it's an easy improvement to make.

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -8,6 +8,7 @@ using System.Diagnostics;
 using System.Globalization;
 using MonoGame.Framework.Utilities;
 using System.Runtime.InteropServices;
+using System.Linq;
 
 
 namespace Microsoft.Xna.Framework.Graphics
@@ -672,9 +673,7 @@ namespace Microsoft.Xna.Framework.Graphics
                     // Dispose of all remaining graphics resources before disposing of the graphics device
                     lock (_resourcesLock)
                     {
-                        var allResources = new WeakReference[_resources.Count];
-                        _resources.CopyTo(allResources);
-                        foreach (var resource in allResources)
+                        foreach (var resource in _resources.ToArray())
                         {
                             var target = resource.Target as IDisposable;
                             if (target != null)
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -154,7 +154,7 @@ namespace Microsoft.Xna.Framework.Graphics
         // Use WeakReference for the global resources list as we do not know when a resource
         // may be disposed and collected. We do not want to prevent a resource from being
         // collected by holding a strong reference to it in this list.
-        private readonly List<WeakReference> _resources = new List<WeakReference>();
+        private readonly HashSet<WeakReference> _resources = new HashSet<WeakReference>();
 
         // TODO Graphics Device events need implementing
         /// <summary>
@@ -672,7 +672,9 @@ namespace Microsoft.Xna.Framework.Graphics
                     // Dispose of all remaining graphics resources before disposing of the graphics device
                     lock (_resourcesLock)
                     {
-                        foreach (var resource in _resources.ToArray())
+                        var allResources = new WeakReference[_resources.Count];
+                        _resources.CopyTo(allResources);
+                        foreach (var resource in allResources)
                         {
                             var target = resource.Target as IDisposable;
                             if (target != null)
@@ -804,7 +806,7 @@ namespace Microsoft.Xna.Framework.Graphics
                 }
 
                 // Remove references to resources that have been garbage collected.
-                _resources.RemoveAll(wr => !wr.IsAlive);
+                _resources.RemoveWhere(wr => !wr.IsAlive);
             }
         }
 
--- a/Tests/Framework/Graphics/GraphicsDeviceTest.cs
+++ b/Tests/Framework/Graphics/GraphicsDeviceTest.cs
@@ -852,5 +852,32 @@ namespace MonoGame.Tests.Graphics
 
             CheckFrames();
         }
+
+        [Test]
+        [RunOnUI]
+        public void DisposeReferencedResources()
+        {
+            var rt = new RenderTarget2D(gdm.GraphicsDevice, 5, 5);
+            var vb = new DynamicVertexBuffer(gd, VertexPositionColor.VertexDeclaration, 1, BufferUsage.None);
+
+            var middleIb = new DynamicIndexBuffer(gd, IndexElementSize.SixteenBits, 1, BufferUsage.None);
+
+            var ib = new DynamicIndexBuffer(gd, IndexElementSize.SixteenBits, 1, BufferUsage.None);
+            var rtc = new RenderTargetCube(gd, 1, false, SurfaceFormat.Color, DepthFormat.Depth16);
+
+            middleIb.Dispose();
+            Assert.IsTrue(middleIb.IsDisposed);
+
+            // Remaining referenced resources should be disposed.
+            gd.Dispose();
+
+            Assert.IsTrue(rt.IsDisposed);
+
+            Assert.IsTrue(vb.IsDisposed);
+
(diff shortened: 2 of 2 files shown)


SimonDarksideJ:
Can you please add Unit tests to validate the implementation, ensuring it can do comparisons between the original and proposed method.
Validation also needs to be tested on all current platforms, including the new Native platform which has very different requirements on some runtime platforms.

MonoGame.Framework/Graphics/GraphicsDevice.cs
```diff
@@ -804,7 +805,7 @@ internal void OnDeviceResetting()
                 }
 
                 // Remove references to resources that have been garbage collected.
-                _resources.RemoveAll(wr => !wr.IsAlive);
+                _resources.RemoveWhere(wr => !wr.IsAlive);
```
tomspilman, line 808: To my mind this isn't much faster...  maybe even slower in some cases.

Because of the delegate call here it still has to visit all entries in the HashSet. 

Also HashSet content is not linear in memory like List is...  so it will take longer to iterate over all elements.

riperiperi: The normal use case `RemoveResourceReference` doesn't actually appear in the diff, since it still calls `_resources.Remove`. This case here is `OnDeviceResetting`, which I don't think is called at all anymore.

riperiperi:
`_resources` seems to be used for only two reasons: cleaning up resources when disposing the graphics device, and the seemingly unused `OnDeviceResetting` method which would delete contents of all current resources. The expensive part is removing resources from it on dispose.

I added a test for the GraphicsDevice dispose case, where it makes sure all remaining resources are disposed when the device is.

There was already a test for the device reset case called `ContentLostResources`, but it has been disabled because I think the functionality was dummied out.

MonoGame.Framework/Graphics/GraphicsDevice.cs
tomspilman: Some more thoughts.

The code above already iterates over all the `_resources` to call `target.GraphicsDeviceResetting()`.  Question is can we use this first loop to check `wr.IsAlive` and remove items?  Or is it common for `GraphicsDeviceResetting()` to free other dependent resources?  Maybe it is since a resource can be like a `Model` with lots of internal texture/effect/material references.

Worst case i think keeping it a list is fine...  just do a smarter job with removal.  For example:

```
int size = _resources.Count;
for (int i=0; i < size; )
{
  if (!_resources[i].IsAlive)
  {
     var last = size - 1;
      if (i != last)
          _resources[i] = _resources[last];
      size--;
     continue;
  }

  i++;
}
_resources.RemoveRange(size, _resources.Count - size); 
```

This should make removal a single fast list operation and still retain the cash friendly iteration over resources.

riperiperi: The problem is this `Remove` in another function.
https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/GraphicsDevice.cs#L726

I can't really see a case where a device reset would be performance critical.

riperiperi:
Is anything else needed for this? I guess from the comments maybe I should try to explain the purpose of the change again.

The `_resources` list in `GraphicsDevice` is used to remove disposed resources on device reset, and dispose all resources when disposing the `GraphicsDevice`. Resources are added to and removed from the list as they are created and disposed. Adding is trivial as it just appends to the end, but removing from a list by reference is not.

The optimization applies to when a single resource is removed, which used to call `List<T>.Remove(T elem)` to remove the disposed resource, which linearly scanned through the list to find the item, then removed it. On large arrays, the scan and shift of this method could take much longer than expected, and would then be multiplied by the number of items you want to dispose.

Here's a benchmark comparison for `List<T>.Remove` on 5000 `object`s on the end of a 50000 item list, compared to the same with `HashSet<T>.Remove`.

```csharp
using BenchmarkDotNet.Attributes;

namespace PeriBenchmarks
{
    public class MgDisposeSimulacrum
    {
        private const int ExistingObjectCount = 50000;
        private const int RemoveObjectCount = 5000;

        private List<object> _objs;
        private HashSet<object> _objsHashset;
        private object[] _objsToRemove;

        [IterationSetup]
        public void IterationSetup()
        {
            _objs = new List<object>();
            _objsHashset = new HashSet<object>();

            for (int i = 0; i < ExistingObjectCount; i++)
            {
                var obj = new object();
                _objs.Add(obj);
                _objsHashset.Add(obj);
            }

            _objsToRemove = new object[RemoveObjectCount];

            for (int i = 0; i < RemoveObjectCount; i++)
            {
                var obj = new object();
                _objs.Add(obj);
                _objsHashset.Add(obj);
                _objsToRemove[i] = obj;
            }
        }

        [Benchmark]
        public void RemoveList()
        {
            for (int i = 0; i < _objsToRemove.Length; i++)
            {
                _objs.Remove(_objsToRemove[i]);
            }
        }

        [Benchmark]
        public void RemoveHashSet()
        {
            for (int i = 0; i < _objsToRemove.Length; i++)
            {
                _objsHashset.Remove(_objsToRemove[i]);
            }
        }
    }
}
```

```
// * Summary *

BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.26100.7171)
AMD Ryzen 9 3900X, 1 CPU, 24 logical and 12 physical cores
.NET SDK=10.0.100
  [Host]     : .NET 8.0.22 (8.0.2225.52707), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.22 (8.0.2225.52707), X64 RyuJIT AVX2


|        Method |          Mean |        Error |       StdDev |        Median |
|-------------- |--------------:|-------------:|-------------:|--------------:|
|    RemoveList | 182,795.57 us | 3,466.154 us | 3,852.621 us | 182,573.80 us |
| RemoveHashSet |     176.14 us |    24.563 us |    72.040 us |     157.20 us |
```

This is similar to my real-world scenario where I was loading a large number of assets (50000), then streamed in/out additional assets as the camera is scrolled around the map (5000 disposes triggered periodically as cleanup). I could spread out the dispose over multiple frames, but that would just reduce my overall framerate for something that should be a lot cheaper.

There's a tradeoff where adding and removing all items in the list is a little more expensive due to it being a HashSet. The bulk remove case `GraphicsDeviceResetting` doesn't seem to be called, and disposing the `GraphicsDevice` only happens once, so the `Add` case would be the largest impact. Here's benchmark results for adding these same 5000 items to the List and HashSet:

```
|     Method |      Mean |     Error |    StdDev |
|----------- |----------:|----------:|----------:|
| AddHashSet | 251.25 us | 14.107 us | 39.790 us |
|    AddList |  45.06 us |  1.104 us |  3.060 us |
```

The performance loss for adding 5000 items compared to a list was **206.19us**, but the _gain_ for removing them compared to a list was **182,619.43us**, so it more than balances out. Maybe for a very low number of resources the benefit will not be there, but at that point the cost of actually creating/disposing the real resource will be more significant.

---

In case the numbers seem surprising, the cost for object reference comparisons and shifting is much higher than primitives. The same test with a list and hashset of unique `int`s is about 12x faster for List remove:

```
|        Method |         Mean |        Error |       StdDev |       Median |
|-------------- |-------------:|-------------:|-------------:|-------------:|
|    RemoveList | 14,807.49 us | 1,246.243 us | 3,595.697 us | 13,031.30 us |
| RemoveHashSet |     31.64 us |     0.535 us |     0.446 us |     31.80 us |
|    AddHashSet |     35.49 us |     1.581 us |     4.585 us |     37.50 us |
|       AddList |     15.24 us |     0.319 us |     0.914 us |     15.00 us |
```

tomspilman (approved):

Issue #9000, opened 2025-09-07
[DesktopVK] Images in the swapchain are partly hardcoded


ThomasFOG:
This problem will cause some hardware to fail and is a major compatibility issue.

---

When the swapchain is (re)created, a variable number of images are created depending on what the newly created swapchain requires:

https://github.com/MonoGame/MonoGame/blob/00b09d89eec57e9fb99eb08db7448e2d3194c6ab/native/monogame/vulkan/MGG_Vulkan.cpp#L1662

the number of images depend on the driver and can vary depending on the ```presentMode```.

The problem is that the current implementation is hardcoded to support a maximum of ```kConcurrentFrameCount = 2``` images in the swapchain.  While the image creation code is dynamic and can handle more than 2 images, parts of the swapchain resources (command buffers, semaphores, fences) are hardcoded here using the ```kConcurrentFrameCount```, once at initialization and never when the swapchain is recreated:

https://github.com/MonoGame/MonoGame/blob/00b09d89eec57e9fb99eb08db7448e2d3194c6ab/native/monogame/vulkan/MGG_Vulkan.cpp#L1312

Uniform buffers are also lazy initialized upon first rendering a frame bound to an image in the chain:

https://github.com/MonoGame/MonoGame/blob/00b09d89eec57e9fb99eb08db7448e2d3194c6ab/native/monogame/vulkan/MGG_Vulkan.cpp#L1812

**This will cause a hard crash on drivers that will request more than 2 images.**

---

#8903 introduced a largely incomplete and non-working fix by using ```surface_capabilities.minImageCount```:

https://github.com/MonoGame/MonoGame/blob/00b09d89eec57e9fb99eb08db7448e2d3194c6ab/native/monogame/vulkan/MGG_Vulkan.cpp#L1606

but this value is generally wrong with some ```presentMode``` that require more than this number and if a higher number is required, ```vkCreateSwapchainKHR``` will ignore that number and create a bigger swapchain for the ```presentMode``` to work.

I've reverted this change in #8999 

---

To this issue, we need to remove the ```kConcurrentFrameCount``` and work with a dynamic number of resources per images given by ```vkGetSwapchainImagesKHR()```.

For this, we need to destroy and recreate the semaphore, fence, command buffer, and uniform buffer each time the swapchain is (re)created.

[...]

Issue #8833, opened 2025-06-08
DesktopVK - Textures not correctly bound


TheKrumpet:
### Prerequisites

- [x] I have verified this issue is present in the `develop` branch
- [x] I have searched [open](https://github.com/MonoGame/MonoGame/issues) and [closed](https://github.com/MonoGame/MonoGame/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported.

### MonoGame Version

MonoGame develop - DesktopVK

### Which MonoGame platform are you using?

N/A

### Operating System

Windows

### Description

There appear to be an issue where sometimes a texture doesn't appear to be bound correctly to a shader, resulting in any attempt to sample that texture returning 0.

I'm not sure what exactly is causing it. but I've got a small sample project with a simple deferred renderer which exhibits the behaviour. It's in the VKRepro.TextureNotBound project.

I can see in RenderDoc, that in the forward pass, the maps are being correctly populated, so there's definitely data in the textures:

![Image](https://github.com/user-attachments/assets/8052cfcf-777b-4afd-a626-580d2a3dbd9a)

But when it comes to drawing the final scene, there's no textures bound at all:

![Image](https://github.com/user-attachments/assets/ea4b0a7b-7275-4e34-8b28-2c1f77e7c5bf)

Additionally, there's the following error in renderdoc:

![Image](https://github.com/user-attachments/assets/a0a15cc7-d532-430f-bc6e-9c515a05bff4)

This error references the `vkCmdDraw` where the final frame should be drawn into the back buffer.

### Steps to Reproduce

Run the VKRepro.TextureNotBound in the attached project.

### Minimal Example Repo

https://github.com/KristianFenn/VKRepro

### Expected Behavior

The texture is bound correctly.

### Resulting Behavior

The texture isn't bound, meaning all samples return 0. (It's surprising to me that it doesn't just crash).

### Files

The renderdoc capture is available at https://github.com/KristianFenn/VKRepro/blob/main/VKRepro.TextureNotBound/VKRepro.rdc

ThomasFOG:
Should be fixed with #8957

TheKrumpet:
@ThomasFOG I've retested this after those fixes were merged, and the issue persists. The textures just aren't getting bound in the deferred pass.

tomspilman:
hey @TheKrumpet .

I wonder if your card happens to be one affected by the swapchain regeneration issue.

If you have the time can you try these fixes:

[...]

Issue #9005, opened 2025-09-07
[DesktopVK] Rendering glitch depending on the first presentMode


ThomasFOG:
I am running into a very bizarre rendering glitch that depends on the swapchain ```presentMode``` that is first initialized with the first swapchain created, and that will persist over all the next swapchain recreated even if the ```presentMode``` is changed.

<img width="98" height="68" alt="Image" src="https://github.com/user-attachments/assets/f08c4cdc-1a5c-4773-84c5-d164d0dc2a95" />

<img width="98" height="67" alt="Image" src="https://github.com/user-attachments/assets/cb5df053-83c2-43e4-b7b1-eb6f2ccc5a99" />

Left is glitched, right is correct.

This one and the same texture, no composition, just one single sprite, white background included. The entire screen and every assets are impacted.

The thing is: it doesn't look like a Vulkan problem. When profiled with RenderDoc, the image that is passed to the presentation engine is perfect and shows no glitch. The glitch is introduction by the presentation engine when pushed to the surface.

So *it might* be an SDL or a driver bug.

---

The very bizarre aspect of this bug, is that the type of glitch depends on which ```presentMode``` is first set. Each mode shows a different set of bugs that persist beyond ```presentMode``` change, and it is like the ```presentMode``` itself is broken.

I don't know if it's related to providing an ```oldSwapchain``` to ```VkCreateSwapchainKHR()```, but it doesn't look like it (at least the VK specs doesn't mention anything that seems related to this, but who knows what drivers do behind the scene in that regard).

ThomasFOG:
Note: we never saw this problem because we used to default to MAILBOX ```presentMode```, which doesn't show issues when first set (but I removed MAILBOX because I found it wasn't working on NVIDIA despite the driver telling that it should; and using MAILBOX didn't explain the problem, so I think we shouldn't think of it as a workaround).

PR #9044, opened 2025-10-10
WIP Vulkan Fixes

tomspilman:
This is a WIP fix for #9000 and #8833.


ThomasFOG:
I tested the PR.

The good things:
- it fixes #9000
- #9005 seems fixed, but maybe not (see next point)

The bad things:
- there are new flickers that were not present before. It kinda looks like the same flickers that were there when the descriptor cache had misses. Or maybe it is still #9005 but it took a different form.
- fullscreen and vsync mode switches are acting weird. They both trigger a black screen for a few seconds, and exclusive fullscreen doesn't respect the requested resolution anymore.

tomspilman:
> * there are new flickers that were not present before. It kinda looks like the same flickers that were there when the descriptor cache had misses. Or maybe it is still [[DesktopVK] Rendering glitch depending on the first presentMode #9005](https://github.com/MonoGame/MonoGame/issues/9005) but it took a different form.

I'm going to work on this issue next.  I saw flickers on my system when i forced it to recreate things on resize of the window.  I suspect if i fix this issue on my end it may fix your cases as well.

tomspilman:
@ThomasFOG  So i think the flickers are fixed with the latest commit.  

Let me know how it works for you.

tomspilman:
@ThomasFOG 

> and exclusive fullscreen doesn't respect the requested resolution anymore.

Can you give me a reproduction case for this?  What specifically were you doing?

tomspilman:
@ThomasFOG  Would be a good time for you to try your test case again.  Or we can just call it fixed and merge it as is.

ThomasFOG:
It's much better but still not quite there:
- No problem with vsync and blackscreens are gone;
- I'm still getting flickers every few seconds that look like a corrupted checkerboard image (descriptor cache misses?). I unfortunately couldn't capture a corrupted frame with RenderDoc despite trying for half an hour. I can record a video but I think it won't be useful without a RenderDoc trace;
- Exclusive-fullscreen mode works mostly correctly, the only problem being the Viewport being wrong when using RenderTargets: sometime the Viewport doesn't follow the backbuffer size change and stays the same as before the resolution switch. This also happens when resizing a window, so I think it's not directly related to fullscreen mode (but likely to swapchain resizing not updating the viewport). If I swap resolutions a little, at some point the viewport is correct (and we're talking about only the value of ```GraphicsDevice.Viewport```, the actual drawing surface is correct).

tomspilman:
> It's much better but still not quite there:

Lame.

> I can record a video 

At least it tells me what to look for.  

What device did you have this occur on?

> the only problem being the Viewport being wrong when using RenderTargets
> sometime the Viewport doesn't follow the backbuffer size change and stays the same as before the resolution switch. 

That is something i can likely test and fix.

You mean specifically when using offscreen an `RenderTarget` ... `device.Viewport` is wrong before or after the target is applied?

ThomasFOG:
> You mean specifically when using offscreen an `RenderTarget` ... `device.Viewport` is wrong before or after the target is applied?

- Render to a target
- Switch back to no target
- Clear to green
- Render that target stretched to the current Viewport size: it won't be stretched to the current window size, but to the previous one (so oversized, or undersized with green visible depending on the resolution change upward/backward)

ThomasFOG (approved):

PR #9050, opened 2025-10-15
[Pipeline] Warning Cleanup 1000

harry-cpp:
The title should be pretty obvious no, probably no need for any description :P 

But in all seriousness, this does the following:
- Cleans up all the warnings we have in the MonoGame.Content.Pipeline and MGFXC
- Enables TreatWarningAsErrors on those 2 projects
- 3rd party code like CppNet has not been fixed, instead I've just added the `#pragma warning disable` at the top to disable its warnings
- Does a bit of code cleanup where applicable (and I've used some newer C# stuff that I've been using in my personal projects, new `[]` stuff for collections are a god sent)
  - Focus was mostly on warning cleanup so there is def more that can be done for this
- When console projects update to the latest MG, they will need to fix up any warnings they have in their files or add `#pragma warning disable` at the top of their files
- I AM NOT CRAZY I SWEAHAHAHAHAHA

### Contributor Declaration


arosolino:
@harry-cpp I like this kind of crazy

web3dev1337:
Hi @harry-cpp 

I went through the PR and spotted a few things to double check. I’ve included the relevant code snippets below to make them easy to double-check.

## 🔴 

### 1. Vertex declaration reader leaves the stream misaligned
`MonoGame.Framework/Content/ContentReaders/VertexDeclarationReader.cs:13`

```csharp
var vertexStride = reader.ReadInt32();
if (vertexStride == 0)
{
    return VertexDeclaration.GetOrCreate(0, []);
}

var elementCount = reader.ReadInt32();
```

The writer always emits both values, so returning before we consume `elementCount` leaves four unread bytes on the stream. Suggested tweak:

```csharp
var vertexStride = reader.ReadInt32();
var elementCount = reader.ReadInt32();
if (vertexStride == 0)
{
    return VertexDeclaration.GetOrCreate(0, []);
}
```

### 2. Six interpolated messages lost their `$`

All of these now log the literal `{…}` text:

```csharp
// BitmapContent.cs:143
public override string ToString() => "{GetType().Name}, {Width}x{Height}";

// TextureContent.cs:46
throw new ArgumentException("Type '{newBitmapType}' does not have a constructor...", nameof(newBitmapType));

// ShaderProfile.DirectX11.cs:37 / 44
throw new Exception("Invalid profile '{pass.vsModel}'. Vertex shader '{pass.vsFunction}' must be SM 4.0 level 9.1 or higher!");

// ShaderProfile.OpenGL.cs:37 / 44
throw new Exception("Invalid profile '{pass.vsModel}'. Pixel shader '{pass.psFunction}' must be SM 3.0 or lower!");
```

Adding the `$` prefix to each of the above restores the runtime values.

## ⚠️ Breaking API changes to double-check

### BitmapContent setters
`MonoGame.Framework.Content.Pipeline/Graphics/BitmapContent.cs`

```csharp
public int Height
{
    get => _height;
    internal set
    {
        ArgumentOutOfRangeException.ThrowIfNegativeOrZero(value);
        _height = value;
    }
}
```

These setters were `protected` and are now `internal`. Anyone deriving from `BitmapContent` outside this assembly can no longer set them.

### AnimationContent.Name vs. ContentSerializer
`MonoGame.Framework.Content.Pipeline/Graphics/AnimationContent.cs`

```csharp
[ContentSerializer(Optional = true)]
public required new string Name { get; set; }
```

We now require `Name`, but the serializer still treats it as optional. Callers that relied on the optional contract will start throwing.

### EffectContent.Identity is required without initialization
`MonoGame.Framework.Content.Pipeline/Graphics/EffectContent.cs`

```csharp
public new required ContentIdentity Identity { get; set; }
```

Existing code that does `new EffectContent()` won’t populate this property, so validation will fail at build time.

### PixelBitmapContent `_pixelData` accessibility
`MonoGame.Framework.Content.Pipeline/Graphics/PixelBitmapContent.cs`

```csharp
private readonly T[][] _pixelData;
```

This used to be `internal`. Some of our helpers (e.g., color conversions) still reach in directly; those now fail to compile.

harry-cpp:
> EffectContent.Identity is required without initialization

Better a compile time error then a runtime crash.

> PixelBitmapContent _pixelData accessibility

`internal` is used for, well, being able to internally access some API without expositing it to the public API. If you are relying on any such API, that means you are referencing the source code directly and there is a good chance that something might get broken with their accessors down the line.

If you need access to this variable, please make a separate issue / PR to add the appropriate public property for the things you need.

AristurtleDev:
@harry-cpp need merge conflicts resolved and also the LLM statement added to the PR description

```
### Contributor Declaration

```

harry-cpp:
Merge conflicts resolved, but had to fix a bunch of new files 😅 

Lets see if this passess.

MonoGame.Framework.Content.Pipeline/Graphics/NodeContent.cs
```diff
@@ -10,84 +10,44 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
...
     public class NodeContent : ContentItem
     {
-        Matrix transform;
-        NodeContent parent;
-        NodeContentCollection children;
-        AnimationContentDictionary animations;
+        /// <summary>
+        /// Gets or sets the name of the content item.
+        /// </summary>
+        public required new string Name { get; set; }
```
AristurtleDev: Do we need to hide `ContentItem.Name` with `new` here? Since `ReflectiveWriter` and `ReflectiveSerializer` walk base `ContentItem` and derived members separately, this looks like it could introduce duplicate `Name` state/serialization behavior.

MonoGame.Framework.Content.Pipeline/Graphics/EffectContent.cs
```diff
@@ -10,16 +10,13 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
...
     {
         /// <summary>
-        /// Initializes a new instance of EffectContent.
+        /// Gets or sets the identity of the content item.
         /// </summary>
-        public EffectContent()
-        {
-
-        }
+        public new required ContentIdentity Identity { get; set; }
```
AristurtleDev: Same here for hiding `ContentItem.Identity`

MonoGame.Framework.Content.Pipeline/Graphics/AnimationChannelDictionary.cs
```diff
@@ -7,13 +7,5 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Graphics
     /// <summary>
     /// Collection of animation data channels, one per bone or rigid object.
     /// </summary>
-    public sealed class AnimationChannelDictionary : NamedValueDictionary<AnimationChannel>
-    {
-        /// <summary>
-        /// Initializes a new instance of AnimationChannelDictionary.
-        /// </summary>
-        public AnimationChannelDictionary()
```
AristurtleDev, line 15: This is a dumb nitpick and feel free to ignore it, but removing the ctor here even though it's an empty ctor removes the API documentation that it would generate for the site and for intellesense when someone calls `new AnimationChannelDictionary()`. 

Just noting the observation.

AristurtleDev (commented):
A couple of notes if you don't mind responding @harry-cpp and I can approve and merge

MonoGame.Framework.Content.Pipeline/Graphics/AnimationChannelDictionary.cs
harry-cpp: You'll get the class documentation instead, this constructor documentation told you nothing.

image

MonoGame.Framework.Content.Pipeline/Graphics/NodeContent.cs
harry-cpp: Yea, these were a bit awkward to make work because for everything else, Name and Identity are optional, except in these 2 cases where they are suddenly mandatory 😓 

But now that you mention it, I think it will be better to add some extra checks in effect processor then hiding underlying values.

AristurtleDev (approved):

PR #9051, opened 2025-10-15
Use global.json to pin the SDK version we use to build

dellis1972:
Use global.json to make sure we use a specific .net sdk.

Set the working directory on Cake Settings so that it picks up global.json on MacOS.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,0 +1,6 @@
+{
+    "sdk": {
+        "version": "8.0.404",
+        "rollForward": "disable"
+    }
+}
--- a/global.json
+++ b/global.json
@@ -1,6 +1,0 @@
-{
-    "sdk": {
-        "version": "8.0.404",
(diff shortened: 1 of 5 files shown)


dellis1972:
Hmm Windows does not seem to want to pick up global.json for the iOS build

global.json
```diff
@@ -0,0 +1,6 @@
+{
```
harry-cpp: I think we should generate this file on github actions instead of it being present in the repo, since its only needed for github actions.

harry-cpp: When I say that I mean from main.yaml, not from CAKE script :)

dellis1972: I know, it needs to be done in yaml cos that is where we install all the workloads and dotnet itself.

harry-cpp (approved):

Issue #9001, opened 2025-09-07
[DesktopVK] Fix the unit tests


ThomasFOG:
A few tests are failing and need fixing.

<img width="314" height="282" alt="Image" src="https://github.com/user-attachments/assets/e7e6d5e8-60af-4ff1-b3d0-930f9e6332df" />

tomspilman:
Going to take this on tomorrow on stream.

tomspilman:
Fixed via #9196 and #9324.

PR #9066, opened 2025-10-25
Fix for desktopvk float3x3 effect issue

### LINKED ISSUE #9019: float3x3 does not work with Vulkan shaders
### Prerequisites

### MonoGame Version

develop

DesktopVK

### Operating System

Windows

### Description

If you compile a shader to use a float3x3 parameter, such as:

`
float3x3 Layer1Colors;
`

It results in the error
https://github.com/MonoGame/MonoGame/blob/fb9fa86f3092c833460730da45f2c6b28ec5f3ea/MonoGame.Framework/Graphics/Shader/ConstantBuffer.cs#L95
`
System.ArgumentException: 'Offset and length were out of bounds for the array or count is greater than the number of elements from index to the end of the source collection.'
`

The error happens because the size of the `_buffer` is `36` and it's expecting it to be `48` because of this line
https://github.com/MonoGame/MonoGame/blob/fb9fa86f3092c833460730da45f2c6b28ec5f3ea/MonoGame.Framework/Graphics/Shader/ConstantBuffer.cs#L69

This works correctly in other frameworks such as DesktopGL & WindowsDX

### Steps to Reproduce

1. Use the DesktopVK repo
2. Create a shader that users float3x3 as parameter

### Minimal Example Repo

https://github.com/arosolino/hit-start-monogame/tree/main/issues/desktopvk-constant-buffer

### Expected Behavior

Should calculate the correct row size

### Resulting Behavior

Does not calculate the correct row size

### Files

_No response_

JakeLegendXIII:
Fixes #9019 

### Description of Change
This resolves the DesktopVK runtime issue that shows up in ConstantBuffer.cs for an effect with a float3x3 parameter. I have verified testing with the sample project in the issue. Basically just forcing matrix or array items to have padding to 16 bytes in the ConstantBufferData.Vulkan.cs to better match the expectations for the buffer at runtime instead of using the actual data size. It looks like the other effect compilation implementations work a bit differently, but end up doing the same kind of thing. The DX implementations receive pre-calculated sizes from reflection APIs and the base mojo/openGL version uses a register based calculation. 

I am creating this as a Draft as I am still checking some things and testing it. I am open to suggestions as I am just getting into the VK parts of the code.


tomspilman:
Have we enabled the graphics unit tests for Vulkan yet?  @SimonDarksideJ @ThomasFOG  ?

We really should and flag tests that don't work on vulkan with some special flag to disable them for now.  We can then make a task to fix these tests specifically.

I would like to see the tests on and ensure this case has a unit test.

JakeLegendXIII:
If we want I can add an additional ShaderTest to the suite that replicates this issue and verifies the fix. Will probably just take me a a few days to get to it. Have to do the fx build out for the different platforms and frame capture/reference image stuff I imagine.

The adding tests for CI and tagging/skipping may require some additional discussion. There is issue #9001 for fixing the Unit Tests. When I run the tests locally including the graphics tests only the same 6 are failing that are mentioned in the issue. So the other ShaderTests are passing for VK.

ThomasFOG:
This looks fine to me.

---

Ultimately, we should rewrite the SPIRV parser to actually decode the SPIRV opcodes instead of taking shortcuts. The current parser should work in most cases, but if we want something really robust and not calculating padding manually, we should read the opcodes and unfurl SPIRV pointers to get types and their size.

JakeLegendXIII:
Hey @tomspilman I went ahead and added an additional shader test that covers this case. The xnb that is added for Vulkan was created with my fix in place, so it will just work. I did test it by reverting my change and seeing the test failure with an xnb created with the original code. It gets the same exception as the repro program and makes the test fail in that case. 

Update: I went ahead and rebuilt the xnbs for the other Vulkan shader tests just so they were consistently ran with the new compiler changes.

JakeLegendXIII:
With the merge of #9082 I have some conflicts in here and want to re-test and re-evaluate the fix with the changes. The issue appears to still be happening with the new parsing code, but I may need to abandon this PR and rework a fix on top of the new code. Moving it back to draft for now.

TheKrumpet:
I hadn't seen this PR, apologies. I've currently got this under development, which should fix this issue by taking the offset/sizing information from the SPIR-V:
https://github.com/MonoGame/MonoGame/compare/develop...TheKrumpet:MonoGame:feature/VulkanCBufferSizing

JakeLegendXIII:
> I hadn't seen this PR, apologies. I've currently got this under development, which should fix this issue by taking the offset/sizing information from the SPIR-V: [develop...TheKrumpet:MonoGame:feature/VulkanCBufferSizing](https://github.com/MonoGame/MonoGame/compare/develop...TheKrumpet:MonoGame:feature/VulkanCBufferSizing)

Well that works for me. I can close this one out as that is the better resolution and better testing! Nice work :)

PR #9068, opened 2025-10-26
Reveal amount of textures in TextureCollection

rds1983:
### Description of Change

Reveals amount of textures in the TextureCollection.
Example usage:
```c#
  // Reset the graphics device textures
for(var i = 0; i < GraphicsDevice.Textures.Count; ++i) 
{
  GraphicsDevice.Textures[i] = null;
}
```

The maximum returned amount of textures is limited by 16 for two purposes:
* Make sure the logic with modifying `_dirty` field won't be broken
* Doubtful anyone would need more than 16 textures anyway

### DIFF
--- a/MonoGame.Framework/Graphics/TextureCollection.cs
+++ b/MonoGame.Framework/Graphics/TextureCollection.cs
@@ -16,6 +16,11 @@ namespace Microsoft.Xna.Framework.Graphics
         private ShaderStage _stage;
         private int _dirty;
 
+        /// <summary>
+        /// Amount of textures
+        /// </summary>
+        public int Count => _textures.Length;
+
         internal TextureCollection(GraphicsDevice graphicsDevice, int maxTextures, ShaderStage stage)
         {
             _graphicsDevice = graphicsDevice;
--- a/MonoGame.Framework/Graphics/TextureCollection.cs
+++ b/MonoGame.Framework/Graphics/TextureCollection.cs
@@ -19,7 +19,13 @@ namespace Microsoft.Xna.Framework.Graphics
         /// <summary>
         /// Amount of textures
         /// </summary>
-        public int Count => _textures.Length;
+        public int Count
+        {
+            get
+            {
+                return _textures.Length;
+            }
+        }
 
         internal TextureCollection(GraphicsDevice graphicsDevice, int maxTextures, ShaderStage stage)
         {
--- a/MonoGame.Framework/Graphics/TextureCollection.cs
+++ b/MonoGame.Framework/Graphics/TextureCollection.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Xna.Framework.Graphics
         {
             get
             {
-                return _textures.Length;
+                return Math.Min(_textures.Length, 16);
             }
         }
 


ThomasFOG:
It's capped and initialized to a capacity of 16 on all platforms, I think it's totally safe to just return ```_textures.Length```.
```_isDirty``` will work up to 32, maybe we should add an exception in the ```TextureCollection``` ctor to test if ```maxTextures``` is not greater than 32 just to make sure that it's future-proof (and that an exception is triggered if we ever go beyond that value).

rds1983:
It isn't capped. TextureCollection has 192 entries on my gpu(3060 ti).

ThomasFOG:
It is capped to 16 inside MonoGame for each backend:
- D3D12: https://github.com/MonoGame/MonoGame/blob/0ec05e8033000681e95bec784f126cd982a4b9da/native/monogame/directx12/MGG_DX12.cpp#L55
- D3D11: https://github.com/MonoGame/MonoGame/blob/0ec05e8033000681e95bec784f126cd982a4b9da/MonoGame.Framework/Platform/Graphics/GraphicsDevice.DirectX.cs#L56
- Vulkan: https://github.com/MonoGame/MonoGame/blob/0ec05e8033000681e95bec784f126cd982a4b9da/native/monogame/vulkan/MGG_Vulkan.cpp#L1513

There is only OpenGL that gets the number of available slots (but with the wrong property, it should be ```MaxTextureImageUnits```, which should give a number close to half of the current value from ```MaxCombinedTextureImageUnits```): https://github.com/MonoGame/MonoGame/blob/0ec05e8033000681e95bec784f126cd982a4b9da/MonoGame.Framework/Platform/Graphics/GraphicsDevice.OpenGL.cs#L260


16 is a good number that guarantees compatibility with a great range of hardware.

rds1983:
True. I use DesktoPGL, which has no limits.
I agree that 16 is good number. Hence it's in the PR.

ThomasFOG:
The problem with the PR is that it doesn't cap the number of slots, only its property.

For consistency, I think that a cap should be added after this line:

https://github.com/MonoGame/MonoGame/blob/0ec05e8033000681e95bec784f126cd982a4b9da/MonoGame.Framework/Platform/Graphics/GraphicsDevice.OpenGL.cs#L260

something in the line of ```MaxTextureSlots = Math.Max(MaxTextureSlots, 16)```.

And the property should just be ```return _textures.Legnth;```.

dellis1972:
If we expose the Count on one  `*Collection` class should it be done on all of them for consistency? e.g MonoGame.Framework/Graphics/SamplerStateCollection.cs

SimonDarksideJ:
I think @dellis1972 suggestion is fair @rds1983 , as this is focusing on exposing public readonly values, I think it should be consistent on the other collections.  Are you ok also making the suggested change too?

SimonDarksideJ:
> If we expose the Count on one `*Collection` class should it be done on all of them for consistency? e.g MonoGame.Framework/Graphics/SamplerStateCollection.cs

@rds1983 are you able to update the PR in line with Dean's suggestion above this week for inclusion in `3.8.6`

PR #9082, opened 2025-11-09
Add more robust SPIR-V parsing to Vulkan shader compilation

TheKrumpet:
### Description of Change

This PR adds some more robust parsing of the SPIR-V file generated when compiling a Vulkan shader, with the aim of reducing reliance on unstable, non-guaranteed IDs generated by DXC, preferring to pull any required information from places guaranteed by the spec. This should improve the stability of Vulkan shader compilation against changes in any tools used.

### Outstanding:
- ~Look into adding array support now we have good array type data from the SPIR-V~ - Done
- ~Look into adding multiple cbuffer support (this should be trivial now)~ - To be handled in a seperate PR
- ~Look into DXC failures related to the `-Fd` flag being used with `-spirv`~
- ~Take a pass on errors/warnings coming from the parsing. Given the way the file is structured, it's hard to know if a failure to parse something earlier will cause an issue later, but leaving those breadcrumbs will almost certainly be useful for debugging. I'm curious if there's any standards around using debug/trace/info output streams?~
- ~merge actually important warnings/errors into the warnings/errors used in the shader compilation~
- ~Look at the case when there's no matched sampler state. Previously it would just crash with a `NullReferenceException` when trying to assign the sampler state, but it looks like it was at least trying to default the state previously.~
- ~Look at semantic mapping for input vertex attributes. This is currently relying on the ID provided by DXC, which isn't guaranteed to be stable. It may be this is the best we can do for now, but it is a potential point of weakness.~


Tools/MonoGame.Effect.Compiler/Effect/Spirv/Types/SpirvType.cs
```diff
@@ -0,0 +1,111 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
```
tomspilman: Use the new foundation copyright header in these:

```
﻿// MonoGame - Copyright (C) MonoGame Foundation, Inc
```

Tools/MonoGame.Effect.Compiler/Effect/Spirv/Types/SpirvTypeArray.cs
```diff
@@ -0,0 +1,33 @@
+﻿// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+using System.Diagnostics;
+
+namespace MonoGame.Effect.Compiler.Effect.Spirv
+{
+    public class SpirvTypeArray : SpirvTypeBase
```
tomspilman: Seems like all these classes are internal to the Effect compiler system and shouldn't be `public`.

Tools/MonoGame.Effect.Compiler/Effect/Spirv/Types/SpirvType.cs
TheKrumpet: I'll do that, I just grabbed this from a file at random so I'll also try and catch any others that are wrong.

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.Vulkan.cs
```diff
@@ -262,148 +232,106 @@ internal override ShaderData CreateShader(ShaderResult shaderResult, string shad
...
-                        {
-                            VkDescriptor d;
-                            if (!descriptors.TryGetValue(data[1], out d))
-                            {
-                                d = new VkDescriptor();
-                                d.name = data[1];
-                                d.type = VkDescriptorType.UNIFORM_BUFFER;
-                                descriptors.Add(d.name, d);
-                            }
+                        ConstantBufferData cbuffer = new ConstantBufferData("global");
```
tomspilman, line 259: We need to deal with this soon too in another PR.  See https://github.com/MonoGame/MonoGame/issues/9083.

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.Vulkan.cs
```diff
@@ -262,148 +232,106 @@ internal override ShaderData CreateShader(ShaderResult shaderResult, string shad
...
-                var cbuffer = new ConstantBufferData("global");
+                            // This image is only sampled by one sampler, we can safely use the texture name for the parameter.
+                            if (sampledImages.Count() == 1)
+                            {
+                                sampler.parameterName = imageVariable.Name;
+                            }
+                            // otherwise make a composite name for this image/sampler combo.
+                            else
+                            {
+                                sampler.parameterName = $"{samplerVariable.Name}+{imageVariable.Name}";
```
tomspilman, line 312: I do want to cleanup all the mess with sampler/texture issues and make it consistent across all target shader profiles.  This seems like a good way forward.

Tools/MonoGame.Effect.Compiler/Effect/ShaderProfile.Vulkan.cs
```diff
@@ -262,148 +232,106 @@ internal override ShaderData CreateShader(ShaderResult shaderResult, string shad
...
-                                    errorsAndWarnings += "Building effects for Vulkan currently doesn't support more than one constant buffer (cbuffer) structures. Please consider refactoring your HLSL code.";
-                                    throw new ShaderCompilerException();
-                                }
-
-                                // Gather uniforms.
-                                foreach (var member in globals.members.Values)
-                                    cbuffer.AddParameter(member.name, member.type, 0, member.offset);
+                                // TODO: should we just default the sampler state here?
+                                // would an exception be more appropriate?
+                                samplerStateInfo = new SamplerStateInfo();
```
tomspilman, line 334: Maybe a logged warning saying we're using a default sampler state?

tomspilman:
This is all looking great @TheKrumpet .  I added a few notes, but i'll push to get this merged quickly so we can move forward.

ThomasFOG:
That's a nice SPIRV parser (the SPIRV specs are really cool for that).

It will sure help with supporting multiple constant buffers and other kind of data like structured buffers.

TheKrumpet:
@ThomasFOG  The SPIR-V spec is one of the best pieces of technical documentation I've ever worked with, it made all of this much simpler.

@tomspilman I think I'm pretty happy with where this is now. Things I've changed since the first review:

- Added support for array params
- Added a flag to the DXC invoke which will cause it to emit HLSL semantics into the SPIR-V, and used those over parsing the IDs
- Added a test suite for all of the types of shader parameter that appear to be supported, to make sure everything is getting set correctly.

tomspilman:
@SimonDarksideJ @CartBlanche I've rebuilt the Mac build like 4 or 5 times now...  it always fails with Android SDK issues.

What do we need to do here?

SimonDarksideJ:
The sheer amount of nullability failures is a cause for concern (although not directly related), these should be handled or nullability turned off at the project level to avoid such warnings.

As for Android, I've kicked off one more build, but as we have previously discussed turning off build testing for Android on Mac, I'll see about a PR to fix that, then you can refresh this PR from the updated base without Android Build on Mac

SimonDarksideJ:
Once #9086 is merged, this will need pulling from `develop` into this PR to skip Android building on MacOS

SimonDarksideJ:
@tomspilman can you approve and merge the Mac Android fix PR, so it can be pulled into this PR to continue building?
https://github.com/MonoGame/MonoGame/pull/9086

tomspilman (approved):

PR #9089, opened 2025-11-12
dotnet 10 automation test

SimonDarksideJ:
Dot net 10 automation test


SimonDarksideJ:
Not yet it seems @harry-cpp , not listed as available for automation install yet it seems :D
https://github.com/dotnet/core/blob/main/os-lifecycle-policy.md

Will check back in a few days :P

PR #9093, opened 2025-11-12
Fixed Vulkan validation errors introduced by parsing changes

TheKrumpet:
Recent changes to the Vulkan shader parsing have unearthed a couple of validation errors:

### VK_EXT_SCALAR_BLOCK_LAYOUT
This one is required when using `-fvk-use-dx-layout` with DXC: https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#memory-layout-rules. This could have happened before, but it required having a shader where the parameters would straddle the 16 byte boundaries.

### VK_GOOGLE_HLSL_FUNCTIONALITY1 and VK_GOOGLE_USER_TYPE
These were introduced by the extra reflection flags added in [ac7f3292bc](https://github.com/MonoGame/MonoGame/pull/9082/commits/ac7f3292bc69314262283c5619fd67fc1a2b2332#diff-587d4bf7d0531e8166e9a2b2ebcc3a9f72877e799d5408382c55c03da19c4713R159). These features aren't required to run the shaders, but as they show up in the capabilities it causes validation errors. An alternative fix would be to either:
- Run DXC twice, once with the flag for reflection info and once without for the shader bytecode. 
- Strip the offending lines from the shader bytecode after compilation manually. 

These options do open the possibility of the shader being subtly different in reflection vs compilation, but that's probably _fairly_ unlikely.

As these will likely conflict with changes in #9044 I'll wait for those to merge and then fix any issues.

### DIFF
--- a/native/monogame/vulkan/MGG_Vulkan.cpp
+++ b/native/monogame/vulkan/MGG_Vulkan.cpp
@@ -1229,12 +1229,22 @@ MGG_GraphicsDevice* MGG_GraphicsDevice_Create(MGG_GraphicsSystem* system, MGG_Gr
 	}
 
 	bool swapChainSupported = false;
+	bool scalarBlockLayoutSupported = false;
+	bool hlslFunctionalitySupported = false;
+	bool userTypeSupported = false;
+
 	for (const auto& extension : deviceExtensions)
 	{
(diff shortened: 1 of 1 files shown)


tomspilman (approved):

tomspilman:
Seems ok to me.

PR #9096, opened 2025-11-14
Update installation targets for Visual Studio versions

dellis1972:
Add support for VS2026


SimonDarksideJ:
As this changes from 17 (2019) / 18 (2022) for 17 / 19 (2026), will this still install and work under 2022 (with v18 not listed) @dellis1972 ?

SimonDarksideJ:
> As this changes from 17 (2019) / 18 (2022) for 17 / 19 (2026), will this still install and work under 2022 (with v18 not listed) @dellis1972 ?

Of, question answered offline, the values represent a range, Tested in 2022, just installing 2026 to verify and test there too

SimonDarksideJ (approved):

PR #9098, opened 2025-11-14
Remove Android building/testing on Linux

SimonDarksideJ:
# Summary

Even with the entire pipeline FORCED to use DotNet 9, for some reason Android is using DotNet 10 on Linux and is breaking the build.

Force removing it for now to enable PR's to pass testing

Windows is now the only platform to handle Android

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,11 +40,6 @@ jobs:
         run: |
           dotnet workload install android ios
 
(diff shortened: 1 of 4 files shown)


SimonDarksideJ:
Reverting back to draft until a descion is made on the best course of action

SimonDarksideJ:
Closing in favour of #9102 , the  issue is with using a wildcard for the dotnet version

PR #9116, opened 2025-11-23
Feature Opengl Native

dellis1972:
A work in progress, an implementation of OpenGL for the new native backend. 

- Currently only supports Clearing the screen. 
- Mostly about setting up the infrastructure to build it using Emscripten.
- Only supports WebGL at this point, not tested on Deskop

the plan is for this backend to ONLY support modern OpenGL 4.1+ or WebGL 2.0

Example csproj for WebGL

```xml
<Project>
  <Sdk Name="Microsoft.NET.Sdk.WebAssembly" />
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net9.0</TargetFramework>
    <MonoGamePlatform>Web</MonoGamePlatform>
    <DefineConstants>$(DefineConstants);MONOGAME;MG_$(MonoGamePlatform)</DefineConstants>
    <!-- This is to basically disable globalization to exclude icudt.dat (1.5MB) and reduce size of dotnet.wasm -->
    <InvariantGlobalization>true</InvariantGlobalization>
    <WasmBuildNative>true</WasmBuildNative>
    <WasmAllowUndefinedSymbols>true</WasmAllowUndefinedSymbols>
    <EmccExtraLDFlags>-sFULL_ES3</EmccExtraLDFlags>
    <WasmEmitSymbolMap>true</WasmEmitSymbolMap>
    <EmccFlags>-sVERBOSE=1 -Wbad-function-cast -Wcast-function-type -O2 -g3 -sINITIAL_MEMORY=128MB -sMAXIMUM_MEMORY=2048MB -sALLOW_MEMORY_GROWTH=1</EmccFlags>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="..\MonoGame.Framework\MonoGame.Framework.Native.csproj" />
  </ItemGroup>

  <ItemGroup>
    <LinkerArg Include="-s USE_WEBGL2=1 -s FULL_ES3=1 -s TOTAL_STACK=5242880" />
    <WasmShellExtraEmccFlags Include="--preload-file $(MSBuildThisFileDirectory)..\Content@/Content" />
    <NativeFileReference Include="..\Artifacts\monogame.native\emscripten\desktopwasm\$(Configuration)\monogame.native.a" CopyToOutputDirectory="PreserveNewest" />
    <NativeFileReference Include="..\Artifacts\monogame.native\emscripten\desktopwasm\$(Configuration)\SDL2.a" CopyToOutputDirectory="PreserveNewest" />
  </ItemGroup>

</Project>
```

Example Program.cs, WebGL is Asynchronous so we need a way to pause the dotnet runtime after `Run` returns. Otherwise the app will just exit. This is because calling `emscripten_set_main_loop` with a `true` for `simulateInfiniteLoop` results in a deadlock. 

```csharp
public static class Program
{
    [STAThread]
    static async Task Main()
    {
        TaskCompletionSource<bool> tcs = new TaskCompletionSource<bool>();
        using (var game = new Game1())
        {
            game.Run();
            await tcs.Task;
        }
    }
}
```

image

image

### DIFF
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -537,7 +537,7 @@ namespace Microsoft.Xna.Framework
                 // We only have a precision timer on Windows, so other platforms may still overshoot
 #if WINDOWS && !DESKTOPGL
                 MonoGame.Framework.Utilities.TimerHelper.SleepForNoMoreThan(sleepTime);
-#elif DESKTOPGL || ANDROID || IOS
+#elif DESKTOPGL || ANDROID || IOS || NATIVE
                 if (sleepTime >= 2.0)
                     System.Threading.Thread.Sleep(1);
 #endif
--- a/MonoGame.Framework/Platform/Native/Graphics.Interop.cs
+++ b/MonoGame.Framework/Platform/Native/Graphics.Interop.cs
@@ -6,6 +6,7 @@ using Microsoft.Xna.Framework;
 using Microsoft.Xna.Framework.Graphics;
 using System;
 using System.Drawing;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 
--- a/MonoGame.Framework/Platform/Native/VertexElement.Native.cs
+++ b/MonoGame.Framework/Platform/Native/VertexElement.Native.cs
@@ -3,6 +3,7 @@
 // file 'LICENSE.txt', which is part of this source code package.
 
 using System;
+using System.Runtime.InteropServices;
 using MonoGame.Interop;
 
 namespace Microsoft.Xna.Framework.Graphics;
--- a/native/monogame/opengl/MGG_OpenGL.cpp
+++ b/native/monogame/opengl/MGG_OpenGL.cpp
@@ -1,0 +1,880 @@
+// MonoGame - Copyright (C) The MonoGame Team
+// This file is subject to the terms and conditions defined in
+// file 'LICENSE.txt', which is part of this source code package.
+
+#include "api_MGG.h"
+#include "mg_common.h"
+
+// Effect includes for OpenGL
+// #include "AlphaTestEffect.ogl.mgfxo.h"
+// #include "BasicEffect.ogl.mgfxo.h"
+// #include "DualTextureEffect.ogl.mgfxo.h"
+// #include "EnvironmentMapEffect.ogl.mgfxo.h"
+// #include "SkinnedEffect.ogl.mgfxo.h"
+// #include "SpriteEffect.ogl.mgfxo.h"
+//#include "mg_effect.h"
+
+// Include required headers for OpenGL/Emscripten
+#if defined(MG_EMSCRIPTEN)
+#include <emscripten.h>
+#include <emscripten/html5.h>
+// Include OpenGLES 3.0 headers
+#include <GLES3/gl3.h>
+#else
+// #if defined(__APPLE__)
+// #include <OpenGL/gl.h>
+// #include <OpenGL/glext.h>
+// #else
+// #include <GL/gl.h>
+// #include <GL/glext.h>
+// #endif
+#endif
+
+#if defined(MG_SDL2)
+#define GL_GLEXT_PROTOTYPES 1
+#include <SDL.h>
+#include <SDL_opengl.h>
+#include <SDL_opengl_glext.h>
+#endif
+
+#include <vector>
+#include <string>
+#include <cstring>
+
+// Debug macros
+#ifdef DEBUG
+void GL_CHECK_ERROR() { \
+    GLenum err = glGetError(); \
+    if (err != GL_NO_ERROR) { \
+        fprintf(stderr, "OpenGL error %d at %s:%d\n", err, __FILE__, __LINE__); \
+    } \
+}
+#else
+#define GL_CHECK_ERROR() ((void)0)
+#endif
+// Structures for OpenGL graphics system
+
+struct MGG_GraphicsAdapter {
+    // OpenGL specific adapter info
+    char name[128] = { 0 };
+    char vendor[128] = { 0 };
+    char version[128] = { 0 };
+};
+
+struct MGG_GraphicsSystem {
+    std::vector<MGG_GraphicsAdapter*> adapters;
+};
+
+struct MGG_GraphicsDevice
+{
+    MGG_GraphicsSystem* system = nullptr;
+    MGG_GraphicsAdapter* adapter = nullptr;
+    
+#if defined(MG_EMSCRIPTEN)
+    EMSCRIPTEN_WEBGL_CONTEXT_HANDLE context = 0;
+#else
+    SDL_GLContext context = nullptr;
+    SDL_Window* window = nullptr;
+#endif
+
+    // Viewport
+    int viewportX = 0;
+    int viewportY = 0;
+    int viewportWidth = 0;
+    int viewportHeight = 0;
+    float viewportMinDepth = 0.0f;
+    float viewportMaxDepth = 1.0f;
+    
+    // Scissor rect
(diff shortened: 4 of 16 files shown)


native/monogame/vulkan/MGG_Vulkan.cpp
```diff
@@ -4331,7 +4331,7 @@ static MGG_Buffer* MGVK_Buffer_Create(MGG_GraphicsDevice* device, MGBufferType t
 	return buffer;
 }
 
-MGG_Buffer* MGG_Buffer_Create(MGG_GraphicsDevice* device, MGBufferType type, mgint sizeInBytes)
+MGG_Buffer* MGG_Buffer_Create(MGG_GraphicsDevice* device, MGBufferType type, mgint sizeInBytes, const char* name)
```
tomspilman: Note this is something we should be doing with most graphics resources as assigning names to them make it MUCH easier to debug using the different platform GPU debuggers.

The concern is extra garbage/performance hits from passing strings how we do at the moment.

dellis1972:
Tested , this currently is working on WebGL and Desktop on MacOS.

MonoGame.Framework/Game.cs
```diff
@@ -537,7 +537,7 @@ public void Tick()
                 // We only have a precision timer on Windows, so other platforms may still overshoot
 #if WINDOWS && !DESKTOPGL
                 MonoGame.Framework.Utilities.TimerHelper.SleepForNoMoreThan(sleepTime);
-#elif DESKTOPGL || ANDROID || IOS
+#elif DESKTOPGL || ANDROID || IOS || NATIVE
```
dellis1972, line 540: @tomspilman  I'm not sure if this is the right place for this condition? Should it be the one above? i.e should NATIVE follow the `SleepForNoMoreThan` path?

tomspilman: This is fine for now.

In the future we should probably use `SleepForNoMoreThan` helper on all platforms and the platform specific code into it.

native/monogame/vulkan/MGG_Vulkan.cpp
tomspilman: @dellis1972 

Is `name` unused here?  If so why was it added to the function?

native/monogame/opengl/MGG_OpenGL.cpp
```diff
@@ -0,0 +1,880 @@
...
+    return state;
+}
+
+void MGG_SamplerState_Destroy(MGG_GraphicsDevice* device, MGG_SamplerState* state) {
+    // No OpenGL resources to clean up
+    delete state;
+}
+
+MGG_Buffer* MGG_Buffer_Create(MGG_GraphicsDevice* device, MGBufferType type, mgint sizeInBytes, const char* name) {
+    printf("Creating buffer for OpenGL graphics device: %zu (type=%d, size=%d, name=%s)\n", (size_t)device->context, type, sizeInBytes, name);
```
tomspilman: I see you're only using the buffer name for logging here on the GL backend.

The complication is this breaks all the console platforms...  which will happen, but is it better to break if fully and add `const char* name` to all resource "create" functions here once.

@dellis1972 @harry-cpp ?

dellis1972: At this time we don't use it yes, But when we actually start implementing the Constant Buffer stuff we will need the name I think. That is how the current C# OpenGL stuff works. I think its specifically for constant buffers or Attribute locations. OpenGL needs a "string" name to look up the location. 

https://github.com/MonoGame/MonoGame/blob/a4d088ecd910e24958d3193762c7291c7bab2c24/MonoGame.Framework/Platform/Graphics/Shader/Shader.OpenGL.cs#L63

dellis1972: If there is a better way to do this, then we can remove it :D

dellis1972: In the code I am converting we use the `name` for the following call
```
GLint uniformLocation = glGetUniformLocation(program, buffer->name.c_str());
```
to get the uniform location.

dellis1972: See https://github.com/MonoGame/MonoGame/blob/a4d088ecd910e24958d3193762c7291c7bab2c24/MonoGame.Framework/Platform/Graphics/Shader/ConstantBuffer.OpenGL.cs#L51

tomspilman: Yeah lets remove this.   We can do this in other ways internal to the native code and shader compiler binary format.

native/monogame/vulkan/MGG_Vulkan.cpp
dellis1972: @tomspilman see the discussion at https://github.com/MonoGame/MonoGame/pull/9116#discussion_r2569585884

native/monogame/opengl/MGG_OpenGL.cpp
tomspilman: The native code backend takes this other path:
  https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Graphics/Shader/ConstantBufferCollection.cs#L74

It passes the stage and constant buffer index.  What should happen here is the native GL backend should look up the name from the internal shader data we generate from the shader compile step.

dellis1972: The alternative is to add a new native API call `MGG_Buffer_SetName` or something like that which is only implemented for specific platforms which need it.

tomspilman: > The alternative is to add a new native API call `MGG_Buffer_SetName`

No lets not do that.

dellis1972: ok, I removed this parameter

MonoGame.Framework/Platform/Native/VertexElement.Native.cs
```diff
@@ -20,6 +21,21 @@ internal MGG_InputElement AsInputElement(int vbSlot, int instanceFrequency)
         element.AlignedByteOffset = (uint)_offset;
         element.InstanceDataStepRate = (uint)instanceFrequency;
 
+        // Marshal string into fixed-size char array
+        if (!string.IsNullOrEmpty(name))
```
tomspilman: @dellis1972 Do you know what `name` looks like here in your tests?

Is name like POSITION, POSITION0, TEXCOORD, TEXCOORD0, TEXCOORD1, etc?

If the name from GL is always the semantic then we can derive the name from `element.SemanticUsage` and `element.SemanticIndex` and build the name within the native code.

dellis1972: I'll remove it for now. We can revisit this when we get around to implementing that part

MonoGame.Framework/Platform/Native/Graphics.Interop.cs
```diff
@@ -120,14 +121,15 @@ internal struct MGG_SamplerState_Info
...
-internal struct MGG_InputElement
+internal unsafe struct MGG_InputElement
 {
     public VertexElementUsage SemanticUsage;
     public uint SemanticIndex;
     public uint VertexBufferSlot;
     public VertexElementFormat Format;
     public uint AlignedByteOffset;
     public uint InstanceDataStepRate;
+    public fixed byte Name[32];
```
tomspilman: @dellis1972 if you don't need this in this current PR i recommend removing it.

I think what you need for GL can be derived from the existing input elements and or the shader data.

Making this structure unsafe and the function call using it unsafe and the usages of that function unsafe just for a member we possibly don't need in the end and don't need at all right now...  seems like it doesn't belong in this PR.

PR #9117, opened 2025-11-23
Refactor Game class and cleanup method descriptions

Ch3shireDev:
### Description of Change

I've cleaned up Game.cs file, fixing typos and improving Equals method to use declaration pattern in type test pattern. Also, my IDE removed additional spaces at the end of comments.


dellis1972:
We need to check that the new syntax is compatible with the console codebase (its either on .net 4.5 or netstandard 2.1 I can't remember which). So I've added the Blocked label for now.

SimonDarksideJ:
> We need to check that the new syntax is compatible with the console codebase (its either on .net 4.5 or netstandard 2.1 I can't remember which). So I've added the Blocked label for now.

Updating the branch following @harry-cpp PR to use Native for Console check and running the automation check

MonoGame.Framework/Game.cs
```diff
@@ -276,7 +276,7 @@ public TimeSpan TargetElapsedTime
 
         /// <summary>
         /// Indicates if this game is running with a fixed time between frames.
-        /// 
+        ///
```
mghoffmann, line 279: Can this use `<para>` and `<c>` instead of a blank line and `<code>`?

```
/// <summary>
/// Indicates if this game is running with a fixed time between frames.
/// <para>When set to <c>true</c> the target time between frames is
/// given by <see cref="TargetElapsedTime"/>.</para>
/// </summary>
```

 I see this in IntelliSense with the current version.
image

If `<para>` and `<c>` are used I see this (a little nicer in my opinion).
image

PR #9137, opened 2025-12-15
Add OpenGL ES 3.x support

CartBlanche:
Add OpenGL ES 3.x support, while we wait for Vulkan to mature and become more optimised on mobile devices.

As mentioned here - https://discord.com/channels/1176986388340686978/1176986388781076582/1450141953227489440 Mobile phones have supported OpenGL ES 3.x since 2013. We're 12 years behind the ball!! Also all the major game dev IDEs (Unity, Unreal Godot etc), all dropped ES 1 & 2. 

This should also close quite a few issues that have been asking for GetData support since 2014 or so.

iOS:
Simulator Screenshot - iPad Pro 13-inch (M5) (16GB) - iOS 26 2 - 2025-12-19 at 23 48 33

Android:
Screenshot_1766160104

Before this change, ShipGame would crash when loading the gameplay screen.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsExtensions.cs
+++ b/MonoGame.Framework/Graphics/GraphicsExtensions.cs
@@ -804,7 +804,7 @@ namespace Microsoft.Xna.Framework.Graphics
 
 #endif // OPENGL
 
-                    public static int GetSyncInterval(this PresentInterval interval)
+        public static int GetSyncInterval(this PresentInterval interval)
         {
             switch (interval)
             {
@@ -1060,7 +1060,7 @@ namespace Microsoft.Xna.Framework.Graphics
             }
         }
 #endif
-            }
+    }
 
     internal class MonoGameGLException : Exception
     {
--- a/MonoGame.Framework/Platform/Graphics/GraphicsCapabilities.OpenGL.cs
+++ b/MonoGame.Framework/Platform/Graphics/GraphicsCapabilities.OpenGL.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Xna.Framework.Graphics
             // Framebuffer objects
 #if GLES
             SupportsFramebufferObjectARB = GL.BoundApi == GL.RenderApi.ES && (device.glMajorVersion >= 2 || GL.HasExtension("GL_ARB_framebuffer_object")); // always supported on GLES 2.0+
-            SupportsFramebufferObjectEXT = GL.HasExtension("GL_EXT_framebuffer_object");;
+            SupportsFramebufferObjectEXT = GL.HasExtension("GL_EXT_framebuffer_object");
             SupportsFramebufferObjectIMG = GL.HasExtension("GL_IMG_multisampled_render_to_texture") |
                                                  GL.HasExtension("GL_APPLE_framebuffer_multisample") |
                                                  GL.HasExtension("GL_EXT_multisampled_render_to_texture") |
@@ -95,8 +95,9 @@ namespace Microsoft.Xna.Framework.Graphics
             }
             MaxTextureAnisotropy = anisotropy;
 
-            // sRGB
+            // sRGB and texture formats
 #if GLES
+            // sRGB, float textures, and half-float textures are core in OpenGL ES 3.0+
             SupportsSRgb = GL.HasExtension("GL_EXT_sRGB");
             SupportsFloatTextures = GL.BoundApi == GL.RenderApi.ES && (device.glMajorVersion >= 3 || GL.HasExtension("GL_EXT_color_buffer_float"));
             SupportsHalfFloatTextures = GL.BoundApi == GL.RenderApi.ES && (device.glMajorVersion >= 3 || GL.HasExtension("GL_EXT_color_buffer_half_float"));
@@ -104,8 +105,8 @@ namespace Microsoft.Xna.Framework.Graphics
 #else
             SupportsSRgb = GL.HasExtension("GL_EXT_texture_sRGB") && GL.HasExtension("GL_EXT_framebuffer_sRGB");
             SupportsFloatTextures = GL.BoundApi == GL.RenderApi.GL && (device.glMajorVersion >= 3 || GL.HasExtension("GL_ARB_texture_float"));
-            SupportsHalfFloatTextures = GL.BoundApi == GL.RenderApi.GL && (device.glMajorVersion >= 3 || GL.HasExtension("GL_ARB_half_float_pixel"));;
-            SupportsNormalized = GL.BoundApi == GL.RenderApi.GL && (device.glMajorVersion >= 3 || GL.HasExtension("GL_EXT_texture_norm16"));;
+            SupportsHalfFloatTextures = GL.BoundApi == GL.RenderApi.GL && (device.glMajorVersion >= 3 || GL.HasExtension("GL_ARB_half_float_pixel"));
+            SupportsNormalized = GL.BoundApi == GL.RenderApi.GL && (device.glMajorVersion >= 3 || GL.HasExtension("GL_EXT_texture_norm16"));
 #endif
 
             // TODO: Implement OpenGL support for texture arrays
@@ -122,8 +123,10 @@ namespace Microsoft.Xna.Framework.Graphics
 
 #if GLES
             SupportsSeparateBlendStates = false;
+            SupportsMapBuffer = GL.BoundApi == GL.RenderApi.ES && device.glMajorVersion >= 3;
 #else
             SupportsSeparateBlendStates = device.glMajorVersion >= 4 || GL.HasExtension("GL_ARB_draw_buffers_blend");
+            SupportsMapBuffer = device.glMajorVersion >= 3 || GL.HasExtension("GL_ARB_map_buffer_range");
 #endif
         }
 
--- a/MonoGame.Framework/Platform/Graphics/Texture3D.OpenGL.cs
+++ b/MonoGame.Framework/Platform/Graphics/Texture3D.OpenGL.cs
@@ -11,12 +11,14 @@ namespace Microsoft.Xna.Framework.Graphics
 {
     public partial class Texture3D : Texture
     {
+        private GraphicsDevice _graphicsDevice;
 
         private void PlatformConstruct(GraphicsDevice graphicsDevice, int width, int height, int depth, bool mipMap, SurfaceFormat format, bool renderTarget)
         {
-#if GLES
-            throw new NotSupportedException("OpenGL ES 2.0 doesn't support 3D textures.");
-#else
+            _graphicsDevice = graphicsDevice;
+            if (!_graphicsDevice.GraphicsCapabilities.SupportsMapBuffer)
+                throw new NotSupportedException("Texture3D is not supported on OpenGL ES versions below 3.0.");
+
             this.glTarget = TextureTarget.Texture3D;
 
             Threading.BlockOnUIThread(() =>
@@ -35,7 +37,6 @@ namespace Microsoft.Xna.Framework.Graphics
 
             if (mipMap)
                 throw new NotImplementedException("Texture3D does not yet support mipmaps.");
-#endif
         }
 
         private void PlatformSetData<T>(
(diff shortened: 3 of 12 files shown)


SimonDarksideJ:
Looks ok to me, but I think it needs another set of eyes on this one since it is removing older support.

Although slightly concerned with the sheer number of formatting updates here @CartBlanche , but guessing this file simply hasn't been touched in that much time.

CartBlanche:
@SimonDarksideJ As mentioned here - https://discord.com/channels/1176986388340686978/1176986388781076582/1450141953227489440 Mobile phones have supported OpenGL ES 3.x since 2013. We're 12 years behind the ball!! Also all the major game dev IDEs (Unity, Unreal Godot etc), all dropped ES 1 & 2.

dellis1972:
I am a bit concerned that we are removing support for platforms like GLES 2.x in an incremental release. 
One of the problems is the mobile release cadence, telling people to use old nuget packages for GLES 2.x is not always an option.

My take on this would we either detect the version of GLES we are loading and dynamically support those features like we do for Desktop, or leave this for when we finish off the native OpenGL api. 

We have the ability to detect the GLES version on Android via https://github.com/MonoGame/MonoGame/blob/develop/MonoGame.Framework/Platform/Graphics/OpenGL.Android.cs#L88

tomspilman:
> telling people to use old nuget packages for GLES 2.x is not always an option.

But they always have old source to build it if they need too as well with the latest compiler/.net?

If @CartBlanche is correct and Unity, Unreal, and Godot do not support GLES 1 & 2 anymore... we should seriously consider dropping it as well.  This is something that is holding back MG from having its full feature set on all platforms.

That said...

> My take on this would we either detect the version of GLES we are loading

If this can be done fairly easily that would be fine too.

ThomasFOG:
> Mobile phones have supported OpenGL ES 3.x since 2013

I'm a bit on the fence with this statement, especially considering the Android park. Android, despite supporting it for a while, is very notorious to have a _very lagging_ hardware park and it would seem that many new phones even today don't support that. And let's not talk about Vulkan, which despite being the official graphics API of Android by now, is still very largely unsupported (and the latest Android OS has a software renderer for Vulkan because almost no manufacturer has a driver for it).

One thing that I think is a _real_ strength of MonoGame, is that it has a low GL requirement and will virtually run on all mobiles. We need to be very cautious with API upgrades, because an innocent version bump can fence MonoGame from literally half the mobiles, even recent ones.

I think that an ES3 support should be built in a way that it is backward compatible with ES2 and features made optional.

tomspilman:
> I think that an ES3 support should be built in a way that it is backward compatible with ES2 and features made optional.

Agreed.... do you know how to approach that in this case @CartBlanche  ?

CartBlanche:
@tomspilman I assume I can reinstate the #if GLES blocks and at runtime do something like:
```
#if GLES
  if (GraphicsDevice.glMajorVersion >= 3)
  {
  ...
  }
  else
  {
  ...
  }
#endif
```

Or did you have something else in mind?

CartBlanche:
From Unity's docs:
Screenshot 2025-12-26 at 14 04 22

From Unreal's docs:
Screenshot 2025-12-26 at 14 05 49

tomspilman:
> I assume I can reinstate the #if GLES blocks and at runtime do something like:

Yeah that seems reasonable.


> From Unity's docs:
> From Unreal's docs:

I get it...  they likely did this after many years of supporting 2.x by default.

For now we should support both and maybe later in 2026 we can begin to depreciate it and disable it by default.

CartBlanche:
@tomspilman @ThomasFOG  @SimonDarksideJ @dellis1972 @harry-cpp  What I hadn't realised is that if we support both ES2 and ES3 we'd need to update the shader compiler to generate both profiles, then modify mgfxo to support both ES2 and ES3 shader formats, as well dynamically pick the right shader at runtime. By dropping ES2 we still need to update the shader compiles, but we don't need to cram in 2 formats into the mgfxo. Obviously correct me if I'm wrong here, as I'm new to this side of things.

ThomasFOG:
That would be quite a big hack because the whole pipeline isn't made to mix formats, that would be something specific to just GLES and would require a specific path separated from the non-ES path.

I really don't know if this is worth the effort, it sounds like a lot of overengineering for little gains with a non-zero risk of breaking the GL backend in some ways. I feel like the energy should be put into making the vulkan backend stable.

If the point is to be able to use GetData to handle collisions in the ShipGame, I'd reply that relying on vertex/index buffers GetData to handle collisions is a bad design in the first place and that a BoundingBox could be added to ModelMeshPart at load time (just like the BoundingSphere is).

---

Otherwise my biggest concern with this PR is the removal of the version checks on entry points loading. That's what I meant with optional: no entry point found? Let's flag the feature as not supported and continue normal operations.

tomspilman:
> What I hadn't realised is that if we support both ES2 and ES3 we'd need to update the shader compiler to generate both profiles

Isn't this just the `#version 330` text in the glsl text?  We can do that at runtime.

CartBlanche:
@ThomasFOG, @tomspilman I've re-added the GLES 2.x checks. So this should now continue to work on GLES2 as well as 3 now. 
> Isn't this just the #version 330 text in the glsl text? We can do that at runtime.

I don't know if that is all that is required for it to work, but for now the changes above should allow most, if not all, of our 3D games to work. Obviously more testing required :)

Once we drop ES 2, maybe we can look at adding MRT and other ES 3 abilities or if Vulkan is optimised and supports those things on mobile too, maybe use that.

MonoGame.Framework/Graphics/GraphicsExtensions.cs
```diff
@@ -511,12 +511,13 @@ internal static void GetGLFormat (this SurfaceFormat format,
...
             var supportsPvrtc = graphicsDevice.GraphicsCapabilities.SupportsPvrtc;
             var supportsEtc1 = graphicsDevice.GraphicsCapabilities.SupportsEtc1;
-            var supportsEtc2 = graphicsDevice.GraphicsCapabilities.SupportsEtc2;
+            var supportsEtc2 = graphicsDevice.GraphicsCapabilities.SupportsEtc2;
             var supportsAtitc = graphicsDevice.GraphicsCapabilities.SupportsAtitc;
             var supportsFloat = graphicsDevice.GraphicsCapabilities.SupportsFloatTextures;
             var supportsHalfFloat = graphicsDevice.GraphicsCapabilities.SupportsHalfFloatTextures;
             var supportsNormalized = graphicsDevice.GraphicsCapabilities.SupportsNormalized;
             var isGLES2 = GL.BoundApi == GL.RenderApi.ES && graphicsDevice.glMajorVersion == 2;
+            var isGLES3 = GL.BoundApi == GL.RenderApi.ES && graphicsDevice.glMajorVersion == 3;
```
dellis1972: This is never used?

MonoGame.Framework/Platform/Graphics/Vertices/VertexBuffer.OpenGL.cs
```diff
@@ -49,17 +49,14 @@ private void PlatformGetData<
         >(int offsetInBytes, T[] data, int startIndex, int elementCount, int vertexStride)
             where T : struct
         {
-#if GLES
             // Buffers are write-only on OpenGL ES 1.1 and 2.0.  See the GL_OES_mapbuffer extension for more information.
             // http://www.khronos.org/registry/gles/extensions/OES/OES_mapbuffer.txt
-            throw new NotSupportedException("Vertex buffers are write-only on OpenGL ES platforms");
-#else
+            if (GL.BoundApi == GL.RenderApi.ES && GraphicsDevice.glMajorVersion < 3)
+                throw new NotSupportedException("VertexBuffer.GetData is not supported on OpenGL ES versions below 3.0. Vertex buffers are write-only on those OpenGL ES platforms");
```
tomspilman, line 55: @CartBlanche to make this code a little simpler and faster maybe add `GraphicsCapabilities.SupportsMapBuffer` and use that in the `if` statements.

MonoGame.Framework/Platform/Graphics/Texture3D.OpenGL.cs
```diff
@@ -14,9 +14,9 @@ public partial class Texture3D : Texture
 
         private void PlatformConstruct(GraphicsDevice graphicsDevice, int width, int height, int depth, bool mipMap, SurfaceFormat format, bool renderTarget)
         {
-#if GLES
-            throw new NotSupportedException("OpenGL ES 2.0 doesn't support 3D textures.");
-#else
+            if (GL.BoundApi == GL.RenderApi.ES && graphicsDevice.glMajorVersion < 3)
```
tomspilman: Same here... maybe add `GraphicsCapabilities.SupportsTexture3D` to simplify this code.

MonoGame.Framework/Platform/Graphics/Vertices/VertexBuffer.OpenGL.cs
CartBlanche: @tomspilman  Hope I've dont it correctly.

MonoGame.Framework/Graphics/GraphicsExtensions.cs
CartBlanche: Removed.

CartBlanche:
Unit tests should now be green again.

dellis1972:
@CartBlanche there are a bunch of unrelated C++ changes in this PR now

CartBlanche:
@dellis1972 Seems a rebase fubarred! I'll try to fix it later.

CartBlanche:
@dellis1972 That looks fixed now. I just noticed what is CodeQL that is running against the PR (bottom thing)?
Screenshot 2026-01-22 at 15 12 55

IceReaper:
Just curious, if this adds GLES3 support for android and ios, would it be a problem to also provide a linux-arm64 DesktopGLES package? There are some devices who would benefit from this a lot, and avoid the need to use gl4es wrappers, which cause several issues when trying to run monogame on those devices.

MonoGame.Framework/Platform/Graphics/Texture3D.OpenGL.cs
```diff
@@ -69,14 +70,16 @@ private void PlatformSetData<T>(
...
 
         private void PlatformGetData<T>(int level, int left, int top, int right, int bottom, int front, int back, T[] data, int startIndex, int elementCount)
              where T : struct
         {
 #if GLES
-            throw new NotSupportedException("OpenGL ES 2.0 doesn't support 3D textures.");
+                // Note: glGetTexImage is not available in GLES, even in 3.0+
+                // We would need to use framebuffer + glReadPixels as a workaround
+                // For now, this will only work on desktop OpenGL
+                throw new NotSupportedException("Texture3D.GetData is not supported on OpenGL ES. Use desktop OpenGL instead.");
```
dellis1972: This error is pointless, if the user is targeting a mobile platform, how can they use Desktop OpenGL? 
Just say its not supported and leave it at that.

MonoGame.Framework/Platform/Graphics/GraphicsCapabilities.OpenGL.cs
```diff
@@ -125,8 +126,10 @@ private void PlatformInitialize(GraphicsDevice device)
 
 #if GLES
             SupportsSeparateBlendStates = false;
+            SupportsMapBuffer = GL.BoundApi == GL.RenderApi.ES && device.glMajorVersion >= 3;
 #else
             SupportsSeparateBlendStates = device.glMajorVersion >= 4 || GL.HasExtension("GL_ARB_draw_buffers_blend");
+            SupportsMapBuffer = true;
```
dellis1972: We can check the MajorVersion like above, and if an Extension is present (like the line above). You'll need to research which version and/or extension glMapBufferRange is provided in.

MonoGame.Framework/Platform/Graphics/Texture3D.OpenGL.cs
CartBlanche: @dellis1972 it wasn't telling them to switch to using DesktopGL on mobile, it was suggesting if they want to use Texture3D.GetData() they DesktopGL is where it is available. But I can see there might be some confusion there.

MonoGame.Framework/Platform/Graphics/GraphicsCapabilities.OpenGL.cs
CartBlanche: Fixed.

CartBlanche:
@IceReaper the ES 3 is purely for Android and iOS, right now. What devices/platforms require linux-arm64 and what is the market share of those platforms? We also accept PRs :)

PR #9140, opened 2025-12-21
[CI] Add packaging for the new native backend libraries

harry-cpp:
Stuff done:
- Renamed the `monogame.native` library to `mgruntime`
- Reworked the CI system to setup packing of all the new nugets for the native system

Nuget packaging for native is split between the following packages:
- `MonoGame.Runtime.System.X` packages which just contain a single native library that's gonna get copied to the output folder:
  - MonoGame.Runtime.Windows.DX12
  - MonoGame.Runtime.Windows.Vulkan
  - MonoGame.Runtime.Mac.Vulkan
  - MonoGame.Runtime.Linux.Vulkan
- `MonoGame.Framework` package which references the preferred runtime libraries for each platform as well as `MonoGame.Framework.Native` nuget package itself
  - this means the end user is supposed to just reference this one nuget package and not any of the above `MonoGame.Runtime.` packages, in the templates we will just have this nuget referenced
  - by default this nuget package references `Windows.DX12`, `Mac.Vulkan` and `Linux.Vulkan` as the desired set of runtimes
  - if the user wants to customize runtimes, ie. use vulkan on Windows, they can instead reference the `MonoGame.Framework.Native` and their desired set of runtime nuget packages manually

There are no custom .targets files or magical properties for any of this.


dellis1972:
@harry-cpp question, so in order to build a truly cross platform "Vulkan" app, users will need to manually update the references? Or will we be providing a cross platform template which already has that setup? 

Also what is the reason for the monogame.native -> mgruntime rename? Is there a technical reason for that ?

SimonDarksideJ (approved):

PR #9146, opened 2025-12-28
[Content.Pipeline] Improve the default logger

harry-cpp:
Stuff done:
- Split out the base class logger from the new content builder logger
- Hooked up the `LogLevel` parameter to the content logger
- Hooked up the `LoggerRootDirectory` so we got some proper pathing showing up
- Reworked the default appearance of the log thats:
  - Friendly for msbuild to parse the warnings / errors from
  - Friendly appearance for threaded logging if we ever get to that
  - Honestly even better for looking at for me :D, technically if anyone liked the previous one they can just assign the `ContentBuilder.Logger` property to a custom made logger

Bellow are some examples on appearance based on log level.

Debug:
debug

Info:
info

Warning:
warning

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
+++ b/MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
@@ -2,8 +2,7 @@
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
-using System.Diagnostics;
-using MonoGame.Framework.Content.Pipeline.Builder;
+using System.Globalization;
 
 namespace Microsoft.Xna.Framework.Content.Pipeline;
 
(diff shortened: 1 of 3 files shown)


MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
```diff
@@ -145,7 +81,7 @@ public void Log(string message)
     /// <param name="messageArgs">Arguments for the reported message.</param>
     [Obsolete("LogImportantMessage is deprecated, please use Log instead.")]
     public virtual void LogImportantMessage(string message, params object[] messageArgs)
-        => Log(LogLevel.Error, string.Format(message, messageArgs));
+        => Log(LogLevel.Error, string.Format(CultureInfo.InvariantCulture, message, messageArgs));
```
SimonDarksideJ, line 84: Any reason we are handling the `CultureInfo.InvariantCulture` in the overloads instead of the base `Log` method?  Because you rightfully added the handling of culture (because of all the warnings) but NOT for the new way the logging is meant to work.
Suggest just applying the format in the `Log` methods instead of the overloads

SimonDarksideJ (dismissed):
Good handling Culture string handling, but should not b in the overloads

MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
tomspilman: This is fixed now correct @harry-cpp @SimonDarksideJ ?

tomspilman (approved):

MonoGame.Framework.Content.Pipeline/ContentBuildLogger.cs
harry-cpp: > Suggest just applying the format in the Log methods instead of the overloads

The `Log` method does not accept message + messageArgs, there is nothing for the main Log method to handle.

tomspilman: @SimonDarksideJ can you clear this now and dismiss the change request so we can merge things?

SimonDarksideJ: Done, still think log should have Culture handling, since the ones that were changed are marked obsolete, but let's move on.

tomspilman: > still think log should have Culture handling,

The two log methods don't do any formatting of arguments:

image

Which is what the culture handing is doing... handling how to format arguments.

So what are you suggesting change there?

PR #9154, opened 2026-01-02
[WIP] Native Song Implementation 

tomspilman:
This is the initial implementation of Song for the native backend.

It includes a unit test as well.

Things left to do:

 - ~~Implement mp3 audio decoding.~~
 - ~~Fix name/priority of stream thread.~~
 - ~~Improve unit test to verify pause to play works.~~
 - ~~Test Song position setter/getter.~~
 - ~~Make better test song for unit tests.~~

Then we can can call this feature done.


dellis1972:
Do we need to add Licence information for third party libraries as they are not subodules?

tomspilman:
> Do we need to add Licence information for third party libraries as they are not subodules?

Not sure.  The headers of these single file .h libs includes the URL to the repo and copyright info.  I think that is enough.

ThomasFOG:
They are CC0, there's nothing to do with licenses.

SimonDarksideJ (approved):

PR #9167, opened 2026-01-13
Added nullable return types for #9166, added bool return for RemoveService

### LINKED ISSUE #9166: GameServiceContainer.GetService return types should be marked as nullable
### Intent

Currently, `GameServiceContainer.GetService<T>()` returns `T` and `GameServiceContainer.GetService(Type type)` returns `object`. Both overloads have cases where they return `null`. Using [nullable reference type](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references) annotation would let library users be aware of the null returns.

### Motivation

I keep having to inspect the compiled source to remember whether the methods return null or throw exceptions if a service of the requested type is not found. Marking the return values as nullable makes it clear that they return null if a service can't be found.

mghoffmann:
Fixes #9166

### Description of Change

- Added a `#nullable enable` section to GameServiceContainer.cs to mark the return types of the `GetService` overloads as nullable.
- Changed the return type of the `RemoveService` method to `bool` because the underlying `Dictionary<Type, object>.Remove` returns `bool` and knowing whether something was found and removed can be useful.

### DIFF
--- a/MonoGame.Framework/GameServiceContainer.cs
+++ b/MonoGame.Framework/GameServiceContainer.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Xna.Framework
         {
             services = new Dictionary<Type, object>();
         }
-
+#nullable enable
         /// <summary>
         /// Add a service provider to this container.
         /// </summary>
@@ -55,11 +55,11 @@ namespace Microsoft.Xna.Framework
         /// no suitable service provider is registered in this container.
         /// </returns>
         /// <exception cref="ArgumentNullException">If the specified type is <code>null</code>.</exception>
-        public object GetService(Type type)
+        public object? GetService(Type type)
         {
             if (type == null)
                 throw new ArgumentNullException("type");
-						
+
             object service;
             if (services.TryGetValue(type, out service))
                 return service;
@@ -72,12 +72,13 @@ namespace Microsoft.Xna.Framework
         /// </summary>
         /// <param name="type">The type of the service to remove.</param>
         /// <exception cref="ArgumentNullException">If the specified type is <code>null</code>.</exception>
-        public void RemoveService(Type type)
+        /// <returns><c>true</c> if a service of the indicated type was found and removed, otherwise <c>false</c>.</returns>
+        public bool RemoveService(Type type)
         {
             if (type == null)
                 throw new ArgumentNullException("type");
 
-            services.Remove(type);
+            return services.Remove(type);
         }
         
         /// <summary>
@@ -90,6 +91,9 @@ namespace Microsoft.Xna.Framework
         /// </exception>
         public void AddService<T>(T provider)
         {
+            if (provider == null)
+                throw new ArgumentNullException("provider");
+
             AddService(typeof(T), provider);
         }
 
@@ -101,7 +105,7 @@ namespace Microsoft.Xna.Framework
         /// A service provider of the specified type or <code>null</code> if
         /// no suitable service provider is registered in this container.
         /// </returns>
- 	public T GetService<T>() where T : class
+ 	public T? GetService<T>() where T : class
         {
             var service = GetService(typeof(T));
 
@@ -111,4 +115,5 @@ namespace Microsoft.Xna.Framework
             return (T)service;
         }
     }
+#nullable restore
 }


dellis1972:
Does changing the return type for a nullable break the existing API? 
Will a person upgrading get compiler or runtime errors if they do not change their code?

MonoGame.Framework/GameServiceContainer.cs
```diff
@@ -55,11 +55,11 @@ public void AddService(Type type, object provider)
         /// no suitable service provider is registered in this container.
         /// </returns>
         /// <exception cref="ArgumentNullException">If the specified type is <code>null</code>.</exception>
-        public object GetService(Type type)
+        public object? GetService(Type type)
```
tomspilman, line 58: I guess this doesn't break other code.

This class implements [IServiceProvider](https://learn.microsoft.com/en-us/dotnet/api/system.iserviceprovider?view=net-10.0) which itself is now defined as:

`public object? GetService(Type serviceType)`

Our code base calls `IServiceProvider.GetService` in lots of places:

https://github.com/search?q=repo%3AMonoGame%2FMonoGame%20GetService&type=code

Yet this all seems to compile.

So maybe this change is safe and doesn't break existing code?

@SimonDarksideJ @dellis1972 @CartBlanche ?

SimonDarksideJ: I would think so, adding a nullable return does not usually result in a breaking change, removing it however would.

So, as long as we agree this is a good move forward we can accept it, but there is no "going back" as that would be breaking.

Thoughts @tomspilman / @dellis1972 ?

SimonDarksideJ (approved):

MonoGame.Framework/GameServiceContainer.cs
mghoffmann: Correct, this will not break anything for existing callers unless they are also in a nullable context *and* don't check for null before using the return value, in which case they probably want an error about this. That was my use case.

SimonDarksideJ:
Reserved for `3.8.6` where we are addressing "nullables" on a wider scale

mghoffmann:
@SimonDarksideJ Is there a specific roadmap for nullable reference types? This is something I can help with and I would be happy to spend a couple of hours a week on it. (Or more if there were a bounty attached, unfortunately I have bills 😁)

Should I make an issue about it and push small PRs linked to the issue? I would follow [this guide](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies) to update incrementally but there are a few ways to proceed. My preference would be to enable nullable by default and add `#nullable disable` to the top of all existing files in one initial very big (but readable) PR, so all new code has nullable enabled without breaking existing code. And then remove `#nullable disable` file-by-file (or by group of files where it makes sense) for smaller PRs building up to the 3.8.6 release.

Enabling nullable warnings and then PR'ing fixes for those warnings in logical groups is workable too.

I'm happy to take ownership of this, I'm just not familiar with the planning process yet. What's the best way to contribute?

tomspilman:
> Is there a specific roadmap for nullable reference types?

Not really.

Our goal at the moment is to get 3.8.5 out stable as soon as possible.

That said there would be nothing wrong with you setting up some PRs with nullable fixes and begin preparing things for this change.  We just may wait a few more weeks until 3.8.5 is out before merging them.

mghoffmann:
Sounds good, I'll make a master issue about it sometime today and start pushing small PRs over the next few weeks.

SimonDarksideJ:
Small pr's, compete with unit tests and a single tackling issue would be best and easier to review and merge.

harry-cpp:
I noticed that you opened a couple PRs about the nullable types, and I am sorry in advance, but I will need to close them as they will conflict with the proper process of enabling it.

Now the process of enabling it properly will go as folllows:
 - First we enabled the property for all the `MonoGame.Framework.X.csproj` projects, fully project wide, this will cause a bunch of warnings for the main framework which will be fine
   - We really don't want `#nullable enable` scattered around all the .cs files
 - Next we slowly start fixing issues that are only in the `MonoGame.Framework.Native.csproj` project
   - Fixing warnings in this project will resolve any nullable public API facing code for the other platforms
   - There is no reason to fix any code for other platform projects because as its currently states, they will get replaced by the .Native one eventually
   - Unless absolutely necessary, no `!.` type of fixes

mghoffmann:
@harry-cpp  Thanks for looking at those PRs 🙂

Just to make sure I understand the preferred process:
1. One PR should add `<nullable>enable</nullable>` to all of the the `MonoGame.Framework.X.csproj` files.
2. Many small and readable PRs should address the new warnings that causes, in logical groups.
3. We won't worry about the warnings in the platform projects because they'll be replaced by the native project and they don't expose members to the public API.

Do I have all that right? I can get the null-enabling PR out this weekend and probably also some warning fixes.

>Unless absolutely necessary, no `!.` type of fixes
    
[Nullable.Extended](https://github.com/tom-englert/Nullable.Extended) is a simple code analyzer that adds errors when null forgiveness is used without a comment explaining the justification. Is it OK to add that analyzer as a dependency with the initial PR?

harry-cpp:
> Do I have all that right?

Yes, that is correct.

> [Nullable.Extended](https://github.com/tom-englert/Nullable.Extended) is a simple code analyzer

I'd say no, def don't want more dependencies. 

If everything is done proper, there should be little to no `!.` appearing at the end, and any that are needed will only be merged after all the other warnings are fixed (not just nullable warnings, we have a bunch of warnings coming from the project right now).

PR #9169, opened 2026-01-14
Add MGFXC_WINE_PATH to CI workflow

dellis1972:
Set environment variable for MonoGame path in CI workflow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
           - os: macos-26
             platform: macos
             shell: bash
-          - os: ubuntu-24.04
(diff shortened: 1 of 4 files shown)


AristurtleDev:
@dellis1972 @SimonDarksideJ I have submitted the following PR

https://github.com/MonoGame/monogame.github.io/pull/272

The issue appears to be that the wine setup scripts are adding an additional `\` in the shebang for the `wine_wrapper.sh` script.  When `WineHelper` falls back to using this, the script fails to execute because the `\` is treated as an esacpe sequence in the script.

Merging that PR and then re-running this should 🤞 fix the CI builds now on ubuntu

PR #9170, opened 2026-01-14
Fix Color.FromHSV hue progression and parameter normalization

### LINKED ISSUE #9131: Color.FromHSV() is missing components
```Color.FromHSV()``` is missing the hue progression and clamps values to their wheel segment.

https://github.com/MonoGame/MonoGame/blob/f2d35b696ebdbf031223f66f75237c9df990ded4/MonoGame.Framework/Color.cs#L2098

Correct conversion should go like something like this:

```csharp
            float w = (h % 60.0f) / 60.0f;

            switch (i)
            {
                case 0:
                    r = v;
                    g = t + w;
                    b = p;
                    break;
                case 1:
                    r = q - w;
                    g = v;
                    b = p;
                    break;
                case 2:
                    r = p;
                    g = v;
                    b = t + w;
                    break;
                case 3:
                    r = p;
                    g = q - w;
                    b = v;
                    break;
                case 4:
                    r = t + w;
                    g = p;
                    b = v;
                    break;
                default:
                    r = v;
                    g = p;
                    b = q - w;
                    break;
            }
```

AristurtleDev:
## Description

Fixes two bugs in `Color.FromHSV()`

1. Fixes fractional hue calculation within color wheel segments.  The fractional position `f` within each 60deg segment was calcualted as `int f = (int)(h / 60.0f) - i`, which always resulted in a `0` due to integer arithmetic.  Changed to `float f = (h % 60.0f) / 60.0f` to correctly calcualte fractional position within [0,1] range.
2. Removed incorrect normalization of saturation and value parameters.  The documentation for the method states that these values should be in the [0,1] range.  By normalizing them by dividing by 100, this causes colors to be nearly black in all cases (e.g. passing `s=1.0, v=1.0` would result in RGB values of `2` instead of `255`)

Additionally, `Color.FromHSV()`, `Color.FromHSL()` were changed to `static` methods. These are factory methods for creating a `Color` value from a different structure, it doesn't make sense to have them as instance members that require an instance before using.

## Breaking Changes

Since `Color.FromHSV()` and `Color.FromHSL()` are changed to static factory methods, this would be considered a breaking change as it changes the public API that already existed.  Not sure if you want to keep this change since it's technically a breaking change or we can obsolete the existing ones with a warning and forward them to the new static ones, then remove the instance methods in a future release where breaking changes are good.

## Related Issues
- Closes #9131

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -2037,7 +2037,7 @@ namespace Microsoft.Xna.Framework
         /// <param name="x"></param>
         /// <param name="rh"></param>
         /// <returns>the hue for R, G or B</returns>
-        private float HtoRGB(float c, float x, float rh)
+        private static float HtoRGB(float c, float x, float rh)
         {
             if ((6 * rh) < 1)
                 return (c + (x - c) * 6 * rh);
@@ -2056,7 +2056,7 @@ namespace Microsoft.Xna.Framework
         /// <param name="s">Saturation component value, from 0.0f to 100.0f</param>
         /// <param name="l">Luminosity (brightness) component value, from 0.0f to 100.0f</param>
         /// <returns><see cref="Color"/> with the HSL values</returns>
-        public Color FromHSL(float h, float s, float l)
+        public static Color FromHSL(float h, float s, float l)
         {
             s /= 100;
             l /= 100;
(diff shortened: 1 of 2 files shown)


ThomasFOG (approved):
Looking good to me, thanks! Breaking changes are indeed totally fine as this is very new and unrelated to XNA compatibility.

tomspilman (approved):

MrPodunkian:
Just chiming in here because I was using this code on something and noticed there's a typo in ToHL where the calculation of the blue segment of the color wheel's parenthesis are botched:

`h = (float)((60.0 * (((r - g) / delta)) + 4.0));`

Which multiplies the value by 60 before adding it to 4, rather than adding 4 an then multiplying the whole thing by 60. I know the code isn't directly related to the changes you were making @AristurtleDev, but figured you'd want to know!

I'm also wondering if FromHSV taking Saturation and Value values from 0.0 to 1.0 is maybe unintentional. All of the other methods deal with Saturation and Value/Luminosity values from 0.0 to 100.0 (including ToHSV), so users (such as myself) might get a little tangled up converting a color to and from HSV and getting unexpected results.

PR #9174, opened 2026-01-17
Update to Content Builder `.targets` to work for any build configuration

SimonDarksideJ:
# Summary

Templates update to enable the Content Builder `.targets` file to work for any build configuration.

Current update hard codes `net9.0` as this is the target framework of the Content project, which is needed for running the build executable.  (Cannot use `$(OutputPath)` as that is the target framework from the Source project, which could be Android/iOS)

```xml
<Project>
  <Target Name="BuildContent" BeforeTargets="BeforeCompile">
    <PropertyGroup>
      <ContentOutput>$(ProjectDir)$(OutputPath)</ContentOutput>
      <ContentTemp>$(ProjectDir)$(IntermediateOutputPath)</ContentTemp>
      <ContentArgs>build -p $(MonoGamePlatform) -s Content/Assets -o $(ContentOutput) -i $(ContentTemp)</ContentArgs>
      <ContentCommand>$(MSBuildThisFileDirectory)bin\$(Configuration)\net9.0\Content</ContentCommand>
    </PropertyGroup>
    <MSBuild Projects="$(MSBuildThisFileDirectory)Content.csproj" Targets="Build" RemoveProperties="TargetFramework;RuntimeIdentifier;RuntimeIdentifiers" />
    <Exec Command="$(ContentCommand) $(ContentArgs)" WorkingDirectory="$(MSBuildThisFileDirectory)..\" CustomErrorRegularExpression="\[E\] .+" CustomWarningRegularExpression="\[W\] .+" />
   </Target>
</Project>
```


Epicguru:
Thank you, I was running into this issue.  
While it is being investigated, is there a reason for the net9.0 to also be hardcoded there?  
I also ran into issues when publishing (publish path is different from normal build path, and can be set to an arbitrary location).

SimonDarksideJ:
> Thank you, I was running into this issue. While it is being investigated, is there a reason for the net9.0 to also be hardcoded there? 
Another foundation member also joust pointed that out with a better resolution, building now.

Instead using the following:

```xml
      <ContentCommand>$(MSBuildThisFileDirectory)$(OutputPath)\___SafeGameName___.Content</ContentCommand>
```

Utilising the project's "own" output path for the resolution instead of rebuilding it.  We've tested in some of our replication projects and this does seem to work better.

SimonDarksideJ:
Nope, scratch that, it does not work.  It is ok for base projects using `net9.0` or `net10.0`, but for others like Android / iOS which use a superset framework (`net9-ios`) which does not match the framework of the content project, it does not.

Might need to query the csproj of the Content project to make it dynamic

SimonDarksideJ:
Update migrated to #9176

PR #9175, opened 2026-01-17
Allow content builder importers to match more than the extension

Epicguru:
### Description of Change
I wanted to be able to automatically import files such as `Sprites.atlas.json` using a custom importer, but currently the only way to do this is to manually specify the importer: not all `.json` files will be sprite atlases, only the ones ending in `.atlas.json`.  
It is possible, in some cases, to give your files a custom file extension but that is annoying since it will no longer open in the expected programs, might not be displayed correctly in version control etc.

This PR changes the way that importers are selected to allow them to match the end of the file name rather than strictly the extension only. This allows `[ContentImporter(".atlas.json")]` to work as expected, whereas previously this was impossible.

This did also raise the question of how to handle the following kind of setup, which importer should be selected?
- A generic '.png' importer i.e. the built-in texture importer.
- A custom '.lightmap.png' importer.

To solve this I have also added a `Priority` property to the importer attribute.  
Importers with higher priority number are earlier in the list and will therefore match first. In the example above, you would give the lightmap importer higher priority.

The method `FindImporterByExtension(ext)` was left unchanged since this is public API. It will still strictly match on the extension alone.
I have added a `FindImporterByFileName(fn)` method instead that does the same thing but matching on the end of the file name, and the single internal usage of FindImporterByExtension has been replaced by this new method.

This change should not affect the behaviour of any existing importers or content setups.  
I have not included tests for this since there don't seem to be any existing tests for the importer selection system.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/PipelineBuilder/PipelineManager.cs
+++ b/MonoGame.Framework.Content.Pipeline/PipelineBuilder/PipelineManager.cs
@@ -354,26 +354,6 @@ namespace MonoGame.Framework.Content.Pipeline.Builder
         }
 
         /// <summary>
-        /// Returns the importer type name based on the file extension.
-        /// </summary>
-        /// <param name="ext">File extension to search for.</param>
-        /// <returns>Importer type name or <see langword="null"/> if not found.</returns>
-        public string FindImporterByExtension(string ext)
-        {
-            if (_importers == null)
-                ResolveAssemblies();
-
-            // Search for the importer.
-            foreach (var info in _importers)
-            {
-                if (info.attribute.FileExtensions.Any(e => e.Equals(ext, StringComparison.InvariantCultureIgnoreCase)))
-                    return info.type.Name;
-            }
-
-            return null;
-        }
-
-        /// <summary>
         /// Returns the importer type name based on the file name (including extension).
         /// </summary>
         /// <param name="fileNameWithExt">Then name of the file including the extension.</param>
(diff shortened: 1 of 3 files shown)


Epicguru:
I realize that this can also be done with wildcards, which may limit the need for this to exist,
although this does have the advantage of being able to drop in (or add a NuGet package for) custom importers without having to modify the build script.

harry-cpp (changes requested):
I see, good point, I agree on the addition of something for this, but I think we can do without the `Priority` property. 

Instead what we need is to make a list of:
```cs
List<(string extension, ImporterInfo importerInfo)> _importers;
```
that we sort based on extension length. Some attributes will have multiple extensions and they will be added multiple times for each.

This would also allow using something like `_asd.png` as a possible "extension".

Epicguru:
Copy, that makes more sense. I've updated it now to get rid of the explicit priority. I've added a new list of importer-extension pairs rather than updating the existing list since the sorted list is only needed in one place and this minimizes changes required.

MonoGame.Framework.Content.Pipeline/PipelineBuilder/PipelineManager.cs
```diff
@@ -428,7 +464,7 @@ public void ResolveImporterAndProcessor(string sourceFilepath, ref string import
         {
             // Resolve the importer name.
             if (string.IsNullOrEmpty(importerName))
-                importerName = FindImporterByExtension(Path.GetExtension(sourceFilepath));
+                importerName = FindImporterByFileName(Path.GetExtension(sourceFilepath));
```
harry-cpp, line 447: Delete the `FindImporterByExtension` function, this should be its only usage.

harry-cpp:
Glorious work @Epicguru, thank you very much, merging!

PR #9179, opened 2026-01-19
Added functionality to set the offset/position of a Song or SoundEffect(Instance)

Lucasharskamp:
### Description of Change

Allow the programmer to move to a specific position/offset (in TimeSpan format) of an AudioEffect(Instance) or Song to play from, in OpenAL/XAudio/MGA_*. (Or to receive the current position/offset position the player is at). 

Historically it's been noted that changing the offset/position of a audio while it's still playing causes runtime errors, so I've enforced that it cannot be changed **unless** it has been stopped beforehand. 

**OpenAL**
In OpenAL, I use alGetSourcef with AL_SEC_OFFSET to set the offset.  I've also removed a redundant if-statement, given no other value is possible once that code is reached. 

**XAudio**
In XAudio, I reset the buffer and assign a PlayBegin value according to the desired input. (This was recommended online)

**MGA (Native)**

Given the volatility of native code, given the many platforms it can run on, I decided to play it safe; stop the sound, reset the buffer, set the beginning value (similar to XAudio) and then hit play again.  Although the audio shouldn't be playing at that point, I decided to play it safe and presume it is somehow still playing.

### Contributor Declaration

**Note:** I have tested this on MonoGame DesktopGL, DirectX and Native, but I cannot verify it works as intended on Android, iOS or consoles. Also, myy C++ skills are rusty (as I was trained back in the day in C++03 and haven't really touched it since) so I do not know if the code I created is 100% memory-leak or process-issues safe. I would like someone to reproduce locally, to prevent potential issues, I've made it mandatory that one must stop the audio before changing the offset.

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/SoundEffectProcessor.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/SoundEffectProcessor.cs
@@ -24,10 +24,20 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
         public ConversionQuality Quality { get { return quality; } set { quality = value; } }
 
         /// <summary>
+        /// Currently assigned audio profile.
+        /// </summary>
+        private AudioProfile? _profile;
+
+        /// <summary>
         /// Initializes a new instance of SoundEffectProcessor.
         /// </summary>
         public SoundEffectProcessor()
-        {
+        { 
+        }
+
+        public SoundEffectProcessor(AudioProfile profile)
+        { 
+            _profile = profile;
         }
 
         /// <summary>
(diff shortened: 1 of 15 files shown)


Lucasharskamp:
Note; you can see it in action on the Discord, ''developer-diaries'' chat, I posted a video yesterday where the function is shown.

SimonDarksideJ:
https://discord.com/channels/355231098122272778/1365297601234079764/1462831092666073231

Lucasharskamp:
As requested, I've added a unittest. 
**Note**: Native doesn't work properly yet, given the MGA_Voice->format isn't initialized for AudioEffectinstances. @tomspilman All the values are left 0, causing the sound to be restarted instead.

tomspilman:
Ok i added some notes.

I don't see a problem with having a setter for position on a voice, but setting the position should not trigger playback like it is currently implemented.

This same concept should apply to all the implementations.

Lucasharskamp:
> Ok i added some notes.
> 
> I don't see a problem with having a setter for position on a voice, but setting the position should not trigger playback like it is currently implemented.
> 
> This same concept should apply to all the implementations.

Done.

Lucasharskamp:
@SimonDarksideJ I think I'm innocent on account of the MacOS build failing. It says it has to do with XCode and the .NET version.

SimonDarksideJ:
No worries @Lucasharskamp , It's a known thing.  The mac hosts get "weird" sometimes.

SimonDarksideJ:
@Lucasharskamp , are you able to pull from develop and resolve the conflicts so we can get this merged this week for `3.8.6`?

PR #9181, opened 2026-01-19
Added some bitwise operators to keyboard state

### LINKED ISSUE #9180: Adding bitwise operators to KeyboardState
### Intent

To be able to perform faster operations on a `KeyboardState`.

For example if I want to check if the A-Z keys are pressed in the current state and not pressed on the previous state I could first create a `KeyboardState` for comparison that just contains the A-Z keys. (This would only be created once).

```c#
var currentState = //.. Current captured keyboard state
var prevState = //.. Previous captured keyboard state
var emptyState = new KeyboardState(); //.. Empty state for empty comparison
var azKeys = new KeyboardState(new Keys[] {
    Keys.A,
    Keys.B,
    Keys.C,
    Keys.D,
    Keys.E,
    Keys.F,
    Keys.G,
    Keys.H,
    Keys.I,
    Keys.J,
    Keys.K,
    Keys.L,
    Keys.M,
    Keys.N,
    Keys.O,
    Keys.P,
    Keys.Q,
    Keys.R,
    Keys.S,
    Keys.T,
    Keys.U,
    Keys.V,
    Keys.W,
    Keys.X,
    Keys.Y,
    Keys.Z,
});
```

I can then use a bitwise AND to check if any of these keys are pressed.
```c#
if((azKeys & currentState) != emptyState)
{
// A a-z key was pressed
}

if((azKeys & currentState & ~prevState) != emptyState)
{
// A a-z key was pressed this frame
}
```

The empty state could be removed by adding another method to keyboard state to check if it's empty, such as `IsEmpty`.

### Motivation

There is some cases where I want quickly check if a specific set of keys are pressed on a keyboard state. This however requires calling `IsKeyDown` for each key. 

One example of this is I want to record the frame the keyboard was last active by checking if a large set of keys were pressed. Because I am doing this every frame I want it to be quick.

PR to follow.

arosolino:
Fixes #9180

The goal here is to provide quicker way of checking multiple key states by using bitwise operations.

I also am creating an internal constructor that would be very useful on a specific console.

- Should XOR be added for full support?
- Should a `IsEmpty` boolean be added

### DIFF
--- a/MonoGame.Framework/Input/KeyboardState.cs
+++ b/MonoGame.Framework/Input/KeyboardState.cs
@@ -18,10 +18,13 @@ namespace Microsoft.Xna.Framework.Input
         /// <summary>
         /// Returns a <see cref="KeyboardState"/> with no keys or modifiers set.
         /// </summary>
-        public static KeyboardState Empty => default;
+        public static KeyboardState Empty = default;
 
         // Used for the common situation where GetPressedKeys will return an empty array
         private static Keys[] empty = new Keys[0];
+
+        // Used to mask out modifier keys when checking for any key presses
+        private static KeyboardState modifiers = new KeyboardState(Keys.LeftShift, Keys.RightShift, Keys.LeftControl, Keys.RightControl, Keys.LeftAlt, Keys.RightAlt);
 
         #region Key Data
 
@@ -177,6 +180,71 @@ namespace Microsoft.Xna.Framework.Input
         public KeyState this[Keys key]
         {
             get { return InternalGetKey(key) ? KeyState.Down : KeyState.Up; }
+        }
+
+        /// <summary>
+        /// Gets whether any non-modifier key is currently pressed.
+        /// </summary>
+        /// <returns>true if any non-modifier key is pressed; false otherwise.</returns>
+        public bool AnyKeyDown()
+        {
+            return (_keys0 & ~modifiers._keys0) != 0
+                || (_keys1 & ~modifiers._keys1) != 0
+                || (_keys2 & ~modifiers._keys2) != 0
+                || (_keys3 & ~modifiers._keys3) != 0
+                || (_keys4 & ~modifiers._keys4) != 0
+                || (_keys5 & ~modifiers._keys5) != 0
+                || (_keys6 & ~modifiers._keys6) != 0
+                || (_keys7 & ~modifiers._keys7) != 0;
+        }
+
+        /// <summary>
+        /// Gets whether any key or modifier is currently pressed.
+        /// </summary>
+        /// <returns>true if any key or modifier is pressed; false otherwise.</returns>
+        public bool AnyKeyOrModifierDown()
+        {
+            return (_keys0 | _keys1 | _keys2 | _keys3 | _keys4 | _keys5 | _keys6 | _keys7) != 0;
+        }
+
+        /// <summary>
+        /// Gets whether any of the specified non-modifier keys are currently pressed.
+        /// </summary>
+        /// <param name="keys">The keys to query.</param>
+        /// <returns>true if any of the specified non-modifier keys are pressed; false otherwise.</returns>
+        public bool AnyKeyDown(KeyboardState keys)
+        {
+            return (this & keys).AnyKeyDown();
+        }
+
+        /// <summary>
+        /// Gets whether any of the specified keys or modifiers are currently pressed.
+        /// </summary>
+        /// <param name="keys">The keys to query.</param>
+        /// <returns>true if any of the specified keys or modifiers are pressed; false otherwise.</returns>
+        public bool AnyKeyOrModifierDown(KeyboardState keys)
+        {
+            return (this & keys).AnyKeyOrModifierDown();
+        }
+
+        /// <summary>
+        /// Gets whether any non-modifier key, excluding the specified keys, is currently pressed.
+        /// </summary>
+        /// <param name="keys">The keys to query.</param>
+        /// <returns>true if any non-modifier key, excluding the specified keys, is pressed; false otherwise.</returns>
+        public bool AnyKeyDownExcept(KeyboardState keys)
+        {
+            return (this & ~keys).AnyKeyDown();
+        }
+
+        /// <summary>
+        /// Gets whether any key or modifier, excluding the specified keys, is currently pressed.
+        /// </summary>
+        /// <param name="keys">The keys to query.</param>
+        /// <returns>true if any key or modifier, excluding the specified keys, is pressed; false otherwise.</returns>
+        public bool AnyKeyOrModifierDownExcept(KeyboardState keys)
+        {
+            return (this & ~keys).AnyKeyOrModifierDown();
         }
 
         /// <summary>
@@ -338,25 +406,6 @@ namespace Microsoft.Xna.Framework.Input
         public override bool Equals(object obj)
         {
             return obj is KeyboardState && this == (KeyboardState)obj;
-        }
-
-        /// <summary>
-        /// Compares whether current instance shares any keys or modifiers in the specified object.
-        /// </summary>
-        /// <param name="obj">The <see cref="KeyboardState"/> to compare.</param>
-        /// <returns>true if the provided <see cref="KeyboardState"/> instance shares any keys or modifiers as the current; false otherwise.</returns>
-        public bool Any(KeyboardState obj)
-        {
-            return
-                (_keys0 & obj._keys0) != 0
-                || (_keys1 & obj._keys1) != 0
-                || (_keys2 & obj._keys2) != 0
-                || (_keys3 & obj._keys3) != 0
-                || (_keys4 & obj._keys4) != 0
-                || (_keys5 & obj._keys5) != 0
-                || (_keys6 & obj._keys6) != 0
-                || (_keys7 & obj._keys7) != 0
-                || (_modifiers & obj._modifiers) != 0;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


tomspilman:
I'm good with these improvements.

Maybe add a `KeyboardState.Empty` for easy checks?  Some sort of easy `bool Any()` type method?

arosolino:
Added `KeyboardState.Empty` as it's super helpful for my case.

Question though... would you rather see it as (similar to how Vector2 does it)?
```
        public static KeyboardState Empty
        {
            get { return emptyState; }
        }
        private static readonly KeyboardState emptyState = new KeyboardState(0, 0, 0, 0, 0, 0, 0, 0, 0);
```

I like the idea of adding `bool Any(KeyboardState)` as it provides a early exit in the method. Here is how these changes cleaned up my current use case.

```
(largeKeyState & currentState & ~prevState) != KeyboardState.Empty
```
vs
```
(~prevState).Any(largeKeyState & currentState);
```

I think the original is a little easier to read the intent so I am not sure.

MonogameGuy:
Should the modifiers be included?  

If I use this as a guard pattern, say `A` is not pressed, it would be confusing when `shift-A` activates the blocked key.

Most users do not care about the shift, or any modifier, when querying the status of WASD.  

---

To prevent the ambiguity offer both modifier invariant and variant versions via methods instead of a single ambiguous operator.

tomspilman:
> Should the modifiers be included?

If we're improving this we should consider how modifiers are used.

MonoGame.Framework/Input/KeyboardState.cs
```diff
@@ -322,7 +340,88 @@ public override bool Equals(object obj)
             return obj is KeyboardState && this == (KeyboardState)obj;
         }
 
+        /// <summary>
+        /// Compares whether current instance shares any keys or modifiers in the specified object.
+        /// </summary>
+        /// <param name="obj">The <see cref="KeyboardState"/> to compare.</param>
+        /// <returns>true if the provided <see cref="KeyboardState"/> instance shares any keys or modifiers as the current; false otherwise.</returns>
+        public bool Any(KeyboardState obj)
```
TheKrumpet: I think a zero-param version of this which just returns if there's any pressed keys for the given `KeyboardState` might also be a good addition?

arosolino:
> Should the modifiers be included?
> 
> If I use this as a guard pattern, say `A` is not pressed, it would be confusing when `shift-A` activates the blocked key.
> 
> Most users do not care about the shift, or any modifier, when querying the status of WASD.
> 
> To prevent the ambiguity offer both modifier invariant and variant versions via methods instead of a single ambiguous operator.

Do you mean this case?
```
var aState = new KeyboardState(Keys.A);
if ((currentState & aState) == KeyboardState.Empty)
    return;

// Both A and Shift+A reach this point
```

If not can you provide an example? Thanks =)

arosolino:
This also excludes the NumLock & CapsLock modifiers, so now I am wondering if the binary operators should just exclude the modifiers.
```
        public static bool operator ==(KeyboardState a, KeyboardState b)
        {
            return a._keys0 == b._keys0
                && a._keys1 == b._keys1
                && a._keys2 == b._keys2
                && a._keys3 == b._keys3
                && a._keys4 == b._keys4
                && a._keys5 == b._keys5
                && a._keys6 == b._keys6
                && a._keys7 == b._keys7;
        }
```

MonogameGuy:
You have not explored all options.

The problem becomes apparent in the following code:

```
var ks = Keyboard.GetState();
// we will allow any key but A by the not operator:
var NotaState = ~ (new KeyboardState({Keys.A})); // Could include all A-Z

if (ks & NotaState)
    return true; //  Shift+A will reach this point
```

Excluding the modifiers, will exclude capital letters.  There is not a simple solution.

MonoGame.Framework/Input/KeyboardState.cs
MonogameGuy: The any key should be an all of the  private `_keys` OR'ed to 0.  There was a long running joke: I cannot find the any key.  It is well understood that modifier keys need not account for "any key".

arosolino:
Let's explore what has been discussed to see if there is room for improvement.

I am now internally storing a state of modifiers (Shift, Ctrl, Alt) on the struct for quick binary operations.

I have implemented the following methods. Names can be workshopped this is just for demonstrating.

`AnyKeyDown()`
- Gets whether any non-modifier key is currently pressed.

`AnyKeyOrModifierDown()`
- Gets whether any key or modifier is currently pressed.

`AnyKeyDown(KeyboardState keys)`
- Gets whether any of the specified non-modifier keys are currently pressed.

`AnyKeyOrModifierDown(KeyboardState keys)`
- Gets whether any of the specified keys or modifiers are currently pressed.

`AnyKeyDownExcept(KeyboardState keys)`
- Gets whether any non-modifier key, excluding the specified keys, is currently pressed.

`AnyKeyOrModifierDownExcept(KeyboardState keys)`
- Gets whether any key or modifier, excluding the specified keys, is currently pressed.

If we are going down this route then I do think users would also expect the following to be available.
`AnyKeyDown(params Key[] keys)`
`AnyKeyOrModifierDown(params Key[] keys)`
`AnyKeyDownExcept(params Key[] keys)`
`AnyKeyOrModifierDownExcept(params Key[] keys)`

Personally I think this is starting to stray from the original intent of this PR, which was just to expose some basic binary operators so I can do some larger key checks.

I would find `AnyKeyDown()` useful and if needed you could just exclude your own set of modifiers before calling it.

Finally if this is the direction we want to go then I would recommend renaming the `_modifiers` for CapsLock and NumLock to something like `_toggles` because its schematically confusing if we are mixing in these new methods that handle modifiers.

mghoffmann:
Can this also include an XOR operator please?

```csharp
KeyboardState jump = new KeyboardState(Keys.W);
KeyboardState current = Keyboard.GetState();
bool newlyPressed = (jump ^ current).AnyKeyDown() && current.IsKeyDown(Keys.W);
```

Supporting modifiers could maybe be more clean with overloads that take booleans instead of different methods:

```csharp
public bool AnyKeyDown(bool includeModifiers = true);
public bool AnyKeyDown(KeyboardState keys, bool includeModifiers = true);
public bool AnyKeyDownExcept(KeyboardState keys, bool includeModifiers = true);
```

>If we are going down this route then I do think users would also expect the following to be available.

I think supporting params overloads would encourage users to lean on those overloads every frame, which would add the unnecessary overhead of constructing a new `KeyboardState` from the params array. Personally I would prefer to be forced into using `AnyKeyDown(KeyboardStateIConstructedAndMustIntentionallyManage)` than to have a params overload available.

These are good changes, thanks for working on this 😃

arosolino:
I agree 100% with what you are suggesting here. I will make these changes when I get back from my trip in a couple days.

arosolino:
Made the following changes.

- Added XOR operator
- Made including the modifiers as a parameter option
- Renamed CapsLock & NumLock to be referenced as toggles internally so it's not to be confused with the modifiers

I am on the fence of whether or not we should even provide a `includeModifiers` option. If a user wanted to do this they could just maintain their own list of modifiers and exclude it themselves before calling the `AnyKeyDown`. I personally don't have a use case for this and if I did I might want more control of what exactly is a modifier (maybe Fn, Win).

mghoffmann:
> If a user wanted to do this they could just maintain their own list of modifiers and exclude it themselves before calling the AnyKeyDown

Yeah this does make sense. Now that I think about it more I think it's best to let the consumer determine what is considered a modifier. Right now all of the modifier keys have their own enum values in `Keys`, and although it's weird a game could use Caps Lock for a standalone input. Lots of games use Left Shift.

If anything the overload could take an optional collection of keys to exclude. But that's overkill when it's pretty easy to XOR two `KeyboardState`s to pass into `AnyKeyDown(KeyboardState keys)`.

TL;DR My humble opinion is that no special overload is needed for modifier keys because they're just regular keys in a lot of games.

arosolino:
I also think if modifiers are being improved then it should probably be an Enum similar to how SDL does it. I have several areas in my code where I want to check if either Left Shift or Right Shift is pressed.
https://wiki.libsdl.org/SDL3/SDL_Keymod

I think this PR should solely focus on bitwise operations and another PR could focus on improving modifiers if that needs to be explored more.

Let me know if we should split this PR up.

mghoffmann:
I agree it would be nice to be able to build a modifiers flag with their own enum. But that brings it back to the question of what counts as a modifier.

What about just defining some static properties on KeyboardState that could be OR'd together, right next to `Empty`?

```cs
public static readonly KeyboardState LeftShift = new(Keys.LeftShift);
public static readonly KeyboardState RightShift = new(Keys.RightShift);
// Etc. for all of the common modifier keys
````

```cs
private readonly KeyBoardState sprintKeys = KeyboardState.LeftShift | KeyboardState.RightShift;

public override void Update(GameTime gameTime)
{
    bool sprint = KeyboardState.AnyKeyDown(shiftModifiers);
```

These new bitwise operators make it possible to treat any key or combination of keys as a modifier without having to verbosely check each individual Keys value every frame. So a new flags enum just for some keys doesn't seem necessary.

And for what it's worth I'm definitely not an authority around here, just a guy who's been using MonoGame and XNA on and off for years and just recently started contributing, so seeing what others have to say is probably good 😄

SimonDarksideJ:
Any further updates to this @arosolino ?

arosolino:
> Any further updates to this @arosolino ?

Sorry I been busy with porting so not able to get to this at the moment.

I think we should just simplify it to not have modifier logic for now because determining what is a modifier is a little beyond the scope of this PR.

Or we can just close this PR for now and come back to it later if someone else needs it.

PR #9191, opened 2026-01-22
Implements #9190: IVertexType should declare VertexDeclaration as a static property

### LINKED ISSUE #9190: IVertexType should declare VertexDeclaration as a static property
### Intent

Currently, the `IVertexType` interface defines `VertexDeclaration` as a get-only instance property.

https://github.com/MonoGame/MonoGame/blob/3f4423b7414a7f3347959e843d7c7c29281dde2f/MonoGame.Framework/Graphics/Vertices/IVertexType.cs#L11-L14

C# 11 (with .NET 7) introduced [static abstract members in interfaces](https://devblogs.microsoft.com/dotnet/welcome-to-csharp-11/#abstracting-over-static-members). This makes more sense for `IVertexType.VertexDeclaration` because the vertex declaration does not change with each instance of the type; it should be static with the vertex type. And in fact each of the built-in implementations of `IVertexType` already implements the interface with an explicit instance property that just wraps a readonly static property.

```csharp
    /// <summary>
    /// Vertex type interface which is implemented by a custom vertex type structure.
    /// </summary>
    public interface IVertexType
    {
        /// <summary>
        /// Vertex declaration, which defines per-vertex data.
        /// </summary>
        static abstract VertexDeclaration VertexDeclaration
        {
            get;
        }
    }
```

https://github.com/MonoGame/MonoGame/blob/3f4423b7414a7f3347959e843d7c7c29281dde2f/MonoGame.Framework/Graphics/Vertices/VertexPositionColor.cs#L24-L25

https://github.com/MonoGame/MonoGame/blob/3f4423b7414a7f3347959e843d7c7c29281dde2f/MonoGame.Framework/Graphics/Vertices/VertexPositionColor.cs#L38-L44

Will change to

```csharp
 public static VertexDeclaration VertexDeclaration { get; }
```

### Motivation

Library users will no longer need to define `VertexDeclaration` as an instance property, just to wrap a static `VertexDeclaration` property that does not change per-instance.

mghoffmann:
Fixes #9190 

### Description of Change

This changes [`IVertexType.VertexDeclaration`](https://github.com/MonoGame/MonoGame/blob/3f4423b7414a7f3347959e843d7c7c29281dde2f/MonoGame.Framework/Graphics/Vertices/IVertexType.cs#L11) to be a [static abstract property](https://devblogs.microsoft.com/dotnet/welcome-to-csharp-11/#abstracting-over-static-members) instead of an instance property, to align with how vertex declarations are actually constructed and used in `IVertexType` implementations.

This breaks the XNA API so it should be reserved for a future release.

### Contributor Declaration


mghoffmann:
This will not build until the target framework is raised (static abstract members of interfaces are a C# 11 feature not supported by nestandard2.1).

MonoGame.Framework/Graphics/Vertices/VertexDeclaration.cs
```diff
@@ -202,30 +206,28 @@ private static int GetVertexStride(VertexElement[] elements)
...
-			}
+            if (!typeof(IVertexType).IsAssignableFrom(vertexType))
+                throw new ArgumentException($"{nameof(vertexType)} does not implement {nameof(IVertexType)}.");
 
-            var vertexDeclaration = type.VertexDeclaration;
-			if (vertexDeclaration == null)
-			{
-				throw new Exception("VertexDeclaration cannot be null");
-			}
+            var map = vertexType.GetInterfaceMap(vertexTypeInterface);
```
mghoffmann: This reflection dependency may break NativeAoT, I'm not sure. If it does this PR might need changes to be compatible with #8194.

mghoffmann:
I'm pretty confident I've fixed the AOT breaking issue based on my research, but less confident in my own ability to test it, so feedback is welcome.

PR #9193, opened 2026-01-22
Nullable context for GraphicsDevice

### LINKED ISSUE #9192: Introduce nullable aware contexts for nullable reference types
### Intent

### Motivation

mghoffmann:
Fixes #9192 

### Description of Change
This adds `#nullable enable` to the top of `GraphicsDevice.cs` and the platform implementations partial classes, and fixes all warnings that result from doing so.

- The warnings in `GraphicsDevice.cs` were mostly around referencing properties that were not marked as nullable and were not assigned in the constructors. Each of the constructors used the `Setup` and `Initialize` methods to assign values to those properties, so now those properties are backed by nullable fields that are initialized instead. The properties return the backing fields with the null-forgiving operator, which should correctly result in `NullReferenceException` if somehow the backing fields are not initialized before the properties are accessed.

- Warnings in the platform implementation files were mostly about uninitialized private and internal fields, almost all of which were initialized similarly to the properties in `GraphicsDevice.cs`. These fields were changed to nullable with null forgiveness used in every context where they should be expected to have been initialized.

- Public properties where a default value was indicated in the documentation comments now assign that default value to their backing field and return it, only if the backing field is null.

- Fields that were initialized in `Setup` to be clones of static graphics resources now use field initializers so the compiler can guarantee the fields are non-null when they're used as null-coalescing assignment values for the backing values of properties with default values.

- Methods and properties that return null in any cases now have a nullable return type in the signature. This should not break any existing callers until they are also in a nullable context, and then only if they use a returned nullable as if it's guaranteed to not be null.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Graphics/GraphicsDevice.cs
+++ b/MonoGame.Framework/Graphics/GraphicsDevice.cs
@@ -81,6 +81,7 @@ namespace Microsoft.Xna.Framework.Graphics
         private Rectangle _scissorRectangle;
         private bool _scissorRectangleDirty;
 
+        // This is set to a non-null value in Initialize, so null forgiveness is safe after that
         private VertexBufferBindings? _vertexBuffers;
         private bool _vertexBuffersDirty;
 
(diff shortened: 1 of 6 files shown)


harry-cpp:

PR #9197, opened 2026-01-23
Nullable context for Game and GameServiceContainer

### LINKED ISSUE #9192: Introduce nullable aware contexts for nullable reference types

mghoffmann:
Fixes #9192 

### Description of Change

This adds nullable contexts to the `Game` class and the `GameServiceContainer` class (which is exclusively used by the `Game` class).

- There were some private fields that were assigned to null. They are now marked nullable.
- Calls to `Services.GameService` now use the preferred generic overload, which is also now appropriately marked as nullable.
- Added some pattern matching and null propagation to avoid duplicate dereferences and casts in null checks.
- The internal `AddJournalEntry<T>` and `SortingFilteringCollection<T>` types now have a `notnull` constraint on `T`, to prevent future changes from attempting to pass null values where they're not handled.
- Event handlers for events of type `EventHandler<EventArgs>` now have the `sender` parameter nullable to match the delegate signature. Null forgiveness was used where `sender` is expected to never be null.
- Events have a default empty delegate handler to prevent invocation of null events.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Game.cs
+++ b/MonoGame.Framework/Game.cs
@@ -2,6 +2,8 @@
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


harry-cpp:

PR #9207, opened 2026-01-31
#9192: Nullable contexts for Graphics namespace (part 1)

### LINKED ISSUE #9192: Introduce nullable aware contexts for nullable reference types

mghoffmann:
Fixes #9192 

### Description of Change

This adds `#nullable enable` to 20 classes in the Graphics namespace. Most of the classes already handle null values appropriately so many files only have the pragma added and nothing else changed, and most of the rest only have nullable annotation changes, like null propagation instead of if blocks, and marking variables and returns that are already treated as nullable by callers.

### Contributor Declaration


harry-cpp:

PR #9214, opened 2026-02-05
Test switching back to macos-latest

SimonDarksideJ:
# Summary

Macos-26 is now throwing errors on use, trialing automation with macos-latest to see if the mac estate is now updated


SimonDarksideJ:
Losey test, macos-latest is still macos-15, no good for MonoGame

PR #9225, opened 2026-02-18
make GamePadButtons._buttons public

envoidia:
### Description of Change

Make GamePadButtons._buttons public, rename to Buttons, and add documentation.

The internal access of this field is inconvenient in some cases. For example, in order to merge multiple GamePadButtons together, you must check each property individually. With the field public, you can simply or them together.

As far as I can tell, there is no compelling reason for this field to remain internal.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Input/GamePadButtons.cs
+++ b/MonoGame.Framework/Input/GamePadButtons.cs
@@ -9,7 +9,10 @@ namespace Microsoft.Xna.Framework.Input
     /// </summary>
     public struct GamePadButtons
     {
-        internal readonly Buttons _buttons;
+        /// <summary>
+        /// A value representing all currently pressed buttons
+        /// </summary>
+        public readonly Buttons Buttons;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


MonoGame.Framework/Input/GamePadButtons.cs
```diff
@@ -9,7 +9,10 @@ namespace Microsoft.Xna.Framework.Input
     /// </summary>
     public struct GamePadButtons
     {
-        internal readonly Buttons _buttons;
+        /// <summary>
+        /// A value representing all currently pressed buttons
+        /// </summary>
+        public readonly Buttons Buttons;
```
SimonDarksideJ, line 15: TO simplify this change and reduce the impact, rather than altering the existing `internal` property, simply add a new readonly accessor that points to the internal property.

Aim is to limit change where possible, and private/internal backers with a public accessor is the preferred standandard.

SimonDarksideJ (changes requested):

PR #9226, opened 2026-02-19
added missing documentation (CS1591 errors) and CA1834 (use char, not string) warnings in the project.

Lucasharskamp:
**Fixes** 

- Improved documentation for Effect constructor where a thrown Exception was not mentioned and parameters not filled in. Also added some information regarding two parameters which make it easy for programmers to make mistakes they can get stuck on, thinking they need to set the index according to the header of the mgfx file, which the method does for them. 
- Added documentation for SensorBase (Given I work with IoT devices for my dialy work, I felt I could fill this in)
- Fixed missing documentation warnings.
- Removed "Count" from ShaderStage given it was not in use.
- Fixed stringbuilder warnings where a string the size of a single char was added, rather than just a single char. 

  
### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Color.cs
+++ b/MonoGame.Framework/Color.cs
@@ -1,4 +1,4 @@
-// MIT License - Copyright (C) The Mono.Xna Team
+﻿// MIT License - Copyright (C) The Mono.Xna Team
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
@@ -1867,7 +1867,7 @@ namespace Microsoft.Xna.Framework
             sb.Append(B);
             sb.Append(" A:");
             sb.Append(A);
-            sb.Append("}");
+            sb.Append('}');
             return sb.ToString();
         }
 
--- a/MonoGame.Framework/Devices/Sensors/Accelerometer.cs
+++ b/MonoGame.Framework/Devices/Sensors/Accelerometer.cs
@@ -1,4 +1,4 @@
-// MonoGame - Copyright (C) MonoGame Foundation, Inc
+﻿// MonoGame - Copyright (C) MonoGame Foundation, Inc
 // This file is subject to the terms and conditions defined in
 // file 'LICENSE.txt', which is part of this source code package.
 
@@ -68,6 +68,7 @@ namespace MonoGame.Framework.Devices.Sensors
             PlatformStop();
         }
 
+        /// <inheritdoc />
         protected override void Dispose(bool disposing)
         {
             PlatformDispose(disposing);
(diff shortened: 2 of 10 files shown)


MonoGame.Framework/Color.cs
```diff
@@ -1867,7 +1867,7 @@ public override string ToString()
             sb.Append(B);
             sb.Append(" A:");
             sb.Append(A);
-            sb.Append("}");
+            sb.Append('}');
```
SimonDarksideJ, line 1870: What is the purpose of switching from double quotes to single quotes.

If valid, they all instances should change within each class affected

SimonDarksideJ (commented):
Just a query on the quotes change, the docs updates seem fine

MonoGame.Framework/Color.cs
Lucasharskamp: It's a cleanliness problem. If you're appending a string the size of a single char, it's slower than a adding a single char directly. Sure, in this case it's trivial, but the warning is more meant for cases where for example you go through a for-loop of n instances.

All instances btw were noted in the build log as warnings, and have been resolved in this PR.

See more here: [CA1834 documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1834)

MonoGame.Framework/Graphics/Shader/ShaderStage.cs
```diff
@@ -4,11 +4,18 @@
...
+    /// </summary>
     public enum ShaderStage
     {
+        /// <summary>
+        /// Vertex related shader resources, such as vertex shader and vertex buffers.
+        /// </summary>
         Vertex,
-        Pixel,
-
-        Count,
```
SimonDarksideJ, line 12: Thoughts on this @tomspilman , should we be removing values from here, are there impacts on private reports from this change?

MonoGame.Framework/Devices/Sensors/Accelerometer.cs
```diff
@@ -1,4 +1,4 @@
-// MonoGame - Copyright (C) MonoGame Foundation, Inc
+﻿// MonoGame - Copyright (C) MonoGame Foundation, Inc
```
tomspilman, line 1: What is the difference in this line?  It isn't whitespace that i can see.

MonoGame.Framework/Graphics/Shader/ShaderStage.cs
tomspilman: At one point we may have been using the Count for some loop in the code base...  but i guess we're not now.

Seems safe to remove.

MonoGame.Framework/Devices/Sensors/Accelerometer.cs
ghost: Line ending changed (\n vs \r\n)

SimonDarksideJ (approved):

PR #9230, opened 2026-02-20
Add readonly public accessor for GamePadButtons._buttons

envoidia:
(I accidentally closed my last PR about this, so I had to make a new one. Sorry about that.)

The internal access of this field is inconvenient in some cases. For example, in order to merge multiple GamePadButtons together, you must check each property individually. With the public accessor, you can simply or them together.

### DIFF
--- a/MonoGame.Framework/Input/GamePadButtons.cs
+++ b/MonoGame.Framework/Input/GamePadButtons.cs
@@ -9,6 +9,11 @@ namespace Microsoft.Xna.Framework.Input
     /// </summary>
     public struct GamePadButtons
     {
+        /// <summary>
+        /// A value representing all currently pressed buttons
+        /// </summary>
+        public readonly Buttons Buttons => _buttons;
+
         internal readonly Buttons _buttons;
 
         /// <summary>
@@ -196,7 +201,7 @@ namespace Microsoft.Xna.Framework.Input
         /// </summary>
         /// <returns>A hash code for this instance that is suitable for use in hashing algorithms and data structures such as a
         /// hash table.</returns>
-        public override int GetHashCode ()
+        public override int GetHashCode()
         {
             return (int)_buttons;
         }


MonoGame.Framework/Input/GamePadButtons.cs
```diff
@@ -157,46 +162,46 @@ internal GamePadButtons(params Buttons[] buttons) : this()
...
+        }
+
+        /// <summary>
+        /// Returns a value indicating whether this instance is equal to a specified object.
+        /// </summary>
+        /// <param name="obj">An object to compare to this instance.</param>
+        /// <returns>true if <paramref name="obj"/> is a <see cref="GamePadButtons"/> and has the same value as this instance; otherwise, false.</returns>
+        public override bool Equals(object obj)
+        {
+            return (obj is GamePadButtons) && (this == (GamePadButtons)obj);
```
mghoffmann, line 196: This can use [pattern matching](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching) to avoid an unnecessary cast.

```cs
public override bool Equals(object obj) => obj is GamePadButtons gpb && this == gpb;
```

I think in this case the runtime performance is the same but it's fewer parentheses which I personally prefer :)

MonoGame.Framework/Input/GamePadButtons.cs
```diff
@@ -157,46 +162,46 @@ internal GamePadButtons(params Buttons[] buttons) : this()
...
-            return (obj is GamePadButtons) && (this == (GamePadButtons)obj);
+        }
+
+        /// <summary>
+        /// Determines whether two specified instances of <see cref="GamePadButtons"/> are equal.
+        /// </summary>
+        /// <param name="left">The first object to compare.</param>
+        /// <param name="right">The second object to compare.</param>
+        /// <returns>true if <paramref name="left"/> and <paramref name="right"/> are equal; otherwise, false.</returns>
+        public static bool operator ==(GamePadButtons left, GamePadButtons right)
```
mghoffmann, line 173: [Binary operators](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#operator-overloadability) for this struct would be nice if you're feeling generous.

MonoGame.Framework/Input/GamePadButtons.cs
envoidia: Isn't reformatting the existing XNA code out of scope for this PR? (I didn't touch that part of the code, git just thinks I did for some reason.)

mghoffmann: Ah, I thought you changed it to add the equality operators. The different syntax wouldn't change the signature at all, just readability and compile efficiency. But probably out of scope if the equality operators aren't actually new.

MonoGame.Framework/Input/GamePadButtons.cs
mghoffmann: If you didn't add these operators then maybe binary operators are best left for a different feature request. But it also feels germane to this PR because the use cases overlap:

```cs
readonly GamePadButtons JumpButtons = new(Buttons.A, Buttons.X);

bool IsJumpPressed(GamePadButtons currentButtons)
    => (currentButtons & JumpButtons) != GamePadButtons.None;
```
Is a lot more convenient than checking each property individually.

This would also need to be added I guess, and/or a `bool Any` instance property.
```cs
/// <summary>
/// A read-only <see cref="GamePadButtons"/> value with no buttons pressed.
/// </summary>
public static readonly GamePadButtons None { get; } = new();
```

envoidia: With the accessor, you can achieve this effect with `currentButtons.Buttons & JumpButtons.Buttons`. Additionally, with extension operators, the user could add these operators in an extension if desired.

I'm not against adding them; it's just a niche feature, and I try to tread lightly when working with legacy code.

But if you still think I should and that it fits this PR, I can do it.

mghoffmann: I'm just some guy on the Internet giving my 2 cents, not a maintainer or anything 🙂 

You're right that it's a niche feature, and I forgot that extension operators are possible now, we live in exciting times 😎
It's up to you I suppose.

ThomasFOG:
Merging GamePadButtons looks like an edge case and I'm not sure if it desirable to bloat the API with a "Buttons.Buttons" for an edge case. This struct is meant to be immutable by design.

What would motivate you to merge such structs?

envoidia:
> What would motivate you to merge such structs?

In a singleplayer game, you generally want any attached controller to work for input. You could individually check every controller, but it makes for nicer code if you can just fuse them together. There are probably other use cases, but that's the first one I thought of.

I've found that this whole API is really weird and uncomfortable to work with. GamePadButtons is just a Buttons but more awkward to use, and GamePadDPad has no reason to be separate from it. Direct Buttons access should make it feel a bit less like the API is an active obstacle. At least that's the idea.

Note that the struct is still immutable after this change.

arosolino:
I also found this piece of code interesting.
https://github.com/MonoGame/MonoGame/blob/a5a068f5d3522b470d6b393b1408ce4c57c155ce/MonoGame.Framework/Input/GamePadState.cs#L113

Everytime `IsButtonDown` and `IsButtonUp` is called it's merging the `GamePadState` into a new `Buttons`.

Maybe the solution to this problem is being able to Convert a `GamePadState` to `Buttons`.

Then @envoidia you could work with that directly.

arosolino:
> I also found this piece of code interesting.
> 
> https://github.com/MonoGame/MonoGame/blob/a5a068f5d3522b470d6b393b1408ce4c57c155ce/MonoGame.Framework/Input/GamePadState.cs#L113
> 
> Everytime `IsButtonDown` and `IsButtonUp` is called it's merging the `GamePadState` into a new `Buttons`.
> 
> Maybe the solution to this problem is being able to Convert a `GamePadState` to `Buttons`.
> 
> Then @envoidia you could work with that directly.

Sorry this doesn't solve your problem of wanting to access `_buttons`.. but you could also merge them using binary operators similar to the PR I made about KeyboardState https://github.com/MonoGame/MonoGame/issues/9180.

arosolino:
Sorry for the comments my brain is running circles. But since you want to merge multiple gamepads maybe we just expose binary operators on the `GamepadState` as a whole so you can merge two `GamepadState` together.

If we did this in my PR for `KeyboardState` it should be done for `GamepadState` as well.

ThomasFOG:
> In a singleplayer game, you generally want any attached controller to work for input. You could individually check every controller, but it makes for nicer code if you can just fuse them together. There are probably other use cases, but that's the first one I thought of.

I might not have the full picture but as stated I personally think that you're trying to overengineer a problem that doesn't exist. I would recommend to just poll PlayerIndex.One and call it a day if you're making a solo game. You'll be fine likely forever with nobody complaining.

envoidia:
> Maybe the solution to this problem is being able to Convert a `GamePadState` to `Buttons`.

> But since you want to merge multiple gamepads maybe we just expose binary operators on the GamepadState as a whole so you can merge two GamepadState together.

Yeah, these are both good ideas. I just picked GamePadButtons because it was the main obstacle and the smallest change.

envoidia:
> I would recommend to just poll PlayerIndex.One and call it a day if you're making a solo game. You'll be fine likely forever with nobody complaining.

Yeah, maybe. I do sometimes have multiple controllers connected to my computer (for local multiplayer games), and it would be annoying if I opened a singleplayer game while I happened to have multiple controllers connected and I had to figure out which one is the one that works. But it's entirely possible that this would only impact me specifically.

When I set out to do this, I figured that it would be an extremely simple thing to do, so I could just throw it in quickly without it being a big deal or something I'd think about for more than 30 seconds. But because it ended up being more complicated than I anticipated, it does start to feel like a lot of effort for such a niche issue.

SimonDarksideJ (approved):

PR #9231, opened 2026-02-20
ContentManager nullable annotations for #9192

mghoffmann:
Relates to #9192 
Depends on #9227 

### Description of Change

This adds nullable annotations, null pattern matching, and documentation changes to the ContentManager class, and some similar changes to parts of ContentReader and ContentTypeReader that are touched by the ContentManager changes.

This looks big but it's mostly `?` insertions where potentially-null variables were already being handled gracefully, and changing `== null` or `!= null` to `is null` or `is not null`.

### Contributor Declaration


SimonDarksideJ:
@tomspilman is this going to have an impact on Native/Console work? as I believed we were avoiding "nullability" due to consolecheck

PR #9236, opened 2026-02-28
Fix issue #9235(make SamplerState.Clone public)

### LINKED ISSUE #9235: SamplerState.Clone should be public instead of internal
### Intent

Currently SamplerState's Clone method is internal. Clone calls a constructor which takes an existing SamplerState, but the constructor is private so it cannot be used either.

Clone should be made public.

### Motivation

SamplerState provides a number of default types like:

* SamplerState.LinearClamp
* SamplerState.LinearWrap
* SamplerState.PointClamp
* SamplerState.PointWrap
* ... and so on

These are good starting points but if you need more control, such as to control mipmap behavior, then these defaults cannot be used as-is.

Ideally one could do something like this:

```cs
var clone = SamplerState.PointClamp.Clone();
clone.MipMapLevelOfDetailBias = -1;
```

However, this is not possible, so instead one must do this:

```cs

var samplerState = SamplerState.PointClamp;
var clone = new SamplerState();

// Copy all the values, be sure to not miss any!
clone.AddressU = samplerState.AddressU;
clone.AddressV = samplerState.AddressV;
clone.AddressW = samplerState.AddressW;
clone.BorderColor = samplerState.BorderColor;
clone.Filter = samplerState.Filter;
clone.MaxAnisotropy = samplerState.MaxAnisotropy;
clone.MaxMipLevel = samplerState.MaxMipLevel;
clone.MipMapLevelOfDetailBias = samplerState.MipMapLevelOfDetailBias;
clone.ComparisonFunction = samplerState.ComparisonFunction;
clone.FilterMode = samplerState.FilterMode;

// apply the change you want
clone.MipMapLevelOfDetailBias = -1;
```

Of course, I'll probably create a utility function for cloning, but users shouldn't have to do this since it already exists in MonoGame itself.

incomplete-tree:
Fixes #9235 

### Description of Change

Make SamplerState.Clone() public in order to allow people to change only a few vars of a samplerstate, instead of having to copy everything

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Graphics/States/BlendState.cs
+++ b/MonoGame.Framework/Graphics/States/BlendState.cs
@@ -444,7 +444,7 @@ namespace Microsoft.Xna.Framework.Graphics
             Opaque = new BlendState("BlendState.Opaque", Blend.One, Blend.Zero);
 		}
 
-	    internal BlendState Clone()
+	    public BlendState Clone()
 	    {
 	        return new BlendState(this);
 	    }
--- a/MonoGame.Framework/Graphics/States/DepthStencilState.cs
+++ b/MonoGame.Framework/Graphics/States/DepthStencilState.cs
@@ -524,7 +524,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		    None = new DepthStencilState("DepthStencilState.None", false, false);
 		}
 
-        internal DepthStencilState Clone()
+        public DepthStencilState Clone()
         {
             return new DepthStencilState(this);
         }
--- a/MonoGame.Framework/Graphics/States/RasterizerState.cs
+++ b/MonoGame.Framework/Graphics/States/RasterizerState.cs
@@ -236,7 +236,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		    CullNone = new RasterizerState("RasterizerState.CullNone", CullMode.None);
 		}
 
-	    internal RasterizerState Clone()
+	    public RasterizerState Clone()
 	    {
 	        return new RasterizerState(this);
 	    }
--- a/MonoGame.Framework/Graphics/States/SamplerState.cs
+++ b/MonoGame.Framework/Graphics/States/SamplerState.cs
@@ -318,7 +318,7 @@ namespace Microsoft.Xna.Framework.Graphics
             _filterMode = cloneSource._filterMode;
         }
 
-        internal SamplerState Clone()
+        public SamplerState Clone()
         {
             return new SamplerState(this);
         }
--- a/MonoGame.Framework/Graphics/States/TargetBlendState.cs
+++ b/MonoGame.Framework/Graphics/States/TargetBlendState.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Xna.Framework.Graphics
             ColorWriteChannels = ColorWriteChannels.All;
         }
 
-	    internal TargetBlendState Clone(BlendState parent)
+	    public TargetBlendState Clone(BlendState parent)
 	    {
 	        return new TargetBlendState(parent)
 	        {
@@ -302,4 +302,3 @@ namespace Microsoft.Xna.Framework.Graphics
 
 	}
 }
-


mghoffmann:
Is there a reason not to make the private constructor public along with `Clone`?

```cs
private SamplerState(SamplerState cloneSource)
```

incomplete-tree:
> Is there a reason not to make the private constructor public along with `Clone`?
> 
> ```cs
> private SamplerState(SamplerState cloneSource)
> ```

The constructor is only used to clone the samplerState, as is evident looking at the parameter name `cloneSource`. Only one of the two is needed: either make the constructor public and delete the Clone() method, or keep it private. 2 seperate methods doing the same seems weird to me.

dellis1972:
@tomspilman this looks ok to me. A few questions. 1. Is this a good addition to the API (it wasn't in XNA) seems like so to me.. 2. are there any other places where a `.Clone()` method would make sense?

tomspilman:
> 1. Is this a good addition to the API (it wasn't in XNA) seems like so to me.
 
Yeah i think it is fine.

> 2. are there any other places where a .Clone() method would make sense

The other graphics states i would think... blend, raster, and depth-stencil.

dellis1972:
> > 1. Is this a good addition to the API (it wasn't in XNA) seems like so to me.
> 
> Yeah i think it is fine.
> 
> > 2. are there any other places where a .Clone() method would make sense
> 
> The other graphics states i would think... blend, raster, and depth-stencil.

@joukepouke  do you think you could add the .Clone() method to the states tom mentioned above?

incomplete-tree:
Yes, I will

SimonDarksideJ:
Any update @incomplete-tree ?

incomplete-tree:
Oh sorry I completely forgot 😞 
I will make it tonight.

SimonDarksideJ:
Sorry to be a complete pain @incomplete-tree , but could you recreate and re-submit this PR clean

There are a LOT of unrelated formatting changes which are clogging up the PR and making it very dirty.  Happens.

Just make sure to only commit files you KNOW you have changes and ensure the git diff is clean.  As it stands we cannot merge this PR due to these issues

uygary:
@SimonDarksideJ do you think it would make sense to run a linter/formatter based on `.editorconfig` just once throughout the codebase?
It should ensure no one has any of these issues going forward.

SimonDarksideJ:
It's usually not a problem, but it comes up from time to time with everyones different setup and we've tried several improvements to reduce the noise. 
But everyone has their own setup, sorry about this.

uygary:
No, that's all fair. But if we run a single pass of formatter across the codebase and make one big fat PR that solely focuses on line endings in all the cs files, since the editorconfig is already in place, it should prevent this from happening for the most part.

And I'm not only talking about folks accidentally submitting changes in irrelevant files.
In my view, it's almost as important to not force those folks to go over the changes and say "oh, let me see which ones I've actually touched myself, and discard the rest."
Because people make mistakes when it comes to that. I know I do... And then go ahead and make the changes I've just discarded by mistake again.

I mean it happens rarely enough, because it also depends on an earlier mistake of having the knee-jerk reaction of clicking "yes, format the file," but it does happen.

And I just think making sure every C# file in the existing codebase adheres to the standard we already set in the editorconfig is the easiest way of ensuring that.
I do understand that going over such a PR makes this a suggestion that makes us worry. But I still think it's worth the effort of doing it once. Plus, I think we can even automate that by making sure the diffs in that changeset (only for that PR) actually contain only line ending and/or whitespace changes.

Just my two cents/tuppence.

incomplete-tree:
Yeah, I'm sorry. I looked at the files the first time, but i guess I forgot by accident. Maybe a formatter/linter commit hook would prevent it? I will revert the unintended changes.

AristurtleDev (approved):
I'll give my approval.

@SimonDarksideJ  i don't think a new PR is necessary here. There are a lot of places in the MonoGame code base where line endings are not normalized, which make it increadibly hard for anyone to submit a PR without some for of auto formatting or line normalization getting in the way.  It's probably better if we let some of these normalization in so they are not happening in the future.

If you click the cog wheel on the Files Changed view and tell it ti hide the whitespace only changes, it cleans up the view 

Image

Doing that on this PR, you can see the only changes are making the `internal` `Clone` methods `public`, no other actual code changes are made

PR #9252, opened 2026-03-06
WindowsDX12 doesn't properly exit when closing the Window.

arosolino:
### Description of Change
When the window is closed and either `EventType.WindowClose` or `EventType.Quit` is fired it does not properly exit the game. It increments `_isExiting` instead of actually calling the proper `Game.Exit()` method.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Platform/Native/GamePlatform.Native.cs
+++ b/MonoGame.Framework/Platform/Native/GamePlatform.Native.cs
@@ -90,7 +90,7 @@ class NativeGamePlatform : GamePlatform
             switch (event_.Type)
             {
                 case EventType.Quit:
-                    _isExiting++;
+                    Game.Exit();
                     break;
 
                 case EventType.WindowGainedFocus:
@@ -113,7 +113,7 @@ class NativeGamePlatform : GamePlatform
                 { 
                     var window = NativeGameWindow.FromHandle(event_.Window.Window);
                     if (Window == window)
-                        _isExiting++;
+                            Game.Exit();
                     break;
                 }
 
--- a/MonoGame.Framework/Platform/Native/GamePlatform.Native.cs
+++ b/MonoGame.Framework/Platform/Native/GamePlatform.Native.cs
@@ -113,7 +113,7 @@ class NativeGamePlatform : GamePlatform
                 { 
                     var window = NativeGameWindow.FromHandle(event_.Window.Window);
                     if (Window == window)
-                            Game.Exit();
+                        Game.Exit();
                     break;
                 }
 


MonoGame.Framework/Platform/Native/GamePlatform.Native.cs
```diff
@@ -113,7 +113,7 @@ private unsafe void PollEvents()
                 { 
                     var window = NativeGameWindow.FromHandle(event_.Window.Window);
                     if (Window == window)
-                        _isExiting++;
+                            Game.Exit();
```
uygary:
```suggestion
                        Game.Exit();
```

arosolino: Whoops fixed.

dellis1972:
@tomspilman  how would this change impact the behaviour of other native platforms like Vulkan and consoles?

//cc @arosolino

SimonDarksideJ:
This needs further testing as it breaks the standard GamePlatform exiting pattern used by other platforms.  Calling Game.Exit directly is invalid in this case.

arosolino:
This is what the current version of DesktopGL is doing

https://github.com/MonoGame/MonoGame/blob/698442da96f4f542b4807d48331997851626eb69/MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs#L125
https://github.com/MonoGame/MonoGame/blob/698442da96f4f542b4807d48331997851626eb69/MonoGame.Framework/Platform/SDL/SDLGamePlatform.cs#L243

uygary:
Looks to me @arosolino is right, this is the existing behaviour, not a change.
More importantly, this also appears to be how it's designed to work in the native backend as well, and omitting the call into Game.Exit is just an oversight. This change is simply fixing a bug:

1. Calling `Game.Exit()` here [sets _shouldExit = true](https://github.com/MonoGame/MonoGame/blob/698442da96f4f542b4807d48331997851626eb69/MonoGame.Framework/Game.cs#L396) in `Game`.
2. `Game` in turn [calls Platform.Exit()](https://github.com/MonoGame/MonoGame/blob/698442da96f4f542b4807d48331997851626eb69/MonoGame.Framework/Game.cs#L621) in `Tick`, unless someone's subscribed to the `Game.Exiting` event and they set `exitingEventArgs.Cancel`.
3. Then `Exit` here in `NativeGamePlatform` safely [increments _isExiting](https://github.com/MonoGame/MonoGame/blob/698442da96f4f542b4807d48331997851626eb69/MonoGame.Framework/Platform/Native/GamePlatform.Native.cs#L64).
4. The rest of the flow is unchanged.

@SimonDarksideJ: Calling `Game.Exit()`, we simply allow the `Game` instance itself to gracefully shutdown (like we currently do in non-native platforms), raising the [Exiting](https://github.com/MonoGame/MonoGame/blob/698442da96f4f542b4807d48331997851626eb69/MonoGame.Framework/Game.cs#L738) event.

SimonDarksideJ (approved):

PR #9254, opened 2026-03-10
Add SetNativeTexture and FromSharedHandle for zero-copy texture sharing

mfkl:
### Description of Change

Enable external code to wrap an already-opened D3D11 shared texture as a MonoGame Texture2D, avoiding a per-frame CopyResource when interoperating with other D3D11 devices (e.g. LibVLC).

I don't know if this is a change you're interested to take it, and if so if other targets (opengl, vulkan) could be candidate for it as well, just wanted to share in case.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Platform/Graphics/Texture.DirectX.cs
+++ b/MonoGame.Framework/Platform/Graphics/Texture.DirectX.cs
@@ -44,6 +44,13 @@ namespace Microsoft.Xna.Framework.Graphics
             return _resourceView;
         }
 
+        internal void SetNativeTexture(Resource texture)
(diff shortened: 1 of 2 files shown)


dellis1972:
@tomspilman  my guess would be we would need this implemented for all platforms? 

@mfkl  you forgot to check the LLm declaration

tomspilman:
> @tomspilman my guess would be we would need this implemented for all platforms?

We could live with this and add a task to the 3.8.6 release to make this a more formal API for all platforms.

SimonDarksideJ (approved):

PR #9258, opened 2026-03-18
Fix documentation for EffectProcessorDebugMode.Auto

bjornenalfa:
### Description of Change

There is no code that looks for `EffectProcessorDebugMode.Auto` meaning that the current documentation is incorrect.
In order to save others the time of looking for how to control the Auto behaviour I would like to make this change to clarify that there is no way to enable debugging shaders outside of using `EffectProcessorDebugMode.Debug`.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework.Content.Pipeline/Processors/EffectProcessorDebugMode.cs
+++ b/MonoGame.Framework.Content.Pipeline/Processors/EffectProcessorDebugMode.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Xna.Framework.Content.Pipeline.Processors
     public enum EffectProcessorDebugMode
     {
         /// <summary>
-        /// Enables effect debugging when built with Debug profile.
+        /// Will use the Optimize behaviour, disabling debugging and producing optimized shaders.
         /// </summary>
         Auto = 0,
 


tomspilman:
So `EffectProcessorDebugMode.Auto` is a byproduct of how the XNA pipeline worked.  Since content builds were triggered when compiling your game code... it would pick Debug shaders automatically when building Debug C# code.

For MonoGame `EffectProcessorDebugMode.Auto` is more complex.  Because you don't necessarily build your content triggered from VS.

I think for now the comment should be:

```
/// Enables effect debugging when built with Debug profile. (Not supported)
```

In the future because of how MG allows you to use different ways to building your content... this enum option will always be hit or miss.  So maybe we shouldn't support it?  In that case maybe it should be:


```
[Obsolete("Not supported in MonoGame")]
/// Enables effect debugging when built with Debug profile.  (Not supported).
```

@dellis1972 @SimonDarksideJ @harry-cpp ?

dellis1972:
I'm ok with Obsoleting this property/type

tomspilman:
> I'm ok with Obsoleting this property/type

To be clear we would not be obsoleting this Enum or the property that uses it.  It serves a good purpose.

The issue is specifically the enum value `EffectProcessorDebugMode.Auto`  which assumes that the pipeline knows if you're building your game code for release or debug.  We don't always know that.

SimonDarksideJ:
Are you ok to update the comment Using Tom's suggestion above @bjornenalfa ?

SimonDarksideJ:
Any update @bjornenalfa . else this PR risks going stale and will need to be closed.

SimonDarksideJ:
Closing as stale and no further responses from author

PR #9261, opened 2026-03-18
Native SDL Gamepads are improperly mapping from instance ids

arosolino:
There is a bug when a gamepad get's added it doesn't use the proper instance id.

The `SDL_ControllerDeviceEvent.which` returns a joystick device index for `SDL_CONTROLLERDEVICEADDED` and an instance id for `SDL_CONTROLLERDEVICEREMOVED`
https://wiki.libsdl.org/SDL2/SDL_ControllerDeviceEvent

The `SDL_ControllerButtonEvent.which` returns an instance id
https://wiki.libsdl.org/SDL2/SDL_ControllerButtonEvent

### Contributor Declaration

### DIFF
--- a/native/monogame/sdl/MGP_sdl.cpp
+++ b/native/monogame/sdl/MGP_sdl.cpp
@@ -425,10 +425,11 @@ mgbyte MGP_Platform_PollEvent(MGP_Platform* platform, MGP_Event& event_)
             auto controller = SDL_GameControllerOpen(ev.cdevice.which);
             if (controller != nullptr)
             {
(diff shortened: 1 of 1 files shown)


arosolino:
I just realized I way over complicated this and didn't look properly at `GamePad.Native.cs` and how it's using a dictionary there to map device ids.

I will fix it.

arosolino:
Alright fix is in just had to make sure when a gamepad is added it gets the actual instance id.

SimonDarksideJ (approved):

PR #9262, opened 2026-03-19
Centralise SoundEffect shutdown and resetting the _systemState to SoundSystemState.NotInitialized

### LINKED ISSUE #9026: [Android] New Version of OpenAL (1.24.x) - Crash after restarting android app
### Prerequisites

### MonoGame Version

Monogame 3.8.5

MonoGame Android Application (mgandroid)

### Operating System

Win 10

### Description

I hope I can put this in the right words.

Google Play forces 16kb pages until Nov 1st 2025 and this basically forces the use of OpenAL 1.24.x (which is referenced in monogame 3.8.5 dev branch) - but also works in 3.8.4 by manually adding package reference to newer OpenAL (makes no difference for this error)

But there seems to be a lifecycle problem/change with openAL, so whenever the Activity is "finished" and restarted, the app will crash due to a missing audio device in OpenAL. This is the error (when calling SoundEffect.Play after a soft-restart of the app):

`Microsoft.Xna.Framework.Audio.NoAudioHardwareException: 'OpenAL context has failed to initialize. Call SoundEffect.Initialize() before sound operation to get more specific errors.'`

What I could figure out is, that openAL is now always "destroying" the context it lives in but something is not triggered in monogame to reinitialize the sound device (gfuess monogame handles soft-resets as such while openAL isnt any more).

### Steps to Reproduce

Steps to reproduce should be as follows (using OpenAL 1.24.x or monogame 3.8.5 dev branch):

- Have an Monogame/Android App that plays some SoundEffect.
- Call Activity->Finish().
- Now restart the app on the device which plays the SoundEffect, the above error should happen.
- I was able to have it trigger debugger, by starting app with VS2022 on the device, calling Activity.Finish() and manually restart the app on the device - the debugger will trigger on SoundEffect.Play with above error.

(the important thing is a soft-reset in android)

The problem can be ommited by calling 

```
this.FinishAffinity();
Java.Lang.JavaSystem.Exit(0);
```

instead of

`this.Finish()`

in order to fully dispose the activity on exiting or in "OnBackPressed()", so it is some sort of lifecycle issue. Calling SoundEffect.Initialize does not help, because it looks like monogame thinks it is already initialized or something like that.

### Minimal Example Repo

_No response_

### Expected Behavior

Well, I guess there need to be some adaptation to the lifecycle changes in openAL or other means in monogame to detect if OpenAL has fully disposed its device.

### Resulting Behavior

`Microsoft.Xna.Framework.Audio.NoAudioHardwareException: 'OpenAL context has failed to initialize. Call SoundEffect.Initialize() before sound operation to get more specific errors.'`

When playing a SoundEffect after an Activity "Soft-Reset" (calling Activity.Finish() or using the back button to close the app)

I tried calling SoundEffect.Initialize() manually but it did not solve the issue, I have not verified sources, but it looks like Initialize thinks openAL is still initialized)

### Files

_No response_

CartBlanche:
Fixes #9026

### Description of Change

Based on OP's suggestion, but setting:
`_systemState = SoundSystemState.NotInitialized`

In a centralised location, so that all soundeffects follow this pattern, regardless of platform.

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Audio/SoundEffect.cs
+++ b/MonoGame.Framework/Audio/SoundEffect.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Xna.Framework.Audio
         #region Internal Audio Data
 
         private string _name = string.Empty;
-        
+
         private bool _isDisposed = false;
         private readonly TimeSpan _duration;
 
@@ -424,8 +424,8 @@ namespace Microsoft.Xna.Framework.Audio
         /// <para>Each SoundEffectInstance has its own Volume property that is independent to SoundEffect.MasterVolume. During playback SoundEffectInstance.Volume is multiplied by SoundEffect.MasterVolume.</para>
         /// <para>This property is used to adjust the volume on all current and newly created SoundEffectInstances. The volume of an individual SoundEffectInstance can be adjusted on its own.</para>
         /// </remarks>
-        public static float MasterVolume 
-        { 
+        public static float MasterVolume
+        {
             get { return _masterVolume; }
             set
             {
@@ -434,7 +434,7 @@ namespace Microsoft.Xna.Framework.Audio
 
                 if (_masterVolume == value)
                     return;
-                
+
                 _masterVolume = value;
                 SoundEffectInstancePool.UpdateMasterVolume();
             }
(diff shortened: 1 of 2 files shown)


MonoGame.Framework/Audio/SoundEffect.cs
```diff
@@ -537,5 +537,10 @@ void Dispose(bool disposing)
 
         #endregion
 
+        internal static void Shutdown()
```
dellis1972: Can we move this static up with the other static methods (like Initialise) so they are in the same place.

CartBlanche: Done.

SimonDarksideJ (approved):

PR #9263, opened 2026-03-20
Add Missing scalar * Matrix operator

dellis1972:
Turns out we were missing the float * Matrix operator on the Matrix class all these years. We have Matrix * float, but not the other. This adds it.

https://learn.microsoft.com/en-us/previous-versions/windows/xna/bb198123(v=xnagamestudio.40)

### Contributor Declaration

### DIFF
--- a/MonoGame.Framework/Matrix.cs
+++ b/MonoGame.Framework/Matrix.cs
@@ -2281,8 +2281,8 @@ namespace Microsoft.Xna.Framework
         /// <summary>
         /// Multiplies the elements of matrix by a scalar.
         /// </summary>
-        /// <param name="scaleFactor">Scalar value on the right of the mul sign.</param>
-        /// <param name="matrix">Source <see cref="Matrix"/> on the left of the mul sign.</param>
+        /// <param name="scaleFactor">Scalar value on the left of the mul sign.</param>
+        /// <param name="matrix">Source <see cref="Matrix"/> on the right of the mul sign.</param>
         /// <returns>Result of the matrix multiplication with a scalar.</returns>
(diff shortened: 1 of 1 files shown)


MonoGame.Framework/Matrix.cs
```diff
@@ -925,10 +925,10 @@ public static void CreatePerspective(float width, float height, float nearPlaneD
...
 		    {
 		        throw new ArgumentException("nearPlaneDistance >= farPlaneDistance");
-		    }
-
-            var negFarRange = float.IsPositiveInfinity(farPlaneDistance) ? -1.0f : farPlaneDistance / (nearPlaneDistance - farPlaneDistance);
-
+		    }
+
+            var negFarRange = float.IsPositiveInfinity(farPlaneDistance) ? -1.0f : farPlaneDistance / (nearPlaneDistance - farPlaneDistance);
+
```
tomspilman, line 931: Seems like lots of whitespace changes.

Should we allow that in this PR @SimonDarksideJ @CartBlanche ?

SimonDarksideJ (approved):

MonoGame.Framework/Matrix.cs
SimonDarksideJ: I think the author has the right intention and overall it looks ok to me, approved @dellis1972

dellis1972: This was edited in the github web editor... not sure why there are so many whitespace changes

dellis1972: @tomspilman  so I've tried to edit this file in multiple editors and every one of the messes up the line endings.

PR #9267, opened 2026-03-24
Add Vulkan and DX12 to Issue template

dellis1972:
Lets add Vulkan and DX12 to the issue template for bug reports

### Contributor Declaration

### DIFF
--- a/.github/ISSUE_TEMPLATE/01_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/01_bug_report.yml
@@ -33,6 +33,8 @@ body:
       - MonoGame Cross-Platform Desktop Application (mgdesktopgl)
       - MonoGame Windows Desktop Application (mgwindowsdx)
       - MonoGame iOS Application (mgios)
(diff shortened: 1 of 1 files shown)


arosolino:
Thank you =)

SimonDarksideJ:
Can we also add the Content Builder as well in that case @dellis1972 ?

dellis1972:
> Can we also add the Content Builder as well in that case @dellis1972 ?

the Content Builder isn't a "Platform"

uygary:
Just my tuppence:
I think it will be good if we can merge this in.
Then we can decide whether we want to handle content pipeline separately, or how to do that in the template.

Since we're already at preview 4, and are trying to squash those pesky bugs, we may want to have the Vulkan and DX12 in the template already.
Every time I open an issue, I end up selecting `N/A` and then editing it manually to copy over the platform string, which I might have forgot at some point.

SimonDarksideJ (approved):

